PR #5762, opened 2018-06-01
Queue | Use disposition ID

dannysellers:
### Description
One of the reasons the bug described in https://github.com/department-of-veterans-affairs/appeals-deployment/pull/1480 went unnoticed is because issue dispositions were not round-trippable. In multiple places on the FE, we were transforming disposition descriptions to fit the context, resulting in a mismatch between what the BE expected and what the FE provided. This obviates the description formatting problem by passing the disposition ID.

To ensure backwards compatibility, this adds a `disposition_id` attribute to Issues, which is passed to Queue as `disposition` within `WorkQueue::IssueSerializer`.

### Acceptance Criteria 
- [ ] Setting Issue disposition during attorney check out persists when viewing issue later

### Testing Plan
1. Go to attorney Draft Decision checkout, set dispositions for issues
2. Complete check out, go to appeal detail view as the assigned judge
3. Confirm dispositions render correctly in IssueListItem


app/repositories/issue_repository.rb
```diff
@@ -75,7 +75,7 @@ def self.load_remands_from_vacols(vacols_id, vacols_sequence_id)
 
   # rubocop:disable Metrics/CyclomaticComplexity
   def self.perform_actions_if_disposition_changes(record, issue_attrs)
-    case issue_attrs[:disposition]
+    case Constants::VACOLS_DISPOSITIONS_BY_ID[issue_attrs[:disposition]]
```
aroltsch: Should we use codes here as well? I think we should rely on the keys from now on.

dannysellers: I dunno, this feels like a place where it might be better to use the text labels for the sake of readability. I get that that's a point against these changes in general, though it looks like this is the only place (outside of constructing legacy test data) where we refer to disposition values specifically—otherwise, we're just passing an ID around.

aroltsch: I am okay with whatever you decide.

Issue #5685, opened 2018-05-25
Move certification tests to FACOLS


mdbenjam:
Certification tests in the following files are failing. We need to move them to FACOLS.

```
spec/feature/certification/cancel_certification_spec.rb
spec/feature/certification/save_certification_spec.rb
spec/feature/certification/start_certification_spec.rb
spec/jobs/start_certification_job_spec.rb
spec/models/certification_spec.rb
spec/models/form8_spec.rb
```

PR #5763, opened 2018-06-01
Rewrite tests for FACOLS, cancel_certification and save_certification…

nemodjuric:
Connects #5685


spec/factories/vacols/case.rb
```diff
@@ -173,6 +173,14 @@
       bfregoff "RO18"
     end
 
+    trait :has_default_regional_office do
+      bfregoff "DSUSER"
+    end
+
+    trait :has_mismatched_form_9 do
```
mdbenjam: I wouldn't make this a trait, I would just override the property in the one place we use it.

spec/factories/vacols/case.rb
```diff
@@ -173,6 +173,14 @@
       bfregoff "RO18"
     end
 
+    trait :has_default_regional_office do
+      bfregoff "DSUSER"
```
mdbenjam: I wouldn't make this a trait, I would just override the property in the one place we use it.

spec/feature/certification/save_certification_spec.rb
```diff
@@ -202,11 +200,18 @@
         expect(form8.certifying_official_name).to eq "Tom Cruz"
         expect(form8.certification_date.strftime("%m/%d/%Y")).to eq Time.zone.today.strftime("%m/%d/%Y")
 
+        uncertify_appeal
         visit "certifications/#{appeal.vacols_id}/sign_and_certify"
         expect(find_field("Name and location of certifying office").value).to eq "Digital Service HQ, DC"
         expect(find_field("Organizational elements certifying appeal").value).to eq "DSUSER"
         expect(find_field("Name of certifying official").value).to eq "Lauren Roth"
 
+        within_fieldset("Title of certifying official") do
```
mdbenjam: Why do we need to add this new field?

spec/feature/certification/save_certification_spec.rb
```diff
@@ -4,31 +4,29 @@
   before do
     Form8.pdf_service = FakePdfService
     Timecop.freeze(Time.utc(2017, 2, 2, 20, 59, 0))
-    allow(Fakes::PowerOfAttorneyRepository).to receive(:update_vacols_rep_name!).and_call_original
+    FeatureToggle.enable!(:test_facols)
+    PowerOfAttorney.repository = PowerOfAttorneyRepository if FeatureToggle.enabled?(:test_facols)
```
lowellrex: What is the purpose of the condition here if we enable the `test_facols` feature flag immediately before this line?

spec/feature/certification/save_certification_spec.rb
```diff
@@ -216,6 +215,7 @@
       end
 
       scenario "Trying to skip steps" do
+        uncertify_appeal
```
lowellrex: Can we just create a new appeal here that is not certified instead of reusing a single appeal and uncertifying it?

spec/feature/certification/save_certification_spec.rb
```diff
@@ -202,6 +200,7 @@
         expect(form8.certifying_official_name).to eq "Tom Cruz"
         expect(form8.certification_date.strftime("%m/%d/%Y")).to eq Time.zone.today.strftime("%m/%d/%Y")
 
+        uncertify_appeal
```
lowellrex: Why do we expect the appeal to not be certified at this point in the test?

spec/feature/certification/start_certification_spec.rb
```diff
@@ -4,85 +4,51 @@
...
-      form9_date: form9.received_at
-    }
+  let(:appeal_ready) do
+    create(:legacy_appeal, vacols_case: vacols_case)
   end
 
-  let(:appeal_ready_exact_match) do
-    Generators::LegacyAppeal.build(vacols_record: vacols_record_exact_match, documents: documents)
+  let(:vacols_case) do
+    create(:case_with_ssoc, bfregoff: "DSUSER", bfdsoc: 181.days.ago)
```
lowellrex: Are we setting "DSUSER" as the user here so this case is assigned to the currently logged in user?

lowellrex: Asking because I ran into the same issue (wanting to see some set of data assigned to the currently logged in user) and solved it a little differently: https://github.com/department-of-veterans-affairs/caseflow/pull/5729/files#r191917505.

If we're trying to do the same thing here, it may be worth looking into making a simple way to set some user as the currently logged in user (maybe in rails_helper).

spec/feature/certification/start_certification_spec.rb
```diff
@@ -111,7 +77,7 @@
       expect(page).to have_title("Check Documents | Caseflow Certification")
       expect(page).to have_content("All documents found with matching VBMS and VACOLS dates.")
       expect(page).to have_content("SOC and SSOC dates in VBMS can be up to 4 days")
-      expect(page).to have_content("SOC 09/10/1987 09/06/1987")
+      expect(page).to have_content("SOC 07/04/2014 07/01/2014")
```
lowellrex: Won't this content change every day? Can we pull these dates from some field in the case's document?

spec/models/certification_spec.rb
```diff
@@ -1,55 +1,59 @@
...
+  end
+
+  let(:vacols_case_mismatch) do
+    create(:case_with_soc, :has_regional_office, bfd19: 3.months.ago)
+  end
+
+  let(:vacols_case_multiple_mismatch) do
+    create(:case, :has_regional_office, bfdnod: 1.month.ago, bfdsoc: 1.month.ago, bfd19: 3.months.ago,
+                                        bfssoc1: 1.month.ago)
+  end
```
lowellrex: We only use a bunch of these object that we create up here once. Can we create them closer in the code to where they end up being referenced? My thinking being that keeping them close makes clearer which variables are used where, and makes refactoring later on easier.

spec/feature/certification/save_certification_spec.rb
nemodjuric: The old test had it this way (not creating a new one) and the reason for that, I believe, was to test if informations was saved and loaded on reload.

spec/feature/certification/save_certification_spec.rb
nemodjuric: You're correct, the condition should be removed. I think I had dual compatibility on mind (working with and without facols) when I wrote it.

spec/feature/certification/save_certification_spec.rb
nemodjuric: Same as above. The old test used the same appeal and it worked because of the fakes setup. To mimic that and use the same appeal having facols on, we have to uncertify it.

spec/feature/certification/start_certification_spec.rb
nemodjuric: Yes, and I agree it can look better. I will modify rails_helper's method authenticate to accommodate user argument, instead of using stub method in spec itself.

spec/feature/certification/start_certification_spec.rb
nemodjuric: These are coming from case factory and will not change. We are creating a record in FACOLS with arbitral values, then faking these into VBMS. You can look more here: https://github.com/department-of-veterans-affairs/caseflow/blob/master/spec/factories/vacols/case.rb#L62

PR #5787, opened 2018-06-04
Add rake tasks to rebuild local dev env in consolidated steps

lowellrex:
To make it easier to rebuild our local development environments I created two rake tasks to destroy and build those environments. We still need to set a local variable but otherwise this pair of commands will rebuild your local development environment when run within the caseflow directory:
```
caseflow $ RAILS_ENV=development bundle exec rake local:destroy
caseflow $ RAILS_ENV=development bundle exec rake local:build
```

### DIFF
--- a/lib/tasks/local/destroy.rake
+++ b/lib/tasks/local/destroy.rake
@@ -1,0 +1,7 @@
+namespace :local do
+  desc "destroy local development environment to have a clean slate when rebuilding"
+  task :destroy do
+    puts "Tearing down docker volumes"
+    `docker-compose down --rmi all -v --remove-orphans`
+  end
+end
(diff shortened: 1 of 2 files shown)


lib/tasks/local/destroy.rake
```diff
@@ -0,0 +1,7 @@
+namespace :local do
+  desc "destroy local development environment to have a clean slate when rebuilding"
+  task :destroy do
+    puts "Tearing down docker volumes"
+    `docker-compose down --rmi all -v --remove-orphans`
```
evankroske: I don't think that --rmi all is necessary, because it destroys the images. Also, I think that this makes rebuilds much slower because Docker has to redownload all the base images.

PR #5793, opened 2018-06-05
Improve build local dev env script

lowellrex:
Continuation of the work initiated in #5787.

Three major changes here:
1. Added step to set up the FACOLS container that our local tests run against.
2. Added logic to wait for FACOLS containers to be ready in this script instead of only relying on other rake tasks's logic to do that for us (that approach failed occasionally).
3. Call rake tasks by way of backticked system commands instead of ruby interface. This allows us to set the environment variables inside the script instead of when we call this task. This has the side benefit of allowing us to copy and paste individual lines into the command line to easily troubleshoot finicky steps.

Prior to this change, to rebuild the local development we would have to run the following sequence of commands:
```
caseflow $ bundle exec rake local:destroy
caseflow $ RAILS_ENV=development bundle exec rake local:build
caseflow $ RAILS_ENV=test bundle exec rake local:vacols:setup
```

After this change, we will be able to rebuild with the following steps:
```
caseflow $ bundle exec rake local:destroy
caseflow $ bundle exec rake local:build
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,12 +122,7 @@ To rapidly set up your local development (and testing) environment, you can run:
 bundle exec rake local:build
 ```
 
-In order to rebuild your local development run that command after tearing down your existing environment with this command:
-```
-bundle exec rake local:destroy
-```
-
-Both of the above shortcuts run a set of commands in sequence that should build (or destroy) your local environment. If you need to troubleshoot the process, you can copy each individual step out of these tasks and run them independently.
+The above shortcut runs a set of commands in sequence that should build your local environment. If you need to troubleshoot the process, you can copy each individual step out of the task and run them independently.
 
 ## Debugging FACOLS setup
 Sometimes the above setup fails, or the app cannot connect to the DB. Here are some frequently encountered scenarios.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -151,6 +151,10 @@ RAILS_ENV=development rake local:vacols:seed
 
 Note you'll need to setup both the test and development databases, but only need to seed the development database.
 
+## Setup shortcuts
```
evankroske: I think it'd be really helpful to also remove the setup commands that your script replaces. This way, we don't have to worry about the documentation going out of date when the script is updates. This also keeps people from updating the docs without updating the script. What do you think?

README.md
```diff
@@ -151,6 +151,10 @@ RAILS_ENV=development rake local:vacols:seed
...
+
+To rapidly set up your local development (and testing) environment, you can run `bundle exec rake local:build`. In order to rebuild your local development run that command after tearing down your existing environment with this command: `bundle exec rake local:destroy`.
```
evankroske: Could you move the commands to their own lines without leading "$"? This would allow devs to select the commands with triple-click instead of clicking and dragging. As it is, it's sort of difficult to find and copy the commands from this paragraph. What do you think?

lowellrex: Smart!

Issue #5569, opened 2018-05-16
AMA Intake | Review: Handle scenario where claimant is different from Veteran


abbyraskinUSDS:
The Form includes the option to list a claimant other than the Veteran. We need a way for users to view associated relationships from the corporate database, so they can select the relevant claimant/appellant when establishing the appeal. At the VBA level, the claimant may need to be added to corporate database (via Share), so we should provide CAs instructions to do so if the Veteran doesn't have any associated relationships yet.

## AC
- Add a required question to the "Enter details" step for all AMA review types
  - "Is the Claimant someone other than the Veteran?"

If yes:
- Show the next question asking to select relationship from BGS (see mock)
- Show radio button list of Veteran relationships from BGS _(_more discovery/definition needed here @shanear_) _
- User can select a name from the list (only one for now) - required field

## Mocks
![image](https://user-images.githubusercontent.com/4042181/40666947-ec99c9a2-632e-11e8-909f-2a3eaba01d09.png)

## Related
#5695 (empty state)

abbyraskinUSDS:
@shanear passing this over to you for review/to add any technical details about the bgs fields.

shanear:
Looks good to me, a couple considerations:

1. Is "Share" capitalized like that? I always thought it was all caps. Like "SHARE".

2. Also we may have access to their relationship to the veteran, if you want to put that on the radio buttons too.

3. Another case we may have to account for is where a user has no relationships in BGS. Might be for another story though.

abbyraskinUSDS:
- I think it actually is "Share" but I'll confirm.
- Yes, that'd be great! I wasn't sure what info we have
- Yeah, i'll make an empty state story.

leikkisa:
@mdbenjam @shanear 

Hey Mark! We are wondering the proposed data structure for storing claimants. We are thinking of making a claimant table with a one to many relationship to reviews (appeals?), so if a claimant was partaking in multiple appeals, they would appear multiple times in the table.  We are thinking of storing the review_id, the participant_id, and perhaps the relationship, and calling on BGS for any other data needed.  Thoughts?

leikkisa:
I'm also wondering if we should hide this feature if there are zero relationships.

leikkisa:
@abbyraskinUSDS When you're back we want to ask you how you want to present the front-end when a veteran has no relationships.

leikkisa:
[...]

PR #5799, opened 2018-06-05
Ability to choose claimant other than the Veteran - Front end

leikkisa:
Connects #5569

Gets Veteran's relationships from BGS, and renders them on the "Review page" in case the claimant is other than the veteran.  This ticket is for Higher Level Review only.


app/models/veteran.rb
```diff
@@ -138,6 +142,12 @@ def fetch_end_products
     self.class.bgs.get_end_products(file_number).map { |ep_hash| EndProduct.from_bgs_hash(ep_hash) }
   end
 
+  def fetch_relationships
+    self.class.bgs.find_all_relationships(
+      participant_id: Veteran.find_or_create_by_file_number(file_number).participant_id
```
shanear: since this method is inside `Veteran`, `participant_id` should be a method you can just call. Instead of refinding the veteran.

client/app/intake/actions/common.js
```diff
@@ -91,6 +91,20 @@ export const setReceiptDate = (receiptDate) => ({
   }
 });
 
+export const setDifferentClaimantOption = (differentClaimantOption) => ({
+  type: ACTIONS.SET_DIFFERENT_CLAIMANT_OPTION,
```
shanear: change to `CLAIMANT_IS_VETERAN`

client/app/intake/constants.js
```diff
@@ -6,6 +6,13 @@ export const PAGE_PATHS = {
   COMPLETED: '/completed'
 };
 
+export const BOOLEAN_RADIO_OPTIONS = [
+  { value: 'false',
+    displayText: 'No' },
+  { value: 'true',
```
shanear: should this be true string or just true?

leikkisa: The Radio components sadly wants strings

client/app/intake/actions/common.js
leikkisa: 👍

app/models/veteran.rb
leikkisa: 👍 , Also yay, I learned how to use emojis in here.

client/app/intake/components/DifferentClaimant.jsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import RadioField from '../../components/RadioField';
+import { BOOLEAN_RADIO_OPTIONS } from '../constants';
+
+export default class DifferentClaimant extends React.PureComponent {
```
shanear: This is another place we should change to VeteranIsClaimant

leikkisa: I like it for the boolean but for the component I went with "SelectClaimant", seems more appropriate to me.

Issue #5786, opened 2018-06-04
Create script to close appeals that should be closed


shanear:
Appeals that we closed with RAMP are being reopened in the field. We need to automatically reclose these. It would also be nice to know the stats on how often it happens.

- Nightly, we should check all appeals closed with RAMP. If any of them have been reopened, or closed with a different disposition, reclose them with a "P"

## Technical concerns
For post-remands, we will need to check the status of the follow up appeal, not the appeal linked via` RampClosedAppeal`. This will require a little bit of extra logic.

Checking in on all appeals closed with RAMP takes quite a long time (several hours), is the load going to be too heavy for a nightly job?

shanear:
@cmgiven thoughts on this story?

cmgiven:
For post-remands, we could also look at the `RampClosedAppeal`. But the key is not its disposition, but rather whether it remains in HIS status as it should. I assume that would be the one more likely to be reopened, if any, rather than the P record.

I assume several hours is based on iterating through each appeal. Should we instead just run a query to find all of the ones that have been re-opened? That would take only seconds.

shanear:
This has been running successfully for the past couple days and appears to be working well.

PR #5804, opened 2018-06-06
Tracer for reclosure of RAMP closed appeals job

shanear:
connects #5786

The plan here is to merge the code that just queries for appeals that we'd mark for reclosure and report to slack what they are. When we verify that that number is correct, we can go ahead and add the code to actually close them.


spec/models/ramp_closed_appeal_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+        RampClosedAppeal.create!(vacols_id: "SHANE1"),
+        RampClosedAppeal.create!(vacols_id: "SHANE2"),
+        RampClosedAppeal.create!(vacols_id: "SHANE3")
+      ]
+    end
+
+    before do
+      expect(AppealRepository).to receive(:find_ramp_reopened_appeals)
+        .with(%w[SHANE1 SHANE2 SHANE3])
+        .and_return(["STUFF"])
```
leikkisa: Haha

app/models/ramp_closed_appeal.rb
```diff
@@ -2,4 +2,15 @@ class RampClosedAppeal < ApplicationRecord
   belongs_to :ramp_election
 
   delegate :established_at, to: :ramp_election
+
+  def self.reclose_all!
+    appeals_to_reclose = []
+
+    find_in_batches(batch_size: 800) do |batch|
+      appeals_to_reclose += AppealRepository.find_ramp_reopened_appeals(batch.map(&:vacols_id))
```
leikkisa: I don't understand how reclose_all gets the vacols_id but I think that's just me! Also, I have no idea how I turned this yellow :-P

cmgiven: Check out: http://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_in_batches

Did not know this existed, but pretty handy.

Shane, is there a reason to drop batch size to 800? Confirmed that Oracle supports lists of up to 1,000 elements.

app/services/appeal_repository.rb
```diff
@@ -274,6 +274,19 @@ def self.location_after_dispatch(appeal:)
     "98"
   end
 
+  # Finds appeals in the set of vacols_ids passed that have been reopened after
+  # being closed for RAMP
+  def self.find_ramp_reopened_appeals(vacols_ids)
+    VACOLS::Case
+      .where(bfkey: vacols_ids)
+      .where.not(bfdc: "P")
```
cmgiven: This line is duplicative of 284.

app/models/ramp_closed_appeal.rb
shanear: Just to be safe i guess. It doesn't really slow it down much

app/services/appeal_repository.rb
shanear: good catch

Issue #5809, opened 2018-06-06
Record 'source' header for API calls


cmgiven:
## AC

* When an API call includes a `source` header, it should be recorded in the `ApiView`.

laurjpeterson:
looks like a PR was merged here, so closing.

PR #5810, opened 2018-06-06
Add source to ApiViews

cmgiven:
Connects #5809

### Description
Adds a `source` column to `ApiViews` to record a header of the same name if present. This is necessary as Vets.gov will soon be passing through the API to other consumers.

### Acceptance Criteria 
- [ ] `source` is recorded in `ApiViews` table if present in an API call.
- [ ] `source` is null if not present.


cmgiven:
Yes. There's some back-and-forth happening with the vets-api team. Hope to have clarity soon.

Issue #5805, opened 2018-06-06
Re-factor judge case assignment to attorney logic into its own class.


aroltsch:
1. Re-factor judge case assignment logic into its own class
2. Move legacy task ID validation into a shared place

PR #5811, opened 2018-06-06
Re-factor judge case assignment logic into its own class and move tas…

aroltsch:
connects #5805 

### Description
This PR doesn't change existing functionality. It contains refactors to remove duplication around task_id validation and creates a better name for judge case assignment logic to attorneys. FE is not affected by any of these changes.


app/models/queues/attorney_queue.rb
```diff
@@ -1,9 +0,0 @@
-class AttorneyQueue < WorkQueue
```
aroltsch: This file can be deleted as it is not being used.

app/models/legacy_tasks/attorney_legacy_task.rb
```diff
@@ -1,63 +1,7 @@
...
-  validates :assigned_by, :assigned_to, presence: true
-  validate :assigned_by_role_is_valid
-
   def self.from_vacols(case_assignment, user_id)
     task = super
     task.assigned_at = case_assignment.assigned_to_attorney_date
     task
   end
-
-  def self.create(task_attrs)
```
aroltsch: Moved all this logic into its own class: JudgeCaseAssignmentToAttorney

app/models/attorney_case_review.rb
```diff
@@ -13,15 +15,13 @@ class AttorneyCaseReview < ApplicationRecord
 
   attr_accessor :issues
 
-  # task ID is vacols_id concatenated with the date assigned
-  validates :task_id, format: { with: /\A[0-9A-Z]+-[0-9]{4}-[0-9]{2}-[0-9]{2}\Z/i }
```
aroltsch: Moved all this logic around task_id  into a shared concern.

PR #5813, opened 2018-06-06
Add appeal closure logic

shanear:
connects #5786 

## The Plan
1. Wait on the Tracer (https://github.com/department-of-veterans-affairs/caseflow/pull/5804) to land in prod and start giving us slack updates on count of appeals
2. Merge (protected by feature flag), and give this script a test run on a single UAT appeal and verify it's working okay (@cmgiven to assist)
3. Pop open the feature flag on master and let'er fly


app/models/ramp_closed_appeal.rb
```diff
@@ -3,14 +3,49 @@ class RampClosedAppeal < ApplicationRecord
 
   delegate :established_at, to: :ramp_election
 
+  def reclose!
```
cmgiven: This is really limited to reclosing an undecided appeal. I suggest we make that explicit by adding an additional check at the top of the method to check that the disposition on the appeal is not a BVA disposition (a number between 1 and 9) and is a field disposition (a letter between A and Z).

PR #5815, opened 2018-06-07
Allow different claimant from veteran (Part 2)

leikkisa:
Connects #5569 

- Add to other intake flows
- Save data in back end


leikkisa:
This has a bunch of commits from Part 1 listed because I git branched from that branch.  The real commits for this PR start with "Add claimant to other flows".

Issue #4716, opened 2018-02-24
Judges | Assign cases | Select and assign multiple cases to an attorney


laurjpeterson:
Judges will have ~20-30 cases in their 'Unassigned' queue, and can assign any number of them to attorneys on their team. Judges can select different attorneys per case, and those cases will be assigned when the judge clicks the 'Assign' button.

When judges are assigning cases to their attorneys, they need to choose which attorney on their team they should assign the case to, so that attorney can draft an appeals decision.

### Acceptance Criteria
- Verify that a column 'Assign to' exists in the 'Assign Y Cases' table
- Verify that a dropdown exists for each unassigned case in the 'Assign to' column
- Verify that the dropdown only contains attorneys that are on their team, plus an 'Other' option
-  Verify that an 'Assign' button exists on the top and bottom of the judges Assign tables, per mocks
- Verify these buttons are present no matter what table the judge is looking at (i.e. Unassigned cases or one of their attorney's queues)

### Mocks

https://app.mural.co/t/workqueue2001/m/workqueue2001/1524171664613/04929efd3dea35587ae9168016d56fa5e5870613

https://va-gov.invisionapp.com/d/main/#/console/13735727/286320006/preview

evankroske:
## Estimates
- [ ] Add button to judge assign view to assign to attorneys on the judge's team: 6 hours.
- [ ] Extend assignment widget to support assigning to anybody: 8 hours.

evankroske:
I think that this issue was resolved by https://github.com/department-of-veterans-affairs/caseflow/pull/6017 , but I could be wrong. Feel free to reopen this issue if you don't think it's done.

PR #5816, opened 2018-06-07
Refactors Redux state

evankroske:
Part of https://github.com/department-of-veterans-affairs/caseflow/issues/4716

### Description
The purpose of this PR is to move tasks data to a vacolsId-indexed object in the Redux state to make it easy to update them with new data. Specifically, I need to add data from the DECASS records to tasks when they are initially assigned to attorneys.

### Testing Plan
- [x] Log in as BVAAABSHIRE.
- [x] On /queue, confirm that tasks appear.
- [x] After clicking "assign tasks", confirm that tasks appear.
- [x] Log in as BVARERDMAN.
- [x] On /queue, click "assign tasks".
- [x] Click on the attorney with tasks.
- [x] Confirm that tasks are shown.
- [x] Click on an appeal to view case details.
- [x] Confirm that case info is rendered.


client/app/queue/AddEditIssueView.jsx
```diff
@@ -335,14 +335,16 @@ AddEditIssueView.propTypes = {
...
-  task: state.queue.loadedQueue.tasks[ownProps.vacolsId],
-  issue: state.queue.editingIssue,
-  error: state.ui.messages.error,
-  modal: state.ui.modal.deleteIssue
-});
+const mapStateToProps = (state, ownProps) => {
+  return {
+    highlight: state.ui.highlightFormItems,
+    appeal: state.queue.stagedChanges.appeals[ownProps.vacolsId],
+    task: state.queue.tasks[ownProps.vacolsId],
```
lowellrex: Wow. This is so much more intuitive. Can't believe we haven't made this change before.

client/app/queue/reducers.js
```diff
@@ -7,13 +7,14 @@ import { ACTIONS } from './constants';
...
 import caseListReducer from './CaseList/CaseListReducer';
-import uiReducer from './uiReducer/uiReducer';
+import * as uiReducer from './uiReducer/uiReducer';
 
 // TODO: Remove this when we move entirely over to the appeals search.
 import caseSelectReducer from '../reader/CaseSelect/CaseSelectReducer';
 
 export const initialState = {
   judges: {},
+  tasks: {},
```
dannysellers: Do we have a plan for reconciling this with `loadedQueue.tasks`? I'm all for moving away from `loadedQueue`, but it seems like the two serve similar purposes at this point, they're both tasks assigned/"assigned" to the user, populating their main view (`JudgeReviewTaskTable`/`AttorneyTaskTable`)

client/app/queue/JudgeReviewTaskTable.jsx
```diff
@@ -73,12 +73,20 @@ class JudgeReviewTaskTable extends React.PureComponent {
   });
 
   render = () => {
+    const { loadedQueueTasks, appeals, tasks } = this.props;
+    const taskWithId = {};
```
dannysellers: `tasksWithId`? `tasksById`?

client/app/queue/JudgeReviewTaskTable.jsx
```diff
@@ -73,12 +73,20 @@ class JudgeReviewTaskTable extends React.PureComponent {
...
+
+    for (const id of Object.keys(loadedQueueTasks)) {
```
dannysellers: I'm not sure I understand why we're building `taskWithId` from two sources like this. Both `tasks` and `loadedQueueTasks` contain the entire task objects. We could just do `const tasksById = _.keyBy(tasks, 'id')`

client/app/queue/AssignedCasesPage.jsx
```diff
@@ -48,6 +52,14 @@ const AssignedCasesPage = (props) => {
   </React.Fragment>;
 };
 
+const mapStateToProps = (state) => {
+  const { tasksAndAppealsOfAttorney, attorneysOfJudge, tasks } = state.queue;
+
+  return { tasksAndAppealsOfAttorney,
+    attorneysOfJudge,
+    tasks };
```
lowellrex: Do we expect the data for a given task in `state.queue.tasksAndAppealsOfAttorney[attorneyId].data.tasks` to be different from the data for that same task in `state.queue.tasks`?

evankroske: The idea is to change state.queue.tasksAndAppealsOfAttorney[attorneyId].data.tasks to be a list of task IDs of tasks assigned to the attorney. This way, we won't have to update task data in two places. However, this is an intermediate step in which we keep the task data in tasksAndAppealsOfAttorney but don't touch it. The Flow PR enforces this at type-check-time. I could make that change now, but I'd have to make it in the Flow PR and strip the types from it again, because I don't trust myself to not make a mistake.

client/app/queue/JudgeReviewTaskTable.jsx
evankroske: Since there's only one task with a given ID, I wanted to use a singluar task in the variable name. I'm OK with taskById, but I don't want to make tasks plural. If this were a map from IDs to list of tasks, I'd call it tasksWithId. What do you think?

client/app/queue/JudgeReviewTaskTable.jsx
evankroske: The difference between tasks and loadedQueueTasks is that loadedQueueTasks only contains the tasks that are assigned to the current user. tasks stores data for all tasks fetched from the backend. By storing all task data in tasks, it's easy to update data for a task whether it's assigned to the current user or not. Does this make sense?

client/app/queue/reducers.js
evankroske: I'm planning to replace loadedQueue.tasks with a list of task IDs assigned to the current user. I'm trying to move all task data to tasks. This is an intermediate step in which I've configured Flow to only allow developers to access the ID property of tasks in loadedQueue.tasks. The next step is to replace it with an array. Sound good?

client/app/queue/JudgeReviewTaskTable.jsx
dannysellers: `tasksById` to me says it's a collection of tasks indexed by their ID, a unique identifier. Keeping task singular seems like it only represents one task, rather than something to be iterated over. `tasksWithId` sounds a bit redundant—each task already has an ID attr inside/within it, but they may not be ordered or indexed in any particular way.

client/app/queue/JudgeReviewTaskTable.jsx
dannysellers: I see, yeah. I guess I didn't realize this is a filter operation—`tasks` and `loadedQueueTasks` seem pretty similar semantically. Could we rename `loadedQueueTasks` something like `currentUserTasks` or `currentUsersTasks` or something? Or, replace this with a filter function like
```javascript
const tasksById = _.reduce(loadedQueueTasks, (result, value, key) => {
  if (key in tasks) {
    result[key] = tasks[key];
  }

  return result;
}, {});
```
Or, just add a comment explaining this is a filter operation?

Issue #5744, opened 2018-05-31
Case details | Case snapshot | About the case


laurjpeterson:
With the new Case details design, we've introduced a case snapshot that gives the user a quick view of information about the case, user-specific task information (e.g. admin action type and instructions for co-located, and allows the user to start certain actions. 

This Case snapshot section will look different depending on the user and task associated with it. 

### All users and scenarios
- Verify that there is an 'About the case' section left aligned in the Case snapshot section.
- 'About the case' section includes:
   - Type: e.g. AOD, CAVC, original, post remand - moved from the former location in former case details
   - Docket number
- Currently assigned to: information - @mkhandekar to confirm location of this in mocks

### If a user navigates to case details and the case is assigned to them
- 'About the case' section also includes:
   - Calendar days waiting = calendar days the case has been assigned to them / in their queue

## Out of scope for this issue
- Actions
  - e.g. Judges assigning to attorneys in Case details - https://github.com/department-of-veterans-affairs/caseflow/issues/5303

## Mocks
![image](https://user-images.githubusercontent.com/1598889/40918784-5ad1fb6a-67d5-11e8-8bd7-68332104bdc0.png)

## Resources
- [Case details Mural](https://app.mural.co/t/workqueue2001/m/workqueue2001/1525112874684/5765d285f568b5f631990e2c8e514a39d9b6b570)

lowellrex:
@mkhandekar I'm curious to get your thoughts on this first pass I made at the case snapshot element:

| Before | After |
| ------ | ----- |
| ![image](https://user-images.githubusercontent.com/32683958/40866721-55917e00-65cd-11e8-9243-69647a6bd75d.png) | ![image](https://user-images.githubusercontent.com/32683958/40866680-0dba32a2-65cd-11e8-84ce-03bc36cc0f7c.png) |

A few notes:
* I did not include the "task instructions" item because I don't know where we draw that from. I think that might be a new AMA task field, but am happy to add it back if it should apply to legacy VACOLS tasks as well.
* From what I could tell from the mocks the proportional width of the various sections were 1/4, 1/2, 1/4. The "medical request ready for review" action in the actions dropdown wrapped onto a second line when the container was limited to 1/4 width so I changed the proportions to 1/4, 1/4, 1/2. If we are going to leave the "task instructions" item out of the middle section I think this works fine, but I am definitely open to ideas about how this should look.
* I have not yet added the "view diary" link.
[...]

PR #5837, opened 2018-06-07
Add case snapshot to case details page

lowellrex:
Connects #5744.

Moves some information about the appeal to a highlighted box at the top of the page that we are calling Case Snapshot.

| Before | After |
| ------ | ----- |
| image | image |

Right now the application and presentation code is very intertwined and I am definitely open to suggestions for how we could untangle the two. I took a few stabs at doing just that but couldn't come up with a solution that I particularly liked. If we can't think of anything better here then I think it is fine to wait to separate this component out into app/presentation pieces until there are other use cases when the improvements may become more clear.


client/app/queue/AppealDetail.jsx
```diff
@@ -92,9 +91,6 @@ export default class AppealDetail extends React.PureComponent {
 
   getListElements = () => {
     const listElements = [{
-      label: 'Type(s)',
-      value: renderAppealType(this.props.appeal)
-    }, {
```
lowellrex: Moved to CaseSnapshot.

client/app/queue/QueueDetailView.jsx
```diff
@@ -45,52 +41,6 @@ class QueueDetailView extends React.PureComponent {
...
-      loadedQueueAppealIds
-    } = this.props;
-
-    if (featureToggles.phase_two && loadedQueueAppealIds.includes(appeal.attributes.vacols_id)) {
-      return <SelectCheckoutFlowDropdown vacolsId={vacolsId} />;
-    }
-
-    return null;
-  }
-
```
lowellrex: Moved to CaseSnapshot.

client/app/queue/QueueDetailView.jsx
```diff
@@ -104,8 +54,7 @@ class QueueDetailView extends React.PureComponent {
           <React.Fragment>View <AppealDocumentCount appeal={this.props.appeal} /> documents</React.Fragment>
         } />
     </CaseTitle>
-    <p className="cf-lead-paragraph" {...subHeadStyling}>{this.subHead()}</p>
-    {this.getCheckoutFlowDropdown()}
```
lowellrex: Moved to CaseSnapshot.

client/app/queue/CaseSnapshot.jsx
```diff
@@ -0,0 +1,132 @@
...
+      <dt>{COPY.CASE_SNAPSHOT_TASK_ASSIGNEE_LABEL}</dt><dd>{this.props.appeal.attributes.location_code}</dd>
+    </React.Fragment>;
+
+    if (!this.props.task) {
+      return assignedToListItem;
+    }
+
+    const task = this.props.task.attributes;
+
+    if (this.props.userRole === 'Judge') {
```
dannysellers: We now have these in constants as [`USER_ROLES`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/constants.js#L117)

client/app/queue/CaseSnapshot.jsx
```diff
@@ -0,0 +1,132 @@
...
+const aboutHeadingStyling = css({
+  marginBottom: '0.5rem'
+});
+
+export default class CaseSnapshot extends React.PureComponent {
+  daysSinceTaskAssignmentListItem = () => {
+    if (this.props.task) {
+      const today = new Date();
+      const dateAssigned = new Date(this.props.task.attributes.assigned_on);
+      const dayCountSinceAssignment = Math.round(Math.abs((today - dateAssigned) / (24 * 60 * 60 * 1000)));
```
dannysellers: Nit: this might be cleaner/more maintainable using moment's [`diff`](https://momentjs.com/docs/#/displaying/difference/):
```javascript
const today = moment();
const dateAssigned = moment(this.props.task.attributes.assigned_on);
const dayCountSinceAssignment = today.diff(dateAssigned, 'days');
```

client/app/queue/QueueDetailView.jsx
```diff
@@ -43,56 +39,15 @@ class QueueDetailView extends React.PureComponent {
...
-    return null;
-  }
-
   render = () => <AppSegment filledBackground>
     <CaseTitle appeal={this.props.appeal} vacolsId={this.props.vacolsId} redirectUrl={window.location.pathname} />
-    <p className="cf-lead-paragraph" {...subHeadStyling}>{this.subHead()}</p>
-    {this.getCheckoutFlowDropdown()}
+    <CaseSnapshot
+      appeal={this.props.appeal}
+      featureToggles={this.props.featureToggles}
```
dannysellers: I just noticed this... Why pass featureToggles as a prop? We moved them into Redux to avoid passing these kinds of global contextual data down to child components

lowellrex: Didn't know that we had moved them to the global state. I'm working on some more case details stuff and will pull it out in one of those PRs.

dannysellers: Actually, it looks like all of these props could be fetched directly from Redux?

PR #5326, opened 2018-04-27
Consolidate task and appeal details views

lowellrex:
Connects #5309.

PR #5840, opened 2018-06-08
Extracts activeAppeal from state correctly

evankroske:
Fixes https://github.com/department-of-veterans-affairs/caseflow/pull/5326/files#r193836098.

### Description
Fixes an issue that caused the prop activeAppeal of QueueLoadingScreen to always be undefined.


shanear:
stripping ready to merge from this temporarily. Will add it back

PR #5841, opened 2018-06-08
Set open timeout and read timout to longer time

aroltsch:
Fixes https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/1536/

We've been seeing a lot of HTTPClient::ReceiveTimeoutError alerts. 

:open_timeout is the timeout for opening the connection. This is useful if you are calling servers with slow or shaky response times.
:read_timeout is the timeout for reading the answer. 

I set the timeout options to the same value we set for connect_vbms and connect_vva


amprokop:
two thumbs up 👍 👍

PR #5845, opened 2018-06-08
Make CaseTitle self contained

lowellrex:
As we start using CaseTitle in more places in the code it becomes increasingly clear that we pass the same components into CaseTitle. This PR moves those elements inside the CaseTitle component (and create a new CaseTitleScaffolding component) so we only have to pass the bare minimum set of elements to the component for it to display correctly.

### DIFF
--- a/client/app/queue/CaseTitle.jsx
+++ b/client/app/queue/CaseTitle.jsx
@@ -1,7 +1,11 @@
 import { css } from 'glamor';
+import PropTypes from 'prop-types';
 import React from 'react';
 
+import AppealDocumentCount from './AppealDocumentCount';
+import { CATEGORIES } from './constants';
 import { COLORS } from '../constants/AppConstants';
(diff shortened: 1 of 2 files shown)


client/app/queue/CaseTitle.jsx
```diff
@@ -32,6 +36,29 @@ const horizontalRuleStyling = css({
...
+  }
+}
+
+CaseTitle.propTypes = {
+  appeal: PropTypes.object.isRequired,
+  redirectUrl: PropTypes.string.isRequired,
+  vacolsId: PropTypes.string.isRequired
+};
+
+class CaseTitleScaffolding extends React.PureComponent {
```
dannysellers: I'm not sure what kind of logic we'll want inside `CaseTitle` going forwards, but it looks like both of these components could be made into functions as they are now.

Issue #5385, opened 2018-05-03
Judge queue | Handle Acting Judge case


amprokop:
Some attorneys were previously acting judges. An acting judge is an attorney who had applied for the privileges to hear Veteran cases at a hearing and therefore decide/sign those decisions. They do not write the decisions for these cases. Instead, another attorney writes them. The frequency of acting judges has decreased over time.

These attorneys need a way to toggle between the Judge view of queue and Attorney view.

- There are 373 appeals pending where a hearing was held by an active attorney in an acting VLJ capacity. 
- There are 57 different attorneys who held hearings for those appeals. 
- There are roughly ~400,000 active appeals.
- So, this would affect 57 out of ~700 attorneys (soon to be ~1000). 

## AC
- Attorneys who also have the judge role in VACOLS should see a role toggle in their queue page
- Attorneys who also have the judge role in VACOLS should default to the attorney role
- When the attorney toggles to the judge view, they should see the judge Review cases (x) | Assign cases (y) view

## Mocks
To be reviewed by team before development can begin

#### Attorney View
![queue_actingjudge_attorney_inactive](https://user-images.githubusercontent.com/31224257/41553051-a0148c9a-72fe-11e8-9a63-36670f9fe1ec.png)

#### Switching to Acting Judge via toggle
![queue_actingjudge_attorney-toggle](https://user-images.githubusercontent.com/31224257/41553052-a027d930-72fe-11e8-8895-1ca84a59394a.png)

#### Judge View
![queue_actingjudge_inactive](https://user-images.githubusercontent.com/31224257/41553053-a03da4a4-72fe-11e8-9f01-2a994686087a.png)

#### Switching to Attorney via toggle
![queue_actingjudge_judge-toggle](https://user-images.githubusercontent.com/31224257/41553054-a0962ea8-72fe-11e8-9d8f-84bc53802add.png)

### Discusion  5/2
- aiming for the simplest option --> toggling between roles will be the easiest from the back end
- @aroltsch will pull the number of users in VACOLS that are both acting judges and attorneys
- @sneha-pai will explore a role toggle pattern
- goal: solve for acting judges by the time we roll out to judges (aiming for week of 5/14/2018)

evankroske:
Personally, I think that a regular dropdown would be the most clear:
```
Role: | Acting Judge |
      | Attorney     |
```

laurjpeterson:
I agree with the dropdown!

amprokop:
[...]

PR #5862, opened 2018-06-11
Queue | Respect sys admin privileges

dannysellers:
### Description
Currently within Queue, we use the [`can_access_queue?`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/application_controller.rb#L90) helper method to grant access. However, this function only consider's a users VACOLS role, disregarding any permissions set with caseflow-commons' Functions.

In FACOLS environments, we give [all users System Admin privileges](https://github.com/department-of-veterans-affairs/caseflow/blob/master/lib/tasks/local/vacols.rake#L100), so we restrict this check to the production environment.

### Acceptance Criteria 
- [ ] System Admin user who is not an Attorney can access Queue as an Attorney

### Testing Plan
1. Go to Caseflow production as an admin user, confirm Attorney table view appears

TODO: Link the Acting Judge role toggle (#5385) with this functionality

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -88,6 +88,7 @@ class ApplicationController < ApplicationBaseController
   helper_method :certification_header
 
   def can_access_queue?
+    return true if current_user.admin?
     role = current_user.vacols_role
     return true if role == "Attorney"
     return true if role == "Judge" && feature_enabled?(:judge_queue)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -88,7 +88,7 @@ class ApplicationController < ApplicationBaseController
   helper_method :certification_header
 
   def can_access_queue?
-    return true if current_user.admin?
+    return true if current_user.admin? && Rails.env.production?
     role = current_user.vacols_role
     return true if role == "Attorney"
     return true if role == "Judge" && feature_enabled?(:judge_queue)
--- a/app/views/queue/index.html.erb
+++ b/app/views/queue/index.html.erb
@@ -2,7 +2,7 @@
   <%= react_component("Queue", props: {
     userDisplayName: current_user.display_name,
     userId: current_user.id,
-    userRole: current_user.vacols_role,
+    userRole: current_user.admin? ? "Attorney" : current_user.vacols_role,
     userCssId: current_user.css_id,
     userCanAccessQueue: can_access_queue?,
     dropdownUrls: dropdown_urls,
--- a/app/views/queue/index.html.erb
+++ b/app/views/queue/index.html.erb
@@ -2,7 +2,7 @@
   <%= react_component("Queue", props: {
     userDisplayName: current_user.display_name,
     userId: current_user.id,
-    userRole: current_user.admin? ? "Attorney" : current_user.vacols_role,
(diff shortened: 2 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -88,6 +88,7 @@ def certification_header(title)
   helper_method :certification_header
 
   def can_access_queue?
+    return true if current_user.admin? && Rails.env.production?
```
lowellrex: Why do we check for the environment? Why not just check if the current user is an admin?

dannysellers: This would mess up the development environment, where [everyone's an admin](https://github.com/department-of-veterans-affairs/caseflow/blob/master/lib/tasks/local/vacols.rake#L100)

evankroske: Personally, I think that we should change the permissions of users in the development environment instead of changing the permission-checking logic to depend on the environment. In my opinion, the code shouldn't be aware of which environment it's running in, as this logic allows bugs to appear in one environment that don't appear in other environments. I think this makes bugs hard to reproduce. Also, the 12 factor app framework discourages the use of environments: https://12factor.net/config

dannysellers: I definitely agree that the code should in general operate independent of the environment, though this felt cleaner than trying to change FACOLS data. Also, the scope of this change is limited enough that it doesn't seem like we're making a major fork in how the site operates—we're simply allowing production admins to access the site with their given privileges.

If we do decide to alter local data, how should we decide which user(s) are not sys admins in FACOLS? Our impersonation logic isn't complex enough to support checking permissions for one user while loading data for another, so we'd have to be pretty deliberate about whose permissions we change. I suppose I would also need help altering the FACOLS source data in UAT, such that the change persists.

Another thing to note is that this change will not be permanent. Eventually, we'll implement a [toggle for acting judge attorneys](https://github.com/department-of-veterans-affairs/caseflow/issues/5385) to switch between which role they're using—when that time comes, we can just display the toggle for all acting judges, as well as admin users.

evankroske: If I'm understanding this PR correctly, its purpose is to grant system admins (us) permission to access queue in UAT and Prod. However, as of two weeks ago, we couldn't access queue in Prod because we don't have STAFF records in Prod. Do you know if that's still an issue? Did I misunderstand the purpose of this PR?

amprokop: @evankroske and @lowellrex raise some complications.

@dannysellers, I thought about this a little. Could we test the case details page with paper cases if we instead just used the case search page to get to these cases, and avoid this complication? Let's chat tomorrow about this, curious as to your thoughts. Thanks!

dannysellers: As of this morning, we [deployed a change](https://dsva.slack.com/archives/C7VNGMYUQ/p1528810222000342) to enable the Case Details link for paper cases within table views (updated the `<CaseDetailsLink/>` component #5864).

The issue here is for users who are admins but have no VACOLS role (like Lauren). I don't know whether/how that correlates to the VACOLS STAFF records, or what it means to not have them in prod (we fetch STAFF records to find all judges for attorney checkout), but this just came about because we couldn't get into Queue last week to investigate the paper case detail issue.

dannysellers:
Closing this, not sure we need it at this point

Issue #5832, opened 2018-06-07
Change appearance of reader link when user does not have access


lowellrex:
We attempt to show a count of documents associated with an appeal whenever we display a link from an appeal to reader. We load the page from the initial request, then make an additional request to get the count of documents, and add that number in the text of the link when we get a successful response from that request.

| Before request | After response |
| -------------- | -------------- |
| ![image](https://user-images.githubusercontent.com/32683958/41122021-261a1bf6-6a68-11e8-9336-a383af87a9f4.png) | ![image](https://user-images.githubusercontent.com/32683958/41121903-ba533326-6a67-11e8-890f-5fac107ee420.png) |

If that request fails, we continue to show the text as it appears before we make the request. That is, we don't make any indication that the request to get the document count failed (other than not displaying the number in the text of the link). This is a good default behaviour in most cases since the backend could error for any number of legitimate reasons (timeouts... and other weird errors I guess), but when we know why the request resulted in a non-successful response we should indicate that as is appropriate.

[One such case](https://github.com/department-of-veterans-affairs/caseflow/pull/5830) is when somebody is viewing a restricted access case (the Veteran for the case could be a VA employee, a celebrity, high ranking military officials, etc.) that is not assigned to them . In this case the server will respond with a 403 Forbidden and if that person were to click on the link to reader they would be presented with an "Additional access needed" screen. We should instead indicate that they cannot view the documents on the front-end some way. Perhaps we could disable the link and display a message that indicates they do not have the requisite level of access.

## Acceptance criteria
* Indicate that additional access is needed to view documents when `AppealDocumentCount`'s request to `/appeals/{vacols_id}/document_count` returns a 403

lowellrex:
cc @mkhandekar @sneha-pai @amprokop

mkhandekar:
Nice! I like the default fall back. Of the top of anyone's head, what are other examples of data that lands in this category -- _useful to know that X exists, but either do not have access to see or edit it_?

Maybe:
- edit things in case details?

lowellrex:
Quick mock of an option for how this could be displayed:
![image](https://user-images.githubusercontent.com/32683958/41123119-a0621618-6a6b-11e8-8a48-609ee03d0ff2.png)

mkhandekar:
Also, is there something concrete we can tell the user on how to get more access?

lowellrex:
Great question that I don't know the answer to! Maybe just go with the default response of... "talk to your ISO"?

[...]

PR #5863, opened 2018-06-11
Ignore ClientRequestError when hitting document_count endpoint

lowellrex:
Similar to #5830. Example of an instance where #5832 would be beneficial as well.

This change 1) prevents Sentry from sending an alert to slack and 2) returns a 400 response instead of a 500. We are hitting this error because we make a request to the `/appeals/{vacols_id}/document_count` endpoint for an appeal and BGS (by way of efolder) cannot find the Veteran associated with that appeal.

Confirmed this is a legitimate response by bypassing this BGS check and requesting the documents for this Veteran's file number directly from VBMS (v1 and v2 endpoints) and the file number does not exist in VBMS either. Perhaps this indicates that the case is a paper case? (cc @dannysellers)

Link to [the slack discussion](https://dsva.slack.com/archives/C6E41RE92/p1528720066000336) with additional thoughts on the approach.


dannysellers:
[Link](https://github.com/department-of-veterans-affairs/dsva-vacols/issues/11) to discussion on identifying paper cases. The function VACOLS uses is [here](https://github.com/department-of-veterans-affairs/caseflow/blob/master/local/vacols/vacols_copy_5_functions.sql#L1556):
```sql
BEGIN
select count(*) into pcnt from FOLDER where ticknum = folder
  and (tisubj2 = 'Y' or tivbms = 'Y') ;
if pcnt > 0 then
   paperless := 'Y';
else
   paperless := 'N';
end if;

RETURN paperless;
END;
```

Issue #5776, opened 2018-06-04
Work needed to support multiple ramp elections


kahlouie:
# Supporting multiple ramp elections

## What is the problem?
Some users may still have claims being settled after their opt-in to RAMP. This leaves some of their appeals as legacy appeals, and some as converted appeals. We're aiming to support all of their appeals through RAMP.

## What we learned:
In our discovery, we determined that it is possible to support multiple ramp elections. There will need to be some modifications to the code base to support this. Please note that while we are supporting multiple ramp elections. We are NOT supporting multiple ramp refilings. Each veteran will only be entitled to one ramp refiling at the moment, regardless of how many ramp elections they have. The reason we don't intend to support multiple ramp refilings yet is that this will add undue complications at this moment.

### app/models/ramp_election_intake.rb
   In `#find_or_build_initial_detail`, if we find existing ramp election and it has an active EP, add this intake to existing ramp election. We might even want to move the `#matching_ramp_election` up into this method as it's the only place it'll be used.
 
 
   In `#validate_detail_on_start`, remove the if clause `matching_ramp_election.established?` as we should now support extending existing ramp elections and supporting multiple ramp elections if the EP on the existing ramp elections are no longer active.
 
   `#veteran_ramp_elections` may need to be deleted or moved up into `#find_or_build_initial_detail`.
   
### app/models/ramp_refiling.rb

  Take out the assumption that a ramp refiling belongs to a single ramp election. This will involve data model changes as well, though. We will either need to support storing multiple ramp elections in ramp refiling or remove altogether and base it on the veteran_file_number.
 
  In `#election_receipt_date`, we should choose the earliest receipt date of the ramp_elections.
 
  We'll need to modify `#validate_receipt_date` and `#validate_option_selected` to handle multiple ramp elections.

### app/models/ramp_refiling_intake.rb

  We'll need to modify `#preload_intake_data` and `#validate_detail_on_start` to handle multiple ramp elections.

  In `#validate_detail_on_start`, the first if clause should check if ramp_elections is empty. The second should check that there aren't any active ramp elections. The third one needs to make sure cleared ramp elections have no issues. We should continue ensuring there's only a single ramp_refiling per veteran ever (at the moment).

  In `#ramp_refiling_already_processed?`, we'll need to check for the ramp refilings by veteran file numbers instead of by ramp election id.
 
  In `#initial_ramp_refiling`, don't set the ramp election on the ramp refiling.
 
[...]

PR #5869, opened 2018-06-11
5776 remove election from refiling code

kahlouie:
Connects #5776 
part 2/3


spec/factories/ramp_election.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :ramp_election do
+    sequence(:veteran_file_number, &:to_s)
```
shanear: We may want to generate a vet in BGS for this. But this is cool.

spec/models/ramp_refiling_intake_spec.rb
```diff
@@ -17,16 +17,27 @@
...
+           notice_date: 4.days.ago,
+           receipt_date: 3.days.ago,
+           end_product_reference_id: Generators::EndProduct.build(
+             veteran_file_number: veteran_file_number,
+             bgs_attrs: { status_type_code: "CLR" }
+           ).claim_id,
+           established_at: Time.zone.now)
+  end
+
+  let(:second_completed_ramp_election) do
```
shanear: ooh nice 😃

spec/models/ramp_refiling_intake_spec.rb
```diff
@@ -133,12 +158,64 @@
...
 
           context "a saved RampRefiling already exists for the veteran" do
-            let!(:preexisting_ramp_refiling) do
-              RampRefiling.create!(
-                veteran_file_number: "64205555",
-                ramp_election: ramp_election
-              )
+            let!(:preexisting_ramp_refiling) { RampRefiling.create!(veteran_file_number: veteran_file_number) }
+
+            it "adds ramp_election_no_issues and returns false" do
```
shanear: This text seems to be wrong. It adds `ramp_refiling_already_processed `, not `ramp_election_no_issues`

app/models/ramp_refiling_intake.rb
```diff
@@ -87,22 +87,11 @@ def find_or_build_initial_detail
...
     @initial_ramp_refiling ||= RampRefiling.new(
-      veteran_file_number: veteran_file_number,
-      ramp_election: fetch_ramp_election
+      veteran_file_number: veteran_file_number
     )
   end
 
   def ramp_elections
     RampElection.established.where(veteran_file_number: veteran_file_number).all
   end
```
shanear: Is there anywhere where we are pulling the ramp elections associated with the selected issues? I think there might be on the stats pages.

spec/factories/ramp_election.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
```
leikkisa: Cool

spec/factories/ramp_election.rb
kahlouie: @shanear Since you can overwrite the veteran file number for any given ramp_election created by the factory, this is just a default. Do you still feel we should do something different in the factory?

shanear: I think the default should be connected to a veteran that is stubbed correctly in BGS, but it's a nice to have

app/models/ramp_refiling_intake.rb
leikkisa: I don't think the stats page has any reporting for specific issues. It has total counts for RampRefilings of different types (like HLR, SC, etc.), and this doesn't seem to need any info about the Ramp Elections.

kahlouie: It doesn't, but I did find some other areas where it's getting called. I thought I had searched more thoroughly. 😬

spec/factories/ramp_election.rb
kahlouie: I'm going to leave it as is since the veteran should be configured per test, not through the factory and not every test needs the BGS one.

Issue #5773, opened 2018-06-04
Excel Integration | Validate data


sharonwarner:
From the rails console, we should be able to:
1) Download spreadsheet from S3
1) Validate that the spreadsheets uploaded are in the correct format

PR #5880, opened 2018-06-12
Judge Spreadsheet Validation

sharonwarner:
Connects #5773 

### Description
This PR adds validation for the judge spreadsheet.

### Testing Plan
1) Download the Judge Assignments template, fill it out with fake data
1) In the rails console, run S3Service::store_file(<file_name>, <file_path>, :filepath)
1) Create a schedule period with the following command: 'JudgeSchedulePeriod.new(user: User.first, start_date: 90.days.ago, end_date: 1.days.ago, file_name: <file_name>).save!
1) Run 'SchedulePeriod.first.validate_spreadsheet', and confirm you get the expected result
1) Repeat with various errors in the spreadsheet


app/services/hearing_schedule/validate_judge_spreadsheet.rb
```diff
@@ -1,11 +1,62 @@
...
+  HEADER_COLUMNS = 7
+  NAME_COLUMN = 2
+  CSS_ID_COLUMN = 3
+  DATE_COLUMN = 4
+
+  class JudgeDatesNotCorrectFormat < StandardError; end
+  class JudgeTemplateNotFollowed < StandardError; end
+  class JudgeDatesNotUnique < StandardError; end
+  class JudgeDatesNotInRange < StandardError; end
+  class JudgeNotInDb < StandardError; end
```
tejans24: nit: I think `JudgeNotInDatabase` seems a bit more cleaner

app/services/hearing_schedule/validate_judge_spreadsheet.rb
```diff
@@ -1,11 +1,62 @@
...
+  end
+
+  def validate_judge_non_availability_dates
+    unless judge_non_availability_dates.all? { |date| date[2].instance_of?(Date) }
+      fail JudgeDatesNotCorrectFormat
+    end
+    unless judge_non_availability_dates.uniq == judge_non_availability_dates
+      fail JudgeDatesNotUnique
+    end
+    unless judge_non_availability_dates.all? { |date| date[2] > @start_date && date[2] < @end_date }
```
tejans24: Should we want to use constants for some of these magic numbers as well?

app/services/hearing_schedule/validate_judge_spreadsheet.rb
```diff
@@ -1,11 +1,62 @@
...
+    dates = judge_non_availability_template.column(DATE_COLUMN).drop(HEADER_COLUMNS)
+    names.zip(css_ids, dates)
+  end
+
+  def validate_judge_non_availability_template
+    unless judge_non_availability_template.column(1)[0] == "Judge Non-Availability Dates" &&
+           judge_non_availability_template.row(7).uniq == [nil, "Jones, Bernard", "BVAJONESB",
+                                                           Date.parse("02/04/2019")] &&
+           judge_non_availability_template.column(5).uniq == [nil]
+      fail JudgeTemplateNotFollowed
```
tejans24: Right now, these fails look like they return one at a time. Is this expected?

app/services/hearing_schedule/validate_judge_spreadsheet.rb
```diff
@@ -1,11 +1,62 @@
...
+  def judge_non_availability_dates
+    names = judge_non_availability_template.column(NAME_COLUMN).drop(HEADER_COLUMNS)
+    css_ids = judge_non_availability_template.column(CSS_ID_COLUMN).drop(HEADER_COLUMNS)
+    dates = judge_non_availability_template.column(DATE_COLUMN).drop(HEADER_COLUMNS)
+    names.zip(css_ids, dates)
+  end
+
+  def validate_judge_non_availability_template
+    unless judge_non_availability_template.column(1)[0] == "Judge Non-Availability Dates" &&
+           judge_non_availability_template.row(7).uniq == [nil, "Jones, Bernard", "BVAJONESB",
```
tejans24: Is this judge hardcoded data intended?

spec/services/hearing_schedule/validate_judge_spreadsheet_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+  before do
+    user.save!
+  end
+
+  context "when judge non-availaility dates are duplicated" do
+    subject do
+      HearingSchedule::ValidateJudgeSpreadsheet.new(
+        Roo::Spreadsheet.open("spec/support/judgeDuplicateDates.xlsx", extension: :xlsx),
+        Date.parse("01/01/2018"),
+        Date.parse("01/06/2018")
```
tejans24: do we want to add a little more interesting data for the test xlsx files? It looks like most of them just have 2 rows with the same judge name.

app/services/hearing_schedule/validate_judge_spreadsheet.rb
sharonwarner: Yes, this is one of the examples.

spec/services/hearing_schedule/validate_judge_spreadsheet_spec.rb
sharonwarner: I'm not sure if that would add much value to the unit tests. I think when we're writing tests for the algorithm or a larger feature test, more interesting data would be useful.

app/services/hearing_schedule/validate_judge_spreadsheet.rb
sharonwarner: Yeah, for now I think this is fine. I think once we write in the frontend integration it'll be easier to decide how these errors should be handled. I definitely agree that eventually we need to get all the errors at once though.

spec/services/hearing_schedule/validate_judge_spreadsheet_spec.rb
tejans24: yea, as long as we add more interesting data and test out different scenarios at some point, that works.

app/services/hearing_schedule/validate_judge_spreadsheet.rb
tejans24: Nit: I think we should make this a constant and since it's being used in multiple places. And easy to change incase we decide to change this data in the excel sheet as well.

Issue #5531, opened 2018-05-14
Refactor intake logic common to all three AMA lanes


shanear:
There is a lot of duplicate code in both the front end and back end for the three AMA lanes.

Identify different ways to refactor a lot of this. Potential candidates:

- [ ] Refactor issues selection screen into a shared component
- [ ] Change `EstablishesEndProduct` into it's own class instead of a concern so we can unit test it.

PR #5881, opened 2018-06-12
Intake AMA refactoring

leikkisa:
Connects #5531

Creates common completeIntake action function.


leikkisa:
One main question I have is whether I should bother to split out AMA actions into a separate file or just put them into "Common"?  I split it out separately for now because we might want to refactor the RAMP functions too, but we could also put them all into common and name the functions separately.

Issue #5735, opened 2018-05-31
Case details | Issues


laurjpeterson:
Viewing issues on an appeal is the highest priority for many Board users, including attorneys and judges. When attorneys and judges click to view details about a case, they want to see issues front and center, so they can get immediate context on the case. 

## AC
- Incorporate new issues designs
- If issues have a disposition already in VACOLS, show that disposition under the Note field, following the design styling
- If issues do not have a disposition, do not show the disposition row

## Mocks
![image](https://user-images.githubusercontent.com/1598889/41435760-d7dd9d92-6fed-11e8-8ca7-4e4e287291b7.png)

## Resource:
- [Case details Mural](https://app.mural.co/t/workqueue2001/m/workqueue2001/1525112874684/5765d285f568b5f631990e2c8e514a39d9b6b570)

laurjpeterson:
@lowellrex @mkhandekar looks like issues redesign is deployed, so closing. Will cover editing in future issues.

lowellrex:
@laurjpeterson I'm pretty sure this work has been completed. Can you confirm so I can go ahead and close this ticket out?

PR #5610, opened 2018-05-18
Judge check out | Evaluate decision

dannysellers:
Resolves #5151 

### Description
After a judge reviews a decision and confirms issue dispositions, they are required to evaluate the quality of the attorney's decision for their performance evaluation.

### Acceptance Criteria 
- [ ] Calendar days worked = Decision submitted - Date assigned
- [ ] Default for complexity / case quality is unselected
- [ ] Alert appears if the user chooses 1 or 2 case quality
- [ ] Error state appears when user does not select required fields (complexity, quality, additional factors occasionally)

### Testing Plan
1. Go to `/queue` as Judge (`BVAAABSHIRE`)
2. Select 'Ready for Dispatch' from Action dropdown for a case, verify issue dispositions/remand reasons, confirm Evaluate Decision page appears next
3. After filling in required fields in Evaluate Decision page, confirm Submit Decision page is next

#### Screenshots
<details><summary>No error</summary>

</details>

<details><summary>Error 1</summary>

</details>

<details><summary>Error 2</summary>

</details>

### TODO
- Hook up to backend
- Set [decision type](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/SubmitDecisionView.jsx#L107) earlier

PR #5886, opened 2018-06-12
First step of case details redesign

lowellrex:
Connects #5735.

There will be a later PR to clean up the loose ends and make the live app agree with [the mocks](https://app.mural.co/t/workqueue2001/m/workqueue2001/1525112874684/5765d285f568b5f631990e2c8e514a39d9b6b570), but I wanted to get these changes out to door in order to validate this approach for constructing the case details page.

| Before | After | 
| ------ | ----- |
| image | image |

cc @mkhandekar

### DIFF
--- a/client/app/constants/AppConstants.js
+++ b/client/app/constants/AppConstants.js
@@ -10,7 +10,8 @@ export const COLORS = {
   ...COMMON_COLORS,
   GOLD_LIGHTEST: '#FFF1D2',
   GOLD_LIGHT: '#F9C642',
-  GREY_BACKGROUND: '#f9f9f9'
+  GREY_BACKGROUND: '#f9f9f9',
+  PRIMARY: '#0071bc'
 };
 
 export const LOGO_COLORS = {
--- a/client/app/queue/AppellantDetail.jsx
+++ b/client/app/queue/AppellantDetail.jsx
@@ -4,7 +4,7 @@ import { css } from 'glamor';
 import _ from 'lodash';
 
 import BareList from '../components/BareList';
-import { boldText, TASK_ACTIONS } from './constants';
+import { boldText } from './constants';
 import COPY from '../../COPY.json';
 import { DateString } from '../util/DateUtil';
 
@@ -44,7 +44,7 @@ export default class AppellantDetail extends React.PureComponent {
 
   veteranIsAppellant = () => _.isNull(this.getAppealAttr('appellant_full_name'));
 
-  getDetails = ({ nameField, genderField, dobField, addressField, relationField }) => {
+  getDetails = ({ nameField, genderField, dobField, addressField, relationField, regionalOfficeField }) => {
     const details = [{
       label: 'Name',
       value: this.getAppealAttr(nameField)
@@ -74,6 +74,14 @@ export default class AppellantDetail extends React.PureComponent {
         value: this.formatAddress(addressField)
       });
     }
+    if (regionalOfficeField && this.getAppealAttr(regionalOfficeField)) {
+      const { city, key } = this.getAppealAttr(regionalOfficeField);
+
+      details.push({
+        label: 'Regional Office',
+        value: `${city} (${key.replace('RO', '')})`
+      });
+    }
 
     const getDetailField = ({ label, value }) => () => <React.Fragment>
       <span {...boldText}>{label}:</span> {value}
@@ -81,10 +89,6 @@ export default class AppellantDetail extends React.PureComponent {
 
     return <BareList ListElementComponent="ul" items={details.map(getDetailField)} />;
   };
-
-  componentDidMount = () => {
-    window.analyticsEvent(this.props.analyticsSource, TASK_ACTIONS.VIEW_APPELLANT_INFO);
-  }
 
   render = () => {
     let appellantDetails;
@@ -99,7 +103,8 @@ export default class AppellantDetail extends React.PureComponent {
             nameField: 'veteran_full_name',
             genderField: 'veteran_gender',
             dobField: 'veteran_date_of_birth',
-            addressField: 'appellant_address'
+            addressField: 'appellant_address',
+            regionalOfficeField: 'regional_office'
           })}
         </ul>
       </React.Fragment>;
@@ -122,7 +127,8 @@ export default class AppellantDetail extends React.PureComponent {
           {this.getDetails({
             nameField: 'veteran_full_name',
             genderField: 'veteran_gender',
-            dobField: 'veteran_date_of_birth'
+            dobField: 'veteran_date_of_birth',
+            regionalOfficeField: 'regional_office'
           })}
         </ul>
       </React.Fragment>;
@@ -136,6 +142,5 @@ export default class AppellantDetail extends React.PureComponent {
 }
 
 AppellantDetail.propTypes = {
-  analyticsSource: PropTypes.string.isRequired,
   appeal: PropTypes.object.isRequired
 };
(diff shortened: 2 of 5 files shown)


client/app/constants/AppConstants.js
```diff
@@ -10,7 +10,8 @@ export const COLORS = {
   ...COMMON_COLORS,
   GOLD_LIGHTEST: '#FFF1D2',
   GOLD_LIGHT: '#F9C642',
-  GREY_BACKGROUND: '#f9f9f9'
+  GREY_BACKGROUND: '#f9f9f9',
+  PRIMARY: '#0071bc'
```
lowellrex: Primary color taken from [styleguide](http://dsva-appeals-certification-demo-1715715888.us-gov-west-1.elb.amazonaws.com/styleguide#colors).

client/app/queue/AppealDetail.jsx
```diff
@@ -5,7 +5,7 @@ import _ from 'lodash';
 
 import IssueList from './components/IssueList';
 import BareList from '../components/BareList';
-import { boldText, TASK_ACTIONS } from './constants';
+import { boldText } from './constants';
```
lowellrex: Since we are now loading all of the sections of the case details page when the case details page loads (as opposed to loading them when their tab was selected), I've moved the analytics up a level to the case details component.

client/app/queue/AppellantDetail.jsx
```diff
@@ -4,7 +4,7 @@ import { css } from 'glamor';
 import _ from 'lodash';
 
 import BareList from '../components/BareList';
-import { boldText, TASK_ACTIONS } from './constants';
+import { boldText } from './constants';
```
lowellrex: Since we are now loading all of the sections of the case details page when the case details page loads (as opposed to loading them when their tab was selected), I've moved the analytics up a level to the case details component.

client/app/queue/AppellantDetail.jsx
```diff
@@ -74,6 +74,14 @@ export default class AppellantDetail extends React.PureComponent {
...
       });
     }
+    if (regionalOfficeField && this.getAppealAttr(regionalOfficeField)) {
+      const { city, key } = this.getAppealAttr(regionalOfficeField);
+
+      details.push({
+        label: 'Regional Office',
+        value: `${city} (${key.replace('RO', '')})`
+      });
+    }
```
lowellrex: Regional office information was part of the appeal detail information that has been removed so we moved it here for now until--or if--we find a better spot for it to live. Figure it is appropriate to include alongside the Veteran information since the regional office from which this appeal originated is likely to be close to the Veteran's residence.

client/app/queue/CaseHearingsDetail.jsx
```diff
@@ -22,7 +21,7 @@ const noTopBottomMargin = css({
   marginBottom: '1rem'
 });
 
-export default class AppealDetail extends React.PureComponent {
+export default class CaseHearingsDetail extends React.PureComponent {
```
lowellrex: This element only contains information about hearings related to this case. Power of attorney information has been moved to its own section and regional office information is included in the Veteran information.

client/app/queue/QueueDetailView.jsx
```diff
@@ -48,9 +52,14 @@ class QueueDetailView extends React.PureComponent {
...
-      name="queue-tabwindow"
-      tabs={this.tabs()} />
+    <hr {...horizontalRuleStyling} />
+    <StickyNavContentArea>
+      <IssueList title="Issues" appeal={_.pick(this.props.appeal.attributes, 'issues')} />
+      <PowerOfAttorneyDetail title="Power of Attorney" appeal={this.props.appeal} />
+      { this.props.appeal.attributes.hearings.length &&
+        <CaseHearingsDetail title="Hearings" appeal={this.props.appeal} /> }
+      <AppellantDetail title="About the Veteran" appeal={this.props.appeal} />
+    </StickyNavContentArea>
```
lowellrex: This is the heart of the change right here. We've replaced the `TabWindow` with what I am calling a `StickyNavContentArea` (open to changing the name) that accepts child elements whose only requirement is that they must have a `title` property. Curious to hear your thoughts on this approach and how it might be improved.

client/app/queue/StickyNavContentArea.jsx
```diff
@@ -0,0 +1,83 @@
+import { css } from 'glamor';
+import React from 'react';
+
+import StringUtil from '../util/StringUtil';
+import { COLORS } from '../constants/AppConstants';
+
+const sectionNavigationContainerStyling = css({
```
lowellrex: I can't exactly pinpoint why, but this mass of styling declarations strikes me as a code smell. Maybe we should move the common `3rem` padding/margin to a variable, but I don't think this would entirely get rid of the stench. Curious to get your thoughts on this styling stuff.

client/app/queue/QueueDetailView.jsx
dannysellers: I like it! I wonder if this should be moved out into `app/queue/components` or `app/components`. Also, this might be a good place to use a Flow interface when the time comes, to require `title` for the components that get rendered

client/app/queue/StickyNavContentArea.jsx
dannysellers: I totally agree, and don't really have any good solutions. As part of #5610 (not merged yet), I created some functions inside `queue/constants` that've been useful:

```javascript
export const marginTop = (margin) => css({ marginTop: `${margin}rem` });
export const marginBottom = (margin) => css({ marginBottom: `${margin}rem` });
export const marginLeft = (margin) => css({ marginLeft: `${margin}rem` });
export const marginRight = (margin) => css({ marginRight: `${margin}rem` });

...
const hrStyling = css(marginTop(2), marginBottom(3));
```

It's funny to think that one not-so-extreme use for glamor would be to basically come full circle...
```javascript
const sectionNavigationContainerStyling = css(
  paddingRight(3),
  sticky(),
  top(3),
  width('20%'),
  { float: 'left' }
);
```
🤦‍♂️

client/app/queue/StickyNavContentArea.jsx
```diff
@@ -0,0 +1,83 @@
...
+  float: 'left',
+  width: '80%'
+});
+
+const getIdForElement = (elem) => `${StringUtil.parameterize(elem.props.title)}-section`;
+
+export default class StickyNavContentArea extends React.PureComponent {
+  render = () => {
+    // Ignore undefined child elements.
+    const childElements = this.props.children.filter((child) => typeof child === 'object');
```
dannysellers: If you fancied some 🏌️⛳️, you could pass [`_.isObjectLike`](https://lodash.com/docs/4.17.10#isObjectLike) (jk)

client/app/queue/StickyNavContentArea.jsx
```diff
@@ -0,0 +1,83 @@
...
+});
+
+const sectionBodyStyling = css({
+  float: 'left',
+  width: '80%'
+});
+
+const getIdForElement = (elem) => `${StringUtil.parameterize(elem.props.title)}-section`;
+
+export default class StickyNavContentArea extends React.PureComponent {
```
dannysellers: I think this could just be a functional component—it takes no (named) arguments, and has no state or lifecycle methods.

client/app/queue/CaseHearingsDetail.jsx
dannysellers: Should `QueueDetailView` be renamed `AppealDetail`/`AppealDetailView`?

lowellrex: Went with `CaseDetailsView` because that's what we have started to call this page when talking about it and also because we are just now getting rid of the `AppealDetail` component so I want to avoid some confusion there.

PR #5892, opened 2018-06-12
Refactor AMA Intake - submitReview

leikkisa:
Connects #5531

Refactor submitReview action for AMA


client/app/intake/util/index.js
```diff
@@ -78,3 +79,30 @@ export const formatRatingData = (intakeState) => {
...
+      claimant: intakeData.claimant
+    };
+  case 'higherLevelReview':
+    return {
+      informal_conference: intakeData.informalConference,
+      same_office: intakeData.sameOffice,
+      receipt_date: formatDateStringForApi(intakeData.receiptDate),
+      claimant: intakeData.claimant
+    };
+  default:
```
leikkisa: Linting made me put a default, but I wasn't really sure what to put here. Maybe I should throw an error?

client/app/intake/actions/higherLevelReview.js
```diff
@@ -1,10 +1,6 @@
-import { ACTIONS, ENDPOINT_NAMES } from '../constants';
-import ApiUtil from '../../util/ApiUtil';
-import { formatDateStringForApi } from '../../util/DateUtil';
-import { formatRatingData } from '../util';
-import _ from 'lodash';
+import { ACTIONS } from '../constants';
 
-const analytics = true;
+// const analytics = true;
```
kahlouie: Did you mean to leave this in there?

client/app/intake/util/index.js
kahlouie: I think that's fine as is, but I might also add claimant since that should also me in all the intakes.

client/app/intake/actions/higherLevelReview.js
leikkisa: Nope! Good call.

leikkisa: Done

client/app/intake/util/index.js
leikkisa: Done

PR #5893, opened 2018-06-12
Fix judge case review test

aroltsch:
Do not hard code issue sequence ID as it is incremented with every call to create a factory


evankroske:
FYI, I just removed the Ready-to-Merge tag because the merge conflict was causing the auto-merge job to fail. Feel free to add it back once the conflict has been resolved.

PR #5898, opened 2018-06-12
VACOLS Integration - Query for all master records by a date range

oscarAtNava:
Connects #5451

### Description
Implemented ability to query for all hearing master records for a particular date range. End point returns all hearing types.
Included in PR is the refactoring of hearing day creation to a new model. It was previously added to the Hearings model but this model is not going to be used with Hearing Schedule. New model name is HearingDay.

### Acceptance Criteria 
- [ ] Code compiles correctly
- [ ] Test cases pass. Included in PR are both request and model tests.

### Testing Plan
1. Go to rails console and enter following command:
`HearingDay.load_days_for_range(Date.parse('01-Jan-2017'), Date.parse('31-Dec-2017'))`

You should see two result sets: one for case hearings and one for travel hearings.


app/controllers/hearings/hearing_day_controller.rb
```diff
@@ -0,0 +1,54 @@
+class Hearings::HearingDayController < HearingsController
```
sharonwarner: Do we want the parent class to be HearingsController? The way I'm viewing it is that a Hearing would belong to a HearingDay. I don't think this controller has to inherit anything from the HearingsController.

app/controllers/hearings/hearing_day_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    verify_authorized_roles("Hearing Schedule")
+  end
+
+  def invalid_record_error(hearing)
+    render json:  {
+      "errors": ["title": "Record is invalid", "detail": hearing.errors.full_messages.join(" ,")]
+    }, status: 400
+  end
+
+  def json_hearings(hearings)
```
sharonwarner: Do you think this could go in the model?

app/controllers/hearings/hearing_day_controller.rb
oscarAtNava: Agreed. I moved some utility functions to Hearing Day controller.

app/controllers/hearings/hearing_day_controller.rb
oscarAtNava: I modeled this to the work done in tasks_controller for queue. Will keep as is for now as per our conversation.

PR #5899, opened 2018-06-12
Adding factory bot to demo env

mdbenjam:
### Description
Demo deploys are broken because we weren't installing factory bot in demo.

### Testing Plan
1. Deployed this branch to demo, and it succeeded.


Gemfile
```diff
@@ -100,7 +100,7 @@ group :production, :staging do
   gem "rails_stdout_logging"
 end
 
-group :stubbed, :test, :development do
+group :stubbed, :test, :development, :demo do
```
sharonwarner: Do we definitely want to install all these gems in demo?

mdbenjam: Reasonable question. I can go either way. I generally think anything we have for development should be in demo. But if you think we should just include this one, then I'm down to move it to a different section instead

Issue #5914, opened 2018-06-14
Use VACOLS user id instead of regional office when updating location


amprokop:
An attorney reported:

> seems like when an attorney is assigning a case to a judge in caseflow, we are first assigning it to a VACO user and then the VACO user assigns it to the judge. the IDT shows judges a screen with cases assigned to them to review and sign, and who assigned it. but now they cannot see what attorney assigned it. judges are seeing a blank, because this VACO user is an invalid user

This appears to be a bug that was around since Feb 2017. After looking at the original PR, @cmgiven and I both believe that we're mistakenly using `regional_office` as `user_db_id`, instead of the VACOLS username: https://github.com/department-of-veterans-affairs/caseflow/pull/792/files#diff-835bda6b4338488dd5b3a03d10f6b8b9R167

See 
https://github.com/department-of-veterans-affairs/dsva-vacols/issues/14 for the question we asked jed.

# AC
Use the user id in the prior locations table, not the regional office.

laurjpeterson:
Looks like a PR was merged here, so closing.

PR #5915, opened 2018-06-14
Use username when updating location, not RO

amprokop:
Connects https://github.com/department-of-veterans-affairs/caseflow/issues/5914

### Testing Plan
1. Ensure that Dispatch code still works.
2. Test that attorney checkout to judge still works.


sharonwarner:
I confirmed that dispatch users have a vacols_uniq_id. Looks good to me.

Issue #5303, opened 2018-04-25
Judges | Case details | Assign to attorneys


laurjpeterson:
Allow judges to assign to attorneys from case details screen of a case that is assigned to them.

evankroske:
## Implementation Plan
### What data does AssignWidget need?
- [ ] attorneysOfJudge
- [ ] attorneys
- [ ] selectedTasks

PR #5916, opened 2018-06-14
Judge Queue: Harmonizes /queue and /tasks endpoints

evankroske:
Part of https://github.com/department-of-veterans-affairs/caseflow/issues/5303

### Description
Adds appeal IDs to the response of /queue and changes the format of the response of /tasks to match that of /queue.

### Testing Plan
- [x] `bundle exec rspec spec/controllers/tasks_controller_spec.rb`


app/controllers/tasks_controller.rb
```diff
@@ -57,6 +57,18 @@ def create
     task = task_class.create(task_params)
 
     return invalid_record_error(task) unless task.valid?
+    if task_class == JudgeCaseAssignmentToAttorney then
+      case_assignments = VACOLS::CaseAssignment.select_tasks.where("brieff.bfkey = ?", task.vacols_id)
```
aroltsch: Do you think we could abstract this logic out of the controller? I see very similar code in both methods.

app/controllers/tasks_controller.rb
```diff
@@ -57,6 +57,18 @@ def create
...
+      render json: {
+        task:
+          ActiveModelSerializers::SerializableResource.new(
+            JudgeLegacyTask.from_vacols(
```
aroltsch: I think it should be AttorneyLegacyTask because this is what we are creating.

app/controllers/tasks_controller.rb
aroltsch: Ugh, I think I understand why you chose to put it here

aroltsch: Let's chat about it when you have a moment

aroltsch: Would it be cleaner to just reload the tasks? Re-direct to `index`

app/controllers/tasks_controller.rb
evankroske: Thanks! I just fixed this

app/controllers/tasks_controller.rb
evankroske: Done! I extracted the query and the serialization, as we discussed

PR #5925, opened 2018-06-14
VACOLS Integration, added update methods for all hearing schedule types.

oscarAtNava:
Connects #5453 #5452

### Description
Implemented update operations for CaseHearing and TravelBoardSchedule for updates specific to Hearing Schedule. Specifically we assign a judge to an existing master record, or we update the RO in a specific travel board master record. 

### Acceptance Criteria 
- [ ] Code compiles correctly
- [ ] Request test cases included in PR pass.

### Testing Plan
1. Set up current user:
`RequestStore[:current_user] = User.find_by(css_id: "BVASCASPER1")`
2. Go to rails console and find an existing hearing day:
`hearing = HearingDay.find_hearing_day("T", "2018-85-1")`
3. Update the RO value from RO44 to RO17:
`HearingDay.update_hearing_day(hearing, {hearing_type: "T", tbro: "RO17"})`
4. Verify value was changed:
`hearing = HearingDay.find_hearing_day("T", "2018-85-1")`

### DIFF
--- a/app/mappers/hearing_mapper.rb
+++ b/app/mappers/hearing_mapper.rb
@@ -21,7 +21,9 @@ module HearingMapper
         room: hearing_info[:room],
         hearing_date: hearing_info[:hearing_date],
         hearing_type: hearing_info[:hearing_type],
-        representative: hearing_info[:representative]
+        representative: hearing_info[:representative],
+        board_member: hearing_info[:board_member],
+        tbro: hearing_info[:tbro]
       }.select { |k, _v| hearing_info.keys.map(&:to_sym).include? k } # only send updates to key/values that are passed
     end
 
--- a/app/models/vacols/travel_board_schedule.rb
+++ b/app/models/vacols/travel_board_schedule.rb
@@ -2,6 +2,15 @@
 class VACOLS::TravelBoardSchedule < VACOLS::Record
   self.table_name = "vacols.tbsched"
 
+  attribute :tbleg, :integer
+
+  COLUMN_NAMES = {
+      tbyear: :tbyear,
+      tbtrip: :tbtrip,
+      tbleg: :tbleg,
+      tbro: :tbro
+  }.freeze
+
   # :nocov:
   class << self
     def hearings_for_judge(css_id)
@@ -30,6 +39,17 @@ class VACOLS::TravelBoardSchedule < VACOLS::Record
     end
   end
 
+  def update_hearing!(hearing_info)
+    attrs = hearing_info.each_with_object({}) { |(k, v), result| result[COLUMN_NAMES[k]] = v }
+    attrs.delete(nil)
+    MetricsService.record("VACOLS: update_hearing! #{tbyear}-#{tbtrip}-#{tbleg}",
+                          service: :vacols,
+                          name: "update_hearing") do
+      VACOLS::TravelBoardSchedule.where(tbyear: tbyear, tbtrip: tbtrip, tbleg: tbleg)
+          .update_all(attrs.merge(tbmoduser: current_user_slogid, tbmodtime: VacolsHelper.local_time_with_utc_timezone))
+    end
+  end
+
   def master_record_type
     :travel_board
   end
@@ -37,4 +57,12 @@ class VACOLS::TravelBoardSchedule < VACOLS::Record
   def vdkey
     nil
   end
+
+  private
+
+  def current_user_slogid
+    slogid = RequestStore.store[:current_user].vacols_uniq_id
+    slogid.nil? ? "" : slogid.upcase
+  end
+
 end
--- a/app/repositories/hearing_day_repository.rb
+++ b/app/repositories/hearing_day_repository.rb
@@ -7,6 +7,20 @@ class HearingDayRepository
       VACOLS::CaseHearing.create_hearing!(hearing_hash) if hearing_hash.present?
     end
 
+    def update_vacols_hearing!(hearing, hearing_hash)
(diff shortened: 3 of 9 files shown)


app/mappers/hearing_mapper.rb
```diff
@@ -21,7 +21,9 @@ def hearing_fields_to_vacols_codes(hearing_info)
         room: hearing_info[:room],
         hearing_date: hearing_info[:hearing_date],
         hearing_type: hearing_info[:hearing_type],
-        representative: hearing_info[:representative]
+        representative: hearing_info[:representative],
```
sharonwarner: These names seem kind of ambiguous to me. Does board_member correspond to the judge? And who is the representative?

app/models/vacols/travel_board_schedule.rb
```diff
@@ -30,11 +39,29 @@ def load_days_for_range(start_date, end_date)
...
     :travel_board
   end
 
   def vdkey
     nil
   end
+
+  private
+
+  def current_user_slogid
```
sharonwarner: I think we already have this method somewhere else, does it make sense to move it to a util file or a shared file?

spec/requests/hearing_day_spec.rb
```diff
@@ -13,6 +13,37 @@
...
+      Generators::Vacols::Staff.create
+      Generators::Vacols::TravelBoardSchedule.create({})
+    end
+
+    it "Update RO in master TB schedule" do
+      hearing
+      put "/hearings/#{hearing.tbyear}-#{hearing.tbtrip}-#{hearing.tbleg}/hearing_day",
+          params: { hearing_type: "T", tbro: "RO27" }
+      expect(response).to have_http_status(:success)
+      # commented out as there is an issue getting back the updated hearing. Works in rails console.
```
sharonwarner: Hm. Sometimes these errors are related to different gems being installed in the test environment versus development. Are we relying on a gem that doesn't exist in the test environment?

tejans24: That's weird that this wouldn't work, can we double check this response object to see if it has a body?

app/models/vacols/travel_board_schedule.rb
```diff
@@ -2,6 +2,15 @@
...
   self.table_name = "vacols.tbsched"
 
+  attribute :tbleg, :integer
+
+  COLUMN_NAMES = {
+    tbyear: :tbyear,
+    tbtrip: :tbtrip,
+    tbleg: :tbleg,
+    tbro: :tbro
+  }.freeze
```
tejans24: should they be able to update who is on the travel board? `tbmem1`?
And also start and end dates?

app/mappers/hearing_mapper.rb
oscarAtNava: Board member is the judge, and representative is an attribute in the current Hearing model. I did not come up with either names. They are what they are in the DB.

spec/requests/hearing_day_spec.rb
oscarAtNava: Response has a body, and the hearing object is there, just not in an updated mode. It is the prior version. I sense that using the where()...update_all is not updating the hearing object in memory.

app/repositories/hearing_day_repository.rb
```diff
@@ -7,6 +7,20 @@ def create_vacols_hearing!(hearing_hash)
       VACOLS::CaseHearing.create_hearing!(hearing_hash) if hearing_hash.present?
     end
 
+    def update_vacols_hearing!(hearing, hearing_hash)
+      hearing_hash = HearingMapper.hearing_fields_to_vacols_codes(hearing_hash)
+      hearing.update_hearing!(hearing_hash) if hearing_hash.present?
+    end
+
+    def find_hearing_day(hearing_type, hearing_key)
+      if hearing_type.nil?
```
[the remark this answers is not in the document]
oscarAtNava: Sure, will make that change.

app/models/vacols/travel_board_schedule.rb
oscarAtNava: Not for the August MVP.

app/repositories/hearing_day_repository.rb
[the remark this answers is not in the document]
oscarAtNava: Created a constant in the new HearingDay model to not depend on VACOLS.

app/mappers/hearing_mapper.rb
sharonwarner: Definitely, but in the mapper, we could map them to what we'll eventually call them in our database.

spec/requests/hearing_day_spec.rb
```diff
@@ -7,18 +7,52 @@
...
+      Generators::Vacols::TravelBoardSchedule.create({})
+    end
+
+    it "Update RO in master TB schedule" do
+      hearing
+      put "/hearings/#{hearing.tbyear}-#{hearing.tbtrip}-#{hearing.tbleg}/hearing_day",
+          params: { hearing_type: HearingDay::HEARING_TYPES[:travel], tbro: "RO27" }
+      expect(response).to have_http_status(:success)
+      # commented out as there is an issue getting back the updated hearing. Works in rails console.
+      # expect(JSON.parse(response.body)["hearing"]["data"]["attributes"]["tbro"]).to eq("RO27")
```
tejans24: can we just remove this line? There is no point in keeping dead commented code imo.

oscarAtNava: I actually took the time to fix it so that the update end point always returns the updated object.

Issue #5851, opened 2018-06-08
Write unit tests for method RemandReasonRepository.load_remand_reasons_for_appeals


aroltsch:
Ensure all necessary factories are created
All scenarios are tested for these methods

PR #5935, opened 2018-06-18
Write Facols unit tests for remand repository

aroltsch:
Resovles #5851 

### Description
FACOLS unit tests for the RemandReasonRepository.load_remand_reasons_for_appeals

### DIFF
--- a/app/repositories/remand_reason_repository.rb
+++ b/app/repositories/remand_reason_repository.rb
@@ -62,19 +62,6 @@ class RemandReasonRepository
     end
   end
 
-  def self.remand_reason_from_vacols_remand_reason(reason)
-    {
-      code: reason.rmdval,
-      after_certification: reason.rmddev.eql?("R2")
-    }
-  end
-
-  def self.load_remands_from_vacols(vacols_id, vacols_sequence_id)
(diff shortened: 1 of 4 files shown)


app/repositories/remand_reason_repository.rb
```diff
@@ -1,28 +1,4 @@
 class RemandReasonRepository
-  # :nocov:
-  def self.create_remand_reasons!(vacols_id, vacols_sequence_id, remand_reasons)
```
aroltsch: Moved these methods down.

spec/repositories/remand_reason_repository_spec.rb
```diff
@@ -1,4 +1,55 @@
...
+    let(:issues3) do
+      [create(:case_issue, :disposition_remanded),
+       create(:case_issue, :disposition_allowed),
+       create(:case_issue, :disposition_remanded)]
+    end
+
+    let!(:remand_reasons) do
+      [
+        create(:remand_reason,
+               rmdkey: issues1.second.isskey,
```
mdbenjam: This should work, but you created an after action in the case_issue factory to set `remand_reasons`. Shouldn't we just use that instead?

aroltsch: Haha, I knew it you would say something.

PR #5939, opened 2018-06-18
Improve display of case snapshot at slimmer widths

lowellrex:
This change improves the display of the case snapshot element at widths narrower than 1201 pixels, since more than 30% of folks who use caseflow have a screen width of 900 or fewer pixels. [Initial Slack discussion here](https://dsva.slack.com/archives/C6E41RE92/p1529009595000136).

| Before | After |
| ------ | ----- |
| image | image |

### DIFF
--- a/client/app/queue/CaseSnapshot.jsx
+++ b/client/app/queue/CaseSnapshot.jsx
@@ -49,7 +49,7 @@ const snapshotChildResponsiveWrapFixStyling = css({
       margin: '2rem 3rem 0 3rem',
       marginRight: '3rem !important',
       paddingTop: '2rem',
-      width: '100% !important'
+      width: '100%'
     },
     '& > div:nth-child(2)': { borderRight: 'none' }
   }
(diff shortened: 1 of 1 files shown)


client/app/queue/CaseSnapshot.jsx
```diff
@@ -41,6 +42,19 @@ const headingStyling = css({
   marginBottom: '0.5rem'
 });
 
+const snapshotChildResponsiveWrapFixStyling = css({
+  '@media(max-width: 1200px)': {
+    '& > .usa-width-one-half': {
+      borderTop: `1px solid ${COLORS.GREY_LIGHT}`,
+      margin: '2rem 3rem 0 3rem',
+      marginRight: '3rem !important',
```
dannysellers: Are these `!important`s necessary? I don't see any difference after removing them

lowellrex: Good call! Looks like the width important is not needed, but the right margin important is necessary to override the following rule:
image

evankroske:
Hey, I just removed the ready-to-merge tag because this PR is blocked by a merge conflict, and it's causing the merge job to fail. Feel free to reapply the tag once you've fixed the merge conflict

Issue #5766, opened 2018-06-01
Make case search work with AMA appeals


mdbenjam:
Case search currently only searches for legacy appeals. We need to modify it to also look for AMA appeals. The results page should show both types side by side.

mkhandekar:
👀 so I can stay updated!

PR #5944, opened 2018-06-18
Return AMA appeals in queue case search

lowellrex:
Connects #5766.

Return AMA appeals (if they exist) when searching for a Veteran's file in queue case search.

## Test plan
* Fire up the rails console `rails c` in the caseflow directory
* Create one or more AMA appeal records in the database using FactoryBot `FactoryBot.create(:appeal, veteran_file_number: file_num)` (You can use file numbers for cases that exist in our VACOLS test data set in order to see both types of appeal showing up in the same results list!)
* See the new records you created show up in the search results!


app/controllers/appeals_controller.rb
```diff
@@ -60,11 +60,13 @@ def get_appeals_for_file_number(file_number)
...
 
+      # rubocop:disable Lint/HandleExceptions
+      appeals = Appeal.where(veteran_file_number: file_number).to_a
       begin
-        appeals = LegacyAppeal.fetch_appeals_by_file_number(file_number)
+        appeals.concat(LegacyAppeal.fetch_appeals_by_file_number(file_number))
       rescue ActiveRecord::RecordNotFound
-        appeals = []
       end
+      # rubocop:enable Lint/HandleExceptions
```
lowellrex: Currently the results will be displayed in the following order:
* all new AMA appeals first (ordered by primary key)
* all old VACOLS appeals second (ordered by primary key)

I think this is fine behaviour for now unless we already know how we'd like to order this potentially mixed set of results.

PR #5948, opened 2018-06-18
CircleCI: Configures Rspec to fail fast

evankroske:
### Description
Configures Rspec to fail as soon as the first test fails. This should speed up failed builds. On the other hand, it will make it much harder to use CircleCI as a development tool.


evankroske:
The team decided that this wasn't a good idea

Issue #5691, opened 2018-05-29
Move Status tests to FACOLS


mdbenjam:
Status tests in the following files are failing. We need to move them to FACOLS.

```
spec/jobs/take_docket_snapshot_job_spec.rb,
spec/models/appeal_events_spec.rb,
spec/models/appeal_history_spec.rb,
spec/models/appeal_series_alerts_spec.rb,
spec/models/appeal_series_issues_spec.rb,
spec/models/appeal_series_spec.rb,
spec/models/docket_snapshot_spec.rb,
spec/models/docket_tracer_spec.rb,
spec/feature/api/v1/appeals_spec.rb,
spec/feature/api/v2/appeals_spec.rb,
```

PR #5955, opened 2018-06-19
Moving final appeals status test to facols

mdbenjam:
Connects #5691

### Description
Moves remaining status tests to FACOLS.

### Acceptance Criteria 
- [ ] Tests pass


app/services/appeal_repository.rb
```diff
@@ -54,11 +53,9 @@ def self.appeals_by_vbms_id_with_preloaded_status_api_attrs(vbms_id)
 
       cases.map do |case_record|
         appeal = build_appeal(case_record)
-        appeal.aod = case_record["aod"] == 1
```
mdbenjam: I don't think this is actually doing what it's supposed to. Even when the data is setup right, it `case_record["aod]` and `case_record["rem_return"]` don't seem to have those values. Also, we now get these values on every request instead of using the cached attribute on the legacy_appeal model. I think it's better to go through the standard path.

spec/feature/api/v2/appeals_spec.rb
```diff
@@ -139,7 +140,7 @@
       expect(ApiView.count).to eq(0)
     end
 
-    it "returns 404 if veteran with that SSN isn't found" do
+    it "returns 404 if veteran with that SSN isn't found", skip: "I believe this just returns an empty array" do
```
mdbenjam: When an SSN isn't found it seems to return `[]` not a 404.

app/services/appeal_repository.rb
cmgiven: This change will have disastrous performance repercussions. Currently, an appeals status hit triggers 4 SQL queries: brieff, issue, hearingsched, and cova. Not great, but tolerable (typical response times are about 2000-2500ms). With this change, it will now trigger 4+2n queries, where n is the number of appeals owned by the Veteran. This gets really bad really fast given VACOLS latency. Assume about a 400ms hit per query. When you're looking at a one or two appeal Veteran, it's bad enough. But when we get to the 10+ appeal range, we're looking at 20 more queries and 8s added to the response time.

spec/feature/api/v2/appeals_spec.rb
cmgiven: Note that this is about BGS, not VACOLS. If the SSN is present in BGS, but not in VACOLS, it would return a 200 with an empty array.

mdbenjam: Are you sure this is about BGS and not VACOLS? It's thrown when we get an `ActiveRecord::RecordNotFound` error: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/api/v2/appeals_controller.rb#L6

That error is thrown in the fake repository: https://github.com/department-of-veterans-affairs/caseflow/blob/master/lib/fakes/appeal_repository.rb#L148

mdbenjam:
That's a good note. I've updated it to increase the complexity by a constant (2) rather than linearly. I cannot figure out how to get it working in it's current form with the joins. I'm not sure how you can use joins like that in active record without a corresponding select statement. I'm pretty confident that this does not work properly right now. If this still doesn't meet the needs for the API, then would you be able to take a look to see if you can fix the join?

cmgiven:
Tested just now, and in their current form, both joins are working in prod.

Can I come back to this one next week?

Issue #5884, opened 2018-06-12
[BE] Determine if user is a co-located admin and return the vacols role


aroltsch:
When a user makes a request to get a list of tasks, we need to determine the user's vacols role.
Reference: https://github.com/department-of-veterans-affairs/dsva-vacols/issues/2

PR #5957, opened 2018-06-19
Anya/5884 determine if user co located admin

aroltsch:
Connects #5884 

### Description
This PR does two things:
1. Adds co-located vacols role
2. Returns multiple roles for a user


spec/factories/vacols/staff.rb
```diff
@@ -5,14 +5,18 @@
...
     trait :attorney_role do
-      svlj "A"
       sattyid "123"
     end
 
     trait :judge_role do
       svlj "J"
     end
 
+    trait :attorney_judge_role
```
dannysellers: should be a `do` here

Issue #5903, opened 2018-06-12
Establish claim fails for higher level review and supplemental claim flows when adding a non-rating issue


shanear:
When a non-rating issue is added to a HLR or SC, the resulting call to create contentions in VBMS throws an error.

## Acceptance Criteria
- HLR and SC with non-rating issues should correctly create the EP and associated contentions in VBMS

## Technical Notes
This is the XML blob with the VBMS error in it
```
VBMS::HTTPError: status_code=500, body=<env:Envelope xmlns:env="http://schemas.xmlsoap.org/soap/envelope/"><env:Header/><env:Body><env:Fault><faultcode>env:Server</faultcode><faultstring>Error validating ClaimService data input. GUID: f9bc41eb-fafc-482b-a8e0-3f4e6a56e033</faultstring><detail><ns4:dataValidationFaultDetailBean xmlns:ns4="http://vbms.vba.va.gov/external/ClaimService/v4" xmlns:ns3="http://vbms.vba.va.gov/cdm/participant/v4" xmlns:ns1="http://vbms.vba.va.gov/cdm/common/v4" xmlns:ns0="http://vbms.vba.va.gov/cdm/claim/v4"><ns4:formFieldErrors><ns4:fieldName>Title</ns4:fieldName><ns4:errorType>MINIMUM_LENGTH_NOT_SATISFIED</ns4:errorType><ns4:message>The minimum data length for the Title within the contentionsToCreate was not satisfied: The Title must not be empty.</ns4:message></ns4:formFieldErrors></ns4:dataValidationFaultDetailBean></detail></env:Fault></env:Body></env:Envelope>, request=#<VBMS::Requests::CreateContentions:0x007fa818257d90 @veteran_file_number="984562357", @claim_id="600131031", @contentions=[nil, "Service connection for pain disorder is granted with an evaluation of 70 percent effective May 1, 2013.", "Entitlement to individual unemployability is granted effective May 1, 2013."]>
```

leahbannon:
* I was using 984562357 
* enter date
* selected 2 check boxes for issues
* clicked Add Issue > chose Unknown from dropdown > wrote no description
* click Establish EP

leahbannon:
This is related to anything with adding additional non-rated issues. It breaks no matter what category or description I add.

leikkisa:
@leahbannon Does it work when you do add a description?

leikkisa:
@abbyraskinUSDS @leahbannon 

[...]

PR #5965, opened 2018-06-19
Do not allow blank descriptions for non-rated issues

leikkisa:
Connects #5903


leikkisa:
It will just get left out of the data that gets sent to the back end, if either or both of them are left out. I have a conversation started in the actual ticket to figure out more with Abby and Leah. We had a meeting about this before but I think the actual design decisions still need to be determined.

client/app/intake/components/NonRatedIssues.jsx
```diff
@@ -14,6 +14,16 @@ export default class NonRatedIssuesUnconnected extends React.PureComponent {
       setIssueDescription
     } = this.props;
 
+    let disableAddNonRatedIssue;
+
+    if (Object.keys(nonRatedIssues).length === 0) {
+      disableAddNonRatedIssue = false;
+    } else {
+      disableAddNonRatedIssue = Boolean(_.reduce(nonRatedIssues, (result, issue) => {
```
shanear: It's pretty tough to tell at a glance what this statement is supposed to be doing. It looks like result is either 0 if all descriptions are filled in, or -X where X is the number of nil descriptions? Why do we need to calculate the number of nil descriptions?

I think [every](https://lodash.com/docs/4.17.5#every) is what you might want here.

I would think all you'd need is:

```
disableAddNonRatedIssue = _.every(nonRatedIssues, 'description')
```

PR #5966, opened 2018-06-19
Match intake with existing ramp election on complete

kahlouie:
Connects #5776

### DIFF
--- a/app/models/ramp_election_intake.rb
+++ b/app/models/ramp_election_intake.rb
@@ -18,7 +18,7 @@ class RampElectionIntake < Intake
   end
 
   def find_or_build_initial_detail
-    matching_ramp_election
+    new_intake_ramp_election
   end
 
   def review!(request_params)
@@ -33,6 +33,12 @@ class RampElectionIntake < Intake
   def complete!(_request_params)
     return if complete? || pending?
     start_completion!
+
+    # If there is an existing ramp election of the same type. Start using that instead.
+    # Potentially we should be destroying the temporary one here at the same time.
+    if existing_ramp_election
+      use_existing_ramp_election
+    end
 
     create_or_connect_end_product
 
@@ -120,11 +126,7 @@ class RampElectionIntake < Intake
   end
 
   def validate_detail_on_start
-    if matching_ramp_election.established?
-      self.error_code = :ramp_election_already_complete
-      @error_data = { receipt_date: matching_ramp_election.receipt_date }
-
-    elsif active_veteran_appeals.empty?
+    if active_veteran_appeals.empty?
       self.error_code = :no_active_appeals
 
     elsif active_compensation_appeals.empty?
@@ -138,8 +140,21 @@ class RampElectionIntake < Intake
     end
   end
 
-  def matching_ramp_election
-    @ramp_election_on_create ||= veteran_ramp_elections.all.first || veteran_ramp_elections.build
+  def new_intake_ramp_election
+    @ramp_election_on_create ||= veteran_ramp_elections.build
+  end
+
+  def existing_ramp_election
+    first = RampElection.established.where(
+      veteran_file_number: veteran_file_number,
+      option_selected: detail.option_selected).first
+    first == detail ? nil : first
+  end
+
+  def use_existing_ramp_election
+    old_detail = detail
+    update! detail: existing_ramp_election
+    old_detail.destroy!
   end
 
   def veteran_ramp_elections
--- a/app/models/ramp_election_intake.rb
+++ b/app/models/ramp_election_intake.rb
@@ -7,6 +7,8 @@ class RampElectionIntake < Intake
     no_active_compensation_appeals: "no_active_compensation_appeals",
     no_active_fully_compensation_appeals: "no_active_fully_compensation_appeals",
     no_active_appeals: "no_active_appeals",
+    # This is a legacy error_code from when we only allowed one RAMP election
+    ramp_election_already_complete: "ramp_election_already_complete",
     # This status will be set on successful intakes to signify that we had to
     # connect an existing EP, which in theory, shouldn't happen, but does in practice.
     connected_preexisting_ep: "connected_preexisting_ep"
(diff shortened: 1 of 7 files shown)


spec/models/ramp_election_intake_spec.rb
```diff
@@ -165,6 +163,111 @@
...
+          bgs_attrs: {
+            claim_type_code: "683SCRRRAMP",
+            claim_receive_date: former_detail.receipt_date.to_formatted_s(:short_date),
+            end_product_type_code: "683"
+          }
+        )
+      end
+
+      it "closes out legacy appeals" do
+        expect(RampClosedAppeal).to receive(:new).with(
```
shanear: I prefer not mocking method expectations for things that we can easily validate in a more robust way. I think the thing we want to validate here is that the `RampClosedAppeal` was created correctly, and we can do that afterwards via something like:

```
expect(RampClosedAppeal.where(
  vacols_id: appeals_to_close.first.vacols_id,
  ramp_election_id: former_detail.id,
  nod_date: appeals_to_close.first.nod_date
).to_not be_nil
```

spec/models/ramp_election_intake_spec.rb
```diff
@@ -165,6 +163,111 @@
...
+          nod_date: appeals_to_close.first.nod_date
+        ).and_call_original
+
+        expect(RampClosedAppeal).to receive(:new).with(
+          vacols_id: appeals_to_close.last.vacols_id,
+          ramp_election_id: former_detail.id,
+          nod_date: appeals_to_close.last.nod_date
+        ).and_call_original
+
+        expect(Fakes::AppealRepository).to receive(:close_undecided_appeal!).with(
```
shanear: These message expectations are fine, since they are mocking an external service that'd be mocked anyways. Bonus points if we write this test to use FACOLS and validate that the underlying FACOLS appeal is closed afterwards.

spec/models/ramp_election_intake_spec.rb
```diff
@@ -165,6 +163,111 @@
...
+          appeal: appeals_to_close.last,
+          user: intake.user,
+          closed_on: Time.zone.today,
+          disposition_code: "P"
+        )
+
+        subject
+      end
+
+      it "connects to the existing active end product" do
```
shanear: So the scenario here is really confusing, but `connected_preexisting_ep` and the new scenario of connecting to an existing ramp election are two different scenarios, so this isn't exactly what we want.

`connected_preexisting_ep` is a work around to when EPs are erroneously created outside of Caseflow without us knowing. It allows us to use an EP that was already there instead of making them cancel the EP and wait 2 days to reprocess the ramp election. Eg. people should **never** be creating EPs outside of Caseflow, so record an error_code, even though we have a work around. 

In the new scenario, it's not really an error. It's perfectly acceptable for a veteran to submit a ramp

Here's a test structure for this method that could work (you've got a lot of this already):

```
context "when an EP already exists" do
  let (:preexisting_ep) { ... }

  # this is where the `connected_preexisting_ep` code would happen
  it "closes out legacy appeals and connects the existing EP to the ramp election"
end

context "when an active RAMP election of the same type already exists" do
  let(:active_ramp_election) { ... }

  it "deletes the temporary ramp election and connects the intake to the existing election and recloses any new appeals"
end

context "when an active RAMP election of a different type exists" do
   it "creates new EP and closes appeals"
end

context "when an inactive RAMP election of the same type already exists" do
   it "creates new EP and closes appeals"
end
```

app/models/ramp_review.rb
```diff
@@ -43,7 +43,7 @@ def higher_level_review?
   #
   # Returns a symbol designating whether the end product was created or connected
   def create_or_connect_end_product!
-    return connect_end_product! if matching_end_product
+    return connect_end_product! if matching_end_product && matching_end_product.active?
```
shanear: Yeah i think you are getting two different scenarios mixed up here. See the comment below for more info.

spec/models/ramp_election_intake_spec.rb
kahlouie: @shanear It is still calling original. It's just making sure that it was called with the appropriate information. I can add those expectations as well at the end. :)

spec/feature/intake/ramp_election_spec.rb
```diff
@@ -142,13 +142,9 @@
       click_on "Search"
 
       expect(page).to have_content("Search for Veteran by ID")
-      expect(page).to have_content(
```
shanear: I think, we can probably drop this test outright

spec/models/ramp_election_intake_spec.rb
```diff
@@ -145,26 +155,129 @@
...
+    describe "if there are existing ramp elections" do
+      let(:former_detail) do
+        create(:ramp_election,
+               veteran_file_number: veteran_file_number,
+               notice_date: 40.days.ago,
+               option_selected: "supplemental_claim",
+               receipt_date: 38.days.ago,
+               established_at: 38.days.ago)
+      end
+      let!(:former_intake) do
```
shanear: I don't think this is necessary for the test

app/models/ramp_election_intake.rb
```diff
@@ -138,8 +141,22 @@ def validate_detail_on_start
...
+  def new_intake_ramp_election
+    @ramp_election_on_create ||= veteran_ramp_elections.build
+  end
+
+  def existing_ramp_election
+    first = RampElection.established.find_by(
+      veteran_file_number: veteran_file_number,
+      option_selected: detail.option_selected
+    )
+    (first == detail) ? nil : first
```
shanear: this can be done inside the query via:

```
 RampElection.established.where(
      option_selected: detail.option_selected
    ).where.not(id: detail_id).first
```

shanear: There is also nothing that's checking whether or not the EP is active on this election. I'll add that in.

app/models/ramp_election_intake.rb
```diff
@@ -138,8 +141,22 @@ def validate_detail_on_start
...
+  end
+
+  def use_existing_ramp_election
+    old_detail = detail
+    update! detail: existing_ramp_election
+    old_detail.destroy!
```
shanear: this could use a transaction to make these two operations atomic

Issue #5562, opened 2018-05-16
Vacols Integration | Query hearing schedule master records by RO


sharonwarner:
We currently pull hearings in by judge and by appeal. We need to be able to pull hearings in by RO so that ROs can see their upcoming hearings.

AC:
1. From the Rails console, we can run HearingDay.repository.load_days_for_regional_office(RO) and it returns all upcoming hearings for that RO.

PR #5967, opened 2018-06-19
Select hearing days by regional office.

oscarAtNava:
If no dates provided, date range defaulted.

Connects #5562 

### Description
Query for hearings days specific to a regional office. Optional date range may be provided. If dates not provided it defaults to a default date range: one month prior to infinity.

### Acceptance Criteria 
- [ ] Code compiles correctly
- [ ] Request test cases all green (./spec/requests/hearing_day_spec.rb)

### Testing Plan
In Rails console run the following command:

`HearingDay.load_days_for_regional_office("RO17", "2017-01-01", "2017-12-31")`

If your FACOLS data is same as mine you will get one TB record back. There are no video hearings in FACOLS. I modified one CaseHearing record to have the appropriate format to be able to get it back with an RO query so I go the result below: one video and one TB row.

`[[#<VACOLS::CaseHearing:0x007ffd51abac60
   hearing_pkseq: 404611,
   hearing_type: "V",
   folder_nr: "VIDEO RO17",
   hearing_date: Tue, 11 Jul 2017 10:30:00 UTC +00:00,
   board_member: "595",
   room: "4",
   mduser: "JWEHNER",
   mdtime: Tue, 11 Jul 2017>],`

`[#<VACOLS::TravelBoardSchedule:0x007ffd52270d10
   tbyear: "2017",
   tbtrip: 31,
   tbleg: 1,
   tbro: "RO17",
   tbstdate: Thu, 17 Aug 2017,
   tbenddate: Mon, 21 Aug 2017,
   tbmem1: "598",
   tbmem2: "748",
   tbmem3: "738",
   tbmem4: nil,
   tbaty1: nil,
   tbaty2: nil,
   tbaty3: nil,
   tbaty4: nil,
   tbadduser: "VHERMISTO",
   tbaddtime: Fri, 17 Feb 2017,
   tbmoduser: "VHERMISTO",
   tbmodtime: Sun, 26 Mar 2017,
   tbcancel: nil,
   tbbvapoc: "Payton McClure (329) 272-7123 x966",
   tbropoc: "Buck Reichert (774) 659-0916">]]`


app/controllers/hearings/hearing_day_controller.rb
```diff
@@ -5,12 +5,18 @@ class Hearings::HearingDayController < ApplicationController
   # Controller to add and update hearing schedule days.
 
   # show schedule days for date range provided
+  # # rubocop:disable Metrics/MethodLength
```
tejans24: Is it possible to just split this method rather than disabling rubocop?

app/controllers/hearings/hearing_day_controller.rb
```diff
@@ -5,12 +5,18 @@ class Hearings::HearingDayController < ApplicationController
...
-    # rubocop:disable Metrics/LineLength
-    @start_date = params[:start_date].nil? ? (Time.zone.today.beginning_of_day - 365.days) : Date.parse(params[:start_date])
-    # rubocop:enable Metrics/LineLength
-    @end_date = params[:end_date].nil? ? Time.zone.today.beginning_of_day : Date.parse(params[:end_date])
-    video_and_co, travel_board = HearingDay.load_days_for_range(@start_date, @end_date)
+    @start_date = validate_start_date(params[:start_date])
+    @end_date = validate_end_date(params[:end_date])
+    regional_office = validate_regional_office(params[:regional_office])
+
+    if regional_office.nil?
```
sharonwarner: We should move some of this logic into the model

app/controllers/hearings/hearing_day_controller.rb
```diff
@@ -64,16 +71,27 @@ def hearing
...
+
+  def validate_start_date(start_date)
+    start_date.nil? ? (Time.zone.today.beginning_of_day - 30.days) : Date.parse(start_date)
+  end
+
+  def validate_end_date(end_date)
+    end_date.nil? ? (Time.zone.today.beginning_of_day + 365.days) : Date.parse(end_date)
+  end
+
+  def validate_regional_office(regional_office)
```
sharonwarner: Is this just a placeholder?

spec/requests/hearing_day_spec.rb
```diff
@@ -93,4 +96,31 @@
...
+      )
+      Generators::Vacols::TravelBoardSchedule.create(tbmem1: "955")
+      Generators::Vacols::Staff.create(sattyid: "955")
+    end
+
+    it "Get hearings for RO" do
+      hearings
+      headers = {
+        "ACCEPT" => "application/json",     # This is what Rails 4 accepts
+        "HTTP_ACCEPT" => "application/json" # This is what Rails 3 accepts
```
tejans24: Curious, do we need this Rails 3 http accept?

app/models/vacols/case_hearing.rb
```diff
@@ -127,12 +133,11 @@ def select_schedule_days
...
              :hearing_type,
              :folder_nr,
-             :repname,
              :room,
              :board_member,
              :mduser,
-             :mdtime,
-             :canceldate)
+             :mdtime)
+        .where("folder_nr is null or folder_nr like ?", "VIDEO %")
```
sharonwarner: Does this only return master records? I think this will break hearing prep.

app/models/vacols/case_hearing.rb
```diff
@@ -127,12 +133,11 @@ def select_schedule_days
              :hearing_date,
              :hearing_type,
              :folder_nr,
-             :repname,
              :room,
              :board_member,
              :mduser,
-             :mdtime,
-             :canceldate)
```
tejans24: is `:canceldate` not being used anywhere?

app/controllers/hearings/hearing_day_controller.rb
tejans24: is this method needed?

app/controllers/hearings/hearing_day_controller.rb
oscarAtNava: Isn't it better to validate dates as close to the client as possible?

app/models/vacols/case_hearing.rb
oscarAtNava: This method is new and is not used by hearing prep.

app/controllers/hearings/hearing_day_controller.rb
oscarAtNava: It is a place holder for when I flesh out the required validations. Probably tie to what the spreadsheet validation is doing.

oscarAtNava: Yes.

app/models/vacols/case_hearing.rb
oscarAtNava: Cancel date is a hearing concept, not a scheduled hearing concept, in the new model.

app/controllers/hearings/hearing_day_controller.rb
oscarAtNava: Actually I should remove that as I did reduce the method by moving the date logic to separate methods.

spec/requests/hearing_day_spec.rb
oscarAtNava: It seems to only work with both entries.

app/controllers/hearings/hearing_day_controller.rb
oscarAtNava: That was removed in a later version of the file.

app/models/vacols/case_hearing.rb
sharonwarner: Oh sorry! The way the line numbers worked out, I thought you were adding this to select_hearings.

app/controllers/hearings/hearing_day_controller.rb
sharonwarner: Sure, we could also add methods to the controller. I just don't think we should have to disable the method length rubocop metric.

PR #5970, opened 2018-06-20
Refactor legacy task routes

### LINKED ISSUE #5969: [BE] - Refactor legacy tasks endpoints into its own controller
1. create legacy_task_controller.rb with the following actions:
* create (judge case assignment to attorney)
* update (judge case reassignment to another attorney)

2. create case_review_controller with the following
* complete (attorney case review, judge case review)

3. leave tasks_controller.rb for new tasks (co-located, ama appeals, etc)
* leave index the way it is for now (will return legacy tasks for attorneys and judges)

aroltsch:
Fixes #5969 

1. create legacy_task_controller.rb with the following actions:
* create (judge case assignment to attorney)
* update (judge case reassignment to another attorney)

2. create case_review_controller with the following
* complete (attorney case review, judge case review)

3. leave tasks_controller.rb for new tasks (co-located, ama appeals, etc)
* leave index the way it is for now (will return legacy tasks for attorneys and judges)


config/routes.rb
```diff
@@ -138,14 +138,11 @@
     get '/appeals/:vacols_id', to: 'queue#index'
     get '/appeals/:vacols_id/*all', to: redirect('/queue/appeals/%{vacols_id}')
     get '/:user_id(*rest)', to: 'tasks#index'
-
-    post '/appeals/:id/complete', to: 'tasks#complete'
```
aroltsch: I am not seeing this route is used anywhere.

aroltsch: @lowellrex could you confirm please

app/controllers/issues_controller.rb
```diff
@@ -1,5 +1,5 @@
 class IssuesController < ApplicationController
-  before_action :verify_task_completion_access
+  before_action :verify_case_review_access
```
aroltsch: Renamed the method to a better name

app/controllers/tasks_controller.rb
```diff
@@ -1,21 +1,11 @@
 class TasksController < ApplicationController
   before_action :verify_queue_access
-  before_action :verify_task_completion_access, only: :complete
-  before_action :verify_task_assignment_access, only: [:create, :update]
-
-  rescue_from Caseflow::Error::VacolsRepositoryError do |e|
```
aroltsch: Moved this to the application controller

app/models/judge_case_assignment_to_attorney.rb
```diff
@@ -2,7 +2,7 @@ class JudgeCaseAssignmentToAttorney
   include ActiveModel::Model
   include LegacyTaskConcern
 
-  attr_accessor :appeal_id, :assigned_to, :task_id, :assigned_by, :type
```
aroltsch: This class doesn't need `type` anymore.

config/routes.rb
dannysellers: Confirmed!

Issue #5632, opened 2018-05-21
Create appellant model & table


mdbenjam:
After [this discussion](https://github.com/department-of-veterans-affairs/caseflow/issues/5536) we decided that we want a many to many relationship between appeals and appellants. We need to create an appellant table with the following fields.

```
1) id
2) participant_id
```

The participant_id comes from BGS and allows us to make queries to determine appellant information.

The model should implement the following methods:

```
appellant_name,
appellant_address_line_1,
appellant_address_line_2,
appellant_city,
appellant_state,
appellant_zip,
appellant_country,
appellant_relationship
```

This information should all be available in BGS.

Finally, once we have this model we can update the newly created `WorkQueue::AppealSerializer` to use these appellant methods.

# AC
- [ ] There should be a new appellant table.
- [ ] There should be a new appellant model with the methods described above.
- [ ] The `WorkQueue::AppealSerializer` should call the new appellant model methods.

lowellrex:
Future me: don't forget about this puppy: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/concerns/associated_bgs_record.rb

PR #5971, opened 2018-06-20
Fetch claimant information from BGS

lowellrex:
Connects #5632.

Adds attributes for Claimant model using information fetched from BGS, and returns those attributes in the `WorkQueue::AppealSerializer` if a claimant exists for the appeal.

### Testing Plan
1. Deploy branch to UAT
2. Look for test Veteran in [test data](https://github.com/department-of-veterans-affairs/appeals-qa/blob/master/TestData/TestCertificationData.md), find one that looks promising and make a request to BGS to get the participant ID
```
rails> u = User.first
rails> RequestStore[:current_user] = u
rails> bgs = ExternalApi::BGSService.new
rails> bgs.fetch_poa_by_file_number(id) # For C file numbers (no trailing C)
rails> bgs.fetch_file_number_by_ssn(id) # For SSNs
```
3. Check to see if we fetch data from BGS properly
```
rails> u = User.first
rails> RequestStore[:current_user] = u
rails> participant_id = {{12345}}
rails> c = Claimant.create(review_request_type: "review_request_type", review_request_id: 1, participant_id: participant_id)
rails> c.address_line_1 # Kicks off BGS fetch and returns some value
rails> c.relationship # Will be null because BGS UAT sandbox does not have data here. Works in prod.
```


app/models/claimant.rb
```diff
@@ -1,9 +1,25 @@
 class Claimant < ApplicationRecord
+  include AssociatedBgsRecord
+
   belongs_to :review_request, polymorphic: true
 
+  bgs_attr_accessor :name, :relationship,
```
leikkisa: I don't see relationship on the example response, do you mean payee_type_name?

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -8,18 +8,21 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
   end
 
   attribute :appellant_full_name do
-    "not implemented"
+    object.claimants[0]. if object.claimants && object.claimants.any?
```
mdbenjam: I think this missing something after the `object.claimants[0].`?

app/mappers/address_mapper.rb
```diff
@@ -11,4 +11,13 @@ def get_address_from_bgs_address(bgs_address)
       zip: bgs_address[:zip_prefix_nbr]
     }
   end
+
+  # TODO: Confirm that the payeen name is coming in with first and last names reversed.
+  def get_name_and_address_from_bgs_info(bgs_info)
+    return {} unless bgs_info
+    {
+      name: bgs_info[:payee_name] && bgs_info[:payee_name].gsub(/\s+/, " "),
```
mdbenjam: There is also a generic person service. Curious if that gives better formatted output. This one seems fine though.

lowellrex: I don't think that one returns us the relationship which we also want. If this formatting becomes an issue then I think we can make a third call to BGS to get a better formatted name from the Person service.

app/models/claimant.rb
lowellrex: We change `payee_type_name` to `relationship` in the [AddressMapper](https://github.com/department-of-veterans-affairs/caseflow/pull/5971/files#diff-6866c8f05313a2a158bddd00e097a588R20) which gets called at the end of [`ExternalApi::BGSService.fetch_claimant_info_by_participant_id()`](https://github.com/department-of-veterans-affairs/caseflow/pull/5971/files#diff-27c4626c4d3c4c1589aae6dc8c287c39R143) (which in turn gets called by `Claimant.fetch_bgs_record()` by way of the `AssociatedBgsRecord` trait).

Issue #5722, opened 2018-05-30
Queue | Make view appeal id arguments consistent


dannysellers:
Currently, we use both [`vacolsId`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/QueueApp.jsx#L150) and [`appealId`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/QueueApp.jsx#L170) within routes in `QueueApp` and within Queue views. We should make these consistent.

My suggestion is to use `appealId`, which seems more future-proof than `vacolsId`.

This will also allow us to return to `/queue/appeals/${appealId}` when canceling a checkout flow (rather than `/queue`) ([ref](https://github.com/department-of-veterans-affairs/caseflow/pull/5977#pullrequestreview-132447301))

PR #5977, opened 2018-06-20
Queue | Connect judge checkout to dispatch FE to BE

### LINKED ISSUE #5152: Judge check out | OMO confirmation screen
After a judge has assigned a case to OMO, they need to see a confirmation screen that the case has been sent to then next team. 

## AC
- After a judge selects 'Assign to OMO' from the dropdown, the confirmation banner appears at the top of their table.
- Copy: 
  - Title: You have successfully submitted an OMO for [Veteran First Name] [Last name]
  - Body: If you need to make any changes, please manage this case in DAS.

## Mocks
queue confirmation omo

### LINKED ISSUE #5153: Judge check out | Decision confirmation screen
After a judge has finished the 'Evaluate Decision' screen and submitted the case, they need to see a confirmation screen that the case has been sent to then next team. 

## AC
- After a judge selects 'Submit' on the Evaluate Decision screen, they are taken back to their table:
   - the case is assigned to team 31 or 32 TODO - goal: all cases are sent to location 30
   - the case is no longer listed in the table
   - the confirmation banner appears at the top of the table.
- Copy: 
- Title: Thanks for reviewing Vera Marshall's decision
- Body: If you need to make any changes, please manage this case in DAS.

## Mocks
queue confirmation banner dispatch

dannysellers:
Resolves #5152, resolves #5153

### Description
Connects FE & BE for Judge checkout to Dispatch flows

### Acceptance Criteria 
- [ ] Ready for Dispatch and Assign OMO flows work as expected

### Testing Plan
1. Go to `/queue` as `BVAAABSHIRE` / a judge
2. All of `BVAAABSHIRE`'s default cases are draft decisions: Go to case detail view, select Ready for Dispatch, complete flow (review dispositions, review remand reasons if applicable, evaluate decision), confirm you're navigated back to `/queue` with success message
3. To test OMO flow, switch users to attorney (`BVASCASPER1`) and complete Medical Request Ready for Review checkout flow, assigning to Anjali Abshire
4. Switch back to `BVAAABSHIRE`, navigate to case detail view for OMO case, choose Assign OMO from dropdown, confirm you're navigated back to `/queue` with success message.

### TODO
- Enable/disable judge checkout dropdown options based on document id / decass deprod attr #5998 

<details><summary>Assign to OMO</summary>

</details>

<details><summary>Ready for Dispatch</summary>

</details>


spec/factories/vacols/decass.rb
```diff
@@ -7,5 +7,6 @@
     deadusr "TEST"
     deadtim { Time.current.strftime("%Y-%m-%d") }
     deassign { Time.current.strftime("%Y-%m-%d") }
+    deprod nil
```
dannysellers: Initially, I tried doing something like
```ruby
trait :omo_request do
  deprod %w[IME VHA OTI OTV].sample
end

trait :draft_decision do
  deprod %w[DEC OTD OTR... VRM].sample
end
```
but I wasn't sure how to alter the [`create_list` arguments in `factories/vacols/case`](https://github.com/department-of-veterans-affairs/caseflow/blob/e5563c6d411703c376300b37ed4381d9775c963a/spec/factories/vacols/case.rb#L32) to pass either `:omo_request` or `:draft_decision`, instead of just setting the value in a key-value pair

spec/feature/queue/queue_spec.rb
```diff
@@ -504,23 +510,409 @@
     end
   end
 
+  context "loads attorney checkout views" do
```
dannysellers: Attorney checkout flow tests were removed when we removed the start checkout dropdown from the AttorneyTaskTable. These are only updated to start checkout flows from in the case details view

client/app/queue/components/DecisionViewBase.jsx
```diff
@@ -95,6 +95,7 @@ export default function decisionViewBase(ComponentToWrap) {
       this.props.resetDecisionOptions();
       _.each(stagedAppeals, this.props.checkoutStagedAppeal);
 
+      // todo: checkout flow now starts from within case details page--return there on cancel?
```
dannysellers: @sneha-pai @mkhandekar As you know, we recently removed the start-checkout-flow dropdown from the Attorney / Judge review tables, requiring the user to go into the case detail view to start the process. If the user is in the checkout process, and they cancel it, should we send them back to `/queue`? or to the appeal detail page (the start point)? `/queue/appeals/${appealId}`?

spec/factories/vacols/staff.rb
```diff
@@ -33,5 +33,11 @@
     trait :has_location_code do
       slogid "55"
     end
+
+    after(:build) do |staff, evaluator|
```
dannysellers: Not sure if this should be `after(:create)`, or if it'll get called either way. We always call `create(:staff, ...)` @mdbenjam

mdbenjam: This should be fine, build gets called in the create lifecycle. You only need `after(:create)` if you need access to the primary key id (which gets assigned when the record is saved)

client/app/queue/components/DecisionViewBase.jsx
mkhandekar: I'd say if the actions dropdown now _only_ lives in case details, then let's send them back to case details.

dannysellers: It turns out this will be a bit more involved than I anticipated (we have inconsistency in some parameter names, see #5722), but I'll start working on it after this PR is merged

mkhandekar: @dannysellers let us know if this inconsistency is more than code-related!

dannysellers: It's only code-related, you're good! In some views, we're passing the appeal id as `vacolsId`, while in others, we're using `appealId`, which feels more future-proof. Just gotta go with one or the other

client/app/queue/utils.js
```diff
@@ -103,3 +106,40 @@ export const getIssueDiagnosticCodeLabel = (code) => {
...
+
+    _.extend(payload.data.tasks, args);
+  }
+
+  payload.data.tasks.issues = issueList.map((issue) => _.extend({},
+    _.pick(issue, ['vacols_sequence_id', 'remand_reasons', 'type', 'readjudication']),
+    { disposition: _.capitalize(issue.disposition) }
+  ));
+
+  return payload;
```
lowellrex: I'm having a really hard time understanding what is happening in this function and think we should think of different ways to address this complexity (split this up into many different functions and aim for functions with a single conditional? Make this an object that consumes these arguments and has a `buildPayload()` function?).

client/app/queue/utils.js
```diff
@@ -103,3 +106,40 @@ export const getIssueDiagnosticCodeLabel = (code) => {
...
+ * @param {Array} issues
+ * @param {Object} args
+ * @returns {Object}
+ */
+export const buildCaseReviewPayload = (decision, userRole, issues, args = {}) => {
+  const payload = {
+    data: {
+      tasks: {
+        type: `${userRole}CaseReview`,
+        ...decision.opts
```
lowellrex: We want to accept anything in the `decision.opts` attribute for this payload (guessing from the [default value in the state](https://github.com/department-of-veterans-affairs/caseflow/blob/fdabe80845cfcf5c7359386bd7f6f7638f5e64e6/client/app/queue/reducers.js#L35)), but I find the spread operator kind of spooky since I guess we could be putting anything in the tasks hash (including overwriting the `type` value we set directly above, no?). Is there any way we can be more explicit with what keys we want from `decision.opts`?

client/app/queue/utils.js
lowellrex: By my count we have--effectively--seven input arguments to this function:
* decision.type
* decision.opts
* userRole
* issues
* args.factors_not_considere
* args.factors_not_considere
* left over `args` values which will be merged with the `tasks` attribute of the eventual payload

client/app/queue/utils.js
```diff
@@ -103,3 +106,40 @@ export const getIssueDiagnosticCodeLabel = (code) => {
...
+
+  if (userRole === USER_ROLES.ATTORNEY) {
+    issueList = getUndecidedIssues(issues);
+
+    _.extend(payload.data.tasks, { document_type: decision.type });
+  } else {
+    args.factors_not_considered = _.keys(args.factors_not_considered);
+    args.areas_for_improvement = _.keys(args.areas_for_improvement);
+
+    _.extend(payload.data.tasks, args);
```
lowellrex: I don't love how we are blindly expanding the values of the `args` input into the `tasks` attribute of the payload. Is there any way we could be more explicit here?

spec/factories/vacols/case.rb
```diff
@@ -14,12 +14,29 @@
...
-        create_list(:decass, evaluator.decass_count, defolder: vacols_case.bfkey, deadusr: slogid ? slogid : "TEST")
+        create_list(
+          :decass,
+          evaluator.decass_count,
+          evaluator.work_product,
+          defolder: vacols_case.bfkey,
+          deadusr: slogid ? slogid : "TEST",
+          demdusr: assigner_slogid ? assigner_slogid : "ASSIGNER",
+          dereceive: (evaluator.user && evaluator.user.vacols_roles.first.eql?("judge")) ? Time.zone.today : nil
+        )
```
lowellrex: This looks great!

Issue #5936, opened 2018-06-18
Allow an attorney to assign multiple admin actions to a co-located user


aroltsch:
1. Allow an attorney to assign multiple admin actions to a co-located user
2. Create a new task for each admin action

PR #5978, opened 2018-06-20
Anya/5936 multiple admin actions 2

aroltsch:
Connects #5936

Allow attorneys to assign multiple admin actions at once.


app/controllers/tasks_controller.rb
```diff
@@ -33,11 +33,13 @@ def index
   end
 
   def create
+    return required_parameters_missing([:titles]) if task_params[:titles].blank?
```
aroltsch: This parameter cannot be validated at the model level so I included it here.

app/models/tasks/co_located_admin_action.rb
```diff
@@ -6,11 +6,14 @@ class CoLocatedAdminAction < Task
...
       ActiveRecord::Base.multi_transaction do
-        record = super
-        if record.valid?
-          AppealRepository.update_location!(record.appeal, LegacyAppeal::LOCATION_CODES[:caseflow])
+        records = params.delete(:titles).each_with_object([]) do |title, result|
+          result << super(params.merge(title: title))
+          result
         end
-        record
+        if records.map(&:valid?).uniq == [true]
```
aroltsch: Update the location in Vacols only if all records are valid.

Issue #5774, opened 2018-06-04
Excel Integration | Save data to database


sharonwarner:
From the command line, we should be able to:
1) Save the validated spreadsheet's data to the database

PR #5980, opened 2018-06-21
Importing judge non-availability

sharonwarner:
Connects #5774 

### Description
This adds functionality for saving judge non-availability days to the database.


app/services/hearing_schedule/validate_judge_spreadsheet.rb
```diff
@@ -14,48 +10,34 @@ class JudgeDatesNotInRange < StandardError; end
   class JudgeNotInDatabase < StandardError; end
 
   def initialize(spreadsheet, start_date, end_date)
-    @spreadsheet = spreadsheet
+    @spreadsheet_template = HearingSchedule::GetSpreadsheetData.new(spreadsheet).judge_non_availability_template
+    @spreadsheet_data = HearingSchedule::GetSpreadsheetData.new(spreadsheet).judge_non_availability_data
```
tejans24: We can just do this with one instance of `GetSpreadsheetData` rather than creating two different objects

Issue #5927, opened 2018-06-14
NOD appeal - need to check from receipt date not today's date


leahbannon:
When we check for NOD timeliness, we've been checking 370 days from today -- it needs to be from the receipt date of the form.

leikkisa:
Sally to create data with ratings date a week apart in June

leikkisa:
There aren't any ratings that meet our current needs, so I've emailed Colin to ask him to add it.

leikkisa:
The rating is in the data, but it doesn't have the right promulgation_date. I'll update here again once the data is ready.

leikkisa:
Just noting that they "fixed" the data, but accidentally put everything back 10 years too far into the past, so I requested another fix to the data.  I think that fell through the cracks (because it is not fixed) so just sent another request.

leikkisa:
The data for this is ready! It is on file number: 984562385.  Given a receipt date of 12/1/2017, there should be two ratings on it. If today's date was still being used (the bug being fixed), there would only be one rating.  There is also a third rating on this veteran that is not timely and shouldn't be showing up in either case.

suzchap:
Receipt date of 12/1/2017 throws the error "Receipt Date cannot be earlier than the AMA pilot start date."

leikkisa:
Okay, I think this is ready for realz! If you try a receipt date of today, it should have 1 rating, and if you try a receipt date of 4/2/2018, it should have TWO ratings.

PR #5986, opened 2018-06-21
Fix timeliness date to be based on receipt date instead of current date

leikkisa:
Connects #5927

### DIFF
--- a/app/models/concerns/cached_attributes.rb
+++ b/app/models/concerns/cached_attributes.rb
@@ -21,11 +21,22 @@ module CachedAttributes
   end
 
   def cache_id(attr_name)
-    "#{self.class.name}-#{id}-cached-#{attr_name}"
+    "#{self.class.name}-#{cache_key(attr_name)}-cached-#{attr_name}"
+  end
+
+  def cache_key(attr_name)
+    send(self.class.cache_key(attr_name))
   end
 
   module ClassMethods
+    def cache_key(attr_name)
+      @cache_keys[attr_name]
+    end
+
     def cache_attribute(attr_name, write_options = {}, &get_value)
+      @cache_keys ||= {}
+      @cache_keys[attr_name] = write_options.delete(:cache_key) || :id
+
       define_method "#{attr_name}=" do |value|
         set_cached_value(attr_name, value)
       end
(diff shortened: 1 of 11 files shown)


app/models/concerns/cached_attributes.rb
```diff
@@ -21,11 +21,24 @@ def set_cached_value(attr_name, value, write_options = {})
...
-    "#{self.class.name}-#{id}-cached-#{attr_name}"
+    "#{self.class.name}-#{cache_key(attr_name)}-cached-#{attr_name}"
+  end
+
+  def cache_key(attr_name)
+    send(self.class.cache_key(attr_name))
   end
 
   module ClassMethods
+    # rubocop:disable Style/ClassVars
```
shanear: This is sort of awkward but i think its necessary here. I think it might be worth adding a comment about how we actually want class variables here, because we want the variable to be accessible in subclasses

app/models/rating.rb
```diff
@@ -36,15 +36,15 @@ def fetch_issues
...
       response = BGSService.new.fetch_ratings_in_range(
         participant_id: participant_id,
-        start_date: Time.zone.today - TIMELY_DAYS,
+        start_date: from_date - TIMELY_DAYS,
         end_date: Time.zone.today
       )
 
       unsorted = ratings_from_bgs_response(response).select do |rating|
-        rating.promulgation_date > (Time.zone.today - 372)
+        rating.promulgation_date > (from_date - TIMELY_DAYS)
```
shanear: oooh nice catch in using `TIMELY_DAYS`

spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -22,22 +22,38 @@
 
   let(:inaccessible) { false }
 
+  let(:receipt_date) { Date.new(2018, 4, 20) }
```
shanear: Really nice modifications to the feature tests 💯

app/models/concerns/cached_attributes.rb
leikkisa: Good call! Yeah I looked up the linting error and the solution I saw looked like it wouldn't fit for us so decided to keep it.

shanear:
Figured I'd document my debugging process:

Added a pry right before the line in the test that was failing:
screen shot 2018-06-23 at 3 02 56 pm

Then checked out the screen. Looks like there was a 500 which was causing the page to hang:
screen shot 2018-06-23 at 3 02 46 pm

Started up `tail -f /log/test.log` and reran the test:
screen shot 2018-06-23 at 3 02 12 pm

Issue #5688, opened 2018-05-29
Move Queue tests to FACOLS


mdbenjam:
Queue tests in the following files are failing. We need to move them to FACOLS.

- [x] spec/feature/queue/case_details_spec.rb
- [x] spec/feature/queue/checkout_flows_spec.rb
- [x] spec/feature/queue/search_spec.rb
- [x] spec/feature/queue/task_queue_spec.rb
- [x] spec/models/queues/work_queue_spec.rb
- [x] spec/controllers/tasks_controller_spec.rb
- [x] spec/models/attorney_case_review_spec.rb

PR #5995, opened 2018-06-21
Move AttorneyCaseReview test to FACOLS

lowellrex:
Connects #5688.

### DIFF
--- a/spec/models/attorney_case_review_spec.rb
+++ b/spec/models/attorney_case_review_spec.rb
@@ -1,76 +1,53 @@
 describe AttorneyCaseReview do
-  let(:judge) { User.create(css_id: "CFS123", station_id: User::BOARD_STATION_ID) }
-  let(:attorney) { User.create(css_id: "CFS456", station_id: "317") }
+  let(:attorney) { FactoryBot.create(:user) }
+  let(:judge) { FactoryBot.create(:user, station_id: User::BOARD_STATION_ID) }
 
-  before do
-    Fakes::Initializer.load!
-  end
+  before { FeatureToggle.enable!(:test_facols) }
+  after { FeatureToggle.disable!(:test_facols) }
 
   context ".complete" do
+    let(:vacols_staff) { FactoryBot.create(:staff, :attorney_role, sdomainid: attorney.css_id) }
+    let(:case_issues) { [] }
+    let(:vacols_case) { FactoryBot.create(:case, :assigned, staff: vacols_staff, case_issues: case_issues) }
+    let(:document_type) { "omo_request" }
+    let(:work_product) { "OMO - IME" }
+    let(:task_id) { "#{vacols_case.bfkey}-#{vacols_case.decass[0].deadtim.strftime('%F')}" }
+    let(:params) do
+      {
+        document_type: document_type,
+        reviewing_judge: judge,
+        work_product: work_product,
+        document_id: "123456789.1234",
+        overtime: true,
+        note: "something",
+        task_id: task_id,
+        attorney: attorney
+      }
+    end
+
+    before { User.authenticate!(user: attorney) }
+
     subject { AttorneyCaseReview.complete(params) }
 
     context "should validate format of the task ID" do
-      let(:params) do
-        {
-          document_type: "omo_request",
-          reviewing_judge: judge,
-          work_product: "OMO - IME",
-          document_id: "123456789.1234",
-          overtime: true,
-          note: "something",
-          task_id: task_id,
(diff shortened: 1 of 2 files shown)


spec/models/attorney_case_review_spec.rb
```diff
@@ -1,76 +1,53 @@
 describe AttorneyCaseReview do
-  let(:judge) { User.create(css_id: "CFS123", station_id: User::BOARD_STATION_ID) }
-  let(:attorney) { User.create(css_id: "CFS456", station_id: "317") }
+  let(:attorney) { FactoryBot.create(:user) }
```
mdbenjam: FYI in tests you can just call `create` and `build` without the preceding `FactoryBot.`

lowellrex: I like including the module namespace as a reminder of where the function is coming from, at least until it becomes second nature that all of our tests use `FactoryBot`'s methods.

spec/models/attorney_case_review_spec.rb
```diff
@@ -82,89 +59,11 @@
       end
     end
 
-    context "when all parameters are present for draft decision and VACOLS update is successful" do
-      before do
-        allow(Fakes::QueueRepository).to receive(:reassign_case_to_judge!).with(
```
mdbenjam: Now that we're not checking if this method is called, can we check to see if the underlying data is properly changed? Not super familiar with this code, but looks like `deadusr` should be updated to be the judge?

app/mappers/queue_mapper.rb
```diff
@@ -43,15 +43,15 @@ module QueueMapper
...
+    "VHA" => "OMO - VHA"
   }.freeze
 
   OVERTIME_WORK_PRODUCTS = {
-    OTD: "Decision",
-    OTI: "OMO - IME",
-    OTV: "OMO - VHA"
+    "OTD" => "Decision",
+    "OTI" => "OMO - IME",
+    "OTV" => "OMO - VHA"
```
lowellrex: Changed symbols to strings to match what we expect to insert/retrieve from VACOLS.

spec/rails_helper.rb
```diff
@@ -120,18 +120,19 @@ def stub=(user)
...
+          "user" =>
+            { "id" => css_id || "DSUSER",
+              "name" => "Lauren Roth",
+              "station_id" => "283",
+              "email" => "test@example.com",
+              "roles" => roles || ["Certify Appeal"] }
+        )
+      end
+
+      RequestStore.store[:current_user] = user
```
lowellrex: Set the user in the RequestStore because we still use `RequestStore.store[:current_user]` instead of `current_user` in a handful of places. Alternatively, I could change where we call RequestStore instead of current_user [in `VACOLS::Case.update_vacols_location!()`](https://github.com/department-of-veterans-affairs/caseflow/blob/5ddc24571b0aad8ba631d0d1169454343b4c4028/app/models/vacols/case.rb#L219) if that makes more sense.

app/mappers/queue_mapper.rb
mdbenjam: So was this broken before?

spec/rails_helper.rb
mdbenjam: I think this is fine.

app/mappers/queue_mapper.rb
lowellrex: This was not broken before, but we were relying on the database connection to convert the symbol into a string before it was inserted into the database.

Since we are inserting the string and not the symbol into the database (and not using these particular symbols for any other purpose than to be strings that we insert into the database), I thought it made sense to explicitly have them be strings on not rely on some opaque Rails magic (and oracle db connection adapter behaviour) here.

### Demonstration that this wasn't broken before
clip

### Code path for conversion from symbol to string
[ActiveRecord.update_all()](https://apidock.com/rails/ActiveRecord/Base/update_all/class) > [ActiveRecord.sanitize_sql_for_assignment()](https://apidock.com/rails/v3.1.0/ActiveRecord/Base/sanitize_sql_for_assignment/class) > [ActiveRecord.sanitize_sql_hash_for_assignment()](https://apidock.com/rails/ActiveRecord/Base/sanitize_sql_hash_for_assignment/class) > [ActiveRecord.quote_bound_value()](https://apidock.com/rails/v3.1.0/ActiveRecord/Base/quote_bound_value/class) > [ActiveRecordConnection.quote()](https://apidock.com/rails/v1.2.6/ActiveRecord/Base/quote)

Issue #5983, opened 2018-06-21
Dropdown button


sharonwarner:
AC:
- Update DropdownMenu.jsx in the frontend-toolkit to allow button styling (primary and secondary)
- Update the node-module in Caseflow
- Add the dropdown button to the styleguide

![dropdownbutton](https://user-images.githubusercontent.com/4306128/41726241-b724899c-753f-11e8-9f2d-f473ed35d222.png)

nolawi:
So looking at this ticket what is proposed is that we make the dropdown menu look like a button as a secondary option.
<img width="483" alt="screen shot 2018-06-21 at 2 04 49 pm" src="https://user-images.githubusercontent.com/1745984/41737058-2e5e040e-755c-11e8-81f2-4b6ee1b2ad51.png">
 
But most of the styles will be inherited a button. The dropdown menu is already very complex with `searchable ` option as well. 

I think it might be better to create a new dropdown button component that inherits the button but has an arrow and options as the current dropdown menu is very complex already. LMK

sharonwarner:
I'm fine with either approach, but I meant the DropdownMenu component. It's the one we use in the navigation bar to login/logout. It's not nearly as complicated as the SearchableDropdown you're describing.

PR #6001, opened 2018-06-22
Dropdown Btn comp

nolawi:
Connects #5983

### Description
Creates New Dropdown Button component

screen shot 2018-06-26 at 12 56 41 pm

### Acceptance Criteria 
- [ ] Button Dropdown works as expected

### Testing Plan
1. Go tohearings/schedule/build click button

### TODO
1. Add to styleguide - have separate PR


client/app/components/DropdownButton.jsx
```diff
@@ -0,0 +1,68 @@
...
+      {this.state.menu && dropdownButtonList() }
+
+    </div>;
+  }
+}
+
+DropdownButton.propTypes = {
+  list: PropTypes.arrayOf(PropTypes.shape({
+    title: PropTypes.string.isRequired,
+    link: PropTypes.string.isRequired,
```
sharonwarner: Is this required? We're not setting it in the example below.

sharonwarner: What is the difference between link and target?

nolawi: yap need to remove link, it should just be target

Issue #5628, opened 2018-05-21
URL for editing Higher Level Review and Supplemental Claim issues


shanear:
We need a URL pattern for allowing any downstream case workers to edit issues. This pattern should either take the claim id of the Higer

## Acceptance criteria
- When the URL pattern for a review is visited, show a mostly blank page with the name of the veteran.

## Technical notes
- Don't worry if the review is cleared, we will handle that in a future story
- Don't worry about the link for appeals, we will handle that in a future story

leikkisa:
@shanear 
I'm guessing the description above is meant to say "This pattern should either take the claim id of the Higher Level Review or Supplemental Claim", and not "either the claim_id or [another non-claim_id option].  Also, figuring the claim_id is the same as the end_product_reference_id.

leahbannon:
Plan is for Sally  to figure out where it will be placed and how much we can influence the visual design (and interaction related to new tab, etc.) and keep Abby in the loop.

abbyraskinUSDS:
@leikkisa @suzchap we're thinking the label should read "Edit in Caseflow", and the placement would be next to the "Add Contentions" button. It would be reeeally cool if we could add a Caseflow logo to the button, but let us know how much influence we have there.

leikkisa:
@abbyraskinUSDS It's looking like we'll get all of the above! We'll get a screenshot for you when it's ready

suzchap:
@leikkisa Is this done yet and will I be able to validate this based on the screenshot or do I need to see it live?

leikkisa:
@suzchap 
I think for the VBMS side of things, since I can't get that, the screenshot that we sent should be good. For our side of things, you should be able to go to the URL:
/higher_level_reviews/600133239
Now that you made an end_product_reference_id for that (the same as claim_id)

suzchap:
@leikkisa hmmm, that URL didn't work for me so we might have to look at this one together.

leikkisa:
@suzchap I think I forgot to indicate "edit" on the URL.  Assuming this is the correct UAT URL, the link should be this:
https://appeals.cf.uat.ds.va.gov/higher_level_reviews/600133239/edit

suzchap:
This is what I see... It does show Veteran name and ID so is this correct?

[...]

PR #6011, opened 2018-06-26
AMA Review Edit URL

leikkisa:
Connects #5628

### DIFF
--- a/app/controllers/intakes_controller.rb
+++ b/app/controllers/intakes_controller.rb
@@ -62,12 +62,6 @@ class IntakesController < ApplicationController
     }, status: 400
   end
 
-  def edit
-    render json: {
-      claim_id: params[:claim_id]
-    }
-  end
-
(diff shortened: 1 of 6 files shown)


app/models/ama_review.rb
```diff
@@ -50,6 +50,11 @@ def veteran
     @veteran ||= Veteran.find_or_create_by_file_number(veteran_file_number)
   end
 
+  def self.veteran_name_by_claim_id(claim_id:)
```
shanear: i think we can probably just do `find_by_claim_id(claim_id)`, and then get the veteran in the controller.

app/controllers/intakes_controller.rb
```diff
@@ -69,6 +57,18 @@ def error
 
   private
 
+  def set_application
```
shanear: Nice!

Issue #5753, opened 2018-06-01
Make AppealView a polymorphic association


mdbenjam:
Currently the AppealView table has a foreign key into the LegacyAppeal table. We need to make this foreign key a polymorphic association so that it works seamlessly with both BEAAM and Legacy appeals.

PR #6018, opened 2018-06-26
Make appeal view, claims search, and tasks polymorphic

mdbenjam:
Connects #5753

### Description
Now that we have appeals and legacy appeals we need to make DB tables that will have relations to both, polymorphic.

### Testing Plan
1. Have migrated up with data.
2. Have migrated down with data.


app/controllers/reader/claims_folder_searches_controller.rb
```diff
@@ -12,6 +12,6 @@ def create
   private
 
   def appeal
-    LegacyAppeal.find_or_create_by_vacols_id(params[:appeal_id])
+    Appeal.find_appeal_by_id_or_find_or_create_legacy_appeal_by_vacols_id(params[:appeal_id])
```
shanear: lol at this method name. There's got to be a better name

app/models/legacy_appeal.rb
```diff
@@ -7,6 +7,9 @@ class LegacyAppeal < ApplicationRecord
   belongs_to :appeal_series
   has_many :dispatch_tasks, foreign_key: :appeal_id, class_name: "Dispatch::Task"
   has_many :worksheet_issues, foreign_key: :appeal_id
+  has_many :appeal_views, as: :appeal
+  has_many :claims_folder_search, as: :appeal
+  has_many :task, as: :appeal
```
shanear: Are these associations required for the polymorphic association to work properly? I don't think so, but im not sure?

db/migrate/20180626151855_remove_appeal_view_index.rb
```diff
@@ -0,0 +1,12 @@
+class RemoveAppealViewIndex < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+  safety_assured
```
shanear: Is safety really assured here? I imagine we will deploying this under relatively high traffic, and messing around with deleting/adding indexes might cause issues.

We may just want both indexes first and then delete the irrelevant index later.

db/migrate/20180626154654_backfill_appeal_default_type_value.rb
```diff
@@ -0,0 +1,13 @@
+class BackfillAppealDefaultTypeValue < ActiveRecord::Migration[5.1]
```
shanear: Are these migration timestamps such that they'll run in the correct order? It seems like they are but it's hard to tell and just want to make sure.

app/models/appeal.rb
```diff
@@ -1,4 +1,8 @@
 class Appeal < AmaReview
+  has_many :appeal_views, as: :appeal
+  has_many :claims_folder_search, as: :appeal
+  has_many :task, as: :appeal
```
shanear: again, unless we are using these methods, we don't really need to add these. I don't think they're required for the polymorphic association to work properly.

app/controllers/reader/claims_folder_searches_controller.rb
mdbenjam: I know it's super long, but I thought we should be very explicit about what is happening.

db/migrate/20180626151855_remove_appeal_view_index.rb
mdbenjam: Good idea, we can remove this in a follow up PR.

db/migrate/20180626154654_backfill_appeal_default_type_value.rb
mdbenjam: Yup we set the data first, then we say you can't have null columns.

mdbenjam:
I'll split out the index changes into two more PRs.

app/models/appeal.rb
mdbenjam: Doesn't it just make it easier if we define them? Allows us to do nice rails magic stuff without having to think about it.

PR #6034, opened 2018-06-27
Fix form8 spec test

nemodjuric:
Connects #5685


spec/factories/vacols/correspondent.rb
```diff
@@ -6,10 +6,12 @@
       appellant_first_name nil
       appellant_middle_initial nil
       appellant_last_name nil
+      appellant_relationship ""
```
mdbenjam: Do we want to default this to "" rather than nil? I'm thinking if we default to nil we can avoid the check for `if sspare1` below.

spec/models/form8_spec.rb
```diff
@@ -120,7 +122,9 @@
   end
 
   context "#remarks_rolled" do
-    let(:appeal) { Form8.new(remarks: "Hello, World") }
+    let(:appeal) do
```
mdbenjam: Why is this called appeal? Shouldn't  it be called form8?

spec/factories/form8.rb
```diff
@@ -0,0 +1,10 @@
+FactoryBot.define do
+  factory :form8 do
+    factory :default_form8 do
+      appellant_name "Brad Pitt"
+      appellant_relationship "Fancy man"
+      file_number "1234QWERTY"
```
mdbenjam: I think this data is better to live inside the test rather than here. It's opaque in the test that the file number of this form8 is 1234QWERTY, but that's what is needed to link it to the appeal.

spec/factories/vacols/correspondent.rb
nemodjuric: When assigning values from appeal to form8 we do: `appellant_relationship: correspondent_record.sspare1 ? correspondent_record.susrtyp : "",` If we put nil as default, that will be different behavior than what we have now and I don't know if that we'll break any code.

mdbenjam: I'm not sure I understand. This is already defaulting the to nil because it's not specified, right? We shouldn't break any application code by making a change to a factory.

nemodjuric: After the call to `set_vacols_values(appeal)` , if vacols case has sspare1 nil, then in one scenario we end up with appellant_relationship equal to "" (like we have in the current code), 
and
in the second scenario we end up with appellant_relationship equal to nil. 
If, for instance front end reads this variable and shows nil instead of empty string, it might break the code - I haven't checked this, just pointing out potential breakage.

nemodjuric: I meant breaking existing tests that use the factory, not the application code.

spec/factories/vacols/correspondent.rb
```diff
@@ -6,10 +6,12 @@
       appellant_first_name nil
       appellant_middle_initial nil
       appellant_last_name nil
+      appellant_relationship nil
     end
 
     sspare1 { appellant_first_name }
     sspare2 { appellant_middle_initial }
     sspare3 { appellant_last_name }
+    susrtyp { appellant_relationship if sspare1 }
```
mdbenjam: I don't think we need this if statement anymore.

Issue #1055, opened 2017-03-01
Caseflow Help images sizes are too large, and hurts performance


askldjd:
The Caseflow home page is now over **1.5MB**. A significant portion of the download comes from the png files. While png files offer higher quality, it is not suitable these screenshots.

A quick batch conversion from png to jpg results in 60% reduction in size, and has very little loss in quality,

### Batch conversion using Quality = 75
```sh
for i in *.png ; do convert -quality 75 "$i" "${i%.*}.jpg" ; done
```

### Size comparison
```
 56K caseflow-cancelmodal.jpg
 34K caseflow-cancelmodal.png
 23K caseflow-login.jpg
 13K caseflow-login.png
 41K caseflow-mismatched.jpg
 21K caseflow-mismatched.png
 14K caseflow-requiredfield.jpg
6.5K caseflow-requiredfield.png
 14K vacols-certify.jpg
 30K vacols-certify.png
 51K vacols-cleardate.jpg
252K vacols-cleardate.png
 11K vacols-processes.jpg
 28K vacols-processes.png
 21K vacols-search.jpg
 38K vacols-search.png
 66K vacols-travelboard.jpg
 30K vacols-travelboard.png
 17K vacols-update.jpg
 38K vacols-update.png
268K vbms-addlabel.jpg
685K vbms-addlabel.png
 15K vbms-documentproperties.jpg
 25K vbms-documentproperties.png
```

### 1.5 MB to load the Caseflow homepage

![image](https://cloud.githubusercontent.com/assets/3258724/23483225/1135105e-fea0-11e6-9aba-4f201fa72f53.png)

askldjd:
The compressed images.

[compressed.zip](https://github.com/department-of-veterans-affairs/caseflow/files/818790/compressed.zip)

nolawi:
@askldjd is this still an issue

askldjd:
Yes it is. The Certification help page is **2.5MB**. While this is low priority, it is probably 30min of work to fix.

![image](https://user-images.githubusercontent.com/3258724/41253556-c13511b2-6d8e-11e8-9008-9a6066e2ae9f.png)

nolawi:
Looks like the only 2 images have not been compressed, have a PR

PR #6037, opened 2018-06-27
issue1055

nolawi:
Connects #1055


askldjd:
I think all the images can be converted to JPG. The help folder goes from 1.8 MB to 792.4 kB with very little loss in quality.

Here's the converted folder.
[help-jpg.zip](https://github.com/department-of-veterans-affairs/caseflow/files/2142810/help-jpg.zip)

PR #6038, opened 2018-06-27
Move common styling for dl into own component

lowellrex:
Connects #5735.

The rest of the #5735 ticket will redesign the way the issues section of the case details page appears. We will use the same list styling we already use for the two `dl`s in CaseSnapshot so this PR simply moves the styling on the `dl` element out to it's own component for easy reuse.

### Testing Plan
* Go to case details page
* Everything in the case snapshot section should look the same

### DIFF
--- a/client/app/queue/CaseSnapshot.jsx
+++ b/client/app/queue/CaseSnapshot.jsx
@@ -1,10 +1,11 @@
-import { after, css, merge } from 'glamor';
+import { css } from 'glamor';
 import moment from 'moment';
 import PropTypes from 'prop-types';
 import React from 'react';
 import { connect } from 'react-redux';
 import _ from 'lodash';
 
+import CaseDetailsDescriptionList from './components/CaseDetailsDescriptionList';
 import SelectCheckoutFlowDropdown from './components/SelectCheckoutFlowDropdown';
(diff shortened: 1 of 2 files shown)


client/app/queue/CaseTitle.jsx
```diff
@@ -73,11 +73,9 @@ CaseTitle.defaultProps = {
...
 
-class CaseTitleScaffolding extends React.PureComponent {
-  render = () => <div {...containingDivStyling}>
-    <h1 {...headerStyling}>{this.props.heading}</h1>
-    <ul {...listStyling}>
-      {this.props.children.map((child, i) => <li key={i} {...listItemStyling}>{child}</li>)}
-    </ul>
-  </div>;
-}
+const CaseTitleScaffolding = (props) => <div {...containingDivStyling}>
```
dannysellers: It looks like `CaseTitle` could also be functional/stateless if you want:
```javascript
export default function CaseTitle(props) {
  const {
    appeal,
    ...
  } = props;

  return <CaseTitleScaffolding ... />
    ...
  </CaseTitleScaffolding>
}
```

lowellrex: I'll make sure to do that next time we're messing around with the CaseTitle class again.

PR #6057, opened 2018-06-28
Ignores missing first names

evankroske:
### Description
Avoids a null pointer error on the review task table when the assigner of the case doesn't have a first name.


client/app/queue/JudgeReviewTaskTable.jsx
```diff
@@ -25,6 +25,11 @@ class JudgeReviewTaskTable extends React.PureComponent {
   }, {
     header: COPY.JUDGE_QUEUE_TABLE_DOCUMENT_ID_COLUMN_TITLE,
     valueFunction: (task) => {
+      if (!task.attributes.assigned_by_first_name) {
```
amprokop: can we just return nothing instead of a question mark?

amprokop: actually, re-reading this — can we just return the last name if it's present instead?

evankroske: I think that if one name is missing, the other will be, too. I changed the code to return nothing.

Issue #6045, opened 2018-06-28
AMA: Add AOD and remove RO from AMA Case details


mdbenjam:
The AMA appeal table needs an AOD field so that we can surface that information in the case details page.

We've also determined that AMA appeals don't really have RO's since intaking will happen at the board, so for now we'll be removing that info from AMA case details.

PR #6060, opened 2018-06-28
Adding POA and AOD to AMA case details

mdbenjam:
Connects #6045

### Description
Three changes:
1) Added advanced_on_docket to the DB, and return that in the appeal serializer.
2) Removed RO from appeal serializer since AMA appeals don't really have one.
3) Refactored the PowerOfAttorney model to use a new BgsPowerOfAttorney service. We delegate a lot of methods there so that AMA appeals can also use the service. 

screen shot 2018-06-29 at 1 14 17 pm

### Acceptance Criteria 
- [ ] AOD and POA are properly displayed on queue details screen for AMA appeals. RO is not shown.

### Testing Plan
1. Make sure :queue_beaam_appeals is enabled.
1) Run rake db:seed
1) Go to queue/beaam
1) Click into an appeal and ensure that AOD and POA are properly displayed.
1) Ensure that RO is not displayed
1) Ensure that when navigating to Reader and clicking on the claim details dropdown that RO is not displayed.


app/models/power_of_attorney.rb
```diff
@@ -18,41 +18,14 @@ class PowerOfAttorney
...
-
-      # TODO: should this be an exception at all? It's a known case.
-      # Fix ruby-bgs so it doesn't throw here.
-      Raven.capture_exception(e)
-    end
-
-    bgs_representative_address
-  end
+  # By using the prefix command of delegate we make the methods bgs_representative_name etc.
+  delegate :representative_name,
```
amprokop: Cool. If I'm understanding this right, the methods `bgs_representative_name` etc stay the same, we just have a better way of lazy-loading — we don't make the call to BGS until one of those methods are accessed. Is that right?

app/models/power_of_attorney.rb
```diff
@@ -18,41 +18,14 @@ class PowerOfAttorney
...
+           :representative_type,
+           :representative_address,
+           :participant_id, to: :bgs_power_of_attorney, prefix: :bgs
```
amprokop: didn't know about this fun `prefix` method, they thought of everything 😍

app/models/power_of_attorney.rb
mdbenjam: Exactly! In fact now we won't even load the address unless it's specifically needed.

Issue #6027, opened 2018-06-27
Spreadsheet Validation | Allow ROs to have no non-availability dates


sharonwarner:
When filling out the RO non-availability spreadsheet, the hearing coordinators will put "N/A" under each RO that has no non-availability dates.

AC:
- When validating the RO non-availability dates, allow the date field to either be a date or "N/A"
- When importing the RO non-availability dates, skip the entries that have "N/A" as the date

MeredithStewart:
@sharonwarner - Can we also add this validation for Board non-availability dates? There would always likely be some (for fed holidays), but it is possible that there wouldn't be any if the timeframe was shorter.  

Do we need to update the spreadsheets to indicate the n/a options?

sharonwarner:
I'll add a second ticket for the board non-availability dates. I don't know Excel very well, but I think we could add 'N/A' as an option for the row right below the ROs.

PR #6061, opened 2018-06-29
 Allow ROs to have non-availability dates

Sjones352:
Connects #6027

### Acceptance Criteria 
- [ ]  When validating the RO non-availability dates, allow the date field to either be a date or "N/A"

**Throws error**
screen shot 2018-06-29 at 12 19 37 pm


app/services/hearing_schedule/validate_ro_spreadsheet.rb
```diff
@@ -17,6 +17,7 @@ class RoDatesNotUnique < StandardError; end
   class RoDatesNotInRange < StandardError; end
   class RoDatesNotCorrectFormat < StandardError; end
   class RoTemplateNotFollowed < StandardError; end
+  class RoDatesNotApplicable < StandardError; end
```
sharonwarner: I don't think we need to add this new error

Issue #5281, opened 2018-04-24
AMA intake | Designate "same office" as special issue on claim for HLRs


abbyraskinUSDS:
In VBMS Claims Service V5 we will have the ability to pass a special issue designating if the Higher Level Review or Supplemental Claim was requested to be processed at the same station. We should pass this value properly.

## Acceptance Criteria
- When a user intakes a HLR and selects same station, we should designate that the HLR EP should be routed to the same station as where one of the original issues was processed.
- When a user does not select same station, we should not designate that it should be routed to the same station.

## Technical Notes
- We need a way to determine if VBMS correctly received the request.
- This is dependent on #5428 

This can be done via the existing CreateContentions endpoint:

```
<contentionsToCreate ...>
  ...
  <ns2:issue typeCd="SSR" narrative="Same Station Review" inferred="false"/>
</contentionsToCreate>
```

## Validation instructions
Test file numbers: 888451290
Choose Higher Level Review
Select "yes" for same office
Finish establishing an end product

After an EP is created we'll need to figure out how to look at it in VBMS. If we can't validate in the VBMS UI, then I can look at what gets returned by ListContentions and see if it's in there.

abbyraskinUSDS:
@shanear to add details and investigate

shanear:
This is blocked because the create contentions service is down in UAT right now.

leahbannon:
(for testing, look for special issue on contentions created in VBMS)

leikkisa:
888451290 should be a good number for this.  So choose Higher Level Review, select "yes" for same office.  After an EP is created we'll need to figure out how to look at it in VBMS.   If we can't validate in the VBMS UI, then I can look at what gets returned by ListContentions and see if it's in there.

leikkisa:
ListContentions does not show data on this. I feel like we need to ask VBMS how to validate this .

Here are a couple of UAT HLRs that we can use to validate what happened after we passed it over to VBMS:
<HigherLevelReview id: 59, veteran_file_number: "123458308", receipt_date: "2018-07-15", informal_conference: false, same_office: true, established_at: "2018-07-25 20:30:51", end_product_reference_id: "600134040", end_product_status: nil, end_product_status_last_synced_at: nil>

[...]

PR #6065, opened 2018-06-29
Add same office to Higher Level Review contentions special issues

leikkisa:
Connects #5281

### Description
Add same office designation to contention special issues for Higher Level Review using VBMS v5


app/models/higher_level_review.rb
```diff
@@ -8,7 +8,20 @@ class HigherLevelReview < AmaReview
 
   private
 
-  # TODO: Update with real code and modifier data
+  def special_issues
+    return [] unless same_office
```
shanear: a "more ruby-like" way of writing this method might be

```
def special_issues
  same_office ? [{ code: "SSR", narrative: "Same Station Review" }] : []
end
```

Issue #5923, opened 2018-06-14
Schedule Generation | Divide hearing days for the schedule period for ROs


tejans24:
* It’s preferred to have RO hearing days distributed among the selected months (example: if an RO has 2 hearing days; it’s preferred to have each hearing day distributed among different months)
* A priority is given for ROs that have multiple rooms available since they can hold more hearings per day
* It’s preferred to have whole number hearing days even if the hearing days cannot be evenly distributed ~ (if an RO has 25 hearing days, 25/6 = 4.16, each month shall have at least 4 hearing days for this RO and the remainder of days can be distributed among random months selected based on the calendar days available to the RO).
* If the RO hearing days cannot be distributed evenly among the selected months due to increased non-availability days for the month, the remaining hearings are distributed through out the other months based on the available calendars days for the RO)

PR #6066, opened 2018-06-29
RO allocations for hearing schedule

tejans24:
Connects #5923 

### Description
* Filters travel board days for fetched from VACOLS from the available days for each RO.
* Evenly divide the hearing days monthly along the schedule period and create distribution weights
* Distribute the total number of hearing days for each month according to the distribution weights
* Give precedence to days when multiple hearing days can be allocated based on room numbers.
  * It's preferred to have (allocated_days_to_month % num_of_rooms == 0) over even distribution based on percentages.
*  Distribute the Hearing days to random available dates in the month

Errors:
* `NotEnoughAvailableDays` - Thrown when the (total number of available days x number of rooms) for each RO is less than the total num allocated days.
* `RoNonAvailableDaysNotProvided` - Thrown when available days not provided for every RO with hearing allocated days.

### Acceptance Criteria 
- [ ] Code is redable
- [ ] All tests are passing
- [ ] Algorithm generation makes sense
- [ ] Look through rspec test cases to see if edge cases are covered in `HearingSchedule::GenerateHearingDaysSchedule`

### Testing Plan
There is a decent amount of logic in there, it might help if I help walk through the code.


app/services/hearing_schedule/generate_hearing_days_schedule.rb
```diff
@@ -4,18 +4,34 @@
...
 class HearingSchedule::GenerateHearingDaysSchedule
+  include HearingSchedule::RoAllocation
+  include HearingSchedule::RoDistribution
+
+  class RoNonAvailableDaysNotProvided < StandardError; end
+
   attr_reader :available_days
   attr_reader :ros
 
+  MULTIPLE_ROOM_ROS = %w[RO17 RO18].freeze
```
sharonwarner: This probably makes more sense to move to regional_office.rb. We already have some hearing information there.

app/services/hearing_schedule/generate_hearing_days_schedule.rb
```diff
@@ -4,18 +4,34 @@
...
     @available_days = filter_non_availability_days(schedule_period.start_date, schedule_period.end_date)
     @ro_non_available_days = ro_non_available_days
 
     # handle RO information
-    @ros = assign_available_days_to_ros(RegionalOffice::CITIES)
-    @ros = filter_non_available_ro_days
+    assign_and_filter_ro_days(schedule_period)
+  end
+
+  def assign_and_filter_ro_days(schedule_period)
```
sharonwarner: Can't we just access the schedule_period with @schedule_period instead of a parameter?

app/services/hearing_schedule/generate_hearing_days_schedule.rb
```diff
@@ -31,8 +47,110 @@ def filter_non_availability_days(start_date, end_date)
     business_days
   end
 
+  def monthly_distributed_weights(monthly_weights, allocated_days)
+    monthly_weights.map do |month, weight|
+      [month, distribute(weight, allocated_days)]
+    end.to_h
+  end
+
+  def add_allocated_days_and_format(grouped_shuffled_monthly_dates, ro_key)
```
sharonwarner: I think once we have over 2 parameters, we should use named parameters.

sharonwarner: Alternatively, there are so many parameters in this file, could we change more of them to instance variables?

lib/tasks/local/vacols.rake
```diff
@@ -87,7 +87,7 @@ namespace :local do
       read_csv(VACOLS::Staff, date_shift)
       read_csv(VACOLS::Vftypes, date_shift)
       read_csv(VACOLS::Issref, date_shift)
-      read_csv(VACOLS::TravelBoardSchedule, date_shift)
+      read_csv(VACOLS::TravelBoardSchedule)
```
sharonwarner: Why do we not want to date shift this?

app/mappers/travel_board_schedule_mapper.rb
```diff
@@ -0,0 +1,13 @@
+module TravelBoardScheduleMapper
+  class << self
```
sharonwarner: It looks like something is off with this diff. This file already exists on master.

lib/tasks/local/vacols.rake
tejans24: This can be unshifted. I did this to have accurate date for the current time period. I'll undo this.

app/services/hearing_schedule/generate_hearing_days_schedule.rb
tejans24: I made most of the reused stuff into instance variables. These are mostly helper methods.

app/services/hearing_schedule/generate_hearing_days_schedule.rb
```diff
@@ -4,20 +4,33 @@
...
+  class RoNonAvailableDaysNotProvided < StandardError; end
+  class NoDaysAvailableForRO < StandardError; end
+
+  attr_reader :available_days, :ros
+
+  MULTIPLE_ROOM_ROS = %w[RO17 RO18].freeze
+  MULTIPLE_NUM_OF_ROOMS = 2
+  DEFAULT_NUM_OF_ROOMS = 1
 
   def initialize(schedule_period, co_non_availability_days = [], ro_non_available_days = {})
```
tejans24: co_non_availability_days, ro_non_available_days will be fetched through the schedule_period -> non_available days in a future PR

app/services/hearing_schedule/generate_hearing_days_schedule.rb
```diff
@@ -4,20 +4,33 @@
...
-  attr_reader :ros
+  include HearingSchedule::RoAllocation
+  include HearingSchedule::RoDistribution
+
+  class RoNonAvailableDaysNotProvided < StandardError; end
+  class NoDaysAvailableForRO < StandardError; end
+
+  attr_reader :available_days, :ros
+
+  MULTIPLE_ROOM_ROS = %w[RO17 RO18].freeze
```
sharonwarner: Does it make more sense to have this logic in regional_office.rb?

app/services/hearing_schedule/generate_hearing_days_schedule.rb
```diff
@@ -33,16 +46,151 @@ def filter_non_availability_days(start_date, end_date)
...
+  # allocated hearing days for each RO
+  #
+  # @ros[ro_key][:allocated_dates]: is a hash with months as keys and date has with rooms array value
+  # as values.
+  #
+  # rooms array is initally empty.
+  #
+  # Sample @ros[ro_key][:allocated_dates] -> {[1, 2018]=> {Thu, 04 Jan 2018=>[], Tue, 02 Jan 2018=>[]}}
+  #
+  # rubocop:disable Metrics/CyclomaticComplexity:
```
sharonwarner: Can we split this up to avoid disabling rubocop?

app/services/hearing_schedule/generate_hearing_days_schedule.rb
```diff
@@ -33,16 +46,151 @@ def filter_non_availability_days(start_date, end_date)
...
+      end
+
+      monthly_allocations[month] = allocated_days
+    end
+    @ros[ro_key][:allocated_dates] = grouped_shuffled_monthly_dates
+  end
+  # rubocop:enable Metrics/CyclomaticComplexity:
+
+  def get_rooms(num_of_rooms)
+    Array.new(num_of_rooms) { |room_num| { room_num: room_num + 1 } }
```
sharonwarner: Is this creating an array with number of room with room numbers? Maybe a clearer method name would be get_room_numbers?

app/services/hearing_schedule/generate_hearing_days_schedule.rb
```diff
@@ -72,10 +216,19 @@ def assign_available_days_to_ros(ro_cities)
   #     Thu, 05 Apr 2018,
   #     Fri, 06 Apr 2018
   #   ]}
+  # fails with RoNonAvailableDaysNotProvided
```
sharonwarner: This shouldn't fail if non-availability days aren't provided. Some ROs won't have any.

app/services/hearing_schedule/generate_hearing_days_schedule.rb
tejans24: Sure I can move it.

PR #6073, opened 2018-06-29
Queue | Add work product validation in judge check out

### LINKED ISSUE #5998: Add work product validation in judge check out
We'd like to ensure judges send cases to the right location after reviewing and signing. Requests for outside medical opinion (OMO) should go to the OMO office (VACOLS location 20) and signed decisions should go to BVA dispatch.

## AC
- Show `Assign to OMO` in the Actions dropdown when the work product is one of the following:
   - VHA
   - OTV
   - IME
   - OTI
- Show `Ready for dispatch` in the Actions dropdown for any other work product value
   - DEC
   - OTD
   - OTR
   - REM
   - VAC
   - AFI
   - BOT
   - COR
   - DAF
   - DEV
   - DIM
   - DOR
   - DRM
   - DVH
   - INT
   - OTB
   - OTH
   - REA
   - REU
   - RRC
   - SUP
   - VDC
   - VDR
   - VDS
   - VRM

dannysellers:
Resolves #5998 

### Description
Adds a `work_product` attr to `VACOLS::CaseAssignment` / `JudgeLegacyTask` / `LegacyTaskSerializer`, which is `deprod` from the associated DECASS record

### Acceptance Criteria 
- [ ] Within case details view as judge, only "Assign OMO" or "Ready for Dispatch" should appear in the dropdown

### Testing Plan
1. Go to `/queue` as judge/`BVAAABSHIRE`. All default cases for `BVAAABSHIRE` are draft decisions (work_product `"REM"` or `"DEC"`), so to test OMO cases, you'll have to do an attorney OMO checkout to Anjali Abshire
2. Judge checkout flow dropdown should only have one option, corresponding to work product type

Draft Decision case:
screen shot 2018-07-02 at 10 18 52 am

OMO case:
screen shot 2018-07-02 at 10 21 44 am


app/models/legacy_appeal.rb
```diff
@@ -67,6 +67,10 @@ class UnknownLocationError < StandardError; end
     (self.class.repository.remand_return_date(vacols_id) || false) unless active?
   end
 
+  cache_attribute :work_product do
```
dannysellers: Not sure if `cache_attribute` is relevant/useful here, given that the block fetches from the decass table every time. Maybe something like
```ruby
@work_product ||= VACOLS::Decass.where(defolder: vacols_id).pluck(:deprod).first
```
would actually cache it?

Also, in [`QueueRepository.find_decass_record`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/repositories/queue_repository.rb#L142), we locate records using `defolder: vacols_id` and `deadtim: created_in_vacols_date`. I'm not 100% sure of the relation between `LegacyAppeal` and `VACOLS::Decass` / what decass records with different `deadtim`s mean, but I can look into increasing the specificity here if need be.

Issue #5829, opened 2018-06-07
Block substitute appeals from being eligible for RAMP


shanear:
We are being asked by AMO to block appeals where claimant isn't the veteran. May not be too difficult and will prevent me from having to reinstate a bunch of VACOLS appeals later on down the road.

cmgiven:
We look for whether VACOLS has a non-Veteran appellant listed in the CORRES table. `where CORRES.SSPARE1 is null` will select only those appeals where the appellant is the Veteran.

suzchap:
@leikkisa can you help me get a test ID for this one?

leikkisa:
I think this is a file number where the claimant isn't the veteran:
141102030

leikkisa:
@cmgiven Hey Chris, sorry all of the file numbers I post are for UAT, and we are using UAT for validation.  Understood that Prod data is sensitive.  I've been trying to say so when I do so in emails, but I was taking for granted that Suzanne knew that, and I will be more clear going forward to say the file numbers are UAT file numbers.

@shanear I am wondering where the PR is for this ticket?  I want to see how we approached this.  I found a file number that should be eligible, but it has a SSPARE1 that is different from the Veteran's last name, and in the UI it comes up as ineligible. However, the relationship to the Veteran is listed as "Veteran" (as opposed to child or spouse or something).  I don't think having a different last name from the Veteran is a good way to check if the appellant is someone else, because both a child and spouse could have the same last name as the Veteran.

leikkisa:
All data below is fake test data.
I validated this ticket using Veteran 141102030

When I create a LegacyAppeal for them with their vacols_id, and get the ramp_ineligibility_reason, it is :claimant_not_veteran

The SSPARE1 is different from the Veteran's last name (please see note above that I don't think this is the correct way to check for a claimant different than the veteran, but it seems to be what was implemented).  Otherwise, the data indicates that they would be eligible for RAMP.

[...]

Issue #5689, opened 2018-05-29
Move Intake tests to FACOLS


mdbenjam:
Intake tests in the following files are failing. We need to move them to FACOLS.

```
spec/models/ramp_election_intake_spec.rb,
spec/models/ramp_election_rollback_spec.rb,
spec/feature/intake/intake_spec.rb,
spec/feature/intake/ramp_election_spec.rb,
```

Issue #5130, opened 2018-04-11
RAMP election | Allow some ACT appeals to opt into RAMP


abbyraskinUSDS:
Currently, we prevent all activated cases from opting into RAMP. However, the Board and VBA would like to allow certain activated appeals to opt into RAMP - specifically those that meet all of the following criteria:
- have a docket date 2016 or later
- have not had a hearing
- are not in AOD (advance on docket) status
- are not CAVC cases

## Logic
> have a docket date 2016 or later

`BFD19 >= date '2016-01-01'`

> have not had a hearing

If we assume VACOLS is always correct, this should be:

`BFHA is null or BFHA not in ('1', '2', '6')`

But I will check with Jed whether we should implement a more rigorous check.

> are not in AOD (advance on docket) status

We have already built logic for this at `Appeal#aod`.

> are not CAVC cases

`BFAC <> '7'`

abbyraskinUSDS:
@cmgiven could you weigh in on the logic/complexity here? We don't need this urgently, but sometime in the next couple weeks would be great, if feasible.

cmgiven:
> have a docket date 2016 or later

`BFD19 >= date '2016-01-01'`

> have not had a hearing

If we assume VACOLS is always correct, this should be:

`BFHA is null or BFHA not in ('1', '2', '6')`

But I will check with Jed whether we should implement a more rigorous check.

> are not in AOD (advance on docket) status

We have already built logic for this at `Appeal#aod`.

> are not CAVC cases

`BFAC <> '7'`

leikkisa:
@cmgiven 
Hi Chris! We are wanting to set up some test data in UAT for validating this issue for one of our 10 RAMP veterans. Would you be able to help me figure out what I need to do to get that set up?

cmgiven:
Sure thing. You have an hour to pair sometime tomorrow?

leikkisa:
@cmgiven Yes, that would be great! I'm pretty free all day (West coast time)

shanear:
This has been deployed for months. We can assume it's working correctly

PR #6075, opened 2018-06-30
Update RAMP eligibility logic

shanear:
connects #5130
connects #5829
connects #5689

Decided to knock out both RAMP eligibility updates with one commit since they're both in the same ballpark. I made some changes to some core appeals logic so I'd like 2 reviews on this one.


app/services/appeal_repository.rb
```diff
@@ -155,7 +155,7 @@ def self.set_vacols_values(appeal:, case_record:)
       hearing_request_type: VACOLS::Case::HEARING_REQUEST_TYPES[case_record.bfhr],
       video_hearing_requested: case_record.bfdocind == "V",
       hearing_requested: (case_record.bfhr == "1" || case_record.bfhr == "2"),
-      hearing_held: !case_record.bfha.nil?,
+      hearing_held: ["1", "2", "6"].include?(case_record.bfha),
```
shanear: According to @cmgiven this is a more accurate definition of `hearing_held`. Since we were only using this value in certification it was probably okay, but it should also be okay to update it.

app/models/legacy_appeal.rb
```diff
@@ -251,7 +251,13 @@ def hearing_scheduled?
   end
 
   def eligible_for_ramp?
-    (status == "Advance" || status == "Remand") && !in_location?(:remand_returned_to_bva)
+    (
+      (status == "Advance" || status == "Remand") && !in_location?(:remand_returned_to_bva) ||
+      (
+        (docket_date && docket_date.to_date > Date.new(2015, 12, 31)) &&
```
cmgiven: Can we pull this date out and add some context around where it comes from and under what circumstances it might change?

app/services/appeal_repository.rb
cmgiven: 👍

app/models/legacy_appeal.rb
```diff
@@ -39,14 +39,14 @@ class UnknownLocationError < StandardError; end
   vacols_attr_accessor :case_record
   vacols_attr_accessor :outcoding_date
   vacols_attr_accessor :last_location_change_date
-  vacols_attr_accessor :docket_number
+  vacols_attr_accessor :docket_number, :docket_date
```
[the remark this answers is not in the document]
cmgiven: 👍

spec/models/legacy_appeal_spec.rb
```diff
@@ -1007,14 +1007,62 @@
...
       it { is_expected.to be_truthy }
+
+      context "when appellant is not the veteran" do
+        let(:correspondent) do
+          create(
+            :correspondent,
+            appellant_first_name: "David",
+            appellant_middle_initial: "D",
+            appellant_last_name: "Schwimmer"
+          )
```
cmgiven: Ross from the TV show Friends, asking, "Why? Why me?"

app/models/legacy_appeal.rb
shanear: I can, but the problem is i don't really know what its significance is. It seems almost arbitrarily selected 😕

PR #6087, opened 2018-07-02
Caseflow | Sets up Flow

evankroske:
### Description
Sets up Flow and annotates one file.

### Testing Plan
- [x] Introduce a type error to utils.js.
- [x] Confirm that Circleci fails.
- [x] Remove type error.


client/app/queue/reducers.js
```diff
@@ -1,17 +1,80 @@
...
-import caseDetailReducer from './CaseDetail/CaseDetailReducer';
+import * as caseDetailReducer from './CaseDetail/CaseDetailReducer';
 import caseListReducer from './CaseList/CaseListReducer';
-import uiReducer from './uiReducer/uiReducer';
+import * as uiReducer from './uiReducer/uiReducer';
 
 // TODO: Remove this when we move entirely over to the appeals search.
 import caseSelectReducer from '../reader/CaseSelect/CaseSelectReducer';
 
+export type DeprecatedTask = {
```
dannysellers: I feel like these model types (i.e. not `State`/`QueueState`) should go in a dedicated file

client/app/queue/reducers.js
```diff
@@ -1,17 +1,80 @@
...
+    appeals: LoadedQueueAppeals,
+    tasks: LoadedQueueTasks,
+    loadedUserId: string
+  },
+  editingIssue: Object,
+  docCountForAppeal: {[string]: Object},
+  stagedChanges: {
+    appeals: {[string]: Object},
+    taskDecision: {
+      type: '',
```
dannysellers: Should this be `string`? `taskDecision.opts` is a bit overloaded at the moment, so I can create a type for that after merge.

client/app/queue/utils.js
```diff
@@ -38,18 +40,17 @@ export const associateTasksWithAppeals = (serverData = {}) => {
 *  Sort by docket date (form 9 date) oldest to
 *  newest within each group
 */
-export const sortTasks = ({ tasks = {}, appeals = {} }) => {
+export const sortTasks = ({ tasks = {}, appeals = {} }: {tasks: Tasks, appeals: {[string]: Object}}) => {
   const partitionedTasks = _.partition(tasks, (task) =>
     appeals[task.vacolsId].attributes.aod || appeals[task.vacolsId].attributes.type === 'Court Remand'
   );
 
-  _.each(partitionedTasks, _.sortBy('attributes.docket_date'));
```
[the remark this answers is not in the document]
evankroske: This line of code doesn't sort the tasks. Instead, it sorts the string "attributes.docket_date" twice. I think that we should fix this in another PR.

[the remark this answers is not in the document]
dannysellers: Yeah, right, I think that was mentioned in the Flow WW demo. Could you add a todo or create a GH issue?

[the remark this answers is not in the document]
evankroske: Sure! I created a GH issue:

https://github.com/department-of-veterans-affairs/caseflow/issues/6101

client/app/queue/reducers.js
evankroske: Agreed. I just moved all of the types into queue/types.js

client/app/queue/reducers.js
evankroske: Yes, thanks for catching this! I replaced both of these values with types.

client/app/queue/types.js
```diff
@@ -87,3 +87,11 @@ export type QueueState = {
   tasksAndAppealsOfAttorney: TasksAndAppealsOfAttorney,
   isVacolsIdAssignedToUserSelected: {[string]: {[string]: {[string]: boolean}}}
 };
+
+export type State = {
```
dannysellers: @evankroske it looks like `State` isn't used anywhere—should it annotate the main reducer somehow? Or is it more for reference?

evankroske: It's going to be used in mapStateToProps, but there aren't any React components in this PR, so it's not used yet.

PR #6088, opened 2018-07-02
VACOLS::Case: Adds error log statement

evankroske:
### Description
Prints a scolding error message if the programmer fails to validate location before passing it to `Case.update_vacols_location!`.

### Testing Plan
- [x] Code review


app/models/vacols/case.rb
```diff
@@ -209,7 +209,12 @@ def remand_clone_attributes
 
   # rubocop:disable Metrics/MethodLength
   def update_vacols_location!(location)
-    return unless location
+    unless location
+      Rails.logger.error "THERE IS A BUG IN YOUR CODE! It attempted to assign a case to a falsy location. " \
```
amprokop: I looked into where the `return unless location` line was added and suspect that there is no legitimate, expected situation where we would run into that condition. See  https://github.com/department-of-veterans-affairs/caseflow/commit/79d49395fb75d1b56d8fb942ca405bce2bd0c7f5

Have you thought about explictly throwing a Sentry error then returning? If we do that then wait a week or two and see no results, I'd be in favor of just throwing an error if there's a falsy location value.

app/repositories/queue_repository.rb
```diff
@@ -156,6 +156,7 @@ def update_decass_record(decass_record, decass_attrs)
 
     def update_location_to_attorney(vacols_id, attorney)
       vacols_case = VACOLS::Case.find(vacols_id)
+      fail "Attorney #{attorney.id} has a falsy vacols_uniq_id" unless attorney.vacols_uniq_id
```
amprokop: We typically try to fail with some class name e.g. `Caseflow::Error::MissingVacolsUserID` rather than a string

app/models/vacols/case.rb
evankroske: I'm in favor of throwing an exception here if nothing relies on this behavior. However, I don't think that we need to use Sentry to determine whether that's the case. I could search for this log message in a week and add the exception then. What do you think?

app/repositories/queue_repository.rb
evankroske: I changed the code to use the existing exception VACOLS::Case::InvalidLocationError

app/models/vacols/case.rb
amprokop: Sure, that sounds fine!

app/repositories/queue_repository.rb
```diff
@@ -156,6 +156,7 @@ def update_decass_record(decass_record, decass_attrs)
 
     def update_location_to_attorney(vacols_id, attorney)
       vacols_case = VACOLS::Case.find(vacols_id)
+      fail VACOLS::Case::InvalidLocationError, "Invalid location \"#{attorney.vacols_uniq_id}\"" unless attorney.vacols_uniq_id
```
amprokop: hm this will only fail when attorney.vacols_uniq_id is nil or false. is that what you want?

evankroske: Yep! `update_vacols_location!` used to do nothing in this case, so this is the case in which I want to throw an error. Not sure whether blank locations are valid, so I don't want to ban those

evankroske:
Looks like Caseflow relies on this behavior. I found several instances of "THERE IS A BUG IN YOUR CODE!" in the logs. Here's an example request: 4601c469-9d11-4b26-88a0-4fa763dbb2fb.

amprokop:
Interesting and maybe a little worrisome. Did you happen to get a sense of the context? If you don't have time to atm, mind creating a new issue so someone can look into it eventually?

evankroske:
@amprokop, I filed a bug with more context: https://github.com/department-of-veterans-affairs/caseflow/issues/6477

PR #6091, opened 2018-07-02
Remove establishes end product

shanear:
Push the last of `EstablishesEndProduct` logic into `EndProductEstablishment`. Moves any of the methods that require database actions into `RampReview` or `AmaReview`.

connects #5531


app/models/concerns/establishes_end_product.rb
```diff
@@ -1,84 +0,0 @@
-module EstablishesEndProduct
```
leikkisa: Yay, so happy to only have one of these now.

app/models/ramp_review.rb
```diff
@@ -56,8 +58,62 @@ def end_product_base_modifier
     end_product_establishment.valid_modifiers.first
   end
 
+  def end_product_active?
```
leikkisa: Do we (eventually) need the end_product_status related code for AMA review too?  Maybe more refactoring potential there for the future.

shanear: Yeah i think so. I think this will be better when we save end_product_establishment to the DB

Issue #6063, opened 2018-06-29
Spreadsheet Validation | Allow N/A for co non-availability days


sharonwarner:
AC:
- When validating co non-availability days, we'll allow either a date type, or N/A
- When importing co non-availability days, we'll skip N/A

PR #6103, opened 2018-07-04
Spreadsheet Validation | co-non-availability-days

Sjones352:
Connects #6063

### Description
This PR validates  co non-availability days
### Acceptance Criteria 
- [ ]  When validating co non-availability days, we'll allow either a date type, or N/A
- [ ] When importing co non-availability days, we'll skip N/A


app/services/hearing_schedule/validate_ro_spreadsheet.rb
```diff
@@ -88,16 +88,22 @@ def validate_co_non_availability_template
     end
   end
 
-  def validate_co_non_availability_dates
-    unless @co_spreadsheet_data.all? { |date| date.instance_of?(Date) }
```
sharonwarner: I think it makes more sense to return true if the date is "n/a". Then we don't have to recheck it in other places as well.

app/models/hearings/co_non_availability.rb
```diff
@@ -5,6 +5,7 @@ def import_co_non_availability(schedule_period)
       co_non_availability = []
       transaction do
         dates.each do |date|
+          next unless date != "N/A"
```
sharonwarner: Same comment as the other pr, I think it makes more sense to do `next if date == "N/A"`

app/services/hearing_schedule/validate_ro_spreadsheet.rb
Sjones352: Fixed

PR #6105, opened 2018-07-05
Migrate case search to FACOLS

lowellrex:
Connects #5688.


spec/factories/vacols/correspondent.rb
```diff
@@ -2,6 +2,9 @@
   factory :correspondent, class: VACOLS::Correspondent do
     sequence(:stafkey)
 
+    snamef "Joshua"
+    snamel "Chamberlain"
```
lowellrex: Gives legacy appeals a Veteran name. Expect all legacy appeals to have Veteran's with names.

spec/factories/vacols/folder.rb
```diff
@@ -1,5 +1,6 @@
 FactoryBot.define do
   factory :folder, class: VACOLS::Folder do
     sequence(:ticknum)
+    sequence(:tinum)
```
lowellrex: tinum maps to appeal docket_number. Expect all cases to have docket numbers.

spec/feature/queue/search_spec.rb
```diff
@@ -91,20 +43,20 @@
 
       it "clicking on the x in the search bar returns browser to queue list page" do
         click_on "button-clear-search"
-        expect(page).to have_content(COPY::ATTORNEY_QUEUE_TABLE_TITLE)
+        expect(page).to_not have_content("1 case found for")
```
lowellrex: This change removes the assumption that the current user will have cases assigned to them.

spec/factories/vacols/case.rb
```diff
@@ -217,6 +217,13 @@
       end
     end
 
+    trait :paper_case do
+      after(:build) do |vacols_case, _evaluator|
+        vacols_case.folder.tivbms = "N" if %w[Y 1 0].include?(vacols_case.folder.tivbms)
+        vacols_case.folder.tisubj2 = "N" if vacols_case.folder.tisubj2 && vacols_case.folder.tisubj2.eq?("Y")
+      end
+    end
```
lowellrex: From [this code](https://github.com/department-of-veterans-affairs/caseflow/blob/8cdfa45165ef9adaa87f47008f4f52d4e2d4665a/app/services/appeal_repository.rb#L217) in `AppealRepository.folder_type_from()`.

mdbenjam: I don't understand the if statements here. Can't we just set both fields to "N" without checking?

spec/factories/vacols/correspondent.rb
mdbenjam: I'm on the fence about this addition. When we add defaults like this, people can make assertions in tests on these defaults which I think obscures what is happening in the test. Maybe it makes sense in this case since name is so fundamental, but I'm wondering if it makes sense to either make it easier to set like we do with appellant name below, or if we could push this into a trait or subclass factory.

spec/factories/vacols/case.rb
lowellrex: My thought was that there might be other values that we would want to set these fields to (`tivbms` and `tisubj2`), and we would only want to reset those fields if they prevented the case from being described as a `paper_case`.

For instance, in a different test we might want to set `tivbms` to 3 (perhaps to validate that we throw an error if the value of this field is anything other than the psuedo-boolean values it accepts) or `tisubj2` to null but we also want the case to be recognized as a paper case. The way this trait is currently written allows for that those scenarios without overwriting the values of these fields.

lowellrex: Also, the VACOLS database documentation we have suggests that both of these fields should be used as booleans (which is a type Oracle databases [do not implement](https://stackoverflow.com/questions/3726758/is-there-a-boolean-type-in-oracle-databases#3726846), crazy), so maybe we are misinterpreting the value of `tivbms` when it is 0?

spec/factories/vacols/correspondent.rb
lowellrex: The main reason why I favor the approach of setting a default here is that it makes setting up the most common case for an appeal (some valid appeal that references a case that has a Veteran with a name) simpler than without these defaults:

```
# With default Veteran name:
appeal = FactoryBot.create(:legacy_appeal, vacols_case: FactoryBot.create(:case))

# Without default Veteran name:
appeal = FactoryBot.create(
  :legacy_appeal,
  vacols_case: FactoryBot.create(
    :case,
    correspondent: FactoryBot.create(
      :correspondent,
      snamef: "Joshua",
      snamel: "Chamberlain"
  )
)
```

To protect against people testing against this static value, I'm game for using `generate_first_name()` and `generate_last_name()`. What do you think?

spec/factories/vacols/case.rb
mdbenjam: It's totally valuable to have tests where these are set to things other than "N", but I think if you're setting them manually, you aren't going to also set the paper_case trait. If your trait is a no_op then it shouldn't be set.

spec/factories/vacols/correspondent.rb
mdbenjam: But if you're not testing against the value, why do you need a first name/last name? Does the code throw an error if they're not defined?

lowellrex: At the moment we need the Veteran to have a name because our tests fail with accessibility errors because we have links without contents (I think only in the breadcrumbs component) if the Veteran does not have a name. Since the breadcrumbs component is going away soon (end of next week!), I think that particular case will be resolved, but in the meantime it needs to be addressed. What do you think of a factory for something like `case_with_veteran` to handle this case where we expect an appeal to have a case with a named Veteran?

image

spec/factories/vacols/case.rb
lowellrex: I'm not sure what alternative you're advocating for here. Would you like this trait to look like the following?
```
trait :paper_case do
  after(:build) do |vacols_case, _evaluator|
    vacols_case.folder.tivbms = "N" if %w[Y 1 0].include?(vacols_case.folder.tivbms)
    vacols_case.folder.tisubj2 = "N" if vacols_case.folder.tisubj2 && vacols_case.folder.tisubj2.eq?("Y")
  end
end
```

spec/factories/vacols/correspondent.rb
mdbenjam: We already have a `with_veteran` trait, seems like this would go perfectly in it!

spec/factories/vacols/case.rb
mdbenjam: I'd probably put the trait on folder:

```
trait :paper_case do
  tivbms "N"
  tisubj2 = "N"
end
```

spec/factories/legacy_appeal.rb
```diff
@@ -2,14 +2,22 @@
...
 
     vacols_id { vacols_case.bfkey }
     vbms_id { vacols_case.bfcorlid }
 
     trait :with_veteran do
-      after(:create) do |legacy_appeal, _evaluator|
+      after(:create) do |legacy_appeal, evaluator|
         create(:veteran, file_number: legacy_appeal.veteran_file_number)
+
+        if evaluator.vacols_case
```
mdbenjam: I'd probably try to unify the bgs and VACOLS names. Like this: 

```
veteran = create(:veteran, file_number: legacy_appeal.veteran_file_number)
evaluator.vacols_case.correspondent.snamef = veteran.first_name
evaluator.vacols_case.correspondent.snamel = veteran.last_name
```

PR #6106, opened 2018-07-05
Move personalized task queue test to FACOLS

lowellrex:
Connects #5688.

### DIFF
--- a/spec/factories/vacols/case.rb
+++ b/spec/factories/vacols/case.rb
@@ -217,13 +217,6 @@ FactoryBot.define do
       end
     end
 
-    trait :paper_case do
-      after(:build) do |vacols_case, _evaluator|
-        vacols_case.folder.tivbms = "N" if %w[Y 1 0].include?(vacols_case.folder.tivbms)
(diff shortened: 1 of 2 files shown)


spec/factories/vacols/case.rb
```diff
@@ -217,6 +217,13 @@
       end
     end
 
+    trait :paper_case do
+      after(:build) do |vacols_case, _evaluator|
+        vacols_case.folder.tivbms = "N" if %w[Y 1 0].include?(vacols_case.folder.tivbms)
+        vacols_case.folder.tisubj2 = "N" if vacols_case.folder.tisubj2 && vacols_case.folder.tisubj2.eq?("Y")
+      end
+    end
```
lowellrex: Racing against the similar case search migration to FACOLS PR to incorporate this change. https://github.com/department-of-veterans-affairs/caseflow/pull/6105/files#diff-7c3dd7d1da702711c44c25340912ad91R220

mdbenjam: Same comment from previous PR with this change.

PR #6113, opened 2018-07-05
Ignore vulnerability for a week

lowellrex:
CI builds are currently failing due to a vulnerability in a third-party package we use. This PR ignores this vulnerability for a week. After a week `bundle-audit` will fail again and we will be reminded to look to see if a patch for the gem has been released.

image

Link to issue on the Thor gem's github: https://github.com/erikhuda/thor/issues/514

I don't think it makes sense to try to get rid of our dependency on this gem since so many gems we rely on use it:
* bourbon
* bundler-audit
* foreman
* jquery-rails
* neat
* railties
* shoryuken

### DIFF
--- a/lib/tasks/security.rake
+++ b/lib/tasks/security.rake
@@ -11,8 +11,12 @@ task :security_caseflow do
   puts "running bundle-audit to check for insecure dependencies..."
   exit!(1) unless ShellCommand.run("bundle-audit update")
 
-  # TODO(lowell): Remove this ignore after we have upgraded rubocop.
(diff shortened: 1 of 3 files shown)


lib/tasks/security.rake
```diff
@@ -11,8 +11,12 @@ task :security_caseflow do
   puts "running bundle-audit to check for insecure dependencies..."
   exit!(1) unless ShellCommand.run("bundle-audit update")
```
drewish: you could fold this into the next command by using the `bundle-audit check --update` flag

Issue #5443, opened 2018-05-08
Non-rated issues | Finish: Incorporate non-rated issues into count


abbyraskinUSDS:
Adds some lightweight visual feedback near the "Establish claim" button to tell users how many issues have been selected on the page. 

## Acceptance Criteria
On the "review" page for SCs and HLRs:
- Add any non-rated or uncategorized issues to the total issues count
- Remove the word "rated"

## Mock
![image](https://user-images.githubusercontent.com/4042181/41855266-2f9da340-7860-11e8-9711-0b9d11651e87.png)

abbyraskinUSDS:
Don't break it out by rated/non-rated - just one total count of issues - scope this to issues count, not the button label

suzchap:
LGTM

PR #6114, opened 2018-07-05
Include non-rated issues in issue count

leikkisa:
Connects #5443 
Include both rated and non-rated issues in issue count on finish page


client/app/intake/actions/ama.js
```diff
@@ -99,3 +99,65 @@ export const completeIntake = (intakeId, intakeData) => (dispatch) => {
       }
     );
 };
+
```
leikkisa: Moved the AMA-only actions into the actions/ama.js file

client/app/intake/components/RatedIssues.jsx
```diff
@@ -18,7 +17,7 @@ export class RatedIssuesUnconnected extends React.PureComponent {
             label={issue.decision_text}
             name={issue.reference_id}
             key={issue.reference_id}
-            value={issue.isSelected}
+            value={issue.isSelected || false}
```
leikkisa: Fixes react bug (the component went from uncontrolled to controlled, now it starts off controlled)

client/app/intake/components/NonRatedIssues.jsx
```diff
@@ -55,7 +55,7 @@ export default class NonRatedIssuesUnconnected extends React.PureComponent {
 
 class NonRatedIssue extends React.PureComponent {
   handleCategoryChange(event) {
-    this.props.setCategory(this.props.id, event.value);
+    this.props.setCategory(this.props.id, event ? event.value : null);
```
leikkisa: Covers the situation where a user deletes the category after adding it, previously it threw an error and kept the old category (even if the box was blank)

client/app/intake/util/index.js
```diff
@@ -83,6 +83,26 @@ export const formatRatingData = (intakeState) => {
...
 };
 
+export const nonRatedIssueCounter = (state, action) => {
+  const selectedIssues = formatIssues(state).request_issues;
+  const selectedIssueCount = selectedIssues ? selectedIssues.length : 0;
+  const currentIssue = state.nonRatedIssues[action.payload.issueId];
+  const descriptionCounter = !currentIssue.description && currentIssue.category ? 1 : 0;
+  const categoryCounter = !currentIssue.category && currentIssue.description ? 1 : 0;
+
+  if (selectedIssueCount && !action.payload.category && !action.payload.description) {
```
leikkisa: If they delete category or description after they previously added it, this decrements the issue count

client/app/intake/util/index.js
```diff
@@ -83,6 +83,26 @@ export const formatRatingData = (intakeState) => {
   return data;
 };
 
+export const nonRatedIssueCounter = (state, action) => {
+  const selectedIssues = formatIssues(state).request_issues;
+  const selectedIssueCount = selectedIssues ? selectedIssues.length : 0;
+  const currentIssue = state.nonRatedIssues[action.payload.issueId];
+  const descriptionCounter = !currentIssue.description && currentIssue.category ? 1 : 0;
```
leikkisa: These are here so that it only increments the issue count if the current issue isn't already accounted for. For example, if there is a category already, and no description, when they add a description it increases the count. However, if there isn't a category it doesn't increase the count. Also, if they add another letter to the description, it doesn't increase the count

Issue #5819, opened 2018-06-07
Flow 1 | Build Schedule | Upload 


Chingujo:
## User Story 
As Justin and Scott upload files required to build the schedule, they will need to select the type of file they are uploading, insert date range, confirm name of file they uploaded, and have the ability to upload a different file, if they have selected the wrong file. 

## AC
- [x] Verify that the page title is Upload Files
- [x] Verify that the instructional text is the below:
  - Please select the file you are uploading and choose a date range. 
- [x] Verify that there is an hz line after the horizontal line
- [x] Verify that there question is as follows: 
  - What are you uploading? 
- [x] Verify that the there are two options:
  - RO and CO hearings
  - Judge non-availability
- [x] Verify that selecting one does the following:
    - "Please input date range" 
    - Opens two date input boxes in the following : date input to date input
    - Date range default: mm/dd/yyyy
       - should auto format based on numerical string input
- [x] Verify that there is a "Select a file for upload" link next to the last date input box
   - Selecting opens a desktop finder
   - Choosing a file renders the following: 
     - Title of spreadsheet in bold.xls
     - Date range in bottom
- [x] Verify that there is an "Upload another file" link that 
- [x] Cancel link sends you to the landing page
- [x] Continue link moves you forward

### Default
![upload schedule](https://user-images.githubusercontent.com/23316107/41176880-0d5b8230-6b30-11e8-8030-3dae4b389278.png)
### After a choice is selected
![upload schedule version 2](https://user-images.githubusercontent.com/23316107/41176881-0d6c0556-6b30-11e8-9493-aa6b18de3895.png)
### After it is selected
![upload schedule final](https://user-images.githubusercontent.com/23316107/41598706-fee1ef60-739e-11e8-8cd8-0e9f1c0f959b.png)

PR #6117, opened 2018-07-05
Date selector

sharonwarner:
Connects #5819 

### Description
This PR adds a date selector to the upload page.

### Testing Plan
1. Go to 'schedule/build/upload'
1. Select either 'RO/CO hearings' or 'Judge non-availability' 
1. Confirm that a date selector is displayed
1. Confirm that when editing the dates, the redux store is updated


client/app/hearingSchedule/components/UploadDateSelector.jsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import InlineForm from '../../components/InlineForm';
+import DateSelector from '../../components/DateSelector';
+
+export default class UploadDateSelector extends React.Component {
+  render() {
+    return <div>
+      <p><i>Please input a date range</i></p>
+      <InlineForm>
```
tejans24: Do we want to add additional validation messages here?

client/app/hearingSchedule/containers/BuildScheduleUploadContainer.jsx
```diff
@@ -1,24 +1,46 @@
...
+  onJudgeEndDateChange
+} from '../actions';
 import BuildScheduleUpload from '../components/BuildScheduleUpload';
 
 export class BuildScheduleUploadContainer extends React.Component {
   render() {
     return <BuildScheduleUpload
       fileType={this.props.fileType}
       onFileTypeChange={this.props.onFileTypeChange}
+      roCoStartDate={this.props.roCoStartDate}
```
tejans24: Can we just add `mapStateToProps` `mapDispatchToProps` to `BuildScheduleUpload`?
Since they are not being used here.

client/app/hearingSchedule/components/UploadDateSelector.jsx
sharonwarner: I don't think we need to do that for the MVP. Maybe once we get the entire flow finished, we could add some.

client/app/hearingSchedule/containers/BuildScheduleUploadContainer.jsx
sharonwarner: I think it makes sense to have the presentational component separate from the state component. I was going off of this blog post: https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0

tejans24: Yea, I'm fine with that approach. We have had discussions on our team on this topic and it was preferred to add them directly to the source rather than trying to separate the presentation layer.

Let's go with this if you think it makes sense.

sharonwarner: Oh interesting, I don't remember that. Let's go with this for now, but let's set up a discussion for some time next week. I'd be interested in hearing the argument for combining them.

app/assets/stylesheets/_commons.scss
```diff
@@ -994,6 +994,7 @@ fieldset {
   margin-bottom: rem(20px);
 
   label {
+    max-width: 100%;
     // This important can be removed when the latest USWDS
```
nolawi: Looks fine with this, however a safe option would be `max-inline-size: max-content;` - which would get same result

PR #6119, opened 2018-07-05
Implement endpoint to return tasks for colocated admin

### LINKED ISSUE #5929: [BE] Implement endpoint for co-located to view their assigned tasks
Implement endpoint for co-located to view their assigned tasks  
* GET /tasks?user_id=xxxx&role=co_located
* co-located tasks are stored in Caseflow `tasks` table

aroltsch:
Fixes #5929 

### Description
Return a list of tasks for a colocated user

```
GET /tasks?user_id=xxxx&role=co_located
```


app/controllers/application_controller.rb
```diff
@@ -96,6 +96,7 @@ def certification_header(title)
   def can_access_queue?
     return true if current_user.attorney_in_vacols?
     return true if current_user.judge_in_vacols? && feature_enabled?(:judge_queue)
+    return true if current_user.colocated_in_vacols? && feature_enabled?(:co_located_queue)
```
dannysellers: Should we add this to [`enable_features_dev`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/scripts/enable_features_dev.rb)?

aroltsch: Good catch!

PR #6122, opened 2018-07-05
Judge queue | Refactors AssignWidget

evankroske:
### Description
Adds types to AssignWidget and refactors the state it uses.

### Testing Plan
- [x] Become BVARERDMAN.
- [x] View tasks assigned to Sunny.
- [x] Assign a task to Jazlyn.
- [x] Refresh the page to confirm that the task has been reassigned.


client/app/components/AssignWidget.jsx
```diff
@@ -8,78 +8,96 @@ import {
...
 } from '../queue/QueueActions';
 import SearchableDropdown from '../components/SearchableDropdown';
 import Button from '../components/Button';
 import _ from 'lodash';
+import type {
+  AttorneysOfJudge, IsTaskAssignedToUserSelected, Tasks, UiStateError, State
+} from '../queue/types';
 
-class AssignWidget extends React.PureComponent {
+class AssignWidget extends React.PureComponent<{|
```
dannysellers: Could this live as a separate type declaration (within this file)? Something like
```javascript
type Props = {
  previousAssigneeId: string...
}

class AssignWidget extends React.PureComponent<Props> {
  ...
}
```

client/app/components/AssignWidget.jsx
```diff
@@ -8,78 +8,96 @@ import {
...
         <p>Assign to:&nbsp;</p>
         <SearchableDropdown
           name="Assignee"
           hideLabel
           searchable
           options={options}
           placeholder="Select a user"
-          onChange={(option) => this.props.setSelectedAssigneeOfUser({ userId: previousAssigneeId,
-            assigneeId: option.value })}
+          onChange={(option) => this.props.setSelectedAssignee({ assigneeId: option.value })}
```
dannysellers: This call isn't working, I'm getting
`VM19395:1 Uncaught Error: Actions may not have an undefined "type" property. Have you misspelled a constant?`

It looks like the action and the constant aren't in the same folder: [`setSelectedAssignee`](https://github.com/department-of-veterans-affairs/caseflow/blob/0867c9ce644e6e9e3e78f59319f90387ae425aab/client/app/queue/QueueActions.js#L195) is in `QueueActions`, while [`SET_SELECTED_ASSIGNEE`](https://github.com/department-of-veterans-affairs/caseflow/blob/0867c9ce644e6e9e3e78f59319f90387ae425aab/client/app/queue/uiReducer/uiConstants.js#L26) is in `uiActions`. My IDE doesn't catch that the constant's in a different `ACTIONS` than is imported, maybe cause they all share the same name?

On a related note, what's the plan for testing judge assignment?

client/app/components/AssignWidget.jsx
```diff
@@ -1,5 +1,5 @@
+// @flow
```
dannysellers: I just noticed this component's in `app/components`... Is there a plan to use it outside of queue?

client/app/components/AssignWidget.jsx
evankroske: Whoops, I just fixed this. Thanks for pointing it out!

I'll expand the testing plan.

client/app/components/AssignWidget.jsx
evankroske: Ah, I see. I've moved AssignWidget to queue/components

PR #6123, opened 2018-07-05
[WIP] Convert intake tests to use FACOLS

leikkisa:
Connects #5689


leikkisa:
Did this work before had understanding of ticket, started a new one :-)

Issue #6092, opened 2018-07-02
Intake | AMA review values should not be wiped on page reload


shanear:
1) start a higher level review intake
2) fill in values for "Did the Veteran request an informal conference?" and the rest of the questions on the review page.
3) Click "Continue to next step"
4) reload the page.
5) Click back. The receipt date is properly filled out, but all the other fields are blank again. They should stay intact.

suzchap:
LGTM

PR #6126, opened 2018-07-05
Review option data retained for AMA

leikkisa:
Connects #6092

Data selected on the review page if a user moves forward in the process and goes back or reloads the page.
Thought I'd go ahead and do the same for claimant on all AMA flows.


client/app/intake/components/SelectClaimant.jsx
```diff
@@ -12,7 +12,7 @@ export default class SelectClaimant extends React.PureComponent {
       relationships
     } = this.props;
 
-    let showClaimants = claimantNotVeteran === 'true';
+    let showClaimants = ['true', true].includes(claimantNotVeteran);
```
shanear: hmmm `['true', true]` is pretty awkward. I'm thinking we probably want to convert this to a boolean in the reducer when we first process the `ui_hash`. We should avoid using `'true'` in logic around the app.

client/app/intake/components/SelectClaimant.jsx
```diff
@@ -40,7 +40,7 @@ export default class SelectClaimant extends React.PureComponent {
         vertical
         options={BOOLEAN_RADIO_OPTIONS}
         onChange={setClaimantNotVeteran}
-        value={claimantNotVeteran}
+        value={claimantNotVeteran === null ? null : claimantNotVeteran.toString()}
```
shanear: I think this can be written as `claimantNotVeteran && claimantNotVeteran.toString()`?

Same with all the other times in the PR that you use this pattern

client/app/intake/reducers/appeal.js
```diff
@@ -23,6 +23,12 @@ const updateFromServerIntake = (state, serverIntake) => {
     receiptDate: {
       $set: serverIntake.receipt_date && formatDateStr(serverIntake.receipt_date)
     },
+    claimantNotVeteran: {
+      $set: serverIntake.claimant_not_veteran
```
shanear: we could do the conversion of `'true'`/`'false'` to `true`/`false` here

client/app/intake/components/SelectClaimant.jsx
leikkisa: Except that it is false sometimes :-/

client/app/intake/reducers/appeal.js
leikkisa: It comes in from the server as true / false.

client/app/intake/components/SelectClaimant.jsx
leikkisa: So coming in from the server, it's set to true, but if a user changes it on the front-end itself, then it is equal to 'true' because the Radio button (annoyingly) requires the value to be a string.

leikkisa: I could format things to be strings on the serverIntake, but then to do additional logic I have to explicitly check if it's equal to 'true' or 'false'.

Issue #4793, opened 2018-03-08
Intake | Allow partial closure of Appeals when there is a mix of rating and non-rating issues


shanear:
RAMP only applies to rating issues. Currently we aren't allowing any appeals that have just one non-rating issue to be eligible for RAMP. We should instead just close the rating issues for appeals that have a mix of rating and non-rating issues.

## Acceptance Criteria
- If an appeal has a mix of rating and non-rating issues, it should be considered eligible for RAMP.
- For these appeals, instead of closing them, only close the rating issues on the appeal and keep the appeal and the non-rating issues open.
- Appeals with 100% rating issues should close as they always have.
- Appeals with 100% non-rating issues should still be considered ineligible.
- On the confirmation page, only display the rating issues that will be closed out.

## Notes
Related to https://github.com/department-of-veterans-affairs/caseflow/issues/4659

abbyraskinUSDS:
@shanear can you investigate whether this has happened?

shanear:
There are 71 instances of this occurring in prod to date:

```
irb(main):021:0> RampElectionIntake.where(error_code: "no_active_fully_compensation_appeals").count
[2018-06-19 10:57:50 -0400]    (27.1ms)  SELECT COUNT(*) FROM "intakes" WHERE "intakes"."type" IN ('RampElectionIntake') AND "intakes"."error_code" = $1  [["error_code", "no_active_fully_compensation_appeals"]]
=> 71
```

I'm thinking we may need to handle this 🤔

cmgiven:
I take this to mean that we want to close out compensation issues, but not others. In this case, we leave the appeal alone, and only create a closed disposition on compensation issues (those `where ISSPROG = '02'`).

I'd suggest creating a 'close' method on 'Issue', and then calling it for each of the `LegacyAppeal.compensation_issues`.

As we do in `AppealRepository.close_undecided_appeal!`, we then need to update the ISSDC and the ISSDCLS.

```
        issdc: disposition_code,
        issdcls: VacolsHelper.local_time_with_utc_timezone
```

shanear:
I've validated that this is happening in prod.

PR #6127, opened 2018-07-05
Allow closing only the compensation issues for a RAMP VACOLS appeal

shanear:
Connects #4793


app/models/ramp_closed_appeal.rb
```diff
@@ -28,31 +55,42 @@ def reclose!
...
-      RampClosedAppeal.find_by(vacols_id: appeal.vacols_id)
+    def partial
+      where.not(partial_closure_issue_sequence_ids: nil)
     end
 
-    appeals_to_reclose.each(&:reclose!)
+    def reclose_all!
+      appeals_to_reclose = []
+
+      fully_closed.find_in_batches(batch_size: 800) do |batch|
```
shanear: We're only going to run the reclose script on fully closed appeals for now since the SQL seems way too complicated for ones where they were partially closed and i'm not sure it'll be a problem.

app/models/ramp_election_intake.rb
```diff
@@ -44,18 +45,12 @@ def complete!(_request_params)
...
-        RampClosedAppeal.create!(
-          vacols_id: appeal.vacols_id,
-          ramp_election_id: ramp_election.id,
-          nod_date: appeal.nod_date
-        )
-      end
+      create_ramp_closed_appeals!
     end
+
+    @ramp_closed_appeals.each(&:close!)
```
shanear: Changing the order here because we have a script that will reclose the appeals if things go wrong. This allows us to push the closure logic into `RampClosedAppeal` which DRYs things up a bit.

app/services/appeal_repository.rb
```diff
@@ -442,7 +442,8 @@ def self.reopen_undecided_appeal!(appeal:, user:, safeguards:)
     fail AppealNotValidToReopen if %w[50 51 52 53 54 70 96 97 98 99].include? previous_active_location
 
     adv_status = previous_active_location == "77"
-    fail AppealNotValidToReopen unless adv_status ^ (close_disposition == "9")
+    fail AppealNotValidToReopen if adv_status && (close_disposition == "9")
+    fail AppealNotValidToReopen if !adv_status && (close_disposition != "9")
```
shanear: this is functionally the same. However the XOR operator is incredibly confusing, so I just split it out.

app/models/ramp_closed_appeal.rb
```diff
@@ -3,8 +3,35 @@ class NoReclosingBvaDecidedAppeals < StandardError; end
 
   belongs_to :ramp_election
 
+  attr_writer :user
+
   delegate :established_at, to: :ramp_election
 
+  # If an appeal was only partially closed (because it contained some ineligible issues)
+  # then we record the issue ids that were closed.
```
leikkisa: Since the alternative is "fully_compensation", does this record all issues if there is a mix, whether it's compensation or not?

app/models/ramp_closed_appeal.rb
```diff
@@ -3,8 +3,35 @@ class NoReclosingBvaDecidedAppeals < StandardError; end
...
+    !!partial_closure_issue_sequence_ids
+  end
+
+  def close!
+    if partial?
+      partial_closure_issue_sequence_ids.each do |vacols_sequence_id|
+        Issue.close_in_vacols!(
+          vacols_id: vacols_id,
+          vacols_sequence_id: vacols_sequence_id,
+          disposition_code: "P"
```
leikkisa: I guess here you might need to filter out the non-comp issues.

app/models/ramp_closed_appeal.rb
cmgiven: Makes sense. Perhaps the method should be renamed from `all` in that case.

app/models/ramp_closed_appeal.rb
shanear: basically the two options are "fully" or "partial" compensation. If it's not compensation at all then it won't exist (because its invalid)

app/models/ramp_closed_appeal.rb
shanear: We do that here:

https://github.com/department-of-veterans-affairs/caseflow/pull/6127/files/caa4ab9b31fd6217369704474a692a0e8482146d#diff-075e7ac6170908de9f266a62655264b5R105

PR #6128, opened 2018-07-06
Return all active attorneys in VACOLS

### LINKED ISSUE #6118: API changes for assigning tasks to any attorney
To allow judges to assign tasks to any attorney, we need to change the Queue API to support some new use-cases:

- The API needs to support getting a list of all attorneys. Some of these users won't be in the Users table.
- The API needs to support assigning to users not in the Users table because some attorneys don't have records in the Users table.

aroltsch:
Fixes #6118 

### Description
Return all attorneys in VACOLS. 
```
GET /users?role=attorney
```


app/models/attorney.rb
```diff
@@ -0,0 +1,9 @@
+class Attorney
+  class << self
+    def list_all
+      Rails.cache.fetch("#{Rails.env}_list_of_attorneys_from_vacols") do
```
evankroske: All the environments use different Redis instances, so I don't think that you need to include Rails.env in the cache key

app/models/attorney.rb
```diff
@@ -0,0 +1,9 @@
+class Attorney
+  class << self
+    def list_all
```
evankroske: I feel like it's weird to have a model class that isn't instantiated. Since this method returns Users, I think that we should move it to User. What do you think?

app/repositories/attorney_repository.rb
```diff
@@ -0,0 +1,8 @@
+class AttorneyRepository
+  def self.find_all_attorneys
```
evankroske: I don't think we should add a repository for attorneys because we don't have a real attorney model

client/app/queue/constants.js
```diff
@@ -123,8 +124,8 @@ export const ISSUE_DISPOSITIONS = _.fromPairs(_.zip(
 export const ISSUE_DESCRIPTION_MAX_LENGTH = 100;
 
 export const USER_ROLES = {
-  ATTORNEY: 'Attorney',
```
evankroske: Does this change need to be in this PR? It doesn't seem relevant

evankroske: But you also made the role keys lowercase. Is that change necessary? Do we need to update any frontend code?

app/models/attorney.rb
evankroske: OK, since we're already organizing judge code this way, I think it makes sense to be consistent. Let's leave this code and the repository as it is now. Thanks for the context!

client/app/queue/constants.js
aroltsch: Evan and I chatted in person and decided to upper case the roles in the next PR.

PR #6134, opened 2018-07-06
Connect sidecar health check to /health-check endpoint

RobertAlexanderUSDS:
Connects https://github.com/department-of-veterans-affairs/appeals-deployment/issues/1530
Connects https://github.com/department-of-veterans-affairs/caseflow-commons/pull/156

### Description
The AWS Load Balancer performs a health check on each AWS instance by calling the /health-check endpoint. The application should check the health of all sidecars, so that the LB will replace the instance if any sidecar has issues.

### Acceptance Criteria 
- [x] Passes spec tests

### Testing Plan
1. Running tests


app/controllers/health_checks_controller.rb
```diff
@@ -4,11 +4,24 @@ class HealthChecksController < ActionController::Base
   protect_from_forgery with: :exception
   newrelic_ignore_apdex
 
+  def initialize
+    @pushgateway = Caseflow::PushgatewayService.new
+  end
+
+  def healthy?
+    # Check health of sidecar services
+    if !Rails.deploy_env?(:prod)
```
lowellrex: Would it make sense to have this be some environment variable instead of relying on the environment we've deployed the code to?

lowellrex: I think I'm a little confused about the condition here. If we *are not* in prod, we will return the health of the `Caseflow::PushgatewayService`, but if we *are* in prod then we will always return true, even if `Caseflow::PushgatewayService` is unhealthy. Is that correct?

Is this so that prod continues to function if the pushgateway is down, but the other environments will not function? Is the reason why we would want the lower envs to stop reporting as healthy because this would provide some additional alert to us that we need to fix the pushgateway?

lowellrex: I think it might make sense to do [what we do with efolder](https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/1024/files#diff-1a9ddc0602aede44fbf3c9f54061009dR13)'s environment variable here as well.

RobertAlexanderUSDS: Currently the application sends metrics to a pushgateway instance hosted on a single server (dsva-appeals-<env>-prometheus-ec2). I'm looking to send those metrics to a service that is running local to the application and remove the old service.

I'm rolling this out slowly to reduce the chance of an outage, so the steps are probably going to be:

- Deploy dockerized pushgateway on staging [DONE]
- Health check on staging [this PR]
- Wire the prometheus server to connect to the staging pushgateway containers
- Migrate staging metrics from centralized pushgateway to the local container
- Repeat the above for prod

So when I'm done with all the above, there will be no behavior difference staging vs prod.

PR #6136, opened 2018-07-06
Anya/6131 increase receive timeout

### LINKED ISSUE #6131: ORA-12170: TNS: connect timeout occurred in CircleCI

aroltsch:
Fixes #6131 

ORA-12609: TNS: Receive timeout occurred
Cause: The receive or read operation did not complete within the allowed time interval.
Action: Check if the peer host is available. Increase the receive timeout value if necessary.

From https://github.com/rsim/oracle-enhanced#timeouts:

```
You can set receive and send timeouts, in seconds, using the following TNSNAMES parameters:

RECV_TIMEOUT - the maximum time the OCI libraries should wait for data to arrive on the TCP socket. Internally, it is implemented through a setsockopt(s, SOL_SOCKET, SO_RCVTIMEO). You should set this value to an integer larger than the server-side execution time of your longest-running query.
SEND_TIMEOUT the maximum time the OCI libraries should wait for write operations to complete on the TCP socket. Internally, it is implemented through a setsockopt(s, SOL_SOCKET, SO_SNDTIMEO). Values larger than 5 are a sign of poorly performing network, and as such it should be avoided.
```


config/database.yml
```diff
@@ -105,10 +105,15 @@ test:
...
 test_vacols:
   adapter: oracle_enhanced
-  host: localhost
-  port: 1521
   username: VACOLS
   password: VACOLS
-  database: BVAP.localdomain
+  database: "(DESCRIPTION=
+    (ADDRESS_LIST=(ADDRESS=(PROTOCOL=tcp)(HOST=localhost)(PORT=1521)))
+    (RECV_TIMEOUT=120)(SEND_TIMEOUT=5)
```
aroltsch: Setting it to 120. Default is 60

PR #6143, opened 2018-07-09
Removes error message from empty unassigned cases page

evankroske:
### Description
Removes the error message shown on the unassigned cases page when the judge doesn't have any cases.

### Testing Plan
- [x] Become BVARERDMAN.
- [x] View unassigned cases page.
- [x] Confirm that the assign widget is visible.


lowellrex:
For reference here is the before/after of this change:

| Before | After |
| ------ | ----- |
| image | image |

client/app/queue/UnassignedCasesPage.jsx
```diff
@@ -1,46 +1,32 @@
...
-      {featureToggles.judge_assign_cases &&
-        <AssignWidget previousAssigneeId={userId}
-          onTaskAssignment={(params) => props.initialAssignTasksToUser(params)} />}
-      <JudgeAssignTaskTable {...props} />
-    </React.Fragment>;
-  }
-
-  return tableContent;
+  return <React.Fragment>
+    <h2>Cases to Assign</h2>
```
lowellrex: Can we move this to `COPY.json`? Then this component will be entirely free of strings of copy!

evankroske: Done!

Issue #5174, opened 2018-04-16
Judges | Assign cases | Assign to 'Other'


laurjpeterson:
Allow the judge to search for Other users, when they are not assigning a case to someone on their team.

## AC

- Verify that when the user chooses 'Other' from the drop down, they have a text box to enter the name of an attorney or judge.
- Copy:
- Error state: When the user types in an attorney or judge that is not at the Board, they should receive an error.

## Mocks

evankroske:
@sneha-pai, in the design of the assign widget, how important is it to put the attorneys on the judge's team in a separate dropdown from the other attorneys? It would be much easier for me to implement and maintain this feature if I could use just one dropdown. The attorneys on the judge's team would appear above other attorneys, so judges wouldn't have to search. What do you think about this change to the design?

sneha-pai:
@evankroske This is an excellent and efficient idea. Would it be possible to show any kind of visual cue in the dropdown, to differentiate the judge's attorneys from other attorneys? (Bolding, different colors, etc)

evankroske:
@sneha-pai , sure! I could put a blank between attorneys on the judge's team, or I could simulate a horizontal rule with a bunch of em dashes: `—————`. Would that work from a design perspective?

mkhandekar:
Nice! Wondering if we could also **title** the two sections, kinda like this? 

<img width="132" alt="screenshot 2018-07-09 15 16 20" src="https://user-images.githubusercontent.com/1598889/42478839-0fa8ad9c-838b-11e8-8f83-34855992979c.png">

evankroske:
@mkhandekar, sorry, but [the select widget we're using](http://jedwatson.github.io/react-select/) doesn't support sections, and I don't think that adding support would be feasible.

evankroske:
@mkhandekar, I was wrong; react-select does support sections. I'm going to label each section with the judge's name, OK?

mkhandekar:
@evankroske sounds good! can you drop in a screenshot of what that looks like (here or the eventual PR)?

PR #6144, opened 2018-07-09
Judges | Assign cases | Assign to 'Other'

evankroske:
- Fixes https://github.com/department-of-veterans-affairs/caseflow/issues/5174

### Description
Allows judges to assign cases to any attorney at the board.

Screen shot of assign widget when using it to assign to an attorney not on the user's team

### Testing Plan
- [x] Become BVARERDMAN.
- [x] Assign a case to Lela Shields.
- [x] Become BVAAABSHIRE.
- [x] Confirm that the case is assigned to Lela.
- [x] On "Cases to Assign", assign an unassigned case to Jazlyn Schimmel.
- [x] Become BVARERDMAN.
- [x] Confirm that the case has been assigned to Jazlyn.


client/app/queue/reducers.js
```diff
@@ -325,6 +326,29 @@ const workQueueReducer = (state = initialState, action = {}) => {
     const appealId = action.payload.task.id;
     const appeal = state.tasksAndAppealsOfAttorney[action.payload.previousAssigneeId].data.appeals[appealId];
 
+    const tasksAndAppealsOfAssignee = update(
+      state.tasksAndAppealsOfAttorney[action.payload.assigneeId] ||
```
mdbenjam: Why is this change necessary?

evankroske: Otherwise, when the reducer tries to add the task to the assignee's list of tasks, it throws an exception. The old code assumed that tasksAndAppealsOfAttorney[assigneeId] was defined, which isn't true when the assignee isn't on the assigner's team.

PR #6147, opened 2018-07-09
Fix overall quality in judge checkout

aroltsch:
During judge checkout, we have been sending VACOLS codes instead of human readable keys.


client/app/queue/EvaluateDecisionView.jsx
```diff
@@ -69,7 +69,7 @@ class EvaluateDecisionView extends React.PureComponent {
     path: `/queue/appeals/${this.props.appealId}/evaluate`
   });
 
-  qualityIsDeficient = () => this.state.quality > 0 && this.state.quality < 3;
+  qualityIsDeficient = () => this.state.quality === 'needs_improvements' || this.state.quality === 'does_not_meet_expectations';
```
aroltsch: Not sure what is the "react" way of doing this comparison.

dannysellers: My gut would be to do `['needs_improvements', 'does_not_meet_expectations'].includes(this.state.quality)`, but this is alright given that we're only checking 2 values. Though, this line might be too long for the linter

client/app/queue/EvaluateDecisionView.jsx
```diff
@@ -213,19 +213,19 @@ class EvaluateDecisionView extends React.PureComponent {
         styling={css(marginBottom(0), errorStylingNoTopMargin)}
         errorMessage={highlight && !this.state.quality ? 'Choose one' : null}
         options={[{
-          value: '5',
+          value: 'outstanding',
```
dannysellers: I'm fine with these, but now that we're only using keys/values from `JUDGE_CASE_REVIEW_OPTIONS.QUALITY`, we could pass `options` as:
```javascript
_.map(JUDGE_CASE_REVIEW_OPTIONS.QUALITY, (val, key, obj) => ({
  value: key,
  displayText: `${Object.keys(obj).indexOf(key) + 1} - ${val}`
}))
```

I'm happy to make this change / attach it to this PR

dannysellers: Screenshot w/display and updated `value`s
screen shot 2018-07-09 at 9 44 02 am

PR #6154, opened 2018-07-09
Change tasks controller create action to allow creation of multiple t…

### LINKED ISSUE #6152: If attorneys assign more than one admin action for the case, they should only receive the case back when all admin actions are completed for the case
If attorneys assign more than one admin action for the case, they should only receive the case back when all admin actions are completed for the case

aroltsch:
…asks

Closes #6152 

### Description
Create multiple admin actions at once. The controller `create` action accepts both forms now: one task to create and an array of tasks

### DIFF
--- a/app/models/tasks/co_located_admin_action.rb
+++ b/app/models/tasks/co_located_admin_action.rb
@@ -3,11 +3,11 @@ class CoLocatedAdminAction < Task
   validate :assigned_by_role_is_valid
 
   class << self
-    def create(params)
+    def create(tasks)
       ActiveRecord::Base.multi_transaction do
-        params[:assigned_to] = next_assignee
-        records = params.delete(:titles).each_with_object([]) do |title, result|
-          result << super(params.merge(title: title))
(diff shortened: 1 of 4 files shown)


app/models/tasks/co_located_admin_action.rb
```diff
@@ -3,11 +3,11 @@ class CoLocatedAdminAction < Task
...
 
   class << self
-    def create(params)
+    def create(tasks)
       ActiveRecord::Base.multi_transaction do
-        params[:assigned_to] = next_assignee
-        records = params.delete(:titles).each_with_object([]) do |title, result|
-          result << super(params.merge(title: title))
+        assignee = next_assignee
+        records = [tasks].flatten.each_with_object([]) do |task, result|
```
dannysellers: This and the change to `tasks_params` seem redundant—are `tasks`/`params.require("tasks")` not already flat arrays? Or is this preparation for calling this function once to create multiple sets of tasks?

PR #6159, opened 2018-07-09
Move case details test to FACOLS

lowellrex:
Connects #5688.


spec/factories/vacols/staff.rb
```diff
@@ -4,6 +4,7 @@
       user nil
     end
 
+    sequence(:sattyid)
```
lowellrex: All attorneys and judges should have attorney IDs. The only folks who shouldn't have attorney IDs are colocated and other staff. We will have to set those to nil explicitly going forward.

The reason why we needed this change is because the attorney ID is how we associate a user to a hearing, so it needs to be unique and tied to a particular user.

spec/controllers/tasks_controller_spec.rb
```diff
@@ -74,12 +74,12 @@
...
 
-    context "when user is neither judge nor attorney" do
+    context "when user has no role" do
       let(:role) { nil }
 
-      it "should not process the request succesfully" do
+      it "should return a 400 invalid role error" do
         get :index, params: { user_id: user.id, role: "unknown" }
-        expect(response.status).to eq 302
+        expect(response.status).to eq 400
```
lowellrex: Slight change of behaviour here since this was supposed to be testing that the user failed the [`invalid_role_error`](https://github.com/department-of-veterans-affairs/caseflow/blob/32c70a218cbe69c591e8f5f8dc51a3fb59a452be/app/controllers/tasks_controller.rb#L19) check and not the [`verify_queue_access`](https://github.com/department-of-veterans-affairs/caseflow/blob/32c70a218cbe69c591e8f5f8dc51a3fb59a452be/app/controllers/tasks_controller.rb#L2) check as it was previously.

Slack discussion for reference: https://dsva.slack.com/archives/C6E41RE92/p1531222424000039

mdbenjam: So this is because we're now assigning sattyid by default?

spec/factories/vacols/case_hearing.rb
```diff
@@ -29,19 +29,27 @@
       # We need to reload the correct record before trying to modify it.
       hearing.hearing_pkseq = hearing.hearing_pkseq + 1
       hearing.reload
+
       if evaluator.user
-        staff = create(:staff, :attorney_judge_role, user: evaluator.user)
-        hearing.update(board_member: staff.sattyid)
+        existing_staff = VACOLS::Staff.find_by_sdomainid(evaluator.user.css_id)
```
mdbenjam: we should only have this in one spot, only in build, not create. Build is always called if create is called.

spec/factories/vacols/case_hearing.rb
```diff
@@ -29,19 +29,27 @@
...
+    after(:build) do |hearing, evaluator|
       # For video hearings we need to build the master record.
       if hearing.hearing_type == "V"
         master_record = create(:case_hearing, hearing_type: "C", folder_nr: "VIDEO RO13")
         # For some reason the returned record's sequence is one less than what is actually saved.
         hearing.vdkey = master_record.hearing_pkseq
       end
+
+      if evaluator.user
+        existing_staff = VACOLS::Staff.find_by_sdomainid(evaluator.user.css_id)
```
mdbenjam: Why do we need this existing staff check? This feels like it adds complexity.

spec/feature/queue/case_details_spec.rb
```diff
@@ -1,267 +1,267 @@
...
   before do
-    Fakes::Initializer.load!
     FeatureToggle.enable!(:queue_phase_two)
+    FeatureToggle.enable!(:test_facols)
+
+    User.authenticate!(user: attorney_user)
   end
 
   after do
+    User.unauthenticate!
```
mdbenjam: Why do we need this in the after block?

lowellrex: This may be me being overcautious about not leaking state. I'll go ahead and remove this and see if any tests break.

spec/feature/queue/case_details_spec.rb
```diff
@@ -1,267 +1,265 @@
...
+    let!(:appeal) do
+      FactoryBot.create(
+        :legacy_appeal,
+        :with_veteran,
+        vacols_case: FactoryBot.create(
+          :case,
+          :assigned,
+          user: attorney_user,
+          # Need a non-cancelled dispositon to show the full set of hearing attributes.
+          # https://github.com/department-of-veterans-affairs/caseflow/blob/7c3f80b3e30aec1488663d99731c6387b6044e61/client/app/queue/CaseHearingsDetail.jsx#L33
```
mdbenjam: I think just the first part of this comment is sufficient.

spec/feature/queue/case_details_spec.rb
```diff
@@ -1,267 +1,265 @@
...
-    context "displays issue dispositions" do
-      scenario "from appellant details page" do
-        appeal = vacols_appeals.first
-        visit "/queue"
-        click_on "#{appeal.veteran_full_name} (#{appeal.vbms_id})"
-        expect(page.document.text).to match(/Disposition 1 - Allowed/i)
-      end
+    scenario "case details page shows appropriate text" do
+      visit "/queue"
+      click_on "#{appeal.veteran_full_name} (#{appeal.sanitized_vbms_id})"
```
mdbenjam: We should prefer `file_number` over `sanitized_vbms_id` since the latter is confusing.

spec/controllers/tasks_controller_spec.rb
lowellrex: Correct. We are assigning a value to the `sattyid` field of `vacols.staff` which causes the `verify_queue_access` to pass (because this user has the VACOLS role of attorney), but since `params['role']` is `unknown` (and the `invalid_role_error` check uses the role from the input parameters), we will fail the `invalid_role_error` check.

spec/feature/queue/case_details_spec.rb
lowellrex: Haha yeah for sure, this is definitely the residue of a work-in-progress comment.

PR #6160, opened 2018-07-09
Remove all relations between ramp election and ramp refiling

kahlouie:
Connects #5776

This was blocking #6155

### DIFF
--- a/app/models/ramp_election.rb
+++ b/app/models/ramp_election.rb
@@ -29,6 +29,10 @@ class RampElection < RampReview
   end
 
   def recreate_issues_from_contentions!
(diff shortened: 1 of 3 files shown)


app/models/ramp_election.rb
```diff
@@ -30,10 +29,6 @@ def control_time
   end
 
   def recreate_issues_from_contentions!
-    # If there is a saved refiling for this election, then the issues
-    # are locked and cannot be recreated
-    return false if ramp_refilings.count > 0
```
shanear: don't we still want some version of this check?

PR #6161, opened 2018-07-09
Adds the field judge_css_id to User

evankroske:
Part of https://github.com/department-of-veterans-affairs/caseflow/issues/5174

### Description
Adds a field to User that contains the CSS ID of their judge if they have one.

### Testing Plan
- [x] The JSON objects on http://localhost:3000/users?role=Attorney include judge_css_id.


app/models/user.rb
```diff
@@ -176,6 +176,17 @@ def current_case_assignments
     self.class.appeal_repository.load_user_case_assignments_from_vacols(css_id)
   end
 
+  def judge_css_id
```
aroltsch: This is okay for now. Ideally it would be good to utilize Attorney model since this field is attorney specific. We might need to add more fields that are attorney specific.

evankroske:
@aroltsch, sure! I just added a test

Issue #5593, opened 2018-05-17
Show when a Veteran's claims folder has new documents


laurjpeterson:
As any Board employee, I want to see if there are any new documents in a Veteran's claims folder so I can assess if I still can / continue to work on the case.
   - e.g. As an attorney, if a FOIA request comes in the mail I want to be notified, so I can know that I cannot finish the decision, I need to transfer the case to FOIA.
   - e.g. As a co-located team member, after I've sent correspondence to a Veteran for a hearing clarification and placed the case on hold, I want to see if there are new documents added to the eFolder so I can check if that new document is the Veteran's answer, and I can complete the hold.

### Questions to be aware of
- Who is alerted by new documents in eFolder when a case is in "case storage" or not assigned to anyone?
- If we are showing any user when there are new documents in the eFolder, what happens to the mail team? Do all users know how to handle all mail?

### Where do users get notified of new documents?
- [x] within table row
- [x] case details
- [x] In Progress & On Hold tabs (gmail style) - ❌out of scope for now

### Resources
- SPIKE ticket- https://github.com/department-of-veterans-affairs/caseflow/issues/5410

mkhandekar:
I updated the original issue with **where users get notified of new documents**, and will be working on this this sprint.

mkhandekar:
Work in progress that was shared with the team yesterday:

### Visual indicator within tabs when there are new documents associated with the cases within that tab

<img width="1080" alt="screenshot 2018-06-14 15 27 49" src="https://user-images.githubusercontent.com/1598889/41433658-9734e06c-6fe7-11e8-91f5-3d3ddfe05edb.png">

### Visual indicator in the table when there are new documents associated with that case

<img width="1086" alt="screenshot 2018-06-14 15 27 58" src="https://user-images.githubusercontent.com/1598889/41433699-b4a6d9e8-6fe7-11e8-8de9-365d8b842137.png">

---

The team talked about taking out the document count within the table view, so next I'll work on an iteration showing:
- [ ] A link to all Veteran documents (without document number)
- [ ] That the case has a new documented

mdbenjam:
@mkhandekar  I was talking with @laurjpeterson about doing this one in the next sprint. Is this design finalized? Also was telling LP that if we are interested, it should also be easy to show the number of new documents.

mkhandekar:
[...]

PR #6162, opened 2018-07-10
adds icon for new veteran documents

mkhandekar:
to be part of #5593 **Show when a Veteran's claims folder has new documents**


mdbenjam:
I actually don't think this will work for us. We probably should include it in the RenderFunctions component. The reason we started doing that is because of how we package assets like this. I'm happy to take care of that when I'm implement the feature.

mdbenjam:
Yeah I added it already, so I'm going to close this.

Issue #5362, opened 2018-05-01
Enable source maps in Sentry


amprokop:
https://docs.sentry.io/clients/javascript/sourcemaps/#raven-js-sourcemaps

# AC
- When we receive a Sentry frontend error, we should be able to trace the error to an unbundled, human readable line of code

PR #6165, opened 2018-07-10
Fix asset pipeline to enable uploading sourcemaps

mdbenjam:
Connects: https://github.com/department-of-veterans-affairs/caseflow/issues/5362

### Description
We need to ensure the asset pipeline doesn't digest our webpack bundle since that will throw off sourcemaps.

We remove unused js files, move minifing from rails to webpack, and just let the asset pipeline fingerprint the js files.


client/webpack.config.js
```diff
@@ -1,3 +1,57 @@
-module.exports = require(
-  '@department-of-veterans-affairs/caseflow-frontend-toolkit/config/getWebpackConfig'
-)(__dirname, './app/index');
+const webpack = require('webpack');
```
mdbenjam: Moving back in from the frontend toolkit

PR #6183, opened 2018-07-11
Judges | Assign cases | Assign to 'Other'

evankroske:
- Fixes https://github.com/department-of-veterans-affairs/caseflow/issues/5174

### Description
Allows judges to assign cases to any attorney at the board.

### Testing Plan
- [ ] Become BVARERDMAN.
- [ ] Assign a case to Lela Shields.
- [ ] Become BVAAABSHIRE.
- [ ] Confirm that the case is assigned to Lela.
- [ ] On "Cases to Assign", assign an unassigned case to Jazlyn Schimmel.
- [ ] Become BVARERDMAN.
- [ ] Confirm that the case has been assigned to Jazlyn.


client/app/queue/QueueActions.js
```diff
@@ -249,3 +249,27 @@ export const reassignTasksToUser = ({ tasks, assigneeId, previousAssigneeId }) =
...
+    error
+  }
+});
+
+export const fetchAllAttorneys = () => (dispatch) => {
+  return ApiUtil.get(
+    '/users?role=Attorney').
+    then((resp) => resp.body).
+    then(
+      (resp) => dispatch(receiveAllAttorneys(resp.attorneys))).
```
dannysellers: Total nitpick: This one line is formatted differently than the others (as is the first one, actually)

client/app/queue/reducers.js
```diff
@@ -38,7 +38,8 @@ export const initialState = {
   },
   attorneysOfJudge: [],
   tasksAndAppealsOfAttorney: {},
-  isTaskAssignedToUserSelected: {}
+  isTaskAssignedToUserSelected: {},
+  allAttorneys: {}
```
dannysellers: Why not just `attorneys`? We don't have `allJudges`

client/app/queue/reducers.js
```diff
@@ -359,6 +360,22 @@ const workQueueReducer = (state = initialState, action = {}) => {
...
   }
+  case ACTIONS.RECEIVE_ALL_ATTORNEYS:
+    return update(state, {
+      allAttorneys: {
+        $set: {
+          data: action.payload.attorneys
+        }
+      }
+    });
+  case ACTIONS.ERROR_ALL_ATTORNEYS:
```
dannysellers: I feel like reducer actions should be semantically tied to the action they're related to. `ERROR_ALL_ATTORNEYS` doesn't have any kind of verb / indication that it's related to loading, rather than something like form validation. Why not name it something like `ERROR_LOADING_ATTORNEYS`/`ERROR_LOADING_ALL_ATTORNEYS`?

client/app/queue/JudgeAssignTaskListView.jsx
```diff
@@ -58,17 +58,14 @@ class JudgeAssignTaskListView extends React.PureComponent {
...
-    };
-
-    return ApiUtil.get(`/users?role=Attorney&judge_css_id=${this.props.userCssId}`, requestOptions).
+    return this.props.fetchAllAttorneys().
       then(
-        (response) => {
-          const resp = JSON.parse(response.text);
+        (attorneys) => {
+          const attorneysOfJudgeWithCssId = _.groupBy(attorneys, (attorney) => attorney.judge_css_id);
+          const attorneysOfJudge = attorneysOfJudgeWithCssId[this.props.userCssId];
```
dannysellers: Nit: why not just filter? `attorneys.filter(attorney) => attorney.judge_css_id === this.props.userCssId)`

client/app/queue/components/AssignWidget.jsx
```diff
@@ -84,9 +91,50 @@ class AssignWidget extends React.PureComponent<Props> {
...
+        }
+        options.push({
+          label: `${judgeName}:`,
+          value: judgeCssId,
+          disabled: true
+        });
+        options.push(...attorneysOfJudgeWithCssId[judgeCssId].map(optionFromAttorney));
+      }
+      options.push({ label: ASSIGN_WIDGET_OTHER,
+        value: OTHER });
```
dannysellers: Why `OTHER` and not just `ASSIGN_WIDGET_OTHER.toUpperCase()`?

client/app/queue/components/AssignWidget.jsx
```diff
@@ -104,7 +152,7 @@ class AssignWidget extends React.PureComponent<Props> {
           searchable
           options={options}
           placeholder="Select a user"
-          onChange={(option) => this.props.setSelectedAssignee({ assigneeId: option.value })}
+          onChange={handleChange}
```
dannysellers: Nit: why extract this callback / move it to a different part of the file? It looks like this is the only place it's used, but now it's separated by 50 lines of indirectly-related / unrelated code

client/app/queue/components/AssignWidget.jsx
```diff
@@ -84,9 +91,50 @@ class AssignWidget extends React.PureComponent<Props> {
...
+      return <Alert type="error" title="Error fetching attorneys" message="Please refresh the page." />;
+    }
+    const optionFromAttorney = (attorney) => ({ label: attorney.full_name,
+      value: attorney.id.toString() });
+    const handleChange = (option) => this.props.setSelectedAssignee({ assigneeId: option.value });
+    const attorneys = allAttorneys.data;
+
+    const options = [];
+
+    {
```
dannysellers: Why use a block here? It doesn't look like there're any variables declared inside it that need a separate scope, and if it's just a visual/semantic distinction, it should probably just go in its own function

client/app/queue/components/AssignWidget.jsx
```diff
@@ -84,9 +91,50 @@ class AssignWidget extends React.PureComponent<Props> {
...
+      const attorneysOfJudgeWithCssId = _.groupBy(attorneys, (attorney) => attorney.judge_css_id);
+      const judgeCssIds =
+        [userCssId].concat(Object.keys(attorneysOfJudgeWithCssId).filter((cssId) => cssId !== userCssId));
```
dannysellers: I don't think we need to handle `userCssId` separately/specially here—wouldn't it already be included in the set of all judges? Why not something like `judgeCssIds = _.uniq(_.map(attorneys, 'judge_css_id'))`?

client/app/queue/components/AssignWidget.jsx
```diff
@@ -83,10 +90,60 @@ class AssignWidget extends React.PureComponent<Props> {
...
+    }
+
+    const options = [];
+
+    options.push({
+      label: `${this.fullNameOfJudgeWithCssId(judgeCssId) || 'Other'}:`,
+      value: judgeCssId,
+      disabled: true
+    });
+    options.push(...attorneysOfJudgeWithCssId[judgeCssId].map((attorney) => ({ label: attorney.full_name,
```
dannysellers: Thank you for moving these into separate functions. It looks like `options` might not be needed any more here—could it just return the result of this map?

client/app/queue/components/AssignWidget.jsx
```diff
@@ -83,10 +90,60 @@ class AssignWidget extends React.PureComponent<Props> {
...
+
+    {
+      const attorneysOfJudgeWithCssId = _.groupBy(attorneys, (attorney) => attorney.judge_css_id);
+      const judgeCssIds =
+        [userCssId].concat(Object.keys(attorneysOfJudgeWithCssId).filter((cssId) => cssId !== userCssId));
+
+      for (const judgeCssId of judgeCssIds) {
+        options.push(...this.optionsFromJudgeAndAttorneys(judgeCssId, attorneysOfJudgeWithCssId));
+      }
+      options.push({ label: ASSIGN_WIDGET_OTHER,
```
dannysellers: It looks like the textbox described in the AC is missing—this just would assign to Other
> Verify that when the user chooses 'Other' from the drop down, they have a text box to enter the name of an attorney or judge.

client/app/queue/components/AssignWidget.jsx
```diff
@@ -83,10 +90,60 @@ class AssignWidget extends React.PureComponent<Props> {
...
+        value: OTHER });
```
dannysellers: Why use a different constant? `ASSIGN_WIDGET_OTHER.toUpperCase()` should work

client/app/queue/components/AssignWidget.jsx
```diff
@@ -83,10 +90,60 @@ class AssignWidget extends React.PureComponent<Props> {
...
+
+    const options = [];
+
+    {
+      const attorneysOfJudgeWithCssId = _.groupBy(attorneys, (attorney) => attorney.judge_css_id);
+      const judgeCssIds =
+        [userCssId].concat(Object.keys(attorneysOfJudgeWithCssId).filter((cssId) => cssId !== userCssId));
+
+      for (const judgeCssId of judgeCssIds) {
+        options.push(...this.optionsFromJudgeAndAttorneys(judgeCssId, attorneysOfJudgeWithCssId));
```
dannysellers: Nit: It looks like we could build `options` with a `map` at this point

client/app/queue/components/AssignWidget.jsx
```diff
@@ -83,10 +90,60 @@ class AssignWidget extends React.PureComponent<Props> {
...
   render = () => {
-    const { attorneysOfJudge, selectedAssignee, error, success } = this.props;
-    const options = attorneysOfJudge.map((attorney) => ({ label: attorney.full_name,
-      value: attorney.id.toString() }));
+    const { selectedAssignee, error, success, allAttorneys, userCssId } = this.props;
+
+    if (!allAttorneys.data && !allAttorneys.error) {
+      return <SmallLoader message="Loading..." spinnerColor={LOGO_COLORS.QUEUE.ACCENT} />;
+    }
+    if (allAttorneys.error) {
```
dannysellers: Nit: If this was moved above the check for `!data && !error`, we could remove the `!error` check
```javascript
if (error) {
  return
}
if (!data) {
  return
}
continue
```

client/app/queue/components/AssignWidget.jsx
evankroske: The user's attorneys needs to come first so that the user can easily find their attorneys when there are a lot of attorneys in the dropdown. This code just puts the current user's CSS ID first in the list so that there attorneys are first.

client/app/queue/AssignedCasesPage.jsx
```diff
@@ -63,7 +63,8 @@ class AssignedCasesPage extends React.PureComponent {
       {featureToggles.judge_assign_cases &&
         <AssignWidget
           previousAssigneeId={attorneyId}
-          onTaskAssignment={(params) => props.reassignTasksToUser(params)} />}
+          onTaskAssignment={(params) => props.reassignTasksToUser(params)}
+          userCssId={this.props.userCssId} />}
```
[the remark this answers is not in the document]
evankroske: Makes sense. I'll make it global

client/app/queue/JudgeAssignTaskListView.jsx
```diff
@@ -127,12 +124,14 @@ class JudgeAssignTaskListView extends React.PureComponent {
...
-                userId={this.props.userId.toString()} />}
+                userId={this.props.userId.toString()}
+                userCssId={this.props.userCssId} />}
           />
           <PageRoute
             path={`${match.url}/:attorneyId`}
             title="Assigned Cases | Caseflow"
-            component={AssignedCasesPage}
+            render={
+              ({ match: innerMatch }) => <AssignedCasesPage userCssId={this.props.userCssId} match={innerMatch} />}
```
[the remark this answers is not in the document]
evankroske: Sounds good! I'll make this change

client/app/queue/QueueActions.js
evankroske: OK, I'll fix it

client/app/queue/reducers.js
evankroske: Makes sense! I'll rename this field

client/app/queue/reducers.js
evankroske: OK, I'll rename it

client/app/queue/JudgeAssignTaskListView.jsx
evankroske: Good point! I'll change it

client/app/queue/components/AssignWidget.jsx
evankroske: Good point! I'll change it

client/app/queue/components/AssignWidget.jsx
evankroske: This was left over from a previous version of the code. I'll get rid of it

client/app/queue/components/AssignWidget.jsx
evankroske: Thanks for pointing this out! I forgot to remove this option. I talked to the designers, and we decided to just put all of the attorneys in one dropdown instead of adding a second dropdown. This code was left over from an earlier version.

client/app/queue/components/AssignWidget.jsx
evankroske: OK, I'll change this

client/app/queue/components/AssignWidget.jsx
evankroske: OK, I'll change this

client/app/queue/components/AssignWidget.jsx
evankroske: This was left over from an earlier version. I'll change it back

Issue #6093, opened 2018-07-02
Save EndProductEstablishment to the DB


shanear:
Now that we moved a lot of the end product creation logic to the `EndProductEstablishment` model, we can begin saving.

`EndProductEstablishment` should represent both a point in time where we established an end product, as well as a continual synchronization with the current values of that End Product in VBMS/BGS.

This will allow us to create and track multiple end products for the same AMA review, which is something we will eventually need to do

blocks  #6148 

## Plan of attack
1) Create table with established_at, end_product_status, id (primary id), polymorphic source_id and source_type, :veteran_file_number, :reference_id, :claim_date, :code, :valid_modifiers, :station
Add index to retrieve the end_product_establishments in each RAMP and AMA model with source_id, source_type and veteran_file_number

2) Before creating a new end_product_establishment, check to see if there's an existing established end product in the db with veteran file number and reference id, if so, don't create a new one. Update perform to set established_at and save.

3) Add migration to backfill RAMP end product establishments.

4) Remove double writing to AMA models (what usually happens after `perform!` is called)

5) Update sync job -> app/jobs/ramp_election_sync_job.rb (app/jobs/sync_reviews_job.rb) needs to now sync end_product_establishment instead and still needs to do `RampElection.recreate_issues_from_contentions!`. `RampReview.sync_ep_status!` can likely be moved to EndProductEstablishment. If we use source.sync from EndProductEstablishment, then we can do any model specific sync stuff.

6) Remove double writing to the RAMP models.

leahbannon:
all of the existing models need to be refactored to use this 
main thing: need many:one relationship with AMA review (HLR or SC only)

open questions:
- RAMP?

leikkisa:
Regarding validation, I noticed while looking at UAT data earlier today that the EndProductEstablishments table was not saving the modifier to the DB, although the EndProducts were correctly being created, and other data was being saved to the database.

I checked prod data, and there was a record with a modifier on 7/27, which should have been after this was merged, everything since then is missing it though.  So more investigation is required (like when the PR got deployed, whether it was due to other VBMS issues).  

It's possible we will have to backfill prod data, which should hopefully not be too bad because RampElection is just one end product modifier code.

leikkisa:
For validation:
Kara or Sally to look at UAT intakes that have happened after this, and check the EndProductEstablishments table data for correctness
[...]

PR #6192, opened 2018-07-11
Remove double writing of end product reference id on reviews

kahlouie:
Connects #6093


spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -218,8 +218,8 @@
 
     expect(intake).to be_success
 
-    higher_level_review.reload
-    expect(higher_level_review.end_product_reference_id).to eq("IAMANEPID")
+    resultant_end_product_establishment = EndProductEstablishment.find_by(source: higher_level_review.reload)
```
shanear: `resultant` haha, i've never used that word before. Nice.

app/models/higher_level_review.rb
```diff
@@ -23,21 +23,12 @@ def special_issues
...
 
-  def establish_end_product!
-    end_product_establishment.perform!
-
-    update!(
-      end_product_reference_id: end_product_establishment.reference_id,
-      established_at: Time.zone.now
-    )
-  end
-
```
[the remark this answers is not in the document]
kahlouie: @leikkisa That would be something needed in the future, but not on this one. If we did the migration in this pr, it would remove that column before the code changes. What that means is that we'd still be trying to write `end_product_reference_id` on the RAMP models when they don't exist in the db. Does that make sense to you?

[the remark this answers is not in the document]
leikkisa: Yeah, it makes sense that it wouldn't hurt anything (will always be nil, right?), but I think it's okay to leave this line out too.  Still want to check on the ramp election rollback code, but otherwise I think it looks good!

app/controllers/higher_level_reviews_controller.rb
```diff
@@ -4,7 +4,8 @@ class HigherLevelReviewsController < ApplicationController
   private
 
   def higher_level_review
-    @higher_level_review ||= HigherLevelReview.find_by!(end_product_reference_id: params[:claim_id])
+    @higher_level_review ||=
+      EndProductEstablishment.find_by!(reference_id: params[:claim_id], source_type: "HigherLevelReview").source
```
ferlatte: The raw string here feels a little magical; could this be a const variable somewhere instead?

app/controllers/supplemental_claims_controller.rb
```diff
@@ -4,7 +4,8 @@ class SupplementalClaimsController < ApplicationController
   private
 
   def supplemental_claim
-    @supplemental_claim ||= SupplementalClaim.find_by!(end_product_reference_id: params[:claim_id])
+    @supplemental_claim ||=
+      EndProductEstablishment.find_by!(reference_id: params[:claim_id], source_type: "SupplementalClaim").source
```
ferlatte: Same. Perhaps EndProductEstablishment should export the various source types as class variables.

app/models/ramp_review.rb
```diff
@@ -63,7 +63,7 @@ def end_product_active?
   end
 
   def end_product_canceled?
-    sync_ep_status! && end_product_status == "CAN"
+    sync_ep_status! && end_product_establishment.synced_status == "CAN"
```
ferlatte: Maybe EndProductEstablishment.SYNC_STATUS_CANCELED ?

kahlouie: I like that!

spec/factories/ramp_election.rb
```diff
@@ -5,10 +5,6 @@
 
     trait :established do
       established_at Time.zone.now
-
```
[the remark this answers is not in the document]
kahlouie: I'd like to do this in a separate pr, but mostly because I'd like to make an EndProductEstablishment factory as well.

kahlouie:
@ferlatte I've added in those changes. If it looks good to you, could you please add the `Ready-to-Merge` label? Thanks

kahlouie:
Since I handled the magic strings, I'm going to merge this since I've got Sally's :+1:

PR #6193, opened 2018-07-11
Queue | Add document ID validations during attorney check out

### LINKED ISSUE #5993: Add document ID validations during attorney and judge check out
The Board found that a common error at BVA dispatch is incorrect document IDs. Paul Saindon and the IDT team have requested that Caseflow validate the various formats of a document ID during attorney and judge check out to reduce potential for errors at BVA dispatch.

They would also like us to validate that the decision document is located at a particular path (e.g. Located at “\\bvacofil1.dva.va.gov\dt1\Board\” & Left(theID, 5) & theID & “.docx”). This is not in scope for this github issue.
 
## AC
- Verify that all decisions have document IDs with any of the below formats:
  - “#####-########” 
  - "#########.###" 
  - "#########.####" 

- Verify that all VHA work products have document IDs with either of the below formats:
  - "V#######.###" 
  -  "V#######.####"

- Verify that all IME work products have document IDs with either of the below formats:
   - "M#######.###"
   - "M#######.####"

- If the document ID does not have that validation, prompt the user with the required validation

dannysellers:
Resolves #5993 

### Description
Adds validation to `document_id` within Attorney checkout flow (SubmitDecisionView)

All OMO decisions should have a `document_id` with one of two formats:
New:
```regex
/^\d{5}-\d{8}$/
```
Old (decisions not prepared using the IDT):
```regex
/^\d{8}\.\d{3,4}$/
```

VHA work product document id format:
```regex
/^V\d/{7}\.\d{3,4}$/
```

IME work product document id format:
```regex
/^M\d{7}\.\d{3,4}$/
```

### Acceptance Criteria 
- [ ] Within SubmitDecisionView, Document ID should validate to either generic regex / specific old work product type regexes depending on OMO type

### Testing Plan
1. Go to `/queue` as attorney (`BVASCASPER1`), case detail view, Medical Request Ready for Review
2. Confirm Document ID validates correctly

</details>


client/app/queue/SubmitDecisionView.jsx
```diff
@@ -32,18 +33,46 @@ import {
...
+import type {
+  Task,
+  LegacyAppeal,
+  Judges
+} from './types/models';
+import type { UiStateError } from './types/state';
+
+type Props = {|
+  appealId: string,
+  nextStep: string,
```
evankroske: FYI, I've started [splitting the parameters](https://github.com/department-of-veterans-affairs/caseflow/pull/6248/files#diff-ea3ba586136c8e05bb0d8a969c557c59R19) (passed in by the caller) from the properties (from the state and action creators) so that Flow can tell the caller when they're calling the component with the wrong arguments. You'd also need to annotate the return value of connect [like so](https://github.com/department-of-veterans-affairs/caseflow/pull/6248/files#diff-ea3ba586136c8e05bb0d8a969c557c59R125).

client/app/queue/SubmitDecisionView.jsx
```diff
@@ -32,18 +33,46 @@ import {
...
+  appeal: LegacyAppeal,
+  decision: Object,
+  task: Task,
+  judges: Judges,
+  highlightFormItems: Boolean,
+  userRole: string,
+  selectingJudge: Boolean,
+  error: ?UiStateError,
+  // dispatch props
+  setDecisionOptions: Function,
```
evankroske: FYI, if you annotate the action creators [like so](https://github.com/department-of-veterans-affairs/caseflow/pull/6248/files#diff-cd116eea7c5ee31c3d5a80861ff7d711R27), Flow will tell you if you call one with the wrong arguments.

client/app/queue/constants.js
```diff
@@ -98,6 +98,14 @@ export const DRAFT_DECISION_OPTIONS = [{
   value: DECISION_TYPES.OMO_REQUEST
 }];
 
+export const OMO_ATTORNEY_CASE_REVIEW_WORK_PRODUCT_TYPES = [{
+  displayText: 'OMO - VHA',
```
evankroske: I feel like display text belongs in COPY, but I could be wrong. What do you think?

client/app/queue/types/models.js
```diff
@@ -11,6 +11,20 @@ export type User = {
   display_name: string
 };
 
+export type Judge = {
```
evankroske: Are judges different from Users? I thought that they had the same fields. However, if they just have a few extra fields, you can use an [intersection type](https://flow.org/en/docs/types/intersections/#toc-intersections-of-object-types) to reduce duplication.

client/app/queue/types/models.js
```diff
@@ -24,6 +38,22 @@ export type AppellantAddress = {
   country: string
 };
 
+export type Issue = {
```
evankroske: Thanks for adding this!

Issue #5906, opened 2018-06-13
Remove navigation breadcrumbs


mdbenjam:
We've decided to remove the navigational breadcrumbs in Queue.

# AC
1) There are no breadcrumbs while navigating between pages.

PR #6201, opened 2018-07-12
Remove breadcrumbs from queue

lowellrex:
Connects #5906. This change removes breadcrumbs from queue pages.

#### Removed breadcrumbs when navigating to a case details page from the personalized task queue
| Before | After |
| ------ | ----- |
| image | image |

#### Removed breadcrumbs when navigating to a case details page from case search
| Before | After |
| ------ | ----- |
| image | image |

#### Removed breadcrumbs from case checkout flow
| Before | After |
| ------ | ----- |
| image | image |


mkhandekar:
It's a big day on caseflow!

PR #6204, opened 2018-07-12
Split appellant details into separate box when Veteran is not appellant

lowellrex:
[Related slack discussion](https://dsva.slack.com/archives/C6E41RE92/p1531428477000130).

### If the Veteran is not the appellant
| Before | After |
| ------ | ----- |
| image | image |

### If the Veteran IS the appellant
| Before | After |
| ------ | ----- |
| image | image |


lowellrex:
cc @mkhandekar

client/app/queue/VeteranDetail.jsx
```diff
@@ -0,0 +1,106 @@
...
+import COPY from '../../COPY.json';
+import { DateString } from '../util/DateUtil';
+
+const detailListStyling = css({
+  paddingLeft: 0,
+  listStyle: 'none',
+  marginBottom: '3rem'
+});
+const addressIndentStyling = (secondLine) => css({
+  marginLeft: secondLine ? '7.5em' : 0
```
evankroske: Optional: Could you use the string zero instead of the number zero for consistency?

client/app/queue/VeteranDetail.jsx
```diff
@@ -0,0 +1,106 @@
...
+});
+
+export default class VeteranDetail extends React.PureComponent {
+  getAppealAttr = (attr) => _.get(this.props.appeal.attributes, attr);
```
evankroske: FYI: As long as we use _.get, Flow won't be able to save us from field name typos. I can see why we parameterized the field names, but if there's any way to avoid it, Flow will be a lot more useful. Also, it seems like most of this code is was just moved in this PR, not introduced, so it seems like this isn't the right PR to start ripping out _.get.

PR #6209, opened 2018-07-13
Queue | Annotates action creators with types

evankroske:
### Description
Adds type annotations to the action creators of Queue.

### Testing Plan
N/A


client/app/queue/AddEditIssueView.jsx
```diff
@@ -11,8 +11,7 @@ import {
   startEditingAppealIssue,
   cancelEditingAppealIssue,
   saveEditedAppealIssue,
-  deleteEditingAppealIssue,
-  editAppeal
```
evankroske: This action creator isn't used in this file.

PR #6219, opened 2018-07-13
Updates to Readme

aroltsch:
1. Separate command by line otherwise it looks as a one-liner


README.md
```diff
@@ -68,8 +68,10 @@ You'll need to install the libraries required to connect to the VACOLS Oracle da
 
 3) Install via Homebrew:
 
+```
```
aroltsch: Without this change it looks like this: 
image

README.md
```diff
@@ -20,6 +20,7 @@ Install the base dependencies via Homebrew:
     brew install rbenv nodenv yarn
     brew tap ouchxp/nodenv
     brew install nodenv-nvmrc
+    brew install postgresql
```
aroltsch: We need to install here otherwise `bundle install` won't succeed without PG config file.

README.md
```diff
@@ -119,14 +130,16 @@ This should install clean. If you have errors, try ... FIXME.
 
 ### Setup the development Postgres user
 
-Add these to your `.bashrc`:
+Add these to your `.bash_profile`:
```
aroltsch: bashrc didn't work for me

README.md
```diff
@@ -39,6 +40,14 @@ Once you've done that, close your terminal window and open a new one. Verify tha
...
 If you don't see both, stop and debug.
 
+### Install Git
+
+```
+brew install git
+```
+
+We are using 2-factor authentication with Github so, for example, when you access a repository using Git on the command line using commands like `git clone`, `git fetch`, `git pull` or `git push` with HTTPS URLs, you must provide your GitHub username and your personal access token when prompted for a username and password. Follow directions [here](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/) on how to do that.
+
```
ferlatte: git comes with the xcode tools that get installed above (which you need for bundle install to work also); you don't need to install it separately if you don't want to. Do we need it from brew for some other reason?

README.md
```diff
@@ -119,14 +130,16 @@ This should install clean. If you have errors, try ... FIXME.
...
+Add these to your `.bash_profile`:
 
 ```
 export POSTGRES_HOST=localhost
 export POSTGRES_USER=postgres
 export POSTGRES_PASSWORD=postgres
 ```
 
+(Reload the file `source ~/.bash_profile`)
+
```
ferlatte: Depending on what else is in there, this isn't a safe operation. Better to tell folks to close their terminal and make a new one.

README.md
mdbenjam: Really? That sucks :/

README.md
mdbenjam: Oh really? Why might this not be safe? Is this for non idempotent things in there?

ferlatte:
There can be; I’ve had it be an issue on other projects, so I tend to document that folks should just start a new terminal to ensure a clean and debugable environment.

It’s probably not that important.

M

PR #6226, opened 2018-07-14
Add new file indicator

mdbenjam:
Connects #5593 

### Description
Adds the new file indicator to the Reader Link. Adds a new endpoint which currently returns all of the new files, since we think we might want more than just if there are new files in the future (this will make changes to surface what types of files easier.)

### Acceptance Criteria 
- [ ] New file icons appear when there are new files.

### Testing Plan
1. Connect to eFolder
  1. Make sure eFolder is running on port 4000
  1. Turn on the `use_efolder_locally` flag in `development.rb`
  1. Change line 37 of efolder_service to `headers = { "FILE-NUMBER" => "DEMO#{vbms_id}" }`
1. Load a queue and ensure all the files have the new file indicator.
1. Click on one make sure it goes away.
1. Add a file in the eFolder DB to that case and ensure the file indicator comes back.


client/app/queue/QueueActions.js
```diff
@@ -22,6 +22,28 @@ export const onReceiveJudges = (judges: UsersById) => ({
...
+    dispatch({
+      type: ACTIONS.RECEIVE_NEW_FILES,
+      payload: {
+        appealId,
+        newDocuments: resp.new_documents
+      }
+    });
+  }, (error) => {
+    dispatch({
+      type: ACTIONS.ERROR_ON_RECEIVE_NEW_FILES,
```
mdbenjam: @mkhandekar what should we do if we get an error from the server when we try to show if there are new files? I'm thinking we show nothing.

client/app/queue/reducers.js
```diff
@@ -87,6 +88,27 @@ const workQueueReducer = (state = initialState, action = {}) => {
...
+          }
+        }
+      }
+    });
+  case ACTIONS.ERROR_ON_RECEIVE_NEW_FILES:
+    return update(state, {
+      newDocsForAppeal: {
+        [action.payload.appealId]: {
+          $set: {
+            error: action.payload.error
```
mdbenjam: We don't actually use this error anywhere right now... but thought we should still have an error case... Thoughts?

client/app/queue/components/NewFile.jsx
```diff
@@ -0,0 +1,50 @@
...
+    if (this.props.docs && this.props.docs.length > 0) {
+      return <NewFileIcon />;
+    }
+
+    return null;
+
+  }
+}
+
+NewFile.propTypes = {
```
evankroske: FYI, you could use Flow to give users of this component type errors sooner. However, they would only get type errors if they had enabled Flow for the file. Example: https://github.com/department-of-veterans-affairs/caseflow/pull/6248/files#diff-fbcfcbf885226ceffe45ee2726838f4eR34

client/app/queue/reducers.js
```diff
@@ -87,6 +90,27 @@ const workQueueReducer = (state = initialState, action = {}) => {
         }
       }
     });
+  case ACTIONS.RECEIVE_NEW_FILES:
+    return update(state, {
+      newDocsForAppeal: {
+        [action.payload.appealId]: {
+          $set: {
+            error: null,
```
evankroske: I don't think that you need to set error to null here because I don't think that immutability-helper merges objects. If you remove this line, error should be undefined after this update call. newDocsForAppeal[appealId] should be completely replaced when you use `$set: { docs: foo }`.

client/app/queue/QueueActions.js
mkhandekar: I'm wondering how long there might be a potential error?

mkhandekar: In the case that there is an error and we can't show the "new" icon, there are two options:
- show nothing. The user can still cick on "view documents" and continue with their tasks. Don't see many downsides to this, hoping that the error resolves in a few hours, and the next time they come to their Queue they should be seeing things normally.
- we could potentially show a small icon/something (similar size to the "new" icon) with hover text that says _something_ like "We don't know if this case has new documents. Please View Documents to check."

mkhandekar: Oh I completely agree that _we_ should be getting a notification (however you see fit) that the feature is broken.

Issue #5823, opened 2018-06-07
Flow 1 | Build Schedule | Success (saved to VACOLS)


Chingujo:
## User Stories 
When Justin and Scott confirm upload to VACOLS, they need to know if the action has been successful or not, so they can re-upload the spreadsheet, apply algorithm, and save to VACOLS. 

## AC
- [ ] Verify that there is confirmation banner with the following copy: 
#### RO
 - Title: You have successfully assigned hearings between mm/dd/yyyy - mm/dd/yyyy 
 - Body: You can view what you upload by clicking the link below or at the table on the bottom of the page. 

#### Judge
 - Title: You have successfully assigned Judges to hearings between mm/dd/yyyy
- Body: You can view what you upload by clicking the link below or at the table on the bottom of the page. 
#### RO and CO
![final upload](https://user-images.githubusercontent.com/23316107/41117213-5dfd65fe-6a5a-11e8-9021-bd98daa33245.png)
#### Judge
![final upload judges](https://user-images.githubusercontent.com/23316107/41120278-0b69c5ea-6a63-11e8-9238-d4567b332a94.png)

PR #6228, opened 2018-07-14
VACOLS upload success page

sharonwarner:
Connects #5823 

### Description
This PR adds a success page for the VACOLS upload. The page doesn't quite match the mock because of this issue that will be fixed in a different PR: https://github.com/department-of-veterans-affairs/caseflow/issues/6229

### Testing Plan
1. Go to hearings/schedule/build
1. Go through flow 1 twice, once with an ro/co upload and then with a judge upload
1. Confirm that the pages all render correctly (particularly the success page)


tejans24:
* On the `hearings/schedule/build/upload` page
* I selected `RO/CO hearings` radio button
* I selected `validRoSpreadsheet.xlsx` file from the test date in the repository

it looks like `TOGGLE_UPLOAD_CONTINUE_LOADING` is being called twice and the upload won't move forward after selecting `validRoSpreadsheet.xlsx` file

screen shot 2018-07-15 at 3 37 22 pm

sharonwarner:
Yes, the request on the last page will be updated in a separate pr to set finalized to true on the schedule period. On the confirmation page, if the schedule period is finalized, we redirect to the home page.

Issue #6222, opened 2018-07-13
Case Details | Actions dropdown disappears on refresh


evankroske:
## Steps to Reproduce the Issue
- Become BVAAABSHIRE.
- Go to /queue.
- Click on a case.
- Refresh the page.

## Example
![untitled](https://user-images.githubusercontent.com/357369/42710776-765cb008-86b3-11e8-80b5-589d527efa0a.gif)

mdbenjam:
https://media.giphy.com/media/qyuxJs79G0PVC/giphy.gif

PR #6236, opened 2018-07-16
Queue | Adds endpoint to get tasks by VACOLS ID

evankroske:
Part of https://github.com/department-of-veterans-affairs/caseflow/issues/6222

### Description

### Testing Plan
- [ ] Become BVAAABSHIRE.
- [ ] View http://localhost:3000/appeals/3449656/tasks .
- [ ] Confirm that the endpoint returned a JSON task.
- [ ] Repeat as BVARERDMAN.
- [ ] Repeat as BVALSPORER.
- [ ] Repeat as BVASCASPER1.

### DIFF
--- a/app/controllers/appeals_controller.rb
+++ b/app/controllers/appeals_controller.rb
@@ -22,9 +22,11 @@ class AppealsController < ApplicationController
   end
 
   def tasks
+    no_cache
+
     tasks, _ = LegacyWorkQueue.tasks_with_appeals_of_appeal(params[:appeal_id])
     render json: {
-      tasks: json_tasks(tasks)
+      tasks: json_tasks(tasks)[:data]
     }
   end
 
--- a/app/controllers/appeals_controller.rb
+++ b/app/controllers/appeals_controller.rb
@@ -21,6 +21,15 @@ class AppealsController < ApplicationController
     render e.serialize_response
   end
 
+  def tasks
+    no_cache
+
+    tasks, _ = LegacyWorkQueue.tasks_with_appeals_of_appeal(params[:appeal_id])
+    render json: {
+      tasks: json_tasks(tasks)
+    }
+  end
+
   def show
     no_cache
 
@@ -93,6 +102,13 @@ class AppealsController < ApplicationController
   def json_appeals(appeals)
     ActiveModelSerializers::SerializableResource.new(
       appeals
+    ).as_json
+  end
+
+  def json_tasks(tasks)
+    ActiveModelSerializers::SerializableResource.new(
+      tasks,
+      each_serializer: ::WorkQueue::LegacyTaskSerializer
     ).as_json
   end
 end
--- a/app/controllers/appeals_controller.rb
+++ b/app/controllers/appeals_controller.rb
@@ -22,8 +22,6 @@ class AppealsController < ApplicationController
   end
 
   def tasks
-    no_cache
-
     tasks, _ = LegacyWorkQueue.tasks_with_appeals_of_appeal(params[:appeal_id])
     render json: {
       tasks: json_tasks(tasks)
--- a/app/controllers/legacy_tasks_controller.rb
+++ b/app/controllers/legacy_tasks_controller.rb
@@ -2,7 +2,7 @@ class LegacyTasksController < ApplicationController
   before_action :verify_queue_access
   before_action :verify_task_assignment_access, only: [:create, :update]
 
-  ROLES = %w[judge attorney].freeze
+  ROLES = %w[judge attorney colocated].freeze
 
   def set_application
     RequestStore.store[:application] = "queue"
--- a/app/controllers/legacy_tasks_controller.rb
+++ b/app/controllers/legacy_tasks_controller.rb
@@ -2,7 +2,7 @@ class LegacyTasksController < ApplicationController
   before_action :verify_queue_access
   before_action :verify_task_assignment_access, only: [:create, :update]
 
-  ROLES = %w[judge attorney colocated].freeze
+  ROLES = %w[judge attorney].freeze
 
   def set_application
     RequestStore.store[:application] = "queue"
--- a/app/controllers/legacy_tasks_controller.rb
+++ b/app/controllers/legacy_tasks_controller.rb
@@ -2,7 +2,7 @@ class LegacyTasksController < ApplicationController
   before_action :verify_queue_access
   before_action :verify_task_assignment_access, only: [:create, :update]
 
-  ROLES = %w[judge attorney].freeze
+  ROLES = %w[judge attorney colocated].freeze
 
   def set_application
     RequestStore.store[:application] = "queue"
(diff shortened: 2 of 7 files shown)


app/controllers/appeals_controller.rb
```diff
@@ -21,6 +21,15 @@ def document_count
     render e.serialize_response
   end
 
+  def tasks
+    no_cache
+
+    tasks, _ = LegacyWorkQueue.tasks_with_appeals_of_appeal(params[:appeal_id])
```
aroltsch: Are you sending vacols ID or appeal ID? I think we should be sending appeal ID and then getting vacols id from the appeal object

app/controllers/legacy_tasks_controller.rb
```diff
@@ -16,6 +16,17 @@ def index
         render "queue/show"
       end
       format.json do
+        if params[:appeal_id]
```
aroltsch: Why did you add it here as well?

app/models/queues/legacy_work_queue.rb
```diff
@@ -7,17 +7,27 @@ class << self
...
+    end
 
-      tasks = vacols_tasks.zip(vacols_appeals).map do |task, appeal|
-        (role.capitalize + "LegacyTask").constantize.from_vacols(task, appeal, user)
-      end
-      [tasks, vacols_appeals]
+    def tasks_with_appeals_of_appeal(appeal_id)
+      vacols_tasks = repository.tasks_for_appeal(appeal_id)
+      # Look up user and their role
+      tasks_with_appeals_of_vacols_tasks(user, role, vacols_tasks)
```
aroltsch: You left a comment here probably forgot to implement. However, do we even need a user for this? Probably not. We need a role though so we know which class to instantiate.

app/controllers/appeals_controller.rb
aroltsch: If we don't need appeals, should we change the method to return tasks only? I would name it `tasks_by_appeal_id`

evankroske:
I decided to fix this issue another way: https://github.com/department-of-veterans-affairs/caseflow/pull/6245

app/controllers/appeals_controller.rb
evankroske: I'm sending a vacols ID because that's what /queue/appeals/X uses.

I made the method return tasks and appeals even though we don't need appeals because we have to create the appeal objects in order to get the tasks. Since we're creating the objects, I thought that we might as well return them. However, I could be convinced that we should change tasks_with_appeals_of_appeal to only return tasks. What do you think?

app/controllers/legacy_tasks_controller.rb
evankroske: Sorry, I forgot to remove this code. I'll get rid of it

evankroske: Done!

app/controllers/legacy_tasks_controller.rb
```diff
@@ -2,7 +2,7 @@ class LegacyTasksController < ApplicationController
   before_action :verify_queue_access
   before_action :verify_task_assignment_access, only: [:create, :update]
 
-  ROLES = %w[judge attorney].freeze
+  ROLES = %w[judge attorney colocated].freeze
```
aroltsch: Why did you add colocated here?

app/models/queues/legacy_work_queue.rb
```diff
@@ -7,17 +7,30 @@ class << self
 
     def tasks_with_appeals(user, role)
       vacols_tasks = repository.tasks_for_user(user.css_id)
-      vacols_appeals = repository.appeals_from_tasks(vacols_tasks)
+      tasks_with_appeals_of_vacols_tasks(user, role, vacols_tasks)
+    end
 
-      tasks = vacols_tasks.zip(vacols_appeals).map do |task, appeal|
-        (role.capitalize + "LegacyTask").constantize.from_vacols(task, appeal, user)
+    def tasks_with_appeals_of_appeal(appeal_id)
```
aroltsch: It might be me but I think "of" appeal is not what we usually use to name methods. I think "by" is more clear. However, I will leave it up to you.

app/models/queues/legacy_work_queue.rb
```diff
@@ -7,17 +7,30 @@ class << self
...
+      user = User.find_or_create_by(css_id: vacols_tasks[0].staff.sdomainid, station_id: User::BOARD_STATION_ID)
+      tasks_with_appeals_of_vacols_tasks(user, user.vacols_roles.first, vacols_tasks)
     end
 
     def repository
       return QueueRepository if FeatureToggle.enabled?(:test_facols)
       @repository ||= QueueRepository
     end
+  private
+    def tasks_with_appeals_of_vacols_tasks(user, role, vacols_tasks)
```
aroltsch: I would personally rename 
`tasks_with_appeals` => `tasks_with_appeals_by_user_id`
`tasks_with_appeals_of_appeal` => `tasks_with_appeals_by_appeal_id`
`tasks_with_appeals_of_vacols_tasks` => `tasks_with_appeals`

app/models/queues/legacy_work_queue.rb
```diff
@@ -7,17 +7,30 @@ class << self
...
 
-      tasks = vacols_tasks.zip(vacols_appeals).map do |task, appeal|
-        (role.capitalize + "LegacyTask").constantize.from_vacols(task, appeal, user)
+    def tasks_with_appeals_of_appeal(appeal_id)
+      vacols_tasks = repository.tasks_for_appeal(appeal_id)
+      if vacols_tasks.empty?
+        return [], []
       end
-      [tasks, vacols_appeals]
+      user = User.find_or_create_by(css_id: vacols_tasks[0].staff.sdomainid, station_id: User::BOARD_STATION_ID)
```
aroltsch: Hmmm. Not sure if this is 100% correct. So any user can query any task and then we assume this returned user associated with the task works at the board.

app/models/queues/legacy_work_queue.rb
```diff
@@ -7,17 +7,30 @@ class << self
...
+      tasks_with_appeals_of_vacols_tasks(user, user.vacols_roles.first, vacols_tasks)
```
aroltsch: User can have multiple roles, so we need to pass `role` as a parameter to specify which tasks to return.

app/models/queues/legacy_work_queue.rb
aroltsch: Thinking about this code more. I can see the complexity. Let's chat about this in person.

aroltsch: I think if you planning on reusing moss of the code from here we should just move the appeal_id as an optional parameter to GET /legacy_tasks?user_id=xxx&role=xxxx. Then if the `appeal_id` is present, filter out tasks all the way at the end to only return tasks associated with the appeal.

app/models/queues/legacy_work_queue.rb
evankroske: I renamed tasks_with_appeals_of_appeal, but I'd prefer not to rename the rest because tasks_with_appeals is used a few places.

app/models/queues/legacy_work_queue.rb
evankroske: I changed the endpoint to require a role parameter

PR #6241, opened 2018-07-16
Forking access check

mdbenjam:
Connects https://github.com/department-of-veterans-affairs/caseflow/issues/6243

### Description
Added a temporary fork to let us try out this new access control method.


app/services/external_api/bgs_service.rb
```diff
@@ -99,10 +99,10 @@ def can_access?(vbms_id)
     Rails.cache.fetch(cache_key, expires_in: 24.hours) do
       DBService.release_db_connections
 
-      MetricsService.record("BGS: can_access? (find_flashes): #{vbms_id}",
```
shanear: i suppose this string should also be toggled but i don't think it's that important

PR #6248, opened 2018-07-17
AssignWidget | Extracts selected tasks logic

evankroske:
Blocks https://github.com/department-of-veterans-affairs/caseflow/issues/5303

### Description
Decouples AssignWidget from the state to allow me to use it on the case details page.

### Testing Plan
- [x] Become BVARERDMAN.
- [x] Assign a task from one attorney to another.
- [x] Refresh the page to confirm that the task has been reassigned.


client/app/queue/components/AssignWidget.jsx
```diff
@@ -55,15 +58,9 @@ type Props = {|
...
-    return _.flatMap(
-      this.props.isTaskAssignedToUserSelected[this.props.previousAssigneeId] || {},
-      (selected, id) => (selected ? [this.props.tasks[id]] : []));
-  }
-
   handleButtonClick = () => {
-    const { selectedAssignee, selectedAssigneeSecondary } = this.props;
-    const selectedTasks = this.selectedTasks();
+    const { selectedAssignee, selectedAssigneeSecondary, getSelectedTasks } = this.props;
+    const selectedTasks = getSelectedTasks();
```
mdbenjam: why is this still a function seems like we can just pass in the values?

evankroske: I thought this would reduce the number of times that AssignWidget needs to be rendered, but now that I think about it, the complexity probably isn't worth it. I'll change AssignWidget to accept the tasks, instead.

Issue #6098, opened 2018-07-02
Intake | Set the claimant when establishing the end product


shanear:
We need to be able to set the claimaint on the EP we create in VBMS via the EstablishClaim service.

There is no officially supported way to do this right now. However, there is a unofficially supported way of passing the claimant's participant id where the veteran participant id would go. VBMS said they could add official support for this before the December release and potentially before the August release.

## Acceptance criteria
- Set the claimant on the EP we create for AMA HLRs and SCs using the unofficial method.
- Comment clearly in the code that this is a workaround

leahbannon:
If the claimant is different from the vet, we send the claimant in participant ID not the vet.

kahlouie:
From Sally: Can't set non-veteran claimant without a payee code.

abbyraskinUSDS:
@leikkisa would working on #6234 unblock this?

shanear:
<img width="878" alt="screen shot 2018-09-25 at 5 43 10 pm" src="https://user-images.githubusercontent.com/1596259/46045084-96d19d00-c0ea-11e8-9397-35b27cb47596.png">
Works!

PR #6258, opened 2018-07-17
Set claimant on End Product

leikkisa:
Connects #6098 

Sets the claimant on an End Product using the participantPersonId in VBMS, which is to be set to the participant ID of the claimant (either the Veteran, or of a person related to the Veteran).

Note: This is a workaround solution until there is VBMS supports a more official solution for setting the claimant, not using a field that is designed for the Veteran's participant ID.


leikkisa:
This is blocked because we have to set the payee code at the same time as the claimant, and we need more information on how to set the payee code.

PR #6263, opened 2018-07-18
Queue | Move state.queue.loadedQueue tasks/appeals into state.queue

### LINKED ISSUE #6053: Queue | Consolidate state.queue.tasks and state.queue.loadedQueue.tasks
When Queue started, the main reducer structure was
```javascript
initialState = {
  loadedQueue: {
    appeals: {},
    tasks: {}
  }
};
```

This was fine initially, when we only needed to track tasks/appeals assigned to the current user. However, since then, we've had to start tracking tasks for multiple assignees, and moved to storing tasks in `state.queue.tasks`, resulting in some [cases](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/AttorneyTaskTable.jsx#L80) [of](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/AttorneyTaskListView.jsx#L34) [checking](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/JudgeAssignTaskListView.jsx#L39) [both](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/JudgeReviewTaskListView.jsx#L43) [collections](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/JudgeReviewTaskTable.jsx#L50). Maintaining both collections is unnecessary and confusing, and they should be consolidated.

dannysellers:
Resolves #6053 

### Description
As Queue has grown, we've moved from tracking tasks/appeals for a single user, to multiple (as in judge assignment), moving from storing tasks/appeals inside `loadedQueue` (i.e. the user's loaded queue) to `state.queue.tasks`/`appeals`, with assignment determined by other attributes.

This eliminates the distinction between tasks/appeals stored in `loadedQueue` and not, moving them all to `state.queue`

To select tasks/appeals by user id, we use [selectors](https://github.com/department-of-veterans-affairs/caseflow/blob/9b918af7cbe318f58ab0214c7456cacb17aabbe1/client/app/queue/selectors.js) to get by `location_code` (LegacyAppeal [`bfcurloc`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/services/appeal_repository.rb#L164)) or [`user_id`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/legacy_tasks/legacy_task.rb#L18) (Task).

### Acceptance Criteria 
- [ ] Code compiles correctly, tests pass (should be no user-facing impact)


client/app/queue/selectors.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+import type { State } from './types/state';
+import _ from 'lodash';
+
+export const selectedTasksSelector = (state: State, userId: string) => _.flatMap(
```
dannysellers: Initially, I thought I was going to have to write a selector to get tasks/appeals assigned to the current user, but I'm not sure that we'll need that until we support acting judges switching roles—otherwise, contexts where the user views their own tasks and contexts where they manage others' are exclusive / assignment is determined elsewhere

client/app/queue/AttorneyTaskListView.jsx
```diff
@@ -41,7 +41,7 @@ class AttorneyTaskListView extends React.PureComponent {
 
   render = () => {
     const { messages } = this.props;
-    const noTasks = !_.size(this.props.loadedQueueTasks) && !_.size(this.props.appeals);
+    const noTasks = !_.size(this.props.tasks) && !_.size(this.props.appeals);
```
evankroske: I think that there may be tasks in state.queue.tasks that aren't assigned to the current attorney. I added state.queue.tasks as a cache of tasks, so whether a task in state.queue.tasks should only tell you whether it has been loaded. I'm following Redux's advice for normalizing entities in the state: https://redux.js.org/basics/reducers#designing-the-state-shape What do you think?

client/app/queue/JudgeReviewTaskTable.jsx
```diff
@@ -50,47 +51,27 @@ class JudgeReviewTaskTable extends React.PureComponent {
       diff(moment(task.attributes.assigned_on), 'days')
   }];
 
-  render = () => {
-    const { loadedQueueTasks, appeals, tasks } = this.props;
-    const taskWithId = {};
-
-    for (const id of Object.keys(loadedQueueTasks)) {
```
dannysellers: I'm not sure if there was a reason to compare all `tasks` with `loadedQueueTasks` here—is `loadedQueueTasks` used to determine which tasks are assigned to the user?

<details><summary>if we need</summary>

```javascript
export const tasksByAssigneeCssId = createSelector(
  [getTasks, getUserCssId],
  (tasks: Tasks, cssId: string) => _.keyBy(
    _.filter(tasks, (task) => task.attributes.user_id === cssId),
    (task: Task) => task.id
  )
);
```

</details>

client/app/queue/AttorneyTaskListView.jsx
evankroske: It seems like this file still depends on tasks having the same meaning as loadedQueueTasks. Am I missing something?

client/app/queue/AttorneyTaskTable.jsx
```diff
@@ -97,18 +97,13 @@ class AttorneyTaskTable extends React.PureComponent<Props> {
...
-      taskWithId[id] = tasks[id];
-    }
+    const { appeals, tasks } = this.props;
 
     return <Table
       columns={this.getQueueColumns}
       rowObjects={sortTasks({
         appeals,
-        tasks: taskWithId
+        tasks
```
evankroske: It seems like this code depends on the assumption that tasks only contains tasks assigned to the current user. I think that this could cause us problems in the future when we need to allow attorneys to assign tasks to colocated admins. If that works like judge assignment works today, tasks will contain tasks assigned to colocated admins as well as the current user. What do you think?

client/app/queue/types/models.js
```diff
@@ -14,10 +14,6 @@ export type User = {
 
 export type Judges = { [string]: User };
 
-export type DeprecatedTask = {
```
evankroske: 👏

client/app/queue/selectors.js
```diff
@@ -0,0 +1,52 @@
...
+  (tasks: Tasks) => _.keyBy(
+    _.filter(tasks, (task: Task) => task.attributes.task_type === 'Review'),
+    (task: Task) => task.id
+  )
+);
+
+export const appealsByAssigneeCssIdSelector = createSelector(
+  [getAppeals, getUserCssId],
+  (appeals: LegacyAppeals, cssId: string) => _.keyBy(
+    _.filter(appeals, (appeal: LegacyAppeal) => cssId.includes(appeal.attributes.location_code)),
```
evankroske: Just curious, but why can't you compare cssId and location_code for equality?

client/app/queue/AttorneyTaskTable.jsx
dannysellers: Oh, yeah, you're right, we need a selector for this prop too. Thanks, good catch.

Because they're called from within `mapStateToProps`, selectors can [receive props](https://github.com/reduxjs/reselect#accessing-react-props-in-selectors) and select with those instead of just looking through the state.

client/app/queue/selectors.js
dannysellers: `userCssId` is of the form `BVASCASPER1`, while `location_code`s are all just `SCASPER1`. It looks like it might could also [possibly be numbers](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/legacy_appeal.rb#L111).

We might be able to do `cssId.endsWith(appeal.attributes.location_code)`, but I'm not sure if that always holds

mdbenjam: I think ends_with is safer than includes. SCASPER is contained in SCASPER and SCASPER1.

evankroske: It sounds like location_code contains Staff.slogid. In the real data, I don't think that this relationship between Staff.sdomainid (CSS ID) and slogid holds. I think that we'll need to either add CSS ID to LegacyAppeal or use Task/CaseAssignment (which has the CSS ID of the assignee) to find the appeals assigned to the current user. https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/vacols/case_assignment.rb#L95

client/app/queue/BeaamTable.jsx
```diff
@@ -59,6 +59,6 @@ BeaamTable.propTypes = {
   featureToggles: PropTypes.object
 };
 
-const mapStateToProps = (state) => _.pick(state.queue.loadedQueue, 'appeals');
+const mapStateToProps = (state) => _.pick(state.queue, 'appeals');
```
evankroske: Do we need to filter these appeals? It seems like we should filter them by assignee

client/app/queue/CaseSnapshot.jsx
```diff
@@ -126,7 +127,7 @@ export class CaseSnapshot extends React.PureComponent {
           {this.taskAssignmentListItems()}
         </CaseDetailsDescriptionList>
       </div>
-      { this.props.loadedQueueAppealIds.includes(appeal.vacols_id) &&
+      { this.props.appealIds.includes(appeal.vacols_id) &&
```
evankroske: Optional: I feel like appealIds isn't a very descriptive name for this variable. appealsAssignedToCurrentUser might be better.

client/app/queue/JudgeAssignTaskListView.jsx
```diff
@@ -36,24 +39,12 @@ class JudgeAssignTaskListView extends React.PureComponent {
...
-      appeals
-    }).
-      filter((task) => task.attributes.task_type === 'Assign').
-      map((task) => ({
-        task,
-        appeal: this.props.appeals[task.appealId]
-      }));
-  }
+  unassignedTasksWithAppeals = () => sortTasks(
+    _.pick(this.props, 'tasks', 'appeals')
```
evankroske: _.pick isn't type-safe. This means that if, for example, you had misspelled "tasks" in the call above, Flow wouldn't tell you. Instead, I think that we should keep using destructuring here. We can keep existing references to _.pick, but I don't think that we should add it to code that isn't already using it. What do you think?

client/app/queue/QueueApp.jsx
```diff
@@ -208,14 +213,15 @@ QueueApp.propTypes = {
 
 const mapStateToProps = (state) => ({
   ..._.pick(state.caseSelect, 'caseSelectCriteria.searchQuery'),
-  ..._.pick(state.queue.loadedQueue, 'appeals'),
+  ..._.pick(state.queue, 'appeals'),
```
evankroske: This change looks really risky to me because QueueApp passes its props to the components it renders. I think that you should keep the current behavior of QueueApp by filtering appeals by assignee

client/app/queue/reducers.js
```diff
@@ -73,19 +61,15 @@ const workQueueReducer = (state = initialState, action = {}) => {
     });
   case ACTIONS.DELETE_APPEAL:
     return update(state, {
-      loadedQueue: {
-        appeals: { $unset: action.payload.appealId },
-        tasks: { $unset: action.payload.appealId }
-      }
+      appeals: { $unset: action.payload.appealId },
+      tasks: { $unset: action.payload.appealId }
```
evankroske: Optional: I think that this action handler should reassign the appeal and task to a nonsense CSS ID instead of deleting them. As you know, this action is created when a decision is signed. If we ever need to display cases that have been signed or implement undo, we'll need to keep these appeals and tasks in the state. That's why I didn't delete tasks in this action handler when I introduced state.queue.tasks. What do you think?

client/app/queue/reducers.js
```diff
@@ -342,13 +326,8 @@ const workQueueReducer = (state = initialState, action = {}) => {
...
       },
-      loadedQueue: {
-        tasks: {
-          $unset: [appealId]
-        },
-        appeals: {
-          $unset: [appealId]
-        }
+      appeals: {
+        $unset: [appealId]
```
evankroske: I don't think that we should delete the appeal here. We still need it to display the tasks and appeals of attorneys on the judge's team. Reassigning it (which should already be happening) should make it disappear from the unassigned cases page

client/app/queue/types/state.js
```diff
@@ -2,18 +2,15 @@
...
 } from './models';
 
-export type LoadedQueueTasks = { [string]: DeprecatedTask };
-export type LoadedQueueAppeals = { [string]: Object };
-
 export type TasksAndAppealsOfAttorney = {
   [string]: {
     state: string,
-    data: {tasks: LoadedQueueTasks, appeals: LoadedQueueAppeals},
+    data: {tasks: Tasks, appeals: LegacyAppeals},
```
evankroske: I don't think that we should change the type of data.tasks. I'm trying to refactor the state to replace tasks with references, and I introduced DeprecatedTask to keep people from using any fields besides ID of tasks in tasksAndAppealsOfAttorney. What do you think?

client/app/queue/types/state.js
```diff
@@ -44,7 +41,10 @@ export type UiState = {
   },
   featureToggles: Object,
   selectedAssignee: ?string,
-  selectedAssigneeSecondary: ?string
+  selectedAssigneeSecondary: ?string,
+  loadedUserId: ?string,
```
evankroske: I'm pretty sure that loadedUserId is a number

client/app/queue/QueueApp.jsx
dannysellers: That's fair, and is definitely something we should move away from. It actually looks like `appeals` isn't used at all within QueueApp 👍

client/app/queue/reducers.js
dannysellers: Yeah, this action was not easy for me to understand when refactoring—we previously were deleting a task and appeal from `loadedQueue`, which I guess was effectively just re-assignment to a user who isn't loaded / isn't the primary user? Thanks for pointing this out

client/app/queue/reducers.js
dannysellers: I'm not opposed, though I might defer that to a future ticket. For one thing, I don't think we need to track 'deleted' appeals/tasks just yet—within Attorney checkout, for example, the success message says something like "if you need to make changes, do so in DAS".

Also, I suspect most actions we're doing / that could be undone would involve more than we could currently do on the front end (whether because there're models/attributes that aren't sent to the FE, or endpoints that don't exist, etc). At a certain point, it might just be simpler to have the server undo whatever action, and then reload the FE collection of tasks/appeals to reflect the 'new' state.

Definitely something to keep in mind though 👍

client/app/queue/types/state.js
dannysellers: Honestly, I would not have figured that out just from reading the code... I was under the impression that `LoadedQueueTasks`/`Appeals` were just to indicate that the components that used those hadn't been converted to the new tasks/appeals structure yet.

I'm happy to work towards fetching tasks from `state.queue.tasks` using just IDs, but if that's the case, we should be explicit about it, both in terms of discussing a strategy (like using/normalizing selectors), and in terms of naming variables.

For example, there's no indication to me that `DeprecatedTask` would be a new scheme of accessing tasks, or that it would be fundamentally different from `Task`, rather than just having a different set of attributes but behaving similarly. Also, `tasksAndAppealsOfAttorney` feels a bit misleading if we're storing IDs, load state, and error messages within it.

Is there anything you think should be done for this PR to address this? I'm happy to make a new ticket too.

client/app/queue/QueueApp.jsx
```diff
@@ -208,14 +213,14 @@ QueueApp.propTypes = {
 
 const mapStateToProps = (state) => ({
   ..._.pick(state.caseSelect, 'caseSelectCriteria.searchQuery'),
-  ..._.pick(state.queue.loadedQueue, 'appeals'),
```
evankroske: Are you sure that none of the components that receive the props of QueueApp depend on appeals?

client/app/queue/selectors.js
```diff
@@ -0,0 +1,56 @@
...
+  LegacyAppeal,
+  LegacyAppeals
+} from './types/models';
+
+export const selectedTasksSelector = (state: State, userId: string) => _.flatMap(
+  state.queue.isTaskAssignedToUserSelected[userId] || {},
+  (selected, id) => selected ? [state.queue.tasks[id]] : []
+);
+
+const getTasks = (state: State): Tasks => state.queue.tasks;
```
evankroske: Optional: I don't think that you need to annotate the return values of these functions. Flow should be inferring these types, so the annotations seem like duplicated code.

client/app/queue/BeaamTable.jsx
dannysellers: @mdbenjam I had to remove the selector I'd added here because these appeals don't have Tasks (using [`appealsByAssignedTaskSelector `](https://github.com/department-of-veterans-affairs/caseflow/blob/8da9f1065d83fb02e12edb4a7ddc58b3488c9ac0/client/app/queue/selectors.js#L46)). Should we try to select appeals based on their `location_code`?

mdbenjam: This is a temporary page, can we just not touch it? They don't have tasks, and they're new appeals so they won't have location codes.

Issue #6254, opened 2018-07-17
Create prototype of authentication with IDT  


amprokop:
Flow:

1. User launches IDT.
1. If there is a token saved for the user, IDT attempts to use it to make the neccessary API call.
1. If there is no token, or if a 403 status code is returned by the API (token expired), IDT makes a GET request to caseflow.ds.va.gov/idt/api/v1/token
1. Caseflow returns a proposed token and a one-time key.
1. IDT opens a browser window with the URL caseflow.ds.va.gov/idt/auth?key=#{ONE_TIME_KEY}
1. If the user is not logged in, Caseflow's existing logic will redirect the user to the CSS login page, and redirect back to caseflow.ds.va.gov/idt/auth?key=#{ONE_TIME_KEY} after login.
1. After the user is logged in, Caseflow will mark the token as valid and display a message like "You have successfully authenticated with the IDT. You can now return to the IDT"
1. In the meantime, every 10 seconds, the IDT will make a new call with the proposed token. If the token is not yet validated, Caseflow will return a 401. Otherwise, Caseflow will return a 200.
1. The user can use the IDT without having to log in again until the token is no longer valid.

PR #6264, opened 2018-07-18
Creates IDT auth framework

amprokop:
Connects #6254

### Description
Adds a prototype of authentication functionality for the IDT to make authenticated API requests to Caseflow. No DB data is returned at this point, so we don't need to feature flag anything.

### Test plan
1. `curl localhost:3000/idt/api/v1/token`
1. Verify response is like `{ one_time_key: "dfsafjslakfs", token: "fdsakfjsa"}`
1. `curl localhost:3000/idt/api/v1/appeals`
1. Verify 400 is returned
1. `curl localhost:3000/idt/api/v1/appeals -H "token: fdsakfjsa"`
1. Verify 403 is returned
1. Open `localhost:3000/idt/auth?one_time_key=dfsafjslakfs`
1. Verify success message is shown
1. Open `localhost:3000/idt/auth?one_time_key=dfsafjslakfs`
1. Verify error is shown
1. `curl  -H "TOKEN: fdsakfjsa" localhost:3000/idt/api/v1/appeals`
1. Verify 200 is returned


app/models/idt/token.rb
```diff
@@ -0,0 +1,47 @@
+class Idt::Token
+  VALID_TOKENS_KEY = "valid_tokens_key".freeze
+  ONE_TIME_KEYS_KEY = "one_time_keys_key".freeze
+  TOKEN_VALIDITY_IN_SECONDS = 60 * 60 * 24 * 3
+
+  def self.generate_proposed_token_and_one_time_key
+    one_time_key = SecureRandom.uuid.delete("-")
+    token = SecureRandom.uuid.delete("-")
```
mdbenjam: Love the creativity... but: https://ruby-doc.org/stdlib-1.9.3/libdoc/securerandom/rdoc/SecureRandom.html#method-c-hex

Also maybe should make it longer like 64 digits?

app/models/idt/token.rb
```diff
@@ -0,0 +1,47 @@
...
+    # to activate the token.
+    client.set(ONE_TIME_KEYS_KEY + one_time_key, token)
+
+    [one_time_key, token]
+  end
+
+  def self.activate_proposed_token(one_time_key)
+    token = client.get(ONE_TIME_KEYS_KEY + one_time_key)
+
+    raise Caseflow::Error::InvalidOneTimeKey unless token && token.length == 32
```
mdbenjam: Can abstract the 32 to a variable if we use the above method.

PR #6266, opened 2018-07-18
Remove queue_phase_two and judge_queue feature toggles because they w…

aroltsch:
Remove two feature toggles:
* queue_phase_two
* judge_queue


client/app/queue/AttorneyTaskTable.jsx
```diff
@@ -38,7 +38,7 @@ class AttorneyTaskTable extends React.PureComponent<Props> {
   tableStyle = css({
```
dannysellers: I think we should actually just remove `tableStyle` completely

<details><summary>with 25% width</summary>

</details>

<details><summary>without specified width</summary>

</details>

In which case, the `bodyStyling` prop should also be removed from the Table below in `render`:
```javascript
return <Table
  ...
  rowClassNames={(task) => task.attributes.task_id ? null : 'usa-input-error'} />;
```

PR #6267, opened 2018-07-18
Queue | Adds endpoint to get tasks by VACOLS ID

evankroske:
Part of https://github.com/department-of-veterans-affairs/caseflow/issues/6222

### Description

### Testing Plan
- [x] Become BVAAABSHIRE.
- [x] Confirm that http://localhost:3000/appeals/3449656/tasks?role=judge returned a JSON task.
- [x] Confirm that http://localhost:3000/appeals/3449656/tasks?role=attorney returned a JSON task.
- [x] Become BVARERDMAN.
- [x] Confirm that http://localhost:3000/appeals/3449656/tasks?role=judge returned a JSON task.
- [x] Become BVASCASPER1.
- [x] Confirm that http://localhost:3000/appeals/3449656/tasks?role=attorney returned a JSON task.


app/models/queues/legacy_work_queue.rb
```diff
@@ -7,17 +7,38 @@ class << self
...
       return QueueRepository if FeatureToggle.enabled?(:test_facols)
       @repository ||= QueueRepository
     end
+
+    private
+
+    MODEL_CLASS_OF_ROLE = {
+      "Attorney" => AttorneyLegacyTask,
+      "Judge" => JudgeLegacyTask,
+      "Colocated" => LegacyTask
```
aroltsch: Colocated don't have LegacyTask, the only have Caseflow tasks

app/models/queues/legacy_work_queue.rb
```diff
@@ -7,17 +7,38 @@ class << self
...
+    }.freeze
+
+    def tasks_with_appeals_of_vacols_tasks(user, role, vacols_tasks)
+      vacols_appeals = repository.appeals_from_tasks(vacols_tasks)
+
+      tasks = vacols_tasks.zip(vacols_appeals).map do |task, appeal|
+        MODEL_CLASS_OF_ROLE[role.capitalize].from_vacols(task, appeal, user)
```
aroltsch: Do you need to call capitalize if you declare the above hash with lower case keys `MODEL_CLASS_OF_ROLE`?

app/models/queues/legacy_work_queue.rb
```diff
@@ -7,17 +7,38 @@ class << self
...
 
-      tasks = vacols_tasks.zip(vacols_appeals).map do |task, appeal|
-        (role.capitalize + "LegacyTask").constantize.from_vacols(task, appeal, user)
+    def tasks_with_appeals_by_appeal_id(appeal_id, role)
+      vacols_tasks = repository.tasks_for_appeal(appeal_id)
+      if vacols_tasks.empty?
+        return [], []
       end
-      [tasks, vacols_appeals]
+      user = User.find_or_create_by(css_id: vacols_tasks[0].staff.sdomainid, station_id: User::BOARD_STATION_ID)
```
aroltsch: So as I mentioned before this will create additional query. You can return `sdomainid` for assigned_to in CaseAssignment select_tasks (`s2.sdomainid as assigned_to_css_id`)

app/models/vacols/case_assignment.rb
```diff
@@ -65,11 +65,19 @@ def tasks_for_user(css_id)
...
+      select_tasks.where("brieff.bfkey = #{id}")
+    end
+
+    # rubocop:disable Metrics/MethodLength
     def select_tasks
       select("brieff.bfkey as vacols_id",
              "brieff.bfcorlid as vbms_id",
              "brieff.bfd19 as docket_date",
              "brieff.bfdloout as assigned_to_location_date",
+             "brieff.bfcurloc as bfcurloc",
```
aroltsch: You can remove this if you decide to add `s2.sdomainid as assigned_to_css_id`

app/models/queues/legacy_work_queue.rb
```diff
@@ -7,17 +7,38 @@ class << self
...
     end
 
     def repository
       return QueueRepository if FeatureToggle.enabled?(:test_facols)
       @repository ||= QueueRepository
     end
+
+    private
+
+    MODEL_CLASS_OF_ROLE = {
```
aroltsch: I like this change!

app/controllers/legacy_tasks_controller.rb
```diff
@@ -1,8 +1,10 @@
 class LegacyTasksController < ApplicationController
+  include Errors
```
aroltsch: You can even put that in ApplicationController. I really like the idea of creating a concern for shared errors!

app/controllers/appeals_controller.rb
```diff
@@ -1,7 +1,11 @@
 class AppealsController < ApplicationController
+  include Errors
+
   before_action :react_routed
   before_action :set_application, only: [:document_count, :new_documents]
 
+  ROLES = Constants::ROLES["roles"]
```
aroltsch: So colocated should not be allowed here for now.

app/models/queues/legacy_work_queue.rb
evankroske: The current code supports uppercase and lowercase roles, and some tests pass in capitalized role names. I think it would be pretty difficult to remove support for capitalized role names.

app/models/queues/legacy_work_queue.rb
evankroske: Sorry, I forgot that we talked about this. I'll fix it

evankroske: Done!

app/controllers/appeals_controller.rb
evankroske: Done!

app/models/queues/legacy_work_queue.rb
evankroske: Good point! I removed this code

app/models/vacols/case_assignment.rb
evankroske: Done!

app/controllers/legacy_tasks_controller.rb
evankroske: We discussed this offline. I don't think we should put it in ApplicationController because it's only useful in a few controllers

Issue #6277, opened 2018-07-19
Make Taks's assigned_to field a polymorphic association with User/Organization


mdbenjam:
Tasks need to be assignable to an organization. Specifically we need to be able to assign to VSOs.

Issue #6276, opened 2018-07-19
Build organization table


mdbenjam:
We want an organization table with the following fields:

```
- name
- role
- functions
- url
- type
```

Example:

```
name: American Legion
role: "VSO"
functions: vso-pilot
url: american-legion
type: vso
```

PR #6283, opened 2018-07-19
Adding organization

mdbenjam:
Connects #6276 #6277 

### Description
Added a new organization table that has a polymorphic association with the task table. This means we can assign tasks to an organization instead of just a user.

### Testing Plan
1. Create a task (Had to create a temporary generic task class for our type)
1. Create an organization
1. Ensure we can associate the organization with the task


db/migrate/20180719153039_create_organizations_table.rb
```diff
@@ -0,0 +1,12 @@
+class CreateOrganizationsTable < ActiveRecord::Migration[5.1]
+  def change
+    create_table :organizations do |t|
+      t.string :type
+      t.string :name
+      t.string :role
+      t.string :function
+      t.string :url
+      t.string :participant_id
```
mdbenjam: I'm not sure about this, this is pretty particular to VSOs, but I figure we can always refactor if this turns out to be unnecessary for generic organizations.

We want participant ids so we can see if this vso matches the claimants rep in BGS.

Issue #6280, opened 2018-07-19
Create VSO as subclass of organization


mdbenjam:
We need to create a sub organization called VSO that corresponds to the vso type on the organization model.

This class will handle the authentication piece of whether a user is able to view a given queue. Steps in authentication:

1) Ensure the user has the VSO role
1) Call the `findPtcpntId` endpoint in the security service to get the user's participant id
1) Call `findPOAsByPtcpntId` in the org service to get the VSOs the user is a part of
1) Ensure the name of the POA returned matches the name of the POA in the table.

Issue #6278, opened 2018-07-19
Create organization routes


mdbenjam:
We need a new route so user's can see the tasks associated with an organization:

`/tasks/organization/:id`

We also a new url that a user can go to, to see the tasks associated to the organization:

`/queue/organization/:id`

PR #6291, opened 2018-07-19
Adding VSO as an organization and a new organizational tasks controller

mdbenjam:
Connects #6280 #6278

### Description
This change adds a new VSO organization subclass that defines a user can access method which determines if a user belongs to the given VSO.

This PR also defines a new organizational/tasks controller that returns all the tasks associated with a given organization.

### Testing Plan
1. The new automated tests!


app/mappers/power_of_attorney_mapper.rb
```diff
@@ -8,14 +8,14 @@ def self.included(base)
   end
 
   def get_poa_from_bgs_poa(bgs_rep = {})
-    return {} unless bgs_rep[:power_of_attorney]
+    return {} unless bgs_rep
```
mdbenjam: Have callers pass in what they want mapped, this allows us to use it for our new `fetch_poas_by_participant_id` method

app/services/external_api/bgs_service.rb
```diff
@@ -60,17 +60,32 @@ def fetch_poa_by_file_number(file_number)
...
 
     unless @poas[file_number]
-      bgs_poa = MetricsService.record("BGS: fetch veteran info for file number: #{file_number}",
+      bgs_poa = MetricsService.record("BGS: fetch poa for file number: #{file_number}",
                                       service: :bgs,
                                       name: "org.find_poas_by_file_number") do
         client.org.find_poas_by_file_number(file_number)
       end
-      @poas[file_number] = get_poa_from_bgs_poa(bgs_poa)
+      @poas[file_number] = get_poa_from_bgs_poa(bgs_poa[:power_of_attorney])
```
mdbenjam: Updating based on our changes to the mapping method.

db/seeds.rb
```diff
@@ -188,6 +188,8 @@ def seed
     create_tags
     create_beaam_appeals
 
+    User.create(css_id: "VSO", station_id: 101)
```
mdbenjam: I'm going to change how we add our VACOLS users cause running this seeds file overwrites them.

app/controllers/organization_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    redirect_to "/unauthorized" unless FeatureToggle.enabled?(organization.feature.to_sym, user: current_user)
+  end
+
+  def set_application
+    RequestStore.store[:application] = "queue"
+  end
+
+  def organization_url
+    params[:id]
```
aroltsch: I am confused why you would not be sending a DB ID instead? Since it appears organization would already be saved at our DB at this point.

app/controllers/organization_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  def verify_role_access
+    verify_authorized_roles(organization.role)
+  end
+
+  def verify_feature_access
+    return unless organization.feature
+
+    redirect_to "/unauthorized" unless FeatureToggle.enabled?(organization.feature.to_sym, user: current_user)
```
aroltsch: You would be storing feature names at the org table? Curious why you decided to do that? Feature toggles are something temporary but you will be persisting them?

app/controllers/organization_controller.rb
mdbenjam: I want this to be a human readable path so that users can go to `queue/organization/american-legion` rather than `queue/organization/3`

aroltsch: Then perhaps renaming the route `GET ogranizations/:url` would be more clear?

```
resources :organizations, param: :url
```

config/routes.rb
```diff
@@ -164,6 +164,11 @@
 
   resources :legacy_tasks, only: [:create, :update]
   resources :tasks, only: [:index, :create, :update]
+
+  resources :organization, only: [:show] do
```
aroltsch: I think we should make it plural to stay consistent.

app/models/tasks/generic.rb
```diff
@@ -0,0 +1,2 @@
+class Generic < Task
```
aroltsch: Is it used anywhere? Then perhaps removing it for now

app/controllers/organization_controller.rb
aroltsch: Can you use organization.name?

app/models/tasks/generic.rb
mdbenjam: It's used in tests for now. I can name it VsoTask for now.

app/controllers/organization_controller.rb
mdbenjam: The only problem there, is I want to by default not use a feature toggle and only test it if it's defined. However, the name will always be defined so we'll have to always look for a feature toggle. I'm going to leave it as is for now. But will remove it if we don't find it immediately useful.

app/controllers/organizations_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  def verify_role_access
+    verify_authorized_roles(organization.role)
+  end
+
+  def verify_feature_access
+    return unless organization.feature
+
+    redirect_to "/unauthorized" unless FeatureToggle.enabled?(organization.feature.to_sym, user: current_user)
```
lowellrex: This is a super smart way to handle the issue of organization-specific feature toggles. I dig it!

Issue #6296, opened 2018-07-20
Refactor queue frontend to make appeal first class citizen


mdbenjam:
Our frontend treats tasks as the primary object instead of appeal. I believe using the appeal as the primary object is more intuitive and leads to simpler code. We can associate tasks to be an array on appeal. We can leave our state denormalized in redux i.e. tasks and appeals are separate objects, but use a selector to combine them in this way. This selector will also deal with legacy/ama appeals, and legacy/ama tasks outputting objects in a common form.

PR #6322, opened 2018-07-23
Refactoring tasks and appeals

mdbenjam:
Connects #6296

### Description
This PR makes a few (okay a lot of) changes.

1) Creates a new selector to get appeals, and put associated tasks as a child of the appeal object.
1) Creates a new TaskTable component that is used by all of our table views.
1) Pulls sorting into the TaskTable component so that it is available everywhere.
1) Removes unnecessary state, and continues the pattern of adding appeals/tasks to the first level state objects.

Some other notes.
1) We change how we do type sort to make it a little simpler.
1) We do not use reselect for some of our selectors. I tried to, but it got complicated (i.e. we needed deep comparison) and I don't actually think we need memoization for performance of our app at this point. We can always rewrite them in the future.
1) The `requireDasRecord` boolean on the TaskTable is pretty specific to the attorney table. However, we should be able to remove this code soon, once we've moved off of legacy tasks.
1) In a lot of places we reference `appeal.tasks[0]`. This is ugly, and not ideal, however since we originally believed that tasks and appeals had a one to one mapping it was necessary. Moving forward we'll need to support multiple tasks.

### Testing Plan
1. Automated tests still pass.


client/app/components/Table.jsx
```diff
@@ -54,7 +54,7 @@ const HeaderRow = (props) => {
         let columnContent = <span {...tooltipProps}>{column.header || ''}</span>;
 
         if (column.getSortValue) {
-          const sortIndicator = props.sortAscending ? <SortArrowDown /> : <SortArrowUp />;
+          const sortIndicator = props.sortAscending ? <SortArrowUp /> : <SortArrowDown />;
```
mdbenjam: Flipping the default arrows, but passing 'asc' and 'desc' correctly to the `_.orderBy` function below.

client/app/queue/components/TaskTable.jsx
```diff
@@ -0,0 +1,320 @@
...
+      header: COPY.CASE_LIST_TABLE_APPEAL_TYPE_COLUMN_TITLE,
+      valueFunction: (appeal) => this.appealHasDASRecord(appeal) ?
+        renderAppealType(appeal) :
+        <span {...redText}>{COPY.ATTORNEY_QUEUE_TABLE_TASK_NEEDS_ASSIGNMENT_ERROR_MESSAGE}</span>,
+      span: (appeal) => this.appealHasDASRecord(appeal) ? 1 : 5,
+      getSortValue: (appeal) => {
+        // We append a * before the docket number if it's a priority case since * comes before
+        // numbers in sort order, this forces these cases to the top of the sort.
+        if (appeal.attributes.aod || appeal.attributes.type === 'Court Remand') {
+          return '*' + appeal.docket_number;
```
mdbenjam: I think this is a cleaner way of doing the sort.

client/app/queue/components/TaskTable.jsx
```diff
@@ -0,0 +1,320 @@
...
+        if (!task) {
+          return 0;
+        }
+
+        return moment().diff(moment(task.attributes.assigned_on), 'days');
+      }
+    } : null;
+  }
+
+  caseDaysWaitingColumn = () => {
```
mdbenjam: We have duedate and days waiting, I've been told that eventually we'll just have days waiting.

client/app/queue/components/TaskTable.jsx
```diff
@@ -0,0 +1,320 @@
...
+
+  caseTypeColumn = () => {
+    return this.props.includeType ? {
+      header: COPY.CASE_LIST_TABLE_APPEAL_TYPE_COLUMN_TITLE,
+      valueFunction: (appeal) => this.appealHasDASRecord(appeal) ?
+        renderAppealType(appeal) :
+        <span {...redText}>{COPY.ATTORNEY_QUEUE_TABLE_TASK_NEEDS_ASSIGNMENT_ERROR_MESSAGE}</span>,
+      span: (appeal) => this.appealHasDASRecord(appeal) ? 1 : 5,
+      getSortValue: (appeal) => {
+        // We append a * before the docket number if it's a priority case since * comes before
```
dannysellers: 👍

client/app/queue/reducers.js
```diff
@@ -36,7 +36,7 @@ export const initialState = {
     }
   },
   attorneysOfJudge: [],
-  tasksAndAppealsOfAttorney: {},
+  attorneyAppealsLoadingState: {},
```
dannysellers: 👍

client/app/queue/JudgeReviewTaskListView.jsx
```diff
@@ -65,23 +73,18 @@ class JudgeReviewTaskListView extends React.PureComponent {
 }
 
 JudgeReviewTaskListView.propTypes = {
-  tasks: PropTypes.object.isRequired,
   appeals: PropTypes.object.isRequired
```
dannysellers: `Warning: Failed prop type: Invalid prop 'appeals' of type 'array' supplied to 'JudgeReviewTaskListView', expected 'object'.`

client/app/queue/AttorneyTaskListView.jsx
```diff
@@ -73,7 +82,7 @@ class AttorneyTaskListView extends React.PureComponent {
 
 AttorneyTaskListView.propTypes = {
   tasks: PropTypes.object.isRequired,
-  appeals: PropTypes.object.isRequired
+  appeals: PropTypes.array.isRequired
```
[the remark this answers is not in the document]
mdbenjam: I'm going to leave it for the time being since we aren't using flow in here yet.

client/app/queue/JudgeAssignTaskListView.jsx
```diff
@@ -72,12 +59,14 @@ class JudgeAssignTaskListView extends React.PureComponent {
   }
 
   caseCountOfAttorney = (attorneyId) => {
-    const { tasksAndAppealsOfAttorney } = this.props;
+    const { attorneyAppealsLoadingState, appealsByUserId } = this.props;
+
+    if (attorneyId in attorneyAppealsLoadingState &&
+      attorneyAppealsLoadingState[attorneyId].state === 'LOADED') {
+      return appealsByUserId[attorneyId] ? appealsByUserId[attorneyId].length.toString() : 0;
```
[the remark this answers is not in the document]
mdbenjam: I think it should be zero, because this means we've loaded the appeals, but couldn't find any. Agree with the toString not being necessary.

Issue #6200, opened 2018-07-12
VACOLS Integration | Vacols Upload


sharonwarner:
- Create a route in the schedule_period controller called `finalize` that takes the cache attribute `algorithm_output` and sends the results to VACOLS

PR #6325, opened 2018-07-24
Persist derived hearing schedule for schedule period

oscarAtNava:
Connects #6200 

### Description
Create a route in the schedule_period controller for both judges and ro/co called finalize that
takes the cache attribute algorithm_output, sends the results to VACOLS and sets 'finalized' to true.

### Acceptance Criteria 
- [ ] Code compiles correctly
- [ ] Tests pass successfully in CI

### Testing Plan
TBD


app/controllers/hearings/schedule_periods_controller.rb
```diff
@@ -51,6 +47,18 @@ def create
     render json: { id: schedule_period.id }
   end
 
+  def update
+    schedule_period = SchedulePeriod.find(params[:schedule_period_id])
+    schedule_period.finalized = if schedule_period.type == "RoSchedulePeriod"
```
sharonwarner: I think this would be cleaner if we moved this logic to ro_schedule_period.rb and judge_schedule_period.rb. So in the controller, it would be something like `schedule_period.finalized = schedule_period.finalize_algorithm`, and then in each model, we would have a `finalize_algorithm` method that actually sends the information to VACOLS.

spec/controllers/hearings/schedule_periods_controller_spec.rb
```diff
@@ -42,4 +42,37 @@
       expect(response_body["id"]).to eq id
     end
   end
+
+  context "persist full schedule for a schedule period", focus: true do
```
sharonwarner: Nit: focus: true! 😂

spec/controllers/hearings/schedule_periods_controller_spec.rb
```diff
@@ -42,4 +42,37 @@
...
+        schedule_period: {
+          type: "RoSchedulePeriod",
+          start_date: "2015/10/24",
+          end_date: "2016/10/24"
+        },
+        file: base64_header + Base64.encode64(File.open("spec/support/validRoSpreadsheet.xlsx").read)
+      }
+      response_body = JSON.parse(response.body)
+      response_body["id"]
+    end
```
sharonwarner: I think we can just use the factories that we defined at the top of the file instead of creating a new schedule_period.

Issue #6341, opened 2018-07-25
API | Tasks are returned with the wrong assignee


evankroske:
When tasks are created or reassigned using LegacyTasksController, the updated task object is returned with the wrong assignee. The assignee is incorrectly set to the assigner. However, this problem only affects the frontend. The task is assigned correctly in the DBs.

PR #6342, opened 2018-07-25
Queue API | Fixes the assignee of created and updated tasks

evankroske:
Fixes https://github.com/department-of-veterans-affairs/caseflow/issues/6341

### Description
Sets the assignee fields to the right assignee.


mdbenjam:
Ha I noticed this in my PR as well, thanks for the fix!

PR #6359, opened 2018-07-25
update pointer to webpack for local tests

nolawi:
Updates webpack bundle location to match local build


nolawi:
Local spec tests are passing for me am not sure what else could be missing. Marking this as [WIP] for now

client/package.json
```diff
@@ -16,7 +16,7 @@
     "build:test": "NODE_ENV=test yarn run build",
     "build:demo": "NODE_ENV=development yarn run build",
     "build:production": "NODE_ENV=production yarn run build",
-    "dev:clear-build": "rm ../app/assets/webpack/* || true",
+    "dev:clear-build": "rm ../app/assets/javascripts/* || true",
```
mdbenjam: hmmm maybe outputting this to the javascripts folder is a bad idea. I think we might still have a stats.js file in there. 🤔

spec/rails_helper.rb
```diff
@@ -231,7 +231,7 @@ def read_csv(klass, date_shift)
   # If it does, it will not execute ReactOnRails, since that slows down tests
   # Thus this will only run once (to initially compile assets) and not on
   # subsequent test runs
-  if Dir["#{::Rails.root}/app/assets/webpack/*"].empty?
+  if File.exist?("#{::Rails.root}/app/assets/javascripts/webpack-bundle.js")
```
mdbenjam: This should be if it doesn't exist, not if it does exist.

nolawi: duh that makes sense

client/package.json
nolawi: i see, we can just remove `bundle` and sourcemaps instead of the whole folder. LMK if i should do that

mdbenjam: Hmmm yeah, I shouldn't have put generated js here as well as non-generated js...  Yeah it probably should just look for things with webpack in the name. Hopefully we can remove the stats.js file in the future.

Issue #6344, opened 2018-07-25
VACOLS Integration | Validations


sharonwarner:
- We should not allow `update` to run if `finalized == true`
- We should not allow `update` to run if `created_at > 4.days.ago`

PR #6366, opened 2018-07-25
Added validations to not update unless finalized == false and cached …

oscarAtNava:
…attribute was created four days or less.

Connects #6344

### Description
Cached data is not persisted to VACOLS unless finalized property is false and cached attribute was created 4 days or less ago.

### Acceptance Criteria 
- [ ] Code compiles correctly

### Testing Plan
Best way to test is using the back page in browser, then check the number of rows in the DB. It should not increment for a finalized schedule period.


app/controllers/hearings/schedule_periods_controller.rb
```diff
@@ -49,11 +49,18 @@ def create
...
+      schedule_period.schedule_confirmed(schedule_period.ro_hearing_day_allocations)
+    end
     render json: { id: schedule_period.id }
   end
 
   def schedule_period_params
     params.require(:schedule_period).permit(:type, :file, :start_date, :end_date)
   end
+
+  def valid_schedule_period(schedule_period)
```
sharonwarner: I just added a bunch of validate methods. I think this name would be clearer if it were something like `can_be_finalized?`. Also, I think it's better to have this method in the model, but up to you.

Issue #6360, opened 2018-07-25
IDT API should return a list of VACOLS appeals by assigned attorney or by file number


amprokop:
Create routes to return simple appeal info for the IDT

`GET /idt/api/v1/appeals --header "TOKEN: 254tretrewrew"` 
`GET /idt/api/v1/appeals --header "FILE: 22222222" --header "TOKEN: 254tretrewrew"`

# Sample output

```
{
  "data": [
    {
      "id": "2556987",
      "type": "legacy_appeals",
      "attributes": {
        "veteran_first_name": "Jay",
        "veteran_middle_name": "Q",
        "veteran_last_name": "Lemke",
        "file_number": "750998600",
        "docket_number": "2053383",
        "number_of_issues": 1
      }
    },
    {
      "id": "3019694",
      "type": "legacy_appeals",
      "attributes": {
        "veteran_first_name": "Tatum",
        "veteran_middle_name": "C",
        "veteran_last_name": "Schumm",
        "file_number": "849386058",
        "docket_number": "7592395",
        "number_of_issues": 1
      }
    }
  ]
}
```

# AC
- IDT endpoint returns appeals with veteran name, claim number, # of issues, docket_number, id.
- The user must be an attorney or a judge.
- A valid token is required.
- If no file number is passed in the headers, it will return appeals assigned to the user.
- If a file number is passed in the headers, it will return appeals with that file number.
- Only return appeals in ACT status.

laurjpeterson:
This is done.

PR #6373, opened 2018-07-26
adds real appeal data to endpoint

amprokop:
Connects https://github.com/department-of-veterans-affairs/caseflow/issues/6361
Connects https://github.com/department-of-veterans-affairs/caseflow/issues/6360

# Test plan
1. `curl localhost:3000/idt/api/v1/token`
1. note one time key
1. open `localhost:3000/idt/auth?one_time_key=KEY` in the browser while logged in as attorney user
1. observe key validation succeeding
1. `curl -H "TOKEN: token" localhost:3000/idt/api/v1/appeals`
1. Verify that cases returned match cases in the attorney's queue, if all those cases are activated
1. `curl -H "FILE: some_file_number" -H "TOKEN: token" localhost:3000/idt/api/v1/appeals`
1. Verify that the cases returned match those returned by Case Search in the UI, if all those cases are activated


app/models/legacy_appeal.rb
```diff
@@ -385,6 +385,11 @@ def dispatch_decision_type
     return "Remand" if remand_on_dispatch?
   end
 
+  def activated?
+    # An appeal will be in this status if the board is currently working on it.
```
mdbenjam: Any chance you could add a comment to active? to explain what that one means. Having active? and activated? is kind of confusing.

app/models/serializers/idt/v1/appeal_serializer.rb
```diff
@@ -0,0 +1,19 @@
+class Idt::V1::AppealSerializer < ActiveModel::Serializer
```
mdbenjam: Can we not use our existing serializers?

app/models/legacy_appeal.rb
amprokop: Cool, will do.

app/models/serializers/idt/v1/appeal_serializer.rb
amprokop: In my eyes, serializers should be pretty tightly coupled to the endpoint. If we wanted to change what the server returns for the Queue list view, we shouldn't need to have the IDT make changes. 

Also, we want to return the minimal amount of information here so we won't require any BGS calls for either AMA or VACOLS appeals.

mdbenjam: Just FYI I found it difficult to figure out how to mix legacy and ama appeals with different serializers without adding the `serializer_class` method on the models and letting the active serializer method automatically pick the right serializer. So this might be tricky when you add ama appeals.

PR #6379, opened 2018-07-26
add endpoints for judge info and user info

amprokop:
Connects https://github.com/department-of-veterans-affairs/caseflow/issues/6395

### Testing Plan

 `curl --header "TOKEN: invalid_token" localhost:3000/idt/api/v1/judges` returns 403
 `curl --header "TOKEN: invalid_token" localhost:3000/idt/api/v1/user` returns 403

 `curl --header "TOKEN: valid_token" localhost:3000/idt/api/v1/judges` returns output like:

```{"data":[{"first_name":"Larissa","middle_name":"J","last_name":"Howell","vacols_attorney_id":"801"},{"first_name":"Ferne","middle_name":"U","last_name":"Weber","vacols_attorney_id":"647"},{"first_name":"Anastacio","middle_name":"H","last_name":"Bode","vacols_attorney_id":"876"}]```

 `curl --header "TOKEN: valid_token" localhost:3000/idt/api/v1/user` returns output like:

```
{"data":{"first_name":"Sunny","middle_name":"A","last_name":"Ritchie","attorney_id":"1860","judge_status":"none","css_id":"BVASRITCHIE"}}%         
```

### CODE REVIEW

File: app/controllers/idt/api/v1/base_controller.rb [hunk 1]
```diff
@@ -0,0 +1,29 @@
+class Idt::Api::V1::BaseController < ActionController::Base
```
amprokop: Refactored to break out a base controller so we can share some helper methods and auth logic

app/controllers/idt/api/v1/appeals_controller.rb
```diff
@@ -1,33 +1,18 @@
...
-    user = User.find_by(css_id: css_id)
-    return render json: { message: "User must be attorney" }, status: 403 unless user.attorney_in_vacols?
-    appeals = file_number ? appeals_by_file_number : appeals_assigned_to_user(user)
+    appeals = file_number ? appeals_by_file_number : appeals_assigned_to_user
 
     render json: json_appeals(appeals)
   end
 
-  def appeals_assigned_to_user(user)
+  def details
```
aroltsch: Should we remove it since it is empty?

app/controllers/idt/api/v1/base_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+  protect_from_forgery with: :exception
+  before_action :validate_token
+
+  def validate_token
+    return render json: { message: "Missing token" }, status: 400 unless token
+    return render json: { message: "Invalid token" }, status: 403 unless Idt::Token.active?(token)
+  end
+
+  def verify_attorney_user
```
aroltsch: Attorney is the only person who will be using it in the future? If not, we should name it verify_access to be more generic

app/controllers/idt/api/v1/base_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+    return render json: { message: "User must be attorney" }, status: 403 unless user.attorney_in_vacols?
+  end
+
+  def user
+    @user ||= User.find_by(css_id: css_id)
+  end
+
+  def file_number
+    request.headers["FILE"]
```
aroltsch: I like to spell things out to make more clear. I would call it FILE_NUMBER

app/repositories/judge_repository.rb
```diff
@@ -1,10 +1,20 @@
...
       User.find_or_create_by(css_id: record.sdomainid, station_id: User::BOARD_STATION_ID)
     end
   end
-  # :nocov:
+
+  def self.find_all_judges_with_name_and_id
+    records = VACOLS::Staff.where(svlj: %w[J A], sactive: "A")
+    # Our user model only contains a full name field, but for certain applications
+    # (like the IDT), we need the separate name fields from VACOLS.
+    records.map do |record|
```
aroltsch: This will not create users in our users table, could that be a problem?

app/repositories/judge_repository.rb
```diff
@@ -1,10 +1,20 @@
...
   def self.find_all_judges
     records = VACOLS::Staff.where(svlj: %w[J A], sactive: "A")
     records.select(&:sdomainid).map do |record|
       User.find_or_create_by(css_id: record.sdomainid, station_id: User::BOARD_STATION_ID)
     end
   end
-  # :nocov:
+
+  def self.find_all_judges_with_name_and_id
+    records = VACOLS::Staff.where(svlj: %w[J A], sactive: "A")
```
aroltsch: Hmmm, not a huge fan that we are duplicating logic from the method above

app/controllers/idt/api/v1/appeals_controller.rb
amprokop: I'll be implementing that in the next PR

app/controllers/idt/api/v1/base_controller.rb
amprokop: Yep that's what I originally had. I had problems with curl stripping out the header because it had an underscore in the name. I try to avoid headers and underscores in header names because of pesky problems. Would you be happy with FILENUMBER?

app/repositories/judge_repository.rb
amprokop: I do not think so, we don't need them to be users in our user table

app/repositories/judge_repository.rb
amprokop: I can abstract out this line into a separate method, even though it'll be more code overall :)

app/controllers/idt/api/v1/base_controller.rb
amprokop: sure i can change that.

app/models/serializers/idt/v1/appeal_serializer.rb
```diff
@@ -1,4 +1,5 @@
 class Idt::V1::AppealSerializer < ActiveModel::Serializer
+  # :nocov
```
aroltsch: You can try removing `nocov` because I've implemented the test for it and merged this morning.

Issue #5868, opened 2018-06-11
Fix ratings call when there are zero ratings within 1 year timeframe and display a message 


abbyraskinUSDS:
Currently, if a Vet has no ratings in BGS, users see a normal error, but then their account is stuck in limbo (they can't do anything else until it's cleared out in BGS).

If the Veteran does not have any rating issues from the past year, we should show a message to users, so they can notify the Veteran or take any necessary actions.

## Acceptance Criteria
- Fix the backend issue to allow AMA intakes to happen even if there are no timely ratings in BGS 

```
NoMethodError: undefined method `[]' for nil:NilClass
	from app/models/rating.rb:65:in `ratings_from_bgs_response'
	from app/models/rating.rb:46:in `fetch_timely'
	from app/models/veteran.rb:104:in `timely_ratings'
	from (irb):27
```
- Display a message on this page (copy below) under the decision issues header

### Copy
> This Veteran has no rated, disability issues within the one-year timeframe for requesting a review. If the Veteran listed non-rated issues, add them below. Otherwise, cancel the intake and follow your standard operating procedures for notifying the Veteran.

abbyraskinUSDS:
Asking Rachel for suggestions.

abbyraskinUSDS:
Mock to follow

abbyraskinUSDS:
I think we should rethink this interaction altogether.

abbyraskinUSDS:
Since they can now add non-rated issues, i think this approach makes sense (I was thinking we'd want to detect this earlier in the process)

abbyraskinUSDS:
@leikkisa here's the no ratings ticket

leikkisa:
Should I keep the "Select from previously rated issues" heading?
<img width="866" alt="screen shot 2018-07-30 at 3 45 50 pm" src="https://user-images.githubusercontent.com/6589610/43427745-d37f1e6a-940f-11e8-9226-6faa3c6f6841.png">

suzchap:
Abby and I discussed... agree the language here is awkward in this scenario but I plan to do a full design/content audit in the next few weeks so I added this to my [Audit & Style Guide ticket](https://github.com/department-of-veterans-affairs/appeals-design-research/issues/938) so we're fine leaving it as-is for now. Thanks for noticing it!

leikkisa:
Can validate this using Veteran 933732230

PR #6390, opened 2018-07-27
Allow AMA intakes when Veteran has no rated issues

leikkisa:
Connects #5868

This allows a Veteran without Ratings to still go through the intake process.  If they have no ratings, a message will be displayed to the user notifying them of this in case they need to add ratings.  Otherwise they can proceed with non-rated issues only.

Note: this does not create non-rated End Products for the non-rated issues yet.


leikkisa:
Waiting to hear back if we want to remove the "Select from previous decision issues" header if there are no ratings.

screen shot 2018-07-30 at 3 45 50 pm

app/models/rating.rb
```diff
@@ -3,6 +3,7 @@ class Rating
 
   attr_accessor :participant_id, :profile_date, :promulgation_date
 
+  # One week buffer was added
```
leikkisa: This is not related to this PR, and I didn't put much because I couldn't remember the whole conversation about it (something to do with mail timing), but I thought it was useful to note that this is 1 year plus 1 week

app/models/rating.rb
```diff
@@ -33,6 +34,8 @@ def fetch_issues
     [response[:rating_issues]].flatten.map do |issue_data|
       RatingIssue.from_bgs_hash(issue_data)
     end
+  rescue Savon::Error
```
kahlouie: Should we rescue a very specific Savon error or all of them? Not entirely sure what they look like, so I'd be curious if we always get the same error if they have no ratings.

app/models/rating.rb
```diff
@@ -48,6 +51,8 @@ def fetch_timely(participant_id:, from_date:)
       end
 
       unsorted.sort_by(&:promulgation_date).reverse
+    rescue Savon::Error
```
kahlouie: here too.

PR #6394, opened 2018-07-27
Anya/update colocated seeds

aroltsch:
Seed colocated tasks with legacy appeals.

To test:
1. Become a colocated user
2. In your browser http://localhost:3000/tasks?role=colocated&user_id=7
3. You will only need appeals with `appeal_type: "LegacyAppeal"`

### DIFF
--- a/app/models/serializers/work_queue/task_serializer.rb
+++ b/app/models/serializers/work_queue/task_serializer.rb
@@ -23,4 +23,12 @@ class WorkQueue::TaskSerializer < ActiveModel::Serializer
   attribute :docket_number do
     object.appeal.docket_number
   end
+
+  attribute :veteran_name do
+    object.appeal.veteran_name
+  end
+
+  attribute :veteran_file_number do
+    object.appeal.veteran_file_number
+  end
 end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -205,6 +205,11 @@ class SeedDB
       veteran: FactoryBot.create(:veteran),
       request_issues: FactoryBot.build_list(:request_issue, 1, description: "Knee pain")
     )
+
+    LegacyAppeal.create(vacols_id: "2096907", vbms_id: "228081153S")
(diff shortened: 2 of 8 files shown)


app/models/serializers/work_queue/task_serializer.rb
```diff
@@ -23,4 +23,12 @@ class WorkQueue::TaskSerializer < ActiveModel::Serializer
   attribute :docket_number do
     object.appeal.docket_number
   end
+
+  attribute :veteran_name do
+    object.appeal.veteran_name
+  end
+
+  attribute :veteran_file_number do
+    object.appeal.veteran_file_number
```
aroltsch: appeal is polymorphic so it will either call Appeal or LegacyAppeal depending on the appeal_type

db/seeds.rb
```diff
@@ -205,6 +205,11 @@ def create_ama_appeals
       veteran: FactoryBot.create(:veteran),
       request_issues: FactoryBot.build_list(:request_issue, 1, description: "Knee pain")
     )
+
+    LegacyAppeal.create(vacols_id: "2096907", vbms_id: "228081153S")
```
aroltsch: These are taken from Facols data

spec/controllers/tasks_controller_spec.rb
```diff
@@ -41,10 +41,14 @@
         expect(response_body.first["attributes"]["status"]).to eq "on_hold"
         expect(response_body.first["attributes"]["assigned_by"]["id"]).to eq user.id
         expect(response_body.first["attributes"]["placed_on_hold_at"]).to_not be nil
+        expect(response_body.first["attributes"]["veteran_name"]).to_not be nil
```
evankroske: It seems like this test could pass even if there were a serious problem with veteran_name. For example, veteran_name could be an empty string instead of the name of the veteran. How hard would it be to change these expectations to check that these attributes have the correct values?

aroltsch: Good point! I think it will be fairly simple to change it.

PR #6400, opened 2018-07-27
When a task is assigned to colocated user and appeal is ama, do not update Vacols location

aroltsch:
When a task is assigned to colocated user and appeal is ama, do not update Vacols location


spec/factories/task.rb
```diff
@@ -54,6 +54,7 @@
       type "AttorneyTask"
       appeal_type "Appeal"
       appeal { create(:appeal) }
+      parent { create(:ama_judge_task) }
```
aroltsch: It fails validation if I don't have it here because ama tasks must have a parent task based on our design.

Issue #6199, opened 2018-07-12
Algorithm | Output to UI


sharonwarner:
- Add a cache_attribute method to ro_schedule_period.rb and judge_schedule_period.rb that 
    - Confirms the schedulePeriod has finalized == false (if finalized == true, you can return an empty array)
    - Runs the algorithm 
    - Saves the output
- Update `show` in schedule_periods_controller.rb to
    - Run the cache_attribute method
    - Send the cache_attribute json object to the frontend
- Confirm frontend renders correctly

The judge algorithm should output a json object with the following format:
```
[{hearing_date: '', hearing_type: '', regional_office: '', room: '', judge_id: ''}, 
{hearing_date: '', hearing_type: '', regional_office: '', room: '', judge_id: ''}]
```

The ro/co algorithm should output a json object with the following format:
```
[{hearing_date: '', hearing_type: '', regional_office: '', room: ''}, 
{hearing_date: '', hearing_type: '', regional_office: '', room: ''}]
```

PR #6404, opened 2018-07-27
showing hearing days in the UI

tejans24:
Connects #6199 

### Description
Shows hearing

### Acceptance Criteria 
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...


app/controllers/hearings/schedule_periods_controller.rb
```diff
@@ -5,35 +5,14 @@ def index
...
-          hearing_date: "2018-06-04",
-          hearing_type: "Video",
-          regional_office: "Portland, OR",
-          room: "1",
-          judge: "Sarah Smith"
-        }
-      ]
+    schedule_period = SchedulePeriod.find(params[:schedule_period_id])
+    render json: { schedule_period: schedule_period.to_hash.merge(
+      hearing_days: schedule_period.ro_hearing_day_allocations.map do |hearing_day|
```
sharonwarner: This will break for JudgeSchedulePeriods. We should have a more generic name for that method (maybe `algorithm`?) so that this code will work for both.

app/models/ro_schedule_period.rb
```diff
@@ -17,4 +21,29 @@ def schedule_confirmed(hearing_schedule)
...
+              regional_office: ro_key
+            )
+          end
+        end
+      end
+      acc
+    end
+  end
+
+  def genenerate_ro_hearing_schedule
```
sharonwarner: Nit: should be 'generate'

PR #6405, opened 2018-07-27
Fixing appeal factory

mdbenjam:
### Description
This test regularly fails because in a recent change we updated the factory to change the veteran file number after setting documents for the old file number. We move setting the file number into the build part instead of the create part of the factory.

### Acceptance Criteria 
- [ ] The failing test in ama_queue_spec passes

### DIFF
--- a/spec/factories/appeal.rb
+++ b/spec/factories/appeal.rb
@@ -32,7 +32,6 @@ FactoryBot.define do
     after(:build) do |appeal, evaluator|
       if evaluator.veteran
         appeal.veteran_file_number = evaluator.veteran.file_number
-        appeal.save
       end
 
       Fakes::VBMSService.document_records ||= {}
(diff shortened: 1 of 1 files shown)


spec/factories/appeal.rb
```diff
@@ -18,22 +18,23 @@
...
     end
 
     transient do
       documents []
     end
 
     after(:build) do |appeal, evaluator|
+      if evaluator.veteran
+        appeal.veteran_file_number = evaluator.veteran.file_number
+        appeal.save
```
aroltsch: build doesn't save the appeal in the DB so calling `save` will defend the purpose of the build function.

aroltsch: and the test that is failing is calling create
https://github.com/department-of-veterans-affairs/caseflow/blob/master/spec/feature/queue/ama_queue_spec.rb#L37

mdbenjam: I agree we should remove the same. However even though the test calls create build gets called first. So we get

1) Appeal generates a `veteran_file_number`
1) build hook is called, and assigns documents to `Fakes::VBMSService.document_records[appeal.veteran_file_number]`
1) create hook is called and changes the `veteran_file_number`.

Issue #6374, opened 2018-07-26
Fix relationship map bug in Prod


leikkisa:
When a user tries to intake an AMA intake in Prod, the app gets stuck spinning in loading on the Veteran search screen.  This happened on both Internet Explorer (the user's browser), and also Chrome on Abby's browser.

Ideas to troubleshoot:
Look at what gets returned in Prod when we get relationships for a Veteran and see if it's the same as UAT
Try the lodash version of map

https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2132/
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2131/

kahlouie:
It looks like the object isn't mappable (not an iterable). It's looking for an array. The object might be a singular object.

abbyraskinUSDS:
Fix broke caseflow details page - further investigation needed

kahlouie:
It didn't break caseflow details page, but it did get rolled back in the deploy. Will wait until the next deploy for it to go out in prod. Also fun news is that I should have access to test prod.

leikkisa:
For validation, I think @abbyraskinUSDS should try starting one of the BEAAM intakes, and then if she can get to the Review page (meaning the error is fixed), then she can cancel the intake.

leikkisa:
Abby validated this in Prod

PR #6406, opened 2018-07-27
Fix feature toggle bug

leikkisa:
Connects #6374 

Fix bug where AMA Intake breaks if feature toggle is only enabled for selected users.


leikkisa:
I just want to note that we still want to write tests for this.  

We should also see how we are checking if the feature toggle is enabled elsewhere (if we're checking it based on current user) as a preventative measure.

PR #6421, opened 2018-07-30
Use extra context instead of including ramp id in title

sunil-sadasivan:
This should allow us to see each sentry error caused by the Ramp Background job https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2138/ 

As multiple errors of the same error type.

The ramp election ids will be stored as extra context in the sentry error.


kahlouie:
Nice, thanks @sunil-sadasivan

PR #6425, opened 2018-07-30
Judge assignment algorithm connection

tejans24:
Connects #6199 

### Description
Connects the judge assignment algorithm to the schedule period. The data should be cached, shows judge algorithm in the UI and can be later submitted to VACOLS.

### Acceptance Criteria 
- [ ] 

### Testing Plan
1. Go to ...


app/repositories/hearing_day_repository.rb
```diff
@@ -23,7 +21,8 @@ def create_schedule(scheduled_hearings)
 
     def update_schedule(updated_hearings)
       updated_hearings.each do |hearing|
-        hearing_to_update = VACOLS::CaseHearing.find(hearing.hearing_pkseq)
+        hearing_to_update = VACOLS::CaseHearing.find(hearing[:hearing_pkseq])
+        hearing.delete(:hearing_pkseq)
```
sharonwarner: And why are deleting hearings here?

app/models/judge_schedule_period.rb
```diff
@@ -11,10 +15,21 @@ def import_spreadsheet
     JudgeNonAvailability.import_judge_non_availability(self)
   end
 
-  # :nocov:
   def schedule_confirmed(hearing_schedule)
-    HearingDay.update_schedule(hearing_schedule)
+    hearing_days = hearing_schedule.map do |hearing_day|
+      hearing_day.delete(:judge_name)
```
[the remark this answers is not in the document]
tejans24: because they are used in the update method in the case hearing that matches the column names with attributes. 

if the attribute name doesn't exist, it becomes nil and update method throws an error.

app/repositories/hearing_day_repository.rb
tejans24: Same reason as above: 
https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/vacols/case_hearing.rb#L152

https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/vacols/case_hearing.rb#L36 - doesn't contain hearing_pkseq

tejans24: it doesn't need that attribute to update the record. Since we already used it to find the particular hearing

app/models/judge_schedule_period.rb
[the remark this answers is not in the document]
tejans24: The regional office needs to added, we don't have a column listed in there.

Issue #5430, opened 2018-05-07
Send associated rating issue with each contention created for Higher Level Reviews and Supplemental Claims


shanear:
In order for the VBMS NWQ to be able to route same station Higher Level Reviews correctly, they need to know the rating issue being reviewed for each contention. These will be passed

## Acceptance criteria
- Pass the correct rating issue id (`rba_issue_id`) along with each contention being created as part of a higher level review or supplemental claim. 

## Technical notes
- We need some way of verifying that VBMS received the data correctly.
- This issue is dependent on https://github.com/department-of-veterans-affairs/caseflow/issues/5428

shanear:
We discovered a bug with this service where the wrong id is being used. VBMS is fixing the bug, but said it might take a while.

shanear:
This should be fixed tomorrow. Going to retest.

shanear:
Retested this today and there is now another bug. Should be retestable in 1-2 weeks 😢

leikkisa:
The order of the process should be:
establishClaim
createContentions (against the returned claimId)
associateRatedIssues (with the contentionId and claimId's returned previously)

leikkisa:
We cannot validate this ourselves, because success doesn't mean it shows up on the associated rated issues in VBMS (according to Joe).  It sounds like he would need to manually validate this somehow.

leikkisa:
Have asked Joe to help validate this on test file number: 984562385

leikkisa:
Joe says it worked!  He checked rated issue ID 113584 (test data)

PR #6445, opened 2018-07-31
[WIP] Map contentions to rated issues

leikkisa:
Connects #5430

Add code to map contentions to rated issues after contentions are created


leikkisa:
Rebuilding the front end for tests is off again (sad face)

Issue #6396, opened 2018-07-27
Refactor Queue to enable pulling only basic appeal information from tasks endpoint and task info from the appeals endpoint


mdbenjam:
In thinking through access control, we've decided the need to separate basic appeal info (name, docket number etc) from more sensitive information (appellant, rep, issues, etc.) In order to get to this point we need to separate endpoints. The first endpoint is the tasks endpoint which will return basic information on each appeal. This will not do a sensitivity access check before returning. The second endpoint is the appeals endpoint which will start returning any tasks of the appeal assigned to the current user, and will return the full appeal information. This will do a sensitivity access check before returning.

The redux store will store denormalized state in three places.
1) `queue.basic_appeal_info`
1) `queue.full_appeal_info`
1) `queue.tasks`

We'll make requests to the tasks controller on `queue` pages and requests to the appeal controller on the `details` page.

evankroske:
Will full_appeal_info store a superset of the information in basic_appeal_info?

mdbenjam:
I'm thinking yes. Normally against duplicating state... But I think this might be a reasonable use case. We could also have them store a disjoint set of data.

PR #6446, opened 2018-07-31
Refactor appeals into appealDetails and appeals

mdbenjam:
Connects #6396

### Description
This PR does several things:
1) Defines a new `appealDetails` object in `state.queue`. Now the `appeals` object only contains basic (non-bgs related) appeal info and `appealDetails` includes all appeal info.
1) Process the basic info that goes into `appeals` and `tasks` to get rid of the `attributes` and underscored variables.
1) Moves the appeal store for case search from it's own reducer to `state.queue.appealDetails`
1) Adds new selectors that can grab new subsets of appeals. These are somewhat specific now, but will hopefully be generalized when we start building other queues. i.e. `tasksForAppealAssignedToAttorneySelector` -> `tasksForAppealAssignedToDirectReportSelector`
   - getTasksForAppeals
   - tasksForAppealAssignedToUserSelector
   - tasksForAppealAssignedToAttorneySelector
   - appealsByCaseflowVeteranId
1) We remove the constraint that all appeals can only have one task, as that constraint is going away.

### Acceptance Criteria 
- [ ] Everything still works as expected


client/app/queue/CaseDetail/CaseDetailActions.js
```diff
@@ -1,15 +0,0 @@
-import * as Constants from './actionTypes';
```
mdbenjam: We no longer need to set appeals/tasks as active. We get the necessary information from the URL.

client/app/queue/CaseDetailsLink.jsx
```diff
@@ -1,41 +1,28 @@
...
     return <React.Fragment>
       <Link
-        to={`/queue/appeals/${appeal.vacols_id}`}
+        to={`/queue/appeals/${appeal.externalId}`}
         disabled={disabled}
-        onClick={this.props.onClick || this.setActiveAppealAndTask}>
+        onClick={this.props.onClick}>
         {this.props.getLinkText(appeal)}
       </Link>
-      {!_.isNull(_.get(appeal, 'appellant_full_name')) && <React.Fragment>
```
mdbenjam: We will no longer show if the appellant is the veteran as that requires another call to BGS.

client/app/queue/CaseDetailsLoadingScreen.jsx
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    if (!tasks || _.filter(tasks, (task) => task.externalAppealId === appealId).length > 0) {
+      promises.push(taskPromise);
+    }
+
+    return Promise.all(promises);
+  };
+
+  loadAttorneysOfJudge = () => {
```
mdbenjam: I eventually would like to move these calls into a different loading screen, or component. However, for now it's easiest just to add them here.

client/app/queue/CaseDetailsView.jsx
```diff
@@ -77,21 +72,16 @@ CaseDetailsView.propTypes = {
...
-  const { appeals, tasks } = state.queue;
+const mapStateToProps = (state, ownProps) => {
+  const { appealDetails } = state.queue;
   const { success, error } = state.ui.messages;
 
   return {
-    appeal: activeAppeal ? appeals[activeAppeal.attributes.vacols_id] : activeAppeal,
-    task: activeTask ? tasks[activeTask.id] : activeTask,
+    appeal: appealDetails[ownProps.appealId],
+    task: tasksForAppealAssignedToUserSelector(state, ownProps)[0],
```
mdbenjam: Our old code assumed there could only be one task per appeal. This is no longer going to be true in AMA. So for now we just take the first task returned since we only expect there to be one task.

client/app/queue/CaseList/CaseListReducer.js
```diff
@@ -6,7 +6,6 @@ export const initialState = {
     searchQuery: ''
   },
   isRequestingAppealsUsingVeteranId: false,
-  receivedAppeals: [],
```
mdbenjam: Moving the appeals from the search queries to the appealDetails section in queue.

client/app/queue/CaseSnapshot.jsx
```diff
@@ -86,48 +83,46 @@ export class CaseSnapshot extends React.PureComponent<Props> {
...
       return false;
     }
-    if (this.props.appealsAssignedToCurrentUser.
-      some((appealIterator) => appealIterator.attributes.vacols_id === this.props.appeal.attributes.vacols_id)) {
+    if (this.props.task) {
       return true;
     }
-    if (!this.props.task) {
-      return false;
+    if (this.props.taskAssignedToAttorney) {
```
mdbenjam: I think this is what the 
```
_.some(	    return false;
      this.props.attorneysOfJudge, (attorney) => attorney.id === this.props.task.attributes.assigned_to_pg_id);
```
was trying to determine.

client/app/queue/QueueLoadingScreen.jsx
```diff
@@ -74,38 +67,6 @@ class QueueLoadingScreen extends React.PureComponent<Props> {
       }));
   };
 
-  loadActiveAppealAndTask = (appealId) => {
```
mdbenjam: Moving this to the case details loading screen.

client/app/queue/CaseListTable.jsx
```diff
@@ -70,9 +66,5 @@ CaseListTable.propTypes = {
 
 const mapStateToProps = () => ({});
 
-const mapDispatchToProps = (dispatch) => bindActionCreators({
-  setActiveAppeal
-}, dispatch);
-
-export default connect(mapStateToProps, mapDispatchToProps)(CaseListTable);
+export default connect(mapStateToProps)(CaseListTable);
```
evankroske: Since CaseListTable doesn't depend on information from the store anymore, I think that we can remove mapStateToProps and connect

client/app/queue/CaseListView.jsx
```diff
@@ -82,8 +83,8 @@ CaseListView.defaultProps = {
   caseflowVeteranId: ''
 };
 
-const mapStateToProps = (state) => ({
-  appeals: state.caseList.receivedAppeals
+const mapStateToProps = (state, ownProps) => ({
+  appeals: appealsByCaseflowVeteranId(state, ownProps)
```
evankroske: I feel like passing all of ownProps here makes this code difficult to understand

client/app/queue/CaseSnapshot.jsx
evankroske: Did you mean to change this code to always show the actions dropdown if the parameter task is defined? It seems like this would cause the dropdown to appear in a lot of situations in which it shouldn't. For example, it shouldn't appear when a judge looks at a task assigned to an attorney on another judge's team. What do you think?

client/app/queue/CaseSnapshot.jsx
```diff
@@ -170,23 +165,14 @@ export class CaseSnapshot extends React.PureComponent<Props> {
...
-  appeal: PropTypes.object.isRequired,
-  featureToggles: PropTypes.object,
-  appealsAssignedToCurrentUser: PropTypes.array,
-  task: PropTypes.object,
-  userRole: PropTypes.string,
-  hideDropdown: PropTypes.bool
-};
-
-const mapStateToProps = (state: State) => {
+const mapStateToProps = (state: State, ownProps: Object) => {
```
evankroske: FYI, I'm pretty sure that ownProps is an instance of Params

client/app/queue/CaseDetailsLink.jsx
dannysellers: Do we plan on displaying that information anywhere else (like in Case Details)? Or should we remove `COPY.CASE_DIFF_VETERAN_AND_APPELLANT`?

client/app/queue/components/TaskTable.jsx
```diff
@@ -95,15 +95,17 @@ class TaskTable extends React.PureComponent<Props> {
           }
 
           return <Checkbox
-            name={task.id}
```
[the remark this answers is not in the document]
mdbenjam: Hmm shoot, I'll change that back then.

client/app/queue/components/TaskTable.jsx
```diff
@@ -95,15 +95,17 @@ class TaskTable extends React.PureComponent<Props> {
...
+            name={task.taskId}
             hideLabel
-            value={this.isTaskSelected(task.id)}
+            value={this.isTaskSelected(task.taskId)}
             onChange={
-              (checked) =>
```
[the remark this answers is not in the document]
mdbenjam: Whoops, no I didn't. Fixed.

client/app/queue/selectors.js
```diff
@@ -13,21 +13,29 @@ import type {
...
-export const selectedTasksSelector = (state: State, userId: string) => _.flatMap(
-  state.queue.isTaskAssignedToUserSelected[userId] || {},
-  (selected, id) => selected ? [state.queue.tasks[id]] : []
-);
+export const selectedTasksSelector = (state: State, userId: string) => {
+  const flatTasks = _.flatten(_.values(state.queue.tasks));
+
+  return _.flatMap(
+    state.queue.isTaskAssignedToUserSelected[userId] || {},
+    (selected, id) => selected ? [_.find(flatTasks, { taskId: id })] : []
```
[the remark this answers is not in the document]
mdbenjam: Based on your previous comment I changed the `isTaskAssignedToUserSelected` to use `externalAppealId` instead of `taskId`. So I can return it to what it was like before.

That said, the original incarnation is not many-task-per-appeal friendly. If there are multiple tasks for a single appeal, looking tasks up by the vacols_id will not work. So this is a change we'll need to make in the future.

client/app/queue/selectors.js
```diff
@@ -39,40 +47,61 @@ export const amaTasksByAssigneeId: (State) => { [number]: AmaTask } = createSele
...
-    const taskMap = _.reduce(tasks, (map, task) => {
-      const taskList = map[task.attributes.appeal_id] ? [...map[task.attributes.appeal_id], task] : [task];
-
-      return { ...map,
-        [task.attributes.appeal_id]: taskList };
-    }, {});
-
     return _.map(appeals, (appeal) => {
-      appeal.tasks = taskMap[appeal.id];
+      appeal.tasks = _.filter(tasks, (task) => task.externalAppealId === appeal.externalId);
```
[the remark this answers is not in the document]
mdbenjam: I like that.

client/app/queue/utils.js
```diff
@@ -20,61 +22,90 @@ import DIAGNOSTIC_CODE_DESCRIPTIONS from '../../constants/DIAGNOSTIC_CODE_DESCRI
...
+          isAdvancedOnDocket: task.attributes.aod,
+          issueCount: task.attributes.issue_count,
+          docketNumber: task.attributes.docket_number,
+          veteranName: task.attributes.veteran_name,
+          veteranFileNumber: task.attributes.veteran_file_number,
+          isPaperCase: task.attributes.paper_case
+        };
       }
-    }
-    if (Array.isArray(outerTasks)) {
```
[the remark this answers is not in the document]
mdbenjam: Yup, you're right, I tested all the other views, but forgot the BEAAM appeals 😞. I'm going to fix that now.

client/app/queue/CaseDetailsView.jsx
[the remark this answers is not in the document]
mdbenjam: I agree. This is how reselect suggests doing it. But maybe we can setup our own patterns to pass an object with just the necessary props.

client/app/queue/CaseDetailsLoadingScreen.jsx
```diff
@@ -0,0 +1,140 @@
...
+
+  createLoadPromise = () => Promise.all([
+    this.loadActiveAppealAndTask(),
+    this.maybeLoadJudgeData()
+  ]);
+
+  reload = () => window.location.reload();
+
+  render = () => {
+    const failStatusMessageChildren = <div>
```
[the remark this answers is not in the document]
mdbenjam: I'm going to leave this as is for now. I don't want to make another copy change (removing the refresh button) until syncing with designers on it. But long term I think that makes sense.

client/app/queue/selectors.js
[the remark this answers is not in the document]
evankroske: Good point. We'll have to rethink this

Issue #6068, opened 2018-06-29
Upgrade Ruby 


nemodjuric:
Caseflow team will upgrade Ruby from current 2.2.4 version to the latest stable version 2.5.1
Release notes for Ruby 2.5 can be found here: https://www.ruby-lang.org/en/news/2017/12/25/ruby-2-5-0-released/
Ruby upgrade should include security fixes and upgrades, it should bring some performance gains and overall reliability. 
Also, it will keep our ruby ecosystem up to date and easier to maintain. Beside that, all support of the Ruby 2.2 series has ended. 

Once everything installed, with a passing test suite and no deprecation warnings, 
Ruby upgrade should be deployed and tested in UAT as separate branch
After successful sanity check in UAT, it should be deployed to Production during after-business hours. 

Guideline for sanity checks:
https://github.com/department-of-veterans-affairs/caseflow/issues/6067

ferlatte:
I admit, I don't know much about the production environment setup. Does it honor .ruby-version automatically? I don't believe that CircleCI does; we'll need to make a new base container with the appropriate version and deploy it when we want to commit to 2.5.1.

Does it make sense to get CircleCI running our tests using both Ruby versions during the transition period? It looks like that's possible with rbenv and may be worth doing so that we can keep future Ruby upgrades easy.

nemodjuric:
Yes, our production environment does ruby-version-lookup where it looks into .ruby_version, then ruby-existence-check to see if that version is installed, and then if it's not it runs ansible role called ruby which includes installation of set default ruby version. 
Running both versions of Ruby makes sense, I'm eager to see how will it work in practice. Would there be two separate builds using different containers? Just curious - what about approach to have CircleCI read .ruby-version file similar to what we do in production environment setup?

ferlatte:
I think we can make the CircleCI environment honor .ruby-version without it being too painful; that seems like a good thing to do regardless.

I'd need to see what CircleCI 2.0 provides now for multiple versions; I know they have support for parallel builds using different base containers when they added workflow support, which might do the trick.

ferlatte:
Nemo and I chatted this morning and have a plan for getting CircleCI to honor .ruby-version. I'm going to do some exploratory work on the Caseflow CircleCI container tomorrow morning and see if I can get something working. We aren't going to setup parallel test runs with multiple Rubies for the time being.

shanear:
closing this since its upgraded in prod. @enriquemanuel will be handling fixing CI

PR #6448, opened 2018-07-31
Mwf support ruby version

ferlatte:
Connects #6068 

### Description
Modified the CircleCI container to support .ruby-version by installing rbenv.
Modified the CircleCI config to use .ruby-version.

### Acceptance Criteria 
- [ ] Code compiles correctly
- [ ] Tests pass
- [ ] Ruby versions are used and installed correctly.

### Testing Plan
1. Submit job to CircleCI using existing config. Tests should pass.
2. Use .ruby-version 2.2.4. Tests should pass.
3. Try a different ruby version (2.5.1). Tests should run with new ruby version, may not pass.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,6 +39,19 @@ jobs:
 
       - restore_cache:
           keys:
+            - rbenv-cache-v1-{{ arch }}-{{ checksum ".ruby-version" }}
+
+      - run:
+          name: Setup Ruby
+          command: rbenv install --skip-existing $(cat ~/project/.ruby-version)
+
+      - save_cache:
+          key: rbenv-cache-v1-{{ arch }}-{{ checksum ".ruby-version" }}
+          paths:
+            - .rbenv
+
+      - restore_cache:
+          keys:
             - gem-cache-v1-{{ arch }}-{{ .Branch }}-{{ checksum "Gemfile.lock" }}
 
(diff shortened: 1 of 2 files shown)


ferlatte:
Deploying this will involve: building the new Docker container and uploading it, modifying this diff to use circleci:latest instead of circleci:mwftest, and then merging.

.circleci/config.yml
```diff
@@ -37,11 +37,34 @@ jobs:
           name: Setup testfiles directory
           command: ~/project/ci-bin/capture-log "mkdir -p ~/project/tmp/testfiles"
 
+      - restore_cache:
```
mdbenjam: This is a good idea, but do we know how much it adds to the build? We rarely upgrade ruby, so if this adds too much to the build time I'd advocate building this into the image. If since it'll be cached makes it negligible then maybe this is the right way to go.

ferlatte: The ruby build adds about 10 minutes when you try a new version; after that, the build time is the same.

I considered building it into the container initially and decided that requiring an additional step for the upgrade was probably not ok, especially if we want to keep Ruby up to date (which, like Rails, I think we probably do).

ferlatte:
I'm going to split this PR into two separate ones in order to make the deployment of this easier; one for the CircleCI container, and then one for the new CircleCI config.

Issue #5991, opened 2018-06-21
Add decision date for non-rated issues


abbyraskinUSDS:
CAs will need to enter the last decision date for each non-rated issue, so these issues can be tracked for timeliness. This story covers the component for adding the date, but does not handle the timeliness validation.

## Acceptance Criteria
- Issue category and issue description are mandatory fields
- A date picker field with the label "Decision date" is present when adding a new non-rated issue
- This field is mandatory, unless it's an uncategorized issue
- Uncategorized issues do not require a decision date 

## Design
![image](https://user-images.githubusercontent.com/4042181/42457669-0a9bfb3a-8366-11e8-9d1e-b38f30fb6990.png)

leahbannon:
note - we need to also write a story for checking timeliness

leahbannon:
pair with Sally on this! :)

suzchap:
I've needed to make some adjustments on this to accommodate issues around editing issues... so hold off on this until I can update this ticket. Please and thank you :)

kahlouie:
@suzchap noted.

suzchap:
Ok, sounds like y'all should just move forward with this issue as-is... just keep in mind we'll be rearranging the fields and adding the remove issue in a later issue. Thanks!

kahlouie:
The current work as is in prod behind the intakeAma toggle allows users to submit a decision date that's not complete, like `03/` as the date. We'll need to fix that before continuing forward.

kahlouie:
@suzchap Looks like if decision date is required, there's a weird formatting issue that occurs. Would you like us to add space to decisionDate so that it has room for the `Required` or would you like to move Required label on all sections to a different place?
![Screen Shot 2018-07-31 at 2.02.59 PM.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5b0329701e56750012d0676e/125516c66e82c728ace21e0d46db978826878de9bce1b057e40fa0c5710b77246457f26b7191a37af112294f5a565ecc3c4a4356bcbbd671f3e57862804e0ee5d33c5d22c33413a5b8d41bbf50bd208b1d87b3bc891ed2f42f67b6d5db1063474f4dcbed5727a19041c95df548e14a56a4113162e57e2431.png)

suzchap:
We're actually going to remove the required label moving forward. We may introduce an "optional" label at some point but not yet. Does that fully fix your problem?

kahlouie:
[...]

PR #6459, opened 2018-08-01
Increment issue count to handle decision date

kahlouie:
Connects #5991

### DIFF
--- a/client/app/intake/reducers/appeal.js
+++ b/client/app/intake/reducers/appeal.js
@@ -65,6 +65,8 @@ export const mapDataToInitialAppeal = (data = { serverIntake: {} }) => (
 );
 
 export const appealReducer = (state = mapDataToInitialAppeal(), action) => {
+  let updatedState;
+
   switch (action.type) {
   case ACTIONS.START_NEW_INTAKE:
     return mapDataToInitialAppeal();
@@ -213,7 +215,7 @@ export const appealReducer = (state = mapDataToInitialAppeal(), action) => {
       }
     });
   case ACTIONS.SET_ISSUE_CATEGORY:
-    var updatedState = update(state, {
+    updatedState = update(state, {
       nonRatedIssues: {
         [action.payload.issueId]: {
           category: {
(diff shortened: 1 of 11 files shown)


client/app/intake/reducers/appeal.js
```diff
@@ -213,33 +215,39 @@ export const appealReducer = (state = mapDataToInitialAppeal(), action) => {
       }
     });
   case ACTIONS.SET_ISSUE_CATEGORY:
-    return update(state, {
+    updatedState = update(state, {
```
mdbenjam: Just out of curiosity, was it necessary to do this state update in two parts? It's not wrong, just wondering why you chose to split it up.

mdbenjam: JK I see why you split it up. That said I think the `allIssueCounter` function makes more sense as a selector than setting an issueCount value in the state. `issueCount` now duplicates state, we can already calculate it from other values in the state. If you haven't used selectors before let me know, and I'm happy to go into more detail.

kahlouie: Yes, I'd love to pair to understand selectors!

kahlouie:
@mdbenjam Validation on a successful issue is meant to be a different story. So other than using selectors instead, I think we should handle validations in the other story.

client/app/intake/selectors.js
```diff
@@ -34,3 +35,13 @@ export const getIntakeStatus = createSelector(
     return status || INTAKE_STATES.NONE;
   }
 );
+
+const appealState = (state) => state;
```
kahlouie: @mdbenjam I'm cheating on the memoization here. Is there a better way to do what I'm doing? I'm now calling the selector twice, and feeling like there's a better way to do this.

mdbenjam: So the problem with doing this is that you'll never memoize because any time any part of the state changes you'll trigger this function to run. Options
1) Refactor formatIssues a little bit to take issues rather than the whole state.
2) Don't use reselect here. Honestly the page is so small not-memoizing is probably totally okay. No need to over optimize if we don't need it.

kahlouie:
@mdbenjam I think I figured out the right way to memoize it! Thanks for the help! Can you give it another look-see?

PR #6460, opened 2018-08-01
VLJ # in spreadsheet instead of css_id

sharonwarner:
This PR updates the spreadsheet so accept VLJ ids instead of css ids.

To test:
- Import spreadsheet with VLJ ids
- Confirm there are no errors
- Confirm the database still stores CSS ids


app/services/hearing_schedule/validate_judge_spreadsheet.rb
```diff
@@ -26,9 +26,9 @@ def validate_judge_non_availability_template
     end
   end
 
-  def find_user(css_id, name)
-    User.where(css_id: css_id,
-               full_name: name.split(", ").reverse.join(" ")).count > 0
+  def judges_in_vacols?(vlj_ids)
+    judges = UserRepository.css_ids_by_vlj_ids(vlj_ids)
```
oscarAtNava: Don't we usually access the repository functions through the main model, in this case the User model?

Issue #6077, opened 2018-07-02
Co-located table | Establish table for co-located users


laurjpeterson:
When a co-located team member is planning what to work on, they need to see all tasks assigned to them in one place and the types of tasks, so they can prioritize what to work on.

## AC
- Establish the table component that will be used in each of the tabs (New, Pending action, On hold) in the designs. 
- Verify the table contains the following columns:
   - Case details
   - Task
   - Type(s)
   - Docket number
   - Days waiting (definition here: https://github.com/department-of-veterans-affairs/caseflow/issues/5888)
   - Veteran documents
- Verify the link in the Veteran documents column navigates the user to the correct Veteran claims folder in Reader

## Out of scope
- Implementing the tabs
   - https://github.com/department-of-veterans-affairs/caseflow/issues/6078 Colocated New Tab
   - https://github.com/department-of-veterans-affairs/caseflow/issues/6079 Colocated Pending Action Tab
   - https://github.com/department-of-veterans-affairs/caseflow/issues/6080 Colocated On Hold Tab

##### New Tab
![colocated-table-new-tab-single-actions-768](https://user-images.githubusercontent.com/31224257/43283661-80eb7dae-90e7-11e8-9f71-01abb998f958.png)

##### New Tab with Tooltips
![colocated-table-new-tab-single-actions-wtooltips-768](https://user-images.githubusercontent.com/31224257/43283662-81092944-90e7-11e8-95ca-4bbfcc7a9cd1.png)

## Open Question
- @sneha-pai - should the column be `Days waiting` like it is for attorneys and judges, or `Total days waiting`?

mkhandekar:
I've updated the link to the design under "mocks" in the original issue.

evankroske:
## Implementation Plan
- [ ] Load the user's colocated tasks into the store. Estimate: 2 days.
- [ ] Add a page component for colocated admins.
- [ ] Add a table with the admin's tasks.
- [ ] Code review. Estimate: 2 days.

PR #6466, opened 2018-08-02
Caseflow API | Adds external_id field to Task

evankroske:
Blocks https://github.com/department-of-veterans-affairs/caseflow/issues/6077

### Description
Adds the field external_id to Task. external_id can be either the UUID or the VACOLS ID of the task's appeal.

### Testing Plan
- [x] Check that all tasks returned by http://localhost:3000/tasks?user_id=7&role=colocated include external_id.


app/models/serializers/work_queue/task_serializer.rb
```diff
@@ -31,4 +31,8 @@ class WorkQueue::TaskSerializer < ActiveModel::Serializer
   attribute :veteran_file_number do
     object.appeal.veteran_file_number
   end
+
+  attribute :external_id do
+    if object.appeal.respond_to? :uuid then object.appeal.uuid else object.appeal.vacols_id end
```
aroltsch: I think a better way to make use of polymorphism is to create a method external_id in each model. Each model will return its own external_id. Then you can just call `object.appeal.external_id` and that's it. The same as I did for other fields (e.g, object.appeal.type)

evankroske: Good point. Done!

Issue #6302, opened 2018-07-23
RO Algorithm | Code cleanup


sharonwarner:
- genenerate_ro_hearing_schedule's spelling should be fixed
- Clean up generate_hearing_days_schedule_spec.rb
    - I skipped a test on line 130. subject.available_days returns May 31st for Togus, ME even though it is listed as a non-availability day for that RO.
    - I skipped another test on line 323. It looks like there was a copy/paste error with the test name.
    - Non-deterministic test failure on 'too many allocated days for an RO'

![Screen Shot 2018-07-25 at 11.53.23 AM.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/582a398e7d9c5f1300c030ee/125516c66e82c728ace21e0d46db978826878de9bce1b057e40fa0c5710b77246456f66b7191a379ee0c2e52415d4dbc30254d58eeb9cb74fce36f34985619e8d96f0e76903818aeb18249bb01eb278b4fd7b4b98a4f92fa2137a5dad11b61435c52dcbd4f3fb2984cc259f348e34856a0133167e476346e0a.png)

<img width="1171" alt="screen shot 2018-08-02 at 9 50 15 am" src="https://user-images.githubusercontent.com/4306128/43588055-a746981a-9639-11e8-92af-ff7d436e569d.png">

<img width="1173" alt="screen shot 2018-08-02 at 10 26 52 am" src="https://user-images.githubusercontent.com/4306128/43590386-e24fd994-963e-11e8-9b41-2f80bbc53630.png">

<img width="1164" alt="screen shot 2018-08-02 at 10 42 52 am" src="https://user-images.githubusercontent.com/4306128/43591241-e6222ce6-9640-11e8-9b96-133034858c5d.png">

<img width="1137" alt="screen shot 2018-08-02 at 11 01 08 am" src="https://user-images.githubusercontent.com/4306128/43592385-6e1bf7ba-9643-11e8-8aae-597655d7e97b.png">

PR #6473, opened 2018-08-02
Flaky test fixes

sharonwarner:
Connects #6302

### DIFF
--- a/app/models/ro_schedule_period.rb
+++ b/app/models/ro_schedule_period.rb
@@ -3,7 +3,7 @@ class RoSchedulePeriod < SchedulePeriod
   after_create :import_spreadsheet
 
   cache_attribute :algorithm_assignments, expires_in: 4.days do
-    genenerate_ro_hearing_schedule
+    generate_ro_hearing_schedule
   end
 
   def validate_spreadsheet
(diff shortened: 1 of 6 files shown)


app/models/ro_schedule_period.rb
```diff
@@ -3,7 +3,7 @@ class RoSchedulePeriod < SchedulePeriod
   after_create :import_spreadsheet
 
   cache_attribute :algorithm_assignments, expires_in: 4.days do
-    genenerate_ro_hearing_schedule
+    generate_ro_hearing_schedule
```
tejans24: 👍🏼

tejans24: 👍🏼

app/services/hearing_schedule/assign_judges_to_hearing_days.rb
```diff
@@ -153,12 +151,8 @@ def get_hearing_type(is_central_hearing)
...
 
   def get_judge_name(css_id)
-    if @judges[css_id][:user_info]
-      @judges[css_id][:user_info].full_name
-    else
-      staff_info = @judges[css_id][:staff_info]
-      "#{staff_info.snamef} #{staff_info.snamemi} #{staff_info.snamel}"
-    end
+    staff_info = @judges[css_id][:staff_info]
+    "#{staff_info.snamef} #{staff_info.snamemi} #{staff_info.snamel}"
```
sharonwarner: We check the judges' name from the spreadsheet in VACOLS, so I think it makes sense to just default to that name. It makes the logic simpler too.

PR #6482, opened 2018-08-02
Judge can reassign AMA case to another attorney

### LINKED ISSUE #6312: Judge should be able to reassign AMA appeal to another attorney
PATCH /tasks

aroltsch:
Fixes #6312 

### Description
Ability for a judge to reassign a case to another attorney

### DIFF
--- a/app/controllers/tasks_controller.rb
+++ b/app/controllers/tasks_controller.rb
@@ -28,6 +28,16 @@ class TasksController < ApplicationController
     render json: { tasks: json_tasks(tasks) }
   end
 
+  # e.g, for legacy appeal => POST /tasks,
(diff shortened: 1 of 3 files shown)


app/controllers/tasks_controller.rb
```diff
@@ -37,8 +47,17 @@ def create
     render json: { tasks: json_tasks(tasks) }, status: :created
   end
 
+  # e.g, for ama appeal => PATCH /tasks/:id,
+  # { type: AttorneyTask,
+  #   assigned_to_id: 23
+  # }
+  # e.g, for ama appeal => PATCH /tasks/:id,
```
dannysellers: for legacy appeal?

aroltsch: it is the same for legacy. I will add the comment

Issue #6285, opened 2018-07-19
CHECK ON AUGUST 18: VBMS requires postal codes for Germany and Caseflow does not


leikkisa:
Discovered this bug with this Sentry error:
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/1917/

The Veteran model only requires USA and Canada to have zip codes in validation.

leikkisa:
@abbyraskinUSDS  @suzchap Can one of you check this out and see what is required, and if we are missing any other countries?  If we don't know, let me know and I can ask VBMS knows.

abbyraskinUSDS:
Hm yeah, this seems like a VBMS question.

abbyraskinUSDS:
phoning a friend named @aroltsch who dealt with something similar for Dispatch a long time ago...

aroltsch:
It appears to me other countries require postal code. This error is for Germany and zip code we are sending is nil.

```
:country=>"Germany", :zip_code=>nil
```

However, I don't remember dealing with it personally. I found a few tickets Shane files with VBMS team:
https://github.com/department-of-veterans-affairs/dsva-vbms/issues/77
https://github.com/department-of-veterans-affairs/dsva-vbms/issues/80

leikkisa:
I am double checking with VBMS if we just need to add Germany (and potentially other countries) to the list of countries requiring postal code in the Veteran model.  

I looked at the other tickets and I think this is unrelated to the OVR address type issue because it's an INT address type and I don't think it's a military address (the Veteran is not in the military anymore and I didn't see any of those other codes).

leikkisa:
VBMS checked and it appears that Germany is allowed to come through without a zip code in V5, but for some reason we were still using v4. Some things to check:
- Are we setting it to use V5 when we make the request?
- Is there a V5 feature flag that needs to be turned on in prod?

leikkisa:
Happening for Australia too.

kahlouie:
Sally has emailed VBMS to see if the bug fix on their side got deployed this weekend. Will get a pr up for the revert with the blocked label until we get confirmation.

PR #6488, opened 2018-08-02
Upcase countries for postal code validation

leikkisa:
Connects #6285

We got another error on this, and I checked the country for the Veteran that had an error, and it doesn't look uppercased.  It looks like for USA we manually set the country to "USA" so it didn't matter then.

Hopefully this is a real temporary fix!

Still will plan on removing this after the VBMS bug is fixed.


app/models/veteran.rb
```diff
@@ -34,7 +34,7 @@ def name
   end
 
   def country_requires_zip?
-    COUNTRIES_REQUIRING_ZIP.include?(country)
+    COUNTRIES_REQUIRING_ZIP.include?(country.upcase)
```
kahlouie: Looks like this is just a fix for zipcode, but if country is not all uppercase, it won't require a state either. Can we `#upcase` it there as well? Otherwise, looks great!

app/models/veteran.rb
```diff
@@ -34,11 +34,11 @@ def name
...
   end
 
   def country_requires_state?
-    country == "USA"
+    country.upcase == "USA"
```
kahlouie: Nice!!! 💃

PR #6496, opened 2018-08-03
Queue Colocated | Adds type column

evankroske:
Fixes https://github.com/department-of-veterans-affairs/caseflow/issues/6077

## Description
Adds the "Type(s)" column to the task table for colocated admins and refactors renderAppealType.

## Testing Plan
- [x] Become Bvalsporer.
- [x] Go to /queue.
- [x] Confirm that all tasks have something in the column "Type(s)".
- [x] Confirm that one task has a red "AOD" in the types column.


app/models/appeal.rb
```diff
@@ -25,7 +25,11 @@ def self.find_appeal_by_id_or_find_or_create_legacy_appeal_by_vacols_id(id)
   end
 
   def type
-    [advanced_on_docket? ? "AOD" : nil].compact.join(" ,")
+    "Original"
+  end
+
+  def aod
```
mdbenjam: Could we flip this? Instead of making a method on appeal called aod, could we make a method on legacy appeal called advanced_on_docket? I think spelling it out adds clarity.

evankroske: Good point. Done!

PR #6498, opened 2018-08-03
Flow | Annotates QueueApp

evankroske:
### Description
Annotates types of QueueApp and removes redundant propTypes.


client/app/queue/QueueApp.jsx
```diff
@@ -108,13 +129,31 @@ class QueueApp extends React.PureComponent {
...
+  routedOrganization = (props) => <QueueLoadingScreen
+    {...this.propsForQueueLoadingScreen()}
+    urlToLoad={`${props.location.pathname}/tasks`}>
     <SearchBar feedbackUrl={this.props.feedbackUrl} />
     <OrganizationQueue {...this.props} />
   </QueueLoadingScreen>
 
   queueName = () => this.props.userRole === USER_ROLES.ATTORNEY ? 'Your Queue' : 'Review Cases';
 
+  propsForQueueLoadingScreen = () => {
```
mdbenjam: Does this function add clarity?

evankroske: I think that passing all props, what we were doing, was dangerous. Any time that you added or changed a prop of QueueApp, it could break Queue. I wanted to stop passing all props, but I also didn't want to pass the same four arguments to every call to QueueLoadingScreen, so I think that this is a practical solution.

dannysellers: I completely agree that passing `...this.props` is a bad idea, but it seems like this function does reduce clarity to a degree. `userRole` and `userCssId` are already set in Redux to avoid [prop drilling](https://blog.kentcdodds.com/prop-drilling-bb62e02cb691)—is there any reason not to add `userRole` or `userCanAccessQueue` to Redux as well?

evankroske: Yep, I agree that adding userRole and userCanAccessQueue is a great idea! I'll think about making that change the next time I've got some time for a refactoring

PR #6503, opened 2018-08-03
Anya/6493 ama attorney checkout

### LINKED ISSUE #6493: Attorney can complete attorney checkout for ama appeals

aroltsch:
Closes #6493 

1. Attorney should be able to complete attorney checkout for AMA appeals by updating attorney_case_reviews table and updating issue dispositions. 
2. Remand reasons are not part of this PR
3. When attorney completes draft decision, task should be marked as completed. This will be implemented in the next PR.


app/controllers/case_reviews_controller.rb
```diff
@@ -17,7 +17,11 @@ def complete
     return invalid_record_error(record) unless record.valid?
 
     response = { task: record }
-    response[:issues] = record.appeal.issues
+    response[:issues] = if record.legacy?
+                          record.appeal.issues
+                        else
+                          record.appeal.decision_issues
```
aroltsch: Ideally we would like to have `issues` method on AMA appeal that returns both tables. What were your thoughts on this?

app/models/judge_case_assignment_to_attorney.rb
```diff
@@ -1,10 +1,10 @@
 class JudgeCaseAssignmentToAttorney
   include ActiveModel::Model
-  include LegacyTaskConcern
```
aroltsch: I've renamed LegacyTaskConcern to be CaseReviewConcern because most of functionality belongs to the case review models

app/models/concerns/case_review_concern.rb
```diff
@@ -1,22 +1,21 @@
-module LegacyTaskConcern
+module CaseReviewConcern
   extend ActiveSupport::Concern
 
-  included do
-    # task ID is vacols_id concatenated with the date assigned
-    validates :task_id, format: { with: /\A[0-9A-Z]+-[0-9]{4}-[0-9]{2}-[0-9]{2}\Z/i }, allow_blank: true
```
aroltsch: Removing the validation because now task_id could also be Caseflow DB id

app/controllers/case_reviews_controller.rb
mdbenjam: request and decision issues? We'll need to definitely return both. Maybe an issues method that returns a hash like: 
```
{ decision_issues: [], request_issues: [] }
```

client/app/queue/utils.js
```diff
@@ -179,8 +179,9 @@ export const buildCaseReviewPayload = (
   }
 
   payload.data.tasks.issues = getUndecidedIssues(issues).map((issue) => _.extend({},
-    _.pick(issue, ['vacols_sequence_id', 'remand_reasons', 'type', 'readjudication']),
-    { disposition: _.capitalize(issue.disposition) }
+    _.pick(issue, ['remand_reasons', 'type', 'readjudication']),
+    { disposition: _.capitalize(issue.disposition) },
+    { id: issue.vacols_sequence_id }
```
mdbenjam: shouldn't this be `issue.id`? Since we're now passing it back as id not `vacols_sequence_id`? Might be wrong on this.

app/models/judge_case_assignment_to_attorney.rb
```diff
@@ -32,8 +32,12 @@ def reassign_to_attorney!
     end
   end
 
+  def created_in_vacols_date
```
mdbenjam: Is this used anywhere?

client/constants/ISSUE_DISPOSITIONS.json
```diff
@@ -0,0 +1,10 @@
+{
+    "allowed": "Allowed",
```
mdbenjam: Should sync with Alex or Chris about what dispositions they'd like to see in AMA (unless this list already came from them :D)

app/controllers/case_reviews_controller.rb
aroltsch: Good idea.

app/models/judge_case_assignment_to_attorney.rb
aroltsch: Yes. I've moved it here from LegacyTaskConcern.

client/app/queue/utils.js
aroltsch: For legacy appeals, this is correct because we are returning vacols_sequence_id for issues. However, FE changes will need to happen for AMA attorney checkout.

client/constants/ISSUE_DISPOSITIONS.json
aroltsch: Good point!

aroltsch: @amprokop or @cmgiven could you chime in? Is this list accurate for AMA appeals.

cmgiven: I don't know, but I do think Alex has done some work on that one. If not, Lauren?

laurjpeterson: These ones are still accurate for AMA ([slack thread](https://dsva.slack.com/archives/CA1QBDLEL/p1530293085000005
)):
"1" => "Allowed",
"3" => "Remanded",
"4" => "Denied",
"5" => "Vacated",
"6" => "Dismissed, Other",
"8" => "Dismissed, Death",
"9" => "Withdrawn",

Emily Tamlyn raised a few dispositions that are in VACOLS but not in our list yet: Merged Appeal, Motion to Vacate Denied, Motion to Vacate w/d, reconsideration by letter, and stay ([slack thread](https://dsva.slack.com/archives/CA1QBDLEL/p1530295928000258)). But, these additional dispositions are types of legal action with the dispositions attached. 

I want to verify that these additional dispositions are necessary at the issue level, or if they are at the appeal level.

Can we move forward with the original list, and I can update the requirements after resurfacing this with LRP?

aroltsch: I will stick to the above list for now, we can update it any time.

app/controllers/case_reviews_controller.rb
```diff
@@ -17,11 +17,7 @@ def complete
     return invalid_record_error(record) unless record.valid?
 
     response = { task: record }
-    response[:issues] = if record.legacy?
-                          record.appeal.issues
-                        else
-                          record.appeal.decision_issues
-                        end
+    response[:issues] = record.appeal.issues
```
mdbenjam: Love it.

client/app/queue/utils.js
mdbenjam: Ah I see, you changed that the frontend passes it back as `id`, but we still receive it as `vacols_sequence_id`

spec/factories/decision_issue.rb
```diff
@@ -0,0 +1,7 @@
+FactoryBot.define do
+  factory :decision_issue do
+    request_issue
+    disposition { Constants::ISSUE_DISPOSITIONS.keys.sample }
```
mdbenjam: I'm hesitant about adding randomness to FactoryBot. Can we just make this a static disposition?

mdbenjam: Or if this is not a required field we can just leave it off. One of the FactoryBot suggestions is to leave your factories as basic as possible.

aroltsch: Great point sir

Issue #6467, opened 2018-08-02
Return more detailed info about VACOLS appeals from /idt/api/v1/appeals/:appeal_id


amprokop:
# AC
For an appeal requested from /idt/api/v1/appeals/:appeal_id, return the following information:

- Veteran name: separate fields for Veteran first, middle, last, and suffix
- Veteran gender
- Veteran alive/deceased
- Boolean representing if the appellant is the veteran
- Appellant name: (if different from Veteran) separate fields for first, middle, last and suffix
- Claim Number (SS or C)
- Docket Number

- Number of issues
- Array of issues
- Issue Key
- Issue Program Code and Description
- Issue Code Code and Description
- Issue Level1, Level2, Level3 Codes and Descriptions
- Issue Description
- Status (ACT, ADV, etc)

- AOD Status
- CAVC remand status
- Public citation number

- Representative
- Type (Attorney, Agent, VSO, special)
- First, Middle, Last, Suffix - or - VSO name

- Document Numbers for all documents for an appeal
- Outstanding mail

PR #6505, opened 2018-08-03
IDT appeal details endpoint

amprokop:
Connects https://github.com/department-of-veterans-affairs/caseflow/issues/6467

### Description
Implements most appeal details functionality for the IDT API. Future PRs will include document ids from DECASS, outstanding mail, and a refactor of the POA functionality.

### Testing
- `/idt/api/v1/appeals -H TOKEN:dfsafsa` should continue to return a list of appeals assigned to the user
- `/idt/api/v1/appeals/:appeal_id -H TOKEN:dfsafsa` should return more detailed information for a single appeal


app/models/certification_stats.rb
```diff
@@ -3,6 +3,7 @@
 # it is responsible for aggregating and caching statistics.
 #
 class CertificationStats < Caseflow::Stats
+  # :nocov:
```
amprokop: adding this because this file was failing code coverage. this file is not quite deprecated, but not worth adding tests for.

config/routes.rb
```diff
@@ -37,7 +37,6 @@
       namespace :v1 do
         get 'token', to: 'tokens#generate_token'
         get 'appeals', to: 'appeals#index'
-        get 'appeal_details/:appeal_id', to: 'appeals#details'
```
amprokop: decided against using this name for the route

app/controllers/idt/api/v1/appeals_controller.rb
```diff
@@ -2,14 +2,17 @@ class Idt::Api::V1::AppealsController < Idt::Api::V1::BaseController
...
     appeals = file_number ? appeals_by_file_number : appeals_assigned_to_user
 
     render json: json_appeals(appeals)
   end
 
   def details
-    # TODO: implement
+    # TODO: add AMA appeals
+    tasks, appeals = LegacyWorkQueue.tasks_with_appeals_by_appeal_id(params[:appeal_id], "attorney")
+    return render json: {}, status: 404 if appeals.empty?
```
lowellrex: I know this is for the IDT, but it might be helpful to have a message in case we are ever called in to troubleshoot the IDT, so we can track down where the error came from.

amprokop: sure i can add that

app/controllers/idt/api/v1/appeals_controller.rb
```diff
@@ -2,14 +2,17 @@ class Idt::Api::V1::AppealsController < Idt::Api::V1::BaseController
...
+    render json: json_appeal_details(tasks[0], appeals[0])
```
lowellrex: Do we need to protect against the scenario where no tasks are returned as well? I'm not super familiar with this code and it looks like `LegacyWorkQueue.tasks_with_appeals_by_appeal_id()` returns an empty array for both tasks and appeals if tasks is empty and we will return a 404 in that case, so I guess we're fine.

lowellrex: Actually, maybe it makes sense to include this check on line 14. Something like:
```
return render json: {}, status: 404 if appeals.empty? || tasks.empty?
```

amprokop: That's actually a good catch — I think we should probably still return the appeal even if there are no tasks for it.

Issue #6470, opened 2018-08-02
Queue Colocated | Add sorting


evankroske:

evankroske:
Fixed by https://github.com/department-of-veterans-affairs/caseflow/pull/6510

PR #6510, opened 2018-08-06
Queue Colocated | Add sorting

evankroske:
Connects https://github.com/department-of-veterans-affairs/caseflow/issues/6470

## Description
Makes the columns of the colocated queue table sortable.

## Testing Plan
- [x] Become BVALSPORER.
- [x] View their queue.
- [x] Confirm that the AOD case is the first case in the table.
- [x] Sort the column "Type(s)".
- [x] Confirm that the AOD case is now last.
- [x] Sort the column "Case Details" by ascending.
- [x] Confirm that the veteran names are in ascending alphabetical order.

### DIFF
--- a/client/app/queue/components/AmaTaskTable.jsx
+++ b/client/app/queue/components/AmaTaskTable.jsx
@@ -37,7 +37,8 @@ class AmaTaskTable extends React.PureComponent<Props> {
 
   caseTaskColumn = () => ({
     header: COPY.CASE_LIST_TABLE_TASKS_COLUMN_TITLE,
-    valueFunction: (task: AmaTask) => CO_LOCATED_ADMIN_ACTIONS[task.attributes.title]
+    valueFunction: (task: AmaTask) => CO_LOCATED_ADMIN_ACTIONS[task.attributes.title],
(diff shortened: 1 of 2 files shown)


client/app/queue/components/AmaTaskTable.jsx
```diff
@@ -30,32 +30,44 @@ class AmaTaskTable extends React.PureComponent<Props> {
       header: COPY.CASE_LIST_TABLE_VETERAN_NAME_COLUMN_TITLE,
       valueFunction:
         (task: AmaTask) => <a href={`/queue/appeals/${task.attributes.external_id}`}>
```
dannysellers: I know it's out of scope, but is there any reason not to use a `Link` here? With a basic `a`, we lose SPA behavior / have to reload all the data

evankroske: The case details page doesn't yet support AMA tasks, so I didn't want to take the change that visiting the case details page would corrupt the state and break the back button. This way, even if someone opens the case details page, the queue page will still work. Does that make sense?

PR #6518, opened 2018-08-06
Queue Colocated | Format veteran names as first name last name

evankroske:
Fixes https://github.com/department-of-veterans-affairs/caseflow/issues/6509

### Description
Formats veteran names as "first name middle initial last name".

### Testing Plan
- [ ] Become BVALSPORER.
- [ ] Go to /queue.
- [ ] Confirm that veteran names have the form "first name middle initial last name".

### DIFF
--- a/client/app/queue/components/AmaTaskTable.jsx
+++ b/client/app/queue/components/AmaTaskTable.jsx
@@ -25,12 +25,22 @@ type Props = Params;
 class AmaTaskTable extends React.PureComponent<Props> {
   getKeyForRow = (rowNumber, task) => task.id
 
+  formattedVetNameOfTask = (task: AmaTask) => {
+    if (!task.attributes.veteran_name) {
+      return '';
+    }
+
+    const [ln, fn, mi] = task.attributes.veteran_name.split(', ')
+
+    return _.compact([fn, mi, ln]).join(' ');
+  }
+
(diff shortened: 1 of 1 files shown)


lowellrex:
Why do we have this problem for `amaTasks` we pull from the global state and not `appeals`?

Perhaps it has something to do with the formatting we do in the [appeal model](https://github.com/department-of-veterans-affairs/caseflow/blob/78a9d2c59ff07dbe1a01a6f1205c8432004c6a38/app/models/appeal.rb#L41)? If that is the case, maybe it would be more robust to make a new method (or change the existing `veteran_name` method) to use the [`:readable_full` format](https://github.com/department-of-veterans-affairs/caseflow/blob/78a9d2c59ff07dbe1a01a6f1205c8432004c6a38/app/models/full_name.rb#L25) that already exists and I does--I think--the same thing on the backend that you're getting after on the front-end here.

lowellrex:
Looks like we do this name formatting in the serializer for the attorney and judge queues: https://github.com/department-of-veterans-affairs/caseflow/blob/78a9d2c59ff07dbe1a01a6f1205c8432004c6a38/app/models/serializers/work_queue/appeal_serializer.rb#L40

evankroske:
@lowellrex , good point. I'll make this change on the backend, instead

Issue #6524, opened 2018-08-06
Ignore errors from new_documents endpoint


lowellrex:
The `/appeals/{ID}/new_documents` endpoint occasionally triggers errors (for example, [this one](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2129/) is the result of a 504 response). Since the call the the `new_documents` endpoint is a user experience enhancement and not critical to the functioning of the application, we should treat errors at this endpoint as non-critical.

## Acceptance criteria
* Errors resulting from requests to the `new_documents` endpoint are not sent to sentry/slack
* `new_documents` endpoint errors *are* recorded as a metric in datadog

If it is possible to know that these errors are non-critical when we throw them, it would be nice to indicate on the error object itself that the error is not critical so we can later decide how to handle the error based on its non-criticality instead of its type (similar to [what we already do](https://github.com/department-of-veterans-affairs/caseflow/blob/cc4ff55e52d8200363d23db5e69fd0a4d99625bf/app/controllers/reader/documents_controller.rb#L23) with serializeable errors).

lowellrex:
For reference, Dave Cheney has a presentation with some [best practices for error handling](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully) that I find myself returning to frequently.

lowellrex:
Completed by #6528.

PR #6528, opened 2018-08-06
Ignore non-showstopping errors to document count and new files endpoints

lowellrex:
Connects #6524.

Errors resulting from requests to the `new_documents` and `document_count` endpoints are not sent to sentry/slack.


app/controllers/appeals_controller.rb
```diff
@@ -111,6 +111,14 @@ def file_number_not_found_error
     }, status: 400
   end
 
+  def handle_non_fatal_error(err)
```
lowellrex: I think this is the place to fire off a request to DataDog (and Prometheus?) to let it know that we have seen another of these events. However, I don't really know how I'd go about doing that (and couldn't find any examples in our codebase), so if you've got any ideas I'd love to hear 'em!

lib/caseflow/error.rb
```diff
@@ -8,10 +8,11 @@ def initialize(args)
     end
 
     def serialize_response
-      { json: { "errors": ["status": code, "title": message, "detail": message] }, status: code }
+      { json: { "errors": [{ "status": code, "title": message, "detail": message }] }, status: code }
```
mdbenjam: Can we get rid of the [] if we're making this an object? How did this work before... hmmm

lowellrex: image

It appears as if we generally expect errors to be an array in other parts of our front-end code (I think this is somewhat of a standard for http responses that return errors, maybe?) so I think it would be best to return an array here (even though it will only have one element) to maintain parity with the rest of our code.

I think this worked before because we never did anything with the error objects we pulled out (which I think would just be `{status: code}`, right?). `NewFile` [accesses the error variable](https://github.com/department-of-veterans-affairs/caseflow/blob/93609b0f0ef27e37128dd165071333ee4639eabd/client/app/queue/components/NewFile.jsx#L44) in the global state but never does anything with it as far as I can tell. `AppealDocumentCount` [makes no attempt](https://github.com/department-of-veterans-affairs/caseflow/blob/93609b0f0ef27e37128dd165071333ee4639eabd/client/app/queue/AppealDocumentCount.jsx#L32) to look at any errors it might have received. I'm not entirely sure how Reader wasn't bitten by this before.

app/controllers/appeals_controller.rb
lowellrex: I prefer handling these errors in this controller because I think it is more clear that we are effectively ignoring them because of the two endpoints that are being accessed, and not because of the type of error that is being generated.

lib/caseflow/error.rb
mdbenjam: Thanks for the thorough explanation!

app/controllers/appeals_controller.rb
```diff
@@ -111,6 +111,23 @@ def file_number_not_found_error
     }, status: 400
   end
 
+  def handle_non_critical_error(endpoint, err)
+    if !err.class.method_defined? :serialize_response
+      err = Caseflow::Error::SerializableError.new(code: 500, message: err.to_s)
+    end
+
+    DataDogService.increment_counter(
+      metric_group: "business",
```
mdbenjam: Not sure this is best classified as a business metric. Maybe make `errors` the group?

lib/caseflow/error.rb
lowellrex: You're welcome!

PR #6537, opened 2018-08-07
Fixing extra vacols calls

mdbenjam:
### Description
The frontend refactor started making a lot of VACOLS requests. This is because the legacy_task does not memoize its relationship with appeal. And while we grab appeals in bulk we need to assign them to the task model preemptively.

### Acceptance Criteria 
- [ ] Should not see any `load_vacols_data` calls when loading a Queue.

### DIFF
--- a/app/models/legacy_tasks/legacy_task.rb
+++ b/app/models/legacy_tasks/legacy_task.rb
@@ -45,7 +45,8 @@ class LegacyTask
       assigned_to: user,
       task_id: record.created_at ? record.vacols_id + "-" + record.created_at.strftime("%Y-%m-%d") : nil,
       document_id: record.document_id,
-      assigned_by: record.assigned_by
+      assigned_by: record.assigned_by,
+      appeal: appeal
     )
   end
 
--- a/app/models/legacy_tasks/legacy_task.rb
+++ b/app/models/legacy_tasks/legacy_task.rb
@@ -6,6 +6,7 @@ class LegacyTask
            :docket_date, :added_by, :task_id, :type, :document_id, :assigned_by, :work_product].freeze
 
   attr_accessor(*ATTRS)
+  attr_writer :appeal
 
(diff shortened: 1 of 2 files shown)


app/models/queues/legacy_work_queue.rb
```diff
@@ -37,8 +37,11 @@ def tasks_with_appeals_of_vacols_tasks(user, role, vacols_tasks)
       vacols_appeals = repository.appeals_from_tasks(vacols_tasks)
 
       tasks = vacols_tasks.zip(vacols_appeals).map do |task, appeal|
-        MODEL_CLASS_OF_ROLE[role.capitalize].from_vacols(task, appeal, user)
+        return_task = MODEL_CLASS_OF_ROLE[role.capitalize].from_vacols(task, appeal, user)
```
aroltsch: Should we set the appeal in from_vacols method?

mdbenjam: I'd like to leave it here for now since it's only one place instead of two?

aroltsch: I was suggesting to move it there

mdbenjam: Oh I get in the base class's from_vacols. Sure that makes sense 👍

app/models/legacy_tasks/legacy_task.rb
```diff
@@ -44,7 +45,8 @@ def self.from_vacols(record, appeal, user)
       assigned_to: user,
       task_id: record.created_at ? record.vacols_id + "-" + record.created_at.strftime("%Y-%m-%d") : nil,
       document_id: record.document_id,
-      assigned_by: record.assigned_by
+      assigned_by: record.assigned_by,
+      appeal: appeal
```
aroltsch: Nice!

PR #6539, opened 2018-08-07
Queue | Return colocated tasks from AppealsController.tasks

### LINKED ISSUE #6517: Queue | Server Error loading Case Details view as Colocated user
As a colocated user (`BVALSPORER`), going from `/queue` to `/queue/appeals/${appealId}` causes a server error within [`LegacyWorkQueue.tasks_with_appeals_of_vacols_tasks`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/queues/legacy_work_queue.rb#L36): `undefined method 'from_vacols' for nil:NilClass`.

[slack thread](https://dsva.slack.com/archives/C6E41RE92/p1533579267000245)

dannysellers:
Resolves #6517 

### Description
When loading the case detail view as a colocated user, we hit `/appeals/${appealId}/tasks?role=Colocated`, which would attempt and fail to get legacy tasks from `LegacyWorkQueue`, because `ColocatedTask`s have no `from_vacols` method / are post-VACOLS.

This change returns `ColocatedTask`s for colocated users when visiting `/queue/appeals/${appealId}`. It also renames the Flow type `Task` to `LegacyTask`

### Acceptance Criteria 
- [ ] `http://localhost:3000/appeals/${appealId}/tasks?role=Colocated` returns tasks

### Testing Plan
1. Go to `/queue` as `BVALSPORER`, navigate to case detail view with dev tools open
2. Confirm response to `/appeals/${appealId}/tasks?role=Colocated` has task data (no UI implemented yet)


app/models/serializers/work_queue/task_serializer.rb
```diff
@@ -11,6 +11,11 @@ class WorkQueue::TaskSerializer < ActiveModel::Serializer
   attribute :placed_on_hold_at
   attribute :instructions
   attribute :appeal_type
+  attribute :previous_task do
```
dannysellers: I added this both to match `LegacyTaskSerializer`, and because `queue/utils.prepareTasksForStore` threw an error trying to set [`previousTaskAssignedOn: task.attributes.previous_task.assigned_on`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/utils.js#L44)

app/controllers/appeals_controller.rb
```diff
@@ -111,16 +113,39 @@ def file_number_not_found_error
     }, status: 400
   end
 
+  def json_colocated_tasks_by_appeal_id(appeal_id)
```
aroltsch: In theory, this endpoint will be used by ama appeals as well so we should make it generic for Caseflow tasks. Here is the logic: 
* if role is [attorney or judge] and if appeal is legacy, call legacy queue
* in every other case call queue class based on a passed role, similar to https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/tasks_controller.rb#L12.

If we want to move this logic to FE, we can rename `tasks` to be `legacy_tasks` and create `tasks` for our caseflow tasks.

app/models/serializers/work_queue/task_serializer.rb
aroltsch: Unfortunately, this will be wrong. We've added this field for a judge legacy task only to tell a judge when a decass record was assigned to the attorney. For AMA appeals, we would return children for a judge based on the parent_id. It would be better if we omit this field for task_serializer for now?

app/controllers/appeals_controller.rb
dannysellers: I'll rename `Tasks` -> `LegacyTasks` on the FE in a separate PR

app/controllers/appeals_controller.rb
```diff
@@ -36,10 +36,17 @@ def tasks
...
     return invalid_role_error unless ROLES.include?(role)
-    tasks, = LegacyWorkQueue.tasks_with_appeals_by_appeal_id(params[:appeal_id], role)
-    render json: {
-      tasks: json_tasks(tasks)[:data]
-    }
+
+    if %w[attorney judge].include?(role)
+      json_tasks_by_legacy_appeal_id_and_role(params[:appeal_id], role)
+    else
+      appeal_db_id = if Appeal::UUID_REGEX.match(params[:appeal_id])
```
aroltsch: Danny, this controller already has this logic https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/appeals_controller.rb#L102. Take a look at the method.

app/controllers/appeals_controller.rb
```diff
@@ -36,10 +36,17 @@ def tasks
...
+                       Appeal.all.where(uuid: params[:appeal_id]).pluck(:id).first
+                     else
+                       LegacyAppeal.all.where(vacols_id: params[:appeal_id]).pluck(:id).first
+                     end
+      json_tasks_by_appeal_id(appeal_db_id)
+    end
```
aroltsch:
```
if %w[attorney judge].include?(role) && appeal.class == "LegacyAppeal"
    json_tasks_by_legacy_appeal_id_and_role(params[:appeal_id], role)
    return
end
json_tasks_by_appeal_id(appeal_db_id)

File: app/controllers/appeals_controller.rb [hunk 5]
```diff
@@ -111,6 +118,26 @@ def file_number_not_found_error
     }, status: 400
   end
 
+  def queue_class
+    TasksController::QUEUES[params[:role].downcase.try(:to_sym)]
+  end
+
+  def json_tasks_by_appeal_id(appeal_db_id)
+    tasks = queue_class.new.tasks_by_appeal_id(appeal_db_id)
```
aroltsch: Nice!

app/controllers/appeals_controller.rb
dannysellers: Mmm, I guess we do need the check for `LegacyAppeal`, yeah. Thanks

app/models/queues/colocated_queue.rb
```diff
@@ -4,6 +4,16 @@ class ColocatedQueue
   attr_accessor :user
 
   def tasks
-    ColocatedTask.where.not(status: "completed").where(assigned_to: user)
+    incomplete_tasks.where(assigned_to: user)
+  end
+
+  def tasks_by_appeal_id(appeal_id)
+    incomplete_tasks.where(appeal_id: appeal_id, appeal_type: "Appeal")
```
dannysellers: `ColocatedTask`s have `appeal_id` and `appeal_type` attrs, where `appeal_id` is the database id of the appeal, rather than it's `vacols_id`/`uuid`. Given 2 appeals both with `id: 1` and
```ruby
ColocatedTask(appeal_id: 1, appeal_type: "LegacyAppeal", vacols_id: "123456")
ColocatedTask(appeal_id: 1, appeal_type: "Appeal", uuid: "...")
```
`incomplete_tasks.where(appeal_id: appeal_id)` might return too many tasks, unless we also pass `appeal_type`. I'm not sure whether this is an issue in reality or just in testing, where I've [created ColocatedTasks for an Appeal and a LegacyAppeal](https://github.com/department-of-veterans-affairs/caseflow/blob/029e1220a5b618148d35bfb2ef7ee29972e8c7f3/spec/controllers/appeals_controller_spec.rb#L147), both of which have a db id of 1.

We call this function for users who aren't attorneys or judges who want to get tasks for a LegacyAppeal, so we do need to differentiate somehow...

app/controllers/appeals_controller.rb
```diff
@@ -36,10 +36,12 @@ def tasks
 
     role = params[:role].downcase
     return invalid_role_error unless ROLES.include?(role)
-    tasks, = LegacyWorkQueue.tasks_with_appeals_by_appeal_id(params[:appeal_id], role)
-    render json: {
-      tasks: json_tasks(tasks)[:data]
-    }
+
+    if %w[attorney judge].include?(role) && appeal.class == LegacyAppeal
```
aroltsch: Ooops Danny, it should be `appeal.class.name`

client/app/queue/utils.js
```diff
@@ -22,35 +24,39 @@ import DIAGNOSTIC_CODE_DESCRIPTIONS from '../../constants/DIAGNOSTIC_CODE_DESCRI
...
-        assignedByFirstName: task.attributes.assigned_by_first_name,
-        assignedByLastName: task.attributes.assigned_by_last_name,
-        workProduct: task.attributes.work_product,
-        previousTaskAssignedOn: task.attributes.previous_task.assigned_on
-      };
-    });
-
-    return _.pickBy(_.keyBy(mappedLegacyTasks, (task) => task.externalAppealId), (task) => task);
-  };
+export const prepareTasksForStore = (tasks: Array<Object>): AmaTasks => tasks.reduce((acc, curr: AmaTask) => {
```
dannysellers: For some reason Flow didn't like me using `_.keyBy` here :\

Also, `AmaTask` still has the old format (parameterized attr names within `attributes`), so I didn't alter them here yet

Issue #6078, opened 2018-07-02
Co-located table | New tab


laurjpeterson:
As a co-located user, I want to distinguish which cases and tasks are newly assigned to me, from those on hold or those that are pending action to come off of hold. 

## AC
- Establish the 'New' tab
- Verify newly assigned case and tasks exist in the co-located table in the new tab

### User Stories
- When a co-located team member is planning what to work on, they need to see all tasks assigned to them (mail, VACOLS, 3230s, etc.) in one place, so they can prioritize what to work on.
  - These tasks include appeals assigned to them to work AND appeals that they placed on hold.

- When a co-located team member is planning what to work on, I need to see what type of tasks are assigned to me, so I can prioritize what to work on.
  - e.g. task type column
  - this task type would initially come from the VACOLS field, and in the future come from when attorneys assign the tasks in queue

##### New Tab
![colocated-table-new-tab-single-actions-768](https://user-images.githubusercontent.com/31224257/43283661-80eb7dae-90e7-11e8-9f71-01abb998f958.png)

##### New Tab with Tooltips
![colocated-table-new-tab-single-actions-wtooltips-768](https://user-images.githubusercontent.com/31224257/43283662-81092944-90e7-11e8-95ca-4bbfcc7a9cd1.png)

mkhandekar:
I've updated the link to the design under "mocks" in the original issue.

PR #6548, opened 2018-08-08
Co-located table | New tab

evankroske:
Fixes https://github.com/department-of-veterans-affairs/caseflow/issues/6078

### Description
Adds a tab to the colocated queue page that includes only new tasks.

### Testing Plan
- [x] Become BVALSPORER.
- [x] Go to /queue.
- [x] Confirm that only 4 tasks are shown.


client/app/queue/selectors.js
```diff
@@ -37,9 +38,14 @@ export const tasksByAssigneeCssIdSelector = createSelector(
...
 
-export const amaTasksByAssigneeId: (State) => { [number]: AmaTask } = createSelector(
-  [getAmaTasks],
-  (tasks: AmaTasks) => _.groupBy(tasks, (task) => task.attributes.assigned_to.id)
+export const amaTasksAssignedTo: (State, { userId: number }) => { [number]: AmaTask } = createSelector(
+  [getAmaTasks, getUserId],
+  (tasks: AmaTasks, userId: number) => _.filter(tasks, (task) => task.attributes.assigned_to.id === userId)
+);
+
+export const amaTasksNewAssignedTo: (State, { userId: number }) => { [number]: AmaTask } = createSelector(
```
dannysellers: It looks like we're passing `userId` as an argument here, which is just threaded down from Rails template -> `QueueApp` -> `ColocatedTaskListView` -> `NewTasksTab` (i.e. prop drilling). Can we not just pull it from Redux / use the `getUserId` selector?

client/app/queue/selectors.js
```diff
@@ -37,9 +38,14 @@ export const tasksByAssigneeCssIdSelector = createSelector(
   )
 );
 
-export const amaTasksByAssigneeId: (State) => { [number]: AmaTask } = createSelector(
-  [getAmaTasks],
-  (tasks: AmaTasks) => _.groupBy(tasks, (task) => task.attributes.assigned_to.id)
+export const amaTasksAssignedTo: (State, { userId: number }) => { [number]: AmaTask } = createSelector(
```
dannysellers: I think it's redundant to have both `userId` as an argument and `getUserId`. Also, I don't think either of these need `State` as an argument here

client/app/queue/selectors.js
evankroske: OK, I'll add userId to the state and get it from there

dannysellers: Oh, it's not there yet, I see. I think I was looking at `userCssId`... But yeah, I'd say we should do that for all properties accessed by [`QueueApp.propsForQueueLoadingScreen `](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/QueueApp.jsx#L142) (not necessarily all within this PR)

client/app/queue/selectors.js
evankroske: Now that I'm planning to add userId to the state, I'll remove the userId argument.

Regarding the state parameter, I thought that selectors had to take the state as the first argument. I forgot to pass the state initially, and I don't think the code worked then. But I'm not an expert on selectors. Am I missing something?

dannysellers: That is right, though `createSelector` returns a (memoized) function that takes `state, props` as arguments, you don't need to wrap it in another function.

The [docs](https://github.com/reduxjs/reselect#connecting-a-selector-to-the-redux-store) are really verbose / the examples are a bit complex, but within example `mapStateToProps` fns, you can see that they're directly calling the function returned by `createSelector`

client/app/queue/selectors.js
```diff
@@ -25,6 +25,7 @@ const getAmaTasks = (state: State) => state.queue.amaTasks;
 const getAppeals = (state: State) => state.queue.appeals;
 const getAppealDetails = (state: State) => state.queue.appealDetails;
 const getUserCssId = (state: State) => state.ui.userCssId;
+const getUserId = (state: State, props: Object) => props.userId;
```
dannysellers: This should be
```javascript
const getUserId = (state: State) => state.ui.userId;
```
or wherever it ends up

client/app/queue/selectors.js
evankroske: I decided to filter the tasks by userCssId, instead. I think that we might as well keep the new selectors consistent with the existing ones.

client/app/queue/selectors.js
evankroske: I've removed the unnecessary parameter

client/app/queue/selectors.js
evankroske: I switched to using getUserCssId and deleted this selector

client/app/queue/ColocatedTaskListView.jsx
```diff
@@ -31,26 +29,33 @@ class ColocatedTaskListView extends React.PureComponent<Props> {
...
+  }
+
   render = () => {
-    const tableContent = <div>
-      <h1 {...fullWidth}></h1>
-      <AmaTaskTable tasks={this.props.amaTasks} />
-    </div>;
+    const tabs = [{
+      label: 'New',
+      page: this.pageNew()
```
dannysellers: Nit: this might as well just be `page: <NewTasksTab />` at this point

client/app/queue/ColocatedTaskListView.jsx
```diff
@@ -7,21 +7,19 @@ import AmaTaskTable from './components/AmaTaskTable';
...
 } from './selectors';
 import { clearCaseSelectSearch } from '../reader/CaseSelect/CaseSelectActions';
+import TabWindow from '../components/TabWindow';
+import COPY from '../../COPY.json';
 
-import { fullWidth } from './constants';
 import type { AmaTask } from './types/models';
 import type { State } from './types/state';
 
 type Params = {|
```
dannysellers: Do we still need this?

evankroske: Nope. I've removed it

client/app/queue/ColocatedTaskListView.jsx
evankroske: Fixed!

PR #6553, opened 2018-08-08
Update brieff table from decass when bva dispatch checkout

aroltsch:
It turns out we are missing some field updates on the brieff table during a judge checkout. We also not updating judge board member id on decass

AC:
* Update decass.dememid when judge charges the case to 4E and then set it to the Judges staff.sattyid
* When the DAS final review is done by the VLJ and the case is charged to 4E the VLJ, Attorney and Team get updated in the BRIEFF table.


app/repositories/queue_repository.rb
```diff
@@ -186,6 +171,32 @@ def update_location_to_attorney(vacols_id, attorney)
...
+      # Attorney and Team get updated in the BRIEFF table
+      decass_record.case.update(
+        bfmemid: decass_record.dememid,
+        bfattid: decass_record.deatty,
+        bfboard: decass_record.deteam
+      )
+      assign_case_for_quality_review(decass_record.case) if location == :quality_review
+    end
+
+    def assign_case_for_quality_review(vacols_case)
```
aroltsch: Moved this method under private

Issue #6234, opened 2018-07-16
Allow users to set payee code during intake


abbyraskinUSDS:
## Story
For 3 AMA lanes, when entering a claimant other than the Veteran, I want to select the payee code so the VSR can process the claim.

If the claimant is other than Veteran, once the claimant is selected from presented options, it will show a list of relationships and the type of relationship. The user will then need to select the payee code. For now, we will show the entire list of possible payee codes, and optimize in the future.

## Technical notes
Payee code should be stored in the database, and sent to VBMS when establishing a claim
Depending on VBMS validation, it's possible that this will have to happen concurrently with setting the claimant participant ID when establishing the claim in VBMS

## Acceptance criteria
- Once the Claimant is selected, a searchable drop-down menu appears that contains the entire payee code list.

## Copy
- Field label: "What is the payee code for this claimant?"
- Drop down contents: [listed here](https://github.com/department-of-veterans-affairs/caseflow/issues/5878#issuecomment-409051820)

## Mocks
<img width="618" alt="screen shot 2018-07-31 at 9 14 47 am" src="https://user-images.githubusercontent.com/11219417/43461747-3a5b5be8-94a2-11e8-8cbb-1e4fecd3c132.png">

## Related issues
- #5878 

## How to Validate
- Test user criteria: AMA with dependents
- Steps to get there: Original intake flow, 3 AMA lanes (SC, HLR, Board) > Choose claimant isn't Veteran > select name > see payee codes
- Choose a payee code that matches the claimant relationship, e.g. choose "Spouse" and "Spouse"
- Dev team to check that the payee is stored in the database
- Check if the payee code is set on the claim (VBMS to help)

Validation for #5281 can happen at the same time

abbyraskinUSDS:
@suzchap this would be a good one to get ready for dev soon if possible

kahlouie:
Should be sent at the same time as the other claimant and we should store payee code in our db.

kahlouie:
Only adding the payee code selection

suzchap:
I began validation with test user 564930680 and chose payee code 99. When I went to create the EP on the "finish" page, I got the "something went wrong" error so I canceled the intake.

Once that's resolved, I guess I'll need to submit again? Or will we be able to check the last two bullets in the "how to validate" section without actually creating an ep?

@kahlouie or @shanear can one of you help with this or should we wait for Sally to be back?

shanear:
Let's wait for sally to get back

[...]

PR #6555, opened 2018-08-08
Add ability to set payee code for the claimant

leikkisa:
Connects #6234
connects #6098


app/models/ama_review.rb
```diff
@@ -38,6 +38,12 @@ def claimant_not_veteran
     claimant_participant_id && claimant_participant_id != veteran.participant_id
   end
 
+  def payee_code
+    # binding.pry
```
shanear: remove this

app/models/appeal_intake.rb
```diff
@@ -8,6 +8,7 @@ def ui_hash(ama_enabled)
       receipt_date: detail.receipt_date,
       claimant: detail.claimant_participant_id,
       claimant_not_veteran: detail.claimant_not_veteran,
+      payee: detail.payee_code,
```
shanear: why is this called `payee` and not `payee_code`?

app/models/ama_review.rb
```diff
@@ -20,9 +20,9 @@ def start_review!
     @saving_review = true
   end
 
-  def create_claimants!(claimant_data:)
+  def create_claimants!(participant_id:, payee_code:)
```
shanear: I think we may just want to keep these arguments unspecified like so:

```
def create_claimants!(attrs)
  ...
end
```

This keeps us from having to keep modifying the arguments list all the way up and down the method call chain

app/models/ramp_review.rb
```diff
@@ -153,6 +154,10 @@ def end_product_modifier
     (END_PRODUCT_DATA_BY_OPTION[option_selected] || {})[:modifier]
   end
 
+  def payee_code
```
shanear: we don't need this anymore since `payee_code` defaults to "00"

leikkisa: EndProductEstablishment still needs this, so keeping it here (it will just default on EndProduct for dispatch claims established)

app/models/ama_review.rb
leikkisa: I'm keeping this because there's a logic step where we check if they've been submitted in the params or not (because if the user doesn't select the option for there to be another claimant, they don't get sent back with the params).

PR #6558, opened 2018-08-09
wip refactor poa

amprokop:
Connects #{github issue number}

### Description

### Acceptance Criteria 
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...


app/models/vacols/case.rb
```diff
@@ -56,15 +56,15 @@ class InvalidLocationError < StandardError; end
...
+    "O" => { full_name: "Other Service Organization", short: "Other" },
     "P" => { full_name: "Army & Air Force Mutual Aid Assn.", short: "Army Mut Aid" },
     "Q" => { full_name: "Catholic War Veterans", short: "Catholic War Vets" },
     "R" => { full_name: "Fleet Reserve Association", short: "Fleet Reserve" },
     "S" => { full_name: "Marine Corp League", short: "Marine Corps League" },
     "T" => { full_name: "Attorney", short: "Attorney", rep_name_in_rep_table: true },
     "U" => { full_name: "Agent", short: "Agent", rep_name_in_rep_table: true },
     "V" => { full_name: "Vietnam Veterans of America", short: "VVA" },
-    "W" => { full_name: "One Time Representative", short: "One Time Rep", rep_name_in_rep_table: true },
+    "W" => { full_name: "One Time Representative", short: "One Time Rep" },
```
amprokop: Out of 1102 appeals marked "One Time Representative", only 220 have rows in the rep table (~1/5), and from inspecting the data, it looks like most of those are very old. We probably shouldn't assume that we'll find rep name data in that table, so I'd like to remove this designation.

There are frequent contradictions in VACOLS POA data. In the future, I think we should move to treating VACOLS POA like so:

If BRIEFF.BFSO contains a specific service org value, display that service org
AND
If REP contains 1 or more values for the appeal, display all of those values.

That way we surface everything in VACOLS instead of trying to interpret what's right.

app/models/vacols/case.rb
```diff
@@ -4,7 +4,7 @@ class VACOLS::Case < VACOLS::Record
   self.primary_key = "bfkey"
 
   has_one    :folder,          foreign_key: :ticknum
-  has_one    :representative,  foreign_key: :repkey
+  has_many   :representatives, foreign_key: :repkey
```
amprokop: This changes the relationship from 1:1 to 1:many.

app/services/appeal_repository.rb
```diff
@@ -133,7 +133,6 @@ def self.set_vacols_values(appeal:, case_record:)
       vbms_id: case_record.bfcorlid,
       type: VACOLS::Case::TYPES[case_record.bfac],
       file_type: folder_type_from(folder_record),
-      contested_claim: case_record.representative.try(:reptype) == "C",
```
amprokop: Since this call to `representative` was called in the `set_vacols_values` method, we were lazy loading REP table for every appeal. Moving this to a method on the object and correcting the logic — we should be looking for *any* rep row, not just one.

spec/factories/vacols/case.rb
```diff
@@ -4,7 +4,6 @@
     sequence(:bfcorkey)
     sequence(:bfcorlid, 100_000_000) { |n| "#{n}S" }
 
-    association :representative, factory: :representative, repkey: :bfkey
```
amprokop: the `association` method won't work for a 1:many, you need to manually create those

spec/factories/vacols/representative.rb
```diff
@@ -1,6 +1,7 @@
 FactoryBot.define do
   factory :representative, class: VACOLS::Representative do
     sequence(:repkey)
-    sequence(:repaddtime) { Time.now.utc }
+    sequence(:repaddtime) { |n| 2.years.ago - n.days }
```
amprokop: repaddtime is part of a uniqueness constraint in the VACOLS DB, so we need to make sure we aren't trying to generate the same value for it every time

app/mappers/power_of_attorney_mapper.rb
```diff
@@ -1,4 +1,6 @@
+# rubocop:disable Metrics/ModuleLength
 module PowerOfAttorneyMapper
+  # TODO: break this module up into BGS and VACOLS-specific POA parsers
```
amprokop: In a future PR, I intend to break up the Power of Attorney model into a base model and BGS and VACOLS models, which inherit from the base model. The base model will enforce a consistent interface, and BGS and VACOLS models will contain their own code. This refactor is temporary/halfway.

app/models/legacy_appeal.rb
```diff
@@ -188,10 +188,15 @@ def disposition_remand_priority
...
+    # TODO: this will only return a single power of attorney. There are sometimes multiple values, eg.
+    # when a contesting claimant is present. Refactor so we surface all POA data.
     @poa ||= PowerOfAttorney.new(file_number: veteran_file_number, vacols_id: vacols_id).tap do |poa|
       # Set the VACOLS properties of the PowerOfAttorney object here explicitly so we only query the database once.
-      poa.class.repository.set_vacols_values(poa: poa, case_record: case_record)
-      poa
+      poa.class.repository.set_vacols_values(
+        poa: poa,
+        case_record: case_record,
+        representative: VACOLS::Representative.appellant_representative(vacols_id)
```
amprokop: This will be refactored away as well in a future PR.

app/models/legacy_appeal.rb
```diff
@@ -235,6 +240,12 @@ def representative_type
     power_of_attorney.vacols_representative_type
   end
 
+  delegate :representatives, to: :case_record
+
+  def contested_claim
```
amprokop: Basically, before we checked if the first rep row for an appeal was a contesting claimant — but many appeals have 2+ rows, we need to check all of them.

app/models/vacols/representative.rb
```diff
@@ -3,7 +3,35 @@ class VACOLS::Representative < VACOLS::Record
   self.table_name = "vacols.rep"
   self.primary_key = "repkey"
 
-  class InvalidRepTypeError < StandardError; end
+  class RepError < StandardError; end
+  class InvalidRepTypeError < RepError; end
+
+  # mapping of values in REP.REPTYPE
+  ACTIVE_REPTYPES = {
```
amprokop: Previously, we were not parsing the value from REPTYPE, which should indicate what type of representative the REP row is.

app/models/vacols/representative.rb
```diff
@@ -26,58 +54,93 @@ def self.update_vacols_rep_type!(bfkey:, rep_type:)
...
-                REPADDR2 = #{address_two},
-                REPCITY = #{city},
-                REPST = #{state},
-                REPZIP = #{zip}
-          WHEN NOT MATCHED THEN INSERT (REPKEY, REPADDR1, REPADDR2, REPCITY, REPST, REPZIP)
-            VALUES ( #{case_id}, #{address_one}, #{address_two}, #{city}, #{state}, #{zip} )
-        SQL
-      end
-    end
+  def self.create_rep!(bfkey, rep_attrs)
```
amprokop: These methods are cribbed from @aroltsch's case_issue.rb.

app/services/power_of_attorney_repository.rb
```diff
@@ -1,87 +1,53 @@
 class PowerOfAttorneyRepository
   include PowerOfAttorneyMapper
 
-  # :nocov:
-  def self.poa_query(poa)
-    VACOLS::Case.includes(:representative).find(poa.vacols_id)
```
amprokop: This query only found the first REP row for a vacols_id. As a temporary measure to avoid breaking anything and increase data accurary, we'll now surface the first REP row for a vacols_id that is a rep for the appellant (rather than for a contesting claimant)

app/services/power_of_attorney_repository.rb
```diff
@@ -1,87 +1,53 @@
...
-        state: address[:state][0, 4],
-        zip: address[:zip][0, 10]
-      }
-    )
-  end
-  # :nocov:
-
-  # TODO: Consider changing this logic. Move the entire name into one field.
-  def self.update_vacols_rep_table!(appeal:, representative_name:, address:)
+  def self.update_vacols_rep_table!(appeal:, representative_name:, address:, rep_type:)
```
amprokop: Combine the updates into one method — we weren't calling them separately anyway.

spec/factories/vacols/representative.rb
```diff
@@ -1,6 +1,9 @@
 FactoryBot.define do
   factory :representative, class: VACOLS::Representative do
     sequence(:repkey)
-    sequence(:repaddtime) { Time.now.utc }
+    sequence(:repaddtime) do |n|
```
amprokop: VACOLS has a uniqueness constraint on repkey + repaddtime

spec/models/vacols/representative_spec.rb
```diff
@@ -0,0 +1,13 @@
+describe VACOLS::Representative do
+  let(:vacols_case) { create(:case_with_multiple_rep_rows) }
+
+  context ".representatives" do
```
amprokop: this needs to be filled out

Issue #6546, opened 2018-08-07
Using back and forward in browser while uploading to VACOLS


tejans24:
Steps:
1). Upload a validated RO schedule
2). Confirm uploading to VACOLS
3). Click the back button before it finishes
4). Proceed forward again (browser control)

It will keep uploading many many records to VACOLS as long as that process is repeated.

MeredithStewart:
Seems like something that should be stopped. Justin and Scott wouldn't purposely try to mess up the system, but seems like we shouldn't let possibly do either.

oscarAtNava:
@sharonwarner Unless you already started working on this one, it seems related to the 6485 as it is all in the same components. You want me to take a look at this as well?

PR #6564, opened 2018-08-09
Adding transaction for finalized / vacols upload

sharonwarner:
Connects #6546 

### Description
This PR adds a transaction block around setting a schedule period to finalized and uploading the algorithm results to VACOLS.

### Testing Plan
1. Confirm a VACOLS upload through the UI
1. Click the back button
1. Confirm through the rails console that finalized is set to true and the VACOLS updates have gone through
1. Click confirm again
1. Confirm the UI displays a success message but no additional hearings have been added to VACOLS


sharonwarner:
Yeah, there's definitely a couple different ways to do it, I don't think it's worth prioritizing right now since we only have two users, and we can tell them to not hit the back button. I'll create a backlog ticket to come up with a better solution when we have more time.

Issue #6370, opened 2018-07-26
Veteran Documents column iteration: include link, doc count and "new" icon


mkhandekar:
We removed the document count because in colocated research, we found that folks did not mention using document count to help them prioritize cases, and because `View documents` is a much clearer & actionable link than `1234 documents`. 

But that might have been because we didn't actively inquire about it - and that's a small research failure on our part, because we started getting support tickets asking for this feature back. [See slack thread](https://dsva.slack.com/archives/C6E41RE92/p1532379929000041).

Here's a middle-of-the-road approach that meets these needs:

## AC
- actionable link `View Docs`
- leaving space for "new" icon when needed, to the right of the link. [Download SVG here](https://github.com/department-of-veterans-affairs/appeals-design-research/tree/master/Projects/Caseflow%20Queue/images). 
  - on hover, tooltip informs user that "this case has new documents" 
- In table: document count (2 pts smaller than link, in color `$color-gray`)
- In case details: document count (same size as link, in color `$color-gray`)

### In table 
<img width="318" alt="screenshot 2018-07-31 11 13 07" src="https://user-images.githubusercontent.com/1598889/43468747-be3035e6-94b2-11e8-8859-937474d5bfaf.png">

### In case details
<img width="1468" alt="screenshot 2018-07-31 11 17 51" src="https://user-images.githubusercontent.com/1598889/43469072-641cca32-94b3-11e8-87bf-8f020a483081.png">

cc: @sneha-pai @laurjpeterson

laurjpeterson:
This looks good to me! I added the Acceptance Criteria heading to your three bullet points - I think it's ready to be picked up by an engineer!

mkhandekar:
Thank you, LP!

mdbenjam:
What should the loading state for the number of documents be? Is the text `Loading number of docs...` okay?

mkhandekar:
We decided to have the loading text on the table, but not in case details.

lowellrex:
Completed by #6571.

PR #6571, opened 2018-08-09
Unify tooltip appearance and add to new docs icon and Veteran ID copy button

lowellrex:
Connects #6370.

* Unifies the styling of tooltips
* Adds tooltip to new documents icon
* Adds tooltip to Veteran ID copy button

Additionally, this PR creates a new `Tooltip` component which wraps `ReactTooltip`, requires a text attribute (the text of the tooltip) and a child element (the element that triggers the tooltip to appear when it is hovered).

### New documents tooltip
| Table view | Case details title |
| --- | --- |
| image | image |

### Column header tooltip
| Before | After |
| --- | --- |
| image | image |

### Veteran ID copy button tooltip
image

### DIFF
--- a/client/app/queue/components/NewFile.jsx
+++ b/client/app/queue/components/NewFile.jsx
@@ -1,15 +1,14 @@
 // @flow
 import * as React from 'react';
 import { connect } from 'react-redux';
-import { css } from 'glamor';
-import ReactTooltip from 'react-tooltip';
 import { NewFileIcon } from '../../components/RenderFunctions';
-import { COLORS } from '../../constants/AppConstants';
+import Tooltip from '../../components/Tooltip';
 import { bindActionCreators } from 'redux';
 import { getNewDocuments } from '../QueueActions';
 import type {
   BasicAppeal
 } from '../types/models';
+import COPY from '../../../COPY.json';
 
 type Params = {|
   appeal: BasicAppeal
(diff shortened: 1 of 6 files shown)


client/app/queue/components/NewFile.jsx
```diff
@@ -19,6 +22,17 @@ type Props = Params & {|
...
 
+const tooltipID = 'newfile-tip';
+
+const tooltipStyling = css({
+  [`& > #${tooltipID}`]: {
+    backgroundColor: COLORS.GREY_DARK,
+    padding: '0.5rem 1rem',
+    textAlign: 'center'
+  },
+  [`& > #${tooltipID}:after`]: { display: 'none' }
```
lowellrex: Removes the caret from the bottom of the tooltip.

mdbenjam: Is it important to remove the caret? I know the mocks don't have it, but we keep the caret on our task table. I'd like to be consistent here. If we decide there shouldn't be a caret then can we wrap `ReactTooltip` in our own component with these changes? I don't want every usage to have to specify its own CSS.

mdbenjam: The same goes for the text align and color.

mdbenjam:
There is one other place we use the react tooltip: https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/hearings/components/WorksheetHeader.jsx#L140-L156 We could try to unify that as well... Don't think that's necessary with this PR though.

Issue #6004, opened 2018-06-25
Case details | make Veteran ID copy-able


mkhandekar:
Case details includes the Veteran ID prominently next to the Veteran name. We've learned that users often copy and paste it to do their work. Let's use / upgrade if needed the copy-able functionality already available elsewhere in Caseflow, and translate that for case details. 

## Case details (current)
![image](https://user-images.githubusercontent.com/1598889/40918784-5ad1fb6a-67d5-11e8-8bd7-68332104bdc0.png)

## Copy-able Veteran ID (for reference)
<img width="143" alt="screen shot 2018-05-22 at 2 57 03 pm" src="https://user-images.githubusercontent.com/27704093/40798717-b524341a-64d9-11e8-827e-5219842e91c2.png">

## To do:

- [ ] design a slimmer, more appropriate copy-able text / button for the Veteran ID

mkhandekar:
@amprokop do you know who worked on the copy-able Veteran ID component?  I'd love to see if we can update it with something like this. Or if it would be ok to develop a slightly alternative version for Queue. The changes include a hover state with a tooltip telling users of what to expect and Veteran ID label to the left instead of above (like the reference image in the original comment ^).

## Copy-able Veteran ID [normal]
<img width="752" alt="screenshot 2018-07-09 20 32 05" src="https://user-images.githubusercontent.com/1598889/42487786-73557b46-83b7-11e8-93da-7dbe18a3086f.png">

## Copy-able Veteran ID [hover]
<img width="751" alt="screenshot 2018-07-09 20 32 20" src="https://user-images.githubusercontent.com/1598889/42487787-736e379e-83b7-11e8-8118-d06ea73e92f3.png">

cc: @laurjpeterson

laurjpeterson:
@lowellrex implemented this last week, closing

PR #6572, opened 2018-08-09
Lowell/6004 add tooltip veteran id button

lowellrex:
Connects #6004.

### Screenshot with tooltip
image

### DIFF
--- a/client/app/queue/CaseTitle.jsx
+++ b/client/app/queue/CaseTitle.jsx
@@ -12,6 +12,7 @@ import { CATEGORIES } from './constants';
 import { COLORS } from '../constants/AppConstants';
 import ReaderLink from './ReaderLink';
 import { ClipboardIcon } from '../components/RenderFunctions';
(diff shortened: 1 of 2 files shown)


client/app/queue/CaseTitle.jsx
```diff
@@ -63,6 +65,17 @@ const clipboardButtonStyling = css({
   '&:hover > svg path': { fill: COLORS.PRIMARY }
 });
 
+const clipboardTooltipID = 'clipboard-tooltip';
+
+const tooltipStyling = css({
+  display: 'inline-block',
```
lowellrex: Need this to prevent the "View all cases" text at the end of the case title from jumping onto another line (I think it was because this `span` element was being rendered as a block, maybe?).

mdbenjam: Darn, that sucks.

lowellrex:
Closing this ticket out as this work will be included in PR #6571.

Issue #6086, opened 2018-07-02
Send back to attorney (back end completed), send to other teams (front end only)


laurjpeterson:
When co-located admin support users finish a task, they need to send it back to the attorney.

Note: co-located will also send tasks to other teams in the future, but that is out of scope for this ticket.

### Send back to attorney
- On case details, the Actions dropdown contains `Send back to attorney`
- When they select that option, the modal per designs pops up so the user can confirm sending 

### Send to another team
- Establish a page dedicated to sending the case to another team
- Provide a dropdown that includes `Hearing Branch` and `Translation` - verified
- Backend ticket - https://github.com/department-of-veterans-affairs/caseflow/issues/6632

All tasks should have the option to send back to the assigning attorney. Tasks that are not on hold should also have an option to place a hold, and tasks of `:translation` or `:schedule_hearing` types should have an option to send to the corresponding team.

### Send back to attorney
![co-located_modal_sending-case_attorney](https://user-images.githubusercontent.com/31224257/43865169-c6aa700a-9b2f-11e8-9d53-f622241cc5f9.png)

### Send action to another team
<img alt="screen_shot_2018-08-24_at_2 36 39_pm_720" src="https://user-images.githubusercontent.com/8533004/44662564-2eed4100-a9dc-11e8-8f03-c02700daa93b.png">
<details><summary>old mock</summary>
<img src="https://user-images.githubusercontent.com/31224257/43915960-3cd6e952-9bda-11e8-9d7a-f8f92ca015ad.png">
</details>

## Reference - sending to other teams
- We will support co-located assigning `Schedule hearing` tasks to co-located right now re: https://github.com/department-of-veterans-affairs/caseflow/issues/5931
- Might expand to Translation - need to confirm that attorneys don't assign directly to Translation, what VACOLS location to assign to, etc.

evankroske:
## Implementation Plan for Assigning Back to Attorney
- [ ] Finish PR. Estimate: 2 more days.
- [ ] Get PR approved. Estimate: 1 day.

- Total estimate: 3 days.

## Implementation Plan for Assigning to Other Teams
[insert plan here]

PR #6599, opened 2018-08-10
Colocated | Send back to assigning attorney

dannysellers:
Connects #6086 

### Description
This adds a `SendToAssigningAttorneyModal` which is displayed when Send to Attorney is selected from the `ColocatedActionsDropdown` in `CaseDetailsView`. This also replaces `queue/constants.USER_ROLES` with `USER_ROLE_TYPES.json`

### Acceptance Criteria 
- [ ] Code compiles correctly

### Testing Plan
1. Go to `/queue` as a colocated user (`BVALSPORER`), go to case details screen
2. Choose Send back to Attorney from dropdown, confirm modal appears
3. Confirm assignment happens successfully / user is returned to `/queue` on success

</details>


client/app/queue/components/ColocatedActionsDropdown.jsx
```diff
@@ -0,0 +1,92 @@
...
+    } = this.props;
+    const actionType = props.value;
+
+    this.stageAppeal();
+
+    if (actionType === CO_LOCATED_ACTIONS.SEND_BACK_TO_ATTORNEY) {
+      return this.props.showModal('sendToAttorney');
+    }
+
+    const routes = {
```
dannysellers: This isn't used yet

app/models/serializers/work_queue/task_serializer.rb
```diff
@@ -4,14 +4,22 @@ class WorkQueue::TaskSerializer < ActiveModel::Serializer
...
   attribute :started_at
   attribute :completed_at
   attribute :placed_on_hold_at
   attribute :instructions
   attribute :appeal_type
 
+  attribute :assigned_by do
+    {
+      first_name: object.assigned_by.full_name.split(" ").first,
+      last_name: object.assigned_by.full_name.split(" ").last,
```
lowellrex: Could we pass only the first initial of the first name here instead of the whole first name? My quick survey of the front-end code suggests that we only use the first initial when we access the first name of the assigner.

With the current approach folks with two first names (like Mary Beth Hurt) would be given the first name of "Mary" instead of the correct "Mary Beth". This isn't a new issue being introduced here (and is by no means a show stopper), but I think we have the opportunity to fix it here by only sending the first initial to the front-end.

lowellrex: Similarly, this approach improperly renders a single last name for folks with two last names. For instance, Martin Van Buren would become "Martin Buren". I don't know how to handle this situation, but once again this is not a dealbreaker for this PR, just something to noodle on.

dannysellers: It might be worthwhile to just send the name formatted with [`FullName.format(:readable_fi_last_formatted)`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/full_name.rb#L21), assuming we don't need to display the first and last names separately (which looks to be the case)

lowellrex: Haha `BVAAABSHIRE` would be J. Cases. I kinda like that.

I think the issue with using FullName is that [the constructor](https://github.com/department-of-veterans-affairs/caseflow/blob/5ec8e7b20aff726889d82801da9d7eea875500cf/app/models/full_name.rb#L6) requires three separate names (of any word length though!) which we don't have in the `User` table.

For now this is fine, the thought just occurred to me while reading through this code so figured I'd share. Also, if we get a chance to rethink how we store/display name information further down the line we can point to concrete examples like this.

Issue #6606, opened 2018-08-10
BGS End Product bug


leikkisa:
Handle case when BGS returns [{:number_of_records=>"0", :return_code=>"SHAR 9999", :return_message=>"Records found"}] for end_products

leikkisa:
High priority due to VBMS testing deadline of today

leikkisa:
Validated that this is working now with Sandra while she was doing VBMS testing

PR #6609, opened 2018-08-10
Handle zero end products response from BGS

leikkisa:
Connects #6606


leikkisa:
Thanks! Should have caught it the first time, doh

Issue #6613, opened 2018-08-13
error when RO doesn't exist for CO hearing days


tejans24:
![screen shot 2018-08-13 at 11 58 50 am](https://user-images.githubusercontent.com/1976755/44043220-48af2bea-9ef0-11e8-8f8e-9b45b1da2687.png)

PR #6614, opened 2018-08-13
handling a case when RO doesn't exist for CO hearing days

tejans24:
Connects #6613 

### Description
Adding an additional check for CO hearing days when RO doesn't exist

### Acceptance Criteria 
- [ ] Tests pass
- [ ] CO days are handled.

### Testing Plan
1. Go to ...

### DIFF
--- a/app/controllers/hearings/schedule_periods_controller.rb
+++ b/app/controllers/hearings/schedule_periods_controller.rb
@@ -7,21 +7,17 @@ class Hearings::SchedulePeriodsController < HearingScheduleController
   end
 
   def show
-    begin
-      sp = if schedule_period.can_be_finalized?
-            schedule_period.to_hash.merge(
-              hearing_days: schedule_period.algorithm_assignments.map do |hearing_day|
-                hearing_day[:regional_office] = ro_information(hearing_day[:regional_office])
-                hearing_day
-              end
-            )
-          else
-            schedule_period.to_hash
-          end
-      render json: { schedule_period: sp }
-    rescue HearingSchedule::Errors::NotEnoughAvailableDays => error
-      render json: { error: error.message, details: error.details }, :status => 422
-    end
+    sp = if schedule_period.can_be_finalized?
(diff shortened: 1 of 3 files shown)


app/controllers/hearings/schedule_periods_controller.rb
```diff
@@ -56,4 +59,15 @@ def schedule_period_params
   def schedule_period
     SchedulePeriod.find(params[:schedule_period_id])
   end
+
+  private
+
+  def ro_information(ro_key)
```
sharonwarner: I think this makes more sense in regional_office.rb.

app/controllers/hearings/schedule_periods_controller.rb
```diff
@@ -56,4 +59,15 @@ def schedule_period_params
...
+    regional_office = RegionalOffice::CITIES[ro_key]
+    if regional_office
+      "#{regional_office[:city]}, #{regional_office[:state]}"
+    else
```
sharonwarner: Is the else section definitely necessary? I think it would return nil regardless, but I'm not sure.

app/controllers/hearings/schedule_periods_controller.rb
```diff
@@ -7,18 +7,21 @@ def index
...
+              hearing_days: schedule_period.algorithm_assignments.map do |hearing_day|
+                hearing_day[:regional_office] = ro_information(hearing_day[:regional_office])
+                hearing_day
+              end
+            )
+          else
+            schedule_period.to_hash
+          end
+      render json: { schedule_period: sp }
+    rescue HearingSchedule::Errors::NotEnoughAvailableDays => error
```
sharonwarner: I think this will break the frontend error handling. Right now the frontend sends 500 errors to an error page, but if we catch this error, the Review Assignments page will error out.

tejans24: This snuck in from another PR. I removed that code.

Issue #6427, opened 2018-07-30
Backfill and End product modifier not getting saved to End Product Establishments


leikkisa:
Noticed when poking through UAT that End Products are getting created correctly with the correct modifier, but that the end product modifier in EndProductEstablishments is nil.

kahlouie:
Maybe the right thing is to drop the end product modifier column as we don't use it, even for getting the end product or for showing end product contentions on the page.

kahlouie:
Fixed this in https://github.com/department-of-veterans-affairs/caseflow/pull/6619

leikkisa:
Validated that this is working, still need to backfill data

shanear:
Backfilled the data via the following commands:
```
EndProductEstablishment.where(modifier: nil, code: "682HLRRRAMP").update_all(modifier: "682")
EndProductEstablishment.where(modifier: nil, code: "683SCRRRAMP").update_all(modifier: "683")
```

This fixed all the data:
```
irb(main):022:0> EndProductEstablishment.where(modifier: nil).count
[2018-08-22 12:59:42 -0400]    (6.3ms)  SELECT COUNT(*) FROM "end_product_establishments" WHERE "end_product_establishments"."modifier" IS NULL
=> 0
```

Issue #6148, opened 2018-07-09
AMA Intake | Create a non-rating EP if Supplemental Claim or HLR contains non-rating issues


shanear:
If a HLR or SC contains non-rating issues, we will need to create an EP with a special non-rating code so it can be routed properly to case workers who can handle it.

## Acceptance Criteria
- If a compensation non-rating issue was added to a HLR or SC during intake, create a separate non-rating EP during the completion step. The non rating issues should be represented as contentions on that EP.
- If there are also rating issues on the HLR or SC, create a rating EP as well, which only contains contentions representing the rating issues.
- If there are any "Unknown" issues, these should be created under a rated EP

## Technical Notes
 The non rating codes are:
`040SCNR Supplemental Claim Non-Rating`  
`030HLRNR - Higher-Level Review Non-Rating`

## Additional Notes
Updating the success page as a result of the non-rating EP will be handled in a separate ticket:
blocks https://github.com/department-of-veterans-affairs/caseflow/issues/6149

## Validation
- Validate for both SC and HLR
- Use test file number 888451290 for both
- Confirm in VBMS that the correct non-rating codes were used (see technical notes)

leikkisa:
Some of the edit functionality is blocked by this

leikkisa:
Adding a couple of notes from conversations (not sure if this is part of this ticket, or if contentions would be part of another ticket).
There is a contention_category field on contentions, so it'd be good to check if we can use it for the non-rated issue category.  

Also, the Queue team would like to know what happens with the description and category for non-rated issues.

suzchap:
Validated with Sally, LGTU

<img width="1419" alt="screen shot 2018-09-12 at 2 43 21 pm" src="https://user-images.githubusercontent.com/11219417/45446187-409c3d00-b69a-11e8-99fc-e286b77fcee8.png">

PR #6619, opened 2018-08-13
Add non rated eps for hlr and sc flows

kahlouie:
Connects #6148 and #6427


kahlouie:
needs validation instructions

Issue #6622, opened 2018-08-13
Prevent requests that assume user works at board when loading Queue


lowellrex:
Currently, we make a handful of requests from the Queue frontend when loading the personalized task queue. Some of these requests assume the user works at the board and fail when the user is neither an attorney nor a judge (or colocated admin?). In order to support VSO users, we need to be able to load the queue for the various VSO organizations.

We can handle this a number of ways, but the simplest is probably just to avoid making requests that assume the person accessing the page works at the board by default.

mdbenjam:
This line causes a 403: https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/QueueLoadingScreen.jsx#L45 because we can't access that endpoint.

This line makes us not make the call to the endpoint we need, because VSO are not attorneys or judges:
https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/QueueLoadingScreen.jsx#L59

This no longer works since it is for legacy appeals not ama appeals and ama tasks which is what VSO tasks/appeals are: https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/QueueLoadingScreen.jsx#L72

We probably should just make an organizational loading screen to avoid all these problems, and replace it here: https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/QueueApp.jsx#L133

PR #6639, opened 2018-08-14
Enable access to VSO organizational queue for VSO users

lowellrex:
Connects #6622.

Enables access to VSO organizational queues for VSO users by avoiding API calls that fail.

### Testing Plan
1. Log in as `VSO` user
2. In local dev env, set feature flag for the American Legion organization `psql -W -h 0.0.0.0 -p 5432 -U postgres -d caseflow_certification_development -c "update organizations set feature = 'org_queue_for_american-legion' where id = 1 and url = 'american-legion';"`
3. In local rails console, enable the American Legion organizational queue feature toggle for the VSO user `org = Vso.where(url: 'american-legion').first; u = User.where(css_id: 'VSO').first; FeatureToggle.enable!(org.feature.to_sym, users: [u.css_id])`
4. Navigate to `http://localhost:3000/organizations/american-legion`
5. Marvel at the beauty of the organizational task queue.

### DIFF
--- a/client/app/queue/OrganizationQueueLoadingScreen.jsx
+++ b/client/app/queue/OrganizationQueueLoadingScreen.jsx
@@ -1,0 +1,64 @@
+// @flow
+import * as React from 'react';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+
+import LoadingDataDisplay from '../components/LoadingDataDisplay';
+import { LOGO_COLORS } from '../constants/AppConstants';
+import ApiUtil from '../util/ApiUtil';
+import { extractAppealsAndAmaTasks } from './utils';
+
+import { onReceiveQueue } from './QueueActions';
+
+type Params = {|
+  children: React.Node,
+  urlToLoad: string
+|};
+
+type Props = Params & {|
+  // Action creators
+  onReceiveQueue: typeof onReceiveQueue
+|};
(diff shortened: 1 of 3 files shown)


client/app/queue/OrganizationQueueLoadingScreen.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+      this.props.onReceiveQueue(extractAppealsAndAmaTasks(tasks));
+    }
+  );
+
+  reload = () => window.location.reload();
+
+  render = () => {
+    const failStatusMessageChildren = <div>
+      It looks like Caseflow was unable to load your cases.<br />
```
evankroske: It seems like there are a few copies of this copy. Do you think it would make sense to move it to COPY.json?

lowellrex: Might be a little tricky to put it in COPY.json because of the html elements, but I definitely think it would be a good idea to make it a default in `LoadingDataDisplay` at the very least. Will make a PR in a moment.

lowellrex: Created a ticket because I'm not going to be able to make a PR immediately. https://github.com/department-of-veterans-affairs/caseflow/issues/6646

PR #6640, opened 2018-08-14
adds remaining attributes to IDT API— document information, outstanding review, and QR info

amprokop:
Connects https://github.com/department-of-veterans-affairs/caseflow/issues/6645


app/controllers/idt/api/v1/appeals_controller.rb
```diff
@@ -25,10 +29,18 @@ def appeals_by_file_number
...
-    json_details = json_appeal(appeal)
-    json_details[:data][:attributes][:assigned_by] = task.added_by.try(:name)
-    json_details
+  def json_appeal_details(tasks, appeal)
+    appeal_details = ActiveModelSerializers::SerializableResource.new(
+      appeal,
+      serializer: ::Idt::V1::AppealDetailsSerializer
+    ).as_json
+
+    appeal_details[:data][:attributes][:documents] = ActiveModelSerializers::SerializableResource.new(
```
amprokop: Adding this other serializer here isn't ideal, but it seemed like the least awkward thing to do since it's two totally separate models.

Issue #6615, opened 2018-08-13
RAMP Intake | Automatically rollback RAMP election when EP is canceled


shanear:
In order to make an entire class of wild RAMPs be a little easier to deal with, and test a feature that will be used in the end state, we will be rolling back RAMP elections whenever the corresponding EP is canceled.

## Acceptance Criteria
- When the EP status of a RAMP election is being synced, and the status is updated to canceled from any other status, then the corresponding election should be rolled back.
- The description for the rollback should be "Automatic roll back due to EP [682/683] cancelation"

## Technical notes
There is a class we've been using for manual rollbacks of RAMP elections, `RampElectionRollback` that should be used for rolling back the RAMP election.

## Notes
- For now, this will only occur for RAMP elections. We will assess whether we need to do it for RAMP refillings too soon.
- The end state story for this same functionality is https://github.com/department-of-veterans-affairs/caseflow/issues/6097

kahlouie:
Add a feature toggle to turn it on and/or off for testing.

kahlouie:
Shane can help pair on this.

kahlouie:
To test this, you'll need to run `FeatureToggle.enable!(:automatic_ramp_rollback)` in the environment's rails console. or you'll need to make a pr to add it into the environment in appeal-deployment repo.

kahlouie:
Jennifer Williams from AMO wants to see it herself.

shanear:
This was validated by Erik Tonne and me in prod. Turning the feature on in prod today

PR #6643, opened 2018-08-14
Rollback ramp election on canceled ep

kahlouie:
Connects #6615 

Please note that this does change the system user by allowing it to be saved and found if it already exists. Due to this, I'm planning on mentioning it in engineering slack tomorrow morning. EDIT: Done, just waiting for feedback

### DIFF
--- a/app/models/ramp_review.rb
+++ b/app/models/ramp_review.rb
@@ -74,6 +74,10 @@ class RampReview < ApplicationRecord
     ## TODO: Remove this once all the data is backfilled
     if (saved_end_product_establishment = EndProductEstablishment.find_by(source: self))
       saved_end_product_establishment.sync!
+      if FeatureToggle.enabled?(:automatic_ramp_rollback) && saved_end_product_establishment.status_canceled?
+        rollback_ramp_review
+      end
+      true
     end
   end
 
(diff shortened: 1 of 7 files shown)


app/models/ramp_review.rb
```diff
@@ -74,6 +74,10 @@ def sync_ep_status!
     ## TODO: Remove this once all the data is backfilled
     if (saved_end_product_establishment = EndProductEstablishment.find_by(source: self))
       saved_end_product_establishment.sync!
+      if FeatureToggle.enabled?(:automatic_ramp_rollback) && saved_end_product_establishment.status_canceled?
+        rollback_ramp_review
+      end
+      true
```
kahlouie: @shanear I don't like just auto-returning true. Is there something better I can do here to still keep to the guidelines of a truthy return response when everything is successful?

leikkisa: Can we add the code to check that it's enabled for the current user (like we had to do for the intakeAma feature toggle?). If it's enabled for everybody this will automatically be true.

leikkisa: Also noticing that this is temporary, so if you're just going to enable it for everybody while backfilling data then nevermind

kahlouie:
@leikkisa It creates it and runs the rollback on creation, including validation that it can be rolled back.

Issue #5994, opened 2018-06-21
Limit how long a user can hold onto a RAMP intake


abbyraskinUSDS:
Currently you can hold onto a case forever, so if you go on vacation (or quit) no one else can process the form in intake, which is not ideal.

## Acceptance Criteria
- If a user has not completed an intake within a day, the intake is "released " so other users can process it
- This rule should apply to RAMP elections and refilings, as well as AMA reviews

## To Validate
- Start an intake (any type) with one user for a veteran
- Attempt to start an intake of the same type for the same veteran that day (it should say it is currently being worked on)
- Wait one day, and attempt to start an intake of the same type for the same veteran again. It should work this time

kahlouie:
One day, "release" intake when a user is searching for the veteran to start a new intake for that same veteran.

leikkisa:
@leahbannon 
Note: Validate this before this 4064, because then you'll be able to use the same file number

Test ID: 561349911
Choose supplemental claim

suzchap:
Validation part 1:
I started a SC for the test id ^.... 

I started the claim with caseflow test user CF_Q_283 and then using caseflow test user CF_KIRK_283, I tried to start a SC for the same test veteran ID and successfully got the message "TEST ONE already started processing this form We noticed that TEST ONE may be in the middle of processing the same form for this Veteran in Caseflow. Please confirm they will complete this intake, then move on to the next mail item."

Will do validation part 2 tomorrow (and then #4064 )

suzchap:
Validation part 2: I returned 25 hrs later, again as CF_KIRK_283 and tried to start intake for this test user --- and was able to just fine. Validation complete.

PR #6644, opened 2018-08-14
Expire intakes after 1 day.

shanear:
connects #5994

### DIFF
--- a/app/models/intake.rb
+++ b/app/models/intake.rb
@@ -5,13 +5,13 @@ class Intake < ApplicationRecord
   belongs_to :detail, polymorphic: true
 
   COMPLETION_TIMEOUT = 5.minutes
+  IN_PROGRESS_EXPIRES_AFTER = 1.day
 
   enum completion_status: {
     success: "success",
     canceled: "canceled",
     error: "error",
-    # TODO: This status is now unused. Remove after we verify no intakes have it.
-    pending: "pending"
+    expired: "expired"
   }
 
   ERROR_CODES = {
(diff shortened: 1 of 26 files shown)


app/models/intake.rb
```diff
@@ -5,13 +5,13 @@ class FormTypeNotSupported < StandardError; end
...
 
   COMPLETION_TIMEOUT = 5.minutes
+  IN_PROGRESS_EXPIRES_AFTER = 1.day
 
   enum completion_status: {
     success: "success",
     canceled: "canceled",
     error: "error",
-    # TODO: This status is now unused. Remove after we verify no intakes have it.
-    pending: "pending"
```
shanear: validated this in prod:
```
irb(main):001:0> Intake.where(completion_status: "pending")
[2018-08-14 19:49:44 -0400]   Intake Load (58.8ms)  SELECT  "intakes".* FROM "intakes" WHERE "intakes"."completion_status" = $1 LIMIT $2  [["completion_status", "pending"], ["LIMIT", 11]]
=> #<ActiveRecord::Relation []>
```

app/models/intake.rb
```diff
@@ -33,7 +33,11 @@ class FormTypeNotSupported < StandardError; end
   attr_reader :error_data
 
   def self.in_progress
-    where(completed_at: nil).where.not(started_at: nil)
+    where(completed_at: nil).where(started_at: IN_PROGRESS_EXPIRES_AFTER.ago...Time.zone.now)
```
shanear: Decided to lazily update expired intakes. In the mean time we just don't recognize them as in progress.

PR #6648, opened 2018-08-15
changes 1:1 to 1:many

amprokop:
Connects https://github.com/department-of-veterans-affairs/caseflow/issues/6655

This is the first PR that breaks up https://github.com/department-of-veterans-affairs/caseflow/pull/6558 into more easily reviewable chunks.

#Test plan
- Unit tests pass


app/models/vacols/representative.rb
```diff
@@ -1,10 +1,39 @@
 class VACOLS::Representative < VACOLS::Record
   # :nocov:
   self.table_name = "vacols.rep"
-  self.primary_key = "repkey"
+  # self.primary_key = "repkey"
```
lowellrex: Do we not need the primary key anymore?

PR #6660, opened 2018-08-15
Redirect VSO users to organizational queue

lowellrex:
This PR makes it so requests to `/queue` from VSO users will redirect them to the URL of the organizational queue for the VSO they are associated with. This enables us to point all VSO users to the `/queue` endpoint in training so we don't have to remember the URLs for each VSO. Additionally, this prevents VSO users from accessing personalized task queues until (if?) those are enabled.

### DIFF
--- a/app/controllers/queue_controller.rb
+++ b/app/controllers/queue_controller.rb
@@ -1,6 +1,7 @@
 class QueueController < ApplicationController
   before_action :react_routed, :check_queue_out_of_service
   before_action :verify_queue_access
+  before_action :redirect_vso_queue_requests
 
   def set_application
     RequestStore.store[:application] = "queue"
@@ -12,5 +13,15 @@ class QueueController < ApplicationController
 
   def check_queue_out_of_service
     render "out_of_service", layout: "application" if Rails.cache.read("queue_out_of_service")
+  end
+
+  private
+
+  def redirect_vso_queue_requests
+    return unless current_user.vso_employee?
+
+    Vso.where.not(feature: nil).each do |vso|
+      return redirect_to vso.path if FeatureToggle.enabled?(vso.feature.to_sym, user: current_user)
+    end
   end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/queue_controller.rb
```diff
@@ -13,4 +14,16 @@ def index
   def check_queue_out_of_service
     render "out_of_service", layout: "application" if Rails.cache.read("queue_out_of_service")
   end
+
+  private
+
+  def redirect_vso_queue_requests
+    return unless current_user.vso_employee?
+
+    Vso.where.not(feature: nil).each do |vso|
```
amprokop: what does this "feature" column refer to?

app/controllers/queue_controller.rb
```diff
@@ -13,4 +14,16 @@ def index
   def check_queue_out_of_service
     render "out_of_service", layout: "application" if Rails.cache.read("queue_out_of_service")
   end
+
+  private
+
+  def redirect_vso_queue_requests
```
amprokop: nit: the only request that is served to this controller is the request to `/queue`, correct? i might have opted to make this a condition of the `index` method rather than a `before_action`, just because making it a hook it will apply by default to any other route we later serve from this controller, and I'm not confident we want that. but this is fine for now too.

lowellrex: Smart! I'll make that change

app/controllers/queue_controller.rb
lowellrex: The feature column of the organizations table (we use single-table inheritance for the VSO model) contains the name of the [feature flag users will have enabled to have](https://github.com/department-of-veterans-affairs/caseflow/blob/f6bb7832d2c1c30a22e47fdbfe77784c225d1db1/app/controllers/organizations_controller.rb#L24) to visit that VSO's organizational queue. So if a VSO does not have a feature flag, then they cannot have a organizational queue so we can just skip them. This probably doesn't matter much since I don't think the list of VSOs will ever be that large, but I figured we could do a minimal amount of filtering at the very least.

PR #6665, opened 2018-08-16
Queue | Limit judge case review comment length

dannysellers:
Resolves https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2300/

### Description
In the DECASS table, `DEBMCOM` (VLJ comments), comment field on Evaluate Decision view, has a max length of 600 characters

### Acceptance Criteria 
- [ ] character limit imposed in comment field on Evaluate Decision view

### Testing Plan
1. Go to ...

### DIFF
--- a/client/app/queue/constants.js
+++ b/client/app/queue/constants.js
@@ -145,6 +145,8 @@ export const ISSUE_DESCRIPTION_MAX_LENGTH = 100;
 export const ATTORNEY_COMMENTS_MAX_LENGTH = 350;
 // max length for document id `DECASS.DEDOCID`
 export const DOCUMENT_ID_MAX_LENGTH = 30;
+// max length for VLJ comment `DECASS.DEBMCOM`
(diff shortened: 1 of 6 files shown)


client/app/queue/constants.js
```diff
@@ -145,6 +145,8 @@ export const ISSUE_DESCRIPTION_MAX_LENGTH = 100;
 export const ATTORNEY_COMMENTS_MAX_LENGTH = 350;
 // max length for document id `DECASS.DEDOCID`
 export const DOCUMENT_ID_MAX_LENGTH = 30;
+// max length for VLJ comment `DECASS.DEBMCOM`
+export const JUDGE_CASE_REVIEW_COMMENT_MAX_LENGTH = 600;
```
lowellrex: Can we make this a shared constant in the `client/constants/` directory so we can use it in `app/models/judge_case_review.rb` as well?

Issue #4293, opened 2018-01-05
Finish Refiling | Show error message on unknown error


shanear:
## Acceptance Criteria
- When submitting a ramp refiling, there should be an error message if the server has an unknown error. This message should look like the one on the Ramp Election finish page:

![screen shot 2018-01-08 at 1 55 44 pm](https://user-images.githubusercontent.com/1596259/34687067-a37adfe4-f47b-11e7-982a-cb0633a6c06f.png)

see: https://github.com/department-of-veterans-affairs/caseflow/pull/4272#discussion_r159923315

suzchap:
Waiting for Sally to set a test ID to have the right criteria.

leikkisa:
We will need to validate this locally since we need to create an unknown error.  However, to validate this locally I need to figure out how to make a decision on a RampElection locally (since that happens outside of Caseflow).

leikkisa:
@leahbannon 
I think this is ready to validate this on my local environment with test file number 239120550. We'll want to pair on this.

shanear:
This has been deployed for months. We can assume it's working correctly

PR #6669, opened 2018-08-17
Test data set up for ramp_refiling

leikkisa:
Connects #4293

Some test data setup so that we can do Ramp Refilings in the local environment.

Tested in local


kahlouie:
Looks like you might need to add `# rubocop:disable Metrics/AbcSize` to the top of the file to ignore the arbitrary code styling issue Rubocop has. For setup, I think this is fine. For complex business logic, that would be less than ideal, though.

PR #6677, opened 2018-08-20
Create Task methods to enable generic task completion

lowellrex:
Connects #6678.

We need to support each of the three generic task actions (`Mark complete`, `Assign to team`, `Assign to individual`--I foresee `Assign to self` just being a shortcut on the frontend for "assign to individual" with the current user's ID in the assignee parameter) in the `Task` model. This is a first pass at doing that with the `Mark complete` task action.

I still need to write tests, but I'd like a second set of eyes validating this approach given our tech spec.


app/models/task.rb
```diff
@@ -37,6 +37,19 @@ def ama?
...
 
+  def mark_complete!
+    update!(status: :completed)
+    parent.child_completed
+  end
+
+  # For now, only take action if all child tasks have been completed
+  def child_completed
+    unless children.reject { |t| t.status == :completed }.length
+      return mark_complete! if assigned_to.is_a?(Organization)
```
aroltsch: why are you calling mark_complete! here again?

app/models/task.rb
```diff
@@ -37,6 +37,19 @@ def ama?
     appeal_type == "Appeal"
   end
 
+  def mark_complete!
```
aroltsch: shall we call this `complete!`?

app/models/task.rb
aroltsch: oh I see.

app/models/task.rb
```diff
@@ -37,6 +37,19 @@ def ama?
...
+    update!(status: :completed)
+    parent.child_completed
+  end
+
+  # For now, only take action if all child tasks have been completed
+  def child_completed
```
aroltsch: The name of the method is confusing.

app/models/task.rb
lowellrex: My only concern is that naming this method `complete!` is that such a name might obscure the fact that this method does more than just change this task's status to `completed`. If you think, such a name would not be confusing then I'm totally up to change the name.

app/models/task.rb
lowellrex: Do you have a preferred name we change this to?

aroltsch: `update_parent`? We should also get rid of `update_parent_status` method or merge these two together.

app/models/task.rb
aroltsch: `mark_as_complete!` is also a good name

app/models/task.rb
```diff
@@ -37,6 +37,19 @@ def ama?
     appeal_type == "Appeal"
   end
 
+  def mark_complete!
+    update!(status: :completed)
+    parent.child_completed
```
aroltsch: In some cases `parent` might be nil. For example, root node.

app/models/task.rb
lowellrex: I'm hesitant to name this method `update_parent` because the `update` method already has a [well defined meaning](https://apidock.com/rails/ActiveRecord/Base/update/class) in Rails and I fear that naming this method something similar implies that it works similarly. On the other hand, I think it is appropriate for `update_parent_status` to be named as it is because it in fact does act similarly to the standard `update` active record method.

The idea behind having this method delegating the action that happens when this event happens (that is, when a task is completed) is to allow for the parent task to control what action it takes when a child task is completed. For instance, if we wanted to subclass some `AlertTask` so that it sends an email to somebody when all tasks are complete we would only have to override the `child_completed` method in `AlertTask`.

Issue #6683, opened 2018-08-20
Create database table for task actions


lowellrex:
As part of the effort to build [generic organizational task queues](https://github.com/department-of-veterans-affairs/caseflow/issues/6573), we will store actions that can be taken on a given action in the database so we can quickly add new actions and make actions available to different work roles. For the first phase of this effort we will only have three task actions: `Mark complete`, `Assign to team`, and `Assign to individual` (I foresee `Assign to self` just being a shortcut on the frontend for "assign to individual" with the current user's ID in the assignee parameter). Each of those actions will create a task whose sole purpose is to indicate current ownership for an appeal.

So this table (tentatively dubbed `task_actions`) should have the following fields:
* name
* status_after (to indicate what the status of the task should be after this action is taken)
* child_task_type (to fill the `assigned_to_type` field in the `Task` model)

There may be some additional fields down the line (say, when not all task actions are available to be taken on every task we may need to add a `task_definitions` table with a mapping of task actions that can be taken on which types of tasks), but for now this should do.

## Acceptance criteria
* Create `task_actions` table
* Create `TaskAction` model
* Add rows in the database for the three generic task actions

lowellrex:
Wrote up some thoughts on why we closed this ticket in the tech spec: https://github.com/department-of-veterans-affairs/caseflow/issues/6573#issuecomment-414786211

PR #6685, opened 2018-08-20
Create TaskAction model and table

lowellrex:
Connects #6683.

Creates `TaskAction` model and table. Three generic task actions would be created with the following commands in the rails console:
```
TaskAction.create!(name: 'Mark task complete', status_after: 'completed')
TaskAction.create!(name: 'Assign to team', status_after: 'on_hold', child_task_owner_type: 'Organization')
TaskAction.create!(name: 'Assign to individual', status_after: 'on_hold', child_task_owner_type: 'User')
```

TaskActions can be called from a controller in the following way:
```
task = Task.find(params[:task_id])
a = TaskAction.find(params[:task_action_id])
a.act(task, params[:child_task_owner_id])
```

### DIFF
--- a/app/models/task_action.rb
+++ b/app/models/task_action.rb
@@ -1,0 +1,37 @@
+class TaskAction < ApplicationRecord
+  enum status_after: {
+    assigned: "assigned",
+    in_progress: "in_progress",
+    on_hold: "on_hold",
+    completed: "completed"
+  }
+
+  def act(task, child_task_owner)
+    update_task_status(task)
+    create_child_task(task, child_task_owner)
+  end
+
+  def update_task_status(task)
+    return unless status_after
+
+    case status_after
+    when "completed"
+      task.mark_complete!
+    when "on_hold"
+      task.update(status: :on_hold)
+    end
+  end
+
+  def create_child_task(task, child_task_owner)
+    return unless child_task_type
+
+    Task.create!(
+      appeal_id: task.appeal_id,
+      assigned_by_id: task.assigned_to_id,
+      appeal_type: task.appeal_type,
+      parent_id: task.id,
+      assigned_to_id: child_task_owner,
+      assigned_to_type: child_task_owner_type
+    )
+  end
+end
--- a/db/migrate/20180820170206_add_task_actions_table.rb
+++ b/db/migrate/20180820170206_add_task_actions_table.rb
@@ -1,0 +1,11 @@
+class AddTaskActionsTable < ActiveRecord::Migration[5.1]
+  def change
+    create_table :task_actions do |t|
+      t.string "name", null: false
+      t.string "status_after"
+      t.string "child_task_owner_type"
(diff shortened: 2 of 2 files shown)


lowellrex:
Still need to write tests for this but am looking for some feedback on this approach to the generic tasks and organizational queues.

db/migrate/20180820170206_add_task_actions_table.rb
```diff
@@ -0,0 +1,12 @@
+class AddTaskActionsTable < ActiveRecord::Migration[5.1]
+  def change
+    create_table :task_actions do |t|
+      t.string "name", null: false
+      t.string "type"
```
lowellrex: Single table inheritance allows us to do something like:
```
class CreateSiblingTaskNotChild < TaskAction
  def act(task, child_task_assignee_id)
    create_child_task(task.parent, child_task_assignee_id)
    update_task_status(task)
  end
end
```
so we could create a task action to "Reassign to team member" that would close the current task and create a new child from the shared parent. I think this is fairly common pattern that could be used when a judge submits a decision that should be routed to dispatch, for instance.

app/models/task_action.rb
```diff
@@ -0,0 +1,37 @@
+class TaskAction < ApplicationRecord
+  enum status_after: {
```
aroltsch: `status_after` is the status after the action is taken?

app/models/task_action.rb
```diff
@@ -0,0 +1,37 @@
...
+    case status_after
+    when "completed"
+      task.mark_complete!
+    when "on_hold"
+      task.update(status: :on_hold)
+    end
+  end
+
+  def create_child_task(task, child_task_assignee_id)
+    return unless child_task_type
```
aroltsch: child_task_assignee_type?

app/models/task_action.rb
```diff
@@ -0,0 +1,37 @@
+class TaskAction < ApplicationRecord
+  enum status_after: {
+    assigned: "assigned",
+    in_progress: "in_progress",
+    on_hold: "on_hold",
+    completed: "completed"
+  }
+
+  def act(task, child_task_assignee_id)
```
aroltsch: It appears to me `child_task_assignee_id` is an optional parameter. We should make it optional.

app/models/task_action.rb
aroltsch: Shall we call it `status_after_action`?

app/models/task_action.rb
lowellrex: I think for now I would rather make `child_task_assignee_id` a required parameter. However, this means that we will have to explicitly pass `nil` in the controller. I imagine the controller will look something like the following:
```
t = Task.find(params['task_id'])
a = TaskAction.find(params['action_id'])
a.act(t, params['assignee_id'])
```

If this stops making sense down the line when we start implementing the controller, I am all for making this an optional parameter.

Issue #6688, opened 2018-08-20
Add hook to create root parent Task from AMA appeals


lowellrex:
The task tree that will enable [generic organizational task queues](https://github.com/department-of-veterans-affairs/caseflow/issues/6573) for AMA appeals relies on a root parent task existing for the entire lifetime of an appeal. The function that creates this root `Task` should be called during the intake process so we can attach child tasks to it as the appeal moves through its life at the board.

## Acceptance criteria
* Add a method to create a new `Task` that will serve as the root task for an AMA appeal

PR #6689, opened 2018-08-20
Add hook to create root parent Task from AMA appeals

lowellrex:
Connects #6688.

Expect this function to be called at the intake process like `t = Task.create_root_task(a.id)`


app/models/task.rb
```diff
@@ -21,6 +21,10 @@ class Task < ApplicationRecord
     completed: "completed"
   }
 
+  def self.create_root_task(appeal_id)
+    create(appeal_id: appeal_id, action: "Root task", assigned_to_type: "Organization", appeal_type: "Appeal")
```
lowellrex: `assigned_to_type` is a required field and I think the type that makes the most sense is "Organization". This field can really be any string, so maybe it should be something more distinct to signal that this is the root task for an appeal (though we already signal that with the `action` field so maybe that would be overkill).

`appeal_type` field is also required. We only allow root tasks for AMA appeals, but maybe we should force the calling code to explicitly state what type of appeal we're creating the root task for.

Thoughts?

amprokop: Hmm "BVA"? "VA"? Either one is fine prob

app/models/task.rb
```diff
@@ -84,4 +85,8 @@ def set_timestamps
       self.completed_at = updated_at if completed?
     end
   end
+
+  def skip_assigned_to_validation
+    false
```
amprokop: nit: if we don't plan on skipping this validation in other cases, it could be a little clearer to have the naming directly apply to this case

`validates :assigned_to, presence: true, unless: :root_task`

Issue #6645, opened 2018-08-15
Add additional attributes to /idt/api/v1/appeals/:appeal_id


amprokop:
The first PR that added the appeal details left out a few attributes.

# AC
- Return all DECASS document assignments (OMO Request & draft decisions) associated with an appeal, including the writing attorney, and document id
- Return a boolean indicating outstanding mail
- Return a boolean indicating if a case has ever been to quality review

PR #6694, opened 2018-08-21
Filters documents array by doc id, corrects "written_by" attribute

amprokop:
Connects to https://github.com/department-of-veterans-affairs/caseflow/issues/6645

Previously, we were returning DECASS records without document ids, we should filter those out. Also, we were returning the CSS ID of the person who the appeal was currently assigned to, not the attorney who wrote the doc.

We also only need the CSS ID of the user who assigned the current task, not for all documents.


app/controllers/idt/api/v1/appeals_controller.rb
```diff
@@ -35,18 +35,32 @@ def json_appeal_details(tasks, appeal)
...
+
+    appeal_details[:data][:attributes][:assigned_by] = assigned_by_user(tasks)
+    appeal_details[:data][:attributes][:documents] = json_documents(tasks)
 
     appeal_details
   end
 
+  def assigned_by_user(tasks)
+    css_id = tasks.sort_by(&:created_at).last.assigned_by_css_id
+    css_id ? User.find_by(css_id: css_id).full_name : ""
```
aroltsch: Are we sure user exists at this point? We also return assigned_by first and last names, you can use that: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/vacols/case_assignment.rb#L103

app/controllers/idt/api/v1/appeals_controller.rb
```diff
@@ -35,18 +35,32 @@ def json_appeal_details(tasks, appeal)
...
+    return appeal_details if tasks.empty?
+
+    appeal_details[:data][:attributes][:assigned_by] = assigned_by_user(tasks)
+    appeal_details[:data][:attributes][:documents] = json_documents(tasks)
 
     appeal_details
   end
 
+  def assigned_by_user(tasks)
+    css_id = tasks.sort_by(&:created_at).last.assigned_by_css_id
```
aroltsch: Theoretically, tasks should be already filtered https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/repositories/queue_repository.rb#L140

app/models/serializers/idt/v1/document_serializer.rb
```diff
@@ -0,0 +1,9 @@
+class Idt::V1::DocumentSerializer < ActiveModel::Serializer
+  def id
+    object.vacols_id
+  end
+  attribute :written_by do
+    object.written_by_css_id ? User.find_by(css_id: object.written_by_css_id).full_name : ""
```
aroltsch: User might not exist in the DB so you will get an error calling full_name on it.

app/controllers/idt/api/v1/appeals_controller.rb
amprokop: Ah, it would be better to use that method, but I'm not using it now. Will change.

app/controllers/idt/api/v1/appeals_controller.rb
amprokop: Ok, yeah using that would make sense.

app/models/serializers/idt/v1/document_serializer.rb
amprokop: Hmm. In practice, I'd bet this will never happen. Attorneys and judges will be using this API, and they will have already signed in to Caseflow. If you think it's worth changing I can do that though!

amprokop: I suppose I could just return the first and last name from the case assignments query, though— that would be less complicated anyway I think.

Issue #6611, opened 2018-08-13
Refactor POA and return first/middle/last separately for appellant


amprokop:
Related to https://github.com/department-of-veterans-affairs/caseflow/issues/6358

# AC
- Refactor POA to use a 1:many relationship instead of 1:1.
- Ensure that we're returning the appellant's POA instead of a potential contesting claimant/etc
- Return first/middle/last of POA to IDT API
- Ensure that updates to POA made during certification are safe, and will not update every POA row if there are multiple rows

PR #6695, opened 2018-08-21
Ensures safe updates to POA

amprokop:
Connects https://github.com/department-of-veterans-affairs/caseflow/issues/6611

Return first/middle/last of POA to IDT API
Ensures that updates to POA made during certification are safe, and will not update every POA row if there are multiple rows

# Test plan
- Unit tests pass


app/models/power_of_attorney.rb
```diff
@@ -37,16 +37,24 @@ def update_vacols_rep_info!(appeal:, representative_type:, representative_name:,
...
-        representative_name: representative_name,
-        address: address
+        rep_name: representative_name,
+        address: address,
+        rep_type: rep_type
       )
     end
   end
 
+  def vacols_representative_name
```
amprokop: this method is here to keep backwards compatibility

app/models/power_of_attorney.rb
```diff
@@ -37,16 +37,24 @@ def update_vacols_rep_info!(appeal:, representative_type:, representative_name:,
       vacols_rep_type: vacols_code
     )
 
-    # If the POA should be stored in the REP table, update that too.
-    if repo.rep_name_found_in_rep_table?(vacols_code)
+    # Update VACOLS with an attorney or agent we found from BGS.
+    if representative_type == "Attorney" || representative_type == "Agent"
```
amprokop: This change restricts the theoretical number of situations we are going to update the rep table. But in practice, we should only update the rep table with attorneys or agents from BGS.

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -85,10 +85,6 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
     false
   end
 
-  attribute :power_of_attorney do
```
amprokop: this looked like a mistake

app/models/vacols/representative.rb
```diff
@@ -55,58 +56,93 @@ def self.update_vacols_rep_type!(bfkey:, rep_type:)
...
-    conn = connection
-    first_name = conn.quote(first_name)
-    middle_initial = conn.quote(middle_initial)
-    last_name = conn.quote(last_name)
-    case_id = conn.quote(bfkey)
+  def self.update_vacols_rep_table!(bfkey:, name:, address:, type:)
+    fail(InvalidRepTypeError) if ACTIVE_REPTYPES[type].empty?
 
-    MetricsService.record("VACOLS: update_vacols_rep_first_name! #{case_id}",
+    MetricsService.record("VACOLS: update_vacols_rep_name! #{bfkey}",
```
aroltsch: Is it update_vacols_rep_table! now?

amprokop:
@aroltsch — there are existing unit tests for the power of attorney repository that pass, and i added new ones for the representative VACOLS model!

Issue #6181, opened 2018-07-11
Editing VBA issues | VBMS links to "Edit Claim Issues" page


abbyraskinUSDS:
When a VBMS user needs to modify claim issues after intake, they can navigate to a "Edit Claim Issues" page in Caseflow, so they can update the the issues on the EP while maintaining the connection to the original rated issue.

## Copy

shanear:
After assessing our client code, it looks like there's enough different about the edit flow to warrant it's own React/Redux "application".

However, there is a lot of shared logic between HLR/SC intake and the edit issues page. For that reason, I'm proposing an `intakeCommon` folder in `client/app` where we can put code and components that are used by multiple intake apps.

The folder structure would then look like:
<img width="208" alt="screen shot 2018-08-17 at 11 58 25 am" src="https://user-images.githubusercontent.com/1596259/44276565-d36dc700-a215-11e8-8b7a-e7e2a28d6c8a.png">

Two components that would definitely be shared are `<RatedIssues/>` and `<NonRatedIssues/>`.. We could also afford to refactor some reducer logic out that deals with ratings and non-rating issues out of `intake/reducers/appeal.js`, `intake/reducers/supplementalClaim.js` and `intake/reducers/higherLevelReview.js` into `intakeCommon` so it could be used in conjunction with the edit page.

kahlouie:
I like that strategy, but also note that this is highly coupled with the work on https://github.com/department-of-veterans-affairs/caseflow/issues/5991 that we may be deferring. I'm not sure if @suzchap and @leikkisa have decided on that. After some discussion with @mdbenjam, it might be better to move some stuff into the selectors file as well to make the client app more in line with what queue is doing.

kahlouie:
To test this, visit `https://appeals.cf.uat.ds.va.gov/intake/higher_level_reviews/600132141/edit`. While you are testing this, also validate #6210 as they are highly connected.

suzchap:
@leikkisa @kahlouie @shanear -- Do I start at VBMS UAT? I tried searching that for the ID provided above and got this message "Share exception thrown in searchBirlsVeteran for FileNumber."

leikkisa:
@suzchap Hi Suzanne. You want to make the intake in Caseflow first, for 123458303 and then go to VBMS.  Not sure what that error is though.

suzchap:
[...]

PR #6705, opened 2018-08-22
Basic content for edit landing page

leikkisa:
Connects #6181


leikkisa:
To Do:
- Fix Veteran name (put first before last), check with Suzanne if this is necessary
- Styling (after talking to Suzanne)
- Fix issue if there are two EPs (rated and non-rated)
- Tests
- Clean up / refactoring
- Check hyphenation/spelling of Higher-level Reviews in mock ups

leikkisa:
screen shot 2018-08-22 at 1 27 59 pm
screen shot 2018-08-22 at 1 28 31 pm

shanear:
screen shot 2018-08-22 at 6 58 17 pm
Can we remove the left padding for the numbered list of issues?

app/views/higher_level_reviews/edit.html.erb
```diff
@@ -1,12 +1,21 @@
...
     feedbackUrl: feedback_url,
-    buildDate: build_date
+    buildDate: build_date,
+    formType: "higher_level_review",
+    veteran: {
+      formName: higher_level_review.veteran.name.formatted(:form),
+      name: higher_level_review.veteran.name,
+      fileNumber: higher_level_review.veteran_file_number
+    },
+    intake: {
```
shanear: I'm not sure we need this data all inside of an `intake` key, especially since we aren't really working with an intake object in the edit page.

All of these values should probably just live in the base level, or under the `review` key

app/views/higher_level_reviews/edit.html.erb
```diff
@@ -1,12 +1,21 @@
...
+      claimId: higher_level_review.end_product_claim_id,
+      receiptDate: higher_level_review.receipt_date,
+      issues: higher_level_review.request_issues,
+      sameOffice: higher_level_review.same_office,
+      informalConference: higher_level_review.informal_conference
```
shanear: it might be nice to break all of this out into a `higher_level_review.ui_hash` method? We could then set that to the `review` key.

app/views/supplemental_claims/edit.html.erb
```diff
@@ -1,12 +1,19 @@
...
     feedbackUrl: feedback_url,
-    buildDate: build_date
+    buildDate: build_date,
+    formType: "supplemental_claim",
+    veteran: {
+      formName: supplemental_claim.veteran.name.formatted(:form),
+      name: supplemental_claim.veteran.name,
+      fileNumber: supplemental_claim.veteran_file_number
+    },
+    intake: {
```
shanear: Same feedback as on `higher_level_reviews/edit.html.erb`

app/assets/stylesheets/_intake.scss
```diff
@@ -61,3 +61,41 @@
...
+  }
+
+  p {
+    margin: .5rem;
+    font-family: 'Source Sans Pro';
+    font-size: 20px;
+    line-height: 1.3em;
+    color: $usa-color-gray;
+  }
+}
```
shanear: this is a lot of additional css here. Were we not able to utilize the style guide for some of this?

client/app/intakeEdit/pages/landing.jsx
```diff
@@ -1,17 +1,100 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
+import Table from '../../components/Table';
+import { FORM_TYPES } from '../../intake/constants';
```
kahlouie: looks like you added this to `client/app/intakeEdit/constants`, should it be taking it in from here? I've created a folder called `client/app/intakeCommon` that can have the shared constants (or other shared stuff).

client/app/intakeEdit/constants.js
```diff
@@ -2,3 +2,8 @@ export const PAGE_PATHS = {
   BEGIN: '/',
   SELECT_ISSUES: '/select_issues'
 };
+
+export const FORM_TYPES = {
+  higher_level_review: '/',
+  supplemental_claim: '/select_issues'
```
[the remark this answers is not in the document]
kahlouie: Also see my comment in the landing page.

app/views/higher_level_reviews/edit.html.erb
kahlouie: I like `review` key as well. Feels more accurate since the intake has already been completed.

leikkisa: I put them in a review key. I was just noticing that we were renaming things a lot for the reducer/connecting/declaring props in variables and this seemed better.  It also works well because some properties are only there for HLR.

app/views/higher_level_reviews/edit.html.erb
leikkisa: I was thinking that too! I would be down for this option.

app/assets/stylesheets/_intake.scss
leikkisa: Hey Shane! I took some of it out.  I talked to Suzanne about this for a while to figure out which Caseflow-wide styles to use, and it seems like there's some discrepancies or issues that she is going to work on, so we decided that the styling on this is somewhat temporary and that we'll revisit.

Some of the styling is pretty specific to the way we did the table and I think that might stay.  I moved the numbers over, they were leaving space for numbers with more than 1 digit, but I think that'll be uncommon.

client/app/intakeEdit/constants.js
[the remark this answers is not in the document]
leikkisa: Lol. I am going to say this is a post-concussion mess up. Thanks!

client/app/intakeEdit/pages/landing.jsx
leikkisa: I was going to wait and do the refactor separately but am okay doing it here too, and we can move things piece by piece. I do want to check in with you on what you set up (so we can coordinate) since it wasn't part of the original framework PR.

leikkisa:
screen shot 2018-08-23 at 4 07 23 pm

Issue #6079, opened 2018-07-02
Co-located table | Pending action tab


laurjpeterson:
As a co-located user, I want to distinguish which cases and tasks are are pending action to come off of hold, so I can potentially reduce the amount of time a Veteran is waiting for a decision.

When an admin action's hold is up, co-located team members want to be alerted, so they can continue working on the case ASAP and move the appeal along in the process.

## AC
- Establish the 'Pending action' tab
- Verify that cases come to the Pending action tab when:
   - Their hold time has completed (days on hold = hold time)
   - New documents have been added to that Veteran's eFolder (#5593)

### User Stories
- When an admin action's hold is up, co-located team members want to be alerted, so they can continue working on the case ASAP and move the appeal along in the process.

### Mocks
https://app.mural.co/t/workqueue2001/m/workqueue2001/1533518524532/52ce941ed69a8a4887b2f6d778e00a4c3597c284

<details><summary>Out of date mock</summary>
![colocated-table-pending-action-tab-768](https://user-images.githubusercontent.com/31224257/43294059-7e8bf330-910b-11e8-8e19-8c384e0f231e.png)
</details>

evankroske:
## Implementation Plan
- [ ] Add selector for pending tasks: 1 hour.
- [ ] Add tab. Estimate: 1 hour.
- [ ] Add column "Total Days Waiting". Estimate: 1 hour.
- [ ] Add a test. Esimate: 1 hour.
- [ ] Get PR approved. Estimate: 1 day.

- Total estimate: 1.5 days.

evankroske:
@mkhandekar , I've got a question about the "Total Days Waiting" column. When a case appears on the pending tab because it has new documents, what should be shown in the column?

mkhandekar:
Answer to Evan's question is: "#A of #"B . (A = number of overdue days, B = number of total days waiting)

See [slack conversation](https://dsva.slack.com/archives/C6E41RE92/p1535035633000200) for context.

PR #6706, opened 2018-08-22
Co-located table | Hide tasks with new documents from on-hold tab

evankroske:
Blocks https://github.com/department-of-veterans-affairs/caseflow/issues/6079

### Description
Hides tasks with new documents from the on-hold tab. (They'll be on the pending tab, instead.)

### Testing Plan
- [ ] Become BVALSPORER.
- [ ] Go to /queue.
- [ ] Open the on-hold tab.
- [ ] Confirm that there aren't any tasks on the tab.

### DIFF
--- a/client/app/queue/QueueLoadingScreen.jsx
+++ b/client/app/queue/QueueLoadingScreen.jsx
@@ -9,9 +9,9 @@ import { LOGO_COLORS } from '../constants/AppConstants';
 import ApiUtil from '../util/ApiUtil';
 import { associateTasksWithAppeals } from './utils';
 
-import { onReceiveQueue, setAttorneysOfJudge, fetchAllAttorneys, fetchAmaTasksOfUser } from './QueueActions';
+import { onReceiveQueue, setAttorneysOfJudge, fetchAllAttorneys, fetchAmaTasksOfUser, getNewDocuments } from './QueueActions';
 import { setUserId } from './uiReducer/uiActions';
-import type { Appeals, Tasks } from './types/models';
+import type { BasicAppeals, Tasks } from './types/models';
 import type { State, UsersById } from './types/state';
 import USER_ROLE_TYPES from '../../constants/USER_ROLE_TYPES.json';
 
@@ -28,7 +28,7 @@ type Params = {|
 type Props = Params & {|
   // From state
   tasks: Tasks,
-  appeals: Appeals,
+  appeals: BasicAppeals,
   amaTasks: Tasks,
   loadedUserId: number,
   judges: UsersById,
@@ -36,7 +36,7 @@ type Props = Params & {|
   onReceiveQueue: typeof onReceiveQueue,
   setAttorneysOfJudge: typeof setAttorneysOfJudge,
   fetchAllAttorneys: typeof fetchAllAttorneys,
-  fetchAmaTasksOfUser: Function,
+  fetchAmaTasksOfUser: (number, string) => Promise<{ payload: { amaTasks: Tasks, appeals: BasicAppeals } }>,
   setUserId: typeof setUserId
 |};
 
@@ -54,8 +54,9 @@ class QueueLoadingScreen extends React.PureComponent<Props> {
       return Promise.resolve();
     }
 
+    this.props.setUserId(userId);
     return this.props.fetchAmaTasksOfUser(userId, userRole).
-      then(() => this.props.setUserId(userId));
+      then(({ payload: { appeals } }) => _.map(appeals, (appeal) => this.props.getNewDocuments(appeal.externalId)));
   }
 
   maybeLoadLegacyQueue = () => {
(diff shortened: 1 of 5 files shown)


client/app/queue/QueueLoadingScreen.jsx
```diff
@@ -54,8 +61,12 @@ class QueueLoadingScreen extends React.PureComponent<Props> {
       return Promise.resolve();
     }
 
+    this.props.setUserId(userId);
+
     return this.props.fetchAmaTasksOfUser(userId, userRole).
-      then(() => this.props.setUserId(userId));
+      then(
+        ({ payload: { appeals: appealsFromServer } }) =>
+          _.map(appealsFromServer, (appeal) => this.props.getNewDocuments(appeal.externalId)));
```
dannysellers: Was this moved earlier in the loading process for the selector? It looks like we end up making this request twice per appeal now (presumably because the first one doesn't finish by the time the `NewFile` is rendered)

client/app/queue/selectors.js
```diff
@@ -100,10 +100,15 @@ export const newTasksByAssigneeCssIdSelector = createSelector(
...
 export const onHoldTasksByAssigneeCssIdSelector: (State) => Array<Task> = createSelector(
-  [tasksByAssigneeCssIdSelector],
-  (tasks: Array<Task>) =>
-    tasks.filter((task) => moment().diff(moment(task.placedOnHoldAt), 'days') < task.onHoldDuration)
+  [tasksByAssigneeCssIdSelector, getNewDocsForAppeal],
+  (tasks: Array<Task>, newDocsForAppeal: NewDocsForAppeal) =>
+    tasks.filter(
+      (task) =>
+        moment().diff(moment(task.placedOnHoldAt), 'days') < task.onHoldDuration &&
+          (!newDocsForAppeal[task.externalAppealId] || newDocsForAppeal[task.externalAppealId].length === 0))
```
dannysellers: This feels pretty unwieldy. Could the no-documents selector be extracted? We might not need to use it anywhere else / might not need to export it, but no-documents feels like a different check than status/on-hold or assignment.

client/app/queue/selectors.js
```diff
@@ -100,10 +100,15 @@ export const newTasksByAssigneeCssIdSelector = createSelector(
...
 );
 
+const getNewDocsForAppeal = (state: State) => state.queue.newDocsForAppeal;
+
 export const onHoldTasksByAssigneeCssIdSelector: (State) => Array<Task> = createSelector(
-  [tasksByAssigneeCssIdSelector],
-  (tasks: Array<Task>) =>
-    tasks.filter((task) => moment().diff(moment(task.placedOnHoldAt), 'days') < task.onHoldDuration)
+  [tasksByAssigneeCssIdSelector, getNewDocsForAppeal],
+  (tasks: Array<Task>, newDocsForAppeal: NewDocsForAppeal) =>
```
dannysellers: Is there an advantage to annotating this argument, rather than annotating the return value of `getNewDocsForAppeal`? Like
```javascript
const getNewDocsForAppeal = (state: State): NewDocsForAppeal => state.queue.newDocsForAppeal;
```

client/app/queue/QueueLoadingScreen.jsx
```diff
@@ -28,16 +34,17 @@ type Params = {|
...
   amaTasks: Tasks,
   loadedUserId: number,
   judges: UsersById,
   // Action creators
   onReceiveQueue: typeof onReceiveQueue,
   setAttorneysOfJudge: typeof setAttorneysOfJudge,
   fetchAllAttorneys: typeof fetchAllAttorneys,
-  fetchAmaTasksOfUser: Function,
-  setUserId: typeof setUserId
+  fetchAmaTasksOfUser: (number, string) => Promise<{ payload: { amaTasks: Tasks, appeals: BasicAppeals } }>,
```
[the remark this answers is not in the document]
evankroske: The problem is that the signature of fetchAmaTasksOfUser in QueueActions is `(number, string) => (Dispatch) => Promise<{ payload: { amaTasks: Tasks, appeals: BasicAppeals } }>`. bindActionCreators strips the dispatch argument, but I don't know how to do the same. We can usually use typeof because we don't care about the return value of the action creator

client/app/queue/QueueLoadingScreen.jsx
evankroske: I had to start preloading documents for appeals because NewFile only loads the documents of an appeal when it's rendered. If I didn't preload the documents, when someone switched to the on-hold tab, some appeals would disappear in front of their eyes. Also, if they looked at the pending tab, some appeals would be missing. All of the appeals with new documents would be missing because their documents wouldn't be loaded.

However, I'll fix the double request problem by adding a loading field to short-circuit the second request for an appeal's documents

client/app/queue/selectors.js
evankroske: Yeah, the document count logic was out of control. I extracted it into its own function. I didn't put it in a new selector because the pending tab wouldn't be able to reuse it. However, I will be able to use the new function in the selector for pending tasks. What do you think?

client/app/queue/selectors.js
evankroske: Unfortunately, we don't have the Flow library definition for reselect installed, so Flow can't infer the type of newDocsForAppeal

client/app/queue/QueueLoadingScreen.jsx
evankroske: Fixed!

PR #6713, opened 2018-08-22
Enable generic task actions for organizational queues on the backend

### LINKED ISSUE #6704: Add endpoints to receive generic task actions
In order to build [generic organizational task queues](https://github.com/department-of-veterans-affairs/caseflow/issues/6573), we need to support a small set of generic task actions for teams and individuals using organizational task queues. On the back-end we need to be able to receive requests for each of the three generic task actions. At a bare minimum these actions need to have the following parameters and consequences:

### Mark task complete
*Parameters*: task ID
*Consequences*: Change state of task to "complete". Update parent task if applicable (because of the work in #6677 simply calling `Task.find(params['task_id']).mark_as_complete!` should do the trick).

### Assign to team
*Parameters*: task ID, assignee ID
*Consequences*: Change state of task to "on hold". Create child task assigned to the specified assignee.

### Assign to individual
*Parameters*: task ID, assignee ID
*Consequences*: Change state of task to "on hold". Create child task assigned to the specified assignee. If the current assignee is an individual, open a sibling task off of the parent task assigned to the organization and close the other child task (if it is easier to add this as a separate route then we can move this logic to the front-end as well).

lowellrex:
Resolves #6704.

The routes and logic in the controller exists to support generic task actions, we just need to add a new task type (`GenericTask`) with a few methods.

This PR enables the three generic task actions by calling the appropriate endpoints with request bodies as follows:

### Mark task complete
```
PATCH /tasks/:id
{ type: "GenericTask" }
```

### Assign to team
```
POST /tasks
{
  type: "GenericTask",
  parent_id: {parent_id},
  assigned_to_id: {assigned_to_id},
  assigned_to_type: "Organization",
  status: {status}  # (optional) status of parent task
}
```

### Assign to individual
```
POST /tasks
{
  type: "GenericTask",
  parent_id: {parent_id},
  assigned_to_id: {assigned_to_id},
  assigned_to_type: "User",
  status: {status}  # (optional) status of parent task
}
```


app/models/task.rb
```diff
@@ -29,6 +29,14 @@ def assigned_by_display_name
...
   end
 
+  def create_from_params(params)
+    create(params)
+  end
+
+  def update_from_params(params)
+    update(params)
+  end
+
```
lowellrex: Renamed these methods so we could easily overwrite them in `GenericTask`. Alternatively we could overwrite the built-in ActiveRecord methods in `GenericTask`, but then we lose the ability to call those built-ins for `GenericTask`.

app/models/tasks/generic_task.rb
```diff
@@ -0,0 +1,39 @@
...
+    # Create an assignee from the input arguments so we throw an error if the assignee does not exist.
+    assignee = Object.const_get(params[:assigned_to_type]).find(params[:assigned_to_id])
+    parent = Task.find(params[:parent_id])
+
+    GenericTask.create!(
+      appeal_id: parent.appeal_id,
+      assigned_by_id: parent.assigned_to_id,
+      appeal_type: parent.appeal_type,
+      parent_id: parent.id,
+      assigned_to_id: assignee.id,
```
aroltsch: I think if you pass an object from line 16 you don't need to specify the assigned_to_type anymore. 

```
assigned_to: assignee
```

The same for the appeal object.

PR #6718, opened 2018-08-23
Jobs | Sets Sentry context on the worker thread

evankroske:
### Description
Changes ApplicationJob to add the app name to the Sentry context from the job's worker thread. This ensures that job errors are routed to the right channel. See also https://github.com/department-of-veterans-affairs/caseflow/pull/6649 and https://github.com/department-of-veterans-affairs/caseflow/pull/6624.

### Testing Plan
- [ ] Merge this PR.
- [ ] Deploy Caseflow.
- [ ] Confirm that BGSEndProductSyncError doesn't start showing up in #appeals-app-alerts.

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -1,10 +1,14 @@
 class ApplicationJob < ActiveJob::Base
-  def self.application_attr(app_name)
-    @app_name = app_name
+  class << self
+    def application_attr(app_name)
+      @app_name = app_name
+    end
+    
+    attr_reader :app_name
   end
 
   before_perform do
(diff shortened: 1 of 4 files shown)


app/jobs/application_job.rb
```diff
@@ -1,6 +1,14 @@
...
-  def self.application_attr(app_name)
+  class << self
+    def application_attr(app_name)
+      @app_name = app_name
+    end
+
+    attr_reader :app_name
+  end
+
+  before_perform do
```
sunil-sadasivan: ah didn't realize we could do `before_perform` very cool

PR #6729, opened 2018-08-24
Anya/6699 ama appeals for idt

### LINKED ISSUE #6699: Return AMA appeals from /idt/api/v1/appeals

aroltsch:
Resolves #6699 

### Description
Allow attorneys and judges to view AMA appeals assigned to them

Judge BVAAABSHIRE
image

Attorney BVASCASPER1
image


app/models/serializers/idt/v1/appeal_serializer.rb
```diff
@@ -1,7 +1,6 @@
 class Idt::V1::AppealSerializer < ActiveModel::Serializer
-  # TODO: serialize AMA appeals with this serializer
   def id
-    object.vacols_id
+    object.is_a?(LegacyAppeal) ? object.vacols_id : object.id
```
amprokop: should this be the uuid (e.g. "00000000-0000-0000-0000-000000000000" and not the db id (e.g. 1)? I think we'd expect the IDT to pass this back to our API to look up details for an appeal

aroltsch: you are probably right sir!

Issue #6679, opened 2018-08-20
Add back-end support for generic task actions dropdown lists


lowellrex:
In order to build [generic organizational task queues](https://github.com/department-of-veterans-affairs/caseflow/issues/6573), we need to pass the list of available teams and users within a team that can be assigned tasks from the back-end to the front (the list of available teams could just be defined in code, but we almost certainly need to go to the back-end for the list of users).

## Acceptance criteria
* Add endpoint for fetching the list of users capable of being assigned a task
* Add endpoint (or hard-code) to return list of teams that can be assigned a task

Issue #6724, opened 2018-08-24
Determine which organizations a Caseflow user belong to


lowellrex:
In order to allow folks to visit [generic organizational task queues](https://github.com/department-of-veterans-affairs/caseflow/issues/6573) for various organizations within the Board, we need to know if they are a part of those organizations. The ways by which we determine if somebody can access a queue differs for each queue that currently exists. For VSOs, we check that the user has the [VSO CSEM function](https://github.com/department-of-veterans-affairs/caseflow/blob/9042b27f3284cddd7bae38ab2fc0d469843dc3a2/app/models/user.rb#L128) and that the user is [associated with the VSO in BGS](https://github.com/department-of-veterans-affairs/caseflow/blob/9042b27f3284cddd7bae38ab2fc0d469843dc3a2/app/models/organizations/vso.rb#L2). For attorneys, judges, and co-located admin, we rely on different fields in VACOLS `staff` table to [determine the board employee's role](https://github.com/department-of-veterans-affairs/caseflow/blob/9042b27f3284cddd7bae38ab2fc0d469843dc3a2/app/repositories/user_repository.rb#L70). We will probably do something similar to the latter approach as we enable access to organizational task queues for other Board teams.

In addition to determining if a single user is a member of an organization, we need to determine which users belong to an organization so we can build the list of folks to populate the "Assign to team member" dropdown.

## Acceptance criteria
* Determine if a user belongs to an organization
* Retrieve list of users who belong to an organization

laurjpeterson:
Closing. Mark has built a way for our users to map individuals to teams and we acquired the October 2018 BVA org chart.

PR #6730, opened 2018-08-24
Add route to return list of members in organization/organizations

lowellrex:
Connects #6679.

I'd recommend starting at the routes and following them through the controller to the model in this review. Still need to add tests but am looking to validate this approach before I dive to deep into test writing.

Tested manually by doing the following:
```
rails c> o = Organization.create!(name: "Translation", feature: "org_queue_translation", url: "translation")
rails c> u = User.create!(station_id: "101", css_id: "ORG_QUEUE_USER", full_name: "Org Q User")
rails c> FeatureToggle.enable!(o.feature.to_sym, users: [u.css_id])
```

### /members route
* Visit `http://localhost:3000/organizations/translation/members` as "Org Q User" and see the list of user associated with the Translation organization.
* Visit `http://localhost:3000/organizations/translation/members` as anybody else and get redirected to unauthorized page.

### /index route
* Visit `http://localhost:3000/organizations` as any user and see the list of internal Board organizations using queue.


app/controllers/organizations_controller.rb
```diff
@@ -8,14 +8,18 @@ def show
     render "organizations/show"
   end
 
+  def members
+    render json: { members: organization.members.map { |m| { id: m.id, css_id: m.css_id, full_name: m.full_name } } }
```
lowellrex: Initially had this as a serializer, but then we have to unpack the attributes property on the front-end. This is also far less boilerplate because the few fields we are accessing are all simple types (int, str, str).

app/models/organization.rb
```diff
@@ -1,7 +1,17 @@
 class Organization < ApplicationRecord
   has_many :tasks, as: :assigned_to
 
-  def user_has_access?(_user)
-    false
+  def user_has_access?(user)
+    members.pluck(:id).include?(user.id)
+  end
```
lowellrex: Until work on #6724 is complete, we will be using FeatureToggles to grant access to different organization queues.

app/controllers/organizations_controller.rb
```diff
@@ -8,14 +8,24 @@ def show
...
+  end
+
+  def list
+    render json: {
+      organizations: Organization.all.reject { |o| o.type && o.type == "Vso" }.map { |o| { id: o.id, name: o.name } }
+    }
```
lowellrex: Do not return VSOs in the list of organizations because this we only want to get a list of internal Board teams.

app/controllers/organizations_controller.rb
```diff
@@ -8,14 +8,24 @@ def show
...
+  end
+
   private
 
   def verify_organization_access
-    redirect_to "/unauthorized" unless organization.user_has_access?(current_user)
+    redirect_to "/unauthorized" unless organization && organization.user_has_access?(current_user)
```
lowellrex: Redirect if there is no organization for the given address.

app/controllers/organizations_controller.rb
```diff
@@ -8,14 +8,24 @@ def show
...
   end
 
   def verify_role_access
-    verify_authorized_roles(organization.role)
+    verify_authorized_roles(organization.role) if organization.role
```
lowellrex: Allow for folks with many different roles to have access to a given organization.

app/controllers/organizations_controller.rb
```diff
@@ -33,7 +43,9 @@ def organization_url
   end
 
   def organization
-    Organization.find_by(url: organization_url)
+    # Allow the url to be the ID of the row in the table since this will be what is associated with
+    # tasks assigned to the organization in the tasks table.
+    Organization.find_by(url: organization_url) || Organization.find(organization_url)
```
lowellrex: Allow us to create the Organization object with the ID as well as the URL slug. This is necessary when we navigate directly to a case details page and the currently active task for that case is assigned to a team (and the current user is a member of the organization to which the task is assigned so we can see the task actions dropdown). The task properties return an `assigned_to_id` and `assigned_to_type` but no property that indicates the organization's URL slug (if the task is assigned to an organization).

This addition (find by ID as well as URL) allows us fetch the members of the organization with the information returned by the task so that we can build the "Assign to team member" task action without having to make an additional request to determine the organization's URL slug.

app/controllers/organizations_controller.rb
```diff
@@ -1,21 +1,31 @@
...
+  before_action :verify_organization_access, except: [:index]
+  before_action :verify_role_access, except: [:index]
+  before_action :verify_feature_access, except: [:index]
   before_action :set_application
 
+  def index
+    render json: {
+      organizations: Organization.all.reject { |o| o.type && o.type == "Vso" }.map { |o| { id: o.id, name: o.name } }
+    }
+  end
```
lowellrex: Do not return VSOs in the list of organizations because this we only want to get a list of internal Board teams.

app/controllers/organizations_controller.rb
```diff
@@ -1,21 +1,31 @@
 class OrganizationsController < ApplicationController
-  before_action :verify_organization_access
-  before_action :verify_role_access
-  before_action :verify_feature_access
+  before_action :verify_organization_access, except: [:index]
+  before_action :verify_role_access, except: [:index]
+  before_action :verify_feature_access, except: [:index]
```
lowellrex: Allow non-organizational queue users access to the list of organizations at the Board.

PR #6746, opened 2018-08-27
Determine is user is a dispatch user

aroltsch:
For IDT, we need to check if user is a Dispatch user.

### DIFF
--- a/app/repositories/user_repository.rb
+++ b/app/repositories/user_repository.rb
@@ -68,6 +68,7 @@ class UserRepository
     private
 
     def roles_based_on_staff_fields(staff_record)
+      return [] if staff_record.sactive != "A"
       case staff_record.svlj
       when "J"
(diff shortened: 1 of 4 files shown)


app/repositories/user_repository.rb
```diff
@@ -68,6 +68,7 @@ def css_id_by_full_name(full_name)
     private
 
     def roles_based_on_staff_fields(staff_record)
+      return [] if staff_record.sactive != "A"
```
evankroske: Hey, I'm not really sure what the purpose of this code is. It looks like it prohibits all inactive staff from accessing Queue, but I'm not sure why that's necessary. Doesn't CSS do this for us? I'm asking because some of our test users don't have sactive set to "A", and I'm worried that some of our prod users might not, either. What do you think?

PR #6747, opened 2018-08-27
Queue | Better manage browser history during checkout flows

dannysellers:
### Description
During usability testing last week with a user, we observed that the user would occasionally click the browser's Back button after canceling or completing a checkout flow, only for the site to crash. This happened because during attorney/judge checkout flows, we 'stage' appeals to differentiate them from the version loaded from the server / to allow editing of appeal attributes while preserving a clean copy. Because staging appeals only happens when starting a checkout flow, going backwards to a page that requires a staged appeal will cause the site to crash.

This PR adjusts how we manage browser history during attorney/judge checkout flows. Previously, we would `push` new pages, allowing the user to click the browser back/forwards buttons to navigate between pages within a flow. Now, though, we `push` _one_ new history entry at the start of a checkout flow, and `replace` the path of that entry when navigating through flows. So, if the user clicks the back button, they would be taken back to `/queue/appeals/${appealId}` regardless of where they are in a flow. This PR also adds a confirmation popup when clicking the back button while inside a flow indicating to the user that navigating away will cancel the checkout flow.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to `/queue` as attorney or judge
2. Begin case review checkout flow, confirm clicking back button within flow returns to `/queue/appeals/${appealId}` after prompt
3. After completing flow, confirm clicking back button returns to `/queue/appeals/${appealId}`, reloading the case (task assignment information should update)

</details>


client/app/queue/CaseSnapshot.jsx
```diff
@@ -10,7 +10,7 @@ import {
   tasksForAppealAssignedToUserSelector
 } from './selectors';
 import CaseDetailsDescriptionList from './components/CaseDetailsDescriptionList';
-import SelectCheckoutFlowDropdown from './components/SelectCheckoutFlowDropdown';
+import AttorneyActionsDropdown from './components/AttorneyActionsDropdown';
```
evankroske: 👏

PR #6748, opened 2018-08-27
Improve root task for appeal

lowellrex:
Connects #6688.

Improves on the work done in PR #6689 by making the "Vso" type into a real type and assigning the root task to that organization.


db/seeds.rb
```diff
@@ -296,6 +296,7 @@ def create_organizations
       url: "american-legion",
       participant_id: "2452415"
     )
+    Bva.create(name: "Board of Veterans' Appeals")
```
lowellrex: When this PR is merged I will run this statement in prod to create the record we will depend on in `RootTask`.

mdbenjam: How about adding a migration here to add that row, so we don't need to run it in the prod console?

PR #6750, opened 2018-08-27
Queue | Misc checkout flow fixes

dannysellers:
Resolves a few issues Anya noticed during testing

### Description
- For AMA tasks, we need to pass a parent ID when creating a Colocated admin action
- Removes `is_legacy_appeal` from Legacy/AppealSerializer, uses `docket_name` instead
- When displaying AMA tasks for attorneys, `on_hold` tasks should be hidden from New / default tab (On hold tab doesn't exist yet for attorneys)
- Update copy on Colocated Pending tab to indicate presence of tasks with new documents

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. As an attorney (`BVASCASPER1`), test adding colocated admin action for AMA appeal. Confirm `parent_id` sent
2. As an attorney, confirm on hold AMA appeals are hidden from table view


client/app/queue/AddColocatedTaskView.jsx
```diff
@@ -56,14 +61,28 @@ class AddColocatedTaskView extends React.PureComponent<Props, ComponentState> {
 
   validateForm = () => Object.values(this.state).every(Boolean);
 
+  buildPayload = () => {
+    const { tasks, appeal } = this.props;
+
+    return _.map([tasks[0]], (task: Task) => {
```
dannysellers: This is anticipating adding multiple colocated actions

client/app/queue/AddColocatedTaskView.jsx
```diff
@@ -56,14 +61,28 @@ class AddColocatedTaskView extends React.PureComponent<Props, ComponentState> {
...
+      const mapped: Object = {
```
dannysellers: Without this annotation, Flow complains "Cannot assign `parent_id` to `mapped.parent_id` because property `parent_id` is missing in object literal" 😕

aroltsch:
Hey @dannysellers, thank you so much 

Testing Results:
1. If I assign AMA appeal to a colocated user, the screen goes blank for attorney
2. When sending a parent ID to create a collocated task, it should be an ID of the attorney task not the judge task ID.  Attorney task is a parent for a colocated task.

dannysellers:
@aroltsch thanks for catching that error. As for sending the parent ID of an attorney task vs a judge task—I'm not sure we're able to make that distinction on the front end. We only maintain one store of `tasks`, which, depending on the user and their role, might be `ColocatedTask`s, `JudgeTask`s, etc, but we don't differentiate between task types. We'd have to adjust the `TaskSerializer`

aroltsch:
Testing results:
1. When creating a collocated admin action for a legacy appeal, it sets the task’s parent to vacols ID. For legacy appeals, do not pass parent ID
image

Not required as part of this PR 
2. 2. When clicked into details of the task, we need to PATCH /tasks/:id with status: :in_progress to move the task from assigned to in_progress

app/models/serializers/work_queue/legacy_appeal_serializer.rb
```diff
@@ -86,4 +82,8 @@ class WorkQueue::LegacyAppealSerializer < ActiveModel::Serializer
   attribute :caseflow_veteran_id do
     object.veteran ? object.veteran.id : nil
   end
+
+  attribute :docket_name do
```
dannysellers: This is needed because `CaseDetailsLoadingScreen` calls [`utils.prepareAppealForStore`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/utils.js#L138), which attempts to use `appeal.attributes.docket_name`. Appeal/LegacyAppealSerializers have never had a `docket_name` attribute (it's part of TaskSerializers), so any page that relies on `appeal.docketName` which is wrapped by `CaseDetailsLoadingScreen` / comes after a page wrapped by it, would previously fail to access that attribute... 🤦‍♂️

Issue #6186, opened 2018-07-11
AMA Decisions | Sync dispositions for AMA higher level reviews and supplemental claims


shanear:
As part of our end product sync job, we should also be syncing the dispositions for the contentions on those end products

## Acceptance Criteria
- When an end product is being synced with VBMS, if the EP is in cleared status "CLR", and the EP is for an AMA higher level review or supplemental claim and the dispositions for that EP haven't already been recorded, then the system syncs the dispositions for each of the issues on that HLR or SC by querying the dispositions for the associated contentions.
- If one or more of the dispositions does not exist, then an error should be raised. We should always have a disposition for each contention.

## Additional Notes
If a disposition is a DTA error, then this will eventually trigger the creation of a supplemental claim. This will be handled by the following story
blocks #6167

## Validation
True validation for this story is going to be fairly tricky, since we need the claim to go through the entire rating and awards process to get cleared and have dispositions. I think we should validate all of the decisions stories at one time with Mary Kate or Colin

- Validate that this works for comp & non-comp

kahlouie:
Edit: This issue assumed that we were already syncing Claim Reviews, which we weren't. We now need to move sync job to end product establishments table.
This is also changing the way we sync ramp election eps.

Then we can find the source, if it's a claim review, sync the disposition. Don't worry about error handling yet.

shanear:
Validating this with the following EP:

```
EndProductEstablishment.find_by(veteran_file_number: "123458311", modifier: "032")
```

shanear:
This is blowing up 😢

```
[2018-10-04 13:52:42 -0400] STARTED sent VBMS request VBMS::Requests::GetDispositions for 600138634
[2018-10-04 13:52:43 -0400] FINISHED sent VBMS request VBMS::Requests::GetDispositions for 600138634:   0.018987   0.000308   0.019295 (  0.449921)

[2018-10-04 13:52:43 -0400]    (1.6ms)  ROLLBACK
Traceback (most recent call last):
       10: from (irb):9
        9: from app/models/end_product_establishment.rb:145:in `sync!'
        8: from app/models/end_product_establishment.rb:151:in `block in sync!'
        7: from app/models/end_product_establishment.rb:300:in `sync_source!'
        6: from app/models/higher_level_review.rb:52:in `on_sync'
        5: from app/models/claim_review.rb:73:in `on_sync'
        4: from app/models/higher_level_review.rb:92:in `sync_dispositions'
[...]

PR #6761, opened 2018-08-27
Fix ramp election ep syncing and add syncing of claim review eps

kahlouie:
Connects #6186


spec/models/ramp_election_spec.rb
```diff
@@ -82,20 +82,19 @@
...
       expect(ramp_election).to receive(:recreate_issues_from_contentions!)
-      expect(ramp_election).to receive(:sync_ep_status!)
 
       subject
     end
 
     context "when error is raised" do
       before do
-        expect(ramp_election).to receive(:sync_ep_status!).and_raise(ActiveRecord::RecordInvalid)
+        # TODO: What kind of error would be realistic to be raised here?
```
kahlouie: Please note the question I have posed here. Wanted to make sure this is still the right way to check the rescue case, but also that we're checking with a potentially realistic error.

app/models/ramp_election.rb
```diff
@@ -74,42 +74,33 @@ def rollback!
   end
 
   def sync!
-    create_end_product_establishment_if_missing
```
shanear: confirmed this is no longer needed

app/jobs/ramp_election_sync_job.rb
```diff
@@ -4,8 +4,7 @@ class RampElectionSyncJob < CaseflowJob
   application_attr :intake
 
   def perform(ramp_election_id)
-    RequestStore.store[:current_user] = User.system_user
-
-    RampElection.find(ramp_election_id).sync!
+    # TODO: DELETE ME
+    # we only didn't remove this to prevent a bunch of queued up job failures
```
shanear: well delete this in a follow up pr

app/models/ramp_election.rb
```diff
@@ -49,9 +40,7 @@ def recreate_issues_from_contentions!
   end
 
   def successful_intake
-    @successful_intake ||= intakes.where(completion_status: "success")
-      .order(:completed_at)
-      .last
+    @successful_intake ||= intakes.where(completion_status: "success").order(:completed_at).last
```
shanear: just changing a multi-liner to a single liner for 💅

app/models/ramp_closed_appeal.rb
```diff
@@ -38,7 +38,7 @@ def reclose!
 
     # If the end product was canceled, don't re-close the VACOLS appeal.
     # Instead rollback the RAMP election data from Caseflow
-    return ramp_election.rollback! if ramp_election.end_product_canceled?
+    return ramp_election.rollback! if ramp_election.end_product_establishment.status_canceled?
```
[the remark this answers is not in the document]
kahlouie: I'll fix this in the next pr, so we can get this out sooner.

PR #6762, opened 2018-08-28
adding an additional check to see if sp can be finalized

### LINKED ISSUE #6522: Review Assignments | Cancel Schedule Period
A user could, theoretically, create a schedule period, cancel out on the Review Assignments page, upload a different schedule period to VACOLS, then return to the cancelled schedule period's page, and accidentally upload more hearings. We need to prevent that from happening.

tejans24:
Resolves #6522 

### Description
Adds another check for schedule period to see if it can be processed to submit to VACOLS

### Acceptance Criteria
- [ ] Test plan executed

### Testing Plan
1. Upload the RO spreadsheet for periods 10-01-2018 to 04-30-2019
2. Upload the same spreadsheet again for the same periods
3. Submit the 1st set of assignments to VACOLS
4. Try to submit the 2nd set of assignments to VACOLS
5. Notice it fail with an error
5. Try entering the URLs of the uploads i.e (http://localhost:3000/hearings/schedule/build/upload/1 and 2)
6. Page expiration needs to be shown to users

[ROAssignmentTemplate-final.xlsx](https://github.com/department-of-veterans-affairs/caseflow/files/2329798/ROAssignmentTemplate-final.xlsx)


app/models/schedule_period.rb
```diff
@@ -38,8 +38,15 @@ def schedule_confirmed(*)
     update(finalized: true)
   end
 
+  def dates_already_finalized?
+    sps = SchedulePeriod.all.select(&:finalized)
```
sharonwarner: We have similar logic in validate_schedule_period on line 14. I think we could just run that validation on update as well instead of adding this check. 

We definitely need to include type in the check too.

tejans24: I updated both of the methods to use the same logic. But on update would mean that the VACOLS upload will still happen before the SchedulePeriod is updated to finalized.

sharonwarner: If it's in a transaction, the VACOLS updates should rollback if the update doesn't go through correctly.

Issue #6725, opened 2018-08-24
Prevent unauthorized users from taking actions on generic tasks


lowellrex:
#6713 enabled back-end support for taking action on generic tasks. However, there are no limits on who can take those actions. We should constrain task actions from happening for generic tasks unless the user is the task owner (if the task is assigned to an individual) or belongs to that organization (if the task is assigned to an organization). In addition to enforcing this on the back-end, we should not display "illegal" task actions on the front-end.

## Acceptance criteria
* Prevent users from action on tasks they don't own (or their org. doesn't own)
* Hide illegal actions from the actions dropdown for a task

PR #6766, opened 2018-08-28
Add authorization checks to GenericTask

lowellrex:
Connects #6725 (only the back-end portion of the ticket).

Adds check to confirm that the user attempting to update or create a `GenericTask` is allowed to. I think a good place to start the review is in GenericTask.rb.


app/controllers/application_controller.rb
```diff
@@ -15,6 +15,12 @@ class ApplicationController < ApplicationBaseController
     render json: { "errors": ["title": e.class.to_s, "detail": e.message] }, status: 400
   end
 
+  rescue_from StandardError do |e|
+    fail e unless e.class.method_defined?(:serialize_response)
+    Raven.capture_exception(e)
+    render(e.serialize_response)
+  end
```
lowellrex: This is not strictly necessary for this PR, but it does add the ability for errors that are uncaught here to serialize themselves. Which I think is a nice addition.

app/controllers/tasks_controller.rb
```diff
@@ -2,7 +2,6 @@ class TasksController < ApplicationController
   include Errors
 
   before_action :verify_queue_access
-  before_action :verify_task_assignment_access, only: [:create]
```
lowellrex: Moved to the Task model so that we don't do this for GenericTasks going through the create action.

lowellrex: Also, the routine being called here (`verify_task_assignment_access`) still needs to stick around because we use it in the LegacyTasksController.

mdbenjam: Let's remove the `verify_task_assignment_access` method as well.

PR #6772, opened 2018-08-28
Make VSO user in seed data functional

lowellrex:
@nikitarockz discovered that our VSO user in the test data does not actually have access to the VSO queue. This PR fixes that.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -44,7 +44,14 @@ class SeedDB
 
     Functions.grant!("System Admin", users: User.all.pluck(:css_id))
 
-    User.create(css_id: "VSO", station_id: 101, full_name: "VSO user associated with american-legion")
+    u = User.create(
+      css_id: "VSO",
+      station_id: 101,
+      full_name: "VSO user associated with american-legion",
+      roles: ["VSO"]
+    )
+    FeatureToggle.enable!(:vso_queue_aml, users: [u.css_id])
+    Functions.grant!("VSO", users: [u.css_id])
   end
 
   def create_dispatch_tasks(number)
@@ -292,6 +299,7 @@ class SeedDB
   def create_organizations
     Vso.create(
       name: "American Legion",
+      feature: "vso_queue_aml"
       role: "VSO",
       url: "american-legion",
       participant_id: "2452415"
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -51,7 +51,6 @@ class SeedDB
       roles: ["VSO"]
     )
     FeatureToggle.enable!(:vso_queue_aml, users: [u.css_id])
-    Functions.grant!("VSO", users: [u.css_id])
   end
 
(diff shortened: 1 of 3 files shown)


db/seeds.rb
```diff
@@ -44,7 +44,14 @@ def create_users
 
     Functions.grant!("System Admin", users: User.all.pluck(:css_id))
 
-    User.create(css_id: "VSO", station_id: 101, full_name: "VSO user associated with american-legion")
+    u = User.create(
+      css_id: "VSO",
+      station_id: 101,
+      full_name: "VSO user associated with american-legion",
+      roles: ["VSO"]
```
lowellrex: Add role to pass [`OrganizationsController.verify_role_access()`](https://github.com/department-of-veterans-affairs/caseflow/blob/2fc9a9ac7681c6475b70d1ae58943c4cfea63592/app/controllers/organizations_controller.rb#L17) check.

db/seeds.rb
```diff
@@ -44,7 +44,14 @@ def create_users
...
+    )
+    FeatureToggle.enable!(:vso_queue_aml, users: [u.css_id])
```
lowellrex: Enable feature to pass [`OrganizationsController. verify_feature_access()`](https://github.com/department-of-veterans-affairs/caseflow/blob/2fc9a9ac7681c6475b70d1ae58943c4cfea63592/app/controllers/organizations_controller.rb#L21) check.

db/seeds.rb
```diff
@@ -44,7 +44,14 @@ def create_users
...
+    Functions.grant!("VSO", users: [u.css_id])
```
lowellrex: Grant function to pass [`ApplicationController. can_access_queue?()`](https://github.com/department-of-veterans-affairs/caseflow/blob/2fc9a9ac7681c6475b70d1ae58943c4cfea63592/app/controllers/application_controller.rb#L97) check.

mdbenjam: So I don't think we want to use functions. I was noticing that earlier in the code. I think https://github.com/department-of-veterans-affairs/caseflow/blob/2fc9a9ac7681c6475b70d1ae58943c4cfea63592/app/models/user.rb#L133 this line should be `roles.include?("VSO")`

db/seeds.rb
mdbenjam: Weird, thought we already had a role defined. Don't know how this was ever working. Maybe it wasn't...

db/seeds.rb
lowellrex: The difference being that Functions come from CSEM and roles come from VACOLS?

db/seeds.rb
lowellrex: Not sure how it was either.

db/seeds.rb
lowellrex: Ahhhh gotcha. I see what you're saying.

Issue #6357, opened 2018-07-25
Move create contentions and connect contentions to rating issues into End Product Establishment


ferlatte:
Since the success of establishing an end product is dependent on the contentions also being created and connected to rating issues, we need to wrap all of those operations into the `EndProductEstablishment` class, so we can know if we need to cancel the EP if one of the operations failed.

## Acceptance Criteria
- Data required to create contentions and connect contentions to rating issues should be saved on `end_product_establishments` before any requests are sent to VBMS.
- `established_at` field on `EstablishEndProduct` should be set and saved when all required operations are complete.

## Validation
- This story should be validated by engineers

Issue #6212, opened 2018-07-13
Editing VBA issues | Saving edited issues as contentions on existing EP


abbyraskinUSDS:
This is the simplest outcome of successfully editing EP issues - issues are added or removed under the existing EP and reflected as contentions in VBMS. (rating -> rating or non-rating -> non-rating)

## Acceptance Criteria
- The "Save" button is grayed out until a change is made, then it's the default blue button color.
- Clicking the "Save" button should trigger changing the issues.
- When the issues are saved, if issues of the same type (rated or non-rated) are added or removed to/from an existing EP, the new contentions are saved in VBMS under the same EP code and label

## Technical Notes
- To do this, we will have to remove any contentions for issues that were removed. This could be done via the `removeContention` endpoint in the VBMS claims service.

requires https://github.com/department-of-veterans-affairs/connect_vbms/issues/190

## Mocks
<img width="778" alt="edit issues" src="https://user-images.githubusercontent.com/11219417/42765672-8dd06fc4-88e6-11e8-9d44-ed7d0ef30b06.png">

## Related stories
https://github.com/department-of-veterans-affairs/appeals-design-research/issues/862

## Validation

1. Visit https://appeals.cf.uat.ds.va.gov/higher_level_reviews/600134030/edit/select_issues
1. Select/deselect issues to test the following cases
   1. No change -> button is disabled gray
   1. Deselect all issues -> hitting save displays a 'no issues selected' error message
   1. Select some issues (a different selection than original) -> hitting save redirects back to /edit when the request completes
   1. Note that while saving, the button should be disabled, though this might be very brief
1. Go to VBMS, look up test file number 123458304
1. look up claim id 600134030, the higher level review we just updated
1. Verify that contentions match the issues selected above in 2.3
1. An engineer could check for:
   1. the removed_at timestamp being set for any request_issues that were deselected
   1. review_request_id is set to nil for any request_issues that were deselected
   1. the rating_issue_associated_at is set for any newly created request_issues

leikkisa:
Wait until edit framework is done to get started on this

shanear:
@suzchap what happens if a user doesn't change anything and clicks save?

shanear:
Asked this in person, the decision is that the Save button should be disabled until a change is made.

kencheeto:
[...]

PR #6781, opened 2018-08-29
Add RequestIssuesUpdate model

shanear:
connects #6212
connects #6357

### Description
Added `RequestIssuesUpdate` class to capture an instance of a user updating the issues on the case. It is stored in the database to make the operation fully auditable, retryable, and able to be rolled back.

Also added a few more things:
- Begun moving create contentions functionality into the End Product Establishment
- Added `ClaimReview` abstract class
- Added some top level description comments for some of our classes

### DIFF
--- a/db/migrate/20180824224113_allow_nil_review_on_request_issues.rb
+++ b/db/migrate/20180824224113_allow_nil_review_on_request_issues.rb
@@ -1,0 +1,6 @@
+class AllowNilReviewOnRequestIssues < ActiveRecord::Migration[5.1]
+  def change
+    change_column_null :request_issues, :review_request_id, true
+    change_column_null :request_issues, :review_request_type, true
+  end
+end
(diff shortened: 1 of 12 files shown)


db/migrate/20180824224113_allow_nil_review_on_request_issues.rb
```diff
@@ -0,0 +1,6 @@
+class AllowNilReviewOnRequestIssues < ActiveRecord::Migration[5.1]
+  def change
+    change_column_null :request_issues, :review_request_id, true
```
kahlouie: Why are we allowing these to be nil? I don't see the reason for changing it?

kahlouie:
Looks like there's some linting errors too

db/migrate/20180824224113_allow_nil_review_on_request_issues.rb
shanear: Because when we remove an issue, instead of hard deleting it, we just set review_request to nil (to soft delete it)

PR #6786, opened 2018-08-29
Enable VSOs to view case details page, also start setting up restrictions on VSO access.

mdbenjam:
Connects #6787

### Description
This change enables VSOs to view case details pages and sets up some restrictions on what they're able to pull.

The next PR will continue down this path, ensuring that a VSO can see appeals ONLY when they are assigned as a POA to an appellant.

### Acceptance Criteria
- [ ] The VSO user can access the case details page.


lib/fakes/bgs_service.rb
```diff
@@ -18,6 +18,10 @@ class Fakes::BGSService
   # rubocop:disable Metrics/MethodLength
   # rubocop:disable Metrics/AbcSize
   def self.create_veteran_records
+    return if @veteran_records_created
```
mdbenjam: Prevent this method on running on every request, should just run on the first.

app/controllers/appeals_controller.rb
```diff
@@ -3,6 +3,8 @@ class AppealsController < ApplicationController
 
   before_action :react_routed
   before_action :set_application, only: [:document_count, :new_documents]
+  # Only whitelist endpoints VSOs should have access to.
+  before_action :deny_vso_access, except: [:index, :show_case_list, :show, :tasks]
```
lowellrex: So this will prevent VSO employees from accessing the `/document_count` and `/new_documents` endpoints, correct? And we want to do this because VSO folks won't have access to Reader so they are irrelevant?

app/controllers/appeals_controller.rb
```diff
@@ -55,6 +62,9 @@ def show
                               service: :queue,
                               name: "AppealsController.show") do
           appeal = Appeal.find_appeal_by_id_or_find_or_create_legacy_appeal_by_vacols_id(id)
+
+          return file_access_prohibited_error if !BGSService.new.can_access?(appeal.veteran_file_number)
```
lowellrex: One of the pieces of feedback we got from last week's user interviews was that the case details page loads slowly so it might be worthwhile to limit this check to only VSO employees. Something like the following perhaps?
```
return file_access_prohibited_error if current_user.vso_employee? && !BGSService.new.can_access?(appeal.veteran_file_number)
```

app/controllers/tasks_controller.rb
```diff
@@ -1,7 +1,8 @@
 class TasksController < ApplicationController
   include Errors
 
-  before_action :verify_queue_access
+  before_action :verify_queue_access, except: [:index]
+  before_action :verify_queue_access_or_vso, only: [:index]
```
lowellrex: I find this a little hard to parse. What do you think about using a combination of `verify_queue_access()` (with VSO employees still being included in the `can_access_queue?()` function) and `deny_vso_access()` (moving that to `ApplicationController`)? Perhaps something like the following is a little more explicit about the behaviour we are preventing and follows the pattern established in `AppealsController`?
```
before_action :deny_vso_access, except: [:index]
```

app/controllers/appeals_controller.rb
mdbenjam: Yeah, it's probably okay to allow it, but since they don't have reader access at the moment, I don't want to make these calls.

app/controllers/appeals_controller.rb
mdbenjam: Damn I wonder how many BGS calls we're making on the case details page. Removing this per our discussion. It should just fail naturally when it calls fetch_veteran_info.

PR #6790, opened 2018-08-29
[WIP] Starting to add check that user is part of a VSO

mdbenjam:

### Description

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -76,6 +76,14 @@ class User < ApplicationRecord
     nil
   end
 
+  def participant_id
+    bgs.get_participant_id_for_user(user)
+  end
+
+  def vsos
+    participant_ids = bgs.fetch_poas_by_participant_id(participant_id)
+  end
+
   def access_to_task?(vacols_id)
     self.class.user_repository.can_access_task?(css_id, vacols_id)
   end
@@ -201,6 +209,10 @@ class User < ApplicationRecord
   end
 
   private
+
+  def bgs
+    @bgs ||= BGSService.new
+  end
 
   def user_info
     @user_info ||= self.class.user_repository.user_info_from_vacols(css_id)
--- a/app/models/organizations/vso.rb
+++ b/app/models/organizations/vso.rb
@@ -1,17 +1,10 @@
 class Vso < Organization
   def user_has_access?(user)
-    user_participant_id = bgs.get_participant_id_for_user(user)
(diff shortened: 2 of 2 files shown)


app/models/veteran.rb
```diff
@@ -67,12 +67,12 @@ def age
     now.year - dob.year - ((now.month > dob.month || (now.month == dob.month && now.day >= dob.day)) ? 0 : 1)
   end
 
-  def self.bgs
```
mdbenjam: I'm pretty sure we don't want to have bgs initialized on a per-class-basis. I think that's what got us into trouble a long time ago, I *think* we risk using the creds for the first user to make request for all future requests.

app/models/user.rb
```diff
@@ -76,6 +76,14 @@ def vacols_full_name
     nil
   end
 
+  def participant_id
+    @participant_id ||= bgs.get_participant_id_for_user(self)
+  end
+
+  def vsos
```
lowellrex: Might it make sense to rename this method something like `vso_poas` to make it clear that what it returns is an array of Power of Attorney hashes and not Vso objects?

app/models/veteran.rb
```diff
@@ -105,6 +105,20 @@ def timely_ratings(from_date:)
     @timely_ratings ||= Rating.fetch_timely(participant_id: participant_id, from_date: from_date)
   end
 
+  def appeals_vso_has_access_to(vso_participant_ids)
```
lowellrex: I think this method is a little more generic than the name suggests. Perhaps we could rename it something like `accessible_appeals_for(participant_ids)`? Also, are there any downsides other than losing the batching function that would happen if we move this to the Appeal model?

app/models/user.rb
mdbenjam: Though it's the vsos the user belongs to, not which ones they have representing them. Maybe the name `vsos_user_represents` is better?

app/models/veteran.rb
mdbenjam: It's just the batching functionality. I'm planning on building out a similar thing for the appeal level to make POA work on a claimant rather than a veteran basis. We can remove VSO, but should include POA in the name since I think that's the only users it applies to. `accessible_appeals_for_poa(poa_participant_ids)`

PR #6792, opened 2018-08-29
Ensure GenericTasks are assigned by human beings

### LINKED ISSUE #6782: Ensure GenericTasks are assigned by human beings
When we create a child task for a GenericTask, we use the [parent's assignee to set the child's assigner](https://github.com/department-of-veterans-affairs/caseflow/blob/8e9f6bba985c0f927a69202286b35e955502fe0a/app/models/tasks/generic_task.rb#L22). This is mostly correct except in the case where the parent task is assigned to an organization. In that case, we should set the child task's assigner to be the person who initiated the action in the frontend.

Worth noting that we will likely eventually create child tasks programmatically, so we will need some sensible fallback if there is no current user.

## Acceptance criteria
* Set current user as GenericTask assigner when creating new tasks

lowellrex:
Resolves #6782.

Prior to this change we would use the parent task's assignee as the child task's assigner. This allowed for the situation where a task assigned to an organization could create a child task assigned by an organization. Currently, the `Task` model [expects all assigners to be humans](https://github.com/department-of-veterans-affairs/caseflow/blob/63c528e93c50e24665095ff83fc04e0f75901eca/app/models/task.rb#L5).

This PR makes is so GenericTasks can only be assigned by a User (or nil).


lowellrex:
I think leaving assigned_by as a User is fine for now and maybe even provides some value. For example, when somebody assigns an organization task to another User the assigned_by field lets us know who did that. If the we instead used the parent task to determine who it was assigned by then we wouldn't know which human being create the child task.

Though maybe this just uncovers another deficiency in the Task model as currently implemented where we find ourselves attaching a fair amount of information around actions taken on a task (who assigned it, notes for the assignee) to the task itself. Perhaps we should move that type of information to a different table (`task_actions`?).

mdbenjam:
My counter would be, we're duplicating information right now. assigned_by duplicates information in the task tree. We can answer both the question who assigned this task `parent.assigned_to` and which user (not organization) assigned this task by recursing up the tree till we find a user assigned_to. I feel this implementation is a little brittle since there's no guarantee that the parent's parent is a user (organizations could in theory route directly to another organization).

Agreed that a separate task action table might be necessary, for notes etc. We'll see how it develops.

lowellrex:
How would we tell who assigned a task if the person who assigned it is a member of the same organization as the person they assigned the task to (maybe a team has two people who are responsible for assigning work)? The parent task would be assigned to their organization and the child task would be assigned to somebody else in the organization.

mdbenjam:
Ah you're totally right that assigned_by can provide new information. i.e. a user can assign a task from the organization page to another user. What do you think about recursing up the tree to find a user parent instead of just looking at one parent to see if they're a user, and otherwise assigning nil. Maybe that's overkill since we don't know of a scenario where an organization will assign to an organization. I'll approve it and leave it up to you.

PR #6805, opened 2018-08-31
Moving Circle CI to 2.5.1

enriquemanuel:
Resolves Circle CI 2.5.1 not working with caseflow

### Description
A few changes were made here.
1. No longer using an ECR image with Oracle installed but using the CircleCI image and installing Oracle library dependencies on the run. This change *should* not impact performance or timing since the libraries are in the repository at checkout.
2. The above change gives us the flexibility to test newer / older versions or Ruby on the fly, we can essentially try a Ruby 2.6 or a 2.2 on whatever run, just by changing the tag on the base image
3. Made Parallelism 5, to be standard
4. Improved syntax readibility

### Acceptance Criteria
- Code compiles with any and all executions

#### Tagging people for visibility:
@mdbenjam @nemodjuric @aroltsch @boris-ning-usds


enriquemanuel:
Dropping this PR due to libraries

Issue #6703, opened 2018-08-21
Add task actions dropdown for organizational queues


lowellrex:
In order to build [generic organizational task queues](https://github.com/department-of-veterans-affairs/caseflow/issues/6573), we need to support a small set of generic task actions for teams and individuals using organizational task queues. On the front-end we need to add a component that is similar to the [task action dropdowns for the existing queues](https://github.com/department-of-veterans-affairs/caseflow/blob/359be6c6867de94b4676b68de19d62f184cde917/client/app/queue/CaseSnapshot.jsx#L158).

## Acceptance criteria
* Task actions component appears on organization task queues (except VSO queues)
* Following task actions exist in the dropdown (adding back-end functionality for these actions will be the work of a different ticket)
  * Mark task complete
  * Assign to team
  * Assign to individual

PR #6806, opened 2018-08-31
Move AppealsController#tasks to TasksController#for_appeal

lowellrex:
Consolidates task-related actions in the TasksController. Refactor that seemed timely given the amount of related work to come from #6703.


app/controllers/application_controller.rb
```diff
@@ -100,6 +100,15 @@ def certification_header(title)
   end
   helper_method :certification_header
 
+  # https://stackoverflow.com/a/748646
+  def no_cache
```
mdbenjam: I do not think we should make every endpoint not cache. I would say most endpoints should.

mdbenjam: Maybe this is a reasonable alternative: https://stackoverflow.com/questions/11966766/hitting-browser-back-button-shows-json-rather-than-html-using-rails-d3-js

mdbenjam: Oh JK for some reason I thought this was a before action applied to everything 😅 ignore what I said.

lowellrex: Definitely still in favor of looking into a different approach for the no-cache stuff... though maybe that'll be a later PR

Issue #6214, opened 2018-07-13
Editing VBA issues | Save edits scenario where we need to cancel irrelevant EPs


abbyraskinUSDS:
While fixing the issues originally created during intake for the claim, if the issues change from rated to non-rated or vice-versa, the EP label would need to change too. This story captures the "Save" scenario where we need to cancel the existing EP and create a new one with the accurate claim label. (i.e., rating -> non-rating)

## Acceptance Criteria
- If the edited EP issues remove an entire category (rating or non-rating) of issues from the review, then Caseflow should cancel that EP

## Example
A user edits a Supplemental claim with rating issue, X, and non-rating issue, Y. They change issue Y to a rating issue, leaving no more non-rating issues as part of the supplemental claim. Caseflow should cancel the non-rating supplemental claim EP associated with that supplemental claim.

## Technical notes
You can cancel an EP by using the `claims.cancel_end_product` method in Ruby BGS (see: https://github.com/department-of-veterans-affairs/ruby-bgs/blob/master/lib/bgs/services/benefit.rb#L37)

requires #6213 
requires #7269

shanear:
This is failing. I think due to this error: https://sentry.uat.ds.va.gov/sentry/caseflow-uat/issues/1061/

shanear:
new error 🙁 

https://sentry.uat.ds.va.gov/sentry/caseflow-uat/issues/1087/

shanear:
woohoo, it worked 🔥 🔥 
<img width="772" alt="screen shot 2018-11-06 at 4 51 37 pm" src="https://user-images.githubusercontent.com/1596259/48096074-49197b80-e1e4-11e8-88eb-70eccb32de57.png">

Issue #6213, opened 2018-07-13
Editing VBA issue | Saving edited issues triggers new EP establishment


abbyraskinUSDS:
While fixing the issues originally created during intake for the claim, if the issues change from all rated to a mixture of non-rated and rated (or vice-versa), there needs to be a new EP to track both types of claim issues. This story captures the "Save" scenario where we need to create a new EP. (i.e. rating -> rating + non-rating)

## Acceptance Criteria
- If the edited EP issues are edited to include a category of issue (Rating or Non-rating) that did not previously have any issues on the EP, then Caseflow should create the appropriate new EP and attach the new issues to that EP.
- If the EP created above was part of a Higher Level Review where an informal conference was requested, Caseflow should also generate an Informal Conference tracked item (see: https://github.com/department-of-veterans-affairs/caseflow/issues/7118 for more information)

## Example
If an HLR with 1 non-rating issue, X, is edited to have both non-rating issue, X, and a new rating issue, Y, then we would create an additional rating EP and attach issue Y to it.

requires #6148 
requires #7269

jiahuang:
Looked through some code and I believe that this is already being done.

Upon a higher level review or supplemental claim update, we call `request_issues_update.perform!` which in turn calls `review.create_issues!(new_issues)` which creates new eps if needed for each issue.

Will add tests to confirm functionality.

leikkisa:
Yeah, during sprint planning we suspected that but thought it'd be good for someone to confirm!  Thanks for looking at it and adding tests!

shanear:
<img width="811" alt="screen shot 2018-11-01 at 6 56 15 pm" src="https://user-images.githubusercontent.com/1596259/47884724-472c7280-de08-11e8-937d-1d8feaea4fc8.png">

<img width="797" alt="screen shot 2018-11-01 at 6 59 58 pm" src="https://user-images.githubusercontent.com/1596259/47884754-6aefb880-de08-11e8-8949-dea26576d744.png">

Nice! This is working 👍

PR #6807, opened 2018-09-02
Refactor how end product establishments are handled on claim reviews

### LINKED ISSUE #6187: Refactor end products out of AMA review model
End products are only found in supplemental claims and higher level. With the work done for #6093, we can have a separate module that inherits from AMAReview, called AMAReviewWithEP. SupplementalClaim and HigherLevelReview can then inherit from that, removing unnecessary EP logic from AMAAppeals.

shanear:
We used to establish end products and create contentions in a way that was not easily usable for adding and removing request issues after the original intake. It was not possible to determine the EP Establishment that a request issue's contention was created on. It also didn't easily support adding more than 2 EPs on the same review.

To fix this, the refactor determines or creates the correct EP establishment for each issue when they are created in the `create_issues!` method. No VBMS calls are made in this method, all the data is required for making the appropriate VBMS calls is saved though.

Then in `process_end_product_establishments!` it creates any contentions or EPs that have not been created. It's idempotent, meaning it can be called again to add any contentions or EPs that need to be added after the original intake.

Also, moved all of the appropriate logic into `ClaimReview`.

This PR became unfortunately large, which I apologize for but I'm not sure it's worth trying to break it down since a lot of the logic is intertwined.

connects #6212
connects #6213 
connects #6214 

resolves #6187


app/models/ama_review.rb
```diff
@@ -43,42 +43,12 @@ def payee_code
     claimants.first.payee_code
   end
 
-  def create_issues!(request_issues_data:)
```
shanear: Moved this logic to `ClaimReview` since most of the logic applies only to reviews with end product establishments

app/models/claim_review.rb
```diff
@@ -2,20 +2,81 @@
 # higher level review as defined in the Appeals Modernization Act of 2017
 
 class ClaimReview < AmaReview
+  has_many :end_product_establishments, as: :source
+
   self.abstract_class = true
 
+  # Save issues and assign it the appropriate end product establishment.
```
shanear: `create_issues!` and `process_end_product_establishments!` are the two key methods of this PR. It allows us to reuse the same logic for intaking claim reviews that we use to update the issues.

app/models/claim_review.rb
```diff
@@ -2,20 +2,81 @@
...
+  # Create that end product establishment if it doesn't exist.
+  def create_issues!(new_issues)
+    new_issues.each do |issue|
+      issue.update!(end_product_establishment: end_product_establishment_for_issue(issue))
+    end
+  end
+
+  # Send the appropriate calls to VBMS to create the end products and contentions for any
+  # outstanding end product establishment. If all end products have been created then this
+  # method does nothing
```
shanear: This method is mostly idempotent, the only problem that hasn't been solved is the if the rating issue assignment fails. It will not attempt to retry that.

app/models/claim_review.rb
```diff
@@ -2,20 +2,81 @@
...
+  def end_product_establishment_for_issue(issue)
+    ep_code = issue_code(issue.rated?)
+    end_product_establishments.find_by(code: ep_code) || new_end_product_establishment(ep_code)
+  end
+
+  def issue_code(_rated)
+    fail Caseflow::Error::MustImplementInSubclass
+  end
+
+  def create_contentions_for_end_product_establishment(end_product_establishment)
```
shanear: Most of this was just moved from `AmaReview` and updated slightly for the new way of dealing with end product establishments.

app/models/ama_review.rb
```diff
@@ -43,42 +43,12 @@ def payee_code
...
-    request_issues.destroy_all unless request_issues.empty?
-
-    request_issues_data.map { |data| request_issues.create_from_intake_data!(data) }
-  end
-
-  def create_end_products_and_contentions!
```
shanear: Instead of going through each potential end product that can be created, we now create these when the issues are created

app/models/supplemental_claim.rb
```diff
@@ -11,52 +11,45 @@ def ui_hash
       veteranName: veteran.name.formatted(:readable_short),
       veteranFileNumber: veteran_file_number,
       claimId: end_product_claim_id,
-      receiptDate: receipt_date,
+      receiptDate: receipt_date && receipt_date.to_formatted_s(:json_date),
```
shanear: This fixes the stupid date issue we've been having.

app/models/claim_review.rb
```diff
@@ -2,20 +2,81 @@
...
+  def create_associated_rated_issues!(end_product_establishment, issues)
+    request_issues_to_associate = issues.select(&:rated?)
+
+    return if end_product_establishment.code != issue_code(true)
+    return if request_issues_to_associate.empty?
+
+    VBMSService.associate_rated_issues!(
+      claim_id: end_product_establishment.reference_id,
+      rated_issue_contention_map: rated_issue_contention_map(request_issues_to_associate)
+    )
```
ferlatte: We should add request_issue.rating_issue_associated_at here that we set when the VBMSService succeeds; then we know that this call worked. Otherwise we can't replay failed requests later.

app/models/end_product_establishment.rb
```diff
@@ -61,29 +66,36 @@ def perform!
...
+    # the created contentions. Instead find the issue by matching text.
+    create_contentions_in_vbms(for_objects.pluck(:description)).each do |contention|
+      matching_object = for_objects.find { |object| object.description == contention.text }
+      matching_object && matching_object.update!(contention_reference_id: contention.id)
+    end
+
+    fail ContentionCreationFailed if for_objects.any? { |object| !object.contention_reference_id }
+  end
+
+  def remove_contention!(for_object)
```
ferlatte: We should add and save a contention_removed_at field for anything that has contentions so that we can properly log that VBMSService removed them (and so we can deal with cases where the removal call fails).

PR #6815, opened 2018-09-04
Implement IDT AMA details - version 1

aroltsch:
First PR to return IDT details for AMA appeals. 

TODO:
1. Get information about assignor and documents for AMA appeals
2. Once we make a decision on how to handle issues, implement issues for AMA appeals
3. Lots of TODO comments


app/controllers/idt/api/v1/appeals_controller.rb
```diff
@@ -13,10 +13,13 @@ def details
     # We query the case assignment table here so we can get information for
     # who wrote the case decision docs/OMO request and what their doc ids are.
     # For AMA appeals, we should get that information from our attorney and judge case review tables.
-    tasks = QueueRepository.tasks_for_appeal(params[:appeal_id])
-    appeals = QueueRepository.appeals_by_vacols_ids([params[:appeal_id]])
```
aroltsch: I don't think this is necessary of we have the logic on line 55.

app/controllers/idt/api/v1/appeals_controller.rb
```diff
@@ -13,10 +13,13 @@ def details
...
-    return render json: { message: "Appeal not found" }, status: 404 if appeals.empty?
-    render json: json_appeal_details(tasks, appeals[0])
+    tasks = if appeal.is_a?(LegacyAppeal)
+              QueueRepository.tasks_for_appeal(appeal.vacols_id)
+            else
+              []
```
aroltsch: This is not implemented yet.

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -65,10 +65,6 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
     object.docket_number
   end
 
-  attribute :status do
```
aroltsch: This method lives on Appeal now.

app/models/proper_name.rb
```diff
@@ -0,0 +1,5 @@
+class ProperName < String
```
amprokop: Hmmm is this class necessary? My hope is to avoid name splitting for its inaccurate edge cases e.g. Martin Van Buren

app/models/serializers/idt/v1/appeal_details_serializer.rb
```diff
@@ -9,50 +9,52 @@ def id
...
+  attribute :aod do
+    object.advanced_on_docket
+  end
   attribute :cavc
   attribute :status
   attribute :previously_selected_for_quality_review
 
   attribute :outstanding_mail do
-    object.outstanding_vacols_mail?
+    object.is_a?(LegacyAppeal) ? object.outstanding_vacols_mail? : "not implemented"
```
amprokop: should we return `false` instead of "not implemented" since this is an API response that presumably will have a conditional that turns on a truthy value

app/models/serializers/work_queue/appeal_serializer.rb
amprokop: word

app/models/proper_name.rb
aroltsch: Yeah. But in your IDT API, you are returning first and last names separately. I've also checked that BGS returns a full name for claimants. https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/claimant.rb#L6

app/models/serializers/idt/v1/appeal_details_serializer.rb
aroltsch: good point.

app/models/proper_name.rb
aroltsch: I will implement a BGS call to get first and last names. https://github.com/department-of-veterans-affairs/ruby-bgs/blob/master/lib/bgs/services/person.rb#L30

app/models/serializers/idt/v1/appeal_details_serializer.rb
```diff
@@ -9,50 +9,52 @@ def id
...
-      ).as_json[:data][:attributes]
+    if object.is_a?(LegacyAppeal)
+      object.issues.map do |issue|
+        ActiveModelSerializers::SerializableResource.new(
+          issue,
+          serializer: ::WorkQueue::IssueSerializer
+        ).as_json[:data][:attributes]
+      end
+    else
+      object.request_issues
```
amprokop: @aroltsch, can we return { program: "Compensation" } for every request issue for now?

aroltsch: Done. 
image

app/models/serializers/idt/v1/appeal_details_serializer.rb
```diff
@@ -9,50 +9,56 @@ def id
...
+        ActiveModelSerializers::SerializableResource.new(
+          issue,
+          serializer: ::WorkQueue::IssueSerializer
+        ).as_json[:data][:attributes]
+      end
+    else
+      object.request_issues.map do |issue|
+        # Hard code program for October 1st Pilot, we don't have all the info for how we'll
+        # break down request issues yet but all RAMP appeals will be 'compensation'
+        { id: issue.id, disposition: issue.disposition, program: "Compensation" }
```
amprokop: ah oops @aroltsch — can we add in the issue `description` as well

PR #6816, opened 2018-09-04
S3 configs

### LINKED ISSUE #6813: Spreadsheet Integrations | New S3 bucket
The bucket we're storing spreadsheets in right now has a configuration that deletes files after 5 days. We need to create a new bucket for hearing schedule that doesn't delete files.

sharonwarner:
Resolves #6813 

### Description

This PR stores files in two new S3 buckets for form_8s and hearing_schedule documents so we can use different permissions for different documents.

### Testing Plan

In UAT, 
1) Upload a spreadsheet in hearing schedule and confirm it's stored in S3 in a 'folder' called hearing_schedule
1) Certify a case through certification and confirm the form 8 is stored in a 'folder' called form_8


spec/factories/judge_schedule_period.rb
```diff
@@ -62,19 +62,19 @@
     end
 
     before(:create) do
-      S3Service.store_file("validJudgeSpreadsheet.xlsx",
+      S3Service.store_file("hearing_schedule/validJudgeSpreadsheet.xlsx",
                            "spec/support/validJudgeSpreadsheet.xlsx", :filepath)
```
oscarAtNava: Why are we not using the constant S3_SUB_BUCKET in this class?

spec/factories/ro_schedule_period.rb
```diff
@@ -1,7 +1,7 @@
 FactoryBot.define do
   before(:create) do
-    S3Service.store_file("validRoSpreadsheet.xlsx", "spec/support/validRoSpreadsheet.xlsx", :filepath)
-    S3Service.store_file("blankRoSpreadsheet.xlsx", "spec/support/blankRoSpreadsheet.xlsx", :filepath)
+    S3Service.store_file("hearing_schedule/validRoSpreadsheet.xlsx", "spec/support/validRoSpreadsheet.xlsx", :filepath)
```
oscarAtNava: Ditto here, S3_SUB_BUCKET constant?

Issue #6623, opened 2018-08-13
Assign task to VSO during intake of RAMP appeals when Veteran is represented by VSO


lowellrex:
In order for VSOs to see RAMP appeals for Veterans they represent in caseflow, we need to create tasks associated with the appeal and assigned to the VSO that represents the Veteran. We should do this during the intake step for all new appeals (RAMP and later AMA), as well as backfilling tasks for BEAAM cases.

These tasks are only placeholders and do not imply that the VSOs are required to take any action on the appeal. This is simply a way to provide VSOs with a view of the RAMP appeals they represent that leverages existing frameworks for convenience's sake. A later ticket will be written to create an IHP task assigned to a VSO when we intake a direct appeal case which will represent a real task with some required action to be taken by the VSO.

## Acceptance criteria
* Create tasks assigned to VSOs during intake for all appeals where Veteran is represented by VSO
* Create tasks for appeals already in caseflow where VSO represents Veteran

laurjpeterson:
We've completed this! Closing.

Issue #5958, opened 2018-06-19
Re-intake RAMP appeals so they are AMA appeals


laurjpeterson:
## Goals
- Allow users to search and find all AMA appeals (including RAMP appeals)

## AC
- Move RAMP appeals into the AMA appeal table, re-intake them so they can go through the intake flow for AMA appeals
- Add RAMP to the Type (e.g. AOD, CAVC)

## Notes 6/19
- RAMP data issues:
   - RAMP appeals are not mapped to the Veteran's disability profile.
   - Don't track any issue dispositions for RAMP EPs in Caseflow.
      - No way to know if an issue on an HLR EP has been granted or denied. VBA just closes the EP.
   - Caseflow creates contentions + EP --> VBA user goes to a rating portion --> creating a new rating issue
      - Caseflow is suggesting that for these contentions, they need to add a disposition
- Shane's idea: Translate RAMP appeal data to AMA data. If a RAMP election was completed, this will result in a rating. That rating will turn into AMA (so no RAMP appeals would go straight to AMA appeals)
- Shane's idea: Intake RAMP appeals

### Definitions
- Contention - Veteran thinks this is what they deserve
- Rating issue - What the Veteran ends up getting

lowellrex:
When we re-intake these RAMP appeals we will need to create the root task of the task tree (how we will track work that needs to be done on an appeal post-VACOLS). We added a method to create the root task in PR #6689, so all that needs to be done is one call with an `Appeal` object that looks like `Task.create_root_task(appeal)`.

Let me know if there are any questions!

laurjpeterson:
Re-intaking is in process! going to close this ticket.

Issue #6573, opened 2018-08-09
Build out generic queues for admin teams


mdbenjam:
# Overview
In discussions over the past few weeks we've determined that at our current rate of progress we're not going to be able to rollout Queue to all admin teams in the timeframe required. However, by thinking of queues in the most generic form, we think we can get an MVP into the hands of users, giving us more time to come back and make Queue work for users' specific needs.

**What is a generic queue?**
Generic functionality that is made possible by our task tree data model that tracks responsibility for tasks, so that we can roll queues out to multiple teams at the board quickly and simultaneously

Resource: [presentation](https://docs.google.com/presentation/d/1__UukJy9ts2qbvHVhbLd3lnInaiZ3UqAqAYdLmC7vDM/edit#slide=id.g3fe0b62e2f_0_222) to BVA leadership about Caseflow tasks 10/17/2018

**What functionality is needed to have a generic queue?**
- Task tree data model
- Viewing tasks in an organization's queue (e.g. quality review)
   - Includes tabs available: New, On Hold, Completed
- Viewing tasks in a user's queue (e.g. quality review user)
   - Includes tabs available: New, On Hold, Completed
- Creating a task on the root task
- Creating child tasks
- Task actions
   - Assign to person
   - Assign to team
   - Mark complete
   - (Place on hold?)
   - (Cancel?)
- Adding notes/instructions to tasks
- Mapping people to organizations

**What functionality is made possible by a generic queue + task tree model?**
- Case timeline - https://github.com/department-of-veterans-affairs/caseflow/issues/5592

# Initiatives
Generic queues is a big project with quite a few initiatives tracked in separate epics/issues. We've assigned point people for each:
- VSOs - https://github.com/department-of-veterans-affairs/caseflow/issues/5581 - @mdbenjam 
- Creating generic task on root (for mail, lit support) - https://github.com/department-of-veterans-affairs/caseflow/issues/7404 - @lowellrex 
- Case timeline - https://github.com/department-of-veterans-affairs/caseflow/issues/5592 - @lowellrex 
- Organization management - https://github.com/department-of-veterans-affairs/caseflow/issues/7427, https://github.com/department-of-veterans-affairs/caseflow/issues/7455 - @mdbenjam 
- Refactoring co-located queue (eventually) to generic queue - TBD

# Rollout plan
1. VSOs
2. Quality review, AOD, FOIA - #7355 begins this work
3. Mail, Lit support (requires creating task on root task), Translation
[...]

PR #6818, opened 2018-09-04
Building tasks from intake

mdbenjam:
Connects #5958 and #6623

### Description
1) Call our new `create_initial_tasks!` method on appeal from the appeal intake flow.
1) Create a new ClaimantPowerOfAttorney service to start shifting the determination of POA from veteran to claimant where it belongs.
1) New code to create VSO tasks based on the claimant's VSOs.

### Testing Plan
1. Intake any appeal, and select the claimant Bob Vance
1. Check the resultant appeal and it should have a VSO task created for it


app/models/appeal_intake.rb
```diff
@@ -38,6 +38,7 @@ def complete!(request_params)
 
     detail.create_issues!(request_issues_data: request_params[:request_issues] || [])
     detail.update!(established_at: Time.zone.now)
+    detail.create_initial_tasks!
```
mdbenjam: @shanear is this the right place to put an integration with Queue. We need this to create an appeals' tasks.

app/services/claimant_power_of_attorney.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  attr_accessor :claimant_participant_id
+
+  private
+
+  # We need to override the BgsPowerOfAttorney's fetch_bgs_record to use participant id.
+  # All of caseflow should probably move to using this since POA is determined by claimant
+  # not by Veteran. That said, until we can make that change this is our placeholder.
+  def fetch_bgs_record
+    bgs.fetch_poas_by_participant_ids([claimant_participant_id])[claimant_participant_id]
```
mdbenjam: @amprokop I was hesitant to make this change across the code base in BgsPowerOfAttorney. Instead I made this new subclass so I could change this method do you think this is a good idea?

app/models/appeal.rb
```diff
@@ -81,13 +81,28 @@ def docket_number
...
+  # For now power_of_attorney returns the first claimant's power of attorney
   def power_of_attorney
-    @bgs_poa ||= BgsPowerOfAttorney.new(file_number: veteran_file_number)
+    claimants.first.power_of_attorney
   end
-
   delegate :representative_name, :representative_type, :representative_address, to: :power_of_attorney
 
+  def power_of_attorneys
+    claimants.map(&:power_of_attorney)
```
shanear: powers of attorney, power of attorneys, or power of attornies? lol

app/models/tasks/root_task.rb
```diff
@@ -1,5 +1,5 @@
 class RootTask < Task
```
shanear: It'd be nice to have a description of what RootTask means in the comments, cause i don't know.

app/models/appeal_intake.rb
shanear: Yeah i like the idea here. I think it might be more descriptive to name it `create_tasks_on_intake_success!` or something like that.

app/models/tasks/vso_task.rb
```diff
@@ -1,2 +1,11 @@
 class VsoTask < Task
+  class << self
+    def create_tasks_for_appeal!(appeal)
+      vso_participant_ids = appeal.power_of_attorneys.map(&:participant_id)
```
shanear: Might be helpful to make `appeal.vsos` to capture a lot of this logic. More reusable, and makes this method simpler:

```
appeal.vsos.each do |vso|
  create(appeal: appeal, status: "in_progress", assigned_to: vso)
end
```

app/models/tasks/root_task.rb
lowellrex: `RootTask` will be the... root of the task tree that will replace location codes as the way to track responsibility for action on an appeal at the Board in Caseflow. Here is the tech spec detailing how that looks: https://github.com/department-of-veterans-affairs/caseflow/issues/6573

app/models/appeal.rb
amprokop: Hmm or should we just dodge this awkwardness and call it `representatives`?

app/services/claimant_power_of_attorney.rb
amprokop: This looks fine — we could also do something like

use the poa_by_file_number service if file_number is set

else use the participant_id service

amprokop: Also, looking forward, it'd be nice to make it easy to fetch multiple POAs in one call — either for all claimants on an appeal or for all claimants on a list of appeals. Any thoughts on whether there's a step that makes sense to push us in that direction?

app/models/organizations/bva.rb
```diff
@@ -1,2 +1,5 @@
 class Bva < Organization
+  def self.singleton
+    Bva.first || Bva.create(name: "Board of Veterans' Appeals")
```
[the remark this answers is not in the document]
lowellrex: Sure. Makes sense to me!

app/models/tasks/vso_task.rb
```diff
@@ -1,2 +1,11 @@
 class VsoTask < Task
+  class << self
+    def create_tasks_for_appeal!(appeal)
```
lowellrex: I like this! Real smart solution to the situation where a claimant has multiple POAs.

Can we add the task's parent task (the root task) here so we can attach this task to the task tree? We could still keep a single argument (root_task) and create the appeal from `root_task.appeal`.

app/models/appeal.rb
mdbenjam: I lean against representative. Other systems seem to call it poa or power of attorney (specificially BGS) so I'd like to be in line with that. Hopefully we can eventually get rid of `power_of_attorney` since picking the first one is wrong haha. We just currently need it for the UI.

mdbenjam: Maybe we could rename `power_of_attorney` to `first_claimants_power_of_attorney` to be explicit.

app/services/claimant_power_of_attorney.rb
mdbenjam: Using either or depending on whether participant_id or file_number is set seems reasonable. I might change that.

In terms of querying for multiple POAs at once, we have that code already here: https://github.com/department-of-veterans-affairs/caseflow/blob/c6226f53e4f954c20a59319196a571c19f1e0ca7/app/models/veteran.rb#L108

That said, I don't think it's necessary for us to use that same thing here, I think it just makes it more complicated, and in the vast majority of cases there is only one claimant. Something we can consider in the future if the case details page is too slow.

amprokop: Makes sense, thanks!

app/models/tasks/root_task.rb
mdbenjam: The real reason we're calling it root task is because ramps are a root vegetable.

shanear: I think I understand, but it's hard to tell from the tech spec. Root task means just a generic task assigned to an org?

lowellrex: There will be one `RootTask` for each appeal that all of the other tasks for that appeal will be connected back to (for AMA appeals, and eventually for legacy appeals as well).

The RootTask will be assigned to the `Bva` organization subclass, but this is simply for convenience's sake because we expect tasks to be assigned to somebody (organization or person). We do not expect actions to be taken directly on the `RootTask` itself (or by somebody representing this `Bva` user to which the `RootTask` is assigned).

PR #6828, opened 2018-09-05
Allow VSO users access to switch user screen

lowellrex:
### Test plan:
* Log in as VSO User
* Go to your queue (notice the redirect! How cool!)
* Select "Switch User" from the dropdown in the top right
* It works. Marvel at the majesty of web development in the 21st century


app/controllers/test/users_controller.rb
```diff
@@ -6,6 +6,7 @@
 class Test::UsersController < ApplicationController
   before_action :require_demo, only: [:set_user, :set_end_products, :reseed, :toggle_feature]
   before_action :require_global_admin, only: :log_in_as_user
+  skip_before_action :deny_vso_access, only: :index
```
aroltsch: `set_user` should also be skipped.

Issue #4932, opened 2018-03-23
USWDS upgrade


mdbenjam:
Time-boxed exploration of trying to upgrade to the latest USWDS. Currently we use an outdated USWDS gem that hasn't been updated in 2 years. We should see if we can import the NPM package directly, instead of through the gem so that we can receive regular updates. Let's keep this exploration to two days and see what happens. After that time, when we understand the difficulties of migrating, we can make a decision on which way to go.

nolawi:
# USWDS update and Use NPM
Broken the level of effort needed to upgrade USWDS into four parts. (Difficulty 1-5)

### 1. USWDS package via NPM in FE (3)

- Attempted to import and run the package in the client side using webpack and add styles to load on all pages. Breaks UI in some pages, and some style guide components  like table, accordion, and navigation header. 

- We will also need to explore adding some kind of style linter.

### 2. Remove Outdated Gem in Caseflow (4)

There was two issues here I found. 

- Caseflow commons has outdated USWDS gem as a [dependency also](https://github.com/department-of-veterans-affairs/caseflow-commons/blob/master/Gemfile). 
and it needs the USWDS gem also installed for it to run in caseflow. I didn’t explore this much. 
- There are non style issues in some global components. Javascript errors rendering some components, including ones imported from caseflow-frontend toolkit

### 3. Remove Gem as a Dependency from caseflow-commons(2)
There are style mixing and variables that caseflow common needs to work independently, this would essentially mean removing imported [USWD style overrides](https://github.com/department-of-veterans-affairs/caseflow-commons/blob/master/app/assets/stylesheets/caseflow_main.scss).


**_Consider deprecating Casflow-commons_** 
Most of styles can just be imported directly in the sass file or just adding it local to our project and eventually removing most of these styles because we are moving in the direction of inline javascript styles. 

**4. Styleguide (4)**
All of styleguide does not render at all without USWDS. We should consider react based library like [Storybook](https://wix.github.io/wix-style-react/ ) or [pattern lab](http://demo.patternlab.io/) - easy to manage as we move to inline Javascript styled component.

## Proposal
....

mdbenjam:
Some questions:

> and it needs the USWDS gem also installed for it to run in caseflow. I didn’t explore this much.
When you say it needs the USWDS gem installed, what does that mean. Aren't we just using this as a conduit for including USWDS? If we install it via NPM don't we get the same thing?

[...]

PR #6832, opened 2018-09-05
Move pdf_viewer sym

nolawi:
Related to #4932

### Description
imports pdf_viewer css via css loader

### Acceptance Criteria
- [ ] Code compiles correctly


client/webpack.config.js
```diff
@@ -63,6 +63,19 @@ const config = {
...
+      {
+        test: /\.css?$/,
+        use: [{
+          loader: 'style-loader'
+        },
+        {
+          loader: 'css-loader',
+          options: {
+            sourceMap: true,
+            url: false
```
nolawi: css-loader, can turn off the `url` import but it has to be loaded outside scss file because we need that for `scss` files.

PR #6836, opened 2018-09-05
Determine AOD by grants or age

### LINKED ISSUE #6274: Automatically mark cases AOD due to age
Veterans who are 75 years or older are automatically Advanced on the Docket (AOD). 

Caseflow should:
- mark cases cases AOD due to age when they are established
- mark cases AOD if they turn 75 at any point in the appeals process 

For reference, today, Jed runs a report each week to identify Veterans who are 75 years or older by their birthdate and adds an AOD diary to their case. 

## AC
- Verify that Veterans who are 75 years or older at the time of case establishment at the Board
- Verify that a Veteran who turns 75 years old at any point in the appeals process is marked AOD

mdbenjam:
Resolves #6274

### Description
We need to change how we determine AOD. Currently we have a column on the Appeal table that marks the case as AOD or not. We've determined this is a bad structure for a few reasons:
1) AOD is attached to a claimant not an appeal
1) AOD needs to be automatically marked when a claimant is over 75

In order to achieve this we're removing the AOD field on the appeal table and creating a new advance_on_docket_grants table which will have a one-to-many relationship with claimants. This new table will contain the reason a case is being marked AOD.

In addition we're adding a date_of_birth field to the claimants table so we can determine when a case should be marked AOD due to age. There's a chance we'll need to build a UI in Intake for this in case the DOB field on the NOD is the claimants DOB. More to come there.

Note I know the DB migration is not backwards compatible with the application code around BEAAM appeals. But I think it's okay since it should only impact BEAAM appeals which only have a very small number of users. There will be a 5-10 minute span where anyone who requests BEAAM appeals might get an error. I can split this up if we deem that risk to high.

### Testing Plan
1. The automated tests exercise the feature.


app/models/claimant.rb
```diff
@@ -2,25 +2,43 @@ class Claimant < ApplicationRecord
...
 
   bgs_attr_accessor :name, :relationship,
                     :address_line_1, :address_line_2, :city, :country, :state, :zip
 
   def self.create_from_intake_data!(participant_id:, payee_code:)
     create!(
       participant_id: participant_id,
-      payee_code: payee_code
+      payee_code: payee_code,
+      date_of_birth: BGSService.new.fetch_person_info(participant_id)[:brthdy_dt]
```
mdbenjam: @shanear since this will impact Intake.

shanear: Shouldn't we be using the whole `AssociatedBgsRecord` stuff to get this value?

Otherwise looks good :100:

mdbenjam: I don't want to add this extra call to the `fetch_bgs_record` since that will be called frequently and having 3 calls in there will slow down a lot of page loads. We only need this called the first time we create a claimant record.

shanear: `fetch_bgs_record` should cache itself. That's the point of `AssociatedBgsRecord`, to prevent duplicative calls.

mdbenjam: Sorry meant to say `fetch_person_info`. `fetch_bgs_record` gets called every time we want claimant information, which will be on lots of page loads. Going from 2->3 BGS calls seems not worth it, when we only need to get this value one time.

PR #6839, opened 2018-09-05
Add generic mark task complete action

lowellrex:
Connects #6703. (Still need to add the other 2 task actions to complete that ticket, as well as doing the cleanup that will result from that).

Adds "Mark task complete" task action for generic tasks assigned to organizations, and allows folks assigned to that organization to take action on those tasks.

This PR can *almost* be cleanly divided into 3 pieces:
1. Backend access control changes to allow for organizational users (recommend starting by looking at the [ApplicationController](https://github.com/department-of-veterans-affairs/caseflow/pull/6839/files#diff-55c5b7aecfb519d0e4880eaf2788eb6e) stuff then working your way through the [TasksController](https://github.com/department-of-veterans-affairs/caseflow/pull/6839/files#diff-f4416012d567d02da854cfbc9e2df2ca) and supporting changes to the [GenericTask](https://github.com/department-of-veterans-affairs/caseflow/pull/6839/files#diff-9f7b57df79bdb2c539ae04506d0f5a67) and [User](https://github.com/department-of-veterans-affairs/caseflow/pull/6839/files#diff-4676c008b11a5480d73d4a6de01e45b9) models).
2. Frontend changes to display the task actions dropdown for organizational queues (recommend starting by looking at [CaseSnapshot](https://github.com/department-of-veterans-affairs/caseflow/pull/6839/files#diff-5b5c7473c1568400a8104eccd0693108), then [the new selector](https://github.com/department-of-veterans-affairs/caseflow/pull/6839/files#diff-2a3f4e296b35f6015a2d1d3ef72e5506) for tasks assigned to an organization, how we're adding an [`organizationId` to state](https://github.com/department-of-veterans-affairs/caseflow/pull/6839/files#diff-34473acdd3a8771528f5978987311644), and finally the new view for the [MarkTaskCompleteView](https://github.com/department-of-veterans-affairs/caseflow/pull/6839/files#diff-fffbc4982a6e26763f9b1bbc24681c4c)
3. And finally, some changes we had to make to [task serializers](https://github.com/department-of-veterans-affairs/caseflow/pull/6839/files#diff-67e5f4d5cfceaa8c093045c1abcd5100) and [store preppers](https://github.com/department-of-veterans-affairs/caseflow/pull/6839/files#diff-d91d3970f54a1fbc30a05ad360bd5444) to accommodate assignees that are Organizations and not Users.

### Test Plan
1. Create a bunch of tasks for the translation organization
```
# In the rails console
o = Organization.where(name: "Translation").first
# Need to select a User this because tasks must be assigned by somebody
u = User.first
t = FactoryBot.create_list(:generic_task, 5, assigned_by: u, assigned_to: o)
```
2. Log in as "ORG Q USER" (You may need to run the FeatureToggle steps in `db/seeds.rb` manually in the rails console since they don't always take during the seed job... don't know why)
3. Navigate to `/organizations/translation`
4. Select any case and notice the task actions dropdown in the case snapshot section of the case details page (!!!)
5. Select the "Mark task complete" option in the dropdown menu.
6. Select "Mark complete" on the following page.
7. Notice that the task actions dropdown has disappeared when you return to the case details page.


app/controllers/tasks_controller.rb
```diff
@@ -73,7 +74,15 @@ def create
...
-    if task.assigned_to != current_user && task.assigned_by != current_user
+    # GenericTasks may be assigned to the organization and not the user, and not assigned by the current user either.
+    # GenericTask.update_from_params() will throw an error if the current user is not allowed to act on it though so
+    # we should be covered in that case.
+    #
+    # TODO: This logic should probably live in the Task model itself (when we call update_from_params probably).
+    # Rescue from the error and redirect to unauthorized when we throw an error.
+    if task.assigned_to != current_user &&
+       task.assigned_by != current_user &&
+       (task.class.name != GenericTask.name || !task.verify_user_access(current_user))
```
lowellrex: Added the `verify_user_access()` check here only for `GenericTask`s since the logic here for `Task`s is a little different from what exists in `Task.verify_user_access()` (task is assigned to or by current_user instead of the VACOLS role checks in the `Task` function).

This should probably be consolidated in a different PR, but until then I've gone with this approach.

app/models/tasks/generic_task.rb
```diff
@@ -6,11 +6,13 @@ def update_from_params(_params, current_user)
   end
 
   def verify_user_access(user)
-    return if assigned_to && assigned_to == user
+    return true if assigned_to && assigned_to == user
```
lowellrex: I don't love overloading this function (expecting a return value as well as a side effect) and think it might make sense to bust this out into two different functions. One that returns a boolean based on whether a user can access the task (something like `user_can_access?()`) and the other (`verify_user_access()`) that throws an exception if `user_can_access?()` returns false. Thoughts?

app/models/user.rb
```diff
@@ -141,6 +141,10 @@ def vso_employee?
     roles.include?("VSO")
   end
 
+  def organization_queue_user?
+    FeatureToggle.enabled?(:organization_queue, user: self)
```
lowellrex: Not sure this is the ultimate way we will determine who has access to organizational queues, but for now I can't think of anything better.

client/app/queue/components/DecisionViewBase.jsx
```diff
@@ -47,7 +47,7 @@ export default function decisionViewBase(ComponentToWrap, topLevelProps = defaul
 
     blockTransitions = () => this.unblockTransitions = this.props.history.block((location) => {
       const { pathname } = location;
-      const newPathInCheckoutFlow = /^\/queue\/appeals\/[a-zA-Z0-9-]+(?:\/\S+)/;
+      const newPathInCheckoutFlow = /^\/queue\/appeals\/[a-zA-Z0-9-]+(?:\/\S+)?/;
```
lowellrex: In order to avoid showing the cancel modal we allow the next path to be the case snapshot page.

client/app/queue/components/GenericTaskActionsDropdown.jsx
```diff
@@ -0,0 +1,98 @@
...
+    switch (props.value) {
+    case 'mark-task-complete':
+      history.push(`/queue/appeals/${appealId}/mark_task_complete`);
+      break;
+    // case "assign-to-team":
+    //   history.push(`/queue/appeals/${appealId}/assign_to_team`);
+    //   break;
+    // case "assign-to-person":
+    //   history.push(`/queue/appeals/${appealId}/assign_to_person`);
+    //   break;
```
lowellrex: Commented out for now because these will have to be implemented before we turn on the organizational task queues for anybody.

client/app/queue/selectors.js
```diff
@@ -88,6 +89,13 @@ export const tasksForAppealAssignedToAttorneySelector = createSelector(
   }
 );
 
+export const tasksForAppealAssignedToOrganizationSelector = createSelector(
+  [getTasksForAppeal, getOrganizationId],
+  (tasks: Tasks, id: Number) => {
+    return _.filter(tasks, (task) => task.assignedTo.id === id && task.status !== 'completed');
```
lowellrex: Added `task.status !== 'completed'` to avoid showing the task actions dropdown for tasks that have been completed and therefore should not have actions taken on them anymore. We might need a similar check on the backend.

client/app/queue/components/GenericTaskActionsDropdown.jsx
```diff
@@ -0,0 +1,98 @@
...
+|};
+
+class GenericTaskActionsDropdown extends React.PureComponent<Props> {
+  onChange = (props) => {
+    const {
+      appealId,
+      history
+    } = this.props;
+
+    this.props.stageAppeal(appealId);
```
dannysellers: We shouldn't need to stage appeals if we aren't going to modify them before saving changes to the server.

I've been trying to think of how we might be able to move away from staging appeals in general, as it can lead to site breakage after unexpected behavior (i.e. using the back button to go to a page that requires a staged appeal), though I haven't come up with any better solutions (besides managing browser history better).

Also, staged appeals don't play nice with our selectors, which all draw from `state.queue.appeals`/`appealDetails`

client/app/queue/selectors.js
dannysellers: Nit: we have [`incompleteTasksByAssigneeCssIdSelector`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/selectors.js#L104) which is `tasksByAssigneeCssIdSelector` -> `tasks.select(status !== 'completed')`. We could just move an `incompleteTasksSelector` further up the chain

app/controllers/application_controller.rb
```diff
@@ -109,13 +109,15 @@ def no_cache
     # :nocov:
   end
 
+  # rubocop:disable Metrics/CyclomaticComplexity
   def can_access_queue?
-    return true if current_user.vso_employee?
+    return true if current_user.organization_queue_user? || current_user.vso_employee?
```
mdbenjam: We determined that we couldn't just get rid of this function, right?

client/app/queue/selectors.js
mdbenjam: Yeah I like that idea. I think we can expand upon it further. So this currently could select a user task if a user and organization have the same id (unlikely). We really should also select for assignedToType, I think it would be nice to have the last level of selectors be for users or organizations based on this assignedToType i.e.
`incompleteTasksSelector` -> `incompleteTasksByAssigneeIdSelector` -> `incompleteUserTasksByAssigneeIdSelector` and `incompleteOrganizationTasksByAssigneeIdSelector`

client/app/queue/components/GenericTaskActionsDropdown.jsx
```diff
@@ -0,0 +1,98 @@
...
+    name={`start-generic-task-action-flow-${this.props.appealId}`}
+    placeholder="Select an action&hellip;"
+    options={this.getOptions()}
+    onChange={this.onChange}
+    hideLabel
+    dropdownStyling={dropdownStyling} />;
+}
+
+const mapStateToProps = (state: State, ownProps: Params) => ({
+  task: getTasksForAppeal(state, ownProps)[0],
```
mdbenjam: Do we use task at the moment? Maybe we'll need it when we support multiple tasks on a given appeal. In fact that's going to be a challenge for the URL situation. We might need a task id after the `mark_task_complete` part of the URL. Something to think about for the future.

client/app/queue/MarkTaskCompleteView.jsx
```diff
@@ -0,0 +1,95 @@
...
+      task,
+      appeal
+    } = this.props;
+
+    const successMsg = {
+      title: 'Task marked as complete',
+      detail: `Task for ${appeal.veteranFullName} completed`
+    };
+
+    const payload = { data: { task: { status: 'VALUE DOESNT MATTER' } } };
```
mdbenjam: Maybe pass '' instead of 'VALUE DOESNT MATTER'?

client/app/queue/OrganizationQueueLoadingScreen.jsx
```diff
@@ -26,6 +30,9 @@ class OrganizationQueueLoadingScreen extends React.PureComponent<Props> {
     (response) => {
       const { tasks: { data: tasks } } = JSON.parse(response.text);
 
+      if (tasks[0]) {
+        this.props.setOrganizationId(tasks[0].attributes.assigned_to.id);
```
mdbenjam: Haha this is janky, but it works for now. In the future we should probably move this here: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/views/queue/index.html.erb

app/controllers/tasks_controller.rb
mdbenjam: So I think we should re-arrange this to bring `verify_user_access` for Tasks and GenericTasks into line. I think `verify_user_access` should see if (in the case of Task) the user is who the task is assigned to or assigned by. In the case of GenericTask it can be what you currently have. Either the assigned to is the user (maybe also the assigned by) or it's assigned to an organization the user belongs to. What is currently in `Task.verify_user_access` should be moved somewhere else since it's more feature management.

app/controllers/application_controller.rb
lowellrex: I determined that it wouldn't be as straightforward as I'd hoped so I didn't attempt to do it in this PR.

client/app/queue/components/GenericTaskActionsDropdown.jsx
lowellrex: Good catch. It looks like we don't need task yet. I'm happy to remove this if you'd like.

mdbenjam: We'll need it going forward (at some point)

client/app/queue/selectors.js
```diff
@@ -103,7 +106,17 @@ export const tasksByAssigneeCssIdSelector = createSelector(
 
 export const incompleteTasksByAssigneeCssIdSelector = createSelector(
   [tasksByAssigneeCssIdSelector],
-  (tasks: Array<Task>) => tasks.filter((task) => task.status !== 'completed')
+  (tasks: Array<Task>) => incompleteTasksSelector(tasks)
+);
+
+export const organizationTasksByAssigneeIdSelector = createSelector(
+  [getTasksForAppeal, getOrganizationId],
+  (tasks: Tasks, id: Number) => _.filter(tasks, (task) => task.assignedTo.id === id)
```
mdbenjam: We need to check here that the assignedTo.type is "Organization" right?

PR #6840, opened 2018-09-05
Delete issues on cancel

### LINKED ISSUE #6709: Canceling RampRefilings or AMA Reviews should delete issues too
When we cancel a RampRefilingIntake it deletes the associated RampRefiling, but not the RampIssues associated to that refiling. Those RampIssues should be deleted as well.

As another item on this ticket, we should also clean up any orphaned RampIssues in prod.

leikkisa:
connects #6709

Delete RAMP or Request Issues when the intake is cancelled or rolled back.


app/models/ramp_election.rb
```diff
@@ -57,7 +57,7 @@ def rollback!
       # End product should already be cancelled, so we don't need to pay attention to the establishment that we already
       # had created. We will create a new one if the ramp election is recreated.
       end_product_establishment.destroy!
-
+      issues.destroy_all unless issues.empty?
```
leikkisa: Also adding this to ramp_election_rollback

app/models/intake.rb
```diff
@@ -126,6 +126,7 @@ def cancel!(reason:, other: nil)
 
   def cancel_detail!
     detail.destroy!
+    detail.remove_issues!
```
shanear: I think you may be able to add this as a `before_destroy` callback

https://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-destroy

this would transactionalize all of this.

app/models/ramp_election_intake.rb
```diff
@@ -58,6 +58,7 @@ def cancel_detail!
       receipt_date: nil,
       option_selected: nil
     )
+    remove_issues!
```
shanear: i don't think the ramp election would have issues at this point, so it think this might be a little misleading

leikkisa: I did notice that there are ramp_issues connected to RampElections, but I am guessing that they get added afterwards then?  I can take this out. I am still looking at the data clean up step, so I suppose I'll notice if there are any instances of this happening.

leikkisa:
We are running this script to clean up the RampRefiling issues:

RampIssue.joins("LEFT OUTER JOIN ramp_refilings on ramp_issues.review_id = ramp_refilings.id").where("ramp_refilings.id IS NULL").where(review_type: "RampRefiling").each(&:delete)

Issue #6851, opened 2018-09-06
Unify front-end task action dropdown components


lowellrex:
While implementing [generic task actions](https://github.com/department-of-veterans-affairs/caseflow/issues/6573), I noticed that we were repeating a lot of the same patterns and code that already existed for the [attorney](https://github.com/department-of-veterans-affairs/caseflow/blob/857e9c05eb55ce738f930977b855e50ef7670a71/client/app/queue/components/AttorneyActionsDropdown.jsx), [judge](https://github.com/department-of-veterans-affairs/caseflow/blob/857e9c05eb55ce738f930977b855e50ef7670a71/client/app/queue/components/JudgeActionsDropdown.jsx), and [co-located administrative staff](https://github.com/department-of-veterans-affairs/caseflow/blob/857e9c05eb55ce738f930977b855e50ef7670a71/client/app/queue/components/ColocatedActionsDropdown.jsx) task actions flows. This ticket exists to unify those components (and any new ones that will exist for `GenericTask`s).

I think this can be achieved in moderately small chunks.
* Make all `...ActionsDropdown`s push some URL to the address bar when they change, and create ReactRouter routes that map to components which call whatever actions were being called in place of the URLs that will live there now.
* Make a single `TaskActionsDropdown` component that determines the options and `onChange()` URLs itself.
* Consider moving the determination of options and URLs to the backend (where we have access to the entire database, can walk the full task tree, get more information about related users/organizations, etc.) determine which actions should be available as options in the dropdown.

lowellrex:
Work completed by #7102.

PR #6853, opened 2018-09-06
Make ColocatedActionsDropdown redirect to URL onChange

lowellrex:
Connects #6851.

First step in unifying the task action dropdowns for the various roles.

### Test plan
1. Log in as a co-located admin staffer
2. Select a case
3. Select the "Send back to attorney" option from the dropdown
4. Notice how it works! Huzzah!


client/app/queue/ColocatedSendTaskToAttorneyView.jsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+
+import { SEND_TO_LOCATION_MODAL_TYPES } from './constants';
+import { showModal } from './uiReducer/uiActions';
+
+class ColocatedSendTaskToTeamView extends React.Component {
```
dannysellers: These could probably be `PureComponent`s

client/app/queue/ColocatedSendTaskToAttorneyView.jsx
```diff
@@ -0,0 +1,21 @@
...
+  constructor(props) {
```
dannysellers: Could we do something like this?
```javascript
constructor = (props) => {
  super(props);
  this.props.showModal(ownProps.modalType);
}

componentWillUnmount = () => {
  if (this.props.modal) {
    this.props.hideModal(this.props.modalType);
  }
}

  modal: getActiveModalType(state)
});
```

In `SendToLocationModal`, we use `getActiveModalType` to get just the active modal type, which we then use to get copy from an object (`SEND_TO_LOCATION_MODAL_TYPE_ATTRS`).

It seems like the only difference between these two views is which modal they display. If we had a `{ taskType: modalType }` mapping, we could have one generic send to location modal for all actions?

client/app/queue/QueueApp.jsx
```diff
@@ -263,6 +273,16 @@ class QueueApp extends React.PureComponent<Props> {
             path="/queue/appeals/:appealId/place_hold"
             title="Place Hold | Caseflow"
             render={this.routedColocatedPlaceHold} />
+          <PageRoute
```
dannysellers: As long as we put the route after the other specific routes, we could have a generic path like `"/queue/appeals/:appealId/:actionType"`, and just validate the action against a set of possible actions (a la `SendToLocationModal. SEND_TO_LOCATION_MODAL_TYPE_ATTRS`)

client/app/queue/TriggerModal.jsx
```diff
@@ -0,0 +1,22 @@
...
+import { bindActionCreators } from 'redux';
+import { withRouter } from 'react-router-dom';
+
+import { showModal } from './uiReducer/uiActions';
+
+class TriggerModal extends React.Component {
+  constructor(props) {
+    super(props);
+    this.props.showModal(props.modal);
+    this.props.history.replace(this.props.history.location.state.previous);
```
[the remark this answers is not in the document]
dannysellers: Given that the url doesn't visibly change, I'm not sure we need to worry about the back button closing the modal at this point.

But, it looks like if we just do `this.props.history.goBack()` instead of `replace`ing the `/queue/modal` entry, we'll end up at `/queue/appeals/${appealId}` with the modal open, and the previous history entry will be `/queue`

PR #6859, opened 2018-09-07
changes to the judge algorithm - Assign all hearing days first based on UA

tejans24:

### Description

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...


app/services/hearing_schedule/assign_judges_to_hearing_days.rb
```diff
@@ -95,22 +114,20 @@ def sort_hearing_days_by_non_avail
...
 
-  def verify_assignments(unassigned_hearing_days, prior_iter_days_assigned, assigned_hearing_days)
-    if (prior_iter_days_assigned == assigned_hearing_days.length) &&
-       assigned_hearing_days.length != @video_co_hearing_days.length
-      dates = unassigned_hearing_days.map(&:hearing_date)
-
+  def verify_assignments
+    if @assigned_hearing_days.length != @video_co_hearing_days.length
       if @algo_counter >= 20
+        dates = @unassigned_hearing_days.map(&:hearing_date)
```
cmgiven: Note that this no longer works, as `@ unassigned_hearing_days` is not updated by `assign_remaining_hearing_days`. However, this is no longer a likely error state: this new algorithm will virtually always find a way to assign every hearing day, the only problem would be if it ends up assigning too many or too few to some judges as a result of their availability and a cranked `days_of_separation` value.

app/services/hearing_schedule/assign_judges_to_hearing_days.rb
```diff
@@ -95,22 +114,20 @@ def sort_hearing_days_by_non_avail
...
     end
   end
 
-  def verify_assignments(unassigned_hearing_days, prior_iter_days_assigned, assigned_hearing_days)
-    if (prior_iter_days_assigned == assigned_hearing_days.length) &&
-       assigned_hearing_days.length != @video_co_hearing_days.length
-      dates = unassigned_hearing_days.map(&:hearing_date)
-
+  def verify_assignments
+    if @assigned_hearing_days.length != @video_co_hearing_days.length
```
cmgiven: We should probably add another error condition when a judge is assigned too many hearings. We'd probably like no one to get more than `(num_hearings / num_judges).ceil + 1` hearings. Fortunately, we can reuse `get_sorted_judges_by_assignments` to run this check.

app/services/hearing_schedule/assign_judges_to_hearing_days.rb
```diff
@@ -21,60 +22,78 @@ def initialize(schedule_period)
     fetch_hearing_days_for_schedule_period
   end
 
-  # rubocop:disable Metrics/MethodLength
   def match_hearing_days_to_judges
-    assigned_hearing_days = []
-    unassigned_hearing_days = fetch_hearing_days_for_matching
+    @assigned_hearing_days = []
+    @unassigned_hearing_days = @video_co_hearing_days.shuffle
```
cmgiven: If we're not going to use it, we should delete `fetch_hearing_days_for_matching`. I'd suggest the same probably applies to `fetch_judges_for_matching` on L38.

app/services/hearing_schedule/assign_judges_to_hearing_days.rb
```diff
@@ -21,60 +22,78 @@ def initialize(schedule_period)
...
-            end
-            index += 1
-            throw :hearing_days_assigned if all_hearing_days_assigned
-          end
+    sorted_judges.each do |css_id|
+      days_assigned = 0
+
+      @unassigned_hearing_days.delete_if do |current_hearing_day|
+        break if days_assigned >= max_days_per_judge
+        unless can_day_be_assigned(current_hearing_day, @assigned_hearing_days, css_id)
```
cmgiven: Flip the logic of `can_day_be_assigned` to match its name, and use an `if` here.

app/services/hearing_schedule/assign_judges_to_hearing_days.rb
```diff
@@ -21,60 +22,78 @@ def initialize(schedule_period)
...
+  private
+
+  def evenly_assign_judges_to_hearing_days
     sorted_judges = fetch_judges_for_matching
 
+    judge_count = sorted_judges.length
     total_hearing_day_count = @video_co_hearing_days.length
+    max_days_per_judge = (total_hearing_day_count.to_f / judge_count).floor
+
+    days_of_separation = 3
```
cmgiven: Good call to make this a constant. We can delete this local then.

app/services/hearing_schedule/assign_judges_to_hearing_days.rb
```diff
@@ -6,6 +6,7 @@ class HearingSchedule::AssignJudgesToHearingDays
 
   TB_ADDITIONAL_NA_DAYS = 3
   CO_ROOM_NUM = "1".freeze
+  DAYS_OF_SEPERATION = 3
```
cmgiven: `SEPARATION`. 😄

Issue #6854, opened 2018-09-06
Create attorney special issues page


laurjpeterson:
For the RAMP appeals pilot, we want to experiment with attorneys entering special issues (that today are entered by ARC staff in Caseflow Dispatch) because (1) this simplifies the amount of work we have to do for dispatching RAMP appeals today and (2) we plan to deprecate Caseflow Dispatch down the line anyway.

## AC
- Create a screen titled `Select special issues`
- `Select special issues` screen comes after selecting remand reasons in the checkout flow
- The following special issues should be listed on this page (TODO)

## Mocks

See [mural](https://app.mural.co/t/workqueue2001/m/workqueue2001/1536261263374/a50fb8424a7bff4e04c49d0b4023a6a63e519f09) for flow.

This is how special issues are displayed in Caseflow Dispatch. Let's replicate it.
<img width="670" alt="screen shot 2018-09-06 at 7 12 30 pm" src="https://user-images.githubusercontent.com/27704093/45189842-e833ee00-b208-11e8-85f9-5d7b36a437c3.png">

laurjpeterson:
@mdbenjam - I got the answer from AMO that these special issues do not apply for RAMP appeals, so we should remove them from the Special issues page:

- Vocational  Rehab
- DIC - death, or accrued benefits - United States
- Education - GI Bill, dependents educational assistance
- Foreign pension, DIC - Mexico, Central and South America,   Caribbean
- Foreign pension, DIC - all other foreign countries
- Home Loan Guaranty
- Insurance
- National Cemetery Administration
- Pension - United States
- VAMC
- Waiver of   Overpayment
- Scholarship, transfer of entitlement

laurjpeterson:
@mdbenjam 

Option 1: Remove the below for AMA appeals:
- Hearing
- Manlincon Compliance
Both of these are not relevant for AMA. 

Option 2: If you don't want to fork these two pages based on if an appeal is legacy or AMA in the future, you can append them
- Hearing ... (legacy appeals only)
- Manlincon Compliance (legacy appeals only)

allyceh:
@laurjpeterson Is this the one we should re-open for design + content for the special issues page?

When shadowing an attorney working a RAMP case yesterday, she saw the special issues page for the first time and noticed it was new. She understood it after we explained it to her, but some additional context around 1) what these special issues are, 2) what the attorney needs to do on this page, 3) how we use any special issues they check (routing purposes) could help orient first time users. 

[...]

PR #6871, opened 2018-09-07
Adding special issue page

mdbenjam:
Connects #6854 

### Description
This PR is the first part of adding special issues. It creates a new special issues table that is polymorphicly associated with both the appeal and legacy appeal tables. It also creates a special issues endpoint the page can use to update and get the currently selected special issues.

The PR also creates the frontend for special issues. It currently doesn't live within the attorney & judge checkout flows but will in the next PR. 

### Testing Plan
1. Insert the page into the flow by changing `SelectDispositionsView`'s `prevPage` to `/queue/appeals/${props.match.params.appealId}/special_issues` and updating the `      draft_decision` key in AttorneyActionsDropdown to be `special_issues` instead of `dispositions`.
1. Go through attorney checkout flow, check all the boxes, click next
1. Restart the flow and ensure all the boxes are still checked.


app/controllers/special_issues_controller.rb
```diff
@@ -0,0 +1,49 @@
+class SpecialIssuesController < ApplicationController
+  # before_action :validate_access_to_task
+
+  def create
+    return record_not_found unless appeal
+
+    if appeal.special_issue
+      appeal.special_issue.update(speical_issue_params)
```
amprokop: typo

app/controllers/special_issues_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+    else
+      appeal.special_issue = SpecialIssue.create(speical_issue_params)
```
amprokop: hmm this is a common typo

app/controllers/issues_controller.rb
```diff
@@ -48,7 +48,7 @@ def json_issues
   end
 
   def validate_access_to_task
-    current_user.access_to_task?(appeal.vacols_id)
+    redirect_to "/unauthorized" unless current_user.access_to_legacy_task?(appeal.vacols_id)
```
mdbenjam: Bug I noticed while doing this fix. We weren't redirecting on this check.

client/app/queue/components/DecisionViewBase.jsx
```diff
@@ -142,7 +142,7 @@ export default function decisionViewBase(ComponentToWrap, topLevelProps = defaul
         goToNextStep: nextStepHook = null
       } = this.state.wrapped;
 
-      if (!validation || !validation()) {
+      if (validation && !validation()) {
```
mdbenjam: We should only fail validation if a validation function was given to us.

client/app/queue/reducers.js
```diff
@@ -24,6 +24,7 @@ export const initialState = {
   editingIssue: {},
   docCountForAppeal: {},
   newDocsForAppeal: {},
+  specialIssues: {},
```
mdbenjam: I think it's okay to store here in a non specific appeal way since we always have the loading screen to get the latest special issues before showing the checklists.

app/controllers/issues_controller.rb
amprokop: 👍

app/models/appeal.rb
```diff
@@ -3,6 +3,7 @@ class Appeal < AmaReview
   has_many :claims_folder_searches, as: :appeal
   has_many :tasks, as: :appeal
   has_many :decision_issues, through: :request_issues
+  has_one :special_issue
```
amprokop: only one? can't an appeal deal with more than one of these?

app/models/user.rb
```diff
@@ -84,10 +84,20 @@ def vsos_user_represents
     @vsos_user_represents ||= bgs.fetch_poas_by_participant_id(participant_id)
   end
 
-  def access_to_task?(vacols_id)
+  def access_to_legacy_task?(vacols_id)
     self.class.user_repository.can_access_task?(css_id, vacols_id)
   end
 
+  def access_to_appeal?(appeal)
```
amprokop: hmm does this confuse the concept of having access to an appeal (all BVA users should have access to all open appeals prob) with the concept of being currently assigned an appeal? not sure if this is just a naming issue

client/app/queue/SpecialIssueLoadingScreen.js
```diff
@@ -0,0 +1,63 @@
+// @flow
```
amprokop: should this be a jsx file?

client/app/queue/components/DecisionViewBase.jsx
amprokop: 👍

client/app/queue/reducers.js
amprokop: sounds good

Issue #6822, opened 2018-09-04
Add removed_at and rating_issue_associated_at to request issues


shanear:
We need to record in the caseflow DB when certain VBMS actions for request issues succeeded. This way we know if we need to revert or retry. 

## Acceptance criteria
- When the contention for a request issue is successfully removed, set `removed_at` on that request issue with the timestamp
- If the removal of the contention fails, `removed_at` should not be set.
- If the request issue is a rating request issue, when the rating issue is successfully associated to it, set the `rating_issue_associated_at` timestamp. 

## Technical notes
This should probably happen in `EndProductEstablishment.remove_contention!`

related to https://github.com/department-of-veterans-affairs/caseflow/issues/6212

## Validation instructions
- Flow: AMA, pick a claim_review (Higher Level Review or Supplemental Claim)
- Test file number: 123458310
- Go through the flow, add at least one rated issue, establish the claim
- Go to the edit page for the claim
- Remove one of the contentions, please note here which contention you removed.
- Ask an engineer to check the data

For an engineer:
- Connect to the UAT Jumpbox, find the most recent rating end_product_establishment for the file number above
- Find the rating request issues associated with the end_product_establishment, ensure that there is a timestamp in rating_issue_associated_at , and for the removed contention ensure that there is a removed_at timestamp

shanear:
This is blocked on https://github.com/department-of-veterans-affairs/caseflow/issues/6212

shanear:
<img width="1428" alt="screen shot 2018-10-01 at 4 27 53 pm" src="https://user-images.githubusercontent.com/1596259/46313860-fae6dc00-c596-11e8-9939-0103a64b04ab.png">

`removed_at` is working 👍

PR #6872, opened 2018-09-08
Record issue removed_at and rating_associated_at timestamps

leikkisa:
connects #6822


app/models/claim_review.rb
```diff
@@ -79,6 +79,10 @@ def create_associated_rated_issues!(end_product_establishment, issues)
       claim_id: end_product_establishment.reference_id,
       rated_issue_contention_map: rated_issue_contention_map(request_issues_to_associate)
     )
+
+    request_issues_to_associate.each do |request_issue|
+      request_issue.update!(rating_issue_associated_at: Time.zone.now)
+    end
```
shanear: It might be better to do this as an `update_all` so we only have to send one DB query.

Something like this?
```
RequestIssue.where(id: request_issues_to_associate.map(&:id)).update_all(rating_issue_associated_at: Time.zone.now)
```

leikkisa:
Added a VBMS error test for remove_contentions

PR #6879, opened 2018-09-10
Handle non poa responses

aroltsch:
When there was no POA, the response is in the following format: 

```{:file_number=>"710332055", :message=>"No POA found for 600231213", :ptcpnt_id=>"600231213"}```

However, the code expected an array of poas. The fix is to handle No POA responses.

### CODE REVIEW

File: app/models/appeal.rb [hunk 1]
```diff
@@ -114,7 +114,7 @@ def docket_number
 
   # For now power_of_attorney returns the first claimant's power of attorney
   def power_of_attorney
-    claimants.first.power_of_attorney
+    claimants.first.power_of_attorney if claimants.first
```
aroltsch: Not all appeals have claimants.

app/models/appeal.rb
```diff
@@ -74,7 +74,7 @@ def appellant
     claimants.first
   end
 
-  delegate :first_name, :last_name, :middle_name, :name_suffix, to: :appellant, prefix: true
+  delegate :first_name, :last_name, :middle_name, :name_suffix, to: :appellant, prefix: true, allow_nil: true
```
aroltsch: Not all appeals have claimants.

PR #6880, opened 2018-09-10
Move logic out of AttorneyActionsDropdown.changeRoute()

lowellrex:
Connects #6851.

Move logic specific to the attorney task actions out of the `AttorneyActionsDropdown` component in preparation for the unification of the task action dropdown components.


client/app/queue/SubmitDecisionView.jsx
```diff
@@ -89,6 +93,11 @@ class SubmitDecisionView extends React.PureComponent<Props> {
     return !missingParams.length;
   };
 
+  componentDidMount = () => {
+    this.props.resetDecisionOptions();
+    this.props.setCaseReviewActionType('omo_request');
```
dannysellers: We probably can't get away with hardcoding this, unfortunately, as `SubmitDecisionView` is used for both attorney checkout flows

PR #6885, opened 2018-09-10
Hide links to Reader from VSO employees

### LINKED ISSUE #6847: Hide Veteran Documents column in VSO task queue table
Also hide the following two links on the case details page.

image

lowellrex:
Resolves #6847.

This PR hides links to Reader from the task table and the case details page for VSO employees.

The changes that actually hide the elements are in the [`TaskTable` component](https://github.com/department-of-veterans-affairs/caseflow/pull/6885/files#diff-7af7b35ac72a61f05d3c01eb77d25333R221) and the [`CaseTitle` component](https://github.com/department-of-veterans-affairs/caseflow/pull/6885/files#diff-f95f729a7c1f0718fc557099430f8dabR94).

hide_reader_links


app/views/queue/index.html.erb
```diff
@@ -5,6 +5,7 @@
     userRole: (current_user.vacols_roles.first || "").capitalize,
     userCssId: current_user.css_id,
     userCanAccessQueue: can_access_queue?,
+    userIsVsoEmployee: current_user.vso_employee?,
```
lowellrex: I think passing this information from the back-end explicitly is easier than trying to determine if the current user is a VSO employee from some other information (for instance, if we relied on the first task's assignee type how would we know to hide the links in the case details page when a VSO employee searches for an appeal they should have access to).

client/app/queue/CaseTitle.jsx
```diff
@@ -134,6 +135,6 @@ export default connect(mapStateToProps, mapDispatchToProps)(CaseTitle);
 const CaseTitleScaffolding = (props) => <div {...containingDivStyling}>
   <h1 {...headerStyling}>{props.heading}</h1>
   <ul {...listStyling}>
-    {props.children.map((child, i) => <li key={i} {...listItemStyling}>{child}</li>)}
+    {props.children.map((child, i) => child && <li key={i} {...listItemStyling}>{child}</li>)}
```
lowellrex: Don't draw a list item if the child element evaluates to false.

client/app/queue/CaseTitle.jsx
```diff
@@ -91,19 +91,19 @@ class CaseTitle extends React.PureComponent {
...
         appealId={appealId}
         analyticsSource={CATEGORIES[analyticsSource.toUpperCase()]}
         redirectUrl={redirectUrl}
         appeal={appeal}
         taskType={taskType}
-        longMessage />
+        longMessage /> }
 
-      <span {...viewCasesStyling}>
+      { !this.props.userIsVsoEmployee && <span {...viewCasesStyling}>
```
mdbenjam: Why do we need to hide this?

PR #6895, opened 2018-09-10
Create BVA dispatch task after judge checkout

### LINKED ISSUE #6886: AMA Judges | Create BVA Dispatch task when judge checkout is compelete
To ensure that BVA Dispatch team members can view appeals assigned to them, when a judge completes the checkout flow for an AMA appeal, we should create tasks for BVA Dispatch team members. This will probably be using the new organizational queue backend.

# AC
- When judge checkout is complete for AMA appeals, we should create new tasks for BVA dispatch users 
- Caseflow perform round-robin assignment for BVA Dispatch users (there are currently only two in production)
- Bonus: create seed data in the local dev environment to seed tasks for BVA Dispatch users

lowellrex:
Resolves #6886.

## TODO
* Explain the changes made in this PR
* Write tests

### DIFF
--- a/app/controllers/case_reviews_controller.rb
+++ b/app/controllers/case_reviews_controller.rb
@@ -16,6 +16,9 @@ class CaseReviewsController < ApplicationController
     record = case_review_class.complete(complete_params)
     return invalid_record_error(record) unless record.valid?
 
+    root_task = Task.find(params[:task_id]).root_task
+    BvaDispatchTask.create_and_assign(root_task)
+
     response = { task: record }
     response[:issues] = record.appeal.issues
     render json: response
--- a/app/models/organizations/bva_dispatch.rb
+++ b/app/models/organizations/bva_dispatch.rb
@@ -1,0 +1,5 @@
+class BvaDispatch < Organization
+  def self.singleton
+    BvaDispatch.first || BvaDispatch.create(name: "Board Dispatch")
+  end
+end
--- a/app/models/tasks/bva_dispatch_task.rb
+++ b/app/models/tasks/bva_dispatch_task.rb
@@ -1,0 +1,20 @@
+class BvaDispatchTask < GenericTask
+  include RoundRobinAssigner
+
+  class << self
+    def create_and_assign(root_task)
+      parent = create!(assigned_to: BvaDispatch.singleton, parent_id: root_task.id)
+      create!(
+        appeal: parent.appeal,
+        parent_id: parent.id,
+        assigned_to: next_assignee
+      )
+    end
+
+    private
+
+    def list_of_assignees
+      Constants::BvaDispatchTeams::USERS[Rails.current_env]
+    end
+  end
+end
--- a/lib/constants/bva_dispatch_teams.rb
+++ b/lib/constants/bva_dispatch_teams.rb
@@ -1,0 +1,9 @@
+module Constants::BvaDispatchTeams
+  USERS = {
+    preprod: %w[CF_KIRK_317],
+    uat: %w[CF_SPOCK_317 CF_KHAN_397],
+    test: %w[BVATEST1 BVATEST2 BVATEST3],
+    development: %w[BVALSPORER],
+    prod: %w[]
+  }.freeze
+end
--- a/spec/factories/task.rb
+++ b/spec/factories/task.rb
@@ -27,6 +27,14 @@ FactoryBot.define do
       completed_at Time.zone.now
     end
 
+    factory :root_task do
+      type "RootTask"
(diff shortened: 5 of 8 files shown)


app/controllers/case_reviews_controller.rb
```diff
@@ -16,13 +16,21 @@ def complete
...
     response[:issues] = record.appeal.issues
     render json: response
   end
 
   private
 
+  def root_task_for_case_review(record)
+    return record.task.root_task if record.task
+    RootTask.create!(appeal: record.appeal, assigned_to: Bva.singleton)
+  end
```
lowellrex: Tasks for `LegacyAppeal`s will not have `RootTask`s so we will need to create them.

app/models/organizations/bva_dispatch.rb
```diff
@@ -0,0 +1,5 @@
+class BvaDispatch < Organization
+  def self.singleton
+    BvaDispatch.first || BvaDispatch.create(name: "Board Dispatch")
```
lowellrex: Similar pattern to what we established with the [`BVA organization`](https://github.com/department-of-veterans-affairs/caseflow/blob/4b32bf6f2830debe7f7363750fcca1fa93297d48/app/models/organizations/bva.rb#L3).

app/models/tasks/bva_dispatch_task.rb
```diff
@@ -0,0 +1,20 @@
+class BvaDispatchTask < GenericTask
+  include RoundRobinAssigner
+
+  class << self
+    def create_and_assign(root_task)
+      parent = create!(assigned_to: BvaDispatch.singleton, parent_id: root_task.id, appeal: root_task.appeal)
```
lowellrex: Adding the organization task here allows us to have visibility into child tasks for all members of this organization (so another BVA dispatcher could complete their coworker's task if the coworker goes on vacation, for instance). I don't think creating this organizational task adds much complexity, but I am happy to remove it if you think it wise.

lib/constants/bva_dispatch_teams.rb
```diff
@@ -0,0 +1,9 @@
+module Constants::BvaDispatchTeams
+  USERS = {
+    preprod: %w[CF_KIRK_317],
+    uat: %w[CF_SPOCK_317 CF_KHAN_397],
+    test: %w[BVATEST1 BVATEST2 BVATEST3],
+    development: %w[BVALSPORER],
+    prod: %w[]
+  }.freeze
```
lowellrex: No idea who this list of users should be, so I've just nabbed what we've got for Colocated teams at the moment.

app/controllers/case_reviews_controller.rb
```diff
@@ -16,13 +16,21 @@ def complete
...
+    BvaDispatchTask.create_and_assign(root_task)
+
     response = { task: record }
     response[:issues] = record.appeal.issues
     render json: response
   end
 
   private
 
+  def root_task_for_case_review(record)
```
lowellrex: Tasks for `LegacyAppeal`s will not have `RootTask`s so we will need to create them.

mdbenjam: Do we want to be creating these tasks for legacy appeals? Don't we already handle those via VACOLS?

app/models/tasks/bva_dispatch_task.rb
mdbenjam: I don't like it... I love it.

app/models/tasks/root_task.rb
```diff
@@ -1,9 +1,15 @@
 class RootTask < Task
+  after_initialize :set_assignee
```
mdbenjam: Are we sure after_initialize runs before the save of create? I just haven't used this one so am unsure.

spec/factories/task.rb
```diff
@@ -27,6 +27,14 @@
       completed_at Time.zone.now
     end
 
+    factory :root_task do
+      type "RootTask"
+      appeal_type "Appeal"
+      appeal { create(:appeal) }
+      assigned_by { nil }
+      assigned_to { Bva.singleton }
```
mdbenjam: Maybe even remove this to more fully test our after_initialize action.

app/models/tasks/root_task.rb
lowellrex: Correct, this does indeed run before the save of `create()`. Tested this by manually in the rails console (and will add a test to our automated test suite), and the validation failed (and I imagine the database constraint on `assigned_to_type` would even if the validation didn't exist).

PR #6896, opened 2018-09-10
6351 add name to veteran table

### LINKED ISSUE #6351: Save Veteran first, middle, last, and suffix in the Caseflow DB when an appeal goes through intake
When an appeal goes through Caseflow Intake, we should save the most current veteran first name, middle name, last name, and suffix from BGS in a new veteran record or in one that already exists. 

For veteran rows already in our DB, the next time we fetch BGS information, we should save the Veteran name to backfill data.

# AC
- `first_name`, `middle_name` `last_name`, and `name_suffix` are added to the Veterans table
- Middle name and suffix are fetched from BGS (currently we only fetch first name and last name)
- When Caseflow Intake is used to intake an appeal, we fetch the Veteran name from BGS and either save it to  a new Veterans row or update the existing row
- For Veteran rows already in the DB, if the name is blank, the next time we fetch data from BGS, save the Veteran name in the DB

kahlouie:
Connects #6351

### DIFF
--- a/app/models/veteran.rb
+++ b/app/models/veteran.rb
@@ -5,8 +5,8 @@
 class Veteran < ApplicationRecord
   include AssociatedBgsRecord
 
-  bgs_attr_accessor :ptcpnt_id, :sex, :first_name, :last_name, :name_suffix,
-                    :ssn, :address_line1, :address_line2, :address_line3, :city,
+  bgs_attr_accessor :ptcpnt_id, :sex, :ssn,
+                    :address_line1, :address_line2, :address_line3, :city,
                     :state, :country, :zip_code, :military_postal_type_code,
                     :military_post_office_type_code, :service, :date_of_birth,
                     :date_of_death
@@ -137,7 +137,15 @@ class Veteran < ApplicationRecord
       return nil unless veteran.found?
 
       before_create_veteran_by_file_number # Used to simulate race conditions
-      veteran.tap { |v| v.update!(participant_id: v.ptcpnt_id) }
+      veteran.tap do |v|
+        v.update!(
+          participant_id: v.ptcpnt_id,
+          first_name: v.bgs_record[:first_name],
+          last_name: v.bgs_record[:last_name],
+          middle_name: v.bgs_record[:middle_name],
+          name_suffix: v.bgs_record[:name_suffix]
+        )
+      end
     rescue ActiveRecord::RecordNotUnique
       find_by(file_number: file_number)
     end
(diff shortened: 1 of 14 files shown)


leikkisa:
It looks like there's a bunch of interactions on this. I'm wondering a couple of things, do we want to add them to the database as new attributes, so people can switch over to using our database optionally, but otherwise the default will be to use the original bgs attributes?

app/models/veteran.rb
```diff
@@ -5,7 +5,7 @@
 class Veteran < ApplicationRecord
   include AssociatedBgsRecord
 
-  bgs_attr_accessor :ptcpnt_id, :sex, :first_name, :last_name, :ssn,
+  bgs_attr_accessor :ptcpnt_id, :sex, :ssn,
```
amprokop: What happens to Veteran rows already in the DB before this is deployed? I think we'll be missing their names from the DB and no longer fetch them from BGS, correct me if I'm wrong.

If this is truly a problem, I suggest that in `find_or_create_by_file_number`, if we find an existing veteran row we still update the name from BGS if the veteran row has a nil value for `first_name`

amprokop:
One comment, about backfilling existing rows in the Veteran table (there are already some from BEEAM and other Intake activity)— looks like we need to make a change to fetch name if needed.

amprokop:
Thank you!

spec/factories/veteran.rb
```diff
@@ -1,5 +1,9 @@
 FactoryBot.define do
   factory :veteran do
+    first_name "Bob"
```
leikkisa: I don't completely understand why you'd want to define it here and in the bgs_veteran_record separately. Do you still need it in the bgs_veteran_record below?

lib/fakes/appeal_repository.rb
```diff
@@ -355,7 +355,7 @@ def self.seed_intake_data!
       file_number = "#{i + 1}5555555"
       claim_id = "FAKEEP123#{i}"
 
-      Generators::Veteran.build(file_number: file_number)
+      Generators::Veteran.build(file_number: file_number, first_name: "Bob", last_name: "Smith")
```
leikkisa: I'm curious why you needed to explicitly define the first and last name when the Veteran generator is being used, it seems like it should pick a random'ish name?  If they don't have a name in the database, it should get created when necessary right (hence all of the reloads?)

spec/feature/queue/ama_queue_spec.rb
```diff
@@ -56,7 +56,12 @@
         create(
           :appeal,
           :advanced_on_docket,
-          veteran: create(:veteran, participant_id: veteran_participant_id, bgs_veteran_record: { first_name: "Pal" }),
+          veteran: create(
+            :veteran,
+            participant_id: veteran_participant_id,
+            first_name: "Pal",
```
leikkisa: is the bgs_vet_record still needed?

spec/factories/veteran.rb
kahlouie: the bgs_veteran_record is used in the Fakes::BGSService, and could potentially test whether or not the name has changed in some way. Editing one, but not the other. The second one is supposed to handle a base for what it should be returning from bgs.

PR #6899, opened 2018-09-10
onboarding readme fixes

kencheeto:
### Description
Two small fixes after onboarding:

* reordering db setup in `rake local:build` since the FACOLS container needs a bit more time before it is responsive. I haven't dug into this, but @jiahuang may have more insight
* added a mention of how to accept the license agreement for Oracle

### Acceptance Criteria
- [x] `bundle exec rake local:build` works

### Testing Plan
N/A


README.md
```diff
@@ -168,18 +167,6 @@ This should install clean. If you have errors, you probably missed a dependency
...
 This should install clean. If you have errors, try ... FIXME.
 
-### Setup the development Postgres user
-
-Add these to your `.bash_profile`:
-
-```
-export POSTGRES_HOST=localhost
-export POSTGRES_USER=postgres
-export POSTGRES_PASSWORD=postgres
```
pkarman: I find I still need these env vars set with the Docker image. It looks like `config/database.yml` is looking for them, and the `POSTGRES_HOST` in particular is necessary to coerce the tcp connection rather than the socket connection.

pkarman: In addition I found:
```
export NLS_LANG=American_America.UTF8
```
helpful to quiet one of the Oracle warnings on startup.

lib/tasks/local/build.rake
```diff
@@ -13,6 +13,8 @@ namespace :local do
       print "."
       sleep 1
     end
+    # Wait a bit longer - the FACOLS container is not responsive yet
+    sleep 30
```
pkarman: could we maybe change the `sleep 1` line 14 instead to `sleep 2`? That would add a potential 180 seconds to the total wait time, w/o the extra sleep line.

pkarman:
looking good @kencheeto -- I ran into these same issues.

README.md
kencheeto: Ah, good catch, added in d7311e83d

lib/tasks/local/build.rake
kencheeto: I think the issue here is that `docker-compose ps | grep 'health: starting'`.strip.chomp.empty?` becomes truthy, exiting the sleep loop, but the FACOLS container is still not responsive causing the subsequent `Setting up development FACOLS` step to fail. That means increasing the `sleep 1` to `sleep 2` would still attempt to start FACOLS db setup as soon as the health check passes.

I removed this sleep and moved the FACOLS db setup after the caseflow db setup which should accomplish about the same thing without waiting explicitly, adding a note. wdyt?

README.md
```diff
@@ -176,8 +175,11 @@ Add these to your `.bash_profile`:
 export POSTGRES_HOST=localhost
 export POSTGRES_USER=postgres
 export POSTGRES_PASSWORD=postgres
+export NLS_LANG=American_America.UTF8
```
ferlatte: DANGER. Production uses ASCII, not UTF8. Maybe set it do US7ASCII to make the warning go away?

README.md
```diff
@@ -64,7 +64,6 @@ Install the base dependencies via Homebrew:
     brew install rbenv nodenv yarn
     brew tap ouchxp/nodenv
     brew install nodenv-nvmrc
-    brew install postgresql
```
ferlatte: You might still want these for the psql utility?

lib/tasks/local/build.rake
pkarman: that's better.

I guess what I'm really after is an approach that doesn't rely on the "about 30 seconds" logic or the side effect of shuffling the order, but instead has some way of explicitly checking that FACOLS really is available, and if not, keeps trying for a reasonable amount of time, and then gives a helpful fatal error if ultimately unsuccessful. Right now the code is sort of "best effort".

That said, maybe that's all way out of scope for what you're trying to do here in this PR, which is to be friendlier for us first-timers.

README.md
pkarman: I think `rails dbconsole` relies on the presence of `psql` so yes.

README.md
pkarman: yes `US7ASCII` is the default in the warning -- good to know, I will change mine.

README.md
kencheeto: Fair - also you're right about the pg gem needing postgres, will readd

README.md
ferlatte: To be clear: I think the "correct" setting is this setting and we're kind of just getting lucky because everything in VACOLS is effectively ASCII data (so we're just happening to throw UTF8 bytes from Ruby at Oracle that happen to be safe to interpret as ASCII bytes). But production is using the US7ASCII setting now, so dev probably needs to match.

PR #6906, opened 2018-09-11
Increase timeout for claim_establishment idempotency migration

leikkisa:
connects #6822


nolawi:
Looks like there is a syntax error here, @leikkisa extra end line 28. We can't migrate otherwise.

PR #6907, opened 2018-09-11
Preparing the case details page for AMA appeals

mdbenjam:
### Description
The case details page is currently tailored for legacy appeals and tasks. This makes a few changes to bring AMA appeals and tasks in line.

In particular it does two things:
1) Starts displaying document id for in the case snapshot for AMA appeals
1) Simplifies the task section to only show the task, assigned by, instructions, and in some cases the preparer of the decision.

While working on this we also made the document id click to copy since judges have asked for that.

### Testing Plan
1. Go to BVASCASPER's queue. The first appeal is an AMA appeal. Click on it.
1. Ensure things look right on this case detail page.
1. Go through a checkout flow and assign it to another judge
1. Go to that judge's queue and ensure their case details page looks correct and the document id is present.


app/controllers/tasks_controller.rb
```diff
@@ -170,7 +170,7 @@ def json_tasks_by_legacy_appeal_id_and_role(appeal_id, role)
   end
 
   def json_tasks_by_appeal_id(appeal_db_id, appeal_type)
-    tasks = queue_class.new.tasks_by_appeal_id(appeal_db_id, appeal_type)
+    tasks = queue_class.new(user: current_user).tasks_by_appeal_id(appeal_db_id, appeal_type)
```
mdbenjam: This is needed to make this method actually get tasks

app/models/serializers/work_queue/task_serializer.rb
```diff
@@ -59,4 +59,15 @@ class WorkQueue::TaskSerializer < ActiveModel::Serializer
   attribute :issue_count do
     object.appeal.issues.count
   end
+
+  attribute :document_id do
```
mdbenjam: These are both temporary... Not clear these should be on the task, and we need a more generic way to handle prepared by. It's really just who completed the sub-task. But for now we're using this hack. See this conversation: https://dsva.slack.com/archives/C6E41RE92/p1536681826000100

client/app/queue/CaseSnapshot.jsx
```diff
@@ -86,12 +88,66 @@ export class CaseSnapshot extends React.PureComponent<Props> {
...
+        </React.Fragment> }
+      { assignedByAbbrev &&
+        <React.Fragment>
+          <dt>{COPY.CASE_SNAPSHOT_TASK_FROM_LABEL}</dt><dd>{assignedByAbbrev}</dd>
+        </React.Fragment> }
+      { taskAssignedToUser.instructions &&
+        <React.Fragment>
+          <dt>{COPY.CASE_SNAPSHOT_TASK_INSTRUCTIONS_LABEL}</dt><dd>{taskAssignedToUser.instructions}</dd>
+        </React.Fragment> }
+      { preparedByAbbrev &&
```
mdbenjam: Hoping for a more generic solution here in the future.

client/app/queue/CaseSnapshot.jsx
```diff
@@ -86,12 +88,66 @@ export class CaseSnapshot extends React.PureComponent<Props> {
...
+        <React.Fragment>
+          <dt>{COPY.CASE_SNAPSHOT_DECISION_PREPARER_LABEL}</dt><dd>{preparedByAbbrev}</dd>
+        </React.Fragment> }
+    </React.Fragment>
+  }
+
+  legacyTaskInformation = () => {
+    // If this is not a task attached to a legacy appeal, use taskInformation.
+    if (!this.props.appeal.locationCode) {
```
mdbenjam: Leaving the legacy case mostly the same.

client/app/queue/components/TaskTable.jsx
```diff
@@ -118,11 +118,14 @@ class TaskTable extends React.PureComponent<Props> {
     return this.props.includeDocumentId ? {
       header: COPY.CASE_LIST_TABLE_DOCUMENT_ID_COLUMN_TITLE,
       valueFunction: (task) => {
-        if (!task.assignedBy.firstName) {
+        const firstName = task.decisionPreparedBy ? task.decisionPreparedBy.firstName : task.assignedBy.firstName;
```
mdbenjam: In the legacy case the preparer is put into `assignedBy`, this is not true in the modern task case. AssignedBy would be who created this task, not who completed the sub task (i.e. the attorney writing the decision)

client/app/queue/CaseSnapshot.jsx
```diff
@@ -86,12 +88,66 @@ export class CaseSnapshot extends React.PureComponent<Props> {
     return null;
   };
 
-  taskAssignmentListItems = () => {
+  getAbbrevName = ({ firstName, lastName }) => {
+    return `${firstName.substring(0, 1)}. ${lastName ? lastName : ""}`;
```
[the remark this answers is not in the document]
mdbenjam: I'm not sure, but this comment made me realize I don't need that conditional anymore anyway!

PR #6908, opened 2018-09-11
Return veteran suffix name, return veteran death date, return outstan…

### LINKED ISSUE #6902: [IDT API] Add more information to certain fields
IDT API:
- [x] If veteran_is_deceased is true, I also need the date of death. 
- [x] If outstanding_mail is true, I need the type of mail with description
- [x] I need a suffix field for the Veteran (already have one for the Appellant)
- [ ] Need separate last/first name fields  for representatives when they are Agent/Attorney.

aroltsch:
…ding mail

Resolves #6902


app/models/appeal.rb
```diff
@@ -66,6 +66,8 @@ def veteran_gender
     veteran && veteran.sex
   end
 
+  delegate :name_suffix, to: :veteran, prefix: true
```
amprokop: why are we delegating this method but not the others?

aroltsch: Good point, I should change that.

PR #6922, opened 2018-09-12
Add seed data for BVA Dispatch tasks

lowellrex:
Connects #6886.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -44,6 +44,12 @@ class SeedDB
 
     Functions.grant!("System Admin", users: User.all.pluck(:css_id))
 
+    create_vso_user
+    create_org_queue_user
+    create_bva_dispatch_user_with_tasks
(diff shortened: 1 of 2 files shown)


db/seeds.rb
```diff
@@ -44,19 +44,50 @@ def create_users
...
 
+  def create_org_queue_user
     q = User.create!(station_id: 101, css_id: "ORG_QUEUE_USER", full_name: "Org Q User")
     FeatureToggle.enable!(:org_queue_translation, users: [q.css_id])
     FeatureToggle.enable!(:organization_queue, users: [q.css_id])
   end
 
+  def create_bva_dispatch_user_with_tasks
+    u = User.create(
+      css_id: "BVA_DISPATCHER",
```
aroltsch: We should probably use a user from Facols that has BVA Dispatch role.

PR #6925, opened 2018-09-12
HearingDay Data Model Implementation

### LINKED ISSUE #6741: Data Structures | Caseflow parent records and hearing days
Meet with Oscar, Sharon, and Teja to confirm data structures, and then create them in Caseflow.

oscarAtNava:
Resolves #6741 

### Description
Implement the HearingDay model and the hearing_days Postgres table. The model needs to be compatible with the HearingDay behavior prior to this issue where data is stored into VACOLS. An implementation date is defined as a constant to control the move from VACOLS to Caseflow data storage.

### Acceptance Criteria
- [ ] Code compiles correctly
- [ ] Defined tests pass

### Testing Plan
1. Build an RO Schedule
2. Add judge assignments to the schedule
3. Check both FACOLS and Caseflow databases depending on schedule dates.

Do the above for both dates before April 1, 2019 and after, as well as combination. All work.


db/schema.rb
```diff
@@ -382,7 +394,6 @@
     t.date "receipt_date"
     t.boolean "informal_conference"
     t.boolean "same_office"
-    t.datetime "established_at"
```
sharonwarner: Do you know why this field was deleted? It seems like an unrelated change.

db/schema.rb
```diff
@@ -613,7 +624,6 @@
   create_table "supplemental_claims", force: :cascade do |t|
     t.string "veteran_file_number", null: false
     t.date "receipt_date"
-    t.datetime "established_at"
```
sharonwarner: Same with this one.

db/schema.rb
oscarAtNava: Look at migration file ./db/migrate/20180907003818_add_claim_establishment_idempotency_fields.rb

leikkisa: This is correct, it accidentally got added back into schema.rb

db/schema.rb
leikkisa: Same as above

Issue #6205, opened 2018-07-12
AMA Intake | Set the EP benefit type correctly


shanear:
Right now, we always set `benefitTypeCd` to `1` which represents a "C&P Live" claim. However, if the veteran is deceased we need to create a "C&P Death" claim represented by a `benefitTypeCd` of `2`.

## Acceptance criteria
- Check if the Veteran is deceased (somewhere in BGS)
- When processing an AMA HLR or SC, and the veteran is deceased (BGS), the `benefitTypeCd` for the EP should be set to `2` for all EPs created

## Technical notes
- Here is the value that should be passed to connect_vbms to change the benefitTypeCd: https://github.com/department-of-veterans-affairs/connect_vbms/blob/master/lib/vbms/requests/establish_claim.rb#L74

shanear:
Nate said this is set in the UI by users. Would AMO want us to set this too manually?

abbyraskinUSDS:
Erik understood the question, but did not want to make a unilateral decision without programs folk on the line. So, I'll send a follow-up email.

kahlouie:
Still waiting on AMO. Big HVAC hearing is helping to cause delays.

abbyraskinUSDS:
VBMS sets it as death if the veteran is not alive. VBMS defaults payee code to spouse instead of Veteran in this case. Date of death. If deceased death, the benefit type defaults to death. no reason to change default. So we should probably check if Veteran is deceased and default to 2 if so - and remove Veteran from payee code options?

jiahuang:
@pkarman `date_of_death` was just added to the veteran model: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/veteran.rb#L12

According to the commit, looks like its nil for alive and has a value for deceased.

pkarman:
Answered via slack: yes, regardless of form, set `benefitTypeCd`

shanear:
I need to get some test data from Colin for a deceased vet before i can validate this

shanear:
Sent an email to Colin. Waiting on that

shanear:
Looks like this failed:

```
[...]

PR #6927, opened 2018-09-12
VBMS benefit_type_code based on date_of_death

pkarman:
connects #6205 

### Description
Rather than hardcoding the `benefit_type_code` and assuming all requests refer to living veterans, check for the `Veteran.date_of_death` and set the `benefit_type_code` appropriately.

### Acceptance Criteria
- [ ] Code compiles correctly
- [ ] Presence of `Veteran.date_of_death` triggers `benefit_type_code == 2`

### Testing Plan
1. Verify tests pass.

### DIFF
--- a/app/models/end_product_establishment.rb
+++ b/app/models/end_product_establishment.rb
@@ -166,9 +166,13 @@ class EndProductEstablishment < ApplicationRecord
   end
 
   def benefit_type_code
-    @benefit_type_code ||= @veteran.date_of_death.nil?
-      ? EndProduct::BENEFIT_TYPE_CODE_LIVE
-      : EndProduct::BENEFIT_TYPE_CODE_DEATH
+    @benefit_type_code ||= begin
+      if veteran.date_of_death.nil?
+        EndProduct::BENEFIT_TYPE_CODE_LIVE
+      else
+        EndProduct::BENEFIT_TYPE_CODE_DEATH
+      end
+    end
   end
 
   def establish_claim_in_vbms(end_product)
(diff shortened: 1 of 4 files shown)


app/models/end_product_establishment.rb
```diff
@@ -165,6 +165,16 @@ def veteran
     @veteran ||= Veteran.find_or_create_by_file_number(veteran_file_number)
   end
 
+  def benefit_type_code
+    @benefit_type_code ||= begin
```
shanear: might be more terse to use the ternary operator. Its a pattern we use a lot through out the app

```
veteran.date_of_death.nil? ? EndProduct::BENEFIT_TYPE_CODE_LIVE :EndProduct::BENEFIT_TYPE_CODE_DEATH
```

pkarman: ha! I had the ternary originally (58d0e0ed2adbc1ebfe91be1b277368f8572d8bee ) and then was afraid I was going to trigger the rubocop linelength rule. I've also been told that the ternary is less idiomatic Rubyish. But if this code base uses it, happy to restore it.

pkarman: yeah, the ternary breaks the linelength rule:
```
app/models/end_product_establishment.rb:169:121: C: Metrics/LineLength: Line is too long. [128/120]
    @benefit_type_code ||= veteran.date_of_death.nil? ? EndProduct::BENEFIT_TYPE_CODE_LIVE : EndProduct::BENEFIT_TYPE_CODE_DEATH
                                                                                                                        ^^^^^^^^
```
Let's keep the if/else, if you don't mind. I find it easier to ready than the multi-line ternary, and the multi-line ternary loses some of the advantage of terse expression.

PR #6932, opened 2018-09-12
Fix case search access control

### LINKED ISSUE #6933: Non-queue users can't access case details page
Folks with access to Caseflow but without queue access (for example, somebody with access only to the case search home page) cannot view the case details page after searching (or navigating there directly).

## Acceptance criteria
* Somebody with access to the case search page should be able to see the case details page

lowellrex:
Resolves #6933. Makes it possible for non-queue users to access case details pages.

Additionally, this PR removes the concept of a "queue user". If somebody who would previously have been prevented from visiting `/queue` navigates to `/queue` they will now see an empty page with a case search bar instead of being redirected to `/unauthorized`. This potentially allows more people to access case search than were previously able to, but I think that group of people is small and if we can determine who they are we should explicitly limit their access to which cases they can search for (as we do with VSO employees).

Below is the list of people we were explicitly allowing access to `/queue` and, by extension, case search:
* System admins
* Organizational queue users
* VSO employees (explicitly prevented from seeing cases that they do not have BGS permissions for)
* Attorneys
* Judges
* Co-located administrative staff
* Anyone with the `case_search_home_page` function (no access to queue before)

## Test plan
* Log in as `CASE_SEARCHER_ONLY`
* Search for a Veteran's case (e.g. 239120550)
* Click on the link in the results page
* See the page successfully load


app/controllers/reader/appeal_controller.rb
```diff
@@ -2,7 +2,6 @@ class Reader::AppealController < Reader::ApplicationController
   def index
     respond_to do |format|
       format.html do
-        return redirect_to "/queue" if can_access_queue?
```
lowellrex: Just needed to pick one since the `can_access_queue?` is going away. Maybe there right thing to do here is to always redirect to `/queue`?

app/controllers/tasks_controller.rb
```diff
@@ -106,9 +105,7 @@ def queue_class
   end
 
   def user_role
-    return params[:role].downcase unless params[:role].to_s.empty?
-
-    current_user.organization_queue_user? ? "generic" : nil
+    params[:role].to_s.empty? ? "generic" : params[:role].downcase
```
lowellrex: Default to "generic" role unless we explicitly pass a role. This change, along with the removal of the `verify_queue_access` hook, enable access to the case details page for `case_search_home_page` users.

app/views/queue/show.html.erb
```diff
@@ -1,17 +0,0 @@
-<% content_for :full_page_content do %>
```
lowellrex: Removed this file because I am pretty sure it was not being used any more.

lowellrex: Turns out, I was wrong... This file is back now.

app/controllers/reader/appeal_controller.rb
```diff
@@ -2,8 +2,8 @@ class Reader::AppealController < Reader::ApplicationController
   def index
     respond_to do |format|
       format.html do
+        return redirect_to "/" if FeatureToggle.enabled?(:case_search_home_page, user: current_user)
```
mdbenjam: I don't think we should make this redirect. The logical page here is `/queue` never the search page.

PR #6935, opened 2018-09-12
Adding address to IDT

### LINKED ISSUE #6701: Allow dispatch users to receive all claimant and POA addresses from BGS for AMA appeals
# AC
- When requesting appeal details for AMA appeals from /idt/api/v1/appeals/:appeal_id, Dispatch users should be able to receive addresses along with name information for all appellants (including contesting claimants) and POAs for all appellants
- This information should be fetched from BGS
- For now, I'm thinking we should *not* rely on cached information for the addresses — we want the most recent possible information 
- Attorney and Judge users should not receive address information.
- If BGS returns an error, we should return an error to the IDT API that indicates that we couldn't fetch address information

mdbenjam:
Resolves #6701

### Description
This creates a new BgsAddressService to improve our ability to lazily load addresses (instead of bulk querying all the BGS endpoints we need, we now only query for address when we need it). It also adds a new parameter to our IDT appeal details serializer that determines whether or not we need to pull addresses.

### Testing Plan
1. Automated tests cover the address and no address scenarios.


app/models/claimant.rb
```diff
@@ -26,11 +25,18 @@ def self.bgs
...
     name_info = self.class.bgs.fetch_person_info(participant_id)
 
-    bgs_record.merge(general_info).merge(name_info)
+    general_info.merge(name_info)
+  end
+
+  private
+
+  def bgs_address_service
+    @bgs_address_service ||= BgsAddressService.new(participant_id: participant_id)
```
mdbenjam: Splitting out this service so that we can use AssociatedBgsRecord for addresses without having to always query for it when calling `fetch_bgs_record`

Issue #6191, opened 2018-07-11
Non-compensation LOBs | Track line of business during intake


abbyraskinUSDS:
When processing a HLR or SC form at the ARC, I want to select the business line, so the data can be tracked in Caseflow and routed to the correct office for processing. This input also will tell the system which data to pre-populate later in the intake flow (e.g., the issue categories), to make it easier for claims assistants to enter issues.

Screenshot from HLR form:
![image](https://user-images.githubusercontent.com/4042181/43153288-7d56fcac-8f36-11e8-9b74-17ed0fece0b2.png)

## Acceptance Criteria
- Benefit types are presented as radio buttons (only one can be selected at a time). 
- Benefit type question is listed first on the Review page.
- This is a required field (but no "required" label is needed. 

## Mocks
<img width="559" alt="screen shot 2018-07-31 at 11 09 45 am" src="https://user-images.githubusercontent.com/11219417/43468540-44bd656c-94b2-11e8-8c5f-4d9de6244c42.png">

## Copy
Field label: What is the Benefit Type?
Drop down contents:
- Compensation
- Pension
- Fiduciary
- Insurance
- Education
- Vocational Rehab. & Employment 
- Loan Guaranty
- VHA
- NCA

## Validation Instructions
Flows: AMA Higher Level Review AND Supplemental Claim
Test file number: 123458311

Fill out the review page, check that the Benefit Claim radio buttons match design standards.  First try going to the next page without selecting an option. Then selection and option and proceed to the finish step.

If you are pairing with an engineer, they can check that it is working without you having to establishing EPs.  If not, finish the intake, and have an engineer check that the benefit_type is set in the supplemental_claims and higher_level_reviews tables for the intakes you did.

suzchap:
<img width="803" alt="screen shot 2018-07-25 at 1 36 34 pm" src="https://user-images.githubusercontent.com/11219417/43217502-d0304e54-900f-11e8-89c6-83a8a881a3f5.png">

shanear:
![screen shot 2018-09-24 at 2 05 48 pm](https://user-images.githubusercontent.com/1596259/45969874-09674d80-c003-11e8-85c4-1fa2038c7d0b.png)

lookin' good.

PR #6945, opened 2018-09-12
Add ability to select non-compensation LOBs for benefit_type

leikkisa:
connects #6191


client/app/intake/constants.js
```diff
@@ -120,6 +121,27 @@ export const CANCELLATION_REASONS = {
   }
 };
 
+export const BENEFIT_TYPES = [
```
shanear: we should probably put these in a place where they are accessible on the front and the back end. But i don't think we need to right now

PR #6950, opened 2018-09-13
Schedule Prevent 5 digit years

### LINKED ISSUE #6740: Date Selector | Prevent 5 digit years
The date selector should not allow 5 digit years.

Sjones352:
Resolves #6740

### Description
This PR limits 5 digit years

### Acceptance Criteria
- [ ] The date selector should not allow 5 digit years.

### Testing Plan
1. Go to schedule and enter years in the date input area.


client/app/hearingSchedule/components/BuildScheduleUpload.jsx
```diff
@@ -34,6 +34,7 @@ const inlineFormStyling = css({
   }
 });
 
+
```
sharonwarner: Let's recheck out this file so we don't commit any unnecessary changes.

client/app/components/DateSelector.jsx
```diff
@@ -51,6 +51,7 @@ export default class DateSelector extends React.Component {
       type,
       validationError,
       value,
+      max,
```
sharonwarner: This variable is unused because we hardcode max to equal '2100-12-26'. I'm fine with either adding it as a prop, or hardcoding the value below, but we should pick one or the other.

client/app/components/DateSelector.jsx
```diff
@@ -66,6 +67,7 @@ export default class DateSelector extends React.Component {
       placeholder={DEFAULT_TEXT}
       required={required}
       {...passthroughProps}
+      max="2100-12-26"
```
sharonwarner: I think it makes more sense to set this to 9999-12-31, but let's confirm with @Chingujo.

PR #6957, opened 2018-09-13
Rename after_certification to post_aoj for remand reasons

aroltsch:


client/app/queue/components/IssueRemandReasonsOptions.jsx
```diff
@@ -219,42 +253,57 @@ class IssueRemandReasonsOptions extends React.PureComponent {
         </div>
       }
 
-      <div {...flexContainer}>
+      {appeal.docketName === 'legacy' ? <div {...flexContainer}>
```
mdbenjam: I think using a function instead of giant ternary might be cleaner

client/app/queue/components/IssueRemandReasonsOptions.jsx
```diff
@@ -66,17 +102,13 @@ class IssueRemandReasonsOptions extends React.PureComponent {
...
+  validate = () => {
     const chosenOptions = this.getChosenOptions();
-    const chosenOptionsWithCertification = _.filter(chosenOptions, (opt) => !_.isNull(opt.after_certification));
 
-    return chosenOptions.length === chosenOptionsWithCertification.length;
-  };
-
-  validate = () => this.getChosenOptions().length >= 1 &&
-    this.validateChosenOptionsHaveCertification();
+    return chosenOptions.length >= 1 && _.every(chosenOptions, (opt) => !_.isNull(opt.post_aoj));
```
mdbenjam: Could you just explain this line? Confused by what this is checking. Can things only be pre aoj?

dannysellers: The default [`post_aoj`](https://github.com/department-of-veterans-affairs/caseflow/pull/6957/files#diff-efb80ef80a79fc5dae64a4f198af5d0eR91) value is `null`, so that neither radio button is selected by default. This just confirms that this attr has a value

Issue #6961, opened 2018-09-13
Script is trying to reprocess ramp refilings for an appeal (which don't actually need to be reprocessed)


shanear:
We aren't setting `establishment_processed_at` for ramp refilings that are appeals so they are registering as `need_to_reprocess`, which is incorrect and causing the error below. We should set them.

https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2432/

PR #6962, opened 2018-09-13
Fix issue with the RampRefiling process script

shanear:
connects #6961


app/models/ramp_refiling_intake.rb
```diff
@@ -64,7 +64,11 @@ def ui_hash(ama_enabled)
   private
 
   def create_end_product_and_contentions
```
jiahuang: Should this function name be changed since the else statement doesn't do a create?

shanear: yeah, but it sometimes didn't do a create before this change, so I'll leave that for another PR.

Issue #6964, opened 2018-09-13
AMA | Evaluate Decision | Show the date that the case was assigned to an attorney


evankroske:
<img width="687" alt="screen shot 2018-09-13 at 11 39 24 am" src="https://user-images.githubusercontent.com/357369/45508123-04341400-b762-11e8-977a-481242e4a42b.png">

Logic for legacy cases: https://github.com/department-of-veterans-affairs/caseflow/blob/6b02606c71ca03b7f175c2165600c4e82f734a50/app/models/legacy_tasks/judge_legacy_task.rb#L8

PR #6969, opened 2018-09-13
AMA | Evaluate Decision | Show the date that the case was assigned to an attorney

evankroske:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/6964

### Description
Shows the date that the judge asked the attorney to write a decision on the Evaluate Decision page for AMA cases.

### Testing Plan
- [x] `rake db:setup db:seed`
- [x] Become Bvaaabshire.
- [x] /queue
- [x] Open an AMA case.
- [x] Select Ready for Dispatch.
- [x] Click Continue.
- [x] Set both dispositions to Allowed.
- [x] Click Continue.
- [x] "Date assigned to attorney" should be the date at which you ran the rake command.


spec/factories/task.rb
```diff
@@ -53,7 +53,7 @@
       appeal { create(:appeal) }
     end
 
-    factory :ama_judge_task do
+    factory :ama_judge_task, class: JudgeTask do
       type "JudgeTask"
```
aroltsch: Oh cool! Can we remove the `type` from line 57 then?

app/models/tasks/judge_task.rb
```diff
@@ -9,4 +9,9 @@ def when_child_task_completed
     update!(action: :review)
     super
   end
+
+  def previous_task
+    fail Caseflow::Error::TooManyChildTasks, task_id: id if children.length > 1
+    children[0]
```
aroltsch: A judge task might have another type of child. I think we talked about having a judge to assign directly to the hearing schedule team or similar. We should filter by the children types here. 
```children.where(type: "AttorneyTask")```

spec/factories/task.rb
evankroske: No, I don't think that FactoryBot understands STI. I think that we need type and class

app/models/tasks/judge_task.rb
evankroske: OK, I just made this change

PR #6980, opened 2018-09-14
Add endpoint for completing BvaDispatchTasks through the IDT API

### LINKED ISSUE #6888: Allow dispatch users to complete their tasks using the IDT API
At the end of the IDT for BVA Dispatch, the IDT should make an call to Caseflow to complete the BVA Dispatch task, so we can run a query every week to determine which cases to send to ARC.

`/idt/api/v1/appeals/:appeal_id/outcode`?

# AC
- Create a way for the IDT to mark a BVA Dispatch task as complete

lowellrex:
Resolves #6888. Adds `/idt/api/v1/appeals/:appeal_id/outcode` endpoint to allow folks to complete `BvaDispatchTask`s through the IDT API.

### DIFF
--- a/app/controllers/idt/api/v1/appeals_controller.rb
+++ b/app/controllers/idt/api/v1/appeals_controller.rb
@@ -2,6 +2,12 @@ class Idt::Api::V1::AppealsController < Idt::Api::V1::BaseController
   protect_from_forgery with: :exception
   before_action :verify_access
 
+  rescue_from StandardError do |e|
+    fail e unless e.class.method_defined?(:serialize_response)
+    Raven.capture_exception(e)
+    render(e.serialize_response)
+  end
+
   def list
     appeals = file_number ? appeals_by_file_number : appeals_assigned_to_user
 
@@ -10,6 +16,11 @@ class Idt::Api::V1::AppealsController < Idt::Api::V1::BaseController
 
   def details
     return render json: { message: "Appeal not found" }, status: 404 unless appeal
+    render json: json_appeal_details
+  end
+
+  def outcode
+    BvaDispatchTask.outcode(appeal, user)
     render json: json_appeal_details
   end
 
--- a/app/models/tasks/bva_dispatch_task.rb
+++ b/app/models/tasks/bva_dispatch_task.rb
@@ -22,7 +22,7 @@ class BvaDispatchTask < GenericTask
     private
 
     def list_of_assignees
-      @list_of_assignees || Constants::BvaDispatchTeams::USERS[Rails.current_env]
+      Constants::BvaDispatchTeams::USERS[Rails.current_env]
     end
   end
 end
--- a/app/models/tasks/bva_dispatch_task.rb
+++ b/app/models/tasks/bva_dispatch_task.rb
@@ -11,10 +11,18 @@ class BvaDispatchTask < GenericTask
       )
     end
 
+    def outcode(appeal, user)
+      tasks = where(appeal: appeal, assigned_to: user)
+      if tasks.count != 1
+        fail Caseflow::Error::BvaDispatchTaskCountMismatch, appeal_id: appeal.id, user_id: user.id, tasks: tasks
+      end
+      tasks[0].mark_as_complete!
+    end
+
     private
 
     def list_of_assignees
-      Constants::BvaDispatchTeams::USERS[Rails.current_env]
+      @list_of_assignees || Constants::BvaDispatchTeams::USERS[Rails.current_env]
     end
   end
 end
--- a/spec/controllers/idt/api/appeals_controller_spec.rb
+++ b/spec/controllers/idt/api/appeals_controller_spec.rb
@@ -249,4 +249,56 @@ RSpec.describe Idt::Api::V1::AppealsController, type: :controller do
       end
     end
   end
+
+  describe "POST /idt/api/v1/appeals/:appeal_id/outcode" do
+    let(:user) { FactoryBot.create(:user) }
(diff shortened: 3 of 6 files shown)


app/controllers/idt/api/v1/appeals_controller.rb
```diff
@@ -13,6 +19,11 @@ def details
     render json: json_appeal_details
   end
 
+  def outcode
+    BvaDispatchTask.outcode(appeal, user)
+    render json: json_appeal_details
```
lowellrex: Not entirely sure what to return in the success case here.

amprokop: This is probably fine.

app/models/tasks/bva_dispatch_task.rb
```diff
@@ -11,6 +11,14 @@ def create_and_assign(root_task)
       )
     end
 
+    def outcode(appeal, user)
```
amprokop: 👍

app/controllers/idt/api/v1/appeals_controller.rb
```diff
@@ -13,6 +19,11 @@ def details
     render json: json_appeal_details
   end
 
+  def outcode
```
amprokop: It might be good to add a check to make sure it's a BVA Dispatch user hitting this route. What do you think?

spec/controllers/idt/api/appeals_controller_spec.rb
```diff
@@ -249,4 +249,56 @@
...
   end
+
+  describe "POST /idt/api/v1/appeals/:appeal_id/outcode" do
+    let(:user) { FactoryBot.create(:user) }
+    let!(:vacols_atty) { FactoryBot.create(:staff, :attorney_role, sdomainid: user.css_id) }
+    let(:root_task) { FactoryBot.create(:root_task) }
+    let(:params) { { appeal_id: root_task.appeal.external_id } }
+
+    before do
+      allow(BvaDispatchTask).to receive(:list_of_assignees).and_return([user.css_id])
```
amprokop: Nice 👍

spec/controllers/idt/api/appeals_controller_spec.rb
```diff
@@ -249,4 +249,56 @@
...
+
+      key, t = Idt::Token.generate_one_time_key_and_proposed_token
+      Idt::Token.activate_proposed_token(key, user.css_id)
+      request.headers["TOKEN"] = t
+    end
+
+    context "when single BvaDispatchTask exists for user and appeal combination" do
```
amprokop: I'm curious — can you think of a situation where more than one task would exist? Or is this more of a hypothetical scenario?

app/controllers/idt/api/v1/appeals_controller.rb
amprokop: Ah, nevermind. I see that we'd only complete a task if it's assigned to the user. Sounds good.

app/models/tasks/bva_dispatch_task.rb
```diff
@@ -11,6 +11,14 @@ def create_and_assign(root_task)
...
+      tasks = where(appeal: appeal, assigned_to: user)
```
amprokop: Should we also check if it's already complete? Otherwise, could a case be double-outcoded? That could lead to bad things if so.

spec/controllers/idt/api/appeals_controller_spec.rb
```diff
@@ -249,4 +249,56 @@
...
+        expect(response_detail).to eq("Expected 1 BvaDispatchTask received #{task_count} tasks for appeal "\
+                                      "#{root_task.appeal.id}, user #{user.id}")
+      end
+    end
+
+    context "when no BvaDispatchTasks exists for user and appeal combination" do
+      let(:task_count) { 0 }
+      it "should throw an error" do
+        post :outcode, params: params
+        expect(response.status).to eq(500)
```
amprokop: nit: Wondering if a 500 is actually the best code to return here. This seems like a pretty known situation — if the IDT tries to outcode the wrong appeal or outcode an appeal that has already been outcoded. This feels more like a "400 bad request" kind of situation to me.

spec/controllers/idt/api/appeals_controller_spec.rb
lowellrex: Mostly hypothetical, but it is possible that the frontend could send the request to the backend more than once (a feisty user double clicks a submit decision button, or maybe two tabs open to the same page and the submit decision button is pressed twice).

If we start seeing this at all often, we should probably just send this information to DataDog (like we do in the standard [AppealsController](https://github.com/department-of-veterans-affairs/caseflow/blob/4f41607e9496c7730607698f21c1d29960110a8c/app/controllers/appeals_controller.rb#L124)) and just mark them all complete since it probably doesn't matter. I'd like to alert us to these situations if they do happen though so we at least have a chance to investigate the circumstances that led to it happening.

app/models/tasks/bva_dispatch_task.rb
lowellrex: I'm up for checking if it is already complete, but I'm not entirely sure what the harm would be for attempting to complete an already completed BvaDispatchTask (which is all that happens in [`Task.mark_as_complete!()`](https://github.com/department-of-veterans-affairs/caseflow/blob/4f41607e9496c7730607698f21c1d29960110a8c/app/models/task.rb#L68)).

spec/controllers/idt/api/appeals_controller_spec.rb
lowellrex: Smart. Will change!

Issue #6168, opened 2018-07-10
AMA Decisions | Create a supplemental claim on a DTA error for a decided higher level review


shanear:
When a higher level review is decided, if the disposition of that issue is in the list below, we should create a follow up supplemental claim with those issues.

DTA Error – PMRs
DTA Error – Fed Recs
DTA Error – Other Recs
DTA Error – Exam/MO

## Acceptance Criteria
- When syncing the dispositions of request issues of a higher level review, if any of the request issues have dispositions of DTA Error (see the list above), Caseflow should automatically create a follow on supplemental claim with request issues linked to those issues.
- If the supplemental claim is the result of DTA error the EP created should have a special code (040HDENR for non rating EPs, 040 HDER  for rating EPs)

## Technical notes
- We will have to determine how to link request issues in the database. It would be good to agree on this structure before we begin coding this task.

blocks #6169

abbyraskinUSDS:
Here is the list of AMA end products. It looks like the options for a supplemental claim stemming from a Board remand decision are:
- 040BDE
- 040BDEIMO

![eps for ama](https://user-images.githubusercontent.com/4042181/45230581-e3b21880-b28e-11e8-8ab9-4c680de023fd.png)

shanear:
This didn't work 😢 

PR incoming.

shanear:
<img width="1176" alt="screen shot 2018-10-10 at 12 56 41 pm" src="https://user-images.githubusercontent.com/1596259/46752732-f81d7280-cc8b-11e8-82d8-f955fafb769c.png">
woohoo!

Checked everything linked up correctly in the DB as well

PR #6992, opened 2018-09-14
Create supplemental review on DTA error

jiahuang:
Connects #6168

### Description
On sync with contentions, if it is a higher level review, create a new supplemental review with associated request issues.


app/models/higher_level_review.rb
```diff
@@ -46,8 +48,39 @@ def valid_modifiers
...
+    return unless dta_issues
+    rating_code_type = dta_issues.first.rated? ? :rating : :nonrating
+    sc = create_supplemental_claim
+    ep = sc.new_end_product_establishment(DTA_SUPPLEMENTAL_CLAIM_CODES[rating_code_type])
+    ep.perform!
+
+    # create duplicate dta issues that link to original
+    new_issues = dta_issues.map do |dta_issue|
+      new_issue = dta_issue.dup
+      new_issue.assign_attributes(
```
jiahuang: are there other attributes that new dta issues should set?

app/models/supplemental_claim.rb
```diff
@@ -36,8 +36,6 @@ def valid_modifiers
     END_PRODUCT_MODIFIERS
   end
 
-  private
-
   def new_end_product_establishment(ep_code)
```
jiahuang: Note: changed this function to be public so it can be used in higher level reviews

db/schema.rb
```diff
@@ -550,6 +550,7 @@
     t.integer "end_product_establishment_id"
     t.datetime "removed_at"
     t.datetime "rating_issue_associated_at"
+    t.integer "dta_issue_id"
```
jiahuang: is "dta_issue_id" the right name for this column?

spec/models/claim_review_spec.rb
```diff
@@ -297,37 +297,154 @@
...
+      let(:issues) { [rating_request_issue, second_rating_request_issue, non_rating_request_issue] }
+
+      let(:rating_contention) do
         Generators::Contention.build(
           claim_id: end_product_establishment.reference_id,
-          text: "hello",
-          disposition: "Granted"
-        ),
+          text: "decision text",
+          disposition: "DTA Error – PMRs"
```
jiahuang: there are 4 different DTA error dispositions but we only test 3 in our tests, should the 4th also be tested?

spec/models/claim_review_spec.rb
```diff
@@ -297,37 +297,154 @@
...
+        claim_review.create_issues!(issues)
+      end
+
+      def verify_followup_request_issue(end_product_id, orig_request_issue, contention)
+        follow_up_issue = RequestIssue.find_by(
+          review_request_id: end_product_id,
+          dta_issue_id: orig_request_issue.id
+        )
+
+        expect(follow_up_issue).to have_attributes(
```
jiahuang: are there further tests we should do on the resulting dta issue?

db/schema.rb
pkarman: the naming conventions I've seen here are `*_reference_id` for points to external systems, and `*_id` for internal FKs. Per Rails conventions, I would expect there to be a `DTAIssue` model somewhere if I saw a column with this name.

app/models/claim_review.rb
```diff
@@ -43,6 +43,7 @@ def invalid_modifiers
   def on_sync(end_product_establishment)
     if end_product_establishment.status_cleared?
       sync_dispositions(end_product_establishment.reference_id)
+      create_dta_supplemental_claim(end_product_establishment) if self.class == HigherLevelReview
```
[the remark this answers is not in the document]
shanear: I have acronym guilt and am wondering if we should spell out `duty_to_assist` in the code instead of `dta`. Not sure, but just something to think about.

app/models/claim_review.rb
```diff
@@ -94,10 +95,18 @@ def rated_issue_contention_map(request_issues_to_associate)
 
   def sync_dispositions(reference_id)
     fetch_dispositions_from_vbms(reference_id).each do |disposition|
-      matching_request_issue(disposition[:contention_id]).update!(disposition: disposition[:disposition])
+      request_issue = matching_request_issue(disposition[:contention_id])
+      request_issue.update!(disposition: disposition[:disposition])
+      if (self.class == HigherLevelReview) && dta_errors.include?(disposition[:disposition])
```
shanear: Additionally, we can probably lazily calculate the dta_issues instead of having to work this out in the loop

```
def dta_issues
  @dta_issues ||= request_issues.select(&:needs_dta_follow_up?)
end
```

```
class RequestIssue
  has_one :follow_up_request_issue, as: :parent_request_issue

  def needs_dta_follow_up?
    !follow_up_request_issue && HigherLevelReview::DTA_ERRORS.include?(disposition)
  end
end
```

app/models/end_product.rb
```diff
@@ -54,7 +59,7 @@ class EndProduct
     "070RMBVAGPMC" => "PMC Remand with BVA Grant"
   }.freeze
 
-  CODES = DISPATCH_CODES.merge(RAMP_CODES).merge(AMA_CODES)
+  CODES = DISPATCH_CODES.merge(RAMP_CODES).merge(AMA_CODES).merge(DTA_CODES)
```
shanear: lol this is getting more fun 👾

app/models/higher_level_review.rb
```diff
@@ -46,8 +48,39 @@ def valid_modifiers
     END_PRODUCT_MODIFIERS
   end
 
+  def dta_errors
+    DTA_ERRORS
+  end
+
+  def create_dta_supplemental_claim(end_product_establishment)
```
shanear: I like your style of changing function names if it doesnt always happen. Might be better named as `create_dta_supplemental_claim_if_necessary(end_product_establishment)`

app/models/higher_level_review.rb
```diff
@@ -46,8 +48,39 @@ def valid_modifiers
...
+    return unless dta_issues
```
shanear: `[]` resolves to truthy. So this wouldn't return if `dta_issues == []` 

```
irb(main):002:0> puts "HI" if []
HI
```

should be

```
return if dta_issues.empty?
```

Additionally, we should have a test testing this functionality. A test would have caught this issue.

db/schema.rb
shanear: Additionally, this column will likely be used to connect issues that were re-reviewed. I'm leaning towards calling this `parent_request_issue_id` or `parent_issue_id`

app/models/claim_review.rb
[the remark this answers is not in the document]
jiahuang: Oh neat, didn't realize that distinction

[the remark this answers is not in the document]
pkarman: I should also say, it's a little odd to me to have the parent class knowing about the subclass. would it make sense to override `on_sync` in the one class that behaves differently?

app/models/higher_level_review.rb
```diff
@@ -46,8 +48,39 @@ def valid_modifiers
...
+    rating_code_type = dta_issues.first.rated? ? :rating : :nonrating
+    sc = create_supplemental_claim
+    ep = sc.new_end_product_establishment(DTA_SUPPLEMENTAL_CLAIM_CODES[rating_code_type])
+    ep.perform!
```
shanear: A lot of this logic feels more like the responsibility of the `SupplementalClaim` than the `higher_level_review`. It sort of goes against the idiomatic principle of "tell dont ask".

There are also a lot of methods in SupplementalClaim that can be used here to cut down on the duplicate logic.

Instead i would think it should be something like:

```
  def dta_supplemental_claim
    @dta_supplemental_claim ||= SupplementalClaim.create!(
      receipt_date: Time.zone.now.to_date
      dta_error_follow_up: true
   )
  end

  def def create_dta_supplemental_claim(end_product_establishment)
    return if dta_issues.empty?

    transaction do
      dta_supplemental_claim.create_issues!(build_follow_up_dta_issues)
      dta_supplemental_claim.process_end_products_and_contentions!
    end
  end

  def build_dta_follow_up_issues
     dta_issues.map do |dta_issue|
       # ...
     end
  end
```
Then the `SupplementalClaim` can be responsible for determining it's rating code depending on if `dta_error_follow_up` is true or not.

See: https://robots.thoughtbot.com/tell-dont-ask for more context

db/migrate/20180914120847_add_dta_id_to_request_issue.rb
```diff
@@ -0,0 +1,5 @@
+class AddDtaIdToRequestIssue < ActiveRecord::Migration[5.1]
+  def change
+    add_column :request_issues, :dta_issue_id, :integer, :null => true
```
shanear: small nit: `:null => true` is the default and superfluous

app/models/supplemental_claim.rb
shanear: see my comments above on why we can move this back to being private

spec/models/claim_review_spec.rb
```diff
@@ -297,37 +297,162 @@
...
+          it "creates a supplemental claim for rated issues" do
+            claim_review.on_sync(end_product_establishment)
+
+            supplemental_claim = SupplementalClaim.find_by(
+              veteran_file_number: claim_review.veteran_file_number,
+              receipt_date: Time.zone.now.to_date)
+            expect(supplemental_claim).to_not be_nil
+
+            end_product_establishment = EndProductEstablishment.find_by(
+              code: "040HDER",
```
leikkisa: This looks like if there are non-rated issues, it should still only be created one rating End Product, and that end_product will have the non-rating issue on it.  This doesn't seem right to me, why wouldn't it make the rating issues on a rated issue EP and non-rated issues on a non-rated EP?

app/models/higher_level_review.rb
```diff
@@ -46,8 +48,39 @@ def valid_modifiers
     END_PRODUCT_MODIFIERS
   end
 
+  def dta_errors
+    DTA_ERRORS
+  end
+
+  def create_duty_to_assist_supplemental_claim_if_needed(end_product_establishment)
+    return if dta_issues.empty?
+    rating_code_type = dta_issues.first.rated? ? :rating : :nonrating
```
leikkisa: This seems strange to me, the rating_code_type is determined by whether the first issue is rated or non-rated?

spec/models/claim_review_spec.rb
leikkisa: In this case I think it's okay since you're setting the error in your tests manually and they're all handled the same way.

app/models/higher_level_review.rb
jiahuang: from my understanding, the contentions we get back at once is either all non-rated or all rated so this is the same as checking for all issues since they are on different work queues and its unlikely that multiple would finish at the same time..? 🤷‍♀️  I'm not too sure on this, maybe @kahlouie can give some more info.

I suppose that we could do a sync that contains both rated & non-rated.

spec/models/claim_review_spec.rb
jiahuang: Not sure I understand your question. This test looks for code 040HDER (rated) vs 040HDENR (nonrated), so its checking that at least a different ep is created for rated and non-rated. Unless you mean that another test before this could be creating an ep that this matcher works for?

app/models/higher_level_review.rb
kahlouie: Correct. Since in the initial HLR we can make two end product establishments (one rating and one nonrating), the end products are worked independently of one another in NWQ. Our sync job is checking by end product establishments, so all the issues on a given end product establishment should be rating or non-rating, but not a mix.

app/models/supplemental_claim.rb
```diff
@@ -37,6 +37,12 @@ def valid_modifiers
     END_PRODUCT_MODIFIERS
   end
 
+  def create_end_product_establishment!(ep_code)
+    ep = new_end_product_establishment(ep_code)
+    ep.perform!
```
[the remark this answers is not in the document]
pkarman: if we call `perform!` here will that run afoul of `ClaimReview.process_end_product_establishments!` (the parent class method) that also will call `perform!` along with other things?

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 201809112051441) do
+ActiveRecord::Schema.define(version: 20180914201526) do
```
pkarman: this version timestamp seems like it is missing a digit?

spec/models/claim_review_spec.rb
```diff
@@ -297,37 +297,166 @@
...
+            description: contention.text,
+            contention_reference_id: contention.id
+          )
+        end
+      end
+
+      it "changes request issue dispositions" do
+        subject
+
+        expect(request_issues.first.reload.disposition).to eq("Granted")
```
pkarman: is the `first`/`last` order significant? or just a by-product of how the test fixtures were instantiated? I guess I'm asking: is the array important? Or simply a convenience in writing the test code?

spec/models/claim_review_spec.rb
```diff
@@ -297,37 +297,166 @@
...
+          receipt_date: Time.zone.now.to_date
         )
+        expect(supplemental_claim).to be_nil
       end
-    end
 
-    it "should add dispositions to the issues" do
-      subject
+      context "when it gets back dispositions with DTAs" do
+        def verify_followup_request_issue(end_product_id, orig_request_issue, contention)
```
pkarman: I like this re-use of testing code.

db/schema.rb
jiahuang: Hah, it was actually the other version file that had an extra digit: https://dsva.slack.com/archives/C2ZAMLK88/p1537294951000100

I'm going to wait on that to get merged in before I pull master back into this branch

spec/models/claim_review_spec.rb
```diff
@@ -297,37 +297,166 @@
...
+        context "for non-rated issues" do
+          before do
+            RequestIssue.find_by(
+              description: non_rating_contention.text
+            ).update!(
+              contention_reference_id: non_rating_contention.id
+            )
+          end
+
+          it "creates a supplemental claim for rated issues" do
```
leikkisa: This needs updated too

spec/models/claim_review_spec.rb
```diff
@@ -297,37 +297,166 @@
...
-      expect(request_issues.first.reload.disposition).to eq("Granted")
-      expect(request_issues.last.reload.disposition).to eq("Denied")
+        context "for rated issues" do
+          before do
+            [rating_contention, second_rating_contention].each do |contention|
+              RequestIssue.find_by(description: contention.text).update!(contention_reference_id: contention.id)
+            end
+          end
+
+          it "creates a supplemental claim for non rated issues" do
```
leikkisa: Don't forget to update these titles ;-)

app/models/supplemental_claim.rb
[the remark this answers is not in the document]
jiahuang: I don't think so. A supplemental claim created along this path should never be calling `process_end_product_establishments` since that calls `create_contentions_for_end_product_establishment` which creates request issues. DTA request issues are structured differently from the regular request issues.

🤔 maybe there should be some kind of enforcement on this so that certain methods for dta supp claims can't be called? Could do another subclass/override for DTA supplemental claim 🤷‍♂️

spec/models/claim_review_spec.rb
jiahuang: It's just the byproduct of the test fixture.

app/models/higher_level_review.rb
```diff
@@ -53,30 +52,33 @@ def dta_errors
...
+      # review request, or removed_at
+      RequestIssue.new(
+        review_request_id: dta_supplemental_claim.id,
+        review_request_type: "SupplementalClaim",
         parent_request_issue_id: dta_issue.id,
-        review_request_id: ep.id,
-        review_request_type: ep.source_type
+        rating_issue_reference_id: dta_issue.rating_issue_reference_id,
+        rating_issue_profile_date: dta_issue.rating_issue_profile_date,
+        contention_reference_id: dta_issue.contention_reference_id,
```
shanear: this would not be the same

app/models/higher_level_review.rb
```diff
@@ -53,30 +52,33 @@ def dta_errors
...
-  def build_follow_up_dta_issues(ep)
+  def build_follow_up_dta_issues
     dta_issues.map do |dta_issue|
-      new_issue = dta_issue.dup
-      new_issue.assign_attributes(
+      # do not copy over end product establishment id,
+      # review request, or removed_at
+      RequestIssue.new(
+        review_request_id: dta_supplemental_claim.id,
+        review_request_type: "SupplementalClaim",
```
shanear: you can do

```
review_request: dta_supplemental_claim
```

to make these two lines shorter

app/models/higher_level_review.rb
```diff
@@ -53,30 +52,33 @@ def dta_errors
...
         parent_request_issue_id: dta_issue.id,
-        review_request_id: ep.id,
-        review_request_type: ep.source_type
+        rating_issue_reference_id: dta_issue.rating_issue_reference_id,
+        rating_issue_profile_date: dta_issue.rating_issue_profile_date,
+        contention_reference_id: dta_issue.contention_reference_id,
+        description: dta_issue.description,
+        issue_category: dta_issue.issue_category,
+        decision_date: dta_issue.decision_date,
+        disposition: dta_issue.disposition,
```
shanear: the disposition would also not be the same, since there would be a new decision that isn't "DTA error"

app/models/supplemental_claim.rb
```diff
@@ -57,6 +52,10 @@ def new_end_product_establishment(ep_code)
   end
 
   def issue_code(rated)
-    rated ? END_PRODUCT_RATING_CODE : END_PRODUCT_NONRATING_CODE
+    if is_dta_error
+      rated ? DTA_SUPPLEMENTAL_CLAIM_CODES[:rating] : DTA_SUPPLEMENTAL_CLAIM_CODES[:nonrating]
```
shanear: i don't really care the format, but it'd be nice if we used the same constant format for all these codes. Either change these to something like `END_PRODUCT_CODES[:dta_rating]` on line 58 to `END_PRODUCT_CODES[:rating]` or change these to something like `END_PRODUCT_DTA_RATING_CODE`.

app/models/supplemental_claim.rb
```diff
@@ -4,6 +4,7 @@ class SupplementalClaim < ClaimReview
   END_PRODUCT_RATING_CODE = "040SCR".freeze
   END_PRODUCT_NONRATING_CODE = "040SCNR".freeze
   END_PRODUCT_MODIFIERS = %w[040 041 042 043 044 045 046 047 048 049].freeze
+  DTA_SUPPLEMENTAL_CLAIM_CODES = { rating: "040HDER", nonrating: "040HDENR" }.freeze
```
shanear: this is in the class `SupplementalClaim` so we don't need `SUPPLEMENTAL_CLAIM` in the constant name

app/models/higher_level_review.rb
```diff
@@ -47,8 +48,58 @@ def valid_modifiers
     END_PRODUCT_MODIFIERS
   end
 
+  def dta_errors
```
shanear: this could be a private method

app/models/higher_level_review.rb
```diff
@@ -47,8 +48,58 @@ def valid_modifiers
...
+    DTA_ERRORS
+  end
+
+  def create_dta_supplemental_claim
```
shanear: this could be a private method

app/models/higher_level_review.rb
```diff
@@ -47,8 +48,58 @@ def valid_modifiers
...
+        contention_reference_id: dta_issue.contention_reference_id,
+        description: dta_issue.description,
+        issue_category: dta_issue.issue_category,
+        decision_date: dta_issue.decision_date,
+        disposition: dta_issue.disposition,
+      )
+    end
+  end
+
+  def sync_dispositions(reference_id)
```
shanear: i think we can probably do this without the need to yield. But this is okay too

app/models/higher_level_review.rb
```diff
@@ -47,8 +48,55 @@ def valid_modifiers
...
+
   private
 
+  def dta_errors
+    DTA_ERRORS
+  end
+
+  def create_dta_supplemental_claim
+    return if dta_issues.empty?
+    dta_supplemental_claim.create_issues!(build_follow_up_dta_issues)
```
shanear: One last think I think we probably want to also `process_end_product_establishments!` for the supplemental claim too

jiahuang: Is that to run `end_product_establishment.perform!` or grab contentions? Seems like there wouldn't be any contentions for these.

PR #6996, opened 2018-09-14
Fixing up VSO queues and search -> case details page

mdbenjam:
### Description
Fixing some minor bugs with tasks showing up in the VSO queue (our VSO user needs to be a part of PVA not American Legion) and viewing the case details as a VSO or from case search.

### Testing Plan
1. Intake a case, choose appellant Bob Vance
1. Log in as VSO user
1. Go to `/organizations/paralyzed-veterans-of-america`
1. Ensure there is the appeal that was intook
1. Click on the case details link and ensure the page loads properly

### DIFF
--- a/app/models/serializers/work_queue/task_serializer.rb
+++ b/app/models/serializers/work_queue/task_serializer.rb
@@ -15,8 +15,8 @@ class WorkQueue::TaskSerializer < ActiveModel::Serializer
     {
       first_name: object.assigned_by_display_name.first,
       last_name: object.assigned_by_display_name.last,
-      css_id: object.assigned_by.css_id,
-      pg_id: object.assigned_by.id
+      css_id: object.assigned_by.try(:css_id),
+      pg_id: object.assigned_by.try(:id)
     }
   end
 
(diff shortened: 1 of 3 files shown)


app/models/serializers/work_queue/task_serializer.rb
```diff
@@ -15,8 +15,8 @@ class WorkQueue::TaskSerializer < ActiveModel::Serializer
     {
       first_name: object.assigned_by_display_name.first,
       last_name: object.assigned_by_display_name.last,
-      css_id: object.assigned_by.css_id,
-      pg_id: object.assigned_by.id
+      css_id: object.assigned_by.try(:css_id),
```
mdbenjam: Not all tasks have `assigned_by`

client/app/queue/CaseSnapshot.jsx
```diff
@@ -109,6 +109,11 @@ export class CaseSnapshot extends React.PureComponent<Props> {
     const {
       taskAssignedToUser
     } = this.props;
+
+    if (!taskAssignedToUser) {
```
mdbenjam: From case search you might view a case you don't have tasks for.

PR #7011, opened 2018-09-17
Amprokop fix assign bug

amprokop:
### Description
- Fixes a bug where any status change in an attorney task would cause a judge task to be moved to "review" status
- Refactors seed data to give more cases to assign, and for clarity
- Refines selectors to correctly filter out on hold cases for judges

### Testing Plan
1. Sign in as BVAAABSHIRE, view 4 ama cases to assign, 2 to review
1. Sign in as BVALSPORER, view 1 ama colocated task
1. Sign in as BVASCASPER1, view 2 ama tasks to review


client/app/queue/selectors.js
```diff
@@ -164,15 +164,21 @@ export const onHoldTasksByAssigneeCssIdSelector: (State) => Array<Task> = create
 export const judgeReviewTasksSelector = createSelector(
   [tasksByAssigneeCssIdSelector],
   // eslint-disable-next-line no-undefined
-  (tasks) => _.filter(tasks, (task: TaskWithAppeal) => [null, undefined, 'review'].includes(task.action))
+  (tasks) => _.filter(tasks, (task: TaskWithAppeal) => {
+    if (task.appealType === 'Appeal') {
+      return [null, undefined, 'review'].includes(task.action) && (task.status === 'in_progress' || task.status === 'assigned');
```
amprokop: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/legacy_tasks/judge_legacy_task.rb#L4


https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/tasks/judge_task.rb#L2

@dannysellers — do you think maybe we used to need the `[null, undefined, 'review']` bit, but now we don't? You're the closest one to the code here, but i don't necessarily see how we might need it.

app/models/task.rb
```diff
@@ -123,7 +123,7 @@ def on_hold_duration_is_set
   end
 
   def update_parent_status
-    parent.when_child_task_completed if saved_change_to_status? && parent
+    parent.when_child_task_completed if saved_change_to_status? && completed? && parent
```
amprokop: @lowellrex, can you take a look at this change? PR description explains why we think we need it. Thanks.

lowellrex: Yeah. Great catch. Don't know how I didn't notice this before.

client/app/queue/selectors.js
dannysellers: Yeah, we definitely needed it at one point to accommodate some seed data, but I don't see any tasks w/undefined or null action any more. I think just checking `task.action == 'review'` should be fine

db/schema.rb
```diff
@@ -613,7 +612,6 @@
   create_table "supplemental_claims", force: :cascade do |t|
     t.string "veteran_file_number", null: false
     t.date "receipt_date"
-    t.datetime "established_at"
```
aroltsch: @amprokop I am not sure why there are changes to schema as part of the changes you made :/

amprokop: I bet someone reverted a change. I'll take these out.

PR #7015, opened 2018-09-18
[wip] layering to improve circleci

enriquemanuel:

### Description

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,9 +3,10 @@ jobs:
   build:
     docker:
       # The first image listed is the image that all steps run inside of.
-      # This can be modified by steps, or can be built as a separate
-      # customized container.
-      - image: 008577686731.dkr.ecr.us-gov-west-1.amazonaws.com/circleci:latest
+      # If we want to use circle ci ruby 2.2 just change this image to:
+      # - image: circleci/ruby:2.2-node-browsers
+      # running circle ci ruby 2.5.1
+      - image: circleci/ruby:2.5.1-node-browsers
         environment:
           - DBUS_SESSION_BUS_ADDRESS: /dev/null
           - RAILS_ENV: test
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -3,9 +3,10 @@ jobs:
   build:
     docker:
       # The first image listed is the image that all steps run inside of.
-      # This can be modified by steps, or can be built as a separate
-      # customized container.
-      - image: 008577686731.dkr.ecr.us-gov-west-1.amazonaws.com/circleci:latest
+      # If we want to use circle ci ruby 2.2 just change this image to:
+      # - image: circleci/ruby:2.2-node-browsers
+      # running circle ci ruby 2.5.1
+      - image: circleci/ruby:2.5.1-node-browsers
```
nemodjuric: My experience with 2.5.1-node-browsers image was that it gave time outs, as you can see from your first build. On the other side, 2.5 image didn't cause these time outs. See: https://github.com/department-of-veterans-affairs/caseflow/pull/6461
Also, they got new files on Docker Hub yesterday, the new name is 2.5.1-node-browsers-legacy

.circleci/config.yml
```diff
@@ -29,138 +30,168 @@ jobs:
 
       # This is the circleci provided Redis container.
       - image: circleci/redis:4.0.10
-    parallelism: 5
+    parallelism: 6
```
nemodjuric: Is there are reason why did we increment this (and why by one only)?

enriquemanuel: no there isnt a reason, just a theory that i was testing

kencheeto:
ah, interesting pr! is this an effort to use a stock CircleCI build image so our chances of having cached images increases?

I've noticed that downloading all of the docker images can take a long time on builds, I've seen some take 7+ minutes just downloading images, while others have everything cached and take only a few seconds.

enriquemanuel:
that is correct, but most of the stuff was removed since we can't store the oracle dependencies in the same repo since we don't own them

pkarman:
@enriquemanuel should this PR be closed?

PR #7017, opened 2018-09-18
Edit request issues on claims

kencheeto:
Connects #6212 

### Description 🚧 
Adds and hooks up a button for saving selected rated issues on a higher level claim or supplemental claim.

Implements:
* SaveButton and SelectIssuesButtons components for editing rated issues for an intake
* Associated redux actions / reducers
* PATCH /higher_level_reviews/:claim_id endpoint for updating rated issues
* PATCH /supplemental_claims/:claim_id endpoint for updating rated issues

This demo shows all the error states, but the 'no change' should not be possible to reach in the UI as the button is disabled until changes are made.
demo

### Acceptance Criteria
- [x] Code compiles correctly
- [x] Feature tests
- [x] Clicking the "Save" button should trigger changing the issues.
- [x] Error messaging
- [x] Success messaging? (could come in a followup)
- [x] The "Save" button is grayed out until a change is made, then it's the default blue button color.
- [x] When the issues are saved, if issues of the same type (rated or non-rated) are added or removed to/from an existing EP, the new contentions are saved in VBMS under the same EP code and label (done via prior pr https://github.com/department-of-veterans-affairs/caseflow/pull/6781)

### Testing Plan
1. Select BVAAABSHIRE user (Judge with hearings and cases)
1. Pick a vbms_id out of bgs_setup.csv with has_rating (e.g. 231439628)
1. Complete intake of a request for a higher level review with the selected vbms_id
1. Look up the claim_id of the new EP
1. Visit /higher_level_reviews/:claim_id/edit
1. Select/deselect issues and save to test the following cases
   1. No change -> button is disabled gray
   1. No issues selected -> hitting save displays a 'no issues selected' error message
   1. Some or all issues selected -> hitting save displays a 'success' message
   1. While saving, the button should be disabled


app/controllers/higher_level_reviews_controller.rb
```diff
@@ -2,6 +2,23 @@ class HigherLevelReviewsController < ApplicationController
   before_action :verify_access, :react_routed, :verify_feature_enabled, :set_application
   SOURCE_TYPE = "HigherLevelReview".freeze
 
+  def update_issues
```
shanear: We may just want to use the idiomatic `update` route here.

https://guides.rubyonrails.org/v5.1/routing.html#crud-verbs-and-actions

app/controllers/higher_level_reviews_controller.rb
```diff
@@ -2,6 +2,23 @@ class HigherLevelReviewsController < ApplicationController
...
+    request_issues_update = RequestIssuesUpdate.new(
+      user: current_user,
+      review: higher_level_review,
+      request_issues_data: params[:request_issues]
+    )
```
shanear: we generally lean towards private methods over variable assignment, so i'm feeling like this would be better as a private method

client/app/intakeEdit/constants.js
```diff
@@ -3,3 +3,21 @@ export const PAGE_PATHS = {
...
+};
+
+export const ACTIONS = {
+  SET_ISSUE_SELECTED: 'SET_ISSUE_SELECTED',
+  REQUEST_ISSUES_UPDATE_START: 'REQUEST_ISSUES_UPDATE_START',
+  REQUEST_ISSUES_UPDATE_SUCCEED: 'REQUEST_ISSUES_UPDATE_SUCCEED',
+  REQUEST_ISSUES_UPDATE_FAIL: 'REQUEST_ISSUES_UPDATE_FAIL'
+};
+
+export const REQUEST_STATE = {
```
shanear: This is also in `/intake` might be worth combining into `/intakeCommon`

client/app/intakeEdit/reducers/index.js
```diff
@@ -1,12 +1,17 @@
...
 
 export const mapDataToInitialState = function(props = {}) {
   return {
     formType: props.formType,
     review: props.review,
-    ratings: formatRatings(props.ratings, props.ratedRequestIssues)
+    ratings: formatRatings(props.ratings, props.ratedRequestIssues),
+    requestStatus: {
+      requestIssuesUpdate: REQUEST_STATE.NOT_STARTED
+    },
```
shanear: Since there's just one call on this page, I'm cool if this is just something like:

`submitRequestStatus: REQUEST_STATE.NOT_STARTED`

but what you have is also more consistent with intake and I like that too. Just a thought.

spec/feature/intake/edit_spec.rb
```diff
@@ -112,6 +112,9 @@ def click_cancel(visit_page)
       visit "supplemental_claims/#{supplemental_claim.end_product_claim_id}/edit/select_issues"
       expect(find_field("PTSD denied", visible: false)).to_not be_checked
       expect(find_field("Left knee granted", visible: false)).to be_checked
+      find("label", text: "PTSD denied").click
+
+      safe_click("#button-submit-update")
```
shanear: I presume we'll write more tests when we pair 😄

client/app/intakeEdit/pages/selectIssues.jsx
```diff
@@ -1,24 +1,34 @@
...
     } = this.props;
 
     const reviewForm = _.find(FORM_TYPES, { key: formType });
 
     return <div>
       <h1>Issues on { veteranName }'s { reviewForm.name }</h1>
 
+      <p>status of request: { requestStatus }</p>
+      <p>response data: { responseErrorData }</p>
+      <p>response code: { responseErrorCode }</p>
```
shanear: also, we should probably delete this stuff eventually 😄

shanear: We can make another story on error handling on the edit page

client/app/intakeEdit/reducers/index.js
```diff
@@ -26,6 +31,38 @@ export const intakeEditReducer = (state = mapDataToInitialState(), action) => {
...
+  case ACTIONS.REQUEST_ISSUE_UPDATE_FAIL:
+    const serverIssues = [];
+    return update(state, {
+      requestStatus: {
+        requestIssuesUpdate: {
+          $set: REQUEST_STATE.FAILED
+        }
+      },
+      responseErrorCode: { $set: action.payload.responseErrorCode },
+      responseErrorData: { $set: action.payload.responseErrorData }
```
shanear: I don't think we are using `responseErrorData` anywhere. We can probably strip it from this entire PR.

app/controllers/higher_level_reviews_controller.rb
shanear: also we need to add this action to `supplemental_claims_controller`. I don't see that anywhere

client/app/intakeEdit/actions/selectIssues.js
```diff
@@ -0,0 +1,52 @@
...
+      },
+      (error) => {
+        let responseObject = {};
+
+        try {
+          responseObject = JSON.parse(error.response.text);
+        } catch (ex) { /* pass */ }
+
+        const responseErrorCode = responseObject.error_code;
+        const responseErrorData = responseObject.error_data;
```
shanear: yeah we aren't sending this in the controller anywhere so we can take it out.

client/app/intakeEdit/actions/selectIssues.js
```diff
@@ -0,0 +1,52 @@
...
+
+const analytics = true;
+
+export const requestIssuesUpdate = (claimId, state) => (dispatch) => {
+  dispatch({
+    type: ACTIONS.REQUEST_ISSUES_UPDATE_START,
+    meta: { analytics }
+  });
+
+  const data = formatIssues(state);
```
shanear: It's sweet that we were able to reuse this 💯

client/app/intakeEdit/pages/selectIssues.jsx
kencheeto: I copied the way finishing a higher level review shows error messages, can edit that as necessary!

kencheeto:
ah, I've actually introduced a bug, fixing now

app/models/end_product_establishment.rb
```diff
@@ -8,6 +8,7 @@
 
 class EndProductEstablishment < ApplicationRecord
   class EstablishedEndProductNotFound < StandardError; end
+  class ContentionCreationFailed < StandardError; end
```
leikkisa: 👍

spec/feature/intake/edit_spec.rb
```diff
@@ -67,6 +67,51 @@
...
+      expect(page).to have_button("Save", disabled: true)
+
+      # reload to verify that the new issues populate the form
+      visit current_path
+      expect(find_field("PTSD denied", visible: false)).to be_checked
+      expect(find_field("Left knee granted", visible: false)).to_not be_checked
+
+      # assert server has updated data
+      expect(higher_level_review.reload.request_issues.first.description).to eq("PTSD denied")
+      expect(request_issue.reload.review_request_id).to be_nil
```
leikkisa: I think we don't want to disconnect it from the review_request, but to add a removed_at timestamp.  Maybe you meant to check that the contention_reference_id is nil?

spec/feature/intake/edit_spec.rb
```diff
@@ -67,6 +67,51 @@
...
+      # expect contentions to reflect issue update
```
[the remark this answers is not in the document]
leikkisa: We can check the calls that get made to VBMS, I think it will be easier to chat about this in Slack.

spec/feature/intake/edit_spec.rb
```diff
@@ -113,5 +158,68 @@ def click_cancel(visit_page)
       expect(find_field("PTSD denied", visible: false)).to_not be_checked
       expect(find_field("Left knee granted", visible: false)).to be_checked
     end
+
+    it "enables save button only when dirty" do
```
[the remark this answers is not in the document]
leikkisa: That's a good point. I am not sure to what extent Rails magic is happening here.  There are some differences between HLR and SC, but not sure if those will be part of what is editable in the future.  Let's keep an eye on it!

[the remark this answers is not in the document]
leikkisa: For the feature tests, I agree that we don't need to duplicate all of the feature tests for both flows for edit, especially since they're sharing the same code.

client/app/intakeEdit/pages/selectIssues.jsx
```diff
@@ -30,6 +39,39 @@ class SelectIssues extends React.PureComponent {
   }
 }
 
+class SaveButtonUnconnected extends React.PureComponent {
+  handleClick = () => {
+    this.props.requestIssuesUpdate(this.props.claimId, this.props.formType, {
+      ratings: this.props.ratings,
+      nonRatedIssues: this.props.nonRatedIssues
```
leikkisa: I'm not sure why we need to include this here since we're not updating nonRatedIssues.

spec/feature/intake/edit_spec.rb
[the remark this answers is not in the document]
kencheeto: Okay - without knowing more about what's next for editing, I think I will leave them separate as copies of each other. If we get further with feature work and they end up being the same, we'll be able to extract out a lot of common code and tests 🎉

spec/feature/intake/edit_spec.rb
kencheeto: Ah, I was following what I saw here, which is incomplete: https://github.com/department-of-veterans-affairs/caseflow/blob/b4f2417e054ede86e6fb0a1ad8b57348de2449b6/app/models/request_issues_update.rb#L90-L94

I'll add the other checks that you showed me in Slack.

client/app/intakeEdit/pages/selectIssues.jsx
kencheeto: Ah good catch, this was from fiddling around early on

client/app/intakeEdit/components/RequestIssuesUpdateErrorAlert.jsx
```diff
@@ -0,0 +1,25 @@
...
+        title: 'Something went wrong',
+        body: 'Please try again. If the problem persists, please contact Caseflow support.'
+      }
+    }[this.props.responseErrorCode || 'default'];
+
+    return <Alert title={errorObject.title} type="error" lowerMargin>
+      {errorObject.body}
+    </Alert>;
+  }
+}
```
shanear: Awesome, we got ahead of this! 💯

spec/feature/intake/edit_spec.rb
```diff
@@ -113,5 +180,90 @@ def click_cancel(visit_page)
...
+      end
+
+      scenario "from landing page" do
+        click_cancel("/")
+      end
+
+      scenario "from select_issues page" do
+        click_cancel("/select_issues")
+      end
+    end
```
shanear: Since feature tests are expensive we sometimes wrap multiple "tests" up into one scenario. This is cool though, just FYI

spec/feature/intake/edit_spec.rb
```diff
@@ -67,6 +67,73 @@
...
+        special_issues: []
+      )
+      expect(Fakes::VBMSService).to have_received(:associate_rated_issues!).with(
+        claim_id: higher_level_review.end_product_claim_id,
+        rated_issue_contention_map: {
+          new_request_issue.rating_issue_reference_id => new_request_issue.contention_reference_id
+        }
+      )
+      expect(Fakes::VBMSService).to have_received(:remove_contention!).once
+    end
```
shanear: These are some awesome tests 👍:+1:

leikkisa:
I suppose we could keep the contention reference ID for now since we're indicating elsewhere that it's been removed and I can't think of any instances where it would hurt anything, but we should be aware that the contention it's referring to no longer exists.  If we decide to remove it later, then we can clean up the data.

spec/feature/intake/edit_spec.rb
kencheeto: @leikkisa basically wrote them for me 😄

PR #7021, opened 2018-09-18
Queue | Fix serializing request_issues count for ama appeals

dannysellers:
### Description
Within `TasksController.index`, tasks for both legacy and AMA appeals are serialized using `TaskSerializer`, which was returning `issue_count` as `object.appeal.issues.count`. However, for Appeals, [`issues`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/appeal.rb#L33) returns a hash containing `request_issues` and `decision_issues`, so the `count` would always be 2. This fixes the count, returning `request_issues.count` for AMA appeals and `issues.count` for legacy.

### Acceptance Criteria
- [ ] Issue count is correct in task table views

### Testing Plan
1. Go to `/queue` as `BVASCASPER1`
2. Confirm 2 AMA appeals have issue counts != 2
3. Click in to case details view, confirm issue counts match

<details><summary>before</summary>

</details>
<details><summary>after</summary>

</details>


app/models/serializers/work_queue/task_serializer.rb
```diff
@@ -57,7 +57,11 @@ class WorkQueue::TaskSerializer < ActiveModel::Serializer
   end
 
   attribute :issue_count do
-    object.appeal.issues.count
+    if object.appeal.class.name.eql? "LegacyAppeal"
```
evankroske: I think it would be cleaner to add a method issue_count to Appeal and LegacyAppeal that returns the number of issues. That way, if we added a new type of appeal, it wouldn't inherit the behavior of Appeal. But I don't think this issue is very important. It's up to you

PR #7024, opened 2018-09-18
Anya/bva dispatch idt

### LINKED ISSUE #6890: Ensure BVA Dispatch users can view their AMA appeals when making a call to /idt/api/v1/appeals
# AC
- When making a call to /idt/api/v1/appeals, BVA Dispatchers should get a list of their AMA appeals
- We don't *need* to return legacy appeals in that queue for now (although how easy is it?)

aroltsch:
Resolves #6890 

### Description
BVA Dispatch users should be able to use IDT for AMA appeals


lib/constants/bva_dispatch_teams.rb
```diff
@@ -3,7 +3,7 @@ module Constants::BvaDispatchTeams
     preprod: %w[CF_KIRK_317],
     uat: %w[CF_SPOCK_317 CF_KHAN_397],
     test: %w[BVATEST1 BVATEST2 BVATEST3],
-    development: %w[BVALSPORER],
-    prod: %w[]
+    development: %w[BVAGWHITE],
+    prod: %w[BVAEJAMES VSCWDAVI]
```
aroltsch: Adding two pilot users.

db/seeds.rb
```diff
@@ -219,59 +224,59 @@ def create_ama_appeals
         FactoryBot.build(:claimant, participant_id: "OTHER_CLAIMANT")
       ],
       veteran_file_number: "701305078",
-      request_issues: FactoryBot.build_list(:request_issue, 3, description: "Head trauma")
```
aroltsch: build_list won't save it in the DB

app/controllers/tasks_controller.rb
```diff
@@ -101,7 +98,7 @@ def for_appeal
   private
 
   def queue_class
-    QUEUES[user_role.try(:to_sym)]
+    QUEUES[user_role.try(:to_sym)] || QUEUES[:generic]
```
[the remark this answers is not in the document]
lowellrex: I think that should work!

mdbenjam:
Can you also remove this line? https://github.com/department-of-veterans-affairs/caseflow/blob/master/spec/controllers/idt/api/appeals_controller_spec.rb#L3

Issue #7023, opened 2018-09-18
RampElectionRollback should not attempt to reopen appeals that are already open


shanear:
We have an issue where we are attempting to reopen VACOLS appeals that are already open. This throws the error you can see in the sentry link below.

We should instead skip those.

https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2440/

## Technical Notes

I believe the fix would be to filter the VACOLS id's to reopen better here:
https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/ramp_election_rollback.rb#L29

PR #7027, opened 2018-09-18
Only reopen closed appeals when rolling back ramp elections

shanear:
connects #7023


spec/models/ramp_election_rollback_spec.rb
```diff
@@ -107,6 +107,20 @@
...
+      create(
+        :legacy_appeal,
+        vacols_case: create(
+          :case_with_soc,
+          :type_original,
+          bfboard: "00",
+          bfkey: "34567"
+        )
+      )
+    end
```
shanear: Adding this causes the existing test to fail in all the right ways. No need for additional tests.

Issue #6965, opened 2018-09-13
Intake | Style change to make form labels bold


suzchap:
## Story
CSS for all intake form labels should be bold (`<strong>`).

## Technical notes
This will require adding the `strongLabel` configuration to `SearchableDropdown`

## Acceptance criteria
- form labels display as bold (style is currently set to just regular, not bold). Note, not all text that's above a form field has been coded as a form label so these won't be affected but we will change those to form labels in other tickets.

## Notes
- Since we are short on design time, the engineer working on this issue should check what labels in intake are not bold.

## How to Validate
- We have labels on almost every page so I'm content if we validate that some that were not bold before, now are. One form that does have a proper form label but isn't bold, is the "Issue category" field that you can find by going to Higher-level Review form > search any ID > answer questions any way you want > on the Finish page, click "add issue". If "Issue category" is bold, consider this validated.

suzchap:
This is waiting on me to add a list of all the places where a label needs to be bold.

shanear:
Everywhere I've checked looks good!

Issue #6960, opened 2018-09-13
Search page: copy and style changes


suzchap:
## Story
Fine-tuning of content and styles.

## Acceptance criteria
- Label for search box is "Enter the Veteran’s ID"
- "Enter the Veteran’s ID" is coded as a form label

## Copy
- Enter the Veteran’s ID

## Mocks
<img width="610" alt="screen shot 2018-09-13 at 12 59 08 pm" src="https://user-images.githubusercontent.com/11219417/45503444-d72d3480-b754-11e8-9080-21731e3dfa36.png">

## How to Validate
- Select any AMA form to land on the Search page.

suzchap:
<img width="1100" alt="screen shot 2018-09-27 at 12 07 07 pm" src="https://user-images.githubusercontent.com/11219417/46159277-e7154000-c24d-11e8-8ce7-7e35a77eb64a.png">

carolaponce:
LGTM

carolaponce:
Thank you @suzchap for walking me through these!

PR #7039, opened 2018-09-18
Search bar text update

pkarman:
connects #6960 

Note that form label is not (yet) bold but will be via #6965

### Description

screen shot 2018-09-18 at 3 42 01 pm

### Acceptance Criteria
- [ ] Code compiles correctly
- [ ] Tests continue to pass
- [ ] Intake Search form looks like mock


kahlouie:
@suzchap It's still a WIP since we're unclear if the form label should be bolded

PR #7042, opened 2018-09-18
add bootsnap gem for faster rails boot

kencheeto:
### Description
Rails 5.2 ships with a new default gem called `bootsnap`, which [speeds up Rails boot time by 50%](https://github.com/rails/rails/pull/29313). I've added it here to see if it'll work for Caseflow. Faster boot times mean a shorter development loop, less CI contention, and faster boot/restart during deploys.

[How it works](https://github.com/Shopify/bootsnap#how-does-this-work)
tl;dr avoiding slow scanning of $LOAD_PATH and caching of yaml files (both are slow parts of Rails boot)

Results for caseflow, locally testing a couple results of `time bundle exec rails s` until the 'Ctrl-C' prompt appears:

before:
real  0m8.811s
real  0m8.325s

after:
real  0m3.939s
real  0m4.098s

~53% decrease in bootup time, seems legit.

### Potential risks
* This gem does affect how files are loaded during app boot, so there is some risk that app boot in a non-dev environment goes awry. A test deploy should alleviate those concerns.
* It's possible that bootsnap negatively affects development by caching filepaths too aggressively (see 'stable entries' under https://github.com/Shopify/bootsnap#path-pre-scanning). I don't know of a use case where those paths need to be uncached.

I'm adding this gem to every group, since there's nothing development-specific about it and it seems better to both take advantage of speed gains in other envs and run as similar of a setup in dev and prod as possible. Shopify uses this gem in every environment and it's the new Rails default.

Tangentially, I'm also curious if we've tried upgrading to Rails 5.2 outright.

### Acceptance Criteria
- [x] Tests pass

### Testing Plan
In addition to working with this gem enabled locally for development, I'd like to do a test deploy to an AWS environment and restart to confirm that both of these succeed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -170,6 +170,6 @@ gem "rubyzip", "~> 1.2.2"
 gem "business_time", "~> 0.9.3"
 
 # Bootsnap speeds up app boot (and started to be a default gem in 5.2).
-gem 'bootsnap', require: false
+gem "bootsnap", require: false
 
 # rubocop:enable Metrics/LineLength
--- a/Gemfile
+++ b/Gemfile
@@ -143,8 +143,6 @@ group :stubbed, :test, :development, :demo do
   # Added at 2018-05-16 22:09:10 -0400 by mdbenjam:
   gem "factory_bot_rails", "~> 4.8"
 
-  # Bootsnap speeds up app boot (and started to be a default gem in 5.2).
-  gem 'bootsnap', require: false
 end
 
 group :stubbed, :development do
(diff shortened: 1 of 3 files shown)


sunil-sadasivan:
say what!

kencheeto:
@shanear is there a pre-production environment I could test a deploy & restart on?

kencheeto:
I will keep an eye on this - I've been working with it locally and haven't noticed any differences yet.

PR #7043, opened 2018-09-18
add --parallel to speed up rubocop

kencheeto:
### Description
I noticed that builds spend a lot of time linting (2 minutes total, about 1m30 of that is rubocop). I tried adding the --parallel flag when running rubocop locally on my 4 core laptop and I get a great speedup:
```
kenshiro15:caseflow kencheeto$ time rubocop --parallel
...
real	0m21.356s
```
vs
```
kenshiro15:caseflow kencheeto$ time rubocop
...
real	1m10.200s
```
I'm not sure how many CPUs each CircleCI container has (I get 404s when I try to look at project configs) but testing it out here. Looks like we save just under a minute off of each container's step:

before:
screen shot 2018-09-18 at 4 35 33 pm
after:
screen shot 2018-09-18 at 4 35 02 pm

Seems like a cheap free win for build times! Should take us from ~16 minute to ~15 minute builds in CI.

### Acceptance Criteria
- [x] Tests pass

### Testing Plan
None, internal CI change

### DIFF
--- a/lib/tasks/lint.rake
+++ b/lib/tasks/lint.rake
@@ -9,7 +9,7 @@ task :lint do
 
   opts = ENV["CI"] ? "" : "--auto-correct"
   puts "running rubocop..."
-  rubocop_result = ShellCommand.run("rubocop #{opts} --color")
+  rubocop_result = ShellCommand.run("rubocop #{opts} --color --parallel")
 
   puts "\nrunning eslint..."
   eslint_cmd = ENV["CI"] ? "lint" : "lint:fix"


mdbenjam:
That's awesome! Thanks for looking into it! One note, FACOLS takes a long time to setup, we moved linting to the beginning of the build a while ago in order to run at the same time as FACOLS was starting up. If this change speeds things up enough we might lose that same time to FACOLS setup. If this is the case we can move other things up in the process. Specifically, we could move the js tests up.

kencheeto:
@mdbenjam ah, okay. Do you mean linting and FACOLS setup run in parallel and tests won't start until FACOLS is done being setup anyway, meaning linting speed improvements don't actually globally reduce CI time?

mdbenjam:
Exactly, but I'm not sure how long it takes FACOLS to setup on circle. We should keep an eye on this:
screen shot 2018-09-19 at 1 49 47 pm

If the time increases a lot then we're probably not gaining anything globally.

kencheeto:
Ok, will do. At first glance the timing varies widely for these individual steps (0:50 to 2:00 for linting and 0:12 to 3:00 for facols setup from three or four I looked at), so I'm going to note that the overall median time is 15:54 and have made a note to check in a week or two.

Issue #7056, opened 2018-09-19
Allow closed CAVC cases to be reopened


shanear:
Right now, we are throwing an error whenever we try and reopen a case that was closed while in CAVC. Verified with the Board that it's safe to reopen these back to CAVC.

This is currently causing this error:
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2461/

PR #7058, opened 2018-09-19
Allow CAVC appeals to be reopened

shanear:
Also remove another frivolous check.

connects #7056


cmgiven:
Shane is a king of glorius commits. Bow unto the majesty of his code.

PR #7060, opened 2018-09-19
Do not overwrite Task instructions when updating Task

### LINKED ISSUE #7055: Do not overwrite Task instructions when updating Task
We are overwriting task instructions when we take any action on the task. One common workflow for co-located staffers is receiving a task from an Attorney to grant an extension request, sending the Judge a word doc for their signature via email, putting the task on hold for 1 day as they waits for the Judge to respond to the email and adding a note to that effect in the instructions field. This overwrites the previous instructions from the Attorney. After the judge has responded with the signed document (and the hold has expired) co-located staffers put the task on hold for 90 days (as per the extension request) and add another note in the instructions field to that effect (again, overwriting the note that explained why there was a 1 day hold). 

Slack discussion here: https://dsva.slack.com/archives/C6E41RE92/p1537385995000100

## Acceptance criteria:
* Retain the instructions in each of these steps.

lowellrex:
Resolves #7055.

Changes `Task.instructions` field to an array of text fields and appending new instructions to that array instead of overwriting the `Task.instructions` field on updates.


db/migrate/20180919213618_change_task_instructions_field_type.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeTaskInstructionsFieldType < ActiveRecord::Migration[5.1]
+  def change
+    change_column :tasks, :instructions, :text, array: true, default: [], using: "(string_to_array(instructions, ','))"
```
aroltsch: Is this gonna migrate existing instructions correctly?

db/schema.rb
```diff
@@ -169,6 +169,7 @@
     t.string "participant_id", null: false
     t.string "payee_code"
     t.date "date_of_birth"
+    t.index ["date_of_birth"], name: "index_claimants_on_date_of_birth"
```
aroltsch: You have a few changes here that are not related to your PR.

app/models/task.rb
```diff
@@ -32,10 +32,12 @@ def assigned_by_display_name
 
   def self.create_from_params(params, current_user)
     verify_user_can_assign(current_user)
+    params["instructions"] = [params["instructions"]] if params.key?("instructions")
```
aroltsch: I would personally leave as a text field and just append to the previous one. This way we don't have to write any migrations to change the type of the field in the DB. I think this is temporary solution so at some point we will have to migrate back to the text field. If you don't foresee any problems with that, I am okay with making it an array.

db/migrate/20180919213618_change_task_instructions_field_type.rb
lowellrex: Yup! I ran this locally and it worked properly.

db/schema.rb
lowellrex: Right. I think somebody did not include this file in a previous PR that included a migration.

app/models/task.rb
lowellrex: The problem I ran into when appending new instructions to the existing instructions was getting them to show up properly on the frontend. Newline characters are not rendered and inline HTML line break elements aren't interpreted, so all of the different sets of instructions would run together when displaying them on the frontend.

I'm open to ideas for how to make the single text field approach work though. Maybe we could insert some magic character (two newlines?) and use that to split the text field into an array on the frontend so we could display it properly there?

db/schema.rb
aroltsch: No bueno.

Issue #7063, opened 2018-09-19
Return docket type for AMA appeals from the IDT API


amprokop:
# AC
- Return docket type for AMA appeals

Issue #7065, opened 2018-09-20
Catch unexpected errors from IDT API 


amprokop:
If there's an unexpected IDT API error, we currently return a big ol Rails error HTML page

# AC
- Return a 500 and an "Unexpected Error" message if we have an unhandled error in the IDT API

laurjpeterson:
@aroltsch is this ticket still relevant?

Issue #7064, opened 2018-09-19
BVA Dispatch IDT appeal details returns error


amprokop:
Steps to reproduce:
- Review a case as a judge
- Find that case's appeal id
- Create a token associated with a BVA dispatch user 
- Curl localhost:3000/idt/api/v1/appeals/:appeal_id 
- See that an error is returned

When creating BVA Dispatch tasks assigned to the user, it looks like we do not set assigned_by_id, since the task is assigned round robin.

Looks to me like the test seed data we have sets that field, but it might be the case that it would never be set in a real life situation, so localhost:3000/idt/api/v1/appeals/:appeal_id for a dispatch user might always fail.

Not sure what will happen for legacy appeals, but we aren't handling them quite yet.

# AC
- localhost:3000/idt/api/v1/appeals/:appeal_id should succeed when passing a token associated with a BVA dispatch user

laurjpeterson:
@aroltsch is this ticket still relevant?

PR #7066, opened 2018-09-20
odds and ends

amprokop:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/7063, https://github.com/department-of-veterans-affairs/caseflow/issues/7064, https://github.com/department-of-veterans-affairs/caseflow/issues/7065

### Testing Plan
- Do a judge case review and verify that BVA Dispatch task is created associated with the correct appeal
- curl localhost:3000/idt/api/v1/appeals/:appeal_id with a BVA Dispatch user token succeeds
- IDT API returns docket name
- when inserting `fail` somewhere in the IDT API, a 500 with `Unexpected Error` is returned rather than the Rails error html page


db/seeds.rb
```diff
@@ -287,23 +287,23 @@ def create_ama_appeals
     LegacyAppeal.create(vacols_id: "2306397", vbms_id: "779309925S")
   end
 
-  def create_root_task
-    FactoryBot.create(:root_task)
+  def create_root_task(appeal)
```
amprokop: these root tasks were not assigned to the correct appeal before, so the handoff to BVA Dispatch from judge case review did not happen correctly in local dev

spec/factories/task.rb
```diff
@@ -76,6 +76,7 @@
       type "BvaDispatchTask"
       appeal_type "Appeal"
       appeal { create(:appeal) }
+      assigned_by nil
```
amprokop: Since BVA Dispatch user tasks are assigned round robin, this will never have a value.

Issue #6940, opened 2018-09-12
Make SC/HLR establishment idempotent


shanear:
In order to be able to retry processing the VBMS calls for a claim review, we need that function to be idempotent.

## Acceptance Criteria
- The `process_end_product_establishments!` should be idempotent, and not make calls that have already succeeded.

pkarman:
Some notes to make sure I am understanding the scope of this ticket. Please correct/challenge my assumptions.

## Current behavior

The `ClaimReview` `process_end_product_establishments!` method encapsulates 4 tasks:

 1. create an EndProduct corresponding to the Claim via `end_product_establishment.perform!`
 1. create contentions for the new EndProduct
 1. create rated issues for any of the new contentions
 1. update the RequestIssues for each contention for `rating_issue_associated_at`

The first 3 tasks all involve one or more `VBMSService` calls, which may fail explicitly with a non-2xx response or may timeout altogether, leaving VBMS in some unknown state.

In the happy path, all 4 tasks succeed, in that order.

In the unhappy path, the first `VBMSService` call to fail will raise an exception and interrupt the flow.

In either case, calling `process_end_product_establishments!` again could create duplicate VBMS records.


## New behavior

We want to be able to call `process_end_product_establishments!` repeatedly and trust that any already-successful tasks are skipped.
This new behavior requires that either:

 * every VBMSService call is self-aware --- it has some way of tracking its completeness and/or recognizing itself in existence already at VBMS, OR
 * some orchestrating class (could be `ClaimReview`, `EndProductEstablishment` or some as-yet-undefined class) tracks the state of each task

## Proposal

As described in #5339 there are two common models for this kind of [distributed transaction pattern](https://blog.couchbase.com/saga-pattern-implement-business-transactions-using-microservices-part/). When considering what direction to take, I think it's helpful to consider the additional requirements mentioned in other tickets.

Tasks should be:

 * idempotent (this ticket)
 * asynchronous (#6941)
 * reversible (#6942)

We could write several more business-specific methods on `ClaimReview` and/or `EndProductEstablishment` to accomplish this, but I think in the long run those classes play more of the orchestrator role and we can abstract the tasks away from the task master.

[...]

Issue #6942, opened 2018-09-12
Retry failed VBMS requests for HLR and SC establishments


shanear:
Once we begin processing VBMS requests for HLR and SC establishments asynchronously, we will need a way to retry them when they fail.

## Acceptance Criteria
- When any part of the VBMS establishment for an HLR or SC fails, a job should periodically attempt to retry processing them.
- An HLR or SC should be retried every 3 hours for 4 consecutive days. After that, it should no longer be retried.

requires #6941

pkarman:
Moving the tasks to ActiveJob in #7111 will get us some automatic retries per https://github.com/department-of-veterans-affairs/appeals-deployment/blob/master/docs/setting-up-a-caseflow-cron-job.md#step-1-implement-your-active-job

That only goes to a max of 5 hours though, so #7111 makes it possible to go for days instead.

However, for #6944 we need to capture the last error for any job so this ticket should be scoped to:

* rescue from any exceptions
* capture and save the error message
* re-throw the exception to trigger retry

pkarman:
looks like it re-ran with no error:
```
> puts hlr.pretty_inspect
#<HigherLevelReview:0x0000558b2e78aca8
 id: 214,
 veteran_file_number: "123458312",
 receipt_date: Sun, 09 Sep 2018,
 informal_conference: false,
 same_office: false,
 establishment_submitted_at: Wed, 03 Oct 2018 22:21:21 UTC +00:00,
 establishment_processed_at: Thu, 04 Oct 2018 01:22:29 UTC +00:00,
 benefit_type: "compensation",
 establishment_attempted_at: Thu, 04 Oct 2018 01:22:01 UTC +00:00,
 establishment_error: nil>
=> nil
```

shanear:
`pretty_inspect`! I'm totally gonna use that in the future lol


![screen shot 2018-10-05 at 5 35 31 pm](https://user-images.githubusercontent.com/1596259/46561313-13bbfe00-c8c5-11e8-8a3f-1a9701149155.png)

Issue #6941, opened 2018-09-12
Process VBMS requests for SCs/HLR establishments asynchronously


shanear:
Instead of processing the VBMS calls for HLRs and SCs synchronously when completing an intake, we should process them asynchronously to save time, since we'll retry failed requests anyways.

## Acceptance Criteria
- On the completion of an HLR/SC intake, a job should be triggered that completes the VBMS establishment actions asynchronously.
- This job should also be used to process the establishment of SCs triggered by DTA errors as described in #6168 

requires #6940
requires #6168

pkarman:
As I read the code, the establishment of SCs triggered by DTA errors happens via `SyncReviewsJob` -> `EndProductSyncJob` so it is already async.

shanear:
This is working! Successfully processed an HLR, SC intake as well as an issues update

PR #7074, opened 2018-09-20
Idempotent end product artifact processing

pkarman:
connects #6940

### Description
Tests to validate idempotency for `ClaimReview` `process_end_product_establishments!`

### Acceptance Criteria
- [x] Code compiles correctly
- [x] Tests continue to pass

### Testing Plan
TODO: can we simulate VBMS failure in UAT?

### DIFF
--- a/app/models/claim_review.rb
+++ b/app/models/claim_review.rb
@@ -24,18 +24,20 @@ class ClaimReview < AmaReview
     end
   end
 
-  # Send the appropriate calls to VBMS to create the end products and contentions for any
-  # outstanding end product establishment. If all end products have been created then this
-  # method does nothing
+  # Idempotent method to create all the artifacts for this claim.
+  # If any external calls fail, it is safe to call this multiple times until
+  # establishment_processed_at is successfully set.
   def process_end_product_establishments!
     return if establishment_processed_at
 
     end_product_establishments.each do |end_product_establishment|
+      next if end_product_establishment.committed?
       end_product_establishment.perform!
-      create_contentions_for_end_product_establishment(end_product_establishment)
+      end_product_establishment.create_contentions!
+      end_product_establishment.create_associated_rated_issues!
+      end_product_establishment.commit!
     end
 
-    end_product_establishments.each(&:commit!)
     update!(establishment_processed_at: Time.zone.now)
   end
 
@@ -66,16 +68,6 @@ class ClaimReview < AmaReview
   def end_product_establishment_for_issue(issue)
     ep_code = issue.rated? ? self.class.rated_issue_code : self.class.nonrated_issue_code
     end_product_establishments.find_by(code: ep_code) || new_end_product_establishment(ep_code)
-  end
-
-  def create_contentions_for_end_product_establishment(end_product_establishment)
-    request_issues_without_contentions = request_issues.where(
-      end_product_establishment: end_product_establishment,
-      contention_reference_id: nil
-    )
-
-    end_product_establishment.create_contentions!(request_issues_without_contentions)
-    end_product_establishment.create_associated_rated_issues!
   end
 
   def sync_dispositions(reference_id)
--- a/app/models/claim_review.rb
+++ b/app/models/claim_review.rb
@@ -6,17 +6,7 @@ class ClaimReview < AmaReview
 
   self.abstract_class = true
 
-  class << self
-    def rated_issue_code
-      self::END_PRODUCT_RATING_CODE
-    end
-
-    def nonrated_issue_code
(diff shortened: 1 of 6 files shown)


app/models/claim_review.rb
```diff
@@ -14,16 +24,21 @@ def create_issues!(new_issues)
...
     end_product_establishments.each do |end_product_establishment|
+      next if end_product_establishment.committed?
       end_product_establishment.perform!
-      create_contentions_for_end_product_establishment(end_product_establishment)
+      end_product_establishment.create_contentions!
+      end_product_establishment.create_associated_rated_issues!
+      end_product_establishment.commit!
     end
 
-    end_product_establishments.each(&:commit!)
```
shanear: The whole `committed_at` thing was a previous attempt at solving this problem. But if we were still doing that then we would set it here. Honestly, I don't even think we need `committed_at` any more, but that's probably for another PR.

app/models/claim_review.rb
```diff
@@ -14,16 +24,21 @@ def create_issues!(new_issues)
     end
   end
 
-  # Send the appropriate calls to VBMS to create the end products and contentions for any
-  # outstanding end product establishment. If all end products have been created then this
-  # method does nothing
+  # Idempotent method to create all the artifacts for this claim.
+  # If any external calls fail, it is safe to call this multiple times until
+  # establishment_processed_at is successfully set.
```
shanear: nice explanation 💯

app/models/claim_review.rb
```diff
@@ -49,49 +64,10 @@ def on_sync(end_product_establishment)
   private
 
   def end_product_establishment_for_issue(issue)
-    ep_code = issue_code(issue.rated?)
+    ep_code = issue.rated? ? self.class.rated_issue_code : self.class.nonrated_issue_code
```
shanear: This change will cause issues with this pr (here: https://github.com/department-of-veterans-affairs/caseflow/pull/6992/files#diff-dc75f311ca26386d12dba1e248641cffR58), since `issue_code()` gets a little more complicated

app/models/end_product_establishment.rb
```diff
@@ -70,17 +70,20 @@ def perform!(commit: false)
 
   # VBMS will return ALL contentions on a end product when you create contentions,
   # not just the ones that were just created.
-  def create_contentions!(for_objects)
+  def create_contentions!
```
shanear: I avoided moving a lot of this logic into `EndProductEstablishment` to try and keep it as generic as possible, but honestly, this is nicer lol 👍

spec/models/claim_review_spec.rb
```diff
@@ -211,6 +217,84 @@
           end
         end
       end
+
+      context "when called multiple times" do
```
shanear: nice test. A tad verbose, but this logic is complicated so i think it's warranted

app/models/claim_review.rb
pkarman: I think it's still helpful to see that a particular `EndProductEstablishment` is complete (committed). It avoids re-running already-complete EPE if the whole method is re-run.

Why would you wait till they are all complete to flag them all? Isn't that the point of `ClaimReview.establishment_processed_at`? To flag the atomic complete-ness of the ClaimReview?

pkarman: I do think the names are a little off. `commit` suggests to me that something could be rolled back, or that it was provisional until that moment, but now it's real. I think we really mean `completed_at`.

And I think it should be plural `ClaimReview.establishments_completed_at`

shanear: > Why would you wait till they are all complete to flag them all? Isn't that the point of ClaimReview.establishment_processed_at? To flag the atomic complete-ness of the ClaimReview?

Since there can be multiple EPs that are part of one atomic action, the original idea was that a epe wouldn't be "commited" until all the EPs as part of it's atomic action are complete.

I like your idea too though. I think we should probably record whatever decision we come to in the comments somewhere.

pkarman: Anything we can do to explicitly flag things as "done" is a win, IMO. It's a performance optimization (don't need to examine the EPE in detail anymore, just the flag), and it helps in debugging a "stuck" ClaimReview, which will become more important once we get #6941 and #6942 implemented.

pkarman: Unfortunately, `RequestIssuesUpdate` requires that we *not* check `EndProductEstablishment.committed_at` or `ClaimReview.establishment_completed_at` because we must re-examine them both in order to update associated issues.

So things aren't really "done" as long as their related artifacts can be removed (e.g. `remove_contention`).

PR #7086, opened 2018-09-20
Display judge and attorney assigned to case in case snapshot

### LINKED ISSUE #7069: Display VLJ and senior counsel on case details page
Many co-located administrative tasks require knowing who the VLJ and senior counsel are for a given case (to send them emails, for instance). This information is not visible on the case details page so co-located staff have to use VACOLS to determine who those folks are. That is an acceptable workaround for legacy appeals, but since AMA appeals will never enter VACOLS we need to add some way to determine who the judge and attorney working the case are.

Slack thread: https://dsva.slack.com/archives/C6E41RE92/p1537386021000100

## Acceptance criteria
* Display VLJ and senior counsel for an appeal on the case details page

lowellrex:
Resolves #7069. Displays VLJ and senior counsel for an appeal on the case details page.


app/models/concerns/taskable.rb
```diff
@@ -0,0 +1,11 @@
+module Taskable
```
lowellrex: Open to a name change on this or even moving these methods inside the two appeal classes. Just thought they should live out here so that we don't change the implementation for `Appeal` and forget to update the similar method for `LegacyAppeal`.

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -1,4 +1,7 @@
 class WorkQueue::AppealSerializer < ActiveModel::Serializer
+  attribute :assigned_attorney
+  attribute :assigned_judge
```
lowellrex: It would be really nice to be able to compose serializers so we don't have to duplicate the attributes as we currently do. I tried to make a little `TaskableSerializer` trait but couldn't get it working properly. Open to thoughts on how to make this work!

PR #7089, opened 2018-09-20
Enable task actions for VSOs

lowellrex:
The three main changes that were required to make the "Mark task as complete" action appear for VSOs were:
1. [Allowing VSO users access](https://github.com/department-of-veterans-affairs/caseflow/pull/7089/files#diff-f4416012d567d02da854cfbc9e2df2caL5) to the `tasks#update` action
2. [Including VSO tasks](https://github.com/department-of-veterans-affairs/caseflow/pull/7089/files#diff-2a3f4e296b35f6015a2d1d3ef72e5506L125) in the selector that plucked out tasks belonging to the current organization.
3. [Removing the explicit hiding](https://github.com/department-of-veterans-affairs/caseflow/pull/7089/files#diff-5b5c7473c1568400a8104eccd0693108L221) of the task actions dropdown for VSOs.


spec/feature/queue/task_queue_spec.rb
```diff
@@ -77,4 +73,47 @@
...
+    let!(:vso_task) { FactoryBot.create(:ama_vso_task, :in_progress, assigned_to: vso) }
+    before do
+      FeatureToggle.enable!(vso.feature.to_sym, users: [vso_employee.css_id])
+      User.authenticate!(user: vso_employee)
+      allow_any_instance_of(Vso).to receive(:user_has_access?).and_return(true)
+      visit "/queue"
+    end
+
+    it "should redirect to VSO's organizational task queue" do
+      expect(page.current_path).to eq(vso.path)
```
mdbenjam: This is a potential race condition. Checking for current_path doesn't wait. So we should check for content first. Also I don't think there's much value in doing this in a separate feature test. I would suggest just combining this and the one below.

I had a VSO test here: https://github.com/department-of-veterans-affairs/caseflow/blob/master/spec/feature/queue/ama_queue_spec.rb#L200 not sure if they overlap at all or can be combined.

spec/feature/queue/task_queue_spec.rb
```diff
@@ -77,4 +73,47 @@
...
+    end
+
+    it "should display task actions dropdown" do
+      case_details_link = page.find(:xpath, "//tbody/tr/td[1]/a")
+      case_details_link.click
+      expect(page).to have_content(COPY::CASE_SNAPSHOT_ACTION_BOX_TITLE)
+
+      # Marking the task as complete correctly changes the task's status in the database.
+      dropdown = page.find(".Select-control")
```
mdbenjam: I think it's a little more clear to do something like this:

```
find(".Select-control", text: "Dropdown name").click
find("div", class: "Select-option", text: "Option").click
```

That way we know what's clicked on.

spec/feature/queue/task_queue_spec.rb
```diff
@@ -77,4 +73,47 @@
...
+      dropdown.sibling(".Select-menu-outer").find("div[id$='--option-0']").click
+
+      complete_button = page.find("#button-next-button")
+      complete_button.click
+
+      expect(page).to have_content(COPY::TASK_MARKED_COMPLETE_NOTICE_TITLE)
+      expect(Task.find(vso_task.id).status).to eq("completed")
+    end
+
+    # it "should mark task complete when selecting the appropriate task action" do
```
mdbenjam: We should delete this if we want to leave it commented out.

PR #7091, opened 2018-09-21
adds judge assignment function

amprokop:
Connects https://github.com/department-of-veterans-affairs/caseflow/issues/7105

# Test plan
- Created 4 appeals with intake
- Saw that appeals were assigned round robin to judges
- Ran `unassigned_ramp_tasks` query in production and inspected the tasks to make sure they were correct


app/models/tasks/judge_task.rb
```diff
@@ -15,4 +17,64 @@ def previous_task
...
+
+    assign_judge_tasks_for_root_tasks(tasks)
+  end
+
+  def self.assign_judge_tasks_for_root_tasks(root_tasks)
+    root_tasks.each do |root_task|
+      Rails.logger.info("Assigning judge task for appeal #{root_task.appeal.id}")
+      task = create(appeal: root_task.appeal,
+                    parent: root_task,
+                    appeal_type: "Appeal",
```
lowellrex: Let's use `Appeal.name` here so this code will fail with a helpful message if we ever rename the `Appeal` class.

app/models/tasks/judge_task.rb
```diff
@@ -1,6 +1,8 @@
 class JudgeTask < Task
   validates :action, inclusion: { in: %w[assign review] }
 
+  include RoundRobinAssigner
```
lowellrex: Now that we're using `RoundRobinAssigner` in 3 places I think it might be time to address the behaviour we currently implement [when the previous CSS ID does not exist in the list of assignees](https://github.com/department-of-veterans-affairs/caseflow/blob/f0cde48dad8ffd63105abbba8d8e1a1013c3b27d/app/models/concerns/round_robin_assigner.rb#L25). Currently, we just restart at the beginning of the list but I think we should probably just start at a random index in that situation so we don't assign way more tasks to judges at the beginning of the list.

app/models/tasks/judge_task.rb
```diff
@@ -15,4 +17,64 @@ def previous_task
...
+  # This function to be manually run in production when we need to fetch all RAMP
+  # appeals that are eligible for assignment to judges, and assign them.
+  def self.assign_ramp_judge_tasks(dry_run: false, batch_size: 10)
+    # Find all root tasks with no children, that means they are not assigned.
+    tasks = unassigned_ramp_tasks[0..batch_size - 1]
+
+    if dry_run
+      Rails.logger.info("Dry run. Found #{unassigned_ramp_tasks.length} tasks to assign.")
+      evidence_count = unassigned_ramp_tasks.select { |task| task.appeal.docket_name == "evidence_submission" }.count
+      direct_review_count = unassigned_ramp_tasks.select { |task| task.appeal.docket_name == "direct_review" }.count
```
lowellrex: Can we move these strings (evidence_submission, direct_review, hearing) to a shared constants file? We use them on the front-end already as well so I think that'd be nice.

app/models/tasks/judge_task.rb
```diff
@@ -15,4 +17,64 @@ def previous_task
...
+    # If it's an evidence submission case, we need to wait until the
+    # evidence submission window is over
+    if task.appeal.docket_name == "evidence_submission"
+      return false if task.appeal.receipt_date > 90.days.ago
+    end
+
+    # If the task already has been assigned to a judge, or if it
+    # is a VSO task, it will have children tasks. We only want to
+    # assign tasks that have not been assigned yet.
+    task.children.empty?
```
lowellrex: Am I correct in reading this that we will not assign appeals to judges if that appeal has a VSO task attached to it, even if that VSO task has been completed? Is that a behaviour we will get rid of at some point down the line (when we are no longer manually running this code, for example)?

app/models/tasks/judge_task.rb
```diff
@@ -15,4 +17,64 @@ def previous_task
     fail Caseflow::Error::TooManyChildTasks, task_id: id if children_attorney_tasks.length > 1
     children_attorney_tasks[0]
   end
+
+  #:nocov:
```
lowellrex: What is the reason for not adding tests for this code?

amprokop: I look at this method as a script that requires manual handling rather than a true piece of our codebase. I'll be running the `dry_run` method to test the output every week. Additionally, we'll be phasing it out as we build case assignment. Due to all those factors, it doesn't seem worthwhile to unit test this.

app/models/tasks/judge_task.rb
amprokop: Yes, this method does not assign judge tasks for VSO appeals. We're not sure yet when we'll be assigning those tasks to judges, but not next week for sure.

app/models/tasks/judge_task.rb
amprokop: sure

app/models/tasks/judge_task.rb
amprokop: Cool, thanks for the info! I'm not particularly concerned about that quirk for the purposes of this script, so I'm not sure if it really fits in to this PR.

PR #7094, opened 2018-09-21
Adding Advanced on Docket UI

### LINKED ISSUE #6257: Editing AMA appeal data | Editing AOD status
When Veterans send in AOD motions for reasons other than age, Board users need to be able to edit the AOD field and mark the case AOD, so the case can be advanced on the docket later in the appeals process. 

The proposed workflow for this is as follows (work in progress) and included for context. The crux of this github issue is the mechanism for changing the AOD field, recognizing the Veteran is now AOD. 
 
- Veteran sends in Motion to Advance on the Docket
- Caseflow recognizes a document has been added to VBMS and routes the case to a Caseflow user to review the document
- Caseflow user reviews the document and recognizes it is a motion for AOD
  - If necessary, Caseflow user sends the motion via email to a judge to approve
- Caseflow user, in Case details, clicks to edit the AOD field
- Caseflow user updates AOD field to reflect that the Veteran is now AOD

Updated 8/30/2018:
- We're building an AOD table with the following fields (see #6274 )
  - `reason`
  - `created_by`
  - `created_at`
  - `veteran_id`
  - `active` (<- to enable a toggle back and forth? is this necessary?)

## Assumptions
- Editing the AOD field in Case details will suffice for October RAMP appeals pilot, rather than a more fully fledged flow.
- Editing the AOD field should be behind a feature flag, because not every user with access to Caseflow will be able to edit this.

## AC
- Create AOD editing feature flag (perhaps `decide_aod`)
- Users with AOD editing feature flag can select "Decide AOD Motion" from the Actions dropdown of case details.
- Create edit AOD screen
- User navigates to the edit AOD screen after clicking "Decide AOD Motion"
- Per mocks, user can grant or deny the AOD motion. 
- Per mocks, user can select an AOD reason:
   - age
   - serious illness
   - financial distress
   - other
- When a user clicks continue at the bottom of the page, the following fields are updated:
  - `created_by`
  - `created_at`

## Mocks
See https://github.com/department-of-veterans-affairs/caseflow/issues/6257#issuecomment-422108911

## Resources
- Edit AOD sketches [mural](https://app.mural.co/t/workqueue2001/m/workqueue2001/1536605643302/759a692551fff392640860e504d25842db09dec7)
- Full, legacy appeals Advance on Docket flow captured in this [Mural](https://app.mural.co/t/workqueue2001/m/workqueue2001/1505149828607/368eaa58e44ff29dabff797bd9b29ad52a232cb0) from last year's discovery.

mdbenjam:
Resolves #6257

### Description
This renames the `advance_on_docket_grants` to `advance_on_docket_motions`, it creates a UI to grant/deny these motions, and a backend to receive those requests.

We also create a higher ordered component called EditModalBase which enables us to quickly create modals that edit values on the appeal. It's the modal equivalent of DecisionReviewBase.

### Testing Plan
1. Go to case details for an AMA appeal
1. Start the AOD editing flow, and grant AOD
1. Ensure the appeal is marked as AOD


app/models/advance_on_docket_grant.rb
```diff
@@ -1,4 +0,0 @@
-class AdvanceOnDocketGrant < ApplicationRecord
```
mdbenjam: Changing the name

client/app/queue/CaseDetailsView.jsx
```diff
@@ -36,7 +38,10 @@ const horizontalRuleStyling = css({
 const PowerOfAttorneyDetail = ({ poa }) => <p>{poa.representative_type} - {poa.representative_name}</p>;
 
 class CaseDetailsView extends React.PureComponent {
-  componentDidMount = () => window.analyticsEvent(CATEGORIES.QUEUE_TASK, TASK_ACTIONS.VIEW_APPEAL_INFO);
+  componentDidMount = () => {
+    window.analyticsEvent(CATEGORIES.QUEUE_TASK, TASK_ACTIONS.VIEW_APPEAL_INFO);
+    this.props.resetErrorMessages();
```
mdbenjam: Error message should appear on the page their on. Before this change if you get an error then cancel the flow, the error never goes away! Not sure if we were intentionally showing errors on this page before...

client/app/queue/uiReducer/uiActions.js
```diff
@@ -63,7 +63,8 @@ const saveFailure = (resp: Object) => (dispatch: Dispatch) => {
   let responseObject = {
     errors: [{
       title: 'Error',
-      detail: 'There was an error processing your request.'
+      detail: 'There was an error processing your request. ' +
+        'Please retry your action and contact support if errors persist.'
```
mdbenjam: Giving the user an action item here.

client/app/queue/QueueApp.jsx
```diff
@@ -228,10 +234,18 @@ class QueueApp extends React.PureComponent<Props> {
...
+            path="/queue/appeals/:appealId/modal/advanced_on_docket_motion"
+            render={this.routedAdvancedOnDocketMotion} />
           <PageRoute
             exact
             path="/queue/appeals/:appealId"
             title="Case Details | Caseflow"
+            render={this.routedQueueDetailWithLoadingScreen} />
+          <PageRoute
+            exact
+            path="/queue/appeals/:appealId/modal/:modalType"
```
mdbenjam: Allow us to still match the case details screen even when there is a modal open.

client/app/queue/QueueApp.jsx
```diff
@@ -228,10 +234,18 @@ class QueueApp extends React.PureComponent<Props> {
             path="/queue/:userId/assign"
             title="Unassigned Cases | Caseflow"
             render={this.routedJudgeQueueList('assign')} />
+          <Route
```
mdbenjam: We use this partial match to display an edit modal.

client/app/styles/_commons.scss
```diff
@@ -454,7 +454,7 @@ svg title {
   margin: 0 auto;
   width: 490px;
   max-height: 100%;
-  overflow-y: auto;
+  overflow-y: visible;
```
mdbenjam: I want dropdowns to appear in modals even if they're too long for the modal height.

client/app/styles/_commons.scss
```diff
@@ -472,8 +472,13 @@ svg title {
   padding: .1rem;
   position: absolute;
   right: rem(10px);
+  top: rem(10px);
   width: 1.5rem;
 
+  &[type="button"] {
```
mdbenjam: Moving the `x` close button more into the corner.

app/models/person.rb
```diff
@@ -1,7 +1,11 @@
 class Person < ApplicationRecord
-  has_many :advance_on_docket_grants
+  has_many :advance_on_docket_motions
```
mdbenjam: We need to move these since Anya just made a new person table! 🎉

client/app/queue/QueueApp.jsx
mdbenjam: We currently don't load the case details here since they should already be up to date since the only way to get to this page is to start on the case details page. But in the future I want to improve the case details loading page so that if things are already loaded, it doesn't re-load them.

client/app/queue/components/EditModalBase.jsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import Modal from '../../components/Modal';
+
+export default function editModalBase(ComponentToWrap, title) {
+  class WrappedComponent extends React.Component {
```
mdbenjam: Here's our new higher order component. This should allow us to make some quick edit modals.

client/app/queue/utils.js
```diff
@@ -246,8 +246,8 @@ export const renderLegacyAppealType = ({ aod, type }: {aod: boolean, type: strin
   const cavc = type === 'Court Remand';
 
   return <React.Fragment>
-    {aod && <span><span {...redText}>AOD</span>, </span>}
     {cavc ? <span {...redText}>CAVC</span> : <span>{type}</span>}
+    {aod && <span><span {...redText}>AOD</span>, </span>}
```
mdbenjam: Swapping the order so AOD is last to make it clearer what is being edited.

client/app/styles/_commons.scss
```diff
@@ -502,6 +507,7 @@ svg title {
 .cf-modal-controls {
   position: relative;
   bottom: 0;
+  padding-bottom: 3rem;
```
mdbenjam: Need to get the submit/cancel buttons to show up in the modal. Changing the overflow-y type made them hang off the modal.

db/migrate/20180921174854_add_granted_to_advance_on_docket_motions.rb
```diff
@@ -0,0 +1,6 @@
+class AddGrantedToAdvanceOnDocketMotions < ActiveRecord::Migration[5.1]
+  def change
```
mdbenjam: Changing the name since we're adding a granted field.

client/app/queue/AdvancedOnDocketMotionView.jsx
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  render = () => {
+    const {
+      error,
+      highlightFormItems
+    } = this.props;
+
+    return <React.Fragment>
+      {error && <Alert type="error" title={error.title} message={error.detail} />}
```
dannysellers: Is there any reason this shouldn't live inside `editModalBase`?

Also, the vertical alignment of this Alert is a bit off (it has a top margin of `1.5rem` by default)

<details><summary>screenshot</summary>

</details>

client/app/queue/components/EditModalBase.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+      this.wrappedComponent.submit().then(() => {
+        this.setState({ loading: false });
+        this.closeHandler();
+      }, () => {
+        this.setState({ loading: false });
+      });
+    }
+
+    render = () => <React.Fragment>
```
dannysellers: Nit: Not sure we need the `<React.Fragment>` here

client/app/queue/components/EditModalBase.jsx
```diff
@@ -0,0 +1,49 @@
...
+    constructor(props) {
+      super(props);
+
+      this.state = { loading: false };
+    }
+
+    getWrappedComponentRef = (ref) => this.wrappedComponent = ref;
+
+    closeHandler = () => {
+      this.props.history.push(`/queue/appeals/${this.props.appeal.externalId}`);
```
dannysellers: I think this should probably be `replace` instead of `push`. The back button does successfully reopen the modal, which is good, but I suspect the user would find that confusing / if they want to go back to `/queue` after editing AOD, they now have to click back 3 times (Case Details -> AOD modal -> Case Details -> `/queue`)

app/controllers/advance_on_docket_motions_controller.rb
```diff
@@ -0,0 +1,19 @@
+class AdvanceOnDocketMotionsController < ApplicationController
+  def create
+    appeal.claimants.first.advance_on_docket_motions.create(
```
dannysellers: I'm getting a 500 here:
```
NoMethodError in AdvanceOnDocketMotionsController#create
undefined method `advance_on_docket_motions' for #<Claimant:0x00007ff87d947000>
Did you mean?  advanced_on_docket
```

mdbenjam: Whoops yeah I'll fix that!

client/app/queue/AdvancedOnDocketMotionView.jsx
mdbenjam: Good call. Moved it into EditModalBase and added some bottom padding.

client/app/queue/components/EditModalBase.jsx
mdbenjam: Solid catch. I started using goBack so that it's like they never went to the modal.

Issue #7095, opened 2018-09-21
Don't include ratings with no rating issues


leikkisa:

shanear:
We're gonna have the intake folks at BVA try these again and follow up to see if it worked.

shanear:
This fixed the issue they were having in prod. Good enough for me

PR #7096, opened 2018-09-21
Do not return ratings with no issues

leikkisa:
connects #7095


spec/models/rating_spec.rb
```diff
@@ -43,6 +43,21 @@
...
+      let(:issues) do
+        nil
+      end
+
+      it "should have no issues" do
+        is_expected.to match(
+          participant_id: rating.participant_id,
+          profile_date: rating.profile_date,
+          promulgation_date: rating.promulgation_date,
+          issues: []
```
leikkisa: This isn't really testing the code because this is a result of the generator.

shanear: fixed this issue. Kinda janky, but works

Issue #7081, opened 2018-09-20
Intake | Don't require city and state for military addresses


shanear:
We've been getting reports that users can't process veterans with military addresses because we are requiring that they have city and state set. We should get rid of this requirement for military addresses.

## Validation
1. Find a veteran with a military address and no city/state (for example, a non-US military address)
2. Confirm that we can process an intake for this veteran

Or

1. Ask Katlin to reattempt intake with one of these veterans

kencheeto:
@shanear you mentioned that there are examples of this, should I look for those in Sentry once I confirm access?

shanear:
Yeah, I think the best thing to do would be to inform Katlin that they can try intaking forms for those vets again and see if it's working for them, (since it doesn't actually throw an exception, it was just incorrectly requiring them to enter a city and state)

kencheeto:
@shanear okay! I'm looking for "Katlin" but I don't see them in Slack. How do I find them?

shanear:
She’s at AMO and not in Slack. I notified her via email. I'll CC you if/when she responds.

kencheeto:
okay, thanks!

shanear:
validated in Prod this is working! Katlin didn't get the email but told us in the AMO meething.

kencheeto:
🎉 thanks!

PR #7097, opened 2018-09-22
allow nil city and state for veterans with military addresses

kencheeto:
Connects #7081 

### Description
Allows veterans fetched from BGS to be created with nil city and state if they have a military address.

### Acceptance Criteria
- [x] Code compiles correctly
- [x] Unit tested
- [ ] Create and process a suitable veteran locally

### Testing Plan
1. Find/create a veteran with a military address (`military_postal_type_code` is present) and a blank `city` and/or `state` in BGS
2. Process a claim for this veteran, ensure that there is no error


app/models/veteran.rb
```diff
@@ -38,6 +39,10 @@ def country_requires_zip?
     COUNTRIES_REQUIRING_ZIP.include?(country && country.upcase)
   end
 
+  def state_is_required?
+    military_address? ? false : country_requires_state?
```
shanear: This could be simplified to:

```
!military_address? && country_requires_state?
```

PR #7109, opened 2018-09-24
re-enable guard-rspec since we are on ruby 2.5

kencheeto:
### Description
guard-rspec is nice for adding `focus: true` to tests then running `bundle exec guard` to automatically rerun tests when watched files change.

### Acceptance Criteria
- [x] `bundle exec guard` will rerun tests on file changes

### Testing Plan
N/A


Gemfile
```diff
@@ -119,12 +119,12 @@ group :stubbed, :test, :development, :demo do
   gem "bundler-audit"
 
   # Testing tools
+  gem "capybara"
```
kencheeto: reorganizing these bc the commented `guard-rspec` line was signaling a new group of gems to rubocop, which enforces alpha ordering of gems within a 'group' as delimited by comments or newlines.

Guardfile
```diff
@@ -2,8 +2,8 @@
 # More info at https://github.com/guard/guard#readme
 
 ## Uncomment and set this to only include directories you want to watch
-# directories %w(app lib config test spec features) \
-#  .select{|d| Dir.exists?(d) ? d : UI.warning("Directory #{d} does not exist")}
+directories(%w[app lib config spec])\
```
kencheeto: without this line, there's an error that some node_modules directory is already being watched. I don't know the `directories` call is required (shouldn't guard only watch the files specified in `watch` clauses?) but it does get rid of the error.

pkarman:
Maybe worth a mention in the README?

kencheeto:
good tip, thanks @pkarman !

Issue #6966, opened 2018-09-13
AMA Intake | Create an "Add issues" landing page


suzchap:
## Story
For any AMA forms, when a user needs to add issues, they will be walked through a multi-screen selection process to help them make the best choice. This landing page will be their starting point. This will address a number of usability issues and set the stage for future additions to accommodate edge cases.  

## Technical notes
Even though we are no longer showing the old Add Issues UI, we will still be using parts of it in follow along stories (see https://github.com/department-of-veterans-affairs/caseflow/issues/6981 and https://github.com/department-of-veterans-affairs/caseflow/issues/6988) so it's probably good to keep a lot of the state structure and components around. The old feature tests can go though.

## Acceptance criteria
- After a user submits a review page, they are shown this "Add Issues" landing page. 
- "Complete intake" button is disabled
- "Cancel intake" link opens the cancel model (same as it works right now),
- "Add issues" button doesn't need to do anything yet (that will be in another ticket)
- The contents of the page show a summary of every field that was completed earlier in the process. Fields are displayed in the order that they appear in the intake flow and will vary for each form type as shown below.

## Copy
**NoD**
- Form ––––– `Form name (Form #)`
- Veteran –––– `Veteran Name (ID # XXXXXXXX)`
- Form receipt date –––– `MM/DD/YYYY`
- Review option –––– `Hearing`, `Direct Review`, or `Evidence Submission`
- Claimant –––– `if Claimant = veteran, display Veteran name; if Claimant = not Veteran, display Claimant name (payee code = XXXX)`
- Requested issues –––– `("+ Add issue" button only / no content for this ticket)`

**SC / Non-Comp**
- Form ––––– `Form name (Form #)`
- Veteran –––– `Veteran Name (ID # XXXXXXXX)`
- Form receipt date –––– `MM/DD/YYYY`
- Benefit type ––– `LoB selected`
- Requested issues –––– `("+ Add issue" button only / no content for this ticket)`

**HLR / Non-Comp**
- Form ––––– `Form name (Form #)`
- Veteran –––– `Veteran Name (ID # XXXXXXXX)`
- Form receipt date –––– `MM/DD/YYYY`
- Benefit type ––– `LoB selected`
- Informal conference –––– `Yes | No`
- Review by same office –––– `Yes | No`
- Requested issues –––– `("+ Add issue" button only / no content for this ticket)`

**SC / Comp**
- Form ––––– `Form name (Form #)`
[...]

PR #7110, opened 2018-09-24
Add issues landing page

leikkisa:
connects #6966

Adds a page at /issues/add_issues which is not yet linked in the main flow.

Does not implement progress bar or add button.

NoD
image

HLR Comp
image

HLR non-comp
image

SC Comp
image

SC non-comp
image


spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -476,4 +480,43 @@
...
     expect(page).to have_content("Request for Higher-Level Review (VA Form 20-0988) has been processed.")
   end
+
+  context "For new Add Issues page" do
+    def check_row(label, text)
+      row = find("tr", text: label)
+      expect(row).to have_text(text)
+    end
+
+    let!(:timely_ratings) do
```
kencheeto: should this be a `before` ?

spec/feature/intake/supplemental_claim_spec.rb
```diff
@@ -348,26 +348,30 @@
     expect(page).to have_current_path("/intake/review_request")
   end
 
-  it "Allows a Veteran without ratings to create an intake" do
+  def start_sc(test_veteran, is_comp = true)
```
kencheeto: maybe an old habit but i prefer `start_supplemental_claim`

spec/feature/intake/higher_level_review_spec.rb
jiahuang: Good thought. I'm fine either way, but I think in the next round of tickets we'll need to reference these timely_ratings in order to test the add modal properly.

PR #7111, opened 2018-09-24
async Intake jobs

pkarman:
connects #6941 

### Description
Creates a new class `ClaimReviewProcessJob` that calls `claim_review.process_end_product_establishments!`. The new job can be run asynchronously, to background VBMS interactions.

Similarly, the class `RequestIssuesUpdateJob` allows async running of `claim_review.process_end_product_establishments!`.

New columns added to `:ramp_elections, :ramp_refilings, :higher_level_reviews, :supplemental_claims` to track async attempts.

The `SyncReviewsJob` class has been refactored to check for Claim Reviews and Request Issues Updates requiring processing and queue a job for each.

The classes `SupplementalClaimIntake` and `HigherLevelReviewIntake` have been refactored into a common superclass `ClaimReviewIntake`.

### Acceptance Criteria
- [x] Code compiles correctly
- [x] Tests continue to pass

### Testing Plan

Manual observation in UAT.


app/jobs/complete_intake_job.rb
```diff
@@ -0,0 +1,33 @@
+# This job runs Intake.complete! steps
+class CompleteIntakeJob < CaseflowJob
```
shanear: I think we just want to do the VBMS establishment processing stuff in this step. That way we aren't relying on redis to hold on to critical data.

shanear: Additionally, only `detail. process_end_product_establishments!` is idempotent. All the other stuff isn't necessarily.

app/models/end_product_establishment.rb
```diff
@@ -140,10 +140,9 @@ def sync!
 
       sync_source!
     end
-
-    # TODO: This is sort of janky. Let's rethink the error handling logic here
+  rescue EstablishedEndProductNotFound => e
+    raise e
```
shanear: Nice! Does is this exception ignored by the rescue on line 145?

app/models/intake.rb
```diff
@@ -242,4 +242,8 @@ def find_or_build_initial_detail
   def build_issues(request_issues_data)
     request_issues_data.map { |data| detail.request_issues.from_intake_data(data) }
   end
+
+  def run_async?
```
shanear: nice way to make this way less annoying 😄

app/models/end_product_establishment.rb
pkarman: I don't think re-throwing an exception inside a `rescue` chain will trigger any other links in the chain. The tests still pass in `spec/models/end_product_establishment_spec.rb` anyway.

app/jobs/complete_intake_job.rb
pkarman: I started with just making a job for `process_end_product_establishments!` but the start/end steps around it suggested to me it was being tried as part of a "transaction" of some kind. I.e. we call `intake.complete_with_status!(:success)` just afterwards. If `process_end_product_establishments!` were performed later, should we be marking the `intake` as successfully complete? We don't yet know if it is.

app/jobs/complete_intake_job.rb
```diff
@@ -0,0 +1,33 @@
...
+    RequestStore.store[:current_user] = user if user
+
+    @intake = intake
+    perform_intake(request_issues)
+  end
+
+  private
+
+  def perform_intake(request_issues)
+    # TODO: wrap this all in a transaction?
```
pkarman: these steps appear to be transaction-like, since we flag the status as "successful" at the end, assuming no exceptions interrupt the flow.

should we make it a "real transaction"?

shanear: Since we can't transactionalize the VBMS/BGS calls, we basically have two steps to completing the HLR/SC intake.

1) Write all data to our DB.
2) Perform the necessary VBMS/BGS requests to represent that data in those dependencies.

All actions in step 1 should be a transaction I think. Actions in step 2 can't be a transaction because of the nature of the services involved. This is why we are creating a retry loop.

app/jobs/complete_intake_job.rb
shanear: the start/end steps would still be relevant even if some parts of the job are processed async, because they would tell us when the intake was submitted, and when it was finally processed (step 2 was completed successfully).

They'd also help us find submitted, but not processed intakes

spec/jobs/sync_reviews_job_spec.rb
```diff
@@ -29,6 +48,17 @@
       end
     end
 
+    context "where there are claim reviews awaiting processing" do
+      it "ignores completed and older expired reviews" do
+        expect(EndProductSyncJob).to receive(:perform_later).twice.and_return(true)
```
pkarman: probably can remove this line if it passes at CI. probably a local issue.

app/jobs/sync_reviews_job.rb
```diff
@@ -25,4 +35,12 @@ def perform(args = {})
...
     end
   end
+
+  def preform_claim_review_processing(limit)
+    [HigherLevelReview, SupplementalClaim].each do |klass|
+      klass.requires_processing.limit(limit).each do |claim_review|
+        ClaimReviewProcessJob.perform_later(claim_review)
+      end
+    end
+  end
```
shanear: love this refactor 💯

app/models/higher_level_review_intake.rb
```diff
@@ -40,11 +40,16 @@ def review_errors
...
-    detail.update!(establishment_submitted_at: Time.zone.now)
-    detail.process_end_product_establishments!
-    complete_with_status!(:success)
+    req_issues = request_params[:request_issues] || []
+    transaction do
+      intake.start_completion!
+      detail.request_issues.destroy_all unless detail.request_issues.empty?
+      detail.create_issues!(build_issues(req_issues))
+      detail.requires_processing!
+      unless run_async?
```
shanear: It might be best to kick off a ClaimReviewProcessJob right here so there is as little of a lag as possible. Then use `attempted_previously` as the scope to find ones that need a follow up attempt.

db/migrate/20180926182000_add_claim_establishment_async_fields.rb
```diff
@@ -0,0 +1,13 @@
+class AddClaimEstablishmentAsyncFields < ActiveRecord::Migration[5.1]
+  def change
+    ActiveRecord::Base.connection.execute "SET statement_timeout = 1800000" # 30 minutes
+
+    [:ramp_elections, :ramp_refilings, :higher_level_reviews, :supplemental_claims].each do |tbl|
+      add_column tbl, :establishment_attempted_at, :datetime
+      add_column tbl, :establishment_error, :string
+    end
```
shanear: does this kind of logic work with `change`? did you attempt to migrate and rollback this migration?

spec/jobs/sync_reviews_job_spec.rb
```diff
@@ -12,6 +12,25 @@
       create(:end_product_establishment, last_synced_at: nil, established_at: 4.days.ago)
     end
 
+    let!(:higher_level_review_requiring_processing) do
+      hlr = create(:higher_level_review)
+      hlr.requires_processing!
+      hlr
```
shanear: a fancy functional way to achieve this effect is using `tap`

```
create(:higher_level_review).tap { |hlr| hlr.requires_processing! }
```

app/models/claim_review.rb
```diff
@@ -6,6 +6,65 @@ class ClaimReview < AmaReview
...
+
+    def expired_without_processing
+      where(establishment_processed_at: nil)
+        .where("establishment_submitted_at <= ?", REQUIRES_PROCESSING_WINDOW_DAYS.days.ago)
+        .order("establishment_submitted_at ASC")
+    end
+  end
+
+  def requires_processing!
+    update!(establishment_submitted_at: Time.zone.now, establishment_processed_at: nil)
```
leikkisa: I think we want to keep the establishment_submitted_at set by when the user presses the "Establish EP" button, right?  In which case we wouldn't update this field.

app/models/claim_review.rb
```diff
@@ -6,6 +6,65 @@ class ClaimReview < AmaReview
...
+    update!(establishment_processed_at: Time.zone.now) unless processed?
+  end
+
+  def attempted!
+    update!(establishment_attempted_at: Time.zone.now)
+  end
+
+  def processed?
+    !!establishment_processed_at
+  end
```
shanear: I really love all these clear, simple methods 👍

app/models/claim_review.rb
leikkisa: Nevermind, I'm guessing this is in the new model.

db/migrate/20180926182000_add_claim_establishment_async_fields.rb
pkarman: I didn't think it would, but rails continually surprises me:
```
[2018-09-26 18:41:21 -0500] Migrating to AddClaimEstablishmentAsyncFields (20180926182000)
[2018-09-26 18:41:21 -0500]    (0.4ms)  BEGIN
== 20180926182000 AddClaimEstablishmentAsyncFields: reverting =================
[2018-09-26 18:41:21 -0500]    (0.5ms)  SET statement_timeout = 1800000
[2018-09-26 18:41:21 -0500]    (0.4ms)  SET statement_timeout = 30000
-- remove_column(:supplemental_claims, :establishment_error, :string)
[2018-09-26 18:41:21 -0500]    (0.9ms)  ALTER TABLE "supplemental_claims" DROP "establishment_error"
   -> 0.0011s
-- remove_column(:supplemental_claims, :establishment_attempted_at, :datetime)
[2018-09-26 18:41:21 -0500]    (0.6ms)  ALTER TABLE "supplemental_claims" DROP "establishment_attempted_at"
   -> 0.0007s
-- remove_column(:higher_level_reviews, :establishment_error, :string)
[2018-09-26 18:41:21 -0500]    (0.7ms)  ALTER TABLE "higher_level_reviews" DROP "establishment_error"
   -> 0.0008s
-- remove_column(:higher_level_reviews, :establishment_attempted_at, :datetime)
[2018-09-26 18:41:21 -0500]    (0.5ms)  ALTER TABLE "higher_level_reviews" DROP "establishment_attempted_at"
   -> 0.0007s
-- remove_column(:ramp_refilings, :establishment_error, :string)
[2018-09-26 18:41:21 -0500]    (0.6ms)  ALTER TABLE "ramp_refilings" DROP "establishment_error"
   -> 0.0007s
-- remove_column(:ramp_refilings, :establishment_attempted_at, :datetime)
[2018-09-26 18:41:21 -0500]    (0.5ms)  ALTER TABLE "ramp_refilings" DROP "establishment_attempted_at"
   -> 0.0006s
-- remove_column(:ramp_elections, :establishment_error, :string)
[2018-09-26 18:41:21 -0500]    (0.5ms)  ALTER TABLE "ramp_elections" DROP "establishment_error"
   -> 0.0007s
-- remove_column(:ramp_elections, :establishment_attempted_at, :datetime)
[2018-09-26 18:41:21 -0500]    (0.5ms)  ALTER TABLE "ramp_elections" DROP "establishment_attempted_at"
   -> 0.0006s
== 20180926182000 AddClaimEstablishmentAsyncFields: reverted (0.0108s) ========

[2018-09-26 18:41:21 -0500]   SQL (0.6ms)  DELETE FROM "schema_migrations" WHERE "schema_migrations"."version" = $1  [["version", "20180926182000"]]
```

pkarman:
thanks @shanear -- I agree about the naming of things. I made some changes in 65785f9 that I hope clarify better.

app/models/concerns/claim_review_completeable.rb
```diff
@@ -0,0 +1,25 @@
+module ClaimReviewCompleteable
```
shanear: any reason why this is a separate concern and not part of `ClaimReview`?

pkarman: yeah I'm unhappy with that concern and have refactored (again) locally. will push later today.

app/models/claim_review.rb
```diff
@@ -2,10 +2,71 @@
 # higher level review as defined in the Appeals Modernization Act of 2017
 
 class ClaimReview < AmaReview
+  include Asyncable
+
   has_many :end_product_establishments, as: :source
 
   self.abstract_class = true
 
+  # establishment_submitted_at - when our db is ready to push to exernal services
```
pkarman: a lot of these class methods for scoping are identical in `ClaimReview` and `RequestIssuesUpdate` -- it would be nice to refactor into the `Asyncable` concern in future.

pkarman: done!

pkarman:
@shanear the addition of `RequestIssuesUpdateJob` made this PR bigger than I would have liked, but I was able to use the `Asyncable` concern to DRY up some code re-use. I'm happy enough with it. Would appreciate one more look through from you.

config/environments/test.rb
```diff
@@ -73,4 +73,6 @@
 
   # Allow health check to pushgateway
   ENV["ENABLE_PUSHGATEWAY_HEALTHCHECK"] = "true"
+
+  config.active_job.queue_adapter = :inline
```
pkarman: w/o this the tests for `spec/jobs/sync_reviews_job_spec.rb` and `EndProductSyncJob` were failing for me locally with errors about missing AWS creds.

alternately, we could set
```
  # set to true to create queues and override the sqs endpiont
  config.sqs_create_queues = true
  config.sqs_endpoint = "http://localhost:4576"
```
in `config/environments/test.rb` similar to how they are in `development.rb`.

Not sure why they work in CI unless AWS creds are set there?

app/jobs/request_issues_update_job.rb
```diff
@@ -0,0 +1,21 @@
...
+    request_issues_update.attempted!
+
+    request_issues_update.review.process_end_product_establishments!
+
+    request_issues_update.removed_issues.each do |request_issue|
+      request_issue.end_product_establishment.remove_contention!(request_issue)
+    end
+
+    request_issues_update.processed!
+  end
```
shanear: awesome. I've been wanting to do this 💯

app/models/request_issues_update.rb
```diff
@@ -19,9 +22,14 @@ def perform!
         before_request_issue_ids: before_issues.map(&:id),
         after_request_issue_ids: after_issues.map(&:id)
       )
+      submit_for_processing!
```
shanear: This is a pattern i've been conflicted about for a while. Since `submit_for_processing!` is just an `update!` statement under the hood, is it better to combine `submit_for_processing!` into the update statement on 21? which would save a DB call, or just use `submit_for_processing!`? which makes it a little more DRY.

I think I lean towards how you are doing it, because I'm biased towards the ruby/rails ideology of readability over performance haha, but I can see the other way.

config/environments/test.rb
shanear: could you ask #appeals-engineering about this? I feel like the tests should be able to work without this. But I don't know enough about the mechanism we developed there. I feel like we stubbed the aws stuff with docker.

app/models/claim_review_intake.rb
```diff
@@ -0,0 +1,52 @@
+class ClaimReviewIntake < Intake
```
shanear: We've needed this class for a while now lol. Nice work 👍

app/jobs/request_issues_update_job.rb
shanear: I think next steps here are to put a lot of this logic directly into `RequestIssuesUpdate`, but that is for another PR.

app/models/request_issues_update.rb
pkarman: I think since it is all wrapped in a transaction, that the normal transaction in `update!` is ignored and so the performance effect is negligible. the performance hit in dbs (in my experience) is on the `COMMIT`, and there should be only one here. If it weren't wrapped in a transaction already, I could see merging the 2 updates into 1 (though I still prefer the readability, like you say).

app/models/claim_review_intake.rb
```diff
@@ -0,0 +1,52 @@
...
+  def review_errors
+    detail.errors.messages
+  end
+
+  def complete!(request_params)
+    return if complete? || pending?
+    complete_claim_review_async(request_params)
+  end
+
+  def complete_claim_review_async(request_params)
```
shanear: could this be a private method?

app/models/concerns/asyncable.rb
```diff
@@ -0,0 +1,87 @@
+# Mixin module to apply to an ActiveRecord class, to make it easier to process via
+# an ActiveJob and retry it beyond the retry logic of ActiveJob.
+# This becomes necessary when a Job has multiple external service calls, each of
+# which may fail and cause retries beyond the "normal" retry window.
+# See ClaimReview and RequestIssuesUpdate e.g.
+module Asyncable
```
shanear: I like this abstraction a lot.

pkarman:
thanks @shanear -- going to merge this as-is and address some of the changes in my follow on PR for #6942

PR #7113, opened 2018-09-24
Case timeline

### LINKED ISSUE #6856: Lightweight Case timeline for RAMP appeals pilot
Let's try to design and implement a lightweight Case timeline for RAMP appeals pilot (does not need to be ready exactly on Oct 1).

For AMA appeals only.

### AC
- Create another "tab" in Case details for Case timeline
- Case timeline elements:
   - NOD date
   - Decision date

mdbenjam:
Resolves #6856 

### Description
This adds a basic case timeline for legacy and ama appeals.

There's a lot to add here, but this is the starting place for the pilot.
screen shot 2018-09-25 at 11 08 28 am

### Testing Plan
1. Go to the case details screen and scroll to the bottom!


app/models/serializers/work_queue/legacy_appeal_serializer.rb
```diff
@@ -89,4 +89,11 @@ class WorkQueue::LegacyAppealSerializer < ActiveModel::Serializer
   attribute :docket_name do
     "legacy"
   end
+
+  attribute :events do
```
mdbenjam: In the future, I want to move this to use appeal status code. This is just a simple start.

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -92,4 +93,10 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
   attribute :caseflow_veteran_id do
     object.veteran ? object.veteran.id : nil
   end
+
+  attribute :events do
+    {
+      nod_receipt_date: object.receipt_date
```
mdbenjam: In the future I want to move to a separate function that gets the events of the ama appeal. We'll need to move this in parity with how legacy appeal events are constructed. This will enable us to get ama appeals to work with appeal status.

client/app/queue/CaseTimeline.jsx
```diff
@@ -0,0 +1,77 @@
...
+  verticalAlign: 'top',
+  padding: '3px'
+});
+
+export default class CaseTimeline extends React.PureComponent {
+  getEventRow = ({ title, pendingTitle, date }, lastRow) => {
+    const formattedDate = date ? moment(date).format('MM/DD/YYYY') : null;
+    const eventImage = date ? <GreenCheckmark /> : <GrayDot />;
+
+    return <tr key={title}>
```
mdbenjam: This is a pretty special table without any borders, feels reasonable to not use the table component.

Issue #7108, opened 2018-09-24
Modifier being passed incorrectly for HLR and SC


shanear:
When I try and establish an HLR or SC in UAT, I'm getting the "something went wrong" error. Looking into sentry, it looks like we are passing an entire array of modifiers into the VBMS request, which is incorrect.

![screen shot 2018-09-24 at 2 17 50 pm](https://user-images.githubusercontent.com/1596259/45970510-bc847680-c004-11e8-9187-f6cba0d4e651.png)

https://sentry.uat.ds.va.gov/sentry/caseflow-uat/issues/959/

shanear:
![screen shot 2018-09-28 at 12 09 12 pm](https://user-images.githubusercontent.com/1596259/46220223-75092d80-c317-11e8-9559-4ab09b2ebeed.png)
got there

PR #7115, opened 2018-09-24
Raise error when there are no available modifiers

shanear:
This was happening in UAT, and it was very painful to debug. Putting this in as a stand in right now.

connects #7108

### DIFF
--- a/spec/models/end_product_establishment_spec.rb
+++ b/spec/models/end_product_establishment_spec.rb
@@ -210,6 +210,21 @@ describe EndProductEstablishment do
       end
     end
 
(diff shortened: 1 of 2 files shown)


spec/models/end_product_establishment_spec.rb
```diff
@@ -210,6 +210,21 @@
       end
     end
 
+    context "when there are no open modifiers" do
+      before do
+        %w[030 031 032].each do |modifier|
```
leikkisa: You should use up 030 through 039 before you won't have any available

Issue #7138, opened 2018-09-25
Fix Edit Issues page in UAT


shanear:
The edit issues page is broken for HLR and SC in UAT. Investigating why.

```
<return>
  <errorLevel>0</errorLevel>
  <anyRatingsWithinRangeFlag>Y</anyRatingsWithinRangeFlag>
  <combinedPercentCurrentList>
    <combinedPercentCurrent>
      <beneficiaryParticipantID>600233666</beneficiaryParticipantID>
      <combinedPctWithoutRECA>30</combinedPctWithoutRECA>
      <combinedPctWithoutSeverance>30</combinedPctWithoutSeverance>
      <combinedPctWithoutTort>30</combinedPctWithoutTort>
      <combinedPctWithoutWorkersComp>30</combinedPctWithoutWorkersComp>
      <compensationServicePeriodType>4</compensationServicePeriodType>
      <effectiveDate>2017-11-01T00:00:00-05:00</effectiveDate>
      <notServiceConnBilateralValue>0.0</notServiceConnBilateralValue>
      <notServiceConnCombinedPct>30</notServiceConnCombinedPct>
      <recaWithholdingPct>0</recaWithholdingPct>
      <serviceConnBilateralPercent>0.0</serviceConnBilateralPercent>
      <serviceConnCombinedPct>30</serviceConnCombinedPct>
      <severanceWithholdingPct>0</severanceWithholdingPct>
      <tortWithholdingPct>0</tortWithholdingPct>
      <veteranParticipantID>600233666</veteranParticipantID>
      <workersCompWithholdingPct>0</workersCompWithholdingPct>
    </combinedPercentCurrent>
  </combinedPercentCurrentList>
  <differencesFound>N</differencesFound>
  <earliestDependencyEligibilityDate>2017-11-01T00:00:00-05:00</earliestDependencyEligibilityDate>
  <multipleRatingsFlag>N</multipleRatingsFlag>
  <prfilDt>2018-07-27T15:43:49-05:00</prfilDt>
  <promulgationDt>2018-07-30T07:25:52-05:00</promulgationDt>
  <ptcpntVetID>600233666</ptcpntVetID>
  <RatingDetails>
    <errorLevel>0</errorLevel>
    <EffectiveDt>2017-11-01T00:00:00-05:00</EffectiveDt>
    <scPercent>30</scPercent>
  </RatingDetails>
  <ratingProfileList>
    <ratingProfile>
[...]

PR #7139, opened 2018-09-25
Make rating_issue_profile_date a datetime.

shanear:
Well, this is awkward. Profile date is not a date. Saving it as a date in our database causes matching request issues to rating issues to break in the edit page because the profile dates don't match (one is a date, one is a datetime)

connects #7138


spec/feature/intake/edit_spec.rb
```diff
@@ -29,7 +29,7 @@
     Generators::Rating.build(
       participant_id: veteran.participant_id,
       promulgation_date: receipt_date + 1.day,
-      profile_date: receipt_date + 4.days,
+      profile_date: 24.days.ago,
```
shanear: This causes tests to fail before the migration.

db/schema.rb
```diff
@@ -569,7 +569,7 @@
     t.string "review_request_type"
     t.bigint "review_request_id"
     t.string "rating_issue_reference_id"
-    t.date "rating_issue_profile_date"
+    t.datetime "rating_issue_profile_date"
```
jiahuang: Should there be a migration file in here? Thought the schema isn't used in upper envs

shanear: good call 😄 

i committed it

leikkisa:
Oh, interesting, good catch.

PR #7145, opened 2018-09-26
Add decisions table

lowellrex:
Connects #7148.

We will be needing the `decisions` table to hold information for AMA appeals we will be getting from the IDT API. This PR creates that table but does not yet include any logic around putting things into that table.

### DIFF
--- a/db/migrate/20180926143801_create_decisions.rb
+++ b/db/migrate/20180926143801_create_decisions.rb
@@ -1,7 +1,7 @@
 class CreateDecisions < ActiveRecord::Migration[5.1]
   def change
     create_table :decisions do |t|
-      t.belongs_to :request_issue
+      t.belongs_to :appeal
       t.string :citation_number
       t.date :decision_date
       t.string :redacted_document_location
--- a/db/migrate/20180926143801_create_decisions.rb
+++ b/db/migrate/20180926143801_create_decisions.rb
@@ -7,5 +7,7 @@ class CreateDecisions < ActiveRecord::Migration[5.1]
       t.string :redacted_document_location
       t.timestamps null: false
     end
+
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -637,6 +647,13 @@
     t.index ["appeal_type", "appeal_id"], name: "index_special_issue_lists_on_appeal_type_and_appeal_id"
   end
 
+  create_table "staff_field_for_organizations", force: :cascade do |t|
```
mdbenjam: I think your DB already had the staff field table in it so it put it in schema.

db/migrate/20180926143801_create_decisions.rb
```diff
@@ -0,0 +1,11 @@
+class CreateDecisions < ActiveRecord::Migration[5.1]
+  def change
+    create_table :decisions do |t|
+      t.belongs_to :request_issue
```
amprokop: Why "request_issue", instead of "appeal"?

lowellrex: Copy/paste error! Great catch!

db/schema.rb
```diff
@@ -187,6 +187,16 @@
     t.integer "request_issue_id"
   end
 
+  create_table "decisions", force: :cascade do |t|
+    t.bigint "request_issue_id"
+    t.string "citation_number"
```
amprokop: potentially we could add a uniqueness constraint on this now, or do it when we implement the validation logic

lowellrex: "This" being the citation number?

mdbenjam: That's a good point, we should put the unique constraint on the citation number on the DB level.

PR #7154, opened 2018-09-26
Anya/7142 7143 idt addons

aroltsch:
1. Adding SSN to the appeals details response.
2. Adding days waiting to the appeals list response
2. Moving assigned_by and documents info to the appeals list


app/controllers/idt/api/v1/appeals_controller.rb
```diff
@@ -12,9 +12,11 @@ class Idt::Api::V1::AppealsController < Idt::Api::V1::BaseController
   end
 
   def list
-    appeals = file_number ? appeals_by_file_number : appeals_assigned_to_user
-
-    render json: json_appeals(appeals)
+    if file_number.present?
+      render json: json_appeals(appeals_by_file_number)
+    else
+      render json: { data: json_tasks(tasks_assigned_to_user) }
```
aroltsch: Instead of appeals, we will use tasks serializer because now we need tasks information for this endpoint.

app/controllers/idt/api/v1/appeals_controller.rb
```diff
@@ -63,20 +58,19 @@ def appeals_by_file_number
...
-    appeal_details = ActiveModelSerializers::SerializableResource.new(
+    ActiveModelSerializers::SerializableResource.new(
       appeal,
       serializer: ::Idt::V1::AppealDetailsSerializer,
       include_addresses: Constants::BvaDispatchTeams::USERS[Rails.current_env].include?(user.css_id),
       base_url: request.base_url
     ).as_json
+  end
 
-    assigned_by_name, documents = appeal.is_a?(LegacyAppeal) ? legacy_appeal_details : ama_appeal_details
```
aroltsch: I will double check with Paul if it is okay to remove from the appeals details.

app/models/serializers/idt/v1/task_serializer.rb
```diff
@@ -0,0 +1,42 @@
+class Idt::V1::TaskSerializer < ActiveModel::Serializer
+  def id
+    appeal.is_a?(LegacyAppeal) ? appeal.vacols_id : appeal.uuid
+  end
+
+  def appeal
+    object.appeal
+  end
+
+  attribute :appeal_type
```
aroltsch: Tiny problem here is that we can't override `type` in the response so now it will have a diff. type from what we had before. We used to return `appeals` and `legacy_appeals` but now `type` will be a task type (e.g, `judge_legacy_tasks`, etc). We can tell Paul to check against `appeal_type` and not `type` to determine the type of the appeal.

aroltsch: I am actually gonna look into it more. I am sure there is a way to override it.

aroltsch: I have a better solution.

app/controllers/idt/api/v1/appeals_controller.rb
aroltsch: Yes, he said okay.

aroltsch:
image

app/models/serializers/idt/v1/appeal_serializer.rb
```diff
@@ -14,4 +14,22 @@ def id
   attribute :docket_number
   attribute :docket_name
   attribute :number_of_issues
+
+  attribute :day_waiting do
```
amprokop: *days_waiting, instead of day_waiting, maybe?

PR #7161, opened 2018-09-27
Skip search queue spec

shanear:
These flaking over 50% of the time on circle. Skipping temporarily until we can find the solution.


pkarman:
@shanear should this PR be closed?

PR #7168, opened 2018-09-27
Queue | Indicate hearings on other appeals

### LINKED ISSUE #6780: Investigate hearing data in Case details
We've gotten some support tickets about certain appeal records in Caseflow not showing hearing data. It seems to be related to a Post Remand record vs. the Original appeal record - the hearing is on the Original appeal record but the data is not on the Post Remand record.

Let's add some copy in the Hearing section of Case details to direct users to `View all cases` so they can see other cases associated with the Veteran and therefore the hearing information. 

### AC
- Show the Hearing section for all legacy Post Remand cases (this is an appeal type / VACOLS action)
- Include the copy: "This Post Remand data might have hearing data associated with a previous, original appeal. `View all cases` to see other cases associated with this Veteran."
- When the user clicks `View all cases`, they are scrolled up to the top, and the View all cases link has been expanded. 

### Out of scope for now, but potential future scope
- We add the Hearing label, or other hearing related data, to the case list in View all cases, so a user can easily see if there was a hearing on the case. 
- We should check with Hearing team, because they have ideas for additional columns for search @sneha-pai 

### Related support tickets
- https://github.com/department-of-veterans-affairs/appeals-support/issues/2833 
- https://github.com/department-of-veterans-affairs/appeals-support/issues/2728
- https://github.com/department-of-veterans-affairs/appeals-support/issues/2901
- https://github.com/department-of-veterans-affairs/appeals-support/issues/3080

dannysellers:
Resolves #6780

### Description
For vets with a Post Remand and Original appeal, the Post Remand will appear in the user's queue. If the Original appeal had a hearing held, we do not show any indication of that in the Post Remand's Case Details view. This adds text directing the user to View All Cases for more hearing info.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Reseed FACOLS (`rake local:vacols:seed`)
2. Go to `/queue` as `BVAAABSHIRE`, Case Details for Nola L Stroman (994806951) (the case assigned to Abshire is the Post Remand).
3. Scroll to Hearings section, confirm copy appears. Click link, confirm page scrolls to top, View All Cases panel shows

</details>


app/services/appeal_repository.rb
```diff
@@ -40,6 +40,15 @@ def self.appeals_by_vbms_id(vbms_id)
     cases.map { |case_record| build_appeal(case_record) }
   end
 
+  def self.appeals_with_hearings_by_vbms_id(vbms_id)
+    vacols_ids = VACOLS::Case.where(bfcorlid: vbms_id).pluck(:bfkey)
+    hearings = HearingRepository.hearings_for_appeals(vacols_ids)
```
dannysellers: [`HearingRepository.hearings_for_appeals`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/repositories/hearing_repository.rb#L39) does a bunch of other stuff besides just fetching their `folder_nr`s. Most of this involves [creating `Hearing` objects](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/repositories/hearing_repository.rb#L112), which we don't need for the purposes of this PR/functionality. But, it also fetches hearings using [`VACOLS::CaseHearing.select_hearings`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/vacols/case_hearing.rb#L115), which seems important (filters `where(hearing_type: HEARING_TYPES.keys)`)

app/models/serializers/work_queue/legacy_appeal_serializer.rb
```diff
@@ -26,6 +26,11 @@ class WorkQueue::LegacyAppealSerializer < ActiveModel::Serializer
     end
   end
 
+  attribute :appeal_ids_with_hearings do
+    # AppealRepository.appeals_with_hearings_by_vbms_id(object.veteran_file_number)
```
aroltsch: Danny, you can pass object.vacols_id which is a bfkey so you won't need to query for that.

aroltsch: Oh, you want to find all bfkeys with that vbms_id?

app/services/appeal_repository.rb
```diff
@@ -40,6 +40,17 @@ def self.appeals_by_vbms_id(vbms_id)
...
+
+    hearings.map do |vacols_id, case_hearings|
+      if case_hearings.flatten.all?(&:held?)
```
dannysellers: Not sure whether this should be `.all?` or `.any?`

app/models/serializers/work_queue/legacy_appeal_serializer.rb
```diff
@@ -26,6 +26,11 @@ class WorkQueue::LegacyAppealSerializer < ActiveModel::Serializer
     end
   end
 
+  attribute :appeal_ids_with_hearings do
+    AppealRepository.appeals_with_hearings_by_vbms_id(object.veteran_file_number)
+      .reject { |vacols_id| vacols_id.eql? object.vacols_id }
+  end
```
lowellrex: Do we plan to use the values of any of these appeal IDs we pass to the front-end? As far as I can tell, this code in this PR only uses the length (0 or greater than 0) of the array of IDs as a condition on the front-end.

I am asking because I think we can simplify this code significantly if we change this to a boolean (`has_completed_hearing_on_previous_appeal` or something similar?).

client/app/queue/CaseHearingsDetail.jsx
```diff
@@ -88,32 +105,60 @@ export default class CaseHearingsDetail extends React.PureComponent {
...
-          paddingBottom: '1.5rem',
-          paddingTop: '1rem',
-          borderBottom: '1px solid grey'
-        },
-        '> li:last-child': {
-          borderBottom: 0
-        }
-      })} />;
+    return <React.Fragment>
+      {caseType === 'Post Remand' && Boolean(appealIdsWithHearings.length) && <React.Fragment>
```
lowellrex: Can we put this case type into a shared constant? (Unless we go the route of having this be a pure boolean, in which case we will not need the case type at all here).

client/app/queue/CaseHearingsDetail.jsx
```diff
@@ -88,32 +105,60 @@ export default class CaseHearingsDetail extends React.PureComponent {
...
+      appeal: {
+        caseType,
+        hearings,
+        appealIdsWithHearings
+      }
+    } = this.props;
+
     const listElements = [{
-      label: this.getAppealAttr('hearings').length > 1 ? 'Hearings (Oldest to Newest)' : '',
+      label: hearings.length > 1 ? 'Hearings (Oldest to Newest)' : '',
```
lowellrex: Can we put this in COPY.json? I'm on a quest to kill all strings!

app/models/serializers/work_queue/legacy_appeal_serializer.rb
dannysellers: I was thinking we might want to allow for the possibility of linking directly to another appeal / case details view, but yeah, we don't currently need the actual values

lowellrex: I think it would make sense to put this logic in the `LegacyAppeal` model and change it to a boolean so we can short circuit a trip to VACOLS and simplify the logic on the front-end. What do you think?

lowellrex:
Can we add the code you included in the test plan in the database seed data?

app/models/serializers/work_queue/legacy_appeal_serializer.rb
dannysellers: I completely agree that this feels more like a class method, though I put it in the repository because it's fetching information from appeals besides self. I'm still not totally clear on how/when/why class methods should be separate from repository methods, but I'm happy to move this into `LegacyAppeal` and make it a boolean 👍

client/constants/LEGACY_APPEAL_TYPES_BY_ID.json
```diff
@@ -0,0 +1,7 @@
+{
+  "original": "Original",
+  "post_remand": "Post Remand",
+  "reconsideration": "Reconsideration",
+  "cavc_remand": "Court Remand",
+  "cue": "Clear and Unmistakable Error"
+}
```
lowellrex: 👍

lib/tasks/local/vacols.rake
```diff
@@ -226,6 +227,27 @@ namespace :local do
...
+        vacols_case: FactoryBot.create(
+          :case,
+          :assigned,
+          :type_original,
+          user: user,
+          bfcorlid: veteran_file_number,
+          case_hearings: [FactoryBot.create(:case_hearing, :disposition_held, user: user)]
+        )
+      )
+    end
```
lowellrex: We've changed the order of the database seed jobs we run with a recent PR (https://github.com/department-of-veterans-affairs/caseflow/pull/7217), so I think this logic will need to go into `db/seeds.rb`.

mkhandekar:
@dannysellers what is the copy for post remand's hearing section?  the gif is too speedy for me 😅

dannysellers:
@mkhandekar oh, sorry about that! It comes from the [issue AC](https://github.com/department-of-veterans-affairs/caseflow/issues/6780), I should've just copied those into here:
> This Post Remand data ~might have~ has hearing data associated with a previous, original appeal. `View all cases` to see other cases associated with this Veteran.

The `View all cases` link scrolls to the top of the page and opens the VeteranCasesView (the panel toggled by clicking the other View all cases link, in the page header)

Issue #7118, opened 2018-09-24
Create Informal Conference tracked item when establishing a claim


leikkisa:
## Story
In order to be able to create an informal conferenced tracked item for Higher-Level Review claims, we need to make a couple of calls to BGS

Add call to manageClaimantLetter_V2 and generateTrackedItems to BGSService
Update the ruby_bgs version in the Gemfile

If a user selects the "Informal Conference" option for a Higher Level Review, then the tracked item should be created as part of establishing the claim (after the claim is established).

Necessary steps should be taken to include it in the idempotency for establishing a claim (tracking success of each of these calls, re-trying if necessary).

requires department-of-veterans-affairs/ruby-bgs#65

shanear:
![screen shot 2018-10-03 at 1 04 43 pm](https://user-images.githubusercontent.com/1596259/46426651-30153a80-c70d-11e8-9cdc-c4c994b5d825.png)
i never thought i would see this day. so happy. 🙂

PR #7171, opened 2018-09-27
Higher-level review informal conference tracked item

leikkisa:
connects #7118


app/services/external_api/bgs_service.rb
```diff
@@ -212,6 +212,39 @@ def get_participant_id_for_user(user)
     end
   end
 
+  def find_claimant_letters(document_id)
+    DBService.release_db_connections
+    MetricsService.record("BGS: find claimat letter for document #{document_id}",
```
pkarman: typo: should be "claimant"

db/migrate/20180927004213_add_hlr_informal_conference_tracked_item.rb
```diff
@@ -0,0 +1,7 @@
+class AddHlrInformalConferenceTrackedItem < ActiveRecord::Migration[5.1]
+  def change
+    add_column :end_product_establishments, :doc_id, :string
```
pkarman: should this be `doc_reference_id`? the convention seems to be `*_id` for local int FKs and `*_reference_id` for strings pointing at records in external systems.

db/migrate/20180927004213_add_hlr_informal_conference_tracked_item.rb
```diff
@@ -0,0 +1,7 @@
...
+    add_column :end_product_establishments, :development_item_id, :string
```
pkarman: same question: should this be `development_item_reference_id`?

spec/models/claim_review_spec.rb
```diff
@@ -59,6 +70,14 @@
     )
   end
 
+  let!(:claimant) do
+    Claimant.create!(
```
pkarman: this could use FactoryBot instead?
```ruby
create(
  :claimant,
 ...
)
```

app/models/claim_review.rb
```diff
@@ -69,6 +73,10 @@ def on_sync(end_product_establishment)
 
   private
 
+  def informal_conference?
```
shanear: nice, i like this pattern

app/models/veteran.rb
```diff
@@ -24,6 +24,10 @@ class Veteran < ApplicationRecord
   # Germany and Australia should be temporary additions until VBMS bug is fixed
   COUNTRIES_REQUIRING_ZIP = %w[USA CANADA].freeze
 
+  # C&P Live = '1', C&P Death = '2'
+  BENEFIT_TYPE_CODE_LIVE = "1".freeze
+  BENEFIT_TYPE_CODE_DEATH = "2".freeze
```
shanear: I know these were already like this, but it might be nicer to make this a hash, since these values are related?

```
BENEFIT_TYPE_CODES = {
  live: "1",
  death: "2" 
}.freeze
```

app/models/veteran.rb
```diff
@@ -73,6 +77,16 @@ def age
...
   end
 
+  def benefit_type_code
+    @benefit_type_code ||= begin
+      if date_of_death.nil?
+        BENEFIT_TYPE_CODE_LIVE
+      else
+        BENEFIT_TYPE_CODE_DEATH
+      end
+    end
```
shanear: might be good to add a helper method `deceased?` to bring a little better meaning to `date_of_death.nil?`

As an added bonus, that would let this whole function fit nicely on one line
```
@benefit_type_code ||= deceased? ? BENEFIT_TYPE_CODE_DEATH : BENEFIT_TYPE_CODE_LIVE
```

app/services/external_api/bgs_service.rb
```diff
@@ -212,6 +212,39 @@ def get_participant_id_for_user(user)
     end
   end
 
+  def find_claimant_letters(document_id)
```
shanear: since this isn't used anywhere else in the code, might be nice to add a comment that describes why it's here.

spec/models/claim_review_spec.rb
shanear: I go back and forth on whether it's better to used FactoryBot when a simple ole `create!` will do.

spec/models/claim_review_spec.rb
```diff
@@ -363,6 +382,41 @@ def epe
...
         end
       end
+
+      context "when informal conference is true" do
+        let(:informal_conference) { true }
+
+        it "generates claimant letter and tracked item" do
+          subject
+          epe = claim_review.end_product_establishments.last
+          expect(epe.reload).to have_attributes(
```
shanear: you shouldn't need to reload this. Since you loaded it after the action of `subject`.

app/models/higher_level_review.rb
```diff
@@ -58,6 +58,10 @@ def issue_code(rated)
 
   private
 
+  def informal_conference?
+    !!informal_conference
```
shanear: isn't this already a boolean? `!!` is typically used to convert non boolean values to booleans

app/models/higher_level_review.rb
```diff
@@ -100,8 +104,14 @@ def dta_supplemental_claim
     @dta_supplemental_claim ||= SupplementalClaim.create!(
       veteran_file_number: veteran_file_number,
       receipt_date: Time.zone.now.to_date,
-      is_dta_error: true
-    )
+      is_dta_error: true,
+      benefit_type: benefit_type
+    ).tap do |sc|
```
shanear: `tap` ❤️

app/models/end_product_establishment.rb
```diff
@@ -180,6 +186,24 @@ def create_associated_rated_issues!
     )
   end
 
+  def generate_claimant_letter!
+    return if doc_id
+    generate_claimant_letter_in_bgs.tap do |result|
+      update!(
```
shanear: I'm sort of a "as few lines as possible" guy, so I'm thinking 192-194 could be combined, but this is cool too if you think it's more readable.

app/models/end_product.rb
```diff
@@ -78,7 +74,7 @@ class EndProduct
   validates :gulf_war_registry, :suppress_acknowledgement_letter, inclusion: { in: [true, false] }
 
   def benefit_type_code
-    @benefit_type_code ||= BENEFIT_TYPE_CODE_LIVE
+    @benefit_type_code ||= Veteran::BENEFIT_TYPE_CODE_LIVE
   end
```
shanear: Now that `Veteran` implements `benefit_type_code`, shouldn't we be able to just delegate this method to `veteran`?

ie:
```
delegate :benefit_type_code, to: :veteran
```

app/models/end_product_establishment.rb
```diff
@@ -180,6 +186,24 @@ def create_associated_rated_issues!
...
+        doc_id: result
+      )
+    end
+  end
+
+  def generate_tracked_item!
+    return if development_item_id
```
shanear: i don't see this logic path in the tests. I think its probably fine if we don't have unit tests for `generate_tracked_item!` or `generate_claimant_letter!` since they are subsumed by `claim_review.process_end_product_establishments!`

However if we do that, i think we want to make sure the BGS calls get skipped in `remains idempotent despite multiple VBMS failures` test on `claim_review.process_end_product_establishments!`

app/models/end_product_establishment.rb
```diff
@@ -180,6 +186,24 @@ def create_associated_rated_issues!
     )
   end
 
+  def generate_claimant_letter!
+    return if doc_id
```
shanear: The comment on line 199 applies here as well

app/models/higher_level_review.rb
leikkisa: Peter suggested this as a safety just in case it's ever null.  We have validation preventing that... so I think we can take it out.  I figured it didn't hurt anything but will clean it up.

db/migrate/20180927004213_add_hlr_informal_conference_tracked_item.rb
leikkisa: Adding "reference" to the field names

spec/models/claim_review_spec.rb
leikkisa: Hmm... will leave this for now (all the related pieces are using create/new too) and we can refactor later if we want to.

app/models/end_product_establishment.rb
leikkisa: Really good catch!  Will add tests for this!

app/models/veteran.rb
```diff
@@ -183,6 +191,10 @@ def before_create_veteran_by_file_number
 
   private
 
+  def deceased?
+    !date_of_death.nil?
```
pkarman: it might be clearer to write `date_of_death.present?` to avoid the double negative.

Issue #7102, opened 2018-09-24
Refactor task action dropdowns


mdbenjam:
Currently our task action dropdowns are very specific to role. Each dropdown operates differently. We propose the following pattern:

The initial request to the backend will get a list of actions that user can perform and the URLs those actions take the user to. This helps us solve a couple of problems: 1) If a user has multiple roles, we can give them tasks from all of those roles. 2) We can add tasks that should be accessible to everyone. 3) It scales better than making a new component for every role.

See the following slack conversation for more context: https://dsva.slack.com/archives/C6E41RE92/p1537542173000100

laurjpeterson:
@lowellrex i moved this to backlog, but, should it actually be closed?

lowellrex:
I think we can close this puppy out!

Issue #6903, opened 2018-09-11
EP routing for legacy and AMA


laurjpeterson:
Caseflow Dispatch currently creates EPs (when they can) and routes them to particular ROs for Board grants and partial grants depending on if there are special issues or not. There are quite a few special issues that require emails to be sent, because the AOJs adjusting the claim do not use VBMS - these are most commonly other business lines like VHA, NCA, Fiduciary, etc.

We want to know if this routing should be updated for AMA appeals, and correct it for legacy appeals as well.

# AMA
**Open questions:**
- Dayna said just SeaTac and St. Pete, Josh says all three DROCs SeaTac, St. Pete, and DC. Which is correct?
- [x] What else do we need to add to the EP in order for NWQ to apply it's routing rules?
   - Is what we're doing in Caseflow Dispatch enough?
   - Are Corp DB flashes enough?
- Will all flashes already be there for the Veteran? Can we assume that? (We have an endpoint to read out the flashes on a Veteran)
- What business line does Waiver of Overpayment correspond to?
- Where do `Foreign pension, DIC - Mexico, Central and South America, Caribbean` and `Foreign pension, DIC - all other foreign countries` go through NWQ Station 499 or to the PMC ROs depending on region - PMC: RO81 - Philadelphia PMC, RO82 - Milwaukee PMC, or RO83 - St. Paul PMC depending on region 

### Appeals without Special Issues
- All Board Grants (compensation and pension management center (PMC)) are routed to DROCs - SeaTac (Station 346) or St. Pete (Station 317) via the NWQ (Station 499)
    - Note: We would like to create new station numbers for DROCs SeaTac and St. Petersburg but for now we are using the codes of the existing RO.
     - **Note: Josh said all 3 DROCs, including DC, while Dayna just said SeaTac and St. Pete**
- NWQ can distinguish Pension EPs because of their claim label
- NWQ can distinguish DTA errors because of their claim label
- We have to add a contention to each EP to chain issues

#### Comp & Pen --> Creating VBMS EPs
Special Issue Name |Routing logic for Board Grants
--- |---
Incarcerated Veterans | Caseflow relies on there being an `Incarcerated` Flash on the Veteran. NWQ sees this flash and routes accordingly. 
Non-rating issue | Use Board Grant Non-Rating EP codes for comp and pen
Private Attorney or Agent | Caseflow relies on there being `Private Attorney - Fee PayablePrivate Attorney - No Fee Payable` Flash
[...]

PR #7175, opened 2018-09-27
Make generic task dropdown

mdbenjam:
Connects #7102 

### Description
First PR to fix the task dropdown. This change adds a new task drop down that will be used for all cases. We replace the attorney dropdown with this new one, and set up the structures to add new options.

We also move checkout flows to have unique URL paths to move state from Redux into the URL about which flow we're in.

### Testing Plan
1. Manually test various checkout flows.


app/models/legacy_tasks/legacy_task.rb
```diff
@@ -47,6 +47,18 @@ def days_waiting
     (Time.zone.today - assigned_at.to_date).to_i if assigned_at
   end
 
+  def get_allowed_actions(role)
+    puts "TEST", role, Constants::TaskActionList::LEGACY_ACCESS_CONTROL.first[:vacols_role].include?(role)
+    Constants::TaskActionList::LEGACY_ACCESS_CONTROL.
```
mdbenjam: Legacy tasks will continue to be determined by role 🤷‍♂️

client/app/queue/CaseSnapshot.jsx
```diff
@@ -254,17 +255,18 @@ export class CaseSnapshot extends React.PureComponent<Props> {
...
       userRole
     } = this.props;
-    let CheckoutDropdown = <React.Fragment />;
+    let ActionDropdown;
     const dropdownArgs = { appealId: appeal.externalId };
 
+
     if (userRole === USER_ROLE_TYPES.attorney) {
-      CheckoutDropdown = <AttorneyActionsDropdown {...dropdownArgs} />;
+      ActionDropdown  = <ActionsDropdown task={taskAssignedToUser} appealId={appeal.externalId} />;
```
mdbenjam: In this PR we're only replacing the attorney dropdown. The others to follow in subsequent PRs!

client/app/queue/utils.js
```diff
@@ -332,7 +330,7 @@ export const buildCaseReviewPayload = (
   }
 
   if (userRole === USER_ROLE_TYPES.attorney) {
-    _.extend(payload.data.tasks, { document_type: decision.type });
+    _.extend(payload.data.tasks, { document_type: checkoutFlow });
```
mdbenjam: We no longer get the type from decision.type but instead encode it in the URL.

lib/constants/task_action_list.rb
```diff
@@ -0,0 +1,39 @@
+module Constants::TaskActionList
+  ACCESS_CONTROL = [
+    {
+      task_type: ["AttorneyTask"],
```
mdbenjam: We can add more conditions here in the future. I'm thinking we'll need to add things css_role, task_action, and possibly others.

client/app/constants/SpecialIssues.js
```diff
@@ -72,7 +72,8 @@ const SPECIAL_ISSUES = [
     specialIssue: 'hearingIncludingTravelBoardVideoConference',
     stationOfJurisdiction: 'regional',
     snakeCase: 'hearing_including_travel_board_video_conference',
-    unhandled: null
+    unhandled: null,
+    nonCompensation: true
```
mdbenjam: Unrelated change. Just wanted to get it in: https://github.com/department-of-veterans-affairs/caseflow/issues/6903#issuecomment-420700931

lib/constants/task_action_list.rb
lowellrex: Would it be possible to move this access control to the various Task classes themselves? Perhaps calling something like `generic_task.available_actions(current_user)` would return this list. One advantage I see in making these instance methods is that it allows us to be flexible around which actions are returned based on the current state of the task itself.

Issue #7157, opened 2018-09-26
Case Review | Judges | Caseflow doesn't set the completion date of OMO reviews


evankroske:
When a judge reviews an OMO request and creates an OMO, [Caseflow doesn't set the completion date of the corresponding Decass record](https://github.com/department-of-veterans-affairs/caseflow/blob/caccf19a7db3c318d6f9e12de47a9f6f1c40eef1/app/repositories/queue_repository.rb#L76). This keeps attorneys from getting credit for OMO requests.

Also, Caseflow doesn't update deoq, either.

PR #7182, opened 2018-09-28
Case Review | Allow judges to rate the quality of OMO requests

evankroske:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/7157

### Description
Changes the OMO assignment flow to require the judge to rate the quality of OMO requests.

### Testing Plan
- Become BVASCASPER1.
- Create an OMO request for a legacy case with document ID "12345-12345678" and assign it to "Judge with hearings and cases" (BVAAABSHIRE).
- Become BVAAABSHIRE.
- Open the case.
- Select "Assign OMO".
- Fill out the evaluation form.
- Click "Continue".
- Confirm that the success message mentions OMO.


client/app/queue/components/JudgeActionsDropdown.jsx
```diff
@@ -89,28 +89,18 @@ class JudgeActionsDropdown extends React.PureComponent<Props, ComponentState> {
 
     this.props.setCaseReviewActionType(actionType);
 
+    let nextPage = 'special_issues';
+    if (appeal.isLegacyAppeal) {
```
dannysellers: Should these two conditionals be an if/else?

client/app/queue/components/JudgeActionsDropdown.jsx
```diff
@@ -89,28 +89,18 @@ class JudgeActionsDropdown extends React.PureComponent<Props, ComponentState> {
...
+    }
     if (actionType === DECISION_TYPES.OMO_REQUEST) {
-      const payload = buildCaseReviewPayload(decision, userRole, appeal.issues, {
-        location: 'omo_office',
-        attorney_id: task.assignedBy.pgId,
-        isLegacyAppeal: appeal.isLegacyAppeal
-      });
-      const successMsg = sprintf(COPY.JUDGE_CHECKOUT_OMO_SUCCESS_MESSAGE_TITLE, appeal.veteranFullName);
-
-      this.props.requestSave(`/case_reviews/${task.taskId}/complete`, payload, { title: successMsg }).
```
dannysellers: Nit: `requestSave` and `deleteAppeal` are both used in `mapDispatchToProps` (but not the component), so the linter won't catch them

app/repositories/queue_repository.rb
```diff
@@ -169,11 +167,7 @@ def update_location_to_attorney(vacols_id, attorney)
       vacols_case.update(bfattid: attorney.vacols_attorney_id)
     end
 
-    def update_vacols_for_bva_dispatch(decass_record, location, decass_attrs)
-      unless decass_record.draft_decision?
-        msg = "The work product is not decision"
-        fail Caseflow::Error::QueueRepositoryError, msg
-      end
+    def update_vacols(decass_record, location, decass_attrs)
```
aroltsch: The method name is very generic. Maybe move this logic into `sign_decision_or_create_omo` since the method is almost empty. Also, can we please double check we need to update the following for OMO:

```
# When the DAS final review is done by the VLJ and the case is charged to 4E the VLJ,
      # Attorney and Team get updated in the BRIEFF table
      decass_record.reload.case.update(
        bfmemid: decass_record.dememid,
        bfattid: decass_record.deatty,
        bfboard: decass_record.deteam
      )
```

aroltsch:
@evankroske would it be possible to add tests to the judge_case_review.spec to make sure all the fields are updated correctly for OMO.

client/app/queue/components/JudgeActionsDropdown.jsx
evankroske: I refactored the logic to use `else if`

app/repositories/queue_repository.rb
evankroske: Yes, DAS updates those fields when the judge reviews an OMO request: https://github.com/department-of-veterans-affairs/VACOLS/blob/1d31c01e5410d9b1c5a52fd5d454914ab6f56da4/src/decass.pbl-dump/w_update_final_2017.srw#L502

PR #7183, opened 2018-09-28
Add tasks/:id/assignable_... endpoints

lowellrex:
Connects #6703. Adds endpoints for tasks to call to get the list of people or organizations that the task can be assigned to.

## To the reviewer
The meat of the review starts with the [new actions in the `TasksController`](https://github.com/department-of-veterans-affairs/caseflow/pull/7183/files#diff-f4416012d567d02da854cfbc9e2df2caR98).


app/controllers/application_controller.rb
```diff
@@ -129,15 +129,18 @@ def deny_vso_access
...
     # This feature toggle control access of attorneys to create admin actions for co-located users
     return true if current_user.attorney_in_vacols? && feature_enabled?(:attorney_assignment_to_colocated)
     # This feature toggle control access of judges to assign cases to attorneys
-    return true if current_user.judge_in_vacols? && feature_enabled?(:judge_assignment_to_attorney)
-    redirect_to "/unauthorized"
-    # :nocov:
+    true if current_user.judge_in_vacols? && feature_enabled?(:judge_assignment_to_attorney)
+  end
+
+  def verify_task_assignment_access
```
lowellrex: Split into two functions so we can call the function that returns the boolean in `Organizations::TasksController. verify_task_access()`

app/controllers/organizations_controller.rb
```diff
@@ -6,15 +6,15 @@ class OrganizationsController < ApplicationController
...
+    render json: { organizations: Organization.assignable_hash }
   end
 
   def show
     render "organizations/show"
   end
 
   def members
-    render json: { members: organization.members.map { |m| { id: m.id, css_id: m.css_id, full_name: m.full_name } } }
+    render json: { members: organization.assignable_members_hash }
```
lowellrex: Considering getting rid of these two methods since we don't currently use them on the front-end and are using the `TasksController` to return these two pieces of data.

app/controllers/organizations_controller.rb
```diff
@@ -1,22 +1,14 @@
...
+  before_action :verify_organization_access
+  before_action :verify_role_access
+  before_action :verify_feature_access
   before_action :set_application
   skip_before_action :deny_vso_access
 
-  def index
-    render json: { organizations: Organization.where(type: nil).map { |o| { id: o.id, name: o.name } } }
-  end
-
```
lowellrex: Removed these actions since we don't use them on the front-end and the functions that serve the same purpose have now been built into the `TasksController`.

app/models/organization.rb
```diff
@@ -1,6 +1,10 @@
 class Organization < ApplicationRecord
   has_many :tasks, as: :assigned_to
 
+  def self.assignable
+    where(type: nil)
```
mdbenjam: Is this just to exclude VSO tasks right now?

lowellrex: As well as the Board's org (and actually BVA dispatch, fixing that now...).

PR #7187, opened 2018-09-28
Fix syntax error

shanear:
`ExternalAPI::VBMSService` is a blind spot in our test blanket.

Fixes the issue described here: https://github.com/department-of-veterans-affairs/caseflow/issues/6212#issuecomment-425500816

connects #6212


kencheeto:
Dang, good to know! thanks for debugging & validating

PR #7190, opened 2018-09-28
[WIP] Rescue/stash VBMS errors in async jobs, rethrow for retry

pkarman:
connects #6942 

### Description
In order to surface errors and allow for manual retry of background tasks, capture exceptions, stash the message, and re-throw the error.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan

TODO


spec/jobs/claim_review_process_job_spec.rb
```diff
@@ -0,0 +1,37 @@
+require "rails_helper"
+
+class AClaimReview
```
shanear: ooh, I like that you made a new class so that this tests the concern 💯

app/jobs/claim_review_process_job.rb
```diff
@@ -7,6 +8,11 @@ def perform(claim_review)
     RequestStore.store[:application] = "intake"
     RequestStore.store[:current_user] = User.system_user
 
-    claim_review.process_end_product_establishments!
+    begin
+      claim_review.process_end_product_establishments!
+    rescue VBMS::ClientError => err
+      claim_review.update_error!(err.to_s)
```
shanear: does `VBMS::ClientError.to_s` capture all the info we want? I think so, but just wanted to see if you checked.

app/jobs/claim_review_process_job.rb
```diff
@@ -7,6 +8,11 @@ def perform(claim_review)
...
+      raise err
```
shanear: Is it necessary to rely on the `ActiveJob` retry on exception logic when reviews that fail to process would be caught by `.requires_processing` [here](
https://github.com/department-of-veterans-affairs/caseflow/pull/7190/files#diff-035aa12863938cbbacadb6d85f570423R41)?

pkarman: It depends on how often you want them to retry and if you want Sentry to know about failures.

If we swallowed the exception (did not re-`raise`), then the normal `ActiveJob` retry logic would not apply, and we'd try again in 3+ hours via `.requires_processing`. We also would not see Sentry alerts at all.

If we re-throw, then we'll try again in 3 seconds, 30 seconds, etc. https://github.com/department-of-veterans-affairs/caseflow/blob/master/config/initializers/shoryuken.rb#L8

app/jobs/claim_review_process_job.rb
pkarman: the `to_s` should return (by default) whatever the message was passed to the `Exception` class initially, w/o the backtrace. I assume we do not want the backtrace.

app/jobs/claim_review_process_job.rb
shanear: > It depends on how often you want them to retry and if you want Sentry to know about failures.

This shouldn't play into our decisionmaking, as we can send the error to sentry without needing to raise it (see an example of how we do that here: https://github.com/department-of-veterans-affairs/caseflow/blob/a5560fa436ea8f65fa6c56c23f066fd089bd40a2/app/controllers/intakes_controller.rb#L37)

> If we re-throw, then we'll try again in 3 seconds, 30 seconds, etc.

I honestly have no idea whether this is better or if we should just try every ~3 hours. Seems fine either way. I'm open to whatever you think is best.

shanear: I think the most common errors are:

- Timeouts (a quick retry should fix)
- Caching bugs with our dependencies (issue resolves in 0-2 days)
- Dependency outages (retry should fix when the system comes back up (usually no longer than a day)
- Bugs with our dependencies that will never resolve (e.g. incident flash)

Unfortunately these all are sort of different categories, I don't really know about the frequency of each since this is just from looking at the XML output. We could start tallying different kinds in the code if we wanted to.

I do think server load can be problematic for our dependencies, but we've never experienced it so far for this feature.

I'm leaning towards every ~3 hours since it's the least aggressive and we can always change back to the sidekiq exponential rolloff, but I'll defer to whatever you think.

pkarman: I we ignore the ActiveJob retry, do we want to log errors to Sentry?

One goal of automatic retry is that self-correcting jobs means less noise from alerts because human intervention is less necessary. OTOH, we probably want to know sooner than later if we are suffering from some systemic upstream failure like a service outage.

shanear: > do we want to log errors to Sentry?

I think we do. We can always ignore them if they get out of hand.

pkarman:
thanks for the feedback @shanear -- I changed the exception handling in af63050ff

PR #7196, opened 2018-10-01
Schedule CAVC/AOD formatting

### LINKED ISSUE #7137: Assign Hearings | CAVC/AOD formatting
In the 'Scheduled' and 'Assign Hearings' tabs in the center of the page, if the appeal type is either 'AOD' or 'CAVC, the text should be red.

Sjones352:
Resolves #7137

### Description
This PR shows 'AOD' or 'CAVC in red and all other types in black text. 

### Acceptance Criteria
- [ ] In the 'Scheduled' and 'Assign Hearings' tabs in the center of the page, if the appeal type is either 'AOD' or 'CAVC, the text should be red.

### Testing Plan
1. Go to Schedule/assign

screen shot 2018-10-01 at 12 30 01 pm


client/app/hearingSchedule/components/AssignHearings.jsx
```diff
@@ -81,7 +81,7 @@ export default class AssignHearings extends React.Component {
   tableRows = (veterans) => {
     return _.map(veterans, (veteran) => ({
       caseDetails: veteran.name,
-      type: this.veteranTypeColor(veteran.type),
+      type: this.veteranTypeColor(veteran),
```
sharonwarner: I think it makes more sense the way you had this set up before. We should only send parameters to methods that are necessary, so we should only have to send `veteran.type`.

client/app/hearingSchedule/components/AssignHearings.jsx
```diff
@@ -66,13 +66,13 @@ export default class AssignHearings extends React.Component {
     </div>;
   };
 
-  veteranTypeColor = (type) => {
-    let veteranType;
+  veteranTypeColor = (docketType) => {
+    let veteranType = docketType.type;
```
sharonwarner: We shouldn't need this variable if we only send veteran.type.

client/app/hearingSchedule/components/AssignHearings.jsx
```diff
@@ -66,13 +66,13 @@ export default class AssignHearings extends React.Component {
...
 
-    if (type === 'CAVC') {
-      veteranType = <span {...colorAOD}>CAVC</span>;
-    } else if (type === 'AOD') {
-      veteranType = <span {...colorAOD}>AOD</span>;
+    if (docketType.type === 'CAVC') {
+      return veteranType = <span {...colorAOD}>CAVC</span>;
```
sharonwarner: Similarly here, we don't need to set `<span />` equal to a variable, we can just return it.

client/app/hearingSchedule/components/AssignHearings.jsx
Sjones352: Fixec

PR #7202, opened 2018-10-01
We must go through appeal to get attorney acse reviews

aroltsch:


app/models/serializers/idt/v1/appeal_serializer.rb
```diff
@@ -24,10 +24,14 @@ def id
   end
 
   attribute :documents do
-    if @instance_options[:task]
+    if object.is_a?(LegacyAppeal) && @instance_options[:task]
```
aroltsch: Attorney case reviews in AMA should be associated with the appeal not a single task. I think the idea of adding appeal_id to the attorney_case_reviews table is a good one.

PR #7203, opened 2018-10-01
Don't try and match profile dates when updating request issues

shanear:
connects #6212

When updating request issues, Caseflow was always removing all contentions and readding them, instead of reusing certain contentions when they didn't change. This was because the equality on profile date was never matching, and instead of attempting to make that work, I just took it out since it's not necessary, since `rating_issue_reference_id`s are unique without it.


spec/models/request_issues_update_spec.rb
```diff
@@ -74,8 +74,6 @@
     existing_request_issues.map do |issue|
       {
         reference_id: issue.rating_issue_reference_id,
-        # TODO: validate the string format this comes in
```
pkarman: 🎉

Issue #6982, opened 2018-09-14
AMA Intake | Clicking "Add this issue", and displaying added issues


suzchap:
## Story
For any AMA forms, when a user needs to add issues, they will be walked through a multi-screen selection process to help them make the best choice. The landing page created in #6966 will be their starting point and a modal on top of it will walk the user through rated, non-rated, and an "unknown" option. 

This ticket is for activating the "Add this issue" button that's on the rated issues modal (created in #6981) and how the issue displays on the "Add issue" page (created in #6966).

requires #6981

## Acceptance criteria
- clicking the "Add this issue" button on the rated issue modal closes the modal
- on the "Add issues" page, the previously selected rated issue and it's decision date are added to the "Requested issues" section.
- The issues listed are sequentially numbered and shown in the order they were created

## Notes
- Out of scope / in another ticket
   - Remove trash icon and link
   - Cancel link
   - "None of these..." button
   - Notes field

## Copy
- no new static copy

## Mocks
**With just one issue added**
<img width="599" alt="screen shot 2018-09-14 at 11 39 58 am" src="https://user-images.githubusercontent.com/11219417/45560382-eecaf280-b812-11e8-95ad-c2e30669055d.png">

**With two issues added**

## How to Validate
- Doesn't require a specific test user
- Form select > search id > review, any answers > on "Add issue page, click "+Add issue" > on modal, select a rated issue > click "Add this issue" button > add another issue (so you can check numbering on Add issue page)
- TBD: does this need to be tested with all 3 AMA forms or will one suffice?

shanear:
![screen shot 2018-10-03 at 1 10 33 pm](https://user-images.githubusercontent.com/1596259/46426879-d3664f80-c70d-11e8-84f1-442a8f5e21ba.png)
looking good

PR #7205, opened 2018-10-01
Hook up add issues button

jiahuang:
Connects #6982

add_issues_click


client/app/intake/components/AddIssuesModal.jsx
```diff
@@ -1,12 +1,45 @@
...
+    };
+  }
+
+  radioOnChange = (value) => {
+    this.setState({
+      referenceId: value
+    });
+  }
+
+  getProfileDateForIssue = (referenceId) => {
```
jiahuang: There are a few alternatives to get the profile date for an issue.
1. set it as part of the "value" key for the radio button, ex something like: `value: {issue.id}#{rating.profileDate}`, and then deconstruct the string to pull out the fields when the "add issue" button is clicked. 
2. modify the existing radio field to pass in more data fields on click

This method seemed to be the least hacky and didn't touch any shared components, but I'm happy to change it.

client/app/intake/reducers/common.js
```diff
@@ -11,11 +11,24 @@ export const commonReducers = (state) => {
     });
   };
 
+  actionsMap[ACTIONS.ADD_ISSUE] = () => {
```
jiahuang: Note that this adds a new action & reducer. The current way we do issues is by setting a "selected" field in the state for a particular issue. I thought about reusing that, but then realized that the ordering of the issues would change when the rendering is displayed on the add issues landing page, so instead I made a new array to keep track of selected issues.

client/app/intake/components/AddIssuesModal.jsx
shanear: Since the rating_issue id is unique on its own, I think what you are doing is fine.

The only thing I would change is that I would put this logic in the reducer (or maybe just the action when we submit the data) and not the component.

client/app/intake/components/AddIssuesModal.jsx
```diff
@@ -1,12 +1,45 @@
...
+    let foundDate = _.filter(this.props.ratings, (ratingDate) => {
+      let foundIssue = _.filter(ratingDate.issues, (issue) => referenceId === issue.reference_id);
+
+      return foundIssue.length === 1;
```
shanear: I think you can use `_.some` to achieve the effect of 29-31:

https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/intakeCommon/util/index.js#L14

client/app/intake/reducers/common.js
leikkisa: I like this way of doing it, I the code for sending it to the back end will be cleaner as well.

client/app/intake/components/AddIssuesModal.jsx
leikkisa: I think this is better than the alternatives, thanks for thinking it through.

client/app/intake/util/index.js
```diff
@@ -145,3 +145,21 @@ export const getAddIssuesFields = (formType, veteran, intakeData) => {
 
   return fields.concat(claimantField);
 };
+
+export const getListOfAddedIssues = (intakeData) => {
```
shanear: The naming of this method is a little confusing. It seems like the purpose for this is to get the values used to display the issues. I think I'd name this something like `addedIssuesDisplayValues(intakeData)`

spec/feature/intake/appeal_spec.rb
```diff
@@ -290,5 +290,11 @@ def check_row(label, text)
     safe_click "#button-add-issue"
     expect(page).to have_content("Left knee granted")
     expect(page).to have_content("PTSD denied")
+
+    # adding an issue should show the issue
+    find("label", text: "Left knee granted").click
+    safe_click ".add-issue"
+
+    expect(page).to have_content("1. Left knee granted")
```
shanear: simple and good 👍

client/app/intake/reducers/common.js
shanear: I agree with this approach :100:

client/app/intake/util/index.js
```diff
@@ -145,3 +145,21 @@ export const getAddIssuesFields = (formType, veteran, intakeData) => {
...
+  let issues = intakeData.addedIssues;
```
kencheeto: if you assign `let issues = intakeData.addedIssues || [];` here you can avoid the ternary below

client/app/intake/components/AddIssuesModal.jsx
kencheeto: not blocking, not urgent, might be something to watch for - i wonder if normalizing `ratingDate` would avoid nested loops on `ratings`? so instead of 
```js
ratings = {
  "rating date 1": {
    ...
    issues: [
      { reference_id: 'id1' },
      { reference_id: 'id2' }
    ]
  },
  "rating date 2": {
    ...
    issues: [
      { reference_id: 'id3' }
    ]
  }
};
```
we have
```js
ratings = [
  {
    ratingDate: 'rating date 1',
    reference_id: 'id1'
  },
  {
    ratingDate: 'rating date 1',
    reference_id: 'id2'
  },
  {
    ratingDate: 'rating date 2',
    reference_id: 'id3'
  }
];
```

had a similar nested loop here too: https://github.com/department-of-veterans-affairs/caseflow/blob/f59a07e6d435450cf00be5514dcafb796d41ea48/client/app/intakeCommon/util/index.js#L94-L110

client/app/intake/pages/addIssues.jsx
```diff
@@ -23,16 +24,22 @@ class AddIssues extends React.PureComponent {
...
-        legacyStyling={false}
-        classNames={['usa-button-secondary']}
-        onClick={this.props.toggleAddIssuesModal}
-      >
-        + Add issue
-      </Button>;
+    const issuesComponent = () => {
+      let issues = getListOfAddedIssues(intakeData);
+
+      return <div>
```
kencheeto: if you don't need the div you could use `React.Fragment` so you're still returning a single element but it won't create the extra node

jiahuang: need the div for styling

Issue #6986, opened 2018-09-14
AMA Intake | Add a notes field to the Rated Issues page


suzchap:
## Story
For any AMA forms, when a user needs to add issues, they will be walked through a multi-screen selection process to help them make the best choice. The landing page created in #6966 will be their starting point and a modal on top of it will walk the user through rated, non-rated, and an "unknown" option.

This ticket is for adding an optional notes field to the first modal for selecting rated issues (created in #6981).

requires #6981
requires #7000

## Technical notes
- Notes are stored in the database for the request issue

## Acceptance criteria
- User can add notes as optional.

## Notes
- There may be a character limit on contention text. We'll have to figure that out and put it in a follow up story

## Copy
- Notes
- Optional

## Mocks
<img width="760" alt="screen shot 2018-09-14 at 12 05 45 pm" src="https://user-images.githubusercontent.com/11219417/45561903-d65cd700-b816-11e8-8c7f-849ba9733d66.png">

## How to Validate
- Doesn't require a specific test user
- Form select > search id > review page: any answers > on "Add issue page, click "+Add issue" to display modal > enter text in Notes field > click "Add this issue" (notes don't currently display anywhere so Dev will need to check that they were logged.
- TBD: does this need to be tested with all 3 AMA forms or will one suffice?

leikkisa:
I believe this requires #7000, since it requires saving to the database, so I am going to wait to continue working on it

leikkisa:
@suzchap @carolaponce Should I be displaying the notes on the list of added issues?

shanear:
```
[...]

PR #7207, opened 2018-10-01
Add "notes" field for request issues

leikkisa:
connects #6986


leikkisa:
screen shot 2018-10-03 at 6 27 08 pm

local/vacols/bgs_setup.csv
```diff
@@ -18,7 +18,7 @@ vbms_id,bgs_key
 767574947S,
 209179363S,
 626343664S,
-849386058S,
+849386058S,has_many_ratings
```
pkarman: I have re-introduced these locally as well but I'm not sure we want them back in the repo? I don't know why they were removed in the first place.

client/app/intake/components/AddIssuesModal.jsx
```diff
@@ -24,8 +26,14 @@ class AddIssuesModal extends React.Component {
...
 
+  notesOnChange = (value) => {
+    this.setState({
+      notes: value
+    });
+  }
+
   onAddIssue = () => {
-    this.props.addIssue(this.state.referenceId, this.props.ratings, true);
+    this.props.addIssue(this.state.referenceId, this.props.ratings, true, this.state.notes);
```
pkarman: this many arguments to a method seems to beg for named pairs instead.
```javascript
this.props.addIssue({refId: this.state.referenceId, ratings: this.props.ratings, isRated: true, notes: this.state.notes})
```

local/vacols/bgs_setup.csv
leikkisa: Will clean this up for now. I learned that this file should be generated by cases.csv, but re-generating it requires using another RAILS_ENV and haven't gotten around to trying it yet :-P.

leikkisa:
Yeah I asked about displaying the notes in the UI afterwards and didn't get a response, but my intuition is that we should put them there.

leikkisa:
screen shot 2018-10-04 at 9 17 22 am

Issue #7000, opened 2018-09-14
AMA Intake | Clicking "Complete Intake" saves and processes claim review


shanear:
## Story
For any AMA forms, when a user needs to add issues, they will be walked through a multi-screen selection process to help them make the best choice. The landing page created in #6966 will be their starting point and a modal on top of it will walk the user through rated, non-rated, and an "unknown" option. 

This ticket is for when the user clicks "Complete Intake", which should save the intake, and attempt to process the appropriate requests to VBMS.

requires #6982

## Acceptance criteria
- When at least 1 issue is added, the "Complete Intake" button should not be disabled. 
- clicking the "Complete Issue" button on the Add Issues screens should save the all the issues listed on the page at the time. It should do this is in the same way as it did on the old select issues page for AMA reviews.

## Notes
Saving non-rating and unknown issues is out of scope and will be handled in future stories (#6988 and #6990)

## Mocks
<img width="686" alt="screen shot 2018-09-18 at 11 40 46 am" src="https://user-images.githubusercontent.com/11219417/45700725-4763fd80-bb3b-11e8-961b-a5afd972bf95.png">

## How to Validate
- Doesn't require a specific test user
- Form select > search id > review, any answers > on "Add issue page, click "+Add issue" > on modal, select a rated issue > click "Add this issue" button > click "Complete Intake"
- Go to VBMS and verify the EP was created with the correct contentions
- TBD: does this need to be tested with all 3 AMA forms or will one suffice?

shanear:
Validated this in all 3 lanes 👍

PR #7223, opened 2018-10-02
complete intake from add issues page

jiahuang:
Connects #7000

complete_intake


client/app/intake/reducers/common.js
```diff
@@ -13,14 +13,16 @@ export const commonReducers = (state, action) => {
...
 
     return {
       ...state,
-      addedIssues: [...listOfIssues, {
-        isRated: action.payload.isRated,
-        id: action.payload.issueId,
-        profileDate: action.payload.profileDate
-      }]
+      addedIssues,
+      issueCount: addedIssues.length
```
jiahuang: Note that this isn't currently used for anything. Right now we actually get the issueCount from calling `formatIssues` in `intakeCommon/util/index.js` since unrated issues need additional validation to get the count correct. Once unrated issues is added to the new add issues page, we can use issueCount instead of having to call `formatIssues` since validation on the unrated issue will already be done once the modal exits and the user is back on the main page.

leikkisa: Yay

client/app/intakeCommon/util/index.js
```diff
@@ -56,18 +56,30 @@ export const validNonRatedIssue = (issue) => {
   return true;
 };
 
+const getRatedIssues = (state) => {
+  if (state.addedIssues && state.addedIssues.length > 0) {
+    // we're using the new add issues page
+    return state.addedIssues.map((issue) => {
```
leikkisa: Since addedIssues is for all issues (I'm guessing since there is an isRated property), should this map through only the non-rated addedIssues?

spec/feature/intake/appeal_spec.rb
```diff
@@ -296,5 +296,9 @@ def check_row(label, text)
     safe_click ".add-issue"
 
     expect(page).to have_content("1. Left knee granted")
+
+    safe_click "#button-finish-intake"
+
+    expect(page).to have_content("Notice of Disagreement (VA Form 10182) has been processed.")
```
leikkisa: Can we add a check here that the issues are correct in the database?  I don't think it's necessary for all flows, but since there's a change to the function handling the data before it gets sent I think it'd be good to have one.

client/app/intakeCommon/util/index.js
```diff
@@ -56,18 +56,30 @@ export const validNonRatedIssue = (issue) => {
   return true;
 };
 
+const getRatedIssues = (state) => {
```
shanear: since this is the part of `formatIssues` that formats the rating issues, then it seems like `formatRatedIssues` might be a more appropriate name for this function

Issue #7218, opened 2018-10-02
[IDT] Add documents and assigned by name


aroltsch:
Return documents and assigned by name in the appeals details
Return documents and assigned by name in the appeals list when searching by file number

PR #7225, opened 2018-10-02
Return assigned by name and documents in the rest of IDT endpoints

aroltsch:
#7218


spec/factories/task.rb
```diff
@@ -64,7 +64,12 @@
       type "AttorneyTask"
       appeal_type "Appeal"
       appeal { create(:appeal) }
-      parent { create(:ama_judge_task) }
+
+      after(:create) do |task, _evaluator|
```
amprokop: hm. i think we already create parent tasks manually in the seeds file for attorney tasks. will this cause some confusion?

amprokop: i *think* this hook will override the parent tasks we create in the seed file, so it might break things. up to you to test.

aroltsch: good catch, we should only create it if it is not there

aroltsch: let me check a few things

PR #7231, opened 2018-10-02
Adds docket type to seeds, returns real citation number, returns empty response from /outcode instead of appeal details

amprokop:
Adds docket type to seeds, removes citation number method since we don't need it right now, returns empty response from bva dispatch outcode so we don't wait for BGS calls.

# Test plan

image


app/models/serializers/idt/v1/appeal_details_serializer.rb
```diff
@@ -72,7 +72,6 @@ def id
   attribute :file_number do
     object.is_a?(LegacyAppeal) ? object.sanitized_vbms_id : object.veteran_file_number
   end
-  attribute :citation_number
```
amprokop: we don't need to return this to the IDT anymore, the IDT passes it to us

app/models/appeal.rb
```diff
@@ -99,9 +99,8 @@ def appellant
 
   delegate :first_name, :last_name, :middle_name, :name_suffix, to: :appellant, prefix: true, allow_nil: true
 
-  # TODO: implement for AMA
   def citation_number
-    "not implemented for AMA"
+    decisions.present? ? decisions.sort_by(&:created_at).last.citation_number : nil
```
mdbenjam: Do we use this anywhere? If not, I wouldn't even implement it. If we use it, I feel we should rename the method to `latest_citation_number`.

amprokop: cool, that sounds fine, i can just delete this method. i'm thinking we'll need it pretty soon

Issue #7028, opened 2018-09-18
AMA Intake | Add Issue Landing page "remove" button


suzchap:
## Story
For any AMA forms, when a user needs to add issues, they will be walked through a multi-screen selection process to help them make the best choice. The landing page created in #6966 will be their starting point and a modal on top of it will walk the user through rated, non-rated, and an "unknown" option.

When a user adds an issue and it displays on the Add Issue Landing page, they are able to remove that issue if needed.

requires #6982 

## Technical notes
- trash icon: https://fontawesome.com/icons/trash-alt?style=regular&from=io

## Acceptance criteria
- When a user adds an issue and it appears on the Add Issue Landing page, they can then remove that issue by clicking "Remove" next to it. 
- Clicking the Remove link will make that issue go away (we aren't doing a warning modal here)

## Copy
- Remove

## Mocks
<img width="686" alt="screen shot 2018-09-18 at 11 40 46 am" src="https://user-images.githubusercontent.com/11219417/45699340-c2c3b000-bb37-11e8-87f9-31001efd213b.png">

## How to Validate
- Doesn't require a specific test user
- Form select > search id > review page: any answers > on "Add issue page, click "+Add issue" > Add a rated issue > back on the Add Issue Landing page click "remove" link & that issue should disappear.
- TBD: does this need to be tested with all 3 AMA forms or will one suffice?

shanear:
works like a charm 🍀

PR #7246, opened 2018-10-03
Remove issue button

pkarman:
connects #7028 

### Description
Adds a Remove button to each added issue so that it can be removed from the current list.

### Acceptance Criteria
- [x] Code compiles correctly
- [x] Each issue can be added and removed.

screen shot 2018-10-03 at 11 04 16 am


client/app/intake/reducers/common.js
```diff
@@ -24,6 +24,26 @@ export const commonReducers = (state, action) => {
...
+    let issueToRemove = action.payload.issue;
+    let i;
+
+    for (i = 0; i < listOfIssues.length; ++i) {
+      let issue = listOfIssues[i];
+
+      if (issue.id !== issueToRemove.referenceId) {
+        newListOfIssues.push(issue);
+      }
+    }
```
shanear: We can use `_.filter` here to achieve the same effect. Something like this:

```
return {
  ...state,
  addedIssues: _.filter(listOfIssues || [], (issue) => issueToRemove.referenceId != issue.id)
};
```

client/app/styles/_intake.scss
```diff
@@ -64,8 +64,55 @@
   td {
     vertical-align: top;
 
-    div {
-      padding-bottom: 0.5em;
+    table {
```
shanear: It'd be nice to be able to use the styleguide `<Table>` element if possible to avoid writing all of this css. I think we should aim to have as little custom intake css as possible

client/app/intake/pages/addIssues.jsx
```diff
@@ -27,19 +33,37 @@ class AddIssues extends React.PureComponent {
...
-        </Button>
-      </div>;
+      return <table className="usa-table-borderless issues">
+        <tbody>
+          { issues.map((issue, index) => {
+            return <tr className="issue" key={issue.referenceId}>
+              <td>{index + 1}. {issue.text}</td>
+              <td className="cf-text-r">
+                <Button
+                  onClick={() => this.onRemoveIssue(issue)}
```
jiahuang: if remove issue is just going to be a wrapper around a prop function, I'd rather just use the prop:

```
<Button onClick={() => this.props.removeIssue(issue)}
```

client/app/intake/actions/ama.js
```diff
@@ -156,6 +156,13 @@ export const addIssue = (issueId, ratings, isRated) => (dispatch) => {
   });
 };
 
+export const removeIssue = (issue) => (dispatch) => {
```
jiahuang: This can be simplified to

```
export const removeIssue = (issue) => ({
  type: ACTIONS.REMOVE_ISSUE,
  payload: { issue }
});

```

pkarman:
Thanks everyone for the review feedback. Changes in 1a536bf8f87cf91eb6426bfaf47652fa3ccc5191

pkarman:
@jiahuang @leikkisa this look ok to you?

pkarman:
going to wait for #7207 to get merged and see if we have conflicts here

pkarman:
resolved conflicts with master and added a little style to new notes section to visually set it apart. Happy to do follow-up PR to change styles @carolaponce @suzchap 

screen shot 2018-10-04 at 12 58 20 pm

suzchap:
Works for me. You're officially on the design team now. :)

Issue #7073, opened 2018-09-20
Video and CO Release | Epic Daily Docket | Default View


Chingujo:
As a hearing coordinator, I need to able to see all the veterans that are scheduled for a hearing that day. 

### Acceptance Criteria
- [x] Verify that the title is Daily Docket ( Day Date)
- [x] Verify that underneath the title there is a "back to schedule" link underneath the title
- [x] Verify that in the far right we have the following fields: 
 - VLJ: (not in hearing prep)
 - Coordinator: ( not in hearing prep) 
-  Hearing Type (in hearing prep)
  - This is a newer field in VACOLS that Jed added
- [x] Verify that we are using the Hearing Prep docket table with the following modifications to the date
   - Remove the check boxes next to the number
   - The Actions tab has the following: 
     - Hearing location (with a change link at the bottom) 
     - Hearing Day as a dropdown 
     - Disposition as a Drop down

### Mocks
**Video Default**
![daily docket default](https://user-images.githubusercontent.com/23316107/45948874-b623d980-bfc7-11e8-8131-9687c34eeb71.png)

#Related Issues
#6873 Epic Daily Docket

PR #7250, opened 2018-10-03
Daily docket UI

sharonwarner:
Connects #7073 

### Description
This PR adds the UI for the Daily Docket page. The information is currently stubbed out and not connected to Redux. Those updates will come in future PRs.

### Testing Plan
1. Go to /hearings/schedule/docket/RO04/2018-10-02
1. Confirm UI looks correct


client/app/components/Table.jsx
```diff
@@ -104,9 +104,15 @@ const HeaderRow = (props) => {
 };
 
 const getCellValue = (rowObject, rowId, column) => {
+  if (column.valueName && rowObject[column.valueName] && rowObject[column.valueName].span === 0) {
+    return '';
+  }
```
lowellrex: Can we take advantage of the [existing `getCellSpan()` function](https://github.com/department-of-veterans-affairs/caseflow/blob/0702ef5bfa02c4a3bc0951755341d207164bbe74/client/app/components/Table.jsx#L117) to achieve the same result? It looks like the `TaskTable` component [does something similar](https://github.com/department-of-veterans-affairs/caseflow/blob/0702ef5bfa02c4a3bc0951755341d207164bbe74/client/app/queue/components/TaskTable.jsx#L143) entirely within that component so maybe we could do something like the following in `DailyDocket`:
```
const dailyDocketColumns = [
  ...
  {
    header: 'Time/RO(s)',
    align: 'left',
    valueName: 'hearingTime',
    span: (row) => row.hideRepColumn ? 2 : 1
  },
  {
    header: 'Representative',
    align: 'left',
    valueName: 'representative',
    span: (row) => row.hideRepColumn ? 0 : 1
  },
  ...

# And add the hideRepColumn property to the last row we push onto the array in 
# getDailyDocketRows()
```

Would that get the same result this change is trying to achieve or am I missing some other behaviour here?

sharonwarner: I think that only defines the column span based on the column itself, but we need it defined by the row.

PR #7251, opened 2018-10-03
Add generic task ui

### LINKED ISSUE #6703: Add task actions dropdown for organizational queues

mdbenjam:
Resolves #6703

### Description
Add organizational tasks actions. Note we don't yet hide the actions dropdown when the task is on hold. For a future PR.

### Testing Plan
1. Go to /organization/translation with translation user
1. Assign task to user
1. Go to /queue
1. See task and assign to organization

genericui


app/models/tasks/generic_task.rb
```diff
@@ -17,13 +36,18 @@ def can_user_access?(user)
   end
 
   class << self
-    def create_from_params(params, current_user)
-      parent = Task.find(params[:parent_id])
-      parent.verify_user_access(current_user)
+    def create_from_params(params_array, current_user)
+      params_array.map do |params|
```
mdbenjam: The create function sends an array in.

mkhandekar:
Can we add a gif here to view functionality? Thank you!

mdbenjam:
@mkhandekar added a gif

app/models/organization.rb
```diff
@@ -2,16 +2,24 @@ class Organization < ApplicationRecord
   has_many :tasks, as: :assigned_to
   has_many :staff_field_for_organization
 
-  def self.assignable
-    where(type: [nil, BvaDispatch.name])
+  def self.assignable(task)
+    organizations = where(type: [nil, BvaDispatch.name])
+
```
lowellrex: Could we add a comment to the effect of "Exclude the current organization from the list of assignable organizations if the task is assigned to this organization or the task is a child of a task assigned to this organization. Prevents assignment loops."

I don't think it will be all that obvious what or why this conditional is doing what it is doing in a year's time.

client/app/queue/QueueApp.jsx
```diff
@@ -94,9 +96,10 @@ class QueueApp extends React.PureComponent<Props> {
       return <AttorneyTaskListView />;
     } else if (userRole === USER_ROLE_TYPES.judge) {
       return <JudgeReviewTaskListView {...this.props} />;
-    } else if (userRole === USER_ROLE_TYPES.colocated) {
-      return <ColocatedTaskListView />;
     }
+
+    return <ColocatedTaskListView />;
```
lowellrex: Tasks in this generic queue disappear after we create a child task for them ("assign task to team"). I think we need to update the [`taskIsOnHold()` method](https://github.com/department-of-veterans-affairs/caseflow/blob/f889bb09a9742e9df88391e750b37c4fc20941db/client/app/queue/utils.js#L425) to account for tasks that are on hold but do not have a `onHoldDuration` (which is the case in this instance). Happy to have that be part of a different PR though.

client/app/queue/AssignToView.jsx
```diff
@@ -0,0 +1,138 @@
...
+    } = this.props;
+    const payload = {
+      data: {
+        tasks: [{
+          type: 'GenericTask',
+          external_id: appeal.externalId,
+          parent_id: task.taskId,
+          assigned_to_id: this.state.selectedValue,
+          assigned_to_type: isTeamAssign ? 'Organization' : 'User'
+        }]
```
lowellrex: Do we want to pass in a `status` parameter here so we change the status of parent task to `on_hold` when we create a child task?

client/app/queue/QueueApp.jsx
mdbenjam: Yeah I've noticed that too. I want to fix that in the next PR.

client/app/queue/AssignToView.jsx
mdbenjam: I thought this automatically happened 🤔 guess not. Should we pass the status? Or just set this up in the generic task?

app/models/organization.rb
mdbenjam: added

client/app/queue/AssignToView.jsx
mdbenjam: Added the code for this

Issue #7211, opened 2018-10-02
Add tab for "Completed" and "On hold" tasks


lowellrex:
When somebody marks a task as complete that case disappears from their queue because we filter out completed tasks from the task table. Our co-located pilot folks have indicated that they often want to reference those cases for a little while after they have marked them as complete (for sending emails, making notes, charging the case back to themselves, etc.). We should allow co-located (and organizational) task queue folks to see tasks they have recently (past 30 days? past 20 tasks?) completed.

In a similar vein cases disappear from attorneys' queues if they have been assigned to co-located. We should allow attorneys (and organizational queue folks) to see tasks that they have assigned to somebody else so that they do not lose visibility on a case that they have some responsibility for. I imagine this could be accomplished in another tab similar to the co-located "on hold" tab (since that is essentially the state those tasks are in).

## Acceptance criteria
* Enable folks to see tasks they have recently completed
* Enable folks to see tasks that do not require action of them at the current moment (on hold, assigned to a different team)

lowellrex:
I believe the only work left to be done here is to add the "on hold"-ish tab to generic queues.

PR #7252, opened 2018-10-03
Add tab for completed colocated tasks

lowellrex:
Connects #7211. Returns recently completed tasks for co-located queues and adds a tab for those recently completed tasks.

### TODO:
* Write tests

completed_tab

### DIFF
--- a/app/models/queues/colocated_queue.rb
+++ b/app/models/queues/colocated_queue.rb
@@ -3,19 +3,29 @@ class ColocatedQueue
 
   attr_accessor :user
 
-  # rubocop:disable Rails/FindEach
   # rubocop:disable Style/SymbolProc
   def tasks
-    incomplete_tasks.where(assigned_to: user).each { |t| t.update_if_hold_expired! }
+    relevant_tasks.select { |t| t.assigned_to == user }.each { |t| t.update_if_hold_expired! }
   end
 
   def tasks_by_appeal_id(appeal_id, appeal_type)
-    incomplete_tasks.where(appeal_id: appeal_id, appeal_type: appeal_type).each { |t| t.update_if_hold_expired! }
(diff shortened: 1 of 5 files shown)


app/models/queues/colocated_queue.rb
```diff
@@ -3,20 +3,30 @@ class ColocatedQueue
...
-  # rubocop:disable Rails/FindEach
   # rubocop:disable Style/SymbolProc
   def tasks
-    incomplete_tasks.where(assigned_to: user).each { |t| t.update_if_hold_expired! }
+    relevant_tasks.select { |t| t.assigned_to == user }.each { |t| t.update_if_hold_expired! }
   end
 
   def tasks_by_appeal_id(appeal_id, appeal_type)
-    incomplete_tasks.where(appeal_id: appeal_id, appeal_type: appeal_type).each { |t| t.update_if_hold_expired! }
+    tasks = relevant_tasks.select { |t| t.appeal_id == appeal_id && t.appeal_type == appeal_type }
```
mdbenjam: Not a huge deal since I think we're dealing with small arrays here, but it would probably be better to make relevant_tasks an ActiveRecord statement, so we could continue to run the where rather than select on the app layer. But doesn't really matter for this case.

app/models/queues/colocated_queue.rb
```diff
@@ -3,20 +3,30 @@ class ColocatedQueue
...
   # rubocop:enable Style/SymbolProc
-  # rubocop:enable Rails/FindEach
 
   private
 
+  def relevant_tasks
+    [incomplete_tasks, recently_completed_tasks].flatten
+  end
+
+  def recently_completed_tasks
```
mdbenjam: I think this should go in task, not colocated_queue. We'll want this for generic tasks.

PR #7259, opened 2018-10-04
Throw error if trying to create duplicate das records

aroltsch:
When a judge assigns a case to attorney, make sure this case is still assigned to the judge.


app/controllers/issues_controller.rb
```diff
@@ -48,7 +48,7 @@ def json_issues
   end
 
   def validate_access_to_task
-    current_user.access_to_legacy_task?(appeal.vacols_id)
+    current_user.fail_if_no_access_to_legacy_task!(appeal.vacols_id)
```
aroltsch: Renamed this method on the user object to make it more accurate. It doesn't return boolean so `?` is not correct. It throws an exception so `!` is more accurate.

app/models/judge_case_assignment_to_attorney.rb
```diff
@@ -59,7 +59,10 @@ class << self
 
     def create(task_attrs)
       task = new(task_attrs)
-      task.assign_to_attorney! if task.valid?
+      if task.valid?
+        task.assigned_by.fail_if_no_access_to_legacy_task!(task.vacols_id)
```
aroltsch: Check that appeal is still assigned to a judge on the assign action only.

app/controllers/issues_controller.rb
amprokop: ❗️

app/models/judge_case_assignment_to_attorney.rb
amprokop: 👍

PR #7262, opened 2018-10-04
sets button into loading state when waiting

amprokop:
After a judge clicked the "Assign N Cases" button, when waiting for the VACOLS judge assignment to attorney query to finish, we were not putting the button into a disabled "loading" state. This resulted in a case where multiple clicks on the button could result in erroneous duplicate DECASS records.

This PR fixes that issue by preventing multiple clicks from triggering multiple assignments.

# Testing

- added `sleep 10`in the transaction block of `QueueRepository.assign_case_to_attorney!`

assignment

### DIFF
--- a/client/app/queue/components/AssignWidget.jsx
+++ b/client/app/queue/components/AssignWidget.jsx
@@ -122,7 +122,6 @@ class AssignWidget extends React.PureComponent<Props> {
   }
 
   render = () => {
-
     const {
       attorneysOfJudge,
       selectedAssignee,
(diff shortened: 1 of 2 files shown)


client/app/queue/components/AssignWidget.jsx
```diff
@@ -92,19 +94,31 @@ class AssignWidget extends React.PureComponent<Props> {
   assignTasks = (selectedTasks: Array<Task>, assigneeId: string) => {
     const { previousAssigneeId } = this.props;
 
+    if (!this.props.savePending) {
```
mdbenjam: nit: I don't think you really need this if

amprokop: good point

Issue #7255, opened 2018-10-03
AMA Intake | Update form names


shanear:
The form names have changed! We need to update them intake to match the names below:

## Acceptance Criteria
- The form types selection should represent the new form names (shown below) for the AMA forms. Those names should be reflected other places where the form is used (such as the review page)

## Copy
Decision Review Request: Supplemental Claim — VA Form 20-0995
Decision Review Request: Higher-Level Review — VA Form 20-0996
Decision Review Request: Board Appeal (Notice of Disagreement) — VA Form 10182

shanear:
![screen shot 2018-10-05 at 2 27 48 pm](https://user-images.githubusercontent.com/1596259/46554025-0c88f600-c8ad-11e8-9d41-276018900660.png)
Checked here as well as a bunch more places. Looks great 👍

PR #7265, opened 2018-10-04
Update intake form names

pkarman:
connects #7255 

### Description
Update Form names and use the constants everywhere.

### Acceptance Criteria
- [x] Code compiles correctly
- [x] Tests continue to pass


pkarman:
oh I like that @leikkisa -- thanks for pointing it out.

PR #7267, opened 2018-10-04
Fix disposition sync

shanear:
This was happening because we were not correctly mocking connectVBMS. ConnectVBMS
returns a list of `Disposition` objects, not contentions.

connects #6186

### DIFF
--- a/lib/generators/contention.rb
+++ b/lib/generators/contention.rb
@@ -25,11 +25,11 @@ class Generators::Contention
         if disposition
           Fakes::VBMSService.disposition_records ||= {}
           Fakes::VBMSService.disposition_records[claim_id] ||= []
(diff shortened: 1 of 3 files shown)


lib/generators/contention.rb
```diff
@@ -25,11 +25,11 @@ def build(attrs = {})
         if disposition
           Fakes::VBMSService.disposition_records ||= {}
           Fakes::VBMSService.disposition_records[claim_id] ||= []
-          Fakes::VBMSService.disposition_records[claim_id] << {
+          Fakes::VBMSService.disposition_records[claim_id] << OpenStruct.new(
```
kencheeto: ok so instead of hashes we are expecting objects? how does that map to `ConnectVBMS
returns a list of Disposition objects, not contentions.` ?

Issue #7201, opened 2018-10-01
Edit VBA Issues | Too many updates too fast can cause bugs


shanear:
When validating #6212, I noticed that submitting 2 or more edits in rapid succession can cause the processing of `RequestIssuesUpdate` to bug out (not being able to find contentions because they were already removed).

I think (?) the solution here might be to prevent going to the edit issues screen if there is a `RequestIssuesUpdate` that is pending processing. There may be other solutions too

jiahuang:
We discussed this and came to the conclusion that on submit we should:

1. Check the submitted_at & processed_at fields for a RequestIssuesUpdate
2. If the submitted_at field is filled but processed_at is not, make the submit action a no-op and flash an error message telling the user to wait a few min.

pkarman:
Tech implementation note: could probably add an `submitted?` method to the `app/models/concerns/asyncable.rb` concern so you end up with:

```ruby
if submitted? && !processed?
```

shanear:
![screen shot 2018-10-10 at 3 35 32 pm](https://user-images.githubusercontent.com/1596259/46761369-48073400-cca2-11e8-913f-c10d3c955bbb.png)
nice!

Issue #6944, opened 2018-09-12
Page to view and retry failed asyncable operations


shanear:
We need a landing page to see all the HLRs and SCs that failed to establish, and potentially an error message about what went wrong. This way a manager can look through them, diagnose the issue and attempt to retry processing them.

## Acceptance Criteria
- There should be a page that shows all Asyncable models that have been attempted, but not successful. It should show the following information about the model:
  - The type of operation
  - The error message
  - Started at time
  - Last attempted at time
  - the veteran file number (if applicable)

## Notes
- We should determine a URL for this page.
- This page should be accessible to anyone with the "Admin Intake" permission

## Mocks
- This should use the standard style guide table for now. We can improve the styling later.

requires #6942

kencheeto:
If it's quick to add sorting to the table, that would be nice

shanear:
<img width="1552" alt="screenshot 2019-01-22 10 48 28" src="https://user-images.githubusercontent.com/1596259/51558041-52090900-1e33-11e9-8ccc-8ccac63d39b9.png">
If there are no failed jobs it looks like this. A small message might be helpful?

shanear:
We can pull it out into a seperate ticket if you want, but might be pretty easy

shanear:
Hmm... giving me an error on UAT:

<img width="1552" alt="screenshot 2019-01-22 10 57 52" src="https://user-images.githubusercontent.com/1596259/51558591-afea2080-1e34-11e9-8a43-8d12e0de5d56.png">

shanear:
<img width="1552" alt="screenshot 2019-01-22 10 59 44" src="https://user-images.githubusercontent.com/1596259/51558653-dad47480-1e34-11e9-938a-026fb0847f1f.png">

https://sentry.uat.ds.va.gov/sentry/caseflow-uat/issues/1258/

pkarman:
failing due to bad data bug reported in #8802

shanear:
woot this is working.
<img width="1552" alt="screenshot 2019-01-23 16 18 50" src="https://user-images.githubusercontent.com/1596259/51645773-93c8ab00-1f2a-11e9-8339-ac6902cd830f.png">

PR #7274, opened 2018-10-04
adding check for in progress updates

jiahuang:
Connects #7201

image

### Testing Plan
1. Manually removed the processed_at field for an existing update
2. tried to update
3. verified that the error message showed up and that the update did not go through

To do
- [x] add tests


client/app/intakeEdit/components/RequestIssuesUpdateErrorAlert.jsx
```diff
@@ -12,6 +12,10 @@ export default class RequestIssuesUpdateErrorAlert extends React.PureComponent {
         title: 'No changes were selected',
         body: 'Please select at least one change and try again.'
       },
+      previous_update_not_done_processing: {
+        title: 'Previous update not yet done processing',
+        body: 'A previously submitted update has not yet finished processing. Please wait a few minutes and try again.'
```
pkarman: depending on *why* it has not finished, it might be hours, not minutes. Maybe instead we could say "Please try back later."

Or, we can leave it as is until #6944 is more thought through.

app/models/request_issues_update.rb
```diff
@@ -96,6 +96,8 @@ def validate_before_perform
       @error_code = :request_issues_data_empty
     elsif !changes?
       @error_code = :no_changes
+    elsif not RequestIssuesUpdate.find_by(review: review, processed_at: nil).nil?
```
pkarman: is there ever a case where `submitted_at` is not null? I'm not sure of the edge cases here.

shanear: Not currently, but i also think its probably just safer to use the `Asyncable` scope `processable`

client/app/intakeEdit/components/RequestIssuesUpdateErrorAlert.jsx
shanear: I think the current wording is correct for the happy path. We don't really want them to context switch before trying again if they don't need to. However, if there was an error, we'd really want them to contact support. Maybe a message like this?

"A previously submitted update has not yet finished processing. Please wait a few minutes and try again.

If it's still not working after that, please contact customer support"

app/models/request_issues_update.rb
```diff
@@ -96,6 +96,8 @@ def validate_before_perform
       @error_code = :request_issues_data_empty
     elsif !changes?
       @error_code = :no_changes
+    elsif !RequestIssuesUpdate.where(review: review).processable.empty?
```
shanear: can use `.exists?` to prevent the double negative:

```
https://apidock.com/rails/ActiveRecord/Base/exists%3F/class
```

PR #7214, opened 2018-10-02
Judge assign modal

mdbenjam:
Connects #7102

### Description
Move the assign widget to a modal on the case details page. Also noticed we were only passing one task to the case details page. We want to return all tasks.

Note: Assigning the task to an `other` attorney, and then trying to assign to a different `other` attorney fails because the assign/reassign logic doesn't work in that case. But I think this is probably broken on master as well.

### Testing Plan
1. Login as a judge
1. Go to the assign queue
1. Click on an appeal
1. Select the task `assign to user`
1. Assign to a user in the modal

PR #7275, opened 2018-10-04
Queue | Colocated and Judge generic actions

dannysellers:
### Description
Moves Judge and Colocated tasks over to the `available_actions` generic task model

### Acceptance Criteria
- [ ] ActionsDropdown options for Judges and Colocated users begin checkout flows / navigate to modals as expected

```ruby
routes = {
  assignToUser: "/modal/assign_to_user" (#7214),
  sendToTeam: "/modal/sendToTeam",
  sendToAttorney: "/modal/sendToAttorney",
  placeHold: "place_hold",
  evaluateDecision: "omo_request/evaluate",
  selectDispositions: "dispatch_decision/dispositions",
  specialIssues: "dispatch_decision/special_issues",
}
```

| Docket | Role | Action | Next Route |
|--------|------|--------|------------|
| Legacy | Judge | Assign | `assignToUser` |
| | | Review | `evaluateDecision \|\| selectDispositions` |
| AMA | Judge | Assign | `assignToUser` |
| | | Review | `specialIssues` |
| * | Colocated | * | `sendToAttorney \|\| placeHold` |
| Legacy | Colocated | Translation/Schedule Hearing | `sendToTeam` |

### Testing Plan
1. Go to `/queue` as `BVAAABSHIRE`
2. Confirm correct dropdown options appear in Case Details view

#### Known Issues
Judge assignment to attorney does not work yet (#7214)


client/app/queue/components/ActionsDropdown.jsx
```diff
@@ -54,8 +54,11 @@ class ActionsDropdown extends React.PureComponent<Props> {
     this.props.stageAppeal(appealId);
     this.props.resetDecisionOptions();
 
-    history.push('');
-    history.replace(`/queue/appeals/${appealId}/${option.value}`);
+    const path = option.value.startsWith('/') ?
+      option.value :
+      `/appeals/${appealId}/${option.value}`;
+
+    history.push(`/queue${path}`);
```
lowellrex: Can we avoid this condition if we just always have the tasks' `allowed_actions()` methods return the full value of the path? The tasks should have access to the appeal to which they are attached on the backend so I think we have all the information we need.

client/app/queue/constants.js
```diff
@@ -183,6 +170,7 @@ export const PAGE_TITLES = {
 
 export const COLOCATED_HOLD_DURATIONS = [15, 30, 45, 60, 90, 120, 'Custom'];
 
+// todo: make shared constant? consolidate w/ColocatedTask.available_actions
```
lowellrex: Yes please!

dannysellers: I'm happy to move these into a shared constant, but I suspect they might change soon, and it might be better to consolidate all of these task routes into a constant (it'd be really great to have all queue routes in a constant, but that might be overkill).

For @mdbenjam: we currently have two ways of rendering modals:
- `/queue/modal/:modalType`, which sets `modals` in Redux and redirects to `/queue/appeals/:appealId`
- `/queue/appeals/:appealId/:modalType`, which renders modal views (i.e. `assign_to_team`) over top of the case details view

These modals (`SendToLocationModal`) are of the first type, while the edit AOD / generic task send to team/person modals (based on `editModalBase`) are of the second. I'm happy to help consolidate these, but I'm not really familiar with the `AssignToView` modals. Also, we have to figure out how to integrate [judge assignment](https://github.com/department-of-veterans-affairs/caseflow/pull/7214) into this url scheme

lowellrex: Makes sense to me!

client/app/queue/components/ActionsDropdown.jsx
```diff
@@ -54,8 +54,7 @@ class ActionsDropdown extends React.PureComponent<Props> {
     this.props.stageAppeal(appealId);
     this.props.resetDecisionOptions();
 
-    history.push('');
-    history.replace(`/queue/appeals/${appealId}/${option.value}`);
+    history.push(`/queue${option.value}`);
```
lowellrex: 👍

lowellrex: Just curious, was there some reason why we pushed an empty string onto the browser history and then placed that entry with the correct route? Do you know if it provides some type of side benefit with react router?

dannysellers: Yeah, I believe I added that initially, but no, I can't think of any reason to push `''` and then replace w/the real url 🤷‍♂️

dannysellers:
FYI this shouldn't be merged yet: Judge assignment doesn't work, is pending resolution of #7214 / some other way to integrate case assignment into the url scheme

client/app/queue/components/ActionsDropdown.jsx
```diff
@@ -54,8 +54,7 @@ class ActionsDropdown extends React.PureComponent<Props> {
     this.props.stageAppeal(appealId);
     this.props.resetDecisionOptions();
 
-    history.push('');
-    history.replace(`/queue/appeals/${appealId}/${option.value}`);
+    history.push(`/queue/appeals/${appealId}/${option.value}`);
```
mdbenjam: Changing this guy up since we use `goBack` when we cancel a modal, and we want to go back to the case details page. And use `replace` on a successful action.

client/app/queue/components/EditModalBase.jsx
```diff
@@ -19,6 +23,10 @@ export default function editModalBase(ComponentToWrap, { title, button, propsToT
 
     getWrappedComponentRef = (ref) => this.wrappedComponent = ref;
 
+    cancelHandler = () => {
+      this.props.history.goBack();
```
mdbenjam: Separating out the cancel handler from the close handler as they now go different places... But may not in the future!

PR #7281, opened 2018-10-04
Add QR team and task

lowellrex:
Creates a QR organization, QR-specific task, and creates a task for QR instead of BVA dispatch when a judge completes case review. Related slack discussion: https://dsva.slack.com/archives/C6E41RE92/p1538686092000100

### DIFF
--- a/app/models/tasks/quality_review_task.rb
+++ b/app/models/tasks/quality_review_task.rb
@@ -1,0 +1,10 @@
+class QualityReviewTask < GenericTask
+  def create_from_root_task(root_task)
+    create!(assigned_to: QualityReview.singleton, parent_id: root_task.id, appeal: root_task.appeal)
+  end
+
+  def mark_as_complete!
+    BvaDispatchTask.create_and_assign(root_task)
+    super
+  end
+end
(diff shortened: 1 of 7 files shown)


app/models/tasks/quality_review_task.rb
```diff
@@ -0,0 +1,10 @@
+class QualityReviewTask < GenericTask
+  def create_from_root_task(root_task)
+    create!(assigned_to: QualityReview.singleton, parent_id: root_task.id, appeal: root_task.appeal)
+  end
+
+  def mark_as_complete!
+    BvaDispatchTask.create_and_assign(root_task)
+    super
+  end
```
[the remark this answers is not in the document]
mdbenjam: So easy, let's hear what they want and then decide

db/migrate/20181004221403_add_quality_review_org_row.rb
```diff
@@ -0,0 +1,9 @@
+class AddQualityReviewOrgRow < ActiveRecord::Migration[5.1]
```
amprokop: Just a post-merge question — was this DB migration strictly necessary?

amprokop: It seems like `QualityReview.singleton` would create the row the first time it was called.

lowellrex: Correct. This is the pattern we've been doing with the other "singleton" database rows for organizations (BvaDispatch, Bva) so I just continued it here. Happy to stop doing that though.

PR #7283, opened 2018-10-04
fix nonrated issue styling

kencheeto:
### Description
- Noticed that the styling for the nonrated issue list changed (`usa-grid` was added in https://github.com/department-of-veterans-affairs/caseflow/pull/7189) but it's inconsistent with the rest of the page, so I'm removing that class here. (@nolawi please let me know if that is incorrect)
~- Completely unrelatedly, adding the has_many_ratings bgs_key to a vet in bgs_setup.csv since it's useful for developing Intake to have vets with ratings. (@sharonwarner these bgs_key values were all removed from this file in https://github.com/department-of-veterans-affairs/caseflow/pull/7151, not sure if that was required or not, is this okay?)~ nevermind, this is happening in https://github.com/department-of-veterans-affairs/caseflow/pull/7287

intake before/after:
demo

### Acceptance Criteria
- [x] Code compiles correctly

### Testing Plan
N/A


jiahuang:
I had a convo with Sharon earlier and she said that the bgs.csv is generated from a script that runs with cases.csv as it's input. So we should make the updates properly in there & the generation scripts to avoid bgs.csv from getting wiped again.

kencheeto:
wow thanks for digging, here's the pr that will fix the root cause: https://github.com/department-of-veterans-affairs/caseflow/pull/7287

PR #7284, opened 2018-10-05
Fix DTA error strings

shanear:
Apparently there are two different kinds of hyphens :(

Also adding some refactoring explained in PR comments.

connects #6168


app/models/claim_review.rb
```diff
@@ -89,11 +89,9 @@ def end_product_establishment_for_issue(issue)
 
   def sync_dispositions(reference_id)
     fetch_dispositions_from_vbms(reference_id).each do |disposition|
-      request_issue = matching_request_issue(disposition.contention_id)
-      request_issue.update!(disposition: disposition.disposition)
-
-      # allow higher level reviews to do additional logic on dta errors
-      yield(disposition, request_issue) if block_given?
```
shanear: We no longer need to tally up the DTA errors at the time of sync, removing the need for this callback

app/models/higher_level_review.rb
```diff
@@ -7,7 +7,7 @@ class HigherLevelReview < ClaimReview
   END_PRODUCT_RATING_CODE = "030HLRR".freeze
   END_PRODUCT_NONRATING_CODE = "030HLRNR".freeze
   END_PRODUCT_MODIFIERS = %w[030 031 032 033 033 035 036 037 038 039].freeze
-  DTA_ERRORS = ["DTA Error – PMRs", "DTA Error – Fed Recs", "DTA Error – Other Recs", "DTA Error – Exam/MO"].freeze
+  DTA_ERRORS = ["DTA Error - PMRs", "DTA Error - Fed Recs", "DTA Error - Other Recs", "DTA Error - Exam/MO"].freeze
```
shanear: yep `–` is different from `-`

😑

app/models/request_issue.rb
```diff
@@ -14,6 +14,10 @@ def self.nonrated
       .where.not(issue_category: [nil, "Unknown issue category"])
   end
 
+  def self.no_follow_up_issues
+    where.not(id: select(:parent_request_issue_id).uniq)
+  end
```
shanear: Rails magic converts this all to one SQL statement ✨which should be fast due to the added index.

app/models/higher_level_review.rb
```diff
@@ -88,16 +90,8 @@ def build_follow_up_dta_issues
...
-      if DTA_ERRORS.include?(disposition.disposition)
-        dta_issues << request_issue
-      end
-    end
-  end
-
-  def dta_issues
-    @dta_issues ||= []
+  def dta_issues_needing_follow_up
+    @dta_issues_needing_follow_up ||= request_issues.no_follow_up_issues.where(disposition: DTA_ERRORS)
```
shanear: instead of tallying these at the time of sync, it's a little more robust to lazily load them. That way there are no sequential dependencies for this method.

db/schema.rb
```diff
@@ -602,6 +602,7 @@
     t.integer "parent_request_issue_id"
     t.text "notes"
     t.index ["end_product_establishment_id"], name: "index_request_issues_on_end_product_establishment_id"
+    t.index ["parent_request_issue_id"], name: "index_request_issues_on_parent_request_issue_id"
```
pkarman: is there supposed to be a migration file with this?

app/models/higher_level_review.rb
pkarman: is it worth adding a comment about the code point for this "new" hyphen? It looks like ASCII `\x2d` instead of Unicode `\xe2 \x80 \x93`

pkarman: would it be worth defining each string as a separate constant so that the tests can evaluate constants instead of strings? just thinking about long-term local maintenance of what seem to be upstream string values.

db/schema.rb
shanear: nice catch. I'm so bad at remembering that 😑

pkarman:
awesome @shanear looks great

PR #7285, opened 2018-10-05
break out linting and js tests into separate circleci jobs

kencheeto:
### Description
I noticed that our parallelization on Circle CI could be more efficient:

- linting happens in every container (so with 5x parallelization, 5x linting per build)
- mocha js tests are parallelized, but there are only 3 test files right now, and they run super quickly
- karma js tests only happen in the primary container but are a blocking step, with ~50% of that time going into a webpack build

These fall into two main issues: **repeated work** and **blocking steps**. I was reading about Circle CI's [workflows](https://circleci.com/docs/2.0/workflows/) feature and it sounded like a good way to speed up our builds by breaking out sequential blocking tasks out of the critical path of the primary container. In order to do this, the old `build` job is broken out into 3 jobs:

1. `lint` - just the dependencies needed for linting, runs danger/lint/security checks, takes about 2 minutes
1. `js_tests` - runs all js tests, which are fast, totaling about 2 minutes
1. `rspec` - everything else from the old `build` job; gets the entire app setup and distributes rspec tests across 5 containers

And instead of triggering one `build` job on each push to a PR, we will now kick off a `workflow` which runs the three jobs above in parallel.

### Results

These changes take the total build time from **~14 minutes to ~9 minutes, or 35% faster builds** by reducing duplicated effort (2 minutes of linting each on n-1 containers) and getting as much off the critical path of the slow ruby tests as possible. 

You can see the new workflow timings here: https://circleci.com/gh/department-of-veterans-affairs/workflows/caseflow/tree/kenshiro-circleci-workflow

One additional bonus of breaking linting and js_tests into separate jobs is that we can retry specific jobs within a workflow.

### Caveats

Each test run will try to consume 7 containers (5x rspec, 1x lint, 1x js_tests) instead of 5, so there will be higher container utilization per build, but the overall decrease in work time should lead to less contention.

It's also possible that some things are wonky with artifact collection / test summaries and the circleci config could probably be cleaned up, but things are passing and I will keep an eye on any issues that come up.

### Acceptance Criteria
- [x] Tests pass

### Testing Plan
N/A


lib/tasks/security.rake
```diff
@@ -16,7 +16,7 @@ task :security_caseflow do
   Time.zone = "Eastern Time (US & Canada)"
 
   # Only ignore this vulnerability for a week.
-  audit_cmd = "bundle-audit check --ignore CVE-2018-1000201"
+  audit_cmd = "bundle-audit check --ignore CVE-2018-14404"
```
kencheeto: bundle-audit started failing on CVE-2018-14404 as of ~8 hours ago because nokogiri released a new update with security fixes for this vulnerability. I'm ignoring the failure here but I've another PR that updates nokogiri: https://github.com/department-of-veterans-affairs/caseflow/pull/7286

kencheeto:
screen shot 2018-10-05 at 1 53 10 am


this is stuck as pending [because it's a Circle CI 1.0 check - a repo admin will need to disable it](https://circleci.com/docs/2.0/workflows-waiting-status/) here once we're happy with this PR and want to merge it to master:

lib/tasks/security.rake
kencheeto: this is outdated, i've rebased off master now which has patched the vuln

PR #7305, opened 2018-10-05
Move case search to link in Navigation bar

### LINKED ISSUE #7016: Streamline case search ux
**https://github.com/department-of-veterans-affairs/caseflow/pull/6932 "Fix case search access control"** inspired a couple of conversations about simplifying how users experience search in Queue as well as how things are organized on the back end. 

## Photos from conversation:

image from ios

image from ios 1

## AC

- search should appear on case details using "view all cases" style expand/hide functionality (on both case details and table view) with additional explanatory text
- search should appear on table view for 
  - queue/reader user who has cases
  - queue/reader user who doesn't have cases at a given point
  - search/reader user who doesn't have cases

lowellrex:
Connects #7016.

case_search_update


client/app/queue/CaseListTable.jsx
```diff
@@ -30,12 +33,15 @@ const labelForLocation = (locationCode, userId) => {
 };
 
 class CaseListTable extends React.PureComponent {
+  componentWillUnmount = () => this.props.clearCaseListSearch();
```
lowellrex: Moved this into the table component so we don't clear the search field when we load the search results page.

mkhandekar:
🎉 

1. Can you update the instructions text per screenshots? 

   ```Please enter a valid 9-digit Veteran SSN or claims file number to search for all available cases.```

2. Also, can we show which lane the case is part of (using our little gray icons before the docket number)?

amprokop:
@mkhandekar — can we *always* search by SSN? My impression was that SSN only works if it is the file number as well. @lowellrex, can you confirm?

If we can't always search by SSN but we can always search by file number, should we leave SSN out of the modified copy?

mkhandekar:
I included it to make the instruction as specific as possible. If it is indeed true that a Veteran will always have a file number (and only sometimes an SSN), then I'm 👍 to removing it or switching the order in which it is said in the sentence!

amprokop:
```Please enter a valid 9-digit Veteran claims file number to search for all available cases.```?

File: client/app/components/RenderFunctions.jsx [hunk 2]
```diff
@@ -355,6 +355,10 @@ export const downloadIcon = (color = COLORS.WHITE) => {
   </svg>;
 };
 
+export const MagnifyingGlass = ({ color = COLORS.GREY_MEDIUM }) => <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 216 146">
+  <path fill={color} d="M172.77 123.025L144.825 95.08c6.735-9.722 10.104-20.56 10.104-32.508 0-7.767-1.51-15.195-4.527-22.283-3.014-7.09-7.088-13.2-12.22-18.336s-11.243-9.207-18.33-12.22c-7.09-3.016-14.52-4.523-22.286-4.523-7.768 0-15.196 1.508-22.284 4.523-7.09 3.014-13.2 7.088-18.332 12.22-5.132 5.134-9.206 11.245-12.22 18.333-3.015 7.088-4.522 14.515-4.522 22.282 0 7.766 1.507 15.192 4.522 22.282 3.014 7.088 7.088 13.197 12.22 18.33 5.134 5.134 11.245 9.207 18.333 12.222 7.09 3.015 14.516 4.522 22.283 4.522 11.95 0 22.786-3.37 32.51-10.105l27.943 27.863c1.955 2.064 4.397 3.096 7.332 3.096 2.824 0 5.27-1.03 7.332-3.096 2.064-2.063 3.096-4.508 3.096-7.332 0-2.877-1.002-5.322-3.013-7.33zm-49.413-34.668C116.214 95.5 107.62 99.07 97.57 99.07c-10.048 0-18.643-3.57-25.786-10.713C64.64 81.214 61.07 72.62 61.07 62.57c0-10.047 3.572-18.643 10.714-25.785C78.926 29.642 87.522 26.07 97.57 26.07c10.048 0 18.643 3.573 25.787 10.715 7.143 7.142 10.715 15.738 10.715 25.786 0 10.05-3.573 18.647-10.715 25.79z" />
```
amprokop: 👍

mkhandekar:
@lowellrex i remember you saying the file number is not always 9 digits? can you refresh my memory?

lowellrex:
We are always searching by "file number" which is either the Veteran's 9 digit SSN or the 7 or 8 digit claims file number. I think we should update the text to read something like:
> Please enter a valid Veteran ID (claims file number or 9-digit SSN) to search for all available cases

What do y'all think?

lowellrex:
@mkhandekar, I updated the gif to include the current copy and docket type badge.

amprokop:
Hmm. VA folks are used to using the file number to search in VACOLS and in VBMS — can we just say "claims file number" and be done with it? Do we need to train our users re: what this means on this page?

lowellrex:
That works for me!

mkhandekar:
Works for me, thank you!

client/app/queue/CaseListTable.jsx
amprokop: 👍

client/app/components/CaseSearchLink.jsx
```diff
@@ -0,0 +1,32 @@
...
+const textStyling = css({
+  color: COLORS.PRIMARY,
+  display: 'table-cell',
+  fontSize: '1.7rem',
+  lineHeight: '4em',
+  marginBottom: 0
+});
+
+const CaseSearchLink = () => <div {...containerStyling}>
+  <Link to="/search">
```
dannysellers: Hey, this only works for `BrowserRouter` instances with no `basename` that have a `PageRoute` of `"/search"`. In Hearing Scheduling, for example, which has a [basename of `"/hearings"`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/hearingSchedule/HearingScheduleApp.jsx#L34), this will direct the user to `"/hearings/search"`.

We could change the Link property to `href`, which would [return an `<a/>` rather than a `<RouterLink/>`](https://github.com/department-of-veterans-affairs/caseflow-frontend-toolkit/blob/master/components/Link.jsx#L99), but then the Queue -> Search behavior would be worse...

lowellrex: I think it might make sense to add a property on `CaseSearchLink` that changes the `to` to an `href` so we get the best of both worlds. Similar to what we did for the logo in the navigation bar: https://github.com/department-of-veterans-affairs/caseflow-frontend-toolkit/pull/43/files#diff-721c0ca1751e1fb73eafb435684e8d1bR60

How does that approach strike you?

dannysellers: That works for me, though I guess it means we'll have to have that logic be based on the route? Something like
```jsx
const linkProps = {
  [window.location.pathname.includes('queue') ? 'to' : 'href']: '/search'
};

<Link {...linkProps}>
```

lowellrex: I imagine other products that were including the search link in the navigation bar would pass that as a property. Something like `<CaseSearchLink outsideCurrentRouter />` and the component would have some logic like the following:
```
const linkProps = {
  [this.props.outsideCurrentRouter ? 'href' : 'to']: '/search'
};

<Link {...linkProps}>
```

dannysellers: Sure, though it looks like `NavigationBar` is set up to always display `CaseSearchLink`. As a result, the link will appear [all over Caseflow](https://github.com/department-of-veterans-affairs/caseflow/search?q=components%2FNavigationBar&unscoped_q=components%2FNavigationBar), and will take the user to a blank page

<details><summary>case search link in reader</summary>

</details><br/>

I'm happy to start on a fix, but do you have a preference between adding a `searchOutsideCurrentRouter`-type prop to `NavigationBar`, or doing it on the fly using the route?

Issue #7306, opened 2018-10-05
AMA Intake | Don't show "Notes:" on issues that don't have notes


shanear:
We are showing "Notes:" regardless of whether there are notes or not:

<img width="1035" alt="screen shot 2018-10-05 at 2 36 57 pm" src="https://user-images.githubusercontent.com/1596259/46560976-a5c30700-c8c3-11e8-90d6-33b6e9efe43f.png">

## Acceptance Criteria
- If there are no notes on an issue, don't show "Notes:" on the add issues page.

shanear:
<img width="1074" alt="screen shot 2018-10-09 at 5 18 49 pm" src="https://user-images.githubusercontent.com/1596259/46699313-69efb080-cbe7-11e8-954c-d41e99f5bd21.png">
works like a charm

PR #7307, opened 2018-10-05
do not display notes when there are no notes

jiahuang:
Connects #7306 

screen shot 2018-10-05 at 2 38 33 pm


leikkisa:
Haha, oops. Thanks Jia!

Issue #7320, opened 2018-10-09
Caseflow creates duplicate Decass records


evankroske:
There are a lot of duplicate Decass records in VACOLS that were created by Caseflow. We need to keep Caseflow from creating more and then delete the existing duplicates.

evankroske:
Duplicate of https://github.com/department-of-veterans-affairs/dsva-vacols/issues/43

PR #7321, opened 2018-10-09
QueueRepository | Prevent the creation of duplicate Decass records

evankroske:
- Part of https://github.com/department-of-veterans-affairs/caseflow/issues/7320
- See also https://github.com/department-of-veterans-affairs/caseflow/pull/7318

### Description
Confirms that the corresponding case is assigned to the current user before creating a new 
Decass record in `assign_case_to_attorney!`.

### Testing Plan
- [x] Circleci


app/repositories/queue_repository.rb
```diff
@@ -1,4 +1,6 @@
 class QueueRepository
+  class CaseAlreadyAssignedError < StandardError; end
```
aroltsch: @evankroske you should probably inherit from VacolsRepositoryError because we have logic to nicely catch it in the controller and return 400: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/application_controller.rb#L19

aroltsch: or you can fail with https://github.com/department-of-veterans-affairs/caseflow/blob/bbb7abb941459b35b537f88f32cc44b69fdad8c0/app/repositories/queue_repository.rb#L81
this error already inherits from VacolsRepositoryError: https://github.com/department-of-veterans-affairs/caseflow/blob/master/lib/caseflow/error.rb#L122

evankroske: Thanks! I changed the code to use QueueRepositoryError

Issue #7268, opened 2018-10-04
AMA Edit | Add Issue Landing


suzchap:
## Story
For any AMA forms, when a user needs to edit issues, they will click the "Edit in Caseflow" button in VBMS to return to the "Add issues landing page" where they will be able to add new issues and/or remove issues added previously. There are a few "are you sure?" type modals to help reduce errors. [Entire flow via Mural](https://app.mural.co/t/workqueue2001/m/workqueue2001/1537795938997/f5834747503f5a248340b9f3426fbac652812400)

This ticket is focused on the content of the landing page and functionality of the "Add issue" button.

## Technical notes

- We should consider that the intake (/add_issues) and intake edit (:id/edit) flows are similar; it may make sense to share some code.
- Unlike with intake screen (a la #6966), the edit page is not currently being used in prod. We don't have to support the old flow in prod.

## Acceptance criteria
- When you click "Edit in Caseflow" in VBMS, you go to the Add Issues Landing page (URL path should remain /{claim type}/{claim id}/edit where claim type is `higher_level_reviews` and `supplemental_claims`)
- The Add Issues Landing page should display the same fields as what displays on the Add Issues Landing page that's used in the original intake flow (as described in #6966) but should not have the progress bar at the top, the primary button is "save", and the cancel link is "cancel edit"  (see mock).
- The "+Add issue" button should behave the same as how it works in the original intake flow.
- The 'save' and 'cancel' buttons do not do anything (that comes in #7269)

## Notes
- Out of scope:
-- Page title is being changed from "Add Issues" to "Add/Remove Issues" for original and edit flows via #7266 

## Copy
n/a

## Mocks
<img width="625" alt="screen shot 2018-10-04 at 3 33 13 pm" src="https://user-images.githubusercontent.com/11219417/46498227-d4b97a00-c7ea-11e8-83c1-957ae71a3d5d.png">

## How to Validate
 
(no need, Shane's got this :muscle:)

kencheeto:
> For any AMA forms,

@shanear this means HLR and SC but not appeal, right? Since AMA appeals don't have end products?

shanear:
> @shanear this means HLR and SC but not appeal, right? Since AMA appeals don't have end products?

yes, appeals will come in a later story

steve-gov:
[...]

PR #7325, opened 2018-10-09
Edit landing page

leikkisa:
connects #7268


app/views/higher_level_reviews/edit.html.erb
```diff
@@ -4,9 +4,6 @@
     dropdownUrls: dropdown_urls,
     feedbackUrl: feedback_url,
     buildDate: build_date,
-    formType: "higher_level_review",
-    review: higher_level_review.ui_hash,
-    ratings: higher_level_review.cached_serialized_timely_ratings,
-    ratedRequestIssues: higher_level_review.request_issues.rated.map(&:ui_hash)
+    serverIntake: higher_level_review.ui_hash(FeatureToggle.enabled?(:intakeAma, user: current_user))
```
jiahuang: 👍 looks cleaner this way

client/app/intake/IntakeFrame.jsx
```diff
@@ -90,7 +89,7 @@ class IntakeFrame extends React.PureComponent {
                   exact
                   path={PAGE_PATHS.ADD_ISSUES}
                   title="Add Issues | Caseflow Intake"
-                  component={AddIssues} />
+                  component={IntakeAddIssuesPage} />
```
jiahuang: 👍 on the naming consistency

leikkisa:
edit-modals

leikkisa:
> Also note for the future -- we're testing the same functionality on the issues page a lot (3x for regular intake on hlr, appeal & sc, and now another 2x for editing hlr and sc). Eventually, we should break this out into js tests or something so it only gets repeated once. The ruby feature specs would then just test a simple path of adding one issue.

I think that's a good call! It was really useful for me in this case, testing helped me find things that were broken in intake vs. edit and HLR vs SC.  Happy to start converting to JS tests in the future.

PR #7330, opened 2018-10-10
Changing highlight colors

mdbenjam:
Resolves https://github.com/department-of-veterans-affairs/appeals-support/issues/3169

### Description
In our recent refacotring we changed the ordering of an override style so that it no longer overrode what we wanted it to 😬. This meant the opacity of the textlayer in Reader made highlighted words on search too faint for people to see. This PR fixes that and makes the colors darker so that they're more visible.
screen shot 2018-10-10 at 1 43 18 pm


client/app/styles/reader/_reviewer.scss
```diff
@@ -524,24 +524,6 @@ input[type="search"].cf-search-input-with-close {
...
 }
 
-// This is override some bad styling in PDFJS. Note selection is not
-// supported in current CSS standards. But with PDFJS using it, we have
-// to override it. When opacity isn't 1, then the highlighting for find
-// is way too light.
-.textLayer {
-  opacity: 1;
-
-  ::selection {
```
mdbenjam: This is no longer needed now that we don't use ctrl+f browser search.

client/app/styles/reader/_reviewer.scss
```diff
@@ -524,24 +524,6 @@ input[type="search"].cf-search-input-with-close {
   display: block;
 }
 
-// This is override some bad styling in PDFJS. Note selection is not
-// supported in current CSS standards. But with PDFJS using it, we have
-// to override it. When opacity isn't 1, then the highlighting for find
-// is way too light.
-.textLayer {
-  opacity: 1;
```
mdbenjam: This has been moved to inline styling so that it always has the highest priority and isn't overriden by the styling in pdfjs.

Issue #7178, opened 2018-09-27
AMA Decisions | Save request issue to resulting rating issue mapping


shanear:
When a rating is made (i.e. the decision is made) on a HLR or SC in VBMS-R, users will be forced to map each rating issue to a contention (which would be represented by a request issue generated at intake). We can query and use this mapping to store which rating issues were the result of which request issues (via #6545).

This will allow us to perform certain eligibility checks such as preventing previously appealed issues from being re-appealed, and preventing previously higher level reviewed issues from being higher level reviewed again (#5244).

requires #6545

## Acceptance Criteria
- When the EP for a HLR or SC is cleared, Caseflow should query which rating issues resulted from which contentions, and thus which request issues. That relationship of rating issues to contentions should be saved in the database.

## Notes
- There can be multiple rating issues mapped to a single request issue

## Technical Notes
This should happen in the exact same place where we are generating a follow up Supplemental Claim.

pkarman:
@shanear by

> This should happen in the exact same place where we are generating a follow up Supplemental Claim.

did you mean? https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/higher_level_review.rb#L72

Or some where else?

shanear:
> did you mean? https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/higher_level_review.rb#L72

> Or some where else?

Since it needs to happen for SC and HLR, probably a little further up the call stack. Like here: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/claim_review.rb#L71

shanear:
working! 

```
irb(main):018:0> sc.request_issues.map { |i| RatingIssue.find_by(request_issue: i) }
[2018-10-12 17:26:57 -0400]   RatingIssue Load (0.8ms)  SELECT  "rating_issues".* FROM "rating_issues" WHERE "rating_issues"."request_issue_id" = 317 LIMIT $1  [["LIMIT", 1]]
[2018-10-12 17:26:57 -0400]   RatingIssue Load (0.5ms)  SELECT  "rating_issues".* FROM "rating_issues" WHERE "rating_issues"."request_issue_id" = 316 LIMIT $1  [["LIMIT", 1]]
[...]

Issue #7299, opened 2018-10-05
Add issue edge cases | Show all ratings on add issue modal, not just timely ones


suzchap:
## Story
When a user adds issues, they may encounter edge cases where the issue is ineligible. We will not indicate that an issue isn't eligible on any of the add issue selection pages but will indicate it on the Add issue landing page and Confirmation Page so they have the information they need to inform the applicant in the letter they send following intake data entry. If an issue is ineligible in will get logged in Caseflow but not VBMS. [Mural showing edge cases](https://app.mural.co/t/workqueue2001/m/workqueue2001/1537791678087/94469b0cdf1a02a7f57a7628702d36db2f079430).

This ticket is focused on the scenario where the issue being requested may not be eligible for that lane due to 1 year eligibility date requirements -- and how we display that on the Add rated issues page. This can occur in HLR and Board Appeal lanes (not SC because there are no date restrictions) and only for rated issues. 

## Acceptance criteria
- On the add issues modal, show all prior rating issues regardless of timeliness and group by decision date just like other issues

## Mocks
(section titled 7/19/2017)
<img width="565" alt="screen shot 2018-10-05 at 2 20 53 pm" src="https://user-images.githubusercontent.com/11219417/46552819-e6158b80-c8a9-11e8-93f7-34e041abb71c.png">

## Background issues
- https://github.com/department-of-veterans-affairs/appeals-design-research/issues/915

## How to Validate
(no need, Shane's got this :muscle:)

shanear:
![screen shot 2018-10-31 at 5 35 03 pm](https://user-images.githubusercontent.com/1596259/47820256-5e4d6080-dd33-11e8-95f6-3856045acf3b.png)
Works!

PR #7348, opened 2018-10-10
Add RatingIssue table, link contentions

pkarman:
connects #7178

### Description

* Add `rating_issues` table, re-use existing model class
* Sync `Veteran` rating issues based on matching `contention_reference_id`
* make `contention_reference_id` a unique integer

### Acceptance Criteria
- [x] Code compiles correctly
- [x] Tests continue to pass

### Testing Plan
See the specific UAT examples in #7178

### DIFF
--- a/app/models/claim_review.rb
+++ b/app/models/claim_review.rb
@@ -71,6 +71,7 @@ class ClaimReview < AmaReview
   def on_sync(end_product_establishment)
     if end_product_establishment.status_cleared?
       sync_dispositions(end_product_establishment.reference_id)
+      sync_rating_issues(end_product_establishment.veteran_file_number)
       # allow higher level reviews to do additional logic on dta errors
       yield if block_given?
     end
@@ -93,6 +94,10 @@ class ClaimReview < AmaReview
         disposition: disposition.disposition
       )
     end
+  end
+
+  def sync_rating_issues(veteran_file_number)
+    Veteran.find_by!(file_number: veteran_file_number).sync_rating_issues!
   end
 
   def fetch_dispositions_from_vbms(reference_id)
--- a/app/models/claim_review.rb
+++ b/app/models/claim_review.rb
@@ -71,7 +71,7 @@ class ClaimReview < AmaReview
   def on_sync(end_product_establishment)
     if end_product_establishment.status_cleared?
       sync_dispositions(end_product_establishment.reference_id)
-      sync_rating_issues(end_product_establishment.veteran_file_number)
+      veteran.sync_rating_issues!
       # allow higher level reviews to do additional logic on dta errors
       yield if block_given?
     end
@@ -94,10 +94,6 @@ class ClaimReview < AmaReview
         disposition: disposition.disposition
       )
     end
-  end
-
-  def sync_rating_issues(veteran_file_number)
-    Veteran.find_by!(file_number: veteran_file_number).sync_rating_issues!
   end
 
   def fetch_dispositions_from_vbms(reference_id)
--- a/app/models/end_product_establishment.rb
+++ b/app/models/end_product_establishment.rb
@@ -220,7 +220,7 @@ class EndProductEstablishment < ApplicationRecord
 
   def rated_issue_contention_map(request_issues_to_associate)
     request_issues_to_associate.inject({}) do |contention_map, issue|
-      contention_map[issue.rating_issue_reference_id] = issue.contention_reference_id
+      contention_map[issue.rating_issue_reference_id] = issue.contention_reference_id.to_s
       contention_map
     end
   end
@@ -310,7 +310,7 @@ class EndProductEstablishment < ApplicationRecord
   end
 
   def contention_for_object(for_object)
-    contentions.find { |contention| contention.id == for_object.contention_reference_id }
+    contentions.find { |contention| contention.id == for_object.contention_reference_id.to_s }
   end
 
   # These are values that need to be determined based on the source right before the end
--- a/app/models/end_product_establishment.rb
+++ b/app/models/end_product_establishment.rb
@@ -220,7 +220,7 @@ class EndProductEstablishment < ApplicationRecord
 
   def rated_issue_contention_map(request_issues_to_associate)
     request_issues_to_associate.inject({}) do |contention_map, issue|
-      contention_map[issue.rating_issue_reference_id] = issue.contention_reference_id.to_s
+      contention_map[issue.rating_issue_reference_id] = issue.contention_reference_id
       contention_map
     end
   end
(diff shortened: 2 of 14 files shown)


app/models/veteran.rb
```diff
@@ -137,6 +137,19 @@ def participant_id
     super || ptcpnt_id
   end
 
+  def sync_rating_issues!
+    timely_ratings(from_date: Time.zone.today).each do |rating|
+      rating.issues.each do |rating_issue|
+        next unless rating_issue.contention_reference_id
```
shanear: If you prefer a more functional style, this can be captured on line `142` by doing something like

```
rating_issues.select(&:contention_reference_id).each do
  # ...
end
```

app/models/claim_review.rb
```diff
@@ -95,6 +96,10 @@ def sync_dispositions(reference_id)
     end
   end
 
+  def sync_rating_issues(veteran_file_number)
+    Veteran.find_by!(file_number: veteran_file_number).sync_rating_issues!
```
leikkisa: You should be able to access "veteran" directly since it's a method in ama_review.
```veteran.sync_rating_issues!```

File: app/models/claim_review.rb [hunk 3]
```diff
@@ -71,6 +71,7 @@ def invalid_modifiers
   def on_sync(end_product_establishment)
     if end_product_establishment.status_cleared?
       sync_dispositions(end_product_establishment.reference_id)
+      sync_rating_issues(end_product_establishment.veteran_file_number)
```
leikkisa: See comment below, it seems like you could put veteran.sync_rating_issues! here directly.

app/models/veteran.rb
leikkisa: I'm wondering if it's a bad idea for the from_date to be Time.zone.today in case there is a delay between when the award is promulgated, then the end_product cleared, and then when we succeed at syncing.  This searching ratings within a very narrow band of just today (from today, to today).

app/services/external_api/bgs_service.rb
```diff
@@ -257,13 +257,14 @@ def init_client
...
       application: "CASEFLOW",
-      client_ip: Rails.application.secrets.user_ip_address,
+      client_ip: ENV.fetch("USER_IP_ADDRESS", Rails.application.secrets.user_ip_address),
       client_station_id: current_user.station_id,
       client_username: current_user.css_id,
       ssl_cert_key_file: ENV["BGS_KEY_LOCATION"],
       ssl_cert_file: ENV["BGS_CERT_LOCATION"],
       ssl_ca_cert: ENV["BGS_CA_CERT_LOCATION"],
       forward_proxy_url: forward_proxy_url,
+      jumpbox_url: ENV["RUBY_BGS_JUMPBOX_URL"],
```
leikkisa: I'm curious what this is here for. If there's a forward_proxy_url I don't think the jumpbox_url does anything.

db/schema.rb
```diff
@@ -23,8 +23,8 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.boolean "granted"
-    t.index ["person_id"], name: "index_advance_on_docket_grants_on_person_id"
-    t.index ["user_id"], name: "index_advance_on_docket_grants_on_user_id"
+    t.index ["person_id"], name: "index_advance_on_docket_motions_on_person_id"
```
leikkisa: These seem like they might be from another PR?

app/models/veteran.rb
```diff
@@ -137,6 +137,19 @@ def participant_id
...
+        request_issue = RequestIssue.find_by(contention_reference_id: rating_issue.contention_reference_id)
+        if request_issue
+          rating_issue.request_issue = request_issue
+          rating_issue.save!
```
leikkisa: 👍

app/services/external_api/bgs_service.rb
pkarman: I was using `jumpbox_url` to tunnel through the UAT jumpbox in order to test from the rails console, in a similar way to https://github.com/department-of-veterans-affairs/appeals-design-research/blob/master/Projects/Intake/onboarding/testing-bgs.md

db/schema.rb
pkarman: Yes the table name changed in a previous migration but was not fixed in the schema. When I ran `rake db:migrate` locally it re-created the indexes based on the new table name.

pkarman:
@shanear @leikkisa you both made comments about `from_date` being `today...today`

I read the code otherwise: https://github.com/department-of-veterans-affairs/caseflow/pull/7348#discussion_r224493911

If I am reading the code correctly, it seems like `from_date` is not a great param name since you both misinterpreted it. Maybe we should rename it `end_date`?

app/models/rating_issue.rb
```diff
@@ -1,7 +1,14 @@
...
+class RatingIssue < ApplicationRecord
+  belongs_to :request_issue
 
-  attr_accessor :reference_id, :decision_text
+  # this local attribute is used to resolve the related RequestIssue
+  attr_accessor :contention_reference_id
+
+  def save_with_request_issue!
+    return unless related_request_issue
+    self.request_issue = related_request_issue
```
pkarman: we use `save` rather than `update` because Rating uses `new` rather than `create`. I.e. the object may not have been previously `insert`ed.

shanear:
> If I am reading the code correctly, it seems like from_date is not a great param name since you both misinterpreted it. Maybe we should rename it end_date?

I agree, but timely_ratings will be going away soon anyways, so I'm not sure we need to fix it: https://github.com/department-of-veterans-affairs/caseflow/issues/7299

pkarman:
@shanear @leikkisa I removed the WIP on this one -- care to give it an official review?

app/models/end_product_establishment.rb
```diff
@@ -310,7 +310,7 @@ def create_contentions_in_vbms(descriptions)
   end
 
   def contention_for_object(for_object)
-    contentions.find { |contention| contention.id == for_object.contention_reference_id }
+    contentions.find { |contention| contention.id.to_i == for_object.contention_reference_id.to_i }
```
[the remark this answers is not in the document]
pkarman: no, unfortunately. all the SOAP responses are parsed as strings, so `contention.id` is a string.

spec/models/end_product_establishment_spec.rb
```diff
@@ -370,19 +370,20 @@
     end
 
     let(:reference_id) { "stevenasmith" }
+    let(:contention_ref_id) { SecureRandom.random_number(1_000_000) }
```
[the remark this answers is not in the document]
pkarman: I was trying to mock what the fakes lib does, but I see your point in this particular test.

Issue #6990, opened 2018-09-14
AMA Intake | Create "Unidentified Issue" modal and add entry to Add Issues Summary page


suzchap:
## Story
For any AMA forms, when a user needs to add issues, they will be walked through a multi-screen selection process to help them make the best choice. The landing page created in #6966 will be their starting point and a modal on top of it will walk the user through rated, non-rated, and an "unknown" option.

When a user doesn't find a match for an issue written on the form with a non rated issue displayed on #6981 OR on the non-rated (#6988), they can click the "None of these match, see more options" button to advance to a new modal that displays "Unknown issue" options where they can transcribe the issue from the form and add optional notes. This content then shows on the Add Issue Summary page. 

requires #6988 

## Technical notes
- we will need to figure out the DB structure for this

## Acceptance criteria
- From the Add issue modal for non-rated issues, clicking the "None of these match, see more options" button moves to show another modal with a field to transcribe the issue as it's written on the form and an optional Notes field.
- "Add this issue" button is disabled until all the required fields are filled out.
- "Add this issue" button, when clicked, will close the modal, and the Add Issues summary page will now display the transcription and note (if there is one) in the "Requested issues" section.'' (it will be in harsh red)

## Notes
- don't worry about the contention text if you don't want to (it will be handled https://github.com/department-of-veterans-affairs/caseflow/issues/7010)
- Out of scope:
   - Dynamic issue ("Add issue 3") and numbering doesn't need to work yet, it'll be done in another issue.
- this will also be for edit.

## Copy
- Add issue modal
   - Describe the issue to mark it as needing further review.
   - Transcribe the issue as it's written on the form
   - Notes
   - Optional
- Add issue summary page
   - Unidentified issue: no issue matched for "[TRANSCRIPTION FIELD]"

## Mocks
**Add issue modal for unknown issues**
<img width="628" alt="screen shot 2018-09-14 at 1 06 05 pm" src="https://user-images.githubusercontent.com/11219417/45564651-1cb63400-b81f-11e8-821e-ec3b2a2b62f0.png">

**Add issues summary page**
<img width="716" alt="screen shot 2018-10-05 at 9 25 49 am" src="https://user-images.githubusercontent.com/11219417/46537981-b3a36880-c880-11e8-8ac4-a5bc51431230.png">

## How to Validate
- Doesn't require a specific test user
[...]

PR #7349, opened 2018-10-11
add flag for unidentified issues

jiahuang:
Backend portion of #6990

### Description
- Adds an `is_unidentified` column to request issues. Unidentified issues have this column set to true
- Removes `Unknown issue category` from being saved in the issue category column for unidentified issues
- Changes unidentified issues from being on a nonrated EP to a rated EP

### Todo
- [ ] verify that we do not need a further migration to migrate issues with "unknown issue category" to use the boolean flag

### DIFF
--- a/app/models/request_issue.rb
+++ b/app/models/request_issue.rb
@@ -10,8 +10,8 @@ class RequestIssue < ApplicationRecord
   end
 
   def self.nonrated
-    where(rating_issue_reference_id: nil, rating_issue_profile_date: nil)
-      .where.not(issue_category: nil, is_unidentified: true)
+    where(rating_issue_reference_id: nil, rating_issue_profile_date: nil, is_unidentified: [nil, false])
+      .where.not(issue_category: nil)
   end
 
   def self.unidentified
@@ -24,10 +24,6 @@ class RequestIssue < ApplicationRecord
 
   def rated?
     rating_issue_reference_id && rating_issue_profile_date
-  end
-
-  def unidentified?
-    is_unidentified
   end
 
   def self.from_intake_data(data)
--- a/app/models/request_issue.rb
+++ b/app/models/request_issue.rb
@@ -6,12 +6,16 @@ class RequestIssue < ApplicationRecord
 
   def self.rated
     where.not(rating_issue_reference_id: nil, rating_issue_profile_date: nil)
-      .or(where(issue_category: "Unknown issue category"))
+      .or(where(is_unidentified: true))
   end
 
   def self.nonrated
     where(rating_issue_reference_id: nil, rating_issue_profile_date: nil)
-      .where.not(issue_category: [nil, "Unknown issue category"])
+      .where.not(issue_category: nil, is_unidentified: true)
+  end
+
+  def self.unidentified
+    where(rating_issue_reference_id: nil, rating_issue_profile_date: nil, is_unidentified: true)
   end
 
   def self.no_follow_up_issues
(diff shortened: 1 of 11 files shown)


client/app/intakeCommon/util/index.js
```diff
@@ -57,11 +57,22 @@ export const validNonRatedIssue = (issue) => {
 };
 
 const formatRatedIssues = (state) => {
+  // unidentified issues are counted as rated issues
   if (state.addedIssues && state.addedIssues.length > 0) {
     // we're using the new add issues page
     return state.addedIssues.
-      filter((issue) => issue.isRated).
+      filter((issue) => issue.isRated || issue.isUnidentified).
```
jiahuang: not super sold on the way I'm combining rated & undefined here, would be fine breaking this out into a separate "formatUnidentifiedIssues" function

client/app/intakeCommon/constants.js
```diff
@@ -56,10 +55,8 @@ export const ISSUE_CATEGORIES = issueCategoriesArray.map((category) => {
   };
 });
 
-// Removes the "Unknown issue category"
-// which is temporary until we activate the new "Add issues" flow
 export const NON_RATED_ISSUE_CATEGORIES = issueCategoriesArray.
```
leikkisa: This is a duplicate of the ISSUE_CATEGORIES constant above, so we could just get rid of this if you are getting rid of "Unknown issue category".  However, "Unknown issue category" is still present in the "Select issues" flow that is the live flow right now.  It sounds like it hasn't been used yet, so we could ask about getting rid of it, and then we could get rid of this constant and use the one above.

client/app/intakeCommon/util/index.js
leikkisa: I would be okay with that, seems like it would be a little easier to follow.  This is okay too.

db/schema.rb
```diff
@@ -23,8 +23,8 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.boolean "granted"
-    t.index ["person_id"], name: "index_advance_on_docket_grants_on_person_id"
-    t.index ["user_id"], name: "index_advance_on_docket_grants_on_user_id"
+    t.index ["person_id"], name: "index_advance_on_docket_motions_on_person_id"
```
leikkisa: These were on one of Peter's PRs too, seems like it belongs somewhere else. DB migration weirdness.

client/app/intakeCommon/util/index.js
shanear: I agree with this. It'd be nice to format them in separate methods

spec/feature/intake/supplemental_claim_spec.rb
```diff
@@ -532,6 +532,13 @@ def check_row(label, text)
                decision_date: 1.month.ago.to_date,
                end_product_establishment_id: non_rating_end_product_establishment.id
       )).to_not be_nil
+
+      expect(RequestIssue.find_by(
+               review_request: supplemental_claim,
+               description: "This is an unidentified issue",
+               is_unidentified: true,
+               end_product_establishment_id: end_product_establishment.id
+      )).to_not be_nil
```
shanear: Nice simple tests 💯

app/models/request_issue.rb
```diff
@@ -6,12 +6,16 @@ class RequestIssue < ApplicationRecord
...
   end
 
   def self.nonrated
     where(rating_issue_reference_id: nil, rating_issue_profile_date: nil)
-      .where.not(issue_category: [nil, "Unknown issue category"])
+      .where.not(issue_category: nil, is_unidentified: true)
+  end
+
+  def self.unidentified
+    where(rating_issue_reference_id: nil, rating_issue_profile_date: nil, is_unidentified: true)
```
shanear: Do these scopes have unit tests? I think it'd probably nice to have tests for them.

Issue #7312, opened 2018-10-09
Fix Cancel Intake on Add Issues page


leikkisa:
## Story
When on the Add Issues page, canceling the intake seems to redirect to /intake/add_issues instead of to /intake and does not render a page

## Acceptance criteria
Cancel intake from the add_issues page cancels the intake and redirects to /intake

PR #7353, opened 2018-10-11
canceling intake from add_issues should redirect

kencheeto:
Connects #7312 

### Description
The current Finish page will redirect back to intake start (`/intake`) if the formType is undefined, which is cleared when canceling an intake. The new AddIssues page should be consistent with that behavior.

I'm not really sold on using a falsy `formType` to redirect since this is navigation by side effect, but this is simple for now. I think once we switch over to the new AddIssues page by default we can edit the `submitCancel` action to do the redirection.

demo

### Acceptance Criteria
- [x] Canceling any intake from the AddIssues page will redirect back to `/intake` as the current Finish page does

### Testing Plan
1. Start any intake, get to the review screen
2. Visit /intake/add_issues
3. Cancel the intake and verify that you are redirected back to /intake (currently there is a blank screen)

### DIFF
--- a/client/app/intake/pages/addIssues.jsx
+++ b/client/app/intake/pages/addIssues.jsx
@@ -1,6 +1,7 @@
 import _ from 'lodash';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
+import { Redirect } from 'react-router-dom';
 import React from 'react';
 
 import AddIssuesModal from '../components/AddIssuesModal';
@@ -9,6 +10,7 @@ import UnidentifiedIssuesModal from '../components/UnidentifiedIssuesModal';
 import Button from '../../components/Button';
 import RequestIssuesUpdateErrorAlert from '../../intakeEdit/components/RequestIssuesUpdateErrorAlert';
 import { FORM_TYPES } from '../constants';
+import { PAGE_PATHS } from '../constants';
 import { formatDate } from '../../util/DateUtil';
 import { formatAddedIssues, getAddIssuesFields } from '../util/issues';
 import Table from '../../components/Table';
@@ -19,7 +21,7 @@ import {
   toggleUnidentifiedIssuesModal
 } from '../actions/addIssues';
 
-class AddIssuesPage extends React.PureComponent {
+export class AddIssues extends React.PureComponent {
   render() {
     const {
       intakeForms,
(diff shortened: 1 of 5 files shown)


client/app/intake/pages/addIssues.jsx
```diff
@@ -1,28 +1,34 @@
...
+import { PAGE_PATHS } from '../constants';
 import { formatDate } from '../../util/DateUtil';
 import { formatAddedIssues, getAddIssuesFields } from '../util';
 
 import Table from '../../components/Table';
 import { toggleAddIssuesModal, toggleNonRatedIssueModal, toggleUnidentifiedIssuesModal } from '../actions/common';
 import { removeIssue } from '../actions/ama';
 
-class AddIssues extends React.PureComponent {
+export class AddIssues extends React.PureComponent {
```
kencheeto: exporting here so we can unit test the undecorated component

client/test/karma/intake/addIssues-test.js
```diff
@@ -0,0 +1,13 @@
...
+import { expect } from 'chai';
+import { shallow } from 'enzyme';
+import { AddIssues } from '../../../app/intake/pages/addIssues';
+import { Redirect } from 'react-router-dom';
+
+describe('AddIssues', () => {
+  it('renders Redirect when formType is undefined', () => {
+    const wrapper = shallow(<AddIssues />);
+
+    expect(wrapper.find(Redirect)).to.have.lengthOf(1);
```
pkarman: I confess I don't know enough React to make this sound intelligent, but the `lengthOf` test surprised me. We don't care what the redirect value is? Just that it was created?

kencheeto: ah, i guess this test could be less minimal, esp. with future devs in mind! fixed

client/test/karma/intake/addIssues-test.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
```
leikkisa: Whoa, a Karma test!

kencheeto: yeah! I added a karma test instead of a feature test because i thought a feature test would be slow. canceling isn't cheap to tack onto an existing feature test because an intake has to get to the /add_issues page and canceling clears the intake, plus there are multiple form types and a feature test would have to test each kind of form in order to get the same coverage as this karma test.

What do you think? happy to go either way, I admit I was also keen on trying out a js test

shanear: I think feature tests are a little bit more manageable, since we already use the rspec DSL in unit tests. I also like that they incorporate more of the stack. I don't think the speed is too much of an issue since we can just crank up the parallelism to compensate for it.

pkarman:
I got bit by this again this morning -- +1 to merge while its all 💚

PR #7358, opened 2018-10-11
Adding instruction box for generic tasks

mdbenjam:
### Description
Users should be able to add instructions when assigning a task to a user or team. This PR adds an instructions text box on the assign to user/team modal which allows users to specify the instructions. These instructions will be autofilled when assigning to a user within the organization since you likely want the same instructions that the parent task had. These instructions will appear in the task information section in case snapshot.

### Testing Plan
1. Assign an organization task to a user.


client/app/queue/AssignToView.jsx
```diff
@@ -35,20 +37,27 @@ type Props = Params & {|
...
 
+    // Autofill the instruction field if assigning to a person on the team. Since they will
+    // probably want the instructions from the assigner.
+    const instructions = this.props.task.instructions;
+    const existingInstructions = instructions.length > 0 ? instructions[instructions.length - 1] : '';
+
     this.state = {
-      selectedValue: null
+      selectedValue: null,
+      instructions: !this.props.isTeamAssign ? existingInstructions : ''
```
lowellrex: Can we combine these two conditional statements into one? Perhaps something like:
```
let existingInstructions = '';
if (instructions.length > 0 && !this.props.isTeamAssign) {
  existingInstructions = instructions[instructions.length - 1];
}
```

client/app/queue/AssignToView.jsx
```diff
@@ -35,20 +37,27 @@ type Props = Params & {|
...
     };
   }
 
   validateForm = () => {
-    return this.state.selectedValue !== null;
+    return this.state.selectedValue !== null && this.state.instructions !== '';
```
lowellrex: This enforces that all tasks are required to have instructions now, correct?

mdbenjam: Correct

PR #7361, opened 2018-10-11
Work around some circular dependencies in create/seed dbs

pkarman:
### Description
`bundle exec rake local:build` was failing because the local psql db was not yet created. There was also a flaky timing issue with Oracle startup.

### DIFF
--- a/lib/tasks/local/build.rake
+++ b/lib/tasks/local/build.rake
@@ -16,6 +16,9 @@ namespace :local do
     # Add a new line so that this scipt's output is more readable.
     puts ""
 
+    puts "Creating local caseflow dbs"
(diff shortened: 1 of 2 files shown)


lib/tasks/local/build.rake
```diff
@@ -16,6 +16,9 @@ namespace :local do
     # Add a new line so that this scipt's output is more readable.
     puts ""
 
+    puts "Creating local caseflow dbs"
+    system("bundle exec rake db:create db:schema:load") || abort
```
pkarman: `db:setup` does these 2 plus `db:seed` but we intentionally do not want `seed` yet until the vacols dbs are ready.

PR #7365, opened 2018-10-11
Return all tasks and change indexing on FE tasks

mdbenjam:
Connects #7102

### Description
We split this out from https://github.com/department-of-veterans-affairs/caseflow/pull/7214.

This PR fixes a bug where we only return a subset of tasks from the task_controller's for_appeal method. However, once we start returning all tasks we need to change how we index them. This moves us to use the externalAppealId to index legacy tasks and the task id to index modern tasks.

### Testing Plan
1. Test Co-located flow for AMA and legacy tasks
1. Test attorney to judge flow for AMA and legacy tasks


app/models/serializers/work_queue/legacy_task_serializer.rb
```diff
@@ -1,4 +1,7 @@
 class WorkQueue::LegacyTaskSerializer < ActiveModel::Serializer
+  attribute :is_legacy do
```
mdbenjam: We had whether or not an appeal was legacy, but not if the task is legacy.

client/app/queue/AddColocatedTaskView.jsx
```diff
@@ -94,11 +94,12 @@ class AddColocatedTaskView extends React.PureComponent<Props, ComponentState> {
     };
 
     this.props.requestSave('/tasks', payload, successMsg).
-      then((resp) => {
-        const response = JSON.parse(resp.text);
```
mdbenjam: I do not think this is right. In the case of a legacy task, the response has the newly created ama task, not an updated version of the legacy task. In the case of an ama task, the response has the newly created ama task, not the original ama task.

client/app/queue/CaseSnapshot.jsx
```diff
@@ -229,6 +230,16 @@ export class CaseSnapshot extends React.PureComponent<Props> {
     </React.Fragment>;
   };
 
+  showActionsForLegacyTasks = (legacyTasks: Array<Task>): boolean => {
+    // users can end up at case details for appeals with no DAS
+    // record (!task.taskId). prevent starting checkout flows
+    return Boolean(legacyTasks.length && _.every(legacyTasks, (task) => task.taskId));
+  }
+
+  showActionsForTaskAssignedToUser = (tasks: Array<Task>): boolean => {
```
mdbenjam: Different criteria need to be used to see if a task or legacytask are actionable.

client/app/queue/QueueActions.js
```diff
@@ -345,9 +353,15 @@ export const initialAssignTasksToUser = ({
         tasks: allTasks.tasks,
         amaTasks: allTasks.amaTasks
       }));
+      if (!oldTask.isLegacy) {
```
mdbenjam: The response from above is not an updated version of the existing task, but the newly created task. AMA tasks need to be marked as on hold.

client/app/queue/ReaderLink.jsx
```diff
@@ -24,11 +24,10 @@ export default class ReaderLink extends React.PureComponent {
       appeal,
       longMessage
     } = this.props;
-    const externalId = appeal.externalId || appeal.attributes.external_id;
```
mdbenjam: This was only needed before our previous appeal refactor back in August. Now we can remove it.

client/app/queue/reducers.js
```diff
@@ -335,30 +342,22 @@ const workQueueReducer = (state = initialState, action = {}): QueueState => {
         }
       }
     });
-  case ACTIONS.SET_TASK_ASSIGNMENT: {
```
mdbenjam: This reducer was never used.

client/app/queue/utils.js
```diff
@@ -154,15 +158,15 @@ export const prepareLegacyTasksForStore = (tasks: Array<Object>): Tasks => {
...
+  return _.pickBy(_.keyBy(mappedLegacyTasks, (task) => task.uniqueId), (task) => task);
 };
 
 export const prepareAllTasksForStore = (tasks: Array<Object>): { amaTasks: Tasks, tasks: Tasks } => {
   const amaTasks = tasks.filter((task) => {
-    return task.attributes.appeal_type === 'Appeal';
+    return !task.attributes.is_legacy;
   });
   const legacyTasks = tasks.filter((task) => {
-    return task.attributes.appeal_type === 'LegacyAppeal';
```
mdbenjam: This is not correct. LegacyAppeals can have AMA tasks.

client/app/queue/CaseSnapshot.jsx
```diff
@@ -229,6 +230,16 @@ export class CaseSnapshot extends React.PureComponent<Props> {
...
   };
 
+  showActionsForLegacyTasks = (legacyTasks: Array<Task>): boolean => {
+    // users can end up at case details for appeals with no DAS
+    // record (!task.taskId). prevent starting checkout flows
+    return Boolean(legacyTasks.length && _.every(legacyTasks, (task) => task.taskId));
+  }
+
+  showActionsForTasksAssignedToUser = (tasks: Array<Task>): boolean => {
+    return _.every(tasks, (task) => task.status !== TASK_STATUSES.on_hold);
```
evankroske: I don't really understand this code. I thought that attorneys could take action on on-hold tasks, and I'm not sure what this code would do with completed tasks.

client/app/queue/MarkTaskCompleteView.jsx
```diff
@@ -76,7 +77,8 @@ const mapStateToProps = (state: State, ownProps: Params) => {
 
   return {
     error,
-    task: getTasksForAppeal(state, ownProps)[0],
+    task: tasksForAppealAssignedToUserSelector(state, ownProps)[0] ||
+      organizationTasksByAssigneeIdSelector(state, ownProps)[0],
```
evankroske: Does this change belong in this PR?

client/app/queue/reducers.js
```diff
@@ -85,13 +85,20 @@ const workQueueReducer = (state = initialState, action = {}): QueueState => {
         $set: action.payload.judges
       }
     });
-  case ACTIONS.DELETE_APPEAL:
+  case ACTIONS.DELETE_APPEAL: {
+    const amaTasksIds = _.map(
+      _.filter(
+        state.amaTasks, (task) => task.externalAppealId === action.payload.appealId
+      ), 'uniqueId'
```
evankroske: FYI, using the string form of _.map isn't type-safe, so Flow wouldn't warn you if you had misspelled "uniqueId". I think that we should try to use the function form of lodash functions

client/app/queue/MarkTaskCompleteView.jsx
mdbenjam: Yes. Choosing the 0th entry on `getTasksForAppeal` no longer is correct. It might be a task that is not assigned to you. Instead we should pick either the first task assigned to you, or the first task assigned to an organization.

client/app/queue/CaseSnapshot.jsx
mdbenjam: Fair, I can't remember the exact reason I changed it. Seems to work fine for now without it.

Issue #7323, opened 2018-10-09
Fix undefined method bug in Rating model


pkarman:
BGS may return a nil value instead of an empty list, in https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/rating.rb#L72 so we need a little defensive check there. Not sure whether it means we should raise an exception or return an empty array.

ref https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2531/

leikkisa:
We should raise an exception, returning an empty array would erroneously make it seem like there are no timely ratings for the Veteran.

shanear:
I think this was due to BGS returning the locked rating error. (it's still a 200 though)

shanear:
![screen shot 2018-10-12 at 4 32 33 pm](https://user-images.githubusercontent.com/1596259/46892707-728cf480-ce3c-11e8-9d04-7f357747ed76.png)
This looks like it's still happening in UAT. 

https://sentry.uat.ds.va.gov/sentry/caseflow-uat/issues/984/

shanear:
Aaaaand, this is because the code hasn't deployed to UAT, because jenkins is broken. hanging on to this one to re-test.

shanear:
![screen shot 2018-10-12 at 5 12 45 pm](https://user-images.githubusercontent.com/1596259/46894316-11681f80-ce42-11e8-9fe8-c43166aa6ab1.png)
super

PR #7367, opened 2018-10-11
If BGS does not return rating_profile_list structure, raise exception

pkarman:
connects #7323 

### Description
When a Rating is locked, we fail with Ruby exception about no method for `[]`. Catch that case and throw an error of our own instead.

### Acceptance Criteria
- [x] Code compiles correctly
- [x] Tests continue to pass

### Testing Plan
Wait for a locked rating?

### DIFF
--- a/spec/models/rating_spec.rb
+++ b/spec/models/rating_spec.rb
@@ -127,5 +127,14 @@ describe Rating do
         )
       end
     end
+
+    context "when a rating is locked" do
+      it "throws NilRatingProfileListError" do
+        allow_any_instance_of(Fakes::BGSService).to receive(:fetch_ratings_in_range).and_return({})
+        expect do
+          Rating.fetch_timely(participant_id: "DRAYMOND", from_date: receipt_date)
+        end.to raise_error(Rating::NilRatingProfileListError)
+      end
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


spec/models/rating_spec.rb
```diff
@@ -127,5 +127,14 @@
         )
       end
     end
+
+    context "when a rating is locked" do
+      it "throws NilRatingProfileListError" do
+        allow_any_instance_of(Fakes::BGSService).to receive(:fetch_ratings_in_range).and_return({})
```
pkarman: would be nice to mock the actual SOAP response instead of just an empty hash.

shanear: i can get this for you

leikkisa:
I think it'd be really nice if we could distinguish between Locked and Converted or Backfilled, or maybe include the intake ID in the error message

I have been assuming this error is for Locked rating, but I think the one from today was Converted or Backfilled.
{:error_level=>"0", :any_ratings_within_range_flag=>"N", :differences_found=>"N", :earliest_dependency_eligibility_date=>Sun, 01 Feb 1987 00:00:00 -0600, :multiple_ratings_flag=>"N", :ptcpnt_vet_id=>"1234", :reject_reason=>"Converted or Backfilled Rating - no promulgated ratings found", :rating_profile_list=>nil}

pkarman:
@leikkisa would e7799d1a2 do the trick?

PR #7368, opened 2018-10-11
Dockerizing Caseflow App for demo and development

enriquemanuel:

### Description

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...


pkarman:
@enriquemanuel I like where you're headed with this.

Might this help? https://store.docker.com/images/oracle-instant-client?tab=description

enriquemanuel:
The circleci issue might be due to the changes that I did to the yml files so connectivity might be not good since this now depends on the `docker-compose` file

docker-compose.yml
```diff
@@ -6,20 +6,21 @@ services:
     image: vacols_db
     tty: true
     stdin_open: true
-    command: bash -c "/home/oracle/setup/dockerInit.sh"
+    command: bash -c "/home/oracle/setup/dockerInitWithSchema.sh"
```
mdbenjam: Do we need to change this? I don't want devs to have to wait for a schema load every time they start up the container. If they've already done it in the past the result should already be in the volumes and we don't need to init with schema again.

docker-compose.yml
```diff
@@ -51,11 +55,21 @@ services:
     volumes:
       - ./local/sqs/conf:/conf
 
+  appeals-app:
```
mdbenjam: So if folks use `docker-compose up` locally they'll try to start up caseflow in a container? I'm not sure that's desirable.

enriquemanuel: Yes, this will bring the file, what I can do, is to make a different compose file used for our use case and leave intact this. Do you think that would be a good approach?

docker-compose.yml
enriquemanuel: By doing the other docker-compose file mentioned below, will resolve this. Ideally, we want the demo to be idempotent so that is why I was doing it.

docker-compose.yml
mdbenjam: Yup I think that's the move. Keep this docker-compose file for developers and have a slightly different one for demo.

sunil-sadasivan: Agreed, I think it's preferable for devs to run Caseflow on the host machine and have dependencies dockerized. 

@enriquemanuel @mdbenjam perhaps we need a separate docker-compose file for demo that includes caseflow in a container?

enriquemanuel:
I have created the separate compose file and updated the documentation above.
Also have reverted the changes back to the `docker-compose.yml`

docker-compose.yml
enriquemanuel: just corrected this

docker-compose.yml
enriquemanuel: reverted to default and created compose file for demo

docker-compose-demo.yml
```diff
@@ -0,0 +1,59 @@
...
+    ports:
+      - "4567-4583:4567-4583"
+      - "8082:${PORT_WEB_UI-8080}"
+    environment:
+      - SERVICES=sqs
+    volumes:
+      - ./local/sqs/conf:/conf
+
+  appeals-app:
+    image: caseflow:latest
```
RobertAlexanderUSDS: Would it make sense to parameterize which image is used?

Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+    curl -sL https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
+    echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \ 
+    apt -y update && \
+    curl -sL https://deb.nodesource.com/setup_6.x | bash - && \
+    apt install -y ${CASEFLOW} &&  \
+    curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
+    echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+
+# Copy all the files
+COPY . .
```
RobertAlexanderUSDS: Can we check what files this ends up including? We probably don't want the .git/ folder to get copied over. Are there any sensitive files you may have locally, like hidden files?

docker-compose-demo.yml
enriquemanuel: now cause i'm doing the tagging in another script upon deployment

Dockerfile
enriquemanuel: i can double check and add them to the dockerignore file

enriquemanuel: I have created a file called `.dockerignore` and it includes several things that were in the `.gitignore` that has lowered the docker image almost 400Mb

docker-compose-demo.yml
enriquemanuel: this `docker-compose` file is a template to be used in ECS, so we always need to use `latest` unless we specify it but again its just a template.

lisac-usds:
Whoa!! This looks really cool. Running through the build/startup this morning, will report back.

lisac-usds:
Ran the `Build Caseflow Docker Image` steps locally, verified `git status` showed no artifacts that needed to be committed. 

@enriquemanuel some comments re: the s3 and ecr usage:
1. Is it true, `aws` credentials need to be available as shell variables in order to run these commands? If so: could consider reminding the user of this, maybe via a comment in the script, as part of an error msg, or in a `docker-bin/README.md`?
1. Are these s3 buckets accessible only to the Appeals Modernization team? If so: could consider flagging this (I noticed this repo is public, so hard to pin down who might try to run the script)

Still working through the `Start it up` steps (getting a strange error, probably just my local setup - a "pull access denied for store/oracle/database-enterprise...", will investigate this afternoon).

aroltsch:
Just an FYI, you pass `-f` option before `up`:
```
docker-compose -f docker-compose.yml up
```

lisac-usds:
Oops. To explain the issue I noted at the end of my previous message - I hadn’t signed up for `store/oracle/database-enterprise`, which led to my problem with running the vacols [Dockerfile](https://github.com/department-of-veterans-affairs/caseflow/blob/1c2878c0db0b3e8f3aa69b0af1435ce9789b4066/local/vacols/Dockerfile#L1)  🤦‍♀️  Solved by signing up for [Oracle Database Enterprise Edition](https://store.docker.com/images/oracle-database-enterprise-edition).

```
$ docker ps -a
CONTAINER ID        IMAGE                   COMMAND                  CREATED             STATUS                    PORTS                                                      NAMES
f6f4e7e15d69        caseflow:latest         "/bin/bash /caseflow…"   31 minutes ago      Up 31 minutes             0.0.0.0:3000->3000/tcp                                     appeals-app
6e1ee73fa0e2        redis:2.8.23            "docker-entrypoint.s…"   31 minutes ago      Up 31 minutes             0.0.0.0:6379->6379/tcp                                     appeals-redis
4531dda26a07        postgres:9.3.14         "/docker-entrypoint.…"   31 minutes ago      Up 31 minutes             0.0.0.0:5432->5432/tcp                                     appeals-db
f487aeafaab5        vacols_db               "bash -c /home/oracl…"   31 minutes ago      Up 31 minutes (healthy)   5500/tcp, 0.0.0.0:1522->1521/tcp                           VACOLS_DB-development
9d55d3a3643d        localstack/localstack   "/usr/bin/supervisor…"   31 minutes ago      Up 31 minutes             0.0.0.0:4567-4583->4567-4583/tcp, 0.0.0.0:8082->8080/tcp   localstack

```


You figured out how to dockerize Caseflow, @enriquemanuel! 

Looks good to me. I left some [comments](https://github.com/department-of-veterans-affairs/caseflow/pull/7368#issuecomment-435427479) about the aws/ecr access.

Issue #7301, opened 2018-10-05
Add issue edge cases | if untimely | untimely exemption


suzchap:

suzchap:
requires #7300
requires #7299

shanear:
![screen shot 2018-11-06 at 5 43 34 pm](https://user-images.githubusercontent.com/1596259/48098741-8e8d7700-e1eb-11e8-849a-7dfefb8c06f3.png)
I think the AC was misleading but this should also work for non-rating issues as well. The modal did not show up for the third issue i added here.

PR #7377, opened 2018-10-12
Add Rating.fetch_all method to get all ratings regardless of date

pkarman:
connects #7299 

### Description
Adds new method `Rating.fetch_all` and uses that instead of `fetch_timely` within the `Veteran` model.

### Acceptance Criteria
- [x] Code compiles correctly
- [x] Tests continue to pass

### Testing Plan
1. Identify rating issue older than one year.
2. Verify it appears for the related veteran.

### DIFF
--- a/app/models/rating.rb
+++ b/app/models/rating.rb
@@ -7,8 +7,8 @@ class Rating
   # It is a datetime, not a date.
   attr_accessor :participant_id, :profile_date, :promulgation_date
 
-  # One week buffer was added
-  TIMELY_DAYS = 372.days
+  ONE_YEAR_PLUS_DAYS = 372.days
+  TWO_LIFETIMES_DAYS = 250.years
 
   def issues
     @issues ||= fetch_issues
@@ -45,15 +45,20 @@ class Rating
   end
 
   class << self
+    def fetch_all(participant_id)
+      fetch_timely(participant_id: participant_id, from_date: (Time.zone.today - TWO_LIFETIMES_DAYS))
+    end
+
     def fetch_timely(participant_id:, from_date:)
+      start_date = from_date - ONE_YEAR_PLUS_DAYS
       response = BGSService.new.fetch_ratings_in_range(
         participant_id: participant_id,
-        start_date: from_date - TIMELY_DAYS,
+        start_date: start_date,
         end_date: Time.zone.today
       )
 
       unsorted = ratings_from_bgs_response(response).select do |rating|
-        rating.promulgation_date > (from_date - TIMELY_DAYS)
+        rating.promulgation_date > start_date
       end
 
       unsorted.sort_by(&:promulgation_date).reverse
--- a/app/models/veteran.rb
+++ b/app/models/veteran.rb
@@ -116,7 +116,7 @@ class Veteran < ApplicationRecord
   end
 
   def timely_ratings
-    @timely_ratings ||= Rating.fetch_all(participant_id: participant_id)
+    @timely_ratings ||= Rating.fetch_all(participant_id)
   end
 
   def accessible_appeals_for_poa(poa_participant_ids)
--- a/app/models/veteran.rb
+++ b/app/models/veteran.rb
@@ -115,8 +115,12 @@ class Veteran < ApplicationRecord
     @zip_code || (@address_line3 if @address_line3 =~ /(?i)^[a-z0-9][a-z0-9\- ]{0,10}[a-z0-9]$/)
   end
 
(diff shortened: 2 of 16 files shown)


app/models/veteran.rb
```diff
@@ -115,8 +115,8 @@ def zip_code
     @zip_code || (@address_line3 if @address_line3 =~ /(?i)^[a-z0-9][a-z0-9\- ]{0,10}[a-z0-9]$/)
   end
 
-  def timely_ratings(from_date:)
-    @timely_ratings ||= Rating.fetch_timely(participant_id: participant_id, from_date: from_date)
+  def timely_ratings
```
shanear: I don't think we should call this method `timely_ratings` any more. just `ratings` will do.

spec/feature/intake/appeal_spec.rb
```diff
@@ -60,8 +60,8 @@
       promulgation_date: receipt_date - untimely_days,
       profile_date: receipt_date - untimely_days + 3.days,
       issues: [
-        { reference_id: "abc123", decision_text: "Untimely rating issue 1" },
-        { reference_id: "def456", decision_text: "Untimely rating issue 2" }
```
shanear: ahhh, yeah keeping this rating around will probably be useful for future tests 👍

app/models/veteran.rb
leikkisa: I think this should be renamed, since "timely" has a specific meaning in this case.

spec/models/rating_spec.rb
```diff
@@ -137,4 +137,41 @@
...
+        participant_id: "DRAYMOND",
+        promulgation_date: receipt_date - 100.years
+      )
+    end
+
+    it "returns rating objects for all ratings" do
+      expect(subject.count).to eq(2)
+    end
+
+    context "when multiple ratings exist" do
```
shanear: multiple ratings exist in the test on line 160, so this test can probably combined with that.

app/models/veteran.rb
jiahuang: should this still be called timely ratings?

app/models/rating.rb
```diff
@@ -45,15 +45,20 @@ def fetch_issues
   end
 
   class << self
+    def fetch_all(participant_id)
+      fetch_timely(participant_id: participant_id, from_date: (Time.zone.today - TWO_LIFETIMES_DAYS))
+    end
+
     def fetch_timely(participant_id:, from_date:)
```
leikkisa: Do we still need this method? I think it could actually be useful to have, such as for the sync, but at the moment it looks like it's no longer needed.

app/models/veteran.rb
```diff
@@ -138,7 +138,7 @@ def participant_id
   end
 
   def sync_rating_issues!
```
leikkisa: We could limit this to ratings that have only been created after the end product was established, since we know no ratings from before that would possibly be a result of the end product.  I am not sure if there will be a big performance impact from having to go through all of a veteran's ratings.

app/models/rating.rb
```diff
@@ -45,15 +45,20 @@ def fetch_issues
...
+      start_date = from_date - ONE_YEAR_PLUS_DAYS
```
jiahuang: It doesn't look like fetch_timely is used besides in fetch_all, so maybe we can just have start_date be the from_date passed into this function instead of -251.5 years?

pkarman:
There's a certain theme in these reviews. That's why naming things is the Hardest Problem! :)

I'd like to keep the `Rating.fetch_timely` method around, because it makes it more clear that we are hacking the VBMS API to get "all" the ratings by picking arbitrarily large date range.

Timeliness is also still a category (see #7301) that we will have to contend with, and I'd be surprised if `fetch_timely` did not end up being used again in some capacity.

The "timely" word is the hard part. Mostly that's because `AMAReview` uses it to do some aggressive caching with `Rails.cache`. I'm not opposed to removing the word from `AMAReview` entirely since we will now be caching all ratings. OTOH we might be caching *both* timely and all ratings, so maybe it's a both/and not an either/or. (This is why Cache Invalidation is the 2nd Hardest Problem.)

What I've done in 49bcbcf13 is add both `ratings` and `timely_ratings` to the Veteran model and corresponding caching in `AMARecord` so that the model offers both options, and it's up to the controller to pick what it wants.

What do you think?

jiahuang:
👍 keeping around both ratings and timely_ratings lgtm esp if we think we'll need both going forward.

PR #7380, opened 2018-10-12
Change assign to person action to reassign for individual generic tasks

### LINKED ISSUE #7210: Create "Re-assign task to team member" task action
For tasks assigned to a team (organization) at the board we will present an "Assign task to team member" option in the task actions dropdown which will create a child task assigned to a person on that team. For tasks assigned to a person having that same option ("Assign task to team member") presents an issue because the child task will return to the original assignee. When the original assignee has left the board (or is on parental leave, military leave, etc.) there will be no way in the UI to complete the original assignee's action and return it back into the flow.

Either in addition to, or in place of the "Assign to team member" option, when a task is assigned to a member of a team and the parent task belongs to that team, we should present an option to "Re-assign to team member" for the person the task is assigned to as well as all other members of the team. The "Re-assign to team member" action should close out the current task and create a new task that is the child of the parent (team) task. This will require work on the front-end as well as the back-end.

## Acceptance criteria
* Somebody can re-assign a task assigned to a member of the same team

lowellrex:
Resolves #7210.

task_reassignment


app/models/tasks/generic_task.rb
```diff
@@ -1,28 +1,33 @@
...
+
+    if assigned_to == user
+      return [
+        Constants.TASK_ACTIONS.ASSIGN_TO_TEAM.to_h,
+        Constants.TASK_ACTIONS.REASSIGN_TO_PERSON.to_h,
+        Constants.TASK_ACTIONS.MARK_COMPLETE.to_h
+      ]
+    end
+
+    if assigned_to.is_a?(Organization)
```
lowellrex: Come to think of it, I think we need to check if the current user is a member of the organization. IF the current user is not a member of the organization we probably should not show the task actions.

lowellrex: Not new with this PR, but we should probably fix this if it really is an issue.

app/models/tasks/generic_task.rb
```diff
@@ -1,28 +1,33 @@
...
+      ]
+    end
+
+    []
   end
 
   def update_from_params(params, current_user)
     verify_user_access!(current_user)
 
+    return reassign(params[:reassign], current_user) if params[:reassign]
```
lowellrex: Return early to avoid the default behaviour of cascading marking parent task as complete calls.

app/models/tasks/generic_task.rb
```diff
@@ -31,6 +36,12 @@ def update_from_params(params, current_user)
     end
   end
 
+  def reassign(reassign_params, current_user)
+    child = self.class.create_child_task(parent, current_user, reassign_params)
+    update!(status: Constants.TASK_STATUSES.completed)
+    children.reject { |t| t.status == Constants.TASK_STATUSES.completed }.each { |t| t.parent_id = child.id }
+  end
```
lowellrex: Meat of the backend change here.

client/app/queue/AssignToView.jsx
```diff
@@ -72,12 +79,51 @@ class AssignToView extends React.Component<Props, ViewState> {
...
+    });
+    const successMsg = { title: `Task reassigned to ${assignee}` };
+
+    return this.props.requestPatch(`/tasks/${task.taskId}`, payload, successMsg).
+      then((resp) => {
+        const response = JSON.parse(resp.text);
+        const preparedTasks = prepareTasksForStore(response.tasks.data);
+
+        this.props.setTaskAttrs(task.externalAppealId, preparedTasks[task.externalAppealId]);
+      });
```
lowellrex: Bulk of the frontend change here.

client/constants/TASK_ACTIONS.json
```diff
@@ -0,0 +1,6 @@
+{
+    "ASSIGN_TO_PERSON": { "label": "Assign to person", "value": "modal/assign_to_person" },
+    "ASSIGN_TO_TEAM": { "label": "Assign to team", "value": "modal/assign_to_team" },
+    "MARK_COMPLETE": { "label": "Mark task complete", "value": "mark_task_complete" },
+    "REASSIGN_TO_PERSON": { "label": "Re-assign to person", "value": "modal/reassign_to_person" }
+}
```
lowellrex: Maybe overkill, but we could use the value of these hashes in the frontend routes so we only have one source of truthy strings for that info... if we'd like.

config/initializers/constants.rb
```diff
@@ -18,4 +18,13 @@ def initialize(hash)
...
   end
+
+  def to_h
+    h = {}
+    singleton_methods.each do |m|
+      val = singleton_method(m).call
+      h[m] = val.is_a?(Subconstant) ? val.to_h : val
+    end
+    h
+  end
```
lowellrex: Added this method so we could pull hashes from our constants while still using the function chain way of accessing constants. Gives us the following:
```
[1] pry(main)> Constants.TASK_ACTIONS.MARK_COMPLETE.to_h
=> {:label=>"Mark task complete", :value=>"mark_task_complete"}
[2] pry(main)> Constants.TASK_ACTIONS.MARK_COMPLETED.to_h
NoMethodError: undefined method `MARK_COMPLETED' for #<Subconstant:0x00007f869aefda80>
Did you mean?  MARK_COMPLETE
from (pry):2:in `<main>'

# Old way of accessing constant hashes doesn't throw exceptions
[3] pry(main)> Constants::TASK_ACTIONS["MARK_COMPLETE"]
=> {"label"=>"Mark task complete", "value"=>"mark_task_complete"}
[4] pry(main)> Constants::TASK_ACTIONS["MARK_COMPLETED"]
=> nil
```

lowellrex: Maybe this should cast hash keys to strings to maintain parity with the old way of accessing constant hashes...

app/models/tasks/generic_task.rb
lowellrex: I tried to replicate this scenario (somebody visits the case details page of an appeal with tasks assigned to an organization that the person visiting the page does not belong to) and could not view the task actions. We must be hiding this on the frontend somehow. Good for us!

client/constants/TASK_ACTIONS.json
mdbenjam: I think this makes sense.

client/app/queue/AssignToView.jsx
```diff
@@ -72,12 +79,51 @@ class AssignToView extends React.Component<Props, ViewState> {
...
+      task,
+      isTeamAssign
+    } = this.props;
+
+    const payload = {
+      data: {
+        task: {
+          reassign: {
+            assigned_to_id: this.state.selectedValue,
+            assigned_to_type: isTeamAssign ? 'Organization' : 'User'
```
mdbenjam: Isn't this always going to be a user?

client/app/queue/AssignToView.jsx
```diff
@@ -72,12 +79,51 @@ class AssignToView extends React.Component<Props, ViewState> {
...
+      }
+    });
+    const successMsg = { title: `Task reassigned to ${assignee}` };
+
+    return this.props.requestPatch(`/tasks/${task.taskId}`, payload, successMsg).
+      then((resp) => {
+        const response = JSON.parse(resp.text);
+        const preparedTasks = prepareTasksForStore(response.tasks.data);
+
+        this.props.setTaskAttrs(task.externalAppealId, preparedTasks[task.externalAppealId]);
```
mdbenjam: This is kind of tricky. We actually need to make a lot of potential changes to the store. Since we updated child tasks as well, and created a new task.

client/app/queue/AssignToView.jsx
```diff
@@ -72,12 +79,51 @@ class AssignToView extends React.Component<Props, ViewState> {
...
+    };
+
+    let assignee = 'person';
+
+    this.options().forEach((opt) => {
+      if (opt.value === this.state.selectedValue) {
+        assignee = opt.label;
+      }
+    });
+    const successMsg = { title: `Task reassigned to ${assignee}` };
```
mdbenjam: I don't know if it's necessary to put the name. If we do, we should also fix this for assigning to a person.

client/app/queue/AssignToView.jsx
mdbenjam: When the instructions field is merge (will be soon) then we should submit the instructions as well as prefill the instructions box with the previous task's instructions.

client/app/queue/AssignToView.jsx
lowellrex: I find it a little more reassuring to see the name of the person/team the task gets assigned to, and adding it doesn't take that much logic so I figured it was worth it. Going to go ahead and make the change to the other assign modals, but happy to not do that as well.

app/controllers/organizations/tasks_controller.rb
```diff
@@ -22,7 +22,8 @@ def organization_url
   def json_tasks(tasks)
     ActiveModelSerializers::SerializableResource.new(
       tasks,
-      each_serializer: ::WorkQueue::TaskSerializer
+      each_serializer: ::WorkQueue::TaskSerializer,
+      user: current_user
```
lowellrex: I think we just forgot to add this when we made this change for the other tasks controllers: https://github.com/department-of-veterans-affairs/caseflow/pull/7175/files#diff-f4416012d567d02da854cfbc9e2df2caR205

client/app/queue/AssignToView.jsx
mdbenjam: I've merged the change with how we index these tasks. This should be 
```suggestion
        preparedTasks.map((preparedTask) => {
          this.props.setTaskAttrs(preparedTask.uniqueId, preparedTask);
        }
```

client/app/queue/QueueApp.jsx
```diff
@@ -242,11 +245,14 @@ class QueueApp extends React.PureComponent<Props> {
             path="/queue/appeals/:appealId/modal/advanced_on_docket_motion"
             render={this.routedAdvancedOnDocketMotion} />
           <Route
-            path="/queue/appeals/:appealId/modal/assign_to_team"
+            path={`/queue/appeals/:appealId/${TASK_ACTIONS.ASSIGN_TO_TEAM.value}`}
```
mdbenjam: Another thing we can do here instead of the multiple routedComponents, and maybe what I should've done the first time is:

```
path={`/queue/appeals/:appealId/modal/:assignType(reassign|assign_to_user|assign_to_team)`}
```

Then in the component you can just check `this.props.assignType`. But this way totally works as well.

client/app/queue/AssignToView.jsx
```diff
@@ -83,16 +90,58 @@ class AssignToView extends React.Component<Props, ViewState> {
         }]
       }
     };
-    const successMsg = {
-      title: `Task assigned to ${this.props.isTeamAssign ? 'team' : 'person'}`
-    };
+
+    let assignee = this.props.isTeamAssign ? 'team' : 'person';
```
mdbenjam: Getting the assignee can be split out into its own method, since it's used below as well.

client/app/queue/QueueApp.jsx
lowellrex: I don't feels strongly about this but do have a slight preference towards explicitly defining each modal as it's own route. While it does cause some repetition in the route declaration and the component to be rendered, it separates the routing logic and the component logic. We jump through the hoop of mapping a route to an `AssignToView` component with optional properties in this file so that the component can implement logic based on the `isTeamAssign` and `isReassignAction` properties instead of the `assignType` matching some value based on the route we used to reach this component.

mdbenjam: I don't feel super strongly either. My case for the other way, is I think it's especially nice for longer flows that can be accessed from a lot of different places. (Think our checkout flows that can be accessed by going through an OMO request, dispatch, and checkout to judge). For modals either way is fine.

PR #7389, opened 2018-10-12
Change message that displays when there are no cases in a queue

lowellrex:

### Description

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...

### DIFF
--- a/client/app/queue/AttorneyTaskListView.jsx
+++ b/client/app/queue/AttorneyTaskListView.jsx
@@ -7,6 +7,7 @@ import _ from 'lodash';
 import StatusMessage from '../components/StatusMessage';
 import TaskTable from './components/TaskTable';
 import AppSegment from '@department-of-veterans-affairs/caseflow-frontend-toolkit/components/AppSegment';
+import Link from '@department-of-veterans-affairs/caseflow-frontend-toolkit/components/Link';
 import Alert from '../components/Alert';
 
 import { workableTasksByAssigneeCssIdSelector } from './selectors';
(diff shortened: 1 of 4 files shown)


client/app/queue/AttorneyTaskListView.jsx
```diff
@@ -57,14 +58,15 @@ class AttorneyTaskListView extends React.PureComponent<Props> {
   render = () => {
     const { messages } = this.props;
     const noTasks = !_.size(this.props.tasks);
-    let tableContent;
 
     if (noTasks) {
-      tableContent = <StatusMessage title={COPY.NO_TASKS_IN_ATTORNEY_QUEUE_TITLE}>
-        {COPY.NO_TASKS_IN_ATTORNEY_QUEUE_MESSAGE}
+      return <StatusMessage>
```
mdbenjam: Maybe not possible, but any chance we could move this text into TaskTable itself? That way we don't have to maintain this in multiple places?

lowellrex: The biggest thing holding us back from doing that is how we separate tables and tabs when displaying cases. Each tab renders its own table ([example from ColocatedTaskView](https://github.com/department-of-veterans-affairs/caseflow/blob/4fefb82e0b9c1303607b9d619a411150ad2f2160/client/app/queue/ColocatedTaskListView.jsx#L97)), so we could have the default tab displaying the "You have no cases assigned to you" message even though the other 2 tabs have cases in them (with different states than the default tab). I think that would be somewhat confusing and worth the cost of maintaining this message in a few different places. Willing to change it though if you disagree.

lowellrex: Although maybe this is the perfect time to move the handling of tabs into the `TaskTable` component itself... We could add a new `tabs` parameter that accepts an array of `TaskTableTab` components that have titles, optional descriptions, and functions to filter the full set of tasks into what is displayed in the table below that one tab.

mdbenjam: That's an interesting idea. I think that makes sense to me. Or maybe even simpler, just a flag that says whether or not we want tabs. Since I think in the future we will have the tabs be the same for all users. Hopefully that future is soon :)

mdbenjam: But you don't have to do it in this PR if you want to just stick with this.

lowellrex: Ticketed: https://github.com/department-of-veterans-affairs/caseflow/issues/7391

PR #7390, opened 2018-10-15
Fixing Reader sidebar

mdbenjam:
Connects https://github.com/department-of-veterans-affairs/appeals-support/issues/3193#issuecomment-427146861

### Description
We have a bug on AMA cases where the sidebar crashes.

### Testing Plan
1. Load Reader with AMA case and open sidebar.


sneha-pai:
@mdbenjam I think this should be all we need for now. We can check on support tickets and other signals that surfacing certain doc details should happen outside the right-side panel.

Issue #7291, opened 2018-10-05
Add issue edge cases | Add Issue Landing | if duplicate lane


suzchap:
## Story
When a user adds issues, they may encounter edge cases where the issue is ineligible. We will not indicate that an issue isn't eligible on any of the add issue selection pages but will indicate it on the Add issue landing page and Confirmation Page so they have the information they need to inform the applicant in the letter they send following intake data entry. If an issue is ineligible in will get logged in Caseflow but not VBMS. [Mural showing edge cases](https://app.mural.co/t/workqueue2001/m/workqueue2001/1537791678087/94469b0cdf1a02a7f57a7628702d36db2f079430).

This ticket is focused on the scenario where the issue being requested is already in another AMA lane and how we display it's ineligibility on the Add Issues Landing page. This scenario can occur in HLR, SC, or Board Appeal lanes.

## Technical notes

- we will need to query for request issues with the same rating_issue_reference_id is on an active decision review (Appeal, SC, HLR), and include this information so the client can render them as ineligible
- we may need to store additional metadata about the issue's ineligibility (like reference ids, etc)

## Acceptance criteria
- when rendering the list of issues on the add issue landing page, if an issue is already active in an AMA lane, we should use red text and add the 'ineligible' copy below
- when the intake is completed and the issue is being saved, save it in the caseflow db with the ineligibility reason but do not create a contention

## Notes
- Identical wording will be used on the confirmation page #7308 
- this only applies to rating issues for now because we don't have a way to know if nonrating issues are duplicates from another AMA lane

## Copy
[Issue selected] is ineligible because it's already under review as a [SC/HLR/Appeal].

## Mocks
<img width="759" alt="screen shot 2018-10-06 at 10 59 08 am" src="https://user-images.githubusercontent.com/11219417/46572620-24b54f80-c957-11e8-8921-671248963682.png">

## Background issues
- #5228 

## How to Validate
(no need, Shane's got this :muscle:)

kencheeto:
@shanear how do we tell if an issue is already being requested in another AMA lane?

shanear:
> @shanear how do we tell if an issue is already being requested in another AMA lane?

There is a request issue with the same rating_issue_reference_id is on an active decision review (Appeal, SC, HLR).

For non rating issues, we'll probably need to show open issues as options. @suzchap have you worked on this at all?

kencheeto:
[...]

Issue #6988, opened 2018-09-14
AMA Intake | Add non-rated issues modal


suzchap:
## Story
For any AMA forms, when a user needs to add issues, they will be walked through a multi-screen selection process to help them make the best choice. The landing page created in #6966 will be their starting point and a modal on top of it will walk the user through rated, non-rated, and an "unknown" option.

When a user doesn't find a match for an issue written on the form with a non rated issue displayed on #6981, they can click the "None of these match, see more options" button to advance to a new modal that displays non-rated issues. The non-rated fields are the same as we've currently have in production except they should not include the "Unknown Issue" option. 

## Acceptance criteria
- From the Add issue modal for rated issues, clicking the "None of these match, see more options" button moves to show another modal with non-rated issues. 
- The non-rated issues are the same as previously but now no longer include "Unknown Issue" as an option.
- "Add this issue" button is grayed out until all the required fields are filled out.

## Copy
- Does issue 2 match any of these issues from past decisions?
- [rest same as before]

## Mocks
<img width="528" alt="screen shot 2018-09-14 at 12 43 57 pm" src="https://user-images.githubusercontent.com/11219417/45563606-df03dc00-b81b-11e8-8b82-d035ed01f617.png">

## Background issues
(especially related design tickets)

## How to Validate
- Doesn't require a specific test user
- Form select > search id > review page: any answers > on "Add issue page, click "+Add issue" > click "None of these match.." to view the non-rated modal to be tested
- TBD: does this need to be tested with all 3 AMA forms or will one suffice?

pkarman:
If there are no predefined issues to connect on the first tab of the modal, should it automatically skip to the 2nd tab?

![screen shot 2018-10-09 at 9 38 32 am](https://user-images.githubusercontent.com/1205061/46677422-3a6d8380-cba8-11e8-9b1a-9a1ca6aac6a3.png)

suzchap:
I think it should just skip to the unrated issue page but would like to check that with @shanear first. Only alternate I could think of would be to have a message that said no rated issues were found which seems like unnecessary work for the user to click through.

shanear:
> If there are no predefined issues to connect on the first tab of the modal, should it automatically skip to the 2nd tab?

really good point. I think we need to do some thinking around this, we'll create a follow up issue to support this scenario.

shanear:
[...]

PR #7393, opened 2018-10-15
Add inActiveReview flag to client and app

pkarman:
connects #7291 connects #6988 

### Description
* Adds `in_active_review` to `RatingIssue.ui_hash`
* Adds style to client for issues in active review.

screen shot 2018-10-15 at 11 43 09 am

### Acceptance Criteria
- [x] Code compiles correctly

### Testing Plan
See #7291

### DIFF
--- a/app/models/rating_issue.rb
+++ b/app/models/rating_issue.rb
@@ -13,11 +13,12 @@ class RatingIssue < ApplicationRecord
   # If you change this method, you will need
   # to clear cache in prod for your changes to
   # take effect immediately.
-  # See AmaReview#cached_serialized_timely_ratings.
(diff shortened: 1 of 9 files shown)


app/models/rating_issue.rb
```diff
@@ -30,6 +31,12 @@ def self.from_bgs_hash(data)
     )
   end
 
+  def in_active_review?
+    ri = RequestIssue.find_by(rating_issue_reference_id: reference_id, removed_at: nil)
+    return ri.review_request_type.underscore.titleize if ri
+    false
```
jiahuang: would prefer nil vs false, otherwise you're returning a string if ri exists or a boolean if it doesn't.

pkarman: fair point. I changed method name to omit the `?` and make it return `nil`.

PR #7395, opened 2018-10-15
Turning on FACOLS for all tests

### LINKED ISSUE #5676: FACOLS Integrations | Hearing prep tests should run on FACOLS
We want to avoid having any hearing scheduling code run off of the fake repositories. Because the hearing scheduling code relies on already-existing models, we need to update those models' unit tests to run off of FACOLS. 

AC:
The following tests run off of FACOLS:
- spec/controllers/hearings/worksheets_controller_spec.rb,
- spec/controllers/hearings_controller_spec.rb,
- spec/feature/hearings_spec.rb,
- spec/models/hearing_spec.rb,
- spec/repositories/hearing_repository_spec.rb,
- spec/models/judge_spec.rb

mdbenjam:
Resolves #5676 

### Description
This PR fixes the remaining tests that need to run on FACOLS. Then it turns on FACOLS for all tests!

### Testing Plan
1. Circle should pass!


app/models/judge.rb
```diff
@@ -27,7 +27,7 @@ def docket?(date)
 
   def upcoming_hearings_on(date, is_fetching_issues = false)
     upcoming_hearings(is_fetching_issues).select do |hearing|
-      hearing.date.between?(date, date.end_of_day)
+      hearing.date.between?(date.beginning_of_day, date.end_of_day)
```
mdbenjam: App change. But was making tests fail. Does this seem right to you?

sharonwarner: We've had a lot of problems with Oracle date matching in these queries. Let me confirm that works in the UI.

sharonwarner: Yep, looks good

Issue #7269, opened 2018-10-04
AMA Edit | Add Issue Landing | save functionality


suzchap:
## Story
For any AMA forms, when a user needs to edit issues, they will click the "Edit in Caseflow" button in VBMS to return to the "Add issues landing page" where they will be able to add new issues and/or remove issues added previously. There are a few "are you sure" type modals to help reduce errors.

This ticket is focused on the landing page's save button and it's destination page.

requires #7268 

## Technical notes
There shouldn't need to be much additional back end logic to support this. We should continue to use `RequestIssuesUpdate`

## Acceptance criteria
- On the Add Issues Landing page, the "Save" button is disabled until an issue is either added or removed.
- When you click the "Save" button, the intake is completed and you go to an Edit Confirmation page.
- When you click "Cancel edit", the intake is canceled and you go to an Edit Canceled page.

## Notes
- The Edit confirmation page can be a very simple dummy page until we work on #7273 

- Out of scope:
-- Design/content for Edit Canceled page will be handled by #7276
-- Design/content for Edit Confirmation page will be handled by #7273 

## Copy
n/a

## Mocks
Same as #7268 

## How to Validate
(no need, Shane's got this :muscle:)

shanear:
So it looks like create contentions fails when you save an edit. This is editing an EP created with the new internal domain code :/

https://sentry.uat.ds.va.gov/sentry/caseflow-uat/issues/1013/

shanear:
works!

PR #7401, opened 2018-10-15
Save button for edit page

leikkisa:
connects #7269


client/app/intake/actions/addIssues.js
```diff
@@ -33,13 +33,13 @@ export const addUnidentifiedIssue = (description, notes) => (dispatch) => {
...
-export const addRatedIssue = (issueId, ratings, isRated, notes) => (dispatch) => {
-  let foundDate = _.filter(ratings, (ratingDate) => _.some(ratingDate.issues, { reference_id: issueId }));
+export const addRatedIssue = (id, ratings, isRated, notes) => (dispatch) => {
+  let foundDate = _.filter(ratings, (ratingDate) => _.some(ratingDate.issues, { reference_id: id }));
 
   dispatch({
     type: ACTIONS.ADD_ISSUE,
     payload: {
-      issueId,
+      id,
```
leikkisa: Needed to rename this so the data structure in addedIssues was the same as the original requestIssues, in order to see if there were changes and if the Save button should be disabled

client/app/intake/pages/addIssues.jsx
```diff
@@ -129,6 +130,7 @@ export const IntakeAddIssuesPage = connect(
     },
     formType: intake.formType,
     veteran: intake.veteran,
+    requestState: null,
```
leikkisa: Just focused on "Edit" errors in this PR, but this should get updated for the Intake Add Issues page too

client/app/intake/reducers/common.js
```diff
@@ -31,16 +31,7 @@ export const commonReducers = (state, action) => {
...
-      isRated: action.payload.isRated,
-      id: action.payload.issueId,
-      profileDate: action.payload.profileDate,
-      category: action.payload.category,
-      description: action.payload.description,
-      decisionDate: action.payload.decisionDate,
-      notes: action.payload.notes,
-      isUnidentified: action.payload.isUnidentified
-    }];
+    let addedIssues = [...listOfIssues, action.payload];
```
leikkisa: Don't want the extra fields when they're not filled in (for the comparison for whether to disable the Save button)

app/models/end_product_establishment.rb
```diff
@@ -86,9 +86,14 @@ def create_contentions!
 
     # Currently not making any assumptions about the order in which VBMS returns
     # the created contentions. Instead find the issue by matching text.
-    create_contentions_in_vbms(issues_without_contentions.pluck(:description)).each do |contention|
+
+    issue_descriptions = request_issues_without_contentions.map do |issue|
+      issue.issue_category ? "#{issue.issue_category} - #{issue.description}" : issue.description
```
pkarman: this should be easier now that the `contention_text` method exists on the `RequestIssue` class.

app/models/end_product_establishment.rb
```diff
@@ -86,9 +86,14 @@ def create_contentions!
...
+    end
+
+    create_contentions_in_vbms(issue_descriptions).each do |contention|
       issue = issues_without_contentions.find do |i|
-        i.description == contention.text && i.contention_reference_id.nil?
+        i.description == contention.text[-i.description.length..-1] && i.contention_reference_id.nil?
```
pkarman: I don't understand this array math?

app/models/request_issue.rb
```diff
@@ -29,7 +29,12 @@ def rated?
     rating_issue_reference_id && rating_issue_profile_date
   end
 
+  def nonrated?
+    issue_category && decision_date
+  end
+
   def contention_text
+    return "#{issue_category} - #{description}" if nonrated?
     is_unidentified ? UNIDENTIFIED_ISSUE_MSG : description
```
pkarman: I like this "return early" pattern. The other option is switching to a `case/when` pattern, which I suspect we would need eventually for the increasing number of use cases.

Could we change line 38 then to:
```ruby
return UNIDENTIFIED_ISSUE_MSG if is_unidentified
description
```

I know @shanear will regret the loss of his beloved ternary operator, but it would be a more consistent pattern.

client/app/intakeEdit/pages/canceled.jsx
```diff
@@ -12,7 +12,7 @@ class Canceled extends Component {
 
     return <div>
       <StatusMessage
-        title="Claim Edit Cancelled"
+        title="Claim Edit Canceled"
```
pkarman: @abbyraskinUSDS would be proud!

spec/models/request_issue_spec.rb
```diff
@@ -54,7 +55,7 @@
     it "changes based on is_unidentified" do
       expect(unidentified_issue.contention_text).to eq(RequestIssue::UNIDENTIFIED_ISSUE_MSG)
       expect(rated_issue.contention_text).to eq("a rated issue")
-      expect(unrated_issue.contention_text).to eq("an unrated rated issue")
+      expect(unrated_issue.contention_text).to eq("a category - a non-rated issue description")
```
pkarman: looks like the variable name needs changing to `non_rated_issue`

app/models/request_issue.rb
leikkisa: Lol, sure, I think I like this better too

spec/models/request_issue_spec.rb
leikkisa: Thanks!

PR #7402, opened 2018-10-15
Only fetch tasks if there are no vacols or caseflow tasks in state when loading case details page

lowellrex:


client/app/queue/CaseDetailsLoadingScreen.jsx
```diff
@@ -55,7 +57,8 @@ class CaseDetailLoadingScreen extends React.PureComponent<Props> {
       );
     }
 
-    if (!tasks || _.filter(tasks, (task) => task.externalAppealId === appealId).length === 0) {
+    if ((!vacolsTasks || _.filter(vacolsTasks, (task) => task.externalAppealId === appealId).length === 0) &&
```
mdbenjam: In the future we may always need to make this request. Where the presence of existing tasks doesn't mean the tasks are up to date. That said, I think for now this is 👌

Issue #7010, opened 2018-09-17
AMA Intake | Contentions for Unidentified issue


suzchap:
As a claims worker, I need to be alerted if there are unidentified issues on a claim review that I need to fix in Caseflow.

requires #6990 

## Acceptance criteria
- When a contention is created for unknown issue, it should have the text "UNIDENTIFIED ISSUE - please click "Edit in Caseflow" button to fix"
- These should also always go on the Rating EP, even if there are no other rating issues.

## Copy
UNIDENTIFIED ISSUE - Please click "Edit in Caseflow" button to fix

## How to Validate
- Doesn't require a specific test user
- Form select > search id > review page: any answers > on "Add issue page, click "+Add issue" > Add an unknown issue.
- Complete the intake
- Go to VBMS and validate the contention text

suzchap:
@shanear this was on our list but I don't think it's a UI ticket so I'll pass back to you to fill out.

shanear:
![screen shot 2018-10-26 at 2 24 39 pm](https://user-images.githubusercontent.com/1596259/47585621-cd8a1580-d92b-11e8-9cef-46ac545a5136.png)
Looking good 👍

shanear:
oh no, it looks like when you add an unidentified contention in the Edit page, it does not use this text:

```
[...]

PR #7405, opened 2018-10-15
Add RequestIssue.contention_text for unidentified issues

pkarman:
connects #7010 

### Description
If a RequestIssue is unidentified, establish the contention in VBMS with string indicating the user must resolve the conflict in Caseflow.

### Acceptance Criteria
- [x] Code compiles correctly

### Testing Plan
See #7010


app/models/request_issue.rb
```diff
@@ -5,6 +5,8 @@ class RequestIssue < ApplicationRecord
   has_many :remand_reasons
   has_many :rating_issues
 
+  UNIDENTIFIED_ISSUE_MSG = "UNIDENTIFIED ISSUE - Please click \"Edit in Caseflow\" button to fix".freeze
```
jiahuang: curious how `"`s show up in vbms. Might be easier to not do the quotes 🤷‍♀️

spec/models/request_issue_spec.rb
```diff
@@ -49,4 +49,12 @@
       expect(unidentified_issues.find_by(id: unidentified_issue.id)).to_not be_nil
     end
   end
+
+  context "#contention_text" do
+    it "changes based on is_unidentified" do
+      expect(unidentified_issue.contention_text).to eq(RequestIssue::UNIDENTIFIED_ISSUE_MSG)
```
jiahuang: Is there another test that verifies that this message doesn't show up for identified issues?

app/models/request_issue.rb
```diff
@@ -27,6 +29,14 @@ def rated?
     rating_issue_reference_id && rating_issue_profile_date
   end
 
+  def contention_text
+    if is_unidentified
+      UNIDENTIFIED_ISSUE_MSG
```
jiahuang: should the text that is in vbms not include the description at all? Donno if something like "UNIDENTIFIED ISSUE - Please click \"Edit in Caseflow\" button to fix : <decision text>" would be clearer or not

app/models/request_issue.rb
```diff
@@ -27,6 +29,14 @@ def rated?
...
+    else
+      description
+    end
```
shanear: maybe my ❤️ for the tertiary operator is clouding my judgement here, but this is a good deal shorter 🙂

`is_unidentified ?  UNIDENTIFIED_ISSUE_MSG : description`

app/models/request_issue.rb
shanear: Good thought @jiahuang. We can validate in UAT this and get back to you on this one.

app/models/request_issue.rb
shanear: We purposely don't want to show them anything about the contention to avoid them actually using as part of a rating or something like that. So it's intentionally unclear lol.

spec/models/request_issue_spec.rb
shanear: it looks like that is happening on line 56/57

PR #7406, opened 2018-10-15
Fix validation on unidentified issues

jiahuang:
Connects #6990

Fixes two issues found in validation:
- "decision date" field does not render if there is no date
- input text for unidentified issues requires that at least a character is entered

add_unidentified_issue


client/app/intake/components/UnidentifiedIssuesModal.jsx
```diff
@@ -51,7 +58,7 @@ class UnidentifiedIssuesModal extends React.Component {
           { classNames: ['usa-button', 'usa-button-secondary', 'add-issue'],
             name: 'Add this issue',
             onClick: this.onAddIssue,
-            disabled: !this.state.description
```
shanear: we could probably use `_.trim` here, and not need some of the other code. Would just change this line to:

```
disabled: !_.trim(this.state.description)
```

client/app/intake/pages/addIssues.jsx
```diff
@@ -58,7 +58,7 @@ export class AddIssuesPage extends React.PureComponent {
               <div className={issueKlasses.join(' ')}>
                 <span className="issue-num">{index + 1}.&nbsp;</span>
                 {issue.text} {addendum}
-                <span className="issue-date">Decision date: {issue.date}</span>
+                { issue.date && <span className="issue-date">Decision date: {issue.date}</span> }
```
shanear: 👍

client/app/intake/components/UnidentifiedIssuesModal.jsx
jiahuang: gonna keep it as is for now since it also does some validation about characters being necessary.

PR #7412, opened 2018-10-16
Rating issue rba_issue_contentions might be array or hash

pkarman:
Addresses production issue https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2645/

### Description
There may be multiple contentions.

### Acceptance Criteria
- [x] Code compiles correctly

### DIFF
--- a/app/models/rating_issue.rb
+++ b/app/models/rating_issue.rb
@@ -23,10 +23,11 @@ class RatingIssue < ApplicationRecord
   end
 
   def self.from_bgs_hash(data)
+    rba_contentions = [data.dig(:rba_issue_contentions) || {}].flatten
     new(
       reference_id: data[:rba_issue_id],
-      profile_date: data.dig(:rba_issue_contentions, :prfil_dt),
-      contention_reference_id: data.dig(:rba_issue_contentions, :cntntn_id),
+      profile_date: rba_contentions.first.dig(:prfil_dt),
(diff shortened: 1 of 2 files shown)


app/models/rating_issue.rb
```diff
@@ -23,10 +23,11 @@ def ui_hash
   end
 
   def self.from_bgs_hash(data)
+    rba_contentions = [data.dig(:rba_issue_contentions) || {}].flatten
     new(
       reference_id: data[:rba_issue_id],
-      profile_date: data.dig(:rba_issue_contentions, :prfil_dt),
-      contention_reference_id: data.dig(:rba_issue_contentions, :cntntn_id),
+      profile_date: rba_contentions.first.dig(:prfil_dt),
```
pkarman: picking the first one is purely arbitrary. if there are multiple, I think we should somehow log/warn and skip the resolution -- unless we can determine that there is only one *legitimate* contention reference.

leikkisa: Don't think this is necessary for this PR, but it seems like you would be able to iterate through them and save any of them that match into the rating_issues table.

pkarman: @leikkisa that's a good point, from a data integrity POV. This method is probably not the place though, since it is intended to return a single `RatingIssue`.

I think we need more real data as examples to know how best to handle it.

Issue #7308, opened 2018-10-06
Add issue edge cases | Confirmation | all ineligible issues


suzchap:
## Story
When a user adds issues, they may encounter edge cases where the issue is ineligible. We will not indicate that an issue isn't eligible on any of the add issue selection pages but will indicate it on the Add issue landing page and Confirmation Page so they have the information they need to inform the applicant in the letter they send following intake data entry. If an issue is ineligible in will get logged in Caseflow but not VBMS. [Mural showing edge cases](https://app.mural.co/t/workqueue2001/m/workqueue2001/1537791678087/94469b0cdf1a02a7f57a7628702d36db2f079430).

This ticket is focused on adding adding a section for ineligible issues to the new confirmation page (#7303). There are currently 5 scenarios where an issue would be displayed in this section (detailed in copy section). These will be saved in caseflow but not VBMS. This applies to all AMA lanes. 

requires #7303 
requires #7313

## Technical notes
- font-awesome [fa-asterisk](https://fontawesome.com/icons/asterisk?style=solid&from=io)

## Acceptance criteria
- When an issue is added that's ineligible, on either the intake or the edit page, it displays on the confirmation page in it's own section, in red, with the reason it's ineligible.

## Notes
- The copy for the ineligible issues should be the same as on the add_issues page
- Out of scope: displaying issues ineligible due to Appeal -> HLR and Appeal -> Appeal, which will happen in #7344 
- Out of scope: displaying issues ineligible due to opt-ins that haven't been approved by the veteran (#7336 )

## Copy
(note: this copy is also used on the Add Issues landing page as specified in issues #7291 , #7293 , #7300 )

Scenario | Display copy 
--- | --- 
(ineligible) If the request is an HLR and same issue was last processed as an HLR | [Issue selected] is ineligible because it was last processed as a Higher-Level Review and this can't be done twice in a row.
(ineligible) If the request is a Board Appeal and the same issue was last processed as a Board Appeal | [Issue selected] is ineligible because it was last processed as a Board Appeal and this can't be done twice in a row.
(Ineligible) If the request is a HLR and same issue was last processed as a Board Appeal | [Issue selected] is ineligible because it was last processed as a Board Appeal which can't be followed by a Higher-Level Review.
(duplicate) If the request is a SC/HLR/Board and same issue is already being processed in another lane | [Issue selected] is ineligible because it's already under review as a [SC/HLR/Appeal].
[...]

PR #7413, opened 2018-10-16
Add ineligible columns to track RequestIssues

pkarman:
connects #7291 

follow on to #7393 

addresses parts of #7301 

### Description

We want to avoid creating duplicate active RequestIssues for a single RatingIssue.

Adds 2 columns to `RequestIssue` to track why we are storing it but not creating a corresponding VBMS contention.

The `ineligible_request_issue` relationship allows one `RequestIssue` to blame another.

Note that this acceptance criteria

> when the intake is completed and the issue is being saved, save it in the caseflow db with the ineligibility reason but do not create a contention

from #7291 will be addressed in a follow-on PR.

### Acceptance Criteria
- [x] Code compiles correctly

### Testing Plan
See #7291


client/constants/REQUEST_ISSUES.json
```diff
@@ -0,0 +1,4 @@
+{
```
leikkisa: Should this file be named differently?  Seems like it is limited to ineligible request issue errors.

app/models/rating_issue.rb
```diff
@@ -33,12 +33,16 @@ def self.from_bgs_hash(data)
...
-    ri = RequestIssue.find_by(rating_issue_reference_id: reference_id, removed_at: nil)
-    return ri.review_request_type.underscore.titleize if ri
+    ri = request_issue_in_review
+    return ri.review_title if ri
   end
 
   private
 
+  def request_issue_in_review
+    RequestIssue.find_by(rating_issue_reference_id: reference_id, removed_at: nil)
```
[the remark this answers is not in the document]
pkarman: Yes you are correct.

This query is the opposite: afaik, we are not allowing multiple active request_issues per a single rating_issue.

Or am I misunderstanding the intent of your question?

app/models/request_issue.rb
```diff
@@ -33,6 +38,20 @@ def contention_text
...
 
+  def update_as_ineligible!(other_request_issue:, reason:)
+    update!(ineligible_request_issue_id: other_request_issue.id, ineligible_reason: reason)
+  end
+
+  def review_title
+    review_request_type.underscore.titleize
+  end
+
+  def ineligible_msg
```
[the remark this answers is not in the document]
pkarman: I can delete the `ineligible_msg` method. It's not strictly necessary.

My assumption was that it might be helpful to render the same message on both the backend and frontend, since there seemed to be a pattern of sharing constants already. But since we have `ineligible_reason` we don't strictly need the human text too.

client/constants/REQUEST_ISSUES.json
pkarman: I can rename. There will be plenty of strings per #7308

pkarman:
thanks for the feedback @leikkisa -- I think I hit everything.

app/models/rating_issue.rb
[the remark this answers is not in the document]
pkarman: I'm seeking clarification here: https://github.com/department-of-veterans-affairs/caseflow/issues/7291#issuecomment-430406865

app/models/request_issue.rb
```diff
@@ -39,16 +62,12 @@ def contention_text
...
-      issue_category: data[:issue_category],
-      notes: data[:notes],
-      is_unidentified: data[:is_unidentified]
-    )
+  def update_as_ineligible!(other_request_issue:, reason:)
+    update!(ineligible_request_issue_id: other_request_issue.id, ineligible_reason: reason)
+  end
+
+  def review_title
+    review_request_type.try(:constantize).try(:review_title)
```
leikkisa: "try" looks handy!

spec/models/request_issue_spec.rb
```diff
@@ -49,6 +49,24 @@
       expect(unidentified_issues.length).to eq(1)
       expect(unidentified_issues.find_by(id: unidentified_issue.id)).to_not be_nil
     end
+
+    context ".in_review_for_rating_issue" do
+      let(:rating_issue) { RatingIssue.new(reference_id: rated_issue.rating_issue_reference_id) }
```
leikkisa: Lol... naming is so fun.

client/app/intake/pages/addIssues.jsx
```diff
@@ -74,7 +75,7 @@ export class AddIssuesPage extends React.Component {
             }
             if (issue.inActiveReview) {
               issueKlasses.push('in-active-review');
-              addendum = `is ineligible because it's already under review as a ${issue.inActiveReview}.`;
+              addendum = INELIGIBLE_REQUEST_ISSUES.in_active_review.replace('{review_title}', issue.inActiveReview);
```
leikkisa: Oh I see why you needed to get the review title from the back-end now, because the review that an issue is under isn't related to whatever front-end flow someone is on.

PR #7414, opened 2018-10-16
Return POA info from BGS for IDT and Queue Case Details

aroltsch:

### Description

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...


app/models/legacy_appeal.rb
```diff
@@ -272,20 +272,26 @@ def outcoded_by_name
...
+    define_method(method_name) do 
+      if use_representative_info_from_bgs?
+        power_of_attorney.send("bgs_#{method_name}".to_sym) 
+      else
+        power_of_attorney.send("vacols_#{method_name}".to_sym) 
+      end
+    end
   end
 
-  def representative_code
```
aroltsch: Do we need to return representative_code for IDT?

app/models/serializers/work_queue/legacy_appeal_serializer.rb
```diff
@@ -73,7 +73,8 @@ class WorkQueue::LegacyAppealSerializer < ActiveModel::Serializer
   attribute :power_of_attorney do
     {
       representative_type: object.representative_type,
-      representative_name: object.representative_name
+      representative_name: object.representative_name,
+      representative_address: object.representative_address
```
mdbenjam: We no longer return the poa information in the appeal serializer for AMA appeals. This is because the BGS calls slow down the case details page load. Instead we have a new endpoint. https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/appeals_controller.rb#L34

app/models/serializers/work_queue/legacy_appeal_serializer.rb
```diff
@@ -70,13 +70,6 @@ class WorkQueue::LegacyAppealSerializer < ActiveModel::Serializer
     object.file_type.eql? "Paper"
   end
 
-  attribute :power_of_attorney do
```
aroltsch: If you remove this, FE will call poa endpoint.

Issue #6984, opened 2018-09-14
AMA Intake | Modal specifications


suzchap:
The modal should scroll to fit large lists of ratings/rating issues. The modal should be completely viewable on different screen sizes and the appropriate width.

## Story
For any AMA forms, when a user needs to add issues, they will be walked through a multi-screen selection process to help them make the best choice. The landing page created in #6966 will be their starting point and a modal on top of it will walk the user through rated, non-rated, and an "unknown" option.

This ticket is to provide additional specifications for how the modal operate for all modal variations.

requires #6990 

## Technical notes
- Add a scenario where a veteran has a lot of ratings and rating issues to demo

## Acceptance criteria
- The entire modal should be viewable on different screen sizes. 
- All modals within the "Add issues" section should be the same size.
- When the modal has more content than can be seen, a scroll bar is added within the modal window.
- When the user scrolls within the modal, the page behind it does not scroll too.

## Copy
N/A

## How to Validate
- Doesn't require a specific test user
- Form select > search id > review, any answers > on "Add issue page, click "+Add issue" to display modal to be tested, then also... > advance through the "None of these match" path to view the non-rated modal and Unknown issue modal.
- TBD: does this need to be tested with all 3 AMA forms or will one suffice?

suzchap:
This should be done after all modals have been created so they can all be validated as being a consistent size.

leikkisa:
Sometimes Caseflow users open it on half of the screen

leikkisa:
For reference, my Windows GFE has a screen resolution of 1366x768

shanear:
Nice!
![screen shot 2018-10-26 at 1 03 58 pm](https://user-images.githubusercontent.com/1596259/47581560-afb6b380-d91f-11e8-829f-42b5e3fd9821.png)

shanear:
responsive AF

PR #7417, opened 2018-10-16
Add Issues Modal Styling

leikkisa:
connects #6984


leikkisa:
Carola and I decided to make the buttons smaller when the screen is smaller, so they still all fit on one line.

Smaller buttons
screen shot 2018-10-17 at 11 54 39 am

Regular
screen shot 2018-10-17 at 11 55 03 am

client/app/styles/_intake.scss
```diff
@@ -53,8 +53,17 @@
 }
 
 .intake-add-issues .cf-modal-body {
-  width: 770px;
-  overflow-y: auto;
+  width: 750px;
+  max-height: 95%;
+
+  @media only screen and (max-width: 760px) {
+    width: 650px;
```
pkarman: what happens when the screen width is less than 650px? I wonder if we should use `max-width: 95%` for the modal so it is always relative to the screen width?

leikkisa: I wasn't doing fully responsive design, just designing for the case where someone might use Caseflow on half of their screen.  Admittedly, I determined this number based on my Windows laptop resolution, which probably isn't the most thorough.  We got some feedback that the Queue research team designs for a common browser width of 768px.

If we use max-width, then the three modals won't have the same width (the ones with less wide content will be narrower), but we could try to do that for the smaller sizes.  The buttons will move at a smaller size, but it doesn't look that bad.  I can put in another case for smaller than 650px where that happens.

pkarman: cool. let's try it and see what feedback we get.

PR #7418, opened 2018-10-16
Rename Add Issues Page

leikkisa:
connects #7266


leikkisa:
I left the page filename the same (addIssues), but let me know if you feel like we should update that as well.

PR #7420, opened 2018-10-16
Adds script to delete duplicate Decass records

evankroske:
Part of https://github.com/department-of-veterans-affairs/dsva-vacols/issues/43

## Description
Adds a script to delete some duplicate Decass records that have been created since August 6.

### Duplicate a Decass Record
```sql
insert into decass
select *
from decass
where rownum <= 1;
```

### Run the Script
```bash
bundle exec rails runner scripts/duplicate_decass_deleter.rb --nodry_run
```

### Confirm That the Duplicate Decass Records is Gone
```sql
select count(*)
from decass
where defolder = '<defolder of a record deleted by the script>';
```


scripts/duplicate_decass_deleter.rb
```diff
@@ -0,0 +1,119 @@
...
+
+DRY_RUN = !ARGV.include?("--nodry_run")
+if !DRY_RUN
+  puts "WARNING: This is NOT a dry run."
+end
+
+defolders = VACOLS::Decass.select("defolder")
+  .where("deadtim >= ?", Date.new(2018, 8, 16))
+  .group("defolder").having("count(*) > 1").map(&:defolder)
+puts "Found #{defolders.length} cases with too many Decass records"
```
amprokop: Hmm.. what about OMO requests? Won't there be more than one DAS record for an OMO request?

evankroske:
I think this will delete most of the duplicates. It's going to delete 80 records out of at most 152 duplicates. The other duplicates will be harder to identify.

PR #7421, opened 2018-10-16
Consolidate modals

mdbenjam:
Connects https://github.com/department-of-veterans-affairs/caseflow/pull/7275

### Description
While working on moving the actions to the backend, I realized that we could consolidate a couple of our modals. This PR consolidates the mark complete actions for generic queues with the send back to an attorney/send to team co-located actions.

### Testing Plan
1. Go to co-located user and send a task back to the attorney and send a task to an organization.
1. Go to generic user and complete a task.


client/app/queue/components/CompleteTaskModal.jsx
```diff
@@ -36,25 +36,23 @@ type Params = {|
...
     getContent: ({ assignerName }: { assignerName: string }) => <React.Fragment>
-      {COPY.COLOCATED_ACTION_SEND_BACK_TO_ATTORNEY_COPY}&nbsp;
-      <b>{sprintf(COPY.COLOCATED_ACTION_SEND_BACK_TO_ATTORNEY_COPY_ATTORNEY_NAME, assignerName)}</b>
+      {sprintf(COPY.MARK_TASK_COMPLETE_COPY, assignerName)}
     </React.Fragment>,
-    buttonText: COPY.COLOCATED_ACTION_SEND_BACK_TO_ATTORNEY_BUTTON
+    buttonText: COPY.MARK_TASK_COMPLETE_BUTTON
   },
-  [SEND_TO_LOCATION_MODAL_TYPES.team]: {
+  send_colocated_task: {
```
mdbenjam: In the not-to-distant future this will be assign task to an organization instead. But currently it just completes the task and the backend handles the complexity. Once this is spun off into an assign to org task, we can massively simplify this file.

client/app/queue/components/CompleteTaskModal.jsx
```diff
@@ -98,53 +93,45 @@ class SendToLocationModal extends React.Component<Props> {
...
   saveState: state.ui.saveState.savePending
 });
 
 const mapDispatchToProps = (dispatch) => bindActionCreators({
-  hideModal,
   requestPatch,
   setTaskAttrs
 }, dispatch);
 
+const propsToText = (props) => {
```
mdbenjam: This is something we'll be able to get rid of when we split this component up into complete vs assign.

client/app/queue/components/EditModalBase.jsx
```diff
@@ -20,9 +20,13 @@ export default function editModalBase(ComponentToWrap, title) {
     getWrappedComponentRef = (ref) => this.wrappedComponent = ref;
 
     closeHandler = () => {
-      this.props.history.goBack();
+      this.props.history.replace('/queue');
     }
 
+    title = () => title || (propsToText && propsToText(this.props).title);
```
mdbenjam: I don't love this propsToText method. Hoping to get rid of it.

client/COPY.json
```diff
@@ -187,8 +187,9 @@
 
   "HEARING_SCHEDULE_SUCCESS_MESSAGE": "You can view your uploaded schedule by clicking the link below",
 
-  "TASK_MARKED_COMPLETE_NOTICE_TITLE": "Task marked as complete",
-  "TASK_MARKED_COMPLETE_NOTICE_DETAIL": "Task for %s's case completed",
+  "MARK_TASK_COMPLETE_TITLE": "Mark this task complete",
```
mdbenjam: Making the language apply to both colocated and generic tasks.

client/app/queue/components/EditModalBase.jsx
```diff
@@ -20,9 +20,13 @@ export default function editModalBase(ComponentToWrap, title) {
     getWrappedComponentRef = (ref) => this.wrappedComponent = ref;
 
     closeHandler = () => {
-      this.props.history.goBack();
+      this.props.history.replace('/queue');
```
lowellrex: I'm not sure we should be sending the user to `/queue`. Hearing scheduling will soon be sending folks to the case details page where they will have available task actions, so they could conceivably end up invoking this handler and landing at `/queue`.

I think you floated this question in Slack yesterday, but I think it might be most desirable to remain on the case details page when we close the modal. Doing so would remove the complexity of handling cases like the above.

client/app/queue/QueueApp.jsx
```diff
@@ -160,9 +158,7 @@ class QueueApp extends React.PureComponent<Props> {
 
   routedAssignToUser = (props) => <AssignToView {...props.match.params} />;
 
-  routedMarkTaskComplete = (props) => <MarkTaskCompleteView
-    nextStep={`/queue/appeals/${props.match.params.appealId}`}
-    {...props.match.params} />;
+  routedCompleteTaskModal = (props) => <CompleteTaskModal {...props.match.params} />;
 
   triggerModal = (props) => <TriggerModal modal={props.match.params.modalType} />;
```
lowellrex: I think we can remove this line now.

client/app/queue/components/CompleteTaskModal.jsx
```diff
@@ -98,53 +93,45 @@ class SendToLocationModal extends React.Component<Props> {
...
-      {SEND_TO_LOCATION_MODAL_TYPE_ATTRS[modalType].getContent(this.getContentArgs())}
-    </Modal>;
+    return this.props.task ? SEND_TO_LOCATION_MODAL_TYPE_ATTRS[this.props.modalType].getContent(this.getContentArgs()) : null;
   };
 }
 
 const mapStateToProps = (state: State, ownProps: Params) => ({
-  task: tasksForAppealAssignedToUserSelector(state, ownProps)[0],
+  task: tasksForAppealAssignedToUserSelector(state, ownProps)[0] ||
+    incompleteOrganizationTasksByAssigneeIdSelector(state, { appealId: ownProps.appealId })[0],
```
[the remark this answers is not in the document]
lowellrex: FWIW, I don't think we need to solve this problem in this PR, it's just something that I've been noodling on so I figured it was worth a mention here in case you had ideas.

client/app/queue/components/CompleteTaskModal.jsx
```diff
@@ -98,53 +93,45 @@ class SendToLocationModal extends React.Component<Props> {
...
+      teamName: (props.task && props.task.action) ? CO_LOCATED_ADMIN_ACTIONS[props.task.action] : ''
+    }),
+    button: SEND_TO_LOCATION_MODAL_TYPE_ATTRS[props.modalType].buttonText
+  }
+}
+
 export default (withRouter(
-  connect(mapStateToProps, mapDispatchToProps)(SendToLocationModal)
+  connect(mapStateToProps, mapDispatchToProps)(editModalBase(
+    CompleteTaskModal, { propsToText }
```
lowellrex: I think `propsToText()` returns an object. I'm not well versed enough in Javascript to know whether it will automatically collapse the nested object, but if it doesn't I think this line might need to be:
```jsx
    CompleteTaskModal, propsToText()
```

client/app/queue/components/CompleteTaskModal.jsx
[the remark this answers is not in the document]
mdbenjam: Yeah I agree we should solve this in a single place. The easiest right now is probably just a selector. Though also on my radar is all these places need to change to support multiple tasks as well :(

client/app/queue/components/CompleteTaskModal.jsx
mdbenjam: I want to pass it as a function, not call it here since it needs the underlying props :/

client/app/queue/components/EditModalBase.jsx
mdbenjam: The weird thing right now, and the reason I changed it is the current behavior for the colocated modals is to redirect to queue. Currently if you still get back to the case details screen you can still take actions even though it's been assigned to someone else. We need to prevent this properly, but haven't gotten around to that yet so I figured the best thing to do would be to just continue the go back to queue.

mdbenjam: We can also maybe make this configurable.

client/app/queue/components/CompleteTaskModal.jsx
lowellrex: Gotcha gotcha

client/app/queue/components/EditModalBase.jsx
lowellrex: Makes sense to me. I think making this configurable would be ideal, but that could probably live in another PR.

Issue #7270, opened 2018-10-04
AMA Edit | Add Issue Landing | Remove alert modal


suzchap:
## Story
For any AMA forms, when a user needs to edit issues, they will click the "Edit in Caseflow" button in VBMS to return to the "Add issues landing page" where they will be able to add new issues and/or remove issues added previously. There are a few "are you sure?" type modals to help reduce errors. [Entire flow via Mural](https://app.mural.co/t/workqueue2001/m/workqueue2001/1537795938997/f5834747503f5a248340b9f3426fbac652812400)

This ticket is focused on the "Are you sure?" modal that shows anytime you click the "Remove" link.

requires #7268 

## Acceptance criteria
- From the Add Issues Landing page, when you click "remove" a modal appears asking the user to confirm that they really want to remove that issue. 
- When you click "Cancel" the modal closes to reveal the Add Issues Landing page with no changes.
- When you click "Yes, remove issue" the modal closes to reveal the Add Issues Landing page with the selected issue removed.

## Copy
Remove issue
The contention you selected will be removed from the EP in VBMS.
Are you sure you want to remove this issue?
Cancel
Yes, remove issue

## Mocks
<img width="661" alt="screen shot 2018-10-05 at 9 51 33 am" src="https://user-images.githubusercontent.com/11219417/46539376-54475780-c884-11e8-906e-0cb04b40dc46.png">

## How to Validate
(no need, Shane's got this :muscle:)

suzchap:
@shanear this modal should also appear via the summary during the original intake flow... so should that be added to this ticket or a new one created?

kencheeto:
They're quite small but it's worth breaking them up anyway. I took a stab here: https://github.com/department-of-veterans-affairs/caseflow/issues/7311

shanear:
![screen shot 2018-10-26 at 2 45 08 pm](https://user-images.githubusercontent.com/1596259/47591782-d46e5380-d93e-11e8-932d-5789f3fae730.png)
works like a charm 👍

PR #7422, opened 2018-10-16
Remove issue modal

jiahuang:
Connects #7270

remove_issue_modal


client/app/styles/_intake.scss
```diff
@@ -87,13 +87,12 @@
         border-bottom: 1px solid $color-gray-lighter;
 
         div {
-          display: table-column;
-          float: left;
+          display: table-cell;
```
jiahuang: Change around the css here. Otherwise issues with short descriptions like "left knee" would have the row not take up 100% of the width.

client/app/styles/_intake.scss
```diff
@@ -137,6 +135,7 @@
 
   td:last-child {
     min-width: 130px;
+    max-width: 500px;
```
jiahuang: 🤷‍♂️ just based on a guess that looked ok

client/app/intake/pages/addIssues.jsx
```diff
@@ -119,6 +141,11 @@ export class AddIssuesPage extends React.PureComponent {
         intakeData={intakeData}
         closeHandler={this.props.toggleUnidentifiedIssuesModal} />
       }
+      { this.props.toggleIssueRemoveModal && intakeData.removeIssueModalVisible && <RemoveIssueModal
```
leikkisa: Don't think you need ```this.props.toggleIssueRemoveModal```  here

Issue #7404, opened 2018-10-15
Create generic tasks / mail rollout


laurjpeterson:
# Background
Veterans and appellants send in all different types of mail at any point in the appeals process. Mail is typically sent to the central mail processing center in Janesville, Wisconsin. Mail comes into the VBA or BVA P.O. Box. Employees in Janesville upload the mail in the Mail Portal. If the mail is for BVA, BVA employees process it in the following way:
- Mail supervisor assigns mail to individual mail staff in the Mail Portal
- Mail employee opens the mail in the mail portal 
- Mail employee searches for the Veteran to confirm there is an appeal record. 
- Mail employee adds a "mail tab" in VACOLS and assigns it based on routing rules. This is a mail task for someone at the Board to do
- Mail employee uploads the mail to VBMS utilizing centralized mail portal

We need to build a way to create these tasks in Caseflow so (1) the mail team can create and assign tasks and (2) so other Board teams can create and assign tasks from these mail tasks.

# Strategy
- Build a manual mail task creation piece up front first before diving into VBMS document types
  - As Alex pointed out, it helps us de-risk the business end by checking the correctness of an automated mail task creation system later on. For example, if users were checking the mail portal and creating MailTasks, theoretically we could build this automated piece which would create FakeMailTasks, then check to see how the results are different. Potentially if we had time we could also use VACOLS mail to accomplish the same end.

# Suggested flow
- Mail supervisor assigns mail to individual mail staff in the Mail Portal
- Mail employee opens the mail in the mail portal 
- Mail employee searches the appellant's ID in Caseflow 
- Mail employee clicks into the appellant's correct `Case details` page.
- Mail employee creates a mail task from the Actions dropdown
- Mail employee fills in a variety of data about the mail task and assigns it to an organization

**Open Questions:**
- Can we auto-fill task assignees like VACOLS does today? Is it worth it?
   - Note: we are planning to do this, yes.

# Rollout ideas
- Start with initial list of mail tasks that can be worked by VLJ Support Staff
       - Implementing active tasks is required before this can begin
       - Note: we think that we can allow mail team to create tasks for AOD + Privacy teams right off the bat, too
- Progressively add additional teams mail tasks can be assigned to
- Progressively add additional mail task types to the list that can be chosen

### 1. AMA first
See routing logic, here https://github.com/department-of-veterans-affairs/caseflow/issues/7906

1. Allow mail team to create mail tasks and assign to VLJ Support Staff
[...]

PR #7423, opened 2018-10-16
Create mail task

lowellrex:
Connects #7404.

Adds ability for any member of the Mail team to assign a task to any Organization branching off of the RootTask.

## Test plan
* Log in as "JOLLY D. POSTMAN"
* Search for an appeal
* Click on the "Create mail task" action
* Assign it to a team
* Voila!

mail_task


app/models/tasks/mail_task.rb
```diff
@@ -0,0 +1,30 @@
+class MailTask < GenericTask
+  class << self
+    def create_from_params(params_array, current_user)
+      verify_user_can_assign!(current_user)
+
+      params_array.map do |params|
+        root_task = RootTask.find_by(appeal_id: params[:appeal].id)
```
lowellrex: If we want to start accepting mail for legacy cases immediately we could create a `RootTask` if one does not already exist (which it wouldn't for legacy cases).

client/app/queue/AssignToView.jsx
```diff
@@ -9,7 +9,8 @@ import COPY from '../../COPY.json';
 import {
   appealWithDetailSelector,
   tasksForAppealAssignedToUserSelector,
-  incompleteOrganizationTasksByAssigneeIdSelector
+  incompleteOrganizationTasksByAssigneeIdSelector,
+  rootTasksForAppealIfMailTeamMemberSelector
```
lowellrex: I think we have to do this because we need to show the `RootTask` if the current user is a member of the mail team. Part of the issue we discussed in this other PR: https://github.com/department-of-veterans-affairs/caseflow/pull/7421#discussion_r225953388

app/views/queue/index.html.erb
```diff
@@ -5,6 +5,7 @@
     userRole: (current_user.vacols_roles.first || "").capitalize,
     userCssId: current_user.css_id,
     userIsVsoEmployee: current_user.vso_employee?,
+    userIsMailTeamMember: MailTeam.singleton.user_has_access?(current_user),
```
lowellrex: I think we need to pass this property to the frontend in order to determine whether to show the `RootTask` or not. This feels clumsy to me, so I'd love to hear if you have thoughts about how to avoid doing this.

app/models/tasks/mail_task.rb
mdbenjam: I think we hold off on legacy for now. Find by is good.

client/app/queue/CaseSnapshot.jsx
```diff
@@ -319,16 +328,18 @@ export class CaseSnapshot extends React.PureComponent<Props> {
...
 
 const mapStateToProps = (state: State, ownProps: Params) => {
-  const { featureToggles, userRole } = state.ui;
+  const { featureToggles, userRole, userIsMailTeamMember } = state.ui;
 
   return {
     appeal: appealWithDetailSelector(state, { appealId: ownProps.appealId }),
     featureToggles,
     userRole,
+    userIsMailTeamMember,
```
mdbenjam: I don't think we actually need this in the component. The root task selector takes care of this for us.

app/models/task.rb
```diff
@@ -42,9 +42,15 @@ def self.recently_completed
     where(status: Constants.TASK_STATUSES.completed, completed_at: (Time.zone.now - 2.weeks)..Time.zone.now)
   end
 
-  def self.create_from_params(params, current_user)
-    verify_user_can_assign!(current_user)
-    params = params.each { |p| p["instructions"] = [p["instructions"]] if p.key?("instructions") }
+  def self.create_many_from_params(params_array, current_user)
+    params_array.map { |params| create_from_params(params, current_user) }.flatten
```
mdbenjam: Do we need to flatten? No harm in doing so, just curious.

app/models/tasks/mail_task.rb
```diff
@@ -0,0 +1,33 @@
+class MailTask < GenericTask
+  class << self
+    def create_from_params(params, user)
+      verify_user_can_assign!(user)
+
+      root_task = RootTask.find_by(appeal_id: params[:appeal].id)
+      unless root_task
+        fail(Caseflow::Error::NoRootTask, message: "Could not find root task for appeal with ID #{params[:appeal]}")
```
mdbenjam: Should we be creating one if it doesn't exist? Maybe not yet, but in the future this might be a path forward for legacy appeals.

app/views/queue/index.html.erb
mdbenjam: We shouldn't show a task that has no available actions. That's my suggestion.

app/models/tasks/mail_task.rb
lowellrex: I think we decided not to do this last week, but am happy to reverse course if you think it best. https://github.com/department-of-veterans-affairs/caseflow/pull/7423#discussion_r226014594

app/models/task.rb
lowellrex: I don't recall, removing it.

app/views/queue/index.html.erb
lowellrex: I really like this idea! This is a much better approach than the way I was going with PR #7432.

Issue #6939, opened 2018-09-12
Editing VBA Issues | Re-associate rated issues on edit issues


leikkisa:
## Story
When a claim review is edited, and issues are added or removed, the entire map for associate_rated_issues needs to be re-sent.  It will replace the previous map.

## Acceptance Criteria
- When a claim review is edited successfully (meaning issues were either added or removed), and contentions were added or removed to any EPs connected to that claim review, Caseflow should call the VBMS request to associate all contentions to rated issues for each EP on the claim review.

## Technical Notes
- There is no logical problem with sending the same data to AssociateRatingIssues twice (it will just affect performance).

## Validation
- Edit a claim_review issues
- Check with VBMS that the new rated issue map matches the new issues selected

shanear:
RemoveContentions is not working with the permissions VBMS gave our system user. This is blocked until we fix that 😞

shanear:
Confirmed this is working with Joe 👍

PR #7424, opened 2018-10-16
Re-do rating issue contention map when request issues are edited

leikkisa:
connects #6939

### DIFF
--- a/app/models/claim_review.rb
+++ b/app/models/claim_review.rb
@@ -43,6 +43,10 @@ class ClaimReview < AmaReview
     end
   end
 
+  def unassociate_rated_request_issues!
+    request_issues.select(&:rated?).each { |ri| ri.update!(rating_issue_associated_at: nil) }
+  end
+
   # Idempotent method to create all the artifacts for this claim.
   # If any external calls fail, it is safe to call this multiple times until
   # establishment_processed_at is successfully set.
--- a/app/models/end_product_establishment.rb
+++ b/app/models/end_product_establishment.rb
@@ -176,18 +176,16 @@ class EndProductEstablishment < ApplicationRecord
   end
 
   def create_associated_rated_issues!
-    request_issues_to_associate = unassociated_rated_request_issues
-
     is_rated = true
     return if code != source.issue_code(is_rated)
-    return if request_issues_to_associate.empty?
+    return if unassociated_rated_request_issues.count == 0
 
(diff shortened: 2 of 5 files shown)


app/models/claim_review.rb
```diff
@@ -43,6 +43,10 @@ def create_issues!(new_issues)
     end
   end
 
+  def unassociate_rated_request_issues!
```
shanear: I think the name of this method is a little confusing, since we aren't actually unassociating them. Maybe `mark_rated_request_issues_to_reassociate`?

app/models/end_product_establishment.rb
```diff
@@ -176,18 +176,16 @@ def status_active?(sync: false)
   end
 
   def create_associated_rated_issues!
```
shanear: maybe not part of this PR, but this method name is also a little misleading. It implies that we are creating request issues, which we are not.

shanear: Should just be called `associate_rated_issues!`

app/models/rating_issue.rb
```diff
@@ -34,7 +34,7 @@ def self.from_bgs_hash(data)
 
   def in_active_review
     ri = RequestIssue.find_by(rating_issue_reference_id: reference_id, removed_at: nil)
-    return ri.review_request_type.underscore.titleize if ri
+    return ri.review_request_type.underscore.titleize if ri && ri.review_request_type
```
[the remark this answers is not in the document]
shanear: should this change have a unit test?

spec/models/request_issues_update_spec.rb
```diff
@@ -265,6 +280,15 @@
...
         expect(Fakes::VBMSService).to have_received(:remove_contention!).with(request_issue_contentions.last)
+
+        new_map = rated_end_product_establishment.send(:rated_issue_contention_map, review.request_issues.reload)
+
+        expect(Fakes::VBMSService).to have_received(:associate_rated_issues!).with(
+          claim_id: rated_end_product_establishment.reference_id,
+          rated_issue_contention_map: new_map
+        )
+
+        expect(review.request_issues.first.rating_issue_associated_at).to be_within(1.hour).of(Time.current)
```
shanear: lol this is one way to do it i guess. I think we can use `Timecop.freeze` in this spec to match exactsies

PR #7425, opened 2018-10-17
Make labels strong

leikkisa:
connects #6965


client/app/intake/components/CancelIntakeModal.jsx
```diff
@@ -115,40 +115,42 @@ class CancelIntakeModal extends BaseForm {
...
-    return <Modal
-      buttons={[
-        { classNames: ['cf-modal-link', 'cf-btn-link', 'close-modal'],
-          name: 'Close',
-          onClick: closeHandler
-        },
-        { classNames: ['usa-button', 'usa-button-secondary', 'confirm-cancel'],
-          name: 'Cancel intake',
-          onClick: this.handleSubmitCancel
+    return <div className="intake-cancel">
```
leikkisa: Adding a div because the default modal body has different label colors, but looks weird to have two label colors

PR #7432, opened 2018-10-17
[WIP] Determine which tasks for an appeal are actionable on the back-end

lowellrex:


app/models/task.rb
```diff
@@ -94,15 +94,25 @@ def when_child_task_completed
     update_status_if_children_tasks_are_complete
   end
 
+  def actionable?(user)
+    return false if completed? || on_hold?
+    can_be_accessed_by_user?(user)
```
lowellrex: We may need to add a function called `can_be_actioned_by_user?()` so that the assigners and assignees of parent tasks cannot action the child/assigned task.

app/models/appeal.rb
```diff
@@ -167,6 +167,10 @@ def create_tasks_on_intake_success!
     RootTask.create_root_and_sub_tasks!(self)
   end
 
+  def actionable_tasks_for_user(user)
+    tasks.select { |t| t.actionable?(user) }.sort_by { |e| e.assigned_to && e.assigned_to == user }
```
lowellrex: Sort tasks that are assigned to the current user first so that plucking the first task off of this list returns the same task as before.

client/app/queue/CaseDetailsLoadingScreen.jsx
```diff
@@ -57,11 +72,17 @@ class CaseDetailLoadingScreen extends React.PureComponent<Props> {
...
+      ((!vacolsTasks || _.filter(vacolsTasks, (task) => task.externalAppealId === appealId).length === 0) &&
+      (!caseflowTasks || _.filter(caseflowTasks, (task) => task.externalAppealId === appealId).length === 0))
+    ) {
       const taskPromise = ApiUtil.get(`/appeals/${appealId}/tasks?role=${userRole}`).then((response) => {
         const legacyTasks = _.every(response.body.tasks, (task) => task.attributes.appeal_type === 'LegacyAppeal');
 
+        _.forEach(response.body.actionable_tasks, (val, key) => {
+          this.props.setActionableTasksForAppeal(key, tasksFromObjectArray(val));
+        });
+
```
lowellrex: We will make a request for actionable tasks for this appeal unless we already have them. I figure that makes sense, but maybe we should try to be smarter about how often we make this call?

PR #7439, opened 2018-10-18
Do not create contentions for ineligible issues

pkarman:
connects #7291 

### Description
Adds feature tests and prevents ineligible issues from generating contentions.

### Acceptance Criteria
- [x] Code compiles correctly
- [x] No contentions are created for ineligible issues

### Testing Plan

see #7291

### DIFF
--- a/app/models/end_product_establishment.rb
+++ b/app/models/end_product_establishment.rb
@@ -79,7 +79,7 @@ class EndProductEstablishment < ApplicationRecord
   # VBMS will return ALL contentions on a end product when you create contentions,
   # not just the ones that were just created.
   def create_contentions!
-    issues_without_contentions = request_issues_without_contentions
+    issues_without_contentions = request_issues_ready_for_contentions
     return if issues_without_contentions.empty?
 
     set_establishment_values_from_source
@@ -218,8 +218,8 @@ class EndProductEstablishment < ApplicationRecord
     rated_request_issues.select { |ri| ri.rating_issue_associated_at.nil? }
   end
 
-  def request_issues_without_contentions
-    request_issues.select { |ri| ri.contention_reference_id.nil? }
+  def request_issues_ready_for_contentions
+    request_issues.select { |ri| ri.contention_reference_id.nil? && ri.eligible? }
   end
 
   def rated_issue_contention_map(request_issues_to_associate)
--- a/app/models/request_issue.rb
+++ b/app/models/request_issue.rb
@@ -4,9 +4,8 @@ class RequestIssue < ApplicationRecord
   has_many :decision_issues
   has_many :remand_reasons
   has_many :rating_issues
-  belongs_to :ineligible_request_issue, class_name: "RequestIssue"
 
-  enum ineligible_reason: { in_active_review: 0, untimely: 1 }
+  enum ineligible_reason: { duplicate_of_issue_in_active_review: 0, untimely: 1 }
 
   UNIDENTIFIED_ISSUE_MSG = "UNIDENTIFIED ISSUE - Please click \"Edit in Caseflow\" button to fix".freeze
 
@@ -43,19 +42,14 @@ class RequestIssue < ApplicationRecord
         notes: data[:notes],
         is_unidentified: data[:is_unidentified]
       )
-      if data[:reference_id]
-        existing_request_issue = in_review_for_rating_issue(data[:reference_id])
-        if existing_request_issue
-          request_issue.ineligible_request_issue = existing_request_issue
-          request_issue.ineligible_reason = :in_active_review
-        end
-      end
+      request_issue.check_for_existing_request_issue!
       request_issue
     end
 
-    def in_review_for_rating_issue(rating_issue)
-      reference_id = rating_issue.is_a?(RatingIssue) ? rating_issue.reference_id : rating_issue
-      unscoped.find_by(rating_issue_reference_id: reference_id, removed_at: nil, ineligible_reason: nil)
+    def find_active_by_reference_id(reference_id)
+      request_issue = unscoped.find_by(rating_issue_reference_id: reference_id, removed_at: nil, ineligible_reason: nil)
+      return unless request_issue && request_issue.status_active?
+      request_issue
     end
   end
 
@@ -78,10 +72,6 @@ class RequestIssue < ApplicationRecord
     description
   end
 
-  def update_as_ineligible!(other_request_issue:, reason:)
-    update!(ineligible_request_issue_id: other_request_issue.id, ineligible_reason: reason)
-  end
-
   def review_title
     review_request_type.try(:constantize).try(:review_title)
   end
@@ -100,5 +90,13 @@ class RequestIssue < ApplicationRecord
       notes: notes,
       is_unidentified: is_unidentified
     }
+  end
+
+  def check_for_existing_request_issue!
+    return unless rating_issue_reference_id
+    existing_request_issue = self.class.find_active_by_reference_id(rating_issue_reference_id)
+    if existing_request_issue
+      self.ineligible_reason = :duplicate_of_issue_in_active_review
+    end
   end
 end
--- a/app/models/request_issue.rb
+++ b/app/models/request_issue.rb
@@ -10,8 +10,6 @@ class RequestIssue < ApplicationRecord
 
   UNIDENTIFIED_ISSUE_MSG = "UNIDENTIFIED ISSUE - Please click \"Edit in Caseflow\" button to fix".freeze
 
-  delegate :status_active?, to: :end_product_establishment
-
   class << self
     def rated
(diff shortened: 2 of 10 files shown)


app/models/request_issue.rb
```diff
@@ -32,7 +30,11 @@ def no_follow_up_issues
     end
 
     def from_intake_data(data)
-      new(
+      # there's a race condition here: if 2 intakes were being entered simultaneously that
```
shanear: We have a limitation on multiple intakes happening at the same time for the same veteran for the same form. This doesn't prevent a HLR & SC to happen simultaneously, but I don't see this being a big problem.

leikkisa: Two intakes can't be happening on the same Veteran at the same time, will that prevent this race condition from happening?

app/models/end_product_establishment.rb
```diff
@@ -79,7 +79,7 @@ def perform!(commit: false)
   # VBMS will return ALL contentions on a end product when you create contentions,
   # not just the ones that were just created.
   def create_contentions!
-    issues_without_contentions = request_issues_without_contentions
+    issues_without_contentions = request_issues_ready_for_contentions
```
leikkisa: issues_without_contentions seems superfluous, is it here to be more self-documenting or is there another reason?

app/models/request_issue.rb
```diff
@@ -41,13 +43,27 @@ def from_intake_data(data)
...
+        existing_request_issue = in_review_for_rating_issue(data[:reference_id])
+        if existing_request_issue
+          request_issue.ineligible_request_issue = existing_request_issue
+          request_issue.ineligible_reason = :in_active_review
+        end
+      end
+      request_issue
     end
 
     def in_review_for_rating_issue(rating_issue)
```
shanear: hmm, it's sort of confusing that `rating_issue` can also be a string? I've seen the Rails code base use the pattern of using `or` in a variable name if it can be two things. Something like `in_review_for_rating_issue (rating_issue_or_rating_issue_reference_id)`

shanear: It's also confusing that `in_review` and `active` seem to mean the same thing. We should choose one term or the other.

spec/feature/intake/appeal_spec.rb
```diff
@@ -351,6 +361,14 @@ def check_row(label, text)
     expect(page).to have_content("3 issues")
     expect(page).to have_content("This is an unidentified issue")
 
+    # add ineligible issue
+    safe_click "#button-add-issue"
+    find_all("label", text: "Old injury in review").first.click
+    safe_click ".add-issue"
+
+    expect(page).to have_content("4 issues")
+    expect(page).to have_content("4. Old injury in review is ineligible")
```
shanear: it'd be nice to check for more of the ineligibility reason here, so we know it captured the right reason

app/models/request_issue.rb
```diff
@@ -41,13 +43,27 @@ def from_intake_data(data)
         notes: data[:notes],
         is_unidentified: data[:is_unidentified]
       )
+      if data[:reference_id]
+        existing_request_issue = in_review_for_rating_issue(data[:reference_id])
+        if existing_request_issue
+          request_issue.ineligible_request_issue = existing_request_issue
+          request_issue.ineligible_reason = :in_active_review
+        end
+      end
```
shanear: It feels like this is something that could be captured in a private method specifically designed to validate the request issue. Or something that could be calculated lazily? Right now we require the RequestIssue to be created via `from_intake_data` if you want the validations.

def ineligible_reason

spec/feature/intake/supplemental_claim_spec.rb
```diff
@@ -533,6 +554,26 @@ def check_row(label, text)
                is_unidentified: true,
                end_product_establishment_id: end_product_establishment.id
       )).to_not be_nil
+
+      xyz789_request_issues = RequestIssue.where(rating_issue_reference_id: in_active_review_rating_issue_reference_id)
+      expect(xyz789_request_issues.count).to eq(2)
+
+      ineligible_issue = xyz789_request_issues.select(&:in_active_review?).first
```
shanear: `in_active_review` is a really confusing scope. For almost the whole time reading this PR, i thought it meant it was, well, in active review. Maybe this could be renamed `conflicting_with_another_active_review` or something like that.

spec/feature/intake/appeal_spec.rb
```diff
@@ -381,6 +399,15 @@ def check_row(label, text)
...
     )).to_not be_nil
+
+    xyz789_request_issues = RequestIssue.where(rating_issue_reference_id: in_active_review_rating_issue_reference_id)
+
+    expect(xyz789_request_issues.count).to eq(2)
+
+    ineligible_issue = xyz789_request_issues.select(&:in_active_review?).first
+
+    expect(ineligible_issue).to be_in_active_review
+    expect(ineligible_issue.ineligible_request_issue).to eq(request_issue_in_progress)
```
leikkisa: I know this might be a pain to change at this point, but I think "ineligible_request_issue" might be a confusing name for the first (in progress) request issue (because that one isn't ineligible... right?). I almost might not be understanding fully.

app/models/end_product_establishment.rb
pkarman: `request_issues_ready_for_contentions` performs a couple of selects to filter out the necessary `request_issues`. To avoid running those selects every time we access the array, we just create a copy of the array locally and operate on that. It's the same thing as memoizing the array like this:

```
def request_issues_ready_for_contentions
  @request_issues_ready_for_contentions ||= ...
end
```

except that in this case, we are modifying the array elements in the `each` loop at the end, so that they are no longer "ready for contentions" -- because contentions were just assigned to them. If they were memoized, the next time you called `request_issues_ready_for_contentions` you would get an array of request issues that already had contentions assigned, which is not what you would expect.

spec/feature/intake/appeal_spec.rb
pkarman: Probably a better name would be `duplicate_of_request_issue`

pkarman: On 3rd thought, I don't really think we need it all. We can look up the original active request issue by reference id if we need to. The FK doesn't actually satisfy any business requirement.

app/models/request_issue.rb
pkarman: I thought about adding it as a `validates` using AR feature, but we want to mutate the object, not prevent it from being saved.

It seemed to me to be important to guard the integrity of no duplicates at the boundary where data enters the app, which is in `from_intake_data`. Otherwise we must enforce that integrity outside the `RequestIssue` class by calling an additional method somewhere, and IMO that introduces a risk of forgetting later on.

I did refactor it out into its own instance method. I couldn't make it private and still call it from inside the class method though.

pkarman:
Thanks @shanear and @leikkisa for the good feedback.

I agree about the naming. We already have such similar names ("request issue", "rated issue", "rating issue") that I conflated the "in active review" status between `RequestIssue` and `RatingIssue`.

A `RatingIssue` may be in active review already, and we know that because we have an existing `RequestIssue` with the same reference_id.

I changed the naming to use `duplicate_of_issue_in_active_review` on `RequestIssue` and `in_active_review` on the `RatingIssue`. It's duplicate `RequestIssue`s that we are trying to avoid, so that seemed clearer to me.

Let me know what you think.

I also dropped the `ineligible_request_issue` column altogether because we don't need to know, specifically, which `RequestIssue` was the duplicate at that moment. The `reference_id` serves as enough of a FK if we need to query it sometime in the future. I didn't put the migration in this PR yet though. If you agree, I'll follow up with it later as a false-start vestigial clean up.

app/models/request_issue.rb
```diff
@@ -41,13 +42,22 @@ def from_intake_data(data)
         notes: data[:notes],
         is_unidentified: data[:is_unidentified]
       )
+      request_issue.check_for_existing_request_issue!
+      request_issue
     end
 
-    def in_review_for_rating_issue(rating_issue)
-      find_by(rating_issue_reference_id: rating_issue.reference_id, removed_at: nil, ineligible_reason: nil)
+    def find_active_by_reference_id(reference_id)
```
shanear: yeah i like this name a lot

spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -509,18 +509,31 @@ def check_row(label, text)
       expect(row).to have_text(text)
     end
 
+    let(:in_active_review_rating_issue_reference_id) { "xyz789" }
```
shanear: this variable name seems to be a hold out from the old naming (there are a bunch of other places in the tests where we do this)

pkarman:
going to merge under the rationale of Good Enough Naming For Now. we can revisit if we need to.

PR #7453, opened 2018-10-18
Call veteran endpoint separately for the case details

### LINKED ISSUE #7426: Return veteran address from BGS for legacy appeals
Veteran name and address should be pulled from BGS for legacy appeals

aroltsch:
Resolves #7426 

### Description
Case details - make a call to get veteran's information separately. 
Pull veteran address, date of birth and date of death from BGS


app/controllers/appeals_controller.rb
```diff
@@ -39,6 +39,35 @@ def power_of_attorney
     }
   end
 
+  # Veteran address and death date are the only data that is being pulled from BGS,
+  # The rest are from VACOLS for now
+  def veteran
```
aroltsch: I will move this into a serializer.

client/app/queue/AppellantDetail.jsx
```diff
@@ -22,24 +22,12 @@ export default class AppellantDetail extends React.PureComponent {
...
     COPY.CASE_DETAILS_GENDER_FIELD_VALUE_MALE;
 
-  getDetails = ({ nameField, genderField, dobField, addressField, relationField, regionalOfficeField }) => {
+  getDetails = ({ nameField, addressField, relationField }) => {
     const details = [{
       label: 'Name',
       value: this.getAppealAttr(nameField)
     }];
 
-    if (genderField && this.getAppealAttr(genderField)) {
```
aroltsch: I think some of these were mistakenly added.

app/controllers/appeals_controller.rb
```diff
@@ -39,6 +39,35 @@ def power_of_attorney
...
+        address_line_2: appeal.veteran_address_line_2,
+        city: appeal.veteran_city,
+        state: appeal.veteran_state,
+        zip: appeal.veteran_zip,
+        country: appeal.veteran_country
+      },
+      regional_office: nil
+    }
+
+    if appeal.regional_office
```
mdbenjam: It's a little weird to me that regional office is lumped in with veteran info. But I guess that's the way it's been!

app/controllers/appeals_controller.rb
```diff
@@ -111,6 +140,14 @@ def appeal
     @appeal ||= Appeal.find_appeal_by_id_or_find_or_create_legacy_appeal_by_vacols_id(params[:appeal_id])
   end
 
+  def formatted_date_of_birth
```
mdbenjam: We should probably just format them the same. Same with date of death.

client/app/queue/AppellantDetail.jsx
mdbenjam: Agreed that regional office seems not appropriate here. But dob seems relevant and maybe gender as well?

client/app/queue/VeteranDetail.jsx
```diff
@@ -5,61 +5,83 @@ import _ from 'lodash';
...
+
+    if (genderValue) {
       details.push({
         label: COPY.CASE_DETAILS_GENDER_FIELD_LABEL,
-        value: this.getGenderValue(genderField)
+        value: genderValue
       });
     }
-    if (dobField && this.getAppealAttr(dobField)) {
+    const dod = this.props.veteranInfo.date_of_birth;
```
mdbenjam: probably should be `dob` instead of `dod`

client/app/queue/VeteranDetail.jsx
```diff
@@ -5,61 +5,83 @@ import _ from 'lodash';
...
-      });
-    }
-    if (relationField && this.getAppealAttr(relationField)) {
-      details.push({
-        label: 'Relation to Veteran',
-        value: this.getAppealAttr(relationField)
-      });
-    }
-    if (addressField && this.getAppealAttr(addressField)) {
+    const address = this.props.veteranInfo.address;
```
mdbenjam: So one thing we can do to make this a little more straightforward is to deconstruct the object at the top of the function. This is a pattern we use on a lot of the frontend:

```
const {
  address, // This one is named the same in the object (this.props) as what we want to call it.
  full_name: name, // The key is the name in the object, the value is what we want to call it.
  date_of_birth: dob
} = this.props;
```

mdbenjam: Might just help to clean it up a little.

client/app/queue/AppellantDetail.jsx
aroltsch: I don't think using veteran's gender and veteran's dob is correct under the claimant's tab. The component is only using three appellant's fields anyways: https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/AppellantDetail.jsx#L72. Or are you saying to leave it for the future when we add appellant's gender and dob?

app/controllers/appeals_controller.rb
aroltsch: Go figure

app/controllers/appeals_controller.rb
aroltsch: Let me see what I can do. I've noticed that some of them are strings not dates.

client/app/queue/VeteranDetail.jsx
aroltsch: Oh yeah, good suggestion

app/controllers/appeals_controller.rb
aroltsch: Yeah, the dates that are coming from Vacols are dates and the ones from BGS are already formatted strings. I can add a comment.

aroltsch: I can probably switch to using BGS date of birth for both the same as date of death.

client/app/queue/AppellantDetail.jsx
mdbenjam: Oh I didn't realize this is pulling the veteran gender/dob. I was thinking we were also pulling appellant gender/dob. Yeah agreed then.

aroltsch: Yupss

Issue #7427, opened 2018-10-17
Cleaning up user organizations


mdbenjam:
## Problem
We currently have a mishmash of ways in which we control who is in each organization. For example:

- Attorney <-> Judge team mappings are stored in `Constants::AttorneyJudgeTeams`
- Co-located team for round robin purposes is stored in `Constants::CoLocatedTeams`
   - Co-located is also retrieved from VACOLS using the condition `staff_record.stitle == "A1" || staff_record.stitle == "A2"`
- Dispatch is determined four our pilot by `Constants::BvaDispatchTeams`
   - Dispatch is also determined from VACOLS using the condition `staff_record.sdept == "DSP"`
- Attorneys are determined by the presence of a `sattyid` in VACOLS
- VSOs are determined by their roles in BGS

This complex web is difficult to track, not particularly extensible, and requires a deploy to change. 

We've generally abided by the principle that it's best to rely on VACOLS for this information. However, VACOLS poses two problems.
1) VACOLS doesn't actually store team information for a lot of users. Lowell in talking with Jed has had to ask him to add people for Quality Review.
2) VACOLS doesn't provide a clean way of allowing a person to be on multiple teams.

## Solutions
I propose a new way to keep track of user relationships to organizations. I know we've been reluctant to manage users <-> organization mappings in Caseflow because there is a new contract out to manage the HR system within VACOLS. While ideally this team would eventually handle this problem, I think we need an interim solution that allows us to easily change these team mappings.

I see two options:
1) Maintain all organization information in a giant ruby hash:
```
{
  <organzation1>: [<user1>, <user2>, ...],
  <organzation2>: [<user1>, <user2>, ...],
  ...
}
```
2) Maintain this mapping in the DB. Create an OrganizationalUser join table that stores this mapping. And build a simple UI to manage it. The UI would look like this:
```
[Organizations | Users] <- Toggle button to show list of organizations or users

Organization 1:
   - User 1
   - User 2
   - [Add User] <- Text box to add users

Organization 2:
   - User 1
   - User 3
   - [Add User] <- Text box to add users

--------- User version of the page --------

User 1: Organization1, Organization2, [Add Organization] <- Text box to add organization
User 2: Organization1, Organization3, [Add Organization] <- Text box to add organization
```

I lean towards option 2 for a few reasons
1) It allows us to store this relationship information in our DB which lets us simply use ActiveRecord magic to understand the relationships. `user.organization` and `organization.users` simply work.
[...]

PR #7454, opened 2018-10-18
Add user organizations

mdbenjam:
Connects: https://github.com/department-of-veterans-affairs/caseflow/issues/7455
Resolves: https://github.com/department-of-veterans-affairs/caseflow/issues/7427

### Description
This PR creates a new Organization-User join table to manage what users are part of which organization. It also creates a UI at `/organizations/<org url>/users` to manage the users in an organization. Any user within the organization can add or remove other users from this page.

Next steps: we need to change our organization class to return the members of the organization from this new join table instead of using VACOLS.

### Testing Plan
1. Log in as the org queue user
1. Go to `/organizations/translation/users`
1. Add and remove users to your hearts content

organizationusers


client/app/queue/OrganizationUsers.jsx
```diff
@@ -0,0 +1,157 @@
...
+import ApiUtil from '../util/ApiUtil';
+import Alert from '../components/Alert';
+import Button from '../components/Button';
+import SearchableDropdown from '../components/SearchableDropdown';
+
+export default class OrganizationUsers extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.state = {
```
mdbenjam: I didn't want to add this stuff to the store. It really doesn't belong in our existing reducers and is completely isolated (for the time being). In the future if we need to expand this we should create a new reducer and move the state there. It's unclear if this should even belong in queue, so I'm trying to keep it as lightweight as possible.

app/controllers/organizations/users_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+  before_action :verify_organization_access
+  before_action :verify_role_access
+  before_action :verify_feature_access
+
+  def index
+    respond_to do |format|
+      format.html { render template: "queue/index" }
+      format.json do
+        organization_users = organization.users
+        remaining_users = User.where.not(id: organization_users.pluck(:id))
```
lowellrex: Do we have any concerns about revealing too much information through this interface? We will be returning 2,000+ users CSS IDs and names. Does that matter?

mdbenjam:
Lowell those are both super valid concerns! Chris seems to think we're okay from a privacy perspective, and I'm pretty confident we'll be okay from a browser perspective. Though we will need some better long term solutions!

Issue #7407, opened 2018-10-15
AMA Edit | Edit link should work for all EPs associated with a claim review


shanear:
Right now, only the edit link for the rating EP associated with a claim review properly takes you to the edit page. We need that to work for the non-rating EP and any other EPs associated with the claim review.

PR #7468, opened 2018-10-19
Edit URL

leikkisa:
connects #7407


app/views/higher_level_reviews/edit.html.erb
```diff
@@ -4,6 +4,7 @@
     dropdownUrls: dropdown_urls,
     feedbackUrl: feedback_url,
     buildDate: build_date,
-    serverIntake: higher_level_review.ui_hash(FeatureToggle.enabled?(:intakeAma, user: current_user))
+    serverIntake: higher_level_review.ui_hash(FeatureToggle.enabled?(:intakeAma, user: current_user)),
+    claimId: end_product_establishment.reference_id
```
leikkisa: I could just use the url param directly here since the rest of the data is from the higher_level_review. Does that make more sense to do?

Also if not, should this be in a ui_hash? It didn't seem necessary to me.

leikkisa: Okay, I agree with not using the URL param directly even though that's the only thing I'm using right now (also it's possible in the future we'll get more from the EP establishment).

The claimId is actually part of an EndProductEstablishment. A higher level review can have many claimIds.  The edit page shows the data for the whole higher level review with all the claimIds.  I guess the reason we look it up by claimId is because that's what VBMS has.

Before it was coded to just return one rated claimId, which is part of why the edit page wasn't working for non-rated end products.

leikkisa: In this case it's the other way around, the HLR is coming from the EP establishment.  Funny enough, the claimId isn't really used for much else besides the URL :-P.

pkarman: ok that makes sense. thanks for the explanation. 👍

leikkisa:
Talked to Shane and he said using the param would be okay for the claim id since it's only used in the URL, and that it would be preferable from having to send it back and forth

PR #7472, opened 2018-10-19
Introduce better filtering for decass records

aroltsch:
Paul pointed out that we cannot use `created_at` date for decass records for attorneys; we need to use attorney_id. And as far as judges, we will use the latest updated decass record.


app/repositories/queue_repository.rb
```diff
@@ -151,14 +151,25 @@ def reassign_case_to_attorney!(judge:, attorney:, vacols_id:, created_in_vacols_
...
     def filter_duplicate_tasks(records)
-      # Keep the latest assignment if there are duplicate records
+      # Keep the latest updated assignment if there are duplicate records
       records.group_by(&:vacols_id).each_with_object([]) do |(_k, v), result|
-        result << v.sort_by(&:created_at).last
+        sorted = v.sort_by(&:updated_at)
+
+        # If the user is an attorney, find all associated with the user's attorney_id
+        if attorney_id_match_found?(sorted)
+          sorted.select! { |task| task.attorney_id == RequestStore[:current_user].vacols_attorney_id }
```
amprokop: nit: do we typically use the in-place modifying `!` functions? not sure but i thought we avoided them, it's ok if you want to keep this though.

aroltsch: why do you think we should be avoiding them?

Issue #7300, opened 2018-10-05
Add issue edge cases | if untimely | Add issue landing page


suzchap:

shanear:
- We are not accounting for the case where the issue "is ineligible because the decision was made before the Appeals Modernization Act go live date on Feb 14, 2019"

Here's a good reference for the timeliness rules:

A) If the rating issue is part of a RAMP decision,
then it is untimely if it's more than 1 year ago.
(also do not show the VACOLS matching step)

B) If a VACOLS issue is matched to the issue,
then it can not be untimely,
however, it is invalid for opt-in if the soc date is more than 60 days ago and the nod date is more than 1 year ago.

C) If it is neither A or B,
then it is untimely if it's more than 1 year ago,
it is also invalid if the decision was made prior to Feb 13, 2019 (the AMA start date)

pkarman:
@shanear We have been using `Rating.promulgation_date` to determine timeliness.

There is currently no `promulgation_date` for the child `RatingIssue`, only a `profile_date`. Do we always inherit the `Rating` parent's `promulgation_date`?

Is there a `promulgation_date` in the BGS response for each issue that we are currently not using?

What is the relationship with `profile_date`? Could we be using that to calculate timeliness for each issue?

leikkisa:
@pkarman I believe the profile_date is when the decision is made, and the promulgation_date is when it's made official and the letter is sent out. So promulgation_date (which I think is generally a few days after profile_date) is used for determining timeliness, but the date that Vets know about (and fill in on their forms) is the decision_date, so we show the decision_date on the UI so that the claim assistants are better able to match the decisions being appealed.

All issues will have both of them, but I don't think we're using the promulgation_date for anything but timeliness so we might not continue making it available after that.

pkarman:
@shanear I will need some help understanding how to do this:

> B) If a VACOLS issue is matched to the issue,

shanear:
Something is off about how this is working with nonrating issues. They all appear to be invalid.

<img width="1112" alt="screen shot 2018-10-26 at 1 12 06 pm" src="https://user-images.githubusercontent.com/1596259/47581921-d5908800-d920-11e8-9752-9951256f5b53.png">

pkarman:
Should be fixed in master now @shanear -- please give it another go

shanear:
[...]

PR #7473, opened 2018-10-19
Add RatingIssue promulgation_date

pkarman:
connects #7300 

### Description
The promulgation_date is on the VBMS Rating, but we need it to evaluate timeliness on all the children RatingIssue. So store it in the database so we can always perform timeliness calculation locally.

This the first of multiple PRs to implement the timeliness check.

### Acceptance Criteria
- [x] Code compiles correctly

### DIFF
--- a/app/models/ama_review.rb
+++ b/app/models/ama_review.rb
@@ -15,7 +15,7 @@ class AmaReview < ApplicationRecord
   before_destroy :remove_issues!
 
   cache_attribute :cached_serialized_ratings, cache_key: :timely_ratings_cache_key, expires_in: 1.day do
-    ratings_with_issues.map(&:ui_hash)
+    receipt_date && ratings_with_issues.map(&:ui_hash)
   end
 
   def self.review_title
@@ -60,6 +60,8 @@ class AmaReview < ApplicationRecord
   private
 
   def ratings_with_issues
+    return unless receipt_date
+
     veteran.ratings(receipt_date: receipt_date).reject { |rating| rating.issues.empty? }
   end
 
--- a/app/models/ama_review.rb
+++ b/app/models/ama_review.rb
@@ -64,7 +64,7 @@ class AmaReview < ApplicationRecord
   private
 
   def ratings_with_issues
-    veteran.ratings.reject { |rating| rating.issues.empty? }
+    veteran.ratings(receipt_date: receipt_date).reject { |rating| rating.issues.empty? }
   end
 
   # disabled for simplecov sake
--- a/app/models/ama_review.rb
+++ b/app/models/ama_review.rb
@@ -14,11 +14,7 @@ class AmaReview < ApplicationRecord
 
   before_destroy :remove_issues!
 
-  # cache_attribute :cached_serialized_timely_ratings, cache_key: :timely_ratings_cache_key, expires_in: 1.day do
-  #   receipt_date && timely_ratings_with_issues.map(&:ui_hash)
-  # end
-
-  cache_attribute :cached_serialized_ratings, cache_key: :ratings_cache_key, expires_in: 1.day do
+  cache_attribute :cached_serialized_ratings, cache_key: :timely_ratings_cache_key, expires_in: 1.day do
     ratings_with_issues.map(&:ui_hash)
   end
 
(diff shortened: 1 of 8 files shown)


app/models/rating_issue.rb
```diff
@@ -38,8 +41,16 @@ def in_active_review
...
   end
 
+  def timely?
+    less_than_one_year_old?
+  end
+
   private
 
+  def less_than_one_year_old?
+    promulgation_date >= (Time.zone.today - Rating::ONE_YEAR_PLUS_DAYS)
```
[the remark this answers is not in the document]
pkarman: > Timeliness is based on receipt_date (this was actually a past bug we had to fix). 

I probably need a conversation about this, since the code suggests that promulgation_date is the important column. https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/rating.rb#L60

Re: the "duplicate" methods there are going to be [additional checks](https://github.com/department-of-veterans-affairs/caseflow/issues/7300#issuecomment-428352741) added in a follow on PR, so `less_than_one_year_old?` will not be the only method called in `timely?`.

app/models/ama_review.rb
```diff
@@ -14,12 +14,8 @@ class AmaReview < ApplicationRecord
 
   before_destroy :remove_issues!
 
-  # cache_attribute :cached_serialized_timely_ratings, cache_key: :timely_ratings_cache_key, expires_in: 1.day do
-  #   receipt_date && timely_ratings_with_issues.map(&:ui_hash)
-  # end
-
-  cache_attribute :cached_serialized_ratings, cache_key: :ratings_cache_key, expires_in: 1.day do
-    ratings_with_issues.map(&:ui_hash)
+  cache_attribute :cached_serialized_ratings, cache_key: :timely_ratings_cache_key, expires_in: 1.day do
```
shanear: `cache_key: :timely_ratings_cache_key` we no longer need this, since we aren't using the receipt_date in our fetch from BGS. This will make the cache a lot cleaner.

app/models/ama_review.rb
```diff
@@ -14,12 +14,8 @@ class AmaReview < ApplicationRecord
...
+    receipt_date && ratings_with_issues.map(&:ui_hash)
```
shanear: Not sure we should require the `receipt_date` here. Seems like unnecessary entangling. I think it's nicer if we just return the entire list of issues regardless.

app/models/veteran.rb
```diff
@@ -121,11 +121,13 @@ def zip_code
...
   def timely_ratings(from_date:)
-    @timely_ratings ||= Rating.fetch_timely(participant_id: participant_id, from_date: from_date)
+    @timely_ratings ||= Rating.fetch_timely(participant_id: participant_id, from_date: from_date).each do |rating|
+      rating.receipt_date = from_date
+    end
   end
 
-  def ratings
-    @ratings ||= Rating.fetch_all(participant_id)
+  def ratings(receipt_date:)
```
shanear: Hmm, I see that passing `receipt_date` through to the `Rating` and then to the `RatingIssue` allows rating issues to determine for themselves whether they are timely. But the hoops we're having to go through in order to do that seem overly cumbersome. It also doesn't seem like `receipt_date` is conceptually a value of a `Rating` or a `RatingIssue`, it feels much more natural as an argument to `timely?`. Something like this: `RatingIssue.timely?(for_receipt_date: Time.zone.today)`

Because we are asking if a `RatingIssue` is timely for a decision review with a certain receipt date.

db/migrate/20181018191048_add_promulgation_date_to_rating_issue.rb
```diff
@@ -0,0 +1,5 @@
+class AddPromulgationDateToRatingIssue < ActiveRecord::Migration[5.1]
+  def change
+    add_column :rating_issues, :promulgation_date, :datetime, null: false
```
shanear: We already have `rating_issues` in the DB. Will this require any backfilling of data?

app/models/ama_review.rb
```diff
@@ -64,27 +60,18 @@ def remove_issues!
...
-    veteran.ratings.reject { |rating| rating.issues.empty? }
-  end
-
-  # disabled for simplecov sake
-  # def timely_ratings_with_issues
-  #  return unless receipt_date
-  #
-  #  veteran.timely_ratings(from_date: receipt_date).reject { |rating| rating.issues.empty? }
-  # end
+    return unless receipt_date
```
shanear: Same as the comment on line 18. I'm not sure we should make receipt_date a requirement for finding a list of all ratings.

pkarman:
sorry @shanear this got merged before I saw your review. I can reapproach in my next PR.

shanear:
No worries 🙂  My fault for getting to it so late

Issue #6938, opened 2018-09-12
RELEASE DEC 9: Add user ID to establish_claim and create_contentions


leikkisa:
## Story
There's an issue where a VBMS claim gets rolled back, but the BGS claim is still being created.  Right now when we create a claim, the user is listed as efolder_express.  We would like to send the actual user ID of the person logged into Caseflow, for both establish_claim and create_contentions, because these two calls use external domains.

VBMS needs to develop the functionality for this, and then provide documentation on how to send the data in the calls.

## Acceptance criteria
- establish_claim call to VBMS includes the user_id of the person logged into Caseflow
- create_contentions call to VBMS also includes the user id
- remove_contention call to VBMS should use the user id of the system user

## Technical Notes
- This should  be toggleable by a feature flag.

This can be done by adding the following user information to the VBMS security header:

```
  <soapenv:Header>
  	<ext:Security xmlns:ext="http://vbms.vba.va.gov/external">
		<ext:cssUserName>nherro_ssuper</ext:cssUserName>
		<ext:cssStationId>317</ext:cssStationId>
		<ext:SecurityLevel>9</ext:SecurityLevel>
	</ext:Security>
  </soapenv:Header>
```

We will likely need to modify `VBMS::Requests::EstablishClaim`, `VBMS::Requests::CreateContentions`, and `VBMS::Requests::RemoveContention` to take in a user object so the information can be injected into the header.

shanear:
Nathan says this should be ready to go

shanear:
Next steps are to reach out to Nate/Joe about how to pass the information, and how we can test that on UAT.

leikkisa:
I sent them a message to ask about this

leikkisa:
From Joe today:
This work isn’t complete, the plan is to make available, during establishClaim and createContentions, and user and station number field that we can then pass through to BGS. Once this work is complete we will include some documentation/examples regarding how it will work from the CASEFLOW side

shanear:
This should be released to UAT by 10/8, instructions should have been sent from Joe on how to use that service.

leikkisa:
Here's how the headers should look:
```
<ext:Security xmlns:ext="http://vbms.vba.va.gov/external">
                <ext:cssUserName>nherro_ssuper</ext:cssUserName>
                <ext:cssStationId>317</ext:cssStationId>
                <ext:SecurityLevel>9</ext:SecurityLevel>
</ext:Security>
```
[...]

PR #7476, opened 2018-10-19
send user id to vbms

jiahuang:
Connects #6938

* bumps vbms
* sends user that did the intake with establish claim & create contention for hlr & sc
  * if sc is created through dta, sends system user
* sends current user with dispatch on establish claim
* sends system user with remove contention


app/models/establish_claim.rb
```diff
@@ -185,7 +185,8 @@ def update_claim_establishment!(attrs = {})
   def establish_claim_in_vbms(end_product)
     VBMSService.establish_claim!(
       claim_hash: end_product.to_vbms_hash,
-      veteran_hash: appeal.veteran.to_vbms_hash
+      veteran_hash: appeal.veteran.to_vbms_hash,
+      is_claim_review: false
```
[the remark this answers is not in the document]
shanear: It looks like it is. Should be able to grep for `VBMSService.establish_claim!` to see.

app/services/external_api/vbms_service.rb
```diff
@@ -93,16 +93,17 @@ def self.clean_document(location)
     File.delete(location)
   end
 
-  def self.establish_claim!(veteran_hash:, claim_hash:)
+  def self.establish_claim!(veteran_hash:, claim_hash:, is_claim_review:)
```
shanear: We should be able to tell if it is a claim review by looking at the claim_hash and looking at the EP code.

app/services/external_api/vbms_service.rb
```diff
@@ -93,16 +93,17 @@ def self.clean_document(location)
...
     @vbms_client ||= init_vbms_client
 
     request = VBMS::Requests::EstablishClaim.new(
       veteran_hash,
       claim_hash,
-      v5: FeatureToggle.enabled?(:claims_service_v5)
+      v5: FeatureToggle.enabled?(:claims_service_v5),
+      send_userid: FeatureToggle.enabled?(:vbms_include_user)
```
shanear: nice 👍

app/services/external_api/vbms_service.rb
```diff
@@ -93,16 +93,17 @@ def self.clean_document(location)
     File.delete(location)
   end
 
-  def self.establish_claim!(veteran_hash:, claim_hash:)
+  def self.establish_claim!(veteran_hash:, claim_hash:, is_claim_review:)
     @vbms_client ||= init_vbms_client
```
shanear: now that we are doing this: 

```
is_claim_review ? vbms_client_with_user : vbms_client_with_system_user
```

do we need the `@vbms_client` variable?

app/services/external_api/vbms_service.rb
```diff
@@ -173,6 +176,20 @@ def self.vbms_client_with_user
...
     )
+
+    FeatureToggle.enabled?(:vbms_include_user) ? @vbms_client_with_user : @vbms_client
+  end
+
+  def self.vbms_client_with_system_user
+    @vbms_client_with_system_user ||= VBMS::Client.from_env_vars(
+      logger: VBMSCaseflowLogger.new,
+      env_name: ENV["CONNECT_VBMS_ENV"],
+      css_id: Rails.deploy_env?(:prod) ? "CSFLOW" : "CASEFLOW1",
```
shanear: We should use `User.system_user` to figure out the user names here

app/services/external_api/vbms_service.rb
```diff
@@ -173,6 +176,20 @@ def self.vbms_client_with_user
...
+      station_id: "283",
```
shanear: Same. We should use `User.system_user` to determine this station id

app/services/external_api/vbms_service.rb
```diff
@@ -173,6 +176,20 @@ def self.vbms_client_with_user
...
+      use_forward_proxy: FeatureToggle.enabled?(:vbms_forward_proxy)
+    )
+
+    FeatureToggle.enabled?(:vbms_include_user) ? @vbms_client_with_system_user : @vbms_client
```
shanear: do we need this toggle when we are already using `send_userid:` to toggle it?

app/models/end_product_establishment.rb
```diff
@@ -238,9 +238,17 @@ def veteran
...
+      found_intake = Intake.find_by(detail_id: source_id, detail_type: source_type)
+      # if no intake is found such as for DTAs, use the system user
+      user = found_intake ? User.find_by(id: found_intake.user_id) : User.system_user
+    end
+
     VBMSService.establish_claim!(
       claim_hash: end_product.to_vbms_hash,
-      veteran_hash: veteran.to_vbms_hash
+      veteran_hash: veteran.to_vbms_hash,
+      user: user
```
jiahuang: I thought about putting the logic to decide the user in vbms service. Decided against it since I wanted that layer to stay thin. I'm open to changing it though.

app/services/external_api/vbms_service.rb
```diff
@@ -124,21 +132,23 @@ def self.create_contentions!(veteran_file_number:, claim_id:, contention_descrip
       claim_id: claim_id,
       contentions: contention_descriptions,
       special_issues: special_issues,
-      v5: FeatureToggle.enabled?(:claims_service_v5)
+      v5: FeatureToggle.enabled?(:claims_service_v5),
+      send_userid: FeatureToggle.enabled?(:vbms_include_user)
     )
 
-    send_and_log_request(claim_id, request)
+    send_and_log_request(claim_id, request, vbms_client_with_current_user)
```
jiahuang: Create contentions current uses the current user. Should this also be the user that did the intake for hlr and sc?

app/models/end_product.rb
```diff
@@ -158,6 +158,10 @@ def contentions
     @contentions ||= claim_id ? VBMSService.fetch_contentions(claim_id: claim_id) : nil
   end
 
+  def claim_review?
+    (DTA_CODES.keys + AMA_CODES.keys).include?(claim_type_code)
```
shanear: I think there are future codes to come here. Worry that we won't be able to keep this method up to date as we add those. I don't have a good solution, just pointing that out.

app/models/end_product_establishment.rb
```diff
@@ -238,9 +238,17 @@ def veteran
   end
 
   def establish_claim_in_vbms(end_product)
+    user = nil
+    if end_product.claim_review?
+      found_intake = Intake.find_by(detail_id: source_id, detail_type: source_type)
+      # if no intake is found such as for DTAs, use the system user
+      user = found_intake ? User.find_by(id: found_intake.user_id) : User.system_user
+    end
```
shanear: I think this method should take a user, like so:
```
def establish_claim_in_vbms(end_product, user)
```

Which would allow us to skip a lot of this nitty gritty logic, by using our polymorphism to pass the correct user in.

I'm thinking we should add `user_id` as a value on `end_product_establishments`, this would allow us to set the user when we create each individual end_product_establishment (like here: https://github.com/department-of-veterans-affairs/caseflow/blob/416bdb9c9100e6d7f5f96a8a4316bd8a80bed5af/app/models/higher_level_review.rb#L131)

app/models/end_product_establishment.rb
shanear: I think we want to go even further out. Read my comment above.

app/services/external_api/vbms_service.rb
shanear: > Should this also be the user that did the intake for hlr and sc?

yes, if we save the user on the EndProductEstablishment, it shouldn't be too hard to pass it into this function

app/services/external_api/vbms_service.rb
```diff
@@ -93,16 +93,24 @@ def self.clean_document(location)
...
       claim_hash,
-      v5: FeatureToggle.enabled?(:claims_service_v5)
+      v5: FeatureToggle.enabled?(:claims_service_v5),
+      send_userid: FeatureToggle.enabled?(:vbms_include_user)
     )
 
-    send_and_log_request(veteran_hash[:file_number], request, FeatureToggle.enabled?(:vbms_include_user))
+    # get the correct client for the request
+    client = @vbms_client
+    if !user.nil?
```
shanear: Should we even allow this if there is no user? I think we always want to be using `vbms_client_with_user`

app/services/external_api/vbms_service.rb
```diff
@@ -175,6 +195,16 @@ def self.vbms_client_with_user
     )
   end
 
+  def self.vbms_client_with_system_user
```
shanear: I think we can probably factor all of these down to just use `vbms_client_with_user(user)`. Doesn't seem to hard to pass the `User.system_user` as the arg to that function for `vbms_client_with_system_user`.

I realize that removes the memoization, but that isn't that important and we can worry about the memoization based on user later.

lib/fakes/vbms_service.rb
```diff
@@ -110,7 +110,8 @@ def self.clean_document(_location)
     # noop
   end
 
-  def self.establish_claim!(claim_hash:, veteran_hash:)
+  # rubocop:disable UnusedMethodArgument
```
shanear: instead of the rubocop workaround, why not just log the username or something? That might be helpful.

app/models/claim_review.rb
```diff
@@ -87,6 +87,11 @@ def informal_conference?
     false
   end
 
+  def intake_processed_by
+    found_intake = Intake.find_by(detail_id: id, detail_type: self.class.name)
```
shanear: this could be a rails `has_one` relationship

app/models/supplemental_claim.rb
```diff
@@ -60,6 +60,10 @@ def issue_code(rated)
 
   private
 
+  def end_product_created_by
+    is_dta_error ? User.system_user : intake_processed_by
```
shanear: `is_dta_error` could be `is_dta_error?` to signify it is a boolean

app/services/external_api/vbms_service.rb
```diff
@@ -116,29 +118,33 @@ def self.fetch_contentions(claim_id:)
...
 
     request = VBMS::Requests::RemoveContention.new(
       contention: contention,
-      v5: FeatureToggle.enabled?(:claims_service_v5)
+      v5: FeatureToggle.enabled?(:claims_service_v5),
+      send_userid: FeatureToggle.enabled?(:vbms_include_user)
     )
 
-    send_and_log_request(contention.claim_id, request)
+    client = FeatureToggle.enabled?(:vbms_include_user) ? vbms_client_with_system_user : @vbms_client
```
shanear: Shouldn't we just always use `vbms_client_with_system_user` and let the feature toggle on `send_userid` do the work for us? (like here: https://github.com/department-of-veterans-affairs/caseflow/pull/7476/files#diff-bac97804e7edd3f5537c3dc2d53cd1b9R103)

app/services/external_api/vbms_service.rb
shanear: bump

app/models/supplemental_claim.rb
jiahuang: is_dta_error is a field from the db. I think I'd have to define a function `is_dta_error?` that returns `is_dta_error` to get the `?`. Donno if that's worth it.

app/models/ramp_review.rb
```diff
@@ -1,5 +1,6 @@
 class RampReview < ApplicationRecord
   belongs_to :user
+  has_one :intake, ->(review) { where(detail_type: review.class.name) }, :foreign_key => "detail_id"
```
jiahuang: all ramp reviews have 1 intake?

PR #7477, opened 2018-10-19
remove eps that no longer have ratings

jiahuang:
Connects #6214

This PR does not handle the edge case of re-adding a rated/non-rated issue that has been canceled. That will be addressed in a follow up PR


app/models/end_product_establishment.rb
```diff
@@ -142,6 +142,16 @@ def preexisting_end_product
     @preexisting_end_product ||= veteran.end_products.find { |ep| end_product_to_establish.matches?(ep) }
   end
 
+  def cancel!
+    transaction do
```
jiahuang: So cancel is only called from cancel_unused_end_products which is only called from RequestIssueUpdate.perform! which is already wrapped in a transaction. Should I still put this in a transaction? Seems like at worst it's a no-op, and it kinda future proofs this method if we ever need to call it elsewhere.

app/models/claim_review.rb
```diff
@@ -44,6 +44,18 @@ def create_issues!(new_issues)
     end
   end
 
+  def cancel_unused_end_products!(removed_issues)
+    # get all potential removed issue eps
+    potential_end_products_to_remove = removed_issues.map(&:end_product_establishment_id).uniq
+    removed_issue_ids = removed_issues.map(&:id)
+    potential_end_products_to_remove.each do |end_product_establishment_id|
```
jiahuang: An alternative is to skip this for loop and make a bigger sql statement which returns end product establishment ids to delete if they have no request issues. This way seemed easier to read & debug.

app/models/request_issues_update.rb
```diff
@@ -18,6 +18,7 @@ def perform!
       review.create_issues!(new_issues)
       strip_removed_issues!
       review.mark_rated_request_issues_to_reassociate!
+      review.cancel_unused_end_products!(removed_issues)
```
shanear: we don't want to do this synchronously, since it's a remote call that's atomic with other remote calls

app/models/claim_review.rb
shanear: I think we probably want to do this in a similar place to where we are removing the contention. I think logic should go something like:

```
request_issue.end_product_establishment.remove_contention!(request_issue)
request_issue.end_product_establishment.cancel! if (no more active contentions on the epe)
```

shanear: see: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/request_issues_update.rb#L44

app/models/end_product_establishment.rb
shanear: yeah i like this method a lot.

app/models/end_product_establishment.rb
```diff
@@ -142,6 +142,16 @@ def preexisting_end_product
...
+      # delete end product in bgs & set sync status to canceled
+      BGSService.new.cancel_end_product(:veteran_file_number, :code, :modifier)
+      update!(synced_status: CANCELED_STATUS)
+    end
+  rescue StandardError => e
+    raise BGSSyncError.new(e, self)
```
shanear: not sure this should be re-raised as a sync error though. Since we aren't really syncing anything, we are canceling the EP.

spec/feature/intake/edit_spec.rb
```diff
@@ -148,6 +148,80 @@ def check_row(label, text)
         higher_level_review.process_end_product_establishments!
       end
 
+      scenario "creates new ep for new rating types" do
```
shanear: I don't think we need to test the new EP functionality in a feature test, a unit test on `RequestIssuesUpdate` would suffice i think.

However, we may end up changing the confirmation page based on what happens, so may want to keep this one around.

app/models/end_product_establishment.rb
```diff
@@ -142,6 +142,16 @@ def preexisting_end_product
     @preexisting_end_product ||= veteran.end_products.find { |ep| end_product_to_establish.matches?(ep) }
   end
 
+  def cancel_unused_end_product!(removed_issues)
+    removed_issue_ids = removed_issues.map(&:id)
+    active_request_issues_ids = request_issues.select { |request_issue| request_issue.removed_at.nil? }.map(&:id)
```
leikkisa: It seems like this could also just check if there are no more active request issues instead of seeing if active issues - removed issues is empty.  That would mean not having to pass the removed issues around.

jiahuang: oh good point. I originally had this getting called inside the transaction that was updating the removed request issues, so the issues weren't actually "removed" yet, but now that it's moved to the async job I can clean this up!

app/models/request_issues_update.rb
```diff
@@ -44,6 +44,11 @@ def process_end_product_establishments!
       request_issue.end_product_establishment.remove_contention!(request_issue)
     end
 
+    potential_end_products_to_remove = removed_issues.map(&:end_product_establishment).uniq
+    potential_end_products_to_remove.each do |end_product_establishment|
```
pkarman: sometimes you'll see this written as a single line:

```ruby
potential_end_products_to_remove.each(&:cancel_unused_end_product!)
```

spec/models/request_issues_update_spec.rb
```diff
@@ -291,6 +307,45 @@
...
+          source: review,
+          code: "030HLRNR"
+        )
+
+        nonrating_request_issue_contention = Generators::Contention.build(
+          claim_id: nonrating_end_product_establishment.reference_id,
+          text: "Unrated issue"
+        )
+
+        RequestIssue.create!(
```
pkarman: I think the preference has been to use the FactoryBot `create()` method instead of instantiating directly from the AR model in tests.

app/models/end_product_establishment.rb
```diff
@@ -142,6 +142,15 @@ def preexisting_end_product
     @preexisting_end_product ||= veteran.end_products.find { |ep| end_product_to_establish.matches?(ep) }
   end
 
+  def cancel_unused_end_product!
+    active_request_issues = request_issues.select { |request_issue| request_issue.removed_at.nil? }
+
+    # if all active request issues were removed, cancel
```
pkarman: this comment seems a little redundant

app/services/external_api/bgs_service.rb
```diff
@@ -34,6 +34,20 @@ def get_end_products(vbms_id)
       end
   end
 
+  def cancel_end_product(vbms_id, end_product_code, end_product_modifier)
```
shanear: can we call `vbms_id`, `veteran_file_number`? this is the up to date terminology

spec/models/request_issues_update_spec.rb
```diff
@@ -291,6 +307,45 @@
...
+        expect(rated_end_product_establishment.synced_status).to eq(nil)
+      end
+
+      it "cancels end products with no request issues" do
+        nonrating_end_product_establishment = create(
+          :end_product_establishment,
+          veteran_file_number: veteran.file_number,
+          source: review,
+          code: "030HLRNR"
+        )
```
shanear: These could be `let` statements

Issue #7463, opened 2018-10-18
Display "returned from QR" tasks in judge queues


lowellrex:
As part of [the QR queue rollout effort](https://github.com/department-of-veterans-affairs/caseflow/issues/7290) we anticipate some cases returning to the signing judges for corrections. We need to ensure that these tasks show up in the judges' individual task queues after they are returned to them by a quality reviewer.

## Acceptance criteria
* Cases returned to the singing judge from QR display in judges' individual task queues.

Issue #7462, opened 2018-10-18
Add custom actions for individual's QR tasks


lowellrex:
As part of [the QR queue rollout effort](https://github.com/department-of-veterans-affairs/caseflow/issues/7290) we need to allow individual quality reviewers to do two actions:
1. Return the case to the judge who signed the decision (specific to the QR process) and
2. Mark the task complete (maybe called something like "ready for dispatch" eventually)

Issue #7459, opened 2018-10-18
Ensure individual QR tasks show up in individual attorney's queues


lowellrex:
As part of [the QR queue rollout effort](https://github.com/department-of-veterans-affairs/caseflow/issues/7290) we expect QR team tasks will spawn child tasks assigned to individual attorneys. We need to ensure that QR tasks show up in their individual task queues after they are distributed from the QR team queue. Related slack thread: https://dsva.slack.com/archives/C6E41RE92/p1539895128000200

## Acceptance criteria
* Individual QR tasks display in individual task queues.

PR #7484, opened 2018-10-20
Adding QR return to judge flow

mdbenjam:
Resolves: #7463 
Resolves: #7462 
Resolves: #7459  

### Description
This completes the QR flow. When QR reviews a case they can now see it in their queues, send it back to a judge if need be, and that judge can see it in their queue.


app/models/queues/attorney_queue.rb
```diff
@@ -18,7 +18,8 @@ def tasks
       end
       result
     end
-    ama_tasks = AttorneyTask.where.not(status: Constants.TASK_STATUSES.completed).where(assigned_to: user)
+    ama_tasks = Task.where.not(status: Constants.TASK_STATUSES.completed)
```
mdbenjam: For now limiting this expansion to QualityReview tasks since we know QR people are attorneys.

app/models/tasks/quality_review_task.rb
```diff
@@ -1,4 +1,21 @@
...
+  def available_actions(user)
+    return super if assigned_to != user
+
+    [
+      {
+        label: COPY::ACTION_READY_FOR_DISPATCH,
+        value: "modal/mark_task_complete"
+      },
+      Constants.TASK_ACTIONS.RETURN_TO_JUDGE.to_h.merge(
+        data: {
```
mdbenjam: New technique to pass needed data to a task. Thoughts?

app/views/organizations/show.html.erb
```diff
@@ -4,6 +4,7 @@
     userId: current_user.id,
     userRole: (current_user.vacols_roles.first || "").capitalize,
     userCssId: current_user.css_id,
+    organizationIds: current_user.organizations.pluck(:id),
```
mdbenjam: This is a more robust way to determine organization id(s).

client/app/queue/AssignToView.jsx
```diff
@@ -82,7 +84,7 @@ class AssignToView extends React.Component<Props, ViewState> {
     const payload = {
       data: {
         tasks: [{
-          type: 'GenericTask',
+          type: this.props.isTeamAssign ? 'GenericTask' : task.type,
```
mdbenjam: We want to propagate the parent tasks' type if assigning to a person on the same team.

client/app/queue/OrganizationQueueLoadingScreen.jsx
```diff
@@ -30,9 +28,6 @@ class OrganizationQueueLoadingScreen extends React.PureComponent<Props> {
     (response) => {
       const { tasks: { data: tasks } } = JSON.parse(response.text);
 
-      if (tasks[0]) {
```
mdbenjam: This isn't a robust way to set organization ids.

client/app/queue/selectors.js
```diff
@@ -138,7 +138,7 @@ export const completeTasksByAssigneeCssIdSelector = createSelector(
 export const organizationTasksByAssigneeIdSelector = createSelector(
   [getTasksForAppeal, getOrganizationId],
   (tasks: Tasks, id: Number) =>
-    _.filter(tasks, (task) => task.assignedTo.id === id && ['Organization', 'Vso'].includes(task.assignedTo.type))
+    _.filter(tasks, (task) => task.assignedTo.id === id && 'User' !== task.assignedTo.type)
```
mdbenjam: Let's be more generic here.

client/app/queue/AssignToCustomUser.jsx
```diff
@@ -0,0 +1,137 @@
+// @flow
```
mdbenjam: Maybe this should just be incorporated into the assign to view modal that already exists. And we should just auto select the judge from the dropdown.

app/models/queues/attorney_queue.rb
```diff
@@ -18,7 +18,8 @@ def tasks
...
+                    .where(assigned_to: user, type: ["AttorneyTask", "QualityReviewTask"])
```
lowellrex: What task types are we excluding by doing this?

lowellrex: Also, can we make these `AttorneyTask.name` and `QualityReviewTask.name`?

app/models/tasks/generic_task.rb
```diff
@@ -77,7 +77,8 @@ def create_child_task(parent, current_user, params)
 
       parent.update!(status: :on_hold)
 
-      GenericTask.create!(
+      Task.create!(
+        type: name,
```
lowellrex: Curious about what motivated this change?

app/models/tasks/judge_task.rb
```diff
@@ -23,8 +23,10 @@ def available_actions(user)
     end
   end
 
-  def self.create(params)
-    super(params.merge(action: "assign"))
+  def self.create(params_array)
+    [params_array].flatten.map do |params|
+      super(params.merge(action: "assign"))
+    end
```
lowellrex: Why does this need to accept an array of parameters? Won't the parent class handle looping across the array of parameters that comes in? Or am I missing something (perhaps we create `JudgeTask`s outside of the `TasksController`?)?

https://github.com/department-of-veterans-affairs/caseflow/blob/916b4cae44aea069e7019c5a226c01eac331bb0b/app/models/task.rb#L45

app/models/tasks/generic_task.rb
lowellrex: This allows the subclasses of `GenericTask` to create child tasks of the same type as the parent task. Smart!

app/models/tasks/quality_review_task.rb
```diff
@@ -1,4 +1,21 @@
...
+          user: completing_judge,
+          type: "JudgeTask"
```
lowellrex: `JudgeTask.name`!

app/models/tasks/quality_review_task.rb
lowellrex: Nice and lightweight, works for me!

app/models/tasks/quality_review_task.rb
```diff
@@ -7,4 +24,10 @@ def mark_as_complete!
     BvaDispatchTask.create_and_assign(root_task)
     super
   end
+
+  private
+
+  def completing_judge
+    root_task.children.find { |task| task.type == "JudgeTask" }.assigned_to
```
lowellrex: `JudgeTask.name` for compile time warnings!

app/models/tasks/quality_review_task.rb
```diff
@@ -1,4 +1,21 @@
 class QualityReviewTask < GenericTask
+  def available_actions(user)
+    return super if assigned_to != user
+
+    [
+      {
+        label: COPY::ACTION_READY_FOR_DISPATCH,
+        value: "modal/mark_task_complete"
```
lowellrex: It appears like we only use `COPY::ACTION_READY_FOR_DISPATCH` here so perhaps we could move this whole hash into TASK_ACTIONS.json?

client/app/queue/AssignToCustomUser.jsx
lowellrex: In addition to moving this logic into the `AssignToView` component, would it be possible to move the options that we select in the dropdown into the TASK_ACTIONS.json file with some function in the base `Task` model that executes methods if the task actions have associated methods. Perhaps like the following:
```ruby
# available_actions() returns an array of options from selected by the subclass
# from TASK_ACTIONS that looks something like:
# [ { "label": "Assign to person", "value": "modal/assign_to_person", "func": "assignable_users" }, ... ]
def available_actions_unwrapper(user)
  available_actions.each do |a|
    { label: a.label, value: a.value, options: a.func ? send(a.func) : nil }
  end
end
```

This way we could avoid longer passing `assignableOrganizations` and `assignableUsers` properties of the task, instead relying on the options made available as an options property of the action we are selecting.

In this particular case the return to judge task action could look something like:
```ruby
# { "label": "Return to judge", "value": "modal/return_to_judge", "func": "assigning_judge" }
def assigning_judge
  judge = root_task.children.find { |task| task.type == "JudgeTask" }.assigned_to
  [{label: judge.full_name, value: judge.id}]
end
```

app/models/tasks/judge_task.rb
mdbenjam: I wrote this before you made that change, I think you're right that it's safe to change back to a normal one creation thing. Sidenote, I wish we weren't overriding the create method. I think that's bad rails practice.

client/app/queue/selectors.js
```diff
@@ -142,7 +142,7 @@ export const actionableTasksForAppeal = createSelector(
 export const organizationTasksByAssigneeIdSelector = createSelector(
   [getTasksForAppeal, getOrganizationId],
   (tasks: Tasks, id: Number) =>
-    _.filter(tasks, (task) => task.assignedTo.id === id && ['Organization', 'Vso'].includes(task.assignedTo.type))
+    _.filter(tasks, (task) => task.assignedTo.id === id && task.assignedTo.type !== 'User')
```
lowellrex: I think we explicitly included `Organization` and `Vso` to avoid including the `Bva` organization, so I think we might want to do that here now as well.

Issue #7397, opened 2018-10-15
AMA Intake/Edit | Error handling for save


leikkisa:
When saving changes on the Add/Remove Issues screen of the edit flow, there are a couple reasons why the changes may be invalid.

## Acceptance criteria
- When a user saves an AMA review edit or intake, if there is an error with that submission, caseflow should display an error message in an alert.

## Technical notes
- The backend for these validations already exists here: (https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/request_issues_update.rb#L94)
- The component to show the correct error message also exists(https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/intakeEdit/components/RequestIssuesUpdateErrorAlert.jsx#L4)

## Copy
Use the same copy we were using before:
found here: https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/intakeEdit/components/RequestIssuesUpdateErrorAlert.jsx#L4

## Mocks
We should show an error alert like this at the top of the page, below the "Add/Remove Issues" title.
<img width="823" alt="screen shot 2018-10-16 at 2 43 11 pm" src="https://user-images.githubusercontent.com/1596259/47040224-bf7e0d00-d153-11e8-9741-51fa7b444188.png">

shanear:
![screen shot 2018-10-26 at 2 50 20 pm](https://user-images.githubusercontent.com/1596259/47586773-1394a880-d92f-11e8-8ef9-5e9882de25c4.png)
![screen shot 2018-10-26 at 2 54 36 pm](https://user-images.githubusercontent.com/1596259/47586775-142d3f00-d92f-11e8-8d0e-7f9188993d91.png)
Got there 👍👍

PR #7498, opened 2018-10-22
Add Complete Intake Error Alert

leikkisa:
connects #7397

### DIFF
--- a/client/app/intake/pages/addIssues.jsx
+++ b/client/app/intake/pages/addIssues.jsx
@@ -9,6 +9,7 @@ import NonRatedIssueModal from '../components/NonRatedIssueModal';
 import RemoveIssueModal from '../components/RemoveIssueModal';
 import UnidentifiedIssuesModal from '../components/UnidentifiedIssuesModal';
 import Button from '../../components/Button';
+import CompleteIntakeErrorAlert from '../components/CompleteIntakeErrorAlert';
 import RequestIssuesUpdateErrorAlert from '../../intakeEdit/components/RequestIssuesUpdateErrorAlert';
 import { REQUEST_STATE, FORM_TYPES, PAGE_PATHS } from '../constants';
 import INELIGIBLE_REQUEST_ISSUES from '../../../constants/INELIGIBLE_REQUEST_ISSUES.json';
@@ -46,6 +47,7 @@ export class AddIssuesPage extends React.Component {
 
   render() {
     const {
+      intakeOrEdit,
       intakeForms,
       formType,
       veteran,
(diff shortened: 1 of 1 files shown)


client/app/intake/pages/addIssues.jsx
```diff
@@ -109,6 +111,27 @@ export class AddIssuesPage extends React.Component {
...
+            completeIntakeErrorCode={completeIntakeErrorCode}
+            completeIntakeErrorData={completeIntakeErrorData} />;
+      } else if (intakeOrEdit === "edit") {
+        const requestIssuesUpdate = intakeData.requestStatus.requestIssuesUpdate;
+        const responseErrorCode = intakeData.responseErrorCode;
+
+        return requestIssuesUpdate === REQUEST_STATE.FAILED &&
+          <RequestIssuesUpdateErrorAlert responseErrorCode={responseErrorCode} />
+      }
+    };
```
shanear: I think we could probably factor `RequestIssuesUpdateErrorAlert` and `CompleteIntakeErrorAlert` into the same class. Then we could reuse `default` and `request_issues_data_empty` messages.

Maybe something like this?
`<SubmitIssuesErrorAlert errorCode={...} errrorData={...}>`

Also we should probably add a `request_issues_data_empty` type error for intakes too.

client/app/intake/pages/addIssues.jsx
```diff
@@ -109,6 +111,27 @@ export class AddIssuesPage extends React.Component {
       </div>;
     };
 
+    const failedRequestAlert = () => {
+      if (intakeOrEdit === "intake") {
```
pkarman: +1 to reusing the component.

my only hesitancy with this approach is using string comparison as a boolean toggle seems to end up being brittle, IME. Alternatives I've seen include:

* use an explicit boolean:

```
isIntake: true
isEdit: false
```

or a constant
```
const INTAKE = 1;
const EDIT = 2;
intakeOrEdit = INTAKE;
if (intakeOrEdit === INTAKE) {
// ...
}
```

shanear: not sure we need this variable. I think we can infer whether we are on edit or intake from the presence of `intakeData.requestStatus.requestIssuesUpdate` or `intakeData.requestStatus.completeIntake`

spec/feature/intake/supplemental_claim_spec.rb
```diff
@@ -575,6 +575,24 @@ def check_row(label, text)
...
+
+      ## Validate error message when complete intake fails
+      expect_any_instance_of(SupplementalClaimIntake).to receive(:complete!).and_raise("A random error. Oh no!")
+
+      safe_click "#button-finish-intake"
+
+      expect(page).to have_content("Something went wrong")
+      expect(page).to have_current_path("/intake/add_issues")
+    end
+
```
shanear: Solid tests :100:

PR #7499, opened 2018-10-23
Untimely issues flagged in UI client

pkarman:
connects #7300 

follow on to #7473 

### Description
Moves `timely_rating?` to `AMAReview` class. Injects the `timely` flag to rating issues via new method `serialized_ratings`.

Adds UI element to highlight untimely issues.

screen shot 2018-10-22 at 5 39 47 pm

### Acceptance Criteria
- [x] Code compiles correctly


app/models/ama_review.rb
```diff
@@ -14,14 +14,29 @@ class AmaReview < ApplicationRecord
...
+    return unless receipt_date
+
+    cached_serialized_ratings.each do |rating|
+      rating[:issues].each do |rating_issue|
+        rating_issue[:timely] = timely_rating?(Date.parse(rating_issue[:promulgation_date].to_s))
+      end
+    end
+  end
+
+  def timely_rating?(promulgation_date)
```
shanear: this could probably be a private method

app/models/request_issue.rb
```diff
@@ -86,12 +86,38 @@ def ui_hash
...
     existing_request_issue = self.class.find_active_by_reference_id(rating_issue_reference_id)
     if existing_request_issue
       self.ineligible_reason = :duplicate_of_issue_in_active_review
     end
-    self
+  end
+
+  def check_for_untimely!
+    return if review_request && review_request.is_a?(SupplementalClaim)
+    rating_issue = original_rating_issue
```
shanear: could `original_rating_issue` be memoized so we don't have to use assignment when we use it?

app/models/request_issue.rb
```diff
@@ -86,12 +86,38 @@ def ui_hash
...
+    if rating_issue && review_request && !review_request.timely_rating?(rating_issue[:promulgation_date])
+      self.ineligible_reason = :untimely
+    end
+  end
+
+  def original_rating_issue
```
shanear: had a little difficulty figuring out what `original_rating_issue` means. It looks like it's fetching the rating_issue hash from the serialized ratings that matches the rating_issue reference_id?

I feel like there might be a cleaner way to do this (maybe via a method that converts the serialized rating hash back into an array of ratings?). If not might be best to explain what's happening here by comment

app/models/ama_review.rb
pkarman: it gets called from `RequestIssue` so alas, no.

app/models/request_issue.rb
pkarman: this is the rabbit hole I went down earlier. 🐰 

Ideally the `RatingIssue` would be saved in the db and we could simply look it up by `rating_issue_reference_id` but we currently do not save `RatingIssue` until they are assigned contentions (later in the flow). So yes, this is looking up the matching `RatingIssue` from the cache that was just used to populate the UI client.

I added some comments and renamed things to (hopefully) clarify.

Issue #7333, opened 2018-10-10
Legacy Opt-in | Review page | add opt-in question


suzchap:
## Story
[Mural showing all opt-in screens](https://app.mural.co/t/workqueue2001/m/workqueue2001/1538676663322/098da798af658862105ea8d4d5aeb03fa7aaf99a)

When a user has a legacy appeal, they will be allowed to "opt-in" to move one or more of those issues into an AMA lane. The AMA paper forms will have a question on them for the applicant to acknowledge that by checking that box, they understand that any legacy issues listed on that form will be moved out of the legacy lane and into the new AMA lane.

This ticket is focused on adding the opt-in question to the review page. 

## Technical notes
- This should be behind a feature flag

## Acceptance criteria
- When creating an intake for any AMA review, add question in the mock below to the bottom of the review page
- When completing an intake for any AMA review, save the answer to the new question to the database

## Notes
- Out of scope: doing anything with the answer to this question. We'll use the answer in the add issues flow in the future

## Copy
Did they agree to withdraw their issues from the legacy system? 

## Mocks
<img width="545" alt="screen shot 2018-10-10 at 2 08 54 pm" src="https://user-images.githubusercontent.com/11219417/46756711-17b99880-cc96-11e8-9963-46ed4d683c6f.png">

## How to Validate
(no need, Shane's got this :muscle:)

nicholasholtz:
Does this question show up for all AMA review intakes, or does this show up specifically because of something the user has done to indicate that this will be an opt-in?

shanear:
need the feature flag set in UAT to test this

shanear:
![screen shot 2018-10-31 at 1 27 21 pm](https://user-images.githubusercontent.com/1596259/47806717-c8a0d980-dd10-11e8-85c2-967e0a30b5e2.png)

need to validate in the db on the supplemental claim on 123458319 in UAT

shanear:
```
#<SupplementalClaim id: 99, veteran_file_number: "123458319", receipt_date: "2018-10-22", establishment_submitted_at: "2018-10-31 17:51:34", establishment_processed_at: "2018-10-31 20:52:20", benefit_type: "compensation", is_dta_error: nil, establishment_attempted_at: "2018-10-31 20:52:08", establishment_error: nil, legacy_opt_in_approved: true>
```

nailed it 💯

PR #7513, opened 2018-10-23
Legacy opt-in option for AMA Intakes

leikkisa:
connects #7333 

- Adds a Legacy opt-in option to all AMA intakes
- Saves the selection in the database
- Some error handling and ama review refactoring
- Puts Legacy opt-in behind a feature toggle


leikkisa:
screen shot 2018-10-23 at 4 45 51 pm

app/models/ama_intake.rb
```diff
@@ -0,0 +1,22 @@
+class AmaIntake < Intake
```
shanear: Can we call this `DecisionReviewIntake`? 🙂

shanear: P.S. love this refactor

db/migrate/20181023204155_add_legacy_opt_in_to_ama_intake.rb
```diff
@@ -0,0 +1,7 @@
+class AddLegacyOptInToAmaIntake < ActiveRecord::Migration[5.1]
+  def change
+    add_column :higher_level_reviews, :legacy_opt_in, :boolean
+    add_column :supplemental_claims, :legacy_opt_in, :boolean
+    add_column :appeals, :legacy_opt_in, :boolean
```
shanear: I think this value is probably more accurately named something like `legacy_opt_in_approved` or `opt_in_approved` or `opt_in_aggrement` or something like that, because just because they approve, doesn't mean they actually opt in anything

client/app/intake/components/LegacyOptIn.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import RadioField from '../../components/RadioField';
+import { BOOLEAN_RADIO_OPTIONS } from '../constants';
+
+export default class LegacyOptIn extends React.PureComponent {
```
shanear: Shouldn't this be called `LegacyOptInQuestion` or `LegacyOptInRadio` to more accurately say what it is?

app/models/supplemental_claim.rb
```diff
@@ -11,23 +14,9 @@ class SupplementalClaim < ClaimReview
...
-      benefitType: benefit_type,
-      claimant: claimant_participant_id,
-      claimantNotVeteran: claimant_not_veteran,
-      payeeCode: payee_code,
-      ratings: serialized_ratings,
-      requestIssues: request_issues.map(&:ui_hash)
-    }
+    super.merge(
+      formType: "supplemental_claim"
+    )
```
shanear: beautiful 👍

spec/feature/intake/supplemental_claim_spec.rb
```diff
@@ -353,7 +351,8 @@ def start_supplemental_claim(test_veteran, is_comp: true)
     supplemental_claim = SupplementalClaim.create!(
       veteran_file_number: test_veteran.file_number,
       receipt_date: 2.days.ago,
-      benefit_type: is_comp ? "compensation" : "education"
+      benefit_type: is_comp ? "compensation" : "education",
+      legacy_opt_in: false
```
shanear: Solid tests :100:

db/migrate/20181023204155_add_legacy_opt_in_to_ama_intake.rb
leikkisa: Good point! I like "approved"

app/models/ama_intake.rb
leikkisa: Lol, yeah

Issue #7072, opened 2018-09-20
Video and CO Release | Epic 2  Case Details - Assign Hearing


Chingujo:
As a hearing coordinator, I need to be able to assign a hearing to a specific slot in the schedule. 

### Acceptance Criteria
- [ ] Verify that when a user clicks the Assign Hearings in the action drop down the following populates for central office: 
   - Hearing location: Washington D.C. Change
       - If the user change it shows the dropdown of RO's. If selected, it will automatically turn the date
         field into a dropdown matching that RO's non availability days
   - Date: Defaults to the one selected in the table and a link text for change 
       - Clicking change renders a dropdown that shows exactly the same availability as the dates on the table 
        - The dates are in the following format: Day Date (2 open slots)
    - Time: 9:00 am or 1:00 pm options
- [ ] Verify the same exact behavior for Video hearings with the following exceptions: 
    - Hearing Location: Defaults to the selected RO 
     - Time: 8:00 am and 12:00 pm
- [ ] Verify that there is a button called Assign and when clicked confirms the veteran into the schedule

mdbenjam:
One pattern we've been using is to make the actions dropdown options open up either a modal or new page, instead of putting the options right in the task area of case details.

Issue #7506, opened 2018-10-23
Use new organization membership in application


lowellrex:
In #7427 we implemented a new method for recognizing which folks are members of which teams. We should now use that instead of relying on constants for pilot users, and VACOLS (Board employees) and BGS (VSOs) for team membership.

It remains an open question about whether or not we want to move [the attorney-judge team mapping](https://github.com/department-of-veterans-affairs/caseflow/blob/master/lib/constants/attorney_judge_teams.rb) into this model as well.

## Acceptance criteria
* [x] Replace `Organization`s reliance on the `VACOLS.staff` table with `OrganizationsUser` model
* [x] Remove `staff_field_for_organizations` table after the above step is complete
* [x] Move pilot membership into the `OrganizationsUser` model
* [x] Replace VSOs' reliance on BGS with the `OrganizationsUser` model (by way of #7974)

PR #7514, opened 2018-10-23
Case Details - Base Page

### LINKED ISSUE #6874: Video and CO Release | Epic 2  Case Details - Base Page
As a hearing coordinator, I want to be able to see detailed information about a Veteran as part of the hearings review process, such as checking the form 9. 

### Acceptance Criteria
- [ ] Verify that clicking on a Veteran Name in the table takes you to the case details page
- [ ] Verify that the header of this page is Veteran Name | View Veterans Document | View assign hearings (see mock)
- [ ] Verify the following information in the snap shot area of the Case Details (see mock)
  - About the Case: 
    - Type
    - Docket 
    - Hearing Type
  - Veteran Address
  - Action with a dropdown called Select: 
    - This Action bar has three states depending on how the user enters this case detail view. 
    - Via the table, the dropdown should show "Assign Hearing" 
    - If a veteran has been scheduled, then the action should show "View Hearing Date" 
- [ ] Verify that the link next to the Veteran ID is "Assign Hearings"
- [ ]  Verify that the rest of the Case Details shows the way it is in Queue (see mock)
   - [ ] Verify in the hearings section the following logic: 
      - All prior hearings with a disposition are shown 
      - When a hearing is scheduled it shows with a null disposition to signify that it is the most current hearing

### Notes
- [ ] How are queue handling user permissions? Should we mimic this?

### Mocks
**Default**
assign hearing co default 1

** Dropdown Expanded**
assign hearing dropdown

** Hearings area in case details**
assign hearing co default 1

## Notes
Please see

oscarAtNava:
Resolves #6874 

### Description
Implement flow between Assign Hearings and Case Details. Hearings coordinator can click on veteran link in Assign Hearings to be taken to Case Details to review the appeal and confirm the hearing schedule assignment.

In this PR when the coordinator selects the Assign Hearing action in Case Details the task is completed in the back-end. The modal functionality will be part of the PR associated with issue #7072 .

### Acceptance Criteria
- [ ] Code compiles correctly
- Test case for ScheduleHearingTask passes.

### Testing Plan
1. Go to Assign Hearings and click on new link for veteran.
2. Confirm the CaseDetails page is rendered and that it has a task in the top section.
3. Confirm that when you select the Assign Hearing action the Mark Complete task modal displays.
4. Click on Complete Task and confirm the database task row is set to completed.


app/models/organizations/hearings_management.rb
```diff
@@ -0,0 +1,9 @@
+class HearingsManagement < Organization
+  def self.singleton
+    HearingsManagement.first || HearingsManagement.create(name: "Hearings Management")
+  end
+
+  def user_has_access?(_user)
+    true
```
lowellrex: Am I misreading this, or will everybody be a member of the hearings management team?

oscarAtNava: This is a temporary fix until I am able to figure out why the function in the parent class fails all the time regardless of how I set up my user.

lowellrex: I think we can start taking advantage of the organization-team relationship Mark created in #7454 by adding rows for folks who are on the Hearings Management team to the `organizations_users` table and overriding the members method in this class:
```ruby
def members
  organizations_users
end
```

There is a PR (#7506) to do this for all of the existing organizations which will implement the above solution for all existing organizations once we populate that table.

app/models/tasks/root_task.rb
```diff
@@ -34,4 +34,10 @@ def create_vso_subtask!(appeal, parent)
       end
     end
   end
+
+  def can_be_accessed_by_user?(user)
+    return true if HearingsManagement.singleton.user_has_access?(user)
+
+    super(user)
+  end
```
lowellrex: I find this a little confusing and I think it may be because of how we try to implement access controls on the target task's parent task. The reason we have to add this method to `RootTask` now is so that members of the Hearings Management team can create `ScheduleHearingTask`s that are children of `RootTask`s. To me that control logic should probably live in the `ScheduleHearingTask`, but we are currently constrained by the interface of `verify_user_access!()`. However I think we can fix it!

We can change `Task.verify_user_access!()` to accept a task argument in addition to the user argument it already accepts (we would have to change the two places this function is called in `GenericTask`, but that's totally doable), and then we can override `verify_user_access!()` in `ScheduleHearingTask` to do exactly what the method above does, but we'll have moved it into the task which it is more closely related to.

How does that sound?

client/app/hearingSchedule/components/AssignHearings.jsx
```diff
@@ -174,7 +202,13 @@ export default class AssignHearings extends React.Component {
       {
         header: 'Case details',
         align: 'left',
-        valueName: 'caseDetails'
+        valueName: 'caseDetails',
+        valueFunction: (veteran) => <Link
+          href={`/queue/appeals/${veteran.vacolsId}`}
+          name={veteran.vacolsId}
+          onClick={this.onClick.bind(this, veteran.vacolsId)} >
```
mdbenjam: This might be better as a button instead of a link that pushes on the history object instead. This pattern doesn't handle failures of the onClick method well. I believe onClick fires and the API request is sent, but I don't think it waits for the promise to complete. And If the request fails, there's no way to know. Not an urgent fix in this PR just something to keep in mind.

client/app/hearingSchedule/components/AssignHearings.jsx
```diff
@@ -55,12 +57,36 @@ export default class AssignHearings extends React.Component {
...
+            assigned_to_type: 'User',
+            assigned_to_id: this.props.userId
+          }
+        ]
+      }
+    };
+
+    ApiUtil.post('/tasks', payload).
+      then((resp) => {
+        const response = JSON.parse(resp.text);
```
mdbenjam: Probably don't need this anymore now that we aren't unifying the stores.

app/models/organizations/hearings_management.rb
oscarAtNava: Thanks. I will take a look at it.

client/app/queue/CaseDetailsLink.jsx
```diff
@@ -83,16 +83,13 @@ CaseDetailsLink.propTypes = {
   task: PropTypes.object,
   appeal: PropTypes.object.isRequired,
   disabled: PropTypes.bool,
+  userRole: PropTypes.string.isRequired,
```
mdbenjam: Do we still need to be passing userRole around?

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -0,0 +1,40 @@
+class ScheduleHearingTask < GenericTask
+  class << self
+    def create_from_params(params, current_user)
+      root_task = RootTask.find_by(appeal_id: params[:appeal].id)
+      if !root_task
+        root_task = RootTask.create!(appeal_id: params[:appeal].id,
+                                     appeal_type: "LegacyAppeal",
```
lowellrex: Let's use `LegacyAppeal.name` so we can throw a compile-time error if the name of this class ever changes or we ever make a typo on some later change.

client/app/queue/components/TaskTable.jsx
```diff
@@ -96,6 +97,7 @@ class TaskTable extends React.PureComponent<Props> {
       valueFunction: (task: TaskWithAppeal) => <CaseDetailsLink
         task={task}
         appeal={task.appeal}
+        userRole={this.props.userRole}
```
mdbenjam: Do we still need to pass userRole?

client/app/hearingSchedule/components/AssignHearings.jsx
oscarAtNava: You are correct. It has already happened to me where I had issues with the onClick logic and plan to fix it in the next PR where I have to manage modals and error handling.

app/models/tasks/schedule_hearing_task.rb
lowellrex: Actually I think we can just use the appeal itself to get the type and the ID here: `create!(appeal: params[:appeal], ...`

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -0,0 +1,40 @@
...
+      create!(
+        appeal: parent.appeal,
+        appeal_type: "LegacyAppeal",
+        assigned_by_id: child_assigned_by_id(parent, current_user),
+        parent_id: parent.id,
+        assigned_to: assignee,
+        instructions: params[:instructions]
+      )
+    end
+  end
```
lowellrex: After we make the change suggested here I don't think we'll need to overwrite `create_child_task` in `ScheduleHearingTask`. The one exception would be if we want to create a task of a certain type, but I think we should add that ability in to `GenericTask.create_child_task()` if that is the case. Maybe something like:
```ruby
create!(
  type: params[:task_type] || GenericTask.name,
  appeal: parent.appeal,
  ...
```

So the only subclassing we would need to do would be adding the `task_type` parameter.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def available_actions(_user)
+    [
+      {
+        label: "Assign hearing",
+        value: "modal/mark_task_complete"
+      }
+    ]
```
lowellrex: Let's put this in the task actions file: https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/constants/TASK_ACTIONS.json

Issue #7293, opened 2018-10-05
Add issue edge cases | Add Issue Landing | Ineligible lanes after prior HLR


suzchap:
## Story
When a user adds issues, they may encounter edge cases where the issue is ineligible. We will not indicate that an issue isn't eligible on any of the add issue selection pages but will indicate it on the Add issue landing page and Confirmation Page so they have the information they need to inform the applicant in the letter they send following intake data entry. If an issue is ineligible in will get logged in Caseflow but not VBMS. [Mural showing edge cases](https://app.mural.co/t/workqueue2001/m/workqueue2001/1537791678087/94469b0cdf1a02a7f57a7628702d36db2f079430).

This ticket is focused on the scenario where the issue being requested isn't eligible for that lane and how we display it's ineligibility on the Add Issues Landing page, as well as how we treat it differently when completing an intake. If a rating issue has previously been involved in a HLR, it is ineligible.

## Technical notes
- look at #7178 for how to determine whether or not a rating issue was involved in a previous HLR
- we may need to store additional metadata about the issue's ineligibility (like reference ids, etc)

## Acceptance criteria
- when rendering the list of issues on the add issue landing page, if an issue was involved in a previous HLR, we should use red text and add the 'ineligible' copy below
- when the intake is completed and the issue is being saved, save it in the caseflow db with the ineligibility reason but do not create a contention

## Notes
- Identical wording will be used on the confirmation page #7308
- Only HLR -> HLR is in scope for this issue; Appeal -> HLR and Appeal -> Appeal ineligibility will be added in https://github.com/department-of-veterans-affairs/caseflow/issues/7344

## Copy
Scenario | Display copy 
--- | --- 
If the request is an HLR and same issue was last processed as an HLR | [Issue selected] is ineligible because it was last processed as a Higher-Level Review and this can't be done twice in a row.

## Mocks
(Sample show what the first scenario would look like)
<img width="759" alt="screen shot 2018-10-06 at 10 43 34 am" src="https://user-images.githubusercontent.com/11219417/46572492-ba031480-c954-11e8-98aa-90cd5e927c86.png">

## Background issues
- #5244 

## How to Validate
(no need, Shane's got this :muscle:)

suzchap:
[...]

PR #7518, opened 2018-10-24
Do not allow repeat Higher-Level Review

pkarman:
connects #7293

### Description
If a 2nd Higher-Level Review on the same issue is attempted during intake, flag it as ineligible.

### Acceptance Criteria
- [x] Code compiles correctly

### DIFF
--- a/app/models/rating_issue.rb
+++ b/app/models/rating_issue.rb
@@ -1,12 +1,12 @@
 class RatingIssue < ApplicationRecord
   belongs_to :request_issue
 
-  # this local attribute is used to resolve the related RequestIssue
+  # this local attribute is used to resolve the contesting RequestIssue
   attr_accessor :contention_reference_id
 
   def save_with_request_issue!
-    return unless related_request_issue
-    self.request_issue = related_request_issue
+    return unless contesting_request_issue
+    self.request_issue = contesting_request_issue
     # if it already exists, update rather than attempt to insert a duplicate
     if existing_rating_issue
       existing_rating_issue.update!(request_issue: request_issue)
@@ -34,7 +34,7 @@ class RatingIssue < ApplicationRecord
     rba_contentions = [data.dig(:rba_issue_contentions) || {}].flatten
     new(
       reference_id: data[:rba_issue_id],
-      profile_date: rba_contentions.first.dig(:prfil_dt),
+      profile_date: rba_contentions.first.dig(:prfil_dt) || data[:profile_date],
       contention_reference_id: rba_contentions.first.dig(:cntntn_id),
       decision_text: data[:decn_txt],
       promulgation_date: data[:promulgation_date],
@@ -50,8 +50,12 @@ class RatingIssue < ApplicationRecord
 
   def prior_higher_level_review
     return unless reference_id
-    return unless related_request_issue
-    return related_request_issue.id if related_request_issue.review_request.is_a?(HigherLevelReview)
+    return unless contesting_request_issue
+    return contesting_request_issue.id if contesting_request_issue.review_request.is_a?(HigherLevelReview)
+  end
+
+  def contesting_request_issue
+    request_issue || find_contesting_request_issue
   end
 
   private
@@ -60,12 +64,8 @@ class RatingIssue < ApplicationRecord
     @existing_rating_issue ||= RatingIssue.find_by(participant_id: participant_id, reference_id: reference_id)
   end
 
-  def related_request_issue
-    request_issue || find_related_request_issue
-  end
-
-  def find_related_request_issue
+  def find_contesting_request_issue
     return if contention_reference_id.nil?
-    @related_request_issue ||= RequestIssue.find_by(contention_reference_id: contention_reference_id)
+    @contesting_request_issue ||= RequestIssue.find_by(contention_reference_id: contention_reference_id)
   end
 end
--- a/app/models/rating_issue.rb
+++ b/app/models/rating_issue.rb
@@ -1,12 +1,12 @@
 class RatingIssue < ApplicationRecord
   belongs_to :request_issue
 
-  # this local attribute is used to resolve the contesting RequestIssue
+  # this local attribute is used to resolve the source RequestIssue
   attr_accessor :contention_reference_id
 
   def save_with_request_issue!
-    return unless contesting_request_issue
-    self.request_issue = contesting_request_issue
+    return unless source_request_issue
+    self.request_issue = source_request_issue
     # if it already exists, update rather than attempt to insert a duplicate
     if existing_rating_issue
       existing_rating_issue.update!(request_issue: request_issue)
@@ -50,12 +50,12 @@ class RatingIssue < ApplicationRecord
 
   def prior_higher_level_review
     return unless reference_id
-    return unless contesting_request_issue
-    return contesting_request_issue.id if contesting_request_issue.review_request.is_a?(HigherLevelReview)
+    return unless source_request_issue
+    source_request_issue.review_request.is_a?(HigherLevelReview) ? source_request_issue.id : nil
   end
 
-  def contesting_request_issue
-    request_issue || find_contesting_request_issue
+  def source_request_issue
+    request_issue || find_source_request_issue
   end
 
   private
@@ -64,8 +64,8 @@ class RatingIssue < ApplicationRecord
     @existing_rating_issue ||= RatingIssue.find_by(participant_id: participant_id, reference_id: reference_id)
   end
 
-  def find_contesting_request_issue
+  def find_source_request_issue
     return if contention_reference_id.nil?
-    @contesting_request_issue ||= RequestIssue.find_by(contention_reference_id: contention_reference_id)
+    @source_request_issue ||= RequestIssue.find_by(contention_reference_id: contention_reference_id)
   end
 end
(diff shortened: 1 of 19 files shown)


client/app/intake/actions/addIssues.js
```diff
@@ -47,6 +47,7 @@ export const addRatedIssue = (args) => (dispatch) => {
       isRated: args.isRated,
       inActiveReview: foundDate[0].issues[args.issueId].in_active_review,
       timely: foundDate[0].issues[args.issueId].timely,
+      priorHigherLevelReview: foundDate[0].issues[args.issueId].prior_higher_level_review,
```
leikkisa: This isn't really a comment on your PR, but wanted to let you know that once (when I was having pains with the "foundDate" and date formatting woes I made a util function called ratingIssuesById.  I wonder if foundDate at the top could be replaced with the rating issue being worked on at this moment, to reduce the repetition here.

client/app/styles/_intake.scss
```diff
@@ -141,6 +141,7 @@
 
         .unidentified-issue,
         .in-active-review,
+        .not-eligible,
```
leikkisa: I'm wondering if there just needs to be one class for the issues with an alert.

client/app/intake/pages/addIssues.jsx
```diff
@@ -78,6 +78,9 @@ export class AddIssuesPage extends React.Component {
             } else if (!issue.timely && formType !== 'supplemental_claim') {
               issueKlasses.push('untimely');
               addendum = INELIGIBLE_REQUEST_ISSUES.untimely;
+            } else if (issue.priorHigherLevelReview && formType === 'higher_level_review') {
```
leikkisa: Starting to think this could an opportunity to refactor into some sort of validation function, might be good to wait until we're done seeing what all edge cases come up though.

client/app/styles/_intake.scss
pkarman: yeah, I'm at that point too.

client/app/intake/pages/addIssues.jsx
pkarman: Agreed. Was thinking the same. There are several more cases coming.

pkarman:
@leikkisa agreed on your points. The client code needs some refactoring now for all these issue attributes/cases. Ok to do that in a follow-on PR? I find it easier to read the history when the refactors are in their own commits.

app/models/rating_issue.rb
```diff
@@ -40,9 +41,19 @@ def in_active_review
...
+  def prior_higher_level_review
+    return unless reference_id
+    return unless related_request_issue
+    return related_request_issue.id if related_request_issue.review_request.is_a?(HigherLevelReview)
+  end
+
   private
 
   def related_request_issue
+    request_issue || find_related_request_issue
```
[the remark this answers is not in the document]
leikkisa: To add onto this, if it is
C (HLR) -> D (SC)
That means it becomes available for an HLR again... right?

In which case what we want to keep track of is the most recent review type for an issue.

app/models/rating_issue.rb
```diff
@@ -1,12 +1,12 @@
...
 
-  # this local attribute is used to resolve the related RequestIssue
+  # this local attribute is used to resolve the source RequestIssue
   attr_accessor :contention_reference_id
 
   def save_with_request_issue!
-    return unless related_request_issue
-    self.request_issue = related_request_issue
+    return unless source_request_issue
+    self.request_issue = source_request_issue
```
shanear: If we are following naming conventions, I feel like this column should be named `source_request_issue_id`,  and the relationship should be named `source_request_issue` and the function that fetches it should be named `fetch_soruce_request_issue`.

so,
`source_request_issue` becomes `fetch_source_request_issue`
`request_issue` becomes `source_request_issue`
`request_issue_id` becomes `source_request_issue_id`

app/models/rating_issue.rb
```diff
@@ -47,14 +48,24 @@ def in_active_review
     request_issue.review_title if request_issue
   end
 
+  def prior_higher_level_review
```
shanear: this should be named `source_higher_level_review`?

app/models/rating_issue.rb
```diff
@@ -47,14 +48,24 @@ def in_active_review
...
+    return unless reference_id
+    return unless source_request_issue
+    source_request_issue.review_request.is_a?(HigherLevelReview) ? source_request_issue.id : nil
+  end
+
+  def source_request_issue
+    request_issue || find_source_request_issue
```
shanear: ahh, okay we already have what i referenced in my comment above.

It's sort of strange that we are both lazy loading it, and eagerly setting it in line 9. I think we may want to choose one way or the other

app/models/rating_issue.rb
```diff
@@ -47,14 +48,24 @@ def in_active_review
...
 
   def existing_rating_issue
     @existing_rating_issue ||= RatingIssue.find_by(participant_id: participant_id, reference_id: reference_id)
   end
 
-  def related_request_issue
+  def find_source_request_issue
     return if contention_reference_id.nil?
-    @related_request_issue ||= RequestIssue.find_by(contention_reference_id: contention_reference_id)
+    @source_request_issue ||= RequestIssue.find_by(contention_reference_id: contention_reference_id)
```
shanear: yeah we should be setting this to the relation we already have on this class.

app/models/request_issue.rb
```diff
@@ -89,27 +93,62 @@ def ui_hash
...
     check_for_untimely!
+    check_for_prior_higher_level_review!
     self
   end
 
+  # the rating issue that this RequestIssue contested.
+  # It may not yet exist in the db as a RatingIssue so we pull hash from the serialized_ratings.
+  def contested_rating_issue
+    return unless review_request
+    @contested_rating_issue ||= begin
```
shanear: The `begin ... end` block might be more readable as a private method `fetch_contested_rating_issue`?

app/models/request_issue.rb
```diff
@@ -89,27 +93,62 @@ def ui_hash
...
+      rating_with_issue = review_request.serialized_ratings.find do |rating|
+        rating[:issues].find { |issue| issue[:reference_id] == rating_issue_reference_id }
+      end || { issues: [] }
+      rating_with_issue[:issues].find { |issue| issue[:reference_id] == rating_issue_reference_id }
```
shanear: It might also be nice if we converted the hash into an actual `RatingIssue` object when it's returned from this method. That seems intuitive to me.

app/models/request_issue.rb
```diff
@@ -89,27 +93,62 @@ def ui_hash
...
+
   private
 
+  def check_for_prior_higher_level_review!
+    return unless rated?
+    return unless eligible?
+    check_for_activity!(:prior_higher_level_review)
+  end
+
+  def check_for_activity!(type)
```
shanear: I don't understand the name of this method, what does `activity` mean?

client/app/intake/pages/addIssues.jsx
```diff
@@ -44,6 +44,20 @@ export class AddIssuesPage extends React.Component {
     }
   }
 
+  checkIfEligible = (issue, formType) => {
```
shanear: This much logic seems misplaced in a component, might be more idomatic to be in a reducer or selector?

spec/models/request_issue_spec.rb
```diff
@@ -99,12 +118,50 @@
...
+        :request_issue,
+        review_request: prior_higher_level_review,
+        rating_issue_reference_id: higher_level_review_reference_id,
+        contention_reference_id: contention_reference_id
+      )
+    end
+
+    it "looks up the chain to the immediately previous request issue" do
+      veteran.sync_rating_issues!
+      expect(rated_issue.previous_request_issue).to eq(prior_request_issue)
```
shanear: `prior_request_issue` 🙂

spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -710,11 +736,14 @@ def check_row(label, text)
...
-      expect(RequestIssue.find_by(rating_issue_reference_id: old_reference_id).eligible?).to eq(false)
+      expect(RequestIssue.find_by(rating_issue_reference_id: old_reference_id).untimely?).to eq(true)
+      expect(
+        RequestIssue.find_by(rating_issue_reference_id: higher_level_review_reference_id).prior_higher_level_review?
+      ).to eq(true)
 
       expect(Fakes::VBMSService).to_not have_received(:create_contentions!).with(
         hash_including(
-          contention_descriptions: array_including("Old injury", "Really old injury")
+          contention_descriptions: array_including("Old injury", "Really old injury", "Already reviewed injury")
```
shanear: Really solid feature tests, comprehensive yet simple.

app/models/request_issue.rb
pkarman: we can do that sometimes, and sometimes we can't. That's because the `RatingIssue` might not be saved in the db yet (because it has no decision), but we need to get access to the already-calculated flags in the `ui_hash`.

app/models/request_issue.rb
pkarman: I renamed to `previous_review` instead. This is a generalized method for when we have `previous_appeal` or another other preview review that disqualifies the current request issue.

PR #7519, opened 2018-10-24
Display data correctly if veteran does not exist in BGS

aroltsch:
For legacy cases, veteran info might not be present in BGS. 

image


app/models/serializers/work_queue/veteran_serializer.rb
```diff
@@ -6,20 +6,22 @@ class WorkQueue::VeteranSerializer < ActiveModel::Serializer
     object.veteran_gender
   end
   attribute :date_of_birth do
-    object.veteran.try(:date_of_birth)
+    object.veteran ? object.veteran.date_of_birth : nil
```
mdbenjam: can't we just call `object.veteran.date_of_birth` now? It'll just return nil if veteran is nil.

PR #7537, opened 2018-10-25
Adds feature flag for new edit page on intake

jiahuang:
Connects https://github.com/department-of-veterans-affairs/appeals-design-research/issues/1038

Adds a feature toggle for `intake_enable_add_issues_page` that switches from using the current finish page to the new add_issues page.

Also refactored all the review buttons to go from 5 -> 1 button.


spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -514,6 +514,17 @@ def start_higher_level_review(test_veteran, is_comp: true, claim_participant_id:
     expect(page).to have_content("#{Constants.INTAKE_FORM_NAMES.higher_level_review} has been processed.")
   end
 
+  scenario "redirects to add_issues with feature flag enabled" do
```
jiahuang: Just added one test case for the new feature flag. Donno if its worth it to add it to all the review types.

client/app/intake/pages/rampRefiling/review.jsx
```diff
@@ -1,15 +1,14 @@
...
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
-import CancelButton from '../../components/CancelButton';
 import RadioField from '../../../components/RadioField';
 import DateSelector from '../../../components/DateSelector';
 import Button from '../../../components/Button';
 import Alert from '../../../components/Alert';
 import { Redirect } from 'react-router-dom';
 import _ from 'lodash';
 import { PAGE_PATHS, INTAKE_STATES, REVIEW_OPTIONS, REQUEST_STATE } from '../../constants';
```
leikkisa: Just noting that REQUEST_STATE stopped getting imported in the other flows with the refactor

client/app/intake/components/IntakeProgressBar.jsx
```diff
@@ -21,7 +21,7 @@ class IntakeProgressBarInner extends React.PureComponent {
       },
       {
         title: '4. Finish',
```
leikkisa: Cool way to fix this. I don't think anyone asked for this, but I wonder if the title should change too or if it's good staying as "Finish"?

jiahuang: I asked about it earlier, and I think we're going to redo this but its a low pri design task so I just made it look less broken for now.

PR #7542, opened 2018-10-25
fix duplicate DAS creation

amprokop:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/7571

 # Testing
- Tested that judge -> judge assignment works and does not create a DAS record by assigning from BVAAABSHIRE to Betty Wolff
- Tested that judge -> attorney assignment is still successful


app/repositories/queue_repository.rb
```diff
@@ -165,6 +165,13 @@ def filter_duplicate_tasks(records, css_id = nil)
       end
     end
 
+    def update_location_to_judge(vacols_id, judge)
+      vacols_case = VACOLS::Case.find(vacols_id)
+      fail VACOLS::Case::InvalidLocationError, "Invalid location \"#{judge.vacols_uniq_id}\"" unless
+        judge.vacols_uniq_id
+      vacols_case.update_vacols_location!(judge.vacols_uniq_id)
```
aroltsch: Do we have to update Decass record with the new judge? Are there any fields?

app/controllers/legacy_tasks_controller.rb
```diff
@@ -41,6 +46,21 @@ def create
...
   end
 
+  def assign_to_judge
+    # If the user being assigned to is a judge, do not create a DECASS record, just
+    # update the location to the assigned judge.
+    appeal = LegacyAppeal.find(legacy_task_params[:appeal_id])
+    QueueRepository.update_location_to_judge(appeal.vacols_id, legacy_task_params[:assigned_to])
+
+    render json: {
+      task: json_task(AttorneyLegacyTask.from_vacols(
```
aroltsch: Are we returning AttorneyLegacyTask? Perhaps it should be JudgeLegacyTask?

app/repositories/queue_repository.rb
```diff
@@ -165,6 +165,13 @@ def filter_duplicate_tasks(records, css_id = nil)
       end
     end
 
+    def update_location_to_judge(vacols_id, judge)
```
aroltsch: Can you pass `appeal` here and then call vacols_case on the appeal to avoid this extra `find` query.

app/repositories/queue_repository.rb
aroltsch: There could be a Decass record, Paul told me that judges assign to other judges in DAS and this creates a new Das record. He says the DAS record creation is very inconsistent. However, in this case I think we can assume there is no DECASS record and see what happens.

app/repositories/queue_repository.rb
amprokop: wait, this wouldn't save any VACOLS queries... are you just asking for this because it might be a little bit clearer?

amprokop: if so, could you look at the rest of this file? we run `VACOLS::Case.find(vacols_id)` in a few other places too... I think I'd rather just keep it as is.

amprokop: Chatted with Anya offline.

The suggested change *would* save a roundtrip later on since we later query VACOLS for the case, we could save the results of this query in memory and avoid going back to VACOLS. 

However, it wouldn't be safe, since `update_location_in_vacols!` updates the case record using raw SQL — the ActiveRecord model wouldn't know about the new changes, and we'd have to call `reload` anyway to get the new data, which would mean another VACOLS roundtrip.

PR #7547, opened 2018-10-25
Revert

amprokop:
Revert https://github.com/department-of-veterans-affairs/caseflow/commit/8dacbd07c10f2cda59b6896dfa8f052fd88d2a29, it's causing many errors.


amprokop:
@mdbenjam — it definitely will. But this wasn't the right fix. It's created a situation where we're creating invalid DECASS records (judge -> judge) but preventing legitimate ones (judge -> attorney) from being created.

PR #7558, opened 2018-10-25
Hearing location for central office

### LINKED ISSUE #7452: Hearing Prep/Schedule | CO hearing location
We currently have different logic in AssignHearings.jsx, DailyDocket.jsx, and ListSchedule.jsx determine the hearing location for Central Office hearings. We should move all that logic to one backend function.

Sjones352:
Resolves #7452

### Description
This pr shows the hearing location for central office in the backend

### Acceptance Criteria
- [ ] We currently have different logic in AssignHearings.jsx, DailyDocket.jsx, and ListSchedule.jsx determine the hearing location for Central Office hearings. We should move all that logic to one backend function.

### Testing Plan
1. Go to Schedule


app/models/hearing.rb
```diff
@@ -74,6 +74,14 @@ def regional_office_timezone
     HearingMapper.timezone(regional_office_key)
   end
 
+  def hearing_location
```
sharonwarner: Let's just call this location. Since we're already in the hearing model, we'll usually call it like `hearing.location`, so `hearing.hearing_location` is a bit repetitive.

PR #7572, opened 2018-10-26
Fix seed data to not create vacols case that is already there

aroltsch:
Turns out that the case that we were trying to create in Facols is already there from the Facols seed data. Take a looky here: https://github.com/department-of-veterans-affairs/caseflow/blob/master/local/vacols/VACOLS::Case_dump.csv#L226

So do not create it!

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -484,23 +484,12 @@ class SeedDB
 
   def create_previously_held_hearing_data
     user = User.find_by_css_id("BVAAABSHIRE")
-    veteran_file_number = "994806951S"
-    appeals = LegacyAppeal.where(vbms_id: veteran_file_number)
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -484,23 +484,12 @@ def setup_dispatch
...
-    return if (appeals.map(&:type) - ["Post Remand", "Original"]).empty? &&
-              appeals.flat_map(&:hearings).map(&:disposition).include?(:held)
+    return if ([appeal.type] - ["Post Remand", "Original"]).empty? &&
+              appeal.hearings.map(&:disposition).include?(:held)
 
-    FactoryBot.create(
-      :legacy_appeal,
-      vacols_case: FactoryBot.create(
-        :case,
-        :assigned,
```
aroltsch: do we need it to be assigned to the user?

PR #7576, opened 2018-10-26
Verify in active review rating issue logic

pkarman:
connects #7291 

### Description
Verifying the `in_active_review` logic really works.

This PR fixes multiple issues:

* timezone is now always reset after each test
* Appeal has a different definition of "active" than ClaimReview (caused #7291 to fail validation)
* fixes flakey HLR feature test

### DIFF
--- a/app/models/ama_review.rb
+++ b/app/models/ama_review.rb
@@ -29,9 +29,7 @@ class AmaReview < ApplicationRecord
       rating[:issues].each do |rating_issue_hash|
         rating_issue_hash[:timely] = timely_rating?(Date.parse(rating_issue_hash[:promulgation_date].to_s))
         # always re-compute flags that depend on data in our db
-        rating_issue = RatingIssue.from_ui_hash(rating_issue_hash)
-        rating_issue_hash[:in_active_review] = rating_issue.in_active_review
-        rating_issue_hash[:source_higher_level_review] = rating_issue.source_higher_level_review
+        rating_issue_hash.merge!(RatingIssue.from_ui_hash(rating_issue_hash).ui_hash)
       end
     end
   end
--- a/app/models/ama_review.rb
+++ b/app/models/ama_review.rb
@@ -26,8 +26,12 @@ class AmaReview < ApplicationRecord
     return unless receipt_date
 
     cached_serialized_ratings.each do |rating|
-      rating[:issues].each do |rating_issue|
-        rating_issue[:timely] = timely_rating?(Date.parse(rating_issue[:promulgation_date].to_s))
+      rating[:issues].each do |rating_issue_hash|
+        rating_issue_hash[:timely] = timely_rating?(Date.parse(rating_issue_hash[:promulgation_date].to_s))
+        # always re-compute flags that depend on data in our db
+        rating_issue = RatingIssue.from_ui_hash(rating_issue_hash)
+        rating_issue_hash[:in_active_review] = rating_issue.in_active_review
+        rating_issue_hash[:source_higher_level_review] = rating_issue.source_higher_level_review
       end
     end
   end
--- a/app/models/rating_issue.rb
+++ b/app/models/rating_issue.rb
@@ -49,7 +49,9 @@ class RatingIssue < ApplicationRecord
       reference_id: reference_id,
       decision_text: decision_text,
       promulgation_date: promulgation_date,
-      contention_reference_id: contention_reference_id
+      contention_reference_id: contention_reference_id,
+      in_active_review: in_active_review,
+      source_higher_level_review: source_higher_level_review
     }
   end
 
--- a/app/models/rating_issue.rb
+++ b/app/models/rating_issue.rb
@@ -22,7 +22,8 @@ class RatingIssue < ApplicationRecord
         participant_id: ui_hash[:participant_id],
         reference_id: ui_hash[:reference_id],
         decision_text: ui_hash[:decision_text],
-        promulgation_date: ui_hash[:promulgation_date]
+        promulgation_date: ui_hash[:promulgation_date],
+        contention_reference_id: ui_hash[:contention_reference_id]
       )
     end
   end
@@ -47,7 +48,8 @@ class RatingIssue < ApplicationRecord
       participant_id: participant_id,
       reference_id: reference_id,
       decision_text: decision_text,
-      promulgation_date: promulgation_date.to_date,
+      promulgation_date: promulgation_date,
+      contention_reference_id: contention_reference_id
     }
   end
 
--- a/app/models/rating_issue.rb
+++ b/app/models/rating_issue.rb
@@ -19,6 +19,7 @@ class RatingIssue < ApplicationRecord
 
     def from_ui_hash(ui_hash)
       new(
+        participant_id: ui_hash[:participant_id],
         reference_id: ui_hash[:reference_id],
         decision_text: ui_hash[:decision_text],
         promulgation_date: ui_hash[:promulgation_date]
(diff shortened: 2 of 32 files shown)


app/models/rating_issue.rb
```diff
@@ -73,6 +74,6 @@ def existing_rating_issue
 
   def fetch_source_request_issue
     return if contention_reference_id.nil?
-    self.source_request_issue ||= RequestIssue.find_by(contention_reference_id: contention_reference_id)
+    self.source_request_issue ||= RequestIssue.unscoped.find_by(contention_reference_id: contention_reference_id)
```
pkarman: this was causing feature test to fail

app/models/request_issue.rb
```diff
@@ -168,4 +170,8 @@ def check_for_nonrated_untimely!
       self.ineligible_reason = :untimely
     end
   end
+
+  def appeal_active?
```
pkarman: The `Appeal` model has a different definition of "active"

spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -737,9 +739,15 @@ def check_row(label, text)
       expect(ineligible_issue.contention_reference_id).to be_nil
 
       expect(RequestIssue.find_by(rating_issue_reference_id: old_reference_id).untimely?).to eq(true)
-      expect(
-        RequestIssue.find_by(rating_issue_reference_id: higher_level_review_reference_id).previous_higher_level_review?
```
pkarman: this was nondeterministic, since `find_by` will return only one match, even if there are more.

spec/rails_helper.rb
```diff
@@ -247,9 +247,14 @@ def read_csv(klass, date_shift)
     Rails.cache.clear
   end
 
+  config.before(:each) do
```
pkarman: fix flakey test ordering where setting a non-default timezone was not getting restored after the test completes.

app/models/request_issue.rb
```diff
@@ -168,4 +172,8 @@ def check_for_nonrated_untimely!
...
+    review_request.tasks.where.not(status: Constants.TASK_STATUSES.completed).count > 0
```
shanear: Seems like this logic should live on the `Appeal` not the request issue.

So it'd look like this:

`review_request.active?`, then we could just define `active?` for each type of review request (`Appeal`, `SupplementalClaim` and `HigherLevelReview`) which would clean up the need for a specific `appeal_active?` method and some of the other logic here.

spec/rails_helper.rb
```diff
@@ -250,12 +250,14 @@ def read_csv(klass, date_shift)
...
   config.before(:each) do
     FeatureToggle.enable!(:test_facols)
+    @spec_time_zone = Time.zone
   end
 
   config.after(:each) do
     Timecop.return
     Rails.cache.clear
     FeatureToggle.disable!(:test_facols)
+    Time.zone = @spec_time_zone
```
shanear: nice! 💯

app/models/ama_review.rb
```diff
@@ -26,8 +26,12 @@ def serialized_ratings
     return unless receipt_date
 
     cached_serialized_ratings.each do |rating|
```
shanear: If we are now converting the serialized values back into objects via:
```
rating_issue = RatingIssue.from_ui_hash(rating_issue_hash)
```
It seems like we could just push all of this logic back into `RatingIssue`

so:

```
def cached_ratings
  cached_serialized_ratings.map do |rating_hash|
    Rating.from_ui_hash(rating_hash)
  end
end
```

Then we can do:
```
cached_ratings.map(&:ui_hash) 
```
whenever we need the serialized values.

app/models/request_issue.rb
pkarman: I agreed at first, but then I realize looking at this now that we are conflating the status of the `RequestIssue` and its parent `review_request`.

Since a `review_request` may have more than one `RequestIssue` associated with it, I think what we want to know is whether the current `RequestIssue` is active, regardless of what the status of any other `RequestIssue` on its parent `review_request` is.

To apply that logic to `Appeal`-based `RequestIssue` instances, I need to do some discovery.

app/models/ama_review.rb
leikkisa: Agree with Shane, that seems more intuitive to me. Also should we call it something other than from_ui_hash since these are being created from BGS?

pkarman: The `timely` check is on the `AmaReview` though, since the dates involved are on the `AmaReview` and we didn't want to pass them down to `Rating` or `RatingIssue`.

We could move the other calculations though, and end up with:

```ruby
rating_issue_hash = RatingIssue.from_ui_hash(rating_issue_hash).ui_hash
```

app/models/rating_issue.rb
leikkisa: I know you mentioned this, but I am curious where RequestIssue is getting scoped. I am curious because it's not intuitive that it is, and am wondering if it will have impacts elsewhere.

app/models/request_issue.rb
```diff
@@ -147,6 +150,7 @@ def check_for_active_request_issue!
     existing_request_issue = self.class.find_active_by_reference_id(rating_issue_reference_id)
     if existing_request_issue
       self.ineligible_reason = :duplicate_of_issue_in_active_review
+      self.ineligible_request_issue_id = existing_request_issue.id
```
leikkisa: Yay, this will be useful.

app/models/request_issue.rb
```diff
@@ -4,6 +4,8 @@ class RequestIssue < ApplicationRecord
   has_many :decision_issues
   has_many :remand_reasons
   has_many :decision_rating_issues, foreign_key: "source_request_issue_id", class_name: "RatingIssue"
+  has_many :duplicate_but_ineligible, class_name: "RequestIssue", foreign_key: "ineligible_request_issue_id"
+  belongs_to :ineligible_due_to, class_name: "RequestIssue", foreign_key: "ineligible_request_issue_id"
```
leikkisa: I might not be understanding correctly.  If an issue is a duplicate... it will have the original eligible issue in "ineligible_request_issue_id"?  It seems like that would be the "eligible_request_issue_id".

app/models/request_issue.rb
leikkisa: Yeah it seems like it might be different for Appeal. For the EPs it's based on whether the entire EP has been cleared... right? But since Appeals don't have EPs maybe it's per issue.

pkarman: Based on my conversation with @mdbenjam and @shanear I think we should leave this `appeal_active?` method here for now. The proper place probably is the `Appeal` model, but it sounds like the business flow for Appeal and HLR/SC is different enough that to preserve the status check on the `RequestIssue` itself, we need to keep the checks in this class.

app/models/request_issue.rb
pkarman: How about `ineligible_due_to_id`?

app/models/ama_review.rb
pkarman: They are not being created directly from the BGS response. That case is handled by `from_bgs_hash`.

The `from_ui_hash` takes the output of `ui_hash` and re-hydrates the original instance. I.e. from a serialized hash, create a model instance.

app/models/rating_issue.rb
pkarman: I think it's when the `create` is called via the association method:

https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/request_issues_update.rb#L18

in that context, it always has the scope of the `review` which in this case means it will never find anything unless we `unscoped` it.

app/models/request_issue.rb
leikkisa: That sounds good!

PR #7581, opened 2018-10-26
Fix hearing AOD

### LINKED ISSUE #7525: Assign Hearings | AOD Status
- The AOD query on line 281 in appeal_repository.rb is not correct. We need that query to pull *only* AOD cases in location 57. 
- We need to send AOD status to the frontend for all of the cases returned in `appeals_ready_for_hearing_schedule`. It should be a boolean stored on each of the appeals (for example, appeal.aod = true/false)
- On the frontend, if AOD is set to true, we need to display that as part of appeal type

mdbenjam:
Resolves #7525

### Description
Three changes in this PR
1) We updated the AOD cases query to actually be a `where` on AOD cases.
2) We started passing whether a case was AOD to the frontend
3) We started displaying whether a case was AOD on the frontend

### Testing Plan
1. Go to the central office location
1. See that on both tabs we see some AOD cases.


app/services/appeal_repository.rb
```diff
@@ -279,11 +279,17 @@ def self.appeals_ready_for_hearing_schedule(regional_office)
...
+      .where("aod = 1").where(bfregoff: regional_office, bfcurloc: "57").order("folder.tinum").limit(30)
     other_cases = VACOLS::Case.joins(:folder).where(bfregoff: regional_office, bfcurloc: "57")
       .order("folder.tinum").limit(30)
 
-    (cavc_cases + aod_cases + other_cases).uniq.first(30).map { |case_record| build_appeal(case_record, true) }
+    aod_vacols_ids = aod_cases.pluck(:bfkey)
+
+    (cavc_cases + aod_cases + other_cases).uniq.first(30).map do |case_record|
+      build_appeal(case_record, true).tap do |appeal|
+        appeal.aod = aod_vacols_ids.include?(appeal.vacols_id)
```
mdbenjam: Feels like there should be a better way to do this.... but aod is complicated. I'm not sure how to select `aod` in the above ActiveRecord query without it stopping the selection of all the other fields. So for now I think this is a reasonable solution.

client/app/hearingSchedule/components/AssignHearings.jsx
```diff
@@ -138,23 +138,12 @@ export default class AssignHearings extends React.Component {
     </div>;
   };
 
-  veteranTypeColor = (appealType) => {
```
mdbenjam: Get rid of this in favor of the Queue method that does a similar thing.

client/app/hearingSchedule/components/AssignHearings.jsx
```diff
@@ -138,23 +138,12 @@ export default class AssignHearings extends React.Component {
...
-
-    if (appealType === 'Court Remand') {
-      return <span {...colorAOD}>CAVC</span>;
-    } else if (appealType === 'AOD') {
-      return <span {...colorAOD}>AOD</span>;
-    }
-
-    return appealType;
+  getHearingTime = (date, regionalOfficeTimezone) => {
```
mdbenjam: Just fixing indentation.

mdbenjam:
What tests would you like me to add?

client/app/hearingSchedule/components/AssignHearings.jsx
```diff
@@ -13,13 +13,9 @@ import { css } from 'glamor';
 import { COLORS } from '../../constants/AppConstants';
 import { getTime, getTimeInDifferentTimeZone } from '../../util/DateUtil';
 import ApiUtil from '../../util/ApiUtil';
-import { prepareTasksForStore } from '../../queue/utils';
+import { prepareTasksForStore, renderAppealType } from '../../queue/utils';
```
sharonwarner: Should we move some of these util functions to client/app/util if they're not queue specific?

mdbenjam: I think longterm that might be a good idea. Though I'm curious if there are better ways for us to more fundamentally share some of the components.

PR #7582, opened 2018-10-29
Display decision date for AMA appeals in search

### LINKED ISSUE #7575: Populate decision date for AMA appeals in search
We are not showing the decision date in the Search results/case list for decided AMA appeals.

lowellrex:
Resolves #7575.

When we [create BvaDispatchTasks](https://github.com/department-of-veterans-affairs/caseflow/blob/505a496eef2b77eeb0bd6726d68b2e86050a37c6/app/models/tasks/bva_dispatch_task.rb#L5), we do not set the action of those tasks to "review". Removing that constraint returns a task for decided appeals.

Currently there are 14 decided AMA appeals but when the query is constrained on the action column we return none of them. Ran a quick confirmation in the prod rails console:
```
rails c> BvaDispatchTask.where(action: "review", status: "completed").count
=> 0
rails c> BvaDispatchTask.where(status: "completed").count
=> 14
```

## Test plan:
* Locally create a decided appeal:
```
rails c> a = Appeal.find(5)
rails c> t = BvaDispatchTask.create!(appeal: a, status: Constants.TASK_STATUSES.completed, assigned_to: BvaDispatch.singleton)
rails c> t.update!(completed_at: t.created_at)
```
* Search for Veteran ID 228081153
* Notice that the decision date now appears in the table for the AMA appeal.


amprokop:
Hmm. I'm unsure if this is the right code from a business perspective. 

We'll return will be the date that BVA dispatch completes their task — which is usually but not always the same date as the date we're saving in https://github.com/department-of-veterans-affairs/caseflow/blob/master/db/schema.rb#L194. That date is manually set with an IDT API call by BVA Dispatch users to match the date an appeal is mailed. Sometimes they batch process appeals (e.g. during a WFH day)and mail them all later, so unfortunately the distinction is important here.

PR #7584, opened 2018-10-29
Cleaning up permissions for organizations

mdbenjam:
### Description
1) Removing the feature toggle entry in the organization table since we don't use it.
2) Allowing System Admins to access organization pages.

### Testing Plan
1. Log in as a system admin ensure that you can access the organization users page
1. Log in as an organization user and ensure you can access the organization page still
1. Log in as a non-organization user and ensure you cannot access the organization page


app/controllers/organizations/users_controller.rb
```diff
@@ -1,8 +1,4 @@
 class Organizations::UsersController < OrganizationsController
-  before_action :verify_organization_access
```
mdbenjam: These were redundant since we inherit from the organizations controller.

app/controllers/organizations_controller.rb
```diff
@@ -12,17 +11,15 @@ def show
   private
 
   def verify_organization_access
+    return if current_user.admin?
```
mdbenjam: I'm not sure if I should put this here (giving system admins access to all org queues) or if I should override these methods in the user organization sub class and allow system admin access only for the user page. Thoughts?

spec/feature/queue/user_organization_spec.rb
```diff
@@ -4,30 +4,53 @@
...
 
-    expect(user_with_role.organizations.count).to eq(0)
+      click_on "Remove-user-#{user_with_role.id}"
+      expect(page).to_not have_content(user_with_role.full_name)
+
+      expect(user_with_role.organizations.count).to eq(0)
+    end
+  end
+
+  context "When user is not in the organization" do
```
mdbenjam: Not sure I should put this feature test in or a test in `tasks_controller_spec.rb` above, since the change does effect all organization related actions.

app/controllers/organizations_controller.rb
lowellrex: Slight preference to overriding this method in `Organizations::UsersController`. I don't think we really care that system admins will be able to view other organizations' queues, but I can imagine the scenario months from now that we do want to add some action on the organization controller that we do want to be limited to only "real" members of that team and we might forget this exception exists.

That being said, I think we're fine either way.

spec/feature/queue/user_organization_spec.rb
lowellrex: I think it makes sense here.

PR #7585, opened 2018-10-29
Remove Organizations' reliance on staff table for membership

lowellrex:
Connects #7596. Replaces the `Organization` models reliance on the `vacols.staff` table for membership with the caseflow `organizations_users` table.

I added the members of the QR team to the new table by running the following in the prod rails console:
```
rails c> org = QualityReview.singleton
rails c> org.members.each { |u| OrganizationsUser.add_user_to_organization(u, org) }
```

### TODO:
* Make tests pass

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -164,9 +164,9 @@ class Task < ApplicationRecord
 
   def assign_to_user_data
     users = if assigned_to.is_a?(Organization)
-              assigned_to.members
+              assigned_to.users
             elsif parent && parent.assigned_to.is_a?(Organization)
-              parent.assigned_to.members.reject { |member| member == assigned_to }
+              parent.assigned_to.users.reject { |u| u == assigned_to }
             else
               []
             end
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -20,23 +20,6 @@ class Organization < ApplicationRecord
   end
 
   def user_has_access?(user)
-    members.pluck(:id).include?(user.id)
-  end
-
-  def members
-    @members ||= User.where(css_id: member_css_ids.uniq)
-  end
-
-  private
-
-  def member_css_ids
-    return [] if staff_field_for_organization.empty?
-
-    staff_records = VACOLS::Staff.where(sactive: "A")
-    staff_field_for_organization.each do |sfo|
-      staff_records = sfo.filter_staff_records(staff_records)
-    end
-
-    staff_records.pluck(:sdomainid)
+    users.pluck(:id).include?(user.id)
   end
 end


app/models/organizations_user.rb
```diff
@@ -3,10 +3,14 @@ class OrganizationsUser < ApplicationRecord
   belongs_to :user
 
   def self.add_user_to_organization(user, organization)
-    create(organization_id: organization.id, user_id: user.id)
+    existing_record(user, organization) || create(organization_id: organization.id, user_id: user.id)
```
lowellrex: Added this so that we just return the existing row instead of throwing an error if the user/organization combination exists. Ran into this case while testing and it seems to me like the row already existing in the database should be considered a successful result when trying to add a user to the organization.

app/models/tasks/judge_task.rb
```diff
@@ -27,6 +27,10 @@ def self.create(params)
     super(params.merge(action: "assign"))
   end
 
+  def self.verify_user_can_assign!(user)
+    QualityReview.singleton.user_has_access?(user) || super(user)
```
lowellrex: Members of the QR team should be allowed to assign judge tasks to other folks (judges). The tests incorrectly passed before because the "QR" user we were using to assign the task [was an attorney](https://github.com/department-of-veterans-affairs/caseflow/blob/06947972611cf449fdd13913a6fc0d64d3ad70a8/spec/factories/vacols/staff.rb#L7) so the [base `Task.verify_user_can_assign!()`](https://github.com/department-of-veterans-affairs/caseflow/blob/06947972611cf449fdd13913a6fc0d64d3ad70a8/app/models/task.rb#L129) allowed the assignment.

spec/feature/queue/ama_queue_spec.rb
```diff
@@ -381,7 +382,7 @@
       fill_in "taskInstructions", with: quality_review_instructions
 
       click_on "Submit"
-      expect(page).to have_content("You have no cases assigned")
+      expect(page).to have_content("Task assigned to #{judge_user.full_name}")
```
lowellrex: Judge tasks do not change the status of their parent when they are created like Generic Tasks do, so we expect to see the parent QR case in this users queue with a success message above it. I am not entirely sure why this test worked before.

app/models/tasks/judge_task.rb
mdbenjam: I thought qr folks were attorneys

db/seeds.rb
```diff
@@ -454,19 +455,10 @@ def create_organizations
...
-    StaffFieldForOrganization.create!(organization: dispatch, name: "stitle", values: %w[A1 A2], exclude: true)
-
-    mail_team = MailTeam.singleton
-    StaffFieldForOrganization.create!(organization: mail_team, name: "sdept", values: %w[MAIL])
-
-    quality_review = QualityReview.singleton
-    StaffFieldForOrganization.create!(organization: quality_review, name: "sdept", values: %w[QR])
-
+    Organization.create!(name: "Translation", url: "translation")
+    BvaDispatch.singleton
```
mdbenjam: Why do we need these lines we already call this method above

spec/feature/queue/ama_queue_spec.rb
mdbenjam: Hmmm interesting. Should they change the status? I feel like they should

db/seeds.rb
lowellrex: Removed!

app/models/tasks/judge_task.rb
lowellrex: Great point! So I guess the tests were passing correctly previously then.

I still think this check is valuable for two reasons:
1. The only types of folks we want to explicitly allow to assign JudgeTasks are members of the QR team (for now, the only other way JudgeTasks should be created is by the application), and I think explicitly checking that the person assigning the task is a member of the QR team more clearly describes the behaviour we want to allow.
2. Before this change we were relying only on [this class' parent's method](https://github.com/department-of-veterans-affairs/caseflow/blob/3a7a0dc7499214f27f1098b484bc09088c6dee59/app/models/task.rb#L128) to do this authorization check for us. Currently, that method allows attorneys to assign tasks (and all members of the QR team are attorneys), but I can imagine that we may want to restrict that in the future so that attorneys do not have the ability to assign all of the tasks they please. If we do that, I think it would not be obvious why QR folks can no longer assign tasks to judges (especially since we expect QR assigning tasks to judges to be a relatively infrequent occurrence, so there may be a fair amount of time in between when we change `Task.verify_user_can_assign!()` and when we notice this error).

app/models/task.rb
```diff
@@ -57,10 +57,15 @@ def self.create_many_from_params(params_array, current_user)
...
+    params = modify_params(params)
+    create(params)
+  end
+
+  def self.modify_params(params)
     if params.key?("instructions") && !params[:instructions].is_a?(Array)
       params["instructions"] = [params["instructions"]]
     end
-    create(params)
+    params
```
lowellrex: We transformed incoming instructions params [in a few places](https://github.com/department-of-veterans-affairs/caseflow/pull/7585/commits/366d29e0844d86f9f7f6d4f0b0ff90f09e1a498d) so I busted them out to their own little function.

app/models/tasks/colocated_task.rb
```diff
@@ -10,14 +10,12 @@ class ColocatedTask < Task
...
   class << self
     # Override so that each ColocatedTask for an appeal gets assigned to the same colocated staffer.
+    # rubocop:disable Lint/UselessAssignment
     def create_many_from_params(params_array, _)
-      params_array.each do |params|
-        if params.key?("instructions") && !params[:instructions].is_a?(Array)
-          params["instructions"] = [params["instructions"]]
-        end
-      end
+      params_array.each { |params| params = modify_params(params) }
```
lowellrex: There may be a more _railsy_ way to do this, but I needed to assign this to something to persist the changes for some reason.

client/app/queue/selectors.js
```diff
@@ -153,7 +153,7 @@ export const incompleteOrganizationTasksByAssigneeIdSelector = createSelector(
 
 export const newTasksByAssigneeCssIdSelector = createSelector(
   [incompleteTasksByAssigneeCssIdSelector],
-  (tasks: Array<Task>) => tasks.filter((task) => !task.placedOnHoldAt)
+  (tasks: Array<Task>) => tasks.filter((task) => !taskIsOnHold(task))
```
lowellrex: Result of a false start about why a test was failing, but still a useful change I think.

app/models/tasks/colocated_task.rb
mdbenjam: Maybe just map and pass the resulting array to the create?

mdbenjam: Also why can't we just use the inherited one in `task.rb`?  Seems like it does the same thing? Or modify the `create_from_params` instead of `create_many_from_params`

lowellrex: Map, smart! I'll make that change.

We can't use `Task.create_many_from_params` or `Task.create_from_params` because `ColocatedTask.create` expects to receive an array of parameters instead of a single parameter hash. Maybe `ColocatedTask.create` could be cleaned up at some point but I think that would have to be a later PR.

app/models/tasks/judge_task.rb
```diff
@@ -23,10 +23,25 @@ def available_actions(user)
     end
   end
 
-  def self.create(params)
+  def self.create_from_params(params, user)
+    new_task = super(params, user)
+
+    parent = Task.find(params[:parent_id]) if params[:parent_id]
+    if parent && parent.type == QualityReviewTask.name
+      parent.update!(status: :on_hold)
```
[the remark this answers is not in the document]
lowellrex: Yup, you're absolutely right! Ticketed #7630.

client/app/queue/utils.js
```diff
@@ -431,6 +432,11 @@ export const taskHasNewDocuments = (task: Task, newDocsForAppeal: NewDocsForAppe
   return newDocsForAppeal[task.externalAppealId].docs.length > 0;
 };
 
-export const taskIsOnHold = (task: Task) =>
-  moment().startOf('day').
-    diff(moment(task.placedOnHoldAt), 'days') < task.onHoldDuration;
+export const taskIsOnHold = (task: Task) => {
+  if (task.onHoldDuration && task.placedOnHoldAt) {
```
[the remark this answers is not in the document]
lowellrex: Some (all?) tasks are placed on hold when we "assign to team/person". Those holds do not have a hold duration, but should still be counted as "on hold". This change fixes that.

PR #7592, opened 2018-10-29
Case Details - Assign Veteran and complete task.

### LINKED ISSUE #7072: Video and CO Release | Epic 2  Case Details - Assign Hearing

oscarAtNava:
Resolves #7072 

### Description
Integrates hearings and Queue's Case Details. Created a Schedule Hearing task with an assign hearing action. Expanded the task model to have many business payload items. Promoted the Regional Office selector to common components and defined a common reducer for any common component to have. Added this common reducer to both hearings and queue reducers. 

### Acceptance Criteria
- [ ] Code compiles correctly
- [ ] Test cases pass

### Testing Plan
1. Go to Assign Hearings page.
2. Click on a veteran ready to get a hearing scheduled.
3. Verify you go to Case Details and see a task with the Assign Hearing action.
4. When Assign Hearing is selected a modal pops up with the hearings info. 
5. Click on Assign in the modal.
6. Confirm the veteran now shows up in the Schedule Hearings tab in Assign Hearings.
7. Check the database that the hearsched row was created and that the location is now either 38 or 36 depending on whether it is a video or central office hearing.


app/models/tasks/schedule_hearing_task.rb
```diff
@@ -29,12 +33,47 @@ def create_child_task(parent, current_user, params)
     end
   end
 
+  def mark_as_complete!
+    hearing_pkseq = task_business_payloads[0].values[2]
+    Rails.logger.info("OARVT hearing_pkseq #{hearing_pkseq} .")
+    hearing_type = task_business_payloads[0].values[3]
+    hearing = VACOLS::CaseHearing.find(hearing_pkseq)
+
+    if hearing_type == "Central"
```
sharonwarner: Can we create a constant in hearing.rb for hearing_type? We're currently using 'CO' in hearing prep and 'Central' in hearing schedule. I already have a different ticket to decide on one and update existing uses, but for now, I think we should avoid adding more instances of a hardcoded hearing type.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -29,12 +33,47 @@ def create_child_task(parent, current_user, params)
...
+      HearingRepository.update_vacols_hearing!(hearing, folder_nr: appeal.vacols_id)
+    else
+      hearing_hash = to_hash(hearing)
+      hearing_hash[:folder_nr] = appeal.vacols_id
+      # hearing date should be from payload.
+      hearing_hash[:hearing_date] = task_business_payloads[0].values[4].to_s
```
sharonwarner: I'm not sure exactly how the payloads work, but if it's easy to make this into a hash, I think that's preferable. It's not obvious to me that the fourth value is going to be the hearing date. Alternatively, we could create a constant that maps those values.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -29,12 +33,47 @@ def create_child_task(parent, current_user, params)
...
+      HearingRepository.create_vacols_child_hearing(hearing_hash)
+    end
+
+    AppealRepository.update_location!(appeal, location_based_on_hearing_type(hearing_type))
+
+    super
+  end
+
+  def location_based_on_hearing_type(hearing_type)
+    if hearing_type == "Central"
```
sharonwarner: Let's use the constant above here as well.

app/repositories/hearing_repository.rb
```diff
@@ -55,6 +55,14 @@ def update_vacols_hearing!(vacols_record, hearing_hash)
       vacols_record.update_hearing!(hearing_hash.merge(staff_id: vacols_record.slogid)) if hearing_hash.present?
     end
 
+    def create_vacols_child_hearing(parent_hearing_hash)
```
sharonwarner: How do we set the child record's folder_nr or date field?

client/app/queue/components/AssignHearingModal.jsx
```diff
@@ -0,0 +1,253 @@
...
+        //this.props.setTaskAttrs(task.uniqueId, preparedTasks[task.uniqueId]);
+        this.props.history.goBack();
+      });
+  };
+
+  render = () => {
+    const hearingDateStr = formatDate(this.props.task.taskBusinessPayloads[0].values[4]);
+    const timeStr = getTime(this.props.task.taskBusinessPayloads[0].values[4]);
+
+    const timeOptions = this.props.task.taskBusinessPayloads[0].values[3] === 'Video' ?
```
sharonwarner: Can we make this a constant in a shared file somewhere? I'll need the same values on the Daily Docket.

app/repositories/hearing_repository.rb
oscarAtNava: We set them in the schedule hearing task before calling the repository method.

app/models/tasks/schedule_hearing_task.rb
oscarAtNava: Done

app/models/tasks/schedule_hearing_task.rb
oscarAtNava: Storing a json object now.

app/models/tasks/schedule_hearing_task.rb
oscarAtNava: Done

client/app/queue/components/AssignHearingModal.jsx
oscarAtNava: Done

app/repositories/hearing_repository.rb
sharonwarner: I see. I think it's a little ambiguous to have this parameter named parent_hearing_hash then. Could we just call it hearing_hash?

local/vacols/VACOLS::Representative_dump.csv
```diff
@@ -23,13 +23,11 @@ repkey,repaddtime,reptype,repso,replast,repfirst,repmi,repsuf,repaddr1,repaddr2,
 2874031,2017-03-27,A,,Reinger,Carol,L,I,57452 Bergstrom Mountain,Suite 485,Leafort,MD,90785,538.615.0459,Iusto voluptas ut aut.,SBEAHAN,2017-03-27,,,,,,CK1281722,
 2874889,2017-09-12,A,,Gottlieb,Wilburn,R,Sr.,39982 Dibbert Hills,Apt. 858,Leuschkeville,KY,65314,568.102.2351,Vitae magni temporibus corrupti quae sint.,KBERGNAUM,2017-09-12,,,,,,CK759075,
 2892519,2016-01-05,G,,Wunsch,Mara,S,DDS,493 Powlowski Island,Apt. 582,Lake Aiyana,HI,60887,(227) 408-4657 x946,Ex temporibus modi illum inventore praesentium.,RO09,2016-01-05,,,,,,CK1111639,
-2895173,2016-10-19,A,,Glover,Freida,Z,III,60597 Wunsch Square,Apt. 329,Hirthemouth,NC,99087,(756) 443-9391 x62989,,RO17,2016-10-19,,,,,,CK1037809,
```
mdbenjam: This shouldn't be necessary any more. The new FACOLS container doesn't have this unique constraint.

db/migrate/20181030123030_create_task_business_payload.rb
```diff
@@ -0,0 +1,11 @@
+class CreateTaskBusinessPayload < ActiveRecord::Migration[5.1]
+  safety_assured
+
+  def change
```
mdbenjam: When we chatted the other day I thought you were suggesting a way to store business specific logic for hearings that was going to be needed outside of just sharing context between Queue and Hearing Schedule. i.e. I thought this business payload table was going to be something like a scheduled hearing table.

However, it seems this is just a temporary way to share information between the two apps. Is the information in this table useful outside of this flow? I'm not sure we need this to be such a general solution.

local/vacols/VACOLS::Representative_dump.csv
oscarAtNava: Row restored.

app/models/serializers/work_queue/task_serializer.rb
```diff
@@ -83,4 +83,10 @@ class WorkQueue::TaskSerializer < ActiveModel::Serializer
   attribute :available_actions do
     object.available_actions_unwrapper(@instance_options[:user])
   end
+
+  attribute :task_business_payloads do
```
mdbenjam: This is okay, though I'm not sure we need to always pass this information. It might be overkill. We do have a mechanism to give data to specific tasks:

We grab an optional `func` param from task actions:
https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/task.rb#L34

And return the result of calling it in the task action itself.
https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/constants/TASK_ACTIONS.json#L2

client/app/components/common/reducers.js
```diff
@@ -0,0 +1,25 @@
+import { ACTIONS } from './actionTypes';
```
mdbenjam: I think this reducer needs a more descriptive name.

db/migrate/20181030123030_create_task_business_payload.rb
oscarAtNava: I believe you will find use cases for this in the future, specially when you start reassigning tasks and you want to reference business objects that are not associated to an appeal directly. Just my two cents.

mdbenjam: For something impermanent I think a simpler solution like query params might be the best way forward. It's a little hacky, but I think in the end a lot simpler.

client/app/components/common/reducers.js
oscarAtNava: The name of the reducer exported is commonComponentsReducer. As far as the file names for supporting files I followed what is used in other products (queue).

db/migrate/20181030123030_create_task_business_payload.rb
oscarAtNava: query params entails passing the data through to the back-end over and over. In this implementation you only have to move the task and the business data is already available to the back-end as it travels along the flow. 
Happy to change it at a later point if you feel strongly about it.

PR #7597, opened 2018-10-30
Untimely issue exemption modal

leikkisa:
connects #7301


leikkisa:
screen shot 2018-11-01 at 3 55 51 pm

app/models/request_issue.rb
```diff
@@ -7,6 +7,9 @@ class RequestIssue < ApplicationRecord
   has_many :duplicate_but_ineligible, class_name: "RequestIssue", foreign_key: "ineligible_due_to_id"
   belongs_to :ineligible_due_to, class_name: "RequestIssue", foreign_key: "ineligible_due_to_id"
 
+  # enum is symbol, but validates requires a string
+  validates :ineligible_reason, exclusion: { in: ["untimely"] }, if: proc { |reqi| reqi.untimely_exemption }
```
shanear: not sure we need this validation, especially if application logic, not user input is adding the `ineligible_reason`

app/models/request_issue.rb
```diff
@@ -42,12 +45,16 @@ def from_intake_data(data)
         decision_date: data[:decision_date],
         issue_category: data[:issue_category],
         notes: data[:notes],
-        is_unidentified: data[:is_unidentified]
+        is_unidentified: data[:is_unidentified],
```
shanear: it's not particularly idiomatic to prepend booleans with `is_` in ruby, but I'm okay with keeping it this way too.

app/models/request_issue.rb
```diff
@@ -42,12 +45,16 @@ def from_intake_data(data)
...
+        untimely_exemption: data[:untimely_exemption],
+        untimely_exemption_notes: data[:untimely_exemption_notes]
       ).validate_eligibility!
     end
 
     def find_active_by_reference_id(reference_id)
-      request_issue = unscoped.find_by(rating_issue_reference_id: reference_id, removed_at: nil, ineligible_reason: nil)
+      request_issue = RequestIssue.where(
+        rating_issue_reference_id: reference_id, removed_at: nil, ineligible_reason: nil
```
shanear: I don't think `removed_at` is the proper way to search for an unremoved issue. Removed at is only set when the contention is removed, not when the user says to remove it. This is confusing and should probably be fixed, but it's when the `review_request_id` is nil.

client/app/intake/actions/addIssues.js
```diff
@@ -38,11 +43,7 @@ export const addUnidentifiedIssue = (description, notes) => (dispatch) => {
 };
 
 export const addRatedIssue = (args) => (dispatch) => {
-  let currentRating = _.filter(
-    args.ratings,
-    (ratingDate) => _.some(ratingDate.issues, { reference_id: args.issueId })
-  )[0];
-  let currentIssue = currentRating.issues[args.issueId];
+  const currentIssue = issueById(args.ratings, args.issueId);
```
shanear: nice function!

app/models/request_issue.rb
pkarman: @shanear did you mean "when the `review_request_id` is *not* nil"?

app/models/request_issue.rb
leikkisa: This line just got highlighted because I added the comma and it's in the database from a previous PR (so not going to change it here). But will keep that in mind.

client/app/intake/components/AddIssuesModal.jsx
```diff
@@ -32,14 +37,30 @@ class AddIssuesModal extends React.Component {
...
+    if (this.requiresUntimelyExemption()) {
+      const currentIssue = issueById(this.props.intakeData.ratings, this.state.referenceId);
+
+      this.props.toggleUntimelyExemptionModal({ currentIssue,
+        notes: this.state.notes });
+    } else {
+      this.props.addRatingRequestIssue({
+        issueId: this.state.referenceId,
+        ratings: this.props.intakeData.ratings,
+        isRated: true,
```
pkarman: we need to rename this `isRating`

client/app/intake/components/UntimelyExemptionModal.jsx
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  onAddIssue = () => {
+    const currentIssue = this.props.intakeData.currentIssueAndNotes.currentIssue;
+    const notes = this.props.intakeData.currentIssueAndNotes.notes;
+
+    this.props.addRatingRequestIssue({
+      issueId: currentIssue.reference_id,
+      ratings: this.props.intakeData.ratings,
+      isRated: true,
```
pkarman: should be `isRating`

PR #7603, opened 2018-10-30
Stop creating duplicate BvaDispatchTasks

lowellrex:
Currently we are creating BvaDispatchTasks every time a QualityReviewTask is marked complete. Doing so creates duplicate BvaDispatchTasks when QualityReviewTasks are assigned to members of the QR team as well as the organizational QualityReviewTask. This PR fixes that.

Related slack conversation: https://dsva.slack.com/archives/C6E41RE92/p1540909625107100


mdbenjam:
Ahhh shoot. I think that was my fault. Good call.

PR #7609, opened 2018-10-30
Schedule | Fix Daily docket notes bug 

### LINKED ISSUE #7574: Daily Docket | Notes field bug
Currently when we edit the notes field, we save the edits on a field called editedNotes. The value of the notes field is editedNotes if it's set, and notes if it's not set. The problem is that the user is no longer able to delete the notes field because editedNotes isn't displayed when it's empty. We need to display editedNotes when editedNotes is set to ''.

Sjones352:
Resolves #7574

### Description
This PR fixes the notes field bug and now user can edit and save notes.

### Acceptance Criteria
- [ ] Currently when we edit the notes field, we save the edits on a field called editedNotes. The value of the notes field is editedNotes if it's set, and notes if it's not set. The problem is that the user is no longer able to delete the notes field because editedNotes isn't displayed when it's empty. We need to display editedNotes when editedNotes is set to ''.

### Testing Plan
1. Go to daily docket page in schedule.
screen shot 2018-10-30 at 12 48 05 pm


client/app/hearingSchedule/components/DailyDocket.jsx
```diff
@@ -186,11 +186,20 @@ export default class DailyDocket extends React.Component {
   };
 
   getNotesField = (hearing) => {
+    let noteToDisplay;
+    const notes = !_.isUndefined(hearing.editedNotes);
```
sharonwarner: I think that this logic is simple enough to include directly in the if statement. So, if (!_.isUndefined(hearing.editedNotes)) { ... }

I think that's a little bit cleaner because the variable name `notes` is pretty ambiguous.

sharonwarner: You also could use the ternary operator So something like this: `_.isUndefined(hearing.editedNotes) ? hearing.notes : hearing.editedNotes || '' `

PR #7622, opened 2018-10-30
Replace nil docket numbers with label

### LINKED ISSUE #7610: When docket number is missing from VACOLS, case details should still be clickable
Don't know why appeals would be missing docket numbers! Here's what it looks like in prod: 

image

Go to https://appeals.cf.ds.va.gov/cases/122055

# AC
- if an appeal has no docket number, a link should say "No docket number"

lowellrex:
Resolves #7610.

Mirrors [what we do for AMA appeals](https://github.com/department-of-veterans-affairs/caseflow/blob/4ed9d4c43259c86c0429fba102d172a6aed8ade1/app/models/appeal.rb#L149) if we do not have a docket number.


amprokop:
Hmm approved but I also wonder if this is the best place to do this- would it be better to do in the reducers on the frontend? Im not sure if other applications want this behavior

PR #7624, opened 2018-10-30
change ep station to 397 for supplemental claims with dta errors

jiahuang:
connects #6213


app/models/supplemental_claim.rb
```diff
@@ -52,6 +52,10 @@ def end_product_created_by
     is_dta_error? ? User.system_user : intake_processed_by
   end
 
+  def end_product_station
+    is_dta_error? ? "397" : super
```
shanear: solid

Issue #7587, opened 2018-10-29
AMA Edit | Timely issues showing as untimely


shanear:
I created a higher level review with two rating issues, which worked correctly. When I went to edit those issues, they registered as untimely. This is a bug.

030 HLR on 123458315 in UAT

Here's a screenshot of the edit page:
![screen shot 2018-10-29 at 12 12 05 pm](https://user-images.githubusercontent.com/1596259/47664576-b474a500-db75-11e8-90a8-c09cc6852829.png)

shanear:
this is working!

PR #7626, opened 2018-10-30
Ineligibility reasons not showing up correctly on Edit page

leikkisa:
connects #7587

Originally this bug was just for ineligibility due to timeliness on the edit page, but I noticed that it would affect the other ineligibility reasons as well so there is code for those too.

### DIFF
--- a/app/models/request_issue.rb
+++ b/app/models/request_issue.rb
@@ -74,6 +74,12 @@ class RequestIssue < ApplicationRecord
     review_request_type.try(:constantize).try(:review_title)
   end
 
+  def in_active_review
+    return unless rating_issue_reference_id
+    request_issue = self.class.find_active_by_reference_id(rating_issue_reference_id)
+    request_issue.review_title if request_issue
+  end
+
   def eligible?
     ineligible_reason.nil?
   end
@@ -86,7 +92,9 @@ class RequestIssue < ApplicationRecord
       decision_date: decision_date,
       category: issue_category,
       notes: notes,
-      is_unidentified: is_unidentified
+      is_unidentified: is_unidentified,
+      ineligible_reason: ineligible_reason,
+      in_active_review: in_active_review
     }
   end
 
--- a/client/app/intake/pages/addIssues.jsx
+++ b/client/app/intake/pages/addIssues.jsx
@@ -47,9 +47,11 @@ export class AddIssuesPage extends React.Component {
   checkIfEligible = (issue, formType) => {
     if (issue.isUnidentified) {
       return false;
-    } else if (issue.inActiveReview) {
-      return INELIGIBLE_REQUEST_ISSUES.in_active_review.replace('{review_title}', issue.inActiveReview);
-    } else if (!issue.timely && formType !== 'supplemental_claim') {
+    } else if (issue.inActiveReview || issue.ineligibleReason === 'duplicate_of_issue_in_active_review') {
+      return INELIGIBLE_REQUEST_ISSUES.duplicate_of_issue_in_active_review.replace('{review_title}', issue.inActiveReview);
+    } else if (issue.ineligibleReason) {
+      return INELIGIBLE_REQUEST_ISSUES[ineligibleReason];
+    } else if (issue.timely === false && formType !== 'supplemental_claim') {
       return INELIGIBLE_REQUEST_ISSUES.untimely;
     } else if (issue.sourceHigherLevelReview && formType === 'higher_level_review') {
       return INELIGIBLE_REQUEST_ISSUES.previous_higher_level_review;
--- a/client/app/intake/util/issues.js
+++ b/client/app/intake/util/issues.js
@@ -79,12 +79,14 @@ export const validNonRatedIssue = (issue) => {
 
 export const formatRequestIssues = (requestIssues) => {
   return requestIssues.map((issue) => {
+    console.log("issue::", issue)
     if (issue.category) {
       return {
         isRated: false,
         category: issue.category,
         description: issue.description,
-        decisionDate: issue.decision_date
+        decisionDate: issue.decision_date,
+        ineligibleReason: issue.ineligible_reason
       };
     }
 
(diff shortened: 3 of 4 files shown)


app/models/request_issue.rb
```diff
@@ -74,6 +74,17 @@ def review_title
     review_request_type.try(:constantize).try(:review_title)
   end
 
+  def in_active_review
```
pkarman: As you mentioned elsewhere, #7576 will make this much easier.

```ruby
ineligible_due_to.try(:review_title)
```

Also, the method naming is somewhat misleading. This `RequestIssue` is not in active review. It is an ineligible duplicate of a different `RequestIssue` that is in active review. That's why the `ineligible_reason` is named `duplicate_of_issue_in_active_review`.

app/models/request_issue.rb
```diff
@@ -86,7 +97,9 @@ def ui_hash
       decision_date: decision_date,
       category: issue_category,
       notes: notes,
-      is_unidentified: is_unidentified
+      is_unidentified: is_unidentified,
+      ineligible_reason: ineligible_reason,
+      in_active_review: in_active_review
```
pkarman: I would change this to (using @shanear favorite ternary pattern):

```ruby
duplicate_of_issue_in_active_review: duplicate_of_issue_in_active_review? ? ineligible_due_to.review_title : nil
```

client/app/intake/pages/addIssues.jsx
```diff
@@ -48,8 +48,12 @@ export class AddIssuesPage extends React.Component {
     if (issue.isUnidentified) {
       return false;
     } else if (issue.inActiveReview) {
```
pkarman: same comment about naming. Instead: `duplicateOfIssueInActiveReview`

client/app/intake/util/issues.js
```diff
@@ -105,7 +106,9 @@ export const formatRequestIssues = (requestIssues) => {
       id: issue.reference_id,
       profileDate: issueDate.toISOString(),
       notes: issue.notes,
-      description: issue.description
+      description: issue.description,
+      ineligibleReason: issue.ineligible_reason,
+      inActiveReview: issue.in_active_review
```
pkarman: `duplicateOfIssueInActiveReview: issue.duplicate_of_issue_in_active_review`

client/constants/INELIGIBLE_REQUEST_ISSUES.json
```diff
@@ -1,5 +1,5 @@
 {
-  "in_active_review": "is ineligible because it's already under review as a {review_title}",
+  "duplicate_of_issue_in_active_review": "is ineligible because it's already under review as a {review_title}",
```
pkarman: I can go either way on this one, but you're right, probably best to make it consistent.

spec/feature/intake/edit_spec.rb
```diff
@@ -83,6 +93,93 @@ def check_row(label, text)
       )
     end
 
+    context "when there are ineligible issues" do
```
pkarman: great test!

spec/models/request_issue_spec.rb
```diff
@@ -146,6 +146,37 @@
...
+      )
+    end
+
+    let!(:ineligible_request_issue) do
+      create(
+        :request_issue,
+        review_request: new_higher_level_review,
+        rating_issue_reference_id: higher_level_review_reference_id,
+        contention_reference_id: contention_reference_id,
+        ineligible_reason: 0
```
pkarman: I think it's best to use the `:duplicate_of_issue_in_active_review` symbol rather than the integer value.

spec/models/request_issue_spec.rb
```diff
@@ -146,6 +146,37 @@
     end
   end
 
+  context "#in_active_review" do
+    let(:previous_higher_level_review) { create(:higher_level_review, id: 10) }
```
pkarman: setting the `id` explicitly is very unusual. why here?

spec/models/request_issue_spec.rb
```diff
@@ -146,6 +146,37 @@
     end
   end
 
+  context "#in_active_review" do
```
pkarman: I think the `in_active_review` method can go away entirely if you use `ineligible_due_to` instead.

pkarman: Maybe parts of this could be re-used for a `ui_hash` test?

client/app/intake/actions/addIssues.js
```diff
@@ -49,7 +49,7 @@ export const addRatedIssue = (args) => (dispatch) => {
     payload: {
       id: args.issueId,
       isRated: args.isRated,
-      inActiveReview: currentIssue.in_active_review,
```
pkarman: I see the dilemma here. We're trying to use the same client code to manage 2 different kinds of models: rating request issues (contentions), and rating issues (decisions).

I think it is vital to maintain the clarity of naming on the server side, because those models represent 2 different aspects of the business process (the VA side (the decision) and the Vet side (the contention)).

What about if we did:

```javascript
inActiveReview: currentIssue.in_active_review || currentIssue.duplicate_of_issue_in_active_review
```

which maintains the clarity of naming on the model, but allows the client side to present the visual warning as needed.

client/constants/INELIGIBLE_REQUEST_ISSUES.json
pkarman: I would leave this as `in_active_review` then because it is for the client side visuals only and that is the key describing the *error* not the status of the model.

app/models/rating_issue.rb
```diff
@@ -50,12 +50,12 @@ def ui_hash
       decision_text: decision_text,
       promulgation_date: promulgation_date,
       contention_reference_id: contention_reference_id,
-      in_active_review: in_active_review,
+      active_review_of_duplicate_issue: active_review_of_duplicate_issue,
```
pkarman: The rating issue is not the duplicate. It's the rating request issue that is the duplicate.

PR #7633, opened 2018-10-31
Improve the appearance of the organizational queue if there are no cases

lowellrex:
| Before | After |
| --- | --- |
| image | image |


sneha-pai:
This looks great. Can we keep the body copy unbolded, and only bold the hyperlink? 
( I'm referring to the "You have no cases assigned to you. You can **search for cases**" )

lowellrex:
@sneha-pai ticketed #7643!

PR #7646, opened 2018-10-31
Only show tasks with available actions

mdbenjam:
### Description
This PR does three things:

1) Makes case snapshot pull the first task that has available actions
2) Make all flows take a task parameter
3) Fixes reassign action on case details for judges.


app/models/legacy_tasks/judge_legacy_task.rb
```diff
@@ -22,7 +22,7 @@ def available_actions(role)
       [
         {
           label: COPY::JUDGE_CHECKOUT_ASSIGN_TO_ATTORNEY_LABEL,
-          value: "assign"
+          value: "modal/assign_to_attorney"
```
mdbenjam: This was wrong.

app/models/tasks/attorney_task.rb
```diff
@@ -7,6 +7,15 @@ class AttorneyTask < Task
   validate :parent_attorney_child_count, on: :create
 
   def available_actions(user)
+    if parent.type == JudgeTask.name && parent.assigned_to == user
```
mdbenjam: Judges can reassign attorney tasks.

mdbenjam: In the future we should make this a separate reassign action... or even better, align it with the generic assign/reassign.

client/app/queue/CaseDetailsLoadingScreen.jsx
```diff
@@ -57,22 +55,19 @@ class CaseDetailLoadingScreen extends React.PureComponent<Props> {
...
-            tasks: prepareLegacyTasksForStore(response.body.tasks) });
-        } else {
-          this.props.onReceiveTasks({ tasks: {},
-            amaTasks: prepareTasksForStore(response.body.tasks) });
-        }
-      });
-
-      promises.push(taskPromise);
-    }
+    const taskPromise = ApiUtil.get(`/appeals/${appealId}/tasks?role=${userRole}`).then((response) => {
```
mdbenjam: Always try to get new tasks, since the tasks may have changed on the backend.

client/app/queue/components/AssignWidget.jsx
```diff
@@ -115,9 +120,14 @@ class AssignWidget extends React.PureComponent<Props> {
       }, () => {
         this.props.resetSaveState();
 
+        const errorDetail = this.props.isModal && userId ?
```
mdbenjam: We know reassign is broken for legacy tasks, so we link folks to the assign page where they can use the checkboxes to reassign

mdbenjam: https://dsva.slack.com/archives/C6E41RE92/p1541088357243800

app/models/tasks/attorney_task.rb
```diff
@@ -7,6 +7,15 @@ class AttorneyTask < Task
...
+      return [
+        {
+          label: COPY::JUDGE_CHECKOUT_ASSIGN_TO_ATTORNEY_LABEL,
+          value: "modal/assign_to_attorney"
+        }
```
lowellrex: Can we move this into TASK_ACTIONS.json?

client/app/queue/AddColocatedTaskView.jsx
```diff
@@ -64,41 +65,41 @@ class AddColocatedTaskView extends React.PureComponent<Props, ComponentState> {
...
       };
 
       if (!appeal.isLegacyAppeal) {
-        mapped.parent_id = task.taskId;
+        mapped.parent_id = singleTask.taskId;
       }
 
       return mapped;
     });
   }
```
lowellrex: I know this isn't new with this PR, but I find this function very confusing. It seems to me all `buildPayload()` is doing is the following:
```jsx
buildPayload = () => {
  const { task, appeal } = this.props;
  return {
    ...this.state,
    type: 'ColocatedTask',
    external_id: appeal.externalId,
    ...(!appeal.isLegacyAppeal && {parent_id: task.taskId})
  };
}
```

Am I missing something else this function is doing?

client/app/queue/QueueApp.jsx
```diff
@@ -128,27 +128,34 @@ class QueueApp extends React.PureComponent<Props> {
...
     checkoutFlow={props.match.params.checkoutFlow} />;
 
   routedSelectSpecialIssues = (props) => <SpecialIssueLoadingScreen appealExternalId={props.match.params.appealId}>
     <SelectSpecialIssuesView
       appealId={props.match.params.appealId}
+      taskId={props.match.params.taskId}
       prevStep={`/queue/appeals/${props.match.params.appealId}`}
-      nextStep={`/queue/appeals/${props.match.params.appealId}/${props.match.params.checkoutFlow}/dispositions`} />
+      nextStep={`/queue/appeals/${props.match.params.appealId}/tasks/${props.match.params.taskId}` +
+        `/${props.match.params.checkoutFlow}/dispositions`} />
```
lowellrex: For readability purposes it might make sense to do something like:
```jsx
routedSelectSpecialIssues = (props) => {
  const {
    appealId,
    checkoutFlow,
    taskId
  } = props.match.params;

  return <SpecialIssueLoadingScreen...
```

client/app/queue/SubmitDecisionView.jsx
```diff
@@ -228,7 +230,7 @@ const mapStateToProps = (state, ownProps) => {
     appeal,
     judges,
     // Attorneys should only have one task assigned to them from this appeal.
```
lowellrex: I think we can ditch this comment now.

client/app/queue/AddColocatedTaskView.jsx
mdbenjam: As in why is there a map... yeah idk.

PR #7649, opened 2018-10-31
Only show available_actions for active tasks

lowellrex:
Limited in scope to only GenericTasks and ColocatedTasks for now since we expect those to be the most frequent occurrences. GenericTasks will be assigned through organizations which should not have available actions when there are active child tasks (this PR fixes this situation), and ColocatedTasks should not be actionable once they have been returned to the assigning attorney even though colocated staffers will still be able to view the appeals to which they are connected.

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -31,9 +31,15 @@ class Task < ApplicationRecord
   # from TASK_ACTIONS that looks something like:
   # [ { "label": "Assign to person", "value": "modal/assign_to_person", "func": "assignable_users" }, ... ]
   def available_actions_unwrapper(user)
+    return [] if no_actions_available(user)
+
     available_actions(user).map do |a|
       { label: a[:label], value: a[:value], data: a[:func] ? send(a[:func]) : nil }
     end
+  end
+
+  def no_actions_available(_user)
+    [Constants.TASK_STATUSES.on_hold, Constants.TASK_STATUSES.completed].include?(status)
   end
 
   def assigned_by_display_name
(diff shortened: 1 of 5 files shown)


app/models/tasks/generic_task.rb
```diff
@@ -1,6 +1,9 @@
 class GenericTask < Task
   # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/CyclomaticComplexity
   def available_actions(user)
+    return [] if [Constants.TASK_STATUSES.on_hold, Constants.TASK_STATUSES.completed].include?(status)
```
mdbenjam: I think we need to encapsulate this logic outside of available actions so we don't have to reimplement it in every inheriting class. Some method that checks this condition then if it's not true calls available actions.

app/models/task.rb
```diff
@@ -31,11 +31,17 @@ def available_actions(_user)
...
+    return [] if no_actions_available(user)
+
     available_actions(user).map do |a|
       { label: a[:label], value: a[:value], data: a[:func] ? send(a[:func]) : nil }
     end
   end
 
+  def no_actions_available(_user)
+    [Constants.TASK_STATUSES.on_hold, Constants.TASK_STATUSES.completed].include?(status)
+  end
```
lowellrex: I don't love this function name, so if you have an idea for a better one I'm all ears.

mdbenjam: Maybe has_allowed_actions?

Issue #7594, opened 2018-10-29
AMA Edit | Create appeal edit link and modify save to update appeal issues as well


shanear:
We have a URL that directs a user to the edit screen in Caseflow for supplemental claims and higher level reviews, but 

## Technical Notes
- This will likely require changes to `RequestIssuesUpdate` in order to make it work for updates to appeals, which do not have EPs

## Acceptance Criteria
- For any appeal, there should be a url `/appeals/[appeal_uuid]/edit` that takes the user to the edit issues page for that appeal.
- When save is clicked, it should redirect the user to the claim details page for that appeal `/queue/appeals/[appeal_uuid]`

## Notes
- This url will be used in queue to direct attorneys to the edit page, to allow them to correct issues.
- There are 2 appeals right now that need corrections, so this is fairly high priority

allyceh:
@laurjpeterson @shanear @amprokop 

I went through this Mural of the Intake flow (https://app.mural.co/t/workqueue2001/m/workqueue2001/1540909080413/09248d7c6c7c77d1b5b21867b8573ce503e71e6b) and a sketch file that contains the more of the validation scenarios and messaging. Below, I've pulled out screens that I feel we can better tailor to attorneys who will be correcting the issues on appeal. Tweaks include adjustments to copy and removing data from the Change Issues page. 

A couple things I'll need: 
- Copy review: from the perspective of an attorney, does this copy make sense? What feedback do you have so that we can make things even clearer?
- Did I miss any key screens or scenarios from the Mural or what's in UAT for SC/HLR that will also apply to Board appeals and that should have copy tweaks? If so, please let me know what you think of.

Other notes: 
- I did not make copy tweaks for the confirmation screens that appear after Intake/attorney saves an appeal. @laurjpeterson , @shanear and I decided to eliminate them in favor of bringing the attorney directly back to the Case Details screen where they can see the changes they just made. 

![image](https://user-images.githubusercontent.com/40216950/47751418-14e40f00-dc68-11e8-880b-55c8f7fd6625.png)
![image](https://user-images.githubusercontent.com/40216950/47751646-c1be8c00-dc68-11e8-8f88-d4b380a00921.png)
![image](https://user-images.githubusercontent.com/40216950/47751462-347b3780-dc68-11e8-9970-b5a50dfb3589.png)
![image](https://user-images.githubusercontent.com/40216950/47751483-3f35cc80-dc68-11e8-82ae-6c6fe858dc63.png)
[...]

PR #7653, opened 2018-10-31
adding edit page to appeals

jiahuang:
Connects #7594 

appeal_edit1

To do:
- [x] add tests


app/controllers/appeals_controller.rb
```diff
@@ -67,8 +67,36 @@ def show
     end
   end
 
+  helper_method :appeal, :url_appeal_uuid
+
+  def appeal
+    @appeal ||= Appeal.find_appeal_by_id_or_find_or_create_legacy_appeal_by_vacols_id(params[:id])
```
shanear: lol `find_appeal_by_id_or_find_or_create_legacy_appeal_by_vacols_id` is a hell of a method name

app/models/appeal.rb
```diff
@@ -32,6 +32,13 @@ def self.find_appeal_by_id_or_find_or_create_legacy_appeal_by_vacols_id(id)
     end
   end
 
+  def ui_hash(ama_enabled)
+    super.merge(
+      docketType: docket_type,
```
shanear: shouldn't these be snake_case per our discussion :)

spec/feature/intake/edit_spec.rb
```diff
@@ -43,6 +45,54 @@ def check_row(label, text)
...
+      # going back to edit page should show those issues
+      visit "appeals/#{appeal.uuid}/edit/"
+      expect(page).to have_content("Left knee granted")
+      expect(page).not_to have_content("nonrating description")
+
+      # canceling should redirect to queue
+      click_on "Cancel edit"
+      expect(page).to have_current_path("/queue/appeals/#{appeal.uuid}")
+    end
+  end
```
shanear: nice test. Love the comments

app/models/request_issues_update.rb
```diff
@@ -38,14 +38,18 @@ def perform!
   def process_end_product_establishments!
     attempted!
 
-    review.process_end_product_establishments!
+    # appeals do not have process_end_product_establishments
+    if review.respond_to?(:process_end_product_establishments!)
```
shanear: I'm thinking we shouldn't even do this async stuff for appeals. Can we just do `attempted!` and `processed!` synchronously for appeals and guard this whole method on `return unless review.respond_to?(:process_end_product_establishments!)`?

app/models/decision_review_intake.rb
```diff
@@ -19,4 +19,21 @@ def cancel_detail!
   def review_errors
     detail.errors.messages
   end
+
+  def complete!(request_params)
```
shanear: I don't think it's super clear what conceptually belongs in the yield block here. Maybe take a block through the `&block` way  (see: https://mixandgo.com/learn/mastering-ruby-blocks-in-less-than-5-minutes) and name it something more descriptive? (like `&inside_transaction`) Or add a comment?

app/models/appeal.rb
leikkisa: 👍

PR #7655, opened 2018-10-31
make stdout sync so we can see security rake task output

kencheeto:
### Description
Like https://github.com/department-of-veterans-affairs/caseflow/pull/7324, but for the security rake task

### Acceptance Criteria
- [ ] Security rake task failures show up in CI

### Testing Plan
N/A


kencheeto:
frustrating that this check is passing on a known bad commit. Could be CircleCI caching or something..?

PR #7664, opened 2018-11-01
Add check for whether an issue is after the AMA Activation Date

leikkisa:
connects #7309 

Testing plan
- Add an issue that came from a RAMP decision, but is before the AMA_ACTIVATION_DATE, and it gets saved as an eligible issue
- Add an issue that didn't come from a RAMP decision, and is before the AMA_ACTIVATION_DATE, and it gets saved as an ineligible issue
- Test all AMA flows
- Test both intake and edit flows
- Test both rating and nonrating issues


app/models/decision_review.rb
```diff
@@ -18,6 +16,13 @@ class DecisionReview < ApplicationRecord
     ratings_with_issues.map(&:ui_hash)
   end
 
+  AMA_ACTIVATION_DATE =
+    if use_ama_activation_date?
```
shanear: we probably want this inside of a function, as this feature flag will be evaluated on deploy and never again.

app/models/rating_issue.rb
```diff
@@ -4,19 +4,24 @@
...
-  attr_accessor :reference_id, :decision_text, :profile_date,
+  attr_accessor :reference_id, :decision_text, :profile_date, :ramp_claim_id,
                 :promulgation_date, :participant_id, :contention_reference_id
 
   class << self
     def from_bgs_hash(data)
       rba_contentions = [data.dig(:rba_issue_contentions) || {}].flatten
+      associated_claims = [data.dig(:associated_claims) || {}].flatten
+      is_ramp_decision = EndProduct::RAMP_CODES.key?(associated_claims.first.dig(:bnft_clm_tc))
+      associated_claim_id = associated_claims.first.dig(:clm_id)
```
shanear: instead of doing these calculations here, it might be better factored by passing data in a more raw state into the class, and let getter methods work this out.

P.S. then you could have a getter method named `from_ramp_decision?` 🙂

app/models/request_issue.rb
```diff
@@ -147,6 +151,25 @@ def check_for_previous_review!(review_type)
     end
   end
 
+  def check_for_before_ama!
+    return unless eligible?
+    check_for_rating_before_ama! if rating?
```
jiahuang: how does this work for unidentified issues? They're created as rating issues now but `rating_issue_profile_date` is nil I think. Would the check always fail?

app/models/request_issue.rb
```diff
@@ -147,6 +151,25 @@ def check_for_previous_review!(review_type)
...
+    if rating_issue_profile_date < DecisionReview::AMA_ACTIVATION_DATE
+      self.ineligible_reason = :before_ama
+    end
+  end
+
+  def check_for_nonrating_before_ama!
+    if decision_date < DecisionReview::AMA_ACTIVATION_DATE
+      self.ineligible_reason = :before_ama
+    end
+  end
```
shanear: It seems like we could have one method that gets the decision date for rating and nonrating request issues. That way we don't need `check_for_rating_before_ama!` and `check_for_nonrating_before_ama!` and the same for the timeliness check methods.

app/models/request_issues_update.rb
```diff
@@ -98,8 +98,12 @@ def calculate_after_issues
         decision_date: issue_data[:decision_date],
         issue_category: issue_data[:issue_category],
         notes: issue_data[:notes],
-        is_unidentified: issue_data[:is_unidentified]
+        is_unidentified: issue_data[:is_unidentified],
+        untimely_exemption: issue_data[:untimely_exemption],
+        untimely_exemption_notes: issue_data[:untimely_exemption_notes],
+        ramp_claim_id: issue_data[:ramp_claim_id]
       ).tap do |request_issue|
+        request_issue.validate_eligibility!
```
shanear: interesting, this seems like it's a big change. Why did we add this? (not saying we shouldn't, just curious 🙂)

app/views/appeals/edit.html.erb
```diff
@@ -5,6 +5,9 @@
     feedbackUrl: feedback_url,
     buildDate: build_date,
     serverIntake: appeal.ui_hash,
-    claimId: url_appeal_uuid
+    claimId: url_appeal_uuid,
```
shanear: `claim id = appeal uuid`

this is a little awkward lol.

app/models/request_issue.rb
```diff
@@ -95,6 +97,7 @@ def ui_hash
       category: issue_category,
       notes: notes,
       is_unidentified: is_unidentified,
+      ramp_claim_id: ramp_claim_id,
```
shanear: I feel like we may want a comment somewhere in this class describing the significance of `ramp_claim_id`

Maybe:
```
ramp_claim_id is set when the contested rating issue is part of a ramp decision. It is set to the claim id of the RAMP EP.
```

app/views/appeals/edit.html.erb
```diff
@@ -5,6 +5,9 @@
...
+    featureToggles: {
+      useAmaActivationDate: FeatureToggle.enabled?(:use_ama_activation_date, user: current_user)
```
shanear: Instead of passing the feature toggle, we could just pass `amaActivationDate`, this way it'd save us some logic on the front end

lib/fakes/bgs_service.rb
```diff
@@ -65,6 +66,17 @@ def self.create_veteran_records
...
+          participant_id: veteran.participant_id,
+          profile_date: DecisionReview::AMA_ACTIVATION_DATE - 10.days,
+          promulgation_date: DecisionReview::AMA_ACTIVATION_DATE - 5.days,
+          issues: [
+            { decision_text: "Issue before AMA not from a RAMP Review", reference_id: "before_ama_ref_id" },
+            { decision_text: "Issue before AMA from a RAMP Review",
+              associated_claims: { bnft_clm_tc: "683SCRRRAMP", clm_id: "ramp_claim_id" },
+              reference_id: "ramp_reference_id" }
+          ]
+        )
```
shanear: nice

app/models/request_issue.rb
leikkisa: Good catch! I thought it didn't but the rating category does include unidentified

spec/models/request_issue_spec.rb
```diff
@@ -306,5 +320,32 @@
...
+      context "rating issue is from a RAMP decision" do
+        let(:ramp_claim_id) { "ramp_claim_id" }
+
+        it "does not flag rating issues before AMA from a RAMP decision" do
+          rating_request_issue.rating_issue_reference_id = "ramp_ref_id"
+          rating_request_issue.validate_eligibility!
+          expect(rating_request_issue.ineligible_reason).to be_nil
+        end
+      end
+    end
```
shanear: solid tests

app/models/request_issue.rb
leikkisa: Oh wait, I was looking at the class methods: rating? doesn't include unidentified. So unidentified will just never get this error.

app/models/request_issues_update.rb
leikkisa: There was a bug because the Edit page was being neglected :-P. So if an issue got added on the Edit page, although it could show up as ineligible there, it wouldn't get any ineligible reason in the database. Also other data wasn't being added to the database if an issue was added from the Edit page.

app/views/appeals/edit.html.erb
leikkisa: Haha. This is from a different PR... I don't know too much about the decision making that went into that.

app/models/rating_issue.rb
```diff
@@ -26,10 +28,19 @@ def from_ui_hash(ui_hash)
...
         profile_date: ui_hash[:profile_date],
-        contention_reference_id: ui_hash[:contention_reference_id]
+        associated_claims_data: ui_hash[:associated_claims_data],
+        rba_contentions_data: ui_hash[:rba_contentions_data]
       )
     end
+
+    private
+
+    def ensure_array_of_objects(data)
```
pkarman: everything in Ruby is an object. everything is data. maybe instead this could be named like:

```ruby
def ensure_array_of_hashes(array_or_hash_or_nil)
  [array_or_hash_or_nil || {}].flatten
end
```

app/models/request_issues_update.rb
shanear: Did we add an automated test for this issue?

pkarman: the call to `validate_eligibility!` should happen, but after the set of `profile_date`.

See https://github.com/department-of-veterans-affairs/caseflow/pull/7769/commits/d5c373737d80913dd7b7347c43cd8f0445632e7c

app/models/request_issue.rb
leikkisa: Did this for before_ama but not for timeliness checks because they are based on promulgation_date

app/models/request_issues_update.rb
```diff
@@ -98,10 +98,11 @@ def calculate_after_issues
...
         notes: issue_data[:notes],
-        is_unidentified: issue_data[:is_unidentified]
-      ).tap do |request_issue|
-        request_issue.rating_issue_profile_date ||= issue_data[:profile_date]
-      end
+        is_unidentified: issue_data[:is_unidentified],
+        untimely_exemption: issue_data[:untimely_exemption],
+        untimely_exemption_notes: issue_data[:untimely_exemption_notes],
+        ramp_claim_id: issue_data[:ramp_claim_id]
+      ).tap(&:validate_eligibility!)
```
pkarman: I like this! More explicitly the *last* thing to happen.

app/models/request_issues_update.rb
```diff
@@ -98,10 +98,11 @@ def calculate_after_issues
         decision_date: issue_data[:decision_date],
         issue_category: issue_data[:issue_category],
         notes: issue_data[:notes],
-        is_unidentified: issue_data[:is_unidentified]
-      ).tap do |request_issue|
-        request_issue.rating_issue_profile_date ||= issue_data[:profile_date]
```
leikkisa: This seemed to be due to previously replacing the rating issue profile_date with the rba_contention profile_date, but it should still be based on the rating issue profile date so this should no longer be necessary

app/models/request_issue.rb
shanear: Hmm... I think we'd want them to be based off of the same date. Not sure why they aren't?

spec/feature/intake/appeal_spec.rb
```diff
@@ -43,7 +43,7 @@
 
   let(:untimely_days) { 372.days }
```
pkarman: is `untimely_days` needed any more?

app/models/rating_issue.rb
```diff
@@ -81,8 +93,26 @@ def decision_issue
...
+  def from_ramp_decision?
+    associated_claims_data && EndProduct::RAMP_CODES.key?(associated_claims_data.first.dig(:bnft_clm_tc))
+  end
+
+  def ramp_claim_id
+    @ramp_claim_id ||= calculate_ramp_claim_id
+  end
+
+  def contention_reference_id
+    @contention_reference_id ||= rba_contentions_data.first.dig(:cntntn_id) if rba_contentions_data
```
pkarman: this logic is a bit awkward to read. maybe instead:

```ruby
def contention_reference_id
  return unless rba_contentions_data
  @contention_reference_id ||= rba_contentions_data.first.dig(:cntntn_id)
end
```

spec/models/request_issues_update_spec.rb
```diff
@@ -82,7 +82,7 @@
   let(:request_issues_data_with_new_issue) do
     existing_request_issues_data + [{
       reference_id: "issue3",
-      profile_date: Time.zone.local(2017, 4, 7),
+      profile_date: Time.zone.local(2017, 11, 7),
```
pkarman: since this date is now a significant date, it might help to name it so that a future developer doesn't assume it is Just a Date.

```ruby
let(:ama_start_date) { Time.zone.local(2017, 11, 7) }
...
  profile_date: ama_start_date
```

app/models/request_issue.rb
leikkisa: Ok am doin it!

Issue #7450, opened 2018-10-18
Discussion: Naming conventions


leikkisa:
Let's discuss coming up with a standard for naming non-rating issues. Some known variations:
non-rating
nonrating
non-rated (nonrated?)

There may be other naming conventions we'd like to standardize so this discussion is open to those as well.

Comment in the ticket, and let's plan to discuss in a week.

shanear:
Since the EP is called "Nonrating", I'm thinking we either want that, or have it hyphenated (i.e. "Non-rating"). I definitely don't think it should be non-rated or nonrated.

Also:

1) Use "AMA" as little as possible in the codebase, this will future proof the codebase, and reduce the number of acronyms used.

2) Decision review: Term used to refer to an AMA appeal, HLR, or SC

leikkisa:
Decision
rating issue - the issue we get from BGS


request issues - the issue we generate during a Decision Review in Caseflow
- rating request issue
- nonrating request issue (no hyphens)
- unidentified request issue

Word for AMA reviews (Appeal/NOD, Higher-level Review, Supplemental Claim)
- Decision Review
This is used in the world

The reviews that create an End Product, only compensation and pension (Higher-level Review, Supplemental Claim)
- Claim Review
This is primarily used internally in the code

Other business lines do not create EPs (e.g. Education), but could create a Higher-level Review or Supplemental Claim.  For now we'll keep calling them claim reviews.

Request issue chain
If an issue is part of a chain of issues, what the user is actually appealing is the last decision in the chain.
HLR can't be followed by an HLR
Appeal can't be followed by an HLR
SC can follow anything?

The request issue's relationship to the ....? Currently related_request_issue
A (Original decision) -> B (SC request issue) -> C (decision issue/rating issue) -> E (HLR request issue)
                                                                            -> D (decision issue/rating issue)
Previous -> Current -> Next

shanear:
Thoughts on refactoring stories needed:

- Non-rated, Non-rating -> Nonrating
- use rating_request_issue, nonrating_request_issue, and unidentified_request_issue where appropriate
- AMAReview -> DecisionReview
- Create decision_issues and decision_rating_issues for HLRs and SCs (will probably need to wait until we know more about how decision issues should work for appeals and if we want to combine them?)

shanear:
https://app.mural.co/t/workqueue2001/m/workqueue2001/1540421981879/ba39075927b4039387697e094436e6a236fe7fa4

PR #7672, opened 2018-11-01
Rename rated rating

pkarman:
connects #7450

### Description
Rename variables per Intake team decision.

NOTE that user-facing text has been left as-is.

### Acceptance Criteria
- [x] Code compiles correctly
- [x] Tests continue to pass


app/models/claim_review.rb
```diff
@@ -76,7 +76,7 @@ def process_end_product_establishments!
     end_product_establishments.each do |end_product_establishment|
       end_product_establishment.perform!
       end_product_establishment.create_contentions!
-      end_product_establishment.associate_rated_issues!
+      end_product_establishment.associate_rating_request_issues!
```
shanear: I think this really is talking about a `rating_issue`

app/models/end_product_establishment.rb
```diff
@@ -187,17 +187,17 @@ def status_active?(sync: false)
     !EndProduct::INACTIVE_STATUSES.include?(synced_status)
   end
 
-  def associate_rated_issues!
-    is_rated = true
-    return if code != source.issue_code(is_rated)
-    return if unassociated_rated_request_issues.count == 0
+  def associate_rating_request_issues!
+    is_rating = true
+    return if code != source.issue_code(is_rating)
```
shanear: not related to this PR, but it's very awkward that we hardcode a variable to true, and the only place we use that variable is the line right after 🙃

pkarman: Kara and I did that to make it more obvious what the single arg indicated. it was non-obvious looking at it

alternately, we could name it `source.issue_code(is_rating: true)`

shanear: Yeah I like that way, even more I like:

`source.issue_code(:rating)` or `source.issue_code(:nonrating)`

.circleci/config.yml
```diff
@@ -262,6 +262,7 @@ jobs:
           path: ~/all_logs.log
 
 workflows:
+  version: 2
```
jiahuang: wait didn't this get removed by Ken's pr here: https://github.com/department-of-veterans-affairs/caseflow/commit/45d39d7c8f7e36fb1e93117921e5cc9a0896f83a

kencheeto: yeah I think that was an attempt to debug the build: https://circleci.com/gh/department-of-veterans-affairs/caseflow/19186

I was following the advice here: https://github.com/CircleCI-Public/config-preview-sdk/blob/master/docs/whats-new.md#whats-new-in-21-configuration

(current docs say that it's required: https://circleci.com/docs/2.0/configuration-reference/#version-1)

It doesn't seem to hurt to leave the version in, and it seems that people get inconsistent results without it, so I think it's fine to re-add here.

PR #7677, opened 2018-11-01
Hide ineligible issues on case details page

### LINKED ISSUE #7625: Filter ineligible issues from Queue
In Intake, we save ineligible issues on appeals to keep a record of how often veterans request ineligible issues. However, these issues should not appear in the rest of queue. The should also not require decisions

lowellrex:
Resolves #7625. Hides ineligible issues on the case details page.

### DIFF
--- a/client/app/queue/components/CaseDetailsIssueList.jsx
+++ b/client/app/queue/components/CaseDetailsIssueList.jsx
@@ -25,7 +25,7 @@ const headingStyling = css({
 
 export default function CaseDetailsIssueList(props) {
   return <React.Fragment>
-    {props.issues.map((issue, i) =>
+    {props.issues.filter((iss) => iss.is_eligible).map((issue, i) =>
       <div key={i} {...singleIssueContainerStyling}>
         <h3 {...headingStyling}>Issue {1 + i}</h3>
         { props.isLegacyAppeal ?
--- a/spec/feature/queue/case_details_spec.rb
+++ b/spec/feature/queue/case_details_spec.rb
@@ -273,6 +273,26 @@ RSpec.feature "Case details" do
     end
   end
 
+  context "when an appeal has an issue that is ineligble" do
+    let(:eligble_issue_cnt) { 5 }
+    let(:ineligble_issue_cnt) { 3 }
+    let(:issues) do
+      [
+        build_list(:request_issue, eligble_issue_cnt, description: "Knee pain"),
+        build_list(:request_issue, ineligble_issue_cnt, description: "Sunburn", ineligible_reason: "untimely")
+      ].flatten
+    end
+    let!(:appeal) { FactoryBot.create(:appeal, request_issues: issues) }
+
+    scenario "only eligble issues should appear in case details page" do
+      visit "/queue/appeals/#{appeal.uuid}"
+
+      # byebug
+      expect(page).to have_content("Issue #{eligble_issue_cnt}")
+      expect(page).to_not have_content("Issue #{eligble_issue_cnt + 1}")
+    end
+  end
+
   context "loads judge task detail views" do
     let!(:vacols_case) do
       FactoryBot.create(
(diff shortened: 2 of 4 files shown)


client/app/queue/components/CaseDetailsIssueList.jsx
```diff
@@ -25,7 +25,7 @@ const headingStyling = css({
 
 export default function CaseDetailsIssueList(props) {
   return <React.Fragment>
-    {props.issues.map((issue, i) =>
+    {props.issues.filter((iss) => iss.is_eligible).map((issue, i) =>
```
lowellrex: This is the heart of the change on the frontend.

spec/feature/queue/case_details_spec.rb
```diff
@@ -273,6 +273,25 @@
     end
   end
 
+  context "when an appeal has an issue that is ineligble" do
+    let(:eligble_issue_cnt) { 5 }
+    let(:ineligble_issue_cnt) { 3 }
+    let(:issues) do
+      [
+        build_list(:request_issue, eligble_issue_cnt, description: "Knee pain"),
+        build_list(:request_issue, ineligble_issue_cnt, description: "Sunburn", ineligible_reason: "untimely")
```
pkarman: one nit: I think `:untimely` instead of "untimely" is the preferred Rails enum convention

spec/feature/queue/case_details_spec.rb
```diff
@@ -273,6 +273,25 @@
...
+    let(:ineligble_issue_cnt) { 3 }
+    let(:issues) do
+      [
+        build_list(:request_issue, eligble_issue_cnt, description: "Knee pain"),
+        build_list(:request_issue, ineligble_issue_cnt, description: "Sunburn", ineligible_reason: :untimely)
+      ].flatten
+    end
+    let!(:appeal) { FactoryBot.create(:appeal, request_issues: issues) }
+
+    scenario "only eligble issues should appear in case details page" do
```
amprokop: nit: eligible

spec/feature/queue/case_details_spec.rb
```diff
@@ -273,6 +273,25 @@
     end
   end
 
+  context "when an appeal has an issue that is ineligble" do
```
amprokop: nit: ineligible

spec/feature/queue/case_details_spec.rb
```diff
@@ -273,6 +273,25 @@
...
+    let(:eligble_issue_cnt) { 5 }
+    let(:ineligble_issue_cnt) { 3 }
+    let(:issues) do
+      [
+        build_list(:request_issue, eligble_issue_cnt, description: "Knee pain"),
```
amprokop: nit: eligible

spec/feature/queue/case_details_spec.rb
```diff
@@ -273,6 +273,25 @@
...
+        build_list(:request_issue, ineligble_issue_cnt, description: "Sunburn", ineligible_reason: :untimely)
```
amprokop: nit: ineligible

amprokop:
Is there any known context where the frontend would need these issues? To save frontend complexity would it be simpler to just not send these issues? 

Also, don't we count the issues in some places? If so, filtering on the backend would save us from future bugs

pkarman:
@amprokop on the intake side we send all the issues and then have a (ever increasingly complex) mechanism for flagging what is ineligible and why. Not sure what the desired UX is on the queue side, but it might be misleading for users if they create things with ineligible flags on the intake side, and then can't see them at all on the queue side. Or maybe that's what users would expect. FWIW.

amprokop:
Interesting — effectively this PR seems to hide the ineligible issues from all users searching for a case. If there is some reason we want to display those issues to Queue users, then front-end filtering makes sense. Wondering who would know about that— @laurjpeterson maybe?

app/models/appeal.rb
```diff
@@ -45,6 +45,10 @@ def reviewing_judge_name
     task ? task.assigned_to.try(:full_name) : ""
   end
 
+  def eligible_request_issues
+    request_issues.select(&:eligible?)
+  end
+
   def issues
     { decision_issues: decision_issues, request_issues: request_issues }
```
lowellrex: I wonder if we should be using `eligible_request_issues` here instead of including ineligible issues here as well. For now I think this is fine because it is not a regression (since the behaviour isn't going to change, though maybe is should).

amprokop: Good question- don’t know!

PR #7686, opened 2018-11-01
[WIP] Remove task action column

lowellrex:
Currently we use the action column of the task table as a psuedo-class that we use to determine various behaviours. This PR makes it so the code does not rely on that column any more.

Since the `Task.action` field is currently referenced across a broad swath of the codebase, I propose doing this overhaul in a few steps to avoid introducing a mass of changes that cause some bug that would be hard to troubleshoot or undo if we did the whole overhaul at once.

## Release plan
* [x] Reference the Task `action` field as `label` on the frontend (#7970)
* [ ] Split `JudgeTask` into two subclasses (JudgeAssignTask and JudgeReviewDecisionTask perhaps) and have those subclasses determine the label field instead of the database (#8033)
* [ ] Create named subclasses for the different `ColocatedTask`s currently defined by their action field
* [ ] Migrate existing `ColocatedTask`s to new subclasses (combination of `type` and `action` fields)
* [ ] Make sure LegacyTasks are handled properly
* [ ] Remove the `action` column from the database


app/models/tasks/colocated_task.rb
```diff
@@ -101,3 +83,132 @@ def assigned_by_role_is_valid
...
 end
+
+class MovableColocatedTask < ColocatedTask
+  def available_actions(user)
+    actions = super(user)
+
+    # Expect the first action to be the mark task complete action which we want to replace with the task-specific
+    # relocation logic.
+    if appeal.class.eql?(LegacyAppeal)
+      actions[0] = {
```
mdbenjam: This is janky haha if we ever change the ordering above we'll break this. I'd either not rely on super or find the the one we want to replace in a more robust way.

mdbenjam: Also isn't mark task complete at index `1`?

app/models/tasks/judge_task.rb
```diff
@@ -110,3 +98,30 @@ def self.list_of_assignees
   #:nocov:
   # rubocop:enable Metrics/AbcSize
 end
+
+# TODO: I actually don't think actions serve any purpose other than describing which task type we should use.
```
mdbenjam: Yeah it might be good to keep this around as the human readable version. We can pass that to the frontend for the table view of the task. Though action is probably not the right name.

app/models/tasks/judge_task.rb
```diff
@@ -31,16 +17,18 @@ def self.create_from_params(params, user)
     new_task
   end
 
+  # TODO: Should we explicitly create the AssignJudgeTask instead of the JudgeTask we are creating right now?
   def self.modify_params(params)
-    super(params.merge(action: "assign"))
+    super(params.merge(type: AssignJudgeTask.name))
```
mdbenjam: Clever, yeah I think we should figure out what we want the task tree to look like with this. In fact, I'm not even sure the distinction between these two tasks makes a ton of sense. The way we've built it, judges can't send a case back to an attorney for reworking 😬

lowellrex:
Going to split this into two PRs since I think it will be significantly easier to remove the action field for judge tasks than it will be for colocated tasks. Also, to confirm that judge and colocated tasks are the only ones with action values I ran the following query:
```
psql> select distinct type, action from public.tasks where action is not null order by 1, 2;
     type      |        action         
---------------+-----------------------
 ColocatedTask | aoj
 ColocatedTask | extension
 ColocatedTask | foia
 ColocatedTask | hearing_clarification
 ColocatedTask | ihp
 ColocatedTask | missing_records
 ColocatedTask | other
 ColocatedTask | poa_clarification
 ColocatedTask | schedule_hearing
 ColocatedTask | translation
 JudgeTask     | assign
 JudgeTask     | review
```

pkarman:
hey @lowellrex can we close this?

lowellrex:
Yeah. I was a little ambitious when I made this. Thought it'd be easier than it turned out to be. I'll transform this into a ticket to do the work.

Issue #7665, opened 2018-11-01
Make queue table view description more generic


lowellrex:
Currently the descriptions for the different tabs in the queue table view are very specific to the co-located admin team. Since this view will be the basis of all generic queues that we roll out to all of the different teams (and individuals on those teams) going forward we need to make these more generic. If it is possible to retain the current text for the colocated queue table view then we should do that. However, since they already have some experience using this view I don't think it would be a terrible outcome if we had to change the text for them as well.

| Tab | Co-located text | Generic text |
| --- | --- | --- |
| New | These are new administrative actions that have been assigned to you | These are cases that require some action from you |
| Pending action | When a hold is complete, the case will appear here. Cases with new documents will also appear here | These are cases that have been assigned to some other team (or placed on hold) require some action from you again |
| On hold | When a hold is complete, the case will automatically return to the Pending action tab | These are cases that do not require immediate action from you |
| Completed | Cases with tasks that were marked complete in the last two weeks will appeal here | _No change_ |

@sneha-pai @allyceh @laurjpeterson @joeyyang What do y'all think of this approach?

## Acceptance criteria
* [ ] More generic descriptions are displayed for all generic queues

joeyyang:
Hi everyone! Not 100% sure I'm getting the meaning correct, but I took a stab at reducing the UI copy down for scannability and comprehension:

| Tab | Co-located text | Generic text |
| --- | --- | --- |
| New | These are new administrative actions that have been assigned to you | New cases assigned to you |
| Pending action | When a hold is complete, the case will appear here. Cases with new documents will also appear here | Cases assigned back to you *or* Cases requiring action |
| On hold | When a hold is complete, the case will automatically return to the Pending action tab | Cases on hold (will return to Pending when hold is released) |
| Completed | Cases with tasks that were marked complete in the last two weeks will appear here | Cases completed (last two weeks) |

laurjpeterson:
> These are cases that have been assigned to some other team (or placed on hold) require some action from you again
- Can you explain this a bit further? Why would something assigned to some other team require some action from the individual again and be in the pending action tab, instead of the New tab?

joeyyang:
Hi everyone — threw together a PR here: https://github.com/department-of-veterans-affairs/caseflow/pull/7697. Let me know your thoughts on copy and I can change. Thanks!

Issue #7391, opened 2018-10-15
Move tabs logic inside TaskTable component


lowellrex:
Currently when we draw tabs, we create separate `TaskTable` components for each tab. This ticket exists to move the tabs logic inside of the `TaskTable` (first discussed as [a comment on a PR](https://github.com/department-of-veterans-affairs/caseflow/pull/7389#discussion_r225183195)).

Two ideas for how this could work are:
1. Calling code passes `TaskTable` an array of objects that have titles, optional descriptions, and functions that filter which tasks to display in the table (and maybe eventually columns as well if we want to display different columns in each view?).
2. Display a pre-defined set of tabs if a `showTabs` flag is set on the component by the calling code.

## Acceptance criteria
* `TaskTable` handles the rendering of tabs within the component itself.
* `TaskTable` continues to work and appear as it does now for use cases where it does not handle tab logic internally.

laurjpeterson:
@lowellrex - i see you merged a PR here. can this be closed?

lowellrex:
Closing this out since this work will be completed as part of #10049.

PR #7687, opened 2018-11-01
Add tabs to organizational view

### LINKED ISSUE #7526: Add tabs for organization queues
Currently, organizational queue views only display one table with all cases that have incomplete tasks assigned to that team, even if the task assigned to the organization has a child task that has been assigned to a member of that team.

We should add tabs to this the organizational queue views that correctly indicate the state of the organization's task. For example, after someone assigns a case from organization queue to individual queue, the case should move to an On Hold tab (perhaps called "assigned" or something similar), so users know that the case has been assigned to a person and will not assign the case a second time.

image

joeyyang:
Resolves #7526 

Now: 
nov-01-2018 15-49-10

Before: 
screenshot 2018-11-01 15 52 48

### Description
Adds tabs to the organizational view, separating cases into "Unassigned," "Assigned," and "Completed".

A few notes:
- Please see code comments in the "Files changed" view
- Please don't hold back if I've done anything that's not idiomatic in this codebase, bad style, etc. 
- Wasn't able to test this with more data besides the one case — is there a test case I can use that will have more data?
- This PR does not update the copy in the table, per https://github.com/department-of-veterans-affairs/caseflow/issues/7665. That will be addressed in a separate PR.

### Acceptance Criteria
- Not sure what goes here, can you please advise?

### Testing Plan
- Not sure what goes here, can you please advise?


client/app/queue/OrganizationQueue.jsx
```diff
@@ -23,24 +29,30 @@ class OrganizationQueue extends React.PureComponent {
...
+        page: <CompletedTasksTab />
+      }
+    ];
 
     return <AppSegment filledBackground>
       <div>
         <h1 {...fullWidth}>{sprintf(COPY.ORGANIZATION_QUEUE_TABLE_TITLE, this.props.organizationName)}</h1>
-        {content}
+        <TabWindow 
+          name="tasks-organization-queue"
```
joeyyang: Wasn't sure what to name this or if there was a convention — is this ok?

client/app/queue/OrganizationQueue.jsx
```diff
@@ -63,3 +75,87 @@ const mapDispatchToProps = (dispatch) => ({
...
+        includeReaderLink
+        tasks={props.tasks}
+      />;
+
+    return <React.Fragment>
+      {content}
+    </React.Fragment>;
+  });
+
+const countCasesByType = (tasks) => {
```
joeyyang: Was wondering if there was a more elegant way to count up the # of cases in each category — any ideas for something better?

client/app/queue/OrganizationQueue.jsx
```diff
@@ -5,42 +5,79 @@ import { bindActionCreators } from 'redux';
...
   tasksWithAppealSelector
 } from './selectors';
+
 import { clearCaseSelectSearch } from '../reader/CaseSelect/CaseSelectActions';
+import TASK_STATUSES from '../../constants/TASK_STATUSES.json';
 
 import { fullWidth } from './constants';
 import COPY from '../../COPY.json';
 
+const countCasesByType = (tasks) => {
```
joeyyang: Was wondering if there was a more elegant way to count up the # of cases in each category — any ideas for something better?

client/COPY.json
```diff
@@ -88,6 +88,11 @@
   "COLOCATED_QUEUE_PAGE_ON_HOLD_TASKS_DESCRIPTION": "When a hold is complete, the case will automatically return to the Pending action tab.",
   "COLOCATED_QUEUE_PAGE_COMPLETE_TASKS_DESCRIPTION": "Cases with tasks that were marked complete in the last two weeks will appeal here.",
 
+  "ORGANIZATIONAL_QUEUE_PAGE_UNASSIGNED_TAB_TITLE": "Unassigned (%d)",
+  "ORGANIZATIONAL_QUEUE_PAGE_ASSIGNED_TAB_TITLE": "Assigned (%d)",
+  "ORGANIZATIONAL_QUEUE_PAGE_COMPLETE_TAB_TITLE": "Completed",
+  "ORGANIZATIONAL_QUEUE_EMPTY_STATE_MESSAGE": "This queue is empty. You can ",
```
lowellrex: Can we add some copy here that explains what the three different tabs show? I think we can use the `COLOCATED_QUEUE_PAGE_COMPLETE_TASKS_DESCRIPTION` for the completed tab. For the "Unassigned" and "Assigned" tabs how does "Cases with tasks to be completed by the %s team that have not been assigned to any individual person" and "Cases assigned to a member of the %s team".

client/app/queue/OrganizationQueue.jsx
lowellrex: We can just get the length of the arrays returned by the selectors that you created in this PR (`getUnassignedOrganizationalTasks`, etc.).

For what it's worth, the mapping isn't 1:1 for tasks/cases, but right now we expect there to be only one task assigned to an organization for each case so this will work for now and be a very close approximation even when we start allowing the creation of multiple tasks for each case. Though we do need to keep this in mind as we move forward so we can create a more robust count when the time comes.

client/app/queue/selectors.js
```diff
@@ -99,6 +99,23 @@ export const getTasksForAppeal = createSelector(
   }
 );
 
+export const getUnassignedOrganizationalTasks = createSelector(
+  [tasksWithAppealSelector],
```
lowellrex: `tasksWithAppealSelector()` returns all of the tasks in the global store regardless of who they are assigned to. Currently this only returns the tasks assigned to the organization by virtue of the fact that accessing the organizational queue table view forces a global state reload because we must directly navigate to their queue by typing in the organization's URL (or by doing some other browser-based navigation which will reload the webpage). However, should we ever change the way we navigate to the organizational queue we could have a global store with tasks for other cases not assigned to the organization.

This problem wasn't introduced in this PR, but I think we can handle this situation cleanly now by adding a selector that only picks out the organization's tasks here.

([Somewhat related ticket](https://github.com/department-of-veterans-affairs/caseflow/issues/7709) that aims to simplify how we manage task state between frontend and backend).

client/app/queue/OrganizationQueue.jsx
```diff
@@ -63,3 +100,75 @@ const mapDispatchToProps = (dispatch) => ({
...
+        includeType
+        includeDocketNumber
+        includeDaysWaiting
+        includeReaderLink
+        tasks={props.tasks}
+      />;
+
+    return <React.Fragment>
+      {content}
+    </React.Fragment>;
```
lowellrex: Shouldn't be addressed in this PR, but I think the fact that we've got another place with task tables and tabs build the case for the [PR that consolidates that into a single component](https://github.com/department-of-veterans-affairs/caseflow/issues/7391).

client/app/queue/OrganizationQueue.jsx
```diff
@@ -63,3 +100,75 @@ const mapDispatchToProps = (dispatch) => ({
...
+  });
+
+const CompletedTasksTab = connect(
+  (state: State) => ({ tasks: getCompletedOrganizationalTasks(state) }))(
+  (props: { tasks: Array<TaskWithAppeal> }) => {
+    const noTasks = !_.size(props.tasks);
```
lowellrex: I don't think we need to display text if there are not tasks for a certain tab. Eventually I think we'd like to show some text if there are no tasks at all for of the tabs, but until then (maybe #7391 makes this easy) I think it is fine to always show the table even if there are no rows in the table.

client/app/queue/selectors.js
```diff
@@ -99,6 +99,23 @@ export const getTasksForAppeal = createSelector(
...
+  (tasks: Tasks) => _.filter(tasks, (task) => {
+    return (task.status === TASK_STATUSES.assigned || task.status === TASK_STATUSES.in_progress);
+  })
+);
+
+export const getAssignedOrganizationalTasks = createSelector(
```
lowellrex: Can we change the name of this function to something like `getOnHoldOrganizationalTasks`? Currently all tasks with the status of "on hold" should have a child task that is assigned to a member of that organization (or assigned to a different organization), but this might not always be the case.

lowellrex:
This looks really good! Left a few small comments that should be addressed but this is well on its way!

client/app/queue/selectors.js
joeyyang: created new selector in our pairing session — see updated code

client/COPY.json
joeyyang: Copy updated — lmk what you think!

joeyyang:
This PR was changed to only get tasks matching the organization ID. Summary of changes (mostly just for my understanding of the codebase):
1. changed `tasks_controller.rb` to send back the organization ID from backend
2. change Redux stores to accept a notion of "active organization ID" to global state:
    1. change `uiReducer.js` to add `activeOrganizationId` field to global state
    2. add `SET_ACTIVE_ORGANIZATION_ID` to `uiConstants.js`
    3. add `setActiveOrganizationId` to `uiActions.js`
3. get that information in `OrganizationQueueLoadingScreen.jsx` — get it back from the response and then use the new `setActiveOrganizationId` action to set it as a prop in the global store
    1. also then bind that action creator to the global store
4. create a new selector which matches the current organization ID (passed in from above steps) to the global list of tasks, and display only those tasks

Updated screenshots — @lowellrex can you please review the copy?
screenshot 2018-11-05 12 35 31
screenshot 2018-11-05 12 35 40
screenshot 2018-11-05 12 35 44



New

client/COPY.json
```diff
@@ -86,7 +86,14 @@
   "COLOCATED_QUEUE_PAGE_NEW_TASKS_DESCRIPTION": "These are new administrative actions that have been assigned to you.",
   "COLOCATED_QUEUE_PAGE_PENDING_TASKS_DESCRIPTION": "When a hold is complete, the case will appear here. Cases with new documents will also appear here.",
   "COLOCATED_QUEUE_PAGE_ON_HOLD_TASKS_DESCRIPTION": "When a hold is complete, the case will automatically return to the Pending action tab.",
-  "COLOCATED_QUEUE_PAGE_COMPLETE_TASKS_DESCRIPTION": "Cases with tasks that were marked complete in the last two weeks will appeal here.",
+  "COLOCATED_QUEUE_PAGE_COMPLETE_TASKS_DESCRIPTION": "Cases with tasks that were marked complete in the last two weeks will appear here.",
+  "COLOCATED_QUEUE_PAGE_UNASSIGNED_TASKS_DESCRIPTION": "Cases owned by the %s team that are unassigned to a person.",
+  "COLOCATED_QUEUE_PAGE_ASSIGNED_TASKS_DESCRIPTION": "Cases assigned to a member of the %s team.",
```
lowellrex: I think it might make more sense to label these two pieces of copy with the prefix `ORGANIZATIONAL_QUEUE...` since they only appear on the org queue page for now.

PR #7692, opened 2018-11-02
Legacy Automatic Case Distribution

cmgiven:
`Distribution.create(judge: judge)`


app/models/distributed_case.rb
```diff
@@ -0,0 +1,12 @@
+class DistributedCase < ApplicationRecord
+  belongs_to :distribution
+
+  validates :distribution, :case_id, :docket, :priority, :genpop, :genpop_query, :ready_at, presence: true
```
amprokop: Purely for standardization, `case_id` could be changed to `appeal_id`, since we refer to it as an `appeal` almost everywhere else.

app/models/distributed_case.rb
```diff
@@ -0,0 +1,12 @@
+class DistributedCase < ApplicationRecord
```
amprokop: What value do you see in persisting `:priority, :genpop, :genpop_query`? Is this for analytics, so we can easily show how the proportions change over time and per judge?

app/models/distribution.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  private
+
+  def distribute
+    legacy_distribution
+    update(statistics: legacy_statistics, completed_at: Time.zone.now)
+  end
+
+  def user_is_judge
+    judge.judge_in_vacols?
```
amprokop: This method will return true for acting judges, since we treat acting judges as judges in some cases, not entirely correctly. I don't believe we'd want to allow them to use this (though requirements may change)

I think the easiest way to filter out acting judges would be to add `&& !judge.attorney_in_vacols`, but we can also clarify.

app/models/distribution.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def judge_has_no_unassigned_cases
+    return false if JudgeTask.where(assigned_to: judge, action: "assign").any?
```
amprokop: I think we'd need to tweak this: `JudgeTask.where(assigned_to: judge, action: "assign", status: ["assigned", "in_progress"])`  — @lowellrex or @aroltsch, could you confirm offhand that that's the correct way to find if a judge has any cases in the `Cases to Assign` queue?

lowellrex: Would prefer to use constant-ized versions of those statuses, but that is correct!

app/models/vacols/case_docket.rb
```diff
@@ -65,5 +99,191 @@ def self.docket_counts_by_month
...
+                                                     query,
+                                                     judge.vacols_attorney_id,
+                                                     (genpop.nil? || !genpop) ? 1 : 0,
+                                                     (genpop.nil? || genpop) ? 1 : 0,
+                                                     range,
+                                                     range.nil? ? 1 : 0,
+                                                     limit
+                                                   ])).to_hash
+      vacols_ids = appeals.map { |appeal| appeal["bfkey"] }
+      batch_update_vacols_location(conn, judge.vacols_uniq_id, vacols_ids)
```
amprokop: Suggest that we add a dry run parameter to the top level `Distribution.create` that, if true, will skip the vacols location updating/any writes, but print out the appeals that will be distributed.

app/models/vacols/case_docket.rb
```diff
@@ -2,6 +2,40 @@ class VACOLS::CaseDocket < VACOLS::Record
   # :nocov:
   self.table_name = "vacols.brieff"
 
+  # rubocop:disable Metrics/MethodLength
+  def self.counts_by_priority_and_readiness
+    query = <<-SQL
+      select count(*) N, PRIORITY, READY
+      from (
+        select case when BFAC = '7' or nvl(AOD_DIARIES.CNT, 0) + nvl(AOD_HEARINGS.CNT, 0) > 0 then 1 else 0 end as PRIORITY,
+          case when BFCURLOC in ('81', '83') then 1 else 0 end as READY
```
amprokop: Location 81 is case storage, right? What is location 83 again?

amprokop: image
Nevermind, looked it up. You're confident that we don't need to count any of these other locations?

app/models/distributed_case.rb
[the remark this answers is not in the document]
cmgiven: Except that would lead me to believe that it is `appeals.id`, when it's either `vacols.brieff.bfkey` or `appeals.uuid` (which can then be retrieved using `Appeal.find_appeal_by_id_or_find_or_create_legacy_appeal_by_vacols_id`). Is there a better term for that key?

app/models/legacy_docket.rb
```diff
@@ -0,0 +1,68 @@
...
+    cases = LegacyAppeal.repository.distribute_priority_appeals(distribution.judge, genpop, limit).map do |record|
+      DistributedCase.create(distribution: distribution,
+                             case_id: record["bfkey"],
+                             docket: "legacy",
+                             priority: true,
+                             ready_at: VacolsHelper.normalize_vacols_datetime(record["bfdloout"]),
+                             genpop: !record["vlj"].nil?,
+                             genpop_query: maybe_boolean_to_string(genpop))
+    end
+    DistributedCase.create(cases)
```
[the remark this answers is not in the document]
cmgiven: Indeed. Can you tell the parts of the PR that I haven't tested?

app/models/concerns/legacy_case_distribution.rb
```diff
@@ -0,0 +1,58 @@
...
+  def priority_count
+    docket.count(priority: true, ready: true)
+  end
+
+  def net_docket_range
+    [total_batch_size - priority_count, 0].max
+  end
+
+  def priority_target
+    proportion = [1.0 * priority_count / total_batch_size, 1.0].min
```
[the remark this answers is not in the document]
cmgiven: :+1:

app/models/concerns/legacy_case_distribution.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  def priority_count
+    docket.count(priority: true, ready: true)
+  end
+
+  def net_docket_range
+    [total_batch_size - priority_count, 0].max
+  end
+
+  def priority_target
```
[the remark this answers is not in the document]
cmgiven: :+1:

app/models/concerns/legacy_case_distribution.rb
```diff
@@ -0,0 +1,58 @@
...
+                                                                  genpop: false,
+                                                                  limit: rem)
+    rem -= priority_hearing_appeals.count
+
+    nonpriority_hearing_appeals = docket.distribute_nonpriority_appeals(self,
+                                                                        genpop: false,
+                                                                        range: net_docket_range,
+                                                                        limit: rem)
+    rem -= nonpriority_hearing_appeals.count
+    priority_rem = (priority_target - priority_hearing_appeals.count).clamp(0, rem)
```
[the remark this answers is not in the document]
cmgiven: Maybe? Note that I still need to do the min operation to prevent exceeding `rem`. I could go either way, but when I learned that splatting a `nil` for the return value gives an empty array, I thought that was nifty, so I'll do it your way.

app/models/distributed_case.rb
cmgiven: Yup.

`:priority` is whether the case was considered priority at the time it was distributed.
`:genpop` is whether the case was considered genpop at the time it was distribtued.
`:genpop_query` is the yes/no/any `genpop` argument to `distribute_*priority_appeals`, which provides sufficient information to understand at what step of the algorithm the case was distributed.

app/models/distribution.rb
cmgiven: Good flag. Eventually we'll need this to work for acting judges as well, but they should only be distributed the cases that are tied to them. We don't yet support that.

app/models/distribution.rb
cmgiven: :+1:

app/models/vacols/case_docket.rb
cmgiven: I wasn't going to do this, because I had no use for it, as there's no way to guarantee that the appeals returned by the dry run would be returned by the write-enabled run, but I just realized that I do have a use for this for the AMA algorithm, at least for `distribute_priority_appeals`. I'll add.

app/models/vacols/case_docket.rb
cmgiven: Yes. Until next week sometime, 81 is the only location that matters. Next week, Jed will be enabling functionality for case review by paralegals, who will place their completed cases in 83.

app/models/vacols/case_docket.rb
```diff
@@ -65,5 +99,191 @@ def self.docket_counts_by_month
...
+              from HEARSCHED
+              where HEARING_TYPE in ('C', 'T', 'V')
+                and AOD in ('G', 'Y')
+              group by FOLDER_NR
+            ) AOD_HEARINGS on AOD_HEARINGS.FOLDER_NR = BRIEFF.BFKEY
+          ) BRIEFF
+          inner join FOLDER on FOLDER.TICKNUM = BRIEFF.BFKEY
+          where BRIEFF.BFMPRO <> 'HIS' and BRIEFF.BFCURLOC in ('81', '83')
+            and BFAC <> '7' and AOD = '0'
+          order by case when substr(TINUM, 1, 2) between '00' and '29' then 1 else 0 end, TINUM
```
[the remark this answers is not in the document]
cmgiven: Nice. I'll add a "happy 100th birthday to the Board" message.

app/models/distribution.rb
cmgiven: Added support for acting judges.

app/models/concerns/legacy_case_distribution.rb
```diff
@@ -0,0 +1,69 @@
...
+                                                                       limit: priority_rem)
+      rem -= priority_nonhearing_appeals.count
+    end
+
+    nonpriority_appeals = docket.distribute_nonpriority_appeals(self, limit: rem)
+
+    [*priority_hearing_appeals, *nonpriority_hearing_appeals, *priority_nonhearing_appeals, *nonpriority_appeals]
+  end
+
+  def legacy_acting_judge_distribution
```
amprokop: FYI: acting judges currently only see the attorney view in Caseflow.

We know there's a need to have acting judges toggle between attorney view and judge view, but have delayed it, since we've heard that the Board is going to re-evaluate the usefulness of acting judges at the end of the year.

However, this code is still useful and neat to have since it'll be backwards compatible — there's no reason why we couldn't do distribution in Caseflow and then have acting judges assign the cases in VACOLS.

cc @laurjpeterson

app/models/vacols/case_docket.rb
```diff
@@ -65,5 +101,217 @@ def self.docket_counts_by_month
...
+        ) BRIEFF
+        left join (
+          select distinct TITRNUM, TINUM,
+            first_value(BOARD_MEMBER) over (partition by TITRNUM, TINUM order by HEARING_DATE desc) VLJ
+          from HEARSCHED
+          inner join FOLDER on FOLDER.TICKNUM = HEARSCHED.FOLDER_NR
+          where HEARING_TYPE in ('C', 'T', 'V') and HEARING_DISP = 'H'
```
amprokop: When would hearing_type *not* be one of C, T, or V?

app/models/vacols/case_docket.rb
```diff
@@ -65,5 +101,217 @@ def self.docket_counts_by_month
...
+      )
+      where ((VLJ = ? and 1 = ?) or (VLJ is null and 1 = ?))
+        and (DOCKET_INDEX <= ? or 1 = ?)
+        and rownum <= ?
+    SQL
+
+    fmtd_query = sanitize_sql_array([
+                                      query,
+                                      judge.vacols_attorney_id,
+                                      (genpop.nil? || !genpop) ? 1 : 0,
```
amprokop: nit: I might add a default value to the argument at the top of the method definition `genpop = false` and make this line just `!genpop ? 1 : 0`

app/models/vacols/case_docket.rb
```diff
@@ -65,5 +101,217 @@ def self.docket_counts_by_month
...
+            first_value(BOARD_MEMBER) over (partition by TITRNUM, TINUM order by HEARING_DATE desc) VLJ
+          from HEARSCHED
+          inner join FOLDER on FOLDER.TICKNUM = HEARSCHED.FOLDER_NR
+          where HEARING_TYPE in ('C', 'T', 'V') and HEARING_DISP = 'H'
+        ) VLJ_HEARINGS
+          on VLJ_HEARINGS.VLJ not in ('000', '888', '999')
+            and VLJ_HEARINGS.TITRNUM = BRIEFF.TITRNUM
+            and (VLJ_HEARINGS.TINUM is null or VLJ_HEARINGS.TINUM = BRIEFF.TINUM)
+      )
+      where ((VLJ = ? and 1 = ?) or (VLJ is null and 1 = ?))
```
amprokop: nit:  why do we need to need to handle the case when VLJ is null here? Seems like we should always pass a judge here. not a big deal, but it just makes the query a bit more complex i think?

app/models/vacols/case_docket.rb
cmgiven: F is an RO formal hearing. It’s basically a completely different thing that happens to be recorded in the same table.

app/models/vacols/case_docket.rb
```diff
@@ -65,5 +101,217 @@ def self.docket_counts_by_month
...
+          order by case when substr(TINUM, 1, 2) between '00' and '29' then 1 else 0 end, TINUM
+        ) BRIEFF
+        left join (
+          select distinct TITRNUM, TINUM,
+            first_value(BOARD_MEMBER) over (partition by TITRNUM, TINUM order by HEARING_DATE desc) VLJ
+          from HEARSCHED
+          inner join FOLDER on FOLDER.TICKNUM = HEARSCHED.FOLDER_NR
+          where HEARING_TYPE in ('C', 'T', 'V') and HEARING_DISP = 'H'
+        ) VLJ_HEARINGS
+          on VLJ_HEARINGS.VLJ not in ('000', '888', '999')
```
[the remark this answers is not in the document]
cmgiven: :+1:

app/models/vacols/case_docket.rb
```diff
@@ -65,5 +101,217 @@ def self.docket_counts_by_month
...
+            and VLJ_HEARINGS.TITRNUM = BRIEFF.TITRNUM
+            and (VLJ_HEARINGS.TINUM is null or VLJ_HEARINGS.TINUM = BRIEFF.TINUM)
+      )
+      where ((VLJ = ? and 1 = ?) or (VLJ is null and 1 = ?))
+        and (DOCKET_INDEX <= ? or 1 = ?)
+        and rownum <= ?
```
amprokop: So this is just basically oracle's version of `limit`? 😬

app/models/vacols/case_docket.rb
cmgiven: Ah, but nil actually has meaning here. Nil will accept both genpop and non-genpop cases. True or false will accept one and block the other.

app/models/vacols/case_docket.rb
cmgiven: The query is set up such that VLJ will only have a value when the case is tied to the VLJ. VLJ is null indicates that the case is genpop. Take a look at the specs for case_docket to see what it’s trying to do.

app/models/vacols/case_docket.rb
cmgiven: Ayup.

app/models/vacols/case_docket.rb
amprokop: Oh i see. I didn't understand this fully.

this clause is more like:

where vlj from hearing record is the judge we're passing and we're looking for genpop appeals OR there is no vlj on the hearing record and we're looking for genpop appeals
and the docket index is less than or equal to the range we passed OR we didn't pass a range
and the rownum is less than the limit we passed

cmgiven: Exactly. One small tweak:

where vlj from hearing record is the judge we're passing and we're looking for *non-*genpop appeals OR there is no vlj on the hearing record and we're looking for genpop appeals
and the docket index is less than or equal to the range we passed OR we didn't pass a range
and the rownum is less than the limit we passed

amprokop: Hm, actually I'm a bit confused, maybe you can help me clear this up.

What do the two genpop arguments to the sanitized SQL array actually do?

app/models/vacols/case_docket.rb
```diff
@@ -65,5 +101,217 @@ def self.docket_counts_by_month
...
+                                      range,
+                                      range.nil? ? 1 : 0,
+                                      limit
+                                    ])
+
+    conn.transaction do
+      if dry_run
+        conn.exec_query(fmtd_query).to_hash
+      else
+        conn.execute("lock table BRIEFF in row exclusive mode")
```
amprokop: Are we sure that this locks selected rows if the select query does not mention `FOR UPDATE`? Brief googling suggests that we might need a `FOR UPDATE` to have the row locks get applied, but I haven't tested.

app/models/vacols/case_docket.rb
cmgiven: The first parens tests whether the cases is non-genpop and is assigned to our judge (we never care about cases that are non-genpop and assigned to someone else). So the first genpop argument is asking, “Will you accept cases that are tied to the judge (that are not genpop)?”

The second parens is testing whether the case is genpop (is not tied to any judge). So the second genpop argument is asking, “Will you accept genpop cases?”

I can answer both those questions in the affirmative by passing genpop: nil.

cmgiven: If I pass genpop: false, I would expect to receive cases that are tied to our judge (the first parens). These are non-genpop cases.

local/vacols/VACOLS::Representative_dump.csv
```diff
@@ -24,13 +24,11 @@ repkey,repaddtime,reptype,repso,replast,repfirst,repmi,repsuf,repaddr1,repaddr2,
 2874889,2017-09-12,A,,Gottlieb,Wilburn,R,Sr.,39982 Dibbert Hills,Apt. 858,Leuschkeville,KY,65314,568.102.2351,Vitae magni temporibus corrupti quae sint.,KBERGNAUM,2017-09-12,,,,,,CK759075,
 2892519,2016-01-05,G,,Wunsch,Mara,S,DDS,493 Powlowski Island,Apt. 582,Lake Aiyana,HI,60887,(227) 408-4657 x946,Ex temporibus modi illum inventore praesentium.,RO09,2016-01-05,,,,,,CK1111639,
 2895173,2016-10-19,A,,Glover,Freida,Z,III,60597 Wunsch Square,Apt. 329,Hirthemouth,NC,99087,(756) 443-9391 x62989,,RO17,2016-10-19,,,,,,CK1037809,
-2895173,2017-10-02,D,,Glover,Freida,Z,III,60597 Wunsch Square,Apt. 329,Hirthemouth,NC,99087,(756) 443-9391 x62989,,WBARROWS,2017-10-02,,,,,,CK1037809,
```
amprokop: Hmm I *think* we discourage the removal of these rows since that will get wiped out the next time we sync with the FACOLS DB. @mdbenjam can you confirm?

amprokop: (On the other hand, everyone who seeds VACOLS will have to remove these rows manually, so I'm not sure it's bad to remove them as a stopgap)

app/models/vacols/case_docket.rb
cmgiven: This should lock the entire table. I can’t use for update because the select query uses groups and nested queries.

amprokop:
Not *quite* ready to approve this yet but getting close. Lemme know if there's any short-term urgency, from our conversations I'm guessing there's not.

local/vacols/VACOLS::Representative_dump.csv
cmgiven: Yeah, I accidentally committed these fixes at some point. But FACOLS doesn’t work if these duplicates exist, so I’m not sure how they got committed in the first place.

cmgiven:
No urgency. And yes, I’m not wild about placing a lock on the entirety of BRIEFF for ~2 seconds. Reads will still work, but that’ll delay any writes until it finishes moving cases. So that might be worth looking into alternatives.

app/models/vacols/case_docket.rb
cmgiven: A small improvement would be to do select for update on the innermost query. This would have the effect of locking everything in case storage, instead of what we really want to do which is to just lock the rows we’re interested in, but at least it wouldn’t lock the entirety of BRIEFF.

amprokop: Agreed. Locking Case Storage rows is actually pretty reasonable.

local/vacols/VACOLS::Representative_dump.csv
amprokop: I think the reason is that the rep table has a uniqueness constraint on the added at date & bfkey. But in a post-create hook, the DB always sets the added at date to the current system date, overwriting whatever value you use. 

So these values could be legitimate in the DB, but the hook breaks seeding.

app/models/vacols/case_docket.rb
amprokop: 👍

app/models/distributed_case.rb
amprokop: 👍

app/models/legacy_docket.rb
```diff
@@ -0,0 +1,65 @@
...
+  def count(priority: nil, ready: nil)
+    counts_by_priority_and_readiness.inject(0) do |sum, row|
+      next sum unless (priority.nil? || (priority ? 1 : 0) == row["priority"]) &&
+                      (ready.nil? || (ready ? 1 : 0) == row["ready"])
+      sum + row["n"]
+    end
+  end
+  # rubocop:enable Metrics/CyclomaticComplexity
+
+  def weight
```
amprokop: I don't see this method used anywhere yet — will this be used to balance legacy and AMA appeals?

app/models/vacols/case_docket.rb
```diff
@@ -2,6 +2,42 @@ class VACOLS::CaseDocket < VACOLS::Record
   # :nocov:
   self.table_name = "vacols.brieff"
 
+  class DocketNumberCentennialLoop < StandardError; end
```
amprokop: Love the name.

local/vacols/VACOLS::Representative_dump.csv
mdbenjam: @cmgiven if you rebuild your FACOLS container these should work. I think this should do it.

```
docker-compose down -v
docker-compose build --no-cache
rake local:build
```

app/models/vacols/case_docket.rb
amprokop: Ah I see!

My whole confusion rabbit hole here and above was from not quickly grokking that nil was a meaningful value.

How would you feel about making the parameter called `genpop_preference` and accepting `only_genpop`, `not_genpop`, or `any` as values? 

then changing this to
```
(genpop_preference == "any" || genpop_preference == "only_genpop" ) ? 1 : 0,
(genpop_preference == "any" || genpop_preference == "not_genpop" ) ? 1 : 0,
```

that way we surface the difference between these values easily.

app/models/vacols/case_docket.rb
amprokop: Ahh I see. Sorry it took me awhile. Suggest that to surface the three possible preferences we make them explicit.

app/models/legacy_docket.rb
amprokop: I don't see this used anywhere — did you write this method in advance to prepare for legacy + AMA case distribution?

app/models/vacols/case_docket.rb
```diff
@@ -65,5 +101,217 @@ def self.docket_counts_by_month
...
+  def self.distribute_priority_appeals(judge, genpop, limit, dry_run = false)
+    conn = connection
+
+    query = <<-SQL
+      select BFKEY, BFDLOOUT, VLJ
+      from (
+        select BFKEY, BFDLOOUT,
+          case when BFHINES is null or BFHINES <> 'GP' then nvl(VLJ_HEARINGS.VLJ, VLJ_PRIORDEC.VLJ) end VLJ
+        from (
+          select BFKEY, BFDLOOUT, BFMPRO, BFCURLOC, BFHINES, AOD
```
amprokop: Been thinking about how to make the two long queries here more fathomable.

My guess is that they'd become significantly more understandable if there was some way to extract out this set of subqueries and the one starting on line 126 into a separate SQL statement that is concatenated here (they're the same except for the top-level `SELECT`ed values, it looks). 

My guess is it would make a huge difference for comprehension just with this one change. What to you think?

app/models/legacy_docket.rb
cmgiven: That’s right.

cmgiven: Yup.

app/models/vacols/case_docket.rb
cmgiven: They’re also ordered differently, and one is only priority and the other excludes priority. But I could try and align the innermost query.

app/models/vacols/case_docket.rb
cmgiven: Yes, indeed. :+1:

cmgiven:
@amprokop Back to you.

app/models/legacy_docket.rb
amprokop: Whoops sorry for the double comment :)

app/jobs/start_distribution_job.rb
```diff
@@ -0,0 +1,16 @@
+class StartDistributionJob < ApplicationJob
```
amprokop: Interesting. What's the motivation behind making this into a job? A little bit rusty on our SQS setup — do we get retries for free then?

app/models/vacols/case.rb
```diff
@@ -210,51 +210,58 @@ def remand_clone_attributes
...
-  # rubocop:disable Metrics/MethodLength
   def update_vacols_location!(location)
+    self.class.batch_update_vacols_location(location, [bfkey])
+  end
+
+  # rubocop:disable Metrics/MethodLength
+  def self.batch_update_vacols_location(location, vacols_ids)
     unless location
-      Rails.logger.error "THERE IS A BUG IN YOUR CODE! It attempted to assign a case to a falsy location. " \
+      Rails.logger.error "THERE IS A BUG IN YOUR CODE! It attempted to assign a case to a falsey location. " \
```
amprokop: 👍

local/vacols/VACOLS::Representative_dump.csv
amprokop: @cmgiven — sorry if this thread lost you. I think we can wipe the changes to this file from this PR due to what Mark said.

If we do merge this in, I believe we'll have a little bit of a weird situation where the next person who pulls from our FACOLS cloud DB will have changes to this file that aren't in master. Not too bad, but worth avoiding.

app/jobs/start_distribution_job.rb
cmgiven: We could, although I have retries disabled to start with until we see what we're working with. The basic problem is that with typical VACOLS latency, I expect this job to take about 8500ms (my rule of thumb from appeals status is 500ms per query, and this job runs a max of 17 queries). That's maybe not too long, but it is on the upper limit of what I'm comfortable with. As I thought through the frontend, I liked the idea of breaking out three steps: create and validate the distribution, wait for the job to complete, and then retrieve the newly distributed cases.

local/vacols/VACOLS::Representative_dump.csv
cmgiven: :+1:

Issue #7652, opened 2018-10-31
Intake Refactor | Store decision issues instead of rating issues in the DB


shanear:
Right now, in order to connect the rating issues that resulted from a request issue on a HLR or SC, we store that connection issue in the `rating_issues` table. After many fruitful discussions about the data model, we decided we should be storing this relationship in the same table that will eventually be used by appeals, which is `decision_issues`.

## Acceptance criteria
- When an EP for a supplemental claim or higher level review is cleared, decision issues are created for each rating issue made on the resulting rating. They should be connected in the DB to the request issues matching their associated contentions.
- If a rating issue is added and not connected to one of the request issue contentions, then it should still be created as a decision issue and connected directly to the decision review. (this shouldn't happen, but is technically possible in VBMS)
- The `rating_issues` table in DB should be deleted, and `RatingIssue` should return to an ephemeral class. 

## Technical Notes
This is a mural describing the relationships between the objects and models: https://app.mural.co/t/workqueue2001/m/workqueue2001/1540421981879/ba39075927b4039387697e094436e6a236fe7fa4

There is unknown time delay on when the rating is retrievable from when the EP status is changed to clear. We need to factor this time delay in when retrieving the rating and the decision issues.

There should not be any records in the `rating_issues` table in prod, as this functionality is not yet live, so no need to worry about migrating data.

leikkisa:
Need to have a conversation with Queue team since they use the Decision Issues model and now we want to do it. Mark is going to kick off a conversation.

Suggestions for collaboration:
- Start reviewing each others' PRs (Queue <> Intake).
- Have an ambassador on each team to attend each others' meetings (sprint planning)?

shanear:
```
[...]

PR #7694, opened 2018-11-02
Move rating issues to decision issues

pkarman:
connects #7652 

### Description
Moves most `RatingIssue` columns to `DecisionIssue` and returns `RatingIssue` to an ephemeral model with no table.

### Acceptance Criteria
- [x] Code compiles correctly
- [x] Tests continue to pass


app/models/rating_issue.rb
```diff
@@ -1,8 +1,11 @@
-class RatingIssue < ApplicationRecord
-  belongs_to :source_request_issue, class_name: "RequestIssue"
+# ephemeral class used for caching Rating Issues for client,
+# and for creating DecisionIssues when a Rating Issue has a contention_reference_id
```
shanear: nice! love these class descriptions

app/models/rating_issue.rb
```diff
@@ -23,21 +26,28 @@ def from_ui_hash(ui_hash)
...
         promulgation_date: ui_hash[:promulgation_date],
+        profile_date: ui_hash[:profile_date],
         contention_reference_id: ui_hash[:contention_reference_id]
       )
     end
   end
 
-  def save_with_source_request_issue!
-    fetch_source_request_issue
+  def save_decision_issue
```
shanear: this feels bang `!` worthy, but i guess there isn't really a good definition of when to use it lol.

app/models/rating_issue.rb
```diff
@@ -23,21 +26,28 @@ def from_ui_hash(ui_hash)
...
     return unless source_request_issue
 
-    # if a RatingIssue already exists, update rather than attempt to insert a duplicate
-    if existing_rating_issue
-      existing_rating_issue.update!(source_request_issue: source_request_issue)
-      self.attributes = existing_rating_issue.attributes
+    # if a DecisionIssue already exists, update rather than attempt to insert a duplicate
+    if decision_issue
```
shanear: this is awesome

app/models/rating_issue.rb
pkarman: My understanding of the `!` convention is to indicate when a method modifies the object on which it is called.

Since this method no longer modifies the `RatingIssue` I removed it.

app/models/decision_issue.rb
```diff
@@ -1,4 +1,4 @@
 class DecisionIssue < ApplicationRecord
-  validates :disposition, inclusion: { in: Constants::ISSUE_DISPOSITIONS_BY_ID.keys.map(&:to_s) }
-  belongs_to :request_issue
+  validates :disposition, inclusion: { in: Constants::ISSUE_DISPOSITIONS_BY_ID.keys.map(&:to_s) }, allow_nil: true
```
pkarman: the db does not have a `not null` constraint, so re-enforcing that here. 

The `inclusion` validation will ignore nil values.

app/models/rating_issue.rb
shanear: ahh i see, i interpret it to mean that it's a method used to alter the DB or a dependency. i.e. it has a side effect. I don't care too much though

PR #7700, opened 2018-11-02
Fixing job error routing

mdbenjam:
### Description
Active job errors were not being routed properly. We lost our context when we caught errors in middleware. We're moving the rescue to within the job.

### Testing Plan
1. Print out the context before reporting an error locally.


config/initializers/shoryuken.rb
```diff
@@ -12,14 +11,18 @@
   Shoryuken::Client.sqs.config[:endpoint] = URI(Rails.application.config.sqs_endpoint)
 end
 
+if Rails.application.config.sqs_create_queues
```
mdbenjam: I'd removed this in the past, don't know why... putting it back.

PR #7702, opened 2018-11-02
Schedule Veteran -  Central Office date offset bug

### LINKED ISSUE #7693: Schedule Veteran - Date offset bug prevents CO hearing update
When confirming the scheduling of a veteran from the Case Details modal the date is being converted to UTC in the back-end and we cannot find the row to update. Modal correctly reports an error.

oscarAtNava:
Resolves #7693 

### Description
Fixed a date offset bug in Central Office update logic.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Select a veteran in Central Office.
2. In Case Details schedule the veteran
3. After receiving successful confirmation verify DB row is updated with veteran vacols id.


app/models/tasks/schedule_hearing_task.rb
```diff
@@ -45,26 +50,36 @@ def update_from_params(params, current_user)
...
-      loaded_hearing = VACOLS::CaseHearing.load_hearing(hearing.hearing_pkseq)
-      HearingRepository.update_vacols_hearing!(loaded_hearing, folder_nr: appeal.vacols_id)
+      update_co_hearing(hearing_date_str)
     else
-      hearing = VACOLS::CaseHearing.find(hearing_pkseq)
-      hearing_hash = to_hash(hearing)
-      hearing_hash[:folder_nr] = appeal.vacols_id
-      hearing_hash[:hearing_date] = hearing_date
-      HearingRepository.create_vacols_child_hearing(hearing_hash)
+      create_child_video_hearing(hearing_pkseq, hearing_date)
```
sharonwarner: Let's map hearing_pkseq to id so that it's easier to refactor when we switch to Caseflow hearings. Up to you if you make the update here or in a separate PR.

oscarAtNava: Duly noted. I will make it in a subsequent PR where we also need to use the model and not the repository to create or update the child.

Issue #7703, opened 2018-11-02
Failing RAMP test


mdbenjam:
```
Edit issues Higher-Level Reviews when there is a rating end product shows request issues and allows adding/removing issues - spec.feature.intake.edit_spec
spec/feature/intake/edit_spec.rb
Failure/Error:
  expect(Fakes::VBMSService).to have_received(:create_contentions!).once.with(
    veteran_file_number: veteran.file_number,
    claim_id: rating_epe.reference_id,
    contention_descriptions: [
      RequestIssue::UNIDENTIFIED_ISSUE_MSG,
      "PTSD denied",
      "Left knee granted"
    ],
    special_issues: [],
    user: current_user

  #<Fakes::VBMSService (class)> received :create_contentions! with unexpected arguments
    expected: ({:veteran_file_number=>"12341234", :claim_id=>"FAKECLAIMID", :contention_descriptions=>["UNIDENTIFIED...e: "Lauren Roth", email: "test@example.com", roles: ["Mail Intake"], selected_regional_office: nil>})
         got: ({:veteran_file_number=>"12341234", :claim_id=>"FAKECLAIMID", :contention_descriptions=>["Active Duty ...e: "Lauren Roth", email: "test@example.com", roles: ["Mail Intake"], selected_regional_office: nil>}) (1 time)
              ({:veteran_file_number=>"12341234", :claim_id=>"FAKECLAIMID", :contention_descriptions=>["Left knee gr...e: "Lauren Roth", email: "test@example.com", roles: ["Mail Intake"], selected_regional_office: nil>}) (1 time)
./spec/feature/intake/edit_spec.rb:357:in `block (4 levels) in <top (required)>'
```

https://circleci.com/gh/department-of-veterans-affairs/caseflow/19584?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

pkarman:
The exact behavior is that the rating and nonrating request issues are getting assigned to the wrong EndProductEstablishment records. I am investigating.

pkarman:
This was a regression caused by `profile_date` not being sent as part of `RatingIssue.ui_hash`.

PR #7712, opened 2018-11-05
Fix flakey test

pkarman:
connects #7703 

### Description

Fixes regression in skipped test.


lib/fakes/bgs_service.rb
```diff
@@ -73,7 +74,8 @@ def self.create_veteran_records
           ]
         )
         hlr = HigherLevelReview.find_or_create_by!(
-          veteran_file_number: veteran.file_number
+          veteran_file_number: veteran.file_number,
+          receipt_date: in_active_review_date
```
pkarman: we needed something to make this unique, otherwise the "find" of "find or create" will never work.

app/models/rating_issue.rb
```diff
@@ -58,6 +58,7 @@ def ui_hash
       reference_id: reference_id,
       decision_text: decision_text,
       promulgation_date: promulgation_date,
+      profile_date: profile_date,
```
pkarman: this is the fix. everything else is clean up.

lib/fakes/bgs_service.rb
leikkisa: This sounds like the date something was put in active review. Maybe "in_active_review_receipt_date"?

PR #7713, opened 2018-11-05
Anya/7588 upload decision document

### LINKED ISSUE #7588: As a BVA dispatch user [or as a judge], I need to upload the final, errorless BVA decision to VBMS, so that BVA has the correct system of record.
As a BVA dispatch user [or as a judge], I need to upload the final, errorless BVA decision to VBMS, so that BVA has the correct system of record.

aroltsch:
Resolves #7588 

### Description
Upload a decision document to VBMS and S3 when a dispatcher outcodes an AMA appeal.


app/models/concerns/uploadable_document.rb
```diff
@@ -10,6 +10,10 @@ def upload_date
     Time.zone.now
   end
 
+  def source
+    "VACOLS"
```
amprokop: it's **ok** to put this here, but might be better to define this method in form8.rb instead? not sure VACOLS is the best default value. up to you though.

amprokop:
Hm. I might be reading this wrong — does the IDT pass the file as a key?

app/models/concerns/uploadable_document.rb
aroltsch: Good point!

app/models/appeal.rb
```diff
@@ -36,6 +36,10 @@ def type
     "Original"
   end
 
+  def sanitized_vbms_id
```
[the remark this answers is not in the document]
aroltsch: yeah, I agree with you.

PR #1234, opened 2017-03-18
Refactor Appeals Repository, add Generators

shanear:
connects #1064 

Cleaned up the tests in `establish_claim_spec.rb`:
 - Reorganized them to better map to the flow
 - Added clarity to test names and variable names
 - Eliminated a lot of unnecessary test data setup
 - Combined some overlapping tests
 - Added generators, described below

With a crude test on my machine it seems to have shaved off 10-15 seconds from the test time
of `establish_claim_spec.rb`
Before refactor:
```
Finished in 1 minute 10.69 seconds (files took 3.98 seconds to load)
Finished in 1 minute 13.19 seconds (files took 4.2 seconds to load)
Finished in 1 minute 17.42 seconds (files took 4.5 seconds to load)
```

After refactor:
```
Finished in 1 minute 0.09 seconds (files took 4.05 seconds to load)
Finished in 58.26 seconds (files took 4 seconds to load)
Finished in 57.37 seconds (files took 3.98 seconds to load)
```

There is still some more work to do here, but this is a good chunk of the refactoring.

## Generators
Description for Generators in `/lib/generators/base.rb`:
```
# Base class for a generator.
#
# The purpose of generators is to easily setup data for testing.
# They are used in a similar way to FactoryGirl, but are also used
# for instrementing faked dependecies.
#
# They are useful for:
# - Setting up default boilerplate data like ids
# - Setting up and connecting the appropriate data in Fakes
# - Containing complex data fixtures
#
# Each Generator class must implement the `.build(attrs)` method
# which must return an unsaved ActiveRecord object. Generators will
# get the `create(attrs)` method for free by extending Generators::Base
```

We decided to go with Generators instead of a library like FactoryGirl because they are relatively lightweight and deal much more with setting up `Fakes` than setting up complex ActiveRecord structures (which I'd argue we want to avoid).

PR #7723, opened 2018-11-05
Removing need to choose payee_code

mdbenjam:
Connects #7524

### Description
Added a new method in the `SelectClaimant` component which only shows the PayeeCode dropdown if it is needed (i.e. the form type is either supplement claim, or higher level review and the benefit type is either pension or compensation).

Also removed the backend's inferring of payee code, and now always send it from the frontend. i.e. we no longer guess that the payee code is "00" if unspecified by the frontend, and always specify a payee code of "00" when the claimant is the Veteran.

### Testing Plan
1. Go through the supplement claim or higher level review paths.
   1. Once when choosing pension/compensation with and without non-veteran claimants
      1. Ensure that with a veteran claimant the payee code is recorded as 00
      1. Ensure that with a non veteran claimant the payee code is whatever is specified on the frontend.
   1. Once when choosing fiduciary
      1. Ensure that the payee code is nil
1. Go through the appeal path
    1. Ensure the payee code is nil


kencheeto:
A note for sierra issues - you'll want to use `Connects #1234` instead of `Resolves #1234` because we don't want the issue to auto-resolve (we manually move it over to Validation, and manually into Done after validating)

mdbenjam:
Not sure this is the best way to do it. Puts a lot of power in the frontend. Also not sure the best way to test this flow. Is it a feature test? If so should I create a new one, or add to an existing one?

app/models/appeal_intake.rb
```diff
@@ -10,7 +10,7 @@ def ui_hash(ama_enabled)
   def review!(request_params)
     detail.create_claimants!(
       participant_id: request_params[:claimant] || veteran.participant_id,
-      payee_code: request_params[:payee_code] || "00"
+      payee_code: nil
```
leikkisa: So there can still be claimants for appeals, but not payee code?

spec/models/higher_level_review_intake_spec.rb
```diff
@@ -98,7 +98,7 @@
         expect(intake.detail.claimants.count).to eq 1
         expect(intake.detail.claimants.first).to have_attributes(
           participant_id: intake.veteran.participant_id,
-          payee_code: "00"
+          payee_code: nil
```
leikkisa: I thought we still want the payee code for HLR and SC if the benefit type is compensation?  This does depend on the front-end for that, could be handled in the back end.  It's a little confusing to have a model test that doesn't exhibit the desired behavior.  Maybe add a non-comp test as well.

mdbenjam: I think handling it on the backend is cleaner, however I couldn't figure out the right way to do it. How do we differentiate: "The frontend didn't send a payee_code because it's the Veteran who is appealing and we should default to "00"" from "The frontend didn't send a payee_code because we don't need one."? Very open to suggestions here!

app/models/appeal_intake.rb
mdbenjam: Correct. I think the Board just doesn't use payee codes.

leikkisa:
Just a thought that there might be some data clean up required for Appeals

spec/models/higher_level_review_intake_spec.rb
leikkisa: I guess the easiest solution is to just specify the payee code in this test to be 00. That way the test at least makes sense.  Or you could modify create_claimants or review! in claim_review_intake.  claim_review_intake might be easiest, where you add a check for the benefit_type.

app/models/appeal_intake.rb
leikkisa: 👍

app/models/claim_review_intake.rb
```diff
@@ -12,7 +12,7 @@ def review!(request_params)
     detail.start_review!
     detail.create_claimants!(
       participant_id: request_params[:claimant] || veteran.participant_id,
-      payee_code: request_params[:payee_code] || "00"
+      payee_code: request_params[:payee_code]
```
leikkisa: Here you could add a check for the benefit type, and if it's in the right category and the claimant is the veteran, you could add "00"

app/models/claim_review_intake.rb
```diff
@@ -10,10 +10,19 @@ def ui_hash(ama_enabled)
...
+    # If there's a claimant use it, otherwise the claimant is the Veteran
+    if request_params[:claimant]
+      detail.create_claimants!(
+        participant_id: request_params[:claimant],
+        payee_code: request_params[:payee_code]
+      )
+    else
+      detail.create_claimants!(
+        participant_id: veteran.participant_id,
+        payee_code: "00"
```
leikkisa: I think this should be nil if the benefit type isn't compensation or pension.

app/models/claim_review_intake.rb
mdbenjam: Bleh sorry you'd already mentioned that and I totally forgot to do it! I was too hasty, my apologizes!

spec/models/higher_level_review_intake_spec.rb
```diff
@@ -116,6 +116,20 @@
           payee_code: "10"
         )
       end
+
+      context "And benefit type is non-compensation" do
```
mdbenjam: @leikkisa what do you think of this test instead of a feature test?

mdbenjam:
Wondering about the data cleanup. Should I add a migration to remove payee code for appeals? My guess is there is not harm in leaving it 🤷‍♂️ . But happy to add if we think it's necessary.

leikkisa:
@mdbenjam I don't think the data clean up will be hard. I'm always a fan of keeping it clean, because we don't know how it will be used downstream.  We can ask @shanear too.  Sometimes for stuff like this we just run a script and put the script in the ticket instead of running a migration.

mdbenjam:
Sounds good in that case, I think the script would just be:

```
Claimant.where(review_request_type: "Appeal").update_all(payee_code: nil)
```

mdbenjam:
As long as removing that old data is cool with @shanear

shanear:
@mdbenjam yep im good with it. You know i love squeaky clean data.

mdbenjam:
The data is updated!

Issue #7386, opened 2018-10-12
AMA Edit | Prevent editing issues on claim reviews with EPs that have already been decided


shanear:
When a HLR or SC has two EPs associated with it, it's possible the user will want to edit the issues when one of the EPs has already been cleared. We should prevent removing any issues that are part of the cleared EP.

## Technical Notes
* there is a `status_cleared?` method on the EP model
* the `sync!` method will ensure that the EP in Caseflow's db mirrors the current state in BGS

## Acceptance Criteria 
- When a user clicks the "Edit in Caseflow" link from VBMS (`/<claim review type>/:claim_id/edit`), if any of EPs associated with the claim review are cleared, then the user should instead be redirected to a splash page telling them they cannot edit the issues.

## Copy
Title: Issues Not Editable
Text: Other end products associated with this [higher level review|supplemental claim] have already been decided, so issues are no longer editable. If this is a problem, please contact Caseflow support. 

## Mocks 
The page should follow the status message style in the styleguide:
<img width="853" alt="screen shot 2018-10-16 at 4 13 17 pm" src="https://user-images.githubusercontent.com/1596259/47044857-513f4780-d15f-11e8-9471-1715b60d5672.png">

shanear:
should be validatable in UAT tomorrow via 123458318

shanear:
![screen shot 2018-11-20 at 12 40 32 pm](https://user-images.githubusercontent.com/1596259/48792201-a243dd80-ecc1-11e8-9c3e-b870df95b0f8.png)
🎆

PR #7735, opened 2018-11-05
Prevent edits on hlr/sc with cleared ep

jiahuang:
Connects #7386

image


app/models/claim_review.rb
```diff
@@ -12,6 +12,13 @@ class ClaimReview < DecisionReview
   def ui_hash(ama_enabled)
     super.merge(
       benefitType: benefit_type,
+      claimant: claimant_participant_id,
+      claimantNotVeteran: claimant_not_veteran,
+      payeeCode: payee_code,
+      legacyOptInApproved: legacy_opt_in_approved,
+      ratings: serialized_ratings,
+      requestIssues: request_issues.map(&:ui_hash),
+      hasClearedEP: has_cleared_ep?,
```
shanear: Why did we add all of this stuff here? Doesn't seem like a lot of it is getting used in this PR.

client/app/intakeEdit/pages/message.jsx
```diff
@@ -6,33 +6,38 @@ import _ from 'lodash';
 import StatusMessage from '../../components/StatusMessage';
 import { FORM_TYPES } from '../../intake/constants';
 
-class Canceled extends Component {
+class ClearedEndProducts extends Component {
```
shanear: Shouldn't this be called `Message`?

app/models/claim_review.rb
jiahuang: that was a bad conflict fix with master. I've added them in where they should be.

app/models/claim_review.rb
```diff
@@ -85,6 +86,10 @@ def on_sync(end_product_establishment)
     end
   end
 
+  def cleared_ep?
+    !end_product_establishments.select(&:status_cleared?).empty?
```
shanear: I'm wondering if we want to go with the cached value or not here? `status_cleared?` could be up to 48 hours out of date here.

pkarman: This might also be written as:

```ruby
  end_product_establishments.any?(&:status_cleared?)
```

client/app/intakeEdit/IntakeEditFrame.jsx
```diff
@@ -19,6 +17,25 @@ const textAlignRightStyling = css({
 });
 
 export default class IntakeEditFrame extends React.PureComponent {
+  displayClearedEpMessage = (details) => {
```
pkarman: it would be nice if these strings were in Constants so that we could more easily test for them.

app/models/claim_review.rb
pkarman: @shanear what is the alternative? re-syncing on demand?

client/app/intakeEdit/IntakeEditFrame.jsx
jiahuang: Some of these messages have injected variables. I guess I could do some string replacement to slot in those variables, but currently we're only using constants for static strings & I'd like to keep it that way for now.

app/models/claim_review.rb
jiahuang: updated to resync before ui display.

PR #7737, opened 2018-11-05
fix bgs call

jiahuang:
Connects #6214

🤦‍♀️ fixes the bgs cancel call to actually pass in data.


spec/models/request_issues_update_spec.rb
```diff
@@ -348,6 +348,12 @@
           issue_category: "Apportionment"
         )
 
+        expect_any_instance_of(Fakes::BGSService).to receive(:cancel_end_product).with(
+          veteran.file_number,
+          "030HLRNR",
+          "030"
+        )
```
shanear: nice 👍

Issue #7534, opened 2018-10-24
Make case details sidebar responsive


lowellrex:
Currently, at small screen widths the table of contents sidebar become hard to read and rob us of valuable screen real estate. This ticket exists to make the sidebar collapse into a smaller version of itself at small screen widths.

### Current state
![image](https://user-images.githubusercontent.com/32683958/47465768-58e49900-d7bc-11e8-910b-95530629cd4f.png)

### Future mocks
![image](https://user-images.githubusercontent.com/1598889/42772397-4c5eb7b4-88f8-11e8-8c17-d5014316502c.png)

PR #7740, opened 2018-11-06
Make case details sidebar responsive

joeyyang:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/7534

### Description
Makes case details sidebar responsive. Sidebar collapses to top and sticks to top of page when scrolled to, and main content section assumes full width. Sidebar links jump to appropriate section with scroll offset.

Question:
- Made the sidebar text size smaller per the design in https://github.com/department-of-veterans-affairs/caseflow/issues/7534 (`13px`). Is the text size too small?

nov-06-2018 08-10-56


client/app/queue/StickyNavContentArea.jsx
```diff
@@ -76,8 +120,18 @@ const sectionHeadingStyling = css({
...
+  paddingTop: '60px',
+  textDecoration: 'none',
+  pointerEvents: 'none',
+  cursor: 'default'
+});
+
 const ContentSection = ({ element }) => <React.Fragment>
-  <h2 id={`${getIdForElement(element)}`} {...sectionHeadingStyling}>{element.props.title}</h2>
+  <h2 {...sectionHeadingStyling}>
+    <a id={`${getIdForElement(element)}`} {...anchorJumpLinkStyling}>{element.props.title}</a>
```
joeyyang: Unfortunately there's no way to do the jump-with-offset and still have a top rounded border on the `h2` without adding an anchor tag inside. The `anchorJumpLinkStyling` nulls out the styles and interactivity added from the link.

client/app/queue/StickyNavContentArea.jsx
```diff
@@ -9,7 +9,23 @@ const sectionNavigationContainerStyling = css({
   paddingRight: '3rem',
   position: 'sticky',
   top: '3rem',
-  width: '20%'
+  width: '20%',
+  '@media (max-width: 920px)': {
```
lowellrex: Let's use a max-width of 1200px to [match when we move](https://github.com/department-of-veterans-affairs/caseflow/blob/b4f7b55bfb115e3ce4525dd9411d211b4b8ab64c/client/app/queue/CaseSnapshot.jsx#L55) the task dropdown underneath the rest of the case snapshot.

client/app/queue/StickyNavContentArea.jsx
```diff
@@ -9,7 +9,23 @@ const sectionNavigationContainerStyling = css({
...
+    top: '1rem',
```
lowellrex: Let's hide the content of the page when it goes "underneath" the nav bar like in the screenshot below:
image

I think we can do this by replacing `top: '1rem'` with `paddingTop: '1rem'` and adding a `background: COLORS.WHITE` property here as well.

client/app/queue/StickyNavContentArea.jsx
```diff
@@ -30,12 +46,40 @@ const sectionNavigationListStyling = css({
...
     content: '〉',
     float: 'right'
+  },
+  '@media (max-width: 920px)': {
+    display: 'flex',
+    flexFlow: 'row',
+    borderBottom: 'none',
+    borderTop: 'none',
+    width: '100%',
+    marginBottom: '30px',
```
lowellrex: We prefer to use `rem` units or percentages when possible (which I think is possible for these pixels units).

client/app/queue/StickyNavContentArea.jsx
sneha-pai: What about a warm-grey container around the case timeline menu? 
screen shot 2018-11-06 at 5 32 13 pm

sneha-pai: Is this already in there? It's hard to see it in the gif captured above.

sneha-pai:
Re: Made the sidebar text size smaller per the design in #7534 (13px). Is the text size too small?

This text may be smaller than usual at 13px (let's keep it for now - there's a styleguide conversation that will happen soon to see if we want to adopt 13/14px as one of our standard sizes (right now our body is 17px) 

 Also, the category of Hearings is missing in the menu option. When that is added, even at 13px there seem to be too many items width-wise. Would love to create either better wrapping or overflow for this horizontal menu treatment. Will send some sketches your way shortly

client/app/queue/StickyNavContentArea.jsx
joeyyang: Just want to make sure I'm thinking of the same thing — does the warm-grey container refer to the dark-ish gray around the word "issues", or something else? If so, it's already there, but only on hover, per the inherited styles from the `usa-sidebar-list` class.

joeyyang:
Made a few changes:

screenshot 2018-11-07 13 35 14
screenshot 2018-11-07 13 35 21

joeyyang:
> Made a few changes, please review and approve:
> 
> screenshot 2018-11-07 13 35 14
> 
> screenshot 2018-11-07 13 35 21

@sneha-pai, I don't see a "Hearings" menu item — is it supposed to be in this view?

screenshot 2018-11-07 11 03 28

laurjpeterson:
@joeyyang @sneha-pai - I think this is just related to the test/dummy data in this appeal record. 

The screenshot above shows a Direct Review AMA appeal, so there won't be hearing data associated with it. Hearing AMA appeals would have hearing data associated with it.

Legacy appeals where the Veteran requested a hearing will have data associated with it.

client/app/queue/StickyNavContentArea.jsx
joeyyang: Ideally, the highlighted tab should change as you scroll through. This would be a fairly complex change, so I'm going to punt it to another PR.

Re: warm grey and darker grey: could you point me to which colors in this file match those descriptions? https://github.com/department-of-veterans-affairs/caseflow/blob/47e661470a49a90ead41460ee64d0fd6cddcea92/client/app/constants/AppConstants.js

client/app/queue/StickyNavContentArea.jsx
```diff
@@ -30,12 +47,40 @@ const sectionNavigationListStyling = css({
...
+  '@media (max-width: 1020px)': {
+    display: 'flex',
+    flexFlow: 'row',
+    borderBottom: 'none',
+    borderTop: 'none',
+    width: '100%',
+    marginBottom: '1rem',
+    '& > li': {
+      padding: '8.5px 10px',
+      fontSize: '13px',
```
lowellrex: I think we should redefine these padding values and font size in rem.

client/app/queue/StickyNavContentArea.jsx
```diff
@@ -76,8 +121,18 @@ const sectionHeadingStyling = css({
   padding: '1rem 2rem'
 });
 
+const anchorJumpLinkStyling = css({
+  color: COLORS.GREY_DARK,
+  paddingTop: '72px',
```
lowellrex: Let's use rem.

joeyyang:
Great! I've made @lowellrex's suggested changes. @sneha-pai, can you give a final approval?

screenshot 2018-11-16 18 29 16

sneha-pai:
@joeyyang to confirm, the light warm-grey that's being used in many places in case details is hex: #F8F8F8 

This is the same warm grey in the inactive horizontal bar. I'm not sure this has ever been added to the official styleguide, but will soon be (since it's already being used in many places)

PR #7743, opened 2018-11-06
Upcoming Hearings - Available slots logic

### LINKED ISSUE #7667: Upcoming Hearing Days | Available slots logic
- For upcoming hearing days - show all hearing days with available slots
- Available slots = total slots - postponed hearings - cancelled hearings
- Move cancelled hearings to the Previously Scheduled section on the Daily Docket

oscarAtNava:
Resolves #7667 

### Description
Fixed in back-end. Return to UI Scheduled Hearings that meet the following criteria:
- Count CO hearings where folder_nr is not null.
- Count Video child hearings where hearing disposition is neither postponed or cancelled.

Return the hearing day to the UI if and only if total slots for RO/CO is greater than the number of hearings counted with above logic.

UI calculates the available slots by subtracting the total_slots value with the number of scheduled hearings it received (current logic, not changed).

### Acceptance Criteria
- [ ] Code compiles correctly
- [ ] Additional tests pass

### Testing Plan
1. Set up test data in FACOLS for CO. Eleven slots. 
2. Have all slots with folder_nr not null.
3. Confirm the date is not shown in UI
4. Nullify folder_nr in a couple of rows.
5. Confirm date does appear in UI and is available for scheduling.
6. Confirm the available slots is equal to the number of rows you nullified.
7. Add child video rows for any RO parent record
8. Set hearing disposition to P or C
9. Confirm only child rows with disposition other than P or C return to the UI.
10. Confirm available slots calculation equals the total_slots for RO minus the number of hearings returned by back-end for Scheduled Hearings tab.


app/models/hearing_day.rb
```diff
@@ -66,21 +66,40 @@ def load_days(start_date, end_date, regional_office = nil)
 
     def load_days_with_hearings(start_date, end_date, regional_office = nil)
```
sharonwarner: Let's change this method name to reflect that we're only pulling hearing days with available slots.

oscarAtNava: Renamed to load_days_with_open_hearing_slots

PR #7751, opened 2018-11-06
explain how to fix security audit failures

kencheeto:
### Description
We've seen a couple times where Brakeman fails CI on a security audit of the installed gems. Adding an explanation of how to address when this comes up so people don't get stuck.

In addition to the audit findings (which list a CVE and the gem in violation), the rake task will also print the following:
```
Failed. Security vulnerabilities were found. Find the dependency in Gemfile.lock, then specify a safe version of the dependency in the Gemfile (preferred) or snooze the CVE in security.rake for a week.
See https://github.com/department-of-veterans-affairs/caseflow/pull/7639 for an example.
```

### Acceptance Criteria
N/A

### Testing Plan
N/A

### DIFF
--- a/lib/tasks/security.rake
+++ b/lib/tasks/security.rake
@@ -14,11 +14,11 @@ task :security_caseflow do
 
   snoozed_cves = [
     # Example:
-    # { cve_name: 'CVE-2018-1000201', until: Time.zone.local(2018, 9, 10) }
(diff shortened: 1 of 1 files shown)


kencheeto:
I should note that I'm basically recycling code here: https://github.com/department-of-veterans-affairs/caseflow/pull/7657 but generalized so it's easy to temporarily ignore CVEs

Issue #7753, opened 2018-11-07
Intake | BGSyncError


pkarman:
Steady stream of these production errors in the last 72 hours.

```
EndProductEstablishment::BGSSyncError

PG::QueryCanceled: ERROR:  canceling statement due to statement timeout
CONTEXT:  while updating tuple (1062,54) in relation "end_product_establishments"
: UPDATE "end_product_establishments" SET "last_synced_at" = $1 WHERE "end_product_establishments"."id" = $2
```

Likely due to the fact that all the upstream syncing is now taking longer than the default PG transaction length.

pkarman:
may be mitigated by #7652

pkarman:
this is still happening as of 11/13 so #7652 did not help (enough).

kencheeto:
Blocked because we are waiting on BGS to replicate these issues. cc @mc-usds for visibility

kencheeto:
Action item while we wait for BGS: investigate other ways to process these intakes that are faster

pkarman:
PR #8086 should work around the slow call, and VBMS 1.6 has some improvements that we should evaluate for performance time.

pkarman:
putting this in validation so we can monitor alerts channel and see if it gets quieter.

steve-gov:
Dec 4: 26 `BGSyncError` alerts
Dec 10: 3 `BGSyncError` alerts

How long do you think we should monitor before passing validation?

pkarman:
The particular error that spawned this ticket has been worked around, but the overall slow issue remains. Let's close this one. I opened #8227 to track the larger issue.

PR #7754, opened 2018-11-07
Do not wrap the upstream sync and the db update in a transaction

pkarman:
connects #7753 

### Description
If upstream syncs take longer than the psql transaction timeout length, the whole transaction will fail.

This may be the desired behavior, because it ensures that the sync will be attempted again later.

Alternative would be to increase the psql transaction timeout length, which is currently 30 seconds.

in `config/database.yml`
```
statement_timeout: 30_000 # 30 seconds
```

### Acceptance Criteria
- [ ] Code compiles correctly

### DIFF
--- a/app/models/end_product_establishment.rb
+++ b/app/models/end_product_establishment.rb
@@ -160,14 +160,8 @@ class EndProductEstablishment < ApplicationRecord
 
     fail EstablishedEndProductNotFound unless result
 
-    transaction do
-      update!(
-        synced_status: result.status_type_code,
-        last_synced_at: Time.zone.now
-      )
-
-      sync_source!
-    end
+    sync_source!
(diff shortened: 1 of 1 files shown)


app/models/end_product_establishment.rb
```diff
@@ -160,14 +160,8 @@ def sync!
...
-    transaction do
-      update!(
-        synced_status: result.status_type_code,
-        last_synced_at: Time.zone.now
-      )
-
-      sync_source!
-    end
+    sync_source!
+    update!(synced_status: result.status_type_code, last_synced_at: Time.zone.now)
```
shanear: I think the fact that we are doing this all inside a transaction was intentional. Do we know what data issues could arise by making this block not atomic for our DB?

PR #7759, opened 2018-11-07
Support delayed rating issue sync

pkarman:
connects #7652 

### Description
Sometimes an EndProductEstablishment gets its status set to cleared (meaning a decision was rendered on the contention) but the corresponding RatingIssue won't be posted for some indeterminate amount of time (days).

This PR makes the decision sync eventually consistent, by adding the `Asyncable` concern to `RequestIssue` and adding a new `DecisionRatingIssueSyncJob` class.

### Acceptance Criteria
- [x] Code compiles correctly
- [x] Tests continue to pass


app/jobs/decision_rating_issue_sync_job.rb
```diff
@@ -0,0 +1,21 @@
...
+  application_attr :intake
+
+  def perform(request_issue)
+    RequestStore.store[:application] = "intake"
+    RequestStore.store[:current_user] = User.system_user
+
+    veteran = request_issue.review_request.veteran
+    begin
+      request_issue.attempted!
+      veteran.sync_rating_issues!([request_issue])
```
shanear: I had a gut reaction to batch these, but i think since we are caching the ratings it should be fine.

Which brings me to another point. Should we reset the rating cache some how as part of this whole process?

app/models/end_product_establishment.rb
```diff
@@ -215,6 +213,14 @@ def generate_tracked_item!
     end
   end
 
+  def request_issues
+    source.request_issues.select { |ri| ri.end_product_establishment == self }
+  end
+
+  def active_request_issues
+    request_issues.select { |request_issue| request_issue.removed_at.nil? }
```
shanear: i don't think this is the test for `active`. Which makes me think we want to use a method here:

```suggestion
    request_issues.select(&:status_active?)
```

app/models/veteran.rb
```diff
@@ -150,9 +150,21 @@ def participant_id
     super || ptcpnt_id
   end
 
-  def sync_rating_issues!
```
shanear: I feel like this method is batched awkwardly, and feels out of place as a method on `Veteran`. Since everything is cached, it seems like there isn't a downside to making `RequestIssue.sync_decisions!` and loop over that:

```
RequestIssue.requires_processing.map(&:sync_decisions!)
```

app/jobs/decision_rating_issue_sync_job.rb
pkarman: `veteran.sync_rating_issues!` does not use or touch the cached rating issues. It pulls real-time response from BGS, using the "timely" window (the last year). The cache ignores the timely window.

Your point about batching is well taken, though. If there are multiple request issues unsynced for each veteran, it's going to make a real-time BGS call for each one. I think that is what we want, but it might have some unforeseen consequences. E.g., if there are 2+ request issues in the processing queue, the first one touched is going to trigger a sync on all of them. That means there will be needless BGS calls and db updates. We can probably add an optimization to look for the `request_issue.processed?` flag *before* we call `veteran.sync_rating_issues!` and return if it is true.

app/models/end_product_establishment.rb
pkarman: I agree that "active" and "not removed" are different things. I want to make sure the existing logic is respected though, since the name "active" is specifically attached to "not removed" and the intent may just be mis-named.

For *this* PR though, we do want `status_active?` and I will use that.

app/models/veteran.rb
pkarman: we do it on the `Veteran` because `timely_ratings` is a method there. It's not using the same cached Ratings that are on the `DecisionReview` class.

app/models/end_product_establishment.rb
shanear: yeah we need to figure out the wording there. I'm thinking:

active = not resolved and not removed
complete = has decision and is "finalized"

i don't have a good word for "not removed"

pkarman: What do you mean by "not resolved"? That `status_active?` is true?

```diff
@@ -218,7 +221,7 @@ class EndProductEstablishment < ApplicationRecord
   end

   def active_request_issues
-    request_issues.select { |request_issue| request_issue.removed_at.nil? }
+    request_issues.select { |request_issue| request_issue.removed_at.nil? && request_issue.status_active? }
   end
```

app/models/veteran.rb
pkarman: I have moved the sync to `EndProductEstablishment.sync_decision_issues!`

PR #7764, opened 2018-11-07
ineligible enum should be string not integer

pkarman:
### Description
Our convention is to use `enum` columns as string values for easier use outside the Rails app.

### DIFF
--- a/db/migrate/20181107182512_request_issue_enum_as_string.rb
+++ b/db/migrate/20181107182512_request_issue_enum_as_string.rb
@@ -1,0 +1,20 @@
+class RequestIssueEnumAsString < ActiveRecord::Migration[5.1]
+  def up
+    # add a new string column, convert the data, then drop the old column and rename the new
+    add_column :request_issues, :ineligible_reason_str, :string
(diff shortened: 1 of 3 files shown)


db/migrate/20181107182512_request_issue_enum_as_string.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def down
+    add_column :request_issues, :ineligible_reason_int, :integer
+    execute "UPDATE request_issues SET ineligible_reason_int=0 WHERE ineligible_reason='duplicate_of_issue_in_active_review'" 
+    execute "UPDATE request_issues SET ineligible_reason_int=1 WHERE ineligible_reason='untimely'"
+    execute "UPDATE request_issues SET ineligible_reason_int=2 WHERE ineligible_reason='previous_higher_level_review'"
+    remove_column :request_issues, :ineligible_reason
+    rename_column :request_issues, :ineligible_reason_int, :ineligible_reason
+  end
```
shanear: there may be a blip where the data and the code don't match up, but I think I'm okay with that. May deploy at 4 or 5 PT tonight to prevent the effects of that.

PR #7767, opened 2018-11-07
Add more VSO actions

### LINKED ISSUE #7480: Allow VSO users to assign cases to VSO users
When a VSO user is viewing AMA appeals needing an IHP, they want to assign the case to a member of their organization, so the IHP can be written.

## AC
- In the Actions dropdown of Case details, a VSO user has the option to assign the task to a user within their organization

## Related
https://github.com/department-of-veterans-affairs/caseflow/issues/9693

mdbenjam:
Resolves #7480 

### Description
We introduce a new task type `InformalHearingPresentationTask` which will be used to assign to VSOs. We allow these tasks to have the `assign to person` action

### Testing Plan
1. Log in as VSO user
1. Go to /organizations/american-legion
1. Click into only task
1. Assign task to the user
1. Go to the user page and complete the task.


app/controllers/tasks_controller.rb
```diff
@@ -2,7 +2,7 @@ class TasksController < ApplicationController
   include Errors
 
   before_action :verify_task_access, only: [:create]
-  skip_before_action :deny_vso_access, only: [:index, :update, :for_appeal]
+  skip_before_action :deny_vso_access, only: [:create, :index, :update, :for_appeal]
```
mdbenjam: Do we need some more permission checking on create? Maybe we should check if this is a VSO user that they can only create a task assigned to another VSO user?

lowellrex: We could check in the create function like the following:
```ruby
  raise Caseflow::Error::ActionForbiddenError if current_user.vso_employee? && task_class != InformalHearingPresentationTask
```

db/seeds.rb
```diff
@@ -66,8 +66,7 @@ def create_vso_user
       full_name: "VSO user associated with PVA",
```
lowellrex: Since we change the name of the organization the user is associated with I think we should change the name of this user as well.

PR #7768, opened 2018-11-07
Moving all tasks assigned to VSOs to IHP tasks

mdbenjam:
Connects #7480

### Description
Moving tasks assigned to VSOs to IHP tasks, and removing the check for VSOs in GenericTasks

### Testing Plan
1. Run the migration locally and ensure the GenericTasks assigned to VSO users are changed to IHP tasks.

### DIFF
--- a/db/migrate/20181107185230_change_generic_tasks_assigned_to_vsos_to_ihps.rb
+++ b/db/migrate/20181107185230_change_generic_tasks_assigned_to_vsos_to_ihps.rb
@@ -1,0 +1,8 @@
+class ChangeGenericTasksAssignedToVsosToIhps < ActiveRecord::Migration[5.1]
+  def up
+    GenericTask.joins("JOIN organizations ON assigned_to_id=organizations.id").
+      where("assigned_to_type=?", "Organization").
+      where("organizations.type=?", "Vso").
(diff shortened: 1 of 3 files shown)


db/migrate/20181107185230_change_generic_tasks_assigned_to_vsos_to_ihps.rb
```diff
@@ -0,0 +1,8 @@
+class ChangeGenericTasksAssignedToVsosToIhps < ActiveRecord::Migration[5.1]
+  def up
+    GenericTask.joins("JOIN organizations ON assigned_to_id=organizations.id").
+      where("assigned_to_type=?", "Organization").
+      where("organizations.type=?", "Vso").
+      update_all(type: "InformalHearingPresentationTask")
```
lowellrex: I think we can just do 
```ruby
GenericTask.
  where("assigned_to_type=?", "Vso").
  update_all(type: "InformalHearingPresentationTask")
```

But I think this way works as well.

mdbenjam: That's what I thought as well, but looking at tasks locally it seems `assigned_to_type` is still organization 🤷‍♂️

PR #7769, opened 2018-11-07
Add untimely exemption modal to nonrating issues

pkarman:
connects #7301 

### Description
The nonrating issue should also prompt the untimely exemption modal.

See https://github.com/department-of-veterans-affairs/caseflow/issues/7301#issuecomment-436436770

### Acceptance Criteria
- [x] Code compiles correctly
- [x] Adding a nonrating issue with a decision date outside the "timely" definition should prompt the untimely exemption modal.


spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -690,7 +690,9 @@ def check_row(label, text)
       fill_in "Decision date", with: "04/19/2016"
       expect(page).to have_button("Add this issue", disabled: false)
       safe_click ".add-issue"
+      add_untimely_exemption_response("No", "I am a nonrating exemption note")
       expect(page).to have_content("6 issues")
+      expect(page).to have_content("I am a nonrating exemption note")
```
leikkisa: I would add the check that this gets added to the database properly.

leikkisa:
Oh, I forgot (the whole reason you asked me to review). The untimeliness exemption field needs to be added to request_issue_update where request issues are created.

app/models/request_issue.rb
```diff
@@ -196,7 +196,7 @@ def check_for_active_request_issue!
     return unless rating?
     return unless eligible?
     existing_request_issue = self.class.find_active_by_reference_id(rating_issue_reference_id)
-    if existing_request_issue
+    if existing_request_issue && existing_request_issue.review_request != review_request
```
pkarman: this should fix the remove and re-add bug

leikkisa: We might want to also add some tests for that bug too.  If we go with removing and re-adding the issue if they do it in Edit, because they could have changed notes, there are still a couple of options. I think what Shane suggested is removing it altogether which would mean removing the contention, and creating it again, so we could check that.  However, that's not really necessary since notes are just in the Caseflow DB, so we could also just update the notes.  Also checking the front end presentation as an eligible issue.

I'll put these notes in the other ticket.

Issue #6842, opened 2018-09-05
Intake | Only show valid payee codes for HLRs and SCs


shanear:
Right now, we are showing the complete list of possible payee codes. However there is only a subset of payee codes that are valid for any EP code. There are a variety of BGS endpoints where we can retrieve this, but it might be worth hard coding them in for now.

## Acceptance Criteria
- On the HLR and SC intake flow, the list of payee codes is filtered to only valid payee codes for those EPs & benefit type (valid list needs to be determined).

## Investigation
- We need to determine from the VBMS UI the list of payee codes that are valid for these modifiers. There are potentially other ways to find out

## Notes
See more information here: https://github.com/department-of-veterans-affairs/caseflow/issues/5878

leikkisa:
Some info here: https://github.com/department-of-veterans-affairs/caseflow/issues/5878

shanear:
Works for living vets. Not working for deceased vets:

Tried with this test vet
<img width="669" alt="screen shot 2018-11-09 at 6 17 00 pm" src="https://user-images.githubusercontent.com/1596259/48293155-bbd86000-e44b-11e8-8f5e-bfe1c3f0e896.png">

Showing 00 as an option:
<img width="1212" alt="screen shot 2018-11-09 at 6 17 06 pm" src="https://user-images.githubusercontent.com/1596259/48293159-c266d780-e44b-11e8-814e-0f54a0511691.png">

shanear:
UAT file number for a deceased veteran
564930680 (deceased, with relationships)

shanear:
123458300 living

steve-gov:
Validates.

HLR living:

<img width="849" alt="screen shot 2018-11-16 at 2 18 38 pm" src="https://user-images.githubusercontent.com/25468689/48642480-f772ad00-e9aa-11e8-83c5-7479d0ce7263.png">

SC living:

<img width="850" alt="screen shot 2018-11-16 at 2 19 15 pm" src="https://user-images.githubusercontent.com/25468689/48642481-f772ad00-e9aa-11e8-9f3e-8b6b8b33bdbb.png">

HLR deceased:

<img width="844" alt="screen shot 2018-11-16 at 2 19 56 pm" src="https://user-images.githubusercontent.com/25468689/48642482-f772ad00-e9aa-11e8-82d2-3b520aa9edd7.png">

SC deceased:

[...]

PR #7770, opened 2018-11-07
filter out valid payee codes

jiahuang:
Connects #6842


jiahuang:
Not sure the best way to test this, probably just a JS test? Feature test is hard w/ capybara since we're making up the select dropdown by inserting divs in the dom.

PR #7781, opened 2018-11-08
One touch hearing

### LINKED ISSUE #7691: Support One Touch Hearings
We have been asked to implement a checkbox, like Jed has done in DAS, to identify a One Touch Hearing. I think we have two options:

**Option 1:** 
- Replicate what Jed implemented in DAS and add a check box to Caseflow judge check out screen for legacy cases
- Show in Case details, hearing section, that the case is a One Touch Hearing

### VACOLS logic
- The field is DECASS.DE1TOUCH, it's a checkbox 'Y' if checked 'N' or null if not checked.
- https://github.com/department-of-veterans-affairs/dsva-vacols/issues/44

## AC
- Checkbox `One Touch Initiative` exists on the `Evaluate attorney screen` of Judge checkout 
- When clicked, field DECASS.DE1TOUCH, it's a checkbox 'Y' if checked 'N' or null if not checked

#### VACOLS logic
- The field is DECASS.DE1TOUCH, it's a checkbox 'Y' if checked 'N' or null if not checked.
- https://github.com/department-of-veterans-affairs/dsva-vacols/issues/44

#### SOP
image

## Previous ideas:
**Option 2:**
- In Hearing Prep, give judges the option to make it a One Touch Hearing case (for legacy). that case should show up in the Assign cases part of the judge's queue so they can assign it to an attorney to draft
- Show in Case details, hearing section, that the case is a One Touch Hearing

mdbenjam:
Resolves #7691 

### Description
Adds a checkbox for Judges to mark a (legacy only) case as a one touch hearing. This change propagates to a DECASS record. Since the column in VACOLS was not there yet, this PR also updates the VACOLS schema.

### Testing Plan
1. Go to BVAAABSHIRE's Queue and mark a legacy case as ready for dispatch
1. Click the one touch hearing box
1. Ensure the DECASS record's `de1touch` column is now "Y"


db/migrate/20181107225536_add_one_touch_initiative_to_judge_case_reviews.rb
```diff
@@ -0,0 +1,5 @@
+class AddOneTouchInitiativeToJudgeCaseReviews < ActiveRecord::Migration[5.1]
+  def change
+    add_column :judge_case_reviews, :one_touch_initiative, :boolean
```
aroltsch: If you don't set a default value, would it set existing records to nil when you run this migration?

mdbenjam: I think so, should I set it to false? I'm wondering if nil is better since that represents the fact that we don't actually know.

aroltsch: Excellent question.

PR #7782, opened 2018-11-08
Rkreyhsig/7681 hide link to edit aod from people who cannot edit aod

### LINKED ISSUE #7681: Hide link to edit AOD from people who cannot edit AOD
Currently everybody can see the link to edit a case's type and set AOD status. We should limit who can see this link to only members of the AOD team because they are the only ones who should be able to change AOD status.

image

## Acceptance Criteria
* Limit the set of people who can see the edit AOD link to only AOD team

rkreyhsig:
Resolves #7681 - Hide link to edit AOD from people who cannot edit AOD


client/app/queue/components/OnHoldLabel.jsx
```diff
@@ -6,6 +6,7 @@ export const numDaysOnHold = (task: Task) => moment().startOf('day').
   diff(task.placedOnHoldAt, 'days');
 
 type Props = { task: Task };
-const OnHoldLabel = (props: Props) => `${numDaysOnHold(props.task)} of ${props.task.onHoldDuration || '?'}`;
+const OnHoldLabel = (props: Props) => `${numDaysOnHold(props.task)}${props.task.onHoldDuration ? ` of ${
+  props.task.onHoldDuration}` : ''}`;
```
lowellrex: I think this change can be reverted since it will be included in #7742 when that is merged.

Gemfile.lock
```diff
@@ -188,7 +188,7 @@ GEM
       terminal-table (~> 1)
     database_cleaner (1.7.0)
     diff-lcs (1.3)
-    docile (1.1.5)
+    docile (1.3.1)
```
lowellrex: I think we can revert this file (Gemfile.lock) because I think these changes were the result of us trying to troubleshoot a local environment.

lowellrex:
One more thing, since we now have an AodTeam organization let's also check that the current user is a member of the AodTeam when they attempt to [edit the AOD status in the controller](https://github.com/department-of-veterans-affairs/caseflow/blob/1928b8312cb745ea420a96a804c46559a16c9edc/app/controllers/advance_on_docket_motions_controller.rb#L3) as well.

app/controllers/advance_on_docket_motions_controller.rb
```diff
@@ -9,6 +11,11 @@ def create
     render json: {}
   end
 
+  def verify_aod_access
+    fail Caseflow::Error::ActionForbiddenError, message: "User does not belong to AOD team" unless
+      AodTeam.singleton.user_has_access?(current_user)
```
lowellrex: I like it!

lowellrex: Let's add a test for this scenario in a new test file for this controller that mirrors [other controller tests](https://github.com/department-of-veterans-affairs/caseflow/blob/master/spec/controllers/beaam_appeals_controller_spec.rb).

Issue #7360, opened 2018-10-11
Add Assigned to for AMA/RAMP appeals


laurjpeterson:
When Lit Support is trying to understand the story of an appeal so they can answer status/congressional inquiries, they want to see who at the Board currently has responsibility for the appeal, so they can complete their investigation.

## AC
- Populate the CSS ID of the user who currently has the appeal in the `Assigned to` column of the Case list for AMA appeals.

<img width="1072" alt="screen shot 2018-10-11 at 1 40 47 pm" src="https://user-images.githubusercontent.com/27704093/46823235-9e847900-cd5b-11e8-86bc-6729462306e8.png">

lowellrex:
If there is no attorney assigned to an appeal (or if the attorney has already drafted a decision or requested OMO), then display the CSS ID or organization name of the entity with most immediate responsibility for the appeal (most-nested child task, tie-breaking on most recently created if need be).

lowellrex:
After chatting with @laurjpeterson the final criteria should actually just be whoever has the most recent open task for an appeal.

@laurjpeterson if there are no open tasks for an appeal (case hasn't been distributed or the decision has been dispatched) what should this "assigned to" field read?

lowellrex:
I like `Case storage` for cases that haven't been distributed yet because it is language folks already know, and I think `Post decision` is slightly preferable to `N/A` since the former is a little more precise about what state the case is in.

laurjpeterson:
Sounds good - let’s go with that! @sneha-pai, @cmgiven, or Amanda might have opinions to change it down the line, but I think we can go with these terms for now to get this out there

PR #7783, opened 2018-11-08
Add Assigned to for AMA/RAMP appeals

joeyyang:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/7360

### Description
Populate the CSS ID of whoever has the most recent task for an appeal in the Assigned to column of the Case list for AMA appeals. If case is assigned to the current user, show `Assigned to you`. If case has not distributed yet, show `Case storage`, and show `Post decision` if the root task is marked as completed.

screenshot 2018-11-08 16 32 39

@lowellrex — please let me know any way I can clean up any of the `.rb` here — including if there's more idiomatic ways to write the tests and if there's a way to clean up the logic in `appeal.rb`.

/cc @laurjpeterson


app/models/appeal.rb
```diff
@@ -43,6 +43,25 @@ def type
...
+    return nil if !root_task
+
+    # if RootTask status:completed, then return "Post-decision"
+    return "Post-decision" if root_task.status == Constants.TASK_STATUSES.completed
+
+    active_tasks = tasks.where(status: [Constants.TASK_STATUSES.in_progress, Constants.TASK_STATUSES.assigned])
+    return nil if active_tasks.empty?
+
+    # if the only active case is a RootTask, then return "Case storage"
+    return "Case storage" if active_tasks == [root_task]
```
lowellrex: Let's put these two strings ("Post-decision" and "Case storage") in COPY.json so they can be more easily edited by non-developers.

app/models/appeal.rb
```diff
@@ -43,6 +43,25 @@ def type
     "Original"
   end
 
+  def location_code
+    return nil if tasks.empty?
+
+    root_task = tasks.first.root_task
+    return nil if !root_task
+
+    # if RootTask status:completed, then return "Post-decision"
```
lowellrex: I think we can remove both of these code comments since they describe the single line beneath both of them. I generally find code comments most useful when they aim to describe the "why" behind what is going on in the code. If we want to add a comment along those lines (and I think we might here) I think it could read something like "RootTasks are marked complete when the decision is dispatched from the Board, so we are using that as our signal that we are "Post-decision"".

app/models/appeal.rb
```diff
@@ -43,6 +43,25 @@ def type
     "Original"
   end
 
+  def location_code
```
lowellrex: Let's add a quick comment that explains why we have decided to add a function called "location_code" to the new AMA appeal class. Perhaps something like, "Returns the most directly responsible party for an appeal when it is at the Board, mirroring Legacy Appeals' location code in VACOLS".

spec/models/appeal_spec.rb
```diff
@@ -185,4 +185,59 @@
...
+      end
+      it "returns Case storage" do
+        expect(appeal.location_code).to eq("Case storage")
+      end
+    end
+
+    context "if there is an assignee" do
+      let(:organization) do create(:organization, name: "VSO") end
+      let(:appeal_organization) do create(:appeal) end
+      let(:user) do create(:user, css_id: "69") end
```
lowellrex: The user and organization factory automatically generate [user css_ids](https://github.com/department-of-veterans-affairs/caseflow/blob/1b9a77a5aab5f1b5847b838e560ff5f1275ae4be/spec/factories/user.rb#L3) and [organization names](https://github.com/department-of-veterans-affairs/caseflow/blob/1b9a77a5aab5f1b5847b838e560ff5f1275ae4be/spec/factories/organization.rb#L3) so we don't have to set those here. We can also reference the variables in the tests as `organization.name` instead of `"VSO"` and `user.css_id` instead of `"69"` so that the test is a little more resilient to future changes.

joeyyang:
thanks @lowellrex, back to you!

spec/models/appeal_spec.rb
```diff
@@ -185,4 +185,59 @@
...
+  context "#location_code" do
+    context "if the RootTask status is completed" do
+      let(:appeal) do create(:appeal) end
+
+      before do
+        create(:root_task, appeal: appeal, status: :completed)
+      end
+
+      it "returns Post-decision" do
+        expect(appeal.location_code).to eq("Post-decision")
```
lowellrex: Let's use the constants from COPY.json here as well so these tests won't fail if we decide to change how that copy reads.

Issue #7334, opened 2018-10-10
Legacy Opt-in | Add issue | Rated and non-rated issue modals


suzchap:
## Story
[Mural showing all opt-in screens](https://app.mural.co/t/workqueue2001/m/workqueue2001/1538676663322/098da798af658862105ea8d4d5aeb03fa7aaf99a)

When a user has a legacy appeal, they will be allowed to "opt-in" to move one or more of those issues into an AMA lane. The AMA paper forms will have a question on them for the applicant to acknowledge that by checking that box, they understand that any legacy issues listed on that form will be moved out of the legacy lane and into the new AMA lane.

This ticket is focused on changing the primary call to action button label, it's destination, and what happens if there isn't a rated issue match.

## Technical notes
- A legacy appeal is active if `active?` is true 🙂(see: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/legacy_appeal.rb#L494)
- A legacy appeal is eligible if the `soc_date` is set AND either the `soc_date` is within the last 60 days OR the `nod_date` is within the last 372 days. Note that the legacy appeal does not have to be active to be eligible.

## Acceptance criteria
- On the add rating issue and the add non-rating issue modals, the primary call-to-action button is normally labeled "Add this issue" but if the veteran has an active and/or eligible legacy appeal, the button instead should say "next". See technical notes for definitions of eligible and active

## Notes
- Clicking "next" will eventually take the user to the VACOLS modal #7335 , but make it a placeholder for now
- On the "unidentified issue" modal, there is no change (they do NOT get presented with the VACOLS modal)
- If there is no eligible or active legacy appeal for the veteran, there is no change (no VACOLS modal for any sort of issue)

## Mocks
<img width="657" alt="screen shot 2018-10-10 at 2 15 11 pm" src="https://user-images.githubusercontent.com/11219417/46757054-00c77600-cc97-11e8-9354-fb370b328043.png">

## How to Validate
(no need, Shane's got this :muscle:)

leikkisa:
I think that this is currently only handling rating request issues.

shanear:
![screen shot 2018-11-19 at 7 19 47 pm](https://user-images.githubusercontent.com/1596259/48743220-1897ff80-ec30-11e8-93a9-a021dac373f4.png)
This is still showing up even when the feature flag is turned off 🙁

leikkisa:
[...]

PR #7786, opened 2018-11-09
add in legacy issues

jiahuang:
Connects #7334

The add issues button changes to "Next" when there is an active or eligible legacy appeal
image


app/models/decision_review.rb
```diff
@@ -95,8 +98,23 @@ def mark_rating_request_issues_to_reassociate!
     request_issues.select(&:rating?).each { |ri| ri.update!(rating_issue_associated_at: nil) }
   end
 
+  def legacy_issues
```
shanear: we may want to call this `serialized_legacy_issues`, since it's not returning instances of  `Issue`

app/models/decision_review.rb
```diff
@@ -95,8 +98,23 @@ def mark_rating_request_issues_to_reassociate!
...
+        issues: legacy_appeal.issues.map(&:attributes)
+      }
+    end
+  end
+
   private
 
+  def active_or_eligible_legacy_appeals
+    @active_or_eligible_legacy_appeals ||= LegacyAppeal.fetch_appeals_by_file_number(veteran_file_number).select do |appeal|
+      appeal.active? || (appeal.soc_date > APPEAL_ISSUE_SOC_ELIGIBLE || appeal.nod_date > APPEAL_ISSUE_NOD_ELIGIBLE)
```
shanear: we can add a method directly to `LegacyAppeal` called `eligible_for_soc_opt_in?` to capture this logic?

Issue #7779, opened 2018-11-08
Allow member of same team to re-assign an already assigned task


lowellrex:
#7210 allowed folks to reassign tasks that were assigned to themselves to other members of their team. This ticket exists to extend that functionality to other members of the same team. That is, if Amy Translation goes on vacation then Billie Translation should be able to reassign a task that is assigned to Amy.

## Acceptance criteria
* When a task is assigned to a member of a team and the parent task belongs to that team, we should present an option to "Re-assign to team member" for the person the task is assigned to as well as all other members of the team

PR #7799, opened 2018-11-09
Adding reassign

mdbenjam:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/7779

### Description
This PR adds a new task action when a task's parent is assigned to an organization you belong to. This action allows you to reassign the task.

This PR changes the return of the tasks's update controller to return all relevant tasks. This way the frontend can just update all the tasks, instead of trying to cherrypick the right updates since which tasks update on an action can be complex. And since other user's actions can change which tasks are still available.

@lowellrex Do we want existing instructions to appear in the reassign box? They currently don't but I think they should.

### Testing Plan
1. Log in as translation user 1
1. Assign a task to translation user 2
1. Now reassign task to translation user 3


app/models/tasks/generic_task.rb
```diff
@@ -6,22 +6,24 @@ def available_actions(user)
...
+      [
         Constants.TASK_ACTIONS.ASSIGN_TO_TEAM.to_h,
         Constants.TASK_ACTIONS.REASSIGN_TO_PERSON.to_h,
         Constants.TASK_ACTIONS.MARK_COMPLETE.to_h
       ]
-    end
-
-    if assigned_to.is_a?(Organization) && assigned_to.user_has_access?(user)
-      return [
+    elsif parent && parent.assigned_to.is_a?(Organization) && parent.assigned_to.user_has_access?(user)
```
mdbenjam: If the parent task is within your organization, then you can reassign it.

app/controllers/tasks_controller.rb
```diff
@@ -82,7 +82,9 @@ def update
     tasks = task.update_from_params(update_params, current_user)
     tasks.each { |t| return invalid_record_error(t) unless t.valid? }
 
-    render json: { tasks: json_tasks(tasks) }
+    tasks_to_return = queue_class.new(user: current_user).tasks
+
+    render json: { tasks: json_tasks(tasks_to_return) }
```
mdbenjam: @lowellrex what do you think of this? Not sure if this is going to mess something else up.

app/models/tasks/generic_task.rb
```diff
@@ -1,30 +1,37 @@
 class GenericTask < Task
+  # rubocop:disable Metrics/MethodLength
   # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/PerceivedComplexity
```
mdbenjam: 🤦‍♂️ Maybe we split this into several methods... idk I kind of think it's clearer in one.

client/app/queue/AssignToView.jsx
```diff
@@ -105,8 +104,11 @@ class AssignToView extends React.Component<Props, ViewState> {
     }
 
     return this.props.requestSave('/tasks', payload, successMsg).
-      then(() => {
-        this.props.setTaskAttrs(task.uniqueId, { status: 'on_hold' });
+      then((resp) => {
+        const response = JSON.parse(resp.text);
+        const preparedTasks = prepareTasksForStore(response.tasks.data);
+
+        this.props.onReceiveAmaTasks(preparedTasks);
```
mdbenjam: This new action just replaces the tasks with what we receive from the server.

client/app/queue/components/AssignHearingModal.jsx
```diff
@@ -178,7 +178,7 @@ class AssignHearingModal extends React.PureComponent<Props, LocalState> {
         const preparedTasks = prepareTasksForStore(response.tasks.data);
 
         // Review with team to see why this is failing.
-        this.props.setTaskAttrs(task.uniqueId, preparedTasks[task.uniqueId]);
+        this.props.onReceiveAmaTasks(preparedTasks);
```
mdbenjam: Just trying to replace this pattern everywhere.

app/models/tasks/generic_task.rb
```diff
@@ -53,9 +58,7 @@ def reassign(reassign_params, current_user)
   end
 
   def can_be_accessed_by_user?(user)
-    return true if assigned_to && assigned_to == user
-    return true if user && assigned_to.is_a?(Organization) && assigned_to.user_has_access?(user)
-    false
+    available_actions(user) != []
```
mdbenjam: What do you think of this?

app/controllers/tasks_controller.rb
```diff
@@ -82,7 +82,9 @@ def update
     tasks = task.update_from_params(update_params, current_user)
     tasks.each { |t| return invalid_record_error(t) unless t.valid? }
```
lowellrex: Somewhat unrelated thought here, but can we just have any validations throw serializable invalid record errors? That would make it so we can get rid of this validation check and just use standard ruby exception handling.

app/controllers/tasks_controller.rb
lowellrex: Also, I think `AttorneyQueue` and `JudgeQueue` both reach into VACOLS, are we concerned this operation will take much longer after this change? (I haven't gone deep down this path, and we may already be incurring these VACOLS costs the way it is currently implemented, so maybe not an issue at all).

app/models/tasks/generic_task.rb
```diff
@@ -90,6 +93,17 @@ def create_child_task(parent, current_user, params)
 
     private
 
+    def task_is_assigned_to_user_within_organiztaion?(user)
+      parent &&
+        parent.assigned_to.is_a?(Organization) &&
+        parent.assigned_to.user_has_access?(user) &&
+        assigned_to.is_a?(User)
```
lowellrex: I think we can potentially avoid unnecessary calls to BGS for VSOs if we move `assigned_to.is_a?(User)` above `parent.assigned_to.user_has_access?(user)`.

app/models/tasks/generic_task.rb
```diff
@@ -90,6 +93,17 @@ def create_child_task(parent, current_user, params)
...
+    end
+
+    def task_is_assigned_to_users_organization?(user)
+      assigned_to.is_a?(Organization) && assigned_to.user_has_access?(user)
+    end
```
lowellrex: Love these two helper functions! So explicit about what they are checking!

app/models/tasks/generic_task.rb
```diff
@@ -1,30 +1,35 @@
...
+        Constants.TASK_ACTIONS.REASSIGN_TO_PERSON.to_h
+      ]
+    elsif task_is_assigned_to_users_organization?
+      [
         Constants.TASK_ACTIONS.ASSIGN_TO_TEAM.to_h,
         Constants.TASK_ACTIONS.ASSIGN_TO_PERSON.to_h,
         Constants.TASK_ACTIONS.MARK_COMPLETE.to_h
       ]
+    else
+      []
```
lowellrex: Is our coding standard in this situation to use else-if blocks instead of returning early? If so can we encode this in rubocop?

Have a slight personal preference towards separate conditions, but that is mainly driven by my distaste for implicit returns (but that is a battle I lost with Matz, so I don't think it matters much here).

lowellrex: Also, can we move the first case inside this cascade of if statements (which I guess could also be a switch statement where we check for validity and have each case be an anonymous function?).

app/models/tasks/generic_task.rb
```diff
@@ -1,30 +1,35 @@
...
       ]
-    end
-
-    if assigned_to.is_a?(Organization) && assigned_to.user_has_access?(user)
-      return [
+    elsif task_is_assigned_to_user_within_organiztaion?
+      [
+        Constants.TASK_ACTIONS.REASSIGN_TO_PERSON.to_h
+      ]
+    elsif task_is_assigned_to_users_organization?
```
lowellrex: I think both this function and the one above it expect a `user` argument.

mdbenjam: Part of a last minute refactor that I didn't test 😬

app/models/tasks/generic_task.rb
mdbenjam: I'll move it back to individual conditions. I don't have a strong preference.

app/models/tasks/generic_task.rb
mdbenjam: Interesting. I hadn't thought of that. Hopefully this doesn't cause non-vso task endpoints to become slow. I'm going to add something to the vso model to prevent that.

app/controllers/tasks_controller.rb
mdbenjam: Maybe, not sure if you can specify what rails validations throw?

app/controllers/tasks_controller.rb
mdbenjam: I don't think `AttorneyQueue` and `JudgeQueue` request anything from VACOLS. I think that's all in `LegacyWorkQueue`. Your first point is interesting. I lean towards just implementing this for other queues. If we just do the intersection we might miss tasks that were previously completed and not in this transaction.

app/models/tasks/generic_task.rb
lowellrex: I like this solution a lot! Drew up a little truth table to confirm that this works and noticed an edge case:

| Scenario | `can_be_accessed_by_user?` |  `available_actions().any?` |
| --- | --- | --- |
| task assigned to current user | ✅ | ✅ |
| task assigned to organization, current user is a member of that organization | ✅ | ✅ |
| task assigned to VSO, current user is a member of that VSO | ✅ (by virtue of this being a subset of the above scenario) | ✅ |
| task assigned to organization, current user is a member of that organization | ❇️ (new behaviour) | ❇️|
| task is a `ScheduleHearingTask` task and current user is not task assignee and does not belong to organization schedule hearings organization | ❌ | ✅ |

Can we overwrite `can_be_accessed_by_user?` in `ScheduleHearingTask` or otherwise fix that class so this approach covers that scenario?

mdbenjam: Ah nice catch. I'm thinking we should just have a better check in available actions on ScheduleHearingTasks. Since we also don't folks to be able to see that possible action but not actually be able to take it?

app/controllers/tasks_controller.rb
lowellrex: I'm suggesting we replace rails validations with methods that raise exceptions instead of active record errors. This is probably a discussion for a later day, and almost certainly just inspired by my dislike for Rails' introduction of a second way to handle exceptions. 

Right now every action we take on records we persist to the database have to account for two types of failures (application and database) and those different types have their own control flows. Considers the following statement:
```ruby
  ...
begin
  task = GenericTask.create(id: existing_task_id, assigned_to: valid_user, appeal: valid_appeal)
rescue ActiveRecord::RecordNotUnique => e
  # Handle cases where we failed to create a row because of a database error.
end
  if !task.valid?
    # Handle cases where we failed application-level validations (say we forgot
    # to include an assignee).
  end
  ...
``` 

We should do our best to capture all possible database, but we will inevitably fail so we will need to account for database errors and will be stuck with this duplicated error handling. Unless we make all validations enter the ruby-standard exception flow by throwing exceptions instead of attaching errors to ActiveRecord objects.

app/controllers/tasks_controller.rb
lowellrex: Very much onboard with adding `recently_completed_tasks` to all queues. 👍

app/models/task.rb
```diff
@@ -61,6 +61,14 @@ def self.recently_completed
     where(status: Constants.TASK_STATUSES.completed, completed_at: (Time.zone.now - 2.weeks)..Time.zone.now)
   end
 
+  def self.incomplete
+    where.not(status: Constants.TASK_STATUSES.completed)
+  end
+
+  def self.incomplete_or_recently_completed
+    incomplete.or(recently_completed)
+  end
```
lowellrex: Love this approach!

app/models/tasks/generic_task.rb
lowellrex: Totally agree.

app/controllers/tasks_controller.rb
mdbenjam: I think we should just be using `create!` instead of `create` as I think that will throw an error if the validation fails.

client/app/queue/AssignToView.jsx
lowellrex: It looks like everywhere we call `onReceiveAmaTasks()` we receive "unprepared" tasks and prepare them before passing the result of the `prepareTasksForStore()` function to `onReceiveAmaTasks()`. Do we ever expect to receive "pre-prepared" tasks before we call `onReceiveAmaTasks()`? If not, can we just move the preparation of the tasks inside `onReceiveAmaTasks()`?

client/app/queue/components/AssignHearingModal.jsx
lowellrex: I think we need to update [`initialAssignTasksToUser()`](https://github.com/department-of-veterans-affairs/caseflow/blob/723bd2f4b1c80d59f520c81ff68666a81f51be9a/client/app/queue/QueueActions.js#L348) since I'm not sure it will work properly if we return all actionable tasks for the current user (we pluck off the first task which is not guaranteed to be the one we just acted on). Also, when we update that function I think we can get rid of `setTaskAttrs()` since that appears to be the last place we use it.

app/models/tasks/generic_task.rb
lowellrex: I think this reads much easier as a single method. I think ignoring this rubocop rule makes a lot of sense here.

lowellrex:
> Do we want existing instructions to appear in the reassign box? They currently don't but I think they should.

Are you referring to the instructions box on the frontend (I think we properly [copy existing instructions on the backend](https://github.com/department-of-veterans-affairs/caseflow/blob/723bd2f4b1c80d59f520c81ff68666a81f51be9a/app/models/tasks/generic_task.rb#L38))? I don't have a strong preference either way, but if we display them on the frontend we might have to amend the backend code so we don't duplicate the existing instructions (maybe it is time to move task instructions to a different table--task_events?!--so it would be easier to copy existing instructions and add new instructions in the same operation).

mdbenjam:
Gotcha it definitely is copied correctly on the backend. I do think solving this general diary problem would be good.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -73,9 +73,14 @@ def location_based_on_hearing_type(hearing_type)
     end
   end
 
-  def available_actions(_user)
-    [
-      Constants.TASK_ACTIONS.SCHEDULE_VETERAN.to_h
-    ]
+  def available_actions(user)
+    if (assigned_to && assigned_to == user) ||
```
mdbenjam: Updated hearing schedule's available actions here: @oscarAtNava. We also change the permission check from the `can_be_accessed_by_user?` method to use the existing of `available_actions`.

client/app/queue/QueueActions.js
```diff
@@ -400,14 +402,22 @@ export const reassignTasksToUser = ({
   return ApiUtil.patch(url, params).
     then((resp) => resp.body).
     then((resp) => {
-      const task = resp.tasks ? resp.tasks.data[0] : resp.task.data;
+      if (oldTask.appealType === 'Appeal') {
```
mdbenjam: Need to test this manually... but I think this works.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -73,9 +73,14 @@ def location_based_on_hearing_type(hearing_type)
...
+       (assigned_to.is_a?(Organization) && assigned_to.user_has_access?(user))
```
lowellrex: I think we can use the newly created `task_is_assigned_to_users_organization?` method here.

app/models/tasks/generic_task.rb
lowellrex: Does this function work the same if we call `available_actions.any?`? I find that a little more direct.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -73,9 +73,14 @@ def location_based_on_hearing_type(hearing_type)
...
+      return [
+        Constants.TASK_ACTIONS.SCHEDULE_VETERAN.to_h
+      ]
+    end
+
+    []
```
lowellrex: Do we want to support reassigning for `ScheduleHearingTask`s as well? If so, I think we can just change this fallthrough to call `super` and the only change will be that we will return `[Constants.TASK_ACTIONS.REASSIGN_TO_PERSON.to_h]` if `task_is_assigned_to_user_within_organiztaion?(user)`.

mdbenjam: For now, I'm going to hold off on this. I think they're going to do everything at the organizational level.

app/models/tasks/generic_task.rb
mdbenjam: I changed it to `.empty?`

client/app/queue/QueueActions.js
mdbenjam: It works now. Had to add a few things to make sure we pull in relevant attorney tasks

app/models/judge.rb
```diff
@@ -34,7 +34,11 @@ def upcoming_hearings_on(date, is_fetching_issues = false)
   def attorneys
     return [] unless user
     (Constants::AttorneyJudgeTeams::JUDGES[Rails.current_env][user.css_id].try(:[], :attorneys) || []).map do |css_id|
-      User.find_or_create_by(css_id: css_id, station_id: User::BOARD_STATION_ID)
+      begin
+        User.find_or_create_by(css_id: css_id, station_id: User::BOARD_STATION_ID)
+      rescue ActiveRecord::RecordNotUnique
```
mdbenjam: A test was failing because of a race condition here. It's possible to get a unique key violation when multiple requests hit this at the same time.

Issue #7595, opened 2018-10-29
AMA Intake | Payee code is required when claimant not veteran


shanear:
When the Payee code is needed, it should also be required, because we'll need it to create the EP properly.

## Acceptance Criteria
- When intaking a compensation or pension claim review, if the claimant is not the veteran, the payee code should be required.

## Notes
Related to https://github.com/department-of-veterans-affairs/caseflow/issues/7524

leikkisa:
Adding inline validation for this field.

leikkisa:
@shanear is the payee_code still technically required when the claimant is the Veteran (and defaulted to "00")?

shanear:
![screen shot 2018-11-19 at 7 50 24 pm](https://user-images.githubusercontent.com/1596259/48744233-63b41180-ec34-11e8-8a97-afeeab9d52e8.png)

It seems to not throw an error when no claimant was selected. I'm thinking we should probably require both the claimant and the payee code in this case.

leikkisa:
Ah, on the veteran part it looks like it's halfway, doesn't require it for the payee_code error, but will add one in the benefit_type (sorry we restructured and I missed that).

Claimant part makes sense!

steve-gov:
Validated in all three AMA forms by selecting one or the other fields, or neither:

<img width="927" alt="screen shot 2018-12-05 at 4 39 18 pm" src="https://user-images.githubusercontent.com/25468689/49553528-f13a6700-f8ac-11e8-8d08-5ec6c2cc318e.png">

<img width="933" alt="screen shot 2018-12-05 at 4 40 32 pm" src="https://user-images.githubusercontent.com/25468689/49553529-f13a6700-f8ac-11e8-8cb2-a269d7593724.png">

<img width="892" alt="screen shot 2018-12-05 at 4 42 20 pm" src="https://user-images.githubusercontent.com/25468689/49553530-f13a6700-f8ac-11e8-828a-a46283356434.png">

PR #7816, opened 2018-11-13
Make Payee code required for HLR/SC compensation and pension

leikkisa:
connects #7595


leikkisa:
I tried various ways of validating this in the claim_review or claimant, but it seems like since the benefit_type is part of the same request and it needs to calculate whether the claimant is not the veteran, it is flaky as to whether it validates correctly. So I moved validation to right where we're getting the params in the first place.  Open to other suggestions.

app/models/claim_review_intake.rb
```diff
@@ -43,6 +45,14 @@ def need_payee_code?(request_params)
     request_params[:benefit_type] == "compensation" || request_params[:benefit_type] == "pension"
   end
 
+  def validate_payee_code(request_params)
```
pkarman: it seems backwards to me to validate the incoming params *after* we call `detail.update`

It also seems like the validation ought to be on the `detail` model class, and it ought to use the ActiveRecord validations.

What am I missing?

leikkisa: I did it after detail update so it would have the other validation errors also if there were any.

Yeah I wanted it to be on the detail model class also. Maybe we can pair on it? I tried it a handful of different ways and it seemed to have flakey results, mostly not working.  I think this is because some things weren't calculated in time.

It seemed to work pretty well before I needed to add the check if the claimant was not the veteran.

app/models/claim_review_intake.rb
```diff
@@ -40,9 +42,21 @@ def complete!(request_params)
...
+    # payee_code is only required for claim reviews where the veteran is
+    # not the claimant and the benefit_type is compensation or pension
+    return if !request_params[:claimant] || request_params[:claimant] == detail.veteran.participant_id
     request_params[:benefit_type] == "compensation" || request_params[:benefit_type] == "pension"
   end
 
+  def validate_payee_code(request_params)
+    if need_payee_code?(request_params) && !request_params[:payee_code]
+      detail.errors.add(:payee_code, "blank")
+      detail.remove_claimants!
```
jiahuang: i dont think you need this since the next create_claimants! call will call remove_claimants!.

app/models/validators/claimant_validator.rb
```diff
@@ -0,0 +1,26 @@
+class ClaimantValidator < ActiveModel::Validator
```
jiahuang: neat! 👍

PR #7829, opened 2018-11-13
Throw exceptions if ActiveRecord validations fail

lowellrex:
Enables [the `Rails/SaveBang` rubocop rule](https://rubocop.readthedocs.io/en/latest/cops_rails/#railssavebang) which ensures that we throw exceptions if we have validation failures when modifying an ActiveRecord object.

Right now every time we modify an ActiveRecord object we have to account for two types of failures (application and database) and those different types have their own control flows. Consider the following code:
```ruby
  ...
begin
  task = GenericTask.create(id: existing_task_id, assigned_to: valid_user, appeal: valid_appeal)
rescue ActiveRecord::RecordNotUnique => e
  # Handle cases where we failed to create a row because of a database error.
end
  if !task.valid?
    # Handle cases where we failed application-level validations (say we forgot
    # to include an assignee).
  end
  ...
``` 

We should do our best to prevent all possible database errors, but we will inevitably fail so we will need to account for database errors that occur as a result of modification operations and will be stuck with this duplicated error handling. We can avoid this duplicated error handling by making all modification operations raise exceptions instead of using ActiveRecord's special error handling.

The [only change made here](https://github.com/department-of-veterans-affairs/caseflow/pull/7829/files#diff-11a0d7906801d9dea0eccb85667ad811) was adding the `Rails/SaveBang` rubocop rule and then running `rubocop -a` to automatically update existing code to conform to this rule. Some tests will probably fail, so we still need to replace calls to `active_record_object.valid?` with explicit error handling or by removing the bang (!) and calling `active_record_object.persisted?`.


mdbenjam:
I think I'm in favor of this. There is a little risk here that we break something because of unhandled exceptions. But as long as we monitor sentry after merging this, we should be able to clean any of those up. And we'll be better off because of it.

app/controllers/hearings_controller.rb
```diff
@@ -9,7 +9,7 @@ def update
       HearingRepository.slot_new_hearing(params["hearing"]["master_record_updated"], hearing.appeal)
     end
 
-    hearing.update(update_params)
+    hearing.update!(update_params)
```
sharonwarner: I'm not sure if other models are set up like this, but this would break hearings functionality. We have an `update` function in hearing.rb that calls a vacols update method and a caseflow update method. This would only call the caseflow update.

sharonwarner: Sounds good! Happy to help with the hearings updates.

mdbenjam: Yeah in general I don't think we should be overriding ActiveRecord methods. I like the idea of either a `before_update`, or better yet, just rename this method to `update_caseflow_and_vacols` or something. It's more explicit. Otherwise folks will just assume it's a normal update.

lowellrex:
Yup! Making a ticket for it!

Issue #7731, opened 2018-11-05
Assign Hearings | Optimizations


sharonwarner:
Yikes. We need to have fewer BGS calls, fewer VACOLS calls, and faster load time.

PR #7830, opened 2018-11-13
Case Hearings Vacols SQL Optimizations

lomaxap:
Resolves (partially) #7731

### Description
- Adds appellant fields to initial SQL query for Case Hearings, limiting use of vacols_attr_accessor


app/models/hearing.rb
```diff
@@ -125,18 +128,16 @@ def vacols_attributes
...
     :number_of_documents, \
     :number_of_documents_after_certification, \
     :veteran,  \
     :sanitized_vbms_id, \
-    :docket_number, \
-    :appellant_address_line_1, \
-    :appellant_city, \
-    :appellant_state, \
-    :appellant_zip, \
+    # :docket_number, \
```
sharonwarner: Do you mind just deleting these lines instead of commenting them out?

PR #7831, opened 2018-11-13
Daily docket fixes

### LINKED ISSUE #7667: Upcoming Hearing Days | Available slots logic

sharonwarner:
Resolves #7667 

### Description
This PR filters the daily docket such that canceled and postponed hearings are placed in the 'Previously Scheduled' section. It also updates the method name for updating hearings to not override the Active Record functionality.

### DIFF
--- a/app/controllers/hearings_controller.rb
+++ b/app/controllers/hearings_controller.rb
@@ -9,7 +9,9 @@ class HearingsController < ApplicationController
       HearingRepository.slot_new_hearing(params["hearing"]["master_record_updated"], hearing.appeal)
     end
 
-    hearing.update(update_params)
+    hearing.update_caseflow_and_vacols(update_params)
+    # Because of how we map the hearing time, we need to refresh the VACOLS data after saving
+    HearingRepository.load_vacols_data(hearing)
     render json: hearing.to_hash(current_user.id)
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/hearings_controller.rb
```diff
@@ -9,7 +9,9 @@ def update
       HearingRepository.slot_new_hearing(params["hearing"]["master_record_updated"], hearing.appeal)
     end
 
-    hearing.update(update_params)
+    hearing.update_caseflow_and_vacols(update_params)
+    # Because of how we map the hearing time, we need to refresh the VACOLS data after saving
+    HearingRepository.load_vacols_data(hearing)
```
sharonwarner: I don't particularly like this solution, but I think it's fine to get us through March. Previously, after updating a hearing time from the daily docket, it was only displaying properly after refreshing the page.

app/models/hearing.rb
```diff
@@ -66,10 +66,10 @@ def active_appeal_streams
     appeals << self.class.repository.appeals_ready_for_hearing(appeal.vbms_id)
   end
 
-  def update(hearing_hash)
+  def update_caseflow_and_vacols(hearing_hash)
```
sharonwarner: Updating this method name to not override Active Record functionality.

PR #7832, opened 2018-11-13
Remove reference to shared Constants in BGS Setup

leikkisa:
Using Constants::Dates was causing an "uninitialized constant" bug in some folks' dev setup, so removing it from BGS Service where the fake ratings are set up.

### DIFF
--- a/lib/fakes/bgs_service.rb
+++ b/lib/fakes/bgs_service.rb
@@ -68,12 +68,23 @@ class Fakes::BGSService
         )
         Generators::Rating.build(
           participant_id: veteran.participant_id,
-          profile_date: DecisionReview.ama_activation_date - 10.days,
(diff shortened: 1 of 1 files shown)


lib/fakes/bgs_service.rb
```diff
@@ -68,15 +68,26 @@ def self.create_veteran_records
         )
         Generators::Rating.build(
           participant_id: veteran.participant_id,
-          profile_date: DecisionReview.ama_activation_date - 10.days,
-          promulgation_date: DecisionReview.ama_activation_date - 5.days,
+          profile_date: Date.new(2019, 2, 14) - 10.days,
```
kencheeto: I've also seen Time.zone.local(2019, 2, 14) used elsewhere, rubocop might complain about this

Issue #7631, opened 2018-10-31
Prevent creating duplicate Tasks on backend


lowellrex:
Some tasks have a uniqueness constraint that is not represented in the database and, as a result, we can create multiple tasks when really we should only create a single task. This ticket exists to prevent the creation of duplicate tasks on the backend (either in controllers or models, whichever makes most sense).

An example of a "somewhat unique" task is the BvaDispatchTask (part of a solution introduced in #7603) where only one BvaDispatchTask should be created for each appeal (maybe this isn't true for remanded appeals if they share the same appeal ID).

Part of the work on this ticket includes collecting the list of tasks that are "somewhat unique" and describing in what ways they are unique. 
* BvaDispatchTask (only one org task per appeal)
* QualityReviewTask (only one org task per appeal)
* JudgeTask (only one active task per appeal)
* AttorneyTask (only one active task per appeal)

mdbenjam:
I very much agree with this. In fact some of this can be built into a validate method of GenericTask. In general we can have a validation that says we cannot create a new child task if this current task is on hold or complete. There would still be a narrow window for a race condition but it feels pretty small. Then in some cases we can override this i.e. when we want multiple admin actions.

PR #7833, opened 2018-11-13
Prevent the creation of duplicate generic tasks at the organization level

lowellrex:
Connects #7631. This PR prevents the creation of multiple active generic tasks assigned to the same organization. This solves [recent problems we have had](https://dsva.slack.com/archives/CD5DAQNCU/p1542115396005000) where multiple `BvaDispatchTasks` have been created for the same appeal.

This _does not_ prevent duplicate JudgeTasks or AttorneyTasks from being created, those changes will have to be a part of a different PR.


app/models/tasks/bva_dispatch_task.rb
```diff
@@ -3,7 +3,12 @@ class BvaDispatchTask < GenericTask
 
   class << self
     def create_and_assign(root_task)
-      parent = create!(assigned_to: BvaDispatch.singleton, parent_id: root_task.id, appeal: root_task.appeal)
+      parent = create!(
+        assigned_to: BvaDispatch.singleton,
+        parent_id: root_task.id,
+        appeal: root_task.appeal,
+        status: Constants.TASK_STATUSES.on_hold
```
lowellrex: Making the parent task status `on_hold` here so we can't create any more children tasks for this organization task.

Issue #7795, opened 2018-11-09
Intake | Remove the stats page


mdbenjam:
We have a flakey test with the stats page. Noone uses the stats page, we instead use looker, so we can get rid of it.

```
Failure/Error:
  RampClosedAppeal.create!(
    ramp_election_id: 5,
    vacols_id: "12345",
    nod_date: 365.days.ago
  )

ActiveRecord::InvalidForeignKey:
  PG::ForeignKeyViolation: ERROR:  insert or update on table "ramp_closed_appeals" violates foreign key constraint "fk_rails_d2dc5ed04b"
  DETAIL:  Key (ramp_election_id)=(5) is not present in table "ramp_elections".
  : INSERT INTO "ramp_closed_appeals" ("vacols_id", "ramp_election_id", "nod_date") VALUES ($1, $2, $3) RETURNING "id"
./spec/feature/intake/intake_stats_spec.rb:44:in `block (2 levels) in <top (required)>'
------------------
--- Caused by: ---
PG::ForeignKeyViolation:
  ERROR:  insert or update on table "ramp_closed_appeals" violates foreign key constraint "fk_rails_d2dc5ed04b"
  DETAIL:  Key (ramp_election_id)=(5) is not present in table "ramp_elections".
  ./spec/feature/intake/intake_stats_spec.rb:44:in `block (2 levels) in <top (required)>'
```

https://circleci.com/gh/department-of-veterans-affairs/caseflow/21201#tests/containers/0

## Acceptance Criteria
- Remove the Intake stats page and all the Intake stats tests

shanear:
No need to validate this 👍

PR #7835, opened 2018-11-13
Remove Intake stats page

leikkisa:
connects #7795


app/jobs/calculate_intake_stats_job.rb
```diff
@@ -1,10 +0,0 @@
-class CalculateIntakeStatsJob < ApplicationJob
```
kencheeto: I was wondering if this job gets run anywhere automatically, but I don't see any mention of it in appeals-deployment 👍

PR #7837, opened 2018-11-13
Move BvaDispatch membership to OrganizationsUser model

lowellrex:
Connects #7506. Moves BvaDispatch membership to OrganizationsUser model (similar to #7585).

Added current members of the pilot to the organization by running the following in prod:
```ruby
%w[BVAEJAMES BVARDAVIS].map { |css_id| User.find_by(css_id: css_id) }.each { |u| OrganizationsUser.add_user_to_organization(u, BvaDispatch.singleton) }
# Confirm that it worked:
BvaDispatch.singleton.users
```


app/controllers/idt/api/v1/appeals_controller.rb
```diff
@@ -88,7 +88,7 @@ def json_appeal_details
     ActiveModelSerializers::SerializableResource.new(
       appeal,
       serializer: ::Idt::V1::AppealDetailsSerializer,
-      include_addresses: Constants::BvaDispatchTeams::USERS[Rails.current_env].include?(user.css_id),
+      include_addresses: BvaDispatch.singleton.user_has_access?(user),
```
aroltsch: We are using `dispatch_user_in_vacols?` in a few places. I assume we are planning on moving away from that at some point but not now?

lowellrex: [Mentioned in slack](https://dsva.slack.com/archives/C6E41RE92/p1542145677061800?thread_ts=1542143926.058700&cid=C6E41RE92) that everybody who uses the IDT makes a request to the IDT API so we still need to use the VACOLS properties to determine whether folks can make a request to the IDT API instead of membership in the `BvaDispatch` organization.

Issue #7810, opened 2018-11-13
Intake | Wording changes


shanear:
Cleveland from AMO looked over some of the screens I shared for the edge cases and submitted some language suggestions (for edge case specific language _and_ other things). These are changes we decided to make.

Since we are demo'ing to AMO on Thursday, we'd like to have these done and deployed to UAT before then.

## Acceptance Criteria
The following copy changes are made:

- On the radio options under "What is the Benefit Type?"
   - Pension -> Pension & Survivor’s Benefits
   - VHA -> Veterans Health Administration
   - NCA -> National Cemetery Association
- "Did the Veteran request an informal conference?" => "Was an informal conference requested?"
- "Did the Veteran request review by the same office?" => "Was an interview by the same office requested?"
- "Did the applicant request an exemption to the date requirements?" = "Did the applicant request an extension to the date requirements?"
- In the navbar, "Add Issues" => "Add/Remove Issues"

## Mocks
![review - hlr - 110918](https://user-images.githubusercontent.com/11700545/48295335-d278ba00-e43f-11e8-8cdc-2ca41f2b46de.png)
## Notes
Background: https://github.com/department-of-veterans-affairs/appeals-design-research/issues/1023

kencheeto:
![Screen Shot 2018-11-13 at 11.55.08.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5b9702ff932b230012942254/125516c66e82c728ace21e0d46db978826878de9bce1b057e40fa0c5710b76226455f06b7191a379ee0c2e54415f46be5d4a1f06a3a3dc78e2a436628f5409bcd33b5a7dc43411fabc824eb801ba72894a85b5ee9850d2aa6c7bb3d4d216744b470a83ed5c3daa984ecc5af748e84d54a61b3062f43b.png)

fyi deleted this word @shanear

kencheeto:
Curious where
> In the navbar, "Add Issues" => "Add/Remove Issues"

"navbar" is - it could either be the browser title bar:

[...]

PR #7839, opened 2018-11-13
update copy for intake based on AMO feedback

kencheeto:
Connects #7810 

### Description
Copy changes as requested by stakeholders

- On the radio options under "What is the Benefit Type?"
   - Pension -> Pension & Survivor’s Benefits
   - VHA -> Veterans Health Administration
   - NCA -> National Cemetery Association
screen shot 2018-11-13 at 13 24 52
<hr>
- "Did the Veteran request an informal conference?" => "Was an informal conference requested?"
- "Did the Veteran request review by the same office?" => "Was an interview by the same office requested?"
screen shot 2018-11-13 at 13 25 49
<hr>
- In the navbar, "Add Issues" => "Add/Remove Issues"
    > i interpreted this to be both the browser title and the breadcrumb, but could easily be wrong

screen shot 2018-11-13 at 13 22 03

demo

The breadcrumb is on the long side and tends to wrap so this might be better as just 'add issues'. We use `add_issues` as the url path so I think that would be OK.

<hr>
- "Did the applicant request an exemption to the date requirements?" => "Did the applicant request an extension to the date requirements?"
screen shot 2018-11-13 at 13 29 50

### Testing Plan
N/A


client/app/intake/components/IntakeProgressBar.jsx
```diff
@@ -20,7 +20,7 @@ class IntakeProgressBarInner extends React.PureComponent {
         path: PAGE_PATHS.REVIEW
       },
       {
-        title: '4. Finish',
+        title: '4. Add / Remove Issues',
```
jiahuang: Maybe "Modify Issues" if "Add / Remove Issues" is too long?

kencheeto: I'm going to set it to 'Add Issues' since the adding/removing starts from 0 and you need at least 1

PR #7843, opened 2018-11-13
Move path() helper method from VSO to Organization

lowellrex:
I was noodling on how we could use the Organization model to determine judge team membership and I noticed that the `path()` method only existed for VSOs and not generic Organizations. This PR changes that and creates the `user_admin_path` helper function as well.


app/models/organization.rb
```diff
@@ -21,4 +21,12 @@ def self.assignable(task)
...
     users.pluck(:id).include?(user.id)
   end
+
+  def path
+    "/organizations/#{url ? url : id}"
+  end
+
+  def user_admin_path
+    "#{path}/users"
+  end
```
lowellrex: These turn out to mostly just be convenience functions for tests, but I don't think that is a bad thing and I don't think having them around incurs much cost at all.

PR #7847, opened 2018-11-13
[WIP] Move judge teams into Organizations model

lowellrex:
Connects #7506. One idea for how we could use the Organizations concept to accommodate judge teams. Early PR, just looking for feedback on this approach.

One nice thing about this approach is that it allows us to use the `JudgeTeam.path` as the URL for the judge assign view when we get to that point.

Before moving over to this model completely we would need to create the records in the database. We could do this by running the following commands in a prod console:
```ruby
Constants::AttorneyJudgeTeams::JUDGES[Rails.current_env].each_pair do |judge_css_id, h|
  judge = User.find_by(css_id: judge_css_id)
  judge_team = JudgeTeam.for_judge(judge) || JudgeTeam.create_for_judge(judge)
  h[:attorneys].each do |css_id| 
    OrganizationsUser.add_user_to_organization(User.find_by(css_id: css_id), judge_team)
  end
end
```


app/models/organizations/judge_team.rb
```diff
@@ -0,0 +1,9 @@
+class JudgeTeam < Organization
+  def self.for_judge(user)
+    find_by(name: user.css_id)
+  end
+
+  def self.create_for_judge(user)
+    create!(name: user.css_id)
```
lowellrex: We rely on the name of the JudgeTeam organization being the CSS ID of the judge user. I think this is brittle but am struggling to think of a better way to have JudgeTeams use the Organizations framework.

mdbenjam: I think this is a reasonable way to do this. Maybe a little brittle, but I don't think that's a huge concern. An alternative solution could be to add an `admin` column to the organizations_user table which gives that user admin power over the organization.

This might be useful as a more general solution. As there are likely teams we only want certain admins editing membership and tasks within. The mass-assign UI could be shown to all admin users.

lowellrex:
Closing this out in favor of completing the work as part of #8063.

PR #7848, opened 2018-11-13
Do not send exception to Sentry when judge/attorney doesn't have access to the legacy task

aroltsch:
For this particular error, there is nothing can be done on our end so don't send it to Sentry anymore

Fixes https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2614/


app/controllers/case_reviews_controller.rb
```diff
@@ -4,6 +4,10 @@ class CaseReviewsController < ApplicationController
     JudgeCaseReview: JudgeCaseReview
   }.freeze
 
+  rescue_from Caseflow::Error::UserRepositoryError do |e|
+    render json: { "errors": ["title": e.class.to_s, "detail": e.message] }, status: 400
+  end
```
lowellrex: Can we make `UserRepositoryError` into a serializable error and move this error handling into the base `ApplicationController`? At [the top of the block](https://github.com/department-of-veterans-affairs/caseflow/blob/bb1606b5b8e3601493c3878d473d19c9d4bdf650/app/controllers/application_controller.rb#L19) in `ApplicationController` where we `rescue_from Caseflow::Error::VacolsRepositoryError`, we can add a line that renders the error's native response (something like `render(e.serialize_response) if e.class.method_defined?(:serialize_response)` perhaps) if the error has one.

This approach would require making `UserRepositoryError` serializable (which I think is pretty trivial, though would probably require making `SerializableError` into a concern), but I think that is probably for the best.

spec/controllers/case_reviews_controller_spec.rb
```diff
@@ -258,13 +257,39 @@
...
+
+          it "should not be successful" do
+            allow_any_instance_of(User).to receive(:fail_if_no_access_to_legacy_task!)
+              .and_raise(Caseflow::Error::UserRepositoryError)
+            expect(Raven).to_not receive(:capture_exception)
+            post :complete, params: { task_id: task_id, tasks: params }
+            expect(response.status).to eq 400
+          end
+        end
+
```
lowellrex: Tests look great!

Issue #7718, opened 2018-11-05
Incorporate hearing info in case search


laurjpeterson:
We think Caseflow Queue users (attorneys, judges, admin support, etc.) will gain value from seeing hearing information in case search results. The Hearing Schedule team has also designed and implemented search results that include hearing data to meet their users' needs. 

This design could also be leveraged in the queue table.

## Requirements
@sneha-pai completed an initial design exploration based on requirements listed in this [comment](https://github.com/department-of-veterans-affairs/caseflow/issues/7016#issuecomment-428678275). 

#### Case search (with appeals AND VBA reviews)
![search results-proposed](https://user-images.githubusercontent.com/31224257/48272942-d390f380-e40d-11e8-8528-50f9fe8dbf7c.png)

#### On Hover or click (if possible)
![search results copy 2](https://user-images.githubusercontent.com/31224257/48272947-d68be400-e40d-11e8-9837-b21916a130ce.png)

sneha-pai:
@mdbenjam @lowellrex 
@MeredithStewart @sharonwarner 
Please review above mocks for feasibility, concerns, feedback.

MeredithStewart:
@sneha-pai -- I think the designs are great. I'm wondering for AMA hearing docket, would we also implement the H tag with the scroll over info? I'm thinking it might be good (even if a bit duplicative). Curious what you thought.

lowellrex:
I don't think this will affect loading/performance at all. We already get all of the information from VACOLS and return it to the frontend so we aren't requesting any additional information.

As for how complex it is to implement, I don't think it will be all that tough. I imagine most of the work here will be concerned with making the visual appearance match the mocks and not finessing the data (since that is probably already in the form we need).

Currently the only thing left in the chute for case search/queue list view this sprint is #6710 (adding HLR and supplemental claims to the search results), so we can tackle this ticket after we complete that puppy (or before if we want to prioritize this ticket?).

PR #7849, opened 2018-11-13
Add "H" badge to case search view for cases with hearings

joeyyang:
Partially addresses https://github.com/department-of-veterans-affairs/caseflow/issues/7718, but figured since this chunk is tidy and adds value I'd submit it first.

### Description
Adds an "H" badge to case search view to indicate cases that have an associated hearing. Hovering over the "H" reveals a (standard) tooltip with additional information.

*Updated* screenshots: 
screenshot 2018-11-16 13 35 15
screenshot 2018-11-16 13 35 19
screenshot 2018-11-16 13 35 33


client/app/hearings/constants/constants.js
```diff
@@ -67,3 +67,9 @@ export const SERVER_ERROR_CODES = {
 
 export const VIDEO_HEARING = 'Video';
 export const CENTRAL_OFFICE_HEARING = 'Central';
+
+export const HEARING_OPTIONS = {
+  central_office: 'Central Office',
+  video: 'Video',
+  travel_board: 'Travel Board'
```
joeyyang: Took a guess here — are these labels accurate?

client/app/queue/CaseListTable.jsx
```diff
@@ -40,13 +41,25 @@ class CaseListTable extends React.PureComponent {
...
       header: COPY.CASE_LIST_TABLE_DOCKET_NUMBER_COLUMN_TITLE,
-      valueFunction: (appeal) => <React.Fragment>
-        <DocketTypeBadge name={appeal.docketName} number={appeal.docketNumber} />
-        <CaseDetailsLink
-          appeal={appeal}
-          userRole={this.props.userRole}
-          getLinkText={() => appeal.docketNumber} />
-      </React.Fragment>
+      valueFunction: (appeal) => {
+        const mostRecentHearing = appeal.hearings[0];
```
joeyyang: Wanted to confirm — if there's more than one hearing, what do we do here? Most recent? Show all?

Also: the test data I was using only had one hearing, but if there's another test user / case with more than one hearing I can use let me know.

lowellrex: I think you're absolutely right that we want to use the most recent hearing if there are multiple. [Started a slack thread](https://dsva.slack.com/archives/C42P25D6H/p1542206903042800) asking the question of how the `hearings` property of `LegacyAppeal`s are sorted.

client/app/hearings/constants/constants.js
lowellrex: It may be more future proof to call something like [lodash's startCase method](https://lodash.com/docs/4.17.10#startCase) on `appeal.type` so that we don't have to maintain this mapping (I doubt these three type will change all that often, but perhaps the keys may).

sneha-pai:
48450546-7e314b00-e75b-11e8-9d47-ae5d8f690de4
@laurjpeterson @lowellrex @joeyyang Would it be possible to shorten the first column header to be "Docket" since it's both number and type (the current length is very wide)?

If we did this, we'd have some space to indent column 1 to the right a bit, so that labels (such as Hearings label) live in a space that have no column headers associated with them.

client/app/queue/components/HearingBadge.jsx
```diff
@@ -0,0 +1,52 @@
...
+  '& > li': {
+    marginBottom: 0,
+    '& > strong': {
+      color: COLORS.WHITE
+    }
+  }
+});
+
+const DocketTypeBadge = ({ hearing, id }) => {
+  const dispositionText = _.find(DISPOSITION_OPTIONS, ['value', hearing.disposition]).label;
```
lowellrex: I think we can do the same uppercasing [I suggested above](https://github.com/department-of-veterans-affairs/caseflow/pull/7849/files#r233477713) here.

client/app/hearings/constants/constants.js
sneha-pai: Yes, I believe so, especially if this is an exhaustive list of labels for BOTH legacy and AMA appeals. We do know that no AMA appeal will have a travel board hearing option.

client/app/queue/CaseListTable.jsx
```diff
@@ -40,13 +41,25 @@ class CaseListTable extends React.PureComponent {
...
+        let hearingBadge;
+
+        if (mostRecentHearing) {
+          hearingBadge = <HearingBadge
+            hearing={mostRecentHearing}
+            id={mostRecentHearing.id} />;
+        }
+
+        return <React.Fragment>
+          {hearingBadge}
```
lowellrex: Let's move the decision about whether to show an H icon into the `HearingBadge` component itself. Perhaps the interface could look a little something like:
```jsx
<HearingBadge hearing={appeal.hearings[0]} />
```

And in the `HearingBadge` component we will return `null` early if the hearing property is undefined or null. This saves us from doing any logic here.

lowellrex:
@sneha-pai I think we could definitely put the hearing badge in its own column with no header. That definitely works.

lowellrex:
One other comment @joeyyang can we add this hearing badge to the `TaskTable` component as well?

laurjpeterson:
@joeyyang - will this also have the hover over information that @sneha-pai designed in this issue? https://github.com/department-of-veterans-affairs/caseflow/issues/7718

sneha-pai:
@laurjpeterson The second mock in this issue already reflects the on-hover tooltip mentioned in #7718  ? Were you referring to another hearing data point?

laurjpeterson:
I did not scroll up :( so sorry to throw you off!

client/app/hearings/constants/constants.js
joeyyang: Oh great! `_.startCase` is great!

client/app/queue/components/HearingBadge.jsx
joeyyang: Great! Lodash is magic and I love it.

joeyyang:
Thanks everyone! Here's a new screenshot:
screenshot 2018-11-14 10 21 16

Currently working on getting the padding to go away by the H badge, will push that shortly.

joeyyang:
Updated the view here to only show the column for the H badge if any results have hearings.

screenshot 2018-11-14 13 02 14
screenshot 2018-11-14 13 02 24

client/app/queue/CaseListTable.jsx
lowellrex: No response to yesterday's thread, so [I've re-asked the question here](https://dsva.slack.com/archives/C42P25D6H/p1542288642049700).

client/app/queue/CaseListTable.jsx
```diff
@@ -18,6 +20,12 @@ const currentAssigneeStyling = css({
   color: COLORS.GREEN
 });
 
+const badgeStyling = css({
+  '& > tbody td:first-child ': {
+    paddingRight: 0
+  }
+});
```
lowellrex: Let's put this styling on the `HearingBadge` itself so we don't have to re-style this if we put it in other tables.

I think we can achieve this by wrapping the `HearingBadge` tooltip in a div and setting the margin on that div:
```jsx
return <div {...css({xmarginRight: '-1.5rem'})}>
  <Tooltip ...
```

One downside is that we are matching the 1.5rem value of the existing padding (from the USWDS styles) to reduce the size of that gutter, but I don't think that tight coupling is terrible thing because we don't expect those styles to change (they haven't in at least the past 14 months) and hopefully when they do we will do an audit of the visual appearance of the site and notice this change.

client/app/queue/CaseListTable.jsx
```diff
@@ -37,47 +45,71 @@ class CaseListTable extends React.PureComponent {
...
+        valueFunction: (appeal) => labelForLocation(appeal.locationCode, this.props.userCssId)
+      }
+    ];
+
+    const doAnyAppealsHaveHearings = Boolean(_.find(this.props.appeals, (appeal) => {
+      return appeal.hearings.length;
+    }));
+
+    if (!doAnyAppealsHaveHearings) {
+      columns.shift();
```
lowellrex: Right now this operation relies on the hearings column being the first column. Why don't we do the inverse here so we aren't coupled to the ordering of the columns? That is:
```jsx
if (doAnyAppealsHaveHearings) {
  columns.unshift({ valueFunction: ...
```

client/app/queue/components/HearingBadge.jsx
```diff
@@ -0,0 +1,55 @@
...
+    '& > strong': {
+      color: COLORS.WHITE
+    }
+  }
+});
+
+const DocketTypeBadge = ({ hearing }) => {
+  let badge = null;
+
+  if (hearing) {
```
lowellrex: Personal preference here, so take it or leave it, but I generally prefer to just return null early if we don't have a hearing:
```jsx
if (!hearing) {
  return null;
}
```

client/app/queue/components/TaskTable.jsx
```diff
@@ -76,6 +78,13 @@ class TaskTable extends React.PureComponent<Props> {
 
   collapseColumnIfNoDASRecord = (task) => this.taskHasDASRecord(task) ? 1 : 0
 
+  caseHearingColumn = () => {
+    return this.props.includeHearingBadge ? {
+      header: '',
+      valueFunction: (task) => <HearingBadge hearing={task.appeal.hearings[0]} />
+    } : null;
+  }
+
```
lowellrex: Excellent! Great addition!

lowellrex:
I think we're really close here, all that is left is responding to a few comments and adding some tests (displaying/not displaying hearing badges, showing information for the correct hearing when there are multiple).

client/app/queue/CaseListTable.jsx
joeyyang: Moved the style — I was playing around with the margin and I like it at `-3rem`. It looks like the default left/right padding on a cell is `1.5rem`, and using `-3rem` compensates for the table's `td:first-child` having a `padding-left: 0` and uses the left-padding of the next column to balance it to my eye. What do you think / @sneha-pai, any thoughts? 

`-3rem`
screenshot 2018-11-15 20 11 54

`-1.5rem`
screenshot 2018-11-15 20 11 46

sneha-pai: I like the -3rem version! thanks @joeyyang!

client/app/queue/CaseListTable.jsx
```diff
@@ -37,47 +39,69 @@ class CaseListTable extends React.PureComponent {
...
+    if (doAnyAppealsHaveHearings) {
+      const hearingColumn = {
+        valueFunction: (appeal) => {
+          const hearings = appeal.hearings.sort((h1, h2) => h1.date < h2.date ? 1 : -1);
+
+          return <HearingBadge hearing={hearings[0]} />;
+        }
+      };
+
+      columns.unshift(hearingColumn);
```
lowellrex: Excellent!

client/app/queue/components/HearingBadge.jsx
```diff
@@ -0,0 +1,55 @@
...
+    This case has a hearing associated with it.
+    <ul {...listStyling}>
+      <li>Judge: <strong>{hearing.heldBy}</strong></li>
+      <li>Disposition: <strong>{_.startCase(hearing.disposition)}</strong></li>
+      <li>Date: <strong><DateString date={hearing.date} /></strong></li>
+      <li>Type: <strong>{_.startCase(hearing.type)}</strong></li>
+    </ul>
+  </div>;
+
+  return <div {...css({ marginRight: '-3rem' })} className="cf-hearing-badge">
```
lowellrex: Let's add a code comment explaining that this -3rem right margin is because we expect this badge to be shown in a table view. I suspect we will end up reusing this on the case details page and might be confused about why we've added this margin when we do that.

spec/feature/queue/search_spec.rb
```diff
@@ -1,5 +1,6 @@
 require "rails_helper"
 # rubocop:disable Style/FormatString
+# rubocop:disable Style/FormatStringToken
```
lowellrex: Do we need to disable this rule because of [the strftime call we make](https://github.com/department-of-veterans-affairs/caseflow/pull/7849/files#diff-a788200fa86c73f4fdf0613faa71e84bR80) in the new test?

spec/feature/queue/search_spec.rb
```diff
@@ -39,6 +40,62 @@
...
+
+      context "no cases in the search view have hearings" do
+        before do
+          visit "/search"
+          fill_in "searchBarEmptyList", with: appeal.sanitized_vbms_id
+          click_on "Search"
+        end
+
+        it "table does not display a column for a badge if no cases have hearings" do
+          docket_column_header = page.find(:xpath, "//thead/tr/th[1]/span")
```
lowellrex: Don't we expect the first column to be the `CASE_LIST_TABLE_DOCKET_NUMBER_COLUMN_TITLE`? Or does this work because xpath indexes from 1 instead of 0 (I really hope xpath doesn't do this)?

lowellrex: Oh my goodness the [index wackiness](https://en.wikipedia.org/wiki/XPath#Abbreviated_syntax) is true 😱:
> Note also, index values in XPath predicates (technically, 'proximity positions' of XPath node sets) start from 1, not 0 as common in languages like C and Java.

Issue #7335, opened 2018-10-10
Legacy Opt-in | Add issue | VACOLS modal


suzchap:
## Story
[Mural showing all opt-in screens](https://app.mural.co/t/workqueue2001/m/workqueue2001/1538676663322/098da798af658862105ea8d4d5aeb03fa7aaf99a)

When a user has a legacy appeal, they will be allowed to "opt-in" to move one or more of those issues into an AMA lane. The AMA paper forms will have a question on them for the applicant to acknowledge that by checking that box, they understand that any legacy issues listed on that form will be moved out of the legacy lane and into the new AMA lane.

This ticket is focused on the modal that displays VACOLS issues that need to be matched.

requires #7334 

## Acceptance criteria
- The modal reached by the flow in #7334 should show the modal in the mock below.
- A notice of disagreement header should show for each active and eligible legacy appeal (see #7334 for definitions of active and eligible) for the veteran, ordered from most recent to least recent. Any issues on those appeals should be shown below the appropriate header as a radio button.
- The last radio button in the list should be "None of these match"
- Only one item should be selectable from the radio button lists, even if there are multiple appeals.
- If "None of these match" is selected, the "Add this issue" button is enabled, and if clicked, it should be treated as a normal, eligible issue.
- "Add this issue" button is disabled until any item is selected 

## Notes
- Out of scope: if a VACOLS issue is selected from the list, what happens next will be handled by follow up stories (#7336 and #7337)

## Copy
Does issue 1 match any of these VACOLS issues?
None of these match

## Mocks
<img width="549" alt="screen shot 2018-10-10 at 2 53 48 pm" src="https://user-images.githubusercontent.com/11219417/46759044-581c1500-cc9c-11e8-9187-9c54063d213a.png">

## How to Validate
(no need, Shane's got this :muscle:)

shanear:
this and
https://github.com/department-of-veterans-affairs/caseflow/issues/7335

aren't working with prod

shanear:
![screen shot 2018-11-19 at 7 05 47 pm](https://user-images.githubusercontent.com/1596259/48742757-3cf2dc80-ec2e-11e8-856b-52d142952192.png)
There is something off with the radio buttons.

steve-gov:
Radio buttons validated working, can be selected independently. (Mac/ Chrome)

[...]

PR #7851, opened 2018-11-14
adding legacy opt in modal

jiahuang:
Connects #7335 

image


client/app/intake/components/LegacyOptInModal.jsx
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  radioOnChange = (value) => {
+    this.setState({
+      vacolsId: value
+    });
+  }
+
+  onAddIssue = () => {
+    // currently just adds the issue
```
jiahuang: I think I also need to do a timelyness check here if "None of these match" is selected and show the timelyness alert module if it's not timely...?

spec/factories/legacy_appeal.rb
```diff
@@ -2,10 +2,12 @@
   factory :legacy_appeal do
     transient do
       vacols_case nil
+      date nil
     end
 
     vacols_id { vacols_case.bfkey }
     vbms_id { vacols_case.bfcorlid }
+    nod_date { nil unless date }
```
jiahuang: maybe this should be vacols_case.nod_date instead?

spec/feature/intake/supplemental_claim_spec.rb
```diff
@@ -746,7 +746,21 @@ def check_row(label, text)
 
     context "with active legacy appeal" do
       before do
-        create(:legacy_appeal, vacols_case: create(:case, bfcorlid: "#{veteran.file_number}S"))
+        # create two legacy appeals with 2 issues each
```
jiahuang: should this be pulled out into the intake helper?

client/app/intake/components/LegacyOptInModal.jsx
leikkisa: I think that timeliness is exempted if they choose "Yes" on opting into their legacy issues.

leikkisa: I think that's correct if none of them match.

spec/feature/intake/supplemental_claim_spec.rb
leikkisa: I like that idea

app/models/decision_review.rb
```diff
@@ -103,12 +103,13 @@ def mark_rating_request_issues_to_reassociate!
   end
 
   def serialized_legacy_issues
-    active_or_eligible_legacy_appeals.map do |legacy_appeal|
-      {
-        date: legacy_appeal.nod_date,
-        issues: legacy_appeal.issues.map(&:attributes)
-      }
-    end
+    active_or_eligible_legacy_appeals.reject { |legacy_appeal| legacy_appeal.issues.empty? }
```
shanear: i feel like an appeal with no issues should also not be eligible, and this logic should live a little deeper in the code?

spec/factories/legacy_appeal.rb
shanear: We shouldn't need to do this, we should just set `vacols_case.nod_date` like you suggested

PR #7869, opened 2018-11-15
Rkreyhsig/7654 display completed tasks in case timeline

### LINKED ISSUE #7654: Display completed tasks in case timeline
Currently case timeline only displays events for the receipt of the form 9 and NOD (image below). We should return additional information about any tasks (judge, attorney, dispatch, etc.) that have been completed and show those in the timeline as well (mocks for this exist in [the mural related to case timeline](https://app.mural.co/t/workqueue2001/m/workqueue2001/1531849576358/38b753bff1e25963a76d8ebecf9679d6445c39b3)).

image

## Acceptance criteria
* Display completed dates for tasks related to the appeal

rkreyhsig:
Resolves #7654 - Display completed tasks in case timeline


app/controllers/appeals_controller.rb
```diff
@@ -51,15 +51,15 @@ def veteran
...
     respond_to do |format|
       format.html { render template: "queue/index" }
       format.json do
         id = params[:appeal_id]
         MetricsService.record("Get appeal information for ID #{id}",
                               service: :queue,
                               name: "AppealsController.show") do
-          render json: { appeal: json_appeals([appeal])[:data][0] }
+          render json: { appeal: json_appeals([appeal])[:data][0],
+                         can_edit_aod: AodTeam.singleton.user_has_access?(current_user) }
```
lowellrex: I think these are coming from an old PR and we can probably get rid of these changes for this PR.

app/models/tasks/attorney_task.rb
```diff
@@ -27,6 +27,10 @@ def available_actions(user)
     ]
   end
 
+  def friendly_name
+    "Decision drafted (Attorney)"
```
lowellrex: Maybe "Decision drafted by attorney"? Also, let's put this in [COPY.json](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/COPY.json) so it can be more easily changed on the fly.

app/models/tasks/judge_task.rb
```diff
@@ -20,6 +20,10 @@ def available_actions(user)
     end
   end
 
+  def friendly_name
+    "Decision taken (Judge)"
```
lowellrex: I think this is often referred to as a decision being "signed" so maybe something like "Decision signed by judge" would be a little clearer. Also, let's put this in COPY.json as well.

app/models/tasks/attorney_task.rb
```diff
@@ -27,6 +27,10 @@ def available_actions(user)
...
+  end
```
lowellrex: We expect all completed `AttorneyTask`s related to AMA appeals to be caused by the attorney drafting a decision and submitting it for review by their judge. We will have to handle OMO requests for legacy appeals in a different way. That work will happen in a different PR.

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -94,4 +94,8 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
       nod_receipt_date: object.receipt_date
     }
   end
+
+  attribute :timeline do
+    object.tasks.where(status: "completed").order("completed_at DESC").map(&:prepare_for_timeline)
```
lowellrex: Let's move this to the Appeal (and LegacyAppeal) class to make it a little more testable (in model tests) and since I think it will have a little more logic than it currently does when all is said and done.

client/app/queue/utils.js
```diff
@@ -265,6 +265,7 @@ export const prepareAppealForStore =
           nodReceiptDate: appeal.attributes.events.nod_receipt_date,
           form9Date: appeal.attributes.events.form9_date
         },
+        timeline: appeal.attributes.timeline,
```
lowellrex: I think we need to map over this attribute (expect timeline to be an array) so we can camelCase the snake_cased "pending_title" key of each element of the timeline.

app/models/appeal.rb
```diff
@@ -212,6 +212,22 @@ def create_tasks_on_intake_success!
...
+    [
+      {
+        title: COPY::CASE_TIMELINE_DISPATCHED_FROM_BVA,
+        pendingTitle: COPY::CASE_TIMELINE_DISPATCH_FROM_BVA_PENDING,
+        date: decision_date
+      },
+      tasks.where(status: "completed").order("completed_at DESC").map(&:timeline_title),
+      {
+        title: COPY::CASE_TIMELINE_NOD_RECEIVED,
+        pendingTitle: COPY::CASE_TIMELINE_NOD_PENDING,
```
lowellrex: Let's remove `pending_title` element from these hashes and use the presence of a decision/receipt date to determine the title. Something like the following should do the trick:
```ruby
title: receipt_date ? COPY::CASE_TIMELINE_NOD_RECEIVED : COPY::CASE_TIMELINE_NOD_PENDING,
```

We will also need to make this change [in the LegacyAppeal model](https://github.com/department-of-veterans-affairs/caseflow/pull/7869/files#diff-5a94be9eb6a667117856c5bd2d145da7R706) as well, but by including this logic on the backend we can remove some complexity from the frontend by getting rid of the `pendingTitle` argument of [the `getEventRow()` function](https://github.com/department-of-veterans-affairs/caseflow/blob/9bdacdd29d427884737a7cef7cac8056e1bc9a81/client/app/queue/CaseTimeline.jsx#L30).

app/models/tasks/judge_task.rb
```diff
@@ -20,6 +20,10 @@ def available_actions(user)
     end
   end
 
+  def friendly_name
+    COPY::CASE_TIMELINE_JUDGE_TASK
```
lowellrex: Love this! So happy that this text is in COPY.json!

client/app/queue/CaseTimeline.jsx
```diff
@@ -40,31 +40,13 @@ export default class CaseTimeline extends React.PureComponent {
...
-    ].filter((event) => !event.legacyOnly || appeal.isLegacyAppeal);
 
     return <React.Fragment>
       {COPY.CASE_TIMELINE_HEADER}
       <table>
         <tbody>
-          {events.map((event, index) => {
-            return this.getEventRow(event, index === events.length - 1);
+          {appeal.timeline.map((event, index) => {
+            return this.getEventRow(event, index === appeal.timeline.length - 1);
```
lowellrex: Since we've moved so much logic out of this component I think it makes sense to change `CaseTimeline` into a [purely functional component](https://hackernoon.com/react-stateless-functional-components-nine-wins-you-might-have-overlooked-997b0d933dbc). Function signatures could look something like the following:
```jsx
const getEventRow = ({ title, date }, lastRow) => <tr key={title}>
...

const CaseTimeline = ({ appeal }) => <React.Fragment>
...
```

client/app/queue/utils.js
```diff
@@ -265,6 +265,11 @@ export const prepareAppealForStore =
           nodReceiptDate: appeal.attributes.events.nod_receipt_date,
           form9Date: appeal.attributes.events.form9_date
         },
+        timeline: appeal.attributes.timeline.map((event) => ({
+          title: event.title,
+          pendingTitle: event.pending_title,
```
lowellrex: If we [remove the `pendingTitle` element](https://github.com/department-of-veterans-affairs/caseflow/pull/7869/files#r234967140) of the objects in the timeline array we can remove this unpacking and end up with just `timeline: appeal.attributes.timeline,` here.

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -94,4 +94,8 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
       nod_receipt_date: object.receipt_date
     }
   end
+
+  attribute :timeline do
+    object.timeline
+  end
```
lowellrex: Now that we've added these timeline properties of the serialized appeal objects we should remove the `events` properties of the [serializers](https://github.com/department-of-veterans-affairs/caseflow/blob/9bdacdd29d427884737a7cef7cac8056e1bc9a81/app/models/serializers/work_queue/appeal_serializer.rb#L92) and the [related unpacking code](https://github.com/department-of-veterans-affairs/caseflow/blob/9bdacdd29d427884737a7cef7cac8056e1bc9a81/client/app/queue/utils.js#L264) in `queue/utils.js`.

spec/feature/queue/case_details_spec.rb
```diff
@@ -448,5 +450,41 @@
...
+
+      it "should display judge & attorney tasks" do
+        visit "/queue/appeals/#{appeal.uuid}"
+        expect(page).to have_content(COPY::CASE_TIMELINE_ATTORNEY_TASK)
+        expect(page).to have_content(COPY::CASE_TIMELINE_JUDGE_TASK)
+      end
+
+      it "should NOT display judge & attorney tasks" do
+        visit "/queue/appeals/#{appeal2.uuid}"
+        expect(page).not_to have_content(COPY::CASE_TIMELINE_JUDGE_TASK)
```
lowellrex: Tests look great!

spec/feature/queue/case_details_spec.rb
```diff
@@ -448,5 +450,41 @@
...
+      end
+      let!(:qr) { QualityReview.singleton }
+      let!(:task) do
+        FactoryBot.create(:qr_task, appeal: appeal, assigned_to: user, status: "completed",
+                                    completed_at: Time.zone.now - 2.days)
+      end
+
+      before do
+        OrganizationsUser.add_user_to_organization(user, qr)
+        attorney_task.update!(status: "completed")
```
lowellrex: Let's add a code comment here that explains why we need to update the status here instead of just setting the status when we create the task (because of [this `after_update` hook](https://github.com/department-of-veterans-affairs/caseflow/blob/9bdacdd29d427884737a7cef7cac8056e1bc9a81/app/models/task.rb#L15) in the Task model). Relatedly, let's not set the status of the `attorney_task` [when we create it](https://github.com/department-of-veterans-affairs/caseflow/pull/7869/files#diff-0230c2b77318163b80f56281c4b2d168R461) because that status doesn't persist (hence this update action here) and it may confuse us down the line as to why we set that status in the first place.

spec/feature/queue/case_details_spec.rb
```diff
@@ -448,5 +450,41 @@
...
+                                 assigned_to: user, status: "completed", completed_at: Time.zone.now)
+      end
+      let!(:qr) { QualityReview.singleton }
+      let!(:task) do
+        FactoryBot.create(:qr_task, appeal: appeal, assigned_to: user, status: "completed",
+                                    completed_at: Time.zone.now - 2.days)
+      end
+
+      before do
+        OrganizationsUser.add_user_to_organization(user, qr)
```
lowellrex: Do we still need to do this in order to view the case details page? Would this test work if we removed all of this QR stuff (`OrganizationsUser...`, `let!(:task)`, `let!(:qr)...`)?

spec/feature/queue/case_details_spec.rb
```diff
@@ -448,5 +450,41 @@
...
+      let!(:appeal) { FactoryBot.create(:appeal) }
+      let!(:appeal2) { FactoryBot.create(:appeal) }
+      let!(:root_task) { create(:root_task, appeal: appeal, assigned_to: user) }
+      let!(:attorney_task) do
+        FactoryBot.create(:task, appeal: appeal, type: AttorneyTask.name, parent: root_task,
+                                 assigned_to: user, status: "completed", completed_at: Time.zone.now - 4.days)
+      end
+      let!(:judge_task) do
+        FactoryBot.create(:task, appeal: appeal, type: JudgeTask.name, parent: attorney_task,
+                                 assigned_to: user, status: "completed", completed_at: Time.zone.now)
```
lowellrex: We have a [set of constants](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/constants/TASK_STATUSES.json) that map to these status strings, so let's use `Constants.TASK_STATUSES.completed` in place of `"completed"` here. The advantages of using a constant instead of a string literal are that we only have to change this value in one place if we ever want to change it, we remove the possibility that we could forget to change this value somewhere in the code if we do decide to change it, and we will get a `NoMethodError` if we mistype the name of the constant instead of possibly succeeding (or getting some less direct error) if we mistype the string literal.

app/models/appeal.rb
```diff
@@ -212,6 +212,22 @@ def create_tasks_on_intake_success!
...
   end
 
+  def timeline
+    [
+      {
+        title: COPY::CASE_TIMELINE_DISPATCHED_FROM_BVA,
+        pendingTitle: COPY::CASE_TIMELINE_DISPATCH_FROM_BVA_PENDING,
+        date: decision_date
+      },
+      tasks.where(status: "completed").order("completed_at DESC").map(&:timeline_title),
```
lowellrex: Let's use the Constant version of the completed status here (more detail about this approach in [the same suggestion](https://github.com/department-of-veterans-affairs/caseflow/pull/7869/files#r234974615) in the test).

app/models/legacy_appeal.rb
```diff
@@ -703,6 +703,23 @@ def external_id
...
+      },
+      {
+        title: form9_date ? COPY::CASE_TIMELINE_FORM_9_RECEIVED : COPY::CASE_TIMELINE_FORM_9_PENDING,
+        date: form9_date
+      },
+      {
+        title: nod_date ? COPY::CASE_TIMELINE_NOD_RECEIVED : COPY::CASE_TIMELINE_NOD_PENDING,
+        date: nod_date
+      }
+    ].flatten
```
lowellrex: Minor note, but we don't need the `flatten` here now that we've removed the nested array of tasks that was previously contained within this array.

app/models/appeal.rb
```diff
@@ -212,6 +212,20 @@ def create_tasks_on_intake_success!
...
+      {
+        title: decision_date ? COPY::CASE_TIMELINE_DISPATCHED_FROM_BVA : COPY::CASE_TIMELINE_DISPATCH_FROM_BVA_PENDING,
+        date: decision_date
+      },
+      tasks.where(status: Constants.TASK_STATUSES.completed).order("completed_at DESC").map(&:timeline_details),
+      {
+        title: receipt_date ? COPY::CASE_TIMELINE_NOD_RECEIVED : COPY::CASE_TIMELINE_NOD_PENDING,
+        date: receipt_date
+      }
+    ].flatten
```
lowellrex: Love it! This looks great!

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -89,9 +89,7 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
     object.veteran ? object.veteran.id : nil
   end
 
-  attribute :events do
-    {
-      nod_receipt_date: object.receipt_date
-    }
+  attribute :timeline do
+    object.timeline
```
lowellrex: Not a dealbreaker, but I think we can even just do `attribute :timeline` without the block that calls `object.timeline` here (like we do [in the legacy serializer](https://github.com/department-of-veterans-affairs/caseflow/pull/7869/files#diff-cf9a90d36a0bad8f006b5b878a6161f2R5)) since that is the default behaviour for serializers.

app/models/task.rb
```diff
@@ -207,6 +207,17 @@ def assign_to_judge_data
...
+  def timeline_title
+    "#{type} completed"
+  end
+
+  def timeline_details
+    {
+      title: timeline_title,
+      date: completed_at
+    }
+  end
```
lowellrex: Excellent! I love it!

PR #7871, opened 2018-11-15
7595 require payee code

pkarman:
refactor of #7816


app/models/claim_review_intake.rb
```diff
@@ -23,6 +23,7 @@ def review!(request_params)
   rescue ActiveRecord::RecordInvalid => _err
     # propagate the error from invalid column to the user-visible reason
     if detail.errors.messages[:benefit_type].include?(ClaimantValidator::PAYEE_CODE_REQUIRED)
+      detail.validate
```
pkarman: elegant fix!

PR #7872, opened 2018-11-15
Limit org task unique check to same appeal

lowellrex:
#7833 introduced logic to prevent the creation of duplicate generic tasks at the organization level, however it did not scope that check to only the current appeal. This PR changes that oversight.

[Example error caused by this bug](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2964/).
> Caseflow::Error::DuplicateOrgTask: Appeal 555 already has an active task of type BvaDispatchTask assigned to BvaDispatch. No action necessary


shanear:
This bug caused an intake prod outage for several hours, I think it might be best to write up a post mortem for this.

PR #7877, opened 2018-11-15
Fixes the legacy opt in modal displaying when legacy opt in is not set

jiahuang:
Connects https://github.com/department-of-veterans-affairs/caseflow/issues/7335

Also renables tests

### DIFF
--- a/app/models/decision_review.rb
+++ b/app/models/decision_review.rb
@@ -103,7 +103,7 @@ class DecisionReview < ApplicationRecord
   end
 
   def serialized_legacy_issues
-    return [] if not legacy_opt_in_approved
+    return [] if !legacy_opt_in_approved
     active_or_eligible_legacy_appeals.map do |legacy_appeal|
       {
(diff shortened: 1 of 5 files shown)


app/models/decision_review.rb
```diff
@@ -103,6 +103,7 @@ def mark_rating_request_issues_to_reassociate!
   end
 
   def serialized_legacy_issues
+    return [] if !legacy_opt_in_approved
```
shanear: this isn't correct. It should show the issues modal regardless if the opt in was approved or not. It should not show them if the feature flag was turned off.

```suggestion
    return [] if !FeatureToggle.enabled?(:intake_legacy_opt_in)
```

shanear: `legacy_opt_in_approved` will toggle whether the matched issue is eligible or not.

leikkisa: I forgot that we still need the legacy issues for ineligibility

PR #7878, opened 2018-11-15
fix legacy opt in with feature disabled

jiahuang:
Connects #7335

Checks the feature toggle instead of legacy_opt_in_approved


app/models/decision_review.rb
```diff
@@ -103,7 +103,7 @@ def mark_rating_request_issues_to_reassociate!
   end
 
   def serialized_legacy_issues
-    return [] if !legacy_opt_in_approved
+    return [] if !FeatureToggle.enabled?(:intake_legacy_opt_in, user: current_user)
```
kencheeto: ruby has `unless` which reads nicer sometimes (`return [] unless foo` vs `return [] if !foo`) but i'm ok either way

spec/feature/intake/appeal_spec.rb
```diff
@@ -606,8 +606,9 @@ def check_row(label, text)
       expect(page).to have_content("Left knee granted")
     end
 
-    scenario "adding issue with legacy opt in not set" do
-      start_appeal(veteran, legacy_opt_in_approved: nil)
+    scenario "adding issue with legacy opt in disabled" do
+      FeatureToggle.disable!(:intake_legacy_opt_in)
```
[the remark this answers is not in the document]
kencheeto: Or a way to stub them that is automatically reset after tests!

spec/feature/intake/appeal_spec.rb
```diff
@@ -607,7 +607,9 @@ def check_row(label, text)
     end
 
     scenario "adding issue with legacy opt in disabled" do
-      FeatureToggle.disable!(:intake_legacy_opt_in)
+      allow(FeatureToggle).to receive(:enabled?).and_call_original
+      allow(FeatureToggle).to receive(:enabled?).with(:intake_legacy_opt_in, user: current_user).and_return(false)
```
pkarman: 👍

Issue #7884, opened 2018-11-16
[BE] Create a join model so request issues and decision issues can have many-to-many relationship


aroltsch:
1. Create a join model RequestDecisionIssue and the associated table. 
2. Start writing to the new table for HLR and SC as well as to the `source_request_issue_id` column
3. Deploy
4. Then write a migration to migrate existing decision_issues to use the new table
5. Deploy
6. Start reading from the new table for all cases
7. Deploy
8. Remove `source_request_issue_id` from the `decision_issues`

pkarman:
It seems important to be able to explicitly know if a `RequestIssue` is in response to a `DecisionIssue` (its "source") or vice versa. We might need additional column(s) on the join model in order to designate that relationship.

PR #7886, opened 2018-11-16
Create a join model for request issues and decision issues

aroltsch:
This is the first PR for https://github.com/department-of-veterans-affairs/caseflow/issues/7884

### Description
1. Create a join model RequestDecisionIssue and the associated table.
2. Start writing to the new table for HLR and SC as well as to the source_request_issue_id column

### DIFF
--- a/app/models/rating_issue.rb
+++ b/app/models/rating_issue.rb
@@ -47,20 +47,22 @@ class RatingIssue
     # if a DecisionIssue already exists then do not touch it. These should be immutable.
     return if decision_issue
 
-    created_decision_issue = DecisionIssue.create!(
-      source_request_issue: source_request_issue,
-      rating_issue_reference_id: reference_id,
-      participant_id: participant_id,
-      promulgation_date: promulgation_date,
-      decision_text: decision_text,
-      profile_date: profile_date
-    )
+    ActiveRecord::Base.transaction do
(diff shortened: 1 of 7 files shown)


app/models/rating_issue.rb
```diff
@@ -47,14 +47,20 @@ def save_decision_issue
     # if a DecisionIssue already exists then do not touch it. These should be immutable.
     return if decision_issue
 
-    DecisionIssue.create!(
+    created_decision_issue = DecisionIssue.create!(
```
pkarman: we probably want to wrap the 2 `create` calls in a `transaction`

db/migrate/20181116155733_create_request_decision_issues.rb
```diff
@@ -0,0 +1,9 @@
+class CreateRequestDecisionIssues < ActiveRecord::Migration[5.1]
+  def change
+    create_table :request_decision_issues do |t|
+      t.integer :request_issue_id
+      t.integer :decision_issue_id
+      t.timestamps
```
pkarman: just an observation that we haven't consistently used the AR timestamps on our models. I like doing it, and maybe we should formally endorse that pattern system-wide.

app/models/rating_issue.rb
aroltsch: good point sir!

PR #7891, opened 2018-11-16
Fix flaky intake stats spec

kencheeto:


spec/feature/intake/intake_stats_spec.rb
```diff
@@ -34,21 +34,23 @@
            established_at: Time.zone.now,
            end_product_status: "HELLA_ACTIVE").issues.create!(description: "this is the only issue here")
 
-    create(:ramp_election,
```
kencheeto: sometimes this is not the fifth ramp election created

Issue #7889, opened 2018-11-16
Intake | Create decision issues for request issues with no matched rating issue


shanear:
When syncing decision issues, handle the scenario where there is no matching request issue:

<img width="784" alt="screen shot 2018-11-16 at 2 11 20 pm" src="https://user-images.githubusercontent.com/1596259/48641989-94cce180-e9a9-11e8-9168-063e422e4869.png">
<img width="779" alt="screen shot 2018-11-16 at 2 11 27 pm" src="https://user-images.githubusercontent.com/1596259/48642000-98f8ff00-e9a9-11e8-998c-641dc356cee4.png">
<img width="776" alt="screen shot 2018-11-16 at 2 11 36 pm" src="https://user-images.githubusercontent.com/1596259/48642004-9ac2c280-e9a9-11e8-8124-35beb304ab14.png">
<img width="778" alt="screen shot 2018-11-16 at 2 11 40 pm" src="https://user-images.githubusercontent.com/1596259/48642007-9c8c8600-e9a9-11e8-885f-56fb31752af3.png">

## Acceptance Criteria
When syncing decision issues for cleared End Products:
- For Rating EPs, check to see if a rating associated with that end product exists:
  - If the rating exists, then decision issues should be created for each request issue, regardless if a matching rating issue is found inside of that rating or not. If no matching rating issue is found, use the disposition of the request issue contention to create a decision issue. 
  - If the rating does not exist, we should continue to poll the veteran's ratings until the rating associated with the cleared HLR or SC is found.

For Nonrating EPs
- Create decision issues using the dispositions of the request issues.

## Technical notes
A lot of the AC is handled by https://github.com/department-of-veterans-affairs/caseflow/issues/7652, how ever the solution implemented for that ticket assumes that all rating request issue contentions will be mapped to a rating issue. I found out after digging into the specifics of the VBMS-R changes, that this assumption is not always true.

We will likely need to change the logic here https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/end_product_establishment.rb#L174 to look for a rating mapped to the EP establishment, and if found, the decision issues should be synced either to a matching rating issue or using the disposition of the contention.

shanear:
holy crap this is working 👍

```
[...]

PR #7898, opened 2018-11-16
connecting decision issues with contentions

jiahuang:
Connects https://github.com/department-of-veterans-affairs/caseflow/issues/7889

Decision Issues are created from Request Issues by the following logic:
- If the request issue is a rating request issue 
  - if the rating request issue's ep has a rating
     - if the rating has a matching rating issue, use that to create the decision issue
     - if there is no matching rating issue, look for a matching contention to create the decision issue
     - if neither rating issue or contention exists, error out
  - if rating does not exist, wait until the rating does exist
- If the request issue is a nonrating request issue
  - look for a matching contention to create the decision issue
  - if no matching contentions exist, error out

This PR also does the following:
* Makes Decision issue's disposition validation apply only on Appeals
* Adds decision_review to Decision Issue which references the hlr or sc


app/models/end_product_establishment.rb
```diff
@@ -184,6 +183,29 @@ def sync_decision_issues!
...
   end
 
+  def sync_decision_issues_from_contentions!
+    # rating issues may not exist for every contention
+    # - rating eps may not have matching ratings for each contention
+    # - nonrating eps don't have ratings
+    # in these cases, use the contention desposition to create a decision issue
+
+    # find request issues with no matching ratings
+    unmatched_request_issues = request_issues - potential_decision_ratings(&:source_request_issue)
```
jiahuang: I don't think this logic is correct because `potential_decision_ratings` could be `[]` but those ratings may still be active at a later time.

I think the correct way is:
- for nonrating request issues just check the contention
- for rating request issues, wait until `potential_decision_ratings` has something in it, then do this logic

open qs:
- Will the contentions exist at this point or can they be like `potential_decision_ratings` where it might be blank?
- will all the potential_decision_ratings be populated at once? Or can it go from an empty array to an array with 2 things to an array with 5 things?

app/models/request_issue.rb
```diff
@@ -149,6 +149,27 @@ def previous_request_issue
...
+    # do not create another decision issue for this request issue if one already exists
+    return if decision_issue_reference_id
+
+    ActiveRecord::Base.transaction do
+      created_decision_issue = DecisionIssue.create!(
+        source_request_issue: self,
+        participant_id: review_request.veteran.participant_id,
+        disposition: contention.disposition,
+        # contention has start_date and submit_date
+        disposition_date: contention.start_date
```
jiahuang: not sure what the disposition date should be. contention has a start_date and a submit_date according to https://github.com/department-of-veterans-affairs/connect_vbms/blob/8b292928138e6e51d2e8fb6b0ac1f3aa77f0db43/spec/requests/list_contentions_spec.rb#L33-L34

app/models/request_issue.rb
```diff
@@ -149,6 +149,27 @@ def previous_request_issue
 
   private
 
+  def save_decision_issue(contention)
+    # do not create another decision issue for this request issue if one already exists
+    return if decision_issue_reference_id
```
jiahuang: not sure if this is right either given that we're going to have many <-> many between request issue and decision issues

app/models/end_product_establishment.rb
leikkisa: Contentions will exist once an eligible request issue is submitted and processed
Don't know the answer to the second question. I think it's safest to assume the latter.  I think that judges do make most decisions at once, but that Veterans sometimes have some period of extra time after a hearing to submit something (I forget what that's called), and that is fairly common.

leikkisa: To answer the comment below - All eligible request issues (that haven't been removed) should have a contention reference id.  There should always be a matching contention, because contentions made from Caseflow have to be edited in Caseflow (I suspect that extends to removal).

app/models/end_product_establishment.rb
```diff
@@ -184,6 +183,29 @@ def sync_decision_issues!
     resolve_synced_decisions(synced_rating_issues)
   end
 
+  def sync_decision_issues_from_contentions!
+    # rating issues may not exist for every contention
+    # - rating eps may not have matching ratings for each contention
+    # - nonrating eps don't have ratings
+    # in these cases, use the contention desposition to create a decision issue
```
leikkisa: small typo in "desposition"

app/models/request_issue.rb
leikkisa: I believe you are right, that we would want to be able to have it be many-to-many, so if there's another decision issue, you would still save this one.

app/models/end_product_establishment.rb
```diff
@@ -184,6 +183,29 @@ def sync_decision_issues!
...
+
+    # find request issues with no matching ratings
+    unmatched_request_issues = request_issues - potential_decision_ratings(&:source_request_issue)
+    unmatched_request_issues.each do |request_issue|
+      found_contention = contentions.detect { |contention| contention.contention_id == request_issue.contention_reference_id }
```
leikkisa: Maybe check that there is a disposition also?

app/models/request_issue.rb
leikkisa: It would be useful to get the contention data for a contention with a disposition, maybe there's a new date altogether?  Because I don't see the disposition in the spec above so potentially there's more data that's not listed there.

spec/models/end_product_establishment_spec.rb
```diff
@@ -594,43 +595,177 @@
...
+            :request_issue,
+            review_request: higher_level_review,
+            end_product_establishment: end_product_establishment,
+            contention_reference_id: contention_ref_id_2
+          )
+        ]
+      end
+      let(:rating){Generators::Rating.build(issues: [])}
+
+      it "will retry rating request issue later" do
```
jiahuang: is this the only test scenario needed if the rating request issue does not have any rating issues? We can just assume that it will eventually have some rating issue?

Or can there be a situation where there's no rating issues but we need to create decision issues off of the contention?

app/models/end_product_establishment.rb
```diff
@@ -184,6 +183,36 @@ def sync_decision_issues!
     resolve_synced_decisions(synced_rating_issues)
   end
 
+  def has_nonrating_code
+    code === HigherLevelReview::END_PRODUCT_NONRATING_CODE ||
+      code === SupplementalClaim::END_PRODUCT_CODES[:nonrating]
```
shanear: I feel like it might be best to put these in a single array constant somewhere, since there will likely be more to come. It also makes this method nicer:

```
DecisionReview.nonrating_ep_codes.include?(code)
```

app/models/end_product_establishment.rb
```diff
@@ -184,6 +183,36 @@ def sync_decision_issues!
...
+      end
+    end
+  end
+
+  def sync_decision_issues!
+    return unless status_cleared?
+    if has_nonrating_code
+      sync_decision_issues_from_dispositons!(request_issues)
+    else
+      request_issues_to_sync_by_disposition = sync_decision_issues_from_ratings!
```
shanear: it's not super clear that the return value of `sync_decision_issues_from_ratings!` is the list of request issues that didn't get a decision issue. I feel like there might be a more clearer way to structure these methods

app/models/end_product_establishment.rb
```diff
@@ -250,13 +279,33 @@ def potential_decision_ratings
...
+    # we can assume that any rating request issues that are not matched
+    # will never get a matching rating issue
+    rating_request_issues_to_match_by_disposition = []
+    if synced_rating_request_issue_for_this_ep.length > 0 || rating_request_issues.length > 0
+      rating_request_issues_to_match_by_disposition = request_issues.reject(&:processed?) - rating_request_issues - synced_rating_request_issue_for_this_ep
+    end
+    # if synced_rating_request_issue_for_this_ep is empty
+    # we need to wait until some rating issues are returned
+
+    rating_request_issues_to_match_by_disposition
```
shanear: this method is a beast. We should try and factor it down.

Part of me thinks we can push a lot of this logic into `RequestIssue`, then `EndProductEstablishment` is just responsible for mapping over that?

app/models/end_product_establishment.rb
```diff
@@ -171,8 +171,7 @@ def sync!
     raise BGSSyncError.new(e, self)
   end
 
-  def sync_decision_issues!
-    return unless status_cleared?
+  def sync_decision_issues_from_ratings!
```
shanear: There is no place where we seem to be checking for the rating associated with the EP. Is that coming in a later PR?

spec/models/end_product_establishment_spec.rb
```diff
@@ -578,6 +578,7 @@
     end
 
     let(:contention_ref_id) { "123456" }
+    let(:contention_ref_id_2) { "234567" }
```
shanear: it seems like we could define this inside of the test scope where it's first referenced.

spec/models/end_product_establishment_spec.rb
leikkisa: If a contention gets a disposition of denied, would that be a case where it doesn't get a new rating, or would it still get a new rating?

PR #7901, opened 2018-11-17
Change display name of VLJ support team (colocated)

lowellrex:
During a meeting with several Board teams yesterday, a Board representative informed us that the VLJ support management is now referred to as the "VLJ Support Staff". This PR changes their display name throughout the application ("assign to team" task actions, co-located team queue, and one place in the copy of the case details page).

### DIFF
--- a/app/models/organizations/colocated.rb
+++ b/app/models/organizations/colocated.rb
@@ -1,5 +1,5 @@
 class Colocated < Organization
   def self.singleton
-    Colocated.first || Colocated.create(name: "VLJ Support Staff", url: "vlj-support-staff")
+    Colocated.first || Colocated.create(name: "VLJ Support Staff", url: "vlj-support")
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/organizations/colocated.rb
```diff
@@ -1,5 +1,5 @@
 class Colocated < Organization
   def self.singleton
-    Colocated.first || Colocated.create(name: "VLJ Support Management", url: "vlj-support-management")
+    Colocated.first || Colocated.create(name: "VLJ Support Staff", url: "vlj-support-staff")
```
lowellrex: This change has already been made to the row in the live database. This will mostly affect demo, testing, and local development environments.

laurjpeterson: I think we can shorten name to `VLJ Support`, and the url to `vlj-support`, just in case they change their name in the future

PR #7910, opened 2018-11-19
Migrate existing decision issues to use request_decision_issues table

aroltsch:
This is step #4 for the ticket https://github.com/department-of-veterans-affairs/caseflow/issues/7884


db/migrate/20181119181014_move_existing_records_to_request_decision_issues.rb
```diff
@@ -0,0 +1,7 @@
+class MoveExistingRecordsToRequestDecisionIssues < ActiveRecord::Migration[5.1]
+  def change
+    DecisionIssue.all.find_each do |decision_issue|
+  	  RequestDecisionIssue.find_or_create_by(request_issue: decision_issue.source_request_issue, decision_issue: decision_issue)
```
pkarman: I don't think this is reversible via normal Rails `change` logic, is it?

pkarman: i.e. you might need a separate `up` and `down` method

aroltsch: it is not reversible. I thought about having up and down method but how would you reverse it? The only solution is to delete all records from request_decision_issues, no?

pkarman: Yes, I think you're right. Still, that seems appropriate to me.

If you changed the `find_or_create_by` to a simple `create` in `up` then you could feel more confident to do a `destroy` in the `down` method.

pkarman: i.e. you'd have to do more a more complex `find_by` in both `up` and `down` -- maybe using the presence of an existing `source_request_issue`?

pkarman: otoh, I'm not sure I see the negatives to leaving the "extra" `request_decision_issues` around after a rollback, since the next migration would just find them.

Maybe it's ok to leave as-is.

aroltsch: As I mentioned to @pkarman  on slack, I can't use `create` because currently in Production there is code writing to both tables: request_decision_issues and decision_issues. And because we have a uniqueness constraint at the DB level on request_issue_id/decision_issue_id we cannot insert duplicates. We have to use find_or_create_by to only insert the "old" ones that were never inserted into request_decision_issues

PR #7912, opened 2018-11-19
Anya/7792 decision issues editing

aroltsch:

### Description

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...


app/models/concerns/case_review_concern.rb
```diff
@@ -21,22 +21,46 @@ def update_task_and_issue_dispositions
...
+      delete_and_create_decision_issues
+    else
+      update_issue_dispositions
+    end
+  end
+
+  def delete_and_create_decision_issues
+    return unless appeal
+    # We will always delete and re-create decision issues on attorney/judge checkout
+    decision_issue_ids_to_delete = appeal.decision_issues.map(&:id)
```
mdbenjam: Not sure, but can this just be `appeal.decision_issues.delete_all`?

mdbenjam: Or maybe it's `destroy_all` and in decision_issues we mark the join table rows to be destroyed when we destroy a decision issue record?

aroltsch: appeal.decision_issues.delete_all doesn't work!
 ```Cannot modify association 'Appeal#decision_issues' because it goes through more than one other association.```

pkarman: It would be nice to have a cascade delete constraint on the join table, either at the Rails level or the db level. It seems like the pattern so far in the app is to do constraints at the Rails level. I think that's `dependent: destroy` on the `decision_issues` association.

aroltsch: Well that's because I have to use `destroy_all` and not `delete_all`. `destroy` runs callbacks

Issue #7892, opened 2018-11-16
Enable admin action functionality for VLJ support + 2 chief group's attorneys


laurjpeterson:
We are rolling out admin action functionality to all VLJ support staff and attorneys from Chief Trueba and Chief Buck's groups. 

## AC
- update features.config file - https://github.com/department-of-veterans-affairs/appeals-deployment/blob/master/ansible/vars/features-config.yml
- VLJ support staff - feature: "colocated_queue",
- attorneys - feature: "attorney_assignment_to_colocated"

### VLJ Support Staff
<img width="181" alt="screen shot 2018-11-19 at 10 21 13 am" src="https://user-images.githubusercontent.com/27704093/48716239-df866d80-ebe4-11e8-83f4-22ee519e4926.png">

### Chief Buck's group

From: Buck, Bethany <Bethany.Buck@va.gov>
Date: Friday, Nov 16, 2018, 10:47 AM
[...]

PR #7913, opened 2018-11-19
Show personalized task queue as home page for all VLJ staff

lowellrex:
Connects #7892.

Currently if a member of the VLJ support staff does not have the [`colocated_queue` feature enabled](https://github.com/department-of-veterans-affairs/appeals-deployment/blob/d93ab9655a883d246b9bb960ddbd63a4148e2cb9/ansible/vars/features-config.yml#L154) then we will show them the search page (`/search`) when they navigate to `appeals.cf.ds.va.gov`. If the `colocated_queue` feature is enabled, then that member of the VLJ support staff would see their personalized task queue (`/queue`). This PR changes that redirect behaviour so that all VLJ support staff will see their personalized task queues if they navigate to `appeals.cf.ds.va.gov`.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -138,7 +138,6 @@ class ApplicationController < ApplicationBaseController
       return false if current_user.admin?
       return false if current_user.organization_queue_user? || current_user.vso_employee?
       return false if current_user.attorney_in_vacols? || current_user.judge_in_vacols?
-      return false if current_user.colocated_in_vacols? && feature_enabled?(:colocated_queue)
       return true
     end
     false


amprokop:
Hm. My interpretation of what you're saying is:


```
previously if the user is colocated and has the case search toggle enabled but didn't have the colocated queue feature toggle enabled, we would send `/` to `/search`. 
previously if the user is colocated and has the case search toggle enabled but has the colocated queue feature toggle enabled, we send `/` to `/queue`

now if the user is colocated and has the case search toggle enabled we always send `/` to `/queue`
```

But when i read this code it looks like

```
now if the user is colocated and has the case search toggle enabled we send `/` to the search page 
```

Can you help me understand if i'm wrong?

lowellrex:
You're 100% correct. Changing the code now

PR #7919, opened 2018-11-19
Allow creation of mail tasks as children of on_hold root tasks

### LINKED ISSUE #7900: Allow creation of mail tasks as children of on_hold root tasks
Currently, [we do not display any options](https://github.com/department-of-veterans-affairs/caseflow/blob/2eef73ea49765c58b93c5f9bccbd2212692ba0d9/app/models/task.rb#L39) on the case details page for RootTasks that are on hold. Even though RootTasks are created with a status of "assigned", they have their status changed to "on_hold" [whenever a child task is created](https://github.com/department-of-veterans-affairs/caseflow/blob/2eef73ea49765c58b93c5f9bccbd2212692ba0d9/app/models/task.rb#L232) (when VSO or judge tasks are created, for instance). This prevents mail team members (and us in demo and our local environment sometimes) from creating mail tasks on cases.

## Acceptance criteria
* Members of the mail team are able to create mail tasks from RootTasks that are on hold

lowellrex:
Resolves #7900.

Currently, [we do not display any options](https://github.com/department-of-veterans-affairs/caseflow/blob/2eef73ea49765c58b93c5f9bccbd2212692ba0d9/app/models/task.rb#L39) on the case details page for RootTasks that are on hold. Even though RootTasks are created with a status of "assigned", they have their status changed to "on_hold" [whenever a child task is created](https://github.com/department-of-veterans-affairs/caseflow/blob/2eef73ea49765c58b93c5f9bccbd2212692ba0d9/app/models/task.rb#L232) (when VSO or judge tasks are created, for instance). This prevents mail team members (and us in demo and our local environment sometimes) from creating mail tasks on cases. This PR addresses that issue.


app/models/tasks/root_task.rb
```diff
@@ -15,6 +15,11 @@ def available_actions(user)
     []
   end
 
+  def no_actions_available?(user)
+    return false if status == Constants.TASK_STATUSES.on_hold && MailTeam.singleton.user_has_access?(user)
```
lowellrex: I don't love how we've added logic into the `RootTask` class that determines actions we can take with `MailTask`s, but I think it is fine for now.

I've been kicking around some thoughts around separating task actions from the tasks they apply to because task actions often have to have knowledge about more than one task (can be available on one task type and result in the creation of a different task type). This isn't the cleanest problem to solve (maybe we would register all of the task actions with some type of mediator when the rails server initializes?) but I think this deserves to be addressed in a tech spec at some point in the near future and this should be a quick fix in the meantime. cc @mdbenjam

Issue #7336, opened 2018-10-10
Legacy Opt-in | Add Issue landing and conf | if ineligible


suzchap:
## Story
[Mural showing all opt-in screens](https://app.mural.co/t/workqueue2001/m/workqueue2001/1538676663322/098da798af658862105ea8d4d5aeb03fa7aaf99a)

When a user has a legacy appeal, they will be allowed to "opt-in" to move one or more of those issues into an AMA lane. The AMA paper forms will have a question on them for the applicant to acknowledge that by checking that box, they understand that any legacy issues listed on that form will be moved out of the legacy lane and into the new AMA lane.

This ticket is focused on what happens if the answer to the "Did they agree to withdraw their issues from the legacy system?" question on the Review page = NO (because this means they didn't agree to opt-in their legacy issues so we're considering them ineligible). 

requires #7308 
requires #7335 
requires #7333 

## Technical Notes

- we will need to style ineligible issues consistently with different kinds of ineligibility reasons

## Acceptance criteria
- When we intake any AMA lane, when adding an issue, if a legacy issue is selected (#7335):
    - if the opt-in checkbox on #7333 is "no", display the issue as ineligible on the add issues page (see copy below)
    - if the opt-in checkbox is "yes" but the legacy issue selected is ineligible (see #7334 for definitions of eligible), display the issue as ineligible on the add issues page (see copy below)
- When an intake is completed for any AMA lane, ineligible issues should be stored in the database with the reason along with the matched legacy issue id
- Ineligible issues should be shown on the confirmation page using the same design introduced in #7308 (there's a section for ineligible issues)

## Copy
[issue selected] is ineligible because the same issue is under review as a Legacy Appeal but the claimant didn’t choose to withdraw their issue from the legacy system. 

[issue selected] is ineligible because the same issue is under review as a Legacy Appeal and that appeal is outside the window of eligibility.

## Mocks
<img width="799" alt="screen shot 2018-10-11 at 12 27 02 pm" src="https://user-images.githubusercontent.com/11219417/46819163-01bcde00-cd51-11e8-92ac-9f7a39e94fbb.png">

<img width="819" alt="screen shot 2018-10-11 at 12 24 39 pm" src="https://user-images.githubusercontent.com/11219417/46819111-e3ef7900-cd50-11e8-86aa-b83508947b57.png">

## How to Validate
(no need, Shane's got this :muscle:)

leikkisa:
[...]

PR #7924, opened 2018-11-20
Handle ineligible legacy opt-in issues

leikkisa:
connects #7336

### DIFF
--- a/client/app/intake/components/LegacyOptInModal.jsx
+++ b/client/app/intake/components/LegacyOptInModal.jsx
@@ -5,6 +5,7 @@ import React from 'react';
 import { formatDateStr } from '../../util/DateUtil';
 import {
   addRatingRequestIssue,
+  addNonratingRequestIssue,
   toggleUntimelyExemptionModal,
   toggleLegacyOptInModal } from '../actions/addIssues';
 import Modal from '../../components/Modal';
@@ -20,10 +21,13 @@ class LegacyOptInModal extends React.Component {
     };
   }
 
+  const { currentIssueAndNotes } = this.props.intakeData
+
   radioOnChange = (value) => {
     this.setState({
       vacolsId: value
     });
+    currentIssueAndNotes.currentIssue.vacolsId = value;
   }
 
   requiresUntimelyExemption = () => {
@@ -31,26 +35,35 @@ class LegacyOptInModal extends React.Component {
       return false;
     }
 
-    return !this.props.intakeData.currentIssueAndNotes.currentIssue.timely;
+    return !currentIssueAndNotes.currentIssue.timely;
   }
 
   onAddIssue = () => {
     // currently just adds the issue & checks for untimeliness
-    // if vacols issue is selected, logic to be implemented by 7336 & 7337 
-    const currentIssue = this.props.intakeData.currentIssueAndNotes.currentIssue;
-    const notes = this.props.intakeData.currentIssueAndNotes.notes;
+    // if vacols issue is selected, logic to be implemented by 7336 & 7337
+    const currentIssue = currentIssueAndNotes.currentIssue;
+    const notes = currentIssueAndNotes.notes;
 
     if (this.requiresUntimelyExemption()) {
       return this.props.toggleUntimelyExemptionModal({ currentIssue,
         notes: this.state.notes });
+    } else if (currentIssue.isRating) {
+        this.props.addRatingRequestIssue({
+        issueId: currentIssue.reference_id,
+        ratings: this.props.intakeData.ratings,
+        isRating: true,
+        vacolsId: this.state.vacolsId,
+        notes
+      });
+    } else {
+      this.props.addNonratingRequestIssue({
+        category: this.state.category.value,
+        description: this.state.description,
+        decisionDate: this.state.decisionDate,
+        timely: true,
+        vacolsId: this.state.vacolsId
+      });
     }
-
-    this.props.addRatingRequestIssue({
-      issueId: currentIssue.reference_id,
-      ratings: this.props.intakeData.ratings,
-      isRating: true,
-      notes
-    });
 
     this.props.toggleLegacyOptInModal();
   }
@@ -80,7 +93,7 @@ class LegacyOptInModal extends React.Component {
 
       return <RadioField
         vertical
-        label={<h3>Notice of Disagreement Date { formatDateStr(legacyIssue.date) }</h3>}
+        label={<h3>Notice of Disagreement Date { formatDateStr(legacyIssue.nod_date) }</h3>}
         name="rating-radio"
         options={radioOptions}
         key={`${index}legacy-opt-in`}
@@ -121,6 +134,7 @@ export default connect(
   null,
   (dispatch) => bindActionCreators({
     addRatingRequestIssue,
+    addNonratingRequestIssue,
     toggleUntimelyExemptionModal,
     toggleLegacyOptInModal
   }, dispatch)
--- a/app/models/decision_review.rb
+++ b/app/models/decision_review.rb
@@ -106,7 +106,8 @@ class DecisionReview < ApplicationRecord
     return [] unless FeatureToggle.enabled?(:intake_legacy_opt_in, user: RequestStore.store[:current_user])
     active_or_eligible_legacy_appeals.map do |legacy_appeal|
       {
-        date: legacy_appeal.nod_date,
+        nod_date: legacy_appeal.nod_date,
+        soc_date: legacy_appeal.soc_date,
         issues: legacy_appeal.issues.map(&:intake_attributes)
       }
     end
--- a/client/app/intake/components/AddIssuesModal.jsx
+++ b/client/app/intake/components/AddIssuesModal.jsx
@@ -72,7 +72,7 @@ class AddIssuesModal extends React.Component {
   }
 
   getNextButtonText = () => {
-    if (this.props.intakeData.legacyIssues.length > 0) {
+    if (this.hasLegacyIssues()) {
       return 'Next';
     }
 
--- a/client/app/intake/components/NonratingRequestIssueModal.jsx
+++ b/client/app/intake/components/NonratingRequestIssueModal.jsx
@@ -5,7 +5,8 @@ import React from 'react';
 import {
   addNonratingRequestIssue,
   toggleUnidentifiedIssuesModal,
-  toggleUntimelyExemptionModal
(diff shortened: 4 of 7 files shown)


app/models/request_issue.rb
```diff
@@ -255,6 +261,26 @@ def check_for_before_ama!
     end
   end
 
+  def check_for_legacy_issue_not_withdrawn!
+    return unless eligible?
+
```
pkarman: I think it would be a little clearer to put a similar guard logic here:
```
return unless vacols_id
```

app/models/request_issue.rb
```diff
@@ -255,6 +261,26 @@ def check_for_before_ama!
...
+    if !review_request.legacy_opt_in_approved && vacols_id
+      self.ineligible_reason = :legacy_issue_not_withdrawn
+    end
+  end
+
+  def check_for_legacy_appeal_not_eligible!
+    return unless eligible?
+    return unless vacols_id
+    return unless review_request.serialized_legacy_appeals
```
pkarman: should this be `review_request.serialized_legacy_appeals.any?` or are you really checking for nil?

app/models/request_issue.rb
```diff
@@ -255,6 +261,26 @@ def check_for_before_ama!
...
+
+    legacy_appeal = review_request.serialized_legacy_appeals.select { |appeal| appeal[:vacols_id] == vacols_id }
```
pkarman: nit: `legacy_appeals` plural since `select` returns an array. Otherwise if you just want the first one, use `find` instead of `select`.

client/app/intake/components/AddedIssue.jsx
```diff
@@ -51,6 +51,10 @@ class AddedIssue extends React.PureComponent {
       errorMsg = INELIGIBLE_REQUEST_ISSUES.previous_higher_level_review;
     } else if (issue.beforeAma) {
       errorMsg = INELIGIBLE_REQUEST_ISSUES.before_ama;
+    } else if (!legacyOptInApproved && issue.vacolsId) {
```
pkarman: nit: glancing over these conditions continues to be a little tricky. these could be grouped like this:

```javascript
} else if (issue.vacoldId) {
  if (!legacyOptInApproved) {
    errorMsg = INELIGIBLE_REQUEST_ISSUES.legacy_issue_not_withdrawn;
  } else if (!issue.eligibleForSocOptIn) {
    errorMsg = INELIGIBLE_REQUEST_ISSUES.legacy_appeal_not_eligible;
  }
}
```

which IMO helps to visually break up the monotony of all the cases and makes it a little easier to read.

spec/feature/intake/appeal_spec.rb
```diff
@@ -8,15 +8,15 @@
     FeatureToggle.enable!(:intake)
     # Test that this works when only enabled on the current user
     FeatureToggle.enable!(:intakeAma, users: [current_user.css_id])
-    FeatureToggle.enable!(:intake_legacy_opt_in)
+    FeatureToggle.enable!(:intake_legacy_opt_in, users: [current_user.css_id])
 
     Time.zone = "America/New_York"
-    Timecop.freeze(Time.utc(2018, 5, 20))
+    Timecop.freeze(Time.utc(2018, 11, 28))
```
pkarman: nit: I don't know what's magical about these particular dates. It might help to name them a little, either as constants or fixtures of some kind. I think we discussed this at our last retro.

spec/feature/intake/appeal_spec.rb
```diff
@@ -41,16 +41,16 @@
                               participant_id: "44444444")
   end
 
-  let(:receipt_date) { Date.new(2018, 4, 20) }
+  let(:receipt_date) { Date.new(2018, 9, 20) }
 
   let(:untimely_days) { 372.days }
 
-  let(:profile_date) { Date.new(2017, 11, 20).to_time(:local) }
+  let(:profile_date) { Date.new(2018, 9, 15).to_time(:local) }
```
pkarman: same comment about dates.

spec/feature/intake/appeal_spec.rb
```diff
@@ -382,46 +382,46 @@ def check_row(label, text)
...
+    expect(page).to have_content("Left knee granted 2 (already selected for issue 1)")
     expect(page).to have_css("input[disabled][id='rating-radio_xyz123']", visible: false)
 
     # Add nonrating issue
     click_intake_no_matching_issues
     add_intake_nonrating_issue(
       category: "Active Duty Adjustments",
       description: "Description for Active Duty Adjustments",
-      date: "04/19/2018"
+      date: "10/27/2018"
```
pkarman: same comment about dates

spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -11,7 +11,7 @@
     FeatureToggle.disable!(:intake_enable_add_issues_page)
 
     Time.zone = "America/New_York"
-    Timecop.freeze(Time.utc(2018, 5, 26))
+    Timecop.freeze(Time.utc(2018, 11, 28))
```
pkarman: same comment about dates

spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -48,14 +48,12 @@
     User.authenticate!(roles: ["Mail Intake"])
   end
 
-  let(:profile_date) { Date.new(2017, 11, 20).to_time(:local) }
-
-  let(:timely_promulgation_date) { Date.new(2017, 11, 30) }
```
pkarman: I like this var name -- helpful to understand the significance. that's an example of what I'm describing for named dates.

spec/feature/intake/supplemental_claim_spec.rb
```diff
@@ -11,7 +11,7 @@
     FeatureToggle.disable!(:intake_enable_add_issues_page)
 
     Time.zone = "America/New_York"
-    Timecop.freeze(Time.utc(2018, 5, 26))
+    Timecop.freeze(Time.utc(2018, 11, 28))
```
pkarman: same comment about dates

spec/support/intake_helpers.rb
```diff
@@ -90,21 +91,76 @@ def expect_eligible_issue(number)
     expect(find_intake_issue_by_number(number)).to_not have_css(".not-eligible")
   end
 
+  # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/MethodLength
   def setup_legacy_opt_in_appeals(veteran_file_number)
-    # create two legacy appeals with 2 issues each
+    # Active and eligible
```
leikkisa: (per Linting's suggestion) am going to break this up into multiple methods

client/app/intake/components/LegacyOptInModal.jsx
```diff
@@ -17,54 +17,71 @@ class LegacyOptInModal extends React.Component {
...
+      vacolsSequenceId: null,
+      radioKey: '',
+      eligibleForSocOptIn: null
     };
   }
 
   radioOnChange = (value) => {
-    // legacy opt in are keyed off of a combo of both id & vacolsSequenceId
+    // legacy opt in are keyed off of a combo of both vacolsId & vacolsSequenceId
     // NO_MATCH_TEXT does not have a vacolsSequenceId
```
jiahuang: i dont think NO_MATCH_TEXT is used anymore, so this comment along with line 14 can be removed

client/app/intake/components/LegacyOptInModal.jsx
```diff
@@ -17,54 +17,71 @@ class LegacyOptInModal extends React.Component {
...
     const legacyValues = value.split('-');
+    const vacolsSequenceId = legacyValues.length > 1 ? legacyValues[1] : null;
+    const legacyAppeal = this.props.intakeData.legacyAppeals.find((appeal) => appeal.vacols_id === legacyValues[0]);
+    const eligibleForSocOptIn = legacyAppeal ? legacyAppeal.eligible_for_soc_opt_in : null;
```
jiahuang: nit: can also be simplified to `legacyAppeal && legacyAppeal.eligible_for_soc_opt_in` which would return false if `legacyAppeal` does not exist. False is clearer than null imo

app/models/request_issue.rb

app/models/request_issue.rb
leikkisa: When the feature flag is turned off it is told to return an empty array, which makes it so we can't find the appeal's eligibility status.

leikkisa: Good call!

app/models/request_issue.rb

client/app/intake/components/AddedIssue.jsx
leikkisa: 👍

client/app/intake/components/LegacyOptInModal.jsx
leikkisa: Did that happen in another PR? It is still used in this PR

client/app/intake/components/LegacyOptInModal.jsx
leikkisa: 👍

spec/feature/intake/appeal_spec.rb
leikkisa: Yeah, I want to do some test refactoring, but I think we should do that in a separate PR.

leikkisa: PS, some of the changes matter but I think there were some that didn't that I didn't completely clean up.  Took some experimentation to fix the broken things :-P.

spec/feature/intake/higher_level_review_spec.rb
leikkisa: I think it will make more sense to do this in a refactor, because currently not all of the tests are using the named variables that do exist.  I am a big fan of making them more consistent.

app/models/issue.rb
```diff
@@ -185,15 +185,14 @@ def description_attributes
     }
   end
 
-  def intake_attributes(is_untimely: false)
+  def intake_attributes
     {
-      id: id,
+      vacols_id: id,
```
shanear: I think this might be even more descriptive as `appeal_vacols_id`

leikkisa: Oh, sorry I didn't respond to this sooner!  We decided to stay consistent with the naming that was already used elsewhere in the app, but I agree that's more descriptive (I didn't realize that it was appeal-level until you told me). I'll add a comment in my current PR to clarify.

PR #7925, opened 2018-11-20
Change which legacy issues show up

leikkisa:
connects #7334

Show all active or eligible issues if the Veteran agrees to withdraw their legacy issues.
Only show active issues if not.


app/models/decision_review.rb
```diff
@@ -102,22 +102,32 @@ def mark_rating_request_issues_to_reassociate!
...
-  def active_or_eligible_legacy_appeals
-    @active_or_eligible_legacy_appeals ||= LegacyAppeal
+  def matchable_legacy_appeals
+    @matchable_legacy_appeals ||= LegacyAppeal
       .fetch_appeals_by_file_number(veteran_file_number)
-      .select(&:eligible_for_soc_opt_in?)
+      .select(&:matchable_to_request_issue?)
+  end
+
+  def active_matchable_legacy_appeals
```
shanear: this can probably just be `active_legacy_appeals`

app/models/decision_review.rb
```diff
@@ -102,22 +102,32 @@ def mark_rating_request_issues_to_reassociate!
     request_issues.select(&:rating?).each { |ri| ri.update!(rating_issue_associated_at: nil) }
   end
 
-  def serialized_legacy_issues
+  def serialized_legacy_appeals
     return [] unless FeatureToggle.enabled?(:intake_legacy_opt_in, user: RequestStore.store[:current_user])
-    active_or_eligible_legacy_appeals.map do |legacy_appeal|
+
+    # If a Veteran does not opt-in to withdraw legacy appeals, do not show inactive appeals
+    available_legacy_appeals = legacy_opt_in_approved ? matchable_legacy_appeals : active_matchable_legacy_appeals
```
shanear: this could probably be split out into a private method

PR #7927, opened 2018-11-20
Use RequestDecisionIssue model for decision issues and request issues

aroltsch:
This is step 6 for https://github.com/department-of-veterans-affairs/caseflow/issues/7884


app/models/request_issue.rb
```diff
@@ -144,7 +144,7 @@ def contested_decision_issue
   end
 
   def previous_request_issue
-    contested_decision_issue.try(:source_request_issue)
+    contested_decision_issue && contested_decision_issue.request_issues.first
```
aroltsch: This method is used in tests only.

db/migrate/20181119212851_remove_source_request_issue_id_index.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveSourceRequestIssueIdIndex < ActiveRecord::Migration[5.1]
+  def change
+  	remove_index :decision_issues, column: :source_request_issue_id
+  	change_column_null :decision_issues, :source_request_issue_id, true
```
pkarman: why not just drop this column? or is that a follow-on PR?

aroltsch: It has to be done in the next PR to avoid downtown during deployment. Because the existing code in production could still be trying to write to this column and if we drop the column before the new code is deployed we will get an error.

aroltsch: Although I don't think we will get an error because we have not been even creating decision issues :). Better be safe though

Issue #7313, opened 2018-10-09
AMA Edit | Confirmation for HLR and SC


suzchap:
## Story
For any AMA forms, when a user needs to edit issues, they will click the "Edit in Caseflow" button in VBMS to return to the "Add issues landing page" where they will be able to add new issues and/or remove issues added previously. There are a few "are you sure?" type modals to help reduce errors. Entire flow via [Mural](https://app.mural.co/t/workqueue2001/m/workqueue2001/1537791678087/94469b0cdf1a02a7f57a7628702d36db2f079430)

This ticket is focused on the Edit Confirmation page that you get after adding or removing at least one issue and clicking "Save" from the Add Issue Landing page to update it to match the new design and additional copy.

requires #7303 

## Technical Notes
- keep an eye out for test duplication refactors across feature tests, but put any significant new work into a new issue

## Acceptance criteria
- When the edit issues page for a Higher Level Review or Supplemental Claim is submitted:
  - The confirmation page should be styled according to the mock below
  - When an EP will be canceled due to the edit (see: https://github.com/department-of-veterans-affairs/caseflow/issues/6214) then show that on the confirmation page according to the copy below.
  - When an EP will be created due to the edit (see: https://github.com/department-of-veterans-affairs/caseflow/issues/6213) then show that on the confirmation page according to the copy below.
  - When the contentions on an EP are updated due to the edit (see: https://github.com/department-of-veterans-affairs/caseflow/issues/6212), then show that according to the copy below.

## Notes
- The ordering of these messages is not particularly important.
- The styling of the confirmation page should match the styling of the original intake confirmation page (see: https://github.com/department-of-veterans-affairs/caseflow/issues/7303)
- Out of scope: confirmation page for Appeals
## Copy

[Veteran name]’s (ID #XXXXXXXXX) Request for [FORM NAME] (VA Form XX-XXXX) has been processed. Go to VBMS claim details and reload to view updates.
Edit the notice letter to reflect the status of requested issues.

[if an EP was canceled]
[ep_type] EP is being canceled.
[subtext] All contentions on this EP were removed

[if an EP was created]
A [ep type] EP is being established:
[subtext for each contention]
Contention: [contention text]
[if informal conference tracked item was added]
Informal Conference Tracked Item

[if the contentions on an EP were updated]
Contentions on [ep_type] EP are being updated:
[subtext for each contention]
Contention: [contention text]

[...]

Issue #7304, opened 2018-10-05
Add issue edge cases | Confirmation | if Unidentified issue


suzchap:
(Not technically about the edge cases but needed to complete them.)

## Story
When a user adds issues, they may encounter edge cases where the issue is ineligible. We will not indicate that an issue isn't eligible on any of the add issue selection pages but will indicate it on the Add issue landing page and Confirmation Page so they have the information they need to inform the applicant in the letter they send following intake data entry. If an issue is ineligible in will get logged in Caseflow but not VBMS. [Mural showing edge cases](https://app.mural.co/t/workqueue2001/m/workqueue2001/1537791678087/94469b0cdf1a02a7f57a7628702d36db2f079430).

This ticket is focused on adding an alert to the new confirmation page (#7303) when an "Unidentified issue" (via this screen #6990 ) has been added. This applies to all AMA lanes. 

requires #7303 
requires #6990
requires #7313 

## Technical notes
- see details for how we store whether or not an issue is unidentified in https://github.com/department-of-veterans-affairs/caseflow/pull/7349

## Acceptance criteria
- After an intake has been completed for any AMA lane, if the review contains any unidentified issues, the confirmation screen will show it as an alert box with instructions on how to come back and edit. 
- After an intake has been edited for any AMA lane, if the review contains any unidentified issues, the confirmation screen will show it as an alert box with instructions on how to come back and edit. 
- If there are multiple unidentified issues, show them on separate lines in the same alert box

## Copy
Unidentified issue
There is still an unidentified issue that needs to be resolved before sending the notice letter.  To edit, go to VBMS claim details and click the “Edit in Caseflow” button.  
Unidentified issue: no issue matched for requested “service connection left shoulder condition.”

## Mocks
<img width="744" alt="screen shot 2018-10-06 at 9 57 15 am" src="https://user-images.githubusercontent.com/11219417/46572094-3ba37400-c94e-11e8-8c00-c2c35ea92256.png">

## How to Validate
(no need, Shane's got this :muscle:)

shanear:
Looks good!

![screen shot 2018-11-26 at 12 32 31 pm](https://user-images.githubusercontent.com/1596259/49031238-6b654000-f177-11e8-844d-805db84c1dc7.png)

PR #7936, opened 2018-11-20
Add alert for unidentified issues on intake confirmation

kencheeto:
Connects #7304 

### Description
Displays an alert for unidentified issues on intake completion. The issue says to add this to both intake and edit, but the edit page is coming later in #7313, so this just adds to the intake confirmation page.

Two notes:
1. I think the confirmation page needs some refactoring soon, but that would come in a followup
1. The spacing is too big, that will also get tweaked in a followup

screen shot 2018-11-20 at 14 05 57

### Acceptance Criteria
- After an intake has been completed for any AMA lane, if the review contains any unidentified issues, the confirmation screen will show it as an alert box with instructions on how to come back and edit.
- If there are multiple unidentified issues, show them on separate lines in the same alert box

### Testing Plan
1. Complete any decision review with unidentified issues, then view the confirmation page.

### DIFF
--- a/client/app/intake/pages/decisionReviewIntakeCompleted.jsx
+++ b/client/app/intake/pages/decisionReviewIntakeCompleted.jsx
@@ -10,6 +10,7 @@ import Alert from '../../components/Alert';
 
 const leadMessageList = ({ veteran, formName, requestIssues }) => {
   const unidentifiedIssues = requestIssues.filter((ri) => ri.isUnidentified);
+
   if (unidentifiedIssues.length === 0) {
     return [
       `${veteran.name}'s (ID #${veteran.fileNumber}) ` +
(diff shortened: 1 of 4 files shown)


client/app/intake/pages/decisionReviewIntakeCompleted.jsx
```diff
@@ -6,6 +6,35 @@ import { PAGE_PATHS, INTAKE_STATES, FORM_TYPES } from '../constants';
...
+        'If you need to edit this, go to VBMS claim details and click the “Edit in Caseflow” button.',
+      <strong>Edit the notice letter to reflect the status of requested issues.</strong>
+    ];
+  }
+
+  const unidentifiedIssuesAlert = <Alert type="warning">
+    <h2>Unidentified issue</h2>
+    <p>There is still an unidentified issue that needs to be resolved before sending the notice
+    letter. To edit, go to VBMS claim details and click the “Edit in Caseflow” button.</p>
+    {unidentifiedIssues.map((ri, i) => <p className="cf-red-text" key={i}>
```
jiahuang: can the key be something like "{i + "-alert"} or something? I just like namespaced keys in case there's another iterator somewhere on the page

PR #7941, opened 2018-11-20
Do not throw exception when completing QRTasks

lowellrex:
Resolves the bug causing the `DuplicateOrgTask` errors when members of the Quality Review team attempted to mark a `QualityReviewTask` complete.

The reason we were throwing these errors is because we were calling the `when_child_task_completed()` method twice on the `QualityReviewTask` assigned to the `QualityReview` organization if that task had a child `QualityReviewTask` assigned to a real person. After the assignee [marks that child `QualityReviewTask` complete](https://github.com/department-of-veterans-affairs/caseflow/blob/b6823101b70bd7ece9d88c6c8344c0588dba25c5/app/models/task.rb#L124), [the `after_update` hook](https://github.com/department-of-veterans-affairs/caseflow/blob/b6823101b70bd7ece9d88c6c8344c0588dba25c5/app/models/task.rb#L15) calls [`update_parent_staus()`](https://github.com/department-of-veterans-affairs/caseflow/blob/b6823101b70bd7ece9d88c6c8344c0588dba25c5/app/models/task.rb#L228) which [calls `mark_as_complete!()`](https://github.com/department-of-veterans-affairs/caseflow/blob/b6823101b70bd7ece9d88c6c8344c0588dba25c5/app/models/task.rb#L223) on the parent task (by way of `when_child_task_completed` and `update_status_if_children_tasks_are_complete`). We then return into the child task's `mark_as_complete!()` method call to run the [final line of that function](https://github.com/department-of-veterans-affairs/caseflow/blob/b6823101b70bd7ece9d88c6c8344c0588dba25c5/app/models/task.rb#L126) which again calls `mark_as_complete!()` on the parent task. It is this second call to `mark_as_complete!()` on the parent that raises the `DuplicateOrgTask` error.

This change eliminates the duplicate creation of the `BvaDispatchTask` by removing the `after_update` hook. Alternatively we could remove the call to `parent.mark_as_complete!` in the `mark_as_complete!` function. I prefer the approach implemented here because 1) we already call `mark_as_complete!` almost everywhere we change a task's status to complete (cleaned up one place we didn't in this PR), 2) I find active record callbacks indirect and non-obvious which can lead to bugs (like this one, for example), and most importantly in my mind 3) the active record callback prevents us from logging into a rails console and changing the status of a task to complete without running the `after_update` callback which we might not always want to do (I can imagine a scenario where an organization-level task is completed, but the individual's task is not so we would want to change that status to `completed` but not re-run `mark_as_complete!` on the parent task).

[Example error](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2969/):
> Caseflow::Error::DuplicateOrgTask - app/models/tasks/generic_task.rb in verify_org_task_unique
> Caseflow::Error::DuplicateOrgTask: Appeal 225 already has an active task of type BvaDispatchTask assigned to BvaDispatch. No action necessary


aroltsch:
@lowellrex Some of the tests are failing. The PR looks okay at the first glance but I am concerned some of the tests are failing.

spec/models/attorney_case_review_spec.rb
```diff
@@ -24,7 +27,7 @@
     subject { AttorneyCaseReview.complete(params) }
 
     context "when ama" do
-      let(:task_id) { create(:ama_attorney_task).id }
+      let(:task_id) { create(:ama_attorney_task, assigned_by: judge, assigned_to: attorney).id }
```
lowellrex: The move from calling `task.update(status: :completed)` to `task. mark_as_complete!` in `CaseReviewConcern.update_task_and_issue_dispositions` resulted in us raising exceptions if the update operation failed due to model validation failures (since we now call `task.update!`). The two changes in this test make it so the task is valid. Prior to this change the task was invalid but we never checked the task's validity so we didn't notice.

PR #7942, opened 2018-11-20
Remove unused judge_assignment_to_attorney feature flag from code

lowellrex:
We rolled judge assignment to all judges [in late September](https://github.com/department-of-veterans-affairs/appeals-deployment/pull/1702). This PR removes this feature flag from our code since we now expect it to always evaluate to true.


spec/controllers/tasks_controller_spec.rb
```diff
@@ -316,13 +308,14 @@
...
+            "action": "address_verification",
             "type": ColocatedTask.name
           }]
         end
 
-        it "should not be successful" do
+        it "should fail assigned_by validation" do
           post :create, params: { tasks: params }
-          expect(response.status).to eq 302
+          expect(response.status).to eq(400)
```
lowellrex: Had to change this from 302 to 400 because [we no longer fail](https://github.com/department-of-veterans-affairs/caseflow/pull/7942/files#diff-55c5b7aecfb519d0e4880eaf2788eb6eL160) in `ApplicationController.can_assign_task?`. Instead we make it all the way into `ColocatedTask.assigned_by_role_is_valid` where we fail because the assigner is a judge. I think this is actually the correct behaviour here.

PR #7950, opened 2018-11-21
Schedule Veteran Default Values as Params

### LINKED ISSUE #7853: Case Details | Assign veterans query param
- Use query param instead of business payload. Send hearing date and parent record id in query param when moving from Assign Hearings page to Case Details page.
- Add query param to modal URL too
- Make regional office in the Schedule Veteran modal read only
- Don't worry about the change Date of Hearing functionality

lomaxap:
Resolves #7853 

### Description
- Passes hearing date, hearing time, and RO key from Assign Veterans as query params for Case Details view, stores values in redux store for use as default values in Schedule Veteran modal. 
- adds Hearing Date dropdown that populates select dropdown with available dates for a given RO
- does NOT change task creation logic


app/controllers/regional_offices_controller.rb
```diff
@@ -4,4 +4,20 @@ def index
       regional_offices: RegionalOffice.ros_with_hearings
     }
   end
+
+  def open_hearing_dates
+    ro = params[:regional_office]
```
sharonwarner: In the HearingDayMapper, we have a regional office validator, can we add that to this logic?

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -1,6 +1,7 @@
 class WorkQueue::AppealSerializer < ActiveModel::Serializer
   attribute :assigned_attorney
   attribute :assigned_judge
+  attribute :sanitized_hearing_request_type
```
sharonwarner: Let's confirm with the queue team that this is the best way to set this up. I'm not sure how they pull data from VACOLS, and I want to make sure we're not affecting their page load times.

db/schema.rb
```diff
@@ -661,12 +661,12 @@
     t.bigint "ineligible_due_to_id"
     t.boolean "untimely_exemption"
     t.text "untimely_exemption_notes"
+    t.string "ineligible_reason"
```
sharonwarner: It looks like nothing was actually changed here, do you mind rechecking this file out?

client/app/hearingSchedule/components/AssignHearings.jsx
```diff
@@ -163,15 +165,23 @@ export default class AssignHearings extends React.Component {
 
   veteransReadyForHearing = () => {
 
+    const { selectedHearingDay, selectedRegionalOffice } = this.props;
+    const date = moment(selectedHearingDay.hearingDate).format('YYYY-MM-DD');
+    const time = getTime(selectedHearingDay.hearingDate);
```
sharonwarner: Something about the time looks off. The hearings are all scheduled for 1 hour later than the user selects.

client/app/components/HearingDateDropdown.jsx
```diff
@@ -0,0 +1,113 @@
...
+    const { readOnly, staticOptions, onChange, value, placeholder } = this.props;
+    const hearingDateOptions = this.hearingDateOptions();
+    const selectedHearingDate = _.find(hearingDateOptions, (opt) => opt.value === value) || {};
+
+    if (!this.props.changePrompt || this.state.editable) {
+      return (
+        <SearchableDropdown
+          name="hearing_date"
+          label="Date of Hearing"
+          staticOptions={staticOptions}
```
sharonwarner: What do these staticOptions do? I don't see them in the SearchableDropdown component.

client/app/components/HearingDateDropdown.jsx
```diff
@@ -0,0 +1,128 @@
+import React from 'react';
```
mdbenjam: I think this should probably reside in `client/app/hearing/components`, since it's connect to the hearings redux store.

client/app/components/RoSelectorDropdown.jsx
```diff
@@ -2,12 +2,21 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import _ from 'lodash';
 import SearchableDropdown from './SearchableDropdown';
+import InlineForm from './InlineForm';
```
mdbenjam: I think this should also probably reside in `client/app/hearing/components`, since it's connect to the hearings redux store, but since it already exists here, no need to move it now.

mdbenjam: Oh just kidding I see this is connected to a common reducer.

client/app/components/HearingDateDropdown.jsx
```diff
@@ -0,0 +1,128 @@
...
+      });
+    });
+
+    if (this.props.staticOptions) {
+      hearingDateOptions.push(...this.props.staticOptions);
+    }
+
+    return hearingDateOptions.sort((d1, d2) => {
+
+      if (d1.value > d2.value) {
```
mdbenjam: nit: I think you can just do `return d1.value - d2.value;` to cover all these cases.

client/app/queue/components/AssignHearingModal.jsx
```diff
@@ -113,11 +119,17 @@ class AssignHearingModal extends React.PureComponent<Props, LocalState> {
   };
 
   formatHearingDate = () => {
-    const dateParts = this.state.selectedDate.split('-');
+    const { selectedHearingDate, selectedHearingTime } = this.props;
```
mdbenjam: Can we do all this with moment instead? https://momentjs.com/

client/app/components/HearingDateDropdown.jsx
mdbenjam: Ah I see we're connecting this to the common store. I think my preference is for common components to just be normal React components if we want to reuse them between apps. i.e. just handle state locally. I think that's simpler than trying to share a reducer.

mdbenjam: If we do use a shared reducer, I think it should be called something other than `common`. These are pretty hearing specific things. Maybe `hearings_common`

client/app/queue/components/AssignHearingModal.jsx
```diff
@@ -86,24 +76,40 @@ class AssignHearingModal extends React.PureComponent<Props, LocalState> {
     super(props);
 
     this.state = {
-      selectedDate: '',
-      selectedTime: '',
-      roEdit: false,
-      dateEdit: false
+      timeOptions: props.appeal.sanitizedHearingRequestType === 'video' ?
```
mdbenjam: Does this belong in state? Does it ever change? Seems like a method `timeOptions(hearingType)` might be better.

client/app/components/RoSelectorDropdown.jsx
lomaxap: Ok, I think for both of the RoSelector and HearingDateDropdown I'll just create a new ticket to change where the dropdown's state is stored. @sharonwarner / @oscarAtNava

client/app/components/HearingDateDropdown.jsx
lomaxap: could have *sworn* that didn't work when I tried it. works now 🤷.

PR #7951, opened 2018-11-21
Throw error if appeals search response text is empty

joeyyang:
### Description
If the request to `/appeals` returns a `response.text` that is empty (`""`), throw an error rather than fail silently.

(`JSON.parse(response.text)` fails silently if `response.text === ""`)

screenshot 2018-11-21 10 55 24


joeyyang:
This change was implemented in https://github.com/department-of-veterans-affairs/caseflow/pull/8098.

Issue #7879, opened 2018-11-15
Refactor Issue eligibility rendering


pkarman:
The code in `client/app/intake/pages/addIssues.jsx` is getting spaghetti like and needs a refactor. In particular the logic for applying eligibility errors is hard to read.

PR #7959, opened 2018-11-21
Refactor added issue into its own component class

pkarman:
connects #7879 

### Description
Refactor added issue rendering and eligibility check into their own component.

### Acceptance Criteria
- [x] Code compiles correctly
- [x] Tests keep passing

### DIFF
--- a/client/app/intake/components/AddedIssue.jsx
+++ b/client/app/intake/components/AddedIssue.jsx
@@ -1,0 +1,87 @@
+import _ from 'lodash';
+import React from 'react';
+
+import INELIGIBLE_REQUEST_ISSUES from '../../../constants/INELIGIBLE_REQUEST_ISSUES.json';
+
+class AddedIssue extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      errorMsg: '',
+      cssKlasses: ['issue-desc']
+    };
+  }
+
+  needsEligibilityCheck() {
+    let { issue, requestIssues } = this.props;
+
+    if (!requestIssues) {
+      return false;
+    }
+    if (issue.ineligibleReason) {
+      return true;
+    }
+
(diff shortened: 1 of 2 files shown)


client/app/intake/components/AddedIssue.jsx
```diff
@@ -0,0 +1,87 @@
...
+          cssKlasses: state.cssKlasses.concat(['not-eligible'])
+        };
+      });
+    }
+  }
+
+  render() {
+    let { issue, issueIdx } = this.props;
+
+    if (this.needsEligibilityCheck) {
```
jiahuang: i think you need to do `this.needsEligibilityCheck()` otherwise `this.needsEligibilityCheck` will return a function which will always evaluate to truthy

client/app/intake/components/AddedIssue.jsx
```diff
@@ -0,0 +1,87 @@
...
+    let existingRequestIssue = _.some(requestIssues, { reference_id: issue.referenceId });
+
+    if (existingRequestIssue) {
+      return false;
+    }
+
+    return true;
+  }
+
+  checkIfEligible() {
```
jiahuang: this function name is throwing me off since it doesn't return a true/false like I would expect from check. Maybe rename to "updateEligibleState"?

client/app/intake/components/AddedIssue.jsx
jiahuang: can also be simplified to `this.needsEligibilityCheck() && this.checkIfEligible();`

kencheeto:
@pkarman I think this is causing unidentified issues to render in black, not red -
screen shot 2018-11-26 at 15 58 13
vs from the commit before this merge commit:
screen shot 2018-11-26 at 16 00 48

pkarman:
good catch. fix in #7993

Issue #7958, opened 2018-11-21
Create add decisions page with just request issues


laurjpeterson:
Part of [Editing AMA issues](https://github.com/department-of-veterans-affairs/caseflow/issues/5961)

When attorneys are ready to check a case out to their judge, they need to add decisions (issue + disposition) for each request issue. 

## AC

Attorneys:
- When the attorney clicks "Decision ready for review", they navigate to the new `Add decisions` screen (and not the `Select dispositions` screen that exists for legacy appeals)

Judges:
- This is replicated for judges. When the judge clicks "Ready for dispatch", they navigate to this page. However, let's title it `Review decisions` instead of `Add decisions` like we do for legacy. 

Both:
- Populate all request issues on the Add decisions screen per mocks
- A button exists beneath each request issue, per mocks. 
- When there isn't a note captured by Intake, hide the "Notes" field. 

## Mocks

![image](https://user-images.githubusercontent.com/40216950/49025336-64373580-f169-11e8-8d0b-63cb6aa977a2.png)

Issue #7957, opened 2018-11-21
Case details | Redesign request issues


laurjpeterson:
Part of [Editing AMA issues](https://github.com/department-of-veterans-affairs/caseflow/issues/5961)

Now that we're introducing the ability to add decision issues to AMA appeals, we want to introduce a new design that will show both contested issues (request issues) from the Veteran's NOD alongside decision issues, that the attorney is deciding. 

## AC
- The banner `Issue` exists above each `request_issue`
- For each issue, we display the issue description, benefit type, diagnostic code, and note.

## Mocks
Issues: No dispositions
Latest updates: 
- Changed "contested issue" header in grey box to "issue"
- Added diagnostic code to each request issue

![image](https://user-images.githubusercontent.com/40216950/50249701-c2dc8180-03ac-11e9-9a78-37bbda72e579.png)

Issues: With dispositions
Latest updates: 
- Changed "contested issue" header in grey box to "issue"
- Added diagnostic code to each request issue
- Added link icon with copy to call out decisions that are connected to more than one request issue

![image](https://user-images.githubusercontent.com/40216950/50249751-e0115000-03ac-11e9-9b0e-270f25cbbd89.png)

Issues: Shows decision with different benefit type
Latest update: 
- Example mock that displays decision issue with different benefit type, for reference

![image](https://user-images.githubusercontent.com/40216950/50249807-08994a00-03ad-11e9-84ba-9716627fb758.png)

allyceh:
@mdbenjam @aroltsch Re-opened, updated mocks + ACs.

PR #7962, opened 2018-11-21
Issue editing frontend

mdbenjam:
Connects: #7957, Connects: #7958, Connects: #7965

### Description
This PR begins adding the editing issue frontend.

1) Update the case details page to show the new AMA issue style.
2) Create a new select disposition page (screenshot below).
3) Build a new modal for users to add decisions.

Next up is to make the modal actually do something. Then show decision issues alongside the request issues.

screen shot 2018-11-26 at 1 31 54 pm

screen shot 2018-11-26 at 1 33 19 pm

### Testing Plan
1. Existing tests pass when feature toggle is off.


client/app/queue/LegacySelectDispositionsView.jsx
```diff
@@ -0,0 +1,210 @@
+import React from 'react';
```
mdbenjam: This is the exact same as the old `SelectDispositionsView.jsx` file.

client/app/queue/AmaSelectDispositionsView.jsx
```diff
@@ -0,0 +1,178 @@
+import React from 'react';
```
mdbenjam: This file is a WIP, but since it's hidden behind a feature toggle, it should be okay to merge.

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -11,7 +11,19 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
...
         description: issue.description,
-        remand_reasons: issue.remand_reasons
+        notes: issue.notes,
+        remand_reasons: issue.remand_reasons,
+        decision_issue_ids: issue.request_decision_issues
+      }
+    end
+  end
+
+  attribute :decision_issues do
```
mdbenjam: We need to start returning decision issues. Though this is not really used in this PR. It will be used in the next one!

db/seeds.rb
```diff
@@ -240,6 +241,9 @@ def create_api_key
   end
 
   def create_ama_appeals
+    description = "Service connection for pain disorder is granted with an evaluation of 70\% effective May 1 2011"
```
mdbenjam: Making these more realistic

app/views/queue/show.html.erb
```diff
@@ -13,7 +13,9 @@
     featureToggles: {
       judge_case_review_checkout: FeatureToggle.enabled?(:judge_case_review_checkout, user: current_user),
       automatic_case_distribution: FeatureToggle.enabled?(:automatic_case_distribution, user: current_user),
-      attorney_assignment_to_colocated: FeatureToggle.enabled?(:attorney_assignment_to_colocated, user: current_user)
+      attorney_assignment_to_colocated: FeatureToggle.enabled?(:attorney_assignment_to_colocated, user: current_user),
+      ama_decision_issues: FeatureToggle.enabled?(:ama_decision_issues, user: current_user)
+
```
amprokop: nitty nit nit

amprokop: ooh interesting are these two files totally identical? should we delete one of them at some point?

client/app/queue/AmaSelectDispositionsView.jsx
amprokop: 👍  — question, are we still trying to avoid "Ama" in names on the frontend? Totally understood if that fell by the wayside.

db/seeds.rb
amprokop: thank you!

db/schema.rb
```diff
@@ -661,12 +661,12 @@
     t.bigint "ineligible_due_to_id"
     t.boolean "untimely_exemption"
     t.text "untimely_exemption_notes"
+    t.string "ineligible_reason"
```
amprokop: question for my own edification — any idea why this happens sometimes?

client/app/queue/AmaSelectDispositionsView.jsx
mdbenjam: Yeah, I try to, but I was sort of at a loss to know what to name it since `SelectDispositionsView.jsx` already exists as the file that forks between the legacy and ama versions. Do you have any naming suggestions here?

db/schema.rb
mdbenjam: I don't know :/ I'll probably just checkout the master version of this file so I don't actually change it.

client/app/queue/AmaSelectDispositionsView.jsx
pkarman: Intake renamed `AMAReview` to `DecisionReview` in order to keep AMA out of names.

amprokop: ah interesting

amprokop: maybe rename the outer SelectDispositionsView into SelectDispositionsContainerView or something? not a huge deal though.

mdbenjam: I like it. Renamed

PR #7964, opened 2018-11-21
Create JudgeTeam class

lowellrex:
Connects #7506.

In preparation for [moving attorney membership in judge teams](https://github.com/department-of-veterans-affairs/caseflow/pull/7847) over to the organization model this PR creates a new `JudgeTeam` class.

### DIFF
--- a/app/models/organizations/judge_team.rb
+++ b/app/models/organizations/judge_team.rb
@@ -4,9 +4,9 @@ class JudgeTeam < Organization
   end
 
   def self.create_for_judge(user)
-    org = create!(name: user.css_id)
-    OrganizationsUser.make_user_admin(user, org)
-    org
+    create!(name: user.css_id).tap do |org|
+      OrganizationsUser.make_user_admin(user, org)
+    end
   end
 
   def can_receive_task?(_task)
--- a/app/models/organizations/judge_team.rb
+++ b/app/models/organizations/judge_team.rb
@@ -8,4 +8,8 @@ class JudgeTeam < Organization
     OrganizationsUser.make_user_admin(user, org)
     org
   end
+
+  def can_receive_task?(_task)
+    false
+  end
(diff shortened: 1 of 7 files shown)


app/models/user.rb
```diff
@@ -209,6 +209,10 @@ def current_case_assignments
     self.class.appeal_repository.load_user_case_assignments_from_vacols(css_id)
   end
 
+  def administrated_teams
```
lowellrex: The name of this function seems a little clumsy to me, but I struggle to think of a better one.

app/models/organizations/judge_team.rb
```diff
@@ -0,0 +1,15 @@
...
+    user.administrated_teams.select { |team| team.is_a?(JudgeTeam) }.first
+  end
+
+  def self.create_for_judge(user)
+    create!(name: user.css_id).tap do |org|
+      OrganizationsUser.make_user_admin(user, org)
+    end
+  end
+
+  def can_receive_task?(_task)
```
mdbenjam: Why is this false?

spec/models/judge_team_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    context "when user is admin of JudgeTeam" do
+      let!(:judge_team) { JudgeTeam.create_for_judge(user) }
+
+      it "should return judge team" do
+        expect(JudgeTeam.for_judge(user)).to eq(judge_team)
+      end
+    end
+
+    # Limitation of the current approach is that users are effectively limited to being the admin of a single JudgeTeam.
```
mdbenjam: I think this is fine

app/models/user.rb
mdbenjam: I think I would either name it: `administered_teams` or `teams_admin_for`

PR #7968, opened 2018-11-23
Issue order in Case Details (#7776)

### LINKED ISSUE #7776: Issue order in Case Details
@nicholasholtz and I observed an attorney who was changing issues on appeal yesterday using the functionality leveraged from the Intake team. When she added two additional VA decisions (issues) to the veteran's appeal, she was actually adding a 7th and 8th issue, but the issues showed up as issue 1 and 2 on Case Details. 

Is it possible for us to add new issues to the end of the issues list so that they reflect the order in which they are selected in the Intake flow? That was the attorney's preference.

rkreyhsig:
Resolves #7776 - Issue order in Case Details


rkreyhsig:
@lowell maybe there is a better way to check the correct order in the test?

app/models/appeal.rb
```diff
@@ -82,7 +82,8 @@ def reviewing_judge_name
   end
 
   def eligible_request_issues
-    request_issues.select(&:eligible?)
+    # Orders by created_at ASC leaving nil values at the end
+    request_issues.select(&:eligible?).sort_by { |issue| [issue.created_at ? 0 : 1, issue] }
```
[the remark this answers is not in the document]
lowellrex: If we are sorting request issues by `created_at` time, would sorting by the table's autoincrementing primary key ID net us the same result? Do we ever expect the `created_at` time order to be different than the order of the table's ID?

db/migrate/20181121222514_add_created_at_timestamp_to_request_issues.rb
```diff
@@ -0,0 +1,5 @@
+class AddCreatedAtTimestampToRequestIssues < ActiveRecord::Migration[5.1]
+  def change
+    add_column :request_issues, :created_at, :datetime
```
lowellrex: One way we could avoid [the nil ordering issue](https://github.com/department-of-veterans-affairs/caseflow/pull/7968/files#r236082447) I raised above would be to populate the `created_at` value for all existing rows after we add this field. I'd personally prefer to leave those values as `null` since I find that an obvious indication that these rows existed before the created_at field was added.

spec/feature/queue/case_details_spec.rb
```diff
@@ -450,5 +450,25 @@
...
+                                            review_request_type: "Appeal")
+        end
+        let!(:request_issue2) do
+          FactoryBot.create(:request_issue, review_request_id: appeal.id, description: issue_description2,
+                                            review_request_type: "Appeal")
+        end
+
+        it "should display sorted issues" do
+          visit "/queue/appeals/#{appeal.uuid}"
+          expect(page).to have_content(issue_description + " Issue 2 DESCRIPTION " + issue_description2)
```
lowellrex: I think this is a fine way to check for ordering. An alternative could be to pluck out the page elements and make sure they are ordered properly ([like we do in this test](https://github.com/department-of-veterans-affairs/caseflow/blob/915dbb3892b8f46d2455aab28d92d24809cdd1ff/spec/feature/queue/task_queue_spec.rb#L73)).

lowellrex: Also, I think we should add a test to make sure sorting works properly when some request issues have `created_at` values and others don't.

lowellrex: No longer needed since we are sorting by ID now.

PR #7970, opened 2018-11-26
Rename Task.action to Task.label on frontend

lowellrex:
Connects #7686. This is the first step in the removal of the `action` column from the `tasks` table. This PR changes all references to the `task.action` attribute to the `task.label` attribute on the frontend.

### DIFF
--- a/app/controllers/tasks_controller.rb
+++ b/app/controllers/tasks_controller.rb
@@ -156,13 +156,17 @@ class TasksController < ApplicationController
 
   def create_params
     @create_params ||= [params.require("tasks")].flatten.map do |task|
-      task = task.permit(:type, :instructions, :action, :assigned_to_id,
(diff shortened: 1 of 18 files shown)


app/controllers/tasks_controller.rb
```diff
@@ -156,14 +156,18 @@ def task
 
   def create_params
     @create_params ||= [params.require("tasks")].flatten.map do |task|
-      task = task.permit(:type, :instructions, :action, :assigned_to_id,
+      task = task.permit(:type, :instructions, :action, :label, :assigned_to_id,
```
amprokop: we're not removing `:action` in this PR  for backwards compatibility yes?

amprokop: ok seen below, resolving

PR #7979, opened 2018-11-26
Add claimant validation when claimant is not veteran

leikkisa:
connects #7595

Require that a claimant is selected if the user selects the option that the claimant is not the veteran.


app/models/claim_review_intake.rb
```diff
@@ -54,16 +61,17 @@ def create_claimant!
     else
       Claimant.create!(
         participant_id: veteran.participant_id,
-        payee_code: need_payee_code? ? "00" : nil,
+        payee_code: nil,
         review_request: detail
       )
     end
+    update_person!
```
leikkisa: On the last PR we moved to Claimant.create! directly, but previously used DecisionReview's create_claimants! which also updates the person.

app/models/appeal_intake.rb
```diff
@@ -8,12 +10,27 @@ def ui_hash(ama_enabled)
...
+        participant_id: claimant_participant_id,
+        payee_code: nil,
+        review_request: detail
+      )
+      update_person!
+      detail.save(context: :intake_review)
+    end
+  rescue ActiveRecord::RecordInvalid => _err
+    # propagate the error from invalid column to the user-visible reason
+    if detail.errors.messages[:veteran_is_not_claimant].include?(ClaimantValidator::CLAIMANT_REQUIRED)
```
pkarman: the logic here is slightly different than in `ClaimReviewIntake` where the `detail.validate` only gets called if there is a `CLAIMANT_REQUIRED` error. Was that intentional?

app/models/appeal_intake.rb
```diff
@@ -8,12 +10,27 @@ def ui_hash(ama_enabled)
...
+      claimant_error = "blank"
```
pkarman: we've started to use this string often enough that it feels like it should be a constant somewhere.

app/models/decision_review.rb
```diff
@@ -82,6 +82,8 @@ def claimant_participant_id
   end
 
   def claimant_not_veteran
+    # This is being replaced by veteran_is_not_claimant, but keeping it temporarily
```
pkarman: +1

spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -8,6 +8,7 @@
     FeatureToggle.enable!(:intake)
     FeatureToggle.enable!(:intakeAma)
     FeatureToggle.enable!(:intake_legacy_opt_in)
+    FeatureToggle.disable!(:intake_enable_add_issues_page)
```
pkarman: are we adding tests to the old intake flow?

spec/models/appeal_intake_spec.rb
```diff
@@ -125,6 +129,18 @@
           payee_code: nil
         )
       end
+
+      context "and claimant is nil" do
```
pkarman: a nit: the "and" here seems to be implied everywhere else `context` is used.

spec/feature/intake/higher_level_review_spec.rb
leikkisa: I modified the existing test that had this section covered (the review page validation).

spec/models/appeal_intake_spec.rb
leikkisa: Will remove!

app/models/appeal_intake.rb
leikkisa: I don't see what you're talking about.

pkarman: I see now that both models have changed in order to escalate the error messages from child object validation into something the view can use. Seems ok for now.

PR #7986, opened 2018-11-26
CO parent records to Caseflow!

### LINKED ISSUE #7989: CO Parent Records
Starting January 1st, we're going to store CO parent records in Caseflow! We need to update update_co_hearing to create child co records in VACOLS.

sharonwarner:
Resolves #7989 

### Description
This PR allows us to create parent records in Caseflow starting January 1st.

### Testing Plan
1. Create a parent record for a CO hearing in Caseflow for January 5th.
2. Go to the View Schedule page and confirm the hearing day is displayed.
3. Go to the Assign Hearings page, schedule a veteran for the parent record, and confirm the record saves successfully, shows up on the daily docket, and shows up on the Scheduled Hearings tab.


app/models/hearing_day.rb
```diff
@@ -3,18 +3,25 @@
...
 class HearingDay < ApplicationRecord
+  belongs_to :judge, class_name: "User"
+
   HEARING_TYPES = {
     video: "V",
     travel: "T",
     central: "C"
   }.freeze
 
   CASEFLOW_SCHEDULE_DATE = Date.new(2019, 3, 31).freeze
```
lomaxap: should this be called `CASEFLOW_V_PARENT_DATE` to be consistent?

lomaxap: may also want to comment that these are the dates where CF handles _saving_ hearings to Vacols and not the date hearings are saved to a CF db.

app/models/hearing_day.rb
```diff
@@ -3,19 +3,28 @@
...
     def create_hearing_day(hearing_hash)
       hearing_date = hearing_hash[:hearing_date]
-      hearing_date = hearing_date.is_a?(DateTime) ? hearing_date : Time.zone.parse(hearing_date).to_datetime
-      if hearing_date > CASEFLOW_SCHEDULE_DATE
+      hearing_date = if hearing_date.is_a?(DateTime) | hearing_date.is_a?(Date)
+                       hearing_date
+                     else
+                       Time.zone.parse(hearing_date).to_datetime
+                     end
+      if hearing_date > CASEFLOW_V_PARENT_DATE
```
oscarAtNava: We are only creating parent records in Caseflow if date is April 1 2019 or later? Thought we were creating them for CO as well after Dec 31.

sharonwarner: Created an issue to fix this.

PR #7988, opened 2018-11-26
Add test data for Quality Review

lowellrex:
Adds two realistic Quality Review tasks, one assigned to `QR_USER` and one assigned to the Quality Review team.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -79,12 +79,12 @@ class SeedDB
   end
 
   def create_qr_user
-    u = User.create!(station_id: 101, css_id: "QR_USER", full_name: "QR User")
-    OrganizationsUser.add_user_to_organization(u, QualityReview.singleton)
+    qr_user = User.create!(station_id: 101, css_id: "QR_USER", full_name: "QR User")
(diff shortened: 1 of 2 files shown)


db/seeds.rb
```diff
@@ -81,6 +81,10 @@ def create_org_queue_users
   def create_qr_user
     u = User.create!(station_id: 101, css_id: "QR_USER", full_name: "QR User")
     OrganizationsUser.add_user_to_organization(u, QualityReview.singleton)
+
+    # Create two QR tasks. One assigned to the organization and one assigned to both the organization and a QR user.
+    create_task_at_quality_review
+    create_task_at_quality_review(u)
```
amprokop: hm 🤔 i thought our linter prohibited single letter variable names. not necessary to change if you prefer it here though

lowellrex: Our javascript linter does not let us use single letter variable names, but rubocop doesn't have a similar constraint. I'm fine with the single letter variable here since we define it 2 lines above and only use that variable once so I don't think it's too hard to figure out what it represents. Happy to change it to something else if you'd prefer.

lowellrex: Twice I guess. I'll go ahead and change it to `qr_user`.

Issue #7772, opened 2018-11-07
Edit page - removing and adding same issue


leikkisa:
On the Edit page, if a user removes the issue, and then re-adds it, it is shown as ineligible because it is already in active review.  However, it's in active review from the same intake.

This should be treated more like it was removed by accident (e.g. never removed at all).

### Acceptance criteria:
- Removing and re-adding an eligible issue on the edit page remains eligible on the UI and in the backend
- An issue that is removed and re-added should not be included in the create_contentions call

leikkisa:
Peter fixed the back-end ineligibility for this situation in https://github.com/department-of-veterans-affairs/caseflow/pull/7769.

The front end presentation still needs to be fixed, most likely in "title_of_active_review".

We also need to check that either both remove and create contention are being called, or neither of them are being called.

pkarman:
> We also need to check that either both remove and create contention are being called, or neither of them are being called.

I think the behaviour we want is to remove/create because the CA may add comments.

shanear:
![bug](https://user-images.githubusercontent.com/1596259/49035443-bc2e6600-f182-11e8-9ec4-a820ade08b04.gif)
Some strange behavior might have to do with this one. 

If an issue was valid at the time it was added, it should continue to be valid on the edit page.

steve-gov:
Validated as working:

![screen](https://user-images.githubusercontent.com/25468689/49970857-7c38e400-fee1-11e8-8e45-d9f6421bea3a.gif)

PR #7993, opened 2018-11-26
Fix React bug with state and props out of sync

pkarman:
connects #7772 

### Description
The refactor in 5270977f8e4cbbea0c2f4f1a5ff5474fcd21b96c introduced a regression due to how React components are rendered. When a list of child components is rendered, and then one is removed, all the child components bump up the stack. That meant that issues might inherit the state of the component just before them, causing them to appear as eligibility errors when they are not.

The solution is to calculate eligibility errors at every `render` instead of assuming that an issue and a component will stay together forever.

### Acceptance Criteria
- [x] Code compiles correctly


client/app/intake/components/AddedIssue.jsx
```diff
@@ -29,7 +29,8 @@ class AddedIssue extends React.PureComponent {
     let errorMsg = '';
 
     if (issue.isUnidentified) {
-      return;
+      return { errorMsg,
+        cssKlasses: ['issue-desc', 'not-eligible'] };
```
pkarman: this needs a feature test for the css class since that is the only thing used on an unidentified issue.

pkarman: added in f13bea7990ad020feaf1f17f8578d9eb2cfa492e

leikkisa:
I don't have the test data setup for checking the timeliness thing (if the timeliness exemption were to change on a request issue), but I suspect it will also be an error.

Here's a different example though. This had an issue that was ineligible before, and if it gets removed and readded, it's no longer shown as ineligible.

edit-modals

pkarman:
that's a great catch @leikkisa ! fixed in c784bcd

I need to write a lot of specific feature / css tests to really firm this up.

Issue #7973, opened 2018-11-26
Reassign cases to other vlj-support staff


laurjpeterson:
When I assigned users to the VLJ support organization (`organizations/vlj-support-staff/users`), I added Rolandus and Jack to the list. However, since they are supervisors, they should not be receiving any cases during the round robin assignment. 

## AC
- Reassign cases that are currently assigned to Rolandus (bvarbranch) and Jack (bvajkenner) to other members of the VLJ support team

lowellrex:
## Additional AC
* Allow VLJ staff to reassign tasks to members of the same team.

lowellrex:
The additional work to allow VLJ support staff to reassign tasks to other members of the same team will be completed in #7999.

PR #7995, opened 2018-11-27
Do not assign ColocatedTasks to admins

lowellrex:
Connects #7973. Prevents `ColocatedTask`s from being assigned to admins on the VLJ support team (`Colocated`).

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -3,6 +3,14 @@ class Organization < ApplicationRecord
   has_many :organizations_users, dependent: :destroy
   has_many :users, through: :organizations_users
 
+  def admins
+    organizations_users.select(&:admin?).map(&:user)
+  end
+
+  def non_admin_users
+    organizations_users.reject(&:admin?).map(&:user)
+  end
+
   def self.assignable(task)
     select { |org| org.can_receive_task?(task) }
   end
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -7,7 +7,7 @@ class Organization < ApplicationRecord
     organizations_users.select(&:admin?).map(&:user)
   end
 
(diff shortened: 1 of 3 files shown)


app/models/organization.rb
```diff
@@ -3,6 +3,14 @@ class Organization < ApplicationRecord
   has_many :organizations_users, dependent: :destroy
   has_many :users, through: :organizations_users
 
+  def admins
+    organizations_users.select(&:admin?).map(&:user)
+  end
+
+  def non_admin_users
+    organizations_users.reject(&:admin?).map(&:user)
```
lowellrex: I played around with using associations to do this in a more rails-y fashion, but couldn't get it to work. Open to ideas for how to make this more rails-y if that's something we want to do.

amprokop: 👍this is good

app/models/organization.rb
```diff
@@ -3,6 +3,14 @@ class Organization < ApplicationRecord
   has_many :organizations_users, dependent: :destroy
   has_many :users, through: :organizations_users
 
+  def admins
```
amprokop: Hm — possible to name these two functions more consistently?

app/models/tasks/colocated_task.rb
```diff
@@ -36,7 +36,7 @@ def create_many_from_params(params_array, user)
     private
 
     def list_of_assignees
-      Colocated.singleton.users.order(:id).pluck(:css_id)
```
amprokop: any reason we changed `order` to `sort_by`?

app/models/organization.rb
lowellrex: I think admins and non_admins makes more sense.

Issue #7965, opened 2018-11-21
Create add decisions page with request issues + decision issues


laurjpeterson:
Part of [Editing AMA issues](https://github.com/department-of-veterans-affairs/caseflow/issues/5961)

After an attorney has added decision issues in the modal, we should display them back on the `Add decisions` page in the checkout flow.

## AC
- Users can see both request issues and the decision issues they are connected to in the `Add decisions` page
- All decision issues also display a disposition (allowed, remanded, etc.)
- If the user adds one decision to many request issues, display the decision under all issues to which it's added. Also display a link icon + "Added to # issues" in each instance of the decision.
- When the user tries to "Continue" without adding a decision, display an error message.
- When the benefit type on the decision issue differs from the benefit type on the request issue, display the update benefit type to the user for reference. Example: Request issue is Compensation, and user adds an Education decision issue. Display "Education" in decision issue. 
  - If the benefit type on decision issue does not differ from the benefit type on the request issue, don't don't display benefit type in the decision issue. (Note: This will be the case 99% of the time.) 

## Mocks

Latest updates: 
- Added diagnostic code field to each request and decision issue
- Added Delete button to the left of Edit button to delete the decision issue

![image](https://user-images.githubusercontent.com/40216950/50249195-50b76d00-03ab-11e9-8b96-ead49381e951.png)

allyceh:
@mdbenjam @aroltsch Re-opened and updated AC + mock.

aroltsch:
Opened a new ticket

Issue #7966, opened 2018-11-21
Edit/remove decision issues


laurjpeterson:
Part of [Editing AMA issues](https://github.com/department-of-veterans-affairs/caseflow/issues/5961)

**Editing decision issues**
- As an attorney, I need the ability to **delete/remove a new decision issue** that I've added by mistake so that I can fix any accidental errors. 
- As an attorney, I need the ability to **edit a new decision issue** that I've created so that I can fix any accidental errors. 

## AC
- An `Edit` and `Delete` link is displayed next to the decision issue on the `Add decisions` page
- When the user clicks `Edit`, a modal titled `Edit decision` pops up. This modal has the same design as the `Add decision` modal. 
  - The existing decision description, disposition, and benefit is pre-populated in the `Edit decision` modal. 
  - The user can edit the description and change the disposition. 
  - When the user clicks Save, they navigate back to the `Add decisions` page and the edited decision issue displays.
- When the user clicks `Delete`, a confirmation modal pops up.
  - If the user confirms that they want to delete the decision, all of the decision issue information (description, disposition, benefit type) is cleared and removed from the Add decisions page. 
  - If the user deletes a decision issue that's connected to more than one request issue, we remove all instances of that decision that are connected to request issues. 

## Mocks

Editing decision for one issue. All other issues have decisions. 
Latest updates (reflected below): 
- Removed delete button
- Added diagnostic code field
- Revised hierarchy of fields

![image](https://user-images.githubusercontent.com/40216950/50248578-79d6fe00-03a9-11e9-85f8-ce2e06633189.png)

Editing decision for one issue. Issues that don't have decisions appear in drop down. 
Latest updates: 
- Removed delete button
- Added diagnostic code field
- Revised hierarchy of fields
- Updated copy and styling of "Do you want to add this decision to other....?" section at the bottom.

![image](https://user-images.githubusercontent.com/40216950/50248671-d5a18700-03a9-11e9-8857-aedd31547199.png)

Editing decision for 2+ issues. All other issues have decisions. 
Latest updates: 
- Removed delete button
- Added diagnostic code field
- Revised hierarchy of fields

![image](https://user-images.githubusercontent.com/40216950/50248801-39c44b00-03aa-11e9-8a89-c9f4222890de.png)

Editing decision for 2+ issues. Issues that don't have decisions appear in drop down.
Latest updates: 
- Removed delete button
- Added diagnostic code field
- Revised hierarchy of fields
[...]

PR #8003, opened 2018-11-27
Show decision issues

mdbenjam:
Connects: #7966, Resolves: #7965, Resolves: #7957, Resolves: #7958

### Description
This change actually makes the add/edit decision issue modal functional. It saves results to the redux store and sends them to the server on completing the flow. It also updates the remand reason select screen to work with these new decision issues.

It also displays decision issues on the case details and disposition selection screen.

### Testing Plan
1. Don't break existing functionality


client/app/queue/SelectDispositionsContainer.jsx
```diff
@@ -8,7 +8,7 @@ class SelectDispositionsContainer extends React.PureComponent {
   render = () => {
     const { appeal, featureToggles, ...otherProps } = this.props;
 
-    if (appeal.isLegacy || !featureToggles.ama_decision_issues) {
+    if (appeal.isLegacyAppeal || !featureToggles.ama_decision_issues) {
```
mdbenjam: This was wrong, but the feature toggle was off so it still worked 😥

client/app/queue/SelectDispositionsView.jsx
```diff
@@ -27,16 +25,24 @@ import {
...
+import uuid from 'uuid';
 
 class SelectDispositionsView extends React.PureComponent {
   constructor(props) {
     super(props);
 
     this.state = {
-      issuesOpen: null
+      openRequestIssueIds: null,
+      decisionIssue: null
```
mdbenjam: Create the decisionIssue in local state and save it to redux state when the issue editing is complete.

client/app/queue/SelectDispositionsView.jsx
```diff
@@ -27,16 +25,24 @@ import {
...
     };
   }
 
+  componentDidMount = () => {
+    if (this.props.userRole === USER_ROLE_TYPES.attorney) {
+      this.props.setDecisionOptions({ work_product: 'Decision' });
```
mdbenjam: I think there has to be a better way to do this... Probably should belong in the submit component, but this mirrors how the legacy disposition component works.

client/app/queue/SelectDispositionsView.jsx
```diff
@@ -74,25 +80,58 @@ class SelectDispositionsView extends React.PureComponent {
     return `/queue/appeals/${appealId}/tasks/${taskId}/${checkoutFlow}/special_issues`;
   }
 
-  addDecisionHandler = (issueIds) => () => {
+  openDecisionHandler = (requestIssueIds, decisionIssue) => () => {
+    const newDecisionIssue = {
+      id: uuid.v4(),
```
mdbenjam: When creating a new decision issue, we need to have an id. It seems easiest to just create a UUID. The real id will be set when it's sent to the server.

client/app/queue/SelectDispositionsView.jsx
```diff
@@ -74,25 +80,58 @@ class SelectDispositionsView extends React.PureComponent {
...
+      description: '',
+      disposition: null,
+      request_issue_ids: requestIssueIds
+    };
+
     this.setState({
-      issuesOpen: issueIds
+      openRequestIssueIds: requestIssueIds,
+      decisionIssue: decisionIssue || newDecisionIssue
```
mdbenjam: If we're editing an issue we use the existing decisionIssue, otherwise we use the new decision issue.

client/app/queue/SelectDispositionsView.jsx
```diff
@@ -74,25 +80,58 @@ class SelectDispositionsView extends React.PureComponent {
...
     this.setState({
-      issuesOpen: null
+      openRequestIssueIds: null,
+      decisionIssue: null
     });
   }
 
+  saveDecision = () => {
+    let decisionIssueFound = false;
+    let newDecisionIssues = this.props.appeal.decisionIssues.map((decisionIssue) => {
```
mdbenjam: This is really clunky. Thinking of a simpler way to do this. We're either appending the new issue, or replacing the existing one.

client/app/queue/SubmitDecisionView.jsx
```diff
@@ -233,7 +237,8 @@ const mapStateToProps = (state, ownProps) => {
     decision,
     error,
     userRole,
-    highlightFormItems
+    highlightFormItems,
+    amaDecisionIssues: state.ui.featureToggles.ama_decision_issues
```
mdbenjam: I think the way we do feature toggles is dumb. They shouldn't be in state, I think they should be global. But for another time.

client/app/queue/components/ContestedIssues.jsx
```diff
@@ -2,48 +2,136 @@ import * as React from 'react';
...
+  margin: `0 ${TEXT_INDENTATION}`
 });
 
 const buttonDiv = css({
   textAlign: 'right',
-  marginBottom: '20px'
+  margin: '20px 0'
+});
+
+const outerDiv = css({
```
mdbenjam: There is a lot of styling here. Diverging a lot from existing style guide components. Don't feel great about it.

client/app/queue/components/IssueRemandReasonsOptions.jsx
```diff
@@ -174,12 +185,16 @@ class IssueRemandReasonsOptions extends React.PureComponent<Params, State> {
...
-    [event.target.id.split('-')[1]]: {
-      checked,
-      post_aoj: null
-    }
-  });
+  toggleRemandReason = (checked, event) => {
+    const splitId = event.target.id.split('-');
+
+    this.setState({
+      [splitId[splitId.length - 1]]: {
```
mdbenjam: The new issue ids have dashes in them since they're UUIDs. We want to get the last segment of the id. Don't love this approach, probably should just use a closure.

client/app/queue/SubmitDecisionView.jsx
amprokop: Agreed, that'd be better.

client/app/queue/components/IssueRemandReasonsOptions.jsx
```diff
@@ -322,21 +337,22 @@ class IssueRemandReasonsOptions extends React.PureComponent<Params, State> {
 
 const mapStateToProps = (state, ownProps) => {
   const appeal = state.queue.stagedChanges.appeals[ownProps.appealId];
-  const issues = appeal.issues;
+  const issues = (state.ui.featureToggles.ama_decision_issues && !appeal.isLegacy) ?
```
amprokop: `isLegacy` or `isLegacyAppeal`?

client/app/queue/SelectRemandReasonsView.jsx
```diff
@@ -108,7 +108,8 @@ SelectRemandReasonsView.propTypes = {
 
 const mapStateToProps = (state, ownProps) => {
   const appeal = state.queue.stagedChanges.appeals[ownProps.appealId];
-  const issues = appeal.issues;
+  const issues = (state.ui.featureToggles.ama_decision_issues && !appeal.isLegacy) ?
```
amprokop: `isLegacy` or `isLegacyAppeal`?

client/app/queue/components/ContestedIssues.jsx
amprokop: I feel your pain. Potentially this is something to raise with design, it looks necessary at the moment though.

client/app/queue/components/IssueRemandReasonsOptions.jsx
amprokop: hm why do you need only the last segment of the id? I suppose it's still unique enough to be fine for all practical purposes. Not sure what the restriction is here that necessitates it.

client/app/queue/components/IssueRemandReasonsOptions.jsx
mdbenjam: isLegacyAppeal 🤦‍♂️

client/app/queue/components/IssueRemandReasonsOptions.jsx
mdbenjam: I *think* (not super familiar with this component) that they're using that section of the id as what is sent back as a key to the server.

client/app/queue/SelectDispositionsView.jsx
mdbenjam: Prepending `temporary-id-` so that folks know it's temporary.

client/app/queue/SelectDispositionsContainer.jsx
aroltsch: Uh oh

client/app/queue/components/IssueRemandReasonsOptions.jsx
```diff
@@ -96,11 +95,23 @@ class IssueRemandReasonsOptions extends React.PureComponent<Params, State> {
...
+    const { appeal, appealId, issueId, amaDecisionIssues } = this.props;
+    const useDecisionIssues = !appeal.isLegacyAppeal && amaDecisionIssues;
+    const issues = useDecisionIssues ? appeal.decisionIssues : appeal.issues;
 
-    this.props.startEditingAppealIssue(appealId, issueId, attributes);
-    this.props.saveEditedAppealIssue(appealId);
+    const updatedIssues = issues.map((issue) => {
+      if (issue.id === issueId) {
+        return { ...issue,
+          remandReasons };
```
mdbenjam: gah, this should be `remand_reasons: remandReasons` fixing in a follow up.

PR #8011, opened 2018-11-28
Make RootTask inherit from GenericTask

lowellrex:
Working on a broader goal of cleaning up the Task model (ticket coming soon!). This seemed like a nice contained place to start.


app/models/tasks/root_task.rb
```diff
@@ -7,22 +7,17 @@ def set_assignee
...
-
-    []
+  def available_actions(_user)
+    [Constants.TASK_ACTIONS.CREATE_MAIL_TASK.to_h]
   end
 
   def no_actions_available?(user)
-    return false if status == Constants.TASK_STATUSES.on_hold && MailTeam.singleton.user_has_access?(user)
-    super
+    !(MailTeam.singleton.user_has_access?(user) && status != Constants.TASK_STATUSES.completed)
```
lowellrex: I find the logic in these `no_actions_available?` methods a little confusing and think it is because I am having trouble reasoning through the double negatives that happen in `available_actions_unwrapper`. I am considering making a future ticket that changes the name of these functions to `actions_available?` (or something similar) and accordingly reverses the result.

lowellrex:

app/models/tasks/generic_task.rb
```diff
@@ -75,7 +73,7 @@ def reassign(reassign_params, current_user)
   end
 
   def can_be_accessed_by_user?(user)
-    !available_actions(user).empty?
+    available_actions_unwrapper(user).any?
```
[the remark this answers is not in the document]
mdbenjam: nit: I think `empty?` is better than `any?` since `any?` implies were checking the array for truthy values.

app/models/tasks/root_task.rb
lowellrex: Ticketed my note about `ColocatedTask`s in #8022.

PR #8012, opened 2018-11-28
Enable navigation between personal task queues & organizational queues

### LINKED ISSUE #7904: Enable navigation between different queues
Implement navigation between personal task queues and organizational queues for folks who have access to an organizational queue, and between judge and attorney queues for attorneys temporarily acting as judges. This design was originally proposed in #5385 and affirmed in #7503.

## Acceptance criteria:
* Display a dropdown that switches active queues for folks with access to an organizational queue.
* Display a dropdown that switches between judge and attorney queues for folks acting as judges.

image

_Note: Let's also explore how it looks--and if it is technically easier--to move the dropdown down into the white container instead of leaving it above that container._

joeyyang:
Partially resolves #7904 

### Description
Implement navigation between personal task queues and organizational queues for users who have access to an organizational queue. Per @lowellrex guidance, this PR does not implement switching between judge and attorney queues for attorneys temporarily acting as judges — this will be addressed in a separate PR. 

Questions:
- @sneha-pai: 
    - I was wondering if there were any better copy options for the dropdown — perhaps "Switch queues"?
- @lowellrex: 
    - I'm using the `Link`'s `href` property to change routes and the performance is a bit slow — does this use React Router / if not, is there a way to handle the routing in React Router and show the loading screen immediately while loading the next route?
    - I was considering stripping out the `organizationIds` from the codebase — it's only used in one place, and I figured we could just send the entire `organization` payload and get the IDs from that. If you agree, I can address this in a followup PR. 
    - Any test coverage I'm missing?

Screenshot:
screenshot 2018-11-27 16 29 25

Screencap:
nov-27-2018 16-33-56


client/app/queue/components/QueueSelectorDropdown.jsx
```diff
@@ -0,0 +1,98 @@
...
+      return <ul className="cf-dropdown-menu active" {...styles.dropdownList}>
+        <li key={0}>
+          <Link className="usa-button-secondary usa-button"
+            href={queueHref} onClick={this.onMenuClick}>
+            {COPY.CASE_LIST_TABLE_QUEUE_DROPDOWN_OWN_CASES_LABEL}
+          </Link>
+        </li>
+
+        {orgs.map((org, index) => {
+          const orgHref = (location === org.url) ? '#' : `/organizations/${org.url}`;
```
lowellrex: Love this! So nice!

client/app/queue/components/QueueSelectorDropdown.jsx
```diff
@@ -0,0 +1,98 @@
...
+              {sprintf(COPY.CASE_LIST_TABLE_QUEUE_DROPDOWN_TEAM_CASES_LABEL, org.name)}
+            </Link>
+          </li>;
+        })}
+      </ul>;
+    };
+
+    if (organizations.length < 1) {
+      return null;
+    }
```
lowellrex: If we move this condition to the top of the function, we can inline the `dropdownButtonList()` function which I think simplifies this component.

app/views/organizations/show.html.erb
```diff
@@ -5,6 +5,7 @@
     userRole: (current_user.vacols_roles.first || "").capitalize,
     userCssId: current_user.css_id,
     organizationIds: current_user.organizations.pluck(:id),
+    organizations: current_user.organizations,
```
lowellrex: Let's only pass the fields of the organizations we definitely care about here. Perhaps something like:
```ruby
organizations: current_user.organizations.map { |o| o.slice(:id, :name, :url) },
```

This is probably over cautious, but I'd rather not expose the BGS participant ID of an organization if we can avoid it (and we don't need the organization's role on the frontend at the moment either).

client/app/queue/QueueApp.jsx
```diff
@@ -85,6 +88,7 @@ class QueueApp extends React.PureComponent<Props> {
     this.props.setUserRole(this.props.userRole);
     this.props.setUserCssId(this.props.userCssId);
     this.props.setOrganizationIds(this.props.organizationIds);
+    this.props.setOrganizations(this.props.organizations);
```
lowellrex: Not totally related to this PR, but let's go ahead and rip out everything related to `organizationIds`. It looks like we don't use that attribute anywhere in code anymore, and if we do need it we can grab it from `organizations`.

joeyyang: Ok cool — We do use it to fetch some data somewhere, so I'd like to do this in a follow-on PR if that's alright.

PR #8013, opened 2018-11-28
Move access check inside Task.update_from_params

lowellrex:
[`GenericTask.update_from_params`](https://github.com/department-of-veterans-affairs/caseflow/blob/f64005a530c0406d8c29492350fc12ae17ac6fdc/app/models/tasks/generic_task.rb#L51) and [`ScheduleHearingTask.update_from_params`](https://github.com/department-of-veterans-affairs/caseflow/blob/f64005a530c0406d8c29492350fc12ae17ac6fdc/app/models/tasks/schedule_hearing_task.rb#L36) both already verify that the current user can update the selected task. This PR adds that check to `Task.update_from_params`.

This PR also removes the redirect to `/unauthorized` in `TasksController.update` since we will be returning an error if `can_be_accessed_by_user?` returns false (by way of [`verify_user_access!`](https://github.com/department-of-veterans-affairs/caseflow/blob/f64005a530c0406d8c29492350fc12ae17ac6fdc/app/models/task.rb#L145)).


lowellrex:

mdbenjam:
I'm late on this but LGTM too.

PR #8014, opened 2018-11-28
Move update_status from GenericTask to Task

lowellrex:
Noticed this optimization while poking around in the Task models and thought it was worth making on its own.


lowellrex:

lowellrex:
In the absence of a team standard around using dirty methods and active record hooks, I prefer calling a function directly that does the work we want to. I actually removed the `after_update` hook from `Task` last week when we fixed a particularly slippery bug that evaded us for a little while. I outlined some thoughts on the general topic there:

> This change eliminates the duplicate creation of the `BvaDispatchTask` by removing the `after_update` hook. Alternatively we could remove the call to `parent.mark_as_complete!` in the `mark_as_complete!` function. I prefer the approach implemented here because 1) we already call `mark_as_complete!` almost everywhere we change a task's status to complete (cleaned up one place we didn't in this PR), 2) I find active record callbacks indirect and non-obvious which can lead to bugs (like this one, for example), and most importantly in my mind 3) the active record callback prevents us from logging into a rails console and changing the status of a task to complete without running the `after_update` callback which we might not always want to do (I can imagine a scenario where an organization-level task is completed, but the individual's task is not so we would want to change that status to `completed` but not re-run `mark_as_complete!` on the parent task).

I'm opening to establishing standards around this type of thing and doing these type of manipulations by way of hooks or dirty methods if that's what we commit to though.

Issue #8015, opened 2018-11-28
Simplify Task.update_from_params authorization logic


lowellrex:
We have two methods for authorizing operations on the `Task` model. [`verify_user_can_assign!`](https://github.com/department-of-veterans-affairs/caseflow/blob/f64005a530c0406d8c29492350fc12ae17ac6fdc/app/models/task.rb#L151) is run before creating a new task and verifies that whoever is attempting to create the new task is allowed to. [`verify_user_access!`](https://github.com/department-of-veterans-affairs/caseflow/blob/f64005a530c0406d8c29492350fc12ae17ac6fdc/app/models/task.rb#L145) is run before updating a task in [`GenericTask`](https://github.com/department-of-veterans-affairs/caseflow/blob/f64005a530c0406d8c29492350fc12ae17ac6fdc/app/models/tasks/generic_task.rb#L51), [`ScheduleHearingTask`](https://github.com/department-of-veterans-affairs/caseflow/blob/f64005a530c0406d8c29492350fc12ae17ac6fdc/app/models/tasks/schedule_hearing_task.rb#L36), and soon [`Task`](https://github.com/department-of-veterans-affairs/caseflow/pull/8013) (additionally `verify_user_access!` is called in [`GenericTask.create_from_params`](https://github.com/department-of-veterans-affairs/caseflow/blob/f64005a530c0406d8c29492350fc12ae17ac6fdc/app/models/tasks/generic_task.rb#L100) on the parent of the new task, and that parent task will have its status updated, so `verify_user_access!` is effectively checking whether we can update a task there as well). Both of these functions should be named more precisely, but `verify_user_access!` could use additional finessing.

The conditional logic in `verify_user_access!` is implemented by the function `can_be_accessed_by_user?` (which should similarly be renamed). Since we only expect tasks to be updated through the task actions exposed on the frontend, we can use the presence of `available_actions` for a given user to determine whether a task can_be_updated_by_user (tasks can also be updated by the application without direct user manipulation--closing parent organization tasks, for example--but those updates do not run through `update_from_params` so they will not be affected by this change). [GenericTask already does this](https://github.com/department-of-veterans-affairs/caseflow/blob/f64005a530c0406d8c29492350fc12ae17ac6fdc/app/models/tasks/generic_task.rb#L77) so we can follow that example as we move this logic into the base `Task` model.

## Acceptance criteria:
* [x] Rename `verify_user_can_assign!` -> `verify_user_can_create!`
[...]

PR #8018, opened 2018-11-28
Remove extra code from ScheduleHearingTask

lowellrex:
I've been working on cleaning up the `Task` model in #8015 and noticed that we duplicate some code here. This PR removes the excess code.


app/models/tasks/schedule_hearing_task.rb
```diff
@@ -15,22 +10,6 @@ def create_from_params(params, current_user)
...
-
-      create!(
-        appeal: parent.appeal,
-        appeal_type: LegacyAppeal.name,
-        assigned_by_id: child_assigned_by_id(parent, current_user),
-        parent_id: parent.id,
-        assigned_to: assignee,
-        instructions: params[:instructions]
-      )
-    end
```
lowellrex: This function is identical to [`GenericTask.create_child_task`](https://github.com/department-of-veterans-affairs/caseflow/blob/f64005a530c0406d8c29492350fc12ae17ac6fdc/app/models/tasks/generic_task.rb#L108) so we can just remove it and rely on ruby calling the inherited class' method. This allows us to make changes to the base class method and get the benefit of those changes here as well.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -1,12 +1,7 @@
 class ScheduleHearingTask < GenericTask
   class << self
     def create_from_params(params, current_user)
-      root_task = RootTask.find_by(appeal_id: params[:appeal].id)
-      if !root_task
-        root_task = RootTask.create!(appeal_id: params[:appeal].id,
-                                     appeal_type: LegacyAppeal.name,
-                                     assigned_to_id: current_user.id)
-      end
+      root_task = RootTask.find_or_create_by!(appeal: params[:appeal])
```
oscarAtNava: Will the find_or_create_by set the assigned_to attribute? Is it necessary to be set at this level?

lowellrex: It is not necessary to set the `assigned_to` attribute here because [`RootTask` overrides it anyway](https://github.com/department-of-veterans-affairs/caseflow/blob/490fd9e41c9107177ce0d29dc0bb0e39b3ac6ba5/app/models/tasks/root_task.rb#L2) and sets it to `Bva.singleton`.

Issue #7337, opened 2018-10-10
Legacy Opt-in | Add Issue landing and conf | if eligible


suzchap:
## Story
[Mural showing all opt-in screens](https://app.mural.co/t/workqueue2001/m/workqueue2001/1538676663322/098da798af658862105ea8d4d5aeb03fa7aaf99a)

When a user has a legacy appeal, they will be allowed to "opt-in" to move one or more of those issues into an AMA lane. The AMA paper forms will have a question on them for the applicant to acknowledge that by checking that box, they understand that any legacy issues listed on that form will be moved out of the legacy lane and into the new AMA lane.

This ticket is focused on what happens if the answer to the "Did they agree to withdraw their issues from the legacy system?" question on the Review page = YES (which means we're considering this issue eligible). 

requires #7335 
requires #7300
requires #7309
requires https://github.com/department-of-veterans-affairs/appeals-deployment/issues/1712

## Technical notes
- We already have a lot of logic to close VACOLS issues (look at `RampClosedAppeal#close!`)
- The legacy opt-in flow is already behind a feature flag

## Acceptance criteria
- When we intake any AMA lane, when adding an issue, if a legacy issue is selected (#7335):
    - if the opt-in checkbox is "yes" and the legacy issue selected is eligible (see #7334 for definition of eligible), display the issue as eligible on the add issues page, noting that the VACOLS issue will be closed (see copy below)
    - eligible issues with a legacy issue selected ignore timeliness eligibility checks as described in #7300 and #7309 
- When we complete an intake in any AMA lane, if there are eligible issues with a legacy issue selected, the legacy issue id (bfkey) should be stored in the DB.
    - during asynchronous processing of this intake completion, close the legacy issue in VACOLS with a disposition code of "O"
    - if that issue was the last issue on the VACOLS appeal, also close the VACOLS appeal with a disposition code of "O"
    - The contention created should have a "VACOLS Opt-in" special issue issue on it. This will use the same mechanism as the same station special issue (see: https://github.com/department-of-veterans-affairs/caseflow/issues/5281). A service request to create the special issue is being tracked in https://github.com/department-of-veterans-affairs/appeals-deployment/issues/1712. 
- Any closed VACOLS issues should be shown on the confirmation page using the design below

## Notes
- The VACOLS disposition "O" should be added which stands for "AMA Opt-in"
[...]

PR #8027, opened 2018-11-28
Legacy issue optin - update VACOLS

pkarman:
connects #7337 

### Description
When a `RequestIssue.vacols_id` is not nil, create a `LegacyOptinProcessJob` and add a special issue to the created contention.

### Acceptance Criteria
- [x] Code compiles correctly

### DIFF
--- a/app/jobs/legacy_optin_process_job.rb
+++ b/app/jobs/legacy_optin_process_job.rb
@@ -1,0 +1,17 @@
+# This job will call perform! on a LegacyIssueOptin
+class LegacyOptinProcessJob < CaseflowJob
+  queue_as :low_priority
+  application_attr :intake
+
+  def perform(legacy_optin)
+    RequestStore.store[:application] = "intake"
+    RequestStore.store[:current_user] = User.system_user
+
+    begin
+      legacy_optin.perform!
+    rescue BadThing => err # TODO: define exceptions
+      legacy_optin.update_error!(err.to_s)
+      Raven.capture_exception(err)
+    end
+  end
+end
--- a/app/jobs/legacy_optin_process_job.rb
+++ b/app/jobs/legacy_optin_process_job.rb
@@ -9,7 +9,7 @@ class LegacyOptinProcessJob < CaseflowJob
 
     begin
       legacy_optin.perform!
-    rescue BadThing => err # TODO: define exceptions
+    rescue StandardError => err # TODO: define exceptions
       legacy_optin.update_error!(err.to_s)
       Raven.capture_exception(err)
     end
--- a/app/models/concerns/legacy_optinable.rb
+++ b/app/models/concerns/legacy_optinable.rb
@@ -3,7 +3,7 @@ module LegacyOptinable
 
   def create_legacy_issue_optin(request_issue)
     legacy_optin = LegacyIssueOptin.create!(request_issue: request_issue).tap(&:submit_for_processing!)
-    if run_async?
+    if LegacyIssueOptin.run_async?
       LegacyOptinProcessJob.perform_later(legacy_optin)
     else
       LegacyOptinProcessJob.perform_now(legacy_optin)
--- a/app/models/concerns/legacy_optinable.rb
+++ b/app/models/concerns/legacy_optinable.rb
@@ -1,6 +1,8 @@
 module LegacyOptinable
   extend ActiveSupport::Concern
 
+  private
+
   def create_legacy_issue_optin(request_issue)
     legacy_optin = LegacyIssueOptin.create!(request_issue: request_issue).tap(&:submit_for_processing!)
     if LegacyIssueOptin.run_async?
--- a/app/models/concerns/legacy_optinable.rb
+++ b/app/models/concerns/legacy_optinable.rb
@@ -2,11 +2,19 @@ module LegacyOptinable
   extend ActiveSupport::Concern
 
   def create_legacy_issue_optin(request_issue)
-    legacy_optin = LegacyIssueOptin.create!(request_issue: request_issue)
+    legacy_optin = LegacyIssueOptin.create!(request_issue: request_issue).tap(&:submit_for_processing!)
     if run_async?
       LegacyOptinProcessJob.perform_later(legacy_optin)
     else
       LegacyOptinProcessJob.perform_now(legacy_optin)
     end
+  end
+
+  def vacols_optin_special_issue
+    { code: "VO", narrative: "VACOLS Opt-in" }
+  end
+
+  def needs_vacols_optin_special_issue?
+    request_issues.any?(&:vacols_id)
   end
 end
--- a/app/models/concerns/legacy_optinable.rb
+++ b/app/models/concerns/legacy_optinable.rb
@@ -1,0 +1,12 @@
+module LegacyOptinable
+  extend ActiveSupport::Concern
+
+  def create_legacy_issue_optin(request_issue)
+    legacy_optin = LegacyIssueOptin.create!(request_issue: request_issue)
+    if run_async?
+      LegacyOptinProcessJob.perform_later(legacy_optin)
+    else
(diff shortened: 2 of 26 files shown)


client/constants/VACOLS_DISPOSITIONS_BY_ID.json
```diff
@@ -14,6 +14,7 @@
     "G": "Advance Failure to Respond",
     "L": "Manlincon Remand",
     "M": "Merged Appeal",
+    "O": "AMA Opt-in",
```
amprokop: This looks good. FYI the first 7 entries in this file will appear in the dropdown that attorneys use during case review (SelectIssueDispositionDropdown.jsx). But this one won't, which is I think what we want.

app/jobs/legacy_optin_process_job.rb
```diff
@@ -0,0 +1,17 @@
+# This job will call perform! on a LegacyIssueOptin
+class LegacyOptinProcessJob < CaseflowJob
+  queue_as :low_priority
+  application_attr :intake
+
+  def perform(legacy_optin)
+    RequestStore.store[:application] = "intake"
+    RequestStore.store[:current_user] = User.system_user
```
jiahuang: Would this cause a problem if another job were to be executed on this thread which relied on request store? Guess those jobs should also set the request store to be what their methods need before execution.

app/models/issue.rb
```diff
@@ -202,8 +202,23 @@ def remand_reasons
     @remand_reasons ||= self.class.remand_repository.load_remands_from_vacols(id, vacols_sequence_id)
   end
 
+  # For status (BFMPRO) of ADV or REM, for the most part, having a disposition means the issue is closed.
+  # On appeal where the status is REM (remanded) the issues with disposition "3" are still active.
```
jiahuang: 🤔 🤷‍♀️

app/models/legacy_issue_optin.rb
```diff
@@ -2,4 +2,43 @@ class LegacyIssueOptin < ApplicationRecord
...
+      appeals: [legacy_appeal],
+      user: RequestStore.store[:current_user],
+      closed_on: Time.zone.today,
+      disposition: Constants::VACOLS_DISPOSITIONS_BY_ID[VACOLS_DISPOSITION_CODE]
+    )
+  end
+
+  def legacy_appeal_needs_closing?
+    # if all the issues are closed, the appeal should be closed.
+    closed_issues = legacy_appeal.issues.select(&:closed?)
```
jiahuang: I'm a 1/5 on a scale for caring about this, but I think `legacy_appeal.issues.reject(&:closed?).is_empty?` would be clearer

app/models/concerns/legacy_optinable.rb
```diff
@@ -0,0 +1,20 @@
+module LegacyOptinable
```
shanear: Lol this name. So rails.

app/models/issue.rb
```diff
@@ -202,8 +202,23 @@ def remand_reasons
...
+  # rubocop:disable Metrics/CyclomaticComplexity
+  def closed?
```
jiahuang: this method is oddly similar but not quite what's on app/models/vacols/case_issue.rb. Are they linked in some way? Should there be a comment here saying to also update case_issue if this gets updated?

app/jobs/legacy_optin_process_job.rb
pkarman: yes, that seems to be the pattern, to set the `RequestStore` up themselves.

app/models/issue.rb
pkarman: see discussion in https://github.com/department-of-veterans-affairs/caseflow/issues/7337#issuecomment-442643381

app/models/decision_review.rb
```diff
@@ -117,6 +118,12 @@ def serialized_legacy_appeals
     end
   end
 
+  def special_issues
+    specials = []
+    specials << vacols_optin_special_issue if needs_vacols_optin_special_issue?
+    specials
```
shanear: this is a prime use for tap, to avoid awkward mutation code.

```
[].tap do |specials|
  specials << vacols_optin_special_issue if needs_vacols_optin_special_issue?
end
```

app/models/legacy_issue_optin.rb
```diff
@@ -2,4 +2,43 @@ class LegacyIssueOptin < ApplicationRecord
...
+
+  VACOLS_DISPOSITION_CODE = "O".freeze # oh not zero
+
+  def perform!
+    attempted!
+    close_legacy_issue_in_vacols
+    close_legacy_appeal_in_vacols if legacy_appeal_needs_closing?
+    clear_error!
+    processed!
+  end
```
shanear: This is awesome. Very easy to read.

app/models/issue.rb
pkarman: good call. I will add comment to xref them.

they are related insofar as an `Issue` is the Caseflow shadow of a `VACOLS::CaseIssue`, just as a `LegacyAppeal` is the Caseflow shadow to a `VACOLS::Case`. The logic should be identical.

app/models/vacols/case_issue.rb
```diff
@@ -141,6 +143,15 @@ def destroy
...
   end
 
+  # For status (BFMPRO) of ADV or REM, for the most part, having a disposition means the issue is closed.
+  # On appeal where the status is REM (remanded) the issues with disposition "3" are still active.
+  def closed?
+    return false if issdc.nil?
+    return false if issdc == "3" && vacols_case.remanded?
+    return true if vacols_case.status_advanced_or_remanded_or_completed?
+    false
+  end
```
shanear: It's weird that we have to duplicate such a complex piece of logic between here and `issue.rb`. Is there any way we can factor one away?

app/models/concerns/legacy_optinable.rb
```diff
@@ -0,0 +1,20 @@
...
+  extend ActiveSupport::Concern
+
+  def create_legacy_issue_optin(request_issue)
```
shanear: these methods seem like they could all be private.

app/models/decision_review.rb
```diff
@@ -117,6 +118,12 @@ def serialized_legacy_appeals
     end
   end
 
+  def special_issues
```
shanear: since this is a public method, it could probably use a unit test

app/models/vacols/case_issue.rb
shanear: Additionally, this method is not tested. I also don't really see where this is used in the code? We may be able to just delete it?

app/models/legacy_issue_optin.rb
```diff
@@ -2,4 +2,43 @@ class LegacyIssueOptin < ApplicationRecord
   include Asyncable
 
   belongs_to :request_issue
+
+  VACOLS_DISPOSITION_CODE = "O".freeze # oh not zero
```
shanear: 0.O

nice catch on the confusion this could cause

spec/models/legacy_issue_optin_spec.rb
```diff
@@ -55,4 +55,47 @@
...
+
+    subject { create(:legacy_issue_optin, request_issue: request_issue) }
+
+    it "closes VACOLS issue with disposition O" do
+      subject.perform!
+
+      vacols_case_issue.reload
+      vacols_case.reload
+      expect(vacols_case_issue.issdc).to eq(LegacyIssueOptin::VACOLS_DISPOSITION_CODE)
+      expect(vacols_case_issue).to be_closed
```
pkarman: this is where `VACOLS::CaseIssue` `closed?` method is tested.

app/models/issue.rb
pkarman: after talking with @shanear just deleted it from CaseIssue

PR #8030, opened 2018-11-29
Remove organizationIds from codebase

joeyyang:
Follow-on PR to https://github.com/department-of-veterans-affairs/caseflow/pull/8012 — removes references to `organizationIds` from codebase.

### DIFF
--- a/client/app/queue/reducers.js
+++ b/client/app/queue/reducers.js
@@ -45,7 +45,7 @@ export const initialState = {
   pendingDistribution: null,
   attorneys: {},
   organizationId: null,
-  organizations: null,
+  organizations: [],
   loadingAppealDetail: {}
 };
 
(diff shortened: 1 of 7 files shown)


client/app/queue/reducers.js
```diff
@@ -45,6 +45,7 @@ export const initialState = {
   pendingDistribution: null,
   attorneys: {},
   organizationId: null,
```
lowellrex: Since we're already cleaning up the code with this PR, I think we can remove `organizationId` here and [in `queue/types/state.js` as well](https://github.com/department-of-veterans-affairs/caseflow/blob/dd30675c4ee4458acbb7ca077b57375c5c82a20a/client/app/queue/types/state.js#L92).

PR #7785, opened 2018-11-08
Create colocated tasks assigned to team and individual

lowellrex:
Resolves #7528. Creates a `ColocatedTask` assigned to the Colocated team as the parent of every ColocatedTask that is automatically assigned to a member of the Colocated team via round robin distribution. This PR also moves from membership in the colocated team pilot from code to the database (I added the two existing Colocated pilot folks to the `organizations_users` table in the production database).

PR #8053, opened 2018-11-29
Fix bug that prevented VLJ support staff from taking any actions on tasks

lowellrex:
This PR fixes a bug that has prevented members of the VLJ support team from taking any action tasks assigned to them since we added the `Colocated` organization and started assigning individual tasks as [children of the organization's task](https://github.com/department-of-veterans-affairs/caseflow/pull/7785) (roughly 3 weeks). This was also reported by one of the folks who piloted VLJ support staff queue and really puzzled me at the time.

Relevant slack thread: https://dsva.slack.com/archives/CCPHFTQTH/p1543521756002300
> This is caused by a recent change to how we organize members of the VLJ support staff and assign tasks to them. They are all now part of the VLJ support staff team, and when we create a `ColocatedTask` we assign that task to the individual as a child of another task assigned to the VLJ support staff (so their can be team-wide visibility into all of the tasks that are being worked on at the moment). This bug is caused because we are trying to act on the **organization's** task instead of the **individual's** task. Working on a fix now

### DIFF
--- a/app/models/tasks/colocated_task.rb
+++ b/app/models/tasks/colocated_task.rb
@@ -40,9 +40,7 @@ class ColocatedTask < Task
     end
   end
 
-  def available_actions(user)
-    return [] unless user.colocated_in_vacols?
-
+  def available_actions(_user)
     actions = [
       {
         label: COPY::COLOCATED_ACTION_PLACE_HOLD,
@@ -65,8 +63,8 @@ class ColocatedTask < Task
     actions
   end
 
-  def no_actions_available?(_user)
-    completed?
+  def no_actions_available?(user)
+    completed? || assigned_to != user
   end
 
   def update_if_hold_expired!
--- a/spec/feature/queue/task_queue_spec.rb
+++ b/spec/feature/queue/task_queue_spec.rb
@@ -221,4 +221,36 @@ RSpec.feature "Task queue" do
       expect(page).to have_content(COPY::CASE_LIST_TABLE_QUEUE_DROPDOWN_LABEL)
     end
   end
+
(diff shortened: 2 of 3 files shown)


app/models/tasks/colocated_task.rb
```diff
@@ -65,8 +63,8 @@ def available_actions(user)
     actions
   end
 
-  def no_actions_available?(_user)
-    completed?
+  def no_actions_available?(user)
+    completed? || assigned_to != user
```
lowellrex: Previously, any member of the VLJ support staff could take action on any `ColocatedTask`. This change makes it so that actions are only available to the person who is assigned the task. I think this is correct because members of the VLJ support staff should only be taking action on tasks assigned to themselves (for now). This solves the problem because now VLJ support staff will not see the organization tasks and will instead only see the individually assigned task, which they can take action on (unlike the parent organization task).

spec/feature/queue/task_queue_spec.rb
```diff
@@ -221,4 +221,36 @@
...
+      end
+
+      it "should be actionable" do
+        visit("/queue/appeals/#{appeal.external_id}")
+
+        find(".Select-control", text: "Select an action…").click
+        find("div", class: "Select-option", text: COPY::COLOCATED_ACTION_SEND_BACK_TO_ATTORNEY).click
+        find("button", text: COPY::MARK_TASK_COMPLETE_BUTTON).click
+
+        expect(page).to have_content(format(COPY::MARK_TASK_COMPLETE_CONFIRMATION, appeal.veteran_full_name))
```
lowellrex: Marking this task complete fails under the old code.

Issue #8050, opened 2018-11-29
Rescue BGS fetch rating errors


leikkisa:
## Story
If a user gets a Locked Rating or Converted or Backfilled rating error from BGS, the user can potentially get stuck in a state where they have an intake in progress, and they can't cancel it because they are stuck on the app-wide "Something went wrong" page.  This locks them out of the Intake app until the problem intake expires, after a day.

## Acceptance Criteria
- If a user tries to intake a veteran that has a Converted or Backfilled rating error, the intake gets canceled and the user is able to start a new intake
- The user does not get stuck on the "Something went wrong" page

steve-gov:
@kencheeto Can you validate this plz?

steve-gov:
User validated in the field, will keep an eye on it.

Issue #7385, opened 2018-10-12
AMA Intake | Add better error message on intake when a user has a locked rating


shanear:
It might be a little while before https://github.com/department-of-veterans-affairs/caseflow/issues/6221 gets fixed. In the mean time we may want to show a user friendly message on what to do.

kencheeto:
an example error in sentry: https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2918/events/

leikkisa:
Closing this, because the user will no longer see an error when they have a locked rating.

leikkisa:
I am re-opening this, because we may want to let the user know that there was a locked rating (even if there were regular promulgated ratings), because maybe those do have issues that can be intaken.

related https://github.com/department-of-veterans-affairs/caseflow/issues/9817

pkarman:
e.g. https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3976/events/270463/

pkarman:
@leikkisa This may be moot now but I am too far removed to know anymore.

leikkisa:
@pkarman I'd be down for closing this (I see that I re-opened it at some point lol).  I don't think we have a good way to know if there are locked ratings now, unless the only rating is locked, so that would make this hard.

Issue #7054, opened 2018-09-19
Investigate cause of "Converted or Backfilled rating" error from BGS


leikkisa:
Handle the situation where we get the below response when looking for timely_ratings from BGS.

```{:error_level=>"0", :any_ratings_within_range_flag=>"N", :differences_found=>"N", :multiple_ratings_flag=>"N", :ptcpnt_vet_id=>"123456", :reject_reason=>"Converted or Backfilled Rating - no promulgated ratings found", :rating_profile_list=>nil}```

When a Veteran has no ratings at all, we're still getting the same Savon error as before.  Some of the Veteran's are also returning a "locked rating" error (which is in another ticket).

leikkisa:
Another response we can get that causes a similar problem is:
{:error_level=>"0", :reject_reason=>"Locked Rating Detected "}

leikkisa:
This seems to be affecting Veterans which we know have ratings.  I checked a Veteran that does not have ratings, and we're getting the same Savon error as before.  If a Veteran with ratings isn't getting this error, they're getting the locked rating error.  I haven't been able to successfully get ratings for any Veteran since this issue started.

I have asked VBMS and they're going to ask their ratings person tomorrow.

shanear:
This is happening in prod and preventing work, bumping to high priority

shanear:
This is now no longer happening in prod?? We should investigate why this happens.

leikkisa:
After talking to BGS, we heard that this should only happen for really old ratings (like 10 years old), but BGS also mentioned that it is happening for newer ratings.  They said that we do not want these ratings to show up.  This shouldn't block us from getting other ratings, but we need to confirm that this is the case.

It seems like this might be caused by improper test data set up.

It is somewhat a mystery to me why it was happening on some Veterans and is no longer happening for them (but is happening for new test data that has been set up), but will not worry about that until it happens again.

The next step is to send BGS of some examples of this happening.

leikkisa:
Test file numbers have been sent:
700032764
700032765
700032767
700032768
700032770
700032771
700032773
700032774

leikkisa:
If we get this error,  it means the ratings have to be part of an award and one hasn't yet been generated and authorized.

We still don't understand why some that have worked before broke, and other mysteries, but I am going to close this for now. It seems like this error should still cause an error in Caseflow, and if we need to fix it, we can have someone go into VBMS-A and generate and authorize an award for the rating.

[...]

Issue #6221, opened 2018-07-13
Locked ratings will block AMA intake


shanear:
Among the craziness of deploying the ratings services, we discovered an unfortunate bug.

The ratings endpoint we are using seems to fail if there is a rating currently in progress. It says that the rating is "Locked". This will cause the entire flow to fail.

We asked AMO how common it might be to have a rating locked at the same time an AMA review would be received, and they seemed to think that it could be significant. This is because it's locked when it's in a raters queue, which could be multiple days.

The solution here is to request that BGS allow the update the rating services to be allow reads even when the rating is locked.

abbyraskinUSDS:
> The ratings endpoint we are using seems to fail if there is a rating currently in progress.

Does this mean if the rating for that specific issue is in progress, or any rating for that Veteran?

abbyraskinUSDS:
From AMO call:
Basically, it sounds like we should explore getting BGS to change their service
- Mike Palmer: What constitutes it being locked? You can have several claims pulled in, but only actively working on one. Big difference -- "locked" vs. actively working on it. 
- Joshua Will thinks it's locked if it's pulled into the VBMS-R queue...

mc-usds:
Will work with Nikita to submit the SR. 

Also as a side note, it seems like this bug would be affecting the Claim for Increase work being done on the Veteran Facing Tools side. I'm going to reach out to the EVSS guys to see if they have encountered it and if they've brought it to up with BGS or how they've worked around it, if at all.

mc-usds:
service request id: SR-001373

shanear:
Looks like we'll be waiting on an estimate of when this will be released to UAT. Earliest it could be released to prod would be with the December coordinated install.

mc-usds:
Tested in UAT with (700017078 700017075) and the fix appears to working.

PR #8057, opened 2018-11-29
Cancel in progress intake on NilRatingProfileListError

leikkisa:
connects #8050

Currently, if a user intakes a Veteran that gets the Locked rating error, or the Backfilled or Converted error (more generally the NilRatingProfileListError), then their Intake app gets stuck on the "Something went wrong" screen because they have an intake in progress that they can't cancel.

This is a little bit of a bandaid fix which just cancels the intake so that the user can at least continue to work on other intakes.  Ideally we will also improve the error messaging in the future so the user knows when it's related to the specific intake.

Related:
https://github.com/department-of-veterans-affairs/caseflow/issues/7385
https://github.com/department-of-veterans-affairs/caseflow/issues/7054
https://github.com/department-of-veterans-affairs/caseflow/issues/6221


spec/feature/intake/appeal_spec.rb
```diff
@@ -71,8 +71,24 @@
...
 
+  let(:no_ratings_err) { Rating::NilRatingProfileListError.new("none!") }
+
+  it "cancels an intake in progress when there is a NilRatingProfileListError" do
+    allow_any_instance_of(Fakes::BGSService).to receive(:fetch_ratings_in_range).and_raise(no_ratings_err)
+    start_appeal(veteran)
+    intake = Intake.find_by(veteran_file_number: veteran_file_number)
+
+    visit "/intake"
+    expect(page).to have_content("Something went wrong")
```
kencheeto: We could set a [Rails flash message](https://api.rubyonrails.org/classes/ActionDispatch/Flash.html) and display that on the 500 page to give the user a specific message and to ask them to reload! happy to pair on that if you'd like

leikkisa: Sweet. Yeah I started to look at doing a more specific error message but then I felt like I was going to take too long on it and really want to finish my other ticket. Would love to pair on that. Going to go ahead and merge this for now, knowing that it's more of a bandaid than a full fix.

kencheeto: that's the right call, i'll setup some pairing after the legacy optin stuff is finished

PR #8060, opened 2018-11-30
add granular error handling for reclosing ramp appeals

kencheeto:
### Description
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2951/

We're seeing some ramp reclosings fail with an exception, causing the job to fail and skip any remaining appeals. I've added granular error handling so we can see individual reclosings fail in Sentry but the job will continue to run and process the rest of the reclosings.

This does not _fix_ the appeals failing to reclose, but it does prevent the failure from affecting other reclosings.

### Testing Plan
After the SyncIntakeJob runs, check that all reclosable appeals are reclosed


app/jobs/sync_intake_job.rb
```diff
@@ -9,7 +9,20 @@ def perform
...
-    reclosed_appeals = RampClosedAppeal.reclose_all!
-    slack_service.send_notification("Intake: Reclosed RAMP VACOLS appeals (count: #{reclosed_appeals.count})")
+
+    appeals_to_reclose = RampClosedAppeal.appeals_to_reclose
+    reclosed_appeals = []
+    appeals_to_reclose.each do |appeal|
+      begin
+        appeal.reclose!
+        reclosed_appeals << appeal
+      rescue StandardError => e
```
pkarman: +1 to this approach

app/models/ramp_closed_appeal.rb
```diff
@@ -77,20 +77,18 @@ def partial
...
     end
 
-    def reclose_all!
-      appeals_to_reclose = []
+    def appeals_to_reclose
+      output = []
 
-      fully_closed.find_in_batches(batch_size: 800) do |batch|
-        appeals_to_reclose += AppealRepository.find_ramp_reopened_appeals(batch.map(&:vacols_id))
+      RampClosedAppeal.fully_closed.find_in_batches(batch_size: 800) do |batch|
```
pkarman: curious why you fully qualify with the class name since it's already inside a class method?

app/models/ramp_closed_appeal.rb
```diff
@@ -77,20 +77,18 @@ def partial
       where.not(partial_closure_issue_sequence_ids: nil)
     end
 
-    def reclose_all!
-      appeals_to_reclose = []
+    def appeals_to_reclose
+      output = []
```
pkarman: a nit: this variable name doesn't really tell me anything. I had to read all the code a couple times to figure out it's an array of appeals to reclose. Maybe just call it `appeals`? Or keep the existing var name?

spec/jobs/sync_intake_job_spec.rb
```diff
@@ -1,15 +1,42 @@
...
+        expect(appeal_1).to receive(:reclose!)
+        expect(appeal_2).to receive(:reclose!)
+
+        SyncIntakeJob.perform_now
+      end
+
+      context "when an error is thrown" do
+        it "captures it and carries on" do
+          allow(appeal_1).to receive(:reclose!).and_raise(StandardError.new)
+          expect(Raven).to receive(:capture_exception)
```
pkarman: nice!

app/models/ramp_closed_appeal.rb
kencheeto: Ah - was copy/pasting when testing this in a repl, forgot to change it back

app/models/ramp_closed_appeal.rb
kencheeto: Yeah - I thought the method name would be sufficient here, and in general I am wary of method names matching local var names. I could change this to `appeals` as a nice middle ground.

PR #8067, opened 2018-11-30
AMA Case Distribution

cmgiven:


app/models/ama_direct_review_docket.rb
```diff
@@ -0,0 +1,21 @@
+class AmaDirectReviewDocket < Docket
+  TIME_GOAL = 365
+  BECOMES_DUE = -60
```
amprokop: Was this number 45 in the presentation? It means "due for distribution", right?

cmgiven: That's right. It has now been pushed back to 60 days before the goal.

app/models/appeal.rb
```diff
@@ -89,8 +89,12 @@ def self.find_appeal_by_id_or_find_or_create_legacy_appeal_by_vacols_id(id)
     end
   end
 
-  def self.non_priority_decisions_in_the_last_year
-    all_nonpriority.joins(:decision_documents).where("receipt_date > ?", 1.year.ago).count
```
cmgiven: Noticed that this isn't the expected behavior. We want to count decisions in the last year. This would count decided appeals received in the last year. Take a look at the rewrite and confirm that's doing what I think it's doing? The tests here can only reinforce the assumptions I'm making.

spec/models/appeal_spec.rb
```diff
@@ -850,37 +850,27 @@
     end
   end
 
-  context "#non_priority_decisions_in_the_last_year" do
+  context "#nonpriority_decisions_per_year" do
```
cmgiven: Changed the name because conceptually, we're looking for a rate. The fact that it is the last year doesn't matter. There might be cause to lag it by a week, or something. That change in how we're calculating it wouldn't change what we're looking to calculate.

cmgiven:
@amprokop Ready for your eyes here. I still need to finish tests for `ama_case_distribution.rb`, but I have a clear strategy for doing so, now that I've broken out all of the docket proportion logic. I'll aim to do that tomorrow morning.

app/models/appeal.rb
amprokop: Dammit. You're right. Thanks for the catch.

app/models/concerns/proportion_hash.rb
```diff
@@ -0,0 +1,72 @@
...
+  # Set one or more proportions to the fixed values, normalizing the remainder
+  # among the other proportions, proportionate to their prior values.
+
+  def add_fixed_proportions!(fixed)
+    except!(*fixed.keys)
+      .normalize!(to: 1.0 - fixed.values.reduce(0, :+))
+      .merge!(fixed)
+  end
+
+  # Translate the ProportionHash into a Hash where the values are Integers, roughly
```
amprokop: Thank you for this comment!

app/models/dockets/direct_review_docket.rb
```diff
@@ -0,0 +1,49 @@
+class DirectReviewDocket < Docket
+  DAYS_TO_DECISION_GOAL = 365
+  BECOMES_DUE = -60
```
amprokop: nit: maybe a little simpler to call this `DAYS_BEFORE_GOAL_DUE_FOR_DISTRIBUTION` or something less verbose but semantically equivalent

Then you could do `DAYS_BEFORE_GOAL_DUE_FOR_DISTRIBUTION.days.after`

and

time_until_due = Integer(oldest_target - Time.zone.today.to_date) - DAYS_BEFORE_GOAL_DUE_FOR_DISTRIBUTION

cmgiven: Agreed. I guessed at which direction would be least awkward, and I guessed wrong.

app/models/concerns/ama_case_distribution.rb
```diff
@@ -0,0 +1,131 @@
...
+    distribute_appeals(:hearing, @rem, priority: false, genpop: "not_genpop")
+
+    # If we haven't yet met the priority target, distribute additional priority appeals.
+    priority_rem = (priority_target - @appeals.count(&:priority)).clamp(0, @rem)
+    oldest_priority_appeals_by_docket(priority_rem).each do |docket, n|
+      distribute_appeals(docket, n, priority: true)
+    end
+
+    # As we may have already distributed nonpriority legacy and hearing docket cases, we adjust the docket proportions.
+    deduct_distributed_actuals_from_remaining_docket_proportions(:legacy, :hearing)
```
amprokop: It looks like `docket_proportions` isn't used until this line. Is that correct? If so, why do we initialize it on line 21?

cmgiven: Because we want to run all of our counts and things before we go in and start moving cases around.

cmgiven:
The CI tests are failing for good reason. I'm unclear why I'm not getting the same unique index error locally. But I think the issue lies here, where we create the JudgeTask. I don't believe this closes out the DistributionTask, so as our test scenario deliberately makes multiple requests of the same dockets, it attempts to redistribute the same cases.

https://github.com/department-of-veterans-affairs/caseflow/pull/8067/files#diff-3e53c012f6f63af022774da59128a7baL75

app/models/concerns/ama_case_distribution.rb
amprokop: Also, rather than relying on accessing an instance variable below on line 44, would it be easier to reason about if we just returned the remaining_docket_proportions from `deduct_distributed_actuals_from_remaining_docket_proportions` and worked from that? 

Might be clearer to rely on explicitly returned values rather than shared instance variables here, I was having trouble tracking where `remaining_docket_proportions` is used.

amprokop: Oh, got it. Sorry I didn't see your response before my second comment.

amprokop: In that case. Might be better to explictly pass docket_proportions to `distribute_appeals_according_to_remaining_docket_proportions` and `deduct_distributed_actuals_from_remaining_docket_proportions`, and return the new docket proportions where possible, so those methods don't rely on side effects as much. Probably a nit though.

amprokop: On the other hand, @appeals and @rem are also shared instance variables where we rely on side effects.  Personally, I'd probably opt to make those local variables and pass them where needed — but those two seem less confusing to me subjectively.

app/models/concerns/ama_case_distribution.rb
```diff
@@ -0,0 +1,131 @@
...
+      evidence_submission_proportion: docket_proportions[:evidence_submission],
+      hearing_proportion: docket_proportions[:hearing],
+      pacesetting_direct_review_proportion: pacesetting_direct_review_proportion,
+      interpolated_minimum_direct_review_proportion: interpolated_minimum_direct_review_proportion,
+      nonpriority_iterations: @nonpriority_iterations
+    }
+  end
+
+  def distribute_appeals(docket, num, priority: false, genpop: "any", range: nil)
+    return [] unless num > 0
```
amprokop: Is the return value of this method currently used anywhere? IMO it should be used, but it's not currently right?

app/models/concerns/ama_case_distribution.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    if range.nil?
+      appeals = dockets[docket].distribute_appeals(self, priority: priority, genpop: genpop, limit: num)
+    elsif docket == :legacy && priority == false
+      return [] unless range > 0
+
+      appeals = dockets[:legacy].distribute_nonpriority_appeals(self, genpop: genpop, range: range, limit: num)
+    else
```
amprokop: So basically, this function is a no-op if 

- we pass a value for range and the docket is not legacy
- we pass a value for range and the docket is legacy and priority 

Might we consider throwing an error in those cases, or potentially moving all the no-op cases to the top to get them out of the way?

In that case it would be a bit simpler down here- something like

```
if docket == :legacy
  appeals = dockets[:legacy].distribute_nonpriority_appeals(self, genpop: genpop, range: range, limit: num)
else 
   appeals = dockets[docket].distribute_appeals(self, priority: priority, genpop: genpop, limit: num)
end
```

app/models/concerns/ama_case_distribution.rb
```diff
@@ -0,0 +1,131 @@
...
+    end
+  end
+
+  def distribute_appeals_according_to_remaining_docket_proportions
+    @remaining_docket_proportions
+      .normalize!
+      .stochastic_allocation(@rem)
+      .each do |docket, n|
+        appeals = distribute_appeals(docket, n, priority: false)
+        @remaining_docket_proportions[docket] = 0 if appeals.count < n
```
amprokop: This line handles the case when we run out of appeals to distribute on the docket, right?

app/models/distributed_case.rb
```diff
@@ -1,12 +1,24 @@
 class DistributedCase < ApplicationRecord
   belongs_to :distribution
+  belongs_to :task
 
-  validates :distribution, :case_id, :docket, :genpop_query, :ready_at, presence: true
-  validates :priority, :genpop, inclusion: [true, false]
+  validates :distribution, :case_id, :docket, :ready_at, presence: true
+  validates :genpop, inclusion: [true, false], if: :hearing_docket
```
amprokop: Just wondering — we don't need to validate that `genpop` and `genpop_query` are present for legacy appeals?

amprokop: ah nvm — i see you defined `hearing_docket` to include the legacy docket as well. Could consider changing the name.

app/models/docket.rb
```diff
@@ -22,10 +22,9 @@ def appeals(priority: nil, ready: nil)
 
   def count(priority: nil, ready: nil)
     # The underlying scopes here all use `group_by` statements,
-    # so the result of `count` will be a hash of key value pairs
-    # e.g. {{[65, 65]=>2, [66, 66]=>2, [67, 67]=>2}
-    # We want a # returned here, so we count the number of key value pairs.
-    appeals(priority: priority, ready: ready).distinct.count.length
+    # so we count using a subquery finding the relevant ids.
+    appeal_ids = appeals(priority: priority, ready: ready).select("appeals.id")
```
amprokop: Hooray!

app/models/docket_coordinator.rb
```diff
@@ -0,0 +1,103 @@
+class DocketCoordinator
+  # MINIMUM_LEGACY_PROPORTION + MAXIMUM_DIRECT_REVIEW_PROPORTION cannot exceed 1.
+  MINIMUM_LEGACY_PROPORTION = 0.1
+  MAXIMUM_DIRECT_REVIEW_PROPORTION = 0.8
```
amprokop: Curious about where these values come from — are they just reasonable figures you came up with?

app/models/docket_coordinator.rb
```diff
@@ -0,0 +1,103 @@
...
+
+    t = 1 - (dockets[:direct_review].time_until_due_of_oldest_appeal.to_f /
+             dockets[:direct_review].time_until_due_of_new_appeal)
+
+    @interpolated_minimum_direct_review_proportion =
+      (pacesetting_direct_review_proportion * t * INTERPOLATED_DIRECT_REVIEW_PROPORTION_ADJUSTMENT)
+        .clamp(0, MAXIMUM_DIRECT_REVIEW_PROPORTION)
+  end
+
+  def pacesetting_direct_review_proportion
```
amprokop: I'd appreciate a comment here explaining the meaning of the return value — looks like it's a figure representing whether or not we're lagging behind or ahead of processing direct review appeals.

app/models/dockets/direct_review_docket.rb
```diff
@@ -0,0 +1,49 @@
...
+  def due_count
+    appeal_ids = appeals(priority: false, ready: true)
+      .where("target_decision_date <= ?", DAYS_BEFORE_GOAL_DUE_FOR_DISTRIBUTION.days.from_now)
+    Appeal.where(id: appeal_ids).count
+  end
+
+  def time_until_due_of_oldest_appeal
+    oldest_target = appeals(priority: false).limit(1).first.target_decision_date
+    time_until_due = Integer(oldest_target - Time.zone.today.to_date) - DAYS_BEFORE_GOAL_DUE_FOR_DISTRIBUTION
+    time_until_due.clamp(0, time_until_due_of_new_appeal)
```
amprokop: The only case I can think of where `time_until_due` would exceed `time_until_due_of_new_appeal` is if we lower `DAYS_TO_DECISION_GOAL`. Is that right?

spec/models/appeal_spec.rb
amprokop: 👍

app/models/docket_coordinator.rb
```diff
@@ -0,0 +1,103 @@
...
+  end
+
+  def direct_review_due_count
+    @direct_review_due_count ||= dockets[:direct_review].due_count
+  end
+
+  def interpolated_minimum_direct_review_proportion
+    return @interpolated_minimum_direct_review_proportion if @interpolated_minimum_direct_review_proportion
+
+    t = 1 - (dockets[:direct_review].time_until_due_of_oldest_appeal.to_f /
```
amprokop: Mind clarifying what `t` represents?

app/models/docket_coordinator.rb
cmgiven: Pretty much. (We went through a process to get stakeholder buy-in.) Note, though, that particularly the MAXIMUM_DIRECT_REVIEW_PROPORTION is not a "this is how the system should operate" value, but rather a "this is the point at which the system breaks" value.

app/models/concerns/ama_case_distribution.rb
cmgiven: It is used in `distribute_appeals_according_to_remaining_docket_proportions`, where if the array is empty, we won't try to distribute any more appeals from that docket.

app/models/concerns/ama_case_distribution.rb
cmgiven: We actually don't need the no-op behavior. I probably should just throw there.

app/models/concerns/ama_case_distribution.rb
cmgiven: That's right.

app/models/distributed_case.rb
cmgiven: Yeah, maybe `docket_has_hearing_option`.

app/models/docket_coordinator.rb
cmgiven: Yup. It basically aims to calculate how many direct review would we need to work to keep pace with what's coming in the door. We won't actually work that many, because we want to allow the docket to age until it reaches the timeliness goal, but the number we do work will be a percentage of that pacesetting proportion.

app/models/dockets/direct_review_docket.rb
cmgiven: That's right. It is indeed unlikely.

app/models/docket_coordinator.rb
cmgiven: This is an old D3 convention for an interpolator, a number between 0 and 1. I guess I can just call it `interpolator` or `interpolation_factor`.

app/models/concerns/ama_case_distribution.rb
cmgiven: Oh, the length of the array is also used to know how many appeals we distributed.

cmgiven:
Green!

PR #8069, opened 2018-11-30
Remove source request issue id from decisions

aroltsch:
This is last step for this ticket https://github.com/department-of-veterans-affairs/caseflow/issues/7884


aroltsch:
While I was away, someone already removed it.

PR #8076, opened 2018-12-03
Assign Hearings Modal | Time should be required

### LINKED ISSUE #8025: Assign Hearings Modal | Time should be required
When the user does not select time on the modal, we should display an error.

Sjones352:
Resolves #8025
### Description
This pr covers time error message when time is not selected.

### Acceptance Criteria
- [ ] When the user does not select time on the modal, we should display an error.

### Testing Plan
1. Load Caseflow hearings
2. Go to Case details page and select `Schedule Veteran`

screen shot 2018-12-03 at 10 11 10 am


client/app/queue/components/AssignHearingModal.jsx
```diff
@@ -237,6 +238,15 @@ class AssignHearingModal extends React.PureComponent<Props, LocalState> {
 
     const timeOptions = this.getTimeOptions();
 
+    const getTimeError = () => {
```
sharonwarner: We shouldn't need this logic anymore. There's a form validator on line 152 that should do this.

client/app/hearingSchedule/components/AssignHearings.jsx
```diff
@@ -167,9 +167,20 @@ export default class AssignHearings extends React.Component {
...
-    const time = getTime(selectedHearingDay.hearingDate);
     const SROVal = selectedRegionalOffice.value;
-    const qry = `?hearingDate=${date}&regionalOffice=${SROVal}&hearingTime=${time}`;
+    const timer = () => {
+      let time = getTime(selectedHearingDay.hearingDate);
+
+      if (time === '12:00 am ET') {
+        return '';
+
+      } else if (selectedHearingDay) {
```
sharonwarner: Why do we need this additional logic? Shouldn't there always be a selectedHearingDay?

Issue #7905, opened 2018-11-19
Display all AMA appeals for which VSO is the representative in queue


lowellrex:
VSOs currently only see cases in their standard queue tables (they see everything in search) if there has been a task assigned to that VSO (to prepare an informal hearing presentation) that is attached to that case. VSOs want to be able to view all cases for which they are the assigned representative in Caseflow, even if there are no tasks to be completed by the VSO associated with that case.

Discussion 12/17:
- Use a `TrackVeteranTask`

## Acceptance criteria:
* Display all AMA appeals for which VSO is the representative in queue
* Exclude `TrackVeteranTask` from Case Timeline and Task Snapshot

## Rollout plan
* [x] Enable hiding of tasks from Case Timeline/Task Snapshot (#9018)
* [x] Create `TrackVeteranTask` when creating RootTasks for appeals represented by VSOs (#9062)
* [x] Show all tracking tasks for VSO in the new "All cases" tab (#9115)
* [x] Close `TrackVeteranTask` when a decision is dispatched (#9173)
* [x] Create job to update `TrackVeteranTask` nightly when representation changes in BGS (#9190)
* [x] Run this job nightly! (https://github.com/department-of-veterans-affairs/appeals-deployment/pull/1991)

![image](https://user-images.githubusercontent.com/32683958/48709455-84984a80-ebd3-11e8-8be3-da38f5251198.png)

## Blocks:
* Add a job that cancels/creates new `TrackVeteranTask`s when the VSO changes (#7784, #8336)

joeyyang:
@laurjpeterson — clarifying question on this task: how will we know when we need to create an Informal Hearing Presentation task? Our current thought is to create an IHP task if there are no hearings and the VSO is an IHP-writing VSO. Is this correct?

Also, how do we know if a VSO is not an IHP-writing VSO?

laurjpeterson:
- Create IHP tasks for all IHP-writing VSOs if the case is an Evidence Submission docket or Direct Review docket case.
   - one exception: For appeals with Paralyzed Veterans of America (PVA) as the VSO, always create an IHP task.

- Right now, this is just a list of VSOs. Most of these are "national" rather than regional or state-based, like Texas Department of Veterans Affairs.

1. Armed Forces Services Corporation, Room 2W.240J
2. American Legion (AL), Room 2E.102B
3. American Red Cross (ARC), Room 2W.240K
4. American Veterans (AMVETS), Room 2W.240G
[...]

PR #8085, opened 2018-12-03
[Tech Spec] Display all AMA appeals for which VSO is the representative in queue 

joeyyang:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/7905

### Description
Per https://github.com/department-of-veterans-affairs/caseflow/issues/7905, VSOs currently only see cases in their standard queue tables if there has been a task assigned to that VSO (to prepare an informal hearing presentation) that is attached to that case. VSOs want to be able to view all cases for which they are the assigned representative in Caseflow, even if there are no tasks to be completed by the VSO associated with that case.

This PR sketches out an approach to show all AMA appeals for which VSO is the representative in queue, rather than only showing tasks assigned to a user by:
- Creating two new tables:
  - `ihp_writing_vsos`: marks down `organization_id`s of IHP-writing VSOs
  - `vso_appeals`: maps `appeal_id`s to `organization_id`s
- Creates a `VsoAppeal` upon `RootTask` creation.

We chose this approach because the code, as currently written in `root_task.rb`, creates an `InformalHearingPresentation` task when we create a `RootTask`, regardless of whether that VSO writes IHPs at all. The reason why this currently works is because the only VSOs who are included in Caseflow right now are IHP-writing VSOs, but we should change this to reflect future users.

Right now we are creating `InformalHearingPresentation` tasks for every appeal, even if that appeal already has (or will have) a hearing. We should not be creating IHP tasks in this case.

Alternative approaches we considered:
- when creating a `RootTask`, we can create a `VsoVisibilityTask` that doesn't require any action. However, this isn't really a task, and we are creating extra data in the model just to show in the UI. The UI should derive the correct display from the data that already exists.

### DIFF
--- a/app/controllers/organizations/tasks_controller.rb
+++ b/app/controllers/organizations/tasks_controller.rb
@@ -3,10 +3,10 @@ class Organizations::TasksController < OrganizationsController
   before_action :verify_role_access, only: [:index]
 
   def index
-    relevant_tasks = GenericQueue.new(user: organization).relevant_tasks
+    tasks = GenericQueue.new(user: organization).tasks
 
     render json: {
-      tasks: json_tasks(relevant_tasks),
+      tasks: json_tasks(tasks),
       id: organization.id
     }
   end
--- a/app/controllers/organizations/tasks_controller.rb
+++ b/app/controllers/organizations/tasks_controller.rb
@@ -3,10 +3,10 @@ class Organizations::TasksController < OrganizationsController
   before_action :verify_role_access, only: [:index]
 
   def index
-    tasks = GenericQueue.new(user: organization).tasks
+    relevant_tasks = GenericQueue.new(user: organization).relevant_tasks
 
     render json: {
-      tasks: json_tasks(tasks),
+      tasks: json_tasks(relevant_tasks),
       id: organization.id
     }
   end
--- a/app/models/organizations/vso.rb
+++ b/app/models/organizations/vso.rb
@@ -1,4 +1,7 @@
 class Vso < Organization
+  has_one :ihp_writing_vso
+  has_many :appeals, through: :vso_appeals
+
   def user_has_access?(user)
     return false unless user.roles.include?("VSO")
 
--- a/app/models/queues/generic_queue.rb
+++ b/app/models/queues/generic_queue.rb
@@ -11,8 +11,6 @@ class GenericQueue
     relevant_tasks.where(appeal_id: appeal_id, appeal_type: appeal_type)
   end
 
-  private
-
   def relevant_tasks
     Task.incomplete_or_recently_completed
   end
--- a/app/models/queues/generic_queue.rb
+++ b/app/models/queues/generic_queue.rb
@@ -11,6 +11,8 @@ class GenericQueue
     relevant_tasks.where(appeal_id: appeal_id, appeal_type: appeal_type)
   end
 
+  private
+
   def relevant_tasks
     Task.incomplete_or_recently_completed
   end
--- a/app/models/tasks/root_task.rb
+++ b/app/models/tasks/root_task.rb
@@ -25,12 +25,19 @@ class RootTask < GenericTask
 
     def create_vso_subtask!(appeal, parent)
       appeal.vsos.each do |vso_organization|
-        InformalHearingPresentationTask.create(
-          appeal: appeal,
-          parent: parent,
-          status: Constants.TASK_STATUSES.in_progress,
-          assigned_to: vso_organization
+        VsoAppeal.create(
+          organization: vso_organization,
+          appeal: appeal
         )
+
+        if appeal.hearings.empty? && vso_organization.ihp_writing_vso
+          InformalHearingPresentationTask.create(
+            appeal: appeal,
+            parent: parent,
+            status: Constants.TASK_STATUSES.in_progress,
+            assigned_to: vso_organization
+          )
+        end
       end
     end
   end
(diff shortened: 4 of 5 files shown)


amprokop:
Nice! Wondering where this `VSOAppeal` class is going to be used. The VSO can change in the middle of the appeal, will this present problems here? Will we need to update this table? 

We'll probably want to keep a historical record of which VSO worked an appeal even if it changes, too. So maybe we need something like `effective_start_date` etc here? 

Curious how y'all think we could handle complexity.

mdbenjam:
VSO changes is a tricky subject. Curious how y'all choose to handle it.

To be honest I prefer the alternative approach of creating a `VsoVisibilityTask`. In fact, I'd probably say for VSOs that get IHPs, make the IHP a subtask of `VsoVisibilityTask`. My reasoning is twofold.
1) I just think it'll be a lot easier. The UI just works, there doesn't need to be any special casing for VSO Queue UI. I worry about the extra code that's needed to make this work, and worry it sets a dangerous precedent for special casing other teams.
2) While I hear the argument that `this isn't really a task, and we are creating extra data in the model just to show in the UI.` I don't think I agree with it. Tracking appeals does feel like a task to me. VSOs might want to complete these tasks at some point. They might want to assign them to people on their team. And just the act of tracking them feels task-y.

lowellrex:
### VSO changes
Adding the ability to hide cases from VSOs when an appellant changes their representation will not be accomplished by this particular ticket (it is not properly handled with `InformalHearingPresentationTask`s at the moment, so this will not be a regression), however I think one advantage the `vso_appeals` table has over tasks is the flexibility to build this into the model. In order to handle this case we could add a `hidden` field (or `no_longer_represents` or something similar) to the `vso_appeals` table that would be true if the appellant's representation changes. `InformalHearingPresentationTask`s (and any future VSO tasks) could then have logic that checks to see if the VSO still represents the appellant based on the information in the `vso_appeals` table (maybe something like `ihp_task.appeal.vso_appeals.where(organization: assigned_to, hidden: false)` in the `verify_user_access` method). It's not clear to me how we would accomplish this behaviour using the existing task model.

### Historical record of which VSO worked an appeal
I don't think we should ever delete rows from the `vso_appeals` (or really any other table for that matter). I think the historical record component can be accomplished by adding created_at and updated_at (and hidden_at?) timestamps to this table. Perhaps I misunderstand what this historical record would include and this is an insufficient solution?

### VsoVisibilityTask
> Tracking appeals does feel like a task to me.

That's fair. Maybe we could use the `vso_appeals` table to filter visibility for VSO-related tasks and create a `VsoVisibilityTask` for ease of use with the system as currently written?

mdbenjam:
The source of truth for which VSO represents a Veteran will always be BGS. Given that, it feels to me that the `vso_appeals` table is acting like a caching layer between us and BGS. That's why I don't feel it's strictly necessary.

> It's not clear to me how we would accomplish this behaviour using the existing task model.

We probably will need a background job, however we solve this problem, that goes through all open appeals and checks their POA to see if it's new. If it is, it'll need to close some IHP tasks and open new ones. Similarly it would have to do something similar with `VsoVisibilityTask`. To me having a `vso_appeals` table doesn't make this process simpler.

lowellrex:
> [a background job will] need to close some IHP tasks and open new ones. Similarly it would have to do something similar with `VsoVisibilityTask`. 

@mdbenjam How would we hide the IHP and `VsoVisibilityTask`s from VSOs who no longer have access to those appeals? Would we add a new status to the `Task` model or maybe add a `deleted` boolean field to the `tasks` table?

mdbenjam:
So even if a IHP or visibility task exists for a user, they won't be able to see the appeal if they're not allowed to in BGS. So there's no risk there. Then our background job could close those tasks and open up new ones for the relevant VSO.

lowellrex:
So a visibility task with the status `assigned` represents a case where that VSO represents the appellant, and a visibility task with the status `completed` represents a case where that VSO used to represent the appellant but no longer represents the appellant?

mdbenjam:
Maybe using a canceled state would be better here. To differentiate between finishing or reassigning.

lowellrex:
Would visibility tasks make use of any task statuses other than `assigned` and `cancelled`?

mdbenjam:
I tend to think we should still allow VSOs to complete them. We don't know what their workflow is like. After an appeal is finished, they may not want to track it anymore.

lowellrex:
Interesting. And then after a visibility task is completed the only way a VSO could recall that case would be to use the search?

mdbenjam:
Yep. Maybe I'm jumping ahead on functionality. But it feels like that could be a useful feature, that fits in with what we've already developed.

lowellrex:
@laurjpeterson Do we know why VSOs want to see all cases or how they intend to use the "All cases" tab? Are they interested in marking cases complete to dismiss those cases from the "All cases" tab (those cases would still be reachable through search)?

amprokop:
> We probably will need a background job, however we solve this problem, that goes through all open appeals and checks their POA to see if it's new. If it is, it'll need to close some IHP tasks and open new ones. Similarly it would have to do something similar with VsoVisibilityTask. To me having a vso_appeals table doesn't make this process simpler.

I agree that this is probably necessary (altho it might just be appeals with open IHP tasks that we should check, not all appeals).

Let's confirm with @laurjpeterson that indeed Veterans can change VSOs in the middle of the IHP and get a sense of how often this might happen.

laurjpeterson:
@lowellrex - VSOs want to see all of their cases to either help prep for a hearing or write IHPs. I’d imagine they’d also find value in the Case Timeline feature to help them understand what is going on with the appeal to help answer Veteran questions.

@amprokop - yes, POA could change at any time, including during the IHP. I’ve heard more often that the VSO is assigned cases but are improperly listed as the Veteran’s POA because of discrepancies in VBMS. So the POA listed changes, but wasn’t actually correct in the first place.

lowellrex:
Closing this PR. Work will be continued in other PRs connected to #7905.

Issue #7982, opened 2018-11-26
AMA Intake | Store decision date for each decision issue


shanear:
We'll need to keep track of the decision date on each issue for https://github.com/department-of-veterans-affairs/caseflow/issues/7978. We need to begin finding and storing this value.

shanear:
```
=> #<EndProduct:0x0000562e1a1625a8 @claim_id="600149145", @claim_date=Fri, 01 Feb 2019 00:00:00 UTC +00:00, @claim_type_code="030HLRNR", @modifier="030", @status_type_code="CLR", @last_action_date=Fri, 01 Feb 2019 00:00:00 UTC +00:00, @claimant_first_name="BERRY", @claimant_last_name="MCGARRAH", @payee_code="00">
irb(main):009:0> HigherLevelReview.find_by(veteran_file_number: "700035233").decision_issues
[2019-02-01 14:39:09 -0500]   HigherLevelReview Load (0.9ms)  SELECT  "higher_level_reviews".* FROM "higher_level_reviews" WHERE "higher_level_reviews"."veteran_file_number" = $1 LIMIT $2  [["veteran_file_number", "700035233"], ["LIMIT", 1]]
[2019-02-01 14:39:09 -0500]   DecisionIssue Load (1.0ms)  SELECT  "decision_issues".* FROM "decision_issues" WHERE "decision_issues"."decision_review_id" = $1 AND "decision_issues"."decision_review_type" = $2 LIMIT $3  [["decision_review_id", 1054], ["decision_review_type", "HigherLevelReview"], ["LIMIT", 11]]
[...]

PR #8091, opened 2018-12-04
Add approx_decision_date to DecisionIssue

leikkisa:
connects #7982


app/models/request_issue.rb
```diff
@@ -192,7 +192,9 @@ def create_decision_issue_from_disposition
         # use epe last_synced_at as a proxy for when the decision was made
         disposition_date: end_product_establishment.last_synced_at,
         decision_review: review_request,
-        benefit_type: benefit_type
+        benefit_type: benefit_type,
+        # save the end product's last action date to be used as an approximate decision date
```
shanear: no need for this guy

app/models/request_issue.rb
```diff
@@ -192,7 +192,9 @@ def create_decision_issue_from_disposition
...
+        last_action_date: end_product_establishment.result.last_action_date
```
shanear: save this on both

db/schema.rb
```diff
@@ -195,6 +195,7 @@
     t.string "decision_review_type"
     t.integer "decision_review_id"
     t.string "benefit_type"
+    t.date "last_action_date"
```
shanear: rename to `end_product_last_action_date`

PR #8093, opened 2018-12-04
Adding where to find documentation to the Caseflow README

mdbenjam:
### Description
Lists locations to find documentation.


README.md
```diff
@@ -459,3 +459,12 @@ When Caseflow Monitor starts working again, switch the banner back to automatic
 ```
 Rails.cache.write(:degraded_service_banner, :auto)
 ```
+
+## Documentation
+We have a lot of technical documentation spread over a lot of different repositories. Here is a non-exhaustive mapping of where to find documentation:
+
+- [Local Caseflow Setup](https://github.com/department-of-veterans-affairs/caseflow/tree/master/docs)
+- [Test data setup in lower environments](https://github.com/department-of-veterans-affairs/appeals-qa/tree/master/docs)
+- [Caseflow specific devops documentation](https://github.com/department-of-veterans-affairs/appeals-deployment/tree/master/docs) This folder also contains our [first responder manual](https://github.com/department-of-veterans-affairs/appeals-deployment/blob/master/docs/first-responder-manual.md), which is super in understanding our production systems.
```
enriquemanuel: we have 2 devops repos
- appeals-deployments
- devops

we are trying to consolidate this, but for now, its a bit confusing having both, but maybe we need to point to both.
Also to share that this documentation is shared with the vets.gov team as well (this applies to the following line as well)

mdbenjam: I'll update that the devops repo is shared with the vets.gov team.

PR #8094, opened 2018-12-04
adds queue dropdown selector to attorney queue table views

### LINKED ISSUE #8065: Filter out JudgeTeam membership from queue selector
#8012 added a dropdown to the queue table views to allow folks to switch between queues for organizations to which they belong. We will soon be [moving judge team membership](https://github.com/department-of-veterans-affairs/caseflow/pull/8063) to the same organizations model and need to prevent judge teams from appearing in that dropdown since attorneys should not navigate to judge team queues (currently there is no concept of a "judge team queue" so the application would display an empty queue, but it is still probably best to avoid this potential confusion... and also to leave open the possibility of moving judge assign view to this "judge team queue").

Currently, we do not show the queue dropdown selector on attorney queue table views but we probably will have to soon since they can be members of Board teams (Quality Review for example). So this will become an issue at that point even though it appears less pressing at the current moment.

## Acceptance criteria
* [x] Hide `JudgeTeam`s from queue selector
* [x] Display queue selector on attorney queue table views

tomas-nava:
Resolves #8065

### Description
Adds queue dropdown selector to attorney queue table views for attorneys who're in appropriate organizations.

### Testing Plan
Automated tests.

#### dropdown selector interaction
smaller


joeyyang:
hey @tomas-nava — do you have a screenshot?

joeyyang:
LGTM, pending screenshot!

Issue #6710, opened 2018-08-22
Case search / Case list | Add Higher Level Reviews and Supplemental Claims


shanear:
As a VBA employee I want to be able to see all the higher level reviews, supplemental claims and appeals for a veteran, and the details about those reviews.

### AC
- Users can view a list of SCs and HLRs in search results
- Users can view a list of SCs and HLRS in case list within Case details

#### Option 1 (All data revealed in table form)
![search-results-sc-hlr-1](https://user-images.githubusercontent.com/31224257/49529520-31392400-f884-11e8-9f4e-30395a699038.png)

#### Option 2a (Claim high level information on display (and accordion behavior reveals EP code-level information)
![search-results-sc-hlr-2a](https://user-images.githubusercontent.com/31224257/49529685-9260f780-f884-11e8-936f-13257e037893.png)

#### Option 2b ( (Claim high level information on display (and accordion behavior reveals EP code-level information) Opened accordions
![search-results-sc-hlr-2b](https://user-images.githubusercontent.com/31224257/49529694-9856d880-f884-11e8-8f5b-fbb20bd12972.png)

mkhandekar:
adding myself to this one so we can circle back in a few weeks

laurjpeterson:
@shanear - circling back here. @sneha-pai put together some designs for this on Friday and is going to post them to this issue for your and @carolaponce's review.

After her exploration of combining all of the requirements listed in this [comment](https://github.com/department-of-veterans-affairs/caseflow/issues/7016#issuecomment-428678275), she came to the conclusion that we could not fit all SC, HLR and appeal results in one table. Instead, SCs and HLRs would be in one table, while appeals live in another.

sneha-pai:
@shanear here's a proposal for how higher level reviews and supplemental claims might appear when a veteran ID is searched for in Caseflow Search. [ @laurjpeterson ]

### Questions for your consideration: 
- what would you call that list? Right now it's tentatively **VBA Reviews**. 
- Is it important to see **Appellant Name** for HR reviews and Supplemental Claims?
- Are any table columns missing?

### Proposed design:
![search results-proposed](https://user-images.githubusercontent.com/31224257/48099914-3ce6eb80-e1ef-11e8-85ae-7ce175841579.png)

shanear:
> what would you call that list? Right now it's tentatively VBA Reviews.

[...]

PR #8098, opened 2018-12-04
Add Higher Level Reviews and Supplemental Claims to search view

joeyyang:
Resolves (partially) https://github.com/department-of-veterans-affairs/caseflow/issues/6710

### Description
- Users can view a list of SCs and HLRs in search results
screenshot 2018-12-17 10 25 09

Notes / review questions:
- The original ticket says "Users can view a list of SCs and HLRS in case list within Case details". @sneha-pai, are there designs for this?
- For a code review, I'm concerned primarily about these things:
  - **Logic:** 
    - I am not 100% clear on the data model here, so I'd like someone to closely review my logic to make sure it is correct
    - Are there any more cases I should test for?
  - **Idiomatic code:** Please confirm that my Ruby code conforms to best practices & this codebase's styleguide


app/models/claim_review.rb
```diff
@@ -48,6 +48,13 @@ def processed_at_column
     def error_column
       :establishment_error
     end
+
+    def find_all_by_file_number(file_number)
+      claim_reviews =
+        HigherLevelReview.where(veteran_file_number: file_number) +
+        SupplementalClaim.where(veteran_file_number: file_number)
```
shanear: The map here fits better in the controller than the model, since it's serializing the returned objects for the UI. Additionally, the name `find_all_by_file_number` seems like it should be returning `ClaimReview` objects.

```
def find_all_by_file_number(file_number)
  HigherLevelReview.where(veteran_file_number: file_number) +
    SupplementalClaim.where(veteran_file_number: file_number)
end
```

```
ClaimReview.find_all_by_file_number(veteran_file_number).map(&:search_table_ui_hash)
```

app/models/claim_review.rb
```diff
@@ -104,6 +111,19 @@ def find_request_issue_by_description(description)
     request_issues.find { |reqi| reqi.description == description }
   end
 
+  def search_table_ui_hash
+    {
+      claim_id: id,
+      veteran_file_number: veteran_file_number,
+      veteran_full_name: Veteran.find_by(file_number: veteran_file_number).name.formatted(:readable_full),
+      ep_codes: ep_codes,
```
shanear: It might be best to just return each ep as a seriailzed object, instead of a series of arrays:

```
{
  ...
  end_products: end_product_establishments.map(&: search_table_ui_hash)
}
```

app/models/claim_review.rb
```diff
@@ -104,6 +111,19 @@ def find_request_issue_by_description(description)
...
+      claimant_names: claimants.map(&:name),
+      review_type: type_of_claim_review(self.class),
```
shanear: The private method here is overkill. No need for if/else statements:

```suggestion
       review_type: type.underscore,
```

see: see: https://apidock.com/rails/String/underscore

app/models/claim_review.rb
```diff
@@ -134,4 +154,16 @@ def fetch_dispositions_from_vbms(reference_id)
   def matching_request_issue(contention_id)
     RequestIssue.find_by!(contention_reference_id: contention_id)
   end
+
+  def ep_codes
+    end_product_establishments.map(&:modifier)
+  end
+
+  def type_of_claim_review(review_class)
```
shanear: the complexity here is unnecessary, see comment above.

client/app/queue/OtherReviewsTable.jsx
```diff
@@ -0,0 +1,137 @@
...
+    const styles = {};
+
+    this.props.reviews.forEach((review, i) => {
+      if (review.epCodes.length > 1) {
+        styles[`& > tbody > tr:nth-of-type(${i + 1}) > td:nth-of-type(3)`] = { padding: 0 };
+        styles[`& > tbody > tr:nth-of-type(${i + 1}) > td:nth-of-type(4)`] = { padding: 0 };
+        styles[`& > tbody > tr:nth-of-type(${i + 1}) > td:nth-of-type(5)`] = { padding: 0 };
+        styles[`& > tbody > tr:nth-of-type(${i + 1})`] = { verticalAlign: 'top' };
+      }
+    });
```
shanear: not as much of an expert on y'alls UI, but this seems like extremely complex custom CSS for what it's trying to achieve.

app/models/claim_review.rb
joeyyang: awesome, i was looking for something exactly like this!

app/models/claim_review.rb
joeyyang: makes sense to me!

app/models/claim_review.rb
joeyyang: yup, that ensures all the data lines up as well, will make this change!

client/app/queue/OtherReviewsTable.jsx
joeyyang: agreed, the change here was either going to be adding styles to the base `Table` class or doing something like this — unfortunately, this is the most exact way to apply these styles :(

joeyyang:
thanks @shanear, back to you!

app/controllers/appeals_controller.rb
```diff
@@ -14,7 +14,13 @@ def show_case_list
...
       format.html { render template: "queue/index" }
       format.json do
-        return get_appeals_for_file_number(Veteran.find(params[:caseflow_veteran_id]).file_number)
+        caseflow_veteran_id = params[:caseflow_veteran_id]
+        veteran_file_number = Veteran.find(caseflow_veteran_id).file_number
+
+        return render json: {
+          appeals: get_appeals_for_file_number(veteran_file_number),
+          claim_reviews: ClaimReview.find_all_by_file_number(veteran_file_number).map(&:search_table_ui_hash)
+        }
```
lowellrex: I'm running into the error below when I search for cases. I think it might be because we need to duplicate this logic in `AppealsController.index`.

image

client/app/queue/CaseListView.jsx
```diff
@@ -42,7 +50,15 @@ class CaseListView extends React.PureComponent {
       then((response) => {
         const returnedObject = JSON.parse(response.text);
 
+        const claimReviews = returnedObject.claim_reviews.map((review) => {
+          return {
+            ...review,
+            caseflow_veteran_id: this.props.caseflowVeteranId
+          };
+        });
```
lowellrex: Can we include the `caseflow_veteran_id` in the response we get from the backend so we don't have to add that attribute to each claim review here on the frontend?

client/app/queue/CaseListView.jsx
```diff
@@ -63,6 +80,12 @@ class CaseListView extends React.PureComponent {
       </div>;
     }
 
+    const claimReviewsCount = this.props.claimReviews && this.props.claimReviews.length;
+
+    if (claimReviewsCount) {
+      otherReviewsTable = <OtherReviewsTable reviews={this.props.claimReviews} />;
+    }
```
lowellrex: Let's move the logic to show/hide the `OtherReviewsTable` into the `OtherReviewsTable` component itself (as well as the header), so that we just call `<OtherReviewsTable reviews={this.props.claimReviews} />` here and within the other component we decide whether we render the table or just return null.

client/app/queue/CaseListView.jsx
```diff
@@ -73,7 +96,12 @@ class CaseListView extends React.PureComponent {
       {this.searchPageHeading()}
       <br /><br />
       <h2 className="cf-push-left" {...fullWidth}>{heading}</h2>
+
+      <h3 className="cf-push-left" {...fullWidth}>Appeals</h3>
```
lowellrex: Let's move "Appeals" and "Other Decision Reviews" to COPY.json.

client/app/queue/OtherReviewsTable.jsx
```diff
@@ -0,0 +1,141 @@
...
+import { DateString } from '../util/DateUtil';
+import COPY from '../../COPY.json';
+import EP_STATUSES from '../../constants/EP_STATUSES.json';
+
+class SubdividedTableRow extends React.PureComponent {
+  render = () => {
+    const styling = {
+      boxSizing: 'content-box',
+      height: '22px',
+      padding: '10px 15px'
```
lowellrex: Let's use `rem`s if we can here for consistency across the codebase.

client/app/queue/OtherReviewsTable.jsx
```diff
@@ -0,0 +1,141 @@
...
+      header: COPY.OTHER_REVIEWS_TABLE_EP_CODE_COLUMN_TITLE,
+      valueFunction: (review) => {
+        const doesReviewHaveEndProducts =
+          review.reviewType === 'higher_level_review' &&
+          review.endProducts && review.endProducts.length > 0;
+
+        if (doesReviewHaveEndProducts) {
+          return review.endProducts.map((endProduct, i) => {
+            return <SubdividedTableRow rowNumber={i}>{`${endProduct.code} ${endProduct.modifier}`}</SubdividedTableRow>;
+          });
```
lowellrex: Actually can we just simplify this logic to check for the existence of end products or do we expect supplemental claims to also have end products? If we don't expect supplemental claims to ever have end products we could do something like:
```jsx
if (review.endProducts && review.endProducts.length) {
  return review.endProducts.map((ep, i) => <SubdividedTableRow rowNumber={i}>
    {`${ep.code} ${ep.modifier}`}
  </SubdividedTableRow>
}

// Do we need to check for the type of review in our fall through case?
return <em>{COPY.OTHER_REVIEWS_TABLE_SUPPLEMENTAL_CLAIM_NOTE}</em>;
```

client/app/queue/selectors.js
```diff
@@ -142,6 +150,13 @@ export const appealsByCaseflowVeteranId = createSelector(
       appeal.caseflowVeteranId.toString() === caseflowVeteranId.toString())
 );
 
+export const claimReviewsByCaseflowVeteranId = createSelector(
+  [claimReviewsSelector, getCaseflowVeteranId],
+  (claimReviews: ClaimReviews, caseflowVeteranId: ?string) =>
+    _.filter(claimReviews, (claimReview: ClaimReview) => claimReview.caseflowVeteranId && caseflowVeteranId &&
+      claimReview.caseflowVeteranId.toString() === caseflowVeteranId.toString())
+);
```
lowellrex: Could we store the claim reviews in a hash of arrays with the caseflow veteran ID as the key?

client/app/queue/utils.js
```diff
@@ -280,6 +280,28 @@ export const prepareAppealForStore =
     };
   };
 
+export const prepareClaimReviewForStore = (claimReviews: Array<Object>) => {
+  const claimReviewHash = claimReviews.reduce((accumulator, claimReview) => {
+    const key = `${claimReview.review_type}-${claimReview.claim_id}`;
```
lowellrex: Do we ever pull things out of the store by this ID? Or do we just use this to ensure we have some deterministic way of identifying claim reviews so we can appropriately overwrite them if they change from request to request?

client/app/queue/OtherReviewsTable.jsx
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    const styles = {};
+
+    this.props.reviews.forEach((review, i) => {
+      if (review.endProducts && review.endProducts.length > 1) {
+        styles[`& > tbody > tr:nth-of-type(${i + 1}) > td:nth-of-type(3)`] = { padding: 0 };
+        styles[`& > tbody > tr:nth-of-type(${i + 1}) > td:nth-of-type(4)`] = { padding: 0 };
+        styles[`& > tbody > tr:nth-of-type(${i + 1}) > td:nth-of-type(5)`] = { padding: 0 };
+        styles[`& > tbody > tr:nth-of-type(${i + 1})`] = { verticalAlign: 'top' };
```
lowellrex: Could we just use a `rowspan` on the first column and render these rows at `tr` elements instead of divs?

lowellrex:
I couldn't get the test data to work properly for me (even after the fix [suggested here](https://github.com/department-of-veterans-affairs/caseflow/pull/8098#discussion_r242548727)). Any tips on how I can get other decision reviews to show up?

image

app/controllers/appeals_controller.rb
joeyyang: Ah, great catch, thank you — fixed!

client/app/queue/CaseListView.jsx
joeyyang: done!

db/seeds.rb
```diff
@@ -390,6 +390,122 @@ def create_ama_appeals
...
+      contention_reference_id: "12345",
+      decision_date: Date.new(2018, 5, 1),
+      ineligible_reason: :untimely
+    )
+
+    higher_level_review.create_issues!([
+                                         eligible_request_issue,
+                                         untimely_request_issue
+                                       ])
+    higher_level_review.establish!
```
lowellrex: I'm getting the following error when I run `bundle exec rake db:seed`:

```
rake aborted!
EndProductEstablishment::NoAvailableModifiers: EndProductEstablishment::NoAvailableModifiers
~/Projects/caseflow/app/models/end_product_establishment.rb:454:in `find_open_modifier'
~/Projects/caseflow/app/models/end_product_establishment.rb:402:in `end_product_to_establish'
~/Projects/caseflow/app/models/end_product_establishment.rb:141:in `perform!'
~/Projects/caseflow/app/models/claim_review.rb:97:in `block in establish!'
~/.rvm/gems/ruby-2.5.1@certification/gems/activerecord-5.1.6.1/lib/active_record/relation/delegation.rb:39:in `each'
~/.rvm/gems/ruby-2.5.1@certification/gems/activerecord-5.1.6.1/lib/active_record/relation/delegation.rb:39:in `each'
~/Projects/caseflow/app/models/claim_review.rb:96:in `establish!'
~/Projects/caseflow/db/seeds.rb:500:in `create_higher_level_reviews_and_supplemental_claims'
~/Projects/caseflow/db/seeds.rb:771:in `seed'
~/Projects/caseflow/db/seeds.rb:789:in `<main>'
```

Any thoughts about how to resolve this?

client/app/queue/CaseList/CaseListActions.js
```diff
@@ -90,14 +103,22 @@ export const fetchAppealsUsingVeteranId = (searchQuery) =>
...
+        let isResponseEmpty;
+        const returnedObject = (response.text) ? JSON.parse(response.text) : null;
 
-        if (_.size(returnedObject.appeals) === 0) {
+        if (returnedObject) {
+          isResponseEmpty = _.size(returnedObject.appeals) === 0 &&
+            _.size(returnedObject.claim_reviews) === 0;
+        }
+
+        if (!returnedObject || isResponseEmpty) {
```
lowellrex: I think we can simplify this logic. Since this is the first function of the then block, we know we're in the happy path and received a successful response from the server. This request hits `/appeals` which routes to `AppealsController.index` and there is only [one successful response we return](https://github.com/department-of-veterans-affairs/caseflow/pull/8098/files#diff-6c45b90fc8f3b5532860540b192918abR16) from that function so we can assume that is our response here. If we do that I think we can simplify this logic to something like:

```jsx
const returnedObject = JSON.parse(response.text);

if (_.size(returnedObject.appeals) === 0 && _.size(returnedObject.claim_reviews) === 0) {
  dispatch(fetchedNoAppealsUsingVeteranId(veteranId));

  return reject();
}
```

I'm not super experienced with Javascript so this may be incorrect though. Let me know what you think.

client/app/queue/OtherReviewsTable.jsx
```diff
@@ -0,0 +1,141 @@
...
+    const styling = {
+      boxSizing: 'content-box',
+      height: '22px',
+      padding: '10px 15px'
+    };
+    const topBorderStyle = '1px solid #D6D7D9';
+
+    if (this.props.rowNumber > 0) {
+      styling.borderTop = topBorderStyle;
+    }
```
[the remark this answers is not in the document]
lowellrex: What're your thoughts on this?

client/app/queue/OtherReviewsTable.jsx
lowellrex: What're your thoughts on this?

app/models/claim_review.rb
```diff
@@ -143,6 +148,18 @@ def find_request_issue_by_description(description)
     request_issues.find { |reqi| reqi.description == description }
   end
 
+  def search_table_ui_hash
+    {
+      caseflow_veteran_id: Veteran.find_by(file_number: veteran_file_number).id,
+      claim_id: id,
+      veteran_file_number: veteran_file_number,
+      veteran_full_name: Veteran.find_by(file_number: veteran_file_number).name.formatted(:readable_full),
+      end_products: end_product_establishments,
```
shanear: we may want to map a serialization method over these?

spec/feature/queue/search_spec.rb
```diff
@@ -39,6 +41,26 @@
...
+      it "shows a higher level review" do
+        expect(page).to have_content("Higher Level Reviews &")
+        expect(page).to have_content(COPY::OTHER_REVIEWS_TABLE_TITLE)
+        expect(find(".cf-other-reviews-table > tbody")).to have_content("Higher Level Review")
+      end
+
+      it "shows a supplemental claim" do
+        expect(page).to have_content("Higher Level Reviews &")
+        expect(page).to have_content(COPY::OTHER_REVIEWS_TABLE_TITLE)
+        expect(find(".cf-other-reviews-table > tbody")).to have_content("Supplemental Claim")
```
shanear: I feel like these are incomplete tests. Shouldn't we be testing that the correct information is displayed?

Whether 2 EPs on the same supplemental claim or HLR displays correctly?

If there are no HLRs or SCs it doesn't show the heading?

app/models/claim_review.rb
```diff
@@ -143,6 +148,18 @@ def find_request_issue_by_description(description)
     request_issues.find { |reqi| reqi.description == description }
   end
 
+  def search_table_ui_hash
```
shanear: this could use a unit test

app/models/claim_review.rb
```diff
@@ -37,6 +37,11 @@ def ui_hash
     )
   end
 
+  def self.find_all_by_file_number(file_number)
```
shanear: this could use a small unit test

db/seeds.rb
joeyyang: @jiahuang advised me to run `Fakes::EndProductStore.new.clear!`

Jia — should we add this line to the `seeds.rb` file somewhere?

client/app/queue/CaseList/CaseListActions.js
joeyyang: I don't love this either but I did encounter a condition where `response.text` is equal to `""`, and `JSON.parse("")` fails silently and exits, which led me to handle this condition specifically.

client/app/queue/OtherReviewsTable.jsx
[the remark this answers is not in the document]
joeyyang: I'm not sure we can do it this way in this case, because we only want to apply the styles conditionally to certain table row indices, depending on the result from the server. 

Perhaps in a later PR we can take a look at modifying the `Table` component to handle this case, `rowspan`s and all the other functionality.

client/app/queue/OtherReviewsTable.jsx
joeyyang: see above comment

db/seeds.rb
```diff
@@ -578,6 +694,7 @@ def create_vsos
 
   def clean_db
     DatabaseCleaner.clean_with(:truncation)
+    Fakes::EndProductStore.new.clear!
```
lowellrex: Let's clear all redis keys when we reseed the database! Maybe something like the following?
```ruby
r = Redis.new(url: Rails.application.secrets.redis_url_cache)
r.keys.each { |k| r.del(k) }
```

sunil-sadasivan:
@joeyyang, I _think_ this broke the `rake db:seed` command. I'm now getting a 

```
PG::NotNullViolation: ERROR:  null value in column "benefit_type" violates not-null constraint
```

I _think_ we need to include a `benefit_type` in creating the request_issues in the seed file.  At least that helped fix it for me:

https://github.com/department-of-veterans-affairs/caseflow/pull/8098/files#diff-f7a25bc30a409fc77050703fc048ca8cR478

Issue #7605, opened 2018-10-30
Release 2 | Build Schedule Remove VLJs


Chingujo:
As a hearing coordinator, I want to be able to remove a Veteran Law Judge from the schedule to account for changes in VLJ availability.

### Acceptance Criteria
- [ ] Verify that there is a way for the user to remove a VLJ from the Build schedule UI 
- [ ] Verify that there is a confirmation when the action is completed

Chingujo:
@MeredithStewart 

We talked about removing judges when they retire or leave, but we would want to add as well? Otherwise they would have to run the judge algorithm every time they added/removed dates?

MeredithStewart:
@Chingujo yes, need ability to add VLJ to the schedule - this will happen as frequently as the remove functionality. The judge algorithm wouldn't be good to run for these -- it's designed for bulk, 1 x use prior to the quarter.

MeredithStewart:
Updated user stories and acceptance criteria 11/15/18.

laurenrussell:
https://va-gov.invisionapp.com/share/EPP1JSLWN48

MeredithStewart:
Updated to add the functionality to remove the hearing day (with no Veterans selected).

oscarAtNava:
Completed together with Add ticket.

PR #8100, opened 2018-12-04
Add / Edit a Hearing Day

### LINKED ISSUE #7602: Release 2 | Editing the schedule | Add hearing day
As a hearing admin, I want to be able to add a hearing date to the schedule to accommodate for varying availability at Regional Offices, and add dates as needed to accommodate AOD Veterans.  

### Mocks 
* https://va-gov.invisionapp.com/share/G6P70ASQTR3

### Acceptance Criteria 
- [ ] Verify that there can select date for new hearing day  
- [ ] Verify that the user can select hearing type (Video or Central) 
- [ ] Verify that the user can select RO from a list of ROs that hold hearings 
- [ ] Verify that an VLJ can be added from of all VLJs 
- [ ] Verify that a hearing day can be created without selecting a VLJ
- [ ] Verify that a room number is automatically assigned based on available room numbers 
- [ ] Verify that a coordinator can select if they do not want a Board hearing room assigned 
- [ ] Verify that an error is shown if there are no hearing rooms available (and no hearing room hasn't been selected) 
- [ ] Verify that a coordinator can be added (free text)
- [ ] Verify that the hearing day can be created without selecting a coordinator 
- [ ] Verify that User can save text in the hearing day's notes field. 
- [ ] Verify that there is a confirmation when the hearing day is added 
- [ ] Verify that errors are shown when user does not select a required fields (date, type, RO if type is video). 
- [ ] Verify that error is shown when hearing day cannot be saved  
- [ ] Verify that only hearing admin can add hearing day to schedule (not hearing coordinators)

Users should only be able to add hearing days through Caseflow past January 1st for CO hearings and past April 1st for video hearings.

oscarAtNava:
Resolves #7602 and #7605 

### Description
Implemented UI and back-end functionality to add a new hearing day (parent record), and edit an exiting hearing day. 

### Acceptance Criteria
- [ ] Code compiles correctly
- [ ] Test pass

### Testing Plan
Acceptance criteria is documented in the following tickets: #7602, #7605.


client/app/hearingSchedule/components/DailyDocket.jsx
```diff
@@ -376,6 +380,12 @@ export default class DailyDocket extends React.Component {
         <h1>Daily Docket ({moment(this.props.dailyDocket.hearingDate).format('ddd M/DD/YYYY')})</h1> <br />
         <div {...backLinkStyling}>
           <Link to="/schedule">&lt; Back to schedule</Link>&nbsp;&nbsp;
+          <Link
+            {...editLinkStyling}
+            linkStyling="true"
```
sharonwarner: This can just be `linkStyling`.

client/app/components/Dropdown.jsx
```diff
@@ -15,7 +15,8 @@ export default class Dropdown extends React.Component {
       required,
       value,
       defaultText,
-      readOnly
+      readOnly,
```
sharonwarner: I don't think it makes a lot of sense to support both this component and the SearchableDropown component. Currently this component is only used in one place in the repository, so I think we should move towards deprecating this component. Could we switch to using the SearchableDropdown instead?

client/app/hearingSchedule/components/HearingDayEditModal.jsx
```diff
@@ -0,0 +1,220 @@
...
+  onHearingDayModified
+} from '../actions';
+
+const notesFieldStyling = css({
+  height: '100px',
+  fontSize: '10pt'
+});
+
+// May explore building API to get from back-end at future point.
+const roomOptions = [
```
sharonwarner: I think we should move towards what queue is doing to solve this problem: we can create a json object in client/constants and it will be accessible from the backend and the frontend.

client/app/components/Dropdown.jsx
oscarAtNava: I changed all dropdowns to use the Searchable one. This change was done before I made that change. I will leave it there as there is no harm to our functionality.

client/app/hearingSchedule/components/DailyDocket.jsx
oscarAtNava: But there is a backLinkStyling defined now so this name reflects there are two link stylings in this file.

client/app/hearingSchedule/components/HearingDayEditModal.jsx
oscarAtNava: Sounds good to me. I will create a ticket to fix this in a subsequent PR as we probably will have to make back-end changes to code that uses the existing rb file.

oscarAtNava: Ticket created: https://github.com/department-of-veterans-affairs/caseflow/issues/8122

client/app/hearingSchedule/components/DailyDocket.jsx
```diff
@@ -379,6 +383,12 @@ export default class DailyDocket extends React.Component {
         <h1>Daily Docket ({moment(this.props.dailyDocket.hearingDate).format('ddd M/DD/YYYY')})</h1> <br />
         <div {...backLinkStyling}>
           <Link to="/schedule">&lt; Back to schedule</Link>&nbsp;&nbsp;
+          <Link
```
sharonwarner: I think this should be a button instead of a link. We can still style it as a link. The way it's set up now, the cursor doesn't make it obvious it's clickable.

client/app/hearingSchedule/containers/DailyDocketContainer.jsx
```diff
@@ -72,7 +129,86 @@ export class DailyDocketContainer extends React.Component {
...
+        data.bva_poc = this.props.coordinator.label;
+      }
+
+      if (this.props.notes) {
+        data.notes = this.props.notes;
+      }
+
+      ApiUtil.put(`/hearings/${this.props.dailyDocket.id}/hearing_day`, { data }).
+        then({}, () => {
+          this.setState({ serverError: true });
```
sharonwarner: It doesn't look like this data is saved in the store correctly after it's saved. When I update the room in the modal, hit save, and then edit the hearing day again, the room information still displays the old room. I have to refresh the page to see the updates.

sharonwarner: It also looks like the notes don't save to the database.

client/app/hearingSchedule/components/ListSchedule.jsx
```diff
@@ -255,6 +256,11 @@ class ListSchedule extends React.Component {
               <FilterRibbon
                 filteredByList={this.state.filteredByList}
                 clearAllFilters={this.clearFilteredByList} />
+              <Link
```
sharonwarner: This should also be a button.

client/app/hearingSchedule/containers/ListScheduleContainer.jsx
```diff
@@ -47,31 +73,173 @@ export class ListScheduleContainer extends React.Component {
...
+      bva_poc: this.props.coordinator.label,
+      notes: this.props.notes,
+      assign_room: !this.props.roomNotRequired
+    };
+
+    if (this.props.selectedRegionalOffice && this.props.selectedRegionalOffice.value !== '') {
+      data.regional_office = this.props.selectedRegionalOffice.value;
+    }
+
+    ApiUtil.post('/hearings/hearing_day.json', { data }).
```
sharonwarner: Same with this update, we should save the hearing day in redux when we get a successful response so the user doesn't have to refresh the page.

client/app/hearingSchedule/components/ListSchedule.jsx
oscarAtNava: Done

client/app/hearingSchedule/components/DailyDocket.jsx
oscarAtNava: Done

app/controllers/hearings/hearing_day_controller.rb
```diff
@@ -237,4 +245,57 @@ def format_for_client(json_hash)
...
+
+  def rooms_are_available
+    # Coming from Add Hearing Day modal but no room required
+    if do_not_assign_room
+      params.delete(:assign_room)
+      params[:room_info] = ""
+      return true
+    end
+    # Coming from regular create from RO algorithm
+    return if no_assign_room_logic_required
```
mdbenjam: Should this return `true` or `false`? Currently it's returning nil.

app/controllers/hearings/hearing_day_controller.rb
```diff
@@ -237,4 +245,57 @@ def format_for_client(json_hash)
...
+    !params.key?(:assign_room) && !params[:room_info].nil?
+  end
+
+  def do_not_assign_room
+    params.key?(:assign_room) && (!params[:assign_room] || params[:assign_room] == "false")
+  end
+
+  def select_available_room(hearing_count_by_room)
+    available_room = nil
+    (1..13).each do |n|
```
mdbenjam: Are these the room numbers? Can this live in a constant somewhere? Feels like a magic number.

mdbenjam: If so, maybe `room_number` is a better name than `n`

app/controllers/hearings/hearing_day_controller.rb
```diff
@@ -237,4 +245,57 @@ def format_for_client(json_hash)
...
+      room_count = hearing_count_by_room[n.to_s]
+      Rails.logger.info("room_count is: #{room_count}")
```
mdbenjam: If this is needed in production seems like we should add the room number as well?

app/controllers/users_controller.rb
```diff
@@ -6,6 +6,11 @@ def index
     when Constants::USER_ROLE_TYPES["attorney"]
       return render json: { attorneys: Judge.new(judge).attorneys } if params[:judge_css_id]
       return render json: { attorneys: Attorney.list_all }
+    when Constants::USER_ROLE_TYPES["hearing_coordinator"]
```
mdbenjam: Maybe not relevant right now. But as I mentioned  earlier Queue is moving towards maintaining our own user <-> team mappings, so at some point we might be able to replace this stuff with https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/organizations/users_controller.rb#L2

client/app/components/Checkbox.jsx
```diff
@@ -18,7 +18,8 @@ export default class Checkbox extends React.Component {
       errorMessage,
       unpadded,
       hideLabel,
-      styling
+      styling,
+      strongLabel
```
mdbenjam: I'm generally against adding new styling to components since it hurts the overall consistency in the look and feel of our app. Do we really need a strong label instead of a regular label? If so, can we just pass a label wrapped in a `<strong>` tag to the component?

client/app/components/Dropdown.jsx
mdbenjam: I think if we're not going to use this change, it's better to remove it. No need to have dead code around.

client/app/hearingSchedule/components/HearingDayAddModal.jsx
```diff
@@ -0,0 +1,317 @@
...
+
+  modalMessage = () => {
+    return <React.Fragment>
+      <div {...fullWidth} {...css({ marginBottom: '0' })} >
+        <p {...spanStyling} >Please select the details of the new hearing day </p>
+        <b {...titleStyling} >Select Hearing Date</b>
+        <DateSelector
+          name="hearingDate"
+          label={false}
+          errorMessage={((this.state.dateError && this.state.typeError) || (this.state.dateError)) ?
```
mdbenjam: I think `((this.state.dateError && this.state.typeError) || (this.state.dateError))` is equivalent to `this.state.dateError`

client/app/hearingSchedule/components/HearingDayAddModal.jsx
```diff
@@ -0,0 +1,317 @@
...
+    >Confirm
+    </Button>;
+  };
+
+  onClickConfirm = () => {
+    let errorMessages = [];
+    let roErrorMessages = [];
+
+    if (this.props.selectedHearingDay === '') {
+      this.setState({ dateError: true });
```
mdbenjam: Do we really need both a boolean `dateError` and to store the error text in a different part of state? Can we just put the error message into `dateError` and `nil` if there's no error? That way the error is truthy when set?

app/controllers/hearings/hearing_day_controller.rb
oscarAtNava: It should return true. I have a regression in this logic which I am going to investigate this morning. I noticed it Friday afternoon after I refactored some logic into methods to resolve a complexity error in the lint scripts.

client/app/components/Dropdown.jsx
oscarAtNava: It is not dead code. It puts a commonly used property in similar components  lacking in this component: the ability to bolden the label.

app/controllers/hearings/hearing_day_controller.rb
oscarAtNava: It is not needed, it will be removed as soon as I get the logic right.

client/app/components/Checkbox.jsx
oscarAtNava: Hmm, I see. I was not aware that the components already imbedded the style guide for the entire app. The strong label is necessary to satisfy the designer mocks. I can push back with the designers if you want me to remove the strong label. @sharonwarner curious to understand your opinion.

client/app/hearingSchedule/components/HearingDayAddModal.jsx
oscarAtNava: I want the error to show on top of the date field if there is a date error or both a date error and a type error. If I only include dateError in this line of code I will get the error message and title in both fields not satisfying the requirements.

client/app/hearingSchedule/components/HearingDayAddModal.jsx
oscarAtNava: I place them in an array to be able to format a ul with the appropriate list <li> items. I can show you the effect I am looking for and perhaps you can let me know if there is another way.

client/app/hearingSchedule/components/HearingDayAddModal.jsx
```diff
@@ -0,0 +1,317 @@
...
+    case 'V':
+      this.setState({ videoSelected: true });
+      this.setState({ centralOfficeSelected: false });
+      break;
+    case 'C':
+      this.setState({ videoSelected: false });
+      this.setState({ centralOfficeSelected: true });
+      break;
+    default:
+      this.setState({ videoSelected: false });
```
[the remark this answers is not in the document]
oscarAtNava: These two booleans are controlling the visibility of the regional office, vlj and coordinator dropdowns hence booleans. The default section happens if the user selects "" in the hearing type dropdown. That is when no selection has been made yet(ListScheduleComponent sets the dropdown to "" when user opens modal to ensure redux has the correct state).

client/app/hearingSchedule/components/HearingDayAddModal.jsx
```diff
@@ -0,0 +1,317 @@
...
+    this.resetErrorState();
+  };
+
+  resetErrorState = () => {
+    this.setState({ dateError: false });
+    this.setState({ typeError: false });
+    this.setState({ roError: false });
+  };
+
+  onVljChange = (value) => {
```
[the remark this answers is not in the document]
oscarAtNava: Thanks

client/app/hearingSchedule/components/HearingDayAddModal.jsx
```diff
@@ -0,0 +1,317 @@
...
+        {(this.state.videoSelected || this.state.centralOfficeSelected) &&
+        <SearchableDropdown
+          name="vlj"
+          label="Select VLJ (Optional)"
+          strongLabel
+          value={this.props.vlj}
+          onChange={this.onVljChange}
+          options={this.props.activeJudges} />
+        }
+        {(this.state.videoSelected || this.state.centralOfficeSelected) &&
```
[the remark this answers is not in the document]
oscarAtNava: I get an error in compilation when I include two components within conditional brackets. Is there some special syntax I need to use?

app/controllers/users_controller.rb
oscarAtNava: I did mention this to Meredith and Sharon. It will be part of a larger discussion I guess.

app/controllers/hearings/hearing_day_controller.rb
oscarAtNava: I will use a constant and improve the variable name.

client/app/hearingSchedule/components/HearingDayAddModal.jsx
mdbenjam: All I'm saying is given the truth table

| a        | b           | a && b \|\| a |
| ------------- |-------------| -----|
| true      | false | true |
| true      | true | true |
| false      | false | false |
| false      | true | false |

`a` has the same value as `a && b || a`

client/app/components/Checkbox.jsx
sharonwarner: I'm fine sending the label wrapped in a <strong> tag.

sharonwarner: Lol. I meant `<strong>`.

client/app/hearingSchedule/components/HearingDayEditModal.jsx
mdbenjam: Yeah I think the constants sharing has been a really great improvement on the Queue side.

client/app/hearingSchedule/components/ListSchedule.jsx
```diff
@@ -234,7 +235,7 @@ class ListSchedule extends React.Component {
                 data={hearingScheduleRows}
                 target="_blank"
                 filename={`HearingSchedule ${this.props.startDate}-${this.props.endDate}.csv`}>
-                Download current view
+                Download Current View
```
mdbenjam: Question for the designer, but I thought buttons were supposed to be sentence case across Caseflow.

client/app/hearingSchedule/components/HearingDayEditModal.jsx
```diff
@@ -0,0 +1,241 @@
...
+  onClickConfirm = () => {
+    this.props.closeModal();
+  };
+
+  modalCancelButton = () => {
+    return <Button linkStyling onClick={this.onCancelModal}>Go back</Button>;
+  };
+
+  onCancelModal = () => {
+    this.props.cancelModal();
```
mdbenjam: Why not just pass `this.props.cancelModal` to the button above. Same with the confirm button.

client/app/hearingSchedule/containers/DailyDocketContainer.jsx
```diff
@@ -72,7 +129,86 @@ export class DailyDocketContainer extends React.Component {
...
+    // find labels in options before passing values to modal
+    const coordinator = _.find(this.props.activeCoordinators, { label: this.props.dailyDocket.bvaPoc });
+
+    this.props.selectVlj(this.props.dailyDocket.judgeId);
+    this.props.selectHearingCoordinator(coordinator);
+    this.props.setNotes(this.props.dailyDocket.notes);
+    this.props.onHearingDayModified(false);
+  };
+
+  closeModal = () => {
```
mdbenjam: Could this code mostly live in the modal component itself? Let the modal handle sending the request, and show the error state. Then just call a simple `closeModal` function here that runs `this.setState({ modalOpen: false });` to close the modal? Might help pass less state back and forth.

spec/factories/vacols/staff.rb
```diff
@@ -31,6 +31,14 @@
       sactive "A"
     end
 
+    trait :hearing_coordinator do
+      stitle "HRG"
+      sactive "A"
+      snamel "Smith#{Random.rand(1..5)}"
```
mdbenjam: I'd use a sequence rather than random: https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#sequences

client/app/components/Checkbox.jsx
mdbenjam: Yeah I think ideally I'd like to hear from design about why this case needs to be bolded when our normal labels do not, but if we do need bold, passing it in as a tag seems 👍

client/app/components/Dropdown.jsx
mdbenjam: I would echo my comment from above, that I just don't think a `strongLabel` is a necessary prop. Folks should just pass in a label in a `<strong>` tag. Adding props that aren't used feels to me like it clutters our components.

client/app/hearingSchedule/components/DailyDocket.jsx
mdbenjam: Is this changed? Maybe you just haven't pushed the updates yet. Just don't want this to go forgotten.

client/app/hearingSchedule/components/HearingDayAddModal.jsx
mdbenjam: Gotcha, no that seems fine then.

client/app/hearingSchedule/components/HearingDayAddModal.jsx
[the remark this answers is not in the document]
mdbenjam: Gotcha, yeah I think I just get concerned when I see us needing to set two variables in every instance of the switch statement. Seems that the underlying state is `selectedHearingType`. But will defer to you if you prefer the two boolean approach.

Regardless, we can probably combine these into one setState call that changes both variables at once.

client/app/hearingSchedule/components/HearingDayAddModal.jsx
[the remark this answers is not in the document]
mdbenjam: Wrap them in a `<React.Fragment></React.Fragment>`

client/app/hearingSchedule/components/HearingDayAddModal.jsx
oscarAtNava: Fixed

client/app/components/Dropdown.jsx
oscarAtNava: Reverted my changes.

client/app/hearingSchedule/components/DailyDocket.jsx
oscarAtNava: Yes, these changes are in the branch now.

client/app/hearingSchedule/containers/ListScheduleContainer.jsx
oscarAtNava: Done

app/controllers/hearings/hearing_day_controller.rb
oscarAtNava: This is fixed now.

app/controllers/hearings/hearing_day_controller.rb
oscarAtNava: Fixed

client/app/hearingSchedule/components/HearingDayAddModal.jsx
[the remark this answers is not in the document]
oscarAtNava: Done

client/app/hearingSchedule/components/HearingDayAddModal.jsx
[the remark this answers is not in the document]
oscarAtNava: Done

client/app/hearingSchedule/components/ListSchedule.jsx
oscarAtNava: Reverted back to master here.

client/app/hearingSchedule/components/HearingDayEditModal.jsx
oscarAtNava: Done

client/app/hearingSchedule/containers/DailyDocketContainer.jsx
oscarAtNava: I could look at refactoring this in the future. I was under the impression we did not want modals making server calls as they may be long in duration. If you feel strongly about this please ask Sharon to open a refactor ticket.

client/app/hearingSchedule/containers/ListScheduleContainer.jsx
```diff
@@ -47,31 +73,173 @@ export class ListScheduleContainer extends React.Component {
...
+      let activeJudges = [];
+
+      _.forEach(resp.hearingJudges, (value) => {
+        activeJudges.push({
+          label: `${value.firstName} ${value.middleName} ${value.lastName}`,
+          value: value.vacolsAttorneyId
+        });
+      });
+
+      this.props.onReceiveJudges(_.orderBy(activeJudges, (judge) => judge.label, 'asc'));
```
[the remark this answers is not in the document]
oscarAtNava: We take this approach in many areas in Hearing Schedule. Perhaps this can be a refactor exercise to review all areas in HS where we sort in the UI instead of the datastore.

client/app/hearingSchedule/containers/ListScheduleContainer.jsx
```diff
@@ -47,31 +73,173 @@ export class ListScheduleContainer extends React.Component {
...
+          this.setState({ serverError: true });
+        }
+      });
+  };
+
+  cancelModal = () => {
+    this.setState({ modalOpen: false });
+  };
+
+  getAlertTitle = () => {
```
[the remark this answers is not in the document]
oscarAtNava: The alert has two props one for title and one for message. I use each separate function to set the appropriate prop. How would having one function work unless I put some sort of conditional logic?

spec/factories/vacols/staff.rb
oscarAtNava: Done

client/app/hearingSchedule/containers/DailyDocketContainer.jsx
sharonwarner: I don't think this is working still, I have to refresh the modal to see the updates.

client/app/hearingSchedule/containers/ListScheduleContainer.jsx
sharonwarner: I don't think this works as expected. In order for the added hearing to show up on the ListSchedule view, the user has to refresh the page.

client/app/hearingSchedule/containers/DailyDocketContainer.jsx
mdbenjam: Why does it matter if a modal makes a long running call?

local/vacols/VACOLS::Staff_dump.csv
```diff
@@ -221,7 +221,7 @@ KPROHASKA1,,,,,Keyshawn,C,Prohaska,KPROHASKA1,SBODE,,BDCOUNSEL,,,,,,,,,,,,,,,,,,
 CHERMISTO1,,,,,Cooper,W,Hermiston,CHERMISTO1,011W,,CHAIR,,,,,,,,,,,,,,,,,,,,,,,,,I,,,,,,,,,,,
 EZIEMANN,,,,,Edwin,R,Ziemann,EZIEMANN,DV-DIR,,AMC,,,,,,,,,,,,,,,,,,,,,,,,,A,,,,,,,,,,,BVAEZIEMANN
 KLUBOWITZ,,,,,Kiara,L,Lubowitz,KLUBOWITZ,AHUDSON,16-621,BDCOUNSEL,,,,,,,,,,,,WLANGOSH,1997-11-25,WLANGOSH,1997-11-25,0,0,0,0,,,,,,I,,,,,,,,621,J,,
-BSCHUMM,,,,,Broderick,A,Schumm,BSCHUMM,HRG,,CDLEGCLK,,,,,,,,,,(221) 483-1165 x044,,,,,,,,,,,,,,,I,,,,,,,,,,,
+BSCHUMM,,,,,Broderick,A,Schumm,BSCHUMM,HRG,,CDLEGCLK,,,,,,,,,,(221) 483-1165 x044,,,,,,,,,,,,,,,A,,,,,,,,,,,BVABSCHUMM
```
mdbenjam: Do these changes correspond to actual changes you made in the VACOLS copy in RDS?

app/controllers/hearings/hearing_day_controller.rb
mdbenjam: Is the logic here correct now?

client/app/hearingSchedule/containers/ListScheduleContainer.jsx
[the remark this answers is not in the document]
mdbenjam: Something like

```
getAlertMessage = () => {
    if (this.state.serverError) {
      return {title: 'An Error Occurred', message: 'You are unable to complete this action.'};
    } else if () { ... }
}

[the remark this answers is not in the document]
oscarAtNava: And you place that function in which prop in the component? Or I put it at the end? Not familiar with that syntax.

local/vacols/VACOLS::Staff_dump.csv
oscarAtNava: No, it is just mock data I am setting up so coordinators show up in FACOLS for anyone testing this section. Not sure what I will find in UAT copy of VACOLS.

client/app/hearingSchedule/containers/ListScheduleContainer.jsx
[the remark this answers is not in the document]
oscarAtNava: This seems to me like a style issue. You really want me to spend the time to refactor this before I merge all the rest of the functionality?

local/vacols/VACOLS::Staff_dump.csv
oscarAtNava: The file has been reverted.

client/app/hearingSchedule/containers/DailyDocketContainer.jsx
oscarAtNava: No spinner?

client/app/hearingSchedule/containers/ListScheduleContainer.jsx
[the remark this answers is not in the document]
mdbenjam: The way it's written now has an implicit parallel structure. Both methods need to have the same set of if-statements aligned. To me that feels a little brittle. I agree at some level it's a matter of style, and I'm okay with you leaving it as is. That said, I disagree with the assertion that reviewers should feel pressure to avoid commenting on details just because a PR is large. This is why I like to avoid PRs this big, it makes it much more difficult to effectively review it.

client/app/hearingSchedule/containers/DailyDocketContainer.jsx
mdbenjam: I see, yeah we usually just use the submit button's loading state in Queue, but if that's not the pattern in HS then this seems reasonable.

oscarAtNava:
I made the changes to update redux after add and edit.

PR #8101, opened 2018-12-04
Schedule Veterans on Action Click

### LINKED ISSUE #8051: Create Schedule Veteran Task on CaseDetails and not from AssignHearing
Previously, schedule veteran tasks where created when a user navigates from the AssignHearings page to the CaseDetail Page. 

In order to allow a Schedule Veteran Task to be created no matter how a user navigates to the CaseDetail page, the task should be created on the CaseDetail page when the user clicks "Schedule a Veteran"

lomaxap:
Resolves #8051 

### Description
- Moves adding schedule and completing schedule task creation to AssignHearingsModal.jsx
- Adds ScheduleVeteran as action option to RootTask on legacy appeals
- Creates RootTask for veteran on CaseDetailsView (on `task_controller#for_appeal`) on page load if one does not exist

- fixes string / object value inconsistencies on HearingDayDropdown and RoSelector
- consistent naming of HearingDay vs. hearing_date


lomaxap:
@lowellrex one thing I need to fix before this is merge: when we create a RootTask for the first time on the appeals_controller, it does not return the task / actions on first page load. You have to refresh the page to see it.

app/controllers/appeals_controller.rb
```diff
@@ -53,7 +53,9 @@ def show
     no_cache
     respond_to do |format|
       format.html { render template: "queue/index" }
+
       format.json do
+        RootTask.find_or_create_by!(appeal: appeal)
```
lowellrex: I don't think it matters much because of how we call this (we make the API request for the JSON response immediately after we request the HTML response whenever we request the HTML page), but it might make sense to move this outside the `respond_to` block so that we always call this function.

client/app/queue/components/AssignHearingModal.jsx
```diff
@@ -73,6 +76,103 @@ const centralOfficeStaticEntry = [{
...
+          }
+        }
+      }
+    };
+
+    return this.props.requestPatch(`/tasks/${scheduleHearingTask.taskId}`, payload, this.getSuccessMsg()).
+      then((resp) => {
+        const response = JSON.parse(resp.text);
+
+        // Review with team to see why this is failing.
```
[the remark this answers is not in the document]
lomaxap: It just updates the Redux store with the latest tasks. Kept it from Oscar's version, but it does seem unnecessary since going back to CaseDetailsView refetches all the data.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -21,7 +21,12 @@ def update_from_params(params, current_user)
       Time.zone.parse(hearing_date)
     end
     task_payloads[:values][:hearing_date] = new_date
-    task_business_payloads.update(task_payloads)
+
+    if !task_business_payloads.empty?
```
sharonwarner: Do we will need the payloads here? I thought we were going to get rid of those since we switched to the query params.

lomaxap: was concerned it might be possible that there are still tasks that were created the old way, so a payload would exist

app/models/tasks/root_task.rb
```diff
@@ -7,11 +7,23 @@ def set_assignee
...
+    []
   end
 
   def no_actions_available?(user)
+    if HearingsManagement.singleton.user_has_access?(user) && legacy?
+      hearing_task = children.find { |task| task.is_a?(ScheduleHearingTask) }
+      return hearing_task.status == Constants.TASK_STATUSES.completed if hearing_task
+      return false
+    end
+
```
lowellrex: I think it might be cleaner to encapsulate all of this logic in `available_actions`. Maybe something like the following:
```ruby
...
if HearingsManagement.singleton.user_has_access?(user) && legacy? && children.select { |t| t.is_a?(ScheduleHearingTask) && t.status == Constants.TASK_STATUSES.completed }.empty?
  return [Constants.TASK_ACTIONS.SCHEDULE_VETERAN.to_h]
end
...
```

And `no_actions_available?` could be simply:
```ruby
def no_actions_available?(_user)
  completed?
end
```

lomaxap: beautiful

spec/controllers/tasks_controller_spec.rb
```diff
@@ -664,7 +664,7 @@
         expect(task["attributes"]["assigned_to"]["css_id"]).to eq vso_user.css_id
         expect(task["attributes"]["appeal_id"]).to eq appeal.id
 
-        expect(appeal.tasks.count).to eq 2
+        expect(appeal.tasks.count).to eq 3
```
lowellrex: The number of tasks we expect has increased by 1 because we are creating `RootTask` when we make this request and we weren't before? Is that correct?

lomaxap: Yes

Issue #7863, opened 2018-11-15
[Tech spec] Allow TaskTable to accept tabs


lowellrex:
## Summary

I'm interested in moving the logic for displaying tabs inside the `TaskTable` component itself. A quick sketch of how this could look [using `ColocatedTaskListView` as an example](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/ColocatedTaskListView.jsx#L50) is below.

**Before**
```jsx
const NewTasksTab = connect(
  (state) => ({ tasks: newTasksByAssigneeCssIdSelector(state) }))(
  (props) => {
    return <React.Fragment>
      <p className="cf-margin-top-0">{COPY.COLOCATED_QUEUE_PAGE_NEW_TASKS_DESCRIPTION}</p>
      <TaskTable
        includeDetailsLink
        includeTask
        includeType
        includeDocketNumber
        includeDaysWaiting
        includeReaderLink
        tasks={props.tasks}
      />
    </React.Fragment>;
  });

...

class Container extends React.PureComponent {
  render = () => {
    const tabs = [
      {
        label: sprintf(COPY.COLOCATED_QUEUE_PAGE_NEW_TAB_TITLE, numNewTasks),
        page: <NewTasksTab />
      },
      ...
    ];

    return <TabWindow tabs={tabs} />;
  }
}
```

**After**
```jsx
const NewTasksTab = <TaskTableTab
  customColumns={[detailsLink(), taskLabel(), taskType(), docketNumber(), daysWaiting(), readerLink()]}
  filterFunction={(task) => !taskIsOnHold(task)}
  description={COPY.COLOCATED_QUEUE_PAGE_NEW_TASKS_DESCRIPTION}
  title={COPY.COLOCATED_QUEUE_PAGE_NEW_TAB_TITLE}
  titleHasCount
/>

...

class Container extends React.PureComponent {
  render = () => {
    const tabs = [ <NewTasksTab />, ... ];

    return <TaskTable tabs={tabs} allTasks={this.props.tasksByAssigneeCssIdSelector} />;
  }
}
```

## Meat and potatoes

[...]

PR #8111, opened 2018-12-05
[WIP] non-comp queue

mdbenjam:

### Description

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...

### DIFF
--- a/app/controllers/appeals_controller.rb
+++ b/app/controllers/appeals_controller.rb
@@ -68,7 +68,8 @@ class AppealsController < ApplicationController
   helper_method :appeal, :url_appeal_uuid
 
   def appeal
-    @appeal ||= Appeal.find_appeal_by_id_or_find_or_create_legacy_appeal_by_vacols_id(params[:appeal_id])
+    @appeal = Appeal.find_appeal_by_id_or_find_or_create_legacy_appeal_by_vacols_id(params[:appeal_id]) if params[:appeal_type] == "Appeal" || params[:appeal_type] == "LegacyAppeal"
+    @appeal = HigherLevelReview.find(params[:appeal_id]) if params[:appeal_type] == "HigherLevelReview"
   end
 
   def url_appeal_uuid
--- a/app/models/claim_review.rb
+++ b/app/models/claim_review.rb
@@ -109,7 +109,7 @@ class ClaimReview < DecisionReview
   end
 
   def external_id
-    id
+    id.to_s
   end
 
   private
--- a/app/models/claim_review.rb
+++ b/app/models/claim_review.rb
@@ -108,6 +108,10 @@ class ClaimReview < DecisionReview
     request_issues.find { |reqi| reqi.description == description }
   end
 
+  def external_id
+    id
+  end
+
   private
 
   def informal_conference?
--- a/client/app/queue/CaseDetailsView.jsx
+++ b/client/app/queue/CaseDetailsView.jsx
@@ -21,6 +21,8 @@ import { resetErrorMessages, resetSuccessMessages, setHearingDay } from './uiRed
 import { CaseTimeline } from './CaseTimeline';
 import { getQueryParams } from '../util/QueryParamsUtil';
 
+import HigherLevelReviewDetailsView from './HigherLevelReviewDetailsView';
+
 import { CATEGORIES, TASK_ACTIONS } from './constants';
 import { COLORS } from '../constants/AppConstants';
 
@@ -56,10 +58,15 @@ class CaseDetailsView extends React.PureComponent {
     const {
       appealId,
       appeal,
+      appealType,
       error,
       success,
       featureToggles
     } = this.props;
+
+    if (appealType === 'HigherLevelReview') {
+      return <HigherLevelReviewDetailsView appealId={appealId} appealType={appealType} />;
+    }
 
     return <AppSegment filledBackground>
       <CaseTitle appeal={appeal} appealId={appealId} redirectUrl={window.location.pathname} />
--- a/client/app/queue/CaseSnapshot.jsx
+++ b/client/app/queue/CaseSnapshot.jsx
@@ -298,7 +298,7 @@ export class CaseSnapshot extends React.PureComponent<Props> {
       {this.showActionsSection() &&
         <div className="usa-width-one-half">
           <h3>{COPY.CASE_SNAPSHOT_ACTION_BOX_TITLE}</h3>
-          <ActionsDropdown task={primaryTask} appealId={appeal.externalId} />
+          <ActionsDropdown task={primaryTask} appeal={appeal} />
         </div>
       }
     </div>;
@@ -309,7 +309,7 @@ const mapStateToProps = (state: State, ownProps: Params) => {
   const { featureToggles, userRole, canEditAod } = state.ui;
 
   return {
-    appeal: appealWithDetailSelector(state, { appealId: ownProps.appealId }),
+    appeal: appealWithDetailSelector(state, { appealId: ownProps.appealId, appealType: ownProps.appealType }),
     featureToggles,
     userRole,
     primaryTask: actionableTasksForAppeal(state, { appealId: ownProps.appealId })[0],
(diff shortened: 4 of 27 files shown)


app/controllers/appeals_controller.rb
```diff
@@ -68,7 +68,8 @@ def show
   helper_method :appeal, :url_appeal_uuid
 
   def appeal
-    @appeal ||= Appeal.find_appeal_by_id_or_find_or_create_legacy_appeal_by_vacols_id(params[:appeal_id])
+    @appeal = Appeal.find_appeal_by_id_or_find_or_create_legacy_appeal_by_vacols_id(params[:appeal_id]) if params[:appeal_type] == "Appeal" || params[:appeal_type] == "LegacyAppeal"
```
mdbenjam: Need to abstract this somewhere, we'll need to send `appeal_type` everywhere.

app/models/claim_review.rb
```diff
@@ -108,6 +108,10 @@ def find_request_issue_by_description(description)
     request_issues.find { |reqi| reqi.description == description }
   end
 
+  def external_id
```
mdbenjam: Thinking we might need a better name for external_id

client/app/queue/CaseDetailsView.jsx
```diff
@@ -56,11 +58,16 @@ class CaseDetailsView extends React.PureComponent {
     const {
       appealId,
       appeal,
+      appealType,
       error,
       success,
       featureToggles
     } = this.props;
 
+    if (appealType === 'HigherLevelReview') {
```
mdbenjam: This component can become a wrapper for all of the "detailed versions"

client/app/queue/CaseSnapshot.jsx
```diff
@@ -309,7 +309,7 @@ const mapStateToProps = (state: State, ownProps: Params) => {
   const { featureToggles, userRole, canEditAod } = state.ui;
 
   return {
-    appeal: appealWithDetailSelector(state, { appealId: ownProps.appealId }),
+    appeal: appealWithDetailSelector(state, { appealId: ownProps.appealId, appealType: ownProps.appealType }),
```
mdbenjam: We should be passing appeal down, not appealId/type and then going to the redux store. We should be able to make our components more reusable if they don't need to go to the store.

client/app/queue/HigherLevelReviewDetailsView.jsx
```diff
@@ -0,0 +1,85 @@
+import { css } from 'glamor';
```
mdbenjam: This file would belong to Intake.

client/app/queue/NonCompOrganizationQueue.jsx
```diff
@@ -0,0 +1,109 @@
+import React from 'react';
```
mdbenjam: This file would belong to Intake.

client/app/queue/OrganizationQueue.jsx
```diff
@@ -26,12 +28,20 @@ const containerStyles = css({
   position: 'relative'
 });
 
+const NON_COMP_ORGANIZATIONS = ['nca'];
```
mdbenjam: Need to think of a better way to fork this.

client/app/queue/QueueActions.js
```diff
@@ -19,24 +19,26 @@ import type {
 } from './types/models';
 
 export const onReceiveQueue = (
-  { tasks, amaTasks, appeals }:
+  { tasks, amaTasks, appeals, higherLevelReviews }:
```
mdbenjam: Want to think of a more generic way to do this.

client/app/queue/QueueApp.jsx
```diff
@@ -276,36 +281,36 @@ class QueueApp extends React.PureComponent<Props> {
             title="Unassigned Cases | Caseflow"
             render={this.routedJudgeQueueList('assign')} />
           <Route
-            path="/queue/appeals/:appealId/modal/advanced_on_docket_motion"
+            path="/queue/:appealType/:appealId/modal/advanced_on_docket_motion"
```
mdbenjam: One of the big changes would be updating the URLs for the case details pages to be `queue/:appealType/:appealId`

client/app/queue/components/TaskTable.jsx
```diff
@@ -272,8 +273,30 @@ class TaskTable extends React.PureComponent<Props> {
     } : null;
   }
 
+  claimantLinkColumn = () => {
```
mdbenjam: New columns are defined here. Lowell has a good idea for a refactor here that will make it easier to place columns in an arbitrary order.

client/app/queue/reducers.js
```diff
@@ -22,6 +22,7 @@ export const initialState = {
   amaTasks: {},
   appeals: {},
   appealDetails: {},
+  higherLevelReviews: {},
```
mdbenjam: higher level reviews and supplement claims can be top level objects in our redux store.

client/app/queue/selectors.js
```diff
@@ -58,8 +60,8 @@ export const getActiveModalType = createSelector(
 );
 
 export const tasksWithAppealSelector = createSelector(
-  [getTasks, getAmaTasks, getAppeals, getAppealDetails],
-  (tasks: Tasks, amaTasks: Tasks, appeals: BasicAppeals, appealDetails: AppealDetails) : Array<TaskWithAppeal> => {
+  [getTasks, getAmaTasks, getAppeals, getAppealDetails, getHigherLevelReviews],
```
mdbenjam: Some of our selectors need to change to pull the necessary data.

client/app/queue/utils.js
```diff
@@ -99,7 +99,32 @@ const extractAppealsFromTasks =
   (tasks: Array<Object>):
     BasicAppeals => {
     return tasks.reduce((accumulator, task) => {
-      if (!accumulator[task.attributes.external_appeal_id]) {
+      if (!accumulator[task.attributes.external_appeal_id] && (task.attributes.appeal_type === 'Appeal' || task.attributes.appeal_type === 'LegacyAppeal')) {
```
mdbenjam: Need to think of a more elegant way to do this.

jiahuang:
@mdbenjam I'm looking into what it takes to do the frontend with a new set of reducers. Looks like the majority of the page is the table. My current proposal for reusing TaskTable is:

In order for TaskTable to be reusable, it needs to be decoupled with any particular state. Kind of problematic to do it right now since a bunch of things in queue rely on the current implementation. Easiest proposal is to 

1. add another export in TaskTable that exports the component without bindings. 
2. add a function in task table props that allows for declaration of columns outside of TaskTable.jsx. Maybe take in something like
```
additionalColumns: [
  {
     header: 
     span:
     valueFunction: () => {}
     getSortValue:   },
  {
     header: 
     span:
     valueFunction: () => {}
     getSortValue   }...
]
```
3. If additionalColumns is present, properties within it would also get compacted into getQueueColumns. Can also just add a switch on the current columns compacted vs new additionalColumns param if current code execution would be problematic with the props being extracted out.

I think this approach would leave keep queue functionality while allowing custom modifications of the TaskTable. In the future we probably want to extract out the current column logic into separate functions that are imported into each TaskTable as needed.

mdbenjam:
> In the future we probably want to extract out the current column logic into separate functions that are imported into each TaskTable as needed.

@lowellrex and I have talked about this and are very much in favor of it! If you want to just make that change I think that'd be 👍 . I don't think it would be a large lift, and would just keep things cleaner. Otherwise, the plan you've laid out with having some of our own functions passed in, and some of the current column-props passed in seems fine.

lowellrex:
Here is a ticket I whipped up with general thoughts on [how to refactor the `TaskTable` component](https://github.com/department-of-veterans-affairs/caseflow/issues/7863). The second bullet there specifically calls out passing columns into the component (and is maybe the only refactor in that ticket worth doing in the near term).

PR #8115, opened 2018-12-05
Add methods to find all AOD and all non-AOD appeals

amprokop:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/8201

### Testing Plan
- Unit tests cover the major categories


app/models/appeal.rb
```diff
@@ -16,8 +16,38 @@ class Appeal < DecisionReview
...
+  scope :not_aod, lambda {
+    joins(claimants: :person)
+      .joins("LEFT OUTER JOIN advance_on_docket_motions on advance_on_docket_motions.person_id = people.id")
+      .where("people.date_of_birth >= ?", 75.years.ago)
+      .where("advance_on_docket_motions.id IS NULL")
+  }
+
   UUID_REGEX = /^\h{8}-\h{4}-\h{4}-\h{4}-\h{12}$/
 
+  def all_aod
```
amprokop: Unfortunately, you can't string two `.joins` together with an `.or` clause in ActiveRecord, it throws an error.

https://github.com/rails/rails/issues/24055

Could probably work on getting more bare metal here and make it work. But maybe let's leave it for now.

app/models/person.rb
```diff
@@ -15,12 +16,12 @@ def date_of_birth
     end
   end
 
-  private
-
   def advanced_on_docket_based_on_age
-    date_of_birth && date_of_birth < 75.years.ago
+    date_of_birth && date_of_birth <= 75.years.ago
```
amprokop: @mdbenjam, does this change look right to you?

app/models/appeal.rb
```diff
@@ -16,8 +16,38 @@ class Appeal < DecisionReview
...
+      .where("advance_on_docket_motions.created_at > appeals.established_at")
+  }
+  scope :aod_due_to_age, lambda {
+    joins(claimants: :person)
+      .where("people.date_of_birth < ?", 75.years.ago)
+  }
+
+  scope :not_aod, lambda {
+    joins(claimants: :person)
+      .joins("LEFT OUTER JOIN advance_on_docket_motions on advance_on_docket_motions.person_id = people.id")
```
amprokop: For unknown reasons, activerecord doesn't allow me to chain `.joins` with `.left_outer_joins`.

app/models/claimant.rb
```diff
@@ -2,7 +2,7 @@ class Claimant < ApplicationRecord
   include AssociatedBgsRecord
 
   belongs_to :review_request, polymorphic: true
-
+  belongs_to :person, primary_key: :participant_id, foreign_key: :participant_id
```
amprokop: @mdbenjam — does this look right to you?

app/models/person.rb
mdbenjam: That seems fine, but probably won't change much since 75.years.ago has a time associated with it.

```
[11] pry(main)> 75.years.ago
=> Sun, 05 Dec 1943 21:14:02 UTC +00:00
```

So I think this change gives them AOD a second earlier.

amprokop: **technically** right is the best kind of right. lol thanks for checking.

app/models/appeal.rb
```diff
@@ -16,6 +16,31 @@ class Appeal < DecisionReview
...
+      .where("advance_on_docket_motions.granted = ?", true)
+  }
+  scope :aod_due_to_age, lambda {
+    joins(claimants: :person)
+      .where("people.date_of_birth <= ?", 75.years.ago)
+  }
+
+  scope :all_priority, lambda {
+    # TODO: add cavc appeals
+    [aod_motions + aod_due_to_age].flatten.uniq(&:id)
```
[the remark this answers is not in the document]
amprokop: Ah I see you were suggesting combining the joins. The reason I didn't do that initially is that `joins` does an inner join, and for finding AOD due to age that doesn't work since we'd be excluding appeal rows without any `AdvanceOnDocketMotions`.

However, I *think* we could still combine the two if we do a left outer join to advance on docket motions, then group as you are suggesting? I can play around with this if you prefer.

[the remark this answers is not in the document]
cmgiven: Yes, that's what I was thinking. I think scopes will be much more useful for us than arrays of appeals. We want to be able to combine them (the intersection of active, ready, priority, and direct docket, e.g.) and only retrieve appeals up to a limit. We want to avoid instantiating appeals that we don't need.

In order to make that possible, we may need to unify the base join across all of these scopes so that they are compatible. But maybe that's not necessary, I don't know.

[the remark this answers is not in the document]
amprokop:
```
Ah, but that would potentially result in multiple records, wouldn't it? This probably needs a .group() thing like the nonpriority scope, yes?
```

Are you asking about whether the use of `.or` might cause duplicates in the returned Appeals array? I think it won't. See:

```
[8] pry(main)> Appeal.where(receipt_date: "10/12/2018").length
[2018-12-12 09:31:43 -0500]   Appeal Load (5.4ms)  SELECT "appeals".* FROM "appeals" WHERE "appeals"."receipt_date" = $1  [["receipt_date", "2018-12-10"]]
=> 6
[9] pry(main)> Appeal.where( veteran_file_number: "500000003").length
[2018-12-12 09:31:49 -0500]   Appeal Load (1.1ms)  SELECT "appeals".* FROM "appeals" WHERE "appeals"."veteran_file_number" = $1  [["veteran_file_number", "500000003"]]
=> 1
[10] pry(main)> Appeal.where(receipt_date: "10/12/2018").or(Appeal.where( veteran_file_number: "500000003")).length
[2018-12-12 09:31:53 -0500]   Appeal Load (1.0ms)  SELECT "appeals".* FROM "appeals" WHERE ("appeals"."receipt_date" = $1 OR "appeals"."veteran_file_number" = $2)  [["receipt_date", "2018-12-10"], ["veteran_file_number", "500000003"]]
=> 6
[11] pry(main)> 
```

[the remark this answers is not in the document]
amprokop: @cmgiven — made the change. Looks like it works fine. Thanks for the input.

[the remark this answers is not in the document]
cmgiven: :+1:

PR #8116, opened 2018-12-05
remove now-obsolete JudgeTask action references

### LINKED ISSUE #8033: Create subclasses of JudgeTask for two JudgeTask actions
In order to [remove the `action` column](https://github.com/department-of-veterans-affairs/caseflow/pull/7686) from the `tasks` table, we need to move that information into the various Task models' `label` methods. In order to support Task classes with multiple actions for a single class we will need to split those into separate classes. This ticket exists to do just that for `JudgeTask`'s `assign` and `review` actions (renamed `JudgeAssignTask` and `JudgeReviewTask` perhaps?).

This will probably have to be a multi-part release so that we maintain support for existing `JudgeTask`s. I'd imagine something like the following:
- [x] Stop creating plain `JudgeTask`s (create either of the two subclasses instead).
- [x] Transform existing `JudgeTask`s in the database to the appropriate subclass.
- [x] Remove all logic in `JudgeTask` that relies on the `action` property (should be handled by subclasses at this point).

## Acceptance criteria
* `JudgeTask`s no longer rely on the `tasks.action` database column

tomas-nava:
Resolves #8033

### Description
This is the last step in #8033


app/controllers/idt/api/v1/appeals_controller.rb
```diff
@@ -54,7 +54,7 @@ def tasks_assigned_to_user
     if feature_enabled?(:idt_ama_appeals)
       tasks += Task.where(assigned_to: user).where.not(status: [:completed, :on_hold])
     end
-    tasks.reject { |task| task.action == "assign" || task.is_a?(JudgeAssignTask) }
```
amprokop: Hmm... how will this affect VACOLS tasks for legacy appeals? Not sure but I thought they had `action = "assign"`. Want to make sure we're still filtering out legacy appeal tasks correctly.

app/models/distribution.rb
```diff
@@ -72,7 +72,7 @@ def validate_judge_has_no_pending_distributions
   def judge_has_no_unassigned_cases
     pending_statuses = [Constants.TASK_STATUSES.assigned, Constants.TASK_STATUSES.in_progress]
     assigned_tasks = judge.tasks.select do |t|
-      ((t.is_a?(JudgeTask) && t.action == "assign") || t.is_a?(JudgeAssignTask)) && pending_statuses.include?(t.status)
```
amprokop: Same question as above — are legacy appeal judge tasks `JudgeAssignTask`s too?

app/controllers/idt/api/v1/appeals_controller.rb
lowellrex: Great catch! Since this set of `tasks` includes both VACOLS-stored `JudgeLegacyTask`s and caseflow-stored `JudgeTask`s, we definitely have to allow for either the action to be set to assign (to support `JudgeLegacyTask`s) in addition to `task.is_a?(JudgeAssignTask)`.

app/models/distribution.rb
lowellrex: Yup, right again! I think the corrected version of this line should read:
```ruby
(t.action == "assign" || t.is_a?(JudgeAssignTask)) && pending_statuses.include?(t.status)
```

PR #8120, opened 2018-12-06
Subclass transient BGSSyncErrors

kencheeto:
### Description
We get a lot of transient errors during end product establishment EP sync, which is polluting the alerts channel. I've been collecting which errors are transient or expected due to routine BGS maintenance, and a subclass of BGSSyncError will help treat these errors differently within our error reporting tools.

Note that this PR alone does not increase/decrease the number of reported errors, just creates a subclass so we can control how they report/are consumed.

### Acceptance Criteria
- Transient errors during BGS EP sync are reported separately in Sentry


app/models/end_product_establishment.rb
```diff
@@ -7,36 +7,85 @@
...
+        #  "Error: 504 for url http://localhost:10001/BenefitClaimServiceBean/BenefitClaimWebService?WSDL"
+        #
+        # Example: https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2928/
+        TransientBGSSyncError.new(error, epe)
+      else
+        new(error, epe)
+      end
+    end
+  end
+  class TransientBGSSyncError < BGSSyncError; end
```
pkarman: why define this here, rather than with all the other exception classes above?

kencheeto: I can't subclass `BGSSyncError` before I've defined it, but I can reference `TransientBGSSyncError` in a `BGSSyncError` method.

PR #8123, opened 2018-12-06
Allow attorneys and judges to update benefit type for decision issues

aroltsch:
Allow attorneys and judges to update benefit type for decision issues


client/app/queue/SelectDispositionsView.jsx
```diff
@@ -270,14 +270,16 @@ class SelectDispositionsView extends React.PureComponent {
         <br />
         <h3>{COPY.DECISION_ISSUE_MODAL_BENEFIT_TYPE}</h3>
         <SearchableDropdown
-          name="Benefit type dropdown"
+          name="Benefit type"
+          placeholder={COPY.DECISION_ISSUE_MODAL_BENEFIT_TYPE}
+          hideLabel
```
aroltsch: If `hideLabel` is not passed, it will display `name`

Issue #7545, opened 2018-10-25
AHL | Dev Implementation


sharonwarner:
Because the VETS360 and Facility Locator APIs are only available with the proper certificates, we'll need to have a faked dev mode version of this functionality. We currently do this with our vbms_service and our bgs_service. 

We should create two files called locator service, one in lib/fakes and one in app/services/external_api. For now, the external_api file can just have one function that returns an empty json object. The fakes file should contain one function that returns a json object with the veteran's RO's alternate hearing locations and a randomly generated distance to each one.

```
{
name: "Berks County VA Clinic",
address_line1: "2762 Century Boulevard",
address_line2: "Ridgeview Professional Center",
address_line3: "Suite 120",
city: "Wyomissing",
state: "PA",
zip: "19610-3345" 
}
```

We should also create an initializer in config/initializers so that we use the fake environment in dev mode and the real environment in production.

PR #8127, opened 2018-12-06
Alternate Hearing Location ExternalAPIs Setup

lomaxap:
First part of #7545 

### Description
- basic setup for interacting with Vets360 and FacilitiesLocator endpoints


lib/fakes/facilities_locator_service.rb
```diff
@@ -0,0 +1,35 @@
+class Fakes::FacilitiesLocatorService < ExternalApi::FacilitiesLocatorService
+
+  def self.get_nearest(point, ids)
```
sharonwarner: I think we should name this get_distances since it's going to return all the alternate hearing locations, not just the nearest one.

PR #8130, opened 2018-12-06
Make an attempt to fix a judge checkout flake

aroltsch:
Based on the screenshot from the build: https://24782-51449239-gh.circle-artifacts.com/1/home/circleci/project/tmp/capybara/screenshot_2018-12-03-20-23-55.880.png
a required field was never clicked.

### DIFF
--- a/spec/feature/queue/checkout_flows_spec.rb
+++ b/spec/feature/queue/checkout_flows_spec.rb
@@ -571,8 +571,8 @@ RSpec.feature "Checkout flows" do
       find("label", text: Constants::JUDGE_CASE_REVIEW_OPTIONS["COMPLEXITY"]["easy"]).click
       find("label", text: "1 - #{Constants::JUDGE_CASE_REVIEW_OPTIONS['QUALITY']['does_not_meet_expectations']}").click
 
-      # areas of improvement
-      find("#issues_are_not_addressed", visible: false).sibling("label").click
+      expect(page).to have_content("Please provide more details about quality of work")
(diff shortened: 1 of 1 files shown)


spec/feature/queue/checkout_flows_spec.rb
```diff
@@ -641,15 +640,10 @@ def select_issue_level_options(opts)
         expect(page).to have_content("One Touch Initiative")
         find("label", text: COPY::JUDGE_EVALUATE_DECISION_CASE_ONE_TOUCH_INITIATIVE_SUBHEAD).click
 
-        click_on "Continue"
-        sleep 1
-
-        expect(page).to have_content("Choose one")
```
aroltsch: Clicking `Continue` without selecting a required option causes FE to display an error. Then clicking the radio button never succeeds - I don't know why! Removing lines 644-647 makes a test succeed on each run.

mdbenjam:
Hmmm looking at the original failures it seems like they're failing somewhere else: https://25073-51449239-gh.circle-artifacts.com/1/home/circleci/project/tmp/capybara/screenshot_2018-12-04-23-45-58.717.png. This is the line that fails: https://github.com/department-of-veterans-affairs/caseflow/blob/master/spec/feature/queue/checkout_flows_spec.rb#L661 which I think means from the screenshot that `Continue` was hit but no area of improvement was selected. But I can't figure out why `find("#issues_are_not_addressed", visible: false).sibling("label").click` isn't successfully clicking on the checkbox sometimes... 🤷‍♂️ . Let's chat about this more on Monday.

aroltsch:
Yup, at first I saw that the failure was somewhere else but I assumed it is still connected to the fact that we hit Continue without selecting required fields. Also from the screenshot above, we cannot tell if two other radio buttons were selected, right?

PR #8132, opened 2018-12-06
use approx_decision_date

jiahuang:
Fixes bug when decision_issue.profile_date is not set


jiahuang:
closing in favor of #8136

PR #8134, opened 2018-12-06
populate fake ratings with unique issue ids

kencheeto:
### Description
Generator code was producing the same nonunique reference ids, which were showing up in a buggy Intake local env:

demo

### DIFF
--- a/lib/generators/rating.rb
+++ b/lib/generators/rating.rb
@@ -103,7 +103,7 @@ class Generators::Rating
     def populate_issue_ids(attrs)
       return unless attrs[:issues]
       # gives a unique id to each issue that is tied to a specific participant_id
-      attrs[:issues].each_with_index.map do |issue, i|
+      attrs[:issues].map do |issue|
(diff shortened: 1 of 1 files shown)


lib/generators/rating.rb
```diff
@@ -100,16 +100,11 @@ def generate_profile_datetime(participant_id)
...
-    def get_prev_issues_count(participant_id)
-      Fakes::BGSService.rating_profile_records[participant_id].values.flatten(1).count
-    end
-
     def populate_issue_ids(attrs)
       return unless attrs[:issues]
       # gives a unique id to each issue that is tied to a specific participant_id
-      prev_count = get_prev_issues_count(attrs[:participant_id])
       attrs[:issues].each_with_index.map do |issue, i|
-        issue[:reference_id] ||= "#{attrs[:participant_id]}#{prev_count + i}"
```
kencheeto: alternatively could have made this `prev_count++`

PR #8135, opened 2018-12-07
Implement edit confirmation page for HLRs and SCs

kencheeto:
Connects #7313 

### Description
Editing a claim review now shows a more informative confirmation page akin to the intake confirmation page. The styling is similar, but the content is a bit different - in addition to established EPs, we also want to show when an EP is being canceled (all request issues removed) or updated (request issues changed).

In order to accomplish this, the update request now returns the issues before and after the update, and the frontend sorts out which EPs are created/updated/canceled.

This PR also establishes an example for writing feature tests that can be more tightly scoped:
* tests common to claim reviews is DRY without losing coverage
* tests specific to a kind of decision review are still simple to write
* any top level setup is generic, minimal, and overridable

screen shot 2018-12-24 at 12 49 51

### Acceptance Criteria
- Updating a HLR or SC leads to a new confirmation page
- Updating an appeal keeps the existing behavior (route back to the appeal sans /edit)
- The new confirmation page shows when a rating/nonrating EP is being established/canceled/updated

#### Claim reviews
1. Intake a HLR and SC with one issue
1. Edit the claim review to add an EP and remove the existing EP
1. Check the confirmation page
1. Edit the claim review to update the existing EP
1. Check the confirmation page

#### Appeals
1. Intake an appeal with one issue
1. Edit the appeal with any valid change
1. Check that you are redirected to the appeal

### DIFF
--- a/client/app/intake/components/IneligibleIssuesList.jsx
+++ b/client/app/intake/components/IneligibleIssuesList.jsx
@@ -3,7 +3,7 @@ import INELIGIBLE_REQUEST_ISSUES from '../../../constants/INELIGIBLE_REQUEST_ISS
 
 const ineligibilityCopy = (issue) => {
   if (issue.titleOfActiveReview) {
-    return INELIGIBLE_REQUEST_ISSUES.duplicate_of_issue_in_active_review.replace(
+    return INELIGIBLE_REQUEST_ISSUES.duplicate_of_rating_issue_in_active_review.replace(
       '{review_title}', issue.titleOfActiveReview
     );
   } else if (issue.ineligibleReason) {
--- a/client/app/intake/components/IneligibleIssuesList.jsx
+++ b/client/app/intake/components/IneligibleIssuesList.jsx
@@ -1,0 +1,27 @@
+import React, { Fragment } from 'react';
+import INELIGIBLE_REQUEST_ISSUES from '../../../constants/INELIGIBLE_REQUEST_ISSUES.json';
+
+const ineligibilityCopy = (issue) => {
+  if (issue.titleOfActiveReview) {
+    return INELIGIBLE_REQUEST_ISSUES.duplicate_of_issue_in_active_review.replace(
+      '{review_title}', issue.titleOfActiveReview
+    );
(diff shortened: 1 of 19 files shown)


spec/feature/intake/intake_edit_confirmation_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+
+    before do
+      decision_review.create_issues!([request_issue])
+      decision_review.establish!
+    end
+
+    describe "given common behavior for claim reviews" do
+      let(:edit_path) { "#{claim_review_type.to_s.pluralize}/#{get_claim_id(decision_review)}/edit" }
+
+      [:higher_level_review, :supplemental_claim].each do |claim_review_type|
```
leikkisa: Cool!

spec/support/intake_helpers.rb
```diff
@@ -141,7 +163,22 @@ def add_intake_rating_issue(description, note = nil)
     safe_click ".add-issue"
   end
 
-  def add_intake_nonrating_issue(category:, description:, date:, legacy_issues: false)
+  def get_claim_id(claim_review)
```
leikkisa: I think you can look up the claim id from the claim review directly, instead of looking for an issue and it's EP code.
```EndProductEstablishment.find_by(source: claim_review).reference_id```

File: client/app/intake/components/IneligibleIssuesList.jsx [hunk 3]
```diff
@@ -0,0 +1,27 @@
+import React, { Fragment } from 'react';
```
leikkisa: Nice

spec/support/intake_helpers.rb
```diff
@@ -103,6 +103,24 @@ def start_appeal(
   # rubocop: enable Metrics/MethodLength
   # rubocop: enable Metrics/ParameterLists
 
+  def setup_intake_edit_flags
```
leikkisa: We can probably call these setup_intake_flags without the "edit" (same for teardown) since the same flags apply to regular intake.

Issue #8032, opened 2018-11-29
Allow attorneys to create multiple admin actions for a single case


lowellrex:
Currently, we only allow attorneys to create a single admin action for a given appeal. We should allow attorneys to create multiple admin actions for each appeal.

| Current | Future |
| --- | --- |
| ![image](https://user-images.githubusercontent.com/32683958/49228287-026d0a80-f3b9-11e8-9233-1b8f4072cf13.png) | ![attorneys_assign-admin-action-1](https://user-images.githubusercontent.com/31224257/49353914-4187ae00-f68e-11e8-882e-05fd4bdd3e7a.png) |

## Acceptance criteria
* Attorneys can create multiple admin actions for each appeal

### Non-goals
* Allow attorneys to create admin actions for appeals that are on hold. This is complicated because (on the back-end) we currently assign multiple admin actions that have been created to a single VLJ support staff member, so we would probably like to assign any new admin actions for that case to the same VLJ support staff member as well. This is totally doable, but may be a little bit of work that is less urgent than allowing the creation of multiple admin actions for a single appeal.

### Default view (user sees button to add second action)
![attorneys_assign-admin-action-1](https://user-images.githubusercontent.com/31224257/49353941-6845e480-f68e-11e8-83fc-0c21e5e3de41.png)

### Add second action fields open up + option to remove 2nd action, first action collapses.
![attorneys_assign-admin-action-2](https://user-images.githubusercontent.com/31224257/49353942-6845e480-f68e-11e8-8099-41ba9d7369e4.png)

### User attempts to edit first action after creating second action
![attorneys_assign-admin-action-2a](https://user-images.githubusercontent.com/31224257/49353943-6845e480-f68e-11e8-9b9e-d76172f88672.png)

### User adds third action
![attorneys_assign-admin-action-3](https://user-images.githubusercontent.com/31224257/49353945-6845e480-f68e-11e8-9e91-a5bf87bc2e10.png)

PR #8140, opened 2018-12-07
8032 create multiple admin actions

tomas-nava:
Connects #8032

### Description
We're looking for feedback on the approach to this ticket. See attached GIF for flow.

Current implementation doesn't match mocks precisely; will create follow-on PR to address those differences.

### Test plan
- login as an attorney
- go to a case
- add admin actions
- delete admin actions
- submit multiple admin actions in one click

## TODO:
- [x] refactor?
- [x] write tests

new-multiple-tasks-flow


client/app/queue/AddColocatedTaskView.jsx
```diff
@@ -47,32 +54,56 @@ type Props = Params & {|
...
-  validateForm = () => Object.values(this.state).every(Boolean);
+  getVisibleAdminActions = () => this.state.adminActions.filter((action) => !action.isHidden);
+
+  removeAdminActionFields = (index) => {
+    this.setState(update(this.state, { adminActions: { [index]: { isHidden: { $set: true } } } }));
+  }
+
+  addAdminActionFields = () => {
+    this.props.highlightInvalidFormItems(false);
+    this.setState(update(this.state, { adminActions: { [this.state.adminActions.length]: { $set: adminActionTemplate() } } }));
```
mdbenjam: Kind of a nitty thing, but I've started to feel `immutability-helper` is more trouble than it's worth. We've generally only used it in Redux and then done a hodgepodge of different solutions for local state. Totally up to you if you want to use here too, but I think for some situations like this one it's a little over verbose. I prefer just using splat. (This is untested so it might not actually work :D)

```
this.setState({ adminActions: [...this.state.adminActions, adminActionTemplate()] });
```

client/app/queue/AddColocatedTaskView.jsx
```diff
@@ -47,32 +54,56 @@ type Props = Params & {|
...
-      instructions: ''
+      adminActions: [adminActionTemplate()]
     };
   }
 
-  validateForm = () => Object.values(this.state).every(Boolean);
+  getVisibleAdminActions = () => this.state.adminActions.filter((action) => !action.isHidden);
+
+  removeAdminActionFields = (index) => {
+    this.setState(update(this.state, { adminActions: { [index]: { isHidden: { $set: true } } } }));
```
mdbenjam: Is there a reason we mark it as `isHidden` rather than just remove it? Feels like it makes some of our other code more complex.

lowellrex:
Merging this is blocked until we can display multiple actionable tasks in the same case snapshot #8214.

spec/support/feature_helper.rb
```diff
@@ -0,0 +1,13 @@
+module FeatureHelper
+  def click_dropdown(options = {}, container = page)
```
mdbenjam: Love this. It might be useful to add an option to select which dropdown in case there are multiple, so that you can avoid the `within` statements you used above. Though those are really fine too. So either way.

PR #8153, opened 2018-12-07
Fix judge decision issues

mdbenjam:
Judges currently cannot submit decision issues. Furthermore, we need a way of allowing judge to see decision issues & the old way of just adding dispositions to request issues.

Two solutions here:

1) Fix the problem of judges not submitting decision issues. This was happening because we had only changed the submission on the SubmitView page and not the EvaluateDecision page that judges use.
1) Never turn on the feature toggle for judges. Instead we'll act as if the feature toggle is on if the case has existing decision issues!

### Testing Plan
1. Update our existing attorney magic test to test the judge flow


app/controllers/case_reviews_controller.rb
```diff
@@ -79,9 +79,39 @@ def judge_case_review_params
...
+      [
+        :disposition,
+        :description,
+        :benefit_type,
+        request_issue_ids: [],
+        remand_reasons: [
+          :code,
+          :post_aoj,
+          :id,
+          :request_issue_id,
```
mdbenjam: We were missing a lot of fields in here.

app/controllers/case_reviews_controller.rb
```diff
@@ -79,9 +79,39 @@ def judge_case_review_params
...
+        ]
+      ]
+   else
+      [
+        :id,
+        :disposition,
+        :description,
+        :readjudication,
+        :benefit_type,
+        request_issue_ids: [],
```
mdbenjam: Added the decision issue fields here as well since judges might send decision issues without having the feature toggle.

app/models/concerns/case_review_concern.rb
```diff
@@ -23,7 +23,8 @@ def update_task_and_issue_dispositions
     end
 
     # Remove this check when feature flag 'ama_decision_issues' is enabled for all
-    if FeatureToggle.enabled?(:ama_decision_issues, user: RequestStore.store[:current_user])
+    if FeatureToggle.enabled?(:ama_decision_issues, user: RequestStore.store[:current_user]) ||
+       issues.first[:request_issue_ids]
```
mdbenjam: If the first sent issue has a request id, then this user was going through the decision issue flow even though they didn't have the feature toggle enabled.

client/app/components/RadioField.jsx
```diff
@@ -44,7 +44,7 @@ export default class RadioField extends React.Component {
     let labelClass = '';
 
     if (hideLabel) {
-      labelClass += ' hidden-field';
+      labelClass += ' usa-sr-only';
```
mdbenjam: This was wrong. Marking something as `display: hidden` makes it unavailable to screen readers. Sniffybara was complaining.

app/controllers/case_reviews_controller.rb
```diff
@@ -78,11 +78,18 @@ def judge_case_review_params
   end
 
   def issues_params
-    if ama? && feature_enabled?(:ama_decision_issues)
-      [:disposition, :description, request_issue_ids: [], remand_reasons: [:code, :post_aoj]]
-    else
-      [:id, :disposition, :readjudication, remand_reasons: [:code, :post_aoj]]
-    end
+    [
```
aroltsch: @mdbenjam would you please add a comment here explaining the logic

Issue #8155, opened 2018-12-07
When AMA tasks are completed, we still display them in the judge view of attorney tasks


amprokop:
Relates to https://github.com/department-of-veterans-affairs/appeals-support/issues/3605

![image](https://user-images.githubusercontent.com/3781835/49674443-08f62480-fa40-11e8-9959-7dfb0b01f30d.png)

PR #8156, opened 2018-12-07
Hide completed AMA attorney tasks when judge is looking at an attorney's tasks

amprokop:
Connects https://github.com/department-of-veterans-affairs/caseflow/issues/8155

### Test plan
- Completed task as attorney
- Saw that task disappeared from the judge's view of the attorney's tasks, and only appeared in the Review Cases list


amprokop:
closing, wrong branch

PR #8166, opened 2018-12-08
Fix only one remand reason gets sent

mdbenjam:
Currently remand reasons are broken :/. We only send the final remand's reason to the backend. This happens because our remand frontend components are built kind of "backwards" where a parent component reaches into the child components to call things outside our normal React flow. This combined with a change to use a new action the other day broke this flow.

Once this is merged: https://github.com/department-of-veterans-affairs/caseflow/pull/8153, I'll add to our test to have multiple decision issues with remand reasons.


client/app/queue/components/IssueRemandReasonsOptions.jsx
```diff
@@ -100,20 +96,10 @@ class IssueRemandReasonsOptions extends React.PureComponent<Params, State> {
...
-          remand_reasons: remandReasons
-        };
-      }
-
-      return issue;
-    });
-
-    const attributes = useDecisionIssues ? { decisionIssues: updatedIssues } : { issues: updatedIssues };
-
-    this.props.editStagedAppeal(appealId, attributes);
```
mdbenjam: So I changed this component to use `editStagedAppeal` instead of a more specific edit issue reducer. However, this broke remand reasons when they're are multiple remands with reasons. This component is built sort of backwards, where the parent is reaching in and calling lots of functions. So `editStagedAppeal` gets called for every issue in an each loop. But that doesn't give time for these components to update with the result of the last call to `editStagedAppeal`, so it overrides the data from the previous call to it and we're only left with the final remand reason.

PR #8167, opened 2018-12-09
Prevent SCT cases from being distributed

cmgiven:
@laurjpeterson Went ahead and made this change. I forget who else on Queue is working on SCT issues, but can you flag this PR for them, as the new join in VACOLS::Case may be useful to them.


laurjpeterson:
@joeyyang - adding you to this PR, in response to chris' comment above. The new join in VACOLS::Case may be useful to you.

PR #8168, opened 2018-12-10
Set timeout to 5 min (new_documents and document_count) and catch error properly

### LINKED ISSUE #7945: [FE] Response timeout of 60000ms exceeded
Sentry: https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2989/

aroltsch:
Resolves #7945

### Description
The call to new_documents and document_count can last up to 4 minutes. From New Relic

image

image

Nginx has default timeout of 60 seconds for its connection to the server side. If the back end server does not respond in 60 seconds, Nginx tears down its connection to the server, and further the connection from the client.
We can increase it by setting `proxy_read_timeout`. As an example from the devops repo:
https://github.com/department-of-veterans-affairs/devops/blob/04e08585ffea68e4477c12cb75d4a8b2756c78ee/ansible/roles/revproxy-configure/templates/nginx_revproxy.conf#L210

Increase timeout on the client for the above calls and catch error properly.

image

Debugging further shows that a lot of requests are in the stalled state

```
Stalled/Blocking
Time the request spent waiting before it could be sent. This time is inclusive of any time spent in proxy negotiation. Additionally, this time will include when the browser is waiting for an already established connection to become available for re-use, obeying Chrome's maximum six TCP connection per origin rule.
```

image


client/app/queue/reducers.js
```diff
@@ -148,6 +148,14 @@ export const workQueueReducer = (state: QueueState = initialState, action: Objec
         }
       }
     };
+  case ACTIONS.ERROR_ON_RECEIVE_DOCUMENT_COUNT:
+    return update(state, {
+      docCountForAppeal: {
+        [action.payload.appealId]: {
+          $set: 'Failed to load'
```
mdbenjam: This is definitely good to add!

Issue #8177, opened 2018-12-10
Legacy Opt-in | logic mismatch in eligibility check


pkarman:
The server does this:

https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/request_issue.rb#L302

The client does this:

https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/intake/components/AddedIssue.jsx#L64

which makes it appear to the client that the VACOLS issue will be closed, but the server will flag it as ineligible.

PR #8180, opened 2018-12-10
Fix case where server/client ineligible VACOLS logic mixed

pkarman:
connects #8177 #7337 

### Description
Fix bug where client UI shows eligible but server logic flags ineligible.


client/app/intake/components/LegacyOptInModal.jsx
```diff
@@ -30,7 +32,15 @@ class LegacyOptInModal extends React.Component {
     const legacyValues = value.split('-');
     const vacolsSequenceId = legacyValues.length > 1 ? legacyValues[1] : null;
     const legacyAppeal = this.props.intakeData.legacyAppeals.find((appeal) => appeal.vacols_id === legacyValues[0]);
-    const eligibleForSocOptIn = legacyAppeal && legacyAppeal.eligible_for_soc_opt_in;
+
+    let eligibleForSocOptIn, vacolsIssue;
+
+    if (vacolsSequenceId) {
```
leikkisa: The thing that bothers me about this is that there should never "not" be a vacolsSequenceId.  Because it's a required part of the identifier for identifying a vacols issue.

So if it's "null" they might have picked "none of these match", or something other weird.

It seems like this is here to be a safeguard.  You could also initialize the value to the legacy appeals value.  But in the case where this is null, I'm guessing there won't be a legacy appeal that has been found either (again, maybe because the option selected is none of these match).

client/app/intake/components/LegacyOptInModal.jsx
```diff
@@ -30,7 +32,15 @@ class LegacyOptInModal extends React.Component {
...
+      vacolsIssue = _.filter(legacyAppeal.issues, { vacols_sequence_id: parseInt(vacolsSequenceId, 10) })[0];
```
leikkisa: This works but I think using "find" is more representative of what's happening.

client/app/intake/components/LegacyOptInModal.jsx
pkarman: you are correct: the only time there is not a `vacolsSequenceId` is the case where the radio button is for "None of these match".

I am up to change that logic too, but I was trying to touch the Least Amount of Code here.

Issue #7659, opened 2018-11-01
Intake Refactor | Remove old AMA Intake finish page code


shanear:
Now that we've launched the AMA Add Issues flow, we can clear out the code for the old flow.

## Technical Notes
- There may be test coverage gaps between the old tests using the finish page and the feature tests exercising the add issues page
- There may be opportunity to refactor some of the delta above into lower level tests (controller, etc)

## Acceptance Criteria
- Remove the code for `:intake_enable_add_issues_page` feature toggle
- Remove the code for the old AMA intake finish page code

shanear:
the RAMP intake flows don't use the old Add issues flow either

PR #8203, opened 2018-12-11
Remove old AMA Intake finish page code

pkarman:
connects #7659 

### Description
Removes the `:intake_enable_add_issues_page` feature toggle and the code it was toggling.


client/app/intake/IntakeFrame.jsx
```diff
@@ -82,16 +82,16 @@ class IntakeFrame extends React.PureComponent {
...
-                  exact
-                  path={PAGE_PATHS.FINISH}
-                  title="Finish Processing | Caseflow Intake"
-                  component={FinishPage} />
                 <PageRoute
                   exact
                   path={PAGE_PATHS.ADD_ISSUES}
                   title="Add / Remove Issues | Caseflow Intake"
                   component={IntakeAddIssuesPage} />
+                <PageRoute
```
jiahuang: i would expect either the PAGE_PATHS.ADD_ISSUES or PAGE_PATHS.FINISH route to go away. Looks like ramp is still using the finish page though?

pkarman: yeah, RAMP still is. I had tried to yank `FINISH` completely but those features remain.

Issue #8183, opened 2018-12-10
Intake | Claimant bug on Add Issues flow when intake in progress


pkarman:
There's a bug preserving the state of a claimant when you reload the intake partway through.

To reproduce:

* Start an intake for HLR
* select a claimant other than the veteran
* successfully arrive at the `/intake/add_issues` step
* change the URL to `/intake`
* click the Continue button (like you did in step 2)
* a JS exception is thrown

![caseflow-intake-review-request](https://user-images.githubusercontent.com/1205061/49768316-8a002680-fca1-11e8-8c40-cdf48acb614e.gif)

pkarman:
fixed in 2e0335039cc5221f99d9f15aa5f5938437a13fdb part of #8203 -- big 💯 to @jiahuang for finding and squashing this bug!

PR #8206, opened 2018-12-11
fix claimants

jiahuang:
Connects #8183

The radio button sets a string "true"/"false" but our rails model has boolean true/false. This PR makes the redux store use boolean true/false instead of both string & boolean types.


client/app/intake/reducers/appeal.js
```diff
@@ -119,12 +120,13 @@ export const appealReducer = (state = mapDataToInitialAppeal(), action) => {
       }
     });
   case ACTIONS.SET_VETERAN_IS_NOT_CLAIMANT:
+    const veteranIsNotClaimant = convertStringToBoolean(action.payload.veteranIsNotClaimant);
```
pkarman: +1

jiahuang:
Closing in favor of 8203

PR #8212, opened 2018-12-11
Rkreyhsig/7943 hide duplicate tasks from case timeline

### LINKED ISSUE #7943: Hide duplicate tasks from case timeline
After we merge #7869 we will display all completed tasks in case timeline. This is a huge addition as it allows Board employees to get a more full picture of a case's progression through the process. However it creates a scenario where we can display duplicates in the case timeline when tasks are assigned to both organizations and members of those organizations (as in the example of the QR tasks below). 

image

A first thought about how we could handle this situation would be to add logic to the `Task. timeline_details` method to hide tasks assigned to organizations from the case timeline. There are certainly problems with this approach (for instance, organization-level tasks can be completed without ever having a child task assigned to a person created off of the organization task) so the solution here will probably be a little more involved.

## Acceptance criteria
* Do not display duplicate tasks in case timeline

rkreyhsig:
Resolves #7943  - Hide duplicate tasks from case timeline


app/models/appeal.rb
```diff
@@ -222,13 +222,18 @@ def create_tasks_on_intake_success!
     RootTask.create_root_and_sub_tasks!(self)
   end
 
+  def tasks_for_timeline
+    tasks.where(status: Constants.TASK_STATUSES.completed).order("completed_at DESC")
+      .reject { |t| t.assigned_to_type == "Organization" && t.children.pluck(:assigned_to_type).include?("User") }
```
lowellrex: Let's use `t.assigned_to.is_?(Organization)` and `User.name` here to get some of the benefits of type errors here.

lowellrex: Also let's add a code comment here about why we are only showing completed tasks and why we are hiding these organizational tasks.

spec/models/appeal_spec.rb
```diff
@@ -287,4 +287,22 @@
...
+        create(:qr_task, appeal: appeal, status: Constants.TASK_STATUSES.completed, assigned_to_type: "Organization")
+      end
+      let!(:task3) do
+        create(:qr_task, assigned_to: judge, appeal: appeal, status: Constants.TASK_STATUSES.completed,
+                         parent_id: task2.id)
+      end
+
+      subject { appeal.tasks_for_timeline.first }
+      it { is_expected.to eq task3 }
+    end
```
lowellrex: Let's add a test that confirms that an completed organization task without child tasks is included in the list of returned tasks.

spec/models/appeal_spec.rb
```diff
@@ -287,4 +287,22 @@
       it { is_expected.to eq judge }
     end
   end
+
+  context "case timeline returns unique tasks containing only child user task" do
```
lowellrex: We usually name the top level context in a model test with the name of the method we are testing and then the subsequent contexts with the specific scenarios we are testing. So this would become something like:
```ruby
context ".tasks_for_timeline" do
  context "when there are completed organization tasks with completed child tasks assigned to people" do
```

spec/models/appeal_spec.rb
rkreyhsig: Done!

app/models/appeal.rb
```diff
@@ -222,13 +222,19 @@ def create_tasks_on_intake_success!
     RootTask.create_root_and_sub_tasks!(self)
   end
 
+  # Hides duplicate cases from timeline that have assigned_to_type both "User" and "Organization"
```
lowellrex: Let's craft this comment together!

PR #8222, opened 2018-12-11
Create mail task modal routes to team based on task type

### LINKED ISSUE #7906: Create mail task modal routes to team based on task type
As we [roll out access to queue to the mail team](https://github.com/department-of-veterans-affairs/caseflow/issues/7404), we need to add actions that support their workflow. Currently, clicking on the "Create mail task" task action brings up the generic "Assign to team" modal. Mail is used to creating a mail task that is routed automatically to a Board team based on the type of mail task that is being created (mail types listed in #7404), so we need to support that way of working by automatically routing the mail task to the correct team based on the type of task that is being created.

## Acceptance criteria:
* Determine the routing rules for each mail task
* Automatically route new mail tasks to the correct team based on the mail task type

## Reference - rollout progression
1. Allow mail team to assign tasks to `VLJ Support`, `AOD`, `Privacy`
2. Probably, next allow mail team to assign tasks to `Hearings`
3. Allow mail team to assign tasks to `Lit Support`

lowellrex:
Resolves #7906.

## How to create a mail task
* Log in as user `JOLLY_POSTMAN`.
* Search for a Veteran ID associated with any AMA appeal (`500000001` is always a good choice).
* Click on the Veteran's name to access the case details page.
* Select the "Create mail task" option in the task actions dropdown.

mail_task_creation


app/models/tasks/mail_task.rb
```diff
@@ -1,28 +1,146 @@
...
+  end
+end
+
+class AddressChangeMailTask < ColocatedMailTask
+  class << self
+    def label
+      COPY::ADDRESS_CHANGE_MAIL_TASK_LABEL
+    end
+
+    def get_child_task_assignee(parent, params)
```
lowellrex: `MailTask.get_child_task_assignee` is where we will put all of the routing logic.

app/models/tasks/mail_task.rb
```diff
@@ -1,28 +1,146 @@
...
+    def get_child_task_assignee(parent, _params)
+      parent_colocated_task_already_created?(parent) ? next_assignee : Colocated.singleton
+    end
+
+    def parent_colocated_task_already_created?(parent)
+      parent.is_a?(name.constantize) && parent.assigned_to == Colocated.singleton
+    end
+  end
+
+  def create_grandchild_task
```
amprokop: hmm naming nit — we're inside the `ColocatedMailTask` class here so we're just creating a child task from this class, right? Even though I can see in the larger context we'd be creating a grandchild task of the whole set of mail tasks.

app/models/tasks/mail_task.rb
```diff
@@ -1,28 +1,297 @@
...
+        PowerOfAttorneyRelatedMailTask,
+        ReturnedUndeliverableCorrespondenceMailTask,
+        StatusInquiryMailTask,
+        DeathCertificateMailTask,
+        ExtensionRequestMailTask,
+        PrivacyActRequestMailTask,
+        OtherMotionMailTask,
+        PrivacyComplaintMailTask,
+        VacateMotionMailTask,
+        AppealWithdrawalMailTask
```
lowellrex: Preserving ordering of task types according to [their previous number](https://github.com/department-of-veterans-affairs/caseflow/issues/7906#issuecomment-440442213).

app/models/tasks/generic_task.rb
```diff
@@ -98,22 +98,21 @@ def create_from_params(params, user)
...
         assigned_by_id: child_assigned_by_id(parent, current_user),
         parent_id: parent.id,
-        assigned_to: assignee,
+        assigned_to: get_child_task_assignee(parent, params),
         instructions: params[:instructions]
       )
     end
 
-    private
+    def get_child_task_assignee(_parent, params)
```
tomas-nava: This could be just `child_task_assignee`

app/models/tasks/mail_task.rb
tomas-nava: Are you using this instead of [the built-in subclasses method](https://api.rubyonrails.org/classes/Class.html#method-i-subclasses) just to preserve the order? How dynamic will this list be? Would it be better to add a `select_order` column to the database or something like that?

tomas-nava: Yep, agreed, alphabetical is best.

Issue #8223, opened 2018-12-11
DecisionReview establish! and legacy opt in processing


leikkisa:
## Story
Set up a DecisionReview-level asynchronous processing job for appeals and claim reviews so that it can be used to process legacy issue opt ins.

## Acceptance criteria
- Legacy opt ins happen as part of a decision review establishment and RequestIssuesUpdate
- Legacy opt ins are rolled back as part of RequestIssuesUpdate
- Multiple issues on a legacy appeal are processed in a batch, to prevent race conditions affecting whether the legacy appeal is closed or reopened

## Technical notes
This can be done in five parts:

* [x] Refactor to prepare
 - change process_end_product_establishments! to establish!
 - change ClaimReviewProcessJob to DecisionReviewProcessJob

* [x] Add async establishment task to Appeal
 - add establish! on Appeal
 - add asyncable columns to appeal
 - add code to kick off and retry DecisionReviewProcessJob for establishing appeals

* [x] Move from individual processing of legacy opt ins to batch processing
 - make a class for CloseLegacyOptIns that batches the closing of issues/appeals
 - run CloseLegacyOptIns on establish! on all decision reviews
 - remove code that closes vacols issues in LegacyOptIn
 - get rid of LegacyOptIn job

* [x] Reopen VACOLS issues as part of RequestIssuesUpdate
 - make a class for ReopenLegacyOptIns that batches the reopening of issues/appeals
 - run CloseLegacyOptIns on establish! when processing RequestIssuesUpdate

* [ ] Schema cleanup
 - remove VACOLS issue data from request_issues
 - get rid of asyncable data on legacy_opt_ins, also appeal json
 - create virtual setter for vacols_issue_opt_ins (allow data to be passed fluidly from front-end)

steve-gov:
Technical Validation: Will be validated as part of other completed tickets.

PR #8232, opened 2018-12-12
Asyncable appeals

pkarman:
connects #8223 

### Description
Step 2 of refactoring of async Appeals and Legacy Opt Ins.


app/models/appeal.rb
```diff
@@ -237,7 +236,9 @@ def timeline
   end
 
   def establish!
-    # currently a no-op
+    attempted!
+    # TODO: actual stuff
```
shanear: 😄

PR #8248, opened 2018-12-12
Tomas/8015 Simplify task auth logic

### LINKED ISSUE #8015: Simplify Task.update_from_params authorization logic

tomas-nava:
Resolves #8015

### Description
Renames and simplifies the methods that control access to tasks.


app/models/task.rb
```diff
@@ -33,18 +33,18 @@ def label
...
+    actions_available?(user) ? available_actions(user).map { |action| build_action_hash(action) } : []
   end
 
   def build_action_hash(action)
     { label: action[:label], value: action[:value], data: action[:func] ? send(action[:func]) : nil }
   end
 
-  def no_actions_available?(user)
-    return true if [Constants.TASK_STATUSES.on_hold, Constants.TASK_STATUSES.completed].include?(status)
+  def actions_available?(user)
```
lowellrex: YESSS!

app/models/task.rb
```diff
@@ -33,18 +33,18 @@ def label
...
+    return false if [Constants.TASK_STATUSES.on_hold, Constants.TASK_STATUSES.completed].include?(status)
 
     # Users who are assigned a subtask of an organization don't have actions on the organizational task.
-    assigned_to.is_a?(Organization) && children.any? { |child| child.assigned_to == user }
+    !assigned_to.is_a?(Organization) || children.all? { |child| child.assigned_to != user }
```
lowellrex: This might read easier if we just return false if the previous condition is true and then fall through and return true. Something like:
```ruby
  ...
  return false if assigned_to.is_a?(Organization) && children.any? { |child| child.assigned_to == user }
  true
end
```

I think that's mostly stylistic, so it's totally up to you whether we make that change or not.

app/models/task.rb
```diff
@@ -141,23 +141,17 @@ def when_child_task_completed
...
-  def verify_user_access!(user)
-    unless can_be_accessed_by_user?(user)
+  def verify_user_can_update!(user)
+    unless can_be_updated_by_user?(user)
       fail Caseflow::Error::ActionForbiddenError, message: "Current user cannot access this task"
     end
   end
 
-  def self.verify_user_can_assign!(user)
+  def self.verify_user_can_create!(user)
```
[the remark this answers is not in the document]
lowellrex: Moved into a separate ticket here: https://github.com/department-of-veterans-affairs/caseflow/issues/8275

spec/spec_helper.rb
```diff
@@ -54,12 +54,12 @@
...
-  #   # Allows RSpec to persist some state between runs in order to support
-  #   # the `--only-failures` and `--next-failure` CLI options. We recommend
-  #   # you configure your source control system to ignore this file.
-  #   config.example_status_persistence_file_path = "spec/examples.txt"
-  #
+
+  # Allows RSpec to persist some state between runs in order to support
+  # the `--only-failures` and `--next-failure` CLI options. We recommend
+  # you configure your source control system to ignore this file.
+  config.example_status_persistence_file_path = "tmp/examples.txt"
```
monfresh: 💯

PR #8256, opened 2018-12-12
Refactor way we are uploading hearing coordinators

### LINKED ISSUE #8255: Duplicate User Error When Querying for Coordinators
Duplicate user error when getting hearing coordinators from staff table and trying to create them in Caseflow. We need to remove the name from the method call as we may already have the user in vacols with a slightly different name.

oscarAtNava:
to Caseflow to avoid duplicate entry errors.

Resolves #8255 

### Description
Find or create only by css_id and station, then update the names in a subsequent loop if the names are null or blank.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Verify hearings home page loads without errors.


app/repositories/user_repository.rb
```diff
@@ -70,11 +70,19 @@ def css_id_by_full_name(full_name)
...
+      user_records = coordinator_records.select(&:sdomainid).map do |record|
         User.find_or_create_by(css_id: record.sdomainid,
-                               full_name: "#{record.snamef} #{record.snamemi} #{record.snamel}",
                                station_id: User::BOARD_STATION_ID)
       end
+
+      user_records.map do |user|
+        if user.full_name.nil? || user.full_name == ""
+          coordinator = coordinator_records.find_by(sdomainid: user.css_id)
+          user.update(full_name: `#{coordinator.snamef} #{coordinator.snamemi} #{coordinator.snamel}`)
```
sharonwarner: Let's update this to use the vacols_full_name method. That's what is used in full_name in user.rb.

sharonwarner: Actually, I just realized, we don't need this logic. The full_name method overwrites the active record method to pull from VACOLS if the field is blank.

PR #8259, opened 2018-12-12
LegacyOptinManager

pkarman:
connects #8223 

### Description
Batch processing of `LegacyIssueOptin` to avoid race conditions.

Step 3 of #8223


app/services/legacy_optin_manager.rb
```diff
@@ -0,0 +1,84 @@
...
+    @decision_review = decision_review
+  end
+
+  # we operate on LegacyIssueOptin rows, within a single VACOLS transaction
+  # for now, just close VACOLS records
+  def process!
+    VACOLS::Case.transaction do
+      affected_legacy_appeals.each do |legacy_appeal|
+        # track which issues we close during this transaction
+        open_legacy_issues = legacy_appeal.issues.reject(&:closed?)
```
pkarman: should probably add a `next unless open_legacy_issues.any?` check here so we are idempotent

leikkisa: 👍

pkarman:
@leikkisa thanks for fixing the tests here -- I think it just needs your review and we're good to go.

Issue #7984, opened 2018-11-26
AMA Intake | Use different EP labels for Pension HLRs and SCs


shanear:
Pension HLRs and SCs will need to use different labels than Compensation so they can be routed properly.

## Acceptance Criteria
- If "Pension" was selected as the benefit type for a HLR or SC intake, when the intake is completed, the EPs created should have the appropriate PMC EP codes (see the table below)
- When a request issue on an HLR or SC with the benefit type of "Pension" has a duty to assist error, the resulting EP should use the appropriate PMC EP code below.

## Notes

If it's easy, update the EP labels displayed on the confirmation pages with the labels below

| Pension EP Code | Definition  |
| ------------- | ------------- | 
 030HLRNRPMC | PMC Higher-Level Review Non-Rating |
| 030HLRRPMC | PMC Higher-Level Review Rating |
| 040BDEIMOPMC | PMC Board DTA Error - w/IMO |
| 040HDENRPMC | PMC HLR DTA Error - Non-Rating |
| 040HDERPMC | PMC HLR DTA Error - Rating |
| 040SCNRPMC | PMC Supplemental Claim Non-Rating |
| 040SCRPMC | PMC Supplemental Claim Rating |

## Background
https://github.com/department-of-veterans-affairs/appeals-design-research/issues/1032

steve-gov:
First two codes validated (created a NR-HLR and a R-HLR for Veteran 123458314): 

<img width="434" alt="screen shot 2019-01-03 at 11 13 50 am" src="https://user-images.githubusercontent.com/25468689/50656807-3e136b00-0f49-11e9-860a-d99ddf83fa28.png">
<img width="478" alt="screen shot 2019-01-03 at 11 15 57 am" src="https://user-images.githubusercontent.com/25468689/50656808-3e136b00-0f49-11e9-9f8d-eaaf00cbf1d1.png">

steve-gov:
Contacted Nate for help with validating `040HDERPMC` and `040HDENRPMC`. Will leave it to @shanear to validate `040BDEIMOPMC`.

leikkisa:
040BDEIMOPMC actually has it's own ticket, so it's not part of this one.
https://github.com/department-of-veterans-affairs/caseflow/issues/6167

steve-gov:
Otherwise validated; will close and validate 6167 when it comes up.

PR #8260, opened 2018-12-12
Add Pension EP codes

leikkisa:
connects #7984 

Adds new EP codes if a Benefit Type is "Pension"


spec/models/claim_review_spec.rb
```diff
@@ -208,6 +210,34 @@ def random_ref_id
...
+        let(:benefit_type) { "pension" }
+
+        it "creates issues and assigns pension end product codes to them" do
+          subject
+
+          expect(rating_request_issue.reload.end_product_establishment).to have_attributes(code: "030HLRRPMC")
+          expect(non_rating_request_issue.reload.end_product_establishment).to have_attributes(code: "030HLRNRPMC")
+        end
+
+        context "for a supplemental claim" do
```
kencheeto: this is a little surprising in the claim_review_spec - i would expect tests here to be covering only generic claim_review functionality. Is there a better place to put this test?

app/models/supplemental_claim.rb
```diff
@@ -15,27 +19,10 @@ def ui_hash
...
-  end
-
-  def end_product_base_modifier
-    valid_modifiers.first
-  end
-
-  def valid_modifiers
-    END_PRODUCT_MODIFIERS
-  end
-
```
kencheeto: 😍 love this cleanup

spec/models/higher_level_review_spec.rb
```diff
@@ -269,6 +273,21 @@
...
+        let(:benefit_type) { "pension" }
+
+        it "creates end product establishment with pension ep code" do
+          subject
+
+          first_dta_request_issue = RequestIssue.find_by(rating_issue_reference_id: "rating1")
+          second_dta_request_issue = RequestIssue.find_by(rating_issue_reference_id: "rating2")
+
+          expect(first_dta_request_issue.end_product_establishment.code).to eq("040HDERPMC")
+          expect(second_dta_request_issue.end_product_establishment.code).to eq("040HDERPMC")
```
kencheeto: is it important to test two dta request issues?

leikkisa: Nope, I was just following the test above, but will simplify this.

spec/models/claim_review_spec.rb
leikkisa: Good call, I can add specific tests to HLR and SC.

PR #8276, opened 2018-12-13
[WIP] Alert for VVA and VBMS document list retrievals

### LINKED ISSUE #6049: Add alert when VBMS or VVA documents don't download
Sometimes we fail to pull documents from one of our dependencies (VVA or VBMS). We still load Reader with the documents that succeeded loading. However, this can cause confusion. Users see that documents loaded, and don't realize that all of the documents are not there.

We currently present [two timestamps at the bottom of the application](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/reader/LastRetrievalInfo.jsx#L15) that show when the last successful read from each service occurred. However, this is subtle and often missed by users. We should have an [alert](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/components/Alert.jsx) at the [top of the document list view](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/reader/PdfListView.jsx#L68-L69), which clearly says when we couldn't fetch documents from a service.

hschallhorn:
Resolves #6049

### Description
Added error in the document list for when the was never a response from VBMS or VVA. Added a similar warning for when the last response was older than the eFolder cache limit. Reimplemented retrieval timestamps.


client/app/reader/LastRetrievalAlert.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+    let staleCacheTimestamp = staleCacheTime.getTime() / 1000,
+      vbmsManifestTimeString = this.props.manifestVbmsFetchedAt,
+      vvaManifestTimeString = this.props.manifestVvaFetchedAt;
+
+    let parsableVbmsManifestTimeString = vbmsManifestTimeString.slice(0, -4) +
+      TIMEZONES[vbmsManifestTimeString.slice(-4)],
+      parsableVvaManifestTimeString = vvaManifestTimeString.slice(0, -4) +
+        TIMEZONES[vvaManifestTimeString.slice(-4)],
+      vbmsManifestTimestamp = moment(parsableVbmsManifestTimeString, 'MM/DD/YY HH:mma Z').unix(),
```
mdbenjam: Are you sure moment can't do the parsing of the original timestamp? I'm pretty hesitant to add our own timezone parsing code.

client/app/reader/LastRetrievalAlert.jsx
```diff
@@ -0,0 +1,65 @@
...
+        Please refresh your browser at a later point to view a complete list of documents in the claims
+        folder.
+      </Alert>;
+    }
+
+    let staleCacheTime = new Date();
+
+    staleCacheTime.setHours(staleCacheTime.getHours() - CACHE_TIMEOUT_HOURS);
+
+    let staleCacheTimestamp = staleCacheTime.getTime() / 1000,
```
mdbenjam: nit: We like to use `const` wherever possible. Looks like `staleCacheTimestamp` isn't mutated change.

mdbenjam: This also applies to some of the other variables in this method.

client/app/reader/LastRetrievalInfo.jsx
```diff
@@ -4,12 +4,10 @@ import { connect } from 'react-redux';
 
 class UnconnectedLastRetrievalInfo extends React.PureComponent {
   render() {
-    if (!this.props.manifestVbmsFetchedAt) {
-      return null;
-    }
-
     return [
-      <div id="vbms-manifest-retrieved-at" key="vbms">Last VBMS retrieval: {this.props.manifestVbmsFetchedAt}</div>,
+      this.props.manifestVbmsFetchedAt ?
```
mdbenjam: 👍

client/app/styles/reader/_document_list.scss
```diff
@@ -242,6 +242,10 @@
       background: inherit;
     }
   }
+
+  .usa-alert {
```
mdbenjam: So I want to get away from adding css properties for specific cases. Instead, I'd wrap the alert in a div with the required spacing. Also we can use `glamor` to put our CSS directly into our React component!

client/app/reader/LastRetrievalAlert.jsx
hschallhorn: 🎯 I was too. Unfortunately, moment doesn't handle timezone abbreviations. If you have a better suggestion on this, I'm all ears.

client/app/reader/LastRetrievalAlert.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+    let parsableVbmsManifestTimeString = vbmsManifestTimeString.slice(0, -4) +
+      TIMEZONES[vbmsManifestTimeString.slice(-4)],
+      parsableVvaManifestTimeString = vvaManifestTimeString.slice(0, -4) +
+        TIMEZONES[vvaManifestTimeString.slice(-4)],
+      vbmsManifestTimestamp = moment(parsableVbmsManifestTimeString, 'MM/DD/YY HH:mma Z').unix(),
+      vvaManifestTimestamp = moment(parsableVvaManifestTimeString, 'MM/DD/YY HH:mma Z').unix();
+
+    // Check that manifest results are fresh
+    if (vbmsManifestTimestamp < staleCacheTimestamp || vvaManifestTimestamp < staleCacheTimestamp) {
```
mdbenjam: http://momentjs.com/docs/#/query/is-before/

client/app/reader/LastRetrievalAlert.jsx
```diff
@@ -0,0 +1,65 @@
...
+        <br />
+        <br />
+        Please refresh your browser at a later point to view a complete list of documents in the claims
+        folder.
+      </Alert>;
+    }
+
+    let staleCacheTime = new Date();
+
+    staleCacheTime.setHours(staleCacheTime.getHours() - CACHE_TIMEOUT_HOURS);
```
mdbenjam: http://momentjs.com/docs/#/manipulating/subtract/

client/app/reader/LastRetrievalAlert.jsx
```diff
@@ -0,0 +1,70 @@
...
+      parsableVvaManifestTimeString = vvaManifestTimeString.slice(0, -4) +
+        TIMEZONES[vvaManifestTimeString.slice(-4)],
+      vbmsManifestTimestamp = moment(parsableVbmsManifestTimeString, 'MM/DD/YY HH:mma Z'),
+      vvaManifestTimestamp = moment(parsableVvaManifestTimeString, 'MM/DD/YY HH:mma Z');
+
+    // Check that manifest results are fresh
+    if (vbmsManifestTimestamp.isBefore(staleCacheTime) || vvaManifestTimestamp.isBefore(staleCacheTime)) {
+      return <div {...alertStyling}>
+        <Alert title="Warning" type="warning">
+          You may be viewing an outdated list of claims folder documents. Please refresh the page to load
```
mdbenjam: @sneha-pai I'm thinking we should alter this text a little bit. I don't want people to think they need to constantly refresh. I'm wondering if we can say something like "We last synced with VBMS and VVA X hours ago. If you'd like to check for new documents, refresh the page."

client/app/reader/LastRetrievalAlert.jsx
```diff
@@ -0,0 +1,70 @@
...
+
+    const staleCacheTime = moment().subtract(CACHE_TIMEOUT_HOURS, 'h'),
+      vbmsManifestTimeString = this.props.manifestVbmsFetchedAt,
+      vvaManifestTimeString = this.props.manifestVvaFetchedAt;
+
+    const parsableVbmsManifestTimeString = vbmsManifestTimeString.slice(0, -4) +
+      TIMEZONES[vbmsManifestTimeString.slice(-4)],
+      parsableVvaManifestTimeString = vvaManifestTimeString.slice(0, -4) +
+        TIMEZONES[vvaManifestTimeString.slice(-4)],
+      vbmsManifestTimestamp = moment(parsableVbmsManifestTimeString, 'MM/DD/YY HH:mma Z'),
```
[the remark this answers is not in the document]
hschallhorn: `tz()` unfortunately doesn't take the abbreviated time zone names. Here's moment's thoughts on why not: https://github.com/moment/moment-timezone/issues/289
We could change the format of manifestVbmsFetchedAt/manifestVvaFetchedAt to be less ambiguous. Timestamps from Time.zone.now are already parseable by moment (eg: `2018-12-18 12:52:19 -0500`) but we're reformatting the string to be readable but not parseable before it gets to the front end.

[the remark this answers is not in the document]
mdbenjam: HA well that makes a lot of sense. I think that solution is better. Let's send the timestamp normally without this extra formatting then so that moment can parse it.

hschallhorn:
@mdbenjam Timezoning issues fixed! Timestamps still show the abbreviated timezone, but the timezone offset is also passed up to be parsed by moment for when time calculations are needed.

client/app/reader/LastRetrievalAlert.jsx
sneha-pai: aha! that sounds like a better alternative. It's less ominous-sounding.

Let's go with that!

Issue #8216, opened 2018-12-11
Show "+ Add a new task" button in active tasks title bar


lowellrex:
![image](https://user-images.githubusercontent.com/32683958/49831646-66e97b80-fd62-11e8-969b-0dd5f09c28cb.png)

## Acceptance criteria
* [ ] Show "+ Add a new task" button in active tasks title bar if the current user can take action on the root task.
* [ ] Hide the root task from the active tasks component.

## Non-goal
* Showing/hiding the task instructions when clicking the "View task instructions" link will not be implemented as part of this ticket.

Issue #5151, opened 2018-04-12
Judge check out | Evaluate decision


laurjpeterson:
After a judge reviews a decision and confirms issue dispositions, judges are required to evaluate the quality of the attorney's decision for their performance evaluation. 

Judges will get to this screen in a few ways:
- If there are no remanded issues, directly after the 'Review dispositions' screen
- If there are remanded issues, directly after the 'Select remand reason(s)' for the last remanded issue.

### Header area 
- Verify that the title of the Page follows the following convention: Evaluate Decision - Veteran Name
 - Underneath the title in body text verify the following: 
    -  Counsel: Counsel Name
    - Document ID: Document ID associated with the decision 
- Verify that they are three horizontal sections (see mock)
  - Case details:
    - Docket number: 
    - Number of issues: 
    - Number of documents
   - Timeliness: 
     - Date assigned: 
     - Decision submitted: 
     - Calendar days worked: 
   - Attorney Notes: This is pulled from the notes field in the attorney check out flow submit page 
- Verify that there is a horizontal line separating the header and the body area 
### Body area
- Verify that the following question is displayed: "How would you rate the overall complexity of the case?" Underneath this question should be the following three circles (these options are blank by default): 
  - Easy
  - Medium
  - Hard
- When the judge makes a selection, the final complexity rating in VACOLS should be updated. The initial complexity rating can remain untouched.
- Verify that the next question is How would you rate the overall quality of this case? With the following options (these options are blank by default): 
  - 5 - Outstanding
  - 4 - Exceeds Expectations
  - 3 - Meets expectations
  - 2 - Needs improvement 
  - 1 - Does not meet expectations 
  - If a user chooses quality rating 1 or 2, the following note displays: 
     - Please provide more details about the quality of work below. If none of the below apply, please add your thoughts in the comments box below. 
     - There is no error state
- Verify that they are two horizontal sections after the quality question with the following titles: 
   - Where there any factors not considered?
   - Where there any areas of improvement? 
   - Please see mock for the options
- Verify that there is a comments section titled Additional factors affecting complexity and quality

### General notes
- Calendar days worked = Decision submitted - Date assigned
- The default state of the complexity and quality states is that no buttons are selected
- There is no error state if the user does not select a box after a 1 or 2 is selected
- Please follow the navigation patterns in the attorney check out flow

## Mocks
_See most recent comment_

Chingujo:
@laurjpeterson 

[...]

PR #8281, opened 2018-12-13
7682/task snapshot component mvp

### LINKED ISSUE #7682: Move active tasks from case snapshot into a lockup below universal case title
_Lowell: The final version of case timeline (mocks in [this mural](https://app.mural.co/t/workqueue2001/m/workqueue2001/1531849576358/38b753bff1e25963a76d8ebecf9679d6445c39b3)), we will include the incomplete tasks above the completed tasks in a continuous linear view. In order to get there we need to be able to view multiple tasks at once. I think the mocks provide a good example of how this could look. I propose we move the task information and actions dropdown out of the case snapshot area with this ticket and move it into this lockup._

_An open question is what case snapshot looks like without the task information. Perhaps docket type and AOD/CAVC (and VLJ and attorney information) get moved into [the proposed universal case title](https://github.com/department-of-veterans-affairs/caseflow/issues/6963)? @sneha-pai what do you think of this approach?_ [Resolved by Universal Case title]

## Acceptance criteria
* [ ] Primary task is displayed in new lockup on case details page.

image

## Non-goal
* Showing/hiding the task instructions when clicking the "View task instructions" link. This version will always display task instructions.
* "+ Add a new task" button will not be implemented as part of this ticket.

image

Please see https://app.mural.co/t/workqueue2001/m/workqueue2001/1543851111020/bbd6df5b95ca9a9ad179cd991d40c9314f35c929 for more details about the active-task logic.

rkreyhsig:
Resolves #7682 - Move active tasks from case snapshot into a lockup below universal case title

| Before | After |
| --- | --- |
|screen shot 2018-12-17 at 10 56 16 am|screen shot 2018-12-17 at 10 55 40 am|


lowellrex:
Let's add a before/after screenshot to this PR

rkreyhsig:
Before/after screenshot added!

client/app/queue/EvaluateDecisionView.jsx
```diff
@@ -211,7 +210,6 @@ class EvaluateDecisionView extends React.PureComponent {
       {error && <Alert title={error.title} type="error" styling={css(marginTop(0), marginBottom(1))}>
         {error.detail}
       </Alert>}
-      <CaseSnapshot appealId={appealId} hideDropdown />
```
lowellrex: Why did we decide to remove the case snapshot from this page instead of replacing it with the task snapshot?

| Before | After |
| --- | --- |
| image | image |

rkreyhsig: I was just removing the Case Snapshot completely like we had talked about.  Didn't think to add the TaskSnapshot to replace it here.  Should I add it?

lowellrex:
How do we want to handle the case where there are no currently active tasks? 

I'd be in favor of either 1) hiding the entire task snapshot box or 2) hiding only the contents of the box and changing the text of the title to reflect that the current user does not have any tasks assigned to them for the current case.

| Current | Option 1 | Option 2 |
| --- | --- | --- |
| image | image | image |

Curious to get thoughts from @eric-nava @allyceh and @sneha-pai here as well.

eric-nava:
Option 2 seems a bit more explicit, so I am leaning towards that. Do we have a component that is more visually like an "FYI" or do we have to use a header here?

lowellrex:
@eric-nava Maybe one of the alerts makes more sense than just the header here? http://www.caseflowdemo.com/styleguide#alerts

One reason to prefer using the header here instead of an alert is because [we intend to put the "+ Add a new task" button](https://github.com/department-of-veterans-affairs/caseflow/issues/8216) in that header for members of the mail team to create mail tasks for any appeal they search for, so I thought it might provide a little visual continuity so that we don't have to change the appearance of that section if the user is a member of the mail team. However, since folks will be using Caseflow logged into their own accounts most of the time and not seeing how other folks view the same pages, maybe having different UIs to convey the same piece of information is fine.

Mocks of how a case details page would look when the current user has no tasks assigned to them but is a member of the mail team:

| Not on the mail team | On the mail team |
| --- | --- |
| image | image |

client/app/queue/TaskSnapshot.jsx
```diff
@@ -23,35 +22,43 @@ import { DateString } from '../util/DateUtil';
...
+import { GrayDot } from '../components/RenderFunctions';
+import {
+  sectionSegmentStyling,
+  sectionHeadingStyling,
+  anchorJumpLinkStyling
+} from './StickyNavContentArea';
+
+export const grayLine = css({
+  width: '5px',
+  minHeight: '50px',
```
lowellrex: I think we can get rid of the `minHeight` property since we are positioning this absolutely now. This will improve how this line appears at smaller sizes as well:

| With `min-height: 50px` | Without any `min-height` |
| --- | --- |
| image | image |

lowellrex:
Because I was curious, this is how the task snapshot will appear for members of the mail team until [we add the "+ Add a new task" button](https://github.com/department-of-veterans-affairs/caseflow/issues/8216). It's not gorgeous but it works and we'll replace this with something prettier in the near future.

image

client/app/queue/TaskSnapshot.jsx
rkreyhsig: Removed!

allyceh:
@lowellrex @eric-nava How about having the header "Currently active tasks" stay the same all the time, keep the snapshot box exposed, and say "None" or "No active tasks" in the snapshot box?

The reason I am proposing this is because it is consistent with how we'll show that a veteran doesn't have a POA (Header: Power of Attorney, Body copy: None). Since these two elements are on the same Case Details page, this will ensure were consistent with how we're presenting these.

client/app/queue/EvaluateDecisionView.jsx
rkreyhsig: @laurjpeterson @sneha-pai @eric-nava @allyceh. Do we keep this here or remove it?  https://github.com/department-of-veterans-affairs/caseflow/issues/5151#issuecomment-395145113 for context around why we originally added it.

lowellrex:
@allyceh works for me!

client/app/queue/EvaluateDecisionView.jsx
allyceh: @laurjpeterson @sneha-pai , trying to get more context here.  

With the context you have from working on this page in June, why do we include case/task information on the Evaluation page at this point in the flow? 

For legacy and AMA cases, prior to seeing the Evaluation page, the judge will also see the Special Issues and Decisions pages, both of which don't contain the Case Snapshot or Task Snapshot yet. Is our new case/task snapshot something that judges would expect to persist throughout the checkout flow?

Just chatted with @lowellrex and @rkreyhsig . We have a couple of questions for you!

If we displayed Task Snapshot on the Evaluation page:
1. Would "assigned on" and "days waiting" be the day it was assigned to the judge and how long it's been in their queue? Lowell and I would be in favor of dropping this section if this information isn't needed or useful at this point. 
2. "Assigned to" and "task instructions" would probably be the most useful. Want to confirm that attorney notes would appear here and would be useful at this point in time. 
3. We would remove the "actions" drop down because the judge can't take any actions at this point.

lowellrex: We're going to add the `TaskSnapshot` back into this view (replacing `CaseSnapshot`) for the time being so we can get this code live without any regressions while we wait for the decision on whether the snapshot stays or goes in the long term.

laurjpeterson: I think we can nix showing currently active tasks on this page. I think the reason it was shown was to show the notes that the attorney wrote the judge, which was in the "Task Information" section of the previous Case Snapshot. 

I think we still need to show the notes the attorney wrote to the judge when submitting their draft decisions somewhere on this page to maintain feature parity.

rkreyhsig:
Title text has been changed:

| Before | After |
| --- | --- |
|screen shot 2018-12-18 at 3 45 47 pm|screen shot 2018-12-18 at 3 59 03 pm|

lowellrex:
@rkreyhsig Let's leave the heading as is and put the text "No active tasks" in the body of that container (where the tasks would normally go).

sneha-pai:
How about in the body of the container: 
"There are no active tasks at this moment."

@allyceh please refine if you get a chance

client/app/queue/TaskSnapshot.jsx
```diff
@@ -120,14 +129,19 @@ export class CaseSnapshot extends React.PureComponent<Props> {
       this.getAbbrevName(primaryTask.decisionPreparedBy) : null;
 
     return <React.Fragment>
-      { primaryTask.label &&
-        <React.Fragment>
-          <dt>{COPY.CASE_SNAPSHOT_TASK_TYPE_LABEL}</dt><dd>{this.getActionName()}</dd>
-        </React.Fragment> }
+      <dt>Assigned to</dt><dd>{primaryTask.assignedTo.cssId}</dd>
```
lowellrex: We already have this string as `CASE_SNAPSHOT_TASK_ASSIGNEE_LABEL` in COPY.json so let's use that here.

client/app/queue/TaskSnapshot.jsx
```diff
@@ -242,49 +246,64 @@ export class CaseSnapshot extends React.PureComponent<Props> {
...
-            </React.Fragment> }
-          {this.daysSinceTaskAssignmentListItem()}
-        </CaseDetailsDescriptionList>
+    return <div className="usa-grid" {...css({ marginTop: '3rem' })}>
+      <h2 {...sectionHeadingStyling}>
+        <a id="our-elemnt" {...anchorJumpLinkStyling}>Currently active tasks</a>
+      </h2>
+      <div {...sectionSegmentStyling}>
+        { this.props.primaryTask ? '' : 'No active tasks'}
+        { this.props.primaryTask && <table {...css({ width: '100%',
```
lowellrex: These two lines effectively define an if/else condition. Our standard pattern here is to move these type of things into an explicit if/else block outside of the return statement. Perhaps something like:
```jsx
let sectionBody = 'No active tasks';
if (this.props.primaryTask) {
  sectionBody = <table ...
```

client/app/queue/TaskSnapshot.jsx
```diff
@@ -242,49 +246,64 @@ export class CaseSnapshot extends React.PureComponent<Props> {
...
-              <dd>{appeal.assignedAttorney.full_name}</dd>
-            </React.Fragment> }
-          {this.daysSinceTaskAssignmentListItem()}
-        </CaseDetailsDescriptionList>
+    return <div className="usa-grid" {...css({ marginTop: '3rem' })}>
+      <h2 {...sectionHeadingStyling}>
+        <a id="our-elemnt" {...anchorJumpLinkStyling}>Currently active tasks</a>
+      </h2>
+      <div {...sectionSegmentStyling}>
+        { this.props.primaryTask ? '' : 'No active tasks'}
```
lowellrex: Let's move this string into `COPY.json`.

client/app/queue/TaskSnapshot.jsx
```diff
@@ -242,49 +246,64 @@ export class CaseSnapshot extends React.PureComponent<Props> {
...
-          { !taskAssignedToVso && appeal.assignedAttorney &&
-            <React.Fragment>
-              <dt>{COPY.CASE_SNAPSHOT_ASSIGNED_ATTORNEY_LABEL}</dt>
-              <dd>{appeal.assignedAttorney.full_name}</dd>
-            </React.Fragment> }
-          {this.daysSinceTaskAssignmentListItem()}
-        </CaseDetailsDescriptionList>
+    return <div className="usa-grid" {...css({ marginTop: '3rem' })}>
+      <h2 {...sectionHeadingStyling}>
+        <a id="our-elemnt" {...anchorJumpLinkStyling}>Currently active tasks</a>
```
lowellrex: Let's move this heading string into `COPY.json`.

client/app/queue/TaskSnapshot.jsx
```diff
@@ -242,49 +246,64 @@ export class CaseSnapshot extends React.PureComponent<Props> {
...
+      </h2>
+      <div {...sectionSegmentStyling}>
+        { this.props.primaryTask ? '' : 'No active tasks'}
+        { this.props.primaryTask && <table {...css({ width: '100%',
+          marginTop: 0 })}>
+          <tbody>
+            <tr>
+              <td {...leftTableCell} {...css({ width: '20%' })}>
```
lowellrex: Let's combine these two sets of css rules by including the width attribute in the `leftTableCellStyling` definition.

client/app/queue/TaskSnapshot.jsx
```diff
@@ -242,49 +246,64 @@ export class CaseSnapshot extends React.PureComponent<Props> {
...
+                <CaseDetailsDescriptionList>
+                  <dt>Assigned on</dt>
+                  <dd>{primaryTask && primaryTask.assignedOn &&
+                      moment(primaryTask.assignedOn).format('MM/DD/YYYY')}</dd>
+                  {this.daysSinceTaskAssignmentListItem()}
+                </CaseDetailsDescriptionList>
+              </td>
+              <td {...tableCellWithIcon}><GrayDot /><div {...grayLine} /></td>
+              <td {...tableCell} {...css({ width: '25%' })}>
```
lowellrex: Same thought as above. Let's combine these two sets of css rules by including the width attribute in the `tableCellStyling` definition (and we should probably rename this style to better represent what it is styling `taskActionCellStyling`).

client/app/queue/TaskSnapshot.jsx
```diff
@@ -242,49 +248,69 @@ export class CaseSnapshot extends React.PureComponent<Props> {
...
+                  </React.Fragment> }
+                { !taskAssignedToVso && appeal.assignedAttorney &&
+                  <React.Fragment>
+                    <dt>{COPY.TASK_SNAPSHOT_ASSIGNED_ATTORNEY_LABEL}</dt>
+                    <dd>{appeal.assignedAttorney.full_name}</dd>
+                  </React.Fragment> }
+                {this.legacyTaskInformation()}
+              </CaseDetailsDescriptionList>
+            </td>
+            <td {...tableCell} {...css({ width: '50%' })}>
```
lowellrex: Let's consolidate these styles and rename the styling to something that is more explicit about what the styling represents (the first style probably makes sense to be `taskInformationContainerStyling` and this style which will probably inherit and extend that style could be something like `taskActionsContainerStyling`). Let's do similar renaming with the other styles as well.

client/app/queue/TaskSnapshot.jsx
```diff
@@ -23,35 +22,44 @@ import { DateString } from '../util/DateUtil';
...
+  position: 'relative',
+  verticalAlign: 'top',
+  width: '45px'
+});
+
+const taskInformationContainerStyling = css({
+  border: 'none',
+  verticalAlign: 'top',
+  padding: '3px',
+  width: '25%'
```
lowellrex: If the only difference between this styling and `taskTimeContainerStyling` is the width, I think we could bust out the shared values into a different variable and pull them into both of these variables so that we don't forget to change a second place when we change the first.

client/app/queue/EvaluateDecisionView.jsx
allyceh: I agree seeing the notes + the name of the person they're evaluating would be helpful. I'll create a ticket to address that design-wise.

allyceh: Ticket to revisit design: https://github.com/department-of-veterans-affairs/caseflow/issues/8382

PR #8282, opened 2018-12-13
Remove extra RootTask method

lowellrex:
As currently written, we expect `RootTask.can_be_updated_by_user?` to always return the same result as `GenericTask.can_be_updated_by_user?` except in the case where [the condition in `RootTask.available_actions` evaluates to false](https://github.com/department-of-veterans-affairs/caseflow/blob/bd3d4a6c493932fb9f47029fb7a58018a557d148/app/models/tasks/root_task.rb#L13). This change is desirable because we should not be creating child tasks in those circumstances. This PR makes that change.

### DIFF
--- a/app/models/tasks/root_task.rb
+++ b/app/models/tasks/root_task.rb
@@ -42,10 +42,4 @@ class RootTask < GenericTask
       end
     end
   end
-
-  def can_be_updated_by_user?(user)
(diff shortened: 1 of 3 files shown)


lowellrex:
Thanks a bunch! The tests look great!

Issue #7985, opened 2018-11-26
Non-Comp Intake | Don't create EP for non-VBMS HLRs and SCs


shanear:
Compensation and Pension use VBMS to process HLRs and SCs. Other business lines do not, thus we should not create EPs for those. Instead we will use the DB objects directly in the Caseflow non-comp UI.

## Acceptance Criteria
When processing the intake of an HLR or SC, and the benefit type is not "Compensation" or "Pension":
- No rating issues should show in the first modal of the Add Issue flow. This means the first modal should only show up when there are prior completed decision issues for that benefit type (see: #7978)
- The "Establish EP" button should instead say "Establish Higher-Level Review" or "Establish Supplemental Claim" (depending on the form).
- When the "Establish [form]" button is clicked. No end products or contentions should be created.

## Technical Notes
- We may need to look at any logic for HLRs or SCs that assumes the presence of at least one `EndProductEstablishment`, as this is no longer guaranteed.

requires #7978

steve-gov:
Bullet 1 validated for HLR and SC.

Bullet 2 validated for HLR and SC.

<img width="938" alt="screen shot 2019-01-03 at 3 46 39 pm" src="https://user-images.githubusercontent.com/25468689/50667608-015a6a80-0f6f-11e9-9a17-e4bd5873b616.png">

Bullet 3 is technical and assumed correct.

Issue #5882, opened 2018-06-12
Intake | Show appropriate issue categories in Add Issues modal for HLRs and SCs


abbyraskinUSDS:
We should only show issue categories appropriate to the selected business line.

## Acceptance Criteria
- When intaking or editing issues on an HLR or SC, the "Issue Category" field on the second modal in the Add Issue flow should have only the options for the business line selected in the "Benefit Type" question on the previous page. See the category lists in https://github.com/department-of-veterans-affairs/appeals-design-research/issues/1043.

## Notes
requires https://github.com/department-of-veterans-affairs/appeals-design-research/issues/1043

leikkisa:
Are the current issue categories all of the "Compensation" issue categories?

steve-gov:
Validated. There will likely be some revisions before launch but no bigs.

PR #8291, opened 2018-12-13
No End Products on non-comp SC and HLR

pkarman:
connect #7985 

### Description
Adds new method `DecisionReview.non_comp?` and applies it to UI text and processing logic.

Note that `Appeal` now has a hardcoded `benefit_type` in order to satisfy the `non_comp?` method. That is intended, and will be resolved when #5882 is implemented.

### DIFF
--- a/client/app/intake/pages/higherLevelReview/finish.jsx
+++ b/client/app/intake/pages/higherLevelReview/finish.jsx
@@ -5,7 +5,7 @@ import Button from '../../../components/Button';
 import CancelButton from '../../components/CancelButton';
 import IssueCounter from '../../components/IssueCounter';
 import { completeIntake } from '../../actions/decisionReview';
-import { REQUEST_STATE } from '../../constants';
+import { REQUEST_STATE, FORM_TYPES } from '../../constants';
 import { issueCountSelector } from '../../selectors';
 
 class FinishNextButton extends React.PureComponent {
@@ -19,6 +19,14 @@ class FinishNextButton extends React.PureComponent {
     );
   }
 
+  buttonText = () => {
+    if (this.props.higherLevelReview.nonComp) {
+      return `Establish ${FORM_TYPES.HIGHER_LEVEL_REVIEW.shortName}`;
+    }
+
+    return 'Establish EP';
+  }
+
   render = () =>
     <Button
       name="finish-intake"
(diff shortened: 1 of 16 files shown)


client/app/intake/pages/higherLevelReview/finish.jsx
```diff
@@ -19,14 +19,22 @@ class FinishNextButton extends React.PureComponent {
     );
   }
 
+  buttonText = () => {
+    if (this.props.higherLevelReview.nonComp) {
+      return `Establish ${FORM_TYPES.HIGHER_LEVEL_REVIEW.shortName}`;
+    }
+
+    return 'Establish EP';
```
jiahuang: could shorten this to
return `Establish ${this.props.higherLevelReview.nonComp ? FORM_TYPES.HIGHER_LEVEL_REVIEW.shortName : "EP"};`

spec/feature/intake/supplemental_claim_spec.rb
```diff
@@ -780,13 +780,33 @@ def check_row(label, text)
...
+      sc, = start_supplemental_claim(veteran, is_comp: false)
       visit "/intake/add_issues"
 
       expect(page).to have_content("Add / Remove Issues")
       check_row("Form", Constants.INTAKE_FORM_NAMES.supplemental_claim)
       check_row("Benefit type", "Education")
+
+      expect(page).to_not have_content("Left knee granted")
+
+      click_intake_add_issue
```
jiahuang: maybe there should also be a test that sets up decision issues? those decision issues should show up on the add rating issues modal right?

client/app/intake/pages/higherLevelReview/finish.jsx
pkarman: true. I find a single-line ternary expression like that harder to read, personally.

spec/feature/intake/supplemental_claim_spec.rb
pkarman: do you mean as part of this `scenario`? or to cover some other aspect of this PR?

client/app/intake/pages/higherLevelReview/finish.jsx
jiahuang: i'm like a 2/10 on caring about this so if its easier to not be ternary that's fine with me.

spec/feature/intake/supplemental_claim_spec.rb
jiahuang: can either be part of this scenario or creating a new one. seems easiest just to add it in here

Issue #7978, opened 2018-11-26
AMA Intake | Let request issues contest decision issues (as well as rating issues)


shanear:
In order to chain request issues that don't have rating issues, we need decision issues to be contestable directly.

<img width="1151" alt="screen shot 2018-11-26 at 1 51 15 pm" src="https://user-images.githubusercontent.com/1596259/49037159-89d33780-f187-11e8-821b-9d1a294d145b.png">
<img width="1156" alt="screen shot 2018-11-26 at 1 51 28 pm" src="https://user-images.githubusercontent.com/1596259/49037168-8d66be80-f187-11e8-8061-d28f695dfa37.png">
<img width="1152" alt="screen shot 2018-11-26 at 1 51 34 pm" src="https://user-images.githubusercontent.com/1596259/49037171-8e97eb80-f187-11e8-9932-fc4d63c6f816.png">

## Acceptance Criteria
- When intaking or editing issues on a decision review, if there are any prior completed decision reviews with decision issues that are **not** connected to a rating issue, those issues should show up in the first modal of the Add Issues flow (the one in the presentation panel above).

- Those issues that appear on the Add Issues flow should be scoped by the selected benefit type. (i.e. if "Pension is selected as the benefit type, only prior "Pension" decision issues should show up). If the decision review is an a appeal, all benefit types should be included.

- If the decision issue was connected to a rating EP. Put the decision issue in the same list as the rest of the rating issues that are part of the rating associated with that EP.

- If the decision issue was not connected to a rating EP put them under a separate "Past decisions from [date]" section. Where [date] is the approximate decision date of the decision issue. (see https://github.com/department-of-veterans-affairs/caseflow/issues/7982)

- The copy for those issues should be as follows:
   - If there is a custom written appeals decision description, use that. If not,
      - If request issue does not have notes:
      "[Disposition] - [Request issue description]"
      - If request issue has notes:
      "[Disposition] - [Request issue description]. Notes: [Request issue notes]"

- The following existing validation patterns should apply to prior non-rating issues:
  - Multiple active issues reviewing the same non-rating decision should be ineligible (https://github.com/department-of-veterans-affairs/caseflow/issues/7291)
  - If HLR, then issues reviewing prior non-rating HLR decisions should be ineligible (https://github.com/department-of-veterans-affairs/caseflow/issues/7293)

## Technical Notes
It's also possible we want to store the decision issue of the contested rating issue if it exists. Not required, but worth considering.

[...]

PR #8292, opened 2018-12-14
ineligible decision issues

jiahuang:
Connects #7978

Adds an ineligibility message if there is another request issue contesting the same decision issue

decision_issue_ineligible

### DIFF
--- a/app/models/contestable_issue.rb
+++ b/app/models/contestable_issue.rb
@@ -58,12 +58,12 @@ class ContestableIssue
 
   def conflicting_request_issue_by_rating
     return unless rating_issue_reference_id
-    RequestIssue.find_active_by_rating_issue_reference_id(rating_issue_reference_id)
+   potentially_conflicting_request_issues.find_active_by_rating_issue_reference_id(rating_issue_reference_id)
   end
 
   def conflicting_request_issue_by_decision_issue
     return unless decision_issue_id
-    RequestIssue.find_active_by_contested_decision_id(decision_issue_id)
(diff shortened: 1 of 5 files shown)


app/models/contestable_issue.rb
```diff
@@ -48,19 +48,38 @@ def serialize
...
+  end
+
   def title_of_active_review
     conflicting_request_issue.try(:review_title)
   end
 
-  def conflicting_request_issue
+  def conflicting_request_issue_by_rating
     return unless rating_issue_reference_id
+    RequestIssue.find_active_by_rating_issue_reference_id(rating_issue_reference_id)
```
shanear:
```suggestion
   potentially_conflicting_request_issues.find_active_by_rating_issue_reference_id(rating_issue_reference_id)
```

shanear:
```
def potentially_conflicting_request_issues
  RequestIssue.where.not(review_request: contesting_decision_review)
end
```

app/models/contestable_issue.rb
```diff
@@ -48,19 +48,38 @@ def serialize
...
+  end
+
+  def conflicting_request_issue_by_decision_issue
+    return unless decision_issue_id
+    RequestIssue.find_active_by_contested_decision_id(decision_issue_id)
```
shanear:
```suggestion
   potentially_conflicting_request_issues.find_active_by_rating_issue_reference_id(rating_issue_reference_id)
```

Issue #7369, opened 2018-10-11
Legacy Opt-in | Reopen VACOLS issues if AMA issue is removed


kencheeto:
in #7337 we are going to start closing VACOLS issues and/or appeals when legacy issues are linked - if those AMA issues are removed, we will need to reopen the VACOLS issues / appeals.

requires https://github.com/department-of-veterans-affairs/caseflow/issues/7337

## Acceptance Criteria
- If an issue associated with an "opted-in" VACOLS issue (via #7337) is removed from any decision review via the Edit issues flow, then the "opted-in" VACOLS issue should be reopened with the previous disposition code

## Notes
We should confirm with Chris which code we should reopen with

## Technical Notes
- This also means that if the parent appeal was closed, because the opted-in issue was the last issue on that appeal, it must be reopened as well.
- There are methods that do this already (look at Ramp election code RampClosedAppeal, RampRollback)
- (An edge case that we don't need to address in this issue: the parent appeal was closed for some other reason (not "O"))

shanear:
This is working like a champ :100:

tested partial/full on remands and originals.

Everything works 👍

PR #8294, opened 2018-12-14
Reopen closed legacy issues if request issue is removed

leikkisa:
connects #7369
connects #8223


leikkisa:
Some comments:
- Left the vacols_id and vacols_sequence_id on request_issue for this PR, to keep the scope down on getting this piece done.
- I left the "error" column in the legacy issue opt ins table, but I am not using it yet. Should I take it out? We could also leave that for the last step which is some clean-up.
- We might want to consider another "renaming" at some point because "opt-in" is pretty mixed between "opt_in" and "optin", which has been going on before this PR, and also didn't want to include that here to scope it down, doesn't seem high priority at the very moment

leikkisa:
Also have a question if a review has one contention that has the legacy opt in special issue, are they all supposed to have the special issue, or just the one that is opted in??

Right now if it's there from the beginning, they'll all get it.  If it doesn't get added until the Edit flow, I think only any issues that have been changed as of that edit session will have the special issue.

app/models/appeal.rb
```diff
@@ -210,8 +210,10 @@ def benefit_type
 
   def create_issues!(new_issues)
     new_issues.each do |issue|
+      # temporary until ticket for appeals benefit type by issue is implemented
```
leikkisa: I don't think I added this comment, so looks like maybe it has been removed from master since.

app/models/legacy_issue_optin.rb
```diff
@@ -1,3 +1,91 @@
...
+
+  def legacy_appeal
+    LegacyAppeal.find_or_create_by_vacols_id(vacols_id)
+  end
+
+  private
+
+  def revert_open_remand_issues
+    # if this is happening, all remanded issues should have a disposition
+    # of "3" on a "HIS" appeal. This is rolling back and putting them back at "O"
```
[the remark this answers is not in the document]
leikkisa: Oh, I should update this comment. Technically the appeal gets re-opened in the step before this (before which the appeal has a HIS status). And re-opening it does check that it has the HIS status before it's re-opened.

I would be down to add a check that it's a "3" in the legacy appeal before putting it back at "O" again (wanted to do this for rollbacks too but took it out for now).  If it's not, then that would indicate that somebody else out there in the world changed the disposition on a HIS appeal, not sure if that's a thing that would or should happen.

app/mappers/issue_mapper.rb
```diff
@@ -64,6 +64,8 @@ def rename(issue_attrs)
     end
 
     def disposition_to_vacols_format(disposition)
+      # allow nil for rolling back issues
+      return disposition if disposition.nil?
```
pkarman: I think `return if disposition.nil?` is the same thing and maybe a trifle less misleading.

app/models/issue.rb
```diff
@@ -202,15 +206,6 @@ def remand_reasons
     @remand_reasons ||= self.class.remand_repository.load_remands_from_vacols(id, vacols_sequence_id)
   end
 
-  # For status (BFMPRO) of ADV or REM, for the most part, having a disposition means the issue is closed.
-  # On appeal where the status is REM (remanded) the issues with disposition "3" are still active.
-  def closed?
-    return false if disposition.nil?
-    return false if in_remand?
-    return true if legacy_appeal.remand? || legacy_appeal.advance? || !legacy_appeal.active?
```
pkarman: I see the definition of `active?` includes the checks for `disposition.nil?` and `in_remand` but are we now disregarding the status of the parent `legacy_appeal`?

app/models/issue.rb
```diff
@@ -318,6 +313,17 @@ def close_in_vacols!(vacols_id:, vacols_sequence_id:, disposition_code:)
       )
     end
 
+    def rollback_opt_in!(opt_in_issue)
```
pkarman: I like this

app/models/legacy_issue_optin.rb
```diff
@@ -1,3 +1,91 @@
 class LegacyIssueOptin < ApplicationRecord
   belongs_to :request_issue
+
+  VACOLS_DISPOSITION_CODE = "O".freeze # oh not zero
+
+  class << self
+    def related_remand_issues(vacols_id)
+      where(vacols_id: vacols_id, original_disposition_code: "3")
```
pkarman: could we use a named constant instead of `"3"` since at a glance I don't know what is special about `3`.

app/models/legacy_issue_optin.rb
```diff
@@ -1,3 +1,91 @@
...
+        vacols_id: vacols_id,
+        vacols_sequence_id: remand_issue.vacols_sequence_id,
+        disposition_code: VACOLS_DISPOSITION_CODE
+      )
+    end
+  end
+
+  def legacy_appeal_needs_reopened?
+    return false unless [nil, "3"].include? original_disposition_code
+    legacy_appeal.case_record.bfmpro == "HIS" && legacy_appeal.case_record.bfcurloc == "99"
```
pkarman: same comment about named constants instead of hardcoded strings here.

db/schema.rb
```diff
@@ -706,7 +711,7 @@
     t.string "decision_sync_error"
     t.string "ineligible_reason"
     t.string "vacols_id"
-    t.string "vacols_sequence_id"
```
pkarman: I see we are now storing `vacols_id` and `vacols_sequence_id` on both `RequestIssue` and `LegacyIssueOptIn` -- must we store the same data in both places?

spec/models/request_issues_update_spec.rb
```diff
@@ -4,8 +4,6 @@
   before do
     Time.zone = "America/New_York"
     Timecop.freeze(Time.utc(2018, 5, 20))
-
-    review.create_issues!(existing_request_issues)
```
pkarman: Why remove this here and copy it to `before` blocks inside each context?

spec/models/request_issues_update_spec.rb
```diff
@@ -226,7 +252,7 @@
           hash_including(
             veteran_file_number: review.veteran_file_number,
             contention_descriptions: ["Service connection for cancer was denied"],
-            special_issues: []
+            special_issues: [{ code: "VO", narrative: "AMA SOC/SSOC Opt-in" }]
```
pkarman: I know this changes because one of the 2 existing request issues now has a legacy opt-in, but maybe we need another test where it doesn't, to verify that `special_issues` changes depending on whether a request issue has a legacy opt-in.

spec/services/legacy_optin_manager_spec.rb
```diff
@@ -1,47 +1,252 @@
 describe LegacyOptinManager do
+  before do
+    Timecop.freeze(Time.utc(2018, 4, 24, 12, 0, 0))
```
pkarman: anything special about this date?

spec/services/legacy_optin_manager_spec.rb
```diff
@@ -1,47 +1,252 @@
...
+          expect(already_closed_case.bfddec).to eq(1.year.ago.to_date)
+        end
+
+        context "when issues are rolled back on a closed appeal" do
+          before do
+            LegacyOptinManager.new(decision_review: appeal).process!
+          end
+
+          it "reopens an undecided appeal and rollsback the dispositions" do
+            expect(vacols_issue("undecided", 2).issdc).to eq("O")
```
pkarman: why `O` instead of `LegacyIssueOptin::VACOLS_DISPOSITION_CODE` ?

spec/services/legacy_optin_manager_spec.rb
```diff
@@ -1,47 +1,252 @@
...
+    create(:request_issue, vacols_id: remand_case.bfkey, vacols_sequence_id: hlr_remand_issue.issseq)
+  end
+
+  let(:already_closed_case) do
+    create(:case,
+           :status_complete,
+           bfkey: "closed",
+           case_issues: [closed_issue1, closed_issue2],
+           bfdsoc: 1.day.ago,
+           bfddec: 1.year.ago)
```
pkarman: since we repeat this date a lot `1.year.ago` it might be clearer to make it its own `let` named var, something like `original_disposition_date` or something else meaningful. When I see it tested for later, I don't understand the significance of `1.year.ago`

spec/services/legacy_optin_manager_spec.rb
```diff
@@ -1,47 +1,252 @@
...
+          it "rollsback the disposition and date and does not change appeals" do
+            undecided_optin1.create_rollback!
+            remand_optin1.create_rollback!
+            closed_optin1.create_rollback!
+
+            subject
+
+            expect(vacols_issue("undecided", 1).issdc).to be_nil
+            expect(vacols_issue("undecided", 1).issdcls).to be_nil
+            expect(vacols_issue("remand", 1).issdc).to eq("3")
```
pkarman: instead of `3` or `G` could we use named vars (e.g. `original_disposition`) so that it's more meaningful?

spec/services/legacy_optin_manager_spec.rb
```diff
@@ -1,47 +1,252 @@
...
+          subject
+          expect(vacols_issue("remand", 1).reload.issdc).to eq("3")
+          expect(vacols_issue("remand", 1).issdcls).to eq(remand_optin1.original_disposition_date)
+          expect(vacols_issue("remand", 2).issdc).to eq("3")
+          expect(vacols_issue("remand", 2).issdcls).to eq(remand_optin2.original_disposition_date)
+
+          # check that remand opted in from another decision review is also rolled back
+          expect(hlr_remand_issue.reload.issdc).to eq("3")
+
+          expect(remand_case.reload.bfmpro).to eq("HIS")
```
pkarman: again, because these tests are all 100+ lines separated from their `let` definitions, I don't know the significance of the hardcoded strings. I think the test is comparing the original values, so something named that way would help me.

spec/models/request_issues_update_spec.rb
leikkisa: For some reason, doing it here was giving me problems with the vacols case and vacols issues not existing yet when the create_issues was running, but it wouldn't let me put those "lets" into the before do block either.

db/schema.rb
leikkisa: Shane and I talked about a few different approaches to cleaning up the data model on this (also keeping in mind that request_issue is also currently for the connected but not opted in issues).  I think I got a little overwhelmed though, because it was another layer of complexity, so he said it was okay to do it in another PR if it would help get this one done faster.

app/mappers/issue_mapper.rb
leikkisa: 👍

app/models/issue.rb
leikkisa: I don't feel qualified to determine this :-P. It was causing me problems that an issue with a disposition on an active appeal was not considered closed, so I added a check if it had an "O" disposition, it would be closed.  I asked Shane to check on it, and he said if all of the issues on an active appeal have a disposition, then we can close it, and this change is a result of that conversation.

I think we should have a chat though, in case we're forgetting/losing something by taking this out.  Like, maybe we shouldn't use this version of "closed?" for determining whether to close an appeal, but maybe we do want to use it to determine issue eligibility for opt in, and if something has a disposition on an active appeal it's still eligible.  I think this might be missing the "whitelist" codes in that case though.

app/models/legacy_issue_optin.rb
leikkisa: 👍

spec/models/request_issues_update_spec.rb
leikkisa: There are tests for the decision reviews that test the various paths for whether to include this. I added the opt-in here not to test the special issue, but to test the rollback since it's happening here. I'll take more of a look at those tests though.

This did bring up other questions for me on the special issue though... like it seems to me like only the opted-in issues should get this designation, or if a veteran selects "yes" to withdrawing their issues then maybe everything should get it, but right now it's in a weird in between state where some issues could not have it, and some issues could have it, but not all of the issues that have it are actually opted in.

spec/services/legacy_optin_manager_spec.rb
leikkisa: No, I just wanted the date to be stable. Not positive that I really need this, but I there is some date dependent logic and I wasn't sure if without this if the date would get changed during tests.

spec/services/legacy_optin_manager_spec.rb
leikkisa: 👍

spec/services/legacy_optin_manager_spec.rb
leikkisa: 👍

app/models/legacy_issue_optin.rb
leikkisa: I updated the "3" but not "HIS" or "99".

app/models/legacy_issue_optin.rb
```diff
@@ -1,3 +1,92 @@
...
+    Issue.close_in_vacols!(
+      vacols_id: vacols_id,
+      vacols_sequence_id: vacols_sequence_id,
+      disposition_code: VACOLS_DISPOSITION_CODE
+    )
+    update!(optin_processed_at: Time.zone.now)
+  end
+
+  def create_rollback!
+    update(rollback_created_at: Time.zone.now)
```
kencheeto: does this need to be `update!`?

app/services/legacy_optin_manager.rb
```diff
@@ -51,35 +26,25 @@ def process!
 
   def affected_legacy_appeals
     legacy_appeals = []
-    request_issues_with_legacy_issues.each do |request_issue|
-      legacy_appeals << legacy_appeal(request_issue.vacols_id) # TODO: get from legacy_issue_optin
+    legacy_issue_opt_ins.each do |legacy_issue_opt_in|
+      legacy_appeals << legacy_issue_opt_in.legacy_appeal
```
kencheeto: this could be refactored a bit:
```ruby
def affected_legacy_appeals
  legacy_issue_opt_ins.map(&:legacy_appeal).uniq
end
```

app/mappers/issue_mapper.rb
```diff
@@ -11,7 +11,7 @@ module IssueMapper
     vacols_id: :isskey
   }.freeze
 
-  ALLOWED_DISPOSITION_CODES = %w[1 3 4 5 6 8 O P].freeze
+  ALLOWED_DISPOSITION_CODES = %w[1 3 4 5 6 8 O P G X L].freeze
```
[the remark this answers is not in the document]
leikkisa: They are in a document called VACOLS DISPOSITIONS BY ID, the new ones are the "failure to respond" and the "manlincon remand" which in conversation we decided are the ones that can be opted in in addition to regular remand and no disposition (still pending feedback from AMO).  I added them here because we're using previously existing code for updating the dispositions that checks against this list.

Do you want me to add a comment in the code as well or did you mean it would be useful to add a comment right here?

app/models/issue.rb
```diff
@@ -132,15 +132,20 @@ def type_hash
...
+    !active?
+  end
+
   def allowed?
     disposition == :allowed
   end
 
   def remanded?
-    disposition == :remanded
+    disposition == :remanded || disposition == :manlincon_remand
```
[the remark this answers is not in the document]
leikkisa: No, this is a symbolized version of the disposition description in vacols dispositions by ID (one says "Remanded" and the other says "Manlicon remand", so if I put :manlicon_remanded it wouldn't work.

app/models/issue.rb
leikkisa: Just to close the loop on this, I check with Chris, and added something that prevents issues on appeals in history status from appearing active, and other than that it looks correct based on our current understanding, pending any additional information when Shane checks in with AMO.

app/services/legacy_optin_manager.rb
leikkisa: 👍

spec/models/request_issues_update_spec.rb
leikkisa: Made another ticket for this to keep this one scoped down.

spec/services/legacy_optin_manager_spec.rb
leikkisa: 👍

app/models/legacy_issue_optin.rb
leikkisa: 👍

app/models/decision_review.rb
```diff
@@ -165,12 +164,30 @@ def serialized_legacy_appeals
...
+  end
+
   def special_issues
     [].tap do |specials|
-      specials << vacols_optin_special_issue if needs_vacols_optin_special_issue?
+      specials << vacols_optin_special_issue if request_issues.any?(&:legacy_issue_optin)
     end
   end
 
+  def create_legacy_issue_optin(request_issue)
```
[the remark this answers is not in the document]
leikkisa: 👍

spec/services/legacy_optin_manager_spec.rb
```diff
@@ -1,47 +1,257 @@
...
+    Timecop.freeze(Time.zone.now)
+  end
+
+  let(:user) { Generators::User.build }
+  let(:closed_disposition_date) { 1.year.ago }
+  let!(:appeal) { create(:appeal, request_issues: request_issues) }
+  let(:request_issues) { [undecided_ri1, undecided_ri2, remand_ri1, remand_ri2, closed_ri1, closed_ri2] }
+  let(:legacy_opt_in_manager) { LegacyOptinManager.new(decision_review: appeal) }
+
+  let!(:undecided_issue1) { create(:case_issue, issseq: 1) }
```
kencheeto: it's really hard to follow what test setup is needed for each test case. Could we push any testcase-specific setup into the testcases themselves?

spec/services/legacy_optin_manager_spec.rb
```diff
@@ -1,47 +1,257 @@
...
+               vacols_id: closed_ri1.vacols_id,
+               vacols_sequence_id: closed_ri1.vacols_sequence_id)
+      end
 
-    subject { LegacyOptinManager.new(decision_review: appeal) }
+      let(:issue) { Issue.load_from_vacols(vacols_issue("undecided", 1).attributes) }
 
-    it "closes VACOLS issue with disposition O" do
-      subject.process!
+      context "when there are still open issues on the appeal" do
```
kencheeto: One code smell for detecting when test setup could be pushed down is to see context/describe blocks that describe setup but there isn't any test setup here. I would expect some setup within this context to be creating open issues on the appeal, for example

spec/services/legacy_optin_manager_spec.rb
```diff
@@ -1,47 +1,257 @@
...
+          follow_up_appeal_issues = VACOLS::CaseIssue.where(isskey: "remandP")
+          expect(follow_up_appeal_issues.count).to eq(3)
+        end
+
+        it "closes the closed issues and does not re-close the already closed appeal" do
+          subject
+          expect(vacols_issue("closed", 1).issdc).to eq(LegacyIssueOptin::VACOLS_DISPOSITION_CODE)
+          expect(vacols_issue("closed", 1).issdcls).to eq(Time.zone.today)
+          expect(vacols_issue("closed", 2).issdc).to eq(LegacyIssueOptin::VACOLS_DISPOSITION_CODE)
+          expect(vacols_issue("closed", 2).issdcls).to eq(Time.zone.today)
```
kencheeto: does this test case require having multiple vacols issues? or are we asserting properties on both because there happens to be two?

app/models/legacy_issue_optin.rb
```diff
@@ -1,3 +1,92 @@
...
+  def opt_in!
+    Issue.close_in_vacols!(
+      vacols_id: vacols_id,
+      vacols_sequence_id: vacols_sequence_id,
+      disposition_code: VACOLS_DISPOSITION_CODE
+    )
+    update!(optin_processed_at: Time.zone.now)
+  end
+
+  def create_rollback!
```
kencheeto: would a better name for this be `flag_for_rollback!` or something? `create` idiomatically refers to creating a new db record or instance of something

app/models/legacy_issue_optin.rb
```diff
@@ -1,3 +1,92 @@
...
+    self.class.related_remand_issues(vacols_id).each do |remand_issue|
+      Issue.close_in_vacols!(
+        vacols_id: vacols_id,
+        vacols_sequence_id: remand_issue.vacols_sequence_id,
+        disposition_code: VACOLS_DISPOSITION_CODE
+      )
+    end
+  end
+
+  def legacy_appeal_needs_reopened?
```
kencheeto: `legacy_appeal_needs_to_be_reopened?` ?

app/services/legacy_optin_manager.rb
```diff
@@ -1,47 +1,22 @@
...
-          next unless vacols_id == legacy_appeal.vacols_id
-
-          # gut checks
-          unless open_legacy_issues.map(&:vacols_sequence_id).map(&:to_s).include?(vacols_sequence_id)
-            fail "VACOLS issue #{vacols_id} sequence #{vacols_sequence_id} is already closed"
-          end
-
-          # close it
-          close_legacy_issue_in_vacols(request_issue.legacy_issue_optin)
+      pending_rollbacks.each(&:rollback!)
```
kencheeto: We're wrapping all the VACOLS changes in a transaction, but do the caseflow db changes also need to be in a transaction? Thinking about a partial processing that fails, we might mark an issue in caseflow as rolled back even though the VACOLS db changes don't get committed.

app/mappers/issue_mapper.rb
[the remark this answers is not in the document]
leikkisa: Ok, I'm going to add a comment about that document here. I don't fully understand why certain dispositions are here in the allowed list, but can at least point to what they codes mean.

app/services/legacy_optin_manager.rb
leikkisa: Goooood catch! Adding now.

app/models/legacy_issue_optin.rb
leikkisa: 👍

app/models/legacy_issue_optin.rb
leikkisa: 👍

Issue #7515, opened 2018-10-23
AMA Intake | Show active non-rating issues in Add Issue flow


shanear:

shanear:
@carolaponce this is based on our conversation yesterday. We initially leaned towards option 1, but after writing the story, I think there are some big problems with that approach.

Modal 1 is centered around past decisions. We are essentially asking "Does the claimant disagree with any of these decisions?". The question would be different for active issues, if we put them on that page. We'd be asking "Is the claimants issue the same as any of these issues?". I'm not sure we should be asking two questions on the same page.

This is why me and ken were leaning towards putting them on another modal, between 1 and 2.

carolaponce:
**UI for Adding non-rating issues**  
If user selects an existing issue in the system, then the screen refreshes. 
If user selects no it doesn't match, then go to existing screen for entering an issue.

carolaponce:
![intake active non-rating issue category](https://user-images.githubusercontent.com/11700545/48168990-5bf48480-e2a6-11e8-8392-d70fbd256719.png)
![intake active non-rating issue no](https://user-images.githubusercontent.com/11700545/48168991-5bf48480-e2a6-11e8-8a60-ce7e9954fc4e.png)
![intake active non-rating issue select](https://user-images.githubusercontent.com/11700545/48168992-5c8d1b00-e2a6-11e8-9973-bcade7f9fc08.png)
![intake active non-rating issue yes](https://user-images.githubusercontent.com/11700545/48168993-5c8d1b00-e2a6-11e8-82c4-9538a56e07cb.png)

jiahuang:
This can be validated by:
1. creating a decision review with a nonrating issue
2. create another decision review for the same veteran. Go to the nonrating issues modal and select the same category as the issue used for step 1
3. should see the radio buttons for step 1 pop up

steve-gov:
@jiahuang This validates except for bullet point 1.

> "Issue description" field, "Decision date" field, and "Add this issue" button should be hidden if the "Issue category" field is not set.

![screencapture3](https://user-images.githubusercontent.com/25468689/50845317-8e6c3d80-133a-11e9-917d-2e2dde5d71d2.gif)

PR #8301, opened 2018-12-15
adding active nonrating issues to modal

jiahuang:
Connects #7515

### DIFF
--- a/app/models/decision_review.rb
+++ b/app/models/decision_review.rb
@@ -170,8 +170,7 @@ class DecisionReview < ApplicationRecord
   end
 
   def active_nonrating_request_issues
-    @active_nonrating_request_issues ||= EndProductEstablishment.by_veteran_file_number(veteran.file_number)
-      .select(&:status_active?).map(&:request_issues).flatten.select(&:nonrating?)
+    @active_nonrating_request_issues ||= RequestIssue.nonrating.where(veteran_file_number: veteran_file_number).where.not(id: request_issues.map(&:id)).select(&:status_active?)
   end
 
   private
--- a/app/models/decision_review.rb
+++ b/app/models/decision_review.rb
@@ -55,8 +55,13 @@ class DecisionReview < ApplicationRecord
     end
   end
 
+  def non_comp?
+    !ClaimantValidator::BENEFIT_TYPE_REQUIRES_PAYEE_CODE.include?(benefit_type)
+  end
+
   def serialized_ratings
     return unless receipt_date
+    return if non_comp?
 
     cached_serialized_ratings.each do |rating|
       rating[:issues].each do |rating_issue_hash|
@@ -141,7 +146,7 @@ class DecisionReview < ApplicationRecord
       {
         vacols_id: legacy_appeal.vacols_id,
         date: legacy_appeal.nod_date,
-        eligible_for_soc_opt_in: legacy_appeal.eligible_for_soc_opt_in?,
+        eligible_for_soc_opt_in: legacy_appeal.eligible_for_soc_opt_in?(receipt_date),
         issues: legacy_appeal.issues.map(&:intake_attributes)
       }
     end
@@ -166,6 +171,7 @@ class DecisionReview < ApplicationRecord
   end
 
   def contestable_issues
+    return contestable_issues_from_decision_issues if non_comp?
     contestable_issues_from_ratings + contestable_issues_from_decision_issues
   end
 
@@ -205,7 +211,7 @@ class DecisionReview < ApplicationRecord
   def matchable_legacy_appeals
     @matchable_legacy_appeals ||= LegacyAppeal
       .fetch_appeals_by_file_number(veteran_file_number)
-      .select(&:matchable_to_request_issue?)
+      .select { |appeal| appeal.matchable_to_request_issue?(receipt_date) }
   end
 
   def active_matchable_legacy_appeals
--- a/app/models/decision_review.rb
+++ b/app/models/decision_review.rb
@@ -176,7 +176,10 @@ class DecisionReview < ApplicationRecord
   end
 
   def active_nonrating_request_issues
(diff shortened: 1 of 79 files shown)


app/models/decision_review.rb
```diff
@@ -168,6 +169,11 @@ def contestable_issues
     contestable_issues_from_ratings + contestable_issues_from_decision_issues
   end
 
+  def active_nonrating_request_issues
+    @active_nonrating_request_issues ||= EndProductEstablishment.by_veteran_file_number(veteran.file_number)
```
jiahuang: Should we add both end product active status & veteran file number to request issue to make this query not as big?

app/models/decision_review_intake.rb
```diff
@@ -9,6 +9,7 @@ def ui_hash(ama_enabled)
       legacyAppeals: detail.serialized_legacy_appeals,
       ratings: detail.serialized_ratings,
       requestIssues: detail.request_issues.map(&:ui_hash),
+      activeNonratingRequestIssues: detail.active_nonrating_request_issues.map(&:ui_hash),
```
jiahuang: Currently sending up all active nonrating request issues. If any of these are selected, the ineligibility note & ineligible_due_to_id comes from the selected issue.

app/models/decision_review.rb
shanear: Yeah i think we probably want this query made from a `RequestIssue` perspective. Since not all request issues will have an `EndProductEstablishment`, but we still want to select those.

```suggestion
    @active_nonrating_request_issues ||= RequestIssue.nonrating.select(&:status_active?)
```

app/models/decision_review.rb
```diff
@@ -168,6 +169,11 @@ def contestable_issues
...
+      .select(&:status_active?).map(&:request_issues).flatten.select(&:nonrating?)
```
shanear:
```suggestion
```

app/models/end_product_establishment.rb
```diff
@@ -94,6 +94,10 @@ def order_by_sync_priority
       active.order("last_synced_at IS NOT NULL, last_synced_at ASC")
     end
 
+    def by_veteran_file_number(veteran_file_number)
+      where(veteran_file_number: veteran_file_number)
+    end
```
shanear: Not needed if the change above is made.

app/models/request_issue.rb
```diff
@@ -15,6 +15,7 @@ class RequestIssue < ApplicationRecord
   validates :ineligible_reason, exclusion: { in: ["untimely"] }, if: proc { |reqi| reqi.untimely_exemption }
 
   enum ineligible_reason: {
+    duplicate_of_nonrating_issue_in_active_review: "duplicate_of_nonrating_issue_in_active_review",
     duplicate_of_issue_in_active_review: "duplicate_of_issue_in_active_review",
```
shanear: might want to rename the tag here, and then add a task to maybe go migrate the values eventually.

```suggestion
    duplicate_of_rating_issue_in_active_review: "duplicate_of_issue_in_active_review",
```

app/models/request_issue.rb
```diff
@@ -211,7 +217,9 @@ def contested_issue
   end
 
   def title_of_active_review
-    duplicate_of_issue_in_active_review? ? ineligible_due_to.review_title : nil
+    if duplicate_of_issue_in_active_review? || duplicate_of_nonrating_issue_in_active_review?
```
shanear: Maybe we should rename `duplicate_of_issue_in_active_review?` to `duplicate_of_rating_issue_in_active_review?` and then make this method:

```
def duplicate_of_issue_in_active_review?
  duplicate_of_rating_issue_in_active_review? || duplicate_of_nonrating_issue_in_active_review?
end
```

client/app/intake/components/NonratingRequestIssueModal.jsx
```diff
@@ -106,10 +149,55 @@ class NonratingRequestIssueModal extends React.Component {
...
+            onChange={this.decisionDateOnChange} />
+        </div>
+
+        <TextField
+          name="Issue description"
+          strongLabel
+          value={description}
+          onChange={this.descriptionOnChange} />
+      </React.Fragment> : null;
+
```
shanear: This blurb is getting pretty beefy. I feel like it might be best to factor out some of this stuff into another method or subcomponent or something. Cool if that is too much though

client/app/styles/_intake.scss
```diff
@@ -141,11 +141,17 @@
...
 
+.add-nonrating-request-issue-description {
   .decision-date {
     margin-top: 20px;
     width: 8em;
   }
+
+  .cf-form-radio {
+    margin-top: 20px;
+  }
```
shanear: Custom CSS?

client/app/intake/components/NonratingRequestIssueModal.jsx
jiahuang: going to keep it as it is for this pr for now. there's lots of events and state tied to the current component in here, and i feel like chaining it to another component doesn't really fix the problem.

client/app/styles/_intake.scss
jiahuang: its just stylizing the radio button when its in our modal, need some kind of margin otherwise everything looks too cluttered. Open to suggestion of better ways to do this

client/app/intake/components/NonratingRequestIssueModal.jsx
```diff
@@ -106,10 +148,55 @@ class NonratingRequestIssueModal extends React.Component {
...
+        vertical
+        label={<h3>Does issue {issueNumber} match any of the issues actively being reviewed?</h3>}
+        name="rating-radio"
+        options={nonratingRequestIssueOptions}
+        key={category}
+        value={selectedNonratingIssueId}
+        onChange={this.selectedNonratingIssueIdOnChange}
+      /> : null;
+
+    const additionalDetails = selectedNonratingIssueId === NO_MATCH_TEXT || !nonratingRequestIssueSelection ?
```
pkarman: I know we love the ternary pattern, but this is pretty hard to read. Could we change it to an if/else pattern?

PR #8305, opened 2018-12-15
detect more transient bgs errors, encode as utf-8

kencheeto:
Detects a couple additional kinds of transient errors that have been occurring lately, and also encodes the error body before checking for known messages to avoid masking encoding errors.

Starting to feel like it'll be good to extract this error handling out of the EPE class.

### DIFF
--- a/app/models/end_product_establishment.rb
+++ b/app/models/end_product_establishment.rb
@@ -45,6 +45,7 @@ class EndProductEstablishment < ApplicationRecord
     # rubocop:disable Metrics/MethodLength
     def self.from_bgs_error(error, epe)
       error_message = if error.try(:body)
+                        # https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3124/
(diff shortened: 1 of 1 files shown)


app/models/end_product_establishment.rb
```diff
@@ -42,8 +42,15 @@ def initialize(error, end_product_establishment)
     end
 
     # rubocop:disable Metrics/CyclomaticComplexity
+    # rubocop:disable Metrics/MethodLength
     def self.from_bgs_error(error, epe)
-      case error.try(:body) || error.message
+      error_message = if error.try(:body)
+                        # https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3124/
+                        error.body.encode("UTF-8", invalid: :replace, undef: :replace, replace: "")
```
kencheeto: https://robots.thoughtbot.com/fight-back-utf-8-invalid-byte-sequences

PR #8311, opened 2018-12-17
Fix alignment of errors in the reader sidebar

### LINKED ISSUE #4034: Category error state looks weird
When there is a server error with updating a category, the error state messes up the UI:
 screen shot 2017-11-30 at 11 12 55 pm

hschallhorn:
Resolves #4034

### Description
display: float was inconsistent with the the rest of the css alignment of the sidebar, causing overlap in the elements.

### Acceptance Criteria
- [ ] Errors display correctly (see below)
screen shot 2018-12-17 at 4 08 40 pm
screen shot 2018-12-17 at 11 22 52 am

### Testing Plan
1. In document_controller.rb's update method, add `render status: 500`
2. Select or deselect a document category to display the alert
3. In tag_controller.rb's create method, add `render status: 500`
4. Add a tag to document to display the alert
5. In annotation_controller.rb's create method, add `render status: 500`
4. Add a comment to document to display the alert


client/app/reader/SideBarComments.jsx
```diff
@@ -29,21 +30,22 @@ class SideBarComments extends PureComponent {
         <Button
           name="AddComment"
           onClick={this.handleAddClick}>
-          <span>{ plusIcon() } &nbsp; Add a comment</span>
+          <span>{plusIcon()} &nbsp; Add a comment</span>
         </Button>
       </span>
+      <div style={{ clear: 'both' }}></div>
```
mdbenjam: What does this line do?

mdbenjam: Or rather is it still necessary if we remove the float property below?

hschallhorn: Because the span above the alert is floated to the right, the alert will overlap the button.
screen shot 2018-12-18 at 11 44 06 am
Clearing float forces the alert to occupy a new row rather than overlapping the button.
We could alternatively float the rest of the content (dropdown and tag picker) to remain consistent in the positioning rather than remove the alert's float and use this hack. I'm now thinking that may have been the better solution.

mdbenjam: Gotcha, I see the reason now. This is fine, one other thing that might be worth trying is flipping the location of the alert and the button. Putting the alert at the top. But the change looks good. So I'll leave it up to you.

PR #8312, opened 2018-12-17
Add more BE validations for judge checkout

### LINKED ISSUE #8179: [BE] Add BE validation to ensure that remanded issues have remand reasons
AC:
BE validation should be present to ensure that remanded issues have remand reasons

### LINKED ISSUE #8235: [BE] Add validation that every request issue has a decision issue for new attorney magic
AC:
Validation is added that every request issue has a decision issue for new attorney magic

### LINKED ISSUE #8270: [BE] Add validation that during checkout number of issues passed match what the appeal has in the DB
Add validation that during checkout number of issues passed match what the appeal has in the DB

aroltsch:
Resolves #8270 
Resolves #8235 
Resolves #8179 

### Description
1. Add validation that during checkout number of issues passed match what the appeal has in the DB for all three paths: legacy checkout, ama "old" checkout and ama "new" checkout
2. Add validation that remand reasons are present when disposition is remanded for all three paths
3. Add validation that at least one decision issue is present for each request issue


app/models/concerns/case_review_concern.rb
```diff
@@ -15,98 +15,6 @@ def appeal
                 end
   end
 
-  def update_task_and_issue_dispositions
```
aroltsch: Moved it into its own concern

app/models/concerns/issue_updater.rb
```diff
@@ -0,0 +1,129 @@
...
+    return unless appeal
+    # We will always delete and re-create decision issues on attorney/judge checkout
+    decision_issue_ids_to_delete = appeal.decision_issues.map(&:id)
+    DecisionIssue.where(id: decision_issue_ids_to_delete).destroy_all
+
+    issues.each do |issue_attrs|
+      request_issues = appeal.request_issues.where(id: issue_attrs[:request_issue_ids])
+      next if request_issues.empty?
+      decision_issue = DecisionIssue.create!(
+        disposition: issue_attrs[:disposition],
```
aroltsch: `disposition` is a required field for `decision_issues` so this will fail if missing

app/models/concerns/issue_updater.rb
```diff
@@ -0,0 +1,129 @@
...
+        description: issue_attrs[:description],
+        benefit_type: issue_attrs[:benefit_type],
+        participant_id: appeal.veteran.participant_id
+      )
+      request_issues.each do |request_issue|
+        RequestDecisionIssue.create!(decision_issue: decision_issue, request_issue: request_issue)
+      end
+      create_remand_reasons(decision_issue, issue_attrs[:remand_reasons] || [])
+    end
+    unless appeal.every_request_issue_has_decision?
```
aroltsch: This will check that every request issue has a decision

app/models/decision_issue.rb
```diff
@@ -1,8 +1,9 @@
 class DecisionIssue < ApplicationRecord
   validates :disposition, inclusion: { in: Constants::ISSUE_DISPOSITIONS_BY_ID.keys.map(&:to_s) },
-                          allow_nil: true, if: :appeal?
+                          if: :appeal?
   validates :benefit_type, inclusion: { in: Constants::BENEFIT_TYPES.keys.map(&:to_s) },
-                           allow_nil: true, if: :appeal?
+                           if: :appeal?
+  validates :description, presence: true, if: :appeal?
```
aroltsch: @pkarman I just want to make sure these changes are okay from your teams' side. I also noticed decision_issues have a direct reference to an appeal. Was it to avoid multiple joins?

aroltsch: We can't really enforce that at the DB level because we need to check the decision_review type to be an Appeal. 

Yes, I was referring to the `decision_review`

app/models/concerns/issue_updater.rb
```diff
@@ -0,0 +1,143 @@
...
+
+  def fail_if_no_dispositions!
+    if (issues || []).map { |issue| issue[:disposition].present? }.uniq != [true]
+      msg = "Issues in the request are missing dispositions"
+      fail Caseflow::Error::AttorneyJudgeCheckoutError, message: msg
+    end
+  end
+
+  def fail_if_count_mismatch!
+    if (issues || []).count != (legacy? ? appeal.issues.count : appeal.request_issues.count)
```
aroltsch: I need to exclude previously decided issues for this count - legacy cases. Working on it now

app/models/concerns/issue_updater.rb
```diff
@@ -0,0 +1,144 @@
...
+
+  def fail_if_no_dispositions!
+    if (issues || []).map { |issue| issue[:disposition].present? }.uniq != [true]
+      msg = "Issues in the request are missing dispositions"
+      fail Caseflow::Error::AttorneyJudgeCheckoutError, message: msg
+    end
+  end
+
+  def fail_if_count_mismatch!
+    existing_issues_count = legacy? ? appeal.undecided_issues.count : appeal.eligible_request_issues.count
```
aroltsch: For legacy cases, only use issues that were not previously decided. For ama appeals, only eligible request issues.

app/controllers/case_reviews_controller.rb
```diff
@@ -8,6 +8,11 @@ class CaseReviewsController < ApplicationController
     handle_non_critical_error("case_reviews", e)
   end
 
+  rescue_from Caseflow::Error::AttorneyJudgeCheckoutError do |e|
```
mdbenjam: Do we really need this? Doesn't this all happen without needing this rescue from?

app/models/appeal.rb
```diff
@@ -78,6 +78,10 @@ def attorney_case_reviews
     tasks.map(&:attorney_case_reviews).flatten
   end
 
+  def every_request_issue_has_decision?
+    request_issues.map { |request_issue| request_issue.decision_issues.present? }.uniq == [true]
```
mdbenjam: I think you can just use `all?` https://apidock.com/ruby/Enumerable/all%3F

app/models/concerns/issue_updater.rb
```diff
@@ -0,0 +1,144 @@
...
+  def update_issue_dispositions_in_caseflow!
+    use_ama_decision_issues? ? delete_and_create_decision_issues! : update_issue_dispositions_deprecated!
+  rescue ActiveRecord::RecordInvalid => e
+    raise Caseflow::Error::AttorneyJudgeCheckoutError, message: e.message
+  end
+
+  def delete_and_create_decision_issues!
+    return unless appeal
+    # We will always delete and re-create decision issues on attorney/judge checkout
+    DecisionIssue.where(id: appeal.decision_issues.map(&:id)).destroy_all
```
mdbenjam: I think we talked about this last time but `appeal.decision_issues.destroy_all` doesn't work right?

mdbenjam: And is this all in a transaction block?

app/models/concerns/issue_updater.rb
```diff
@@ -0,0 +1,144 @@
...
+  end
+
+  def fail_if_invalid_issues_attrs!
+    return if is_a?(AttorneyCaseReview) && omo_request?
+
+    fail_if_count_mismatch!
+    fail_if_no_dispositions!
+  end
+
+  def fail_if_no_dispositions!
```
mdbenjam: And disposition existing are enforced using rails validations for decision issues 👍

app/models/concerns/issue_updater.rb
aroltsch: Yes and yes!

app/controllers/case_reviews_controller.rb
aroltsch: you are right it does!

PR #8326, opened 2018-12-17
Show multiple tasks in active tasks component

### LINKED ISSUE #8214: Show multiple tasks in active tasks component
image

## Acceptance criteria
* [ ] If there are multiple actionable tasks for a single appeal, show them all.

## Non-goal
* Showing/hiding the task instructions when clicking the "View task instructions" link will not be implemented as part of this ticket.
* "+ Add a new task" button will not be implemented as part of this ticket.

rkreyhsig:
Resolves #8214 - Show multiple tasks in active tasks component

| Before | After |
| --- | --- |
|screen shot 2018-12-26 at 11 37 02 am| screen shot 2018-12-26 at 11 29 42 am|


client/app/queue/TaskSnapshot.jsx
```diff
@@ -245,48 +245,52 @@ export class TaskSnapshot extends React.PureComponent<Props> {
...
-    if (this.props.primaryTask) {
-      sectionBody = <table {...css({ width: '100%',
-        marginTop: 0 })}>
+    sectionBody = [];
+    this.props.tasks.map((task, index) => (
+      sectionBody.push(<table {...css({ width: '100%',
+        marginTop: '0px',
+        marginBottom: '5px' })}>
         <tbody>
           <tr>
```
lowellrex: Let's map over each actionable task and return a new table row instead of a new table. We can use a [map within the JSX code here](https://github.com/department-of-veterans-affairs/caseflow/blob/6bea2aab4a10d6bf3de6b4b26ab1f229c9f5637a/client/app/queue/StickyNavContentArea.jsx#L87) as well so that we don't have to bother with transforming `sectionBody` into an array (which I think is causing the "No active tasks" text to not show up properly when it should).

client/app/queue/TaskSnapshot.jsx
```diff
@@ -110,131 +119,122 @@ export class TaskSnapshot extends React.PureComponent<Props> {
...
     }
 
     const {
-      userRole,
-      primaryTask
+      userRole
     } = this.props;
 
-    if (!primaryTask) {
+    if (!task) {
```
lowellrex: Since we're now mapping across a collection of tasks, we will always pass a `task` and can now remove this block.

client/app/queue/TaskSnapshot.jsx
```diff
@@ -51,12 +56,16 @@ const taskInformationContainerStyling = css(taskContainerStyling, { width: '25%'
...
+  padding: '0px',
   position: 'relative',
   verticalAlign: 'top',
   width: '45px'
 });
 
+const titleLabel = css({
+  fontWeight: 'bold'
+});
+
```
lowellrex: Are we adding this style for each `dl` tag? Can we just apply that styling to some parent element so we only have to define it once (perhaps `CaseDetailsDescriptionList` if we don't use that anywhere else or if wherever else we use it is also fine with being bolded).

client/app/queue/TaskSnapshot.jsx
```diff
@@ -305,7 +309,8 @@ const mapStateToProps = (state: State, ownProps: Params) => {
   return {
     appeal: appealWithDetailSelector(state, { appealId: ownProps.appealId }),
     userRole,
-    primaryTask: actionableTasksForAppeal(state, { appealId: ownProps.appealId })[0]
+    primaryTask: actionableTasksForAppeal(state, { appealId: ownProps.appealId })[0],
```
lowellrex: I don't think we need to pluck off the `primaryTask` if we are looping over all actionable tasks.

client/app/queue/TaskSnapshot.jsx
```diff
@@ -34,11 +34,16 @@ export const grayLineStyling = css({
   background: COLORS.GREY_LIGHT,
   margin: 'auto',
   position: 'absolute',
-  top: '35px',
+  top: '20px',
```
lowellrex: Let's make this `25px` to match the change we made to the padding below.

client/app/queue/TaskSnapshot.jsx
```diff
@@ -51,7 +56,7 @@ const taskInformationContainerStyling = css(taskContainerStyling, { width: '25%'
 const taskInfoWithIconContainer = css({
   textAlign: 'center',
   border: 'none',
-  padding: '10px',
+  padding: '0px',
```
lowellrex: Let's only override the top padding to 0 and leave the rest at 10px.

client/app/queue/TaskSnapshot.jsx
```diff
@@ -136,136 +130,147 @@ export class TaskSnapshot extends React.PureComponent<Props> {
...
+          <td {...taskTimeContainerStyling}>
+            <CaseDetailsDescriptionList>
+              <dt>{COPY.TASK_SNAPSHOT_TASK_ASSIGNMENT_DATE_LABEL}</dt>
+              <dd>{task && task.assignedOn &&
+                moment(task.assignedOn).format('MM/DD/YYYY')}</dd>
+              {this.daysSinceTaskAssignmentListItem(task)}
+            </CaseDetailsDescriptionList>
+          </td>
+          <td {...taskInfoWithIconContainer}><GrayDot />
+            <div {...grayLineStyling} className={tskLength - 1 === index ? lastTask : ''} /></td>
```
lowellrex: Let's use a condition here to selectively show the gray line instead of applying a style. Something like the following should work:
```jsx
{ (index + 1 < taskLength) && <div {...grayLineStyling} /> }
```

client/app/queue/TaskSnapshot.jsx
```diff
@@ -136,136 +130,147 @@ export class TaskSnapshot extends React.PureComponent<Props> {
...
+          <td {...taskInformationContainerStyling}>
+            <CaseDetailsDescriptionList>
+              {this.legacyTaskInformation(task)}
+            </CaseDetailsDescriptionList>
+          </td>
+          <td {...taskInformationContainerStyling} {...css({ width: '50%' })}>
```
lowellrex: Let's replicate the pattern we have with `taskTimeContainerStyling` and `taskInformationContainerStyling` and create a specific style for this table cell. Something like the following should work:
```jsx
const taskActionsContainerStyling = css(taskContainerStyling, { width: '50%' });
```

client/app/queue/TaskSnapshot.jsx
```diff
@@ -136,136 +130,147 @@ export class TaskSnapshot extends React.PureComponent<Props> {
...
+        </tr>);
     }
 
     return <div className="usa-grid" {...css({ marginTop: '3rem' })}>
       <h2 {...sectionHeadingStyling}>
         <a id="our-elemnt" {...anchorJumpLinkStyling}>{COPY.TASK_SNAPSHOT_ACTIVE_TASKS_LABEL}</a>
       </h2>
       <div {...sectionSegmentStyling}>
-        { sectionBody }
+        <table {...css({ width: '100%',
```
lowellrex: Let's move this styling out to the top of the file where we have the rest of the styles.

client/app/queue/TaskSnapshot.jsx
```diff
@@ -136,136 +130,147 @@ export class TaskSnapshot extends React.PureComponent<Props> {
...
+          marginTop: '0px',
+          marginBottom: '5px' })}>
```
lowellrex: Why did we add the bottom margin here?

client/app/queue/TaskSnapshot.jsx
```diff
@@ -136,136 +130,147 @@ export class TaskSnapshot extends React.PureComponent<Props> {
...
+    }
+
+    const {
+      userRole
+    } = this.props;
+
+    // users can end up at case details for appeals with no DAS
+    // record (!task.taskId). prevent starting attorney checkout flows
+    return userRole === USER_ROLE_TYPES.judge ? Boolean(task) : Boolean(task.taskId);
+  }
```
lowellrex: I think this section of the file is behind the current state of the master branch. I think this function should read simply:
```jsx
showActionsSection = (task) => (task && !this.props.hideDropdown);
```

client/app/queue/TaskSnapshot.jsx
```diff
@@ -136,136 +130,147 @@ export class TaskSnapshot extends React.PureComponent<Props> {
...
-      { taskIsOnHold(primaryTask) &&
+      { taskIsOnHold(task) &&
         <React.Fragment>
           <dt>{COPY.CASE_LIST_TABLE_TASK_DAYS_ON_HOLD_COLUMN_TITLE}</dt>
-          <dd><OnHoldLabel task={primaryTask} /></dd>
+          <dd><OnHoldLabel task={task} /></dd>
         </React.Fragment>
       }
-      { primaryTask.instructions &&
+      { task.instructions &&
```
lowellrex: We expect `task.instructions` to be an array that may be empty. Let's filter out the cases where the array is empty with something like `task.instructions && task.instructions.length > 0 &&`. And let's do the same for `legacyTaskInformation` as well.

Issue #7916, opened 2018-11-19
Fix 90-day evidence window task for Evidence Submission AMA appeals


laurjpeterson:
There are two AMA dockets that give appellants 90-days to submit evidence:
- Hearing docket - 90 days from the day of the hearing
- Evidence Submission docket - 90 days from the receipt date of the NOD

We'd like to create 90-day evidence window tasks for a few reasons:
1. Tasks populate currently active tasks in Case details and in Case timeline. With this task, users can easily see that a Veteran has a 90-day evidence window and no action can be taken until it completes
2. Veterans can choose to waive their 90-day evidence window. We can resolve this task as a way to track that they've elected to waive it
3. We can rely on the status of this task to know if a case is ready to be automatically distributed to a judge

## Notes
- This github issue only encompasses Evidence Submission docket cases. Hearing docket 90-day evidence window tasks will be created if a hearing has been held, and we have not yet implemented that functionality because AMA Hearings will not be held until ~April 2019.
- Subsequent github issues will also be created to handle the resolving of 90-day evidence window tasks if a Veteran decides to waive it, including which users have permission to act on these tasks.

## AC
- When an Evidence Submission appeal is established and a root task is created, also create a `90-day Evidence Window` task that begins with the `receipt_date` of the NOD and lasts for 90 days
- The `90-day Evidence Window` task blocks automatic case distribution
- The `90-day Evidence Window` task resolves to be closed/completed after 90 days

laurjpeterson:
@amprokop - can i close this? do all newly established Evidence Submission appeals get this task?

amprokop:
I reread the reqs for this — there's one piece missing — we set the timer for the task to be 90 days after the appeal is created in Caseflow, not 90 days after the receipt date. Sadly, I did not anticipate that need, and it will be some work to make it happen. 

What's the priority to make that change on your end?

laurjpeterson:
Assuming established date is later than the receipt date, if we don't make the change then we could delay appeals from being distributed/worked/dispatched. I think it's pretty high priority for this to be accurate to meet AMA requirements. 

When do you think you'd be able to work on it / how long do you think it might take?

lowellrex:
This work was completed by #9340.

PR #8333, opened 2018-12-18
Creates distribution, evidence window, and informal hearing presentation tasks for direct review and evidence submission appeals 

amprokop:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/8334
Related to https://github.com/department-of-veterans-affairs/caseflow/issues/7916 but does not complete it, will be continued in future PR

### Description
If we have ama auto case distribution on, create Distribution, IHP, and evidence submission window tasks as necessary. This PR just creates the tasks, later PRs will introduce the logic to expire/complete them. 

### Testing Plan
1. unit tests pass


app/models/task.rb
```diff
@@ -67,6 +68,18 @@ def children_attorney_tasks
     children.where(type: AttorneyTask.name)
   end
 
+  def complete?
```
lowellrex: I think this function is already available at `completed?`

app/models/task.rb
```diff
@@ -67,6 +68,18 @@ def children_attorney_tasks
...
+    status == Constants.TASK_STATUSES.completed
+  end
+
+  def in_progress?
```
lowellrex: I think this function is already available as `in_progress?`. I think this is because ActiveRecord know that `status` fields are special?

app/models/tasks/judge_task.rb
```diff
@@ -1,5 +1,9 @@
 class JudgeTask < Task
-  include RoundRobinAssigner
+  include Timer
+
+  def when_expired
+
+  end
```
lowellrex: I think this might be detritus from another sketch.

client/constants/TASK_STATUSES.json
```diff
@@ -2,5 +2,6 @@
   "assigned": "assigned",
   "in_progress": "in_progress",
   "on_hold": "on_hold",
-  "completed": "completed"
+  "completed": "completed",
+  "canceled": "canceled"
```
lowellrex: **cancelled

amprokop: 😆

app/models/organizations/auto_case_distributor.rb
```diff
@@ -0,0 +1,5 @@
+class AutoCaseDistributor < Organization
+  def self.singleton
+    AutoCaseDistributor.first || AutoCaseDistributor.create(name: "AutoCaseDistributor", url: "auto_case_distributor")
+  end
```
lowellrex: What is the advantage of creating a new `AutoCaseDistributor` organization instead of using the `Bva` organization we use as a placeholder for other tasks that we don't expect to be assigned to any person or team at the Board?

Two possible benefits would be the ability to see the list of all cases ready for distribution in the organizational queue view, and to limit the group of people who can see that view by using the organizational membership controls. Could we just use the same functions of the `Bva` organization? And rely on the table filters (coming soon!) to filter out `RootTask`s from that view? Are there other benefits to the `AutoCaseDistributor` organization that I'm missing?

app/models/task.rb
```diff
@@ -67,6 +68,10 @@ def children_attorney_tasks
     children.where(type: AttorneyTask.name)
   end
 
+  def all_children_complete
+    children.all?(&:complete?)
+  end
```
lowellrex: I don't think we use this function anywhere.

app/models/tasks/root_task.rb
```diff
@@ -30,20 +30,75 @@ def actions_available?(_user)
...
-    def create_vso_subtask!(appeal, parent)
-      appeal.vsos.each do |vso_organization|
-        InformalHearingPresentationTask.create(
+    def create_vso_subtask!(appeal, parent, status = Constants.TASK_STATUSES.in_progress)
+      appeal.vsos.map do |vso_organization|
+        InformalHearingPresentationTask.create!(
           appeal: appeal,
           parent: parent,
-          status: Constants.TASK_STATUSES.in_progress,
+          status: status,
```
lowellrex: I actually don't think we need to set the status to `in_progress` here since the default value for a task's `status` field is `assigned` and we treat both of those values as essentially "active" values (with the major exception being that we bold docket numbers in the queue table view if the task is `assigned` and not `in_progress`).

app/models/tasks/root_task.rb
```diff
@@ -30,20 +30,75 @@ def actions_available?(_user)
...
           assigned_to: vso_organization
         )
       end
     end
+
+    def create_evidence_submission_task!(appeal, parent)
+      EvidenceSubmissionWindowTask.create!(
+        appeal: appeal,
+        parent: parent,
+        status: Constants.TASK_STATUSES.in_progress,
```
lowellrex: Same comment about default status [as above](https://github.com/department-of-veterans-affairs/caseflow/pull/8333/files#r242980409).

app/models/tasks/root_task.rb
```diff
@@ -30,20 +30,75 @@ def actions_available?(_user)
...
+        assigned_to: MailTeam.singleton
```
lowellrex: What do we expect the MailTeam do with the `EvidenceSubmissionWindowTasks`? I would have thought that these tasks would be for the VSOs (or other representatives in which case this should probably be assigned to `Bva` or some other non-representative entity) to accomplish since they will be the ones submitting the evidence.

app/models/tasks/root_task.rb
```diff
@@ -30,20 +30,75 @@ def actions_available?(_user)
...
+
+        if appeal.needs_ihp?
+          vso_tasks = create_vso_subtask!(appeal, current_parent, ihp_status)
+          # TODO: when or if there are more than one vso, vso tasks
+          # should expire at the same time. which one should be the
+          # blocking parent of evidence submission tasks?
+          current_parent = vso_tasks.first
+        end
+
+        create_evidence_submission_task!(appeal, current_parent) if appeal.evidence_submission_docket?
```
lowellrex: Why do we want to make `EvidenceSubmissionWindowTask`s children of IHP tasks? Can VSOs not work IHPs and evidence submission tasks in parallel?

If IHP tasks and evidence submission tasks _can_ be worked in parallel then we could make both of those types of tasks have the `DistributionTask` as their parent. Doing this would also allow us to have the feature flag only control whether the parent of these tasks is a `RootTask` or `DistributionTask` in the calling function.

```ruby
def create_root_and_sub_tasks!(appeal)
  parent = create!(appeal: appeal)
  if FeatureToggle.enabled?(:ama_auto_case_distribution)
    parent = create_distribution_task!(appeal, parent)
  else
  create_subtasks!(appeal, parent)
end
```

app/models/organizations/auto_case_distributor.rb
amprokop:
```
Two possible benefits would be the ability to see the list of all cases ready for distribution in the organizational queue view
```

Yeah, or to query for all appeals that are between intake or distribution.

I don't think it's a big difference from using `BVA` as the organization 🤷‍♀️. If you feel strongly, I', happy to change it, but it seems like a pretty small detail to me.

app/models/task.rb
amprokop: Ok I can remove it

app/models/organizations/auto_case_distributor.rb
lowellrex: I think we can still do that query based solely on the task type, right?

My preference is to avoid creating additional organizations if existing ones will suffice. I can foresee a proliferation of organizations and would like to get out in front of that if possible.

app/models/tasks/root_task.rb
amprokop: The Mail team would be the ones completing these tasks in the case there is a waiver of the evidence submission period.

As far as I know, VSOs would submit evidence, but that would happen completely outside of Caseflow.

app/models/tasks/root_task.rb
amprokop:
```
Why do we want to make EvidenceSubmissionWindowTasks children of IHP tasks? Can VSOs not work IHPs and evidence submission tasks in parallel?
```

Correct, these periods cannot happen in parallel. There is a window to submit evidence, and immediately following that there is a period where the IHP is written.

app/models/tasks/root_task.rb
amprokop: Good thought, thanks for clarifying and for the context. How confident are you that if I change the default IHP task status from in_progress to assigned as you suggest, it will not break anything? If "very confident", I'm happy to make the change. 

Otherwise, my default behavior would be to not change anything about the way it currently works.

app/models/tasks/root_task.rb
amprokop: Wondering if you think there's some other reason we should assign these tasks to VSOs. Can always change it later, there is more to work out here for sure, just looking for a tentative answer really.

app/models/tasks/root_task.rb
```diff
@@ -30,20 +30,75 @@ def actions_available?(_user)
...
+    def needs_subtask(appeal)
+      appeal.needs_ihp? || appeal.hearing_docket? || appeal.evidence_submission_docket?
+    end
+
+    def create_subtasks!(appeal, parent)
+      distribution_status = if needs_subtask(appeal)
+                              Constants.TASK_STATUSES.on_hold
+                            else
+                              Constants.TASK_STATUSES.in_progress
+                            end
```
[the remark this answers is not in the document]
amprokop: Cool, thanks for the suggestion! I'll look into this

app/models/organizations/auto_case_distributor.rb
amprokop: Sure, I can change it back to BVA then.

app/models/tasks/judge_task.rb
```diff
@@ -91,6 +91,7 @@ def self.unassigned_ramp_tasks
   end
 
   def self.eligible_for_assigment?(task)
+    return false if task.appeal.nil?
```
amprokop: unrelated change, just a small condition to get the judge task assignment script to succeed when running it manually

app/models/tasks/root_task.rb
lowellrex: The reason why I thought this task should be assigned to VSOs instead of the Mail Team is that by assigning these tasks to VSOs we enable the VSOs to be the ones to notify the application that the evidence has been submitted or that they are waiving their evidence submission window.

If we assign this task to the mail team we are relying on them to trigger the closing of this task which will have to happen programmatically (maybe when an `EvidenceOrArgumentMailTask` is created for this appeal?) or will rely on a member of the mail team to search for this appeal and close the task when additional evidence is received (and if the VSO submits evidence in more than one batch then we may close this task prematurely). I think both of those courses of action are fraught, and I think the simplest solution to both of those potential problems would be allowing VSOs control over when this task is marked complete.

app/models/tasks/root_task.rb
lowellrex: How about if we create IHP tasks off of a common parent task (the `DistributionTask`) when the evidence submission task is marked as complete. I think this reduces the complexity of this function dramatically and maintains the correct model of tasks.

```ruby
def create_subtasks!(appeal, parent)
  transaction do
    distribution_task = create_distribution_task!(appeal, parent)
    appeal.evidence_submission_docket? ? create_evidence_submission_task!(appeal, distribution_task) : create_vso_subtask!(appeal, distribution_task)
  end
end
```

app/models/tasks/root_task.rb
amprokop: Not all evidence submission appeals have VSOs. In your mind, where would we assign tasks for evidence submission appeals without VSOs? (And, I’m forgetting- can Appeals have multiple VSOs too?)

amprokop: Also, just for clarification- submitting evidence would not, I believe, end the evidence submission window - what would end it is submitting a waiver explicitly saying that you want to waive the period.

So there shouldn’t be a problem with closing this task prematurely, do you agree?

app/models/tasks/root_task.rb
amprokop: Sure, that sounds fine.

There is one added benefit here!!! 

If we decouple the creation of these tasks, we can check for POA changes in a retrying job after the evidence window ends and before the IHP window starts.

app/models/tasks/root_task.rb
amprokop: In the end, I think this touches on product and process, so I’m fine going with whatever for now, but we’ll have to circle back with Lauren and BVA regardless

lowellrex: Interesting, I did not know about the explicit waiver. I think it makes sense to assign the task to the mail team in that case.

> can Appeals have multiple VSOs too?

Yes. Each issue being appealed can have different a representative. [We don't currently handle these situations correctly](https://github.com/department-of-veterans-affairs/caseflow/blob/3a3adfa70b540c460fc8005e003593eda8375a23/app/models/appeal.rb#L218) in the application though.

lowellrex: This approach (one evidence submission window task per appeal assigned to the mail team) works for now, but I wonder if we will need to adapt it to be on the request issue-level instead of the appeal-level...

app/models/tasks/distribution_task.rb
```diff
@@ -0,0 +1,10 @@
+class DistributionTask < GenericTask
+  # Prevent this task from being marked complete
+  # when a child task (e.g. evidence submission)
+  # is marked complete.
+  def update_status_if_children_tasks_are_complete
```
amprokop: @lowellrex — i had to add this here because distribution tasks were being marked as complete when evidence submission tasks were marked as complete, due to the `update_status_if_children_tasks_are_complete` method in `task.rb`. 

I'm not confident of all of the implications of making a change to `task.rb`, so I overwrote the method here. Not sure if you'd prefer to handle this in a different way.

app/models/task.rb
```diff
@@ -141,6 +141,8 @@ def when_child_task_completed
     update_status_if_children_tasks_are_complete
   end
 
+  def when_child_task_created; end
```
amprokop: I had to add this here because some tests were failing — apparently some GenericTasks have Tasks as parents.

app/models/task.rb
```diff
@@ -255,7 +257,7 @@ def automatically_assign_org_task?
     assigned_to.is_a?(Organization) && assigned_to.automatically_assign_to_member?(self.class)
   end
 
-  def update_parent_status
+  def on_complete
```
lowellrex: Maybe we can rename this `notify_parent_task_child_task_completed` or something to give us some indication of what the function is doing instead of when we expect it to run?

app/models/tasks/evidence_submission_window_task.rb
```diff
@@ -0,0 +1,6 @@
+class EvidenceSubmissionWindowTask < GenericTask
+  def on_complete
+    RootTask.create_vso_subtask!(appeal, parent)
+    super
+  end
```
lowellrex: Let's attach this to a different `after_update` hook that would just be specific to the `EvidenceSubmissionWindowTask`. Maybe something like:
```ruby
after_update :create_vso_subtask, if: :status_changed_to_completed_and_has_parent?

def create_vso_subtask
  RootTask.create_vso_subtask!(appeal, parent)
end
```

app/models/tasks/generic_task.rb
```diff
@@ -16,6 +17,10 @@ def verify_org_task_unique
     end
   end
 
+  def when_child_task_created
+    update!(status: Constants.TASK_STATUSES.on_hold) unless on_hold?
+  end
```
lowellrex: I think [we already have a function](https://github.com/department-of-veterans-affairs/caseflow/pull/8333/files#diff-c95c9451cf097bdd58f73d0f9b114f97R12) that sets the parent task to on hold when we create a child task.

app/models/tasks/distribution_task.rb
lowellrex: I think this change makes sense here. Maybe we could add a little more color in the code comment about why we want to keep this organizational task open. Maybe something like: Do not automatically complete DistributionTasks when children are completed because we DistributionTasks are used to signal that cases are ready to be distributed to judges.

app/models/task.rb
lowellrex: I think [we may not need this function](https://github.com/department-of-veterans-affairs/caseflow/pull/8333/files#r246953874) at all.

app/models/tasks/evidence_submission_window_task.rb
amprokop: sure, sounds fine if you prefer

app/models/tasks/distribution_task.rb
amprokop: sure

Issue #7990, opened 2018-11-26
Non-comp Flow | Reviews landing page


shanear:
Compensation and Pension use VBMS to process HLRs and SCs. Other business lines do not, thus we should not create EPs for those. Instead we will use the DB objects directly in the Caseflow non-comp UI. This is the page where those HLRs and SCs will appear.

## Acceptance Criteria
- When a SC or HLR is created and has a benefit type other than Compensation or Pension, it should appear as a task on the business line review page associated with that benefit type.
- Visiting the `/decision_reviews/[business-line-slug]` url should show the review page associated with the key for that business line (i.e. `/decision_reviews/vocational_rehab_and_employment` should show the Vocational Rehab and Employment reviews page)
- "Days Waiting" column should be the number of days since the task was added to the queue (which would be the date the decision review intake was completed)
- The "Intake New Form" link should route the user to the intake flow `/intake`

## Notes
Out of scope for this story:
- The "Completed" tab
- The search bar
- The link for clicking on an individual task
- "Board Grant" tasks
- order of the tasks
- pagination

Full non-comp flow: https://va-gov.invisionapp.com/share/PFPCPJ7MEGC

## Mocks

![returning disposition main](https://user-images.githubusercontent.com/11700545/49621592-01b91300-f97b-11e8-98db-00f8cf0702cd.png)

## Technical Notes
While issues typically do not contain details about technical implementation, there's enough open questions that could have long term implications that we've discussed a rough approach:

1. We will build this noncomp flow into a new Redux app, separate from Intake/IntakeEdit/Queue
2. We will reuse whatever Queue frontend code we can, refactoring to make generic if necessary. Look to reuse reducers, UI components, logic. See TaskTable as an example
3. We will reuse Queue backend code - Organizations can map to business lines, Task can be used directly with some modifications to support a polymorphic appeal/HLR/SC association (currently only appeal)

There will be potentially be many different kinds of tasks on this page.

See additional notes here: https://github.com/department-of-veterans-affairs/appeals-design-research/issues/1081

There's a [sample PR](https://github.com/department-of-veterans-affairs/caseflow/pull/8111/files) that Mark put together that shows how this might be built into Queue - a good reference to use for what components can be shared, though we want to build this into a new Redux app.

References:
* [models/tasks/generic_task.rb](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/tasks/generic_task.rb)
[...]

PR #8345, opened 2018-12-18
Non comp reviews landing page

pkarman:
connects #7990 

### Description
This is the first of multiple PRs. This one sets up the basic backend functionality for assigning Tasks to DecisionReviews.

### DIFF
--- a/app/controllers/decision_reviews_controller.rb
+++ b/app/controllers/decision_reviews_controller.rb
@@ -16,7 +16,7 @@ class DecisionReviewsController < ApplicationController
   end
 
   def business_line
-    @business_line ||= NonComp.find_by(url: business_line_slug)
+    @business_line ||= BusinessLine.find_by(url: business_line_slug)
   end
 
   def set_application
--- a/app/controllers/decision_reviews_controller.rb
+++ b/app/controllers/decision_reviews_controller.rb
@@ -1,0 +1,42 @@
+class DecisionReviewsController < ApplicationController
+  before_action :verify_access, :react_routed, :verify_feature_enabled, :set_application
+
+  def index
+    if business_line
+      render json: { tasks: business_line.tasks }
+    else
+      render json: { error: "#{business_line_slug} not found" }, status: 404
+    end
+  end
+
+  private
+
+  def business_line_slug
+    params.permit(:business_line_slug)[:business_line_slug]
+  end
+
+  def business_line
+    @business_line ||= NonComp.find_by(url: business_line_slug)
+  end
+
+  def set_application
+    RequestStore.store[:application] = "decision_reviews"
+  end
+
+  # TODO: authz rules for this space
+  def verify_access
+    return false unless business_line
+    return true if current_user.admin?
+    return true if business_line.user_has_access?(current_user)
+
+    Rails.logger.info("User with roles #{current_user.roles.join(', ')} "\
+      "couldn't access #{request.original_url}")
+
+    session["return_to"] = request.original_url
+    redirect_to "/unauthorized"
+  end
+
+  def verify_feature_enabled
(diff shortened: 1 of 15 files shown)


app/models/organizations/non_comp.rb
```diff
@@ -0,0 +1,2 @@
+class NonComp < Organization
```
shanear: I think calling this `BuisnessLine` is probably safer

app/controllers/decision_reviews_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+      render json: { tasks: business_line.tasks }
+    else
+      render json: { error: "#{business_line_slug} not found" }, status: 404
+    end
+  end
+
+  private
+
+  def business_line_slug
+    params.permit(:business_line_slug)[:business_line_slug]
```
shanear: I don't think permit is necessary here

app/models/serializers/work_queue/decision_review_serializer.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  attribute :aod do
+    object.try(:advanced_on_docket)
+  end
+
+  attribute :issue_count do
+    object.number_of_issues
+  end
+end
```
shanear: Call me old fashioned, but this feels far more cumbersome and verbose than our old serialization technique

app/models/serializers/work_queue/task_serializer.rb
```diff
@@ -32,15 +32,15 @@ class WorkQueue::TaskSerializer < ActiveModel::Serializer
   end
 
   attribute :docket_name do
-    object.appeal.docket_name
+    object.appeal.try(:docket_name)
```
shanear: We've been using `&.` for this

app/models/decision_review.rb
```diff
@@ -174,6 +187,11 @@ def contestable_issues
     contestable_issues_from_ratings + contestable_issues_from_decision_issues
   end
 
+  # do not confuse ui_hash with serializer. ui_hash for intake and intakeEdit. serializer for work queue.
```
jiahuang: ooph, good callout

spec/controllers/decision_reviews_controller_spec.rb
```diff
@@ -0,0 +1,48 @@
+describe DecisionReviewsController, type: :controller do
+  before do
+    FeatureToggle.enable!(:decision_reviews)
```
jiahuang: do we need a feature toggle for this? might not since there's no users that have the url & there's no way to get to it from our current flows.

app/controllers/decision_reviews_controller.rb
pkarman: using it makes rails happier and less noisy about unpermitted params in the logs.

app/models/serializers/work_queue/decision_review_serializer.rb
pkarman: ok, you're old fashioned! :)

truly, it is more verbose. otoh, it is the Approved ActiveRecord way of doing it.

app/models/serializers/work_queue/task_serializer.rb
pkarman: I believe Ruby will die on `&.` for an unavailable method. `try` will just return `nil` which in this case is what we want.

see https://stackoverflow.com/questions/37977721/why-is-safe-navigation-better-than-using-try-in-rails

spec/controllers/decision_reviews_controller_spec.rb
pkarman: It might be overkill, but security through obscurity feels less explicit and riskier. The code is open, after all, so if a user wanted to, they could see that the route exists.

PR #8349, opened 2018-12-18
Convert decass notes to ASCII before saving to DB

### LINKED ISSUE #8128: Limit attorney's notes during checkout on FE to 350 chars
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2709/

Screenshot:

image

For legacy cases, we update Vacols DECASS table. One of the fields on that table has a length limit. 

AC:
1. On BE, when setting `decass.deatcom` limit to be no longer than 350 chars
2. On FE, limit attorney's notes during checkout to be less than 350 chars

monfresh:
Resolves #8128

**Why**:
The VACOLS.DECASS.DEATCOM DB column is limited to 350 characters. If an attorney
submits a note with 350 characters, where some are multi-byte characters, such 
as curly quotes or accented characters, the VACOLS Oracle DB will treat the 
string as having more than 350 characters, which results in an exception due to
the DB constraint.

**How**:
- On the backend, only use the first 350 characters, then convert them all to
ASCII using the `stringex` gem which has a handy `to_ascii` method.

On the frontend, the Notes field was already limited to 350 characters.

The truncation and ASCII conversion happens in the `QueueMapper` class, which 
was refactored using the following principles:

- Use a class instead of a module to avoid passing arguments around
- Give the class a single public method and only test that method
- Move the data transformation to individual short methods with clear names
- Keep all methods under 10 lines of code
- Order methods by level of abstraction for better readability

I noticed that other VACOLS columns have length limits, so we might need to
convert those to ASCII as well. Or perhaps it might make sense to have a generic
ASCII transformation that gets applied to all VACOLS string columns.


aroltsch:
This is a very nice refactor @monfresh! It is looking much cleaner - I like it a lot! I will take one more look tomorrow AM.

monfresh:
@pkarman I added the stringex gem and used the `to_ascii` method. Is that what you had in mind?

spec/repositories/queue_repository_spec.rb
```diff
@@ -147,6 +147,18 @@ def vacols_id
...
+
+    context "when note with multi-byte characters causes it to be greater than 350 characters" do
+      let(:decass_attrs) { { note: ("a" * 341) + "Veteran’s" } }
+      let(:date_added) { "2018-04-18".to_date }
+      let!(:decass) { create(:decass, defolder: vacols_case.bfkey, deadtim: date_added) }
+
+      it "converts multi-byte characters to ASCII (VACOLS Oracle DB is in ASCII format)" do
+        subject
+
+        expect(decass.reload.deatcom).to eq(("a" * 341) + "Veteran's")
```
pkarman: nice!

Issue #7981, opened 2018-11-26
Schedule AMA Veterans | AMA Veterans Waiting


MeredithStewart:
### User Stories
As a hearing coordinator, I need to schedule AMA Veterans as required into any hearing day on the schedule. 

### AC 

- [ ] Verify that Veterans in the AMA Hearing Docket are prioritized in order by NOD date
- [ ] Verify that Veterans in the Hearing Docket date are prioritized with CAVC first, then AOD. 
- [ ] Verify that CAVC hearings and AOD hearings are also prioritized in order of NOD date 
- [ ] Verify that Veterans in the AMA Hearing Docket can be scheduled for any available hearing day.

### Mocks
- https://va-gov.invisionapp.com/share/AYPASC0CDWG
Tabs
<img width="987" alt="screen shot 2018-11-29 at 2 13 11 pm" src="https://user-images.githubusercontent.com/44249412/49245647-f7c66b80-f3e0-11e8-915c-520c42e7ef40.png">

PR #8353, opened 2018-12-18
Code cleanup to prep for AMA hearing schedule

sharonwarner:
Connects #7981 

# Description
This PR refactors the Assign Hearings page to use the queue appeal serializer. This will set us up to easily add AMA appeals as well.

# To test
1) Go to assign hearings page
1) Confirm you can still assign veterans


mdbenjam:
So the problem with using the Queue serializer is that it pulls in a lot of fields hearings don't need. I think we'll need to create a separate serializer for hearings. The other option is to treat it more like how Queue handles tasks. We might be able to reuse the task serializer which does pass limited appeal information.

PR #8354, opened 2018-12-19
[WIP] NonComp frontend

jiahuang:
Connects #7990

Also see https://github.com/department-of-veterans-affairs/caseflow/pull/8111#issuecomment-448422089 for plans around TaskTable

### DIFF
--- a/client/app/nonComp/components/NonCompTabs.jsx
+++ b/client/app/nonComp/components/NonCompTabs.jsx
@@ -1,0 +1,30 @@
+import TabWindow from '../../components/TabWindow';
+import TaskTable from '../../queue/components/TaskTable';
+
+class NonCompTabs extends React.PureComponent {
+  render = () => {
+    const tabs = [{
+      label: 'In progress tasks',
+      page: <TaskTableTab
+        description="In progress"
+        tasks={this.props.unassignedTasks}/>
+    },{
+      label: 'Completed tasks',
+      page: <TaskTableTab
+        description="Completed"
+        tasks={this.props.completedTasks}/>
+    }]
+  }
+}
+
+const TaskTableTab = ({ description, tasks }) => <React.Fragment>
+   <p className="cf-margin-top-0">{description}</p>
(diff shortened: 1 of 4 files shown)


client/app/nonComp/components/NonCompTabs.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+    return <TabWindow
+       name="tasks-organization-queue"
+       tabs={tabs}
+     />;
+  }
+}
+
+const TaskTableTab = ({ description, tasks }) => <React.Fragment>
+   <p className="cf-margin-top-0">{description}</p>
```
jiahuang: task table will go in here, but right now it'll break since we don't have the props required. I can either fix up task table in this PR or make another PR to do that. I'm leaning towards putting it in another PR.

app/views/decision_reviews/index.html.erb
```diff
@@ -0,0 +1,13 @@
+<% content_for :full_page_content do %>
+  <%= react_component("NonComp", props: {
+    userDisplayName: current_user.display_name,
+    dropdownUrls: dropdown_urls,
+    feedbackUrl: feedback_url,
+    buildDate: build_date,
+    serverNonComp: {
+      businessLine: business_line.name,
```
pkarman: good placeholder for now

spec/controllers/decision_reviews_controller_spec.rb
```diff
@@ -29,15 +29,15 @@
...
         get :index, params: { business_line_slug: non_comp_org.url }
 
         expect(response.status).to eq 200
-        expect(response.body).to eq({ tasks: [] }.to_json)
       end
     end
 
-    context "business-line-slug is not found" do
+    # todo, swap out this test with a feature test
+    context "business-line-slug is not found", skip: true do
```
pkarman: can't we do both? I think we want the controller behavior and this is the place to test it.

PR #8357, opened 2018-12-19
Remove therubyracer gem

monfresh:
**Why**: It's only needed if a JS runtime is not available on the server, and
assuming we have Node installed everywhere, we don't need it.


monfresh:
Closing this for now while I do more research. It looks like at one point, react-rails recommended `therubyracer` on MRI for best performance (https://github.com/shakacode/react_on_rails/issues/21), but the document linked in that GitHub issue now recommends `mini_racer` based on benchmarks from Discourse. However, I think those benchmarks are old.

PR #8374, opened 2018-12-19
Example TaskTimer using Asyncable

pkarman:

### Description

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...

### DIFF
--- a/app/models/concerns/asyncable.rb
+++ b/app/models/concerns/asyncable.rb
@@ -74,6 +74,13 @@ module Asyncable
     end
   end
 
+  def process_after(at_when)
+    transaction do
+      submit_for_processing!
+      update!(self.class.attempted_at_column => (at_when - REQUIRES_PROCESSING_RETRY_WINDOW_HOURS.hours))
+    end
+  end
+
   def submit_for_processing!
     update!(self.class.submitted_at_column => Time.zone.now, self.class.processed_at_column => nil)
   end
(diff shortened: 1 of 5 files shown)


amprokop:
Thanks for whipping this up, @pkarman!

Based on my reading of the code, it looks like we're essentially leveraging `attempted_at`'s logic to sort of trick Asyncable into thinking the first run of a job is actually the first retry. Is that right?

Wondering if you see this as sort of a stopgap solution that we can expand. I do think it would narrowly work for the use case I described, but we'd sort of lose the retry capability if we did want to use it for timed tasks that might require retries (like the POA scenario I described), since the timer we'd set would be for large values like 45 or 90 days. Am I reading into this correctly? 

Another item here that could be addressed in the timer (this is one that I'd probably address later, more just want your thoughts). I'm speculating a bit here, but I think we'd potentially want to use business days for our timers, rather than a strict hourly interval, to avoid tasks being yanked from a user's queue in the middle of the day. So for example, if an evidence submission appeal with no VSO was intook at 3pm EST, we might want it not to expire at 3pm EST + 90 days, but instead 3 pm EST + 90 days + the time until midnight on that day. Does that make sense to you?

cc @laurjpeterson on this last point, since it is a slight policy issue here, this sort of thing usually only comes up in my experience when it's time to write the code :)

pkarman:
> Based on my reading of the code, it looks like we're essentially leveraging attempted_at's logic to sort of trick Asyncable into thinking the first run of a job is actually the first retry. Is that right?

correct. @shanear is working on an alternate approach too.

we could instead change the `submitted_at` logic to only include those rows that are not null *and* dated in the past.

PR #8380, opened 2018-12-19
[WIP] Improve document count and new documents performance

### LINKED ISSUE #8192: [BE] Improve document count and new documents performance
We have been seeing `Response timeout of 60000ms exceeded` errors for new_documents and document_count endpoints. 

Note: We have a nightly Reader background job that downloads VBMS documents, saves the list in the Caseflow documents table and uploads these documents to s3. This job runs for users with a Reader role only.

Improvements to the above endpoints:
1. Include AMA appeals to the Reader background job.
1. Ensure all queue users are part of the Reader background job. How can we get a list of Queue users? Can we query for users who have logged in to Caseflow for the past 2 weeks and assume these users will need documents? Can we use attorney/judge mapping table or organizations table?
1. For the `document_count` and `new_documents` endpoints, do not make calls to VBMS and instead use Caseflow `documents` table to get results. 
1. Remove any VACOLS calls from the endpoints (?).

Attention! The above approach to use a Reader job might result in outdated content - 24 hours max.

hschallhorn:
References #8192

### Description
Change calls to VBMS to calls to the caseflow documents table to improve the performance of `new_documents` and `document_count`.

### Acceptance Criteria
- [x] AMA appeals are included in nightly reader job.
- [ ] No calls made to VBMS
- [ ] No calls made to VACOLS (?)
- [ ] No `Response timeout of 60000ms exceeded` errors


app/jobs/fetch_documents_for_reader_user_job.rb
```diff
@@ -15,8 +15,13 @@ def perform(reader_user)
 
     setup_debug_context(reader_user)
     update_fetched_at(reader_user)
-    appeals = reader_user.user.current_case_assignments
-    fetch_documents_for_appeals(appeals)
+    legacy_appeals = reader_user.user.current_case_assignments
+
+    ama_user_tasks = Task.where(assigned_to_id: reader_user.user.id)
```
aroltsch: You could use objects instead of IDs
```
assigned_to: reader_user.user
```

aroltsch: Rails magic :)

app/jobs/fetch_documents_for_reader_user_job.rb
```diff
@@ -15,8 +15,13 @@ def perform(reader_user)
...
+      .where.not(status: Constants.TASK_STATUSES.completed).all
+    ama_appeals = ama_user_tasks.map { |user_task| Appeal.find(user_task.appeal_id) }.uniq
```
aroltsch: You could use `map` to get appeal objects from tasks
```
ama_user_tasks.map(&:appeal).uniq
```

hschallhorn: Duh. Thanks!

Issue #8262, opened 2018-12-12
Appeal Outcoding | Creating outcoding job that uploads decision document


shanear:
There will be a handful of actions that need to be performed in external systems as a result of appeals outcoding. The first one is uploading a document, which already happens, but should be performed asynchronously.

## Acceptance Criteria
- When an appeal is outcoded, a job should be kicked off that uploads the decision document to VBMS.

## Notes
This will be the same job that also creates the Board Grant end product and contentions.

## Techincal Notes
- The decision document code should already be there, we just need to make it asynchronous.
- We should use `Asyncable` to do this. And add this to the list of retry actions.
- Outcoding appeals currently happens in BvaDispatchTask
- setting the attempted_at date might be one way to control when a job executes

steve-gov:
Update: Compensation nonrating issue pending dispatch on https://appeals.cf.uat.ds.va.gov/queue/appeals/098cf858-9e94-4337-aa40-e3c5e0d3d457

steve-gov:
From VBMS Document Properties:
```
Shipping Number: 
New Mail: false
Document ID: {CDF47CA6-DFC5-4B1E-9EA7-D11EAAA7F9B9}
Date of Upload: Wed Jan 23 12:08:48 EST 2019
Content Source: BVA
Source Comment: 
System Source: BVA
Moved by: 
```

steve-gov:
Leaving this open until the next appeal is outcoded and doc uploaded.

shanear:
looks like the queue team decided it should not be delayed. Took that off of the acceptance criteria:
https://github.com/department-of-veterans-affairs/caseflow/pull/8440

steve-gov:
Validated through several other stories. 

<img width="863" alt="screen shot 2019-01-28 at 12 18 26 pm" src="https://user-images.githubusercontent.com/25468689/51863801-d73e6300-22f6-11e9-9d90-ae5b8c6ed4a4.png">

PR #8381, opened 2018-12-19
Process decision document upload in separate job

shanear:
connects #8262


app/jobs/sync_reviews_job.rb
```diff
@@ -13,9 +13,10 @@ def perform(args = {})
     limit = args["limit"] || DEFAULT_EP_LIMIT
 
     perform_end_product_syncs(limit)
-    perform_ramp_refiling_reprocessing
     perform_decision_review_processing(limit)
     perform_decision_rating_issues_syncs(limit)
+    reprocess_decision_documents(limit)
```
shanear: Going to continue to put the reprocessing stuff in this job. We may want to take a look at this later, but i don't see the downside right now, since it's far simpler

app/jobs/sync_reviews_job.rb
```diff
@@ -13,9 +13,10 @@ def perform(args = {})
...
+    perform_ramp_refiling_reprocessing
```
shanear: moved this one to the back since its the only one that's actually doing procesing, instead of just starting more jobs

app/models/decision_document.rb
```diff
@@ -1,35 +1,44 @@
...
-  def source
-    "BVA"
+  def submit_for_processing!
+    cache_file!
+    super(delay: DECISION_OUTCODING_DELAY)
   end
 
-  def upload!
-    return unless file
-    S3Service.store_file(s3_location, Base64.decode64(file))
```
shanear: This will happen synchronously in `submit_for_processing!`, since we will need to store the file content somewhere while the job waits. S3 seemed like the best/safest place to do this.

spec/models/tasks/bva_dispatch_task_spec.rb
```diff
@@ -55,21 +59,24 @@
...
       it "should complete the BvaDispatchTask assigned to the User and the task assigned to the BvaDispatch org" do
-        expect(Caseflow::Fakes::S3Service).to receive(:store_file)
-          .with("decisions/" + root_task.appeal.external_id + ".pdf", /PDF/)
-        allow(VBMSService).to receive(:upload_document_to_vbms)
-        BvaDispatchTask.outcode(root_task.appeal, params, user)
+        delayed_decision_time = Time.zone.now + DecisionDocument::DECISION_OUTCODING_DELAY
+
+        expect do
+          BvaDispatchTask.outcode(root_task.appeal, params, user)
+        end.to have_enqueued_job(ProcessDecisionDocumentJob).at(delayed_decision_time)
```
shanear: decided to use the active job expectations since they validate we are actually using ActiveJob correctly.

app/models/decision_document.rb
```diff
@@ -38,7 +47,16 @@ def pdf_name
...
+  def s3_location
+    DecisionDocument::S3_SUB_BUCKET + "/" + pdf_name
+  end
+
   def output_location
     File.join(Rails.root, "tmp", "pdfs", pdf_name)
   end
+
+  def cache_file!
+    fail NoFileError unless @file
```
aroltsch: We don't want to throw this error just yet. We have to coordinate with Paul on this https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/idt/api/v1/appeals_controller.rb#L78

spec/models/decision_document_spec.rb
```diff
@@ -17,6 +17,7 @@
 
   context "#submit_for_processing!" do
     subject { decision_document.submit_for_processing! }
+    before { FeatureToggle.enable!(:decision_document_upload) }
```
aroltsch: It is always a good idea to disable the flag in the `after` block to ensure other tests don't get messed up.

app/jobs/process_decision_document_job.rb
```diff
@@ -0,0 +1,11 @@
+class ProcessDecisionDocumentJob < CaseflowJob
+  queue_as :low_priority
+  application_attr :intake
+
+  def perform(decision_document)
+    RequestStore.store[:application] = "intake"
```
aroltsch: Perhaps `idt`, not `intake`?

PR #8383, opened 2018-12-19
make task table work with custom fields

jiahuang:
Connects #7990 

image

with a data blob of:

```
[
      {
        claimant: "bob",
        appeal: {
          id: 1,
          isLegacyAppeal: false,
          issueCount: 9
        },
        url: "/intake",
        veteranSSN: "12345",
        assignedOn: Time.zone.now - 2.days,
        type: "higher level review"
      },
      {
        claimant: "bobbet",
        appeal: {
          id: 2,
          isLegacyAppeal: false,
          issueCount: 5
        },
        url: "/intake",
        veteranSSN: "54321",
        assignedOn: Time.zone.now - 3.days,
        type: "supplemental claim"
      }
    ]
```

Modifies task table to do the following:
* all columns can have an "order" field which determines order it shows up in the table
* task table can take in a `customColumns` prop which has additional columns


client/app/queue/components/TaskTable.jsx
```diff
@@ -55,7 +55,7 @@ type Props = Params & {|
   userRole: string
 |};
 
-class TaskTable extends React.PureComponent<Props> {
+export class TaskTableUnconnected extends React.PureComponent<Props> {
```
pkarman: does changing the class name make the file name misleading?

jiahuang: the connected task table is still the one exported by default

client/app/nonComp/components/NonCompTabs.jsx
```diff
@@ -23,8 +26,44 @@ class NonCompTabsUnconnected extends React.PureComponent {
...
+};
+
+const appealTypeColumn = () => {
+  return {
+    header: 'Type',
+    valueFunction: (task) => task.type,
+    getSortValue: (task) => task.type,
+    // order determines where this column displays
+    // make it 100 so this column is always last
+    order: 100
```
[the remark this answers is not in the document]
jiahuang: can't make it 0 because js is silly. Made it -1 since thats at least a little better than 100..

```js
> 0 == null
false
> 0 < null
false
> 0 > null
false
```

client/app/nonComp/components/NonCompTabs.jsx
```diff
@@ -23,8 +26,44 @@ class NonCompTabsUnconnected extends React.PureComponent {
   }
 }
 
-const TaskTableTab = ({ description }) => <React.Fragment>
+const claimantColumn = () => {
```
mdbenjam: One thought. I think we should consider having one file where we put all of these in. I worry about us reimplementing the same columns in different places.

jiahuang: Good point. I'm going to move this into a utils folder inside of noncomp for now since these also rely on the data being formatted in a way specific to non-comp. When queue does a refactor we can look into pulling out this + the task table component into the components folder.

PR #8389, opened 2018-12-20
Serialize decision review tasks with ActiveSerializer

pkarman:
connects #7990 

### Description
Preps backend payload to match client expectations.


app/models/serializers/work_queue/decision_review_task_serializer.rb
```diff
@@ -0,0 +1,38 @@
+class WorkQueue::DecisionReviewTaskSerializer < ActiveModel::Serializer
```
shanear: The time for this may have already passed but `DecisionReviewTask` is super generic. Does it represent any task for a decision review? Or just the one that goes on the business line pages?

app/models/serializers/work_queue/decision_review_task_serializer.rb
```diff
@@ -0,0 +1,38 @@
...
+  def task
+    object
+  end
+
+  def decision_review
+    task.appeal
+  end
+
+  attribute :claimant do
```
jiahuang: Does `decision_review.claimants` return the veteran if the payee code is "00"? Or is this nil if the claimant is the veteran?

app/models/serializers/work_queue/decision_review_task_serializer.rb
pkarman: Right now it is any task. I just the most generic name on purpose for that reason. We could subclass (or rename) it for `BusinessLineTask` e.g. if that makes more sense.

app/models/serializers/work_queue/decision_review_task_serializer.rb
pkarman: I think it depends on what we want to show the user.

I assume it should return the veteran as you describe. I can change.

Issue #8107, opened 2018-12-05
Create subclass for JudgeTasks created by Quality Review


lowellrex:
`JudgeAssignTask`s can be created when cases are distributed to judges or when the Quality Review team returns a task to the signing judge.
* #8106 (PR #8108) added logic to handle the different actions that should be available in these two cases.
* PR #8407 made sure the correct actions are shown when a judge assigns a QR task to an attorney. 
* #8033 (PR #8072) created subclasses for `JudgeAssignTask` and `JudgeReviewTask`, so we can follow their lead and create another subclass to encapsulate the quality review-related logic we added in #8106.

## Acceptance criteria
* [x] Create `JudgeQualityReviewTask` that encapsulates quality review-related logic
* [x] `JudgeQualityReviewTask`s show up in judge review table
* [x] Rename `JudgeReviewTask` to `JudgeDecisionReviewTask`
* [x] Migrate existing `JudgeReviewTask`s to `JudgeDecisionReviewTask`s

lowellrex:
If we are able to come up with a satisfactory workaround for changing a task's status manually (maybe through the database directly?) without triggering a cascade of changes to parent tasks, we should consider moving [the `mark_as_complete!` method](https://github.com/department-of-veterans-affairs/caseflow/blob/d5e4c7463b45416dc1dea57e8c2b65a99de153df/app/models/task.rb#L137) back into an `after_update` hook.

PR #8391, opened 2018-12-20
Display link to edit request issues

### LINKED ISSUE #7920: Add link to correct issues (add/remove issues) from Case details
Part of [Editing AMA issues](https://github.com/department-of-veterans-affairs/caseflow/issues/5961)

When attorneys are viewing issues in Case details and they notice that the issues do not accurately reflect what the Veteran requested on their NOD, they need to add/remove issues to correct them.

The intake team has already deployed the ability to do this correcting, we just need to add a link in Case details for attorneys to do so.

Right now, only attorneys will do this editing, but it's possible other BVA users should have this permission in the future, like VLJ support, Lit Support, etc. I think we can put this behind a feature flag for now, but am open to other ideas - @mdbenjam what do you think?

## AC
- Link exists to correct issues, reading `Correct issues` per mocks
- When a user clicks the link, they navigate to the `/edit` URL for the case to add/remove issues (note: this link does not have `/queue` in it)
- When the user finishes adding/editing issues, they navigate back to the Case details page (note: this already works)

Tech notes:
1. an attorney can edit request issues before they checked out the case to the judge
2. a judge can edit a request issue before he/she finishes his/her checkout.

## Mocks
@allyceh to add

aroltsch:
Resolves #7920 

### Description
Display a link to edit request issues for attorneys and judges only if they have not completed the checkout

image


app/models/user.rb
```diff
@@ -78,6 +78,15 @@ def vacols_full_name
     nil
   end
 
+  def can_edit_request_issues?(appeal)
+    Task.where(
+      type: [JudgeReviewTask, AttorneyTask],
```
lowellrex: Is there some reason we are excluding `JudgeAssignTask`s from this list of types? Can judges not edit request issues before a decision has been drafted on an appeal?

app/models/user.rb
```diff
@@ -78,6 +78,15 @@ def vacols_full_name
...
+      appeal: appeal,
+      assigned_to: self,
+      status: [Constants.TASK_STATUSES.assigned, Constants.TASK_STATUSES.in_progress]
```
lowellrex: Are we intentionally preventing attorneys and judges from editing request issues for tasks that are on hold?

app/models/user.rb
lowellrex: If we want to include `JudgeAssignTask`s (and [eventually `JudgeQualityReviewTask`s](https://github.com/department-of-veterans-affairs/caseflow/issues/8107)) we could modify this function to look something like:
```ruby
appeal.tasks.where(
  assigned_to: user,
  status: [Constants.TASK_STATUSES.assigned, Constants.TASK_STATUSES.in_progress]
).select { |t| t.is_a?(JudgeTask) || t.is_a?(AttorneyTask) }
```

client/app/queue/CaseDetailsLoadingScreen.jsx
```diff
@@ -50,6 +52,7 @@ class CaseDetailLoadingScreen extends React.PureComponent<Props> {
     if (!appealDetails || !(appealId in appealDetails)) {
       promises.push(
         ApiUtil.get(`/appeals/${appealId}`).then((response) => {
+          this.props.setCanEditRequestIssues(response.body.can_edit_request_issues);
```
lowellrex: This approach runs into the same problem [we had with the edit AOD button](https://github.com/department-of-veterans-affairs/caseflow/pull/8390) where we don't show the edit request issues button if we land on the case details page from the search page.

lowellrex: I think that is probably fine for now since we expect attorneys and judges to end up on the case details page through their queues, and navigating to the case details page from the queue views properly shows the edit button.

client/app/queue/StickyNavContentArea.jsx
```diff
@@ -120,9 +121,17 @@ const anchorJumpLinkStyling = css({
...
+  fontSize: '1.5rem',
+  fontWeight: 'normal',
+  margin: '5px'
+});
+
 const ContentSection = ({ element }) => <React.Fragment>
   <h2 {...sectionHeadingStyling}>
     <a id={`${getIdForElement(element)}`} {...anchorJumpLinkStyling}>{element.props.title}</a>
+    {element.props.editLink && <span className="cf-push-right" {...anchorEditLinkStyling}>
+      <Link href={element.props.editLink}>{element.props.editLinkTitle}</Link></span>}
```
lowellrex: Could we instead offload the rendering of this additional header content on the element itself? I can imagine a situation where different content areas would both want to display something in the title bar in addition to the section heading (perhaps a note of explanation or a [differently styled link](https://github.com/department-of-veterans-affairs/caseflow/issues/8216)).

The [`NavigationBar` establishes a pattern](https://github.com/department-of-veterans-affairs/caseflow-frontend-toolkit/blob/2cd92fbf82f9de212f8ff1bdc406d5b5f2cd583a/components/NavigationBar.jsx#L88) that I think we can replicate with something like:
```jsx
// Rendered in StickyNavContentArea.jsx
{element.additionalHeaderContent && element.additionalHeaderContent()}

// CaseDetailsIssueList.jsx
additionalHeaderContent = () => <span className="cf-push-right" {...anchorEditLinkStyling}>
  <Link href={element.props.editLink}>{element.props.editLinkTitle}</Link>
</span>
```

client/app/queue/CaseDetailsView.jsx
```diff
@@ -81,9 +85,11 @@ class CaseDetailsView extends React.PureComponent {
       <hr {...horizontalRuleStyling} />
       <StickyNavContentArea>
         <CaseDetailsIssueList
-          amaIssueType={featureToggles.ama_decision_issues || !_.isEmpty(appeal.decisionIssues)}
+          amaIssueType={amaIssueType}
           title="Issues"
           isLegacyAppeal={appeal.isLegacyAppeal}
+          editLink={amaIssueType && canEditRequestIssues && `/appeals/${appealId}/edit`}
+          editLinkTitle={COPY.CORRECT_REQUEST_ISSUES_LINK}
```
lowellrex: Can we move the displaying of edit issues link entirely inside the `CaseDetailsIssueList` component? I think the [`PowerOfAttorney`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/PowerOfAttorneyDetail.jsx) and [`VeteranDetail`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/VeteranDetail.jsx) components establish a pattern we can reproduce here. We could default to displaying nothing, make a request to a new [`can_edit_issues` endpoint](https://github.com/department-of-veterans-affairs/caseflow/blob/ffedf94503aaf3a0ee3e6fa8deda2886b392727a/app/controllers/appeals_controller.rb#L28), (maybe show an [ellipsis](http://bookofzeus.com/articles/css/animate-ellipsis-using-css3/) while the request is awaiting a response if we're feeling fancy???), and then show the link if the response tells us we should. One benefit of this approach is that we could offload a lot (all?) of the logic onto the backend.

client/app/queue/uiReducer/uiReducer.js
```diff
@@ -29,6 +29,7 @@ export const initialState = {
   selectedAssigneeSecondary: null,
   veteranCaseListIsVisible: false,
   canEditAod: false,
+  canEditRequestIssues: false,
```
lowellrex: Since the ability to edit issues depends on the combination of user and appeal can we include the appeal ID here (we don't have to worry about the user piece of the equation because the global store will be reset when the user logs out/in). I'm imagining the shape of this looking like:
```jsx
canEditRequestIssues: {
  "0d99920c-8355-4449-8b46-a9b20aa77717": true,
  "c4487080-c8ad-4996-badd-35bd58f8b4c2": false,
  "3565723": false
},
```

Alternatively, [if we encapsulate the displaying of the edit issues link](https://github.com/department-of-veterans-affairs/caseflow/pull/8391/files#r243598990) in the `CaseDetailsIssueList` component then we don't even have to keep that information in the global redux store and can instead keep that information in the component's local state.

app/models/user.rb
aroltsch: yes

client/app/queue/CaseDetailsLoadingScreen.jsx
aroltsch: yes, once a new developer joins the Decisions team, I will ask him to re-factor this into its own endpoint

client/app/queue/CaseDetailsView.jsx
aroltsch: I don't think we should have `can_edit_issues` endpoint. I think we should make it more general and return information on what a user can edit (e.g, AOD, request issues, perhaps power of attorney at a later date, etc). I don't see a problem with this current approach for now. You should be able to pass editLink and editLinkTitle to the StickyNavContentArea component and this should work for PowerOfAttorney and VeteranDetail

app/models/user.rb
aroltsch: That's a good point. I think we should let judges edit request issues when it is in the assign status.

client/app/queue/CaseDetailsLoadingScreen.jsx
mdbenjam: FWIW I think this is an argument for any editing to exist in the task action drop-down. We already have a lot of structures in place to pass the right information when loading tasks. Though admittedly there is a design question there.

app/models/appeal.rb
```diff
@@ -100,7 +100,7 @@ def attorney_case_reviews
   end
 
   def every_request_issue_has_decision?
-    request_issues.all? { |request_issue| request_issue.decision_issues.present? }
+    eligible_request_issues.all? { |request_issue| request_issue.decision_issues.present? }
```
aroltsch: This change is not related to this PR, btw.

aroltsch: I can create a new PR if you feel strongly about it.

client/app/queue/CaseDetailsView.jsx
```diff
@@ -81,9 +91,13 @@ class CaseDetailsView extends React.PureComponent {
       <hr {...horizontalRuleStyling} />
       <StickyNavContentArea>
         <CaseDetailsIssueList
-          amaIssueType={featureToggles.ama_decision_issues || !_.isEmpty(appeal.decisionIssues)}
+          amaIssueType={amaIssueType}
           title="Issues"
           isLegacyAppeal={appeal.isLegacyAppeal}
+          additionalHeaderContent={amaIssueType && appeal.canEditRequestIssues &&
```
aroltsch: Let me know what you think of this approach. I looked into creating a method on the `CaseDetailsIssueList` component but in order to access it outside the component I need to use a `Ref` attribute. 
React docs recommend against it: https://reactjs.org/docs/refs-and-the-dom.html

```
You can define any public method on your component classes (such as a reset method on a Typeahead) and call those public methods through refs (such as this.refs.myTypeahead.reset()). In most cases, it's clearer to use the built-in React data flow instead of using refs imperatively.
```

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -32,6 +32,10 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
     end
   end
 
+  attribute :can_edit_request_issues do
+    RequestStore[:current_user]&.can_edit_request_issues?(object)
```
lowellrex: Let's pass the user as [an argument to the serializer](https://github.com/department-of-veterans-affairs/caseflow/blob/6bea2aab4a10d6bf3de6b4b26ab1f229c9f5637a/app/controllers/tasks_controller.rb#L236) like we do for [the `TaskSerializer`](https://github.com/department-of-veterans-affairs/caseflow/blob/6bea2aab4a10d6bf3de6b4b26ab1f229c9f5637a/app/models/serializers/work_queue/task_serializer.rb#L84).

client/app/queue/CaseDetailsView.jsx
lowellrex: Interesting. Maybe we could use a higher order component in `StickyNavContentArea` to accomplish the encapsulation we'd like (@mdbenjam would know better than I about how to appropriately do this in React), but I think this looks good for now!

app/models/serializers/work_queue/appeal_serializer.rb
mdbenjam: I generally agree we should pass the user as a parameter instead of using request store. The problem is the request store is an implicit parameter that it's not clear we rely on. Let's say a job were to use this serializer (admittedly unlikely), we'd throw 500s if we forget to set user in the request store. Having the parameter makes sure we don't forget to set it.

aroltsch: Passing a parameter to the serializer is also optional. You could make the same mistake.

PR #8393, opened 2018-12-20
Legacy Case Distribution Tweaks

cmgiven:
* We've learned more from Jed about what cases need to be prevented from being distributed. These changes reflect that.
* Before merging, we also may adjust the batch size per attorney, so that judges receive fewer cases each request.

### DIFF
--- a/app/models/vacols/case_docket.rb
+++ b/app/models/vacols/case_docket.rb
@@ -14,9 +14,48 @@ class VACOLS::CaseDocket < VACOLS::Record
     select BFKEY, BFDLOOUT, BFMPRO, BFCURLOC, BFAC, BFHINES, TINUM, TITRNUM, AOD
     from BRIEFF
     #{VACOLS::Case::JOIN_AOD}
-    #{VACOLS::Case::JOIN_SPECIALTY_CASE_TEAM_ISSUES}
+    #{JOIN_MAIL_BLOCKS_DISTRIBUTION}
+    #{JOIN_DIARY_BLOCKS_DISTRIBUTION}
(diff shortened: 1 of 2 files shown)


app/models/vacols/case_docket.rb
```diff
@@ -14,9 +14,48 @@ class DocketNumberCentennialLoop < StandardError; end
...
+        from MAIL
+        MLCOMPDATE is null and MLTYPE not in ('02', '05', '08', '13')
+        group by MLFOLDER
+      ) MAIL
+      on MAIL.MLFOLDER = BRIEFF.BFKEY
+    )
+    on MAILKEY = BFKEY
+  ".freeze
+
+  JOIN_DIARY_BLOCKS_DISTRIBUTION = "
```
amprokop: Can we add a comment explaining that we're blocking diaries with certain types? Guessing EXT = extension, POA = poa clarification, HCL = ??

cmgiven:
Realized this is still pending your approval, @amprokop.

spec/models/vacols/case_docket_spec.rb
```diff
@@ -16,6 +16,7 @@
   let(:another_judge) { create(:user) }
   let!(:another_vacols_judge) { create(:staff, :judge_role, sdomainid: another_judge.css_id) }
 
+  let(:nonpriority_ready_case_bfbox) { nil }
```
mdbenjam: Could we add a test for the diary blocking?

cmgiven: The reason I didn't do this is that we don't have a factory for the ASSIGN table (or a model, etc.). Thoughts there?

PR #8395, opened 2018-12-20
[wip] add task timer and time evidence submission tasks

amprokop:

### Description

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...


app/models/concerns/timeable_task.rb
```diff
@@ -0,0 +1,23 @@
...
+      fail Caseflow::Error::MissingTimerInfo unless method_defined?(:when_timer_ends)
+      fail Caseflow::Error::MissingTimerInfo unless respond_to?(:timer_delay)
+
+      super(args).tap do |task|
+        TaskTimer.create!(task: task, submitted_at: Time.zone.now + timer_delay)
+      end
+    end
+
+    def create!(args)
+      fail Caseflow::Error::MissingTimerInfo unless method_defined?(:when_timer_ends)
```
aroltsch: If these two methods are identical why don't you just call one of the methods inside another method to not duplicate the logic. 


```
def create!(args)
       create(args)
end
```

app/models/concerns/timeable_task.rb
```diff
@@ -0,0 +1,23 @@
+module TimeableTask
+  extend ActiveSupport::Concern
+
+  module ClassMethods
+    def create(args)
+      fail Caseflow::Error::MissingTimerInfo unless method_defined?(:when_timer_ends)
```
aroltsch: `Caseflow::Error::MissingTimerInfo` might not be an accurate error name. Perhaps, `MissingRequiredMethodDefinitions`. You can also do something like that https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/concerns/uploadable_document.rb#L18

app/jobs/task_timer_job.rb
```diff
@@ -0,0 +1,21 @@
+class TaskTimerJob < CaseflowJob
+  queue_as :low_priority
+  application_attr :queue
+
+  def perform
+    RequestStore.store[:application] = "queue"
+    RequestStore.store[:current_user] = User.system_user
+
+    TaskTimer.lock.requires_processing.includes(:task).each do |task_timer|
```
aroltsch: Do you need to lock the entire table or just one row? ```ActiveRecord::Base#lock! method to lock one record by id```

app/models/concerns/timeable_task.rb
amprokop: oh, I know why — because they're not identical.

`create` doesn't throw errors if creation fails, `create!` does. i didn't want to change any behavior. but maybe it wouldn't matter here.

app/jobs/task_timer_job.rb
amprokop: cool ok. locking by row now. i think this adds another N queries. but we can optimize later.

app/models/concerns/timeable_task.rb
aroltsch: But both of them call `create!`, no? Also if your method raises an error, it should end with `!`, your `create` raises MissingTimerInfo error if methods are not defined. 

You can also mimic what's done in this module: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/concerns/uploadable_document.rb#L18

amprokop: Made the change to remove `create`

PR #8397, opened 2018-12-20
Sort gems alphabetically within groups in Gemfile

monfresh:
**Why**: It makes it easier to find gems visually.

**How**: Add the `Bundler/OrderedGems` cop to the Rubocop config, remove all
blank lines in the Gemfile, and run `rubocop -a` to automatically sort the gems
within each `group`. Note that if you separate gems with a blank line, Rubocop
considers those 2 different groups, so they won’t be sorted together.


.rubocop.yml
```diff
@@ -13,6 +7,13 @@ AllCops:
     - 'vendor/**/*'
     - './.simplecov'
     - 'client/node_modules/**/*'
+  TargetRubyVersion: 2.5
```
tomas-nava: if you don't set `TargetRubyVersion` here, RuboCop will get the value from our `.ruby-version` file, which I think is good? That way we don't have to update ruby version in two places.

monfresh: Ah, nice. All this time I missed that memo :)

monfresh: I removed it.

app/controllers/certification_cancellations_controller.rb
```diff
@@ -6,9 +6,9 @@ def create
     @certification_cancellation = CertificationCancellation.new(certification_cancellation_params)
 
     if @certification_cancellation.save
-      render json: { is_cancelled: true }, status: 201
+      render json: { is_cancelled: true }, status: :created
```
aroltsch: I personally prefer using numbers for HTTP codes.

app/mappers/hearing_day_mapper.rb
```diff
@@ -3,22 +3,22 @@ class InvalidRegionalOfficeError < StandardError
...
-    board_member:      :judge_id,
-    team:              :team,
-    adduser:           :created_by,
-    addtime:           :created_at,
-    mduser:            :updated_by,
-    mdtime:            :updated_at,
-    vdbvapoc:          :bva_poc,
-    notes:             :notes,
-    judge_last_name:   :judge_last_name,
+    hearing_pkseq: :id,
```
aroltsch: I find the above more readable.

app/mappers/issue_mapper.rb
```diff
@@ -56,7 +56,8 @@ def validate!(issue_attrs)
     def rename(issue_attrs)
       COLUMN_NAMES.keys.each_with_object({}) do |k, result|
         # skip only if the key is not passed, if the key is passed and the value is nil - include that
-        next unless issue_attrs.keys.include? k
+        next unless issue_attrs.key?(k)
```
aroltsch: Nice

app/models/end_product_establishment.rb
```diff
@@ -451,7 +455,8 @@ def find_open_modifier
   end
 
   def sync_source!
-    return unless source && source.respond_to?(:on_sync)
+    return unless source&.respond_to?(:on_sync)
```
aroltsch: Nice

Gemfile
```diff
@@ -82,7 +82,7 @@ group :test, :development, :demo do
   # Testing tools
   gem "capybara"
   gem "capybara-screenshot"
-  gem "danger", "5.5.5"
+  gem "danger", "~> 5.10"
   gem "database_cleaner"
```
tomas-nava: The new version of Danger has an annoying bug 😞

https://github.com/danger/danger/issues/1052

monfresh: Fixed here: https://github.com/department-of-veterans-affairs/caseflow/pull/8545

tomas-nava: Thanks!

Issue #8079, opened 2018-12-03
AHL | Facility Locator Prod Implementation


sharonwarner:
Requires #8078 

AC:
From the rails console, we should be able to run `appeal.veteran.hearing_locations` and receive a json object with the veteran's RO's alternate hearing locations and their associated distances.

PR #8402, opened 2018-12-20
VADotGovService

### LINKED ISSUE #8078: AHL | Vets 360 Prod Implementation
From the rails console, we should be able to run `appeal.veteran.geolocation` and save the lat long the vets 360 API returns.

lomaxap:
Resolves #8078 and #8079 

### Description
- Adds VADotGovService with 2 methods
  - `#geocode` which takes `address:, city:, state:, zip_code:, country:` params and returns a lat/long point 
   - `#get_distance` takes a lat/long point and a list of facility ids and returns all those facilities and their distance from that point from closest to furthest


app/services/external_api/va_dot_gov_service.rb
```diff
@@ -0,0 +1,156 @@
+require "json"
+
+class ExternalApi::VADotGovService
```
mdbenjam: Meta point. I think one way we could structure this better than we've built these services in the past is to move more of the code into a testable place. Most of this code could live in `VADotGovService` and then we could have `Fakes::VADotGovEndpoint` and `ExternalApi::VADotGovEndpoint`. `Fakes::VADotGovEndpoint` could literally mock out the responses i.e. you call get_facilities with a page number and it returns a json object for that page. Then we can test things like calling pages until we find the ids. I don't think it's necessarily worth making that change now, but it's just something I was thinking of that we could improve in a lot of our services.

app/services/external_api/va_dot_gov_service.rb
```diff
@@ -0,0 +1,156 @@
...
+          query: { lat: point[0], long: point[1], page: page },
+          ids: ids
+        )
+
+        ids -= results[:facilities].pluck(:id)
+        facility_results += results[:facilities]
+
+        break if !results[:has_next]
+        page += 1
+        sleep 1
```
mdbenjam: Is this to deal with the rate limiting?

app/services/external_api/va_dot_gov_service.rb
```diff
@@ -0,0 +1,156 @@
+require "json"
+
+class ExternalApi::VADotGovService
+  class << self
+    # :nocov:
+    def get_distance(point, ids)
+      page = 1
+      facility_results = []
+
+      until ids.empty?
```
mdbenjam: Nit: I don't love mutating a parameter of the function. Maybe create a `remaining_ids` variable?

app/services/external_api/va_dot_gov_service.rb
```diff
@@ -0,0 +1,156 @@
...
+
+        ids -= results[:facilities].pluck(:id)
+        facility_results += results[:facilities]
+
+        break if !results[:has_next]
+        page += 1
+        sleep 1
+      end
+
+      Rails.logger.info("Unable to find api.va.gov facility data for: #{ids.join(', ')}.") if !ids.empty?
```
mdbenjam: Do we want to throw an error or something here? If this is a case that should never happen, we probably want to know about it.

app/services/external_api/va_dot_gov_service.rb
```diff
@@ -0,0 +1,156 @@
...
+    def check_for_error(response_body:, code:)
+      case code
+      when 200 # rubocop:disable Lint/EmptyWhen
+      when 400
+        fail Caseflow::Error::VaDotGovRequestError, code: code, message: response_body
+      when 500
+        fail Caseflow::Error::VaDotGovServerError, code: 502, message: response_body
+      else
+        msg = "Error: #{response_body}, HTTP code: #{code}"
+        fail Caseflow::Error::VaDotGovServerError, code: 502, message: msg
```
mdbenjam: I think we should use the code as code. Instead of 502 here.

app/services/external_api/va_dot_gov_service.rb
```diff
@@ -0,0 +1,156 @@
...
+      { facilities: facilities, has_next: query[:page] + 1 <= total_pages }
+    end
+
+    def send_va_dot_gov_request(query: {}, headers: {}, endpoint:, method: :get, body: nil)
+      url = URI.escape(base_url + endpoint)
+      request = HTTPI::Request.new(url)
+      request.query = query
+      request.open_timeout = 600 # seconds
+      request.read_timeout = 600 # seconds
+      request.auth.ssl.ssl_version  = :TLSv1_2
```
mdbenjam: I'd see if we can get rid of this and the cert below.

app/services/external_api/va_dot_gov_service.rb
```diff
@@ -0,0 +1,156 @@
...
+      end
+
+      { facilities: facilities, has_next: query[:page] + 1 <= total_pages }
+    end
+
+    def send_va_dot_gov_request(query: {}, headers: {}, endpoint:, method: :get, body: nil)
+      url = URI.escape(base_url + endpoint)
+      request = HTTPI::Request.new(url)
+      request.query = query
+      request.open_timeout = 600 # seconds
```
mdbenjam: Maybe lower this. Not sure we want a 10 minute timeout for this endpoint. Sounds like it's reasonably fast.

app/services/external_api/va_dot_gov_service.rb
```diff
@@ -0,0 +1,156 @@
...
+          requestCountry: {
+            country_code: country
+          },
+          zipCode5: zip_code
+        }
+      }
+    end
+
+    def facility_response_data(resp_body)
+      [
```
mdbenjam: nit: I'd use a hash instead of a returned array. The array assumes the caller knows what is in position 0, 1, and 2. The hash allows you to provide context.

app/services/external_api/va_dot_gov_service.rb
```diff
@@ -0,0 +1,156 @@
...
+        break if !results[:has_next]
+        page += 1
+        sleep 1
+      end
+
+      Rails.logger.info("Unable to find api.va.gov facility data for: #{ids.join(', ')}.") if !ids.empty?
+      facility_results
+    end
+
+    def geocode(**args)
```
mdbenjam: Why not use named args here? Feels sort of mystic what should be passed in.

app/services/external_api/va_dot_gov_service.rb
lomaxap: yes

app/services/external_api/va_dot_gov_service.rb
lomaxap: Just to simplify the code since the arguments are all just passed to another method. The arguments are named in the geocode_body method.

app/services/external_api/va_dot_gov_service.rb
lomaxap: True. But then we can’t write that nice, clean destructuring statement below!

mdbenjam: Fair enough. I think I value making it a little more clear over conciseness, but it's also a personal preference, so I'll leave it up to you!

app/services/external_api/va_dot_gov_service.rb
mdbenjam: I'd use the same argument as above that I value clarity over conciseness, but again, it's a personal preference, so it's up to you!

app/services/external_api/va_dot_gov_service.rb
lomaxap: the code shouldn't fail if there are a handful of ids that aren't found, but we do want to know about it so we can check the data. would we get an alert from `Rails.logger.error`?

mdbenjam: We could throw a sentry error without failing. Here's code that sends a sentry error: https://github.com/department-of-veterans-affairs/caseflow/blob/bf90bb2acd8219dc46b36c73eb09a63cd79e6f6f/app/controllers/application_controller.rb#L12

And we wouldn't have to fail then. Though I would counter that maybe failing is good here. If we don't fail, we might not show the closest facility to the HS coordinator.

lomaxap: ok. failing works for me then.

app/services/external_api/va_dot_gov_service.rb
```diff
@@ -0,0 +1,153 @@
+require "json"
+
+class ExternalApi::VADotGovService
+  class << self
+    # :nocov:
+    def get_distance(point, ids)
```
mdbenjam: I'd say the same thing about point. Clearer if it's a hash instead of an array.

PR #8403, opened 2018-12-20
Add non-compensation issue categories

leikkisa:
connects #5882


app/services/legacy_optin_manager.rb
```diff
@@ -26,7 +26,7 @@ def process!
           next unless vacols_id == legacy_appeal.vacols_id
 
           # gut checks
-          unless open_legacy_issues.map(&:vacols_sequence_id).map(&:to_s).include?(vacols_sequence_id)
+          unless open_legacy_issues.map(&:vacols_sequence_id).map(&:to_s).include?(vacols_sequence_id.to_s)
```
leikkisa: Temporary fix because of a failing test, this will be obsolete once the legacy issue opt in PR is merged.

leikkisa:
I made an assumption that the existing issue categories are for "Compensation", and that those are the ones to still use for Appeals because I didn't see "Compensation" in the list.

I am also wondering if I should add "Unknown issue category" at the top of every list? I kind of remember that coming up in conversation, but I couldn't remember if it was that or "Other" which is in some of the lists.

spec/support/intake_helpers.rb
```diff
@@ -1,5 +1,108 @@
 # rubocop:disable Metrics/ModuleLength
 module IntakeHelpers
+  # rubocop: disable Metrics/MethodLength
```
pkarman: I like to see these helpers consolidated.

spec/feature/intake/nonrating_request_issue_modal_spec.rb
```diff
@@ -0,0 +1,91 @@
+require "rails_helper"
```
pkarman: I like this!

PR #8405, opened 2018-12-21
Edit request issues by id

jiahuang:
Does the following:
* Going to intake/edit sends previously added request issues ids. If these ids exist, don't create a new issue. If these ids do not exist (such as when removing or adding a new issue), remove/create the issue 
* Fixes a bug where if you have an existing request issue A contesting a decision issue and save via the edit page, request issue A gets removed and readded.
* removed some extra js code used for the old intake edit page


app/models/request_issue.rb
```diff
@@ -76,9 +76,9 @@ def from_intake_data(data)
     end
 
     def find_or_build_from_intake_data(data)
-      find_or_initialize_by(
-        attributes_from_intake_data(data)
-      ).tap(&:validate_eligibility!)
+      # request issues on edit have ids
+      # but newly added issues do not
+      data[:request_issue_id] ? RequestIssue.find_by(id: data[:request_issue_id]) : from_intake_data(data)
```
pkarman: when you use `RequestIssue.find(data[:request_issue_id])` it's the same as `find_by(id: ...)`

spec/feature/intake/edit_spec.rb
```diff
@@ -176,6 +176,56 @@ def check_row(label, text)
       expect(page).to have_button("Save", disabled: false)
     end
 
+    context "with multiple request issues with same data fields" do
+      let!(:duplicate_nonrating_request_issue) do
+        create(:request_issue,
+               review_request: appeal,
```
pkarman: would it make maintenance clearer if instead of hardcoding all the duplicate values, instead we just used the original record's attributes?

spec/feature/intake/edit_spec.rb
```diff
@@ -356,6 +406,55 @@ def verify_decision_issues_can_be_added_and_removed(page_url,
...
+
+    expect(first_not_modified_request_issue).to_not be_nil
+    expect(second_not_modified_request_issue).to_not be_nil
+
+    non_modified_ids = [first_not_modified_request_issue.id, second_not_modified_request_issue.id]
+    request_issue_update = RequestIssuesUpdate.find_by(review: decision_review)
+
+    # existing issues should not be added or removed
+    expect(request_issue_update.created_issues
+      .select { |created| non_modified_ids.include? created.id }.empty?).to eq(true)
```
pkarman: you can also pass a block to `any?` which might make this a little clearer

app/models/appeal.rb
```diff
@@ -204,6 +204,7 @@ def create_issues!(new_issues)
       issue.update!(benefit_type: benefit_type, veteran_participant_id: veteran.participant_id)
       issue.create_legacy_issue_optin if issue.legacy_issue_opted_in?
     end
+    request_issues.reload
```
jiahuang: ended up having to do a reload here due to https://dsva.slack.com/archives/C754H751T/p1546472019040600

Issue #8202, opened 2018-12-11
Schedule Hearing Task | Post response returns all Schedule Hearing Tasks


sharonwarner:
When we create or update a ScheduleHearingTask through create in the tasks_controller, the response returns all the ScheduleHearingTasks that are incomplete or were finished in the last two weeks. That slows down the response time significantly. We should update these responses to only include the task created.

amprokop:
Read a comment from @lomaxap that indicates that each Task serialization does 2 calls to VACOLS. This seems like the underlying problem IMO. Potentially we should think about batching those queries so there are a constant number of VACOLS queries. 

See  https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/services/appeal_repository.rb#L597 and https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/services/appeal_repository.rb#L45 for places in the codebase where that is done

lowellrex:
The two fields which are causing us to incur this VACOLS request overhead at serialization at `case_type` and `docket_number`. I think we need both of those on the frontend whenever we have a task, so I don't think we can lazy load them. I agree with @amprokop that batching is the way to go. Maybe we could do this in the TaskController so that we get the benefit of this for all legacy appeals. Perhaps something like the following (I don't think this works exactly, but I think it is a decent enough sketch):

```ruby
  def json_tasks(tasks)
    # Make a single request to VACOLS to grab all of the rows we want here?
    legacy_appeals = tasks.select { |t| t.appeal.is_a?(LegacyAppeal) }.map { |t| t.appeal }

    # Load these puppies into memory in a single batch.
    cases = VACOLS::Case.includes(:folder, :correspondent, :representatives).find(legacy_appeals.pluck(:vacols_id))

    # The vacols attributes for the legacy appeals don't stay in memory so we go ahead
    # and re-fetch them one by one when we run the serializer. Not entirely sure why.
    tasks.each do |t|
      if t.appeal.is_a?(LegacyAppeal)
        case_record = cases.select { |cr| cr.id == t.appeal.vacols_id }.first
        AppealRepository.set_vacols_values(appeal: t.appeal, case_record: case_record)
      end
    end

    ActiveModelSerializers::SerializableResource.new(
      tasks,
      each_serializer: ::WorkQueue::TaskSerializer,
      user: current_user
    ).as_json
  end
```

PR #8417, opened 2018-12-21
Limit Tasks Returned by Create/Update Methods

lomaxap:
Temporarily addresses #8202

- Adds `return_related_tasks` param to create/update methods on Tasks controller


lowellrex:
I think we can close this since the root problem should be addressed by #8427.

PR #8424, opened 2018-12-21
Noncomp landing page

jiahuang:
Connects #7990 

Also switched to using participant id instead of ssn

image


client/app/nonComp/pages/NonCompPage.jsx
```diff
@@ -1,13 +1,36 @@
 import React from 'react';
 import { connect } from 'react-redux';
+
 import NonCompTabs from '../components/NonCompTabs';
+import Button from '../../components/Button';
+
+const textAlignRightStyling = {
```
leikkisa: Why are we putting styling in here like this?

jiahuang: these are both kinda one offs for this specific page due to the button layout. Didn't seem useful to make a css class for them.

Issue #8199, opened 2018-12-11
Universal Case Title (additional elements)


sneha-pai:
#8150 began the process of moving our case-related information into a new element called Universal Case Title. Upon modifying other areas of case details, a few more elements were requested to be in this title area:
- VLJ on the case
- Attorney on the case
- if the veteran is the appellant or not
- if the case is a paper case

#### Narrow Width (768px)
![universal-case-title-expanded-768](https://user-images.githubusercontent.com/31224257/49819139-a4d6a780-fd42-11e8-8493-76d0f7853f97.png)

#### Wide width (1200px)
![universal-case-title-expanded-1200](https://user-images.githubusercontent.com/31224257/49819220-dc455400-fd42-11e8-825a-aa260024a95b.png)

@lowellrex @rkreyhsig @laurjpeterson @allyceh @eric-nava please provide feedback in comments below. Feedback specifically needed on:
- Feasibility of this data being able to display in these two different widths (depending on browser width on-load)
- This element has gotten quite tall - is that a problem?
- Overall composition critique
- any alternative approaches not proposed

lowellrex:
> * Feasibility of this data being able to display in these two different widths (depending on browser width on-load)

I personally don't know how to achieve the effect of different background colors for the first and second line of that list when we rely on the browser width to determine which elements get pushed onto the second row (and couldn't find any readily-googleable techniques to do that a few minutes of searching). So that might be tough to do (maybe @joeyyang has thoughts on how to accomplish this?). Would it be bad to have the same light grey background for both lines?

sneha-pai:
It wouldn't be bad to have the same light grey for both lines. I can update the mock to reflect this.

allyceh:
This looks good! A couple comments: 

- We could shorten the "The veteran is not the appellant" and "This is a paper case" to "Contested claim" and "Paper case". As I learn more about contested claims, that is the verbiage that seems to resonate most with everyone. "The veteran is not the appellant" can mean two things, 1) the claim is contested, or 2) the veteran is deceased, and a surviving spouse wants the benefits (not a contested claim). 
  - Are we looking to communicate both here, or just one of those things?
- We could consider using the same text style for both, unless there is a good reason for making one italics + one non-italics

PR #8426, opened 2018-12-21
Still needs polishing and tests

### LINKED ISSUE #8289: Add VLJ and Attorney working case to Universal Case Title
image

## Acceptance criteria
* [ ] Show VLJ in universal case title if one has been assigned for this case.
* [ ] Show attorney in universal case title if one has been assigned for this case.

## Non-goal
* Different color for first and second lines of the universal case title (`CaseTitleDetails` component) if the list of elements in that component spill over onto a second line. The background for all of the elements of this list will be the same light grey color even if they spill over onto a second line.

rkreyhsig:
Resolves #8289 - Add VLJ and Attorney working case to Universal Case Title

### DIFF
--- a/client/COPY.json
+++ b/client/COPY.json
@@ -67,6 +67,8 @@
   "CASE_DETAILS_POA_VSO": "Veteran Service Organization",
   "CASE_DETAILS_POA_AGENT": "Private Agent",
   "CASE_DETAILS_POA_ATTORNEY": "Private Attorney",
+  "CASE_DETAILS_VET_NOT_APELLANT": "The veteran is not the apellant",
+  "CASE_DETAILS_PAPER_CASE": "This is a paper case",
 
   "CASE_DETAILS_GENDER_FIELD_LABEL": "Gender",
   "CASE_DETAILS_GENDER_FIELD_VALUE_FEMALE": "Female",
--- a/client/app/queue/CaseDetailsView.jsx
+++ b/client/app/queue/CaseDetailsView.jsx
@@ -15,6 +15,8 @@ import CaseHearingsDetail from './CaseHearingsDetail';
 import PowerOfAttorneyDetail from './PowerOfAttorneyDetail';
 import CaseTitle from './CaseTitle';
 import CaseTitleDetails from './CaseTitleDetails';
+import CaseTitleSupportingDetails from './CaseTitleSupportingDetails';
+
 import TaskSnapshot from './TaskSnapshot';
 import CaseDetailsIssueList from './components/CaseDetailsIssueList';
 import StickyNavContentArea from './StickyNavContentArea';
@@ -71,6 +73,7 @@ class CaseDetailsView extends React.PureComponent {
         {success.detail}
       </Alert>}
       <CaseTitleDetails appealId={appealId} redirectUrl={window.location.pathname} />
+      <CaseTitleSupportingDetails appealId={appealId} redirectUrl={window.location.pathname} />
       { this.props.veteranCaseListIsVisible &&
         <VeteranCasesView
           caseflowVeteranId={appeal.caseflowVeteranId}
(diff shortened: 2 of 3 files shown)


client/COPY.json
```diff
@@ -67,6 +67,7 @@
   "CASE_DETAILS_POA_VSO": "Veteran Service Organization",
   "CASE_DETAILS_POA_AGENT": "Private Agent",
   "CASE_DETAILS_POA_ATTORNEY": "Private Attorney",
+  "CASE_DETAILS_VET_NOT_APELLANT": "The veteran is not the apellant",
```
lowellrex: One small style suggestion. We prefer to use the full term "Veteran" (capitalized) when referring to Veterans and don't usually shorten in (`git grep -i VETERAN | grep -i 'VET[^E]'`) so let's maintain that convention in the text of the copy string as well as the variable name here.

lowellrex:
Looks like we're making progress for sure!
image

A few notes though. First, as [mentioned in the ticket](https://github.com/department-of-veterans-affairs/caseflow/issues/8289#issue-390877265) (and [confirmed by Sneha](https://github.com/department-of-veterans-affairs/caseflow/issues/8199#issuecomment-446371576) on the original ticket) the work this ticket aims to accomplish is only adding the VLJ and attorney to the line of additional information below the primary case title (in the component `CaseTitleDetails`). If the elements inside that component are too wide to fit on a single line the elements should wrap onto a second line but should still have the grey background.

Second, we should hide VLJ and attorney if we do not have that information.

Third, we don't need to include the paper case/appellant not Veteran information in this PR.

client/COPY.json
lowellrex: Also, I think appellant is missing a P in the text above.

client/COPY.json
```diff
@@ -67,6 +67,7 @@
   "CASE_DETAILS_POA_VSO": "Veteran Service Organization",
   "CASE_DETAILS_POA_AGENT": "Private Agent",
   "CASE_DETAILS_POA_ATTORNEY": "Private Attorney",
+  "CASE_DETAILS_VETERAN_NOT_APPELLANT": "The Veteran is not the apellant",
```
lowellrex: Let's remove this bit of copy until we start using it in the application.

client/app/queue/CaseDetailsView.jsx
```diff
@@ -15,6 +15,7 @@ import CaseHearingsDetail from './CaseHearingsDetail';
 import PowerOfAttorneyDetail from './PowerOfAttorneyDetail';
 import CaseTitle from './CaseTitle';
 import CaseTitleDetails from './CaseTitleDetails';
+
```
lowellrex: Let's undo this change.

client/app/queue/CaseTitleDetails.jsx
```diff
@@ -121,6 +121,20 @@ export class CaseTitleDetails extends React.PureComponent {
...
+        <h4>{COPY.TASK_SNAPSHOT_ASSIGNED_JUDGE_LABEL}</h4>
+        <div>
+          {appeal.assignedJudge ? appeal.assignedJudge.full_name : ''}
+        </div>
+      </React.Fragment> }
+
+      { appeal.assignedAttorney && <React.Fragment>
+        <h4>{COPY.TASK_SNAPSHOT_ASSIGNED_ATTORNEY_LABEL}</h4>
+        <div>
+          {appeal.assignedAttorney ? appeal.assignedAttorney.full_name : ''}
```
lowellrex: Since we're inside a conditional block that already checks to see if the `assignedAttorney` attribute is truthy I think we can rewrite this line as `<div> appeal.assignedAttorney.full_name</div>`.

lowellrex:
Addressed the wrapping issue:
image

lowellrex:
Two last suggestions:
1. Let's remove this information from [the `TaskSnapshot` component](https://github.com/department-of-veterans-affairs/caseflow/blob/6c71e6febac121ef263f9d51e2a645559bb39a32/client/app/queue/TaskSnapshot.jsx#L248) so we only show it in one place on the case details page and
2. Let's hide these fields from the current user if they are a member of a VSO using the `userIsVsoEmployee` property

PR #8427, opened 2018-12-22
Reduce number of VACOLS requests when calling tasks endpoints

### LINKED ISSUE #8202: Schedule Hearing Task | Post response returns all Schedule Hearing Tasks

lowellrex:
Resolves #8202. Eager loads all of the `VACOLS::Case`s associated with a set of `Task`s so that the serializer does not make requests to VACOLS for every individual task.

I am still struggling to write a failing test for this, but I have been able to confirm that we only make one request after this change.

## Steps to reproduce:
* Log in as `BVASCASPER1` or another user with tasks associated with `LegacyAppeals`
* Set a breakpoint in `AppealsRepository.find_case_record()` so we can count the number of requests we make to fetch `VACOLS::Case`s
* Visit `/queue`
* Observe that we only make 1 request to fetch `VACOLS::Case`s
* Replace [`eager_load_legacy_appeals_batch(tasks)`](https://github.com/department-of-veterans-affairs/caseflow/pull/8427/files#diff-f4416012d567d02da854cfbc9e2df2caR234) with `tasks` to simulate the way we were building the response prior to this change
* Visit `/queue`
* Observe that we only make many requests to fetch `VACOLS::Case`s, 1 request for each task associated with `LegacyAppeals` (8 in the current state of my development environment)


aroltsch:
Looks great!

PR #8428, opened 2018-12-24
Add veteran ahl job

### LINKED ISSUE #8414: Job to download AHL for Appeals
# Problem
Before a Veteran can be scheduled for a hearing, we need to alert the hearing coordinator of any possible alternative hearing locations (AHL). Determining the AHLs is slow (about 1-2 seconds per Veteran). Unfortunately there are a lot of Veterans that need to be scheduled at once, making the requests to get the AHL for all these Veterans will take on the order of minutes. This means we cannot make all of the requests on page load.

# Solution
Background job. We can write a background job that gets the AHL for all of the Veterans' that need to be scheduled. We can save these AHL in a DB table so that they're easy to query.

## AHL Table
Create a new `hearing_location` table. Each `appeal` can have multiple records in the `hearing_location` table. Each record will have the following columns.

- `distance` 
- `facility_id`
- `name`
- `address`

## Job
We should run the job on a cron every hour for a limited set of Veterans. Running it frequently for small numbers of Veterans has some advantages
1) We are rate limited, so spreading the load out will allow us to avoid hitting these limits.
1) When we start the job the first time there will be thousands of Veterans to query. Similarly if the job ever fails for multiple days we'll have a backlog of Veterans to schedule. Spreading this out over hour chunks will ensure we don't hit any rate limits, nor overwhelm any dependencies.

Query for the Appeals we need to schedule.
```
vacols_ids = VACOLS::Case.where(bfcurloc: 57).pluck(:id)

# Find VACOLS cases that don't have appeals in our system.
existing_vacols_ids = LegacyAppeal.where(vacols_id: vacols_ids).pluck(:vacols_id)
missing_vacols_ids = vacols_ids - existing_vacols_ids

# Create missing legacy appeals
missing_vacols_ids.each do |vacols_id| # Schedule no more than 500/hour
  LegacyAppeal.find_or_create_by_vacols_id(vacols_id)
end

appeals = LegacyAppeal.where(vacols_id: vacols_ids).includes(:hearing_locations).where(hearing_location: { id: nil }).limit(500)

appeals.each do |appeal|
  veteran = appeal.veteran
  lat, long = ExternalApi::VADotGovService.geocode(
    address_line_1: veteran.address_line1,
    address_line2: veteran.address_line2,
    address_line3: veteran.address_line3,
    city: veteran.city,
    state: veteran.state,
    country: veteran.country,
    zip_code: veteran.zip_code
  )

  ids = ids_for_ro # to be defined
  ExternalApi::VADotGovService(lat, long, ids).each do |alternate_hearing_location|
    AlternateHearingLocation.create(appeal_id: appeal.id, **alternate_hearing_location)
  end
end
```

mdbenjam:
Resolves #8414 and merges #8402

### Description
- Adds `VADotGovService`
- Adds `FetchHearingLocationsForVeteransJob`


app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -0,0 +1,77 @@
...
+  end
+
+  def create_available_locations_for_veteran(veteran, lat, long, ids)
+    VADotGovService.get_distance(lat: lat, long: long, ids: ids).each do |alternate_hearing_location|
+      AvailableHearingLocations.create(
+        veteran_file_number: veteran.file_number,
+        distance: alternate_hearing_location[:distance],
+        facility_id: alternate_hearing_location[:id],
+        name: alternate_hearing_location[:name],
+        address: alternate_hearing_location[:address]["address_1"]
```
mdbenjam: Do we want to store the full address? What do we need to show in the UI?

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -0,0 +1,77 @@
...
+        address_line3: veteran.address_line3,
+        city: veteran.city,
+        state: veteran.state,
+        country: veteran.country,
+        zip_code: veteran.zip_code
+      )
+
+      facility_ids = facility_ids_for_veteran veteran
+
+      create_available_locations_for_veteran veteran, lat, long, facility_ids
```
mdbenjam: nit: I feel like we normally pass parameters with param, but if it's not a rubocop rule then it's up to you.

app/services/external_api/va_dot_gov_service.rb
```diff
@@ -0,0 +1,160 @@
...
+
+        remaining_ids -= results[:facilities].pluck(:id)
+        facility_results += results[:facilities]
+
+        break if !results[:has_next]
+        page += 1
+        sleep 1
+      end
+
+      unless remaining_ids.empty?
```
mdbenjam: If we're concerned about how many pages we need to search, we can add a DataDog metric here.

app/services/external_api/va_dot_gov_service.rb
```diff
@@ -0,0 +1,160 @@
...
+        msg = "Unable to find api.va.gov facility data for: #{remaining_ids.join(', ')}."
+        fail Caseflow::Error::VaDotGovAPIError, code: 500, message: msg
+      end
+
+      facility_results
+    end
+
+    def geocode(**args)
```
mdbenjam: I will point this out again that I'm not a fan of using `**args`. I know it makes things a little more concise in this method, but it makes it super unclear what you're supposed to pass. Anyone who wants to call geocode now has to first look at the body of this method, see that you pass `**args` to `geocode_body` and then look at the signature for `geocode_body`. I feel that's a pretty bad developer experience. It also makes it much easier to pass the wrong values.

lib/fakes/va_dot_gov_service.rb
```diff
@@ -0,0 +1,144 @@
+class Fakes::VADotGovService < ExternalApi::VADotGovService
+  def self.send_va_dot_gov_request(endpoint:, **_args)
```
mdbenjam: Hmm feels like we're mocking out the wrong functions. Don't we want to mock out the fuctions that our job calls? Rather than this internal function?

spec/jobs/fetch_hearing_locations_for_veterans_job_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    end
+
+    context "#file_numbers" do
+      it "returns bfcorlid in CF file_number format" do
+        expect(@job.file_numbers[0]).to eq @bfcorlid_file_number
+      end
+    end
+
+    context "#missing_veteran_file_numbers" do
+      it "returns list of file_numbers with no associated veteran" do
```
mdbenjam: Might be useful to have to file numbers, one with a veteran and one without, to make sure we only create for veterans that don't exist.

spec/jobs/fetch_hearing_locations_for_veterans_job_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+
+      Fakes::BGSService.veteran_records = { @bfcorlid_file_number => veteran_record }
+    end
+
+    before(:each) do
+      @job = FetchHearingLocationsForVeteransJob.new
+    end
+
+    context "#file_numbers" do
+      it "returns bfcorlid in CF file_number format" do
```
mdbenjam: This should really be testing that we get file numbers for all appeals in location 57. Maybe create two cases, one in location 57 one in a different location. Make sure we return the file number for the one in 57.

spec/jobs/fetch_hearing_locations_for_veterans_job_spec.rb
```diff
@@ -0,0 +1,74 @@
+require "rails_helper"
+require "faker"
+
+describe FetchHearingLocationsForVeteransJob do
+  context "when there are appeals in locations 57 with no associated veteran" do
+    before(:all) do
+      @regional_office_id = "RO01"
+      @bfcorlid = "123456789S"
+      @bfcorlid_file_number = "123456789"
+      create(:case, bfcurloc: 57, bfregoff: @regional_office_id, bfcorlid: @bfcorlid)
```
mdbenjam: I'd move a lot of this to let statements that are closer to the tests instead of defining it all in a block at the top.

spec/jobs/fetch_hearing_locations_for_veterans_job_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    context "#create_missing_veterans" do
+      it "creates a veteran" do
+        @job.create_missing_veterans
+        expect(Veteran.count).to eq 1
+      end
+    end
+
+    context "#perform" do
+      before do
+        @distance = VADotGovService.fake_facilities_data[:meta][:distances][0][:distance]
```
mdbenjam: While we can use the fake service, I generally recommend mocking this out. That way we can change up the development fakes without breaking tests.

spec/services/external_api/va_dot_gov_service_spec.rb
```diff
@@ -0,0 +1,4 @@
+require "rails_helper"
+
+describe ExternalApi::VADotGovService do
```
mdbenjam: Do you want to write tests for this?

app/services/external_api/va_dot_gov_service.rb
lomaxap: I really do think this is a matter of preference between writing DRY code or being clearer sooner. The flip side is that if I want to add a parameter to this method, I'd need to remember to edit it in 3 places. But I don't feel as strongly about this, so I'm fine changing.

spec/jobs/fetch_hearing_locations_for_veterans_job_spec.rb
lomaxap: i'm not sure what you mean? these will be the same for all tests, but I can change it use `let` instead of `before`

lib/fakes/va_dot_gov_service.rb
lomaxap: Does it matter for the fake? I did this thinking I could write something like 
```
allow(VADotGovService).to receive(:fake_facilities_data).and_return(custom_mock_data)
```
to mock out HTTP response, but I think you're right, I should just test ExternalAPI::VADotGovService rather than the fake.

app/models/hearings/available_hearing_locations.rb
```diff
@@ -0,0 +1,3 @@
+class AvailableHearingLocations < ApplicationRecord
+  belongs_to :veteran, foreign_key: :file_number, primary_key: :veteran_file_number
```
sharonwarner: This seems more complicated than necessary, is there a downside to just using the veteran id?

app/services/external_api/va_dot_gov_service.rb
sharonwarner: I'd also prefer not using **aargs. It's easy enough to search for all the usages of this function if we want to add another parameter.

app/models/hearings/available_hearing_locations.rb
lomaxap: hmmm. sounds like it's fine to me. @mdbenjam, any reason we shouldn't use the veteran's id?

lib/fakes/va_dot_gov_service.rb
mdbenjam: The purpose of these fakes is less for tests and more for doing things in local dev. If you want to run this AHL job in the rails console right now, you can't because the functions the job calls are not defined. In your tests you can  mock out the public function i.e. `geocode` and `get_distance`, or you could mock out the `send_va_dot_gov_request` on the real service.

app/models/hearings/available_hearing_locations.rb
mdbenjam: Mixed feelings here. I wanted us to join all our tables that link to the veteran table via the primary key. But Shane was pretty adamant about using the veteran_file_number. Now I'm hesitant to have some tables connect with id and some with veteran_file_number. That's the context, but I'll leave it up to you all.

lomaxap: ok. i'm fine with staying consistent.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -0,0 +1,85 @@
...
+
+    RegionalOffice::CITIES[ro][:alternate_locations] || [] << RegionalOffice::CITIES[ro][:facility_locator_id]
+  end
+
+  def full_address_for(address)
+    address_1 = address["address_1"]
+    address_2 = address["address_2"] ? " " + address["address_2"] : ""
+    address_3 = address["address_3"] ? " " + address["address_3"] : ""
+
+    "#{address_1}#{address_2}#{address_3} #{address['city']}, #{address['state']}"
```
mdbenjam: I think I prefer storing these as different fields in the DB since that gives us more flexibility later. For example what if we want to show the address on multiple lines? Or just the city? But leave it up to you.

spec/jobs/fetch_hearing_locations_for_veterans_job_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+        create(:veteran, file_number: "987654321")
+      end
+
+      describe "#missing_veteran_file_numbers" do
+        it "returns list of file_numbers with no associated veteran" do
+          expect(job.missing_veteran_file_numbers).to match_array [bfcorlid_file_number]
+        end
+      end
+    end
+  end
```
mdbenjam: Lastly might be good to test this function:

```
  def veterans
    @veterans ||= Veteran.where(file_number: file_numbers)
      .left_outer_joins(:available_hearing_locations)
      .where("available_hearing_locations.updated_at < ? OR available_hearing_locations.id IS NULL", 1.month.ago)
      .limit(500)
  end
```
Especially to ensure that it gets veterans that both have no hearing locations and who have old hearing locations.

lib/fakes/va_dot_gov_service.rb
lomaxap: oh, ok. This is inheriting from ExternalApi::VADotGovService, so it's just overriding the `send_va_dot_gov_request` method. `geocode` and `get_distance` methods are still defined

mdbenjam: Ah I see, I hadn't noticed that. Then I think this works fine then!

PR #8429, opened 2018-12-24
[WIP] Comment deep links

### LINKED ISSUE #8396: Reader deep link to comment
The IDT wants users to be able to link directly to a comment in Reader. We need to add a button that generates a deep link to that comment. Then when that deep link is hit from the browser, we need to scroll the user to the relevant comment. We should use the comment's database id as the query param value, so that the IDT can link to a comment with just knowing the comment's id.

# AC
- [ ] Add a `share` button next to the `edit` and `delete` buttons which opens a modal with the deep link to that comment.
- [ ] When navigating to the deep linked comment, the page should scroll to the comment after loading.

hschallhorn:
Resolves #8396

### Description
Adds the ability to access and follow a link to a comment with only the comment's database id.

### Acceptance Criteria
- [x] Add a share button next to the edit and delete buttons which opens a modal with the deep link to that comment.
- [x] Link created opens document containing the annotation
- [x] When navigating to the deep linked comment, the page should scroll to the comment after loading.

### Testing Plan
1. Go to any reader document longer than one page
2. Create a comment on any page other than the first page
3. Click share next to created comment
4. Click the link in the modal to copy it to your clipboard
5. Follow the copied link and confirm the newly created comment is shown.


client/app/reader/PdfViewer.jsx
```diff
@@ -22,6 +27,10 @@ import { CATEGORIES, ACTION_NAMES, INTERACTION_TYPES } from './analytics';
 
 const NUMBER_OF_DIRECTIONS = 4;
 
+const copyTextButtonStyling = css({
```
mdbenjam: Can we use overflow-wrap instead? Seems to preserve short words. And then let's just add this as a style directly in the copy button. I'm trying to have us avoid adding optional styles to components.

https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

spec/feature/reader/reader_spec.rb
```diff
@@ -761,6 +772,20 @@ def placing_annotation_icon_position
...
+      scenario "Follow comment deep link" do
+        annotation = documents[1].annotations[0]
+        visit "/reader/appeal/#{appeal.vacols_id}/documents/#{documents[1].id}?annotation=#{annotation.id}"
+
+        expect(page).to have_content(annotation.comment)
+        expect(page).to have_css(".page")
+        expect_in_viewport("commentIcon-container-#{annotation.id}")
+      end
+
+      scenario "404 on bad comment id" do
```
mdbenjam: I don't think we need this test anymore. It looks like we got rid of this route.

client/app/reader/PdfViewer.jsx
hschallhorn: 👍

Issue #8329, opened 2018-12-18
Intake | Only show decision issues made before receipt date


shanear:
On intake and edit, we are showing decision issues that were made before the receipt date. It's impossible for them to have contested a decision that happened in the future.

## Acceptance Criteria
- On the intake or the intake edit page, the only decision issues that should show up as contestable are those where the decision date (or approximated decision date) is before the receipt date of the decision review being intook.

## Notes
Background story https://github.com/department-of-veterans-affairs/caseflow/issues/7978

steve-gov:
I created an HLR with an intake date earlier than the displayed decision dates and at least one rated-issue effective date:

![screencapture](https://user-images.githubusercontent.com/25468689/50658532-8aad7500-0f4e-11e9-9db0-d5c3ee1ce8f2.gif)

Shouldn't I NOT be able to select any of the issues with a decision date > 05/10/2018 in this case? I may be misunderstanding this ticket.

steve-gov:
Confirmed that this is not the desired behavior. In this example, I should not be able to select among issues decided on 10/10/2018 if my receipt date is 5/10/2018.

leikkisa:
@steve-gov Sorry about that, the spec said "decision issues" which is a specific thing, but I will make a change to apply to rating issues as well.

steve-gov:
Validated: Confirmed that I was not able to select among issues decided in the future.

PR #8432, opened 2018-12-25
Do not show decision issues from future in contestable issues

leikkisa:
connects #8329


app/models/appeal.rb
```diff
@@ -274,6 +274,7 @@ def establish!
   def contestable_decision_issues
     DecisionIssue.where(participant_id: veteran.participant_id)
       .where.not(decision_review_type: "Appeal")
+      .select { |issue| issue.approx_decision_date ? issue.approx_decision_date < receipt_date : false }
```
pkarman: I think the ternary within the select is a little duplicative.

`issue.approx_decision_date && issue.approx_decision_date < receipt_date` would do the same thing, I think.

leikkisa: 👍

app/models/appeal.rb
```diff
@@ -274,6 +274,7 @@ def establish!
   def contestable_decision_issues
     DecisionIssue.where(participant_id: veteran.participant_id)
       .where.not(decision_review_type: "Appeal")
+      .select { |issue| issue.approx_decision_date && issue.approx_decision_date < receipt_date }
```
shanear: it's worth noting that this is not a database filter at this point. I don't think that should matter too much since the result should always be low

app/models/claim_review.rb
```diff
@@ -118,6 +118,7 @@ def find_request_issue_by_description(description)
   def contestable_decision_issues
     DecisionIssue.where(participant_id: veteran.participant_id, benefit_type: benefit_type)
       .where.not(decision_review_type: "Appeal")
+      .select { |issue| issue.approx_decision_date ? issue.approx_decision_date < receipt_date : false }
```
shanear: also I can't find a test for this. It looks like we updated some tests to continue to work, but i feel like there should be a unit test testing that decisions in the future aren't returned

app/models/appeal.rb
leikkisa: Noted!

app/models/claim_review.rb
leikkisa: Ok! I had started that, but it told me it was a private method, and since it wasn't tested before, I ended up deciding to test it in one of the feature tests.  But I added two model tests now.

shanear: oops i should have been more explicit here. We don't want to be testing private methods, but there is a public method we can test this private method through. it looks like `ClaimReview#contestable_issues` should work

leikkisa: Ok I switched it! I just tested this one part of it, but I moved it to contestable_issues

PR #8433, opened 2018-12-25
example of backup error messages for users

kencheeto:
### Description
We've had a couple instances where obscured errors make it impossible for Intake users to know what went wrong (transient? bad data?), leading them to being blocked for up to a day until the intake expires. Additionally, if the specific veteran has an issue where any intake will fail, the Intake user might simply run into the same problem again.

While a better outcome would be to avoid these errors altogether or fix/workaround them automatically, a good first step is to give users useful information if we can. This PR illustrates how we can use Rails flash messages to do this:

* Capture specific errors and set `flash[:error]` in the controller
* Pass the flash error to the Error redux app
* Show the error when we render the 500 page

Again, this pattern is not intended to replace proper error messages integrated into the UI, but as a cheap fallback. The screenshot below shows what renders if an unhandled error is raised with a message:
screen shot 2018-12-24 at 17 28 36

### DIFF
--- a/app/controllers/claim_review_controller.rb
+++ b/app/controllers/claim_review_controller.rb
@@ -1,5 +1,13 @@
 class ClaimReviewController < ApplicationController
   before_action :verify_access, :react_routed, :verify_feature_enabled, :set_application
+  helper_method :props
+
+  def edit
+    props
+  rescue StandardError => e
+    Raven.capture_exception(e)
+    flash[:error] = e.message
+  end
 
   def update
     if request_issues_update.perform!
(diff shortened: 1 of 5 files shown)


app/controllers/claim_review_controller.rb
```diff
@@ -26,6 +48,7 @@ def request_issues_update
   end
 
   def claim_review
+    raise StandardError.new("oh no")
```
kencheeto: illustrative; see https://github.com/department-of-veterans-affairs/caseflow/pull/8057 for examples of errors we could handle more nicely

app/controllers/claim_review_controller.rb
```diff
@@ -1,5 +1,13 @@
 class ClaimReviewController < ApplicationController
   before_action :verify_access, :react_routed, :verify_feature_enabled, :set_application
+  helper_method :props
+
+  def edit
+    props
```
kencheeto: calling this here so we can rescue the error (previously some errors would occur when rendering the template)

config/environments/development.rb
```diff
@@ -10,7 +10,7 @@
   config.eager_load = true
 
   # Show full error reports and disable caching.
-  config.consider_all_requests_local       = true
+  config.consider_all_requests_local       = false
```
kencheeto: need to set this locally so you can see the same 500 error page that users do, otherwise you'll see a more detailed Rails error page

PR #8441, opened 2018-12-26
Move condition to decide if we show/hide task action dropdown to backend

lowellrex:
This PR moves logic onto the backend now that we have a robust model for determining available actions for each task. When this PR was created we were still using the [combination of individual, attorney, and organization tasks](https://github.com/department-of-veterans-affairs/caseflow/pull/7275/files#diff-5b5c7473c1568400a8104eccd0693108R241) in order to determine whether to show the task actions dropdown. Now that we are using the `Task. available_actions()` and the `actionableTasksForAppeal()` selector, we can move a bunch of this code to the backend.

Before | After
--- | ---
image | image

### DIFF
--- a/app/models/legacy_tasks/attorney_legacy_task.rb
+++ b/app/models/legacy_tasks/attorney_legacy_task.rb
@@ -1,6 +1,6 @@
 class AttorneyLegacyTask < LegacyTask
   def available_actions(role)
-    return [] if role != "attorney"
+    return [] if role != "attorney" || !task_id
 
     actions = [
       {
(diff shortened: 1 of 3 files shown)


app/models/legacy_tasks/attorney_legacy_task.rb
```diff
@@ -1,6 +1,6 @@
 class AttorneyLegacyTask < LegacyTask
   def available_actions(role)
-    return [] if role != "attorney"
+    return [] if role != "attorney" || !task_id
```
lowellrex: This change results in not showing any information about this task in the task snapshot on the case details page, but since this task is not actionable to the current user then I think it is okay (I also think there will not be any information associated with this task).

aroltsch: Could you please add a comment to the code explaining the logic.

Issue #5695, opened 2018-05-29
AMA intake | Claimant is not veteran and veteran has no relationships


abbyraskinUSDS:
Right now, if the veteran has no relationships, and the claimant is not the veteran, we show this:

<img width="1139" alt="screen shot 2018-10-03 at 1 12 10 pm" src="https://user-images.githubusercontent.com/1596259/46427053-48d22000-c70e-11e8-9d06-8ed31348b350.png">

We prompt the user for a payee code, but don't allow them to select the claimant. We need to update this to properly handle the situation.

## Acceptance Criteria:
- When intaking any decision review, and "Is the claimant someone other than the Veteran?" is set, and the veteran has no dependents (`veteran.relationships`) in CorpDB/BGS:
  - Don't show the payee code prompt
  - Disable the "Next" button
  - Show the error message below the radio button according to the comps, "There are no Claimants in our records. You can add the claimant in VBMS and refresh this page."

## Copy
There are no Claimants in our records. You can add the claimant in VBMS and refresh this page.

## Mocks
![REVIEW: no relationship alert.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5b91284c50ee5a0014e9de9d/125516c66e82c728ace21e0d46dab1ac0aa7b4f3cfe7b003b658fc953d4f287d3a0caa3b3084ef2dad56394d4d1f0cf31b0d0c00eef39b2eaeb46d72dc5902efd86d0970c66c16acb8d01dba01b972995397f4edd4189fb5646fb2c99d452c1f0d1288fa5934a89848c95ff64ae84f47e9.png)

## Related
#5569 (Adds "other claimant" question)

abbyraskinUSDS:
- Sally is holding off on developing any validations for now
- Assigning to Leah to figure out how AMO wants CAs to handle this scenario

leikkisa:
Potential plan:
If there are no relationships, keep the "yes/no" radio button, and if they select "yes", show a message that no relationships have been found and that they need to add them and refresh (very similar to current message).
If they select "yes", the "continue" button is greyed out until a relationship is selected

shanear:
@carolaponce need thoughts or designs on what alert should happen on this screen when the veteran has no relationships recorded

carolaponce:
Introducing a style from style-guide I don't think we are using, but I see applicable in this case.
I don't want to use red text because it is not an error - It is an alert.

Proposed text: 
[...]

PR #8442, opened 2018-12-26
When there are no relationships and claimant is not Veteran

leikkisa:
connects #5695


leikkisa:
screen shot 2018-12-26 at 4 39 53 pm

spec/feature/intake/review_page_spec.rb
```diff
@@ -0,0 +1,190 @@
+require "rails_helper"
+require "support/intake_helpers"
+
+RSpec.feature "Intake Review Page" do
```
pkarman: fwiw I don't think the `RSpec` is required here since the helpers import all the methods, but I have seen it in lots of our tests anyway.

spec/feature/intake/review_page_spec.rb
```diff
@@ -0,0 +1,190 @@
+require "rails_helper"
```
pkarman: I think this `require` is redundant.

spec/support/intake_helpers.rb
```diff
@@ -103,6 +108,24 @@ def start_appeal(
   # rubocop: enable Metrics/MethodLength
   # rubocop: enable Metrics/ParameterLists
 
+  def setup_intake_flags
+    FeatureToggle.enable!(:intake)
+    FeatureToggle.enable!(:intakeAma)
+    FeatureToggle.enable!(:intake_legacy_opt_in)
+
+    Time.zone = "America/New_York"
+    Timecop.freeze(Time.utc(2018, 5, 26))
```
pkarman: if the specific date doesn't matter, can we use `Time.zone.today` -- otherwise, let's start naming these days so they are meaningful.

PR #8445, opened 2018-12-27
Consolidate Queue classes

lowellrex:
`ColocatedQueue`, `GenericQueue`, and `JudgeQueue` all scratch the same itch in almost identical ways. This PR consolidates those three classes into one as a first step towards hopefully doing the same with `LegacyWorkQueue` and `AttorneyQueue`.


app/models/queues/generic_queue.rb
```diff
@@ -3,17 +3,23 @@ class GenericQueue
 
   attr_accessor :user
 
+  # rubocop:disable Style/SymbolProc
   def tasks
-    relevant_tasks.where(assigned_to: user)
-  end
-
-  def tasks_by_appeal_id(appeal_id, appeal_type)
-    relevant_tasks.where(appeal_id: appeal_id, appeal_type: appeal_type)
```
lowellrex: We never called this function.

app/models/task.rb
```diff
@@ -233,6 +233,15 @@ def timeline_details
...
   end
 
+  def update_if_hold_expired!
+    update!(status: Constants.TASK_STATUSES.in_progress) if on_hold_expired?
+  end
+
+  def on_hold_expired?
+    return true if placed_on_hold_at && on_hold_duration && placed_on_hold_at + on_hold_duration.days < Time.zone.now
+    false
+  end
```
lowellrex: Moved these two functions from the `ColocatedTask` subclass up into the base `Task` class.

spec/controllers/tasks_controller_spec.rb
```diff
@@ -16,36 +16,19 @@
...
-    let!(:task15) do
-      create(:ama_judge_task, :completed, assigned_to: user, assigned_by: user, completed_at: 3.weeks.ago)
-    end
-
-    let!(:task11) { create(:ama_attorney_task, assigned_to: user) }
-    let!(:task12) { create(:ama_attorney_task, :in_progress, assigned_to: user) }
-    let!(:task13) { create(:ama_attorney_task, :completed, assigned_to: user) }
-    let!(:task16) { create(:ama_attorney_task, :completed, assigned_to: user, completed_at: 3.weeks.ago) }
-    let!(:task14) { create(:ama_attorney_task, :on_hold, assigned_to: user) }
-
```
lowellrex: Every single one of these tasks was being created for each test that was run. With this move from `ColocatedQueue` and `JudgeQueue` to `GenericQueue` we no longer scope the returned set of tasks to only `ColocatedTask` and `JudgeTask`s respectively. As a result we were returning attorney and colocated tasks when we expected to return only the type of task that related to the specific queue. Returning all relevant tasks (tasks assigned to the current user or, in the case of judges, cases assigned to an attorney on their team) regardless of type is the correct behaviour. I've updated the test here to only create the tasks within the tests that actually test for them.

app/models/queues/generic_queue.rb
aroltsch: Perhaps it was removed from the controller at some point.

app/models/queues/generic_queue.rb
```diff
@@ -3,17 +3,23 @@ class GenericQueue
...
+    (relevant_tasks + relevant_attorney_tasks).each { |t| t.update_if_hold_expired! }
```
aroltsch: This doesn't work?

```
(relevant_tasks + relevant_attorney_tasks).each(&:update_if_hold_expired!)
```

lowellrex: It works! I thought the `GenericQueue` test would fail but that test doesn't seem bothered by this change at all! Can't remember why I did that in the first place...

PR #8457, opened 2018-12-27
Strip whitespaces document ID

### LINKED ISSUE #8448: [FE] Trim spaces before and after document ID

aroltsch:
Resolves #8448


spec/models/attorney_case_review_spec.rb
```diff
@@ -144,7 +144,7 @@
   context ".complete" do
     let(:document_type) { Constants::APPEAL_DECISION_TYPES["OMO_REQUEST"] }
     let(:work_product) { "OMO - IME" }
-    let(:document_id) { "123456789.1234" }
+    let(:document_id) { "     123456789.1234    " }
```
lowellrex: Might make sense to create two variables here. One that is the real document ID and the second that is the padded version of the document ID so that we don't have to replicate this string everywhere. Maybe something like the following?

```ruby
let(:document_id) { "123456789.1234" }
let(:padded_document_id) { "     #{document_id}    " }
```

aroltsch: Good suggestion

PR #8460, opened 2018-12-27
adds the JudgeQualityReviewTask

tomas-nava:
This is a new subclass for JudgeTasks created by Quality Review.

Also renames `JudgeReviewTask` to `JudgeDecisionReviewTask`

Connects #8107


app/models/tasks/judge_task.rb
```diff
@@ -16,17 +13,6 @@ def timeline_title
...
-    new_task = super(params, user)
-
-    parent = Task.find(params[:parent_id]) if params[:parent_id]
-    if parent && parent.is_a?(QualityReviewTask)
-      parent.update!(status: :on_hold)
-    end
-
-    new_task
-  end
-
```
lowellrex: We can stop overriding this because the logic to update the parent task's status to on hold has since [moved into the base task class](https://github.com/department-of-veterans-affairs/caseflow/blob/426d8afda0b2d14b9d5632ea804ef8bce4ad62d3/app/models/task.rb#L12). Is that correct?

app/models/tasks/judge_task.rb
```diff
@@ -35,19 +21,8 @@ def self.verify_user_can_create!(user)
...
-    update_status(params.delete("status")) if params.key?("status")
-    update(params)
-
-    [self]
-  end
-
   def when_child_task_completed
-    update!(type: JudgeReviewTask.name)
+    update!(type: JudgeReviewTask.name) if type == JudgeAssignTask.name
```
lowellrex: Let's move this into the `JudgeAssignTask` class so that we don't have to check for the type here.

client/app/queue/AssignToAttorneyModalView.jsx
```diff
@@ -38,7 +38,7 @@ class AssignToAttorneyModalView extends React.PureComponent<Props> {
   ) => {
     const previousAssigneeId = tasks[0].assignedTo.id.toString();
 
-    if (tasks[0].label === 'assign') {
+    if (tasks[0].label === 'assign' || tasks[0].label === 'quality review') {
```
lowellrex: I think we can use [the includes array method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) here.

client/app/queue/selectors.js
```diff
@@ -221,12 +221,12 @@ export const judgeReviewTasksSelector = createSelector(
   [tasksByAssigneeCssIdSelector],
   (tasks) => _.filter(tasks, (task: TaskWithAppeal) => {
     if (task.appealType === 'Appeal') {
-      return task.label === 'review' &&
+      return (task.label === 'review' || task.label === 'quality review') &&
```
lowellrex: Same note about the [includes array method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) here.

app/models/tasks/judge_quality_review_task.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def self.create_from_params(params, user)
+    Task.find(params[:parent_id])&.update!(status: :on_hold)
+    super(params, user)
+  end
+
+  def self.modify_params(params)
+    super(params).merge(type: JudgeQualityReviewTask.name)
+  end
```
lowellrex: I think we can get rid of this if we also get rid of [`JudgeTask.modify_params()`](https://github.com/department-of-veterans-affairs/caseflow/blob/8b435b61e2ea615de233957b43e423584b385227/app/models/tasks/judge_task.rb#L30) that is trying to change the type of `JudgeAssignTask`.

It appears to me that `JudgeTask.modify_params()` only gets called by the following functions (in reverse order):
* `Task.create_from_params()`
* `Task.create_many_from_params()`
* `TasksController.create()`

I am pretty sure the object type that will be calling `create_many_from_params()` in the `TasksController` will always be either `JudgeAssignTask` or `JudgeQualityReviewTask`, so we can get rid of `JudgeTask.modify_params()`. I think this is an artifact when we used the `action` field to change `JudgeTask` behaviour and set that value when we created the task record. Since we are now relying only on the task type. Let's trash this! Please challenge me on this one because I might be wrong.

app/models/tasks/judge_quality_review_task.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def update_from_params(params, _current_user)
+    params[:instructions] = [instructions, params[:instructions]].flatten if params.key?(:instructions)
+
+    update_status(params.delete(:status)) if params.key?(:status)
+    update(params)
+
+    [self]
+  end
```
lowellrex: Why do we need to override this function here? Can we not just rely on the base `Task. update_from_params()` method?

app/models/tasks/judge_task.rb
```diff
@@ -2,10 +2,7 @@ class JudgeTask < Task
   include RoundRobinAssigner
 
   def available_actions(_user)
-    actions = baseline_actions
-    actions << Constants.TASK_ACTIONS.MARK_COMPLETE.to_h if parent && parent.is_a?(QualityReviewTask)
-
-    actions
+    baseline_actions
```
lowellrex: Two thoughts:
1. Let's have this return an empty array and offload the `available_actions` method definitions onto the subclasses (same thought as [this comment](https://github.com/department-of-veterans-affairs/caseflow/pull/8460/files#r244234314)).
2. This change will remove some available actions for existing judge tasks, but I think that is okay. Let's just plan to babysit the deploy and update the tasks that have been affected after this change goes out.

Currently it only affects 6 tasks and the change would only result in an option not appearing in the task actions dropdown box. Not a terrible failure mode.
```ruby
rails c> impacted_tasks = Task.where(type: [JudgeTask.name, JudgeAssignTask.name, JudgeReviewTask.name]).where.not(status: "completed").select { |t| t.parent && t.parent.is_a?(QualityReviewTask) }
rails c> impacted_tasks.count
# 6

# And then the migration we will run to update the tasks to be the right type will look something like:
old_qr_judge_tasks = Task.where(type: [JudgeTask.name, JudgeAssignTask.name, JudgeReviewTask.name]).select { |t| t.parent && t.parent.is_a?(QualityReviewTask) }
rails c> old_qr_judge_tasks.count
# 13
rails c> old_qr_judge_tasks.each { |t| t.update!(type: JudgeQualityReviewTask.name) }
```

app/models/tasks/judge_quality_review_task.rb
tomas-nava: It looks like the base method on `Task` does everything except [this line](https://github.com/department-of-veterans-affairs/caseflow/pull/8460/files/8799fbe8e1a2197204cdba75a331aa70c35b464b#diff-3dac5160eca78d80ec3d56fe03c44413R18), which looks like it's a guard against setting a nil status?

I pulled it down from `JudgeTask` where it was added with [this commit](https://github.com/department-of-veterans-affairs/caseflow/commit/6ba7f655d3e1f3e476d9a32c3c5b726a873c8f60#diff-70ac1e271f4909395fdbf16693720a3cR45)

app/models/tasks/judge_quality_review_task.rb
tomas-nava: I think you're right...verifying and will report back.

tomas-nava: Yeah, I'm confident you're right. Resolving this.

app/models/tasks/judge_quality_review_task.rb
lowellrex: When I added this PR we were cascading completed task information back up to the parent by way of the `mark_task_complete` function (via `update_status`) instead of relying on an `after_update` hook like we do now (after this behaviour was restored [3 weeks ago](https://github.com/department-of-veterans-affairs/caseflow/pull/8164)). So I think the original motivation for overriding this method in the subclass no longer exists.

As for protecting against `nil` values in the status field, I don't think we ever expect to receive those values from the frontend (`git grep status client/app/queue/`) and if we do then we should probably protect against `nil` status values for all tasks (which I don't think we need to do yet).

tomas-nava:
I've added the change from `JudgeReviewTask` to `JudgeDecisionReviewTask`; here's the script we can run to update existing tasks in production:

```ruby
Task.find_by(type: 'JudgeReviewTask').each do |task|
  puts "changing JudgeReviewTask #{task.id} into a JudgeDecisionReviewTask"
  task.update!(type: JudgeDecisionReviewTask.name)
end
```

I could also put this into a migration if that's preferable.

lowellrex:
Ran the following commands in the rails console to change `JudgeReviewTask`s to `JudgeDecisionReviewTask`s (see [related slack discussion](https://dsva.slack.com/archives/C6E41RE92/p1546531054021300)):
```ruby
# Had to create a dummy class since we removed the existing JudgeReviewTask class in this PR
rails c> class JudgeReviewTask < Task; end

Task.where(type: 'JudgeReviewTask').each do |task|
  puts "changing JudgeReviewTask #{task.id} into a JudgeDecisionReviewTask"
  task.update!(type: JudgeDecisionReviewTask.name)
end
```

Ran the following commands in the rails console to change the type of judge tasks that had quality review task parents:
```ruby
rails c> msgs = []
rails c> JudgeTask.select { |t| t.parent.is_a?(QualityReviewTask) }.each do |task|
  msgs << "changing #{task.type} #{task.id} into a JudgeQualityReviewTask"
  task.update!(type: JudgeQualityReviewTask.name)
end
rails c> msgs
=> ["changing JudgeDecisionReviewTask 1183 into a JudgeQualityReviewTask", "changing JudgeAssignTask 1844 into a JudgeQualityReviewTask", "changing JudgeAssignTask 1847 into a JudgeQualityReviewTask", "changing JudgeDecisionReviewTask 1839 into a JudgeQualityReviewTask", "changing JudgeAssignTask 2014 into a JudgeQualityReviewTask", "changing JudgeDecisionReviewTask 1625 into a JudgeQualityReviewTask", "changing JudgeDecisionReviewTask 4893 into a JudgeQualityReviewTask", "changing JudgeAssignTask 5176 into a JudgeQualityReviewTask", "changing JudgeAssignTask 5724 into a JudgeQualityReviewTask", "changing JudgeAssignTask 9126 into a JudgeQualityReviewTask", "changing JudgeAssignTask 9128 into a JudgeQualityReviewTask", "changing JudgeAssignTask 7953 into a JudgeQualityReviewTask", "changing JudgeDecisionReviewTask 10361 into a JudgeQualityReviewTask", "changing JudgeAssignTask 12810 into a JudgeQualityReviewTask", "changing JudgeAssignTask 12973 into a JudgeQualityReviewTask", "changing JudgeAssignTask 13018 into a JudgeQualityReviewTask"]
```

Issue #7623, opened 2018-10-30
AMA Edit | Edit appeal copy changes


shanear:
There should be some copy differences when editing the issues on an appeal. These will help the edit flow cater better to attorneys who will be using it.

## Acceptance Criteria
When editing any decision review:
- The call to action buttons on the bottom should read "Save" and "Cancel"

When editing the issues on a Board Appeal:
- The "Form being processed" row should be hidden
- "Receipt date on this form" should instead say "NOD receipt date"
- The copy on the "Remove Issue" modal should be updated to reflect copy below (see mocks below)

## Notes
this was created because of requests by @allyceh on this issue https://github.com/department-of-veterans-affairs/caseflow/issues/7594

## Copy
The issue you selected will be removed from the list of issues on appeal.

Are you sure that this issue is not listed on the veteran's NOD and that you want to remove it?

## Mocks
![image](https://user-images.githubusercontent.com/40216950/47751418-14e40f00-dc68-11e8-880b-55c8f7fd6625.png)
![image](https://user-images.githubusercontent.com/40216950/47751646-c1be8c00-dc68-11e8-8f88-d4b380a00921.png)

allyceh:
A second set of eyes would be great before we're ready for dev. @laurjpeterson or @sneha-pai , can you take a look? Sneha, if you need more context, happy to walk you through things, too.

laurjpeterson:
looks good to me!

allyceh:
Great. I’d say we’re good to start dev.

carolaponce:
@allyceh  and I reviewed her UI/copy changes today and the needs of her users. There are a few things we want to clarify:

1. Based on the user (clerk, attorney, etc) a different version of this edit issues screens would appear.  We will have variations of copy based on the user. Correct? @shanear 

2. Issues that are not valid or duplicates will not show up in the case details page of Queue, correct?

3. Whenever there is an invalid issue, or any issue error, the Intake team (Ivy's team) is responsible for notifying the Veteran.  Does anybody know this part of the process? If not I can check with Ivy.

**Next steps:**

- Clarify questions above.
- I will review our error messages and send @allyceh  feedback.

carolaponce:
My comments in #7594

carolaponce:
@allyceh for our review today:
![issues-sumary-exploration error copy 2](https://user-images.githubusercontent.com/11700545/48430620-b88bef80-e73d-11e8-9e94-5e50a88ffcf0.png)
[...]

PR #8463, opened 2018-12-27
Edit appeal copy updates

leikkisa:
connects #7623


leikkisa:
Do we need feature tests for this, since it's just copy changes?

client/app/intake/util/issues.js
```diff
@@ -265,10 +266,18 @@ export const formatIssues = (state) => {
 
 export const getAddIssuesFields = (formType, veteran, intakeData) => {
```
pkarman: nice refactor here!

PR #8464, opened 2018-12-28
Simplify LegacyWorkQueue

lowellrex:
`LegacyWorkQueue` exists to fetch tasks from VACOLS associated with `LegacyAppeal`s. This PR attempts to make that behaviour clearer by:
1. Renaming its functions more precisely
2. Ignoring the current user's role when possible
3. Fetching the user from the VACOLS task assignee and not the other way around


app/controllers/tasks_controller.rb
```diff
@@ -89,11 +89,8 @@ def for_appeal
       return json_vso_tasks
     end
 
-    tasks = appeal.tasks
-    if %w[attorney judge].include?(user_role) && appeal.is_a?(LegacyAppeal)
-      legacy_appeal_tasks = LegacyWorkQueue.tasks_with_appeals_by_appeal_id(params[:appeal_id], user_role)
-      tasks = (legacy_appeal_tasks + tasks).uniq
-    end
+    legacy_appeal_tasks = appeal.is_a?(LegacyAppeal) ? LegacyWorkQueue.tasks_by_appeal_id(appeal.vacols_id) : []
```
lowellrex: Always fetch legacy tasks for legacy appeals regardless of the current user's primary role in VACOLS.

app/models/queues/legacy_work_queue.rb
```diff
@@ -1,21 +1,14 @@
-# Called "WorkQueue" instead of "Queue" to not conflict with the
-# "Queue" class that ships with Ruby.
 class LegacyWorkQueue
   include ActiveModel::Model
   class << self
-    def tasks_with_appeals(user, role)
+    def tasks_for_user(user)
```
lowellrex: Renamed because we no longer return appeals.

app/models/queues/legacy_work_queue.rb
```diff
@@ -24,16 +17,19 @@ def repository
...
-    def tasks_with_appeals_of_vacols_tasks(user, role, vacols_tasks)
       vacols_appeals = repository.appeals_by_vacols_ids(vacols_tasks.map(&:vacols_id))
 
       vacols_tasks.zip(vacols_appeals).map do |task, appeal|
-        MODEL_CLASS_OF_ROLE[role.capitalize].from_vacols(task, appeal, user)
+        user = if task.assigned_to_css_id
+                 User.find_or_create_by(css_id: task.assigned_to_css_id, station_id: User::BOARD_STATION_ID)
+               end
+
+        task_class = (user&.vacols_roles&.first&.downcase == "judge") ? JudgeLegacyTask : AttorneyLegacyTask
```
lowellrex: If user associated with the task does not have a role in VACOLS default to attorney like we [previously in the IDT controller](https://github.com/department-of-veterans-affairs/caseflow/pull/8464/files#diff-145e4f8c3deece57cdc484c1cc4bd5caL61).

app/models/queues/legacy_work_queue.rb
```diff
@@ -24,16 +17,19 @@ def repository
...
-    }.freeze
+    def tasks_from_vacols_tasks(vacols_tasks)
+      return [] if vacols_tasks.empty?
 
-    def tasks_with_appeals_of_vacols_tasks(user, role, vacols_tasks)
       vacols_appeals = repository.appeals_by_vacols_ids(vacols_tasks.map(&:vacols_id))
 
       vacols_tasks.zip(vacols_appeals).map do |task, appeal|
-        MODEL_CLASS_OF_ROLE[role.capitalize].from_vacols(task, appeal, user)
+        user = if task.assigned_to_css_id
```
lowellrex: Previously we passed in the user argument. Now we fetch the user from the brieff table (current location). I think this is more precise especially when we are fetching tasks for an appeal and not tasks for a user.

app/controllers/tasks_controller.rb
lowellrex: Fair point! Undoing this now.

PR #8468, opened 2018-12-28
non-comp SC and HLR create DecisionReviewTask on complete!

pkarman:
connects #7990 

### Description
Create DecisionReviewTask when intake completes for non-comp HLR and SC.


app/models/claim_review.rb
```diff
@@ -48,6 +48,12 @@ def create_issues!(new_issues)
     end
   end
 
+  def create_non_comp_task!
+    return if tasks.any? { |task| task.is_a?(DecisionReviewTask) } # TODO: more specific check?
+    # TODO: better user?
```
pkarman: not sure who should own this task, or if it matters at this point.

app/models/claim_review.rb
```diff
@@ -48,6 +48,12 @@ def create_issues!(new_issues)
     end
   end
 
+  def create_non_comp_task!
+    return if tasks.any? { |task| task.is_a?(DecisionReviewTask) } # TODO: more specific check?
```
leikkisa: This is just in case someone completes a claim review twice?

app/models/claim_review.rb
leikkisa: Makes sense. Do we even need to assign it to anyone yet?

pkarman: apparently `assigned_to` is required to pass validation.

PR #8471, opened 2018-12-28
More test coverage for attorney checkout

### LINKED ISSUE #8268: [BE] Capybara tests for legacy and ama checkout
Ensure capybara tests are written to test multiple issue dispositions and remand reasons for legacy/ama checkout and that the updates are propagated to the DB

aroltsch:
Resolves #8268

### DIFF
--- a/spec/feature/queue/attorney_checkout_flow_spec.rb
+++ b/spec/feature/queue/attorney_checkout_flow_spec.rb
@@ -480,7 +480,6 @@ RSpec.feature "Attorney checkout flow" do
         expect(appeal.issues.second.remand_reasons.size).to eq 2
         expect(appeal.issues.third.remand_reasons.size).to eq 0
 
-
         expect(VACOLS::CaseIssue.where(isskey: appeal.vacols_id)[0].issdc).to eq "3"
         expect(VACOLS::CaseIssue.where(isskey: appeal.vacols_id)[1].issdc).to eq "3"
         expect(VACOLS::CaseIssue.where(isskey: appeal.vacols_id)[2].issdc).to eq "1"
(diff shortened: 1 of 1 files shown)


spec/feature/queue/attorney_checkout_flow_spec.rb
```diff
@@ -65,12 +65,30 @@
...
 
       click_on "Continue"
+
+      find_field("Service treatment records", visible: false).sibling("label").click
+      find_field("Post AOJ", visible: false).sibling("label").click
+
+      click_on "Continue"
+      # For some reason clicking too quickly on the next remand reason breaks the test.
+      # Adding sleeps is bad... but I'm not sure how else to get this to work.
+      sleep 1
```
aroltsch: I've been struggling with this all morning. For some reason nothing works except putting a sleep statement.

lowellrex: Does adding a call to `expect` not solve the problem? Something like `expect(page).to have_content("Medical examinations")` instead of the sleep?

aroltsch: It won't work because if there are two issues with remanded disposition, the page would have "Medical examinations" for the first issue. It fails for the second issue.

lowellrex: Thought here being that `all` does not wait for any elements to be visible where `expect` (or `find`) would wait. https://www.rubydoc.info/github/jnicklas/capybara/master/Capybara%2FNode%2FFinders:all

aroltsch: I thought that `all` does wait. How would you write this test with `find`? `find` usually works when you have only one single element on the page. The problem that we are facing here is that there are multiple matches.

lowellrex: I don't think `all` waits (because it doesn't know how many elements to wait for). We could find the elements we want using something like `page.find(:xpath, "//label[@for='1-medical_examinations']")` I think.

aroltsch: Continuing this discussion on slack

PR #8477, opened 2018-12-28
Use named and relative dates rather than hardcoded strings

pkarman:
Refactor tests to use named and relative dates rather than hardcoded strings. This should make them easier to read and maintain since it should provide more meaningful names and make it easier to shift specs as dates move.

### DIFF
--- a/spec/feature/intake/appeal_spec.rb
+++ b/spec/feature/intake/appeal_spec.rb
@@ -193,7 +193,7 @@ feature "Appeal Intake" do
 
     expect(appeal.request_issues.count).to eq 2
 
-    rating_request_issue = appeal.request_issues.find(&:ri.rating_issue_reference_id)
+    rating_request_issue = appeal.request_issues.find(&:rating_issue_reference_id)
     nonrating_request_issue = appeal.request_issues.find { |ri| ri.rating_issue_reference_id.nil? }
 
     expect(rating_request_issue).to have_attributes(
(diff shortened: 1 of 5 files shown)


spec/support/date_time_helper.rb
```diff
@@ -0,0 +1,21 @@
+module DateTimeHelper
+  def post_ama_start_date
+    ama_start_date + 30.days
+  end
+
+  def ama_start_date
+    Time.new(2019, 2, 14).in_time_zone
```
leikkisa: Should these be dates instead of times?

spec/feature/intake/appeal_spec.rb
```diff
@@ -41,11 +40,15 @@
...
+  let(:future_date) { Time.zone.now + 30.days }
+
+  let(:receipt_date) { post_ramp_start_date - 30.days }
 
   let(:untimely_days) { 372.days }
 
-  let(:profile_date) { Time.zone.local(2018, 9, 15) }
+  let(:profile_date) { post_ramp_start_date - 35.days }
+
+  let(:untimely_date) { receipt_date - 373.days }
```
leikkisa: I guess this could use the untimely_days variable.

spec/support/date_time_helper.rb
pkarman: The existing pattern seems to be using `Time` objects everywhere instead of `Date` so that the timezone is present.

That's a good point about UTC. The dates are important because of the law. I always assumed that the effective date was midnight ET (i.e. local to DC).

PR #8479, opened 2018-12-31
Show "+ Add a new task" button in active tasks title bar #8216 (PR)

### LINKED ISSUE #8216: Show "+ Add a new task" button in active tasks title bar

rkreyhsig:
Resolves #8216 - Show "+ Add a new task" button in active tasks title bar

| Before | After |
| --- | --- |
|old|new|

| Before | After |
| --- | --- |
|old|new|


spec/feature/queue/task_queue_spec.rb
```diff
@@ -151,8 +151,7 @@
       it "should allow us to assign a mail task to a user" do
         visit "/queue/appeals/#{appeal.uuid}"
 
-        find(".Select-control", text: "Select an action…").click
-        find("div", class: "Select-option", text: Constants.TASK_ACTIONS.CREATE_MAIL_TASK.label).click
+        find("button", text: "+ Add a new task").click
```
rkreyhsig: @lowellrex - This looks like the test I was going to write anyway.. Creating an appeal with a root task and clicking to open the modal and verify functionality.

client/app/queue/TaskSnapshot.jsx
```diff
@@ -237,15 +265,21 @@ export class TaskSnapshot extends React.PureComponent<Props> {
...
 
+    /* console.log(this.props.tasks);
+    console.log(this.props.tasks[0].type);
+    console.log(this.props.tasks[0].type !== 'RootTask');
+    console.log(!this.state.addNewTaskIsVisible);*/
+
     if (taskLength) {
       sectionBody = this.props.tasks.map((task, index) =>
-        <tr>
+        task.type !== 'RootTask' && !(rootTask[0] && rootTask[0].availableActions[0]) && <tr key={task.uniqueId}>
```
lowellrex: Let's filter the root task out before we enter this conditional block. That way we will still see the "no active tasks" message if the only task is a root task.

client/app/queue/TaskSnapshot.jsx
```diff
@@ -284,6 +318,12 @@ export class TaskSnapshot extends React.PureComponent<Props> {
     return <div className="usa-grid" {...css({ marginTop: '3rem' })}>
       <h2 {...sectionHeadingStyling}>
         <a id="our-elemnt" {...anchorJumpLinkStyling}>{COPY.TASK_SNAPSHOT_ACTIVE_TASKS_LABEL}</a>
+        { rootTask[0] && rootTask[0].availableActions[0] && <Button
+          linkStyling
+          styling={css({ float: 'right',
+            paddingRight: '10px' })}
+          name={COPY.TASK_SNAPSHOT_ADD_NEW_TASK_LABEL}
+          onClick={this.changeRoute} /> }
```
lowellrex: Let's move the "Add new task" button to its own component and pass in the root task from the `TaskSnapshot` component. This will allow us to better encapsulate the logic around showing/hiding the component, validating the inputs, and the actions it takes when clicked.

client/app/queue/TaskSnapshot.jsx
```diff
@@ -237,15 +265,21 @@ export class TaskSnapshot extends React.PureComponent<Props> {
...
     } = this.props;
 
     let sectionBody = COPY.TASK_SNAPSHOT_NO_ACTIVE_LABEL;
     const taskLength = this.props.tasks.length;
 
+    /* console.log(this.props.tasks);
+    console.log(this.props.tasks[0].type);
+    console.log(this.props.tasks[0].type !== 'RootTask');
+    console.log(!this.state.addNewTaskIsVisible);*/
+
```
lowellrex: Looks like some leftovers here.

client/app/queue/TaskSnapshot.jsx
```diff
@@ -234,6 +240,7 @@ export class TaskSnapshot extends React.PureComponent<Props> {
     return <div className="usa-grid" {...css({ marginTop: '3rem' })}>
       <h2 {...sectionHeadingStyling}>
         <a id="our-elemnt" {...anchorJumpLinkStyling}>{COPY.TASK_SNAPSHOT_ACTIVE_TASKS_LABEL}</a>
+        { rootTask[0] && rootTask[0].availableActions[0] && <AddNewTaskButton rootTask={rootTask[0]} /> }
```
lowellrex: Let's move the control over whether to display this component into the component itself so we can just call `<AddNewTaskButton appealId={appealId}>`.

client/app/queue/TaskSnapshot.jsx
```diff
@@ -198,15 +200,19 @@ export class TaskSnapshot extends React.PureComponent<Props> {
...
+    const tasks = this.props.tasks.filter((task) => {
+      return task.type !== 'RootTask';
+    });
+    const taskLength = tasks.length;
 
     if (taskLength) {
-      sectionBody = this.props.tasks.map((task, index) =>
-        <tr key={task.uniqueId}>
+      sectionBody = tasks.map((task, index) =>
+        task.type !== 'RootTask' && !(rootTask[0] && rootTask[0].availableActions[0]) && <tr key={task.uniqueId}>
```
lowellrex: Let's move this filtering into it's own selector. Maybe something like `nonRootActionableTasksForAppeal` which can do this filtering in the selector file.

client/app/queue/selectors.js
```diff
@@ -173,6 +173,10 @@ export const actionableTasksForAppeal = createSelector(
   [getTasksForAppeal], (tasks: Tasks) => _.filter(tasks, (task) => task.availableActions.length)
 );
 
+export const rootTaskForAppeal = createSelector(
```
lowellrex: Let's rename this to be plural, `rootTasksForAppeal`.

lowellrex: Also, since we only use this in one place so far let's tailor it to our needs and have it only return actionable root tasks.

spec/feature/queue/task_queue_spec.rb
lowellrex: Let's use the constant from `COPY.json` here.

client/app/queue/components/AddNewTaskButton.jsx
```diff
@@ -0,0 +1,44 @@
...
+} from '../QueueActions';
+
+class AddNewTaskButton extends React.PureComponent {
+  changeRoute = () => {
+    const {
+      rootTask,
+      history
+    } = this.props;
+
+    this.props.stageAppeal(rootTask.appealId);
```
lowellrex: We don't change any elements of the appeal here (we only create a new task) so we don't need to stage the appeal.

lowellrex: Same goes for the `resetDecisionOptions` function.

client/app/queue/components/AddNewTaskButton.jsx
```diff
@@ -0,0 +1,44 @@
...
+    const modalPath = rootTask.availableActions[0].value;
+
+    history.push(`/queue/appeals/${appealId}/tasks/${uniqueId}/${modalPath}`);
+  }
+
+  render = () => {
+    return <Button
+      linkStyling
+      styling={css({ float: 'right',
+        paddingRight: '10px' })}
```
lowellrex: Let's move the styling out to an explicit declared variable outside the component.

client/app/queue/TaskSnapshot.jsx
```diff
@@ -252,7 +258,9 @@ const mapStateToProps = (state: State, ownProps: Params) => {
   return {
     appeal: appealWithDetailSelector(state, { appealId: ownProps.appealId }),
     userRole,
-    tasks: actionableTasksForAppeal(state, { appealId: ownProps.appealId })
+    tasks: actionableTasksForAppeal(state, { appealId: ownProps.appealId }),
```
lowellrex: Let's set the `tasks` property here to be `nonRootActionableTasksForAppeal`. That way we don't need to pull in the two properties below (`rootTask` and `nonRootTasks`).

client/app/queue/TaskSnapshot.jsx
```diff
@@ -198,15 +201,17 @@ export class TaskSnapshot extends React.PureComponent<Props> {
...
-    const taskLength = this.props.tasks.length;
+    let sectionBody = <tr><td {...css({ borderStyle: 'none' })}>{COPY.TASK_SNAPSHOT_NO_ACTIVE_LABEL}</td></tr>;
+    const nonRootTasks = this.props.nonRootTasks;
+    const taskLength = nonRootTasks.length;
 
     if (taskLength) {
-      sectionBody = this.props.tasks.map((task, index) =>
-        <tr key={task.uniqueId}>
+      sectionBody = nonRootTasks.map((task, index) =>
+        !(rootTask[0]) && <tr key={task.uniqueId}>
```
lowellrex: I think we can just call `<tr key...` here since our selector filters out the non-root tasks.

client/app/queue/TaskSnapshot.jsx
```diff
@@ -198,15 +201,17 @@ export class TaskSnapshot extends React.PureComponent<Props> {
...
-      appeal
+      appeal,
+      rootTask
     } = this.props;
 
-    let sectionBody = COPY.TASK_SNAPSHOT_NO_ACTIVE_LABEL;
-    const taskLength = this.props.tasks.length;
+    let sectionBody = <tr><td {...css({ borderStyle: 'none' })}>{COPY.TASK_SNAPSHOT_NO_ACTIVE_LABEL}</td></tr>;
+    const nonRootTasks = this.props.nonRootTasks;
+    const taskLength = nonRootTasks.length;
```
lowellrex: I think we can revert all of this since we're using the non-root selector now.

client/app/queue/TaskSnapshot.jsx
```diff
@@ -5,11 +5,14 @@ import { connect } from 'react-redux';
 
 import {
   actionableTasksForAppeal,
-  appealWithDetailSelector
+  appealWithDetailSelector,
+  rootTasksForAppeal,
+  nonRootActionableTasksForAppeal
```
lowellrex: I think we can import only `appealWithDetailSelector` and `nonRootActionableTasksForAppeal` now.

client/app/queue/components/AddNewTaskButton.jsx
```diff
@@ -0,0 +1,52 @@
...
+      styling={buttonStyling}
+      name={COPY.TASK_SNAPSHOT_ADD_NEW_TASK_LABEL}
+      onClick={this.changeRoute} /> : null;
+  }
+}
+
+const mapStateToProps = (state: State, ownProps: Params) => {
+
+  return {
+    rootTask: rootTasksForAppeal(state, { appealId: ownProps.appealId })
```
lowellrex: Let's pull the first element off of the `rootTasksForAppeal` array in the `mapStateToProps` function so that we don't have to do that in the other functions in this class.

client/app/queue/selectors.js
```diff
@@ -173,6 +173,16 @@ export const actionableTasksForAppeal = createSelector(
...
 );
 
+export const rootTasksForAppeal = createSelector(
+  [getTasksForAppeal], (tasks: Tasks) => _.filter(tasks, (task) => task.type === 'RootTask' &&
+    task.availableActions.length)
+);
+
+export const nonRootActionableTasksForAppeal = createSelector(
+  [getTasksForAppeal], (tasks: Tasks) => _.filter(tasks, (task) => task.type !== 'RootTask' &&
+    task.availableActions.length)
```
lowellrex: Let's take advantage of `actionableTasksForAppeal` filtering out tasks that do not have available actions here by doing something like:
```jsx
export const rootTasksForAppeal = createSelector(
  [actionableTasksForAppeal], (tasks: Tasks) => _.filter(tasks, (task) => task.type === 'RootTask')
);

export const nonRootActionableTasksForAppeal = createSelector(
  [actionableTasksForAppeal], (tasks: Tasks) => _.filter(tasks, (task) => task.type !== 'RootTask')
);
```

client/app/queue/TaskSnapshot.jsx
rkreyhsig: @lowellrex If I remove the tr td, we get a console error, and we use taskLength two other places.  I can change line 209 to const tasks = this.props.tasks;

Issue #7344, opened 2018-10-10
Add issue edge cases | Add Issue Landing | if ineligible lane (just Appeal -> HLR and Appeal -> Appeal)


kencheeto:
## Story
When a user adds issues, they may encounter edge cases where the issue is ineligible. We will not indicate that an issue isn't eligible on any of the add issue selection pages but will indicate it on the Add issue landing page and Confirmation Page so they have the information they need to inform the applicant in the letter they send following intake data entry. If an issue is ineligible in will get logged in Caseflow but not VBMS. [Mural showing edge cases](https://app.mural.co/t/workqueue2001/m/workqueue2001/1537791678087/94469b0cdf1a02a7f57a7628702d36db2f079430).

This ticket is focused on the scenario where the issue being requested isn't eligible for that lane and how we display it's ineligibility on the Add Issues Landing page. There are 3 scenarios where this might occur that are detailed in the copy section -- note that they will only occur for HLR and Board Appeal lanes and only for rated issues.

requires https://github.com/department-of-veterans-affairs/appeals-design-research/issues/998

## Acceptance criteria
- If a decision issue that was decided as part of an appeal is contested as part of a higher level review, it should be considered ineligible. This means it should show as ineligible with the copy below, and show up as ineligible on the confirmation page described in #7308  
- If a decision issue that was decided as part of an appeal is contested as part of an appeal, it should be considered ineligible. This means it should show as ineligible with the copy below, and show up as ineligible on the confirmation page

## Notes
- Identical wording will be used on the confirmation page #7308

## Technical note
- Could write these as independent testcases (not mixed with other ineligibility test coverage)

## Copy
Scenario | Display copy 
--- | ---
If the request is a Board Appeal and the same issue was last processed as a Board Appeal | [Issue selected] is ineligible because it was last processed as a Board Appeal and this can't be done twice in a row.
If the request is a HLR and same issue was last processed as a Board Appeal | [Issue selected] is ineligible because it was last processed as a Board Appeal which can't be followed by a Higher-Level Review.

## Mocks
(Sample show what the first scenario would look like)
<img width="759" alt="screen shot 2018-10-06 at 10 43 34 am" src="https://user-images.githubusercontent.com/11219417/46572492-ba031480-c954-11e8-98aa-90cd5e927c86.png">

## Background issues
- #5244 

## How to Validate
(no need, Shane's got this :muscle:)

steve-gov:
[...]

PR #8481, opened 2018-12-31
7344 appeal ineligible lane

leikkisa:
connects #7344


app/models/request_issue.rb
```diff
@@ -297,7 +300,8 @@ def create_decision_issues_from_rating
   end
 
   # RatingIssue is not in db so we pull hash from the serialized_ratings.
-  def fetch_contested_rating_issue_ui_hash
+  def contested_rating_issue_ui_hash
+    return unless review_request.serialized_ratings
```
shanear: Trying to figure out when this would happen. Shouldn't this raise an exception if there was a `rating_issue_reference_id` recorded, and we can't find the `serialized_ratings`?

spec/models/request_issue_spec.rb
```diff
@@ -365,15 +367,98 @@
...
+          let(:review) do
+            create(
+              :appeal,
+              veteran: veteran,
+              legacy_opt_in_approved: legacy_opt_in_approved
+            )
+          end
+
+          it "is not eligible after an appeal" do
+            rating_request_issue.rating_issue_reference_id = higher_level_review_reference_id
```
shanear: this is a candidate for `let` since were doing this in a couple tests

spec/models/request_issue_spec.rb
```diff
@@ -365,15 +367,98 @@
...
+            rating_request_issue.validate_eligibility!
```
shanear: also, i realize this was already written this way so we don't have to fix it, but this should be in `subject`

app/models/request_issue.rb
```diff
@@ -19,6 +19,8 @@ class RequestIssue < ApplicationRecord
     duplicate_of_rating_issue_in_active_review: "duplicate_of_rating_issue_in_active_review",
     untimely: "untimely",
     previous_higher_level_review: "previous_higher_level_review",
+    appeal_to_appeal: "appeal_to_appeal",
+    appeal_to_higher_level_review: "appeal_to_higher_level_review",
```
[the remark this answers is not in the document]
leikkisa: Ok! Yeah the reason I didn't use previous_appeal is because there were two, but I like that suggestion.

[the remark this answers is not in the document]
leikkisa: I think "previous_higher_level_review" might have been a little misleading, because the code that was in there was checking if the previous review was a higher level review no matter what the current review was (so it would also make a supplemental claim issue ineligible if it came from a higher level review). I can put previous in all of them, but it seems unnecessary.

app/models/request_issue.rb
leikkisa: I replaced it with rating_issue_reference_id and then realized what it was there for. serialized_ratings is set up to return nil if there's no receipt date

app/models/request_issue.rb
```diff
@@ -297,35 +298,44 @@ def create_decision_issues_from_rating
...
-    return decision_date if nonrating?
-    return contested_rating_issue.try(:promulgation_date) if rating?
+    if review_request.is_a?(Appeal) && contested_issue.source_review_type == "Appeal"
+      self.ineligible_reason = :appeal_to_appeal
+    end
+
+    self.ineligible_due_to_id = contested_issue.source_request_issue.id if ineligible_reason
   end
+  # rubocop:enable Metrics/CyclomaticComplexity
+  # rubocop:enable Metrics/PerceivedComplexity
```
leikkisa: So I combined the three checks, and it scored high on complexity in the linter, as long as that's ok.  We could also make some sort of matrix that checks the cross of the source review and review request for all options.

spec/feature/intake/edit_spec.rb
```diff
@@ -576,14 +577,6 @@ def verify_decision_issues_can_be_added_and_removed(page_url,
         visit "higher_level_reviews/#{ep_claim_id}/edit"
 
         expect(page).to have_content("8 issues")
-        expect_ineligible_issue(1)
```
leikkisa: These weren't always showing up in the same order

spec/models/request_issue_spec.rb
leikkisa: 👍

spec/models/request_issue_spec.rb
leikkisa: I think the reason it's not is because sometimes it's tested on rating_request_issue and sometimes it's on nonrating_request_issue. I know that can be fixed too but I'm going to leave it for now.

spec/factories/request_issue.rb
```diff
@@ -26,7 +26,7 @@
                rating_issue_reference_id: request_issue.rating_issue_reference_id,
                profile_date: request_issue.rating_issue_profile_date,
                benefit_type: request_issue.review_request.benefit_type,
-               decision_text: "rating decision issue",
+               decision_text: "a rating decision issue",
```
leikkisa: "rating decision issue" looks too much like "nonrating decision issue" to rspec :-P

app/models/appeal.rb
```diff
@@ -278,15 +275,16 @@ def establish!
...
+    root_task && root_task.status == Constants.TASK_STATUSES.completed
+  end
 
-  def contestable_decision_issues
-    return [] unless receipt_date
-    DecisionIssue.where(participant_id: veteran.participant_id)
-      .where.not(decision_review_type: "Appeal")
-      .select { |issue| issue.approx_decision_date && issue.approx_decision_date < receipt_date }
+  def root_task
+    tasks.first.root_task if !tasks.empty?
```
mdbenjam: I'd probably just do `RootTask.where(appeal_id: id)` feels a little cleaner.

app/models/appeal.rb
```diff
@@ -278,15 +275,16 @@ def establish!
     processed!
   end
 
-  private
+  def outcoded?
+    root_task && root_task.status == Constants.TASK_STATUSES.completed
```
mdbenjam: I think this is correct, but @lowellrex for confirmation.

spec/factories/appeal.rb
```diff
@@ -64,6 +64,13 @@
       end
     end
 
+    trait :outcoded do
```
mdbenjam: This feels like fairly unique test setup, I'd ask if it can be in the spec file instead of here. I'm hesitant about adding too many traits.

app/models/appeal.rb
leikkisa: Ok! I'm guessing there is one root task per appeal. That makes more sense to me too, thanks.

app/models/appeal.rb
leikkisa: Confirmed w/ Lowell

PR #8482, opened 2019-01-02
Simplify task status change notification

tomas-nava:
Resolves #8173

### DIFF
--- a/app/models/tasks/generic_task.rb
+++ b/app/models/tasks/generic_task.rb
@@ -59,7 +59,7 @@ class GenericTask < Task
   def reassign(reassign_params, current_user)
     reassign_params[:instructions] = [instructions, reassign_params[:instructions]].flatten
     sibling = self.class.create_child_task(parent, current_user, reassign_params)
-    mark_as_complete!
+    update!(status: Constants.TASK_STATUSES.completed)
 
     children_to_update = children.reject { |t| t.status == Constants.TASK_STATUSES.completed }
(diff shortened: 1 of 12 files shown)


app/models/tasks/judge_task.rb
```diff
@@ -40,7 +40,7 @@ def update_from_params(params, _current_user)
 
     params["instructions"] = [instructions, params["instructions"]].flatten if params.key?("instructions")
 
-    update_status(params.delete("status")) if params.key?("status")
+    update!(status: params.delete("status")) if params.key?("status")
     update(params)
```
tomas-nava: This'll get wiped out when we merge #8460

app/models/tasks/generic_task.rb
```diff
@@ -51,15 +51,15 @@ def update_from_params(params, current_user)
 
     return reassign(params[:reassign], current_user) if params[:reassign]
 
-    update_status(params[:status])
+    update!(params[:status]) if params[:status]
```
lowellrex: Does this work? I would have expected we'd need to include the parameter we are updating here:

```ruby
update!(status: params[:status]) if params[:status]
```

app/models/tasks/judge_task.rb
lowellrex: Can we just remove line 43? As written we are updating the status and then updating the rest of the fields, can we just consolidate that into one `update` (or `update!`) call?

PR #8491, opened 2019-01-02
AMA hearings table

### LINKED ISSUE #8421: Create AMA Hearings Table!
Woo!

sharonwarner:
Resolves #8421 

### Description
This PR adds the new hearings table for AMA hearings! I've also updated the hearings routes to use external_id (vacols_id for legacy hearings, uuid for ama hearings) instead of the table id so that we can differentiate between legacy and ama hearings.

### Testing Plan
1. Go to hearing prep
1. Confirm that updating the daily docket and the hearing worksheet still works


app/models/hearing.rb
```diff
@@ -0,0 +1,23 @@
+class Hearing < ApplicationRecord
+  belongs_to :hearing_day
+  belongs_to :appeal
+  belongs_to :judge, class_name: "User"
+
+  UUID_REGEX = /^\h{8}-\h{4}-\h{4}-\h{4}-\h{12}$/
+
+  def self.find_hearing_by_uuid_or_vacols_id(id)
```
lomaxap: 👏 👏 👏

app/models/legacy_hearing.rb
```diff
@@ -31,6 +31,10 @@ def venue
     self.class.venues[venue_key]
   end
 
+  def external_id
+    vacols_id
+  end
+
```
lomaxap: there are so there are 3 different ids that I can make out:

1. `external_id`
2. `uuid`
3. `vacols_id`

if `external_id` meant to cover either `uuid` or `vacols_id`, could we just name this `uuid` and nix `external_id` altogether? Especially since `external_id` on AMA hearings isn't external, right?

sharonwarner: I agree with you that it's weird. I went with external_id because it matches the model we went with for ama appeals. On the appeals models, external_id links to UUID for ama appeals and vacols_id for legacy appeals.

Issue #8263, opened 2018-12-12
Appeal Outcoding | Create Board Grant EPs and contentions


shanear:
When a board appeal is outcoded, this should create a board grant used to communicate what changes should be "effectuated" at VBA.

## Acceptance Criteria
When a board appeal is outcoded, as part of the outcoding job in #8262:
- For each granted compensation decision issue, a contention with the description of the decision issue should be created on an EP.
- If the decision issue is on a rating request issue, the contention should be created on a rating EP. If the decision issue is on a nonrating request issue, the contention should be created on a nonrating EP. Those EPs should be created if they do not already exist for that appeal. See the EP codes below.
- The created contentions should be connected to the decision issues by reference id.

## Notes
The EP Codes are:
 EP | Claim Label
-- | -- 
030BGR | Board Grant Rating 
030BGNR | Board Grant Non-Rating 

requires #8262

steve-gov:
Update: Compensation nonrating issue pending dispatch on https://appeals.cf.uat.ds.va.gov/queue/appeals/098cf858-9e94-4337-aa40-e3c5e0d3d457

steve-gov:
Issue outcoded, but not sure how to validate whether an EP was created. 

<img width="774" alt="screen shot 2019-01-23 at 11 00 27 am" src="https://user-images.githubusercontent.com/25468689/51630092-2010a900-1efe-11e9-9927-cc93f6309629.png">

steve-gov:
EP was not created. Shane's taking a look.

shanear:
This is cause the EP code doesn't seem to exist, will be following up on this tomorrow.

```
[...]

PR #8502, opened 2019-01-03
Create board grant effectuations, EPs, and contentions on outcoding

shanear:
connects #8263


app/models/appeal.rb
```diff
@@ -3,7 +3,6 @@ class Appeal < DecisionReview
 
   has_many :appeal_views, as: :appeal
   has_many :claims_folder_searches, as: :appeal
-  has_many :decision_issues, through: :request_issues
```
shanear: this is in `DecisionReview`

app/models/decision_issue.rb
```diff
@@ -8,6 +8,13 @@ class DecisionIssue < ApplicationRecord
   has_many :request_issues, through: :request_decision_issues
   has_many :remand_reasons, dependent: :destroy
   belongs_to :decision_review, polymorphic: true
+  has_one :effectuation, class_name: "BoardGrantEffectuation", foreign_key: :granted_decision_issue_id
+
+  def self.granted
+    # TODO: "allowed" is the disposition for BVA grants, not necessarily the disposition for granted HLRs and SCs
+    #       we need to add that
```
shanear: This represents a bigger issue, where `allowed` and `Granted` basically mean the same thing. Do we want to do a translation layer? Needs more thought

app/models/end_product_establishment.rb
```diff
@@ -152,30 +152,43 @@ def perform!(commit: false)
...
+    when DecisionDocument then source.effectuations.where(end_product_establishment: self)
+    end
+  end
+
   # VBMS will return ALL contentions on a end product when you create contentions,
   # not just the ones that were just created.
   def create_contentions!
-    issues_without_contentions = request_issues_ready_for_contentions
-    return if issues_without_contentions.empty?
+    return if records_ready_for_contentions.empty?
```
shanear: needed to refactor this method to use a generic contention record instead of request issues. More of this to come.

app/models/decision_document.rb
```diff
@@ -34,16 +37,46 @@ def submit_for_processing!
...
+
   private
 
+  def create_board_grant_effectuations!
+    appeal.decision_issues.granted.each do |granted_decision_issue|
+      BoardGrantEffectuation.find_or_create_by(granted_decision_issue: granted_decision_issue)
+    end
+  end
+
+  def process_board_grant_effectuations!
```
jiahuang: It is slightly confusing to me that `create_board_grant_effectuations!` and `process_board_grant_effectuations!` aren't operating on the same object types. Decision documents create board grant effectuations which create eps, but those eps are associated with the parent decision document instead of the effectuation.

Does it make more sense to have the ep source as the board grant effectuation? I see that in the comments for board grant effectuations that:

> Some are represented as contentions on an EP in VBMS. Others are tracked via Caseflow tasks

So not all effectuations have eps. Does that mean that `process_board_grant_effectuations!` would operate on tasks in the future? If there's multiple things that processing a effectuation can do, it seems better to store that logic in the effectuation instead of the decision document.

app/models/board_grant_effectuation.rb
```diff
@@ -0,0 +1,70 @@
+# BoardGrantEffectuation represents the work item of updating records in response to a granted issue on a Board appeal.
+# Some are represented as contentions on an EP in VBMS. Others are tracked via Caseflow tasks.
+
+class BoardGrantEffectuation < ApplicationRecord
+  belongs_to :appeal
+  belongs_to :granted_decision_issue, class_name: "DecisionIssue"
+  belongs_to :decision_document
+  belongs_to :end_product_establishment
+
+  validates :granted_decision_issue, presence: true
```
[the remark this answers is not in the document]
shanear: they are set automagically on create via `hydrate_from_granted_decision_issue`

app/models/decision_document.rb
shanear: > Does it make more sense to have the ep source as the board grant effectuation? I see that in the comments for board grant effectuations that:

yeah i would have liked to do that, but there can only be 1 `source` per end product establishment. Since effectuations are at a issue level, it wouldn't work. Plus, the ep and contentions need to be made via 1 VBMS call. Trying to do them per effectuation makes that extremely difficult.

I could rename the method though? I sort of like it how it's named

Does that mean that process_board_grant_effectuations! would operate on tasks in the future?

No. It would just get skipped. Tasks are just database objects and don't require API calls

app/models/decision_document.rb
```diff
@@ -34,16 +37,46 @@ def submit_for_processing!
   end
 
   def process!
+    return if processed?
+
     attempted!
-    VBMSService.upload_document_to_vbms(appeal, self)
+    upload_to_vbms!
+    create_board_grant_effectuations!
```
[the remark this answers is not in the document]
shanear: done

app/models/appeal.rb
shanear: @aroltsch it's worth noting that this change caused some of the queue tests to break (i fixed them). I think it's still correct, but it's worth noting this change to your team.

see: https://github.com/department-of-veterans-affairs/caseflow/pull/8502/commits/188c6835f858b6b883daab38464b8a08314cb896 for more deets

PR #8504, opened 2019-01-03
Set benefit type to compensation for all request issues

aroltsch:
All request issues created before Feb 14 are compensation only.


db/migrate/20190103200519_set_benefit_type_to_compensation_for_all_request_issues.rb
```diff
@@ -0,0 +1,5 @@
+class SetBenefitTypeToCompensationForAllRequestIssues < ActiveRecord::Migration[5.1]
+  def change
+    RequestIssue.where(benefit_type: nil).update_all(benefit_type: "compensation")
```
jiahuang: should there also be a migration to prevent `benefit_type` from being nil?

Issue #7994, opened 2018-11-26
Non-comp Flow | Decision page


shanear:
For each non-comp HLR and SC, the users need a way to record the decision being made for each request issue.

Full non-comp flow: https://va-gov.invisionapp.com/share/PFPCPJ7MEGC

## Acceptance Criteria
- On the non-comp reviews page, the claimant name should be a link to the dispositions page for that review.
- The disposition page should look like the mock below, it should show each issue on the review, and have a dropdown to select the disposition, as well as a place to write an optional decision description.
- When the "Edit Issues" button is pressed, it should redirect the user to the edit issues page for that decision review.
- When "Cancel" is pressed, it should redirect the user back to the reviews landing page.
- The decision date should default to today's date.

## Notes
Disposition Options:
Granted
Denied
DTA Error
Dismissed
Withdrawn

Out of scope:
- The "Complete" button

## Technical notes

We could likely reuse some Queue components here - the mock is based on LegacySelectDispositionsView

## Copy
There are some changes to the copy in the mocks:
'Decision' at the top to 'Select Dispositions'

## Mocks
*Higher-Level Review*
![disposition main - hlr](https://user-images.githubusercontent.com/11700545/49259068-b045dc80-f3ec-11e8-8b98-f50f1ce6ee6e.png)

*Higher-Level Review, selecting disposition*
![disposition main - hlr dd](https://user-images.githubusercontent.com/1596259/49190848-11fd3c80-f342-11e8-95be-99fcf03ba3cf.png)

*Supplemental Claim*
![disposition main - sc](https://user-images.githubusercontent.com/11700545/49259067-b045dc80-f3ec-11e8-8b24-e53d755fe9f0.png)

requires #7990

carolaponce:
Updated Mocks

![disposition main - hlr dd](https://user-images.githubusercontent.com/11700545/49258808-aa033080-f3eb-11e8-9088-fbf1dcdac667.png)
![disposition main - hlr](https://user-images.githubusercontent.com/11700545/49258815-ad96b780-f3eb-11e8-9a27-5345d1b440e4.png)
![disposition main - sc](https://user-images.githubusercontent.com/11700545/49258818-b091a800-f3eb-11e8-816a-b79b9f6e6d0e.png)

carolaponce:
Updated mocks
Change - copy change to: Review each issue and select a disposition.
[...]

PR #8508, opened 2019-01-03
Non-comp decision page

pkarman:
connects #7994 

### Description
New page to handle setting decisions on non-comp tasks.


pkarman:
screen shot 2019-01-03 at 5 13 24 pm

PR #8509, opened 2019-01-03
delete decision issue when request issue is deleted

jiahuang:
See https://dsva.slack.com/archives/C754H751T/p1546552804074900?thread_ts=1546545287.060900&cid=C754H751T

When a request issue is deleted, does the following:
1. deletes all of the request issue's associations from request_decision_issues table
2. deletes a decision issue if the decision issue has no other associated request issue

### DIFF
--- a/app/models/decision_issue.rb
+++ b/app/models/decision_issue.rb
@@ -27,6 +27,11 @@ class DecisionIssue < ApplicationRecord
     associated_request_issue&.issue_category
   end
 
+  def delete_on_removed_request_issue(request_issue_id)
+    # delete if the request issue is deleted and there are no other request issues associated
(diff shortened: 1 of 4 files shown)


app/models/decision_issue.rb
```diff
@@ -27,6 +27,11 @@ def issue_category
     associated_request_issue&.issue_category
   end
 
+  def delete_on_removed_request_issue(request_issue_id)
+    # delete if the request issue is deleted and there are no other request issues associated
+    delete if request_issues.length == 1 && request_issues.first.id == request_issue_id
```
pkarman: do we want to use `destroy` instead of `delete` here? https://stackoverflow.com/questions/22757450/difference-between-destroy-and-delete

PR #8512, opened 2019-01-04
Do not show rating issues from the future

leikkisa:
connects #8329


spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -427,7 +439,7 @@
       find("label", text: "Compensation", match: :prefer_exact).click
     end
 
-    fill_in "What is the Receipt Date of this form?", with: "04/20/2018"
+    fill_in "What is the Receipt Date of this form?", with: receipt_date.strftime("%D")
```
pkarman: +1

PR #8514, opened 2019-01-04
Complete the decision issue feature test

aroltsch:
Based on https://github.com/department-of-veterans-affairs/caseflow/commit/188c6835f858b6b883daab38464b8a08314cb896

### DIFF
--- a/app/models/concerns/issue_updater.rb
+++ b/app/models/concerns/issue_updater.rb
@@ -14,7 +14,7 @@ module IssueUpdater
   def delete_and_create_decision_issues!
     return unless appeal
     # We will always delete and re-create decision issues on attorney/judge checkout
-    DecisionIssue.where(id: appeal.decision_issues.map(&:id)).destroy_all
+    appeal.decision_issues.destroy_all
     create_decision_issues!
     fail_if_not_all_request_issues_have_decision!
   end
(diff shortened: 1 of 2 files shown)


app/models/concerns/issue_updater.rb
```diff
@@ -14,7 +14,7 @@ def update_issue_dispositions_in_caseflow!
   def delete_and_create_decision_issues!
     return unless appeal
     # We will always delete and re-create decision issues on attorney/judge checkout
-    DecisionIssue.where(id: appeal.decision_issues.map(&:id)).destroy_all
+    appeal.decision_issues.destroy_all
```
aroltsch: Since now we have a direct association to decision_issues we can use the above.

shanear: beautiful

spec/feature/queue/attorney_checkout_flow_spec.rb
```diff
@@ -320,8 +322,10 @@
 
         expect(page).to have_content(COPY::JUDGE_CHECKOUT_DISPATCH_SUCCESS_MESSAGE_TITLE % appeal.veteran_full_name)
 
-        # The decision issue should have the new content the judge added
+        # Two request issues are merged into 1 decision issue
+        expect(appeal.decision_issues.count).to eq 3
```
shanear: I was fixing this spec yesterday and wondering if this is what we actually wanted. Nice.

PR #8517, opened 2019-01-04
Create non-comp LOB on demand

pkarman:
connects #7990 

### Description
Uses a `BusinessLine` for each non-comp task assignment, creating it if necessary.

Also adds a UUID to HLR and SC tables so we can use that for URLs rather than an EPE (which will not exist for non-comp intakes).

### DIFF
--- a/app/controllers/claim_review_controller.rb
+++ b/app/controllers/claim_review_controller.rb
@@ -26,9 +26,7 @@ class ClaimReviewController < ApplicationController
   end
 
   def claim_review
-    @claim_review ||=
-      source_type.constantize.find(url_claim_id) ||
-      EndProductEstablishment.find_by!(reference_id: url_claim_id, source_type: source_type).source
+    @claim_review ||= source_type.constantize.find_by_uuid_or_reference_id!(url_claim_id)
   end
 
   def url_claim_id
--- a/app/controllers/claim_review_controller.rb
+++ b/app/controllers/claim_review_controller.rb
@@ -27,6 +27,7 @@ class ClaimReviewController < ApplicationController
 
   def claim_review
     @claim_review ||=
+      source_type.constantize.find(url_claim_id) ||
       EndProductEstablishment.find_by!(reference_id: url_claim_id, source_type: source_type).source
   end
 
--- a/app/models/claim_review.rb
+++ b/app/models/claim_review.rb
@@ -22,7 +22,7 @@ class ClaimReview < DecisionReview
   class << self
     def find_by_uuid_or_reference_id!(claim_id)
       claim_review = find_by(uuid: claim_id) ||
-                     EndProductEstablishment.find_by(reference_id: claim_id, source_type: to_s).source
+                     EndProductEstablishment.find_by(reference_id: claim_id, source_type: to_s).try(:source)
       fail ActiveRecord::RecordNotFound unless claim_review
       claim_review
     end
@@ -48,7 +48,7 @@ class ClaimReview < DecisionReview
   # Create that end product establishment if it doesn't exist.
   def create_issues!(new_issues)
     new_issues.each do |issue|
-      if non_comp?
+      if caseflow_only?
         issue.update!(benefit_type: benefit_type, veteran_participant_id: veteran.participant_id)
       else
         issue.update!(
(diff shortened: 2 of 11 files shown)


app/models/claim_review.rb
```diff
@@ -55,8 +55,14 @@ def create_issues!(new_issues)
 
   def create_non_comp_task!
     return if tasks.any? { |task| task.is_a?(DecisionReviewTask) } # TODO: more specific check?
-    # TODO: better user?
-    DecisionReviewTask.create!(appeal: self, assigned_at: Time.zone.now, assigned_to: User.system_user)
+    DecisionReviewTask.create!(appeal: self, assigned_at: Time.zone.now, assigned_to: non_comp_business_line)
```
shanear: since `non_comp` is sort of a misnomer (we actually mean non-comp/non-pension), i'd prefer avoiding that terminology.

app/controllers/claim_review_controller.rb
```diff
@@ -27,6 +27,7 @@ def request_issues_update
 
   def claim_review
     @claim_review ||=
+      source_type.constantize.find(url_claim_id) ||
       EndProductEstablishment.find_by!(reference_id: url_claim_id, source_type: source_type).source
```
shanear: I'm thinking this logic would probably be better in `ClaimReview`, maybe as `ClaimReview#find_by_id_or_reference_id` or something like that.

That way it's more reusable, and we can test it in a unit test and not an awkward controller test.

app/models/claim_review.rb
```diff
@@ -28,7 +28,7 @@ def ui_hash
   end
 
   def edit_issues_url
-    "/#{self.class.to_s.underscore.pluralize}/#{end_product_establishments.first.reference_id}/edit"
+    "/#{self.class.to_s.underscore.pluralize}/#{uuid}/edit"
```
shanear: I don't see where `uuid` is defined on `ClaimReview` or any of its subclasses

app/models/claim_review.rb
shanear: I think we should use the term `vbms` or `non_vbms` or `caseflow`? Because the real differentiating factor is what system it's tracked in

app/models/claim_review.rb
shanear: ohh, it was added in this PR. Why are we doing that?

pkarman:
Thanks for the feedback @shanear 

The addition of the UUID is because the new non-comp flow allows for HLR and SC to exist w/o any End Products, thus no End Product Establishment records, and therefore no existing way to edit the claim issues because the URL would require the EPE reference_id.

The way this PR approaches it allows for either the existing EPE reference_id URL syntax, or an explicit UUID on the claim review itself, to support both cases. The assumption is that the UUID is more unique so it is preferred in the `find_by_uuid_or_reference_id` method.

app/models/claim_review.rb
pkarman: this `edit_issues_url` is only used on the non-comp dispositions page, to allow the issues to be edited.

app/models/claim_review.rb
pkarman: how about `caseflow_only?` or `upstream_end_products?`

shanear:
> The way this PR approaches it allows for either the existing EPE reference_id URL syntax, or an explicit UUID on the claim review itself, to support both cases. The assumption is that the UUID is more unique so it is preferred in the find_by_uuid_or_reference_id method.

Sounds good. We just need to support the reference id, since VBMS won't have the UUID, and they're the ones who need to make the url

app/models/claim_review.rb
shanear: oh, that's sort of confusing. Should we update the name or add a comment to signal that?

app/models/claim_review.rb
```diff
@@ -53,10 +62,16 @@ def create_issues!(new_issues)
...
-  def create_non_comp_task!
+  def create_decision_review_task!
     return if tasks.any? { |task| task.is_a?(DecisionReviewTask) } # TODO: more specific check?
-    # TODO: better user?
-    DecisionReviewTask.create!(appeal: self, assigned_at: Time.zone.now, assigned_to: User.system_user)
+    DecisionReviewTask.create!(appeal: self, assigned_at: Time.zone.now, assigned_to: business_line)
+  end
+
+  def business_line
+    return unless caseflow_only?
```
shanear: `caseflow_only?` love that.

spec/controllers/higher_level_reviews_controller_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    it "finds by EPE reference_id" do
+      hlr.end_product_establishments.first.update!(reference_id: "abc123")
+
+      get :edit, params: { claim_id: hlr.end_product_establishments.first.reference_id }
+
+      expect(response.status).to eq 200
+    end
+  end
```
shanear: This controller test is now largly redundant of both the feature and unit tests. I'm cool with deleting it, but if you think it should stay, i'm cool with that too.

PR #8530, opened 2019-01-05
Remove v1 Status API

cmgiven:
In preparation for v3, removing all of the dead v1 code paths. Most v1 API tests were removed during the FACOLS transition, thus having only a single test remaining to delete.


app/models/legacy_appeal.rb
```diff
@@ -161,10 +161,6 @@ def vbms_id
     end
   end
 
-  def v1_events
-    @v1_events ||= AppealEvents.new(appeal: self, version: 1).all.sort_by(&:date)
-  end
-
```
mc-usds: Should this be updated or deleted? https://github.com/department-of-veterans-affairs/caseflow/blob/31eea6f16358e90288834e961cd116e1b8255547/app/models/legacy_appeal.rb#L666

cmgiven: Deleted. Good catch.

PR #8536, opened 2019-01-07
Make test less fragile

lowellrex:
Address test fragility as mentioned here https://github.com/department-of-veterans-affairs/caseflow/pull/8489#discussion_r245174678

Happy to consider an approach where we explicitly set the `created_at` timestamp instead as well.


spec/models/round_robin_task_distributor_spec.rb
```diff
@@ -27,11 +27,13 @@
...
     context "when task assigned to Organization is most recent task" do
-      let!(:user_task) { FactoryBot.create(:task, assigned_to: assignee) }
-      let!(:org_task) { FactoryBot.create(:task, assigned_to: FactoryBot.create(:organization)) }
+      before do
+        FactoryBot.create(:task, assigned_to: assignee)
+        FactoryBot.create(:task, assigned_to: FactoryBot.create(:organization))
+      end
 
       it "should return the most recent task assigned to a User" do
-        expect(round_robin_distributor.latest_task.id).to eq(user_task.id)
```
tomas-nava: I would add this expectation above line 36 so that the test will fail if we create the tasks in reverse order:

```ruby
expect(Task.all.max_by(&:created_at).assigned_to_type).to eq(Organization.name)
```

Issue #8283, opened 2018-12-13
Appeal Outcoding | Create Board Grant tasks on outcoding for non comp issues


shanear:
If a board grant is made on a decision issue that is not compensation or pension, the effectuation will not be tracked as part of an EP, but as part of a Caseflow task in the non comp queue described in #7990.

requires #7990

## Acceptance Criteria
When a board appeal is outcoded, after the decision document has been uploaded in the outcoding job referenced in #8262:
- If there are any granted issues on the appeal that have a benefit type other than compensation or pension, a "Board Grant" task should be created for the appropriate business line(s) based on the benefit type(s) of the decision issue(s).
- This task should appear on the non-comp landing page referenced here: https://github.com/department-of-veterans-affairs/caseflow/issues/7990 with the same columns

## Notes
- Issues on appeals are not limited to just one benefit type, so there may need to be multiple board grant tasks for one appeal.
- The task is more accurately known as an effectuation task. That word might be helpful when naming things.
- _Out of scope:_ what happens when you click on the task. That will be handled in https://github.com/department-of-veterans-affairs/caseflow/issues/8299

## Mocks
![returning disposition main](https://user-images.githubusercontent.com/1596259/50028070-478d6100-ffa3-11e8-984e-d2f46ede9f97.png)

jiahuang:
@steve-gov it will be easier to wait on validation for this ticket until #8299 is done

steve-gov:
Creating validation data for this story:

Added two issues to a Board Appeal in the Direct Review docket for `123458319 ` (LOB: Category: Decision Date):
- Education: Accrued: 12/31/2018
- Loan Guaranty: Other: 11/30/2018

Created appeal: https://appeals.cf.uat.ds.va.gov/queue/appeals/79630ab6-b9dd-443c-b38b-39e49fe59843

steve-gov:
Validated: both issues created a task in their respective LOBs.

<img width="839" alt="screen shot 2019-01-28 at 12 02 14 pm" src="https://user-images.githubusercontent.com/25468689/51862943-9fceb700-22f4-11e9-95f9-27f4487ebc02.png">

<img width="839" alt="screen shot 2019-01-28 at 12 02 33 pm" src="https://user-images.githubusercontent.com/25468689/51862954-a78e5b80-22f4-11e9-9cda-faf5d4b2315e.png">

PR #8548, opened 2019-01-07
Create BoardGrantEffectuation tasks for noncompensation issues

leikkisa:
connects #8283


app/models/board_grant_effectuation.rb
```diff
@@ -23,8 +23,12 @@ def contention_text
 
   private
 
+  def benefit_type
+    granted_decision_issue.benefit_type
+  end
+
   def effectuated_in_vbms?
-    %w[compensation pension].include?(granted_decision_issue.benefit_type)
+    %w[compensation pension].include?(benefit_type)
```
shanear: I think there might be an opportunity to factor this list with some of the stuff peter was working, on but cool if you don't want to do it in this PR

app/models/board_grant_effectuation.rb
```diff
@@ -35,17 +39,43 @@ def hydrate_from_granted_decision_issue
 
     if effectuated_in_vbms?
       self.end_product_establishment = find_or_build_end_product_establishment
+    else
+      find_or_build_noncomp_task
```
shanear: would like to avoid using the term `noncomp`, can we say `caseflow` or `queue`?

spec/models/board_grant_effectuation_spec.rb
```diff
@@ -66,6 +66,9 @@
 
       it "does not create end product establishment" do
```
shanear: maybe update this too something about making a task?

spec/models/board_grant_effectuation_spec.rb
```diff
@@ -66,6 +66,9 @@
...
         expect(subject.end_product_establishment).to be_nil
+        expect(BoardGrantEffectuationTask.find_by(appeal: decision_document.appeal)).to have_attributes(
+          assigned_to: BusinessLine.find_by(url: benefit_type)
+        )
```
shanear: I feel like we could do with a little more unit testing here to test the cases where there is already a task and ones where there is no task

app/models/board_grant_effectuation.rb
leikkisa: 👍 I wanted to differentiate it from the other effectuations but I don't think that's really necessary. I think in that case I'll call it effectuation_task

spec/models/board_grant_effectuation_spec.rb
leikkisa: 👍

app/models/board_grant_effectuation.rb
leikkisa: @pkarman Maybe we can chat in the morning so I know what you're up to?

spec/models/board_grant_effectuation_spec.rb
leikkisa: 👍

app/models/tasks/board_grant_effectuation_task.rb
```diff
@@ -0,0 +1,13 @@
+class BoardGrantEffectuationTask < DecisionReviewTask
+  def label
+    "Board Grant"
+  end
+
+  def serializer_class
+    ::WorkQueue::DecisionReviewTaskSerializer
```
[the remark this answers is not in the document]
leikkisa: 👍

Issue #8518, opened 2019-01-04
VBMS decisions with no rating issue break DTA contentions


shanear:
If a decision issue synced from VBMS is not connected to a rating issue, it is also missing a `description` and a `decision_text`.

If that decision issue is a DTA error, then the follow up contention on the EP breaks because the description is blank

shanear:
This is related to #4 on https://github.com/department-of-veterans-affairs/caseflow/issues/8239.

Going to tackle that first

shanear:
testing on fn: 635842354

shanear:
Confirmed this is working now

Issue #8239, opened 2018-12-12
Intake | Audit database columns


shanear:
`request_issues` and `decision_issues` have a lot of columns now, and as we've been developing, we've noticed that there isn't a tight definition of what they represent. Additionally, many of them may not be necessary anymore.

Prioritizing this because PA&I is going to retrieving metrics from a direct connection to our database. It needs to be intelligible and well defined.

Additionally, well defined columns will prevent confusion as we continue to build features and fix bugs

## Proposals
**request_issues**
1) __DONE__ change `review_request_type` and `review_request_id` to `decision_review_type` and `decision_review_id`. _This matches the terminology used everywhere else in Caseflow_

1) __DONE__ change from `rating_issue_profile_date` to `contested_rating_issue_profile_date` and `rating_issue_reference_id` to `contested_rating_issue_reference_id`. _This is to better match the naming of `contested_decision_issue_id`. Also adding `contested_` as a prefix better describes the relationship of the request issue to the rating issue._

1) __DONE__ change `contested_rating_issue_profile_date` from datetime to string. _This is because this field is used as an identifier to retrieve the appropriate rating from BGS. Having it as a datetime causes time zone logic to break the identifier, since it needs to be in the exact timezone it was returned in_

1) __DONE__ Add `contested_rating_issue_description`, and use that instead of `description` for the rating issue description. _The `description` column is overloaded and causes `description` to sometimes not represent the actual description. Instead, we should reserve `RequestIssue#description` for a method that combines `contested_rating_issue_decision_text` or `nonrating_description` properly._

1) __In Progress__ change `issue_category` to `nonrating_issue_category`, `decision_date` to `nonrating_decision_date`, and `description` to `nonrating_description` _This better designates that these fields are only used if the issue is a nonrating issue._
Note - we are now using decision_date for rating_issue_promulgation_dates, so we will not rename that at the moment.

1) __DONE__ Remove `disposition`. _This is tracked via `decision_issues` now. If we need request issue dispositions, we can calculate those as a function of the decision issues._

1) __DONE__ Remove `parent_request_issue_id`.  _The only place this is used is `RequestIssue.no_follow_up_issues`, which is not used. So we can delete that method too._

## decision_issues

1. __In progress__ rename `promulgation_date` to `rating_issue_promulgation_date`, and `profile_date` to `rating_issue_profile_date` _This better signifies that these fields apply to the rating issue._

[...]

PR #8549, opened 2019-01-07
Add description columns and begin writing to them

shanear:
since `RequestIssue#description` was far too overloaded, split it into 3 columns per the discussion with @pkarman and @leikkisa.

This is just the DB migration and the writes. Another data migration will be coming, and a follow up PR for switching over the reads.

connects #8239
connects #8518

### DIFF
--- a/app/models/request_issue.rb
+++ b/app/models/request_issue.rb
@@ -112,6 +112,7 @@ class RequestIssue < ApplicationRecord
 
     private
 
+    # rubocop:disable Metrics/MethodLength
     def attributes_from_intake_data(data)
       contested_issue_present = data[:rating_issue_reference_id] || data[:contested_decision_isssue_id]
 
@@ -143,6 +144,7 @@ class RequestIssue < ApplicationRecord
       }
     end
   end
+  # rubocop:enable Metrics/MethodLength
 
   def status_active?
     return appeal_active? if review_request.is_a?(Appeal)
--- a/app/models/request_issue.rb
+++ b/app/models/request_issue.rb
@@ -114,7 +114,7 @@ class RequestIssue < ApplicationRecord
 
     # rubocop:disable Metrics/MethodLength
     def attributes_from_intake_data(data)
-      contested_issue_present = data[:rating_issue_reference_id] || data[:contested_decision_isssue_id]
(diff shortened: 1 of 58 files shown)


app/models/request_issue.rb
```diff
@@ -113,14 +113,20 @@ def find_active_by_contested_decision_id(contested_decision_issue_id)
     private
 
     def attributes_from_intake_data(data)
+      contested_issue_present = data[:rating_issue_reference_id] || data[:contested_decision_isssue_id]
```
leikkisa: Isssue

leikkisa: isssue (too many S's?)

spec/support/intake_helpers.rb
```diff
@@ -318,6 +319,7 @@ def setup_request_issue_with_rating_decision_issue(decision_review, contested_ra
            contested_rating_issue_reference_id: contested_rating_issue_reference_id,
            contested_rating_issue_profile_date: decision_review.receipt_date - 1.day,
            description: "Test rating decision issue",
+           contested_issue_description: "Test nonrating decision issue",
```
leikkisa: Should this say rating?

spec/models/request_issues_update_spec.rb
```diff
@@ -311,6 +319,7 @@
             end_product_establishment: nonrating_end_product_establishment,
             contention_reference_id: nonrating_request_issue_contention.id,
             description: nonrating_request_issue_contention.text,
+            nonrating_issue_description: "Service connection for cancer was denied",
```
leikkisa: I think this is supposed to be nonrating_request_issue_contention.text

spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -950,6 +963,7 @@ def check_row(label, text)
 
       expect(RequestIssue.find_by(
                review_request: higher_level_review,
+               contested_issue_description: "Issue before AMA Activation from RAMP",
                description: "Issue before AMA Activation from RAMP",
```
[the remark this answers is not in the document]
shanear: I'm doing this because we need to continue to read from `description` until past data is migrated. There will be a follow on PR to fix this

PR #8550, opened 2019-01-07
Add hearing_regional_office column to Veteran and AHL job

### LINKED ISSUE #8484: AMA Veterans need to be associated with an RO
## User Stories: 

* A Hearing coordinator needs to schedule AMA Veterans for Board hearings by Regional Office and alternate locations associated with that RO. 

## Notes: 
*  VBA and the Board currently use this tool that matches a Veteran by their home address zip code to the VBA RO facility. Note that a Veteran isn't matched to an RO facility across state lines. 
* We investigated using BGS data field -- but none seem to actually be tied to the Veterans updated address. Those that contain station of jurisdiction are manually updated and often don't reflect the RO closest to the Veterans address. 
* We could also use geomatching to associate a Veteran to an RO (within state), but that seems to be less specific than the VA tool of zip codes, where the VA has determined which zip codes are associated with which RO. 

## VBA Jurisdiction Look up tool: 
http://vbacoweb03.dva.va.gov/bl/21/SOJ/default.asp?

Open Questions 
* If the VBA Jurisdiction tool can only be used manually (no API or data source), then it seems like we need someone in the Intake process checking the Veteran's address against the tool.
* Can this be done at Intake? Can this be added as part of the mail process?

lomaxap:
Resolves #8484

### Description
- Adds `hearing_regional_office` field to veteran model
- Adds logic to update `hearing_regional_office` in AHL job


app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -55,29 +61,57 @@ def perform
...
+  def find_legacy_ro_and_update_for_veteran(veteran)
+    ro = bfcorlid_to_ro_hash[LegacyAppeal.convert_file_number_to_vacols(veteran.file_number)]
+
+    veteran.update(hearing_regional_office: ro) unless ro.nil?
+
+    ro
+  end
+
+  def fetch_and_update_ro_for_veteran(veteran, lat, long)
+    distances = VADotGovService.get_distance(lat: lat, long: long, ids: ro_facility_ids)
```
mdbenjam: I think there was a way to pass in a type instead of a list of ids. I'm wondering if it would be better to use that type and pass regional office in the query instead of a list of ids. That way if a regional office is ever added, we'll include it in our options of possible closest ROs. Admittedly, we'd throw an error at some point because we wouldn't find it in our `RegionalOffice::CITIES` list, but I think that might still be preferable. What are your thoughts?

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -55,29 +61,57 @@ def perform
...
+
+    unless distances.empty?
```
mdbenjam: Do we want to silently fail in this case? Seems like there should always be a return value in distances, and if there isn't we'd probably want the job to fail?

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -55,29 +61,57 @@ def perform
...
+  end
 
-    VACOLS::Case.where(bfcorlid: bfcorlids).pluck(:bfcorlid, :bfregoff).to_h
+  def bfcorlid_to_ro_hash
+    @bfcorlid_to_ro_hash ||= begin
+      bfcorlids = veterans.pluck(:file_number).map do |file_number|
+        LegacyAppeal.convert_file_number_to_vacols(file_number)
+      end
+
+      VACOLS::Case.where(bfcorlid: bfcorlids).pluck(:bfcorlid, :bfregoff).to_h
```
mdbenjam: This is not a new problem, but I just had a realization that we can have a problem if there are multiple appeals in our list for the same veteran. Probably a very small case, but something to just be aware of.

db/migrate/20190107214300_add_regional_office_to_veteran.rb
```diff
@@ -0,0 +1,5 @@
+class AddRegionalOfficeToVeteran < ActiveRecord::Migration[5.1]
+  def change
+    add_column :veterans, :hearing_regional_office, :string
```
mdbenjam: I'm thinking that we might want to name this `closest_regional_office`, or `closest_regional_office_within_veteran_state`. I think the Queue team is interested in the same basic property: @tomas-nava . But still want us to avoid using `regional_office` without qualifiers.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
lomaxap: I was actually thinking about adding exception handling around VADotGovService so the entire job doesn't fail (thinking if it's just an issue specific to that veteran's data, we'll still cover the others). I suppose if it's a more fundamental bug, we wouldn't want to keep trying. Do you think that is a better approach?

app/jobs/fetch_hearing_locations_for_veterans_job.rb
lomaxap: Yea, this occurred to me too, in the future it may be easier to handle if we can limit it to just cases that haven't been distributed? And is this equally true for AMA and Legacy?

app/jobs/fetch_hearing_locations_for_veterans_job.rb
lomaxap: I don't think regional office means anything to the facility locator. The valid types now are 
- Health facilities
- Benefits facilities
- Cemeteries
- Vet Centers
I'm not sure what those are, but it might be that Benefits facilities are all ROs? Can check with Meredith or do some data validation if she's unsure.

mdbenjam: Hmm gotcha, darn. Yeah if benefits facilities works that'd be great, otherwise, it's not really a big deal.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
mdbenjam: We need some way to alert us that it's failing. I think at first we can start by just failing the job. Once we see how common failures are, if they're causing noise, we can report the failures to Sentry and not fail the entire job. I just want us to feel pressure to fix things if they're broken. And  broken job definitely does that :)

app/jobs/fetch_hearing_locations_for_veterans_job.rb
mdbenjam: This would only effect legacy. Do cases get removed from location 57 when they're distributed?

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -26,26 +26,32 @@ def create_missing_veterans
...
-        distance: alternate_hearing_location[:distance],
-        facility_id: alternate_hearing_location[:id],
-        name: alternate_hearing_location[:name],
-        address: full_address_for(alternate_hearing_location[:address]),
-        city: alternate_hearing_location[:address]["city"],
-        state: alternate_hearing_location[:address]["state"],
-        zip_code: alternate_hearing_location[:address]["zip"]
-      )
-    end
+  def find_or_update_ro_for_veteran(veteran, va_dot_gov_address:)
```
mdbenjam: Realized there's another bug in the job. 
I think `def veterans` should not be memoized and `def missing_veteran_file_numbers` should be:

```
def missing_veteran_file_numbers
  existing_veteran_file_numbers = Veteran.where(file_number: file_numbers).pluck(:file_number)
  file_numbers - existing_veteran_file_numbers
end
```

Right now the first job is going to be creating thousands of missing veterans because we don't limit the query at all.

mdbenjam: Actually this is still not right. We need to then limit the number of missing file numbers we return:

```
  def missing_veteran_file_numbers
    existing_veteran_file_numbers = Veteran.where(file_number: file_numbers).pluck(:file_number)
    (file_numbers - existing_veteran_file_numbers)[0..(500 - existing_veteran_file_numbers.length)]
  end
```

mdbenjam: Just tested this all and it seems to work. We should probably put the 500 in a constant since it's used twice.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -55,34 +61,63 @@ def perform
...
-    end
+  def facility_ids_for_veteran(veteran, va_dot_gov_address:)
+    ro = find_or_update_ro_for_veteran(veteran, va_dot_gov_address: va_dot_gov_address)
+
+    RegionalOffice::CITIES[ro][:alternate_locations] || [] << RegionalOffice::CITIES[ro][:facility_locator_id]
+  end
 
-    VACOLS::Case.where(bfcorlid: bfcorlids).pluck(:bfcorlid, :bfregoff).to_h
+  def ro_facility_ids_for_state(state_code)
+    RegionalOffice::CITIES.values.reject { |ro| ro[:facility_locator_id].nil? || ro[:state] != state_code }
```
mdbenjam: @MeredithStewart any idea what we should do for Veterans that are outside the country?

mdbenjam: We don't have to answer this now, but something we should consider.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -55,34 +61,63 @@ def perform
...
+                 when "GQ", "AQ", "RM", "FM", "CQ", "PS"
+                   "HI"
+                 when "PH", "RP", "PI"
+                   "PI"
+                 when "VI", "VQ", "PR"
+                   "PR"
+                 when "US"
+                   va_dot_gov_address[:state_code]
+                 else
+                   msg = "#{va_dot_gov_address[:country_code]} is not a valid country code."
```
mdbenjam: We might need a better story for Vets outside the country not in these places.

mdbenjam: But let's see when we error!

lomaxap: @MeredithStewart and I discussed this briefly. I think we agreed it would be good to create block for the ScheduleVeteranTask if the Vet's address is outside US territories. @sharonwarner should we add this to a sprint? We might be able to get away with just failing silently if we can't fit this in before Feb 14th.

lomaxap: created issue #8614

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -33,6 +36,7 @@ def find_or_update_ro_for_veteran(veteran, va_dot_gov_address:)
   def create_available_locations_for_veteran(veteran, va_dot_gov_address:, ids:)
     VADotGovService.get_distance(lat: va_dot_gov_address[:lat], long: va_dot_gov_address[:long], ids: ids)
       .each do |alternate_hearing_location|
+        AvailableHearingLocations.where(veteran_file_number: veteran.file_number).delete_all
```
mdbenjam: Maybe put this and the create in a transaction block? Only destroy if create is successful?

Also in general (we're bad at doing this on the project) it's better to use `destroy_all` since any `on_destroy` hooks don't run on `delete`

Issue #8386, opened 2018-12-19
Legacy Opt-in | Assign VACOLS opt in special issue per individual contention


leikkisa:
When we create contentions on a claim, we add can add special issue designations.  The "review by same office" special issue exists at the review-level and gets added to all contentions on a claim.

The vacols opt in special issue should only get added on the contentions for the issues with a vacols opt in associated with it.  Currently, it's getting added to all of the contentions at the time contentions are being created, if any one of the request issues on the decision review has a legacy issue opt in.

## Acceptance criteria
When a contention is created (whether during intake or edit), it should only have the vacols opt in special issue if the request issue has a legacy issue opt in.

## Validation
Create a Claim Review (HLR or SC), select yes for opting in legacy issues, and add more than one issue on the add issues page. For one issue, select a legacy issue to opt in.  For the other issue, do not. Then finish the review, and check the special issues on the contentions in VBMS. Only one of the issues should have the opt in special issue (SOC/SSOC opt in)

shanear:
<img width="608" alt="screen shot 2019-01-16 at 2 58 47 pm" src="https://user-images.githubusercontent.com/1596259/51284133-551d7880-199f-11e9-9c66-d13e04effd40.png">

why hello there.

shanear:
works also with some optins some non-opt ins

PR #8552, opened 2019-01-08
Make special issues per contention instead of per claim for VACOLS opt in

leikkisa:
connects #8386 
related https://github.com/department-of-veterans-affairs/connect_vbms/pull/208


app/models/request_issue.rb
```diff
@@ -180,6 +180,13 @@ def eligible?
     ineligible_reason.nil?
   end
 
+  def special_issues
+    specials = []
+    specials << { code: "VO", narrative: Constants.VACOLS_DISPOSITIONS_BY_ID.O } if legacy_issue_opted_in?
+    specials << { code: "SSR", narrative: "Same Station Review" } if decision_review.try(:same_office)
+    return specials unless specials.empty?
```
jiahuang: so this returns nil or an array? better to return the empty array imo

leikkisa: Most contentions don't have special issues, so I decided not to include it in the contentions data when there aren't any (could have done that downstream too).

Issue #8553, opened 2019-01-08
Assign RAMP tasks with VSO tasks complete to judges


amprokop:
Now that IHP-writing VSOs have started marking those tasks complete, we can adjust the JudgeTasks  to assign those RAMP cases. 

# AC
- `JudgeTask.assign_ramp_judge_tasks` assigns otherwise eligible cases where VSO tasks are complete to judges

PR #8555, opened 2019-01-08
update assignment method

amprokop:
Resolves #https://github.com/department-of-veterans-affairs/caseflow/issues/8553

### Testing Plan
- Manually tested using a RootTask with a completed IHP task as its child

### DIFF
--- a/app/models/tasks/judge_task.rb
+++ b/app/models/tasks/judge_task.rb
@@ -61,6 +61,7 @@ class JudgeTask < Task
   end
 
   def self.eligible_for_assigment?(task)
+    return false if task.appeal.nil?
     return false if task.appeal.class == LegacyAppeal
     return false if task.appeal.docket_name.nil?
     # Hearing cases will not be processed until February 2019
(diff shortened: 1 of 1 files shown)


app/models/tasks/judge_task.rb
```diff
@@ -71,12 +72,18 @@ def self.eligible_for_assigment?(task)
...
+
+    if has_completed_vso_task(task)
+      return true
+    end
+    
     task.children.empty?
   end
 
+  def self.has_completed_vso_task(task)
+    task.children.all? { |t| t.type == "InformalHearingPresentationTask" && t.status == "completed" }
```
lowellrex: What do you think about dropping this in where `task.children.empty?` lives at the current moment? I think it covers this case as well.

```ruby
task.children.all? { |t| !t.type.is_a?(JudgeTask) && t.completed? }
```

amprokop: wow yes!

PR #8558, opened 2019-01-08
Use VACOLS task eager loading functionality for organizations

lowellrex:
Currently we cannot load the VLJ support staff team queue because it takes too long to load. This PR fixes that by using the VACOLS eager loading stuff we added in #8427 for normal `TasksController` in `Organizations::TasksController` as well.

### DIFF
--- a/app/controllers/organizations/tasks_controller.rb
+++ b/app/controllers/organizations/tasks_controller.rb
@@ -19,8 +19,7 @@ class Organizations::TasksController < OrganizationsController
 
   def json_tasks(tasks)
     ActiveModelSerializers::SerializableResource.new(
-      tasks,
-      each_serializer: ::WorkQueue::TaskSerializer,
+      AppealRepository.eager_load_legacy_appeals_for_tasks(tasks),
       user: current_user
     ).as_json
   end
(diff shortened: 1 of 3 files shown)


app/controllers/organizations/tasks_controller.rb
```diff
@@ -19,8 +19,7 @@ def organization_url
 
   def json_tasks(tasks)
     ActiveModelSerializers::SerializableResource.new(
-      tasks,
-      each_serializer: ::WorkQueue::TaskSerializer,
+      AppealRepository.eager_load_legacy_appeals_for_tasks(tasks),
```
lowellrex: I would rather call `TasksController.json_tasks` here so that this mistake doesn't happen again (where we forget to transfer changes for one controller's serializer method to the other, but I couldn't find a clean way to do that. Open to suggestions.

app/services/appeal_repository.rb
```diff
@@ -15,6 +15,22 @@ def self.transaction
     end
   end
 
+  def self.eager_load_legacy_appeals_for_tasks(tasks)
```
mdbenjam: I think this is fine for now. I know you're just moving this in here. But I think in the future I'd love for us to create a class method on LegacyAppeal that mass loads an array of legacy appeals. That could be a more general solution to this problem that I think is super common across Caseflow.

PR #8564, opened 2019-01-08
Update document_count to use Caseflow Documents table

hschallhorn:
References #8192 

### Acceptance Criteria
- [x] Calls to `document_count` to not make calls to vacols or vbms
- [x] `document_count` returns the correct number of documents


hschallhorn:
Looks like I missed a few other tests that check doc_count. Checking them out now.

spec/models/legacy_hearing_spec.rb
```diff
@@ -128,12 +128,14 @@
...
     subject { hearing.to_hash_for_worksheet(nil).with_indifferent_access }
 
+    let(:documents) { create_list(:document, 3) }
     let(:appeal) do
       create(:legacy_appeal, :with_veteran, vacols_case:
         create(
           :case_with_form_9,
           bfcorlid: "12345678",
-          case_issues: [create(:case_issue)]
+          case_issues: [create(:case_issue)],
```
aroltsch: Ah, I see. It seems as the Hearing Prep application is also asking for a number of documents. I think all users for the hearing prep should have a Reader role although I can't recall how cases are being pulled for the hearing prep users. I am afraid we might break hearing prep. Two options:
1. Ensure Reader job runs for the hearing prep users correctly meaning that hearing prep users have a reader role and that we pull all assigned appeals for the hearing prep users.
2. Create two methods: number_of_documents_from_caseflow and number_of_documents_from_vbms. Queue can use number_of_documents_from_caseflow while hearing prep can use number_of_documents_from_vbms.

cc @mdbenjam @sharonwarner

hschallhorn: We have a preference on this?

mdbenjam: This is a Sharon question. My guess is it's fine. The document count might be 3 hours out of date, but that's likely sufficient.

aroltsch: I think we face a different problem here. If we don't pull documents into Caseflow for a particular appeal through the  Reader job, the document count for that will always be zero, correct? This code here: ```Document.where(file_number: veteran_file_number).size``` assumes there are documents in Caseflow DB and it never calls eFolder to refresh.

mdbenjam: Ah good call. We should fall back to calling eX directly if we don't have the documents. i.e.

```
  def number_of_documents
    Document.where(file_number: veteran_file_number).size || document_fetcher.number_of_documents
  end
```

or something.

aroltsch: @hschallhorn I chatted with Mark and we agreed it is better to have two separate methods.

hschallhorn: 👍

client/app/queue/AppealDocumentCount.jsx
```diff
@@ -29,7 +29,7 @@ class AppealDocumentCount extends React.PureComponent {
         timeout: { response: 5 * 60 * 1000 }
       };
 
-      ApiUtil.get(`/appeals/${this.props.externalId}/document_count`, requestOptions).then((response) => {
+      ApiUtil.get(`/appeals/${this.props.externalId}/document_count_from_caseflow`, requestOptions).then((response) => {
```
aroltsch: Hey @hschallhorn, I don't think hearing prep is using this endpoint so you can leave it as is. They are using this method `number_of_documents`directly from `legacy_hearing` model which you already handled correctly.

app/controllers/appeals_controller.rb
```diff
@@ -31,6 +31,12 @@ def document_count
     return handle_non_critical_error("document_count", e)
   end
 
+  def document_count_from_caseflow
+    render json: { document_count: appeal.number_of_documents_from_caseflow }
+  rescue StandardError => e
```
aroltsch: I would not catch this error like this. 
`handle_non_critical_error` is for errors that we don't want to see in Sentry but instead count them in Datadog. As I mentioned in my previous comment, you don't need to implement a new method in the controller.

app/controllers/appeals_controller.rb
```diff
@@ -20,9 +20,7 @@ def show_case_list
   end
 
   def document_count
-    render json: { document_count: appeal.number_of_documents }
-  rescue StandardError => e
-    handle_non_critical_error("document_count", e)
```
aroltsch: I see. It was here initially like this. Interesting!

spec/models/legacy_appeal_spec.rb
```diff
@@ -426,9 +426,10 @@
       create(:case, documents: documents)
     end
 
-    subject { appeal.number_of_documents }
```
aroltsch: I think you would still need a test for both methods. I'd suggest leaving the test for `number_of_documents` the same and just adding one tiny test for `number_of_documents_from_caseflow`

app/models/appeal.rb
```diff
@@ -46,9 +46,13 @@ def document_fetcher
     )
   end
 
-  delegate :documents, :number_of_documents, :manifest_vbms_fetched_at,
+  delegate :documents, :manifest_vbms_fetched_at, :number_of_documents,
            :new_documents_for_user, :manifest_vva_fetched_at, to: :document_fetcher
 
+  def number_of_documents_from_caseflow
+    Document.where(file_number: veteran_file_number).size
```
aroltsch: Even though we think Reader job should update documents for all appeals, let's add this check: 
if there are no documents in Caseflow for this appeal, use `number_of_documents` from the document_fetcher.

app/models/appeal.rb
```diff
@@ -46,9 +46,14 @@ def document_fetcher
     )
   end
 
-  delegate :documents, :number_of_documents, :manifest_vbms_fetched_at,
+  delegate :documents, :manifest_vbms_fetched_at, :number_of_documents,
            :new_documents_for_user, :manifest_vva_fetched_at, to: :document_fetcher
 
+  def number_of_documents_from_caseflow
```
aroltsch: Could you add a comment in both methods explaining the logic please. I don't think it would super clear for a new developer looking at it.  Another thing you could is to move these two methods into a document_fetcher this way you don't have to repeat the same logic twice in appeal.rb and legacy_appeal.rb.

hschallhorn: When document_fetcher is initialized, it calls `fetch_documents_from_service`. So moving these methods to document_fetcher will make a call to eFolder or VBMS.

PR #8565, opened 2019-01-08
Authorize child task creation with available task actions

### LINKED ISSUE #8273: Use available task actions to determine whether we can create a child task
Before we create `Task` objects we confirm that the person attempting to create the task can do so (by way of `Task.verify_user_can_create!` and the overridden methods in its various subclasses). The only way users create tasks is through the available task actions dropdown on the case details page [1], which we populate through with the `Task.available_actions_unwrapper` method. Since we limit the set of actions that appear in that dropdown to only those actions that the current user can take, we can use the presence of a task in that set of actions to determine if the task we are trying to create can be created without including additional logic in the task models.

Using `MailTask` as an example, [`verify_user_can_create!`](https://github.com/department-of-veterans-affairs/caseflow/blob/afb9e9aa636a18da8bd1cc12a0c504139796e6e1/app/models/tasks/mail_task.rb#L22) (I'm using the new names for these functions after #8248 is merged) limits the set of folks who can create `MailTask`s to members of the `MailTeam`. The only place `MailTask`s are made available to be created is [in `RootTask.available_actions`](https://github.com/department-of-veterans-affairs/caseflow/blob/afb9e9aa636a18da8bd1cc12a0c504139796e6e1/app/models/tasks/root_task.rb#L11) which only displays the option to create a `MailTask` if the current user is a member of the `MailTeam`. Instead of duplicating this logic we could use the existence of `MailTask` in the parent task's set of available actions to determine whether the current user can create the task. Perhaps something like:

```ruby
def create_from_params(params, user)
  parent = Task.find(params[:parent_id])
  verify_user_can_create!(parent, user)
  ...

def verify_user_can_create!(parent, user)
  parent.available_actions_unwrapper(user).any? do |opt|
    opt.data && opt.data.type && opt.data.type == type
  end
end
```

## Benefits of this approach
* Less code in our codebase.
* Fewer paths through our code.
* Eliminates logic that effectively accomplishes the same thing twice.
* Allows us to make a change to who can create a task in a single place, reducing the likelihood that, for example, we want to allow another group of people to create `MailTask`s so we add them to the condition in `RootTask.available_actions` but forget to change `MailTask.verify_user_can_create!`.

## Caveats
* We will need a fallback for when we create `ColocatedTask`s from `AttorneyLegacyTask`s because those tasks will not have parent tasks.
* We will need to include the type of task we are creating in each option we put into the array of available actions. Currently the shape of this array looks like:
```ruby
[
  {
    label: "Mark task complete",
    value: "modal/mark_task_complete"
  },
  {
    label: "Human-readable label",
    value: "URL path", 
    data: { # Optional field
      selected: "value of currently selected option in dropdown",
      options: [ { label: "label", value: 12 }, ... ],
      type: GenericTask.name
    },
  },
  ...
]
```
* It may make sense to consider alternative approaches building this set of available actions (maybe moving from anonymous hashes to objects for the individual options and the `available_actions_unwrapper`?) if that makes it easier to access the information we want.
* We have a [`before_create`](https://github.com/department-of-veterans-affairs/caseflow/blob/afb9e9aa636a18da8bd1cc12a0c504139796e6e1/app/models/task.rb#L12) hook that sets the `assigned_at` value of the task and changes the parent task's status to `on_hold`. We will need to split this into two different hooks (maybe `before_create :set_assigned_at` and `after_create :put_parent_task_on_hold`).

## Acceptance criteria
* [ ] `Task.verify_user_can_create!` determines if somebody can create a task based on the parent task's `available_actions` instead of bespoke `verify_user_can_create!` methods for each task subclass.

[1] Tasks can also be created with the judge bulk assign functionality and automatically by the application in a handful of places. But the automatic task creation will bypass `Task.create_from_params` and any task that is judge bulk assign-able will have the "Assign to attorney" option as one of the available actions for that task.

tomas-nava:
Use the task actions made available to users to decide whether a user is authorized to create a task through the interface.

Resolves #8273


app/models/legacy_tasks/attorney_legacy_task.rb
```diff
@@ -7,22 +7,10 @@ def available_actions(role)
...
-      {
-        label: COPY::ATTORNEY_CHECKOUT_OMO_LABEL,
-        value: "omo_request/submit"
-      }
-    ]
+    actions = [Constants.TASK_ACTIONS.REVIEW_DECISION.to_h, Constants.TASK_ACTIONS.SUBMIT_OMO_REQUEST_FOR_REVIEW.to_h]
 
     if FeatureToggle.enabled?(:attorney_assignment_to_colocated, user: assigned_to)
-      actions.push(
-        label: COPY::ATTORNEY_CHECKOUT_ADD_ADMIN_ACTION_LABEL,
```
lowellrex: Now that we've moved these variables to the TASK_ACTIONS file, let's remove them from COPY.json (and update the tests that reference those variables to use `Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.label`).

app/models/task.rb
```diff
@@ -155,10 +157,11 @@ def verify_user_can_update!(user)
...
   end
 
-  def self.verify_user_can_create!(user)
-    unless user.attorney_in_vacols? || user.judge_in_vacols?
-      fail Caseflow::Error::ActionForbiddenError, message: "Current user cannot assign this task"
+  def self.verify_user_can_create!(user, parent)
+    can_create = parent.available_actions_unwrapper(user).any? do |action|
+      action.dig(:data, :type) == name || action.dig(:data, :options)&.any? { |option| option.dig(:value) == name }
     end
+    fail Caseflow::Error::ActionForbiddenError, message: "Current user cannot assign this task" unless can_create
```
lowellrex: What do you think about adding the user ID, task ID, and task type to this error message? I guess we can get that information from Sentry so maybe it's not worth it.

app/models/task.rb
```diff
@@ -155,10 +157,11 @@ def verify_user_can_update!(user)
     end
   end
 
-  def self.verify_user_can_create!(user)
-    unless user.attorney_in_vacols? || user.judge_in_vacols?
-      fail Caseflow::Error::ActionForbiddenError, message: "Current user cannot assign this task"
+  def self.verify_user_can_create!(user, parent)
+    can_create = parent.available_actions_unwrapper(user).any? do |action|
```
lowellrex: Do we need to protect against the `parent` parameter from being `nil` here? It looks like [if there is no `parent_id` field](https://github.com/department-of-veterans-affairs/caseflow/pull/8565/files#diff-c95c9451cf097bdd58f73d0f9b114f97R90) in the set of request parameters `parent` might be `nil` here.

app/models/tasks/colocated_task.rb
```diff
@@ -43,24 +51,14 @@ def automatically_assign_org_task?
...
-    ]
+    actions = [Constants.TASK_ACTIONS.PLACE_HOLD.to_h, Constants.TASK_ACTIONS.ASSIGN_TO_PRIVACY_TEAM.to_h]
 
     if %w[translation schedule_hearing].include?(action) && appeal.class.name.eql?("LegacyAppeal")
-      actions.unshift(
-        label: format(COPY::COLOCATED_ACTION_SEND_TO_TEAM, Constants::CO_LOCATED_ADMIN_ACTIONS[action]),
-        value: "modal/send_colocated_task"
-      )
+      send_to_team = Constants.TASK_ACTIONS.SEND_TO_TEAM.to_h
+      send_to_team[:label] = format(COPY::COLOCATED_ACTION_SEND_TO_TEAM, Constants::CO_LOCATED_ADMIN_ACTIONS[action])
```
lowellrex: Clever way to handle the label here! I like it!

app/models/task.rb
```diff
@@ -155,10 +157,11 @@ def verify_user_can_update!(user)
...
+      action.dig(:data, :type) == name || action.dig(:data, :options)&.any? { |option| option.dig(:value) == name }
```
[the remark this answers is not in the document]
tomas-nava: Yeah, I think this'd be a productive way to go. I'd been thinking of adding a `creatable_tasks` style parameter to the data objects, but that would've been redundant and difficult to maintain.

app/models/task.rb
tomas-nava: Good catch! Should we fail if there's no parent, or fall back to checking for judge/attorney in VACOLS like I've done in [the `ColocatedTask` implementation of `verify_can_create`](https://github.com/department-of-veterans-affairs/caseflow/blob/a0098c6813b8c7d3dd5d932cdc00082c74e60a01/app/models/tasks/colocated_task.rb#L34)?

lowellrex: Let's fail. The only time I think there will be no parent is for `ColocatedTask`s attached to `LegacyAppeal`s and your change to support that for `ColocatedTask` should cover us here.

PR #8575, opened 2019-01-08
Allow editing document ID

### LINKED ISSUE #7606: Edit document ID in Queue
## Background

We've gotten 3 different attorneys asking how to edit Document ID. How might we help attorneys edit Document ID in Case details?

## Mocks

Context: These mocks show what editing individual data points in Case Details could look like. As we're reviewing these designs, let's consider how they would scale across different editing scenarios. 

User first clicks "Edit" from Case details:
screen shot 2018-12-05 at 12 01 08 pm

Editing modals in case details consist of a: 
- Header (format: "Edit" + "(data being edited)"
- Optional instructional text to describe what the user is editing and the impact (if applicable)
- Text field pre-filled with the values being edited
- Save/cancel actions

image
- When the modal opens, we display a focused view of the text field so that the user can quickly edit. 
- We enable the save button when the user changes something in the text field. This is a visual queue to reinforce that we acknowledge their change. 

image
- Error state (above). 
  - Open question: How can we make this error message more helpful? From my understanding of the document ID formats in #5993, they can vary based on new and old cases. Ideally, we would give the user the correct format so they can fix the error. How many possible formats are there?

## AC 

- When a user clicks Edit next to the Decision Document ID field, the modal (above) appears on top of the Case Details page. 
- When the modal appears, the text field is displayed in a focused state.
- When the user makes any edit, we enable the save button. 
- Clicking "Save" validates the document ID field (see #5993 for validations that were implemented in the attorney-judge checkout flow).
  - If the format is invalid, present an error message. 
- Clicking "Save" returns the user to the Case Details page where they will see the change reflected. 
- Clicking "Cancel" or "x" returns the user to the Case Details page where they won't see any change reflected.

monfresh:
**Why**: Sometimes, attorneys enter the wrong document ID. They need to
be able to correct the ID after the fact.

**How**:
For now, we only display the Edit option for AMA cases.
After the user enters the new document ID, the backend performs two
validations:
- The current user must be either the assigned attorney or judge.
Although all users can view any appeal, only those who are assigned to
it are able to change the document ID.
- The document ID must follow a specific format:
12345-12345678 or 12345678.123 or 12345678.1234 for Draft Decisions

Resolves #7606

Testing Plan:
- View queue as user BVASCASPER1
- Click on any AMA case and complete the checkout flow
- Go back to the appeal details page
You should now see an "Edit" button next to the Document ID
- Click on the Edit button
You should see a modal with a text field pre-filled with the document
ID. The Save button should be disabled by default.
- Start typing in the field
The Save button should now be enabled.
- Enter an invalid ID, such as "1234" and click Save
You should see an error message in the modal letting you know which
formats are valid.
- Enter a valid ID, such as "12345678.123" and click Save
The modal should be dismissed, and the new document ID should appear on
the details page


monfresh:
@laurjpeterson @aroltsch Here are some screenshots:

screen shot 2019-01-08 at 4 51 41 pm
screen shot 2019-01-08 at 4 51 19 pm
screen shot 2019-01-08 at 4 13 36 pm
screen shot 2019-01-08 at 4 11 30 pm

Things that still need to be cleaned up:

- The Edit link should just be a pencil icon, not a button according to the mocks in #7606. 
- The alignment of the Edit button is off
- The format validation currently only exists on the appeals details page, but we should also add it on the checkout page

monfresh:
Also, I wasn't able to make the text field in focus automatically after the modal appears. @mdbenjam can you help with that please?

config/routes.rb
```diff
@@ -207,6 +207,7 @@
   end
 
   post '/case_reviews/:task_id/complete', to: 'case_reviews#complete'
+  post '/case_reviews/:task_id/update', to: 'case_reviews#update'
```
aroltsch: Is it going to be `attorney_case_reviews.task_id` or `attorney_case_reviews.id`?

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -108,4 +108,12 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
   attribute :caseflow_veteran_id do
     object.veteran ? object.veteran.id : nil
   end
+
+  attribute :document_id do
+    object.attorney_case_reviews.map(&:document_id).last
+  end
+
+  attribute :attorney_case_review_id do
+    object.attorney_case_reviews.last&.id
```
aroltsch: Should we perhaps create a method `last_attorney_case_review` that we can reuse to get a document_id and an id? It might be cleaner that way.

app/services/update_attorney_case_review.rb
```diff
@@ -0,0 +1,93 @@
...
+    FormResponse.new(success: success, errors: errors.messages)
+  end
+
+  private
+
+  attr_reader :id, :document_id, :user_id
+  attr_writer :user_id
+
+  def case_review
+    @case_review ||= AttorneyCaseReview.find_by(id: id)
```
aroltsch: Should we throw an error if not found? I usually use `find` method that will throw an exception if the item is not found.

config/routes.rb
aroltsch: It appears to me it is going to be `attorney_case_reviews.id` so I would say `patch '/case_reviews/:id'` would be better?

app/services/update_attorney_case_review.rb
```diff
@@ -0,0 +1,93 @@
+class UpdateAttorneyCaseReview
```
aroltsch: As I mentioned previously I really like this type of classes! I've used the same pattern at my previous job - we always aimed for less code at the controller level. I think creating a separate directory called `workflows` would be better instead of including these in `services`.

laurjpeterson:
Thanks @monfresh! I'd like @allyceh to take a look too. Functionally it looks correct, but the blue edit button looks different than design outlined in https://github.com/department-of-veterans-affairs/caseflow/issues/7606. Can you help explain why that is? My hunch is that you're using an existing button component, which is quicker and simpler to implement.

app/services/update_attorney_case_review.rb
monfresh: Yes! Thanks for the reminder.

config/routes.rb
monfresh: Yes. Good point! I was using `patch` at first but changed it to get the frontend working with the `requestSave` function, but I can switch to `requestPatch`.

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -108,4 +108,16 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
...
+  attribute :document_id do
+    latest_attorney_case_review.document_id
+  end
+
+  attribute :attorney_case_review_id do
+    latest_attorney_case_review.id
+  end
+
+  def latest_attorney_case_review
+    AttorneyCaseReview.where(task_id: Task.where(appeal: object).pluck(:id)).order(:created_at).last
```
aroltsch: I think putting this method in the appeal model would be better so it can be re-used by other parts of the application.

spec/system/queue/case_details/update_document_id_spec.rb
```diff
@@ -0,0 +1,22 @@
+require "rails_helper"
+
+RSpec.describe "Updating Document ID", type: :system do
```
aroltsch: We have been putting feature specs in the `feature` directory. I'd prefer to have one place for it.

app/models/serializers/work_queue/appeal_serializer.rb
monfresh: The reason for putting it in the serializer is that currently this is the only place where it is being used. I believe in keeping models lean and free of business logic. I noticed that the Appeal model was already much longer than the recommended 100 lines, and I didn't want to make it worse.

If in the future we end up needing this method elsewhere, we could put it in a View model or Query object, as described in this article that I reference often:
https://codeclimate.com/blog/7-ways-to-decompose-fat-activerecord-models/

Thoughts?

aroltsch: Ok, sounds good!

spec/system/queue/case_details/update_document_id_spec.rb
aroltsch: Great point!

monfresh:
@laurjpeterson Here is an updated screenshot:

screen shot 2019-01-10 at 11 32 47 am

spec/system/queue/case_details/update_document_id_spec.rb
monfresh: I ended up using a feature spec for now because the system test didn't work out of the box. I'll do some more research to figure out how they work.

mdbenjam: Could you talk a little more about the differences here between System and Feature tests? What advantages do system tests give?

client/app/queue/CaseTitleDetails.jsx
```diff
@@ -68,16 +77,70 @@ const CaseDetailTitleScaffolding = (props) => <div {...containingDivStyling}>
 </div>;
 
 export class CaseTitleDetails extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      showModal: false,
+      buttonCss: 'usa-button-disabled',
```
mdbenjam: Not sure we need this anymore.

client/app/queue/CaseTitleDetails.jsx
```diff
@@ -119,10 +182,41 @@ export class CaseTitleDetails extends React.PureComponent {
...
-      { !userIsVsoEmployee && primaryTask && primaryTask.documentId &&
+      { !userIsVsoEmployee && appeal && appeal.documentID &&
         <React.Fragment>
           <h4>{COPY.TASK_SNAPSHOT_DECISION_DOCUMENT_ID_LABEL}</h4>
-          <div><CopyTextButton text={primaryTask.documentId} /></div>
+          <div><CopyTextButton text={this.state.value || appeal.documentID} />
+            <Button
+              linkStyling
+              onClick={this.handleModalClose} >
+              <span {...css({ position: 'absolute' })}>{pencilSymbol()}</span>
```
mdbenjam: Why do we need absolute positioning here?

spec/system/queue/case_details/update_document_id_spec.rb
mdbenjam: Want to make sure we understand all the advantages before switching since having our tests split incurs some cost on the teams understanding of the structure of our code.

mdbenjam:
To make something take focus you probably need to use a ref. This is a good stack overflow example:
https://stackoverflow.com/questions/28889826/set-focus-on-input-after-render

spec/system/queue/case_details/update_document_id_spec.rb
monfresh: Apart from the reasons quoted above, I don't have specific details, but I will find out. Both the Rails and RSpec teams are recommending this as the new way forward because feature tests involve complexity created by RSpec maintainers that was not supported by official Rails APIs, whereas system tests rely on clean APIs exposed by Rails and maintained by the Rails team.

client/app/queue/CaseTitleDetails.jsx
monfresh: Good catch!

client/app/queue/CaseTitleDetails.jsx
monfresh: I copied and pasted it from another file where we use the pencil icon. Without this css directive, the alignment is not right. UI positioning is not my strong suit, so I'm not sure how else to make everything align.

Issue #8543, opened 2019-01-07
Appeal Intake | Select benefit type per issue


shanear:
We need user to select benefit type when they are entering a non-comp issue.
Add another dropdown field to the nonrating modal, the first field. Based on what the user selects, the Category dropdown options will change.

## Acceptance Criteria
When a user is intake an appeal, and adding an issue:
- On the on non-comp modal, there should be a drop down at the top of the modal called "Benefit Type" with options representing each possible benefit type.
- The dropdown should default to "Compensation"
- The options in "Category", should filter to just the options that belong to the selected "Benefit type"
- The selected benefit type should save to the request issue when the appeal is saved

## Comps
No comps since it's relatively straightforward. @carolaponce can provide tweaks if necessary.

carolaponce:
I had this solution in my designs for that option. Let me know if you need anything else.
![issues_noncomp_1](https://user-images.githubusercontent.com/11700545/50992430-260d8f80-14e5-11e9-9034-d799aa7e8c93.png)

pkarman:
Thanks @carolaponce 

I note that this mock uses a dropdown to show the benefit type, while elsewhere we use a group of radio buttons.

![screen shot 2019-01-10 at 1 49 39 pm](https://user-images.githubusercontent.com/1205061/50993179-a250a480-14de-11e9-9b79-ce27d0541570.png)

Is the difference in these 2 presentations intentional? Could we standardize on one?

carolaponce:
My intention was to keep user from scrolling in the modal.

steve-gov:
Validated:

![screencapture](https://user-images.githubusercontent.com/25468689/51219531-35764980-18e6-11e9-8bcb-87c3ecd44474.gif)

shanear:
Still need to test whether the benefit type saves correctly

PR #8577, opened 2019-01-08
Benefit type per Appeal issue

pkarman:
connects #8543 

### Description
Adds Benefit Type picker to Nonrating issue modal for Appeal form type.

Also fixes bug with untimely exemption modal being shown on every nonrating issue.

screen shot 2019-01-10 at 3 01 32 pm


spec/feature/intake/appeal_spec.rb
```diff
@@ -744,7 +771,6 @@ def check_row(label, text)
         expect(page).to have_content("Does issue 3 match any of these VACOLS issues?")
 
         add_intake_rating_issue("None of these match")
-        add_untimely_exemption_response("Yes")
```
pkarman: this untimely modal should not have been appearing in this scenario. now it doesn't.

app/models/appeal.rb
```diff
@@ -207,14 +207,16 @@ def previously_selected_for_quality_review
...
 
   def benefit_type
-    # temporary until ticket for appeals benefit type by issue is implemented
-    # https://github.com/department-of-veterans-affairs/caseflow/issues/5882
-    "compensation"
+    fail "Multiple benefit_types possible"
+  end
+
+  def caseflow_only?
+    false
```
leikkisa: Should this be true since Appeals are (almost) all in Caseflow?  Not sure if the outcoding is included. Maybe add a comment here to indicate what this means.

leikkisa: Should this be true since Appeals are (almost) all in Caseflow? Not sure if the outcoding is included. Maybe add a comment here to indicate what this means, or it might benefit from a renaming.

app/models/appeal.rb
```diff
@@ -207,14 +207,16 @@ def previously_selected_for_quality_review
   end
 
   def benefit_type
-    # temporary until ticket for appeals benefit type by issue is implemented
-    # https://github.com/department-of-veterans-affairs/caseflow/issues/5882
-    "compensation"
+    fail "Multiple benefit_types possible"
```
pkarman: alternately we could just remove this method altogether -- I left it here to help id code paths that depend on it.

app/models/request_issue.rb
```diff
@@ -125,6 +125,7 @@ def attributes_from_intake_data(data)
         unidentified_issue_text: data[:is_unidentified] ? data[:decision_text] : nil,
         decision_date: data[:decision_date],
         issue_category: data[:issue_category],
+        benefit_type: data[:benefit_type] || "compensation",
```
leikkisa: I think if we add the benefit type selection, it should be required instead of defaulted to "compensation", or left blank, and in the case of HLR or SC I'm guessing it gets added elsewhere from the review.  

Maybe this is here to say that if it's an issue from the ratings issues modal, it should have a benefit type of "compensation", but that modal also shows the decision_issues which could have different benefit types :-/.

app/models/decision_review.rb
```diff
@@ -62,7 +62,7 @@ def caseflow_only?
 
   def serialized_ratings
     return unless receipt_date
-    return if caseflow_only?
+    return if caseflow_only? && include_ratings_based_on_benefit_type?
```
leikkisa: I think this is getting a little convoluted. I think we're wanting to return ratings if it's an Appeal or benefit type is compensation?  I actually don't know if we want to show ratings for Pension, we should ask.

pseudocode but something more like return unless decision_review_type=='Appeal' || benefit_type == 'compensation'

app/models/request_issue.rb
leikkisa: I guess for DecisionIssues we'd want to maintain the same benefit type they had before.

app/models/appeal.rb
```diff
@@ -207,14 +207,16 @@ def previously_selected_for_quality_review
...
+  end
+
+  def caseflow_only?
```
jiahuang: I realize you probably named it this as a result of the other `caseflow_only` function in decision review, but maybe both could be named better? Something like `compensation_benefit_types_only?`

spec/feature/intake/supplemental_claim_spec.rb
```diff
@@ -854,7 +854,6 @@ def check_row(label, text)
           expect(page).to have_content("Does issue 3 match any of these VACOLS issues?")
 
           add_intake_rating_issue("None of these match")
-          add_untimely_exemption_response("Yes")
```
pkarman: this should not have been here. untimely exemption does not apply to Supplemental Claim

app/models/appeal.rb
pkarman: yes I think it needs to be renamed. it's really like `effectuated_in_vbms?`

client/app/intake/components/NonratingRequestIssueModal.jsx
```diff
@@ -113,53 +121,51 @@ class NonratingRequestIssueModal extends React.Component {
     let receiptDate = new Date(this.props.intakeData.receiptDate);
     let isTimely = (receiptDate - decisionDate) <= ONE_YEAR_PLUS_MS;
 
-    return !isTimely;
+    return isTimely;
```
jiahuang: nit: can this variable be named something different from the function name?

app/models/appeal.rb
pkarman: I will rename and make the concern we discussed for BoardGrantEffectuation since they both rely on benefit_type

client/app/intake/components/NonratingRequestIssueModal.jsx
```diff
@@ -113,53 +121,51 @@ class NonratingRequestIssueModal extends React.Component {
...
-      this.props.addNonratingRequestIssue({
-        category: this.state.category.value,
-        description: this.state.description,
-        decisionDate: this.state.decisionDate,
-        ineligibleDueToId: this.state.ineligibleDueToId,
-        ineligibleReason: this.state.ineligibleReason,
-        reviewRequestTitle: this.state.reviewRequestTitle,
-        timely: true
-      });
+      this.props.addNonratingRequestIssue(currentIssue);
```
jiahuang: 💯

pkarman:
thanks @leikkisa and @jiahuang 

I renamed `caseflow_only?` and split it into 2 methods `effectuated_in_caseflow?` and `benefit_type_requires_payee_code?` to distinguish the 2 ways it was being used.

per @leikkisa comment:

> I think this doesn't handle benefit_type for DecisionIssues

the ticket was limited to the nonrating modal only, as afaik that would exclude dealing with DecisionIssues. Maybe you meant on editing an Appeal?

app/models/claim_review.rb
```diff
@@ -70,12 +70,9 @@ def create_decision_review_task!
   end
 
   def business_line
-    return unless caseflow_only?
+    return if benefit_type_requires_payee_code?
```
leikkisa: I don't think you need to have this here, the business line will still return correctly if the benefit type requires a payee code.  The logic for whether a task get's created checks that it's effectuated in Caseflow.

app/models/appeal.rb
leikkisa: yay, the new naming is clearer to me

app/models/appeal.rb
shanear: I think it's probably fine, a more descriptive failure might be good if we're going to do that then

app/models/appeal.rb
```diff
@@ -297,4 +299,14 @@ def root_task
   def bgs
     BGSService.new
   end
+
+  def contestable_decision_issues
+    return [] unless receipt_date
+
+    DecisionIssue.where(participant_id: veteran.participant_id).select do |issue|
+      next if issue.decision_review.is_a?(Appeal) && !issue.decision_review.outcoded?
```
shanear: I feel like this should just be one method on `DecisionIssue`, since we may need it in the future.

maybe `issue.final?` or `issue.finalized?`

app/models/board_grant_effectuation.rb
```diff
@@ -2,6 +2,8 @@
 # Some are represented as contentions on an EP in VBMS. Others are tracked via Caseflow tasks.
 
 class BoardGrantEffectuation < ApplicationRecord
+  include Benefitable
```
shanear: I like this because it's railsy haha, but i think it's not super clear. Maybe `HasBenefitType` or `HasBusinessLine`?

app/models/concerns/benefitable.rb
```diff
@@ -0,0 +1,20 @@
+module Benefitable
+  extend ActiveSupport::Concern
+
+  def business_line
+    business_line_name = Constants::BENEFIT_TYPES[benefit_type]
+    @business_line ||= BusinessLine.find_or_create_by(url: benefit_type, name: business_line_name)
+  end
+
+  def effectuated_in_vbms?
```
shanear: I would rename this to `processed_in_vbms?` since we're making it more generic. `effectuated` is a term we use to describe the board grant effectuation process only. It'd be confusing otherwise.

app/models/decision_review.rb
```diff
@@ -56,13 +57,9 @@ def review_title
...
   end
 
-  def caseflow_only?
-    !ClaimantValidator::BENEFIT_TYPE_REQUIRES_PAYEE_CODE.include?(benefit_type)
-  end
-
   def serialized_ratings
     return unless receipt_date
-    return if caseflow_only?
+    return unless benefit_type_requires_payee_code? || is_a?(Appeal)
```
shanear: yeah this is incorrect. We want to return ratings for appeals.

Also, I'm not super fond of using `benefit_type_requires_payee_code?` where we aren't directly asking whether the benefit type requires the payee code. I much prefer `processed_in_caseflow?` here.

app/models/decision_review_intake.rb
```diff
@@ -5,7 +5,7 @@ def ui_hash(ama_enabled)
       claimant: detail.claimant_participant_id,
       veteran_is_not_claimant: detail.veteran_is_not_claimant,
       payeeCode: detail.payee_code,
-      nonComp: detail.caseflow_only?,
+      nonComp: !detail.benefit_type_requires_payee_code?,
```
shanear: I know you didn't do this, but I'm not a fan of using `nonComp`, since it's not technically correct (it's non comp & non pension)

spec/feature/intake/appeal_spec.rb
```diff
@@ -683,6 +683,33 @@ def check_row(label, text)
...
+    expect(page).to have_content("Description for Accrued")
+
+    click_intake_finish
+
+    expect(page).to have_content("Intake completed")
+
+    request_issue = intake.reload.detail.request_issues.first
+
+    expect(request_issue.description).to eq("Accrued - Description for Accrued")
+    expect(request_issue.benefit_type).to eq("education")
```
shanear: It might be nice to add an issue of another business line and test that it has a different benefit type

shanear: This actually reminds me, we are going to have to set the benefit type even for rating issues 😢 which can be compensation or pension depending on the rating issue. Non necessary in this PR, lets just set them all to "compensation", but we will need to do that.

spec/feature/intake/appeal_spec.rb
```diff
@@ -683,6 +683,33 @@ def check_row(label, text)
...
+    expect(page).to have_content("Add / Remove Issues")
+
+    click_intake_add_issue
+    click_intake_no_matching_issues
+    add_intake_nonrating_issue(
+      benefit_type: "Education",
+      category: "Accrued",
+      description: "Description for Accrued",
+      date: profile_date.strftime("%D")
+    )
```
shanear: we also may want to add a test that validates that the issue categories change when you change the benefit type

app/models/appeal.rb
pkarman: IIUC, if the `decision_review` isa Appeal and is outcoded, then the `DecisionIssue` is finalized.

OTOH, if the `decision_review` is not an Appeal, under what conditions is the `DecisionIssue` finalized?

spec/feature/intake/appeal_spec.rb
pkarman: I agree that would be a good test. However, our use of react-select library makes it tricky since we do not have "normal" `<select>` element to check. I have verified it works manually as expected. Should I spend more time on this now, or open a backlog issue to write tests for this?

app/models/decision_review_intake.rb
pkarman: agreed. var name changed.

spec/feature/intake/appeal_spec.rb
leikkisa: Yeah, and in that PR we might want to confirm that it's being set for DecisionIssues too for all of the LOB.

app/models/decision_review.rb
pkarman: new method `eligible_for_serialized_ratings?` allows per-class control (which we need because not all DecisionReview subclasses act the same way).

app/models/appeal.rb
```diff
@@ -109,6 +109,17 @@ def attorney_case_reviews
     tasks.map(&:attorney_case_reviews).flatten
   end
 
+  # the processed_in_* logic is per-RequestIssue for Appeals,
+  # so we always say "false" at this level and allow each RequestIssue
+  # to further specify.
+  def processed_in_caseflow?
+    false
+  end
```
shanear: this is sort of awkward, since appeals are processed in caseflow

app/models/appeal.rb
```diff
@@ -215,14 +226,12 @@ def previously_selected_for_quality_review
   end
 
   def benefit_type
-    # temporary until ticket for appeals benefit type by issue is implemented
-    # https://github.com/department-of-veterans-affairs/caseflow/issues/5882
-    "compensation"
+    fail "benefit_type on Appeal is set per RequestIssue"
```
shanear: love this

app/models/appeal.rb
```diff
@@ -305,4 +314,19 @@ def root_task
   def bgs
     BGSService.new
   end
+
+  # we always want to show ratings on intake
+  def eligible_for_serialized_ratings?
```
shanear: This is not really what we are asking for. I feel like something like these would be more accurate:

`rating_issues_contestable?`
`eligible_to_contest_ratings?`
`can_contest_rating_issues?`

app/models/appeal.rb
```diff
@@ -305,4 +314,19 @@ def root_task
...
+    true
+  end
+
+  def contestable_decision_issues
+    return [] unless receipt_date
+
+    DecisionIssue.where(participant_id: veteran.participant_id)
+      .select(&:finalized?)
```
shanear: nice

app/models/concerns/has_business_line.rb
```diff
@@ -0,0 +1,22 @@
+module HasBusinessLine
```
shanear: Love this

app/models/decision_issue.rb
```diff
@@ -38,6 +38,14 @@ def rating?
     request_issues.none?(&:nonrating?)
   end
 
+  def finalized?
+    if appeal?
+      decision_review.outcoded?
+    else
+      disposition.present?
+    end
```
shanear: i prefer the ternary operator, but i suppose this is fine too 😝

app/models/request_issue.rb
```diff
@@ -140,6 +143,10 @@ def attributes_from_intake_data(data)
     # rubocop:enable Metrics/MethodLength
   end
 
+  def set_default_attributes
+    self.benefit_type ||= "compensation" # must match default value in db
```
shanear: why is this here. Seems like we'd be in danger of falsely setting a benefit type as compensation. I feel like we should always explicitly set this, or at least default to the benefit_type of it's decision_review

spec/feature/intake/appeal_spec.rb
shanear: no worries if it's too tricky

app/models/appeal.rb
pkarman: right. the correct answer is "maybe" not false or true, since we don't know, at the Appeal level, whether any of its request_issues will be processed in caseflow or not.

The problem here is parity in naming between `ClaimReview` and `Appeal`, since they share lots of logic via `DecisionReview` but don't always share a data model. In some ways it feels like all the `DecisionReview` shared code would have been better implemented as concerns to mix in.

shanear:
looks great

PR #8579, opened 2019-01-08
Read from new rating issue description columns

### LINKED ISSUE #8518: VBMS decisions with no rating issue break DTA contentions

shanear:
yay for renaming things. 😭

This PR will be merged after #8549 is deployed to prod and the following migration script is run:

```
RequestIssue.all.each do |r|
  if r.contested_rating_issue_reference_id
    r.update!(contested_issue_description: r.description)
  elsif r.issue_category
    r.update!(nonrating_issue_description: r.description)
  elsif r.is_unidentified
    r.update!(unidentified_issue_text: r.description)
  else
    raise "bad issue found #{r.id}"
  end
end
```

This also fixes #8518 and an unrecorded issue where non rating request issues weren't displaying properly on queue (re: @aroltsch)

### DIFF
--- a/app/models/decision_issue.rb
+++ b/app/models/decision_issue.rb
@@ -10,6 +10,9 @@ class DecisionIssue < ApplicationRecord
   belongs_to :decision_review, polymorphic: true
   has_one :effectuation, class_name: "BoardGrantEffectuation", foreign_key: :granted_decision_issue_id
 
+  # Attorneys will be entering in a description of the decision manually for appeals
+  before_save :calculate_and_set_description, unless: :appeal?
+
   def self.granted
     # TODO: "allowed" is the disposition for BVA grants, not necessarily the disposition for granted HLRs and SCs
     #       we need to add that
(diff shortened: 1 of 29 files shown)


app/models/decision_issue.rb
```diff
@@ -43,20 +40,21 @@ def rating?
...
+    self.description ||= calculate_description
   end
 
-  def nonrating_description
-    "#{disposition}: #{issue_category} - #{associated_request_issue.description}"
+  def calculate_description
+    return decision_text if decision_text
+    return nil unless associated_request_issue
+
+    "#{disposition}: #{associated_request_issue.description}"
```
leikkisa: This reminds me, this is where I got confused before (between DecisionIssue description and decision_text).  I think I understand it now though.

Would we ever want this to include info from all of the request issues?

PR #8601, opened 2019-01-10
assign ColocatedTasks w/auto child task creation

### LINKED ISSUE #8532: Rely on automatic child task creation when assigning ColocatedTasks
#8489 added the ability for tasks to [automatically create a child task when they are created](https://github.com/department-of-veterans-affairs/caseflow/blob/9adda2bd49f033c04f459c847d378f7fdbdbfb1c/app/models/task.rb#L13) so that we don't have to write that code for each task class that wants to implement this functionality. This ticket exists to use the new pattern established by #8489 to replace `ColocatedTask`s [existing logic that assigns a task to a member](https://github.com/department-of-veterans-affairs/caseflow/blob/9adda2bd49f033c04f459c847d378f7fdbdbfb1c/app/models/tasks/colocated_task.rb#L12) of the `Colocated` team.

## Acceptance criteria
* [ ] `ColocatedTask` no longer explicitly create a task assigned to an individual user when the parent organization task is created.

tomas-nava:
Resolves #8532


app/controllers/tasks_controller.rb
```diff
@@ -60,6 +60,8 @@ def create
     tasks_to_return = (queue_class.new(user: current_user).tasks + tasks).uniq
 
     render json: { tasks: json_tasks(tasks_to_return) }
+  rescue ActiveRecord::RecordInvalid => error
+    render json: { errors: [title: error.class.to_s, detail: error.message] }, status: :bad_request
```
tomas-nava: Rescuing errors generated by the change from `create` to `create!` in `Task.create_from_params`; is this the right pattern to follow?

app/models/task.rb
```diff
@@ -88,7 +88,7 @@ def self.create_many_from_params(params_array, current_user)
   def self.create_from_params(params, user)
     verify_user_can_create!(user)
     params = modify_params(params)
-    create(params)
+    create!(params)
```
tomas-nava: Are there any consequences of this change I'm not thinking of?

app/controllers/tasks_controller.rb
lowellrex: We were [previously returning an `invalid_record_error`](https://github.com/department-of-veterans-affairs/caseflow/pull/8601/files#diff-f4416012d567d02da854cfbc9e2df2caR58) if the task was invalid so we can just return that here. Additionally, since the task can no longer be invalid (if it is we will throw this error), we can get rid of the `...unless task.valid` condition on line 58.

app/models/task.rb
lowellrex: The only thing I can think of is the `valid?` checks but I think those are covered by other changes in this PR. I love this change!

app/models/colocated_task_distributor.rb
```diff
@@ -3,4 +3,14 @@ def initialize(list_of_assignees: Colocated.singleton.non_admins.sort_by(&:id).p
                  task_class: Task)
     super
   end
+
+  def next_assignee(_task_class = nil, appeal = nil)
```
lowellrex: Do we need to change the function signature to expect these two parameters in the [parent class](https://github.com/department-of-veterans-affairs/caseflow/pull/8601/files#diff-eaffcb231bed7d0bddf84ac52f858285L33) as well?

app/models/organizations/colocated.rb
```diff
@@ -3,7 +3,7 @@ def self.singleton
     Colocated.first || Colocated.create(name: "VLJ Support Staff", url: "vlj-support")
   end
 
-  def next_assignee(_task_class)
-    ColocatedTaskDistributor.new.next_assignee
+  def next_assignee(task_class = nil, appeal = nil)
```
lowellrex: Do we need to update the base class (`Organization`) to have the same function signature as well?

lowellrex: Also, while we're doing that let's get rid of the `task_class` argument. We don't use it yet and I think it is confusing that it is included and yet we don't use it.

app/models/tasks/colocated_task.rb
```diff
@@ -12,34 +12,24 @@ class << self
...
+        all_tasks
+      end
     end
   end
 
-  def automatically_assign_org_task?
-    false
+  def set_assigned_at_and_update_parent_status
+    self.assigned_at = created_at unless assigned_at
+    parent&.update(status: :on_hold)
```
lowellrex: Let's make this change to the base class' version of this method.

| . | VACOLS tasks | Caseflow tasks | 
| --- | --- | --- |
| Legacy Appeals | ✅ | ✅ |
| AMA appeals | 🚫 | ✅ |

The existing code assumes that we cannot have caseflow tasks for legacy appeals, but that is incorrect. Checking for the existence of the parent is the perfect way to handle this!

app/models/tasks/colocated_task.rb
```diff
@@ -84,6 +74,14 @@ def assign_to_privacy_team_data
 
   private
 
+  def create_and_auto_assign_child_task
+    dup.tap do |child_task|
+      child_task.assigned_to = assigned_to.next_assignee(self.class, appeal)
```
lowellrex: Is the only difference between this method and the base class' version the inclusion of the `appeal` parameter?

app/models/colocated_task_distributor.rb
```diff
@@ -3,4 +3,14 @@ def initialize(list_of_assignees: Colocated.singleton.non_admins.sort_by(&:id).p
...
+    open_assignee = appeal
+      &.tasks
+      &.where&.not(status: Constants.TASK_STATUSES.completed)
+      &.where(assigned_to_type: "User")
+      &.find_by(assigned_to_id: User.where(css_id: list_of_assignees).map(&:id))
+      &.assigned_to
```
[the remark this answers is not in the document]
tomas-nava: yep, I think that's more human-readable, and the generated query is similarly efficient.

PR #8609, opened 2019-01-10
Hearing Worksheet | Case Details Link

### LINKED ISSUE #8592: Hearing Worksheet | Case Details Link
On the hearing worksheet, for both legacy and ama appeals, we need to add a button in the top righthand corner that links to the Case Details page. We should use the Link component stylized as a button.

screen shot 2019-01-09 at 3 26 44 pm

Sjones352:
Resolves #8592

### Description
This pr adds a Case detail  link on the Hearing worksheet 

### Acceptance Criteria
- [ ] On the hearing worksheet, for both legacy and ama appeals, we need to add a button in the top righthand corner that links to the Case Details page. We should use the Link component stylized as a button.

### Testing Plan
1. Go to Hearing worksheet and see the Case details link.
screen shot 2019-01-10 at 12 03 23 pm

2. Click on the `View case detail` link and see if it takes you to Case details page.


client/app/hearings/components/WorksheetHeaderVeteranSelection.jsx
```diff
@@ -114,6 +114,13 @@ class WorksheetHeaderVeteranSelection extends React.PureComponent {
         />
       </div>
       <div className="cf-push-right">
+        <Link
+          name="view-case-detail"
+          onClick={this.onClickReviewClaimsFolder}
```
sharonwarner: I don't think we need this line since we're using a link instead of a button.

PR #8621, opened 2019-01-10
Update README with note about hot reloading

monfresh:

### DIFF
--- a/README.md
+++ b/README.md
@@ -389,6 +389,38 @@ To run the test suite:
 
     bundle exec rake
 
+### Testing frontend changes in feature specs
+
+Making frontend changes requires Webpack to compile the assets in order for the
+spec to pick up the changes. To automatically compile assets after every change,
+you can turn on Webpack's [Hot Module Replacement](https://webpack.js.org/concepts/hot-module-replacement/)
+by setting the env var `REACT_ON_RAILS_ENV` to `HOT` in your Terminal session.
+
+For example:
+```console
+export REACT_ON_RAILS_ENV=HOT
+bundle exec rspec spec/feature/queue/case_details_spec.rb:350
+```
+or
+```console
+REACT_ON_RAILS_ENV=HOT bundle exec rspec spec/feature/queue/case_details_spec.rb:350
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -389,6 +389,16 @@ To run the test suite:
 
     bundle exec rake
 
+### Testing frontend changes in feature specs
+
+Making frontend changes requires Webpack to compile the assets in order for the
+spec to pick up the changes. To automatically compile assets after every change,
+set the env var `REACT_ON_RAILS_ENV` to `HOT` in your shell profile. For example,
+if you're using Bash, add this to your `.bash_profile`:
```
mdbenjam: I'm a little hesitant to ask people to put this in their bash profile. Maybe start with just saying "to run tests when webpack is running with hot module reloading you can use `REACT_ON_RAILS_ENV=HOT rake`"?

monfresh: Do people run `rake` for tests? What is the downside of setting this in your shell profile?

I run tests directly within Sublime Text, where I can choose to run either one spec or all specs in the current file. If I were to do it from the command line, I would run something like `bundle exec rspec spec/path/to/feature/spec.rb:specific_line_number`

How about we give options? Perhaps something like this: 

Set the env var in your terminal session before running tests. For example:
```console
export REACT_ON_RAILS_ENV=HOT
bundle exec rspec spec/feature/queue/case_details_spec.rb:350
```
or
```console
REACT_ON_RAILS_ENV=HOT bundle exec rspec spec/feature/queue/case_details_spec.rb:350
```
For less typing, create an alias in your shell profile (such as `.bash_profile`):
```
alias ber='REACT_ON_RAILS_ENV=HOT bundle exec rspec'
```
To run a specific test with this alias:
```console
ber spec/feature/queue/case_details_spec.rb:350
```

README.md
```diff
@@ -389,6 +389,32 @@ To run the test suite:
 
     bundle exec rake
 
+### Testing frontend changes in feature specs
+
+Making frontend changes requires Webpack to compile the assets in order for the
```
mdbenjam: Last thing we should just mention that the webpack server needs to be running. Either `foreman start` or `yarn run dev:hot`.

PR #8642, opened 2019-01-11
Fix search test

lowellrex:
This PR ensures the link we click on contains only the `appeal.docket_number` and nothing else. This test was failing occasionally because we were trying to click on the text of the docket number but accidentally clicked on the dropdown menu when the CSS ID contained the `appeal.docket number`.

image

### DIFF
--- a/spec/feature/queue/search_spec.rb
+++ b/spec/feature/queue/search_spec.rb
@@ -255,7 +255,7 @@ RSpec.feature "Search" do
       end
 
       it "clicking on docket number sends us to the case details page" do
-        click_on appeal.docket_number
+        find("a", text: /\A#{appeal.docket_number}\z/).click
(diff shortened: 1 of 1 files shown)


spec/feature/queue/search_spec.rb
```diff
@@ -255,7 +255,7 @@
       end
 
       it "clicking on docket number sends us to the case details page" do
-        click_on appeal.docket_number
+        find("a", text: /\A#{appeal.docket_number}\z/).click
```
pkarman: I think you can also do:

```ruby
find("a", exact_text: appeal.docket_number).click
```

which might be slightly easier to read. See https://www.rubydoc.info/github/jnicklas/capybara/Capybara/Node/Finders#find-instance_method

lowellrex: Oh that's much better!

PR #8662, opened 2019-01-11
new_documents endpoint without VBMS calls

hschallhorn:
References #8380 

### Acceptance Criteria
- [x] Calls to new_documents to not make calls to vacols or vbms
- [x] new_documents returns the correct number of documents


app/models/legacy_appeal.rb
```diff
@@ -145,6 +145,20 @@ def number_of_documents_from_caseflow
     (count != 0) ? count : number_of_documents
   end
 
+  def new_documents_from_caseflow(user)
```
aroltsch: I think we should move these two methods into a concern so they are in one place and we don't duplicate code. We can name it `DocumentConcern`, we can also move `documents_from_caseflow` method there.

hschallhorn: Letsssss do it. Also, I assume you mean `number_of_documents_from_caseflow`.

aroltsch: Yes! You know what I mean!

PR #8663, opened 2019-01-11
Do no throw issue repository errors to sentry

### LINKED ISSUE #8204: Do not throw Sentry error when issue is not found
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2365/

image

image

### LINKED ISSUE #8567: Do not send this exception to Sentry - Cannot find issue with vacols ID
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3284/

aroltsch:
Resolves #8204

### Description
image

This error doesn't happen very often. I was not able to re-produce it using the normal flow. I was only able to re-produce it if I had two open tabs.


lowellrex:
Why do we want to silence these errors? It seems to me like they are identifying a bug in our code. From what I can tell, the call stack (in reverse) when we hit our error looks like this:

* `CaseReviewsController.complete`
    *  Accepts parameters from the request that [includes an array of issues hashes](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/case_reviews_controller.rb#L71)
* `AttorneyCaseReview.complete`
* `AttorneyCaseReview.update_in_vacols!`
* `IssueUpdater.update_issue_dispositions_in_vacols!`
    * [We loop over each element](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/concerns/issue_updater.rb#L26) of that array of issues hashes here
* `Issue.update_in_vacols!`
* `IssueRepository.update_vacols_issue!`
* `IssueRepository.validate_issue_presence!`
* `VACOLS::CaseIssue.find_by`
    * We fail to find a `VACOLS::CaseIssue` here because the issue ID does not exist in the database

It seems to me like the request from the frontend is bad (perhaps because we didn't update the global state on the frontend after an API response that changed issues? [This may be an example of that case](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3253/events/215108/)) and we should probably be attempting to correct the request we send instead of being more lenient with the input on the backend.

app/controllers/issues_controller.rb
```diff
@@ -3,11 +3,10 @@ class IssuesController < ApplicationController
 
   rescue_from ActiveRecord::RecordInvalid do |e|
     Rails.logger.error "IssuesController failed: #{e.message}"
-    Raven.capture_exception(e)
     render json: { "errors": ["title": e.class.to_s, "detail": e.message] }, status: :bad_request
   end
 
-  rescue_from Caseflow::Error::UserRepositoryError do |e|
+  rescue_from Caseflow::Error::VacolsRepositoryError do |e|
```
lowellrex: It looks like this change makes it so that we are rescuing from [`VacolsRecordNotFound`, `RemandReasonRepositoryError`, `QueueRepositoryError`, and `MissingRequiredFieldError`](https://github.com/department-of-veterans-affairs/caseflow/blob/2500ced5f71d6258fc6d61ccf200a45762b2b219/lib/caseflow/error.rb#L185) in addition to the `UserRepositoryError` we used to rescue from here and the `IssueRepositoryError` this PR is specifically targeting. Why do we want to rescue from all VACOLS repository errors instead of just those two?

aroltsch: The rest of the errors are not thrown in the code path for the issue controller. I can be more specific though and catch just these two.

Issue #7997, opened 2018-11-27
Non-comp Flow | Review page filter by type


shanear:
It's useful for non-comp users to be able to filter their list of reviews by type.

Full non-comp flow: https://va-gov.invisionapp.com/share/PFPCPJ7MEGC

## Acceptance Criteria
- On both the "In Progress" and "Completed" tabs on the non comp reviews page, there should be a filter icon (see the mocks below). When clicked it should show a filter checklist.
- If any item on the check list is checked, the list should be filtered to only the types that are checked (i.e. if "Supplemental Claim" and "Higher Level Review" are checked, it should only show tasks of that type)
- If "Clear category filter" is clicked, it should remove all checked filters.

## Notes
The filter icon and functionality is also used in Reader

requires #7990
requires #7996

carolaponce:
Updated comp

Changing the Intake form button to a main button style.
Changing text to:
Review each issue and select a disposition.
Adding - to HLR dropdown

![returning disposition main copy](https://user-images.githubusercontent.com/11700545/49257592-9d7cd900-f3e7-11e8-87cf-34ada5d480e1.png)

carolaponce:
Final mock.  Also deleting previous to avoid any confusion.

![returning disposition main copy](https://user-images.githubusercontent.com/11700545/49620664-c7e60d80-f976-11e8-9597-2693bfed38eb.png)

steve-gov:
Looks like this in UAT (Chrome 70 Mac), no checkboxes and weirdly large font:

<img width="900" alt="screen shot 2019-01-08 at 11 57 36 am" src="https://user-images.githubusercontent.com/25468689/50846236-acd33880-133c-11e9-8408-8ac24668d2e7.png">

steve-gov:
Slight padding issues, but gonna clear it:

<img width="339" alt="screen shot 2019-01-15 at 10 57 19 am" src="https://user-images.githubusercontent.com/25468689/51202773-5cb42300-18b4-11e9-96f8-49bece51081f.png">

shanear:
Some tweaks:
<img width="1029" alt="screen shot 2019-01-15 at 2 01 57 pm 1" src="https://user-images.githubusercontent.com/1596259/51217307-62bdfa00-18dc-11e9-8ce7-1005c29e448b.png">

1. if there is no EP, we should not say "null" in EP Code
1. if the EP is not yet established the status should read "Establishing..."
[...]

PR #8671, opened 2019-01-12
fix type filter styling

jiahuang:
Connects #7997

* Added a new list picker component that includes a checkbox

filter


client/app/components/ListItemPickerCheckbox.jsx
```diff
@@ -0,0 +1,57 @@
+// similar to ListItemPicker only this saves
+// the state of the current selected item and shows it as a checkbox
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css } from 'glamor';
+
+import Checkbox from './Checkbox';
+// import update from 'immutability-helper';
```
leikkisa: Leftover?

PR #8687, opened 2019-01-14
Fixing nil return

mdbenjam:
We were returning nil: https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3346/

### Description
We returned from a map thinking that would return from the map and not the function. So we were returning nil from the function instead of an array with fewer elements.

### Testing Plan
1. Created a new test for this scenario.

### DIFF
--- a/app/models/hearing_day.rb
+++ b/app/models/hearing_day.rb
@@ -109,9 +109,11 @@ class HearingDay < ApplicationRecord
       regional_office_keys = total_video_and_co.map { |hearing_day| hearing_day[:regional_office] }
       regional_office_hash = HearingDayRepository.ro_staff_hash(regional_office_keys)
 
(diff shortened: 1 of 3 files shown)


app/models/hearing_day.rb
```diff
@@ -119,11 +119,13 @@ def load_days_with_open_hearing_slots(start_date, end_date, regional_office = ni
           regional_office_hash[hearing_day[:regional_office]], hearing_day
         )
 
-        return nil if scheduled_hearings.length >= total_slots || hearing_day[:lock]
```
mdbenjam: `return` returns from the whole function not the map... I didnt know that, I can't believe I haven't made this mistake elsewhere (or maybe I did! 😬 )

sharonwarner: Ah I see.

Issue #8299, opened 2018-12-14
Appeal Outcoding | Non-comp effectuation task page


shanear:
If a board grant is made on a decision issue that is not compensation or pension, the effectuation will not be tracked as part of an EP, but as part of a Caseflow task in the non comp queue described in #7990. The effectuation of these granted issues will need to be recorded via this page.

requires #8283 

## Acceptance Criteria
- When a user clicks on a "Board Grant" task on the non-comp landing page, it should take them to the "effectuation" page (see the mock below)
- The page should only show granted issues for the business line the task was clicked from.
- The "Cancel" button should send the user back to the non-comp landing page.
- When on the effectuation page, if the "I certify these benefits have been effectuated" checkbox is clicked the "Confirm" button should become active.
- When the "Confirm" button is clicked. The task should be completed, which should record the time it was completed and the user who completed it. It should move the task from the "In Progress" tab to the "Completed" tab. 
- When the "Confirm" button is clicked, the user should be redirected back to the non-comp landing page.

## Mocks
![effectuation](https://user-images.githubusercontent.com/1596259/50025016-10b24d80-ff99-11e8-826a-5d961e3c5e01.png)

kencheeto:
@shanear i added `When the "Confirm" button is clicked, the user should be redirected back to the non-comp landing page.` but lmk if that is wrong

steve-gov:
Will validate after #8283 validates and one of its issues is outcoded as a Board Grant.

steve-gov:
Outcoding successfully created the effectuation tasks. 
- Acceptance criteria 1, 3, and 4 validated.
- Tasks from two separate business lines appeared on the same effectuation task page. The process should have created two separate tasks/ task pages, one for each business line.
- Saving fails with the attached error.
- The failed-to-save error reads `The dispositions for this task could not be saved` but should read `The effectuation confirmation(s) for this task could not be saved`

steve-gov:
<img width="830" alt="screen shot 2019-01-28 at 11 59 07 am" src="https://user-images.githubusercontent.com/25468689/51863218-4fa42480-22f5-11e9-9762-457d32b10598.png">

![screencapture1](https://user-images.githubusercontent.com/25468689/51863222-52067e80-22f5-11e9-9304-180af5a4df12.gif)

steve-gov:
Awesome! Validated!

[...]

Issue #8031, opened 2018-11-29
Non-comp Flow | Save decision


shanear:
For each non-comp HLR and SC, the users need a way to record the decision being made for each request issue.

## Acceptance Criteria
- On the non-comp decision page, the "Complete" button should be disabled unless a disposition is set for each issue, and the decision date is filled in.
- When the "Complete" button is active and pressed, it should save decision issues for each decision as well as the decision date and the user who saved the decision. The review should be considered complete.
- After saving, it should redirect to the "Completed" tab of the reviews page. It should show a small confirmation message according to the mocks below.

## Copy
Decision completed
You successfully added a decision for [claimant name]'s [form name] (Veteran SSN: [veteran ssn])

## Mocks
*Decision page*
![disposition main - hlr](https://user-images.githubusercontent.com/1596259/49190856-17f31d80-f342-11e8-85e9-ee8541a8e620.png)

*Confirmation after save*
![confirmation - disposition - ftu](https://user-images.githubusercontent.com/1596259/49191538-ed569400-f344-11e8-8ccd-a46b10d3c8e5.png)

Full non-comp flow: https://va-gov.invisionapp.com/share/PFPCPJ7MEGC

requires #7994

carolaponce:
![confirmation - disposition copy](https://user-images.githubusercontent.com/11700545/49251988-b8475180-f3d7-11e8-845c-802356792c64.png)

carolaponce:
![disposition main - hlr](https://user-images.githubusercontent.com/11700545/49252046-da40d400-f3d7-11e8-8e4a-596b708ad994.png)

carolaponce:
![disposition main - hlr](https://user-images.githubusercontent.com/11700545/49252397-ca75bf80-f3d8-11e8-9415-10d931e0a7ee.png)

jiahuang:
Doh, there's a bug where upon completion the `Date Sent` field isn't populated properly. I will fix this in a subsequent PR

steve-gov:
Seems like there are a couple ways the button-disabling can be fooled:

- I can select a disposition, then clear it, and can push the button
- I can clear the date, enter a date, clear the date, and push the button

Button should only be enabled if the date is present and each disposition is selected.

![screencapture](https://user-images.githubusercontent.com/25468689/51646645-45b5a680-1f2e-11e9-8a00-de2e1208b0a3.gif)

steve-gov:
[...]

PR #8697, opened 2019-01-14
Disabling tasks when they are completed

jiahuang:
Connects #8031 
Connects #8299

After tasks are completed, they become uneditable & the complete button goes away

Dispositions:
image

Board grants
image


client/app/nonComp/components/Disposition.jsx
```diff
@@ -131,9 +134,24 @@ class NonCompDispositions extends React.PureComponent {
...
+      decisionIssuesStatus,
+      task
     } = this.props;
 
+    let completeDiv = null;
+
+    if (!task.completed_at) {
+      completeDiv = <React.Fragment>
+        <div className="cf-txt-r">
+          <a className="cf-cancel-link" href={`/decision_reviews/${businessLineUrl}`}>Cancel</a>
```
pkarman: there will be a new method in #8786 called `BusinessLine.tasks_url` that you can call instead of hardcoding, but that's only in the model right now. Maybe we should add to the serializer so the frontend can use?

spec/feature/non_comp/board_grants_spec.rb
```diff
@@ -86,6 +86,13 @@ def submit_form
     in_progress_task.reload
     expect(in_progress_task.status).to eq("completed")
     expect(in_progress_task.completed_at).to eq(Time.zone.now)
+
+    # click on completed task and verify that it is not editable
+    click_link "Tom Brady"
```
pkarman: this feels brittle. Could we click link by cell number or `in_progress_task.ui_hash[:claimant_name]`?

spec/feature/non_comp/dispositions_spec.rb
```diff
@@ -102,6 +107,16 @@ def fill_in_disposition(num, disposition, description = nil)
       expect(hlr.decision_issues.find_by(disposition: "Granted", description: nil)).to_not be_nil
       expect(hlr.decision_issues.find_by(disposition: "Granted", description: "test description")).to_not be_nil
       expect(hlr.decision_issues.find_by(disposition: "Denied", description: "denied")).to_not be_nil
+
+      # verify that going to the completed task does not allow edits
+      click_link "Bob Smith"
```
pkarman: same comment about claimant name hardcoded in tests.

app/models/tasks/board_grant_effectuation_task.rb
```diff
@@ -12,6 +12,18 @@ def ui_hash
...
 
   def complete_with_payload!(_decision_issue_params, _decision_date)
+    return false unless validate
+
     update!(status: Constants.TASK_STATUSES.completed, completed_at: Time.zone.now)
   end
+
+  private
+
+  def validate
```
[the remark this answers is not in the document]
jiahuang: nice catch, i'll change the name

app/models/serializers/work_queue/decision_review_task_serializer.rb
```diff
@@ -37,6 +37,10 @@ def claimant_relationship
     }
   end
 
+  attribute :tasks_url do
+    task.assigned_to.tasks_url
```
pkarman: nice!

Issue #8338, opened 2018-12-18
Shepherd appeals from intake to auto-case distribution


amprokop:
### Creating IHP, evidence submission, and hearing tasks
 - [X] Create IHP and evidence submission window tasks on intake for direct docket and evidence submission appeals https://github.com/department-of-veterans-affairs/caseflow/issues/8334 https://github.com/department-of-veterans-affairs/caseflow/issues/7916
 - [X] Create task timing functionality days and make evidence submission tasks expire in 90 days https://github.com/department-of-veterans-affairs/caseflow/issues/8684
  - [X] Create `ready` scope, usable by auto-case distribution, that can find all appeals that are ready for distribution 
- [x] Track when an appeal became ready for distribution
 - [x] Create `ScheduleHearing` tasks, `HoldHearing` tasks, and `TranscribeHearing` tasks (@sharonwarner to coordinate completion)
 - [x] Ensure hearing tasks expire as necessary, ensure they are returned by the ready scope (@sharonwarner to coordinate completion)

### Blocking distribution of appeals with open mail tasks
- [X] When a mail task for an undistributed appeal is created, we should potentially create it as a child of the Distribution task, thus ensuring that a distribution will not happen if there is an open mail task 

### Handling POA changes
We will check POA and create IHP tasks as late as possible:
 - [X] for direct review appeals, we check POA and create necessary IHP tasks that block distribution when the appeal is intook
 - [X] for evidence submission appeals, we check POA and create necessary IHP tasks after the evidence submission window ends or is waived (so if POA changes during that window, we'll handle it correctly
- [x] for hearing appeals, check POA and create necessary IHP tasks (at least for PVA, which is apparently the only IHP-writing VSO for hearing appeals) **after** the hearing is held and evidence is submitted
 ^ The above tasks will ensure that we correctly handle a POA change before the IHP period starts. However, if POA changes after the IHP period would have started, we don't currently have a solution.
- [x] handle the case when POA changes *after* the IHP period starts 

### Supporting waiving of evidence submission period
- [ ] complete the evidence submission task early, opening up the appeal for IHP writing or distribution before the period is over
- [ ] ensure we can track any necessary side information about the waiver

### Supporting waiving of hearing
- [x] When the hearing appeal is ready to be scheduled, check if the hearing is withdrawn. If it is, don't create further hearing tasks and mark appeal as ready for distribution

### Resilience and error handling
[...]

PR #8698, opened 2019-01-14
adds ready scope

amprokop:
Related to https://github.com/department-of-veterans-affairs/caseflow/issues/8338


app/models/tasks/root_task.rb
```diff
@@ -70,7 +71,10 @@ def create_subtasks!(appeal, parent)
         if appeal.evidence_submission_docket?
           create_evidence_submission_task!(appeal, distribution_task)
         else
```
cmgiven: Could this be an else if just to make clear that this applies only to the direct review docket?

amprokop: Yep definitely. I'll update that when I add the hearing part next.

Issue #8695, opened 2019-01-14
Intake | Add limit for contention text


shanear:
There is a BGS enforced character limit for contention text of 255 characters. We need to build a way that shortens contention text of the contentions that we create if they are too long.

pkarman:
Do we rely on this text for matching caseflow records at all?

One pattern I have seen before is to truncate text to something like 239 bytes, and save the last 16 bytes for a checksum (first part of a md5 or sha digest) of the entire untruncated text, so that matching can still work later.

shanear:
![ss6](https://user-images.githubusercontent.com/1596259/51627237-55b19400-1ef6-11e9-8dc4-22b4e43d1d26.PNG)
boom shakalaka

PR #8705, opened 2019-01-14
Limit contention text to 255 characters

leikkisa:
connects #8695 

We want to be able to maintain having long descriptions within Caseflow data, but contention_text to 255 characters for BGS


app/models/board_grant_effectuation.rb
```diff
@@ -19,9 +19,7 @@ class BoardGrantEffectuation < ApplicationRecord
     pension_nonrating: "030BGNRPMC"
   }.freeze
 
-  def contention_text
-    granted_decision_issue.description
-  end
+  delegate :contention_text, to: :granted_decision_issue
```
shanear: nice 🎇

app/models/contention.rb
```diff
@@ -0,0 +1,14 @@
+class Contention
```
pkarman: I like this pattern.

FWIW, in other Rails apps you might see this as a Decorator rather than a Model. See https://robots.thoughtbot.com/evaluating-alternative-decorator-implementations-in

Since we don't have that pattern, it's fine with me to leave in the Model space.

spec/models/ramp_refiling_spec.rb
```diff
@@ -182,14 +190,19 @@
         expect(Fakes::VBMSService).to have_received(:create_contentions!).with(
           veteran_file_number: "64205555",
           claim_id: "1337",
-          contentions: [{ description: "Arm" }, { description: "Arm" }, { description: "Leg" }],
+          contentions: [
+            { description: "Arm" },
+            { description: "Arm" },
+            { description: "Leg" },
+            { description: "#{('Long Description' * 20).slice(0, 252)}..." }
```
pkarman: alternately, `"x" * 252` might be a tad easier to read.

app/models/contention.rb
leikkisa: Cool, I was wondering about how this would usually be approached. Thanks!

PR #8706, opened 2019-01-14
Sync board grant effectuations to decision issues 

### LINKED ISSUE #8265: Appeal Outcoding | Poll for promulgated rating when Board Grant EPs are cleared
The purpose of creating EPs and contentions is so we can figure out which rating issues were created as a result of which decision issues. In this story we will be retrieving the promulgated rating and matching the rating issues to their decision issues.

## Acceptance Criteria
- All board grant EPs created as part of #8263 should be continuously polled to check whether they have cleared.
- When a board grant EP has been cleared, we should then begin checking for the promulgated rating associated with that EP.
- Once the promulgated rating is found, connect each rating issue with its matching decision issue in the database. Decision issues can be matched based on contention reference id.

## Notes
- If there are any rating issues where a matching decision issue cannot be found. Do nothing for now, this will be handled in a future story.
- rating EPs when cleared don't always get ratings (we might want to add a timeout of 10 days or something), should do this work in a followup

## Technical Notes
- Most of the logic required for this story already exists, we just need to repurpose it to work with board grant EPs.

requires #8263
requires #8266

shanear:
connects #8265


spec/models/board_grant_effectuation_spec.rb
```diff
@@ -3,23 +3,109 @@
...
+        it "attempts sync but doesn't finish processing" do
+          subject
+          expect(board_grant_effectuation).to be_attempted
+          expect(granted_decision_issue).to have_attributes(
+            promulgation_date: nil,
+            profile_date: nil,
+            decision_text: nil,
+            rating_issue_reference_id: nil
+          )
+          expect(board_grant_effectuation).to be_processed
```
jiahuang: Is this test correct?  "but doesn't finish processing" makes it seem like this should be `not_to be_processed`.

app/models/board_grant_effectuation.rb
```diff
@@ -19,12 +20,67 @@ class BoardGrantEffectuation < ApplicationRecord
...
+    end
+  end
+
+  def sync_decision_issues!
+    return if processed?
+
+    attempted!
+    return unless associated_rating
+
+    update_from_matching_rating_issue!
```
jiahuang: should `processed!` still get called if `update_from_matching_rating_issue!` returns nil due to there being no matching rating issue?

spec/models/board_grant_effectuation_spec.rb
leikkisa: Oh thanks! Apparently, if there is an associated rating, but not matching issue, it should still be considered finished for processing. Will update the spec description.

app/models/board_grant_effectuation.rb
leikkisa: In the case where there's a rating, but not a matching rating issue, we're saying that the matching rating issue isn't ever expected to be created so we consider it processed.

shanear:
I approve

PR #8717, opened 2019-01-15
Improve VLJ support staff test data

### LINKED ISSUE #8611: Improve VLJ support staff test data
Test data for our VLJ support staff user is fairly wonky. Improving this would make for a better training experience. The acceptance criteria includes a few examples but if we notice more improvements let's tackle those here:

## Acceptance criteria
* [ ] Make task instructions match the type of task (example of us doing this wrong is an Arneson type but instructions read "poa clarification")
* [ ] Create tasks of type "schedule hearing" and "translation" and any other types that have special behaviour
* [ ] Fix "Days on hold" column in the queue table view and the label in the task instructions area. Both read "NaN" for some cases (do we use a field that may be nil when computing that?)

lowellrex:
Resolves #8611.

This PR addresses confusion caused by confusing test data that were raised by members of the VLJ support staff during a training session last week.

* Makes task instructions agree with the task type
* Populates the `placed_on_hold_at` field for test `ColocatedTask`s
* Creates a `ColocatedTask` of type `translation` so that every different type of `ColocateTask` exists in the test data

Before | After
--- | ---
image | image


spec/factories/task.rb
```diff
@@ -41,15 +41,37 @@
...
-      action { Constants::CO_LOCATED_ADMIN_ACTIONS.keys.sample }
-      instructions ["poa is missing"]
-      appeal { create(:appeal) }
+      factory :ama_colocated_task do
+        appeal { create(:appeal) }
+      end
+
+      after(:build) do |task|
+        # rubocop:disable Metrics/LineLength
+        example_instructions = {
```
mdbenjam: I think this is probably fine. One possible suggestion would be to make these task traits instead of an after(:build). That way we can still easily build all types of tasks. i.e. now it will be impossible to build a task with a title but no instructions.

PR #8718, opened 2019-01-15
Special issues copy changes and new component

### LINKED ISSUE #7216: Make copy changes for special issues page for legacy appeals
Allow attorneys to select special issues during the checkout flow for legacy appeals like we do for AMA appeals. This sets us up to deprecate Caseflow dispatch.

## AC
- Create a screen titled `Select special issues`
- `Select special issues` screen comes after selecting 'Decision Ready for Review' in the attorney checkout flow
- Special issues include all in the below mock EXCEPT `Hearing - including travel board and videoconference`

## Background
See [mural](https://app.mural.co/t/workqueue2001/m/workqueue2001/1536261263374/a50fb8424a7bff4e04c49d0b4023a6a63e519f09) for flow.

This is how special issues are currently displayed in Caseflow Dispatch. 
screen shot 2018-09-06 at 7 12 30 pm

## Mocks

image

## Reference - AMA issue
https://github.com/department-of-veterans-affairs/caseflow/issues/6854

youngfreezy:
Resolves #7216

### Description
Update special issues copy in attorney/judge checkout flow per new mockups.
Since special issues are re-used in different applications, created a new file called `QUEUE_SPECIAL_ISSUES` and added in keys related to the section they should be in and the order. Used the existing `CheckboxGroup` component for the sections of checkboxes.
followed existing pattern of setting snakecase value of special issue in to `special_issues` node in redux store and setting to true/false if user has checked/unchecked that option.

new special issues: `committee_on_waivers_and_compromises` and `dic_united_states`

these are being sent in the post body, but are not being returned by the special issues api (`/appeals/${appeal.externalId}/special_issues`) on a 200 repsonse which means the backend needs to be updated to handle these new special issues? @aroltsch @lowellrex 

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Login as BVASCASPER1
2. Go to their queue /queue
3. Click on a legacy appeal
4. Under available actions, select ready for decision option
5. check the redux store in redux dev tools or check the console output to see that special issues are persisted in redux when toggling the checkboxes
6. when hitting continue, ensure that the special issues api returns a 200


client/app/constants/QueueSpecialIssues.js
```diff
@@ -0,0 +1,301 @@
...
+  {
+    display: 'Contaminated water at Camp LeJeune',
+    specialIssue: 'contaminatedWaterAtCampLejeune',
+    stationOfJurisdiction: {
+      key: '327',
+      location: 'Louisville, KY'
+    },
+    snakeCase: 'contaminated_water_at_camp_lejeune',
+    unhandled: null,
+    section: 'issuesOnAppeal',
```
aroltsch: Instead of creating a new file, I think it is okay to add these keys to the main special issues file. You can introduce a new boolean named `used_for_queue_view` and set that to true for queue special issues. I am skeptical to create a new file in this case because it is a ton of information that we are duplicating.

youngfreezy: OK sounds good! I assumed the changes in text might impact other parts of the application, it seems like that’s not the case. I will also double check.

youngfreezy: @aroltsch just made/pushed the changes we discussed

client/app/constants/SpecialIssues.js
```diff
@@ -2,14 +2,16 @@ import React from 'react';
 
 const SPECIAL_ISSUES = [
   {
-    display: 'Contaminated Water at Camp LeJeune',
+    display: 'Contaminated water at Camp LeJeune',
```
aroltsch: Do you need `queueDisplay` here?

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
aroltsch: I think you probably want to delete this file

client/app/queue/SelectSpecialIssuesView.jsx
```diff
@@ -63,8 +77,42 @@ class SelectSpecialIssuesView extends React.PureComponent {
         {this.getPageNote()}
       </p>
       {error && <Alert type="error" title={error.title} message={error.detail} />}
-      <div className="cf-multiple-columns">
-        {specialIssueCheckboxes}
+      <div {...flexContainer} className="special-options">
+        <div {...flexColumn}>
+          <CheckboxGroup
+            label={<h3>About the appellant: </h3>}
+            name="About the appellant"
```
aroltsch: Let's move these constants to our COPY.json file so it would be easy for designers to make changes.

aroltsch:
Hey Fareez, for FE tickets it is also a good idea to post screenshots.

youngfreezyVA:
@aroltsch all tests pass! :)

aroltsch:
1. Based on the mocks, "Has a private attorney or agent" should not be capitalized"
Mocks:
image

Your view: 
image


2. Same here
Mocks: 
image


Your view:
image

3. DIC or Pension is spelled out.

Mocks: 
image


Your view:
image

4. No comma at the end

Mocks:

image


Your view:
image

PR #8721, opened 2019-01-15
Allow judges to create admin actions

### LINKED ISSUE #8608: Allow judges to send admin actions to VLJ support
When judges identify admin actions on a case, they want to send the case to VLJ support, so VLJ support can act on it as quickly as possible. 

This will be relevant for both Legacy and AMA cases, but if we need to, we can just do AMA cases first. 

## AC
- Judges have the option to `Add admin action` from the Case details page, currently active tasks component, Actions dropdown
- When judges click `Add admin action`, they navigate to the add admin action page (same behavior as attorneys)

lowellrex:
Resolves #8608. This PR allows judges to send tasks directly to the VLJ support staff.

## Test path
* Log in as `BVAAABSHIRE`
* Notice that you can add an admin action for any case in the assign or review view

### DIFF
--- a/app/models/tasks/judge_assign_task.rb
+++ b/app/models/tasks/judge_assign_task.rb
@@ -1,5 +1,5 @@
 class JudgeAssignTask < JudgeTask
-  def available_actions(_user)
+  def additional_actions(_user)
     [Constants.TASK_ACTIONS.ASSIGN_TO_ATTORNEY.to_h]
   end
 
--- a/app/models/tasks/judge_assign_task.rb
+++ b/app/models/tasks/judge_assign_task.rb
@@ -1,5 +1,5 @@
 class JudgeAssignTask < JudgeTask
-  def additional_actions(_user)
+  def additional_available_actions(_user)
     [Constants.TASK_ACTIONS.ASSIGN_TO_ATTORNEY.to_h]
   end
 
--- a/spec/feature/queue/task_queue_spec.rb
+++ b/spec/feature/queue/task_queue_spec.rb
@@ -323,6 +323,9 @@ RSpec.feature "Task queue" do
 
       before do
         visit("/queue/appeals/#{appeal.external_id}")
+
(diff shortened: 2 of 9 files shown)


spec/feature/queue/task_queue_spec.rb
```diff
@@ -336,6 +339,22 @@
...
+        click_dropdown(text: action)
+        fill_in(COPY::ADD_COLOCATED_TASK_INSTRUCTIONS_LABEL, with: "Please complete this task")
+        find("button", text: COPY::ADD_COLOCATED_TASK_SUBMIT_BUTTON_LABEL).click
+
+        # Expect to see a success message and have the task in the database
+        expect(page).to have_content(format(COPY::ADD_COLOCATED_TASK_CONFIRMATION_TITLE, "an", "action", action))
+        expect(judge_task.children.length).to eq(1)
+        expect(judge_task.children.first).to be_a(ColocatedTask)
+      end
     end
```
tomas-nava: should we have a test for legacy appeals as well?

app/models/tasks/judge_assign_task.rb
```diff
@@ -1,5 +1,5 @@
 class JudgeAssignTask < JudgeTask
-  def available_actions(_user)
+  def additional_actions(_user)
     [Constants.TASK_ACTIONS.ASSIGN_TO_ATTORNEY.to_h]
```
tomas-nava: I think it'd be clearer if this method had `available_actions` in its title, like `additional_available_actions`

spec/feature/queue/task_queue_spec.rb
lowellrex: Great catch! I'll add one tomorrow.

lowellrex: Actually, this PR doesn't doesn't include the work to make this work for Legacy appeals and it looks like that is alright by the acceptance criteria in the original ticket (#8608). I'll make a new ticket to do that work.

PR #8730, opened 2019-01-15
Use strong_migrations gem

monfresh:
**Why**: `strong_migrations` is better maintained than
`zero_downtime_migrations` and seems to have better advice. The author
of the gem has written many other fantastic gems that are used in
production at Instacart, so they should be trustworthy.

The gem also comes with a handy rake task that will automatically
alphabetize the columns in `schema.rb`. This will eliminate issues
where after pulling the latest code and running recent migrations
locally, your schema.rb will change.

### DIFF
--- a/config/initializers/strong_migrations.rb
+++ b/config/initializers/strong_migrations.rb
@@ -1,0 +1,1 @@
+StrongMigrations.start_after = 20190111000717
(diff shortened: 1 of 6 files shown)


config/initializers/strong_migrations.rb
```diff
@@ -0,0 +1 @@
+StrongMigrations.start_after = 20190111000717
```
aroltsch: Nice.

PR #8733, opened 2019-01-15
fix judge script

amprokop:


amprokop:
closing, not merging, we set these roottask statuses to complete in the prod DB

Issue #8710, opened 2019-01-15
Intake | system error thrown when NilRatingProfileListError


shanear:
Currently we're seeing a "Something went wrong" when going from the review page to the add issues page of an intake for any decision review if the `NilRatingProfileListError` is thrown.

Even if there are no ratings, we should still allow a user to intake a decision review for a veteran.

Alright, so there are 4 scenarios that we need tested/covered:

### 1. has both a backfilled/converted and promulgated rating
VBMS has said that this is difficult to test because it's difficult to make a backfilled/converted rating in UAT. We are trusting them that this DOES send back ratings.

### 2. has only a backfilled/converted rating
This throws an error, but we should allow the intake to continue

### 3. has both a locked and a promulgated rating
This should return ratings as normal, now that BGS has fixed their issue. We still need to test this in UAT.

### 4.  has only a locked rating
Like with backfilled ratings, we should allow the intake to continue, just with no ratings

## Technical Notes
A UAT veteran to test this on is: 700036051

jiahuang:
blocked on getting some more example data from bgs

nicholasholtz:
@jiahuang are we still waiting on this example data? Who did we ask for this from?

leikkisa:
Sally has asked for example response, haven't received a response yet, going to escalate and include Nicholas and Jia

nicholasholtz:
Sally, could you forward me what you asked for?

jiahuang:
I followed up on the email again. I'm trying to get sample data for a veteran that has both a backfilled rating as well as promulgated ratings.

nicholasholtz:
Linda Ciston responded on 1/31, indicating a response would be forthcoming today or tomororow

nicholasholtz:
Linda Ciston provided response on 2/4

jiahuang:
Got data for #1 (has both a backfilled/converted and promulgated rating), it returns ratings. Waiting on data for #3 (has both a locked and a promulgated rating), but if it behaves the same way as #1, we will handle it correctly.

nicholasholtz:
Colin provided sample data on 2/8

leikkisa:
Looked at the rating that Colin added on the file number, and it's working as expected, we're getting the promulgated rating, and not the backfilled (not even with any reject reason). 

[...]

PR #8743, opened 2019-01-15
do not error out on backfilled ratings

jiahuang:
Connects #8710

### DIFF
--- a/app/models/decision_review.rb
+++ b/app/models/decision_review.rb
@@ -268,7 +268,8 @@ class DecisionReview < ApplicationRecord
     veteran.ratings.reject { |rating| rating.issues.empty? }
 
     # return empty list when there are no ratings
-  rescue Rating::BackfilledRatingError
+  rescue Rating::BackfilledRatingError => e
+    Raven.capture_exception(e)
     []
   end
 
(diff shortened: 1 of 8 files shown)


app/models/decision_review_intake.rb
```diff
@@ -12,9 +12,6 @@ def ui_hash(ama_enabled)
       activeNonratingRequestIssues: detail.active_nonrating_request_issues.map(&:ui_hash),
       contestableIssuesByDate: detail.contestable_issues.map(&:serialize)
     )
-  rescue Rating::NilRatingProfileListError
-    cancel!(reason: "system_error")
```
pkarman: the `cancel` was to workaround a specific error that left the intake a stuck state. Does that issue no longer exist in this case?

leikkisa:
So if there are just no ratings, we do already allow then to continue. This error happens when there's a Locked Rating or a Backfilled or Converted Rating. We should get confirmation that we want the user to be able to continue if they have a Locked Rating.

app/models/decision_review.rb
```diff
@@ -266,6 +266,10 @@ def ratings_with_issues
     return [] unless veteran
 
     veteran.ratings.reject { |rating| rating.issues.empty? }
+
+    # return empty list when there are no ratings
+  rescue Rating::BackfilledRatingError
+    []
```
leikkisa: Can we do a Raven.capture_exception so that we can still get some sort of alert when there is a backfilled rating?  We still want to investigate individual cases to learn more about them (for example, the last one we investigated debunked BGS's previous assumption that this only happened for really old ratings).

PR #8744, opened 2019-01-15
Create hearing task at case details load

### LINKED ISSUE #8573: ScheduleVeteranTask should be created on Case Details page load for legacy cases and become a top-level Task
A recent update to the case details page creates a timeline view of all open tasks with available actions. #8101 adds `ScheduleVeteranTask` as an action for the `RootTask`, but this is inconsistent with how Tasks are intended to be used, and offers no way to complete or block a `ScheduleVeteranTask` with additional admin/actions. 

A `ScheduleVeteranTask` needs to be actionable on the Case Details page for all **veterans that have a legacy case in location 57 and do not have a scheduled hearing date**. 

### Plan
- Remove the `ScheduleVeteranTask` as an action from the `RootTask` and instead check on page load if veteran can have hearings scheduled and create a task if so
- Add a `SCHEDULE_VETERAN` action to the `ScheduleVeteranTask` as an available action
- Unskip the schedule veteran feature tests

### Implementation

 To prevent having to force a page reload or refetch data if a `ScheduleVeteranTask` needs to be created, I think we should add this logic to the `TasksController`.

In [`TasksController#for_appeal`](https://github.com/department-of-veterans-affairs/caseflow/pull/8101) add something like
```ruby
ScheduleVeteranTask.create_if_eligible(appeal)
```

mdbenjam:
Resolves #8573

### Description
This moves the task creation of ScheduleHearingTasks from clicking on a task action to loading the case details page. This will allow us to add admin actions to the schedule hearing task.

### Testing Plan
1. Manually ensure that the task action box appears when an appeal that needs a hearing schedule is navigated to.

### DIFF
--- a/app/models/tasks/schedule_hearing_task.rb
+++ b/app/models/tasks/schedule_hearing_task.rb
@@ -1,14 +1,14 @@
 class ScheduleHearingTask < GenericTask
   class << self
-    def create_from_params(params, current_user)
-      root_task = RootTask.find_or_create_by!(appeal: params[:appeal])
-      params[:parent_id] = root_task.id
-
-      task_payloads = params.delete(:business_payloads)
-      child_task = super(params, current_user)
-      child_task.task_business_payloads.create(task_payloads) if task_payloads
-
-      child_task
+    def create_if_eligible(appeal)
+      if appeal.is_a?(LegacyAppeal) && appeal.case_record.bfcurloc == "57"
+        ScheduleHearingTask.find_or_create_by!(appeal: appeal) do |task|
+          task.update(
+            assigned_to: HearingsManagement.singleton,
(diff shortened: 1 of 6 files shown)


app/models/tasks/schedule_hearing_task.rb
```diff
@@ -28,25 +28,9 @@ def update_from_params(params, current_user)
       task_business_payloads.create(task_payloads)
     end
 
-    super(params, current_user)
-  end
-
-  def update_parent_status
```
mdbenjam: I don't think this logic really belongs here. We want to do this on completing the task not when we update the parents status.

mdbenjam: I think queue renamed the method and made this logic not really belong here anymore.

Issue #8747, opened 2019-01-16
Case Search | When a veteran has HLRs or SCs and no Appeals, the case search page shows nothing


shanear:
When a veteran has no appeals, but does have claim reviews, they do not show up on the case search page.

![bug2](https://user-images.githubusercontent.com/1596259/51218399-0f01df80-18e1-11e9-98a3-1b71b1a9daaa.gif)

Here is the response payload from the search request for the GIF above:
```
{
    "appeals": [],
    "claim_reviews": [
        {
            "caseflow_veteran_id": 1253,
            "claim_id": 879,
            "claimant_names": [
                "Rebelo Devin"
            ],
            "end_products": [
                {
                    "benefit_type_code": "1",
                    "claim_date": "2019-01-01",
                    "claimant_participant_id": "600229055",
                    "code": "030HLRR",
                    "committed_at": "2019-01-15T17:47:25.068-05:00",
                    "development_item_reference_id": null,
                    "doc_reference_id": null,
                    "established_at": "2019-01-15T17:47:21.697-05:00",
                    "id": 667,
                    "last_synced_at": "2019-01-15T18:20:10.130-05:00",
                    "modifier": "030",
                    "payee_code": null,
                    "reference_id": "600147333",
                    "source_id": 879,
                    "source_type": "HigherLevelReview",
                    "station": "499",
                    "synced_status": "CLR",
                    "user_id": 405,
                    "veteran_file_number": "124592531"
                }
            ],
            "review_type": "higher_level_review",
            "veteran_file_number": "124592531",
            "veteran_full_name": "Rebelo Devin"
        }
    ]
}
```

## Technical Notes
try 124592531 in UAT to reproduce this.

PR #8749, opened 2019-01-16
Show HLRs / SCs if there aren't any appeals for a veteran

joeyyang:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/8747

cc @shanear 

### Description
This PR fixes a bug caused by this condition:
- When a veteran has no appeals, but does have claim reviews, they do not show up on the case search page.

screenshot 2019-01-15 18 30 08


shanear:
Good tests. Looks good to me but didn't do a thorough review

Issue #8080, opened 2018-12-03
Queue | Hearing Schedule Tasks


sharonwarner:
We need to work with the intake team to ensure that Hearing Schedule tasks are created for each veteran who has selected that AMA lane.

PR #8750, opened 2019-01-16
Create hearing task at intake

mdbenjam:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/8080

### Description
Creates a schedule hearing task when an appeal in the hearing docket is intook.

### Testing Plan
1. Manually go through the flow from AMA appeal with schedule hearing task to actually schedule the hearing.
2. Add automated tests to exercise completing the task.


spec/models/hearing_day_spec.rb
```diff
@@ -141,9 +141,8 @@
 
     subject { HearingDay.load_days(Time.zone.today, Time.zone.today, "C") }
 
-    it "gets hearings for a date range" do
-      hearings
-      expect(subject.size).to eq 2
+    it "shouldn't load any since we're past HearingDay::CASEFLOW_CO_PARENT_DATE" do
+      expect(subject[:vacols_hearings].size).to eq 0
```
mdbenjam: This was not testing the right thing before. It was simply testing if the method returned an array of 2, which it did. But we should've been checking if the first array within the array had a size of three, which it no longer does since we're past the cutoff date.

spec/models/hearing_day_spec.rb
```diff
@@ -120,15 +120,15 @@
...
+    let!(:hearings) do
       [create(:case_hearing),
        create(:case_hearing)]
     end
 
     subject { HearingDay.load_days(Time.zone.today, Time.zone.today, "RO13") }
 
     it "gets hearings for a date range" do
-      expect(subject.size).to eq 2
+      expect(subject[:vacols_hearings].size).to eq 2
```
mdbenjam: This test was passing because we were testing the wrong thing. See below comment.

app/models/hearing_day.rb
```diff
@@ -111,33 +131,34 @@ def load_days_with_open_hearing_slots(start_date, end_date, regional_office = ni
...
 
     def filter_non_scheduled_hearings(hearings)
-      filtered_hearings = []
-      hearings.each do |hearing|
-        if hearing.vacols_record.hearing_type == REQUEST_TYPES[:central]
-          if !hearing.vacols_record.folder_nr.nil?
-            filtered_hearings << hearing
+      hearings.select do |hearing|
+        if hearing[:vacols_record].nil?
+          true
```
mdbenjam: Keeping all caseflow hearings

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -10,25 +10,41 @@ def create_if_eligible(appeal)
         end
       end
     end
+
+    def tasks_for_ro(regional_office)
```
mdbenjam: Will write a test for this.

app/models/hearing_day.rb
```diff
@@ -81,28 +96,33 @@ def update_schedule(updated_hearings)
...
-                                regional_office, start_date, end_date).each_with_object([])
+                                regional_office, start_date, end_date)
         video_and_co, travel_board =
           HearingDayRepository.load_days_for_regional_office(regional_office, start_date, end_date)
       end
-      cf_video_and_co = enrich_with_judge_names(cf_video_and_co)
-      total_video_and_co = video_and_co + cf_video_and_co
-      [total_video_and_co, travel_board]
+
+      {
```
mdbenjam: I think returning arrays is a bad idea. You have to know what exactly is being returned in what order. I much perfer returning hashes.

app/models/hearing_day.rb
```diff
@@ -111,33 +131,34 @@ def load_days_with_open_hearing_slots(start_date, end_date, regional_office = ni
...
+          regional_office_hash[hearing_day.regional_office], hearing_day
         )
 
         if scheduled_hearings.length >= total_slots || hearing_day[:lock]
           nil
         else
-          hearing_day.slice(:id, :scheduled_for, :request_type, :room).tap do |day|
-            day[:hearings] = scheduled_hearings
+          HearingDay.to_hash(hearing_day).slice(:id, :scheduled_for, :request_type, :room).tap do |day|
+            day[:hearings] = scheduled_hearings.map { |hearing| hearing.to_hash(RequestStore.store[:current_user].id) }
```
mdbenjam: I do not like using RequestStore in our models... But for now going to do it. Don't want to make this PR larger.

app/models/hearing_day.rb
sharonwarner: I think we can also get rid of travel board here. We don't use it anywhere.

app/models/hearing_day.rb
sharonwarner: Ah, this will create a bug later. We use hearings count to see how many hearings are scheduled for that day. If a hearing gets postponed or cancelled, then that means the hearing coordinators should be able to schedule another hearing. The logic is something like `allow_user_to_schedule_a_hearing if scheduled_hearings.count < hearing_day.total_slots`. This means for AMA hearings, they won't be able to schedule over a hearing that was postponed or canceled. Not a problem for now, but let's make sure we create a ticket to fix this.

app/repositories/hearing_day_repository.rb
```diff
@@ -19,35 +19,29 @@ def find_hearing_day(request_type, hearing_key)
 
     def load_days_for_range(start_date, end_date)
       video_and_co = VACOLS::CaseHearing.load_days_for_range(start_date, end_date)
-        .each_with_object([]) do |hearing, result|
-        result << to_canonical_hash(hearing)
-      end
+
       removed_children_records = video_and_co.reject do |hearing_day|
-        hearing_day[:request_type] == "C" && hearing_day[:scheduled_for] > HearingDay::CASEFLOW_CO_PARENT_DATE
```
sharonwarner: Oh no... is it possible to add a mapper that changes hearing_type -> request_type and hearing_date -> scheduled_for in VACOLS::CaseHearing? I guess it's not a big deal to have VACOLS column names in the repository, this just feels like a change in the wrong direction.

app/repositories/hearing_repository.rb
```diff
@@ -102,7 +102,7 @@ def slot_new_hearing(parent_record_id, time, appeal)
     end
 
     def update_co_hearing(hearing_date_str, appeal)
-      if hearing_date_str.to_date > HearingDay::CASEFLOW_CO_PARENT_DATE
+      if hearing_date_str.to_date > HearingDay::CASEFLOW_CO_PARENT_DATE || appeal.is_a?(Appeal)
```
sharonwarner: At this point, we can actually update this to always create a child co hearing. We're no longer slotting any CO hearings through the fake VACOLS hearing days.

app/repositories/hearing_repository.rb
```diff
@@ -131,37 +131,57 @@ def create_child_co_hearing(hearing_date_str, appeal)
...
+          room: hearing.room,
+          board_member: hearing.board_member,
+          vdbvapoc: hearing.vdbvapoc
+        )
+      else
+        Hearing.create!(
+          appeal: appeal,
+          hearing_day_id: hearing_day.id,
+          room: hearing.room,
+          judge_id: hearing_day.judge.id,
```
sharonwarner: This won't necessarily always exist, we should do hearing_day.judge.try(:id).

app/repositories/hearing_repository.rb
```diff
@@ -131,37 +131,57 @@ def create_child_co_hearing(hearing_date_str, appeal)
...
-        vdbvapoc: hearing_day.bva_poc
-      )
+      if (appeal.is_a?(LegacyAppeal))
+        VACOLS::CaseHearing.create_child_hearing!(
+          folder_nr: appeal.vacols_id,
+          hearing_date: VacolsHelper.format_datetime_with_utc_timezone(hearing_date),
+          vdkey: hearing_day.id,
+          hearing_type: hearing_day.request_type,
+          room: hearing_day.room,
+          board_member: hearing_day.judge ? hearing_day.judge.vacols_attorney_id : nil,
```
sharonwarner: Same for this, the judge won't always exist.

app/repositories/hearing_repository.rb
```diff
@@ -131,37 +131,57 @@ def create_child_co_hearing(hearing_date_str, appeal)
...
+          vdbvapoc: hearing_day.bva_poc
+        )
+      else
+        Hearing.create!(
+          appeal: appeal,
+          hearing_day_id: hearing_day.id,
+          room: hearing.room,
+          judge_id: hearing_day.judge.id,
```
sharonwarner: And here 😂

spec/models/hearing_day_spec.rb
sharonwarner: Yikes. Thanks for catching that!

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -10,25 +10,41 @@ def create_if_eligible(appeal)
...
+      # Get all legacy tasks for this RO
+      legacy_appeal_tasks = AppealRepository.appeals_ready_for_hearing_schedule(regional_office).map do |appeal|
```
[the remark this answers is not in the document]
sharonwarner: Just remembered to come back to this - At least for now, we're pulling in legacy appeals by location, will this cause those appeals to show up twice?

app/repositories/hearing_day_repository.rb
mdbenjam: Good call. I added methods on case_hearing called `scheduled_for` and `request_type` to alias their real names!

app/models/tasks/schedule_hearing_task.rb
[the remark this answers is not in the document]
mdbenjam: Let's talk about this tomorrow. Not entirely sure what you mean by this.

app/repositories/hearing_repository.rb
mdbenjam: I think this one works. the logic didn't change here :) It checks for judge's existence first

app/models/hearing_day.rb
mdbenjam: So I tried to, but there are some places in the assign algorithm where we do check travel board :/

app/models/hearing_day.rb
mdbenjam: Is there a way to fix this now? Do we have names for the dispositions? `:postponed` and `:canceled`?

spec/models/hearing_day_spec.rb
```diff
@@ -251,15 +250,21 @@
...
       it "returns hearings are mapped to days" do
-        expect(subject.size).to eq 2
-        expect(subject[1][:hearings].size).to eq 2
-        expect(subject[1][:request_type]).to eq "V"
-        expect(subject[1][:hearings][0][:appeal_id]).to eq appeal.id
-        expect(subject[1][:hearings][0][:hearing_disp]).to eq nil
-        expect(subject[1][:hearings][1][:appeal_id]).to eq appeal_today.id
-        expect(subject[0][:hearings][0][:appeal_id]).to eq appeal_tomorrow.id
+        expect(subject.size).to eq 3
+        expect(subject[0][:hearings][0]["appeal_id"]).to eq ama_appeal.id
```
mdbenjam: This makes me think we'll need to do some rejiggering on the frontend to pass appeal_type so we can differentiate between ama and legacy.

app/models/hearing_day.rb
sharonwarner: 😭

app/models/hearing_day.rb
sharonwarner: Yep!

app/models/vacols/case_hearing.rb
```diff
@@ -177,6 +177,18 @@ def select_schedule_days
...
 
+  def scheduled_for
+    hearing_date
+  end
+
+  def request_type
+    hearing_type
+  end
+
+  def judge_id
```
sharonwarner: Is this definitely correct? We use Caseflow's judge ids for hearing days / hearings, but sattyids in VACOLS.

app/repositories/hearing_repository.rb
```diff
@@ -102,23 +102,15 @@ def slot_new_hearing(parent_record_id, time, appeal)
...
-      if hearing_date_str.to_date > HearingDay::CASEFLOW_CO_PARENT_DATE
-        return create_child_co_hearing(hearing_date_str, appeal)
-      end
-
-      # Get the next open slot for that hearing date and time.
-      hearing = VACOLS::CaseHearing.find_by(hearing_date: hearing_date_str, folder_nr: nil)
-      fail NoOpenSlots, message: "No available slots for this hearing day." if hearing.nil?
-
-      loaded_hearing = VACOLS::CaseHearing.load_hearing(hearing.hearing_pkseq)
-      HearingRepository.update_vacols_hearing!(loaded_hearing, folder_nr: appeal.vacols_id)
```
sharonwarner: 😂 🎉

sharonwarner: This doesn't have to be done in this PR, but we can remove this method entirely. It's just calling the method right below it.

app/repositories/hearing_repository.rb
sharonwarner: Ah yup!

PR #8753, opened 2019-01-16
tracks distribution task ready time, catches timer task job errors

amprokop:
Resolves #https://github.com/department-of-veterans-affairs/caseflow/issues/8755, https://github.com/department-of-veterans-affairs/caseflow/issues/8756

### Manual testing Plan

To test the locking behavior of the task timer job:

Run in console window 1:
```
timer = TaskTimer.find(1)
timer.update!(task_id: 666)
timer.with_lock { sleep 30; timer.update!(task_id: 888) }
```

Run in console window 2:
```
timer.with_lock { puts timer.task_id }
```
Observe that console window 2 hangs until console window 1 finishes, then prints `888`

### DIFF
--- a/app/jobs/task_timer_job.rb
+++ b/app/jobs/task_timer_job.rb
@@ -6,19 +6,23 @@ class TaskTimerJob < CaseflowJob
     RequestStore.store[:application] = "queue"
     RequestStore.store[:current_user] = User.system_user
 
-    TaskTimer.transaction do
-      TaskTimer.requires_processing.includes(:task).each do |task_timer|
-        task_timer.lock!
-        process(task_timer)
-      end
+    TaskTimer.requires_processing.includes(:task).each do |task_timer|
+      # TODO: if this job's runtime gets too long, spawn individual jobs for each task timer.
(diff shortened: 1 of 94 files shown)


app/jobs/task_timer_job.rb
```diff
@@ -6,19 +6,24 @@ def perform
...
     end
   end
 
   def process(task_timer)
-    # the reload creates an N+1 query situation
-    return if task_timer.reload.processed?
+    # Calling ".with_lock" will block the current thread until no other threads have a lock on the row.
+    task_timer.with_lock do
+      # Reloading after we acquire a lock will ensure that we don't double-process jobs.
+      return if task_timer.reload.processed?
```
aroltsch: It seems as `with_lock` does the reload. Something worth testing. 

```
Under the hood, this is what happens within a with_lock block:

opens up a database transaction
reloads the record instance
requests exclusive access to the record
```

amprokop: 👍 👍 yes, this doesn't need to be here. thanks.

PR #8760, opened 2019-01-16
Common Components for Hearing Dropdowns that Require Data from AJAX Calls

### LINKED ISSUE #8038: Move RoSelectorSelector + HearingDateDropdown state from redux to component.
See [Mark's comment](https://github.com/department-of-veterans-affairs/caseflow/pull/7950/files/70dce969e418efddab51f0ce74009d1ad24358c9#r237534514). 

Storing the available hearings dates / regional office in redux is most likely unnecessary. We should move the state from redux to component.

lomaxap:
Resolves #8038

### Description
- Creates dropdown components for Regional Office, HearingDates, Hearing Coordinators, Judges, and Hearing Rooms. Fetches data internally for each component rather than requiring parent component to fetch and pass data.
- Creates a shared heardingDropdownData state with list of dropdown options from backend in redux store

### Acceptance Criteria
- [ ] If more than one of the same type of dropdown is rendered on a page, only fetches data once
- [ ] Dropdowns take string values

### DIFF
--- a/client/app/hearingSchedule/components/DataDropdowns/RegionalOffice.jsx
+++ b/client/app/hearingSchedule/components/DataDropdowns/RegionalOffice.jsx
@@ -52,13 +52,14 @@ class RegionalOfficeDropdown extends React.Component {
   }
 
   render() {
-    const { name, label, onChange, regionalOffices: { options } } = this.props;
+    const { name, label, onChange, regionalOffices: { options }, readOnly } = this.props;
 
     return (
       <SearchableDropdown
         name={name}
         label={label}
         strongLabel
+        readOnly={readOnly}
         value={this.getSelectedOption()}
         onChange={(option) => onChange(option.value)}
         options={options} />
@@ -70,7 +71,8 @@ RegionalOfficeDropdown.propTypes = {
   name: PropTypes.string,
   label: PropTypes.string,
   value: PropTypes.string,
-  onChange: PropTypes.func.isRequired
+  onChange: PropTypes.func.isRequired,
+  readOnly: PropTypes.bool
 };
 
 RegionalOfficeDropdown.defaultProps = {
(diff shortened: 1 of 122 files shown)


client/app/hearingSchedule/components/DataDropdowns/constants.js
```diff
@@ -0,0 +1,59 @@
+export const ACTIONS = {
+  RECEIVE_DROPDOWN_DATA: 'RECEIVE_DROPDOWN_DATA',
+  FETCH_DROPDOWN_DATA: 'FETCH_DROPDOWN_DATA'
+};
+
+export const HEARING_ROOM_OPTIONS = [
```
sharonwarner: You should be able to pull these from client/constants/HEARING_ROOMS_LIST.json.

client/app/hearingSchedule/components/DataDropdowns/RegionalOffice.jsx
```diff
@@ -0,0 +1,111 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+import { onReceiveDropdownData, onFetchDropdownData } from './actions';
+import ApiUtil from '../../../util/ApiUtil';
+import _ from 'lodash';
+
```
oscarAtNava: We already have a regional office component in the common components section of our code: client/app/components/RoSelectorDropdown.jsx. Are we discontinuing its use?

client/app/hearingSchedule/components/DataDropdowns/constants.js
sharonwarner: We use it on the backend to map room numbers too.

client/app/hearingSchedule/components/DataDropdowns/RegionalOffice.jsx
sharonwarner: Is it possible to do that as part of this PR? It will be confusing to have multiple components that do essentially the same thing.

PR #8766, opened 2019-01-16
VACOLS issues should also ignore the AMA timeliness reqs

pkarman:
connects #7337 

### Description
VACOLS (legacy) issues do not get AMA timeliness check.

### DIFF
--- a/app/models/request_issue.rb
+++ b/app/models/request_issue.rb
@@ -499,9 +499,7 @@ class RequestIssue < ApplicationRecord
   end
 
   def should_check_for_before_ama?
-    return false if is_unidentified || ramp_claim_id || vacols_id
-
-    true
+    !is_unidentified && !ramp_claim_id && !vacols_id
   end
 
   def check_for_legacy_issue_not_withdrawn!
--- a/app/models/request_issue.rb
+++ b/app/models/request_issue.rb
@@ -491,13 +491,17 @@ class RequestIssue < ApplicationRecord
   # rubocop:enable Metrics/PerceivedComplexity
 
   def check_for_before_ama!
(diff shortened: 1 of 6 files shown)


app/models/request_issue.rb
```diff
@@ -491,15 +491,19 @@ def check_for_eligible_previous_review!
...
-    return if ramp_claim_id
+    return unless eligible? && should_check_for_before_ama?
 
     if decision_or_promulgation_date && decision_or_promulgation_date < DecisionReview.ama_activation_date
       self.ineligible_reason = :before_ama
     end
   end
 
+  def should_check_for_before_ama?
+    return false if is_unidentified || ramp_claim_id || vacols_id
```
shanear: I feel like this might be a cleaner way to say this:
```
!is_unidentified && !ramp_claim_id && !vacols_id
```

spec/feature/intake/supplemental_claim_spec.rb
```diff
@@ -867,6 +867,17 @@ def start_supplemental_claim(
...
+          # add before_ama ratings
+          click_intake_add_issue
+          add_intake_rating_issue("Non-RAMP Issue before AMA Activation")
+          add_intake_rating_issue("limitation of thigh motion (extension)")
+          add_untimely_exemption_response("Yes")
+
+          expect(page).to have_content("Non-RAMP Issue before AMA Activation")
+          expect(page).to_not have_content(
+            "Non-RAMP Issue before AMA Activation #{ineligible_constants.before_ama}"
+          )
```
shanear: Interesting, this doesn't seem to test any functionality associated with the fix. Was this test not already there?

Issue #8739, opened 2019-01-15
Unprocessable Entity after SUBMIT_REVIEW_FAIL


kencheeto:
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3252

jiahuang:
Looked through the kibana logs. I can see 1 request fail and then immediately the next request completing in 30s, with the difference being this in their params:

Failed request:
Parameters: {"option_selected"=>"appeal", "appeal_docket"=>"evidence_submission", "id"=>"122971", "intake"=>{}}

Successful request:
Parameters: {"option_selected"=>"appeal", "receipt_date"=>"2018-11-20", "appeal_docket"=>"evidence_submission", "id"=>"122971", "intake"=>{}}

In [`rampRefiling.js`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/intake/actions/rampRefiling.js#L54-L58) we're trying to set receipt_date. My guess is that we're not setting it properly somehow..

jiahuang:
I was able to reproduce us sending a payload without the required fields. The "continue to next step" should be grayed out until the receipt date filled out & checkbox is selected

![image](https://user-images.githubusercontent.com/577629/51285621-6917a900-19a4-11e9-8fb1-d36106cdb7f1.png)

jiahuang:
Sentry is reporting 4xxs that aren't actual errors.

In the case of intake `SUBMIT_REVIEW_FAIL`, we handle the error in the frontend and then notify the user that the form has to be filled out. Failed logs are usually followed by a successful retry 30s later in kibana.

There are other cases of sentry logging 4xxs that may not be real errors.

On reader, seems like some metrics don't properly post https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3252/events/221015/
```
> %c Action with type "SET_OPENED_ACCORDION_SECTIONS" took 1.00 milliseconds. background: #bada55; color: #222
> POST /metrics/v1/histogram 422
```

Search on reader returns 422? Maybe when a search can't return data?
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3252/events/220356/
```
> POST /reader/appeal/3013540/claims_folder_searches 422
```

leikkisa:
I'm closing this one - Jia submitted a fix for the specific case of that brought up this ticket. I think the issue then became about not reporting these to Sentry.  

[...]

PR #8780, opened 2019-01-17
require receiptDate to be set

jiahuang:
Connects #8739

This is a stopgap for the sentry error. Ideally we'd have validation on each type of form type to make sure all fields are filled out.

### DIFF
--- a/client/app/intake/pages/review.jsx
+++ b/client/app/intake/pages/review.jsx
@@ -79,8 +79,8 @@ class ReviewNextButton extends React.PureComponent {
         this.handleClick(selectedForm, intakeData);
       }}
       loading={intakeData ? intakeData.requestStatus.submitReview === REQUEST_STATE.IN_PROGRESS : true}
-      disabled={formType === 'ramp_refiling' ?
-        toggleIneligibleError(intakeData.hasInvalidOption, intakeData.optionSelected) : needsRelationships}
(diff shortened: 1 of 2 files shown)


leikkisa:
Don't they currently get kicked back with the validation error if they submit without a receipt date?

jiahuang:
Yeah they do, I think maybe we should just not have sentry report this error instead

client/app/intake/pages/review.jsx
```diff
@@ -79,8 +79,8 @@ class ReviewNextButton extends React.PureComponent {
         this.handleClick(selectedForm, intakeData);
       }}
       loading={intakeData ? intakeData.requestStatus.submitReview === REQUEST_STATE.IN_PROGRESS : true}
-      disabled={formType === 'ramp_refiling' ?
-        toggleIneligibleError(intakeData.hasInvalidOption, intakeData.optionSelected) : needsRelationships}
+      disabled={!intakeData.receiptDate || (formType === 'ramp_refiling' ?
+        toggleIneligibleError(intakeData.hasInvalidOption, intakeData.optionSelected) : needsRelationships)}
```
kencheeto: i think this is complicated enough that we should extract this into a helper rather than inline. Fine with that being a followup tho!

jiahuang:
closing in favor of #8786

PR #8786, opened 2019-01-17
do not report err on SUBMIT_REVIEW_FAIL

jiahuang:
Connects #8739

catch the error so it doesn't show up on sentry


jiahuang:
doh, looks like we're using throwing the error to avoid the page redirect.

https://docs.sentry.io/clients/javascript/usage/#raven-js-promises
> By default, Raven.js capture unhandled promise rejections as described in official ECMAScript 6 standard.

Now I'm catching the error and not doing anything so that it doesn't get reported. Shouldn't change any functionality.

PR #8790, opened 2019-01-17
Add asyncable jobs admin page

pkarman:
connects #6944 

### Description
A new route for asyncable jobs that have expired. If the user has an `Admin Intake` role, they may view a page showing expired jobs and their last error messages, and restart any of them.

screen shot 2019-01-18 at 4 16 06 pm


app/controllers/asyncable_jobs_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+      expired_jobs << klass.expired_without_processing
+    end
+    expired_jobs.flatten.sort_by(&:submitted_at_dtim)
+  end
+
+  def asyncable_models
+    ActiveRecord::Base.descendants.select { |c| c.included_modules.include?(Asyncable) }
+      .reject(&:abstract_class?)
+      .map(&:name)
+      .map(&:constantize)
```
shanear: This is a cool way to do this. why is the `.map(&:name).map(&:constantize)` needed?

app/controllers/asyncable_jobs_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+  def jobs
+    @jobs ||= gather_jobs
+  end
+
+  def job
+    @job ||= allowed_params[:asyncable_job_klass].constantize.find(allowed_params[:id])
+  end
+
+  # TODO: how to support paging when coallescing so many different models?
+  def gather_jobs
```
shanear: This might be better positioned as a class method on `Asyncable`. This way it'd be easier to unit test. Not a fan of a lot of controller logic.

maybe `Asyncable.all_unprocessed` or something like that?

spec/controllers/asyncable_jobs_controller_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      context "asyncable klass does not include Asyncable concern" do
+        it "returns 404 error" do
+          get :index, params: { asyncable_job_klass: "Intake" }
+
+          expect(response.status).to eq 404
+        end
+      end
+    end
+  end
+end
```
shanear: I think i'd rather this be a feature spec than a controller spec. Something along the lines of:

- test the access control
- create a failed SupplementalClaim
- create a failed HigherLevelReview
- create a successful job
- test that they show up on the page 
- test the retry button

app/controllers/asyncable_jobs_controller.rb
pkarman: you are right! :doh:! fixed

app/models/concerns/asyncable.rb
```diff
@@ -113,6 +111,10 @@ def submitted?
     !!self[self.class.submitted_at_column]
   end
 
+  def submitted_at_dtim
```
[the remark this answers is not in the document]
shanear: Hmm that's def not clear. Why is another column needed?

PR #8801, opened 2019-01-18
Replacing HearingCoordinator and Judges Dropdowns

lomaxap:
Much larger PR than I was intending.

I have been frustrated by having to find the value AND label for a dropdown before I can set a value. Let me know what you all think about these updates to resolve that:

1. All of these dropdowns take only a string as the value
2. the `onChange` method that you define is passed the `value` (as string) and `label` as separate parameters (e.g `onChange={(value, label)=> doSomething(value, label)}`). This way, you can use the `label` in your component if you need to, but you do not need to pass it back to the selector's value prop
3. Added `validateValueOnMount` property to RegionalOffice and HearingDate. This just ensures the initial value you pass is a valid option and runs `onChange` if it is

Smaller changes: 
- Moves all the dropdowns / reducers from `/hearingSchedule` directory to the `/common` directory. 
- Adds "Central" as an RO in the `RegionalOffice.rb`


client/app/components/DataDropdowns/HearingCoordinator.jsx
```diff
@@ -0,0 +1,96 @@
...
+    if (options || isFetching) {
+      return;
+    }
+
+    this.props.onFetchDropdownData('hearingCoordinators');
+
+    ApiUtil.get('/users?role=HearingCoordinator').then((resp) => {
+      const coordinatorOptions = _.values(ApiUtil.convertToCamelCase(resp.body.coordinators)).map((coor) => ({
+        label: coor.fullName,
+        value: coor.cssId
```
lomaxap: What value should be saved to the db for coordinator (bva_poc) fields?

sharonwarner: For now, let's leave it as the coordinator's name. In the future, we should refactor this to be foreign keyed to our users table.

app/controllers/regional_offices_controller.rb
```diff
@@ -1,7 +1,7 @@
 class RegionalOfficesController < ApplicationController
   def index
     render json: {
-      regional_offices: RegionalOffice.ros_with_hearings
+      regional_offices: RegionalOffice.ros_with_hearings.merge("C" => RegionalOffice::CITIES["C"])
```
sharonwarner: Would it make more sense to just add central office as a regional office?

app/models/regional_office.rb
```diff
@@ -168,14 +168,22 @@ class NotFoundError < StandardError; end
...
     timezone: "America/New_York",
     hold_hearings: true,
     facility_locator_id: "vba_314",
+    alternate_locations: ["vba_372"] },
+    "C" =>
+  { label: "Central",
+    city: "Washington",
+    state: "DC",
+    timezone: "America/New_York",
+    hold_hearings: false,
```
sharonwarner: Ah, looks like you did. Can we just set hold_hearings to true?

lomaxap: setting as a regional office with hearings caused [this `validate_ros_with_hearings` method](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/services/hearing_schedule/validate_ro_spreadsheet.rb#L43-L57) to fail tests

Issue #8675, opened 2019-01-13
Intake | Store the disability code for rating request issues


shanear:
Queue needs the ability to see the disability code for contested rating request issues. We should store these when the rating request issue is created.

## Acceptance Criteria
- When saving a rating request issue on an Appeal, the disability code of the contested rating issue should saved with the rating request issue (if it is present)

## Technical Notes
- To maintain consistency between the other columns, it's probably best to name the column `contested_rating_issue_disability_code`.

- This will require a little more complex parsing of the rating information that is returned from BGS.

- The disability code is stored on the disability evaluations of the disability structure passed from BGS. There can be multiple disability evaluations (and thus multiple disability codes) per rating issue. Always use the latest one.

- Not all rating issues involve disabilities. If it doesn't do not store any code.

## Validation instructions
- Intake an Appeal with one or more rating issues. On the case details page, the diagnostic codes of the contested rating issues should show up.

jiahuang:
Relevant example data blob from `Veteran.ratings.map{ |rating| rating.send(:rating_profile) }`

Note that this doesn't seem to be under each individual rating issue

```
"disabilities": [
      {
        "call_id": "17",
        "jrn_dt": "2018-09-19T10:52:02.000-05:00",
        "jrn_lctn_id": "346",
        "jrn_obj_id": "VBMS-cp_dsblty2_pkg.do_update",
        "jrn_stt_tc": "U",
        "jrn_user_id": "RMENDO_RVSR",
        "parent_name": "CP_RBA_PRFIL2",
        "row_cnt": "2",
        "row_id": "1",
        "additional_decision_bases": {
          "call_id": "17",
          "jrn_dt": "2018-07-17T08:31:44.000-05:00",
          "jrn_lctn_id": "328",
          "jrn_obj_id": "VBMS-cp_dsblty_decn_pkg.do_creat",
          "jrn_stt_tc": "I",
          "jrn_user_id": "AR328_RVSR",
          "parent_id": "1",
          "parent_name": "CP_DSBLTY2",
          "row_id": "5",
          "dsblty_decn_basis_tc": "INC",
          "dsblty_decn_sn": "30071",
          "dsblty_dt": "2018-07-17T08:31:43.000-05:00",
          "dsblty_id": "1122656"
        },
        "combat_ind": "N",
        "decn_basis_tc": "INC",
        "decn_basis_tn": "Incurred",
        "decn_tc": "SVCCONNCTED",
[...]

PR #8804, opened 2019-01-18
[WIP] add disability code to ratings

jiahuang:
Connects #8675

### DIFF
--- a/app/models/rating_issue.rb
+++ b/app/models/rating_issue.rb
@@ -5,7 +5,7 @@ class RatingIssue
   include ActiveModel::Model
 
   attr_accessor :reference_id, :decision_text, :profile_date, :associated_end_products,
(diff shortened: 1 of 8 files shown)


app/models/rating_issue.rb
```diff
@@ -18,7 +18,8 @@ def from_bgs_hash(rating, bgs_data)
         decision_text: bgs_data[:decn_txt],
         associated_end_products: rating.associated_end_products,
         promulgation_date: rating.promulgation_date,
-        participant_id: rating.participant_id
+        participant_id: rating.participant_id,
+        disability_code: bgs_data[:disability_code]
```
leikkisa: I think this is a little bit misleading, since disability_code isn't part of the BGS data.

PR #8806, opened 2019-01-18
Enable judges to return cases to attorneys for corrections

### LINKED ISSUE #8233: Add "Return to attorney for corrections" option for JudgeDecisionReviewTasks
After a decision has been drafted judges will occasionally return cases to attorneys to make corrections or rewrite the decision. We need to support this action in Caseflow for AMA cases. On 12 December 2018, we manually accomplished the same result by updating the judge task's type and status and the attorney task's status (see the [related slack conversation](https://dsva.slack.com/archives/C6E41RE92/p1544627868008000)). For this ticket we will implement a more robust solution explicitly creating a specific `AttorneyRewriteTask` in this scenario.

## Acceptance criteria
* [ ] Judges have an option to return a case to an attorney after that attorney has drafted a decision on the case
* [ ] Returning the case to the attorney creates a new `AttorneyRewriteTask`

tomas-nava:
Resolves #8233

### Description
Cases that've had decisions drafted by an attorney now have a "Return to attorney" task available to the judge. This creates a task (of new type `AttorneyRewriteTask`) that allows the attorney to review and re-submit the decision.

### Outstanding
- [x] If there are completed `AttorneyTasks` and current `AttorneyRewriteTasks`, the case will show up in both the _Assigned_ and _Completed_ tabs on the attorney's queue. **We'll file this as a separate ticket.**
- [x] Should the `AttorneyRewriteTask` have a different label in the case timeline? "Decision resubmitted by attorney"? **For the time being, we'll leave it with the same label. We may want to hide it from the case timeline altogether.**


app/models/task.rb
```diff
@@ -269,6 +269,18 @@ def schedule_veteran_data
     }
   end
 
+  def return_to_attorney_data
+    judge_task = root_task.children.find { |t| t.is_a?(JudgeTask) }
```
mdbenjam: Isn't this action defined on a judge task? Seems to me like we just want to use the `assign_to_user_data` function we already have? The task should already be assigned to the judge, so I don't think we need to do anything fancy to figure out which judge it's assigned to. In fact this could find the wrong judge task if there are multiple

client/app/queue/QueueActions.js
```diff
@@ -333,6 +333,39 @@ export const setSelectionOfTaskOfUser =
...
+  };
+
+  return ApiUtil.post(url, params).
+    then((resp) => resp.body).
+    then((resp) => {
+      dispatch(onReceiveAmaTasks(
+        resp.tasks.data
+      ));
+
+      dispatch(setSelectionOfTaskOfUser({
```
mdbenjam: Is this to maintain checkboxes on the judge assign page?

spec/support/feature_helper.rb
```diff
@@ -1,23 +1,41 @@
...
+  end
+
   def generate_words(n_words)
     Array.new(n_words).map do
       word_length = [rand(12), 3].max
       generate_text(word_length)
     end.join(" ")
   end
 
+  def step(_title)
```
mdbenjam: I like it. Maybe print out some test output indicating the step is completed? I'm not sure. Either way it's a good organizational step.

client/app/queue/QueueActions.js
tomas-nava: Nope, that was a sloppy copy/paste! Removing it now.

app/models/task.rb
tomas-nava: This is fixed with [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/8806/commits/3006a786d81f893e68600697349d6e5545e159b8).

client/app/queue/QueueActions.js
tomas-nava: This is fixed with [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/8806/commits/12ff4094e2a4048a93bf04198ca841b61f9ac3b1).

spec/support/feature_helper.rb
tomas-nava: I added the ability to include step output by passing a flag to rspec in [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/8806/commits/e1d4c2c41ed791ce6b623a78aca454378964b127).

example:
`SHOW_STEPS=true bundle exec rspec spec/feature/queue/ama_queue_spec.rb:583`

app/models/task.rb
```diff
@@ -270,10 +270,9 @@ def schedule_veteran_data
   end
 
   def return_to_attorney_data
-    judge_task = root_task.children.find { |t| t.is_a?(JudgeTask) }
-    judge = judge_task&.assigned_to
-    assignee = judge_task&.children&.select { |t| t.is_a?(AttorneyTask) }&.max_by(&:created_at)&.assigned_to
-    attorneys = JudgeTeam.for_judge(judge)&.attorneys || []
+    assignee = children.select { |t| t.is_a?(AttorneyTask) }.max_by(&:created_at)&.assigned_to
+    attorneys = JudgeTeam.for_judge(assigned_to)&.attorneys || []
+    attorneys |= [assignee] if assignee.present?
```
mdbenjam: I've never seen `|=` didn't know XOR worked on arrays in ruby. Pretty cool!

app/models/task.rb
```diff
@@ -269,6 +269,17 @@ def schedule_veteran_data
     }
   end
 
+  def return_to_attorney_data
+    assignee = children.select { |t| t.is_a?(AttorneyTask) }.max_by(&:created_at)&.assigned_to
+    attorneys = JudgeTeam.for_judge(assigned_to)&.attorneys || []
+    attorneys |= [assignee] if assignee.present?
+    {
+      selected: assignee,
```
lowellrex: What happens if the previous `AttorneyTask`'s assignee is no longer in the same judge team? I think this will be rare, but maybe it is worth considering?

client/app/queue/QueueActions.js
```diff
@@ -333,6 +333,33 @@ export const setSelectionOfTaskOfUser =
     }
   });
 
+export const returnCaseToAttorney = ({
```
lowellrex: Where does this function get called?

app/models/task.rb
tomas-nava: If the previous assignee isn't in the judge's team, they will still be available in the dropdown (the `|=` on line `275` merges/dedupes the assignee into the attorneys array)

client/app/queue/QueueActions.js
tomas-nava: ugh 😅good catch; that was a vestige of an abandoned solution I was trying. I removed it with [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/8806/commits/1f8b2f420e6eb887aca0483652c6cfef9e9680c5).

lowellrex:
Outstanding issue one (cases showing up in both the _assigned_ and _completed_ tabs) ticketed as #8973.

PR #8812, opened 2019-01-18
issue.eligibleForSocOptIn is undefined for RequestIssues on edit

pkarman:
connects #7337 

### Description
Fix subtle JS bug where `undefined !== false`.

This is because the same code gets applied to `ContestedIssue` and `RequestIssue` when both creating and editing an Intake. On edit, the `RequestIssue` has an undefined value for `eligibleForSocOptIn` and was triggering an ineligible warning message for the issue, even though it was eligible.


pkarman:
> is it worth changing all of them to === false?

probably not. this is a weird case where `undefined` because we are re-using code that expects a particular data structure. In this case, there's some circular logic around the relationship between the VACOLS issue, the Rating issue, and the RequestIssue that only shows up when you try and edit a RequestIssue that's been associated. It can't be opted in because it is already opted in, and yet we don't want to show the error message that normally applies when you can't opt something in.

Issue #8807, opened 2019-01-18
Legacy Appeals should be serialized with the type LegacyAppeal


cmgiven:
Currently, legacy appeals returned by the API use the type "AppealSeries", reflecting the model being serialized. This implementation detail should not be exposed. Instead, it should return "LegacyAppeal," in preparation for clients needing to use this information to differentiate types of decision reviews.

PR #8814, opened 2019-01-18
Use legacyAppeal instead of appealSeries as the type in API responses

cmgiven:
Connects #8807


cmgiven:
Yup. So the JSON object that gets returned is in the format:

```
{
  "id": 1
  "type": "legacyAppeal"
  "attributes": { ... }
}
```

We want to change `type` in the root object. There's also a `type` in `attributes`, which is what the `attribute :type_code, key: :type` is adding.

Issue #8771, opened 2019-01-16
Display correct EP codes & EP statuses for HLR / SCs in search results


joeyyang:
from @shanear:

<img width="1029" alt="screen shot 2019-01-15 at 2 01 57 pm 1" src="https://user-images.githubusercontent.com/2928395/51275497-ffd56d00-1986-11e9-9b38-8596e38996a4.png">

1. ~~If there is no EP, don't show `null` (fixed in https://github.com/department-of-veterans-affairs/caseflow/pull/8817)~~
1. If the EP is not yet established (`established_at == nil`), (https://github.com/department-of-veterans-affairs/caseflow/pull/8768)
    1. and there is any value in `establishment_error`, the status should read "Establishment failed. Support notified."
    1. and there is no `establishment_error`, the status should read "Establishing..." 
1. ~~Remove "Decision Date" column because data is unreliable (fixed in https://github.com/department-of-veterans-affairs/caseflow/pull/8818)~~
1. ~~remove duplicate `Appellant Name`, if any (https://github.com/department-of-veterans-affairs/caseflow/pull/8867)~~

PR #8817, opened 2019-01-18
default EP modifier to empty string if it is null

joeyyang:
Resolves (partially) https://github.com/department-of-veterans-affairs/caseflow/issues/8771

Don't allow an EP modifier to come through to the frontend as `null` by setting null values to `""`.


joeyyang:
thanks @kencheeto , back to you!

Issue #8456, opened 2018-12-27
Hearing Details (for AMA hearings)


MeredithStewart:
## User Stories
* As part of hearing process, Hearing Coordinators sometimes need to update data for individual hearings, including room location, VLJ, and hearing coordinator. 
* As part of the hearing process, data around transcription for individual hearings also needs to be recorded and updated as needed.

## Mocks
* Page 4 of https://va-gov.invisionapp.com/share/TKPRX5QUMRW#/screens 

## Acceptance Criteria - Hearing Details 
- [x] Verify that the Hearing Details are only shown for AMA hearings. 
- [x] Verify that the user can select the date of the hearing and can access the daily docket for that hearing day. 
- [x] Verify that the user can see hearing details that are not editable in this view, including docket type. See mocks for data included.  
- [x] Verify that the user can edit the VLJ, Hearing Coordinator, and Hearing Room so that it is out of sync with the overall hearing day parent record. 
- [x] Verify that changes to the overall hearing day (vlj, coodinator, room) will write to children, overwriting previous changes made at the child record level. 
- [x] Verify that the user can select VLJ from a list of VLJs. 
- [x] Verify that the user can select hearing coordinator from a list of coordinators. 
- [x] Verify that the user can select a room from the list of all rooms. 
- [x] Verify that the user can see and edit the notes on the individual hearing - notes field for the child record. 
- [x] Verify that the user can select that the Veteran has waived their 90 day evidence hold. 
- [x] Verify that the default is unselected for the 90 day evidence hold. 
- [x] Verify that the 90 day evidence hold reflects updates made to this field in Hearing Prep. 
 
## Acceptance Criteria - Transcription Details  
- [x] Verify that the user can manually enter a task number (letters, numbers, dashes) 
- [x] Verify that the user can select from a list of available transcribers: 
   * Genesis Government Solutions, Inc.
   * Jamison Professional Services  
   * The Ravens Group, Inc.
- [x] Verify that the user can enter or select the date, "Sent to Transcriber," with a default unselected and user able to revert to unselected.  
- [x] Verify that the user can enter or select the date, "Expected Return Date," with a default unselected and user able to revert to unselected.    
- [x] Verify that the user can enter or select the date, "Transcript Uploaded," with a default unselected and user able to revert to unselected.  
- [x] Verify that the user can select Transcription Problem Type from list: 
       * No audio 
       * Poor Audio Quality
       * Incomplete Hearing
       * Other (see notes)
[...]

PR #8821, opened 2019-01-18
Hearing Details

lomaxap:
Resolves mostly #8456 

- Creates `/hearings/:hearingId/details` route with form
- Creates Transcription table


app/controllers/hearings_controller.rb
```diff
@@ -4,21 +4,20 @@ class HearingsController < ApplicationController
...
-    end
+    slot_new_hearing
 
     if hearing.is_a?(LegacyHearing)
       hearing.update_caseflow_and_vacols(update_params_legacy)
       # Because of how we map the hearing time, we need to refresh the VACOLS data after saving
       HearingRepository.load_vacols_data(hearing)
     else
       hearing.update!(update_params)
+      update_transcription if params["transcription"]
```
sharonwarner: I think we can make this a little cleaner by following the pattern in hearings/appeals_controller.rb. We can create the transcription record in the show method, and then update the transcription record from the hearing directly.

client/app/hearingSchedule/components/Details.jsx
```diff
@@ -0,0 +1,203 @@
...
+
+export default class HearingDetails extends React.Component {
+
+  constructor(props) {
+    super(props);
+
+    const { hearing } = this.props;
+    const transcription = hearing.transcription || {};
+
+    this.state = {
```
sharonwarner: I think this makes more sense to store and edit through redux. Generally over the past year, we've tried to store/edit everything through redux. Over the past few months we've started moving ui state out of redux, but I still think it makes sense to edit data that we send/receive from the backend through redux.

db/migrate/20190117151019_add_hearing_day_details_to_hearings.rb
```diff
@@ -0,0 +1,20 @@
...
+    t.belongs_to :hearing, index: true
+    t.string :task_number
+    t.string :transcriber
+    t.date :sent_to_transcriber_date
+    t.date :expected_return_date
+    t.date :uploaded_to_vbms_date
+    t.string :problem_type
+    t.date :problem_notice_sent_date
+    t.string :requested_remedy
+    t.boolean :copy_requested
```
sharonwarner: This field is already set on the hearing as transcript_requested. You can pick which field you want to keep, though if we decide to keep this field on the transcriptions table, we'll have to update hearing prep to use this field for ama hearings. It's probably easier to keep the other field, but up to you!

Issue #8776, opened 2019-01-16
Update Add Decisions page in the attorney-judge checkout flow


allyceh:
This ticket is a new ticket linked to #7965. 

### Latest Mocks

Updates to make to Add Decisions screen: 
- Add diagnostic code field to each request and decision issue
- Add Delete button to the left of Edit button to delete the decision issue

![image](https://user-images.githubusercontent.com/40216950/50249195-50b76d00-03ab-11e9-8b96-ead49381e951.png)

PR #8823, opened 2019-01-18
8776 update add decisions page

### LINKED ISSUE #8778: Update Edit Decision modal and add "delete" modal for decision issues
This ticket is linked to #7966 . 

## Latest Mocks

Edit modal

- Remove delete button from modal
- Added diagnostic code field
- Revised hierarchy of fields (order should be: Dispo, Description, Diagnostic Code, Benefit Type
- When we display the drop-down field to add the decision to other issues, updated copy should read: "Do you want to add this decision to other issues on appeal? ...."

image

Modal for deleting one decision issue
- Display modal when user clicks "delete" on the Add Decisions page

image

Modal for deleting 2+ decision issues 
- Display modal when user clicks "delete" on the Add Decisions page, and the decision is attached to 2 or more issues. 

image

youngfreezyVA:
Resolves #8778 and #8776

### Description
Made the changes described in this ticket:
https://github.com/department-of-veterans-affairs/caseflow/issues/8778

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. run `bundle exec rails runner scripts/enable_features_dev.rb`
2.  login as judge or attorney
3. select AMA appeal
4. select decision ready for review
5. no special issues
6. on add decisions page, add a decision to any issue
7. verify a delete option now appears next to the edit button:
screen shot 2019-01-18 at 3 35 19 pm
8. on clicking delete, verify this modal appears and that the 'cancel' and 'Yes, delete decision' buttons work:
screen shot 2019-01-18 at 3 36 21 pm
9. verify that the appropriate text appears if there are connected issues associated with the issue you are trying to delete. (you can check if there are connected issues by seeing if there is a 'select issues' dropdown when you go to add an issue)


youngfreezyVA:
> The typeface looks like Helvetica rather than source sans pro. Confirming that will be updated before we push to prod.

Hi @allyceh,

so I have control over the copy that appears in the modal, but to change the modal's title and buttons font, I would need to make changes to the modal component itself. let me know if I'm missing something, but if that's what needs to be done, I think a separate task should be made for that, and I can go ahead and make these changes to the modal component (which will impact all modals).

also, please let me know if there is any sort of UX tool we use to inspect the specs of mockups like fonts, etc (such as Invision).

thanks!

client/app/queue/components/ContestedIssues.jsx
```diff
@@ -81,14 +82,26 @@ export default class ContestedIssues extends React.PureComponent {
...
-            {openDecisionHandler && <span>
-              <Button
-                name="Edit"
-                id={`edit-issue-${requestIssue.id}-${decisionIssue.id}`}
-                onClick={openDecisionHandler(requestIssue.id, decisionIssue)}
-                classNames={['cf-btn-link']}
-              />
-            </span>}
+            <div>
+              {this.props.isAmaAppeal && openDeleteAddedDecisionIssueHandler && <span>
```
mdbenjam: I think this will always be an AMA appeal on this screen. So I'm not sure that we need to test for that.

client/app/styles/_main.scss
```diff
@@ -388,3 +388,8 @@ select {
   margin-left: 1.5em;
   text-indent: -1.7em;
 }
+
+.delete-decision-modal {
```
mdbenjam: I'm trying to get us to add less CSS, is this styling really necessary?

client/app/queue/components/ContestedIssues.jsx
youngfreezyVA: ok cool, thought that might be the case, will update

client/app/styles/_main.scss
youngfreezyVA: I think this CSS might go away depending on @allyceh's response to the modal changes in my comment above :)

PR #8829, opened 2019-01-19
add app tagging for debugging

kencheeto:
It's hard to to map latency issues spotted in Datadog to specific parts of the application. It's also hard to differentiate which data points originate from user interaction vs jobs (where latency is less critical). In order to make debugging these a bit easier, I'm doing two things:

1. tagging the application name in Dispatch controllers
1. reading `application_attr` inside the base `ApplicationJob#before_perform` method to set the application name
1. for jobs, appending `_job` so we can differentiate between user interaction and jobs

Jobs that want to explicitly use a different app name still can by setting it in `perform` (see https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/jobs/process_decision_document_job.rb)

I'm wary of breaking any alerting/reporting that is observing application names - I can just use the application name directly if that is the case.

### DIFF
--- a/app/controllers/dispatch/tasks_controller.rb
+++ b/app/controllers/dispatch/tasks_controller.rb
@@ -2,7 +2,7 @@ class Dispatch::TasksController < ApplicationController
   class InvalidTaskClassError < StandardError; end
   class InvalidTaskStateError < StandardError; end
 
-  before_action :check_dispatch_out_of_service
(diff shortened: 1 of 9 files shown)


lisac-usds:
connecting this to issue https://github.com/department-of-veterans-affairs/caseflow/issues/8852 - it's my understanding this app tagging came about while investigating that issue.

PR #8848, opened 2019-01-21
AMA hearing worksheet

### LINKED ISSUE #8475: Hearing Prep Hearing Worksheet Updates
## User Stories 
* As a VLJ, I need to be able to determine when an appeal is an AMA appeal. 
* As a VLJ, I need to be able to take notes on AMA issues and have options only available for AMA appeals. 
* As a VLJ, I need to be able to correct issues through Intake if I find an administrative error was made. 
* As a VLJ, I need to be able to see the Case Details for each Veteran. 

## Mocks 
* For mocks see (page 3): https://va-gov.invisionapp.com/share/JMPS9DG29DA#/338987658_Worksheet_Current_Issues-Issues

## Acceptance Criteria 
- [x] Verify user has a link to the "View Case Details" for that Veteran. Verify that the button/link opens a new tab in the browser. Note: This is the same behavior as the "Review Claims Folder" button.  Verify that "View Case Details" button is implemented for both AMA and Legacy appeals. 
- [x] Verify user can see the docket number and docket type, reflecting H for AMA appeals and L for Legacy appeals. 
- [x] Verify that Procedural History is removed for AMA appeals. Note: Legacy appeals continue to have Procedural History section. 
- [x] Verify for AMA appeals, the new Issues section of the Hearings Worksheet is formatted similar to the Case Details Issues section. 
- [ ] Verify for AMA appeals, the user can select, "Correct Issues," which brings the user to a modal within Intake that allows the VLJ to correct administrative errors.  Note: URL is the same one that will be implemented on Case Details. This treatment will be a secondary button.
- [ ] Verify that after VLJ closes the "Correct Issues" modal, that the user is brought back to the Hearing Prep Hearing Worksheet. Note: While in Modal, the background will appear as if they are in Intake. 
- [x] Verify that the user can edit in-line for each AMA issue the, "Hearing Worksheet Notes," and Preliminary impressions. 
- [x] Verify that for issues on AMA appeals the user does not have the OMO option, ability to remove the issue (trash can), or the ability to add a new issue. Note: These options are retained for legacy appeals. 
- [x] Verify that for AMA appeals, the limitation does not exist of having to pull issues 30 days prior to the hearing like it does for Legacy appeals. Note: Because AMA doesn't have the same editing issues functionality, the issues section for AMA should dynamically update as changes are made to issues through Intake. It needs to function this way. We assume the 30 days in advance pulling in issues does not appeal to AMA hearing worksheets.
- [x] Verify the title is re-worded to say "Veteran/Appellant Information" above the information in both Legacy and AMA Hearing Prep views.
- [x] Verify all headings are h4 tags above the case, vet, VLJ, etc. info on both Legacy and AMA Hearing Prep views.
- [x] Verify the "Representative Org." title is changed to "Power of Attorney" in both Legacy and AMA Hearing Prep views. 
- [x] Verify the "Docket Number" title is changed to "Docket" in both Legacy and AMA Hearing Prep views. 

## Epic 
* Part of Epic #8473

sharonwarner:
Resolves #8475 

### Description
This PR updates the hearing worksheet to work for AMA appeals.

### Testing Plan
1. Create an AMA hearing, go to the hearing worksheet, confirm functionality works as expected
1. Create a legacy hearing, go to the hearing worksheet, confirm functionality works as expected


client/app/components/ContentSection.jsx
```diff
@@ -0,0 +1,34 @@
...
+});
+
+export const sectionSegmentStyling = css({
+  border: `1px solid ${COLORS.GREY_LIGHT}`,
+  borderTop: '0px',
+  marginBottom: '3rem',
+  padding: '1rem 2rem'
+});
+
+export default class ContentSection extends React.Component {
```
sharonwarner: This is all copied almost directly from StickyNavContentArea. I didn't want to include that refactor in this PR, but I think that the StickyNavContentArea could use this component.

client/app/hearings/HearingWorksheet.jsx
```diff
@@ -114,6 +115,32 @@ export class HearingWorksheet extends React.PureComponent {
...
+    return <div>
+      <HearingWorksheetDocs {...this.props} />
+      <HearingWorksheetStream {...this.props} print={this.props.print} />
+    </div>;
+  };
+
+  getHearingWorksheet = () => {
+    let worksheetIssuesArray = [];
+
+    _.forEach(this.props.worksheetIssues, (value) => {
```
sharonwarner: 😭 

The ContestedIssues component expects the issues to be an array, the hearing prep state expects them to be a hash keyed by id. Changing hearing prep would be a major refactor, and I don't want to update the ContestedIssues component any more than I have to.

client/app/hearings/components/HearingWorksheetAmaIssues.jsx
```diff
@@ -0,0 +1,80 @@
...
+            minRows={5}
+            maxRows={8}
+            maxLength={300}
+          />
+        </div></div>,
+      preliminaryImpressions: <HearingWorksheetPreImpressions ama issue={issue} />
+    }];
+
+    return <div>
+      <Table
```
sharonwarner: The table is basically unnecessary -- except that's how the existing styling works. Can refactor later.

client/app/hearings/HearingWorksheet.jsx
lomaxap: i think you can use `_.values(this.props.worksheetIssues)`

client/app/hearings/actions/Issue.js
```diff
@@ -165,8 +173,16 @@ export const toggleIssueDeleteModal = (issueId, isShowingModal) => ({
 export const saveIssues = (worksheetIssues) => (dispatch) => {
   _.forEach(worksheetIssues, (issue) => {
```
lomaxap: might be time constraint on implementing this, but wouldn't it be better to have an endpoint that takes multiple worsheetIssues rather than looping through and calling the endpoint on each loop? Fewer calls/responses and db transactions.

sharonwarner: yes! will create a tech backlog ticket for this (and the other things i commented above).

client/app/hearings/HearingWorksheet.jsx
```diff
@@ -114,6 +115,26 @@ export class HearingWorksheet extends React.PureComponent {
...
+      <HearingWorksheetDocs {...this.props} />
+      <HearingWorksheetStream {...this.props} print={this.props.print} />
+    </div>;
+  };
+
+  getHearingWorksheet = () => {
+    return <div className="cf-hearings-worksheet-data cf-hearings-worksheet-issues">
+      <ContentSection
+        header={<div>Issues</div>}
+        content={<ContestedIssues
```
mdbenjam: nit: take the content as a child component. It just makes it more fluid.

client/app/hearings/actions/Issue.js
```diff
@@ -165,8 +173,16 @@ export const toggleIssueDeleteModal = (issueId, isShowingModal) => ({
...
     if (issue.edited) {
-      ApiUtil.patch(`/hearings/appeals/${issue.appeal_id}`, { data: { appeal: {
-        worksheet_issues_attributes: [issue] } } }).
+
+      let url = `/hearings/appeals/${issue.appeal_id}`;
+      let data = { appeal: { worksheet_issues_attributes: [issue] } };
+
+      if (issue.docket_name === 'hearing') {
+        url = `/hearings/worksheets/${issue.hearing.external_id}`;
```
mdbenjam: So we have two different endpoints for AMA/legacy?

client/app/hearings/components/HearingWorksheetAmaIssues.jsx
```diff
@@ -0,0 +1,83 @@
...
+import Table from '../../components/Table';
+
+const tableRowStyling = css({
+  '& > tr': {
+    '& > td:nth-child(1)': { width: '50%' },
+    '& > td:nth-child(2)': { width: '50%' }
+  }
+});
+
+class HearingWorksheetAmaIssues extends PureComponent {
```
mdbenjam: Is this just a copy of an existing component tweaked to work for AMA?

db/migrate/20190118155859_create_hearing_issue_notes.rb
```diff
@@ -0,0 +1,14 @@
+class CreateHearingIssueNotes < ActiveRecord::Migration[5.1]
+  def change
+    create_table :hearing_issue_notes do |t|
```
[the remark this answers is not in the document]
sharonwarner: Would love to talk through this... worksheet_issues and hearing_issue_notes are so different that I don't think it makes sense to try to combine them into one table. I think ideally we would have legacy_worksheet_issues and worksheet_issues, but there wasn't enough time to get all those updates in before the hearing today.

client/app/hearings/components/HearingWorksheetAmaIssues.jsx
sharonwarner: It's pretty different. Though I think at some point we should refactor the hearing worksheet to make the more similar. We think of worksheet_issues so differently for legacy appeals that it's really hard to combine functionality.

client/app/hearings/actions/Issue.js
sharonwarner: Yeah. I hate it, but it would take a huge refactor to fix it.

client/app/hearings/HearingWorksheet.jsx
sharonwarner: Makes sense. Will fix that today.

PR #8849, opened 2019-01-22
[WIP] adding hlr skeleton for appeal status

mc-usds:
Resolves #{8726}

### Description
Added in skeleton model for HLR. Updated the serializer to get the id and type from the object passed in. It is not hooked up to the path to serialize it.

### Acceptance Criteria
- [x] Code compiles correctly

### Testing Plan
tested on command line to see it is would serialize:

=> {:data=>
  [{:id=>"HLR9",
    :type=>"higherLevelReviews",
    :attributes=>
     {:appealIds=>[],
      :updated=>"2019-01-21T21:30:16-05:00",
      :incompleteHistory=>false,
      :type=>nil,
      :active=>true,
      :description=>nil,
      :aod=>nil,
      :location=>nil,
      :aoj=>nil,
      :programArea=>nil,
      :status=>nil,
      :alerts=>nil,
      :docket=>nil,
      :issues=>[],
      :evidence=>[]}},
   {:id=>"HLR10",
    :type=>"higherLevelReviews",
    :attributes=>
     {:appealIds=>[],
      :updated=>"2019-01-21T21:30:16-05:00",
      :incompleteHistory=>false,
      :type=>nil,
      :active=>false,
      :description=>nil,
      :aod=>nil,
      :location=>nil,
      :aoj=>nil,
      :programArea=>nil,
      :status=>nil,
      :alerts=>nil,
      :docket=>nil,
      :issues=>[],
      :evidence=>[]}}]}

### DIFF
--- a/app/models/higher_level_review.rb
+++ b/app/models/higher_level_review.rb
@@ -29,6 +29,69 @@ class HigherLevelReview < ClaimReview
     create_dta_supplemental_claim
   end
 
+## needed for appeal status api
+  def review_status_id
+    return "HLR#{id}"
+  end
+
+  def vacols_ids
+    #doesn't apply to HLRs
+    []
+  end
+
+  def incomplete
+    return false
+  end
+
+  def type_code
+    # TODO: add logic to return - original, post_remand, post_cavc_remand, reconsideration, cue
+  end
+
+  def active?
+    end_product_establishments.any? { |ep| ep.status_active?(sync: false) }
+  end
+
+  def description
+    #need to impelement
+  end
+
+  def aod
+    #doesn't apply to HLRs
+  end
+
+  def location
+    #will this always be aoj?
+  end
+
+  def aoj
+    #add logic to return proper enum: - vba, vha, nca, other 
+  end
+
+  def program
+    #maps to benefit_type?
+  end
+
+  def status_hash
+    #add logic to return: hlr_received, hlr_dta_error, hlr_decision, hlr_closed
+  end
+
+  def alerts
+    #add logic to return alert enum
+  end
+
+  def docket_hash
+    #doesn't apply to HLRs
+  end
+
+  def issues
+    #get request and corresponding rating issue
+    []
+  end
+
+  def events
+    # hlr_request, hlr_decision, hlr_dta_error, or hlr_other_close
+  end
+
   private
 
   def create_dta_supplemental_claim
--- a/app/models/serializers/v2/appeal_serializer.rb
+++ b/app/models/serializers/v2/appeal_serializer.rb
@@ -1,8 +1,19 @@
 class V2::AppealSerializer < ActiveModel::Serializer
-  type :legacy_appeal
+
+  def type
+    if appeal_status_v3_enabled?
+      object.class.to_s.camelize(:lower)
+    else
+      :legacy_appeal
+    end
+  end
 
   def id
-    object.vacols_id
+    if appeal_status_v3_enabled?
+      object.review_status_id
+    else
+      object.vacols_id
+    end
   end
 
   attribute :vacols_ids, key: :appeal_ids
@@ -31,5 +42,9 @@ class V2::AppealSerializer < ActiveModel::Serializer
   # Stubbed attributes
   attribute :evidence do
     []
+  end
+
+  def appeal_status_v3_enabled?
+    FeatureToggle.enabled?(:api_appeal_status_v3)
   end
 end
--- a/app/controllers/api/v2/appeals_controller.rb
+++ b/app/controllers/api/v2/appeals_controller.rb
@@ -37,6 +37,10 @@ class Api::V2::AppealsController < Api::ApplicationController
     end
   end
 
+  def hlrs
+    @hlrs ||= HigherLevelReview.where(veteran_file_number: vbms_id)
+  end
+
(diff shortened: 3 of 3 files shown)


app/models/higher_level_review.rb
```diff
@@ -29,6 +29,69 @@ def on_decision_issues_sync_processed(_end_product_establishment)
     create_dta_supplemental_claim
   end
 
+## needed for appeal status api
+  def review_status_id
+    return "HLR#{id}"
+  end
+
+  def vacols_ids
+    #doesn't apply to HLRs
```
cmgiven: It does (although we might want to rename the method name). For now, it should just be the `review_status_id` nested inside an array. If we start to do linking and combining of reviews, this array would also have other IDs the review was formerly identified under.

app/models/higher_level_review.rb
```diff
@@ -29,6 +29,69 @@ def on_decision_issues_sync_processed(_end_product_establishment)
...
+    []
+  end
+
+  def incomplete
+    return false
+  end
+
+  def type_code
+    # TODO: add logic to return - original, post_remand, post_cavc_remand, reconsideration, cue
```
cmgiven: This does not apply to HLRs.

app/models/higher_level_review.rb
```diff
@@ -29,6 +29,69 @@ def on_decision_issues_sync_processed(_end_product_establishment)
...
+  def description
+    #need to impelement
+  end
+
+  def aod
+    #doesn't apply to HLRs
+  end
+
+  def location
+    #will this always be aoj?
```
cmgiven: Yup!

app/models/higher_level_review.rb
```diff
@@ -29,6 +29,69 @@ def on_decision_issues_sync_processed(_end_product_establishment)
...
+  end
+
+  def aoj
+    #add logic to return proper enum: - vba, vha, nca, other 
+  end
+
+  def program
+    #maps to benefit_type?
```
cmgiven: Yes indeed.

app/models/higher_level_review.rb
```diff
@@ -29,6 +29,69 @@ def on_decision_issues_sync_processed(_end_product_establishment)
...
+  end
+
+  def status_hash
+    #add logic to return: hlr_received, hlr_dta_error, hlr_decision, hlr_closed
```
cmgiven: Note that this method returns the `details` object for a given status. The status type should be determined elsewhere.

app/models/serializers/v2/appeal_serializer.rb
```diff
@@ -32,4 +43,8 @@ def id
   attribute :evidence do
     []
   end
+
+  def appeal_status_v3_enabled?
+    FeatureToggle.enabled?(:api_appeal_status_v3)
```
cmgiven: Woot!

app/models/serializers/v2/appeal_serializer.rb
```diff
@@ -1,8 +1,19 @@
...
+    if appeal_status_v3_enabled?
+      object.class.to_s.camelize(:lower)
+    else
+      :legacy_appeal
+    end
+  end
 
   def id
-    object.vacols_id
+    if appeal_status_v3_enabled?
```
cmgiven: Could we just rename this in AppealSeries so that we don't need the feature flag here?

app/models/serializers/v2/appeal_serializer.rb
```diff
@@ -1,8 +1,19 @@
 class V2::AppealSerializer < ActiveModel::Serializer
-  type :legacy_appeal
+
+  def type
+    if appeal_status_v3_enabled?
+      object.class.to_s.camelize(:lower)
```
cmgiven: This will return `appeal_series` instead of `legacy_appeal` for legacy appeals.

We use snake case for values. We use camel case for keys. Yes, this is weird and shouldn't be the case.

Also note that this returns a plural, `higherLevelReviews`, instead of the anticipated `higher_level_review`. If that's preferable, we should just switch the Swagger spec to use the plurals.

mc-usds: It doesn't work and I don't know why!!!! That def is completely ignored. I can't figure out how to get it to work in a reasonable amount of time, so created a new serializer and override it there.

app/models/serializers/v2/appeal_serializer.rb
mc-usds: Removed this logic and am overriding in the new HLR serializer.

app/models/higher_level_review.rb
mc-usds: Ok, will implement in a separate PR.

app/models/higher_level_review.rb
mc-usds: Implemented. Added an new issue so I remember to add "fiduciary" to the schema (8861).

app/models/higher_level_review.rb
mc-usds: moved the the HLR status serializer since it will always be the same value.

app/models/higher_level_review.rb
mc-usds: Fixed. Created a new serializer for HLR status and override the attribute there.

PR #8850, opened 2019-01-22
Seperate document counts for efolder and and caseflow

### LINKED ISSUE #8820: When on the queue page use document count from Caseflow, when on case details page call VBMS for the document count

hschallhorn:
Resolves #8820

### Description
* Show efolder count in case details
* Show caseflow count in queue

### Acceptance Criteria
- [ ] When a user navigates to their queue page, we should use a document count from Caseflow because all appeals on this page are assigned to a current user
- [ ] When a user navigates to the case details page, we should use a document count from VBMS

### DIFF
--- a/app/controllers/appeals_controller.rb
+++ b/app/controllers/appeals_controller.rb
@@ -2,7 +2,7 @@ class AppealsController < ApplicationController
   include Errors
 
   before_action :react_routed
-  before_action :set_application, only: [:document_count, :new_documents]
+  before_action :set_application, only: [:document_count_from_efolder, :document_count_from_caseflow, :new_documents]
   # Only whitelist endpoints VSOs should have access to.
   skip_before_action :deny_vso_access, only: [:index, :power_of_attorney, :show_case_list, :show, :veteran]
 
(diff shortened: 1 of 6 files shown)


app/controllers/appeals_controller.rb
```diff
@@ -2,7 +2,7 @@ class AppealsController < ApplicationController
   include Errors
 
   before_action :react_routed
-  before_action :set_application, only: [:document_count, :new_documents]
+  before_action :set_application, only: [:document_count_from_efolder, :document_count_from_caseflow, :new_documents]
```
aroltsch: You should probably change it back :)

hschallhorn: Knew I'd miss something 🙏 TY

app/controllers/appeals_controller.rb
```diff
@@ -36,7 +36,11 @@ def show_case_list
   end
 
   def document_count
-    render json: { document_count: appeal.number_of_documents_from_caseflow }
+    if params[:cached].present? && params[:cached] == "true"
```
aroltsch: I think if you do this 
```if params[:cached]```
it should check the presence of the parameter and that is it set to true

Comment (hschallhorn) [reply app/controllers/appeals_controller.rb]: This is what I had at first, but the query params come in as strings, and `'false'` is truthy in ruby. I could at least get rid of `.present?` though.

File: app/controllers/appeals_controller.rb [hunk 3]
```diff
@@ -36,7 +36,11 @@ def show_case_list
   end
 
   def document_count
-    render json: { document_count: appeal.number_of_documents_from_caseflow }
+    if params[:cached] && params[:cached] == "true"
```
aroltsch: I think you can remove your first check. Also it is better to avoid if/else statements because they are not as easily readable. 
```
if params[:cached] == "true"
      render json: { document_count: appeal.number_of_documents_from_caseflow }
      return
    end
    render json: { document_count: appeal.number_of_documents }```

PR #8857, opened 2019-01-22
Fix date parsing on non comp dispositions page

pkarman:
The JS util method `formatDate` expects a full ISO8601 string. When you pass it a m/d/Y string it will parse it into the local browser time zone, which is typically the "day before" since the negative UTC offset will count backwards from the implicit midnight time.


pkarman:
@kencheeto agreed. Going to work on that in a second PR though, because I think there are other breaking instances in the intake utils.

PR #8859, opened 2019-01-22
kick users out of Completed before trying to read data

kencheeto:
Users can navigate directly to /completed - in this case we should redirect before trying to read some data that might not be there.

avoids https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3407/
and fixes https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3439/

### DIFF
--- a/client/app/intake/pages/decisionReviewIntakeCompleted.jsx
+++ b/client/app/intake/pages/decisionReviewIntakeCompleted.jsx
@@ -124,14 +124,18 @@ class DecisionReviewIntakeCompleted extends React.PureComponent {
     case INTAKE_STATES.STARTED:
       return <Redirect to={PAGE_PATHS.REVIEW} />;
     case INTAKE_STATES.REVIEWED:
-      return <Redirect to={PAGE_PATHS.FINISH} />;
+      if (formType === FORM_TYPES.RAMP_ELECTION.key || formType === FORM_TYPES.RAMP_REFILING.key) {
+        return <Redirect to={PAGE_PATHS.FINISH} />;
+      } else {
+        return <Redirect to={PAGE_PATHS.ADD_ISSUES} />;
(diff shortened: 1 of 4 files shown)


pkarman:
A test? Hard to tell what is really changing here since it looks like it's simply the order of the statements?

kencheeto:
I'm putting the redirect checks based on intake state first, rather than trying to iterate over request issues, which can be undefined if you visit /completed directly. I've also updated the redirect to properly redirect decision reviews back to /add_issues; we were also kicking users back to /finish which doesn't work for decision reviews anymore.

client/app/intake/pages/decisionReviewIntakeCompleted.jsx
```diff
@@ -117,20 +117,25 @@ class DecisionReviewIntakeCompleted extends React.PureComponent {
...
 
     switch (intakeStatus) {
     case INTAKE_STATES.NONE:
       return <Redirect to={PAGE_PATHS.BEGIN} />;
     case INTAKE_STATES.STARTED:
       return <Redirect to={PAGE_PATHS.REVIEW} />;
     case INTAKE_STATES.REVIEWED:
-      return <Redirect to={PAGE_PATHS.FINISH} />;
+      if (formType === FORM_TYPES.RAMP_ELECTION.key || formType === FORM_TYPES.RAMP_REFILING.key) {
+        return <Redirect to={PAGE_PATHS.FINISH} />;
```
jiahuang: ramp election goes to `PAGE_PATHS.FINISH` but ramp refiling does not? Just checking cuz idk

kencheeto: both should go to FINISH

Issue #8802, opened 2019-01-18
DTA Error supplemental claim establishment_submitted_at field is null


shanear:
Looking at the data for a DTA Error supplemental claim EP, it looks like the `establishment_submitted_at` column is null. This makes it hard to back track what happened :(
 
```
irb(main):026:0> pp SupplementalClaim.where(veteran_file_number: "124592531").first
[2019-01-17 19:12:51 -0500]   SupplementalClaim Load (1.3ms)  SELECT  "supplemental_claims".* FROM "supplemental_claims" WHERE "supplemental_claims"."veteran_file_number" = $1 ORDER BY "supplemental_claims"."id" ASC LIMIT $2  [["veteran_file_number", "124592531"], ["LIMIT", 1]]
#<SupplementalClaim:0x0000557b0bd4d498
 id: 271,
 veteran_file_number: "124592531",
 receipt_date: Wed, 16 Jan 2019,
 establishment_submitted_at: nil,
 establishment_processed_at: Fri, 18 Jan 2019 00:00:49 UTC +00:00,
 benefit_type: "compensation",
 establishment_attempted_at: Fri, 18 Jan 2019 00:00:26 UTC +00:00,
 establishment_error: nil,
 legacy_opt_in_approved: true,
 veteran_is_not_claimant: false,
 uuid: "6fc18a98-2934-4bf8-870e-054d180b123f",
 decision_review_remanded_type: "HigherLevelReview",
 decision_review_remanded_id: 879>
```

shanear:
looks good!
```
#<ActiveRecord::Associations::CollectionProxy [#<EndProductEstablishment id: 769, established_at: "2019-01-30 17:54:03", synced_status: "PEND", source_type: "SupplementalClaim", source_id: 328, veteran_file_number: "700035231", reference_id: "600148876", claim_date: "2019-01-29", code: "040HDER", modifier: "040", station: "397", last_synced_at: "2019-01-30 21:52:11", claimant_participant_id: "600247088", payee_code: "00", committed_at: "2019-01-30 17:54:06", doc_reference_id: nil, development_item_reference_id: nil, benefit_type_code: "1", user_id: 124>]>
```

PR #8864, opened 2019-01-22
Ignore nil submitted at when sorting asyncable jobs

pkarman:
connects #6944 

### Description
When the submitted_at column is nil, Ruby was throwing a bad comparison error on sort. This is due to bad data (see #8802) but work around it any way.

### DIFF
--- a/app/models/concerns/asyncable.rb
+++ b/app/models/concerns/asyncable.rb
@@ -111,8 +111,8 @@ module Asyncable
     !!self[self.class.submitted_at_column]
   end
 
-  def asyncable_submitted_at
-    self[self.class.submitted_at_column]
+  def sort_by_submitted_at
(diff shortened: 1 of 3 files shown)


pkarman:
> Should the expired jobs just not include the ones that never got submitted?

The definition of "expired" is a non-null `attempted_at` value more than N days in the past. If a job is attempted but not submitted, then there is a flaw in the code somewhere (e.g. #8802). So no, we should always include all expired jobs, because this page is for fixing things.

pkarman:
> is this something we could address with a followup migration to backfill the timestamps (and remove the fallback)?

I assume the fix for #8802 will do whatever migrations are necessary.

IMO, the "jobs" page should show all jobs regardless of the quality of their data, so that the underlying problems can be addressed.

PR #8867, opened 2019-01-22
remove duplicate claimant names in a claim if they exist

joeyyang:
Resolves (partially) #8771

### Description
Removes duplicate `Appellant Name`, if a claim for some reason has multiple appellants with the same name.


app/models/claim_review.rb
```diff
@@ -128,7 +128,7 @@ def search_table_ui_hash
       veteran_file_number: veteran_file_number,
       veteran_full_name: claim_veteran&.name&.formatted(:readable_full),
       end_products: end_product_establishments,
-      claimant_names: claimants.map(&:name),
+      claimant_names: claimants.map(&:name).uniq,
```
shanear: hmm... wondering why there were two claimaints to begin with here

shanear: this is probably fine. Lets add a comment here about how there shouldn't be two claimants, but this helps for the time being.

Issue #8649, opened 2019-01-11
Additional validation matching payee code to claimant


shanear:
There are ways to default the payee code to the last used payee code for a claimant. This would help Claims Assistants not to make mistakes in adding an incorrect payee code for a claimant.

kencheeto:
We should ask VBMS to figure out what endpoint to hit to find the right default payee code

leikkisa:
Have a current PR that addresses the first approach listed by Ken above. We would also like to add ability to default based on relationship type when there isn't a previous claim with the same claimant.

VBMS is working on getting us the logic for this.  It is also mentioned in this other ticket, so might save that part of the work for that ticket (and will post updates about the logic there):
https://github.com/department-of-veterans-affairs/caseflow/issues/8676

leikkisa:
Currently keeping this in "in progress", will see what we hear from VBMS

leikkisa:
Here's the logic from VBMS: ```switch(benefitType) {
   case $scope.cestEditorData.BENEFIT_TYPE_CD.CPL:
       payeeCode = "00";
       break;
   case $scope.cestEditorData.BENEFIT_TYPE_CD.CPD:
       payeeCode = "10";
       break;
   case $scope.cestEditorData.BENEFIT_TYPE_CD.CH18:
       payeeCode = "11";
       break;
}

SPOUSE("Spouse", Stream.concat(Arrays.asList("10").stream(), getStringRangeStream(70,78)).collect(Collectors.toList())),
CHILD("Child", Stream.concat(getStringRangeStream(11,29), getStringRangeStream(70,78)).collect(Collectors.toList())),
PARENT("Parent", Arrays.asList("50","60").stream().collect(Collectors.toList()));






function isClaimSignifyingAClaimantPayeeAssociation(payeeCd, epCd) {
 //If it's not a veteran payee code AND (either the ep code hasn't been selected yet or it is not a burial or accrued claim)
 // then assume the claim signifies a relationship.
 // Examples: 99 is selected (which is a veteran payee code) and the ep code is not selected, assume not signifying
 //           10 is selected (which is not a veteran payee code) and the ep code is not selected, assume signifying
 //           10 is selected (which is not a veteran payee code) and the ep code of accrued is selected, assume not signifying
   return !claimUtil.isVeteranPayeeCode(payeeCd) && (!epCd||!claimUtil.isBurialOrAccrual(epCd));
}

[...]

PR #8873, opened 2019-01-22
Default payee code

leikkisa:
connects #8649

### DIFF
--- a/app/models/end_product.rb
+++ b/app/models/end_product.rb
@@ -81,9 +81,10 @@ class EndProduct
   DISPATCH_MODIFIERS = %w[070 071 072 073 074 075 076 077 078 079 170 171 175 176 177 178 179 172].freeze
 
   attr_accessor :claim_id, :claim_date, :claim_type_code, :modifier, :status_type_code, :last_action_date,
-                :station_of_jurisdiction, :gulf_war_registry, :suppress_acknowledgement_letter
+                :station_of_jurisdiction, :gulf_war_registry, :suppress_acknowledgement_letter, :payee_code,
+                :claimant_last_name, :claimant_first_name
 
-  attr_writer :payee_code, :claimant_participant_id, :benefit_type_code
+  attr_writer :claimant_participant_id, :benefit_type_code
 
   # Validators are used for validating the EP before we create it in VBMS
   validates :modifier, :claim_type_code, :station_of_jurisdiction, :claim_date, presence: true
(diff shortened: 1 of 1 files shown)


app/models/establish_claim.rb
```diff
@@ -253,7 +253,8 @@ def saved_end_product
     EndProduct.new(
       claim_id: outgoing_reference_id,
       station_of_jurisdiction: appeal.dispatched_to_station,
-      claim_type_code: claim_establishment.ep_code
+      claim_type_code: claim_establishment.ep_code,
+      payee_code: "00"
```
leikkisa: Dispatch uses payee code "00", but was defaulting it in the EndProduct model. Moving it here so that we can use payee_code for the claim's payee code coming back to us as well.

app/models/veteran.rb
```diff
@@ -124,6 +124,13 @@ def relationships
     @relationships ||= fetch_relationships
   end
 
+  def relationships_with_default_payee_codes
+    @relationships_with_default_payee_codes ||= relationships.each do |r|
```
shanear: i think we can make this a lazily loaded function in `Relationship`. This way it's always there if you need it.

spec/feature/intake/review_page_spec.rb
```diff
@@ -48,25 +54,39 @@
...
 
-        context "supplemental claim" do
-          it "requires payee code" do
-            start_supplemental_claim(
-              veteran,
-              veteran_is_not_claimant: veteran_is_not_claimant
-            )
-            check_pension_and_compensation_payee_code
+      context "when benefit type is pension or compensation" do
+        [:higher_level_review, :supplemental_claim].each do |claim_review_type|
```
shanear: I know you are making things DRY here. 

It's counter intuitive, but i think we want to avoid doing too much looping and stuff in tests. It makes them really hard to read and reason about.

client/app/intake/util/index.js
```diff
@@ -49,11 +49,16 @@ export const formatRelationships = (relationships) => {
...
       value: relationship.participant_id,
-      displayText: `${first} ${last}, ${type}`
+      displayText: `${first} ${last}, ${type}`,
+      defaultPayeeCode: relationship.default_payee_code
     };
   });
 };
 
+export const getDefaultPayeeCode = (state, claimant) => {
+  return _.find(state.relationships, { value: claimant }).defaultPayeeCode;
```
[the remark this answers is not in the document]
leikkisa: In this case, the value of claimant on the front-end is just the participant ID.

app/models/veteran.rb
```diff
@@ -124,6 +124,13 @@ def relationships
...
+      eps = end_products.select { |ep| ep.claimant_first_name == r.first_name && ep.claimant_last_name == r.last_name }
+      eps && r.default_payee_code = eps.max_by(&:claim_date).try(:payee_code)
```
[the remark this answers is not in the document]
leikkisa: The try is there for a reason. I think when I added it, it was actually because eps.max_by(&:claim_date) was nil, and nil.try(:something) resolves to nil.

Per Shane's suggestion, I'm going to move stuff to a relationship ui_hash, so that should handle the second comment.

PR #8888, opened 2019-01-23
Add RequestIssuesUpdate.veteran method for asyncable.ui_hash

pkarman:
Revealed in UAT via live-testing error.


app/models/request_issues_update.rb
```diff
@@ -10,6 +10,8 @@ class RequestIssuesUpdate < ApplicationRecord
   attr_writer :request_issues_data
   attr_reader :error_code
 
+  delegate :veteran, to: :review
```
jiahuang: neat! I didn't know this existed

Issue #8630, opened 2019-01-10
Individual Queues Case Details Actions


Chingujo:
As a hearing coordinator, I want to be able to place tasks on hold, complete them, or re-assign them to someone else from my individual queue so I can effectively close out tasks, assign them as needed in case of error or sickness, and keep track of pending tasks. 

### Mocks
https://va-gov.invisionapp.com/share/TUPWZLI7GBP#/screens/340231214 (4 to 12)

### General Page and Layout
- [x] Verify that we are adopting the most up to date Case Details page
- [x] Verify that the task shows up in the Current Tasks Section underneath the Case Details Veteran Snap Shot
- [x] Verify that the task has the following: 
 - Assigned on : Date assigned
 - Days Waiting: Number of days since assigned
- Task history: A timeline of the task 
- Assigned to: Who it is currently assigned to
- Assigned by: Who assigned it
- Task: The name of the task
- Task instructions: When clicked shows the task instructions for that task 
- Actions section 
- [ ] Verify that clicking on action has the following three options: 
- Place on hold - launches the place on hold flow
- Mark as complete - completes the task
- Re-assign - Assigns the case to someone else or release back to the organizational queue 

### Place on hold
- [x] Verify that there is a page titled and header area follows Queue on hold
- [x] Verify that there is a drop down called Select days on hold with the following options: 
- 15 days 
- 30 days
- 45 days
- Other 
  - Verify that clicking other renders another text input where a custom day can be entered
- [x] Verify that there is a Notes field where the user can type additional notes
- [x] Verify that there is a Cancel Button and a Place on Hold button on the bottom right
- [x] Verify the following confirmation when the user clicks place on hold
- Title: You have placed Veteran First Name, Last Name's [Task Name] on hold for [Number of Days]
- Body: You can find this case in the on hold tab of your queue

### Mark as Complete
- [x] Verify that when the user selects "Mark as Complete" a modal pops up with the same Title
- [x] Verify the following copy: 
- Marking this task as complete will remove this admin action from your active tasks
- [x] Verify that cancel closes the modal
- [x] Verify that Mark as Complete moves it to the completed tab of the individual queue
- [x] Verify that there is a modal 
- [x] Verify that there is a confirmation banner with the following copy:
- Title: You have successfully completed Veteran First Name, Last Name's [ task name]
- Body: You can find this case in the completed tab of your queue 

### Re-assign
- [x] Verify that clicking on the "re-assign" action renders a modal 
[...]

Issue #8626, opened 2019-01-10
Organizational Queues


Chingujo:
As a hearing branch management team, I need to have a place where all of the orgs tasks are clearly visible both to assign my cases to myself and others and to keep track of them. 

### Mocks
https://va-gov.invisionapp.com/share/TUPWZLI7GBP (pages 13 to 15)

#### General page and layout
- [x] Verify that the title of the page is Hearing Management Branch Queue
- [x] Verify that there is dropdown button called Switch Views in the upper right of the table
   - Verify that clicking on this renders a dropdown with the following: 
      - Hearing Management Branch Queue
      - Your Queue
- [x] Verify that there is a Queue table with three tabs: 
 - Unassigned: All of the tasks that have yet to be assigned to a hearing team member
_Note: Please ensure that the default state of all unopened tasks are bolded and clicking them unbolds the link_
 - Assigned: All of the tasks currently assigned to an individual 
 - Completed: All tasks that are completed 
- [ ] Verify that there is pagination for the Org queue at the bottom of the screen
  - Verify that we are showing 50 cases per page 

### Unassigned
- [x] Verify the following columns are in the unassigned table in the following order: 
- Case details: A link to the case details in Veteran Last Name Last Name | Appeal ID
- Task: The task type 
- Regional Office: The RO of where the task was originated. This would be null if foreign veteran case
- Type: The same type field and prioritization in Queue showing if the case is CAVC, AOD, or other
- Days Waiting: The number of days waiting since the task was created 
- [ ] Verify that the task column and the regional office is a filterable column and when clicked a searchable filter appears

### Assigned Tab
- [x] Verify that the same information from the Unassigned appears on this page with the following exceptions: 
 - Verify that there is an Assigned to column after the Regional Office 
 - Verify that that the Days Waiting shows the number of days since assigned
- [ ] Verify that the Assigned to column is filterable and clicking on that list shows all of the hearing branch team members in the following format: 
- First Name Last Name (# assigned) 

### Completed Tab
- [x] Verify that the completed tab has all of the same information as the Unassigned tab without the days waiting
- [x] Verify cases appear same as the default as queue 

### Related Issues
* #8618 - EPIC | Queues and Admin Actions 
* #8623 Individual Queues

mdbenjam:
Remaining features already have other issues. Closing this issue so we can better track the remaining aspects:

Filtering: #8215
Pagination: #8242

Issue #8625, opened 2019-01-10
Assignment of Tasks from Case Details


ShellySmithDSVA:
## User Stories
As a Hearings Branch user, I should be able to assign open tasks to myself when I am viewing the case details of a particular case. This will allow me to work specific cases easily.

As a Hearings Branch user, I should be able to assign open tasks to other members of my team in the case details view. This will allow me to to assign or re-route tasks easily and break up works on a case. 

## Mocks
See pages 16-22 of the following InVision file: https://va-gov.invisionapp.com/share/TUPWZLI7GBP#/screens/340231214

## Related Epic
See Epic #8618 for the other associated items with this ticket.

## Acceptance Criteria
- [x] Verify that each open task has an in line drop down that allows users the options to "Assign task to self" or "Assign task to" 
- [x] Verify that when the user selects Assign task to self the user is taken back to the Team queue and is given a confirmation message at the top of the page. 
- [x] Verify this confirmation message includes a link to the user's queue.
- [x] Verify when the user selects Assign task to a modal appears on the case details page.
- [x] Verify the Assign Task modal contains a drop down list of all Hearing team users and a field for notes.
- [x] Verify the user is brought back to the Hearing Team queue and a confirmation message is given at the top of the screen showing what what Vet and task type was assigned to what user. 

## Related Epic 
* #8618 - EPIC | Queues and Admin Actions

mdbenjam:
Everything was completed except for the "Assign to self" action, which we're going to hold off on implementing since the "Assign to" modal defaults to yourself.

Issue #8623, opened 2019-01-10
Individual Queues


Chingujo:
As a hearing coordinator, I need to be able to see all of my tasks so I can keep track of their statuses.

### Mocks
https://va-gov.invisionapp.com/share/TUPWZLI7GBP (pages 1 to 3)

### Table and page layout
- [ ] Verify the banner has an arrow next to Caseflow, that when selected has two options: 
  - Your Queue
   - Hearing Schedule 
- [x] Verify that the title of the page is "Your Queue"
- [ ] Verify that the queue table has three tabs: 
    - Assigned: Contains all of the tasks that are not on hold or just assigned
    - On hold: Contains all of the tasks that the user placed "on hold" and are awaiting further review or action
    - Completed: Contains all of the tasks the user has marked as complete 
- [x] Verify that in the upper right of the table there is a toggle called "Switch Views"
- [x] Verify that when the user clicks this toggle a dropdown appears with the following: 
      - Your Queue
      - Hearing Management Branch 
- [x] Verify that when the user clicks this link, it takes them to the organizational queue 
- [ ] Verify that there is pagination for the individual queue at the bottom of the screen
  - Verify that we are showing 10 cases per page 

### Assigned Tab 
- [x] Verify the following columns in the Assigned tab: 
     - Case details: A link to the case details in the following format Veteran First Name Last Name | Appeal ID
      - Task: The name of the task
      - Regional Office: Displays the regional office and it should be null 
      - Type: The type of the tasks with the same prioritization from Queue
      - Docket Number: Shows in the following format Docket label #
      - Days waiting: Shows the number of days passed since it was assigned 
- [ ] Verify that the task and regional office is filterable

### On Hold Tab
- [x] Verify that all of the same information in Assigned tab shows in the On Hold Tab 
- [ ] Verify that the task and regional office is filterable

### Completed tab
- [x] Verify that this tab has all of the same information from the Assigned tab 
- [ ] Verify that the task and regional office is filterable

### Related Issues
* #8618 - EPIC | Queues and Admin Actions 
* #8626 Organizational Queues

mdbenjam:
Most of this is being built by Queue. I'll work with them to make sure things are on track. The only thing here I don't think they're going to tackle is the regional office column. I'll circle back to that.

mdbenjam:
Remaining features already have other issues. Closing this issue so we can better track the remaining aspects:
[...]

PR #8898, opened 2019-01-23
Adding first pass at hearing admin actions

mdbenjam:
Connects #8623 #8625 #8626 #8630 #8622

### Description
This modifies the admin action UI to accept a generic set of actions, then creates the new HearingAdminAction tasks to pass to the frontend as options.

### Testing Plan
1. Add a new automated test to test the flow from the Hearing Schedule, Assign Hearing page to putting an admin action on hold.


client/app/queue/AddColocatedTaskView.jsx
```diff
@@ -108,9 +119,9 @@ class AddColocatedTaskView extends React.PureComponent<Props, ComponentState> {
         return {
           label: action.actionLabel,
           instructions: action.instructions,
-          type: 'ColocatedTask',
+          type: taskActionData(this.props).type || action.actionLabel,
```
mdbenjam: Hacky way to get this done until we get rid of the action column for colocated admin actions.

client/app/queue/AddColocatedTaskView.jsx
```diff
@@ -60,6 +59,18 @@ type Props = Params & {|
   setAppealAttrs: typeof setAppealAttrs
 |};
 
+const taskActionData = (props) => {
```
mdbenjam: @lowellrex do you want me to put this in a `utils.js` file?

app/models/tasks/generic_task.rb
```diff
@@ -98,20 +98,26 @@ def create_from_params(params, user)
...
+      end
     end
 
-    def child_task_assignee(_parent, params)
-      Object.const_get(params[:assigned_to_type]).find(params[:assigned_to_id])
+    def child_task_assignee(parent, params)
+      if params[:assigned_to_id]
+        Object.const_get(params[:assigned_to_type]).find(params[:assigned_to_id])
+      else
+        parent.assigned_to
```
mdbenjam: @lowellrex is this an okay default?

app/models/tasks/generic_task.rb
```diff
@@ -51,7 +51,7 @@ def update_from_params(params, current_user)
 
     return reassign(params[:reassign], current_user) if params[:reassign]
 
-    update!(status: params[:status]) if params[:status]
+    update!(params)
```
mdbenjam: @lowellrex any concerns with this? This allows us to put generic tasks on hold.

app/models/tasks/hearing_admin_action_task.rb
```diff
@@ -0,0 +1,54 @@
...
+  validates :assigned_by, presence: true
+  validates :parent, presence: true
+  validate :on_hold_duration_is_set, on: :update
+
+  # rubocop:disable Metrics/AbcSize
+  def available_actions(user)
+    if assigned_to == user
+      [
+        Constants.TASK_ACTIONS.PLACE_HOLD.to_h,
+        Constants.TASK_ACTIONS.MARK_COMPLETE.to_h
```
sharonwarner: Ah, missing comma here.

app/controllers/tasks_controller.rb
```diff
@@ -151,7 +152,10 @@ def user
   helper_method :user
 
   def task_class
-    mail_task_classes = Hash[*MailTask.subclasses.map { |subclass| [subclass.to_s.to_sym, subclass] }.flatten]
+    mail_task_classes = Hash[
```
lowellrex: `additional_task_classes` or some other more precise name maybe?

app/models/tasks/generic_task.rb
lowellrex: I can't think of any reason why not. We'll save updates to other fields if they are passed in the parameters, but that shouldn't be an issue.

lowellrex: Actually, can we just remove this method from `GenericTask` and rely on [the same method in `Task`](https://github.com/department-of-veterans-affairs/caseflow/blob/ec0bc4ece3e193e171cd74f821fbe642cf350e15/app/models/task.rb#L103)? We'd only have to add the reassign line in that method (as well as defining that as a no-op function in `Task`) but we would get some more reassurance around the updating of instructions which would be nice.

app/models/tasks/generic_task.rb
lowellrex: Why do we need a default value? We were previously throwing an error if there were no assignee specified in the request parameters, and [we override this function](https://github.com/department-of-veterans-affairs/caseflow/blob/03ad4f62298fda550df130fe1688ab66cdd96582/app/models/tasks/mail_task.rb#L60) when we don't expect to have those parameters. What is the situation where we would want to use this default?

app/models/tasks/hearing_admin_action_task.rb
```diff
@@ -0,0 +1,54 @@
+class HearingAdminActionTask < GenericTask
+  validates :assigned_by, presence: true
+  validates :parent, presence: true
```
lowellrex: We could move this up to the `Task` model and just override it in `RootTask` and `ColocatedTask` (for legacy appeals only) if we'd like.

app/models/tasks/hearing_admin_action_task.rb
```diff
@@ -0,0 +1,54 @@
...
+      ]
+    end
+  end
+  # rubocop:enable Metrics/AbcSize
+
+  private
+
+  def on_hold_duration_is_set
+    if saved_change_to_status? && on_hold? && !on_hold_duration && assigned_to.is_a?(User)
+      errors.add(:on_hold_duration, "has to be specified")
```
lowellrex: We want to add this to the errors attribute instead of raising an error directly because we might be updating multiple `HearingAdminActionTask`s at once?

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -60,13 +62,23 @@ def location_based_on_hearing_type(hearing_type)
...
   end
 
+  def add_admin_action_data
+    {
+      selected: nil,
+      options: HearingAdminActionTask.subclasses.sort_by(&:label).map do
+        |subclass| { value: subclass.name, label: subclass.label }
+      end
+    }
+  end
```
lowellrex: I like it!

client/app/queue/AddColocatedTaskView.jsx
lowellrex: I think it's fine here. Totally up to you!

client/app/queue/AddColocatedTaskView.jsx
```diff
@@ -125,7 +136,8 @@ class AddColocatedTaskView extends React.PureComponent<Props, ComponentState> {
     };
     const msgTitle = COPY.ADD_COLOCATED_TASK_CONFIRMATION_TITLE;
     const msgSubject = pluralize(COPY.ADD_COLOCATED_TASK_CONFIRMATION_SUBJECT, this.state.adminActions.length);
-    const msgActions = this.state.adminActions.map((action) => CO_LOCATED_ADMIN_ACTIONS[action.actionLabel]).join(', ');
+    const msgActions = this.state.adminActions.map((action) =>
```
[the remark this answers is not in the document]
mdbenjam: Should be fixed now.

spec/factories/appeal.rb
```diff
@@ -112,11 +112,5 @@
         )]
       end
     end
-
-    factory :intaked_appeal do
```
mdbenjam: I made this in a previous PR. Turns out it was a duplicate of `:with_tasks`

app/models/tasks/hearing_admin_action_task.rb
mdbenjam: I completely copied this code from `colocated_task.rb`. Maybe this should be moved up to `task.rb`

app/models/tasks/hearing_admin_action_task.rb
mdbenjam: Do you know of an easy way to override validations in subclasses?

client/app/queue/AddColocatedTaskView.jsx
mdbenjam: decided to move it, since it was completely copied from the assign to view file.

app/models/tasks/generic_task.rb
mdbenjam: Gotcha, changing this to override it!

app/models/tasks/hearing_admin_action_task.rb
lowellrex: Maybe we could use [custom validation methods](https://guides.rubyonrails.org/active_record_validations.html#custom-methods)? If it's too much work or would result in the code being too confusing them I'm all for limiting this validation to this task class only.

app/models/tasks/hearing_admin_action_task.rb
lowellrex: I don't think we want to move this up to the `Task` class because it will fail when we put a task on hold when we create a child task for this task... although I would imagine `ColocatedTask`s should be failing right now if we copied this directly... or maybe there have been no child tasks created for `ColocatedTask`s yet. That's entirely possible.

app/models/tasks/hearing_admin_action_task.rb
mdbenjam: I'm going to defer this for now. Though I think going forward it might be a good move. Just not going to add it to this PR.

app/models/tasks/hearing_admin_action_task.rb
mdbenjam: Good point. Yeah I don't think our admin actions have subtasks yet. This probably needs a refactor then in the future. Great reason to make on_hold a subtask instead of a state on a task.

app/models/tasks/hearing_admin_action_task.rb
```diff
@@ -0,0 +1,82 @@
+class HearingAdminActionTask < GenericTask
+  validates :assigned_by, presence: true
+  validates :parent, presence: true
+  validate :on_hold_duration_is_set, on: :update
+
+  def self.child_task_assignee(parent, params)
```
lowellrex: When will `HearingAdminActionTask`s have `assigned_to...` fields? Is this preparing for the future or is there just some code path that I am missing?

lowellrex: Ahhh gotcha! Makes sense to me!

spec/feature/hearing_schedule/schedule_veteran_spec.rb
```diff
@@ -89,4 +89,125 @@
       expect(VACOLS::CaseHearing.first.folder_nr).to eq vacols_case.bfkey
     end
   end
+
+  context "when scheduling an AMA hearing" do
```
sharonwarner: 😂

Issue #7478, opened 2018-10-20
Add Hearing icon to queue/case views


laurjpeterson:
When judges are assigning work to their attorneys, they want to see if a hearing has been held for a case. We think this will be useful to more users than just judges, so will apply this to more than just judge views. 

## AC
- Wherever there is a table (attorneys, judges assign/review, co-located, generic), there is an icon indicating that a hearing has been held on the case
- This indicator only shows up when a hearing has the disposition type of held

Note:
- We only have this for legacy appeals right now.
- We may need to adjust this when AMA appeals start to have hearings. This is because the hearing docket type indicator and hearing held indicator could be redundant. 

## Mocks
@sneha-pai to add by 10/24. This will be an engineering task that someone can pick up when they have downtime.

joeyyang:
Hi @sneha-pai! I'm almost ready to get to this in the next day or two, just checking in on mocks.

lowellrex:
Perhaps we will use the "H" badge from this mock from ticket #6963?
![image](https://user-images.githubusercontent.com/1598889/45507289-a6063180-b75f-11e8-951c-5f5b73afe2ba.png)

laurjpeterson:
i think we can use mocks from this search results tickets for incorporating the `H` badge into queue table views

https://github.com/department-of-veterans-affairs/caseflow/issues/7718

lowellrex:
The hearing badge currently only shows up on the case search page. The remaining work to be done here is to make the badge show up on the various queue pages.

## Steps to reproduce:
* Log in as `BVALSPORER` in local test environment
* Search for Veteran ID `213912991`
* Observe the hearing badge appears
* Return to `/queue` and find the row for "Jadyn N Daugherty (213912991)" in one of the tabs
* Observe that there is no hearing badge

cc @joeyyang adding this to the work we have to do next sprint

lowellrex:
Re-opening this to resolve the bug [temporarily fixed by hiding the hearing badge](https://github.com/department-of-veterans-affairs/caseflow/pull/9271) from all pages.

lowellrex:
🐛 🔫 SMALL? The effort here is in finding the root cause. Reproducing it and making sure it doesn't happen again will likely be quicker than the investigation.

Examples of the error are [in Sentry](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2991/). Perhaps this is a result of the appeal ID not being available at some point during the React lifecycle.

[...]

PR #8904, opened 2019-01-23
Add Hearing icon to queue/case views

joeyyang:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/7478

### Description
Adds the `HearingBadge` "H" to all `<TaskTable>`s. 

ex.
screenshot 2019-01-23 17 31 59
screenshot 2019-01-23 17 32 17

### DIFF
--- a/app/models/serializers/work_queue/task_serializer.rb
+++ b/app/models/serializers/work_queue/task_serializer.rb
@@ -43,6 +43,21 @@ class WorkQueue::TaskSerializer < ActiveModel::Serializer
     object.appeal.try(:docket_number)
   end
 
+  attribute :hearings do
+    object.appeal.hearings.map do |hearing|
+      {
+        held_by: hearing.judge.present? ? hearing.judge.full_name : "",
+        # this assumes only the assigned judge will view the hearing worksheet. otherwise,
+        # we should check `hearing.hearing_views.map(&:user_id).include? judge.css_id`
+        viewed_by_judge: !hearing.hearing_views.empty?,
+        date: hearing.scheduled_for,
+        type: hearing.readable_request_type,
+        external_id: hearing.external_id,
+        disposition: hearing.disposition
+      }
+    end
+  end
+
   attribute :veteran_full_name do
     object.appeal.veteran_full_name
   end
--- a/client/app/queue/components/HearingBadge.jsx
+++ b/client/app/queue/components/HearingBadge.jsx
@@ -46,7 +46,7 @@ const HearingBadge = ({ hearing }) => {
   </div>;
 
   // We expect this badge to be shown in a table, so we use this to get rid of the standard table padding.
-  return <div {...css({ marginRight: '-3rem' })} className="cf-hearing-badge">
+  return <div {...css({ marginRight: '-2.5rem' })} className="cf-hearing-badge">
     <Tooltip id={`badge-${hearing.id}`} text={tooltipText} position="bottom">
       <span {...badgeStyling}>H</span>
     </Tooltip>
--- a/client/app/queue/components/HearingBadge.jsx
+++ b/client/app/queue/components/HearingBadge.jsx
@@ -31,7 +31,7 @@ const listStyling = css({
 });
 
 const HearingBadge = ({ hearing }) => {
-  if (!hearing) {
+  if (!hearing || hearing.disposition !== 'held') {
     return null;
   }
 
--- a/client/app/queue/components/TaskTable.jsx
+++ b/client/app/queue/components/TaskTable.jsx
@@ -85,7 +85,9 @@ export class TaskTableUnconnected extends React.PureComponent<Props> {
   caseHearingColumn = () => {
     return this.props.includeHearingBadge ? {
       header: '',
-      valueFunction: (task: TaskWithAppeal) => <HearingBadge hearing={task.appeal.hearings[0]} />
+      valueFunction: (task: TaskWithAppeal) => {
+        return <HearingBadge hearing={task.hearings && task.hearings[0]} />;
+      }
     } : null;
   }
 
--- a/client/app/queue/AssignedCasesPage.jsx
+++ b/client/app/queue/AssignedCasesPage.jsx
@@ -84,6 +84,7 @@ class AssignedCasesPage extends React.Component<Props> {
         onTaskAssignment={(params) => props.reassignTasksToUser(params)}
         selectedTasks={selectedTasks} />
       <TaskTable
+        includeHearingBadge
         includeSelect
         includeDetailsLink
         includeType
--- a/client/app/queue/AttorneyTaskListView.jsx
+++ b/client/app/queue/AttorneyTaskListView.jsx
@@ -163,6 +163,7 @@ export default (connect(mapStateToProps, mapDispatchToProps)(AttorneyTaskListVie
 const TaskTableTab = ({ description, tasks }) => <React.Fragment>
   <p className="cf-margin-top-0" >{description}</p>
   <TaskTable
+    includeHearingBadge
     includeDetailsLink
     includeType
     includeDocketNumber
--- a/client/app/queue/BeaamAppealListView.jsx
+++ b/client/app/queue/BeaamAppealListView.jsx
@@ -23,6 +23,7 @@ class BeaamListView extends React.PureComponent {
           {messages.error.detail}
         </Alert>}
         <TaskTable
+          includeHearingBadge
           includeDetailsLink
           includeType
           includeDocketNumber
--- a/client/app/queue/ColocatedTaskListView.jsx
+++ b/client/app/queue/ColocatedTaskListView.jsx
@@ -116,6 +116,7 @@ const NewTasksTab = connect(
     return <React.Fragment>
       <p className="cf-margin-top-0">{COPY.COLOCATED_QUEUE_PAGE_NEW_TASKS_DESCRIPTION}</p>
       <TaskTable
+        includeHearingBadge
         includeDetailsLink
         includeTask
         includeType
@@ -133,6 +134,7 @@ const PendingTasksTab = connect(
     return <React.Fragment>
       <p className="cf-margin-top-0">{COPY.COLOCATED_QUEUE_PAGE_PENDING_TASKS_DESCRIPTION}</p>
       <TaskTable
+        includeHearingBadge
         includeDetailsLink
         includeTask
         includeType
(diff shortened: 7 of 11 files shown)


client/app/queue/components/HearingBadge.jsx
```diff
@@ -31,7 +31,7 @@ const listStyling = css({
 });
 
 const HearingBadge = ({ hearing }) => {
-  if (!hearing) {
+  if (!hearing || hearing.disposition !== 'held') {
```
lowellrex: Where does this string come from? Can we make it into a [shared constant](https://github.com/department-of-veterans-affairs/caseflow/tree/master/client/constants) between the front and back-end somehow?

client/app/queue/components/HearingBadge.jsx
```diff
@@ -46,7 +46,7 @@ const HearingBadge = ({ hearing }) => {
   </div>;
 
   // We expect this badge to be shown in a table, so we use this to get rid of the standard table padding.
-  return <div {...css({ marginRight: '-3rem' })} className="cf-hearing-badge">
+  return <div {...css({ marginRight: '-2.5rem' })} className="cf-hearing-badge">
```
lowellrex: Why did we need to make this change?

app/models/serializers/work_queue/task_serializer.rb
```diff
@@ -43,6 +43,21 @@ class WorkQueue::TaskSerializer < ActiveModel::Serializer
...
+        # this assumes only the assigned judge will view the hearing worksheet. otherwise,
+        # we should check `hearing.hearing_views.map(&:user_id).include? judge.css_id`
+        viewed_by_judge: !hearing.hearing_views.empty?,
+        date: hearing.scheduled_for,
+        type: hearing.readable_request_type,
+        external_id: hearing.external_id,
+        disposition: hearing.disposition
+      }
+    end
+  end
```
lowellrex: Let's define a serializer for a hearing and send the set of appeals through that serializer [like we do for issues for LegacyAppeals](https://github.com/department-of-veterans-affairs/caseflow/blob/0d2b1e1abcb5bee627473fa84267abb2d5398f17/app/models/serializers/work_queue/legacy_appeal_serializer.rb#L8). This way we can reuse this exact same serialization for tasks, appeals, and legacy appeals.

client/app/queue/components/TaskTable.jsx
```diff
@@ -85,7 +85,9 @@ export class TaskTableUnconnected extends React.PureComponent<Props> {
   caseHearingColumn = () => {
     return this.props.includeHearingBadge ? {
       header: '',
-      valueFunction: (task: TaskWithAppeal) => <HearingBadge hearing={task.appeal.hearings[0]} />
+      valueFunction: (task: TaskWithAppeal) => {
+        return <HearingBadge hearing={task.hearings && task.hearings[0]} />;
```
lowellrex: Do we need to sort the hearings by date [like we do for the search results list](https://github.com/department-of-veterans-affairs/caseflow/blob/0d2b1e1abcb5bee627473fa84267abb2d5398f17/client/app/queue/CaseListTable.jsx#L87)? Maybe we could put that sorting [in the serializer](https://github.com/department-of-veterans-affairs/caseflow/pull/8904/files#r250428923)?

app/models/serializers/work_queue/task_serializer.rb
lowellrex: Or maybe we could use the `Hearing` model's [existing `to_hash` method](https://github.com/department-of-veterans-affairs/caseflow/blob/0d2b1e1abcb5bee627473fa84267abb2d5398f17/app/models/hearing.rb#L99).

app/models/serializers/work_queue/task_serializer.rb
```diff
@@ -43,6 +43,21 @@ class WorkQueue::TaskSerializer < ActiveModel::Serializer
     object.appeal.try(:docket_number)
   end
 
+  attribute :hearings do
```
lowellrex: Let's lazy load hearings. I'm afraid that if we make requests to load hearings for every task we load for somebody's queue the request could take an incredibly long time. Loading all hearings for AMA appeals is probably pretty cheap (though we would have to join to a few tables), but [loading hearings for legacy appeals](https://github.com/department-of-veterans-affairs/caseflow/blob/0d2b1e1abcb5bee627473fa84267abb2d5398f17/app/models/legacy_appeal.rb#L239) is likely much slower. Let's explore lazy loading hearing information like [we do with document counts](https://github.com/department-of-veterans-affairs/caseflow/blob/0d2b1e1abcb5bee627473fa84267abb2d5398f17/client/app/queue/components/TaskTable.jsx#L264).

client/app/queue/components/HearingBadge.jsx
joeyyang: This spacing works well in all the contexts that the badge shows up in.

client/app/queue/components/HearingBadge.jsx
joeyyang: you know, now that we're filtering these out on the backend, it's probably easier just to get rid of this second condition

joeyyang:
back to you!

client/app/queue/components/HearingBadgeLoader.jsx
```diff
@@ -0,0 +1,47 @@
...
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+
+import ApiUtil from '../../util/ApiUtil';
+
+import HearingBadge from './HearingBadge';
+import { setMostRecentlyHeldHearingForAppeal, errorFetchingHearingForAppeal } from '../QueueActions';
+
+class HearingBadgeLoader extends React.PureComponent {
+  componentDidMount = () => {
```
lowellrex: Let's skip this request if we already have something in the Redux store for this appeal.

lowellrex:
Running this locally it looks like we aren't correctly setting this in the store (maybe we aren't specifying which appeal it should be set for?). Every case shows the following on a fresh seed of `BVASCASPER1`'s assigned tab.

image

client/app/queue/types/state.js
```diff
@@ -82,6 +82,7 @@ export type QueueState = {|
   claimReviews: ClaimReviews,
   editingIssue: Object,
   docCountForAppeal: {[string]: Object},
+  docCountForAppeal: Object,
```
lowellrex: Why do we need to add this?

joeyyang:
Significant changes to this PR — here's a screencap of the new behavior:
jan-31-2019 17-25-36

client/app/queue/components/HearingBadgeLoader.jsx
```diff
@@ -0,0 +1,47 @@
...
+    ApiUtil.get(`/appeals/${this.props.externalId}/hearings`).then((response) => {
+      this.props.setMostRecentlyHeldHearingForAppeal(this.props.externalId, JSON.parse(response.text));
+    }, (error) => {
+      this.props.errorFetchingHearingForAppeal(this.props.externalId, error);
```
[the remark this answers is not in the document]
joeyyang: I decided to have it fail silently — can you think of a better way to handle the error?

joeyyang:
> Running this locally it looks like we aren't correctly setting this in the store (maybe we aren't specifying which appeal it should be set for?). Every case shows the following on a fresh seed of `BVASCASPER1`'s assigned tab.
> 
> image

Fixed this bug.

client/app/queue/CaseListTable.jsx
```diff
@@ -77,14 +77,19 @@ class CaseListTable extends React.PureComponent {
...
 
-    const doAnyAppealsHaveHearings = Boolean(_.find(this.props.appeals, (appeal) => {
-      return appeal.hearings.length;
-    }));
-
-    if (doAnyAppealsHaveHearings) {
+    const doAnyAppealsHaveHeldHearings = Boolean(
+      _.find(this.props.appeals, (appeal) => {
+        return appeal.hearings.
+          filter((hearing) => hearing.disposition === 'held').
```
joeyyang: still need to check `disposition` here because we don't get this result thru the `appeals_controller`

client/app/queue/components/HearingBadge.jsx
```diff
@@ -1,11 +1,23 @@
...
 
+import ApiUtil from '../../util/ApiUtil';
 import { DateString } from '../../util/DateUtil';
+import { setMostRecentlyHeldHearingForAppeal, errorFetchingHearingForAppeal } from '../QueueActions';
+
+/**
+ * This component can accept either a Hearing object or a Task object.
+ * e.g.,
+ *   <HearingBadge hearing={hearing} />
+ *   <HearingBadge task={task} />
```
joeyyang: let me know what you think of this API

client/app/queue/components/HearingBadge.jsx
```diff
@@ -30,27 +42,64 @@ const listStyling = css({
...
+HearingBadge.propTypes = {
+  task: PropTypes.object,
+  hearing: PropTypes.object
 };
 
-export default HearingBadge;
+const mapStateToProps = (state, ownProps) => {
+  let externalId, hearing;
+
+  if (ownProps.hearing) {
```
joeyyang: accept either a hearing or a task

client/app/queue/components/HearingBadge.jsx
lowellrex: I like it!

lowellrex:
Looks good! Can we add a few tests to make sure the hearing icon shows up as we expect (and doesn't) on the case search page and on a queue table view?

joeyyang:
> Looks good! Can we add a few tests to make sure the hearing icon shows up as we expect (and doesn't) on the case search page and on a queue table view?

Case search hearing badge tests can be found here in [search_spec.rb](https://github.com/department-of-veterans-affairs/caseflow/blame/master/spec/feature/queue/search_spec.rb#L183), and I'll add tests for the queue table view now!

spec/feature/queue/case_details_spec.rb
```diff
@@ -79,7 +79,7 @@
 
       scenario "Entire set of attributes for hearing are displayed" do
         visit "/queue"
-        page.find(:xpath, "//tr[@id='table-row-#{appeal.vacols_id}']/td[1]/a").click
+        page.find(:xpath, "//tr[@id='table-row-#{appeal.vacols_id}']/td[2]/a").click
```
lowellrex: What do you think about spinning this pattern out to its own helper function? I imagine something like [the `click_dropdown` function](https://github.com/department-of-veterans-affairs/caseflow/blob/f77595d4f1a3e128a11da98134195c0614882f51/spec/support/feature_helper.rb#L2) perhaps with a signature like `find_table_cell(id: appeal.vacols_id, row_header: COPY::CASE_LIST_TABLE_VETERAN_NAME_COLUMN_TITLE)`.

PR #8906, opened 2019-01-23
Hearing Prep Header Updates

### LINKED ISSUE #8742: Daily Docket | Header updates
The daily docket header in hearing prep needs some added information: the notes on the hearing day, the bva coordinator, and the room number. 

*This information only needs to be displayed for Caseflow hearing days.*

screen shot 2019-01-15 at 4 45 14 pm

oscarAtNava:
Resolves #8742 

### Description
Added hearing day attributes to the hearing prep header area.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to hearing prep
2. Confirm coordinator, room and notes attributes are present for the hearing day


app/controllers/hearings/dockets_controller.rb
```diff
@@ -13,12 +13,20 @@ def show
     date = date_from_string(params[:docket_date])
     return not_found unless date && judge.docket?(date)
 
+    hearing_day = HearingDay.find_by(scheduled_for: date)
```
sharonwarner: This will have to include type and regional office as well. We'll have lots of hearing days for a particular date.

sharonwarner: Actually, all of the hearings should be associated with a hearing day, that would be an easier way to find it.

app/controllers/hearings/dockets_controller.rb
```diff
@@ -13,18 +13,44 @@ def show
...
+        }
       end
     end
   end
 
   private
 
+  def hearing_day(first_hearing)
+    hearing_day_object = {}
+    if first_hearing["docket_name"] == "hearing"
```
sharonwarner: This functionality should work for legacy hearings as well, as long as there's a Caseflow hearing day. We can use the April 1st date to determine if there's a Caseflow hearing day.

oscarAtNava: Ok, the ticket specifically mentioned to only do the for Caseflow Hearing Days.

sharonwarner: Yes, that's correct - Caseflow hearing days, but Caseflow and legacy hearings.

app/repositories/hearing_repository.rb
```diff
@@ -266,6 +266,7 @@ def vacols_attributes(vacols_record)
         regional_office_key: ro,
         request_type: vacols_record.hearing_type,
         scheduled_for: date,
+        vdkey: vacols_record.vdkey,
```
sharonwarner: Could we update this to parent_record_id? That way we won't need the additional logic in hearing_day in the dockets controller.

oscarAtNava: Changing the name to something other than vdkey breaks the logic that loads the data from VACOLS. (associated_vacols_model.rb)

Issue #8307, opened 2018-12-17
Geo matching | Schedule Veterans | Case details modal error


Chingujo:
As a hearing coordinator, I need validations in the modal to ensure that I am not making any errors when I schedule. 

### AC
- [ ] Verify that if the user tries to schedule without selecting a hearing location, date, or time we render an in line error. 
- [ ] The format of the error should follow: 
  - Title
   - Resolution:

### Mocks
![schedule veteran error](https://user-images.githubusercontent.com/23316107/50095156-d0461000-01e2-11e9-9373-0ff912dba2a8.png)
 

### Related Issues
#8041 Case details schedule modal

Issue #8043, opened 2018-11-29
Geo matching | Daily Docket | Change hearing location


Chingujo:
As a hearing coordinator, I need to be able to change the regional office and the related hearing locations to account for postponments and other veteran requests

### Acceptance Criteria
- [ ] Verify that the user can change the hearing location via dropdown and it shows
  - City, state miles
- [ ] Verify that if the user changes the Regional Office, it also changes the hearing location options and forces the user to select a date
- [ ] Verify that a save button and cancel link appears when a user makes a change
- [ ] Verify that clicking save confirms the change with a success message at the top of the page

### Related issue
#8042  Add hearing location fields

Issue #8042, opened 2018-11-29
Geo matching | Daily Docket | Add Hearing Location fields


Chingujo:
As a hearing coordinator, I need to see the hearing location in the Daily Docket both in the column in the right and the actions column in the left to edit. 

### Acceptance Criteria
- [ ] Verify that in the RO/hearing time column that it also displays the hearing location 
- [ ] Verify that there is a hearing location dropdown in the actions column 
- [ ] Verify the user can change the hearing location for that Regional Office

Chingujo:
@Chingujo 

 - Issue with printing (is vertical print alignment okay)

 - Bold notes/make general 

- modal?

Issue #8041, opened 2018-11-29
Geo matching | Schedule Veterans | Case Details Modal


Chingujo:
As a hearing coordinator, I need a way to schedule veterans by alternate hearing location and ensure that the Veteran is going to their closest hearing location. 

### Acceptance Criteria
- [ ] Verify that when the user clicks "Schedule Veterans" from case details the schedule modal comes up
- [ ] Verify the following in the modal:
   - Title: Schedule Veteran Name
   - Veteran address underneath the title 
   - Regional Office: 
      Regional Office 
   - Suggested Hearing Location
      - Dropdown shows: Suggested hearing location (facility name) distance in miles
   - Hearing date: 
      Hearing day 
   - Hearing Time: 
      8:00 am and 12:30 p.m as radio buttons 
- [ ] Verify that there is a change link next to Regional Office and hearing day 
- [ ] Verify that when this change link is selected, a drop down appears
- [ ] Verify that changing the Regional Office, changes the suggested location and the available hearing day 

### Mocks

Chingujo:
Remove the other time selector to be addressed in #8653

PR #8909, opened 2019-01-23
AHL on Assign Hearings, Schedule Veterans Modal, and Daily Docket

### LINKED ISSUE #8035: Geo matching | Scheduled Veterans Tab | Add Hearing Location Column
As a hearing coordinator, I want to be able to see the hearing location on the "Scheduled Veterans" column. So, I can build the desired schedule for that day. 

### Acceptance Criteria
- [ ] Verify that there is a "Hearing Location" column in the "Scheduled Veterans" tab
- [ ] Verify that there the hearing locations are displayed in the following format: 
- City, State
- (Facility) 
- Verify that the following logic exists for facility type 
    - If regional office, show (RO)
    - If related to a clinic or health facility show (VHA)
    - If vet center, show (Vet Center)
    - If related to benefits other than RO show (VBA) 
- [ ] Verify that time is the RO and the column name is RO time

### Mocks
**Scheduled Page**
geo matching current schedule

### Related Issues
#8034  Geo matching epic

lomaxap:
Resolves  #8035, #8037, #8036, #8042, #8043, #8041, and #8307

- adds hearing_location table to legacy and ama hearings
- adds suggested location columns to assign hearings page
- Refactors assign hearings tab

- Adds regional office and hearing location dropdowns
- Adds VeteranHearingLocationDropdown component
- queries VADotGovService if no veteran AHL info exists for an RO
- Minor refactor of daily docket columns

- Adds hearing location to Schedule Veteran modal
- creates LegacyHearing when hearing is scheduled
- slight refactor of slot_new_hearing
- uses slot_new_hearing in ScheduleVeteranTask


app/models/legacy_hearing.rb
```diff
@@ -38,7 +40,7 @@ def external_id
     vacols_id
   end
 
-  def location
+  def vacols_location
```
sharonwarner: I think this makes more sense as something like `request_type_location`. It's not really the VACOLS location because this location works for Caseflow hearings as well.

app/models/serializers/work_queue/task_serializer.rb
```diff
@@ -63,6 +63,14 @@ class WorkQueue::TaskSerializer < ActiveModel::Serializer
     object.appeal.number_of_issues
   end
 
+  attribute :closest_regional_office do
+    object.appeal.veteran_closest_regional_office
+  end
+
+  attribute :suggested_hearing_location do
```
sharonwarner: Should we send all available_hearing_locations to the frontend and then pick the suggested location there? I believe we need all the options for the dropdowns.

app/services/appeal_repository.rb
```diff
@@ -307,6 +307,7 @@ def self.appeals_ready_for_hearing_schedule(regional_office)
       return appeals_ready_for_co_hearing_schedule
     end
 
+    # do we need to not limit this to 30 for assign hearings page?
```
sharonwarner: Mark made some updates that improved the page load time for this. I don't think we need the limit anymore. Either way the limit is definitely going away when we add pagination.

app/models/serializers/work_queue/task_serializer.rb
lomaxap: true

PR #8911, opened 2019-01-24
Change Service Endpoints

lomaxap:
- removes `services` from endpoints
- pass `ids` param as comma-separated string
- check if missing_veterans exist


app/services/external_api/va_dot_gov_service.rb
```diff
@@ -72,11 +72,11 @@ def base_url
     end
 
     def facilities_endpoint
-      "services/va_facilities/v0/facilities"
+      "va_facilities/v0/facilities"
```
mdbenjam: This needs to use the ENV var. It should be available now.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -68,6 +68,17 @@ def perform
       rescue Caseflow::Error::VaDotGovLimitError
         sleep 60
         va_dot_gov_address = validate_veteran_address(veteran)
+      rescue Caseflow::Error::VaDotGovRequestError => e
+        case e.message["messages"][0]["key"]
+        when "AddressCouldNotBeFound"
+          # do something when address is not valid (1234 Fake st.)
```
mdbenjam: At the very least we should sent the exception to sentry: `Raven.capture_exception(e)`, but I think failing the job is fine. It'll force us to work through the various errors.

app/services/external_api/va_dot_gov_service.rb
lomaxap: i don't see the endpoints there, just the base url

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -19,8 +19,10 @@ def file_numbers
   end
 
   def missing_veteran_file_numbers
-    existing_veteran_file_numbers = Veteran.where(file_number: file_numbers).pluck(:file_number)
-    (file_numbers - existing_veteran_file_numbers)[0, (QUERY_LIMIT - existing_veteran_file_numbers.length)]
+    @missing_veteran_file_numbers ||= begin
```
mdbenjam: what's the begin for?

app/services/external_api/va_dot_gov_service.rb
mdbenjam: 🤦‍♂️ my bad, thought this was the base url stuff. Totally right. Didn't see the function above.

PR #8917, opened 2019-01-24
Differentiate between VACOLS calls for metrics

lowellrex:
Requests to VACOLS for appeals and hearings data are currently clumped into one bucket in our metrics. This PR separates them so we can determine where these requests are truly coming from.

image


lowellrex:
Link to related slack discussion: https://dsva.slack.com/archives/C4JECDLSE/p1548348333345200

PR #8929, opened 2019-01-24
Allow editing document ID for legacy cases

### LINKED ISSUE #8640: A user should be able to edit document ID for legacy appeals
AC:
1. Display `edit` icon for users who have permissions to edit a document ID. Can use similar pattern as here: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/serializers/work_queue/appeal_serializer.rb#L35
2. When editing a document ID for lagacy cases, update attorney_case_reviews table and VACOLS::Decass table with the correct document ID. To find associated VACOLS::Decass table use `task_id` which is a combination of defolder_id and created_at date.

monfresh:
**Why**: This is a follow-up to the feature where we added the ability
to edit document IDs for AMA cases. Now, we are supporting legacy cases
as well.

Resolves #8640

**Acceptance Criteria**:
- A judge or attorney assigned to a legacy case can edit the Document ID
- A user that is not associated with a case (whether AMA or legacy)
should not see the "Edit" link
- When the document ID for a legacy case is updated, both the
`AttorneyCaseReviews` and `VACOLS::Decass` tables should be updated.
- The format of Document IDs is validated based on the work product

**Test Plan**:
1. Sign in as an attorney and complete the checkout flow for a legacy
case
2. Visit the appeals details page
Expected Result: you should be able edit the document ID
3. Enter an invalid ID, such as "123", and click Save
Expected Result: an error message appears
4. Enter a valid ID and click Save
Expected Result: the new document ID appears on the page
5. Reload the page
Expected Result: the new document ID appears on the page
6. Switch to the judge user assigned to the case
7. Repeat steps 2-5
8. Switch to a user not associated with the case
Expected Result: you should not be able to edit the document ID

### DIFF
--- a/app/controllers/case_reviews_controller.rb
+++ b/app/controllers/case_reviews_controller.rb
@@ -26,9 +26,10 @@ class CaseReviewsController < ApplicationController
   end
 
   def update
-    result = UpdateAttorneyCaseReview.new(
+    case_review_class = (params[:legacy] == true) ? UpdateLegacyAttorneyCaseReview : UpdateAttorneyCaseReview
+    result = case_review_class.new(
       id: params[:id],
-      user_id: current_user.id,
+      user: current_user,
       document_id: params[:document_id]
     ).call
 
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -4,7 +4,7 @@ class Organization < ApplicationRecord
   has_many :users, through: :organizations_users
 
   def admins
-    organizations_users.select(&:admin?).map(&:user)
+    organizations_users.includes(:user).select(&:admin?).map(&:user)
   end
 
   def non_admins
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -69,6 +69,8 @@ Rails/Present:
 
 Metrics/MethodLength:
   Max: 20
+  Exclude:
+    - "spec/**/*"
 
 Metrics/ClassLength:
   Max: 400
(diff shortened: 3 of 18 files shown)


app/models/organization.rb
```diff
@@ -4,7 +4,7 @@ class Organization < ApplicationRecord
   has_many :users, through: :organizations_users
 
   def admins
-    organizations_users.select(&:admin?).map(&:user)
+    organizations_users.includes(:user).select(&:admin?).map(&:user)
```
monfresh: I did this to stop the N+1 queries while testing locally. It's not related to this PR, so I'm happy to move it to a separate PR if that's preferable.

aroltsch: Thank you!

app/controllers/case_reviews_controller.rb
```diff
@@ -26,9 +26,10 @@ def complete
   end
 
   def update
-    result = UpdateAttorneyCaseReview.new(
+    case_review_class = (params[:legacy] == true) ? UpdateLegacyAttorneyCaseReview : UpdateAttorneyCaseReview
```
aroltsch: Not a huge fan passing this parameter like that but I guess it would hard to determine that based on the `id`

aroltsch:
We should be working on deprecating Decass records very soon so hopefully we can work on improving the workflow around updating attorney_case_reviews and storing document IDs

app/controllers/case_reviews_controller.rb
monfresh: Alternatively, we could pass in the appeal Id and have the controller determine whether or not it's legacy.

PR #8930, opened 2019-01-24
Adding Appeal skeleton model for appeal status api

mc-usds:
Resolves #{8728}

### Description
These are updates for Appeals to be supported in Appeal Status API.
Added a new serializer for appeal and skeleton methods to Appeal to support it.
Also updated the controller -- renamed appeals to legacy_appeals and "appeals" will refer to ama appeals and added the Appeal json to be returned along with hlr and sc.

Added a simple test. I will fill it in with as the attributes are implemented.

### Acceptance Criteria
- [x] Code compiles correctly

### Testing Plan
1. ran lint
2. tested on the command line

appeals = Appeal.where(veteran_file_number: 701305078)

        appeals,
        each_serializer: ::V2::AppealStatusSerializer,
        key_transform: :camel_lower
      ).as_json

Output: 
 {:data=>
  [{:id=>"A1",
    :type=>"appeal",
    :attributes=>
     {:appealIds=>["A1"],
      :updated=>"2019-01-24T15:52:12-05:00",
      :incompleteHistory=>false,
      :type=>"original",
      :active=>false,
      :description=>nil,
      :aod=>nil,
      :location=>nil,
      :aoj=>"other",
      :programArea=>nil,
      :status=>nil,
      :alerts=>nil,
      :docket=>nil,
      :issues=>[],
      :events=>nil,
      :evidence=>[]}}]}

### DIFF
--- a/app/controllers/api/v2/appeals_controller.rb
+++ b/app/controllers/api/v2/appeals_controller.rb
@@ -26,7 +26,7 @@ class Api::V2::AppealsController < Api::ApplicationController
         all_reviews_and_appeals
       else
         ActiveModelSerializers::SerializableResource.new(
-          appeals,
+          legacy_appeals,
           each_serializer: ::V2::AppealSerializer,
           key_transform: :camel_lower
         ).as_json
@@ -34,9 +34,9 @@ class Api::V2::AppealsController < Api::ApplicationController
     end
   end
 
-  def appeals
+  def legacy_appeals
     # Appeals API is currently limited to VBA appeals
-    @appeals ||= AppealHistory.for_api(vbms_id: vbms_id).select do |series|
+    @legacy_appeals ||= AppealHistory.for_api(vbms_id: vbms_id).select do |series|
       series.aoj == :vba
     end
   end
@@ -47,6 +47,10 @@ class Api::V2::AppealsController < Api::ApplicationController
 
   def supplemental_claims
     @supplemental_claims ||= SupplementalClaim.where(veteran_file_number: vbms_id.sub("S", ""))
+  end
+
+  def appeals
+    @appeals ||= Appeal.where(veteran_file_number: vbms_id.sub("S", ""))
   end
 
   def all_reviews_and_appeals
@@ -62,7 +66,13 @@ class Api::V2::AppealsController < Api::ApplicationController
       key_transform: :camel_lower
     ).as_json
 
-    { data: hlr_json[:data] + sc_json[:data] }
+    appeal_json = ActiveModelSerializers::SerializableResource.new(
+      appeals,
+      each_serializer: ::V2::AppealStatusSerializer,
+      key_transform: :camel_lower
+    ).as_json
+
+    { data: hlr_json[:data] + sc_json[:data] + appeal_json[:data] }
   end
 
   def vbms_id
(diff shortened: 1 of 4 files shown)


app/controllers/api/v2/appeals_controller.rb
```diff
@@ -49,6 +49,10 @@ def supplemental_claims
     @supplemental_claims ||= SupplementalClaim.where(veteran_file_number: vbms_id.sub("S", ""))
   end
 
+  def appeals
+    @appeals ||= Appeal.where(veteran_file_number: vbms_id.sub("S", ""))
```
cmgiven: In addition to ending with an "S", a file number can also end with a "C".

cmgiven: This is an older numbering system that was in place before VA switched to SSNs.

We might also want to check whether C numbers are zero-padded to eight digits. They are not in VACOLS, but they are in VBMS. I don't know what Appeal is using.

app/controllers/api/v2/appeals_controller.rb
```diff
@@ -62,7 +66,13 @@ def all_reviews_and_appeals
...
     ).as_json
 
-    { data: hlr_json[:data] + sc_json[:data] }
+    appeal_json = ActiveModelSerializers::SerializableResource.new(
+      appeals,
+      each_serializer: ::V2::AppealStatusSerializer,
+      key_transform: :camel_lower
+    ).as_json
+
+    { data: hlr_json[:data] + sc_json[:data] + appeal_json[:data] }
```
cmgiven: Nice.

amprokop:
High-level comment: should we be making this into v3 of appeal status? Seems like the changes might be pretty extensive. I'm not adamant either way, just pointing it out.

app/controllers/api/v2/appeals_controller.rb
amprokop: Hmm to add on this — for AMA appeals, I'm not sure we have to bother with this S and C business.

For legacy appeals, we do, because that's the format used in VACOLS. The existing code path for legacy appeals runs like:

1. Consumer (vets.gov) sends us an SSN
2. We fetch the file number using the SSN from BGS
3. We convert it to the VACOLS format by adding an S or a C
4. We query VACOLS using the VACOLS format

For hlrs and supplemental claims and ama appeals, it looks like you're doing:
1. Consumer (vets.gov) sends us an SSN
2. We fetch the file number using the SSN from BGS
3. We convert it to the VACOLS format by adding an S or a C
4. We convert it back to the file number by stripping the S (as @cmgiven points out i think this isn't quite right, we'd need to skip the C too)
5. We query our own DB

If my understanding is correct we should edit by skipping 3 and 4

cmgiven:
My view of this is that the schema remains the same, we're mostly just changing the enum possibilities, which I think should be considered non-breaking changes, thus not a major version. But I don't have strong feelings there.

app/controllers/api/v2/appeals_controller.rb
mc-usds: Fixed this to get the veteran file number from BGS and use that for hlr/sc/ama and vbms_id is the converted version used by legacy appeals.

Issue #8933, opened 2019-01-24
Intake | Payee code is null for claim review intakes


shanear:
This appears to be blocking all EP Establishments in prod.

Here's an example in UAT:
```
irb(main):005:0> HigherLevelReview.find_by(veteran_file_number: "700052108").end_product_establishments
[2019-01-24 17:36:00 -0500]   HigherLevelReview Load (0.8ms)  SELECT  "higher_level_reviews".* FROM "higher_level_reviews" WHERE "higher_level_reviews"."veteran_file_number" = $1 LIMIT $2  [["veteran_file_number", "700052108"], ["LIMIT", 1]]
[2019-01-24 17:36:00 -0500]   EndProductEstablishment Load (0.7ms)  SELECT  "end_product_establishments".* FROM "end_product_establishments" WHERE "end_product_establishments"."source_id" = $1 AND "end_product_establishments"."source_type" = $2 LIMIT $3  [["source_id", 978], ["source_type", "HigherLevelReview"], ["LIMIT", 11]]
=> #<ActiveRecord::Associations::CollectionProxy [#<EndProductEstablishment id: 727, established_at: nil, synced_status: nil, source_type: "HigherLevelReview", source_id: 978, veteran_file_number: "700052108", reference_id: nil, claim_date: "2019-01-22", code: "030HLRNR", modifier: nil, station: "499", last_synced_at: nil, claimant_participant_id: "600264086", payee_code: nil, committed_at: nil, doc_reference_id: nil, development_item_reference_id: nil, benefit_type_code: "1", user_id: 412>, #<EndProductEstablishment id: 726, established_at: nil, synced_status: nil, source_type: "HigherLevelReview", source_id: 978, veteran_file_number: "700052108", reference_id: nil, claim_date: "2019-01-22", code: "030HLRR", modifier: nil, station: "499", last_synced_at: nil, claimant_participant_id: "600264086", payee_code: nil, committed_at: nil, doc_reference_id: nil, development_item_reference_id: nil, benefit_type_code: "1", user_id: 412>]>
```

shanear:
validated in UAT. Works.

PR #8935, opened 2019-01-24
Fix default payee code

pkarman:
connects #8933 

*Description*

The `end_product_establishments.payee_code` column was getting null values. This PR adds a db constraint to prevent that.

The dilemma is that we are using `DecisionReview.payee_code` to set the value of new `EndProductEstablishment` records, but we are also using it to populate the client UI form. We want the UI form to default to `nil` so that the user is prompted to select a value, but that means that `nil` values may somehow slip into the db.

What we do not have is a reproducible test for how nil values get into the db. The hope is that by adding the db constraint we can discover the code path that is leading to that state.

### DIFF
--- a/app/models/end_product_establishment.rb
+++ b/app/models/end_product_establishment.rb
@@ -148,7 +148,11 @@ class EndProductEstablishment < ApplicationRecord
     end
   end
 
-  validates :payee_code, presence: true
+  before_save :set_default_values
+
+  def set_default_values
+    self.payee_code ||= EndProduct::DEFAULT_PAYEE_CODE
+  end
 
   def perform!(commit: false)
     return if reference_id
--- a/app/models/end_product_establishment.rb
+++ b/app/models/end_product_establishment.rb
@@ -151,6 +151,13 @@ class EndProductEstablishment < ApplicationRecord
   before_save :set_default_values
 
   def set_default_values
+    if payee_code.nil?
+      if source && source.try(:veteran_is_not_claimant) == true
+        fail "nil payee_code should be 02"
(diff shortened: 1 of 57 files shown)


app/models/end_product_establishment.rb
```diff
@@ -148,6 +148,8 @@ def active
     end
   end
 
+  validates :payee_code, presence: true
```
shanear: I don't think this is necessary if we're setting it by default in the code

spec/models/relationship_spec.rb
```diff
@@ -15,7 +15,7 @@
     subject { relationship.ui_hash }
 
     context "when there are no prior claims for that relationship" do
-      it "returns a hash with nil default_payee_code" do
+      it "returns a hash with a nil default_payee_code" do
```
kencheeto: this is because it's nil until we try to save the end product, right?

pkarman: yes we keep it nil so that the user is prompted to select it. otherwise, if we default it the user won't be prompted and the default might be incorrect.

spec/support/intake_helpers.rb
```diff
@@ -30,10 +30,11 @@ def start_higher_level_review(
     )
 
     if claim_participant_id
-      Claimant.create!(
+      create(
+        :claimant,
         review_request: higher_level_review,
-        participant_id: claim_participant_id || test_veteran.participant_id,
```
pkarman: since we already know `claim_participant_id` is truthy, the `||` constructor will never be called.

Issue #8652, opened 2019-01-11
Display task column in judge queue table view


lowellrex:
The only things that appeared in judges' queue table views prior to the quality review pilot were decisions to be signed. Now that judges can get quality review tasks we should show the task label in the queue table view to indicate what action they need to take on the appeal at a glance.

## Acceptance criteria
* [ ] Show task label in the judge queue view
* [ ] [Title case the task label](https://github.com/department-of-veterans-affairs/caseflow/issues/8652#issuecomment-453603556)

Before | After
--- | ---
![image](https://user-images.githubusercontent.com/32683958/51050422-c1af0680-159e-11e9-87e8-700cc4857c2b.png) | ![image](https://user-images.githubusercontent.com/32683958/51050723-b7413c80-159f-11e9-931c-2c9fa475309e.png)

lowellrex:
cc @allyceh @eric-nava for thoughts on the visual appearance of this new tab (we don't have the ability to change the sort order yet, as a heads up). Maybe we should be upper-casing these strings though?

allyceh:
Visual appearance looks good to me. Can we use title case for these strings since "type" is also using title case? Ex: "Review", and "Quality Review"

PR #8944, opened 2019-01-25
add task column to judge queue view

joeyyang:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/8652
/cc @allyceh 

### Description
Adds the task column to judge queue view (with the task name in `_.startCase`)

screenshot 2019-01-25 10 16 18


client/COPY.json
```diff
@@ -22,6 +22,7 @@
   "CASE_LIST_TABLE_APPEAL_ISSUE_COUNT_COLUMN_TITLE": "Issues",
   "CASE_LIST_TABLE_APPEAL_DOCUMENT_COUNT_COLUMN_TITLE": "Veteran Documents",
   "CASE_LIST_TABLE_TASK_ACTION_COLUMN_TITLE": "Action",
+  "CASE_LIST_TABLE_TASK_LABEL_COLUMN_TITLE": "Task(s)",
```
lowellrex: Can we re-use the existing `CASE_LIST_TABLE_TASKS_COLUMN_TITLE` copy here instead?

client/app/queue/components/TaskTable.jsx
```diff
@@ -131,6 +132,13 @@ export class TaskTableUnconnected extends React.PureComponent<Props> {
     } : null;
   }
 
+  caseTaskLabelColumn = () => {
+    return this.props.includeTaskLabel ? {
+      header: COPY.CASE_LIST_TABLE_TASK_LABEL_COLUMN_TITLE,
+      valueFunction: (task: TaskWithAppeal) => _.startCase(task.label)
+    } : null;
+  }
+
```
lowellrex: This seems very similar to the existing [`caseTaskColumn` task table function](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/components/TaskTable.jsx#L126). Can we re-use that? Perhaps by title-casing the label in [in the `actionNameOfTask` function](https://github.com/department-of-veterans-affairs/caseflow/blob/36e9250177ddce645b96fa59fe919128f1449e6d/client/app/queue/components/TaskTable.jsx#L124) if the task label does not exist in the set of `CO_LOCATED_ADMIN_ACTIONS` we map the label into?

client/COPY.json
joeyyang: done!

client/app/queue/components/TaskTable.jsx
joeyyang: done!

Issue #8241, opened 2018-12-12
Add user column to team queue table


laurjpeterson:
We want to add the user column to team queues so supervisors and team members know who is working on tasks. 

This is especially important if teams decide to have users self-assign work (like the Hearings Management Branch has suggested), or reassign work when users are out on leave. 

## AC
- User column exists on team queue, `Assigned` and `Completed` tab

#### User column on "Assigned" tab
![colocated-table-team-cases-768](https://user-images.githubusercontent.com/31224257/51272127-4337d780-1997-11e9-87ef-f900cd68ae14.png)

#### Considerations and implications:
- to add this column (and especially to suit our narrowest browser width specifications), I removed "Type" column (Original, CAVC, Reconsideration, etc). [now outdated]
   [UPDATE]:  I've reintroduced type column and omitted veteran documents column.

## Open questions [all resolved as of 01/16/19]
- is it easier to add user column to all tabs, even though the `Unassigned` tab doesn't have assigned users? 
          - Sneha: Adding this column requires sacrificing another important column that may be useful when viewing team cases. I think we should limit adding this column only to assigned and completed tabs.
- Is it okay to remove "type" column? If not, which other column would we want to remove in order to add the "assigned to" column?

sneha-pai:
@laurjpeterson can you please have a look at the proposed mock, and questions around if we include this column, which other column do we not show?

laurjpeterson:
thank you @sneha-pai! 

>is it easier to add user column to all tabs, even though the Unassigned tab doesn't have assigned users?
>- Sneha: Adding this column requires sacrificing another important column that may be useful when viewing team cases. I think we should limit adding this column only to assigned and completed tabs.
- Ok

>Is it okay to remove "type" column? If not, which other column would we want to remove in order to add the "assigned to" column?
- I think I'd prefer to remove the link to Reader, since users can navigate to Reader from Case Details
- Type is important for AOD, CAVC, etc. for legacy cases. 
- I'd love @MeredithStewart @Chingujo @laurenrussell 's eyes on this too, for their thoughts on this w/r/t hearing branch

laurenrussell:
@laurjpeterson I would agree, in that Type is important to stay on track when filling the docket.

laurjpeterson:
Note: VLJ support staff would benefit from the user column on the team queue soon as well, so they can easily know when they need to reassign cases from user to user. cc: @sneha-pai @mdbenjam

PR #8947, opened 2019-01-25
add 'assigned to' column to organization queue view (assigned and complete tabs)

joeyyang:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/8241
/cc @laurjpeterson 

### Description
Adds an 'assigned to' column to organization queue view (assigned and complete tabs).

screenshot 2019-01-28 16 32 36


client/COPY.json
```diff
@@ -13,6 +13,7 @@
   "CASE_LIST_TABLE_TITLE": "Appeals",
   "CASE_LIST_TABLE_VETERAN_NAME_COLUMN_TITLE": "Case Details",
   "CASE_LIST_TABLE_DOCKET_NUMBER_COLUMN_TITLE": "Docket",
+  "CASE_LIST_TABLE_ASSIGNED_TO_COLUMN_TITLE": "Assigned to",
```
lowellrex: Can we re-use the existing `CASE_LIST_TABLE_APPEAL_LOCATION_COLUMN_TITLE` copy here? I think they would be used in all the same places.

lowellrex:
Noticed a bug (#8979) while reviewing this PR. Even though that bug doesn't strictly block this PR, let's fix it because I think it will make the result of this PR easier to discern.

app/models/serializers/work_queue/task_serializer.rb
```diff
@@ -27,6 +27,7 @@ class WorkQueue::TaskSerializer < ActiveModel::Serializer
   attribute :assigned_to do
     {
       css_id: object.assigned_to.try(:css_id),
+      name: object.assigned_to.try(:name),
```
lowellrex: I think this will always show the name of the current organization and not the name of whoever the child task is assigned to.

lowellrex: Using [the `Appeal.location_code` method](https://github.com/department-of-veterans-affairs/caseflow/blob/36e9250177ddce645b96fa59fe919128f1449e6d/app/models/appeal.rb#L87) should give us a more accurate result here.

joeyyang:
> Noticed a bug (#8979) while reviewing this PR. Even though that bug doesn't strictly block this PR, let's fix it because I think it will make the result of this PR easier to discern.

This was the reload / Redux bug we had discussed a while back, correct? Was this the behavior that we see in table views throughout the app, or am I thinking of something different?

lowellrex:
> > Noticed a bug (#8979) while reviewing this PR. Even though that bug doesn't strictly block this PR, let's fix it because I think it will make the result of this PR easier to discern.
> 
> This was the reload / Redux bug we had discussed a while back, correct? Was this the behavior that we see in table views throughout the app, or am I thinking of something different?

I think this is a different beast entirely.

app/models/serializers/work_queue/task_serializer.rb
joeyyang: yes, much better!
screenshot 2019-01-28 16 32 36

joeyyang:
Screencap of fix for https://github.com/department-of-veterans-affairs/caseflow/issues/8979

jan-29-2019 12-45-48

joeyyang:
back to you!

lowellrex:
I'm still seeing the bug described by #8979 (return to the org queue by pressing the back button).

org-queue-bug

Let's split this PR into two separate PRs with one addressing #8979 and the other addressing #8241 (I think we're 95% the way to complete on #8241).

client/app/queue/OrganizationQueue.jsx
```diff
@@ -109,3 +109,16 @@ const TaskTableTab = ({ description, tasks }) => <React.Fragment>
     tasks={tasks}
   />
 </React.Fragment>;
+
+const NotUnassignedTaskTableTab = ({ description, tasks }) => <React.Fragment>
```
lowellrex: Maybe `TaskTableWithUserColumnTab`?

joeyyang: done

client/app/queue/components/TaskTable.jsx
```diff
@@ -170,6 +171,13 @@ export class TaskTableUnconnected extends React.PureComponent<Props> {
     } : null;
   }
 
+  caseAssignedToColumn = () => {
+    return this.props.includeAssignedTo ? {
+      header: COPY.CASE_LIST_TABLE_APPEAL_LOCATION_COLUMN_TITLE,
+      valueFunction: (task: TaskWithAppeal) => task.assignedTo.name
```
lowellrex: Can we add a `getSortValue` here as well (same as the `valueFunction`) so we can sort on this field?

PR #8962, opened 2019-01-25
fix seeds and run rake db:seed in lint task

kencheeto:


kencheeto:
Wrapping up my current investigation: I'm not sure why rake db:seed isn't working, but currently bgs fakes code only runs in the dev environment, so I think we'll want to exercise that code by calling it explicitly in a test somewhere.

PR #8966, opened 2019-01-25
Add last_submitted_at to asyncable models

pkarman:
connects #6944 

### Description
The `submitted_at` column is used for reporting metrics on the time it takes to complete EPs, so we cannot reset it.

This adds a new column `last_submitted_at` that is used exclusively by the Asyncable concern for managing jobs, including restarting them.

### DIFF
--- a/db/migrate/20190125223733_add_asyncable_last_submitted_at.rb
+++ b/db/migrate/20190125223733_add_asyncable_last_submitted_at.rb
@@ -1,5 +1,5 @@
 class AddAsyncableLastSubmittedAt < ActiveRecord::Migration[5.1]
-  def change
+  def asyncable_tables
     [
       :request_issues,
       :board_grant_effectuations,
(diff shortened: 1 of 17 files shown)


spec/feature/intake/higher_level_review/edit_spec.rb
```diff
@@ -928,7 +928,7 @@
         before_request_issue_ids: [request_issue.id],
         after_request_issue_ids: [request_issue.id],
         attempted_at: Time.zone.now,
-        submitted_at: Time.zone.now,
```
shanear: should we keep the checks on both `submitted_at` and `last_submitted_at` here?

db/migrate/20190125223733_add_asyncable_last_submitted_at.rb
```diff
@@ -0,0 +1,30 @@
...
+        execute "UPDATE #{tbl} SET last_submitted_at=#{submitted_at_column}"
+      end
+    end
+  end
+
+  def down
+    asyncable_tables.each do |tbl|
+      remove_column tbl, :last_submitted_at
+    end
+  end
```
shanear: nice! this was a tricky migration. Might want to keep a close eye on this when we deploy.

spec/feature/asyncable_jobs/asyncable_jobs_index_spec.rb
```diff
@@ -52,26 +54,29 @@
...
       safe_click "#job-HigherLevelReview-#{hlr.id}"
 
       expect(page).to have_content("Restarted")
-      expect(page).to_not have_content(hlr.establishment_submitted_at.strftime(date_format))
+      expect(page).to_not have_content(hlr.last_submitted_at.strftime(date_format))
       expect(page).to_not have_content("oops!")
 
-      expect(hlr.reload.establishment_submitted_at).to eq(now)
+      expect(hlr.reload.last_submitted_at).to eq(now)
+      expect(hlr.establishment_submitted_at).to eq(8.days.ago)
```
pkarman: here's the test that `establishment_submitted_at` did *not* change when the job was restarted.

spec/feature/intake/higher_level_review/edit_spec.rb
pkarman: I think this fixture just triggers the correct messaging on the confirmation page during intake.

there's a separate test specifically for the un-changing `submitted_at` value in the asyncable_jobs feature test.

PR #8982, opened 2019-01-28
Daily docket updates for AMA hearings

### LINKED ISSUE #8797: Daily Docket | AMA hearings
We need to ensure that the hearing schedule daily docket functionality works for AMA hearings. We should also add a feature test for AMA hearings.

Also, CO times are wrong on the daily docket.

sharonwarner:
Resolves #8797 

To test: 

Add an AMA hearing, Navigate to its daily docket, Update the hearing, Confirm the updates save.
Add an AMA hearing, Navigate to its daily docket, Postpone the hearing, Confirm the new hearing is created.

### DIFF
--- a/app/models/hearing.rb
+++ b/app/models/hearing.rb
@@ -87,6 +87,10 @@ class Hearing < ApplicationRecord
   end
   #:nocov:
 
+  def slot_new_hearing(hearing_day_id, scheduled_time, appeal)
+    Hearing.create!(hearing_day_id: hearing_day_id, scheduled_time: scheduled_time, appeal: appeal)
+  end
+
   def external_id
     uuid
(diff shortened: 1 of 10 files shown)


app/models/hearing.rb
```diff
@@ -87,6 +93,13 @@ def current_issue_count
   end
   #:nocov:
 
+  def slot_new_hearing(hearing_day_id, _scheduled_time_unused, _appeal_unused)
+    # These fields are needed for the legacy hearing's version of this method
```
lomaxap: should this be a class method and not an instance method?

lomaxap: Should this be a class method and not an instance method?

PR #8983, opened 2019-01-28
Design Updates to Order docket

Sjones352:
Resolves ##8788

### Schedule Veteran 
- [ ]  Verify that the "Scheduled Veterans" tab sorts Veterans scheduled by hearing time. 
- [ ] Verify that the "Scheduled Veterans" tab has numbers (1-11, for example) listed next to each Veteran as they are scheduled.
- [ ] Verify that there is a link to that hearing day's daily docket in the "Scheduled Veterans" tab. 

### Daily Docket 
- [  ] Verify that the Appellant/Veteran is listed in order by time slots for CO (9:00s first; 1:00s second) and Video hearings (8:30s first; 12:30s second)

### Testing Plan
1. Go to Daily Docket
2. Check if Appellant/Veteran is listed in order by time slots

screen shot 2019-01-28 at 10 17 45 am

### DIFF
--- a/client/app/hearingSchedule/components/AssignHearings.jsx
+++ b/client/app/hearingSchedule/components/AssignHearings.jsx
@@ -14,6 +14,7 @@ import { renderAppealType } from '../../queue/utils';
 import StatusMessage from '../../components/StatusMessage';
 import DocketTypeBadge from '../../components/DocketTypeBadge';
 
+
 const sectionNavigationListStyling = css({
   '& > li': {
     backgroundColor: COLORS.GREY_BACKGROUND,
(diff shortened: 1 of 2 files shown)


client/app/hearingSchedule/components/AssignHearings.jsx
```diff
@@ -289,12 +336,16 @@ export default class AssignHearings extends React.Component {
...
           {
             label: 'Scheduled Veterans',
-            page: <Table
-              columns={tabWindowColumns}
-              rowObjects={this.tableScheduledHearingsRows(scheduledOrder)}
-              summary="scheduled-hearings-table"
-              slowReRendersAreOk
-            />
+            page: <div>
+              <Link to={`/schedule/docket/${scheduledLink()}`}>
```
sharonwarner: I don't think we need a function for this, does it work if we just use ${selectedHearingDay.id}?

Sjones352: Fixed

Issue #8910, opened 2019-01-24
Intake | Change "No" to "N/A" in "Did the Veteran opt in" question


shanear:
This is a simple copy change. We want to change "No" to "N/A" for the field in the screenshot below:

<img width="701" alt="screenshot 2019-01-23 16 49 36" src="https://user-images.githubusercontent.com/1596259/51646795-eefc9c80-1f2e-11e9-83aa-21d52ac0ab9c.png">

steve-gov:
Validated on all 3 forms.

<img width="947" alt="screen shot 2019-01-28 at 5 29 49 pm" src="https://user-images.githubusercontent.com/25468689/51877843-6ca31c80-2322-11e9-9bec-961dc4188a89.png">

PR #9006, opened 2019-01-28
change "No" text to "N/A"

kencheeto:
Connects #8910 

Copy change for boolean radio option label.

screen shot 2019-01-28 at 15 30 29

and the completed intake still works:

screen shot 2019-01-28 at 15 31 35


spec/feature/intake/appeal_spec.rb
```diff
@@ -136,7 +136,7 @@
     end
 
     within_fieldset("Did they agree to withdraw their issues from the legacy system?") do
```
pkarman: this block seems ripe for an intake helper.

PR #9007, opened 2019-01-28
implementing active for HLR and added tests.

mc-usds:
Resolves #{8834}

### Description
Implemented active attribute for HLR and added tests.
SC was already done.

### Acceptance Criteria
- [x] Code compiles correctly

### Testing Plan
1. added unit test and updated the feature test.

### DIFF
--- a/app/models/higher_level_review.rb
+++ b/app/models/higher_level_review.rb
@@ -44,7 +44,7 @@ class HigherLevelReview < ClaimReview
   end
 
   def active?
-    end_product_establishments.any? { |ep| ep.status_active?(sync: false) }
+    end_product_establishments.any? { |ep| ep.status_active?(sync: false) } || dta_claim_active?
   end
 
   def description
@@ -137,5 +137,11 @@ class HigherLevelReview < ClaimReview
       benefit_type_code: veteran.benefit_type_code,
       user: intake_processed_by
     )
+  end
+
+  def dta_claim_active?
+    dta_claim = SupplementalClaim.where(veteran_file_number: veteran_file_number,
+                                        decision_review_remanded: self)
+    dta_claim.empty? ? false : dta_claim.first.active?
   end
 end
--- a/app/models/higher_level_review.rb
+++ b/app/models/higher_level_review.rb
@@ -140,8 +140,8 @@ class HigherLevelReview < ClaimReview
   end
 
   def dta_claim_active?
-    dta_claim = SupplementalClaim.where(veteran_file_number: veteran_file_number,
-                                        decision_review_remanded: self)
-    dta_claim.empty? ? false : dta_claim.first.active?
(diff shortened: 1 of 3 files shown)


app/models/higher_level_review.rb
```diff
@@ -138,4 +138,10 @@ def new_end_product_establishment(ep_code)
       user: intake_processed_by
     )
   end
+
+  def dta_claim_active?
+    dta_claim = SupplementalClaim.where(veteran_file_number: veteran_file_number,
```
pkarman: I think you could use `find_by` instead of `where` here and it would find exactly one or nil, which is what you want.

spec/feature/api/v2/appeals_spec.rb
```diff
@@ -122,6 +122,10 @@
              veteran_is_not_claimant: veteran_is_not_claimant)
     end
 
+    let!(:hlr_ep) do
+      create(:end_product_establishment, source: hlr, synced_status: "PEND")
```
pkarman: you can use `:active` instead of `synced_status: "PEND"`

spec/feature/api/v2/appeals_spec.rb
```diff
@@ -131,6 +135,10 @@
              veteran_is_not_claimant: veteran_is_not_claimant)
     end
 
+    let!(:sc_ep) do
+      create(:end_product_establishment, source: supplemental_claim_review, synced_status: "CLR")
```
pkarman: you can use `:cleared` instead of `synced_status: "CLR"`

spec/models/higher_level_review_spec.rb
```diff
@@ -288,4 +288,64 @@
       end
     end
   end
+
+  context "#active" do
+    let(:synced_status) { "CLR" }
+    let(:dta_ep_sync_status) { "PEND" }
+    let!(:hlr_with_dta_error) { create(:higher_level_review, veteran_file_number: "123456789") }
```
pkarman: if you make `veteran_file_number` also a `let` fixture I would find this easier to read (and fewer duplicated strings).

app/models/higher_level_review.rb
mc-usds: thanks! updated to use find_by

spec/feature/api/v2/appeals_spec.rb
mc-usds: fixed.

spec/feature/api/v2/appeals_spec.rb
mc-usds: fixed

spec/models/higher_level_review_spec.rb
mc-usds: fixed

PR #9008, opened 2019-01-28
Fixed VLJ filtering.

### LINKED ISSUE #8928: View Schedule | Can't filter by VLJ
The view schedule page is no longer filterable by VLJ. We should fix this functionality and add a feature test for it.

oscarAtNava:
Resolves #8928 

### Description
Fixed VLJ filtering method to account for fact that props.hearingSchedule does not contain judge full name.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to View Hearing Schedule
2. Select a VLJ to filter
3. Confirm result set is properly filtered with value selected


oscarAtNava:
Need to add feature test

PR #9010, opened 2019-01-28
Handle Address Errors in AHL Job

lomaxap:
- Handles address validation errors
- Fixes Delaware/TX data inconsistency


app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -7,7 +7,7 @@ class FetchHearingLocationsForVeteransJob < ApplicationJob
   def veterans
     @veterans ||= Veteran.where(file_number: file_numbers)
       .left_outer_joins(:available_hearing_locations)
-      .where("available_hearing_locations.updated_at < ? OR available_hearing_locations.id IS NULL", 1.month.ago)
+      .where("available_hearing_locations.updated_at < ? OR available_hearing_locations.id IS NULL", 1.week.ago)
       .limit(QUERY_LIMIT)
```
lomaxap: meredith thinks 1 month is too long

mdbenjam: Probably still fine. Though we should keep an eye on how many cases are in location 57.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -144,4 +147,18 @@ def get_state_code(va_dot_gov_address)
     msg = "#{state_code} is not a valid state code."
     fail Caseflow::Error::FetchHearingLocationsJobError, code: 500, message: msg
   end
+
+  def handle_error(error, veteran)
+    case error.message.messages[0]["key"]
+    when "DualAddressError", "AddressCouldNotBeFound", "InvalidRequestStreetAddress"
+      Raven.capture_exception(error)
+      LegacyAppeal.where(
```
mdbenjam: I think we need to do this on appeals that are in location 57. We don't necessarily want to add it to cases that are not needing to be scheduled.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -144,4 +147,18 @@ def get_state_code(va_dot_gov_address)
...
+        vbms_id: LegacyAppeal.convert_file_number_to_vacols(veteran.file_number)
+      ).each do |appeal|
+        HearingAdminActionVerifyAddressTask.create(appeal: appeal)
```
mdbenjam: We also should make a schedule hearing task as the parent. You can use this method:
```
https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/tasks/schedule_hearing_task.rb#L5
```

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -68,6 +68,9 @@ def perform
       rescue Caseflow::Error::VaDotGovLimitError
         sleep 60
         va_dot_gov_address = validate_veteran_address(veteran)
+      rescue StandardError => error
```
mdbenjam: Is the error we get more specific than standard error? It would be great to avoid catching StandardError if possible. Not a huge deal since we reraise it later.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -144,4 +147,33 @@ def get_state_code(va_dot_gov_address)
     msg = "#{state_code} is not a valid state code."
     fail Caseflow::Error::FetchHearingLocationsJobError, code: 500, message: msg
   end
+
+  def error_instructions_map
+    { "DualAddressError" => "The veteran's address in VBMS is ambiguous.",
+      "AddressCouldNotBeFound" => "The veteran's address in VBMS could not be found on a map",
```
mdbenjam: super nit: Let's make the user text consistent and add a period.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -144,4 +147,33 @@ def get_state_code(va_dot_gov_address)
...
+
+  def handle_error(error, veteran)
+    key = error.message["messages"][0]["key"]
+
+    case key
+    when "DualAddressError", "AddressCouldNotBeFound", "InvalidRequestStreetAddress"
+      LegacyAppeal.where(
+        vbms_id: LegacyAppeal.convert_file_number_to_vacols(veteran.file_number)
+      ).each do |appeal|
+        schedule_hearing_task = ScheduleHearingTask.create_if_eligible(appeal)
```
mdbenjam: We should ask product people about this. Not sure if we want to create multiple of these tasks for each valid appeal, or only on one appeal. The danger with multiple is that it might make someone work an address twice. The danger with only doing it once is that users might think it's possible to schedule an appeal that is not schedule-able. Could you get guidance from @MeredithStewart on what her preference is?

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -144,4 +147,47 @@ def get_state_code(va_dot_gov_address)
...
+    { "DualAddressError" => "The veteran's address in VBMS is ambiguous.",
+      "AddressCouldNotBeFound" => "The veteran's address in VBMS could not be found on a map.",
+      "InvalidRequestStreetAddress" => "The veteran's address in VBMS does not exist or is invalid." }
+  end
+
+  def multiple_appeals_instructions
+    "
+    Please note that this Veteran has multiple appeals. It’s possible this issue has already been resolved.
+
+    If you see a regional office and alternate hearing location task, then this task can be closed."
```
mdbenjam: take out "task" in this sentence. `If you see a regional office and alternate hearing location for this appeal, then this task can be closed.`

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -144,4 +147,47 @@ def get_state_code(va_dot_gov_address)
...
+        vbms_id: LegacyAppeal.convert_file_number_to_vacols(veteran.file_number)
+      )
+
+      veteran_appeals.each do |appeal|
+        schedule_hearing_task = ScheduleHearingTask.create_if_eligible(appeal)
+        next unless schedule_hearing_task
+
+        HearingAdminActionVerifyAddressTask.create!(
+          appeal: appeal,
+          instructions: instructions(key, has_multiple: veteran_appeals.count > 1),
```
mdbenjam: We should probably only do this if these are appeals are in 57. Since there could be old inactive appeals that aren't going to get these added tasks. They shouldn't affect our message.

Issue #9001, opened 2019-01-28
Decision Sync | Fall back to using contention disposition in the absence of a rating issue disposition


shanear:
When syncing decision issues, it is possible for a rating issue not to have a disposition, as we have seen in UAT. If that happens, we should fall back to using the disposition on the contention, since that is required by VBMS.

shanear:
Update: after discussing with Nate, it sounds like we want to always use the contention disposition

shanear:
Test 1: 700035231 030
- create 3 rating issue contentions
- map them all to the same rating issue
- set two of them to a disposition of "DTA Error - Recs"
- set one of them to "Denied"
- this should create 2 "decision issues" in caseflow, 1 for DTA Error and one for Denied.
- promulgate the rating and clear the EP
- the DTA error should automatically generate an 040

Test 2: 700035273 030
- create 2 nonrating issue contentions
- set one to a disposition of "DTA Error - {whatever}"
- set one to disposition of "Granted"
- clear the EP
- The DTA Error should automatically create an 040


These steps were validated

PR #9012, opened 2019-01-28
Allow multiple contentions to map to 1 request issue

shanear:
We learned today that multiple contentions can map to 1 rating issue. Updating the logic to account for that.

Also cleaned up an issue where we had to use `unscoped` do override scoping. 

connects #9001

### DIFF
--- a/app/models/rating_issue.rb
+++ b/app/models/rating_issue.rb
@@ -1,13 +1,11 @@
 # ephemeral class used for caching Rating Issues for client,
-# and for creating DecisionIssues when a Rating Issue has a contention_reference_id
+# and for creating DecisionIssues when a Rating Issue has contention_reference_ids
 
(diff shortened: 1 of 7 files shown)


app/models/rating_issue.rb
```diff
@@ -95,20 +92,41 @@ def ramp_claim_id
...
+
+    result.reject(&:nil?)
+  end
+
+  def calculate_contention_reference_ids
+    result = rba_contentions_data.map do |contention_data|
+      contention_data.dig(:cntntn_id)
+    end
+
+    result.reject(&:nil?)
```
pkarman: I think `compact` does this http://ruby-doc.org/core-1.9.3/Array.html#method-i-compact

shanear: ooh nice.

Issue #8622, opened 2019-01-10
Hearing Branch Admin Actions


MeredithStewart:
## User Stories 
* As a hearing coordinator, I need to be able to add a task to be completed before a Veteran can be scheduled. 

## Mocks: 
* Pages 26 - 44 here: https://va-gov.invisionapp.com/share/TUPWZLI7GBP 

## Acceptance Criteria 
* Verify that from the "Schedule Veterans" UI, the Hearing Coordinator can choose between "Schedule Veteran" and "Add Admin Action" on the Schedule Veteran task. (Page 30 & 35 of the Mocks) 
* Verify that the user can select an admin action from a list: 
    Verify power of attorney 
    Veteran is incarcerated 
    Contested Claimant issue (correct spelling error) 
    Verify address 
    Missing forms 
    FOIA/Privacy request 
    Foreign Veteran case 
    Other  
* Verify that the user can add task instructions per admin action. 
* Verify that the user can add another action and cancel that action if it is not needed. 
* Verify that the user is allowed to confirm admin actions with the copy, "Adding an admin action will move this case to your team's queue and remove it from the list of Veterans to schedule. (Page 40 of mocks) 
* Verify that the user confirm or cancel without saving admin actions.  (Page 40 of mocks) 
* Verify that the user receives a confirmation banner when the admin action has been added. Copy: "You have successfully added an Admin Action for Joe Snuffy. Joe Snuffy has been placed in your team's queue."  (Page 41 of mocks) 
* Verify that the user can add an admin action, which is blocking for the Schedule Veteran task. 
* Verify that Veterans with admin actions do not show on the list of Veterans to be scheduled (both AMA and Legacy) 
* Verify that when an admin action has been completed, the Veteran is returned to the list of Veterans to be scheduled (both AMA and Legacy) 
* Verify that when an admin action is been assigned for a Legacy appeal, that it moves out of VACOLS Location 57 (Hearing Branch) and into VACOLS Location 34 (Hearing Abeyance). 
* Verify that when an admin action has been completed for a Legacy appeal, that it moves out of VACOLS Location 34 and into Location 57. 
* Verify that when an admin action has been assigned, it appears in the Hearing Team queue. 
* Verify that when an admin action task has been placed on hold, it appears in the hold tab of that person's individual queue. 

## Related Epic 
* #8618 - EPIC | Queues and Admin Actions

MeredithStewart:
[...]

PR #9014, opened 2019-01-28
Updating copy and redirects for hearing admin actions

mdbenjam:
Connects #8622 #8623 #8626 #8630 

### Description
Two main changes:
1) Allows available actions to start passing copy for a number of actions. As well as passing where the completion of that action should take the user (queue page, case details, etc.)
2) Updates the location to "CASEFLOW" when adding an admin action for a legacy appeal.


app/models/task.rb
```diff
@@ -319,6 +326,8 @@ def create_and_auto_assign_child_task(options = {})
       child_task.parent = self
       child_task.save!
     end
+
+    update(status: :on_hold)
```
mdbenjam: We override the `after_create` hook that puts a parent task on hold below, so we need to readd it here.

app/models/tasks/generic_task.rb
```diff
@@ -61,6 +61,10 @@ def can_be_updated_by_user?(user)
     available_actions_unwrapper(user).any?
   end
 
+  # We put the parent on hold in create_many_from_params in a transaction block, so we
+  # do not need to put the parent on hold in this after create defined in task.rb
+  def put_parent_on_hold; end
```
mdbenjam: Would love to know if there's a better way to do this.

app/models/tasks/generic_task.rb
```diff
@@ -74,32 +78,42 @@ def task_is_assigned_to_users_organization?(user)
...
+          child = create_child_task(parent_task, user, params)
 
-      params = modify_params(params)
-      child = create_child_task(parent_task, user, params)
-      parent_task.update!(status: params[:status]) if params[:status]
-      child
+          child
+        end
+
+        params_array.pluck(:parent_id).uniq.map do |parent_id|
```
mdbenjam: Now setting the parent(s) to hold is in a transaction block with creating the tasks.

app/views/queue/index.html.erb
```diff
@@ -8,7 +8,6 @@
     userIsVsoEmployee: current_user.vso_employee?,
     canEditAod: AodTeam.singleton.user_has_access?(current_user),
     caseSearchHomePage: case_search_home_page,
-    organizationName: defined?(organization) ? organization.name : nil,
```
mdbenjam: moving this to the tasks endpoint so that even if a user switches to the org page after originally going to their org queue, the name will still be loaded.

client/app/queue/components/EditModalBase.jsx
```diff
@@ -18,7 +18,10 @@ export default function editModalBase(ComponentToWrap, { title, button, pathAfte
     constructor(props) {
       super(props);
 
-      this.state = { loading: false };
+      this.state = {
+        loading: false,
+        pathAfterSubmit: pathAfterSubmit || (propsToText && propsToText(this.props).pathAfterSubmit) || '/queue'
```
mdbenjam: Unfortunately we have to put  this in the state, if we don't when it goes to get the path after submit, after clicking the submit button, the task has already been updated and the data in the available actions will be missing.

mdbenjam:
Still need to debug this test: https://github.com/department-of-veterans-affairs/caseflow/blob/master/spec/feature/queue/task_queue_spec.rb#L170

app/models/tasks/generic_task.rb
```diff
@@ -74,32 +74,47 @@ def task_is_assigned_to_users_organization?(user)
...
 
   class << self
-    def create_from_params(params, user)
-      parent_task = Task.find(params[:parent_id])
-      fail Caseflow::Error::ChildTaskAssignedToSameUser if parent_task.assigned_to_id == params[:assigned_to_id] &&
-                                                           parent_task.assigned_to_type == params[:assigned_to_type]
+    def create_many_from_params(params_array, user)
+      transaction do
+        # verify access to all of the tasks
+        params_array.each do |params|
```
mdbenjam: We check access to modify tasks first, then modify them. Since you cannot act on most on hold tasks, if we try to check access as we modify them, we will put the parent task on hold and then not have access to modify the children.

app/models/task.rb
```diff
@@ -34,8 +34,8 @@ def label
   # available_actions() returns an array of options from selected by the subclass
   # from TASK_ACTIONS that looks something like:
   # [ { "label": "Assign to person", "value": "modal/assign_to_person", "func": "assignable_users" }, ... ]
-  def available_actions_unwrapper(user)
-    actions = actions_available?(user) ? available_actions(user).map { |action| build_action_hash(action, user) } : []
+  def available_actions_unwrapper(user, allow_actions_while_on_hold = false)
```
mdbenjam: Don't love this approach, but, it seems like the simplest

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -38,8 +40,18 @@ def label
     "Schedule hearing"
   end
 
+  def update_location_in_vacols
```
mdbenjam: @sharonwarner location updating logic.

spec/feature/hearing_schedule/schedule_veteran_spec.rb
```diff
@@ -117,98 +116,94 @@
...
-        click_dropdown(text: "Denver")
-        click_button("AMA Veterans Waiting")
-        click_on "Bob Smith"
-        expect(page).to have_content(COPY::TASK_SNAPSHOT_ACTIVE_TASKS_LABEL)
-
-        # Case details screen
-        click_dropdown(text: Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h[:label])
-
-        # Admin action screen
+    ensure_stable do
```
mdbenjam: Temporary. Trying to understand why this sometimes fails.

app/models/task.rb
lowellrex: Are we setting `allow_actions_while_on_hold` anywhere other than [in `Task.verify_user_can_create!`](https://github.com/department-of-veterans-affairs/caseflow/pull/9014/files#diff-c95c9451cf097bdd58f73d0f9b114f97R166)? If we are not, could we just have `Task.verify_user_can_create!` call `available_actions(user).map {...` instead of this function? The benefit of the latter approach is that I think it removes a level of indirection and makes it a little clearer that we don't care about the state of the parent task when we are checking to see if we can create a child from it.

mdbenjam: We do need to unwrap it though because the `verify_user_can_create!` digs into the data returned by the function to see the types the data function returns. We also probably don't want to allow creation of tasks with completed parents.

Though I agree with the broader point. Maybe I could refactor it to put the core of the unwrapper in another private method called `available_actions_with_data` that is called from `available_actions_unwrapper` and `verify_user_can_create!` but only `available_actions_unwrapper` checks for `actions_avilable?`

client/app/queue/components/EditModalBase.jsx
mdbenjam: It's defined in the higher ordered component definition on line 16.

app/models/task.rb
lowellrex: If that refactor wouldn't take too much effort I think it'd be worth it.

mdbenjam:
Easy enough. I'll make sure to do that refactor. Then write a few more tests and then we're good to go!

Issue #8937, opened 2019-01-25
Hide duplicate organization tasks from case timeline


lowellrex:
When we [added task instructions to case timeline](https://github.com/department-of-veterans-affairs/caseflow/issues/8483) we accidentally started showing organization tasks in the case timeline again (these were previously hidden by the work done in #7943). Additionally, since we've decided to [show incomplete tasks that aren't actionable](https://github.com/department-of-veterans-affairs/caseflow/issues/8220) in the task snapshot instead of the case timeline area as recently planned, we should consider moving the construction of the set of tasks to display in the case timeline back to the backend (#8483 moved that logic to the frontend).

![image](https://user-images.githubusercontent.com/32683958/51717226-170efd00-200e-11e9-9f76-4acb9d5f289f.png)

## Acceptance criteria
* [ ] Do not display duplicate tasks in case timeline

sneha-pai:
Sneha to follow up with Mark and Lowell.
Taking this off of Eric's plate

tomas-nava:
MEDIUM ... some combination of back- and front-end work

PR #9018, opened 2019-01-28
Add method to Task to hide from task snapshot and case timeline

lowellrex:
Refactor preparing for #7905 by adding two functions that will be set to `true` for soon-to-be-added `TrackVeteranTask`.

Instead of relying on the string value of type of the task (like we were with `RootTask`s previously), we will instead set this value to be true or false (default to false). This will allow us to control when a task is shown on the backend when we have full access to the database and the relationships between tasks (will re-enable [hiding of duplicate tasks](https://github.com/department-of-veterans-affairs/caseflow/issues/8937), for instance).

### DIFF
--- a/client/app/queue/TaskSnapshot.jsx
+++ b/client/app/queue/TaskSnapshot.jsx
@@ -3,8 +3,7 @@ import React from 'react';
 import { connect } from 'react-redux';
 import {
   appealWithDetailSelector,
-  nonRootActionableTasksForAppeal,
-  incompleteNonActionableTasks
+  taskSnapshotTasksForAppeal
 } from './selectors';
 import AddNewTaskButton from './components/AddNewTaskButton';
 import TaskRows from './components/TaskRows';
@@ -28,7 +27,6 @@ type Params = {|
 |};
 
 type Props = Params & {|
-  userRole: string,
   appeal: Appeal
 |};
 
@@ -36,14 +34,13 @@ export class TaskSnapshot extends React.PureComponent<Props> {
 
   render = () => {
     const {
-      appeal
+      appeal,
+      tasks
     } = this.props;
 
     let sectionBody = COPY.TASK_SNAPSHOT_NO_ACTIVE_LABEL;
-    const tasks = this.props.tasks.concat(this.props.nonActionableTasks);
-    const taskLength = tasks.length;
 
-    if (taskLength) {
+    if (tasks.length) {
       sectionBody = <table {...tableStyling}>
         <tbody>
           { <TaskRows appeal={appeal} taskList={tasks} timeline={false} /> }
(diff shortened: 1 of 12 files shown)


client/app/queue/TaskSnapshot.jsx
```diff
@@ -64,13 +61,9 @@ export class TaskSnapshot extends React.PureComponent<Props> {
...
 
 const mapStateToProps = (state: State, ownProps: Params) => {
-  const { userRole } = state.ui;
-
   return {
     appeal: appealWithDetailSelector(state, { appealId: ownProps.appealId }),
-    userRole,
-    tasks: nonRootActionableTasksForAppeal(state, { appealId: ownProps.appealId }),
-    nonActionableTasks: incompleteNonActionableTasks(state, { appealId: ownProps.appealId })
+    tasks: taskSnapshotTasksForAppeal(state, { appealId: ownProps.appealId })
```
lowellrex: This is a slight change in how these are ordered. Previously all actionable tasks were at the top of the task snapshot with the non-actionable tasks listed below the actionable ones in `created_at` order. Now all tasks are sorted by `created_at` date. I think this is a fine change. Open to dissent though.

joeyyang: Sounds fine to me — do you have a before / after screenshot?

joeyyang:
Will this PR require any migrations to avoid breaking existing displays / data on the frontend?

lowellrex:
> Will this PR require any migrations to avoid breaking existing displays / data on the frontend?

Nope!

client/app/queue/TaskSnapshot.jsx
lowellrex: Coming right up!

lowellrex: Before | After
--- | ---
image | image

I think this is an improvement because the timeline implied by the grey line is accurate in the after example.

joeyyang: Looking in this related ticket: https://github.com/department-of-veterans-affairs/caseflow/issues/8937, it looks like tasks are arranged most-recent-at-top.

My read of the "After" screenshot is that the most recent is at the bottom (tough to tell w/ the test data), and after looking at the timeline view linked above I'd expect most recent to be at the top. It would also make sense to me to surface any potential action that you could take at the top of the UI.

Am I reading this correctly / would it make sense to flip the order of the timeline here?

lowellrex: Case timeline is newest on top, but task snapshot is oldest on top. I guess you're right that flipping the ordering of task snapshot to match case timeline makes sense. Doing that now.

lowellrex: Oldest on top (old) | Newest on top (new with latest commit)
--- | ---
image | image

joeyyang:
LGTM!

PR #9019, opened 2019-01-29
Add columns to queue table

### LINKED ISSUE #8215: Make all columns filterable
### Context
We want to allow users to filter cases in their view by data in their various columns. One of the main use cases leading to this change is for Hearings Coordinators to filter the Hearings Management Branch team queue list of cases by Regional Office. This is especially important because of the large number of cases that will be in the Hearings Management Branch team queue at a given time.

In the future, we think many teams will want the options to cater their view by filtering a variety of columns, like just by docket type or by task type.

### Guidelines
- utilize filtering component already created in Reader and Hearing Schedule, but ensure that both the filter icon and the filter dropdown match new designs in all uses.
- Each category column which has been green-lit for filtering should automatically display all available filter options in the dropdown. If a filter value is not present in the current table's data, do not display it in the dropdown.
- The action of filtering will affect pagination and total available cases viewed by the user. This creates a need for additional information displayed near pagination component, as well as for the pagination to be truncated to reflect the newly filtered amount.

### User needs
- As a viewer of a table with filtering options, I need to be able to click on a column's filter icon, view a dropdown of items in that column, and select (via checkbox) the items that I'd like the table to be filtered by. 
- As a viewer of a table with filtering options, when I have selected items to filter by:
       -  I need to see a sentence above the table that reads: "Filtering by: <column name 1> (<number of items selected), <column name 2> (<number of items selected)" and so on and so forth, followed by a link that reads: "Clear all filters"
       - I need to see "Viewing <case range> of <total filtered case number> (<x> cases hidden). Clear all filters
         For example: "Viewing 1 - 15 of 120 filtered cases (103 cases hidden)" followed by a Clear all filters link.
- As a viewer of a table with filtering options, when I have selected items to filter by, the filter icon for that category column should display as blue (active filter), vs. black (inactive filter).
- As a viewer of a table with filtering options, when I have selected items to filter by, and I want to remove one or more filters, I can do one of the following:
    - to clear all filters at once, I can click on the "clear all filters" link. 
    - to clear an individual column filter, I can click on the filter icon for that column, and click on "Clear <column name> filter" which will always appear at the top of the filter dropdown.

#### Case Details column filter
filtering-1

#### Task column filter
filtering-2

#### Type(s) column filter
filtering-3

#### Docket column filter
filtering-4

#### Scenario when two columns have been filtered by
Take note of the effects on pagination (seen below the table), the "Filtered by: _____" string above the table, as well as the redundant display of "clear all filters link" (shown both above and below the table)
filtering-5

Similar, related ticket: https://github.com/department-of-veterans-affairs/caseflow/issues/7863

annekainicUSDS:
Resolves #8215

### Description
This PR adds a common component for filtering data in tables by column, and adds it to the required columns in the Queue table. There are several main changes here:

1. Extracted the individual components for filtering (the icon itself, the dropdown list of options, etc.), and wrapped those in a new component, `TableFilter`
2. Created a new component for each filter option in the dropdown, `FilterOption`
3. Created new props that need to be passed to each column in the table in order for filtering to work correctly:
- `enableFilter`: a boolean for whether filtering is turned on for each column,
- `tableData`: the entire data set for the table (required to calculate the options each column can be filtered on),
- `columnName`: the name of the column in the table data,
- `toggleDropdownFilterVisibility`: a function that changes the status of the filter dropdown's visibility, and dispatches an action to change a new value in the store to capture this information
- `filteredByList`: the list of filters that have been selected; this data comes from the store, and is an object where each key is a column name, which then points to an array of the specific options that column is filtered by,
- `updateFilters`: a function that updates the `filteredByList`,
- `isDropdownFilterOpen`: a property from the store that is updated by `toggleDropdownFilterVisibility`, and should receive the specific column name,
- `anyFiltersAreSet`: a boolean that determines whether the "Clear All Filters" option in the dropdown is enabled (this was a historical property, and to me seems a bit confusing, but refactoring it seemed out of scope of this PR),
- `customFilterLabels`: an object with key-value pairs translating the data values to user readable text (as much as possible I try to use the table data itself to determine what to show to the user, but sometimes this isn't possible, so I added an optional flag to accept translations for the data),
- `label`: text used for the aria-label on the icon (we could probably programmatically create this from other props),
- valueName: a string that was a historical property used in a variety ways, we can probably eliminate this prop, but it might break things elsewhere until this component is used by all apps
4. Fix styling (still waiting on the correct svg for the selected state of the filter icon)
5. New functions to store the filters (now stored in the app's store as opposed to in local component state), display the appropriate text for the filters, filter the table data based on selected filters, create a summary of selected filters at the top of the table, and update the state of which filters are selected or not

As much as possible I tried to not create changes that would not break other apps, but I could probably use some eyes on making sure I didn't inadvertently break anything (since I was touching some commonly used components).

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
Once the basic functionality is confirmed, I will write unit tests for the newly created components.


client/app/components/FilterOption.jsx
```diff
@@ -0,0 +1,49 @@
...
+  dropdownFilterViewListStyle, dropdownFilterViewListItemStyle }) => {
+
+  const handleChange = (event) => {
+    setSelectedValue(event.target.value);
+  };
+
+  return <ul {...dropdownFilterViewListStyle} {...listStyling}>
+    {options.map((option, index) => {
+      return <li className="cf-filter-option-row" key={index} {...dropdownFilterViewListItemStyle} {...listItemStyling}>
+        <input
```
mdbenjam: We have a checkbox component that might provide some useful styling here. Auto includes the label etc.

client/app/components/TableFilter.jsx
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * This component can be used to implement filtering for a table column.
+ * The required props are:
+ * - @column {array[string]} array of objects that define the properties
+ *   of the column. Possible attributes for each column include:
+ *   - @enableFilter {boolean} whether filtering is turned on for each column
+ *   - @tableData {array} the entire data set for the table (required to calculate
+ *     the options each column can be filtered on)
+ *   - @columnName {string} the name of the column in the table data
+ *   - @toggleDropdownFilterVisibility {function} changes the status of the filter
```
mdbenjam: We've been inconsistent about what should go in local state vs redux. However, I would suggest moving this into local state of this component. I don't think there's much advantage for keeping this in the redux store.

client/app/components/TableFilter.jsx
```diff
@@ -0,0 +1,161 @@
...
+ *     dropdown's visibility, and dispatches an action to change a new value in the
+ *     store to capture this information
+ *   - @filteredByList {object} the list of filters that have been selected;
```
mdbenjam: Similarly I think this component should maintain this value. I think it'll be rare that we'll need to know this information outside of this component. Maybe we keep the `updateFilters` and `filteredByLIst` props as optional so that a caller can maintain this list outside this component if they need to do something else with the filtering information.

client/app/components/TableFilter.jsx
```diff
@@ -0,0 +1,161 @@
...
+class TableFilter extends React.PureComponent {
+  filterDropdownOptions = (tableDataByRow, columnName) => {
+    let countByColumnName = _.countBy(tableDataByRow, columnName);
+    let uniqueOptions = [];
+    const filtersForColumn = _.get(this.props.column.filteredByList, columnName);
+    const { customFilterLabels } = this.props.column;
+
+    for (let key in countByColumnName) {
+      if (key && key !== 'null' && key !== 'undefined') {
+        if (customFilterLabels && customFilterLabels[key]) {
```
mdbenjam: nit: If the if and else are the same save for the dsiaplyText it might be nice to just build the displayText in the if/else and factor the rest outside. I think it makes it a little clearer what the conditional is doing. But styling preference so up to you.

client/app/components/TableFilter.jsx
```diff
@@ -0,0 +1,161 @@
...
+      }
+    } else {
+      newFilters = newFilters.concat([value]);
+    }
+
+    newList[columnName] = newFilters;
+    this.props.column.updateFilters(newList);
+
+    // For some reason when filters are removed a render doesn't automatically happen
+    this.forceUpdate();
```
mdbenjam: What if we change the component from inheriting from PureComponent to instead inherit from Component? I'm wondering if what we're changing is an complex object where only a subobject changed and so the naive `shouldComponentUpdate` function is returning false.

client/app/components/TableFilter.jsx
```diff
@@ -0,0 +1,161 @@
...
+              setSelectedValue={(value) => this.updateSelectedFilter(value, column.columnName)} />
+          </DropdownFilter>
+        }
+      </span>
+    );
+  }
+}
+
+TableFilter.propTypes = {
+  column: PropTypes.shape({
```
mdbenjam: Why expect a `column` with a predefined shape? Why not just make all these subfields top level props?

client/app/queue/components/TaskTable.jsx
```diff
@@ -126,6 +130,17 @@ export class TaskTableUnconnected extends React.PureComponent<Props> {
   caseTaskColumn = () => {
     return this.props.includeTask ? {
       header: COPY.CASE_LIST_TABLE_TASKS_COLUMN_TITLE,
+      enableFilter: true,
+      tableData: this.props.tasks,
```
mdbenjam: We already pass this to the table component. Can we just make the table component pass this along?

client/app/components/TableFilter.jsx
```diff
@@ -0,0 +1,161 @@
...
+ *   - @toggleDropdownFilterVisibility {function} changes the status of the filter
+ *     dropdown's visibility, and dispatches an action to change a new value in the
+ *     store to capture this information
+ *   - @filteredByList {object} the list of filters that have been selected;
+ *     this data comes from the store, and is an object where each key is a column name,
+ *     which then points to an array of the specific options that column is filtered by
+ *   - @updateFilters {function} updates the filteredByList
+ *   - @isDropdownFilterOpen {object} a property from the store that is updated by
+ *     toggleDropdownFilterVisibility, and should receive the specific column name
+ *   - @anyFiltersAreSet {boolean} determines whether the "Clear All Filters" option
```
mdbenjam: Also seems like something that could be maintained internally

client/app/queue/components/TaskTable.jsx
```diff
@@ -314,28 +349,104 @@ export class TaskTableUnconnected extends React.PureComponent<Props> {
...
-
-    return <Table
-      columns={this.getQueueColumns}
-      rowObjects={tasks}
-      getKeyForRow={this.props.getKeyForRow || this.getKeyForRow}
-      defaultSort={{ sortColIdx: this.getDefaultSortableColumn() }}
-      rowClassNames={(task) =>
-        this.taskHasDASRecord(task) || !this.props.requireDasRecord ? null : 'usa-input-error'} />;
+    let { tasks } = this.props;
+    const { filteredByList } = this.props;
```
mdbenjam: I think I'd move this all inside the table component. Including the:

```
          <div>
            <strong>Filtering by:</strong>
            {filterListContent}
            <span>{clearAllFiltersLink}</span>
          </div>
```

Feels like that should be common across applications.

client/app/components/TableFilter.jsx
annekainicUSDS: I kept this in the store because the other apps that use this property (reader and hearing schedule) also keep this in store and dispatch an action to update it's value. I can change it for queue, we'll just have to go back and change the other apps when this new filtering component eventually is used by them.

client/app/components/TableFilter.jsx
annekainicUSDS: I think this one needs to be stored outside of this component, because we need this information in order to figure out how to filter the table data, which happens outside of this component.

client/app/components/TableFilter.jsx
annekainicUSDS: Yes, that's definitely possible, I'll give that a try

client/app/components/TableFilter.jsx
annekainicUSDS: I don't really know why it needs to be this way, but because these props are passed per column in `TaskTable`, they get nested under `column`. Refactoring this would I think mean refactoring a bunch of things about how the app is put together. If that's something we want to explore, I think it's at least outside the scope of this PR.

client/app/queue/components/TaskTable.jsx
annekainicUSDS: That's a little tricky. The reason why `TableFilter` needs access to `tableData` is for generating the list of possible filter options in the dropdown (https://github.com/department-of-veterans-affairs/caseflow/pull/9019/files#diff-9f502b8ca40c397109e5b1a4302ba58dR38). When filters are selected, the data that is sent to `Table` is a filtered version of the original data, based on whatever filters were selected (). Sending the table data from `Table` to `TableFilter` would mean `TableFilter` only has access to the filtered version, not the complete version, and therefore wouldn't know how to generate the complete set of filter options, which is the functionality we want. So I don't think that's possible :/.

This is one of the things I had to talk through with Sneha; happy to talk through it with you and demo the functionality if it's easier to explain that way.

client/app/components/TableFilter.jsx
annekainicUSDS: I think this one also has to be passed down because it's a flag for whether or not to enable clearing each column's filters, which is something the user needs to have control over. That's why it's passed in from `TaskTable`.

client/app/components/TableFilter.jsx
```diff
@@ -0,0 +1,161 @@
...
+ *     in the dropdown is enabled
+ *   - @customFilterLabels {object} key-value pairs translating the data values to
+ *     user readable text
+ *   - @label {string} used for the aria-label on the icon,
+ *   - @valueName {string} if valueFunction is not defined, cell value will use
```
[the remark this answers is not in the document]
annekainicUSDS: Definitely, this one was a historical prop that is still used by other apps, which is why I didn't remove it for now because I'm worried it might break things if other apps are depending on it being there. Once we refactor other apps to use this filtering functionality, we can remove it.

client/app/queue/components/TaskTable.jsx
annekainicUSDS: I thought about it, but it seemed outside of what should belong in the `Table` component, which seems like it should only contain the table itself and not wrapping elements. Perhaps the best solution is creating a separate component for this functionality that could then be shared?

client/app/components/FilterOption.jsx
annekainicUSDS: I tried used the `Checkbox` component, but it was different enough and I would have to make changes to it that it didn't seem worthwhile.

client/app/queue/components/TaskTable.jsx
annekainicUSDS: Pulled this out into a new component!

annekainicUSDS:
@mdbenjam I've responded to your feedback! Where possible I pushed up new changes, and where it was not I hopefully explained why, but let me know if you have further thoughts on anything!

In regards to the tests not passing, I'm not sure how to resolve this. For some reason I am unable to run my test suite locally, so it's difficult for me to debug these. Also, I'm not sure why the lint tests have not passed CircleCi as they pass locally for me. It seems to be an issue with Flow, which doesn't seem to have run in CircleCi. If you have time to help me figure out how to run the tests locally so I can debug these, that would be awesome!

client/app/components/TableFilter.jsx
mdbenjam: Gotcha, I think I misunderstood what this field means. It seems like the name should be slightly different? Like `allowClearFilters` or something.

client/app/components/TableFilter.jsx
mdbenjam: Gotcha. I think you can use the spread operator to pass the values of an object as properties. i.e.

`<TableFilter {...column}`

client/app/queue/components/TaskTable.jsx
mdbenjam: Gotcha, that makes sense. Though I think this is a good reason for moving the storage of the filtered list out of our store, and into the table itself. Then we can avoid doing this passing.

client/app/components/TableFilter.jsx
mdbenjam: I think we should focus our update around making the ideal filtering component, and not worry about retrofitting Reader and Hearing Schedule for now. Especially if we're duplicating the table for Queue for the time being. Reader in particular has some pretty wacky filtering logic and retrofitting it can be resolved later.

I think we made mistakes in the past by pushing so much into the redux store because it makes it difficult to then reuse the component. I would love for filtering to be drop in. i.e. You say you want filtering on a column and pass i 1-2 other props to make it work.

client/app/queue/components/TaskTable.jsx
mdbenjam: I think the advantage of moving this into the component is it allows this to really be plug and play for other apps. And one of my goals is to get Caseflow to consolidate some of our styling between applications, which I think this helps to achieve.

mdbenjam:
I think it's scss-lint that's causing the linting troubles. Can you try running that locally?

As for the tests, what happens when you try to run tests locally?

client/app/components/FilterSummary.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+const FilterSummary = ({ filteredByList, alternateColumnNames, clearFilteredByList }) => {
+  let filterSummary = null;
+  let filterListContent = [];
+  const clearAllFiltersLink = <a style={{ cursor: 'pointer' }}
+    onClick={() => clearFilteredByList({})}> Clear all filters</a>;
+
+  // Don't show anything if there are no filters.
+  if (!_.isEmpty(filteredByList)) {
+    for (const filter in filteredByList) { // eslint-disable-line guard-for-in
```
mdbenjam: Instead of disabling eslint on this line, could we just use an each loop?

client/app/components/TableFilter.jsx
annekainicUSDS: Okay sounds good, I can move this into local state.

client/app/queue/QueueDropdownFilter.jsx
```diff
@@ -0,0 +1,91 @@
...
+      rootElemWidth: null
+    };
+  }
+
+  render() {
+    const { children, name } = this.props;
+
+    // Some of the filter names are camelCase, which would be displayed to the user.
+    // To make this more readable, convert the camelCase text to regular casing.
+    const displayName = _.capitalize(_.upperCase(name));
```
mdbenjam: Is this the only difference from the existing DropdownFilter?

client/app/components/TableFilter.jsx
annekainicUSDS: Ah didn't think of that, good point! I'll change this

client/app/components/TableFilter.jsx
annekainicUSDS: Definitely, this prop is not clear. It's a historical prop used by other props too, which is why I kept the name for now, but I can change it just for queue.

client/app/queue/QueueDropdownFilter.jsx
annekainicUSDS: No, there were a few other changes as well, mostly styling changes.

client/app/queue/utils.js
```diff
@@ -482,3 +482,16 @@ export const taskActionData = (props: Object) => {
   // We should never get here since any task action the creates this modal should provide data.
   throw new Error('Task action requires data');
 };
+
+export const userReadableColumnNames = {
```
annekainicUSDS: This didn't seem like the best place to put this, but I couldn't find another place that made more sense. Let me know what you think about this.

client/app/queue/utils.js
```diff
@@ -482,3 +482,16 @@ export const taskActionData = (props: Object) => {
...
+  'appeal.caseType': 'Case Type',
+  'appeal.docketName': 'Docket Number',
+  label: 'Task(s)'
+};
+
+export const userReadableDocketNameFilters = {
+  direct_review: 'Direct Review',
```
annekainicUSDS: I think I need confirmation that this is the complete set of options for the Docket Name column, and that these values are correct.

annekainicUSDS:
@mdbenjam Through your comments I came to understand what you were getting at with the concern about storing things in the app store, and I totally get it. I refactored this pretty significantly to pull everything out of the app store and instead store things in local state. Let me know what you think of the new approach!

In terms of your two questions about running tests:
1. I don't know how to run scss-lint locally... there isn't information about this in the README or an npm script in package.json. Short of installing a new package, I don't see a way to do this currently. Can you let me know what you were thinking of?
2. When I try running `bundle exec rake` (according to the README), I get the following error:
```
could not connect to server: No such file or directory (PG::ConnectionBad)
	Is the server running locally and accepting
	connections on Unix domain socket "/tmp/.s.PGSQL.5432"?
```
Perhaps it would be better to debug this off Github, so let me know if you have time to talk this through.

client/app/queue/utils.js
mdbenjam: I think that's fine for now. We could also just put it in a JSON file like: https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/constants/TASK_ACTIONS.json#L14

But not really sure this belongs in that either 🤷‍♂️

client/app/queue/utils.js
mdbenjam: Yes that looks good. There are only these four options.

client/app/queue/QueueTable.jsx
```diff
@@ -0,0 +1,346 @@
...
+  updateFilteredByList = (newList) => {
+    this.setState({ filteredByList: newList });
+  };
+
+  filterTableData = (data: Array<Object>) => {
+    const { filteredByList } = this.state;
+    const filteredData = _.clone(data);
+
+    // Only filter the data if filters have been selected
+    if (!_.isEmpty(filteredByList)) {
```
mdbenjam: A style comment, so take it or leave it, but I think this might be cleaner using a `filter` rather than cloning the data and removing fields.

```
data.filter((row) => {
  _.keys(filteredByList).every((filterColumn) => {
    if (_.isEmpty(filteredByList[filterColumn])) {
      return true;
    } else {
      return filteredByList[filterColumn].includes(row[filterColumn]);
    }
  });
});
```

Untested code so I might be off base :)

client/app/queue/components/TaskTable.jsx
mdbenjam: With the new updates, I think we can get this value inside the table component without re-passing it here.

client/app/queue/QueueTable.jsx
annekainicUSDS: Good suggestion! The code sample isn't working as it is (I think the `return` statements inside `_.key()` are returning out of each key and not returning out of each row of data). I can use `.filter`, but I need to still do that with a loop of `filteredByList` first. The way that I had written it previously was to loop through each filter first, and then loop through the data, checking if each row matched the filter. If not, I would remove that data point. Then I would loop through the data again using the next filter, further filtering it down, and so on. I think it's necessary to follow this order because filters between different columns can conflict with each other, so you want to make sure you're fulfilling them in the order they've been selected, which is why I loop through the filters first. I think using `.filter` makes at least one part of the previous implementation better, so worth doing!

mdbenjam: Interesting. Could you give an example where columns can interfere with each other? I assumed filtering was commutative.

mdbenjam: Yeah I think it's commutative, but the example is a good one. I think we might need to recalculate the list of available filters based on the filtered data instead of the complete set of data.

annekainicUSDS: Yes, definitely! That's what I did in the refactored version I pushed up. Sorry about the confusion, I got myself mixed up so many times thinking through how this should work.

client/app/components/FilterSummary.jsx
```diff
@@ -27,6 +28,13 @@ const FilterSummary = ({ filteredByList, alternateColumnNames, clearFilteredByLi
     // and then later removed, there may still be a key in the filteredByList object
     // pointing to an empty array.
     if (filterListContent.length > 0) {
+      // Add commas between filters
+      for (let i = 0; i < filterListContent.length - 1; i += 2) {
+        let commaSpan = <span key={`comma-${i}`}>, </span>;
```
mdbenjam: This feels like it could be a little cleaner. A for loop where we increment by 2 and change the thing we're iterating over while we're iterating worries me. What about something like:
```
filterListContent.map((element) => <span>{element}, </span>)
```

This will just add the comma to the <span> element already in filterListContent.

client/app/components/RenderFunctions.jsx
```diff
@@ -218,27 +218,15 @@ export class PageArrowRight extends PureComponent {
...
-      <path d="M16.8333333,20 L4.16666667,20 C2.37222222,20 1,18.6277778 1,16.8333333 L1,4.16666667 C1,2.37222222 2.37222222,1 4.16666667,1 L15,1 L16.8333333,1 C18.6277778,1 20,2.37222222 20,4.16666667 L20,16.8333333 C20,18.6277778 18.6277778,20 16.8333333,20 Z" id={pathId}></path>
-      <filter x="-10.5%" y="-10.5%" width="121.1%" height="121.1%" filterUnits="objectBoundingBox" id={filterId}>
-        <feGaussianBlur stdDeviation="1.5" in="SourceAlpha" result="shadowBlurInner1"></feGaussianBlur>
-        <feOffset dx="0" dy="1" in="shadowBlurInner1" result="shadowOffsetInner1"></feOffset>
-        <feComposite in="shadowOffsetInner1" in2="SourceAlpha" operator="arithmetic" k2="-1" k3="1" result="shadowInnerInner1"></feComposite>
-        <feColorMatrix values="0 0 0 0 0   0 0 0 0 0.443137255   0 0 0 0 0.737254902  0 0 0 1 0" type="matrix" in="shadowInnerInner1"></feColorMatrix>
-      </filter>
-    </defs>
-    <g stroke="none" strokeWidth="1" fill="none" fillRule="evenodd">
+    <g id={filterId} stroke="none" strokeWidth="1" fill="none" fillRule="evenodd">
```
mdbenjam: We can probably just remove these ids so we don't need to pass them in as props at all.

client/app/components/FilterSummary.jsx
annekainicUSDS: Yes, I struggled with figuring out how to make this clean. There needs to be some kind of conditional because we don't want to add a comma to the last filter in the list, assuming the list is longer than 1. I think I can do that with `.map` though, so I can edit this.

PR #9022, opened 2019-01-29
Always use the contention disposition for decision issue sync

shanear:
After a discussion with Nate, we decided we should ALWAYS use the contention disposition as the disposition of the decision issue. We also need to support many contentions being connected to one rating issue. This PR does that, and explains why in a comment.

connects #9001

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20190128185846) do
+ActiveRecord::Schema.define(version: 20190129002938) do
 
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -251,7 +251,7 @@
     t.datetime "profile_date"
     t.datetime "promulgation_date"
     t.string "rating_issue_reference_id"
-    t.index ["rating_issue_reference_id", "participant_id"], name: "decision_issues_uniq_idx", unique: true
+    t.index ["rating_issue_reference_id", "disposition", "participant_id"], name: "decision_issues_uniq_by_disposition_and_ref_id", unique: true
```
pkarman: +1 to db constraints!

PR #9028, opened 2019-01-29
Allow tags to contain commas

### LINKED ISSUE #7873: Allow Reader to use commas in tags
I received feedback from a few attorneys that they would like the abillity to add commas to tags in Reader. 

Current behavior: 
Reader doesn't allow users to add commas in tags. When a comma is added (in a date, for example) to a tag, Reader interprets that as a cue to create a new tag. Attorneys are using workarounds to add dates in tags, like reformatting the date or not using a comma when it makes sense to use one.

Desired behavior: 
Users are able to add commas to tags, and Reader doesn't create a new tag when a comma is used. This would help attorneys communicate effectively, avoid workarounds, and avoid accidentally creating a new tag when they use a comma.

hschallhorn:
Resolves #7873

### Description
Modified the keys that can be pressed to complete a tag and start a new one.

### Acceptance Criteria
- [X] Tags in reader can contain commas

### Testing Plan
1. Go to a document in Reader
2. Begin entering a tag that contains a comma (must be typed, not copied/pasted to test correctly)
3. Notice entering a comma does not create a new tag

### DIFF
--- a/client/app/components/SearchableDropdown.jsx
+++ b/client/app/components/SearchableDropdown.jsx
@@ -87,6 +87,18 @@ class SearchableDropdown extends React.Component<Props, ComponentState> {
     }
   }
 
+  // Override the default keys to create a new tag (allows creating options that contain a comma)
+  shouldKeyDownEventCreateNewOption = ({ keyCode }) => {
+    switch (keyCode) {
+    // Tab and Enter only
+    case 9:
+    case 13:
+      return true;
+    default:
+      return false;
+    }
+  };
+
   render() {
     const {
       options,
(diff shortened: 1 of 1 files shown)


client/app/components/SearchableDropdown.jsx
```diff
@@ -87,6 +87,18 @@ class SearchableDropdown extends React.Component<Props, ComponentState> {
     }
   }
 
+  // Override the default keys to create a new tag (allows creating options that contain a comma)
```
hschallhorn: fn being overridden: https://github.com/JedWatson/react-select/blob/v1.0.0-rc.10/src/Creatable.js#L222

PR #9029, opened 2019-01-29
Slot New Hearing API

lomaxap:
Resolves merge conflicts with `slot_new_hearing` methods

### DIFF
--- a/app/controllers/hearings_controller.rb
+++ b/app/controllers/hearings_controller.rb
@@ -9,10 +9,12 @@ class HearingsController < ApplicationController
   end
 
   def update
-    slot_new_hearing
+    if postponed?
+      slot_new_hearing
+      params["hearing"].delete(:hearing_location_attributes)
+    end
 
     if hearing.is_a?(LegacyHearing)
-
       hearing.update_caseflow_and_vacols(update_params_legacy)
       # Because of how we map the hearing time, we need to refresh the VACOLS data after saving
       HearingRepository.load_vacols_data(hearing)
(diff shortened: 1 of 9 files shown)


spec/feature/hearing_schedule/daily_docket_spec.rb
```diff
@@ -73,20 +73,5 @@
       # created a ticket that we can look into after February.
       # expect(page).to have_content("8:30 am")
     end
-
-    scenario "User can postpone a hearing" do
```
sharonwarner: Bad merge?

app/models/hearing.rb
```diff
@@ -98,10 +98,11 @@ def current_issue_count
   end
   #:nocov:
 
-  def slot_new_hearing(hearing_day_id, _scheduled_time_unused, _appeal_unused)
+  def slot_new_hearing(hearing_day_id, hearing_location_attrs: nil, **_args)
     # These fields are needed for the legacy hearing's version of this method
     Hearing.create!(hearing_day_id: hearing_day_id,
                     scheduled_time: scheduled_time,
```
lomaxap: we may actually need to pass scheduled time. On the daily docket, if a user changes the time, they may expect the postponed hearing to have that time.

spec/feature/hearing_schedule/daily_docket_spec.rb
lomaxap: yea, I moved this method up and changed it a little and it didn't see that it was the same.

app/controllers/hearings_controller.rb
```diff
@@ -111,6 +112,17 @@ def update_params_legacy
                                      ])
   end
 
+  def master_record_params
+    params.require("master_record_updated").permit(:id,
+                                                   time: [:h, :m, :offset],
```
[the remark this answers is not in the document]
sharonwarner: Got it. I'd like to carve out some time to refactor this after April 1st, so this is fine for now.

Issue #8330, opened 2018-12-18
Intake Edit | Prevent editing request issues on appeals that have been outcoded


shanear:
If an appeal has been outcoded, then it's issues should no longer be editable.

## Acceptance Criteria
- If a user visits the edit issues page for an appeal that has been outcoded, show them a full page message with the copy below

## Copy
Title: Issues Not Editable
Text: This appeal has been outcoded and the issues are no longer editable.

## Mocks
It should look exactly like the page from https://github.com/department-of-veterans-affairs/caseflow/issues/7386

steve-gov:
Validated on several attempts, see https://github.com/department-of-veterans-affairs/caseflow/issues/6167#issuecomment-459099718 for one example.

PR #9033, opened 2019-01-29
Outcoded Appeals may not be edited

pkarman:
connects #8330 

### Description
Once outcoded, an Appeal may no longer be edited.


pkarman:
screen shot 2019-01-29 at 10 07 46 am

PR #9035, opened 2019-01-29
Navigate between judge assign and review views with the queue selector dropdown component

### LINKED ISSUE #8119: Navigate to judge assign view by the queue selector dropdown
#8012 added the dropdown selector to navigate between a user's personal task queue and queues for any organizations they are a part of. #8117 moved attorney membership on judge team teams into that same organizations model. With the groundwork laid by these two tickets, we can move the link to the judge assign cases view into that dropdown.

In addition to the benefits accrued by unifying around a single queue table view (reducing the front-end's reliance on a user's role, for one), when this is implemented we will be able to accomodate judges acting as attorneys simply by create a `JudgeTeam` for that attorney.

| Before | After |
| --- | --- |
| image | image |

tomas-nava:
Resolves #8119 

### Before:
judge-nav-before

### After:
judge-nav-after


client/app/queue/components/QueueJudgeViewSelectorDropdown.jsx
```diff
@@ -0,0 +1,38 @@
+// @flow
```
joeyyang: nit: I'd prefer a name like `JudgeAssignAndReviewDropdown` that tells you more "what it does" than "what/where it is" — up to you!

client/app/queue/components/QueueJudgeViewSelectorDropdown.jsx
```diff
@@ -0,0 +1,38 @@
...
+        label: COPY.ASSIGN_MODE_LINK_LABEL
+      }
+    ];
+
+    return <QueueSelectorDropdown items={items} />;
+  }
+}
+
+QueueJudgeViewSelectorDropdown.propTypes = {
+  userId: PropTypes.number.isRequired
```
joeyyang: is the `userId` always going to be a `number`?

client/app/queue/components/QueueOrganizationSelectorDropdown.jsx
```diff
@@ -0,0 +1,45 @@
...
+
+    const items = organizations.map((org, index) => {
+      return {
+        key: (index + 1).toString(),
+        href: (location === org.url) ? '#' : `/organizations/${org.url}`,
+        label: sprintf(COPY.CASE_LIST_TABLE_QUEUE_DROPDOWN_TEAM_CASES_LABEL, org.name)
+      };
+    });
+
+    items.unshift(
```
joeyyang: nit: Since `unshift` is a bit uncommon, from a readability perspective I'd love to have this statement come first and then concat the organizations array onto this, so something like:
```javascript
const items = [
  {
    key: '0',
    href: queueHref,
    label: COPY.CASE_LIST_TABLE_QUEUE_DROPDOWN_OWN_CASES_LABEL
  }
];

items.concat(organizations.map(...))
```

or 
```javascript
const firstLink = {
  key: '0',
  ...
};

const organizationLinks = organizations.map(...)

const items = firstLink.concat(organizationLinks) // OR const items = [...firstLink, ...organizationLinks]
```

client/app/queue/components/QueueSelectorDropdown.jsx
```diff
@@ -47,33 +46,26 @@ export default class QueueSelectorDropdown extends React.Component<Props, Compon
...
-            {COPY.CASE_LIST_TABLE_QUEUE_DROPDOWN_OWN_CASES_LABEL}
-          </Link>
-        </li>
-
-        {organizations.map((org, index) => {
-          const orgHref = (location === org.url) ? '#' : `/organizations/${org.url}`;
+        {items.map((item) => {
+          const linkProps = {
+            className: 'usa-button-secondary usa-button',
+            onClick: this.onMenuClick,
```
joeyyang: I know I'm the dummy who wrote the `onMenuClick` method but we should change the name to "what it does", so maybe `toggleDropdownMenu`?

client/COPY.json
```diff
@@ -122,7 +122,8 @@
   "ORGANIZATIONAL_QUEUE_EMPTY_STATE_MESSAGE": "This queue is empty. You can ",
 
   "JUDGE_CASE_REVIEW_TABLE_TITLE": "Review %s Cases",
-  "SWITCH_TO_ASSIGN_MODE_LINK_LABEL": "Switch to Assign Cases",
+  "ASSIGN_MODE_LINK_LABEL": "Assign Cases",
```
joeyyang: maybe something in the name of key here to hint that it's part of a queue switcher dropdown?

client/app/queue/components/QueueJudgeViewSelectorDropdown.jsx
tomas-nava: Yeah, I think so; it definitely is now, and I see the same pattern used elsewhere in the codebase for database IDs.

PR #9039, opened 2019-01-29
8616 remove special issues for ama

### LINKED ISSUE #8616: Remove special issues page for AMA appeals
Note: we can only do this when we've confirmed EP creation for AMA appeals is ready.

we can *hide* the special issues page from attorneys for AMA appeals. attorneys will *not* need to select special issues, nor will we need to pre-populate them, because the VBMS National Work Queue (NWQ) will be able to recognize special issues attributes that already exist in VBMS and route EPs correctly.

for example: one of the special issues is `incarcerated Veteran`. Appeals Management Office (AMO) has agreed that we can rely on VBMS having a "flash" that indicates the Veteran is incarcerated. This "flash" is on the Veteran, irrespective of the appeal.

in the future, surfacing these "flashes" is likely something we'll want to do so BVA employees know more about the Veteran, but that is solving a different problem then routing EPs to the correct VBA regional office for updating the Veteran's benefit.

[Slack post](https://dsva.slack.com/archives/C6E41RE92/p1547074950068100)

## AC
- Special issues page does not show for AMA appeals during attorney or judge check out

youngfreezyVA:
Resolves #8616

### Description

for AMA appeals, remove special issues for attorneys and judges.

This required changing route definitions, changing the way we pushed on to the browser history in the ActionsDropDown component, as well as changing back button urls which referenced special issues.  Redux was used to store the specialIssuesRedirect as well as whether an appeal was legacy or not, this was stored in `state.queue.appealDetails[appealId].isLegacy` and `state.queue.appealDetails[appealId].specialIssuesRedirect`.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Login as an attorney or judge
2. go to queue
3a. select a legacy appeal -> select ready for decision -> you should see special issues -> then when you click next you should see dispositions
3b. select a non legacy appeal -> select ready for decision -> you should NOT see special issues and see dispositions instead. when you click back, you should go back to Case Details View


youngfreezyVA:
@aroltsch  looks like `draft_decision/evaluate` is broken closing PR until I can dig deeper.

client/app/queue/CaseDetailsView.jsx
```diff
@@ -63,6 +64,7 @@ class CaseDetailsView extends React.PureComponent {
         regionalOffice
       });
     }
+    this.props.setAppealType(appeal.isLegacyAppeal, appealId);
```
lowellrex: Won't this already be set by [`extractAppealsFromTasks`](https://github.com/department-of-veterans-affairs/caseflow/blob/2cdbfaf119e695cf9108d2c805a8d3f94e7c303d/client/app/queue/utils.js#L101) and [`prepareAppealForStore`](https://github.com/department-of-veterans-affairs/caseflow/blob/2cdbfaf119e695cf9108d2c805a8d3f94e7c303d/client/app/queue/utils.js#L240)? Why do we need to reset it here?

client/app/queue/QueueApp.jsx
```diff
@@ -428,7 +438,8 @@ QueueApp.propTypes = {
 };
 
 const mapStateToProps = (state: State) => ({
-  reviewActionType: state.queue.stagedChanges.taskDecision.type
+  reviewActionType: state.queue.stagedChanges.taskDecision.type,
+  appealDetails: state.queue.appealDetails
```
lowellrex: Let's use the `appealWithDetailSelector` here [like we do in `CaseDetailsView`](https://github.com/department-of-veterans-affairs/caseflow/blob/2cdbfaf119e695cf9108d2c805a8d3f94e7c303d/client/app/queue/CaseDetailsView.jsx#L131).

client/app/queue/LegacySelectDispositionsView.jsx
```diff
@@ -80,16 +80,16 @@ class LegacySelectDispositionsView extends React.PureComponent {
...
       appeal
     } = this.props;
 
     if (appeal.isLegacyAppeal) {
       return `/queue/appeals/${appealId}`;
     }
-
-    return `/queue/appeals/${appealId}/tasks/${taskId}/${checkoutFlow}/special_issues`;
+    // we will never be in a state where we go back to special_issues from here, right?
+    // return `/queue/appeals/${appealId}/tasks/${taskId}/${checkoutFlow}/special_issues`;
```
lowellrex: What should our fall-through case be here instead? Maybe throw an error?

client/app/queue/CaseDetailsView.jsx
youngfreezyVA: so I was confused because the `isLegacyAppeal` wasn't being set in the `appealDetails` node/hash in the store, but it was in the `appeals` hash. I updated `appealDetails` to also include this information in `prepareAppealForStore` and removed this function call

lowellrex: We generally aim to only store information in the redux store once (either `appeals` or `appealDetails` for instance) and then use a selector to pull that information out of the store if we need the combination of two different hashes for the same entity. We use [`appealWithDetailSelector`](https://github.com/department-of-veterans-affairs/caseflow/blob/6cacef3488b95318666d22590fea41e830618ef2/client/app/queue/selectors.js#L114) to give us the combination of `appeals` and `appealDetails` for any given appeal so let's try to re-use that here.

youngfreezyVA: got it - turns out I don't need the if that had the need for that info anyways, but will definitely use `appealWithDetailSelector` in the future

client/app/queue/QueueApp.jsx
youngfreezyVA: no longer needed here but will use it in the future

client/app/queue/LegacySelectDispositionsView.jsx
youngfreezyVA: i removed the if condition. for ama, you only get here from appeal details, so the if isn't needed and you always go back to `/queue/appeals/${appealId}`

client/app/queue/QueueApp.jsx
```diff
@@ -1,4 +1,5 @@
 // @flow
+
```
lowellrex: I don't think there were any logical changes this file, can we back out these cosmetic changes for a cleaner git history?

youngfreezyVA: done!

client/app/queue/LegacySelectDispositionsView.jsx
```diff
@@ -79,17 +79,11 @@ class LegacySelectDispositionsView extends React.PureComponent {
...
-      taskId,
-      checkoutFlow,
-      appeal
+      appealId
     } = this.props;
 
-    if (appeal.isLegacyAppeal) {
-      return `/queue/appeals/${appealId}`;
-    }
+    return `/queue/appeals/${appealId}`;
```
lowellrex: Do we want to leave this logic in place? We can end up at this page through the legacy appeal checkout flow (special issues first, then disposition selection) or the AMA appeal checkout flow (skip special issues page, start at disposition selection). For AMA appeals won't we want to go directly back to the case details page and for legacy appeals won't we want to go back to the special issues page?

lowellrex: Although the code as it was written did the exact opposite...

youngfreezyVA: haha yeah exactly... so i essentially left the code as is for the legacy appeal, but then the same thing should happen for AMA...so I removed the `if`. i'm not sure from a business standpoint if this is what we want... so feel free to let me know either way. @aroltsch any thoughts?

aroltsch: So for legacy appeals, we should go back to the special issues page when we click the back button on select dispositions view and for ama appeals, we should go back to the case details.

youngfreezyVA: done!

Issue #8674, opened 2019-01-13
Appeal Intake | Create record request task


shanear:
When an appeal intake is processed, if it contains issues for a non-comp non-pension business line, we need to generate a task in their queue to send them documents necessary to process the appeal.

## Acceptance Criteria
- When an appeal intake was successfully processed. If it had any issues of a non-comp non-pension business line, a "Record Request" task should be created in the appropriate business lines queue/queues.
- "Record Request" should be an option in the task filter on the non-comp landing page
 
## Additional Notes
When the board intakes an appeal for a non-VBMS LOB, that LOB must be notified to send the relevant records to the board for review.

To facilitate this:

- Create a "record request" task
- Add the task to the LOB's intake queue
- Link the task to a "record request" confirmation page
- On the confirmation page, allow the user to confirm that they have provided the necessary records
- Once confirmed, mark the task "completed" so that it appears in the Completed tab

Out of scope:
- What happens when you click the link for this task. We will handle the page for this task in a future issue.

carolaponce:
Designs, pending address, will confirms once I have that info confirmed.
As a user, the task here is to verify the records have been sent by clicking on check box.

Copy:
Title - **Request to send Veteran record to the Board**
Description - The Veteran/appellant has filed a Notice of Disagreement at the Board of Veterans' Appeals. In order to decide that appeal, the Board will need the complete records from [BUSINESS LINE]. Please take necessary steps to send this record to the Board.
Mail to: ADDRESS

![non-comp_recordreq_completed](https://user-images.githubusercontent.com/11700545/51857255-99d1d980-22e6-11e9-9671-15c7cfb96eaf.png)
![non-comp_recordreq_completed](https://user-images.githubusercontent.com/11700545/51857257-9b030680-22e6-11e9-8998-b8ac1e36a96d.png)

carolaponce:
![non-comp_recordreq](https://user-images.githubusercontent.com/11700545/51858008-84f64580-22e8-11e9-991e-a84c8a420ccb.png)

steve-gov:
Validated:

<img width="628" alt="screen shot 2019-01-30 at 12 20 09 pm" src="https://user-images.githubusercontent.com/25468689/52010183-ba866480-2489-11e9-9f66-fef77beb3736.png">

[...]

PR #9041, opened 2019-01-29
Create VeteranRecordRequest for non-comp request issues

pkarman:
connects #8674 

### Description
A new task `VeteranRecordRequest` created whenever an Appeal request issue has a benefit_type that is not compensation or pension.


app/controllers/decision_reviews_controller.rb
```diff
@@ -40,15 +40,19 @@ def task_id
   end
 
   def task
-    @task ||= DecisionReviewTask.find(task_id)
+    @task ||= Task.includes([:appeal, :assigned_to]).find(task_id)
```
pkarman: we have more than one kind of task subclass now.

app/controllers/decision_reviews_controller.rb
```diff
@@ -40,15 +40,19 @@ def task_id
...
   end
 
   def in_progress_tasks
-    apply_task_serializer(business_line.tasks.order(assigned_at: :desc).reject(&:completed?))
+    apply_task_serializer(
+      business_line.tasks.includes([:assigned_to, :appeal]).order(assigned_at: :desc).reject(&:completed?)
```
pkarman: this just quiets the bullet warnings about eager loading.

app/models/appeal.rb
```diff
@@ -329,6 +330,18 @@ def program
 
   private
 
+  def create_business_line_tasks
+    request_issues.select(&:requires_record_request_task?).each do |req_issue|
+      business_line = req_issue.business_line
+      VeteranRecordRequest.create!(
+        parent: root_task,
```
pkarman: not sure if we need to make this the child of the Root task or not.

app/models/serializers/work_queue/decision_review_task_serializer.rb
```diff
@@ -17,7 +17,7 @@ def claimant_name
   end
 
   def claimant_relationship
-    return "self" unless decision_review.claimants.any?
+    return "self" unless decision_review.veteran_is_not_claimant
```
pkarman: fixes logic that would bite us eventually.

pkarman:
screen shot 2019-01-29 at 3 43 17 pm

app/models/serializers/work_queue/decision_review_task_serializer.rb
leikkisa: 👍

app/models/appeal.rb
pkarman: confirmed yes with @lowellrex

Issue #8979, opened 2019-01-28
Limit table to organization tasks in organization queue


lowellrex:
In the organization queue's table view we show cases related to [all tasks we have in the redux store](https://github.com/department-of-veterans-affairs/caseflow/blob/36e9250177ddce645b96fa59fe919128f1449e6d/client/app/queue/selectors.js#L135). We should limit this to only showing cases related to tasks that are assigned to the current organization.

### Steps to reproduce
* Navigate to any organization queue
* Click on the link to the case details page for any case which has tasks assigned to somebody else
* Load the case details page (this loads all tasks for the current appeal)
* Return to the queue table view
* Notice that there are now table rows for each task, even the ones not assigned to the current organization

_Example of this behaviour in action:_

![too-many-org-tasks](https://user-images.githubusercontent.com/32683958/51840910-5c1b8380-22db-11e9-9328-292e4bf412fe.gif)

## Acceptance criteria
* [ ] Limit rows in the organization queue table view to tasks assigned to that organization

PR #9043, opened 2019-01-29
only use tasks for the organization in organization queue views

joeyyang:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/8979

### Description
Fix bug where tasks from outside the organization were making their way into the organization queue view. 

jan-29-2019 17-10-14

### DIFF
--- a/client/app/queue/selectors.js
+++ b/client/app/queue/selectors.js
@@ -133,19 +133,19 @@ export const getAllTasksForAppeal = createSelector(
 );
 
 export const getUnassignedOrganizationalTasks = createSelector(
-  [tasksWithAppealSelector],
+  [tasksByOrganization],
   (tasks: Tasks) => _.filter(tasks, (task) => {
     return (task.status === TASK_STATUSES.assigned || task.status === TASK_STATUSES.in_progress);
   })
 );
 
 export const getAssignedOrganizationalTasks = createSelector(
-  [tasksWithAppealSelector],
+  [tasksByOrganization],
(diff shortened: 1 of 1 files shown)


lowellrex:
I think we need to check if the tasks' `assignedTo.type` is an Organization in [the `tasksByOrganization` selector](https://github.com/department-of-veterans-affairs/caseflow/blob/63d64de29978509abace0e6c961d61350a7eb4cc/client/app/queue/selectors.js#L90). We actually might need to add an element to the `assignedTo` property of the task serializer that asks if the task `assigned_to.is_a?(Organization)` on the backend so we don't need to specify ever different type of possible organization we might come across.

joeyyang:
In this example above, all the tasks `assignedTo.type` is equal to `"Colocated"`. This is the intended behavior, correct?

lowellrex:
> In this example above, all the tasks `assignedTo.type` is equal to `"Colocated"`. This is the intended behavior, correct?

That's the reason why we'll need to do this check on the back-end. When we set this value we select the most specific type (`Colocated` instead of `Organization`) which is fine in most cases, but in this particular case we only care that the assignee is an instance of `Organization`. Since we don't have that hierarchy on the front-end (we don't know on the front-end that `Colocated` inherits from `Organization`) we can either hard-code everything that inherits from `Organization` or we can do that check on the back-end and pass it to the front-end as a specific attribute. Maybe in the task serializer something like:

```ruby
attribute :assigned_to do
  {
    is_organization: object.assigned_to.is_a?(Organization),
    ...
```

joeyyang:
great, back to you!

Issue #9027, opened 2019-01-29
Automatic case distribution parameters should be relaxed


amprokop:
Currently, judges can only request cases if they've zeroed out their "Cases to Assign" list. The "Assign cases" button only appears when the "Cases to Assign" list is empty.

We want to relax this — judges can request a new batch of cases *unless* they have 1 or more cases that have been sitting, unassigned, for more than 14 days, or they have more than 8 cases assigned to them.

# AC
- "Assign cases" button is always present
- Use existing design pattern — add two new error states
    - You have cases that have been waiting to be assigned for more than 14 days.
    - You have more than 8 cases.

PR #9046, opened 2019-01-29
tweaks acd

amprokop:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/9027

image

image

### Testing Plan
`FeatureToggle.enable!(:automatic_case_distribution)`


spec/feature/queue/task_queue_spec.rb
```diff
@@ -378,15 +378,26 @@
...
+          parent: root_task,
+          appeal: root_task.appeal
+        )
       end
       let!(:legacy_review_task) do
         FactoryBot.create(:legacy_appeal, vacols_case: FactoryBot.create(:case, :assigned, user: judge_user))
       end
 
-      it "should display both legacy and caseflow review tasks" do
+      it "should display both legacy and caseflow tasks for review and assign tasks page" do
```
amprokop: just adding a little more feature test coverage for this page, just a check that the page loads and that it loads the cases

app/models/distribution.rb
```diff
@@ -61,26 +62,50 @@ def validate_user_is_judge
...
+  def longer_than_two_weeks_ago(date)
+    return false if date.nil?
+
+    date.beginning_of_day < 14.days.ago.beginning_of_day
   end
 
   def assigned_tasks
     pending_statuses = [Constants.TASK_STATUSES.assigned, Constants.TASK_STATUSES.in_progress]
     judge.tasks.select do |t|
-      assigned_legacy_task = t.is_a?(JudgeLegacyTask) && t.action == "assign"
```
amprokop: none of the tasks returned by `judge.tasks` will be legacy tasks — i think the original line of code here was unneeded

db/seeds.rb
```diff
@@ -664,6 +665,12 @@ def create_tasks
     create_task_at_judge_review(@ama_appeals[8], judge, attorney)
     create_task_at_colocated(@ama_appeals[8], judge, attorney, colocated)
 
+    9.times do
```
amprokop: creating some more test data so we can easily test out error states

laurjpeterson:
@amprokop - what happens if a user tries to request when more than 8 AND one or more case has been on their plate for 14+ days? Do we show both error messages?

app/models/distribution.rb
```diff
@@ -61,26 +62,50 @@ def validate_user_is_judge
...
+  end
+
+  def judge_has_eight_or_fewer_unassigned_cases
+    return false if judge_tasks.length > 8
+
+    judge_tasks.length + judge_legacy_tasks.length <= 8
+  end
+
+  def judge_cases_waiting_longer_than_two_weeks
+    return true if judge_tasks.any? { |task| longer_than_two_weeks_ago(task.assigned_at) }
```
cmgiven: Nice optimization.

Issue #8676, opened 2019-01-13
Appeal DTA Errors | Set the DTA payee code for Appeal DTA errors correctly based on historic data


shanear:
For appeals intake, users do not input the payee code of the claimant.

If we create a follow up supplemental claim, then we don't know what payee code the claimant should be when we create the EPs.

Is there a way to properly determine the payee code even though the intake at the Board did not select it. 

AMO Said if it would help, we could narrow the list down to:
<img width="533" alt="screen shot 2019-01-15 at 9 55 27 am" src="https://user-images.githubusercontent.com/1596259/51199708-591c9e00-18ac-11e9-9bdf-505664f75b51.png">

related to https://github.com/department-of-veterans-affairs/caseflow/issues/8649

## Acceptance Criteria
- When setting the payee code on an 040 EP generated by a Board Appeal DTA error, use the logic from https://github.com/department-of-veterans-affairs/caseflow/issues/8649 to determine what it should be set to.

### Validation
- Find or create a HLR or SC with a non-Veteran claimant. This will ensure that Veteran has an end product which stores the payee code for that claimant.
- Create an Appeal with the same Veteran. Select the same claimant as the claimant.
- Have Colin decide an issue on the Appeal with a DTA error
- Check that the payee code on the new Supplemental Claim is populated correctly for the claimant.

leikkisa:
Just taking some notes here:
C&P is for compensation and pension
CH35 is an Education award, which right now does not require payee code
DIC is available to a spouse or child for a deceased veteran - ask AMO if this should always be used for deceased veterans or if we might ever want to use the regular "spouse" or "child"

Institutional Vets seems related to Vets that have a fiduciary, more info here: https://www.knowva.ebenefits.va.gov/system/templates/selfservice/va_ssnew/help/customer/locale/en-US/portal/554400000001018/content/554400000014264/M21-1,-Part-III,-Subpart-v,-Chapter-6,-Section-E---Institutional-Awards-Specific-to-Incompetent-Veterans

leikkisa:
It seems like we can only automatically select the payee code in the case that the claimant has a previous claim for the same Veteran file number.  I wonder if this will always be the case (since it's an appeal and presumably there was an original claim), but if not then we'll need to find another solution for the Appeal DTA Supplemental Claims.

Some info from Nathan Jolly and Mary Kate:
[...]

PR #9047, opened 2019-01-29
Add relationship-based default payee codes

leikkisa:
connects #8676

### DIFF
--- a/app/models/relationship.rb
+++ b/app/models/relationship.rb
@@ -2,7 +2,7 @@ class Relationship
   include ActiveModel::Model
   include ActiveModel::Validations
 
-  attr_accessor :participant_id, :first_name, :last_name, :relationship_type, :veteran_file_number
+  attr_accessor :participant_id, :first_name, :last_name, :relationship_type, :veteran_file_number, :gender
 
   class << self
     def from_bgs_hash(veteran, hash)
@@ -11,7 +11,8 @@ class Relationship
         participant_id: hash[:ptcpnt_id],
         first_name: hash[:first_name],
         last_name: hash[:last_name],
-        relationship_type: hash[:relationship_type]
+        relationship_type: hash[:relationship_type],
+        gender: hash[:gender]
       )
     end
   end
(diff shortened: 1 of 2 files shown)


spec/models/relationship_spec.rb
```diff
@@ -1,32 +1,86 @@
...
   let(:veteran) { create(:veteran) }
+  let(:relationship_type) { "Spouse" }
+  let(:gender) { nil }
 
   let(:relationship) do
     Relationship.new(
       veteran_file_number: veteran.file_number,
       participant_id: "1234",
-      first_name: "BOB",
+      first_name: "TORY",
```
leikkisa: Since the gender is changing around :-P

spec/models/relationship_spec.rb
```diff
@@ -1,32 +1,86 @@
...
+              relationship_type: "Parent",
+              default_payee_code: "50"
+            )
+          end
+        end
+
+        context "when the parent is female" do
+          let(:gender) { "F" }
+
+          it "returns the Father payee code" do
```
jiahuang: should this be "Mother"?

spec/models/relationship_spec.rb
```diff
@@ -1,32 +1,86 @@
...
+        end
+      end
+
+      context "when the claimant is an unhandled relationship type" do
+        let(:relationship_type) { "Other" }
+
+        it "does not have a default payee code" do
+          expect(subject).to include(
+            relationship_type: "Other",
+            default_payee_code: nil
```
jiahuang: just wanted to verify that we don't want "00" here

app/models/relationship.rb
```diff
@@ -11,7 +11,8 @@ def from_bgs_hash(veteran, hash)
         participant_id: hash[:ptcpnt_id],
         first_name: hash[:first_name],
         last_name: hash[:last_name],
-        relationship_type: hash[:relationship_type]
+        relationship_type: hash[:relationship_type],
+        gender: hash[:gender]
```
shanear: did we validate this is present in the bgs hash?

app/models/relationship.rb
```diff
@@ -29,7 +30,27 @@ def ui_hash
...
+    when "Child"
+      "11"
+    when "Parent"
+      case gender
+      when "M"
+        "50"
+      when "F"
+        "60"
+      end
+    end
```
shanear: there is a condensed syntax for `case` that might read better here:
```
case relationship_type
when "Spouse" then "10"
when "Child" then "11"
when "Parent" then gender == "M" ? "50" : "60"
end
```

leikkisa: I started off with a ternary but figured if the gender was blank, it would be better to return nil

spec/models/relationship_spec.rb
leikkisa: Right now it's not defaulting in all cases (although we would eventually want that for appeal DTA supplemental claims).  So if they pick someone with a relationship we're not handling right now, it just doesn't get defaulted.

app/models/relationship.rb
leikkisa: Yes, it is there

PR #9048, opened 2019-01-29
Rename last_submitted_at -> establishment_last_submitted_at

pkarman:
Make column names consistent.


db/migrate/20190129233723_rename_last_submitted_at_pt3.rb
```diff
@@ -0,0 +1,26 @@
+class Appeal < ApplicationRecord
+  self.ignored_columns = ["last_submitted_at"]
+end
+
+class RenameLastSubmittedAtPt3 < ActiveRecord::Migration[5.1]
```
shanear: i like this migration pattern

Issue #8803, opened 2019-01-18
Decision Sync | DTA Error EP fails to establish because of no date


shanear:
We got a claim date error when trying to establish a DTA error EP. This is because `approx_decision_date` from the decision issue was null.

```
irb(main):032:0> SupplementalClaim.where(veteran_file_number: "124592531").first.decision_review_remanded.decision_issues.first.approx_decision_date
[2019-01-17 19:19:18 -0500]   SupplementalClaim Load (1.2ms)  SELECT  "supplemental_claims".* FROM "supplemental_claims" WHERE "supplemental_claims"."veteran_file_number" = $1 ORDER BY "supplemental_claims"."id" ASC LIMIT $2  [["veteran_file_number", "124592531"], ["LIMIT", 1]]
[2019-01-17 19:19:18 -0500]   HigherLevelReview Load (1.0ms)  SELECT  "higher_level_reviews".* FROM "higher_level_reviews" WHERE "higher_level_reviews"."id" = $1 LIMIT $2  [["id", 879], ["LIMIT", 1]]
[2019-01-17 19:19:18 -0500]   DecisionIssue Load (1.0ms)  SELECT  "decision_issues".* FROM "decision_issues" WHERE "decision_issues"."decision_review_id" = $1 AND "decision_issues"."decision_review_type" = $2 ORDER BY "decision_issues"."id" ASC LIMIT $3  [["decision_review_id", 879], ["decision_review_type", "HigherLevelReview"], ["LIMIT", 1]]
=> nil
```

leikkisa:
Have added an error when this happens to help diagnose it.

leikkisa:
I did look at that code before.  These could still be two different problems since the above was for a ClaimReview.

aroltsch:
@shanear when creating decision issues, do we need to set a profile date? 
https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/concerns/issue_updater.rb#L48
If yes, where do I get this info?

leikkisa:
@aroltsch We just merged a PR that will get the decision date for Appeals from the Decision Document, and Appeals shouldn't need the profile date

leikkisa:
For claim reviews which get the end product last action date, we've discovered that the date was being misinterpreted (the day and month were getting switched), resulting in incorrect dates or invalid dates. In the case of invalid dates, this would prevent the new Supplemental Claim from being created, because the Decision Issue was missing the approximate decision date.

[...]

PR #9049, opened 2019-01-29
Use decision document for appeal decision date

leikkisa:
connects #8803


app/models/decision_issue.rb
```diff
@@ -75,6 +75,16 @@ def contention_text
 
   private
 
+  def appeal_decision_date
+    return unless appeal?
+
+    decision_review.decision_document.decision_date
```
shanear: may get burned by some of these objects not existing. should we use `&.`?

leikkisa: 👍

PR #9050, opened 2019-01-30
make board grants only show specific business line

jiahuang:
Connects #8299

Task has a decision_review, which has all the request issues regardless of type. But appeals can have multiple tasks assigned to each business line. This filters out the request issues that are of other benefit types, which prevents the decision issues of the other business types from displaying.


app/models/tasks/board_grant_effectuation_task.rb
```diff
@@ -17,6 +17,18 @@ def complete_with_payload!(_decision_issue_params, _decision_date)
     update!(status: Constants.TASK_STATUSES.completed, completed_at: Time.zone.now)
   end
 
+  def appeal_data
+    # for board grants, filter out request issues
+    # that are not the correct business line
+    business_line = assigned_to.becomes(BusinessLine)
```
shanear: this should be a private method

app/models/tasks/board_grant_effectuation_task.rb
```diff
@@ -17,6 +17,18 @@ def complete_with_payload!(_decision_issue_params, _decision_date)
...
+    request_issues = appeal.request_issues
+      .select { |issue| issue.benefit_type == business_line.url }
+      .map(&:ui_hash)
```
shanear: this should also be a private method: `request_issues_filtered_by_business_line`?

spec/feature/non_comp/board_grants_spec.rb
```diff
@@ -104,4 +105,54 @@ def submit_form
...
+      end
+    end
+
+    let!(:in_progress_vha_task) do
+      create(:board_grant_effectuation_task, :in_progress, appeal: appeal, assigned_to: vha_org)
+    end
+
+    let(:vha_dispositions_url) { "decision_reviews/vha/tasks/#{in_progress_vha_task.id}" }
+
+    scenario "displays board grants page with correct dispositions", focus: true do
```
shanear: `focus: true`

shanear: I think it's probably okay to just have one test where there are multiple organizations, since this test basically tests everything in the test with just one org.

shanear: we also have both cases tested in a unit test.

jiahuang: the existing tests test for actions done on the page. I think having this separate test for multiple issues makes the other tests a bit easier to read since there isn't as much data setup.

app/models/tasks/board_grant_effectuation_task.rb
```diff
@@ -17,6 +17,18 @@ def complete_with_payload!(_decision_issue_params, _decision_date)
     update!(status: Constants.TASK_STATUSES.completed, completed_at: Time.zone.now)
   end
 
+  def appeal_data
```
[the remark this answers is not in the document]
jiahuang: i changed it to appeal_ui_hash since `DecisionReviewTask` also needs to have the same method. I think the call to `request_issues_by_benefit_type` clarifies what this method does for board grants.

PR #9051, opened 2019-01-30
Appeal Status: Add support for events attribute for HLR and SC

mc-usds:
Resolves #{8999}

### Description
Adding support for events attribute in HLR and SC

### Acceptance Criteria
- [x] Code compiles correctly

### Testing Plan
1. updated unit and feature test


app/models/appeal_events.rb
```diff
@@ -5,6 +5,16 @@ class AppealEvents
   attr_accessor :version
 
   def all
+    if appeal.is_a?(LegacyAppeal)
```
cmgiven: Consider writing:

```ruby
case appeal
when LegacyAppeal
  legacy_appeal_events
when SupplementalClaim
  sc_events
when HigherLevelReview
  hlr_events
end
```

app/models/higher_level_review.rb
```diff
@@ -64,8 +64,36 @@ def issues
     []
   end
 
+  def decision_event_date
+    if !dta_claim && decision_issues.any?
```
pkarman: just a stylistic note. elsewhere we seem to use [guard clauses](https://www.thechrisoshow.com/2009/02/16/using-guard-clauses-in-your-ruby-code/) rather than nested conditionals. so e.g.

```ruby
def decision_event_date
  return if dta_claim
  return unless decision_issues.any?

  # etc
end

File: app/models/higher_level_review.rb [hunk 3]
```diff
@@ -64,8 +64,36 @@ def issues
...
+      if end_product_establishments.any?
+        decision_issues.first.approx_decision_date
+      else
+        decision_issues.first.promulgation_date
+      end
+    end
+  end
+
+  def dta_error_event_date
+    if !hlr_ep_active? && dta_claim
```
pkarman: likewise

```ruby
def dta_error_event_date
  return if hlr_ep_active
  return unless dta_claim

  # etc
end
```

app/models/higher_level_review.rb
```diff
@@ -64,8 +64,36 @@ def issues
...
+
+  def other_close_event_date
+    if !active? && decision_issues.empty? && end_product_establishments.any?
+      end_product_establishments.first.last_synced_at
+    end
+  end
+
   def events
-    # need to implement. hlr_request, hlr_decision, hlr_dta_error, or hlr_other_close
+    @events ||= AppealEvents.new(appeal: self).all
```
pkarman: nice!

app/models/supplemental_claim.rb
```diff
@@ -58,8 +58,24 @@ def issues
     []
   end
 
+  def decision_event_date
+    if decision_issues.any?
```
pkarman: `return unless decision_issues.any?` etc

spec/models/higher_level_review_spec.rb
```diff
@@ -351,4 +351,116 @@
...
+               veteran_file_number: veteran_file_number,
+               receipt_date: receipt_date)
+      end
+
+      let(:last_synced_at) { receipt_date + 20.days }
+      let(:synced_status) { "CLR" }
+      let!(:hlr_ep) do
+        create(:end_product_establishment,
+               source: hlr,
+               synced_status: synced_status,
```
pkarman: you can use the `:cleared` trait instead of `synced_status: synced_status`

check out the `spec/factories/end_product_establishment.rb` file for other pre-defined traits.

spec/models/higher_level_review_spec.rb
```diff
@@ -351,4 +351,116 @@
...
+        expect(request_event.date.to_date).to eq(receipt_date.to_date)
+
+        closed_event = events.find { |e| e.type == :hlr_other_close }
+        expect(closed_event.date.to_date).to eq(last_synced_at.to_date)
+      end
+    end
+
+    context "hlr has dta error and remanded sc decision" do
+      let(:synced_status) { "CLR" }
+      let(:veteran_file_number) { "123456789" }
```
pkarman: no need to redefine the `:veteran_file_number` if an enclosing `context` already defines it the same way.

app/models/appeal_events.rb
mc-usds: Thanks, I changed to to a case statement.

app/models/higher_level_review.rb
mc-usds: Fixed to use guard clauses.  thanks!

spec/models/higher_level_review_spec.rb
mc-usds: fixed to use :cleared

spec/models/higher_level_review_spec.rb
mc-usds: removed this.

PR #9052, opened 2019-01-30
Appeal Status: adding support for aod in Appeal

mc-usds:
Resolves #{8835}

### Description
Adding support for aod attribute in Appeal

### Acceptance Criteria
- [x] Code compiles correctly

### Testing Plan
1. updated feature test


app/models/appeal.rb
```diff
@@ -303,7 +303,7 @@ def linked_review_ids
   end
 
   def aod
```
cmgiven: We can alias the attribute in the serializer instead of creating this method. I believe that would look like:

```
attribute :advanced_on_docket, key: :aod
```

Or maybe the other way around.

mc-usds:
this branch got out of sync with my local.  Will close and resubmit.

PR #9053, opened 2019-01-30
Use headless driver by default for tests

monfresh:
**Why**: When tests run in Chrome, focus is stolen from other
applications, which can be frustrating. More often than not, folks
prefer to run tests in the background.

**How**:
- Use an env var to determine which driver to use, and set the
default driver to the headless one.

- Move the capybara setup to its own file in spec/support to make it
easier to find and change

- Remove code related to Sauce Labs, which we aren't using anymore


monfresh:
Good point. I will reverse the logic.

pkarman:
You might also use the `CI` env var instead too, since that already has meaningful impact on linting and other CI-related stuff (like running headless).

monfresh:
In order for this to work in Circle CI, we need to update our Docker Ruby image to 2.5.3 so that we can use a more recent version of Chrome. We will put the Ruby upgrade on hold for now until after AMA day.

Issue #9034, opened 2019-01-29
Non Comp | Process document request task flow


shanear:
This is the flow used to handle and close out tasks created in https://github.com/department-of-veterans-affairs/caseflow/issues/8674

requires https://github.com/department-of-veterans-affairs/caseflow/issues/8674

## Acceptance Criteria
When a user clicks on a Document request task in a non-comp screen,
- they should be routed to a page that looks like the mock below.
- the "Confirm" button should be disabled until the "I certify this record..." checkbox is checked
- when the "Confirm" button is clicked. The tasks should be completed and recorded. (similar toan effectuation task as described in https://github.com/department-of-veterans-affairs/caseflow/issues/8299) The user should be routed back to the non-comp in progress screen.

![non-comp_recordreq](https://user-images.githubusercontent.com/11700545/51858008-84f64580-22e8-11e9-991e-a84c8a420ccb.png)

## Copy
Title - Request to send Veteran record to the Board
Description - The Veteran/appellant has filed a Notice of Disagreement at the Board of Veterans' Appeals. In order to decide that appeal, the Board will need the complete records from [BUSINESS LINE]. Please take necessary steps to send this record to the Board.
Mail to: [ADDRESS]

[425 I St NW, Washington DC, 20001.]
^^ Place holder

steve-gov:
Validated: 

![screencapture](https://user-images.githubusercontent.com/25468689/52018489-730ad300-249f-11e9-9dd1-6958e8a46f9c.gif)

PR #9067, opened 2019-01-30
Pek non comp record request

pkarman:
connects #9034 

### Description
Adds a new page for Veteran Record Request task.

screen shot 2019-01-30 at 12 57 31 pm


client/app/nonComp/constants.js
```diff
@@ -12,3 +12,5 @@ export const DECISION_ISSUE_UPDATE_STATUS = {
   SUCCEED: 'SUCCEED',
   FAIL: 'FAIL'
 };
+
+export const BOA_ADDRESS = '425 I St NW, Washington DC, 20001';
```
jiahuang: Will this need to be displayed as two lines to account for something like:
```
Attn: Board of Appeals
425 I St NW, Washington DC, 20001
```

pkarman: I believe this is a placeholder. @carolaponce is following up.

pkarman: ticket #9072

Issue #8881, opened 2019-01-23
Intake | Errors with BGS relationships request can cause User to get stuck on error screen forever


shanear:
Today someone received an error when fetching the veteran relationships in UAT.

```
Legacy POA for vet not found on Corporate Database. The POA data was not stored on the Corporate Database, however, the rest of the transaction was stored on the Corporate Database
```

The bigger issue though was that she was stuck on the error page and could not cancel the intake. We need to fix this scenario.

![image 3](https://user-images.githubusercontent.com/1596259/51581978-0da35a80-1e7f-11e9-9fbb-a4b68d2e8ac8.png)

## Technical Notes
This is highly related to https://github.com/department-of-veterans-affairs/caseflow/issues/5480. I think the solution is to continue to capture specific dependency related errors instead of a global catch. If it's an issue with caseflow, we should simply fix the issue.

nicholasholtz:
@shanear Is this still being investigated?

shanear:
![test](https://user-images.githubusercontent.com/1596259/52092778-8f754100-256d-11e9-9333-66dc75d26c52.gif)
works 👍

PR #9074, opened 2019-01-30
catch all intake controller errors

jiahuang:
Connects #8881

* Cancels the existing intake so user doesn't get stuck
* shows a detailed message to the user

screen shot 2019-01-30 at 11 58 12 am


app/controllers/intakes_controller.rb
```diff
@@ -60,6 +60,30 @@ def error
...
+      feedbackUrl: feedback_url,
+      buildDate: build_date,
+      featureToggles: {
+        intakeAma: FeatureToggle.enabled?(:intakeAma, user: current_user),
+        legacyOptInEnabled: FeatureToggle.enabled?(:intake_legacy_opt_in, user: current_user),
+        useAmaActivationDate: FeatureToggle.enabled?(:use_ama_activation_date, user: current_user)
+      }
+    }
+
+  rescue Exception => e
```
jiahuang: Maybe too aggressive in cancelling? Should this catch only BGS errors?

Issue #9064, opened 2019-01-30
Intake | Surface DTA HLRs where the Supplemental Claim failed to create in Caseflow


shanear:
We experienced an issue where a DTA Supplemental Claim failed to create as a result of a decision issue sync in caseflow. This did not fail any Async jobs, and thus was not surfaced aside from 

We need to either:

1. fail the Request Issue sync in Caseflow so it will appear in the jobs UI.

or 

2. have some other mechanism for surfacing DTA HLRs where the supplemental claim failed to create.

leikkisa:
Plan is to add validation to Decision Issue, which will prevent Request Issue sync from completing.

leikkisa:
Just putting Jia's PR in here, because it will be relevant when we move away from Promulgation date: https://github.com/department-of-veterans-affairs/caseflow/pull/9111

shanear:
"Decisions issues syncing..." isn't showing up in UAT

PR #9076, opened 2019-01-30
Add syncing message if an ep's request issues are still syncing decision issues

shanear:
connects #9064


client/COPY.json
```diff
@@ -48,6 +48,8 @@
   "OTHER_REVIEWS_TABLE_EMPTY_TEXT": "This Veteran has no higher level reviews or supplemental claims at this time.",
   "OTHER_REVIEWS_TABLE_ESTABLISHMENT_FAILED": "Establishment failed. Support notified.",
   "OTHER_REVIEWS_TABLE_ESTABLISHING": "Establishing...",
+  "OTHER_REVIEWS_TABLE_SYNCING_DECISIONS": "Syncing decisions...",
+  "OTHER_REVIEWS_TABLE_SYNCING_DECISIONS_ERROR": "Decisions sync failed. Support notified.",
```
joeyyang: Just out of curiosity, how does support get notified in this case? Wondering if someone who sees this will know what to do if they see this.

shanear: We see it on the failed jobs page

Issue #8619, opened 2019-01-10
Appeal Intake | Set benefit type on rating issues


pkarman:
When a rating request issue is created, we need to determine whether the contested rating issue is a pension or a compensation rating issue.

## Acceptance Criteria
- When a rating request issue is created, we should record the benefit type as "pension" or "compensation" depending on whether the rating issue is a pension or compensation. 

## Technical Notes
We will need to figure out how to determine whether a rating issue is pension or compensation. There may be some helpful info in here: https://github.com/department-of-veterans-affairs/appeals-design-research/issues/1032

## Notes
Follow on to #8543 

See comment https://github.com/department-of-veterans-affairs/caseflow/pull/8577#discussion_r246584046

pkarman:
suspect SOAP response fields: `bnftClmTc` and `isuDescr`

pkarman:
after much email and investigation, we decided to treat any `bnftClmTc` that ends with `PMC` as a pension type.

shanear:
confirmed our criteria is correct with AMO today 

- any rating issue decided on a PMC EP is considered a PMC issue. (Confirmed by Laila 01/31/2019)

steve-gov:
Intook a NOD against a rated Pension issue:

<img width="219" alt="screen shot 2019-02-06 at 2 13 36 pm" src="https://user-images.githubusercontent.com/25468689/52377498-eb324500-2a19-11e9-87e9-fbd2e4072eb4.png">

Selected Direct Review docket and Pension line of business, and selected the only rated issue on the screen, the Pension issue:

<img width="516" alt="screen shot 2019-02-06 at 2 19 46 pm" src="https://user-images.githubusercontent.com/25468689/52377646-5bd96180-2a1a-11e9-93b5-ccc221059a4f.png">

This created appeal https://appeals.cf.uat.ds.va.gov/queue/appeals/74b1036e-357d-4002-a5d5-748cbd156c8a, which listed the benefit type as Compensation:

<img width="411" alt="screen shot 2019-02-06 at 2 21 38 pm" src="https://user-images.githubusercontent.com/25468689/52377723-904d1d80-2a1a-11e9-8b5c-9dcf8d228528.png">

I'm not sure that's being explicitly validated in this issue, but wanted to add it as a data point.

steve-gov:
@pkarman Is there a way to know the `bnftClmTc` for the rated issue in my last comment -- the one with the 5004 diagnostic code?

[...]

PR #9081, opened 2019-01-30
Rating issue benefit type

pkarman:
connects #8619 

### Description
Use the rating_issue.benefit_type on request issues.


app/models/request_issue.rb
```diff
@@ -491,7 +491,7 @@ def find_or_create_decision_issue_from_rating_issue(rating_issue)
       decision_text: rating_issue.decision_text,
       profile_date: rating_issue.profile_date,
       decision_review: review_request,
-      benefit_type: benefit_type,
+      benefit_type: rating_issue.benefit_type,
```
pkarman: I'm not sure this is correct.

PR #9084, opened 2019-01-30
AMA AHL Job

lomaxap:
- Adds query to get all schedule hearing tasks- 
- Excludes veterans with open Address admin actions


app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -5,19 +5,31 @@ class FetchHearingLocationsForVeteransJob < ApplicationJob
...
   def file_numbers
-    # TODO: will need an AMA equivalent of this query
     @file_numbers ||= VACOLS::Case.where(bfcurloc: 57).pluck(:bfcorlid).map do |bfcorlid|
       LegacyAppeal.veteran_file_number_from_bfcorlid(bfcorlid)
     end
   end
 
+  def schedule_hearing_tasks
+    @schedule_hearing_tasks ||= ScheduleHearingTask.where.not(status: "completed")
+      .joins("
```
mdbenjam: IS this to determine tasks that are not blocked by hearing admin tasks? Why do we need to exclude those tasks? Don't we still want to update the ScheduleHearingTask's regional office etc. even in this case.

If that is the purpose of this call, I think we can achieve it more easily by just looking for when the status is `assigned`. When the ScheduleHearingTask has a child, it'll get put on hold.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -174,7 +186,8 @@ def handle_error(error, veteran)
       tasks = LegacyAppeal.where(
         vbms_id: LegacyAppeal.convert_file_number_to_vacols(veteran.file_number)
       ).map do |appeal|
-        ScheduleHearingTask.create_if_eligible(appeal)
+        ScheduleHearingTask.create_if_eligible(appeal) ||
+          schedule_hearing_tasks.select { |task| task.appeal.id == appeal.id }
```
mdbenjam: I think it might be clearer to just add an AMA branch to the `create_if_eligible` method.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
lomaxap: This is just to exclude the tasks that are blocked specifically by an `HearingAdminActionVerifyAddressTask` and in the future an `HearingAdminActionForeignVeteranCaseTask`, in those cases, we've already determined that we can't work out regional office without the coordinator updating the Veteran's address. 

I don't think `assigned` is specific enough since it could be blocked for reasons other than their address didn't validate.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
lomaxap: i think the method would need to be name `find_or_create_if_eligible`

mdbenjam: I think changing the name makes sense.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
mdbenjam: Gotcha, so if we don't exclude these then we'll always try to get the addresses for the same invalid appeals because they won't have any `hearing_location` records.

PR #9098, opened 2019-01-31
Fix order of events in case timeline

### LINKED ISSUE #9070: Case timeline displays attorney and judge tasks out of order
screen shot 2019-01-30 at 2 12 07 pm

This is a screenshot of Case timeline from a Veteran's case. The judge and attorney tasks are out of order. The judge task should come after the attorney task, per the dates on each task.

lowellrex:
Resolves #9070.

PR #9018 introduced a bug where the events in the case timeline are [sorted in the wrong order](https://github.com/department-of-veterans-affairs/caseflow/pull/9018/files#diff-2a3f4e296b35f6015a2d1d3ef72e5506R208) (by created_at date instead of completed_at date). This PR corrects that by sorting events in case timeline by completed_at date.

Before | After
--- | ---
image | image


client/app/queue/selectors.js
```diff
@@ -220,7 +220,7 @@ export const rootTasksForAppeal = createSelector(
 export const caseTimelineTasksForAppeal = createSelector(
   [getAllTasksForAppeal],
   (tasks: Tasks) => _.orderBy(_.filter(completeTasksSelector(tasks), (task) =>
-    !task.hideFromCaseTimeline), ['createdAt'], ['desc'])
+    !task.hideFromCaseTimeline), ['completedAt'], ['desc'])
```
joeyyang: Is there a chance tasks won't have a `completedAt`, i.e., won't be complete?

lowellrex: The `completeTasksSelector` we call on line 222 ensures that tasks here will be complete.

PR #9101, opened 2019-01-31
Only compare dates when they exist

pkarman:
See https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3545/


app/models/intake_stats.rb
```diff
@@ -115,7 +115,9 @@ def intake_series_status(intakes)
 
     average_election_response_time_by_notice_date: lambda do |range|
       elections = RampElection.established.where(notice_date: offset_range(range))
-      response_times = elections.map { |e| e.receipt_date.to_time.to_f - e.notice_date.to_time.to_f }
+      response_times = elections.select { |e| e.receipt_date && e.notice_date }.map do |e|
+        e.receipt_date.to_time.to_f - e.notice_date.to_time.to_f
+      end
```
monfresh: If `receipt_date` and/or `notice_date` are missing, `response_times` will be an empty array, and `average(response_times)` will return `NaN`. Is that going to be a problem when it is called by the frontend in `app/views/intake_stats/show.html.erb`? Should we add a test for this?

pkarman: good questions @monfresh 

the intake stats page is benignly neglected, but not yet replace-able. The test coverage is ... minimal.

I just added a fix to the `average` private method to avoid dividing by zero.

This image shows what the user sees when there are `0` results. I think we can rely on no regressions in this case.

screen shot 2019-01-31 at 10 08 13 am

PR #9105, opened 2019-01-31
Fix veteran name caching logic

pkarman:
connects #8963 

### Description
For performance reasons we cache the Veteran's name in Caseflow. However, we were never updating the name if it changed upstream.


app/models/veteran.rb
```diff
@@ -176,16 +176,34 @@ def validate_address
...
+
   class << self
     def find_or_create_by_file_number(file_number)
       find_and_maybe_backfill_name(file_number) || create_by_file_number(file_number)
     end
 
     def find_by_file_number_or_ssn(file_number_or_ssn)
       if file_number_or_ssn.to_s.length == 9
-        find_by(file_number: file_number_or_ssn) || find_by_ssn(file_number_or_ssn)
+        find_and_maybe_backfill_name(file_number_or_ssn) || find_by_ssn(file_number_or_ssn)
```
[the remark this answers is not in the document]
pkarman: > I like the idea of doing it when the intake starts, and then again when the EP is created.

The method `find_by_file_number_or_ssn` is currently only used when intake starts, in the [intakes_controller](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/intakes_controller.rb#L105).

It seems like the window is pretty small between when intake starts and the EP is created (minutes or hours). Is the risk of name change upstream great enough to warrant re-fetching the veteran before creating the EP?

We're in the realm of the 2nd hardest problem now: cache invalidation.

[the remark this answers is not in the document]
pkarman: Put another way: if you want to avoid the performance hit, you use `find_by()`.

The `find_by_file_number*` methods necessarily have the performance hit. Seems like it is developer choice at that point.

The bug here, IMO, was that `find_by_file_number_or_ssn` was not checking the cache like `find_or_create_by_file_number` was. Now at least they are consistent, and the definition of "stale cache" has been expanded to include differences, not just nil-in-caseflow.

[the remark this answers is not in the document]
shanear: > It seems like the window is pretty small between when intake starts and the EP is created (minutes or hours). Is the risk of name change upstream great enough to warrant re-fetching the veteran before creating the EP?

That's what I would have thought, but the scenario is that a CA can begin an intake, then notice the name is wrong and fix it in VBMS. If this happens, it would overwrite their change. The business is pretty concerned about this scenario.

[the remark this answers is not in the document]
pkarman: > The business is pretty concerned about this scenario.

Fair enough. I will add a feature test for that scenario and make sure we are re-checking before establishing EPs.

spec/feature/intake/veteran_name_caching_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+        Fakes::BGSService.veteran_records[veteran.file_number][:last_name] = "Changed"
+      end
+
+      click_intake_finish
+
+      expect(page).to have_content("Intake completed")
+    end
+
+    step "EPs use the updated Veteran name" do
+      expect(bgs).to have_received(:fetch_veteran_info).exactly(5).times
```
pkarman: the `sync` flag feature effectively saves 1 call to BGS. Not at all sure it's worth that slim performance savings.

pkarman:
@shanear I added an optional `sync` flag to the `find_by_file_number*` methods. After some profiling it looks like it only saves on 1 call to BGS over the entire intake cycle. Still, at least now developers can choose to sync or not on each Veteran fetch.

app/models/end_product_establishment.rb
```diff
@@ -463,7 +463,7 @@ def invalid_modifiers
   end
 
   def veteran
-    @veteran ||= Veteran.find_or_create_by_file_number(veteran_file_number)
+    @veteran ||= Veteran.find_or_create_by_file_number(veteran_file_number, sync: true)
```
pkarman: we sync right before we create end products.

app/models/veteran.rb
```diff
@@ -176,36 +176,62 @@ def validate_address
...
+      is_stale = true if self[name] != bgs_record[name]
+    end
+    is_stale
+  end
+
   class << self
-    def find_or_create_by_file_number(file_number)
-      find_and_maybe_backfill_name(file_number) || create_by_file_number(file_number)
+    def find_or_create_by_file_number(file_number, sync: false)
+      find_and_maybe_backfill_name(file_number, sync: sync) || create_by_file_number(file_number)
```
shanear: i feel like this method could be written something like this:

```
find_by_file_number(file_number, backfill_name: true)
```

since `sync` seems to just be the thing that triggers the backfilling of the name

app/models/veteran.rb
```diff
@@ -176,36 +176,62 @@ def validate_address
...
       if file_number_or_ssn.to_s.length == 9
-        find_by(file_number: file_number_or_ssn) || find_by_ssn(file_number_or_ssn)
+        find_and_maybe_backfill_name(file_number_or_ssn, sync: sync) || find_by_ssn(file_number_or_ssn, sync: sync)
       else
-        find_by(file_number: file_number_or_ssn)
+        find_and_maybe_backfill_name(file_number_or_ssn, sync: sync)
+      end
+    end
+
+    def find_or_create_by_file_number_or_ssn(file_number_or_ssn, sync: false)
```
shanear: I don't see where this method is used. It seems like it was added in this PR though

pkarman: It was added, yes, mostly just for symmetry. It is used on in the specs right now.

app/models/veteran.rb
pkarman: yes I can rename to `sync_name` instead of `sync`.

You are correct; it basically just gates the call to `bgs_record` to avoid fetching the record unless we really need it.

app/models/veteran.rb
shanear: yeah I'm sort of against adding methods that are not used in the code, because probably YAGNI, and it gets out of hand fast.

PR #9111, opened 2019-01-31
get decision date based on decision issues

jiahuang:
Connects #8031

Tasks don't have decision date, so find it based off of the decision issue. Decision issue dates are saved as UTC and we convert to local time in the browser, which throws off the date. Modified the date format to be able to get a date from UTC time.


app/controllers/decision_reviews_controller.rb
```diff
@@ -66,7 +66,7 @@ def business_line
   def decision_date
     return unless task.instance_of? DecisionReviewTask
 
-    Date.parse(allowed_params.require("decision_date")).to_datetime
+    Date.parse(allowed_params.require("decision_date")).in_time_zone.utc.to_datetime
```
shanear:
```
.in_time_zone.utc
```

this seems like too many things. Shouldn't `.utc` do the same thing?

client/app/nonComp/components/Disposition.jsx
```diff
@@ -138,6 +138,12 @@ class NonCompDispositions extends React.PureComponent {
 
     let completeDiv = null;
 
+    let decisionDate = this.state.decisionDate;
+
+    if (appeal.decisionIssues.length > 0) {
+      decisionDate = formatDateStrUtc(appeal.decisionIssues[0].promulgationDate);
```
shanear: might want to check with sally on whether we are using `promulgation_date` for non-comp issues

client/app/util/DateUtil.js
```diff
@@ -18,16 +18,18 @@ export const formatDate = function(dateString) {
...
+  const day = StringUtil.leftPad(date.getDate(), 2, '0');
+  const year = date.getFullYear();
 
   return `${month}/${day}/${year}`;
 };
 
 // Date format YYYY-MM-DD
-export const formatDateStr = (dateString, dateFormat = 'YYYY-MM-DD', expectedFormat = dateFormatString) => {
+export const formatDateStr = (dateString, dateFormat = 'YYYY-MM-DD',
+  expectedFormat = dateFormatString, utc = false) => {
```
pkarman: it's not clear to me what the `utc` param means here. Does it mean that `dateString` is in the UTC timezone? Or that it should be converted to the UTC timezone?

client/app/util/DateUtil.js
```diff
@@ -43,7 +45,17 @@ export const formatDateStr = (dateString, dateFormat = 'YYYY-MM-DD', expectedFor
     }
   }
 
-  return moment(dateString, dateStringFormat).format(expectedFormat);
+  let date = moment(dateString, dateStringFormat);
+
+  if (utc) {
+    date = date.utc();
```
pkarman: this converts the timezone? does it also shift the time?

spec/feature/non_comp/dispositions_spec.rb
```diff
@@ -3,6 +3,7 @@
 feature "NonComp Dispositions Task Page" do
   before do
     FeatureToggle.enable!(:decision_reviews)
+    Timecop.freeze(Time.zone.today)
```
pkarman: the browser time + zone will not reflect this - does that matter?

spec/feature/non_comp/dispositions_spec.rb
```diff
@@ -86,13 +87,13 @@ def find_disabled_disposition(num, disposition, description = nil)
...
     scenario "saves decision issues" do
       visit dispositions_url
-
       expect(page).to have_button("Complete", disabled: true)
 
       # set description & disposition for each request issue
       fill_in_disposition(0, "Granted")
       fill_in_disposition(1, "Granted", "test description")
       fill_in_disposition(2, "Denied", "denied")
+      fill_in "decision-date", with: "01/01/2019"
```
pkarman: this date is unknown to me - is it just arbitrary?

client/app/util/DateUtil.js
jiahuang: It means that the datestring being passed in doesn't have a timezone attached to it, so moment will attempt to convert to the local timezone unless utc conversion is specified.

client/app/util/DateUtil.js
jiahuang: Correct, the time is shifted to be the same time but in UTC.

client/app/util/DateUtil.js
pkarman: ah ok. Maybe `forceUTC` would be clearer?

Issue #9107, opened 2019-01-31
Intake | Don't wait a day to sync nonrating decision issues


shanear:
Right now we are waiting a day to sync decision issues from BGS, because that's about how long it takes for the promulgated ratings to show up in the service.

However, we don't need to find any promulgated ratings for non-rating EP decision issue syncs, yet we are still waiting a day. We should not be waiting a day.

jiahuang:
note: For non-rating, do not wait at all

shanear:
This worked today

PR #9113, opened 2019-01-31
sync nonrating ep immediately

jiahuang:
Connects #9107 

Sync nonrating request issues immediately.


spec/models/end_product_establishment_spec.rb
```diff
@@ -774,8 +776,9 @@
       it "submits each request issue and starts decision sync job" do
         subject
 
-        expect(request_issues.first.reload.decision_sync_submitted_at).to_not be_nil
-        expect(request_issues.second.reload.decision_sync_submitted_at).to_not be_nil
+        # delay in processing should be 1 day for rating, immediatly for nonrating
+        expect(request_issues.first.reload.decision_sync_submitted_at).to eq(Time.now + 1.day)
```
pkarman: should this be `Time.zone.now` instead?

PR #9115, opened 2019-01-31
Add 'All cases' tab for VSOs

lowellrex:
Connects #7905.

Adds an all cases tab to VSO organization queues which displays all appeals that have been active at the Board in the past 2 weeks which the VSO represents.

Organizational queue landing page | All cases view
--- | ---
image | image

### DIFF
--- a/client/COPY.json
+++ b/client/COPY.json
@@ -127,7 +127,7 @@
   "ORGANIZATIONAL_QUEUE_EMPTY_STATE_MESSAGE": "This queue is empty. You can ",
 
   "ALL_CASES_QUEUE_TABLE_TAB_TITLE": "All cases",
-  "ALL_CASES_QUEUE_TABLE_TAB_DESCRIPTION": "All AMA cases where the appellant is represented by %s",
+  "ALL_CASES_QUEUE_TABLE_TAB_DESCRIPTION": "All AMA cases where the appellant is represented by %s that have been active at the Board of Veterans' Appeals any time in the past 2 weeks",
 
   "JUDGE_CASE_REVIEW_TABLE_TITLE": "Review %s Cases",
   "JUDGE_ASSIGN_DROPDOWN_LINK_LABEL": "Assign Cases",
--- a/client/COPY.json
+++ b/client/COPY.json
@@ -126,6 +126,9 @@
   "ORGANIZATIONAL_QUEUE_PAGE_ASSIGNED_TASKS_DESCRIPTION": "Cases assigned to a member of the %s team.",
   "ORGANIZATIONAL_QUEUE_EMPTY_STATE_MESSAGE": "This queue is empty. You can ",
 
+  "ALL_CASES_QUEUE_TABLE_TAB_TITLE": "All cases",
+  "ALL_CASES_QUEUE_TABLE_TAB_DESCRIPTION": "All AMA cases where the appellant is represented by %s",
+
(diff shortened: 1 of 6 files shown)


client/app/queue/OrganizationQueue.jsx
```diff
@@ -77,27 +99,25 @@ class OrganizationQueue extends React.PureComponent {
...
 
   return {
     success,
     organizationName: state.ui.activeOrganization.name,
+    organizationIsVso: state.ui.activeOrganization.isVso,
     organizations: state.ui.organizations,
     unassignedTasks: getUnassignedOrganizationalTasks(state),
     assignedTasks: getAssignedOrganizationalTasks(state),
     completedTasks: getCompletedOrganizationalTasks(state),
-    tasks: tasksByOrganization(state)
```
lowellrex: This variable wasn't used by component.

client/COPY.json
```diff
@@ -126,6 +126,9 @@
   "ORGANIZATIONAL_QUEUE_PAGE_ASSIGNED_TASKS_DESCRIPTION": "Cases assigned to a member of the %s team.",
   "ORGANIZATIONAL_QUEUE_EMPTY_STATE_MESSAGE": "This queue is empty. You can ",
 
+  "ALL_CASES_QUEUE_TABLE_TAB_TITLE": "All cases",
+  "ALL_CASES_QUEUE_TABLE_TAB_DESCRIPTION": "All AMA cases where the appellant is represented by %s that have been active at the Board of Veterans' Appeals any time in the past 2 weeks",
```
joeyyang: I'd like to propose shortening this copy a bit: 

"All AMA cases represented by %s that have been active at the Board of Veterans' Appeals in the past 2 weeks."

client/app/queue/OrganizationQueue.jsx
```diff
@@ -69,6 +68,29 @@ class OrganizationQueue extends React.PureComponent {
       }
     ];
 
+    // Focus on the first tab in the list of tabs unless we have an "all cases" view, in which case the first tab will
+    // be the "all cases" tab. In that case focus on the second tab which will be the first tab with workable tasks.
+    let focusedTab = 0;
+
+    if (this.props.organizationIsVso) {
```
joeyyang: I'd consider moving this logic into a subroutine so it's clear from a glance that we're conditionally adding a tab, so it looks something like this:

```javascript
tabs = addAllCasesTabIfOrganzationIsVSO(tabs);

const addAllCasesTabIfOrganizationIsVSO(tabs) => {
  if (this.props.organizationIsVSO) {
    tabs.unshift({
       ...
```

joeyyang: perhaps 
```javascript
<TabWindow
  ...
  defaultPage={this.props.organizationIsVSO ? 1 : 0}
/>
```

lowellrex: Hmmm. One nice feature of the current logic flow is that those two things are coupled. If the organization is a VSO then we add the tab and we change the focused tab. I fear decoupling that leads to fragility in the future.

I'm also biased toward procedural code over abstraction so maybe that's just this bias showing through. Happy to adjust to this approach if you'd prefer.

joeyyang: ok! i think the way you have it is fine then!

PR #9118, opened 2019-02-01
8439 catch promise error

### LINKED ISSUE #8439: [Sentry] Do not send this exception to sentry - Error app/queue/uiReducer/uiActions in ?
We need to catch this 400 error properly on the FE and not send it to Sentry.

https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3144/

image

youngfreezyVA:
Resolves #8439 
cc: @lomaxap 
### Description
catch unhandled promise errors in the `requestSave` method

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. open up two tabs and make your way to this url: http://localhost:3000/queue/appeals/3626186/tasks/3626186/draft_decision/dispositions/edit/2
2. Try to edit the same disposition that you've already deleted in another tab
3. you should see an error message appear
4. it should not be sent to sentry any longer since it is caught. (is there any way to verify that it's not sent to Sentry, before this PR is merged?).


client/app/queue/uiReducer/uiActions.js
```diff
@@ -67,19 +67,21 @@ const saveSuccess = (message: UiStateMessage, response: Object) => (dispatch: Di
...
-      detail: 'There was an error processing your request. ' +
-        'Please retry your action and contact support if errors persist.'
-    }]
-  };
+  let errorResponseObject;
 
   try {
-    responseObject = JSON.parse(response.text);
-  } catch (ex) { /* pass */ }
+    errorResponseObject = JSON.parse(response.text);
```
mdbenjam: Interesting, why would there be an error here?

client/app/queue/uiReducer/uiActions.js
```diff
@@ -90,12 +92,15 @@ export const requestSave = (
...
-  return ApiUtil[verb](url, params).then(
-    (resp) => dispatch(saveSuccess(successMessage, resp)),
-    (resp) => dispatch(saveFailure(resp))
-  );
+  return ApiUtil[verb](url, params).
+    then(
+      (resp) => dispatch(saveSuccess(successMessage, resp)),
+
+    ).
+    catch((err) => dispatch(saveFailure(err)));
```
mdbenjam: Does moving the error handler from the second argument of the `then` args to a catch statement change the functionality here? Does it also catch errors in the `saveSuccess` handler?

youngfreezyVA: yup! the catch block will catch either error (in the `.then` block and also in the `saveSuccess` callback). 
https://bytearcher.com/articles/using-promise-then-callback-callback-misses-errors/
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch

client/app/queue/uiReducer/uiActions.js
youngfreezyVA: in the previous implementation (and even in my updated one) the response is really an error response, that's how callback parameters in `catch` callbacks are usually named, ie:

```
.catch((err) => {console.log(err)}
```
I simply updated the naming of that variable to make the code more readable, it was already trying to parse an object which contained data about the error since this method (`saveFailure`) was already in the second/failure callback. after your response, I realized I could make the code even clearer so just pushed some changes to change the names of the parameters

Issue #9139, opened 2019-02-01
Create Docket class with distribution methods


amprokop:
This class will be later subclassed to create hearing, evidence submission, and direct_review dockets. 

It will have methods to achieve the following:

- Return appeals from a docket that are priority (AOD), nonpriority, or either
- Return appeals from a docket that are ready for distribution, or ones that are not ready
- Return those appeals ordered in the order they should be distributed
- Count the # of appeals in a particular state
- Return the N oldest priority appeals and when they became ready
- Distribute appeals to judges

PR #9140, opened 2019-02-01
Adds docket and appeal methods for distribution

amprokop:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/9139


app/models/docket.rb
```diff
@@ -0,0 +1,85 @@
...
+
+    if priority == true
+      scope.merge(Appeal.ordered_by_distribution_ready_date)
+    else
+      scope.order("receipt_date")
+    end
+  end
+
+  def count(priority: nil, ready: nil)
+    appeals(priority: priority, ready: ready).length
```
cmgiven: Does `length` need to instantiate all of the appeals? Or does it smartly count in the database?

amprokop: Good point, it instantiates all the appeals — but `count` does not behave here like i expected, it returns a hash and not a number. I'll look into it.

amprokop: ah, this is why, I believe — https://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-count

amprokop: Sigh. I think `appeals(priority: priority, ready: ready).count.length` would actually return what we want here. `appeals(priority: priority, ready: ready).count` returns
```
{[65, 65]=>2,
 [66, 66]=>2,
 [67, 67]=>2,
 [68, 68]=>2,
 [69, 69]=>2,
 [70, 70]=>2,
 [71, 71]=>2,
 [72, 72]=>2,
 [73, 73]=>2,
 [74, 74]=>2,
 [75, 75]=>2,
 [76, 76]=>2,
 [77, 77]=>2,
 [78, 78]=>2,
 [79, 79]=>2,
 [80, 80]=>2,
 [81, 81]=>2,
 [82, 82]=>2,
 [83, 83]=>2,
 [84, 84]=>2}
```

app/models/appeal.rb
```diff
@@ -44,7 +44,21 @@ class Appeal < DecisionReview
   scope :ready_for_distribution, lambda {
     joins(:tasks)
       .group("appeals.id")
-      .having("count(case when tasks.type = ? and tasks.status = ? then 1 end) = ?", "DistributionTask", "assigned", 1)
+      .having("count(case when tasks.type = ? and tasks.status = ? then 1 end) = ?",
```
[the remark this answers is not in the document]
amprokop: Yeah, I was looking into that. You're right, I think it's harmless and safer to add the `>=1`here.

app/models/docket.rb
```diff
@@ -0,0 +1,89 @@
+class Docket
+  include ActiveModel::Model
+
+  def docket_type
+    fail Caseflow::Error::MustImplementInSubclass
+  end
+
+  def appeals(priority: nil, ready: nil)
+    fail "'ready for distribution' value cannot be false" if ready == false
```
aroltsch: Why are you passing this as a parameter if it can only be `true`

app/models/docket.rb
```diff
@@ -0,0 +1,89 @@
...
+
+    scope = docket_appeals
+    scope = scope.merge(Appeal.ready_for_distribution) if ready == true
+    scope = scope.merge(Appeal.all_priority) if priority == true
```
aroltsch: You could perhaps combine lines 11-20 into:

```
if priority == true
   scope = scope.merge(Appeal.all_priority)
   scope.merge(Appeal.ordered_by_distribution_ready_date)
   return scope
end

scope = scope.merge(Appeal.all_nonpriority)
scope.order("receipt_date")
```

app/models/docket.rb
cmgiven: `ready` can also be `nil`. That's the default.

amprokop: it can be `true` or `nil`

app/models/docket.rb
amprokop: yeah that sounds nice, i can do that

PR #9141, opened 2019-02-01
Do not display available actions if appeal is not assigned to the current user

### LINKED ISSUE #8946: Disable checkout actions for Legacy appeals
Per this slack thread these actions also show for active appeals that haven't been distributed for decision yet. The actions are likely on for all Case details pages because of users' roles, regardless of the step in the appeals process, and because these are Legacy appeals.

Tagging @aroltsch.

What happens in DAS and VACOLS when a judge or attorney user tries to "check out" a legacy appeal that isn't assigned to them, or that isn't active?
What's your initial impression of the lift to remove the actions dropdown for legacy appeals outside of the judge-attorney-judge loop? Wondering if there's a quick fix, and future (or more ideal) fix when we're working on transferring Legacy tasks to Caseflow tasks.

aroltsch:
Resolves #8946 

Do not display actions dropdown if a legacy appeal is not assigned to a current user.


app/models/legacy_tasks/legacy_task.rb
```diff
@@ -11,7 +11,7 @@ class LegacyTask
   TASK_ID_REGEX = /\A[0-9A-Z]+-[0-9]{4}-[0-9]{2}-[0-9]{2}\Z/i.freeze
 
   def available_actions_unwrapper(user, role)
-    available_actions(role).map { |action| build_action_hash(action, user) }
+    available_actions(user, role).map { |action| build_action_hash(action, user) }
```
lowellrex: Can we just pass the `user` here and get the role from the `User` object?

lowellrex:
Cool! Do we want to prevent tasks from even showing up in the first place on the case details for legacy appeals when attorneys and judges are viewing them?

aroltsch:
I think we still want to return tasks information even though a current user is not assigned to them.

app/models/legacy_tasks/legacy_task.rb
aroltsch: I think there was a reason why `role` was passed to the `available_actions_unwrapper`. I don't think this `role` is necessarily coming from VACOLS. I think it is coming from the parameter.

lowellrex: You're definitely right that the role comes from the input parameter, but the input parameter itself [comes from VACOLS](https://github.com/department-of-veterans-affairs/caseflow/blob/a9ea885837d7b548046c0b33cb272399ff05f7b0/app/views/queue/index.html.erb#L5) so I think we can rely on the user object here (and we can further pave the way for getting rid of the role on the front-end entirely!).

aroltsch: I'd prefer not to add additional VACOLS query to pull the role from there. Also I think eventually we will rely on organizations to get user's role?

PR #9143, opened 2019-02-01
[WIP] Do not format end_product_last_action_date when creating decision issue

leikkisa:
connects #8803 

Having trouble reproducing the error locally

I think we should try it again to confirm it's not working.


spec/models/request_issue_spec.rb
```diff
@@ -972,9 +972,10 @@
                 decision_review_type: "HigherLevelReview",
                 decision_review_id: review.id,
                 benefit_type: "compensation",
-                end_product_last_action_date: end_product_establishment.result.last_action_date.to_date
```
leikkisa: I tried this with and without this in the tests and am getting the same results, but don't know if this could be related.

leikkisa: I saw a blog post from someone else saying that this caused a similar problem for them so this is probably it.

leikkisa:
This was due to a UAT problem, and is not really an issue. Closing.

Issue #7638, opened 2018-10-31
Editing intake issue | Adding a request issue with previously canceled ep will create new ep


jiahuang:
Acceptance criteria:
- Adding a new rating/nonrating request issue on the same review should result in the issue being added to the new rating/nonrating ep, not the previously canceled one
- any logic depending on selecting the correct rating/nonrating ep should take into account "canceled" status, and not use any eps that are canceled
- any logic depending on showing the right eps for a review should not show eps with canceled status
- tests should be updated to take into account canceled ep status

pkarman:
One way to test:

* create an intake with a single rating issue
* save
* edit that same intake
* add a nonrating issue
* remove the rating issue
* save
* edit that same intake again
* add a different rating issue (preferably from the same rating but probably works w/ any rating)
* save
* verify that only 2 end products are pending, and 1 is canceled (can you do this in VBMS?)

steve-gov:
Validation steps:
1. Created HLR review for rating issue (decision date 07/29/2018) on `123458314`
2. This created EP 035 in VBMS. Clicked `Edit in Caseflow`
3. Added nonrating issue with `other` category and removed rating issue.

Result:
1. EP 035 canceled
2. EP 036 HLR-NR created with one contention.
--------
<img width="558" alt="screen shot 2019-02-06 at 3 54 25 pm" src="https://user-images.githubusercontent.com/25468689/52381846-e6748d80-2a27-11e9-9e4c-ccab420ca8c5.png">
--------
<img width="769" alt="screen shot 2019-02-06 at 3 55 14 pm" src="https://user-images.githubusercontent.com/25468689/52381848-eaa0ab00-2a27-11e9-863f-e0609598eaac.png">
--------

Further validation:
1. On EP 036 in VBMS, clicked `Edit in Caseflow`
2. Added PTSD rating issue, saved 
3. Got "There are more issues now" modal, confirmed
4. EP 037 HLR-R created with one contention.
--------
<img width="436" alt="screen shot 2019-02-06 at 4 01 32 pm" src="https://user-images.githubusercontent.com/25468689/52382142-0062a000-2a29-11e9-8e69-69db2b08c12b.png">
--------
<img width="745" alt="screen shot 2019-02-06 at 4 02 03 pm" src="https://user-images.githubusercontent.com/25468689/52382150-06f11780-2a29-11e9-9710-815f74708740.png">
--------
Result: Issues were added to new EPs, not previously canceled ones.

PR #9145, opened 2019-02-01
Canceled EPEs are not re-used

pkarman:
connects #7638 

### Description
Skip EPEs that are canceled.


app/models/claim_review.rb
```diff
@@ -193,9 +193,10 @@ def intake_processed_by
   end
 
   def end_product_establishment_for_issue(issue)
-    end_product_establishments.find_by(
-      code: issue.end_product_code
-    ) || new_end_product_establishment(issue.end_product_code)
+    end_product_establishments
+      .where(code: issue.end_product_code)
+      .where.not(synced_status: EndProduct::INACTIVE_STATUSES).first ||
```
jiahuang: I'm assuming we don't need canceled eps for any logic (except maybe updating the modifier code?), and having these saved is mainly just for audits?

app/models/end_product.rb
```diff
@@ -9,7 +9,7 @@ class EndProduct
     "CAN" => "Canceled"
   }.freeze
 
-  INACTIVE_STATUSES = %w[CAN CLR].freeze
+  INACTIVE_STATUSES = %w[CAN CLR CLD].freeze
```
pkarman: this means "CLOSED"

app/models/request_issues_update.rb
```diff
@@ -27,10 +27,9 @@ def perform!
         after_request_issue_ids: after_issues.map(&:id)
       )
       submit_for_processing!
+      process_job
```
pkarman: moving this inside the transaction prevents a "wedged" state in dev/testing if the `process_job` fails.

lib/fakes/vbms_service.rb
```diff
@@ -144,7 +144,7 @@ def self.generate_end_product_for_claim(veteran_hash:, claim_hash:, claim_id:)
         benefit_claim_id: claim_id,
         claim_receive_date: claim_hash[:date].to_formatted_s(:short_date),
         end_product_type_code: claim_hash[:end_product_modifier],
-        end_product_code: claim_hash[:claim_type_code]
+        claim_type_code: claim_hash[:end_product_code]
```
pkarman: this has been a long standing bug but since tests do not rely on these fakes it went unnoticed.

spec/models/claim_review_spec.rb
```diff
@@ -612,9 +630,11 @@ def epe
     context "when there are more than one end product establishments" do
       let(:issues) { [non_rating_request_issue, rating_request_issue] }
 
-      it "establishes the claim and creates the contetions in VBMS for each one" do
+      it "establishes the claim and creates the contentions in VBMS for each one" do
```
pkarman: just a typo

app/models/claim_review.rb
```diff
@@ -194,7 +194,9 @@ def intake_processed_by
 
   def end_product_establishment_for_issue(issue)
     end_product_establishments.find_by(
-      code: issue.end_product_code
+      "code = ? AND synced_status IS NULL OR synced_status NOT IN (?)",
```
jiahuang: Wait is this doing `(code = ? AND synced_status IS NULL) OR (synced_status NOT IN (?))` or `code = ? AND (synced_status IS NULL OR synced_status NOT IN (?))` If it's the 2nd, isn't that or covered by `synced_status NOT IN (?)`.

Can you put parens around this to make the order of operations more obvious?

pkarman: parens added.

I thought the same as you, that `foo NOT IN (?)` would include records where foo is `nil` but apparently that does not work. See also https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/end_product_establishment.rb#L47

PR #9146, opened 2019-02-02
adding prior payee code for appeal remand

jiahuang:
Connects #8676

When there is an appeal dta, attempt to:
- use the latest prior ep's payee code that had the same claimant on the same veteran
- if ☝️ is not found, put both the appeal and the sc in an error state


leikkisa:
I don't think we should default to "00", I think adding the errors like you did is better

PR #9147, opened 2019-02-02
Validate document ID when submitting decision

### LINKED ISSUE #8659: Document ID should be validated on Submit Decision View for both legacy and ama apeals
Currently, when an attorney is ready to submit an AMA or legacy decision, the Document ID field is not validated. Draft Decisions must have a specific format, but an attorney can enter the wrong format and neither the backend nor the frontend will complain.

monfresh:
**Why**: Document IDs follow different formats depending on the work
product. We first introduced Document ID validation in #8575, but that
only applied to editing an existing Document ID. In this PR, we are
making sure Document IDs are validated during the attorney checkout
flow, when they are first created.

Resolves #8659

Testing Plan:
1. As an attorney, select an AMA case in your queue
2. From the Actions dropdown, select Decision ready for review
3. Complete the rest of the checkout flow until you reach the last page
where you can enter the Document ID
4. Enter an invalid ID, such as "123" and click Continue
Expected Result: You get a helpful error message stating what the
expected format is
5. Enter a valid ID, such as "12345-12345678" and click Continue
Expected Result: the case is successfully saved
5. Go back to your queue and select a Legacy case
6. From the Actions dropdown, select "Medical Request Ready for Review"
7. Select "OMO - VHA" or "OMO - IME"
4. Enter an invalid ID, such as "123" and click Continue
Expected Result: You get a helpful error message stating what the
expected format is
5. Enter a valid ID, such as "V1234567.123" (for VHA) and click Continue
Expected Result: the case is successfully saved

### DIFF
--- a/app/controllers/case_reviews_controller.rb
+++ b/app/controllers/case_reviews_controller.rb
@@ -1,9 +1,4 @@
 class CaseReviewsController < ApplicationController
-  CASE_REVIEW_CLASSES = {
-    AttorneyCaseReview: AttorneyCaseReview,
-    JudgeCaseReview: JudgeCaseReview
-  }.freeze
-
   rescue_from Caseflow::Error::UserRepositoryError do |e|
     handle_non_critical_error("case_reviews", e)
   end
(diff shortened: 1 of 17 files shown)


app/controllers/case_reviews_controller.rb
```diff
@@ -13,16 +8,17 @@ def set_application
...
-
-    record = case_review_class.complete(complete_params)
-    return invalid_record_error(record) unless record.valid?
-
-    create_quality_review_task(record)
-
-    response = { task: record }
-    response[:issues] = record.appeal.issues
-    render json: response
+    result = CompleteCaseReview.new(case_review_class: case_review_class, params: complete_params).call
```
aroltsch: I like these changes!

PR #9165, opened 2019-02-04
Fix width of error message on jobs page

pkarman:
Some browsers do not add a horizontal scroll bar for the whole page when the table is too wide to fit. Since the error message is the variable width value, hardcode its max width and tell it to scroll.

screen shot 2019-02-04 at 9 31 11 am

### DIFF
--- a/client/app/styles/_jobs.scss
+++ b/client/app/styles/_jobs.scss
@@ -1,0 +1,5 @@
+.cf-job-error {
+  max-width: 200px;
+  overflow: auto;
+  display: inline-block;
+}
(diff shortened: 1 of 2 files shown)


pkarman:
@kencheeto good question.

IME the wrap only works if the string can be "broken" according to whitespace, or according to browser rules. c.f. https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

I'd prefer we see what users say about this small change and then consider alternatives based on that.

kencheeto:
Ohhh right i can see the text a few rows above, I thought it was a single line that always scrolled 👍

Issue #9135, opened 2019-02-01
AHL Job | Create ScheduleHearingTasks


sharonwarner:
- For any appeal in location 57 that has requested a CO or a video hearing (see logic in the `appeals_ready_for_hearing_schedule` method in appeal_repository.rb), the AHL job should create a ScheduleHearingTask and move the VACOLS location to 'CASEFLOW'
- The Assign Hearings page logic should be updated to pull in any appeals with a ScheduleHearingTask 
- The Assign Hearings page logic should be updated to not pull in appeals with a ScheduleHearingTask that is on hold

PR #9167, opened 2019-02-04
Speed up hearing endpoint

mdbenjam:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/9135

### Description
This PR does three main things:
1) Creates a new parent HearingTask
2) Moves the creation of schedule hearing tasks to our AHL job.
3) Speeds up our `ready_for_hearing_schedule` endpoint by.
   - Adding includes statements where possible
   - Improving eager loading
   - Offloading to the DB where possible

There are some potential future improvements. Specifically, the tasks serializer is a little slow.
```
  attribute :document_id do
    object.latest_attorney_case_review ? object.latest_attorney_case_review.document_id : nil
  end

  attribute :decision_prepared_by do
    {
      first_name: object.prepared_by_display_name ? object.prepared_by_display_name.first : nil,
      last_name: object.prepared_by_display_name ? object.prepared_by_display_name.last : nil
    }
  end

  attribute :available_actions do
    object.available_actions_unwrapper(@instance_options[:user])
  end
```
These are not necessary for Hearings, so we could create a serializer with them removed. We can also try to speed the first two attributes up by more intelligently bulk loading them.

We'll see how these changes fair in production before exploring that more.


app/models/serializers/work_queue/task_serializer.rb
```diff
@@ -102,10 +102,4 @@ class WorkQueue::TaskSerializer < ActiveModel::Serializer
   attribute :available_actions do
     object.available_actions_unwrapper(@instance_options[:user])
   end
-
-  attribute :task_business_payloads do
```
mdbenjam: Removing this since it's no longer used. No need to spend time returning it.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -17,23 +17,29 @@ def find_or_create_if_eligible(appeal)
...
-          assigned_to: HearingsManagement.singleton
-        )
-      end
-
       # Get all tasks associated with AMA appeals and the regional_office
-      appeal_tasks = ScheduleHearingTask.where(
-        appeal_type: Appeal.name,
-        status: Constants.TASK_STATUSES.assigned.to_sym
-      ).joins("INNER JOIN appeals ON appeals.id = appeal_id")
+      incomplete_tasks = ScheduleHearingTask.where(
```
mdbenjam: Get all tasks that are in progress or assigned.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -17,23 +17,29 @@ def find_or_create_if_eligible(appeal)
...
+        "status = ? OR status = ?",
+        Constants.TASK_STATUSES.assigned.to_sym,
+        Constants.TASK_STATUSES.in_progress.to_sym
+      ).includes(:assigned_to, :assigned_by, :appeal, attorney_case_reviews: [:attorney])
+      appeal_tasks = incomplete_tasks.joins("INNER JOIN appeals ON appeals.id = appeal_id")
```
mdbenjam: Get all AMA appeals where the Veteran's closest regional office is the provided RO.

app/services/appeal_repository.rb
```diff
@@ -20,19 +20,28 @@ def self.eager_load_legacy_appeals_for_tasks(tasks)
     legacy_appeal_ids = tasks.select { |t| t.appeal.is_a?(LegacyAppeal) }.map(&:appeal).pluck(:vacols_id)
 
     # Load the VACOLS case records associated with legacy tasks into memory in a single batch.
-    cases = vacols_records_for_appeals(legacy_appeal_ids) || []
+    cases = (vacols_records_for_appeals(legacy_appeal_ids) || []).group_by(&:id)
```
mdbenjam: Change this to a hash with keys that are vacols ids. This allows us to avoid a nested loop below. i.e. change runtime from n^2 -> n.

config/environments/test.rb
```diff
@@ -3,6 +3,7 @@
   config.after_initialize do
     Bullet.enable        = true
     Bullet.bullet_logger = true
+    Bullet.rails_logger  = true
```
mdbenjam: Make it clear when there are better ways for us to load data in our tests.

spec/controllers/tasks_controller_spec.rb
```diff
@@ -762,5 +764,36 @@
         expect(data.first["attributes"]["veteran_available_hearing_locations"]).to eq(nil)
       end
     end
+
+    context "stress test for endpoint" do
```
mdbenjam: This is what I used to test our endpoint's runtime. I'm thinking of deleting it, but we could have a test that looks at the amount of time the endpoint takes and fail if it takes too long... but that seems like it'll be flakey.

spec/models/tasks/schedule_hearing_task_spec.rb
```diff
@@ -100,11 +100,23 @@
...
+          )
+        end
+      end
+
       let!(:non_hearing_cases) do
         create_list(:case, number_of_cases)
       end
 
       it "returns tasks for all relevant appeals in location 57" do
+        AppealRepository.create_schedule_hearing_tasks
```
mdbenjam: This test now needs us to have veterans who have the closest RO listed, so that they'll have tasks created for them.

app/services/appeal_repository.rb
```diff
@@ -20,19 +20,28 @@ def self.eager_load_legacy_appeals_for_tasks(tasks)
...
+    aod = VACOLS::Case.aod(legacy_appeal_ids)
```
mdbenjam: Eager loading AOD saves us a significant amount of time when we serialize legacy appeals.

app/models/tasks/schedule_hearing_task.rb
sharonwarner: Want to confirm: does this exclude tasks that have HearingAdminTasks as children?

app/services/appeal_repository.rb
```diff
@@ -20,19 +20,28 @@ def self.eager_load_legacy_appeals_for_tasks(tasks)
...
+      case_record = cases[t.appeal.vacols_id.to_s].first
+      set_vacols_values(appeal: t.appeal, case_record: case_record) if case_record
+      t.appeal.aod = aod[t.appeal.vacols_id.to_s]
     end
   end
 
   def self.find_case_record(id)
-    VACOLS::Case.includes(:folder, :correspondent, :representatives, :case_issues).find(id)
+    # Oracle cannot load more than 1000 records at a time
+    if id.is_a?(Array)
```
sharonwarner: Ah yup! We made that mistake when we ran the RO algorithm forever ago!

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -17,23 +17,29 @@ def find_or_create_if_eligible(appeal)
...
+      # LegacyAppeal.veteran_file_number_from_bfcorlid function. It is written here in SQL.
+      legacy_appeal_tasks = incomplete_tasks
+        .joins("INNER JOIN legacy_appeals ON legacy_appeals.id = appeal_id")
+        .joins("INNER JOIN veterans ON
+                  CASE
+                    WHEN substring(legacy_appeals.vbms_id from length(legacy_appeals.vbms_id) for 1) = 'C'
+                      THEN lpad(regexp_replace(legacy_appeals.vbms_id, '[^0-9]', ''), 8, '0')
+                    ELSE regexp_replace(legacy_appeals.vbms_id, '[^0-9]', '')
+                  END = veterans.file_number")
+        .where("veterans.closest_regional_office = ?", regional_office)
```
sharonwarner: Previously our logic for this page sorted the appeals such that CAVC appeals were first, then AOD appeals, then the remaining appeals all in docket order. Is there a way to add that sorting on the frontend?

app/services/appeal_repository.rb
```diff
@@ -301,20 +310,54 @@ def self.update_vacols_after_dispatch!(appeal:, vacols_note:)
...
+
   def self.appeals_ready_for_hearing_schedule(regional_office)
     if regional_office == HearingDay::REQUEST_TYPES[:central]
       return appeals_ready_for_co_hearing_schedule
     end
 
     cavc_cases = VACOLS::Case.joins(:folder)
       .where(bfregoff: regional_office, bfcurloc: "57", bfac: "7", bfdocind: "V", bfhr: "2")
-      .order("folder.tinum").limit(100).includes(:correspondent, :case_issues, folder: [:outcoder])
+      .order("folder.tinum").includes(:correspondent, :case_issues, folder: [:outcoder])
```
sharonwarner: I don't think this logic is actually used anywhere anymore!

app/services/appeal_repository.rb
```diff
@@ -301,20 +310,54 @@ def self.update_vacols_after_dispatch!(appeal:, vacols_note:)
     end
   end
 
+  def self.cases_that_need_hearings
+    VACOLS::Case.joins(:folder)
+      .where(bfhr: %w[1 2], bfcurloc: "57").order("folder.tinum")
```
sharonwarner: I don't think this is sufficient... I believe this would also pull in appeals that requested travel board. If bfhr is 2, bfdocind needs to be "V".

mdbenjam: Ah cool, I meant to call this out to you, cause I don't really know what we're looking for :). I'll update it!

app/models/tasks/schedule_hearing_task.rb
mdbenjam: ahhhh great call. Nooo more sql to write 😬

app/models/tasks/schedule_hearing_task.rb
mdbenjam: Frontend can do this easily, in fact I think that's the default sort in Queue. Not sure about HS. If we return everything though we're 👌 since we can actually do that sorting on the frontend.

app/models/tasks/schedule_hearing_task.rb
mdbenjam: Oh wait this will work! Those tasks will be on hold. And this looks to ensure the tasks are either assigned or in progress.

app/services/appeal_repository.rb
sharonwarner: There are probably appeals in location 57 that have already had hearings scheduled through VACOLS that just haven't been moved out yet. Could we restrict this task creation to appeals that don't have a scheduled hearing with no disposition?

app/services/appeal_repository.rb
```diff
@@ -301,47 +310,41 @@ def self.update_vacols_after_dispatch!(appeal:, vacols_note:)
...
-    other_cases = VACOLS::Case.joins(:folder)
-      .where(bfregoff: regional_office, bfhr: "2", bfcurloc: "57", bfdocind: "V")
-      .order("folder.tinum").limit(100).includes(:correspondent, :case_issues, folder: [:outcoder])
-
-    aod_vacols_ids = aod_cases.pluck(:bfkey)
+  def self.cases_that_need_hearings
+    VACOLS::Case.where(bfhr: "1", bfcurloc: "57").or(VACOLS::Case.where(bfhr: "2", bfdocind: "V", bfcurloc: "57"))
+      .joins(:folder).order("folder.tinum")
+      .includes(:correspondent, :case_issues, folder: [:outcoder])
+  end
```
sharonwarner: This should also filter out appeals that already have open hearings.

app/models/tasks/schedule_hearing_task.rb
sharonwarner: Can we include this in this PR? If not, do you mind creating a ticket to fix this?

app/models/tasks/schedule_hearing_task.rb
sharonwarner: Want to confirm, this works for central office as well?

mdbenjam: Good call, talked to @lomaxap and he gave me some pointers on how to fix it.

app/services/appeal_repository.rb
mdbenjam: Sorry I missed some of this the other day. I'll restrict it to hearings with no disposition.

app/models/tasks/root_task.rb
```diff
@@ -84,9 +84,15 @@ def create_distribution_task!(appeal, parent)
     end
 
     def create_hearing_schedule_task!(appeal, parent)
+      hearing_task = HearingTask.create!(
```
mdbenjam: Create a new HearingTask

app/models/tasks/schedule_hearing_task.rb
mdbenjam: Updated the frontend

app/services/appeal_repository.rb
mdbenjam: Should be good now. Can you check my logic?

client/app/hearingSchedule/components/AssignHearingsTabs.jsx
```diff
@@ -177,7 +178,17 @@ export default class AssignHearingsTabs extends React.Component {
       return filteredBy === appeal.attributes.veteranAvailableHearingLocations[0].facilityId;
     });
 
-    return _.map(filtered, (appeal, index) => ({
+    const sortedByAodCavc = _.sortBy(filtered, (appeal) => {
```
mdbenjam: Sort appeals on the frontend

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -71,8 +71,13 @@ def create_available_locations_for_veteran(veteran, va_dot_gov_address:)
     end
   end
 
+  def create_schedule_hearing_tasks
+    AppealRepository.create_schedule_hearing_tasks
```
lomaxap: realizing the `veteran_ids_from_tasks` should not have a `QUERY_LIMIT` once all these tasks are created

Issue #7254, opened 2018-10-03
TaskEvents: A proposal for keeping track of activity on an appeal


lowellrex:
We currently use `Task`s to describe responsibility for completing the various actions it takes to process an appeal through the Board. This is proving to be a good model for meeting that objective, but less than ideal at the job of describing when they were taken and precisely what those actions were. I propose we adopt an additional model whose explicit purpose is tracking activity that has happened on tasks assigned to an appeal.

The proposed model (tentatively dubbed `TaskEvent`s) aspires to improve on three weaknesses with the pure `Task` model:
* *Diary*: Recording actions taken on an appeal that do not result in a change to a task's state.
* *Messaging*: Communicating from the assignee to the assigner why a task was marked as complete.
* *Multiple holds*: Saving information about when/how long holds were put on tasks that had multiple holds.

## What does it look like?

```
create table task_events (
  id          serial,
  task_id     integer,
  time        timestamp,
  description text
);
```

### Diary

One pattern our co-located pilots are employing is placing short term holds on tasks when waiting for email responses, adding a note explaining why the task is on hold when they do so. Currently, this appends the note to the `Task`'s `instructions` field without any information about when the note was made or who made the note. By storing these notes as `TaskEvent`s, we would know when they happened and could infer who created them by looking at who the task was assigned to (or we could add a `created_by` column to `TaskEvent`). This would also remove the confusion caused by us overloading the `instructions` column to include assigner instructions and assignee notes.

### Messaging

Currently there is no way for a `Task`'s assignee to include a message to the assigner when the assignee completes the `Task`. We could offer the assignee the option to add a note (as a `TaskEvent`) to the `Task` they are completing. We could then display that in the Case Timeline and rely on the assigner looking in the Case Timeline for that information, or we could pluck the final `TaskEvent` off of the parent `Task`'s child `Task` and show it in the Case Snapshot.

### Multiple holds

If a co-located staffer creates a hold for `Task` that was previously put on hold we overwrite the previous on hold information. I suspect this on hold information may be important when trying to understand the timeline of an appeal if it took an extraordinarily long time to process (say there are three extension requests granted we will only record one in the `Task` table).

## Further extension of the TaskEvent model

[...]

PR #9169, opened 2019-02-04
Add Task.active? method to determine if a task is complete or cancelled

### LINKED ISSUE #8701: Add ability to cancel tasks in DB
Certain tasks in the appeals process are canceled/resolved/closed rather than completed. Two examples of this are:
- If a Veteran withdraws their hearing request, the Schedule Veteran task is closed
- If a Veteran waives their 90-day evidence window, the 90-day Evidence window is closed

Note: this is different than unassigning a task that was assigned to you, making the top level task back the team queue task, but LP wonders if there is any similarity. 

## Acceptance criteria
* [x] Update `closed_at` timestamp when task is cancelled or completed (PR #9312)
* [x] Add method that does not differentiate between a task being completed successfully and being abandoned in some other way using a new `Task.active?` method (PR #9169)
* [x] Update existing code to use the new `Task.active?` method where applicable (PR #9359)
* [ ] Add backend infrastructure to support cancelling of tasks (PR #9373)
* [ ] Update existing code to set the status to `cancelled` instead of `completed` where applicable (PR #9373)
* [x] Drop the `completed_at` column from the `tasks` table (PR #9181)

lowellrex:
Resolves #8701. Part 2 of 6 of the remaining work to do on this ticket. Adds the `active?` method to the `Task` model to determine if a task has been completed or cancelled.

Blocked by changes to add new `cancelled` status in PR #9312.


lowellrex:
> Reading the requirements on #8701, I wonder if it might make sense to have a separate type of task to mark cancellation; a `CancelScheduleHearingTask`, or more generally, a `CancelTask` that has its own `created_at` and `instructions` and an explicit link to the task it's canceling. We could have `active?` check for an associated `CancelTask`. That'd give us a clearer history of the appeal.
> 
> Otherwise I think we'd have to add a `canceled_at` date to all tasks, and maybe a `canceled_note` text field, and then we'd want to modify the timeline/history to display canceled tasks as two separate events.

Chatted about these points over the phone:
* `completed_at` field will become `closed_at` and will be filled whenever a task's status is changed to `cancelled` or `completed`
* We will not add a `canceled_note` field. We prefer to store cancellation information in a different object entirely (perhaps a `TaskNote` object along the lines of [the `TaskEvent` proposal](https://github.com/department-of-veterans-affairs/caseflow/issues/7254)). If we need to add narrative information when we cancel a task before we implement something like a `TaskNote`, we will append that message to the `instructions` field of the `Task` model.

PR #9180, opened 2019-02-04
Add closed_at column to tasks table

lowellrex:
Connects #8701. Adds the `closed_at` column in preparation to replace the `completed_at` column so that the `completed` and `closed` task statuses can set the same timestamp column. Removing the `completed_at` column will follow in a later PR that will be merged at least 1 day after this change is live to allow enough time for the migration.


db/migrate/20190204193215_add_closed_at_task_column.rb
```diff
@@ -0,0 +1,10 @@
+class AddClosedAtTaskColumn < ActiveRecord::Migration[5.1]
+  def up
+    safety_assured do
+      add_column :tasks, :closed_at, :datetime
+      execute "UPDATE tasks SET closed_at=completed_at"
```
aroltsch: Is it not gonna lock the entire table?

lowellrex: I think this will only lock the individual row as it is updated. That's not totally clear in the [postgres documentation](https://www.postgresql.org/docs/9.4/explicit-locking.html), but if we do end up locking the entire table then I don't think it's a big deal because there are only 30,000 rows in that table so I expect the lock will last on the order of tens of milliseconds.

aroltsch: I remember we ran into issues before while locking an entire table so we came to an agreement to use `find_each`. https://apidock.com/rails/ActiveRecord/Batches/ClassMethods/find_each to avoid downtime during deployment. According to this thread https://stackoverflow.com/questions/1113277/how-do-i-do-large-non-blocking-updates-in-postgresql, all changed rows are locked until the entire update is complete. So I am afraid we might be taking a risk here :)

PR #9185, opened 2019-02-04
Hearing Types | Remove hardcoded hearing types

### LINKED ISSUE #8732: Hearing Types | Remove hardcoded hearing types
Throughout both hearing prep and hearing scheduling, we use "C", "V", and "T" to designate hearing types central, video, and travel. We should remove those hardcoded hearing types and use the HearingDay constant, `REQUEST_TYPES`.

Sjones352:
Resolves #8732

### Description
This pr removes hardcoded hearing types  "C", "V", and "T" from Hearing prep and Schedule
### Acceptance Criteria
- [ ] Throughout both hearing prep and hearing scheduling, we use "C", "V", and "T" to designate hearing types central, video, and travel. We should remove those hardcoded hearing types and use the HearingDay constant, REQUEST_TYPES


app/mappers/hearing_mapper.rb
```diff
@@ -61,9 +61,9 @@ def timezone(regional_office_key)
     private
 
     def code_based_on_request_type(type)
-      return "1" if type == :C
-      return "2" if type == :T
-      return "6" if type == :V
+      return "1" if type == HearingDay::REQUEST_TYPES[:central]
```
sharonwarner: I think this should be type.to_s so that we're not comparing symbols to strings. I'm also fine if you want to just leave the existing logic here for now!

Sjones352: Yes, will leave existing logic for now.

PR #9186, opened 2019-02-04
Add withdraw hearing action

### LINKED ISSUE #8635: Cancel Hearing Request (before Scheduled)
We need to add an action, perhaps at the root task of "Schedule Veteran" that allows the user to close the Schedule Veteran task when the user has opted to withdraw their hearing request - for both AMA and Legacy. 

### Acceptance Criteria
- [ ] Verify that in the "Schedule task" action modal there is an additional task called " Withdraw hearing request"
- [ ] Verify that there is a modal that confirms the withdrawal with the following copy:
   - Title: Confirm withdraw hearing
    - Body: Withdrawing this hearing request will remove this veteran from the hearing pool. 
- [ ] Verify that confirming the withdrawal renders a success message and redirects to the hearing schedule table
    - Title: You have successfully withdrawn [Veteran Name]'s hearing request
     - Body: [Veteran Name] has been removed from the hearing schedule and has been sent to the next appropriate location
- [ ] Verify that this closes the hearing schedule task and removes it from the hearing schedule pool (location 57)
- [ ] Verify that it is moved to location 55 if VSO and location 81 if no VSO
- [ ] Verify that for legacy appeals, we update BRIEFF.BFHR = ‘5’ and BRIEFF.BFHA = ‘5’ so the appeal falls off any pending hearing reports, 5 = none

### Mock
_Add Withdraw hearing request as an option in the schedule hearing task_
add admin action task enabled dropdown

_Modal confirm_
confirm hearing withdrawal

_Final cofirmation_
scheduled veteran

mdbenjam:
Resolves #8635

### Description
This PR adds a new cancel task modal. This modal is intended to be very generic, where the backend can pass any copy it wants to the frontend.

The PR also adds logic in ScheduleHearingTask to handle canceled tasks. There are two pieces to canceling a task.
1) We need to complete the task since there currently is no canceled status.
2) For AMA we need to create a `EvidenceSubmissionWindowTask` and for legacy we need to move the location of the appeal based on whether or not it should get an IHP.


client/app/queue/components/CancelTaskModal.jsx
```diff
@@ -0,0 +1,100 @@
...
+      hearingDay
+    } = this.props;
+    const payload = {
+      data: {
+        task: {
+          status: 'canceled'
+        }
+      }
+    };
+    const hearingScheduleLink = taskActionData(this.props).back_to_hearing_schedule ?
```
mdbenjam: Everything about this modal is generic except this. I didn't know how to add a link from information on the backend :/

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -60,15 +60,18 @@ def update_status_if_children_tasks_are_complete
...
-    hearing_location = task_payloads[:values][:hearing_location]
+      hearing_time = task_payloads[:values][:hearing_time]
+      hearing_day_id = task_payloads[:values][:hearing_pkseq]
+      hearing_type = task_payloads[:values][:hearing_type]
+      hearing_location = task_payloads[:values][:hearing_location]
 
-    if params[:status] == Constants.TASK_STATUSES.completed
       slot_new_hearing(hearing_day_id, hearing_type, hearing_time, hearing_location)
+    elsif params[:status] == "canceled"
+      withdraw_hearing
```
sharonwarner: I just realized, it looks like it's possible to get in a state where we've created an EvidenceSubmissionWindowTask but somehow updating the ScheduleHearingTask to 'completed' errors out, so we'd have an open task that shouldn't be worked. I think this pattern exists in other places too. Have we thought about ways to wrap these in transactions? Or is this a case that we could add to a Looker alert and fix when necessary? I can't imagine this would happen frequently.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -60,15 +60,18 @@ def update_status_if_children_tasks_are_complete
...
+      params[:status] = Constants.TASK_STATUSES.completed
```
sharonwarner: Could we add a canceled status? I'm worried about a case where we have a closed ScheduleHearingTask but no scheduled hearing and we're not sure why.

lowellrex: [In progress](https://github.com/department-of-veterans-affairs/caseflow/issues/8701)! The changing of column names in preparation for the new cancelled status was actually the proximate cause for the outage yesterday.

mdbenjam: Yeah as soon as that's ready we can actually set the status as canceled.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -104,8 +108,37 @@ def add_admin_action_data(_user)
...
 
+  def withdraw_hearing_data(_user)
+    {
+      redirect_after: "/queue/appeals/#{appeal.external_id}",
+      modal_title: COPY::WITHDRAW_HEARING_MODAL_TITLE,
+      modal_body: COPY::WITHDRAW_HEARING_MODAL_BODY,
+      message_title: format(COPY::WITHDRAW_HEARING_SUCCESS_MESSAGE_TITLE, appeal.veteran_full_name),
+      message_detail: format(COPY::WITHDRAW_HEARING_SUCCESS_MESSAGE_BODY, appeal.veteran_full_name),
+      back_to_hearing_schedule: true
+    }
```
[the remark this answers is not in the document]
mdbenjam: Yep!

app/models/tasks/schedule_hearing_task.rb
sharonwarner: Ah awesome! Sounds good!

app/models/tasks/schedule_hearing_task.rb
mdbenjam: Actually decided to add it to just this method. And it looks like we have a `multi_transaction` block defined in ApplicationRecord that should do both a postgres and oracle transaction.

PR #9190, opened 2019-02-04
Add job to update TrackVeteranTasks

lowellrex:
Final piece in the puzzle for #7905.

## Remaining work
* [x] Add tests
* [ ] Run job manually once
* [ ] Add trigger to [cronned jobs in appeals-deployment repo](https://github.com/department-of-veterans-affairs/appeals-deployment/blob/dee9ffeb87f941405e1bb76689b6b863d57ee559/ansible/vars/aws-config.yml#L434)


app/models/appeal.rb
```diff
@@ -161,6 +161,40 @@ def active?
     tasks.where(type: RootTask.name).where.not(status: Constants.TASK_STATUSES.completed).any?
   end
 
+  # Optimized version of Appeal.select(&:active?)
+  def self.all_active
```
amprokop: I have an open PR with  a scope to accomplish the same thing: https://github.com/department-of-veterans-affairs/caseflow/pull/9140/files#diff-d38fdc00b17f61e0a6e1bd027f26224cR51

What do you think of that approach?

app/models/appeal.rb
```diff
@@ -161,6 +161,40 @@ def active?
...
+        distinct(appeals.*)
+      FROM appeals
+      INNER JOIN tasks
+        ON appeals.id = tasks.appeal_id
+      WHERE tasks.type = ?
+        AND tasks.status not in (?)
+      ORDER BY appeals.id", RootTask.name, Constants.TASK_STATUSES.completed])
+  end
+
+  def sync_tracking_tasks
```
amprokop: Eventually, should this close and open IHP writing tasks too?

app/models/appeal.rb
```diff
@@ -161,6 +161,40 @@ def active?
...
+    # Check if there are existing tracking tasks.
+    active_tracking_tasks = tasks.where(type: TrackVeteranTask.name)
```
amprokop: could this logic actually live in `TrackVeteranTask` instead of appeal?

app/models/appeal.rb
lowellrex: Exactly!

app/models/appeal.rb
lowellrex: I think this should live in the `Appeal` model (or maybe even as a standalone helper function? But we don't really have the pattern of standalone helpers, so probably not that) because we will eventually add other logic here when we sync with BGS (like the opening new IHP writing tasks, [as you mention](https://github.com/department-of-veterans-affairs/caseflow/pull/9190#discussion_r)) and we need to be able to create new `TrackVeteranTask`s as well as modifying existing `TrackVeteranTask`s.

app/models/appeal.rb
lowellrex: I like it! Making it a scope makes a lot of sense and the SQL statements are functionally equivalent.

amprokop:
Did you add this job in the jobs controller?

PR #9191, opened 2019-02-04
automatically create translation tasks for veterans in Puerto Rico or Philippines

### LINKED ISSUE #8226: Automatically create Translation task after Caseflow Intake
After Caseflow Intake, create Translation tasks assigned to Translation team queue for cases whose claims originated in the San Juan or Manila Regional Office. 

## AC
- `TranslationTask` has been created on the `RootTask`

tomas-nava:
Resolves #8226

### Description
Establishes new `TranslationTask` task type, and automatically assigns a `TranslationTask` to the Translation organization on successful intake if the veteran's address is in Puerto Rico or Philippines.

The code uses the [`VADotGovService`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/services/external_api/va_dot_gov_service.rb) to validate the veteran's address. If the service is unavailable, it fails silently and falls back to the veteran's address. It does not retry.


tomas-nava:
As I mentioned above, I'm not retrying the `VADotGovService` if it fails or times out. This means that it's possible that appeals that should've automatically generated translation tasks will not. Some possible solutions to that problem:

- retry synchronously, like in the [`FetchHearingLocationsForVeteransJob`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/jobs/fetch_hearing_locations_for_veterans_job.rb#L79)
- retry asynchronously, using shoryuken or asyncable
- add a hook to the [`FetchHearingLocationsForVeteransJob`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/jobs/fetch_hearing_locations_for_veterans_job.rb) to create translation tasks if they haven't already been created

app/models/tasks/translation_task.rb
```diff
@@ -0,0 +1,5 @@
+class TranslationTask < GenericTask
+  def self.create_from_root_task(root_task)
+    create!(assigned_to: Translation.singleton, parent_id: root_task.id, appeal: root_task.appeal)
```
lowellrex: We've been creating [`GenericTask`s assigned to the `Translation`](https://github.com/department-of-veterans-affairs/caseflow/blob/117cd81cb787d58d0e2c8ef4a408d0bfcd61bda6/app/models/task.rb#L268) organization for [`ColocatedTask`s with the "translation" action](https://github.com/department-of-veterans-affairs/caseflow/blob/117cd81cb787d58d0e2c8ef4a408d0bfcd61bda6/app/models/tasks/colocated_task.rb#L49). Can we update that method to instead create a `TranslationTask`?

spec/feature/queue/ama_queue_spec.rb
```diff
@@ -276,6 +276,27 @@ def valid_document_id
         expect(page).to have_content("Task assigned to #{other_organization.name}")
         expect(Task.last.instructions.first).to eq(instructions)
       end
+
+      context "A TranslationTask is assigned to the organization" do
+        let(:veteran_first_name) { "Milivoj" }
+        let(:veteran_last_name) { "Veilleux" }
```
lowellrex: 🤔

spec/models/appeal_spec.rb
```diff
@@ -479,25 +479,114 @@
...
+          end
+        end
+      end
+
+      context "the VADotGovService is not responsive" do
+        let(:message) { { "messages" => [{ "key" => "AddressCouldNotBeFound" }] } }
+        let(:error) { Caseflow::Error::VaDotGovServerError.new(code: "500", message: message) }
+
+        before do
+          allow(VADotGovService).to receive(:send_va_dot_gov_request).and_raise(error)
```
lowellrex: Love this test!

lowellrex:
> As I mentioned above, I'm not retrying the `VADotGovService` request if it fails or times out. This means that it's possible that appeals that should've automatically generated translation tasks will not. Some proposed solutions to that problem:
> 
> * retry synchronously, like [in the `FetchHearingLocationsForVeteransJob`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/jobs/fetch_hearing_locations_for_veterans_job.rb#L79)
> * retry asynchronously, using shoryuken or asyncable
> * add a hook to the [`FetchHearingLocationsForVeteransJob`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/jobs/fetch_hearing_locations_for_veterans_job.rb) to create translation tasks if they haven't already been created (this means that the translation task won't be created immediately after intake; instead it'll be created the next time the `FetchHearingLocationsForVeteransJob` runs)

I think the decision to fail silently here is perfect! We are only sending cases to the translation team for Veterans with addresses in the Philippines or Puerto Rico because the translation team has found that to be an accurate heuristic. We expect there to be some false positives in there and some cases with documents that need translation for cases outside that parameter and VLJ support staff should be able to route those instances to the translation team.

spec/feature/queue/ama_queue_spec.rb
tomas-nava: https://www.fakenamegenerator.com/

app/models/tasks/translation_task.rb
tomas-nava: I made that change with [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/9191/commits/60f43fb75414107e0bd45a7203749b2784895b10)

PR #9199, opened 2019-02-05
Add Wait for Click Dropdowns in Spec

lomaxap:


mdbenjam:
FYI one trick you can use to ensure this fixes the problem is to wrap it in `ensure_stable` and push the code up. i.e.

```
ensure_stable do
  it ..., focus: true do
  end
end
```

Will run the test 20 times on CI and you can ensure it passes every time.

PR #9208, opened 2019-02-05
HoldHearingTask creation

### LINKED ISSUE #9134: Task Flow | Hold hearing tasks
When the ScheduleHearingTask is completed, we should ensure the appeal is in the 'CASEFLOW' location and that we create a HoldHearingTask.

sharonwarner:
Resolves #9134 

### Description
This PR creates a HoldHearingTask assigned to the BvaOrganization after a hearing is scheduled.


app/models/tasks/schedule_hearing_task.rb
```diff
@@ -69,19 +69,12 @@ def update_from_params(params, current_user)
 
     if params[:status] == Constants.TASK_STATUSES.completed
       slot_new_hearing(hearing_day_id, hearing_type, hearing_time, hearing_location)
+      HoldHearingTask.create_hold_hearing_task!(appeal, parent)
     end
 
     super(params, current_user)
   end
 
-  def location_based_on_hearing_type(hearing_type)
```
mdbenjam: So one question, what do they use these locations for right now? Is it only used for us to pull things correctly into our daily dockets?

sharonwarner: I don't think they're used for anything. We don't use them in hearing prep, and the hearings branch is ok with us not using them.

PR #9219, opened 2019-02-05
replace hand-coded xpath selectors with table cell selector function

joeyyang:
### Description
This PR replaces the hand-coded `xpath` table cell selectors with a helper function that will be less breakable by table structure changes.


spec/support/feature_helper.rb
```diff
@@ -1,4 +1,11 @@
 module FeatureHelper
+  def find_table_cell(vacols_id, row_header)
+    header = find(:xpath, "//thead/tr/th", text: row_header)
+    header_index = header.path.split("/").last.delete_prefix("th[").delete_suffix("]")
```
joeyyang: is there a less brittle way to pull the index out of this string?
ex: `header.path`:  `"/html/body/div/div/div/div/main/div/div/section/div/div[2]/table/thead/tr/th[2]"`, where I want the last `2`

lowellrex: A regular expression like `/.*\[(\d+)\]$/.match(header.path)[1]` might be a little more robust, but I think this is fine for now.

Issue #9188, opened 2019-02-04
Update data model for closed issues


shanear:
On the add/remove issues screen, we "soft delete" request issues by setting their `decision_review_id` and `decision_review_type` to `nil`. We will have a variety of new use cases for closed issues with no decisions (examples: https://github.com/department-of-veterans-affairs/caseflow/issues/9179, https://github.com/department-of-veterans-affairs/caseflow/issues/8936) which will require changing the data model to support.

We can start with these soft deleted issues, since the functionality is already live, so this will be mostly a refactor.

## Acceptance Criteria
- When a request issue is removed on the Add/Remove issues screen, `decision_review_id` and `decision_review_type` should not be set to `nil`. Instead, `closed_at` should be set to the current time, and `closed_status` should be set to `removed`.
- All other parts of the app should operate the same way, meaning:
  - Closed request issues should NOT be used in calculating whether an issue is ineligible due to being a duplicate on the intake flow.
  - Closed request issues should NOT be correlated to an active contention in VBMS. If an issue is closed, it's contention should be removed if it hasn't already been removed.
  - Closed request issues should NOT receive any new decisions in Caseflow. They should not appear on any of the non-comp decision flows. They should also not appear in any of the appeals decisions flow.
  - Closed request issues should NOT appear on the add/remove issues page (unless otherwise stated in future stories)

## Notes
- Removing the last issue from a decision review is currently prevented in Caseflow, so we won't worry about the case where the last issue on a decision review is closed in this issue. That's out of scope and will be handled in a follow up issue.

## Technical Notes
- `closed_at` and `closed_status` should be added as new columns. The vision for these columns is that they would eventually be the way to identify if an issue is active or not, and would capture closure states that did not involve decisions.
- The next update we would make is to set `closed_at` and `closed_status` (to `decision_made`) when the decision issues are synced successfully, or when the decision is made in Caseflow.

pkarman:
If `closed_status` only ever has a value of `removed` then why have that column? Does `closed_at` communicate both that it was closed and when?

leikkisa:
Maybe it's the difference between withdrawn and deleted?

shanear:
looks good for claim reviews. Didn't show up on the edit page and didn't block new issues. Data looks good: 

```
 #<RequestIssue:0x0000556793e530f8
  id: 1289,
  review_request_type: "SupplementalClaim",
  review_request_id: 381,
  rating_issue_reference_id: "114210",
[...]

PR #9220, opened 2019-02-05
Add RequestIssues closed_at and closed_status

pkarman:
connects #9188 

### Description
Adds 2 new columns to track closed RequestIssues, instead of disassociating them from the `review_request`.

### DIFF
--- a/app/models/decision_review.rb
+++ b/app/models/decision_review.rb
@@ -190,7 +190,7 @@ class DecisionReview < ApplicationRecord
   end
 
   def active_nonrating_request_issues
-    @active_nonrating_request_issues ||= RequestIssue.nonrating.not_deleted
+    @active_nonrating_request_issues ||= RequestIssue.nonrating.not_closed
       .where(veteran_participant_id: veteran.participant_id)
       .where.not(id: request_issues.map(&:id))
       .select(&:status_active?)
--- a/app/models/decision_review.rb
+++ b/app/models/decision_review.rb
@@ -196,6 +196,10 @@ class DecisionReview < ApplicationRecord
       .select(&:status_active?)
   end
 
+  def open_request_issues
+    request_issues.not_closed
+  end
+
   # do not confuse ui_hash with serializer. ui_hash for intake and intakeEdit. serializer for work queue.
   def serializer_class
     ::WorkQueue::DecisionReviewSerializer
@@ -210,10 +214,6 @@ class DecisionReview < ApplicationRecord
   end
 
   private
-
-  def open_request_issues
-    request_issues.not_closed
-  end
 
   def can_contest_rating_issues?
     fail Caseflow::Error::MustImplementInSubclass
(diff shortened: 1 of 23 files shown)


app/models/decision_review.rb
```diff
@@ -190,12 +190,16 @@ def contestable_issues
...
 
   def active_nonrating_request_issues
-    @active_nonrating_request_issues ||= RequestIssue.nonrating.not_deleted
+    @active_nonrating_request_issues ||= RequestIssue.nonrating.not_closed
       .where(veteran_participant_id: veteran.participant_id)
       .where.not(id: request_issues.map(&:id))
       .select(&:status_active?)
   end
 
+  def open_request_issues
```
shanear: hmm... not sure if this is a necessary layer of indirection, since it's just a scope.

If you rename the scope to `.active` its about the same character.

app/models/request_issue.rb
```diff
@@ -131,6 +136,10 @@ def not_deleted
       where.not(review_request_id: nil)
     end
 
+    def not_closed
```
shanear: I think I prefer `active` or `open`. Adding negatives can be more confusing when another word will do.

shanear: This is such an important term, that it might warrant a discussion/slack comment with the rest of the team

db/migrate/20190205195304_add_closed_at_to_request_issues.rb
```diff
@@ -0,0 +1,15 @@
+class AddClosedAtToRequestIssues < ActiveRecord::Migration[5.1]
+  def up
+    safety_assured do
+      add_column :request_issues, :closed_at, :datetime
+      add_column :request_issues, :closed_status, :string
+      now = Time.zone.now
+      execute "UPDATE request_issues SET closed_at = '#{now}', closed_status = 'removed' WHERE review_request_id IS NULL"
```
shanear: nice

app/models/request_issue.rb
pkarman: I prefer `open` too. I was trying to be consistent with `not_deleted` but have changed it.

pkarman:
@jiahuang wrote:
> Is there a way to alias decision_review.request_issues to always use open_request_issues outside of the model? It doesn't seem like we'd want all request issues during normal operation.

I think that gets into the realm of Rails' [default_scope](https://andycroll.com/ruby/dont-use-default-scope/) which I think we should avoid.

I have added the `.open` scope to all the places where I could find it.

The tests that had to change in this PR are demonstrating that closed issues continue not to appear in screens to the user. There were several explicit tests for that already.

app/models/decision_review.rb
pkarman: I like it for parity with `EndProductEstablishment`'s interface. Since in that class the `request_issues` are not an association but a plain method, the naming is the same for the same effect.

pkarman:
Here's an existing example of [testing that a closed request issue does not appear on the screen](https://github.com/department-of-veterans-affairs/caseflow/blob/master/spec/feature/intake/higher_level_review/edit_spec.rb#L996).

PR #9229, opened 2019-02-06
refactor the click_dropdown feature helper method

tomas-nava:
The method now `expect`s the dropdown to be on the page before trying to `find` it, which will make Capybara wait until its preset timeout for the element to appear before raising an error.


spec/support/feature_helper.rb
```diff
@@ -50,6 +44,25 @@ def xstep(title)
...
+    if options[:prompt].present?
+      keyword_args = { text: options[:prompt] }
+    elsif options[:name].present?
+      selector = ".dropdown-#{options[:name]} .Select-control"
+    end
+
+    if options[:wait].present? && options[:wait] > 0
+      keyword_args[:wait] = options[:wait]
+    else
+      expect(container).to have_selector(selector, **keyword_args)
```
lomaxap: is this because the `find` method does not respect the `wait` argument?

tomas-nava: It was because I *thought* that `find` didn't respect `Capybara.default_max_wait_time`, but [I was wrong](https://www.rubydoc.info/github/jnicklas/capybara/Capybara/Node/Finders#find-instance_method).

`find` respects both `default_max_wait_time` and `wait`.

I've removed this expectation with the latest commit, so now this PR doesn't really "stabilize" the method, it just refactors it.

Issue #9086, opened 2019-01-30
Case Search | Bug (?) | EPs in search results show as "Cleared" when they have a status of CLD


steve-gov:
CLR = Cleared
CLD = Closed

Observed: HLR with status `CLD` in VBMS has an EP Status of `Cleared` in the case search results.

Expected: EPs with status `CLD` should show as Closed, and those with status `CLR` should show as Cleared.

## EPs in VBMS:
<img width="737" alt="screen shot 2019-01-30 at 3 14 25 pm" src="https://user-images.githubusercontent.com/25468689/52019356-63d95480-24a2-11e9-98e6-16a1cc6d2fd8.png">

## EPs in Caseflow search:
<img width="947" alt="screen shot 2019-01-30 at 3 14 09 pm" src="https://user-images.githubusercontent.com/25468689/52019357-63d95480-24a2-11e9-8687-45a3a7ac6f37.png">

kencheeto:
add `RW` as ready to work

steve-gov:
`Ready to Work` validated. But weirdly EPs with status CLD are still showing up as `Cleared` in the interface.

### EPs in VBMS
<img width="766" alt="screen shot 2019-02-06 at 3 28 19 pm" src="https://user-images.githubusercontent.com/25468689/52380745-da86cc80-2a23-11e9-90a3-3c78a11db975.png">

### The 033 and 035 are CLD, and should show as Closed
<img width="674" alt="screen shot 2019-02-06 at 3 26 11 pm" src="https://user-images.githubusercontent.com/25468689/52380747-de1a5380-2a23-11e9-99cf-059fab509764.png">

leikkisa:
When I looked up the EPs showing in VBMS, claim IDs: 600145972 and 600145970, they have a synced_status of "CLR". I checked one directly from BGS as well to make sure it didn't go from CLR to CLD or something weird like that with BGSService.new.client.claims.find_claim_detail_by_id(600145972) and it had <statusTypeCode>CLR</statusTypeCode>

So I think this means that VBMS shows CLR as CLD in their UI.

My next question is whether CLD is an actual status_type_code for an End Product, because I think it'd be misleading to add code to handle it if it isn't. There are no instances of EndProductEstablishments in Caseflow UAT or Prod right now where the synced_status is CLD.

leikkisa:
Got confirmation from VBMS that they show Cleared End Products as CLD, so I'm going to remove the changes related to CLD

[...]

PR #9231, opened 2019-02-06
Add closed and ready to work to end product statuses for search page

leikkisa:
connects #9086


leikkisa:
screen shot 2019-02-05 at 5 27 56 pm

leikkisa:
There are specs for known statuses and unknown status for end product, which also surprisingly look pretty much the same as each other :-P. It felt a little silly adding each one to what was there.

PR #9238, opened 2019-02-06
Don't fetch document count if we don't need it

### LINKED ISSUE #8981: Only fetch document count for appeal if we don't have it
Currently, we fetch the document count for an appeal [every time the `AppealDocumentCount` component is loaded](https://github.com/department-of-veterans-affairs/caseflow/blob/eb1928df757dc17c689b9651285af3bbc244aaea/client/app/queue/AppealDocumentCount.jsx#L19) on the front-end. We do this because when we load the component in the queue table view we use the value that we've cached to the Caseflow database and when we load the component on the case details page we request the latest document count VBMS (which we expect to take longer to respond and therefore don't do when we load the queue table view). We can prevent some of these requests if we use the value we've cached in the redux store whenever we load the queue table view and only make the request on the case details page when the most recent value in the redux store does not come from VBMS.

## Acceptance criteria
* [ ] Fetch document count if the document count does not exist in the redux store
* [ ] Fetch document count when visiting the case details page if value in redux store does not come from VBMS
* [ ] Do not fetch document count if value in redux store comes from VBMS

hschallhorn:
Only request document count from BE if
  a. we don't already have it in the state or
  b. the count we have comes from Documents when we want the vbms count or
     vice versa

Resolves #8981 

### Description
Before requesting document count, check to see if we already have it stored and it came from the correct source (caseflow Documents table vs VBMS)

### Acceptance Criteria
- [X] Fetch document count if the document count does not exist in the redux store
- [X]  Fetch document count when visiting a user's queue if value in redux store does not come from caseflow Documents table
- [X]  Fetch document count when visiting the case details page if value in redux store does not come from VBMS
- [X]  Do not fetch document count if value in redux store comes from VBMS

### Testing Plan
1. Go to a queue with reader links and multiple tabs
2. Switch to a different tab and back to the original one
3. Confirm no requests were made to document_count (reader link should not re-render)


lowellrex:
Is the truth table below correct (✅ means that we use the cached value, ❌ means that we make a request to the backend)?

. | input `cached` -> true | input `cached` -> false 
--- | --- | ---
doc count in store, `cached` value in store -> true | ✅ [Use the value from the store](https://github.com/department-of-veterans-affairs/caseflow/pull/9238/files#diff-4487eca0ba96c962b3a94ec83881d9acR58), and [don't make request](https://github.com/department-of-veterans-affairs/caseflow/pull/9238/files#diff-4487eca0ba96c962b3a94ec83881d9acR29) | ❌ Make a request to the backend
doc count in store, `cached` value in store -> false | ❌ [Make a request to the backend](https://github.com/department-of-veterans-affairs/caseflow/pull/9238/files#diff-4487eca0ba96c962b3a94ec83881d9acR29) | ✅ [Use the value from the store](https://github.com/department-of-veterans-affairs/caseflow/pull/9238/files#diff-4487eca0ba96c962b3a94ec83881d9acR58), and [don't make request](https://github.com/department-of-veterans-affairs/caseflow/pull/9238/files#diff-4487eca0ba96c962b3a94ec83881d9acR29)
doc count and `cached` value not in store | ❌ Make the request | ❌ Make the request

hschallhorn:
Yep. Though the last row should be "doc count OR `cached` value not in store"

Though now that I look at it again, I need to handle the request error differently to make sure that doesn't return true

client/app/queue/reducers.js
```diff
@@ -160,15 +160,20 @@ export const workQueueReducer = (state: QueueState = initialState, action: Objec
...
       }
     });
   case ACTIONS.SET_APPEAL_DOC_COUNT:
     return update(state, {
       docCountForAppeal: {
         [action.payload.appealId]: {
-          $set: action.payload.docCount
+          $set: {
+            count: action.payload.docCount,
+            cached: action.payload.cached
```
lowellrex: I think we need to set the `err` attribute to `null` here or else it will stick around after we get one error.

lowellrex: Might be worth splitting that out into a different object in state as well. Something like `docCountErrorForAppeal`

hschallhorn: Hm, I heard from someone $set will replace the whole object.

All jokes aside, I tested this and it works as intended. Feel free to come shoulder surf for proof.

client/app/queue/AppealDocumentCount.jsx
```diff
@@ -38,22 +41,22 @@ class AppealDocumentCount extends React.PureComponent {
...
+    if (_.isNil(this.props.docCountForAppeal) && !this.props.docCountErrorForAppeal) {
       if (this.props.loadingText) {
         return <span {...documentCountStyling}>Loading number of docs...</span>;
       }
 
       return null;
     }
 
-    return this.props.docCountForAppeal;
+    return this.props.docCountErrorForAppeal || this.props.docCountForAppeal.count;
```
lowellrex: I think we need to unset `docCountErrorForAppeal` at some point. Perhaps we should unset everything (error, count, source/cached) for the appeal when we make the request so we have a clean slate when we receive a response from the backend?

client/app/queue/AppealDocumentCount.jsx
```diff
@@ -28,6 +26,11 @@ class AppealDocumentCount extends React.PureComponent {
       return;
     }
 
+    if (this.props.docCountForAppeal && !this.props.docCountErrorForAppeal &&
```
lowellrex: I think we can leave off the `!this.props.docCountErrorForAppeal` since we should not have an error for the document count when we actually have a document count.

client/app/queue/AppealDocumentCount.jsx
hschallhorn: Error is unset here: https://github.com/department-of-veterans-affairs/caseflow/pull/9238/files#diff-a178f0c699aa614d5c8790997889ed2cR180

hschallhorn: But yeah, unsetting everything would work as well. And would make your other comment true as well

client/app/queue/reducers.js
```diff
@@ -168,7 +169,15 @@ export const workQueueReducer = (state: QueueState = initialState, action: Objec
...
+          }
+        }
+      },
+      docCountErrorForAppeal: {
+        [action.payload.appealId]: {
+          $set: null
```
hschallhorn: @lowellrex Is nulling out `docCountErrorForAppeal` here (and doing the same for `docCountForAppeal` in `case ACTIONS.ERROR_ON_RECEIVE_DOCUMENT_COUNT:` alright for resetting them, or should it be done before a new request is made?
Though I feel like I already know the answer.

client/app/queue/AppealDocumentCount.jsx
```diff
@@ -28,6 +26,11 @@ class AppealDocumentCount extends React.PureComponent {
...
+      (cached || this.props.docCountForAppeal.cached === Boolean(cached))) {
```
lowellrex: One final check on my understanding the distinction between the `this.props.cached` and `this.props.docCountForAppeal.cached`. `this.props.cached` is a boolean value that indicates whether we want the server to return the count of documents that has been cached in Caseflow (`true`), or whether we want to make the request to VBMS (the authoritative source on document counts) in which case this value would be `false`. And `this.props.docCountForAppeal.cached` is a boolean that indicates whether the value we've stored in the global state is from the authoritative source (`false`) or from our backend cache (`true`).

Is that correct?

hschallhorn: Yep. We can also rename/add comments to clarify this. 

So this is checking that we already have a count in the state and that the count we have comes from the source the front end wants this request to come from. With the exception being that we are requesting a doc count from Caseflow, but already have the more accurate doc count in the state.

lowellrex:
Writing up some quick notes about the direction we decided to head in our offline chat.

* Do not make the request to the backend if:
    * We have the precise value in the redux store
    * We have the cached value in the redux store and we are requesting the cached value

* Display (this list is effectively a switch statement, stop after we hit the first applicable case)
    * The precise value if we have it
    * The cached value if we have it and we are requesting the cached value
    * Nothing if the request is in flight
    * The error if we returned an error
    * Nothing (I don't think we should ever hit this fall back)

This will necessitate a slight change to how we are storing the data in the redux store to look something like:

```
docCountForAppeal => {
  id1 => {
    precise => Int?,
    cached => Int?,
    error => String?,
    loading => Boolean,
  },
  id2 => ...
```

Correct me if I'm wrong @hschallhorn !

client/test/karma/queue/ColocatedTaskListView-test.js
```diff
@@ -166,6 +167,19 @@ describe('ColocatedTaskListView', () => {
       expect(docketNumber.text()).to.include(appeal.docketNumber);
       expect(daysWaiting.text()).to.equal('1');
       expect(documents.html()).to.include(`/reader/appeal/${task.externalAppealId}/documents`);
+      expect(documents.text()).to.include('Loading number of docs...');
```
hschallhorn: @lowellrex added a few simple tests

client/app/queue/AppealDocumentCount.jsx
```diff
@@ -17,43 +17,57 @@ const documentCountStyling = css({
...
+
     const requestOptions = {
       withCredentials: true,
       timeout: { response: 5 * 60 * 1000 }
     };
 
     const endpoint = `document_count${cached ? '?cached' : ''}`;
 
+    this.props.loadAppealDocCount(this.props.externalId);
+
```
lowellrex: When we attempt to load the document count for the appeal let's reset the `errorFetchingDocumentCount` for that appeal.

hschallhorn: 👍

client/app/queue/AppealDocumentCount.jsx
```diff
@@ -17,43 +17,57 @@ const documentCountStyling = css({
...
+      loadingText
+    } = this.props;
 
-      return null;
+    if (docCountForAppeal) {
+      if (_.isNumber(docCountForAppeal.precise)) {
+        return docCountForAppeal.precise;
+      } else if (cached && _.isNumber(docCountForAppeal.cached)) {
+        return docCountForAppeal.cached;
+      } else if (loadingText && (docCountForAppeal.loading || docCountForAppeal.error)) {
```
lowellrex: I'm not sure I understand the logic in this final condition? Is it possible to separate these two conditions?

As written if `loadingText` is true (and it will always be true, since the only place we call [`AppealDocumentCount` sets it](https://github.com/department-of-veterans-affairs/caseflow/blob/29e7c4cfa0bc4b844a566d84a2168a90f31a99ad/client/app/queue/ReaderLink.jsx#L59)) and either `docCountForAppeal.loading` or `docCountForAppeal.error` evaluates to true then we will display one of the two two values in the return statement. Can we rewrite this as something like:

```jsx
...
  return docCountForAppeal.cached;
} else if (docCountForAppeal.error) {
  return docCountForAppeal.error;
} else if (docCountForAppeal.loading) {
  return <span {...documentCountStyling}>Loading number of docs...</span>;
}
...
```

client/app/queue/AppealDocumentCount.jsx
```diff
@@ -17,43 +17,57 @@ const documentCountStyling = css({
...
+    }
+
     const requestOptions = {
       withCredentials: true,
       timeout: { response: 5 * 60 * 1000 }
     };
 
     const endpoint = `document_count${cached ? '?cached' : ''}`;
 
+    this.props.loadAppealDocCount(this.props.externalId);
```
lowellrex: Do we ever unset the boolean value that indicates the document count for this appeal is loading?

client/app/queue/AppealDocumentCount.jsx
hschallhorn: 🤔 You're right. I was seeing an odd error in testing, but I'll revisit it

client/app/queue/AppealDocumentCount.jsx
hschallhorn: Yep. any time a doc count comes back or we receive an error https://github.com/department-of-veterans-affairs/caseflow/pull/9238/files#diff-a178f0c699aa614d5c8790997889ed2cR179

client/app/queue/AppealDocumentCount.jsx
hschallhorn: [Here](https://github.com/department-of-veterans-affairs/caseflow/blob/29e7c4cfa0bc4b844a566d84a2168a90f31a99ad/client/app/queue/ReaderLink.jsx#L55) is were we have an `AppealDocumentCount` where loading text is not used. This is on the case details page, where the document count appears within the link. If we have an error, "View Failed to Load docs" will be shown

lowellrex: Ahhh! Great catch!

hschallhorn: Do we want to handle this differently or should I back up to the old logic?

PR #9239, opened 2019-02-06
Mark jobs before Oct 1 processed

pkarman:
Should only really affect UAT since the related tables are not in active use in production yet.

Should reduce the noise on the /jobs page.


pkarman:
will do via console

PR #9240, opened 2019-02-06
Hearings Seeds

lomaxap:


db/seeds.rb
```diff
@@ -318,6 +318,139 @@ def create_hearings
...
 
+  def create_ama_hearing(day)
+    vet = Generators::Veteran.build(
+      file_number: Faker::Number.number(9).to_s,
+      first_name: Faker::Name.first_name,
+      last_name: Faker::Name.last_name
+    )
+    vet.save
+
+    app = FactoryBot.create(
```
sharonwarner: Nit: Can we use hearing_day, veteran, and appeal as these variable names? They're more consistent with what we're doing across the codebase.

db/seeds.rb
```diff
@@ -318,6 +318,139 @@ def create_hearings
...
+    )
+  end
+
+  def create_vacols_hearing_day(day, ro_key)
+    master_record = FactoryBot.create(
+      :case_hearing,
+      hearing_date: day.scheduled_for,
+      hearing_type: "C",
+      room: "001",
+      folder_nr: (ro_key == "C") ? "VIDEO #{ro_key}" : nil
```
sharonwarner: Hm, I'm not totally sure about this logic - We don't need to create any CO master records in VACOLS, only video.

PR #9241, opened 2019-02-06
Create pagination components and add them to Queue table

### LINKED ISSUE #8242: Introduce pagination to all queue tables
### Context
Soon, queues will have 100 to 1000s of cases in a queue. This is especially relevant for team queues, like the hearings management branch queue, and when we support both legacy and AMA cases in the same queue. We'd like to introduce pagination to caseflow.

### Principles
- We'd ideally like to retrieve and locally cache on client-side all available cases at once, and conceal them according to the pagination sequencing. (as opposed to retrieving each page freshly). We'll need to test the speed and reliability of this, and @mdbenjam has some ideas that we can pressure test it with hearing schedule data.
- We'd like to build this component once, and reuse it in any table structure across Caseflow. Current needs are for Queue & Hearing Schedule products, but future needs may arise in Intake product.
- We'd like to integrate filtering with pagination, such that when the user has filtered any of the vertical table categories, the pagination should reflect the filtered range, and shall appear truncated until the filters have been cleared.

### User needs
- As a queue user, I need to see 15 cases per page.
- As a queue user, I need to see a case-range and total case count at the top and bottom of the page (Viewing 1 - 15 of 120 total cases) 
- As a queue user, I need to always see the first and last page of the sequence, as well as the previous and next buttons.
- As a queue user, I need to see the active page (the case-range that I'm viewing) as a dark, active square on the pagination component.
- As a queue user, if I'm on page one of the sequence, the previous button should be disabled. Similarly, if I'm on the last page of the sequence, the next button should be disabled.

#### When user is on page 1
pagination-1

#### When user is somewhere in the middle of the page-range
pagination-2

#### When the user is on the last page available
pagination-3

annekainicUSDS:
Resolves #8242

### Description
Adds pagination components, and enables them on Queue.

This is what the new components look like added **below the table**. You can see the styling on the first page that is selected, as well as the `...` button, which is added for any table that has more than 5 pages. The `Previous` button is also disabled.
screen shot 2019-02-06 at 11 27 51 am

The componens are also added **above the table**.
screen shot 2019-02-06 at 4 05 46 pm

This is what it looks like when a **page in the middle of the list** is selected. Both `Previous` and `Next` buttons are enabled.
screen shot 2019-02-06 at 11 28 09 am

This is what it looks like when the **last page** is selected. The `Next` button is disabled.
screen shot 2019-02-06 at 11 28 15 am

When there are too few data points to create more than 1 page, we show the summary of cases seen, but we **hide the pagination buttons**.
screen shot 2019-02-06 at 11 28 27 am

There is also **scrolling** feature added, which is triggered when the buttons are clicked. The page will scroll to an invisible element at the top of the table, and will also set focus to that element, so that the next time the user tabs it will start at the top of the table as opposed to in the bottom set of buttons.

This feature interacts with **filtering** in the following ways:
1. The data in the table is first filtered, and then broken up into pages. The number of pages will change based on the amount of data that is filtered out.
2. Whenever a new filter is applied, the table goes back to the first page of data. This is useful for avoiding bugs when pages that may have been selected no longer exist once the data is filtered down.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
Added unit tests to test new components and functionality.


client/app/components/TablePagination.jsx
```diff
@@ -0,0 +1,147 @@
...
+    } = this.props;
+    const numberOfPages = paginatedData.length;
+
+    // Render the pagination summary
+    // Get previous number of cases so we can calculate the case range in
+    // the pagination summary
+    let previousCaseCount = 0;
+    let i = 0;
+
+    while (i < currentPage) {
```
mdbenjam: nit: why not use a for loop?

client/app/queue/QueueTable.jsx
```diff
@@ -223,6 +240,10 @@ export default class QueueTable extends React.PureComponent {
 
   updateFilteredByList = (newList) => {
     this.setState({ filteredByList: newList });
+
+    // When filters are added or changed, default back to the first page of data
+    // because the number of pages could have changed as data is filtered out.
+    this.updateCurrentPage(0);
```
mdbenjam: 👍

client/app/queue/QueueTable.jsx
```diff
@@ -248,6 +269,32 @@ export default class QueueTable extends React.PureComponent {
     return filteredData;
   };
 
+  paginateData = (tableData) => {
+    const duplicateTableData = tableData.slice(0);
+    const paginatedData = [];
+    const numberOfPages = (duplicateTableData.length / 15) + (duplicateTableData.length % 15 > 0 ? 1 : 0);
```
mdbenjam: This might be clearer as `Math.ceil(duplicateTableData.length / 15)`

client/app/queue/QueueTable.jsx
```diff
@@ -248,6 +269,32 @@ export default class QueueTable extends React.PureComponent {
...
+
+    _.times(numberOfPages, () => {
```
mdbenjam: I'm wondering if this might be a little cleaner if you just splice in bigger chunks:

```
for(let i = 0; i <= tableData.length; i = i + ITEMS_PER_PAGE) {
  paginatedData.push(tableData.slice(i, i + ITEMS_PER_PAGE));
}
```

Untested, but I think it'd be a little more efficient.

client/app/styles/_commons.scss
```diff
@@ -1239,3 +1247,51 @@ button {
     }
   }
 }
+
+// ===========================*
+// Pagination
+// ============================*/
+
+.cf-pagination-summary {
```
mdbenjam: Could we either move this CSS to a pagination specific file? Or inline using glamor? Trying to reduce the amount of CSS we're adding to the commons file.

client/app/queue/QueueTable.jsx
```diff
@@ -27,6 +29,20 @@ import { COLORS } from '../constants/AppConstants';
  *
  * see StyleGuideTables.jsx for usage example.
  */
+const scrollTo = (dest = this, opts) => scrollToComponent(dest, _.defaults(opts, {
```
mdbenjam: This scrollsTo the top of the table?

client/app/queue/QueueTable.jsx
```diff
@@ -27,6 +29,20 @@ import { COLORS } from '../constants/AppConstants';
...
+  align: 'top',
+  duration: 1000,
+  ease: 'outCube',
+  offset: -35
+}));
+
+const focusElement = (el = this) => {
+  if (el.tabIndex <= 0) {
+    el.setAttribute('tabindex', '-1');
```
mdbenjam: Why do we need this?

client/app/queue/QueueTable.jsx
mdbenjam: Also maybe we could turn this 15 into a constant, or maybe even a prop.

client/app/components/TablePagination.jsx
annekainicUSDS: I don't remember why I made this choice... but looking at it now a for loop might be more appropriate since i'm incrementing an index. I can change it.

client/app/queue/QueueTable.jsx
annekainicUSDS: Yes, that is the default. You can pass something else to scroll to if you want, though. But we're using it just to scroll to the top of the element `<QueueTable/>`.

client/app/queue/QueueTable.jsx
annekainicUSDS: Yes, turning it into a prop is probably a good idea so we can change it.

client/app/queue/QueueTable.jsx
annekainicUSDS: I can try it

client/app/styles/_commons.scss
annekainicUSDS: Yes, I wasn't sure what the best place to put new CSS was, there doesn't seem to be a ton of organization around how site-wide CSS is currently divided up. Maybe what's better to do is to create a file for all table-related data?

client/app/queue/QueueTable.jsx
annekainicUSDS: Setting the `tabindex` for the element we add `focus` to is necessary because this element is not naturally included in the tab order (it's not a link or button or selectable element). If we don't add `tabindex` to it, even though we focus on it, if the user is navigating the page using tabbing, once they click tab again after being scrolled to the top of the page, the element they will tab into will be the next element after the previous one that had focus (e.g., if the user clicks on a new page, they are scrolled to the top of the table, but once they click tab again they will have tabbed into the next page button instead of the next element at the top of the table). Hope that makes sense, easier to understand in interacting with the page then explaining it.

client/app/queue/QueueTable.jsx
annekainicUSDS: Changed this!

client/app/styles/_commons.scss
annekainicUSDS: I created a new file for table-only styles. I think in general some reorganization and thought around how to set up the CSS in caseflow would probably be helpful, but that's outside the scope of this PR, so if that's something we want to do let's tackle that separately. Let me know, though, if the separate file for table-only styles is something you're okay with for now.

client/app/queue/QueueTable.jsx
annekainicUSDS: Turned this into a prop.

client/app/components/TablePagination.jsx
annekainicUSDS: Changed this to a for loop.

annekainicUSDS:
@mdbenjam I made the changes you suggested, so let me know if you want to take another look at this!

client/app/queue/QueueTable.jsx
mdbenjam: Do we definitely want to scroll? It's a little weird to be fighting with the auto scroll if I want to scroll as it's scrolling. At the very least could we scroll it a bit faster?

client/app/queue/QueueTable.jsx
mdbenjam: I'm playing around with it, but still not sure I understand. After changing pages hitting tab should do what? Select `prev`? That's what it seems to do for me.

annekainicUSDS: What happens without these lines is that once you click on a new page, you are scrolled to the top, but the focus is still on the last button you clicked, so when you start tabbing through, it tabs you to an element below where you are, which is confusing. I think if the scrolling is taking you to the top of the page, so should the focus. If you pull down the branch and comment out these lines, it will be more clear.

mdbenjam: Ah I see now. Gotcha. Sounds good 👍

client/app/queue/QueueTable.jsx
annekainicUSDS: Scrolling is something @sneha-pai and I decided on, and I think it's a good idea. If we want to remove it, that's really a conversation for Sneha to be involved in. She mentioned that she plans to do user testing once this is merged, so we'll definitely get feedback from users if this is confusing. But I've seen this used effectively in similar situations for apps on va.gov.

mdbenjam: Sounds good!

annekainicUSDS: I shortened the scrolling length slightly, so hopefully this is a bit better!

PR #9245, opened 2019-02-06
Amprokop add target decision date

amprokop:
- Sets `target_decision_date` on appeals when appeals are intook
- Adds appeal method that counts nonpriority decisions in the last year. This will include RAMP decisions, i verified that we started creating `decision_documents` before RAMP started on Oct. 1st.
- Adds class method to provide the rate at which non-priority direct reviews were received in the last 365 days, regardless of their current status. For the first month of AMA, this should return 38,500. As of April 1, 2019, we can look at the receipts between March 1 and yesterday, and multiply up to get the annual rate. As of February 29th, 2020 (365 days after March 1st, 2019, since it's a leap year), we just count the number of receipts in the last year.


app/models/appeal_intake.rb
```diff
@@ -36,6 +36,7 @@ def review!(request_params)
   def complete!(request_params)
     super(request_params) do
       detail.update!(established_at: Time.zone.now)
+      detail.set_target_decision_date!
```
amprokop: I don't see any reason to put this behind a feature flag — do you, @cmgiven?

spec/factories/decision_document.rb
```diff
@@ -1,7 +1,7 @@
 FactoryBot.define do
   factory :decision_document do
     appeal { create(:appeal) }
-    sequence(:citation_number) { |n| "A181#{n % 10}3456" }
+    sequence(:citation_number) { |n| "A181#{(n % 100_000).to_s.rjust(5, '0')}" }
```
amprokop: This previously threw a unique index error anytime you tried to create more than 10 documents. This just kicks the can down the road but i hope 10000 is fine :)

app/models/appeal.rb
```diff
@@ -31,6 +31,7 @@ class Appeal < DecisionReview
       .where("advance_on_docket_motions.granted = ?", true)
       .or(join_aod_motions
         .where("people.date_of_birth <= ?", 75.years.ago))
+      .distinct
```
amprokop: Small bugfix

app/models/appeal_intake.rb
cmgiven: Nope, seems reasonable.

app/models/direct_review_docket.rb
```diff
@@ -1,5 +1,30 @@
...
+      # Hardcode this figure as a baseline since we won't have enough data to judge.
+      return 38_500
+    elsif today < Date.new(2020, 2, 29)
+      # If it's been fewer than 365 days since March 1st, return a weighted number of appeals.
+      march_first = Date.new(2019, 3, 1)
+      days_since_march_first = Integer(today - march_first)
+      appeals_since_march_first =
+        DirectReviewDocket.all_nonpriority.where("receipt_date > ?", march_first).count.length
+      # appeals_since_march_first / adjusted_appeals = days_since_march_first / 365
+      ((appeals_since_march_first * 365) / days_since_march_first).round
```
cmgiven: The `.round` is unnecessary, since it's dividing integers.

app/models/appeal.rb
```diff
@@ -304,6 +309,10 @@ def establish!
     processed!
   end
 
+  def set_target_decision_date!
+    update!(target_decision_date: DirectReviewDocket::DAYS_TO_DECISION_GOAL.days.from_now) if direct_review_docket?
```
cmgiven: Should this maybe be from `receipt_date`? I'm not sure it would matter long-term, but it seems like we'll probably want to use this method to backfill target_decision_dates?

amprokop: Interesting, I defer to you on that, but `receipt_date` feels right to me as well.

cmgiven: Yeah, let's go with `receipt_date`.

Issue #9249, opened 2019-02-06
Intake | End Product establishment created for HLR with no eligible issues


shanear:
We experienced an issue in UAT where an end product was created for an HLR with no eligible request issues.

See the data below for more:

```
#####################################################

HigherLevelReview.where(veteran_file_number: "612341450").last.request_issues

[#<RequestIssue:0x0000562870848b88
  id: 1255,
  review_request_type: "HigherLevelReview",
  review_request_id: 1097,
  rating_issue_reference_id: "112362",
  rating_issue_profile_date: Wed, 30 May 2018 04:00:00 UTC +00:00,
  contention_reference_id: nil,
  description: nil,
  issue_category: nil,
  decision_date: nil,
  disposition: nil,
  end_product_establishment_id: 836,
  removed_at: nil,
  rating_issue_associated_at: nil,
  parent_request_issue_id: nil,
  notes: "disagrees with effective date",
  is_unidentified: nil,
  ineligible_due_to_id: 1241,
  untimely_exemption: nil,
  untimely_exemption_notes: nil,
  ineligible_reason: "duplicate_of_rating_issue_in_active_review",
  ramp_claim_id: nil,
  decision_sync_submitted_at: nil,
  decision_sync_attempted_at: nil,
  decision_sync_processed_at: nil,
  decision_sync_error: nil,
  vacols_id: nil,
  vacols_sequence_id: nil,
  created_at: Wed, 06 Feb 2019 17:00:41 UTC +00:00,
  benefit_type: "compensation",
  contested_decision_issue_id: nil,
  veteran_participant_id: "600228922",
  contested_rating_issue_diagnostic_code: nil,
  decision_review_type: "HigherLevelReview",
  decision_review_id: 1097,
  contested_rating_issue_reference_id: "112362",
  contested_rating_issue_profile_date: "2018-05-30T09:14:34-05:00",
  contested_issue_description:
   "Service connection for Arm is granted with an evaluation of 30 percent effective January 1, 2017.",
  nonrating_issue_description: nil,
  unidentified_issue_text: nil,
  contested_rating_issue_disability_code: "5201",
  last_submitted_at: nil>]


##############################################
HigherLevelReview.where(veteran_file_number: "612341450").last.end_product_establishments

[#<EndProductEstablishment:0x000056286fd404e8
  id: 836,
  established_at: Wed, 06 Feb 2019 17:00:57 UTC +00:00,
  synced_status: "RFD",
[...]

PR #9259, opened 2019-02-06
Do not create end product establishments for ineligible issues

leikkisa:
connects #9249

### DIFF
--- a/app/models/claim_review.rb
+++ b/app/models/claim_review.rb
@@ -52,7 +52,7 @@ class ClaimReview < DecisionReview
   # Create that end product establishment if it doesn't exist.
   def create_issues!(new_issues)
     new_issues.each do |issue|
-      if processed_in_caseflow?
+      if processed_in_caseflow? || !issue.eligible?
         issue.update!(benefit_type: benefit_type, veteran_participant_id: veteran.participant_id)
       else
         issue.update!(
@@ -199,8 +199,6 @@ class ClaimReview < DecisionReview
   end
 
   def end_product_establishment_for_issue(issue)
-    return unless issue.eligible?
-
     end_product_establishments.find_by(
       "(code = ?) AND (synced_status IS NULL OR synced_status NOT IN (?))",
       issue.end_product_code,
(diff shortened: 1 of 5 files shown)


app/models/claim_review.rb
```diff
@@ -199,6 +199,8 @@ def intake_processed_by
   end
 
   def end_product_establishment_for_issue(issue)
+    return unless issue.eligible?
```
pkarman: I wonder if this is the right place for this?

if you put the filter for `.eligible?` inside `create_issues!` then you could avoid calling this at all.

pkarman: e.g.
```diff
@@ -56,7 +56,7 @@ class ClaimReview < DecisionReview
         issue.update!(benefit_type: benefit_type, veteran_participant_id: veteran.participant_id)
       else
         issue.update!(
-          end_product_establishment: end_product_establishment_for_issue(issue),
+          end_product_establishment: issue.eligible? ? end_product_establishment_for_issue(issue) : nil,
           benefit_type: benefit_type,
           veteran_participant_id: veteran.participant_id
         )
```

I think I would find that more explicit.

pkarman: alternately:
```diff
@@ -52,7 +52,7 @@ class ClaimReview < DecisionReview
   # Create that end product establishment if it doesn't exist.
   def create_issues!(new_issues)
     new_issues.each do |issue|
-      if processed_in_caseflow?
+      if processed_in_caseflow? || !issue.eligible?
         issue.update!(benefit_type: benefit_type, veteran_participant_id: veteran.participant_id)
       else
         issue.update!(
```

spec/models/claim_review_spec.rb
```diff
@@ -325,6 +328,17 @@ def random_ref_id
     context "when there's more than one issue" do
       let(:issues) { [rating_request_issue, non_rating_request_issue] }
 
+      context "when they're all ineligible" do
```
pkarman: good test

app/models/claim_review.rb
leikkisa: I like that!

Issue #9123, opened 2019-02-01
Intake Noncomp | Per-organization user management should be delegated to Admin Intake, not System Admin CSEM function


steve-gov:
Currently a user must have the `System Admin` CSEM function to perform add/ remove users to an "organization," i.e. a noncomp LOB. To mitigate significant risk, change that to the `Admin Intake` function.

steve-gov:
I don't believe this is working. I was able to reach `/organizations/nca/users` with superuser `CF_Q_283`. We created a user with `Mail Intake` and `Admin Intake` (`bg101_dvsr`) and couldn't reach the URL.

pkarman:
This is my fault. Apparently the `granted` feature does not act on roles as I expected.
```
irb(main):007:0> u.granted?("Admin Intake")
=> false
irb(main):008:0> u.roles
=> ["User", "Admin Intake", "Mail Intake"]
```
I will fix.

steve-gov:
Still can't access `/organizations/nca/users` with `bg101_dvsr`. Is there some way I can confirm which CSEM functions this user has?

pkarman:
weird. via the UAT console, that user looks qualified:
```
irb(main):001:0> user = User.find_by(css_id: 'bg101_dvsr'.upcase)
[2019-02-06 10:22:35 -0500]   User Load (1.2ms)  SELECT  "users".* FROM "users" WHERE "users"."css_id" = $1 LIMIT $2  [["css_id", "BG101_DVSR"], ["LIMIT", 1]]
=> #<User id: 416, station_id: "101", css_id: "BG101_DVSR", full_name: "BRIAN GUTHRIE", email: "", roles: ["User", "Admin Intake", "Mail Intake"], selected_regional_office: nil>
irb(main):002:0> user.administer_org_users?
=> true
irb(main):003:0> user.organizations
[2019-02-06 10:22:54 -0500]   Organization Load (1.3ms)  SELECT  "organizations".* FROM "organizations" INNER JOIN "organizations_users" ON "organizations"."id" = "organizations_users"."organization_id" WHERE "organizations_users"."user_id" = $1 LIMIT $2  [["user_id", 416], ["LIMIT", 11]]
[...]

PR #9262, opened 2019-02-06
Fix user authz for admin of org users

pkarman:
connects #9123 

### Description
Fixes user authorization when accessing the /organizations/:business_line/users endpoint.

Actual tests!

### DIFF
--- a/app/controllers/organizations/users_controller.rb
+++ b/app/controllers/organizations/users_controller.rb
@@ -30,7 +30,9 @@ class Organizations::UsersController < OrganizationsController
   end
 
   def verify_organization_access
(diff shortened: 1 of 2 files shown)


app/controllers/organizations/users_controller.rb
```diff
@@ -30,7 +30,9 @@ def destroy
   end
 
   def verify_organization_access
-    redirect_to "/unauthorized" unless current_user.admin? || current_user.administered_teams.include?(organization)
+    return if current_user.administer_org_users?
```
kencheeto: you don't need to `return true` here, right?

pkarman: right. we just avoid issuing the `redirect_to`

Issue #6946, opened 2018-09-13
AMA Intake | Rollback AMA reviews when the EP is cancelled


leikkisa:
If someone cancels an end product, we should cancel the associated issues for that end product.

requires https://github.com/department-of-veterans-affairs/caseflow/issues/9188

## Acceptance criteria
- If the status of an end product becomes canceled (CAN), the request issues for that end product should be marked as closed with a `closed_status` of `end_product_canceled` (according to the rules of https://github.com/department-of-veterans-affairs/caseflow/issues/9188)

## Techical Notes
- We do not currently have any need for marking the entire HLR or SC as closed, so that will be out of scope for now.

shanear:
Not sure if we should hard delete them or soft delete them.

leikkisa:
I'm wondering, if we are keeping the EndProductEstablishment record with the status of CAN, and the review record (not sure if we are keeping this or not), should we just keep the issues too?  I guess keep all the links in tact.  Is there any downside?

shanear:
> I'm actually not sure if we are keeping AMA records around or not. 

Right now we are virtually deleting RampElections and doing nothing for RampRefilings and AMA reviews. I think this issue might need to be converted into a story for how we should roll back AMA reviews (and RampRefilings) in the database.

shanear:
This scenario was brought up again by testers. Probably going to prioritize.

shanear:
1. if there is no EP, we should not say "null" in EP Code
1. if the EP is not yet established the status should read "Establishing..."
1. if the EP is not yet established and there is any value in `establishment_error`, then it should stay "Establishment Failed. Support notified." Bonus points if this is a red color
1. remove "Decision Date" column

steve-gov:
EPs `036` and `037` on file `123458314` have been canceled. Checking statuses.

steve-gov:
Validated that the issues on those EPs have the correct statuses.

PR #9266, opened 2019-02-07
[WIP] Close request issues on canceled EPs

leikkisa:
connects #6946


app/models/end_product_establishment.rb
```diff
@@ -327,9 +327,16 @@ def cancel!
       # delete end product in bgs & set sync status to canceled
       BGSService.new.cancel_end_product(veteran_file_number, code, modifier)
       update!(synced_status: CANCELED_STATUS)
+      close_request_issues_if_canceled!
```
leikkisa: I am thinking this one might not be necessary since at this point all of the request issues would be removed.

leikkisa: Ahh. I am going to keep it in here, request_issue.status_active? is based on the end product establishment, not whether the request issue is closed. But I added a thing that if it was already closed due to being removed, it doesn't get reclosed.

Issue #8434, opened 2018-12-26
Pre-select judge to submit case to in attorney checkout


lowellrex:
During the attorney checkout process attorneys select the judge to which cases should be submitted. We can pre-select the judge using the [`assignedJudge` attribute of the appeal](https://github.com/department-of-veterans-affairs/caseflow/blob/6bea2aab4a10d6bf3de6b4b26ab1f229c9f5637a/client/app/queue/utils.js#L240). Doing this could save attorneys some time and make the process less error prone.

![image](https://user-images.githubusercontent.com/32683958/50448315-7b934b00-08ee-11e9-9ea7-64341c2e0b1d.png)

## Acceptance criteria
* [ ] Auto-populate the judge to return the case to

PR #9267, opened 2019-02-07
8234 persist attorney re write info

### LINKED ISSUE #8234: Display information from previous checkouts when attorney goes through checkout for a rewrite
#8233 allows judges to return cases to attorneys for corrections/rewrites. This ticket exists to display information from the previous attorney checkout on the final page of the checkout process when an attorney goes through the checkout process for a rewrite.

This issue was brought to light as the result of us manually returning a case to an attorney from a judge on 12 December 2018 (see the [related slack conversation](https://dsva.slack.com/archives/C6E41RE92/p1544627868008000)).

## Acceptance criteria
* [ ] Display information from previous checkouts when attorney goes through checkout for a rewrite

youngfreezyVA:
Resolves #8234 and #8434 

### Description
this issue describes it best:
https://github.com/department-of-veterans-affairs/caseflow/issues/8234
this pr only handles AMA appeals

### Acceptance Criteria
- [ ] in the `/submit` view, if checking out as an attorney, and this appeal has been returned to you by a judge, the following fields should be pre populated: 

- the judges instructions `AttorneyRewriteTask.instructions`
- the document id -
- the reviewing judge 
- overtime (?)

### Testing Plan
1. as a judge with cases, select an AMA appeal, return it to an attorney (remember the note)
2. switch users to that attorney, checkout
3. once you're at submit, all fields should be pre populated

@aroltsch @lowellrex please let me know if the naming of the functions/variables is accurate/explicit enough.


spec/feature/queue/ama_queue_spec.rb
```diff
@@ -276,27 +276,6 @@ def valid_document_id
         expect(page).to have_content("Task assigned to #{other_organization.name}")
         expect(Task.last.instructions.first).to eq(instructions)
       end
-
```
youngfreezyVA: I didn't delete this test. no idea how this got in the diff. I can add it back.

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -152,13 +152,22 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
     latest_attorney_case_review&.id
   end
 
+  attribute :attorney_case_review_details do
+    {
+      document_id: latest_attorney_case_review&.document_id,
+      assigned_judge: object.assigned_judge,
+      note: object.attorney_rewrite_task_for_appeal&.instructions&.first,
```
aroltsch: So in this case assigned_judge should be: 
```
object.attorney_rewrite_task_for_appeal&.assigned_by&.full_name
```

app/models/appeal.rb
```diff
@@ -137,6 +137,10 @@ def reviewing_judge_name
     task ? task.assigned_to.try(:full_name) : ""
   end
 
+  def attorney_rewrite_task_for_appeal
+    tasks.find { |t| t.is_a?(AttorneyRewriteTask) && (t.appeal_id == id) }
```
aroltsch: The second part in this find statement is redundant. 

when you call `tasks` inside the appeal class, it returns all tasks where t.appeal_id == id because of an appeal `has_many :tasks` association 

Also you probably want to look for non-completed AttorneyRewriteTask tasks because there are could be multiple.

spec/feature/queue/ama_queue_spec.rb
aroltsch: Yes, please.

aroltsch:
Hey Fareez, nice start! I've added a few comments. Also, we should add some unit tests for the methods you added on the BE.

app/models/serializers/work_queue/appeal_serializer.rb
aroltsch: I think `note` and `instructions` are two separate things and we should display them both. `note` is what attorney entered when going through the checkout and `instructions` is what a judge entered.  We need to ask our design team on where they want instructions to be displayed.

youngfreezyVA:
> Hey Fareez, nice start! I've added a few comments. Also, we should add some unit tests for the methods you added on the BE.

Hey @aroltsch addressed all your comments, please have a look! Thanks.

app/models/appeal.rb
```diff
@@ -137,6 +137,10 @@ def reviewing_judge_name
     task ? task.assigned_to.try(:full_name) : ""
   end
 
+  def attorney_rewrite_task_for_appeal
+    tasks.find { |t| t.is_a?(AttorneyRewriteTask) && (t.status == Constants.TASK_STATUSES.in_progress) }
```
aroltsch: I think more accurate way would be to also include assigned statuses. Also, you could do the filtering in SQL query though ActiveRecord instead of doing it in Ruby.

```
tasks.where(type: "AttorneyRewriteTask", status: [Constants.TASK_STATUSES.in_progress, Constants.TASK_STATUSES.assigned]
```

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -152,13 +152,23 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
...
   attribute :can_edit_document_id do
     AmaDocumentIdPolicy.new(
       user: @instance_options[:user],
       case_review: latest_attorney_case_review
     ).editable?
   end
-
+  # this should probably be moved to `appeal.rb` but didn't want to break anything
   def latest_attorney_case_review
     AttorneyCaseReview.where(task_id: Task.where(appeal: object).pluck(:id)).order(:created_at).last
```
aroltsch: Let's set it to an instance variable and use a conditional assignment operator here so it won't call this query every time you invoke this method. 

```
def latest_attorney_case_review
    @latest_attorney_case_review ||= AttorneyCaseReview.where(task_id: Task.where(appeal: object).pluck(:id)).order(:created_at).last
```

Let's do the same for `attorney_rewrite_task_for_appeal` method.

client/app/queue/SubmitDecisionView.jsx
```diff
@@ -175,25 +186,33 @@ class SubmitDecisionView extends React.PureComponent<Props> {
...
+        judgeSelector={
+          this.getDefaultJudgeSelector()
+        }
+      />
       <TextareaField
         label="Notes:"
         name="notes"
-        value={decisionOpts.note}
+        // the judges instructions take precedence in the note 
+        // over the note that the attorney previously wrote in checkout
```
aroltsch: Is that what design team suggested? Seems a bit weird to me

client/app/queue/SubmitDecisionView.jsx
```diff
@@ -133,15 +133,26 @@ class SubmitDecisionView extends React.PureComponent<Props> {
       });
   };
 
+  getDefaultJudgeSelector = () => {
+    const legacyJudgeSelector = _.get(this.props, 'task.addedByCssId', null);
+    const amaJudgeSelector = _.get(this.props.appeal, 'assignedJudge.full_name');
```
aroltsch: If `assignedJudge` is nil would it still work?

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -152,13 +152,23 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
     latest_attorney_case_review&.id
   end
 
+  attribute :attorney_case_review_details do
```
aroltsch: I would name it `attorney_case_rewrite_details` to differentiate between those two. Although I don't feel strongly about it.

client/app/queue/SubmitDecisionView.jsx
youngfreezyVA: nope! that was my own solution - I didn't speak with them, sorry! let me know what we should do here.

client/app/queue/SubmitDecisionView.jsx
youngfreezyVA: yes, the `_.get` method is basically a null safe operator for javascript, it resolves to undefined by default, it's the same thing as `&` in Rails

client/app/queue/SubmitDecisionView.jsx
aroltsch: I think we should ask our design team (LP, Sneha, Allyce). Could you post this question on slack to them. I personally think we should put judge's instructions somewhere else but we might need design for that and could do it as a separate ticket.

youngfreezyVA: ok, i will remove the logic about the judges rewrite instructions from this PR - resolving this convo since that will be handled in another itcket

app/models/appeal.rb
```diff
@@ -152,6 +152,12 @@ def reviewing_judge_name
     task ? task.assigned_to.try(:full_name) : ""
   end
 
+  def attorney_rewrite_task_for_appeal
+    @attorney_rewrite_task_for_appeal ||=
+      tasks.where(type: "AttorneyRewriteTask",
+                  status: [Constants.TASK_STATUSES.in_progress, Constants.TASK_STATUSES.assigned])
```
aroltsch: let's also sort it by created_at to get the latest one (you can probably use `max_by` method for that)

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -152,14 +152,24 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
     latest_attorney_case_review&.id
   end
 
+  attribute :attorney_case_rewrite_details do
+    {
+      document_id: latest_attorney_case_review&.document_id,
+      assigned_judge: object.attorney_rewrite_task_for_appeal&.first&.assigned_by,
```
aroltsch: If you use `max_by` above it will return on element instead of a collection.

`where` returns a collection

client/app/queue/SubmitDecisionView.jsx
```diff
@@ -116,7 +120,14 @@ class SubmitDecisionView extends React.PureComponent<Props> {
     } = this.props;
 
     const issuesToPass = !isLegacyAppeal && amaDecisionIssues ? decisionIssues : issues;
-    const payload = buildCaseReviewPayload(checkoutFlow, decision, userRole, issuesToPass, { isLegacyAppeal });
+    // not sure why this is happening, it's causing tests to fail
```
aroltsch: I don't understand what this comment relates to

app/models/appeal.rb
aroltsch: I think we should create a new method on task similar to this: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/task.rb#L39

We can call it `open_statuses` or something. @lowellrex do you have thoughts?

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -152,14 +152,24 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
     latest_attorney_case_review&.id
   end
 
+  attribute :attorney_case_rewrite_details do
+    {
+      document_id: latest_attorney_case_review&.document_id,
```
aroltsch: Is it necessary to include it here? There is already a document_id included: 
https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/serializers/work_queue/appeal_serializer.rb#L147

client/app/queue/SubmitDecisionView.jsx
```diff
@@ -37,46 +36,51 @@ import {
...
     const {
       opts: decisionOpts
     } = this.props.decision;
-    const requiredParams = ['document_id', 'reviewing_judge_id', 'work_product'];
 
+    const requiredParams = ['document_id', 'reviewing_judge_id', 'work_product'];
     const missingParams = _.filter(requiredParams, (param) => !_.has(decisionOpts, param) || !decisionOpts[param]);
 
-    return !missingParams.length;
+    const isValid = !missingParams.length;
```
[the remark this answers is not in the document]
youngfreezyVA: initially for debugging and then I figured it's also more explicit

client/app/queue/SubmitDecisionView.jsx
youngfreezyVA: if the frontend sends decision.opts.overtime as null to the backend, a 400 error is returned saying 'overtime is not on the list' -- this only happens on the automated tests though, i wasnt able to reproduce it locally. i think this is a backend bug. to work around this I had to make sure it was false instead of null

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -152,14 +152,24 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
...
+      assigned_judge: object.attorney_rewrite_task_for_appeal&.first&.assigned_by,
+      note_from_judge: object.attorney_rewrite_task_for_appeal&.first&.instructions&.first,
```
lowellrex: We usually keep most task-related information and workflow-related information (like attorney and judge checkout) separate from one another. We currently display notes from the judge back to the attorney in the task snapshot area:

image

Also, it looks like we don't actually end up using `note_from_judge` anywhere on the frontend. Can we get rid of it?

app/models/appeal.rb
```diff
@@ -152,6 +152,12 @@ def reviewing_judge_name
     task ? task.assigned_to.try(:full_name) : ""
   end
 
+  def attorney_rewrite_task_for_appeal
+    @attorney_rewrite_task_for_appeal ||=
+      tasks.where(type: "AttorneyRewriteTask",
```
lowellrex: Let's use `AttorneyRewriteTask.name` here so that the static analyzer can help us catch any issues if the name of this class changes somewhere down the line.

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -152,14 +152,24 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
...
     AmaDocumentIdPolicy.new(
       user: @instance_options[:user],
       case_review: latest_attorney_case_review
     ).editable?
   end
-
   def latest_attorney_case_review
-    AttorneyCaseReview.where(task_id: Task.where(appeal: object).pluck(:id)).order(:created_at).last
+    @latest_attorney_case_review ||=
+      AttorneyCaseReview.where(task_id: Task.where(appeal: object).pluck(:id)).order(:created_at).last
```
lowellrex: Can we take advantage of the [existing `Appeal.attorney_case_reviews` method](https://github.com/department-of-veterans-affairs/caseflow/blob/fa41c7799926426996592a0b8cacc2a7542be0e5/app/models/appeal.rb#L133) here?

client/app/queue/JudgeSelectComponent.jsx
```diff
@@ -48,7 +48,11 @@ class JudgeSelectComponent extends React.PureComponent<Props> {
   }
 
   setDefaultJudge = (judges) => {
-    const judge = _.find(judges, { css_id: this.props.assignedByCssId });
+    const judge =
+       // for legacy
+       _.find(judges, { css_id: this.props.judgeSelector }) ||
+       // for ama
+       _.find(judges, { full_name: this.props.judgeSelector });
```
lowellrex: What is this change?

lowellrex: Ahhh I see that we've [changed the interface](https://github.com/department-of-veterans-affairs/caseflow/pull/9267/files#diff-dc6af3991f74c682c598e9a98c97bf8dR147) for this component.

I think we can use the ID of the judge instead of the name so we guarantee the ID will be unique when the name might not be. This also allows us to not have to worry about changing formatting of names as they appear in this list.

image

client/app/queue/SubmitDecisionView.jsx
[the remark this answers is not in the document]
lowellrex: Can `isValid` be false?

client/app/queue/SubmitDecisionView.jsx
```diff
@@ -133,6 +144,13 @@ class SubmitDecisionView extends React.PureComponent<Props> {
       });
   };
 
+  getDefaultJudgeSelector = () => {
+    const legacyJudgeSelector = _.get(this.props, 'task.addedByCssId', null);
```
lowellrex: I think we can use the existing `isLegacy` attribute of the `this.props.task` property to determine whether to return a CSS ID or the User ID. Maybe something like:

```jsx
getDefaultJudgeSelector = () => this.props.task.isLegacy ? this.props.task.addedByCssId : this.props.task.assignedBy.pgId;
```

client/app/queue/SubmitDecisionView.jsx
```diff
@@ -37,46 +36,51 @@ import {
...
+      });
+    });
+
+  }
+
+  // this handles the case where there is no document_id on this.props.decision.opts
+  // and it comes from this.props.appeal.attorneyCaseRewriteDetails instead
+  // if you don't do this you will get validation errors and a 400 for a missing document_id
+  // this is also needed to keep the onChange values in sync
+  extendDecisionOptsWithAttorneyCheckOutInfo = (decision, attorneyCaseRewriteDetails) => {
```
lowellrex: Can we name this function something along the lines of `setInitialDecisionOptions`?

client/app/queue/SubmitDecisionView.jsx
```diff
@@ -116,7 +120,14 @@ class SubmitDecisionView extends React.PureComponent<Props> {
     } = this.props;
 
     const issuesToPass = !isLegacyAppeal && amaDecisionIssues ? decisionIssues : issues;
-    const payload = buildCaseReviewPayload(checkoutFlow, decision, userRole, issuesToPass, { isLegacyAppeal });
+    // not sure why this is happening, it's causing tests to fail
+    // the backend can't handle a null response at this point
+
+    if (decision.opts.overtime === null) {
+      decision.opts.overtime = false;
```
lowellrex: Let's set this value to false when [we pluck out the initial values](https://github.com/department-of-veterans-affairs/caseflow/pull/9267/files#diff-dc6af3991f74c682c598e9a98c97bf8dR58) of the other fields.

app/models/serializers/work_queue/appeal_serializer.rb
lowellrex: Should be able to access it in `SubmitDecisionView` by `this.props.appeal.documentId`.

app/models/serializers/work_queue/appeal_serializer.rb
lowellrex: I don't think we need to return this value in the serializer because it is already available to us in `SubmitDecisionView` by calling `this.props.task.assignedBy`.

app/models/serializers/work_queue/appeal_serializer.rb
```diff
@@ -152,14 +152,24 @@ class WorkQueue::AppealSerializer < ActiveModel::Serializer
     latest_attorney_case_review&.id
   end
 
+  attribute :attorney_case_rewrite_details do
+    {
+      document_id: latest_attorney_case_review&.document_id,
+      assigned_judge: object.attorney_rewrite_task_for_appeal&.first&.assigned_by,
+      note_from_judge: object.attorney_rewrite_task_for_appeal&.first&.instructions&.first,
+      overtime: latest_attorney_case_review&.overtime,
+      note_from_attorney: latest_attorney_case_review&.note
```
lowellrex: I think `overtime` and `note_from_attorney` are the only two new fields we need to return in the appeal serializer for this PR.

app/models/serializers/work_queue/appeal_serializer.rb
youngfreezyVA: yes! I was planning on doing that for this PR, as the UX team is still deciding if/where we need to place that somewhere (will be a separate task)

app/models/serializers/work_queue/appeal_serializer.rb
youngfreezyVA: based on the fact that we don't need alot the information I thought we did, I can just grab all the info I need from this method: ```def latest_attorney_case_review
    @latest_attorney_case_review ||=
      AttorneyCaseReview.where(task_id: Task.where(appeal: object).pluck(:id)).order(:created_at).last
  end```

youngfreezyVA: oh I may have misunderstood your question - this isn't a new method I created, it was already there, I just modified it per @aroltsch feedback so that it only makes the query once (initially) when that method is called in the class

client/app/queue/SubmitDecisionView.jsx
youngfreezyVA: awesome, thanks!

spec/factories/task.rb
```diff
@@ -105,6 +105,11 @@
       parent { create(:ama_judge_task) }
     end
 
+    factory :ama_attorney_rewrite_task do
```
aroltsch: @youngfreezy I can see you added this factory but I am not seeing a feature test for that. Could you please add the test.

PR #9268, opened 2019-02-07
Add ama_opt_in_eligible alert (closes #8809)

### LINKED ISSUE #8809: Add amaOptInEligible alert to API
When an active appeal has received an SOC or SSOC within the last 60 days, the API should return an alert of type amaOptInEligible. This alert should have a dueDate that matches the logic used by Intake to determine whether an appeal is eligible for statutory opt-in.

This alert must be behind a feature flag.

cmgiven:
Resolves #8809

### Description
Adds an alert to a legacy appeal when there is an SOC or SSOC within the last 60 days.

### DIFF
--- a/spec/models/appeal_series_alerts_spec.rb
+++ b/spec/models/appeal_series_alerts_spec.rb
@@ -16,6 +16,10 @@ describe AppealSeriesAlerts do
     DocketSnapshot.create
   end
 
+  after do
+    FeatureToggle.disable!(:api_appeal_status_v3)
+  end
+
   let(:docket_month) { 1.year.ago.to_date.beginning_of_month }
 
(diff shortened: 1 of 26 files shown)


spec/models/appeal_series_alerts_spec.rb
```diff
@@ -1,5 +1,6 @@
 describe AppealSeriesAlerts do
   before do
+    FeatureToggle.enable!(:api_appeal_status_v3)
```
mc-usds: Should we disable the feature flag in an after block?

cmgiven: Yeah, that seems like a good precaution.

Issue #8170, opened 2018-12-10
Create easy way for team admins to navigate to team user management page


lowellrex:
There is currently no way for admins of the various teams at the Board (and VSOs) to navigate to the user management page. This ticket exists to do the engineering work required to add that link/dropdown/whatever other method we decide on to navigate to the user management page.

_Current look at the user management page:_
![image](https://user-images.githubusercontent.com/32683958/49739815-c5c6cc00-fc60-11e8-8ef8-cde718c43e9e.png)

## Acceptance criteria
* [ ] Team admins and judges see an option in their top dropdown menu with the copy "[Team name] management"
* [ ] Team admins or judges click that link, they navigate to the user management page (`/organizations/vlj-support-staff/users`)

laurjpeterson:
@lowellrex - I connected with @allyceh and @sneha-pai on this and updated the AC above. would love your input on if you think using the Switch Views dropdown will work for this for now.

lowellrex:
Sure! I think using the "Switch Views" dropdown will definitely work for now

joeyyang:
couple questions: 
- how do we tell if someone is a team admin?
- we are putting this dropdown both on the user management page (`/organizations/:organization/users`) and [somewhere else? anywhere else where the switcher lives?]

laurjpeterson:
I can answer your second question: Yes, we should put it everywhere the switcher lives. I think this is individual queues, organization queues, and the user management page.

lowellrex:
> how do we tell if someone is a team admin?

The `OrganizationsUser` table has an `admin` field that tells us if somebody is an admin of a team. We can use the `User.administered_teams` method when we build the list of `selectable_organizations` to determine the URL we should be pointing to (though we will probably have to move the serialized object [we build in the template file](https://github.com/department-of-veterans-affairs/caseflow/blob/0590e73effa48112a2a340d740ad6cb23901268e/app/views/queue/index.html.erb#L7) into the user model).

lowellrex:
Exclude `JudgeTeam` admins from being able to visit the team admin page until @laurjpeterson can confirm that judges should be allowed to determine who is on their team.

joeyyang:
discussed this approach with @sneha-pai — we thought it'd make more sense (and be much simpler to implement) if we put this link in the user's top menu dropdown, here:
[...]

PR #9288, opened 2019-02-07
add link to user management page in top menu for team admins

joeyyang:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/8170
BLOCKED BY: https://github.com/department-of-veterans-affairs/caseflow/pull/9290
/cc @laurjpeterson to coordinate comms

### Description
Adds a link to the team user management page in the user's top bar if they are a team admin or judge.
screenshot 2019-02-07 11 04 23


app/controllers/application_controller.rb
```diff
@@ -29,6 +29,15 @@ class ApplicationController < ApplicationBaseController
 
   private
 
+  def manage_teams_menu_items
+    current_user.administered_teams.map do |team|
+      {
+        title: "#{team.name} team management",
+        link: "/organizations/#{team.url}/users"
```
lowellrex: I think we can just use [the `user_admin_path` method](https://github.com/department-of-veterans-affairs/caseflow/blob/04be94d27d405a44d32259bd9bef68bdc206a5b4/app/models/organization.rb#L49) of the `Organization` model here.

lowellrex:
One small comment but it looks really good otherwise! Can we add a test or two for this?

spec/feature/queue/user_organization_spec.rb
```diff
@@ -21,11 +21,28 @@
...
     before { OrganizationsUser.make_user_admin(user, organization) }
 
+    scenario "Has a link to the team admin page" do
+      visit "/"
+
+      find(".cf-dropdown-trigger").click
+      expect(page).to have_content("#{organization.name} team management")
+
+      find("a", text: "#{organization.name} team management").click
+      expect(page).to have_content("#{organization.name} team")
```
lowellrex: Maybe we could add an additional test here to make sure we ended up at the correct page? Something like the following maybe?

```
expect(page.current_path).to eq(organization.user_admin_path)
```

Issue #9092, opened 2019-01-31
AMA Intake | Any issue older than the current issue in a chain should not be selectable


steve-gov:
An appellant can only appeal the latest issue in a chain of decisions.

As an example, if there is a decision made on 01/01/2019 that is appealed, and that appeal results in a decision made on 02/01/2019, only the latter decision issue may be appealed.

For MVP treatment, any issue that is part of an issue chain and is older than any other issue in the chain should not be selectable. The UI should guide the user to selecting the most recent decision.

## Acceptance Criteria
- If a decision issue on the first Add Issues modal is contested by another decision issue on the first Add Issues modal, then the issue being contested should be disabled. There should be an italicized message beside the option reading "(Please select most recent decision on MM/DD/YYYY)"

## Notes
The visual treatment should match the treatment we already have for preventing duplicate selections:
<img width="591" alt="screenshot 2019-01-31 16 21 10" src="https://user-images.githubusercontent.com/1596259/52094543-64422000-2574-11e9-9e79-0c6f8e8ac3e7.png">

Related to #1094.

steve-gov:
@shanear This is one treatment of the problem -- the "ineligible" route, which covers a number of use cases. We can also assume the appellant _intended_ to appeal the latest issue by default. I can rewrite the issue if that's preferred.

shanear:
>  We can also assume the appellant intended to appeal the latest issue by default.

I prefer this method. Seems better (and less annoying) for vets.

jiahuang:
Just to clarify:

jiahuang:
Was able to verify in UAT with veteran file number 700032874
<img width="980" alt="screen shot 2019-02-21 at 9 14 13 pm" src="https://user-images.githubusercontent.com/577629/53215333-29d11d80-361e-11e9-8768-c6e7198d1014.png">

PR #9292, opened 2019-02-07
Find latest contestable issue in chain

jiahuang:
Connects #9092

Adds a `find_latest_contestable_issue` method to ContestableIssue which walks to the latest contestable issues in a chain.

screen shot 2019-02-21 at 8 06 03 pm


app/services/contestable_issue_chain.rb
```diff
@@ -0,0 +1,26 @@
+class ContestableIssueChain
```
jiahuang: gonna switch to a module

app/services/contestable_issue_chain.rb
```diff
@@ -0,0 +1,52 @@
...
+    @chain.last
+  end
+
+  private
+
+  def build_later_issues_chain(contestable_issue)
+    # this only appends future decision issues to the chain
+    future_decision_issues = []
+    # contestable issues made from rating issues will always be in a chain of only itself
+    return future_decision_issues if !contestable_issue.decision_issue?
```
shanear: this is incorrect, explained in person

app/services/contestable_issue_chain.rb
```diff
@@ -0,0 +1,41 @@
+class ContestableIssueChain
```
jiahuang: ContestableIssueChain isn't really a great name for this since it returns all `decision issues` and not contestable issues. Maybe this should be renamed to `FutureDecisionIssueChain`?

app/services/contestable_issue_chain.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def build_later_issues_chain(contestable_issue)
+    # this only appends future decision issues to the chain
+    # prior decision issues are not included in this chain
+    next_decision_issue = contestable_issue.next_decision_issue
+    return [] unless next_decision_issue
+
+    future_decision_issues = [next_decision_issue]
+    while (next_decision_issue = next_decision_issue.next_decision_issue)
+      if future_decision_issues.map(&:id).include? next_decision_issue.id
```
jiahuang: tried moving this check elsewhere but couldn't really figure out how to do it

app/models/contestable_issue.rb
```diff
@@ -4,13 +4,16 @@ class ContestableIssue
...
+                :rating_issue_disability_code, :rating_issue_profile_date_timestamp
 
   class << self
     def from_rating_issue(rating_issue, contesting_decision_review)
       new(
         rating_issue_reference_id: rating_issue.reference_id,
         rating_issue_profile_date: rating_issue.profile_date.to_date,
+        # keep track of the original profile date with timestamp
+        # we need this to go from contestable issue => request issue
+        rating_issue_profile_date_timestamp: rating_issue.profile_date.to_s,
```
jiahuang: not a big fan of this. Maybe this should be

```
rating_issue_profile_date: rating_issue.profile_date
```

And then the serialize actually converts it to a date?

app/services/contestable_issue_chain.rb
shanear: can we make all of the values lazy loaded? meaning we could modify the class to be the entire chain if we need that one day?

app/models/contestable_issue.rb
```diff
@@ -54,12 +58,40 @@ def source_review_type
...
+                         contested_rating_issue_profile_date: rating_issue_profile_date_timestamp)
+  end
+
   def decision_issue?
     !!decision_issue_id
   end
 
+  def serialize_latest_decision_issue_in_chain
+    {
+      id: contestable_issue_chain.last_issue&.id,
```
shanear: i don't like the `&.` here. Shouldn't there always be a last issue (if there's just one issue it should be itself right?)

Also returning `{ id: nil, date: nil }` seems bad for downstream consumers

app/models/contestable_issue.rb
```diff
@@ -54,12 +58,40 @@ def source_review_type
     decision_issue? ? source_request_issues.first.decision_review_type : source_request_issues.first.review_request_type
   end
 
+  def next_decision_issue
+    decision_issue&.next_decision_issue || request_issue&.next_decision_issue
+  end
```
shanear: It's not clear what this method does, or what `decision_issue` or `request_issue` mean in relation to `contestable_issue`. I think we're going to have to get more specific on the naming here, or at least add a description of what's returned here.

app/models/contestable_issue.rb
```diff
@@ -54,12 +58,40 @@ def source_review_type
...
+    return unless decision_issue?
+
+    DecisionIssue.find(decision_issue_id)
+  end
+
+  def request_issue
+    return if decision_issue?
+
+    RequestIssue.find_by(contested_rating_issue_reference_id: rating_issue_reference_id,
+                         contested_rating_issue_profile_date: rating_issue_profile_date_timestamp)
```
shanear: this seems like it's finding the request_issue that is contesting this issue? This would make more sense maybe named `contested_by_request_issue`

app/models/contestable_issue.rb
shanear: `rating_issue_reference_id` should be unique, so we don't need the timestamp

jiahuang: wait i thought we needed both rating_issue_profile_date & rating_issue_reference_id to uniquely identify a rating issue?

app/models/contestable_issue.rb
```diff
@@ -23,7 +23,7 @@ def from_rating_issue(rating_issue, contesting_decision_review)
     def from_decision_issue(decision_issue, contesting_decision_review)
       new(
         rating_issue_reference_id: decision_issue.rating_issue_reference_id,
-        rating_issue_profile_date: decision_issue.profile_date.try(:to_date),
```
shanear: nice 👍 much cleaner

app/models/contestable_issue.rb
```diff
@@ -54,12 +55,51 @@ def source_review_type
...
+    return unless decision_issue?
+
+    DecisionIssue.find(decision_issue_id)
+  end
+
+  def contested_by_request_issue
+    return [] if decision_issue?
+
+    RequestIssue.where(contested_rating_issue_reference_id: rating_issue_reference_id,
+                       contested_rating_issue_profile_date: rating_issue_profile_date.to_s)
```
shanear: The profile date is only needed for looking up the value in BGS. It's not actually a uniqueness constraint 🙃

This would find the contesting request issue no matter if it was a rating or decision issue or both.

Another thing, is that we don't want to pick up deleted/withdrawn/etc request issues. I think we will need a scope for that. Maybe `.active_or_decided`? That name feels clunky but I can't think of anything better right now.

```
RequestIssue.active_or_decided.where(
  contested_rating_issue_reference_id: rating_issue_reference_id,
  contested_decision_issue_id: decision_issue_id
)
```

app/models/contestable_issue.rb
```diff
@@ -54,12 +55,51 @@ def source_review_type
     decision_issue? ? source_request_issues.first.decision_review_type : source_request_issues.first.review_request_type
   end
 
+  def next_decision_issues
+    associated_decision_issue&.next_decision_issues || contested_by_request_issue.map(&:next_decision_issues).flatten
```
shanear: If the changes made to `contested_by_request_issue` are made then this can be simplified to:

```
def next_decision_issues
  contested_by_request_issue.next_decision_issues
end
```

Another thing to note is that I think we can assume that a decision issue can only be contested by one active or decided request issue, since we are preventing dups.

app/models/contestable_issue.rb
```diff
@@ -54,12 +55,51 @@ def source_review_type
...
+    serialized_issues = latest_contestable_issues.map do |latest_contestable_issue|
+      {
+        id: latest_contestable_issue.decision_issue_id,
+        date: latest_contestable_issue.date
+      }
+    end
+    serialized_issues.sort_by { |issue| issue[:date] }
+  end
+
+  def find_latest_contestable_issue
```
shanear: this should be `find_latest_contestable_issues` right?

app/models/contestable_issue.rb
```diff
@@ -54,12 +55,51 @@ def source_review_type
...
     !!decision_issue_id
   end
 
+  def serialize_latest_decision_issues
+    serialized_issues = latest_contestable_issues.map do |latest_contestable_issue|
+      {
+        id: latest_contestable_issue.decision_issue_id,
+        date: latest_contestable_issue.date
+      }
+    end
```
shanear: this should be an instance method:

```
def serialize_id_and_date
  { id: decision_issue_id, date: date }
end
```

Which is more reusable, and cleans this method up:
```
def serialize_latest_decision_issues
  latest_contestable_issues.map(&:serialize_id_and_date).sort_by { |i| i[:date] }
end
```

app/models/decision_issue.rb
```diff
@@ -87,6 +87,21 @@ def contention_text
     Contention.new(description).text
   end
 
+  def contests_request_issue(request_issue_id)
+    request_issues.any{ |request_issue| request_issue.id == request_issue_id}
+  end
+
+  def next_decision_issues
```
shanear: This method is unnecessary if changes to `ContestableIssue.next_decision_issues` are made.

app/models/decision_issue.rb
```diff
@@ -87,6 +87,21 @@ def contention_text
     Contention.new(description).text
   end
 
+  def contests_request_issue(request_issue_id)
```
shanear: This is also unnecessary.

app/models/request_issue.rb
```diff
@@ -410,6 +410,10 @@ def requires_record_request_task?
     !benefit_type_requires_payee_code?
   end
 
+  def next_decision_issues
```
shanear: This is unnecessary. `RequestIssue.decision_issues` should get you want you want here.

app/models/contestable_issue.rb
shanear: nope, just when we retrieve the the issue from BGS

app/models/contestable_issue.rb
jiahuang: > This would find the contesting request issue no matter if it was a rating or decision issue or both.

I think we need to always select the request issue contesting a decision issue _except_ if the contestable issue was built from a rating issue. Otherwise you end up in an infinite loop where you just find the same request issue contesting a rating issue.

app/models/request_issue.rb
```diff
@@ -185,6 +181,10 @@ def from_intake_data(data, decision_review: nil)
       new(attrs).tap(&:validate_eligibility!)
     end
 
+    def active_or_decided
+      open.where(decision_sync_error: nil).where.not(decision_sync_submitted_at: nil)
```
shanear: are we sure this works for appeal decision issues?

app/models/request_issue.rb
```diff
@@ -185,6 +181,10 @@ def from_intake_data(data, decision_review: nil)
       new(attrs).tap(&:validate_eligibility!)
     end
 
+    def active_or_decided
```
shanear: this might be more consistent as `open_or_decided`

spec/models/contestable_issue_spec.rb
```diff
@@ -128,4 +128,102 @@
       end
     end
   end
+
+  context "#latest_contestable_issues" do
```
shanear: This sets up a lot of data in order to test some scenarios that don't really require it. It also doesn't really hit all the logic paths. It'd be a little tighter as something like this:

```
context "when rating issue" do
  context "when no following issues" do
  end

  context "when one following issue" do
  end

  context "when multiple staggered following issues" do
  end
end

context "when decision issue" do
end
```

shanear: We should use:

```
subject { contestable_issue. latest_contestable_issues }
```

spec/models/contestable_issue_spec.rb
```diff
@@ -128,4 +128,102 @@
...
+    let(:appeal) { create(:appeal) }
```
shanear: isn't this value set at the top of the file?

spec/models/contestable_issue_spec.rb
```diff
@@ -128,4 +128,102 @@
...
+    let(:starting_date) { Time.zone.now - 20.days }
```
shanear: If date isn't relevant in the logic of the method, then I don't think we should be specifying it. Misleads the reader to believe that it's relevant

spec/models/contestable_issue_spec.rb
```diff
@@ -128,4 +128,102 @@
...
+    let!(:rating_contestable_issue) do
+      rating_issue = RatingIssue.new(
+        reference_id: "NBA",
+        participant_id: "123",
+        profile_date: starting_date - 2.days,
+        promulgation_date: starting_date - 3.days,
+        decision_text: "This broadcast may not be reproduced",
+        associated_end_products: [],
+        rba_contentions_data: [{}]
+      )
```
shanear: We should be using the factory here to only specify relevant values 
```
create(:request_issue, ...)
```

shanear: Also we should use any setup code from the top of the spec if possible.

spec/models/contestable_issue_spec.rb
```diff
@@ -128,4 +128,102 @@
...
+                             contested_decision_issue_id: nil)
+
+      create(:decision_issue,
+             decision_review: appeal,
+             description: "decision issue for initial request issue",
+             caseflow_decision_date: starting_date,
+             request_issues: [request_issue])
+    end
+
+    let!(:future_contestable_issues) do
```
shanear: Following from my previous note, it would be a lot easier to see how all of this data is relevant if it was created in the context where it's relevant.

spec/models/contestable_issue_spec.rb
```diff
@@ -128,4 +128,102 @@
...
+      decision_issue = create(:decision_issue,
+                              decision_review: future_contestable_issues.second.contesting_decision_review,
+                              description: "another decision issue",
+                              caseflow_decision_date: starting_date + 4.days,
+                              request_issues: future_contestable_issues.second.source_request_issues)
+      ContestableIssue.from_decision_issue(decision_issue, future_contestable_issues.second.contesting_decision_review)
+    end
+
+    # rubocop:disable Metrics/AbcSize
+    def check_latest_contestable_issues(latest_contestable_issues)
```
shanear: I think sometimes using a function like this is a crutch for writing a simpler spec. It's useful in feature tests where the test is going to be more "integration-y", but unit tests want to stay as focused as possible. It's really difficult to follow this. Can we make this simpler by making the spec more focused?

app/models/request_issue.rb
jiahuang: You're right, this won't work for appeal decision issues since they never get synced. 

I thought about this some more though, don't think we need to filter by both open and active. Filtering by just `open` would accomplish the same thing for finding future decision issues. Request issues for hlr/scs w/ unsynced EPs will just not have any decision issues. The same is true for appeal request issues that have not yet been processed.

An alternative is to do something like this, but I'm not sure how to figure out if an appeal's request issues have been decided or not except by also looking through the decision issue table or maybe tasks.

```
def open_or_decided
      open.decided_request_issues_for_claim_reviews.or.decided_request_issues_for_appeals
end

def decided_request_issues_for_claim_reviews
      # decided request issues for sc/hlrs are request issues with eps & sync times
      where(decision_sync_error: nil).where.not(decision_sync_submitted_at: nil, end_product_establishment_id: nil)
end

def decided_request_issues_for_appeals
      # decided request issues for appeals do not have eps but have tasks? matching decision issues?
      where(end_product_establishment_id: nil) # todo filter by something else
end
```

app/models/contestable_issue.rb
```diff
@@ -44,22 +44,37 @@ def from_decision_issue(decision_issue, contesting_decision_review)
...
       titleOfActiveReview: title_of_active_review,
       sourceReviewType: source_review_type,
-      timely: timely?
+      timely: timely?,
+      latestIssuesInChain: serialize_latest_decision_issues
     }
   end
 
+  def serialize_decision_id_and_date
+    { id: decision_issue&.id, date: approx_decision_date }
```
leikkisa: I think we've decided to call "date" approx_decision_date, but maybe that didn't apply to everywhere.

app/models/contestable_issue.rb
```diff
@@ -68,8 +83,21 @@ def title_of_active_review
...
 
-  def decision_issue?
-    !!decision_issue_id
+  def contested_by_request_issue
+    RequestIssue.open.find_by(contested_rating_issue_reference_id: rating_issue_reference_id,
+                              contested_decision_issue_id: decision_issue&.id)
+  end
+
+  def serialize_latest_decision_issues
+    latest_contestable_issues.map(&:serialize_decision_id_and_date).sort_by { |issue| issue[:date] }
```
leikkisa: Would get updated also if "date" above is updated to approx_decision_date

app/models/contestable_issue.rb
```diff
@@ -68,8 +83,21 @@ def title_of_active_review
...
+  end
+
+  def find_latest_contestable_issues
+    return [self] if next_decision_issues.blank?
+
+    next_decision_issues.map do |decision_issue|
+      ContestableIssue.from_decision_issue(decision_issue, decision_issue.decision_review).latest_contestable_issues
```
leikkisa: should decision_issue.decision_review be the current decision_review (the method calls this argument the contesting_decision_review)? I think the decision_issue's decision review is the previous one, where it was contested in the past.

client/app/intake/components/AddIssuesModal.jsx
```diff
@@ -93,14 +93,31 @@ class AddIssuesModal extends React.Component {
...
+
+            text = `${text} (Please select the most recent decision on ` +
+                   `${dates})`;
+          }
+
           return {
             displayText: text,
             value: issue.index,
-            disabled: foundIndex !== -1
+            disabled: (foundIndex !== -1) || hasLaterIssueInChain
```
leikkisa: Nice job on this page!

spec/models/contestable_issue_spec.rb
```diff
@@ -137,12 +140,96 @@
...
+
+        contestable_issues
+      end
+
+      it "finds the latest contestable issue" do
+        expect(subject.length).to eq(1)
+        expect(subject.first.decision_issue.id).to eq(future_contestable_issues.last.decision_issue.id)
+      end
+
+      context "when there are multiple future decision issues" do
```
leikkisa: Nice

PR #9293, opened 2019-02-07
Hold Hearing Task: hearing association

sharonwarner:
Because appeals can have multiple hearings, we need a way to associate a HearingTask directly to a hearing. This PR updates our hearing schedule process such that when a hearing is scheduled, a HoldHearingTask is created and its parent HearingTask is associated to the scheduled hearing.


app/models/tasks/hold_hearing_task.rb
```diff
@@ -1,11 +1,15 @@
...
   class << self
-    def create_hold_hearing_task!(appeal, parent)
+    def create_hold_hearing_task!(appeal, parent, hearing)
       HoldHearingTask.create!(
         appeal: appeal,
         parent: parent,
         assigned_to: Bva.singleton
       )
+
+      if parent.is_a? HearingTask
```
sharonwarner: @oscarAtNava FYI - this check can be removed once the backfill is complete.

PR #9299, opened 2019-02-07
MVP Transcription Queue

### LINKED ISSUE #8651: Transcription Queue Pre MVP
## User Stories
As a Hearings Schedule user, I need to be able to successfully record a held hearing and make sure that hearing is transcribed fully and successfully.

## Mocks
https://va-gov.invisionapp.com/share/MXPXWUDKSP6

## Acceptance Criteria
- [x] Verify the task can be closed by choosing Mark as complete: Transcribed. This will close the task and release the case to judge review.
- [x] Verify the task can be closed by choosing Mark as complete: Reschedule hearing. Choosing this option will create a ScheduleHearingTask assigned to Hearings Management.

## Related Tickets
See Epic #8650 for the complete scope of work related to this ticket.
See #8655 for the full Postponement flow.

sharonwarner:
Resolves #8651 

### Description
This PR creates a transcription queue and allows the users to complete the transcription task and send the task to case distribution or complete the transcsription task and reschedule the hearing.


spec/factories/task.rb
```diff
@@ -105,6 +105,13 @@
       parent { create(:ama_judge_task) }
     end
 
+    factory :transcription_task do
+      type TranscriptionTask.name
+      appeal { create(:appeal) }
+      parent { create(:root_task, appeal: appeal) }
```
sharonwarner: I'm not sure if you already know this, but there's definitely some weirdness in the factories for other tasks where the appeal for the root task won't match the appeal for the created task.

app/models/tasks/transcription_task.rb
```diff
@@ -0,0 +1,5 @@
+class TranscriptionTask < GenericTask
+  def available_actions(_user)
+    [Constants.TASK_ACTIONS.RESCHEDULE_HEARING.to_h, Constants.TASK_ACTIONS.COMPLETE_TRANSCRIPTION.to_h]
```
lowellrex: Do we want to limit who can act on a transcription task to only members of the transcription team?

spec/factories/task.rb
```diff
@@ -105,6 +105,13 @@
       parent { create(:ama_judge_task) }
     end
 
+    factory :transcription_task do
```
lowellrex: Can we make this into `factory : transcription_task, class: TranscriptionTask do` so that the factory returns an instance of `TranscriptionTask` instead of an instance of `Task`?

app/models/tasks/transcription_task.rb
```diff
@@ -0,0 +1,66 @@
...
+      else
+        super(params, current_user)
+      end
+    end
+
+    [self]
+  end
+
+  def hearing_task
+    return @hearing_task if @hearing_task
```
sharonwarner: I think we could make this a little bit more rigid. The HearingTask should always be 2 levels above the TranscriptionTask. I'm not sure if we gain anything with the flexibility this provides, and we could accidentally mask bugs.

mdbenjam: Cool, I can make that change. My original thought here was if we move this to the hearing_task model and then make all hearing related tasks inherit from hearing_task then we can  have a generic method to find the root hearing_task. But we can refactor as we need to later.

PR #9302, opened 2019-02-08
Revert "Temporarily disable hearing badge"

joeyyang:
Reverts department-of-veterans-affairs/caseflow#9271
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/7478

Adds a check to make sure we don't make a request if `externalId` is `undefined`.


client/app/queue/components/HearingBadge.jsx
```diff
@@ -44,7 +44,7 @@ const listStyling = css({
 
 class HearingBadge extends React.PureComponent {
   componentDidMount = () => {
-    if (!this.props.mostRecentlyHeldHearingForAppeal && !this.props.hearing) {
+    if (!this.props.mostRecentlyHeldHearingForAppeal && !this.props.hearing && this.props.externalId) {
```
joeyyang: here's the change

lowellrex: How could we end up in a situation where we didn't have an `externalId` here? Were we calling this when we pass in `hearing` property [so we don't ever set the `externalId` property](https://github.com/department-of-veterans-affairs/caseflow/blob/f112f82a51c5c19d9f991b87cf28d2406f56ff90/client/app/queue/components/HearingBadge.jsx#L87) but that value `hearing` is false? Or the task does not have an associated appeal [with an `externalId` value](https://github.com/department-of-veterans-affairs/caseflow/blob/f112f82a51c5c19d9f991b87cf28d2406f56ff90/client/app/queue/components/HearingBadge.jsx#L89) so that value is undefined? Or something else entirely?

app/controllers/appeals_controller.rb
```diff
@@ -64,7 +64,6 @@ def power_of_attorney
     }
   end
 
-  # :nocov:
   def hearings
```
lowellrex: Let's count how many request we make so that we are more confident we are making these requests without error. Following the pattern [established by `handle_non_critical_error`](https://github.com/department-of-veterans-affairs/caseflow/blob/f112f82a51c5c19d9f991b87cf28d2406f56ff90/app/controllers/application_controller.rb#L32) it might look something like:

```ruby
DataDogService.increment_counter(
  metric_group: "request_counter",
  metric_name: "hearings_for_appeal",
  app_name: RequestStore[:application]
)
```

lowellrex:
Let's return the hearing badge to the queue table views slowly so that 1) we limit exposure to this error if it comes back and 2) we can pinpoint which particular addition caused the failures if they return. I propose we start by only adding the hearing badge back for the case search results page (`CaseListTable` component), monitor [usage graphs](https://github.com/department-of-veterans-affairs/caseflow/pull/9302#discussion_r255185164) to make sure we are actually hitting that endpoint, monitor slack/sentry for the first sign of errors, and then add the hearing badge back to each table view whenever we are confident that there are no errors for the hearing badge in production (maybe as often as every day?).

client/app/queue/components/HearingBadge.jsx
joeyyang: No idea! I believe I don't have enough test data to verify any of this behavior and I'm not familiar enough with the React lifecycle to know for sure. My read of the code says that if we have a hearing, the call never gets made. Any guesses?

joeyyang:
> I propose we start by only adding the hearing badge back for the case search results page (`CaseListTable` component)

Since the badge was already there (and supposedly not causing problems), should we pick another place to add it back, perhaps somewhere with medium request volume, like `OrganizationQueue`?

joeyyang:
back to you!

client/app/queue/components/HearingBadge.jsx
lowellrex: I'm stumped too! I like the hypothesis we've got--that the absence of this condition caused us to make a request to `/appeals/undefined/hearings`--but let's do a little more work to determine the environment that causes us to fail that condition by reproducing this error locally.

PR #9308, opened 2019-02-08
Add fasterer gem to catch performance issues

monfresh:
**Why**:
In Ruby and in Rails, there are multiple ways of achieving of
the same result. However, they are not always equal in terms of
performance, and choosing the wrong one can have a significant impact.

For example, we recently had daily DB statement timeouts between January
22 and February 6 that caused the AWS RDS CPU utilization to increase by
about 5-10x the normal rate. This was caused by using `.count > 0` to
check if an ActiveRecord query contained any results. This query was on
the Documents table, which has been quickly growing in size. `COUNT`
queries in Postgres are known to be slow, and get slower as the table
grows.

If all we want to know is if the query returned any results at all, that
means we don't care about the actual count, and therefore we don't need
to perform a `COUNT` query. Instead, we can use
`ActiveRecord::Relation#any?` which performs a much faster SQL query
that uses `SELECT 1 AS one` with a `LIMIT` of 1.

The [fasterer](https://github.com/DamirSvrtan/fasterer) gem was inspired
by the [fast-ruby](https://github.com/JuanitoFatas/fast-ruby) repo,
which in turn was inspired by a great
[talk on Ruby performance](https://www.youtube.com/watch?v=fGFM_UrSp70)
by Erik Michaels-Ober.

It analyzes the code for slow idioms and lets you know which version is
faster. This is a great way to learn something new and to level up the
whole team. The gem doesn't currently track all of the examples in the
fast-ruby repo, but it does track some of the most common mistakes. I
plan on contributing to the gem to add more checks.

The 4 most common slow idioms we had were:

- `select`ing from an array and then calling `.first` or `.last` instead
of `detect`
- Using `Hash#merge!` instead of `Hash#[]=` when only merging one
element
- Sorting using the idiom `sort { |a, b| a.foo <=> b.foo }` instead of
`sort_by(&:foo)`. Relatedly, using `sort_by` and then `.first` or
`.last` instead of `min_by` and `max_by` respectively.
- Passing in a block argument and using `block.call` instead of just
calling `yield`.

In most cases, the faster code also had the additional advantage of making the
code easier to read and understand.

### DIFF
--- a/app/models/appeal.rb
+++ b/app/models/appeal.rb
@@ -144,7 +144,7 @@ class Appeal < DecisionReview
   end
 
   def reviewing_judge_name
-    task = tasks.order(:created_at).select { |t| t.is_a?(JudgeTask) }.last
+    task = tasks.order(created_at: :desc).detect { |t| t.is_a?(JudgeTask) }
     task ? task.assigned_to.try(:full_name) : ""
   end
 
--- a/app/models/appeal_history.rb
+++ b/app/models/appeal_history.rb
@@ -53,7 +53,7 @@ class AppealHistory
 
     # Each appeal tree gets a series ID; assign all of its descendant nodes that ID.
     appeal_tree_roots.each_with_index do |root, sid|
-      traverse_appeal_tree(root) { |node| node[:series_id] = sid }
+      traverse_appeal_tree(root, sid) { |node| node[:series_id] = sid }
 
       merge_table[sid] = sid
     end
@@ -124,9 +124,11 @@ class AppealHistory
     remove_instance_variable :@roots
   end
 
-  def traverse_appeal_tree(node, &block)
-    block.call(node)
-    node[:children].each { |child| traverse_appeal_tree(child, &block) }
+  def traverse_appeal_tree(node, sid)
+    yield(node)
+    node[:children].each do |child|
+      traverse_appeal_tree(child, sid) { |child_node| child_node[:series_id] = sid }
+    end
   end
 
   def find_parent_appeal(appeal)
--- a/app/models/appeal_series.rb
+++ b/app/models/appeal_series.rb
@@ -83,17 +83,11 @@ class AppealSeries < ApplicationRecord
     @issues ||= AppealSeriesIssues.new(appeal_series: self).all
   end
 
-  # rubocop:disable CyclomaticComplexity
   def description
     ordered_issues = latest_appeal.issues
       .select(&:codes?)
-      .sort do |a, b|
-        dc_comparison = (a.diagnostic_code.nil? ? 1 : 0) <=> (b.diagnostic_code.nil? ? 1 : 0)
-
-        next dc_comparison unless dc_comparison == 0
-
-        a.vacols_sequence_id <=> b.vacols_sequence_id
-      end
+      .sort_by(&:vacols_sequence_id)
+      .partition(&:diagnostic_code).flatten
 
     return "VA needs to record issues" if ordered_issues.empty?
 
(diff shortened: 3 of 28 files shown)


app/models/appeal.rb
```diff
@@ -133,7 +133,7 @@ def every_request_issue_has_decision?
   end
 
   def reviewing_judge_name
-    task = tasks.order(:created_at).select { |t| t.is_a?(JudgeTask) }.last
+    task = tasks.order(created_at: :desc).detect { |t| t.is_a?(JudgeTask) }
```
pkarman: interesting! I usually use `find` but didn't realize it also goes by the name `detect`

app/models/appeal_history.rb
```diff
@@ -124,9 +124,11 @@ def burn_appeal_trees
     remove_instance_variable :@roots
   end
 
-  def traverse_appeal_tree(node, &block)
-    block.call(node)
-    node[:children].each { |child| traverse_appeal_tree(child, &block) }
+  def traverse_appeal_tree(node, sid)
+    yield(node)
+    node[:children].each do |child|
+      traverse_appeal_tree(child, sid) { |child_node| child_node[:series_id] = sid }
```
pkarman: elegant!

app/models/appeal_series.rb
```diff
@@ -83,17 +83,11 @@ def issues
...
       .select(&:codes?)
-      .sort do |a, b|
-        dc_comparison = (a.diagnostic_code.nil? ? 1 : 0) <=> (b.diagnostic_code.nil? ? 1 : 0)
-
-        next dc_comparison unless dc_comparison == 0
-
-        a.vacols_sequence_id <=> b.vacols_sequence_id
-      end
+      .sort_by(&:vacols_sequence_id)
+      .partition(&:diagnostic_code).flatten
```
pkarman: wow this one is remarkable. I read over the tests for `#description` so I could have more confidence that the logic is preserved here. Seems right.

spec/models/appeal_series_spec.rb
```diff
@@ -448,11 +448,12 @@
         [
           create(:case_issue, issseq: 1, issprog: "02", isscode: "17", isslev1: "02"),
           create(:case_issue, issseq: 2, issprog: "02", isscode: "15", isslev1: "03", isslev2: "5252"),
-          create(:case_issue, issseq: 3, issprog: "02", isscode: "15", isslev1: "03", isslev2: "9432")
+          create(:case_issue, issseq: 3, issprog: "02", isscode: "12", isslev1: "05"),
```
pkarman: I assume you added this `:case_issue` to exercise the method a bit more?

leikkisa:
Thank you! Will keep my eyes open for these examples going forward.

app/models/appeal.rb
monfresh: Same here, but because fasterer and fast-ruby both refer to it as `detect`, that's what I used. I just checked though, and fasterer accepts `find` as well. I'm happy to change it to `find`. Rubocop has a config that allows you to specify which alias you prefer, but it's disabled by default because it can lead to false positives: https://github.com/rubocop-hq/rubocop/blob/master/manual/cops_style.md#stylecollectionmethods. I supposed we could enable it and disable false positives on a case-by-case basis.

spec/models/appeal_series_spec.rb
monfresh: Yeah, I wanted to make sure the ones with a diagnostic code were grouped together first, and the ones without grouped together at the end of the array.

PR #9312, opened 2019-02-08
Update closed_at timestamp when task is cancelled

lowellrex:
Connects #8701. Part 1 of 6 of the remaining work to do on this ticket. This ticket adds the `cancelled` status and makes sure we update the timestamp when we change a task to this status.


app/models/task.rb
```diff
@@ -392,12 +393,14 @@ def put_parent_on_hold
...
   def set_timestamps
     if will_save_change_to_status?
       self.assigned_at = updated_at if assigned?
       self.started_at = updated_at if in_progress?
       self.placed_on_hold_at = updated_at if on_hold?
-      self.closed_at = updated_at if completed?
+      self.closed_at = updated_at if completed? || cancelled?
     end
   end
+  # rubocop:enable CyclomaticComplexity
```
tomas-nava: you could avoid the `CyclomaticComplexity` flag by using a case statement, something like

```ruby
  def set_timestamps
    if will_save_change_to_status?
      case status_change_to_be_saved&.last&.to_sym
      when :assigned
        self.assigned_at = updated_at
      when :in_progress
        self.started_at = updated_at
      when :on_hold
        self.placed_on_hold_at = updated_at
      when :completed, :cancelled
        self.closed_at = updated_at
      end
    end
  end
```

PR #9315, opened 2019-02-08
Allow admins to reassign ColocatedTasks

### LINKED ISSUE #8607: Allow team admins to view and action on currently active tasks
Team admins (like branch chiefs, e.g. Rolandus) need to be able to reassign cases from one of their team members to another. So, when they click into a Case details page for a case assigned to one of their users, they need to see the currently active tasks assigned to their team/people in their team (even if it is not assigned to them) and be able to reassign them.

## AC
- When team admins view Case details page, tasks assigned to their team or people in their team populate in the Currently Active Tasks component
- Team admins can reassign tasks from one person on their team to another using the Actions drop down

## Open questions
- I don't think team admins need to be able to place the task on hold
- They likely will want to be able to complete tasks

lowellrex:
Resolves #8607. This change adds the option for administrators of the `Colocated` team to re-assign `ColocatedTask`s to that are assigned to members of the `Colocated` team to other members of the `Colocated` team.

Before | After
--- | ---
image | image


app/models/tasks/generic_task.rb
```diff
@@ -30,7 +30,7 @@ def available_actions(user)
       ]
     end
 
-    if task_is_assigned_to_user_within_organiztaion?(user)
+    if task_is_assigned_to_user_within_organization?(user)
```
lowellrex: Fix typo from an earlier PR.

app/models/tasks/colocated_task.rb
```diff
@@ -52,12 +58,12 @@ def available_actions(_user)
 
     actions
   end
+  # rubocop:enable Metrics/PerceivedComplexity
+  # rubocop:enable Metrics/CyclomaticComplexity
```
lowellrex: I'm up for refactoring this if we don't want to ignore some rubocop rules here, but I don't find this function overly complex. Curious to see if you do though!

app/models/tasks/colocated_task.rb
```diff
@@ -35,10 +35,16 @@ def verify_user_can_create!(user, parent)
   end
 
   # rubocop:disable Metrics/AbcSize
-  def available_actions(_user)
-    actions = [Constants.TASK_ACTIONS.PLACE_HOLD.to_h, Constants.TASK_ACTIONS.ASSIGN_TO_PRIVACY_TEAM.to_h]
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/PerceivedComplexity
+  def available_actions(user)
+    if assigned_to != user
+      return [Constants.TASK_ACTIONS.REASSIGN_TO_PERSON.to_h] if Colocated.singleton.admins.include?(user)
```
lowellrex: Since this change is limited to only `ColocatedTask`s we can use the `Colocated` organization directly instead of the `task_is_assigned_to_user_within_organization?` method (which is defined in the `GenericTask` model).

db/seeds.rb
```diff
@@ -594,15 +590,7 @@ def create_colocated_legacy_tasks(attorney, colocated_user)
...
-      org_task = FactoryBot.create(:colocated_task, :on_hold, org_task_args)
-
-      personal_task_args = org_task_args.merge(
-        parent: org_task,
-        action: org_task.action,
-        instructions: org_task.instructions,
-        assigned_to: colocated_user
-      )
-      FactoryBot.create(*[:colocated_task, attrs[:trait]].compact, personal_task_args)
+      FactoryBot.create(:colocated_task, :on_hold, org_task_args)
```
lowellrex: Since we've offloaded automatic assignment of `ColocatedTask`s [to the `Colocated` organization](https://github.com/department-of-veterans-affairs/caseflow/blob/89d6f74b715213b07933c6358e9b0ee895ba6e60/app/models/organizations/colocated.rb#L6) we were double assigning these tasks (once by the [automatic task assignment](https://github.com/department-of-veterans-affairs/caseflow/blob/89d6f74b715213b07933c6358e9b0ee895ba6e60/app/models/organization.rb#L29) in the `Colocated` organization, and once here in the seeds file).

app/models/task.rb
```diff
@@ -199,6 +199,31 @@ def self.verify_user_can_create!(user, parent)
     end
   end
 
+  def reassign(reassign_params, current_user)
```
lowellrex: Moved from `GenericTask` into the base `Task` class so `ColocatedTask` has access to it.

app/models/task.rb
```diff
@@ -199,6 +199,31 @@ def self.verify_user_can_create!(user, parent)
...
+    sibling = dup.tap do |t|
+      t.assigned_by_id = self.class.child_assigned_by_id(parent, current_user)
+      t.assigned_to = self.class.child_task_assignee(parent, reassign_params)
+      t.instructions = [instructions, reassign_params[:instructions]].flatten
+      t.save!
+    end
```
lowellrex: Only [change in this method](https://github.com/department-of-veterans-affairs/caseflow/pull/9315/files#diff-9f7b57df79bdb2c539ae04506d0f5a67L51) is changing how we create the new sibling task from a `create_child_task` call to a `dup` call. Did this for a few reasons:

1. `create_child_task` does not set the `action` field which `ColocatedTask` [needs to have set](https://github.com/department-of-veterans-affairs/caseflow/blob/b93da4a081f4d36339a5471ce61bd5fb82013b17/app/models/tasks/colocated_task.rb#L2).
1. Don't have to explicitly set a whole bunch of fields (type, appeal, parent).
1. Don't have to move `create_child_task` into the `Task` model.

All of the other reasons for changing this from `create_child_task` to `dup` could be solved (mostly by just moving those functions to the `Task` model), but solving the first problem (setting the `action` field) would have required a somewhat cumbersome refactor of this function as well as the `create_child_task` method (though maybe it would be as simple as just passing `action` as a parameter and setting it... but we are trying to phase out that field!). Curious to hear what you think!

app/models/task.rb
```diff
@@ -199,6 +199,31 @@ def self.verify_user_can_create!(user, parent)
...
+
+    update!(status: Constants.TASK_STATUSES.completed)
+
+    children_to_update = children.reject { |t| t.status == Constants.TASK_STATUSES.completed }
+    children_to_update.each { |t| t.update!(parent_id: sibling.id) }
+
+    [sibling, self, children_to_update].flatten
+  end
+
+  def self.child_task_assignee(_parent, params)
```
lowellrex: Moved from `GenericTask` into the base `Task` class so `ColocatedTask` has access to it.

app/models/task.rb
```diff
@@ -199,6 +199,31 @@ def self.verify_user_can_create!(user, parent)
...
+    Object.const_get(params[:assigned_to_type]).find(params[:assigned_to_id])
+  end
+
+  def self.child_assigned_by_id(parent, current_user)
```
lowellrex: Moved from `GenericTask` into the base `Task` class so `ColocatedTask` has access to it.

app/models/task.rb
```diff
@@ -170,12 +168,14 @@ def when_child_task_completed
     update_status_if_children_tasks_are_complete
   end
 
-  def can_be_updated_by_user?(user)
-    return true if [assigned_to, assigned_by].include?(user) ||
-                   parent&.assigned_to == user ||
-                   user.administered_teams.select { |team| team.is_a?(JudgeTeam) }.any?
+  def task_is_assigned_to_user_within_organization?(user)
```
lowellrex: Moved from `GenericTask` into the base `Task` class so `ColocatedTask` has access to it.

app/models/task.rb
```diff
@@ -170,12 +168,14 @@ def when_child_task_completed
...
+    parent&.assigned_to.is_a?(Organization) &&
+      assigned_to.is_a?(User) &&
+      parent.assigned_to.user_has_access?(user)
+  end
 
-    false
+  def can_be_updated_by_user?(user)
+    available_actions_unwrapper(user).any?
```
lowellrex: Lifted this directly from the `GenericTask` logic. The only change here is that all judges no longer have the ability to modify all `AttorneyTask`s, `ColocatedTask`s, and `JudgeTask`s (all `Task` classes that do not inherit from `GenericTask`), only those that they can take action on.

The exception for judges was introduced in PR #7329, but has probably been broken for some time since judges are not presented any options that could trigger the modification of any tasks for a while, so I think this change is safe to make. (Slack thread with [some related discussion here](https://dsva.slack.com/archives/C6E41RE92/p1549899413007800))

spec/models/task_spec.rb
```diff
@@ -104,29 +104,14 @@
...
-      let(:task) { create(:task, type: "Task", assigned_to: user) }
+      let(:task) { create(:generic_task, assigned_to: user).becomes(GenericTask) }
 
       it { is_expected.to be_truthy }
     end
 
     context "when user does not have access" do
       let(:user) { create(:user) }
-      let(:task) { create(:task, type: "Task", assigned_to: create(:user)) }
+      let(:task) { create(:generic_task, assigned_to: create(:user)) }
```
lowellrex: Need to make this change so that there we have some default set of task actions instead of none.

spec/models/task_spec.rb
```diff
@@ -104,29 +104,14 @@
...
-      let(:user) { create(:user) }
-      let(:parent) { create(:task, type: "Task", assigned_to: user) }
-      let(:task) { create(:task, type: "Task", parent: parent) }
-
-      it { is_expected.to be_truthy }
-    end
-
-    context "when user is any judge" do
-      let(:user) { create(:user, css_id: "BVABDANIEL") }
-      let(:task) { create(:task, type: "Task", assigned_to: user) }
```
lowellrex: We are now relying on the presence of task actions to determine whether this is true or false so these tests are no longer applicable.

app/models/tasks/colocated_task.rb
tomas-nava: If I were to do anything, I'd break out lines 51-61 into a separate method, but I agree, it's fine as-is.

lowellrex: Made an attempt at it here: https://github.com/department-of-veterans-affairs/caseflow/pull/9315/commits/ac17f5f98a5c17e4fb11d77b3567b2cbaac43fba

Is that along the lines of what you were thinking?

lowellrex:
> Is this already the case? No changes needed?

Correct!

> > "They likely will want to be able to complete tasks"
> 
> This is categorized as an "Open question" so not necessarily a requirement for this PR, but we should confirm with LP.

@laurjpeterson confirms that scoping this PR to only the reassign action is fine for now in [this slack discussion](https://dsva.slack.com/archives/C6E41RE92/p1550073486099100).

app/models/tasks/colocated_task.rb
tomas-nava: yep that works great!

Issue #9317, opened 2019-02-08
Quick copy change | Update VRE long name in benefit types


steve-gov:
In https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/constants/BENEFIT_TYPES.json:

`"voc_rehab": "Vocational Rehab. & Employment",`

becomes

`"voc_rehab": "Vocational Rehabilitation and Employment",`

**ETA**: In addition, please replace the string in the same manner anywhere it is stored in the DB.

**ETA**: ...in both prod and UAT.

pkarman:
Could we also ditch the ampersand? "Vocational Rehabilitation and Employment"

steve-gov:
Works for me!

steve-gov:
Edited criteria to add: Please update instances of the string in the DB as well.

kencheeto:
Going to run this bit of code in a prod rails console:
```ruby
voc_rehab = BusinessLine.find_by_url("voc_rehab")
voc_rehab.update!(name: "Vocational Rehabilitation and Employment")
```

kencheeto:
Approved in Slack: https://dsva.slack.com/archives/C754H751T/p1550083738110500

and done! @steve-gov could you refresh the page? it should be updated now.

steve-gov:
Edited criteria to add: Please update instances of the string in the UAT DB as well.

kencheeto:
Updated in UAT as well!

steve-gov:
<img width="600" alt="screen shot 2019-02-14 at 9 00 40 am" src="https://user-images.githubusercontent.com/25468689/52803496-082cd080-3037-11e9-98bf-0bb9c22c7104.png">

PR #9321, opened 2019-02-09
update voc_rehab string

kencheeto:
Connects #9317


pkarman:
Can we spell it out "Vocational Rehabilitation and Employment" per their website?

PR #9326, opened 2019-02-10
Appeal Status: Adding issues attribute for HLR, SC, and Appeal

### LINKED ISSUE #9160: API: SCs, HLRs, and Appeals have issues
```
"issues": [
  {
    "active": false,
    "lastAction": "allowed",
    "date": "2019-05-01",
    "description": "Tuberculosis of bones and joints",
    "diagnosticCode": "5001"
  },
  {
    "active": true,
    "lastAction": "remand",
    "date": "2019-05-01",
    "description": "Pension issue",
    "diagnosticCode": null
  }
]
```

* An issue begins life as a RequestIssue. When that RequestIssue has one or more DecisionIssues, we use the DecisionIssues instead.
* `lastAction` and `date` refer to the last time the disposition of the issue changed, generally by a decision.
  * If the issue is still active, both `lastAction` and `date` are null.
  * The enum values for `lastAction` are "allowed" (granted), "denied", "remand", and "withdrawn".
  * For Supplemental Claims and Higher-Level Reviews, we do not show the `lastAction` and `date` until the SC or HLR is inactive. This is because a HLR might have a DTA error, and the decision will not be sent to the Veteran until that DTA error is resolved.
  * For Appeals, the `lastAction` and `date` are shown once a decision is dispatched. Note that the `lastAction` might be "remand". This should become "allowed" or "denied" once the AOJ makes a new decision (this should also update the `date` to the date of the AOJ's decision).
  * If an issue is withdrawn, that can be shown immediately.
* `active` is true if the `lastAction` is null or "remand", false otherwise.
* `diagnosticCode` is the diagnostic code of a rating issue, if known.
* `description` is a string describing the issue.
  * If there is a diagnostic code, use the `status_description` from `DIAGNOSTIC_CODE_DESCRIPTIONS`, but capitalize the first letter.
  * If there is not a diagnostic code, but there's a description of the issue available, we might want to use that. If this is free text entered by a VA employee, we don't want to use it.
  * The last resort is to use the claim type of the issue, so `pension` would become "Pension issue".

mc-usds:
Resolves #9160

### Description
Adding the issues hash attribute for HLR, SC and Appeals

### Acceptance Criteria
- [x] Code compiles correctly

### Testing Plan
1. added unit tests and updated feature test


app/models/appeal.rb
```diff
@@ -574,5 +595,24 @@ def contestable_decision_issues
         issue.approx_decision_date && issue.approx_decision_date < receipt_date
       end
   end
+
+  def issue_active_status(issue)
```
cmgiven: Having these methods on Appeal is probably Bad™. I was going suggest an "obvious" solution, and then I slowly realized the good reasons why you did it this way. This is the sort of problem that Shane would probably immediately find the Ruby way of dealing with. He's in town this week, and it might be worth getting his attention on this for a half hour.

But I'm also in a mode to not want to be slowing you down, and this seems pretty thorny, so I don't think this should block anything.

app/models/appeal.rb
```diff
@@ -532,6 +532,27 @@ def events
...
 
+  def issues_hash
+    issue_list = decision_issues.empty? ? request_issues.open : fetch_all_decision_issues
+
+    fetch_issues_status(issue_list)
+  end
+
+  def fetch_all_decision_issues
+    return decision_issues unless remanded_issues?
+    return decision_issues if remanded_issues? && active_remanded_claims?
```
cmgiven: Even having sketched this out with you, it took me a couple minutes to reason my way to this being correct. A couple comments might be useful to explain why we're doing it this way.

app/models/decision_review.rb
```diff
@@ -312,4 +312,38 @@ def validate_receipt_date
...
+
+  def get_issue_last_action_date(issue)
+    return if active?
+
+    issue.approx_decision_date if issue.is_a?(DecisionIssue)
+  end
+
+  def get_issue_status_description(issue)
+    if issue.diagnostic_code && Constants::DIAGNOSTIC_CODE_DESCRIPTIONS[issue.diagnostic_code]
+      Constants::DIAGNOSTIC_CODE_DESCRIPTIONS[issue.diagnostic_code]["status_description"].capitalize
```
cmgiven: Sadly, we can't use `capitalize` here, because it will downcase the remaining letters. There are a lot of medical conditions where multiple words need to be capitalized. Stack Overflow suggestion here: https://stackoverflow.com/a/33617474

app/models/higher_level_review.rb
```diff
@@ -59,9 +59,18 @@ def alerts
     # need to implement. add logic to return alert enum
   end
 
-  def issues
-    # need to implement. get request and corresponding rating issue
-    []
+  def issues_hash
+    issue_list = active? ? request_issues.open : fetch_all_decision_issues
```
cmgiven: If `.open` does what I think it does, it might be wrong? Might this be removing request issues from the list preemptively?

app/models/decision_review.rb
mc-usds: Should I only capitalize the first letter of the first word?
I was looking through DIAGNOSTIC_CODE_DESCRIPTIONS and for almost all of them the whole string is lower case.  The exceptions are:

"status_description": "HIV-related illness"
"status_description": "Hodgkin's disease"
"status_description": "non-Hodgkin's lymphoma"
"status_description": "C-cell hyperplasia"
"status_description": "Parkinson's disease"
"status_description": "Sydenham's chorea"
"status_description": "Alzheimer's disease"

app/models/appeal.rb
mc-usds: Yeah, this whole PR is kinda gnarly.  I went back and forth about either putting the logic in appeal or putting the logic in request_issue and decision_issue respectively. I think I ended up going with putting it in appeal is because it keep just about all the changes at the decision_review level.

Looking back through the code, there is precedent to put it at the issue level:
https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/request_issue.rb#L228

Putting it on request_ and decision_ issue would align better with the current code.

app/models/higher_level_review.rb
mc-usds: Yes, open will filter out issues that have been closed.

app/models/appeal.rb
mc-usds: Will add comments for this method.

app/models/decision_review.rb
cmgiven: That is fewer than I remembered. Of that list, only "HIV-related illness" and "non-Hodgkin's lymphoma" would be problems.

app/models/appeal.rb
cmgiven: :+1:

app/models/appeal.rb
```diff
@@ -528,6 +528,31 @@ def events
...
+
+  def fetch_all_decision_issues
+    return decision_issues unless remanded_issues?
+    # only include the remanded issues they are still being worked on
+    return decision_issues if remanded_issues? && active_remanded_claims?
+
+    # if there were remanded issues and there is a decision available
+    # for them, include the decisions from the remanded SC and do not
+    # include the original remanded decision
+    di_list = decision_issues.reject { |di| di[:disposition] == "remanded" }
```
pkarman: this seems ripe for adding a method to `DecisionIssue`

app/models/appeal.rb
```diff
@@ -528,6 +528,31 @@ def events
...
+
+    remand_supplemental_claims.each do |sc|
+      sc.decision_issues.each do |di|
+        di_list << di
+      end
+    end
+
+    di_list
```
pkarman: should this have a `.uniq` added to it? any way there could be overlap?

app/models/decision_issue.rb
```diff
@@ -83,6 +83,24 @@ def contention_text
     Contention.new(description).text
   end
 
+  def issue_status_active?
```
pkarman: a nit about the naming here. The word "issue" is so overloaded in this app that I'm not sure what these methods refer to. The decision issue? the contested issue? the request issue? If it's the decision issue then I don't think the `issue_` prefix is necessary.

app/models/higher_level_review.rb
```diff
@@ -55,9 +55,18 @@ def alerts
...
+
+    fetch_issues_status(issue_list)
+  end
+
+  def fetch_all_decision_issues
+    all_decision_issues = decision_issues.reject { |di| DTA_ERRORS.include?(di[:disposition]) }
```
pkarman: as noted above, this `reject` call might be better implemented as a class method scope on `DecisionIssue`

app/models/higher_level_review.rb
```diff
@@ -55,9 +55,18 @@ def alerts
...
+
+    all_decision_issues += dta_claim.decision_issues if dta_claim
+
+    all_decision_issues
```
pkarman: since we're concatenating arrays here, might this need a `.uniq` ?

app/models/request_issue.rb
```diff
@@ -415,6 +415,25 @@ def decision_or_promulgation_date
     return contested_rating_issue.try(:promulgation_date) if rating?
   end
 
+  def diagnostic_code
+    contested_rating_issue_diagnostic_code
+  end
+
+  def issue_status_active?
```
pkarman: same comment here about naming. the `issue_` prefix confuses me a little.

app/models/supplemental_claim.rb
```diff
@@ -49,9 +49,10 @@ def alerts
     # need to implement. add logic to return alert enum
   end
 
-  def issues
-    # need to implement. get request and corresponding rating issue
-    []
+  def issues_hash
```
pkarman: this seems identical to the `HigherLevelReview` method. Maybe it belongs on the superclass `ClaimReview`?

spec/models/appeal_spec.rb
```diff
@@ -1139,4 +1139,130 @@
       end
     end
   end
+
+  context "#item_hash" do
```
pkarman: is `#item_hash` a typo?

mc-usds: d'oh yeah, that's a typo fixed it.

app/models/supplemental_claim.rb
mc-usds: good catch, moved to ClaimReview.

app/models/decision_issue.rb
mc-usds: Will use "api_status" in place of "issue" in the naming.

app/models/appeal.rb
mc-usds: Thanks for pairing.  Made this a class method in DecisionIssues

app/models/higher_level_review.rb
mc-usds: Good catch. Added .uniq

PR #9328, opened 2019-02-10
Appeal Status: Adding status description for HLR, SC and Appeal

### LINKED ISSUE #9161: API: SCs, HLRs, and Appeals have a description attribute
```
"description": "Osteomyelitis and 4 others"
```

* If at least one issue has a diagnostic code, we will use the `status_description` found in `DIAGNOSTIC_CODE_DESCRIPTIONS` of the first issue with a diagnostic code as the "marquee issue", i.e. osteomyelitis above. Note that we must capitalize the first letter. (Note similarity to #9160.) If there is more than one issue, we will add text like "and 1 other" or "and 4 others" where the number is the total number of issues minus one (because we don't include the marquee issue in the count).
* If no issue has a diagnostic code, but all issues are of the same claim type, we will create a string of the format: "1 education issue" or "2 pension issues".
* If the claim type is other or there are multiple claim types, we will create a string of the format: "1 issue" or "2 issues".

mc-usds:
Resolves #9161

### Description
Adding implementation for the description attribute.

### Acceptance Criteria
- [x] Code compiles correctly

### Testing Plan
1. updated feature test

### DIFF
--- a/app/models/decision_review.rb
+++ b/app/models/decision_review.rb
@@ -335,8 +335,9 @@ class DecisionReview < ApplicationRecord
     issue_diagnostic_code = issue.contested_rating_issue_diagnostic_code if issue
 
     if issue_diagnostic_code && Constants::DIAGNOSTIC_CODE_DESCRIPTIONS[issue_diagnostic_code]
-      description = Constants::DIAGNOSTIC_CODE_DESCRIPTIONS[issue_diagnostic_code]["status_description"].capitalize
-
+      description = Constants::DIAGNOSTIC_CODE_DESCRIPTIONS[issue_diagnostic_code]["status_description"]
+      description[0] = description[0].upcase
+      
       return description if request_issues.count - 1 == 0
 
       return "#{description} and 1 other" if request_issues.count - 1 == 1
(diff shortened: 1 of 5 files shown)


app/models/decision_review.rb
```diff
@@ -312,4 +312,44 @@ def validate_receipt_date
...
+
+  def fetch_status_description_using_diagnostic_code
+    issue = request_issues.find do |ri|
+      !ri[:contested_rating_issue_diagnostic_code].nil?
+    end
+
+    issue_diagnostic_code = issue.contested_rating_issue_diagnostic_code if issue
+
+    if issue_diagnostic_code && Constants::DIAGNOSTIC_CODE_DESCRIPTIONS[issue_diagnostic_code]
+      description = Constants::DIAGNOSTIC_CODE_DESCRIPTIONS[issue_diagnostic_code]["status_description"].capitalize
```
cmgiven: Same annoying comment about capitalization as on the other PR. Sorry.

mc-usds: No problem. I changed it to only upper case the first letter in the string, the rest of the string is left as is.

PR #9338, opened 2019-02-11
Update README.md

youngfreezyVA:
update readme with info for connecting to DB with Psequel


README.md
```diff
@@ -239,6 +239,11 @@ bundle exec rails dbconsole # password is `postgres`
 
 To connect to FACOLS, we recommend using [SQL Developer](https://www.oracle.com/database/technologies/appdev/sql-developer.html). Connection details can be found in the docker-compose.yml file.
 
+You can also use Psequel (instead of SQL Developer) with the following setup (user and password is postgres):
+
+https://www.dropbox.com/s/zkf8w85yfyjejwl/Screenshot%202019-02-11%2012.19.42.png?dl=0
```
pkarman: any chance we could copy that screen shot to GH so that we're not pointing at an external reference that might some day disappear?

Issue #9297, opened 2019-02-07
Intake | Show the promulgation date and not the profile date on the add issues modal


shanear:
We've been informed by the business to show the promulgation date and not the profile date on the Add Issues modal. Should be a straightforward fix.

pkarman:
@shanear should the DecisionIssue approx decision date use `promulgation_date` as well? see question in https://github.com/department-of-veterans-affairs/caseflow/pull/9345#pullrequestreview-202350148

pkarman:
Testing notes: find a rating with different profile and promulgation dates and verify the promulgation date is the visible "Decision Date" in the Add Issues modal.

steve-gov:
Validated in UAT that rated issue shows promulgated date instead of profile date:

<img width="241" alt="screen shot 2019-02-15 at 12 03 18 pm" src="https://user-images.githubusercontent.com/25468689/52881550-bad55f80-3119-11e9-9462-b35cb135d2f4.png">

<img width="542" alt="screen shot 2019-02-15 at 12 03 11 pm" src="https://user-images.githubusercontent.com/25468689/52881552-bc9f2300-3119-11e9-947a-859f07229265.png">

PR #9345, opened 2019-02-11
Use rating issue promulgation_date on ContestableIssue

pkarman:
connects #9297 

### Description
The Add Issues modal on intake should display the promulgation date for existing rating issues.


spec/models/contestable_issue_spec.rb
```diff
@@ -20,7 +20,6 @@
       rba_contentions_data: [{}]
     )
   end
-  let(:profile_date) { Time.zone.today }
```
pkarman: this was defined twice

spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -41,6 +41,8 @@
 
   let(:receipt_date) { Date.new(2018, 9, 20) }
 
+  let(:promulgation_date) { receipt_date - 10.days }
```
pkarman: the old definition had this equal to the profile_date.

spec/models/contestable_issue_spec.rb
leikkisa: Cleanin up!

spec/models/contestable_issue_spec.rb
```diff
@@ -60,15 +59,15 @@
     end
 
     context "is untimely" do
-      let(:profile_date) { Time.zone.today - 373.days }
```
leikkisa: Ooh, timeliness shouldn't have been based on profile_date in the first place.

spec/models/decision_review_spec.rb
```diff
@@ -76,44 +76,54 @@
...
-        },
+      serialized_contestable_issues = subject.map(&:serialize)
+
+      expect(find_serialized_issue(serialized_contestable_issues, "123")).to eq(
+        # this rating issue got replaced with a decision issue
+        ratingIssueReferenceId: "123",
+        ratingIssueProfileDate: profile_date,
+        ratingIssueDisabilityCode: nil,
+        decisionIssueId: decision_issues.first.id,
+        date: profile_date,
```
leikkisa: Maybe for another PR but I would love to rename this field with what kind of date it's supposed to be. That said, should this be promulgation_date?

spec/models/decision_review_spec.rb
```diff
@@ -76,44 +76,54 @@
...
+        sourceReviewType: nil,
+        timely: true
+      )
+
+      expect(find_serialized_issue(serialized_contestable_issues, "789")).to eq(
+        ratingIssueReferenceId: "789",
+        ratingIssueProfileDate: profile_date + 1.day,
+        ratingIssueDisabilityCode: nil,
+        decisionIssueId: decision_issues.second.id,
+        date: profile_date + 1.day,
```
leikkisa: Should this be promulgation_date?

pkarman:
@leikkisa you asked

> if it's being used in the front end as the decision date it sounds like yes. If it's being used to convey the actual decision date then I guess no?

I'm not sure, TBH, but I think yes.

That will change how the tests work for where you asked about profile-vs-promulgation.

pkarman:
@leikkisa I think I hit all your comments. LMK.

app/models/request_issue.rb
```diff
@@ -298,6 +298,16 @@ def ui_hash
     }
   end
 
+  def approx_decision_date
+    if contested_issue
+      contested_issue.approx_decision_date
+    elsif decision_date
+      decision_date
+    elsif decision_issues.any?
```
pkarman: adding this bit of logic since tests revealed it is possible (from a schema perspective) for a request issue to have neither a contested_issue nor a decision_date but to have decision_issues associated.

client/app/intake/util/issues.js
```diff
@@ -314,22 +315,23 @@ export const formatAddedIssues = (intakeData, useAmaActivationDate = false) => {
...
     } else if (issue.isRating) {
-      // todo: date works for contestable issue
-      // and profile_date works for request issue (for the edit page)
-      // fix this to use same keys
-      const profileDate = new Date(issue.date || issue.profileDate);
+      if (!issue.decisionDate && !issue.approxDecisionDate) {
+        console.warn(issue);
+        throw new Error('no decision date');
+      }
+      const decisionDate = new Date(issue.decisionDate || issue.approxDecisionDate);
```
pkarman: I don't *think* we ever have a case where, for a rating issue, `issue.approxDecisionDate` is null but we have a `issue.decisionDate` so this is mostly defensive.

spec/factories/decision_issue.rb
```diff
@@ -4,7 +4,8 @@
     disposition "allowed"
     benefit_type "compensation"
     diagnostic_code "9999"
-    end_product_last_action_date 5.days.ago.to_date
+    end_product_last_action_date { decision_review.is_a?(Appeal) ? nil : 5.days.ago.to_date }
```
pkarman: the presence/absence of these 2 date attributes is dependent on the decision_review type, but we should have one or the other by default or tests will fail.

client/app/intake/util/issues.js
```diff
@@ -314,22 +315,23 @@ export const formatAddedIssues = (intakeData, useAmaActivationDate = false) => {
         isUnidentified: true
       };
     } else if (issue.isRating) {
-      // todo: date works for contestable issue
-      // and profile_date works for request issue (for the edit page)
-      // fix this to use same keys
-      const profileDate = new Date(issue.date || issue.profileDate);
+      if (!issue.decisionDate && !issue.approxDecisionDate) {
+        console.warn(issue);
+        throw new Error('no decision date');
```
pkarman: we don't want tests to pass unless the data is legit, so throwing this error here. There are zero cases where we want to allow presentation of a rating request issue w/o a date.

spec/feature/intake/intake_edit_confirmation_spec.rb
```diff
@@ -9,34 +9,40 @@
...
+        ]
       )
     end
 
-    let(:request_issue) do
+    let!(:request_issue) do
       create(:request_issue,
-             rating_issue_reference_id: "def456",
-             rating_issue_profile_date: rating.profile_date,
+             contested_rating_issue_reference_id: rating_reference_id,
```
pkarman: we were still using the deprecated attribute names here so that rating lookup logic was failing (silently till now).

spec/feature/intake/intake_edit_confirmation_spec.rb
```diff
@@ -9,34 +9,40 @@
...
+             contested_rating_issue_profile_date: rating.profile_date,
              review_request: decision_review,
              description: "PTSD denied")
     end
 
     before do
       decision_review.create_issues!([request_issue])
       decision_review.establish!
+      decision_review.reload
```
pkarman: this fetches the uuid assigned by psql for easier url construction.

app/models/decision_issue.rb
```diff
@@ -131,13 +131,11 @@ def processed_in_caseflow?
     decision_review.processed_in_caseflow?
   end
 
-  def claim_review_approx_decision_date
-    # there's an end_product_last_action_date when decision issues are created from eps
-    # but only a promulgation date when decision issues are created from noncomp
-    if profile_date
-      profile_date.to_date
+  def processed_in_vbms_decision_date
```
shanear: I think this would be a good place for a comment. Also why'd we not qualify this with `approx` too?

app/models/request_issue.rb
```diff
@@ -298,6 +298,16 @@ def ui_hash
     }
   end
 
+  def approx_decision_date
```
shanear: This method is mixing two different concepts.

lines 302-305 are retrieving the decision date of the contested issue.

line 306 is retrieving the decision date of the decision of this issue.

Obv that is confusing. Maybe renaming this `approx_decision_date_of_contested_issue` would help alleviate that?

pkarman: previously it was just labeled `decision_date` in `ui_hash`. All I did was move the logic to a method.

Most accurately it would be `approx_decision_date_of_contested_issue_or_nonrating_decision_date_or_maybe_decision_issue`
which I would be just as happy with. ;)

shanear: `nonrating_decision_date` is the decision date of the contested issue (if it's nonrating)

pkarman:
Thanks @shanear -- took a stab at comments/renaming in 94cacc0

pkarman:
> I'm feeling much better about this!
> 
> I think this might need to be updated to use approx_decision_date too:
> In noncomp/components/BoardGrant.jsx
> let issueDate = formatDateStr(issue.decisionIssue.promulgationDate);

I will catch that in a follow up!

Issue #9303, opened 2019-02-08
Decision Sync | DTA Supplemental Claim for second EP will create excessive issues


shanear:
When the second EP of an HLR is cleared, if both the first and the second EP had DTA errors, then the second Supplemental Claim will be created with all of the issues from the first HLR. This is incorrect.

### Validation
- Create a HLR with a rating and nonrating request issue
- Have Colin DTA error both of the issues.  Have him clear the end products at different times (a day apart if possible).
- Check that that the issue one the first one that got cleared does not have a duplicate new request issue on a DTA supplemental claim. They should both have one new request issue (can be on one, or two different supplemental claims).

- Can probably combine this with #9142 since they both require a similar set up

leikkisa:
I propose adding a link to the end product establishment remanded in Supplemental Claim.  Otherwise, an HLR could have two DTA Supplemental claims, both that look somewhat identical except the issues on them.  Kind of like how Appeal can have multiple DTA SCs which are separated by the Benefit Type, HLRs can have multiple DTA SCs which are separated by the end product establishment.

This might also allow us to refactor some more of the remand code to be shared between HLRs and Appeals (the appeals code does already check that an issue isn't already being contested).

leikkisa:
This ended up being harder than expected due to parallel development with the API status, which was using the 1:1 paradigm for dta claims. I didn't do anything with connecting the EPE (mentioned above), I'll leave that for the actual ticket that needs that (also in case there's a different approach).

This might be hard to validate. If the original Higher Level Review has two end products and they get decided at different enough times (for them to get synced separately) or get synced separately somehow. Then check that the DTA supplemental claim doesn't have any duplicate issues.

Sjones352:
Combined with #9142

Sjones352:
Colin validated #9142 and it passes. 
<img width="462" alt="Screen Shot 2019-03-20 at 11 47 57 AM" src="https://user-images.githubusercontent.com/23080951/54699113-d069ea80-4b06-11e9-8f28-4fe39adf367c.png">

Issue #8842, opened 2019-01-20
API: Appeal statuses


cmgiven:
An Appeal should have a status attribute:

```
"status": {
  "type": "pending_hearing_scheduling",
  "details": {}
}
```

This issue concerns only the `type` of the status. For each `type` that has a non-empty `details` object, there will be a separate issue.

* **Appeals before decisions**
  * If the appeal has an open task to schedule a hearing, and the hearing request has not been  [~~waived~~](https://dsva.slack.com/archives/C3EAF3Q15/p1548956752003300) withdrawn, the type is `pending_hearing_scheduling`.
  * If the hearing has been scheduled but has not yet occurred, the type is `scheduled_hearing`.
  * If there is an open task to hold the appeal for 90 days while the Veteran adds new evidence, the type is `evidentiary_period`.
  * If there is an open task for a VSO to write an informal hearing presentation (IHP), the type is `at_vso`. (Until we have completed steps discussed in #8288, we are not properly closing expired IHP tasks)
  * If none of the above are true, and the appeal has yet to be distributed to a judge, the type is `on_docket`.
  * If the appeal has been distributed to a judge, but not yet decided, the type is `decision_in_progress`.
* **Appeals after decisions**
  * If the Board decision did not contain a remanded issue, the type is `bva_decision`.
  * If the Board decision did contain at least one remanded issue, the type is `ama_remand`.
  * If the Board decision contained a remand, and an EP was created in VBMS to track work on that remand, and that EP has now been closed, the status is `post_bva_dta_decision`.
  * If the Board decision did not contain a remand but did contain a grant, and an EP was created in VBMS to track work on effectuating that grant, and that EP has now ben closed, the status is `bva_decision_effectuation`.
* **Other closed statuses**
  * If the Veteran withdrew their appeal before a decision, the status is `withdrawn.` (Until #8936 is complete Caseflow does not have a mechanism for recognizing appeals as withdrawn)
  * If the appeal was closed without a decision for any other reason, the status is `other_close.`

PR #9353, opened 2019-02-11
Set up higher level reviews to have many remand supplemental claims

leikkisa:
connects #9303

Combines the HLR DTA code with Appeal code, and ensures that already contested issues DTA errors aren't duplicated.  Fixes bugs related to supporting 1:many relationship with DTA claims.


app/models/appeal.rb
```diff
@@ -496,8 +483,8 @@ def remanded_issues?
     decision_issues.any? { |di| di.disposition == "remanded" }
   end
 
-  def remanded_sc_with_ep
-    @remanded_sc_with_ep ||= remand_supplemental_claims.find(&:processed_in_vbms?)
+  def active_remanded_claims_with_ep?
+    active_remanded_claims&.select(&:processed_in_vbms?)&.any?
```
leikkisa: @mc-usds Are you sure you only want this for active remanded claims that were processed in VBMS here?

spec/models/appeal_spec.rb
```diff
@@ -1075,14 +1080,29 @@
...
+      end
       let!(:remanded_ep) do
         create(:end_product_establishment,
-               :cleared, source: remanded_sc, last_synced_at: receipt_date + 100.days)
+               :cleared, source: remanded_sc_with_ep, last_synced_at: receipt_date + 100.days)
       end
       let!(:remanded_sc_decision) do
         create(:decision_issue,
-               decision_review: remanded_sc,
+               decision_review: remanded_sc_with_ep,
```
leikkisa: @mc-usds This test is failing (line 1117 a little below) because in order to get the right status, it needs to have an active remand with an EP and this one is cleared. If I create an active EP on it, then I think the aojDecisionDate part will fail.

mc-usds: Sorry, I'm having a hard time telling which status this is testing. Could we pair on this one?

It looks like it testing one of the post decision statuses and in that case I think all the EPs would be cleared.

mc-usds: To provide some context: Here were the conditions I was trying to cover post decision, from
https://github.com/department-of-veterans-affairs/caseflow/issues/8842

**Appeals after decisions**
* If the Board decision did not contain a remanded issue, the type is `bva_decision`.
* If the Board decision did contain at least one remanded issue, the type is `ama_remand`.
* If the Board decision contained a remand, and an EP was created in VBMS to track work on that remand, and that EP has now been closed, the status is `post_bva_dta_decision`.
* If the Board decision did not contain a remand but did contain a grant, and an EP was created in VBMS to track work on effectuating that grant, and that EP has now ben closed, the status is `bva_decision_effectuation`.

app/models/appeal.rb
mc-usds: remanded_sc_with_ep was being used to see if there were any remanded SC processed in VBMS at all.

If there is a remand that is no processed in VBMS, is it processed in Caseflow? Is there a way to tell if it is active?

leikkisa: Ok great. That's actually another question I had for someone else, whether active status could be determined by whether there were incompleted tasks left, but I believe the answer is yes.

app/models/appeal.rb
```diff
@@ -333,13 +333,6 @@ def root_task
     RootTask.find_by(appeal_id: id)
   end
 
-  def create_remand_supplemental_claims!
```
leikkisa: Moved to decision_review to share with HLR

app/models/appeal.rb
```diff
@@ -333,13 +333,6 @@ def root_task
...
-    decision_issues.remanded.each(&:find_or_create_remand_supplemental_claim!)
-    remand_supplemental_claims.each(&:create_remand_issues!)
-    remand_supplemental_claims.each(&:create_decision_review_task_if_required!)
-    remand_supplemental_claims.each(&:start_processing_job!)
-  end
-
```
leikkisa: Moved to decision_review to share with HLR

app/models/appeal.rb
```diff
@@ -400,13 +389,11 @@ def fetch_pre_decision_status
   end
 
   def fetch_post_decision_status
-    if !remanded_issues? && effectuation_ep? && !active_ep?
+    if remand_supplemental_claims.any?
+      active_remanded_claims? ? :ama_remand : :post_bva_dta_decision
```
leikkisa: Talked to Melissa. If there are any active remands (whether processed in Caseflow or not) it should return ama_remand, if there are remands and they are all closed then the post_bva one.

app/models/appeal.rb
```diff
@@ -497,11 +484,7 @@ def distributed_to_a_judge?
   end
 
   def remanded_issues?
-    decision_issues.any? { |di| di.disposition == "remanded" }
-  end
-
-  def remanded_sc_with_ep
-    @remanded_sc_with_ep ||= remand_supplemental_claims.find(&:processed_in_vbms?)
```
leikkisa: No longer needed since we have a way to check whether processed_in_caseflow claims are active now

app/models/claim_review.rb
```diff
@@ -186,6 +190,10 @@ def issues_hash
 
   private
 
+  def incomplete_tasks?
+    tasks.reject(&:completed?).any?
```
leikkisa: This assumes all processed_in_caseflow claims have at least one task (I think that's correct)

app/models/claim_review.rb
```diff
@@ -127,6 +127,10 @@ def cleared_ep?
     end_product_establishments.any? { |ep| ep.status_cleared?(sync: true) }
   end
 
+  def active?
```
mc-usds: Thanks for updating this!

app/models/higher_level_review.rb
```diff
@@ -43,10 +37,6 @@ def incomplete
     false
   end
 
-  def active?
-    hlr_ep_active? || dta_claim_active?
-  end
-
```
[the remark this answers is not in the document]
leikkisa: I think we should make that a separate method (there is a separate method now called active_remand_claims?), so that there's consistency for what active? means between all of the decision reviews.  Otherwise I think it's going to cause bugs.  We could update active? to include remand supplemental_claims always too, if active? isn't being used in ways that wouldn't want that for Appeal.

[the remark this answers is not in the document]
leikkisa: Appeal has something called active_status? that seems to be for this. I can add this to claim_review.

fetch_status in Appeal is using active? and I'm wondering if we want that to use active_status?

app/models/decision_issue.rb
```diff
@@ -33,11 +42,20 @@ def granted
...
     def not_remanded
-      where.not(disposition: "remanded")
+      where.not(disposition: REMAND_DISPOSITIONS)
+    end
+
+    def contested
+      joins("INNER JOIN request_issues on request_issues.contested_decision_issue_id = decision_issues.id")
+    end
+
+    def uncontested
```
pkarman: +1 to these scoping methods

app/models/decision_issue.rb
```diff
@@ -19,6 +19,15 @@ class DecisionIssue < ApplicationRecord
   has_one :effectuation, class_name: "BoardGrantEffectuation", foreign_key: :granted_decision_issue_id
   has_one :contesting_request_issue, class_name: "RequestIssue", foreign_key: "contested_decision_issue_id"
 
+  # NOTE: These are the string identifiers for the DTA error dispositions returned from VBMS.
```
pkarman: glad to see these get moved here!

spec/models/higher_level_review_spec.rb
```diff
@@ -566,14 +505,14 @@
         issue = issue_statuses.find { |i| i[:diagnosticCode] == "9999" }
         expect(issue).to_not be_nil
         expect(issue[:active]).to eq(true)
-        expect(issue[:last_action]).to be_nil
-        expect(issue[:date]).to be_nil
+        expect(issue[:last_action]).to eq("DTA Error - PMRs")
```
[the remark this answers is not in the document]
leikkisa: Oops! I was thinking the original decision issues, based on how issues_hash operates for Appeal. I changed it to return original request issues, and added an active_status? method to be used instead of active? for this.

app/models/higher_level_review.rb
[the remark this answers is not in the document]
mc-usds: A separate method is good. I should have given it a more api specific name in the first place, as that version of "active" is appeal status api specific.

You are correct, fetch_status should be using active_status? instead of active?
For appeal status api,  an appeal is considered active until all remanded claims and effectuations are completed.

app/models/appeal.rb
```diff
@@ -616,26 +582,6 @@ def issues_hash
...
-
-    remand_sc_decisions = []
-    remand_supplemental_claims.each do |sc|
-      sc.decision_issues.each do |di|
-        remand_sc_decisions << di
-      end
-    end
-
-    (di_list + remand_sc_decisions).uniq
-  end
```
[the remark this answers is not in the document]
leikkisa: Oh okay, I see that now.

app/models/higher_level_review.rb
```diff
@@ -55,41 +45,15 @@ def alerts
     # need to implement. add logic to return alert enum
   end
 
-  def fetch_all_decision_issues_for_api_status
-    all_decision_issues = decision_issues.reject { |di| DTA_ERRORS.include?(di[:disposition]) }
-
-    all_decision_issues += dta_claim.decision_issues if dta_claim
-
-    all_decision_issues
-  end
```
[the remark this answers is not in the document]
leikkisa: Thanks for the thorough review and catching this stuff.  I feel much better about it now.  I still combined them because the code for the one-to-many version needs a loop that is the same for both of them, but I added the Appeal-specific behavior to an Appeal method that calls super.

app/models/higher_level_review.rb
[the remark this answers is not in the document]
leikkisa: I switched it, but had to move it back because of tests.

Issue #9340, opened 2019-02-11
Tweak the evidence submission task


amprokop:
- Ensure all VSO tasks are created before the evidence task is marked complete, and ensure the `when_timer_ends` hook fails unless those tasks are created.
- When evidence submission window tasks are created, create them using the `receipt_date` of the appeal rather than the created_at time

PR #9354, opened 2019-02-11
Tweaks evidence submission tasks

amprokop:
https://github.com/department-of-veterans-affairs/caseflow/issues/9340


app/models/tasks/evidence_submission_window_task.rb
```diff
@@ -1,16 +1,12 @@
...
   include TimeableTask
-  after_update :create_ihp_task, if: :status_changed_to_completed_and_has_parent?
 
   def when_timer_ends
-    update!(status: :completed)
-  end
-
-  def create_ihp_task
     RootTask.create_ihp_tasks!(appeal, parent)
+    update!(status: :completed)
```
amprokop: Before, we updated the status of the task to `complete`, then attempted to create necessary IHP tasks. Now, we'll attempt to create ihp tasks, which causes a call to BGS. If that fails, we'll retry later.

app/models/tasks/evidence_submission_window_task.rb
```diff
@@ -1,16 +1,12 @@
...
   end
 
-  def self.timer_delay
-    90.days
+  def timer_ends_at
```
amprokop: This used to ensure evidence tasks expired 90 days from the time of task creation. Here, we change the timing to be 90 days from the receipt date of the appeal.

app/models/concerns/timeable_task.rb
```diff
@@ -4,10 +4,10 @@ module TimeableTask
   module ClassMethods
     def create!(args)
       fail Caseflow::Error::MissingTimerMethod unless method_defined?(:when_timer_ends)
-      fail Caseflow::Error::MissingTimerMethod unless respond_to?(:timer_delay)
+      fail Caseflow::Error::MissingTimerMethod unless method_defined?(:timer_ends_at)
 
       super(args).tap do |task|
-        TaskTimer.create!(task: task, last_submitted_at: Time.zone.now + timer_delay)
+        TaskTimer.create!(task: task, last_submitted_at: task.timer_ends_at)
```
amprokop: instead of assuming a delay, we'll assume the method passes a time when the timer is up.

Issue #9163, opened 2019-02-04
API: postBvaDtaDecision and bvaDecisionEffectuation statuses have details object


cmgiven:
```
"status": {
  "type": "bva_decision_effectuation",
  "details": {
    "bvaDecisionDate": "2019-05-01",
    "aojDecisionDate": "2019-07-15"
  }
}
```

* `bvaDecisionDate`: The date of the Board decision.
* `aojDecisionDate`: The date of the AOJ decision effectuating the Board's decision (in the case of a bvaDecisionEffectuation status) or correcting a DTA error (in the case of a postBvaDtaDecision status).
* Note that postBvaDtaDecision details also has an `issues` array, described in #9162.

PR #9356, opened 2019-02-12
Appeal Status: Adding status details to status attribute

### LINKED ISSUE #9162: API: Decision statuses have issues in details object
```
"status": {
  "type": "sc_decision",
  "details": {
    "issues": [
      {
        "description": "Limitation of wrist motion",
        "disposition": "allowed"
      },
      {
        "description": "Rheumatoid arthritis",
        "disposition": "denied"
      }
    ]
  }
}
```

* This applies to the amaRemand, bvaDecision, postBvaDtaDecision, scDecision, and hlrDecision statuses.
* For each issue on a decision, there should be an entry in an `issues` array.
* The `description` is determined according to the same logic as in #9160.
* The `disposition` is one of "allowed" (granted), "denied", and "remand" (only in the case of a Board decision).
* Note that in the case of a Higher-Level Review with a DTA error, we will be combining decision issues from the original HLR as well as the follow-on SC. We will not show a "remand" disposition for those issues where a DTA error was identified, but rather only show the final disposition after the DTA error was corrected.
* For a `postBvaDtaDecision`, the `issues` will only contain those issues that were addressed on the AOJ decision correcting the DTA error, not the `allowed` or `denied` issues from the original Board decision.

mc-usds:
Resolves #9162, #9163

### Description
Addes status details for some of the status types.

### Acceptance Criteria
- [x] Code compiles correctly

### Testing Plan
1. updated unit tests and feature test

### DIFF
--- a/app/models/appeal.rb
+++ b/app/models/appeal.rb
@@ -415,34 +415,37 @@ class Appeal < DecisionReview
 
   def fetch_details_for_status
     case fetch_status
-    when :ama_remand
-      decision_issues.map do |di|
-        {
-          description: di.description_for_status,
-          decision: di.disposition_for_status
-        }
-      end
-    when :post_bva_dta_decision
-      issue_list = remanded_sc_decision_issues
-      issue_list.map do |di|
-        {
-          description: di.description_for_status,
-          decision: di.disposition_for_status
-        }
-      end
     when :bva_decision
-      decision_issues.map do |di|
-        {
-          description: di.description_for_status,
-          decision: di.disposition_for_status
-        }
-      end
+      decision_issues_status_details
+    when :ama_remand
+      decision_issues_status_details
+    when :post_bva_dta_decision
+      post_bva_dta_decision_details
     when :pending_hearing_scheduling
       {
         type: "video"
       }
     else
       {}
+    end
+  end
+
+  def post_bva_dta_decision_status_details
+    issue_list = remanded_sc_decision_issues
+    issue_list.map do |di|
+      {
+        description: di.api_status_description,
+        disposition: di.api_status_disposition
+      }
+    end
+  end
+
+  def decision_issues_status_details
+    decision_issues.map do |di|
+      {
+        description: di.api_status_description,
+        disposition: di.api_status_disposition
+      }
     end
   end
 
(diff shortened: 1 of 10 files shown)


app/models/appeal.rb
```diff
@@ -413,6 +413,60 @@ def fetch_post_decision_status
...
+    end
+  end
+
+  def post_bva_dta_decision_status_details
+    issue_list = remanded_sc_decision_issues
+    issue_list.map do |di|
+      {
+        description: di.api_status_description,
+        disposition: di.api_status_disposition,
+        bvaDecisionDate: decision_event_date,
```
cmgiven: This will return an array of issues, each containing a `bvaDecisionDate` and `aojDecisionDate`. We need to nest that array in an object, and include the decision dates on that object, not on each issue:

```json
{
  "issues": [
    {
      "description": "Compensation issue",
      "disposition": "allowed"
    }
  ],
  "bvaDecisionDate": "2019-01-01",
  "aojDecisionDate": "2019-01-01"
}
```

app/models/appeal.rb
```diff
@@ -413,6 +413,60 @@ def fetch_post_decision_status
...
+        aojDecisionDate: dta_descision_event_date
+      }
+    end
+  end
+
+  def decision_issues_status_details
+    decision_issues.map do |di|
```
cmgiven: Similarly, please nest inside an object as `"issues": [...]`.

app/models/appeal.rb
```diff
@@ -413,6 +413,60 @@ def fetch_post_decision_status
...
+
+  def remanded_sc_decision_issues
+    issue_list = []
+    remand_supplemental_claims.each do |sc|
+      sc.decision_issues.map do |di|
+        issue_list << di
+      end
+    end
+
+    issue_list
```
cmgiven: Same comment.

app/models/decision_issue.rb
```diff
@@ -105,8 +105,28 @@ def api_status_last_action_date
     approx_decision_date
   end
 
+  def api_status_disposition
+    "remand" if disposition == "remanded"
```
cmgiven: Sorry.

app/models/higher_level_review.rb
```diff
@@ -191,9 +191,24 @@ def fetch_status
...
+
+  def fetch_details_for_status
+    case fetch_status
+    when :hlr_decision
+      issue_list = fetch_all_decision_issues_for_api_status
+      issue_list.map do |di|
+        {
+          description: di.api_status_description,
+          disposition: di.api_status_disposition
+        }
```
cmgiven: Same comment here. Array needs to be inside an object.

app/models/supplemental_claim.rb
```diff
@@ -128,4 +129,18 @@ def fetch_status
...
   end
+
+  def fetch_details_for_status
+    case fetch_status
+    when :sc_decision
+      decision_issues.map do |di|
+        {
+          description: di.api_status_description,
+          disposition: di.api_status_disposition
+        }
```
cmgiven: And once more here.

app/models/appeal.rb
mc-usds: Thanks for catching this. I totally misread the gh issue. Fixed it.

app/models/appeal.rb
mc-usds: fixed.

app/models/appeal.rb
mc-usds: fixed

app/models/higher_level_review.rb
mc-usds: fixed

app/models/supplemental_claim.rb
mc-usds: fixed

PR #9362, opened 2019-02-12
Add uniq constraint to claimants table

aroltsch:
In production there are multiple claimant records with the same participant id associated with the same review request.

This is the first PR to add a uniq constraint to the claimants table.

In the subsequent PRs:
1. Write a migration to delete duplicate claimants records
2. Add a unique constraint at the DB level

### DIFF
--- a/app/models/claimant.rb
+++ b/app/models/claimant.rb
@@ -7,6 +7,10 @@ class Claimant < ApplicationRecord
 
   bgs_attr_accessor :first_name, :last_name, :middle_name, :relationship
 
+  validates :participant_id,
+            uniqueness: { scope: [:review_request_id, :review_request_type],
+                          on: :create }
+
   def self.create_from_intake_data!(participant_id:, payee_code:)
     create!(
       participant_id: participant_id,
--- a/spec/models/claimant_spec.rb
+++ b/spec/models/claimant_spec.rb
@@ -120,6 +120,44 @@ describe Claimant do
   end
 
   context "#valid?" do
+    context "participant_id" do
+      let(:participant_id) { "1234" }
+
+      let(:review_request) do
+        build(:higher_level_review,
+              id: 1,
+              benefit_type: "fiduciary",
+              veteran_file_number: create(:veteran).file_number)
+      end
+
+      let!(:claimant) do
+        create(:claimant, review_request: review_request, participant_id: participant_id)
+      end
+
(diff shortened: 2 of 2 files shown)


app/models/claimant.rb
```diff
@@ -7,6 +7,10 @@ class Claimant < ApplicationRecord
 
   bgs_attr_accessor :first_name, :last_name, :middle_name, :relationship
 
+  validates :participant_id,
```
pkarman: do we also want a db constraint (unique index)?

aroltsch: I pointed out in the PR description that I will add that later. I cannot add this right now because it can cause some failures until i remove duplicates.

pkarman: doh! :shameface:

spec/models/claimant_spec.rb
```diff
@@ -120,6 +120,44 @@
...
+
+        it "requires uniqueness" do
+          expect(subject).to_not be_valid
+          expect(subject.errors.messages[:participant_id]).to eq ["has already been taken"]
+        end
+      end
+
+      context "when created with the same participant_id and different review_request" do
+        let(:review_request2) do
+          build(:appeal,
```
pkarman: IUC, using `build` here doesn't write to the db, so I'm not sure how the `find_or_initialize_by` would ever find it?

spec/models/claimant_spec.rb
```diff
@@ -120,6 +120,44 @@
   end
 
   context "#valid?" do
+    context "participant_id" do
+      let(:participant_id) { "1234" }
+
+      let(:review_request) do
+        build(:higher_level_review,
+              id: 1,
```
pkarman: why the explicit `id` value?

spec/models/claimant_spec.rb
aroltsch: This test doesn't test this code https://github.com/department-of-veterans-affairs/caseflow/pull/9362/files#diff-bc46c6c1b2e028b83ffabba96b1f31ddR17

`build` works fine here, because I create it above in the DB. Then build fails validation because there is a claimant with that participant id in the DB already

spec/models/claimant_spec.rb
aroltsch: explicit id because I'd like to test the scenario where two review requests with the same ID but different type

Issue #9227, opened 2019-02-05
Intake Non-comp | Edit Issues | When removing an issue where the benefit type is not compensation or pension, do not reference VBMS in the alert modal.


steve-gov:
Non-VBMS users should not see warnings about EPs or VBMS.

Observed:
1. As an Fiduciary Caseflow Administrator, I go to a disposition task in my task list (eg https://appeals.cf.uat.ds.va.gov/decision_reviews/fiduciary/tasks/114) and click `Edit Issues`
2. I am directed to the Add/ Remove Issues page for that decision review, where I click `Remove` on one of the issues
3. The alert modal reads "The contention you selected will be removed from the EP in VBMS."

Expected:
- In step 3, the alert modal should read "The contention you selected will be removed from the decision review."

![screencapture](https://user-images.githubusercontent.com/25468689/52311859-4d794000-295d-11e9-9081-88a0ce398778.gif)

steve-gov:
Validated on several noncomp issues:

<img width="427" alt="screen shot 2019-02-12 at 4 57 51 pm" src="https://user-images.githubusercontent.com/25468689/52678686-5e8af980-2ee7-11e9-8c8f-a21819fbe2b9.png">

PR #9368, opened 2019-02-12
Use special Remove Issue confirmation text for non-comp reviews

pkarman:
connects #9227 

### Description
Special text in remove issue modal for caseflow-only EPs.

### DIFF
--- a/client/app/intake/components/RemoveIssueModal.jsx
+++ b/client/app/intake/components/RemoveIssueModal.jsx
@@ -4,9 +4,17 @@ import React from 'react';
 
 import { removeIssue } from '../actions/addIssues';
 import Modal from '../../components/Modal';
+import { benefitTypeLooksLikeCompensation } from '../util';
 
-const removeIssueMessage = (formType) => {
-  if (formType === 'appeal') {
+const removeIssueMessage = (intakeData) => {
+  if (intakeData.benefitType && !benefitTypeLooksLikeCompensation(intakeData.benefitType)) {
+    return <div>
+      <p>The contention you selected will be removed from the decision review.</p>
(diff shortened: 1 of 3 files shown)


pkarman:
screen shot 2019-02-12 at 2 04 49 pm

client/app/intake/components/RemoveIssueModal.jsx
```diff
@@ -4,9 +4,17 @@ import React from 'react';
 
 import { removeIssue } from '../actions/addIssues';
 import Modal from '../../components/Modal';
+import { benefitTypeLooksLikeCompensation } from '../util';
```
leikkisa: Should we name this more consistently with the other versions of this? like processed_in_vbms or something like that.  Shane mentioned the possibility that other business lines could move to VBMS.

client/app/intake/components/RemoveIssueModal.jsx
```diff
@@ -4,9 +4,17 @@ import React from 'react';
...
 
-const removeIssueMessage = (formType) => {
-  if (formType === 'appeal') {
+const removeIssueMessage = (intakeData) => {
+  if (intakeData.benefitType && !benefitTypeLooksLikeCompensation(intakeData.benefitType)) {
+    return <div>
+      <p>The contention you selected will be removed from the decision review.</p>
```
leikkisa: Have a tech backlog ticket to move front end copy to client/COPY.json, this could be an opportunity. Don't feel strongly about it though.

pkarman: yeah let's wait to move everything to COPY.json at one go.

PR #9373, opened 2019-02-12
Add generic cancel task action

### LINKED ISSUE #8701: Add ability to cancel tasks in DB

lowellrex:
Resolves #8701. Adds the "Cancel task" option to the list of actions available for `GenericTask`s in preparation for it being added to other tasks. Changes some existing places where we set the status of tasks to `completed` to `cancelled` instead.

full-flow


app/models/appeal.rb
```diff
@@ -603,7 +603,7 @@ def other_close_event_date
     return if active_status?
     return if decision_issues.any?
 
-    root_task.completed_at
+    root_task.closed_at
```
lowellrex: `completed_at` field [no longer exists](https://github.com/department-of-veterans-affairs/caseflow/pull/9181) in the `tasks` table. Don't know why this wasn't causing any test failures...

app/models/task.rb
```diff
@@ -13,8 +13,8 @@ class Task < ApplicationRecord
   after_create :put_parent_on_hold
 
   before_update :set_timestamps
-  after_update :update_parent_status, if: :status_changed_to_completed_and_has_parent?
-  after_update :update_children_status, if: :status_changed_to_completed?
+  after_update :update_parent_status, if: :status_changed_to_closed_and_has_parent?
+  after_update :update_children_status, if: :status_changed_to_closed?
```
lowellrex: We are treating the cancellation of tasks the same as the completion of tasks with regards to how that action bubbles up and down through parent and child tasks. [Related slack discussion](https://dsva.slack.com/archives/C6E41RE92/p1550001536061900) where we settled on this being the correct behaviour.

app/models/task.rb
```diff
@@ -210,7 +210,7 @@ def reassign(reassign_params, current_user)
       t.save!
     end
 
-    update!(status: Constants.TASK_STATUSES.completed)
+    update!(status: Constants.TASK_STATUSES.cancelled)
```
lowellrex: Reassigning tasks now cancels the old task.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -97,9 +97,8 @@ def update_from_params(params, current_user)
 
         slot_new_hearing(hearing_day_id, hearing_type, hearing_time, hearing_location)
         HoldHearingTask.create_hold_hearing_task!(appeal, parent)
-      elsif params[:status] == "canceled"
+      elsif params[:status] == Constants.TASK_STATUSES.cancelled
```
lowellrex: Now that we have the `cancelled` task status we can use it here.

app/models/task.rb
```diff
@@ -394,12 +403,12 @@ def update_parent_status
...
-    saved_change_to_attribute?("status") && completed?
+  def status_changed_to_closed?
+    saved_change_to_attribute?("status") && !active?
   end
 
-  def status_changed_to_completed_and_has_parent?
-    status_changed_to_completed? && parent
+  def status_changed_to_closed_and_has_parent?
+    status_changed_to_closed? && parent
   end
```
tomas-nava: These names (`status_changed_to_closed?` and `status_changed_to_closed_and_has_parent?`) feel misleading to me now; `task_closed?` and `task_closed_and_has_parent?` would be more accurate.

lowellrex: Agreed! Maybe we could go with `task_just_closed?` to convey that the status just changed? `task_closed?` doesn't imply to me as strongly that we are checking to see if a status change just occurred.

Issue #9285, opened 2019-02-07
Show "assigned to" for all AMA caseflow tasks


lowellrex:
We use the [`Appeal.location_code` function](https://github.com/department-of-veterans-affairs/caseflow/blob/becb7df57bda7cca095cdeba5941bfa1c2eaeac1/app/models/appeal.rb#L104) to populate the "Assigned to" column of queue table views for AMA/RAMP appeals. However, sometimes we don't return anything in that field. This ticket exist to always return some value for those appeals.

The first few rows with empty "Assigned to" values are examples of this behaviour in action:

![image](https://user-images.githubusercontent.com/32683958/52430656-999ac080-2ad4-11e9-8cee-d26a0205f4ab.png)

[Related slack conversation](https://dsva.slack.com/archives/C6E41RE92/p1549470790078200).

## Acceptance criteria
* [ ] Create a test scenario that recreates this situation (no assigned to value).
* [ ] Always return some value from the `Appeal.location_code` method for AMA/RAMP appeals.

lowellrex:
MEDIUM. Perhaps we can leverage the new `fetch_pre_decision_status` and `fetch_post_decision_status` Appeal model methods here? https://github.com/department-of-veterans-affairs/caseflow/blob/becb7df57bda7cca095cdeba5941bfa1c2eaeac1/app/models/appeal.rb#L365

PR #9379, opened 2019-02-13
add condition to location_code if there aren't any active tasks

joeyyang:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/9285

### Description
Think I have a solution here, but want to confirm that this logic is correct. Can you verify the logic?

Screenshot and seed data coming shortly.


app/models/appeal.rb
```diff
@@ -121,13 +121,15 @@ def location_code
...
-      elsif !active_tasks.empty?
+      elsif active_tasks.any?
         most_recent_assignee = active_tasks.order(updated_at: :desc).first.assigned_to
         location_code = if most_recent_assignee.is_a?(Organization)
                           most_recent_assignee.name
                         else
                           most_recent_assignee.css_id
                         end
+      else
+        location_code = status_hash[:type].to_s.titleize
```
lowellrex: It looks like this returns the following:

image

I was able to get to this position by placing a task assigned to a member of the VLJ support staff on hold. I think in this case we'd like to have cases in this position display the name of that member of the VLJ support staff since it is a little more precise than "Decision In Progress".

lowellrex: One thought on how we could refactor this method to use some existing functions we've already built. Perhaps we could offload the logic of what to show in that column to the task itself. Maybe something like:

```ruby
class Task < ApplicationRecord
...
  def assigned_to_label
    assigned_to.is_a?(Organization) ? assigned_to.name : assigned_to.css_id
  end
```

With tasks that should display special text overriding the function:

```ruby
class DistributionTask < GenericTask
...
  def assigned_to_label
    COPY::CASE_LIST_TABLE_CASE_STORAGE_LABEL
  end
```

And then the only job for this `Appeal` method would be to determine which task's assigned to label to use:

```ruby
class Appeal < DecisionReview
...
  def location_code
    return COPY::CASE_LIST_TABLE_POST_DECISION_LABEL if root_task&.status == Constants.TASK_STATUSES.completed
    return tasks.active.order(:updated_at).last.assigned_to_label if tasks.active.any?
    return tasks.order(:updated_at).last.assigned_to_label
  end
```

What do you think of that approach?

app/models/appeal.rb
```diff
@@ -115,29 +115,15 @@ def type
...
-                          most_recent_assignee.css_id
-                        end
-      end
-    end
+    active_tasks = tasks.where(status: [Constants.TASK_STATUSES.in_progress, Constants.TASK_STATUSES.assigned])
+    return active_tasks.order(:updated_at).last.assigned_to_label if active_tasks.any?
+    return tasks.order(:updated_at).last.assigned_to_label if tasks.any?
 
-    location_code
+    status_hash[:type].to_s.titleize
```
joeyyang: accounts for the case where `appeal.tasks.empty?`

app/models/tasks/root_task.rb
```diff
@@ -33,6 +33,10 @@ def actions_available?(_user)
     true
   end
 
+  def assigned_to_label
```
joeyyang: you had suggested this live in `distribution_task.rb` but based on my previous understanding if we ever only had a `RootTask` it would be "Assigned to: Case storage"

joeyyang: this corresponds to the old version of `appeal.rb:126-127`

app/models/appeal.rb
```diff
@@ -115,29 +115,15 @@ def type
...
-        location_code = if most_recent_assignee.is_a?(Organization)
-                          most_recent_assignee.name
-                        else
-                          most_recent_assignee.css_id
-                        end
-      end
-    end
+    active_tasks = tasks.where(status: [Constants.TASK_STATUSES.in_progress, Constants.TASK_STATUSES.assigned])
+    return active_tasks.order(:updated_at).last.assigned_to_label if active_tasks.any?
+    return tasks.order(:updated_at).last.assigned_to_label if tasks.any?
```
lowellrex: I think this one should be `return tasks.active.order(:updated_at).last.assigned_to_label if tasks.any?` to include tasks that are on hold.

lowellrex: Upon closer inspection I think this should be something like:

```ruby
on_hold_tasks = tasks.where(status: Constants.TASK_STATUSES.on_hold)
return on_hold_tasks.order(:updated_at).last.assigned_to_label if on_hold_tasks.any?
```

app/models/appeal.rb
lowellrex: I think this makes sense. Let's build a truth table of the various task status states to be sure!

✅ -> One or more tasks with this status
❌ -> No tasks with this status
➖ -> It does not matter if there are tasks with this status

. | root_task completed | assigned or in_progress | on_hold | cancelled or completed | result
--- | --- | --- | --- | --- | ---
1 | ✅ | ➖ | ➖ | ➖ | `COPY::CASE_LIST_TABLE_POST_DECISION_LABEL`
2 | ❌ | ✅ | ➖ | ➖ | `assigned_to_label` of most recently updated task with `assigned` or `in_progress` status
3 | ❌ | ❌ | ✅ | ➖ | `assigned_to_label` of most recently updated task with `on_hold` status
4 | ❌ | ❌ | ❌ | ✅ | `assigned_to_label` of most recently updated task
5 | ❌ | ❌ | ❌ | ❌ | This scenario should not be possible, but I think the fallback to `status_hash[:type].to_s.titleize` makes sense

I think we handle scenarios 1, 2, and 5 correctly but we need to make some slight modifications to the logic above to properly account for 3 and 4. I've made suggestions about how we could do that here.

app/models/appeal.rb
lowellrex: I think we need to insert a few lines between 122 and 123 to account for on hold tasks:

Upon closer inspection I think this should be something like:

```ruby
on_hold_tasks = tasks.where(status: Constants.TASK_STATUSES.on_hold)
return on_hold_tasks.order(:updated_at).last.assigned_to_label if on_hold_tasks.any?
```

app/models/appeal.rb
joeyyang: awesome, i think this makes sense!

PR #9380, opened 2019-02-13
Updates to AHL Job

lomaxap:
- Queries all schedule hearing task with more efficient veteran file_number accessor
- handles no method errors for error messages and legacy appeals without case_records

### DIFF
--- a/app/jobs/fetch_hearing_locations_for_veterans_job.rb
+++ b/app/jobs/fetch_hearing_locations_for_veterans_job.rb
@@ -25,13 +25,7 @@ class FetchHearingLocationsForVeteransJob < ApplicationJob
         AND status != 'completed') admin_actions
         ON admin_actions.parent_id = id")
       .where("admin_actions.parent_id IS NULL")
-      .map do |task|
-        if task.appeal.is_a?(Appeal)
-          task.appeal.veteran_file_number
-        else
-          LegacyAppeal.veteran_file_number_from_bfcorlid(task.appeal.vbms_id)
-        end
-      end.compact
+      .map { |task| task.appeal.veteran_file_number }.compact
   end
 
   def missing_veteran_file_numbers
--- a/app/jobs/fetch_hearing_locations_for_veterans_job.rb
+++ b/app/jobs/fetch_hearing_locations_for_veterans_job.rb
@@ -5,7 +5,7 @@ class FetchHearingLocationsForVeteransJob < ApplicationJob
   QUERY_LIMIT = 500
 
   def veterans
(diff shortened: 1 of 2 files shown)


app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -81,16 +87,15 @@ def perform
...
     end
   end
 
   def perform_once_for(veteran)
     begin
       va_dot_gov_address = validate_veteran_address(veteran)
     rescue Caseflow::Error::VaDotGovLimitError
-      sleep 60
-      va_dot_gov_address = validate_veteran_address(veteran)
+      return false
```
lomaxap: originally told limit if 500/minute. looks more like 500/hour. ending job if limit is reached.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -3,7 +3,7 @@ class ScheduleHearingTask < GenericTask
 
   class << self
     def find_or_create_if_eligible(appeal)
-      if appeal.is_a?(LegacyAppeal) && appeal.case_record.bfcurloc == "57" &&
+      if appeal.is_a?(LegacyAppeal) && appeal.case_record&.bfcurloc == "57" &&
```
lomaxap: any idea why a legacy appeal wouldn't have a case record? we should probably handle this better rather than ignoring it.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -17,15 +17,21 @@ def file_numbers
...
         LEFT OUTER JOIN (SELECT parent_id FROM tasks
         WHERE type IN ('HearingAdminActionVerifyAddressTask', 'HearingAdminActionForeignVeteranCaseTask')
         AND status != 'completed') admin_actions
         ON admin_actions.parent_id = id")
-      .where("admin_actions.parent_id IS NULL").limit(QUERY_LIMIT)
-      .map { |task| task.appeal.veteran&.id }.compact
+      .where("admin_actions.parent_id IS NULL")
+      .map do |task|
+        if task.appeal.is_a?(Appeal)
+          task.appeal.veteran_file_number
```
mdbenjam: I think you can just call `veteran_file_number` on both legacy and ama appeals. On legacy it is just an alias for `sanitized_vbms_id` which calls `veteran_file_number_from_bfcorlid`

app/models/tasks/schedule_hearing_task.rb
mdbenjam: Hmmm that's very weird... we only create legacy appeals when there's an underlying record in VACOLS. Do you have a sentry error where this happened? It'd be good to investigate a little further.

PR #9405, opened 2019-02-13
Create hearing admin branch

### LINKED ISSUE #9255: Create a HearingAdminOrganization
We don't want all of the schedule hearing tasks to clog up the organizational queue. In order to separate these views, we're going to recreate a new HearingAdminOrganization for all the admin actions.

# AC
1. New admin actions should be assigned to HearingAdminActionOrganization
1. Reassign already created admin actions to HearingAdminActionOrganization
1. Add users who are in the HearingsManagementOrganization to the `HearingAdminActionOrganization`

mdbenjam:
Resolves #9255 

### Description
Three changes in this PR
1) Create a new Hearing Admin organization for admin tasks
2) Change the name from HearingsManagement to HearingManagement
3) Stop checking the status when creating a Hearing task so that we don't fail

### DIFF
--- a/app/services/appeal_repository.rb
+++ b/app/services/appeal_repository.rb
@@ -21,7 +21,10 @@ class AppealRepository
 
     # Load the VACOLS case records associated with legacy tasks into memory in a single batch.
     cases = (vacols_records_for_appeals(legacy_appeal_ids) || []).group_by(&:id)
-    aod = VACOLS::Case.aod(legacy_appeal_ids)
+
+    aod = legacy_appeal_ids.in_groups_of(1000, false).reduce({}) do |acc, group|
+      acc.merge(VACOLS::Case.aod(group))
+    end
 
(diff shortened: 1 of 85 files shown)


app/services/appeal_repository.rb
```diff
@@ -340,7 +340,6 @@ def self.create_schedule_hearing_tasks
     LegacyAppeal.where(vacols_id: ids.map(&:first) - vacols_ids_with_schedule_tasks).each do |appeal|
       parent = HearingTask.find_or_create_by!(
         appeal: appeal,
-        status: Constants.TASK_STATUSES.assigned.to_sym,
```
mdbenjam: I think this is what was causing our problems with the Hearing task. It would be immediately placed on hold, so if we try to run this again for the same appeal, we'll try to create an open task which isn't allowed.

Issue #9337, opened 2019-02-11
Create DistributionTasks for all remaining RAMP appeals


amprokop:
We'll plan to distribute all remaining RAMP appeals through auto case distribution, rather than through the RAMP assignment script.

For all RAMP appeals that have not been assigned to judges:

- Run `set_target_decision_date!` against the appeal
- If the appeal is direct review, create a DistributionTask to mark the appeal ready for distribution
- If the appeal is evidence submission, create a DistributionTask and an EvidenceSubmissionWindowTask as a child of the DistributionTask. It's fine if we do this no matter if the appeal has already exceeded the 90 day window or not — the nightly job will just pick it up the next night and mark the task as complete (https://github.com/department-of-veterans-affairs/caseflow/issues/9339).
- Ramp appeals that were already intook will already have an IHP task—we need to ensure that we don't create duplicates. 
  - in `RootTask.create_ihp_tasks!`, before creating new InformalHearingPresentation tasks, check if one already exists for that organization and that appeal, and if so, just return the existing task.
  - Ordinarily for evidence submission appeals we would create IHP tasks *after* the evidence submission period ends. However, for the purpose of backfilling tasks for RAMP appeals, it shouldn't be a problem if the IHP task and the evidence submission task are open simultaneously — the IHP task currently never expires and it will block distribution until it is complete. 
  - Q: Are there appeals that could have multiple IHP tasks for the same organization? If so, this approach may create problems.

- If the appeal is a hearing appeal...
  - Create a HearingTask as a child of the DistributionTask and a ScheduleHearingTask as a child of the HearingTask (#9206)

laurjpeterson:
Q: With this outline above, will the EvidenceWindowSubmissionTask's start date be the receipt date of the NOD?

amprokop:
Yeah, I will make that change first.

amprokop:
https://github.com/department-of-veterans-affairs/caseflow/issues/9340

cmgiven:
Ah, we will also need to run `set_target_decision_date!` against all of these.

amprokop:
Good call.

PR #9406, opened 2019-02-13
Amprokop backfill ramp tasks

amprokop:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/9337

### DIFF
--- a/app/models/tasks/judge_task.rb
+++ b/app/models/tasks/judge_task.rb
@@ -70,5 +70,4 @@ class JudgeTask < Task
     task.children.all? { |t| !t.is_a?(JudgeTask) }
   end
   #:nocov:
-  # rubocop:enable Metrics/AbcSize
 end
--- a/app/models/tasks/judge_task.rb
+++ b/app/models/tasks/judge_task.rb
@@ -22,7 +22,6 @@ class JudgeTask < Task
   end
 
   #:nocov:
-  # rubocop:disable Metrics/AbcSize
   def self.backfill_ramp_appeals_with_tasks(dry_run: true)
     # Find all unassigned tasks and sort them by the NOD date
     tasks = unassigned_ramp_tasks.sort_by { |task| task.appeal.receipt_date }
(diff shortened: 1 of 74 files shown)


app/models/tasks/judge_task.rb
```diff
@@ -22,67 +22,53 @@ def previous_task
...
-    return false if task.appeal.hearing_docket?
-
-    # If it's an evidence submission case, we need to wait until the
-    # evidence submission window is over
-    if task.appeal.evidence_submission_docket?
-      return false if task.appeal.receipt_date > 90.days.ago
-    end
 
-    task.children.all? { |t| !t.is_a?(JudgeTask) && !t.active? }
+    task.children.all? { |t| !t.is_a?(JudgeTask) }
```
amprokop: This is a change — for every Appeal without a JudgeTask, we need to backfill with DistributionTasks — even if that appeal has an open mail task or an open judge task.

app/models/tasks/judge_task.rb
```diff
@@ -22,67 +22,53 @@ def previous_task
...
-      )
-      Rails.logger.info("Assigned judge task with task id #{task.id} to #{task.assigned_to.css_id}")
+  def self.backfill_tasks(root_tasks)
+    transaction do
+      root_tasks.each do |root_task|
+        Rails.logger.info("Creating subtasks for appeal #{root_task.appeal.id}")
+        RootTask.create_subtasks!(root_task.appeal, root_task)
+        distribution_task = DistributionTask.find_by(parent: root_task)
+        # Update any open IHP tasks if they exist so that they block distribution.
+        ihp_task = InformalHearingPresentationTask.find_by(appeal: root_task.appeal)
```
cmgiven: Do we need to do this with mail tasks too? Any other kinds of tasks we should move over?

amprokop: We don't need to do it with mail tasks — not implemented yet, but the approach we agreed on was for me to modify the `ready_for_distribution_at` to exclude mail tasks.

amprokop: @lowellrex — would you mind taking a quick look at this — as far as I know, IHP and mail tasks are the only currently used task types that should block distribution — are there any others?

PR #9410, opened 2019-02-13
Fix for showing draft dispositions through API

cmgiven:
The Board received a report of a VA.gov user being able to see a draft decision on the Issues tab. In addition to recording the draft disposition, VACOLS is marking the issue as having a close date, bypassing the existing check (unclear whether this has always been the case or is a regression). This adds a check of the status of the appeal itself to ensure that we do not show Board dispositions (allowed, denied, remanded) of an issue while it is active at the Board.

### DIFF
--- a/app/models/appeal_series_issues.rb
+++ b/app/models/appeal_series_issues.rb
@@ -59,6 +59,7 @@ class AppealSeriesIssues
       next [] unless ELIGIBLE_TYPES.include? appeal.type
 
       appeal.issues.reject(&:merged?).each do |issue|
+        issue.appeal = appeal
         issue.cavc_decisions = appeal.cavc_decisions.select do |cavc_decision|
           cavc_decision.issue_vacols_sequence_id == issue.vacols_sequence_id
         end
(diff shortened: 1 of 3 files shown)


app/models/appeal_series_issues.rb
```diff
@@ -74,8 +75,11 @@ def last_action_for_issues(issues)
         type = last_action_type_from_disposition(issue.disposition)
 
         if type
-          memo[:date] = issue.close_date
-          memo[:type] = type
+          # Prevent draft decisions from being shared publicly
+          unless [:allowed, :denied, :remand].include?(type) && issue.appeal.activated?
```
amprokop: Will this preclude AOJ grants or field grants from showing up in Appeals Status?

cmgiven: It won't, because they would be of a different type. Only if it's a Board disposition on an active appeal would it be blocked.

amprokop: ah got it

PR #9412, opened 2019-02-13
Fix flakey checkbox test

tomas-nava:
### Description
Explicitly uses the task ID to identify associated checkboxes in a test that was flaking when lingering tasks pushed the test task IDs up.

### DIFF
--- a/spec/feature/queue/judge_assignment_spec.rb
+++ b/spec/feature/queue/judge_assignment_spec.rb
@@ -10,16 +10,15 @@ RSpec.feature "Judge assignment to attorney" do
   let(:appeal_one) { FactoryBot.create(:appeal) }
   let(:appeal_two) { FactoryBot.create(:appeal) }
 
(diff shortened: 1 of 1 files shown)


spec/feature/queue/judge_assignment_spec.rb
```diff
@@ -38,40 +37,44 @@
       case_rows = page.find_all("tr[id^='table-row-']")
       expect(case_rows.length).to eq(2)
 
-      # step "checks both cases and assigns them to an attorney"
-      scroll_element_in_to_view(".usa-table-borderless")
-      check "1", allow_label_click: true
-      check "2", allow_label_click: true
+      step "checks both cases and assigns them to an attorney" do
+        scroll_element_in_to_view(".usa-table-borderless")
+        check judge_task_one.id.to_s, allow_label_click: true
```
pkarman: +1 to these explicit vars instead of assuming hardcoded strings.

PR #9414, opened 2019-02-14
add intake fingerprint to js errors

kencheeto:
Sentry alerts originating from different js apps (intake vs queue vs hearing etc) are often coalesced when they are raised in an external API call (see https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2991/events/)

Sentry [allows providing a custom fingerprint](https://blog.sentry.io/2018/01/18/setting-up-custom-fingerprints) (fingerprints are what group errors together). This PR illustrates one way we could set a custom fingerprint as each different js app boots, and configuring Raven to use the custom fingerprint if available.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -63,7 +63,13 @@
     <script src="https://cdn.ravenjs.com/3.27.0/raven.min.js"></script>
     <script>
       Raven.config("<%= ENV["PUBLIC_SENTRY_DSN"] %>", {
-        release: "<%= Rails.application.config.build_version[:commit][0..6] %>"
+        release: "<%= Rails.application.config.build_version[:commit][0..6] %>",
+        dataCallback: function(data) {
+          if (Raven.caseflowAppName) {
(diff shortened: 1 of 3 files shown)


leikkisa:
Very cool

app/views/layouts/application.html.erb
```diff
@@ -63,7 +63,13 @@
     <script src="https://cdn.ravenjs.com/3.27.0/raven.min.js"></script>
     <script>
       Raven.config("<%= ENV["PUBLIC_SENTRY_DSN"] %>", {
-        release: "<%= Rails.application.config.build_version[:commit][0..6] %>"
+        release: "<%= Rails.application.config.build_version[:commit][0..6] %>",
+        dataCallback: function(data) {
+          if (Raven.caseflowAppName) {
+            data.fingerprint = ["{{ default }}", Raven.caseflowAppName]
```
kencheeto: the `default` at the beginning of this array means that all errors from an external API call will still get grouped together, but have subgroups under it by caseflow app name if provided.

PR #9415, opened 2019-02-14
Fix flakey queue specs

### LINKED ISSUE #8680: Flake: expected to find visible css "label" with text "Before certification" 2 times, found 1 match
https://circleci.com/gh/department-of-veterans-affairs/caseflow/33246?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

image

monfresh:
For the attorney checkout specs, I tried a bunch of things, including
adding unique class names and IDs to the checkboxes and radio buttons,
but the only thing that seemed to work is to call `page.all` an extra
time, possibly combined with changing the order of clicks. For example,
instead of clicking "Current findings" and then one of the radio buttons
beneath it, we click "Current findings", then "Nexus option", then the
radio buttons under each of those checkboxes.

Resolves #8680


spec/feature/queue/attorney_checkout_flow_spec.rb
```diff
@@ -507,20 +507,28 @@
         page.find("div", class: "Select-option", text: "Stay").click
 
         click_on "Continue"
+
         expect(page).to have_content("Select Remand Reasons")
         expect(page).to have_content(appeal.issues.first.note)
 
-        page.all("label", text: "Current findings", count: 1)[0].click
-        page.all("label", text: "After certification", count: 1)[0].click
+        page.all("label", text: "Current findings").size
```
aroltsch: Do you want to check the size with the expectation?

aroltsch: Otherwise it is very confusing on why you are doing that

monfresh: Agreed. I added the expectations and the test is still passing. I ran it 5 times in a row.

PR #9426, opened 2019-02-14
Appeal Status: Add status attr post decision alerts for HLR, SC and Appeal

### LINKED ISSUE #8840: API: amaPostDecision alert
If VA has made a decision that is eligible to be appealed, SCs, HLRs, and Appeals should have an amaPostDecision alert in their `alerts` array.

```
{
  "type": "ama_post_decision",
  "details": {
    "decisionDate": "2019-05-01",
    "availableOptions": ["supplemental_claim", "cavc"],
    "dueDate": "2020-04-30",
    "cavcDueDate": "2019-08-29"
  }
}
```

* `type` is `"ama_post_decision"`
* `decisionDate` is the date of the decision that is eligible to be appealed.
* `availableOptions` is an array of valid decision review options, of `"supplemental_claim"`, `"higher_level_review"`, `"appeal"`, and `"cavc"`.
* `dueDate` is generally 365 days after `decisionDate`, unless the appeal is a contested claim, in which case it is 60 days after `decisionDate`. This should be `null` if the decision is a Board decision on a contested claim or fiduciary claim type.
* `cavcDueDate`  is 120 days after `decisionDate`. It should be `null` unless the decision is a Board decision.

The alert should only be included if either `dueDate` or `cavcDueDate` is greater than or equal to today.

There may be multiple amaPostDecision alerts on an Appeal if the Board decision (decision 1) was a remand or required effectuation, which could create a VBMS EP which when closed would be an appealable decision 2. They should be ordered chronologically, first to last.

### Eligibility for different options

* Supplemental claims, and AOJ decisions after a Board remand or to effectuate a Board decision (creating a VBMS EP) have the options `["supplemental_claim", "higher_level_review", "appeal"]`, unless it is a contested claim or fiduciary claim type.
  * If it is a contested claim, the only option is `"appeal"`.
  * If it is a fiduciary claim type, `"supplemental_claim"` is not an option.
* Higher-Level Review decisions have the options `["supplemental_claim", "appeal"]`, unless it is a fiduciary claim type.
  * If it is a fiduciary claim type, `"supplemental_claim"` is not an option.
* Board Appeal decisions have the options `["supplemental_claim", "cavc"]`, unless it is a fiduciary claim type.
  * If it is a fiduciary claim type, `"supplemental_claim"` is not an option.

mc-usds:
Resolves #8840

### Description
Adding the post decision alert for HLR, SC and Appeals

### Acceptance Criteria
- [X] Code compiles correctly

### Testing Plan
1. updated unit test and feature test.


mc-usds:
Marking at WIP as this will need to be updated once https://github.com/department-of-veterans-affairs/caseflow/pull/9353 is merged.

@cmgiven If an appeal has 1 decision and it's a remand, do we want to have an alert for that decision or only have an alert once the remand has a decision?

app/models/appeal.rb
```diff
@@ -400,7 +400,7 @@ def fetch_pre_decision_status
   end
 
   def fetch_post_decision_status
-    if !remanded_issues? && effectuation_ep? && !active_ep?
+    if !remanded_issues? && effectuation_ep? && !active_effectuation_ep?
```
leikkisa: Looks like we will want to add something for checking the status of effectuation tasks too, if this is similar to what we talked about before.

app/models/higher_level_review.rb
```diff
@@ -100,6 +100,29 @@ def events
     @events ||= AppealEvents.new(appeal: self).all
   end
 
+  def api_alerts_have_decision?
+    fetch_status == :hlr_decision
+  end
+
+  def due_date_to_appeal_decision
+    # the deadline to contest the decision for this claim
+    return dta_decision_event_date + 365.days if dta_claim
```
leikkisa: This might get a little weird depending on when you use it, because you want to check now if there are any dta_claims, and unless they're all closed, the dta_decision_event_date will be nil.

mc-usds: I made and update so that there will be an alert for every remand sc decision that has a different decision date.

app/models/appeal.rb
mc-usds: I was of the understanding for this particular case we only want and alert for effectuations processed in VBMS. I will double check.

app/models/api_status_alerts.rb
```diff
@@ -0,0 +1,79 @@
...
+  def appeal_alerts
+    [
+      post_decision,
+      post_remand_decision,
+      post_effectuation
+    ].flatten.compact.uniq
+  end
+
+  def post_decision
+    return unless decision_review.api_alerts_have_decision?
```
cmgiven: Can you talk through the thought process for why this is necessary?

mc-usds: This is poor naming on my part. I am changing it to be api_alerts_show_decision_alert?

This is to handle the difference logic for when to show decisions when there are remands.
With HLR this alert will not show until the HLR is no longer active (remands have been resolved)
With Appeal this alert will show once the decisions are available, even when there are active remands. And the separate alert will show once the remands have been resolved.

app/models/higher_level_review.rb
```diff
@@ -42,10 +42,12 @@ def status_hash
   end
 
   def alerts
-    # need to implement. add logic to return alert enum
+    @alerts ||= ApiStatusAlerts.new(decision_review: self).all.sort_by { |alert| alert[:details][:decisionDate] }
   end
 
   def active_status?
+    # for the purposes for appeal status api, an HLR is considered active if there are
+    # still active remand claims.
```
leikkisa: Nice!

app/models/higher_level_review.rb
```diff
@@ -68,6 +70,31 @@ def events
     @events ||= AppealEvents.new(appeal: self).all
   end
 
+  def api_alerts_show_decision_alert?
+    # for HLR, only want to show the decision alert when the HLR is no longer active,
+    # meaning any remands have been resolved.
+    !active_status? || decision_issues.any?
```
leikkisa: Based on the comment, should this be && ?  The HLR can have decision issues while it is still has an active_status? of true if there are remands.

mc-usds: fixed.

PR #9429, opened 2019-02-14
Move Available Hearing Locations to Appeals

lomaxap:
Part 3 of moving AHLs to appeals 
see all #9442, #9503

- Updates `ScheduleHearingTasks.tasks_for_ro` to use appeals' closest RO
- Updates frontend dropdowns to use appeals data


app/controllers/hearings_controller.rb
```diff
@@ -35,12 +35,17 @@ def find_closest_hearing_locations
     begin
       HearingDayMapper.validate_regional_office(params["regional_office"])
 
-      veteran = Veteran.find_by(file_number: params["veteran_file_number"])
+      case params["appeal_type"]
+      when "ama"
+        appeal = Appeal.find_by(uuid: params["appeal_id"])
+      when "legacy"
+        appeal = LegacyAppeal.find_by(id: params["appeal_id"])
```
mdbenjam: Is it definitely the id, and not the vacols_id of the appeal. We generally use the vacols_id on the frontend, but not positive if we are here. If that is the case we have a method: `Appeal.find_appeal_by_id_or_find_or_create_legacy_appeal_by_vacols_id`

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -16,51 +16,22 @@ def find_or_create_if_eligible(appeal)
...
         Constants.TASK_STATUSES.in_progress.to_sym
       ).includes(:assigned_to, :assigned_by, :appeal, attorney_case_reviews: [:attorney])
+
       appeal_tasks = incomplete_tasks.joins("INNER JOIN appeals ON appeals.id = appeal_id")
-        .joins("INNER JOIN veterans ON appeals.veteran_file_number = veterans.file_number")
-        .where("veterans.closest_regional_office = ?", regional_office)
+        .where("appeals.closest_regional_office = ?", regional_office)
+
+      legacy_appeal_tasks = incomplete_tasks
+        .joins("INNER JOIN legacy_appeals ON legacy_appeals.id = appeal_id")
```
mdbenjam: these joins can now just be: `joins(:appeals)` and `joins(:legacy_appeals)`

client/app/components/DataDropdowns/AppealHearingLocations.jsx
```diff
@@ -97,7 +97,7 @@ class VeteranHearingLocationsDropdown extends React.Component {
 
     let url = '/hearings/find_closest_hearing_locations?regional_office=';
 
-    url += `${regionalOffice}&veteran_file_number=${veteranFileNumber}`;
+    url += `${regionalOffice}&appeal_id=${appealId}&appeal_type=${appealType}`;
```
mdbenjam: We can likely infer the type based on whether or not it's a uuid

app/controllers/hearings_controller.rb
lomaxap: ah, yea. I'm using the `id`, but should be using `external_id`

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -23,51 +23,21 @@ def find_or_create_if_eligible(appeal)
       end
     end
 
-    def legacy_tasks_for_ro(regional_office, incomplete_tasks)
-      # Joining to legacy appeals is more difficult because we don't store the veteran file number
-      # we only have the bfcorlid, which needs to be modified according to the
-      # LegacyAppeal.veteran_file_number_from_bfcorlid function. It is written here in SQL.
-      legacy_appeal_tasks = incomplete_tasks
-        .joins("INNER JOIN legacy_appeals ON legacy_appeals.id = appeal_id")
-        .joins("INNER JOIN veterans ON
```
mdbenjam: Getting rid of this feels soooooooo good

spec/controllers/hearings_controller_spec.rb
```diff
@@ -135,14 +135,29 @@
...
+      let!(:appeal) { create(:appeal) }
 
-    it "returns an address" do
-      get :find_closest_hearing_locations,
-          as: :json,
-          params: { veteran_file_number: "123456789", regional_office: "RO13" }
+      it "returns an address" do
+        get :find_closest_hearing_locations,
+            as: :json,
+            params: { appeal_id: appeal.external_id, appeal_type: "ama", regional_office: "RO13" }
```
mdbenjam: don't need appeal_type now

app/models/tasks/schedule_hearing_task.rb
lomaxap: FYI, when I change this, I get error 
```
 ActiveRecord::EagerLoadPolymorphicError:
       Cannot eagerly load the polymorphic association :assigned_to
```

mdbenjam: huh, that's weird, I guess just leave it as is then :/

PR #9431, opened 2019-02-14
Change review_request to decision_review on request issues

shanear:
Held back on additional refactoring/data fix efforts.

connects #8239


app/models/decision_review.rb
```diff
@@ -8,7 +8,7 @@ class DecisionReview < ApplicationRecord
 
   attr_reader :saving_review
 
-  has_many :request_issues, as: :review_request
+  has_many :request_issues, as: :decision_review
   has_many :claimants, as: :review_request
```
pkarman: do we care about the claimant association too? maybe in another PR?

shanear: yeah, I'd like to get to claimant, but this one is already supermassive

PR #9447, opened 2019-02-15
Exclude appeals that require(d) IHPs for the purposes of calculating …

cmgiven:
Noticed during testing that `DirectReviewDocket.new.time_until_due_of_oldest_appeal` was returning 0. Turns out that the receipt dates of the oldest ten appeals are: [Thu, 05 Apr 2018, Tue, 10 Apr 2018, Wed, 16 May 2018, Fri, 01 Jun 2018, Thu, 07 Jun 2018, Sat, 30 Jun 2018, Sun, 01 Jul 2018, Tue, 31 Jul 2018, Sun, 05 Aug 2018, Mon, 06 Aug 2018]

This is because they're blocked by IHPs, which the Board has yet to negotiate deadlines for.

The obvious fix is just to exclude non-ready appeals. This PR does that.

But when an IHP is completed for one of these old appeals, it will become ready, and the next time someone requests a distribution, the docket will suddenly appear dramatically older than it in fact is, overdistributing direct review appeals.

So this PR further proposes excluding appeals that have ever had an IHP task from the calculation.

I'll be talking this through with LRP in a bit. But wanted to get the PR in early.


app/models/appeal.rb
```diff
@@ -52,6 +52,13 @@ class Appeal < DecisionReview
               DistributionTask.name, Constants.TASK_STATUSES.assigned, 1)
   }
 
+  scope :non_ihp, lambda {
+    joins(:tasks)
+      .group("appeals.id")
+      .having("count(case when tasks.type = ? then 1 end) >= ?",
+              InformalHearingPresentationTask.name, 1)
```
amprokop: Should this line read `InformalHearingPresentationTask.name, 0)`

amprokop:
Sounds good on the approach — once we enforce an expiration time on IHPs, will this still be a problem?

app/models/appeal.rb
cmgiven: It's greater than or equal to, so 1 is what we're looking for. At least 1 task.

cmgiven:
That is the open question. If there is an expiration time, then yes, this is not a problem, although this fix probably doesn't cause harm in that world either.

app/models/appeal.rb
cmgiven: Oh, shoot, it should be equal to 0. D'oh.

Issue #9448, opened 2019-02-15
Intake | Veteran name failing to backfill


pkarman:
Several support tickets on this recently:
https://github.com/department-of-veterans-affairs/appeals-support/issues/4089

This is difficult to reproduce and current workaround is to call `Veteran.find_by_file_number_or_ssn(file_number, sync_name: true)` manually via console.

We suspect something amiss with the `Veteran.accessible?` logic but why that does not trigger an `:veteran_not_accessible` error is unclear.

pkarman:
Part of the problem (maybe) is that the `BGS.can_access?` method treats all these errors as the same (returns `false`)
```
BGS::ShareError: No Person record found

BGS::ShareError: findFlashes: FileNumber is invalid, must be 8 or 9 digits

BGS::ShareError: This is a Sensitive Record and you do not have high enough access to update it
```
so it's difficult to diagnose the exact cause of the failure.

pkarman:
Not getting reports of this anymore, closing subject to more reports.

PR #9452, opened 2019-02-15
Veteran find or create fix

pkarman:
connects #9448 

## Description

This change refactors the `Intake.build` method to take a `Veteran` instance instead of a file number. This should reduce the number of calls to upstream services.

This change also affects the logic for when a `Veteran` record is created. It will only be created if both `found?` and `accessible?` are true.

### DIFF
--- a/app/controllers/intakes_controller.rb
+++ b/app/controllers/intakes_controller.rb
@@ -117,7 +117,7 @@ class IntakesController < ApplicationController
   def new_intake
     @new_intake ||= Intake.build(
       user: current_user,
-      veteran_file_number: veteran_file_number,
+      veteran: veteran,
       form_type: params[:form_type]
     )
   end
@@ -126,10 +126,9 @@ class IntakesController < ApplicationController
     @intake ||= Intake.where(user: current_user).find(params[:id])
   end
 
-  def veteran_file_number
+  def veteran
     # param could be file number or SSN. Make sure we return file number.
-    veteran = Veteran.find_by_file_number_or_ssn(params[:file_number], sync_name: true)
-    veteran ? veteran.file_number : params[:file_number]
+    @veteran ||= Veteran.find_or_create_by_file_number_or_ssn(params[:file_number], sync_name: true)
   end
 
   def success_message
--- a/app/controllers/intakes_controller.rb
+++ b/app/controllers/intakes_controller.rb
@@ -117,8 +117,7 @@ class IntakesController < ApplicationController
   def new_intake
     @new_intake ||= Intake.build(
       user: current_user,
-      veteran: veteran,
-      veteran_file_number: params[:file_number],
+      veteran_or_file_number: veteran || params[:file_number],
       form_type: params[:form_type]
     )
   end
--- a/app/controllers/intakes_controller.rb
+++ b/app/controllers/intakes_controller.rb
@@ -127,7 +127,6 @@ class IntakesController < ApplicationController
   end
 
   def veteran
-    # param could be file number or SSN. Make sure we return file number.
     @veteran ||= Veteran.find_or_create_by_file_number_or_ssn(params[:file_number], sync_name: true)
   end
 
--- a/app/controllers/intakes_controller.rb
+++ b/app/controllers/intakes_controller.rb
@@ -118,6 +118,7 @@ class IntakesController < ApplicationController
     @new_intake ||= Intake.build(
       user: current_user,
       veteran: veteran,
+      veteran_file_number: params[:file_number],
       form_type: params[:form_type]
     )
   end
--- a/app/models/intake.rb
+++ b/app/models/intake.rb
@@ -31,6 +31,7 @@ class Intake < ApplicationRecord
   }.freeze
 
   attr_reader :error_data
+  attr_writer :veteran
 
   def self.in_progress
     where(completed_at: nil).where(started_at: IN_PROGRESS_EXPIRES_AFTER.ago..Time.zone.now)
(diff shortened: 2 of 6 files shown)


app/controllers/intakes_controller.rb
```diff
@@ -126,10 +126,9 @@ def intake
     @intake ||= Intake.where(user: current_user).find(params[:id])
   end
 
-  def veteran_file_number
+  def veteran
     # param could be file number or SSN. Make sure we return file number.
-    veteran = Veteran.find_by_file_number_or_ssn(params[:file_number], sync_name: true)
-    veteran ? veteran.file_number : params[:file_number]
+    @veteran ||= Veteran.find_or_create_by_file_number_or_ssn(params[:file_number], sync_name: true)
```
pkarman: I don't think we actually want to do this for a search, because it might populate our db with invalid file numbers.

app/models/veteran.rb
```diff
@@ -272,19 +271,17 @@ def create_by_file_number(file_number)
         %(create_by_file_number file_number:#{file_number} found:true accessible:#{veteran.accessible?})
       )
 
+      return veteran unless veteran.accessible?
+
       before_create_veteran_by_file_number # Used to simulate race conditions
       veteran.tap do |v|
-        v.update!(participant_id: v.ptcpnt_id)
```
pkarman: it's not clear to me why we would always update the participant_id even if the veteran was not accessible to the user.

app/models/intake.rb
```diff
@@ -40,13 +41,14 @@ def self.expired
...
-  def self.build(form_type:, veteran_file_number:, user:)
+  def self.build(form_type:, veteran:, user:)
     intake_classname = FORM_TYPES[form_type.to_sym]
 
     fail FormTypeNotSupported unless intake_classname
 
     intake_classname.constantize.new(
-      veteran_file_number: veteran_file_number,
+      veteran: veteran,
+      veteran_file_number: veteran.try(:file_number),
```
jiahuang: shouldn't a veteran with a file number always exist during an intake?

app/controllers/intakes_controller.rb
```diff
@@ -117,7 +117,7 @@ def intake_in_progress
   def new_intake
     @new_intake ||= Intake.build(
       user: current_user,
-      veteran_file_number: veteran_file_number,
+      veteran: veteran,
```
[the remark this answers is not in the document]
pkarman: > Is this also a decision to stop populating the veteran_file_number field in the Intakes table?

No. Just an optimization to avoid hitting BGS multiple times in a single request.

app/models/veteran.rb
pkarman: `.accessible?` is supposed to mean `.authorized?` but it currently has multiple meanings that aren't always clear.

pkarman:
I have added a test to make sure `intake.veteran_file_number` is being set.

I think there's a valid argument to be made that storing a `veteran_file_number` on any Caseflow record w/o a corresponding `Veteran` record is broken behavior. We don't really treat `veteran_file_number` like a strict foreign key, but there are plenty of places we do:

```
@veteran ||= Veteran.find_by(file_number: veteran_file_number)
```

and that will stop working in the case of Intake that has an error on it.

OTOH, we have nearly 30% of our intakes in production with errors on them, which it seems we track for reporting/statistical purposes.

app/controllers/intakes_controller.rb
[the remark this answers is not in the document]
leikkisa: Ok! As long as it still goes there then I'm more okay with this. I saw the veteran_file_number in the more specific intake build, but not in this build so I wasn't sure if it would make it through to the database.

spec/controllers/intakes_controller_spec.rb
```diff
@@ -21,6 +21,54 @@
...
+      end
+
+      let(:file_number) { "999887777" }
+      let!(:veteran) {} # no-op
+
+      it "does not create Veteran db record in Caseflow" do
+        expect(Veteran.find_by_file_number_or_ssn(file_number)).to be_nil
+        post :create, params: { file_number: file_number, form_type: "higher_level_review" }
+        expect(response.status).to eq(422)
+        expect(controller.send(:new_intake).error_code).to eq("veteran_not_accessible")
```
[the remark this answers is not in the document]
leikkisa: 👍

leikkisa:
As for not storing Veteran file number if there's not a corresponding Veteran, I do agree with you, I just feel a little conservative because I use the file number a lot for trouble shooting.  This is way less useful if the Veteran doesn't exist (in which case I think it'd only be in the Intakes table), but if it did exist once upon a time we might still be able to get useful information (like we might have the participant ID on the old file number, which we could use to find the new file number if it changes), or it's still useful if it's not accessible.

app/models/intake.rb
```diff
@@ -40,12 +41,16 @@ def self.expired
     where(completed_at: nil).where(started_at: Time.zone.at(0)...IN_PROGRESS_EXPIRES_AFTER.ago)
   end
 
-  def self.build(form_type:, veteran_file_number:, user:)
+  def self.build(form_type:, veteran_or_file_number:, user:)
```
shanear: I don't love that we modifed this only for the tests. It makes the api a lot weirder.

shanear: It also adds some unnecessary complexity

app/controllers/intakes_controller.rb
```diff
@@ -117,7 +117,7 @@ def intake_in_progress
   def new_intake
     @new_intake ||= Intake.build(
       user: current_user,
-      veteran_file_number: veteran_file_number,
+      veteran_or_file_number: veteran || params[:file_number],
```
shanear: This change feels clunky to me. It doesn't seem like it's needed to solve the problem. I also thought BGS stuff was cached so this doesn't save any calls either

pkarman:
closing in favor of #9660

PR #9455, opened 2019-02-15
Add created_at to decision issues

shanear:
This field is on `request_issues` and even though it's not used, it can be helpful in debugging and understanding the data. So we should probably add it.

connects #8239


shanear:
> Do we want to backfill existing rows?

I was thinking about this, and my thought is no, since we don't have the accurate value. and nil is better than wrong lol.

PR #9464, opened 2019-02-15
groups priority appeals by appeal id

amprokop:
Testing:

In prod, redefined `Appeal.all_priority`, then checked that `Appeal.all_priority.uniq.length == Appeal.all_priority.length`


app/models/docket.rb
```diff
@@ -13,7 +13,8 @@ def appeals(priority: nil, ready: nil)
 
     if priority == true
       scope = scope.merge(Appeal.all_priority)
-      return scope.merge(Appeal.ordered_by_distribution_ready_date)
+      # TODO: remove the `.uniq` when Appeal.all_priority is fixed to return unique results.
+      return scopes.merge(Appeal.ordered_by_distribution_ready_date).uniq
```
cmgiven: Makes sense. I assume you've tested to confirm this solves the issue? Because it's possible that the entries are not unique by virtue of having been joined against different rows from other tables.

amprokop: I have tested that when adding `.uniq` 7 fewer appeals are returned in prod, but i'll double check to ensure that there are no duplicate rows by appeal id just in case.

amprokop:
Bugger — since other methods are chained together with this, i can't insert `uniq` here

amprokop:
Found the correct fix — adding a group by statement removes dupes.

Issue #9386, opened 2019-02-13
Decision Sync | Non-disability issues register as non rating issues when creating DTA error EPs


shanear:
We found an issue in UAT where an HLR with 3 rating issues was recorded all 3 with DTA error.

The resulting Supplemental claim created 2 of the 3 issues in a non-rating EP, which is incorrect. They are all rating issues.

The common factor between the affected issues is that they are both non-disability issues. meaning they don't have a diagnostic code.

```
[#<RequestIssue:0x000056039fa4f6e8
  id: 1321,
  review_request_type: "SupplementalClaim",
  review_request_id: 395,
  rating_issue_reference_id: "123971",
  rating_issue_profile_date: Tue, 12 Feb 2019 14:27:53 UTC +00:00,
  contention_reference_id: 316104,
  description: nil,
  issue_category: nil,
  decision_date: Tue, 12 Feb 2019,
  disposition: nil,
  end_product_establishment_id: 876,
  removed_at: nil,
  rating_issue_associated_at: Wed, 13 Feb 2019 14:52:44 UTC +00:00,
  parent_request_issue_id: nil,
  notes: nil,
  is_unidentified: nil,
  ineligible_due_to_id: nil,
  untimely_exemption: nil,
  untimely_exemption_notes: nil,
  ineligible_reason: nil,
  ramp_claim_id: nil,
  decision_sync_submitted_at: nil,
  decision_sync_attempted_at: nil,
  decision_sync_processed_at: nil,
  decision_sync_error: nil,
  vacols_id: nil,
  vacols_sequence_id: nil,
  created_at: Wed, 13 Feb 2019 14:52:21 UTC +00:00,
  benefit_type: "compensation",
  contested_decision_issue_id: 140,
  veteran_participant_id: "600237113",
  contested_rating_issue_diagnostic_code: nil,
  decision_review_type: "SupplementalClaim",
  decision_review_id: 395,
  contested_rating_issue_reference_id: "123971",
  contested_rating_issue_profile_date: "2019-02-12 14:27:53 UTC",
  contested_issue_description:
   "A duty to assist error has been identified during the Higher Level Review for Schizophrenia.",
  nonrating_issue_description: nil,
  unidentified_issue_text: nil,
  contested_rating_issue_disability_code: nil,
  last_submitted_at: nil,
  closed_at: nil,
  closed_status: nil>,
 #<RequestIssue:0x000056039fa220d0
  id: 1322,
[...]

PR #9466, opened 2019-02-16
Add dta request issues originally from rating request issues to rating end product

leikkisa:
connects #9386


config/initializers/extend_hash.rb
```diff
@@ -0,0 +1,13 @@
+class Hash
+  def find_all_values_for(key)
+    result = []
+    result << self[key]
+    self.values.each do |hash_value|
+      values = [hash_value] unless hash_value.is_a? Array
```
pkarman: I think `values = [hash_value].flatten` would do the right thing too.

spec/models/request_issue_spec.rb
```diff
@@ -320,6 +320,29 @@
           context "when rating" do
             let(:request_issue) { rating_request_issue }
             it { is_expected.to eq "040HDERPMC" }
+
+            context "when missing rating_issue_reference_id but comes from a previous rating request issue" do
```
pkarman: should this comment refer to "contested_rating_issue_reference_id" instead?

config/initializers/extend_hash.rb
```diff
@@ -0,0 +1,13 @@
+class Hash
```
pkarman: it would be good to see an explicit test for this, maybe in `spec/initializers/extend_hash_spec.rb`

I know there's a test that exercises it, but it would be good to see some tests so that if the end product structure ever changes we could rest easy that tests would fail.

spec/models/request_issue_spec.rb
leikkisa: Good call! I think that got renamed recently.

config/initializers/extend_hash.rb
leikkisa: 👍

config/initializers/extend_hash.rb
leikkisa: 👍

PR #9473, opened 2019-02-18
Consider an EP preexisting only if it is active

pkarman:
connects #9476 

Related to:

* https://github.com/department-of-veterans-affairs/appeals-support/issues/3938
* https://github.com/department-of-veterans-affairs/caseflow/issues/7638

### DIFF
--- a/spec/models/ramp_election_intake_spec.rb
+++ b/spec/models/ramp_election_intake_spec.rb
@@ -311,6 +311,7 @@ describe RampElectionIntake do
           expect(intake.detail.established_at).to_not be_nil
           expect(veteran_end_products.count).to eq 3
           expect(veteran_end_products.map(&:claim_id)).to include(preexisting_ep.claim_id)
+          expect(intake.detail.end_product_establishment.reference_id).to_not eq(preexisting_ep.claim_id)
         end
       end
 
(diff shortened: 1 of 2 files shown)


leikkisa:
I think this is correct. I believe it's okay to make a new one once the old one is canceled or closed, and it definitely makes more sense than to connect a new intake to a closed EP.  

For AMA we are able to re-use the modifiers once that happens.  I guess this was put in there to prevent duplicates in the first place, so it's interesting to me that there would be one found with the same claim_date, but maybe that's one way to handle mistakes (to cancel it and re-create it)...?

spec/models/ramp_election_intake_spec.rb
```diff
@@ -274,15 +276,41 @@
...
+          expect(veteran_end_products.count).to eq 2
+          expect(veteran_end_products.map(&:claim_id)).to include(preexisting_ep.claim_id)
+
           subject
 
           expect(intake.reload).to be_success
           expect(intake.detail).to_not eq(existing_ramp_election)
           expect(intake.detail.established_at).to_not be_nil
+          expect(veteran_end_products.count).to eq 3
+          expect(veteran_end_products.map(&:claim_id)).to include(preexisting_ep.claim_id)
```
leikkisa: I guess this is implied by the fact that there are now three end products instead of two, but it might make sense to check that the new RAMP Election's EPE's reference_id is not equal to the preexisting_ep.claim_id in this case.

PR #9474, opened 2019-02-18
Add decision timeliness to decision_in_progress details

### LINKED ISSUE #9430: API: decisionInProgress has details object (post-MVP)
In the status decisionInProgress, the detail object should contain:
- `decisionTimeliness`

```
"status": {
  "type": "decisionInProgress",
  "details": {
    "decisionTimeliness": "..."
  }
}
```

cmgiven:
Resolves #9430 

### Description
This is needed for MVP as the front-end will expect to have this information in order to display the `decision_in_progress` status.

### DIFF
--- a/app/models/appeal.rb
+++ b/app/models/appeal.rb
@@ -403,6 +403,7 @@ class Appeal < DecisionReview
   # rubocop:enable Metrics/PerceivedComplexity
 
   # rubocop:disable Metrics/MethodLength
+  # rubocop:disable CyclomaticComplexity
   def fetch_details_for_status
     case fetch_status
     when :bva_decision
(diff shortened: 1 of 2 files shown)


app/models/appeal.rb
```diff
@@ -417,25 +418,30 @@ def fetch_details_for_status
       post_bva_dta_decision_status_details
     when :bva_decision_effectuation
       {
-        bvaDecisionDate: decision_event_date,
-        aojDecisionDate: decision_effectuation_event_date
+        bva_decision_date: decision_event_date,
+        aoj_decision_date: decision_effectuation_event_date
```
cmgiven: These are converted to CamelCase by the serializer, but we can keep them in snake_case, according to Ruby convention, until that point.

mc-usds: Ah ok. CamelCase vs snake_case gets me again.
I think there are a couple of unit tests in spec/models/appeal_spec.rb that check "bvaDecisionDate" and "aojDecisionDate" that will need to also be updated.

cmgiven: Ah, good catch.

PR #9477, opened 2019-02-18
Refactor dates/timezones in specs

pkarman:
connects #9488 

## Description

Part of an ongoing spec audit to use meaningful date/time names and relative timezones.

Per the ticket:

* avoid hardcoded date strings
* avoid hardcoded time zones
* avoid constructing dates without meaningful names

Discouraged in individual specs:

* `Time.zone = 'America/New_York' # instead, let the spec config handle it`
* `Date.new(2019, 1, 1) # instead use the dates predefined in spec/support/date_time_helper.rb`
* `"02/19/2019" # instead use Time.zone.today.mdY`


app/models/vacols/case_hearing.rb
```diff
@@ -115,9 +115,9 @@ def create_hearing!(hearing_info)
     end
 
     def create_child_hearing!(hearing_info)
-      MetricsService.record("VACOLS: create_hearing!",
+      MetricsService.record("VACOLS: create_child_hearing!",
```
pkarman: make the logging match the method name.

scripts/enable_features_dev.rb
```diff
@@ -54,6 +54,10 @@
           feature: "use_representative_info_from_bgs",
           enable_all: true
         },
+        {
+          feature: "use_ama_activation_date",
```
pkarman: we are currently *not* using this in UAT/prod but enabling locally so that we can start to make sure we are testing against it.

spec/support/intake_helpers.rb
```diff
@@ -122,7 +122,6 @@ def setup_intake_flags
     FeatureToggle.enable!(:intakeAma)
     FeatureToggle.enable!(:intake_legacy_opt_in)
 
-    Time.zone = "America/New_York"
```
pkarman: this is now set globally in the rails_helper.rb

spec/controllers/hearings_controller_spec.rb
```diff
@@ -85,19 +85,25 @@
...
             id: legacy_hearing.external_id, hearing: params, master_record_updated: master_record_params
           }
           expect(response.status).to eq 200
-
           expect(LegacyHearing.last.location.facility_id).to eq "vba_301"
-          expect(VACOLS::CaseHearing.find_by(vdkey: hearing_day[:id]).hearing_date).to eq(
-            Time.new(2019, 4, 2, 10).in_time_zone("Eastern Time (US & Canada)")
-          )
+
+          # VACOLS thinks it is UTC, but the values written to it are Eastern.
```
pkarman: I think this test change reflects zero feature change, but since this is the only test that was failing, I'm not 100% certain of that.

spec/rails_helper.rb
```diff
@@ -49,7 +49,8 @@
   Dir.mkdir cache_directory
 end
 
-ENV["TZ"] ||= "America/New York"
+ENV["TZ"] ||= "America/New_York"
+Time.zone = ENV["TZ"]
```
pkarman: longer term we'll want to use something like https://github.com/alindeman/zonebie to randomize timezone and allow us to verify browser timezones may vary (e.g. https://github.com/kbaum/browser-timezone-rails)

spec/feature/intake/appeal/edit_spec.rb
```diff
@@ -8,8 +8,7 @@
     FeatureToggle.enable!(:intakeAma)
     FeatureToggle.enable!(:intake_legacy_opt_in)
 
-    Time.zone = "America/New_York"
-    Timecop.freeze(Time.utc(2018, 5, 26))
+    Timecop.freeze(post_ama_start_date)
```
jiahuang: 👍

spec/feature/intake/appeal/edit_spec.rb
```diff
@@ -31,8 +30,8 @@
     User.authenticate!(roles: ["Mail Intake"])
   end
 
-  let(:receipt_date) { Time.zone.today - 20 }
-  let(:profile_date) { "2017-11-02T07:00:00.000Z" }
+  let(:receipt_date) { Time.zone.today - 20.days }
```
jiahuang: I think receipt date is supposed to be a date, and profile_date is supposed to be a datetime.

spec/feature/intake/appeal_spec.rb
```diff
@@ -42,15 +41,15 @@
...
-  let(:future_date) { Time.zone.now + 30.days }
+  let(:future_date) { (Time.zone.now + 30.days).to_date }
 
-  let(:receipt_date) { post_ramp_start_date - 30.days }
+  let(:receipt_date) { (post_ramp_start_date - 30.days).to_date }
 
   let(:untimely_days) { 372.days }
 
-  let(:profile_date) { post_ramp_start_date.to_datetime - 35.days }
+  let(:profile_date) { (post_ramp_start_date - 35.days).to_date }
```
jiahuang: I think vbms gives us profile dates as datetimes

spec/rails_helper.rb
jiahuang: Can you add in a comment in here about how this will also change the spawned chrome browser timezone?

spec/support/date_time_helper.rb
```diff
@@ -18,6 +18,14 @@ def ramp_start_date
   def post_ramp_start_date
     Time.new(2017, 12, 8).in_time_zone
   end
+end
+
+class Date
+  # rubocop:disable Naming/MethodName
+  def mdY
```
jiahuang: nice!

PR #9482, opened 2019-02-19
Remove ama_decision_issues flag and remove all old code

### LINKED ISSUE #9306: Clean up ama_decision_issues feature flag
We use feature flags to enable us to push new features to production and enable these features when we are ready. 

`ama_decision_issues` is enabled for all so it would be good to get rid of `ama_decision_issues` flag in the code and any old code around that.

AC:
1. Clean up FE and BE code in Caseflow
2. Remove this feature from appeals-deployment repo.

aroltsch:
Resolves #9306 

### Description
We enabled `ama_decision_issues` flag for all users so we can get rid of the old code.

### DIFF
--- a/client/app/queue/EvaluateDecisionView.jsx
+++ b/client/app/queue/EvaluateDecisionView.jsx
@@ -133,8 +133,7 @@ class EvaluateDecisionView extends React.PureComponent {
       checkoutFlow,
       decision,
       userRole,
-      appealId,
-      amaDecisionIssues
+      appealId
     } = this.props;
 
     let loc = 'bva_dispatch';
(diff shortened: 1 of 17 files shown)


client/app/queue/EvaluateDecisionView.jsx
```diff
@@ -337,8 +337,7 @@ const mapStateToProps = (state, ownProps) => {
     task: taskById(state, { taskId: ownProps.taskId }),
     decision: state.queue.stagedChanges.taskDecision,
     userRole: state.ui.userRole,
-    error: state.ui.messages.error,
-    amaDecisionIssues: state.ui.featureToggles.ama_decision_issues || !_.isEmpty(appeal.decisionIssues)
```
mdbenjam: Do we use this flag somewhere in this file?

aroltsch: Corrected!

PR #9496, opened 2019-02-19
Remove flow from files

### LINKED ISSUE #9305: Remove Flow from front end
AC:

Flow is removed from FE

youngfreezyVA:
Resolves #9305

### Description
Removes Flow from the frontend

### Acceptance Criteria
- [ ]no regression issues introduced by removing flow


youngfreezyVA:
@aroltsch in order to get the lint task to pass the circle ci command that runs `lint.rake` needs to be updated to match this: https://github.com/department-of-veterans-affairs/caseflow/commit/7ca8e49e0dcd0adfc50beb3d8ab1de6de1518f17

mdbenjam:
Did you have to do this by hand? Or was there a script that removes flow?

youngfreezyVA:
I did this by hand and re-ran the unit tests in between every few commits. If there was a script I wish I knew about

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

spec/feature/queue/ama_queue_spec.rb
```diff
@@ -692,114 +692,6 @@ def valid_document_id
         )
       end
 
-      step "judge sees the case in their review queue" do
```
mdbenjam: Why do we need to remove these tests?

spec/factories/task.rb
```diff
@@ -112,6 +112,11 @@
       parent { create(:ama_judge_task) }
     end
 
+    factory :ama_attorney_rewrite_task do
```
mdbenjam: Why do we need to add this factory?

mdbenjam:
I thought I remember Evan saying there was a removal script when he added Flow, but I'm not actually sure what script that is 🤷‍♂️ . Added two comments, but besides that this looks good to me.

spec/feature/queue/ama_queue_spec.rb
youngfreezyVA: I branched off of another branch (that is now merged into master) so not sure why this is showing up in the diff, will resolve

Issue #9310, opened 2019-02-08
Add action to cancel task in actions dropdown


lowellrex:
Every single AMA task should have the option to cancel it which will result in the task being updated with the status `cancelled`.

not implemented:
- `HearingAdminActionTask`: @lomaxap asks us to defer this to later while they map out the rest of the flows

not cancellable:
- `JudgeDecisionReviewTask`
- `JudgeAssignTask`

🚫Blocked by #8701.

lowellrex:
MEDIUM

Issue #7612, opened 2018-10-30
Judge to Judge assignment of RAMP tasks fails


amprokop:
![image](https://user-images.githubusercontent.com/3781835/47740016-1d7b1c00-dc4d-11e8-86c3-7d570632b06e.png)

Since RAMP task assignment was built to create an AttorneyTask, it is a POST to `/tasks`. This correctly fails validation if the assigned to user is a judge. Right now, judges cannot push RAMP tasks to other judges. Unclear if anyone has tried this in production.

Possibilities for remediation:

# Quick fix
- on POST to `/tasks`, if a judge is the assigning user and the user being assigned, instead modify the existing JudgeTask (easy fix, adds tech debt)

# Thorough fix
- `AttorneyRepository.find_all_attorneys` currently (and perhaps unexpectedly) returns judges, acting judges, and attorneys. It queries for staff records where `sattyid` is not null, but judges and attorneys both have attorney ids. Return the type of user to the client. If the user is an attorney, POST to `/tasks` to create an AttorneyTask. If the user is a judge, PATCH `/tasks/:id` to change the assigned to user on the existing JudgeTask. If the user is an acting judge, POST to `/tasks` to create an AttorneyTask (later, we'll likely have to disambiguate assignment to acting judge acting as a judge vs acting judge acting as attorney)

lowellrex:
I am a little confused about what the problem here is. As I understand the scenario above, a judge is trying to assign a case to an attorney for them to draft a decision but we are wrongly identifying that attorney as a judge. So is the problem that attorneys are being misidentified as judges?

amprokop:
Nope.

The class and method names in the codebase implies that we’re only
returning attorneys, so that’s misleading.

But if you look at the Assign Tasks view, it says something like “select an
attorney or a judge” when you select “Other” in the dropdown- we intended
to support judge to judge assignment but it does not work.

lowellrex:
So would judge X be asking judge Z to draft a decision in this example? Or is judge X just trying to change the ownership of this case to judge Z so judge Z can ask one of their attorneys to draft a decision?

lowellrex:
Gotcha gotcha. I didn't know that we supported this behaviour.

[...]

PR #9499, opened 2019-02-19
add cancel task option to all AMA task models

joeyyang:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/9310

### Description
Adds the option to cancel task to all AMA task models (`judge_task.rb`, `generic_task.rb`, etc., excludes `attorney_task.rb` and `{any}_legacy_task.rb`)

Example:
screenshot 2019-02-19 11 33 40


app/models/tasks/generic_task.rb
```diff
@@ -24,22 +24,24 @@ def available_actions(user)
       return [
         Constants.TASK_ACTIONS.ASSIGN_TO_TEAM.to_h,
         Constants.TASK_ACTIONS.REASSIGN_TO_PERSON.to_h,
-        Constants.TASK_ACTIONS.CANCEL_TASK.to_h,
-        Constants.TASK_ACTIONS.MARK_COMPLETE.to_h
+        Constants.TASK_ACTIONS.MARK_COMPLETE.to_h,
+        Constants.TASK_ACTIONS.CANCEL_TASK.to_h
```
joeyyang: make `CANCEL_TASK` go last everywhere it shows up

app/models/tasks/judge_task.rb
```diff
@@ -1,6 +1,6 @@
 class JudgeTask < Task
   def available_actions(user)
-    additional_available_actions(user).unshift(Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h)
+    additional_available_actions(user).unshift(Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h).push(Constants.TASK_ACTIONS.CANCEL_TASK.to_h)
```
joeyyang: there's gotta be a better way to do this

lowellrex: I think this is fine. Though I guess we could do something like the following if you prefer:

```ruby
[
  Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h,
  additional_available_actions(user),
].flatten
```

joeyyang: i like this much better!

app/models/tasks/judge_task.rb
```diff
@@ -1,6 +1,10 @@
 class JudgeTask < Task
   def available_actions(user)
-    additional_available_actions(user).unshift(Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h)
+    [
+      Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h,
+      additional_available_actions(user),
+      Constants.TASK_ACTIONS.CANCEL_TASK.to_h
```
joeyyang: ask lauren

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -110,7 +110,8 @@ def available_actions(user)
       return [
         Constants.TASK_ACTIONS.SCHEDULE_VETERAN.to_h,
         Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h,
-        Constants.TASK_ACTIONS.WITHDRAW_HEARING.to_h
+        Constants.TASK_ACTIONS.WITHDRAW_HEARING.to_h,
```
joeyyang: removing cancel option here, since "Withdraw hearing" also has `"value": "modal/cancel_task"`, which fails `app/models/task.rb#available_actions_unwrapper:58`

app/models/tasks/colocated_task.rb
```diff
@@ -45,7 +45,8 @@ def available_actions(user)
 
     available_actions_with_conditions([
                                         Constants.TASK_ACTIONS.PLACE_HOLD.to_h,
-                                        Constants.TASK_ACTIONS.ASSIGN_TO_PRIVACY_TEAM.to_h
+                                        Constants.TASK_ACTIONS.ASSIGN_TO_PRIVACY_TEAM.to_h,
+                                        Constants.TASK_ACTIONS.CANCEL_TASK.to_h
```
lowellrex: I think we need to update `ColocatedTask.update_location_in_vacols` to allow the cancellation of tasks to return the case to the correct location in VACOLS (as opposed to just the completion of tasks, as it is written now).

lowellrex:
Looks good for the most part! Just a few thoughts:

1. Let's confirm with @laurjpeterson that we are alright to delay adding the ability to cancel `AttorneyTask`s.
2. Let's determine what it means for a `JudgeTask` to be cancelled. What happens if a judge cancels a `JudgeAssignTask`, should it return to the basket of cases to be distributed? Should we allow them to do this?
3. Let's update `ColocatedTask` to make sure we charge cancelled tasks associated with `LegacyAppeal`s back to the appropriate location in VACOLS.
4. Let's add tests for each of these scenarios to confirm that the states of parent tasks change as expected when we cancel child tasks.

joeyyang:
@laurjpeterson should a judge be able to cancel any kind of `JudgeTask` (types are `JudgeQualityReviewTask`, `JudgeAssignTask`, `JudgeDecisionReviewTask`)?  @lowellrex and I thought that `JudgeQualityReviewTask`s were the only ones that could be cancelled in any meaningful way.

joeyyang:
WIP: logic for if a task is cancelled?
- for a `ColocatedTask`, it should go back to the attorney who assigned it
  - for a `LegacyAppeal`, it should go back to whoever assigned it (look in VACOLS)
- for a `QualityReviewTask`, it should create a `BvaDispatchTask`? (should it be cancellable?)
- for a `GenericTask`, 
- for a `HearingAdminActionTask`
- for a `InformalHearingPresentationTask`
- for a `JudgeQualityReviewTask`

laurjpeterson:
> 1. Let's confirm with @laurjpeterson that we are alright to delay adding the ability to cancel `AttorneyTask`s.
- Sure. Judges can just reassign attorney cases if they need to be taken off attorneys' plates.

> @laurjpeterson should a judge be able to cancel any kind of `JudgeTask` (types are `JudgeQualityReviewTask`, `JudgeAssignTask`, `JudgeDecisionReviewTask`)?  
> @lowellrex and I thought that `JudgeQualityReviewTask`s were the only ones that could be cancelled in any meaningful way and would like to propose that only `JudgeQualityReviewTask`s be cancellable.
- Def can't cancel a JudgeAssignTask, because it was assigned to the judge in docket order so it can't go backwards
- I agree that JudgeQR is cancellable
- I agree that shouldn't allow cancelling JudgeDecisionReview task. I might be persuaded otherwise, because I could see cases get assigned to a judge, but the judge goes on leave or travel board, and therefore can't actually sign that case. so they want to cancel it so their attorney can assign it to someone else. other ways to accomplish this are for judges to reassign cases to other judges (though I think this fails right now - https://github.com/department-of-veterans-affairs/caseflow/issues/7612) or for certain caseflow users to be able to reassign cases to other judges

app/models/tasks/colocated_task.rb
```diff
@@ -88,10 +89,12 @@ def create_and_auto_assign_child_task(_options = {})
...
-    if saved_change_to_status? &&
-       completed? &&
-       appeal_type == LegacyAppeal.name &&
-       all_tasks_completed_for_appeal?
+    is_closed_legacy_appeal = saved_change_to_status? &&
+                              !active? &&
+                              appeal_type == LegacyAppeal.name &&
+                              all_tasks_closed_for_appeal?
+
+    if is_closed_legacy_appeal
```
lowellrex: Can we rename this variable to something like `all_colocated_tasks_for_legacy_appeal_complete` or maybe even go back to not assigning the value of those conditions to a variable at all so we can sidestep the issue of precise variable naming in this case?

app/models/tasks/attorney_task.rb
```diff
@@ -24,7 +24,10 @@ def available_actions(user)
                             else
                               Constants.TASK_ACTIONS.REVIEW_LEGACY_DECISION.to_h
                             end
-    [review_decision_label, Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h]
+    [
+      review_decision_label,
+      Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h
+    ]
```
lowellrex: I think we can undo this change for a tidier commit if we'd like.

app/models/tasks/judge_task.rb
```diff
@@ -4,7 +4,10 @@
 
 class JudgeTask < Task
   def available_actions(user)
-    additional_available_actions(user).unshift(Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h)
+    [
+      Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h,
+      additional_available_actions(user)
+    ].flatten
```
lowellrex: I think we can undo this change for a tidier commit if we'd like.

app/models/tasks/attorney_task.rb
joeyyang: I think showing each set of actions on its own line is much more readable, if that's ok!

PR #9503, opened 2019-02-19
Swap AHL Job

lomaxap:
Part 2 of switching to appeals for AHL
see also #9442, #9429 

- updates AHL job to update appeals closest_regional_office and available_hearing_locations
- adds appellant method to legacy_appeal
- pulls BGS address for veteran if they are the appellant and the BGS record exists
- completing VerifyAddress admin actions updates appeals


app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -4,129 +4,156 @@ class FetchHearingLocationsForVeteransJob < ApplicationJob
...
+    ro = fetch_and_update_ro_for_appeal(appeal, va_dot_gov_address: va_dot_gov_address)
     facility_ids = facility_ids_for_ro(ro[:closest_regional_office])
-    AvailableHearingLocations.where(veteran_file_number: veteran.file_number).destroy_all
+    AvailableHearingLocations.where(appeal: appeal).destroy_all
 
     if facility_ids.length == 1
-      create_available_location_by_file_number(veteran.file_number, facility: ro[:facility])
+      create_available_location_for_appeal(appeal, facility: ro[:facility])
     else
+      sleep 1
```
lomaxap: resolves random VADotGovLimitError sentry alerts

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -4,129 +4,156 @@ class FetchHearingLocationsForVeteransJob < ApplicationJob
...
-      .where("admin_actions.parent_id IS NULL")
-      .map { |task| task.appeal.veteran_file_number }.compact
+  def self.get_appellant_address(appeal)
+    appeal.is_a?(LegacyAppeal) ? appeal.appellant[:address] : appeal.appellant.address
   end
 
-  def missing_veteran_file_numbers
-    existing_veteran_file_numbers = Veteran.where(file_number: file_numbers).pluck(:file_number)
-    (file_numbers - existing_veteran_file_numbers)[0, (QUERY_LIMIT - existing_veteran_file_numbers.length)] || []
+  def self.validate_appellant_address(appeal)
```
lomaxap: adding a few class methods for use in HearingLocation dropdown endpoint (see (hearings_controller)[https://github.com/department-of-veterans-affairs/caseflow/pull/a9429/files#diff-8a2c889e412c17fbba7f79b5f1ac6624R48])

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -4,129 +4,156 @@ class FetchHearingLocationsForVeteransJob < ApplicationJob
...
+      LegacyAppeal.find_or_create_by_vacols_id(bfkey)
+    end
+  end
+
+  def find_appeals_ready_for_geomatching(appeal_type)
+    # Appeals that have not had an available_hearing_locations updated in the last week
+    # and have an active ScheduleHearingTask
+    # that is not blocked by an VerifyAddress or ForeignVeteranCase admin action
+    appeal_type.left_outer_joins(:available_hearing_locations)
+      .where("#{appeal_type.table_name}.id IN (SELECT t.appeal_id FROM tasks t
```
mdbenjam: Not sure I understand all this. (I'm bad at SQL)

Why can't we just look for appeals with schedule hearing tasks that aren't completed or cancelled?

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -4,129 +4,156 @@ class FetchHearingLocationsForVeteransJob < ApplicationJob
 
   QUERY_LIMIT = 500
 
-  def veterans
-    @veterans ||= Veteran.where("file_number IN (?)", file_numbers + file_numbers_from_tasks)
-      .left_outer_joins(:available_hearing_locations)
+  def appeal_bfkeys
```
mdbenjam: I'm not sure we need to do all these queries anymore since we're creating schedule hearing tasks for all of these appeals.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
mdbenjam: Interesting. so did you not want to move this to appeal because we have two different types of appeals? Maybe doing a service model like we did with document fetcher: https://github.com/department-of-veterans-affairs/caseflow-certification/blob/master/app/services/document_fetcher.rb#L1

app/jobs/fetch_hearing_locations_for_veterans_job.rb
lomaxap: I guess we could, but we'd still need to add logic somewhere to exclude appeals that are blocked by a ForeignVetCase or VerifyAddress action . This is just feels simpler to me. "only get the appeals we are concerned with" rather than excluding them in the application logic.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
lomaxap: Thought that might be case. Was covering my bases.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -4,129 +4,156 @@ class FetchHearingLocationsForVeteransJob < ApplicationJob
...
+    # and have an active ScheduleHearingTask
+    # that is not blocked by an VerifyAddress or ForeignVeteranCase admin action
+    appeal_type.left_outer_joins(:available_hearing_locations)
+      .where("#{appeal_type.table_name}.id IN (SELECT t.appeal_id FROM tasks t
+          LEFT OUTER JOIN tasks admin_actions
+          ON t.id = admin_actions.parent_id
+          AND admin_actions.type IN ('HearingAdminActionVerifyAddressTask', 'HearingAdminActionForeignVeteranCaseTask')
+          AND admin_actions.status NOT IN ('cancelled', 'completed')
+          WHERE t.appeal_type = '#{appeal_type.name}'
+          AND admin_actions.id IS NULL AND t.type = 'ScheduleHearingTask'
```
mdbenjam: `AND admin_actions.id IS NULL` Why do we want to check that?

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -149,9 +176,10 @@ def valid_states
     @valid_states ||= RegionalOffice::CITIES.values.reject { |ro| ro[:facility_locator_id].nil? }.pluck(:state)
   end
 
-  def create_available_location_by_file_number(file_number, facility:)
+  def create_available_location_for_appeal(appeal, facility:)
     AvailableHearingLocations.create(
-      veteran_file_number: file_number,
+      veteran_file_number: "null", # make migration backwards compatible
```
mdbenjam: You could put the actual veteran_file_number based on the appeal as well. That might solve our cutover job problem.

app/mappers/address_mapper.rb
```diff
@@ -26,6 +26,20 @@ def get_address_from_corres_entry(corres_entry)
     }
   end
 
+  def address_from_veteran_record_or_corres_entry(corres_entry)
```
mdbenjam: I think it'd be a little clearer if you either made an `address` method on the veteran model, or you just added an `address_from_veteran_record` method in this mapper that takes the veteran object. It's a little confusing to me to reference the veteran object directly from within this method.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
lomaxap: the table result without the where statement would be somethings like

| t.appeal_id | t.id | admin_actions.id |
|---|---|---|
| 3 | 1 | 10 |
| 6 | 2 |     |
| 9 | 3 |     |
| 12 | 4 | 9  |
| 12 | 4 | 8 |

we want only appeal ids that don't have admin actions, so we set `admin_actions.id IS NULL`

mdbenjam: Ahhh I see. I understand this now.

mdbenjam: Maybe further comment this SQL query to make it really clear what's going on.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
lomaxap: yep. just wrote that out. definitely simplifies things.

app/jobs/fetch_hearing_locations_for_veterans_job.rb
mdbenjam: So are you suggesting we leave it in here? I feel like that overcomplicates the logic some, since now it lives in two places.

lomaxap: oh, no. I've removed it. was just saying why I added this in the first place.

app/models/tasks/hearing_admin_action_task.rb
```diff
@@ -81,8 +81,7 @@ def self.label
   end
 
   def fetch_closest_ro_and_ahls
-    veteran = appeal.veteran
-    FetchHearingLocationsForVeteransJob.new.perform_once_for(veteran) unless veteran.nil?
+    FetchHearingLocationsForVeteransJob.new.perform_once_for(appeal)
```
mdbenjam: We shouldn't do it in this PR, but I think a better strategy is to refactor the meat of the job out to a service and then we can call that service from the job or here.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -5,10 +5,17 @@ class << self
     def find_or_create_if_eligible(appeal)
       if appeal.is_a?(LegacyAppeal) && appeal.case_record&.bfcurloc == "57" &&
          appeal.hearings.all?(&:disposition)
+
+        parent = HearingTask.active.find_or_create_by!(appeal: appeal) do |task|
```
mdbenjam: oh yikes, didn't realize we were missing this 😮

PR #9507, opened 2019-02-19
Add filtering to other apps

annekainicUSDS:
Addresses https://github.com/department-of-veterans-affairs/caseflow/issues/9536

### Description
There are a few other apps that are using filtering in tables. Now that we've made improvements to filtering functionality for the Queue app, we wanted to use these same common filtering components everywhere.

There are 3 additional apps using filtering (as far as I could tell):
1. Hearing Schedule
2. nonComp
3. Reader

This PR converts the Hearing Schedule and nonComp apps to use the new filtering functionality. Reader was more complex to do, and I decided to leave that for its own PR. It requires making some significant changes to the commonly used filtering components because the assumptions about the structure of the table data that is being filtered are different.

For now, I have created `NewTable` and `NewDropdownFilter` components that will eventually replace `Table` and `DropdownFilter` once Reader has been switched over as well.

I've also removed some historical props and cleaned up some prop names now that they are no longer needed.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
More testing may be needed on the Hearing Schedule table as we wait for more test data for the VLJ column. UPDATE: This has been tested.


client/app/components/NewDropdownFilter.jsx
```diff
@@ -41,10 +41,9 @@ class QueueDropdownFilter extends React.PureComponent {
       <div className="cf-dropdown-filter" style={{ top: '10px' }} ref={(rootElem) => {
         this.rootElem = rootElem;
       }}>
-        {this.props.addClearFiltersRow &&
+        {!this.props.disableClearFilters &&
```
annekainicUSDS: I decided to use this prop to **disable** the clear filters row as opposed to enabling it, so the default option is that the clear filters row is rendered if you don't add this prop to the column.

client/app/hearingSchedule/components/ListSchedule.jsx
```diff
@@ -30,47 +30,14 @@ const formatVljName = (lastName, firstName) => {
   }
 };
 
-const populateFilterDropDowns = (resultSet, filterName) => {
-  let countByFilterName = _.countBy(resultSet, filterName);
-  let uniqueOptions = [];
+// const judgeNameToIdMap = (hearings) => {
```
annekainicUSDS: Waiting to remove this code until we can more thoroughly test the judge name data in the VLJ column (waiting on more test data).

client/app/components/TableFilter.jsx
```diff
@@ -116,11 +118,7 @@ class TableFilter extends React.PureComponent {
       verticalAlign: 'middle'
     }, hover({ cursor: 'pointer' }));
 
-    const filterOptions = tableData && columnName ?
-      this.filterDropdownOptions(tableData, columnName) :
-      // Keeping the historical prop `getFilterValues` for backwards compatibility,
-      // will remove this once all apps are using this new component.
-      getFilterValues;
```
annekainicUSDS: We can now remove this prop because it's no longer used by other apps.

client/app/hearingSchedule/components/ListSchedule.jsx
```diff
@@ -179,57 +124,15 @@ class ListSchedule extends React.Component {
...
-  setLocationSelectedValue = (value) => {
-    this.props.onReceiveHearingSchedule(filterSchedule(this.props.hearingSchedule, 'regionalOffice', value));
-    this.setState({
-      filteredByList: this.state.filteredByList.concat(['Hearing Location'])
-    });
-    this.props.toggleLocationFilterVisibility();
-  };
-
-  /*
-    As props.hearingSchedule does not have judge full name we need to create a full name to judgeId mapping
```
annekainicUSDS: From what I can see of the seed data, and from talking to someone on the hearings schedule team, it is highly unlikely that we would have a value for `judgeId`, but not for `judgeFirstName` or `judgeLastName`.

Given this, I don't think we need this functionality to generate judge names from the `judgeId`. In an ideal world, we should simply be filtering the table based on the data contained in that table and not on a generated value. But this is still a fairly significant change, so if there's any way we can get more certainty that this will not break anything, that would be great!

annekainicUSDS: @lomaxap I would love if you could take a look at this change and make sure that it won't break anything!

client/app/components/NewDropdownFilter.jsx
```diff
@@ -18,7 +18,7 @@ const dropdownFilterViewListItemStyle = css(
   }
 );
 
-class QueueDropdownFilter extends React.PureComponent {
+class NewDropdownFilter extends React.PureComponent {
```
lowellrex: We're naming this component `NewDropdownFilter` to distinguish it from the existing `DropdownFilter`, correct? Do we eventually expect to replace all uses of the existing `DropdownFilter` component (after this PR will it be used anywhere other than Reader?)? If so, can we rename the old `DropdownFilter` to something that implies that it will be going away (`DeprecatedDropdownFilter` or `ReaderDropdownFilter` or something similar) so that we can name this dropdown filter in a way that implies that this component is the default (maybe this component just gets renamed `DropdownFilter`)?

client/app/components/NewTable.jsx
```diff
@@ -190,7 +188,7 @@ class FooterRow extends React.PureComponent {
   }
 }
 
-export default class QueueTable extends React.PureComponent {
+export default class NewTable extends React.PureComponent {
```
lowellrex: Can we name this table something that relates to how this component should be used or something about its nature? How is this component different than the existing `Table` and `TaskTable` components? I'm not very familiar with this component, but at a glance it looks like this table adds pagination (among other things?), so maybe `PaginatedTable` makes sense?

client/app/hearingSchedule/components/ListSchedule.jsx
lomaxap: this looks good to me! 👍

client/app/components/NewTable.jsx
```diff
@@ -264,7 +262,15 @@ export default class QueueTable extends React.PureComponent {
...
         // Only return the data point if it contains the value of the filter
         filteredData = filteredData.filter((row) => {
-          return filteredByList[columnName].includes(_.get(row, columnName));
+          // Sometimes when the cell value is blank, the value is stored as
+          // `undefined`, `null`, or `'null'`. This converts all of these
+          // possible values to 'null' for consistent comparison to the
+          // `filteredByList`, which will always store it as 'null'.
+          let rowValue = typeof _.get(row, columnName) === 'undefined' ? 'null' : _.get(row, columnName);
+
+          rowValue = rowValue === null ? 'null' : rowValue;
```
lowellrex: I find this logic a little hard to follow. Is this doing the same thing as `let rowValue = [undefined, null, 'null'].includes(_.get(row, columnName)) ? 'null' : _.get(row, columnName);`? My javascript knowledge isn't top notch, so maybe there is some reason to not write that as such in javascript?

client/app/components/NewTable.jsx
```diff
@@ -316,18 +322,22 @@ export default class QueueTable extends React.PureComponent {
...
+      rowObjects = rowObjects.length > 0 ? paginatedData[this.state.currentPage] : rowObjects;
+    }
 
     let keyGetter = getKeyForRow;
 
     if (!getKeyForRow) {
       keyGetter = _.identity;
       if (!slowReRendersAreOk) {
-        console.warn('<QueueTable> props: one of `getKeyForRow` or `slowReRendersAreOk` props must be passed. ' +
+        console.warn('<NewTable> props: one of `getKeyForRow` or `slowReRendersAreOk` props must be passed. ' +
```
lowellrex: Why are we logging this to the console? Is this so we notice it when we're testing? Or should Caseflow users do something as a result of this warning?

client/app/components/TableFilter.jsx
```diff
@@ -43,6 +43,8 @@ class TableFilter extends React.PureComponent {
 
     for (let key in countByColumnName) { // eslint-disable-line guard-for-in
       let displayText = `<<blank>> (${countByColumnName[key]})`;
+      const keyValue = key === 'undefined' || key === 'null' ? 'null' : key;
```
lowellrex: It looks like were are doing a similar thing here that [we do in `NewTable`](https://github.com/department-of-veterans-affairs/caseflow/pull/9507/files#r259402858). Is there some reason why we exclude raw `null` and `undefined` values in this case? Also, to prevent drift, it might make sense to bust this logic into a separate helper function we could use across the app.

```jsx
export const consolidatedEmptyValuesFor = (val) => 
  [undefined, null, 'undefined', 'null'].includes(val) ? 'null' : val;
```

client/app/components/NewDropdownFilter.jsx
annekainicUSDS: Good idea! Will do that. And yes, the older version of these components will go away once Reader is switched over. But I will still rename these for now.

client/app/components/NewTable.jsx
annekainicUSDS: Yes, this is very confusing. I tried to describe it in the comment above, but it might not have been clear enough. Basically, what happens is that if there's not a value for a cell, it could be because of 2 reasons: either the value in the data is `undefined` or `null`. It doesn't matter which it is, all that matters is that the cell is blank. When we go to filter the table, we add a value to our `filteredByList` to show what filter has been selected. The filter function then filters the data based on what properties are in the `filteredByList`. The table data could be either `null` or `undefined`, but the `filteredByList` stores both as the string `'null'`, so this check sets either a `null` or `undefined` value in the table data to `'null'` so it can match the `filteredByList`.

What make make this cleaner is creating a mapping of `null` and `undefined` values to `'null'` and set it that way. I'll try that and see if this makes it easier to understand.

client/app/components/NewTable.jsx
annekainicUSDS: I'm not sure! This existed before in the original `<Table/>` component, so I assumed there was a reason for it being there. If not, though, I can remove it!

client/app/components/TableFilter.jsx
annekainicUSDS: Yep, agreed! I can do that.

pkarman:
@lowellrex should we close this? I assume filtering will change anyway under pagination pattern.

Issue #9283, opened 2019-02-07
Show most relevant assignee for caseflow tasks attached to legacy appeals


lowellrex:
We show the VACOLS location code in the "Assigned to" column of queue table views for legacy appeals. This makes sense for legacy appeals that are being worked in VACOLS, but for legacy appeals being processed in Caseflow (by way of Caseflow tasks) we only show "CASEFLOW" in that column. This ticket exists to more precisely show the current assignee for legacy appeals that have tasks in casefow, similar to how the [`Appeal.location_code` function](https://github.com/department-of-veterans-affairs/caseflow/blob/becb7df57bda7cca095cdeba5941bfa1c2eaeac1/app/models/appeal.rb#L104) works right now.

The last few rows with the "Assigned to" value of "CASEFLOW" are an example of this behaviour in action:

![image](https://user-images.githubusercontent.com/32683958/52430656-999ac080-2ad4-11e9-8cee-d26a0205f4ab.png)

[Related slack conversation](https://dsva.slack.com/archives/C6E41RE92/p1549496174101800).

## Acceptance criteria
* [ ] Show most relevant assignee for caseflow tasks attached to legacy appeals in the queue table view
* [ ] Rename the `Appeal.location_code` method to be more descriptive.

lowellrex:
MEDIUM. Don't forget tests.

PR #9523, opened 2019-02-19
Show most relevant assignee for caseflow tasks attached to legacy appeals

joeyyang:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/9283

### Description
- For legacy tasks that have been worked in Caseflow, instead of showing the "Assigned to:" location as "CASEFLOW" on the frontend, show the location it's been assigned in Caseflow
- Change `location_code` in the frontend and anywhere it doesn't refer to `bfcurloc` in VACOLS to `assigned_to_location`


app/models/legacy_appeal.rb
```diff
@@ -715,8 +715,26 @@ def external_id
     vacols_id
   end
 
+  def assigned_to_location
```
joeyyang: still need to add tests for this

app/models/appeal_series.rb
```diff
@@ -174,7 +174,7 @@ def disambiguate_status_advance
   def disambiguate_status_active
     return :scheduled_hearing if latest_appeal.hearing_scheduled?
 
-    case latest_appeal.location_code
+    case latest_appeal.assigned_to_location
```
lowellrex: I think we want to leave this as `location_code` since I think it will [always be associated with a `LegacyAppeal`](https://github.com/department-of-veterans-affairs/caseflow/blob/3c90bbff113bf605d65e2050b17e71d9c3dc21c8/app/models/appeal_series.rb#L2) and we are referencing the VACOLS database field here, not our more human-readable `assigned_to_location` label.

app/models/legacy_appeal.rb
```diff
@@ -715,8 +715,26 @@ def external_id
...
+    return most_recently_assigned_to_label(on_hold_tasks) if on_hold_tasks.any?
+
+    return most_recently_assigned_to_label(tasks) if tasks.any?
+
+    location_code
+  end
+
   private
 
+  def most_recently_assigned_to_label(tasks)
```
lowellrex: Let's rename this argument so that we don't confuse it with the `tasks` relationship the object already has.

app/models/legacy_tasks/legacy_task.rb
```diff
@@ -40,6 +40,10 @@ def assign_to_attorney_data(_user)
     }
   end
 
+  def assigned_to_label
+    assigned_to.is_a?(Organization) ? assigned_to.name : assigned_to.css_id
```
lowellrex: `LegacyTask`s can only be assigned to users so I think we can simplify this into something like `assigned_to&.css_id`

lowellrex:
Looks good! Made a few comments, but once those are addressed (and the tests you mentioned needing to be written are written) this will be ready to rock and roll!

app/models/appeal_series.rb
joeyyang: thank you good catch!

app/models/appeal.rb
```diff
@@ -1,3 +1,8 @@
+##
```
joeyyang: @shanear I'm adding comments as I touch files per this [tech spec](https://github.com/department-of-veterans-affairs/caseflow/issues/9619) — would you mind approving or suggesting changes to the file-level comments I've added below?

app/models/legacy_appeal.rb
```diff
@@ -715,8 +721,28 @@ def external_id
     vacols_id
   end
 
+  def assigned_to_location
+    return location_code unless LOCATION_CODES[:caseflow] == location_code
+
+    active_tasks = tasks.where(status: [Constants.TASK_STATUSES.in_progress, Constants.TASK_STATUSES.assigned])
```
shanear: This should be a private method. It has reusablity value.

app/models/legacy_appeal.rb
```diff
@@ -715,8 +721,28 @@ def external_id
     vacols_id
   end
 
+  def assigned_to_location
+    return location_code unless LOCATION_CODES[:caseflow] == location_code
```
shanear: what does `LOCATION_CODES[:caseflow] == location_code` mean?

Maybe there could be a private method that captures that meaning?

app/models/legacy_appeal.rb
```diff
@@ -715,8 +721,28 @@ def external_id
...
+    # shouldn't happen because if all tasks are closed the task returns to the assigning attorney
+    return most_recently_assigned_to_label(tasks) if tasks.any?
+
+    # shouldn't happen because setting location to "CASEFLOW" only happens when a task is created
+    location_code
+  end
+
   private
 
+  def most_recently_assigned_to_label(tasks_for_appeal)
```
shanear: if `active_tasks` was a private method, then you wouldn't need the argument here. It should always be the tasks for this appeal. It'd be really weird if this private method was called with tasks for another appeal.

app/models/legacy_appeal.rb
```diff
@@ -715,8 +721,28 @@ def external_id
     vacols_id
   end
 
+  def assigned_to_location
+    return location_code unless LOCATION_CODES[:caseflow] == location_code
+
+    active_tasks = tasks.where(status: [Constants.TASK_STATUSES.in_progress, Constants.TASK_STATUSES.assigned])
+    return most_recently_assigned_to_label(active_tasks) if active_tasks.any?
+
+    on_hold_tasks = tasks.where(status: Constants.TASK_STATUSES.on_hold)
```
shanear: again, this could be a private method

app/models/legacy_appeal.rb
shanear: ahh i see, we call this with different sets of tasks for the appeal. I think this would be best as a scope method on task:

```
class Task 
  def self.most_recently_assigned
    order(:updated_at).last
  end
end
```

and then you wouldn't need this method. Just say:

```
on_hold_tasks.most_recently_assigned.assigned_to_location
```

Speaking of which, why is it called `assigned_to_label` on the task and `assigned_to_location` on the appeal. It seems like those should be consistent.

app/models/legacy_appeal.rb
```diff
@@ -715,8 +721,28 @@ def external_id
...
+    return most_recently_assigned_to_label(on_hold_tasks) if on_hold_tasks.any?
+
+    # shouldn't happen because if all tasks are closed the task returns to the assigning attorney
```
shanear: if this shouldn't happen, we should raise an error, not fail gracefully. It means there's bad data and we'd want to know about it.

app/models/legacy_appeal.rb
```diff
@@ -715,8 +721,28 @@ def external_id
...
+    return most_recently_assigned_to_label(tasks) if tasks.any?
+
+    # shouldn't happen because setting location to "CASEFLOW" only happens when a task is created
+    location_code
```
shanear: same as above. If it shouldn't happen, it should error out, so we can know about and resolve the issue, instead of let the bad data linger.

app/models/legacy_appeal.rb
lowellrex: This function is only used to display a label in the queue or case details page, so this situation occurring isn't terrible. I'd be in favor of sending a message to Sentry, but still returning some value here to be super safe.

shanear: Fair. I feel like sending "Error loading location" or something like that is better than a guess that might be wrong or misleading.

lowellrex: I think that makes sense too. I'll defer to @joeyyang about how he'd like to handle it. I think any of those three options makes sense to me (keep it as is, keep the return value as is but send the error to sentry, or use "error loading location" as the return value and send it to sentry).

joeyyang: Let's keep the return value as is and send the error to Sentry, if that works!

app/models/legacy_appeal.rb
joeyyang: My logic was: in the UI, we want to show a label (e.g., a human-readable name versus a number), but the source of truth in the backend is a potentially numerical `location(_code)`, which we want to imply through the name.

joeyyang:
thanks @shanear, back to you! also, can you review the top-of-file comments on `appeal.rb` and `legacy_appeal.rb`?

app/models/appeal.rb
```diff
@@ -1,3 +1,8 @@
...
+# An appeal that a Veteran or appellant for VA decisions on claims for benefits.
```
shanear: I think we could add that it's an appeal to the Board of Veterans' Appeals

app/models/legacy_appeal.rb
```diff
@@ -1,3 +1,9 @@
+##
+# An appeal that a Veteran or appellant for VA decisions on claims for benefits,
+# filed before the Veterans Appeals Improvement and Modernization Act (AMA) went into effect on Feb 19 2019.
```
shanear: I think the timing here is not the most important here. You can still file legacy appeals until Feb 19 2020. The most important is that it's filed according to the laws and policies guiding the process prior to the AMA.

PR #9526, opened 2019-02-19
Production Backfill for HearingTasks

### LINKED ISSUE #9206: ScheduleHearingTask | Backfill production data
There are a few hundred hearings docket AMA appeals in production. We need to backfill these appeals so that they have a HearingTask as a child of the CaseDistributionTask and then a ScheduleHearingTask as a child of the HearingTask.

DistributionTasks will exist in production after this ticket is completed: #9337.

Existing HoldHearingTasks and ScheduleHearingTasks should be updated such that they are children of a HearingTask which is a child of the original task's parent.

oscarAtNava:
Resolves #9206 

### Description
Create HearingTasks for legacy appeals with ScheduleHearingTasks. Make the ScheduleHearingTask and HoldHearingTask, if it exists, children of the newly created HearingTask. Associate HearingTask with the ScheduleHearingTask original parent.

### NB: 
Although ticket mentions tasks for AMA appeals this part was taken care of by Alex's PR (see below). This PR focuses on LegacyAppeals with ScheduleHearingTasks that have RootTask as parent (and HoldHearingTasks if they exist).
A production query shows that there are 696 SHT and 77 HHT with RootTask as parent.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to console.
2. Execute command: bundle exec rails runner scripts/back_fill_hearing_tasks.rb
3. With existing seed data confirm console message 'Total ScheduleHearingTasks to be relinked: 3' displays.
4. Verify Tasks with appeals 5, 6 and 7 were properly moved from RootTask as parent to HearingTask as parent.


scripts/back_fill_hearing_tasks.rb
```diff
@@ -0,0 +1,31 @@
+# Backfill a HearingTask as parent of a ScheduleHearingTask and HoldHearingTask
+# for Legacy Appeals.
+#
+# Author: OAR
+# Date:   Feb 19, 2019
+#
+
+legacy_sched_hear_tasks = Task.where("type='ScheduleHearingTask' AND appeal_type='LegacyAppeal'").select do |task|
+  task.parent.type == "RootTask"
```
sharonwarner: To be safe, let's confirm that there aren't any ScheduleHearingTasks with something other than the RootTask or the HearingTask as its parent.

mdbenjam: Yeah, I'd expect there to be some with a `DistributionTask` as their parent.

scripts/back_fill_hearing_tasks.rb
```diff
@@ -0,0 +1,31 @@
...
+
+legacy_sched_hear_tasks.each do |task|
+  hearing_task = HearingTask.create!(appeal: task.parent.appeal, assigned_to: Bva.singleton)
+
+  # Traverse through all the children for the task parent, which is
+  # the root task for the appeal (see query to identify the ScheduleHearinTasks)
+  root_parent = task.parent
+  hearing_task_status = "completed"
+  root_parent.children.each do |child|
+    next unless child.type == "ScheduleHearingTask" || child.type == "HoldHearingTask"
```
mdbenjam: If there are multiple ScheduleHearingTasks, don't we want each of them to have their own HearingTask parent? I'm not actually sure this case exists right now, if it does it could get pretty complicated because it'll be more difficult to associated ScheduleHearingTasks with their associated HoldHearingTasks.

oscarAtNava: Production data from the beginning of the week did not show this condition. I will run the queries again prior to running the script to make sure.

scripts/back_fill_hearing_tasks.rb
oscarAtNava: I ran queries in production to ascertain all LegacyAppeal related ScheduleHearingTasks either have a HearingTask as a parent or a Root Task. I will run the query again prior to running the script to confirm.

mdbenjam: Cool sounds good.

Issue #9525, opened 2019-02-19
caseflow_decision_date not being set for Appeal decisions


leikkisa:

leikkisa:
This is because the decision document doesn't exist yet at the time of creating the decision issue.

I can't currently tell how long it takes for a DecisionDocument to exist between after the decision issue is created, because the DecisionDocuments are on decision issues that don't have created_at on them.  None of the new Appeal decision issues (with created_at populated) have a Decision Document yet.

We can backfill data for those that do have Decision Documents.  We can also update the decision_date once the DecisionDocument is there, but this would leave the decision_date unpopulated for the gap in between.

We could set the decision date to the day that the decision issue is created, but I would like to confirm that this is correct before doing so, since it's different logic.

leikkisa:
I just checked this out for the latest decision document in prod, and it looks like it's not working :-/. Going to take another look at it.

leikkisa:
It looks like the decision_date is after the DecisionDocument is processed.

leikkisa:
Shane backfilled all of the data just now, we should check it out again tomorrow just to make sure there's not a bug.

leikkisa:
Still does not look like it's working.

leikkisa:
This is working for any new Decision Documents this week. Since this happens in "submit_for_processing" I am wondering if submit_for_processing got called on all of these before the fix, but then they got processed later.... I've noted which Decision Issues, Appeals, and Decision Documents are affected. I am going to backfill them again, and then next time I check, if they are are any new ones I'll be able to see exactly which ones this is happening to (it was hard before due to backfilling everything).

leikkisa:
Checked again today and seems to be good! Moving to closed.

PR #9530, opened 2019-02-19
Update caseflow_decision_date when decision_document is ready

leikkisa:
connects #9525

### DIFF
--- a/app/models/decision_document.rb
+++ b/app/models/decision_document.rb
@@ -33,6 +33,7 @@ class DecisionDocument < ApplicationRecord
   end
 
   def submit_for_processing!
+    appeal.decision_issues.each { |di| di.update!(caseflow_decision_date: decision_date) }
     return no_processing_required! unless upload_enabled?
 
     cache_file!
--- a/spec/models/decision_document_spec.rb
+++ b/spec/models/decision_document_spec.rb
@@ -41,6 +41,7 @@ describe DecisionDocument do
     after { FeatureToggle.disable!(:decision_document_upload) }
 
     let(:expected_path) { "decisions/#{decision_document.appeal.external_id}.pdf" }
+    let!(:decision_issue) { create(:decision_issue, decision_review: appeal, caseflow_decision_date: nil) }
 
     context "when there is a file" do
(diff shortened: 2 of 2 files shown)


app/models/decision_document.rb
```diff
@@ -33,6 +33,7 @@ def pdf_location
   end
 
   def submit_for_processing!
+    appeal.decision_issues.each { |di| di.update!(caseflow_decision_date: decision_date) }
```
leikkisa: I put this outside of process! because not all of the decision_documents get processed.

pkarman: if a job gets restarted, `submit_for_processing!` may be called multiple times. Not sure if that matters here, but maybe a check to `update!(...) unless di.caseflow_decision_date` makes sense?

a nit: putting that one liner in its own method and calling it here would be easier to read.

pkarman: also, should all the `update` calls be wrapped in a single transaction?

leikkisa:
Note - also need to backfill data

leikkisa:
Here's a script we could run:
```
Appeal.select(&:decision_document).each do |appeal|
  appeal.decision_issues.each do |di|
    di.update!(caseflow_decision_date: appeal.decision_document.decision_date) unless di.caseflow_decision_date
  end
end
```

pkarman:
the code logic looks correct, but it will find every Appeal, so it's not as efficient as it could be.

maybe instead something like:

```ruby
DecisionIssue
  .includes(:decision_review)
  .where(decision_review_type: "Appeal")
  .where(caseflow_decision_date: nil).each do |di|
  di.update!(caseflow_decision_date: appeal.decision_document.decision_date)
end
```

leikkisa:
Cool, I will test it on UAT first

leikkisa:
Tested in UAT, worked as expected. Successfully run in prod. The fix deployed to prod this morning so that should cover it.

```
DecisionIssue.includes(:decision_review).where(decision_review_type: "Appeal", caseflow_decision_date: nil).each do |di|
  di.update!(caseflow_decision_date: di.decision_review&.decision_document&.decision_date)
end
```

PR #9532, opened 2019-02-20
Appeal Status: update to handle no decision date 

### LINKED ISSUE #9529: Update to check for decision document in Appeal
Fix for https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3788/

For Appeal:
The decision issues are created before the decision document and the decision date will not be populated on decision issue until the decision document is available.

Update the logic for status and logic around decision date to check for the existence of a decision document.

mc-usds:
Resolves #9529

### Description
An appeal can have decision issues with no decision date before there is a decision document.
The decision issues' decision date is updated when the decision document is available. 
Updating Appeal status to handle this scenario when determining status and what alerts to show.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...


spec/models/appeal_spec.rb
```diff
@@ -1068,11 +1068,22 @@
...
       end
 
-      it "has a decision" do
+      it "status is still in progress since because no decision document" do
         status = appeal.status_hash
-        expect(status[:type]).to eq(:bva_decision)
-        expect(status[:details][:issues].first[:description]).to eq("Dental or oral condition")
-        expect(status[:details][:issues].first[:disposition]).to eq("allowed")
+        expect(status[:type]).to eq(:decision_in_progress)
+        expect(status[:type]).to eq(:decision_in_progress)
```
pkarman: these look like duplicate lines?

app/models/api_status_alerts.rb
```diff
@@ -46,6 +46,7 @@ def post_decision
 
   def post_remand_decision
     return unless decision_review.remand_decision_event_date
+    return unless decision_review.decision_event_date
     return unless Time.zone.today < decision_review.remand_decision_event_date + 365.days
```
pkarman: not in this PR I know but just noticing: should that be `Time.zone.today <= ...` so that on the 1-year date is works?

pkarman: also, what about leap year?

mc-usds: Thanks for catching that, changed to <=

Regarding leap year, I need to figure that out. The github issue called out the deadlines in days from decision date: 365, 120, or 60 depending on the scenario.  And my quick reading through ruby didn't confirm if using 1.year would account for leap year either.  I'll do that in a separate change.

Issue #9519, opened 2019-02-19
Intake | Refresh all AMA EP statuses on start of intake


shanear:
A common use case for cancelling an EP is to free up the issues so they can be re-intook. But because the EP sync script runs so infrequently (once per 1-2 days, per EP), it may take a while for those issues to be properly canceled.

In order to always use the up to date status for all issues when validating, we should sync all EPs for a veteran before the intake begins.

## Acceptance criteria

This one is really hard to validate because it relies a lot on timing.

One thing to try might be to cancel an EP (outside of Caseflow?), and then quickly start another intake. The "quickly" part is to beat the regular cron job that syncs EPs. Then during intake the canceled EP's code should be available to reuse.

* intake anything connected to a rating
* cancel it
* immediately try another intake that would include the same rating

Sjones352:
@pkarman can an Ep be canceled in VBMS?

Sjones352:
I created an Ep and Cancelled in VBMS

**Created in Caseflow**
<img width="872" alt="Screen Shot 2019-03-11 at 1 00 30 AM" src="https://user-images.githubusercontent.com/23080951/54128901-74ef7c80-43e3-11e9-83a0-592499b2077f.png">

**EP Created in VBMS**
<img width="796" alt="Screen Shot 2019-03-11 at 1 00 22 AM" src="https://user-images.githubusercontent.com/23080951/54128900-7456e600-43e3-11e9-9b1e-fe13c781aec6.png">

**Cancel in VBMS**
<img width="663" alt="Screen Shot 2019-03-11 at 12 48 24 AM" src="https://user-images.githubusercontent.com/23080951/54128841-525d6380-43e3-11e9-98d3-2dd631c769b7.png">

<img width="997" alt="Screen Shot 2019-03-11 at 1 01 51 AM" src="https://user-images.githubusercontent.com/23080951/54128860-5be6cb80-43e3-11e9-8d0e-0efb32e02de3.png">

When I create another Ep with the same rating and check VBMS, it never shows up so I believe this ticket passes and beats the  cron job that syncs EPs.

PR #9546, opened 2019-02-20
Refresh EP status

pkarman:
connects #9519 

### Description
Adds hook within the base `Intake` class to run `after_validated_pre_start!` within the `start!` method. For DecisionReview, implement that hook method and sync all active EndProductEstablishment records so that they reflect current status during intake.

### DIFF
--- a/app/models/decision_review_intake.rb
+++ b/app/models/decision_review_intake.rb
@@ -1,6 +1,15 @@
 class DecisionReviewIntake < Intake
   include RunAsyncable
 
+  # run during start!
+  def after_validated_pre_start!
+    veteran.end_products.each do |ep|
+      next unless ep.active? || ep.status_type_code.nil?
+      epe = EndProductEstablishment.find_by(reference_id: ep.claim_id, veteran_file_number: veteran.file_number)
+      epe.sync!
+    end
+  end
+
   def ui_hash(ama_enabled)
     super.merge(
       receipt_date: detail.receipt_date,
--- a/app/models/decision_review_intake.rb
+++ b/app/models/decision_review_intake.rb
@@ -52,7 +52,9 @@ class DecisionReviewIntake < Intake
     veteran.end_products.each do |ep|
       next unless ep.active? || ep.status_type_code.nil?
 
-      EndProductEstablishment.find_by!(reference_id: ep.claim_id, veteran_file_number: veteran.file_number).sync!
+      epe = EndProductEstablishment.find_by!(reference_id: ep.claim_id, veteran_file_number: veteran.file_number)
+      epe.veteran = veteran
+      epe.sync!
     end
   end
 end
--- a/app/models/decision_review_intake.rb
+++ b/app/models/decision_review_intake.rb
@@ -49,10 +49,8 @@ class DecisionReviewIntake < Intake
 
   # run during start!
   def after_validated_pre_start!
-    veteran.end_products.each do |ep|
-      next unless ep.active? || ep.status_type_code.nil?
-
-      epe = EndProductEstablishment.find_by!(reference_id: ep.claim_id, veteran_file_number: veteran.file_number)
+    epes = EndProductEstablishment.active.established.where(veteran_file_number: veteran.file_number)
+    epes.each do |epe|
       epe.veteran = veteran
       epe.sync!
     end
--- a/app/models/decision_review_intake.rb
+++ b/app/models/decision_review_intake.rb
@@ -49,7 +49,7 @@ class DecisionReviewIntake < Intake
 
   # run during start!
   def after_validated_pre_start!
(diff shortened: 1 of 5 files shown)


app/models/decision_review_intake.rb
```diff
@@ -44,4 +44,17 @@ def complete!(request_params)
...
+
+  private
+
+  # run during start!
+  def after_validated_pre_start!
+    veteran.end_products.each do |ep|
+      next unless ep.active? || ep.status_type_code.nil?
+
+      epe = EndProductEstablishment.find_by!(reference_id: ep.claim_id, veteran_file_number: veteran.file_number)
+      epe.veteran = veteran
```
leikkisa: What if there is no end product establishment for the end product (e.g. it wasn't made from an Intake)?  epe will be nil. Also, why do you need to set the EPE's Veteran here, if the epe exists it should already be set.

app/models/decision_review_intake.rb
```diff
@@ -44,4 +44,17 @@ def complete!(request_params)
   def build_issues(request_issues_data)
     request_issues_data.map { |data| RequestIssue.from_intake_data(data, decision_review: detail) }
   end
+
+  private
+
+  # run during start!
+  def after_validated_pre_start!
+    veteran.end_products.each do |ep|
+      next unless ep.active? || ep.status_type_code.nil?
```
leikkisa: I don't think the status_type_code would be nil for an End Product on a Veteran since those come from BGS (unlike an End Product Establishment), because it gets a status automatically once its created.

app/models/decision_review_intake.rb
```diff
@@ -44,4 +44,17 @@ def complete!(request_params)
   def build_issues(request_issues_data)
     request_issues_data.map { |data| RequestIssue.from_intake_data(data, decision_review: detail) }
   end
+
+  private
+
+  # run during start!
+  def after_validated_pre_start!
+    veteran.end_products.each do |ep|
```
leikkisa: Should this be looking at the Veteran's End Product Establishments?

spec/models/higher_level_review_intake_spec.rb
```diff
@@ -21,6 +21,44 @@
...
+      )
+    end
+
+    let!(:existing_canceled_epe) do
+      create(
+        :end_product_establishment,
+        :canceled,
+        veteran_file_number: veteran_file_number
+      )
+    end
```
leikkisa: I think this comment will be moot if you switch to looking at EPEs, but if not, then I would add an End Product on the Veteran that doesn't have an EPE for the tests.

pkarman:
Thanks for the feedback, @leikkisa 

I have refactored to act on EPE instead of EP. I had forgotten there could be End Products with an EPE.

The reason this PR passes around the same `veteran` object is that each EPE will instantiate a new one, and potentially backfill the veteran's name, and then re-fetch all the EPs for that veteran. So it's a performance savings since this method will be done in real-time during the intake flow, rather than in a background job.

app/models/decision_review_intake.rb
```diff
@@ -44,4 +44,15 @@ def complete!(request_params)
   def build_issues(request_issues_data)
     request_issues_data.map { |data| RequestIssue.from_intake_data(data, decision_review: detail) }
   end
+
+  private
+
+  # run during start!
+  def after_validated_pre_start!
+    epes = EndProductEstablishment.active.established.where(veteran_file_number: veteran.file_number)
```
leikkisa: Active checks for established right now (sync also checks that it's active), so might be a bit of redundancy here. I don't know if that hurts anything though.

This might be for a different PR, but I think we don't want active to check for established (because if it's still in the process of establishing we want it to be considered active).  I think that's why we added the check for synced_status of null to the active method.  Although it'd be weird to sync EPs that aren't established yet, so I guess we want the sync to still check that it's both active and established.

Don't think it hurts to be on DecisionReview, but wondering if it's more clear to be on ClaimReview.

spec/models/higher_level_review_intake_spec.rb
```diff
@@ -21,6 +21,41 @@
...
+        :canceled,
+        veteran_file_number: veteran_file_number,
+        established_at: Time.zone.yesterday
+      )
+    end
+
+    before do
+      @synced = []
+      allow_any_instance_of(EndProductEstablishment).to receive(:sync!) do |epe|
+        @synced << epe.id
```
leikkisa: neat test!

app/models/decision_review_intake.rb
pkarman: I removed the `.active` scope since `sync!` will already check for that.

I think Appeals should get the same treatment, in case it affects any of the issues that are being attempted. That's why I put it on DecisionReview. I'm not sure if there's a real use case there, but I don't think it hurts.

PR #9547, opened 2019-02-20
Untimely evidence

### LINKED ISSUE #8340: Need to capture when untimely evidence was received in attorney check-out
Section 5 sub (t) of AMA requires that we capture data on how frequently a Veteran submits evidence at a time when it can't be considered. This includes:

1) submitting evidence for an appeal on the direct review lane;
2) submitting evidence outside the new evidence 90-day window on the evidence lane;
3) submitting evidence outside the post-hearing new evidence 90-day window on the hearing lane; and possibly, 
4) submitting evidence after having waived the post-hearing new evidence window on the hearing lane.

LRP suggested a check-box in the attorney check-out flow, but will leave it in your capable hands to figure out the best solution here. 

## Mock

image

image

youngfreezyVA:
Resolves #8340

### Description
in the submit view, add an option to allow attorneys to indicate that untimely evidence was submitted. per this mock:
image

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
Try checking the untimely evidence checkbox on submit, submit to a judge
go to that judge, return to an attorney
go to that attorney, when in the submit view again, untimely evidence should still be checked
for ama attorney checkout only

screenshots of changes:
image
image
image

arrow icon doesn't exist yet, thus the reason it isn't there atm.


app/mappers/queue_mapper.rb
```diff
@@ -174,6 +174,10 @@ def overtime
     decass_attrs[:overtime]
   end
 
+  def untimely_evidence
```
aroltsch: You don't need it here. queue_mapper is used for legacy cases only.

db/migrate/20180205183203_create_attorney_case_reviews.rb
```diff
@@ -6,6 +6,7 @@ def change
       t.integer "attorney_id"
       t.string  "work_product"
       t.boolean  "overtime", default: false
+      t.boolean "untimely_evidence", default: false
```
aroltsch: In order to add a new column to the table, you need to generate a new migration file and then run db:migrate to update schema.rb file. We don't update schema.rb directly. Here is a tutorial: https://www.tutorialspoint.com/ruby-on-rails/rails-migrations.htm

youngfreezyVA: ah I see so this is the actual generated migration file

PR #9557, opened 2019-02-20
Add columns for two request_issues column changes

shanear:
connects #8239


db/schema.rb
```diff
@@ -245,6 +245,7 @@
   create_table "decision_issues", force: :cascade do |t|
     t.string "benefit_type"
     t.date "caseflow_decision_date"
+    t.datetime "created_at"
```
shanear: somehow this was missing from a previous PR 😬

Issue #9142, opened 2019-02-01
DTA Error EP needs to retain the same POA as the original EP


shanear:
When we are creating 040s from Higher Level Reviews, and that HLR has a claim level (limited) POA, then the follow up POA should have the same POA

### Validation
- Create a Higher Level Review decision review, add a rating and nonrating issue to it.
- Have someone add a limited POA to both end products on the claim.  Unfortunately, I'm not sure who does this or how this happens, maybe it can get added in VBMS or maybe Colin can help.
- Have the decisions on both issues be a DTA error (need Colin's help)
- Check that the new Supplemental Claim that gets created creates two dta end products that both have the same limited POA on them.  EndProductEstablishment will have the fields limited_poa_code and limited_poa_access populated, and VBMS should show the limited POA on the claims as well.

leikkisa:
I created a couple of claims with a limited POA (600149269 and 600149268).  I looked at the raw end_product data from BGS and the claim data from VBMS and did not find it in either.  I am asking VBMS if they can point me to where to find it, and if they might know when it would be set besides when we establish a claim.

It appears that there is a field called powerOfAttorneyId under "Claim" for VBMS's ClaimService V5 in a document that I have, and it is surrounded by all of the other data fields that gets returned with the claim, but alas it was not in the response that I got.  My theory right now is that VBMS might still need to add this.

Other note: establishing the claim with the limited POA failed when I set the limited_poa_access to false (it gave a payee_code error which was a little misleading)

leikkisa:
We can get it from OrgWebService.findLimitedPoasByBnftClaimIds(<list>claimId)

leikkisa:
Validated that the VBMS service for setting limited POA is working in VBMS! BGS service does not seem to be working though, have sent them an email.  The call is succeeding, but not returning anything. I also tried another call with the same service, looking up POA by participant_id because I noticed that VBMS had one, and it also didn't return anything. So I think the whole service might not be working.

Have started working on the DTA SC part, but there's a complicated bug in creating the DTA SCs, and generally think there might need to be some refactoring there, so am going to switch gears.

leikkisa:
Update - BGS service is working, just needed to tweak the code which I was able to do with a combo of some similar documentation from Melissa and some examples BGS sent.  Next step is to code it, but first am fixing a bug we discovered in the process.

nicholasholtz:
@leikkisa Has this one been validated?

pkarman:
[...]

PR #9560, opened 2019-02-20
Add BGS service for limited POA to Caseflow

leikkisa:
connects #9142


app/mappers/power_of_attorney_mapper.rb
```diff
@@ -28,6 +28,21 @@ def get_hash_of_poa_from_bgs_poas(bgs_resp)
...
   end
 
+  def get_limited_poas_hash_from_bgs(bgs_response)
+    return unless bgs_response
+
+    limited_poas_hash = {}
+
+    Array.wrap(bgs_response).map do |lpoa|
+      limited_poas_hash[lpoa[:bnft_claim_id]] = {
+        limited_poa_code: lpoa[:poa_cd],
```
leikkisa: The new keys are what the fields are called in the VBMS establish claim call.

app/mappers/power_of_attorney_mapper.rb
```diff
@@ -28,6 +28,21 @@ def get_hash_of_poa_from_bgs_poas(bgs_resp)
...
+        limited_poa_access: lpoa[:authzn_poa_access_ind]
+      }
+    end
+
+    limited_poas_hash
```
leikkisa: We can submit one claim id or multiple claim IDs in an array for this call. I thought it might be useful to submit multiple since a decision review can have multiple end products, so converted the result from an array to a hash keyed by the claim_id (but we could keep it as an array if people think that's unnecessary).

app/mappers/power_of_attorney_mapper.rb
jiahuang: just double checking, will the bgs fields always exist?

lib/fakes/bgs_service.rb
```diff
@@ -555,6 +555,17 @@ def fetch_poas_by_participant_ids(participant_ids)
...
 
+  def fetch_limited_poas_by_claim_ids(claim_ids)
+    result = {}
+    Array.wrap(claim_ids).each do |claim_id|
+      if claim_id.include? "HAS_LIMITED_POA"
+        result[claim_id] = { limited_poa_code: "OU3", limited_poa_access: "Y" }
+      end
+    end
+
+    result unless result.empty?
```
pkarman: does this return `nil` if `result.empty?` is true?

If so I would find it more clear if it was written as:
```ruby
result.empty? ? nil : result
```

app/mappers/power_of_attorney_mapper.rb
leikkisa: 🤷‍♀️ 
authzn_change_clmant_addrs_ind isn't always there, but we're not using it at the moment.

The benefit claim id will always be there. I'm fairly confident the poa code and access indicator will be there.  If they're not, then they'll be set to nil, which might get weird when we're setting the limited POA on a new claim. Although if they're not there, then I guess that would indicate they're not required.

lib/fakes/bgs_service.rb
leikkisa: Yes it does, but that makes sense! Will update.

Issue #9559, opened 2019-02-20
Intake | Review and edit schema comment spreadsheet


pkarman:
Spreadsheet: https://docs.google.com/spreadsheets/d/1AqJ6peHD2UVOiZnQdgRYHXe1NK0IiMUT97xcM4XLJoE/edit?ts=5c6b42b3#gid=482501305

When you've added your input, check off your name.

* [ ] @leikkisa 
* [ ] @shanear 
* [ ] @carolaponce 
* [x] @pkarman

leikkisa:
I think we'll still have edits to make in the future, but I'm going to go ahead and close this issue!

PR #9562, opened 2019-02-20
Example comment migration for legacy issue optin

leikkisa:
connects #9559


leikkisa:
Note - here's some example code to add comments when creating a table/column in the first place:
```
def change
    create_table :my_table, comment: 'A description of my_table' do |t|
      t.string :a_field, comment: 'A description of a_field'
      t.string :another_field, comment: 'A description of another_field'
    end

    add_index :my_table, :a_field,
              name: 'index_my_table_on_a_field',
              unique: true,
              comment: 'A comment for the index'
  end
```

leikkisa:
Also FYI, here's a query for getting all of the columns with their comments. More fields about the data is available but I figured this covers what's commonly found in a data dictionary.

```
-- fetch all columns with comments for data dictionary
select c.table_name, c.column_name, pgd.description, c.is_nullable, c.data_type
from pg_catalog.pg_statio_all_tables as st
inner join information_schema.columns c
on c.table_schema = st.schemaname
and c.table_name = st.relname
left join pg_catalog.pg_description pgd
on pgd.objoid=st.relid
and pgd.objsubid=c.ordinal_position
where st.schemaname = 'public'
order by 1, 2
;
```

PR #9564, opened 2019-02-20
adding nil check for legacy_hearing

mc-usds:
### Description
Updating to handle when venue is nil. Addresses:
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3795/

Legacy hearings scheduled via Caseflow will have the hearing information stored in hearing_location

### Acceptance Criteria
- [x] Code compiles correctly


mc-usds:
@cmgiven 
I'm not familiar too familiar with the area of the code. Somehow venue is null, my guess is because the venue_key is not found in the hash? I'm adding a nil check before tying to get the type location so it doesn't error out, but then in that case, that value will be nil.  I tried doing a search for which LegacyHearings have venue == nil, but ran into an error:

 %> LegacyHearing.select { |lh| lh.venue.nil? }

Traceback (most recent call last):
        7: from (irb):2
        6: from (irb):2:in `block in irb_binding'
        5: from app/models/legacy_hearing.rb:39:in `venue'
        4: from app/models/concerns/associated_vacols_model.rb:18:in `block (2 levels) in vacols_attr_accessor'
        3: from app/models/concerns/associated_vacols_model.rb:67:in `check_and_load_vacols_data!'
        2: from app/models/concerns/associated_vacols_model.rb:89:in `perform_vacols_request'
        1: from app/repositories/hearing_repository.rb:192:in `load_vacols_data'
Caseflow::Error::VacolsRecordNotFound (Hearing record with vacols id 628944 not found.)

PR #9567, opened 2019-02-21
Allow generic document uploads to VBMS via IDT

### LINKED ISSUE #8641: [BE] Implement BE endpoint to allow IDT to upload any document to VBMS
AC: 

1. Implement a BE endpoint to allow IDT to upload any document to VBMS. Utilize existing functionality that was already implemented for outcoding the appeal. 
https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/idt/api/v1/appeals_controller.rb#L40

2. Work with Paul Saindon to test the endpoint in UAT.

Tech notes: VBMS requires certain fields when uploading a document, make these fields required params.

monfresh:
**Why**: Paul Saindon has requested the ability to upload any type of
document to VBMS via an IDT endpoint.

**How**:
- Create a new `vbms_uploaded_documents` DB table with various
timestamps used by the `Asyncable` module to track the upload process,
similar to the `decision_documents` table (which might be refactored
later to use this new table).
- Create a new endpoint for uploading documents at:
`idt/api/v1/appeals/:appeal_id/upload_document`. This endpoint requires
the following parameters: `appeal_id` (the appeal's UUID), `file`: a
Base64-encoded file, and `document_type`, a string description of the
document.
- The controller calls a Service Object that validates that the appeal
exists, that a file is present, and that the `document_type` is
registered in the Caseflow Commons Document Types:
https://github.com/department-of-veterans-affairs/caseflow-commons/blob/master/app/models/caseflow/document_types.rb
If everything is valid, the file is uploaded to VBMS via a background
job.

Resolves #8641

Test Plan:
Work with Paul to test doc uploads to VBMS in UAT

### DIFF
--- a/app/controllers/idt/api/v1/upload_vbms_document_controller.rb
+++ b/app/controllers/idt/api/v1/upload_vbms_document_controller.rb
@@ -1,0 +1,24 @@
+class Idt::Api::V1::UploadVbmsDocumentController < Idt::Api::V1::BaseController
+  protect_from_forgery with: :exception
+  skip_before_action :verify_authenticity_token, only: [:create]
+  before_action :verify_access
+
+  rescue_from StandardError do |e|
+    Raven.capture_exception(e)
+    if e.class.method_defined?(:serialize_response)
+      render(e.serialize_response)
+    else
+      render json: { message: "Unexpected error: #{e.message}" }, status: :internal_server_error
+    end
+  end
+
+  def create
+    result = PrepareDocumentUploadToVbms.new(request.parameters).call
+
+    if result.success?
+      render json: { message: "Document successfully queued for upload." }
+    else
+      render json: result.errors[0], status: :bad_request
+    end
+  end
+end
--- a/app/jobs/upload_document_to_vbms_job.rb
+++ b/app/jobs/upload_document_to_vbms_job.rb
@@ -1,0 +1,11 @@
+class UploadDocumentToVbmsJob < CaseflowJob
+  queue_as :low_priority
+  application_attr :idt
+
+  def perform(document)
+    RequestStore.store[:application] = "idt"
+    RequestStore.store[:current_user] = User.system_user
+
+    UploadDocumentToVbms.new(document).call
+  end
+end
--- a/app/models/vbms_uploaded_document.rb
+++ b/app/models/vbms_uploaded_document.rb
@@ -1,0 +1,6 @@
+class VbmsUploadedDocument < ApplicationRecord
+  belongs_to :appeal, optional: false
+  validates :document_type, presence: true
+
+  attr_accessor :file
+end
--- a/app/controllers/idt/api/v1/appeals_controller.rb
+++ b/app/controllers/idt/api/v1/appeals_controller.rb
@@ -17,10 +17,6 @@ class Idt::Api::V1::AppealsController < Idt::Api::V1::BaseController
     render(json: { message: e.message }, status: :bad_request)
   end
 
-  rescue_from Caseflow::Error::DocumentUploadFailedInVBMS do |e|
-    render(e.serialize_response)
-  end
-
   rescue_from ActiveRecord::RecordNotFound do |_e|
     render(json: { message: "Record not found" }, status: :not_found)
   end
--- a/app/models/appeal.rb
+++ b/app/models/appeal.rb
@@ -10,6 +10,7 @@ class Appeal < DecisionReview
 
   # decision_documents is effectively a has_one until post decisional motions are supported
   has_many :decision_documents
+  has_many :vbms_uploaded_documents
   has_many :remand_supplemental_claims, as: :decision_review_remanded, class_name: "SupplementalClaim"
 
   has_one :special_issue_list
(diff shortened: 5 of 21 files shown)


app/jobs/upload_document_to_vbms_job.rb
```diff
@@ -0,0 +1,11 @@
+class UploadDocumentToVbmsJob < CaseflowJob
+  queue_as :low_priority
+  application_attr :intake
```
monfresh: I'm not sure if we need this. I copied `ProcessDecisionDocumentJob`.

app/models/vbms_uploaded_document.rb
```diff
@@ -0,0 +1,78 @@
+class VbmsUploadedDocument < ApplicationRecord
+  include Asyncable
+
+  class NotYetSubmitted < StandardError; end
+
+  belongs_to :appeal, optional: false
+  validates :document_type, presence: true
+
+  S3_SUB_BUCKET = "vbms_uploaded_documents".freeze
```
monfresh: Does this sub bucket have to exist already in S3?

monfresh:
Some questions that came to mind:

- How big do we expect this table to get?
- Once a file is uploaded successfully, do we want to keep the corresponding row in the table around?
- Should we have a rake task to periodically remove rows for successful uploads?

app/controllers/idt/api/v1/upload_vbms_document_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+    Raven.capture_exception(e)
+    if e.class.method_defined?(:serialize_response)
+      render(e.serialize_response)
+    else
+      render json: { message: "Unexpected error: #{e.message}" }, status: :internal_server_error
+    end
+  end
+
+  def create
+    result = QueueDocumentUploadToVbms.new(request.parameters).call
```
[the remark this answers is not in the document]
monfresh: What I meant by "queue" was prepare it to be uploaded. This class validates the params, then calls the background job which then calls the `UploadDocumentToVbms` class that does the actual uploading. Perhaps a better name would be `ValidateDocumentUploadToVbms`?

app/models/vbms_uploaded_document.rb
aroltsch: I don't think so.

aroltsch: I think maybe we should use a document type as a name for s3 sub_bucket because `vbms_uploaded_documents` is very generic and we already upload other documents to VBMS

app/controllers/idt/api/v1/upload_vbms_document_controller.rb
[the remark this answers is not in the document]
aroltsch: Well the class does more that just validating. `PrepareDocumentUploadToVbms`? My concern is that there are two workflow classes with a very similar name. `QueueDocumentUploadToVbms` and `DocumentUploadToVbms` - it is unclear what the difference is by simply looking at the name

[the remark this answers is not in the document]
monfresh: I like "prepare". I just pushed a change with that name and updated the job `application_attr` as well.

app/models/vbms_uploaded_document.rb
monfresh: Great idea. The problem is that we won't be able to have the bucket name match the `document_type` exactly because S3 bucket names must follow all of these rules:

- Bucket names must be unique across all existing bucket names in Amazon S3.

- Bucket names must comply with DNS naming conventions.

- Bucket names must be at least 3 and no more than 63 characters long.

- Bucket names must not contain uppercase characters or underscores.

- Bucket names must start with a lowercase letter or number.

- Bucket names must be a series of one or more labels. Adjacent labels are separated by a single period (.). Bucket names can contain lowercase letters, numbers, and hyphens. Each label must start and end with a lowercase letter or a number.

- Bucket names must not be formatted as an IP address (for example, 192.168.5.4).

- When you use virtual hosted–style buckets with Secure Sockets Layer (SSL), the SSL wildcard certificate only matches buckets that don't contain periods. To work around this, use HTTP or write your own certificate verification logic. We recommend that you do not use periods (".") in bucket names when using virtual hosted–style buckets.

This means we would have to do string manipulation to convert something like `Notification Letter (e.g. VA 20-8993, VA 21-0290, PCGL)` into a format that S3 accepts.

Perhaps the best solution would be to name the bucket something like `vbms-doctype-#{document_type_id}`, where `document_type_id` corresponds to the number key in this TYPES hash: https://github.com/department-of-veterans-affairs/caseflow-commons/blob/master/app/models/caseflow/document_types.rb

aroltsch: Ah, makes sense! Let's just name the bucket: `idt_uploaded_documents` then? I don't want to use a `document_type_id` because it would be unclear to a person looking at it what this number stands for.

monfresh: Done, except I used dashes instead of underscore because AWS doesn't allow underscores, which makes me wonder how the `form_8` and `hearing_schedule` buckets are working.

PR #9568, opened 2019-02-21
Add bummr and derailed_benchmarks dev gems

monfresh:
**Why**: bummr is very useful for updating multiple gems at once using
separate commits. derailed_benchmarks is great for finding memory and
other performance issues.

The reason for pointing at the master branch for bummr is that there is
a bug in version 0.4.0 and the fix that is on master hasn't been
released yet. See: lpender/bummr#58

References:
https://github.com/lpender/bummr
https://github.com/schneems/derailed_benchmarks

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -105,6 +105,8 @@ group :test, :development, :demo do
 end
 
 group :development do
+  gem "bummr", "= 0.3.2", require: false
+  gem "derailed_benchmarks"
   gem "dotenv-rails"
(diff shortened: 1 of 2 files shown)


pkarman:
I'm +1 to the use of these gems.

Could we wait a few days and see if a release fix goes out? Pointing at github master makes me a little uneasy since it seems possible to accidentally update.

monfresh:
I locked it to version 0.3.2 for now.

PR #9574, opened 2019-02-21
Schedule-HearingPrep/appellant refactor

### LINKED ISSUE #7711: Hearing Prep/Schedule | Claimant vs Appellant vs Veteran
In hearing prep and hearing schedule we have logic all over the place to display appellant if it exists, otherwise display veteran. There's a method `claimant` that already built this logic in. We should use that instead.

Sjones352:
Resolves #7711

### Description
Changes appellant to claimant method

### Acceptance Criteria
- [x] In hearing prep and hearing schedule we have logic all over the place to display appellant if it exists, otherwise display veteran. There's a method `claimant` that already built this logic in. We should use that instead.

### Testing Plan
1. Go to DailyDocket in Schedule


Sjones352:
@sharonwarner I think this method I am implementing is redundant and maybe leaving appellant as it is, is better.

app/models/legacy_appeal.rb
```diff
@@ -336,18 +336,18 @@ def contested_claim
   def claimant
     if appellant_is_not_veteran
       {
-        first_name: appellant_first_name,
-        middle_name: appellant_middle_initial,
-        last_name: appellant_last_name,
+        appellant_first_name: appellant_first_name,
```
sharonwarner: We want to keep this as first_name. That way on the frontend, we can just say claimant.first_name and we don't need to check for appellant/veteran anymore.

PR #9576, opened 2019-02-21
Seeding Lots of Veterans Ready for Hearing

lomaxap:


db/seeds.rb
```diff
@@ -875,7 +935,10 @@ def create_previously_held_hearing_data
     return if ([appeal.type] - ["Post Remand", "Original"]).empty? &&
               appeal.hearings.map(&:disposition).include?(:held)
 
-    FactoryBot.create(:case_hearing, :disposition_held, user: user, folder_nr: appeal.vacols_id)
+    master_record = FactoryBot.create(:case_hearing, hearing_date: Time.zone.local(2018, 12, 20))
```
sharonwarner: We shouldn't create any parent records in Caseflow. Let's change the hearing date to something after 4/1 and create a Caseflow hearing day.

db/seeds.rb
```diff
@@ -374,25 +374,82 @@ def create_hearing_days
...
+    ) do |task|
+      task.update(
+        parent: hearing_task,
+        assigned_to: HearingsManagement.singleton
+      )
+    end
+  end
+
   def create_legacy_case_with_open_schedule_hearing_task(ro_key)
+    fname = Faker::Name.first_name
```
sharonwarner: We should use the FACOLS logic to create these cases. Documentation for that is here: https://github.com/department-of-veterans-affairs/caseflow/blob/master/docs/FACOLS.md

lomaxap: @mdbenjam I'm trying to create lots of FACOLS records for appeals ready for hearing to test performance on the assign hearings page a little better. Right now, I'm creating 100 / regional office manually. We're wondering if there's a better way to seed an arbitrary number of FACOLS records using the method @sharonwarner linked to.

mdbenjam: Yeah I think there is. We can connect to the DB defined in the documentation and pull specific rows down that meet the requirements we need.

We can pair on it this afternoon if you'd like.

Issue #9571, opened 2019-02-21
Appeal Outcoding | Wait until the decision date to process the outcoding


shanear:
We are seeing errors come in when we process some outcodings.

See:

https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3803/events/250788/

The solution here is to not attempt to process the decision document until the decision date. Confirmed with Lauren and Allyce.

## Acceptance criteria

Not sure if there's a way in UAT to manually set the decision date in the future, but if it is, then the way to test this would be to verify that the decision document record has a submitted_at date value that equals the decision date.

Will need IDT help from Anya

Sjones352:
@pkarman validating this in UAT.
I set the decision date for a future date 
<img width="1119" alt="screen shot 2019-03-08 at 12 26 34 pm" src="https://user-images.githubusercontent.com/23080951/54044944-3ddc5980-419e-11e9-890a-9369abdcea9d.png">
 
**Created Ep**
<img width="800" alt="screen shot 2019-03-08 at 12 32 39 pm" src="https://user-images.githubusercontent.com/23080951/54044979-577da100-419e-11e9-8e40-81f2bdb92097.png">

**Ep Status**
<img width="1642" alt="screen shot 2019-03-08 at 12 33 55 pm" src="https://user-images.githubusercontent.com/23080951/54045042-87c53f80-419e-11e9-8483-77a3d4462c6e.png">

Wondering how I can validate this part `verify that the decision document record has a submitted_at date value that equals the decision date`

pkarman:
This step:
"Will need IDT help from Anya"
is the trick. We need @aroltsch to help us step the decision process through its paces so we end up with a DecisionDocument record. I believe she has done that before for Steve so you might just ask her.

aroltsch:
From what I understand, I need to outcode the appeal with the decision date in the future.

aroltsch:
I outcoded the appeal in UAT and set the decision date to 1.day.from_now. Here is what I did in rails console:

```
uuid = "ed984057-cc2b-473b-a6cf-76b1145667d2".upcase
a = Appeal.find_by(uuid: uuid)
a.tasks.where(type: "BvaDispatchTask").delete_all
RequestStore[:current_user] = User.last
a.request_issues.each do |r|
[...]

PR #9577, opened 2019-02-21
Delay outcoding to decision date

pkarman:
connects #9571 

## Description

Calculates the appropriate delay for decision documents with decision dates in the future.

### DIFF
--- a/app/models/tasks/bva_dispatch_task.rb
+++ b/app/models/tasks/bva_dispatch_task.rb
@@ -30,9 +30,8 @@ class BvaDispatchTask < GenericTask
     def create_decision_document!(params)
       DecisionDocument.create!(params).tap do |decision_document|
         delay = 0
-        decision_time = decision_document.decision_date.to_time(:local)
-        if decision_time > Time.zone.now
-          delay = (decision_time - Time.zone.now).ceil
+        if decision_document.decision_date > Time.zone.today
+          delay = decision_document.decision_date.to_time(:local)
         end
         decision_document.submit_for_processing!(delay: delay)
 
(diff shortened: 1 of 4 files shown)


app/models/tasks/bva_dispatch_task.rb
```diff
@@ -29,7 +29,12 @@ def outcode(appeal, params, user)
 
     def create_decision_document!(params)
       DecisionDocument.create!(params).tap do |decision_document|
-        decision_document.submit_for_processing!
+        delay = 0
+        decision_time = decision_document.decision_date.to_time(:local)
+        if decision_time > Time.zone.now
+          delay = (decision_time - Time.zone.now).ceil
+        end
+        decision_document.submit_for_processing!(delay: delay)
```
shanear: this seems overly complex. What if we added another optional arg on `submit_for_processing` like this:

```
submit_for_processing(delay_until: decision_date)
```

shanear:
```
def submit_for_processing!(delay: 0, delay_until: nil)
    when_to_start = delay_until.to_datetime || (Time.zone.now + delay)

    update!(
      self.class.last_submitted_at_column => when_to_start,
      self.class.submitted_at_column => when_to_start,
      self.class.processed_at_column => nil
    )
  end
```

pkarman: I like passing in a Time object. How about 7a995e583 ?

shanear: I like that, but I'm wondering if we want a `DateTime` instead since that's the more common `Time` object, or to allow any type of time object

shanear: Also this method looks nice here 
https://apidock.com/rails/ActiveSupport/TimeWithZone/future%3F

shanear:
```
decision_document.submit_for_processing!(
  delay: decision_document.decision_date.future? ? decision_document.decision_date : 0
)
```

😄

PR #9586, opened 2019-02-21
Fix for when there are no ready direct reviews

cmgiven:
Resolves an error that judges are reporting.


monfresh:
Can we add a spec please?

cmgiven:
Hotfix. Blocking distributions now.

cmgiven:
Thanks, all!

PR #9599, opened 2019-02-21
Add limited POA to DTA End Products

leikkisa:
connects #9142

### DIFF
--- a/app/models/end_product.rb
+++ b/app/models/end_product.rb
@@ -194,7 +194,7 @@ class EndProduct
   end
 
   def limited_power_of_attorney
-    @limited_power_of_attorney ||= claim_id ? BGSService.new.fetch_limited_poas_by_claim_ids(claim_id)
+    @limited_power_of_attorney ||= fetch_limited_power_of_attorney
   end
 
   def ramp?
@@ -205,6 +205,13 @@ class EndProduct
 
   def label
     @label ||= CODES[claim_type_code]
+  end
+
+  def fetch_limited_power_of_attorney
+    return unless claim_id
+
+    limited_poa = BGSService.new.fetch_limited_poas_by_claim_ids(claim_id)
+    limited_poa ? limited_poa[claim_id] : nil
   end
 
   def near_decision_date_of?(appeal)
--- a/app/models/end_product.rb
+++ b/app/models/end_product.rb
@@ -193,6 +193,10 @@ class EndProduct
     @contentions ||= claim_id ? VBMSService.fetch_contentions(claim_id: claim_id) : nil
   end
 
+  def limited_power_of_attorney
+    @limited_power_of_attorney ||= claim_id ? BGSService.new.fetch_limited_poas_by_claim_ids(claim_id)
+  end
+
   def ramp?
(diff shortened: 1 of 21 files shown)


app/models/higher_level_review.rb
```diff
@@ -19,7 +19,10 @@ def ui_hash
     )
   end
 
-  def on_decision_issues_sync_processed(_end_product_establishment)
+  def on_decision_issues_sync_processed(end_product_establishment)
+    return unless decision_issues.remanded.uncontested
+
+    end_product_establishment.check_for_limited_power_of_attorney!
```
leikkisa: I want to minimize the calls to BGS for this. This checks all of a Higher Level Review's end products if there are any remanded issues.  Theoretically we could put this down at the issue level so it doesn't check all of the end products, only the ones with a remand, but it gets weird, so I am putting here for now.

app/models/end_product.rb
```diff
@@ -102,6 +102,14 @@ def claimant_participant_id
     @claimant_participant_id ||= nil
   end
 
+  def limited_poa_code
+    @limited_poa_code ||= nil
```
leikkisa: These are for establishing a new end product. I don't have these set to fetch data for an existing end product, because we don't need that information unless creating a remand claim. I hope this isn't misleading though.

app/models/end_product.rb
```diff
@@ -183,6 +193,10 @@ def contentions
     @contentions ||= claim_id ? VBMSService.fetch_contentions(claim_id: claim_id) : nil
   end
 
+  def limited_power_of_attorney
+    @limited_power_of_attorney ||= claim_id ? BGSService.new.fetch_limited_poas_by_claim_ids(claim_id)
```
leikkisa: This is enough to prevent fetching it multiple times right?

leikkisa:
This assumes that a decision_issues previous request issues are all on the same end product. I don't know if this is correct given the comment at https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/decision_issue.rb#L75.

However, when we asked if a rating issue would ever become a nonrating issue, Nicholas felt pretty confident that would never happen, so I went with this for now.

app/models/end_product.rb
```diff
@@ -183,6 +193,10 @@ def contentions
     @contentions ||= claim_id ? VBMSService.fetch_contentions(claim_id: claim_id) : nil
   end
 
+  def limited_power_of_attorney
```
jiahuang: would prefer that either all `limited_poa` gets switched to `limited_power_of_attorney` or this gets switched to `limited_poa` just to be consistent

db/schema.rb
```diff
@@ -354,6 +354,8 @@
     t.string "doc_reference_id"
     t.datetime "established_at"
     t.datetime "last_synced_at"
+    t.boolean "limited_poa_access", comment: "Indicates whether the limited Power of Attorney has access to view documents"
```
jiahuang: 👏

app/models/end_product.rb
leikkisa: 👍 Good call, esp. for searchability.

PR #9600, opened 2019-02-22
Update gems with bummr

monfresh:
**Why**: To have the latest and greatest features and security fixes.


monfresh:
Most of the changes in this PR are a result of running this command to automatically use the block version of attributes for FactoryBot factories:
```console
rubocop \
  --require rubocop-rspec \
  --only FactoryBot/AttributeDefinedStatically \
  --auto-correct
```
This was necessary because static attributes are deprecated in FactoryBot 5.x, and the latest 4.x version throws deprecation messages all over the place when running tests. We are planning on upgrading to 5.x after this PR is merged.

More details here:
https://thoughtbot.com/blog/deprecating-static-attributes-in-factory_bot-4-11

PR #9606, opened 2019-02-22
Fix Raven context reporting

pkarman:
Raven.extra_context is a class method that must be cleared after each job.

See https://docs.sentry.io/clients/ruby/context/#rack-http-context


pkarman:
I wrote more tests for this and can't duplicate. False alarm.

Issue #9361, opened 2019-02-12
Improve new documents icon implementation


aroltsch:
Problem:
Case details uses the last time the user clicked into Reader timestamp while the queue page uses the task's placed_on_hold_at date

Figure out the best way to tackle this problem implementation wise:

Idea:
Case details page:
Use appeals_controller and make a request on the appeal because task info is not needed

Queue page:
Use tasks_controller to get new_documents on per task basis

Bonus point: think of ways to improve on `cached` parameter

AC: 
Write a tech spec outlining the best solution

hschallhorn:
So something we also need to consider is the fact that there are some queues that have an implementation of the new documents icon that are not associated with tasks that are on hold (and therefore would not use the on hold date). [Relevant slack thread](https://dsva.slack.com/archives/C6E41RE92/p1548351413362100).

hschallhorn:
So this looks like it should be pretty simple to do.

**On BE:**
1. Add a route to new_documents for tasks.
2. Return new_docs_for_user on the task's associated appeal, passing in cached as true and the task's placed_on_hold date
3. Remove all mention of cached and place_on_hold_date from appeals_controller and passed cached as false to new_docs_for_user (as any time we're requesting new_documents on an appeal is from the case details page)

**On FE:**
1. Remove all parameters passed from the front end as cached and placed_on_hold are now handled exclusively on the back end.
2. Add newDocsForTask and associated actions/reducers

aroltsch:
Excellent!

1. I think we should also move `number_of_documents_from_caseflow` to the `DocumentFetcher`
1. Then instead of using DocumentConcern we should create a new workflow class called `NewDocumentsForUser` that will have a few attributes: user, appeal, query_vbms (instead of cached), date_to_compare_with (instead of placed_on_hold_at - just like you suggested initially). You can create a method `process` inside
1. In the appeals_controller, you would instantiate `NewDocumentForUser.new(appeal: appeal, user: current_user, query_vbms: true, date_to_compare_with: Time.zone.at(0).process`
1. In the tasks_controller, you would instantiate `NewDocumentForUser.new(appeal: task.appeal, user: current_user, query_vbms: false, date_to_compare_with: task.placed_on_hold_at).process`

Then can you think if we can get rid of `cached` in the `document_count` endpoint?

hschallhorn:
[...]

PR #9611, opened 2019-02-22
DocumentConcern-> NewDocumentsForUser

hschallhorn:
References #9361

Remove Document Concern and move logic to NewDocumentsForUser


app/models/appeal.rb
```diff
@@ -89,6 +88,18 @@ def document_fetcher
   delegate :documents, :manifest_vbms_fetched_at, :number_of_documents,
            :manifest_vva_fetched_at, to: :document_fetcher
 
+  def new_documents
+    @new_documents_for_user ||= NewDocumentsForUser.new(
+      appeal: self, user: RequestStore[:current_user], query_vbms: true, date_to_compare_with: Time.zone.at(0)
```
aroltsch: Why don't we initialize it in the controller instead of moving it here because it is the same for ama and legacy appeal.

hschallhorn: Also was in the middle of pinging you on this. Creeper.

This will be the extent of this PR though. To keep it small

app/models/appeal.rb
```diff
@@ -89,6 +88,18 @@ def document_fetcher
...
+    )
+    @new_documents_for_user.process!
+  end
+
+  def number_of_documents_from_caseflow
```
aroltsch: Based on your comment, let's leave this method in the DocumentConcern then!

app/controllers/appeals_controller.rb
```diff
@@ -47,11 +47,10 @@ def document_count
   end
 
   def new_documents
-    render json: { new_documents: appeal.new_documents_for_user(
-      user: current_user,
-      cached: params[:cached],
-      placed_on_hold_at: params[:placed_on_hold_date]
-    ) }
+    @new_documents_for_user ||= NewDocumentsForUser.new(
+      appeal: appeal, user: current_user, query_vbms: true, date_to_compare_with: Time.zone.at(0)
```
aroltsch: You probably want to send `params[:placed_on_hold_date]` otherwise we will break on_hold column

app/controllers/appeals_controller.rb
```diff
@@ -47,11 +47,10 @@ def document_count
   end
 
   def new_documents
-    render json: { new_documents: appeal.new_documents_for_user(
-      user: current_user,
-      cached: params[:cached],
-      placed_on_hold_at: params[:placed_on_hold_date]
-    ) }
+    @new_documents_for_user ||= NewDocumentsForUser.new(
```
aroltsch: I think we can remove `@` and `||`. You only want to use that if you need this variable in memory throughout the request.

app/services/new_documents_for_user.rb
```diff
@@ -0,0 +1,41 @@
+class NewDocumentsForUser
+  include ActiveModel::Model
+
+  attr_accessor :appeal, :user, :query_vbms, :date_to_compare_with
+
+  def initialize(attributes)
```
aroltsch: I don't think you need `initialize` method because ActiveModel::Model takes care of it for you

app/services/new_documents_for_user.rb
```diff
@@ -0,0 +1,41 @@
+class NewDocumentsForUser
```
aroltsch: Let's put this class in the `workflows` directory. It already exists under app

app/services/new_documents_for_user.rb
```diff
@@ -0,0 +1,41 @@
...
+  # Retrieves any documents that have been uploaded more recently than the user has viewed
+  # the appeal or an optional provided date. Try to load docs from caseflow if query_vbms is
+  # false, load from vbms otherwise
+  def process!
+    caseflow_documents = find_or_create_documents
+    appeal_view = appeal.appeal_views.find_by(user: user)
+
+    return caseflow_documents if !appeal_view && !date_to_compare_with
+
+    compare_date = appeal_view ? [date_to_compare_with, appeal_view.last_viewed_at].max : date_to_compare_with
```
aroltsch: You would need to leave the logic for placed_on_hold here for now based on my comment above.

app/services/new_documents_for_user.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    filter_docs_by_date(caseflow_documents, compare_date)
+  end
+
+  def filter_docs_by_date(documents, date)
```
aroltsch: We can make these methods private if we don't have unit tests for these: filter_docs_by_date and find_or_create_documents

aroltsch:
Nice start! A few comments

app/controllers/appeals_controller.rb
```diff
@@ -47,11 +47,11 @@ def document_count
...
 
   def new_documents
-    render json: { new_documents: appeal.new_documents_for_user(
-      user: current_user,
-      cached: params[:cached],
-      placed_on_hold_at: params[:placed_on_hold_date]
-    ) }
+    placed_on_hold_at = params[:placed_on_hold_date] ? DateTime.strptime(placed_on_hold_at, "%s") : Time.zone.at(0)
+    new_documents_for_user = NewDocumentsForUser.new(
+      appeal: appeal, user: current_user, query_vbms: true, date_to_compare_with: placed_on_hold_at
```
aroltsch: The same for `params[:cached]` right? If we want to use that

Issue #9026, opened 2019-01-29
Manually cancel dispatched LegacyAppeals that are still showing up in VLJ Support Staff queues


lowellrex:
VLJ Support Staff have been complaining for a few weeks that tasks related to dispatched appeals are still showing up in their queue table views. This ticket exists to determine why this is happening and chart a course to resolution of this issue.

Three VACOLS IDs for appeals affected by this are 3626725, 3206295, and 3576330.

## Current AC
- [ ] Identify all cases that have been dispatched but have open Caseflow tasks
- [ ] Change all open Caseflow tasks' statuses to be `completed`

## Previous Acceptance criteria
* [ ] Determine why tasks related to dispatched LegacyAppeals are still showing up in VLJ Support Staff queues
* [ ] Write tickets to address code bugs causing this issue
* [ ] Correct production data (mark tasks complete, etc.) to address the issue as needed.

lowellrex:
The problem with one of these cases (VACOLS ID 3576330) was not that it was worked in VACOLS when it was checked out to the Caseflow location (as we first hypothesized), but rather that the `ColocatedTask` assigned to the VLJ Support Staff organization was completed when we expected the child `ColocatedTask` assigned to an individual person to have been completed. This was caused by a confluence of two major factors:

1. The application [only displaying a single task](https://github.com/department-of-veterans-affairs/caseflow/blob/b80a5838aaaf4034ab9c66eaa373e186f4ab0090/client/app/queue/TaskSnapshot.jsx#L280) in the Task Snapshot area even if there were more than one task with available actions for the appeal. This problem was addressed by a change to the code on 27 December--not deployed until 2 January--that [showed all tasks with available actions](https://github.com/department-of-veterans-affairs/caseflow/pull/8326/files#diff-c6245bd098c838610d8b6f4dbce23d56R275) in Task Snapshot instead of just one.
1. The way the application selected which single task to display was indeterminate and resulted in the `ColocatedTask` assigned to the VLJ Support Staff organization being displayed instead of the child `ColocatedTask` assigned to an individual person. As a result, VLJ support staff using Caseflow unknowingly took action on the organizationall-assigned task instead the individually-assigned task. This returned the case to the assigning attorney and allowed it to continue on to dispatch, but the individually-assigned case remained visible in their individual queues.

[...]

PR #9615, opened 2019-02-22
Oar issue 9295 prod backfill part two

### LINKED ISSUE #9295: Task Associated Object | Backfill production data
For open HoldHearingTasks in production, we need to determine what hearing they're associated with and create a hearing_task_association from the HearingTask to the hearing.

oscarAtNava:
Resolves #9295 and #9026 

### Description
Created script to add hearing_task_associations in prod for existing HoldHearingTasks. Also modified backfill script to create HearingTasks in prod with review feedback received.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Tested all scenarios using modified seed data locally.


scripts/back_fill_hearing_task_associations.rb
```diff
@@ -0,0 +1,24 @@
...
+# Identify all HoldHearingTasks
+# Check if their parent HearingTask has an existing HearingTaskAssociation entry
+# Find hearing associated with HearingTask using appeal id
+# Create HearingTaskAssociation with Hearing and HearingTask info
+#
+# Author: OAR
+# Date:   Feb 19, 2019
+#
+
+hearing_tasks = Task.where(type: "HoldHearingTask").map(&:parent)
```
mdbenjam: Are we actually checking if the parent is a hearing task? It doesn't look like we are.

scripts/back_fill_hearing_tasks.rb
```diff
@@ -1,28 +1,40 @@
...
+legacy_sched_hear_tasks = Task
+  .where("(type='ScheduleHearingTask' OR type='HoldHearingTask') AND appeal_type='LegacyAppeal'")
+  .select do |task|
   task.parent.type == "RootTask"
 end
-puts "Total ScheduleHearingTasks to be relinked: #{legacy_sched_hear_tasks.size}"
+puts "Total HoldHearing or ScheduleHearing Tasks with RootTask as parent: #{legacy_sched_hear_tasks.size}"
+
+tasks_grouped_by_appeal_and_date = legacy_sched_hear_tasks.group_by do |task|
+  "#{task.appeal.id}-#{task.created_at.strftime('%Y-%m-%d')}"
```
mdbenjam: So the idea here is that the hold hearing and schedule hearing tasks that are related should have the same created_at date?

scripts/back_fill_hearing_task_associations.rb
oscarAtNava: We are not. Assumption is this script will run after the backfill to assign HearingTask parents to all HoldHearingTasks. This will be double-checked prior to running the script.

scripts/back_fill_hearing_tasks.rb
oscarAtNava: Correct. If they are not we will not cover that use case. After I run the script I will run queries to see what was left behind and fix manually. The data we pulled yesterday shows most data in prod will be covered by this scenario.

mdbenjam: Interesting, that's a little surprising I'd expect the `closed_at` date of the SHT to be closer to the `created_at` date of the HHT. But I guess they used to be created in quick succession together. So if that's the case then this is probably fine.

scripts/back_fill_hearing_task_associations.rb
```diff
@@ -0,0 +1,24 @@
...
+
+puts hearing_tasks.inspect
+
+hearing_tasks.each do |task|
+  hearing = if task.appeal_type == "LegacyAppeal"
+              LegacyHearing.find_by(appeal_id: task.appeal_id)
```
sharonwarner: A lot of these cases will have multiple hearings, this should return a hearing with no disposition. We'll have to confirm that there aren't any appeals with multiple hearings with no dispositions or no hearings with no dispositions.

scripts/back_fill_hearing_tasks.rb
oscarAtNava: I think your observation is accurate. I will change the code to account for this.

scripts/back_fill_hearing_task_associations.rb
```diff
@@ -0,0 +1,48 @@
...
+  # Get all legacy hearings for appeal id
+  hearings = LegacyHearing.where(appeal_id: task.appeal_id)
+
+  # Need to account for multiple hearings for an appeal. Must choose the one with
+  # disposition of null.
+  # Prod data shows that some vacols identifiers in LegacyHearing may not be present
+  # any more in VACOLS.
+  hearings.each do |hearing|
+    vacols_hearing = VACOLS::CaseHearing.find(hearing.vacols_id)
+    if vacols_hearing.hearing_disp.nil?
```
sharonwarner: Let's put another check to make sure the hearing is in the future. In theory, this should already be the case if the disposition is null... but who knows....

PR #9631, opened 2019-02-25
Delaware Closest Regional Office

### LINKED ISSUE #9510: Wilmington, DE shouldn't show as an RO to schedule because it doesn't have hearing days
Currently, Wilmington, DE shows as an RO in the drop down to Schedule Veterans. 

Wilmington, DE should not get assigned hearing days, so should not have a separate list of Veterans to be scheduled. Those Veterans should be geo-matched to the closest RO that holds hearing. Wilmington, DE is an alt hearing location for Philadelphia hearing days. Essentially, the Veterans for Wilmington, DE can't be scheduled for hearings at the moment.

lomaxap:
Resolves #9510

- After discussion with @MeredithStewart, all appeals whose closest RO is in DE are instead given Philadelphia as `closest_regional_office` (so they appear in the Philly assign hearings list) but are still mapped to Wilmington, DE office as the suggested hearing location.
- Refactors job so that the bulk of work is done by validator service (based on [this comment](https://github.com/department-of-veterans-affairs/caseflow/pull/9503#discussion_r259134216))
- Adds admin action if `appellant.addresss` returns `nil`


app/services/va_dot_gov_address_validator.rb
```diff
@@ -47,4 +92,129 @@ def validate_zip_code
...
+        instructions: error_instructions_map[error_key],
+        admin_action_type: HearingAdminActionForeignVeteranCaseTask
+      )
+    else
+      fail error
+    end
+  end
+
+  def create_admin_action_for_schedule_hearing_task(instructions:, admin_action_type:)
+    task = ScheduleHearingTask.find_by(appeal: appeal)
```
sharonwarner: It's possible an appeal could have multiple ScheduleHearingTasks. We should restrict this to only open ScheduleHearingTasks.

app/services/va_dot_gov_address_validator.rb
```diff
@@ -26,13 +71,13 @@ def address
...
-      address_line1: address[:address_line1],
-      address_line2: address[:address_line2],
-      address_line3: address[:address_line3],
+      address_line1: address[:address_line_1],
+      address_line2: address[:address_line_2],
+      address_line3: address[:address_line_3],
       city: address[:city],
       state: address[:state],
       country: address[:country],
-      zip_code: address[:zip_code]
```
lomaxap: In investigating the duplicate admin action, i realized i did not change this method from the veteran address column names 🤦‍♂️. This means all geomatching from the last week has been done using the zip code fallback.

app/services/va_dot_gov_address_validator.rb
```diff
@@ -47,4 +92,129 @@ def validate_zip_code
...
+
+    admin_action_type.create!(
+      appeal: appeal,
+      instructions: [instructions],
+      assigned_to: HearingsManagement.singleton,
+      parent: task
+    )
+  end
+
+  def except_delaware(closest_regional_office)
```
mdbenjam: So instead of excepting Delaware here, why not just stop passing Delaware's RO's id into the `get_distance_to_facilities` method? In fact I wonder if moving the list of facility ids from the hearings_controller to this module could be useful: 

```
facility_ids = (RegionalOffice::CITIES[params["regional_office"]][:alternate_locations] ||
                     []) << RegionalOffice::CITIES[params["regional_office"]][:facility_locator_id]	                     []) << RegionalOffice::CITIES[params["regional_office"]][:facility_locator_id]
```

mdbenjam: just kidding that is already in this class, but it looks like it's also in the hearings_controller, maybe we can consolidate that. Also, maybe I'm wrong about this? Is it just supposed to be everyone in Delaware maps to Philadelphia's RO.

lomaxap: they should show up in Philadelphia's assign hearings page but should not use Philadelphia's alternate locations when finding available locations. We could either do this here, or add logic to `ScheduleHearingTask#tasks_for_ro` that adds `RO60` to Philadelphia queries. 

In keeping with your earlier comment about having `closest_regional_office` data make sense semantically, i'm thinking i should do it that way now. I've been prioritizing simplifying the `tasks_for_ro` logic over that. (in hindsight, maybe the field should have been named `suggested_regional_office`)

mdbenjam: Okay one more thought, maybe it should just be a mapping in `get_state_code` DE can map to PA?

lomaxap: the get state code determines which alternate locations to use. At that point in the logic, we still want to only use DE locations (which is just Wilmington), PA has 6 other locations that DE residents should map to.

app/services/va_dot_gov_address_validator.rb
```diff
@@ -47,4 +92,129 @@ def validate_zip_code
...
+      city: facility[:city],
+      state: facility[:state],
+      zip_code: facility[:zip_code],
+      classification: facility[:classification],
+      facility_type: facility[:facility_type]
+    )
+  end
+
+  def get_state_code(va_dot_gov_address) # rubocop:disable Metrics/CyclomaticComplexity
+    return "DC" if appeal.is_a?(LegacyAppeal) && appeal.hearing_request_type == :central_office
```
mdbenjam: This feels a little strange to me since the field we're setting is called closest regional office. Do you think it's better to set this here vs just determine it on the query in the `tasks_for_ro` in `ScheduleHearingTask` as we're currently doing?

app/services/va_dot_gov_address_validator.rb
lomaxap: as re: facility_ids method. Should I just add a method to `RegionalOffice` model?

app/services/va_dot_gov_address_validator.rb
mdbenjam: ah was this just passing in nil for those address lines?

app/services/va_dot_gov_address_validator.rb
lomaxap: my thinking was this would save us a VACOLS query on every request to tasks_for_ro and make the endpoint faster.

looking at the new relic data, i think these VACOLS call are related
screenshot 2019-02-27 11 45 04
assuming they are, that's a total of 7.7 we can save for each request.

app/services/va_dot_gov_address_validator.rb
lomaxap: yes

app/services/va_dot_gov_address_validator.rb
mdbenjam: I think most of those are not related to this call. We have to load everything from VACOLS for all of the legacy tasks, so I think that's the majority of this time. That said, there probably are some small savings we can get here. So we can go down this path, we should just consider renaming the field in the future maybe? I think it's a little unclear right now, and I don't want other teams to think it's strictly the closest regional office.

app/services/va_dot_gov_address_validator.rb
mdbenjam: I think adding that method to the regional office model makes sense. Overall this explanation makes sense, it's too bad that Delaware is such a special case :/

PR #9635, opened 2019-02-25
Prevent too many hearings from being scheduled for the same day

### LINKED ISSUE #9607: Bug: Daily Docket allowing over-booking of rooms and lists unavailable days
The Daily Docket is allowing Postponements and rescheduling to days that are already fully booked. 
- For ROs with 2 hearing days per date, both are listed in the postponement drop down regardless of slots left open on that day
- Postponement and rescheduling is successful and puts Veterans on the Daily Docket in excess of the available slots
- Attempting to postpone to another room/day with open slots on the same calendar date fails 
- The Schedule Veterans view is correct and removes unavailable days

See Tamisha's email screenshot for details

sharonwarner:
Resolves #9607 

### Description
This PR adds a backend check such that we can't overschedule one hearing day.


app/models/hearing_day.rb
```diff
@@ -37,20 +53,40 @@ def update_children_records
...
+    fail HearingDayHasChildrenRecords if !vacols_hearings.empty? || !hearings.empty?
   end
 
-  def vacols_children_records
+  def vacols_hearings
     HearingRepository.fetch_hearings_for_parent(id)
   end
 
+  def open_vacols_hearings
+    vacols_hearings.reject { |hearing| [:postponed, :canceled].include?(hearing.disposition) }
```
oscarAtNava: No Show is considered an open hearing?

sharonwarner: Hm, maybe I should name this method something else. This is to signify hearings that can be scheduled over.

PR #9636, opened 2019-02-25
Appeal Status: handle case where request issues are not present

### LINKED ISSUE #9632: Fix for Sentry issue where request_issues is nil
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3826/

It is possible that a Decision Review has been created in Caseflow backend but does not have request issues yet if the intake process is in progress.

The front end does block completed a decision review if there are no request issues.  There is no constraint on the back end.  It is reasonable to assume that a decision review should have request issues and appeal status will not show display claims that have no request issues.

mc-usds:
Resolves #9632

### Description
Updating Appeal Status API to filter out decision reviews that do not have request issues.
Also, added checks that request issues are present in methods that expect them to be there.

### Acceptance Criteria
- [x] Code compiles correctly

### Testing Plan
1. unit tests and feature test pass.


mc-usds:
Thanks. I added a test for check that claims and appeals that have no request issues are filtered out.

PR #9637, opened 2019-02-25
Do not enqueue jobs with delayed processing

pkarman:
connects #9571 

### Description

Having already set a delay on the processing time, respect that delay when enqueuing job.

### DIFF
--- a/app/models/tasks/bva_dispatch_task.rb
+++ b/app/models/tasks/bva_dispatch_task.rb
@@ -33,7 +33,7 @@ class BvaDispatchTask < GenericTask
         delay = decision_document.decision_date.future? ? decision_document.decision_date : 0
         decision_document.submit_for_processing!(delay: delay)
 
-        unless decision_document.processed?
+        unless decision_document.processed? || decision_document.decision_date.future?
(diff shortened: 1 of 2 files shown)


leikkisa:
I always thought if there was a delay, it would still get enqueued but not processed until after the delay. If we don't do ProcessDecisionDocumentJob.perform_later(decision_document.id), is there a way to check that it still does get enqueued later?  

Is this behavior different than elsewhere where we have async jobs with a delay?

pkarman:
The `perform_later` method puts it into the shoryuken queue for immediate async processing. Shoryuken does not know anything about the delay. The delay is used by the cron job and the asyncable methods to figure out what to put on the shoryuken queue.

We tend to do belt and suspenders approaches with our asyncable jobs. I.e. put it in the shoryuken queue immediately and also rely on the asyncable+cron logic if the shoryuken job fails.

In the case of an intentional delay, we have to only use asyncable+cron approach.

pkarman:
here's a good example of this pattern: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/end_product_establishment.rb#L253

note how if there is a delay, we only submit for processing, w/o the `perform_later`.

PR #9651, opened 2019-02-26
Postpone Hearings

lomaxap:


lomaxap:
@sharonwarner wanted to get your thoughts on this approach before I implement. I want to create 2 endpoints for each postponement type. I think this is easier to follow rather than having a single `/postpone` endpoint with a bunch of conditionals. It also makes these endpoints easier to re-use elsewhere in the application if don't bake too much daily docket logic into each endpoint.

the frontend would look something like

```javascript
schedule = () => {
  const { hearing } = this.props;
  ApiUtil.put(`/hearings/${hearing.id}/schedule`).then(() => { ... });
}

postpone = () => {
  const { postponementType, hearing } = this.props;
  ApiUtil.put(`/hearings/${hearing.id}/postpone`).then(() => {
    // do something if error
    switch(postponementType){
     case 'schedule':
       schedule();
        break;
      ...
    }
  });
}
```

PR #9660, opened 2019-02-26
Do not create Veteran record unless accessible

pkarman:
connects #9448 

## Description

This change affects the logic for when a `Veteran` record is created. It will only be created if both `found?` and `accessible?` are true.

### DIFF
--- a/spec/controllers/intakes_controller_spec.rb
+++ b/spec/controllers/intakes_controller_spec.rb
@@ -3,7 +3,13 @@ RSpec.describe IntakesController do
     Fakes::Initializer.load!
     FeatureToggle.enable!(:intake)
     User.authenticate!(roles: ["Mail Intake"])
+
+    allow_any_instance_of(Fakes::BGSService).to receive(:fetch_veteran_info).and_call_original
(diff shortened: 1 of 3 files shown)


spec/controllers/intakes_controller_spec.rb
```diff
@@ -21,6 +27,59 @@
...
+      let!(:veteran) {} # no-op
+
+      it "does not create Veteran db record in Caseflow" do
+        expect(Veteran.find_by_file_number_or_ssn(file_number)).to be_nil
+        expect(Intake.find_by(veteran_file_number: file_number)).to be_nil
+        post :create, params: { file_number: file_number, form_type: "higher_level_review" }
+        expect(response.status).to eq(422)
+        expect(controller.send(:new_intake).error_code).to eq("veteran_not_accessible")
+        expect(Veteran.find_by_file_number_or_ssn(file_number)).to be_nil
+        expect(Intake.find_by(veteran_file_number: file_number)).to_not be_nil
```
leikkisa: Nice

Issue #9548, opened 2019-02-20
AMA Intake | Better error message for non unique phone number


pkarman:
The BGS error
```
NonUniqueResultException
The query on candidate type "class gov.va.vba.corporate.entities.PtcpntPhone" with filter "SELECT p from PtcpntPhone AS p WHERE p.compId.ptcpntId = :ptcpntId AND p.compId.phoneTypeNm = :phoneTypeNm AND p.compId.efctvDt IS NOT NULL AND p.endDt IS NULL" was configured to have a unique result, but more than one instance matched the query.
```

Happens because two or more Veteran records have been merged in CorpDB and now there are multiple active phone numbers.

The workaround is to log in to SHARE and update the Veteran record. See https://github.com/department-of-veterans-affairs/appeals-support/issues/4020#issuecomment-464790349

Our error message should include that helpful step.

## Technical Note

* A new method to the BGS library similar `can_access?` but that returns both a boolean and the reason.

leikkisa:
Martina sent us a screenshot! Considering this Validated.

PR #9662, opened 2019-02-26
Add descriptive error if Veteran has multiple active phone numbers

leikkisa:
connects #9548


leikkisa:
I looked at adding another method to the bgs library, but can_access? just calls the same service as fetch_veteran_info so I wasn't sure if there was that much of an advantage to putting it in there, unless we want to replace can_access? altogether.

app/models/veteran.rb
```diff
@@ -126,6 +127,16 @@ def accessible?
     bgs.can_access?(file_number)
```
pkarman: IME the error gets thrown in this call to `bgs.can_access?(file_number)` before `fetch_bgs_record` ever gets called, so `veteran.accessible?` is going to be false.

app/models/veteran.rb
```diff
@@ -126,6 +127,16 @@ def accessible?
...
   end
 
+  def multiple_phone_numbers?
+    access_error&.include?("NonUniqueResultException")
```
pkarman: this is a clever way of calling `fetch_bgs_record` when checking for errors, in order to trigger the exception.

app/models/intake.rb
```diff
@@ -191,7 +194,7 @@ def validate_start
       self.error_code = :veteran_not_found
 
     elsif !veteran.accessible?
-      self.error_code = :veteran_not_accessible
+      self.error_code = veteran.multiple_phone_numbers? ? :veteran_has_multiple_phone_numbers : :veteran_not_accessible
```
pkarman: I suspect you could avoid disabling rubocop if you put this into its own method. e.g. `veteran.accessible_error`

spec/controllers/intakes_controller_spec.rb
```diff
@@ -77,7 +77,6 @@
         expect(controller.send(:new_intake).error_code).to eq("veteran_not_accessible")
         expect(Veteran.find_by_file_number_or_ssn(file_number)).to be_nil
         expect(Intake.find_by(veteran_file_number: file_number)).to_not be_nil
-        expect(bgs).to_not have_received(:fetch_veteran_info)
```
pkarman: could we change this instead of deleting it?
```
expect(bgs).to have_received(:fetch_veteran_info).exactly(1).times
```

spec/feature/intake/intake_spec.rb
```diff
@@ -141,6 +141,29 @@
       end
     end
 
+    context "Veteran records have been merged and Veteran has multiple active phone numbers in SHARE" do
```
pkarman: good test!

app/models/veteran.rb
```diff
@@ -126,6 +127,18 @@ def accessible?
     bgs.can_access?(file_number)
   end
 
+  def access_error
```
pkarman: it would be good to add tests for these methods on the veteran model spec, to keep our coverage up.

app/models/intake.rb
leikkisa: Cool, I thought about that too. Will do!

spec/controllers/intakes_controller_spec.rb
leikkisa: 👍

app/models/veteran.rb
leikkisa: 👍

app/models/intake.rb
```diff
@@ -191,7 +192,7 @@ def validate_start
       self.error_code = :veteran_not_found
 
     elsif !veteran.accessible?
-      self.error_code = :veteran_not_accessible
+      veteran_accessible_error
```
pkarman: nit: could we name this `set_veteran_accessible_error` to make it a verb? it's setting an attribute, not just returning one.

pkarman: or rewrite it as:

```ruby
self.error_code = veteran_accessible_error
```

for consistency?

leikkisa: Makes sense!

PR #9663, opened 2019-02-26
Add global team admin page

### LINKED ISSUE #9626: Create page for creation and management of Organizations
Caseflow has a concept of [`Organization`s](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/organization.rb) that have members with access to certain behaviours of Caseflow based on organization membership. Administrators of individual organizations have access to a user management page for their organization like the one below:

image

This page works well for the administration of individual teams. However, we lack a page where a "super" admin (like @marvokdolor-gov or likely somebody at the Board in the future) can get a quick list of all teams or even create new teams. I think this page could serve two purposes:
1. A centralized place for the creation of new teams (`JudgeTeam`s and `Vso`s in particular) and the modification (changing VSO's BGS participant IDs) or deletion (when a judge leaves the Board) of existing teams.
1. A place to find links to the user administration pages for each existing team.

lowellrex:
Resolves #9626.

image

### How to play with these changes locally
* `bundle exec rake db:seed`
* Log in as `TEAM_ADMIN`
* Visit `/team_management`

### TODO
* [ ] Inline feedback after update operation
* [ ] Lots of validation


app/controllers/application_controller.rb
```diff
@@ -130,21 +130,18 @@ def application_urls
...
-      }
+      { title: "Help", link: help_url },
+      { title: "Send Feedback", link: feedback_url, target: "_blank" }
     ]
 
     if current_user&.administered_teams&.any?
       urls.concat(manage_teams_menu_items)
     end
 
+    if Bva.singleton.user_has_access?(current_user)
```
lowellrex: Access to the new App-wide Team Admin page is currently limited to members of the `Bva` organization (the `Bva` organization currently has no members). I'm not sure if membership in this organization is the correct answer, but I like using membership in *an* organization for access control so that this page is somewhat self-referential (organization membership determines who can view the page that is nominally related to organization membership). The whole point of this page is to allow non-developers on the Caseflow team (and eventually Board employees?!) to create `JudgeTeam`s and `Vso`s without intervention by a developer. If we were to use functions or feature toggles to manage access to this page then it wouldn't be fully self-service like it would be if we use organization membership. The flip side of that is that we remove the ability for developers to act as backstops.

app/controllers/team_management_controller.rb
```diff
@@ -0,0 +1,66 @@
+class TeamManagementController < ApplicationController
```
lowellrex: Heavily inspired by the design of [`Organizations::UsersController`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/organizations/users_controller.rb).

app/controllers/users_controller.rb
```diff
@@ -9,11 +9,22 @@ def index
       return render json: { attorneys: Attorney.list_all }
     when Constants::USER_ROLE_TYPES["hearing_coordinator"]
       return render json: { coordinators: User.list_hearing_coordinators }
+    when "non_judges"
```
lowellrex: Endpoint we hit to get the list of all users who are not currently owners of judge teams.

config/routes.rb
```diff
@@ -204,6 +204,11 @@
     get '/:user_id(*rest)', to: 'legacy_tasks#index'
   end
 
+  resources :team_management, only: [:index, :update]
+  get '/team_management(*rest)', to: 'team_management#index'
+  post '/team_management/judge_team/:user_id', to: 'team_management#create_judge_team'
+  post '/team_management/national_vso', to: 'team_management#create_national_vso'
```
lowellrex: I'm not at all confident about the naming of these routes and am up for suggestions here.

client/app/queue/TeamManagement.jsx
```diff
@@ -0,0 +1,235 @@
...
+import { onReceiveTeamList } from './teamManagement/actions';
+import { withRouter } from 'react-router-dom';
+
+const tableStyling = css({
+  width: '100%',
+  '& td': { border: 'none' },
+  '& input': { margin: 0 }
+});
+
+class TeamManagement extends React.PureComponent {
```
lowellrex: This is probably a decent place to start when reviewing this PR. This is the component we render when we load `/team_management`.

client/app/queue/TeamManagement.jsx
```diff
@@ -0,0 +1,235 @@
...
+            </tbody>
+          </table>
+
+        </div>
+      </AppSegment>
+    </LoadingDataDisplay>;
+  };
+}
+
+const mapStateToProps = (state) => {
```
lowellrex: Put stuff in global redux store so that the additions made in the modals are reflected the table view.

client/app/queue/TeamManagement.jsx
```diff
@@ -0,0 +1,235 @@
...
+      user_admin_path: props.user_admin_path
+    };
+  }
+
+  changeName = (value) => this.setState({ name: value });
+  changeUrl = (value) => this.setState({ url: value });
+  changeParticipantId = (value) => this.setState({ participant_id: value });
+
+  // TODO: Add feedback around whether this request was successful or not.
+  submitUpdate = () => {
```
lowellrex: Still need to add inline success/error messages for the update operations.

app/controllers/application_controller.rb
mdbenjam: I think this generally makes sense. Also not sure if this is the right organization. Maybe there should be a separate administrative org. But for now I think this makes sense.

client/app/queue/TeamManagement.jsx
```diff
@@ -0,0 +1,235 @@
...
+      <td>
+        <TextField
+          name="Name"
+          label={false}
+          value={this.state.name}
+          onChange={this.changeName}
+        />
+      </td>
+      <td>
+        <TextField
```
mdbenjam: Now that we have all this stuff built out and good navigation between various Queues, we can think about going to your original idea of just using the ids in the URLs. Making that switch shouldn't be very painful now. Just a thought.

PR #9667, opened 2019-02-27
Add filtering to reader

annekainicUSDS:

## Description
Pushing up this PR for thoughts.

There are a few main features of the new `FilterableTable` component:
1. **Generating the list of options** a column can be filtered on by going through the set of data in that column and reducing it to a set of unique values.
2. **Keeping track of selected filters** by listening for clicks on each filter checkbox or clicks on the `clearFilters` links and storing its `checked` status in local state
3. **Filtering the table data** based on which filters have been selected

### Challenges
I struggled with converting Reader to using the new `FilterableTable` component because it challenged the basic way a couple of these main features of `FilterableTable` component works.

#### Generating the list of filter options
The `FilterableTable` component assumes a direct relationship between a value in the data store and what the table cell displays. For example, let's say we have data for the table that has a key called `animal`. The value for that key will be a string that lists the animal. So for the following table data:
```
[
  { animal: 'tiger' },
  { animal: 'giraffe' },
  { animal: 'elephant' }
]
```

We'll see the following table:

| Animal | ... |
|---|---|
| tiger | ... |
| giraffe | ... |
| elephant | ... |

Reader has two columns that can be filtered. Both columns depend on unique data structures in the table data:
1. The Category column can show multiple categories in a single cell. Each category is its own key in the data. The table data could look like this:
```
[
  { category_1: true, category_2: false, category_3: false },
  { category_1: false, category_2: true, category_3: false },
  { category_1: true, category_2: false, category_3: true }
]
```

And the corresponding table will look like this:

| Category | ... |
|---|---|
| category_1 | ... |
| category_2 | ... |
| category_1, category_3 | ... |

2. The Tag column can also show multiple tags in a single cell, but instead of separate keys for each tag, there's a single `tag` key that points to an array of tags.
```
[
  {  tags: [{ text: 'something' }] },
  { tags: [{ text: 'something_else' }] },
  { tags: [{ text: 'something' }, { text: 'something_else' }]}
]
```

The table will look like this:

| Tag | ... |
|---|---|
| something | ... |
| something_else | ... |
| something, something_else | ... |

Because of the different data structures for each filterable column, it makes it really difficult to depend on `FilterableTable`'s internal function to generate filter options

**Solution:** I created a new prop called `customFilterOptions` that can be passed to each column to bypass the default function to generate the filter options within `FilterableTable`.

#### Keeping track of selected filters
`FilterableTable` can still do this for Reader. But the trouble is that other components outside of the table need access to this information. The search bar at the top of Reader needs to know which filters are selected because you should only be able to search from the filtered data. The individual view of each document also needs to know which filters are selected because there's functionality to go to the next or previous document from that view, and that will depend on the filtered table data.

**Solution:** I created a new prop called `onFilterChange` that will pass up the updated `filteredByList` in `FilterableTable`'s local state back up to the parent component, where it will then be shared throughout the app as needed.

#### Filtering the table data
Filtering the table data also becomes very difficult because of the unexpected data structures that can arise. So I also relied on the existing functionality in Reader to filter the table data and bypassed the `FilterableTable` component's default functionality to do this.

### Where this stands
What these changes come down to is that I basically had to disable the main features of `FilterableTable` in order to handle the complexities and specificities of how Reader works. It makes me wonder if it's even useful to use `FilterableTable` for Reader, or if we should just keep the `Table` component it was using before. I guess the question is: does the benefit of having only one component for a table across caseflow outweigh the fact that in order to use only one component we'll need to add so many different props and hooks to shape it into exactly what each app needs? Would love thoughts on this!


lowellrex:
I really appreciate the deep dive into this one! And I think you're right that the complexity here might outweigh the desire to consolidate all tables into `FilterableTable`. So, I think we're fine to leave Reader's table as a special case and not roll it into `FilterableTable`.

annekainicUSDS:
Okay sounds good, going to close this PR then!

PR #9673, opened 2019-02-27
Use assigned_at date if placed_on_hold_at is nil.

### LINKED ISSUE #9671: New documents icon is broken for legacy appeals because we don't save AttorneyTasks for legacy appeals in our tasks model
AC:
If placed_on_hold is nil for a record, use task.assigned_at instead. This is due to this code: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/queues/attorney_queue.rb#L6

image

hschallhorn:
Resolves #9671

### Description
In attorney queues FE, on_hold tasks don't have a placed_on_hold_at date but rather use the assigned_at date instead. This ticket uses the assigned_at date on the BE to look for new docs

### Acceptance Criteria
- [ ] No 500s on new_documents in production for attorney queues


hschallhorn:
> Approved if you could add a quick feature test.

"quick". added an extra test in there to make sure placed_on_hold_at works as well

PR #9678, opened 2019-02-27
Include tasks assigned by an attorney in reader user job

### LINKED ISSUE #9672: Document count for on hold tab for attorneys is inaccurate
Document count for an on-hold tab for attorneys is inaccurate for legacy appeals. This is due to how we fetch on hold tasks for attorneys https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/queues/attorney_queue.rb#L6

For legacy appeals, when an attorney assigns a colocated admin action, we mark appeal's location to Caseflow in Vacols and create a colocated task. For legacy appeals, there is no attorney task in caseflow so we need to use assigned_by to search for on hold tasks for attorneys in our reader job.

Make a change here: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/jobs/fetch_documents_for_reader_user_job.rb#L20

hschallhorn:
Resolves #9672 

### Acceptance Criteria
- [ ] Doc count for attorneys is correct in production

### DIFF
--- a/app/jobs/fetch_documents_for_reader_user_job.rb
+++ b/app/jobs/fetch_documents_for_reader_user_job.rb
@@ -20,6 +20,8 @@ class FetchDocumentsForReaderUserJob < ApplicationJob
     ama_user_tasks = Task.active.where(assigned_to: reader_user.user)
     ama_appeals = ama_user_tasks.map(&:appeal).uniq
 
+    # Attorney Legacy Tasks are not yet stored in Caseflow Tasks. However, we can grab
+    # the ones "on hold" by looking for colocated tasks they have assigned
     attorney_user_tasks = []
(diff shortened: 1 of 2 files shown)


app/jobs/fetch_documents_for_reader_user_job.rb
```diff
@@ -20,7 +20,13 @@ def perform(reader_user)
     ama_user_tasks = Task.active.where(assigned_to: reader_user.user)
     ama_appeals = ama_user_tasks.map(&:appeal).uniq
 
-    fetch_documents_for_appeals(legacy_appeals + ama_appeals)
+    attorney_user_tasks = []
+    if reader_user.user.attorney_in_vacols?
```
aroltsch: Let's add a comment explaining why we are doing that.

PR #9681, opened 2019-02-27
Remove Issues Page | Decision dates in modal are one day ahead of decision dates on page

### LINKED ISSUE #9650: Intake Add/ Remove Issues Page | Decision dates in modal are one day ahead of decision dates on page
Observed in UAT on file `123458314 `: 
1. I intake a form and land on the Add/ Remove Issues page (https://appeals.cf.uat.ds.va.gov/intake/add_issues)
2. I select a rated issue "Service connection for ptsd is granted..." under the heading `Past decisions from 07/30/2018`
3. When I add the issue, I see the line `Decision date: 07/29/2018`

This is repeatable for other file numbers and issues.

screencapture

Sjones352:
connects #9650

### Description

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to Add/Remove Issues

**Past Decision**
screen shot 2019-03-04 at 12 11 28 pm

**Rating added showing correct date**
screen shot 2019-03-04 at 12 10 49 pm

**Rating can not be checked**
screen shot 2019-03-04 at 12 11 50 pm


client/app/intake/util/issues.js
```diff
@@ -320,11 +320,11 @@ export const formatAddedIssues = (intakeData, useAmaActivationDate = false) => {
         throw new Error('no decision date');
       }
       const decisionDate = new Date(issue.decisionDate || issue.approxDecisionDate);
-
       return {
         referenceId: issue.id,
         text: issue.description,
-        date: formatDateStr(decisionDate),
+        // formatDatStr without converting to Date time
```
jiahuang: Can you add in another comment in here explaining that the Datetime conversion converts to the local timezone instead of utc, which may result in an incorrect date?

client/app/intake/util/issues.js
```diff
@@ -320,11 +320,11 @@ export const formatAddedIssues = (intakeData, useAmaActivationDate = false) => {
         throw new Error('no decision date');
       }
       const decisionDate = new Date(issue.decisionDate || issue.approxDecisionDate);
```
pkarman: could we instead just change this to:
```
const decisionDate = issue.decisionDate || issue.approxDecisionDate;
```
since the `decisionDate` variable also gets passed? I'm not sure it needs to be a `Date` object at all.

client/app/intake/util/issues.js
leikkisa: I think Jia is saying to update this comment:
// formatDatStr without converting to Date time
to something like this:
// formatDatStr converts to the local timezone instead of UTC

client/app/intake/util/issues.js
```diff
@@ -319,12 +319,13 @@ export const formatAddedIssues = (intakeData, useAmaActivationDate = false) => {
...
       }
-      const decisionDate = new Date(issue.decisionDate || issue.approxDecisionDate);
+      const decisionDate = issue.decisionDate || issue.approxDecisionDate;
 
       return {
         referenceId: issue.id,
         text: issue.description,
-        date: formatDateStr(decisionDate),
+        // formatDatStr without converting to Date time
+        date: formatDateStr(issue.decisionDate || issue.approxDecisionDate),
```
leikkisa: I don't understand why this line is changed since decisionDate is equal to issue.decisionDate || issue.approxDecisionDate in the new code.

client/app/intake/util/issues.js
```diff
@@ -319,12 +319,13 @@ export const formatAddedIssues = (intakeData, useAmaActivationDate = false) => {
         console.warn(issue);
         throw new Error('no decision date');
       }
-      const decisionDate = new Date(issue.decisionDate || issue.approxDecisionDate);
+      const decisionDate = issue.decisionDate || issue.approxDecisionDate;
```
leikkisa: I'm wondering how changing this will impact the other places this is used below, like beforeAma and also just plain decisionDate.

client/app/intake/util/issues.js
Sjones352: That line changed because the `decisionDate` is formatting the time Eastern standard hence the error the one day error we are getting.  It is because `issue.decisionDate || issue.approxDecisionDate` is passed in the new Date object hence converting it in Eastern standard time. `formatDateStr(issue.decisionDate || issue.approxDecisionDate)` converts it UTC time zone and the right `Past decision date` is displayed.

leikkisa: That looks good to me!

PR #9685, opened 2019-02-27
Add check for nil compare date

hschallhorn:
Resolves 500s on new_documents

### Description
Add check for nil compare date


app/workflows/new_documents_for_user.rb
```diff
@@ -12,6 +12,7 @@ def process!
 
     return caseflow_documents if !appeal_view && !date_to_compare_with
 
+    @date_to_compare_with ||= Time.zone.at(0)
```
lowellrex: This is just applying the patch we have in place for the [`AppealsController.new_documents` endpoint](https://github.com/department-of-veterans-affairs/caseflow/blob/43634de6b9135082085d2bc06f4cbf358e939bbd/app/controllers/appeals_controller.rb#L51) to the [similar function on the `TasksController`](https://github.com/department-of-veterans-affairs/caseflow/blob/43634de6b9135082085d2bc06f4cbf358e939bbd/app/controllers/tasks_controller.rb#L138), correct?

lowellrex: Also, why do we want two different `new_documents` endpoints? One accepts a task ID and the other an appeal ID, but do they return the same data if the task is related to the same appeal?

hschallhorn: yep, we could alternatively do `date_to_compare_with: task.placed_on_hold_at || task.assigned_at` as well

hschallhorn: The endpoint for tasks uses the cached documents (as this always comes from queue) and the task's on hold date. The endpoint for appeals (from case details) uses vbms and has no associated task or alternative date. Anya preferred having these split into two endpoints rather than passing parameters with the request. This was done in #9639

lowellrex: Gotcha gotcha. Can we change the [name of the endpoint](https://github.com/department-of-veterans-affairs/caseflow/blob/43634de6b9135082085d2bc06f4cbf358e939bbd/app/controllers/tasks_controller.rb#L142) we send to datadog for one or both of these to differentiate them? Perhaps `appeal_new_documents` and `tasks_new_documents`?

hschallhorn: 💯

aroltsch: @hschallhorn are we sure the 500 errors were caused by `date_to_compare_with` being nil? I don't see how it could be possible.

Issue #8655, opened 2019-01-11
Hearing Disposition - Postponed


ShellySmithDSVA:
## User Stories
As a Hearings Schedule or Hearing Prep user, I need to be able to postpone hearings for many reasons. I may also need to immediately re-schedule the vet, or put them back in the list, or further work the case to make sure the vet gets their hearing. I also need to be able to cancel and close the case should the circumstances and judge decisions point to that path.

## Mocks
Postpone from Case Details: https://va-gov.invisionapp.com/share/HTQRIQBJNMS
Postpone from Hearing Prep Daily Docket: https://va-gov.invisionapp.com/share/78QRITG9ZAR
Postpone from Daily Docket: https://va-gov.invisionapp.com/share/AMQRITPTE43

## Acceptance Criteria
- [ ] Verify when the user marks a hearing as Postponed in case details, a the postponement module is displayed.
- [ ] Verify the postponement module allows for users to select only one of the radio button options. Checking another option toggles the previous option off.
- [ ] Verify the module contains the instruction text outlined in the mocks.
- [ ] In only hearing schedule, verify the user has the option to choose Reschedule Immediately. Choosing this option presents the user with the Schedule Veteran options in the modal.
- [ ] Verify Send to Schedule Veteran is presented as an option. Choosing this option will send the case to the list of Veterans to be scheduled.
- [ ] Verify the Add Admin Action provides the user with a list of tasks in a drop down and a place to add task instructions. 
- [ ] Verify the Select Reason provides users with the list of admin actions identical to the admin actions. Make sure the list includes:
Health issue
Rep/Veteran not available 
- [ ]  Verify confirming Admin Actions sends this case to the Hearings Schedule team queue as unassigned. 

Backend work:
- [ ] If the user selects postpone and reschedule immediately, we need to close the DispositionTask and HearingTask, create a new HearingTask off the original HearingTask's parent associated with the new hearing, and create a new DispositionTask.
- [ ] If the user selects postpone and reschedule later,  we need to close the DispositionTask and HearingTask, create a new HearingTask off the original HearingTask's parent associated with the new hearing, and create a new ScheduleHearingTask.
- [ ] If the user selects postpone with an admin action,  we need to close the DispositionTask and HearingTask, create a new HearingTask off the original HearingTask's parent associated with the new hearing, and create a new ScheduleHearingTask and HearingAdminTask.

Engineering notes:
[...]

PR #9686, opened 2019-02-27
Disposition Task Logic: Postponed

lomaxap:
#8655 (backend pieces)

This PR sets up logic for all `DispositionTask` updates.

`POST /tasks/:dispositionTaskId/` can receive the following payload:
```javascript
{
  status: "cancelled",
   business_payloads: {
     values: {
      disposition: "postponed|held|no_show|cancelled"
      after_disposition_update: {
        action: "reschedule|schedule_later|release"
        // if reschedule
        new_hearing_attrs: {
          hearing_day_id: "4444444", // hearing_day_id
          hearing_location: { facility_id: ... },
          hearing_time: { h: 4, m: 10, offset: "-05:00" }
        }
        // if schedule_later
        with_admin_action_klass: "HearingAdminActionVerifyPoaTask|HearingAdminActionIncarceratedVeteranTask|etc..",
        admin_action_instructions: "Fix. This."
      }
    }
  }
}
```


app/models/tasks/hold_hearing_task.rb
```diff
@@ -12,4 +12,86 @@ def create_hold_hearing_task!(appeal, parent, hearing)
...
+  def schedule_later(with_admin_action_klass: nil)
+    new_hearing_task = hearing_task.cancel_and_recreate
+
+    schedule_task = ScheduleHearingTask.create!(
+      parent: new_hearing_task,
+      appeal: appeal,
+      assigned_to: HearingsManagement.singleton
+    )
+    if with_admin_action_klass.present?
+      with_admin_action_klass.constantize.create!(
```
lomaxap: do we actually need to create any hearing admin action? Or would just creating `HearingAdminActionTask` here be sufficient

app/models/tasks/hold_hearing_task.rb
```diff
@@ -12,4 +12,86 @@ def create_hold_hearing_task!(appeal, parent, hearing)
...
+
+  def no_show() end
+
+  def cancelled() end
+
+  def held() end
+
+  private
+
+  def slot_new_hearing(hearing_day_id, hearing_type, hearing_time, hearing_location)
```
lomaxap: would be nice to refactor this to use the same method / logic as ScheduleHearingTask

app/models/tasks/hearing_task.rb
```diff
@@ -1,3 +1,13 @@
 class HearingTask < GenericTask
   has_one :hearing_task_association
+
+  def cancel_and_recreate
```
lomaxap: dupe of [this](https://github.com/department-of-veterans-affairs/caseflow/pull/9299/files#diff-c95c9451cf097bdd58f73d0f9b114f97)

app/models/tasks/disposition_task.rb
```diff
@@ -27,4 +68,72 @@ def check_parent_type
...
+        instructions: after_disposition_update[:admin_action_instructions]
+      )
+    end
+  end
+
+  def no_show() end
+
+  def cancelled() end
+
+  def held() end
```
tomas-nava: I find the naming of these methods confusing... instead of `postponed`, `no_show`, `cancelled`, `held` would `set_hearing_postponed`, `set_hearing_no_show`, `set_hearing_cancelled`, and `set_hearing_held` make sense?

app/models/tasks/disposition_task.rb
```diff
@@ -27,4 +68,72 @@ def check_parent_type
...
+        instructions: instructions.present? ? [instructions] : nil,
+        assigned_to: HearingsManagement.singleton
+      )
+    end
+  end
+
+  def release() end
+
+  def postponed(after_disposition_update:)
+    hearing = hearing_task.hearing_task_association.hearing
```
tomas-nava: Could we delegate `hearing`? something like:

```ruby
def hearing_task_association
  hearing_task&.hearing_task_association
end
```

and

```ruby
delegate :hearing, to: :hearing_task_association, allow_nil: true
```

...with the caveat that `parent` may not be a `HearingTask` as mentioned in my other comment.

app/models/tasks/disposition_task.rb
```diff
@@ -8,14 +8,55 @@ class DispositionTask < GenericTask
...
+      if parent.is_a? HearingTask
+        HearingTaskAssociation.create!(hearing: hearing, hearing_task: parent)
+      end
+
+      disposition_task
+    end
+  end
+
+  def hearing_task
+    parent
```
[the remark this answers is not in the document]
lomaxap: need to check with @sharonwarner on this one. I'm pretty sure that should never happen. And if it does it should throw an error.

app/models/tasks/disposition_task.rb
lomaxap: FYI, I got [this rubocop error](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/AccessorMethodName) so went with `mark_hearing_postponed` instead

app/models/tasks/disposition_task.rb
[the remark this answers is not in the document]
sharonwarner: Yes, at this point we can assume that the DispositionTasks's parent is always a HearingTask. You might need to merge master into this PR, but Oscar added validations.

PR #9689, opened 2019-02-28
Supplemental Claims table comments

leikkisa:
connects #9559


leikkisa:
When there's crossover with other similar tables (like Higher Level Reviews), I'll take feedback on this one and implement it there too.

db/migrate/20190228004408_add_comments_for_supplemental_claims.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    change_column_comment(:supplemental_claims, :establishment_submitted_at, "Timestamp for when the Supplemental Claim was submitted by a Claims Assistant. This adds the End Product Establishment to a job to finish processing asynchronously.")
+
+    change_column_comment(:supplemental_claims, :legacy_opt_in_approved, "Indicates whether a Veteran opted to withdraw their Supplemental Claim request issues from the legacy system if a matching issue is found. If there is a matching legacy issue and it is not withdrawn, then that issue is ineligible to be a new request issue and a contention will not be created for it.")
+
+    change_column_comment(:supplemental_claims, :receipt_date, "The date that the Supplemental Claim form was received by central mail. Only issues decided prior to the receipt date will show up as contestable issues.  It is also the claim date for any associated end products that are established. Supplemental Claims do not have the same timeliness restriction on contestable issues as Appeals and Higher Level Reviews.")
+
+    change_column_comment(:supplemental_claims, :uuid, "The universally unique identifier for the Supplemental Claim.")
```
leikkisa: I don't really know what the UUID is used for other than for Appeals.

pkarman: The UUID is used in the "public" URL generated for linking to editing a decision review from a non-comp task. Maybe we could amend the comment to:

> The universally unique identifier for the Supplemental Claim. Can be used to link to the claim after it is completed.

PR #9690, opened 2019-02-28
Allow task assignee to send note along with a task when it is completed #8610

### LINKED ISSUE #8610: Allow task assignee to send note along with a task when it is completed
There is currently no way for VLJ support staff to send a message to the attorney who created the admin action when VLJ support staff complete their task and send the case back to the attorney (see [related slack thread](https://dsva.slack.com/archives/C6E41RE92/p1547140305090400)). Attaching a note about what the resolution of the task was from the task assignee back to the task creator is a common pattern I expect we will see frequently.

We've proposed a few ideas about how we could accomplish this (#7552 and #7254) but we should not limit ourselves to only those approaches.

## Acceptance criteria
* [ ] When folks complete a task they can attach a note that the task creator can see
* [ ] Task creator can see notes from folks who completed child tasks in the task snapshot

youngfreezyVA:
Resolves #8610 

### Description
Allow VLJ staff to send note to attorney that sent the case to them

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
as an attorney, add an admin action, assign to ihp, and it will be sent to vlj.
switch to vlj, send to an attorney, enter instructions
you'll see it on attorney timeline details!


spec/feature/queue/colocated_task_queue_spec.rb
```diff
@@ -63,6 +64,13 @@
 
       # Click into case details page. Expect to see draft decision option.
       click_on(appeal.veteran.name.formatted(:readable_full))
+      # verify that the instructions from the VLJ appear on the case timeline
+      sleep 1
+      page.find_all("button", text: "View task instructions")[1].click
+      sleep 1
```
tomas-nava: I would replace lines 68-70 with

```ruby
scroll_element_in_to_view("#case_timeline-section")
view_text = COPY::TASK_SNAPSHOT_VIEW_TASK_INSTRUCTIONS_LABEL
page.find_all("table#case-timeline-table button", text: view_text).each(&:click)
```
This scrolls the page to the timeline and clicks on both view buttons; it doesn't seem to need the sleep.

client/app/queue/components/CompleteTaskModal.jsx
```diff
@@ -3,31 +3,53 @@ import { bindActionCreators } from 'redux';
...
+import { ATTORNEY_COMMENTS_MAX_LENGTH, marginTop } from '../constants';
 import COPY from '../../../COPY.json';
 import CO_LOCATED_ADMIN_ACTIONS from '../../../constants/CO_LOCATED_ADMIN_ACTIONS.json';
 
 import {
   taskById,
   appealWithDetailSelector
 } from '../selectors';
-import { onReceiveAmaTasks } from '../QueueActions';
+import { onReceiveAmaTasks, setDecisionOptions } from '../QueueActions';
```
tomas-nava: I don't think we need `setDecisionOptions` here; I think it'd be better to follow the pattern in [`AssignToView.jsx`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/AssignToView.jsx#L160).

In fact I think you can remove references to decisions throughout the component.

client/app/queue/components/CompleteTaskModal.jsx
```diff
@@ -3,31 +3,53 @@ import { bindActionCreators } from 'redux';
...
 import {
   requestPatch
 } from '../uiReducer/uiActions';
 import editModalBase from './EditModalBase';
 import { taskActionData } from '../utils';
+const renderInstructionsFieldForColocatedTask = (props) => {
+  if (props.task.type === 'ColocatedTask' && !props.task.isLegacy) {
```
tomas-nava: I checked with @lowellrex ... we don't need to restrict this field to only `ColocatedTask`s, and he doesn't think legacy tasks will ever see this modal. So you can remove this conditional.

Issue #9587, opened 2019-02-21
Fix N+1 queries reported by the bullet gem


aroltsch:
AC:
Fix N+1 queries reported by the bullet gem

monfresh:
Slow DB queries I've fixed so far (not just N+1):
https://github.com/department-of-veterans-affairs/caseflow/pull/9847
https://github.com/department-of-veterans-affairs/caseflow/pull/9807

I'm focusing on the slowest queries first as reported by New Relic, and then any others I find while testing locally.

One of the slowest ones is when creating a bunch of Documents at once in `DocumentFetcher` and `AddSeriesIdToDocumentsService`. I researched a possible fix a few weeks ago and realized it requires Postgres 9.5, which then led me to notice that we are using 3 different version of Postgres locally vs Circle CI vs Production. I've been trying to schedule some time with DevoOps to figure out a plan for the upgrade since it requires downtime. Hopefully we'll get something scheduled soon.

monfresh:
Here's another one I fixed yesterday: #9967

monfresh:
Another performance improvement I worked on was replacing `active_model_serializers` with `fast_jsonapi` (#10191). This is the third-party library we use to present the JSON data that the view consumes or that other services consume (like IDT). `fast_jsonapi` was built by folks at Netflix and they have shown it to be at least 25 times faster than the gem we were using previously.

laurjpeterson:
Closing this issue in favor of (1) making this one an epic (2) opening tickets for each test that would fail if bullet was enabled, then (3) enabling bullet overall.

pkarman:
this is perennial and I think we've hit the biggest offenders according to NR (e.g. https://github.com/department-of-veterans-affairs/caseflow/commit/08e176b0ab7537e180c239551c0766c1acc7426a#diff-fd86a0d6982fc8d827021423d942eab4R57)

PR #9703, opened 2019-02-28
Add some eager loading optimizations

pkarman:
connects #9587 

Fixes a few N+1 queries reported by bullet.

### DIFF
--- a/app/services/appeal_repository.rb
+++ b/app/services/appeal_repository.rb
@@ -74,7 +74,7 @@ class AppealRepository
     cases = MetricsService.record("VACOLS: appeals_by_vbms_id",
                                   service: :vacols,
                                   name: "appeals_by_vbms_id") do
-      VACOLS::Case.where(bfcorlid: vbms_id).includes(:folder, :correspondent, :representatives)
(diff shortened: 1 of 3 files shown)


app/services/appeal_repository.rb
```diff
@@ -74,7 +74,7 @@ def self.appeals_by_vbms_id(vbms_id)
     cases = MetricsService.record("VACOLS: appeals_by_vbms_id",
                                   service: :vacols,
                                   name: "appeals_by_vbms_id") do
-      VACOLS::Case.where(bfcorlid: vbms_id).includes(:folder, :correspondent, :representatives)
+      VACOLS::Case.where(bfcorlid: vbms_id).includes(:folder, :correspondent, :representatives, :case_issues)
```
pkarman: I suspect this one may have some performance trade-offs with VACOLS, possibly good in some instances, poor in others.

monfresh: We can measure in New Relic and adjust if necessary. Also, I find that looking at the actual DB calls in the console is the best way to make sure the eager loading worked. I don't always trust Bullet.

Issue #9584, opened 2019-02-21
Bug: video reschedule selected 12:30 pm time, shows up as 1:30 pm time on daily docket.


MeredithStewart:
Tamisha rescheduled Veteran for a legacy video hearing for 12:30 pm on 4/9/19, but it's showing up as 1:30 pm. 

Something is wrong with the time as it's working on daily docket.

Update 2/26/19 - AMA cases are showing up as having 9/1 time slots. I instructed Tamisha to schedule with a note, but issue persists.

mdbenjam:
Want to make sure I understand all the correct values here.

Should there be any difference between AMA and Legacy?

And is it:
Video: 8:30, 12:30?
Central: 9, 1?

mdbenjam:
And, is this is happening for all cases? Or just the one you pointed out?

mdbenjam:
I'm seeing things like this in the fake data:
```
9:30 am ET /
8:30 am CT 
Muskogee, OK
```
Does this seem right? It should be 8:30 AM for whatever timezone the veteran is in?

MeredithStewart:
There should be no difference between AMA and legacy for time. 

Times for video and CO above are correct.

Yes, time selected should be for Veteran’s time zone. 

We were seeing this problem for hearings postponed through the daily docket. I think it was working from the regular schedule Veterans modal.

sharonwarner:
The AMA 9:00/1:00 bug was fixed in a separate PR.

PR #9713, opened 2019-02-28
Fix time on daily docket

mdbenjam:
Connects #9584

### Description
When postponing a hearing, if the user doesn't edit the time we use the `getTime` function to set the time. If the user edits the time we use the `formatTime` function. This PR updates `getTime` to use the same timezone logic as `formatTime`, as before this change we were always sending the central office's time instead of the hearing time.

### Testing Plan
1. Postpone a hearing, don't check any of the time boxes.
2. Ensure the postponed hearing is at the correct time.


client/app/hearingSchedule/containers/DailyDocketContainer.jsx
```diff
@@ -96,14 +96,14 @@ export class DailyDocketContainer extends React.Component {
...
-    const timeObject = moment(hearing.scheduledFor);
+    const timeObject = moment(hearing.scheduledFor).tz(hearing.regionalOfficeTimezone);
 
     return {
       // eslint-disable-next-line id-length
       h: timeObject.hours(),
       // eslint-disable-next-line id-length
       m: timeObject.minutes(),
-      offset: timeObject.format('Z')
+      offset: timeObject.tz('America/New_York').format('Z')
```
pkarman: hardcoding this string in the client code could be problematic.

could we make it a named constant at least so we know *why* we're using it here?

pkarman: looking at the code I see we are already harding that time zone in many places on the client side, so this doesn't make the problem worse.

I'll add that to my list of time things to fix later!

mdbenjam: Yeah, this is super not ideal :/. I think the path forward here might be when we migrate hearings from VACOLS to Caseflow. At that point we can start storing hearings in UTC and use timezones in a sane manner.

PR #9714, opened 2019-02-28
Add decision_review columns to claimants table

leikkisa:
connects #8239 

Step 1 for renaming the review_request columns in claimants to decision_review
Next will be backfilling the data and writing data to both places. Finally will be removing the review_request columns.

### DIFF
--- a/db/migrate/20190228185229_add_decision_review_columns_to_claimants.rb
+++ b/db/migrate/20190228185229_add_decision_review_columns_to_claimants.rb
@@ -1,0 +1,8 @@
+class AddDecisionReviewColumnsToClaimants < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def change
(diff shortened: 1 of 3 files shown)


db/migrate/20190228185229_add_decision_review_columns_to_claimants.rb
```diff
@@ -0,0 +1,8 @@
+class AddDecisionReviewColumnsToClaimants < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :claimants, :decision_review, index: false, polymorphic: true
```
pkarman: is the `index: false` why you must `add_index` explicitly on the next line?

leikkisa: I had it all in one line before, but the rails migration told me to do it this way.  I know that's not a very satisfying answer :-P

PR #9716, opened 2019-03-01
Remove unused RAMP columns

leikkisa:
connects #8239


pkarman:
Test failures like:

```
Error: undefined local variable or method `end_product_reference_id' for #<RampElection:0x00007fa37e6c23b0> Did you mean? end_product_code. Intake has been cancelled, please retry.
```

So some of these columns may still be in use.

Issue #9516, opened 2019-02-19
Intake | Copy change for AMA date error


pkarman:
Instead of

> Receipt Date cannot be earlier than the AMA pilot start date

it should say

> Receipt Date cannot be prior to {ama_activation_date}

PR #9718, opened 2019-03-01
Show AMA date in error if receipt date before AMA date

leikkisa:
connects #9516 

We might want to remove the feature flag altogether since AMA day has happened.


pkarman:
the feature flag is still disabled in uat/prod because we ran into timing issues. I think it will go away eventually but not until those timing issues are resolved.

Sjones352:
feature flag still disabled in UAT so until it is resolved, this is it.

screen shot 2019-03-06 at 9 11 18 pm

leikkisa:
You should still be able to validate this, it's just that the date that shows up in the error message will be further in the past. Between this and the other ticket, I'm wondering if these just aren't deploying properly to UAT.

Sjones352:
**PASSES IN UAT**
screen shot 2019-03-07 at 2 42 00 pm

PR #9720, opened 2019-03-01
Remove duplicates to shrink queries

pkarman:
I noticed large arrays of duplicate IDs in SQL queries while debugging something else.

### DIFF
--- a/app/repositories/hearing_repository.rb
+++ b/app/repositories/hearing_repository.rb
@@ -141,19 +141,20 @@ class HearingRepository
     end
 
     def hearings_for(case_hearings)
-      vacols_ids = case_hearings.map { |record| record[:hearing_pkseq] }.compact.uniq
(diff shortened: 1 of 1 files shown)


app/repositories/hearing_repository.rb
```diff
@@ -153,7 +153,7 @@ def hearings_for(case_hearings)
           .assign_or_create_from_vacols_record(vacols_record,
                                                legacy_hearing: fetched_hearings_hash[vacols_record.hearing_pkseq])
         set_vacols_values(hearing, vacols_record)
-      end.flatten
+      end.flatten.uniq
```
mdbenjam: What do you think about moving the `uniq` to line 149? `case_hearings.uniq.map` to avoid unnecessary queries in this loop?

pkarman: good call. I refactored to call `.uniq` once.

PR #9727, opened 2019-03-01
Fixing incorrect hearing times after a postponement 

### LINKED ISSUE #9584: Bug: video reschedule selected 12:30 pm time, shows up as 1:30 pm time on daily docket.

mdbenjam:
Resolves #9584

### Description
When postponing a hearing it is possible that the new day is in DST while the previously scheduled day was not. This means we might save the date with an incorrect offset. To fix that we need to used the new date to determine the offset.

This is a temporary fix as this component's logic needs significant rethinking.

Also, I think there is a still a potential bug with postponing AMA cases, but more investigation needs to be done. But it appears that changing the time when postponing an AMA hearing does not work. 

### Testing Plan
1. Manually change a hearing from a date before DST to after and ensure the time is still correct.


client/app/hearingSchedule/containers/DailyDocketContainer.jsx
```diff
@@ -92,8 +92,10 @@ export class DailyDocketContainer extends React.Component {
   };
 
   getTime = (hearing) => {
+    const dateForTimezoneCalculation = (hearing.editedDate && hearing.editedDate.scheduledFor) || hearing.scheduledFor;
```
mdbenjam: We need to use the edited date for the timezone calculation since it might be in DST while the current date is not.

PR #9732, opened 2019-03-01
Add Danger rule to detect disabling Rubocop rules

monfresh:
**Why**: We use Rubocop to enforce code quality, and the settings we use
have been agreed upon by the team. If we disable them without a valid
reason, then it defeats the purpose of using a tool like Rubocop.

We are currently disabling in several places important rules that
measure code complexity. This means that those classes or methods can
keep getting more complex since we're not checking their complexity
anymore.

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -12,7 +12,16 @@ if `git diff #{github.base_commit} spec/ | grep -E '(:focus => true)|(focus: tru
   fail("focus: true is left in test")
 end
 
-# We must take care of our VACOLS models.  Remind developers to test this thoroughly
+# We must take care of our VACOLS models. Remind developers to test this thoroughly
 if !git.modified_files.grep(/app\/models\/vacols/).empty?
-  warn("This PR changes VACOLS models.  Please ensure this is tested against a UAT VACOLS instance")
+  warn("This PR changes VACOLS models. Please ensure this is tested against a UAT VACOLS instance")
+end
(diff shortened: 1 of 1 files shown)


Dangerfile
```diff
@@ -12,7 +12,16 @@ if `git diff #{github.base_commit} spec/ | grep -E '(:focus => true)|(focus: tru
...
-  warn("This PR changes VACOLS models.  Please ensure this is tested against a UAT VACOLS instance")
+  warn("This PR changes VACOLS models. Please ensure this is tested against a UAT VACOLS instance")
+end
+
+# We should not disable Rubocop rules unless there's a very good reason
+if `git diff #{github.base_commit} | grep -E 'rubocop:disable'`.length > 1
+  warn(
+    "This PR disables one or more Rubocop rules. " \
+    "If there is a valid reason, please provide it in your commit message. " \
+    "Otherwise, consider refactoring the code."
```
pkarman: good language

app/controllers/application_controller.rb
```diff
@@ -29,6 +29,7 @@ class ApplicationController < ApplicationBaseController
 
   private
 
+  # rubocop:disable Metrics/MethodLength
```
pkarman: is this here just to test the rule?

monfresh: Yep!

Issue #9389, opened 2019-02-13
Allow VSO users to update TrackVeteranTask status on backend


lowellrex:
When a VSO user clicks into a case details page from the "All cases" tab of their VSO's organization queue, we get an `Caseflow::Error::ActionForbiddenError` (like [this one we received today](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3678/)). The user will still be able to visit the case details because [we don't care about the request's result](https://github.com/department-of-veterans-affairs/caseflow/blob/fb478a7d33a9fb14b1825ed20344804c100171b9/client/app/queue/CaseDetailsLink.jsx#L28), but we shouldn't throw raise this error in the first place. This ticket exists to address that issue.

We are raising the `Caseflow::Error::ActionForbiddenError` because we are attempting to update the status of a `TrackVeteranTask`, and since that class does not have [any available actions](https://github.com/department-of-veterans-affairs/caseflow/blob/fb478a7d33a9fb14b1825ed20344804c100171b9/app/models/tasks/track_veteran_task.rb#L2) the [`can_be_updated_by_user ?` check](https://github.com/department-of-veterans-affairs/caseflow/blob/fb478a7d33a9fb14b1825ed20344804c100171b9/app/models/task.rb#L180) fails. We should probably override that method in the `TrackVeteranTask` class to allow all members of the organization to update the status.

Alternatively, we could just always create `TrackVeteranTask`s with the status of `in_progress` instead of the default `assigned` (either in [the `RootTask` creation routine](https://github.com/department-of-veterans-affairs/caseflow/blob/6b2c20030790090b62eb0711acdfcfcf59d97820/app/models/tasks/root_task.rb#L71) or as an `after_initialize` callback on [the `TrackVeteranTask` itself](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/tasks/track_veteran_task.rb)).

## Acceptance criteria
* [ ] Allow members of VSOs to update the status of `TrackVeteranTask`s assigned to their VSO
* [ ] Do not throw `Caseflow::Error::ActionForbiddenError` when a member of a VSO visits a case details page from their team's queue

laurjpeterson:
[...]

PR #9735, opened 2019-03-01
Fix TrackVeteranTask access bug

lowellrex:
Resolves `Caseflow::Error::ActionForbiddenError` bugs (like [this one](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3678/)) that happen because of the same access issues mentioned in #9389 and initially fixed in #9602. This issue popped back up because we started creating `TrackVeteranTask`s [in an additional place](https://github.com/department-of-veterans-affairs/caseflow/blob/8b658970f3b563d3346bae4a096cdce728b6bc71/app/models/appeal.rb#L213) but forgot to set the task status appropriately there.

This PR moves that initial fix into the `TrackVeteranTask` so that we correctly set the task's `status` to `in_progress` every time we create a `TrackVeteranTask`.


app/models/tasks/track_veteran_task.rb
```diff
@@ -5,6 +5,13 @@
 #   - the power of attorney changes on an appeal
 
 class TrackVeteranTask < GenericTask
+  # Avoid permissions errors outlined in Github ticket #9389 by setting status here.
+  after_initialize :set_in_progress_status
+
+  def set_in_progress_status
+    self.status = Constants.TASK_STATUSES.in_progress
+  end
```
tomas-nava: Maybe make `set_in_progress_status` a private method? Can't think of any reason we'd want to use it instead of an `update!` except by a hook.

Also maybe use `update!` here instead of equals assignment?

lowellrex: I think we need to use the assignment operator here because we've only instantiated the object, and haven't yet saved it to the database. Changing this to `update!` I run into the following errors:

```ruby
     Failure/Error: update!(status: Constants.TASK_STATUSES.in_progress)
     
     ActiveRecord::RecordInvalid:
       Validation failed: Assigned to can't be blank, Appeal can't be blank
```

tomas-nava: 👍 ah, duh, I should've realized.

PR #9745, opened 2019-03-02
Add comments to RAMP Election Rollback table

leikkisa:
connects #9559


db/migrate/20190302004820_add_comments_to_ramp_election_rollbacks.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    change_column_comment(:ramp_election_rollbacks, :created_at, "Timestamp for when the rollback was created.")
+
+    change_column_comment(:ramp_election_rollbacks, :ramp_election_id, "The ID of the RAMP Election being rolled back.")
+
+    change_column_comment(:ramp_election_rollbacks, :reason, "The reason for rolling back the RAMP Election. Rollbacks happen automatically for canceled RAMP Election End Products, but can also happen for other reason such as by request.")
+
+    change_column_comment(:ramp_election_rollbacks, :reopened_vacols_ids, "The IDs of any legacy appeals which were reopened as a result of rolling back the RAMP Election, corresponding to the VACOLS BFKEY.")
+
+    change_column_comment(:ramp_election_rollbacks, :updated_at, "Timestamp for when the rollback was last updated.")
```
leikkisa: Don't know if we're really using this column.

pkarman: Rails will autopopulate it. IMO, all active models should have `:created_at` and `:updated_at` because it makes debugging much easier.

PR #9749, opened 2019-03-04
Add default label for all tasks

lowellrex:
Related to [this slack discussion](https://dsva.slack.com/archives/C6E41RE92/p1551712604075100) where we were made aware that some tasks do not have labels in the task snapshot.

Before | After
--- | ---
image | image


app/models/task.rb
```diff
@@ -37,7 +37,7 @@ def available_actions(_user)
   end
 
   def label
-    action
+    self.class.name
```
lowellrex: The only task type that still uses the `action` field is `ColocatedTask`. Lots of other tasks [define a more precise `label`](https://github.com/department-of-veterans-affairs/caseflow/search?q=%22def+label%22), but in case they don't this changes the fall-back to the class name instead of the `action` field which will probably be `null`.

PR #9759, opened 2019-03-04
8853 display data source for veteran address

### LINKED ISSUE #8853: Display data source for Veteran address in Case details
We return Veteran address from BGS for AMA and legacy appeals (https://github.com/department-of-veterans-affairs/caseflow/issues/7426). Let's display this for our users, just like we do for POA address. 

Reference: mocks for POA address data source here - https://github.com/department-of-veterans-affairs/caseflow/issues/6812

## AC
- Data source for Veteran address shown in Case details, about the Veteran section, for both legacy and AMA appeals

youngfreezyVA:
Resolves #8853

### Description
displays data source for veteran address depending on if its legacy appeal or not


client/COPY.json
```diff
@@ -83,6 +83,8 @@
   "CASE_DETAILS_HEARING_ON_OTHER_APPEAL_LINK": "View all cases",
   "CASE_DETAILS_HEARING_ON_OTHER_APPEAL_POST_LINK": " to see other cases associated with this Veteran.",
   "CASE_DETAILS_INCORRECT_POA": "POA data comes from VBMS. To update POA, please send a request to the VLJ support staff.",
+  "CASE_DETAILS_INCORRECT_VETERAN_ADDRESS_LINE1": "Veteran information comes from ", 
+  "CASE_DETAILS_INCORRECT_VETERAN_ADDRESS_LINE2": ". To update the veteran's information, please send a request to the VLJ support staff.",
```
tomas-nava: You can consolidate these into a single variable and interpolate the value you want, like this:

```json
"CASE_DETAILS_INCORRECT_VETERAN_ADDRESS": "Veteran information comes from %s. To update the veteran's information, please send a request to the VLJ support staff.",
```

and then:

```js
sprintf(COPY.CASE_DETAILS_INCORRECT_VETERAN_ADDRESS, this.props.isLegacyAppeal ? legacyDataSource : amaDataSource)
```

tomas-nava: Also, I don't think `CASE_DETAILS_INCORRECT_VETERAN_ADDRESS` is an accurate name for this bit of copy; something more like `CASE_DETAILS_VETERAN_ADDRESS_SOURCE`?

youngfreezyVA:
> We should test this, maybe in `spec/feature/queue/case_details_spec.rb`; also, see review comments.

all set @tomas-nava!

youngfreezyVA:
> Sorry to do this to you, but I realized when reading through the original issue (and confirmed with @lowellrex) that the message should actually be the same for both types of appeals:
> 
> ```
> "Veteran information comes from VBMS. To update the veteran's information, please send a request to the VLJ support staff."
> ```
> So you can remove the parts that display a different message depending on what type of appeal it is, and just show the above message every time.

done! @tomas-nava

PR #9766, opened 2019-03-04
Fix flakey rake script tests

tomas-nava:
### Description
Tests of the tasks/tasks rake script were failing because the rake tasks were being included and running multiple times. This PR fixes the way rake tasks are included in the spec and un-skips the previously flakey tests.

The flake was 100% reproducible by running the following command:

```bash
bundle exec rspec spec/feature/log_in_as_user_spec.rb spec/lib/tasks/tasks_spec.rb
```

I confirmed locally and on Circle that after this change, the tests no longer fail.


spec/lib/tasks/tasks_spec.rb
```diff
@@ -2,9 +2,10 @@
 require "rake"
 
 describe "task rake tasks" do
-  before do
+  before :all do
```
lowellrex: `before()` defaults to `before(:each)`, correct? Was the reason this test was failing consistently because we were running this block before every test block? Or is the fix to the problem what we added on line 6, and this change was just made to speed things up so we don't reset the environment for every test block.

tomas-nava: Correct, `:each` is the default. The test was failing because I wasn't creating a fresh Rake application context, and I believe that change alone would've been enough to make the tests pass. But while I was at it I also added the `:all` parameter to `before` because it really only needs to run once.

lowellrex: Excellent! Makes sense to me!

PR #9788, opened 2019-03-05
Fix AppealsController DoubleRenderError

lowellrex:
Fixes [`AbstractController::DoubleRenderError`s we've been seeing](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3663/) coming from `AppealsController.index`


app/controllers/appeals_controller.rb
```diff
@@ -11,12 +11,20 @@ def index
...
+            claim_reviews: ClaimReview.find_all_by_file_number(veteran_file_number).map(&:search_table_ui_hash)
+          }
+        else
+          render json: {
+            "errors": [
+              "title": "Must include Veteran ID",
+              "detail": "Veteran ID should be included as HTTP_VETERAN_ID element of request headers"
+            ]
+          }, status: :bad_request
+        end
```
lowellrex: Inlined this response since this is the only place we use it.

app/controllers/appeals_controller.rb
```diff
@@ -184,7 +184,10 @@ def get_appeals_for_file_number(file_number)
   end
 
   def get_vso_appeals_for_file_number(file_number)
-    return file_access_prohibited_error if !BGSService.new.can_access?(file_number)
+    if !BGSService.new.can_access?(file_number)
+      msg = "User is prohibited from accessing files associated with provided Veteran ID"
+      fail(Caseflow::Error::ActionForbiddenError, message: msg)
+    end
```
monfresh: The Ruby Style Guide [discourages using exceptions for flow of control](https://github.com/rubocop-hq/ruby-style-guide#no-exceptional-flows). This is one of the practices I'm proposing we stop following in this codebase. Exceptions are slow in Ruby, and it's not clear at first glance what will happen after raising this error if you're not already familiar with how exceptions are raised and then rescued in `ApplicationController`.

This is why I'm advocating for a cleaner approach where the validations are performed by a separate class which can then pass on any errors to the controller without using exceptions.

To the end user, this fix does not make a difference because  the frontend currently doesn't use the messages rendered by the controller. It treats all backend errors the same and displays a generic `Server encountered an error searching for {file_number}. Please retry your search`. I improved this user experience in my PR by having the frontend read the errors from the backend and display them to the user.

From what I've seen, this error doesn't happen very often, and it doesn't affect users since it only happens when a VSO employee is trying to access a file that they don't have access to. Since this fix doesn't change the user experience, what do you think about waiting until after my Wisdom Wednesday talk?

Issue #9709, opened 2019-02-28
Allow Intakes for Veterans with undifferentiated gender


leikkisa:
First VBMS needs to allow Veterans with a nil sex when establishing a claim.  Right now the VBMS validation requires that the value be in ['M','F'].

Nate says we can't pass an empty string, but we can pass "undifferentiated", so the first step might be testing this with 'U'.

Then we can remove the validation for it when searching for a Veteran at the beginning of intake.

leikkisa:
I'm going to pair with Sandra on this tomorrow.

leikkisa:
So if the gender is undifferentiated, we don't want to pass a gender attribute at all in the claim_hash in connect_vbms (meaning we should make it an optional argument).

Sjones352:
@leikkisa can you please put instructions on how I can validate this ticket.

leikkisa:
@Sjones352 

I don't know if you have the ability to delete the gender from a Veteran in another system, but that would be the easiest way to do it. And then complete a Higher Level Review or Supplemental Claim for them, with Compensation as the business line.  Then we'll have to check the data to see if the claim establishment succeeds.

Other than that, we'll need to find a Veteran that has the other required information (primarily address), but is missing a gender. In my experience, if they're missing one they're missing everything so that might be tough.

Sjones352:
Slacked Colin to get create me a veteran without a gender so I will validate once that is done.

Sjones352:
@leikkisa looks like passing 'undifferentiated' works because an EP was created.  Please see screen-shots below.
<img width="1191" alt="Screen Shot 2019-03-14 at 12 16 40 PM" src="https://user-images.githubusercontent.com/23080951/54373180-32cc7200-4653-11e9-97be-779a4363fde2.png">

<img width="1096" alt="Screen Shot 2019-03-14 at 12 08 22 PM" src="https://user-images.githubusercontent.com/23080951/54374403-6f996880-4655-11e9-8a73-18bc0d5eafd9.png">
<img width="1083" alt="Screen Shot 2019-03-14 at 12 08 40 PM" src="https://user-images.githubusercontent.com/23080951/54374405-6f996880-4655-11e9-9d1f-782f27864ac3.png">
<img width="1081" alt="Screen Shot 2019-03-14 at 12 14 22 PM" src="https://user-images.githubusercontent.com/23080951/54374406-7031ff00-4655-11e9-92e1-88d88eee1853.png">

[...]

PR #9792, opened 2019-03-05
Remove gender validation on Veteran for Intake

leikkisa:
connects #9709


leikkisa:
Wait for connect_vbms fix, and update gem before merging this

PR #9797, opened 2019-03-05
Add task labels to Hearings tasks

### LINKED ISSUE #9750: Hearing tasks | Add labels for Case Details page
timeline

The timeline on the case details page displays the name of the task unless a label is explicitly defined. We should define labels for the tasks we've created for hearings.

oscarAtNava:
Resolves #9750 

### Description
Added label property to each hearing task class and set it to verbiage defined in ticket.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to Case Details and confirm you see the new label text.


app/models/tasks/hearing_task.rb
```diff
@@ -5,4 +5,8 @@
 
 class HearingTask < GenericTask
   has_one :hearing_task_association
+
+  def label
+    "All hearing-related tasks"
```
sharonwarner: I'm not sure if this label will make sense to a user. Maybe something like "A hearing request is being processed." would make more sense.

app/models/tasks/disposition_task.rb
```diff
@@ -18,6 +18,10 @@ def create_disposition_task!(appeal, parent, hearing)
     end
   end
 
+  def label
+    "Hearing disposition to be selected"
```
sharonwarner: Maybe this label should be dependent on the hearing date. Before the hearing date, it could be "A hearing is scheduled", after the hearing date it could be "Select a hearing disposition."

app/models/tasks/hearing_task.rb
oscarAtNava: These are per Meredith in the issue ticket. Let's discuss in stand-up tomorrow.

PR #9801, opened 2019-03-05
Use uniqueId over taskID

### LINKED ISSUE #9748: Use task.uniqueId as the ID for the NewFile component
We are returning a bundle of ["non-critical errors"](https://app.datadoghq.com/dashboard/3v6-zzc-s72/non-critical-caseflow-application-errors) from the `tasks_new_documents` endpoint and [a kibana search reveals](https://caseflow-kibana.va.gov/_plugin/kibana/goto/d6e3494c9336824bbd8c794473c73f08) that we are making requests to the `TasksController#new_documents` endpoint with a `null` task ID. I think we can use the `task.uniqueId` property instead of the `task.taskId` property [since the latter can sometimes be `null`](https://github.com/department-of-veterans-affairs/caseflow/blob/b1e2358cd062d38eb6b1e0f83bc3ce8cf3aecadd/app/models/legacy_tasks/legacy_task.rb#L102).

We determined that `null` `taskId`s were the root cause of the `null` ID requests we were making by hopping into a production Rails console and running of the following commands:

```ruby
# Got the list of tasks for a user experiencing this problem.
rails c> u = User.find(889)
rails c> tasks = LegacyWorkQueue.tasks_for_user(u)
rails c> tasks.count
# 25

rails c> legacy_tasks = tasks.select { |t| t.is_a?(LegacyTask) }
rails c> legacy_tasks.count
# 25

# Confirmed that some of the tasks could have null task IDs.
rails c> tasks_with_nil_ids = tasks.select { |t| t.task_id.nil? }
rails c> tasks_with_nil_ids.count
# 12

# Checked to see if the alternative (the unique ID field) would solve this problem.
# Task.appeal.vacols_id is used in the LegacyTaskSerializer to populate the external_appeal_id field.
# https://github.com/department-of-veterans-affairs/caseflow/blob/b1e2358cd062d38eb6b1e0f83bc3ce8cf3aecadd/app/models/serializers/work_queue/legacy_task_serializer.rb#L52
rails c> external_ids = tasks.map { |t| t.appeal.vacols_id }
rails c> external_ids.compact.count
# 25
```

## Acceptance criteria
* [ ] We are no longer making requests from the frontend to `/tasks/null/new_documents`

hschallhorn:
Resolves #9748

See #9748 for details why

Also stopped using `.exists` as it was incorrectly returning true


aroltsch:
I chatted with Hunter and we should close this PR. We should move Assign cases page and Unassigned tab for org queues to use appeals_controller instead of using the `task` because we don't need placed_on_hold_at for those views

Issue #9729, opened 2019-03-01
Intake Edit | Show indicator for user to save after changes are made


shanear:
On the intake edit page, changes aren't saved immediately. We need to notify users to save their changes.

## Acceptance Criteria
- On the intake edit page for any decision review, if an issue is added or removed, then a blue banner should show up informing the user to save their changes (see the mock).

## Mocks
![edit issues - withdrawn](https://user-images.githubusercontent.com/1596259/53833038-7f6dca00-3f3c-11e9-8b2c-3aa06d459f63.png)

## Notes
- This will not affect the initial intake page, only the edit page

Sjones352:
**UAT**

**Before I remove decision**
<img width="1177" alt="Screen Shot 2019-03-12 at 1 11 55 PM" src="https://user-images.githubusercontent.com/23080951/54221030-084ead80-44c9-11e9-882c-2870f4d7c75d.png">
**Modal showing Remove issue**
<img width="1082" alt="Screen Shot 2019-03-12 at 1 12 16 PM" src="https://user-images.githubusercontent.com/23080951/54221031-08e74400-44c9-11e9-9866-70b8dc2c21cc.png">
**Message banner**
<img width="1108" alt="Screen Shot 2019-03-12 at 1 12 28 PM" src="https://user-images.githubusercontent.com/23080951/54221032-08e74400-44c9-11e9-849c-c1ccd7fb37d4.png">
**Click Save and modal shows**
<img width="1102" alt="Screen Shot 2019-03-12 at 1 14 01 PM" src="https://user-images.githubusercontent.com/23080951/54221033-08e74400-44c9-11e9-91c9-a0171f3bee39.png">
**Once saved**
<img width="962" alt="Screen Shot 2019-03-12 at 1 14 29 PM" src="https://user-images.githubusercontent.com/23080951/54221034-097fda80-44c9-11e9-9bf6-72080fa33424.png">

PR #9806, opened 2019-03-06
adding flash to edit page

jiahuang:
Connects #9729 

image

Validation:
* create intake, when adding issues verify that banner does not show up
* edit intake, verify that when modifying issues the banner shows up


client/app/intake/pages/addIssues.jsx
```diff
@@ -171,6 +206,7 @@ export class AddIssuesPage extends React.Component {
       <Table
         columns={columns}
         rowObjects={rowObjects}
+        rowClassNames={issueChangeClassname}
```
leikkisa: If you set the value of haveIssuesChanged to a variable you could use that on line 161 and put the ternary with the classname here instead of making the issueChangeClassname function, which seems a bit more complicated.

client/app/intake/pages/addIssues.jsx
```diff
@@ -30,11 +30,32 @@ export class AddIssuesPage extends React.Component {
...
       issueRemoveIndex: 0
     };
   }
 
+  getFormName = (formType) => {
+    return _.find(FORM_TYPES, { key: formType }).key;
+  }
+
+  haveIssuesChanged = (currentIssues) => {
+    if (currentIssues.length !== this.state.originalIssueIds.length) {
```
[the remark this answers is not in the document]
jiahuang: this wouldn't return true, but the next part will

client/app/intake/pages/addIssues.jsx
jiahuang: rowClassNames has to be a function due to `rowClassNames: PropTypes.func,` in Table.jsx

PR #9823, opened 2019-03-07
Add cancel method to DispositionTask

### LINKED ISSUE #9539: Add application logic when cancelling DispositionTask
The `DispositionTask` model (previously `HoldHearingTask`) needs to have a cancel method. This method should perform the following actions:

1) Confirm the task's associated hearing's disposition is 'canceled'. This hearing can be found through the `DispositionTasks`'s parent (a `HearingTask`) which will have a `HearingTaskAssociation`.
1) If the appeal is a legacy appeal:
    - We cancel the `DispositionTask`
    - The `HearingTask` should automatically close
    - We move the appeal to location 81 if there's no associated VSO and 55 if there is an associated VSO (we already have similar logic in the withdraw_hearing method in `ScheduleHearingTask`)
1) If the appeal is an AMA appeal:
    - We cancel the `DispositionTask`
    - The `HearingTask` should automatically close
    - If the appeal has a VSO, we create an IHP task

For more information, see here: https://github.com/department-of-veterans-affairs/appeals-design-research/blob/master/Projects/Hearings/Hearing%20Schedule/Tech%20Specs/HearingTaskFlow.md

## Acceptance criteria
* [ ] Add method to cancel `DispositionTask`s during job that will run every night
* [ ] Cancelling `DispositionTask`s closes parent `HearingTask` and returns `LegacyAppeal` to location 81 or 55
* [ ] Cancelling `DispositionTask`s closes parent `HearingTask` and maybe creates an IHP task

tomas-nava:
Resolves #9539

### Description
Adds a `cancel!` method to the `DispositionTask` that makes required changes on parent `HearingTask` and associated legacy/ama appeal.


app/models/tasks/disposition_task.rb
```diff
@@ -29,4 +31,25 @@ def check_parent_type
...
+    update!(status: Constants.TASK_STATUSES.cancelled)
+
+    if appeal.is_a?(LegacyAppeal)
+      location = if appeal.vsos.empty?
+                   LegacyAppeal::LOCATION_CODES[:case_storage]
+                 else
+                   LegacyAppeal::LOCATION_CODES[:service_organization]
+                 end
+
+      AppealRepository.withdraw_hearing!(appeal)
```
tomas-nava: Is it necessary to withdraw the hearing here?

app/models/tasks/hearing_task.rb
```diff
@@ -7,4 +7,16 @@
...
   has_one :hearing_task_association
+
+  private
+
+  def update_status_if_children_tasks_are_complete
+    unique_child_types = children&.map(&:type)&.uniq || []
+    if unique_child_types.count == 1 && unique_child_types.first == "DispositionTask"
+      unique_child_status = children&.map(&:status)&.uniq&.first
+      return update!(status: :cancelled) if unique_child_status == Constants.TASK_STATUSES.cancelled
+    end
```
tomas-nava: Does this logic make sense? It will cancel the `HearingTask` if its children are only cancelled `DispositionTask`s

app/models/tasks/disposition_task.rb
sharonwarner: It's not, in fact, we don't want to. Withdrawing a hearing means no hearing was ever scheduled. These cases are handled differently in VACOLS reporting. The VACOLS reporting logic for cancelled hearings is handled in VACOLS::CaseHearing: `after_update :update_hearing_action, if: :saved_change_to_hearing_disp?`, so we don't need to do anything here.

app/models/tasks/hearing_task.rb
sharonwarner: In most cases, the HearingTask will already have a completed child ScheduleHearingTask if a DispositionTask exists. A HearingTask will also never have more than one DispositionTask, so I think we can simplify this logic a little bit. There's a diagram that might be helpful in this doc: https://github.com/department-of-veterans-affairs/appeals-design-research/blob/master/Projects/Hearings/Hearing%20Schedule/Tech%20Specs/HearingTaskFlow.md

app/models/tasks/disposition_task.rb
tomas-nava: Addressed this in [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/9823/commits/cdf6a2b28e84a07f34c8e12c5eb3e60485a6402b)

app/models/tasks/hearing_task.rb
tomas-nava: Addressed this in [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/9823/commits/f17075848da19dce366a4986c0b72a56c4d25dc6); the HearingTask is now canceled if it has no active children and one of those children is a canceled DispositionTask.

Issue #9543, opened 2019-02-20
Create NoShowHearingTask


sharonwarner:
NoShowHearingTasks should have three available actions: the hearing can be rescheduled for a later date, the appeal can be released for case distribution, or the task can be put on hold. 

1) If the appeal is being rescheduled:
    - We complete the DispositionTask 
    - The HearingTask should automatically close
    - We create a new HearingTask off of the original HearingTask's parent
    - We create a new ScheduleHearingTask off of the new HearingTask
1) If the appeal is being released:
    - For an AMA appeal:
        - We close the NoShowHearingTask
        - The DispositionTask and HearingTask should automatically close
        - If the appeal has a VSO, we create a IHP task
    - For a legacy appeal:
        - We close the NoShowHearingTask
        - The DispositionTask and HearingTask should automatically close
        - We move the appeal to location 81 if there is no associated VSO and 55 if there is an associated VSO

### Bite-sized pieces of work:
*To be combined as seems appropriate*
* [x] Create `NoShowHearingTask` class (#9819)
* [x] Add back-end logic for `NoShowHearingTask` class to reschedule a hearing (#9824)
* [x] Add front-end task action option and tests for reschedule hearing action (#9824)
* [x] Add back-end logic for `NoShowHearingTask` class to release an AMA appeal for case distribution (#9905)
* [x] Add back-end logic for `NoShowHearingTask` class to release a Legacy Appeal for case distribution (#9905)
* [x] Add front-end task action option and tests for releasing appeal for case distribution (#9905)
* [ ] Allow `NoShowHearingTask`s to be put on hold (will be completed as part of #9207)

<img width="1075" alt="screen shot 2019-02-20 at 12 10 36 pm" src="https://user-images.githubusercontent.com/4306128/53110253-94903500-3508-11e9-8d5e-f7572e84db91.png">

<img width="810" alt="screen shot 2019-02-20 at 12 11 03 pm" src="https://user-images.githubusercontent.com/4306128/53110287-9f4aca00-3508-11e9-8fa7-969842258f45.png">

Postponement modal work moved to #9614.

For more information, see here: https://github.com/department-of-veterans-affairs/appeals-design-research/blob/master/Projects/Hearings/Hearing%20Schedule/Tech%20Specs/HearingTaskFlow.md

lowellrex:
> For now, we should only allow the user to select 'Send to Schedule Veteran List', we will add more postponement functionality later
[...]

PR #9824, opened 2019-03-07
Allow NoShowHearingTasks to be rescheduled/postponed

lowellrex:
Connects #9543.

This PR implements the logic to reschedule a hearing for an appeal by completing the parent `DispositionTask` and grandparent `HearingTask` and creating a new `HearingTask` and child `ScheduleHearingTask`.

We expect the state of the task tree before calling `no_show_hearing_task.reschedule_hearing()` to look like the following diagram:

```
                   +----------+
                   |          |
                   | RootTask |
                   |          |
                   +----+-----+
                        |
                        |
                        |
              +---------+--------+
              |                  |
              | DistributionTask |
              |                  |
              +---------+--------+
                        |
                        |
                        |
                 +------+------+
                 |             |
                 | HearingTask |
                 |             |
                 +------+------+
                        |
           +------------+----------+
           |                       |
+----------+----------+   +--------+--------+
|                     |   |                 |
| ScheduleHearingTask |   | DispositionTask |
|                     |   |                 |
+---------------------+   +--------+--------+
                                   |
                                   |
                                   |
                         +---------+---------+
                         |                   |
                         | NoShowHearingTask |
                         |                   |
                         +-------------------+

```

reschedule_hearing


app/models/tasks/hearing_task.rb
```diff
@@ -7,4 +7,11 @@
 
 class HearingTask < GenericTask
   has_one :hearing_task_association
+  before_validation :set_assignee
```
lowellrex: Everywhere we were creating `HearingTask`s we were setting the assignee to `Bva`. This change makes it so we no longer have to do that.

app/models/tasks/no_show_hearing_task.rb
```diff
@@ -3,4 +3,16 @@
 # Task created after an appellant no-shows for a hearing. Gives the hearings team the options to decide how to handle
 # the no-show hearing after the judge indicates that the appellant no-showed.
 class NoShowHearingTask < GenericTask
+  before_validation :set_assignee
+
+  def reschedule_hearing
+    update!(status: Constants.TASK_STATUSES.completed)
+    RootTask.create_hearing_schedule_task!(appeal, root_task)
+  end
```
lowellrex: Meat of the change is here.

app/models/tasks/no_show_hearing_task.rb
```diff
@@ -3,4 +3,16 @@
 # Task created after an appellant no-shows for a hearing. Gives the hearings team the options to decide how to handle
 # the no-show hearing after the judge indicates that the appellant no-showed.
 class NoShowHearingTask < GenericTask
+  before_validation :set_assignee
+
+  def reschedule_hearing
+    update!(status: Constants.TASK_STATUSES.completed)
```
lowellrex: Status changes will cascade up through the parents because they are assigned to organizations.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -6,6 +6,7 @@
 # Once completed, a DispositionTask is created.
 
 class ScheduleHearingTask < GenericTask
+  before_validation :set_assignee
```
lowellrex: Everywhere we were creating `ScheduleHearingTask`s we were setting the assignee to `HearingsManagement`. This change makes it so we no longer have to do that.

app/controllers/tasks_controller.rb
```diff
@@ -124,6 +124,18 @@ def ready_for_hearing_schedule
     }
   end
 
+  def reschedule
```
lowellrex: I don't love this design but it allows us to make this change quickly (without making the `data` functions into classes, for instance). Super open to considering alternative approaches here. I think we can do much better.

In summary:
* `NoShowHearingTask` has an action that sends us to a dedicated modal that calls this endpoint
* `tasks/:id/reschedule` effectively calls `NoShowHearingTask.find(:id). reschedule_hearing()`
* `reschedule_hearing()` 1) creates the new `ScheduleHearingTask` and 2) closes the existing branch of the task tree that has the hearings-related tasks

If we were to do only one of those operations (new task or closing existing) then we could easily fit this into one of the existing `TasksController` methods (`new` or `update`).

We could use `TasksController.update` and have special logic in `NoShowHearingTask.update_from_params()` to create the `ScheduleHearingTask` (this is [what we do for `TranscriptionTask`](https://github.com/department-of-veterans-affairs/caseflow/blob/c6195ae99676a634b382e5d2110179b76564bcd9/app/models/tasks/transcription_task.rb#L36), though it is not yet in use). My only fear of going with this approach is that we overload what it means for a `NoShowHearing` to be cancelled.

We could also use `TasksController.new` and pass in a new `ScheduleHearingTask`. We would have to add logic in `ScheduleHearingTask.update_params()` to set the parent task to the root task of the appeal (instead of the `NoShowHearingTask` which it would default to). In addition we would have to add a `before_create` callback to `HearingTask` to close the other branch or we would throw an `DuplicateOrgTask` error when we create it. The approach is very clean (don't need any new controller logic, can re-use `AssignToView`) but it feels a shade too magical for me to be 100% comfortable. This would rely on us overriding a core `Task` method in one class, a callback in a different class, and both of these things would be triggered by an action taken on a third type of task.

I am still on the fence about whether the approach in this PR is better than the one described above. Curious to get your thoughts.

lowellrex: > We should at least be doing actions like this consistently across our controllers.

Yes yes yes! And I really like the approach outlined in the blog post you link... though I think we should wait to do it in a different PR.

PR #9828, opened 2019-03-07
Case flow details function

### LINKED ISSUE #9235: Configure Case Details function
We are adding a CSEM/CSUM function called Caseflow Case Details, so users can have access to search for cases in Caseflow and view their case details. (For context, another example of a CSEM/CSUM function is `Reader` or `HearingPrep`.) They should *not* be able to access to Caseflow Reader. The primary users for this function are VBA call center/customer service folks, so they can answer Veteran questions, but there will likely be more users that eventually use this scoped down version of Caseflow.

Martina is in the process of requesting the Caseflow Case Details function - https://github.com/department-of-veterans-affairs/appeals-pm/issues/2371

## AC
We need to ensure the following criteria are met, when a user has access to Caseflow Case Details:
- User can search for cases at `/search`
- User navigates to `/search` when they click the Caseflow logo (note: this can be done by the `case_search_homepage` feature toggle if need be)
- When user clicks on docket number link, they navigate to the case details page for that case
- `Veteran documents` link does not show for these users, and users cannot access `/reader/appeal`
- Users cannot navigate to an individual or team queue

youngfreezyVA:
Resolves #9235

### Description
progress pr for caseflow case details function

### Acceptance Criteria
see:
https://github.com/department-of-veterans-affairs/caseflow/issues/9235

### Testing Plan
1. Go to ...


spec/feature/queue/ama_queue_spec.rb
```diff
@@ -6,7 +6,17 @@
...
+    let(:no_queue_user) { FactoryBot.create(:user, roles: ["Case Details"]) }
+    it "should not be able to access queue" do
+      step "case details role tries to access queue" do
+        User.authenticate!(user: no_queue_user)
+        visit "/queue"
+        expect(page).to_not have_content("Queue")
+        expect(current_path).to eq "/search"
+      end
+    end
+  end
```
tomas-nava: See my previous comment about testing for the presence of something rather than the absence of something.

I would expand to also test for:
  - clicking the caseflow logo takes the user to `/search`
  - the user can search for a case, click the docket number link, and view the case details page
  - the veteran documents link on the case details page isn't rendered
  - (maybe) the user can't access reader
  - (maybe) the user can't navigate to team queue

app/views/queue/index.html.erb
```diff
@@ -12,6 +12,7 @@
     applicationUrls: application_urls,
     feedbackUrl: feedback_url,
     buildDate: build_date,
+    roles: current_user.roles,
```
tomas-nava: Although passing all the user's roles gives you a lot more flexibility and room to expand, it feels brittle for this feature; for example, we define and check for the `"Case Details"` role on the backend in `QueueController#verify_access` and on the frontend in `utils.hasRestrictedQueueRole()`.

What if we just checked the role on the backend and passed a boolean (named something like `hasCaseDetailsRole`?) to the frontend instead?

youngfreezyVA: what if the front end needs to hide/show additional features based on roles in the future... wouldn't it be easier for the front end to have its own source of truth for restricted roles?

tomas-nava:
Approved for the changes I requested, but would appreciate a second set of eyes before it's merged.

app/views/queue/index.html.erb
youngfreezyVA: @aroltsch @lowellrex please let me know if you have any thoughts before I mark this convo as resolved and add the `ready-to-merge` label

lowellrex: I'm in agreement with Tomas here. We should be able to restrict what pages they can view by checking the user role when they make the request to the back-end for whatever asset we would otherwise restrict on the front-end.

youngfreezyVA: understood - made the necessary changes. @lowellrex @tomas-nava

PR #9829, opened 2019-03-07
Automatically create HearingTask as parent of all new ScheduleHearingTasks

lowellrex:
Connects #9543. Prepares for us to add logic on the front-end that creates a new `ScheduleHearingTask`, automatically creates the parent `HearingTask`, and closes any other active `HearingTask`s and their children.

We only create `ScheduleHearingTask` and `HearingTask`s in 3 places, and every time we create one, we create the other as well. This PR makes it so we automatically create the parent `HearingTask` when we create the child `ScheduleHearingTask`.

### DIFF
--- a/spec/models/tasks/schedule_hearing_task_spec.rb
+++ b/spec/models/tasks/schedule_hearing_task_spec.rb
@@ -23,12 +23,16 @@ describe ScheduleHearingTask do
     end
   end
 
-  context "Create a ScheduleHearingTas with parent other than HearingTask type." do
+  context "Create a ScheduleHearingTask with parent other than HearingTask type." do
     let(:root_parent) { FactoryBot.create(:root_task, appeal: appeal) }
-    let(:schedule_hearing) { create(:schedule_hearing_task, parent: root_parent) }
+    let(:schedule_hearing) do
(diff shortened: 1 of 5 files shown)


spec/models/tasks/schedule_hearing_task_spec.rb
```diff
@@ -23,12 +23,16 @@
...
     let(:root_parent) { FactoryBot.create(:root_task, appeal: appeal) }
-    let(:schedule_hearing) { create(:schedule_hearing_task, parent: root_parent) }
+    let(:schedule_hearing) do
+      FactoryBot.create(:schedule_hearing_task, parent: root_parent, assigned_to: HearingsManagement.singleton)
+    end
 
-    it "should throw an error" do
-      expect { schedule_hearing }.to raise_error(Caseflow::Error::InvalidParentTask)
+    it "creates a HearingTask in between the input parent and the ScheduleHearingTask" do
+      expect { schedule_hearing }.to_not raise_error
```
sharonwarner: Up to you if you want to get rid of it or not, but I don't think this line provides much value. If it errors, the test will fail on the next line anyway.

Issue #9799, opened 2019-03-05
Intake | Turn off transient upstream errors in Slack


pkarman:
Increase the signal-to-noise ratio in #appeals-intake-alerts channel and do not log Sentry alerts for Intake-specific errors that look like transient upstream problems.

If we experience these errors during async jobs, they will retry automatically.

If we experience these errors during sync web interaction, we should send a nice message to the user asking them to try again.

Team discussion lead to these specifics action items:

* [x] Expand on the definition of TransientBGSError to include all "connect timeout" exceptions
* [x] Do not log TransientBGSError message in Sentry
* [x] Create more specific errors for VBMS errors, similar to the `BGSSyncError` class.
* [x] Add the @sierra tag to the daily Asyncable jobs report and report only on expired jobs.

## Validation

Since these tasks all effect day-to-day Slack support we'll validate manually once the changes are deployed to production.

pkarman:
Do we want to track actual sensitive violation errors?
e.g. https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3908/events/266193/

pkarman:
![Screen Shot 2019-03-14 at 8.37.27 AM.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/57968125341e740e0048fb83/125516c66e82c728ace21e0d46db978826878de9bce1b057e40fa0c5700b77206455f76b7191a370f1112c4f5d585edd3c4a4356bcbbd671f3e57862804e0ee5d33c5d22c33413a5b8d41bbf50bd208b1d87b3bc891ed2f42f67b6d5db1063474f4dcbed5727a19041c95df349e44a55a21a3f6ae7762431.png)

PR #9831, opened 2019-03-07
Filter asyncable jobs on expired

pkarman:
connects #9799 

### Description

Adds a new method to the `AsyncableJobs` class to easily filter by error message, useful for investigating failed jobs from the console.

Adds a new method to the `Asyncable` concern to check if a job is expired. Changes the daily cron report to only check the number of expired jobs (those that need manual intervention).


app/models/concerns/asyncable.rb
```diff
@@ -136,6 +136,13 @@ def submitted?
     !!self[self.class.submitted_at_column]
   end
 
+  def expired?
+    last_submitted = self[self.class.last_submitted_at_column]
```
jiahuang: does this also find jobs that were successfully submitted?

pkarman: ah good catch. Since the report is already filtered by not-yet-processed it would not include successful jobs, but this method all by itself would. I will fix to be consistent.

spec/services/asyncable_jobs_spec.rb
```diff
@@ -44,6 +47,20 @@
     it "sorts by the submited_at column, descending order" do
       expect(subject.jobs).to eq([sc_not_attempted_expired, hlr, sc, sc_not_attempted, sc_not_submitted])
     end
+
+    it "includes both expired and non-expired jobs" do
```
leikkisa: I find this test name to be confusing

pkarman: yes `expired_without_processing` is the equivalent class method. Are you suggesting naming the instance method the same as the class method?

leikkisa: Oh, no I just thought it might make sense to use the class method instead of using a new instance method with a select, since it's already there.

For this specific comment, I meant that it "includes both expired and non-expired jobs" is confusing because it seems like it's specifically looking for expired jobs in this test.

Issue #9794, opened 2019-03-05
Intake | DTA SC should have delayed submit_for_processing


pkarman:
Supplemental claims created as the result of a DTA decision may have a decision date in the future and should have an explicit delay in their `submit_for_processing` call.

### Validation
1. Create a HLR or SC with a benefit type that is not compensation or pension
2. Add a nonrating issue
3. Go to the non-compensation page and make a decision on that issue that is a DTA error, with a decision date that is in the future (maybe just one day)
4. After the decision date comes, check that the original decision review now has a remand supplemental claim, and that the submitted_at date is the same as the decision_date

leikkisa:
We do not allow receipt_dates in the future for Decision Reviews, and the receipt_dates are set to the decision_date, which can be in the future. This makes the specific issue above not a problem.  From what I've seen in appeals, if a decision_issue is created, the remand supplemental claim doesn't seem to be created or processed until after that date has arrived.

I need to check what happens in the case of HLRs, because technically this could cause an HLR to fail in creating the remand SC. We haven't seen this yet in prod.

I also can't find the Sentry error that instigated all of this :-/.

Sjones352:
<img width="1386" alt="Screen Shot 2019-03-14 at 7 20 40 PM" src="https://user-images.githubusercontent.com/23080951/54398619-37178000-4691-11e9-82eb-63e16f15561f.png">
<img width="1449" alt="Screen Shot 2019-03-14 at 7 21 31 PM" src="https://user-images.githubusercontent.com/23080951/54398620-37178000-4691-11e9-966f-6dc903df537d.png">
<img width="1505" alt="Screen Shot 2019-03-14 at 7 21 48 PM" src="https://user-images.githubusercontent.com/23080951/54398621-37178000-4691-11e9-9016-e1206a059672.png">
<img width="1680" alt="Screen Shot 2019-03-14 at 7 23 11 PM" src="https://user-images.githubusercontent.com/23080951/54398622-37178000-4691-11e9-8dc5-a8743b9a0d9b.png">
<img width="1341" alt="Screen Shot 2019-03-14 at 7 23 46 PM" src="https://user-images.githubusercontent.com/23080951/54398623-37b01680-4691-11e9-8dd0-8461c4df5cf4.png">
[...]

Issue #9643, opened 2019-02-26
ClaimDateDt invalid value


jiahuang:
See https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3467/

 "The data value of the ClaimDateDt within the claimToEstablish did not satisfy the following condition: The ClaimDateDt value must be a valid date for a claim"

ClaimDateDt value for this ep is "2019-02-26" which looks correct

pkarman:
This smells like a timezone bug so our tests should exercise some shifting time zones.

pkarman:
Related https://github.com/department-of-veterans-affairs/caseflow/pull/9577

leikkisa:
We need to delay submitting for processing for End Product Establishments if the claim_date is in the future.  This can happen because the claim_date is based on the receipt_date, and for DTA claims, this is based on the approx_decision_date.  This is sometimes entered by users in the UI.

leikkisa:
Related #9794

leikkisa:
Hopefully #9794 and tickets before it fixes this, but we'll need to wait and see if this error continues to pop up.

Sjones352:
Hey Sally can you please provide validation instructions.

leikkisa:
@Sjones352 

I think the best way to validate this might be checking if it happens any more in Prod because it seems to be caused by some timezone differences in the servers that will be difficult to reproduce, so I recommend waiting on this and then we can check the alerts for the last time it happened and see if it has happened after this fix. 

However, here are some instructions we could try also. There's no guarantee that these will reproduce the error though.

### Validation instructions
- Do a noncompensation AMA intake
- Add a request issue
- Go to the Noncompensation review page to make a decision on that request issue
- Give it a DTA error decision
- Give it a decision_date that is several days in the future
- Check on the same day that the decision_issue has a submitted_at of the future decision_date
- Check that the DTA Supplemental Claim doesn't get created until that decision_date

Note - if it tries before the decision date, it should fail with the ClaimDateDt isn't valid error.  However, if this happens it will eventually retry and succeed once the decision_date has come to pass.

Sjones352:
**Currently staged in Uat** 
<img width="1219" alt="Screen Shot 2019-03-18 at 12 21 13 PM" src="https://user-images.githubusercontent.com/23080951/54545749-a71b5400-4978-11e9-9605-2376a2ce72ad.png">
[...]

PR #9843, opened 2019-03-07
Handle future decision dates for DTA SC

leikkisa:
connects #9794
connects #9643


app/models/concerns/asyncable.rb
```diff
@@ -95,7 +95,8 @@ def potentially_stuck
   end
 
   def submit_for_processing!(delay: 0)
-    when_to_start = delay.try(:to_datetime) ? delay.to_datetime : Time.zone.now + delay
+    # One minute offset to prevent "this date is in the future" errors with external services
+    when_to_start = delay.try(:to_datetime) ? delay.to_datetime + 1.minute : Time.zone.now + delay
```
pkarman: you're thinking maybe this is a time sync problem? Like their server clocks are drifted? That's a good hypothesis to test out.

spec/models/appeal_intake_spec.rb
```diff
@@ -132,6 +132,7 @@
     end
 
     context "Claimant is different than Veteran" do
+      before {}
```
pkarman: I get why this no-op `before` is here but I wonder if it deserves a comment, or maybe re-thinking the test file structure to avoid it? Was there a specific earlier `before` you were trying to override?

spec/models/higher_level_review_spec.rb
```diff
@@ -295,6 +295,22 @@
         it "creates DecisionReviewTask" do
           expect { subject }.to change(DecisionReviewTask, :count).by(1)
         end
+
+        context "when decision date is in the future" do
```
pkarman: nice test!

app/models/claim_review.rb
```diff
@@ -10,6 +10,7 @@ class ClaimReview < DecisionReview
   has_one :intake, as: :detail
 
   with_options if: :saving_review do
+    validate :validate_receipt_date
```
pkarman: I see the test for "when saving review" -- do we a test for `Appeal`? Otherwise I'm not 100% sure why we needed to split out this validation into the separate subclasses.

spec/models/appeal_intake_spec.rb
leikkisa: Oh, this was mistakenly left in, I changed how I was doing appeal. Good catch.

app/models/claim_review.rb
leikkisa: This has been a thorn in my side before, but Appeal uses a context called intake_review that's been there.  I've tried changing it before so they'd all be consistent but ran into issues (although maybe I'd be more successful now).  Since this only affects SCs I figured it was okay for this to be outside of the intake_review context.

app/models/concerns/asyncable.rb
leikkisa: I am curious too! My new theory is that there is like some rogue server that is in a different timezone, haha (which this wouldn't help, but it seems worth checking).  It's hard to tell because it fixes itself, and we can't see if there were attempts in between that failed. The ones that did though both succeeded 7 hours and 16 minutes later, which is weird.

screen shot 2019-03-07 at 4 33 39 pm

Issue #9730, opened 2019-03-01
Intake Edit | If all issues are removed on a Caseflow decision review, close associated tasks


shanear:
We need to allow users to remove entire appeals (and non-comp decision reviews), not just individual issues. To start, we will allow appeals to be removed by removing all the issues and clicking save.

## Acceptance Criteria
- On the intake edit page for an appeal or non-comp decision review, do not disable the save button when all issues are removed.
- When the save button is pressed, in addition to removing all the issues on the decision review, all open tasks on that appeal should be closed, and the appeal should be considered closed and removed.

## Notes
- We will be working on a short hand way to close the entire appeal later, instead of having to remove each individual issue
- The parallel issue for VBMS reviews is #9731

## Technical Notes
- This functionality should be enabled and disabled using a feature flag named `:remove_decision_reviews`
- Ask @lowellrex how tasks should be closed on the appeal

jiahuang:
Talked to Lowell, open tasks should be cancelled by something like `appeal.tasks.active.update_all!(status: Constants.TASK_STATUSES.cancelled)`

Additionally, we may want to differentiate a manually closed appeal from other appeals by either:
1. add a status column in appeal that gets set to "manually removed"
2. setting the existing `establishment_error` field to be "manually removed". We could also expose this field to the UI so that the worker can put in a more detailed message.

jiahuang:
@Sjones352 A general approach for the backend portion of this ticket would be to first write failing specs that test the following scenarios when the remove_decision_reviews flag is turned on:
* when there is an appeal intake with multiple active tasks and all request issues are removed, ensure that all active tasks are set to a canceled state
* when there is an appeal intake with multiple active tasks and not all request issues are removed, ensure that no active tasks are set to a canceled state
* when there is an appeal intake with completed tasks and all request issues are removed, ensure that completed tasks do not get set to a canceled state
* the 3 above scenarios should be tested for higher level reviews & supplemental claims

The backend portion can be implemented by:
1. Adding a `cancel_active_tasks!` method to the decision_review model. This will cancel all active tasks with the code in the prior comment.
2. Modify the `perform!` method in `RequestIssueUpdate` to run `cancel_active_tasks!` when `after_issues` is empty (which will mean that all issues will be deleted). This should run inside the transaction.

pkarman:
Validated as part of #9733 -- the task created was closed. https://appeals.cf.uat.ds.va.gov/decision_reviews/fiduciary/tasks/1342

PR #9852, opened 2019-03-07
All issues are removed on a Caseflow decision review

Sjones352:
connects #9730

@pkarman and @leikkisa wondering if I can create 3 separate files for the tests instead of adding more tests to supplemental claims and appeals as these files are already too big.


spec/feature/intake/remove_decision_reviews_spec.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
```
jiahuang: this file can be removed

spec/feature/intake/higher_level_review/edit_spec.rb
```diff
@@ -1089,4 +1089,92 @@ def click_cancel(visit_page)
       end
     end
   end
+
+  context "when remove decision reviews is enabled" do
```
jiahuang: this test should be repeated for supplemental claims and appeals

app/models/decision_review.rb
```diff
@@ -185,6 +185,10 @@ def establish!
     # no-op
   end
 
+  def cancel_active_tasks!
+    tasks.active.update_all(status: Constants.TASK_STATUSES.cancelled)
```
leikkisa: Should we use the task version of canceling a task? 
cancel_task_and_child_subtasks

spec/feature/intake/higher_level_review/edit_spec.rb
```diff
@@ -1089,4 +1089,92 @@ def click_cancel(visit_page)
...
+
+    after do
+      FeatureToggle.disable!(:remove_decision_reviews, users: [current_user.css_id])
+    end
+
+    let(:today) { Time.zone.now }
+    let(:last_week) { Time.zone.now - 7.days }
+    let(:higher_level_review) do
+      # reload to get uuid
+      create(:higher_level_review, :with_end_product_establishment, veteran_file_number: veteran.file_number).reload
```
leikkisa: I don't think this should have an end product establishment, because then it wouldn't have tasks. Instead, I would give it a benefit_type of the non_comp_org.

There might be validation preventing an organization called "Non-Comp Org" though so you may have to use a real one.

spec/feature/intake/higher_level_review/edit_spec.rb
```diff
@@ -1089,4 +1089,92 @@ def click_cancel(visit_page)
...
+        # to run the job even though job should be processed sync
+        expect(in_progress_task.reload.status).to eq(Constants.TASK_STATUSES.cancelled)
+
+        # going back to the edit page does not show any requested issues
+        visit "higher_level_reviews/#{higher_level_review.uuid}/edit"
+        expect(page).not_to have_content(existing_request_issues.first.description)
+        expect(page).not_to have_content(existing_request_issues.second.description)
+      end
+
+      scenario "no active tasks cancelled when request issues remain" do
```
leikkisa: This scenario may not apply for Appeals, since their tasks will be per benefit_type, although that may already be handled i the code.

spec/feature/intake/higher_level_review/edit_spec.rb
Sjones352: Added Supplemental claim spec

app/models/decision_review.rb
pkarman: Right now our `DecisionReview` tasks do not have descendents, so calling `cancel_task_and_child_subtasks` isn't strictly necessary, but probably good future-proofing.

```ruby
def cancel_active_tasks!
  tasks.each(&:cancel_task_and_child_subtasks)
end
```

should cover it.

spec/feature/intake/higher_level_review/edit_spec.rb
```diff
@@ -1089,4 +1089,92 @@ def click_cancel(visit_page)
...
+          click_remove_intake_issue(1)
+          click_remove_issue_confirmation
+        end
+
+        safe_click("#button-submit-update")
+        safe_click ".confirm"
+        expect(page).to have_content(Constants.INTAKE_FORM_NAMES.higher_level_review)
+        expect(completed_task.reload.status).to eq(Constants.TASK_STATUSES.completed)
+        # TODO: for some reason this test is flaky, seems like it takes some time
+        # to run the job even though job should be processed sync
```
pkarman: I've seen this before when there are multiple bg jobs processed sync, because the browser is still async in a feature test. I just throw a `sleep 1` here, like in this test.

https://github.com/department-of-veterans-affairs/caseflow/blob/5bbbe90117365a8689e3f894a3cc4223ea229569/spec/feature/intake/nonrating_request_issue_modal_spec.rb#L58

spec/feature/intake/supplemental_claim/edit_spec.rb
```diff
@@ -552,4 +552,93 @@ def click_cancel(visit_page)
...
+          click_remove_intake_issue(1)
+          click_remove_issue_confirmation
+        end
+
+        safe_click("#button-submit-update")
+        safe_click ".confirm"
+        expect(page).to have_content(Constants.INTAKE_FORM_NAMES.supplemental_claim)
+        expect(completed_task.reload.status).to eq(Constants.TASK_STATUSES.completed)
+        # TODO: for some reason this test is flaky, seems like it takes some time
+        # to run the job even though job should be processed sync
```
pkarman: same thing here, a `sleep 1` should address it.

spec/feature/intake/supplemental_claim/edit_spec.rb
```diff
@@ -552,4 +552,93 @@ def click_cancel(visit_page)
...
+        expect(completed_task.reload.status).to eq(Constants.TASK_STATUSES.completed)
+        expect(in_progress_task.reload.status).to eq(Constants.TASK_STATUSES.in_progress)
+      end
+
+      context "when review has no active tasks" do
+        scenario "no tasks are cancelled when all request issues are removed" do
+          visit "supplemental_claims/#{supplemental_claim.uuid}/edit"
+          click_remove_intake_issue(1)
+          click_remove_issue_confirmation
+          safe_click("#button-submit-update")
```
pkarman: you can use `click_edit_submit_and_confirm` instead of these 2 `safe_click` statements.

spec/feature/intake/higher_level_review/edit_spec.rb
```diff
@@ -1089,4 +1089,92 @@ def click_cancel(visit_page)
...
+      end
+    end
+
+    context "when review has no active tasks" do
+      scenario "no tasks are cancelled when all request issues are removed" do
+        visit "higher_level_reviews/#{higher_level_review.uuid}/edit"
+        click_remove_intake_issue(1)
+        click_remove_issue_confirmation
+        safe_click("#button-submit-update")
+        safe_click ".confirm"
```
pkarman: check out `click_edit_submit_and_confirm` instead of these 2 safe_click statements.

app/models/decision_review.rb
```diff
@@ -185,6 +185,10 @@ def establish!
     # no-op
   end
 
+  def cancel_active_tasks!
+    tasks.each(&:cancel_task_and_child_subtasks)
```
pkarman: now that I look at it again, I wonder if we should wrap this in a transaction Just In Case?

```ruby
transaction do
  tasks.each(&:cancel_task_and_child_subtasks)
end
```

so they either all get canceled or none of them do.

spec/feature/intake/appeal/edit_spec.rb
```diff
@@ -374,4 +374,74 @@
       expect(page).to have_content("This appeal has been outcoded and the issues are no longer editable.")
     end
   end
+
+  fcontext "when remove decision reviews is enabled" do
```
pkarman: left over focus?

app/models/request_issues_update.rb
```diff
@@ -110,6 +111,10 @@ def validate_before_perform
     !@error_code
   end
 
+  def allow_zero_request_issues
```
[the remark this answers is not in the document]
monfresh: We can ignore this error. I will update the Reek config to check this offense only for public methods. Having said that, what do you think about adding a question mark to the end of the method name? Methods that return `true` or `false`, also known as predicate methods in Ruby, should end with a question mark.

app/models/decision_review.rb
Sjones352: It is wrapped in a transaction in the perform method in the `request_issues_update.rb`

app/models/request_issues_update.rb
```diff
@@ -125,4 +130,8 @@ def process_legacy_issues!
   def strip_removed_issues!
     removed_issues.each(&:remove_from_review)
   end
+
+  def cancel_active_tasks!
```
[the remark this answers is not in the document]
Sjones352: https://stackoverflow.com/questions/612189/why-are-exclamation-marks-used-in-ruby-methods. Thought I would share why I changed it.

PR #9860, opened 2019-03-08
Automatically set assignee for hearing tasks

lowellrex:
While working on PR #9824 (in service of issue #9543), I made some changes to automatically set the assignee for a few hearings-related tasks. That PR has started to balloon so I've busted those changes out here to keep the size of the other PR down.


app/models/tasks/hearing_task.rb
```diff
@@ -7,9 +7,14 @@
 
 class HearingTask < GenericTask
   has_one :hearing_task_association
+  before_validation :set_assignee
```
lowellrex: We only create `HearingTask`s in [one place anymore](https://github.com/department-of-veterans-affairs/caseflow/pull/9860/files#diff-6295d901c347ce4280a0880ec65238e8R63), but in case we start creating them elsewhere we will no longer have to explicitly set the assignee.

app/models/tasks/no_show_hearing_task.rb
```diff
@@ -3,4 +3,11 @@
 # Task created after an appellant no-shows for a hearing. Gives the hearings team the options to decide how to handle
 # the no-show hearing after the judge indicates that the appellant no-showed.
 class NoShowHearingTask < GenericTask
+  before_validation :set_assignee
+
+  private
+
+  def set_assignee
+    self.assigned_to = assigned_to.nil? ? HearingAdmin.singleton : assigned_to
```
lowellrex: When `NoShowHearingTask`s are created they are assigned to the `HearingAdmin` organization, but we may eventually allow them to be assigned to individuals. This ternary allows for that possibility.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -135,6 +136,10 @@ def withdraw_hearing_data(_user)
 
   private
 
+  def set_assignee
+    self.assigned_to = assigned_to.nil? ? HearingsManagement.singleton : assigned_to
+  end
```
lowellrex: When we create `ScheduleHearingTask `s in the application code they are assigned to the `HearingsManagement ` organization, but they may eventually be assigned to individuals. This ternary allows for that possibility.

PR #9869, opened 2019-03-08
Fix a non-idempotent appeal spec

monfresh:
**Why**: The spec was testing the `ordered_by_distribution_ready_date`
scope in `Appeal`, which only orders by the tasks's `assigned_at` if the
task is of type `DistributionTask`. However, the test was creating
`RootTask`s only, and so the scope was only ordering by primary key ID.
Simply creating 2 factories doesn't always guarantee that the first
one listed in the spec will be the first one created in the DB.

**How**: Use `DistributionTask`s in the spec, and update the
`assigned_at` to a specific date to ensure the test always returns
results in the same order.


amprokop:
Didn't know that we couldn't rely on the ordering there. Thanks for teaching me that, @monfresh!

Issue #9207, opened 2019-02-05
Allow individual queue folks to place tasks on hold


laurjpeterson:
The Translation organization just got their first AMA appeal translation task assigned to them! However, the only actions they have available to them are Re-assign to person, Assign to team, or Mark task complete. We need to add the Place on hold option

## Acceptance Criteria
* [ ] Translation team individuals (and other repeatable queue individuals, like hearing branch individuals) can place tasks assigned to them on hold
* [ ] Team admins can place tasks assigned to them or to their organizations on hold
* [ ] For cases that have been placed on hold, users have all actions they had available to them when the case was in their assigned tab (in the assigned status), plus:
   - **End hold** - this (1) completes the hold task and (2) changes the task's status to "assigned" so the task appears in the user's assigned tab (instead of on hold tab)

## Screenshot from demo
<img width="1124" alt="screen shot 2019-02-20 at 4 18 22 pm" src="https://user-images.githubusercontent.com/27704093/53125144-388ad800-352b-11e9-9789-dae4217a1bac.png">

lowellrex:
Let's try to use the [`TimeableTask` model](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/concerns/timeable_task.rb) here to accomplish this if we can.

lowellrex:
MEDIUM.

amprokop:
![image](https://user-images.githubusercontent.com/3781835/54153311-a4b77800-4415-11e9-90fa-77d19fbe92f0.png)
The organization view doesn't have an "on hold" tab for tasks.

laurjpeterson:
here's the ticket to add on hold tab to team queue - https://github.com/department-of-veterans-affairs/caseflow/issues/9341

is this something that you could take on as well?

laurjpeterson:
@lowellrex - i'm unassigning @amprokop  from this task. let's connect today or tomorrow about how we might finish this ticket this or next sprint!

lowellrex:
Rough rollout plan:
* [ ] Create new `TimedHoldTask`
* [ ] Make sure `GenericTask`s respond properly to having a child `TimedHoldTask`
  * "Days on hold" appears correctly
  * Actions display correctly (all of the same actions as before + "Cancel timed hold")
  * Updating the parent task cancels the `TimedHoldTask` (are there any scenarios where this is not true?)
* [ ] Enable creation of `TimedHoldTask`s for `GenericTask`s
[...]

PR #9888, opened 2019-03-11
enables on hold for user

amprokop:
https://github.com/department-of-veterans-affairs/caseflow/issues/9207

This code is not yet fully tested nor does it have enough automated tests, just surfacing for feedback on the general approach.

Creates a new TimedHoldTask class that imports TimeableTask. To put a GenericTask on hold, we will now call GenericTask.place_on_hold(parent, current_user, params).

Colocated tasks are not changed for now, but I suggest we plan to migrate them when convenient.

I considered sending the TimedHoldTask to the frontend and then just updating the Case Timeline and all the frontend logic to handle this new TimedHoldTask correctly, but decided against it for the moment, out of a sense that the frontend shouldn't need to know about all tasks necessarily, and shouldn't be responsible for sorting through backend implementation details. 

So, the smoothest past forward seemed to be to redefine the `placed_on_hold_at` and `on_hold_duration` methods on GenericTask, so we wouldn't have to change anything from the serializer or the frontend (theoretically). This has the effect that we need to filter out TimedHoldTasks from the list of tasks we return to the frontend.

Suggested path forward for future PRs if we continue with this approach:

- Let this in production for awhile to gain confidence that it works as expected
- Migrate new ColocatedTask holds to use TimedHoldTask (may require DB backfilling for holds that are already in progress)
- Consider removing the following columns - "on_hold_duration",  "placed_on_hold_at" — we should be able to just rely on the TaskTimer we create to keep track of the on hold duration, and the created_at field on the TimedHoldTask to determine when the parent task was placed on hold.
    - Are those columns used anywhere else, though?

Open questions:

- Where in the DB should we put the instructions field? On the TimedHoldTask row or on the parent task?
- What's the best place to filter out TimedHoldTasks before we serialize tasks to send to the frontend?
- Do we need to migrate JudgeTask and AttorneyTask to this approach as well? It doesn't seem like they can be placed on a timed hold.

Test plan (not ran yet):
Screen Shot 2019-03-11 at 2 58 42 PM
Screen Shot 2019-03-11 at 2 59 37 PM

- placed task on hold
- verified that it shows up in the queue
- in the console, called `when_timer_ends` on the on hold task
- verified that the appeal shows up in the queue again


lowellrex:
Did we explore creating a new `OnHoldTask` using the [`TimeableTask` concern](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/concerns/timeable_task.rb) as the child of tasks we want to put on hold in order to avoid some of the problems we run into with the way `ColocatedTask`s implement the holding logic?

Quoting from a slack thread about one of the issues with the current implementation of the on hold ability:
> One problem with the way we currently handle putting tasks on hold (by using fields of the task) is that we don't properly handle what happens if a task gets placed on hold a second time, but it is not a timed hold

> So if we place a 5 day hold for a task on March 1st, hold expires on March 5th, then on March 6th we create a child task off of this task then the task will immediately be taken off hold when somebody loads there queue because the `placed_on_hold_at` value and the `on_hold_days` value indicate that the tasks hold ended already (which it did) without any indication that the new hold (caused by the child task being created) is different than the timed hold

In addition to not distinguishing from the different types of holds we don't currently respect the `placed_on_hold_at` value [when cascading completion status](https://github.com/department-of-veterans-affairs/caseflow/blob/3d5579a87353c42b1205e533a65df85facdccb57/app/models/task.rb#L451) from children tasks to parents.

app/models/tasks/generic_task.rb
```diff
@@ -34,6 +34,7 @@ def available_actions(user)
         Constants.TASK_ACTIONS.ASSIGN_TO_TEAM.to_h,
         Constants.TASK_ACTIONS.REASSIGN_TO_PERSON.to_h,
         Constants.TASK_ACTIONS.MARK_COMPLETE.to_h,
+        Constants.TASK_ACTIONS.PLACE_HOLD.to_h,
```
lowellrex: We override `available_actions` [for a handful of classes](https://github.com/department-of-veterans-affairs/caseflow/search?q=def+available_actions&type=Code) that inherit from `GenericTask` so we'll need to add this option to those classes as well if we want them to be able to be placed on hold.

lowellrex: The [PR that added the ability to cancel tasks](https://github.com/department-of-veterans-affairs/caseflow/pull/9499) might be a good reference for how we could go about accomplishing this.

amprokop:
It may be that creating a new task class and migrating everything over to that is the best approach — or we could directly fix the bugs you mentioned, I have no opinion here.

I'm happy to not merge this and start exploring either approach — but I'm not confident that I'll have time to complete either of the alternative approaches this sprint, my shallow read is that both are much more of a lift. You and/or @laurjpeterson should make that call, is it better to merge this now with some known issues, or wait for a more perfect solution?

app/models/tasks/generic_task.rb
amprokop: Thanks! I'll update the available actions for those tasks.

laurjpeterson:
I'd prefer to merge this now so users can have the ability to place tasks on hold, which is "generic queue" functionality we trained them on, but haven't had. I know refactoring before merging might be the desired way forward, but if we don't have bandwidth now to do it, I worry about delaying this feature for another few months.

lowellrex:
I am strongly opposed to merging this as is until we do more exploration about [the approach I suggested above](https://github.com/department-of-veterans-affairs/caseflow/pull/9888#issuecomment-472860103). I think 2 hours of exploration with a quick summary of findings if it won't pan out would be well worth that time.

I believe we will spend a significant amount of time supporting this approach, and if we eventually move over to a more robust approach that time spent will be negatively productive in getting the refactor in because it would be time spent on things other than developing the robust approach.

amprokop:
@lowellrex — cool, I'm happy to commit 2 hours to this this sprint — I am confident that the approach you proposed will work, but my guess is that implementation and any subsequent fixes will take significantly more than 2 hours, and I can't commit to having it done before next Friday. Does that work for you?

lowellrex:
> I can't commit to having it done before next Friday. Does that work for you?

Yup! That works! I think as long as we can get this out the door before the end of the sprint (Fri 22 March) then we'll be good to go!

amprokop:
@lowellrex — hmm I'm not sure I'm going to be able to do that, it will depend on how big of a task it winds up being.

app/models/appeal.rb
```diff
@@ -705,6 +705,10 @@ def bgs
     BGSService.new
   end
 
+  def tasks_for_frontend
```
amprokop: I think this method should probably be in the task model itself. Will move it. But the general idea is that we'll save the frontend from having to walk the task tree to match up on hold tasks with their parents to get those details.

app/models/tasks/generic_task.rb
```diff
@@ -55,9 +56,23 @@ def available_actions(user)
 
     []
   end
+
   # rubocop:enable Metrics/MethodLength
   # rubocop:enable Metrics/AbcSize
 
+  def placed_on_hold_at
```
amprokop: So we're overriding the methods that would automatically be accessing DB fields (see the task schema) in GenericTask for now — eventually, I'd favor getting rid of the `placed_on_hold_at` and `on_hold_duration` columns entirely.

app/models/tasks/generic_task.rb
```diff
@@ -92,5 +107,24 @@ def create_child_task(parent, current_user, params)
         )
       end
     end
+
+    def place_on_hold(parent, current_user, params)
+      transaction do
+        # Remove any other hold tasks so we don't wind up with more than one
+        if parent.timed_hold_task
```
amprokop: My guess is that we should only permit one hold on a task at a time, and any new hold should blow away the old one. Open to other suggestions though.

app/models/appeal.rb
tomas-nava: It might make sense to [follow this pattern](https://github.com/department-of-veterans-affairs/caseflow/pull/9018/files), where we hide tasks from various views with methods that return booleans. So we have `hide_from_case_timeline` and `hide_from_task_snapshot`; you could add something like `hide_from_task_list` and then filter on that value (whether it's here in the Appeal model, on the Task model, or on the frontend).

app/models/tasks/generic_task.rb
```diff
@@ -92,5 +107,24 @@ def create_child_task(parent, current_user, params)
...
+        parent.update!(status: Constants.TASK_STATUSES.on_hold)
+
+        TimedHoldTask.create!(
+          appeal: parent.appeal,
+          parent_id: parent.id,
+          instructions: params[:instructions],
+          on_hold_duration: params[:on_hold_duration]
+        )
+      end
+    end
```
tomas-nava: I think this approach is right. I have some syntactical/implementation feedback which you may've been considering already, so I'll be brief.

- the task parameter should be named `task` instead of `parent`
- this (plus the `timed_hold_task` getter) will only get the most recent active `TimedHoldTask` child. I know it shouldn't be possible to have more than one active `TimedHoldTask` as a child of a `GenericTask`, but maybe we can get all active tasks and destroy them all to be sure?
- can we cascade destruction of the `TaskTimer` when the `TimedHoldTask` is destroyed (or vice versa)?
- I don't think you need to explicitly set the parent `on_hold`, [this hook should do it for you](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/task.rb#L19).

tomas-nava: Actually, do we want to `destroy` tasks or should we `cancel` them so there's a record?

app/models/appeal.rb
amprokop: oh wait — i think what i actually want is to return true for `hide_from_case_timeline` and `hide_from_task_snapshot` here

amprokop: thanks!

app/models/tasks/timed_hold_task.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+##
+# Shell task that, when created as a child of a GenericTask, places that task on hold.
+# A nightly job queries for and expires
+
+class TimedHoldTask < GenericTask
+  include TimeableTask
+
+  validates :on_hold_duration, presence: true, inclusion: { in: 1..100 }
```
lowellrex: Why did we pick an upper limit of 100 days?

app/models/tasks/timed_hold_task.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def when_timer_ends
+    update!(status: :completed)
+  end
+
+  def timer_ends_at
+    Time.zone.today + on_hold_duration.days
```
lowellrex: Can we update this function to use a static date (maybe the task's creation date?) instead of the current date? It looks like we only call this function [when we create the `TaskTimer`](https://github.com/department-of-veterans-affairs/caseflow/blob/f47c5caac69cbaf79f9063fe992b3bdd4fe1fb02/app/models/concerns/timeable_task.rb#L13), but we may forget about that convention in the future and I think it would make sense to have this function always return the same value regardless of whether we call this function the day the task was created or 3 days later.

app/models/tasks/generic_task.rb
lowellrex: Maybe we can move this into the base `Task` class and override these methods only in `ColocatedTask` since that is the only task class that uses timed holds? Either way, can we write up a ticket that includes a migration plan for abandoning the old (`ColocatedTask`) way of doing timed holds in favor of this new strategy?

app/models/tasks/timed_hold_task.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+##
+# Shell task that, when created as a child of a GenericTask, places that task on hold.
+# A nightly job queries for and expires
+
+class TimedHoldTask < GenericTask
+  include TimeableTask
```
lowellrex: Might we need an `available_actions` item that ends the hold early?

app/controllers/tasks_controller.rb
```diff
@@ -79,7 +79,11 @@ def create
   #   on_hold_duration: "something"
   # }
   def update
-    tasks = task.update_from_params(update_params, current_user)
+    tasks = if task.is_a?(GenericTask) && update_params.status == "on_hold"
+              GenericTask.place_on_hold(task, update_params)
+            else
+              task.update_from_params(update_params, current_user)
+            end
```
lowellrex: Since the only field of the parent task we're updating here is the "status" field [by way of the `after_create` hook](https://github.com/department-of-veterans-affairs/caseflow/blob/f47c5caac69cbaf79f9063fe992b3bdd4fe1fb02/app/models/task.rb#L19) perhaps it would be more precise to have this be a `create` action since we are creating a new instance of the `TimedHoldTask`?

If we use that approach we would have to move [the logic to cancel existing task timers](https://github.com/department-of-veterans-affairs/caseflow/pull/9888/files#diff-9f7b57df79bdb2c539ae04506d0f5a67R120) associated with the task to the `TimedHoldTask` class, but it would allow us to not change `TasksController` at all.

app/models/tasks/timed_hold_task.rb
lowellrex: This work should probably be completed in another PR, along with the front-end work to add the modal to place tasks on hold.

app/models/tasks/timed_hold_task.rb
amprokop: I believe the UI goes up to 90 days. Open to changing it to whatever you like.

app/models/tasks/timed_hold_task.rb
amprokop: Sure, that sounds like a good tweak. Thanks!

app/models/tasks/timed_hold_task.rb
amprokop: I'm not sure, do we currently have an option in the UI to cancel holds on tasks?

app/models/tasks/generic_task.rb
amprokop: Updated this to cancel them.

app/models/tasks/timed_hold_task.rb
lowellrex: Makes sense to me!

app/models/tasks/timed_hold_task.rb
lowellrex: We can't currently cancel holds so this definitely makes sense to be part of a different PR if we want to do it.

lowellrex:
Closing this out in favor of other PRs (linked in #9207) that will continue this work.

Issue #9875, opened 2019-03-08
Sort out what active and closed mean


shanear:
In auditing the DB, i noticed there are different definitions of open and active all over the codebase, and I believe it might be causing issues with the VA.gov API as well as other places in the code.

- For instance, some places in the code are scoping request issues to `open` issues, however the scope should include `open` and `decided` issues. I think we'll want a scope specifically for this. My best idea is `active_or_decided` but I'm open to a better word.

- We are looking at different data to determine whether an issue or decision review is active. Some look at tasks, some look at end products. We should coalesce on what it means for a request issue or decision review to be `active?`

- Ineligible issues should be immediately closed with a closed status of ineligible.

shanear:
Confirmed this is actually causing a bug where we are unable to reintake closed issues

PR #9893, opened 2019-03-11
Refactor/fix duplicate issue check to use open/closed and ineligible

shanear:
connects #9875


app/models/contestable_issue.rb
```diff
@@ -102,37 +102,31 @@ def find_latest_contestable_issues
...
-
-    return unless different_decision_review(found_request_issue)
-
-    found_request_issue
-  end
-
-  def different_decision_review(found_request_issue)
-    return unless found_request_issue
-
-    found_request_issue.decision_review_id != contesting_decision_review.id ||
```
shanear: This is redundant of the scope here: 
https://github.com/department-of-veterans-affairs/caseflow/pull/9893/files#diff-703856fcb983e2497bfd75cf7ffd77afR126

app/models/contestable_issue.rb
```diff
@@ -102,37 +102,31 @@ def find_latest_contestable_issues
...
 
   def conflicting_request_issue_by_decision_issue
     return unless decision_issue&.id
 
-    potentially_conflicting_request_issues.find_active_by_contested_decision_id(decision_issue.id)
+    potentially_conflicting_request_issues.find_by(contested_decision_issue_id: decision_issue.id)
   end
 
   def potentially_conflicting_request_issues
-    RequestIssue.where.not(decision_review: contesting_decision_review)
```
shanear: The scope here was previously not getting used. The logic here was being implemented by:
https://github.com/department-of-veterans-affairs/caseflow/pull/9893/files#diff-703856fcb983e2497bfd75cf7ffd77afL131

app/models/request_issue.rb
```diff
@@ -153,7 +153,15 @@ def not_deleted
       where.not(decision_review_id: nil)
     end
 
+    def eligible
```
leikkisa: I know these are little, but I believe there's a place where these scopes have tests, might be good to add one here.

app/models/contestable_issue.rb
```diff
@@ -102,37 +102,31 @@ def find_latest_contestable_issues
...
 
+  def conflicting_request_issue
+    conflicting_request_issue_by_decision_issue || conflicting_request_issue_by_rating
+  end
+
   def conflicting_request_issue_by_rating
     return unless rating_issue_reference_id
 
-    potentially_conflicting_request_issues.find_active_by_contested_rating_issue_reference_id(rating_issue_reference_id)
+    potentially_conflicting_request_issues.find_by(contested_rating_issue_reference_id: rating_issue_reference_id)
```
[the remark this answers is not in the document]
shanear: `potentially_conflicting_request_issues` does scope it to open issues

app/models/request_issue.rb
shanear: yeah, i have a longer way to go here, I'm gonna add tests for these scopes in a follow up PR.

Issue #9894, opened 2019-03-11
Intake | Increase processing window for Request Issue decision sync


pkarman:
We currently give up attempting to sync decisions after the default 4 days. It may reasonably take longer than that.

Try for 14 days for now, and watch production to consider changing that higher.

### Validation

Peter is watching production to verify that older RequestIssue jobs automatically re-try.

pkarman:
```
decision_sync_last_submitted_at: Thu, 07 Mar 2019 17:46:24 UTC +00:00,
decision_sync_attempted_at: Thu, 14 Mar 2019 02:50:30 UTC +00:00,
```

window has gotten larger.

PR #9895, opened 2019-03-11
Increase retry window for Request Issue decision sync

pkarman:
connects #9894

### Description

Adjusts the processing window up to 14 days for Request Issues.

Also fixes how subclasses can override the `REQUIRES_PROCESSING_WINDOW_DAYS` class var.


spec/models/request_issue_spec.rb
```diff
@@ -118,6 +118,18 @@
       expect(todo).to_not include(rating_request_issue)
       expect(todo).to include(nonrating_request_issue)
     end
+
+    it "keeps trying for #{RequestIssue::REQUIRES_PROCESSING_WINDOW_DAYS} days" do
```
leikkisa: Nice test!

Issue #9444, opened 2019-02-15
Properly handle duplicate records in users table


lowellrex:
Occasionally we run into issues where the wrong user is added to an organization because multiple records exist for different casings of the same CSS ID (`BVAWOODL` vs `bvawoodl`). One such instance is chronicled in [this Slack thread](https://dsva.slack.com/archives/C6E41RE92/p1550069807090000). A quick survey of the `users` table indicates that 10% or more of the records in that table are duplicates:

```ruby
rails c> User.count
# 2519
rails c> User.select("upper(css_id), station_id, count(*)").group("upper(css_id), station_id").having("count(*) > 1").length
# 105
```

That `105` number translates to 210 duplicate records (or more since some of those may have more than 2 casings grouped together). This ticket exists to address this problem.

## Acceptance criteria
* [ ] Determine how `User` records with the same CSS ID reach the database
* [ ] Prevent duplicate `User` records from reaching the database
* [ ] Improve application's resiliency when handling a `User` record that is not precisely the one we want (when CSS ID casing doesn't match, for instance)

pkarman:
I suggest we:

* [ ] Prevent new users from being created with unknown case. Always lowercase (or UPPERCASE, I just prefer lower) the `css_id` in `find_by_css_id_or_create_with_default_station_id` method which seems to be how users are created
* [ ] Resolve the existing duplicates manually (see e.g. #9593)
* [ ] Add [unique constraint](http://shuber.io/case-insensitive-unique-constraints-in-postgres/) on the db, enforcing case

pkarman:
@mdbenjam and I have been working on code to merge duplicates with an `undo` column. The plan at this point is to 

* [ ] merge all the user accounts to the UPPERCASE css id
* [ ] let the undo column sit for a week, then drop it

PR #9907, opened 2019-03-12
Add foreign key on all tables with user_id

pkarman:
connects #9444

### Description
Adds a real foreign key constraint to all tables with a `user_id` column, to prevent accidentally deleting a user referenced elsewhere.

### DIFF
--- a/lib/tasks/users.rake
+++ b/lib/tasks/users.rake
@@ -24,7 +24,7 @@ namespace :users do
   def models_with_user_id
     load_all_models unless @models_with_user_id
     @models_with_user_id ||= ActiveRecord::Base.descendants.reject(&:abstract_class?)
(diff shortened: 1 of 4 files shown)


lib/tasks/users.rake
```diff
@@ -24,7 +24,7 @@ namespace :users do
   def models_with_user_id
     load_all_models unless @models_with_user_id
     @models_with_user_id ||= ActiveRecord::Base.descendants.reject(&:abstract_class?)
-      .select { |c| c.attribute_names.include?("user_id") }
+      .select { |c| c.attribute_names.include?("user_id") }.uniq
```
pkarman: noticed that some of these classes were polymorphic, pointing at the same table.

Issue #9789, opened 2019-03-05
Issues with timeliness exceptions aren't showing up in Queue


shanear:
After migrating all the prior decisions, there are a couple issues remaining that do not have decisions. The thing these issues have in common are that they were untimely, but have exceptions.

It looks like they are not showing up as eligible issues in Queue.

## Technical notes
I believe the reason is because `ineligible_reason` is set, and that's the criteria queue is using to determine whether an issue is eligible.

The affected issues are request issue ids: 4568, 4725

leikkisa:
Issue 4568 is ineligible because it is from before_ama

Issue 4725 has a timeliness_exemption of false, but it does has timeliness_exemption_notes filled out with "This issue was raised in conjunction with the IR claim on appeal in the context of Rice "

There are no request issues with timeliness_exemption set to true that have an ineligible_reason of "untimely".  So there isn't really a bug here, but there are many issues that have timeliness exemption notes, but timeliness exemption set to false.  We are wondering from a usability perspective if this is confusing.

leikkisa:
I would like to close this, but first I want to ask if there are people that we should report these results to, and if there's any desire to ask a product/design person if we should iterate on the design (and if yes, add a ticket for that).

leikkisa:
For the one that has before_ama, if they remove it and re-add it, it should be eligible again for now because we turned off the "before_ama" requirement temporariliy.

carolaponce:
Is part of the issue users writing notes about timeliness even though the veteran has not requested a time extension?

leikkisa:
Yes, we're wondering if it's confusing for folks because there are a bunch of RequestIssues with a note added, but with "no" selected as the option.  This would be relatively easy for me to change too so I can go ahead with it if desired.

![Screen Shot 2019-03-11 at 6.20.00 PM.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5aa9b7c9e7d70200133c2088/125516c66e82c728ace21e0d46db978826878de9bce1b057e40fa0c5700b77206455f26b7191a37ef1102b4f5f5f5ecc3c4a4356bcbbd671f3e57862804e0ee5d33c5d22c33413a5b8d41bbf50bd208b1d87b3bc891ed2f42f67b6d5db1063474f4dcbed5727a19041c95df349e24856a2113765e6732431.png)

leikkisa:
Looks like it could use some spacing above the radio too :-P

[...]

PR #9911, opened 2019-03-12
Only show untimely exemption notes if untimely exemption is selected

leikkisa:
connects #9789


spec/feature/intake/appeal_spec.rb
```diff
@@ -584,8 +582,7 @@ def complete_appeal
     expect(RequestIssue.find_by(
              decision_review: appeal,
              contested_issue_description: "Really old injury",
-             untimely_exemption: false,
-             untimely_exemption_notes: "I am an exemption note"
```
shanear: lol we actually tested that this happened

PR #9912, opened 2019-03-12
Rename open to active, and other refactorings related to request issues

shanear:
connects  #9875


app/models/appeal.rb
```diff
@@ -165,7 +165,7 @@ def reviewing_judge_name
   def eligible_request_issues
     # It's possible that two users create issues around the same time and the sequencer gets thrown off
     # (https://stackoverflow.com/questions/5818463/rails-created-at-timestamp-order-disagrees-with-id-order)
-    open_request_issues.select(&:eligible?).sort_by(&:id)
```
shanear: `active` only returns eligible issues so no more need for `select(&:eligible?)`

app/models/appeal.rb
```diff
@@ -596,11 +596,10 @@ def fetch_docket_type
 
   def docket_switch_deadline
     return unless receipt_date
-    return unless request_issues.open.any?
+    return unless request_issues.active.any?
     return if request_issues.any? { |ri| !ri.closed? && ri.decision_or_promulgation_date.nil? }
 
-    open_request_issues = request_issues.find_all { |ri| !ri.closed? }
```
shanear: @lowellrex I'm not sure what's needed here? Should we only be looking at active, eligible issues, or is it important that we include ineligible issues here too?

app/models/appeal.rb
```diff
@@ -649,7 +648,7 @@ def events
   end
 
   def issues_hash
-    issue_list = decision_issues.empty? ? request_issues.open.select(&:eligible?) : fetch_all_decision_issues
```
shanear: again `select(&:eligible?)` is not needed anymore

app/models/concerns/issue_updater.rb
```diff
@@ -36,7 +36,7 @@ def update_issue_dispositions_in_vacols!
 
   def create_decision_issues!
     issues.each do |issue_attrs|
-      request_issues = appeal.request_issues.open.where(id: issue_attrs[:request_issue_ids])
```
shanear: I believe we only want eligible issues, since they're the only ones that will be receiving decisions (@lowellrex might want to confirm)

app/models/decision_review.rb
```diff
@@ -190,14 +190,9 @@ def contestable_issues
   end
 
   def active_nonrating_request_issues
-    @active_nonrating_request_issues ||= RequestIssue.nonrating.open
+    @active_nonrating_request_issues ||= RequestIssue.nonrating.active
       .where(veteran_participant_id: veteran.participant_id)
       .where.not(id: request_issues.map(&:id))
-      .select(&:status_active?)
```
shanear: redundant, and wrong (since `status_active?` doesn't operate the same as `active?`)

app/models/decision_review.rb
```diff
@@ -190,14 +190,9 @@ def contestable_issues
...
-  end
-
-  def open_request_issues
```
shanear: moved to `request_issues_ui_hash` since that's the only place the includes are needed

app/models/end_product_establishment.rb
```diff
@@ -238,13 +238,9 @@ def generate_tracked_item!
   end
 
   def request_issues
-    return [] unless source.try(:request_issues)
+    return ActiveRecord::NullRelation unless source.try(:request_issues)
```
shanear: Converted this to behave like a Rails relationship. This makes it behave more predictably throughout the app.

app/models/end_product_establishment.rb
```diff
@@ -354,24 +350,12 @@ def fetch_associated_rating
...
   def unassociated_rating_request_issues
-    eligible_rating_request_issues.select { |ri| ri.associated_rating_issue? && ri.rating_issue_associated_at.nil? }
-  end
-
-  def eligible_request_issues
-    open_request_issues.select(&:eligible?)
-  end
-
-  def eligible_rating_request_issues
-    eligible_request_issues.select(&:rating?)
```
shanear: All of these become redundant now that we're using a `ActiveRecord::Relation`

app/models/rating_issue.rb
```diff
@@ -68,20 +68,13 @@ def serialize
...
       ramp_claim_id: ramp_claim_id,
-      title_of_active_review: title_of_active_review,
       rba_contentions_data: rba_contentions_data,
       associated_end_products: associated_end_products.map(&:serialize),
       diagnostic_code: diagnostic_code,
       benefit_type: benefit_type
     }
   end
 
-  def title_of_active_review
```
shanear: unused

app/models/request_issue.rb
```diff
@@ -149,19 +149,17 @@ def nonrating
       ).where.not(issue_category: nil)
     end
 
-    def not_deleted
-      where.not(decision_review_id: nil)
-    end
```
shanear: unused

app/models/serializers/work_queue/decision_review_serializer.rb
```diff
@@ -2,7 +2,7 @@
 
 class WorkQueue::DecisionReviewSerializer < ActiveModel::Serializer
   attribute :issues do
-    object.request_issues.open.map do |issue|
```
shanear: not sure what this is used for but we may want to include active_and_ineligible issues here instead of active issues (which would only be the eligible ones) @lowellrex

app/models/appeal.rb
lowellrex: I'm not very familiar with this code. @aroltsch may know better than I.

app/models/serializers/work_queue/decision_review_serializer.rb
lowellrex: Once again I'm not very familiar with this code so I'll defer to @aroltsch.

app/models/end_product_establishment.rb
```diff
@@ -238,13 +238,9 @@ def generate_tracked_item!
   end
 
   def request_issues
-    return [] unless source.try(:request_issues)
+    return RequestIssue.none unless source.try(:request_issues)
```
pkarman: TIL about `none` !

app/models/request_issue.rb
```diff
@@ -149,19 +149,17 @@ def nonrating
...
-
     def eligible
       where(ineligible_reason: nil)
     end
 
-    def open
+    # "Active" issues are issues that need decisions.
+    # They show up as contentions in VBMS and issues in Caseflow Queue.
```
pkarman: helpful comment!

app/models/appeal.rb
aroltsch: I am not familiar with this code either.

app/models/concerns/issue_updater.rb
aroltsch: Yes. Although theoretically the FE should only be sending eligible request issue IDs. Having extra protection layer on the BE is good.

PR #9913, opened 2019-03-12
Adjust-Options-For-Schedule-Hearing-Admin-Actions-From-Attorneys

### LINKED ISSUE #9558: Adjust options for schedule hearing admin actions from attorneys
When attorneys assign `Schedule hearing` admin actions to VLJ Support, the hearing branch needs VLJ Support to confirm that a hearing actually needs to be scheduled, so they aren't assigned schedule hearing tasks that don't need to be scheduled, and therefore delay the Veteran's appeal.

Related [slack thread](https://dsva.slack.com/archives/C6E41RE92/p1550692146046300).

## AC
- When VLJ Support is assigned a Schedule hearing task, give them the following options in the Actions drop down:
   - `Hearing does not need to be scheduled, send back to attorney`
   - `Hearing does need to be scheduled, send to hearing branch`
- If user selects `Hearing does not need to be scheduled, send back to attorney`, follow current send back to attorney behavior
- If user selects `Hearing does need to be scheduled, send to hearing branch` follow current send to hearing branch behavior 

## Note
- Current send to hearing branch behavior for legacy appeals is changing VACOLS location to 57. We now know that Caseflow Hearing Schedule uses the CASEFLOW location for cases that need a hearing. I don't think we should change the behavior to changing the location to CASEFLOW right now, because I assume that there are additional things that would need to happen in order to replicate what happens during the nightly location 57 --> CASEFLOW location job so the case shows up correctly in Caseflow Hearing Schedule.

youngfreezyVA:
Resolves #9558
progress PR


app/models/tasks/colocated_task.rb
```diff
@@ -73,13 +74,20 @@ def available_actions_with_conditions(core_actions)
 
     core_actions.unshift(Constants.TASK_ACTIONS.COLOCATED_RETURN_TO_ATTORNEY.to_h)
 
+    if action == "schedule_hearing"
```
youngfreezyVA: @aroltsch I'm not sure if BVALSPORER - who is colocated with cases - should also be seeing these new options?

app/models/tasks/colocated_task.rb
```diff
@@ -64,6 +64,7 @@ def available_actions(user)
                                       ])
   end
 
+  # rubocop:disable Metrics/AbcSize
```
aroltsch: This is not good :)

app/models/tasks/colocated_task.rb
aroltsch: Yes, this is for all colocated users. Colocated users == VLJ Support

aroltsch: Although does it apply to legacy and ama or just ama?

client/constants/TASK_ACTIONS.json
```diff
@@ -19,6 +19,8 @@
     "RETURN_TO_JUDGE": { "label": "Return to judge", "value": "modal/assign_to_judge", "func": "assign_to_judge_data" },
     "REVIEW_LEGACY_DECISION": { "label": "Decision ready for review", "value": "draft_decision/special_issues"},
     "REVIEW_AMA_DECISION": { "label": "Decision ready for review", "value": "draft_decision/dispositions"},
+    "SCHEDULE_HEARING_COLOCATED_RETURN_TO_ATTORNEY": { "label": "Hearing does not need to be scheduled, send back to attorney", "value": "modal/mark_task_complete" },
+    "SCHEDULE_HEARING_SEND_TO_TEAM": { "label": "Hearing does need to be scheduled, send to hearing branch", "value": "modal/send_colocated_task" },
```
aroltsch: If the hearing needs to be scheduled, it looks like you need to send it to the hearing team though?

aroltsch: Or am I wrong here?

app/models/tasks/colocated_task.rb
```diff
@@ -73,13 +74,20 @@ def available_actions_with_conditions(core_actions)
 
     core_actions.unshift(Constants.TASK_ACTIONS.COLOCATED_RETURN_TO_ATTORNEY.to_h)
 
+    if action == "schedule_hearing"
+      core_actions.shift
```
aroltsch: Why are you shifting here?

youngfreezyVA: if I don't, there will be two return to attorney options, and there will be an error thrown saying there are two labels with identical paths. it accounts for line 75 which was already there and I didn't want to take it out

client/constants/TASK_ACTIONS.json
youngfreezyVA: I'm not sure, I thought it had to follow the existing `SEND_TO_TEAM` logic.. @laurjpeterson any thoughts?

app/models/tasks/colocated_task.rb
youngfreezyVA: true. gonna change it to this: 
```  if action == "schedule_hearing"
      core_actions = handle_schedule_hearing(core_actions)
    end```

File: app/models/tasks/colocated_task.rb (hunk 1 above)
Comment (youngfreezyVA) [reply app/models/tasks/colocated_task.rb]: legacy. will make the changes

File: client/constants/TASK_ACTIONS.json (hunk 3 above)
Comment (aroltsch) [reply client/constants/TASK_ACTIONS.json]: when `SCHEDULE_HEARING_SEND_TO_TEAM`, we need to display the correct modal and make appropriate BE calls to reassign the task to the hearing team

File: spec/feature/queue/task_queue_spec.rb [hunk 5]
```diff
@@ -419,6 +419,51 @@
...
+
+      it "the new options should appear when a hearing has been scheduled,
+        and the bfcurloc in vacols should be updated to 57" do
+        visit("/queue/appeals/#{appeal.external_id}")
+        find(".Select-control", text: "Select an action…").click
+        expect(page).to have_content(Constants.TASK_ACTIONS.SCHEDULE_HEARING_SEND_TO_TEAM.to_h[:label])
+        find("div", class: "Select-option", text: Constants.TASK_ACTIONS.SCHEDULE_HEARING_SEND_TO_TEAM.label).click
+        find("button", text: "Send case").click
+        expect(page).to have_content("Bob Smith's case has been sent to the Schedule hearing team")
+        expect(vacols_case.reload.bfcurloc).to eq("57")
```
aroltsch: Nice!

spec/feature/queue/task_queue_spec.rb
```diff
@@ -419,6 +419,51 @@
       expect(page).to_not have_content(judgeteam.name)
     end
   end
+  describe "VLJ support staff schedule hearing action" do
+    let!(:attorney) { FactoryBot.create(:user) }
```
aroltsch: You don't need to include the bang operator on all factories. If you leave it out here, the attorney object won't be created until it is called on line 425. You only want to include bang if you need an object to be created immediately.

spec/feature/queue/task_queue_spec.rb
```diff
@@ -419,6 +419,51 @@
...
+        visit("/queue/appeals/#{appeal.external_id}")
+        find(".Select-control", text: "Select an action…").click
+        expect(page).to have_content(Constants.TASK_ACTIONS.CANCEL_TASK.to_h[:label])
+        expect(page).to have_content(Constants.TASK_ACTIONS.SCHEDULE_HEARING_SEND_TO_TEAM.to_h[:label])
+        find("div", class: "Select-option", text: Constants.TASK_ACTIONS.CANCEL_TASK.label).click
+        find("button", text: "Submit").click
+        expect(page).to have_content("Task for Bob Smith's case has been cancelled")
+        User.authenticate!(user: attorney)
+        visit("/queue")
+        expect(page).to have_content(appeal.veteran_file_number)
```
aroltsch: Nice!

spec/feature/queue/task_queue_spec.rb
```diff
@@ -419,6 +419,51 @@
...
+    let!(:vlj_support_staffer) { FactoryBot.create(:user) }
+
+    before do
+      OrganizationsUser.add_user_to_organization(vlj_support_staffer, Colocated.singleton)
+      User.authenticate!(user: vlj_support_staffer)
+    end
+
+    context "when a ColocatedTask has been assigned through the Colocated organization to an individual" do
+      before do
+        ColocatedTask.create_many_from_params([{
```
aroltsch: Why didn't you use a factory to create that?

spec/feature/queue/task_queue_spec.rb
youngfreezyVA: ahh ok understood!

spec/feature/queue/task_queue_spec.rb
youngfreezyVA: I followed the same pattern in this test on line 394

app/models/tasks/colocated_task.rb
```diff
@@ -86,6 +84,31 @@ def actions_available?(_user)
...
+  def legacy_translation_or_hearing_actions(actions)
+    return legacy_schedule_hearing_actions(actions) if action == "schedule_hearing"
+
+    legacy_translation_actions(actions)
+  end
+
+  def legacy_schedule_hearing_actions(actions)
+    task_actions = Constants.TASK_ACTIONS
+    # remove the assign to privacy team option
+    actions.delete_at(1)
```
aroltsch: Deleting at a specific index could break things if another developer adds another action without realizing. Also,  removing the assign to privacy team option - did the design team ask?

spec/feature/queue/task_queue_spec.rb
```diff
@@ -419,6 +419,51 @@
...
+        and the bfcurloc in vacols should be updated to 57" do
+        visit("/queue/appeals/#{appeal.external_id}")
+        find(".Select-control", text: "Select an action…").click
+        expect(page).to have_content(Constants.TASK_ACTIONS.SCHEDULE_HEARING_SEND_TO_TEAM.to_h[:label])
+        find("div", class: "Select-option", text: Constants.TASK_ACTIONS.SCHEDULE_HEARING_SEND_TO_TEAM.label).click
+        find("button", text: "Send case").click
+        expect(page).to have_content("Bob Smith's case has been sent to the Schedule hearing team")
+        expect(vacols_case.reload.bfcurloc).to eq("57")
+      end
+      it "the case should be returned in the attorneys queue when canceled" do
```
aroltsch: We usually add one empty line break between "it" blocks

spec/feature/queue/task_queue_spec.rb
```diff
@@ -419,6 +419,51 @@
...
+    context "when a ColocatedTask has been assigned through the Colocated organization to an individual" do
+      before do
+        ColocatedTask.create_many_from_params([{
+                                                assigned_by: attorney,
+                                                action: :schedule_hearing,
+                                                appeal: appeal
+                                              }], attorney)
+      end
+
+      it "the new options should appear when a hearing has been scheduled,
```
aroltsch: This 'it" description is too long and not 100% accurate. The location is updated to 57 when a user assigns it back to the hearing team

app/models/tasks/colocated_task.rb
youngfreezyVA: ah good point. ill remove it by label or something. yes, in the issue lauren specified the three options that should be there, and didn't mention assign to privacy team: https://github.com/department-of-veterans-affairs/caseflow/issues/9558#issuecomment-476803266

aroltsch: could you double check with LP please

Issue #9307, opened 2019-02-08
Refactor attorney and judge flows FE


aroltsch:
AC: 
Get rid of `EditModalBase` and `DecisionReviewBase`

youngfreezyVA:
I agree, I think we should also think about having all URL actions/mappings in one place, having the urls passed around in different components/places can be hard to track if we ever want to change the UI flow. so passing in `nextPage=urlMappings.flowPage.next`

PR #9915, opened 2019-03-12
[WIP] Refactor EditModalBase

hschallhorn:
References #9307

### Description
Removes the added complexity of having EditModalBase be a higher order component

### Acceptance Criteria
- All modals that used EditModalBase function as perviously intended while now using the updated QueueFlowModal


client/app/queue/AddJudgeTeamModal.jsx
```diff
@@ -13,6 +12,7 @@ import {
   showErrorMessage
 } from './uiReducer/uiActions';
 import { withRouter } from 'react-router-dom';
+import QueueFlowModal from './components/QueueFlowModal';
 
 class AddJudgeTaskModal extends React.Component {
```
annekainicUSDS: This is very minor and not an issue that was addressed in this PR, but I noticed that the name of this file (`AddJudgeTeamModal`) is different from the name of the component exported from it (`AddJustTaskModal`). Do we want to make those consistent?

hschallhorn: Nice catch. I see no reason not to update it now.

PR #9917, opened 2019-03-12
Fix IDT document uploads

monfresh:
**Why**: The background job was receiving as a single argument an object
that doesn't persist the `file` attribute. Once the job was triggered,
`document.file` would be `nil`, causing `Base64.decode64` in
`UploadDocumentToVbms#cache_file!` to throw an exception.

**How**: Pass in the file explicitly to the Job so that it can pass it
to the instance of `UploadDocumentToVbms`. In general, it's recommended
to pass simple attributes to background jobs like strings. As a further
improvement, we should probably pass in the primary key ID of the
document and then look it up in the `UploadDocumentToVbms` class.


app/jobs/upload_document_to_vbms_job.rb
```diff
@@ -4,10 +4,10 @@ class UploadDocumentToVbmsJob < CaseflowJob
   queue_as :low_priority
   application_attr :idt
 
-  def perform(document)
+  def perform(document:, file:)
```
aroltsch: Ah, good catch!

Issue #9733, opened 2019-03-01
Intake Edit | Update save warning when removing entire decision review


shanear:
Users need to be warned when they are removing an entire decision review.

## Acceptance Criteria
- On the intake edit page, when a user clicking save would remove the entire review (according to #9730 and #9731) instead of the Number of Issues Changed warning, show the appropriate warning from the mocks below based on whether the decision review is processed in Caseflow or VBMS.

## Mocks
Processed in Caseflow:
![remove_review_appeal](https://user-images.githubusercontent.com/1596259/53656296-be7fe080-3c07-11e9-8b89-df32e157cbee.png)

Processed in VBMS:
![remove_review_hlr-sc](https://user-images.githubusercontent.com/1596259/53656302-c3449480-3c07-11e9-815c-d4bf04b68156.png)

## Notes
- The copy is subject to change per @carolaponce
- This functionality should be enabled and disabled using a feature flag named `:remove_decision_reviews`

Sjones352:
**Validations Vbms**
- Create a higher level review or supplemental claim with a benefit type of compensation/pension
- Go to the Edit page
- Remove all decision reviews
- click on save
- Modal shows message `This will remove the review and cancel all the End Products associated with it.` 

**Validations Caseflow**
- Create a higher level review or supplemental claim with a benefit type of non-comp
- Go to the Edit page
- Remove all decision reviews
- click on save
- Modal shows message `This review and all tasks associated with it will be removed.`

pkarman:
Validated for compensation HLR:

![Screen Shot 2019-03-26 at 3 33 33 PM](https://user-images.githubusercontent.com/1205061/55031421-986a1800-4fdc-11e9-8697-f51cb764b246.png)

pkarman:
Validated for non-comp in Caseflow:

![Screen Shot 2019-03-26 at 3 35 44 PM](https://user-images.githubusercontent.com/1205061/55031572-df580d80-4fdc-11e9-90a7-d6a50edf7e18.png)

PR #9920, opened 2019-03-12
Remove modal

Sjones352:
Connects #9733
Processed in VBMS:
Screen Shot 2019-03-16 at 12 00 15 AM

Processed in Caseflow:
Screen Shot 2019-03-18 at 3 13 25 PM


client/app/intakeEdit/components/EditButtons.jsx
```diff
@@ -29,11 +30,16 @@ class SaveButtonUnconnected extends React.Component {
...
     let showModals = {
       issueChangeModal: false,
-      unidentifiedIssueModal: false
+      unidentifiedIssueModal: false,
+      reviewRemovedModal: false
     };
 
     if (this.state.originalIssueNumber !== this.props.state.addedIssues.length) {
-      showModals.issueChangeModal = true;
+      if (_.isEmpty(this.props.state.addedIssues.length)) {
```
pkarman: I think you can either use `_.isEmpty(this.props.state.addedIssues)` or `!this.props.state.addedIssues.length` (no isEmpty). The `.length` attribute should never be "empty" afaik but it could be zero.

client/app/intakeEdit/components/EditButtons.jsx
```diff
@@ -109,6 +115,18 @@ class SaveButtonUnconnected extends React.Component {
...
+      { this.state.showModals.reviewRemovedModal && <SaveAlertConfirmModal
+        title="Remove review?"
+        buttonText= "Yes, remove"
+        onClose={() => this.closeModal('reviewRemovedModal')}
+        onConfirm={() => this.confirmModal('reviewRemovedModal')}>
+        <p>
+          The review originally had {this.state.originalIssueNumber}&nbsp;
+          { pluralize('issue', this.state.originalIssueNumber) } but now has {this.props.state.addedIssues.length}.
+        </p>
+        <p>This will remove the review and cancel all the End Products associated with it.</p>
```
pkarman: this text will only apply to claim reviews (HLR and SC) since they are the only ones with End Products processed in VBMS. I think maybe this code should check the `formType`.

pkarman: The ticket describes using `:remove_decision_reviews` feature flag to gate this messaging. Is that already in the client side code but not in this PR?

Sjones352: Yes this is already in the client side. In order to test it please enable Feature toggle in console 
`FeatureToggle.enable!(:remove_decision_reviews)`

client/app/intakeEdit/components/EditButtons.jsx
```diff
@@ -103,6 +103,22 @@ class SaveButtonUnconnected extends React.Component {
 
     const saveDisabled = _.isEqual(addedIssues, originalIssues) || disableDueToIssueCount;
 
+    let removeReviewModal = () => {
```
pkarman: I like making this a method.

Should there be an "else" condition here for the alternate language? Or is that coming in a follow-on PR?

Sjones352: There is no else statement as this function warns users when they are removing an entire decision review. They also have an option to cancel in the modal should not want to remove the entire decision review

pkarman: Sorry, I wasn't clear.

The issue #9733 has 2 mocks for 2 different messages. This PR covers the VBMS case. I'm asking about the Caseflow-only case.

Sjones352: Decided to check by benefit_type.

spec/feature/intake/appeal/edit_spec.rb
```diff
@@ -391,7 +391,7 @@
              closed_at: last_week)
     end
 
-    context "when review has multiple active tasks" do
+    fcontext "when review has multiple active tasks" do
```
pkarman: focus to be removed

client/app/intakeEdit/components/EditButtons.jsx
```diff
@@ -103,24 +104,35 @@ class SaveButtonUnconnected extends React.Component {
...
-            { pluralize('issue', this.state.originalIssueNumber) } but now has {this.props.state.addedIssues.length}.
-          </p>
-          <p>This will remove the review and cancel all the End Products associated with it.</p>
-        </SaveAlertConfirmModal>;
+    const removeReviewCopy = () => {
+      const removeVbmsCopy = 'This will remove the review and cancel all the End Products associated with it.';
+
+      const removeCaseflowCopy = 'This review and all tasks associated with it will be removed.';
+
+      if (VBMS_BENEFIT_TYPES.includes(benefitType)) {
```
pkarman: benefit type is set per-issue on Appeals, so I recommend checking the `processedInCaseflow` prop which should be set on all decision reviews. That should be all you need to check.

client/app/intakeEdit/components/EditButtons.jsx
```diff
@@ -103,24 +104,35 @@ class SaveButtonUnconnected extends React.Component {
...
-          buttonText= "Yes, remove"
-          onClose={() => this.closeModal('reviewRemovedModal')}
-          onConfirm={() => this.confirmModal('reviewRemovedModal')}>
-          <p>
-          The review originally had {this.state.originalIssueNumber}&nbsp;
-            { pluralize('issue', this.state.originalIssueNumber) } but now has {this.props.state.addedIssues.length}.
-          </p>
-          <p>This will remove the review and cancel all the End Products associated with it.</p>
-        </SaveAlertConfirmModal>;
+    const removeReviewCopy = () => {
```
pkarman: does this need to be a closure? Could it just be a string?

client/app/intakeEdit/components/EditButtons.jsx
```diff
@@ -114,6 +137,18 @@ class SaveButtonUnconnected extends React.Component {
...
+      {this.state.showModals.reviewRemovedModal && <SaveAlertConfirmModal
+        title="Remove review?"
+        buttonText= "Yes, remove"
+        onClose={() => this.closeModal('reviewRemovedModal')}
+        onConfirm={() => this.confirmModal('reviewRemovedModal')}>
+        <p>
+          The review originally had {this.state.originalIssueNumber}&nbsp;
+          { pluralize('issue', this.state.originalIssueNumber) } but now has {this.props.state.addedIssues.length}.
+        </p>
+        <p>{removeReviewCopy()}</p>
```
leikkisa: If you move the copy variables out of that function, and add the processedInCaseflow from the props (around line 98) then here you could put something like:
``` { processedInCaseflow ? removeCaseflowCopy : removeVbmsCopy }```

And you wouldn't need the copy function any more, since now you're using processedInCaseflow instead of having to calculate that.

File: spec/feature/intake/appeal/edit_spec.rb [hunk 8]
```diff
@@ -488,6 +488,19 @@
         expect(in_progress_task.reload.status).to eq(Constants.TASK_STATUSES.cancelled)
       end
 
+      scenario "remove all decsions" do
```
leikkisa: Typo here (and below in the test with the same name)

client/app/intakeEdit/components/EditButtons.jsx
```diff
@@ -29,11 +30,16 @@ class SaveButtonUnconnected extends React.Component {
...
     let showModals = {
       issueChangeModal: false,
-      unidentifiedIssueModal: false
+      unidentifiedIssueModal: false,
+      reviewRemovedModal: false
     };
 
     if (this.state.originalIssueNumber !== this.props.state.addedIssues.length) {
-      showModals.issueChangeModal = true;
+      if (this.props.state.addedIssues.length === 0) {
```
leikkisa: This might be a good place to put in a check for the removeDecisionReviews feature toggle

spec/feature/intake/appeal/edit_spec.rb
```diff
@@ -488,6 +488,19 @@
...
+        visit "appeals/#{appeal.uuid}/edit"
+        # remove all request issues
+        appeal.request_issues.length.times do
+          click_remove_intake_issue(1)
+          click_remove_issue_confirmation
+        end
+
+        click_edit_submit
+        expect(page).to have_content("Remove review?")
```
leikkisa: I would add a check for the "processed in caseflow" specific text here as well

spec/feature/intake/higher_level_review/edit_spec.rb
```diff
@@ -1158,6 +1158,22 @@ def click_cancel(visit_page)
...
+      scenario "remove all decsions" do
+        visit "higher_level_reviews/#{higher_level_review.uuid}/edit"
+        # remove all request issues
+        higher_level_review.request_issues.length.times do
+          click_remove_intake_issue(1)
+          click_remove_issue_confirmation
+        end
+
+        click_edit_submit
+        expect(page).to have_content("Remove review?")
```
leikkisa: If this review is processed_in_vbms would add a check for the specific copy that it should have. Also, I think it would be good to add a test for a higher level review that is processed_in_caseflow  (or whatever scenario this test isn't).

spec/feature/intake/appeal/edit_spec.rb
Sjones352: @leikkisa I thought that appeals are processed in Caseflow so maybe that test would not be necessary.

leikkisa: Yes, the test just confirms the expected behavior.

You could replace expect(page).to have_content("Remove review?") with expect(page).to have_content("This review and all tasks associated with it will be removed.") since that's more specific.

spec/feature/intake/higher_level_review/edit_spec.rb
leikkisa: I think it would be good to add a test for the processedInCaseflow case somewhere also, where you just set the benefitType to something processed in Caseflow.

Issue #9916, opened 2019-03-12
Bold new tasks in all queues


lowellrex:
Currently, we distinguish between tasks where somebody has visited the case details page and tasks where they have not [only for members of the VLJ support staff](https://github.com/department-of-veterans-affairs/caseflow/blob/65bfc1f3a3d33ada7136bd290ce4452bd709bbd9/client/app/queue/CaseDetailsLink.jsx#L46). This ticket exists to bold/unbold links to the case details page for an appeal in the task queue for everybody. This would likely require us to change from using the `status` field of the task to something more involved (perhaps a `task_views` table linking a user to a task).

## Acceptance criteria
* [ ] Show links to all cases in bold if they have not been viewed yet by the current user

hschallhorn:
Closed in #14277

PR #9923, opened 2019-03-12
Fix action forbidden error

### LINKED ISSUE #9613: Stop throwing a Caseflow::Error::ActionForbiddenError when attorneys visit ColocatedTask pages
Because of [how we build the list of cases](https://github.com/department-of-veterans-affairs/caseflow/blob/cddcaf4f0aa1837e45a15defc31db66078027da9/app/models/queues/attorney_queue.rb#L10) in an attorney's queue, we [occasionally get `Caseflow::Error::ActionForbiddenError`s](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3678/events/251695/) when an attorney visits the case details page for a `LegacyAppeal` attached to a `ColocatedTask` they created before the assignee visits the case details page ([when that happens](https://github.com/department-of-veterans-affairs/caseflow/blob/cddcaf4f0aa1837e45a15defc31db66078027da9/client/app/queue/CaseDetailsLink.jsx#L19) the status of the `ColocatedTask` would change from `assigned` to `in_progress`). We should not attempt to change the status of these tasks when attorneys visit these tasks since they are not assigned to them as an individual.

## Acceptance criteria
* [ ] Do not attempt to change the status of a task that is not assigned to the current user.
* [ ] Do not throw a `Caseflow::Error::ActionForbiddenError` when an attorney visits the case details page for a `LegacyAppeal` attached to a `ColocatedTask` they created.

tomas-nava:
Resolves #9613

### Description
Only attempt to change a task's status to `in_progress` if the user is colocated, because they're the only users who see bold links in the queue for cases they haven't visited yet. This is a temporary fix to silence errors until we tackle #9916


spec/feature/queue/case_details_spec.rb
```diff
@@ -449,34 +449,55 @@
...
+          assigned_to: colocated_user,
+          assigned_by: attorney_user
+        )
+      end
+
+      scenario "displays task bold in queue" do
+        visit "/queue"
+        vet_name = assigned_task.appeal.veteran_full_name
+        expect(page).to have_content("#{vet_name.split(' ').first} #{vet_name.split(' ').last}")
+        # how to tell if the link is bold?
```
lowellrex: It looks like [capybara has a `style()` method](https://www.rubydoc.info/github/jnicklas/capybara/master/Capybara/Node/Element#style-instance_method) that we might be able to use.

Alternatively, we could just check that the status changed (and doesn't for attorneys viewing `ColocatedTask`s) and assume that we will be using that information to bold/unbold the link or check to see if the function the endpoint calls is called (`expect_any_instance_of(TasksController).to receive(:update).exactly(1).times` or something like that).

PR #9927, opened 2019-03-13
Upgrade Ruby from 2.5.1 to 2.5.3

monfresh:


.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ jobs:
       # The first image listed is the image that all steps run inside of.
       # This can be modified by steps, or can be built as a separate
       # customized container.
-      - image: 008577686731.dkr.ecr.us-gov-west-1.amazonaws.com/circleci:ruby-2.5.1
+      - image: 008577686731.dkr.ecr.us-gov-west-1.amazonaws.com/circleci:ruby-2.5.3-node-browsers
```
nemodjuric: Image URI should be 008577686731.dkr.ecr.us-gov-west-1.amazonaws.com/circleci:2.5.3-node-browsers

monfresh: Yeah. I didn't change the lint and js_tests parts yet. The rspec job is set to the correct image, but it's still failing. I will change the others as well to see if it makes a difference.

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ jobs:
       # The first image listed is the image that all steps run inside of.
       # This can be modified by steps, or can be built as a separate
       # customized container.
-      - image: 008577686731.dkr.ecr.us-gov-west-1.amazonaws.com/circleci:ruby-2.5.1
+      - image: 008577686731.dkr.ecr.us-gov-west-1.amazonaws.com/circleci:2.5.3-node-browsers
```
enriquemanuel: my only concern with this is that now we are breaking naming convention (if we ever had one)

.circleci/config.yml
```diff
@@ -150,7 +150,7 @@ jobs:
...
```
enriquemanuel: my only concern with this is that now we are breaking naming convention (if we ever had one)

.circleci/config.yml
```diff
@@ -212,7 +212,7 @@ jobs:
...
```
enriquemanuel: my only concern with this is that now we are breaking naming convention (if we ever had one)

Issue #9800, opened 2019-03-05
Intake | Remove non-existent contention error


pkarman:
See https://dsva.slack.com/archives/CCB97HZJL/p1551815313002300

We get a 'NoMethodError` Ruby exception when we attempt to remove a contention that does not exist.

There are multiple ways forward on this and it is unclear why this is happening.
 
It might be possible to edit a claim review multiple times and queue multiple RequestIssuesUpdate jobs to remove the same contention.

We should flag issues that have been queued for removal so that users cannot "remove" them twice.

pkarman:
One example:

A SC was created. An unidentified request issue was removed and another added. Then the 2nd unidentified issue was replaced with a rating request issue.

```
[#<RequestIssuesUpdate:0x00005596813d1068
  id: 132,
  user_id: 5144,
  review_type: "SupplementalClaim",
  review_id: 802,
  before_request_issue_ids: [7453],
  after_request_issue_ids: [9671],
  processed_at: Fri, 15 Mar 2019 19:39:56 UTC +00:00,
  attempted_at: Fri, 15 Mar 2019 19:39:50 UTC +00:00,
  submitted_at: Fri, 15 Mar 2019 19:39:50 UTC +00:00,
  error: nil,
  last_submitted_at: Fri, 15 Mar 2019 19:39:50 UTC +00:00>,
 #<RequestIssuesUpdate:0x00005596813d0f28
  id: 133,
  user_id: 5144,
  review_type: "SupplementalClaim",
  review_id: 802,
  before_request_issue_ids: [9671],
  after_request_issue_ids: [9673],
  processed_at: nil,
  attempted_at: Mon, 18 Mar 2019 19:52:59 UTC +00:00,
  submitted_at: Fri, 15 Mar 2019 19:54:37 UTC +00:00,
  error: nil,
  last_submitted_at: Fri, 15 Mar 2019 19:54:37 UTC +00:00>]
```

The 2nd unidentified request issue (9671) re-used the removed contention id assigned to the 1st unidentified request issue (7453). Then when we attempted to remove 9671, because it was pointing to a non-existent contention, we get the error.

pkarman:
fixed by #10041

PR #9931, opened 2019-03-13
When VBMS reports contention does not exist, raise error

pkarman:
connects #9800 

### Description

If we attempt to remove a contention using a reference_id that VBMS cannot find, we should throw an error.


app/models/request_issues_update.rb
```diff
@@ -50,7 +50,12 @@ def establish!
 
     potential_end_products_to_remove = []
     removed_issues.select(&:end_product_establishment).each do |request_issue|
-      request_issue.end_product_establishment.remove_contention!(request_issue)
+      begin
+        request_issue.end_product_establishment.remove_contention!(request_issue)
+      rescue EndProductEstablishment::ContentionNotFound => _err
```
pkarman: alternately, we could report the error to Sentry if we wanted to track the condition.

leikkisa:
I like giving it a more meaningful error.

I am a little bit nervous about "we should remove the request issue".  I looked into one instance of this, and the end product had three contentions on it, and the Decision Review had three request issues on it, but only one of them matched.  I'm worried removing the request issue will obfuscate whatever the real problem is.  If we are going to give them a closed status, I think it should be different than if they are removed by the user.

I would love to look into more of them I'm just having trouble searching for them on Sentry.

leikkisa:
Yeah, the two removed request issues have the same contention_reference_id, which shouldn't be possible, so I think there's a bug in our code somewhere.

PR #9935, opened 2019-03-13
Set ineligible issues as closed (also some refactoring)

shanear:
connects #9875

A follow up PR to migrate all the old data will be coming. For now the code works for both closed and "open" ineligible issues.


app/models/request_issue.rb
```diff
@@ -2,17 +2,6 @@
 
 # rubocop:disable Metrics/ClassLength
 class RequestIssue < ApplicationRecord
-  # TODO: remove this eventually, used to protect caching from screwing up removed columns
```
shanear: figured now was as good of time as ever to get rid of this.

app/models/request_issue.rb
```diff
@@ -387,30 +378,35 @@ def legacy_issue_opted_in?
     eligible? && vacols_id && vacols_sequence_id
   end
 
-  def remove!
-    update!(closed_at: Time.zone.now, closed_status: :removed)
+  def close!(status)
```
shanear: factors a lot of the close logic into one place

spec/models/request_issue_spec.rb
```diff
@@ -71,7 +72,8 @@
       vacols_id: vacols_id,
       vacols_sequence_id: vacols_sequence_id,
       closed_at: closed_at,
-      closed_status: closed_status
+      closed_status: closed_status,
+      ineligible_reason: ineligible_reason
```
shanear: decided to speckle some validations into the feature tests for extra coverage.

app/models/request_issue.rb
leikkisa: Nice

Issue #9832, opened 2019-03-07
Create nightly job to update DispositionTask if hearing disposition changed


lowellrex:
Judges may change the disposition of a hearing multiple times on the day of the hearing, but we expect (and maybe don't even allow) them to not change the disposition of the hearing after the day is over. In order to move the appeal forward in the progress we will run a nightly job that will take the appropriate action (using the set of [actions we've already created](https://github.com/department-of-veterans-affairs/caseflow/issues/7556#issuecomment-467008702)) on active `DispositionTask`s related to hearings that have been held and have non-`null` dispositions.

## Acceptance criteria:
* [x] Create a nightly job that takes the appropriate action on `DispositionTask`s for hearings *that were not held that day* with non-null dispositions
* [x] Run this job nightly! (https://github.com/department-of-veterans-affairs/appeals-deployment/pull/2011)

See the section on the `DispositionTask` in the [main reference document](https://github.com/department-of-veterans-affairs/appeals-design-research/blob/master/Projects/Hearings/Hearing%20Schedule/Tech%20Specs/HearingTaskFlow.md#disposition-task) for more information.

lowellrex:
MEDIUM. Reference ticket #7905 for a recently created nightly job.

PR #9942, opened 2019-03-13
Create nightly job for DispositionTask when disposition changes

lowellrex:
Connects #9832.


app/jobs/hearing_disposition_change_job.rb
```diff
@@ -0,0 +1,95 @@
...
+    log_info(start_time, task_count_for, error_count)
+  rescue StandardError => e
+    log_info(start_time, task_count_for, error_count, e)
+  end
+  # rubocop:enable Metrics/PerceivedComplexity
+  # rubocop:enable Metrics/MethodLength
+  # rubocop:enable Metrics/CyclomaticComplexity
+  # rubocop:enable Metrics/AbcSize
+
+  def log_info(start_time, task_count_for, error_count, err = nil)
```
tomas-nava: I think we should log hearing (and maybe task?) IDs, not just counts. Whether it's here or via independent `Rails.logger` calls in the flow above.

app/jobs/hearing_disposition_change_job.rb
```diff
@@ -0,0 +1,95 @@
...
+            task_count_for[:held_in_past_24_hours] += 1
+          end
+        else
+          # Expect to never reach this block since all dispositions should be accounted for above. If we run into this
+          # case we ignore it and will investigate and potentially incorporate that fix here. Until then we're fine.
+          task_count_for[:unknown_dispositions] += 1
+        end
+      end
+
+    rescue StandardError => e
```
tomas-nava: What kinds of errors are we anticipating here?

app/jobs/hearing_disposition_change_job.rb
```diff
@@ -0,0 +1,95 @@
...
+          task.mark_no_show!
+          task_count_for[hearing.disposition] += 1
+        when nil
+          # We allow judges and hearings staff 2 days to make changes to the hearing's disposition. If it has been more
+          # than 2 days since the hearing was held and there is no disposition then remind the hearings staff.
+          if hearing.scheduled_for < 48.hours.ago
+            # Logic will be added as part of #9833.
+            task_count_for[:stale] += 1
+          else
+            task_count_for[:held_in_past_24_hours] += 1
```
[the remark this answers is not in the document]
lowellrex: Great catch. Will rename

app/jobs/hearing_disposition_change_job.rb
```diff
@@ -0,0 +1,95 @@
...
+      held_in_past_24_hours: 0,
+      stale: 0,
+      unknown_dispositions: 0
+    }
+
+    # Set user to system_user to avoid sensitivity errors
+    RequestStore.store[:current_user] = User.system_user
+
+    # This is inefficient. If it runs slowly or consumes a lot of resources then refactor. Until then we're fine.
+    DispositionTask.active.where.not(status: Constants.TASK_STATUSES.on_hold).each do |task|
```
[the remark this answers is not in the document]
lowellrex: We assume this is true, but we should almost certainly fail here in the code if that is not true. Great catch.

app/jobs/hearing_disposition_change_job.rb
lowellrex: Any and all! One that I foresee us running into is a `null` hearing_task association, though I guess we can just protect against that in the code.

app/models/tasks/disposition_task.rb
```diff
@@ -77,6 +77,10 @@ def update_legacy_appeal_location
   end
 
   def hearing_disposition
-    parent&.hearing_task_association&.hearing&.disposition
+    hearing&.disposition
+  end
+
+  def hearing
+    parent&.hearing_task_association&.hearing
```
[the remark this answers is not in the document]
lowellrex: Cool! I'll definitely take advantage of that when that merges to master.

app/jobs/hearing_disposition_change_job.rb
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    # Set user to system_user to avoid sensitivity errors
+    RequestStore.store[:current_user] = User.system_user
+
+    # This is inefficient. If it runs slowly or consumes a lot of resources then refactor. Until then we're fine.
+    tasks = DispositionTask.active.where.not(status: Constants.TASK_STATUSES.on_hold).select do |t|
+      t.hearing && (t.hearing.scheduled_for < 24.hours.ago)
+    end
+    hearing_ids = tasks.map { |t| t.hearing.id }
```
lowellrex: It looks like we have a [`has_one` association](https://github.com/department-of-veterans-affairs/caseflow/blob/97fde32d60c9fc2ad64d22c1655f04931a2da655/app/models/hearing.rb#L10) between a `Hearing` and a `HearingTask`. I don't know how Rails enforces this association and it [doesn't look like a constraint on the database exists](https://github.com/department-of-veterans-affairs/caseflow/blob/97fde32d60c9fc2ad64d22c1655f04931a2da655/db/schema.rb#L517) to enforce this, so I'd be happy to add more conditions that confirm we don't end up in this situation if you thinks it's better to be more defensive here (I am wary about how much we should trust ActiveRecord to protect us here).

lowellrex: I guess we could just add that database constraint... That's probably the thing to do.

app/jobs/hearing_disposition_change_job.rb
```diff
@@ -0,0 +1,99 @@
...
+
+    msg = "#{self.class.name} #{result} after running for #{duration}."
+    task_count_for.keys.each do |k, v|
+      msg += "Processed #{v} #{k.to_s.humanize} hearings"
+    end
+    msg += " Encountered errors for #{error_count} hearings."
+    msg += " Fatal error: #{err.message}" if err
+
+    Rails.logger.info(msg)
+    Rails.logger.info(hearing_ids)
```
lowellrex: Andrew informs me that we expect on the order of 500 hearings to be held every day so I'd rather just log this to the system logs (where we can get at it in Kibana and AWS logs if we want to) so that we don't end up spamming the appeals-app-alerts channel. Open to other solutions here (maybe a database table that's just an audit trail? or literally anything else).

app/jobs/hearing_disposition_change_job.rb
```diff
@@ -0,0 +1,98 @@
+# frozen_string_literal: true
+
+require "action_view"
+
+class HearingDispositionChangeJob < CaseflowJob
+  # For time_ago_in_words()
+  include ActionView::Helpers::DateHelper
+  queue_as :low_priority
+
+  # rubocop:disable Metrics/MethodLength
```
monfresh: It looks like the Danger rule did not catch this. We added a rule to check for new instances of disabling Rubocop rules because we agreed that we don't want to disable them anymore, and refactor the code instead. Code Climate as a "grep" engine that I plan on trying out to catch these kinds of things.

lowellrex:
Explored an alternative approach using a class to accomplish the same thing that the procedural code does now in #10064 in case you're interested in going that route.

app/jobs/hearing_disposition_change_job.rb
tomas-nava: I think this is fine!

app/jobs/hearing_disposition_change_job.rb
tomas-nava: Let's do this on a separate ticket.

tomas-nava: [ticketed here](https://github.com/department-of-veterans-affairs/caseflow/issues/10152)

Issue #9946, opened 2019-03-13
Intake | RAMP EndProductNotFound error needs EP assignment


pkarman:
e.g. https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3992/events/270606/

When a new RAMP refiling Intake is started, there is an initial check to see if all existing EPs are active. If the `EndProductEstablishment` for an existing RampElection row has a `nil` `reference_id` then it will raise a `EstablishedEndProductNotFound` error.

Before raising that error, we should try and check for an existing EP matching the date and code the veteran and assign it to the EPE. It may be that there are orphans on the Caseflow side and orphans on the VBMS side that merely need to be connected.

### Tech notes

* Could either rescue the exception in https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/ramp_review.rb#L85 or let EPE `.sync!` do the matching attempt itself. Probably better to do it in RampReview since it is RAMP specific.

leikkisa:
We have not fully diagnosed the problem, but here is my attempt at summarizing what we know so far.

There are 2463 established RAMP Elections that have a successful intake, but that are missing an End Product Establishment.

1668 Are not connected to an existing end product. The last one of these was intaken on.  We believe these are from before using end product establishments and were not backfilled.

795 are connected to an existing end product, and so should not have an end product establishment.  There was previously a bug in the code which could lead to these being connected to a canceled end product.  This bug was fixed in late February, allowing only active end products to be connected.  The instances of the missing end product establishments seem to have stopped at that time.

58 of them have multiple intakes, which seems to be caused by the RampElection being rolled back, and intaken again.

leikkisa:
#### Validation
This is going to be difficult to manually validate, because in UAT there are no RampElections missing an EndProductEstablishment, and the code doesn't allow new ones to be created like that.

So I suggest we validate it by looking for this error in Slack. Let's take a look at it after Wednesday, April 17th and see if there were any recent instances of this.

Sjones352:
Looked through the alert slack messages since April  17th and I see no errors so closing the issue.

Issue #9623, opened 2019-02-22
Intake | End Product Description is nil


pkarman:
As reported in https://github.com/department-of-veterans-affairs/appeals-support/issues/4131 it is possible for the `end_product_description` value to be `nil` on a RAMP Review.

![image](https://user-images.githubusercontent.com/44010416/52745029-226d9c80-2fac-11e9-8d87-3302d2b6fc33.png)

## Acceptance criteria

* When an End Product Establishment description is nil, do not display the `Established EP` message on RAMP complete screen.

## Technical notes

* It's not clear if a `nil` description is valid from a business perspective. This will require some investigation. In the support ticket cited above, the data all looked correct in VBMS and VACOLS.

leikkisa:
There was a RAMP Election successfully processed for this Vet on 2018-03-15, a RAMP Refiling on 2019-02-12.  All other intakes attempted on this Veteran have had errors, mostly no_active_appeals.  So the user shouldn't have even gotten to this page in the first place.

The code looks pretty good to me, but our tests for it aren't very good so I think that is a good place to start.

leikkisa:
I'm finding that the Ramp Elections for this user on that day that are missing an end_product_description are either canceled or have an error of "connected_preexisting_ep". If there's connected_preexisting_ep, there is a still a RampElection that says it was established, and there's no End Product Establishment, but if you call End Product Establishment on that Ramp Election it creates a temporary one. However, that EPE doesn't have an End Product, so when it tries to get the description it returns nil.

I think it's bad for reporting that we're reporting the errored intakes as established Ramp Elections.

leikkisa:
Confirmed that the current code shouldn't have this problem, because it's saving the EndProductEstablishment to the database.

pkarman:
By "current code" you mean the problem was fixed sometime after the reported bug?

leikkisa:
#### Validation instructions
- Create a RAMP Election Opt-in for a Veteran
- Create another RAMP election Opt-in for that Veteran (this should cause the End Product to be connected)
- On the confirmation page for the second RAMP Election Opt-In, see if there is an end_product description

Sjones352:
@leikkisa do you have any file number that has a RAMP-opt-election document in VBMS. I actually added a document on this file but I still get this error. Can you provide me a file if you have one that is eligible to participate in RAMP? 

[...]

PR #9947, opened 2019-03-13
Create EPE for RAMP reviews with matching EP

leikkisa:
connects #9946
connects #9623

### DIFF
--- a/app/models/end_product_establishment.rb
+++ b/app/models/end_product_establishment.rb
@@ -142,9 +142,8 @@ class EndProductEstablishment < ApplicationRecord
     RequestIssue::END_PRODUCT_CODES.find_all_values_for(:rating).include?(code)
   end
 
-  # Find an end product that has the traits of the end product that should be created.
   def preexisting_end_product
-    @preexisting_end_product ||= veteran.end_products.find { |ep| ep.active? && end_product_to_establish.matches?(ep) }
+    @preexisting_end_product ||= veteran.end_products.find { |ep| end_product_to_establish.matches?(ep) }
   end
 
   def cancel_unused_end_product!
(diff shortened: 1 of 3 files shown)


app/models/end_product_establishment.rb
```diff
@@ -142,9 +142,8 @@ def rating?
     RequestIssue::END_PRODUCT_CODES.find_all_values_for(:rating).include?(code)
   end
 
-  # Find an end product that has the traits of the end product that should be created.
```
leikkisa: I took this out because we're not creating a new end product here, just looking at the one that is already created.

app/models/ramp_review.rb
```diff
@@ -124,16 +124,16 @@ def new_end_product_establishment
       station: "397", # AMC
       benefit_type_code: veteran.benefit_type_code,
       user: intake_processed_by
-    )
+    ).tap do |new_epe|
+      new_epe.update!(reference_id: new_epe.preexisting_end_product.claim_id) if new_epe.preexisting_end_product
```
shanear: are we sure we want to save the new_epe here? Maybe we should just set the reference_id on the in memory object. Since that's what this method previously returned

app/models/ramp_review.rb
```diff
@@ -124,16 +124,16 @@ def new_end_product_establishment
...
+    end
   end
 
   def veteran
     @veteran ||= Veteran.find_or_create_by_file_number(veteran_file_number)
   end
 
   def connect_end_product!
```
shanear: It looks like `connect_end_product!` basically does nothing now? This is sort of confusing

leikkisa:
Per Shane, the new plan is:
- Find out how many Ramp Reviews are missing End Product Establishments
- Find out if any of them have active end products
- Backfill them
- Make sure new ones are not being created without saving the end product establishment

leikkisa:
closing in favor of https://github.com/department-of-veterans-affairs/caseflow/pull/10190

Issue #8936, opened 2019-01-24
Removing and withdrawing appeals


laurjpeterson:
## Background/Context
Veterans can choose to withdraw legacy or AMA appeals. Since the Veteran is withdrawing, there is no Board decision on this appeal. This means there should not be any decision issues.

BVA currently has a hardship case needing immediate action. The homeless Veteran has a RAMP appeal in the hearing lane. He wants to withdraw his appeal and file a supplemental claim instead. This is definitely allowed under AMA - the Board need to withdraw the current Board appeal, and then create a new EP/intake the supplemental claim. 

## Idea
- Allow users to Withdraw request issues on the Add/Remove issues screen, similar to how we currently allow them to Remove request issues. 
   - For example, in a dropdown next to the issue, the user could choose remove or withdraw. Or, another icon next to the Remove trashcan icon. **Design needed**
   - Withdrawn request issues is kind of like a type of removal. Removed issues can be intaken again in another lane. 
- If all issues are withdrawn, Caseflow closes all open or on hold tasks on the appeal. These two parameters indicate that the _entire _appeal is withdrawn. 
  - Note: this may require Caseflow having the ability to close tasks (instead of completing them). Spec to be worked on by Foxtrot this sprint: https://github.com/department-of-veterans-affairs/caseflow/issues/8701)
- If all issues are withdrawn (and all tasks closed), surface a user-facing error message explaining that the entire appeal is now withdrawn. **Design needed**
- If the user makes a mistake, they can re-add the issue as is done today. **Design needed** How to explain to the user that they can add the issue back in if they made a mistake?
- Show withdrawn request issues on the Add/Remove Issues page like we do for ineligible issues, but grayed out (**Design needed**). Include the date that the request issue was withdrawn. 
- Withdrawn issues do not show in Case details or any check out flows.

### Other considerations
- Search results: when a user searches for a withdrawn appeal. If all issues are withdrawn, update `Assigned to` column in Search results to withdrawn?
- Case timeline: add a withdrawn event?
- 

## Additional thoughts
If we like this idea, taking actions on Request issues, this pattern could apply to other scenarios. These scenarios have not been fully explored yet, and likely require a working session. 

1. **Erroneously docketed appeals** - when an appeal was incorrectly established. 
   - Currently, we are deleting these from the database, but need a long term solution (https://github.com/department-of-veterans-affairs/caseflow/issues/8824)
   - Idea: If all issues are removed, then prompt the user to close the appeal.

[...]

PR #9953, opened 2019-03-14
Tweak withdrawn copy

amprokop:
Makes temporary copy tweak so that withdrawn appeals aren't shown as "Pending dispatch from BVA", which will confuse and unnerve users.

Testing:
- Cancelled the RootTask on an appeal by changing the `status`
- Loaded Case Details and saw the copy change as expected


client/app/queue/components/TaskRows.jsx
```diff
@@ -207,14 +207,25 @@ class TaskRows extends React.PureComponent {
       timeline
     } = this.props;
 
+    const appealCancelled = taskList.find((task) => task.type === 'RootTask' && task.status === 'cancelled');
```
amprokop: Is checking the RootTask for a `cancelled` state the best way to tell that an appeal was withdrawn?

lowellrex: I don't think we've determined [how we will know](https://github.com/department-of-veterans-affairs/caseflow/blob/3d5579a87353c42b1205e533a65df85facdccb57/app/models/appeal.rb#L543) when an appeal has been withdrawn. [Here is the ticket](https://github.com/department-of-veterans-affairs/caseflow/issues/8936) that discusses implementing the ability to fully withdraw an appeal. It does not look like we've built this out yet so I don't think we expect to see AMA appeals that have been withdrawn yet. Or do we already have an AMA appeal that has been withdrawn in Caseflow?

amprokop: We do have one in production right now, that's correct. It was an urgent request from BVA to docket switch, and it was told to me that we're no longer deleting appeals to docket switch. On recommendation from @shanear and @lowellrex, I ran `task.update_columns(status: Constants.TASK_STATUSES.cancelled)` to cancel the open tasks. 

In testing locally, it revealed that after doing so, the appeal would still show up in case search, and would be displayed as pending dispatch from BVA.

amprokop: see https://dsva.slack.com/archives/CA1QBDLEL/p1551822381003500 for the discussion

lowellrex: In general I'm very wary of inferring the state of an appeal using the state of tasks attached to the appeal (I really believe the appeal should not know that tasks exist at all), but that is a discussion for a different time (and we already violate that dream of mine all over the `Appeal` model). So since we're already doing this can we fill in the `Appeal.withdrawn` method to indicate this is what we mean the a `RootTask` being cancelled?

amprokop: sure

pkarman:
@lowellrex this is now in conflict -- worth resolving and merging?

PR #9963, opened 2019-03-14
Remove disabled but valid Rubocop offenses

monfresh:
**Why**: By disabling offenses we actually care about, such as
complexity and class length, it allows classes or methods to get even
bigger and more complex indefinitely.

Now that we are using Code Climate, we can remove all of the
`rubocop:disable` statements we care about because Code Climate will
only complain about new offenses, not existing ones.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -110,7 +110,6 @@ class ApplicationController < ApplicationBaseController
   end
   helper_method :logo_path
 
-  # rubocop:disable Metrics/CyclomaticComplexity
   def application_urls
     urls = [{
       title: "Queue",
@@ -136,7 +135,6 @@ class ApplicationController < ApplicationBaseController
     # Only return the URL list if the user has applications to switch between
     (urls.length > 1) ? urls : nil
   end
-  # rubocop:enable Metrics/CyclomaticComplexity
   helper_method :application_urls
 
   def dropdown_urls
@@ -180,8 +178,6 @@ class ApplicationController < ApplicationBaseController
     # :nocov:
   end
 
-  # rubocop:disable Metrics/CyclomaticComplexity
-  # rubocop:disable Metrics/PerceivedComplexity
   def case_search_home_page
     if feature_enabled?(:case_search_home_page)
       return false if current_user.admin?
@@ -194,8 +190,6 @@ class ApplicationController < ApplicationBaseController
     false
   end
   helper_method :case_search_home_page
-  # rubocop:enable Metrics/PerceivedComplexity
-  # rubocop:enable Metrics/CyclomaticComplexity
 
   def deny_vso_access
     redirect_to "/unauthorized" if current_user&.vso_employee?
--- a/app/controllers/distributions_controller.rb
+++ b/app/controllers/distributions_controller.rb
@@ -59,7 +59,6 @@ class DistributionsController < ApplicationController
     }, status: :forbidden
   end
 
-  # rubocop:disable Metrics/MethodLength
   def json_error(error)
     case error
     when :not_judge
@@ -78,9 +77,8 @@ class DistributionsController < ApplicationController
       {
         "error": error,
         "title": "Cases in your queue are waiting to be assigned",
-        # rubocop:disable Metrics/LineLength
-        "detail": "Please assign all cases that have been waiting in your assignment queue for more than 30 days before requesting more."
-        # rubocop:enable Metrics/LineLength
+        "detail": "Please assign all cases that have been waiting in your " \
+                  "assignment queue for more than 30 days before requesting more."
       }
     when :different_user
       {
@@ -102,7 +100,6 @@ class DistributionsController < ApplicationController
       }
     end
   end
-  # rubocop:enable Metrics/MethodLength
 
   def feature_enabled?
     FeatureToggle.enabled?(:automatic_case_distribution, user: current_user)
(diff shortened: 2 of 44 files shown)


pkarman:
I see where you're headed with this.

I'm uncomfortable removing rubocop from our own lint task and completely out-sourcing the state of our violation history to Code Climate.

My experience with Code Climate in the past is that:

* it's difficult to align the local rubocop version/rules with what runs at CC and makes it impossible to know whether you've introduced new problems *before* you commit and push to GH. To borrow from the technical debt metaphor, it's like you don't know if you're attempting to borrow more unless you go to a physical bank branch to get your balance, rather than being able to check your account from home.
* By outsourcing enforcement, we completely rely on CC to remember for us that we have violations and not to block us from merging on new violations. I'm not sure I trust that touching an existing file with existing (but unmarked) violations will pass muster with CC, even if I don't touch the parts of the file that include the violations. This gives away a lot of control over merging (a critical path in the CI pipeline) to a service over which we don't have much control or visibility.
* CC uses reek and rubocop together, and I've found that the rules they enforce can often be at odds with one another. E.g. rubocop encourages short methods. Reek encourages a limit on the number of methods. So it's possible (and this happened to me with enough frequency on a large codebase that I remember it clearly) that refactoring to satisfy rubocop meant that I offended reek, and vice versa.

I would rather that we maintain our commitment to not introducing any new exceptions, and then pay down our principal by refactoring to remove existing exceptions. That allows us to make judgements about our debt on a PR-by-PR basis, rather than outsourcing management of our debt to a 3rd party.

monfresh:
> it's difficult to align the local rubocop version/rules with what runs at CC 

CC uses the `.rubocop.yml` from the repo, so the local rules are aligned. As for the Rubocop version, it's possible that they might be different if we upgrade the one in our repo to a version that's not yet available in CC. Checking just now, I see that CC is using the latest version 0.67 (https://github.com/codeclimate/codeclimate-rubocop/branches/all?utf8=%E2%9C%93&query=channel%2Frubocop), which they added a week after 0.67.2 was released. That's a reasonable turnaround to me.

> and makes it impossible to know whether you've introduced new problems before you commit and push to GH. 

There are several recommended ways you can find out whether you've introduced new problems before you commit and push to GH:

- Use the Rubocop plugin for your editor
- Run `git diff --name-only --diff-filter=MA | xargs rubocop` in your branch. See https://github.com/rubocop-hq/rubocop/issues/5477 for more details

> I'm not sure I trust that touching an existing file with existing (but unmarked) violations will pass muster with CC, even if I don't touch the parts of the file that include the violations. 

I don't recall for sure what CC would do in that case. However, we are free to ignore violations in a PR if they are clearly not related to the PR. I would emphatically prefer to be warned about something that I can ignore with confidence than to allow existing violations to be in further violation, hence increasing our debt.

> CC uses reek and rubocop together

Only because we tell it to in our `.codeclimate.yml`. We are free to specify the engines we want CC to run.

> I've found that the rules they enforce can often be at odds with one another.

I recall the issue you are referring to, but I don't think it happened as frequently as you remember. How many times have you seen it in this repo? In any case, we are free to adjust the Reek config to our liking. We are also free to ignore certain offenses if we agree that they are false positives.

> I would rather that we maintain our commitment to not introducing any new exceptions, and then pay down our principal by refactoring to remove existing exceptions. 

I consider increasing the length or complexity of a class or method that is already in violation to be a new exception. Is it not part of our commitment to make sure we don't increase the technical debt of existing violations? How can we ensure that we don't keep making existing methods/classes worse if we don't remove the `rubocop:disable` comments?

Perhaps I'm misunderstanding, but it sounds like you are treating refactoring existing offenses as a separate exercise. I think it's better to refactor things as we touch them. It's not worth refactoring code that is rarely used. If we happen to need to modify a method or class that is already in violation, by removing the `rubocop:disable` comments, we can be alerted. 

If we make a clean change to a file that is not in violation but contains a method that is, but untouched by the PR, and if CC reports the previously-existing offense, we can choose to ignore it for that PR. I think that's a much better option than having to rely on humans to see if a PR makes a change to a method or class that is already in violation but is being ignored by Rubocop. We know from experience that these types of "make existing vioations worse" issues have been unchecked, such as allowing classes to grow way past our 400-line limit, which itself is much larger than the recommended limits established by Rubocop (100) and Code Climate (250). 

We've also had new instances of `rubocop:disable` be approved in PRs as recently as 28 days ago: https://github.com/department-of-veterans-affairs/caseflow/pull/10209. That's a separate issue, but goes to show that we should be relying on bots to enforce our rules for us rather than humans.

monfresh:
I tested how CC deals with existing offenses and I can confirm that we can safely remove instances of `rubocop:disable`. I tested this by:

1. forking this repo
2. opened a PR that removes a couple of `rubocop:disable` statements from `ApplicationController`: https://github.com/monfresh/caseflow/pull/2
3. Merged that PR
4. Opened a new PR that makes a clean change to `ApplicationController`:  https://github.com/monfresh/caseflow/pull/3/files

Result: CC does not complain about existing offenses. It only looks at new offenses introduced in the PR. You can see in the PR that the first time it analyzed it, I had added a new offense, which was the only one it complained about. Then, I removed it, and it said there are 0 issues.

monfresh:
@pkarman #10591 has been merged. Can this be approved now?

pkarman:
good to go @monfresh -- please announce in #appeals-engineering when you merge so folks are not caught off guard.

Issue #9972, opened 2019-03-14
Intake | Remove AMA feature flags


pkarman:
Now that AMA is fully in effect, we should remove all the related feature flags in our code.

These include:

* [x] `intake`
* [x] `intakeAma`
* [x] `intake_legacy_opt_in`

leikkisa:
@Sjones352 Once these are all merged, could you do another pull request to remove these from appeals_deployment?

https://github.com/department-of-veterans-affairs/appeals-deployment/blob/master/ansible/vars/features-config.yml

Sjones352:
Pr in appeals-deployment
https://github.com/department-of-veterans-affairs/appeals-deployment/pull/2039

Issue #9816, opened 2019-03-06
Intake | Legacy opt-in modal multi-click fails to register "None of these match"


pkarman:
In the Legacy issue (VACOLS) opt-in modal during intake, if you:

* click on an issue radio button
* then click change to the "None of these match" radio button
* then click Add Issue

it will fail to register the None of these match and instead apply an eligibility error based on the first selected radio button issue.

We observed this during an Intake ridealong.

Sjones352:
@leikkisa Please provide validation instructions. @pkarman I am a little confused about this statement `then click change to the "None of these match" radio button` I thought "None of these match" is a regular button.

leikkisa:
@Sjones352 

### Validation
- Choose a Veteran that has Vacols issues (we might need to find one or create one)
- Do an AMA intake
- Add an issue
- The modal for whether it is connected to any Vacols issues should pop up
- First select one of the VACOLS issues
- Then select "None of these match"
- Then finish the intake
- The Request Issue that was added should have nil for vacols_id in the database

Sjones352:
Checking for a veteran with Vocals issues, If I don't find one then let's create one.

Sjones352:
@pkarman  and @leikkisa unfortunately this fails UAT validation unless I am doing wrong.

<img width="778" alt="Screen Shot 2019-03-13 at 7 09 08 PM" src="https://user-images.githubusercontent.com/23080951/54321366-31edfe80-45c7-11e9-8d4b-69e1a5ebff19.png">
<img width="1252" alt="Screen Shot 2019-03-13 at 7 10 05 PM" src="https://user-images.githubusercontent.com/23080951/54321367-32869500-45c7-11e9-970d-95947c441e99.png">

vacols_id is not nill
<img width="1680" alt="Screen Shot 2019-03-13 at 7 33 17 PM" src="https://user-images.githubusercontent.com/23080951/54321392-51852700-45c7-11e9-8722-941b5bdf6976.png">

pkarman:
thanks @Sjones352 -- the trick for this one is to click the "Dental" and then click "None of these match" and then the "Add this issue" button.

I validated this locally and it works, so I am going to close.

PR #9973, opened 2019-03-14
Remove intake_legacy_opt_in feature flag, update seeds

pkarman:
connects #9972

### Description

While trying to locally validate #9816 I discovered that our legacy opt-in code was (a) not enabled in development, and (b) the seed data was not updated to reflect AMA date passage.


app/models/veteran.rb
```diff
@@ -213,6 +213,7 @@ def find_or_create_by_file_number(file_number, sync_name: false)
     end
 
     def find_by_file_number_or_ssn(file_number_or_ssn, sync_name: false)
+      file_number_or_ssn.sub!(/[CS]$/, "")
```
shanear: why add this?

pkarman: locally I find myself copy/pasting from the `local/vacols/*csv` files using VBMS id values. Those don't work because of the trailing letter. This would Just Do the Right Thing if someone does that.

pkarman: more trouble than it saves right now. I deleted it.

PR #9978, opened 2019-03-14
add green checkmark for legacy dispatched, as well as date to the left

### LINKED ISSUE #9511: Case timeline does not show dispatched legacy appeals correctly
The "Dispatched from BVA" element of Case Timeline does not show completed even though a legacy case has been dispatched from the Board.

Currently we do not show the date when an appeal was dispatched from the Board in case timeline. We [correctly change the text](https://github.com/department-of-veterans-affairs/caseflow/blob/2fb7b02db05e3fe2854e75a6232d19dcfc46f3aa/client/app/queue/components/TaskRows.jsx#L212), but we do not show the date the decision was dispatched nor do we show the green dot indicating that the decision has been dispatched.

image

## Acceptance criteria
* [ ] Show the green check mark instead of grey dot if the decision has been dispatched from the Board
* [ ] Show the date the decision was dispatched from the Board if it has been in the pattern of the other tasks

## Notes
- Could/should we be using Appeals Status code for this at all?

## Examples
- https://appeals.cf.ds.va.gov/queue/appeals/2502986
- https://appeals.cf.ds.va.gov/queue/appeals/2713767

youngfreezyVA:
Resolves #9511
### Description
https://github.com/department-of-veterans-affairs/caseflow/issues/9511

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
For legacy appeals 

With a decision date:
image
without:

image


client/app/queue/components/TaskRows.jsx
```diff
@@ -206,15 +206,22 @@ class TaskRows extends React.PureComponent {
       taskList,
       timeline
     } = this.props;
+    const isLegacyAppealWithDecisionDate = appeal.decisionDate && appeal.isLegacyAppeal;
```
youngfreezyVA: @aroltsch @laurjpeterson in the code here I scoped the changes to only legacy appeals...so the green checkmark/date to the left will only appear if its a legacy appeal.

spec/feature/queue/case_details_spec.rb
```diff
@@ -471,6 +471,28 @@
     end
   end
 
+  context "when there is a dispatch and decision_date" do
```
aroltsch: Very nice test! Is there a test when there is no decision date to test for no green checkmark?

youngfreezyVA: thanks! yup, the test right above it: https://github.com/department-of-veterans-affairs/caseflow/blob/bff38888759c680db26c39ff05721cd135f56046/spec/feature/queue/case_details_spec.rb#L469

PR #9981, opened 2019-03-14
Load queues with tasks associated to delete VACOLS appeals

### LINKED ISSUE #9856: Enable folks to visit queues that contain tasks associated with deleted LegacyAppeals
VLJ support staff are currently unable to load the team queue because the set of tasks it attempts to display includes a task that is associated with a `LegacyAppeal` with a case that is no longer in the `vacols.brieff` table. I am not entirely sure why the case is not in the `brieff` table, however the application should be resilient enough to handle this case and return the rest of the tasks to the frontend without error.

## Acceptance criteria:
* [ ] Render queues even if the queue contains a legacy case that has been removed from VACOLS.

lowellrex:
Resolves #9856.


lowellrex:
Didn't add any tests for the `AppealRepository` changes outside of the feature changes but I'd be happy to if you think it'd be good to (there aren't any preexisting tests).

app/services/appeal_repository.rb
```diff
@@ -32,28 +33,33 @@ def self.eager_load_legacy_appeals_for_tasks(tasks)
...
   end
 
-  def self.find_case_record(id)
+  def self.find_case_record(id, ignore_misses: false)
     # Oracle cannot load more than 1000 records at a time
     if id.is_a?(Array)
       id.in_groups_of(1000, false).map do |group|
-        VACOLS::Case.includes(:folder, :correspondent, :representatives, :case_issues).find(group)
+        if ignore_misses
+          VACOLS::Case.includes(:folder, :correspondent, :representatives, :case_issues).find_by(bfkey: group)
```
tomas-nava: `find_by` doesn't do what we want; it'll always only return [the first matching record](https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find_by) and we want all matching records. I think we want something like this instead:

```ruby
VACOLS::Case.includes(:folder, :correspondent, :representatives, :case_issues).where("bfkey in (?)", group)
```

It isn't causing any test failures because none of our test that pass `ignore_misses: true` are passing more than one unique `bfkey`.

lowellrex: Added a test for this method as well.

Issue #9922, opened 2019-03-12
Supplemental claims seem to be subject to timeliness validations


shanear:
The `before_ama` eligibility check should not be applied to Supplement Claims (form 0995).

We should remove that check for SCs on front and back ends.

### Acceptance

* Create a Supplemental Claim
* Add a nonrating request issue with a date older than Feb 14, 2019
* Should *not* receive the ineligible message for prior to Feb 14.

### Background

From Jane:

> Hi All,  We received a report from one of the DROC intake processing centers (IPC) reporting an issue they are encountering in Caseflow. Here's the scenario: Veteran submits an 0995 on 03/01/2019 for a decision made on 12/05/2017 (outside of the 1 year). IPC tries to process the claim through Caseflow Intake (as the M21 dictates), however, they receive an error message from Caseflow indicating the Veteran is ineligible because he is outside of the year. IPC then sends an ineligibility notice to the Veteran.  We received clarification from our Program Admin team providing that an 0995 can be submitted at anytime; only an 0996 should be rejected by the system after one year. The DROC IPCs should be able to intake for those supplemental claims received on the 0995, but the 12/2017 effective date will not be protected.  Can you please look into this issue?  Thank you!

> Hey @shane-russell-va,  Yes, please. Can you please change this to allow for Caseflow to intake the supplemental claims received on the 0995?  Is there any way to identify these claims that were previously blocked by Caseflow?   Thanks for your help.

pkarman:
To find them in the db:
```
RequestIssue.where(ineligible_reason: :before_ama).select { |reqi| reqi.decision_review.is_a?(SupplementalClaim) }
```

pkarman:
manual clean up performed to resubmit these once-ineligible request issues.

```ruby
screqs = RequestIssue.where(ineligible_reason: :before_ama).select { |reqi| reqi.decision_review.is_a?(SupplementalClaim) }

screqs.each do |r|
  r.transaction do
    r.update!(ineligible_reason: nil)
    sc = r.decision_review
    sc.create_issues!([r])
    sc.establish!
  end
end
```

Sjones352:
**Passes UAT**

<img width="1185" alt="Screen Shot 2019-03-20 at 11 31 50 AM" src="https://user-images.githubusercontent.com/23080951/54697441-f6da5680-4b03-11e9-890b-1ef14e361682.png">
[...]

PR #9982, opened 2019-03-14
Do not apply before_ama check to Supplemental Claims

pkarman:
connects #9922 

### Description
Do not apply `before_ama` eligibility check to Supplemental Claims, per clarification from AMO.


pkarman:
the tests that require a single ineligible issue remain on a claim review no longer work for SC, and those tests will go away Real Soon because of the feature to withdraw a claim. So I changed them to just apply to HLR.

PR #10005, opened 2019-03-15
Add AsyncableJobsReporter

pkarman:
Makes it simpler to summarize what is in the queue, categorized by age and error type.

### DIFF
--- a/app/services/asyncable_jobs.rb
+++ b/app/services/asyncable_jobs.rb
@@ -19,20 +19,6 @@ class AsyncableJobs
     jobs.select { |j| msg_regex.match?(j[j.class.error_column]) }
   end
 
-  # summarize all jobs by class and error and age
-  def summary
-    report = {}
-    jobs.each do |job|
-      cls = job.class.to_s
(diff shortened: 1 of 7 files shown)


app/services/asyncable_jobs.rb
```diff
@@ -19,6 +19,20 @@ def find_by_error(msg)
...
+    jobs.each do |job|
+      cls = job.class.to_s
+      age = ((Time.zone.now - job.sort_by_last_submitted_at) / 3600 / 24).ceil.to_s
+      err = job[job.class.error_column]
+      report[cls] ||= {}
+      report[cls][age] ||= { err => 0 }
+      report[cls][age][err] += 1
+    end
+    report
+  end
```
shanear: Would we want to show this on the UI?

pkarman: yes currently for the console, though I think we should also use this for the daily slack alert report on jobs.

pkarman: will need to scrub all the PII out of the error messages before we put it in the slack job.

Issue #8436, opened 2018-12-26
Use JudgeTeam membership to determine judges and attorneys


lowellrex:
We currently use the `VACOLS.STAFF` table to [determine if a user is a judge or an attorney](https://github.com/department-of-veterans-affairs/caseflow/blob/6bea2aab4a10d6bf3de6b4b26ab1f229c9f5637a/app/repositories/user_repository.rb#L80) (by way of [methods on the User model](https://github.com/department-of-veterans-affairs/caseflow/blob/6bea2aab4a10d6bf3de6b4b26ab1f229c9f5637a/app/models/user.rb#L43)). This leads to problems when employees are [misidentified in VACOLS as attorneys](https://github.com/department-of-veterans-affairs/caseflow/issues/8224) and complexity around when somebody is an [acting judge](https://github.com/department-of-veterans-affairs/caseflow/blob/6bea2aab4a10d6bf3de6b4b26ab1f229c9f5637a/app/repositories/user_repository.rb#L93). #8117 moved membership in judge teams into the `JudgeTeam` model which we should be able to use determine whether somebody is a judge or an attorney without relying on VACOLS.

## Acceptance criteria
* [ ] Use `JudgeTeam` membership to determine if somebody is a judge or an attorney

lowellrex:
LARGE. Will need to confirm that the Caseflow-only way of identifying judges and attorneys (using `JudgeTeam`s) matches what exists in VACOLS and deconflicting the differences with the Board when we discover them (cc @marvokdolor-gov ).

lowellrex:
Did some pairing with Tomas and winnowed down the list of judges in VACOLS that differ from judges with `JudgeTeam`s in Caseflow to 27. More work to do here, but this should give whoever works on this next a good head start.

Worth noting that "acting judges" do not have judge teams, so wherever we check for the judge vacols role (which acting judges have), we will need to allow for them to pass that check as well (maybe by hitting VACOLS then as a backup?).

```
rails c> u = JudgeTeam.first.admins.first
> judges_and_acting_judges_in_vacols = VACOLS::Staff.where(sactive: 'A', svlj: ['J', 'A'])
> judges_and_acting_judges_in_vacols.count
# 282
> JudgeTeam.count
# 99

> judges_in_vacols = VACOLS::Staff.where(sactive: 'A', svlj: 'J')
> judges_in_vacols.count
# 132

# Why are there 33 extra judges in VACOLS?

[...]

PR #10006, opened 2019-03-15
Allow judge to judge assignment of ama cases

### LINKED ISSUE #7612: Judge to Judge assignment of RAMP tasks fails

hschallhorn:
Resolves #7612 

### Description
Allow judges to reassign ama cases to another judge

### Acceptance Criteria
- [x] On the unassigned cases page, allow a judge to reassign a task to another judge
- [x] On the case details cases page for an usassigned task, allow a judge to reassign the task to another judge

### Testing Plan
1. Checkout this branch and reseed the database with `bundle exec rake db:setup`
2. Switch to the user "BVAEBECKER"
3. Open your queue and navigate to assign cases
4. Select a task, select "Other" from the "Assign to:" dropdown, and select "Judge with hearings and cases" from the "Enter the name of an attorney or judge:" dropdown.
5. Hit "Assign 1 case" and notice no errors (This case should then also appear in BVAAABSHIRE's assign cases view)


client/app/queue/components/AssignWidget.jsx
```diff
@@ -65,16 +65,20 @@ class AssignWidget extends React.PureComponent {
...
   assignTasks = (selectedTasks, assigneeId) => {
     const {
+      attorneys,
       previousAssigneeId,
       userId
     } = this.props;
 
     this.props.setSavePending();
 
+    const assigneeIsJudge = _.find(attorneys.data, { id: Number(assigneeId) })['judge_team?'];
```
hschallhorn: Uses judge team membership to identify if the selected assignee is a judge. See #8436

lowellrex:
Did we explore the idea of keeping this mostly on the back-end? On the case details page we could add an [option to "reassign to different judge"](https://github.com/department-of-veterans-affairs/caseflow/blob/572d49ad47f0590f05601cd7baf84ed23c0039b3/client/constants/TASK_ACTIONS.json#L18) and on the judge assign view instead of [submitting bulk update requests to `/tasks`](https://github.com/department-of-veterans-affairs/caseflow/blob/572d49ad47f0590f05601cd7baf84ed23c0039b3/client/app/queue/QueueActions.js#L377) we could POST requests to something like `/judge_bulk_assign/:task_id` (with the assignee in the request body) and have a `JudgeBulkAssignController` determine whether we want to reassign the judge task to another judge or create an attorney task based on the assignee. The advantage of this approach is that we wouldn't have to pass more information to the front-end since the back-end already knows if the user is an attorney or judge. Thoughts?

hschallhorn:
Had not considered handling most of it on the back end. Was mostly following Alex's suggested plan of attack.
RE your suggestions: both the case details page and the judge assign view use the [AssignWidget](https://github.com/department-of-veterans-affairs/caseflow/pull/10006/files?file-filters%5B%5D=.jsx#diff-fbcfcbf885226ceffe45ee2726838f4e) so I think it would make more sense to  use the same option (either adding an option to reassign to a different judge or the judge_bulk_assign route). Adding an option to reassign to a different judge would require us to request all the [judge users](https://github.com/department-of-veterans-affairs/caseflow/blob/572d49ad47f0590f05601cd7baf84ed23c0039b3/app/controllers/users_controller.rb#L7) (another call to VACOLS) but is more explicit to the user what action they're performing.

lowellrex:
> both the case details page and the judge assign view use the AssignWidget so I think it would make more sense to use the same option

Didn't know that, but I absolutely agree!

> Adding an option to reassign to a different judge would require us to request all the judge users (another call to VACOLS) but is more explicit to the user what action they're performing.

I think we could avoid the VACOLS request by building the list of judges using something like `JudgeTeam.all.map(&:judge)`. But I think it probably just makes sense to leave the assign widget as is (even though it is less explicit than adding a new option) and just change the endpoint we submit the request to.

app/controllers/assign_tasks_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  private
+
+  def task
+    @task ||= Task.find(create_params.first[:parent_id])
+  end
+
+  def tasks
+    if assignee_is_judge?
+      tasks = task.update_from_params(update_params, current_user)
```
lowellrex: Can we call `task.reassign(create_params.first, current_user)` here so the old `JudgeTask` is cancelled at the same time as the  when we hit this endpoint?

app/controllers/assign_tasks_controller.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class AssignTasksController < TasksController
```
lowellrex: Can we name this to something a little more precise to the function it serves? Maybe `JudgeAssignTasksController` to differentiate between the tasks we create by the `TasksController.create` function?

app/controllers/assign_tasks_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+      tasks = AttorneyTask.create_many_from_params(create_params, current_user)
+    end
+
+    tasks
+  rescue ActiveRecord::RecordInvalid => error
+    invalid_record_error(error.record)
+  end
+
+  def assignee_is_judge?
+    JudgeTeam.for_judge(User.find(create_params.first[:assigned_to_id]))
```
lowellrex: We're using `create_params.first` in a few places, would it make sense to have a helper function called something like `task_params` as a shortcut to access just that element of the `create_params` array?

app/controllers/assign_tasks_controller.rb
hschallhorn: [Updated the param](https://github.com/department-of-veterans-affairs/caseflow/pull/10006/files#diff-2b698a9ff669fb4944bd0a54b30aa659R391)s so it falls through to reassign()

client/app/queue/QueueActions.js
```diff
@@ -379,15 +379,20 @@ export const initialAssignTasksToUser = ({
...
           parent_id: oldTask.taskId,
           assigned_to_id: assigneeId
-        }]
+        }],
+        task: {
+          reassign: {
+            assigned_to_type: 'User',
+            assigned_to_id: assigneeId
+          }
+        }
```
lowellrex: Does calling `task.reassign(task_params, current_user)` directly in `JudgeAssignTasksController.update_tasks()` work? If we did that then we could get rid of this element of the request body.

hschallhorn: reassign() still [requires params[:reassign]](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/task.rb#L134). I could create `reassign_params` on the back end though to pass directly to it

lowellrex: Gotcha! Sure this works for me!

Issue #10001, opened 2019-03-15
Intake | Sync reviews race condition causes duplicate EPs


shanear:
Today Colin noticed a duplicate EP being created for a Higher Level Review. After digging through the logs, it looks like the there was a `DecisionReviewProcessJob` launched as a result of the of the intake on the main thread.

However, before that job finished, a cron'ed `SyncReviewsJob` noticed an unprocessed Higher Level Review, starting a second `DecisionReviewProcessJob`.

Those two jobs got into a race condition, which ended up creating a duplicate EP.

## Solution
- We are only going to retry never attempted jobs if they were submitted over 3 hours ago

pkarman:
I think this is an argument for only processing jobs via `SyncReviewsJob` and not also calling immediately in the sync flow.

shanear:
That introduces up to an extra 2-3 minute delay for the happy path, which I don't think is great. The business has said often that some CA's want to be able to see the EP very soon after the intake.

I think we can probably change `requires_processing` to put a bigger buffer for `last_attempted_at`. We should only be retrying Asyncables that failed over 30 min to an hour ago. Thoughts?

pkarman:
Maybe we just remove the `never_attempted` condition in the `attemptable` method? That way the cron job would only try things that have failed.

pkarman:
Closing till we see this issue re-emerge.

PR #10007, opened 2019-03-15
Update cron job to not re-processes recently submitted jobs

shanear:
Fixes the race condition described in #10001 

connects #10001

### DIFF
--- a/app/models/concerns/asyncable.rb
+++ b/app/models/concerns/asyncable.rb
@@ -63,7 +63,9 @@ module Asyncable
     end
 
     def never_attempted
-      where(attempted_at_column => nil)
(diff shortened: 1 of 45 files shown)


app/models/concerns/asyncable.rb
```diff
@@ -63,7 +63,9 @@ def processable
     end
 
     def never_attempted
-      where(attempted_at_column => nil)
+      where(attempted_at_column => nil).where(
+        arel_table[submitted_at_column].lt(processing_retry_interval_hours.hours.ago)
```
pkarman: should this be `arel_table[last_submitted_at_column]` ?

PR #10015, opened 2019-03-15
Adding filtering to RO column

### LINKED ISSUE #9874: Bug: Hearing Team Queues and Individual Queues are missing filterable RO column
In the team queues for the Hearing Admin Team, we've lost the Regional Office column and filter, so that the Branch can sort tasks by RO. 

I believe we had this implemented before, but it doesn't appear in production any longer.

Also, we had designed the RO column and filter on individual queues too for the hearing branch users, but that also doesn't appear.

mdbenjam:
Resolves #9874 

### Description
This PR adds the RO column for all hearing admin users (this already existed for members in the hearing management organization).

This also adds filtering, and correctly maps the RO key to the name of the RO.


app/models/serializers/work_queue/task_serializer.rb
```diff
@@ -61,7 +61,7 @@ class WorkQueue::TaskSerializer < ActiveModel::Serializer
   end
 
   attribute :closest_regional_office do
-    object.appeal.closest_regional_office
+    RegionalOffice.find!(object.appeal.closest_regional_office).city
```
lomaxap: won't this throw an error if the closest_regional_office is nil?

mdbenjam: Good call! Updating now.

PR #10019, opened 2019-03-15
Refactoring UpdateAppellantRepresentationJob to work with legacy appeals

### LINKED ISSUE #9896: VSO Schedule - UpdateAppellantRepresentationJob
We are using TrackVeteranTasks to determine which appeals' hearings to display for each VSO. Currently the UpdateAppellantRepresentationJob only creates TrackVeteranTasks for AMA appeals. We need to update that job to create TrackVeteranTasks for legacy appeals as well.

mdbenjam:
Resolves #9896  

### Description
This PR updates the `UpdateAppellantRepresentationJob` to query all legacy appeals with hearings so that we have track veteran tasks for them.


app/models/tasks/track_veteran_task.rb
```diff
@@ -30,6 +30,31 @@ def hide_from_task_snapshot
     true
   end
 
+  def self.sync_tracking_tasks(appeal)
```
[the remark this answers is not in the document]
mdbenjam: This method was simply moved from appeal, so that we can call with legacy appeals as well now.

spec/models/tasks/track_veteran_task_spec.rb
```diff
@@ -42,4 +42,74 @@
       expect(tracking_task.hide_from_case_timeline).to eq(true)
     end
   end
+
+  describe ".sync_tracking_tasks" do
```
mdbenjam: test moved from appeal_spec

app/jobs/update_appellant_representation_job.rb
```diff
@@ -16,8 +18,13 @@ def perform
     # Set user to system_user to avoid sensitivity errors
     RequestStore.store[:current_user] = User.system_user
 
-    Appeal.active.each do |a|
-      appeal_new_task_count, appeal_closed_task_count = a.sync_tracking_tasks
+    appeals_to_update.each do |a|
+      sync_record = a.record_synced_by_job.find_or_create_by(sync_job_name: UpdateAppellantRepresentationJob.name)
+
+      sync_record.attempted!
+      appeal_new_task_count, appeal_closed_task_count = TrackVeteranTask.sync_tracking_tasks(a)
```
sharonwarner: Definitely not blocking, but maybe we should add a todo or an issue to catch some errors here. It looks like one bad record will derail the job indefinitely.

Issue #9983, opened 2019-03-14
Intake Edit | Add select input for removing or withdrawing issues


shanear:
Since there are now multiple actions you can take on a request issue on the edit screen, we need to have a select box next to each issue with the options.

All withdrawal/removal mocks: https://va-gov.invisionapp.com/share/BRQUJR5J5AE#/screens/351540954

## Acceptance Criteria
- Next to each issue on the edit issues screen, there should be a select box with two options, "Withdraw issue" and "Remove issue".
- If "Remove issue" is selected, it should do the same thing as if the "Remove issue" button was clicked.
- Nothing should happen if "Withdraw issue" is selected. We will handle that in a future story

## Mocks
<img width="1223" alt="Screenshot 2019-03-14 15 27 13" src="https://user-images.githubusercontent.com/1596259/54395723-e4789c80-466d-11e9-88f3-ac38be1e0417.png">

## Notes
- This functionality should be enabled and disabled using a feature flag named `:withdraw_decision_reviews`

Sjones352:
**Passes UAT**
<img width="1680" alt="Screen Shot 2019-03-26 at 10 37 14 AM" src="https://user-images.githubusercontent.com/23080951/55006095-8d4ec200-4fb3-11e9-8404-9275783db692.png">
<img width="1438" alt="Screen Shot 2019-03-26 at 10 37 40 AM" src="https://user-images.githubusercontent.com/23080951/55006096-8d4ec200-4fb3-11e9-8893-307f69b74bfe.png">
<img width="1457" alt="Screen Shot 2019-03-26 at 10 38 11 AM" src="https://user-images.githubusercontent.com/23080951/55006097-8d4ec200-4fb3-11e9-8b81-d06b9c024c75.png">
<img width="1456" alt="Screen Shot 2019-03-26 at 10 38 21 AM" src="https://user-images.githubusercontent.com/23080951/55006098-8de75880-4fb3-11e9-80b4-217f260a9992.png">
<img width="1338" alt="Screen Shot 2019-03-26 at 10 38 32 AM" src="https://user-images.githubusercontent.com/23080951/55006100-8de75880-4fb3-11e9-98f0-4264e2a5ba4c.png">

PR #10042, opened 2019-03-19
Add dropdown box for remove/withdraw issue action

leikkisa:
connects #9983


spec/support/intake_helpers.rb
```diff
@@ -237,6 +237,12 @@ def click_remove_intake_issue(number)
     issue_el.find(".remove-issue").click
   end
 
+  def click_remove_intake_issue_dropdown(number)
```
pkarman: nice!

Issue #10051, opened 2019-03-19
Intake | Close Appeal request issues only if closed tasks


pkarman:
Per @aroltsch 

> for the appeal, we should only close the request issue if the appeal has closed BvaDispatchTask tasks

See thread in https://dsva.slack.com/archives/C754H751T/p1553001634188200

> we should also reopen request issues that were closed and we might need to backfill missing decision issues

aroltsch:
### Appeals that have been dispatched but missing decision issues:

```
Appeal.all.select { |a| a.tasks.where(type: "BvaDispatchTask").map(&:status).uniq == ["completed"] && a.decision_issues.empty? }
```

### Appeals that have not been dispatched but have have `decided` request issues: 

```
Appeal.all.select { |a| a.tasks.where(type: "BvaDispatchTask", status: "completed").empty? && a.request_issues.active.empty? }
```

pkarman:
Clean up included:
```
appeals = Appeal.all.select { |a|
  a.tasks.where(type: "BvaDispatchTask", status: "completed").empty? &&
    a.request_issues.decided.any?
}

appeals.map(&:id)

appeals.map(&:request_issues).flatten.select(&:decided?).each { |r|
  r.update!(closed_status: nil, closed_at: nil) if r.decided?
}
```

pkarman:
this problem was due to #9556 where we should have selectively skipped certain Appeals and not set their request issues to "decided".

pkarman:
We identified further Appeals that require decision issues be re-created.

```
Appeal.all.select { |a| a.tasks.where(type: "BvaDispatchTask", status: ["assigned", "in_progress"]).any? && a.decision_issues.empty? }

Appeal.all.select { |a|
  a.decision_documents.any? && a.decision_issues.empty?
}
```

Affected Appeals: `[1040, 1127, 1203, 1220, 1248, 1296, 1273, 1373, 1394, 534, 1113, 1689, 240, 624, 138, 1096, 1542, 1640, 1082]`

@aroltsch is going to talk with @laurjpeterson about next steps.

pkarman:
@aroltsch can we close this one?

aroltsch:
Yes.

Issue #9556, opened 2019-02-20
Intake | Set closed_at and closed_status for decided issues


shanear:
In order to use `closed_at` and `closed_status` as the source of truth on whether an issue is open or not, we need to set decided issues to closed.

## Acceptance Criteria
- When the decision issues are synced for a request issue on a claim review, set `closed_at` to now and `closed_status` to `decided` for all request issues with at least one decision issue.
- When an appeal is outcoded (see: `BvaDispatchTask.outcode`), all request issues with at least one decision issue should have `closed_at` set to now, and `closed_status` set to `decided`

leikkisa:
Looks good to me! I suppose we should also backfill data as part of this ticket.

leikkisa:
Proposed script for data migration, tested in UAT:
```
RequestIssue.joins(:request_decision_issues).where(closed_at: nil).each{ |ri| ri.update!(closed_at: ri.request_decision_issues.first.created_at, closed_status: 'decided') }
```
I started by looking at decision_sync_processed_at for the datetime but it is nil for most of them.  UAT has decision issues without request issues, but checked in prod and don't believe that is the case there.

leikkisa:
Due to https://github.com/department-of-veterans-affairs/caseflow/issues/9767, I would like to do this migration sooner than later, even if that means we have to do it again one after Shane is done backfilling decision issue data.

leikkisa:
I decided to go ahead and just run this.

pkarman:
verified in production:
```
irb(main):002:0> RequestIssue.where(closed_status: "decided").count
[2019-03-11 16:00:50 -0400]    (4.8ms)  SELECT COUNT(*) FROM "request_issues" WHERE "request_issues"."closed_status" = $1  [["closed_status", "decided"]]
=> 889
```

PR #10053, opened 2019-03-19
Do not create decision issues if appeal lacks completed BVA Dispatch task

pkarman:
connects #10051 

### Description

Appeals require an additional task be completed before being closed as "decided".


app/models/appeal.rb
```diff
@@ -673,6 +673,10 @@ def available_review_options
     %w[supplemental_claim cavc]
   end
 
+  def completed_bva_dispatch_task?
+    !!tasks.find { |t| t.is_a?(BvaDispatchTask) }&.completed?
```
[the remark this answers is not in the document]
aroltsch: I think we should only sync when all BvaDispatchTask tasks are completed. An appeal usually has two BvaDispatchTask tasks: one assigned to user and one assigned to the organization.

app/models/request_issue.rb
```diff
@@ -212,7 +212,7 @@ def end_product_code
   end
 
   def status_active?
-    return appeal_active? if decision_review.is_a?(Appeal)
+    return appeal_active? if decision_review_is_appeal?
```
leikkisa: Not related to this PR, but this doesn't seem to handle noncomp claim reviews.

app/models/request_issue.rb
```diff
@@ -440,7 +440,7 @@ def diagnostic_code
 
   def api_status_active?
     return decision_review.active_status? if decision_review.is_a?(ClaimReview)
-    return true if decision_review.is_a?(Appeal)
+    return true if decision_review_is_appeal?
```
leikkisa: Again, not really about your PR, but I don't think the api_status_active should just return true for an Appeal. I think that might be a remnant from Melissa not knowing how to check if an appeal was active.

app/models/request_issue.rb
```diff
@@ -534,7 +534,17 @@ def duplicate_of_issue_in_active_review?
...
+  def decision_review_is_appeal?
+    decision_review.is_a?(Appeal)
+  end
+
+  def decision_review_is_appeal_and_does_not_have_completed_bva_dispatch_task?
+    decision_review_is_appeal? && !decision_review.completed_bva_dispatch_task?
+  end
+
   def create_decision_issues
+    return false if decision_review_is_appeal_and_does_not_have_completed_bva_dispatch_task?
```
leikkisa: Haha. Just a nit but I think the level of abstraction for the two new methods above might not be necessary here (although it is funny that the method name is about as long as the code itself).

pkarman:
Turns out this case was caused by #9556 and would not be arrived at during any normal business flow.

leikkisa:
I have to admit, I don't understand enough. It looks like the BvaDispatchTask gets created for quality review and dispatch/outcoding, but I'm not sure if it gets created for every appeal.  I wonder if someone in Queue would know.  I also don't know if it matters, because maybe noncomp decision issues get created in a different flow.

I think asking Queue is a good step, but if we don't know it might be better to check that the BvaDispatchTask exists before checking that it's completed.

PR #10056, opened 2019-03-19
Add representative code to contested claimant

### LINKED ISSUE #9826: [IDT] Contested claimants missing representatives
Examples, 
vacols_id: 2695328, 3366989, 2469831, 2811717

```
SELECT * FROM REP WHERE REPKEY='2811717'
```
returns only one record that contains information about the contested claimant and their rep in the `repso` field

`repso` field contains a letter that stands for a service organization https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/vacols/case.rb#L46

AC: return representative for contested claimants for IDT

monfresh:
**Why**: IDT needs this info

Resolves #9826

### DIFF
--- a/app/models/vacols/representative.rb
+++ b/app/models/vacols/representative.rb
@@ -157,7 +157,7 @@ class VACOLS::Representative < VACOLS::Record
       address: get_address_from_rep_entry(self),
       representative: {
         code: code,
-        name: VACOLS::Case::REPRESENTATIVES[code][:full_name]
+        name: VACOLS::Case::REPRESENTATIVES.dig(code, :full_name)
       }
     }
   end
--- a/app/models/vacols/representative.rb
+++ b/app/models/vacols/representative.rb
@@ -146,13 +146,19 @@ class VACOLS::Representative < VACOLS::Record
              "Agent"
            end
 
+    code = repso
+
     {
       type: type,
       first_name: repfirst,
       middle_name: repmi,
       last_name: replast,
       name_suffix: repsuf,
-      address: get_address_from_rep_entry(self)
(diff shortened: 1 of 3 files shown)


app/models/vacols/representative.rb
```diff
@@ -152,7 +152,10 @@ def as_claimant
       middle_name: repmi,
       last_name: replast,
       name_suffix: repsuf,
-      address: get_address_from_rep_entry(self)
+      address: get_address_from_rep_entry(self),
+      representative: {
+        code: repso
```
aroltsch: Is Paul okay with us just sending the code instead of a human readable name?

aroltsch: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/vacols/case.rb#L46

app/models/vacols/representative.rb
```diff
@@ -146,13 +146,19 @@ def as_claimant
...
       type: type,
       first_name: repfirst,
       middle_name: repmi,
       last_name: replast,
       name_suffix: repsuf,
-      address: get_address_from_rep_entry(self)
+      address: get_address_from_rep_entry(self),
+      representative: {
+        code: code,
+        name: VACOLS::Case::REPRESENTATIVES[code][:full_name]
```
aroltsch: I think code could be blank in some cases.

monfresh: Could it be blank as in an empty string? If it's `nil`, that's already covered. I will update to handle the missing letter scenario.

monfresh: I updated it to use `dig`, which will handle all nonexistent keys.

PR #10063, opened 2019-03-20
Update capybara from 2.18.0 to 3.15.0

monfresh:
One of the major new changes is that when comparing text content on the
page, Capybara no longer fully normalizes whitespace, instead returning
text as close as possible to "as rendered". For example, if there is a
new line between text, or if the text in the spec has 2 whitespaces but
the rendered text only has one whitespace, the test will fail.

To make these tests pass, the spec needs to include the newline in the
expectation. For text that includes extra whitespace, we can either
remove the extra whitespace from the text in the spec, or if what we're
really testing is that the rendered text removes extra whitespace, we
can call `.squeeze(" ")` on the text in the expectation.

See the changes in this PR for examples of what expectations should look
like.

Because the Capybara API changed in 3.x, we also had to update our
sniffybara gem to match the new API.


spec/feature/reader/reader_spec.rb
```diff
@@ -234,7 +234,7 @@ def add_comment(text)
       scenario "Claims folder details issues show no issues message" do
         visit "/reader/appeal/#{appeal.vacols_id}/documents"
         find(".rc-collapse-header", text: "Claims folder details").click
-        expect(find("#claims-folder-issues").text).to have_content("No issues on appeal")
+        expect(page).to have_css("#claims-folder-issues", text: "No issues on appeal")
```
monfresh: This is the correct way to check for the text in an element that might have changed. Written this way, the expectation tells Capybara to wait for the element, and then wait until its value matches, up to the `Capybara.default_max_wait_time` setting.

The way the expectation was written before causes Capybara to only wait for the matching element, and immediately return whatever value it has, which might still be the old value.

Unsurprisingly, thoughtbot have an article on this topic: https://thoughtbot.com/blog/write-reliable-asynchronous-integration-tests-with-capybara

spec/feature/reader/reader_spec.rb
```diff
@@ -601,7 +601,7 @@ def expect_dropdown_filter_to_be_visible
           Generators::Annotation.create(
             comment: "hello world",
             document_id: documents[0].id,
-            y: 750
+            y: 1000
```
monfresh: This allows the test on line 766 to pass.

spec/feature/reader/reader_spec.rb
```diff
@@ -773,13 +766,12 @@ def placing_annotation_icon_position
...
-        click_on documents[1].type
-
-        expect(page).to have_content("IN THE APPEAL", wait: 10)
+        click_on documents[0].type
+        expect(page).to have_content(annotations[0].comment)
 
         expect(page).to have_css(".page")
         expect(find_field("page-progress-indicator-input").value).to eq "1"
-        scroll_to(class_name: "ReactVirtualized__Grid", value: 2000, index: 1)
+        find("#comment4").click
```
monfresh: For some reason, scrolling with `scroll_to` doesn't work in Capybara 3.x. Instead, what I did was place the comment on the second page (by changing the `y` value on line 604 above) so that when it is clicked, the page will automatically scroll to the second page.

Gemfile
```diff
@@ -102,7 +102,7 @@ group :test, :development, :demo do
   gem "rubocop", "~> 0.52", require: false
   gem "scss_lint", require: false
   gem "simplecov", git: "https://github.com/colszowka/simplecov.git", require: false
-  gem "sniffybara", git: "https://github.com/department-of-veterans-affairs/sniffybara.git", branch: "master"
+  gem "sniffybara", git: "https://github.com/department-of-veterans-affairs/sniffybara.git", branch: "mb-update-capybara-click"
```
pkarman: can we get that branch merged? or is there a backwards compatibility issue?

spec/feature/queue/attorney_checkout_flow_spec.rb
```diff
@@ -434,24 +434,27 @@
...
-        page.all("label", text: "Before certification")[1].click
-        page.all("label", text: "After certification")[2].click
+        # I know we're not supposed to sleep in tests, but this is the only
+        # thing that allows the tests to pass consistently. I think the issue is
+        # that after pressing "Continue" above, the page is moving and we have
+        # to wait until it stops moving before clicking on the checkboxes.
+        # Otherwise, it's not always able to click on the right checkboxes. If
+        # someone knows a better way to wait for the page to stop moving, please
+        # change this.
+        sleep 1
```
pkarman: would the helper `scroll_element_in_to_view` aid here?

monfresh:
Closes #10106

spec/feature/reader/reader_spec.rb
pkarman: would `scroll_element_in_to_view` work?

pkarman: or `safe_click`?

Gemfile
monfresh: I have to update all other repos that use sniffybara to Capybara 3.x first.

spec/feature/queue/attorney_checkout_flow_spec.rb
monfresh: I'll give it a shot.

monfresh: It's not working out of the box. We can explore this later, but I don't think it's worth it for this PR.

spec/feature/reader/reader_spec.rb
monfresh: I don't think so. What the test was doing before was scrolling the document to page 2 using `scroll_to(class_name: "ReactVirtualized__Grid", value: 2000, index: 1)`, so it's not scrolling any particular element into view. It's just trying to scroll down far enough to get to the second page.

What I did here is instead of trying to simulate a long scroll just with the "mouse", I updated the test so that the current document has a comment on page 2, and by clicking that comment, the page automatically ends up on page 2 and we can check the page number progress indicator.

In the grand scheme of things, I think testing things like this are unnecessary and not worth a slow integration test, but because this test already existed, I made it pass the easiest way I could think of.

monfresh: I just realized that this test was suffering from the same affliction as the one I mentioned earlier where we were calling `.value` on a `node` instead of using `expect(page).to_not have_field`. Also, Capybara 3.x now has built-in support for scrolling: https://www.rubydoc.info/gems/capybara/Capybara/Node/Element#scroll_to-instance_method

I was able to get this test to pass by removing our custom ` scroll_to` method and using this instead:
```ruby
all(".ReactVirtualized__Grid").last.scroll_to(0, 2000)
expect(page).to_not have_field("page-progress-indicator-input", with: "1") 
```

Issue #10041, opened 2019-03-18
RequestIssuesUpdate saves the wrong contention_reference_id when removing and adding a contention


leikkisa:
When removing a contention from a decision review during edit, and adding another contention at the same time, the removed contention's ID is saved as the contention ID on the new request issue.

#### Acceptance Criteria
In one edit session, remove a request issue that has a contention in VBMS, and add a new request issue.

The removed issue should have its contention removed from VBMS. The added request issue should have a new contention in VBMS. The correct contention_reference_id for the new request issue should be stored in the Request Issues table. 

Sentry example:
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/4103/events/277138/

This contention was already removed in RequestIssuesUpdate 132, and was attempted again in 133

pkarman:
See also #9800

pkarman:
these contention ids are being assigned to more than one request issue in production:
`["87456051", "87449256", "87271002", "87130493", "87069181", "87087863", "87252459", "87376149", "87536950", "87378529", "87378528", "87378527"]`

pkarman:
Research on the existing duplicates:

### 87456051 

same rating issue
should be 87559687

### 87449256 

both unidentified, different text
has 2 contentions, one with text of active RI [87559840] but submit date a month earlier, one with UNIDENTIFIED [87559706] with same submit date. both have same start date.

### 87271002

both unidentified, different text
has 3 contentions, one with UNIDENTIFIED [87530886] submit date of first RI, one with different text of active RI [87529397], one with text of removed original [87531629]

### 87130493

same rating issue
should be 87376331

### 87069181

same rating issue
should be 87194298

### 87087863

both unidentified, different text
has 3 contentions, all with "UNIDENTIFIED ISSUE - Please click "
87194818 - 5 march
87194817 - 5 march
87260694 - 7 march

### 87252459

same rating issue (active has value for rating_issue_associated_at)
has 2 contentions
87430458 - has rating text
87536803 - dated 2 days after RI

### 87376149

same rating issue (active has value for rating_issue_associated_at)
should be 87376623

### 87536950

same rating issue (active has value for rating_issue_associated_at)
has 2 contentions
87536948 - different text/date
87550278 - same text/date as rating

### 87378529

same rating issue (active has value for rating_issue_associated_at)
has 6 contentions
should be 87452297

 ### 87378528

same rating issue (active has value for rating_issue_associated_at)
should be 87452298

[...]

PR #10079, opened 2019-03-20
Fix contention_id matching bug

pkarman:
connects #10041 

### Description

When identical contention text is used, EPE can fail to correctly distinguish new vs old contentions and misidentify the correct contention id to store on the Request Issue record.

Also adds a unique index on the `contention_reference_id` since VBMS has confirmed that those values are globally unique.

### DIFF
--- a/app/models/claim_review.rb
+++ b/app/models/claim_review.rb
@@ -194,6 +194,10 @@ class ClaimReview < DecisionReview
     fetch_issues_status(issue_list)
   end
 
+  def contention_records(_epe)
+    request_issues.active
+  end
+
   private
 
   def incomplete_tasks?
--- a/app/models/claim_review.rb
+++ b/app/models/claim_review.rb
@@ -194,12 +194,12 @@ class ClaimReview < DecisionReview
     fetch_issues_status(issue_list)
   end
 
-  def contention_records(_epe)
-    request_issues.active
+  def contention_records(epe)
+    epe.request_issues.active
   end
 
-  def all_contention_records(_epe)
-    request_issues
+  def all_contention_records(epe)
+    epe.request_issues
   end
 
   private
--- a/app/models/claim_review.rb
+++ b/app/models/claim_review.rb
@@ -198,6 +198,10 @@ class ClaimReview < DecisionReview
     request_issues.active
   end
 
+  def all_contention_records(_epe)
+    request_issues
+  end
+
   private
 
   def incomplete_tasks?
--- a/app/models/end_product_establishment.rb
+++ b/app/models/end_product_establishment.rb
@@ -79,11 +79,11 @@ class EndProductEstablishment < ApplicationRecord
 
     set_establishment_values_from_source
 
-    contentions = records_ready_for_contentions.map do |issue|
-      contention = { description: issue.contention_text }
-      issue.try(:special_issues) && contention[:special_issues] = issue.special_issues
-      contention
-    end
+    contentions = build_contentions(records_ready_for_contentions)
+
+    # VBMS returns all the contentions on the claim, old and new, so keep track
+    # of existing contentions we know about. That way if text matches we can avoid false positives.
+    existing_contention_reference_ids = all_contention_records.pluck(:contention_reference_id).compact.uniq.map(&:to_s)
 
     # Currently not making any assumptions about the order in which VBMS returns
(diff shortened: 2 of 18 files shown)


db/migrate/20190320140634_add_updated_at_request_issue.rb
```diff
@@ -0,0 +1,5 @@
+class AddUpdatedAtRequestIssue < ActiveRecord::Migration[5.1]
+  def change
+    add_column :request_issues, :updated_at, :datetime
```
pkarman: this is useful in debugging.

lib/fakes/bgs_service.rb
```diff
@@ -192,7 +192,7 @@ def self.create_veteran_records
         sc
       when "has_higher_level_review_with_vbms_claim_id"
         claim_id = "600118951"
-        contention_reference_id = 1234
+        contention_reference_id = veteran.file_number[0..4] + "1234"
```
pkarman: must be unique

app/models/end_product_establishment.rb
```diff
@@ -79,20 +79,24 @@ def create_contentions!
...
 
     # We don't care about duplicate text; we just care that every request issue
     # has a contention.
     create_contentions_in_vbms(contentions).each do |contention|
+      next if existing_contention_reference_ids.include?(contention.id)
+
       record = records_ready_for_contentions.find do |r|
-        r.contention_text == contention.text && r.contention_reference_id.nil?
+        contention.claim_id == reference_id &&
+          r.contention_text == contention.text &&
```
leikkisa: I think this is still going to fail if there are multiple Unidentified Issues created at the same time, because we don't send specific text, we send the same text for all of them to VBMS. So maybe what you can do is also add the new contention_reference_id to the existing_contention_reference_ids after it gets created.

app/models/claim_review.rb
```diff
@@ -194,6 +194,14 @@ def issues_hash
     fetch_issues_status(issue_list)
   end
 
+  def contention_records(_epe)
+    request_issues.active
```
leikkisa: This isn't quite the same as putting
```request_issues.active``` in the EndProductEstablishment model, because the EPE might only have a subset of the entire claim review's request_issues, where here you're getting the ClaimReview's request_issues (which might be on multiple EPEs).

So you could use the epe argument
```epe.request_issues.active```

File: db/migrate/20190319144125_unique_contention_id.rb [hunk 5]
```diff
@@ -0,0 +1,13 @@
+class UniqueContentionId < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def up
+    remove_index :request_issues, [:contention_reference_id, :removed_at]
+    add_index :request_issues, :contention_reference_id, unique: true, algorithm: :concurrently
```
leikkisa: Do we need to fix the ones that are in Prod (and I guess maybe UAT) before adding this?

spec/models/request_issues_update_spec.rb
```diff
@@ -431,6 +431,45 @@
...
+
+        it "does not re-use contention_reference_id" do
+          expect(review.request_issues.pluck(:contention_reference_id).compact.uniq.count).to eq(2)
+          subject
+          review.reload
+          expect(review.request_issues.pluck(:contention_reference_id).compact.uniq.count).to eq(4)
+          expect(review.request_issues.active.count).to eq(2)
+        end
+      end
+
```
leikkisa: Can we add a test for adding two unidentified issues at the same time?

db/migrate/20190319144125_unique_contention_id.rb
pkarman: yes. I am tracking those and will fix before merging.

app/models/end_product_establishment.rb
leikkisa: On second thought, since it's looping through the new contentions, if they're getting created all in one call this might not be necessary.

pkarman: yeah I think as long as are limiting the matches to the current pool of created contentions, it doesn't matter if there are duplicate text strings. It's only when we match outside the current pool.

I've added a test to create 10 identical unidentified request issues, just to exercise this theory.

Issue #10059, opened 2019-03-19
Investigate insertBenefitClaim: City is null


leikkisa:
We are getting this error on establishing a claim.

```An error occurred while establishing the claim: Unable to establish claim: insertBenefitClaim: City is null```

We believe this is due to there being a non-Veteran claimant on the claim, and that dependent not having an address populated in the database.

Sentry example:
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3467/events/277716/

#### Acceptance criteria
If the Veteran chooses a non-Veteran claimant on the review page, we validate the claimant's address and show a descriptive error, and require the address to be valid before allowing the user to proceed.

#### Technical notes
According to VBMS, in the case of the claimants the address validation is just that it exists, and if it exists it assumes that it's valid. It does not necessarily follow the same rules as the Veteran address validation.

leikkisa:
Martina is going to ask the user to update the claimants address for the current case.

leikkisa:
In this specific case, the claimant shows up as one of the Veteran's relationships, but getting the information for the claimant results in:
```{:number_of_award_benes=>"0", :return_code=>"SHAR 0000", :return_message=>"No records found"}```

When I look at a successful call, the address isn't present but there is a mailing and payment address ID (and also, fun, there's a payee code!).

I'm wondering if I should just handle the "No records found" case for now without digging further into the address.

leikkisa:
I handled the nil address, since that was the case for this ticket and VBMS implied that if there's an address at all that is what they look for.  However depending on address validations downstream we may need to do more specific checks since the error we were getting implied it was about the city.  Preferably we could do some UAT testing to determine what is really needed.

Sjones352:
@leikkisa looking at this screenshot, do I create an EP before I can validate the claimant's address in UAT to see  a descriptive error?
<img width="1280" alt="Screen Shot 2019-03-26 at 11 01 36 AM" src="https://user-images.githubusercontent.com/23080951/55008030-c5a3cf80-4fb6-11e9-82d2-fe12f524490e.png">

leikkisa:
Sandra saw this error come up again today, so this requires more investigation still

https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/4174/events/283038/

leikkisa:
Martina has not heard back from the user.  I sent an email to Jennifer to see if she could add the claimant.

leikkisa:
[...]

PR #10080, opened 2019-03-21
Add error for missing Claimant address

leikkisa:
connects #10059


pkarman:
I had created #10074 for this yesterday too. I was trying to keep it more general unless you're 100% sure this nil city string is indeed the claimant issue.

leikkisa:
Closing as a dupe of https://github.com/department-of-veterans-affairs/caseflow/pull/10074

PR #10084, opened 2019-03-21
Fix default Capybara driver setting

monfresh:
**Why**: While testing PR #9503, I removed the conditional and forgot to
set it back before merging.


monfresh:
There was a single reader spec that kept failing. I looked at the screenshot and noticed the layout was different than when viewing the site in my browser full width for example. I thought perhaps the test was failing due to browser size. It worked this time, and hopefully it will continue to work.

Issue #9495, opened 2019-02-19
VSO Hearing Schedule | Daily Docket View 


Chingujo:
As a VSOs POA, I need to see the daily docket to see what veterans are coming for each hearing as well as the alternate hearing location. 

### Acceptance criteria 
_This view of the daily docket is exactly the same as the RO view of the hearing schedule except the following:_

- [x] We filter the daily docket by VSO. For example, if a DAV coordinator clicked on the daily docket for that day then they only see the veterans that have selected DAV as a POA. 
- [x] Verify that previously scheduled Veterans for that VSO are seen on the daily docket. 
- [x] Every field is read only, including the notes field that is writable in all other views. 

### Mocks
![daily docket-disabled dropdown single](https://user-images.githubusercontent.com/23316107/53169195-93174900-35aa-11e9-90f7-2f0256d4576c.png)

PR #10086, opened 2019-03-21
VSO hearing days

### LINKED ISSUE #9493: VSO Schedule - View Schedule
As a hearing coordinator for VSO's, I need to be able to see the up and coming schedule to see which of my veterans have up and coming hearing days to coordinate the day of and prior. 

### Acceptance Criteria
_Please note that VSO's reps will be navigating to the hearing schedule via Caseflow Queue toggle that currently exists in our hearing scheduling application. This view of the schedule is exactly the same as the RO schedule except the following:_

- [ ] Verify the following logic: 
- We sort the views of this schedule by VSO. Example: We look in the database for all hearings with a POA of DAV and this filtered list becomes the DAV hearing schedule. 
 - If in BGS a VSO coordinator has multiple accreditations, then they see the switch view toggle, that when clicked, shows the schedule for the other VSO"s they are accredited for. See switch view version of the mock 
- [ ] Verify the above switch view toggle has the same design and ux of the queue switch view toggle
- [ ] Verify that for the "Number of Hearings" column that it reflects the number of Veterans that are assigned to that POA/VSO only for that day (not the total number of Veterans on the docket). It should reflect only those veterans currently on the docket, not those that were previously scheduled for the day. 

### Mock
schedule view no filter

_If VSO coordinator has more than one BGS credential render the below toggle_
schedule view filter

sharonwarner:
Resolves #9493 #9495 

### Description
This PR adds VSO views for the daily docket and the view schedule page. It also includes some code cleanup for pulling hearing days from VACOLS.

My guess is that this will require further optimizations, but I'll include them in a separate PR.


app/models/hearing_day.rb
```diff
@@ -122,31 +122,36 @@ def update_schedule(updated_hearings)
       end
     end
 
+    def upcoming_days_for_vso_user(start_date, end_date, user)
+      HearingDay.includes(hearings: [appeal: :tasks])
+        .where("DATE(scheduled_for) between ? and ?", start_date, end_date).select do |hearing_day|
+        hearing_day.hearings.any? { |hearing| hearing.assigned_to_vso?(user) } ||
+          hearing_day.vacols_hearings.any? { |hearing| hearing.assigned_to_vso?(user) }
```
sharonwarner: This is currently going to make a request to vacols for every hearing day... will probably have to refactor.

app/repositories/hearing_day_repository.rb
```diff
@@ -15,17 +15,16 @@ def find_hearing_day(request_type, hearing_key)
       end
     end
 
-    def load_days_for_range(start_date, end_date)
-      video_and_co = VACOLS::CaseHearing.load_days_for_range(start_date, end_date)
-      travel_board = VACOLS::TravelBoardSchedule.load_days_for_range(start_date, end_date)
-      [video_and_co, travel_board]
+    def load_video_days_for_range(start_date, end_date)
```
mdbenjam: Nice change, makes this all clearer.

PR #10092, opened 2019-03-22
Adding service to dedup users

### LINKED ISSUE #10066: Bug: Judge doesn't see all hearing days assigned in Hearing Prep. Hearing Schedule Filter for Judge doesn't show all of Judges hearing days.
Judge Michael Herman sees different hearing days in Hearing Prep than we see in Hearing Schedule, even though Veterans are scheduled. One of the hearing days that the Judge was changed to (4/22), is not showing up in the filter for that judge on Hearing Schedule, but does show up inline. 

What Judge Herman can see for upcoming hearing days in Hearing Prep: 
Screen Shot 2019-03-20 at 10 06 26 AM

Judge Herman's hearing days in Hearing Schedule: 
Screen Shot 2019-03-20 at 10 06 40 AM

One we expect to show up: 
* 4/4/19 - https://appeals.cf.ds.va.gov/hearings/schedule/docket/87 - Judge Herman is assigned the Fargo docket, Veterans are scheduled 
* 4/22/19 - Docket for Boston - Judge Herman is on the docket, but this date doesn't appear in the VLJ filter: https://appeals.cf.ds.va.gov/hearings/schedule/docket/225 (this was a date change) 

Dates that should show up when a Veteran is scheduled: 
* 4/10/19 - https://appeals.cf.ds.va.gov/hearings/schedule/docket/131 - Judge Herman is assigned the Houston docket, but no Veterans are yet scheduled (I inquired and they will be postponing Veterans from a snow day to this day) 
* 4/17/19 - https://appeals.cf.ds.va.gov/hearings/schedule/docket/202 - Judge Herman is assigned St. Paul docekt, but not Veterans are scheduled (all postponed)

mdbenjam:
Resolves #10066

### Description
Creates a new UserDedup service which can be used to deduplicate a user's css_id. Currently we have lots of users that share the same css_id with different casing. i.e. "bvaabcde" and "BVAABCDE". We need to merge these users since only the all upper case version should be valid. This new service does this merging, reaches into every other model that has a foreign key into the User table and updates it. This service also stores all the changes that are made in a jsonb field on the merged User object, and the service provides a function to undo the change from the data in that jsonb object.

We'll start the process by changing one or two users and making sure everything is okay before we run the rake task to transform all of the users.

### Testing Plan
1. Added new tests to ensure we can merge records and undo the change when needed.


lib/tasks/users.rake
```diff
@@ -35,4 +35,16 @@ namespace :users do
       reporter.report.each { |ln| puts ln }
     end
   end
+
+  desc "deduplicates all user records merging all records with the same css_id with the capitalized one"
+  task dedupe_users: :environment do
```
pkarman: maybe this task should be called `merge` ? you'd call it like `rake users:merge`

app/services/user_reporter.rb
```diff
@@ -43,6 +43,14 @@ def report_user_related_records(user)
         related_records << "#{user.id} has #{num} #{cls}"
       end
     end
+
+    models_with_named_user_foreign_key.each do |foreign_key|
```
pkarman: is this method in this class?

app/services/user_dedup_service.rb
```diff
@@ -0,0 +1,105 @@
+# frozen_string_literal: true
```
pkarman: I get why you have a 2nd class here, but given how much code is copied and that we intend to throw this particular service away after the cleanup is done, how do you feel about putting the merging code into the `Reporter` class? I guess I don't care if we're going to use it once, but it seems a shame to copy/paste so much code.

mdbenjam: Yeah I originally wrote it all in the reporter but then decided to factor it out, in case we decide to use the reporter for something else. But I think you're right, that it makes more sense to put them together, since so much is reused. I'll go ahead and factor it back in!

app/services/user_reporter.rb
mdbenjam: Whoops missed this when I factored out the dedup service. I'll fix this back up.

app/services/user_reporter.rb
```diff
@@ -33,20 +81,51 @@ def models_with_user_id
...
+              end
+
+      undo_action = {
+        model: foreign_key[:model].name,
+        column: column_id_name,
+        ids: scope.pluck(:id),
+        user_id: old_user.id
+      }
+
+      scope.update(column_id_name => new_user.id)
```
pkarman: I think we want `update!` to raise an exception? or do we care if something fails?

pkarman:
> iirc there was a test for the updater class? 

ah I forgot. I was never able to merge it because of a weird coverage error. See #9997

PR #10100, opened 2019-03-22
Soft delete decision issues

### LINKED ISSUE #9818: Track when we delete decision issues
We should track decision issues that we delete

Options:
1. Introduce a separate table the same as request_issues_updates
2. Introduce soft delete to decision_issues table

aroltsch:
Resolves #9818 

### Description
Soft delete decision issues during checkout


app/models/concerns/issue_updater.rb
```diff
@@ -7,7 +7,7 @@ def update_issue_dispositions_in_caseflow!
     return unless appeal
 
     # We will always delete and re-create decision issues on attorney/judge checkout
-    appeal.decision_issues.destroy_all
+    appeal.decision_issues.each(&:soft_delete)
```
shanear: don't have to fix this in this PR, but I feel like this should be a DB transaction with other stuff in this method.

app/models/decision_issue.rb
```diff
@@ -30,6 +30,8 @@ class DecisionIssue < ApplicationRecord
   REMAND = "remanded"
   REMAND_DISPOSITIONS = [REMAND, DTA_ERROR_PMR, DTA_ERROR_FED_RECS, DTA_ERROR_OTHER_RECS, DTA_ERROR_EXAM_MO].freeze
 
+  default_scope { where(deleted_at: nil) }
```
shanear: maybe a comment to explain this? or excuse why we're using `default_scope` :)

app/models/decision_issue.rb
```diff
@@ -61,6 +63,11 @@ def uncontested
     end
   end
 
+  def soft_delete
+    update(deleted_at: Time.zone.now)
+    request_decision_issues.update_all(deleted_at: Time.zone.now)
```
shanear: these should probably be a transaction

app/models/decision_issue.rb
```diff
@@ -69,9 +76,9 @@ def issue_category
     associated_request_issue&.issue_category
   end
 
-  def destroy_on_removed_request_issue(request_issue_id)
-    # destroy if the request issue is deleted and there are no other request issues associated
-    destroy if request_issues.length == 1 && request_issues.first.id == request_issue_id
+  def mark_deleted_on_removed_request_issue
```
shanear: why not `soft_delete_on_removed_request_issue`?

spec/models/attorney_case_review_spec.rb
```diff
@@ -65,10 +65,14 @@
       it "should create and delete decision issues" do
         subject
         old_decision_issue_ids = [decision_issue1.id, decision_issue2.id, decision_issue3.id, decision_issue4.id]
-        old_remand_reasons_ids = [remand_reason1.id, remand_reason2.id]
         expect(DecisionIssue.where(id: old_decision_issue_ids)).to eq []
         expect(RequestDecisionIssue.where(decision_issue_id: old_decision_issue_ids)).to eq []
-        expect(RemandReason.where(id: old_remand_reasons_ids)).to eq []
+
+        # Ensure soft deleted records are there
```
shanear: nice 👍

app/models/decision_issue.rb
aroltsch: It is in a very big one transaction that is outside of this file.

app/models/concerns/issue_updater.rb
aroltsch: It is part of the transaction that is outside of this file

PR #10110, opened 2019-03-22
align postgresql version to our AWS envs

pkarman:
We can get up to 9.5 in AWS RDS production so making sure we test and dev against the same version.


monfresh:
@enriquemanuel Can we only install exactly version 9.5, or can we install the latest in the 9.5 series, which is 9.5.16? If we only specify `9.5` in this PR, does it mean we can install any minor version?

pkarman:
In my local testing for docker, if I specified 9.5 it installs the latest patch version 9.5.16. I think if we want to automatically get the latest patch in docker and circle, this is ideal for maintenance purposes.

For AWS RDS, I believe patch versions are made available as they are released and can upgraded w/o significant hassle. IMO that could be a regular maintenance window case, which would happen after circle/docker has already exercised the patch.

PR #10120, opened 2019-03-22
Switch Claimant references to review_request to decision_review

leikkisa:
connects #9559


db/schema.rb
```diff
@@ -212,15 +212,12 @@
...
   end
 
-  create_table "claimants", force: :cascade do |t|
-    t.bigint "decision_review_id"
-    t.string "decision_review_type"
-    t.string "participant_id", null: false
-    t.string "payee_code"
-    t.bigint "review_request_id", null: false
-    t.string "review_request_type", null: false
+  create_table "claimants", force: :cascade, comment: "The claimant for each Decision Review, and its payee_code if required." do |t|
```
shanear: claimants is a weird table because it doesn't necessarily represent the claimant but the joining relationship of a person as a claimant on the claim, if that makes sense.

The best way to represent the difference is that if the same person is a claimant on two different claims, there will be a seperate claimant record for each claim, even though both records will map to the same person.

I think we should capture this in the comments.

db/schema.rb
```diff
@@ -212,15 +212,12 @@
...
+    t.bigint "decision_review_id", comment: "The ID of the Decision Review the claimant is on."
+    t.string "decision_review_type", comment: "The type of Decision Review the claimant is on."
+    t.string "participant_id", null: false, comment: "The participant ID of the claimant selected on a Decision Review."
+    t.string "payee_code", comment: "The payee_code for the claimant selected on a Decision Review, if applicable. Payee_code is required for claimants when the claim is processed in VBMS and the Veteran is not the claimant. For Supplemental Claims automatically created due to remanded decisions, this is automatically selected based on the last payee code used for the same claimant on a previous end product, if available."
```
shanear: > Payee_code is required for claimants when the claim is processed in VBMS and the Veteran is not the claimant

Payee code is required period. It's just always "00" if the veteran is the claimant.

shanear: > For Supplemental Claims automatically created due to remanded decisions, this is automatically selected based on the last payee code used for the same claimant on a previous end product, if available.

Not sure if we want to get into this in the comment. This is something that has a decent chance of changing at some point, and if it were to change, there's no way that person would know to update this.

PR #10121, opened 2019-03-22
Migration to remove review_request

leikkisa:
connects #9559

Combined into https://github.com/department-of-veterans-affairs/caseflow/pull/10120, will leave open for now in case anyone feels that they should be separated.


leikkisa:
Closing, combined into another PR.

PR #10122, opened 2019-03-22
Add comments to decision issues

leikkisa:
connects #9559 

Will run the actual migration after review comments.

### DIFF
--- a/db/migrate/20190322225131_add_comments_to_decision_issues.rb
+++ b/db/migrate/20190322225131_add_comments_to_decision_issues.rb
@@ -1,0 +1,33 @@
+class AddCommentsToDecisionIssues < ActiveRecord::Migration[5.1]
+  def change
+    change_table_comment(:decision_issues, "Issues that represent a decision made on a Decision Review's request_issue.")
+
+    change_column_comment(:decision_issues, :benefit_type, "The Benefit Type, also known as the Line of Business for a decision issue. For example, compensation, pension, or education.")
+
+    change_column_comment(:decision_issues, :caseflow_decision_date, "This is a decision date for decision issues where decisions are entered in Caseflow, such as for Appeals or for Decision Reviews with a business line that is not processed in VBMS.")
+
+    change_column_comment(:decision_issues, :created_at, "Automatic timestamp when row was created")
+
+    change_column_comment(:decision_issues, :decision_review_id, "The ID of the Decision Review with the Request Issue that was decided in order to create this Decision Issue.")
+
+    change_column_comment(:decision_issues, :decision_review_type, "The type of the Decision Review with the Request Issue that was decided in order to create this Decision Issue.")
+
+    change_column_comment(:decision_issues, :decision_text, "If decision issue is connected to a rating, this is the rating issue's decision text")
+
+    change_column_comment(:decision_issues, :description, "Optional description that user can input")
+
+    change_column_comment(:decision_issues, :diagnostic_code, "If decision issue is connected to a rating, this is the rating issue's diagnostic code")
+
+    change_column_comment(:decision_issues, :disposition, "The disposition for a decision issue, for example 'granted' or 'denied'.")
+
+    change_column_comment(:decision_issues, :end_product_last_action_date, "After an End Product gets synced with a status of CLR (cleared), we save the End Product's last_action_date on any Decision Issues that are created as a result. We use this as a proxy for decision date for non-rating issues that were processed in VBMS because they don't have a rating profile date, and we do not have access to the exact decision date.")
+
+    change_column_comment(:decision_issues, :participant_id, "Veteran's participant id")
+
+    change_column_comment(:decision_issues, :profile_date, "If a decision issue is connected to a rating, this is the profile_date of that rating. The profile_date is used as an identifier for the rating, and is the date we believe that the Veterans think of as the decision date.")
+
+    change_column_comment(:decision_issues, :promulgation_date, "If a decision issue is connected to a rating, it will have a promulgation date. This represents the date that the decision is legally official. It is different than the decision date. It is used for calculating whether a decision issue is within the timeliness window to be appealed or get a higher level review.")
+
+    change_column_comment(:decision_issues, :rating_issue_reference_id, "If the decision issue is connected to the rating, this ID identifies the specific issue on the rating that is connected to the decision issue (a rating can have multiple issues). This is unique per rating issue.")
+  end
+end


db/migrate/20190322225131_add_comments_to_decision_issues.rb
```diff
@@ -0,0 +1,33 @@
+class AddCommentsToDecisionIssues < ActiveRecord::Migration[5.1]
+  def change
+    change_table_comment(:decision_issues, "Issues that represent a decision made on a Decision Review's request_issue.")
```
shanear: Are we using underscores in our comments? Also, I don't think we need `Decision Review's` in the statement. It sort of clutters it up:

```
Issues that represent a decision made on a request issue.
```

db/migrate/20190322225131_add_comments_to_decision_issues.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    change_column_comment(:decision_issues, :benefit_type, "The Benefit Type, also known as the Line of Business for a decision issue. For example, compensation, pension, or education.")
```
shanear: I think the business line and the benefit type map 1 to 1, but they are not synonyms. Maybe something like this:

```
Classification of the benefit being decided on. Maps 1 to 1 to VA lines of business, and typically used to know which line of business the decision correlates to.
```

db/migrate/20190322225131_add_comments_to_decision_issues.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    change_column_comment(:decision_issues, :caseflow_decision_date, "This is a decision date for decision issues where decisions are entered in Caseflow, such as for Appeals or for Decision Reviews with a business line that is not processed in VBMS.")
+
+    change_column_comment(:decision_issues, :created_at, "Automatic timestamp when row was created")
+
+    change_column_comment(:decision_issues, :decision_review_id, "The ID of the Decision Review with the Request Issue that was decided in order to create this Decision Issue.")
```
shanear: We may not want to be so specific here

Also I'm not sure we want to be capitalizing `Decision Review`?

```
ID of the decision review the decision was made on.
```

```
Type of the decision review the decision was made on.
```

db/migrate/20190322225131_add_comments_to_decision_issues.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    change_column_comment(:decision_issues, :decision_review_type, "The type of the Decision Review with the Request Issue that was decided in order to create this Decision Issue.")
+
+    change_column_comment(:decision_issues, :decision_text, "If decision issue is connected to a rating, this is the rating issue's decision text")
```
shanear: I think `connected` is too vague of a word.

```
If decision resulted in a change to a rating, this is the rating issue's decision text.
```

db/migrate/20190322225131_add_comments_to_decision_issues.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    change_column_comment(:decision_issues, :description, "Optional description that user can input")
```
shanear: Maybe specify that this is only for decision issues entered in Caseflow?

db/migrate/20190322225131_add_comments_to_decision_issues.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    change_column_comment(:decision_issues, :diagnostic_code, "If decision issue is connected to a rating, this is the rating issue's diagnostic code")
```
shanear: Same as the comment above

```
If decision resulted in a change to a rating, this is the rating issue's diagnostic code.
```

db/migrate/20190322225131_add_comments_to_decision_issues.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    change_column_comment(:decision_issues, :disposition, "The disposition for a decision issue, for example 'granted' or 'denied'.")
```
shanear: One thing that's sort of weirded me out is that the VBMS dispositions and the Caseflow dispositions are different. Do we want to get into examples in the definition? Since for comp HLRs and SCs the dispositions will be "Allowed" and "Denied"

db/migrate/20190322225131_add_comments_to_decision_issues.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    change_column_comment(:decision_issues, :end_product_last_action_date, "After an End Product gets synced with a status of CLR (cleared), we save the End Product's last_action_date on any Decision Issues that are created as a result. We use this as a proxy for decision date for non-rating issues that were processed in VBMS because they don't have a rating profile date, and we do not have access to the exact decision date.")
```
shanear: nice 👍

db/migrate/20190322225131_add_comments_to_decision_issues.rb
pkarman: definitely worth a mention. if we're treating it like an enum but with synonymous values, that will confuse people.

db/migrate/20190322225131_add_comments_to_decision_issues.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    change_column_comment(:decision_issues, :participant_id, "Veteran's participant id")
+
+    change_column_comment(:decision_issues, :profile_date, "If a decision issue is connected to a rating, this is the profile_date of that rating. The profile_date is used as an identifier for the rating, and is the date we believe that the Veterans think of as the decision date.")
```
shanear: Same as comment above on `connected` vs `resulted in`

Not sure we want to be putting `we believe` in the DB notes. Muddies up the definition, and people reading this may not know who `we` is.

```
The profile_date is used as an identifier for the rating, and is the date that most closely maps to what Veterans write down as the decision date.
```

db/migrate/20190322225131_add_comments_to_decision_issues.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    change_column_comment(:decision_issues, :promulgation_date, "If a decision issue is connected to a rating, it will have a promulgation date. This represents the date that the decision is legally official. It is different than the decision date. It is used for calculating whether a decision issue is within the timeliness window to be appealed or get a higher level review.")
```
shanear:
```
This represents the date that the decision is legally official.
```

Will probably want confirmation on this. I'm not sure if this is true.

shanear: Also `connected` vs `resulted in`

db/migrate/20190322225131_add_comments_to_decision_issues.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    change_column_comment(:decision_issues, :rating_issue_reference_id, "If the decision issue is connected to the rating, this ID identifies the specific issue on the rating that is connected to the decision issue (a rating can have multiple issues). This is unique per rating issue.")
```
shanear: `connected` vs `resulted in`, but otherwise looks great

leikkisa:

PR #10123, opened 2019-03-22
EPE table comments

leikkisa:
connects #9559

I will run the migration after review comments.

### DIFF
--- a/db/migrate/20190322230352_add_comments_to_end_product_establishments.rb
+++ b/db/migrate/20190322230352_add_comments_to_end_product_establishments.rb
@@ -1,0 +1,41 @@
+class AddCommentsToEndProductEstablishments < ActiveRecord::Migration[5.1]
+  def change
+    change_table_comment(:end_product_establishments, "Keeps track of End Products that need to be established for AMA Decision Review Intakes, when they are successfully established, and updates on the End Product's status.")
+
+    change_column_comment(:end_product_establishments, :benefit_type_code, "The benefit_type_code is 1 if the Veteran is alive, and 2 if the Veteran is deceased. Not to be confused with benefit_type, which is unrelated.")
+
+    change_column_comment(:end_product_establishments, :claim_date, "The claim_date for End Products established is set to the receipt date of the form.")
+
+    change_column_comment(:end_product_establishments, :claimant_participant_id, "The participant ID of the claimant submitted on the End Product.")
+
+    change_column_comment(:end_product_establishments, :code, "The end product code, which determines the type of end product that is established. For example, it can contain information about whether it is rating, nonrating, compensation, pension, created automatically due to a Duty to Assist Error, and more.")
+
+    change_column_comment(:end_product_establishments, :committed_at, "Timestamp indicating other actions performed as part of a larger atomic operation containing the end product establishment, such as creating contentions, are also complete.")
+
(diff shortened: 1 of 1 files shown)


db/migrate/20190322230352_add_comments_to_end_product_establishments.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    change_column_comment(:end_product_establishments, :last_synced_at, "The time that the status of the End Product was last synced with BGS. Once an End Product is cleared or canceled, it will stop being synced.")
+
+    change_column_comment(:end_product_establishments, :modifier, "The end product modifier. For Higher Level Reviews, the modifiers range from 030-039. For Supplemental Claims, they range from 040-049. The same modifier cannot be used twice for an active end product per Veteran.  Once an End Product is no longer active, the modifier can be used again.")
+
+    change_column_comment(:end_product_establishments, :payee_code, "The payee_code of the claimant submitted for this End Product.")
+
+    change_column_comment(:end_product_establishments, :reference_id, "The claim_id of the End Product, which is stored after the End Product is successfully established in VBMS")
+
+    change_column_comment(:end_product_establishments, :source_id, "The ID of the Decision Review that the end product establishment is connected to.")
```
pkarman: should we rename this to `decision_review` to be consistent with other relationships?

db/migrate/20190322230352_add_comments_to_end_product_establishments.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    change_column_comment(:end_product_establishments, :source_type, "The type of Decision Review that the End Product Establishment is for, for example HigherLevelReview.")
+
+    change_column_comment(:end_product_establishments, :station, "The station ID of the End Product's station.")
```
pkarman: should we rename to `station_id` or does that imply a local FK?

db/migrate/20190322230352_add_comments_to_end_product_establishments.rb
```diff
@@ -0,0 +1,41 @@
+class AddCommentsToEndProductEstablishments < ActiveRecord::Migration[5.1]
+  def change
+    change_table_comment(:end_product_establishments, "Keeps track of End Products that need to be established for AMA Decision Review Intakes, when they are successfully established, and updates on the End Product's status.")
```
shanear:
```
Represents end products that have been, or need to be established by Caseflow. Used to track the status of those end products as they are processed in VBMS and/or SHARE.
```

db/migrate/20190322230352_add_comments_to_end_product_establishments.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    change_column_comment(:end_product_establishments, :benefit_type_code, "The benefit_type_code is 1 if the Veteran is alive, and 2 if the Veteran is deceased. Not to be confused with benefit_type, which is unrelated.")
```
shanear: Could be shortened a tad:
```
1 if the Veteran is alive, and 2 if the Veteran is deceased. Not to be confused with benefit_type, which is unrelated.
```

db/migrate/20190322230352_add_comments_to_end_product_establishments.rb
shanear: I think the docs here are a little too specific to decision reviews for this generic class. EPEs could be used for anything in the future.

leikkisa:

PR #10124, opened 2019-03-22
Add intake table comments

leikkisa:
connects #9559

Will run migration after review comments.

### DIFF
--- a/db/migrate/20190322231353_add_comments_to_intake.rb
+++ b/db/migrate/20190322231353_add_comments_to_intake.rb
@@ -1,0 +1,29 @@
+class AddCommentsToIntake < ActiveRecord::Migration[5.1]
+  def change
+    change_table_comment(:intakes, "Keeps track of the initial intake of all Decision Reviews and RAMP Reviews.")
+
+    change_column_comment(:intakes, :cancel_other, "The additional notes a Claims Assistant can enter if they are canceling an intake for a reason other than the options presented.")
+
+    change_column_comment(:intakes, :cancel_reason, "The reason a Claim Assistant is canceling the current intake. Intakes can also be canceled automatically when there is an uncaught error, with the reason 'system_error'.")
+
+    change_column_comment(:intakes, :completed_at, "Timestamp for when the Intake was completed, whether it was successful or not.")
+
+    change_column_comment(:intakes, :completion_started_at, "Timestamp for when the user submitted the Intake to be completed.")
+
+    change_column_comment(:intakes, :completion_status, "Indicates whether the intake was successful, or was closed by being canceled, expired, or due to an error.")
+
+    change_column_comment(:intakes, :detail_id, "The ID of the Decision Review or RAMP Review that the Intake is connected to.")
+
+    change_column_comment(:intakes, :detail_type, "The type of Decision Review or RAMP Review that the Intake is connected to.")
+
+    change_column_comment(:intakes, :error_code, "If the Intake was unsuccessful due to a set of known errors, the error code is stored here. An error is also stored here for RAMP Elections that are connected to a currently active End Product, even though the Intake is a success.")
+
+    change_column_comment(:intakes, :started_at, "Timestamp for when the Intake was created, which happens when a Claims Assistant successfully searches for a Veteran in the Intake app.")
+
+    change_column_comment(:intakes, :type, "The type of Intake this is, for example if the detail type is Appeal, the intake type is AppealIntake.")
+
+    change_column_comment(:intakes, :user_id, "The CSS_ID of the user who created the intake.")
+
+    change_column_comment(:intakes, :veteran_file_number, "The file number of the Veteran which the Intake is for.")
+  end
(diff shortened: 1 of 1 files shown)


db/migrate/20190322231353_add_comments_to_intake.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    change_column_comment(:intakes, :cancel_reason, "The reason a Claim Assistant is canceling the current intake. Intakes can also be canceled automatically when there is an uncaught error, with the reason 'system_error'.")
+
+    change_column_comment(:intakes, :completed_at, "Timestamp for when the Intake was completed, whether it was successful or not.")
+
+    change_column_comment(:intakes, :completion_started_at, "Timestamp for when the user submitted the Intake to be completed.")
+
+    change_column_comment(:intakes, :completion_status, "Indicates whether the intake was successful, or was closed by being canceled, expired, or due to an error.")
+
+    change_column_comment(:intakes, :detail_id, "The ID of the Decision Review or RAMP Review that the Intake is connected to.")
```
pkarman: should we rename to `decision_review` or `source` to match EPE?

db/migrate/20190322231353_add_comments_to_intake.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    change_column_comment(:intakes, :detail_type, "The type of Decision Review or RAMP Review that the Intake is connected to.")
+
+    change_column_comment(:intakes, :error_code, "If the Intake was unsuccessful due to a set of known errors, the error code is stored here. An error is also stored here for RAMP Elections that are connected to a currently active End Product, even though the Intake is a success.")
+
+    change_column_comment(:intakes, :started_at, "Timestamp for when the Intake was created, which happens when a Claims Assistant successfully searches for a Veteran in the Intake app.")
+
+    change_column_comment(:intakes, :type, "The type of Intake this is, for example if the detail type is Appeal, the intake type is AppealIntake.")
```
pkarman: could we use the phrase "class name" instead of "type" to describe the "type" column? It has special meaning in Rails for polymorphic pattern.

db/migrate/20190322231353_add_comments_to_intake.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    change_column_comment(:intakes, :user_id, "The CSS_ID of the user who created the intake.")
```
pkarman: this is not the CSS ID. This is the `id` of the `User` record.

db/migrate/20190322231353_add_comments_to_intake.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    change_column_comment(:intakes, :veteran_file_number, "The file number of the Veteran which the Intake is for.")
```
pkarman: I added the TODO comment in the spreadsheet for veteran participant id.

db/migrate/20190322231353_add_comments_to_intake.rb
```diff
@@ -0,0 +1,29 @@
+class AddCommentsToIntake < ActiveRecord::Migration[5.1]
+  def change
+    change_table_comment(:intakes, "Keeps track of the initial intake of all Decision Reviews and RAMP Reviews.")
```
shanear: Should be more vague:

```
Represents the intake of an form or request made by a veteran.
```

db/migrate/20190322231353_add_comments_to_intake.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    change_column_comment(:intakes, :cancel_other, "The additional notes a Claims Assistant can enter if they are canceling an intake for a reason other than the options presented.")
```
shanear: I think we should use `user` instead of `Claims Assistant`, since it's not always a CA doing intakes (most of the time it is, but not always)

```
Notes that can be added if a user needs to cancel an intake for any reason other than the stock set of options presented.
```

db/migrate/20190322231353_add_comments_to_intake.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    change_column_comment(:intakes, :cancel_reason, "The reason a Claim Assistant is canceling the current intake. Intakes can also be canceled automatically when there is an uncaught error, with the reason 'system_error'.")
```
shanear: `Claim Assistant` => `user`

```
The reason the intake was canceled. Could have been manually canceled by a user, or automatic.
```

db/migrate/20190322231353_add_comments_to_intake.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    change_column_comment(:intakes, :completed_at, "Timestamp for when the Intake was completed, whether it was successful or not.")
```
shanear: `Intake` => `intake`

leikkisa:

PR #10125, opened 2019-03-22
Ramp Election table comments

leikkisa:
connects #9559

Will run migration after review comments.


leikkisa:

Issue #9679, opened 2019-02-27
Intake | Improved error messaging


pkarman:
When users report the "Something went wrong" error, they often take a screenshot to accompany it.

As a user, when I get the "Something went wrong" error, I would always find it helpful to know a little more detail, particularly if it is something that helps me know (a) it wasn't my fault and (b) what I might be able to do (e.g. retry in the case of a transient upstream error).

As an engineer, it would be helpful if the error message contained just enough information to help me more quickly narrow down the actual error. A unique ID (even if it was random) that was logged as well as displayed to the screen, might help me search Sentry and/or filter the AWS logs to the specific stacktrace or exception message.

## Acceptance criteria

* Every "Something went wrong" message should contain at least one more line of detail, even if it is a randomly generated UUID.

Sjones352:
@leikkisa validated this Issue. Please refer to https://github.com/department-of-veterans-affairs/caseflow/pull/10132
<img width="744" alt="Screen Shot 2019-03-28 at 8 01 44 PM" src="https://user-images.githubusercontent.com/23080951/55200418-5e437680-5194-11e9-88f2-669fb70d9c9e.png">

PR #10132, opened 2019-03-23
Add error UUID to generic client error messages

pkarman:
connects #9679 

Help connect what the user sees (and the screensnaps they attach to support tickets) with Sentry and Rails logs. Makes it easier to search for errors using a unique string.

Screen Shot 2019-03-23 at 8 46 27 AM
Screen Shot 2019-03-23 at 10 11 11 AM
Screen Shot 2019-03-25 at 11 25 33 AM


app/controllers/intakes_controller.rb
```diff
@@ -22,6 +22,11 @@ def create
         error_data: new_intake.error_data
       }, status: :unprocessable_entity
     end
+  rescue StandardError => error
+    Raven.capture_exception(error, extra: { error_uuid: error_uuid })
+    Rails.logger.error("Error UUID #{error_uuid} : #{error}")
+    # we name the variable error_code to re-use the client error handling.
+    render json: { error_code: error_uuid }, status: :internal_server_error
```
monfresh: What do you think about moving the Raven and Rails.logger lines into a method called `log_error(error)`? This will eliminate the duplication in the `review` and `create` methods and get rid of one of the Code Climate warnings.

Also, do you know why we use the singular `error_code` in `create` and plural `error_codes` in `review`? If we can make those consistent, we can move all 3 lines into a shared method and eliminate all Code Climate warnings.

pkarman: the client code expects the 2 different variables, as they are used at 2 distinct steps in the intake process.

I'm not against refactoring the client code but that would be more invasive than what I'm trying to accomplish here.

mdbenjam:
Out of scope, but I think this would be a great if we could use React error boundaries to do the same thing for frontend errors. Then the screen wouldn't ever go blank if there was an error in our React code. We'd display "something went wrong" with a uuid for the error. This is a great start!

leikkisa:
Validated:
On HLR 1323 created two request issues, contention reference IDs: 318215 and 318216
On edit, removed those issues and created two new issues that were otherwise identical.
Created 318217 and 318218.

PR #10134, opened 2019-03-25
Board Grant Effectuations comments

pkarman:
connects #9559

### DIFF
--- a/db/migrate/20190325152148_board_grant_effectuation_comments.rb
+++ b/db/migrate/20190325152148_board_grant_effectuation_comments.rb
@@ -1,0 +1,15 @@
+class BoardGrantEffectuationComments < ActiveRecord::Migration[5.1]
+  def change
+    change_table_comment(:board_grant_effectuations, "Tracks all the effects of a Board Grant decision made in Caseflow.")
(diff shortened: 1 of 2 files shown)


db/migrate/20190325152148_board_grant_effectuation_comments.rb
```diff
@@ -0,0 +1,15 @@
+class BoardGrantEffectuationComments < ActiveRecord::Migration[5.1]
+  def change
+    change_table_comment(:board_grant_effectuations, "Tracks all the effects of a Board Grant decision made in Caseflow.")
```
leikkisa: Here's a new description for this:
```Represents the work item of updating records in response to a granted issue on a Board appeal. Some are represented as contentions on an EP in VBMS. Others are tracked via Caseflow tasks.```

File: lib/tasks/local/vacols.rake [hunk 2]
```diff
@@ -23,6 +26,10 @@ namespace :local do
         sleep 1
       end
       # rubocop:enable Lint/HandleExceptions
+
+      unless facols_is_ready
+        fail "Gave up waiting for FACOLS to get ready and we won't leave alone"
+      end
```
lowellrex: Looks good to me!

PR #10137, opened 2019-03-25
Enable user merge to work on unique constraints

### LINKED ISSUE #10066: Bug: Judge doesn't see all hearing days assigned in Hearing Prep. Hearing Schedule Filter for Judge doesn't show all of Judges hearing days.

mdbenjam:
Resolves #10066

### Description
When trying to merge two of our duplicated users we ran into the problem that both users had AppealView records pointing to the same Appeal. Our database has a unique constraint on AppealView where by there can be only one record for each combination of Appeal and User. 
The following tables have unique constraints:

```
AppealView
HearingView
DocumentView
ReaderUser
UserQuota
```

This PR updates it so that any rows that would violate the unique constraint if they were to be merged, are instead deleted. I have not added these deleted rows to the undo blob for two reasons. Adding them would be difficult, and space intensive, since it's not just a matter of mapping old ids to new ids, but storing whole records. Second, the above tables are more for statistics and so if we lose records, it is probably okay. The only table I'm unsure of is the UserQuota table, which is used for a single feature in dispatch, but it believe that if we're going to lose a record it will be okay.

### Testing Plan
1. Add a new automated test for this condition.


app/services/user_reporter.rb
```diff
@@ -105,6 +109,25 @@ def report_user_related_records(user)
     related_records
   end
 
+  def delete_unique_constraint_violating_records(scope, foreign_key, fk_column_name, old_user, new_user)
+    if foreign_key[:unique]
+      if foreign_key[:unique].empty?
```
pkarman: the conditional nesting here is a little hard to ready. since there is no top-level `else` could we just combine the first 2 conditions?

mdbenjam: On second thought these are two different things. We don't want to do any of this if `foreign_key[:unique]` is not present. I'm going to open up another PR to undo that.

PR #10141, opened 2019-03-25
Appeal Status: Filter out remanded Supplemental Claims

### LINKED ISSUE #10140: Appeal Status API: Filter out remanded Supplemental Claims
Current bug: The remanded supplemental claims are shown as a separate claim.

The remanded supplemental claims should not be shown separately. Their information and decisions are displayed as part of the original HLR or Appeal that has the remanded decision.

They should be filtered out when the controller gets the list of supplemental claims associated with the Veteran.

mc-usds:
Resolves #10140

### Description
Update so that Appeal Status API does include remanded supplemental claims in the list. The information for regarding the status of the remanded supplemental claim is included in the origianal decision review.

### Acceptance Criteria
- [x] Code compiles correctly

### Testing Plan
1. updated feature test.


spec/feature/api/v2/appeals_spec.rb
```diff
@@ -623,4 +623,89 @@
       expect(json["data"].length).to eq(0)
     end
   end
+
+  context "Remanded SC filtered out" do
+    before do
+      Timecop.freeze(Time.utc(2018, 11, 28))
```
pkarman: could we use `pre_ama_start_date` unless the date 2018/11/28 is significant?

spec/feature/api/v2/appeals_spec.rb
```diff
@@ -623,4 +623,89 @@
...
+    end
+
+    let(:veteran_file_number) { "111223333" }
+    let(:receipt_date) { Date.new(2018, 9, 20) }
```
pkarman: is this date important? could we make it relative? `Time.zone.today - 30.days` e.g.

spec/feature/api/v2/appeals_spec.rb
```diff
@@ -623,4 +623,89 @@
...
+    let!(:dta_ep) do
+      create(:end_product_establishment, :active, source: dta_sc)
+    end
+
+    before do
+      allow_any_instance_of(Fakes::BGSService).to receive(:fetch_file_number_by_ssn) do |_bgs|
+        veteran_file_number
+      end
+
+      FeatureToggle.enable!(:api_appeal_status_v3)
```
pkarman: status_v3 and URL routes like v2? is that right?

app/controllers/api/v2/appeals_controller.rb
```diff
@@ -48,7 +48,10 @@ def hlrs
   end
 
   def supplemental_claims
+    # Filter out remanded SC because status and information of those are display through
+    # the original HLR or Appeal
     @supplemental_claims ||= SupplementalClaim.where(veteran_file_number: veteran_file_number)
+      .where(decision_review_remanded: nil)
```
pkarman: this seems like it might be a useful scope on the `SupplementalClaim` class itself? Not sure what to name it though. Find to leave as-is for this PR.

spec/feature/api/v2/appeals_spec.rb
mc-usds: Yeah, originally I thought we were going to bump the version number, but we didn't need to and I didn't change the name of the feature flag.

spec/feature/api/v2/appeals_spec.rb
mc-usds: Fixed. It has no significance. I think I grabbed the date from an existing test at some point.

spec/feature/api/v2/appeals_spec.rb
mc-usds: Fixed

Issue #10004, opened 2019-03-15
Separate construction of case timeline data and UI presentation


lowellrex:
Currently we determine what values should populate each field of a case timeline entry at the same place in the code where we determine how those fields should be presented. This causes [the code that does both of these things](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/components/TaskRows.jsx) to be hard to comprehend and [bug prone](https://github.com/department-of-veterans-affairs/caseflow/issues/9968).

An element of the case timeline has 5 attributes that we determine from characteristics of the related task. We should determine each of these attributes first, then move on to displaying them
<img width="603" alt="case timeline element diagram" src="https://user-images.githubusercontent.com/32683958/54448381-f7e53f80-4721-11e9-8d18-9fc7f6581d89.png">

## Acceptance criteria
* [ ] Create a collection of normalized case timeline elements (on the front-end or back-end)
* [ ] Loop over this collection to determine how to display each element

lomky:
_[what is this chart?](https://github.com/department-of-veterans-affairs/caseflow/issues/11533)_
```bash
1 | 
2 | |||
3 | |||||||
5 | 
8 | 
```

Timebox to a 3.

Why 1?

Why 2?
 - Code cleanup

Why 3?
 - Frontend
 
Why 5?

PR #10149, opened 2019-03-25
Task dates

### LINKED ISSUE #9968: Case timeline tasks out of order
In this screenshot, the BVADispatch tasks are ordered before the Judge and Attorney tasks, even though those were completed first. 

image (13)

## AC
- Case timeline tasks are ordered newest to oldest, top to bottom

## Resources 
Related Case Timeline cleanup ticket: https://github.com/department-of-veterans-affairs/caseflow/issues/9070

youngfreezyVA:
Resolves  #9968

### Description
sort tasks on case timeline by closedAt or createdAt dates

### Acceptance Criteria
Case timeline tasks are ordered newest to oldest, top to bottom

### Testing Plan
1. Go to ...


client/app/queue/SubmitDecisionView.jsx
```diff
@@ -133,7 +133,9 @@ class SubmitDecisionView extends React.PureComponent {
       type: checkoutFlow === DECISION_TYPES.DRAFT_DECISION ?
         'decision' : 'outside medical opinion (OMO) request',
       veteran: veteranFullName,
-      judge: judges[decision.opts.reviewing_judge_id].full_name
+      judge: judges[decision.opts.reviewing_judge_id] ?
```
aroltsch: This commit is from your other PR

spec/feature/queue/case_details_spec.rb
```diff
@@ -694,6 +699,12 @@
         expect(page).to have_content(COPY::CASE_TIMELINE_JUDGE_TASK)
       end
 
+      it "should sort tasks properly" do
+        visit "/queue/appeals/#{appeal.uuid}"
+        first_row_with_date = page.find_all("table#case-timeline-table tbody tr")[1]
+        expect(first_row_with_date).to have_content("01/01/2020")
```
aroltsch: could we check for the rest of the tasks in the test

client/app/queue/utils.js
```diff
@@ -528,3 +528,10 @@ export const nullToFalse = (key, obj) => {
 
   return obj;
 };
+
+export const sortTaskList = (taskList) => {
+  return taskList.sort((prev, next) => {
+    return new Date(next.closedAt || next.createdAt).getTime() -
```
aroltsch: do we only display closed tasks on case timeline? Do we need to fall back to created_at?

youngfreezyVA: based on this line, we display other tasks on case timeline as well:
https://github.com/department-of-veterans-affairs/caseflow/blob/f9f110fce2bbe6c10d81a77963287c7dfe83a7d4/client/app/queue/components/TaskRows.jsx#L240
I'm not sure if there's a better property, other than `createdAt` to fall back on if there is no `closedAt` date. @lowellrex

lowellrex: The `TaskRows` component is used by both the Case Timeline and the Task Snapshot. Case Timeline displays only closed tasks so it should sort by the `closed_at` value. Task Snapshot displays only incomplete tasks so it should sort by the `created_at` value. Because these two different use cases use the same component I think we need to support sorting by both. Ideally we would just pass a pre-sorted list of elements to a component which would then render the elements (there is [a ticket to do just this](https://github.com/department-of-veterans-affairs/caseflow/issues/10004)) so that Case Timeline and Task Snapshot could specify how they want to order things and move that outside of the rendering component. Until we separate the construction of the data and the displaying of that data, I think we will have to support sorting by either value.

spec/feature/queue/case_details_spec.rb
youngfreezyVA: done!

Issue #9649, opened 2019-02-26
Intake | Differentiate Rating vs Nonrating Appeal DTA errors


pkarman:
> We are seeing 47 EP040s with the claim label of "Board DTA Error" which for our workload routing 
> workload is an obsolete claim label.  The current claim label for these claims should either have the 
> claim label of "Board DTA Error Rating" or "Board DTA Error Non-Rating".

https://dsva.slack.com/archives/CFBLH2LR4/p1551187534022500

## Acceptance criteria

* When a DTA error EP 040 is created for a nonrating issue, it should be "Board DTA Error Non-Rating".
* When a DTA error EP 040 is created for a rating issue, it should be "Board DTA Error Rating".

## Technical notes

* May need to identify different EP code for nonrating error. https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/end_product.rb#L25
* The DTA error selection mapping will need updating https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/request_issue.rb#L94

shanear:
It seems like IMO is no longer necessary. This is the full list of 040's per Jane:

## 040 Supplemental EP
Board DTA Error Non-Rating
Board DTA Error Rating
HLR DTA Error Non-Rating
HLR DTA Error Rating
PMC Board DTA Error Non-Rating
PMC Board DTA Error Rating
PMC HLR DTA Error Non-Rating
PMC HLR DTA Error Rating
PMC Supplemental Claim Non-Rating
PMC Supplemental Claim Rating
Supplemental Claim Non-Rating
Supplemental Claim Rating

leikkisa:
Ask Nate for the actual codes

leikkisa:
Here are the codes:
Board DTA Error Non-Rating - 040BDENR
Board DTA Error Rating - 040BDER
HLR DTA Error Non-Rating - 040HDENR
HLR DTA Error Rating - 040HDER
PMC Board DTA Error Non-Rating - 040BDENRPM
PMC Board DTA Error Rating - 040BDERPMC
PMC HLR DTA Error Non-Rating - 040HDENRPMC
PMC HLR DTA Error Rating - 040HDERPMC
PMC Supplemental Claim Non-Rating - 040SCNRPMC
PMC Supplemental Claim Rating - 040SCRPMC
Supplemental Claim Non-Rating - 040SCNR
Supplemental Claim Rating - 040SCR

leikkisa:
[...]

PR #10157, opened 2019-03-26
Fix board dta pension nonrating EP code

leikkisa:
connects #9649 

Bug fix

Fortunately, does not appear to affect any prod data. Still investigating the issue that prompted looking into this.


leikkisa:
Waiting to hear back from VBMS if there is a "C" missing at the end of this code.

leikkisa:
Nate says this is how they're stored in corporate, so I'm going to go ahead without the "C".

Issue #9785, opened 2019-03-05
Hide duplicate organization tasks from Task Snapshot


lowellrex:
Currently we display all active tasks in the task snapshot. This results in us displaying some tasks assigned to organizations that duplicate their child tasks assigned to real people (see example below). We should hide these tasks from the task snapshot when there are no actions available for those tasks. We have [an existing ticket](https://github.com/department-of-veterans-affairs/caseflow/issues/8937) to hide duplicate tasks from the case timeline so we can probably re-use whatever we do there in service of completing this ticket.

<img width="1003" alt="example" src="https://user-images.githubusercontent.com/32683958/53818492-24d16f80-3f36-11e9-9477-c5d47111bc6a.png">

## Acceptance criteria
* [ ] Determine what it means for a task to be a "duplicate"
* [ ] Hide duplicate tasks from the task snapshot

tomas-nava:
SMALL (may be able to reuse work done in #8937)

PR #10159, opened 2019-03-26
hide duplicate org tasks from case timeline and active tasks

### LINKED ISSUE #8937: Hide duplicate organization tasks from case timeline

youngfreezyVA:
Resolves #8937 #9785


app/models/task.rb
```diff
@@ -143,12 +143,16 @@ def hide_from_queue_table_view
     false
   end
 
+  def duplicate_org_task
+    assigned_to.is_a?(Organization) && children.pluck(:assigned_to_type).include?(User.name)
```
aroltsch: @lowellrex could you confirm this is correct

app/models/task.rb
```diff
@@ -143,12 +143,16 @@ def hide_from_queue_table_view
...
+  end
+
   def hide_from_case_timeline
-    false
+    duplicate_org_task
   end
 
   def hide_from_task_snapshot
-    false
+    duplicate_org_task
```
lowellrex: This is such a clever way to solve this problem! Bravo!

app/models/task.rb
lowellrex: I think we also need to confirm that the child task that is assigned to a user has the same type as the current task. Maybe something like:
```ruby
assigned_to.is_a?(Organization) && children.any? { |t| User.name == t.assigned_to_type && type == t.type }
```

app/models/task.rb
youngfreezyVA: thank you!

app/models/task.rb
```diff
@@ -143,12 +143,16 @@ def hide_from_queue_table_view
     false
   end
 
+  def duplicate_org_task
```
aroltsch: Let's add a unit test for this method.

youngfreezyVA: donesies

Issue #10148, opened 2019-03-25
Intake | Duplicate EP exists


pkarman:
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/4205/events/281398/

Claim not established. A duplicate claim for this EP code already exists in CorpDB. Please use a different EP code modifier. 

RAMP intake is failing to identify duplicate modifiers during the intake process, and then failing during establishment.

pkarman:
Ramp Refiling 3948 has `nil` modifier on its EPE.

https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/4158/events/281976/

pkarman:
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/4157/events/282034/

Happens during /intake/:id/complete PATCH request

leikkisa:
It looks like we haven't seen this error in a while for RAMP, and we've made some other changes to this code, so I'm going to close this, and we can re-open if it comes up again.

PR #10167, opened 2019-03-26
Add Caseflow::Error::VBMS class to alias body -> message

pkarman:
connects #10148 

### Description

Our new VBMSError wrapper class was masking some special error handling that was looking for a different error class. The tests did not show this because they explicitly mock a different class than the live external VBMS lib uses.


pkarman:
> Do we need to backfill any data?

Not sure yet. Probably some manual fixing of some kind. See #10148

PR #10168, opened 2019-03-26
Update database comments for Intake tables

leikkisa:
connects #9559


db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
+class AddIntakeDatabaseComments < ActiveRecord::Migration[5.1]
+  def change
+    change_table_comment(:appeals, "Decision reviews intaken for appeals to the board (also known as a notice of disagreement).")
```
shanear: I think this should probably indicate that they are AMA and not legacy appeals.

leikkisa: 👍

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+    change_table_comment(:decision_issues, "Issues that represent a decision made on a decision review's request issue.")
```
shanear: They may not always be on request issues in the future. So I'd say it should just be:

```
Issues that represent a decision made on a decision review
```

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+    change_table_comment(:end_product_establishments, "Represents end products that have been, or need to be established by Caseflow. Used to track the status of those end products as they are processed in VBMS and/or SHARE.")
+
+    change_table_comment(:ramp_refilings, "Intake data for RAMP refilings, also known as RAMP selection.")
+
+    change_table_comment(:request_issues_updates, "Keeps track of edits to request issues on a decision review that happen after the initial intake, such as removing and adding issues.  When the decision review is processed in VBMS, this also tracks whether adding or removing contentions in VBMS for the update has succeeded.")
+
+    change_table_comment(:request_decision_issues, "Bridge table to match request issues to decision issues.")
```
shanear: I think it's more common to call these "join tables":

```
Join table for the has and belongs to many relationship between request issues and decision issues
```

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+    change_table_comment(:ramp_elections, "Intake data for RAMP elections.")
+
+    change_table_comment(:intakes, "Represents the intake of an form or request made by a veteran.")
+
+    change_table_comment(:decision_issues, "Issues that represent a decision made on a request issue.")
```
shanear: this is a dup of line 5

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:decision_issues, :promulgation_date, "The promulgation date of the rating that a decision issue resulted in (if applicable). It is used for calculating whether a decision issue is within the timeliness window to be appealed or get a higher level review.")
+
+     change_column_comment(:decision_issues, :rating_issue_reference_id, "Identifies the specific issue on the rating that resulted from the decision issue (a rating can have multiple issues). This is unique per rating issue.")
+
+     change_column_comment(:appeals, :closest_regional_office, "The code for the regional office closest to the Veteran on the appeal.")
+
+     change_column_comment(:appeals, :docket_type, "The docket type selected by the Veteran on their appeal form, which can be hearing, evidence submission, or direct review.")
+
+     change_column_comment(:appeals, :established_at, "Timestamp for when the appeal has successfully been intaken into Caseflow.")
```
shanear: I think this is wrong. It's the time the establishment of the appeal was succesfully processed. Doesn't necessarily correlate to when it was intook.

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:appeals, :establishment_error, "The error message if attempting to establish the appeal resulted in an error. This gets cleared once the establishment is successful.")
+
+     change_column_comment(:appeals, :establishment_last_submitted_at, "Timestamp for when the the job is eligible to run (can be reset to restart the job).")
+
+     change_column_comment(:appeals, :establishment_processed_at, "Timestamp for when the establishment has succeeded in processing.")
+
+     change_column_comment(:appeals, :establishment_submitted_at, "Timestamp for when the intake was submitted by the user.")
+
+     change_column_comment(:appeals, :legacy_opt_in_approved, "Indicates whether a Veteran opted to withdraw their matching issues from the legacy process when submitting them for an decision review. If there is a matching legacy issue and it is not withdrawn then it is ineligible for the decision review.")
```
shanear: Could shorten this a bit
```
Indicates whether a Veteran opted to withdraw matching issues from the legacy process. If there is a matching legacy issue and it is not withdrawn then it is ineligible for the decision review.
```

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:appeals, :receipt_date, "The date that an appeal form was received by central mail. Used to determine which issues are within the timeliness window to be appealed. Only issues decided prior to the receipt date will show up as contestable issues.")
```
shanear: I think `Receipt date of the appeal form` is good enough. I don't think appeal forms are all received via central mail.

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:appeals, :target_decision_date, "If the appeal docket is direct review, this sets the target decision date for the appeal, which is one year after the receipt date.")
```
shanear: sort of weird that we store this lol, since it's just a straight computed value. Is it _always_ 1 year after the receipt date?

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:appeals, :uuid, "The universally unique identifier for the appeal, which can be used to navigate to appeals/appeal_uuid")
```
shanear: We could point out that the UUID also allows a single id to determine a legacy or AMA appeal, since the format for legacy appeal ids is < 10 digits instead of a 40 character UUID.

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:appeals, :veteran_file_number, "The file number of the Veteran for this review. Not unique per Veteran.")
```
shanear: > Not unique per Veteran.

I don't think this is by design. Might be good to point out that there can sometimes erroneously be more than one per veteran. 

```
The VBA corporate file number of the Veteran for this review. There can sometimes erroneously be more than one file number per veteran.
```

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:appeals, :veteran_is_not_claimant, "Selected by the user during intake, indicates whether the Veteran is the claimant, or if the claimant is someone else such as a dependent. Must be TRUE if Veteran is deceased.")
+
+     change_column_comment(:ramp_refilings, :appeal_docket, "When the RAMP refiling option selected is appeal, they can select hearing, direct review or evidence submission as the appeal docket.")
+
+     change_column_comment(:ramp_refilings, :established_at, "Timestamp for when the review successfully established, including any related actions such as establishing a claim in VBMS if applicable.")
```
shanear: I don't think this is right. Shouldn't this be the same as `established_at` for appeals:

https://github.com/department-of-veterans-affairs/caseflow/pull/10168/files#diff-fb81967b591c7669f1399cae78a751caR53

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:ramp_refilings, :establishment_processed_at, "Timestamp for when the end product establishments for the RAMP review finished processing.")
+
+     change_column_comment(:ramp_refilings, :establishment_submitted_at, "Timestamp for when an intake for a review was submitted by the user.")
+
+     change_column_comment(:ramp_refilings, :has_ineligible_issue, "Selected by the user during intake, indicates whether the Veteran has ineligible issues.")
```
shanear:
```
indicates whether the Veteran listed ineligible issues on their refiling
```

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:ramp_refilings, :option_selected, "Which lane the RAMP refiling is for, between appeal, higher level review, and supplemental claim.")
+
+     change_column_comment(:ramp_refilings, :receipt_date, "The date that the RAMP form was received by central mail.")
```
shanear: again, I don't think this always has to be central mail, and "receipt date" is an ubiquitous enough word.

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:ramp_refilings, :veteran_file_number, "The file number of the Veteran for this review. Not unique per Veteran.")
```
shanear: same as veteran_file_number comments above

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:request_issues_updates, :after_request_issue_ids, "An array of the request issue IDs after a user has finished editing a decision review. Used with before_request_issue_ids to determine appropriate actions (such as which contentions need to be added).")
```
shanear: "An array of the __active__ request issue IDs"

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:request_issues_updates, :attempted_at, "Timestamp for when the request issue update was last attempted.")
```
shanear: maybe this should say "Timestamp for when the request issue update __processing__ was last attempted"

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:request_issues_updates, :before_request_issue_ids, "An array of the request issue IDs previously on the decision review before this editing session. Used with after_request_issue_ids to determine appropriate actions (such as which contentions need to be removed).")
```
shanear: "An array of the __active__ request issue IDs"

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:request_issues_updates, :error, "The error message if the last attempt at updating the request issues was not successful.")
```
shanear: "updating the request issues" => "processing the request issues update"

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:request_issues_updates, :last_submitted_at, "Timestamp for when the the job is eligible to run (can be reset to restart the job).")
```
shanear: "Timestamp for when the processing for the request issues update was last submitted. Used to determine how long to continue retrying the processing job. Can be reset to allow for additional retries"

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:request_issues_updates, :processed_at, "Timestamp for when the request issue updated successfully completed processing.")
```
shanear: "request issue updated" => "request issue update"

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:request_issues_updates, :review_id, "The ID of the decision review edited.")
+
+     change_column_comment(:request_issues_updates, :review_type, "The type of the decision review edited.")
+
+     change_column_comment(:request_issues_updates, :submitted_at, "Timestamp when the edit was originally submitted.")
```
shanear: "edit" => "request issues update"

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:request_issues_updates, :user_id, "The ID of the user who edited the decision review.")
+
+     change_column_comment(:request_decision_issues, :created_at, "Automatic timestamp when row was created.")
+
+     change_column_comment(:request_decision_issues, :decision_issue_id, "The ID of the decision issue.")
+
+     change_column_comment(:request_decision_issues, :request_issue_id, "The ID of the request issue.")
+
+     change_column_comment(:request_decision_issues, :updated_at, "Automatically populated when the record is updated.")
```
shanear: I feel like this column isn't needed lol, since we never update this table. But whateves

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:ramp_elections, :established_at, "Timestamp for when the review successfully established, including any related actions such as establishing a claim in VBMS if applicable.")
+
+     change_column_comment(:ramp_elections, :notice_date, "The date that the Veteran was notified of their option to opt their legacy appeals into RAMP.")
+
+     change_column_comment(:ramp_elections, :option_selected, "Indicates whether the Veteran selected for their RAMP election to be processed as a higher level review (with or without a hearing), a supplemental claim, or a board appeal.")
+
+     change_column_comment(:ramp_elections, :receipt_date, "The date that the RAMP form was received by central mail.")
+
+     change_column_comment(:ramp_elections, :veteran_file_number, "The file number of the Veteran for this review. Not unique per Veteran.")
```
shanear: same comment as other veteran_file_number comments

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:intakes, :cancel_reason, "The reason the intake was canceled. Could have been manually canceled by a user, or automatic.")
+
+     change_column_comment(:intakes, :completed_at, "Timestamp for when the intake was completed, whether it was successful or not.")
+
+     change_column_comment(:intakes, :completion_started_at, "Timestamp for when the user submitted the intake to be completed.")
+
+     change_column_comment(:intakes, :completion_status, "Indicates whether the intake was successful, or was closed by being canceled, expired, or due to an error.")
+
+     change_column_comment(:intakes, :detail_id, "The ID of the decision review or RAMP review that resulted from the intake.")
```
shanear: I think we should probably be a bit more vague here.

"The ID of the record created as a result of the intake"

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:intakes, :detail_type, "The type of decision review or RAMP review that the intake resulted in.")
```
shanear: "The type of the record created as a result of the intake"

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:intakes, :error_code, "If the intake was unsuccessful due to a set of known errors, the error code is stored here. An error is also stored here for RAMP elections that are connected to an active end product, even though the intake is a success.")
+
+     change_column_comment(:intakes, :started_at, "Timestamp for when the intake was created, which happens when a user successfully searches for a Veteran.")
+
+     change_column_comment(:intakes, :type, "The class name of the intake.")
+
+     change_column_comment(:intakes, :user_id, "The ID of the user who created the intake.")
+
+     change_column_comment(:intakes, :veteran_file_number, "The file number of the Veteran which the intake is for. Not unique per Veteran.")
```
shanear: same as other veteran_file_number comments

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:end_product_establishments, :benefit_type_code, "1 if the Veteran is alive, and 2 if the Veteran is deceased. Not to be confused with benefit_type, which is unrelated.")
+
+     change_column_comment(:end_product_establishments, :claim_date, "The claim_date for end products established is set to the receipt date of the form.")
```
shanear: I don't think we want to be so specific, since we may establish end products for all sorts of reasons.

db/migrate/20190326202730_add_intake_database_comments.rb
```diff
@@ -0,0 +1,191 @@
...
+
+     change_column_comment(:end_product_establishments, :source_type, "The type of source that resulted in this end product establishment.")
+
+     change_column_comment(:end_product_establishments, :station, "The station ID of the end product's station.")
+
+     change_column_comment(:end_product_establishments, :synced_status, "The status of the end product, which is synced by a job. Once and end product is cleared (CLR) or canceled (CAN) the status is final and the end product will not continue being synced.")
+
+     change_column_comment(:end_product_establishments, :user_id, "The ID of the user who performed the decision review intake.")
+
+     change_column_comment(:end_product_establishments, :veteran_file_number, "The file number of the Veteran submitted when establishing the end product.")
```
shanear: same as other comments

db/migrate/20190326202730_add_intake_database_comments.rb
leikkisa: Haha okay. I think bridge table might be more common in the DB world, but saying many to many should make it clear enough to all audiences.

db/migrate/20190326202730_add_intake_database_comments.rb
leikkisa: This isn't our column, so I looked at the logic for it and yes it is always the case. I could take it out too and leave it blank since someone else has ownership. I'm leaving it in for now but let me know if I should take it out (the comment, not the column).

db/migrate/20190326202730_add_intake_database_comments.rb
leikkisa: Ok, I changed it to the new comment, but left out "erroneously" since that gets determined upstream and I think it still gets the point across.

db/migrate/20190326202730_add_intake_database_comments.rb
leikkisa: In the ramp_refiling_intake model, this gets set after create_end_product_and_contentions. Not sure if that's what we want to happen, but the logic is actually different than appeal where it gets set before trying to do any processing.

db/migrate/20190326202730_add_intake_database_comments.rb
leikkisa: Agree! I noticed that too.

db/migrate/20190326202730_add_intake_database_comments.rb
leikkisa: Ah yes, that's true. I was trying to capture business logic, but this is also not true for decision documents. Will make it more generic.

PR #10190, opened 2019-03-28
Refactor RAMP Review

leikkisa:
connects #9946

### DIFF
--- a/app/models/ramp_election_intake.rb
+++ b/app/models/ramp_election_intake.rb
@@ -41,12 +41,7 @@ class RampElectionIntake < Intake
     return if complete? || pending?
 
     start_completion!
-
-    if existing_ramp_election_active?
-      use_existing_ramp_election
-    else
-      create_or_connect_end_product
-    end
+    create_or_connect_end_product
 
     transaction do
       complete_with_status!(:success)
@@ -141,29 +136,7 @@ class RampElectionIntake < Intake
   end
 
   def new_intake_ramp_election
-    @new_intake_ramp_election ||= matching_ramp_election_with_notice_date || veteran_ramp_elections.build
-  end
-
-  def existing_ramp_election
-    @existing_ramp_election ||= RampElection.established.where(
-      veteran_file_number: veteran_file_number,
-      option_selected: detail.option_selected
-    ).where.not(id: detail_id).first
-  end
-
-  def existing_ramp_election_active?
-    existing_ramp_election&.end_product_active?
-  end
-
-  def use_existing_ramp_election
-    transaction do
-      detail.destroy!
-      update!(detail: existing_ramp_election)
-    end
-  end
-
-  def matching_ramp_election_with_notice_date
-    veteran_ramp_elections.where(established_at: nil).where.not(notice_date: nil).first
+    @new_intake_ramp_election ||= veteran_ramp_elections.build
   end
 
   def veteran_ramp_elections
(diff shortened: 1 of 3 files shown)


app/models/ramp_election_intake.rb
```diff
@@ -41,12 +41,7 @@ def complete!(_request_params)
     return if complete? || pending?
 
     start_completion!
-
-    if existing_ramp_election_active?
-      use_existing_ramp_election
```
leikkisa: existing_ramp_elections were used when they were pre-made for RAMPS with a notice date, but we're not using them anymore.

app/models/ramp_review.rb
```diff
@@ -62,13 +62,14 @@ def on_decision_issues_sync_processed(end_product_establishment)
...
 
-  # If an EP with the exact same traits has already been created. Use that instead
+  # If an EP with the exact same traits has already been created and is still active. Use that instead
   # of creating a new EP. This prevents duplicate EP errors and allows this method
   # to be idempotent
   #
   # Returns a symbol designating whether the end product was created or connected
   def create_or_connect_end_product!
-    return connect_end_product! if end_product_establishment.preexisting_end_product
+    return :connected if preexisting_end_product_establishment
```
leikkisa: If this specific RAMP review already has an end_product_establishment, I feel that we shouldn't establish it again, which was happening before if there was a preexisting end_product_establishment.

This should not happen since we're not going to be using existing ramp_elections anymore, but thought I'd put this in here as a safety. We could use a different error message for these to differentiate them from ones that are connected to an EPE from a different RAMP Election.

app/models/ramp_review.rb
```diff
@@ -104,14 +107,45 @@ def remove_issues!
...
+      modifier: end_product_modifier,
+      station: "397",
+      benefit_type_code: veteran.benefit_type_code
+    )
+  end
+
+  def connected_end_product_establishment
+    @connected_end_product_establishment ||=
+      matching_end_product_establishments.detect do |epe|
+        epe.result&.last_action_date&.nil? || epe.result.last_action_date > established_at
```
leikkisa: connect to an end product that was active when this ramp review was established. Keep this EPE ephemeral.

leikkisa:
The plan is to:
- Do not create EPEs for connected RAMP elections with an existing EPE, but link to the existing connected EPE. 
- If the EPE that it's connected to is missing, then save a new EPE as needed (no need to backfill these), but do not process it (because it was already processed sometime in the past).
- If there are RAMP Elections that are not connected that are missing their EPE, backfill them.

It appears that the data weirdness was largely caused by RampElectionIntakes that had multiple RampElections. Also if there was a preexisting EPE but not a connected one, it could get "processed" again.

app/models/ramp_election_intake.rb
```diff
@@ -141,29 +136,7 @@ def validate_detail_on_start
...
-  def use_existing_ramp_election
-    transaction do
-      detail.destroy!
-      update!(detail: existing_ramp_election)
-    end
-  end
-
-  def matching_ramp_election_with_notice_date
-    veteran_ramp_elections.where(established_at: nil).where.not(notice_date: nil).first
+    @new_intake_ramp_election ||= veteran_ramp_elections.build
```
leikkisa: I will actually move this to "find_or_build_initial_detail"

leikkisa:
Closing this because the main issues have been resolved in other PRs.

PR #10191, opened 2019-03-28
Replace AMS with fast_jsonapi

monfresh:
**Why**: It's purportedly at least 25x faster than AMS, and the Netflix
team wants to maintain that performance with every change. This library
is also more active than AMS.

The API is very similar to AMS, so it's very much worth it to switch.

https://github.com/Netflix/fast_jsonapi


app/controllers/hearings/hearing_day_controller.rb
```diff
@@ -139,15 +139,6 @@ def record_not_found
...
 
-  def json_created_hearings(hearings)
-    json_hash = ActiveModelSerializers::SerializableResource.new(
-      hearings,
-      each_serializer: ::Hearings::HearingDaySerializer
-    ).as_json
-
-    format_for_client(json_hash)
-  end
-
```
monfresh: This code is not used.

app/controllers/hearings/hearing_day_controller.rb
```diff
@@ -169,18 +160,6 @@ def json_hearing(hearing)
...
-      hearing_array = []
-      json_hash[:data].each do |hearing_hash|
-        hearing_array.push({ id: hearing_hash[:id] }.merge(hearing_hash[:attributes]))
-      end
-      hearing_array
-    else
-      { id: json_hash[:data][:id] }.merge(json_hash[:data][:attributes])
-    end
-  end
-
```
monfresh: This code is not used.

app/models/decision_review.rb
```diff
@@ -293,6 +293,20 @@ def find_or_build_request_issue_from_intake_data(data)
...
+    return descripton if descripton
+
+    description = fetch_status_description_using_claim_type
+    return description if description
+
+    return "1 issue" if request_issues.count == 1
+
+    "#{request_issues.count} issues"
+  end
+
```
monfresh: `fast_jsonapi` requires all serializer attributes to be public methods on the object. I'm surprised AMS allows private methods.

app/controllers/tasks_controller.rb
```diff
@@ -111,18 +111,13 @@ def for_appeal
...
-      data: ActiveModelSerializers::SerializableResource.new(
-        tasks,
-        user: current_user,
-        role: user_role,
-        exclude_extra_fields: true
-      ).as_json[:data]
-    }
+    render json: AmaAndLegacyTaskSerializer.new(
+      tasks: tasks, params: params, ama_serializer: WorkQueue::RegionalOfficeTaskSerializer
+    ).call
```
monfresh: By specifying the `ama_serializer` we can adhere to the [Tell, Don't Ask](https://thoughtbot.com/blog/tell-dont-ask) principle, as opposed to passing in the `exclude_extra_fields` param into the `TaskSerializer`, which then needs to have a bunch of conditionals based on the presence of `exclude_extra_fields`. Instead, we create a new `RegionalOfficeTaskSerializer` that inherits from `TaskSerializer` and overrides certain attributes.

app/models/appeal_series.rb
```diff
@@ -249,7 +249,7 @@ def details_for_status
     when :at_vso
       { vso_name: representative_name }
     when :decision_in_progress
-      { decision_timeliness: DECISION_TIMELINESS.dup }
+      { decisionTimeliness: DECISION_TIMELINESS.dup }
```
pkarman: why camelCase here?

app/models/appeal_series_issues.rb
```diff
@@ -46,9 +46,9 @@ def all
 
         {
           description: issues.first.friendly_description,
-          diagnostic_code: issues.first.diagnostic_code,
+          diagnosticCode: issues.first.diagnostic_code,
```
pkarman: camelCase?

app/models/appeal_series.rb
monfresh: Good question. I meant to add a comment about this. With AMS, you can specify a `key_transform` option that will convert all keys into the specified transform, such as `camel_lower`. It applies the transform to keys at all levels. However, `fast_jsonapi` only transforms keys at the top level. For example, in `V2::AppealStatusSerializer`, there is an `issues` attribute that itself returns a hash based on the appeal's `issues_hash`. `fast_jsonapi` doesn't transform the keys in `issues_hash`.

As far as I can tell, the `issues_hash` is only used by serializers, so I thought why not transform them at the source for now? As a follow-up, I want to move the `issues_hash` method out of the model and into a new serializer that itself specifies the key transform.

PR #10192, opened 2019-03-28
Removing hearing day options from daily docket

sharonwarner:
We no longer allow users to postpone from the daily docket, so we don't need to load other open hearing days for a hearing location. This PR removes that logic.


app/models/hearing_day.rb
```diff
@@ -135,15 +136,20 @@ def upcoming_days_for_vso_user(start_date, end_date, user)
...
 
+      caseflow_hearing_ids = vacols_hearings_for_remaining_hearing_days.values.flatten.pluck(:id)
+
+      loaded_caseflow_hearings = LegacyHearing.includes(appeal: :tasks).where(id: caseflow_hearing_ids)
+
       hearing_days_with_vacols_hearings = remaining_hearing_days.select do |hearing_day|
         !vacols_hearings_for_remaining_hearing_days[hearing_day.id.to_s].nil? &&
           vacols_hearings_for_remaining_hearing_days[hearing_day.id.to_s].any? do |hearing|
-            hearing.assigned_to_vso?(user)
+            loaded_caseflow_hearings.find(hearing.id).assigned_to_vso?(user)
```
mdbenjam: What's this change for?

app/models/hearing_day.rb
```diff
@@ -135,15 +136,20 @@ def upcoming_days_for_vso_user(start_date, end_date, user)
         remaining_hearing_days.pluck(:id)
       )
 
+      caseflow_hearing_ids = vacols_hearings_for_remaining_hearing_days.values.flatten.pluck(:id)
+
+      loaded_caseflow_hearings = LegacyHearing.includes(appeal: :tasks).where(id: caseflow_hearing_ids)
```
mdbenjam: These will always be created already?

sharonwarner: Yes, we create them in hearings_for in fetch_hearings_for_parents.

PR #10193, opened 2019-03-28
Pdf reader large files

### LINKED ISSUE #9330: Problems with very large files in Reader
We’re experiencing a few issues  with 1000+ page documents in reader:

1. *Zooming does not preserve the current page + location.*
This is most noticeable on a very high number page such as 1000 (which will jump to page ~947), but affects all pages.

1. *When zoomed in, pressing “Add a comment” will skip to another page*
Attempting to scroll or move the comment cursor while placing will also skip to different pages and locations. Usually occurs at zoom level 1.6.
Support ticket for this one: https://github.com/department-of-veterans-affairs/appeals-support/issues/3890

1. *When zoomed in, pressing “Add a comment” will cause the pdf view to go blank*
Clicking to place the comment will not trigger a ’Place Annotation” action. Scrolling the document will typically bring the view back, but often to the wrong page and will cause the same issues as above with the view going blank again as well. Usually occurs at zoom level 1.8.
Support ticket for this one: https://github.com/department-of-veterans-affairs/appeals-support/issues/4049

1. And finally, if you are actually able to place a comment on the page you intended….
*Comment sidebar shows incorrect page number and the jump to section button does not skip to the correct page*
Hitting the jump to comment button multiple times will just back and forth between incorrect pages if zoomed in.

Notes:
1. Use https://dsvaappeals.1password.com/vaults/all/allitems/qnqd3hkjl36zh5me5sk2zm3hrq document to re-produce
2. Ask @hschallhorn for clarification.

youngfreezyVA:
Resolves #9330


client/app/reader/PdfFile.jsx
```diff
@@ -213,7 +212,10 @@ export class PdfFile extends React.PureComponent {
       const scrollToIndex = this.props.jumpToPageNumber ? pageIndexOfPageNumber(this.props.jumpToPageNumber) : -1;
 
       this.grid.scrollToCell(this.pageRowAndColumn(scrollToIndex));
-      this.props.resetJumpToPage();
+
+      setTimeout(() => {
```
mdbenjam: Do we still need this? I think this change was unrelated to the fix.

Issue #9734, opened 2019-03-01
Intake Edit | Decision review removal confirmation alert


shanear:
We need to provide confirmation when users remove decision reviews.

## Acceptance Criteria
- When an appeal removal is saved, the user should be redirect back to the case details page with an alert based on the mock below. 
- When the removal of a non-comp decision review is saved, the user should be redirected back to the appropriate business line landing page with the alert from the mock below.
- When the removal of a VBMS based decision review is saved, a confirmation page should be shown based on the mocks below.

## Mocks
Caseflow based:
![confirmation_for_queue](https://user-images.githubusercontent.com/1596259/53656533-685f6d00-3c08-11e9-9403-1e4061d2c94c.png)
Copy should be changed based on the type of decision review.

VBMS based:
![confirmaton_text02](https://user-images.githubusercontent.com/1596259/53657762-e7a27000-3c0b-11e9-81dc-1d6742231ec1.png)

## Notes
- Copy subject to change per @carolaponce 

## Technical Notes
- Ask @lowellrex and/or @laurjpeterson what the case details page for a removed appeal should look like. Will it load?
- There will likely be more variations on copy for withdraws and "stayed" reviews. Keep that in mind in the technical design

pkarman:
See also #9888

pkarman:
Validation requires manually toggling feature flag in UAT: `FeatureToggle.enable!(:remove_decision_reviews)`

pkarman:
Validated non-comp:

<img width="767" alt="Screen Shot 2019-04-09 at 3 01 32 PM" src="https://user-images.githubusercontent.com/1205061/55831565-03d1e080-5ad9-11e9-8a55-57c1d2dcc7c1.png">
<img width="1085" alt="Screen Shot 2019-04-09 at 3 01 50 PM" src="https://user-images.githubusercontent.com/1205061/55831577-0df3df00-5ad9-11e9-8f3b-a78e40e82ba0.png">

Validated VBMS:

<img width="802" alt="Screen Shot 2019-04-09 at 3 05 30 PM" src="https://user-images.githubusercontent.com/1205061/55831602-1cda9180-5ad9-11e9-8ceb-3514d760ca53.png">
<img width="857" alt="Screen Shot 2019-04-09 at 3 05 42 PM" src="https://user-images.githubusercontent.com/1205061/55831613-22d07280-5ad9-11e9-9caf-48211746ba76.png">

PR #10197, opened 2019-03-28
Fixes bug non-comp benefit types route to case detail page

Sjones352:
connects #9734

### Description
This pr fixes the non-comp decisions routing to queue case detail page. Before it wouldn't load the case detail page if the benefit-type was a non-compensation. We discovered this bug while working on the appeal alert message. 

Screen Shot 2019-03-28 at 3 15 22 PM


spec/feature/intake/appeal/edit_spec.rb
```diff
@@ -419,6 +419,34 @@
...
+        description: "Description for Accrued",
+        date: 1.day.ago.to_date.mdY
+      )
+
+      expect(page).to_not have_content("The Veteran's profile has missing or invalid information")
+      expect(page).to have_button("Save", disabled: false)
+
+      click_edit_submit_and_confirm
+      expect(page).to have_current_path("/queue/appeals/#{appeal.uuid}")
+      expect(RequestIssue.find_by(
```
pkarman: maybe we need another expect here to verify the page does not have an error loading all the tasks? Loading the RequestIssue doesn't really help verify that.

app/controllers/tasks_controller.rb
```diff
@@ -96,7 +96,7 @@ def for_appeal
       return json_vso_tasks
     end
 
-    tasks = appeal.tasks
+    tasks = appeal.tasks.reject { |t| t.is_a?(DecisionReviewTask) }
```
pkarman: I'd like to hear from @aroltsch whether this should be in a scope of some kind instead. I expect it will apply to more than just this one endpoint. We can also do that in a follow-on PR.

aroltsch: @Sjones352 I am a bit confused about this change. Could you elaborate please?

Sjones352: @aroltsch 
When an appeal is edited and you click the save button, it is supposed to take you to case detail page but currently it provides an error. The error comes from the prepareTaskForStore method.
Screen Shot 2019-03-28 at 10 41 58 AM

aroltsch: I don't quite understand how this change fixes the error.

pkarman: @aroltsch the problem is that non-comp issues get assigned Tasks that inherit from DecisionReviewTask, which do not serialize the same way that traditional queue tasks do. DecisionReviewTask tasks are meant to be viewed on the `/decision_reviews/:line-of-business` route only. So this change filters them out from the Queue page.

aroltsch: Ah, makes sense! I think adding a scope is a good idea. Sandra, could you also add the above comment to the controller so it is clear for others when looking at this change.

Sjones352: I can add the comment yes and add the scope in the follow-on PR. Thanks all

Sjones352: Done.

PR #10206, opened 2019-03-28
Make disposition update job more resilient to errors

lowellrex:
Connects #9832.

While [testing `HearingDispositionChangeJob ` yesterday](https://dsva.slack.com/archives/C4JECDLSE/p1553721158158400) we encountered a case where a hearing was deleted from VACOLS that prevented the job from completing because [it happened outside of our loop](https://github.com/department-of-veterans-affairs/caseflow/pull/10206/files#diff-404984de894d2c27596f02649e0f1db5L19) that catches individual failures. This PR updates the code so that we don't attempt to access any elements of the hearing related to `DispositionTask`s until we are [inside the loop](https://github.com/department-of-veterans-affairs/caseflow/pull/10206/files#diff-93ee34179e4b1242f3647802a67d582bR22) where we catch individual failures and keep chugging through the job.


spec/jobs/hearing_disposition_change_job_spec.rb
```diff
@@ -25,6 +25,20 @@ def create_disposition_task_ancestry(disposition: nil, scheduled_for: nil, assoc
...
+    root_task = FactoryBot.create(:root_task, appeal: appeal)
+    distribution_task = FactoryBot.create(:distribution_task, appeal: appeal, parent: root_task)
+    parent_hearing_task = FactoryBot.create(:hearing_task, appeal: appeal, parent: distribution_task)
+
+    hearing = FactoryBot.create(
+      :legacy_hearing,
+      appeal: appeal
+    )
+    HearingTaskAssociation.create!(hearing: hearing, hearing_task: parent_hearing_task)
+    DispositionTask.create!(appeal: appeal, parent: parent_hearing_task, assigned_to: Bva.singleton)
```
tomas-nava: (optional) Why not use a factory for these two?

PR #10218, opened 2019-03-29
Add outcode for legacy appeals

### LINKED ISSUE #8278: Implement automatic decision document upload for legacy appeals
1. Utilize the same IDT outcode endpoint as for ama appeals
2. Do not create any EPs or complete BvaDispatchTask tasks for legacy appeals

This endpoint should only create a DecisionDocument record and upload it to VBMS for legacy appeals

nemodjuric:
Resolves #8278 

This PR adds appeal_type to decision_documents table (making it polymorphic)
Notified #appeals-schema


app/models/tasks/bva_dispatch_task.rb
```diff
@@ -12,20 +12,20 @@ def create_from_root_task(root_task)
 
     def outcode(appeal, params, user)
       tasks = where(appeal: appeal, assigned_to: user)
-      if tasks.count != 1
-        fail Caseflow::Error::BvaDispatchTaskCountMismatch, appeal_id: appeal.id, user_id: user.id, tasks: tasks
-      end
 
-      task = tasks[0]
+      throw_error_if_no_tasks_or_if_task_is_completed(appeal, tasks, user)
```
aroltsch: This should only be done for ama appeals. There are no BvaDispatchTask tasks for legacy appeals yet

spec/factories/task.rb
```diff
@@ -43,6 +43,12 @@
       assigned_to { Bva.singleton }
     end
 
+    factory :root_task_legacy, class: RootTask do
```
aroltsch: This doesn't exist yet. There are no root tasks for legacy appeals.

app/models/legacy_appeal.rb
```diff
@@ -20,6 +20,8 @@ class LegacyAppeal < ApplicationRecord
   has_many :appeal_views, as: :appeal
   has_many :claims_folder_searches, as: :appeal
   has_many :tasks, as: :appeal
+  has_many :decision_documents, as: :appeal
+  has_many :decision_issues, as: :decision_review
```
aroltsch: Legacy appeal doesn't have decision_issues

app/models/decision_document.rb
```diff
@@ -133,4 +133,12 @@ def cache_file!
 
     S3Service.store_file(s3_location, Base64.decode64(@file))
   end
+
+  def legacy?
```
aroltsch: Is this method used?

app/models/tasks/bva_dispatch_task.rb
```diff
@@ -11,21 +11,21 @@ def create_from_root_task(root_task)
...
       create_decision_document!(params)
 
-      task.update!(status: Constants.TASK_STATUSES.completed)
-      task.root_task.update!(status: Constants.TASK_STATUSES.completed)
-      appeal.request_issues.each(&:close_decided_issue!)
+      if ama?(appeal)
+        task.update!(status: Constants.TASK_STATUSES.completed)
+        task.root_task.update!(status: Constants.TASK_STATUSES.completed)
+        appeal.request_issues.each(&:close_decided_issue!)
+      end
```
aroltsch: Could we move `create_decision_document!(params)` all the way down so we can have only one `if` statement? Would that break things?

app/models/tasks/bva_dispatch_task.rb
```diff
@@ -34,6 +34,10 @@ def outcode(appeal, params, user)
 
     private
 
+    def ama?(appeal)
```
aroltsch: This method looks a little off because you are passing an appeal here. I would just reuse this line `appeal.class.name == Appeal.name` instead of creating a method

app/models/tasks/bva_dispatch_task.rb
nemodjuric: It would if create_decision_document! fails/throws an error, which can include a scenario where document is not uploaded to VBMS, but we closed the task and request issue.

app/models/decision_document.rb
```diff
@@ -133,4 +133,8 @@ def cache_file!
 
     S3Service.store_file(s3_location, Base64.decode64(@file))
   end
+
+  def ama?
+    appeal.class.name == Appeal.name
```
pkarman: how about `appeal.is_a?(Appeal)` instead?

app/models/tasks/bva_dispatch_task.rb
```diff
@@ -11,21 +11,21 @@ def create_from_root_task(root_task)
     end
 
     def outcode(appeal, params, user)
-      tasks = where(appeal: appeal, assigned_to: user)
-      if tasks.count != 1
-        fail Caseflow::Error::BvaDispatchTaskCountMismatch, appeal_id: appeal.id, user_id: user.id, tasks: tasks
+      if appeal.class.name == Appeal.name
```
pkarman: `appeal.is_a?(Appeal)` is more idiomatic imo, and allows for subclassing.

PR #10222, opened 2019-03-29
rename specific error classes to avoid rails reload error

pkarman:
## Description

Fixes local development `reload!` where we got a TypeError with Rails confused about superclass.


leikkisa:
Weird!

PR #10224, opened 2019-03-29
Lazily load rep and appellant info

### LINKED ISSUE #9774: Performance | Daily docket load time
The daily docket queries need to be optimized.

mdbenjam:
Resolves #9774

### Description
Lazily load the rep and appellant info to avoid calls to BGS on the initial data load on the daily docket pages.

Future improvement:
We could add an error state to display when a request fails. Currently it will just display "Loading..." forever.


app/controllers/hearings/hearing_day_controller.rb
```diff
@@ -42,7 +42,8 @@ def show
 
     render json: {
       hearing_day: json_hearing(hearing_day_hash).merge(
-        hearings: hearings.map { |hearing| hearing.to_hash(current_user.id) }
+        hearings: hearings.map { |hearing| hearing.quick_to_hash(current_user.id) },
+        hearing_day_options: hearing_day_options
```
sharonwarner: I think this line is just a bad merge, we don't need hearing_day_options on the daily docket anymore.

mdbenjam: whoooops great catch. Sorry I resolved that conflict tooo quickly.

app/models/hearing.rb
```diff
@@ -101,6 +101,10 @@ def disposition_editable
     disposition_task_in_progress || !hearing_task?
   end
 
+  def representative
+    appeal.representative_name
```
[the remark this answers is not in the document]
mdbenjam: Screen Shot 2019-04-01 at 12 59 38 PM

PR #10232, opened 2019-04-01
Appeal Status adding check for nil

### LINKED ISSUE #10237: Appeal Status API Sentry issue
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3882/

mc-usds:
Resolves #10237

### Description
Checking for nil so that user does not encounter error.

### Acceptance Criteria
- [x] Code compiles correctly


app/models/appeal.rb
```diff
@@ -642,7 +642,7 @@ def other_close_event_date
     return if active_status?
     return if decision_issues.any?
 
-    root_task.closed_at.to_date
+    root_task.closed_at.try(&:to_date)
```
tomas-nava: You can use [a safe navigation operator](https://rubyinrails.com/2017/11/17/safe-navigation-operator-ampersand-dot-in-ruby/) for this: `root_task.closed_at&.to_date`

mc-usds: Fixed. Thanks!

PR #10241, opened 2019-04-01
Appeal removal is saved show alert message

Sjones352:
Connects #9734
Screen Shot 2019-04-01 at 7 46 10 PM


client/app/queue/QueueApp.jsx
```diff
@@ -63,6 +63,7 @@ import COPY from '../../COPY.json';
 import TASK_ACTIONS from '../../constants/TASK_ACTIONS.json';
 import USER_ROLE_TYPES from '../../constants/USER_ROLE_TYPES.json';
 import DECISION_TYPES from '../../constants/APPEAL_DECISION_TYPES.json';
+import { FlashAlerts } from '../nonComp/components/Alerts';
```
leikkisa: Part of me was wondering if we should make this a more general component (instead of a nonComp component, now that we're using it elsewhere), but there are a couple of other "refactors" that could happen there too and I think it'd be okay to put that into a tech backlog ticket.

spec/feature/intake/appeal/edit_spec.rb
```diff
@@ -533,18 +533,30 @@
...
+            click_remove_issue_confirmation
+          end
+
+          click_edit_submit
+          expect(page).to have_content("Remove review?")
+          expect(page).to have_content("This review and all tasks associated with it will be removed.")
+          click_intake_confirm
+
+          expect(page).to have_current_path("/queue/appeals/#{appeal.uuid}")
+          expect(RequestIssue.find_by(
```
leikkisa: Could you add a check that the Request Issue is removed?  Otherwise I'm not sure what the objective of this expect statement is.

Sjones352: Going to remove it because it checks for the request issues on the Case-detail page.

client/app/queue/QueueApp.jsx
Sjones352: I agree but there would be a lot of refactoring which I believe we can do later.

PR #10244, opened 2019-04-01
Removing asyncable from record_synced_by_job

mdbenjam:
Remove Asyncable from record_synced_by_job. We inherited from Asyncable to try and use the same patterns as other jobs. However, it was overkill for this job since it already has a natural retry mechanism. If the job fails, it will retry an hour later. All we really need is a `processed_at` time.

### DIFF
--- a/db/migrate/20190401192800_remove_columns_from_record_synced_by_job.rb
+++ b/db/migrate/20190401192800_remove_columns_from_record_synced_by_job.rb
@@ -1,0 +1,9 @@
+class RemoveColumnsFromRecordSyncedByJob < ActiveRecord::Migration[5.1]
+  def change
+    safety_assured do
+      remove_column :record_synced_by_jobs, :submitted_at
+      remove_column :record_synced_by_jobs, :attempted_at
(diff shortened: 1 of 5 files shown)


db/migrate/20190401192800_remove_columns_from_record_synced_by_job.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveColumnsFromRecordSyncedByJob < ActiveRecord::Migration[5.1]
+  def change
+    safety_assured do
+      remove_column :record_synced_by_jobs, :submitted_at
+      remove_column :record_synced_by_jobs, :attempted_at
+      remove_column :record_synced_by_jobs, :last_submitted_at
```
pkarman: was the `error` column ever present?

mdbenjam: I was thinking of leaving it even if we aren't using it right now. Maybe I should just remove it as well. We currently have no error handling in this job, so I think the next step is to add it and set the error field.

PR #10247, opened 2019-04-01
Dismiss dropdowns when clicking outside of the menu

### LINKED ISSUE #8972: Click outside of queue selector to dismiss dropdown
Currently clicking outside of the queue selector dropdown menu does not dismiss the menu. This ticket exists to make it so clicking outside of queue selector dropdown menu _does_ dismiss the menu. Using [the existing `DropdownMenu` component](https://github.com/department-of-veterans-affairs/caseflow-frontend-toolkit/blob/master/components/DropdownMenu.jsx) should give us this functionality for free.

queue-switcher

## Acceptance criteria
* [ ] Clicking outside of the queue selector dropdown dismisses the menu

hschallhorn:
Resolves #8972

### Description
Added ability to click outside of the dropdown component to dismiss it. Modified the QueueSelectorDropdown to use a component we already have rather then a custom one.

### Acceptance Criteria
- [x] Dropdown in queue can be dismissed by clicking anywhere but the dropdown itself

### Testing Plan
1. Go to a Judge's queue such a BVAEBECKER's
2. Click the "Select view" dropdown
3. Click outside of the dropdown
4. Notice the menu closes

Untitled

### DIFF
--- a/client/app/components/DropdownButton.jsx
+++ b/client/app/components/DropdownButton.jsx
@@ -21,6 +21,21 @@ export default class DropdownButton extends React.Component {
     this.state = {
       menu: false
     };
+    this.wrapperRef = null;
+  }
+
+  componentDidMount = () => document.addEventListener('mousedown', this.onClickOutside);
+
+  componentWillUnmount = () => document.removeEventListener('mousedown', this.onClickOutside);
+
+  setWrapperRef = (node) => this.wrapperRef = node
+
+  onClickOutside = (event) => {
+    if (this.wrapperRef && !this.wrapperRef.contains(event.target) && this.state.menu) {
+      this.setState({
+        menu: false
+      });
+    }
   }
 
   onMenuClick = () => {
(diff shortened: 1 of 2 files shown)


client/app/components/DropdownButton.jsx
```diff
@@ -21,6 +21,21 @@ export default class DropdownButton extends React.Component {
     this.state = {
       menu: false
     };
+    this.wrapperRef = null;
+  }
+
+  componentDidMount = () => document.addEventListener('mousedown', this.onClickOutside);
```
youngfreezyVA: I would add a check for the document being defined as well `document && document.addEventListener('mousedown', this.onClickOutside);`

youngfreezyVA: https://stackoverflow.com/questions/29782884/document-is-not-defined-when-attempting-to-setstate-from-the-return-of-an-async

hschallhorn: Mind explaining why this is necessary? My understanding of `componentDidMount()` is that it occurs after the first `render()`, ensuring that `document` is defined.

youngfreezyVA: it's not necessary - I approved the PR. if we ever move to server side rendering its a defensive check. have run into this same issue before when server side rendering components... check out the stack overflow link (I updated it).

hschallhorn: Makes sense, thanks!

Issue #10163, opened 2019-03-26
Intake | Set task closed_at when cancelling


pkarman:
Follow on to #9730 

When we cancel a task we should set the `closed_at` timestamp. The Task `set_timestamps` hook does not appear be getting triggered.

Example in UAT is DecisionReviewTask id 1342

pkarman:
Validated as part of #9734 

```
irb(main):007:0> pp Task.last
[2019-04-09 16:14:25 -0400]   Task Load (0.8ms)  SELECT  "tasks".* FROM "tasks" ORDER BY "tasks"."id" DESC LIMIT $1  [["LIMIT", 1]]
#<DecisionReviewTask:0x000055f55eea35e0
 id: 1401,
 appeal_id: 1378,
 status: "cancelled",
 type: "DecisionReviewTask",
 action: nil,
 instructions: [],
 assigned_to_id: 10,
 assigned_by_id: nil,
 assigned_at: Tue, 09 Apr 2019 19:58:38 UTC +00:00,
 started_at: nil,
 created_at: Tue, 09 Apr 2019 19:58:38 UTC +00:00,
 updated_at: Tue, 09 Apr 2019 19:58:38 UTC +00:00,
 appeal_type: "HigherLevelReview",
 placed_on_hold_at: nil,
 on_hold_duration: nil,
 assigned_to_type: "Organization",
 parent_id: nil,
 closed_at: Tue, 09 Apr 2019 20:01:41 UTC +00:00>
```

PR #10249, opened 2019-04-02
Intake | Set task closed_at when cancelling

Sjones352:
Connects #10163
Screen Shot 2019-04-02 at 7 21 08 PM

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -265,7 +265,7 @@ class Task < ApplicationRecord
   def cancel_task_and_child_subtasks
     # Cancel all descendants at the same time to avoid after_update hooks marking some tasks as completed.
     descendant_ids = descendants.pluck(:id)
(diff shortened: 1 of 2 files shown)


app/models/task.rb
```diff
@@ -268,11 +268,13 @@ def previous_task
     nil
   end
 
+  # rubocop:disable Metrics/LineLength
   def cancel_task_and_child_subtasks
     # Cancel all descendants at the same time to avoid after_update hooks marking some tasks as completed.
     descendant_ids = descendants.pluck(:id)
-    Task.active.where(id: descendant_ids).update_all(status: Constants.TASK_STATUSES.cancelled)
+    Task.active.where(id: descendant_ids).update_all(status: Constants.TASK_STATUSES.cancelled, closed_at: Time.zone.now)
```
leikkisa: You can split this into multiple lines to avoid disabling rubocop:
```
update_all(
  status: Constants.TASK_STATUSES.cancelled,
  closed_at: Time.zone.now
)
```

PR #10256, opened 2019-04-02
Fixing VSO method when there is non POA

### LINKED ISSUE #10234: Bug: Appeal charged to POA that isn't listed nor IHP writing
This appeal shows assigned to "National Law School Veterans Clinic Consortium," (a POA VSO) but it is believed that this an error:  

https://appeals.cf.ds.va.gov/queue/appeals/3191420 

From Sal 4/1/19: The hearing was successfully cancelled, no change in location in VACOLS but is now assigned to the National Law School Veterans Clinic consortium?  There is no POA for this appeal in VACOLS or VBMS.

Asked Dustin to look and he believes it is in error. He checked file to see if there's any documents for POA and there isn't. Doesn't seem to be any reason for it to be assigned to that particular

mdbenjam:
Resolves #10234

### Description
When a Veteran has no POA, the participant_id is nil. In this case we look up any VSOs in our database with a nil POA. For some reason we've recently added a few. Looks like some field VSOs may not have a participant id. Regardless we don't want to pull in any VSO if the Veteran doesn't have representation.

### DIFF
--- a/app/models/appeal.rb
+++ b/app/models/appeal.rb
@@ -323,7 +323,7 @@ class Appeal < DecisionReview
   end
 
   def vsos
-    vso_participant_ids = power_of_attorneys.map(&:participant_id)
+    vso_participant_ids = power_of_attorneys.map(&:participant_id) - [nil]
     Vso.where(participant_id: vso_participant_ids)
   end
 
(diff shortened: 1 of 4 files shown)


lowellrex:
Do we need to do any data cleanup after this gets merged? Do we need to delete errantly created `TrackVeteranTask`s?

mdbenjam:
I think... it will fix itself? Won't we auto close the VSOs we no longer return?

lowellrex:
> I think... it will fix itself? Won't we auto close the VSOs we no longer return?

Yup! You're absolutely correct! What a nice little system we built!

PR #10258, opened 2019-04-02
Check BGS access before loading case details page

### LINKED ISSUE #7922: Prevent Caseflow users from viewing case details pages with sensitivity levels above their allowance
@mdbenjam [raises an important consequence](https://github.com/department-of-veterans-affairs/caseflow/pull/7921#pullrequestreview-176555333) of us [moving to non-blocking BGS requests](https://github.com/department-of-veterans-affairs/caseflow/pull/7061) on the case details page: It is possible that VSO employees could view the basic pieces of the case details page of a case where their VSO does not represent the appellant.

One way we could introduce this constraint would be to add a block in `AppealsController.show()` that reaches out BGS if the current user is a VSO employee.

## Acceptance criteria
* VSO employees cannot see case details pages for cases where their VSO does not represent the appellant.

lowellrex:
Resolves #7922. Makes a request to BGS to confirm the current user can access the appeal on the case details page (for sensitivity reasons or POA reasons).

Currently just displays the failure to load screen. We should display a more precise error message in the future, but I think this is a fine solution for the time being. Open to discussion though!

### DIFF
--- a/app/controllers/appeals_controller.rb
+++ b/app/controllers/appeals_controller.rb
@@ -3,6 +3,7 @@
 class AppealsController < ApplicationController
   before_action :react_routed
   before_action :set_application, only: [:document_count]
+  before_action :verify_bgs_sensitivity, only: [:show]
   # Only whitelist endpoints VSOs should have access to.
   skip_before_action :deny_vso_access, only: [:index, :power_of_attorney, :show_case_list, :show, :veteran, :hearings]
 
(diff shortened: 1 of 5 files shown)


app/controllers/appeals_controller.rb
```diff
@@ -153,6 +154,10 @@ def set_application
     RequestStore.store[:application] = "queue"
   end
 
+  def verify_bgs_sensitivity
+    fail(ActionForbiddenError) unless BGSService.new.can_access?(appeal.veteran_file_number)
```
mdbenjam: This is fine, but remember that `can_access?` really just gets veteran information. So it might be cool to remove the later async request to get basic veteran information and return it in the first request since we have to make that call anyway.

lowellrex: I'll make a ticket.

lowellrex: Ticketed here: https://github.com/department-of-veterans-affairs/caseflow/issues/10261

Issue #8095, opened 2018-12-04
Improperly updating tasks after checkout flows


mdbenjam:
Currently when completing a checkout flow we don't always properly update the tasks. i.e. if you do an attorney checkout flow the completed task doesn't end up under the completed tab. They do show up after reloading the page. We need to reload tasks after completing a checkout flow.

joeyyang:
Just want to chime in to say that this behavior also exists when adding an admin action to a task — i.e., the tab contents and numbers do not update when a task has been sent away for an admin action.

lowellrex:
Screencap of the on hold count decreasing (legacy appeals disappear) after an attorney creates an admin action for an AMA appeal:

![onholdcountmismatch](https://user-images.githubusercontent.com/32683958/49875410-ef156280-fdee-11e8-8ee9-e6f6d7674790.gif)

lowellrex:
One problem we are able to reproduce and pinpoint is the issue of tasks related to cases with new documents disappearing from the "on hold" tab. It appears as if the [`onHoldTasksByAssigneeCssIdSelector` selector](https://github.com/department-of-veterans-affairs/caseflow/blob/9079cd44ccfb30ec2af56627261b5862f627a345/client/app/queue/selectors.js#L207) is causing this because it is filtering out tasks related to cases with new documents.

### Steps to reproduce:
* Log in as `BVASCASPER1`
* Make sure there is a task related to a case with new documents (can probably spoof this by hard coding some appeal IDs locally [in `AppealsController.new_documents`](https://github.com/department-of-veterans-affairs/caseflow/blob/9079cd44ccfb30ec2af56627261b5862f627a345/app/controllers/appeals_controller.rb#L28)
* Load `/queue` (see a flicker in the count of tasks that are on hold, from larger number to smaller number)
* Comment out the request to [hit the new documents endpoint in `QueueLoadingScreen`](https://github.com/department-of-veterans-affairs/caseflow/blob/9079cd44ccfb30ec2af56627261b5862f627a345/client/app/queue/QueueLoadingScreen.jsx#L68)
* Load `/queue` (see only the larger number in the on hold tab count, no flicker)
* Click on the "on hold" tab.
  - This should fire requests to the new documents endpoint and the on hold count will decrease at this point in time.

lowellrex:
[...]

PR #10263, opened 2019-04-02
Display the correct assigned tasks and on hold tasks after creating an admin action

hschallhorn:
References #8095

### Description
After an attorney or judge create an admin action, update the tasks in their assigned tab to correctly reflect their tasks

### Acceptance Criteria (without reloading a user's queue)
- [x] When a judge creates an admin action for a legacy appeal, the task no longer appears in their assigned tab
- [ ] When an attorney creates an admin action for a legacy appeal, the task no longer appears in their assigned tab
- [x] When an attorney creates an admin action for a legacy appeal, the task appears in their on hold tab
- [x] When an attorney creates an admin action for an ama appeal when there are legacy tasks on hold, the legacy tasks do not disappear from the on hold tab.

### Testing Plan
**Judge assigns legacy task admin action**
1. Go to BVAAABSHIRE's queue
2. Note the number of assigned tasks
3. Go to the case detains of any legacy task
4. Add an admin action to the task
5. Confirm that the task no longer appears in the Assigned tab

**Attorney assigns legacy task admin action**
1. Go to BVASCASPER1's queue
2. Note the number of assigned and on hold tasks
3. Go to the case detains of any legacy task
4. Add an admin action to the task
5. Confirm that the task no longer appears in the Assigned tab
6. Confirm that the task now appears in the On Hold Tab

**Attorney assigns ama task admin action**
1. Go to BVASCASPER1's queue
2. Note the number of assigned and on hold tasks
3. Go to the case detains of any ama task
4. Add an admin action to the task
5. Confirm that the number of on hold tasks increases rather than decreases


client/app/queue/AddColocatedTaskView.jsx
```diff
@@ -106,12 +106,17 @@ class AddColocatedTaskView extends React.PureComponent {
...
+          amaTasks: allTasks.amaTasks
+        });
+
         if (task.isLegacy) {
           this.props.setAppealAttrs(task.externalAppealId, { location: 'CASEFLOW' });
-        } else {
-          const response = JSON.parse(resp.text);
-
-          this.props.onReceiveAmaTasks(response.tasks.data);
+          this.props.deleteTask(task.uniqueId);
```
hschallhorn: Manually removes the legacy task from our store on the front end. Is this the best way to handle this?

youngfreezyVA: I think the backend should handle this. not sure how atm, but we can look into it together if you like.

lowellrex: I haven't worked in this code for a minute but I tend to agree with @youngfreezyVA here. One idea is that since the back-end returns the entire set of tasks a user should have in their queue to the front-end whenever we [GET](https://github.com/department-of-veterans-affairs/caseflow/blob/5ec779179c47cd689b8b0b7f5998b4091ddd5f11/app/controllers/tasks_controller.rb#L32), [POST](https://github.com/department-of-veterans-affairs/caseflow/blob/5ec779179c47cd689b8b0b7f5998b4091ddd5f11/app/controllers/tasks_controller.rb#L63), or [PATCH ](https://github.com/department-of-veterans-affairs/caseflow/blob/5ec779179c47cd689b8b0b7f5998b4091ddd5f11/app/controllers/tasks_controller.rb#L85)to `/tasks` we could change `onReceiveTasks` and `onReceiveAmaTasks` to `$set` those elements of the global redux store instead of merging those changes in [as it does now](https://github.com/department-of-veterans-affairs/caseflow/blob/5ec779179c47cd689b8b0b7f5998b4091ddd5f11/client/app/queue/reducers.js#L58).

hschallhorn: >entire set of tasks
The legacy tasks assigned to the user are not returned with these requests. These are grabbed from [a separate endpoint](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/QueueLoadingScreen.jsx#L39) when the page is loaded, which is why the queue is correct after a reload.

hschallhorn: Also, only the GET endpoint correctly returns all AMA tasks as [queue_class](https://github.com/department-of-veterans-affairs/caseflow/blob/5ec779179c47cd689b8b0b7f5998b4091ddd5f11/app/controllers/tasks_controller.rb#L155) is determined by the `role` param which only passed with GET requests. With POST and PATCH, GenericQueue is used to determine the rest of the user's AMA tasks

Issue #10070, opened 2019-03-20
Intake | No decision date error on edit


pkarman:
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/4141/events/278262/

https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/4415/events/291850/

Unable to identify a decision date for a rating issue.

This is a Supplemental Claim with one rating request issue. The rating issue has a profile date.

When fetching the ratings there is a "Locked Rating Detected" error, which seems related to #9817 .

We need a better error message when zero ratings are available on editing an existing Decision Review that has rating request issues.

### Acceptance criteria

* All unlocked ratings should be returned via the `Rating` model.
* Store the rating promulgation date on the `decision_date` column during initial Intake
* If an existing rating request issue refers to a locked rating (not in the returned array from `Rating`) then present the user with an error that "one or more locked ratings may prevent this claim from being viewed" instead of letting the page throw a JS exception about `no decision date`.

pkarman:
another case https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/4209/

pkarman:
> Store the rating promulgation date on the `decision_date` column during initial Intake

handled in #10274

pkarman:
> present the user with an error that "one or more locked ratings may prevent this claim from being viewed"

in #10351 and #10380

pkarman:
Validation steps:

* identify a locked rating in UAT
* attempt to edit an existing Claim for the affected veteran
* should receive an error message asking you to try again in 24 hours.

Sjones352:
@pkarman I used this file that has a locked rating to validate `984562385` but I never get the error message. Wondering if I am missing something here

leikkisa:
@Sjones352 I think after re-reading this, we need the rating that is now locked to be one of the ratings that a user added to an intake (so at the time it wasn't locked), and then lock the rating, and then try to edit it.

I think we'll need to enlist Colin's help on this one. Let's first try to find a Decision Review that has a rating issue on it, and then ask Colin if he can "Lock" that specific rating, then try to edit the Decision Review again.  This is when I suspect we should see that error message.

Sjones352:
I have created an Ep on this file`700032062` and added a rating. Asking Colin to lock it so that I can try it again

[...]

PR #10274, opened 2019-04-02
Save rating issue decision date on request_issue

pkarman:
connects #10070 

## Description

Save the original rating issue decision date on the request issue so we do not rely on the presence of the rating issue in the future to show the decision date.


client/app/intake/util/issues.js
```diff
@@ -202,6 +202,7 @@ const formatRatingRequestIssues = (state) => {
       return {
         request_issue_id: issue.id,
         rating_issue_reference_id: issue.ratingIssueReferenceId,
+        rating_issue_decision_date: issue.decisionDate,
```
leikkisa: Does this need the same formatting as the nonrating issue?
formatDateStringForApi

leikkisa: Should we just call this decision_date here also?

app/models/request_issue.rb
```diff
@@ -297,6 +298,9 @@ def approx_decision_date_of_issue_being_contested
       decision_date
     elsif decision_issues.any?
       decision_issues.first.approx_decision_date
```
leikkisa: This isn't about your PR but the name of this method seems ambiguous, given that in the top two cases it results in the decision_date of the contested issue, and in the last situation it results in the decision_date of the request issue.

pkarman: I think 99% of the time it's going to be one of the first two cases.

The only way it gets to the 3rd case is if the request issue somehow represents the contention of a decision w/o a rating. I.e. the first link in the chain.

pkarman: we could try removing it and seeing what tests break...

pkarman: nothing broke so I removed that last condition.

PR #10284, opened 2019-04-03
Add common component for AMA issues

### LINKED ISSUE #8591: Create AMA issues component
### Context
We'd like to componentize the way that AMA issues appear in Caseflow. Other teams can leverage the AMA issues component so that AMA issue display is standardized across Caseflow. To start, the Hearings team would like to use this component in the AMA Hearing Prep tool.

### Current state of AMA issues
screen shot 2019-01-09 at 2 59 11 pm
screen shot 2019-01-09 at 2 59 01 pm

### Proposed mock
screen shot 2019-01-09 at 2 44 14 pm

### Specs
screen shot 2019-01-09 at 3 03 33 pm
Size 17 font. Labels are bold. "Note from NOD" is in italics.

annekainicUSDS:
Resolves #8591

### Description
Create a shareable component for AMA issues. To be used in Case Details page and Add Decisions flow.

The way I decided to build this is by separating the display of issues and decisions that go with each issue. This choice to separate them has some pros and cons:

**Pros**: Decisions don't seem to be shown in every situation where issues are, so keeping them separate makes it easier to only use the component for issues when that's all that's needed. It also makes the issue component more flexible to show different components as children instead of just the one view of decisions.
**Cons**: It means a bit more manual configuration when using both together.

I think the trade-off to favor flexibility over reducing configuration is worthwhile, especially given the brief of this ticket, which is to create a shareable component between all apps that display issues, but open to thoughts here!

How this will now look in the Case Details page:
Screen Shot 2019-04-04 at 2 39 01 PM

How this will now look in the Add Decisions flow:
Screen Shot 2019-04-04 at 2 38 49 PM

Screen Shot 2019-04-04 at 2 39 26 PM

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...


client/app/queue/SelectDispositionsView.jsx
```diff
@@ -256,12 +256,21 @@ class SelectDispositionsView extends React.PureComponent {
...
     } = this.state;
-
     const connectedRequestIssues = appeal.issues.filter((issue) => {
       return decisionIssue && decisionIssue.request_issue_ids.includes(issue.id);
     });
     const connectedIssues = this.connectedRequestIssuesWithoutCurrentId(connectedRequestIssues, requestIdToDelete);
     const toDeleteHasConnectedIssue = connectedIssues.length > 0;
+    const issueErrors = {};
+
+    appeal.issues.forEach((issue) => {
```
youngfreezyVA: I'm not a fan of using forEach since it mutates values instead of creating a new copy of an array. can we modify to use `.map` instead?

youngfreezyVA:
> Code Climate has analyzed commit [74e9b4b](https://github.com/department-of-veterans-affairs/caseflow/commit/74e9b4b56a77af8cc83bf8a56efc5eb3b1be8f55) and detected **2 issues** on this pull request.
> 
> Here's the issue category breakdown:
> 
> Category	Count
> Duplication	2
> View more on [Code Climate](https://codeclimate.com/github/department-of-veterans-affairs/caseflow/pull/10284).

copy and pasting my response on slack: 
re: the duplicated code... if they are exactly identical, I would suggest adding that function as a static method on one of the classes for now and call that static method from the other class @annekainic. instead of referencing `this.props` you can just pass in `this.props` as a parameter to that function. or, you could create a base class with just that method. (

client/app/queue/SelectDispositionsView.jsx
annekainicUSDS: Totally fair, this code is confusing! I can add some comments to help clarify what it is doing. At a high level what this code is doing is checking some conditional logic to determine whether there's an error for each issue in the appeal. If there is an error, that issue is added to the `issueErrors` object (that I declared above), which is then passed down to the child component to determine whether or not to show an error message for each issue.

I didn't think `.map` was appropriate because `map` returns a new array, which I don't need here because I'm creating an object instead. I don't think `forEach` is mutating the original object on which it is called here; all it is doing is calling a function on each issue in `appeal.issues`, which is then checking if an error should be added to the `issueErrors` objects for that issue.

client/app/queue/components/DecisionIssues.jsx
```diff
@@ -0,0 +1,141 @@
...
+  justifyContent: 'space-between'
+});
+
+const noteDiv = css({
+  fontSize: '1.5rem',
+  color: COLORS.GREY
+});
+
+export default class DecisionIssues extends React.PureComponent {
+  static generateDecisionIssues = (requestIssue, props) => {
```
annekainicUSDS: @youngfreezy I created the static method here as you suggested, and removed the duplicated code from the other component. Let me know if this looks correct to you!

youngfreezyVA: LGTM!

PR #10290, opened 2019-04-03
Limit organization task uniqueness to same parent task 8984

### LINKED ISSUE #8984: Limit organization task uniqueness to same parent task
Currently, we prevent the creation of duplicate tasks assigned to the same organization for a single appeal by [checking the assignee, appeal, and task type](https://github.com/department-of-veterans-affairs/caseflow/blob/eb1928df757dc17c689b9651285af3bbc244aaea/app/models/tasks/generic_task.rb#L6). This could prevent the creation of tasks assigned to the same organization that may not actually be duplicates (a privacy team task from the mail team, and one from the VLJ support, for instance). To prevent this false positive we should additionally check to see whether the potential duplicate task would share the same parent task.

## Acceptance criteria
* [ ] Only throw `Caseflow::Error::DuplicateOrgTask` error when tasks have same parent.

youngfreezyVA:
Resolves #8984


app/models/tasks/generic_task.rb
```diff
@@ -14,12 +14,17 @@ class GenericTask < Task
   def verify_org_task_unique
     return if !active?
 
-    if appeal.tasks.active.where(type: type, assigned_to: assigned_to).any? && assigned_to.is_a?(Organization)
+    if appeal.tasks.active.where(
+      type: type,
+      assigned_to: assigned_to,
+      parent: parent
+    ).any? && assigned_to.is_a?(Organization)
```
lowellrex: Can we add 2 new passing tests that would have failed with the old code?
1. Completed "duplicate" task
2. Active task with all the same properties except that it has a different parent task

Issue #7708, opened 2018-11-05
Post Standup Discussion: How to coordinate shared DB schema


mdbenjam:
We've had some troubles in the past with coordinating database schema design/changes between teams. For example: https://github.com/department-of-veterans-affairs/caseflow/issues/7625

Another example is the claimant-appeal relationship. Claimants really reflect the relationship between a person and an appeal, but the Queue team assumed that the claimant represented the person appealing.

How can we better prevent these situations from happening in the future?

pkarman:
I think we need a Schema Change Committee that needs to sign off on each PR.

pkarman:
Proposed template for PRs that affect db/schema.rb

* [ ] Column comments updated
* [ ] #appeals-metrics notified
* [ ] Reviewed by:
  * [ ] @pkarman 
  * [ ] @aroltsch 

(list of specific folks TBD)

pkarman:
* PR template :yes:
* Danger alert on PR for schema changes
* Remove danger alert on VACOLS change
* create #appeals-schema channel, drop PR link and summary in channel

pkarman:
we've got processes in place we are now following.

PR #10295, opened 2019-04-03
Add Danger rule for schema change, drop Danger for VACOLS, add schema checklist

pkarman:
connects #7708 

### Description

In order to facilitate better communication around changes to the Caseflow schema, adds a section to the PR template and a Danger rule.

Drops the existing Danger rule about testing VACOLS since we now have FACOLS tests locally.


Dangerfile
```diff
@@ -15,8 +15,8 @@ if `git diff #{github.base_commit} spec/ | grep -E '(:focus => true)|(focus: tru
 end
 
 # We must take care of our VACOLS models. Remind developers to test this thoroughly
-if !git.modified_files.grep(/app\/models\/vacols/).empty?
-  warn("This PR changes VACOLS models. Please ensure this is tested against a UAT VACOLS instance")
+if !git.modified_files.grep(/db\/schema.rb/).empty?
```
lowellrex: Can we change this pattern to anything in the `/db` directory so that we include migrations? Sometimes people (myself included!) forget to run the migration locally before pushing up the PR. I think this condition would miss that situation.

pkarman: OTOH, if db/schema.rb isn't modified and db/migrations/ is, should that be its own Danger rule?

lowellrex: I like that even better!

pkarman: thanks @lowellrex -- I added another rule.

Issue #10173, opened 2019-03-27
Get payee_code from claimant information


leikkisa:
Currently the payee_code for Appeal DTAs is getting set from decision_issue.prior_payee_code, which uses the payee_code for the same participant from a previous end product.  If the previous end_product is not available, we should check if the claimant has a payee_code available and use that.  If that is not available, the current behavior (the "No payee code" error) should continue.

Currently, BGSService uses fetch_claimant_info_by_participant_id, which calls find_general_information_by_participant_id and returns the relationship, we should add the payee_type_code to the result so that it is available.

pkarman:
@leikkisa ideas on validation steps?

leikkisa:
Lol, I found a test Veteran with 24 relationships (3 spouses, 21 children)
### Validation instructions
Do an appeal intake for:
UAT Veteran: 700013444

Select anyone that is not the Veteran as the claimant.

Then we'll need to create a remand decision on the request issue, and see if the payee_code gets populated on the claimant for the remand Supplemental Claim.

leikkisa:
Colin says that he does not have the ability to do this.

leikkisa:
I am going ahead and closing this - see comment above.

PR #10298, opened 2019-04-03
Get payee_code from claimant infor

Sjones352:
Connects #10173

### DIFF
--- a/spec/models/decision_issue_spec.rb
+++ b/spec/models/decision_issue_spec.rb
@@ -390,7 +390,7 @@ describe DecisionIssue do
           end
         end
 
-        fcontext "when there is no prior claim by the claimant" do
+        context "when there is no prior claim by the claimant" do
           context "when there is a bgs payee code" do
             before { allow_any_instance_of(Claimant).to receive(:bgs_payee_code).and_return("12") }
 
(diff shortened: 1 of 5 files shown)


spec/models/decision_issue_spec.rb
```diff
@@ -391,15 +391,37 @@
...
-            # verify that both appeal and newly created dta sc have errors
-            expect(SupplementalClaim.find_by(
-                     veteran_file_number: decision_review.veteran_file_number,
-                     establishment_error: "No payee code"
-                   )).to_not be_nil
-            expect(decision_review.establishment_error).to eq("DTA SC creation failed")
+          context "when there is a bgs payee code" do
+            before { allow_any_instance_of(Claimant).to receive(:bgs_payee_code).and_return("12") }
+
+            it "creates a new supplemental claim" do
```
pkarman: not for this PR but we should change this test to not use `subject` multiple times in a single `it` spec. IIRC that will run the `subject` every time, which in this case is less efficient and does not necessarily exercise the test how we want it to.

Sjones352: Thanks @pkarman will fix this in another pr.

PR #10300, opened 2019-04-03
Use scope method when benefit types route to case detail page

Sjones352:
Connects  #9734

Last part of this comment https://github.com/department-of-veterans-affairs/caseflow/pull/10197#discussion_r270223786


app/models/task.rb
```diff
@@ -34,6 +34,8 @@ class Task < ApplicationRecord
 
   scope :inactive, -> { where(status: inactive_statuses) }
 
+  scope :not_decisions_review, -> { where.not(type: "DecisionReviewTask") }
```
pkarman: I don't think this will do what you mean.

When you use a SQL statement to filter, as in this `where` clause, it does not take subclassing into account the way a `reject` statement does. `reject` is performed on the array *after* it is fetched from the db, so it can interrogate the instance and figure out that (e.g.) `VeteranRecordRequest` is_a `DecisionReviewTask`.

If you add some tests to this PR this will become more clear.

You have a couple options if you want to keep this as a SQL `where` inside a `scope`.

* hardcode all the subclass names so it looks something like:
```ruby
 where.not(type: %w[DecisionReviewTask VeteranRecordRequest BoardGrantEffectuationTask])
```
* interrogate the ActiveModel class directly
```ruby
where.not(type: DecisionReviewTask.descendants.map(&:name) + ["DecisionReviewTask"])
```

The second style is probably better since it is more maintainable and less likely someone will forget to update this query if we add another subclass of `DecisionReviewTask`.

Sjones352: @pkarman not sure what test I can write for this as I had written one already in the previous pr.

pkarman:
this patch should fix the failing spec
```diff
diff --git a/spec/factories/task.rb b/spec/factories/task.rb
index 812869038..6aedab958 100644
--- a/spec/factories/task.rb
+++ b/spec/factories/task.rb
@@ -51,7 +51,7 @@ FactoryBot.define do
       status { Constants.TASK_STATUSES.on_hold }
     end

-    factory :generic_task do
+    factory :generic_task, class: GenericTask do
       type { GenericTask.name }
       appeal { create(:appeal) }
     end
```

PR #10305, opened 2019-04-04
Allow MailTask subclasses to be assigned to individual people

### LINKED ISSUE #9964: Bug: User unable to assign tasks to any individual on their team
### Description
During refresher training with AOD team, Charlene (AOD Team Lead) was unable to assign a case to anyone on her team, including herself. The error message was unreadable - see below.

Additionally, we confirmed with Nina today that she has been unable to assign cases from the Lit Support queue to users on her team. She's been trying for the past week. 

Screenshot: 

Screen Shot 2019-03-14 at 11 49 41 AM

lowellrex:
Resolves #9964.


app/models/tasks/generic_task.rb
```diff
@@ -79,18 +79,14 @@ def create_from_params(params, user)
     end
 
     def create_child_task(parent, current_user, params)
-      transaction do
-        parent.update!(status: Constants.TASK_STATUSES.on_hold)
```
lowellrex: This happens automatically thanks to [an `after_create` hook on `Task`](https://github.com/department-of-veterans-affairs/caseflow/blob/fd445d5c976e0e52a5309a1fc926840d3cc3d6f0/app/models/task.rb#L19).

app/models/tasks/mail_task.rb
```diff
@@ -31,22 +31,27 @@ def subclass_routing_options
     def create_from_params(params, user)
       verify_user_can_create!(user, Task.find(params[:parent_id]))
 
-      root_task = RootTask.find(params[:parent_id])
+      parent_task = Task.find(params[:parent_id])
```
lowellrex: We were previously failing with a 404 because active record could not find a `RootTask` with the ID of the new parent which was an `AodMotionMailTask`.

app/models/tasks/mail_task.rb
```diff
@@ -31,22 +31,27 @@ def subclass_routing_options
...
 
         params = modify_params(params)
-        create_child_task(mail_task, user, params)
+        create_child_task(parent_task, user, params)
       end
     end
 
+    def child_task_assignee(parent, params)
+      super || default_assignee(parent, params)
+    end
```
lowellrex: Allows the assignment of `MailTask`s to individual people (by way of [`Task.child_task_assignee`](https://github.com/department-of-veterans-affairs/caseflow/blob/fd445d5c976e0e52a5309a1fc926840d3cc3d6f0/app/models/task.rb#L236)) if the appropriate parameters are passed in.

spec/feature/queue/mail_task_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      it "successfully assigns the task to team member" do
+
+        visit("queue/appeals/#{aod_team_task.appeal.external_id}")
+        click_dropdown(prompt: COPY::TASK_ACTION_DROPDOWN_BOX_LABEL, text:    Constants.TASK_ACTIONS.ASSIGN_TO_PERSON.label)
+        fill_in("taskInstructions", with: "instructions")
+        click_button("Submit")
+
+        expect(page).to have_content(format(COPY::ASSIGN_TASK_SUCCESS_MESSAGE, user.full_name))
+        expect(page.current_path).to eq("/queue")
+        expect(aod_team_task.assigned_to_id).to eq(user.id)
```
lowellrex: It looks like [this test is failing in CI](https://circleci.com/gh/department-of-veterans-affairs/caseflow/57010) but not locally. I suspect the reason it is passing locally is because we are testing only the assignee ID and not the assignee as a whole and the ID of the `AodTeam` in the `organizations` table is the same as the ID of our `user` in the `users` table.

I think these two assertions (lines 49 and 50) are exactly what we want to test (that we 1) successfully created a child task and 2) that the child task is assigned to the right person) but might need a little modification. Maybe a slight change to do something like the following would work:

```ruby
new_tasks = aod_team_task.children
expect(new_tasks.length).to eq(1)

new_task = new_tasks.first
expect(new_task.assigned_to).to eq(user)
```

db/seeds.rb
```diff
@@ -217,9 +221,22 @@ def create_qr_user
     create_task_at_quality_review(qr_user, "Huilen Concepcion", "Ilva Urrutia")
   end
 
-  def create_aod_user
+  def create_aod_user_and_tasks
     u = User.create!(station_id: 101, css_id: "AOD_USER", full_name: "AOD team member")
     OrganizationsUser.add_user_to_organization(u, AodTeam.singleton)
+
+    # TODO: Create factories for mail task creation
```
lowellrex: Let's remove this TODO comment. This is fine for now.

PR #10309, opened 2019-04-04
Fixing VACOLS logic for postponed hearings

### LINKED ISSUE #10116: Bug: Postpone logic moved to 55 (instead of 57)
For this case: https://appeals.cf.ds.va.gov/queue/appeals/2916606 It was postponed and the option, "Send back to Schedule Veterans" was selected, but the appeal was sent to VACOLS location 55 instead of VACOLS 57. 

From Sal: 
When sending a Veteran back to the list to be scheduled should the appeal be charged to location 55??  I thought the appeal would remain in CaseFlow.  The Veteran below was scheduled for a CO hearing and requested to be rescheduled for a Video.  I associated the mail, postponed and sent back to schedule list through the search, not on the daily docket, and the appeal moved to 55.  I charged the appeal back to location 57.

sharonwarner:
Resolves #10116 

### Description
We were accidentally moving appeals to case storage when their hearings were postponed. This is because we were running VACOLS location logic as part of a callback when we shouldn't have been. That callback now confirms the appeal has no open HearingTasks before moving appeals in VACOLS.


app/models/tasks/schedule_hearing_task.rb
```diff
@@ -64,19 +63,9 @@ def create_parent_hearing_task
     end
   end
 
-  def update_location_in_vacols
-    if saved_change_to_status? && appeal.is_a?(LegacyAppeal) && on_hold?
-      AppealRepository.update_location!(appeal, LegacyAppeal::LOCATION_CODES[:caseflow])
-    end
-  end
```
lowellrex: Why did we add this `after_update` hook and why are we getting rid of it?

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -64,19 +63,9 @@ def create_parent_hearing_task
...
-
   # We only want to take this off hold, not actually complete it, like the inherited method does
   def update_status_if_children_tasks_are_complete
-    if appeal.is_a?(LegacyAppeal)
-      AppealRepository.update_location!(appeal, LegacyAppeal::LOCATION_CODES[:schedule_hearing])
-    end
-
-    return update!(status: :assigned) if on_hold?
+    update!(status: :assigned) if on_hold?
```
lowellrex: We've made this change because scheduling hearing now happens in Caseflow instead of VACOLS for `LegacyAppeal`s, correct?

app/models/tasks/schedule_hearing_task.rb
sharonwarner: When we originally implemented this, we scheduled veterans from location 57. So, when we put a Schedule Hearing Task on hold, we moved the associated appeal to Caseflow, and when we finished the admin task, we moved the associated appeal back to VACOLS. We now schedule veterans from location Caseflow, so this logic is no longer necessary. The appeal should remain in Caseflow for the duration of the hearing process.

app/models/tasks/schedule_hearing_task.rb
sharonwarner: Yep!

app/models/tasks/schedule_hearing_task.rb
lowellrex: Perfect!

app/models/tasks/hearing_task.rb
```diff
@@ -11,13 +11,43 @@ class HearingTask < GenericTask
...
+
+  def when_child_task_completed
+    super
+
+    return unless appeal.tasks.active.where(type: HearingTask.name).empty?
+
+    if appeal.is_a?(LegacyAppeal)
+      update_legacy_appeal_location
+    else
+      RootTask.create_ihp_tasks!(appeal, parent)
```
lowellrex: I think the logic is correct now, but it looks like we have a different `parent` than before. Before it would have been a `HearingTask` and now it is a `RootTask`. I think the we have it correct now, but I'm curious about why we were never impacted by the incorrect parent before. Maybe we just never executed that code path?

PR #10318, opened 2019-04-04
Search for reps using bfcorkey

aroltsch:
Resolves #{10254}

### Description
If a new appeal is created for the same case, repkey doesn't get updated to the new appeal so we need to search for rep record using bfcorkey


app/repositories/queue_repository.rb
```diff
@@ -109,7 +109,7 @@ def tasks_for_appeal_query(appeal_id)
     end
 
     def appeal_info_query(vacols_ids)
-      VACOLS::Case.includes(:folder, :correspondent, :representatives)
```
aroltsch: Now because `representatives` is an instance method, we can't use `includes` with it. I don't even think `representatives` was used anywhere where it was included.

aroltsch: I might be wrong and maybe there is a better approach.

app/models/vacols/case.rb
```diff
@@ -281,6 +280,13 @@ def update_vacols_location!(location)
     self.class.batch_update_vacols_location(location, [bfkey])
   end
 
+  def representatives
```
aroltsch: Searching for representatives by repkey is not enough.

PR #10330, opened 2019-04-04
CacheManager

pkarman:
wip

### Description

A service class to manage buckets of cached data by its origin: VACOLS, VBMS, BGS, Caseflow.

Provides utility methods for clearing an entire bucket.

### DIFF
--- a/app/services/cache_manager.rb
+++ b/app/services/cache_manager.rb
@@ -1,0 +1,46 @@
+class CacheManager
+  class NoSuchBucket < StandardError; end
+
+  BUCKETS = {
+    bgs: [
+
+    ],
+    vbms: [
+
+    ],
+    vacols: [
+      "#{Rails.env}_list_of_judges_from_vacols",
+      "#{Rails.env}_list_of_judges_from_vacols_with_name_and_id",
+      "#{Rails.env}_list_of_attorneys_from_vacols",
+      "#{Rails.env}_list_of_hearing_coordinators_from_vacols"
+    ],
+    caseflow: [
+
+    ]
+  }.freeze
+
+  def self.cache_store
+    @cache_store ||= begin
+      redis_conn = Redis.new(url: Rails.application.secrets.redis_url_cache)
+    end
+  end
+
+  def all_cache_keys
+    # skip some persistent key patterns because they are maintained separately.
+    keys = []
+    self.class.cache_store.scan_each do |key|
+      next if key.match?(/^(feature_toggle|functions):/)
+      keys << key
+    end
+    keys
+  end
+
+  def clear(bucket)
+    key_names = BUCKETS[bucket.to_sym]
+    fail NoSuchBucket, bucket unless key_names
+
+    key_names.each do |key|
+      Rails.cache.delete(key)
+    end
+  end
+end


app/services/cache_manager.rb
```diff
@@ -0,0 +1,63 @@
...
+  def self.cache_store
+    @cache_store ||= begin
+      redis_conn = Redis.new(url: Rails.application.secrets.redis_url_cache)
+      Redis::Namespace.new("cache", redis: redis_conn)
+    end
+  end
+
+  def all_cache_keys
+    self.class.cache_store.keys
+  end
```
monfresh: This method and the one above it don't seem to be used.

app/services/cache_manager.rb
```diff
@@ -0,0 +1,63 @@
...
+      "JudgeSchedulePeriod-*-cached-*",
+      "SchedulePeriod-*-cached-*",
+      "RoSchedulePeriod-*-cached-*",
+      "RampElectionIntake-*-cached-*",
+      "IntakeStats-last-calculated-timestamp"
+    ]
+  }.freeze
+
+  def self.rails_cache
+    @rails_cache ||= ActiveSupport::Cache.lookup_store(Rails.configuration.cache_store)
```
monfresh: Isn't this the same as `Rails.cache`? Why can't we call `Rails.cache.delete(key)` directly on line 59 for example?

app/services/cache_manager.rb
pkarman: valid point. they are use-able but they aren't currently test-able because our test env uses the FileCache not the Redis cache.

I used these during development (and manually in UAT) to look at the kinds of keys being stored in Redis.

spec/services/cache_manager_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+      cache_key = "IntakeStats-last-calculated-timestamp"
+      expect(Rails.cache.exist?(cache_key)).to be_falsey
+      IntakeStats.throttled_calculate_all!
+      expect(Rails.cache.exist?(cache_key)).to be_truthy
+
+      subject.clear(:caseflow)
+
+      expect(Rails.cache.exist?(cache_key)).to be_falsey
+    end
+  end
```
monfresh: Can we please add a spec for the `*` matcher to make sure it's working as expected?

app/services/cache_manager.rb
pkarman: some of my research suggested that if we were accessing `Rails.cache` from a rake task in production there might be some issues. I didn't test that out.

spec/services/cache_manager_spec.rb
pkarman: unfortunately the RedisStore and the FileStore implement `delete_matched` in different ways. Redis uses the underlying redis lib string glob patterns https://api.rubyonrails.org/classes/ActiveSupport/Cache/RedisCacheStore.html#method-i-delete_matched and the FileStore uses actual RegExp.

pkarman: I figured out how to test the redis store specifically so I think I caught all the desired changes @monfresh

app/services/cache_manager.rb
monfresh: We're actually using this [RedisStore](https://github.com/redis-store/redis-activesupport/blob/master/lib/active_support/cache/redis_store.rb) via the `redis-rails` gem, not `RedisCacheStore` that comes built in with Rails 5.2+. As such, I was going to suggest simplifying this to `Rails.cache.data.keys`, but when we upgrade to Rails 5.2, if we end up using the built-in RedisCacheStore instead of the `redis-rails` gem (which the gem recommends), I don't think `Rails.cache.data.keys` will work, but we might be able to do `Rails.cache.keys`. For now, this works fine.

app/services/cache_manager.rb
```diff
@@ -0,0 +1,57 @@
...
+      "SchedulePeriod-*-cached-*",
+      "RoSchedulePeriod-*-cached-*",
+      "RampElectionIntake-*-cached-*",
+      "IntakeStats-last-calculated-timestamp"
+    ]
+  }.freeze
+
+  # NOTE that development and test envs currently do not use the Redis store, so this only applies
+  # to production environments (AWS)
+  def self.cache_store
```
monfresh: Since this method is only called once by `all_cache_keys`, should we remove the unnecessary memoization?

pkarman: the memoization is a class instance variable (cached once per class) and the `all_cache_keys` is an instance method. So I think it's still useful to cache it, since you could then instantiate multiple CacheManager objects which would share a redis connection (much like Rails.cache is shared).

monfresh: Agreed. I overlooked the fact that it was a class ivar.

Issue #10316, opened 2019-04-04
Rating issue put on nonrating Board DTA EP


leikkisa:
https://github.com/department-of-veterans-affairs/appeals-support/issues/4841

After investigation, the original Request Issue is a rating issue, with a rating reference ID, so this appears to be a bug.

Appeal: 1040
Remand SC: 4449
Original request issue: 5234

We should switch to previous_request_issue.rating? now that previous_request_issue is available.

There may also be a request for manual clean up.

pkarman:
@leikkisa would you include some validation steps here?

leikkisa:
This might be very similar to 10187, but this is for a DTA Appeal, so I'll modify those steps here in case there is a difference:

Do an Appeal for UAT Veteran 700032874
Contest the issue with this information:
Decision date: 2019-01-25
Text: "Evaluation of PTSD, which is currently 50 percent disabling, is continued."

Take note of the Appeal ID for the Appeal that is created.

Then create a remand decision for the appeal. This script should help:
```
# put the UUID of the Appeal being decided here
uuid = "ed984057-cc2b-473b-a6cf-76b1145667d2"
disposition="remanded"
# the citation number needs to be unique, so it will need to change everytime you run this
citation_number = "A19000200"

a = Appeal.find_by(uuid: uuid.upcase) #or find by the appeal ID
a.tasks.where(type: "BvaDispatchTask").delete_all
RequestStore[:current_user] = User.last
a.request_issues.each do |r|
decision_issue = DecisionIssue.create!(disposition: disposition, description: r.description, benefit_type: r.benefit_type, diagnostic_code: "5000", participant_id: a.veteran.participant_id, decision_review: a, rating_issue_reference_id: r.contested_rating_issue_reference_id)
RequestDecisionIssue.create!(decision_issue: decision_issue, request_issue: r)
end
root_task = a.tasks.find_by(type: "RootTask")
BvaDispatchTask.create_from_root_task(root_task)
u = a.reload.tasks.find_by(type: "BvaDispatchTask", assigned_to_type: "User").assigned_to
file = "1234"
BvaDispatchTask.outcode(a.reload, { citation_number: citation_number, decision_date: Time.zone.now, redacted_document_location: "\\\\bvacofil1.dva.va.gov\\archdata$\\arch1901\\${citation_number}.txt", file: file }, u)
```

[...]

PR #10331, opened 2019-04-04
Use previous_request_issue to determine if issue is rating

leikkisa:
connects #10316

Previously we checked the previous rating issue's end product establishment to determine if it was rating. Switch to using previous_request_issue which does not depend on the end_product_establishment.


leikkisa:
There are 16 Nonrating Board DTA EPs in Prod right now, and all of them have rating issues on them.

PR #10342, opened 2019-04-05
Fix var name (currently zero spec coverage)

pkarman:
fixes regression in #10191


monfresh:
The `json_appeals` method definitely gets called via tests such as `appeals_controller_spec.rb:285`. It didn't raise an error because we stub the `current_user` method in tests: https://github.com/department-of-veterans-affairs/caseflow/blob/master/spec/rails_helper.rb#L120

PR #10349, opened 2019-04-05
Reset veteran invalid in redux

leikkisa:
The Veteran is getting an invalid Veteran error on the Add Issues page for a valid Veteran.  I don't know if this will fix it, but the user did have one invalid Veteran in their intake history (due to having an invalid file number), and if the redux store didn't get reset maybe the error persisted.


pkarman:
can't hurt!

PR #10351, opened 2019-04-05
Missing decision date error

pkarman:
connects #10070 

## Description

If one or more request issues throw a MissingDecisionDate exception, it's likely because they have a locked rating. So say that to the user.

Screen Shot 2019-04-05 at 2 30 25 PM


pkarman:
example of what this fixes https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/4489/

PR #10356, opened 2019-04-08
Allow MailTasks to be created for dispatched appeals

lowellrex:
Resolves an issue [identified by Ivy's team last week](https://dsva.slack.com/archives/C6E41RE92/p1554323146081200).

We previously did not allow the creation of `MailTask`s for dispatched appeals because we hid the "+ Add a new task" link. This PR enables the creation of `MailTask`s for dispatched appeals by un-hiding that link.

### DIFF
--- a/app/models/tasks/root_task.rb
+++ b/app/models/tasks/root_task.rb
@@ -39,6 +39,10 @@ class RootTask < GenericTask
     true
   end
 
+  def actions_allowable?(_user)
+    true
+  end
+
   def assigned_to_label
     COPY::CASE_LIST_TABLE_CASE_STORAGE_LABEL
   end
--- a/client/app/queue/components/AssignHearingModal.jsx
+++ b/client/app/queue/components/AssignHearingModal.jsx
@@ -30,7 +30,7 @@ import {
 import Link from '@department-of-veterans-affairs/caseflow-frontend-toolkit/components/Link';
 import {
   appealWithDetailSelector,
-  actionableTasksForAppeal
+  scheduleHearingTasksForAppeal
 } from '../selectors';
 import { onReceiveAmaTasks, onReceiveAppealDetails } from '../QueueActions';
 import { prepareAppealForStore } from '../utils';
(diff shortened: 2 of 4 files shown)


app/models/tasks/root_task.rb
```diff
@@ -39,6 +39,10 @@ def actions_available?(_user)
     true
   end
 
+  def actions_allowable?(_user)
+    true
+  end
+
```
lowellrex: Need to add this so that [the assignment action](https://github.com/department-of-veterans-affairs/caseflow/blob/c0a91afbfd37d30bb48a79da6ba0a6dfea66a7d0/app/models/task.rb#L212) in the new test succeeds.

client/app/queue/selectors.js
```diff
@@ -204,7 +200,12 @@ export const completeTasksByAssigneeCssIdSelector = createSelector(
 );
 
 export const actionableTasksForAppeal = createSelector(
-  [getTasksForAppeal], (tasks) => _.filter(tasks, (task) => task.availableActions.length)
+  [getAllTasksForAppeal], (tasks) => _.filter(tasks, (task) => task.availableActions.length)
```
lowellrex: This is the change that fixed the code by returning completed tasks that have actions available.

client/app/queue/components/AssignHearingModal.jsx
```diff
@@ -381,10 +381,7 @@ class AssignHearingModal extends React.PureComponent {
 }
 
 const mapStateToProps = (state, ownProps) => ({
-  scheduleHearingTask: _.find(
-    actionableTasksForAppeal(state, { appealId: ownProps.appealId }),
-    (task) => task.type === 'ScheduleHearingTask' && task.status !== 'completed'
-  ),
+  scheduleHearingTask: scheduleHearingTasksForAppeal(state, { appealId: ownProps.appealId })[0],
```
youngfreezyVA: I would assign a variable in the opening of `mapStateToProps`  and take advantage of es6 array destructuring like this: `[scheduleHearingTask] =  scheduleHearingTasksForAppeal(state, { appealId: ownProps.appealId })`
and then use object shorthand notation for the ` scheduleHearingTask` prop. not necessary but makes use of es6 features and we aren't hardcoding the index (though we implicitly are with the array destructuring).

PR #10363, opened 2019-04-08
Find End Product without End Product Establishment for RAMP

leikkisa:
connects #9964

If the EndProductEstablishment for a RAMP intake isn't saved, find the matching end product when needed


app/models/end_product_establishment.rb
```diff
@@ -173,10 +177,10 @@ def cancel_unused_end_product!
   end
 
   def sync!
+    fail EstablishedEndProductNotFound, id unless result
```
leikkisa: Moved this because there was a circular reference

app/models/ramp_refiling_intake.rb
```diff
@@ -79,7 +79,7 @@ def validate_detail_on_start
       @error_data = veteran_invalid_fields
     elsif ramp_elections.empty?
       self.error_code = :no_complete_ramp_election
-    elsif ramp_elections.all?(&:end_product_active?)
+    elsif ramp_elections.any?(&:end_product_active?)
```
leikkisa: I believe this was a mistake in the original logic

app/models/end_product_establishment.rb
leikkisa: Sounds good! I was thinking it would call status_active? which could call sync! again, but in this case we're calling it with sync being false, I missed that, because in the other case we're calling status_active? with sync being true.

app/models/ramp_refiling_intake.rb
leikkisa: I still think this is a mistake. 

Maybe what we need to check here is if there is at least one cleared RampElection. What we're seeing is that there can be active ones and canceled ones, and it seems to me that if there is an active RampElection (of which there can only be one at a time for each of the two modifiers) and a canceled RampElection, and no cleared RAMP elections, then it's not ready for a RampRefiling.

leikkisa: That said, this change isn't necessary for this PR, I just thought it was wrong logic, so we could also wait on this change for Shane.

leikkisa:
I would be a big fan of doing a backfill for the unconnected RampElections, that would be less hacky.  Shane is nervous about backfilling data at all so this is more of a bandaid.  However, I do feel that the explanation here: https://github.com/department-of-veterans-affairs/caseflow/issues/9946
is satisfying enough to me to justify backfilling the missing unconnected EPEs.

app/models/ramp_refiling_intake.rb
pkarman: if it's not necessary for this PR, let's revert that line for now and open a ticket to remind ourselves to ask Shane.

leikkisa: New ticket: https://github.com/department-of-veterans-affairs/caseflow/issues/10377

PR #10366, opened 2019-04-08
Allow team admins to add other admins

### LINKED ISSUE #7974: Enable way to make user an admin of an organization in web UI
#7584 added a interface for people using Caseflow to manage team membership. #7852 added a new `admin` field in the `organizations_users` table that indicates whether somebody is an admin of a given team. We should add the ability to manage admin privileges in that interface. The admin field is currently only used for determining who is the judge on a judge team, but will soon be used for other purposes (like making sure VLJ support staff admins do not get assigned tasks, and maybe managing team membership).

image

## Acceptance criteria
* People using Caseflow should be able to make members of an team an admin of that team.

lowellrex:
Resolves #7974.

image

Looking for a design review on this approach to adding the "Make user admin" button.


app/controllers/organizations/users_controller.rb
```diff
@@ -22,7 +22,21 @@ def index
...
 
-    render json: { users: json_users([user_to_modify]) }, status: :ok
+    render json: { users: json_managed_users([user_to_modify]) }, status: :ok
+  end
+
+  def update
+    no_cache
+
+    if params.key?(:admin)
+      if params[:admin]
```
monfresh: What do you think about making this an explicit check for `true`? That way, if `params[:admin]` is the string `"false"`, for example, the conditional will fail.

app/controllers/organizations/users_controller.rb
```diff
@@ -56,4 +70,12 @@ def organization_url
   def json_users(users)
     ::WorkQueue::UserSerializer.new(users, is_collection: true)
   end
+
+  def json_managed_users(users)
```
monfresh: What is the reason for treating this as an array if we're only going to pass in a single user every time?

app/models/serializers/work_queue/user_serializer.rb
```diff
@@ -4,4 +4,8 @@ class WorkQueue::UserSerializer
   include FastJsonapi::ObjectSerializer
   attribute :css_id
   attribute :full_name
+
+  attribute :is_admin do |object, params|
+    params[:check_if_admin] ? params[:organization].user_is_admin?(object) : nil
+  end
```
monfresh: Is there a reason the value must be `nil` as opposed to `false`? If not, we could simplify this:
```ruby
attribute :is_admin do |object, params|
  params[:organization].user_is_admin?(object)
end
```
and eliminate the need to pass in the `check_if_admin` param.

monfresh: Even better, I would create a new serializer called `OrganizationUserSerializer` that inherits from `UserSerializer` and only defines the `is_admin` method, which I would personally rename to just `admin`. The Ruby Style Guide discourages prefixing method names with `is`, `does` or `can`: https://github.com/rubocop-hq/ruby-style-guide#bool-methods-prefix

The reason for creating a new serializer is so that you don't have to worry about whether or not the organization param will be present. This allows to adhere to the [Tell, Don't Ask](https://thoughtbot.com/blog/tell-dont-ask) principle.

spec/models/organizations_user_spec.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+describe OrganizationsUser do
+  let(:organization) { FactoryBot.create(:organization) }
+  let(:user) { FactoryBot.create(:user) }
```
monfresh: Out of curiosity, why do you prefer the long-form way of creating factories via `FactoryBot.create(:user)` as opposed to the shortcut `create(:user)`?

lowellrex: I prefer the explicitness of including the module name so that we know where the function is coming from. I'll probably remember that `create` in test files means FactoryBot's `create` method (and not ActiveRecord's `create` function for instance), but I think the explicitness makes it clear to any newcomers to the code where that comes from (and me in moments of weakness).

I'm open to removing the `FactoryBot.` prefix across the board if we think it's better, but I can't think of much downside to keeping it for the time being.

PR #10382, opened 2019-04-10
progress pr - Batch requests in task table

### LINKED ISSUE #8934: [Tech spec] How should we batch async calls from the frontend?
related ticket https://github.com/department-of-veterans-affairs/caseflow/pull/8904

### Description
image

#### The issue
In most `<TaskTable>`s, we do a couple calls per `Task` asynchronously in order to quickly load the Tasks but lazy load information that would come from VACOLS or other expensive sources.

However, Chrome (for instance) can only make 6 requests to the same host simultaneously. If there are 30 tasks, and 3 async calls per task (for the hearing, number of documents, and new files indicator), this process can take a really long time.

@lowellrex @mdbenjam @amprokop @tomas-nava, what's your opinion on how we should handle this in Rails? @lowellrex proposed considering websockets, and I don't have much of a clue how to batch these requests in Rails.

youngfreezyVA:
Resolves #8934

### DIFF
--- a/app/controllers/appeals_controller.rb
+++ b/app/controllers/appeals_controller.rb
@@ -113,7 +113,7 @@ class AppealsController < ApplicationController
   end
 
   def appeal_by_ids(id)
-    @appeal_by_ids ||= Appeal.find_appeal_by_id_or_find_or_create_legacy_appeal_by_vacols_id(id)
+    Appeal.find_appeal_by_id_or_find_or_create_legacy_appeal_by_vacols_id(id)
   end
 
   def url_appeal_uuid
--- a/app/controllers/appeals_controller.rb
+++ b/app/controllers/appeals_controller.rb
@@ -40,7 +40,8 @@ class AppealsController < ApplicationController
     ids = params[:appeal_ids].split(',')
     document_counts_by_id = {}
     ids.each do |id| 
-      document_counts_by_id[id] = appeal_by_ids(id).number_of_documents 
+      doc_count = Appeal.find_appeal_by_id_or_find_or_create_legacy_appeal_by_vacols_id(id).number_of_documents 
+      document_counts_by_id[id] = doc_count
     end
     render json: { document_counts_by_id: document_counts_by_id }
     rescue Caseflow::Error::EfolderAccessForbidden => e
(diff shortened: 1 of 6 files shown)


app/controllers/appeals_controller.rb
```diff
@@ -34,12 +34,23 @@ def show_case_list
...
-  def document_count
-    render json: { document_count: appeal.number_of_documents }
-  rescue Caseflow::Error::EfolderAccessForbidden => e
-    render(e.serialize_response)
-  rescue StandardError => e
-    handle_non_critical_error("document_count", e)
+  def document_counts_by_id
+    render json: { document_counts_by_id: build_document_counts_hash }
+  rescue Caseflow::Error::EfolderAccessForbidden => error
+    render(error.serialize_response)
```
aroltsch: This entire request would error out even if one appeal fails. I think we should catch this error on per appeal basis. If one appeal fails, set it to error and continue with others.

app/controllers/appeals_controller.rb
```diff
@@ -34,12 +34,25 @@ def show_case_list
...
+  end
+
+  def build_document_counts_hash
```
youngfreezyVA: @monfresh do you have any thoughts on how to get code climate to not complain about this having 7 statements?

app/controllers/appeals_controller.rb
```diff
@@ -50,26 +63,31 @@ def power_of_attorney
...
-          type: most_recently_held_hearing.readable_request_type,
-          external_id: most_recently_held_hearing.external_id,
-          disposition: most_recently_held_hearing.disposition
-        }
-      else
-        {}
+    render json: { most_recently_held_hearings_by_id: build_most_recently_held_hearings_hash }
+  end
+
+  def build_most_recently_held_hearings_hash
```
youngfreezyVA: @monfresh same here

monfresh: I will take a look

youngfreezyVA:
Hey moncef, There will be a lot of changes in the next couple of days so please hold off I’ll reach back out when it’s ready thank you

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

app/controllers/appeals_controller.rb
youngfreezyVA: there's going to be alot of changes to this, so please hold off for now i'll reach back out when its ready if i'm still facing this issue, thank you!

PR #10385, opened 2019-04-10
Fix attorney case review caching

aroltsch:
We received a few support tickets where a wrong document ID was displayed on the case details page. I checked DB records and they all look correct for these appeals. Then I thought the problem was here: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/serializers/work_queue/appeal_serializer.rb#L124. Ruby operator `||=` assigns a value if not already assigned. Then I noticed this PR changed the method to be a class method: https://github.com/department-of-veterans-affairs/caseflow/pull/10191/files#diff-d52c63713877ca3617b80d8ea1ae658cR123. I think that's what caused the issue. If I remove `||=`, it displays correct document ID every time.


aroltsch:
@youngfreezy indeed :)

monfresh:
Good catch! I believe there is a way to memoize class methods. It would be good to try to avoid making 6 DB calls to look up the Attorney Case Review. Can you try this?

```ruby
def self.latest_attorney_case_review(appeal)
    @latest_attorney_case_review ||= Hash.new do |hash, key|
      hash[key] = AttorneyCaseReview.where(task_id: Task.where(appeal: appeal).pluck(:id)).order(:created_at).last
    end
    @latest_attorney_case_review[appeal]
  end
```

PR #10392, opened 2019-04-10
Add Docket Cutoff Lines to Assign Hearings Page

### LINKED ISSUE #9753: Daily Docket Range Line AMA
As a hearing coordinator, I need to know what dates to schedule for in the AMA Veterans tab based on the Caseflow Algorithm.

### Acceptance Criteria 
- [ ] Verify that we are dispalying the current schedule date based on the Case Distribution algorithm. 
- [ ] Verify that we show a gray HZ line with the following copy: 
 - Schedule for [date] [year]
- [ ] Verify that line updates as needed to reflect the new cases that should be scheduled by month.
- [ ] Verify that when there are no more veterans to schedule there is copy in the table with the following: 
 - All veterans have been scheduled for this hearing month. Please check back agian tomomorrow
 

### Mocks 
hearing docket date ama

_If all are scheduled_
Hearing docket date no hearings

### Related Tickets 
* #9753 - Hearing schedule date 
* #9752  - Legacy hearing docket date 
* #9754 - Warning message

lomaxap:
Resolves #9753
- add cutoff line to Assign Hearing's AMA Veteran's Waiting tab to indicate end of docket range


app/models/dockets/hearing_request_docket.rb
```diff
@@ -16,4 +16,20 @@ def distribute_appeals(distribution, priority: false, genpop: "any", limit: 1)
...
+  def appeals_in_docket_range
+    target = DocketCoordinator.new.target_number_of_ama_hearings_this_month
+    target = 30 if target == 0 # for testing
+
+    appeals_in_range = appeals(priority: false)
+    held_hearings = Hearing.where(appeal: appeals_in_range, disposition: "held").where.not(
+      hearing_day: HearingDay.where(scheduled_for: Time.zone.now.beginning_of_month..Time.zone.now.end_of_month)
+    )
+
+    appeals_in_range.where.not(id: held_hearings.pluck(:appeal_id)).limit(target)
```
cmgiven: I think I spot an issue here: if an appeal became “in range” in a previous month, it would be counted against this month’s target. But this month’s target is telling us how many additional appeals are now in range, not the total number of appeals in range. So we need some way to know which appeals were previously in range so that we can just select “target” number of new appeals.

lomaxap: Ok. I don't think I understand what "in range" is then. 

My understanding is that `dockets[:hearing].appeals` will give me all appeals that have not had a decision. In the context of hearing schedule, I'm thinking it's irrelevant when exactly the appeal became "in range." For our purposes, we just need to know whether or not the appeal has had a hearing or not. If it hasn't had a held hearing, then it should still show up in this list.

This method would give me all appeals that don't have a decision AND have not had a held hearing. What am I missing?

app/controllers/hearings_controller.rb
```diff
@@ -48,6 +48,25 @@ def find_closest_hearing_locations
     end
   end
 
+  def docket_range
```
sharonwarner: Would this make more sense to have in the appeals controller?

app/models/dockets/hearing_request_docket.rb
cmgiven: When you ask for the target number of appeals for May, that's asking you for Caseflow to schedule an additional _n_ appeals. But we might not have scheduled all of the appeals from April, and even if we did schedule them, they might not be decided, so they'd still appear in `dockets[:hearing].appeals`.

When it became "in range" doesn't matter, except insofar as we want to know whether the appeal was already "in range" last month. Because if it was "in range," then it shouldn't count toward the target number of additional appeals we need to schedule this month.

lomaxap: I *think* I understand. I was assuming `target_number_of_ama_hearings` returned a consistent number based on the number of days (something simple like 20 days * 12 judges * 11 hearings).  But it's literally how many hearings we should have held by now to reach a yearly target, correct?

> if it was "in range," then it shouldn't count toward the target number of additional appeals we need to schedule this month.

I think for what I'm trying to do I want all appeals that should have been scheduled by now. So if we could schedule 30 appeals in April and 30 in May, but only managed to schedule 25 in April, I think I'd want the target to be 35 for May.

cmgiven: That's right. It's actually not too dissimilar from your consistent number, except that there's no judge multiplier and that the hearings multiplier does change over time. But note that that's not how many hearings we should have held by now, but rather the number of additional hearings we need to add to our all-time total.

I'd suggest that what we need to do here is somehow mark those appeals that are in range. That way, each month we just mark the target number of additional appeals, based on docket order but skipping those that have already been marked.

app/models/serializers/work_queue/legacy_task_serializer.rb
```diff
@@ -12,6 +12,7 @@ class WorkQueue::LegacyTaskSerializer
   attribute :assigned_on
   attribute :docket_name
   attribute :docket_date
+  attribute :docket_range_date, &:docket_date
```
sharonwarner: Are both of these definitely eager loaded? I just want to make sure we don't blow up the page load time.

Issue #10215, opened 2019-03-29
Update error message for Selection Form already processed


pkarman:
See https://dsva.slack.com/archives/C754H751T/p1553873240030300

Current text is:

> Selection Form already processed in Caseflow Caseflow does not currently support more than one Selection Form for a Veteran.

New text:

> Please contact your management team

instead of:

> Please contact Caseflow Support

Sjones352:
@leikkisa please provide validation instructions.

Sjones352:
**Passes UAT**

![Screen Shot 2019-04-14 at 6.02.05 PM.png](https://images.zenhubusercontent.com/5b8fddd1b2cc7b6c78ece9f9/fa73562d-63ce-4e80-855a-93317e273374)

PR #10395, opened 2019-04-10
Update copy

leikkisa:
connects #10215


leikkisa:
There are 5 other places that say "Please contact Caseflow Support". Three of then are "something went wrong" errors, but I'm wondering if we would want to update the following two as well:

````Other end products associated with this ${details.formName} have already been decided,
      so issues are no longer editable. If this is a problem, please contact Caseflow support.````
```'Caseflow temporarily does not support closing appeals with any non-compensation issues.' +
          'Please contact Caseflow Support on how to proceed.'```

PR #10396, opened 2019-04-10
9643 delay VBMS submit for processing

leikkisa:
#9643 
Add 7 hours to automatic claims that get established in VBMS, if a delay was added due to the date being in the future.  

Note - these might need an additional 3 hours (because the delay comes up 3 hours short).


spec/models/higher_level_review_spec.rb
```diff
@@ -324,15 +324,15 @@
...
-          it "creates a DTA Supplemental claim, but does not start processing until the claim_date" do
+          it "creates a DTA Supplemental claim, but does not start processing until hours after the claim_date" do
             subject
             dta_sc = SupplementalClaim.find_by(
               receipt_date: caseflow_decision_date,
               decision_review_remanded: higher_level_review
             )
             expect(dta_sc).to_not be_nil
             expect(dta_sc.establishment_submitted_at).to eq(
-              caseflow_decision_date.to_date.to_datetime - 3.hours + 1.minute
```
leikkisa: I'm not sure why these are offset by 3 hours, but I did notice this for decision_documents in prod too. Do I need to add another 3 hours?

pkarman: the initial 3 hour delay is for everything, to avoid race condition where a cron job might start at the same time that the job is being processed manually just after it was submitted for processing.

app/models/decision_review.rb
```diff
@@ -226,7 +226,7 @@ def create_remand_supplemental_claims!
       rsc.create_remand_issues!
       rsc.create_decision_review_task_if_required!
 
-      delay = rsc.receipt_date.future? ? rsc.receipt_date : 0
+      delay = rsc.receipt_date.future? ? (rsc.receipt_date + 7.hours).utc : 0
```
pkarman: let's make the `7` a named constant so it's easier to remember (a) why it exists and (b) why it is 7.

how about `PROCESS_DELAY_VBMS_OFFSET_HOURS = 7` and then call like `PROCESS_DELAY_VBMS_OFFSET_HOUR.hours`

app/models/tasks/bva_dispatch_task.rb
```diff
@@ -36,7 +36,7 @@ def outcode(appeal, params, user)
 
     def create_decision_document!(params)
       DecisionDocument.create!(params).tap do |decision_document|
-        delay = decision_document.decision_date.future? ? decision_document.decision_date : 0
+        delay = decision_document.decision_date.future? ? (decision_document.decision_date + 7.hours) : 0
```
pkarman: same comment here about named constant. we could put it in the Asyncable module itself if that would help centralize.

spec/models/higher_level_review_spec.rb
pkarman: no need to add another 3 hours, but probably the math would be easier if we wrote the test like:

```ruby
caseflow_decision_date.to_date + DecisionReview::PROCESS_DELAY_VBMS_OFFSET_HOURS.hours - SupplementalClaim.processing_retry_interval_hours.hours + 1.minute
```

spec/models/tasks/bva_dispatch_task_spec.rb
```diff
@@ -129,15 +129,15 @@
...
+        let(:decision_date) { 1.day.from_now }
 
         it "sets a delay on the enqueued_job" do
           expect do
             BvaDispatchTask.outcode(root_task.appeal, params, user)
           end.to_not have_enqueued_job(ProcessDecisionDocumentJob)
 
           decision_document = DecisionDocument.find_by(appeal_id: root_task.appeal.id)
-          expect(decision_document.submitted_at).to eq(decision_date + 1.minute - 8.hours)
+          expect(decision_document.submitted_at).to eq((decision_date.to_date + 1.minute + 4.hours))
```
pkarman: same comment here about naming the hours for clarity.

app/models/decision_review.rb
leikkisa: 👍

Sjones352:
@leikkisa can you please provide instructions on how to validate this one.

PR #10397, opened 2019-04-11
Rename issue_category to nonrating_issue_category

leikkisa:
connects #8239

Rename instances where the field is used and remove issue_category.  The data is verified in UAT and Prod.

*Schema Changes Only*

* [x] Column comments updated


Sjones352:
This is a copy change that is also linked to #8239 and therefore does not need UAT validation.

Issue #9795, opened 2019-03-05
Intake | Bust cache on Veteran multiple phone exception


pkarman:
When a Veteran suffers from the multiple active phone exception, bust the cache on the `can_access?` BGS call so that if a user fixes the problem in SHARE, they do not need to wait for 24 hours to proceed in Caseflow.

Sjones352:
@leikkisa please provide validation instructions

pkarman:
Validation instructions:

* identify a Veteran in UAT with duplicate phone number records. IIRC Colin was researching this.
* attempt an Intake for that Veteran
* confirm the special phone number error message
* resolve the duplicate phone numbers in UAT
* immediately try the Intake again in UAT. The error message should not appear.

Sjones352:
Asked Colin to create me a veteran with duplicate phone number.

Sjones352:
This is going to be a hard one to validate as VBMS does not create Veteran with phone number records. @leikkisa do you know anyone at BGS who might be able help us?
<img width="282" alt="Screen Shot 2019-04-25 at 10 21 18 AM" src="https://user-images.githubusercontent.com/23080951/56743246-3bc95c80-6744-11e9-8d3b-072117a8bd54.png">

leikkisa:
@Sjones352 I think we should validate in prod with our users. I asked someone about that, and they're getting an error trying to update the Veteran's phone numbers in VBMS, so we probably can't validate this anyway until that issue is fixed.

leikkisa:
I learned during a BGS/VBMS meeting that users cannot update the phone number themselves, so we won't be able to validate this.  Data clean up is being led by a person named Jack Goldberg.  I've heard that they've merged all instances where the multiple phone numbers matched each other. If they didn't match, but were valid, they picked the newest phone number. The remaining cases have invalid phone numbers.

I am closing this issue since we can't validate it.

PR #10400, opened 2019-04-11
Bust cache when veteran has multiple phone numbers

leikkisa:
connects #9795


spec/feature/intake/intake_spec.rb
```diff
@@ -170,6 +170,20 @@
...
+
+        safe_click ".Select"
+        fill_in "Which form are you processing?", with: Constants.INTAKE_FORM_NAMES.ramp_election
+        find("#form-select").send_keys :enter
+        safe_click ".cf-submit.usa-button"
+
+        fill_in search_bar_title, with: "12341234"
+        click_on "Search"
+
+        expect(page).to have_current_path("/intake/review_request")
```
leikkisa: I confirmed the behavior when bust_can_access_cache isn't called, not sure if I should formally that in a test or if the successful test is good enough.

app/services/external_api/bgs_service.rb
```diff
@@ -187,6 +187,11 @@ def can_access?(vbms_id)
     end
   end
 
+  def bust_can_access_cache(vbms_id)
+    cache_key = "bgs_can_access_#{current_user.css_id}_#{current_user.station_id}_#{vbms_id}"
```
pkarman: this seems ripe to put into a `user_can_access_vbms_id_cache_key` method and re-use it in `can_access?`

spec/feature/intake/intake_spec.rb
pkarman: I think I'd want to see an explicit test for the `cache` behavior, i.e. that the key doesn't exist when the non-unique error is caught.

leikkisa: 👍

app/services/external_api/bgs_service.rb
leikkisa: 👍

spec/feature/intake/intake_spec.rb
```diff
@@ -170,6 +170,24 @@
...
+        expect(Rails.cache.exist?(cache_key)).to eq(true)
```
pkarman: can we mock the non-unique error so we can verify that the cache key does not exist in that case?

leikkisa: This is on line 175 (after the non-unique error), and then the cache gets set again on trying the intake again.

pkarman: ah sorry I misread that originally as checking that the inittial cache state was empty.

looks great!

Sjones352:
This is connected #9795. Once validation instructions are provided, then I will proceed with testing

Issue #9987, opened 2019-03-14
Intake Edit | Withdrawal confirmation for VBMS based decision reviews


shanear:
We need to show a confirmation for users that issues were withdrawn.

All withdrawal/removal mocks: https://va-gov.invisionapp.com/share/BRQUJR5J5AE#/screens/351540954

requires https://github.com/department-of-veterans-affairs/caseflow/issues/9985

## Acceptance Criteria
- When a save is processed on the Add/Remove issues screen for a VBMS based review where one or more issues were withdrawn, a section should appear on the confirmation page (see: https://github.com/department-of-veterans-affairs/caseflow/issues/7313) saying "Issues Withdrawn" with a checkbox, showing all withdrawn issues.
- If an entire decision review is withdrawn, it should behave the same as the "Review Removed" confirmation. Except it should say "Review Withdrawn" (see: https://github.com/department-of-veterans-affairs/caseflow/issues/9734)

## Notes
- Caseflow based withdraws will be handled here: https://github.com/department-of-veterans-affairs/caseflow/issues/9988

Sjones352:
Covers Acceptance Criteria

1.  If an entire decision review is withdrawn, it should behave the same as the "Review Removed" confirmation. Except it should say "Review Withdrawn" (see: https://github.com/department-of-veterans-affairs/caseflow/issues/9734)

![Screen Shot 2019-04-14 at 5.23.48 PM.png](https://images.zenhubusercontent.com/5b8fddd1b2cc7b6c78ece9f9/9ee06b85-03d9-4f22-81b4-b9581aec3125)

@leikkisa I am guessing that first criteria is handled in the next pr?

PR #10418, opened 2019-04-11
VBMS withdraw confirmation

leikkisa:
connects #9987


leikkisa:
Screen Shot 2019-04-12 at 4 11 11 PM

PR #10422, opened 2019-04-12
Fix N+1 queries in Appeal and Task serializers

monfresh:


app/models/appeal.rb
```diff
@@ -159,8 +159,11 @@ def every_request_issue_has_decision?
   end
 
   def latest_attorney_case_review
-    @latest_attorney_case_review ||=
-      AttorneyCaseReview.where(task_id: tasks.pluck(:id)).order(:created_at).last
+    return @latest_attorney_case_review if defined?(@latest_attorney_case_review)
```
monfresh: This allows us to eliminate unnecessary DB calls when there is no latest_attorney_case_review. With regular `||=` memoization, if the value is `nil`, the DB call will still be made, but by using `if defined?`, we can eliminate those calls.

config/environments/development.rb
```diff
@@ -4,6 +4,7 @@
     Bullet.bullet_logger = true
     Bullet.console       = true
     Bullet.rails_logger  = true
+    Bullet.unused_eager_loading_enable = false
```
monfresh: This allows us to focus on the N+1 queries only.

spec/support/bullet.rb
```diff
@@ -0,0 +1,14 @@
...
+    config.before(:each) do
+      Bullet.start_request
+    end
+
+    config.after(:each) do
+      Bullet.perform_out_of_channel_notifications if Bullet.notification?
+      Bullet.end_request
+    end
+  end
+end
```
monfresh: This allows us to test for N+1 queries when running tests by setting `Bullet.enable = true` in `test.rb`.

app/models/queues/attorney_queue.rb
```diff
@@ -18,12 +19,34 @@ def tasks
       end
     end
 
-    caseflow_tasks = Task.incomplete_or_recently_closed
+    caseflow_tasks = Task.includes(*ama_task_includes).incomplete_or_recently_closed
```
pkarman: nice!

app/models/task.rb
```diff
@@ -176,7 +176,12 @@ def days_waiting
   end
 
   def latest_attorney_case_review
-    AttorneyCaseReview.where(task_id: Task.where(appeal: appeal).pluck(:id)).order(:created_at).last
+    return @latest_attorney_case_review if defined?(@latest_attorney_case_review)
```
pkarman: 👍

app/models/appeal.rb
aroltsch: Nice

app/models/legacy_appeal.rb
```diff
@@ -24,6 +24,7 @@ class LegacyAppeal < ApplicationRecord
   has_one :special_issue_list, as: :appeal
   has_many :record_synced_by_job, as: :record
   has_many :available_hearing_locations, as: :appeal, class_name: "AvailableHearingLocations"
+  has_many :claimants, -> { Claimant.none }
```
aroltsch: Interesting. This would allow you to call `includes` I suppose.

app/models/task.rb
aroltsch: egggcellent

spec/support/bullet.rb
aroltsch: I really hope at some point we can enable it for all tests.

monfresh:
Yes, that is my goal. After we merge this one, I will push a branch with Bullet enabled and see how many tests fail and then I'll fix them one by one.

app/models/legacy_appeal.rb
monfresh: Yeah, I thought this would be easier than having conditionals in the various places where we might have a combination of legacy and AMA appeals.

PR #10423, opened 2019-04-12
Batch document counts

### LINKED ISSUE #8934: [Tech spec] How should we batch async calls from the frontend?

youngfreezyVA:
Resolves #8934


app/workflows/document_counts_by_appeal_id_hash.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+class DocumentCountsByAppealIdHash
```
aroltsch: Let's remove `Hash` from the class name.

app/controllers/appeals_controller.rb
```diff
@@ -34,12 +34,14 @@ def show_case_list
...
   end
 
-  def document_count
-    render json: { document_count: appeal.number_of_documents }
-  rescue Caseflow::Error::EfolderAccessForbidden => e
-    render(e.serialize_response)
-  rescue StandardError => e
-    handle_non_critical_error("document_count", e)
+  def document_counts_by_id
+    render json: {
```
aroltsch: Let's raise an error if there are more than 5 appeal IDs in the request. We don't want to spin up more than 5 threads per request

app/controllers/appeals_controller.rb
```diff
@@ -34,12 +34,14 @@ def show_case_list
...
+      document_counts_by_id: DocumentCountsByAppealIdHash.new(
+        hash: {}, appeal_ids: params[:appeal_ids].split(",")
+      ).call
+    }
+  rescue Caseflow::Error::EfolderAccessForbidden => error
```
aroltsch: this error should be caught on the per appeal basis as well

app/controllers/appeals_controller.rb
```diff
@@ -34,12 +34,14 @@ def show_case_list
...
-  def document_count
-    render json: { document_count: appeal.number_of_documents }
-  rescue Caseflow::Error::EfolderAccessForbidden => e
-    render(e.serialize_response)
-  rescue StandardError => e
-    handle_non_critical_error("document_count", e)
+  def document_counts_by_id
+    render json: {
+      document_counts_by_id: DocumentCountsByAppealIdHash.new(
+        hash: {}, appeal_ids: params[:appeal_ids].split(",")
```
aroltsch: You don't need to pass hash here.

app/controllers/appeals_controller.rb
```diff
@@ -34,12 +34,18 @@ def show_case_list
...
   end
 
-  def document_count
-    render json: { document_count: appeal.number_of_documents }
-  rescue Caseflow::Error::EfolderAccessForbidden => e
-    render(e.serialize_response)
-  rescue StandardError => e
-    handle_non_critical_error("document_count", e)
+  def document_counts_by_id
+    appeal_ids = params[:appeal_ids].split(",")
```
aroltsch: @youngfreezy, let's move this logic to the workflow class. Sorry about that. I should have mentioned that earlier.

aroltsch: (the error message logic I mean)

app/workflows/document_counts_by_appeal_id.rb
```diff
@@ -0,0 +1,71 @@
...
+      rescue Caseflow::Error::EfolderAccessForbidden => err
+        handle_document_count_error(err, document_counts_by_id_hash, appeal_id)
+        next
+      end
+    end
+  end
+
+  def handle_document_count_error(err, document_counts_by_id_hash, appeal_id)
+    code = (err.class == ActiveRecord::RecordNotFound) ? 404 : 500
+    document_counts_by_id_hash[appeal_id] = {
```
aroltsch: You already have `hash` object that is accessible in the entire, you could use that instead of passing  document_counts_by_id_hash

app/workflows/document_counts_by_appeal_id.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+class DocumentCountsByAppealId
+  def initialize(appeal_ids:)
+    @hash = {}
```
aroltsch: We should have a better name for this variable.

app/workflows/document_counts_by_appeal_id.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+class DocumentCountsByAppealId
+  def initialize(appeal_ids:)
+    @document_counts_by_appeal_id_hash = {}
+    @appeal_ids = appeal_ids
+    @max_batch_size = 5
```
aroltsch: Let's make it into a constant.

app/workflows/document_counts_by_appeal_id.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  # end
+
+  def build_document_counts_hash(appeal_ids)
+    # Collect appeal objects sequentially so we don't exhaust DB pool
+    appeals = collect_appeal_objects_sequentially(appeal_ids)
+    # Spin up a new thread of each appeal and then call join on each thread
+    # Creating threads without calling join on them will cause the main thread
+    # to continue without waiting and possibly exit before the child threads have finished
+    create_thread_for_each_appeal(appeals).map(&:join)
```
aroltsch: I think you don't need this join since you are calling join on line 44.

app/workflows/document_counts_by_appeal_id.rb
```diff
@@ -0,0 +1,70 @@
...
+  private
+
+  def build_document_counts_hash(appeal_ids)
+    # Collect appeal objects sequentially so we don't exhaust DB pool
+    appeals = collect_appeal_objects_sequentially(appeal_ids)
+    # Spin up a new thread of each appeal and then call join on each thread
+    # Creating threads without calling join on them will cause the main thread
+    # to continue without waiting and possibly exit before the child threads have finished
+    create_thread_for_each_appeal(appeals)
+    @document_counts_by_appeal_id_hash
```
aroltsch: Do you need to return this hash here?

app/workflows/document_counts_by_appeal_id.rb
```diff
@@ -0,0 +1,66 @@
...
+      fail Caseflow::Error::TooManyAppealIds
+    end
+
+    build_document_counts_hash(@appeal_ids)
+    @document_counts_by_appeal_id_hash
+  end
+
+  private
+
+  def build_document_counts_hash(appeal_ids)
```
aroltsch: By the way, appeal_ids are accessible within the class so don't need to pass this variable around.

Issue #10233, opened 2019-04-01
Add canceled_at column to asyncable module


pkarman:
Some asyncable jobs will never successfully finish. Rather than delete them, add a `canceled_at` column that can be set with a `cancel!` method. The normal per-class column name overrides should be supported.

pkarman:
Validation steps:

* On the jobs page, pick one of the oldest jobs. Hover over the "Restart" button to find out its ID value.
* Via the Rails console, find the record and call the `.canceled!` method on it.
* Reload the jobs page and verify the job no longer appears there.

Sjones352:
@pkarman I reloaded the page and the job still exists

![Screen Shot 2019-04-29 at 5.44.47 PM.png](https://images.zenhubusercontent.com/5b8fddd1b2cc7b6c78ece9f9/c5912230-d56c-4662-bc52-a2d99c077109)

leikkisa:
I confirmed Sandra's findings, and it was due to the way "or" was being used in the query, PR out to fix.

Sjones352:
This is now passing UAT Validation

![Screen Shot 2019-05-08 at 12.53.06 PM.png](https://images.zenhubusercontent.com/5b8fddd1b2cc7b6c78ece9f9/0703e28d-4c26-4683-9d4f-bb211f5c5b51)

![Screen Shot 2019-05-08 at 12.54.36 PM.png](https://images.zenhubusercontent.com/5b8fddd1b2cc7b6c78ece9f9/a2b785e1-84df-48ad-a2a0-d7804e2291a7)

Reloaded the jobs page and verified that the job no longer appears there.
https://appeals.cf.uat.ds.va.gov/jobs?page=1

PR #10435, opened 2019-04-12
Add asyncable canceled_at columns

pkarman:
connects #10233 

### Description
Some jobs will never finish processing. Marking them canceled should hide them from the /jobs page.

*Schema Changes Only*

* [x] Column comments updated


app/models/concerns/decision_syncable.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module DecisionSyncable
```
leikkisa: 👍

PR #10440, opened 2019-04-15
Remove can_access_v2 BGS feature toggle

### LINKED ISSUE #10439: Remove can_access_v2 BGS feature flag
This feature has been toggled on for all users [since July 2018](https://github.com/department-of-veterans-affairs/appeals-deployment/pull/1592/files), let's remove the code that checks for this feature toggle and instead just default to it being on.

lowellrex:
Resolves #10439.

This feature has been toggled on for all users [since July 2018](https://github.com/department-of-veterans-affairs/appeals-deployment/pull/1592/files). This PR replaces  the call to `FeatureToggle` which will always return true to the boolean `true`.

### DIFF
--- a/app/services/external_api/bgs_service.rb
+++ b/app/services/external_api/bgs_service.rb
@@ -182,7 +182,7 @@ class ExternalApi::BGSService
       MetricsService.record("BGS: can_access? (find_by_file_number): #{vbms_id}",
                             service: :bgs,
                             name: "can_access?") do
-        client.can_access?(vbms_id, FeatureToggle.enabled?(:can_access_v2, user: current_user))
+        client.can_access?(vbms_id, true)
       end
     end
   end


app/services/external_api/bgs_service.rb
```diff
@@ -182,7 +182,7 @@ def can_access?(vbms_id)
       MetricsService.record("BGS: can_access? (find_by_file_number): #{vbms_id}",
                             service: :bgs,
                             name: "can_access?") do
-        client.can_access?(vbms_id, FeatureToggle.enabled?(:can_access_v2, user: current_user))
+        client.can_access?(vbms_id, true)
```
aroltsch: Should we perhaps remove this logic then https://github.com/department-of-veterans-affairs/ruby-bgs/blob/2d0e08ea4157b725242777a6f876fc199f503b61/lib/bgs/services.rb#L73?

lowellrex: Yup! I imagine that will have a to be a second PR (as well as any changes we would have to make for [efolder, monitor, and appeals-data](https://github.com/department-of-veterans-affairs/ruby-bgs/network/dependents)), but I think that makes sense to make this change on the Caseflow side first for clarity's sake.

aroltsch: Makes sense sir!

PR #10441, opened 2019-04-15
Judge view of hearing schedule

### LINKED ISSUE #10409: Hearing Prep Merge | View schedule
Create judge view for viewing the schedule.

sharonwarner:
Resolves #10409 

### Description
This PR creates a judge view of hearing schedule. Judges are able to see hearing days if either the hearing day is assigned to them or the hearing day contains any hearings that are assigned to them.

### DIFF
--- a/app/models/hearing_day.rb
+++ b/app/models/hearing_day.rb
@@ -124,23 +124,17 @@ class HearingDay < ApplicationRecord
     end
 
     def upcoming_days_for_judge(start_date, end_date, user)
-      caseflow_assigned_hearing_days_and_hearings = HearingDay.includes(:hearings)
-        .where("DATE(scheduled_for) between ? and ?", start_date, end_date).select do |hearing_day|
-        hearing_day.judge == user || hearing_day.hearings.any? { |hearing| hearing.judge == user }
-      end
-
-      remaining_hearing_days = HearingDay.where("DATE(scheduled_for) between ? and ?", start_date, end_date)
-        .where.not(id: caseflow_assigned_hearing_days_and_hearings.pluck(:id)).order(:scheduled_for).limit(1000)
-
-      vacols_hearings_for_remaining_hearing_days = HearingRepository.fetch_hearings_for_parents_assigned_to_judge(
-        remaining_hearing_days.pluck(:id), user
+      hearing_days_in_range = HearingDay.includes(:hearings)
+        .where("DATE(scheduled_for) between ? and ?", start_date, end_date)
+      vacols_hearings = HearingRepository.fetch_hearings_for_parents_assigned_to_judge(
+        hearing_days_in_range.first(1000).pluck(:id), user
       )
 
-      hearing_days_with_vacols_hearings = remaining_hearing_days.reject do |hearing_day|
-        vacols_hearings_for_remaining_hearing_days[hearing_day.id.to_s].nil?
+      hearing_days_in_range.select do |hearing_day|
+        hearing_day.judge == user ||
+          hearing_day.hearings.any? { |hearing| hearing.judge == user } ||
+          !vacols_hearings[hearing_day.id.to_s].nil?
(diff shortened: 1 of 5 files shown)


app/models/hearing_day.rb
```diff
@@ -123,6 +123,26 @@ def update_schedule(updated_hearings)
...
 
+    def upcoming_days_for_judge(start_date, end_date, user)
+      caseflow_assigned_hearing_days_and_hearings = HearingDay.includes(:hearings)
+        .where("DATE(scheduled_for) between ? and ?", start_date, end_date).select do |hearing_day|
+        hearing_day.judge == user || hearing_day.hearings.any? { |hearing| hearing.judge == user }
+      end
+
+      remaining_hearing_days = HearingDay.where("DATE(scheduled_for) between ? and ?", start_date, end_date)
+        .where.not(id: caseflow_assigned_hearing_days_and_hearings.pluck(:id)).order(:scheduled_for).limit(1000)
+
```
lomaxap: do we need to exclude `caseflow_assigned_hearing_days_and_hearings`? is this so we have a unique list of days when you combine?

app/models/hearing_day.rb
```diff
@@ -123,6 +123,26 @@ def update_schedule(updated_hearings)
...
+      vacols_hearings_for_remaining_hearing_days = HearingRepository.fetch_hearings_for_parents_assigned_to_judge(
+        remaining_hearing_days.pluck(:id), user
+      )
+
+      hearing_days_with_vacols_hearings = remaining_hearing_days.reject do |hearing_day|
+        vacols_hearings_for_remaining_hearing_days[hearing_day.id.to_s].nil?
+      end
+
+      caseflow_assigned_hearing_days_and_hearings + hearing_days_with_vacols_hearings
```
lomaxap: this method was hard to follow for me. Is there are way to preload vacols hearings in the first query?
something like:
```ruby
hearing_days = HearingDay.where("DATE(scheduled_for) between ? and ?", start_date, end_date).includes(:hearings)
vacols_hearing_days = HearingRepository.fetch_hearings_for_parents(hearing_days.pluck(:id))
hearing_days_for_judge = hearing_days.select do |hearing_day|
   hearing_day.judge == user ||
   hearing_day.hearings.any? { |hearing| hearing.judge == user } ||
   vacols_hearing_days[hearing_day.id]&.any? { |hearing| some_user_id_comparison_for_vacols_hearing }
end
```

if something like that is possible, you could also get rid of the 2 new methods below

app/models/hearing_day.rb
```diff
@@ -123,6 +123,26 @@ def update_schedule(updated_hearings)
       end
     end
 
+    def upcoming_days_for_judge(start_date, end_date, user)
+      caseflow_assigned_hearing_days_and_hearings = HearingDay.includes(:hearings)
+        .where("DATE(scheduled_for) between ? and ?", start_date, end_date).select do |hearing_day|
+        hearing_day.judge == user || hearing_day.hearings.any? { |hearing| hearing.judge == user }
```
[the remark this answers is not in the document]
sharonwarner: Hearing.judge can't be nil if the HearingDay is not nil. The only time Hearing.judge is different from HearingDay.judge is if a hearing coordinator has overwritten the judge at the hearing level. They do not have the option to set the judge to nil.

app/models/hearing_day.rb
sharonwarner: I did this for performance. There's no need to pull hearings from VACOLS for hearing days we're already planning to display.

app/models/hearing_day.rb
sharonwarner: Yeah definitely agree this is cleaner. I guess it isn't significantly faster to pull ~30 fewer hearing days when we already have to pull a thousand or so. I'll update this.

Issue #10227, opened 2019-04-01
Reduce UpdateAppellantRepresentationJob slack messages


lowellrex:
We [recently changed](https://github.com/department-of-veterans-affairs/appeals-deployment/pull/2016) the `UpdateAppellantRepresentationJob` to run once an hour instead of once a day. As a result we are now sending 24 messages to [the #appeals-app-alerts slack  channel](https://dsva.slack.com/archives/C4JECDLSE/p1554009518116900) every day instead of 1.

![image](https://user-images.githubusercontent.com/32683958/55331482-5bc97100-5461-11e9-9a0d-18fbf909966f.png)

The information in these messages is useful but does not require any immediate action. We should instead send a message to slack only when that message requires some action (or maybe when that message will inform us as to why another error happened?) and move to other monitoring functions this message is performing to Datadog [like we do with dispatch](https://app.datadoghq.com/monitors/4212136).

## Acceptance criteria
* [x] Only send a message to slack if `UpdateAppellantRepresentationJob` encountered an error (#10443)
* [x] Create a Datadog monitor that ensures this job is running successfully

lowellrex:
Added datadog monitors to alert if the [runtime of the job is more than 30 minutes](https://app.datadoghq.com/monitors/9547279) and if [no tasks were modified by the job in the past 24 hours](https://app.datadoghq.com/monitors/9547287).

PR #10443, opened 2019-04-15
Reduce UpdateAppellantRepresentationJob slack messages

lowellrex:
Connects #10227.

This PR updates the hourly `UpdateAppellantRepresentationJob` to send data directly to DataDog and only sends messages to slack if it failed completely.

Once we are sending data to DataDog we will set up monitors and alerts around those monitors. We aren't currently taking any action on the successful slack messages we receive from this job so I don't think there is much risk to there being a delay between this change being released live and the DataDog monitors being created (though ticket #10227 will not be complete until those monitors are created). Happy to entertain differing opinions though!

### DIFF
--- a/app/jobs/update_appellant_representation_job.rb
+++ b/app/jobs/update_appellant_representation_job.rb
@@ -7,13 +7,12 @@ class UpdateAppellantRepresentationJob < CaseflowJob
   include ActionView::Helpers::DateHelper
   queue_as :low_priority
 
+  APP_NAME = "caseflow_job"
+  METRIC_GROUP_NAME = self.class.name.underscore
   TOTAL_NUMBER_OF_APPEALS_TO_UPDATE = 1000
 
   def perform
     start_time = Time.zone.now
-    new_task_count = 0
-    closed_task_count = 0
-    error_count = 0
 
     # Set user to system_user to avoid sensitivity errors
     RequestStore.store[:current_user] = User.system_user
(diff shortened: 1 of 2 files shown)


app/jobs/update_appellant_representation_job.rb
```diff
@@ -78,16 +77,40 @@ def active_appeals
...
-          " Created #{new_task_count} new tracking tasks and closed #{closed_task_count} existing tracking tasks." \
-          " Encountered errors for #{error_count} individual appeals."
+    msg = "UpdateAppellantRepresentationJob failed after running for #{duration}. Fatal error: #{err.message}"
 
-    msg += " Fatal error: #{err.message}" if err
     Rails.logger.info(msg)
-    Rails.logger.info(err.backtrace.join("\n")) if err
+    Rails.logger.info(err.backtrace.join("\n"))
+
     slack_service.send_notification(msg)
```
tomas-nava: We're not verifying in tests that this slack message is sent?

spec/jobs/update_appellant_representation_job_spec.rb
```diff
@@ -34,12 +36,8 @@
...
 
     it "runs the job as expected" do
-      expect_any_instance_of(UpdateAppellantRepresentationJob).to receive(:log_info).with(
-        anything,
-        new_task_count,
-        closed_task_count,
-        error_count
-      )
+      expect_any_instance_of(UpdateAppellantRepresentationJob).to receive(:record_runtime).exactly(1).times
+      expect_any_instance_of(UpdateAppellantRepresentationJob).to_not receive(:log_error).with(anything, anything)
```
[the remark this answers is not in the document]
lowellrex: I think so, yes.

app/jobs/update_appellant_representation_job.rb
lowellrex: Great catch! Adding the test back in!

PR #10450, opened 2019-04-15
[WIP] Enable bullet in tests to see which ones fail

monfresh:

### Description

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...

*Schema Changes Only*

* [ ] Column comments updated

### DIFF
--- a/app/models/concerns/issue_updater.rb
+++ b/app/models/concerns/issue_updater.rb
@@ -92,7 +92,7 @@ module IssueUpdater
   end
 
   def fail_if_no_remand_reasons!(issue, remand_reasons_attrs)
-    if issue.disposition == "remanded" && remand_reasons_attrs.blank?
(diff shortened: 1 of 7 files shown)


pkarman:
thoughts on this @monfresh ?

monfresh:
Ideally, we would fix all the failing specs so we can start enabling it and detecting new N+1 queries. I haven't had time to do this. Maybe when I'm on bat team in 2 weeks, or maybe something the bat team can tackle a bit at a time?

monfresh:
After working on this some more, I'm realizing that this might not be worth the effort. Sometimes there are false positives, and in some cases, it's hard to tell if adding `includes` makes a difference. In some cases, it resulted in more DB queries. Since tests don't use as much data as production, it's not clear if there's a benefit. I think a better approach might be to monitor New Relic for slow queries and controllers, and address them on a case by case basis.

Issue #9527, opened 2019-02-19
Give private attorneys access to Queue


lowellrex:
This tech spec outlines one path to provide private attorneys with access to an "All cases" queue table view and Case Details. We do not expect Private Attorneys to do any tasks in Caseflow at this time.

### PrivateBar as Vso subclass a la JudgeTeam

This approach will allow us to take advantage of code we have already written for VSO-related organization controls (as well as BGS for access control) in a pattern that we have already established with minimal changes to existing code. One disadvantage to this approach is that creating an organization for a single person to give us the benefit of access controls is a somewhat imprecise model and indirect mapping of concepts into code.

A quick sketch of how this would look:

```ruby
class PrivateBar < Vso
  def self.create_for_user(user, ptcpt_id)
    create!(name: user.css_id, url: user.css_id.downcase, participant_id: ptcpt_id).tap do |org|
      OrganizationsUser.make_user_admin(user, org)
      VsoConfig.create!(organization: org, ihp_dockets: [])
    end
  end

  def user_has_access?(user)
    user_is_admin?(user)
  end

  def can_receive_task?(_task)
    false
  end
end
```

And we could create these organizations for private attorneys by running the following commands:

```ruby
rails c> u = User.find_by(css_id: "PRIVATE_ATTORNEY")
rails c> PrivateBar.create_for_user(u, "1234567")
```

`PrivateBar` being a subclass of `Vso` lets us take advantage of [the `Appeal.vsos` method](https://github.com/department-of-veterans-affairs/caseflow/blob/68f72a296916a0c3dc031ad2663baab4e727e016/app/models/appeal.rb#L306) that we use [to create `TrackVeteranTask`s when we create the `RootTask`](https://github.com/department-of-veterans-affairs/caseflow/blob/68f72a296916a0c3dc031ad2663baab4e727e016/app/models/tasks/root_task.rb#L69) for an appeal. With this design we would not have to change any code in the `Appeal` or `RootTask` models in order to start associating `TrackVeteranTask`s with private attorneys.

[...]

PR #10454, opened 2019-04-15
Create PrivateBar class

lowellrex:
Connects #9527 (part 3 of 4).

Creates `PrivateBar` class to give private attorneys access to queue similar to what field VSOs have.


lowellrex:
Closing in favor of #10960.

Issue #10442, opened 2019-04-15
Do not require claimant address for non-comp reviews


pkarman:
Do not validate the claimant's address or send the "claimant_address_required" error to the front end if the review is non-compensation.

Acceptance criteria
For a decision review that is non-compensation with a Veteran or non-Veteran claimant that does not have an address, the user should be able to complete the review without errors.

Related PR that implemented this:
#10091

If the intake /review PATCH request returns 422 (unprocessable) for `claimant_address_required` the error message is not visible to the user when the claimant is the Veteran and it just looks like the page is "stuck".

Here's the console log:

<img width="592" alt="Screen Shot 2019-04-15 at 9 43 06 AM" src="https://user-images.githubusercontent.com/1205061/56141886-4ffea400-5f63-11e9-9300-ab0e35c3a7ce.png">

leikkisa:
This should also put an error message on the users screen saying that an address is required for the claimant. However, the reason for this is that a claim establishment fails if the claimant doesn't have an address, so this may not be required for intakes processed in Caseflow.

I would also like to find out if the Veteran was the claimant in this instance.

leikkisa:
Here's a ticket for this:
https://github.com/department-of-veterans-affairs/caseflow/issues/10449

Sjones352:
**Passes UAT user does not have address**

**Non-comp**
![Screen Shot 2019-04-19 at 2.33.07 PM.png](https://images.zenhubusercontent.com/5b8fddd1b2cc7b6c78ece9f9/6f7decc4-25fd-4434-afea-86fa8154860a)

**Appeal**
![Screen Shot 2019-04-19 at 2.39.56 PM.png](https://images.zenhubusercontent.com/5b8fddd1b2cc7b6c78ece9f9/83d5d8da-e15a-47b7-8c05-b59de55bbafc)

PR #10458, opened 2019-04-16
Do not require address for intakes processed_in_caseflow

leikkisa:
connects #10442


app/models/validators/claimant_validator.rb
```diff
@@ -32,6 +32,8 @@ def validate_participant_id(claimant)
   end
 
   def validate_claimant_address(claimant)
+    return if claimant.decision_review.is_a?(Appeal)
+    return unless benefit_type_requires_payee_code?(claimant)
```
pkarman: in the end, such a simple fix!

pkarman:
looks like some tests need updating.

PR #10459, opened 2019-04-16
Use regional office from session when setting timezone

### LINKED ISSUE #10255: ActiveRecord::ConnectionTimeoutError
This Sentry alert has gone off a few times in the past 24hr: https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/4040/. The alert says, "could not obtain a connection from the pool within 5.000 seconds (waited 5.000 seconds); all pooled connections were in use". Looks like the request to look up a user by css id and station id is timing out.

aroltsch:
Resolves #10255 

### Description
1. When setting a timezone, do not query a user record each time
2. Store user's DB ID in session to make lookups faster

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...

*Schema Changes Only*

* [ ] Column comments updated

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -244,11 +244,8 @@ class ApplicationController < ApplicationBaseController
   end
 
   def set_timezone
-    Time.zone = timezone_based_on_ro || "America/Chicago"
-  end
-
-  def timezone_based_on_ro
-    (RegionalOffice::CITIES[session[:regional_office]] || {})[:timezone] if session[:regional_office]
+    Time.zone = session[:timezone] || current_user&.timezone
+    session[:timezone] ||= current_user&.timezone
   end
 
   # This is used in development mode to:
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -244,7 +244,9 @@ class ApplicationController < ApplicationBaseController
   end
 
   def set_timezone
-    Time.zone = current_user.timezone if current_user
+    if session[:regional_office]
+      Time.zone = (RegionalOffice::CITIES[session[:regional_office]] || {})[:timezone] || "America/Chicago"
+    end 
   end
 
   # This is used in development mode to:
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -246,7 +246,7 @@ class ApplicationController < ApplicationBaseController
   def set_timezone
     if session[:regional_office]
       Time.zone = (RegionalOffice::CITIES[session[:regional_office]] || {})[:timezone] || "America/Chicago"
-    end 
+    end
   end
 
   # This is used in development mode to:
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -244,9 +244,11 @@ class ApplicationController < ApplicationBaseController
   end
 
   def set_timezone
-    if session[:regional_office]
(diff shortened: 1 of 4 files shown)


app/models/user.rb
```diff
@@ -311,9 +307,10 @@ def from_session(session)
 
       return nil if user_session.nil?
 
+      pg_user_id = session["pg_user_id"]
       css_id = user_session["id"]
       station_id = user_session["station_id"]
-      user = find_by_css_id(css_id)
+      user = pg_user_id ? find_by(id: pg_user_id) : find_by_css_id(css_id)
```
aroltsch: searching by ID is faster than by css ID

app/controllers/application_controller.rb
```diff
@@ -244,7 +244,9 @@ def set_raven_user
   end
 
   def set_timezone
-    Time.zone = current_user.timezone if current_user
+    if session[:regional_office]
+      Time.zone = (RegionalOffice::CITIES[session[:regional_office]] || {})[:timezone] || "America/Chicago"
```
aroltsch: Don't need to query user to set timezone.

app/controllers/application_controller.rb
```diff
@@ -244,7 +244,11 @@ def set_raven_user
   end
 
   def set_timezone
-    Time.zone = current_user.timezone if current_user
+    Time.zone = timezone_based_on_ro || "America/Chicago"
```
aroltsch: No need to query user to get regional office, we can get it from session

app/controllers/application_controller.rb
```diff
@@ -244,7 +244,11 @@ def set_raven_user
...
+  end
+
+  def timezone_based_on_ro
+    (RegionalOffice::CITIES[session[:regional_office]] || {})[:timezone] if session[:regional_office]
```
lowellrex: Where do we put the regional_office into the session?

aroltsch: I thought it was here https://github.com/department-of-veterans-affairs/caseflow/blob/9031ac194c3cabf8bcfd9c571f60566c28e4ffac/app/controllers/sessions_controller.rb#L45

aroltsch: and here https://github.com/department-of-veterans-affairs/caseflow/blob/9031ac194c3cabf8bcfd9c571f60566c28e4ffac/app/controllers/sessions_controller.rb#L70

aroltsch: However, now I am looking and it might be only happening when a user needs to select a regional office

aroltsch: I saw here that session has a regional_office https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/user.rb#L323

app/controllers/application_controller.rb
```diff
@@ -244,7 +244,8 @@ def set_raven_user
   end
 
   def set_timezone
-    Time.zone = current_user.timezone if current_user
+    Time.zone = session[:timezone] || current_user&.timezone
+    session[:timezone] ||= current_user&.timezone
```
lowellrex: Could be rewritten `Time.zone = session[:timezone] ||= current_user&.timezone` but then again maybe not because that syntax is a little tough to parse.

app/models/user.rb
pkarman: I am surprised that search by css ID would be noticably slower since there should be an index on it.

also could write `find(pg_user_id)` to be a tiny bit more terse.

PR #10461, opened 2019-04-16
Intake | withdraw feature toggle

Sjones352:
Resolves # https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/4634/

### DIFF
--- a/app/models/request_issues_update.rb
+++ b/app/models/request_issues_update.rb
@@ -73,7 +73,7 @@ class RequestIssuesUpdate < ApplicationRecord
   end
 
   def removed_or_withdrawn_issues
-    return [] unless allow_zero_request_issues? && withdrawn_issues?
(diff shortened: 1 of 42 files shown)


app/models/request_issues_update.rb
```diff
@@ -73,6 +73,8 @@ def removed_issues
   end
 
   def removed_or_withdrawn_issues
+    return [] unless allow_zero_request_issues? && withdrawn_issues?
```
pkarman: maybe this should be:

```ruby
return removed_issues unless withdrawn_issues?
```

instead.

pkarman: and could we rename the method to `allow_withdrawn_issues?` instead to differentiate it from what could be confused with whether `withdrawn_issues` is defined or not.

PR #10462, opened 2019-04-16
fix

youngfreezyVA:
https://dsva.slack.com/archives/C6E41RE92/p1555427772071300

### DIFF
--- a/app/workflows/tasks_for_appeal.rb
+++ b/app/workflows/tasks_for_appeal.rb
@@ -30,7 +30,7 @@ class TasksForAppeal
   def tasks_assigned_to_user_or_any_other_vso_employee
     appeal.tasks
       .includes(*task_includes)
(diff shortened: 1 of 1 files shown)


app/workflows/tasks_for_appeal.rb
```diff
@@ -30,7 +30,7 @@ def call
   def tasks_assigned_to_user_or_any_other_vso_employee
     appeal.tasks
       .includes(*task_includes)
-      .select { |task| task.assigned_to.is_a?(Representative) || current_user == task.assigned_to }
+      .select { |task| task.assigned_to.is_a?(Representative) || @user == task&.assigned_to }
```
aroltsch: you should be able to call `user` without an `@` sign

aroltsch: also, you don't need `&` sign after task because assigned_to should always be present

PR #10465, opened 2019-04-16
Daily Docket Judge View

### LINKED ISSUE #10410: Hearing Prep Merge | Daily docket
AC:
- [ ] Update previously scheduled hearings such that it pulls hearings that were postponed or canceled *and their dispositions are uneditable*.
- [ ] Add judge views for AMA and legacy hearings

Note that we do not currently have functionality for AOD/AOD reason for AMA hearings. That is included in a separate ticket: #8682 For now, let's just include read only dropdowns for those two fields.

AMA hearing:
Screen Shot 2019-04-11 at 3 15 28 PM

Legacy hearing:

https://va-gov.invisionapp.com/share/PMR1ID7XSHT#/screens/353825993

lomaxap:
resolves #10410

- adds AOD, AOD reason, prepped, and VLJ worksheet fields to daily docket
- limits view of those fields with users in `JudgeTeam` org.
- Moves all inputs and display text to `DailyDocketRowInputs.jsx`

this PR also includes an opinionated refactor of the daily docket updates redux.

Previously, each input field for a daily docket row had its own action/reducer associated with it.
For example, updating the `notes` columns includes:

an action:
```javascript
export const onHearingNotesUpdate = (hearingId, notes) => ({
  type: ACTIONS.HEARING_NOTES_UPDATE,
  payload: {
    hearingId,
    notes
  }
});
```
and a reducer
```javascript
  case ACTIONS.HEARING_NOTES_UPDATE:
    return update(state, {
      hearings: {
        [action.payload.hearingId]: {
          editedNotes: { $set: action.payload.notes },
          edited: { $set: true }
        }
      }
    });
```

This PR creates one action/reducer to update all fields
```javascript
  case ACTIONS.UPDATE_DOCKET_HEARING:
    return update(state, {
      hearings: {
        [action.payload.hearingId]: {
          $set: {
            ...state.hearings[action.payload.hearingId],
            ...action.payload.values
          }
        }
      }
    });
```

I think this reduces the overhead of adding/removing/maintaining inputs, and I can't currently think of any drawbacks to rolling all updates into a single function. If there is logic that needs to happen when a particular property is updated, we can use `componentDidUpdate` and compare current and previous props.

I'm curious how this approach compares to how other apps are using redux so may be worth checking with other teams.


app/views/hearing_schedule/index.html.erb
```diff
@@ -8,11 +8,12 @@
         !current_user.can?('Build HearSched') && !current_user.can?('Edit HearSched'),
     userRoleVso: current_user.can?('VSO') && !current_user.can?('RO ViewHearSched') &&
         !current_user.can?('Build HearSched') && !current_user.can?('Edit HearSched'),
+    userInJudgeTeam: !JudgeTeam.for_judge(current_user).nil?,
```
sharonwarner: Thinking about this more, I think we should continue to use the "Hearing Prep" CSEM role to determine if the user should see the judge view. That matches the current functionality much better than using teams. I'm worried that the JudgeTeam might not match the "Hearing Prep" users and we'll spend a lot of time on support tickets adding users to the JudgeTeam. 

Also, in the show method in the hearing_day_controller, we should restrict the hearings returned to only include hearings assigned to the current_user. (I know that method is messy, don't feel like you have to clean it up... we're going to clean it up when we remove VACOLS hearing day functionality.)

client/app/hearingSchedule/reducers.js
```diff
@@ -68,6 +68,17 @@ const hearingScheduleReducer = (state = initialState, action = {}) => {
     return update(state, {
       $unset: ['saveSuccessful', 'displayLockSuccessMessage']
     });
+  case ACTIONS.UPDATE_DOCKET_HEARING:
```
sharonwarner: I don't feel strongly about this, but I agree we should reach out to other teams so we can come up with a consistent practice across applications.

Also, if we decide to go this route, then let's remove the old actions/constants/reducers.

spec/feature/hearing_schedule/daily_docket_spec.rb
```diff
@@ -129,4 +129,22 @@
...
   end
+
+  context "Daily docket for Judge user" do
+    let!(:current_user) { User.authenticate!(css_id: "BVATWARNER", roles: ["Hearing Prep"]) }
+    let!(:hearing) { create(:hearing, :with_tasks) }
+
+    scenario "User has hearing prep fields" do
+      visit "hearings/schedule/docket/" + hearing.hearing_day.id.to_s
+      find(".dropdown-aod")
+      find(".dropdown-aodReason")
```
tomas-nava: Could we express these as expectations instead? Like:

```ruby
expect(page).to have_css(".dropdown-aod")
```

PR #10466, opened 2019-04-16
When appellant representation changes, cancel/duplicate tasks assigned to previous representative (VSO).

### LINKED ISSUE #10214: Cancel all tasks assigned to old VSO when appellant representation changes
In [`TrackVeteranTask.sync_tracking_tasks`](https://github.com/department-of-veterans-affairs/caseflow/blob/50a83ce0f8334f639d5caf3dcdffdce5b258a2b2/app/models/tasks/track_veteran_task.rb#L33) we open and close tracking tasks when appellant representation changes (by way of [the `UpdateAppellantRepresentationJob`](https://github.com/department-of-veterans-affairs/caseflow/blob/50a83ce0f8334f639d5caf3dcdffdce5b258a2b2/app/jobs/update_appellant_representation_job.rb#L25)). However we have not been doing anything for other types of tasks that are assigned to VSO (or the descendants of those tasks).

## Acceptance criteria
* [ ] Create IHP tasks assigned to previous VSO when appellant representation changes
* [ ] Cancel IHP tasks assigned to previous VSO and child tasks when appellant representation changes

kevmo:
Resolves #10214

### Description
Updated business logic in `app/models/tasks/track_veteran_task.rb` to look for active tasks by previous VSO representatives.  The `sync_tracking_tasks` method now finds these tasks, cancels them, then duplicates their attributes for a newly created task, which is assigned to the new representative.

Updated `spec` file as well ➡️  Taking suggestions for additional tests to add. 

### Acceptance Criteria
[] Create IHP tasks assigned to previous VSO when appellant representation changes
[] Cancel IHP tasks assigned to previous VSO and child tasks when appellant representation changes

### Testing Plan
`bundle exec rspec spec/models/tasks/track_veteran_task_spec.rb`


kevmo:
@lowellrex The linters are barking bc `TrackVeteranTask.sync_tracking_tasks` has become a 31 LOC monster. You said you already had a refactor in mind?

app/models/tasks/track_veteran_task.rb
```diff
@@ -33,6 +33,7 @@ def hide_from_task_snapshot
   def self.sync_tracking_tasks(appeal)
     new_task_count = 0
     closed_task_count = 0
+    copied_task_count = 0
 
     active_tracking_tasks = appeal.tasks.active.where(type: TrackVeteranTask.name)
```
lowellrex: I think we need to include `InformalHearingPresentationTask`s here as well. Maybe something like:

```ruby
# Maybe we could rename this variable to active_vso_tasks.
active_tracking_tasks = appeal.tasks.active.where(type: [TrackVeteranTask.name, InformalHearingPresentationTask.name])
```

app/models/tasks/track_veteran_task.rb
```diff
@@ -33,6 +33,7 @@ def hide_from_task_snapshot
...
     cached_vsos = active_tracking_tasks.map(&:assigned_to)
```
lowellrex: I think we will need to make sure the assignees of the tasks are unique because we are now returning IHP tasks and tracking tasks and a single VSO would have both assigned to them. Something like the following could work:

```ruby
cached_vsos = active_tracking_tasks.map(&:assigned_to).uniq
```

app/models/tasks/track_veteran_task.rb
```diff
@@ -52,7 +53,26 @@ def self.sync_tracking_tasks(appeal)
...
+          if success
+            copied_task_count += 1
+          end
+        end
+
+        t.update!(status: Constants.TASK_STATUSES.cancelled)
+      end
+    end
+
+    [new_task_count, closed_task_count, copied_task_count]
```
lowellrex: I think we can include all of this logic in [the loop that starts on line 44](https://github.com/department-of-veterans-affairs/caseflow/pull/10466/files#diff-a0ca90091cb8eea603b91baff656b76eR44) by doing something like:

```ruby
new_vsos.each do |new_vso|
  params = { appeal: appeal, parent: appeal.root_task, assigned_to: new_vso }
  TrackVeteranTask.create!(**params)
  InformalHearingPresentationTask.create!(**params) if new_vso.should_write_ihp?(appeal)
  ...
```

app/models/tasks/track_veteran_task.rb
kevmo: So I removed that for two reasons
* Broke non-TrackVeteranTasks into their own separate loop/unit of code to make it easier to separate into separate function for refactor
* To generalize the code. #10214 says "Cancel all tasks assigned to old VSO". The code below searches for all tasks associated with the appeal that are assigned to the old VSO and duplicates them. Should this PR be limited just to InformalHearingPresentation tasks? I had assumed that the Acceptance Criteria just recommended that particular task as being good for testing here.

lowellrex: There are only two types of tasks that are currently assigned to VSOs: IHP tasks and tracking tasks. Whenever we add other task types that will be assigned to VSOs I would like to explicitly add them to this logic of opening/closing them if representation changes instead of having it happen automatically for all tasks assigned to VSOs, because I think that behaviour could be a little vexing to try to debug.

spec/models/tasks/track_veteran_task_spec.rb
```diff
@@ -44,10 +44,40 @@
...
+          appeal: root_task.appeal,
+          assigned_to: old_vso
+        )
+      end
+
+      before { allow_any_instance_of(Appeal).to receive(:vsos).and_return([new_vso]) }
+
+      it "cancels all tasks of former VSO" do
+        subject
+        expect(ihp_org_task.reload.status).to eq(Constants.TASK_STATUSES.cancelled)
```
lowellrex: Can we check to confirm that the `new_vso` has to tasks assigned to it before we call `subject` and that the `new_vso` does have tasks assigned to it after we call `subject`?

kevmo: Did it in the next `it` block.

PR #10467, opened 2019-04-17
Only propagate changed fields in HearingDays to child Hearings and LegacyHearings

### LINKED ISSUE #10404: Hearing day updates | Only propagate edited fields to children records
In the hearing day model, we have an after_update hook that runs the following method. This method needs to be updated to only edit children record fields that have been edited at the parent level. For example, if only the room number was edited at the parent level, we only want to edit the room number of the children records. 

```  
def update_children_records
    vacols_hearings.each do |hearing|
      hearing.update_caseflow_and_vacols(
        room: room,
        bva_poc: bva_poc,
        judge_id: judge ? judge.vacols_attorney_id : nil
      )
    end

    hearings.each { |hearing| hearing.update!(room: room, bva_poc: bva_poc, judge: judge) }
  end
```

tomas-nava:
Resolves #10404

### Description
Checks to see whether a change has been made in a `HearingDay` before passing that change to an associated `Hearing` or `LegacyHearing`.


spec/models/hearing_day_spec.rb
```diff
@@ -88,25 +101,71 @@
...
+          subject
+        end
+      end
     end
   end
 
   context "confirm_no_children_records" do
-    let!(:hearing_day) { create(:hearing_day) }
-    let!(:hearing) { create(:hearing, hearing_day: hearing_day) }
+    let!(:hearing_day) { FactoryBot.create(:hearing_day) }
```
sharonwarner: Just curious - why do you prefer `FactoryBot.create` to `create`?

tomas-nava: Partly it's a preference (maybe Lowell's?) that I adopted from working on the queue team, partly I like that it disambiguates where the `create` method's defined. Not that there's a ton of confusion on that front, but I think it makes the tests a little less mysterious for newcomers.

PR #10474, opened 2019-04-17
Only send relevant fields to backend from hp daily docket

sharonwarner:
Resolves https://github.com/department-of-veterans-affairs/appeals-support/issues/5012

We've had several support requests where the hearing date has unexpectedly changed. I believe this is happening from the hearing prep daily docket. We previously sent the entire hearing object to the backend when updating a hearing, so if the time was formatted differently because of a timezone, those changes can get persisted to the database. I updated the hearing prep frontend to only send fields that are editable from the hearing prep daily docket.


sharonwarner:
Yes! I also manually tested all fields for both AMA and legacy hearings.

PR #10483, opened 2019-04-18
Fix impersonation in production

aroltsch:
Set `user_pg_id` when impersonating users in production and locally. Fixes logout.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -61,12 +61,14 @@ class SessionsController < ApplicationController
 
   def remove_user_from_session
     session.delete(:regional_office)
+    session.delete(:user_pg_id)
     session.delete("user")
   end
 
   def add_user_to_session(user_id)
     user = User.find(user_id)
     session["user"] = user.to_session_hash
+    session[:user_pg_id] = user.id
     session[:regional_office] = user.users_regional_office
     RequestStore[:current_user] = user
   end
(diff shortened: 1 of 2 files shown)


app/controllers/sessions_controller.rb
```diff
@@ -61,12 +61,14 @@ def destroy
 
   def remove_user_from_session
     session.delete(:regional_office)
+    session.delete(:user_pg_id)
```
aroltsch: This is for global admins to be able to log out.

app/controllers/test/users_controller.rb
```diff
@@ -83,6 +83,7 @@ def set_user
     User.clear_current_user # for testing only
 
     session["user"] = User.authentication_service.get_user_session(params[:id])
+    session[:pg_user_id] = params[:id]
```
aroltsch: This is for changing users locally.

app/controllers/test/users_controller.rb
```diff
@@ -94,6 +95,7 @@ def log_in_as_user
     return head :not_found if user.nil?
 
     session["user"] = user.to_session_hash
+    session[:pg_user_id] = user.id
```
aroltsch: This is for impersonating in production.

app/controllers/sessions_controller.rb
aroltsch: and it probably fixes the logout

PR #10485, opened 2019-04-18
Anya/10287 cache staff record in redis

### LINKED ISSUE #10287: Oracle connections closed when trying to connect
Sentry alert: https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2769/

Looks like Oracle connections were closed when trying to look up records from VACOLS.

aroltsch:
Resolves #10287 

### Description
An attempt to prevent oracle connection errors from happening by caching staff records in redis for 24 hours


app/models/user.rb
```diff
@@ -295,7 +295,8 @@ def css_ids_by_vlj_ids(vlj_ids)
     def create_judge_in_vacols(first_name, last_name, vlj_id)
       return unless Rails.env.development? || Rails.env.demo?
 
-      UserRepository.create_judge_in_vacols(first_name, last_name, vlj_id)
+      sdomainid = ["BVA", first_name.first, last_name].join
```
aroltsch: Removed this method from UserRepository because code climate was complaining about too many methods. I also think we should not put dev/demo methods in the main files and create a helper for that.

Issue #10432, opened 2019-04-12
Ensure VLJ Support Staff users navigate to their queue when clicking Caseflow logo and not Veteran Case Search page


laurjpeterson:
In VLJ Support Staff refresher/training on 4/11, VLJ Support users flagged that when they click the Caseflow logo, they are taken to the case search homepage instead of their queue. 

Initially, I thought this was because they had the `case_search_home_page` turned on for them because I remember doing this for them last summer. But looking at [features-config.yml](https://github.com/department-of-veterans-affairs/appeals-deployment/edit/master/ansible/vars/features-config.yml), it looks like that feature toggle is on for everyone. 

## AC
- [ ] Investigate why VLJ Support Staff are being taken to Veteran Case Search homepage - use BVAPBROWNING as a user to impersonate
- [ ] Define how this could be fixed, hopefully just by the features.config.

lowellrex:
Looks like some of these folks are not identified as members of the VLJ support staff in VACOLS so [`ApplicationController.case_search_home_page` returns true](https://github.com/department-of-veterans-affairs/caseflow/blob/870e4c4c3c2ca523643d0e68c9f6377f4c7e3801/app/controllers/application_controller.rb#L190).

```ruby
rails c> Colocated.singleton.users.map(&:colocated_in_vacols?)
# [false, false, false, false, true, true, true]
```

We can attack this a few different ways. The quickest would be to have somebody else update VACOLS to make sure these folks are identified as belonging to the VLJ support staff team. A more precise, long-term fix is probably to update the [`User.organization_queue_user?` method](https://github.com/department-of-veterans-affairs/caseflow/blob/870e4c4c3c2ca523643d0e68c9f6377f4c7e3801/app/models/user.rb#L182) to check for membership in any organization instead of the feature flag it currently checks for (and which does not relate to [any activated feature](https://github.com/department-of-veterans-affairs/appeals-deployment/blob/master/ansible/vars/features-config.yml)).

PR #10491, opened 2019-04-19
Update User.organization_queue_user? method

lowellrex:
Connects #10432. VLJ support staff were not being properly identified as "organizational queue users". This PR fixes that.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -180,7 +180,7 @@ class User < ApplicationRecord
   end
 
   def organization_queue_user?
-    FeatureToggle.enabled?(:organization_queue, user: self)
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -180,7 +180,7 @@ def vso_employee?
   end
 
   def organization_queue_user?
-    FeatureToggle.enabled?(:organization_queue, user: self)
+    organizations.any?
```
lowellrex: We never actually ended up [using this feature toggle](https://github.com/department-of-veterans-affairs/caseflow/pull/6839#discussion_r215439916).

Issue #9988, opened 2019-03-14
Intake Edit | Withdrawal confirmation for Caseflow based decision reviews


shanear:
We need to show a confirmation for users that issues were withdrawn.

All withdrawal/removal mocks: https://va-gov.invisionapp.com/share/BRQUJR5J5AE#/screens/351540954

requires https://github.com/department-of-veterans-affairs/caseflow/issues/9985

## Acceptance Criteria
- When a save is processed on the Add/Remove issues screen for a Caseflow based review where one or more issues were withdrawn. There should be a notification that there were issue(s) withdrawn on the notification alert on the case details page.
- If an entire decision review is withdrawn, it should behave the same as the "Review Removed" confirmation. Except it should say "Review Withdrawn" (see: https://github.com/department-of-veterans-affairs/caseflow/issues/9734)

## Mocks
<img width="1169" alt="Screenshot 2019-03-14 17 15 31" src="https://user-images.githubusercontent.com/1596259/54399722-ff9ed880-467c-11e9-955e-9501e662a3e1.png">

## Notes
- VBMS based withdraws will be handled here: #9987

leikkisa:
Let's wait until Carola comes back to ask about this. Except for the "this issue:" part, this mock seems to be for withdrawing the entire review.

carolaponce:
Confirmation: This would show the issue withdrawn not the review title

carolaponce:
@leikkisa This is the updated confirmation format - this is the modular confirmation that gets added to the Reviews needing action page for non-comp LOB or if this confirmation is included in queue.

Since we are changing the add/remove title to Edit issues - the confirmation reflects that and the message is simplified.

![simplified_confirmation](https://user-images.githubusercontent.com/11700545/56155455-c6a19e80-5f6f-11e9-8366-0f02d1241e6d.png)

leikkisa:
Validated in UAT:
Could use some space below it though.

![Screen Shot 2019-05-01 at 5.29.13 PM.png](https://images.zenhubusercontent.com/58f6b6eb51aa2c30577bb60e/ba9b66ff-0c37-4265-9393-e50f1a55623c)

PR #10496, opened 2019-04-22
Withdrawn Alert message for Caseflow decision

Sjones352:
Connects #9988

[ ]  If an entire decision review is withdrawn, it should behave the same as the "Review Removed" confirmation. Except it should say "Review Withdrawn" (see: https://github.com/department-of-veterans-affairs/caseflow/issues/9734)

**Processed in Caseflow. Redirects you to business_line**
Screen Shot 2019-04-22 at 1 06 52 PM

**Appeal redirects you to queue page**
Screen Shot 2019-04-22 at 11 56 45 AM

Screen Shot 2019-04-23 at 2 16 35 PM


app/controllers/appeals_controller.rb
```diff
@@ -169,4 +164,38 @@ def review_removed_message
...
+  def review_withdrawn_message
+    "You have successfully withdrawn a review."
+  end
+
+  # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Metrics/AbcSize
+  def removed_or_withdrawn_issues_success
+    if request_issues_update.after_issues.empty?
+      flash[:removed] = review_removed_message
+      render json: {
```
leikkisa: Since this render json section is the same in all cases, it could go after the removed_or_withdrawn_issues_success in update, instead of being repeated here.

app/controllers/appeals_controller.rb
```diff
@@ -169,4 +164,38 @@ def review_removed_message
...
+        issuesBefore: request_issues_update.before_issues.map(&:ui_hash),
+        issuesAfter: request_issues_update.after_issues.map(&:ui_hash),
+        withdrawnIssues: request_issues_update.withdrawn_issues.map(&:ui_hash)
+      }
+    elsif !request_issues_update.withdrawn_issues.empty?
+      request_issues_update.withdrawn_issues.delete_if(&:withdrawal_date)
```
leikkisa: What is this line here for, deleting if there is a withdrawal date?

app/controllers/appeals_controller.rb
```diff
@@ -169,4 +164,38 @@ def review_removed_message
...
+    else
+      render json: {
+        issuesBefore: request_issues_update.before_issues.map(&:ui_hash),
+        issuesAfter: request_issues_update.after_issues.map(&:ui_hash),
+        withdrawnIssues: nil
+      }
+    end
+  end
+  # rubocop:enable Metrics/AbcSize
+  # rubocop:enable Metrics/MethodLength
```
leikkisa: This will likely not be needed after you move the render json outside of this method

app/controllers/claim_review_controller.rb
```diff
@@ -57,22 +57,52 @@ def verify_access
...
+  # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Metrics/AbcSize
+  def removed_or_withdrawn_issues
+    if request_issues_update.after_issues.empty?
       flash[:removed] = decisions_removed_message
       render json: { redirect_to: claim_review.business_line.tasks_url,
                      issuesBefore: request_issues_update.before_issues.map(&:ui_hash),
-                     issuesAfter: request_issues_update.after_issues.map(&:ui_hash) }
+                     issuesAfter: request_issues_update.after_issues.map(&:ui_hash),
+                     withdrawnIssues: claim_review.business_line.tasks_url }
```
leikkisa: Why does withdrawnIssues contain a URL?

app/controllers/claim_review_controller.rb
```diff
@@ -57,22 +57,52 @@ def verify_access
...
+                     issuesBefore: request_issues_update.before_issues.map(&:ui_hash),
+                     issuesAfter: request_issues_update.after_issues.map(&:ui_hash),
+                     withdrawnIssues: claim_review.business_line.tasks_url }
     else
       render json: {
         redirect_to: nil,
         issuesBefore: request_issues_update.before_issues.map(&:ui_hash),
-        issuesAfter: request_issues_update.after_issues.map(&:ui_hash)
+        issuesAfter: request_issues_update.after_issues.map(&:ui_hash),
+        withdrawnIssues: nil
```
leikkisa: Same feedback as the appeals controller on the render json section.

client/app/nonComp/components/Alerts.jsx
```diff
@@ -46,6 +46,8 @@ export class FlashAlerts extends React.PureComponent {
         flashMsg = <Alert key={idx} title="Error" type="error" >{flash[1]}</Alert>;
       } else if (flash[0] === 'removed') {
         flashMsg = <Alert key={idx} title="Review Removed" type="success">{flash[1]}</Alert>;
+      } else if (flash[0] === 'withdrawn') {
```
leikkisa: We may want to check with Carola is the "Review Removed" is also being updated to "Edit Completed" as the title. That was the impression that I got from her.

app/controllers/appeals_controller.rb
```diff
@@ -169,4 +164,38 @@ def review_removed_message
...
+  # rubocop:disable Metrics/AbcSize
+  def removed_or_withdrawn_issues_success
+    if request_issues_update.after_issues.empty?
+      flash[:removed] = review_removed_message
+      render json: {
+        issuesBefore: request_issues_update.before_issues.map(&:ui_hash),
+        issuesAfter: request_issues_update.after_issues.map(&:ui_hash),
+        withdrawnIssues: request_issues_update.withdrawn_issues.map(&:ui_hash)
+      }
+    elsif !request_issues_update.withdrawn_issues.empty?
```
leikkisa: Here you are checking if there are any withdrawn issues, instead of if all issues are withdrawn.  The "review withdrawn" message should only be if all of the issues are withdrawn.

app/controllers/appeals_controller.rb
Sjones352: This line checks if an issue is withdrawn then remove it and show alert.

client/app/intake/pages/addIssues.jsx
```diff
@@ -217,6 +217,8 @@ export class AddIssuesPage extends React.Component {
       </div>;
     };
 
+    const messageHeader = this.props.editPage ? 'Edit Issues' : 'Add / Remove Issues';
```
leikkisa: Nice! I wonder what Carola would think about just "Add Issues" for the intake version?

app/controllers/appeals_controller.rb
```diff
@@ -111,11 +111,12 @@ def url_appeal_uuid
 
   def update
     if request_issues_update.perform!
-      flash[:removed] = review_removed_message if request_issues_update.after_issues.empty?
+      removed_or_withdrawn_issues_success
```
leikkisa: Can we give this a consistent name with the one in the claim_reviews_controller? Also, maybe it could be something like "set_flash_success_message".  I don't feel strongly about the specific name, but I think making them consistent with each other would be good.

client/app/intakeEdit/reducers/index.js
```diff
@@ -63,6 +63,9 @@ export const intakeEditReducer = (state = mapDataToInitialState(), action) => {
       },
       issuesBefore: {
         $set: formatRequestIssues(action.payload.responseObject.issuesBefore)
+      },
+      withdrawnIssues: {
+        $set: action.payload.responseObject.withdrawnIssues
```
leikkisa: Should this also be wrapped with "formatRequestIssues".  I actually don't know that we even need to send this from the backend and set this.

spec/feature/intake/appeal/edit_spec.rb
```diff
@@ -494,6 +494,9 @@
       click_edit_submit
 
       expect(page).to have_current_path("/queue/appeals/#{appeal.uuid}")
+
+      expect(page).to have_content("Edit Completed")
```
leikkisa: I think you should replace this with a test for the withdrawal-specific message, since it would also have "Edit Completed" in other cases.

spec/feature/intake/higher_level_review/edit_spec.rb
```diff
@@ -1365,8 +1365,39 @@ def click_cancel(visit_page)
...
+
+        click_withdraw_intake_issue_dropdown(1)
+        click_withdraw_intake_issue_dropdown(2)
+
+        fill_in "withdraw-date", with: withdraw_date
+        click_edit_submit
+        sleep 1
+
+        expect(current_path).to eq("/decision_reviews/education")
+        expect(page).to have_content("Edit Completed")
```
leikkisa: Same comment as above about having a more specific test.

spec/feature/intake/supplemental_claim/edit_spec.rb
```diff
@@ -768,6 +767,39 @@ def click_cancel(visit_page)
...
+        scenario "show alert message when all decision reviews are withdrawn" do
+          visit "supplemental_claims/#{supplemental_claim.uuid}/edit"
+          click_withdraw_intake_issue_dropdown(1)
+          click_withdraw_intake_issue_dropdown(2)
+          fill_in "withdraw-date", with: withdraw_date
+          click_edit_submit
+          sleep 1
+
+          expect(current_path).to eq("/decision_reviews/education")
+          expect(page).to have_content("Edit Completed")
```
leikkisa: Same comment as above.

spec/feature/intake/appeal/edit_spec.rb
```diff
@@ -650,7 +653,7 @@
           click_intake_confirm
 
           expect(page).to have_current_path("/queue/appeals/#{appeal.uuid}")
-          expect(page).to have_content("Review Removed")
+          expect(page).to have_content("Edit Completed")
```
leikkisa: Same comment, except it would have the removal-specific message.

PR #10498, opened 2019-04-22
Option to navigate between daily docket apps

### LINKED ISSUE #10451: Hearing Prep Merge | Option to navigate between apps
Requires #10410 

Once the proper judge functionality is in place in hearing schedule, we want to allow users to navigate between hearing prep and hearing schedule so we can tease out bugs before moving all users over to hearing schedule. 

In hearing schedule, we want to display the following alert (the button navigates to hearings/dockets) - only for hearing prep users:

Screen Shot 2019-04-15 at 1 53 34 PM

In hearing prep, we want to update the alert to include a button that navigates to hearings/schedule:

Screen Shot 2019-04-15 at 1 54 05 PM

lomaxap:
Resolve #10451

- Adds buttons that navigates between /hearings/schedule and hearings/docket


lomaxap:
> * _Question:_ The alerts are showing on individual docket pages as opposed to `hearings/schedule` and `hearings/dockets`, but the buttons to switch between the views link to those index pages. Is that intentional? I noticed it's not explicitly noted where the alerts are supposed to show up in the issue.

I don't think hearing schedule dockets necessarily have a corresponding hearing prep docket. The schedule lists hearings associated with the same `HearingDay` while prep is hearings grouped by their `scheduled_for` date. I think Sharon left it as `hearings/schedule` and `hearings/dockets` in the AC for #10451 for simplicity. Might be worth investigating if it's easy enough to find the corresponding prep/schedule docket, but have left it off this PR.

> * _Request:_ Can we check for the presence of the alerts in a feature test? Maybe just add a couple lines to existing tests?
👍 

> * _Request:_ The alerts use a warning (`!` inside triangle) symbol instead of a info (`i` inside circle) symbol as shown in the comps.

I think the comps just have an out of date symbol for warnings in the sketch file. I can change it to an info `Alert` but the background color would be blue. I think that makes more sense though, so I'll go with that.

PR #10499, opened 2019-04-22
Create TimedHoldTask

lowellrex:
Connects #9207.

This PR creates the `TimedHoldTask` that enables all later [phases of the related ticket](https://github.com/department-of-veterans-affairs/caseflow/issues/9207#issuecomment-484914183).


app/models/task_timer.rb
```diff
@@ -7,4 +7,14 @@ class TaskTimer < ApplicationRecord
   def veteran
     task.appeal.veteran
   end
+
+  def self.requires_processing
+    # Only process timers for tasks that are active.
+    # Inline original definition of the requires_processing function due to limitations of mixins.
+    active_tasks.processable.attemptable.unexpired.order_by_oldest_submitted
```
aroltsch: This will override this call as well https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/jobs/task_timer_job.rb#L10. Is it intentional? We should probably remove `includes(:task)` piece then

lowellrex: Yup! That's the only place we call this function at the moment.

I'm more than happy to remove the `includes(:task)` call in `TaskTimerJob` but what will that do? Isn't that just an eager load? I'm still not super familiar with all the Rails-isms.

aroltsch: I think you actually want to use `includes` instead of `joins`  since you are accessing task data here: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/jobs/task_timer_job.rb#L24. 

You can read here: http://tomdallimore.com/blog/includes-vs-joins-in-rails-when-and-where/

lowellrex: Looks like I needed to add both. Done!

app/models/task_timer.rb
```diff
@@ -7,4 +7,14 @@ class TaskTimer < ApplicationRecord
...
   end
+
+  def self.requires_processing
+    # Only process timers for tasks that are active.
+    # Inline original definition of the requires_processing function due to limitations of mixins.
+    with_active_tasks.processable.attemptable.unexpired.order_by_oldest_submitted
+  end
+
+  def self.with_active_tasks
+    joins(:task).where("tasks.status not in (?)", Task.inactive_statuses).includes(:task)
```
aroltsch: Do you really need both? Unless I am missing something. 

Would this not work?
```
includes(:tasks).where(tasks: { status: Task.inactive_statuses} )
```

lowellrex: That totally works! Thanks!

Issue #10444, opened 2019-04-15
async jobs page pagination


pkarman:
We have too many jobs on the /jobs page to comfortably scroll through them.

Add some pagination.

## Validation steps:

* visit the /jobs page
* if there are more than 50 async jobs there should be links to the multiple pages at the bottom of the page
* there should be links at the top of the page to each type of async job, and those should also be paginated.

Sjones352:
**Passes UAT**
https://appeals.cf.uat.ds.va.gov/jobs

<img width="1367" alt="Screen Shot 2019-04-29 at 4 05 01 PM" src="https://user-images.githubusercontent.com/23080951/56923473-a823c300-6a98-11e9-9db4-cb035117ce29.png">
<img width="1265" alt="Screen Shot 2019-04-29 at 4 04 43 PM" src="https://user-images.githubusercontent.com/23080951/56923472-a823c300-6a98-11e9-9fed-846f9df281dd.png">

PR #10509, opened 2019-04-23
Paginate async jobs page

pkarman:
connects #10444 

### Description

Our backlog of async jobs can run into the 1000s which makes loading on a single page prohibitive.

This change implements pagination feature for the /jobs page, and adds subnav to easily filter by job type.

Screen Shot 2019-04-23 at 2 04 11 PM

### DIFF
--- a/client/app/components/Pagination.jsx
+++ b/client/app/components/Pagination.jsx
@@ -10,7 +10,7 @@ class Pagination extends React.PureComponent {
 
   // pagination button is zero-based but we are one-based so fake out the currentPage.
   currentPageIndex = () => {
-    this.props.currentPage - 1;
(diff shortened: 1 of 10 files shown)


client/app/components/Pagination.jsx
```diff
@@ -0,0 +1,140 @@
...
+import PropTypes from 'prop-types';
+
+import PaginationButton from './PaginationButton';
+
+class Pagination extends React.PureComponent {
+  handleChange = (index) => {
+    this.props.updatePage(index);
+  };
+
+  // pagination button is zero-based but we are one-based so fake out the currentPage.
```
leikkisa: I wonder if it's worth making a new variable to make this clearer in the code.

Issue #10523, opened 2019-04-24
RAMP intake referring to wrong user


pkarman:
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3276/events/309792/

For intake 153482 the user is `AMCMAWIL` but the EPE refers to user `AMCAFERG` who is no longer active.

Seems like whatever is associating the Intake with (maybe an existing?) EPE needs to coordinate the `user_id`.

leikkisa:
I'm seeing a couple of recent examples of this for a Supplemental Claim, but no more for RAMP (this fix is specifically for RAMP), and I don't see how we would validate this otherwise. So I'm closing this for now, and if it comes up again we can re-open.

PR #10528, opened 2019-04-24
Stop connecting intakes to previous ramp election

leikkisa:
connects #10523

This is causing a bug where a Ramp Election has multiple RampElectionIntakes instead of one RampElectionIntake, which is causing issues for EP Establishment because it's using the user who processed the older intake on "intake_processed_by"

### DIFF
--- a/app/models/ramp_election_intake.rb
+++ b/app/models/ramp_election_intake.rb
@@ -163,8 +163,4 @@ class RampElectionIntake < Intake
       update!(detail: existing_ramp_election)
     end
   end
-
-  def veteran_ramp_elections
-    RampElection.where(veteran_file_number: veteran_file_number)
-  end
 end
--- a/app/models/ramp_election_intake.rb
+++ b/app/models/ramp_election_intake.rb
@@ -141,7 +141,9 @@ class RampElectionIntake < Intake
   end
 
   def new_intake_ramp_election
-    @new_intake_ramp_election ||= matching_ramp_election_with_notice_date || veteran_ramp_elections.build
+    @new_intake_ramp_election ||= RampElection.new(
+      veteran_file_number: veteran_file_number
+    )
   end
 
   def existing_ramp_election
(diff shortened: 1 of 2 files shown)


app/models/ramp_election_intake.rb
```diff
@@ -141,7 +141,9 @@ def validate_detail_on_start
   end
 
   def new_intake_ramp_election
-    @new_intake_ramp_election ||= matching_ramp_election_with_notice_date || veteran_ramp_elections.build
+    @new_intake_ramp_election ||= RampElection.new(
```
pkarman: is `RampElection.new(...)` the same as `veteran_ramp_elections.build`?

spec/models/ramp_election_intake_spec.rb
```diff
@@ -441,13 +441,12 @@ def veteran_end_products
...
-      it "saves intake and sets detail to a new ramp election" do
+      it "creates a new RAMP Election and does not set detail to the existing RAMP Election" do
         expect(subject).to be_truthy
 
         expect(intake.detail).to have_attributes(
-          id: ramp_election.id,
           veteran_file_number: "64205555",
-          notice_date: 5.days.ago.to_date
+          notice_date: nil
         )
```
pkarman: it would be nice to see a new test explicitly for this change.

app/models/ramp_election_intake.rb
leikkisa: Oh, thank you. Yes, we don't need the veteran_ramp_elections method anymore so I meant to remove it.  It's equivalent code, and more consistent with the code for Ramp Refiling.

spec/models/ramp_election_intake_spec.rb
leikkisa: This is actually the specific test for this change. Previously, it checked that the detail was the existing ramp election with a notice date (created on line 441). The test now checks that it creates a new ramp election instead of connecting to the existing one.

spec/models/ramp_election_intake_spec.rb
```diff
@@ -441,13 +441,16 @@ def veteran_end_products
...
         expect(subject).to be_truthy
 
         expect(intake.detail).to have_attributes(
-          id: ramp_election.id,
           veteran_file_number: "64205555",
-          notice_date: 5.days.ago.to_date
+          notice_date: nil
+        )
+
+        expect(intake.detail).to_not have_attributes(
```
pkarman: perfect!

PR #10531, opened 2019-04-24
Display appropriate error for duplicate veteran phone number

### LINKED ISSUE #10424: Investigate why users get 'additional access needed' error but they have full access to document in VBMS
Support ticket: https://github.com/department-of-veterans-affairs/appeals-support/issues/4907

Slack thread: https://dsva.slack.com/archives/C2ZAMLK88/p1554918721214300

AC: Investigate why users get 'additional access needed' error but they have full access to document in VBMS

hschallhorn:
Resolves #10424

### Description
Checks for duplicate veteran phone numbers, previously causing "Additional access needed" errors, and displays appropriate error to the user


client/app/components/LoadingDataDisplay.jsx
```diff
@@ -15,6 +15,12 @@ const accessDeniedMsg = <div>
         It looks like you do not have the necessary level of access to view this information.<br />
         Please check with your application administrator before trying again.</div>;
 
+const duplicateNumberTitle = { title: COPY.DUPLICATE_PHONE_NUMBER_TITLE };
+const duplicateNumberMsg = <div>
+        Duplicate phone numbers documented.<br />
+        Please contact VLJ Support to request that this veterans information be updated to remove duplicate
```
monfresh: should `veterans` be `veteran's` ?

app/controllers/appeals_controller.rb
```diff
@@ -182,4 +182,14 @@ def set_flash_success_message
...
+
+  def render_access_error
+    if appeal.veteran.multiple_phone_numbers?
+      return render(Caseflow::Error::ActionForbiddenError.new(
+        message: COPY::DUPLICATE_PHONE_NUMBER_TITLE
+      ).serialize_response)
+    end
+
+    render(Caseflow::Error::ActionForbiddenError.new.serialize_response)
+  end
```
monfresh: Just a thought, and I'm not sure if passing in a nil message is problematic, but we could eliminate some duplication like this:
```ruby
def render_access_errror
  render(Caseflow::Error::ActionForbiddenError.new(
     message: access_error_message
   ).serialize_response)
end

def access_error_message
  COPY::DUPLICATE_PHONE_NUMBER_TITLE if appeal.veteran.multiple_phone_numbers?
end
```

app/models/veteran.rb
```diff
@@ -114,14 +114,16 @@ def fetch_bgs_record
...
     result && result[:file_number] && result
   rescue BGS::ShareError => error
+    @access_error = error.message
+
     # Now that we are always checking find_flashes for access control before we fetch the
     # veteran, we should never see this error. Reporting it to sentry if it happens
-    Raven.capture_exception(error)
-    @access_error = error.message
+    unless error.message.match?(/Sensitive File/)
+      Raven.capture_exception(error)
```
monfresh: Was this intentional to exclude errors that match `Sensitive File` from Sentry?

app/controllers/appeals_controller.rb
pkarman: If we added a more generic error message as the default in `access_error_message` I'm +1 to that change.

app/models/veteran.rb
pkarman: I'm ok excluding from Sentry. We do similar in Intake, since there's nothing actionable on our end in that case.

PR #10540, opened 2019-04-24
10354/schedule hearings in caseflow

### LINKED ISSUE #10354: Create ScheduleHearingTasks for ColocatedTasks with the schedule_hearing action
Now that we are scheduling hearings for `LegacyAppeal`s using [the `ScheduleHearingTask`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/tasks/schedule_hearing_task.rb) we should make use of that same task for `ColocatedTask`s with the `schedule_hearing` action instead of [changing the VACOLS location to 57](https://github.com/department-of-veterans-affairs/caseflow/blob/fbde547b958f9a787b23618e1775389fd880457c/app/models/tasks/colocated_task.rb#L126).

`ColocatedTask`s associated with `LegacyAppeal`s with the `schedule_hearing` action are created as children of `AttorneyLegacyTask`s (see the results of the query below), so we will need to make sure that the VACOLS location code remains "CASEFLOW" and that the `ScheduleHearingTask` is created as a child of the `RootTask` for the appeal.

```sql
select coalesce(parent_id, 0)
, count(*)
from tasks
where type = 'ColocatedTask'
  and assigned_to_type = 'Organization'
  and appeal_type = 'LegacyAppeal'
  and action = 'schedule_hearing'
group by 1;
```

## Acceptance criteria
* [ ] Determine what state the task tree after a member of the VLJ support staff has selected the "Hearing does need to be scheduled, send to hearing branch" option
* [ ] Stop changing the VACOLS location codes for these `ColocatedTask`s
* [ ] Create a `ScheduleHearingTask` for these `ColocatedTask`s

kevmo:
Resolves #10354

### Testing Plan
`bx rspec spec/models/task/colocated_task_spec.rb`


app/models/tasks/colocated_task.rb
```diff
@@ -119,16 +119,24 @@ def update_location_in_vacols
                                                      all_tasks_closed_for_appeal?
 
     if all_colocated_tasks_for_legacy_appeal_complete
-      AppealRepository.update_location!(appeal, location_based_on_action)
+      new_vacols_location = location_based_on_action
+      AppealRepository.update_location!(appeal, new_vacols_location)
```
lowellrex: Why did we make this change? I don't think there's a functional difference unless I'm missing something?

app/models/tasks/colocated_task.rb
```diff
@@ -119,16 +119,24 @@ def update_location_in_vacols
...
     case action.to_sym
-    when :translation, :schedule_hearing
-      return assigned_by.vacols_uniq_id if status == Constants.TASK_STATUSES.cancelled && action == "schedule_hearing"
+    when :schedule_hearing
+      # Return to attorney if the task is cancelled. For instance, if the VLJ support staff sees that the hearing was
+      # actually held.
+      return assigned_by.vacols_uniq_id if status == Constants.TASK_STATUSES.cancelled
 
-      return LegacyAppeal::LOCATION_CODES[action.to_sym]
+      ScheduleHearingTask.create!(appeal: appeal, parent: appeal.root_task)
```
lowellrex: Can we add a comment about why we're creating this schedule hearing task? Maybe we could include it in the comment below?

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -58,7 +58,7 @@ def label
   end
 
   def create_parent_hearing_task
-    if parent.type != HearingTask.name
+    if !parent || parent.type != HearingTask.name
```
lowellrex: When do we expect there not to be a parent task? A quick grep and I still can't find anywhere where the `ScheduleHearingTask` would not have a parent `git grep -E "ScheduleHearingTask\.create" app`. Maybe we have test data that doesn't reflect a state the application will ever actually end up in?

spec/models/tasks/colocated_task_spec.rb
```diff
@@ -208,7 +208,7 @@
         it "should update location to schedule hearing in vacols" do
           expect(vacols_case.bfcurloc).to_not eq staff.slogid
           colocated_admin_action.update!(status: Constants.TASK_STATUSES.completed)
-          expect(vacols_case.reload.bfcurloc).to eq LegacyAppeal::LOCATION_CODES[:schedule_hearing]
+          expect(vacols_case.reload.bfcurloc).to eq LegacyAppeal::LOCATION_CODES[:caseflow]
```
lowellrex: Can we also check to see that a `ScheduleHearingTask` was created?

Issue #10516, opened 2019-04-24
all the flakey tests


pkarman:
* [x] https://circleci.com/gh/department-of-veterans-affairs/caseflow/59562
* [x] https://circleci.com/gh/department-of-veterans-affairs/caseflow/59503
* [x] https://circleci.com/gh/department-of-veterans-affairs/caseflow/59751
       https://circleci.com/gh/department-of-veterans-affairs/caseflow/64594
* [x] https://circleci.com/gh/department-of-veterans-affairs/caseflow/59776
* [x] https://circleci.com/gh/department-of-veterans-affairs/caseflow/60756
* [x] https://circleci.com/gh/department-of-veterans-affairs/caseflow/60829
* [x] https://circleci.com/gh/department-of-veterans-affairs/caseflow/61697
* [x] https://circleci.com/gh/department-of-veterans-affairs/caseflow/61696 (just the HearingTask one) #10692 
* [x] https://circleci.com/gh/department-of-veterans-affairs/caseflow/62205 (another instance of the HearingTask one in a different spec)
* [x] https://circleci.com/gh/department-of-veterans-affairs/caseflow/63353
* [x] https://circleci.com/gh/department-of-veterans-affairs/caseflow/64726 (vadotgov api)

monfresh:
What does it mean when a checkbox above is checked? If it means that the flake has been fixed, then some of these should be unchecked, like the Hearing Task flake, which happened on my PR today: https://circleci.com/gh/department-of-veterans-affairs/caseflow/63586

The weird thing is it seems to always fail on this particular line in the controller spec, but when I add a `puts` statement locally in `ScheduleHearingTask`, that line never gets called by the test.

pkarman:
the hearing task flake was happening in multiple places. We fixed one, and then another popped up.

I fixed it yesterday in ce376882610bc32f6a5de79146db2af0ab8b8bec

pkarman:
* [ ] https://github.com/department-of-veterans-affairs/caseflow/pull/10079/commits/9c768e12c94d0394ed9376c003beec97bea8aec9
* [ ] https://circleci.com/gh/department-of-veterans-affairs/caseflow/64702 (Reader Short list of documents Rotating documents - reader.reader_spec)
* [ ] https://circleci.com/gh/department-of-veterans-affairs/caseflow/64781 (hearting schedule daily docket dropdown)
[...]

PR #10565, opened 2019-04-25
Fix flakey tests

pkarman:
**Why**: Multiple before blocks must still run synchronously. If we visit the page via
the browser in the initial before block, then we always run the risk that the browser
will load the page before the other before blocks have finished executing. In this
case, that meant the data wasn't yet completely set up before the spec executed.

connects #10516


spec/feature/queue/search_spec.rb
```diff
@@ -160,20 +165,18 @@
 
               context "the EP has been established" do
                 before do
-                  end_product_establishment_1.commit!
```
leikkisa: I don't know what these were for, but guessing it's okay without them.

pkarman: hlr.establish! is calling commit on each EPE so we just need that, but we needed to reload it first.

PR #10570, opened 2019-04-26
Add endpoint for creating timed holds

lowellrex:
Connects #9207. Implements approach 3 outlined in the tech spec from #10512.

Adds an endpoint (`/tasks/:task_id/place_hold`) to create a timed hold task for an existing task.

### DIFF
--- a/app/models/tasks/timed_hold_task.rb
+++ b/app/models/tasks/timed_hold_task.rb
@@ -47,6 +47,7 @@ class TimedHoldTask < GenericTask
 
   # We cancel the siblings after we have created the new task so that the parent task stays on hold.
   def cancel_active_siblings
-    siblings.select(&:active?).each { |sibling_task| sibling_task.update!(status: Constants.TASK_STATUSES.cancelled) }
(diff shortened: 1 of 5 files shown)


app/models/tasks/timed_hold_task.rb
```diff
@@ -32,4 +44,9 @@ def hide_from_task_snapshot
   def verify_parent_task_presence
     fail(Caseflow::Error::InvalidParentTask, message: "TimedHoldTasks must have parent tasks") unless parent
   end
+
+  # We cancel the siblings after we have created the new task so that the parent task stays on hold.
+  def cancel_active_siblings
+    siblings.select(&:active?).each { |sibling_task| sibling_task.update!(status: Constants.TASK_STATUSES.cancelled) }
```
tomas-nava: Is there any chance this could sweep up non-`TimedHoldTask`s that we wouldn't want canceled?

lowellrex: Great catch! I'll add some code and a test to confirm we won't end up doing this.

PR #10577, opened 2019-04-26
Daily Docket Hold Open Field

### LINKED ISSUE #10526: Prevent Postponed Hearings Notes from Being Editable on Schedule Daily Docket
The notes field should not be editable when the hearing is postponed, and the save button should never appear https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/4753/

### LINKED ISSUE #10561: Bug: Judge View Hearing Schedule & Daily Docket Errors
The implementation of the new Judge view of the Daily Docket has several errors:

- [ ] Legacy hearings no longer have the Hold Open option. This should be added back to Legacy hearings. 
- [ ] Legacy hearings have an AOD reason option. This should be removed and only apply to AMA hearings. 
- [ ] The switch to old view banner should be on the schedule page rather than the Daily Docket pages. 

Please see the following designs for corrections:
https://va-gov.invisionapp.com/share/PMR1ID7XSHT#/353825994_Daily_Docket_Saved_Changes
https://va-gov.invisionapp.com/share/RYRI6Y153ZA#/358206234_Updated_Version

lomaxap:
resolves #10561
resolves #10526 

- adds Hold Open dropdown to legacy hearings
- fixes possible FE bug where legacy hearings and hearings may have the same id. Switches to using external_id as key for hearings in redux
- add more data to legacy hearing factories
- ensure all inputs have unique names / ids base on external_id
- moves alert to list page
- makes postponed and cancelled hearings notes disabled

### DIFF
--- a/client/app/hearingSchedule/components/DailyDocketRow.jsx
+++ b/client/app/hearingSchedule/components/DailyDocketRow.jsx
@@ -11,7 +11,7 @@ import {
   DispositionDropdown, TranscriptRequestedCheckbox, HearingDetailsLink,
   AodDropdown, AodReasonDropdown, HearingPrepWorkSheetLink, StaticRegionalOffice,
   NotesField, HearingLocationDropdown, StaticHearingDay, TimeRadioButtons,
-  Waive90DayHoldCheckbox
+  Waive90DayHoldCheckbox, HoldOpenDropdown
 } from './DailyDocketRowInputs';
 
 const SaveButton = ({ hearing, cancelUpdate, saveHearing }) => {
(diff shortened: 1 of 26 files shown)


client/app/hearingSchedule/components/DailyDocketRow.jsx
```diff
@@ -103,7 +103,8 @@ class HearingActions extends React.Component {
     return <React.Fragment>
       <HearingPrepWorkSheetLink hearing={hearing} />
       <AodDropdown {...inputProps} />
-      <AodReasonDropdown {...inputProps} />
+      {hearing.docketName === 'hearing' && <AodReasonDropdown {...inputProps} />}
+      {hearing.docketName !== 'hearing' && <HoldOpenDropdown {...inputProps} />}
```
tomas-nava: maybe set up a `hearingType` variable here like you did [in DailyDocketModals](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/hearingSchedule/components/DailyDocketModals.jsx#L89)?

PR #10580, opened 2019-04-26
Exclude TrackVeteranTasks from Appeal.assigned_to_location determination

### LINKED ISSUE #10368: Exclude TrackVeteranTasks from Appeal.assigned_to_location determination
The [logic in `Appeal.assigned_to_location`](https://github.com/department-of-veterans-affairs/caseflow/blob/5f441cfaf1aa461df3f849aa828a168eebf7e41d/app/models/appeal.rb#L139) incorrectly includes `TrackVeteranTask`s which is causing confusing values to be displayed in the "assigned to" column of some queue table views (see #10344 for an example). Since `TrackVeteranTask`s do not require their assignees to do any work on the appeal we should exclude them from the `Appeal.assigned_to_location` determination.

## Acceptance criteria
* [ ] Exclude `TrackVeteranTask`s from `Appeal.assigned_to_location` determination

kevmo:
Resolves #10368 

### Description
Filtered TrackVeteranTasks from Appeal.assigned_to_location determination using ActiveRecord `where.not` clause/method. Added spec. 

### Testing Plan
`bx rspec spec/models/appeal_spec.rb`


spec/models/appeal_spec.rb
```diff
@@ -694,6 +694,19 @@
...
+      let!(:appeal) { create(:appeal) }
+      let!(:root_task) { create(:root_task, appeal: appeal, status: :in_progress) }
+
+      before do
+        create(:track_veteran_task, appeal: appeal, status: :in_progress)
+      end
+
+      it "does not include TrackVeteranTasks in its determinations" do
+        expect(appeal.assigned_to_location).to eq(COPY::CASE_LIST_TABLE_CASE_STORAGE_LABEL)
+      end
```
lowellrex: Great test!

app/models/task.rb
```diff
@@ -35,6 +35,12 @@ class Task < ApplicationRecord
 
   scope :inactive, -> { where(status: inactive_statuses) }
 
+  scope :active_not_tracking, -> { active.where.not(type: TrackVeteranTask.name) }
+
+  scope :on_hold_not_tracking, lambda {
+    where(status: Constants.TASK_STATUSES.on_hold).where.not(type: TrackVeteranTask.name)
+  }
```
lowellrex: What are the advantages of this approach instead of something like the following where we just chain two existing scopes?
```ruby
# Definition:
scope(:non_tracking, -> { where.not(type: TrackVeteranTask.name) })

# Calling code:
active_tasks = tasks.active.non_tracking
...
on_hold_tasks = tasks.non_tracking.where(status: Constants.TASK_STATUSES.on_hold)
```

PR #10581, opened 2019-04-28
Handle Internal Server Errors from va dot gov Facilities API / Vets360 API

lomaxap:
This PR adds datadog tracking to the `fetch_veteran_hearing_locations` job while also improving error handling for on-the-fly requests to Facilities API and Vets360 API

### DIFF
--- a/app/controllers/hearings_controller.rb
+++ b/app/controllers/hearings_controller.rb
@@ -44,7 +44,10 @@ class HearingsController < ApplicationController
 
       render json: { hearing_locations: locations }
     rescue Caseflow::Error::VaDotGovAPIError => e
-      render json: { message: e.message["messages"][0]["key"] }, status: :bad_request
+      messages = e.message.try("messages") || []
+      render json: { message: messages[0].try("key") || e.message }, status: :bad_request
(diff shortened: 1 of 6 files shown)


app/controllers/hearings_controller.rb
```diff
@@ -44,7 +44,10 @@ def find_closest_hearing_locations
 
       render json: { hearing_locations: locations }
     rescue Caseflow::Error::VaDotGovAPIError => e
-      render json: { message: e.message["messages"][0]["key"] }, status: :bad_request
+      messages = e.message.try("messages") || []
```
tomas-nava:
```suggestion
      messages = e.message.dig(:messages) || []
```

app/controllers/hearings_controller.rb
```diff
@@ -44,7 +44,10 @@ def find_closest_hearing_locations
...
+      render json: { message: messages[0].try("key") || e.message }, status: :bad_request
```
tomas-nava:
```suggestion
      render json: { message: messages[0].dig(:key) || e.message }, status: :bad_request
```

PR #10583, opened 2019-04-29
Batch document counts

youngfreezyVA:
### Description
Implement new endpoint that accepts an array of appeal IDs and uses threads to query document count in parallel.


aroltsch:
Closing in favor of https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/1090/files and Lowell's idea:

```
I think we should reduce the number of retries specifically for this endpoint, so that we attempt to get the number of documents once (for the queue page at least) and display whatever we return. At the beginning of the day (when documents are not cached) this will return a bunch of "unknown" values but it will also trigger those downloads. Every subsequent queue table view should get a response back from efolder that includes accurate document counts because they have already been fired off by the earlier requests. And since we reload the entire queue table view quite regularly (due to react router issues, I think) I think the visibility to the Caseflow users will be minimal.
```

PR #10584, opened 2019-04-29
Update gems with bummr

monfresh:


monfresh:
Yes, they are. The changes are only present in the commit that upgrades Rubocop. Most of the changes are making the exception variable naming consistent.

PR #10587, opened 2019-04-29
Catch duplicate root tasks on appeals

### LINKED ISSUE #10369: Throw error if we attempt to create a second RootTask for an Appeal
Our code currently relies on their being only a single `RootTask` for each `Appeal` but does not enforce that anywhere in the code. This ticket exists to make that reliance explicit by throwing an error if we attempt to create a second `RootTask` for an `Appeal`.

## Acceptance criteria
* [ ] Throw error if we attempt to create a second `RootTask` for an `Appeal`

lomky:
Resolves #10369

### Description
Appeals now check for an existing RootTask before creating another. Uses the error `DuplicateOrgTask`. Considers both active and inactive tasks.

### Acceptance Criteria
- [x] Appeals with existing RootTasks throw an error and do not create a second RootTask
  - [x] Active RootTasks
  - [x] Inactive RootTasks

### Testing Plan
1. None?


lowellrex:
Created a new ticket to split up `root_task.rb` and satisfy Code Climate (#10601).

monfresh:
If this fix is not critical to merge right away, I would take the time in this PR to move the `create_root_and_sub_tasks!` method into a separate class. I believe in the "fix things as you touch them" philosophy. From what I can tell, this is the only method that creates a RootTask, and so the `before_create` callback would only apply to it.

I realize that `create_subtasks!` is also called by JudgeTask, and so we can't move it out of RootTask just yet. We will have some duplication, but that's OK. We can address that later the next time we need to touch `RootTask` or `JudgeTask`. The point is to make incremental improvements to the code every time we touch it.

PR #10595, opened 2019-04-29
Add a rake task and tests for changing task assigned_to in bulk

tomas-nava:
### Description
Adds a rake task for changing the `assigned_to` value of tasks in bulk.


lib/tasks/tasks.rake
```diff
@@ -59,4 +60,64 @@ namespace :tasks do
...
+      fail InvalidOrganization, "No organization with id #{check_id}!" if Organization.find_by(id: check_id).blank?
+    end
+
+    from_organization = Organization.find(from_id)
+    to_organization = Organization.find(to_id)
+
+    target_tasks = if extras.any?
+                     task_class.find(extras).select { |task| task.assigned_to_id = from_id }
+                   else
+                     task_class.all.active.where(assigned_to_id: from_id)
```
lowellrex: Let's check to see if the `assigned_to_type` is `Organization` here as well, so we don't accidentally scoop up any `User`s with the same ID.

PR #10599, opened 2019-04-30
Allow HearingDispositionChangeJob without any tasks

lowellrex:
Fix to resolve [a bug we see from time to time](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/4731/) caused by the script never calling `task_count_for()`, therefore never setting the `@task_count_for` instance variable, causing the program to attempt to loop over `nil`.

### DIFF
--- a/spec/jobs/hearing_disposition_change_job_spec.rb
+++ b/spec/jobs/hearing_disposition_change_job_spec.rb
@@ -350,5 +350,11 @@ describe HearingDispositionChangeJob do
         subject
       end
     end
+
+    context "when there are no DispositionTasks to be processed" do
+      it "runs successfully but does not do any work" do
+        expect { subject }.to_not raise_error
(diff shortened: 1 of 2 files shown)


spec/jobs/hearing_disposition_change_job_spec.rb
```diff
@@ -350,5 +350,11 @@ def create_disposition_task_for_legacy_hearings_ancestry(associated_hearing: nil
         subject
       end
     end
+
+    context "when there are no DispositionTasks to be processed" do
+      it "runs successfully but does not do any work" do
+        expect { subject }.to_not raise_error
+      end
+    end
```
lowellrex: Reproduced this situation by causing the application to never enter the loop that would create the `@task_count_for` variable.

Issue #8169, opened 2018-12-10
Require unique url path for Organizations


lowellrex:
The [organizations table](https://github.com/department-of-veterans-affairs/caseflow/blob/f358ba0275a96e7efd22863b115a41c31bd4ca87/db/schema.rb#L546) does not currently require the `url` field. During the effort to [move judge team membership](https://github.com/department-of-veterans-affairs/caseflow/issues/8063) into the Organization model @marvokdolor-gov noted that judges would like to be able to add/remove attorneys from their teams. Because judges are admins of their judge teams, they already have the capability to do this. However, there is no easy way for them to navigate to the user management page and the URL for these pages is not all that discoverable as it currently stands (the fallback is to use the table ID which results in JudgeTeam urls being something like `/organization/26/users`). Making the JudgeTeam URL the same as the judge's CSS ID would make navigation to the user management page easier until we [address navigating to that page in the UI](https://github.com/department-of-veterans-affairs/caseflow/issues/8170). 

When we create Organizations without explicitly defining the `url` field, we could a `before_create` hook to just fill this in with the ID field (unless we don't have an ID at that point in time...) or a lowercased and hyphen-joined version of the organization name.

## Acceptance criteria
* [ ] Organizations are required to have `url` fields before being inserted in the database
* [ ] JudgeTeams are created with a `url` that matches the judge's CSS ID
* [ ] Existing organizations updated with appropriate `url`s

amprokop:
```
 before_create hook to just fill this in with the ID field (unless we don't have an ID at that point in time...)
``` 

I think (as you're surmising) this would need to be `after_create` but otherwise 👍

mdbenjam:
Sounds good to me 👍

lowellrex:
Added URLs to all `JudgeTeam`s that match their CSS IDs so @marvokdolor-gov can more easily find the user management page associated with the judge.

```
rails c> JudgeTeam.all.each { |team| team.update!(url: team.name.downcase) }
```

And the user management page for the judge team associated with the judge with CSS ID `BVAAABSHIRE` would be `https://appeals.cf.ds.va.gov/organizations/bvaaabshire/users`.

laurjpeterson:
@lowellrex - I see this was done for judge teams. Should we keep this open for other BVA teams?

lowellrex:
> @lowellrex - I see this was done for judge teams. Should we keep this open for other BVA teams?

Absolutely!

[...]

PR #10604, opened 2019-04-30
Organization url & name updates

lomky:
Bumps #8169 

### Description
URL must exist and be unique.
Name must exist.

DB changes coming in another PR to finish ticket.


lowellrex:
It looks like we've got a few failing tests caused by the new validations we added. I think we can resolve them by adding a `name` and `url` when we instantiate those objects (like [here, for instance](https://github.com/department-of-veterans-affairs/caseflow/blob/b61108e5c55fe3470f6d6834f56f6f642bba67e7/spec/controllers/organizations/tasks_controller_spec.rb#L11)).

lomky:
Looks like I have a lot of tests to fix! VSO type are done, I'll dig into the rest tomorrow.

lomky:
I believe I have gotten all the non-feature tests fixed. I also think I got all the feature tests as well, but I will let CircleCI confirm that for me once the FACOLs error is resolved

spec/models/appeal_spec.rb
```diff
@@ -540,7 +540,12 @@
         let(:appeal) do
           create(:appeal, claimants: [create(:claimant, participant_id: participant_id_with_nil)])
         end
-        let!(:vso) { Vso.create(name: "Test VSO") }
+        let!(:vso) do
+          Vso.create(
```
pkarman: There is a `vso` factory already in `spec/factories/organization.rb` -- maybe we make sure the default name and url are set there and use the factory instead?

spec/models/hearing_spec.rb
```diff
@@ -76,7 +76,13 @@
     subject { hearing.assigned_to_vso?(user) }
 
     context "when the hearing is not assigned a vso" do
-      let(:vso) { Vso.create(participant_id: "999") }
+      let(:vso) do
+        Vso.create(
+          participant_id: "999",
```
pkarman: same comment about factory.

spec/models/judge_team_spec.rb
```diff
@@ -50,7 +50,11 @@
     # Limitation of the current approach is that users are effectively limited to being the admin of a single JudgeTeam.
     context "when user is admin of multiple JudgeTeams" do
       let!(:first_judge_team) { JudgeTeam.create_for_judge(user) }
-      let!(:second_judge_team) { JudgeTeam.create_for_judge(user) }
+      let!(:second_judge_team) do
```
pkarman: could we use a factory here? not sure if that would simplify.

PR #10605, opened 2019-04-30
Enforcing Organization URL uniqueness at the db level

### LINKED ISSUE #8169: Require unique url path for Organizations

lomky:
Resolves #8169 

### Description
DB migration to enforce uniqueness at hte DB level, to avoid race conditions from only having it as a validation.

### Schema Changes Only

* [x] Column comments updated
  * Not doing any new queries
* [x] Eyeball Production to confirm unique/existing URLs currently.
  * Waiting for response to [question on slack](https://dsva.slack.com/archives/C2ZAMLK88/p1556644556118700)


db/migrate/20190430152338_add_comments_to_organization_fields.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddCommentsToOrganizationFields < ActiveRecord::Migration[5.1]
+  def change
```
pkarman: I don't see the comments in the `db/schema.rb` change -- maybe another local `rake db:migrate` needed?

lowellrex:
Commands run in prod rails console to confirm uniqueness and presence.

```ruby
rails c> Organization.where(name: [nil, ""]).count
# 1
rails c> Organization.where(name: [nil, ""])
# #<ActiveRecord::Relation [#<BusinessLine id: 307, type: "BusinessLine", name: nil, role: nil, url: "unidentified", participant_id: nil>]>
# TODO: Asked the question in appeals-engineering slack channel. Let's move on with the URL uniqueness check.

rails c> Organization.where(url: [nil, ""]).count
# 0
rails c> Organization.count
# 215
rails c> Organization.all.pluck(:url).uniq.count
# 215
# All organizations have URLs and they are all unique.
```

Link to [the slack thread](https://dsva.slack.com/archives/C2ZAMLK88/p1556644556118700) where we discuss what to do about the `BusinessLine` without a name.

lomky:
Failing JudgeTeam model test corrected in #10604

PR #10610, opened 2019-04-30
Add seed data for lit support user

youngfreezyVA:


app/models/tasks/other_motion_mail_task.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class OtherMotionMailTask < MailTask
```
youngfreezyVA: I had to move  `OtherMotionMailTask` into its own file `app/models/tasks/other_motion_mail_task.rb` to avoid the unitialized constants error when creating the factory. I think it makes sense to pull out lit support related sub-classes in their own file anyways?

PR #10611, opened 2019-04-30
Appeal Status API bug fix and added logging

mc-usds:
Resolves #17976

### Description
Addressing a way that a hash could have possibly been returned instead of a string. Also added logging because this is a case that shouldn't be happening and will need to be investigated further once we have more information.

### Acceptance Criteria
- [ x] Code compiles correctly

### Testing Plan
1. Added a unit test for this edge case.

*Schema Changes Only*

* [ ] Column comments updated


spec/models/issue_spec.rb
```diff
@@ -195,6 +195,11 @@
       let(:codes) { %w[02 15 03 1234] }
       it { is_expected.to be_nil }
     end
+
+    context "when there are more levels than codes" do
+      let(:codes) { %w[01 01] }
+      it { is_expected.to eq("") }
```
pkarman: could we add a test here that Raven gets called as well? You can see an example of this:

https://github.com/department-of-veterans-affairs/caseflow/blob/master/spec/jobs/decision_issue_sync_job_spec.rb#L54

you'd want to stub `capture_message` instead of `capture_exception` and then check for it as else where in that other spec file.

app/models/issue.rb
```diff
@@ -273,6 +281,7 @@ def friendly_description_for_codes(code_array)
     issue_description
   end
   # rubocop:enable Metrics/CyclomaticComplexity
+  # rubocop:enable Metrics/PerceivedComplexity
```
pkarman: we're trying to avoid adding new exceptions, though in this case I think it's ok since the point is to capture data and then go back and fix this.

spec/models/issue_spec.rb
mc-usds: Added the stub call and the check if it's called.

Issue #10594, opened 2019-04-29
Add Edit Contention Title form


pkarman:
First step in allowing users to edit contention text.

## Acceptance criteria

* Change the "Add Issues" page title to "Edit Issues"
* In the UI, the title will be edited in place (not in a modal). Add a "Edit contention title" link below each issue on the page
* When the user clicks the "Edit contention title" link a form should appear with the current title text in a textarea field
* The form should contain a "Cancel" link that hides the form w/o changing the original request issue description.
* The "Submit" button does nothing (will be addressed in a different ticket).

## Mock
![Edit title 02](https://user-images.githubusercontent.com/11700545/55116058-53f57f80-50a3-11e9-9541-8db9df364e22.png)

## Technical Notes
* Feature flag this behind a `edit_contention_text` toggle.

leikkisa:
Validated at same time as 10780

PR #10629, opened 2019-05-01
WIP | Add Edit Contention Title form

Sjones352:
Connects #10594

### DIFF
--- a/client/app/intake/components/EditContentionTitle.jsx
+++ b/client/app/intake/components/EditContentionTitle.jsx
@@ -1,0 +1,82 @@
+import React from 'react';
+import COPY from '../../../COPY.json';
+import Link from '@department-of-veterans-affairs/caseflow-frontend-toolkit/components/Link';
+import TextareaField from '../../components/TextareaField';
+import Button from '../../components/Button';
+import { css } from 'glamor';
+
+const buttonStyling = css({
+  marginBottom: '15px'
+});
+
+const paragraphyStyling = css({
+  paddingLeft: '30px'
+});
+
+export default class EditContentionTitle extends React.PureComponent {
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      showEditTitle: false
+    };
+  }
+
+  editContentionOnClick = () => {
+    this.setState({ showEditTitle: true });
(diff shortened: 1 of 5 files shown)


client/app/intake/components/EditContentionTitle.jsx
```diff
@@ -0,0 +1,85 @@
...
+  marginBottom: '15px'
+});
+
+const paragraphyStyling = css({
+  paddingLeft: '30px'
+});
+
+const linkStyle = css({
+  paddingLeft: '0px'
+});
```
leikkisa: I don't know how I feel about doing the styling like this. I think it makes more sense where the styling is variable, but for something like this it seems like it should be in the stylesheets.

client/app/intake/components/EditContentionTitle.jsx
```diff
@@ -0,0 +1,85 @@
...
+  {this.state.showEditTitle && <div className="issue-text-style">
+    <TextareaField
+      name={title}
+      onChange={(value) => {
+        this.setState({ value });
+      }}
+      strongLabel
+    />
+    <p {...paragraphyStyling}>{this.props.issue.text}</p>
+    {this.props.issue.notes && <p {...css({
```
leikkisa: I don't like how the styling is being handled here either.

spec/feature/intake/appeal/edit_spec.rb
```diff
@@ -624,6 +630,20 @@
...
+      visit "appeals/#{appeal.uuid}/edit"
+      expect(page).to have_content("Edit contention title")
+
+      ri_contention_text = RequestIssue.find_by(contested_issue_description: "PTSD denied")
+
+      within first(".issue-edit-text") do
+        click_button("Edit contention title")
+      end
+
+      expect(ri_contention_text).to_not be_nil
```
leikkisa: I don't think this test makes sense, it is checking whether this request issue exists in the database.

client/app/intake/components/EditContentionTitle.jsx
Sjones352: Fixed Styling

PR #10631, opened 2019-05-01
Do not show canceled jobs in Async jobs page

leikkisa:
connects #10233

Do not show canceled jobs on jobs page.


app/models/concerns/asyncable.rb
```diff
@@ -107,10 +107,10 @@ def with_error
     end
 
     def potentially_stuck
-      processable
-        .where(canceled_at_column => nil)
-        .or(attempted_without_being_submitted)
+      expired_without_processing
         .or(with_error)
+        .or(attempted_without_being_submitted)
+        .where(canceled_at_column => nil)
```
leikkisa: This should get expired jobs, jobs with an error, jobs that were attempted without being submitted, and exclude canceled jobs. I wasn't 100% sure if this was the goal here, or if we only want to show expired jobs.

If we only want errors/attempted without being submitted if they are also expired, then we would do it like this:
```with_error
.or(attempted_without_being_submitted)
.expired_without_processing
.where(canceled_at_column => nil)```

File: spec/services/asyncable_jobs_spec.rb [hunk 2]
```diff
@@ -2,34 +2,35 @@
...
+           establishment_last_submitted_at: expired_day - 2.days,
            establishment_attempted_at: 7.days.ago,
            establishment_error: "bad problem",
            veteran_file_number: veteran.file_number)
   end
   let!(:sc_not_submitted) do
     create(:supplemental_claim,
-           establishment_attempted_at: 7.days.ago,
-           establishment_error: "bad problem",
+           establishment_attempted_at: unexpired_day,
```
leikkisa: Here is an unexpired one that was attempted but not submitted that is counting (an example of something to change if we don't want to show unexpired jobs yet).

spec/services/asyncable_jobs_spec.rb
```diff
@@ -40,41 +41,44 @@
...
       expect(subject.jobs).to be_a(Array)
-      expect(subject.jobs.length).to eq(6)
+      expect(subject.jobs.length).to eq(5)
       expect(subject.jobs).to include(hlr)
       expect(subject.jobs).to include(sc)
       expect(subject.jobs).to include(sc_not_submitted)
       expect(subject.jobs).to include(sc_not_attempted_expired)
-      expect(subject.jobs).to include(sc_not_attempted)
+      expect(subject.jobs).to include(sc_with_error)
+      expect(subject.jobs).to_not include(sc_not_attempted)
```
leikkisa: This one isn't expired yet, and not attempted yet, so it made sense to me that it wouldn't show up.

app/models/concerns/asyncable.rb
```diff
@@ -107,10 +107,10 @@ def with_error
     end
 
     def potentially_stuck
-      processable
-        .where(canceled_at_column => nil)
-        .or(attempted_without_being_submitted)
+      expired_without_processing
```
shanear: If we aren't using `expired_without_processing` here, it looks like there isn't anywhere where it's used. So it can be deleted.

app/models/concerns/asyncable.rb
shanear: I don't think we'd want only the expired errors. This looks good.

pkarman: We were using `processable` because we assumed users would want to see their pending jobs immediately, regardless of whether they had an error or not. Doubter confirmation that async processing works.

I'm not exactly clear on what this is attempting to fix. Are canceled jobs still showing on the /jobs page?

spec/services/asyncable_jobs_spec.rb
```diff
@@ -2,34 +2,35 @@
 
 describe AsyncableJobs do
   let(:veteran) { create(:veteran) }
+  let(:expired_day) { (DecisionReview::REQUIRES_PROCESSING_WINDOW_DAYS + 1).days.ago }
+  let(:unexpired_day) { (DecisionReview::REQUIRES_PROCESSING_WINDOW_DAYS - 1).days.ago }
   let!(:hlr) do
     create(:higher_level_review,
-           establishment_last_submitted_at: 7.days.ago,
+           establishment_last_submitted_at: expired_day - 3.days,
```
pkarman: it's not clear to me what `expired_day - 3.days` is for -- why the extra `- 3.days`? To get the sorting correct?

spec/services/asyncable_jobs_spec.rb
pkarman: When @shanear and I went over this originally, he pushed to have jobs show up if they had not yet been attempted, to re-assure users. If that expectation has changed, that's cool. But that was the original rationale.

spec/services/asyncable_jobs_spec.rb
leikkisa: Yeah, I originally put the expired day to help me tell the difference between which were supposed to be expired or not, and then I saw the next test which was testing the ordering.

leikkisa:
Ok! Happy to make it whatever. The last conversation I had been a part of was that we were going to limit it to expired jobs, but that makes sense to me (to keep the ones that haven't been attempted yet).  I wasn't quite sure what the intent was when I was looking at it.

app/models/concerns/asyncable.rb
```diff
@@ -108,9 +110,9 @@ def with_error
 
     def potentially_stuck
       processable
-        .where(canceled_at_column => nil)
-        .or(attempted_without_being_submitted)
         .or(with_error)
+        .or(attempted_without_being_submitted)
```
leikkisa: Currently, this will show jobs that were attempted without being submitted, even if they have since been processed. I'm not sure when this would happen and I don't see any cases of it for HLR, SC, or request_issues, but I get why it's in there (because jobs won't show up in processable if they are missing the submitted data).

I would suggest fixing this in the next PR.

If we want to show everything that hasn't been processed yet, then we might just want to show everything where processed_at is nil, and canceled_at is not nil.

spec/services/asyncable_jobs_spec.rb
```diff
@@ -41,6 +41,7 @@
   let!(:sc_canceled) do
     create(:supplemental_claim,
            veteran_file_number: veteran.file_number,
+           establishment_error: "bad problem",
```
leikkisa: The reason the canceled claim wasn't showing up in the tests before is because it didn't meet any of the other criteria for "potentially stuck". This shows that even with an error, it doesn't show up since it's canceled.

app/models/concerns/asyncable.rb
leikkisa: That's correct, it didn't pass validation for Sandra and I confirmed it too.

app/models/concerns/asyncable.rb
```diff
@@ -108,9 +110,9 @@ def with_error
...
+        .where(canceled_at_column => nil)
```
pkarman: this is part of `processable` now so duplicative?

leikkisa: This is where the "OR" part affects it.  The way it's written, it is processable OR it has an error OR it was attempted without being submitted, and those don't have the "canceled_at => nil" in them.  I assumed this was in here because the attempted_without_being_submitted wouldn't show up in processable.

I think we'll be better off putting something like where processed_at is nil, and canceled_at is not nil, and depending on what we decided, maybe where attempted_at is not nil.

leikkisa: We could put
```
attempted_without_being_submitted
.or(processable)
.with_error
.where(canceled_at_column => nil)
```

It's all a matter of what the intention is!

Sjones352:
@leikkisa can you please provide validation instructions

Sjones352:
Never mind It is connected to #10233 so validating now

PR #10633, opened 2019-05-02
Rebuilding Facols from Oracle Base Image

enriquemanuel:
## Description
This PR changes entirely the way we are building and will be building FACOLS.
Before we were using the Oracle docker image provided in Docker Hub, but that had its own caveats and it was not being updated as frequently as we expected and in fact was giving us problems in our CI tests. It was failing on average in 1 out of 5 tests if not more.

For the above reason, we decided to build our own Oracle Image-based out of https://github.com/oracle/docker-images/tree/master/OracleDatabase/SingleInstance information and we even modified a bit the Dockerfile provided to include our own set of scripts and have our schema.

### Process
By altering a few scripts we were able to include the creation of our DDL's (tablespaces, tables, indexes, etc) in the first creation of the DB. This although its a lengthy process, we don't expect to run it very frequently. in fact, we shouldn't do it often unless something changes with those DDL' s and then we need to push a new image to our Docker Repository.

#### Just for testing
If you want to build your own image and use it, you need to use the provided script `build_push.sh` script with the `local` parameter. This will download all the required files and start building the Docker image and tag it accordingly for later use.

#### Local dev
If you want to build your own image to actually do some local development, you should run `bundle exec rake local:build`. This will essentially navigate to the correct locations, build the Docker image, start the `docker-compose` process and then even provision the required DB's including FACOLS.

### Gains
With the above configuration, we gained that our tests no longer have to wait for FACOLS to load, in fact in average it takes between 2 to 5 seconds to load up the entire DB with its required structure and then we can provision the DB, but with gains, comes the caveats.

### Caveats
In order to our image to contain everything including the schema, we removed the Volumes, and some separation, which allowed us to speed up the startup time, but this came with a hefty increase in the size of the image in itself.
Considering that the Oracle compiled image is 10Gb with no DDL, our image with DDL's is 15Gb, so we added 5Gb to the equation and that *does not include data*.
We have an open ticket with Oracle https://github.com/oracle/docker-images/issues/1265 but we are still waiting on a reply to see what we can slim or tear down that is not essential for the DB.

### More information
Our previous FACOLS base image was 3Gb, so we really want to know where the extra . 7/8 GB came from in particular from Oracle, cause ideally we want to slim down our FACOLS image to less than 10Gb.

### Other Changes
Not also we created a new image now, but with the help of @aroltsch we were able to combine the `dev` and `test` databases into one single docker container. This helped our local development since now we have fewer things running in our computer with the same amount of CPU+Ram combination.

In essence, now you can log in to **each of the DB schemas**, so in essence, if you want to log in to FACOLS TEST you can use the `vacols_test` user or `vacols_development` for the FACOLS DEVELOPMENT one.

## To test:
In order to test it, you **will** need to remove all containers and clean up as much space as possible from your docker images, volumes, etc. Its recommended that you do the below commands, if not the build *might* fail.

1. `docker system prune`
2.`bundle exec rake local:build`

## Docker configuration
I would assume that some people might have different configuration and it will depend on their computer. We have tested the above commands and images using the following configuration in Docker.
- Disk Space: 96Gb
- CPU's: 4
- Memory: 4Gb
- Swap: 1Gb


local/vacols/setup_vacols.sql
```diff
@@ -1,4 +1,4 @@
-CONNECT sys/Oradoc_db1@BVAP as sysdba
+CONNECT sys/Oradoc_db1@//localhost:1521/BVAP as SYSDBA;
```
pkarman: does this need to be hardcoded host:port? will that affect how our current 2 instances of vacols works?

local/vacols/vacols_copy_4_triggers.sql
```diff
@@ -37,7 +37,7 @@ ALTER TRIGGER "VACOLS"."HEARSCHED_INSERT" ENABLE;
 
 
   CREATE OR REPLACE TRIGGER "VACOLS"."BRIEFF_BFSO_UPD" AFTER
-UPDATE OF bfso ON BRIEFF FOR EACH ROW
+UPDATE OF bfso ON "VACOLS"."BRIEFF" FOR EACH ROW
```
pkarman: will this hardcoded schema name conflict with the work @aroltsch is doing to combine containers?

local/vacols/buildPushFacols2.sh
```diff
@@ -0,0 +1,11 @@
+#! /bin/sh
+
+eval $(aws ecr get-login --no-include-email --region us-gov-west-1)
```
pkarman: now that we know it works, could we combine or remove the other build script?

local/vacols/setup_vacols.sql
enriquemanuel: we are ideally moving away from 2 instances of vacols into one *soon* with the work from @aroltsch

local/vacols/buildPushFacols2.sh
enriquemanuel: will wait until we know it works correctly? since we might need to build VACOLS differently from the entire Libraries from Oracle Github, so the entire script might actually change

.circleci/config.yml
```diff
@@ -22,7 +22,7 @@ jobs:
             POSTGRES_DB: caseflow_certification_test
 
       # This is our homespun VACOLS container. An oracle db with some special sauce.
-      - image: 008577686731.dkr.ecr.us-gov-west-1.amazonaws.com/facols:version_4
+      - image: 008577686731.dkr.ecr.us-gov-west-1.amazonaws.com/facols2:2
```
aroltsch: Why did you choose this name?

local/vacols/buildPushFacols2.sh
aroltsch: Enrique, let's use underscores in the file to stay consistent.

local/vacols/vacols_copy_4_triggers.sql
aroltsch: Why did you have to make this change here?

aroltsch:
Thank you for tackling this Enrique. This is amazing!

local/vacols/vacols_copy_4_triggers.sql
enriquemanuel: I was doing the changes that I implemented in the DMS recovery task, this can probably be omitted.

Makefile.example
```diff
@@ -42,7 +42,7 @@ ready:
 	h=`docker-compose ps | grep healthy`; while [ "$$h" == "" ]; do h=`docker-compose ps | grep healthy`; echo $$h; sleep 1; done
 
 up:
-	docker-compose up -d
+	cd local/vacols/ && ./build_push.sh local && cd ../../ && docker-compose up -d
```
pkarman: will this re-build the local containers every time? that's not what the original `up` command does.

Cypherscircle (approved):

PR #10636, opened 2019-05-02
Search for user by user pg id instead of css id

aroltsch:
### Description
We call `find_by("UPPER(css_id)=UPPER(?)", css_id)` in the User model on every request. This may or may not be causing intermittent `ActiveRecord::ConnectionTimeoutError`. Let's search by User.id to make it more efficient. 

### Testing Plan
1. Tests pass
2. Ensure logout works
3. Ensure can change users locally in the browser


monfresh:
Do we need to add an index on the Users table?

aroltsch:
Index on the Users.id? I don't think it is needed @monfresh

PR #10642, opened 2019-05-02
Add spacing below flash alert

leikkisa:
connects #9988

Some of the alerts didn't have spacing below them and it looked a little weird.


leikkisa:
Screen Shot 2019-05-02 at 3 03 04 PM

PR #10648, opened 2019-05-03
Sorts appeals so job definitely pulls different appeals each run

### LINKED ISSUE #10319: Bug: Appeals not showing in Schedule Veteran lists, but have open schedule hearing tasks
https://appeals.cf.ds.va.gov/queue/appeals/2850494 -- doesn't appear in Togus, ME list or CO 

https://appeals.cf.ds.va.gov/queue/appeals/3031750 - doesn't appear in Phoenix list or CO 

Previously these cases were in 57, but hadn't gotten the tasks they needed to schedule Veterans. Now they have tasks, but aren't showing up in the lists to schedule, so the branch wouldn't see them to schedule unless they looked for them.

sharonwarner:
Resolves #10319 

### Description
We weren't sorting the appeals before pulling them into the job, so it was possible for appeals with higher IDs to not run for weeks. Now we sort the appeals such that appeals with no ROs are run first, and then appeals that have been updated the least recently go next.


app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -23,13 +23,13 @@ def find_appeals_ready_for_geomatching(appeal_type)
           AND admin_actions.id IS NULL AND t.type = 'ScheduleHearingTask'
           AND t.status NOT IN ('cancelled', 'completed')
         )", appeal_type.name)
-      .where("available_hearing_locations.updated_at < ? OR available_hearing_locations.id IS NULL", 1.week.ago)
+      .order("available_hearing_locations.updated_at nulls first")
```
lomaxap: sometimes i forget how nice postgres is. this is handy.

PR #10650, opened 2019-05-03
Users can create change hearing disposition tasks

### LINKED ISSUE #9403: Change Hearing Dispositions Task
## User Stories
As a Hearings Management branch user, I need to be able to change dispositions after the Daily Docket is locked so that the hearings can move correctly through the tasks needed to complete it.

As a Hearings Management branch administrator, I need to control when dispositions are changed after the Daily Docket is locked. This allows the cases to flow correctly through the assigned tasks and limits the amount of changes made to cases after they are locked.

As a Hearings Management branch administrator, I need to track why dispositions a being changed, investigate corrections, and make accurate updates. 

## Mocks
See the InVision links here: https://va-gov.invisionapp.com/share/CMQHR7EP3Z5

## Engineering Notes
- There's an existing ticket to create the ChangeHearingDispositionTask that should cover much of the acceptance criteria below.
- We need to add Change Disposition as an allowable action from the TranscriptionTask and the NoShowHearingTask. When the user selects that action, we should close the existing task, sibling tasks, and the parent DispositionTask and create a ChangeHearingDispositionTask off of the HearingTask.

## Acceptance Criteria
- [x] Verify the Hearings Management branch users can create a Send for hearing disposition change task under the No Show waiting period task. This creates the task in the Hearings Management queue
- [x] Verify the Transcription team can create a Send for hearing disposition change task under the Transcribe hearing task. This removes the hearing from the transcription queue and creates the task in the Hearings management queue
- [x] Verify Superusers can create the Change hearing disposition task on all tasks that remain in the Hearings Management or Transcription queues (anything with an open Hearing Task). This closes the currently open task and creates the Change hearing disposition task in the Hearings Schedule queue
- [x] Verify Change hearing disposition tasks are visible to all users in the team queue
- [x] Verify only Superusers can assign Change hearing disposition tasks to themselves
- [x] Verify the Change hearing disposition task has the option to place on hold / Assign task to (only other Superusers) / Change hearing disposition
- [x] Verify when the Change hearing disposition option is selected, a modal pops up with the option to change the hearing disposition and leave notes. The modal should also list the hearing date and the current disposition.
- [x] Verify all hearing dispositions are available in the drop down
- [x] Verify once the hearing disposition is saved the disposition is changed in the Case details and in the Daily Docket, Hearing Prep Daily Docket, and Hearing Details
- [x] Verify a Change hearing disposition task is automatically created if a hearing isn't marked with a disposition and is saved by the nightly job
- [x] Verify when the hearing disposition is left blank, it is shown as having a disposition of None in the modal
- [x] Verify when the change is confirmed, the task and notes show in the case timeline

tomas-nava:
Resolves #9403 

### Description
Completes or validates the functionality described in the ticket above:

- Ability to create `ChangeHearingDispositionTask`s from any child task of a HearingTask
- Additional tests for validation


app/models/tasks/disposition_task.rb
```diff
@@ -63,9 +63,13 @@ def update_from_params(params, user)
...
   end
 
-  def create_change_hearing_disposition_task_and_complete
+  def create_change_hearing_disposition_task_and_complete(instructions = nil)
     multi_transaction do
-      ChangeHearingDispositionTask.create!(appeal: appeal, parent: parent)
+      ChangeHearingDispositionTask.create!(
+        appeal: appeal,
+        parent: parent,
+        instructions: instructions.present? ? [instructions] : nil
```
sharonwarner: Why do we store instructions like this? Does it make more sense to send instructions to this method as an array?

tomas-nava: We store them like this because `instructions` is a postgres array column, and we've established a pattern of waiting until we're updating/creating to put it into array form.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -93,15 +93,17 @@ def update_from_params(params, current_user)
   end
 
   def available_actions(user)
+    hearing_admin_actions = available_hearing_admin_actions(user)
```
sharonwarner: I *think* we don't want change hearing disposition tasks to be available from the schedule hearing task yet. This parent hearing task won't have any active disposition tasks, so I think the logic will just fail silently in create_change_hearing_disposition_task_and_complete_children.

PR #10654, opened 2019-05-06
Add VBMS::ShareExceptionFindRatingData

pkarman:
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3467/


app/exceptions/vbms_error.rb
```diff
@@ -26,6 +26,7 @@ class CannotDeleteContention < Caseflow::Error::VBMS; end
   class ClaimDateInvalid < Caseflow::Error::VBMS; end
   class FilenumberDoesNotExist < Caseflow::Error::VBMS; end
   class MissingData < Caseflow::Error::VBMS; end
+  class ShareException < Caseflow::Error::VBMS; end
```
leikkisa: I think this class name is too generic for the error message, because ShareExceptions I believe happen for a bunch of other things as well (probably anything that the Share system doesn't like).  

I think including findRatingData in the name would be more appropriate.

Issue #9901, opened 2019-03-12
Refactor / Document how we're handling time


lomaxap:
We currently handle time when creating hearings in a couple of different ways.

On the daily dockets ([hearing sched](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/hearingSchedule/containers/DailyDocketContainer.jsx#L176) / [hearings controller](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/hearings_controller.rb#L13)). In the [Assign Hearing modal ](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/components/AssignHearingModal.jsx#L267)/ [slot_new_hearing](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/repositories/hearing_repository.rb#L98) method. (any where else?)

This has introduced a number of bugs, particularly around timezone and daylight saving time shifts. We should document how a hearing's time is stored and accessed in VACOLS vs. caseflow and create a consistent method for handling both, if it doesn't exist already.

--------------

We should update the hearing model to store times as strings.

nicholasholtz:
@sharonwarner @pshahVA 

This is an example of a docket where the times are way off. I note that this is legacy. Are these cases that were scheduled through Caseflow? Is there a way to know? 

https://appeals.cf.ds.va.gov/hearings/schedule/docket/282

(cc @lomaxap @laurjpeterson )

nicholasholtz:
Another example - https://appeals.cf.ds.va.gov/hearings/schedule/docket/282

sharonwarner:
@nicholasholtz anything scheduled on a caseflow hearing day (after 4/1 for video and 1/1 for co) had to have been scheduled for caseflow, so the docket you linked to above was definitely scheduled through caseflow.

nicholasholtz:
Ah, of course. Thanks.

PR #10660, opened 2019-05-06
Hearing Time Refactor

lomaxap:
Connects #9901

- moves all time logic to backend service (includes conversion to CO timezone, and reformatting for vacols)
- adds `scheduled_time_string` accessor to legacy/hearings
- adds `build_legacy_params_with_time` and `build_params_with_time` methods for reformatting update request parameters in `hearings_controller`


app/services/hearing_time_service.rb
```diff
@@ -0,0 +1,86 @@
...
+
+    def vacols_formatted_datetime(scheduled_for:, scheduled_time_string:)
+      hour, min = scheduled_time_string.split(":")
+
+      time = scheduled_for.to_datetime.change(
+        hour: hour.to_i,
+        min: min.to_i
+      )
+
+      Time.utc(time.year, time.month, time.day, time.hour, time.min, time.sec)
```
lomaxap: I'm not sure why in the [`VacolsHelper`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/helpers/vacols_helper.rb#L19) we're casting it to ET before UTC?

pkarman: iirc there's some offset trickery done to account for how VACOLS stores times (in ET but labeled as UTC) and how Rails thinks it stores times (in UTC). The specs appear to have changed since I last touched this but I left this comment originally. https://github.com/department-of-veterans-affairs/caseflow/pull/9477/files#r260494293

lomaxap: ahh. interesting. this bug makes a lot more sense then.

What I think the cause of this bug is FTR: we're using the central office time (`scheduled_for` in table below) to display the [hearing time](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/hearingSchedule/components/DailyDocketRowDisplayText.jsx#L42-L54) without changing it back to local time. 

For this PR, I'm assuming in both VACOLS and Caseflow time is stored in the DB in local time but encoded UTC, we simply need to get the hour and minute from that time object and pass it along to the client as a string. 

|hearing.scheduled_for|hearing.vacols_record.hearing_date|timezone|
|:-------------------:|:--------------------------------:|:------:|
|2017-05-26 11:30:00 -0400|2017-05-26 08:30:00 UTC|America/Los_Angeles|
|2017-04-25 11:30:00 -0400|2017-04-25 11:30:00 UTC|America/New_York|
|2015-11-05 10:30:00 -0500|2015-11-05 08:30:00 UTC|America/Denver|
|2012-09-10 01:00:00 -0400|2012-09-10 00:00:00 UTC|America/Chicago|
|2019-06-27 11:30:00 -0400|2019-06-27 08:30:00 UTC|America/Boise|
|2014-04-15 12:30:00 -0400|2014-04-15 12:30:00 UTC|America/Puerto_Rico|

Issue #8693, opened 2019-01-14
Allow BVA dispatch to return cases to Judges


lowellrex:
Board dispatch regularly needs to send cases back to judges for corrections. This ticket exists to present them with an option--[similar to the one members of the Quality Review team have](https://github.com/department-of-veterans-affairs/caseflow/issues/7462)--to send cases back to the signing judge.

## Acceptance criteria
* [x] Board dispatchers can select an option in the task actions dropdown on the case details page to send a case back to the signing judge. This option is called "Return to judge"
* [x] Board dispatchers can write a note with instructions about the task, so judges know what they need to do with it
* [x] When this task arrives in the judges queue, the task is called "Revise decision (BVA Dispatch)"

laurjpeterson:
With this github issue, judges will be able to distinguish these tasks from others.
https://github.com/department-of-veterans-affairs/caseflow/issues/8652

laurjpeterson:
@allyceh - based on your work with BVA dispatchers, do you have a suggestion for what this task is called? I have "Correct decision data" but maybe it should be more broad?

allyceh:
Assuming this task would be judge facing // appear in the judges queue, how about "Revise decision (BVA Dispatch)"? In our refreshed task naming ticket https://github.com/department-of-veterans-affairs/caseflow/issues/7508 , I proposed "Revise decision (quality review)" for when member of the QR team need judges to make a change. This task name would follow the same schema. 

For the action available to BVA Dispatch, I'm thinking "Return to judge" would be clear to them. Or, will BVA Dispatch use the "Assign to" option and include notes to the judge?

laurjpeterson:
I love it!

lomky:
Should this task then be able to have a corresponding Attorney subtask, or do Judges handle this directly?

lowellrex:
MEDIUM. Make sure we have adequate test coverage here since this is a new workflow in Queue for BVA Dispatchers. Will need to create a new task class (something like `JudgeDispatchReturnTask`) a la `JudgeQualityReviewTask`.

lowellrex:
> Should this task then be able to have a corresponding Attorney subtask, or do Judges handle this directly?

Yup! I think that makes sense!

lomky:
A couple questions @laurjpeterson. I provided `Quality Review`'s current behavior for reference.

 - Should these `Dispatch Return` tasks end up in the judge's Review pane or their Assign pane?
[...]

PR #10670, opened 2019-05-06
Allow Dispatch to Return Cases to the Judge

lomky:
This PR implements a new flow allowing BVA Dispatch users to return an Appeal to its Judge. The judge may then either act on the task or assign it on to an Attorney.

  - [x] New Judge task `JudgeDispatchReturnTask` or "Revise decision (BVA Dispatch)"
  - [x]  Judge can return task to the Attorney that worked on the decisions as an `AttorneyRewriteTask`

Working on #8693
 
### Description

### Acceptance Criteria
* [x] Board dispatchers can select an option in the task actions dropdown on the case details page to send a case back to the signing judge. This option is called "Return to judge"
* [x] Board dispatchers can write a note with instructions about the task, so judges know what they need to do with it
* [x] When this task arrives in the judges queue, the task is called "Revise decision (BVA Dispatch)"

### Testing Plan
1. Go to ...


app/models/queues/attorney_queue.rb
```diff
@@ -19,9 +19,9 @@ def tasks
       end
     end
 
-    caseflow_tasks = Task.includes(*task_includes).incomplete_or_recently_closed
-      .where(assigned_to: user, type: [AttorneyTask.name, AttorneyRewriteTask.name, QualityReviewTask.name])
-    (colocated_tasks_for_attorney_tasks + caseflow_tasks).flatten
+    caseflow_tasks_attorney = AttorneyTask.includes(*task_includes).incomplete_or_recently_closed
+    caseflow_tasks_quality_review = QualityReviewTask.includes(*task_includes).incomplete_or_recently_closed
+    (colocated_tasks_for_attorney_tasks + caseflow_tasks_attorney + caseflow_tasks_quality_review).flatten
```
lowellrex: Why do we need to define which types of tasks to return here? Can we do something like `user.tasks.includes(*task_includes).incomplete_or_recently_closed`? Is there some reason we want to exclude the judge tasks and generic tasks here? This isn't a change in behaviour, just something I noticed just now.

| count | type
---| ---
639 | AttorneyTask
73 | QualityReviewTask
57 | AttorneyRewriteTask
8 | JudgeDecisionReviewTask
5 | GenericTask
5 | JudgeAssignTask

https://caseflow-looker.va.gov/sql/fn4kwcwxgrq8t7

```sql
select count(*)
, tasks.type
from tasks
join organizations_users
  on organizations_users.user_id = tasks.assigned_to_id
join organizations
  on organizations.id = organizations_users.organization_id
where organizations.type = 'JudgeTeam'
  and organizations_users.admin = 'false'
  and tasks.assigned_to_type = 'User'
group by 2
order by 1 desc;
```

app/models/tasks/bva_dispatch_task.rb
```diff
@@ -5,6 +5,10 @@
 # This indicates that an appeal is decided and the appellant is about to be notified of the decision.
 
 class BvaDispatchTask < GenericTask
+  def available_actions(user)
+    super.unshift(Constants.TASK_ACTIONS.DISPATCH_RETURN_TO_JUDGE.to_h)
```
lowellrex: Will this give users who are not members of the `BvaDispatch` team the ability to return the task to the judge? The [default set of actions](https://github.com/department-of-veterans-affairs/caseflow/blob/003a610fcc6375324df17320ba9185e446fe137c/app/models/tasks/generic_task.rb#L62) for `GenericTask`s that are not assigned to the current user is an empty array, are we able to avoid adding this to that set of actions some other way?

app/workflows/complete_case_review.rb
```diff
@@ -31,7 +31,8 @@ def review_class
   def create_quality_review_task
     return if case_review.appeal.is_a?(LegacyAppeal) ||
               !case_review.is_a?(JudgeCaseReview) ||
-              case_review.task.parent.is_a?(QualityReviewTask)
+              case_review.task.parent.is_a?(QualityReviewTask) ||
+              case_review.task.parent.is_a?(BvaDispatchTask)
```
lowellrex: Can we add a comment to the code that explains why we are returning early here? I think we're doing this because we can end up posting to the `/case_reviews/:task_id/complete` endpoint after we've already created a QR or Dispatch task.

client/app/queue/QueueApp.jsx
```diff
@@ -333,7 +333,13 @@ class QueueApp extends React.PureComponent {
             path={`/queue/appeals/:appealId/tasks/:taskId/${TASK_ACTIONS.REASSIGN_TO_PERSON.value}`}
             render={this.routedReassignToUser} />
           <Route
-            path={`/queue/appeals/:appealId/tasks/:taskId/${TASK_ACTIONS.RETURN_TO_JUDGE.value}`}
+            path={`/queue/appeals/:appealId/tasks/:taskId/${TASK_ACTIONS.QR_RETURN_TO_JUDGE.value}`}
+            render={this.routedAssignToUser} />
+          <Route
+            path={`/queue/appeals/:appealId/tasks/:taskId/${TASK_ACTIONS.DISPATCH_RETURN_TO_JUDGE.value}`}
+            render={this.routedAssignToUser} />
```
lowellrex: Excellent re-use of this function!

app/models/tasks/bva_dispatch_task.rb
lomky: Will add in a check similar to how GenericTask does. 

Would we want to restrict the Return to Judge capability to just the assigned BvaDispatch user, or anyone in the BvaDispatch org?

app/models/queues/attorney_queue.rb
lomky: @laurjpeterson Can you think of any reason why a task of Judge*Task or GenericTask type that is assigned to an attorney _shouldn't_ show up in their queue?

app/models/tasks/bva_dispatch_task.rb
lowellrex: Good question! Will defer to @laurjpeterson on this one.

app/models/queues/attorney_queue.rb
laurjpeterson: Why would a GenericTask ever be put in an attorney's queue?

app/models/tasks/bva_dispatch_task.rb
laurjpeterson: let's allow the assigned dispatch user, and the dispatch team admin

app/models/queues/attorney_queue.rb
lomky: I don't know, but if it is assigned to them and in their queue, wouldn't it be better for them to see it than not?

laurjpeterson: i agree with you and lowell:

> I'm in favor of showing all task types for attorneys and then limiting them if we receive any complaints

spec/feature/queue/task_queue_spec.rb
```diff
@@ -633,6 +643,135 @@ def validate_pulac_cerullo_tasks_created(task_type, label)
...
+        OrganizationsUser.add_user_to_organization(FactoryBot.create(:user), Colocated.singleton)
+      end
+
+      it "should display an option of Ready for Dispatch" do
+        expect(bva_dispatch_person_task.reload.status).to eq(Constants.TASK_STATUSES.on_hold)
+
+        find(".Select-control", text: "Select an action…").click
+        find("div", class: "Select-option", text: Constants.TASK_ACTIONS.JUDGE_AMA_CHECKOUT.label).click
+
+        expect(page).to have_content("Add decisions")
```
lowellrex: Let's use COPY.json where we can. `COPY::DECISION_ISSUE_PAGE_TITLE` should work here.

spec/feature/queue/task_queue_spec.rb
```diff
@@ -633,6 +643,135 @@ def validate_pulac_cerullo_tasks_created(task_type, label)
...
+        visit("/queue/appeals/#{appeal.external_id}")
+
+        # Add a user to the Colocated team so the task assignment will suceed.
+        OrganizationsUser.add_user_to_organization(FactoryBot.create(:user), Colocated.singleton)
+      end
+
+      it "should display an option of Ready for Dispatch" do
+        expect(bva_dispatch_person_task.reload.status).to eq(Constants.TASK_STATUSES.on_hold)
+
+        find(".Select-control", text: "Select an action…").click
```
lowellrex: `COPY::TASK_ACTION_DROPDOWN_BOX_LABEL`

spec/feature/queue/task_queue_spec.rb
```diff
@@ -633,6 +643,135 @@ def validate_pulac_cerullo_tasks_created(task_type, label)
...
+        click_on "Continue"
+
+        expect(page).to have_content(COPY::JUDGE_CHECKOUT_DISPATCH_SUCCESS_MESSAGE_TITLE % appeal.veteran_full_name)
+
+        expect(judge_task.reload.status).to eq(Constants.TASK_STATUSES.completed)
+        expect(bva_dispatch_person_task.reload.status).to eq(Constants.TASK_STATUSES.assigned)
+      end
+
+      it "should display an option to Return to Attorney" do
+        click_dropdown(prompt: "Select an action", text: "Return to attorney")
```
lowellrex: Maybe we can use `COPY::TASK_ACTION_DROPDOWN_BOX_LABEL` and `Constants.TASK_ACTIONS.JUDGE_QR_RETURN_TO_ATTORNEY.label` here?

spec/feature/queue/task_queue_spec.rb
```diff
@@ -633,6 +643,135 @@ def validate_pulac_cerullo_tasks_created(task_type, label)
...
+        expect(dropdown_selected_value(find(".cf-modal-body"))).to eq attorney_user.full_name
+        fill_in "taskInstructions", with: "Please fix this"
+
+        click_on "Submit"
```
lowellrex: `COPY::MODAL_SUBMIT_BUTTON`

spec/feature/queue/task_queue_spec.rb
```diff
@@ -633,6 +643,135 @@ def validate_pulac_cerullo_tasks_created(task_type, label)
...
+
+        expect(page).to have_content("Task assigned to #{attorney_user.full_name}")
```
lowellrex: `expect(page).to have_content(ASSIGN_TASK_SUCCESS_MESSAGE % attorney_user.full_name)`

spec/feature/queue/task_queue_spec.rb
lowellrex: `JUDGE_DISPATCH_RETURN_TO_ATTORNEY` rather.

spec/feature/queue/task_queue_spec.rb
```diff
@@ -633,6 +643,135 @@ def validate_pulac_cerullo_tasks_created(task_type, label)
...
+
+        expect(judge_task.reload.status).to eq(Constants.TASK_STATUSES.on_hold)
+        expect(judge_task.children.first).to be_a(AttorneyDispatchReturnTask)
+        expect(judge_task.children.first.status).to eq(Constants.TASK_STATUSES.assigned)
```
lowellrex: Can we do an end to end test here (can be in a new file and [use `step()`](https://github.com/department-of-veterans-affairs/caseflow/blob/13a2cebc4e244cd475160c0b72030ad5a23598dc/spec/support/feature_helper.rb#L57))?

* Task in BVA dispatch user's queue
* Sends task to judge
* Judge sends it to attorney
* Attorney goes through checkout
* Judge goes through checkout
* Case is actionable in BVA dispatcher's queue and case details page has case timeline that reflects the above tasks have been completed.

If we'd like to get this PR merged and write a ticket to add the test coverage above in a different PR, that makes sense too.

Issue #10546, opened 2019-04-25
Do not show removed reviews in search results/case list


laurjpeterson:
When a user has removed all issues from an appeal/HLR/SC and removed the review entirely, we should not show the appeal/HLR/SC in Caseflow's search results, so we don't confuse users that the appeal is still relevant.

## AC
- Removed appeals/HLRs/SCs do not show in Caseflow search results (or Caseflow case list)

leikkisa:
@laurjpeterson I'm assuming the "Caseflow search results" and the "Caseflow case list" are the same thing, but let me know if there are two places I should check.

laurjpeterson:
@leikkisa - yep, they're the same thing!

Sjones352:
**Fails UAT**

This file had 2 non-ratings removed but when I search the case, I still find it 
![Screen Shot 2019-05-10 at 11.48.55 PM.png](https://images.zenhubusercontent.com/5b8fddd1b2cc7b6c78ece9f9/1491c608-6576-4803-816f-a580dd98b10b)![Screen Shot 2019-05-10 at 11.49.15 PM.png](https://images.zenhubusercontent.com/5b8fddd1b2cc7b6c78ece9f9/d73068b0-1936-4a9b-baf7-26ecd59211b5)

leikkisa:
@Sjones352 This Veteran has 6 HLRs, and one of them is not showing up (probably the removed one), so I think the issue is the validation approach.

I tried this on the same Veteran with a Supplemental Claim, and it showed up on the Case Search page after creating it (there were two Supplemental Claims) and disappeared after removing all of the issues.

Sjones352:
**Passes UAT*
https://appeals.cf.uat.ds.va.gov/cases/1604. The Supplemental Claim does not show up on search page

![Screen Shot 2019-05-14 at 3.12.05 PM.png](https://images.zenhubusercontent.com/5b8fddd1b2cc7b6c78ece9f9/97d09a43-f4c5-4947-a9c6-f09d17b0af03)
![Screen Shot 2019-05-14 at 3.13.26 PM.png](https://images.zenhubusercontent.com/5b8fddd1b2cc7b6c78ece9f9/17045f93-81bd-4488-bc87-e2b073265c69)
[...]

PR #10673, opened 2019-05-07
Do not show removed reviews in case search

leikkisa:
connects #10546


app/models/claim_review.rb
```diff
@@ -33,9 +33,10 @@ def find_by_uuid_or_reference_id!(claim_id)
       claim_review
     end
 
+    # Does not include decision reviews where all request issues were removed
```
leikkisa: Added this comment because I couldn't think of a better name for the method that indicated "find_all_that_are_not_removed" :-P.  I am open to suggestions!

pkarman: maybe `find_all_visible_by_file_number` ?

leikkisa: You are so great at naming! Thank you, I like that.

PR #10675, opened 2019-05-07
Only make 1 request for doc counts from efolder

lowellrex:
Aims to reduce the number of requests queue sends to efolder which I believe will are at the root of the flash outages Caseflow has been having the past few months.

## Why will this help
The queue table view makes a request to the `/appeals/#id/document_count` endpoint for each row (appeal) that loads. This sends a request to efolder to download the documents associated with a case file. After efolder starts this initial download caseflow polls efolder every second for 5 minutes to see if that download has finished. If the download finishes (or has previously finished and is cached in efolder) then caseflow will count the number of documents and return that count to the front-end.

This change reduces the number of retries specifically for the `document_count` endpoint, so that we attempt to get the number of documents only once each time we hit that endpoint and display whatever we return. At the beginning of the day (when documents are not yet cached in efolder) this will return an error and display "Failed to Load" on the queue table view (and no number on the case details page, the only other place we make this request) but it will also trigger efolder to download and cache the documents. Every subsequent queue table view should get a response back from efolder that includes accurate document counts because they have already been fired off by the earlier requests. And since we reload the entire queue table view quite regularly (due to react router issues, I think) I think the visibility to the Caseflow users will be minimal.

## User impact
At the beginning of the day document counts on the queue table view will display "Failed to Load" more frequently than they do currently (currently they will usually read "Loading number of docs..." for several minutes and then display the document count).


app/services/external_api/efolder_service.rb
```diff
@@ -3,18 +3,21 @@
 require "json"
 
 class ExternalApi::EfolderService
-  TRIES = 300
+  def self.quick_document_count_for_appeal(appeal, user)
+    response = generate_efolder_request(appeal.veteran_file_number.to_s, user, 1)
+    response[:documents]&.size
```
aroltsch: If `response[:documents]` is empty, it will return 0 and display View 0 Docs?

lowellrex: Either `0` or `nil` I assume. It will continue to display whatever we previously displayed in that case. I don't think this is a change in behaviour.

PR #10676, opened 2019-05-07
PA was routed incorrectly

### LINKED ISSUE #7366: Private Attorney Special Issue
I need to do some more research on this, but when we pull appeals from VACOLS, we set private_attorney_or_agent equal to true if case_record.bfso == "T" and false otherwise. I believe that introduced a bug such that if a user selects private attorney to true on the special issues page, but case_record.bfso != "T", the special issue gets reset back to false the next time we try to pull the appeal from VACOLS.

Relevant PR: https://github.com/department-of-veterans-affairs/caseflow/pull/2365

sharonwarner:
Resolves https://github.com/department-of-veterans-affairs/appeals-support/issues/4556
Resolves #7366 

### Description
PA was previously using default routing, but it should be routed to the appeal's original RO. We were also not accurately persisting updates to PA because we overwrote any updated values from VACOLS. This PR removes that integration.


pkarman:
@sharonwarner did we want to merge this?

PR #10679, opened 2019-05-07
Lowell/reduce efolder requests from queue doc count

youngfreezyVA:
updates failing tests in master


youngfreezyVA:
> can't quite tell from the GH diff but maybe some extra whitespace on that last patch? in any case, if the linter is ok, I'm ok.
> thanks!

yeah there is! running the lint command now and pushing

PR #10682, opened 2019-05-07
Allow edit on hearings details for legacy hearing

### LINKED ISSUE #10403: Hearing details | Edit judge for legacy hearings
Currently we only allow users to edit a hearing record from the hearing details page for AMA hearings. We'd like to add functionality such that users can edit the judge at an individual hearing level from the hearing details page for legacy hearings.

AC:
- [ ] Users can edit an individual hearing's judge from the hearing details page for a legacy hearing

Current page view:

Screen Shot 2019-04-11 at 11 02 43 AM

New page view:

Engineering notes:
- The frontend code for this logic can be found in `client/app/hearingSchedule/components/Details.jsx`
- The backend code for this logic can be found in `app/controllers/hearings_controller.rb`
- We'll need to ensure we're sending the vacols_attorney_id to VACOLS, not the judge_id from the caseflow database

ferristseng:
Resolves #10403

### Description

Removes the legacy warning on the hearing details page, and makes the judge, room, and hearing coordinator editable on the page. Also changes the mapper to map the Caseflow user attorney id to the judge id.

Also removes `vacols_attributes` from the legacy hearings model because it is unused.

### Acceptance Criteria
- [x]  Users can edit an individual hearing's judge from the hearing details page for a legacy hearing

### DIFF
--- a/spec/feature/hearing_schedule/hearing_details_spec.rb
+++ b/spec/feature/hearing_schedule/hearing_details_spec.rb
@@ -5,6 +5,11 @@ require "rails_helper"
 RSpec.feature "Hearing Schedule Daily Docket" do
   let(:user) { create(:user, css_id: "BVATWARNER", roles: ["Build HearSched"]) }
 
+  before do
+    create(:staff, sdept: "HRG", sactive: "A", snamef: "ABC", snamel: "EFG")
+    create(:staff, svlj: "J", sactive: "A", snamef: "HIJ", snamel: "LMNO")
+  end
+
   context "Hearing details is not editable for a non-hearings management user" do
     let!(:current_user) { User.authenticate!(user: user) }
     let!(:hearing) { create(:hearing, :with_tasks) }
@@ -21,11 +26,6 @@ RSpec.feature "Hearing Schedule Daily Docket" do
       User.authenticate!(user: user)
     end
     let!(:hearing) { create(:hearing, :with_tasks) }
-
-    before do
-      create(:staff, sdept: "HRG", sactive: "A", snamef: "ABC", snamel: "EFG")
-      create(:staff, svlj: "J", sactive: "A", snamef: "HIJ", snamel: "LMNO")
-    end
 
     scenario "User can update fields", skip: "Test is flakey" do
       visit "hearings/" + hearing.external_id.to_s + "/details"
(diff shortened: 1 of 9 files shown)


spec/feature/hearing_schedule/hearing_details_spec.rb
```diff
@@ -62,9 +62,29 @@
...
+    scenario "User can edit Judge" do
       visit "hearings/" + legacy_hearing.external_id.to_s + "/details"
-      expect(page).to have_content("This is a Legacy Case Hearing")
+
+      expect(page).to have_field("judgeDropdown", disabled: false)
+      expect(page).to have_field("hearingCoordinatorDropdown", disabled: false)
+      expect(page).to have_field("hearingRoomDropdown", disabled: false)
+      expect(page).to have_field("Notes", disabled: false)
+      expect(page).to have_no_selector("label", text: "Yes, Waive 90 Day Evidence Hold")
+    end
```
lomaxap: We should test that they can edit the field and save the changes. You can see an example [here](https://github.com/department-of-veterans-affairs/caseflow/blob/1bd0691b63dbbeeb0592d46246a37bc96b28382c/spec/feature/hearing_schedule/hearing_details_spec.rb#L45-L54). There are a few nice utilities like `fill_in`,  `click_dropdown`, and `click_button`.

spec/feature/hearing_schedule/hearing_details_spec.rb
```diff
@@ -62,9 +62,29 @@
...
+      expect(page).to have_no_field("taskNumber")
+      expect(page).to have_no_field("transcriber")
+      expect(page).to have_no_field("sentToTranscriberDate")
+      expect(page).to have_no_field("expectedReturnDate")
+      expect(page).to have_no_field("uploadedToVbmsDate")
+      expect(page).to have_no_field("problemType")
+      expect(page).to have_no_field("problemNoticeSentDate")
+      expect(page).to have_no_field("requestedRemedy")
+      expect(page).to have_no_field("copySentDate")
+      expect(page).to have_no_field("copyRequested")
```
lomaxap: same comment as above

ferristseng: I think I might be able to skip those checks for this test because this is testing if those input fields don't exist.

Issue #10294, opened 2019-04-03
Updating address in VBMS


leikkisa:
https://dsva.slack.com/archives/CFBLH2LR4/p1554318719015200

A user at VBMS UAT reported to that if the address is updated in VBMS before the claim is established in Caseflow, Caseflow will revert the address back to the old address and change the address in VBMS.  However, if the claim is established in Caseflow and then the address is updated in VBMS, then the updated address will stay.

This was noticed during UAT testing as well.

#### Acceptance criteria
- If a user starts an intake and then updates the Veteran's address in VBMS before the claim is established, the Veteran's address should be refreshed when establishing the claim.  
- The address in VBMS should not get changed back to the previous address when establishing the claim.

leikkisa:
According to BGS when an address is updated in VBMS it is also immediately updated in BGS, so we should make sure we refresh the Veteran's address before establishing a claim.

leikkisa:
*Validation instructions*
There are two situations worth testing:

User updates address while intake is in progress (after intake was started)
1. Start an intake that is processed in VBMS (HLR or SC, Compensation)
2. Change the Veteran's address in VBMS
3 Complete the intake

User updates address after intake is completed
For a different Veteran:
1. Complete an intake that is processed in VBMS
2. Change the Veteran's address in VBMS

The next day (or at least after the End Product Establishments have completed processing), check the Veteran's address in VBMS.  It should be the new, updated address, not the previous address.

Sjones352:
**User updates address while intake is in progress (after intake was started)**

File Number `700060407`
<img width="1445" alt="Screen Shot 2019-05-14 at 8 15 28 PM" src="https://user-images.githubusercontent.com/23080951/57740519-45304f80-7686-11e9-8006-3d5bbd9263dc.png">

<img width="1164" alt="Screen Shot 2019-05-14 at 8 22 25 PM" src="https://user-images.githubusercontent.com/23080951/57740522-45c8e600-7686-11e9-9f7b-8dfdc44f1947.png">
<img width="1187" alt="Screen Shot 2019-05-14 at 8 20 35 PM" src="https://user-images.githubusercontent.com/23080951/57740521-45c8e600-7686-11e9-8c06-f1af162363f6.png">

<img width="1228" alt="Screen Shot 2019-05-14 at 8 23 07 PM" src="https://user-images.githubusercontent.com/23080951/57740523-45c8e600-7686-11e9-949c-ebb5742a1931.png">

[...]

PR #10683, opened 2019-05-07
Refresh Veteran address before establishing claim

leikkisa:
connects #10294

When a user updates a Veteran's address in VBMS before a claim is successfully established, we are sometimes sending the old address to BGS when establishing the claim, which surprisingly causes the address to erroneously get reverted to the old address in VBMS.  This is an attempt to prevent that.


leikkisa:
I did not notice the Veteran record getting cached, but I would love any insight if I'm missing another place where we might be storing an old address for a Veteran.

aroltsch:
I think your idea of refreshing the veteran record makes sense.

app/models/veteran.rb
```diff
@@ -325,6 +325,10 @@ def alive?
     !deceased?
   end
 
+  def unload_bgs_record
+    instance_variable_set(:@bgs_record_loaded, false)
```
pkarman: does it not work to just do `@bgs_record_loaded = false` ?

Issue #8892, opened 2019-01-23
Intake | Legacy Opt-in copy change on edit page.


shanear:
For any issues on the edit page that have legacy opt-ins, the copy reads as if the issue is going to reclose the VACOLS issue. Instead it should probably say something like:

"This issue has automatically closed the VACOLS issue"
![ss5](https://user-images.githubusercontent.com/1596259/51627121-08352700-1ef6-11e9-94b4-35f04e3f043f.PNG)

laurjpeterson:
@aroltsch to take a look re: code climate issues

leikkisa:
I am closing this, because it is just a copy change, and we don't currently have any legacy issue opt-ins in UAT to make this easy.

PR #10685, opened 2019-05-08
Update legacy opt in copy on edit

leikkisa:
connects #8892


client/constants/INTAKE_STRINGS.json
```diff
@@ -1,4 +0,0 @@
-{
```
aroltsch: Glad we moved these constants into one place

pkarman:
@leikkisa this is failing Circle pretty dramatically -- any plans to fix?

leikkisa:
@pkarman Yes, I am planning on fixing it when I have a chance.  It was passing when it got reviewed and approved, but then I started trying to fix a Code Climate thing about cognitive complexity.  Running Code Climate on the file locally didn't work for me, and also my tests in general broke last week but started working again yesterday after I updated to the new FACOLS.  The code climate refactor is for the AddedIssue eligibility check which I agree was cognitively complex ;-).  So I am waiting until I have a chance to spend some time on it.

If I could skip Code Climate then I could revert some commits and merge it.

Issue #10596, opened 2019-04-29
New column for edited contention text


pkarman:
Step 2 of editing request issue title. In Step 1 we added the UI element. In this step, we add the database column and server-side changes.

## Definitions

* "original text" here means the generated title text derived from the contested rating issue description.
* "edited text" here means the user-entered title text

## Acceptance criteria

* When a new request issue description is sent to the server, save it to a new column on the `RequestIssue` model.
* When creating contentions prefer the edited text if it is defined.

## Technical notes

* Add a new column `rating_issue_description` for storing the user's edited title. Preserve the `contested_issue_description` for the original text, so that the user can "undo" their edits.
* The `description` method in `RequestIssue` should prefer `rating_issue_description` over `contested_issue_description` if `rating_issue_description` is not nil.
* Use a feature toggle `edit_contention_text`.
* For now working only on Claim Reviews.

leikkisa:
When a nonrating issue gets contested again, it shows up as a contestable issue in the same modal as rating issues, which I believe will have the ability to edit the description.

How about naming this updated_description?

pkarman:
How about `user_description` to indicate it was manually edited? Or `manual_description`? or `edited_description`?

`updated_description` implies a timeliness or something SQL-esque which confuses me a little.

Sjones352:
I think I like `edited_description` as it indicates description has been edited

PR #10686, opened 2019-05-08
New column for edited contention text

Sjones352:
Connects #10596


db/migrate/20190508001228_add_rating_issue_description_to_request_issues.rb
```diff
@@ -0,0 +1,5 @@
+class AddRatingIssueDescriptionToRequestIssues < ActiveRecord::Migration[5.1]
+  def change
+    add_column :request_issues, :rating_issue_description, :string, comment: "Description of the edited contested rating or decision issue. Will be either a rating issue's decision text or a decision issue's description."
```
leikkisa: How about this?
"The edited description for the contested issue, optionally entered by the user."

This part is not correct in either case: Will be either a rating issue's decision text or a decision issue's description.

app/models/request_issue.rb
```diff
@@ -237,6 +237,7 @@ def description
     return contested_issue_description if contested_issue_description
     return "#{nonrating_issue_category} - #{nonrating_issue_description}" if nonrating?
     return unidentified_issue_text if is_unidentified?
+    return rating_issue_description if !rating_issue_description.nil?
```
leikkisa: Per the spec, this would actually be at the top of this method since it's preferred.

Oh boy, I'm seeing some issues with the spec for this issue, because once a nonrating issue is decided, it shows up as a contestable issue in the same place as the rating issues and I believe can have the description edited.  I will comment in the ticket.

client/app/intake/pages/addIssues.jsx
```diff
@@ -184,7 +184,7 @@ export class AddIssuesPage extends React.Component {
                   }
                 </div>
               </div>
-              {editContentionText && <EditContentionTitle
```
Sjones352: If I do !issue.category, the Edit Contention Title button is hidden for higher level Review and Supplemental claims including the appeal.

app/models/request_issue.rb
```diff
@@ -234,6 +234,7 @@ def closed?
   end
 
   def description
+    return edited_description if edited_description
```
pkarman: we might need this to be `return edited_description if edited_description.present?` 

https://blog.appsignal.com/2018/09/11/differences-between-nil-empty-blank-and-present.html

The reason is that if the user edits the description, then edits again to the empty string `""` we do not want to send an empty string to VBMS. Ruby differentiates between `nil` and `""` so `.present?` does what we intend.

client/app/intake/pages/addIssues.jsx
```diff
@@ -184,7 +184,7 @@ export class AddIssuesPage extends React.Component {
...
+              {editContentionText && formType !== FORM_TYPES.APPEAL.key && issue.category && <EditContentionTitle
```
pkarman: maybe encapsulate this boolean logic into something easier to read?

```js
const editableContentionText = formType !== FORM_TYPES.APPEAL.key && issue.category ? true : false;
{editContentionText && editableContentionText && <EditContentionTitle
```

db/migrate/20190508215805_add_edited_description_to_request_issues.rb
```diff
@@ -0,0 +1,5 @@
+class AddEditedDescriptionToRequestIssues < ActiveRecord::Migration[5.1]
+  def change
+    add_column :request_issues, :edited_description, :string, comment: "The updated description for the contested issue, optionally entered by the user."
```
pkarman: maybe replace "updated" with "edited" in the comment too?

spec/models/request_issue_spec.rb
```diff
@@ -641,6 +644,12 @@
       it { is_expected.to eq("a rating request issue") }
     end
 
+    context "when description is updated" do
+      let(:request_issue) { rating_request_issue }
+      let(:edited_description) { "edited description" }
+      it { is_expected.to eq(edited_description) }
+    end
```
pkarman: could add a test here for when `let(:edited_description) { "" }`

client/app/intake/pages/addIssues.jsx
leikkisa: Is that also true for rating issues?

client/app/intake/pages/addIssues.jsx
```diff
@@ -152,6 +152,8 @@ export class AddIssuesPage extends React.Component {
       return <div className="issues">
         <div>
           { requestIssues.map((issue) => {
+            const editableContentionText = Boolean(formType !== FORM_TYPES.APPEAL.key && issue.category);
```
leikkisa: This should be !issue.category

spec/models/request_issue_spec.rb
```diff
@@ -641,6 +644,21 @@
...
+
+    context "when description returns empty string" do
+      let(:request_issue) { rating_request_issue }
+      let(:edited_description) { "" }
+
+      it "should return an empty string" do
+        expect(edited_description).to eq("")
```
leikkisa: This should be 
```it { is_expected.to eq("a rating request issue") }```

File: spec/feature/intake/higher_level_review/edit_spec.rb [hunk 10]
```diff
@@ -1443,6 +1443,33 @@ def click_cancel(visit_page)
...
+      scenario "edit contention text" do
+        visit "higher_level_reviews/#{higher_level_review.uuid}/edit"
+
+        expect(page).to have_content("Edit contention title")
+
+        within first(".issue-edit-text") do
+          click_edit_contention_issue
+        end
+
+        expect(page).to have_content("PTSD")
```
leikkisa: This test is a little weak (because it will have this text whether the feature is working or not).

spec/feature/intake/higher_level_review/edit_spec.rb
```diff
@@ -1443,6 +1443,33 @@ def click_cancel(visit_page)
       end
     end
 
+    context "when rating is withdrawn" do
```
leikkisa: I think this title is wrong

PR #10687, opened 2019-05-08
End holds set with timed hold tasks

tomas-nava:
Connects #9207

### Description
- interface to allow users to end holds set with `TimedHoldTask`s
- consolidates task action data functions in a new `TaskActionRepository` class
- updating the parent task cancels the `TimedHoldTask`
- a `TimedHoldTask` is invalid if it has more than one `TaskTimer`


app/models/task.rb
```diff
@@ -295,136 +326,10 @@ def cancel_task_and_child_subtasks
...
   def timeline_title
     "#{type} completed"
   end
 
-  def timeline_details
-    {
-      title: timeline_title,
-      date: closed_at
-    }
-  end
```
lowellrex: Removed because this function was never called?

app/repositories/task_action_repository.rb
```diff
@@ -0,0 +1,226 @@
+# frozen_string_literal: true
+
+class TaskActionRepository
```
lowellrex: I love this! Such a clean way to separate this stuff out!

client/COPY.json
```diff
@@ -433,6 +433,10 @@
   "NON_COMP_ADDRESS_MESSAGE": "The Veteran/appellant has filed a Notice of Disagreement at the Board of Veterans' Appeals. In order to decide that appeal, the Board will need the complete records from",
   "INTAKE_EDIT_TITLE": "Edit contention title",
 
+  "END_HOLD_MODAL_TITLE": "End hold early",
+  "END_HOLD_MODAL_BODY": "Do you want to end the hold early?",
+  "END_HOLD_SUCCESS_MESSAGE_TITLE": "Success! The hold has been ended early.",
```
lowellrex: 👍

app/models/task.rb
```diff
@@ -22,6 +22,7 @@ class Task < ApplicationRecord
   before_update :set_timestamps
   after_update :update_parent_status, if: :task_just_closed_and_has_parent?
   after_update :update_children_status_after_closed, if: :task_just_closed?
+  after_update :cancel_timed_hold, unless: :task_just_placed_on_hold?
```
lowellrex: If we create a child task for a task that is on a timed hold, will we cancel the existing timed hold? If not, I think it is fine to defer that work to a follow on PR since this looks good (and big!) as is and we aren't creating timed holds yet.

app/models/task.rb
tomas-nava: Yep, it was [defined](https://github.com/department-of-veterans-affairs/caseflow/commit/74be9c6490efe8d72d155c2d3b0834e1931d8f82#diff-c95c9451cf097bdd58f73d0f9b114f97R215) and [used](https://github.com/department-of-veterans-affairs/caseflow/commit/74be9c6490efe8d72d155c2d3b0834e1931d8f82#diff-d38fdc00b17f61e0a6e1bd027f26224cR223) in November '18, then [the usage was removed](https://github.com/department-of-veterans-affairs/caseflow/commit/b9602c993662133a16bb312a78d175998bf0654a#diff-d38fdc00b17f61e0a6e1bd027f26224cL283) in January but the definition never was.

app/models/task.rb
tomas-nava: We will; I added a test in [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/10687/commits/2edb095685dc4ba144d50ff7114ad63aed4d3eed).

PR #10693, opened 2019-05-08
Change task type (FE only)

### LINKED ISSUE #7126: Change task type (front end only)
Related support ticket - https://github.com/department-of-veterans-affairs/appeals-support/issues/3066#issuecomment-423973823

Stories:
* When a co-located user is assigned a task incorrectly identified by the attorney, they need to be able to change the task type, so they have the right actions available to them for the next step.
* When Litigation Support users are assigned a post-decisional motion (like a Motion for Reconsideration) and after reviewing the file they realize that the Veteran actually requested a Motion to Vacate, they need to change the task type to track what the Veteran actually requested about the case.
* When any user is assigned a mail task and notices that the mail team incorrectly categorized the mail task, they want to change the task type so it accurately reflects what the Veteran/mail says.

#### User, in this case a VLJ Support staff member, sees that the task that's been assigned to them by the attorney is inaccurate. They select the actions dropdown for that respective task, to change task type.
changing-task-1

#### They have the opportunity to select from an available list of replacement tasks (in this case, other admin actions). They can also provide a task note.
changing-task-2

#### They select a different admin action from the list
changing-task-3

#### They fill out a note
changing-task-4

#### They see a confirmation message as well as the reflected task changes in Currently active tasks
changing-task-5

## AC
- Front end work for change task type

## Out of scope (and needs discussion)
- Back end work for changing task type according to the above scenarios. We currently have automatic routing of mail tasks, so we need to think through how to handle a change in task type. Also, we need to think through how to trace task changes up the tree given different changes.

hschallhorn:
Resolves #7126

### Description
Give users the ability to change the the type of task in the case that it was assigned incorrectly. Applies to admin action tasks and mail tasks. Front end only, waiting on back end changes.

### Testing Plan
1. Go to ...

Screen Shot 2019-05-08 at 12 15 06 PM
Screen Shot 2019-05-08 at 12 15 20 PM
Screen Shot 2019-05-08 at 12 16 10 PM
Screen Shot 2019-05-08 at 12 16 32 PM


app/models/tasks/colocated_task.rb
```diff
@@ -70,6 +70,7 @@ def available_actions_with_conditions(core_actions)
     end
 
     core_actions.unshift(Constants.TASK_ACTIONS.COLOCATED_RETURN_TO_ATTORNEY.to_h)
+    # core_actions.unshift(Constants.TASK_ACTIONS.CHANGE_TASK_TYPE.to_h)
```
lowellrex: Let's add a comment here that points to this PR and and acknowledges that this is waiting for back-end work before it is enabled.

app/models/tasks/mail_task.rb
```diff
@@ -79,6 +79,14 @@ def most_recent_active_task_assignee(parent)
...
     self.class.label
   end
+
+  def change_task_type_data(_user = nil)
+    mail_assign_to_organization_data
+  end
+
+  def available_actions(user)
+    super(user) # .unshift(Constants.TASK_ACTIONS.CHANGE_TASK_TYPE.to_h)
+  end
```
lowellrex: Let's remove this function or comment it out with a comment like we'll do in `ColocatedTask`.

client/constants/CO_LOCATED_ACTIONS.json
```diff
@@ -1,5 +1,6 @@
 {
   "SEND_TO_TEAM": "send_to_team",
+  "CHANGE_TASK_TYPE": "change_task_type",
```
lowellrex: Does this change enable changing the task type for `ColocatedTask`s?

spec/controllers/tasks_controller_spec.rb
```diff
@@ -631,6 +631,28 @@
...
+
+      it "should update successfully" do
+        patch :update, params: { task: { action: new_action, instructions: new_instructions }, id: admin_action.id }
+        expect(response.status).to eq 200
+        response_body = JSON.parse(response.body)["tasks"]["data"]
+        expect(response_body.first["id"]).to eq admin_action.id.to_s
+        expect(response_body.first["attributes"]["label"]).to eq new_action
+        expect(response_body.first["attributes"]["instructions"]).to include new_instructions
+      end
+    end
```
lowellrex: Let's remove this test entirely. We can leave a note in the back-end ticket to reference the commit that added this test as one possible test to write.

spec/feature/queue/colocated_task_queue_spec.rb
```diff
@@ -168,4 +168,65 @@
       expect(page).to have_content(appeal.veteran.name.formatted(:readable_full))
     end
   end
+
+  describe "vlj support staff changes task type", skip: "needs back end implementation" do
```
lowellrex: Let's make this test work by mocking `ColocatedTask.available_actions` to include the change task type action, and mocking the back-end to respond with a success or failure as appropriate.

spec/feature/queue/mail_task_spec.rb
```diff
@@ -56,4 +56,126 @@
       end
     end
   end
+
+  describe "Changing a mail team task type", skip: "needs back end implementation" do
```
lowellrex: Let's make this test work by mocking as well.

client/constants/CO_LOCATED_ACTIONS.json
hschallhorn: No it does not.
🤔 
In fact, I'm struggling to find a place were we use this file at all...

client/app/queue/ChangeTaskTypeModal.jsx
```diff
@@ -0,0 +1,154 @@
...
+    const { error, ...otherProps } = this.props;
+    const { action } = this.state;
+
+    return <QueueFlowModal
+      validateForm={this.validateForm}
+      submit={this.submit}
+      title={COPY.CHANGE_TASK_TYPE_SUBHEAD}
+      button={COPY.CHANGE_TASK_TYPE_SUBHEAD}
+      pathAfterSubmit={`/queue/appeals/${this.props.appealId}`}
+      {...otherProps}
```
lowellrex: What other property are we passing to `QueueFlowModal` here?

client/app/queue/ChangeTaskTypeModal.jsx
```diff
@@ -0,0 +1,154 @@
...
+      data: {
+        task: this.buildPayload(),
+        role: this.props.role
+      }
+    };
+    const msgTitle = COPY.CHANGE_TASK_TYPE_CONFIRMATION_TITLE;
+    const oldAction = taskActionData(this.props).options.find((option) =>
+      option.value === task.label || option.label === task.label
+    );
+    const newAction = taskActionData(this.props).options.find((option) => option.value === action.actionLabel);
```
lowellrex: Can we rename these variables to something like `newTaskType` since the thing we are changing is the type of the task and not the actions that are available for the task?

client/app/queue/ChangeTaskTypeModal.jsx
```diff
@@ -0,0 +1,154 @@
...
+
+    const payload = {
+      data: {
+        task: this.buildPayload(),
+        role: this.props.role
+      }
+    };
+    const msgTitle = COPY.CHANGE_TASK_TYPE_CONFIRMATION_TITLE;
+    const oldAction = taskActionData(this.props).options.find((option) =>
+      option.value === task.label || option.label === task.label
```
lowellrex: We need to match by value or label in order to support mail task type changes as well as colocated task type changes, is that right?

client/app/queue/ChangeTaskTypeModal.jsx
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  submit = () => {
+    const { task } = this.props;
+    const { action } = this.state;
+
+    const payload = {
+      data: {
+        task: this.buildPayload(),
+        role: this.props.role
```
lowellrex: Why do we need the `role` property?

client/app/queue/ChangeTaskTypeModal.jsx
```diff
@@ -0,0 +1,154 @@
...
+    };
+  }
+
+  submit = () => {
+    const { task } = this.props;
+    const { action } = this.state;
+
+    const payload = {
+      data: {
+        task: this.buildPayload(),
```
lowellrex: Can we move the whole payload inside of the `buildPayload` function?

client/app/queue/ChangeTaskTypeModal.jsx
hschallhorn: Correct!

lowellrex: Do we need to do this `find()` operation? We only use `oldAction` one place where we call `oldAction.label` which I think will always be equivalent to `task.label`, no?

hschallhorn: 🤦‍♀

hschallhorn: No comment. Will fix.

client/app/queue/ChangeTaskTypeModal.jsx
```diff
@@ -0,0 +1,154 @@
...
+
+    return <React.Fragment>
+      <div id={key} key={key}>
+        <div {...marginTop(4)}>
+          <SearchableDropdown
+            errorMessage={highlightFormItems && !actionLabel ? COPY.FORM_ERROR_FIELD_REQUIRED : null}
+            name={COPY.CHANGE_TASK_TYPE_ACTION_LABEL}
+            placeholder="Select an action type"
+            options={taskActionData(this.props).options}
+            onChange={(option) => option && this.updateActionField('actionLabel', option.value)}
```
lowellrex: Can we put the entire `option` object into the local state so that we can pluck out the `label` and `value` without having to `find` one value by the other in the `submit` function?

client/app/queue/ChangeTaskTypeModal.jsx
```diff
@@ -0,0 +1,154 @@
...
+        // handle the error from the frontend
+      });
+  }
+
+  actionForm = (action) => {
+    const { highlightFormItems } = this.props;
+    const { instructions, actionLabel, key } = action;
+
+    return <React.Fragment>
+      <div id={key} key={key}>
```
lowellrex: Why do we need an id and key here?

client/app/queue/ChangeTaskTypeModal.jsx
```diff
@@ -0,0 +1,154 @@
...
+class ChangeTaskTypeModal extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      action: actionTemplate()
+    };
+  }
+
+  updateActionField = (key, value) => {
```
lowellrex: I think by default [`setState` does the "shallow merge"](https://reactjs.org/docs/react-component.html#setstate) that we are doing here so this could be rewritten into something like:

```jsx
updateActionField = (key, value) => this.setState({[key]: value });
```

Though maybe it could just be inlined entirely at that point.

client/app/queue/ChangeTaskTypeModal.jsx
```diff
@@ -0,0 +1,154 @@
...
+  appealWithDetailSelector,
+  taskById
+} from './selectors';
+import { marginTop } from './constants';
+import COPY from '../../COPY.json';
+
+import { taskActionData } from './utils';
+import QueueFlowModal from './components/QueueFlowModal';
+
+const actionTemplate = () => {
```
lowellrex: Do we call `actionTemplate` anywhere outside of the constructor?

client/constants/CO_LOCATED_ACTIONS.json
hschallhorn: We OK with removing this file?

lowellrex: Yup!

client/app/queue/ChangeTaskTypeModal.jsx
hschallhorn: So! `ColocatedTask.label` ends up being poa_clarification, missing_hearing_transcripts, etc., which is why `option.value === task.label` was needed in the first place.

client/app/queue/ChangeTaskTypeModal.jsx
hschallhorn: ⛔️

client/app/queue/ChangeTaskTypeModal.jsx
hschallhorn: We don't! Removed.

client/app/queue/ChangeTaskTypeModal.jsx
hschallhorn: Looks like nothing necessary. Removing.

client/app/queue/ChangeTaskTypeModal.jsx
```diff
@@ -0,0 +1,136 @@
...
+
+  submit = () => {
+    const { task } = this.props;
+    const { actionOption } = this.state;
+
+    const payload = this.buildPayload();
+
+    const msgTitle = COPY.CHANGE_TASK_TYPE_CONFIRMATION_TITLE;
+    const oldTaskType = taskActionData(this.props).options.find((option) =>
+      option.value === task.label || option.label === task.label
```
lowellrex: Ahh, I see what we're trying to do. A pattern we've used in the past for making the task's label value more readable is to search through the `CO_LOCATED_ADMIN_ACTIONS` hash for the corresponding key and use that value if one is present, and if one is not we fall back to the label. Let's try spinning [`TaskTable.actionNameOfTask`](https://github.com/department-of-veterans-affairs/caseflow/blob/444defa51a806f118e9a52e0929820d8d87e6597/client/app/queue/components/TaskTable.jsx#L101) out into a helper method in queue/utils and using it here.

app/models/tasks/colocated_task.rb
```diff
@@ -82,6 +85,10 @@ def actions_available?(_user)
     active?
   end
 
+  def change_task_type_data(_user = nil)
+    add_admin_action_data
+  end
```
lowellrex: #10687 moved the task action builder functions into a new `TaskActionRepository` module so I think this function (as well as the similar one in `MailTask`) will need to move there.

client/app/queue/components/TaskTable.jsx
```diff
@@ -23,7 +23,7 @@ import CaseDetailsLink from '../CaseDetailsLink';
 import ContinuousProgressBar from '../../components/ContinuousProgressBar';
 
 import { setSelectionOfTaskOfUser, bulkAssignTasks } from '../QueueActions';
-import { renderAppealType, taskHasCompletedHold } from '../utils';
+import { renderAppealType, taskHasCompletedHold, actionNameOfTask } from '../utils';
```
lowellrex: This is fantastic!

spec/feature/queue/mail_task_spec.rb
```diff
@@ -56,4 +56,148 @@
       end
     end
   end
+
+  describe "Changing a mail team task type" do
```
lowellrex: Tests look great!

PR #10703, opened 2019-05-08
Show when task is assigned to an organization in case details page.

### LINKED ISSUE #10306: Show assigned to label for tasks assigned to organizations
Currently [we do not show the "assigned to" label](https://github.com/department-of-veterans-affairs/caseflow/blob/51ca798dd93296b484b6fa3089db5cfb93603e1b/client/app/queue/components/TaskRows.jsx#L120) in the task snapshot and case timeline for tasks that are assigned to organizations. We should display the team's name instead. This effort may be aided by [separating the construction and presentation](https://github.com/department-of-veterans-affairs/caseflow/issues/10004) of each element in those tasks lists.

image

## Acceptance criteria
* [ ] Show organization name in assigned to label for tasks assigned to organizations

kevmo:
Resolves #10306 

Slight refactor of assigntedToListItem method in client/app/queue/components/TaskRows.jsx. If a non-legacy appeal is not assigned to an individual, method will now check to see if assigned to an organization and, if so, will display that Org in the Assigned To field in the case details page.

Also added `/* eslint-disable max-lines */` to TaskRows.jsx.  A potential ticket that comes to mind is removing the CSS-in-JavaScript in that file.


client/app/queue/components/TaskRows.jsx
```diff
@@ -116,7 +117,11 @@ class TaskRows extends React.PureComponent {
   }
 
   assignedToListItem = (task) => {
-    const assignee = task.isLegacy ? this.props.appeal.locationCode : task.assignedTo.cssId;
+    let assignee = task.isLegacy ? this.props.appeal.locationCode : task.assignedTo.cssId;
```
lowellrex: Let's just always use `task.assignedTo.name` for the else clause of this condition. I think will [accomplish the CSS ID/organization name logic](https://github.com/department-of-veterans-affairs/caseflow/blob/e5c743408366445909bf5eb116435bf5197606d2/app/models/task.rb#L443) we want.

kevmo: I reworked the logic again, because there are apparently some instances where there can be a non-legacy case that is not assigned to an organization (which was caught by failing tests). In these instances, we can still use CSS ID, though that is the last resort.

kevmo:
Tightened up spacing in a few instances to meet line limit. We should consider refactoring the CSS out of this in a future ticket. I did not do it in this ticket, because it seems like a good idea to take a bigger-picture look at how we handle this CSS-in-JavaScript. Some alternatives: 1) Move that code into SCSS files, 2) Make a separate `.js` file we can export/import styles from.

client/app/queue/components/TaskRows.jsx
```diff
@@ -116,7 +114,15 @@ class TaskRows extends React.PureComponent {
...
-    const assignee = task.isLegacy ? this.props.appeal.locationCode : task.assignedTo.cssId;
+    let assignee;
+
+    if (task.isLegacy) {
+      assignee = this.props.appeal.locationCode;
+    } else if (task.assignedTo.isOrganization) {
+      assignee = task.assignedTo.name;
+    } else {
+      assignee = task.assignedTo.cssId;
+    }
```
lowellrex: Let's move this to the back-end. I think adding an `assignee_name` attribute to the task serializer and legacy task serializers that encompass this logic would make sense. This would allow this front-end function just call `task.assigneeName` without any conditional logic.

PR #10704, opened 2019-05-08
Automatic case distribution for Hearing Docket

### LINKED ISSUE #9999: Include hearings docket cases in auto-case distribution
Hearings will start being held for AMA appeals in the hearings docket on April 1st, so auto-case distribution needs to distribute appeals in the hearings docket on that date as well. In order to do this we will need to build out [the `HearingRequestDocket` class](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/dockets/hearing_request_docket.rb) in the model of [the `DirectReviewDocket` class](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/dockets/direct_review_docket.rb).

For more information about when a case is eligible for distribution see [this related github discussion](https://github.com/department-of-veterans-affairs/caseflow/issues/8288).

## Acceptance criteria
* [ ] Build out the `HearingRequestDocket` class.
* [ ] Auto-case distribution distributes cases from the hearings docket.

monfresh:
Resolves #9999

In order to support distribution of appeals in the Hearing docket, we
needed to define the `age_of_n_oldest_priority_appeals` and
`distribute_appeals` methods to adhere to the conditions that are
specific to Hearing appeals.

In `age_of_n_oldest_priority_appeals`, the first set of base conditions
that get applied are: priority, ready for distribution, and in the
hearing docket. These scopes (defined in the `Appeal` model) get
applied via the `Docket` class. An appeal is considered prioritized if
it is Advanced on the Docket. More details about priority status here:
https://github.com/department-of-veterans-affairs/caseflow/wiki/Automatic-Case-Distribution#priority

For appeals in the Hearing docket, we only want `genpop` appeals, which
are those that adhere to any of these conditions:
- with no hearings
- with hearings, but none that are held
- where the most recent held hearing was held by an inactive judge

This means that if an appeal has multiple hearings, one or more of
which have the `disposition` attribute set to `held`, we only want to
look at the most recent hearing. In order to determine the date the
hearing was held, we need to join on the `HearingDay` table and look
at the `scheduled_for` field.

More background info on `genpop` here:
https://github.com/department-of-veterans-affairs/caseflow/wiki/Automatic-Case-Distribution#genpop

For the `distribute_appeals` method, the base conditions are: ready for
distribution and in the hearing docket. Priority and genpop are passed
in via `AmaCaseDistribution#ama_distribution` because both priority and
nonpriority cases can be distributed, as well as "only genpop",
"not genpop", or "any" (which is a combination of the previous two).

### DIFF
--- a/app/models/dockets/hearing_request_docket.rb
+++ b/app/models/dockets/hearing_request_docket.rb
@@ -5,15 +5,23 @@ class HearingRequestDocket < Docket
     "hearing"
   end
 
-  # CMGTODO: should only return genpop appeals.
-  def age_of_n_oldest_priority_appeals(_num)
-    []
+  def age_of_n_oldest_priority_appeals(num)
+    relation = appeals(priority: true, ready: true).limit(num)
+
+    HearingRequestDistributionQuery.new(
+      base_relation: relation, genpop: "only_genpop"
+    ).call.map(&:ready_for_distribution_at)
   end
 
-  # CMGTODO
-  # rubocop:disable Lint/UnusedMethodArgument
-  def distribute_appeals(_distribution, priority: false, genpop: "any", limit: 1)
-    []
+  def distribute_appeals(distribution, priority: false, genpop: "any", limit: 1)
+    base_relation = appeals(priority: priority, ready: true).limit(limit)
+
+    appeals = HearingRequestDistributionQuery.new(
+      base_relation: base_relation, genpop: genpop, judge: distribution.judge
+    ).call
+
+    HearingRequestCaseDistributor.new(
+      appeals: appeals, genpop: genpop, distribution: distribution, priority: priority
+    ).call
   end
-  # rubocop:enable Lint/UnusedMethodArgument
 end
--- a/spec/models/dockets/hearing_request_docket_spec.rb
+++ b/spec/models/dockets/hearing_request_docket_spec.rb
@@ -3,28 +3,385 @@
 describe HearingRequestDocket do
   before do
     FeatureToggle.enable!(:ama_auto_case_distribution)
+    FeatureToggle.enable!(:ama_acd_tasks)
   end
   after do
     FeatureToggle.disable!(:ama_auto_case_distribution)
+    FeatureToggle.disable!(:ama_acd_tasks)
   end
 
-  let(:judge_user) { create(:user) }
-  let!(:vacols_judge) { create(:staff, :judge_role, sdomainid: judge_user.css_id) }
-  let(:distribution) { Distribution.create!(judge: judge_user) }
-
   describe "#age_of_n_oldest_priority_appeals" do
+    let(:judge_user) { create(:user, last_login_at: Time.zone.now) }
+    let!(:vacols_judge) { create(:staff, :judge_role, sdomainid: judge_user.css_id) }
+
     subject { HearingRequestDocket.new.age_of_n_oldest_priority_appeals(10) }
 
-    it "should return a stubbed value" do
-      expect(subject).to eq([])
+    it "only returns priority, distributable, hearing docket appeals that match the following conditions:
+        where the most recent held hearing was not tied to an active judge
+        OR
+        appeals that have no hearings at all
+        appeals that have no hearings with disposition held" do
+      _another_inactive_judge = create(:user, last_login_at: 70.days.ago)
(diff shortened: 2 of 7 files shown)


app/models/dockets/direct_review_docket.rb
```diff
@@ -11,6 +11,7 @@ def docket_type
   def due_count
     appeal_ids = appeals(priority: false, ready: true)
       .where("target_decision_date <= ?", DAYS_BEFORE_GOAL_DUE_FOR_DISTRIBUTION.days.from_now)
+    # WHY ARE WE DOING THIS BELOW INSTEAD OF COUNTING THE RESULTS ABOVE?
```
cmgiven: IIRC, this was necessary to prevent double-counting.

spec/models/dockets/hearing_request_docket_spec.rb
```diff
@@ -3,28 +3,265 @@
...
+        appeals that have no hearings at all
+        appeals that have no hearings with disposition held" do
+      _another_inactive_judge = create(:user, last_login_at: 70.days.ago)
+      JudgeTeam.create_for_judge(_another_inactive_judge)
+      create_appeals_that_should_not_be_returned_by_query
+      first_appeal = matching_all_conditions_with_inactive_judge
+      second_appeal = create_appeal_with_hearing_docket_type_but_no_hearings
+      third_appeal = create_appeal_with_hearing_docket_type_with_no_held_hearings
+      fourth_appeal = matching_all_conditions_with_no_judge
+      fifth_appeal = create_appeal_with_most_recently_held_hearing_not_tied_to_any_judge(judge_user)
```
cmgiven: It's not clear that these are all priority appeals from the name.

I don't think cases 4 and 5 should ever happen, but they are perfectly reasonable inclusions.

spec/models/dockets/hearing_request_docket_spec.rb
```diff
@@ -3,28 +3,265 @@
...
+        appeal = create_nonpriority_distributable_hearing_appeal_tied_to_distribution_judge(judge_user)
+
+        tasks = subject
+
+        expect(tasks.length).to eq(1)
+        expect(tasks.first.class).to eq(DistributedCase)
+        expect(distribution.distributed_cases.length).to eq(1)
+        expect(judge_user.reload.tasks.map(&:appeal)).to eq([appeal])
+      end
+    end
```
cmgiven: These are great and very readable.

spec/models/dockets/hearing_request_docket_spec.rb
```diff
@@ -3,28 +3,265 @@
...
+    appeal = create(:appeal, :ready_for_distribution, :advanced_on_docket_due_to_motion, docket_type: "hearing")
+    most_recent = create(:hearing_day, scheduled_for: 3.days.ago)
+    create(:hearing, disposition: "held", appeal: appeal, hearing_day: most_recent)
+
+    tied_hearing_day = create(:hearing_day, scheduled_for: 4.days.ago)
+    hearing = create(:hearing, disposition: "held", appeal: appeal, hearing_day: tied_hearing_day)
+    hearing.update(judge: judge)
+
+    appeal
+  end
```
cmgiven: The one test case I'm potentially missing is similar to this one, except the most recent hearing is 1) tied to a judge other than `judge`, but 2) is not held. This tests that we're only looking at held hearings, and that we can ignore the most recent hearing if it wasn't held and look to a previous hearing that was held.

app/models/dockets/hearing_request_docket.rb
```diff
@@ -5,15 +5,23 @@ def docket_type
...
   end
 
-  # CMGTODO: should only return genpop appeals.
-  def age_of_n_oldest_priority_appeals(_num)
-    []
+  def age_of_n_oldest_priority_appeals(num)
+    relation = appeals(priority: true, ready: true).limit(num)
+
+    HearingRequestDistributionQuery.new(
+      base_relation: relation, genpop: "only_genpop"
```
aroltsch: It seems as you are not using `genpop` attribute when calling `only_genpop_appeals` method. Why not to call this method `appeals_to_distribute` and instantiate the class with genpop: "only_genpop" instead of making `only_genpop_appeals` a public method?

monfresh: That's a great point. I will fix this.

PR #10715, opened 2019-05-10
Check whether our test suite passes with DB connections released during BGS calls

lowellrex:
Related slack conversation: https://dsva.slack.com/archives/C2ZAMLK88/p1557438218292700

### DIFF
--- a/app/services/db_service.rb
+++ b/app/services/db_service.rb
@@ -5,15 +5,13 @@
 # This enables us to release connections before making an external call.
 class DBService
   def self.release_db_connections
-    if FeatureToggle.enabled?(:release_db_connections)
-      if VACOLS::Record.connection_pool.active_connection?
-        Rails.logger.info("Releasing VACOLS DB Connection")
-        VACOLS::Record.connection_pool.release_connection if VACOLS::Record.connection.open_transactions == 0
-      end
-      if ActiveRecord::Base.connection_pool.active_connection?
-        Rails.logger.info("Releasing PG DB Connection")
-        ActiveRecord::Base.connection_pool.release_connection if ActiveRecord::Base.connection.open_transactions == 0
-      end
+    if VACOLS::Record.connection_pool.active_connection?
+      Rails.logger.info("Releasing VACOLS DB Connection")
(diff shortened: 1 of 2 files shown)


pkarman:
this seems to mostly work!

lowellrex:
@pkarman I think we should test this feature toggle live with a Board user who uses Caseflow extensively before putting much more time into updating the code to expect database connections to sometimes be released. @laurjpeterson do you know if there is attorney at the Board who would be willing to be part of a "performance pilot"?

laurjpeterson:
yes, many! are there any attributes that would make one attorney a better tester over another? 

like, we want to test with an attorney who has a lot of cases in their queue, so we can see how many seconds it takes their queue to load?

lowellrex:
> are there any attributes that would make one attorney a better tester over another?

An attorney who would be very quick to give us feedback (if they could be on slack that would be great!) and be willing to iterate with us when they raise an issue. If this feature works (which it should!) there will be no visible consequences to the attorney. However failures will be all-or-nothing failures that happen around requests to BGS. So we might see things like navigating to a document in Reader working, but the page never loading if they refresh the page when they are viewing a document. It would be great to have somebody who is vigilant about noticing and reporting those types of errors and doubly great if they were willing to reproduce the steps to cause the error while we fiddle with levers on our side (like turning on and off the feature toggle to see if we can cause the error again for 5 minutes after they report the error).

lowellrex:
@laurjpeterson How might we go about getting say, 3 attorneys to test this with?

laurjpeterson:
@lowellrex - one former decision drafting attorney, former legislation, regulations, and policy person, and current supervisory senior counsel Emily Tamlyn is in Slack. i wonder if she might be ok to work with, even if she isn't currently a decision drafting attorney. 

Two other decision drafting attorneys we've met with a bunch are Robert Behlen and Jane Nichols (in slack, but I don't know how recently she's been using it.) It's also possible to test with Paul, because I think he's started writing decisions again.

laurjpeterson:
@lowellrex - i confirmed with Emily Tamlyn that she has cases and is willing to meet with an engineer! Emily.Tamlyn@va.gov. Or, she's in slack.

PR #10718, opened 2019-05-10
Execute build script from anywhere

lowellrex:
Allows this script to be run from any directory instead of just its parent directory. Tested on a Mac using Bash but it should be portable.

### DIFF
--- a/local/vacols/build_push.sh
+++ b/local/vacols/build_push.sh
@@ -41,20 +41,22 @@ if [[ $# -gt 1 ]]; then
 fi
 
 build(){
+  parent_dir=$(dirname $0)
+  build_facols_dir="${parent_dir}/build_facols"
 
   echo "${bold}Building FACOLS from Base Oracle...${normal}"
 
   echo -e "\tCleaning Up Old dependencies and Bring Required Packages"
-  rm -rf build_facols && mkdir build_facols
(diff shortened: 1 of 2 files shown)


local/vacols/build_push.sh
```diff
@@ -41,20 +41,22 @@ if [[ $# -gt 1 ]]; then
...
+  parent_dir=$(dirname $0)
+  build_facols_dir="${parent_dir}/build_facols"
 
   echo "${bold}Building FACOLS from Base Oracle...${normal}"
 
   echo -e "\tCleaning Up Old dependencies and Bring Required Packages"
-  rm -rf build_facols && mkdir build_facols
-  cp Dockerfile setup_vacols.sql vacols_copy_* build_facols
-  cd build_facols
+  rm -rf build_facols_dir && mkdir build_facols_dir
```
lowellrex: Previously if this script was run from a directory that had a directory named `build_facols` but was not in the same directory as this script we would delete that directory before failing in the copy operation on line 49. This change makes it so things happen relative to the directory this script is living in.

enriquemanuel:
Thanks for the explanation and improvement. When I created it was envisioned to run from there, but with the interaction with the other tools, it makes sense this change.
💯

PR #10728, opened 2019-05-13
Enable all tasks to be placed on timed hold

lowellrex:
Connects #9207.

* Adds "Place task on hold" option to tasks that don't already have that option
  * Creates modal similar to the existing on hold page for `ColocatedTask`s
  * Continues using existing system for `ColocatedTask`s


app/models/tasks/timed_hold_task.rb
```diff
@@ -10,17 +10,22 @@ class TimedHoldTask < GenericTask
...
-      assigned_by: assigned_by || parent_task.assigned_to,
-      assigned_to: parent_task.assigned_to,
-      parent: parent_task,
-      days_on_hold: days_on_hold&.to_i,
-      instructions: [instructions].compact.flatten
-    )
+    multi_transaction do
+      if parent_task.is_a?(Task)
+        parent_task.update!(instructions: [parent_task.instructions, instructions].flatten.compact)
+      end
```
lowellrex: Adds instructions to the task that is being placed on hold so that the instructions display in the task snapshot.

app/models/tasks/timed_hold_task.rb
```diff
@@ -10,17 +10,22 @@ class TimedHoldTask < GenericTask
   after_create :cancel_active_siblings
 
   attr_accessor :days_on_hold
-  validates :days_on_hold, presence: true, inclusion: { in: 1..100 }, on: :create
+  validates :days_on_hold, presence: true, inclusion: { in: 1..120 }, on: :create
```
lowellrex: 120 days is a preset option for the colocated on hold length, so this matches that.

client/app/queue/components/StartHoldModal.jsx
```diff
@@ -0,0 +1,118 @@
...
+
+    return this.props.requestSave(`/tasks/${task.taskId}/place_hold`, payload, successMsg).
+      then((resp) => {
+        const response = JSON.parse(resp.text);
+
+        this.props.onReceiveAmaTasks(response.tasks.data);
+      });
+  }
+
+  render = () => {
```
lowellrex: Majority of this render function is copied from [the existing `ColocatedPlaceHoldView`](https://github.com/department-of-veterans-affairs/caseflow/blob/876dc69d0e3a394dfe0cce7be31f6247ef2d2a54/client/app/queue/ColocatedPlaceHoldView.jsx#L102) component that this modal will eventually replace entirely.

I didn't bother to factor out any of this into shared functions because the old component will be going away in short order as part of #9207.

spec/feature/queue/attorney_queue_spec.rb
```diff
@@ -62,4 +62,80 @@
       end
     end
   end
+
+  describe "timed holds" do
```
lowellrex: Front-end tests for this new feature.

app/models/tasks/generic_task.rb
```diff
@@ -58,6 +61,8 @@ def available_actions(user)
 
     []
   end
+  # rubocop:enable Metrics/AbcSize
+  # rubocop:enable Metrics/MethodLength
```
monfresh: What was the reason for disabling Rubocop here? Why not extract the different arrays into well-named methods that represent the different action types? For example:
```ruby
def available_actions(user)
  return [] unless user

  return actions_when_task_assigned_to_user if assigned_to == user

  if task_is_assigned_to_user_within_organization?(user)
    return actions_when_task_is_assigned_to_user_within_organization
  end

  if task_is_assigned_to_users_organization?(user)
    return actions_when_task_is_assigned_to_users_organization
  end

  []
end

def actions_when_task_is_assigned_to_user
  [
    appropriate_timed_hold_task_action,
    Constants.TASK_ACTIONS.MARK_COMPLETE.to_h,
  ]
end

def actions_when_task_is_assigned_to_user_within_organization
  [
    Constants.TASK_ACTIONS.REASSIGN_TO_PERSON.to_h
  ]
end

def actions_when_task_is_assigned_to_users_organization
  [
    Constants.TASK_ACTIONS.ASSIGN_TO_PERSON.to_h,
    Constants.TASK_ACTIONS.MARK_COMPLETE.to_h,
  ]
end
```

PR #10730, opened 2019-05-13
Add "Cancel and Assign RO" action to Address Verify task

### LINKED ISSUE #9861: Verify Address Task Cancelation Flow
## User Stories
As a Hearings Branch super user, I need to be able to cancel tasks that have been added or generated in error so that our team can proceed with scheduling a hearing for a case.

As a Hearings Branch super user, I need to control access to task cancelations so that tasks aren't erroneously canceled

As a Hearings Branch super user, I need to be able to cancel an Address Verification task and assign a Regional Office for a hearing if an address is valid but not able to be validated by the automatic geomatching. This prevents cases from being stuck in an address verification loop. 

## Mocks
https://va-gov.invisionapp.com/share/HZQUCJ1NCUV

## Acceptance Criteria
- [ ] Verify cancelation of all Hearings Branch admin tasks are available to super users only
- [ ] Verify that super users have an option to "Cancel task and assign RO" in the drop down for a Verify Address task
- [ ] Verify when a user selects "Cancel task and assign RO" a modal is presented that allows the user to select a RO and add notes
- [ ] Verify selecting an RO is required and the action will not complete without RO selection
- [ ] Verify that once these changes are confirmed, the Verify Address task is canceled and the case moves to the correct Regional Office Veterans to be scheduled list

ferristseng:
Resolves #9861

### Description

Create a admin action for an Address Verify task to cancel it and assign it to an RO. When the appeal gets assigned the new RO, new available hearing locations for are created for that RO.

### Acceptance Criteria

 - [x] Verify cancelation of all Hearings Branch admin tasks are available to super users only
 - [x] Verify that super users have an option to "Cancel task and assign RO" in the drop down for a Verify Address task
 - [x] Verify when a user selects "Cancel task and assign RO" a modal is presented that allows the user to select a RO and add notes
 - [x] Verify selecting an RO is required and the action will not complete without RO selection
 - [x] Verify that once these changes are confirmed, the Verify Address task is canceled and the case moves to the correct Regional Office Veterans to be scheduled list


.reek.yml
```diff
@@ -21,6 +21,7 @@ detectors:
   UtilityFunction:
     public_methods_only: true
     exclude:
+      - HearingAdminActionVerifyAddressTask#available_hearing_admin_actions
```
sharonwarner: What checks was this method failing on? We're definitely sure refactoring isn't worth it?

ferristseng: It was saying I should move that method, `available_hearing_admin_actions`, to a utility function, but it was actually something I was trying to override. I was thinking it's a false positive, so it's OK to exclude.

PR #10735, opened 2019-05-13
Inject docket number column into TaskTable instance in OrganizationQueue.

### LINKED ISSUE #10619: Use customColumns to build task table for OrganizationQueue
In order to support of the objective of #10049 (moving task table definitions to the back-end), we will [inject the columns](https://github.com/department-of-veterans-affairs/caseflow/blob/8f168920a70b82899c28b4e99e2551774c1c0c64/client/app/queue/components/TaskTable.jsx#L320) we want the table to display in `OrganizationQueue` instead of [setting properties](https://github.com/department-of-veterans-affairs/caseflow/blob/8f168920a70b82899c28b4e99e2551774c1c0c64/client/app/queue/OrganizationQueue.jsx#L134) that signal to `TaskTable` that we want to display certain columns. In order to do this we first need to redefine [`TaskTable` instance methods](https://github.com/department-of-veterans-affairs/caseflow/blob/8f168920a70b82899c28b4e99e2551774c1c0c64/client/app/queue/components/TaskTable.jsx#L101) into pure functions that can be passed in an array to the `customColumns` property as we do [in the `NonCompTabs` component](https://github.com/department-of-veterans-affairs/caseflow/blob/8f168920a70b82899c28b4e99e2551774c1c0c64/client/app/nonComp/components/NonCompTabs.jsx#L106).

Work in this direction was started in [a WIP branch for #10049](https://github.com/department-of-veterans-affairs/caseflow/compare/10049/wip_pull_tab_configs_out), which may be of use as a reference.

## Acceptance criteria:
* [ ] Change columns that `OrganizationQueue` uses from a methods dependent on instance state into standalone functions that do not rely on instance state.
* [ ] Pass `customColumns` to `TaskTable` in `OrganizationQueue` instead of setting various `include...` parameters.

kevmo:
Partially resolves #10619, which calls for injecting columns into `TaskTable` instances via a `customColumns` array, as opposed to setting properties.  

In order to tackle this more granularly, we have elected to do this one column at a time. This is the first PR in this series of work and represents the injection of the docket number column.

### DIFF
--- a/client/app/queue/components/TaskTable.jsx
+++ b/client/app/queue/components/TaskTable.jsx
@@ -37,7 +37,7 @@ import CO_LOCATED_ADMIN_ACTIONS from '../../../constants/CO_LOCATED_ADMIN_ACTION
 import ORGANIZATION_NAMES from '../../../constants/ORGANIZATION_NAMES.json';
 
 const hasDASRecord = (task, requireDasRecord) => {
-  return (task.appeal.isLegacyAppeal && requireDasRecord) ? task.taskId : true;
+  return (task.appeal.isLegacyAppeal && requireDasRecord) ? Boolean(task.taskId) : true;
 };
 
 const collapseColumn = (requireDasRecord) => (task) => hasDASRecord(task, requireDasRecord) ? 1 : 0;
--- a/client/app/queue/components/TaskTable.jsx
+++ b/client/app/queue/components/TaskTable.jsx
@@ -92,11 +92,7 @@ export class TaskTableUnconnected extends React.PureComponent {
   }
 
   taskHasDASRecord = (task) => {
-    if (task.appeal.isLegacyAppeal && this.props.requireDasRecord) {
-      return task.taskId;
-    }
-
-    return true;
+    return hasDASRecord(task, this.props.requireDasRecord);
   }
 
   collapseColumnIfNoDASRecord = (task) => this.taskHasDASRecord(task) ? 1 : 0
--- a/client/app/queue/components/TaskTable.jsx
+++ b/client/app/queue/components/TaskTable.jsx
@@ -38,7 +38,7 @@ import ORGANIZATION_NAMES from '../../../constants/ORGANIZATION_NAMES.json';
 
 const hasDASRecord = (task, requireDasRecord) => {
   if (task.appeal.isLegacyAppeal && requireDasRecord) {
-    return task.taskId;
+    return !!task.taskId;
   }
 
   return true;
@@ -69,7 +69,7 @@ export const docketNumberColumn = (tasks, requireDasRecord) => {
     },
     span: collapseColumn(requireDasRecord),
     getSortValue: (task) => {
-      if (!taskHasDASRecord(task, requireDasRecord)) {
+      if (!hasDASRecord(task, requireDasRecord)) {
         return null;
       }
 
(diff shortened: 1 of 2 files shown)


client/app/queue/components/TaskTable.jsx
```diff
@@ -36,6 +36,48 @@ import COPY from '../../../COPY.json';
...
+    header: COPY.CASE_LIST_TABLE_DOCKET_NUMBER_COLUMN_TITLE,
+    enableFilter: true,
+    tableData: tasks,
+    columnName: 'appeal.docketName',
+    customFilterLabels: DOCKET_NAME_FILTERS,
+    anyFiltersAreSet: true,
+    label: 'Filter by docket name',
+    valueName: 'docketName',
+    valueFunction: (task) => {
+      if (!taskHasDASRecord(task, requireDasRecord)) {
```
lomky: Style question:  it feels a little weird to me that the function can return either an ID or a boolean.

```
      if (requireDasRecord && taskHasDASRecord?(task)) {
        return task.taskId;
      else {
        return null;
      }
```
And make the function `taskHasDASRecord?` just return `true` or `false`.

If we check the `requreDasRecord` first with `&&` we can shortcut out if needed, and avoid passing the variable further in.

Thoughts?

client/app/queue/components/TaskTable.jsx
```diff
@@ -36,6 +36,48 @@ import COPY from '../../../COPY.json';
...
 
+const taskHasDASRecord = (task, requireDasRecord) => {
+  if (task.appeal.isLegacyAppeal && requireDasRecord) {
+    return task.taskId;
+  }
+
+  return true;
+};
+
+const collapseColumnIfNoDASRecord = (requireDasRecord) =>
```
lomky: For my React education purposes: what does it mean that this file defines `taskHasDASRecord` and `collapseColumnIfNoDASRecord` twice, once with `const` and once without?

client/app/queue/components/TaskTable.jsx
```diff
@@ -135,6 +177,37 @@ export class TaskTableUnconnected extends React.PureComponent {
...
+      getSortValue: (task) => {
+        if (!this.taskHasDASRecord(task)) {
+          return null;
+        }
+
+        return `${task.appeal.docketName || ''} ${task.appeal.docketNumber}`;
+      }
+    } : null;
+  }
+
```
lowellrex: Can we move this function back to where it was previously defined (what is now line 248) so that we reduce the apparent changes in this PR?

client/app/queue/components/TaskTable.jsx
lomky: It belatedly occurs to me that I'm thinking in Rails structure and we're in the React codebase, so let me know if I'm just thinking in the wrong syntax!

client/app/queue/components/TaskTable.jsx
```diff
@@ -135,6 +177,37 @@ export class TaskTableUnconnected extends React.PureComponent {
     } : null;
   }
 
+  caseDocketNumberColumn = () => {
```
lowellrex: Can this method just call the pure function this PR created so that the two functions don't have the opportunity to drift apart from one another? Something like the following perhaps:

```jsx
caseDocketNumberColumn = () => docketNumberColumn(this.props.tasks, this.props.requireDasRecord);
```

client/app/queue/components/TaskTable.jsx
lowellrex: Great question! @kevmo let's rename these functions to reduce confusion and any weird undefined behaviour. I don't expect any weird behaviour because the methods defined within the class definition (those without `const` which are bound to that particular instance) will be called when we call the functions within the context of `this.`.

client/app/queue/components/TaskTable.jsx
lowellrex: Agreed! It looks like we never use the value `taskHasDASRecord` returns in the truthy case (`git grep taskHasDASRecord`) so let's just go ahead and cast that value into a boolean in the function itself.

PR #10738, opened 2019-05-13
Fix regression with var name

pkarman:
Bash requires sigil on its variables.


lowellrex:
Thanks for cleaning this up for me Peter!

PR #10744, opened 2019-05-14
[WIP] Run specs in a random order

monfresh:
**Why**: Specs should not be dependent on any particular order. We
want to surface any order dependencies as bugs.


lowellrex:
For reference we had a related effort for Efolder a little while back: https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/761 and https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/794. Perhaps we can use some lessons from that experience to inform this effort.

pkarman:
picking this up again in #13448

PR #10750, opened 2019-05-14
Pulac

### LINKED ISSUE #10323: Lit Support Queue: Create "Notify OGC of Pulac-Cerullo" task
## User Story
As a Motions Attorney, I need to assign a case to Kate (i.e. our team liaison between OGC/CAVC) if the Veteran also filed a CAVC appeal so that they can make a determination about who has jurisdiction on the case. 

### Acceptance criteria:
- A new task type is created - "Notify OGC of Pulac-Cerullo"
- "Notify OGC of Pulac-Cerullo" is an action available to Motions Attorneys in the actions drop-down on the Case Details page
- Routing: the task is assigned Kate Sosna
- No feature flag is needed
- Display confirmation modal after user clicks "Notify OGC of Pulac-Cerullo" action
- After user clicks "Submit" on confirmation modal, return user to their individual queue and display a confirmation banner.

## Mocks

**Confirmation modal**

image

**Confirmation banner**

image

youngfreezyVA:
resolves #10323


app/models/tasks/mail_task.rb
```diff
@@ -13,6 +13,10 @@ class MailTask < GenericTask
   # Skip unique verification for mail tasks since multiple mail tasks of each type can be created.
   def verify_org_task_unique; end
 
+  def available_actions(_user)
+    super
+  end
```
lowellrex: What does this function do?

youngfreezyVA: it doesn't look like it's needed, @aroltsch added it when she was helping me with this PR when it was still a WIP.

app/models/tasks/mail_task.rb
```diff
@@ -116,6 +116,14 @@ def self.default_assignee(_parent, _params)
 end
 
 class ClearAndUnmistakeableErrorMailTask < MailTask
+  def available_actions(user)
+    if LitigationSupport.singleton.user_has_access?(user)
```
lowellrex: Since `ClearAndUnmistakeableErrorMailTask`s [will always be assigned to](https://github.com/department-of-veterans-affairs/caseflow/blob/81a747f987cc8bedd0bbf6897763cce4f683318c/app/models/tasks/mail_task.rb#L123) members of the  `LitigationSupport` team, I think this check is redundant and we can avoid it. [`GenericTask.available_actions`](https://github.com/department-of-veterans-affairs/caseflow/blob/81a747f987cc8bedd0bbf6897763cce4f683318c/app/models/tasks/generic_task.rb#L34) should take care of the organizational visibility concern.

app/repositories/task_action_repository.rb
```diff
@@ -212,6 +212,16 @@ def add_schedule_hearing_task_admin_actions_data(task, _user)
       }
     end
 
+
+    def assign_to_pulac_cerullo_data(task, _user)
+      org = PulacCurello.singleton
+      {
+        selected: org,
+        options: [{ label: org.name, value: org.id }],
+        type: task.type
```
lowellrex: I think this is the heart of the problem we are observing with the organization not being who we expect. I think we might need a special `PulacCerulloTask` or something similar here to avoid the [automatic child task creation](https://github.com/department-of-veterans-affairs/caseflow/blob/81a747f987cc8bedd0bbf6897763cce4f683318c/app/models/tasks/mail_task.rb#L51) we do in `MailTask`.

app/models/tasks/mail_task.rb
aroltsch: I thought that you needed that part if you wanted to call available_actions from ClearAndUnmistakeableErrorMailTask, etc classes because ClearAndUnmistakeableErrorMailTask is a child of MailTask and MailTask is a child of GenericTask.

aroltsch: You can try removing it and see if it still works :)

app/models/tasks/mail_task.rb
aroltsch: Makes sense.

app/repositories/task_action_repository.rb
aroltsch: If we create a new task PulacCerulloTask, would we need to differentiate between ClearAndUnmistakeableErrorMailTask
ReconsiderationMotionMailTask
VacateMotionMailTask

All three tasks can be assigned to Pulac Cerullo

lowellrex: I don't know much about what a "Pulac Cerullo" task implies, so this question comes from a place of ignorance, but why do we need to differentiate between the different ways a `PulacCerulloTask` is received? Will different actions be available for a `PulacCerulloTask` that is created as the child of a CUE vs a motion for reconsideration?

aroltsch: Yup, I think it is a question to Lauren!

youngfreezyVA: hi @laurjpeterson any thoughts here?

laurjpeterson: No, I don't think the actions on PulacCerullo task are different based on if it is a child of CUE vs. another type of motion. I think we can go with our generic task actions.

app/models/tasks/mail_task.rb
youngfreezyVA: I did before commenting :) in any case this stuff will be moved to a new task type (or rather its not needed because we'll be using the generic actions)

db/seeds.rb
```diff
@@ -283,6 +284,11 @@ def create_lit_support_user
     OrganizationsUser.add_user_to_organization(u, LitigationSupport.singleton)
   end
 
+  def create_pulac_curello_user
```
youngfreezyVA: @aroltsch is there an associated step that needs to happen to add Kate to the `PulacCurello` org in production

app/models/appeal.rb
```diff
@@ -679,7 +679,7 @@ def available_review_options
   private
 
   def most_recently_assigned_to_label(tasks)
-    tasks.order(:updated_at).last.assigned_to_label
+    tasks.order(:created_at).last.assigned_to_label
```
youngfreezyVA: filtering out the active tasks above by `on hold` broke like 33 tests..other tests/scenarios were expecting `on_hold` tasks to be present. so I changed the method from this: ```def most_recently_assigned_to_label(tasks)
    active_tasks = tasks.active.where.not(status: :on_hold)
    active_tasks.order(:created_at).last.assigned_to_label
  end``` to the above one-liner. @aroltsch

lowellrex: I think active, not on-hold tasks is the correct set of tasks we want to be selecting. Would you write a ticket to update this method so that it returns the correct value and fixes any tests that might fail? I don't think we need to do that work in this PR.

db/seeds.rb
aroltsch: Yes, once we deploy this, we can add Kate to PulacCurello in production.

app/models/appeal.rb
youngfreezyVA: awesome, here's the issue fyi https://github.com/department-of-veterans-affairs/caseflow/issues/10825

aroltsch: should we revert the change to sort by updated_at?

lowellrex: > should we revert the change to sort by updated_at?

Yup! Let's revert it so that we stay with the existing incorrect sorting resulting in an incorrect label instead of introducing a different way that we incorrectly sort and end up with an incorrect label.

client/constants/LIT_SUPPORT.json
```diff
@@ -0,0 +1,4 @@
+{
```
aroltsch: Let's rename this file to be ORGANIZATION_NAMES.json

app/models/tasks/mail_task.rb
```diff
@@ -298,6 +302,10 @@ def self.default_assignee(parent, _params)
 end
 
 class ReconsiderationMotionMailTask < MailTask
+  def available_actions(user)
+    super.push(Constants.TASK_ACTIONS.LIT_SUPPORT_PULAC_CERULLO.to_h)
```
aroltsch: I think we need that check after all - who can view this action. We need to check here who is viewing the actions. Here is an example where a mail team member can see Pulac action and it should not be happening: 

image

aroltsch: cc @lowellrex

aroltsch: I think I know what happens. When a mail user creates a new task called for example `VacateMotionMailTask`, it creates one VacateMotionMailTask task for Mail organization and then a child task assigned to Litigation Support. Is that the expected behavior?  Should we be creating  VacateMotionMailTask task assigned to Mail organization? Would it ever come back to the mail org?

client/constants/LIT_SUPPORT.json
lowellrex: If we're going to establish this pattern let's do it for all `Organization`s. I'm not sure we need to pull these values into a json file though since it looks like we only use each of these values in one place (as fallback names for organizations if the organization does not already exist in the database).

app/models/tasks/mail_task.rb
lowellrex: > Is that the expected behavior? Should we be creating VacateMotionMailTask task assigned to Mail organization?

Yup! This is the expected behaviour. We assign the first task to the `MailTeam` because we don't know how the task will be routed (logic in `default_assignee`) at the time we create the task. The task assigned to the `MailTeam` will be automatically closed when the child task assigned to a different team is completed.

> Would it ever come back to the mail org?

Not yet, but #10826 will finish the changes Hunter started to allow `ColocatedTask`s and `MailTask`s to have their type changed so we will use the parent task assigned to the `MailTeam` as a common parent for the old task and the new task in that scenario.

youngfreezyVA: ok, so do I still need to add the check back @aroltsch? I think I need to try to update the test to switch users either way.

PR #10753, opened 2019-05-14
Redirect BusinessLine organizations to /decision_reviews

### LINKED ISSUE #10752: BusinessLine organizations should redirect from /organizations/:name/tasks
The `BusinessLine` Organization type should not have access to any /organizations/:name/tasks. They are designed for `/decision_reviews` only.

If a request comes in for a `/organizations/:name/tasks` and the org is a BusinessLine type, redirect to the `/decision_reviews/:name` URL instead.

pkarman:
Resolves #10752


pkarman:
> a business line wouldn't have tasks at the other URL if there is an appeal for that business line.

Correct. BusinessLine organizations are only assigned DecisionReviewTasks so they should have zero tasks in the Appeal Queue.

PR #10765, opened 2019-05-14
Add VBMSError::BadSOAPMessage

pkarman:
when SOAP response is un-parseable


app/exceptions/vbms_error.rb
```diff
@@ -107,6 +108,9 @@ def initialize(error)
     # https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3728/events/331292/
     "The System has encountered an unknown error" => "Unknown",
 
+    # https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3954/events/329778/
+    "Unable to parse SOAP message" => "BadSOAPMessage",
```
monfresh: Do we know what is causing the bad soap message?

pkarman: Not yet.

PR #10772, opened 2019-05-14
Break out root task creation to reduce file size

### LINKED ISSUE #10601: Create module for creating root tasks for appeals
We currently create `RootTask`s using [a series of class methods](https://github.com/department-of-veterans-affairs/caseflow/blob/254b25166796ebd96cccf586ff1bc88b6a6ee470/app/models/tasks/root_task.rb#L44). In order to reduce the size of that file and more cleanly separate concerns, we should move this set of functions into its own module for creating `RootTask`s from an appeal.

## Acceptance criteria
* [ ] Use new `RootTaskForAppealFactory` (or similar) module for creating `RootTask`s.

hschallhorn:
Resolves #10601 

### Description
Break out root task creation to reduce file size


app/workflows/root_task_for_appeal.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+class RootTaskForAppeal
+  class << self
+    def create_root_and_sub_tasks!(appeal)
+      root_task = RootTask.create!(appeal: appeal)
+      create_vso_tracking_tasks(appeal, root_task)
+      if FeatureToggle.enabled?(:ama_acd_tasks)
```
lowellrex: This feature [has been on for 3 months](https://github.com/department-of-veterans-affairs/appeals-deployment/pull/1974). Let's remove this conditional and remove the `else` block.

app/workflows/root_task_for_appeal.rb
```diff
@@ -0,0 +1,82 @@
...
+          appeal: appeal,
+          parent: parent,
+          assigned_to: vso_organization
+        )
+      end
+    end
+
+    # Make this private again after RAMPs are refilled
+    # https://github.com/department-of-veterans-affairs/caseflow/pull/9406
+    # private
```
lowellrex: Let's make the functions below this private again.

hschallhorn: Already made everything that isn't used outside of this private, but `create_subtasks!` is still used in a [few](https://github.com/department-of-veterans-affairs/caseflow/blob/2d6fc81174a51c640e5e57b04027940fc46a26e2/app/models/tasks/judge_task.rb#L56) [places](https://github.com/department-of-veterans-affairs/caseflow/blob/2d6fc81174a51c640e5e57b04027940fc46a26e2/db/seeds.rb#L91).

hschallhorn: Of course, seeds has no reason not to use `create_root_and_sub_tasks`

app/workflows/root_task_for_appeal.rb
```diff
@@ -0,0 +1,82 @@
...
+          assigned_to: vso_organization
+        )
+      end
+    end
+
+    def create_evidence_submission_task!(appeal, parent)
+      EvidenceSubmissionWindowTask.create!(
+        appeal: appeal,
+        parent: parent,
+        assigned_to: MailTeam.singleton
```
lowellrex: Let's [set a default assignee](https://github.com/department-of-veterans-affairs/caseflow/blob/2d6fc81174a51c640e5e57b04027940fc46a26e2/app/models/tasks/schedule_hearing_task.rb#L112) for `EvidenceSubmissionWindowTask`s.

app/workflows/root_task_for_appeal.rb
```diff
@@ -0,0 +1,82 @@
...
+      )
+    end
+
+    def create_distribution_task!(appeal, parent)
+      DistributionTask.create!(
+        appeal: appeal,
+        parent: parent,
+        assigned_to: Bva.singleton,
```
lowellrex: Let's set a default assignee for `DistributionTask` in the class itself similar to what we do with [`ScheduleHearingTask`](https://github.com/department-of-veterans-affairs/caseflow/blob/2d6fc81174a51c640e5e57b04027940fc46a26e2/app/models/tasks/schedule_hearing_task.rb#L112).

app/workflows/root_task_for_appeal.rb
```diff
@@ -0,0 +1,82 @@
...
+        status: "on_hold"
```
lowellrex: We don't need to set the task status here since it will automatically get set when we create a child task from this task.

app/workflows/root_task_for_appeal.rb
```diff
@@ -0,0 +1,82 @@
...
+      ScheduleHearingTask.create!(appeal: appeal, parent: parent)
+    end
+
+    def create_subtasks!(appeal, parent)
+      distribution_task = create_distribution_task!(appeal, parent)
+
+      if appeal.evidence_submission_docket?
+        create_evidence_submission_task!(appeal, distribution_task)
+      elsif appeal.hearing_docket?
+        create_hearing_schedule_task!(appeal, distribution_task)
```
lowellrex: These three functions never get called outside of this module so let's inline them for readability. Something like:

```ruby
distribution_task = DistributionTask.create!(appeal: appeal, parent: parent)

if appeal.evidence_submission_docket?
  EvidenceSubmissionWindowTask.create!(appeal: appeal, parent: distribution_task)
elsif appeal.hearing_docket?
  ScheduleHearingTask.create!(appeal: appeal, parent: distribution_task)
```

db/seeds.rb
```diff
@@ -88,7 +88,7 @@ def create_ama_distribution_tasks
     )
 
     root_task = RootTask.create!(appeal: appeal)
-    RootTask.create_subtasks!(appeal, root_task)
+    RootTaskForAppeal.create_subtasks!(appeal, root_task)
```
lowellrex: Let's call this as `RootTaskForAppeal.create_root_and_sub_tasks!(appeal)` so that `create_subtasks` can be made a private method.

app/workflows/root_task_for_appeal.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+class RootTaskForAppeal
```
lowellrex: This class has two public functions: `create_root_and_sub_tasks!` and `create_ihp_tasks!` which both accept an `Appeal` one accepts a parent `DistributionTask`. I think we can rearrange this class to more precisely represent what we want it to do.

Let's rename it `InitialTasksFactory` and instantiate it with a single appeal argument. We should be able to get the `RootTask` for every appeal with a `root_task` that calls `RootTask.find_or_create_by(appeal: appeal)` because every appeal will only have a single `RootTask` (and indeed [this is true in our database](https://caseflow-looker.va.gov/sql/9nskh7xmgzn24j)).

We might need to do something a little more flexible than `DistributionTask.find_or_create_by(appeal: appeal, parent: root_task)` because we anticipate that appeals may eventually have more than one `DistributionTask` (motions for reconsideration, judges leaving before assigning all their cases, etc.) even though [no appeals currently have more than 1 `DistributionTask`](https://caseflow-looker.va.gov/sql/kvd6ppqmdxxntm). Perhaps we can just select the most recently created `DistributionTask` for an appeal if one already exists.

This would give us an interface like: `InitialTasksFactory.new(appeal: appeal).create_root_and_sub_tasks!`

What do you think?

app/workflows/root_task_for_appeal.rb
```diff
@@ -0,0 +1,62 @@
...
+        # For some RAMP appeals, this method may run twice.
+        existing_task = InformalHearingPresentationTask.find_by(
+          appeal: appeal,
+          assigned_to: vso_organization
+        )
+        existing_task || InformalHearingPresentationTask.create!(
+          appeal: appeal,
+          parent: parent,
+          assigned_to: vso_organization
+        )
```
hschallhorn: @lowellrex is the reason we don't use find_or_create_by here because the parent task may not be the same as the parent of the existing task?

app/workflows/root_task_for_appeal.rb
hschallhorn: In talking with Moncef about some code climate issues I feel that `create_ihp_tasks!` could be moved into something separate as it's being called in other places when not creating initial tasks.

spec/models/tasks/root_task_spec.rb
```diff
@@ -92,43 +86,76 @@
...
+        it "creates RootTask assigned to Bva organization" do
+          InitialTasksFactory.new(appeal).create_root_and_sub_tasks!
+          expect(RootTask.last.assigned_to).to eq(Bva.singleton)
+        end
+      end
     end
 
     context "when an evidence submission docket appeal is created" do
-      before do
-        FeatureToggle.enable!(:ama_acd_tasks)
```
hschallhorn: Removed but there are [~9 other tests](https://github.com/department-of-veterans-affairs/caseflow/search?q=FeatureToggle.enable%21%28%3Aama_acd_tasks%29&unscoped_q=FeatureToggle.enable%21%28%3Aama_acd_tasks%29) in the codebase that are still toggling this feature

hschallhorn: Will create a separate ticket to resolve this

hschallhorn: here! https://github.com/department-of-veterans-affairs/caseflow/issues/10792

app/workflows/initial_tasks_factory.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def create_root_and_sub_tasks!
+    create_vso_tracking_tasks
+    if FeatureToggle.enabled?(:ama_acd_tasks)
+      ActiveRecord::Base.transaction do
+        create_subtasks!
+      end
+    else
+      IhpTasksFactory.new(@root_task).create_ihp_tasks!
```
lowellrex: Since `ama_acd_tasks` is always on now let's remove this else block.

spec/models/tasks/root_task_spec.rb
```diff
@@ -57,7 +57,7 @@
                  ])
         end
 
-        before { RootTask.create_root_and_sub_tasks!(appeal) }
+        before { InitialTasksFactory.new(appeal).create_root_and_sub_tasks! }
```
lowellrex: Let's move all of these `create_root_and_sub_tasks` tests into a test file specifically for the `InitialTasksFactory`.

app/workflows/initial_tasks_factory.rb
hschallhorn: Removing this caused a bunch of failures in other tests. Added it back in to be handled in #10792

lowellrex: Works for me!

PR #10779, opened 2019-05-15
Add meta_request gem for better debugging

monfresh:
**Why**: This gem is used by RailsPanel, a great Chrome extension I
discovered via the Ruby Weekly newsletter.

https://github.com/dejan/rails_panel


monfresh:
Screenshot of the extension in use on the app locally:

Screen Shot 2019-05-15 at 11 15 21 AM

PR #10782, opened 2019-05-15
Updated node-sass

lomky:
### Description
Ran `yarn upgrade node-sass`.


enriquemanuel:
Once this merges, we will need to create a new "base" ami, or just wait until next morning (this happens automatically every morning at ~4am)

lomky:
I was also surprised by the number of updated packages, I don't immediately know why it was so many.

lomky:
@pkarman looks like Yarn added the integrity sha in late 2017, and a quite git blame shows these packages getting the new integrity line all predate that. [Yarn issue](https://github.com/yarnpkg/yarn/pull/5042). I guess yarn was happy to add that info in a read-only fashion, even though I scoped it to upgrade node-sass?

Issue #9397, opened 2019-02-13
Disable RAMP intake options


shanear:
We need a feature flag that disables both RAMP elections and RAMP refilings as options in intake. All active RAMP intakes should still be okay to go through, it just prevents any new ones.

Once that feature flag `enable_ramp_intake` is created, it should be enabled only for 3 users in production: 

* AMOPLENE
* AMCJVIDR
* AMCCWUND

leikkisa:
Before we disable the options, we should check with the business if this is based on receipt_date, or if it turns off altogether. We may want to add validation to receipt_date for these if it is receipt_date oriented instead (not sure how long it takes for things to go from mail to being intaken).

nicholasholtz:
So, just talked with AMO - it looks like they will want the RAMP establishment functionality to be present until May 1, 2019 (and to confirm with them before it gets turned off).

- they need to continue to process existing forms that have not been processed, and to re-process some forms that were erroneously processed. So this ticket should be moved to the backlog, @shanear.

pkarman:
Moved back to ready-for-dev as they have requested 3 users remain enabled, off for everyone else.

PR #10783, opened 2019-05-15
Only allow RAMP intakes if feature toggle enabled

leikkisa:
connects #9397


app/models/request_issue.rb
```diff
@@ -546,6 +546,8 @@ def create_decision_issues
       create_decision_issues_from_rating
     end
 
+    # We expect all rating request issues on an EP to get an associated rating created when they're decided
```
leikkisa: This comment is not really part of this PR, but I didn't want to make a separate PR just for this.

client/app/intake/pages/selectForm.jsx
```diff
@@ -32,13 +33,25 @@ class SelectForm extends React.PureComponent {
...
+      }
+
+      { enableSearchableDropdown && <SearchableDropdown
         name="form-select"
         label="Which form are you processing?"
         placeholder="Enter or select form"
         options={radioOptions}
         onChange={this.setFormTypeFromDropdown}
         value={this.props.formType} />
+      }
```
leikkisa: This was here before we took out the AMA feature toggle.  From the design team of the past: if we have more than 3 options, it should be a searchable dropdown, if not it should be radio buttons.

spec/support/intake_helpers.rb
```diff
@@ -3,6 +3,19 @@
 # rubocop:disable Metrics/ModuleLength
 module IntakeHelpers
   # rubocop: disable Metrics/ParameterLists
+
+  def select_form(form_name)
```
pkarman: I really like this refactor! Nice and elegant.

Sjones352:
Passes UAT

PR #10789, opened 2019-05-15
Update document series ids with 1 SQL statement

monfresh:
**Why**: One of our slowest transactions is
`Reader::DocumentsController#index` due to the call to `annotations`,
which triggers an update of hundreds of documents via
`AddSeriesIdToDocuments`. Instead of using a separate DB call for
each document, we can batch update the documents with a single DB
call. This is possible thanks to the `activerecord-import` gem which
takes advantage of some features only available in Postgres 9.5+, which
was the main reason I have been pushing for us to upgrade our Postgres
version.

I also took this opporturity to do some refactoring.


app/services/add_series_id_to_documents.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+class AddSeriesIdToDocuments
```
pkarman: a nit: I prefer class names to be nouns. Maybe `BulkSeriesIdAssigner` or something like that? I know there are some in the Rails community who advocate for Rails services to be named with verbs, but that doesn't seem to be the convention in this app and it goes against broader OO naming conventions.

monfresh:
Before, the individual queries looked like this:
```
UPDATE "documents" SET "series_id" = $1 WHERE "documents"."id" = $2 ["series_id", "d41fd22f-5ee8-4069-a7a3-c26de4a8c58a"], ["id", 1]
UPDATE "documents" SET "series_id" = $1 WHERE "documents"."id" = $2 ["series_id", "54efded9-8423-462e-8977-49c882fd7771"], ["id", 2]
UPDATE "documents" SET "series_id" = $1 WHERE "documents"."id" = $2 ["series_id", "29b455f4-e4c7-4c3f-97f5-6789181e05cc"], ["id", 3]
UPDATE "documents" SET "series_id" = $1 WHERE "documents"."id" = $2 ["series_id", "1c3ed0c3-f0ee-4cbb-a6f8-88f12de80b86"], ["id", 4]
UPDATE "documents" SET "series_id" = $1 WHERE "documents"."id" = $2 ["series_id", "884ed8f8-361c-4967-8574-3c0969f29a18"], ["id", 5]
UPDATE "documents" SET "series_id" = $1 WHERE "documents"."id" = $2 ["series_id", "f6456d3c-52ae-4cbd-92e4-59dba3449daf"], ["id", 6]
```
Now, there is only 1 SQL statement for any number of documents:
```
INSERT INTO "documents" ("id","series_id","vbms_document_id")
VALUES
  (1,'fe4da70e-57ca-4a30-b8e9-d60c6deaf9b0','eae35a01-2c19-4d35-a79e-e96944fc695e'),
  (2,'b44edf98-25a2-4398-94c5-4a9f9dab6de2','426279ff-f7f8-4214-ae81-007bb6eca244'),
  (3,'9137a44a-c99b-4c3a-b9e5-47b9bc3dbc3d','74a3aa46-c36c-4658-8369-6dd88add9397'),
  (4,'c6f1e53e-a524-4ae2-870e-3f1950edc63d','2616721e-72aa-44fe-b50f-0c4f6c225ff7'),
  (5,'aeefdd13-adbd-45e4-97b7-f31e71f90abb','b51c3a63-05da-4b89-b3dd-25082f87e97d'),
  (6,'44e7d89e-43b2-4828-b841-20dc76543a4d','44e7d89e-43b2-4828-b841-20dc76543a4d')
ON CONFLICT (id) DO UPDATE SET "series_id"=EXCLUDED."series_id" RETURNING "id"
```

A very basic benchmark shows that this is about 16x faster. I just compared updating 500 records:

```ruby
def fast
  start = Time.zone.now
  documents = Document.where(series_id: nil).limit(500)
  values = documents.map do |doc|
    doc.series_id = "foo"
    doc
  end
  Document.import(values, on_duplicate_key_update: [:series_id])
  end_time = Time.zone.now
  puts "difference: #{end_time - start}"
end
# difference: 0.113942

def slow
  start = Time.zone.now
  documents = Document.where(series_id: nil).limit(500)
  documents.each { |doc| doc.update!(series_id: "foo") }
  end_time = Time.zone.now
  puts "difference: #{end_time - start}"
end
# difference: 1.809335
```
To run this in the console, I first created 500 new Documents like this:
```ruby
(1..500).each { |i|  Document.create(type: "NOD", vbms_document_id: i+1000) }
```
Then, I copied one of the methods, and then ran it with `send(:fast)` or `send(:slow)`.
The `fast` one completed in 0.11 seconds, whereas the `slow` one took 1.8 seconds.

app/services/add_series_id_to_documents.rb
monfresh: Sure, I'm down with that. Maybe `DocumentSeriesIdAssigner`?

Issue #10717, opened 2019-05-10
Refactor EfolderService calling code to expect immediate response


lowellrex:
`ExternalApi::EfolderService` [sleeps for 1 second](https://github.com/department-of-veterans-affairs/caseflow/blob/dc71939c5c084bf027932395b8a01a0f678fd26b/app/services/external_api/efolder_service.rb#L36) for as many as 10 seconds each time the application does not get  the particular response it wants. The application behaves in this way because we want to return the correct data to the calling code even if that data is not available immediately.

The advantage to this approach is that the calling code is more likely (though not guaranteed) to receive the data it expects from the Efolder service and only has to handle the failure case in rare circumstances.

The disadvantages of this approach are:
* We occupy the entire call stack (including database connections #10716) while we wait for the correct response, tightly coupling every step in the path to the next one
* We consume resources that other requests could be using
* The calling code lacks visibility into what is happening and flexibility to respond as it likes

## Acceptance criteria
* [ ] Remove `sleep` from `ExternalApi::EfolderService.generate_efolder_request` so that it responds immediately
* [ ] Refactor calling code to expect immediate response that does not necessarily have correct data

pkarman:
We could achieve the same result if we moved the polling behavior to the browser side. That has the advantage of keeping the HTTP cycle short (releasing the db connection faster).

So instead of
```
browser -> caseflow_server_polling -> efolder
```
we could
```
browser_polls -> caseflow -> efolder
```
we'd end up with more (but shorter) HTTP requests to Caseflow over time, but the same number of HTTP requests to Efolder.

We might want to leave the polling behavior in `ExternalApi::EfolderService.generate_efolder_request` since that might be used from a background job, and instead pass `1` iteration to `ExternalApi::EfolderService.generate_efolder_request` from the affected controller actions.

PR #10790, opened 2019-05-15
EFolderService.document_count uses efolder /api/v2/veterans endpoint

pkarman:
related to #10717 

**Requires** https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/1098

**Why** Efolder now has an API endpoint with a quick response for Veteran document count. Use that for Queue table preview, rather than triggering Efolder manifest fetch for all tasks.


app/services/external_api/efolder_service.rb
```diff
@@ -3,15 +3,10 @@
...
-    response[:documents]&.size
-    # We expect to fail fast. generate_efolder_request() raises an error if we have not gotten a positive response from
-    # Efolder after we are done retrying. Since we only retry once, it is more likely that we will raise that error.
-    # Since this function is only used to get the text that the front-end will display, we catch that error to avoid
-    # returning a 504 response to the front-end and showing the error text.
-  rescue Caseflow::Error::DocumentRetrievalError
-    "Try again later..."
+  def self.document_count(file_number, user)
+    headers = { "FILE-NUMBER" => file_number }
+    response = send_efolder_request("/api/v2/veterans", user, headers)
```
aroltsch: This is a little confusing that we are calling GET `/veterans` and receiving a document_count back

pkarman: would it make more sense if the endpoint was `/api/v2/veterans/document_count` ?

pkarman: I'm not sure of the historical reasons why the `FILE-NUMBER` is in the headers and not the URL, other than keeping potential PII out of the URL (the logs are full of PII so I'm not sure that it matters).

pkarman: or maybe `/api/v2/document_count` makes more sense?

lowellrex: > I'm not sure of the historical reasons why the FILE-NUMBER is in the headers and not the URL, other than keeping potential PII out of the URL (the logs are full of PII so I'm not sure that it matters).

We send URLs to Sentry (which we host) and Sentry sends messages to Slack (which we don't host). We had a PII spillage incident nearly 2 years ago now where we were posting SSNs (by way of claims file numbers if I remember correctly) in Slack via Sentry so we changed our practice to not send any PII to Sentry (cleanup from the incident was a pain as well).

We also send URLs to New Relic (and maybe DataDog as well?) which we do not host so I think it makes sense to leave all PII out of URLs in order to reduce the likelihood of sending those services PII accidentally.

Additionally, we expect Caseflow users to share URLs (and we even ask them to in the course of debugging!) so I think it is a best practice to not have any PII in the URLs even when they aren't user-facing so that we don't get in the habit of putting PII in URLs.

aroltsch: I like `/api/v2/document_count` !

pkarman:
API route changed in https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/1101

PR #10808, opened 2019-05-16
apply stub_const to quiet warnings

pkarman:
more #tech-debt payment


spec/models/hearing_spec.rb
```diff
@@ -60,7 +60,7 @@
     end
 
     before do
-      BGSService = ExternalApi::BGSService
+      stub_const("BGSService", ExternalApi::BGSService)
```
monfresh: We also need to remove the one in the `after` block. This probably applies to other files in this PR.

PR #10814, opened 2019-05-17
AMA AOD functionality

### LINKED ISSUE #8682: Daily Docket | AMA AOD functionality
## User Stories
As a Judge, I need the ability to grant or deny AOD status for a Veteran based on evidence presented to me at a hearing. 

As a Judge, I need to make sure my ruling is recorded properly with a reason, so that this can be referenced in the future and tracked. 

As a Judge, I need to make sure my ruling is applied immediately so that it impacts the way the case is prioritized. 

## Mocks
https://va-gov.invisionapp.com/share/Y3QWT5ARMNJ

## Acceptance Criteria
- [x] Verify this functionality exists for AMA hearings only.
- [x] Verify users can select either Grant or Deny from the AOD drop down. Remove the filed option.
- [x] Verify a dropdown option is available for AOD reason. This dropdown should not be editable unless AOD is selected as Grant or Deny.
- [x] Verify when the user selects an AOD option, the AOD reason drop down is accessible. 
- [x] Verify any changes must be saved via the save button.

lomaxap:
Resolves #8682

## Acceptance Criteria
- [x] Verify this functionality exists for AMA hearings only.
- [x] Verify users can select either Grant or Deny from the AOD drop down. Remove the filed option.
- [x] Verify a dropdown option is available for AOD reason. This dropdown should not be editable unless AOD is selected as Grant or Deny.
- [x] Verify when the user selects an AOD option, the AOD reason drop down is accessible. 
- [x] Verify any changes must be saved via the save button.


app/controllers/hearings_controller.rb
```diff
@@ -21,8 +29,15 @@ def update
       Transcription.find_or_create_by(hearing: hearing)
       hearing.update!(params)
     end
+  end
 
-    render json: hearing.to_hash(current_user.id)
+  def update_advance_on_docket_motion
+    motion = AdvanceOnDocketMotion.find_or_create_by!(person_id: advance_on_docket_motion_params[:person_id])
+    motion.update(advance_on_docket_motion_params)
+  end
```
lomaxap: should one `AdvanceOnDocketMotion` be created per appellant? Or per appellant/user?

PR #10818, opened 2019-05-17
Create bulk assign endpoint

### LINKED ISSUE #9621: Create endpoint that accepts requests to bulk assign hearings management tasks
This ticket exists to create an endpoint that accepts requests for the bulk assignment of tasks assigned to the hearings management branch (more specifics around this effort in [this ticket](https://github.com/department-of-veterans-affairs/caseflow/issues/8620#issuecomment-463414729)).

## Acceptance criteria
* [ ] Endpoint responds to requests as expected

aroltsch:
Resolves #9621 

### Description
BE endpoint to bulk assign tasks


spec/workflows/bulk_task_assignment_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+    let!(:schedule_hearing2) do
+      FactoryBot.create(:no_show_hearing_task, 
+        assigned_to: organization, 
+        created_at: 2.days.ago)
+    end
+    let!(:schedule_hearing3) do
+      FactoryBot.create(:no_show_hearing_task, 
+        assigned_to: organization, 
+        created_at: 1.days.ago)
+    end
```
tomas-nava: small thing, but these should be named `:no_show_hearing1`, `:no_show_hearing2`, `:no_show_hearing3`

app/workflows/bulk_task_assignment.rb
```diff
@@ -0,0 +1,102 @@
...
+    ActiveRecord::Base.transaction do
+      tasks_to_be_assigned.map do |task|
+        Task.create!(
+          type: task_type,
+          appeal: task.appeal,
+          assigned_by: assigned_by,
+          parent: task,
+          assigned_to: assigned_to
+        )
+      end
```
tomas-nava: Looking at [the logic for reassigning individual tasks in the Task model](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/task.rb#L269), reassigned tasks are siblings rather than children of the original task. In fact reassigned tasks are exact duplicates of the original organization task, with the exception of `assigned_by_id`, `assigned_to_id`, and (optionally) `instructions`.

Maybe we could just call `task.reassign(params, assigned_by)` where `params` looks like
```
{"assigned_to_id"=>assigned_to.id, "assigned_to_type"=>"User"}
```

spec/workflows/bulk_task_assignment_spec.rb
aroltsch: good catch!

app/workflows/bulk_task_assignment.rb
aroltsch: I think this reassign logic is used to reassign the task from one user to another user. If the task is reassigned from Organization to User, this User task becomes a child of the Organization task. Let me confirm with Lowell.

app/models/tasks/disposition_task.rb
```diff
@@ -140,7 +140,7 @@ def update_hearing_disposition(disposition:)
   end
 
   def check_parent_type
-    if parent.type != HearingTask.name
+    if parent&.type != HearingTask.name
```
lowellrex: `DispositionTask`s should always have a parent task of some kind. Can we add a validation for that?

app/workflows/bulk_task_assignment.rb
lowellrex: I don't know about this particular business case, but it makes sense to me that the bulk-assigned tasks that are assigned to users would be children of the organization task.

Maybe we could rename the beginning the `reassign` method to something like `duplicate_with_params` so we could re-use that logic? We'd still need to change the parent, but

tomas-nava: Yeah, sorry, my original comment was incorrect. I think the way you're doing it here is fine, though maybe you could explicitly set `assigned_to_type` to `User.name`?

app/models/tasks/disposition_task.rb
aroltsch: Done!

app/workflows/bulk_task_assignment.rb
aroltsch: Theoretically, if you do this `assigned_to: assigned_to` (set the object instead of an ID), rails knows what type of object it is so setting assigned_to_type is unnecessary.

tomas-nava: Ahhh, I could see it was working but couldn't figure out how. I'm glad to know it's Rails magic, thank you.

Issue #10825, opened 2019-05-20
create new scope  in `Task` that finds actionable tasks  that filters out `on_hold` tasks


youngfreezyVA:
create a scope in `Task` that finds actionable tasks - this would entail renaming `active` to `open` and use `active` to mean actionable

- update all usages/instances of `tasks.active.where.not(status: :on_hold)` to use this new scope

- update `most_recently_assigned_to_label` in `app/models/appeal.rb ` to use this new scope

- fix the 30+ failing tests as a result of this change

lowellrex:
Could we include renaming `inactive` to `closed` as well? So we end up with the following taxonomy: 
![image](https://user-images.githubusercontent.com/32683958/58029892-880d7f80-7aeb-11e9-9a67-ca95b37580bd.png)

Where we currently have:
![image](https://user-images.githubusercontent.com/32683958/58029962-b25f3d00-7aeb-11e9-9120-9adfe37865bb.png)

youngfreezyVA:
sounds good! that makes sense. i'll also follow the bullet points above and look to replace usages of the old taxomony of closed/inactive

PR #10831, opened 2019-05-21
Stop cascading task closure at org boundary

### LINKED ISSUE #10324: Stop cascading task closure at organization boundary
We [cascade task closures](https://github.com/department-of-veterans-affairs/caseflow/blob/3e0639a69202486a15e4b2b8bb763754c14b236e/app/models/task.rb#L22) so that organization-level tasks are completed when the task that is assigned to the individual is completed. This works well in most cases, but fails if the task is assigned from an organization to a different team. Consider the following scenario:

```
+-------------------+
|                   |
|      RootTask     |
|                   |
+---------+---------+
          |
          |
+---------+---------+
|                   |
| QualityReviewTask |
|   (organization)  |
|                   |
+---------+---------+
          |
          |
+---------+---------+
|                   |
|   ColocatedTask   |
|   (organization)  |
|                   |
+---------+---------+
          |
          |
+---------+---------+
|                   |
|   ColocatedTask   |
|   (individual)    |
|                   |
+-------------------+
```

OR

```
+-------------------+
|                   |
|      RootTask     |
|                   |
+---------+---------+
          |
          |
+---------+---------+
|                   |
| QualityReviewTask |
|   (organization)  |
|                   |
+---------+---------+
          |
          |
+---------+---------+
|                   |
|   JudgeQRTask     |
|   (individual)    |
|                   |
+-------------------+
```

When the individually-assigned `ColocatedTask` is closed the organizationally-assigned parent `ColocatedTask` will also be closed (good!). However, because [our checks are somewhat naive](https://github.com/department-of-veterans-affairs/caseflow/blob/3e0639a69202486a15e4b2b8bb763754c14b236e/app/models/task.rb#L466) the `QualityReviewTask` will also be closed. We should only close the parent task if it is of the same type as the child task, otherwise we should just change the status from "on hold" to "assigned" (maybe we could pass the child task that just completed to the function [when we call `parent.when_child_task_completed `](https://github.com/department-of-veterans-affairs/caseflow/blob/3e0639a69202486a15e4b2b8bb763754c14b236e/app/models/task.rb#L443)).

_Note:_ Because of [how we assign `MailTask`s](https://github.com/department-of-veterans-affairs/caseflow/blob/3e0639a69202486a15e4b2b8bb763754c14b236e/app/models/tasks/mail_task.rb#L39), we probably need to allow this cascading behaviour to keep happening for `MailTask`s assigned to the `MailTeam`.

## Acceptance criteria
* [ ] Stop cascading task closure if at organization boundary

lowellrex:
Resolves #10324. Prevents [cascading task closures](https://github.com/department-of-veterans-affairs/caseflow/blob/3e0639a69202486a15e4b2b8bb763754c14b236e/app/models/task.rb#L22) from occurring when organizationally-assigned parent task is of a different type than the just-closed child task. Additionally, this PR prevents us from changing the status for organizationally-assigned tasks unless those tasks are "on hold".

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -381,8 +381,8 @@ class Task < ApplicationRecord
   end
 
   def update_status_if_children_tasks_are_complete(child_task)
-    if children.any? && children.select(&:active?).empty? && on_hold?
-      if assigned_to.is_a?(Organization) && task_types_to_cascade_task_completion.include?(child_task.type)
+    if children.any? && children.active.empty? && on_hold?
+      if assigned_to.is_a?(Organization) && cascade_closure_from_child_task?(child_task)
         return update!(status: Constants.TASK_STATUSES.completed)
       end
 
@@ -390,8 +390,8 @@ class Task < ApplicationRecord
     end
   end
 
-  def task_types_to_cascade_task_completion
-    [type]
+  def cascade_closure_from_child_task?(child_task)
+    type == child_task&.type
   end
 
   def set_assigned_at
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -231,8 +231,8 @@ class Task < ApplicationRecord
     ["", ""]
   end
 
-  def when_child_task_completed
-    update_status_if_children_tasks_are_complete
+  def when_child_task_completed(child_task)
+    update_status_if_children_tasks_are_complete(child_task)
   end
 
   def task_is_assigned_to_user_within_organization?(user)
@@ -363,7 +363,7 @@ class Task < ApplicationRecord
   end
 
   def update_parent_status
-    parent.when_child_task_completed
+    parent.when_child_task_completed(self)
   end
 
   def update_children_status_after_closed; end
@@ -380,11 +380,18 @@ class Task < ApplicationRecord
     saved_change_to_attribute?(:placed_on_hold_at)
   end
 
-  def update_status_if_children_tasks_are_complete
-    if children.any? && children.select(&:active?).empty?
-      return update!(status: Constants.TASK_STATUSES.completed) if assigned_to.is_a?(Organization)
-      return update!(status: :assigned) if on_hold?
+  def update_status_if_children_tasks_are_complete(child_task)
+    if children.any? && children.select(&:active?).empty? && on_hold?
+      if assigned_to.is_a?(Organization) && task_types_to_cascade_task_completion.include?(child_task.type)
+        return update!(status: Constants.TASK_STATUSES.completed)
+      end
+
+      update!(status: Constants.TASK_STATUSES.assigned)
     end
+  end
+
+  def task_types_to_cascade_task_completion
+    [type]
   end
 
   def set_assigned_at
--- a/app/models/tasks/distribution_task.rb
+++ b/app/models/tasks/distribution_task.rb
@@ -4,14 +4,6 @@
 # Task that signals that an appeal is ready for distribution to a judge, including for auto case distribution.
 
 class DistributionTask < GenericTask
-  # Prevent this task from being marked complete
-  # when a child task (e.g. evidence submission)
-  # is marked complete because distribution tasks are used
-  # to signal that cases are ready for assignment to judges.
-  def update_status_if_children_tasks_are_complete(_child_task)
-    ready_for_distribution! if children.any? && children.active.none?
-  end
-
   def ready_for_distribution!
     update!(status: :assigned, assigned_at: Time.zone.now)
   end
--- a/app/models/tasks/distribution_task.rb
+++ b/app/models/tasks/distribution_task.rb
@@ -8,7 +8,7 @@ class DistributionTask < GenericTask
   # when a child task (e.g. evidence submission)
   # is marked complete because distribution tasks are used
   # to signal that cases are ready for assignment to judges.
-  def update_status_if_children_tasks_are_complete
+  def update_status_if_children_tasks_are_complete(_child_task)
     ready_for_distribution! if children.any? && children.active.none?
   end
 
--- a/app/models/tasks/hearing_task.rb
+++ b/app/models/tasks/hearing_task.rb
@@ -62,8 +62,8 @@ class HearingTask < GenericTask
 
   private
 
-  def task_types_to_cascade_task_completion
-    [type, NoShowHearingTask.name, DispositionTask.name]
+  def cascade_closure_from_child_task?(_child_task)
+    true
   end
 
   def set_assignee
@@ -71,7 +71,7 @@ class HearingTask < GenericTask
   end
 
   def update_status_if_children_tasks_are_complete(_child_task)
-    if children.select(&:active?).empty?
+    if children.active.empty?
       return update!(status: :cancelled) if children.select { |c| c.type == DispositionTask.name && c.cancelled? }.any?
     end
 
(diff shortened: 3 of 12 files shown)


spec/models/tasks/hearing_admin_action_task_spec.rb
```diff
@@ -68,7 +68,6 @@
     end
 
     it "finds closest_ro for veteran" do
-      VADotGovService = Fakes::VADotGovService
```
lowellrex: Unrelated to other changes in this PR but I noticed it during testing.

Fixes the following ruby warning:
```
spec/models/tasks/hearing_admin_action_task_spec.rb:71: warning: already initialized constant VADotGovService
config/initializers/va_dot_gov.rb:1: warning: previous definition of VADotGovService was here
```

app/models/task.rb
```diff
@@ -380,13 +380,20 @@ def task_just_placed_on_hold?
     saved_change_to_attribute?(:placed_on_hold_at)
   end
 
-  def update_status_if_children_tasks_are_complete
-    if children.any? && children.select(&:active?).empty?
-      return update!(status: Constants.TASK_STATUSES.completed) if assigned_to.is_a?(Organization)
-      return update!(status: :assigned) if on_hold?
+  def update_status_if_children_tasks_are_complete(child_task)
```
lowellrex: Bulk of the change starts here. We now pass a child `Task` to the parent task's `update_status_if_children_tasks_are_complete` method so the parent task can inspect the child task's type and only automatically change its status to `completed` if the child task that just completed was of the same type.

app/models/task.rb
```diff
@@ -380,13 +380,20 @@ def task_just_placed_on_hold?
...
+    if children.any? && children.active.empty? && on_hold?
```
lowellrex: Moved the condition that confirms the task was previously `on_hold` outside the block so that it includes organizational tasks.

app/models/task.rb
```diff
@@ -380,13 +380,20 @@ def task_just_placed_on_hold?
...
+      if assigned_to.is_a?(Organization) && cascade_closure_from_child_task?(child_task)
+        return update!(status: Constants.TASK_STATUSES.completed)
+      end
+
+      update!(status: Constants.TASK_STATUSES.assigned)
     end
   end
 
+  def cascade_closure_from_child_task?(child_task)
```
lowellrex: I may have prematurely extracted this method since we only override it two places and use it in one. Open to suggestions for alternatives to this approach.

app/models/tasks/distribution_task.rb
```diff
@@ -4,14 +4,6 @@
 # Task that signals that an appeal is ready for distribution to a judge, including for auto case distribution.
 
 class DistributionTask < GenericTask
-  # Prevent this task from being marked complete
-  # when a child task (e.g. evidence submission)
-  # is marked complete because distribution tasks are used
-  # to signal that cases are ready for assignment to judges.
-  def update_status_if_children_tasks_are_complete
-    ready_for_distribution! if children.any? && children.active.none?
-  end
```
lowellrex: We needed to override `update_status_if_children_tasks_are_complete` here to avoid the aggressive cascading we were doing without regard for child task type before this change. Now that we have fixed this problem we can remove this method redefinition.

app/models/tasks/hearing_task.rb
```diff
@@ -62,12 +62,16 @@ def disposition_task
...
+  end
+
   def set_assignee
     self.assigned_to = Bva.singleton
   end
 
-  def update_status_if_children_tasks_are_complete
-    if children.select(&:active?).empty?
+  def update_status_if_children_tasks_are_complete(_child_task)
+    if children.active.empty?
```
lowellrex: I think we might be able to get rid of this redefinition if we update `Task. update_status_if_children_tasks_are_complete` to set the parent status to the same status as the child task. Doing that will have to wait for a different PR though since it will involve a fair amount of research into the consequences of this decision.

app/models/tasks/schedule_hearing_task.rb
```diff
@@ -63,11 +63,6 @@ def create_parent_hearing_task
     end
   end
 
-  # We only want to take this off hold, not actually complete it, like the inherited method does
-  def update_status_if_children_tasks_are_complete
-    update!(status: :assigned) if on_hold?
-  end
```
lowellrex: We needed to override `update_status_if_children_tasks_are_complete` here to avoid the aggressive cascading we were doing without regard for child task type before this change. Now that we have fixed this problem we can remove this method redefinition.

spec/models/task_spec.rb
```diff
@@ -56,46 +75,70 @@
...
+
+      context "when task is closed" do
+        let!(:task) { FactoryBot.create(:task, :on_hold, type: Task.name, assigned_to: organization) }
+        let!(:child) { FactoryBot.create(:task, :completed, type: Task.name, parent: task) }
+
+        before { task.update!(status: Constants.TASK_STATUSES.completed) }
+
+        it "does not change the status of the task" do
+          subject
+          expect(task.status).to eq(Constants.TASK_STATUSES.completed)
```
lowellrex: New test to confirm that organizationally-assigned tasks do not change their status if they have already been closed.

spec/models/task_spec.rb
```diff
@@ -56,46 +75,70 @@
...
-          task.when_child_task_completed
+          subject
+        end
+      end
+
+      context "when child task has a different type than parent" do
+        let!(:child) { FactoryBot.create(:generic_task, :completed, parent_id: task.id) }
+        it "sets the status of the parent to assigned" do
+          subject
+          expect(task.reload.status).to eq(Constants.TASK_STATUSES.assigned)
```
lowellrex: New test to confirm cascading task closures stop when task parent task has a different type than the just-closed child task.

spec/models/tasks/no_show_hearing_task_spec.rb
```diff
@@ -84,4 +80,20 @@
...
+        assigned_to: HearingsManagement.singleton
+      )
+    end
+
+    it "sets the status of the parent NoShowHearingTask to assigned" do
+      expect(no_show_hearing_task.status).to eq(Constants.TASK_STATUSES.on_hold)
+      hearing_admin_action_task.update!(status: Constants.TASK_STATUSES.completed)
+      expect(no_show_hearing_task.status).to eq(Constants.TASK_STATUSES.assigned)
+    end
+  end
```
lowellrex: Testing specific task tree state identified during [research into existing examples](https://github.com/department-of-veterans-affairs/caseflow/issues/10324#issuecomment-494192386) where the application might close parent tasks incorrectly.

spec/models/task_spec.rb
```diff
@@ -2,52 +2,71 @@
...
         it "should change task's status to assigned" do
           status_before = task.status
-          task.when_child_task_completed
+          subject
           expect(task.status).to_not eq(status_before)
           expect(task.status).to eq("assigned")
         end
       end
 
+      context "when task is closed" do
```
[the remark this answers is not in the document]
lowellrex: Closed includes tasks that are `completed` and `cancelled`. Here is [a diagram I whipped up](https://github.com/department-of-veterans-affairs/caseflow/issues/10825#issuecomment-494016204) the other day of task statuses.

Issue #10780, opened 2019-05-15
Save edited contention text to server


pkarman:
This is step 4 of editing contention title text.

## Acceptance Criteria

* When an issue's title has been edited and the "Submit" button is clicked, the Save button activates if it isn't already active (see #10597).
* When the "Save" button is clicked the edited title text is sent to the server instead of the original text, and is stored in the `edited_description` column on the `RequestIssue`.
* All other UI alerts remain the same.

leikkisa:
@Sjones352 As part of this PR, can we disable the "edit contention text" feature during regular intakes, and only allow it during "edit"?  Thank you.

Sjones352:
@leikkisa not sure what you mean as the feature toggle is only allowed in `edit` but please send me an example where it shows on regular intake. I will also remove it from appeal as it does not apply to appeals. Please let me know if Colin said that this was good to go.

leikkisa:
Validated in UAT:
![Screen Shot 2019-06-03 at 1.02.14 PM.png](https://images.zenhubusercontent.com/58f6b6eb51aa2c30577bb60e/23201c8b-8862-41fb-80e2-7c0c36d3b6a8)

PR #10832, opened 2019-05-21
Save edited contention text to server

Sjones352:
Connects #10780


app/models/request_issue.rb
```diff
@@ -409,6 +410,10 @@ def withdraw!(withdrawal_date)
     close!(status: :withdrawn, closed_at_value: withdrawal_date.to_datetime)
   end
 
+  def edit_contention_text!(new_description)
```
pkarman: this method name is unconventional. How about `save_edit_contention_text!(new_description)`

Also, it's just one line -- why not just call `update!` instead directly?

app/models/request_issues_update.rb
```diff
@@ -167,6 +189,15 @@ def process_withdrawn_issues!
     withdrawn_issues.each { |ri| ri.withdraw!(withdrawal_date) }
   end
 
+  def process_edited_issues!
+    return if edited_issues.empty?
+
+    edited_issues.each do |ri|
+      edited_issue = @request_issues_data.detect { |issue| issue[:request_issue_id] == ri.id.to_s }
+      ri.edit_contention_text!(edited_issue[:edited_description])
```
pkarman: why not just call `update!` directly here?

spec/feature/intake/higher_level_review/edit_spec.rb
```diff
@@ -1480,6 +1480,26 @@ def click_cancel(visit_page)
...
+        expect(page).to have_button("Submit", disabled: true)
+        expect(page).to have_field(type: "textarea", match: :first, placeholder: "PTSD")
+
+        fill_in(with: "Right Knee")
+        expect(page).to have_button("Submit", disabled: false)
+        click_button("Submit")
+        expect(page).to have_content("Right Knee")
+
+        click_button("Save")
+        expect(RequestIssue.select { |issue| issue.edited_description == "Right Knee" }).to_not be_nil
```
pkarman: I think you want to call 

```ruby
expect(higher_level_review.request_issues.find { |issue| issue.edited_description == "Right Knee" }).to_not be_nil
```

instead, just to make sure you're looking at the right scope of RequestIssue.

app/models/request_issues_update.rb
```diff
@@ -92,11 +94,15 @@ def withdrawn_issues
...
+  def edited_issues
+    @edited_issues ||= edited_request_issue_ids ? fetch_edited_issues : calculate_edited_issues
+  end
+
   private
 
   def changes?
     review.request_issues.active_or_ineligible.count != @request_issues_data.count || !new_issues.empty? ||
-      !withdrawn_issues.empty?
+      !withdrawn_issues.empty? || !edited_issues.empty?
```
leikkisa: Just a style nit - we could change these to withdrawn_issues.any? and edited_issues.any?

app/models/request_issues_update.rb
```diff
@@ -118,12 +124,24 @@ def calculate_withdrawn_issues
...
   def withdrawn_issue_data
     return [] unless @request_issues_data
 
     @request_issues_data.select { |ri| !ri[:withdrawal_date].nil? && ri[:request_issue_id] }
   end
 
+  def edited_issue_data
+    return [] unless @request_issues_data
+
+    @request_issues_data.select { |ri| !ri[:edited_description].nil? && ri[:request_issue_id] }
```
leikkisa: Similar style nit - maybe we could change these to use present? instead of the double negative ```ri[:edited_description].present?```

spec/feature/intake/higher_level_review/edit_spec.rb
```diff
@@ -805,7 +805,7 @@
 
       let(:request_issues) { [request_issue, decision_request_issue, nonrating_decision_request_issue] }
 
-      it "does not remove & readd unedited issues" do
+      it "does not remove & read unedited issues" do
```
leikkisa: I think this was supposed to be re-add

spec/feature/intake/higher_level_review/edit_spec.rb
leikkisa: Here I think it'd be better to find the edited Request issue with: ```RequestIssue.where(edited_description: "Right Knee")```

app/models/request_issues_update.rb
```diff
@@ -167,6 +189,15 @@ def process_withdrawn_issues!
...
 
+  def process_edited_issues!
+    return if edited_issues.empty?
+
+    edited_issues.each do |ri|
+      edited_issue = @request_issues_data.detect { |issue| issue[:request_issue_id] == ri.id.to_s }
+      ri.edit_contention_text!(edited_issue[:edited_description])
+    end
+  end
+
```
leikkisa: Code climate is complaining about the loop inside of a loop here.

We could do 
```
edited_issue_data.each{ |edited_issue| RequestIssue.find(edited_issue[:request_issue_id]).edit_contention_text!(edited_issue[:edited_description]) }
```

PR #10836, opened 2019-05-21
Moving hearing worksheet to hearings application

sharonwarner:
Connects #10640 

### Description
This PR updates the frontend views such that the hearing worksheet and the hearings worksheet use the same erb file and front end application component.


client/app/hearingSchedule/HearingsApp.jsx
```diff
@@ -0,0 +1,162 @@
...
+  userRoleAssign: PropTypes.bool,
+  userRoleBuild: PropTypes.bool,
+  userRoleView: PropTypes.bool,
+  userRoleVso: PropTypes.bool,
+  feedbackUrl: PropTypes.string.isRequired,
+  buildDate: PropTypes.string,
+  dropdownUrls: PropTypes.array,
+  userRole: PropTypes.string,
+  userId: PropTypes.number,
+  userCssId: PropTypes.string
```
ferristseng: I think this is missing a couple prop types: `applicationUrls`, `dailyDocket`, `userRoleHearingPrep`, and `userInHearingsOrganization`, and I think `userRole` isn't being given.

client/app/hearings/containers/HearingWorksheetContainer.jsx
```diff
@@ -88,9 +88,9 @@ export class HearingWorksheetContainer extends React.Component {
 }
 
 const mapStateToProps = (state) => ({
-  worksheet: state.worksheet,
-  worksheetServerError: state.worksheetServerError,
-  fetchingWorksheet: state.fetchingWorksheet
+  worksheet: state.hearings.worksheet,
+  worksheetServerError: state.hearings.worksheetServerError,
+  fetchingWorksheet: state.hearings.fetchingWorksheet
```
ferristseng: Should these be declared in the PropTypes still if they're being mapped from the state?

sharonwarner: I couldn't tell which thumb direction meant "you're right, these shouldn't be declared."

Issue #10609, opened 2019-04-30
Do not treat "unidentified" request issues as non-comp BusinessLine


pkarman:
See https://dsva.slack.com/archives/C2ZAMLK88/p1556644556118700

We should not be attempting to create BusinessLine for unidentified non-comp RequestIssues. That we get to that place in the code smells like a stricter requirement is necessary for tracking unidentified non-comp tasks.

## Acceptance

* [ ] Re-assign existing production env tasks to a real Business Line.
* [ ] Prevent automatic creation of BusinessLine organization for unidentified non-comp issues. Alternate behavior (warning to the user, etc) TBD.
* [ ] Delete the existing BusinessLine data in UAT and prod if it exists.

lowellrex:
This bug popped back up again: https://dsva.slack.com/archives/C4JECDLSE/p1558013859294600

We fixed the immediate issue it was causing by giving a task that did not have an assignee the "unidentified" `BusinessLine`:
```
  rails c> t = Task.find(204662)
  rails c> unidentified_business_line = Organization.find(307)
  rails c> t.update!(assigned_to: unidentified_business_line)
```

lowellrex:
Same bug again (https://dsva.slack.com/archives/C4JECDLSE/p1558105150001000): 

```ruby
  rails c> tasks = Task.where(assigned_to_id: nil)
  rails c> tasks.count
  # 1
  rails c> task = tasks.first
  rails c> pp task.appeal.tasks.order(:created_at).pluck(:id, :parent_id, :type, :assigned_to_id)
  # Yup, it's another VeteranRecordRequest task without an assignee.
  # Is it also unknown benefit_type?
  rails c> task.appeal.request_issues.pluck(:benefit_type)
  # ["compensation", "compensation", "unknown", "compensation"]
  # Yup.
  # Temporary fix:
  rails c> unidentified_business_line = Organization.find(307)
  rails c> task = Task.find(206542)
  rails c> task.update!(assigned_to: unidentified_business_line)
```

pkarman:
@laurjpeterson @leikkisa we should talk through the desired behavior here. We rely on the `benefit_type` for a request issue in order to assign a LOB (BusinessLine) org task. However, unidentified issues (not rating or non-rating) don't have a `benefit_type` -- where should they be assigned?

laurjpeterson:
@laurjpeterson to schedule time to discuss with @pkarman @aroltsch @leikkisa

[...]

PR #10842, opened 2019-05-21
Do not create record request tasks for unidentified issues

leikkisa:
connects #10609

### DIFF
--- a/spec/factories/request_issue.rb
+++ b/spec/factories/request_issue.rb
@@ -25,6 +25,7 @@ FactoryBot.define do
     trait :unidentified do
       is_unidentified { true }
       unidentified_issue_text { "unidentified issue description" }
+      benefit_type { "unknown" }
     end
 
     trait :removed do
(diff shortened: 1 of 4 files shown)


spec/models/appeal_spec.rb
```diff
@@ -566,7 +566,11 @@
     end
 
     context "request issue has non-comp business line" do
-      let!(:appeal) { create(:appeal, request_issues: [create(:request_issue, benefit_type: :fiduciary)]) }
+      let!(:appeal) { create(:appeal, request_issues: [
+        create(:request_issue, benefit_type: :fiduciary),
+        create(:request_issue, benefit_type: :compensation),
+        create(:request_issue, :unidentified)
+        ]) }
```
leikkisa: After adding an unidentified issue, the test below raises an error without the code change to requires_record_request_task? (which is what we're seeing in prod).

spec/factories/request_issue.rb
```diff
@@ -25,6 +25,7 @@
     trait :unidentified do
       is_unidentified { true }
       unidentified_issue_text { "unidentified issue description" }
+      benefit_type { "unknown" }
```
pkarman: shouldn't this be "unidentified" to match `guess_benefit_type` in `RequestIssue`?

leikkisa: 👍

PR #10843, opened 2019-05-21
Automatically populate VLJ support staff correspondence docs

lowellrex:
VLJ support staff regularly request the ability to automatically populate fields for documents they prepare. This is a proof of concept of a first step in that direction.

magic-text-vlj-support

### DIFF
--- a/app/controllers/correspondence_controller.rb
+++ b/app/controllers/correspondence_controller.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class CorrespondenceController < ApplicationController
+
+  def example
+    task = Task.find(params[:id])
+
+    document = ::Caracal::Document.new
+    document.h1('Example document')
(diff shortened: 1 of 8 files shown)


app/controllers/correspondence_controller.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class CorrespondenceController < ApplicationController
+  def example
+    task = Task.find(params[:id])
+
+    document = ::Caracal::Document.new
+    document.h1("Example document")
+    document.p(format("Appeal for Veteran %s requires additional attention.", task.appeal.veteran_full_name))
```
monfresh: Why not use regular string interpolation here?
```ruby
document.p("Appeal for Veteran #{task.appeal.veteran_full_name} requires additional attention.")
```

pkarman:
archiving.

PR #10848, opened 2019-05-22
AttorneyQualityReviewTask

### LINKED ISSUE #10691: Create AttorneyQualityReviewTask class
Currently, when judge receive a quality review task [and assign it to an attorney](https://github.com/department-of-veterans-affairs/caseflow/blob/bfbe0a16ef455291bf28ed7f4de84eb9704cffbe/app/models/tasks/judge_quality_review_task.rb#L9) the application creates `AttorneyTask`s. We should instead create an instance of a new `AttorneyQualityReviewTask` class so that we can apply a specific label and custom actions.

As of this writing we have created 9 tasks that should be `AttorneyQualityReviewTask`s. We'll need to update the task type for those as well.
```ruby
  rails c> AttorneyTask.where(parent_id: JudgeQualityReviewTask.pluck(:id)).count
  # 9
```

## Acceptance Criteria
* [x] Create new `AttorneyQualityReviewTask` class
* [x] Determine what actions that class should have (should attorneys go through checkout process again?)
* [x] Update task type of existing `AttorneyTask`s that should be `AttorneyQualityReviewTask` in the production database

lomky:
Bumps #10691 

### Description
Added in the AttorneyQualityReview type

### Acceptance Criteria
- [x] JudgeQualityReviewTasks children are AttorneyQualityReviewTasks, not AttorneyTasks

### Testing Plan
1. Go to QR user page
2. Return to a Judge.
3. Become judge and go to the case
4. Return to Attorney
5. Become Attorney & confirm type.
6. Draft decision to judge
7. Confirm as Judge
8. Confirm case returns to QR.

After deploy, update the prod cases


spec/feature/queue/quality_review_flow_spec.rb
```diff
@@ -101,16 +104,14 @@
...
 
-    find(".Select-control", text: "Select a user").click
-    first("div", class: "Select-option", text: attorney_user.full_name).click
+    expect(dropdown_selected_value(find(".cf-modal-body"))).to eq attorney_user.full_name
     click_on "Submit"
 
-    expect(page).to have_content("Assigned 1 case")
+    #byebug
+    expect(page).to have_content("Task assigned to #{attorney_user.full_name}")
+    #expect(page).to have_content("Assigned 1 case")
```
lowellrex: Let's clean this up before we merge this.

lomky:
It actually isn't blocked since I made the update in `app/models/queues/attorney_queue.rb` to find this type. (there's a redundant & improved update in the BvaTask PR). Marking this as ready to go, and then we just need to remember to update the prod DB after this goes out.

Issue #9179, opened 2019-02-04
Intake | How do we handle manually Closed/Cleared EPs?


shanear:
There are all sorts of reasons why an EP could be Closed or Cleared, and not receive decisions, awards , or ratings. We are currently expecting any cleared EP to have decisions, which is not true 😢

Here's a list of all the reasons...
![image 8](https://user-images.githubusercontent.com/1596259/52233380-c2704b00-2873-11e9-9427-f3977ba0211b.png)
![image 9](https://user-images.githubusercontent.com/1596259/52233384-c43a0e80-2873-11e9-9887-850d9aaa8ea1.png)
![image 10](https://user-images.githubusercontent.com/1596259/52233387-c69c6880-2873-11e9-8752-06242f19806b.png)

requires #9188

shanear:
- We need to communicate to AMO that all 030's and 040's that received a decision MUST record that decision in either VBMS-R or VBMS-A. Clearing the EP without adding a disposition for each contention will result in loss of data.

- If an EP is cleared or closed with no decisions, then we will close all associated request issues with a closed_status of `no_decision`

shanear:
AMO meeting today, Paul said that the only reason they would clear or close an 040 or 030 without a decision is due to compliance error. It might be that we treat them the same as canceled EPs.

Waiting to follow up on that initial thought

shanear:
Even though AMO followed up on this and said this would not happen. It has already happened a handful of times in prod. We should figure out a solution to this issue, since it's causing Decision issue syncs to fail.

laurjpeterson:
Jane and @leikkisa are going through closed/cleared EPs on a weekly basis to continue investigating.

leikkisa:
SR for the data report from BGS that gives us reasons for why these were cleared: #SR-001884

leikkisa:
I sent a report yesterday to BGS with ~400 new claims.  The person who is assigned is out on vacation for a week though.

leikkisa:
Conversation to add an option to do this from the UI:
https://dsva.slack.com/archives/CFBLH2LR4/p1576000153419300

PR #10856, opened 2019-05-23
Close request issues with no decision on cleared EP

leikkisa:
connects #9179 

For the moment, I am adding this for manually processing these.  

Hopefully triggering this automatically can come soon.


app/models/request_issue.rb
```diff
@@ -406,6 +407,17 @@ def close_after_end_product_canceled!
     end
   end
 
+  def close_with_no_decision!
```
monfresh: Where/how does this method get called?

pkarman: per description:
> For the moment, I am adding this for manually processing these.

monfresh: My bad, I missed that. Thanks!

Issue #10855, opened 2019-05-22
Hearing schedule index times out loading for VSO users


tomas-nava:
The hearing schedule page has started timing out for VSO users.

[Sentry error example](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/5140/events/347499/)  

Related slack threads: [one](https://dsva.slack.com/archives/C42P25D6H/p1557860730271700), [two](https://dsva.slack.com/archives/CHX8FMP28/p1558458958437100)

Closed PRs related to this work: #10796, #10819, #10845, #10850

laurjpeterson:
@pkarman - do you think this ticket can be closed?

pkarman:
closing -- the optimization needed is tracked in other GH issues.

PR #10858, opened 2019-05-23
refactor upcoming_days_for_vso_user

tomas-nava:
- eager load appeal>tasks>assigned_to for both ama and vacols days
- refactor for brevity/readability

Connects #10855


app/models/hearing_day.rb
```diff
@@ -169,30 +169,27 @@ def upcoming_days_for_judge(start_date, end_date, user)
     end
 
     def upcoming_days_for_vso_user(start_date, end_date, user)
-      hearing_days_with_ama_hearings = HearingDay.includes(hearings: [appeal: [tasks: :assigned_to]])
-        .where("DATE(scheduled_for) between ? and ?", start_date, end_date).select do |hearing_day|
-        hearing_day.hearings.any? { |hearing| hearing.assigned_to_vso?(user) }
-      end
+      days_in_range = hearing_days_in_range(start_date, end_date)
```
tomas-nava: `hearing_days_in_range` gives us hearing days with eager-loaded appeal>tasks>assigned_to, which we use for both `ama_days` and `remaining_days` (from which `vacols_days` are selected); previously we were not eager loading these values for `vacols_days`.

app/models/hearing_day.rb
```diff
@@ -169,30 +169,27 @@ def upcoming_days_for_judge(start_date, end_date, user)
...
-        !vacols_hearings_for_remaining_hearing_days[hearing_day.id.to_s].nil? &&
-          vacols_hearings_for_remaining_hearing_days[hearing_day.id.to_s].any? do |hearing|
-            loaded_caseflow_hearings.detect { |legacy_hearing| legacy_hearing.id == hearing.id }.assigned_to_vso?(user)
-          end
+      vacols_days = remaining_days.select do |day|
+        vacols_hearing = vacols_hearings_for_remaining_days[day.id.to_s]
+
+        vacols_hearing&.any? do |hearing|
+          loaded_hearings.detect { |legacy_hearing| legacy_hearing.id == hearing.id }&.assigned_to_vso?(user)
+        end
```
tomas-nava: The use of safe navigation `vacols_hearing&.any?` takes the place of the explicit `nil?` check.

app/models/hearing_day.rb
sharonwarner: I don't think we need these values for vacols_days, but either way, this code is definitely easier to read, so it makes sense to me.

PR #10861, opened 2019-05-23
Replace old style holds with TimedHoldTask

lowellrex:
Resolves (#9207). Create timed holds for `ColocatedTask`s and `HearingAdminActionTask`s using the new timed holds instead of the old-style holds using the `on_hold_duration` field of a task. This step allows begins the transition away from the old-style holds which we will hopefully be able to remove entirely soon!


app/models/task.rb
```diff
@@ -327,10 +327,8 @@ def update_if_hold_expired!
   end
 
   def on_hold_expired?
-    return true if on_hold? && placed_on_hold_at && on_hold_duration &&
-                   placed_on_hold_at + on_hold_duration.days < Time.zone.now
-
-    false
+    !on_timed_hold? && on_hold? && placed_on_hold_at && on_hold_duration &&
+      (placed_on_hold_at + on_hold_duration.days < Time.zone.now)
```
lowellrex: If we are on a timed hold we will not expire any old-style holds.

app/models/tasks/colocated_task.rb
```diff
@@ -49,19 +49,19 @@ def label
   end
 
   def available_actions(user)
-    if assigned_to != user
-      if task_is_assigned_to_user_within_organization?(user) && Colocated.singleton.admins.include?(user)
-        return [Constants.TASK_ACTIONS.REASSIGN_TO_PERSON.to_h]
-      end
+    if assigned_to == user
+      return available_actions_with_conditions([
+                                                 appropriate_timed_hold_task_action,
```
lowellrex: This line is the only functional difference here. The other changes are just reshuffling the order of the conditional blocks.

app/models/tasks/disposition_task.rb
```diff
@@ -76,13 +76,7 @@ def no_show!
...
     end
 
-    no_show_hearing_task = NoShowHearingTask.create!(parent: self, appeal: appeal)
-
-    no_show_hearing_task.update!(
-      status: Constants.TASK_STATUSES.on_hold,
-      on_hold_duration: 25,
-      instructions: ["Mail must be received within 14 days of the original hearing date."]
-    )
+    NoShowHearingTask.create_with_hold(self)
```
lowellrex: Moved this logic inside of the `NoShowHearingTask` class.

app/models/tasks/timed_hold_task.rb
```diff
@@ -15,11 +15,15 @@ class TimedHoldTask < GenericTask
...
+        parent_task.update!(
+          instructions: [parent_task.instructions, instructions].flatten.compact,
+          # Set on_hold_duration to nil so that we override any old-style holds when we create a new timed hold.
+          on_hold_duration: nil
+        )
       end
       create!(
         appeal: parent_task.appeal,
-        assigned_by: assigned_by || parent_task.assigned_to,
+        assigned_by: assigned_by,
```
lowellrex: We were never using the `parent_task.assigned_to` fallback in the code. Removed the fallback since we want to put `NoShowHearingTask`s on hold [when they are created](https://github.com/department-of-veterans-affairs/caseflow/pull/10861/files#diff-320db9966d5a6b73985bea83a2bd9ad3R8) and they are `assigned_to` the `HearingsManagement` organization so that would become the `assigned_by` value but `assigned_by` values are only [allowed to be `User`s or `null`](https://github.com/department-of-veterans-affairs/caseflow/blob/6bca27f60dfe3cc4e6949809b725e8720dca4098/app/models/task.rb#L11).

app/models/task.rb
lowellrex: Added tests here: https://github.com/department-of-veterans-affairs/caseflow/pull/10861/files#diff-ffb5b5242cf0ba53067d359c8b8ead99R704

app/models/tasks/disposition_task.rb
lowellrex: Updated test to check for `calculated_on_hold_duration` instead of raw `on_hold_duration `: https://github.com/department-of-veterans-affairs/caseflow/pull/10861/files#diff-d371fb3a64af821be15c014d7a210242R421

app/models/tasks/colocated_task.rb
```diff
@@ -152,4 +152,9 @@ def on_hold_duration_is_set
       errors.add(:on_hold_duration, "has to be specified")
     end
   end
+
+  # ColocatedTasks on old-style holds can be placed on new timed holds which will not reset the placed_on_hold_at value.
+  def task_just_placed_on_hold?
+    super || (on_timed_hold? && children.active.where.not(type: TimedHoldTask.name).empty?)
```
tomas-nava: It seems like this'll always return `true` if there's an active `TimedHoldTask` child whether it was "just" placed on hold or not. Is there any case where that'd give us behavior we don't want? (Same question applies to `HearingAdminActionTask.task_just_placed_on_hold?`)

lowellrex: Good point. I can't think of any updates we'd do to a task that is on hold where we'd want to leave it on hold, can you?

tomas-nava: I think it's the opposite. This will pass:
```ruby
      let(:task) { FactoryBot.create(:generic_task) }

      before { TimedHoldTask.create_from_parent(task, days_on_hold: 16) }

      it "cancels the timed hold task when the task is updated" do
        task.update!(status: Constants.TASK_STATUSES.assigned)
        expect(TimedHoldTask.first.status).to eq Constants.TASK_STATUSES.cancelled
      end
```

but this will fail:
```ruby
      let(:parent) { FactoryBot.create(:schedule_hearing_task) }
      let(:task) { FactoryBot.create(:hearing_admin_action_incarcerated_veteran_task, parent: parent) }

      before { TimedHoldTask.create_from_parent(task, days_on_hold: 16) }

      it "cancels the timed hold task when the task is updated" do
        task.update!(status: Constants.TASK_STATUSES.assigned)
        expect(TimedHoldTask.first.status).to eq Constants.TASK_STATUSES.cancelled
      end
```

But that's a contrived example; I'm not sure whether we'd ever encounter a circumstance like that.

lowellrex: You're 100% right. Creating a PR to fix this behaviour (#10932).

PR #10873, opened 2019-05-23
Add AppealReporter

pkarman:
example generated SQL:

```sql
SELECT "appeals".* 
FROM   "appeals" 
       LEFT OUTER JOIN "decision_documents" 
                    ON "decision_documents"."appeal_id" = "appeals"."id" 
                       AND "decision_documents"."appeal_type" = "appeal" 
WHERE  "decision_documents"."id" IS NULL 
       AND ( "appeals"."id" NOT IN (SELECT "tasks"."appeal_id" 
                                    FROM   "tasks" 
                                    WHERE  "tasks"."appeal_type" = "Appeal" 
                                           AND "tasks"."status" IN ( 
                                               'completed', 'cancelled' 
                                                                   )) ) 
       AND ( "appeals"."id" NOT IN (SELECT "tasks"."appeal_id" 
                                    FROM   "tasks" 
                                    WHERE  "tasks"."appeal_type" = "Appeal" 
                                           AND "tasks"."type" = "RootTask" 
                                           AND "tasks"."status" = "cancelled") )
```

### DIFF
--- a/app/models/concerns/decision_documentable.rb
+++ b/app/models/concerns/decision_documentable.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module DecisionDocumentable
+  extend ActiveSupport::Concern
+
+  class_methods do
+    def has_zero_decision_documents
+      left_joins(:decision_documents).where(decision_documents: { id: nil })
+    end
+  end
+end
--- a/app/models/concerns/taskable.rb
+++ b/app/models/concerns/taskable.rb
@@ -3,6 +3,14 @@
 module Taskable
   extend ActiveSupport::Concern
 
+  class_methods do
+    def has_active_tasks
+      belongs_to_appeal_type = "belongs_to_#{name.underscore}".to_sym
+      where.not(id: Task.select(:appeal_id).send(belongs_to_appeal_type).inactive).
+        where.not(id: Task.select(:appeal_id).send(belongs_to_appeal_type).cancelled_root_task)
+    end
+  end
+
   def assigned_attorney
     tasks.includes(:assigned_to).detect { |t| t.is_a?(AttorneyTask) }.try(:assigned_to)
   end
--- a/app/models/concerns/taskable.rb
+++ b/app/models/concerns/taskable.rb
@@ -18,4 +18,8 @@ module Taskable
   def assigned_judge
     tasks.includes(:assigned_to).detect { |t| t.is_a?(JudgeTask) }.try(:assigned_to)
   end
+
+  def all_tasks_on_hold?
+    tasks.select(&:on_hold?).size == tasks.size
+  end
 end
--- a/app/models/concerns/taskable.rb
+++ b/app/models/concerns/taskable.rb
@@ -6,8 +6,8 @@ module Taskable
   class_methods do
     def has_active_tasks
       belongs_to_appeal_type = "belongs_to_#{name.underscore}".to_sym
-      where.not(id: Task.select(:appeal_id).send(belongs_to_appeal_type).inactive).
-        where.not(id: Task.select(:appeal_id).send(belongs_to_appeal_type).cancelled_root_task)
+      where.not(id: Task.select(:appeal_id).send(belongs_to_appeal_type).inactive)
+        .where.not(id: Task.select(:appeal_id).send(belongs_to_appeal_type).cancelled_root_task)
     end
   end
 
--- a/app/services/appeal_reporter.rb
+++ b/app/services/appeal_reporter.rb
@@ -1,21 +1,29 @@
 # frozen_string_literal: true
 
 class AppealReporter
-
   def stuck
-    stuck = []
-    # AMA appeal must have 1 or more tasks
-    open_appeals.find_in_batches do |appeals|
-      stuck << appeals.select(&:all_tasks_on_hold?)
-    end
-    # legacy appeal may legitimally have zero tasks
-    open_legacy_appeals.find_in_batches do |appeals|
-      stuck << appeals.select { |appeal| appeal.tasks.count > 0 && appeal.all_tasks_on_hold? }
-    end
-    stuck.flatten
+    [stuck_appeals, stuck_legacy_appeals].flatten
   end
 
   private
+
+  def stuck_appeals
+    # AMA appeal must have 1 or more tasks
+    stuck = []
+    open_appeals.find_in_batches do |appeals|
+      stuck << appeals.select(&:all_tasks_on_hold?)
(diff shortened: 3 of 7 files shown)


app/services/appeal_reporter.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+class AppealReporter
```
monfresh: For queries, I typically reach for a Query Object placed in the `queries` folder. See -[HearingRequestDistributionQuery](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/queries/hearing_request_distribution_query.rb) as an example. For this one, what do you think about the more descriptive name `StuckAppealsQuery`?

app/services/appeal_reporter.rb
```diff
@@ -0,0 +1,35 @@
...
+  def stuck
+    [stuck_appeals, stuck_legacy_appeals].flatten
+  end
+
+  private
+
+  def stuck_appeals
+    # AMA appeal must have 1 or more tasks
```
monfresh: According to this comment, for an AMA appeal to be considered stuck, it must have 1 or more tasks. However, in the spec, we are expecting the appeal with zero tasks to be returned. What do you think about using SQL instead of Ruby to perform this query. Something like this:
```ruby
def stuck_appeals
  open_appeals.joins(:tasks)
    .group("appeals.id")
    .having("count(tasks) = count(case when tasks.status = 'on_hold' then 1 end)")
end
```
This assumes that we only want appeals that have at least one task. Note that if you call `all_tasks_on_hold?` on an appeal with zero tasks, it will return `true`, which is misleading.

app/services/appeal_reporter.rb
```diff
@@ -0,0 +1,35 @@
...
+    stuck = []
+    open_appeals.find_in_batches do |appeals|
+      stuck << appeals.select(&:all_tasks_on_hold?)
+    end
+    stuck
+  end
+
+  def stuck_legacy_appeals
+    # legacy appeal may legitimally have zero tasks
```
monfresh: Same comment here about using SQL:
```ruby
def stuck_legacy_appeals
  open_legacy_appeals.joins(:tasks)
    .group("legacy_appeals.id")
    .having("count(tasks) = count(case when tasks.status = 'on_hold' then 1 end)")
end
```

spec/services/appeal_reporter_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+describe AppealReporter do
+  let!(:appeal_with_zero_tasks) { create(:appeal) }
+  let!(:appeal_with_tasks) { create(:appeal, :with_tasks) }
+  let!(:legacy_appeal_with_zero_tasks) { create(:legacy_appeal, vacols_case: create(:case)) }
+  let!(:legacy_appeal_with_tasks) { create(:task, type: "RootTask").appeal }
```
monfresh: What do you think about adding a legacy appeal with all tasks on hold to make sure it is included in the results:
```ruby
let!(:legacy_appeal_with_all_tasks_on_hold) { create(:task, :on_hold, type: "RootTask").appeal }
```

spec/services/appeal_reporter_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    schedule_hearing_task.parent.update!(parent: appeal.root_task)
+    appeal.root_task.descendants.each(&:on_hold!)
+    appeal
+  end
+
+  describe "#stuck" do
+    subject { described_class.new.stuck }
+
+    it "returns array of appeals that look stuck" do
+      expect(subject).to match_array([appeal_with_zero_tasks, appeal_with_all_tasks_on_hold])
```
monfresh: Assuming we don't want AMA appeals with zero tasks, and if we add the legacy appeal with all tasks on hold, this should be:
```ruby
expect(subject).to match_array([appeal_with_all_tasks_on_hold, legacy_appeal_with_all_tasks_on_hold])
```

app/models/concerns/decision_documentable.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module DecisionDocumentable
+  extend ActiveSupport::Concern
+
+  class_methods do
+    def has_zero_decision_documents
```
monfresh: What do you think about using `with` or `without` instead of `has`, as in `without_decision_documents`?

app/models/concerns/taskable.rb
```diff
@@ -3,11 +3,23 @@
 module Taskable
   extend ActiveSupport::Concern
 
+  class_methods do
+    def has_active_tasks
```
monfresh: `with_active_tasks`?

spec/services/appeal_reporter_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  let!(:legacy_appeal_with_zero_tasks) { create(:legacy_appeal, vacols_case: create(:case)) }
+  let!(:legacy_appeal_with_tasks) { create(:task, type: "RootTask").appeal }
+  let!(:appeal_with_all_tasks_on_hold) do
+    appeal = create(:appeal, :with_tasks)
+    schedule_hearing_task = create(:schedule_hearing_task, appeal: appeal)
+    schedule_hearing_task.parent.update!(parent: appeal.root_task)
+    appeal.root_task.descendants.each(&:on_hold!)
+    appeal
+  end
+
```
monfresh: I think we're missing some more appeals to test the `has_active_tasks` scope. When I change `open_appeals` and `open_legacy_appeals` to only use the `has_zero_decision_documents` scope, the tests still pass.

app/services/appeal_reporter.rb
```diff
@@ -0,0 +1,35 @@
...
+    stuck = []
+    open_legacy_appeals.find_in_batches do |appeals|
+      stuck << appeals.select { |appeal| appeal.tasks.count > 0 && appeal.all_tasks_on_hold? }
+    end
+    stuck
+  end
+
+  def open_appeals
+    Appeal.has_zero_decision_documents.has_active_tasks
```
monfresh: Since these scopes are only used in this context, what do you think about adding them to a new query class?

spec/services/appeal_reporter_spec.rb
monfresh: I would also add appeals with multiple tasks, but only some that are on hold, to make sure those get excluded.

pkarman:
thanks @monfresh going to try a different approach.

Issue #10078, opened 2019-03-20
Case search | Withdrawn status for AMA appeals


laurjpeterson:
When users are searching for appeals, they need to see those appeals have been withdrawn in case search, so they can quickly get context on the case. 

# AC
- If all request issues on an AMA appeal have the status of withdrawn, show the status of "Withdrawn" in Case Search

# Designs
Note: yellow is just a highlight

![status column](https://user-images.githubusercontent.com/31224257/55699110-396eb080-5997-11e9-9cf0-dbe625be22c2.png)

PR #10874, opened 2019-05-23
Withdrawn status for AMA appeals

Sjones352:
Connects #10078

Screen Shot 2019-05-23 at 9 47 12 AM


spec/feature/queue/search_spec.rb
```diff
@@ -486,4 +485,20 @@ def perform_search
...
+
+    def perform_search
+      visit "/search"
+      fill_in "searchBarEmptyList", with: caseflow_appeal.veteran_file_number
+      click_on "Search"
+    end
+   
+    it "shows withdrawn decision reviews" do
+      perform_search
+      expect(withdrawn_request_issue.closed_status).to eq("withdrawn")
```
leikkisa: We know that the withdrawn request issue has a closed status of "withdrawn" because the factory made it that way.  Here we want to check that the withdrawn status shows on the UI of the search screen.

Sjones352: @leikkisa decided to write a long test and it is now passing.

monfresh: @Sjones352 Thanks for writing the feature spec. Given that feature specs are slow, and that the feature of withdrawing request issues is already tested, what do you think about setting up the withdrawn request issues in the DB without going through the whole user flow, and then only testing the new "withdrawn" text on the search page?

spec/feature/queue/search_spec.rb
```diff
@@ -511,51 +498,9 @@ def perform_search
       click_on "Search"
     end
 
-    let(:withdraw_date) { 1.day.ago.to_date.mdY }
-
     scenario "withdraw entire review and show withdrwan on search page" do
```
leikkisa: There's a typo on 501

spec/feature/queue/search_spec.rb
```diff
@@ -511,51 +498,9 @@ def perform_search
...
-      expect(page).to have_content("This review will be withdrawn.")
-      expect(page).to have_button("Withdraw", disabled: false)
-
-      click_edit_submit
-
-      expect(page).to have_current_path("/queue/appeals/#{caseflow_appeal.uuid}")
-      expect(page).to have_content("You have successfully withdrawn a review.")
-
-      # load search page
+      caseflow_appeal.root_task.update(status: Constants.TASK_STATUSES.cancelled)
```
leikkisa: We should double check with someone how this works (I suspect it'll be the same for removed reviews), but otherwise this looks good to me. Since this is how appeal withdrawal is calculated now, I will approve this and make a follow up for this comment.

monfresh: I believe another thing we can do that might be better is to stub the call to `withdrawn?`. That way, if the implementation changes, the spec doesn't have to change. Something like this:
```ruby
allow(caseflow_appeal).to receive(:withdrawn?).and_return(true)
```

PR #10877, opened 2019-05-24
Queue/tomas/9621 create bulk assign endpoint

### LINKED ISSUE #10822: Connect BE with FE for Hearing Bulk Assignment
Tech notes: 
FE and BE is already implemented. We need modify FE to make a BE request and handle the response accordingly. 

BE PR: https://github.com/department-of-veterans-affairs/caseflow/pull/10818

AC:
As any user from Hearing Management organization I should be able to bulk assign tasks.

youngfreezyVA:
Resolves #10822


client/app/queue/components/BulkAssignModal.jsx
```diff
@@ -73,6 +74,32 @@ class BulkAssignModal extends React.PureComponent {
...
+        assigned_to_id: this.state.modal.assignedUser,
+        task_type,
+        task_count }
+      };
+
+      return ApiUtil.post('/bulk_task_assignments', { data }
+      ).
+        then(() => {
+          // make a call to to get the queue from the source of truth
+          window.location.reload();
```
aroltsch: Is that a right way of doing that? Do we want to reload the entire page?

client/app/queue/components/BulkAssignModal.jsx
```diff
@@ -73,6 +74,32 @@ class BulkAssignModal extends React.PureComponent {
...
+      const { organizationId: organization_id } = this.props;
+      const regionalOffice = _.uniq(this.props.tasks.filter((task) => {
+        return locationCity(task) === this.state.modal.regionalOffice;
+      }))[0];
+      const regionalOfficeKey = _.get(regionalOffice, 'closestRegionalOffice.key');
+
+      const data = { bulk_task_assignment: {
+        organization_id,
+        regional_office: regionalOfficeKey,
+        assigned_to_id: this.state.modal.assignedUser,
```
aroltsch: Should we also get assignedUser from the state on line 78?

.codeclimate.yml
```diff
@@ -19,6 +19,8 @@ checks:
     config:
       threshold: 25
     enabled: false
+  attribute:
+    enabled: false
```
aroltsch: Why do you need to do that?

client/app/queue/components/BulkAssignModal.jsx
```diff
@@ -241,7 +272,8 @@ BulkAssignModal.propTypes = {
   tasks: PropTypes.array.isRequired,
   organizationUrl: PropTypes.string,
   assignTasks: PropTypes.func.isRequired,
-  issueCountOptions: PropTypes.array
+  issueCountOptions: PropTypes.array,
```
aroltsch: Are we still using PropTypes everywhere?

youngfreezyVA: yes, this is being used. it's not marked as required so it's optional to pass in

.codeclimate.yml
youngfreezyVA: I didn't add this :) I thought either you or @tomas-nava did

client/app/queue/components/BulkAssignModal.jsx
youngfreezyVA: good call yeah!

client/app/queue/components/BulkAssignModal.jsx
youngfreezyVA: yes in this case I think it is. usually we use the `QueueFlowModal` which has the default `pathAfterSubmit` of `/queue` (which reloads the entire page).

.codeclimate.yml
aroltsch: It looks like you are reusing the branch? I think it is better to create a new branch next time. We should remove this commit.

client/app/hearingSchedule/components/DailyDocketRowInputs.jsx
```diff
@@ -12,6 +12,7 @@ import HearingTime from './modalForms/HearingTime';
 import { pencilSymbol } from '../../components/RenderFunctions';
 
 import { DISPOSITION_OPTIONS } from '../../hearings/constants/constants';
+import { locationCity } from '../../queue/utils';
```
aroltsch: Let's name it regionalOfficeCity to be more clear

client/app/queue/utils.js
```diff
@@ -541,3 +541,7 @@ export const sortTaskList = (taskList) => {
     new Date(prev.closedAt || prev.createdAt).getTime();
   });
 };
+
+export const locationCity = (task) => {
```
aroltsch: It looks like this method operates not only on `task`. Let's make param name more generic

client/app/queue/components/BulkAssignModal.jsx
aroltsch: I think we should remove PropTypes from this file as I think we decided not to use them anymore. We should clean up other files at some point.

client/app/queue/components/BulkAssignModal.jsx
```diff
@@ -71,6 +79,29 @@ class BulkAssignModal extends React.PureComponent {
...
+      const data = { bulk_task_assignment: {
+        organization_id,
+        regional_office: regionalOfficeKey,
+        assigned_to_id,
+        task_type,
+        task_count }
+      };
+
+      return ApiUtil.post('/bulk_task_assignments', { data }).then(() => {
+        createOrgQueueLoadPromise(this.props, '/organizations/hearing-management/tasks');
```
youngfreezyVA: @lowellrex in order to have the org queue update (without a page refresh) after the bulk assign modal is closed, I pulled the org queue load promise method into the utils file and re used that here. let me know if you have thoughts on this approach.

app/controllers/organizations/users_controller.rb
```diff
@@ -50,7 +46,7 @@ def destroy
   end
 
   def verify_organization_access
-    return if current_user.administer_org_users?
+    return if current_user.administer_org_users? || organization.users.include?(current_user)
```
lowellrex: Why do we need to allow members of the team to make requests to these endpoints? As far as I was aware the point of these endpoints is for team administration and should be locked down to only the admins of the team.

client/app/queue/components/BulkAssignModal.jsx
lowellrex: 👍 I think that makes sense.

lowellrex:
Can we add a screen capture of what this flow looks like now?

app/controllers/organizations/users_controller.rb
aroltsch: For the bulk assignment modal any user can assign tasks to another user within their organization so we need a way for any user to pull the users list. Do you see it as a problem @lowellrex?

lowellrex: I don't see any problem with folks being able to bulk assign tasks to somebody else within the organization, but this change allows anybody in the organization to manage membership in that organization.

It looks we only [use the `organization_users` element](https://github.com/department-of-veterans-affairs/caseflow/blob/6edbd5334a4c6f1284b158d1058dc549042ccd72/client/app/queue/components/BulkAssignModal.jsx#L36) of the response so we might be able to use the existing [`UsersController`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/users_controller.rb) or create a new `members` endpoint in `Organizations::UsersController` that only checks for organization membership instead of organization administrative privileges.

Alternatively, we could rename `Organizations::UsersController` to something like `Organizations::ManagementController` so that it more precisely maps to what it represents and have a new `Organizations::UsersController` that returns users in an organization (which is what we need here).

aroltsch: I like this idea! I can fix that @youngfreezy

spec/feature/queue/hearings_tasks_spec.rb
```diff
@@ -90,6 +90,35 @@
...
+            FactoryBot.create(:no_show_hearing_task)
+          end
+          success_msg = 'You have bulk assigned 4 No Show Hearing Task task(s)'
+          visit("organizations/hearing-management/")
+          click_button(text: "Assign Tasks")
+          fill_in_and_submit_bulk_assign_modal
+          expect(page).to have_content(success_msg)
+          # wait for the ajax call to reload the org queue to complete
+          using_wait_time 10 do 
+            expect(page).to have_content("Assigned (4)")
```
aroltsch: Capybara should theoretically wait for the content to appear so you should not need `using_wait_time 10`, We should avoid adding sleep and wait statements. @youngfreezyVA

PR #10879, opened 2019-05-24
Move LOB redirect from tasks.json endpoint to main .html endpoint

pkarman:
This reverts 6b3a1e35625966efde4b203bf6e1d19e61d5f0da which put a redirect on the Organization tasks endpoint. That endpoint was only accessed via XHR for JSON. The redirect properly belongs on the /organizations/:name endpoint instead, which is the main .html response to the browser that triggers the XHR JSON request.


app/controllers/decision_reviews_controller.rb
```diff
@@ -100,10 +100,6 @@ def verify_access
     redirect_to "/unauthorized"
   end
 
-  def verify_feature_enabled
```
pkarman: this feature has been on since AMA day

PR #10888, opened 2019-05-25
[WIP] Move task action definitions out of code

lowellrex:
Work in progress of a first step for how we can move definition of task actions and their conditions out of code and into self-service configurations.

### Benefits:
* Self-service.
* Immediate changes in production.
* Offload decision on business rules to those most familiar with business process.
* Reduces testing burden on developers.

### DIFF
--- a/app/models/tasks/attorney_task.rb
+++ b/app/models/tasks/attorney_task.rb
@@ -20,11 +20,7 @@ class AttorneyTask < Task
       actions: [Constants.TASK_ACTIONS.ASSIGN_TO_ATTORNEY.to_h]
     },
     {
-      conditions: [:not_assigned_to_me],
-      actions: []
-    },
-    {
-      conditions: [:ama_appeal, :on_timed_hold],
+      conditions: [:assigned_to_me, :ama_appeal, :on_timed_hold],
       actions: [
         Constants.TASK_ACTIONS.REVIEW_AMA_DECISION.to_h,
         Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h,
@@ -32,7 +28,7 @@ class AttorneyTask < Task
       ]
     },
     {
-      conditions: [:ama_appeal],
+      conditions: [:assigned_to_me, :ama_appeal],
       actions: [
         Constants.TASK_ACTIONS.REVIEW_AMA_DECISION.to_h,
         Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h,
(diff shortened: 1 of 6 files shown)


app/models/tasks/mail_task.rb
```diff
@@ -336,7 +336,7 @@ def self.default_assignee(_parent, _params)
 class VacateMotionMailTask < MailTask
   def available_actions(user)
     if LitigationSupport.singleton.user_has_access?(user)
-      return super.push(Constants.TASK_ACTIONS.LIT_SUPPORT_PULAC_CERULLO.to_h)
+      return super | [Constants.TASK_ACTIONS.LIT_SUPPORT_PULAC_CERULLO.to_h]
```
lowellrex: Use union of two arrays instead of pushing element on to array because we end up with multiple options when we call `available_actions` multiple times in the call stack.

app/models/tasks/colocated_task.rb
```diff
@@ -75,7 +75,7 @@ def available_actions_with_conditions(core_actions)
     # core_actions.unshift(Constants.TASK_ACTIONS.CHANGE_TASK_TYPE.to_h)
 
     if action == "translation" && appeal.is_a?(Appeal)
-      return ama_translation_actions(core_actions)
+      return core_actions | [Constants.TASK_ACTIONS.SEND_TO_TRANSLATION.to_h]
```
lowellrex: Following pattern of [using unions instead of pushes](https://github.com/department-of-veterans-affairs/caseflow/pull/10888/files#r287602920) and inline function call used only once for readability.

app/models/tasks/attorney_task.rb
```diff
@@ -14,23 +14,51 @@ class AttorneyTask < Task
...
+        Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h,
+        Constants.TASK_ACTIONS.END_TIMED_HOLD.to_h
+      ]
+    },
+    {
+      conditions: [:assigned_to_me, :ama_appeal],
+      actions: [
+        Constants.TASK_ACTIONS.REVIEW_AMA_DECISION.to_h,
+        Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h,
+        Constants.TASK_ACTIONS.PLACE_TIMED_HOLD.to_h
```
lowellrex: If we move logic from `available_actions` into task actions as proposed in #10889 then we could replaced this action with `Constants.TASK_ACTIONS.TOGGLE_TIMED_HOLD.to_h` and remove the hash above this one.

pkarman:
archiving

PR #10889, opened 2019-05-26
Allow TaskActions to change label and URL in repository functions

lowellrex:
Aims to simplify `Task` model by moving logic in `actions_available` method into task actions. Creates `TaskActionHelper` module and supports occasions when we want to have mutually exclusive task actions.

Aims to have `actions_available` always return an array of `TASK_ACTION`s with no intermediate logic (like `appropriate_timed_hold_task_action`) determining an element of that array. That logic gets off-loaded to the `TASK_ACTION` and executed by the `TaskActionHelper`.

This paves the way for us to determine actions available in more declarative (and verbose, and potentially repetitive) way as described in #10888.


app/models/task_action.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def build_data_attribute(func)
+    data = func ? TaskActionRepository.send(func, @task, @user) : nil
+    @data = data
+
+    if data&.delete(:returns_complete_hash)
+      @label = data[:label]
+      @value = data[:value]
+      @data = data[:data]
+    end
```
lowellrex: This implementation strikes me as a little hacky but I'm struggling to come up with something more elegant.

app/models/task_action.rb
```diff
@@ -0,0 +1,40 @@
...
+  include ActiveModel::Model
+
+  def initialize(config, task, user)
+    @task = task
+    @user = user
+
+    @label = config[:label]
+    @value = config[:value]
+
+    build_data_attribute(config[:func])
```
lowellrex: Build data attribute as part of object initialization so that this class can eventually contain the logic [that determines what type of tasks it creates](https://github.com/department-of-veterans-affairs/caseflow/blob/949b06739d7b4c98e12f846bcd762892aab23a9a/app/models/task.rb#L251).

client/constants/TASK_ACTIONS.json
```diff
@@ -145,6 +148,9 @@
         "label": "Decision ready for review",
         "value": "draft_decision/dispositions"
     },
+    "REVIEW_DECISION_DRAFT": {
+        "func": "review_decision_draft"
+    },
```
lowellrex: Replaces "[review_decision_label](https://github.com/department-of-veterans-affairs/caseflow/blob/ba248292a0b66f21da17bd1b1d015f3182543b4a/app/models/tasks/attorney_task.rb#L24)".

client/constants/TASK_ACTIONS.json
```diff
@@ -118,6 +118,9 @@
         "label": "End hold early",
         "value": "modal/end_hold"
     },
+    "TOGGLE_TIMED_HOLD": {
+        "func": "toggle_timed_hold"
+    },
```
lowellrex: Replaces "[appropriate_timed_hold_task_action](https://github.com/department-of-veterans-affairs/caseflow/blob/ba248292a0b66f21da17bd1b1d015f3182543b4a/app/models/task.rb#L76)".

app/repositories/task_action_repository.rb
```diff
@@ -226,6 +226,20 @@ def assign_to_pulac_cerullo_data(_task, _user)
...
+      action = Constants.TASK_ACTIONS.END_TIMED_HOLD.to_h if task.on_timed_hold?
+
+      TaskAction.new(action, task, user).to_h.merge(returns_complete_hash: true)
+    end
+
+    def review_decision_draft(task, user)
+      action = Constants.TASK_ACTIONS.REVIEW_LEGACY_DECISION.to_h
+      action = Constants.TASK_ACTIONS.REVIEW_AMA_DECISION.to_h if task.ama?
+
+      TaskAction.new(action, task, user).to_h.merge(returns_complete_hash: true)
```
lowellrex: I think it would be tidier to make it so the `TaskActionRepository` has no knowledge of the `TaskAction` class, but am struggling to come up with an implementation that does that.

app/helpers/task_action_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+
+      if data&.delete(:returns_complete_hash)
+        data
+      else
+        config.merge(data: data)
+      end
+    else
+      config
+    end
+  end
```
lowellrex: This method is pretty concise but I don't think it's all that clear. Would love some suggestions on how to make it clearer.

pkarman: you could gate the method like:

```ruby
def build_hash(config, task, user)
  return config unless config[:func]

  ret_val = TaskActionRepository.send(config[:func], task, user)
  
  return ret_val if ret_val&.delete(:returns_complete_hash)

  config.merge(data: ret_val)
```

fwiw I typically avoid variable names like `data`

lowellrex: That's way nicer! Changing it now.

PR #10895, opened 2019-05-28
Remove ACD feature toggles

### LINKED ISSUE #10792: ama_acd_task feature is still being toggled in tests
Automatic case distribution [has been enabled](https://github.com/department-of-veterans-affairs/appeals-deployment/pull/1974/files) since ama day, but we [still have tests](https://github.com/department-of-veterans-affairs/caseflow/search?q=ama_acd_tasks+path%3A%2Fspec&unscoped_q=ama_acd_tasks+path%3A%2Fspec) that rely on enabling and disabling this feature.

## AC
- [ ] Remove all references to the feature flag `ama_acd_task`
- [ ] Remove all references to the feature flag `ama_auto_case_distribution `

hschallhorn:
Resolves #10792

### Description
Removes all references to `ama_acd_tasks` and `ama_auto_case_distribution` toggle as they have been enabled in production and will not be disabled in the future.


spec/feature/hearings/postpone_hearing_spec.rb
```diff
@@ -34,7 +34,6 @@
   let!(:appeal) do
     create(
       :appeal,
-      :with_tasks,
```
lowellrex: Why did we need to remove this argument?

spec/feature/queue/judge_assignment_spec.rb
```diff
@@ -229,6 +229,29 @@
...
+      allow_any_instance_of(HearingRequestDocket)
+        .to receive(:distribute_appeals)
+        .and_return([])
+
+      allow_any_instance_of(LegacyDocket).to receive(:weight).and_return(101.4)
+      allow_any_instance_of(DirectReviewDocket).to receive(:weight).and_return(10)
+      allow_any_instance_of(EvidenceSubmissionDocket).to receive(:weight).and_return(95)
+      allow_any_instance_of(HearingRequestDocket).to receive(:weight).and_return(7)
+    end
+
```
lowellrex: Why do tests fail if this isn't here?

spec/models/appeal_intake_spec.rb
```diff
@@ -222,7 +222,7 @@
         decision_date: Date.new(2018, 12, 25),
         nonrating_issue_description: "nonrating request issue decision text"
       )
-      expect(intake.detail.tasks.count).to eq 1
+      expect(intake.detail.tasks.count).to eq 2
```
lowellrex: Why did this number increase?

spec/models/distribution_spec.rb
```diff
@@ -97,22 +97,96 @@
...
       expect(subject.distributed_cases.first.ready_at).to eq(2.days.ago.beginning_of_day)
       expect(subject.distributed_cases.where(priority: true).count).to eq(5)
-      expect(subject.distributed_cases.where(genpop: true).count).to eq(8)
-      expect(subject.distributed_cases.where(priority: true, genpop: false).count).to eq(2)
-      expect(subject.distributed_cases.where(priority: false, genpop_query: "not_genpop").count).to eq(1)
-      expect(subject.distributed_cases.where(priority: false, genpop_query: "any").map(&:docket_index).max).to eq(35)
+      expect(subject.distributed_cases.where(priority: true, docket: "direct_review").count).to eq(1)
+      expect(subject.distributed_cases.where(docket: "legacy").count).to be >= 8
+      expect(subject.distributed_cases.where(docket: "direct_review").count).to be >= 3
+      expect(subject.distributed_cases.where(docket: "evidence_submission").count).to eq(2)
```
lowellrex: Why did we change what we are testing here?

spec/feature/queue/judge_assignment_spec.rb
hschallhorn: When this feature was not enabled, this test was using `legacy_distribution`. Testing `ama_distribution` requires a [fair amount of setup in testing](https://github.com/department-of-veterans-affairs/caseflow/blob/c49f6b9298d6b6b82d50066488759e381cf85c95/spec/models/distribution_spec.rb#L151) for it to function properly.

spec/models/appeal_intake_spec.rb
hschallhorn: Now that `create_subtasks` is always called with the feature enabled, a distribution task is always created when [appeal intake is completed](https://github.com/department-of-veterans-affairs/caseflow/blob/c1788c300475d43fac66ac889ae37dc29393dcc0/app/models/appeal.rb#L339).

spec/models/distribution_spec.rb
hschallhorn: All these changes came from removing the legacy distribution test (as it no longer occurs) and moving the "when ama cases are in the mix" to be the main test.

spec/feature/hearings/postpone_hearing_spec.rb
hschallhorn: [`:with_tasks`](https://github.com/department-of-veterans-affairs/caseflow/blob/c1788c300475d43fac66ac889ae37dc29393dcc0/spec/factories/appeal.rb#L76) previously did not create hearing and schedule hearing tasks without the acd feature enabled. This was resulting in quite a few extra automatically created tasks causing the tests to fail. As these tasks are created manually in the test and removing the `:with_tasks` cleared up any issues, I opted to remove it as I had already spent an hour or so attempting to get the tests back to passing.

PR #10900, opened 2019-05-28
Expand reader job to users who have been active for the past month

aroltsch:
### Description
We have a Reader job that pre-fetches documents for users from VBMS to s3. The job would pre-fetch documents for all users who have a Reader role in Caseflow. It would only pre-fetch documents for cases that are actively assigned to the user (either through Caselow tasks or VACOLS case assignment). 

There are 9136 user records in Caseflow. Only 1461 user records have Reader role. Instead of looking at the Reader role, let's grab users who are actively using Caseflow (if last_login_at is no later than 7 days)

## Numbers from production: 
users = User.where("last_login_at >= ?", 1.week.ago).count
2470

## Important note
Once this is deployed, we should closely monitor the reader job to ensure we are not killing efolder.


app/jobs/fetch_documents_for_reader_user_job.rb
```diff
@@ -6,83 +6,9 @@ class FetchDocumentsForReaderUserJob < ApplicationJob
   queue_as :low_priority
   application_attr :reader
 
-  # if a user has experienced more than DOCUMENT_FAILURE_COUNT, we consider this job as failed
-  DOCUMENT_FAILURE_COUNT = 5
-
   def perform(reader_user)
-    @counts = {
```
aroltsch: Moving all this logic to separate workflow classes for easier readability.

app/workflows/fetch_documents_for_reader_job.rb
```diff
@@ -0,0 +1,53 @@
...
+  rescue Caseflow::Error::EfolderError => e
+    Rails.logger.error "Encountered #{e.class.name} when fetching documents for appeal #{appeal.id}"
+  end
+
+  def setup_debug_context
+    RequestStore.store[:current_user] = user
+    Raven.extra_context(application: "reader")
+    Raven.user_context(
+      email: user.email,
+      css_id: user.css_id,
```
pkarman: could we include the station_id here too?

app/jobs/fetch_documents_for_reader_user_job.rb
```diff
@@ -1,88 +1,14 @@
...
-
-  def log_error
-    Rails.logger.error log_message("ERROR")
-  end
-
-  def log_message(status = "SUCCESS")
-    "FetchDocumentsForReaderUserJob (user_id: #{RequestStore[:current_user].id}) #{status}. " \
-      "Retrieved #{@counts[:appeals_successful]} / #{@counts[:appeals_total]} appeals"
+    appeals = AppealsForReaderJob.new(reader_user.user).process
+    FetchDocumentsForReaderJob.new(user: reader_user.user, appeals: appeals).process
```
pkarman: I like this refactor!

app/jobs/retrieve_documents_for_reader_job.rb
```diff
@@ -6,11 +6,8 @@ class RetrieveDocumentsForReaderJob < ApplicationJob
   queue_as :low_priority
   application_attr :reader
 
-  DEFAULT_USERS_LIMIT = 3
-  def perform(args = {})
```
aroltsch: Limit is never passed as an argument

app/models/users/reader_user.rb
```diff
@@ -3,10 +3,13 @@
 class ReaderUser < ApplicationRecord
   belongs_to :user
 
+  DEFAULT_USERS_LIMIT = 10
```
aroltsch: Move limit here so we don't have it to pass it from function to function. Increase limit to 10 users at a time

app/models/users/reader_user.rb
```diff
@@ -16,19 +19,27 @@ def all_by_documents_fetched_at(limit = 10)
...
+      all_without_records.each do |user|
+        # Create ReaderUser records for these users
         ReaderUser.create(user_id: user.id)
       end
     end
 
-    def all_without_records(limit = 10)
+    # Delete Reader user records that have not been active
+    def delete_records
+      ReaderUser.joins("LEFT JOIN users ON users.id=reader_users.user_id")
```
aroltsch: Ideally I would even remove this table and move document_fetched_at to the users table but not as part of this change

app/workflows/fetch_documents_for_reader_job.rb
aroltsch: done.

app/models/users/reader_user.rb
aroltsch: I think I'll do that in the next PR. I am not a huge fan of this logic of creating and deleting new records

app/models/users/reader_user.rb
```diff
@@ -3,32 +3,43 @@
...
+        .where("last_login_at < ?", 1.week.ago)
+        .limit(DEFAULT_USERS_LIMIT).map(&:delete)
+    end
+
+    # Search through users who have been active for the past month
+    def all_without_records
```
pkarman: do we just assume that every Caseflow user is a potential Reader user? If we're going to drop `reader_users` entirely, why create more here?

aroltsch: This logic was already present here. I suggest we completely get rid of this table.

Issue #10620, opened 2019-05-01
Build OrganizationQueue from static configuration


lowellrex:
As the final step before we send the configuration from the back-end to the front-end in #10049, let's use a config in [`OrganizationQueue` itself](https://github.com/department-of-veterans-affairs/caseflow/blob/8f168920a70b82899c28b4e99e2551774c1c0c64/client/app/queue/OrganizationQueue.jsx#L36) to describe the set of tables we want to display in tabs.

Work in this direction was started in [a WIP branch for #10049](https://github.com/department-of-veterans-affairs/caseflow/compare/10049/wip_pull_tab_configs_out), which may be of use as a reference.

## 🚫 Blocked by
* [x] Use customColumns to build task table for OrganizationQueue #10619

## Acceptance criteria
* [ ] Describe tables we want to display in `OrganizationQueue` in json-like configuration
* [ ] Generate tables from the configuration

lowellrex:
SMALL. Should not need to write new tests because it should not change front-end behaviour.

PR #10912, opened 2019-05-29
Refactor BulkAssignModal

lowellrex:
In preparation for building queues from static configurations (#10620) and ultimately enabling pagination this PR moves where the `BulkAssignModal` component is called out of the `TaskTable` component so that component can be more easily built from a static configuration.

### DIFF
--- a/client/app/queue/components/BulkAssignModal.jsx
+++ b/client/app/queue/components/BulkAssignModal.jsx
@@ -1,4 +1,6 @@
-import * as React from 'react';
+import React from 'react';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
 import PropTypes from 'prop-types';
 import _ from 'lodash';
 
@@ -7,6 +9,8 @@ import ApiUtil from '../../util/ApiUtil';
 import Button from '../../components/Button';
 import Modal from '../../components/Modal';
 import Dropdown from '../../components/Dropdown';
+
+import { bulkAssignTasks } from '../QueueActions';
 
 const BULK_ASSIGN_ISSUE_COUNT = [5, 10, 20, 30, 40, 50];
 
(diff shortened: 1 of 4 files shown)


client/app/queue/components/BulkAssignModal.jsx
```diff
@@ -28,15 +32,9 @@ class BulkAssignModal extends React.PureComponent {
...
-    if (this.props.organizationUrl) {
-      ApiUtil.get(`/organizations/${this.props.organizationUrl}/users.json`).then((resp) => {
-        fetchedUsers = resp.body.organization_users.data;
-
-        this.setState({ users: fetchedUsers });
-      });
-    }
+    ApiUtil.get('/organizations/hearing-admin/users.json').then((resp) => {
+      this.setState({ users: resp.body.organization_users.data });
+    });
```
lowellrex: This component is only ever called when the `organizationUrl` is `hearing-admin`.

client/app/queue/components/BulkAssignModal.jsx
```diff
@@ -71,7 +69,7 @@ class BulkAssignModal extends React.PureComponent {
     this.setState({ showErrors: true });
 
     if (this.generateErrors().length === 0) {
-      this.props.assignTasks(this.state.modal);
+      this.props.bulkAssignTasks(this.state.modal);
```
lowellrex: We only call this component in one place. Moving as many of the attributes that we previously passed to the component into the component itself so that the calling code does not have to care about the component's implementation.

client/app/queue/components/BulkAssignModal.jsx
```diff
@@ -206,9 +204,8 @@ class BulkAssignModal extends React.PureComponent {
   }
 
   render() {
-    const isBulkAssignEnabled = this.props.enableBulkAssign && this.props.organizationUrl;
```
lowellrex: Removed `enableBulkAssign` because if the calling code is attempting to render this component then we should render it.

client/app/queue/components/TaskTable.jsx
```diff
@@ -34,7 +33,6 @@ import {
 } from '../constants';
 import COPY from '../../../COPY.json';
 import CO_LOCATED_ADMIN_ACTIONS from '../../../constants/CO_LOCATED_ADMIN_ACTIONS.json';
-import ORGANIZATION_NAMES from '../../../constants/ORGANIZATION_NAMES.json';
```
lowellrex: Inlined the only place this file was used.

PR #10914, opened 2019-05-29
Refactor TaskTable includeDetailsLink into pure detailsColumn function.

kevmo:
Connects #10619 

OrganizationQueue component then imports detailsColumn and injects into TaskTable via the customColumns property. The information regarding user role, as used in this column, has been moved to the OrganizationQueue component's props.

### DIFF
--- a/client/app/queue/components/TaskTable.jsx
+++ b/client/app/queue/components/TaskTable.jsx
@@ -73,11 +73,19 @@ export const docketNumberColumn = (tasks, requireDasRecord) => {
   };
 };
 
+<<<<<<< HEAD
 export const hearingBadgeColumn = () => {
   return {
     header: '',
     valueFunction: (task) => <HearingBadge task={task} />
   };
+=======
+export const hearingBadgeColumn = (tasks) => {
+  return {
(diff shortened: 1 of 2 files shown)


client/app/queue/components/TaskTable.jsx
```diff
@@ -80,6 +80,23 @@ export const hearingBadgeColumn = () => {
...
 };
 
+export const detailsColumn = (tasks, requireDasRecord, userRole) => {
+  return {
+    header: COPY.CASE_LIST_TABLE_VETERAN_NAME_COLUMN_TITLE,
+    valueFunction: (task) => <CaseDetailsLink
+      task={task}
+      appeal={task.appeal}
+      userRole={userRole}
+      disabled={!hasDASRecord(task, requireDasRecord)} />,
```
lowellrex: Maybe we can add this to the next PR, but can we rename this function to something like `legacyTaskHasAndRequiresDASRecord` or similar so that it more clearly states what it does? Or maybe these could be two functions (`legacyTaskHasDASRecord() && requireDasRecord`)?

Issue #10919, opened 2019-05-29
Non-unique sattyids in staff table


sharonwarner:
Email sent to Jed:

"""
When we are determining who a particular hearing is assigned to, we join the hearings table to the staff table using `hearsched.board_member = staff.sattyid`. Today, I realized there's at least one non-unique sattyid: "999". Is there a better way to join the tables to ensure uniqueness? Or is it possible to delete one of the staff entries for "999"?
"""

https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/5151/events/

Other relevant PRs: https://github.com/department-of-veterans-affairs/caseflow/pull/10850

sharonwarner:
Jed deleted the duplicate sattyid from the STAFF table. I went through our db to see if we have any duplicate records and found 10. 

```
hearing_vacols_ids = LegacyHearing.all.pluck(:vacols_id)
hearing_vacols_ids.select { |id| hearing_vacols_ids.count(id) > 1 }
=> ["602651", "602651", "602651", "611441", "611441", "613425", "613425", "613426", "613426", "613427", "613427", "613428", "613428", "613429", "613429", "622326", "622326", "623237", "623237", "623241", "623241"]
```

I'm going through these tomorrow with @lomaxap to delete unnecessary entries and then will add a unique constraint on vacols_id

PR #10920, opened 2019-05-29
Grouping duplicated hearings

sharonwarner:
Connects #10919 

### Description
We're joining VACOLS hearings to the VACOLS users table using a non-unique column. This is causing us to duplicate hearings. We'll need to figure out the correct way to join these tables, but in the meantime, this PR should silence the Sentry alerts.


app/models/vacols/case_hearing.rb
```diff
@@ -115,6 +115,7 @@ def select_hearings
         .joins("left outer join folder on folder.ticknum = brieff.bfkey")
         .joins("left outer join corres on corres.stafkey = bfcorkey")
         .where(hearing_type: HEARING_TYPES)
+        .group(:hearing_pkseq)
```
pkarman: This doesn't remove the duplicates? Just clusters them? I'm not sure how that helps.

sharonwarner:
Never mind - Jed updated the db so this is no longer needed.

Issue #10863, opened 2019-05-23
Report open Appeal and Legacy Appeals with no in-progress tasks


pkarman:
We may have Appeals or Legacy Appeals that are "stuck" with no tasks assigned or in progress.

Write a service class to report on these, call-able via Rake task or background job.

PR #10925, opened 2019-05-29
AppealsWithNoTasksOrAllTasksOnHoldQuery

pkarman:
connects #10863 

A "stuck" appeal is one that:

* has no decision documents
* has tasks that are all on hold
* in the case of an Appeal, has zero tasks


app/queries/stuck_appeals_query.rb
```diff
@@ -0,0 +1,33 @@
...
+  def stuck_appeals
+    stuck_query("Appeal") + zero_tasks_query("Appeal")
+  end
+
+  def stuck_legacy_appeals
+    stuck_query("LegacyAppeal")
+  end
+
+  def zero_tasks_query(klass_name)
+    klass_name.constantize.where.not(id: Task.select(:appeal_id).where(appeal_type: klass_name))
```
lowellrex: This translates to the following, correct?

```sql
select *
from appeals
where id not in (
  select appeal_id
  from tasks
  where appeal_type = "Appeal"
);
```

app/queries/stuck_appeals_query.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def stuck_appeals
+    stuck_query("Appeal") + zero_tasks_query("Appeal")
+  end
+
+  def stuck_legacy_appeals
+    stuck_query("LegacyAppeal")
+  end
+
+  def zero_tasks_query(klass_name)
```
lowellrex: I think we expect to have `LegacyAppeal`s without tasks in the `legacy_appeals` table if the appeal hasn't yet had any Caseflow tasks created for it (`ColocatedTask`s or `HearingTask`s would be the first, I think).

app/queries/stuck_appeals_query.rb
```diff
@@ -0,0 +1,33 @@
...
+    klass = klass_name.constantize
+    table = klass.table_name
+    klass.left_joins(:decision_documents).where(decision_documents: { id: nil })
+      .where.not(id: Task.select(:appeal_id).where(appeal_type: klass_name).inactive)
+      .where.not(id: Task.select(:appeal_id).where(appeal_type: klass_name)
+                       .where(type: "RootTask", status: Constants.TASK_STATUSES.cancelled))
+      .joins(:tasks)
+      .group("#{table}.id")
+      .having("count(tasks) = count(case when tasks.status = 'on_hold' then 1 end)")
+  end
```
lowellrex: This translates to the following for RAMP/BEAAM/AMA `Appeal`s, correct?

```sql
select appeals.*
from appeals
left join decision_documents
  on decision_documents.appeal_type = 'Appeal'
  and decision_documents.appeal_id = appeals.id
join tasks
  on tasks.appeal_type = 'Appeal'
  and tasks.appeal_id = appeals.id
where decision_documents.id is null
  and appeals.id not in (
    select appeal_id
    from tasks
    where appeal_type = 'Appeal'
      and status in ('cancelled', 'completed')
  )
  and appeals.id not in (
    select appeal_id
    from tasks
    where appeal_type = 'Appeal'
      and status = 'cancelled'
      and type = 'RootTask'
  )
group by appeals.id
having count(tasks) = count(case when tasks.status = 'on_hold' then 1 end);
```

app/queries/stuck_appeals_query.rb
pkarman: yes. that's why `zero_tasks_query` is not applied to `LegacyAppeal` when evaluating "stuck-ness"

app/queries/stuck_appeals_query.rb
lowellrex: This simplifies to the following (which can probably be optimized even further) then, right?

```sql
select appeals.*
from appeals
left join decision_documents
  on decision_documents.appeal_type = 'Appeal'
  and decision_documents.appeal_id = appeals.id
where decision_documents.id is null
  and appeals.id not in (
    select appeal_id
    from tasks
    where appeal_type = 'Appeal'
      and status <> 'on_hold'
  )
  and appeals.id in (
    select appeal_id
    from tasks
    where appeal_type = 'Appeal'
      and status = 'on_hold'
  );
```

https://caseflow-looker.va.gov/sql/cpm6rwq8wshpf7

pkarman: yeah the `and status <> 'on_hold'` vs `status in ('assigned', 'in_progress')` subtlety is where I get a little confused. I think I would like the explicitness of the latter.

OTOH, that last clause basically says "at least one task is on_hold" while the other clause says "no tasks not on_hold". My brain starts to melt a bit.

lowellrex: I actually think we need `status in ('assigned', 'in_progress')` instead of `status <> 'on_hold'` because the latter includes closed tasks and we want to exclude those.

```sql
select appeals.*
from appeals
left join decision_documents
  on decision_documents.appeal_type = 'Appeal'
  and decision_documents.appeal_id = appeals.id
where decision_documents.id is null
  -- If the appeal has active tasks then exclude it.
  and appeals.id not in (
    select appeal_id
    from tasks
    where appeal_type = 'Appeal'
      and status in ('assigned', 'in_progress')
  )
  -- If the appeal has on hold tasks (but no active tasks) include it.
  and appeals.id in (
    select appeal_id
    from tasks
    where appeal_type = 'Appeal'
      and status = 'on_hold'
  );
```

https://caseflow-looker.va.gov/sql/dn6gqhxb9ch53s

app/queries/stuck_appeals_query.rb
```diff
@@ -0,0 +1,33 @@
...
+  def zero_tasks_query(klass_name)
+    klass_name.constantize.where.not(id: Task.select(:appeal_id).where(appeal_type: klass_name))
+  end
+
+  def stuck_query(klass_name)
+    klass = klass_name.constantize
+    table = klass.table_name
+    klass.left_joins(:decision_documents).where(decision_documents: { id: nil })
+      .where.not(id: Task.select(:appeal_id).where(appeal_type: klass_name).inactive)
+      .where.not(id: Task.select(:appeal_id).where(appeal_type: klass_name)
```
aroltsch: I don't know if we really care about performance here but you could move this query `Task.select(:appeal_id).where(appeal_type: klass_name)`  into a method and reuse it in three different places.

app/queries/stuck_appeals_query.rb
aroltsch: Did we see appeals with no root tasks?

app/queries/stuck_appeals_query.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def zero_tasks_query(klass_name)
+    klass_name.constantize.where.not(id: Task.select(:appeal_id).where(appeal_type: klass_name))
+  end
+
+  def stuck_query(klass_name)
+    klass = klass_name.constantize
+    table = klass.table_name
+    klass.left_joins(:decision_documents).where(decision_documents: { id: nil })
```
aroltsch: I think including decision_documents might be an overkill. Focusing on tasks might be sufficient enough to find "stuck" appeal.

app/queries/stuck_appeals_query.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class StuckAppealsQuery
```
aroltsch: I think this name could be confusing for a new developer looking at it :)

pkarman: happy to take suggestions for alternate names!

app/queries/stuck_appeals_query.rb
pkarman: OTOH, if there are decision documents already, the appeal is "done" and not worth considering as stuck at all, even if its task history matched our other criteria.

I got this from @lowellrex originally -- thoughts?

aroltsch: I think even if there is a decision document, a BVA Dispatcher can send the case back to the judge for corrections. I think the appeal should be considered done when all tasks are completed including the root task.

app/queries/stuck_appeals_query.rb
aroltsch: `AppealsWithNoTasksOrWithAllTasksOnHoldQuery`

app/queries/stuck_appeals_query.rb
lowellrex: Great point about BVA dispatchers returning cases that already have decision documents to judges, @aroltsch! Definitely agree that we can rely only on tasks for this query.

pkarman:
@aroltsch @lowellrex this is up to date now. Could you give it one more look over? Going to invoke it in the nightly data integrity checks.

app/queries/appeals_with_no_tasks_or_all_tasks_on_hold_query.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  private
+
+  def stuck_appeals
+    stuck_query("Appeal") + zero_tasks_query("Appeal")
+  end
+
+  def stuck_legacy_appeals
+    stuck_query("LegacyAppeal")
```
lowellrex: Since [we're now using `TimedHoldTask`s](https://github.com/department-of-veterans-affairs/caseflow/pull/10861) for all timed holds (and those tasks will have a status of `assigned` when they are active) we could update the `UntrackedLegacyAppealsChecker` to check for active tasks [instead of simply tasks](https://github.com/department-of-veterans-affairs/caseflow/blob/3013d86189bb801311a3538d743ee3ddebbb88dd/app/services/untracked_legacy_appeals_checker.rb#L14).

app/queries/appeals_with_no_tasks_or_all_tasks_on_hold_query.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def zero_tasks_query(klass_name)
+    klass_name.constantize.where.not(id: Task.select(:appeal_id).where(appeal_type: klass_name))
```
lowellrex: Since we only use this for `Appeal`s, I find it more clear to write this as `Appeal.where.not(id: Task.select(:appeal_id).where(appeal_type: Appeal.name))`

app/queries/appeals_with_no_tasks_or_all_tasks_on_hold_query.rb
```diff
@@ -0,0 +1,35 @@
...
+
+class AppealsWithNoTasksOrAllTasksOnHoldQuery
+  def call
+    [stuck_appeals, stuck_legacy_appeals].flatten
+  end
+
+  private
+
+  def stuck_appeals
+    stuck_query("Appeal") + zero_tasks_query("Appeal")
```
lowellrex: If we end up moving the check for stuck `LegacyAppeal`s to `UntrackedLegacyAppealsChecker` this could become:
```ruby
Appeal.active.where.not(id: Task.select(:appeal_id).where(appeal_type: Appeal.name))
```

lowellrex: I guess we need to account for `RootTask`s so maybe this would have to be something like:
```ruby
Appeal.active.where.not(
  id: Task.visible_in_queue_table_view.select(:appeal_id).where(appeal_type: Appeal.name)
)
```

pkarman: that doesn't pass the test I wrote. I'm going to stick with the original, unless the test is wrong somehow.

PR #10926, opened 2019-05-30
Change Task Type - Back End

### LINKED ISSUE #10826: Add back-end support for changing Task type
#7126 created the front-end for changing `Task` types. This ticket exists to implement the back-end support for this new modal.

## Acceptance criteria
* [ ] Create new endpoint that accepts an existing task ID, a new task type, and any additional instructions we receive from the front-end (We can use the [Tasks/PlaceHoldController](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/tasks/place_hold_controller.rb) as a model here).
* [ ] Add code to duplicate the existing task as a sibling with the new task type and any additional instructions.
* [ ] Turn this feature on by [adding the action to the appropriate tasks](https://github.com/department-of-veterans-affairs/caseflow/pull/10693/files#diff-738d69bc1fa41b93704d5dcc4d805639R83)

hschallhorn:
Resolves #10826

### Description
Implements the back end for changing mail and colocated task types

### Acceptance Criteria
- [ ] Creates new endpoint that accepts an existing task ID, a new task type, and any additional instructions we receive from the front-end
- [ ] Add code to duplicate the existing task as a sibling with the new task type and any additional instructions.
- [ ] Turn this feature on by adding the action to the appropriate tasks


app/controllers/tasks/change_type_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+  end
+
+  def mail_sibling
+    new_class = MailTask.subclasses.find { |mt| mt.name == update_params[:action] }
+
+    default_assignee = new_class.default_assignee(task.parent, update_params)
+
+    new_class.create!(
+      appeal: task.appeal,
```
hschallhorn: Is there a cleaner way to do this? Mail tasks all have their own subclass rather than just a different `action` label.

app/controllers/tasks/change_type_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+      assigned_to: default_assignee.user_has_access?(current_user) ? task.assigned_to : default_assignee,
```
hschallhorn: Ensures the mail task is only reassigned if the current user is not on the `default_assignee` team

app/controllers/tasks/change_type_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+      assigned_by: task.assigned_by,
+      instructions: [task.instructions, update_params[:instructions]].flatten,
+      status: task.status,
+      parent: task.parent,
+      on_hold_duration: task.on_hold_duration,
```
hschallhorn: Do we want to keep a task on hold if the task type is changed?

app/controllers/tasks/change_type_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+      dupe.save!
+    end
+  end
+
+  def mail_sibling
+    new_class = MailTask.subclasses.find { |mt| mt.name == update_params[:action] }
+
+    default_assignee = new_class.default_assignee(task.parent, update_params)
+
+    new_class.create!(
```
hschallhorn: Is there a cleaner way to do this? Mail tasks all have their own subclass rather than just a different action label, so we can't just dup.

app/controllers/tasks/change_type_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+      task.slice(:appeal, :assigned_by, :status, :parent, :on_hold_duration, :placed_on_hold_at)
+      .merge(
+        assigned_to: default_assignee.user_has_access?(current_user) ? task.assigned_to : default_assignee,
```
hschallhorn: Ensures the mail task is only reassigned if the current user is not on the default_assignee team

app/controllers/tasks/change_type_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+  end
+
+  def mail_sibling
+    new_class = MailTask.subclasses.find { |mt| mt.name == update_params[:action] }
+
+    default_assignee = new_class.default_assignee(task.parent, update_params)
+
+    new_class.create!(
+      task.slice(:appeal, :assigned_by, :status, :parent, :on_hold_duration, :placed_on_hold_at)
```
hschallhorn: Do we want to keep a task on hold if the task type is changed?

lowellrex:
Do we want to change the task type of the parent task as well as the child task? `ColocatedTask`s have a parent task that is assigned to an organization (and `MailTask`s have a parent and grandparent task assigned to an organization). This "Change task type" action is available on the individually-assigned task, do we want to traverse the task tree and create a new organizationally-assigned task with the new type as well?

```
+----------------+
|                |
|    RootTask    |
|                |
+-------+--------+
        |
        |
+-------+--------+
|                |
|   JudgeTask    |
|                |
+-------+--------+
        |
        |
+-------+--------+
|                |
|  AttorneyTask  |
|                |
+-------+--------+
        |
        |
+-------+--------+
|                |
| ColocatedTask  |
| (organization) |
|                |
+-------+--------+
        |
        |
+-------+--------+
|                |
| ColocatedTask  |
|  (individual)  |
|                |
+----------------+
```

app/controllers/tasks/change_type_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+  def update_task_type
+    sibling = nil
+
+    if task.is_a? ColocatedTask
+      sibling = colocated_sibling
+    elsif task.is_a? MailTask
+      sibling = mail_sibling
+    else
+      fail Caseflow::Error::ActionForbiddenError, message: "Can only change task type of colocated or mail tasks"
+    end
```
lowellrex: Maybe we could do something like the following here:

```ruby
    if !task.is_a?(ColocatedTask) && !task.is_a?(MailTask)
      fail Caseflow::Error::ActionForbiddenError, message: "Can only change task type of colocated or mail tasks"
    end

    sibling = task.is_a?(ColocatedTask) ? colocated_sibling : mail_sibling
```

Not sure that reads any better, your call.

app/controllers/tasks/change_type_controller.rb
lowellrex: Can we pass type and action from the front-end?

app/controllers/tasks/change_type_controller.rb
lowellrex: I think we do. I think we effectively want to leave the active tree in the same state with regards to status as it was before the task type change, only changing the task type.

app/controllers/tasks/change_type_controller.rb
hschallhorn: Yeah I'm not sure. I figured this way would be easier to add another type of task that can be changed should the situation ever arise.

lowellrex: We could also add a method on each task that tells us if the task can have it's type changed.

```rub
# task.rb
def can_change_type?
  false
end

# colocated_task.rb
def can_change_type?
  false
end
```

and then we could call that here like:

```ruby
unless task.can_change_type?
  fail Caseflow::Error::ActionForbiddenError, message: "Can only change task type of colocated or mail tasks"
end
```

That seems more OOP-y

hschallhorn: Agreed!

lowellrex: Actually! I think it can be even cleaner by offloading the actual task changing function on the task classes themselves:

```ruby
# task.rb
def change_type(params)
  fail Caseflow::Error::ActionForbiddenError, message: "Cannot change task type for tasks of type #{type}"
end

# colocated_task.rb
# and 
# mail_task.rb
def change_type(params)
  # Do the actual type changing here.
end
```

hschallhorn:
> Do we want to change the task type of the parent task as well as the child task? 

Most definitely!

lowellrex:
Just tested this manually and it looks like we're running into an issue where we are double assigning the individually-assigned tasks:

```ruby
  rails c> appeal = FactoryBot.create(:appeal)
  rails c> InitialTasksFactory.new(appeal).create_root_and_sub_tasks!
  # What does our task tree look like?
  rails c> appeal.tasks.order(:created_at).pluck(:id, :parent_id, :type, :status)
  # [[283, nil, "RootTask", "on_hold"],
         [284, 283, "DistributionTask", "on_hold"],
             [285, 284, "EvidenceSubmissionWindowTask", "assigned"]]
  # Let's go create a mail task in the UI and then change its type.
  rails c> appeal.uuid
  # "2e742234-77b2-4c9c-8570-dc6043c42feb"
  # Go to queue/appeals/2e742234-77b2-4c9c-8570-dc6043c42feb and add a new mail task.
  # Created an address change task which will be assigned to VLJ support staff.
  # Let's see what the task tree looks like.
  rails c> appeal.tasks.order(:created_at).pluck(:id, :parent_id, :type, :status)
  # [[283, nil, "RootTask", "on_hold"],
         [284, 283, "DistributionTask", "on_hold"],
             [285, 284, "EvidenceSubmissionWindowTask", "assigned"],
         [286, 283, "AddressChangeMailTask", "on_hold"],
             [287, 286, "AddressChangeMailTask", "on_hold"],
                 [288, 287, "AddressChangeMailTask", "assigned"]]
  # Log in as the VLJ support staffer assigned this task and change it to FOIA.
  rails c> appeal.tasks.order(:created_at).pluck(:id, :parent_id, :type, :status)
  # [[283, nil, "RootTask", "on_hold"],
         [284, 283, "DistributionTask", "on_hold"],
             [285, 284, "EvidenceSubmissionWindowTask", "assigned"],
         [286, 283, "AddressChangeMailTask", "cancelled"],
             [287, 286, "AddressChangeMailTask", "cancelled"],
                 [288, 287, "AddressChangeMailTask", "cancelled"],
         [292, 283, "FoiaRequestMailTask", "on_hold"],
             [290, 292, "FoiaRequestMailTask", "on_hold"],
                 [291, 290, "FoiaRequestMailTask", "on_hold"],
                 [289, 290, "FoiaRequestMailTask", "in_progress"],
```

291 and 289 are both assigned to individuals. I'm pretty sure this is because we are creating the organization task and it is auto-creating a child task. I think we should just reassign the child task in this situation to the current user (if the parent task has the same organization).

Does that make sense?

lowellrex:
Making progress! Two things I noticed while manually testing that I think we need to address:
* It looks as if we are changing the parent of the task on which we called "change task type" and I don't think we should be doing that.
* I don't think we're respecting [the `default_assignee`](https://github.com/department-of-veterans-affairs/caseflow/blob/c89765f35464bbf415459e2ee9de20d0bb977d47/app/models/tasks/mail_task.rb#L151) specified by particular `MailTask` types when we should.

```ruby
  rails c> appeal = FactoryBot.create(:appeal)
  rails c> InitialTasksFactory.new(appeal).create_root_and_sub_tasks!
  # What does our task tree look like?
  rails c> appeal.tasks.order(:created_at).pluck(:id, :parent_id, :type, :status)
  # [[302, nil, "RootTask", "on_hold"],
         [303, 302, "DistributionTask", "on_hold"],
             [304, 303, "EvidenceSubmissionWindowTask", "assigned"]]
  rails c> appeal.uuid
  # "452f3995-20cf-47fa-b6eb-51a7a9e95f54"
  # Go to queue/appeals/452f3995-20cf-47fa-b6eb-51a7a9e95f54 and add a new mail task.
  # Created an address change task which will be assigned to VLJ support staff.
  # Let's see what the task tree looks like.
  rails c> appeal.tasks.order(:created_at).pluck(:id, :parent_id, :type, :status)
  # [[302, nil, "RootTask", "on_hold"],
         [303, 302, "DistributionTask", "on_hold"],
             [304, 303, "EvidenceSubmissionWindowTask", "assigned"],
         [305, 302, "AddressChangeMailTask", "on_hold"],
             [306, 305, "AddressChangeMailTask", "on_hold"],
                 [307, 306, "AddressChangeMailTask", "assigned"]]
  # Log in as the VLJ support staffer assigned this task and change it to death cert.
  rails c> appeal.tasks.order(:created_at).pluck(:id, :parent_id, :type, :status)
  # [[302, nil, "RootTask", "on_hold"],
         [303, 302, "DistributionTask", "on_hold"],
             [304, 303, "EvidenceSubmissionWindowTask", "assigned"],
         [305, 302, "AddressChangeMailTask", "cancelled"],
             [306, 305, "AddressChangeMailTask", "cancelled"],
         [308, 302, "DeathCertificateMailTask", "on_hold"],
             [309, 308, "DeathCertificateMailTask", "on_hold"],
                 [307, 309, "AddressChangeMailTask", "cancelled"],
                 [310, 309, "DeathCertificateMailTask", "in_progress"]]
  # Looks like the tasks have the correct status, but 307 should be child of 306
  # instead of 309.
  rails c> Task.find(310).assigned_to
  # 310 is assigned to the same person as 307, so that works perfectly.
  # Let's reassign this to a different team by changing task type to Congressional.
  rails c> appeal.tasks.order(:created_at).pluck(:id, :parent_id, :type, :status)
  # [[302, nil, "RootTask", "on_hold"],
         [303, 302, "DistributionTask", "on_hold"],
             [304, 303, "EvidenceSubmissionWindowTask", "assigned"],
         [305, 302, "AddressChangeMailTask", "cancelled"],
             [306, 305, "AddressChangeMailTask", "cancelled"],
         [308, 302, "DeathCertificateMailTask", "cancelled"],
             [309, 308, "DeathCertificateMailTask", "cancelled"],
                 [307, 309, "AddressChangeMailTask", "cancelled"],
         [311, 302, "CongressionalInterestMailTask", "on_hold"],
             [312, 311, "CongressionalInterestMailTask", "on_hold"],
                 [310, 312, "DeathCertificateMailTask", "cancelled"],
                 [313, 312, "CongressionalInterestMailTask", "in_progress"]]
  # Task statuses are correct. We changed the parent of 310 to 312 (should be 309).
  # Are the assignees correct?
  rails c> Task.find(311).assigned_to
  # MailTeam
  # Good.
  rails c> Task.find(312).assigned_to
  # VLJ support staff
  # Not good. Should be LitigationSupport.
  rails c> Task.find(313).assigned_to
  # Same assignee as 307 and 310. Member of VLJ support staff... I don't actually think
  # this task should be created because Lit Support does not automatically assign
  # cases unlike VLJ support staff.
```

Below is a simplified (left out unaffected branches) version of the task tree above and what state it should be in after we change its type.

image

I think we can do this by comparing the cancelled tasks to the newly created ones. A quick sketch of this logic could be:
1. Find the first ancestor of current task that has a different type. That task will be the "shared ancestor", and every task leading from the current task to that shared ancestor will the "old branch".
2. Create the new task type as a child of the shared ancestor. We will likely automatically create at least one child task. All the new tasks we create will be the "new branch".
3. Inspect the childmost task in the old branch ("old child task") and make sure it matches the childmost task in the new branch ("new child task").
    * If the old child task is assigned to an organization, do nothing more.
    * If the old child task is assigned to an individual, compare the parent task of the old child task to the childmost organizationally-assigned task in the new branch.
        * If the organizations do not match, do nothing more.
        * If the organizations match, make sure there is a child task assigned to the same individual who was assigned the old child task by creating a new child task or by changing the assignee of any automatically created child task.

lowellrex:
Looking good! The task tree ancestry looks great!

I think we're still running into the issue where we are maintaining the current assignee in some cases where we shouldn't, though.

```ruby
  rails c> appeal = FactoryBot.create(:appeal)
  rails c> FeatureToggle.enable!(:ama_acd_tasks)
  rails c> InitialTasksFactory.new(appeal).create_root_and_sub_tasks!
  # What does our task tree look like?
  rails c> appeal.tasks.order(:created_at).pluck(:id, :parent_id, :type, :status)
  # [[423, nil, "RootTask", "on_hold"],
         [424, 423, "DistributionTask", "on_hold"],
             [425, 424, "EvidenceSubmissionWindowTask", "assigned"]]
  rails c> appeal.uuid
  # "d2518a8f-1fb8-4339-8e73-df66b35f05a6"
  # Go to queue/appeals/d2518a8f-1fb8-4339-8e73-df66b35f05a6 and add a new mail task.
  # Created an address change task which will be assigned to VLJ support staff.

  # Let's see what the task tree looks like.
  rails c> appeal.tasks.order(:created_at).pluck(:id, :parent_id, :type, :status)
  # [[423, nil, "RootTask", "on_hold"],
         [424, 423, "DistributionTask", "on_hold"],
             [425, 424, "EvidenceSubmissionWindowTask", "assigned"],
         [426, 423, "AddressChangeMailTask", "on_hold"],
             [427, 426, "AddressChangeMailTask", "on_hold"],
                 [428, 427, "AddressChangeMailTask", "assigned"]]
  # Looks good.
  # Log in as the VLJ support staffer assigned this task and change it to death cert.

  # What does the task tree look like now?
  rails c> appeal.tasks.order(:created_at).pluck(:id, :parent_id, :type, :status)
  # [[423, nil, "RootTask", "on_hold"],
         [424, 423, "DistributionTask", "on_hold"],
             [425, 424, "EvidenceSubmissionWindowTask", "assigned"],
         [426, 423, "AddressChangeMailTask", "cancelled"],
             [427, 426, "AddressChangeMailTask", "cancelled"],
                 [428, 427, "AddressChangeMailTask", "cancelled"],
         [429, 423, "DeathCertificateMailTask", "on_hold"],
             [430, 429, "DeathCertificateMailTask", "on_hold"],
                 [431, 430, "DeathCertificateMailTask", "in_progress"]]
  # Ancestry tree looks good!
  rails c> Task.find(431).assigned_to
  # Current assignee looks good!

  # Let's reassign this to a different team by changing task type to Congressional.
  rails c> appeal.tasks.order(:created_at).pluck(:id, :parent_id, :type, :status)
  # [[423, nil, "RootTask", "on_hold"],
         [424, 423, "DistributionTask", "on_hold"],
             [425, 424, "EvidenceSubmissionWindowTask", "assigned"],
         [426, 423, "AddressChangeMailTask", "cancelled"],
             [427, 426, "AddressChangeMailTask", "cancelled"],
                 [428, 427, "AddressChangeMailTask", "cancelled"],
         [429, 423, "DeathCertificateMailTask", "cancelled"],
             [430, 429, "DeathCertificateMailTask", "cancelled"],
                 [431, 430, "DeathCertificateMailTask", "cancelled"],
         [432, 423, "CongressionalInterestMailTask", "on_hold"],
             [433, 432, "CongressionalInterestMailTask", "on_hold"],
                 [434, 433, "CongressionalInterestMailTask", "in_progress"]]
  # Tree looks good! How do the assignees look?
  rails c> Task.find(434).assigned_to
  # This is assigned to a member of the VLJ support staff when this task shouldn't exist
  # because LitSupport does not do automatic task assignment.
  rails c> Task.find(433).assigned_to
  # Colocated.
  # This should be assigned to LitigationSupport.
```

app/models/tasks/colocated_task.rb
```diff
@@ -85,6 +83,15 @@ def actions_available?(_user)
     open?
   end
 
+  def change_type(params)
+    dup.tap do |dupe|
+      dupe.action = params[:action]
+      dupe.instructions = params[:instructions]
+      dupe.assigned_to = Colocated.singleton
+      dupe.save!
+    end
```
lowellrex: What fields are we retaining from the duplicated task object? What is the advantage of duplicating instead of creating a new task from scratch?

app/models/task.rb
```diff
@@ -170,6 +170,47 @@ def self.modify_params(params)
     params
   end
 
+  def update_task_type(params)
```
lowellrex: Can we do all of the modifications in a transaction so that if one fails we don't end up in a half-mutated state?

app/controllers/tasks/change_type_controller.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class Tasks::ChangeTypeController < TasksController
+  def update
+    tasks = task.update_task_type(update_params)
+    tasks.each { |task_to_check| return invalid_record_error(task_to_check) unless task_to_check.valid? }
```
lowellrex: I think this is unnecessary if we [do the mutations in `update_task_type` in a transaction](https://github.com/department-of-veterans-affairs/caseflow/pull/10926/files#r290864175) since all tasks should be valid or we should have failed out of that function.

app/models/tasks/colocated_task.rb
hschallhorn: Creating makes more sense. 
`assigned_by`, `parent`, and `appeal` can be set on creation and `status`, `placed_on_hold_at`, and `on_hold_duration` should not be retained as they're automatically handled anyway.

app/models/task.rb
```diff
@@ -170,6 +170,49 @@ def self.modify_params(params)
...
+    ActiveRecord::Base.transaction do
```
lowellrex: Let's use `multi_transaction` here since we might be operating on a task associated with a legacy appeal that touches VACOLS.

app/models/task.rb
```diff
@@ -170,6 +170,49 @@ def self.modify_params(params)
...
+      new_branch_task = first_ancestor_of_type.change_type(params)
```
lowellrex: Can we change the name of the `change_type` function to something like `create_twin` that indicates we are not changing the instance's type but rather creating a duplicate of the existing task with some changes.

app/models/tasks/colocated_task.rb
```diff
@@ -85,6 +83,17 @@ def actions_available?(_user)
     open?
   end
 
+  def change_type(params)
```
lowellrex: We only use two elements from `params` here. Can we pass them as named parameters? The signature could become `change_type(action:, instructions:)` and we would not have to make any changes to the calling code because ruby (Rails?) unpacks hashes in argument lists by default.

app/models/tasks/colocated_task.rb
```diff
@@ -85,6 +83,17 @@ def actions_available?(_user)
...
+    self.class.create!(
+      appeal: appeal,
+      parent: parent,
+      assigned_by: assigned_by,
```
lowellrex: Do we want to inherit the `assigned_by` value of the previous task? Or do we want to make the person changing the task type the new assigner? Maybe a question for LP or Sneha.

app/models/task.rb
```diff
@@ -170,6 +170,49 @@ def self.modify_params(params)
...
+      new_child_task = new_branch_task.last_descendant_of_type
+      new_child_task.update!(status: status)
```
lowellrex: Do we want to inherit the status of the old task? Or do we want to leave this task with the default status of `assigned` because it is effectively a new task?

app/models/task.rb
```diff
@@ -170,6 +170,49 @@ def self.modify_params(params)
...
+      end
+
+      children.active.each { |child| child.update!(parent_id: last_descendant_of_type.id) }
+      first_ancestor_of_type.cancel_descendants
+
+      [first_ancestor_of_type.descendants, new_branch_task.first_ancestor_of_type.descendants].flatten
+    end
+  end
+
+  def assigned_to_same_org?(task_to_check)
```
lowellrex: I love all these helper functions! Super well done!

app/models/task.rb
```diff
@@ -170,6 +170,49 @@ def self.modify_params(params)
...
+      new_child_task = new_branch_task.last_descendant_of_type
+      new_child_task.update!(status: status)
+
+      if assigned_to.is_a?(User) && new_child_task.assigned_to.is_a?(User) &&
+         parent.assigned_to_same_org?(new_child_task.parent)
+        new_child_task.update!(assigned_to: assigned_to)
+      end
+
+      children.active.each { |child| child.update!(parent_id: last_descendant_of_type.id) }
+      first_ancestor_of_type.cancel_descendants
```
lowellrex: Can we add a few code comments around this code explaining why we're doing what we're doing?

app/models/task.rb
hschallhorn: I've been listening to Moncef a little to much it seems 🤔

app/models/tasks/colocated_task.rb
hschallhorn: I could go either way on this. Lets ask LP

client/app/queue/ChangeTaskTypeModal.jsx
```diff
@@ -62,7 +62,7 @@ class ChangeTaskTypeModal extends React.PureComponent {
       detail: COPY.CHANGE_TASK_TYPE_CONFIRMATION_DETAIL
     };
 
-    return this.props.requestPatch(`/tasks/${task.taskId}`, payload, successMsg).
+    return this.props.requestPatch(`/tasks/${task.taskId}/change_type`, payload, successMsg).
```
lowellrex: Can we remove the `external_id` and `type` elements from the payload? It doesn't look like we use them on the back-end.

app/models/tasks/mail_task.rb
```diff
@@ -80,11 +80,27 @@ def label
...
+      parent: parent,
+      assigned_by: assigned_by,
+      assigned_to: MailTeam.singleton
+    )
+
+    task_type.create!(
+      appeal: appeal,
+      parent: parent_task,
+      assigned_by: assigned_by,
+      assigned_to: task_type.default_assignee(parent_task, params),
```
lowellrex: It looks like we don't ever use the `params` argument when determining `default_assignee`. I'm going to prepare another PR to remove that argument from the function.

app/models/tasks/colocated_task.rb
lowellrex: Sounds good! If we need to change what we've got here in the code we can do that in a follow-up PR so we can get this one out the door.

app/models/tasks/mail_task.rb
```diff
@@ -80,11 +80,27 @@ def label
...
-  # Waiting for backend implementation before allowing user access
-  # https://github.com/department-of-veterans-affairs/caseflow/pull/10693
-  # def available_actions(user)
-  #   super(user).unshift(Constants.TASK_ACTIONS.CHANGE_TASK_TYPE.to_h)
-  # end
+  def available_actions(user)
+    super(user).present? ? super(user).unshift(Constants.TASK_ACTIONS.CHANGE_TASK_TYPE.to_h) : []
+  end
+
+  def change_type(params)
```
lowellrex: Same comment [as above](https://github.com/department-of-veterans-affairs/caseflow/pull/10926/files#r291192433) regarding the function signature (and [name](https://github.com/department-of-veterans-affairs/caseflow/pull/10926/files#r291190322)).

app/models/task.rb
```diff
@@ -170,6 +170,49 @@ def self.modify_params(params)
...
+    (parent && same_task_type?(parent)) ? parent.first_ancestor_of_type : self
+  end
+
+  def last_descendant_of_type
+    child_of_task_type = children.open.detect { |child| same_task_type?(child) }
+    child_of_task_type&.last_descendant_of_type || self
+  end
+
+  def same_task_type?(task_to_check)
+    task_to_check.slice(:class, :action, :type).eql? slice(:class, :action, :type)
```
lowellrex: Can we make this function resilient to `nil` values for `task_to_check`? Maybe something like:
```ruby
def same_task_type?(task_to_check)
  slice(:class, :action, :type).eql?(task_to_check&.slice(:class, :action, :type))
end
```

This would let us simplify `first_ancestor_of_type` as well:
```ruby
def first_ancestor_of_type
  same_task_type?(parent) ? parent.first_ancestor_of_type : self
end
```

app/models/task.rb
```diff
@@ -170,6 +170,49 @@ def self.modify_params(params)
...
+  def assigned_to_same_org?(task_to_check)
+    assigned_to.is_a?(Organization) && assigned_to == task_to_check.assigned_to
+  end
+
+  def first_ancestor_of_type
+    (parent && same_task_type?(parent)) ? parent.first_ancestor_of_type : self
+  end
+
+  def last_descendant_of_type
+    child_of_task_type = children.open.detect { |child| same_task_type?(child) }
```
lowellrex: Can we raise an error if there are multiple children for the current task? I think the behaviour in that case is a little unexpected (that it would select the first--by ID? created_at timestamp?--open child task).

app/models/task.rb
```diff
@@ -170,6 +170,49 @@ def self.modify_params(params)
...
+    end
+  end
+
+  def assigned_to_same_org?(task_to_check)
+    assigned_to.is_a?(Organization) && assigned_to == task_to_check.assigned_to
+  end
+
+  def first_ancestor_of_type
+    (parent && same_task_type?(parent)) ? parent.first_ancestor_of_type : self
+  end
```
lowellrex: Can we add a few basic tests for these new helper functions?

spec/feature/queue/mail_task_spec.rb
```diff
@@ -58,146 +58,66 @@
...
+    before do
+      OrganizationsUser.add_user_to_organization(user, Colocated.singleton)
     end
 
-    context "when task does need to be reassigned" do
-      let(:root_task) { FactoryBot.create(:root_task) }
-      let(:old_task_type) { DeathCertificateMailTask }
-      let(:new_task_type) { AodMotionMailTask }
+    it "should update the task type" do
+      # Visit case details page
```
lowellrex: We can use `step "Visit case details page" do` to segment the feature test into blocks if we'd like.

app/models/task.rb
hschallhorn: is `children.active.each { |child| child.update!(parent_id: last_descendant_of_type.id) }` even necessary? Will there ever be any children under the task we're changing the type of?

lowellrex: Good question! I don't know. I think it is nice to be a little defensive in this case (also, I'm now realizing this should probably be `children.open` instead of `children.active`). Alternatively we could raise an error here to prevent the task type change from changing and alerting us to the fact that we did end up in this situation.

app/models/tasks/mail_task.rb
lowellrex: Created PR: https://github.com/department-of-veterans-affairs/caseflow/pull/11034

PR #10931, opened 2019-05-30
Extract TaskTable column-creation methods into pure functions.

### LINKED ISSUE #10619: Use customColumns to build task table for OrganizationQueue

kevmo:
OrganizationQueue now imports pure functions from TaskTable component file and uses them to inject columns into the 3 different variants of TaskTable instances it can create.

Closes #10619. 

Also fixes docket number column misplacement that occurred in an earlier PR.

Current Hearing Management queue with Docket Number column in wrong place, as introduced in #10735:

 

Hearing Management queue with Docket Number column returned to correct place:


client/app/queue/OrganizationQueue.jsx
```diff
@@ -150,39 +151,69 @@ const mapDispatchToProps = (dispatch) => ({
...
-      detailsColumn(tasks, false, userRole),
-      taskColumn(tasks)
-    ]}
-    includeRegionalOffice={organizationName === 'Hearing Management' || organizationName === 'Hearing Admin'}
-    includeType
-    includeAssignedTo
-    includeDaysWaiting
-    tasks={tasks}
-  />
+  { (organizationName === 'Hearing Management' || organizationName === 'Hearing Admin') ?
```
lowellrex: Instead of conditionally rendering one of two components with a varying set of columns can we assign those columns to a variable and pass them into the component? Something like the following maybe:

```jsx

const TaskTableWithUserColumnTab = ({ description, tasks, organizationName, userRole }) => {

let columns = [
        docketNumberColumn(tasks, false),
        hearingBadgeColumn(tasks),
        detailsColumn(tasks, false, userRole),
        taskColumn(tasks)
      ]
if (organizationName === 'Hearing Management' || organizationName === 'Hearing Admin') {
  columns = [
        docketNumberColumn(tasks, false),
        hearingBadgeColumn(tasks),
        detailsColumn(tasks, false, userRole),
        taskColumn(tasks),
        regionalOfficeColumn(tasks)
      ]
}

return <React.Fragment>
  <p className="cf-margin-top-0">{description}</p>

  <TaskTable
    customColumns={columns}
    includeType
    includeAssignedTo
    includeDaysWaiting
    tasks={tasks}
  />
</React.Fragment>;
};
```

PR #10932, opened 2019-05-30
Fix timed hold task cancellation behaviour

lowellrex:
Creating a child task for a task that is on a timed hold should cancel the timed hold. It currently doesn't but this PR fixes that.


app/models/task.rb
```diff
@@ -357,6 +361,10 @@ def update_parent_status
     parent.when_child_task_completed(self)
   end
 
+  def tell_parent_task_child_task_created
+    parent&.when_child_task_created(self)
+  end
```
lowellrex: Mirroring the pattern we established with `update_parent_status`.

app/models/task.rb
```diff
@@ -228,6 +227,11 @@ def when_child_task_completed(child_task)
     update_status_if_children_tasks_are_complete(child_task)
   end
 
+  def when_child_task_created(child_task)
+    cancel_timed_hold unless child_task.is_a?(TimedHoldTask)
+    update!(status: :on_hold)
+  end
```
lowellrex: When we create a child task we update the parent's status to on hold (were already doing this) and cancel any existing timed holds.

app/models/task.rb
```diff
@@ -17,12 +17,11 @@ class Task < ApplicationRecord
 
   before_create :set_assigned_at
   after_create :create_and_auto_assign_child_task, if: :automatically_assign_org_task?
-  after_create :put_parent_on_hold
+  after_create :tell_parent_task_child_task_created
 
   before_update :set_timestamps
   after_update :update_parent_status, if: :task_just_closed_and_has_parent?
   after_update :update_children_status_after_closed, if: :task_just_closed?
-  after_update :cancel_timed_hold, unless: :task_just_placed_on_hold?
```
lowellrex: Bundled this behaviour into the function that was previously called `cancel_timed_hold`

PR #10936, opened 2019-05-30
update docker-demo

enriquemanuel:
docker demo to use latest vacols ecr image


docker-compose-demo.yml
```diff
@@ -14,18 +14,11 @@ services:
       - "${TMPDIR:-/tmp/localstack}:/tmp/localstack"
 
   vacols-db-development:
-    build: ./local/vacols/.
     container_name: VACOLS_DB-development
-    image: vacols_db
-    tty: true
-    stdin_open: true
+    image: vacols_db:latest
     command: bash -c "/home/oracle/setup/dockerInitWithSchema.sh"
```
lisac-usds: just to check: will we still need to run this `dockerInitWithSchema.sh`?

enriquemanuel: You are right, not needed anymore, removing it

PR #10940, opened 2019-05-30
add classes to task factories

tomas-nava:
### Description
Adds class declarations to all the factories that were missing them.


spec/models/tasks/decision_review_task_spec.rb
```diff
@@ -60,7 +60,7 @@
         }
       ]
     end
-    let(:task) { create(:task, appeal: hlr, status: task_status).becomes(described_class) }
+    let(:task) { create(:higher_level_review_task, appeal: hlr, status: task_status) }
```
lomky: Was this a mistake before?

tomas-nava: No, this would've worked as intended, it was just a roundabout way to get there.

PR #10952, opened 2019-05-30
Prettier format JSON constants

ferristseng:
### Description

  * Add prettier as a dev dependency
  * Run prettier in `lint:fix`
  * Run prettier on `client/constants/TASK_ACTIONS.json`


client/package.json
```diff
@@ -11,7 +11,7 @@
     "test:karma:watch": "yarn run test:karma --no-single-run",
     "test:legacy:watch": "yarn run test:legacy --watch",
     "lint": "eslint . --ext .jsx --ext .js --max-warnings 0 --cache",
-    "lint:fix": "yarn run lint --fix",
+    "lint:fix": "yarn run lint --fix && yarn run prettier constants/TASK_ACTIONS.json --write",
```
ferristseng: We could format everything in the constants folder with `constants/*` here. Wondering if it would be a headache if folks were editing these on branches though

lowellrex: Can we make this it's own script that we can call with or without the `--write` flag? That way we can [add it to our linting rake task](https://github.com/department-of-veterans-affairs/caseflow/blob/master/lib/tasks/lint.rake) and fail CI if the `TASK_ACTIONS.json` file is not formatted.

PR #10960, opened 2019-05-31
Add PrivateBar class and creation UI

lowellrex:
Resolves #9517.

add-private-bar

Now that we've [tested that a `FieldVso`-style approach works](https://github.com/department-of-veterans-affairs/caseflow/issues/9527#issuecomment-492644994), we can create a specific class for this entity and give Caseflow admins the ability to create these in the UI.

### TODO:
After this is merged and pushed live, we should update the two existing `FieldVso` organizations for private bar attorneys to have the correct type (`PrivateBar`).


client/app/queue/TeamManagement.jsx
```diff
@@ -28,6 +28,8 @@ class TeamManagement extends React.PureComponent {
 
   addIhpWritingVso = () => this.props.history.push('/team_management/add_vso');
 
+  addPrivateBar = () => this.props.history.push('/team_management/add_private_bar');
+
```
lowellrex: The front-end for this UI is entirely untested. We previously decided that was fine because 1) we test the back-end so we are sure we won't accept or return garbage data, and 2) this will only be used by Caseflow personnel who will let the development team know when something goes wrong, and 3) there are alternative ways to create/update/delete organizations (rails console, for one).

Happy to revisit that discussion if we'd like now since we're adding more functionality to this interface (though we are _not_ opening this administration page up to more users).

app/controllers/team_management_controller.rb
```diff
@@ -9,6 +9,7 @@ def index
       format.json do
         render json: {
           judge_teams: JudgeTeam.all.order(:id).map { |jt| serialize_org(jt) },
+          private_bars: PrivateBar.all.order(:id).map { |private_bar| serialize_org(private_bar) },
           vsos: Vso.all.order(:id).map { |vso| serialize_org(vso) },
```
kevmo: 3 different invocations of the same pattern 
```ruby
orgs: Org.all.order(:id).map { |o| serialize_org(o) }

```
Fine for now, but maybe logic should eventually be moved to the `Organization` active record model, since all 3 of these inherit from it.

client/app/queue/AddPrivateBarModal.jsx
```diff
@@ -0,0 +1,79 @@
...
+import { bindActionCreators } from 'redux';
+import { connect } from 'react-redux';
+import { onReceiveNewPrivateBar } from './teamManagement/actions';
+import {
+  requestSave,
+  showErrorMessage
+} from './uiReducer/uiActions';
+import TextField from '../components/TextField';
+import { withRouter } from 'react-router-dom';
+import QueueFlowModal from './components/QueueFlowModal';
```
kevmo: I generally import in the following order: standard libs, 3rd party libs, and then whatever files I wrote in a project. Not a PR blocker.

spec/models/organizations/private_bar_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+    let(:private_bar) { PrivateBar.create!(name: "Caseflow Law Group", url: "caseflow-law") }
+    let(:appeal) { FactoryBot.create(:appeal) }
+
+    before do
+      allow(appeal).to receive(:representatives).and_return(PrivateBar.where(id: private_bar.id))
+    end
+
+    it "creates a representative that does not write IHPs for appeals they represent" do
+      expect(appeal.representatives.include?(private_bar)).to eq(true)
+      expect(private_bar.should_write_ihp?(appeal)).to eq(false)
```
kevmo: Good.

client/app/queue/TeamManagement.jsx
kevmo: Not opening it to more users? Let's keep going.

app/controllers/team_management_controller.rb
lowellrex: The logic being the ordering? Or the ordering and the serialization?

Issue #10794, opened 2019-05-16
Add Update Contention to VBMS Service


leikkisa:
### Acceptance criteria
- VBMSService in Caseflow has a method for updating contention text in VBMS
- Caseflow's gemfile and gemfile.lock point to the latest version of connect_vbms

### Technical notes
This will be added in caseflow/app/services/external_api/vbms_service.rb
and in caseflow/lib/fakes/vbms_service.rb

### Validation
In the console for Caseflow UAT:
- Find an open End Product Establishment for a HLR or SC
- Get a contention on that EPE
- Update the text
- submit the edited contention to VBMS with the update_contention call
- double check that the contention now has the new contention text

```
c = epe.contentions.first
c.text = "edited description"
VBMSService.update_contention!(c)
```

depends on: https://github.com/department-of-veterans-affairs/connect_vbms/issues/220

leikkisa:
Validated in UAT:

![Screen Shot 2019-06-03 at 12.15.14 PM.png](https://images.zenhubusercontent.com/58f6b6eb51aa2c30577bb60e/e6f81190-661d-4e4c-8e1a-539e99bce104)

PR #10969, opened 2019-05-31
Add Update Contention to VBMS Service

Sjones352:
Connects  #10794
 I need to add tests to this Pr


lib/fakes/vbms_service.rb
```diff
@@ -194,6 +194,13 @@ def self.remove_contention!(contention)
     true
   end
 
+  def self.updated_contention!(contention)
```
leikkisa: this has an extra "d", it should be update_contention! to match the VBMS library.

Gemfile
```diff
@@ -13,7 +13,7 @@ gem "bgs", git: "https://github.com/department-of-veterans-affairs/ruby-bgs.git"
 gem "bootsnap", require: false
 gem "business_time", "~> 0.9.3"
 gem "caseflow", git: "https://github.com/department-of-veterans-affairs/caseflow-commons", ref: "8dde00d67b7c629e4b871f8dcb3617bfe989b3db"
-gem "connect_vbms", git: "https://github.com/department-of-veterans-affairs/connect_vbms.git", ref: "dddc821c2335c7de234a5454e4b4874e3f658420"
+gem "connect_vbms", git: "https://github.com/department-of-veterans-affairs/connect_vbms.git", ref: "603c7385bd5f021286a3f4f58fb2d99c772a4ca1"
```
leikkisa: If you run bundle install it will update the Gemfile.lock file too.

PR #10971, opened 2019-05-31
download functionality

enriquemanuel:
This will allow the rake command to just download the facols image from ECR, instead of trying to build it.


local/vacols/build_push.sh
```diff
@@ -101,6 +101,15 @@ push(){
 
 }
 
+download(){
+  # get circleci latest image from this same repo
+  facols_image=$(cat ../../.circleci/config.yml| grep facols | awk '{print $3}')
```
pkarman: does this relative path require the script be invoked from a particular place? or is it relative to the script itself?

enriquemanuel: is relative path to the script `build_push.sh` so it can go all the way up and read the `.circleci` configs

enriquemanuel:
@pkarman do you mind trying it before we merge it?

PR #10979, opened 2019-06-03
Add efolder_documents_fetched_at to users table

aroltsch:
### Description
Reader job was recently changed to run for users who are active within a week that means we sould have to delete and re-create records in ReaderUser table. ReaderUser model appears to be over-loaded with unnecessary logic. Move efolder_documents_fetched_at to the users table to simplify the logic and code around it. 

### To do:
Delete reader_users table and all the code around it after this code is deployed

*Schema Changes Only*

* [x] Column comments updated


app/workflows/find_users_in_batches_for_reader_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class FindUsersInBatchesForReaderJob
```
pkarman: this seems like it belongs in the `app/queries` space instead since it is basically just SQL generation. Maybe `BatchUsersForReaderQuery` or something like that?

db/migrate/20190603150958_add_efolder_documents_fetched_at_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddEfolderDocumentsFetchedAtToUsers < ActiveRecord::Migration[5.1]
+  def change
+  	add_column :users, :efolder_documents_fetched_at, :datetime, comment: "Date when efolder documents were cached in s3 for this user"
```
pkarman: is that a tab indent?

spec/jobs/retrieve_documents_for_reader_job_spec.rb
```diff
@@ -6,45 +6,25 @@
...
-          Generators::ReaderUser.create(user_id: inactive_user.id)
-        end
-
-        it "should only run FindDocumentsForReaderUserJob for the active user" do
-          expect(FetchDocumentsForReaderUserJob).to receive(:perform_later).once do |reader_user|
-            expect(reader_user.user.id).to eq(active_user.id)
-          end
+      context "if there are active and inactive users" do
+        it "should only run the job for the active user with expired or nil efolder_documents_fetched_at" do
+          expect(FetchDocumentsForReaderUserJob).to receive(:perform_later).exactly(3).times
```
pkarman: could we add a `with` clause to the `expect ... to .. receive` to verify which user object(s) are being passed? i.e. how do we know that the `3` times correspond to the "active user with expired or nil ..." conditions?

spec/workflows/find_users_in_batches_for_reader_job_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require "rails_helper"
```
pkarman: I don't think `rails_helper` is strictly necessary -- rspec always loads it by default.

spec/workflows/find_users_in_batches_for_reader_job_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+    let!(:inactive_user) do
+      create(:user, last_login_at: 3.months.ago )
+    end
+
+    subject { FindUsersInBatchesForReaderJob.process }
+
+    it "should return active user with expired or nil efolder_documents_fetched_at" do
+      # should first process the user with nil efolder_documents_fetched_at
+      expect(subject[0].id).to eq active_user4.id
```
pkarman: how about `expect(subject).to eq([active_user4, active_user2, active_user1])` instead for brevity?

spec/jobs/retrieve_documents_for_reader_job_spec.rb
aroltsch: I don't think it is possible to check for all three objects or is it? @pkarman

aroltsch: I test the specific users in the workflow class so I thought that was enough.

app/workflows/find_users_in_batches_for_reader_job.rb
aroltsch: Good suggestion!

spec/workflows/find_users_in_batches_for_reader_job_spec.rb
aroltsch: you are right. That's what happens when you copy and paste without thinking :)

spec/jobs/retrieve_documents_for_reader_job_spec.rb
pkarman: check out #10989

PR #10981, opened 2019-06-03
Temporary fix for db/seeds

lowellrex:
Allows us to run `bundle exec rake db:seed` in our local development environments again. This will be removed when we get rid of the `ama_acd_tasks` feature flag.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -91,6 +91,7 @@ class SeedDB
       decision_review: appeal
     )
 
+    FeatureToggle.enable!(:ama_acd_tasks)
     InitialTasksFactory.new(appeal).create_root_and_sub_tasks!
 
     # Completing the evidence submission task will mark the appeal ready for distribution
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -91,6 +91,7 @@ def create_ama_distribution_tasks
       decision_review: appeal
     )
 
+    FeatureToggle.enable!(:ama_acd_tasks)
```
lowellrex: There are a handful of rubocop changes in here, but this is one of 2 meaningful changes that fix db/seeds.

db/seeds.rb
```diff
@@ -536,19 +537,15 @@ def create_ama_case_with_open_schedule_hearing_task(ro_key)
 
     vet.save
 
-    appeal = FactoryBot.create(
+    FeatureToggle.enable!(:ama_acd_tasks)
+    FactoryBot.create(
```
lowellrex: This is the other one.

Issue #10049, opened 2019-03-19
Pass columns to display in queue table view from back-end


lowellrex:
In preparation to [consolidate the three individual queues](https://github.com/department-of-veterans-affairs/caseflow/issues/10032) let's pass the set of columns to display in the table from the back-end so the front-end can be agnostic to the user's role. Each of the individual queue views (and the tabs within those queues) differ slightly so passing these from the back-end (where we will always know the user's role) should allow us to present the correct set of columns.

This work also paves the way for a future where individual people or teams can [configure their queue table view](https://github.com/department-of-veterans-affairs/caseflow/issues/8423) to show a custom set of columns.

## 🚫 Blocked by
* [x] Use customColumns to build task table for OrganizationQueue #10619
* [ ] Build OrganizationQueue from static configuration #10620

## Acceptance criteria
* [ ] Pass columns to display in queue table view from back-end

lowellrex:
LARGE. Limit this to only organization queues to start. Be aggressive in looking for other ways to break up PRs into bite-sized chunks.

lowellrex:
Started a branch for some progress on this PR: [`10049/wip_pull_tab_configs_out`](https://github.com/department-of-veterans-affairs/caseflow/tree/10049/wip_pull_tab_configs_out)

PR #10983, opened 2019-06-03
Use established modal pattern for bulk task assignment

lowellrex:
In the course of working on #10049, I noticed that we use a different pattern for displaying the bulk task assign modal. This PR makes it so we use our normal pattern so that we can put the `createOrgQueueLoadPromise()` definition back in the `OrganizationQueueLoadingScreen` component.


client/app/queue/components/BulkAssignModal.jsx
```diff
@@ -78,26 +68,24 @@ class BulkAssignModal extends React.PureComponent {
...
         regional_office: regionalOfficeKey,
         assigned_to_id,
         task_type,
         task_count }
       };
 
       return ApiUtil.post('/bulk_task_assignments', { data }).then(() => {
-        createOrgQueueLoadPromise(this.props, '/organizations/hearing-management/tasks');
+        this.props.history.push(`/organizations/${this.organizationUrl()}`);
+        window.location.reload();
```
lowellrex: We refresh now instead of calling this method so that we can move this method out of `queue/utils` and avoid creating a circular dependency in #10049.

client/app/queue/components/BulkAssignModal.jsx
```diff
@@ -195,7 +183,7 @@ class BulkAssignModal extends React.PureComponent {
 
   generateNumberOfTaskOptions = () => {
     const actualOptions = [];
-    const issueCounts = this.props.issueCountOptions || BULK_ASSIGN_ISSUE_COUNT;
+    const issueCounts = BULK_ASSIGN_ISSUE_COUNT;
```
lowellrex: We didn't pass that property the one place we called this component. We can add it back in later if we end up wanting custom task counts.

spec/feature/queue/hearings_tasks_spec.rb
```diff
@@ -93,23 +93,23 @@
...
         end
 
         it "is able to bulk assign tasks for the hearing management org" do
           3.times do
             FactoryBot.create(:no_show_hearing_task)
           end
-          success_msg = 'You have bulk assigned 4 No Show Hearing Task task(s)'
-          visit("organizations/hearing-management/")
+          success_msg = "You have bulk assigned 4 No Show Hearing Task task(s)"
+          visit("/organizations/hearing-management")
```
lowellrex: This is the meaningful change in this test.

PR #10988, opened 2019-06-03
Organization clean_url means voc_rehab -> voc-rehab

pkarman:
**Why**

E.g. previously we did a `find` using the `url` "voc_rehab". Finding zero results, we tried to create it using "voc-rehab" because the `clean_url` before_save hook in Organization model altered the string. That means we would try to insert a duplicate organization.

This change assumes every BusinessLine type Organization has a unique name and that is sufficient for the initial `find`.

### DIFF
--- a/app/models/concerns/has_business_line.rb
+++ b/app/models/concerns/has_business_line.rb
@@ -5,9 +5,7 @@ module HasBusinessLine
 
   def business_line
     business_line_name = Constants::BENEFIT_TYPES[benefit_type]
-    @business_line ||= BusinessLine.find_or_create_by(name: business_line_name) do |org|
-      org.url = benefit_type
-    end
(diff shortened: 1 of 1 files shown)


app/models/concerns/has_business_line.rb
```diff
@@ -5,7 +5,9 @@ module HasBusinessLine
 
   def business_line
     business_line_name = Constants::BENEFIT_TYPES[benefit_type]
-    @business_line ||= BusinessLine.find_or_create_by(url: benefit_type, name: business_line_name)
+    @business_line ||= BusinessLine.find_or_create_by(name: business_line_name) do |org|
```
lomky: Minor nitpick, but I don't love `||=` on a multiline statement with a block.

PR #10994, opened 2019-06-04
Async jobs report no pagination

pkarman:
The nightly job was paginating by default, which meant we only ever saw max of 50 stuck async jobs.

Passing a value less than zero (e.g. "-1") will turn off pagination.


app/services/asyncable_jobs.rb
```diff
@@ -29,6 +29,12 @@ def gather_jobs
       expired_jobs << klass.potentially_stuck
     end
     jobs = expired_jobs.flatten.sort_by(&:sort_by_last_submitted_at)
+
+    return paginated_jobs(jobs) if page_size > 0
+    return jobs
```
pkarman: if `page_size == 0` this will return all jobs, which seems like what we want.

PR #11006, opened 2019-06-04
Pass queue table config to front-end from back-end

### LINKED ISSUE #10049: Pass columns to display in queue table view from back-end

lowellrex:
Resolves #10049. Passes a specification for the current organization's queue table view to the front-end. Specification ends up in the `queue.queueConfig` element of the global redux store and looks like the following example for VLJ support staff:
```jsx
{
    "active_tab": "unassigned",
    "table_title": "VLJ Support Staff cases",
    "tabs": [
        {
            "allow_bulk_assign": false,
            "columns": [
                "hearingBadgeColumn",
                "detailsColumn",
                "taskColumn",
                "typeColumn",
                "docketNumberColumn",
                "daysWaitingColumn",
                "readerLinkColumn"
            ],
            "description": "Cases owned by the VLJ Support Staff team that are unassigned to a person.",
            "label": "Unassigned (%d)",
            "name": "unassigned",
            "task_group": "unassignedTasks"
        },
        {
            "allow_bulk_assign": false,
            "columns": [
                "hearingBadgeColumn",
                "detailsColumn",
                "taskColumn",
                "typeColumn",
                "assignedToColumn",
                "docketNumberColumn",
                "daysWaitingColumn"
            ],
            "description": "Cases assigned to a member of the VLJ Support Staff team.",
            "label": "Assigned (%d)",
            "name": "assigned",
            "task_group": "assignedTasks"
        },
        {
            "allow_bulk_assign": false,
            "columns": [
                "hearingBadgeColumn",
                "detailsColumn",
                "taskColumn",
                "typeColumn",
                "assignedToColumn",
                "docketNumberColumn",
                "daysWaitingColumn"
            ],
            "description": "Cases completed (last two weeks):",
            "label": "Completed",
            "name": "completed",
            "task_group": "completedTasks"
        }
    ]
}
```


.reek.yml
```diff
@@ -5,6 +5,7 @@ detectors:
   Attribute:
     exclude:
       - BulkTaskAssignment
+      - QueueConfig
```
lowellrex: Disabling because I'm following the standard pattern of how we set instance attributes for `ActiveModel` objects.

.reek.yml
```diff
@@ -20,6 +21,9 @@ detectors:
     enabled: false
   NilCheck:
     enabled: false
+  RepeatedConditional:
+    exclude:
+      - QueueConfig
```
lowellrex: Reek complains that `QueueConfig` calls `organization.show_regional_office_in_queue?` too many times and prefers we encapsulate that conditional in a class. `QueueConfig` is acting as a factory for columns and I prefer we leave it as is because the way it is currently written is more explicit.

app/models/queue_config.rb
```diff
@@ -0,0 +1,122 @@
...
+
+  def initialize(args)
+    super
+
+    if !organization&.is_a?(Organization)
+      fail(
+        Caseflow::Error::MissingRequiredProperty,
+        message: "organization property must be an instance of Organization"
+      )
+    end
```
lowellrex: I want `QueueConfig.new` to fail immediately if there is no organization argument (or it is not actually an `Organization`) and I could not wrangle `ActiveModel` validations to do that for me. Would love to move this back to a validation if it is possible.

app/models/queues/generic_queue.rb
```diff
@@ -9,6 +9,10 @@ def tasks
     (relevant_tasks + relevant_attorney_tasks).each(&:update_if_hold_expired!)
   end
 
+  def config
+    user.is_a?(Organization) ? QueueConfig.new(organization: user).to_h : {}
```
lowellrex: This might be able to be incorporate into the `QueueConfig` class in the future. For now I think it is fine to leave it here as it reduces the complexity of the `QueueConfig` class.

app/models/queue_config.rb
```diff
@@ -0,0 +1,122 @@
+# frozen_string_literal: true
+
+class QueueConfig
```
lowellrex: This is the meat of the change here.

.reek.yml
lomky: I could go either way on this one. I don't think it's more or less explicit either way (assuming a clearly named class). Pinging @kevmo for a third opinion.

kevmo: I prefer the methods between those two options.

kevmo:
Given team shortages this week, timing (<1 hour until sprint retro), failing CodeClimate, & the fact that the API layer doesn't match up to what the `this.queueConfig` function is generating on the front-end, I am going to leave this PR until Lowell is back next week, so we can discuss (am happy to pick it up and finish it on Monday).  I think this is very close to being done, though.

app/models/queue_config.rb
lowellrex: We can check the validity in a custom validator, but I'm not sure how to have the validator raise the exception immediately upon validation instead of doing the whole `obj.new; fail(SomeError) unless obj.valid?` dance.

I'd like to avoid the pattern of checking validity in the calling code, in favor of capturing that logic within the code that is called itself. I prefer to throw the exception as early as we know we we've encountered an error, preserve the original error as long as possible, and handle the exception when we are ready to (which in this case I don't think would be until we hit [the existing logic in `ApplicationController`](https://github.com/department-of-veterans-affairs/caseflow/blob/45725534fe7a631dab22143d9544668031ed946b/app/controllers/application_controller.rb#L12) to return the error to the front-end).

What I'd like to avoid:
```ruby
  # app/models/generic_queue.rb
  def config
    if user.is_a?(Organization)
      queue_config = QueueConfig.new(organization: user)
      if !queue_config.valid?
        fail(SomeError)
      end
      queue_config.to_h
    else
      {}
    end
  end
```

.reek.yml
lowellrex: Let's stick with what we've got then!

app/models/queue_config.rb
lomky: I think the `after_initialize` callback might be what you're after? 

https://guides.rubyonrails.org/v2.3.11/activerecord_validations_callbacks.html#after-initialize-and-after-find

lowellrex: I think that is only available for `ActiveRecord` classes and not `ActiveModel` classes.

PR #11008, opened 2019-06-04
Move task subclasses into their own files

### LINKED ISSUE #11007: Fix single-table inheritance complaints.
Move `MailTask` and `HearingAdminActionTask` subclasses into their own files. Related [slack discussion](https://dsva.slack.com/archives/C2ZAMLK88/p1558735071019000).

lowellrex:
Resolves #11007.


app/models/tasks/clear_and_unmistakeable_error_mail_task.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class ClearAndUnmistakeableErrorMailTask < MailTask
```
aroltsch: shall we create a mail_tasks directory under tasks and put all these guys there?

lowellrex: I have a slight preference towards all the tasks being in the same directory (`/tasks`) just for simplicity's sake, but if you feel strongly that we should put them in their own directory then we can do that too.

If we put these classes in their own directory, would we add a `MailTasks::` prefix to their names or would they stay as-is?

pkarman: if you change the class name, they would need a new dir.

if you keep the class names the same, they could just go in `/tasks`

I prefer the latter personally.

app/models/tasks/hearing_related_mail_task.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class HearingRelatedMailTask < MailTask
```
aroltsch: Can we move hearing tasks under `hearing_tasks` directory the same as `mail_tasks`?

app/models/tasks/clear_and_unmistakeable_error_mail_task.rb
aroltsch: I don't have a strong preference.

PR #11017, opened 2019-06-04
Return promise function to org queue loading component

lowellrex:
Completes the work started in #10912 in preparation for #10049 by moving the loading promise back to the `OrganizationQueueLoadingScreen` component.


client/app/queue/OrganizationQueueLoadingScreen.jsx
```diff
@@ -17,14 +18,34 @@ import {
 class OrganizationQueueLoadingScreen extends React.PureComponent {
   reload = () => window.location.reload();
 
+  createOrgQueueLoadPromise = () => ApiUtil.get(this.props.urlToLoad, { timeout: { response: 5 * 60 * 1000 } }).
```
kevmo: Why not just write `300000`?

lowellrex: To me `5 * 60 * 1000` reads more clearly as 5 minutes than `300000` (milliseconds) does.

kevmo: That's fair.

spec/models/task_spec.rb
```diff
@@ -695,7 +695,7 @@ def next_assignee(_options = {})
 
     it "returns and destroys related timers" do
       expect(TaskTimer.where(task_id: task_id).count).to eq(task_timer_count)
-      expect(task.task_timers).to eq(task_timers)
+      expect(task.task_timers.to_a).to match_array(task_timers)
```
kevmo: :+1:

PR #11018, opened 2019-06-04
Speed up Quality Review feature test

lowellrex:
Speeds up a test that occasionally causes Circle CI failures.

### DIFF
--- a/spec/feature/queue/quality_review_flow_spec.rb
+++ b/spec/feature/queue/quality_review_flow_spec.rb
@@ -109,7 +109,7 @@ RSpec.feature "Quality Review worflow" do
       find(".Select-control", text: "Select an action", match: :first).click
       find("div", class: "Select-option", text: Constants.TASK_ACTIONS.JUDGE_QR_RETURN_TO_ATTORNEY.to_h[:label]).click
 
-      expect(dropdown_selected_value(find(".cf-modal-body"))).to eq
+      expect(dropdown_selected_value(find(".cf-modal-body"))).to eq attorney_user.full_name
       click_on "Submit"
 
       expect(page).to have_content("Task assigned to #{attorney_user.full_name}")
@@ -192,7 +192,7 @@ RSpec.feature "Quality Review worflow" do
       # ensure no duplicate org tasks
       page.go_back
 
-      assert_text("QualityReviewTask", count: 1)
+      page.find("table#case-timeline-table").assert_text("QualityReviewTask", count: 1)
     end
   end
 end
--- a/spec/feature/queue/quality_review_flow_spec.rb
+++ b/spec/feature/queue/quality_review_flow_spec.rb
@@ -63,132 +63,136 @@ RSpec.feature "Quality Review worflow" do
 
   scenario "return case to judge" do
     expect(QualityReviewTask.count).to eq 1
-    # step "QR user visits the quality review organization page and assigns the task to themself"
-    visit quality_review_organization.path
-    click_on veteran_full_name
 
-    find(".Select-control", text: "Select an action").click
-    find("div", class: "Select-option", text: Constants.TASK_ACTIONS.ASSIGN_TO_PERSON.to_h[:label]).click
+    step "QR user visits the quality review organization page and assigns the task to themself" do
+      visit quality_review_organization.path
+      click_on veteran_full_name
 
-    fill_in "taskInstructions", with: "Review the quality"
-    click_on "Submit"
+      find(".Select-control", text: "Select an action").click
+      find("div", class: "Select-option", text: Constants.TASK_ACTIONS.ASSIGN_TO_PERSON.to_h[:label]).click
 
-    expect(page).to have_content("Task assigned to #{qr_user_name}")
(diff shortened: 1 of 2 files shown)


spec/feature/queue/quality_review_flow_spec.rb
```diff
@@ -63,132 +63,136 @@
...
+      # ensure no duplicate org tasks
+      page.go_back
 
-    expect(page).to have_content("#{veteran_full_name}'s case has been marked complete")
-    # ensure no duplicate org tasks
-    page.go_back
-    case_timeline_rows = page.find_all("table#case-timeline-table tbody tr")
-    quality_review_tasks_on_timeline = case_timeline_rows.select do |html_node|
-      html_node.has_content?("QualityReviewTask")
+      assert_text("QualityReviewTask", count: 1)
```
lowellrex: This is the only meaningful change.

spec/feature/queue/quality_review_flow_spec.rb
```diff
@@ -63,132 +63,136 @@
...
+      find("button", text: COPY::TASK_SNAPSHOT_VIEW_TASK_INSTRUCTIONS_LABEL, match: :first).click
+      expect(page).to have_content(qr_instructions)
 
-    find(".Select-control", text: "Select an action", match: :first).click
-    find("div", class: "Select-option", text: Constants.TASK_ACTIONS.MARK_COMPLETE.to_h[:label]).click
+      find(".Select-control", text: "Select an action", match: :first).click
+      find("div", class: "Select-option", text: Constants.TASK_ACTIONS.MARK_COMPLETE.to_h[:label]).click
 
-    expect(page).to have_content("Mark as complete")
+      expect(page).to have_content("Mark as complete")
```
lowellrex: This step is pretty slow, would be nice to speed it up.

spec/feature/queue/quality_review_flow_spec.rb
```diff
@@ -63,132 +63,136 @@
...
-    find("div", class: "Select-option", text: Constants.TASK_ACTIONS.QR_RETURN_TO_JUDGE.to_h[:label]).click
+      click_on veteran_full_name
 
-    expect(dropdown_selected_value(find(".cf-modal-body"))).to eq judge_user.full_name
-    fill_in "taskInstructions", with: qr_instructions
+      find(".Select-control", text: "Select an action").click
+      find("div", class: "Select-option", text: Constants.TASK_ACTIONS.QR_RETURN_TO_JUDGE.to_h[:label]).click
 
-    click_on "Submit"
+      expect(dropdown_selected_value(find(".cf-modal-body"))).to eq judge_user.full_name
```
lowellrex: This step is also pretty slow, would also be nice to speed it up.

spec/feature/queue/quality_review_flow_spec.rb
```diff
@@ -63,132 +63,136 @@
...
+      # ensure no duplicate org tasks
+      page.go_back
 
-    expect(page).to have_content("#{veteran_full_name}'s case has been marked complete")
-    # ensure no duplicate org tasks
-    page.go_back
-    case_timeline_rows = page.find_all("table#case-timeline-table tbody tr")
-    quality_review_tasks_on_timeline = case_timeline_rows.select do |html_node|
-      html_node.has_content?("QualityReviewTask")
+      page.find("table#case-timeline-table").assert_text("QualityReviewTask", count: 1)
```
lowellrex: This step is speedier than the previous loop we had.

spec/feature/queue/quality_review_flow_spec.rb
```diff
@@ -63,132 +63,136 @@
...
 
-    find("button", text: COPY::TASK_SNAPSHOT_VIEW_TASK_INSTRUCTIONS_LABEL, match: :first).click
-    expect(page).to have_content(qr_instructions)
+      find("button", text: COPY::TASK_SNAPSHOT_VIEW_TASK_INSTRUCTIONS_LABEL, match: :first).click
+      expect(page).to have_content(qr_instructions)
 
-    find(".Select-control", text: "Select an action", match: :first).click
-    find("div", class: "Select-option", text: Constants.TASK_ACTIONS.MARK_COMPLETE.to_h[:label]).click
+      find(".Select-control", text: "Select an action", match: :first).click
+      find("div", class: "Select-option", text: Constants.TASK_ACTIONS.MARK_COMPLETE.to_h[:label]).click
```
lowellrex: Turns out it is actually this clicking step that is taking such a long time. The clicking happens immediately, but we do not proceed to the assertion on line 169 for nearly 2 seconds. My current hypothesis is that because [`element.click()` returns `element`](https://www.rubydoc.info/github/teamcapybara/capybara/master/Capybara/Node/Element#click-instance_method) and our element disappears from view after we click it, we are befuddling the machinery and causing it to be slow to return for some reason. I'm not confident about this--or at all sure about how/why this would happen--but this is the only thing I can think of.

lowellrex:
> Looks great! Do you want to file any issues for the line's you ID'd as super slow?

I think it's fine to have the ticket or not, either way, your call!

PR #11019, opened 2019-06-05
Refactor nonpriority_receipts_per_year

### LINKED ISSUE #11002: ACD Refactor: Make DirectReview#nonpriority_receipts_per_year easier to understand
There is a lot of code to parse through and hold in your head when reading that method. What it's returning is not that complex, but the way it's written makes it hard to decipher.

monfresh:
**Why**: To make it easier to understand.

**How**: Use the Replace Temp with Query refactoring recipe to move
all the variables and logic into private methods so that the main
method is understandable at a glance. The well-named methods explain
what is returned, and if one needs to dig deeper, they can look at the
private methods, but it's no longer necessary to read all of that code
in order to understand what the method is doing.

References:
https://refactoring.guru/replace-temp-with-query

Resolves #11002


spec/models/dockets/direct_review_docket_spec.rb
```diff
@@ -54,14 +54,6 @@
   context "#nonpriority_receipts_per_year" do
     subject { DirectReviewDocket.new.nonpriority_receipts_per_year }
 
-    context "before April 1st, 2019" do
```
pkarman: is this test no longer needed because we have moved into a post 2019/4/1 era where we have enough data?

app/models/dockets/direct_review_docket.rb
```diff
@@ -27,26 +27,44 @@ def time_until_due_of_new_appeal
...
+  def prorated_nonpriority_appeals_per_year_based_on_days_since_march_first_2019
+    ((nonpriority_appeals_since_march_first_2019 * 365) / days_since_march_first_2019).round
+  end
+
+  def number_of_nonpriority_appeals_received_in_the_past_year
+    all_nonpriority.where("receipt_date > ?", 1.year.ago).ids.size
+  end
+
+  def today
+    @today ||= Time.zone.today
```
lomky: Any chance this object lives long enough for this to become incorrect?

spec/models/dockets/direct_review_docket_spec.rb
monfresh: Correct.

app/models/dockets/direct_review_docket.rb
monfresh: Not that I can think of.

Issue #10795, opened 2019-05-16
Update edited contentions in VBMS


leikkisa:
### Acceptance criteria
- Anytime a user updates a request issue's contention text, that update is processed in VBMS asynchronously

### Technical notes
- One approach would be to add "updated_request_issue_ids" to RequestIssuesUpdate

Sjones352:
Verified in Intake and VBMS that it updates asynchronously 
![Screen Shot 2019-06-27 at 10.21.01 PM.png](https://images.zenhubusercontent.com/5b8fddd1b2cc7b6c78ece9f9/13d740ea-b95b-429a-a080-e21db2ff266c)

![Screen Shot 2019-06-27 at 10.21.17 PM.png](https://images.zenhubusercontent.com/5b8fddd1b2cc7b6c78ece9f9/e1535e0f-217a-4e88-a915-ac138098be4a)

https://appeals.cf.uat.ds.va.gov/higher_level_reviews/9c528dc7-98a5-48cb-b2d1-bf2a03bb7fa4/edit/confirmation

![Screen Shot 2019-06-27 at 10.27.20 PM.png](https://images.zenhubusercontent.com/5b8fddd1b2cc7b6c78ece9f9/1cadc694-97ba-4339-8044-c6a4d5c2c2d2)

PR #11030, opened 2019-06-05
Update contention text in VBMS in request issues update

leikkisa:
connects #10795

### Description
Update contentions in VBMS when their text is updated in RequestIssuesUpdate
Change RequestIssuesUpdate to perform now instead of asynchronously

*Schema Changes Only*

* [x] Column comments updated


app/models/request_issues_update.rb
```diff
@@ -34,27 +34,20 @@ def perform!
       )
       cancel_active_tasks
       submit_for_processing!
-      process_job
+      DecisionReviewProcessJob.perform_now(self)
```
leikkisa: Attempt to run the RequestIssueUpdate immediately to prevent issues with the state in VBMS changing

pkarman: I don't think we want to do this yet. Let's keep it async till we have more information about why this might be happening. I am working on a PR that will help.

app/models/request_issue.rb
```diff
@@ -413,8 +413,23 @@ def withdraw!(withdrawal_date)
...
+    update!(edited_description: new_description, contention_updated_at: nil)
+  end
+
+  def contention
+    return unless contention_reference_id
+
+    end_product_establishment.contention_for_object(self)
+  end
+
+  def update_contention_text_in_vbms!
```
pkarman: it might merit a comment there that this is called async only by `RequestIssuesUpdate` -- in which case, maybe this method belongs in that class instead?

app/models/request_issues_update.rb
```diff
@@ -34,27 +34,19 @@ def perform!
...
     end
 
     true
   end
 
-  def process_job
```
pkarman: Let's keep this method.

app/models/request_issues_update.rb
```diff
@@ -34,27 +34,19 @@ def perform!
...
-  end
-
   # establish! is called async via DecisionReviewProcessJob.
   # it is queued via submit_for_processing! in the perform! method above.
   def establish!
     attempted!
 
     review.establish!
-
+    edited_issues.each(&:update_contention_text_in_vbms!)
```
pkarman: if we move `updated_contention_text_in_vbms!` to this class, this could become:

```ruby
edited_issues.each { |issue| update_contention_text_in_vbms!(issue) }
```

pkarman: Alternately, we could move `updated_contention_text_in_vbms!` to `EndProductEstablishment` where all the other VBMS API calls happen.

leikkisa: This ended up looking pretty funny:
```
edited_issues.each { |issue| issue.end_product_establishment.update_contention_text_in_vbms!(issue) }
```

I don't think it belongs in request_issues_update. My favorite place is in request_issue, but I am trying out using an extracted class (Moncef showed me when I got a Code Climate thing on another request issue ticket).

What do you think of doing it that way?

app/models/contention.rb
```diff
@@ -7,7 +7,7 @@ def initialize(description)
     self.description = description
   end
 
-  # BGS limits contention text to 255 characters
+  # BGS limits contention text to 255 bytes
```
pkarman: good fix!

app/models/request_issue_contention.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
```
pkarman: I really like this refactor! nice delegation pattern.

PR #11059, opened 2019-06-07
Refactor Timeout Calculation

### LINKED ISSUE #10997: Seeing Null cases in hearings admin team cases view.
When I log into caseflow hearings and switch to "hearings admin team cases" I see a "null".  
Capture

ferristseng:
~Resolves 10997~

### Description

  * Small refactoring moving seconds calculation to utility method
  * ~Increase timeout for `OrganizationQueueLoadingScreen` to 30 minutes. Average time I saw this route take in New Relic was 11 seconds~

### DIFF
--- a/client/app/queue/OrganizationQueueLoadingScreen.jsx
+++ b/client/app/queue/OrganizationQueueLoadingScreen.jsx
@@ -3,7 +3,7 @@ import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 
 import ApiUtil from '../util/ApiUtil';
-import { getDurationFromSeconds } from '../util/DateUtil';
+import { getMinutesToMilliseconds } from '../util/DateUtil';
 import LoadingDataDisplay from '../components/LoadingDataDisplay';
 import { LOGO_COLORS } from '../constants/AppConstants';
 import { extractAppealsAndAmaTasks } from './utils';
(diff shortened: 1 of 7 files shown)


client/app/util/DateUtil.js
```diff
@@ -152,3 +152,7 @@ export const getTimeInDifferentTimeZone = (date, timeZone) => {
 export const getDate = (date) => {
   return moment(date).format('YYYY-MM-DD');
 };
+
+export const getDurationFromSeconds = (seconds) => {
+  return seconds * 60 * 1000;
```
pkarman: the method name confuses me. What unit is "duration"? It seems like this is calculating the number of seconds in `seconds` minutes. I.e. the method argument should be `minutes`.

Is this what you mean?

```javascript
export const getMinutesToMilleseconds = (minutes) => {
  return minutes * 60 * 1000;
}
```

ferristseng: Yeah! Let me make that change

client/app/queue/OrganizationQueueLoadingScreen.jsx
```diff
@@ -18,25 +19,35 @@ import {
...
-          is_vso: isVso,
-          queue_config: queueConfig
-        } = JSON.parse(response.text);
+  createOrgQueueLoadPromise = () => {
+    // Setting timeout duration very high as a workaround for:
+    //
+    //    https://github.com/department-of-veterans-affairs/caseflow/issues/10997
+    //
+    const requestOptions = {
+      timeout: { response: getMinutesToMilliseconds(30) }
```
pkarman: 30 minutes is pretty long eh? I think the original 5 is probably correct and we need to be looking elsewhere for the culprit on this particular bug.

PR #11062, opened 2019-06-07
cancel icon

### LINKED ISSUE #10414: Canceled tasks appear as completed in Case Timeline
When an AMA task is canceled, the task appears as completed in the Case Timeline. This is because cancel task functionality was added after Case Timeline was implemented, and we have not yet finished designing how to display canceled in it. 

### Case Timeline "Cancel Task" (isolated view)
cancel-task-isolated-view

----- 
### Case Timeline "Cancel Task" (In Context)
case-timeline-cancel-task

youngfreezyVA:
Resolves #10414


youngfreezyVA:
@aroltsch this pr is for the mvp version, I think @sneha-pai is going to provide an svg instead of the icon that i'm using here.

Issue #11043, opened 2019-06-06
Consider modifiers on recently closed claims "taken"


pkarman:
Some VBMS modifiers look like they ought to be available (are on canceled EPs) but still throw errors when attempting to establish new EPs.

Update the EndProductModifierFinder workflow to consider a modifier taken if its end product was canceled or cleared within the last day or two.

## AC

* If a veteran's has an end product that was canceled or cleared recently, do not attempt to use it for establishing a new claim of the same type for that veteran. For example, if a veteran has a claim with the modifier 040 that was cleared today, and another with modifier 041 that is pending, try establishing the claim with modifier 042.

leikkisa:
Response from BGS on why this is happening:
This happens when the BDN PIF is still present although the corporate claim has been cleared or cancelled (PCLR or PCAN).  This will ALWAYS happen within a timing window but may happen due to other situations/scenarios.

The timing window is best illustrated via a scenario:

The starting point is when a claim is created on both the BDN and on corporate with a CEST.
A PCAN or PCLR is done off some corporate app (Share, VBMS, etc). Now the claim is no longer active and is basically invisible on corporate. However, the BDN PIF has been marked inactive BUT apps still treat it as active just due to its existence.

The state/status of the PIF on the BDN is not considered.

A second (supplemental) CEST is done with the same EPC etc. This works on corporate and should work on the BDN except for the timing window. Technically, a new claim should be able to be created with the same attributes as one recently cleared or cancelled. That workson corporate. However, on the BDN, the claim still exists until WIPP runs and all apps consider it still “active”. Thus, the error occurs thatan attempt is being made to create a duplicate claim – because on corporate it is no longer there, but it still is on the BDN until WIPP runsand removes it.

That’s the story and has been for many, many years. If a user does a PCAN or PCLR and then tries to re-CEST the same claim before WIPP runs on theBDN, this error will occur. Usually, this means that the user must PCAN or PCLR and attempt re-CEST on the same day, because I believe WIPP runseach night.

leikkisa:
It sounds like this is due to manually canceling or clearing EPs "off corporate", so we can expect CLR and CAN eps to have their modifiers be available again if they are properly closed.  CAN seems more common, but the example I see in prod right now looks like it's from a CLR claim. We may want to respond to that specific error with incrementing the modifier instead of filtering out CLR and CAN from the beginning.

Examples of claim IDs with tied up modifiers:
72466355 (CLR)
[...]

PR #11065, opened 2019-06-07
only re-use modifiers for CLR EPs

pkarman:
connects #11043 

There is some confusion right now about whether BGS should be allowing the re-use of canceled EP modifiers. We know they have in the past, but currently they are not being recycled.

This PR limits the pool of valid modifiers to just cleared EPs.

### DIFF
--- a/app/models/end_product_establishment.rb
+++ b/app/models/end_product_establishment.rb
@@ -451,7 +451,7 @@ class EndProductEstablishment < ApplicationRecord
   end
 
   def taken_modifiers
-    @taken_modifiers ||= veteran.end_products.select(&:active?).map(&:modifier)
+    @taken_modifiers ||= veteran.end_products.select(&:cleared?).map(&:modifier)
   end
 
   def find_open_modifier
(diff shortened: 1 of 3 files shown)


app/models/end_product_establishment.rb
```diff
@@ -451,7 +451,7 @@ def fetch_result
   end
 
   def taken_modifiers
-    @taken_modifiers ||= veteran.end_products.select(&:active?).map(&:modifier)
+    @taken_modifiers ||= veteran.end_products.select(&:canceled?).map(&:modifier)
```
leikkisa: You still want to include the active EPs here as well

spec/models/end_product_establishment_spec.rb
```diff
@@ -130,11 +130,11 @@
       end
     end
 
-    context "when eps with a valid modifiers already exist" do
+    context "when eps with a valid modifier already exist" do
```
leikkisa: I would still include an example of an active EP in tests.

app/models/end_product_establishment.rb
pkarman: doh! you are right. I had the logic inverted. I have changed to be `reject cleared?` since we want all except those that are CLR

PR #11066, opened 2019-06-07
Legacy hearings load appellant address and poa info from BGS instead of VACOLS

### LINKED ISSUE #10473: Correct POA/Address source for Legacy Hearings
In Hearing Schedule, Legacy cases should load Power of Attorney and address information from BGS, not VACOLS. This will match AMA functionality and make our source of information consistent. 

- [x] Verify Legacy cases use BGS as the source on information for POA and Appellant/Veteran Address
- [x] Verify the Schedule Veteran modal uses and displays the BGS address 
- [ ] Verify any load messages match AMA functionality

tomas-nava:
Resolves #10473

### Description
For Legacy Hearings, prefer to load information about appellant address and poa/representative from BGS instead of VACOLS; particularly for the Daily Docket page and the Schedule Hearing modal.


app/models/legacy_appeal.rb
```diff
@@ -52,11 +50,14 @@ class UnknownLocationError < StandardError; end
...
   vacols_attr_accessor :number_of_issues
-
   vacols_attr_accessor :outcoding_date
   vacols_attr_accessor :last_location_change_date
   vacols_attr_accessor :docket_number, :docket_date, :citation_number
 
+  # these values have been delegated to BGS, but we may want to set them in VACOLS
+  vacols_attr_setter :appellant_address_line_1, :appellant_address_line_2
+  vacols_attr_setter :appellant_city, :appellant_state, :appellant_country, :appellant_zip
+
```
sharonwarner: Are these used anywhere? It feels weird to me to be able to write to a field that we don't read from. If we don't use them anywhere, I think I'd prefer to leave them out.

app/models/legacy_hearing.rb
```diff
@@ -5,15 +5,22 @@ class LegacyHearing < ApplicationRecord
...
-  vacols_attr_accessor :docket_number, :appeal_type, :appellant_address_line_1
-  vacols_attr_accessor :appellant_address_line_2, :appellant_city, :appellant_state
-  vacols_attr_accessor :appellant_zip, :appellant_country, :room, :bva_poc, :judge_id
+  vacols_attr_accessor :representative_name, :hearing_day_id
+  vacols_attr_accessor :docket_number, :appeal_type, :room, :bva_poc, :judge_id
+
+  # these values have been delegated to BGS, but we may want to set them in VACOLS
+  vacols_attr_setter :representative
+  vacols_attr_setter :appellant_address_line_1, :appellant_address_line_2
+  vacols_attr_setter :appellant_city, :appellant_state, :appellant_country, :appellant_zip
```
sharonwarner: Same here - I'd rather remove this if we're not currently using it.

app/controllers/hearings_controller.rb
```diff
@@ -6,6 +6,10 @@ class HearingsController < HearingsApplicationController
   before_action :verify_access_to_hearings, except: [:show_print, :show]
   before_action :verify_access_to_reader_or_hearings, only: [:show_print, :show]
 
+  def set_application
```
sharonwarner: We don't need this method because the controller inherits from hearings_application_controller.

PR #11067, opened 2019-06-07
Remove VLJ and attorney names from case title for removed appeals

### LINKED ISSUE #10547: Remove VLJ and attorney names from case title for removed appeals
When appeals have been removed, also remove attorney and judge names from the case title, to prevent confusion about this appeal.

Here is the case title:
image

Sjones352:
connects #10547
Screen Shot 2019-06-07 at 6 44 23 PM
Screen Shot 2019-06-07 at 6 41 30 PM

### DIFF
--- a/spec/feature/queue/case_details_spec.rb
+++ b/spec/feature/queue/case_details_spec.rb
@@ -391,7 +391,6 @@ RSpec.feature "Case details" do
     scenario "case details page shows appropriate text" do
       visit "/queue"
       click_on "#{appeal.veteran_full_name} (#{appeal.veteran_file_number})"
-
       # Call have_content() so we wait for the case details page to load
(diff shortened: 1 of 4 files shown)


spec/feature/queue/case_details_spec.rb
```diff
@@ -932,6 +939,7 @@
     context "when an AMA appeal has been dispatched from the Board" do
       let(:appeal) { FactoryBot.create(:appeal) }
       let(:root_task) { FactoryBot.create(:root_task, appeal: appeal) }
+      let!(:removed_request_issue) { create(:request_issue, decision_review: appeal) }
```
leikkisa: This looks like it wasn't meant to be a part of this.

PR #11068, opened 2019-06-10
AsyncableJobsReporter as_csv

pkarman:
Generates a CSV report of asyncable jobs, with PII filtered out.

### DIFF
--- a/.reek.yml
+++ b/.reek.yml
@@ -10,6 +10,7 @@ detectors:
   FeatureEnvy:
     exclude:
       - FetchDocumentsForReaderJob#fetch_for_appeal
+      - AsyncableJobsReporter
   InstanceVariableAssumption:
     exclude:
       - Appeal
@@ -18,6 +19,9 @@ detectors:
     enabled: false
   MissingSafeMethod:
     enabled: false
+  NestedIterators:
+    exclude:
+      - AsyncableJobsReporter
   NilCheck:
     enabled: false
   TooManyMethods:
--- a/app/services/asyncable_jobs_reporter.rb
+++ b/app/services/asyncable_jobs_reporter.rb
@@ -7,6 +7,10 @@ class AsyncableJobsReporter
     self.jobs = jobs
   end
 
+  def as_csv
+
+  end
+
   # summarize all jobs by class and error and age
   def summary
     @summary ||= build_summary
--- a/app/services/asyncable_jobs_reporter.rb
+++ b/app/services/asyncable_jobs_reporter.rb
@@ -1,14 +1,30 @@
 # frozen_string_literal: true
 
+require "csv"
+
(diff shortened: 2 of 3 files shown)


.reek.yml
```diff
@@ -10,6 +10,7 @@ detectors:
   FeatureEnvy:
     exclude:
       - FetchDocumentsForReaderJob#fetch_for_appeal
+      - AsyncableJobsReporter
```
pkarman: disabling this check because it will always show positives for this class. Its very design is intended to operate on another object rather than itself.

.reek.yml
```diff
@@ -18,6 +19,9 @@ detectors:
     enabled: false
   MissingSafeMethod:
     enabled: false
+  NestedIterators:
+    exclude:
+      - AsyncableJobsReporter
```
pkarman: there are probably ways to flatten out the iteration but I'm not sure they would make the code more readable. Again, because we are creating a multi-tested tree of data, nested iterations are required.

app/services/asyncable_jobs_reporter.rb
```diff
@@ -30,13 +50,17 @@ def summarize
 
   private
 
+  def sanitize_error(job)
+    # keep PII out of output
+    (job[job.class.error_column] || "none").gsub(/\s.+/s, "")
```
leikkisa: Good catch, I completely didn't think of this!

spec/services/asyncable_jobs_reporter_spec.rb
```diff
@@ -78,4 +83,12 @@
       expect(subject.summarize).to eq(msg.chomp)
     end
   end
+
+  describe "#as_csv" do
+    it "returns a CSV-formatted string" do
+      csv = subject.as_csv
+      expect(csv).to match /SupplementalClaim,\d+,#{6.days.ago},#{6.days.ago},#{7.days.ago},bad,#{veteran.participant_id}/
```
leikkisa: It might be nice to refer to the dates these are referring to (e.g. sc.establishment_submitted_at), but am okay with this too.

app/services/asyncable_jobs_reporter.rb
```diff
@@ -1,10 +1,30 @@
...
+      jobs.each do |job|
+        klass = job.class
+        csv << [
+          klass.to_s,
+          job.id,
+          job[klass.submitted_at_column],
+          job[klass.last_submitted_at_column],
+          job[klass.attempted_at_column],
+          sanitize_error(job),
+          job.try(:veteran).try(:participant_id)
```
leikkisa: The errors are the most meaningful, but I thought it was useful to add whether they were expired too, especially for RequestIssue since it's understood that there's some waiting period before we expect a decision.

PR #11069, opened 2019-06-10
Segregate UI tests from non-UI tests

edmangimelli:
### Description

UI tests are now tagged with `ui_test: true`

example; ignore all UI tests:
```
bundle exec rspec --tag ~ui_test
```

example; only run UI tests:
```
bundle exec rspec --tag ui_test
```

Also, `lib/tasks/spec.rake` updated to give the option to ignore tests

examples:
```
bin/rake spec:unit # run all unit tests
bin/rake spec:unit_no_ui # run all unit tests except those which test the UI
bin/rake spec:queue_no_ui
```

### Acceptance Criteria

- calling rspec with or without `ui_test` tag runs correctly
- new rake tasks (*_no_ui) run correctly

### DIFF
--- a/lib/tasks/spec.rake
+++ b/lib/tasks/spec.rake
@@ -3,11 +3,13 @@
 require "rspec/core/rake_task"
 
 namespace :spec do
+  DEFAULT_RSPEC_OPTS = "--tty --color"
+
   feature_spec_folders = Dir.entries("./spec/feature").select do |file|
     ![".", ".."].include?(file) && File.directory?(File.join("./spec/feature", file))
   end
 
-  spec_names = feature_spec_folders + %w[other unit]
+  spec_names = [*feature_spec_folders, 'unit', 'other'].flat_map { |n| [n, "#{n}_no_ui"] }
 
   # Tasks for setting up the environment and running the tests without
   # immediately printing output to the screen. This prevents output from overrunning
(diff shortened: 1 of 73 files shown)


lib/tasks/spec.rake
```diff
@@ -3,11 +3,13 @@
...
 
 namespace :spec do
+  DEFAULT_RSPEC_OPTS = "--tty --color"
+
   feature_spec_folders = Dir.entries("./spec/feature").select do |file|
     ![".", ".."].include?(file) && File.directory?(File.join("./spec/feature", file))
   end
 
-  spec_names = feature_spec_folders + %w[other unit]
+  spec_names = [*feature_spec_folders, 'unit', 'other'].flat_map { |n| [n, "#{n}_no_ui"] }
```
tomas-nava:
```suggestion
  spec_names = [*feature_spec_folders, "unit", "other"].flat_map { |n| [n, "#{n}_no_ui"] }
```

Double quoting strings is house style.

lib/tasks/spec.rake
```diff
@@ -53,24 +55,42 @@ namespace :spec do
...
-  end
 
-  desc "Run all uncategorized feature specs"
-  RSpec::Core::RakeTask.new(:other) do |t|
-    t.pattern = "spec/feature/*_spec.rb"
-    t.rspec_opts = "--tty --color"
+  make_rake_task = proc do |name:, desc:, pattern:, exclude_pattern:|
+    [
+      { name_suffix: nil, extra_rspec_opts: nil },
+      { name_suffix: '_no_ui', extra_rspec_opts: "--tag ~ui_test" }
```
tomas-nava:
```suggestion
      { name_suffix: "_no_ui", extra_rspec_opts: "--tag ~ui_test" }
```
Double quoting strings is house style.

edmangimelli:
> Couldn't we get a similar result by
> 
> 1. using [the `feature` type](https://relishapp.com/rspec/rspec-rails/docs/feature-specs/feature-spec) to mark only our feature tests, then
> 2. [filtering with `--tag type:feature`](https://relishapp.com/rspec/rspec-core/v/2-4/docs/command-line/tag-option#filter-examples-with-a-name:value-tag) or `--tag ~type:feature`
> 
> instead of introducing a custom `ui_test` tag?

Although the vast majority of UI tests are within the `feature` directory, at least a couple aren't--which is why I ended up doing it this way.

lib/tasks/spec.rake
edmangimelli: gotcha.
I, by default, opt for single quoted strings if I don't need interpolation, but I'd rather be consistent!

tomas-nava: Yeah, that was my habit as well, it took me a while to get used to doing it this way.

tomas-nava:
> Although the vast majority of UI tests are within the `feature` directory, at least a couple aren't--which is why I ended up doing it this way.

What if we changed that? 😀 We could move `hearing_admin_action_verify_address_task_spec.rb` into the features directory, and move the feature tests in `hearing_admin_action_foreign_veteran_case_task_spec.rb` into a new file in the features directory, leaving only the 4 model tests.

pkarman:
+1 to @tomas-nava comment. Let's align all our feature tests in the consistent dir structure and just use the built-in differentiator.

edmangimelli:
Follow @tomas-nava 's suggestion, closing this PR and using a simple approach. This branch _could_ be useful in the future --the changes to the spec rake task are set up well for running tests according to a tag
New PR:
https://github.com/department-of-veterans-affairs/caseflow/pull/11085

PR #11072, opened 2019-06-10
Veteran Search by SSN or File Number

### LINKED ISSUE #10930: Allow users to search for claim number OR SSN
Right now, we only allow users to search for Veterans in Veteran Case Search using the Veteran's SSN. Some Veterans have SSNs, some have claim numbers, and some have both. We should allow users to use any of these numbers to search for Veterans, to have better feature parity with VACOLS.

## AC
- Users can search for Veterans by claims file number, if that claims file number is different than their SSN

ferristseng:
Resolves #10930

### Description

  * Changes `fetch_appeals_by_file_number` and `find_all_visible_by_file_number` to accept multiple file numbers.
  * Adds a `VeteranFinderQuery` class that can search for veterans, and their appeals
    * This replaces `AppealsForFileNumber`
  * Updates some messaging around the search box to include "SSN"
  * Refactors some functionality in `CaseSearchResultsForCaseflowVeteranId` and `CaseSearchResultsForVeteranFileNumber` to a base class
  * `CaseSearchResultsForVeteranFileNumber` gets a list of matching veterans by SSN and file number, then does a search for all the associated appeals and claims.
  * Updates tests, and adds new test for case where multiple veterans are found

### Acceptance Criteria

- [x] Users can search for Veterans by claims file number, if that claims file number is different than their SSN


app/queries/veteran_finder_query.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  class << self
+    def find_by_ssn_or_file_number(file_number_or_ssn:)
+      # Multiple possibilities:
+      #
+      #   1. The input is an SSN or File Number that maps directly to the `file_number`
+      #      field for a vet.
+      #   2. The input is an SSN that requires a BGS lookup for the File Number.
+      #   3. Combination of both.
+      veteran_file_number_or_ssn = Veteran.find_by(file_number: file_number_or_ssn)
```
pkarman: why re-implement the logic of `Veteran.find_by_file_number_or_ssn` method here?

app/queries/veteran_finder_query.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+class VeteranFinderQuery
```
pkarman: I think this class is doing too much. The name suggests it would find `Veteran` objects, but it appears to also allow for finding appeals. Maybe split it into 2 smaller classes?

app/queries/veteran_finder_query.rb
ferristseng: I wasn't sure if this was a case that needed to be handled, but it's to handle the situation when there are multiple veteran records, one where the `file_number` is set to the SSN, and one where it is not. It sounded like this could happen, but I'm not sure how frequent it is. I tried using `Veteran.find_by_file_number_or_ssn`, but ran into issues because it short circuits if there's a Veteran with their `file_number` set to their SSN.

ferristseng:
> I imagine the `VeteranFinder` with a simpler interface:
> 
> ...
> 
> and a separate class for finding appeals
> 
> ...
> 
> which would use the `VeteranFinder` internally.

I also felt like the `VeteranFinder` class might be doing too much, and wasn't sure if the appeals related code should bundled together. I like moving that code to a separate class.

> Also I'm not sure if the `VeteranFinder` is properly a [Query class](https://medium.flatstack.com/query-object-in-ruby-on-rails-56ea434365f0) since it is mostly wrapping the Veteran model. Alternately, we could do a larger refactor to move all the `find_*` methods from the Veteran model as they exist today into a `VeteranQuery` class, and then let the `VeteranFinder` class use `VeteranQuery` internally.
> 

I like the idea of moving the `find_*` methods to a different class because it does seem like the `Veteran` class has a lot of responsibilities. I'm not sure about whether or not to create a separate class for it. It seems like the `VeteranFinder` and `VeteranQuery` classes would be similar enough, so maybe all that functionality could live in the `VeteranFinder` class too? Either way, I think that refactoring is probably out-of-scope of this PR, and could probably be done in a follow up PR.

app/services/veteran_finder.rb
```diff
@@ -0,0 +1,22 @@
...
+        #   1. The input is an SSN or File Number that maps directly to the `file_number`
+        #      field for a vet.
+        #   2. The input is an SSN that requires a BGS lookup for the File Number.
+        #   3. Combination of both.
+        veteran_file_number_or_ssn = Veteran.find_by(file_number: one_file_number_or_ssn)
+        veteran_bgs_lookup = if one_file_number_or_ssn.length == 9
+                               Veteran.find_by_ssn(one_file_number_or_ssn)
+                             end
+
+        [veteran_file_number_or_ssn, veteran_bgs_lookup].select { |vet| !vet.nil? }.uniq(&:id)
```
pkarman: I think this could simply be:

```ruby
[veteran_file_number_or_ssn, veteran_bgs_lookup].compact.uniq
```

app/services/veteran_finder.rb
```diff
@@ -0,0 +1,38 @@
...
+
+        # If a veteran exists for the given claim number, there might be
+        # another record where the file number is the veteran's SSN.
+        veteran_file_number_match = Veteran.find_by(file_number: file_number)
+
+        veteran_ssn_match = if veteran_file_number_match
+                              Veteran.find_by(file_number: veteran_file_number_match.ssn)
+                            end
+      end
+
```
pkarman: there's also a case where there are multiple file numbers for a single Veteran, none of which are the same as the SSN. So we'd probably want to do a search by `participant_id` since that should be unique per Veteran.

```ruby
veterans = Veteran.where(participant_id: [veteran_file_number_match, veteran_ssn_match].compact.uniq.map(&:participant_id))
```

app/services/appeal_finder.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  class << self
+    def find_appeals_with_file_numbers(file_numbers)
+      return [] if file_numbers.empty?
+
+      MetricsService.record("VACOLS: Get appeal information for file_numbers #{file_numbers}",
+                            service: :queue,
+                            name: "VeteranFinderQuery.find_appeals_with_file_numbers") do
+        appeals = Appeal.where(veteran_file_number: file_numbers).reject(&:removed?).to_a
+        # rubocop:disable Lint/HandleExceptions
```
pkarman: I'm ok with this rubocop exception, given that so much other code depends on that behavior from `LegacyAppeal`.

pkarman: I'm ok with this rubocop exception, given that so much other code depends on that behavior from `LegacyAppeal`.

app/services/veteran_finder.rb
```diff
@@ -0,0 +1,38 @@
...
+        veteran_file_number_match = Veteran.find_by(file_number: ssn)
+        veteran_ssn_match = Veteran.find_by_ssn(ssn)
+      else
+        # The input is not an SSN. The value should be a claim number.
+        file_number = file_number_or_ssn
+
+        # If a veteran exists for the given claim number, there might be
+        # another record where the file number is the veteran's SSN.
+        veteran_file_number_match = Veteran.find_by(file_number: file_number)
+        veteran_ssn_match = if veteran_file_number_match
```
pkarman: this construct feels a little awkward to me. is this the same thing? maybe easier to read?

```ruby
veteran_ssn_match = Veteran.find_by(file_number: veteran_file_number_match.ssn) if veteran_file_number_match
```

app/services/appeal_finder.rb
ferristseng: @pkarman Do you mean that we should remove the begin/rescue here?

pkarman: No I am ok with the begin/rescue and the accompanying rubocop:disable. I was just explicitly saying I was ok because we are trying to obey rubocop as a matter of policy.

pkarman: I think you should restore the rubocop comments to disable the check.

Issue #9948, opened 2019-03-13
Fix race condition for on_decision_issues_sync_processed


leikkisa:
#### Story
If two decision issues are processed at the same time, they may not kick off the on_decision_issues_sync_processed job that creates remand supplemental claims.

#### Acceptance criteria
If an end product establishment has multiple decision issues, and any of them have a disposition of remanded, creating the remand supplemental claim should be kicked off.  This should not be prevented if the decision issues are synced at the same time.

#### Technical notes
This is happening in the End Product Establishment model at decision_issues_sync_complete?

One approach could be creating the remand supplemental claim once there is a remanded decision issue, and adding to it for subsequent decision issues, instead of waiting for all decision issues to be finished processing.

leikkisa:
I manually kicked off the one in Prod that wasn't started, the result is Supplemental Claim 1234

leikkisa:
Script for detecting these:
```
dis=DecisionIssue.remanded.select {|di| di.request_issues.first.decision_sync_processed_at != nil && di.decision_review.remand_supplemental_claims.count == 0}
```

When I checked today there were 197.

Script for fixing them:
```
dis.each{|di| di.decision_review.create_remand_supplemental_claims!}
```

msj0nes:
@leikkisa is this ticket ready for closure?

GabbyTISTA:
Moved to https://vajira.max.gov/browse/CASEFLOW-1691

PR #11073, opened 2019-06-10
[WIP] Create remand supplemental claims when a remand decision issue is created

leikkisa:
connects #9948 

Previously we were waiting for all decision issues on a claim to be processed, and then kicking off the job of creating remand supplemental claims.  This was failing when two decision issues finished processing at the same time.

This creates the remand_supplemental_claims when any one of the decision_issues is processed, and adding future remand issues to the same remand_supplemental_claim.


leikkisa:
Just a note that I'm planning on closing this, but I'm leaving it up for now to see the changes while working on an alternate PR.

Issue #10866, opened 2019-05-23
Prevent intake using reserved file number


pkarman:
File number 123456789 is so convenient to type for examples and test that it has a real Veteran record in production for BGS. The problem is that the record is so large that it will cause the initial intake response to time out fetching the record from upstream. If a user uses that number to test with, they will get stuck waiting for 24 hours for the intake-in-progress to expire.

As a work around, prevent the user of that particular file number during intake Veteran search.

## AC

* [ ] Put a FeatureToggle on this called `intake_reserved_file_number` so that we can continue to use that file number in other environments.
* [ ] On intake veteran search, if the file number is 123456789 and feature toggle is true, then present an error message for "invalid file number".

PR #11089, opened 2019-06-14
Prevent intake using reserved file number

Sjones352:
Connects #10866
Screen Shot 2019-06-14 at 12 37 05 AM

### DIFF
--- a/app/models/intake.rb
+++ b/app/models/intake.rb
@@ -288,13 +288,13 @@ class Intake < ApplicationRecord
     veteran_file_number =~ /^[0-9]{8,9}$/
   end
 
-  def file_number_reserved
+  def file_number_reserved?
     FeatureToggle.enabled?(:intake_reserved_file_number,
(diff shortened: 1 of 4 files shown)


client/app/intake/pages/search.jsx
```diff
@@ -69,6 +69,10 @@ class Search extends React.PureComponent {
         title: 'Veteran not found',
         body: veteranNotFoundInstructions
       },
+      reserved_veteran_file_number: {
```
Sjones352: I decided to add a `reserved_veteran_file_number` error code in order to show the title `Invalid file number`.

app/models/intake.rb
```diff
@@ -284,6 +288,18 @@ def file_number_valid?
     veteran_file_number =~ /^[0-9]{8,9}$/
   end
 
+  def file_number_reserved
```
pkarman: nit: `file_number_reserved?` with a question mark

Issue #10203, opened 2019-03-28
Check for Incident Flash on Intake start


Sjones352:
Incident Flash is a temporary lock on a Veteran record that prevents new claims from being established.

We should check for the presence of that Flash when an intake starts, similar to our checks for other valid Veteran attributes.

## AC

If Veteran has an Incident Flash on their record, halt the intake and present user with error message about contacting {{TODO identify contact}} to remove the incident flash lock.

leikkisa:
We are waiting for an update from Nate about this.

leikkisa:
Notes from an email:
From Colin: We have a case in production with the 2x CADD flash, which prevents address updates. However, claim establishment is allowed. I tested this in prodtest against the record in VBMS and it established properly.

From Peter: Error is coming from VBMS API

From Joe: This is an existing issue in general for VBMS CEST, not specifically the External Claim Service (claims api). There appear to be a variety of downstream address issues between VBMS CEST and FAS which will still trigger this error message.

Existing defect:
BGS GUIDs when attempting to CEST in VBMS.  Failed with Exception!! ( INC1948403 ) (714927)
https://rtc01.aide.oit.va.gov/jazz/resource/itemName/com.ibm.team.workitem.WorkItem/714927

leikkisa:
Script for finding these:
```drs=HigherLevelReview.where.not(establishment_error: nil).select{|dr| dr.establishment_error.include?("Incident")} + SupplementalClaim.where.not(establishment_error: nil).select{|dr| dr.establishment_error.include?("Incident")}.map(&:veteran_file_number)```

leikkisa:
I sent an email for the following Decision Reviews to request to lift the incident flash restriction temporarily.
VAVBAWAS/CO/IncidenteBenefitsNotification
IncidenteBenefitnotification.VBAVACO@va.gov

SCs - [28454, 32159, 18427, 14179, 19915, 31693, 25425, 8660, 5378, 25377, 4465, 31866, 31874, 22006, 17335, 24865]

HLR - 4976

leikkisa:
The actual email to send these to is IncidentTeam.VBASLC@va.gov (VAVBASLC/RO/VBAIncidentTeam), so that an encrypted email with the veteran SSNs can be sent.

The instructions should include the user contacting the veteran to confirm that it's a valid request, because these veteran's have been previous victims of fraud.

[...]

PR #11096, opened 2019-06-15
Add incident flash error

leikkisa:
connects #10203


leikkisa:
Let me know if you have any ideas on a better message.

Screen Shot 2019-06-14 at 5 10 20 PM

pkarman:
I think the rspec failures are from the refactor of the validation code. That might benefit from another refactor since those nested conditionals are getting harder to read.

PR #11097, opened 2019-06-15
updated README instructions on running caseflow locally

edmangimelli:
added note to README instructing devs to use `docker-compose up -d` if docker containers are stopped for some reason
(a useful note for devs either expecting `foreman start` to do docker-compose up, or devs simply unfamiliar with docker)

### DIFF
--- a/README.md
+++ b/README.md
@@ -361,12 +361,26 @@ To run caseflow:
 ```
 foreman start
 ```
-**Note:**  If you had just run `bundle exec rake local:build` to set up your local environment, docker-compose has already been called, and the docker containers you need are running in the background. If these containers are stopped or killed—restarting your computer, running `docker-compose down`, etc.—you must start them back up.
 
-For example, after restarting your computer, you must instead run:
+**Note:** The Docker containers must always be running in order for the Rails application to start successfully. Some rake tasks will start them automatically, but if you restart your computer or otherwise stop Docker, you'll want to run `docker-compose up -d` to start the containers initially.
+
 ```
 docker-compose up -d
 foreman start
+```
+
+Spinning up the docker containers is fast, but not instantaneous Occasionally, running `foreman start` *immediately* after `docker-compose up -d`, can cause the rails server to fail to start because not all of the containers are ready. For example, the following _may_ not work:
+
+```
+docker-compose up -d && foreman start
+```
+
+`Makefile.example` provides a bunch of useful shortcuts, one of which is the `run` directive. `run` will ensure that all of the dockers containers are ready before running `foreman start`.
+
+Example:
+
+```
+make -f Makefile.example run
 ```
 
 #### Separate Front & Backend Servers ####################################################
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -357,12 +357,24 @@ Your development setup of caseflow runs Redis, Postgres and OracleDB (VACOLS) in
 
 ### Running Caseflow ###################################################
 
-#### All in one ########################################################################
+To run caseflow:
 ```
 foreman start
 ```
+**Note:**  If you had just run `bundle exec rake local:build` to set up your local environment, docker-compose has already been called, and the docker containers you need are running in the background. If these containers are stopped or killed—restarting your computer, running `docker-compose down`, etc.—you must start them back up.
```
pkarman: this note is a little awkwardly phrased. how about something like:

> **Note**: The Docker containers must always be running in order for the Rails application to start successfully. Some `rake` tasks will start them automatically, but if you restart your computer or otherwise stop Docker, you'll want to run `docker-compose up -d` to start the containers initially.

edmangimelli: Yeah! That sounds much better.
What do you think about mentioning `Makefile.example`, or, at the very least, mentioning how you technically need to let those containers spin up and hit ready status before doing `foreman start` (like how the makefile has a while loop that waits for the ready status).

pkarman: I think a mention of `cp Makefile.example Makefile` is fine and the relevant commands. We've sort of refrained from endorsing it and re-writing the README to make use of it (since that would require edits in several places) but happy to entertain a PR for that. The Makefile.example is a late addition to the repo but does seem to be useful for new devs in particular.

README.md
```diff
@@ -357,12 +357,38 @@ Your development setup of caseflow runs Redis, Postgres and OracleDB (VACOLS) in
...
+```
+docker-compose up -d && foreman start
+```
+
+`Makefile.example` provides a bunch of useful shortcuts, one of which is the `run` directive. `run` will ensure that all of the dockers containers are ready before running `foreman start`.
+
+Example:
+
+```
+make -f Makefile.example run
```
pkarman: I just created #11266 that hopefully makes this a little clearer. That way you can just refer to `make run`.

README.md
```diff
@@ -357,12 +357,38 @@ Your development setup of caseflow runs Redis, Postgres and OracleDB (VACOLS) in
...
 ```
+
+**Note:** The Docker containers must always be running in order for the Rails application to start successfully. Some rake tasks will start them automatically, but if you restart your computer or otherwise stop Docker, you'll want to run `docker-compose up -d` to start the containers initially.
+
+```
+docker-compose up -d
+foreman start
+```
+
+Spinning up the docker containers is fast, but not instantaneous Occasionally, running `foreman start` *immediately* after `docker-compose up -d`, can cause the rails server to fail to start because not all of the containers are ready. For example, the following _may_ not work:
```
pkarman: missing period after "instantaneous"

PR #11098, opened 2019-06-17
Canceled hearings - require Evidence Submission Window Tasks

### LINKED ISSUE #10978: All Canceled Hearings Should Get an EvidenceSubmissionWindowTask
Currently, `EvidenceSubmissionWindowTask`s are only created if a `ScheduleHearingTask`is canceled (which means the hearing was withdrawn) or if the hearing is scheduled and held. All hearings that are cancelled should be given `EvidenceSubmissionWindowTask`

sharonwarner:
Resolves #10978 

### Description
This PR updates the canceled hearing logic so that we create an evidence submission window task for AMA appeals.


app/models/tasks/disposition_task.rb
```diff
@@ -60,6 +60,14 @@ def cancel!
       fail HearingDispositionNotCanceled
     end
 
+    if appeal.is_a? Appeal
+      EvidenceSubmissionWindowTask.create!(
+        appeal: appeal,
+        parent: hearing_task.parent,
```
sharonwarner: Is this logic correct? I don't think the Evidence Submission Window Task should be underneath the Hearing Task - it's not really associated with a hearing. This made the most sense to me.

Issue #11049, opened 2019-06-06
Include tasks as part of the organization queue configuration


lowellrex:
#11045 uses the configuration the back-end passes to the front-end to generate the components for the `OrganizationQueue` while still drawing the tasks from the global redux state. This ticket exists to pass tasks that belong in each tab as part of that configuration and use those to populate the tables.

## Acceptance criteria
* [x] (back-end) Return tasks as an element of each tab in the queue config #11105
* [x] (front-end) Use tasks returned as element in each tab from the back-end's configuration instead of the tasks from the global redux store. (completed as part of #11213)

PR #11105, opened 2019-06-17
Include tasks as element in QueueConfig

lowellrex:
Connects #11049 (back-end portion only).

Includes tasks as element within each tab returned from the back-end as part of the queue configuration.


app/controllers/organizations/tasks_controller.rb
```diff
@@ -5,10 +5,8 @@ class Organizations::TasksController < OrganizationsController
   before_action :verify_role_access, only: [:index]
 
   def index
-    queue = GenericQueue.new(user: organization)
-    tasks = queue.tasks
-
     # Temporarily limit hearings-management tasks to AOD tasks, because currently no tasks are loading
```
lomky: small nit: this comment is creeping away from its code

app/controllers/organizations/tasks_controller.rb
```diff
@@ -33,11 +35,7 @@ def json_tasks(tasks)
     params = { user: current_user }
 
     AmaAndLegacyTaskSerializer.new(
-      tasks: tasks, params: params, ama_serializer: serializer
+      tasks: tasks, params: params, ama_serializer: organization.ama_task_serializer
```
lomky: I really like this change to letting the objects figure out their serializer

Issue #10337, opened 2019-04-05
Stop creating duplicate ColocatedTasks


lowellrex:
We [recently discovered](https://github.com/department-of-veterans-affairs/caseflow/issues/10335) that we have created a few dozen duplicate `ColocatedTask`s over the past few months at the rate of a few per week. Duplicate `ColocatedTask`s can result in delays to decisions being dispatched, redundant work being performed by the VLJ support staff (especially if a different administrator is assigned to work on the second task), and confusion to users of the Caseflow application. This ticket exists to prevent the creation of duplicate `ColocatedTask`s.

```sql
select
  appeal_type,
  appeal_id,
  type,
  assigned_to_id,
  assigned_to_type,
  action,
  parent_id,
  max(created_at),
  count(*)
from tasks
  where assigned_to_type = 'Organization'
  and status in ('assigned', 'in_progress', 'on_hold')
  and type = 'ColocatedTask'
group by 1, 2, 3, 4, 5, 6, 7
having count(*) > 1;
```

## Acceptance criteria
* [x] Understand how we've been creating duplicate `ColocatedTask`s
* [x] Update application code to prevent the creation of duplicate `ColocatedTask`s
* [ ] Remove (delete or cancel) duplicated `ColocatedTask`s

laurjpeterson:
Support ticket related to this: https://github.com/department-of-veterans-affairs/appeals-support/issues/4428

tomas-nava:
I've dug into this and it's my conclusion that **this is not a bug in our code.**

## How I investigated

I modified Lowell's query to give me task `id`s and `created_at`s and ran it in Looker:
```sql
SELECT appeal_type, appeal_id, type, assigned_to_id, assigned_to_type, action, parent_id, LISTAGG(created_at, ', ') AS agg_created_at, LISTAGG(id, ', '), max(created_at) AS max_created_at, count(*) FROM tasks WHERE assigned_to_type = 'Organization' AND status IN ('assigned', 'in_progress', 'on_hold') AND type = 'ColocatedTask' GROUP BY 1, 2, 3, 4, 5, 6, 7 HAVING count(*) > 1 ORDER BY max_created_at DESC;
```

[...]

PR #11110, opened 2019-06-17
Soft restriction on creating duplicate colocated tasks

tomas-nava:
Connects #10337

### Description
Rejects creation of a new `ColocatedTask` if the appeal, assigned to, action, parent, and instructions are identical. This is restricted at the application level only, no database constraints are added.


pshahVA:
Will there be an error pop up or something?

tomas-nava:
@pshahVA yep, it looks like this:

Screen Shot 2019-06-17 at 1 47 01 PM

app/controllers/application_controller.rb
```diff
@@ -187,7 +187,7 @@ def deny_vso_access
 
   def invalid_record_error(record)
     render json: {
-      "errors": ["title": "Record is invalid", "detail": record.errors.full_messages.join(" ,")]
+      "errors": ["title": COPY::INVALID_RECORD_ERROR_TITLE, "detail": record.errors.full_messages.join(" ,")]
```
lowellrex: 🙌

app/models/tasks/colocated_task.rb
```diff
@@ -161,4 +162,24 @@ def on_hold_duration_is_set
...
+      appeal_id: appeal_id,
+      assigned_to_id: assigned_to_id,
+      assigned_to_type: assigned_to_type,
+      action: action,
+      parent_id: parent_id,
+      instructions: instructions
+    ).find_each do |duplicate_task|
+      if duplicate_task.open?
+        errors[:base] << format(
+          COPY::ADD_HEARING_ADMIN_TASK_ERROR_DUPLICATE,
```
lowellrex: Maybe this can be renamed to something more generic like: `ADD_TASK_ERROR_DUPLICATE`

tomas-nava: I renamed it to something more accurate but not more generic (because it refers specifically to an "action" and not a "task")

PR #11115, opened 2019-06-18
Anya/7378 decision dates in future

### LINKED ISSUE #7378: AMA Intake | Don't allow non rating decision dates in the future
We currently allow free form decision dates in the future. We should have a way of preventing this, either when the issue is added, or when the intake is completed

aroltsch:
Resolves #7378 

### Description
1. Do not allow decision dates in the future
2. Validate date only if the date is fully entered, otherwise disable the save button


client/app/intake/components/NonratingRequestIssueModal.jsx
```diff
@@ -73,10 +73,20 @@ class NonratingRequestIssueModal extends React.Component {
   decisionDateOnChange = (value) => {
     this.setState({
       decisionDate: value,
-      dateError: !validateDate(value)
+      dateError: this.errorOnDecisionDate(value)
     });
   }
 
+  errorOnDecisionDate = (value) => {
+    if (value.length === 10) {
```
aroltsch: I've noticed that we validated the date on every change of the field. I don't know if it was intended behavior so I've changed that.

spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -736,6 +736,9 @@ def complete_higher_level_review
 
         fill_in "Decision date", with: "13/04/2019"
         expect(page).to have_content("Please enter a valid decision date")
+
+        fill_in "Decision date", with: "12/08/2099"
```
pkarman: how about `Time.zone.tomorrow.mdY` instead? that way we don't have to worry about a "2099 problem".

aroltsch: The current date is hardcoded using Timecop feature to the year of 2015. If I add ~10 years from now, this test will fail at some point because of the hard coded current date. I've tried 100.years.from_now but our date validation doesn't allow dates that are 2100 and later :). So I decided to hard code to 2099.

aroltsch: Perhaps I can unfreeze the current date just for this test

pkarman: I see that the problem is that the browser does not respect the `Timecop` since that is server-side only.

Could we send "today" date from the server so that the browser is always dealing with the same date the server is dealing with?

aroltsch: It seems like an overkill to get the test working? Let me see if I can unfreeze the Timecop

spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -736,6 +736,11 @@ def complete_higher_level_review
 
         fill_in "Decision date", with: "13/04/2019"
         expect(page).to have_content("Please enter a valid decision date")
+
+        Timecop.return
+        future_decision_date = 1.day.from_now.strftime("%m/%d/%Y")
+        fill_in "Decision date", with: future_decision_date
+        expect(page).to have_content("Decision date cannot be in the future")
```
leikkisa: 👍

client/app/intake/util/issues.js
```diff
@@ -52,6 +52,17 @@ export const validateDate = (date) => {
   return null;
 };
 
+export const validateDateNotInFuture = (date) => {
+  const currentDate = new Date();
+  const enteredDate = new Date(date);
```
leikkisa: I remember there can be weird timezone handling with the JS date function depending on how you use it, Peter knows more. But I remember exactly what it was and am not too worried about it, just thought it was worth mentioning.

spec/feature/intake/higher_level_review_spec.rb
pkarman: `1.day.from_now.strftime("%m/%d/%Y")` is the same as `Time.zone.tomorrow.mdY`

client/app/intake/util/issues.js
pkarman: it would depend on the format of the `date` string -- if it isn't strictly ISO8601 it will parse as UTC instead of EST.

spec/feature/intake/higher_level_review_spec.rb
aroltsch: oh nice!!!

PR #11116, opened 2019-06-18
Update Rubocop config to match Ruby Style Guide

monfresh:
**Why**:
Absent our own style guide, we choose to follow the community-driven
Ruby Style Guide, which has recently undergone a nice facelift:
https://rubystyle.guide

Quoting from the Guiding Principles:

> It’s common knowledge that code is read much more often than it is
written. The guidelines provided here are intended to improve the
readability of code and make it consistent across the wide spectrum of
Ruby code. They are also meant to reflect real-world usage of Ruby
instead of a random ideal. When we had to choose between a very
established practice and a subjectively better alternative we’ve opted
to recommend the established practice.

> There are some areas in which there is no clear consensus in the
Ruby community regarding a particular style (like string literal
quoting, spacing inside hash literals, dot position in multi-line
method chaining, etc.). In such scenarios all popular styles are
acknowledged and it’s up to you to pick one and apply it consistently.

> Ruby is famously optimized for programmer happiness. We’d like to
believe that this guide is going to help you optimize for maximum
programmer happiness.

As such, I have updated our Rubocop config to remove many cops that
were globally disabled in favor of only disabling them on a case by
case basis. I have also updated our `MethodLength`, `ClassLength`, and
`AbcSize` cops to match the default Rubocop settings.


ferristseng:
I think I'm going to just open up a PR to upgrade Rubocop. Enabling a bunch of rules seems to be too big a lift. I also feel like the Ruby Style Guide is too strict with regards to things like method length / line length and might not suite our needs.

Issue #11078, opened 2019-06-12
Catch TypeError in IDT veterans_controller in informative and friendly way


nemodjuric:
It seems that we are not catching an error in case some of the BGS data is missing. 
For more details see:
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/5229/

PR #11121, opened 2019-06-18
If BGS returns nil for POA use empty hash in IDT response

pkarman:
connects #11078 

Fixes 500 error when IDT tries to fetch a non-existent POA.

### DIFF
--- a/app/controllers/idt/api/v1/veterans_controller.rb
+++ b/app/controllers/idt/api/v1/veterans_controller.rb
@@ -34,7 +34,11 @@ class Idt::Api::V1::VeteransController < Idt::Api::V1::BaseController
 
     return {} unless poa
 
-    poa.merge(bgs.find_address_by_participant_id(poa[:participant_id]))
+    poa_address = bgs.find_address_by_participant_id(poa[:participant_id])
+
+    return poa.reject { |key| key == :participant_id } unless poa_address
+
+    poa.merge(poa_address)
   end
 
   def json_veteran_details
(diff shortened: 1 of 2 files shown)


app/controllers/idt/api/v1/veterans_controller.rb
```diff
@@ -26,10 +26,15 @@ def veteran
...
-    end
+    @poa ||= fetch_poa_with_address
+  end
+
+  def fetch_poa_with_address
+    poa = bgs.fetch_poa_by_file_number(veteran[:file_number])
+
+    return {} unless poa
+
+    poa.merge(bgs.find_address_by_participant_id(poa[:participant_id]))
```
nemodjuric: We should cover the case where poa is not nil but find_address_by_participant_id returns nil (which I suspect is what happens in this case).

pkarman:
thanks @nemodjuric let me know if you have other changes you'd like to see.

app/controllers/idt/api/v1/veterans_controller.rb
```diff
@@ -26,10 +26,19 @@ def veteran
...
+  def fetch_poa_with_address
+    poa = bgs.fetch_poa_by_file_number(veteran[:file_number])
+
+    return {} unless poa
+
+    poa_address = bgs.find_address_by_participant_id(poa[:participant_id])
+
+    return poa unless poa_address
+
+    poa.merge(poa_address)
```
nemodjuric: Maybe use one line for return statement with ternary operator.

PR #11130, opened 2019-06-18
SSN and File Number Search - Frontend

### LINKED ISSUE #10930: Allow users to search for claim number OR SSN

ferristseng:
Resolves #10930

### Description

- Move `app/validators/validate_vso_employee_can_access_file_number.rb` logic to base class, and update it to validate access for each veteran.
- Move some logic to base class, `CaseSearchResultsBase`.
- Update results workflow to accept multiple veteran IDs, and get each of them.
- Update `CaseListSearch` and `CaseListView` to handle search results for multiple veteran records.
- Update search routes to accept comma-separated list of IDs, and accept `veteran_ids` in query string.
- Fix some tests


client/app/queue/QueueApp.jsx
```diff
@@ -278,14 +289,9 @@ class QueueApp extends React.PureComponent {
           {this.props.flash && <FlashAlerts flash={this.props.flash} />}
           <PageRoute
             exact
-            path="/search"
+            path={['/search', '/cases/:caseflowVeteranIds']}
```
ferristseng: This technically allows comma-separated IDs in `/cases/...`, but it lets us to continue reusing the same render function. I think this is OK... just wanted to make a note of that.

Issue #11056, opened 2019-06-06
BGS error Set read error on cd_devactn_grp


pkarman:
Reported during attempt to establish claim for HLR 6720

```
Service Error - Error: Service cannot proceed: Set read error on cd_devactn_grp.   FILE: CDGenTrackedItemsRule.cpp  LINE: 215
```

full stacktrace

```
Traceback (most recent call last):
        9: from (irb):4
        8: from app/models/claim_review.rb:92:in `establish!'
        7: from app/models/claim_review.rb:98:in `block in establish!'
        6: from app/models/end_product_establishment.rb:256:in `generate_tracked_item!'
        5: from app/models/end_product_establishment.rb:506:in `generate_tracked_item_in_bgs'
        4: from app/services/external_api/bgs_service.rb:279:in `generate_tracked_items!'
        3: from app/services/metrics_service.rb:13:in `record'
        2: from app/services/metrics_service.rb:14:in `block in record'
        1: from app/services/external_api/bgs_service.rb:282:in `block in generate_tracked_items!'
Savon::SOAPFault ((S:Server) Service Error - Error: Service cannot proceed: Set read error on cd_devactn_grp.   FILE: CDGenTrackedItemsRule.cpp  LINE: 215)
```

leikkisa:
We are getting this error because the claim is already cleared.  For this particular job, I canceled the establishment job and cleared the error.

To handle this in the future, I suggest checking if the EP's synced_status is CLR, and if it is canceling anything remaining on the establishment job.

leikkisa:
I asked in the BGS meeting today if we needed to worry about side effects of this claim being closed without the tracked item being generated and we decided that if the claim was closed it was okay. I also believe this was probably manually closed because it would have had to be closed quickly in order for this to repeatedly fail.

Closing this ticket.

PR #11133, opened 2019-06-18
Finish establishment processing if EP is inactive

leikkisa:
connects #11056


pkarman:
I think the code climate warning could be addressed by moving the `end_product_establishments.each` block into a private method called `establish_all_end_products` and just calling it from the public method.

Issue #11125, opened 2019-06-18
Make legacy appeals with an S/SOC date before AMA day ineligible


leikkisa:
### Story
Veterans can only opt-in from an S/SOC if the S/SOC was issued on or after 2/19/2019.

https://dsva.slack.com/archives/C754H751T/p1560861871060000

### Acceptance criteria
1. If a legacy appeal has an S/SOC date before 2/19/2019, it is not eligible to be opted in. If a user selects it, they will get the "legacy_appeal_not_eligible" error message

### Technical notes
See eligible_for_soc_opt_in? on LegacyAppeal

Sjones352:
@leikkisa can you please provide instructions on how this can be validated or rather do you have a file number with an S/SOC date before 2/19/2019 in UAT?

PR #11136, opened 2019-06-18
Make legacy appeals with ssoc dates before ama ineligible for opt in

leikkisa:
connects #11125


spec/models/legacy_appeal_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 describe LegacyAppeal do
   before do
-    Timecop.freeze(Time.utc(2015, 1, 1, 12, 0, 0))
+    Timecop.freeze(post_ama_start_date)
```
pkarman: ❤️

spec/support/date_time_helper.rb
```diff
@@ -6,7 +6,7 @@ def post_ama_start_date
   end
 
   def ama_start_date
-    Constants::DATES["AMA_ACTIVATION_TEST"].to_date.in_time_zone
+    Constants::DATES["AMA_ACTIVATION"].to_date.in_time_zone
```
pkarman: +1

pkarman:
can probably fix the code climate issue by grouping a bunch of those `return if` checks into a single private method.

Issue #11040, opened 2019-06-06
Claimant address validation during intake


pkarman:
We should be validating claimant address is complete on initial Intake. Some examples where Supplemental Claims fail to establish: 5611 and 1571

leikkisa:
Jennifer checked on these two claims, and for both of them said that the spouses shouldn't be the claimants, and therefore aren't required to have an address.  The veteran gets paid, but their benefit is increased due to having a spouse.

This might require clarification with the veteran/claimant.  We may also want to check if this is a "contested claimant" situation.

This still should have been prevented by validation.

leikkisa:
SC 1571's claimant has an address of nil, but it happened before the claimant address validation code was merged on March 22nd, so I think that's why that went through.

SC 5611 has an address that looks like this:
`{:address_line_1=>nil, :address_line_2=>nil, :address_line_3=>nil, :city=>nil, :country=>nil, :state=>nil, :zip=>nil}`

This is different than the address just being nil and wasn't being checked for before.

leikkisa:
I canceled the above two claims. According to Jane SC 60137 was established for one of the above veterans (I believe this replaces SC 5611), and the other one was ineligible.

PR #11137, opened 2019-06-18
Validate claimant address when address is present but fields are empty

leikkisa:
connects #11040


app/models/validators/claimant_validator.rb
```diff
@@ -32,11 +32,11 @@ def validate_participant_id(claimant)
   end
 
   def validate_claimant_address(claimant)
-    return if claimant.address
     return unless claimant.decision_review
     return if claimant.decision_review.is_a?(Appeal)
     return unless benefit_type_requires_payee_code?(claimant)
     return unless claimant.participant_id
+    return false if claimant.address_line_1.nil?
```
pkarman: is the `false` important here? the implicit value is `nil`

app/models/validators/claimant_validator.rb
```diff
@@ -1,52 +1,62 @@
 # frozen_string_literal: true
 
-class ClaimantValidator < ActiveModel::Validator
+class ClaimantValidator
```
leikkisa: The ruby guide says "If your validator is complex enough that you want instance variables, you can easily use a plain old Ruby object instead", and between a code climate warning saying I wasn't using this specific instance, and the number of times the word "claimant" was in here, I figured it was time to switch.

PR #11138, opened 2019-06-19
Allow withdrawn appeals to be distributed

### LINKED ISSUE #10973: Change which tasks we cancel when appeals are withdrawn
Currently, we cancel all open tasks on an appeal when an appeal is withdrawn (all issues are withdrawn). We need to change this logic based on what we've learned about distributing withdrawn appeals to judges for decision. 

## AC
- When all request issues on an appeal are withdrawn, cancel all open tasks except:
   - `RootTask`
   - `TrackVeteranTask`
- Add `DistributionTask` with status assigned

monfresh:
**Why**: Judges need to provide a decision for withdrawn appeals.
The definition of a withdrawn appeal is one where all request issues
have been withdrawn, even if they are ineligible or closed for other
reasons.

Previously, withdrawing all request issues would cancel all of the
appeal's active tasks. However, in order to allow the appeal to be
distributed, we cannot cancel its `RootTask` or `TrackVeteranTask`,
and we need to set its `DistributionTask`'s status to `assigned`.

If an appeal has a mixture of withdrawn and ineligible request issues,
or a mixture of withdrawn and closed request issues, then we should
cancel all active tasks

Resolves #10973

### DIFF
--- a/app/policies/withdrawn_decision_review_policy.rb
+++ b/app/policies/withdrawn_decision_review_policy.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+class WithdrawnDecisionReviewPolicy
+  delegate :request_issues, to: :decision_review
+
+  def initialize(decision_review)
+    @decision_review = decision_review
+  end
+
+  def satisfied?
+    no_active_request_issues? && at_least_one_withdrawn_issue?
+  end
+
+  private
+
+  attr_reader :decision_review
+
+  def no_active_request_issues?
+    request_issues.active.empty?
+  end
+
+  def at_least_one_withdrawn_issue?
+    request_issues.withdrawn.any?
+  end
+end
--- a/app/services/appeal_active_task_cancellation.rb
+++ b/app/services/appeal_active_task_cancellation.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+class AppealActiveTaskCancellation
+  delegate :request_issues, :tasks, to: :appeal
+
+  def initialize(appeal)
+    @appeal = appeal
+  end
+
+  def call
+    if WithdrawnDecisionReviewPolicy.new(appeal).satisfied?
+      # Withdrawn appeals must be included in the automatic distribution pool.
+      # For an appeal to be considered ready for distribution, it must have an
+      # assigned DistributionTask. In order for the appeal to progress after
+      # distribution, its RootTask and TrackVeteranTask cannot be cancelled.
+      cancel_active_tasks_except_those_needed_for_distribution
+      assign_distribution_task
+    elsif no_active_request_issues?
+      cancel_active_tasks
+    end
+  end
+
+  private
+
+  attr_reader :appeal
+
+  def cancel_active_tasks_except_those_needed_for_distribution
+    all_tasks_except_those_needed_for_distribution.each(&:cancel_task_and_child_subtasks)
+  end
+
+  def assign_distribution_task
+    tasks.find_by(type: "DistributionTask").update!(status: "assigned")
+  end
+
+  def all_tasks_except_those_needed_for_distribution
+    tasks.where.not(type: tasks_needed_for_distribution)
+  end
+
+  def tasks_needed_for_distribution
+    %w[TrackVeteranTask RootTask]
+  end
+
+  def no_active_request_issues?
+    request_issues.active.empty?
+  end
+
+  def cancel_active_tasks
+    tasks.each(&:cancel_task_and_child_subtasks)
+  end
+end
--- a/spec/feature/intake/higher_level_review/edit_spec.rb
+++ b/spec/feature/intake/higher_level_review/edit_spec.rb
@@ -1335,6 +1335,7 @@ feature "Higher Level Review Edit issues" do
         click_edit_submit_and_confirm
 
         expect(page).to have_content(Constants.INTAKE_FORM_NAMES.higher_level_review)
+        sleep 1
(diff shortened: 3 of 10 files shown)


spec/workflows/appeal_withdrawal_spec.rb
```diff
@@ -0,0 +1,274 @@
+require "rails_helper"
```
pkarman: I think this is automatically required in all spec files.

pkarman: frozen string literal?

spec/workflows/appeal_withdrawal_spec.rb
```diff
@@ -0,0 +1,274 @@
...
+        decision_review: appeal
+      )
+      appeal.request_issues = [eligible_request_issue, closed_request_issue]
+      appeal.save!
+      appeal
+    end
+  end
+
+  def user
+    @user ||= create(:user)
```
pkarman: does this memoize on each `it` invocation the same way `let` does?

app/services/appeal_active_task_cancellation.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def no_active_request_issues?
+    appeal.request_issues.active.empty?
+  end
+
+  def at_least_one_withdrawn_issue?
+    appeal.request_issues.withdrawn.any?
+  end
+
+  def tasks
```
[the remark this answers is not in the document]
monfresh: Yes! I was thinking about that.

app/models/higher_level_review.rb
```diff
@@ -97,6 +97,10 @@ def available_review_options
     %w[supplemental_claim appeal]
   end
 
+  def cancel_active_tasks
```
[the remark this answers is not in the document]
monfresh: Good call.

spec/workflows/appeal_withdrawal_spec.rb
monfresh: Yes and yes. The reason why CC is not complaining about the missing frozen string literal is because we are ignoring all spec files in the CC config, which is not what we want. I will open a PR to fix that.

spec/workflows/appeal_withdrawal_spec.rb
monfresh: I believe so.

app/policies/withdrawn_decision_review_policy.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def satisfied?
+    all_active_request_issues_withdrawn?
+  end
+
+  private
+
+  attr_reader :decision_review
+
+  def all_active_request_issues_withdrawn?
```
monfresh: I think this method name is confusing. I think I will just move the contents of this method into the `satisfied?` method

spec/feature/intake/higher_level_review/edit_spec.rb
```diff
@@ -1330,6 +1330,7 @@ def click_cancel(visit_page)
         click_edit_submit_and_confirm
 
         expect(page).to have_content(Constants.INTAKE_FORM_NAMES.higher_level_review)
+        sleep 1
```
monfresh: I had to add this to see the spec fail locally before I fixed the bug.

leikkisa: Yeah, I sometimes worry about the extra time it takes to process things with the extracted classes, it's noticeably slower. But I do like having the code more organized.

monfresh: Interesting. I'm curious what makes you feel the class extraction is noticeably slower. Is that based on benchmarks? I could be wrong, but I don't think having extra classes affects performance in a significant way. I'm fairly confident the reason why the task statuses take so long to update is because the existing `RequestIssuesUpdate#perform!` method makes a bunch of DB calls, including always talking to VACOLS even when there aren't any Legacy Appeals being processed.

The reason why this spec now needs a sleep is because before, it would not have detected a bug in the code. In this spec, the app should not attempt to cancel tasks, therefore the task statuses are not expected to change. If you look at the spec above it, you will notice that it already had a sleep before my PR. Someone else noticed how slow the Request Issues Update feature is and added the sleep to make sure the test has enough time to update the tasks.

Earlier, when I broke task cancellation for Claim Reviews, this spec was passing locally because the test did not have enough time to see that the tasks were actually canceled. By adding the sleep, we can properly detect the bug and make sure it is fixed.

app/policies/withdrawn_decision_review_policy.rb
monfresh: I completely agree that `appeal.withdrawn?` is the way to go in `AppealActiveTaskCancellation` once `Appeal#withdrawn?` is updated to call this policy class.

spec/feature/intake/higher_level_review/edit_spec.rb
leikkisa: Ah okay, I might be wrong!  I've used this pattern in a couple of places now (I really like it!), but I have had to change tests away from checking that things are updated with Time.zone.now. I just put within 1 second of Time.zone.now and then it's all good.  It thought you needing to add the sleep here might be related, but it sounds like it's not.

monfresh: Got it. Thinking about this some more, I think these tests belong in a model spec, not a feature spec. Checking for DB changes is typically not something done in a feature spec. Since feature specs are slow, we want to keep them at a minimum. In this case, the cancellation of tasks is not user-facing on the appeal edit page, so it doesn't belong in a feature spec IMO. I will open a new issue to move theses tests.

PR #11140, opened 2019-06-19
Close JudgeAssignTasks upon creating AttorneyTasks rather than changing them to JudgeDecisionReviewTasks

### LINKED ISSUE #8408: Stop changing JudgeTask type
When a judge at the Board receives an appeal to make a decision on, they take a sequence of steps in order to hand the appeal off to the next person in the lifetime of an appeal. The judge is required to sign a decision on the appeal and in order to do so they always assign the appeal to an attorney on their team to draft the decision, and then sign that decision once the attorney has finished drafting it. Judges also occasionally return decisions back to the attorney that drafted it.

We currently represent this in the task model as a task that is created with the type `JudgeAssignTask` and that type [changes to `JudgeReviewTask`](https://github.com/department-of-veterans-affairs/caseflow/blob/ffedf94503aaf3a0ee3e6fa8deda2886b392727a/app/models/tasks/judge_task.rb#L49) once the attorney has finished drafting their decision (in the language of the application, once the child `AttorneyTask` is completed).

### Currently:
```
                                   |
-------------------------------    |    -------------------------------
                                   |
       Root task                   |           Root task
                                   |
-------------------------------    |    -------------------------------
         |                         |             |
         |                         |             |
         v                         |             v
     +-------------------+         |         +-------------------+
     |                   |         |         |                   |  # Task changes type
     |  JudgeAssignTask  |         |         |  JudgeReviewTask  |  # when child attorney
     |                   |         |         |                   |  # task is completed.
     +--------+----------+         |         +--------+----------+
              |                    |                  |    ^
              |                    |                  |    |
              v                    |                  v    |
         +--------------+          |             +--------------+
         |              |          |             |              |
         | AttorneyTask |          |             | AttorneyTask |
         |              |          |             |              |
         +--------------+          |             +--------------+
                                   |
```

We change the type of the judge task so that we can display a different set of available actions to the judge in the available actions dropdown menu. This is a little artificial because the task the judge has to complete has not actually changed even though the type has (the judge still has to deliver a signed decision). I think we should change the way the application represents the different steps in the process of a judge delivering a signed decision to better match the process.

## Option 1: Use child AttorneyTask presence to determine actions

Currently we use the task type alone to determine which actions are available to the judge for a given appeal:

`JudgeAssignTask`
* Assign to attorney

`JudgeReviewTask`
* Ready for Dispatch
* (soon!) Return to attorney for corrections (or something similar)

Instead we could use the presence of a completed child attorney task to make that same determination:

No child `AttorneyTask`
* Assign to attorney

Completed child `AttorneyTask`
* Ready for Dispatch
* (soon!) Return to attorney for corrections (or something similar)

The task tree would look like the following in this case:

```
                                   |
-------------------------------    |    -------------------------------
                                   |
       Root task                   |           Root task
                                   |
-------------------------------    |    -------------------------------
         |                         |             |
         |                         |             |
         v                         |             v
     +-------------------+         |         +-------------------+
     |                   |         |         |                   |
     |     JudgeTask     |         |         |     JudgeTask     |
     |                   |         |         |                   |
     +--------+----------+         |         +--------+----------+
              |                    |                  |    ^
              |                    |                  |    |
              v                    |                  v    |
         +--------------+          |             +--------------+
         |              |          |             |              |
         | AttorneyTask |          |             | AttorneyTask |
         |              |          |             |              |
         +--------------+          |             +--------------+
                                   |
```

## Option 2: Create separate assign and review tasks

If we want to retain the distinct `JudgeAssignTask` and `JudgeReviewTask`s we could create the latter when a judge tags an attorney to draft an appeal. The available actions would remain the same as they are now (determined by the task type), but the task tree would change to look like the following:

```
------------------------------------------------------------

     Root task

------------------------------------------------------------
          |      ^             |
          |      |             |
          v      |             v
   +-------------------+   +-------------------+
   |                   |   |                   |
   |  JudgeAssignTask  |   |  JudgeReviewTask  |
   |                   |   |                   |
   +-------------------+   +-------------------+
                                  |
    # JudgeAssignTask             |
    # marked as complete          v
    # when appeal assigned     +--------------+
    # to an attorney           |              |
                               | AttorneyTask |
                               |              |
                               +--------------+
```

Curious to hear thoughts on these approaches, whether you think this is actually a problem, or if there are alternative ways to address this concern.

hschallhorn:
Resolves #8408

### Description

1. Rather than changing the type of `JudgeAssignTask` to `JudgeDecisionReviewTask` when an `AttorneyDecisionTask` is completed, when a judge assigns cases to an attorney, close the `JudgeAssignTask` and create the `AttorneyDecisionTask` with a `JudgeDecisionReviewTask` as its parent.
2. Ensure all tests adhere to this model
3. Create a new subclass, `AttorneyDecisionTask`, for `AttorneyTask` to more accurately describe what type of task is being created, rather than a generic attorney task (much like how `JudgeTask` is a parent to all other judge tasks)


app/models/tasks/attorney_task.rb
```diff
@@ -14,23 +14,6 @@ class AttorneyTask < Task
   validate :assigned_to_role_is_valid
   validate :child_attorney_tasks_are_completed, on: :create
 
-  class << self
-    def create_many_from_params(params_array, _user)
```
hschallhorn: Was being touched by AttorneyRewriteTask. create_many_from_params, AttorneyQualityReviewTask. create_many_from_params, and AttorneyDispatchReturnTask. create_many_from_params, which we obviously don't want happening, so I moved it to its own subclass as we only want this specific behavior with attorney decision writing tasks

lowellrex:
Can we delay creating the `AttorneyDecisionTask` class to a later PR so this set of changes stays limited to not changing the type of the `JudgeTask`?

hschallhorn:
> Can we delay creating the AttorneyDecisionTask class to a later PR so this set of changes stays limited to not changing the type of the JudgeTask?

app/models/tasks/attorney_task.rb
```diff
@@ -32,11 +32,6 @@ def timeline_title
     COPY::CASE_TIMELINE_ATTORNEY_TASK
   end
 
-  def update_parent_status
-    parent.begin_decision_review_phase if parent&.is_a?(JudgeAssignTask)
-    super
-  end
-
```
lowellrex: What happens to existing `AttorneyTask`s that are children of `JudgeAssignTask`s right now if we remove this code? Will we manually update the task tree for all of the existing data (by creating a new `JudgeDecisionReviewTask`, making the `AttorneyTask` the child of that new task, and then closing the `JudgeAssignTask` that the `AttorneyTask` was previously a child of)?

I think it may be simpler to leave this code in for the time being (maybe we could add a sentry alert to let us know to remove this code a month after we deploy this code?) so that we don't have to update a lot of live data.

I'm definitely up for changing the live data though, if that's the route we prefer. But if it is I think we need to walk through what that looks like in detail before merging this PR.

app/models/tasks/attorney_decision_task.rb
```diff
@@ -0,0 +1,20 @@
...
+          judge_assign_task = JudgeAssignTask.find(params[:parent_id])
+          judge_review_task = JudgeDecisionReviewTask.create!(judge_assign_task.slice(:appeal, :assigned_to, :parent))
+
+          judge_assign_task.update!(status: Constants.TASK_STATUSES.completed)
+
+          attorney_task = create!(params.merge(parent_id: judge_review_task.id))
+
+          [attorney_task, judge_review_task, judge_assign_task]
+        end.flatten
+      end
```
lowellrex: I like this approach!

app/models/tasks/attorney_task.rb
hschallhorn: Looks like theres 138 of these. I'm in for leaving this code in. Just in case something is wonky with our approach.

hschallhorn: Tossed in a sentry alert that links to this conversation. Hello person from the future!

app/models/tasks/judge_assign_task.rb
```diff
@@ -10,6 +10,9 @@ def additional_available_actions(_user)
 
   def begin_decision_review_phase
     update!(type: JudgeDecisionReviewTask.name)
+    # Tell sentry so we know this is still happening. Remove this in a month
+    Raven.capture_message("Still changing JudgeAssignTask type to JudgeDecisionReviewTask.
+      See: https://github.com/department-of-veterans-affairs/caseflow/pull/11140#discussion_r295487938")
```
lowellrex: Let's only start throwing this message a month from now as a reminder to see where we are in the migration. Perhaps something like: `Raven.capture_message(...) if Time.zone.now > Time.new(2019, 8, 1)`?

lowellrex: We can [run a query](https://caseflow-looker.va.gov/sql/9vbvbc9qznxsth) that looks something like the following to determine how many tasks will still trigger this function:
```sql
select count(*)
from tasks atty_tasks
join tasks judge_tasks
  on judge_tasks.id = atty_tasks.parent_id
where atty_tasks.type = 'AttorneyTask'
  and judge_tasks.type = 'JudgeAssignTask'
  and atty_tasks.status not in ('completed', 'cancelled')
  and judge_tasks.status not in ('completed', 'cancelled');
```

PR #11141, opened 2019-06-19
Fix Hearing Location on Hearing Details

### LINKED ISSUE #10073: Update hearing location in Hearing Details
Hearing location needs to be updated to reflect Washington, DC for Central Office hearings. It was showing the VACOLS RO as hearing location.

Screen Shot 2019-03-20 at 11 56 25 AM

ferristseng:
Resolves #10073 

### Description

- Refactors `to_hash` logic to serializers for Hearing and LegacyHearing.
- Implement and serialize `readableLocation` for Hearing. This was being used on the frontend, but didn't exist for non-legacy hearings.

### DIFF
--- a/app/models/legacy_hearing.rb
+++ b/app/models/legacy_hearing.rb
@@ -205,7 +205,7 @@ class LegacyHearing < ApplicationRecord
   end
 
   def to_hash(current_user_id)
-    LegacyHearingSerializer.default(
+    ::LegacyHearingSerializer.default(
       self,
       params: { current_user_id: current_user_id }
     ).serializable_hash[:data][:attributes]
@@ -226,7 +226,7 @@ class LegacyHearing < ApplicationRecord
   end
 
   def to_hash_for_worksheet(current_user_id)
-    LegacyHearingSerializer.worksheet(
+    ::LegacyHearingSerializer.worksheet(
       self,
       params: { current_user_id: current_user_id }
     ).serializable_hash[:data][:attributes]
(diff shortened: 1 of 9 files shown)


app/serializers/hearings/hearing_serializer_base.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module HearingSerializerBase
+  extend ActiveSupport::Concern
+
+  class_methods do
```
ferristseng: This section was causing a lot of issues for reek. These are intended to act as alternate constructors for the serializers. I ignored both the warnings because it seems like it would introduce more complexity if I fixed them.

app/models/legacy_hearing.rb
```diff
@@ -100,7 +100,9 @@ def hearing_day
   end
 
   def regional_office_key
-    return (venue_key || appeal&.regional_office_key) if request_type == "T" || hearing_day.nil?
+    if request_type == HearingDay::REQUEST_TYPES[:travel] || hearing_day.nil?
```
ferristseng: Made this use the variable instead of the `"T"`

client/app/hearings/components/AssignHearingsTabs.jsx
```diff
@@ -175,10 +175,6 @@ export default class AssignHearingsTabs extends React.Component {
 
   };
 
-  getHearingLocation = (hearing) => (
-    hearing.readableRequestType === 'Central' ? 'Washington, DC' : hearing.readableLocation
-  );
-
```
sharonwarner: Yay!

PR #11142, opened 2019-06-19
Use CacheManager rather than blindly nuking Redis on re-seed

pkarman:
see https://dsva.slack.com/archives/C2ZAMLK88/p1560970532153800


enriquemanuel:
lgtm

PR #11143, opened 2019-06-19
Re-enable BGS lookup for POA

pkarman:
This reverts commit ac9bf0016e24e9a9b841e85de7d24889b1e1df90.

see #11126 

When initially deployed, the BGS lookup was (on aggregate) too slow for production use. This re-enables the BGS lookup with 24 hours of Rails caching per-participant. There is also a job (which will eventually be run nightly) that can pre-warm all the caches so that users do not suffer the caching expense.


app/jobs/warm_bgs_participant_address_caches_job.rb
```diff
@@ -0,0 +1,30 @@
...
+  def process(ro_ids)
+    ro_ids.each do |ro_id|
+      regional_office = HearingDayMapper.validate_regional_office(ro_id)
+
+      HearingDay.open_hearing_days_with_hearings_hash(
+        Time.zone.today.beginning_of_day,
+        Time.zone.today.beginning_of_day + 182.days,
+        regional_office,
+        RequestStore.store[:current_user].id
+      )
```
tomas-nava: Why use `HearingDay.open_hearing_days_with_hearings_hash` to cache veterans' addresses?

app/services/bgs_address_service.rb
```diff
@@ -15,11 +15,15 @@ def address
...
-    nil
+    cache_key = "bgs-participant-address-#{participant_id}"
+
+    Rails.cache.fetch(cache_key, expires_in: 24.hours) do
+      bgs.find_address_by_participant_id(participant_id)
+    rescue Savon::Error
+      # If there is no address for this participant id then we get an error.
+      # catch it and return an empty array
+      nil
+    end
```
tomas-nava: Should we add a cache to `BgsPowerOfAttorney.fetch_bgs_record` as well?

app/services/bgs_power_of_attorney.rb
```diff
@@ -21,9 +21,15 @@ def bgs
 
   def fetch_bgs_record
     if claimant_participant_id
-      bgs.fetch_poas_by_participant_ids([claimant_participant_id])[claimant_participant_id]
+      cache_key = "bgs-participant-#{claimant_participant_id}"
```
tomas-nava: Could we name this cache key `bgs-participant-poa-` to use the convention established with `bgs-participant-address-`?

app/jobs/warm_bgs_participant_address_caches_job.rb
pkarman: because that was the endpoint that made us aware of the performance hit.

https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/hearings/hearing_day_controller.rb#L45

the assumption is that hearing day is appropriate because it hits all the POAs for a time window. doing one BGS fetch at a time is not a performance bottleneck. It's when we need to grab dozens at a time on a single HTTP request.

pkarman: caching it at the service level should improve performance for one-off requests as well though, assuming it will hit the cache.

Issue #11152, opened 2019-06-20
Fix ColocatedTasks for LegacyAppeals being charged to wrong location in VACOLS


lowellrex:
`ColocatedTask`s for `LegacyAppeal`s are sometimes being charged to the wrong location in VACOLS when they are marked complete. During initial investigation of this bug [we added a test](https://github.com/department-of-veterans-affairs/caseflow/pull/11134) that sought to confirm the fix deployed in #10853 fixed all scenarios where charging to the wrong location. That test works as expected, but there is still some scenario where this bug still occurs.

The fix in #10853 was deployed on May 24th, but we still see occurrences of this bug after that date.

![image](https://user-images.githubusercontent.com/32683958/59850560-85938500-9338-11e9-831f-3e4975cac42a.png)

The following query (https://caseflow-looker.va.gov/sql/mcfj2dqyssdcgj) reveals an upper bound on the number of `ColocatedTask`s affected by this bug of 281.
```sql
select id
from tasks
where parent_id in ( select parent_id
  from tasks
  where type = 'ColocatedTask'
    and assigned_to_type = 'User'
    and appeal_type = 'LegacyAppeal'
    and status = 'cancelled'
)
  and type = 'ColocatedTask'
  and assigned_to_type = 'User'
  and appeal_type = 'LegacyAppeal'
  and status = 'completed';
```

We can narrow down the number of tasks affected by this bug to just those cases that are still charged to the (presumably) incorrect VACOLS location in the rails console:
```ruby
  rails c> all_potentially_affected_tasks = Task.find_by_sql("select *
from tasks
where parent_id in ( select parent_id
  from tasks
  where type = 'ColocatedTask'
    and assigned_to_type = 'User'
    and appeal_type = 'LegacyAppeal'
    and status = 'cancelled'
)
  and type = 'ColocatedTask'
  and assigned_to_type = 'User'
  and appeal_type = 'LegacyAppeal'
  and status = 'completed';")
  rails c> all_potentially_affected_tasks.count
  # 281
  # The operation below hits vacols for each of the 281 tasks so it might be a little slow
  # to run, 90 seconds or so.
  rails c> still_mischarged_cases = all_potentially_affected_tasks.select { |task| task.assigned_by.vacols_uniq_id == task.appeal.location_code }
  rails c> still_mischarged_cases.count
  # 93 (as of 9am eastern 20 June 2019)
```

[...]

PR #11153, opened 2019-06-20
Add logging for ColocatedTask VACOLS location bug

lowellrex:
Connects #11152.

Adds alerting in Sentry when this situation comes up so we can inspect the state of the task tree to understand why this is happening. Example [post to slack](https://dsva.slack.com/archives/C6E41RE92/p1561040407063800) and [issue in Sentry](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/5383/).

### DIFF
--- a/app/models/tasks/colocated_task.rb
+++ b/app/models/tasks/colocated_task.rb
@@ -149,6 +149,9 @@ class ColocatedTask < Task
       "Data to aid in the investigation of ColocatedTasks for LegacyAppeals being charged to incorrect VACOLS location \
 bug described in https://github.com/department-of-veterans-affairs/caseflow/issues/11152",
 
+      # Task ID.
+      "Task triggering VACOLS location change ID: #{id}",
+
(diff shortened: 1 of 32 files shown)


app/models/tasks/colocated_task.rb
```diff
@@ -131,10 +131,51 @@ def update_location_in_vacols
...
+bug described in https://github.com/department-of-veterans-affairs/caseflow/issues/11152",
+
+      # Task ID.
+      "Task triggering VACOLS location change ID: #{id}",
+
+      # Future location.
+      "Future location: #{location_based_on_action}",
+
+      # Information about the task tree's current state.
+      appeal.tasks.order(:created_at).pluck(
```
aroltsch: I would use a `map` and return array of hashes for easier readability. 
```
appeal.tasks.order(:created_at).map { |t| { id: t.id, parent_id: t.parent_id, ... } }
```

PR #11155, opened 2019-06-20
hardcode limit on tasks returned by /organizations/tasks endpoint

pkarman:
until server-side pagination is complete, hardcode a limit on the number of tasks returned.

### DIFF
--- a/app/controllers/organizations/tasks_controller.rb
+++ b/app/controllers/organizations/tasks_controller.rb
@@ -19,7 +19,8 @@ class Organizations::TasksController < OrganizationsController
   def tasks
     # Temporarily limit hearings-management tasks to AOD tasks, because currently no tasks are loading
     if organization.url == "hearings-management"
-      GenericQueue.new(user: organization, limit: 1000).tasks.select { |task| task.appeal.is_a?(Appeal) || task.appeal.aod }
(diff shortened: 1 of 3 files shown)


app/controllers/organizations/tasks_controller.rb
```diff
@@ -23,6 +16,14 @@ def index
 
   private
 
+  def tasks
+    if organization.url == "hearings-management"
+      GenericQueue.new(user: organization).tasks(1000).select { |task| task.appeal.is_a?(Appeal) || task.appeal.aod }
+    else
+      GenericQueue.new(user: organization).tasks(400)
```
aroltsch: I would pass the limit when instantiating the object and set the default in the initializer to 400

app/controllers/organizations/tasks_controller.rb
```diff
@@ -5,13 +5,6 @@ class Organizations::TasksController < OrganizationsController
   before_action :verify_role_access, only: [:index]
 
   def index
-    tasks = GenericQueue.new(user: organization).tasks
-
-    # Temporarily limit hearings-management tasks to AOD tasks, because currently no tasks are loading
```
aroltsch: We should probably leave the comment

PR #11163, opened 2019-06-21
Automatically add BGS VSO team members to the VSO team in Caseflow

### LINKED ISSUE #11041: Automatically add VSO staff to VSO organization in Caseflow
A user's membership in a VSO is determined by [the relationship stored in BGS](https://github.com/department-of-veterans-affairs/caseflow/blob/9f31758c7b91795d825dfc5bd44b7a1d613eeb54/app/models/representative.rb#L6). We use this BGS relationship to ensure that we don't expose any Veteran data to people who are not staff of the appellant's elected VSO because BGS should always be up to date with that membership information.

However, the ability to assign tasks to VSO staff and the ability for those staffers to navigate from their personal queues depends on those users being identified as members of the VSO organization in Caseflow. Currently, the only way for users to be added as members of a VSO is for them to be added to the organization by a VSO admin using [the user admin interface](https://github.com/department-of-veterans-affairs/caseflow/pull/7454). This PR exists to automatically add VSO staff as members of the VSO organization in Caseflow.

One way we could accomplish this would be to add them as organization members the first time they [visit the VSO organization's team queue](https://github.com/department-of-veterans-affairs/caseflow/blob/b6a4088edc01e1a3cb8d7acabe30a62ef274bf08/app/controllers/organizations_controller.rb#L10).

## Acceptance criteria
* [ ] Automatically add VSO staff to the VSO organization if they are members but not yet identified as such in Caseflow when they visit the VSO team queue.

hschallhorn:
Resolves #11041

### Description
Automatically adds BGS VSO team members to the VSO team in Caseflow when they visit the VSO team page.

### Testing Plan
1. Create a new VSO user 
```ruby
> u = User.create!(css_id: "NEW_VSO", station_id: 101, roles: ["VSO"])
```
2. Login as user
3. Check to make sure the user is not a part of the VSO team
```ruby
> u.organizations
=> []
```
4. Visit the VSO team page at http://localhost:3000/organizations/veterans-service-organization
5. Check to make sure the user is now a part of the VSO team
```ruby
> u.reload.organizations
=> [#<Vso:0x00007fc2187ed948
  id: 6,
  name: "VSO",
  participant_id: "2452415",
  role: "VSO",
  type: "Vso",
  url: "veterans-service-organization">]
```


app/controllers/organizations_controller.rb
```diff
@@ -4,6 +4,7 @@ class OrganizationsController < ApplicationController
   before_action :verify_organization_access
   before_action :verify_role_access
   before_action :verify_business_line, only: [:show]
+  before_action :add_user_to_vso, only: [:show]
```
hschallhorn: Because this action is performed last, the [check to ensure](https://github.com/department-of-veterans-affairs/caseflow/blob/9f31758c7b91795d825dfc5bd44b7a1d613eeb54/app/models/representative.rb#L6) the user is a member of the vso in bgs has already occurred

lomky: I'm mildly in favor of a comment to that effect, since reordering the `before_action`s could cause a problem.

spec/controllers/organizations_controller_spec.rb
```diff
@@ -0,0 +1,84 @@
+# frozen_string_literal: true
+
+describe OrganizationsController, type: :controller do
```
lomky: :1st_place_medal:

PR #11165, opened 2019-06-21
Set application for IDT controllers to trigger BGS AOD lookup

pkarman:
see #11014 

It appears that this method was never being called in the IDT controller space and so `RequestStore[:application]` was never being set.


app/controllers/idt/api/v1/base_controller.rb
```diff
@@ -3,6 +3,7 @@
 class Idt::Api::V1::BaseController < ActionController::Base
   protect_from_forgery with: :exception
   before_action :validate_token
+  before_action :set_application
```
aroltsch: woah

PR #11166, opened 2019-06-21
Don't ignore specs in Code Climate config

monfresh:
**Why**: We don't want to ignore specs globally. We want specs to
adhere to at least a subset of rules, such as style-related cops
and the frozen string literal.


spec/feature/queue/ama_queue_spec.rb
```diff
@@ -455,24 +456,6 @@ def valid_document_id
...
-        click_on veteran_full_name
-
-        click_dropdown(prompt: "Select an action", text: "Assign to attorney")
-        click_dropdown(prompt: "Select a user", text: attorney_user.full_name)
-
-        click_on "Submit"
-
-        expect(page).to have_content("Assigned 1 case")
-      end
-
```
monfresh: All of these removed specs were duplicated exactly word for word earlier in this file.

PR #11169, opened 2019-06-21
Dupe Task Checker

lomaxap:


pkarman:
I like the approach.

A nit about naming. I think there's a nascent trend of putting these in the `app/queries` name space.

PR #11173, opened 2019-06-21
Fix tests in search spec

ferristseng:
### Description

See the comments for the changes


spec/feature/queue/search_spec.rb
```diff
@@ -489,10 +495,8 @@ def perform_search
   end
 
   context "has withdrawn decision reviews" do
-    let!(:caseflow_appeal) do
-      create(:appeal,
-             :with_tasks)
-    end
+    let!(:veteran) { create(:veteran) }
```
ferristseng: The call to fake `BGSService.fetch_file_number_by_ssn` was reversing a SSN number, and causing a veteran to be created with a file number different than the one the test appeal was created with.

Observed path in code:

```
   find with appeal veteran file number 
-> no veteran, tries to create a veteran 
-> fake BGSService.fetch_file_number_by_ssn reverses file number 
-> veteran gets created where the file number is a reverse of the ssn
-> no appeal gets found since the original appeal has the un-reversed file number
```

**Fix:** create a veteran and create the relationship to the appeal before making the search query.

spec/feature/queue/search_spec.rb
```diff
@@ -408,7 +414,7 @@ def perform_search
       end
 
       it "searching in search bar works" do
-        fill_in "searchBarEmptyList", with: appeal.sanitized_vbms_id
+        fill_in "searchBarEmptyList", with: appeal.sanitized_vbms_id, fill_options: { clear: :backspace }
```
ferristseng: Occasionally the search input field would be populated by a long search input. This was happening because multiple search inputs were being concatenated together.

Screen Shot 2019-06-21 at 5 05 58 PM

**Fix:** clear the input before writing the new value to search by. See: https://github.com/erikras/redux-form/issues/686

spec/feature/queue/search_spec.rb
```diff
@@ -289,16 +292,19 @@ def perform_search
...
         click_on "Search"
 
         expect(page).to have_content("1 case found for")
         expect(page).to have_content(COPY::CASE_LIST_TABLE_DOCKET_NUMBER_COLUMN_TITLE)
       end
 
-      it "clicking on the x in the search bar returns browser to queue list page" do
+      it "clicking on the x in the search bar clears the search bar" do
```
ferristseng: This test seemed like it was testing old behavior. Clicking the X on the search bar doesn't cause the browser to navigate back to the queue list page, so I changed to to make sure the input was getting cleared.

PR #11177, opened 2019-06-21
Only update VACOLS location for organization ColocatedTasks

lowellrex:
Connects #11152. This PR limits the changing of VACOLS location codes when closing `ColocatedTask`s associated with `LegacyAppeal`s to organizationally-assigned tasks to circumvent a bug in how we are determining the current VACOLS location of an appeal.

[The logging](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/5409/) that we added in #11153 revealed that the task tree was in the same state for the individually-assigned `ColocatedTask` and its organizationally-assigned parent. This led us to scrutinize how the `after_update` hooks on `Task` and `ColocatedTask` are interacting with one another which we can describe in outline form below.

Individually-assigned `ColocatedTask`
* Status changed to `completed`
* `Task.update_parent_status` (updates parent)
    * Organizationally-assigned `ColocatedTask` status changed to `completed`
        * `Task.update_parent_status` (does nothing since there is no parent task)
        * `ColocatedTask.update_location_in_vacols` (updates the status to the organizationally-assigned task's assigner)
* `ColocatedTask.update_location_in_vacols` (updates the status to the individually-assigned task's assigner)

I don't know why the second `update_location_in_vacols` is passing the check for the [current VACOLS location code](https://github.com/department-of-veterans-affairs/caseflow/blob/2b191245284a9968c2cf4e9daf438ad9ae246f2d/app/models/tasks/colocated_task.rb#L133) being `CASEFLOW` when it should be the organizationally-assigned task's assigner's VACOLS unique ID (perhaps because we are caching that value somewhere? Database connection maybe?), but this PR avoids dealing with that issue with the new `is_a?(Organization)` check.


app/models/tasks/colocated_task.rb
```diff
@@ -125,15 +125,18 @@ def create_and_auto_assign_child_task(_options = {})
     super(appeal: appeal)
   end
 
+  # rubocop:disable Metrics/CyclomaticComplexity
```
aroltsch: Lowell, I think we agreed not to disable Rubocop in files :).

lowellrex: How do I disable this cop for only this single function in `.rubocop.yml`? It doesn't look like we've done that yet.

app/models/tasks/colocated_task.rb
```diff
@@ -125,15 +125,18 @@ def create_and_auto_assign_child_task(_options = {})
...
   def update_location_in_vacols
     if saved_change_to_status? &&
        !open? &&
        all_tasks_closed_for_appeal? &&
        appeal.is_a?(LegacyAppeal) &&
-       appeal.location_code == LegacyAppeal::LOCATION_CODES[:caseflow]
+       appeal.location_code == LegacyAppeal::LOCATION_CODES[:caseflow] &&
+       assigned_to.is_a?(Organization)
```
aroltsch: If you moved this three lines (133-135) into a separate function would it still complain about the complexity?

lowellrex: I don't know. I guess we could. I don't think we combine these three conditions other places and I think doing so would introduce indirection and make the code less comprehensible, so I'd rather avoid doing that by just disabling the rubocop rule here.

app/models/tasks/colocated_task.rb
pkarman: according to the docs, we disable rubocop through special comments in files.

https://github.com/department-of-veterans-affairs/caseflow/pull/11037/files#diff-1d45562ecd42ce63263b4f03d56db21cR14

spec/models/tasks/colocated_task_spec.rb
```diff
@@ -192,7 +192,10 @@
   context ".update" do
     let!(:attorney_2) { FactoryBot.create(:user) }
     let!(:staff_2) { FactoryBot.create(:staff, :attorney_role, sdomainid: attorney_2.css_id) }
-    let!(:colocated_admin_action) { create(:colocated_task, assigned_by: attorney_2) }
+    let(:org_colocated_task) do
+      FactoryBot.create(:colocated_task, assigned_by: attorney_2, assigned_to: Colocated.singleton)
+    end
+    let!(:colocated_admin_action) { org_colocated_task.children.first }
```
lowellrex: More realistic test data setup.

app/models/tasks/colocated_task.rb
```diff
@@ -125,15 +125,18 @@ def create_and_auto_assign_child_task(_options = {})
...
   end
 
+  # rubocop:disable Metrics/CyclomaticComplexity
   def update_location_in_vacols
     if saved_change_to_status? &&
        !open? &&
        all_tasks_closed_for_appeal? &&
        appeal.is_a?(LegacyAppeal) &&
-       appeal.location_code == LegacyAppeal::LOCATION_CODES[:caseflow]
+       VACOLS::Case.find(appeal.vacols_id).bfcurloc == LegacyAppeal::LOCATION_CODES[:caseflow] &&
```
lowellrex: Tests were failing because the `appeal.location_code` value was cached in the application. `appeal.reload.location_code` did not reset the caching we were doing (maybe it is at the Oracle DB connection layer?).

app/models/tasks/colocated_task.rb
aroltsch: You could combine lines 133 and 134 into one method which is more clear I believe:

```
def in_caseflow_location_in_vacols?
   appeal.is_a?(LegacyAppeal) && VACOLS::Case.find(appeal.vacols_id).bfcurloc == LegacyAppeal::LOCATION_CODES[:caseflow]
end
```

app/models/tasks/colocated_task.rb
```diff
@@ -129,12 +129,17 @@ def update_location_in_vacols
...
        all_tasks_closed_for_appeal? &&
-       appeal.is_a?(LegacyAppeal) &&
-       appeal.location_code == LegacyAppeal::LOCATION_CODES[:caseflow]
+       appeal_in_caseflow_vacols_location? &&
+       assigned_to.is_a?(Organization)
       AppealRepository.update_location!(appeal, location_based_on_action)
     end
   end
 
+  def appeal_in_caseflow_vacols_location?
```
aroltsch: I like your name better!

PR #11180, opened 2019-06-24
Update Appeal#withdrawn? to match latest requirements

monfresh:
While working on #11138, we understood the correct definition of a
withdrawn appeal, which is that it has at least one withdrawn request
issue, and no other active request issues. In most cases, when all of
an appeal's request issues are closed, its RootTask gets cancelled,
which is why it made sense for `withdrawn?` to check the RootTask
status. However, in the case of a withdrawal, the RootTask no longer
gets cancelled after implementing #11138.


app/models/appeal.rb
```diff
@@ -488,7 +488,7 @@ def distributed_to_a_judge?
   end
 
   def withdrawn?
-    root_task&.status == Constants.TASK_STATUSES.cancelled
+    WithdrawnDecisionReviewPolicy.new(self).satisfied?
```
leikkisa: Let's add this at the decision_review level since it has the same definition now.

spec/models/appeal_spec.rb
```diff
@@ -1538,15 +1538,15 @@
...
-      it "is withdrawn" do
-        expect(appeal.withdrawn?).to eq(false)
-        root_task.update!(status: Constants.TASK_STATUSES.cancelled)
-        expect(appeal.withdrawn?).to eq(true)
-      end
+      expect(WithdrawnDecisionReviewPolicy).to receive(:new)
+        .with(appeal).and_return(policy)
+      expect(policy).to receive(:satisfied?)
+
+      appeal.withdrawn?
```
leikkisa: This test looks like it's just testing that the code is written a certain way, in which case it seems unnecessary to me (it doesn't feel like it's really testing anything).  It would be nice to test the behavior at a more human level, like with a decision review with request issues.

spec/policies/withdrawn_decision_review_policy_spec.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+describe WithdrawnDecisionReviewPolicy do
+  describe "#satisfied?" do
+    context "withdrawn request issue exists, along with other active request issues" do
```
leikkisa: Why not make a decision review with an active and withdrawn request issue?

app/models/decision_review.rb
```diff
@@ -301,6 +301,14 @@ def removed?
...
   end
 
+  def active_request_issues
+    request_issues.active
+  end
+
+  def withdrawn_request_issues
+    request_issues.withdrawn
+  end
+
```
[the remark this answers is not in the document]
monfresh: This makes it a lot easier to write fast tests. In order to test the policy, now all we need is to pass in any object that responds to `withdrawn_request_issues` and `active_request_issues`. We don't need to instantiate any Rails models or use factories or interact with the DB. Otherwise, we would have to either create a Decision Review factory along with different types of request issues, or create a new fake object that responds to `request_issues`, which in turn responds to `active` or `withdrawn`.

spec/models/appeal_spec.rb
monfresh: I'm writing that test now, but it will be in the `DecisionReview` model spec. We only need one test that actually queries the DB. If `DecisionReview#withdrawn_request_issues` and `DecisionReview#active_request_issues` are unit tested to make sure they return the right data from the DB, then in all the other tests, we just need to make sure the right objects are being called with the right messages. 

In the case of the policy spec, all we need to test is that when an object receives the messages `withdrawn_request_issues` and `active_request_issues`, and returns an array that's either empty or has one or more elements, it returns `true` or `false` depending on the contents of each array. The policy class doesn't actually care what is returned. All it needs to know is whether or not the arrays are populated and that's all it needs to answer `satisfied?`.

leikkisa: Yeah, adding tests for those two new methods at least closes the loop.

app/models/decision_review.rb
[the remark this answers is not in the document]
monfresh: In addition to making the policy a lot easier and faster to test, this change also fixes the violation of the Law of Demeter, which says that an object should only talk to its immediate collaborators. In the policy class, we had this message chain: `decision_review.request_issues.active`. That second dot, where we call methods on `request_issues` is the sign that we might be violating the Law of Demeter. 

A typical example in Rails is when you have a Book that has an Author. Instead of getting the book's author's name by doing `book.author.name`, you would use the Rails `delegate` method so that you can call `book.author_name`:
```ruby
class Book < ApplicationRecord
  has_one :author

  delegate :name, to: :author, prefix: true
end
```
In this case, though, I don't think we can use `delegate` because we would end up with a method named `request_issues_active`, which doesn't sound right to me. I was hoping there would be a `suffix` option but it doesn't look like there is.

Sandi Metz talks about the Law of Demeter in the video we watched during my last Wisdom Wednesday. It starts here: https://youtu.be/PJjHfa5yxlU?t=1309

She also has another great talk about testing, where she goes over dependency injection and only needing to test that the right messages are being passed:
https://www.youtube.com/watch?v=URSWYvyc42M
I think we should watch this at our next Wisdom Wednesday

spec/models/appeal_spec.rb
monfresh: Tests added here: https://github.com/department-of-veterans-affairs/caseflow/pull/11180/commits/eedece460385dc8d35d95f86960aa1fd72472563

app/models/appeal.rb
monfresh: Good point. Done here: https://github.com/department-of-veterans-affairs/caseflow/pull/11180/commits/e98fc5fcfdc2e5e90e4a9b28bfd3fbe21a725d77

PR #11186, opened 2019-06-24
Colocated Duping Test

### LINKED ISSUE #10337: Stop creating duplicate ColocatedTasks

lomaxap:
tests #10337

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -396,6 +396,7 @@ class Task < ApplicationRecord
   private
 
   def create_and_auto_assign_child_task(options = {})
+    # duping here
     dup.tap do |child_task|
       child_task.assigned_to = assigned_to.next_assignee(**options)
       child_task.parent = self
--- a/spec/feature/queue/colocated_task_queue_spec.rb
+++ b/spec/feature/queue/colocated_task_queue_spec.rb
@@ -27,6 +27,24 @@ RSpec.feature "ColocatedTask" do
       )
     end
 
+    it "should only create one colocated task", focus: true do
+      User.authenticate!(user: attorney_user)
+      visit("/queue/appeals/#{appeal.uuid}")
+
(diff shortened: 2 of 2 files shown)


app/models/task.rb
```diff
@@ -396,6 +396,7 @@ def assigned_to_label
   private
 
   def create_and_auto_assign_child_task(options = {})
+    # duping here
     dup.tap do |child_task|
       child_task.assigned_to = assigned_to.next_assignee(**options)
```
lomaxap: not sure what this logic is meant to do for other tasks, but `dup` [creates a new object with no ID](https://apidock.com/rails/ActiveRecord/Core/dup) which is then saved as a new record below.

Issue #11054, opened 2019-06-06
Enable TaskTable to use API requests to paginate and sort tasks


lowellrex:
This is the big "on switch" for organization queue pagination. The `TaskTable` component currently manages filtering, sorting, and pagination on the front-end. This ticket exists to move that to the back-end by way of API requests so that we can load queue table views with thousands of tasks. In order to support the pagination on the front-end (current way) at the same time as pagination on the back-end (near future for org queues) we may need to create a similar `PaginatedTaskTable` to not disrupt the way we currently do things.

In order to reduce the complexity of this ticket, let's not save any tasks in state initially. Later tickets may implement some way to store pages in state so we don't have to make an API request for every page (result sets keyed off of the combination of filters/sorts/pages?), but for now let's make an API call for all new page requests, filter changes, and re-sorting with a loading spinner when those API requests are in flight.

## 🚫 Blocked by
* [x] Allow organization queue table configuration to accept sorting (#11052)
* [x] Allow organization queue tasks endpoint to accept pagination (#11053)

We will ignore the problem of filtering/sorting with the first pass at this ticket. The first pass will include passing a new `build_from_api` (or similar) attribute of [the organization queue's response](https://github.com/department-of-veterans-affairs/caseflow/blob/4310ce79ab4d81493ff87a73521bb59fd210a913/app/controllers/organizations/tasks_controller.rb#L7) (or queue config) that signals to the front-end that we will build the table (with column headings that do not allow filtering/sorting) from API calls instead of from tasks we already have in memory.

## Acceptance criteria
* [x] `TaskTable` component's pagination, filtering, and sorting use API requests instead of happening on the front-end.
* [x] Organization queues with thousands of tasks (VLJ support staff, Hearings Management, American Legion) successfully load within timing out (5 minutes).

Filtering was spun off into a dedicated ticket (#11309).

Issue #10032, opened 2019-03-18
Consolidate individuals' queue table views


lowellrex:
We have three different ways to display an individual Caseflow user's tasks, [one for attorneys](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/AttorneyTaskListView.jsx), [one for judges](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/JudgeDecisionReviewTaskListView.jsx), and [one for everybody else](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/ColocatedTaskListView.jsx). In addition to increasing the amount of nearly redundant front-end code, this forces the [application to care about the current user's role](https://github.com/department-of-veterans-affairs/caseflow/blob/9887a895f97d976a8107bdc5d0fdb593b27b9693/client/app/queue/QueueApp.jsx#L83), and complicates the process of extending parts of the application that interact with a user's role (like [adding support for acting judges](https://github.com/department-of-veterans-affairs/caseflow/issues/10029) in Caseflow).

Attorney queue | Judge review cases view | Generic (Colocated) queue
--- | --- | ---
![image](https://user-images.githubusercontent.com/32683958/54545874-ea75c280-4978-11e9-8cc4-c99ed467afb3.png) | ![image](https://user-images.githubusercontent.com/32683958/54544961-08dabe80-4977-11e9-834d-e00b32627e3a.png) | ![image](https://user-images.githubusercontent.com/32683958/54545935-0e390880-4979-11e9-8577-7dc0af2be125.png)

This ticket exists as a reference for the more tightly-scoped work tickets that will consolidate those three views into one that can handle all three user roles (while being mostly-agnostic to the user's role on the front-end). Before migrating these three components to a single unified component for each role, we can update the existing components to reduce the impact of the switchover and discover any unexpected snags.

The three queue table views differ as follows:

. | Attorney queue | Judge review cases view | Generic (Colocated) queue
--- | --- | --- | ---
Loading error message | Displays error [if legacy appeals need DAS assigmment](https://github.com/department-of-veterans-affairs/caseflow/blob/9887a895f97d976a8107bdc5d0fdb593b27b9693/client/app/queue/AttorneyTaskListView.jsx#L47) | None | None
Columns displayed | TODO | TODO | TODO |
Navigation dropdown | TODO | TODO | TODO |
Tabs displayed | TODO | TODO | TODO |
[...]

PR #11187, opened 2019-06-24
Create new endpoint for returning a page of tasks for an organization

### LINKED ISSUE #11052: Allow organization queue table configuration to accept sorting
#11049 uses tasks from each tab's `tasks` element in the configuration returned from the back-end instead of the tasks from the global redux store. This ticket exists to enable the back-end to return those tasks ordered by their properties (the ability for the front-end to send these requests will be completed as part of a later PR). A quick sketch of one way this could look is below.

If we receive a request to `/organizations/vlj-support-staff/tasks?tab=on_hold&sort_by=case_details_link&order=desc` we would return tasks that correspond to the following ActiveRecord statement.

```ruby
Task.where(assigned_to: Colocated.singleton)
  .sort_by { |t| t.appeal.veteran_full_name } # Sorting
  .reverse # For sort order
```

If it is significantly easier to paginate without sorting and filtering then we should do that and add sorting/filtering back in later so that folks who cannot currently load their queues (VLJ support staff, Hearings branch) have at least some functionality in the short-term. If it is only a little more work to add this then we should add it in. (Spun #11307 out so that we can solve the filtering problem separately).

## Acceptance criteria
* [x] Allow organization queue tasks endpoint to returned sorted set of tasks.

## Work continuing in
* Enable back-end sorting of tasks by docket type and number #11311
* Enable back-end sorting of tasks by Veteran last name #11312
* Enable back-end sorting of tasks by regional office #11313
* Enable back-end sorting of tasks by issue count #11314
* Enable back-end sorting of tasks by appeal type #11315
* Enable back-end sorting of tasks by assignee name #11316
* Enable back-end sorting of tasks by number of days on hold #11317
* Enable back-end sorting of tasks by assigner #11318

### LINKED ISSUE #11053: Allow organization queue tasks endpoint to accept pagination
#11052 introduces the ability to filter and sort sets of tasks assigned to an organization. This ticket exists to allow us to limit those sets of tasks to a single page of a defined size and offset. The [Kaminari](https://github.com/kaminari/kaminari) and [will_paginate](https://github.com/mislav/will_paginate) gems looks promising here.

## 🚫 Blocked by
* [ ] Allow organization queue table configuration to accept sorting and filtering #11052

## Acceptance criteria
* [ ] Enable pagination for a subset of a filtered and sorted tasks.

lowellrex:
Resolves #11053. Creates a new endpoint that accepts a tab name and, optionally, a page number for returning a paged set of tasks. Does not implement filtering or sorting. Those features will be added as part of a later PR associated with #11052 (which is where I thought this PR was going at the start).

PRs from ticket #11054 will contain code to call this endpoint from the front-end.


app/controllers/organizations/tasks_controller.rb
```diff
@@ -4,16 +4,34 @@ class Organizations::TasksController < OrganizationsController
...
       tasks: json_tasks(tasks),
       id: organization.id,
       is_vso: organization.is_a?(::Representative),
-      queue_config: queue_config
+      queue_config: queue_config.to_hash_for_user(current_user)
     }
   end
 
+  # Endpoint for returning some subset of tasks.
+  def tasks
```
pkarman: this is already a private method

lowellrex: For sure for sure. This PR is still very much a work in progress and probably several days away from being ready for review.

app/controllers/organizations/task_pages_controller.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+class Organizations::TaskPagesController < OrganizationsController
```
lowellrex: Majority of this class is lifted directly from `Organizations::TasksController`.

app/controllers/organizations/task_pages_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+    tasks = TaskPage.new(
+      assignee: organization,
+      tab_name: params[:tab],
+      page: params[:page]
+    ).paged_tasks
+
+    render json: {
+      tasks: json_tasks(tasks)
+    }
+  end
```
lowellrex: Starting point of the change right here.

app/models/queue_config.rb
```diff
@@ -51,14 +51,13 @@ def include_tracking_tasks_tab?
     organization.is_a?(Representative)
   end
 
-  def tracking_tasks
-    TrackVeteranTask.active.where(assigned_to: organization)
-  end
-
   def tracking_tasks_tab
+    name = Constants.QUEUE_CONFIG.TRACKING_TASKS_TAB_NAME
+    tasks = TaskPage.new(assignee: organization, tab_name: name).tasks_for_tab
```
lowellrex: Moved the definition of each task into the new `TaskPage` class so that we only define it in one place. Open to moving those definitions out into a different file if you'd prefer, but I'd like to only define it once.

app/models/queue_config.rb
```diff
@@ -67,20 +66,19 @@ def tracking_tasks_tab
...
-
   # rubocop:disable Metrics/AbcSize
   def unassigned_tasks_tab
+    name = Constants.QUEUE_CONFIG.UNASSIGNED_TASKS_TAB_NAME
+    tasks = TaskPage.new(assignee: organization, tab_name: name).tasks_for_tab
+
     {
-      label: format(COPY::ORGANIZATIONAL_QUEUE_PAGE_UNASSIGNED_TAB_TITLE, unassigned_tasks.count),
-      name: Constants.QUEUE_CONFIG.UNASSIGNED_TASKS_TAB_NAME,
+      label: format(COPY::ORGANIZATIONAL_QUEUE_PAGE_UNASSIGNED_TAB_TITLE, tasks.count),
```
lowellrex: Since we display the number of tasks for a given tab in that tab's label I am struggling to think of a way we can consolidate the addition of the tasks to each tab's hash to some single place in the code (say to [when we serialize the tabs](https://github.com/department-of-veterans-affairs/caseflow/pull/11187/files?file-filters%5B%5D=.json&file-filters%5B%5D=.jsx&file-filters%5B%5D=.rb#diff-0e38eeb126d1e7d54ae84ff58a647e0cL20)) even though they are very similar. Open to ideas about how/where we could do that though.

app/models/task.rb
```diff
@@ -43,7 +43,8 @@ class << self; undef_method :open; end
 
   scope :closed, -> { where(status: closed_statuses) }
 
-  scope :not_tracking, -> { where.not(type: TrackVeteranTask.name) }
+  # Equivalent to .reject(&:hide_from_queue_table_view) but offloads that to the database.
+  scope :visible_in_queue_table_view, -> { where.not(type: [TrackVeteranTask.name, TimedHoldTask.name]) }
```
lowellrex: This scope was only being called [in one method](https://github.com/department-of-veterans-affairs/caseflow/pull/11187/files?file-filters%5B%5D=.json&file-filters%5B%5D=.jsx&file-filters%5B%5D=.rb#diff-d38fdc00b17f61e0a6e1bd027f26224cR85) and that method is fine to exclude `TimedHoldTask`s as well (it probably should have been doing so all along).

app/models/task_page.rb
```diff
@@ -0,0 +1,87 @@
...
+class TaskPage
+  include ActiveModel::Model
+
+  validates :tab_name, presence: true
+  validate :assignee_is_user_or_organization
+
+  attr_accessor :assignee, :tab_name, :page
+  # attr_accessor :filters, :sort_by, :sort_order
+
+  TASKS_PER_PAGE = 15
```
lowellrex: Matches [the current page size](https://github.com/department-of-veterans-affairs/caseflow/blob/4310ce79ab4d81493ff87a73521bb59fd210a913/client/app/queue/QueueTable.jsx#L276).

app/models/task_page.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  def initialize(args)
+    super
+
+    fail(Caseflow::Error::MissingRequiredProperty, message: errors.full_messages.join(", ")) unless valid?
```
lowellrex: Fail upon instantiation without having to offload validation to the calling code.

app/models/task_page.rb
```diff
@@ -0,0 +1,87 @@
...
+  # def filtered_tasks(tasks)
+  #   filters&.each do |filter_string|
+  #     filter = Rack::Utils.parse_query(filter_string)
+  #     # TODO: Fail if the filter is not in the correct format
+  #     # TODO: Fail if the column we are filtering on is not in some allowed set of columns.
+  #     tasks = tasks.where(filter["col"] => filter["val"])
+  #   end
+  #
+  #   tasks
+  # end
```
lowellrex: Left this start in the code since we will be extending the code in this direction shortly (before the end of July!). Happy to remove it if you prefer.

app/controllers/organizations/task_pages_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  private
+
+  def organization_url
+    params[:organization_url]
+  end
+
+  def json_tasks(tasks)
+    Rails.logger.debug("starting AppealRepository.eager_load_legacy_appeals_for_tasks")
```
aroltsch: I think we can remove these debug statements. We added them to debug page load times.

app/models/task_page.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+
+class TaskPage
```
pkarman: a name nit: this seems like a `TaskPager` service, not a model. I.e. the object is not an instance of a page, it's an instance of a pagination maker.

app/models/task_page.rb
```diff
@@ -0,0 +1,87 @@
...
+  TASKS_PER_PAGE = 15
+
+  def initialize(args)
+    super
+
+    fail(Caseflow::Error::MissingRequiredProperty, message: errors.full_messages.join(", ")) unless valid?
+  end
+
+  def paged_tasks
+    @page ||= 1
```
pkarman: I'd expect to see the default `page` value set in `initialize`

app/models/task_pager.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+
+class TaskPager
```
aroltsch: I would personally name `PaginateTasksService` and move it under services.

app/models/task.rb
aroltsch: Cool

app/models/task_pager.rb
```diff
@@ -0,0 +1,87 @@
...
+  def assigned_tasks
+    Task.visible_in_queue_table_view.where(assigned_to: assignee).on_hold
+  end
+
+  def recently_completed_tasks
+    Task.visible_in_queue_table_view.where(assigned_to: assignee).recently_closed
+  end
+
+  def assignee_is_user_or_organization
+    unless assignee.is_a?(User) || assignee.is_a?(Organization)
```
aroltsch: This is probably an overkill to validate that but it is okay, no harm in it.

app/models/task_pager.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  private
+
+  def tracking_tasks
+    TrackVeteranTask.active.where(assigned_to: assignee)
+  end
+
+  def unassigned_tasks
+    Task.visible_in_queue_table_view.where(assigned_to: assignee).active
```
aroltsch: So I think here we want to make sure we search for `organization` assigned_to_type.

app/models/task_pager.rb
aroltsch: Or do we want assignee to be an Organization only?

config/routes.rb
```diff
@@ -233,6 +233,7 @@
 
   resources :organizations, only: [:show], param: :url do
     resources :tasks, only: [:index], controller: 'organizations/tasks'
+    resources :task_pages, only: [:index], controller: 'organizations/task_pages'
```
aroltsch: Small nit: this name task_pages might be a little confusing. Or is it just me?

app/models/task_pager.rb
lowellrex: I think ActiveRecord knows to unwrap `assigned_to` into `...where assigned_to_id: #{id} and assigned_to_type: #{type}...` because the `assigned_to` field is polymorphic.

app/models/task_pager.rb
lowellrex: We want it to be both so that we can eventually consolidate all queue table view handling to receive their configuration from the back-end and be drawn by the same component on the front-end (#10032).

config/routes.rb
lowellrex: I'm up for renaming this to something else if you have an alternative you prefer.

lowellrex: We can rename it as part of a later PR if we want. I'm merging this into master so that we can move onto the next piece of the pagination effort that relies on this endpoint.

app/models/task_pager.rb
aroltsch: Yes it does but here `unassigned_tasks` should only be filtered by Organization, correct? If you accidentally pass assigned_to as a User, this logic won't work right.

config/routes.rb
aroltsch: How about `PaginatedTasksController`?

app/models/task_pager.rb
lowellrex: I'm pretty sure it works with a `User` as well as an `Organization`.

```ruby
rails c> assignee = Task.visible_in_queue_table_view.where(assigned_to_type: User.name).first.assigned_to
rails c> Task.visible_in_queue_table_view.where(assigned_to: assignee).active
```

aroltsch: The name might be misleading then. It says `unassigned_tasks` which means you'd like to return tasks that are not assigned to any user.

aroltsch: I mean yes I know it will work for a User type. I am talking about the business logic here

lowellrex: Definitely agree that the name is misleading. Adding a todo to change the name now.

As for the business logic, is there an example where it would fail? I'm struggling to think of one.

Issue #11154, opened 2019-06-20
Add feature tests to ensure case time line is sorted correctly. 


aroltsch:
Reference: https://github.com/department-of-veterans-affairs/caseflow/issues/10975
https://github.com/department-of-veterans-affairs/caseflow/pull/11080

PR #11197, opened 2019-06-25
Add feature tests to ensure case time line is sorted correctly

Sjones352:
Connects #11154
Screen Shot 2019-06-25 at 10 38 01 PM


spec/feature/queue/case_details_spec.rb
```diff
@@ -504,6 +504,33 @@
...
+      end
+
+      scenario "displays Pulac Cerullo task in order on  case timeline" do
+        visit "/queue/appeals/#{appeal.external_id}"
+
+        case_timeline_rows = page.find_all("table#case-timeline-table tbody tr")
+        first_row_with_task = case_timeline_rows[0]
+        second_row_with_task = case_timeline_rows[1]
+        third_row_with_task = case_timeline_rows[2]
+        expect(first_row_with_task).to have_content("Dispatch from BVA pending")
```
aroltsch: I think here we'd like to specifically test for an appeal that has a decision date that is earlier than a completed date of the Pullac Cerullo task and ensure that Pullac Cerullo task shows up at the top above the Dispatch from BVA.

Sjones352: Great explanation. Helped me understand the concept better.

client/app/queue/components/TaskRows.jsx
```diff
@@ -305,7 +305,7 @@ class TaskRows extends React.PureComponent {
           </CaseDetailsDescriptionList>
         </td>
         <td {...taskInfoWithIconTimelineContainer}
-          {...(appeal.withdrawalDate ? redWithdrawStyling : greyDotTimelineStyling)}>
+          {...(appeal.withdrawalDate || appeal.decisionDate ? timelineLeftPaddingStyle : greyDotTimelineStyling)}>
```
Sjones352: I discovered a little bug with the styling hence the change to check for decisionDate too. This is was what I saw
Screen Shot 2019-06-25 at 10 43 13 PM

PR #11200, opened 2019-06-25
Improve Print View of Hearing Worksheet

### LINKED ISSUE #8885: Hearing Prep Worksheet > "Save to PDF" // inconsistent spacing & display bugs
As a judge, I sometimes like to print the hearing prep worksheet to take paper notes, which allows me to make eye contact with the veteran during the hearing. 

I sat in on a hearing and noticed a few print display bugs, which the judge printed from "Save to PDF" in Windows / Chrome:
- The "Current Issues" tab has inconsistent / compressed spacing
- The "Hearing Summary" section reproduces a carriage return with extra spaces in the print view, which takes up extra paper (but in this case allows the judge to take handwritten notes in the margin)

screenshot 2019-01-23 10 25 15
screenshot 2019-01-23 10 46 51

### LINKED ISSUE #10291: Fix Contrast on VLJ Worksheet Print/Attorney View
## User Story
As an attorney, I need to be able to view the VLJ Hearing Worksheet in high contrast and view the issue checkboxes easily and quickly. Currently, the Attorney/Print view of the VLJ Hearing Prep worksheet displays very low contrast check boxes, so it is hard to see. 

## Acceptance Criteria
- [ ] The background color for the check boxes should be changed to #212121 for better readability. 

## Associated Tickets
See the bug report ticket here: https://github.com/department-of-veterans-affairs/appeals-support/issues/4422

### LINKED ISSUE #10314: Print Hearing Prep Worksheet to Single Page
Re-factor the VLJ Hearing Prep worksheet to print all on one page. Right now the worksheet prints on 2-3 pages depending on the number of issues. 

Original user request can be found here: https://github.com/department-of-veterans-affairs/appeals-support/issues/4837#event-2253343701

### LINKED ISSUE #10555: Optimize print so the Hearing Prep Worksheet doesn't cut off issues
User story: As a Judge, I need to be able to print the Hearing worksheets as a back up if Caseflow is unavailable so that I am prepared for a hearing if the system goes down. 

Background: Currently, the worksheets print and cut off hearing notes and issues so that it pushes to another page in an disorganized way, needs a cleaner page break. 

Example page: https://appeals.cf.ds.va.gov/hearings/644195/worksheet

AC: 
[ ] Verify that worksheet prints so that notes and issues aren't cut off, but pushed to another page as needed.

ferristseng:
Resolves #8885
Resolves #10291
Resolves #10314
Resolves #10555

### Description

  * Remove the print view for many different components related to the Hearing Worksheet, and instead, have an entirely separate layout for the printout of the Hearing Worksheet in `HearingWorksheetPrinted.jsx`
    * Most of the styling defined in `hearings_worksheet.scss` no longer applies, and there's a lot of new styling as well
  * Fix an issue in the hearing serializers where they weren't creating the `veteran_fi_last_formatted` field needed for the worksheet footer
  * **Had to remove the footer page count because it isn't working in Chrome apparently. Previously, we were setting it up to print 2 sheets each time, but now it should be compact enough to print enough on one page most of the time.**
  * Update worksheet header to use flex box for better control of element spacing.
  * Lots of refactoring and moving things around for readability and so I could reuse different functionality in the new component
  * Make checkmark blue instead of white for #10291. This check is an SVG, and there was no gray one.
  * Changes to increase the amount that can be displayed on each page
    * Decrease margins between elements
    * Decrease page margins
    * Decrease font size of most headers (had to increase the font size of some to make them more legible)
    * Use abbreviations in headers to prevent them from spilling over to multiple lines
    * Redesign the issues and procedural history sections to take up less space
    * Remove extra new lines in summary as described in #8885

### Acceptance Criteria

8885:

  - [x] The "Current Issues" tab has inconsistent / compressed spacing
  - [x] The "Hearing Summary" section reproduces a carriage return with extra spaces in the print view, which takes up extra paper (but in this case allows the judge to take handwritten notes in the margin)

10291:

  - [x]  Make check more readable

10314:

  - [x] Print on a single page

10555:

  - [x] Verify that worksheet prints so that notes and issues aren't cut off, but pushed to another page as needed.


ferristseng:
These are a bit difficult to tackle separately with the approach I'm taking!

app/assets/stylesheets/print/hearings_worksheet.scss
```diff
@@ -1,307 +1,249 @@
 @import 'uswds/src/stylesheets/lib/addons/_font-stacks.scss';
 @import 'uswds/src/stylesheets/core/_variables.scss';
 
+$cell-padding: 10px;
+$body-font-size: 14px;
+
 @media print {
-  body {-webkit-print-color-adjust: exact;}
-  h3 {
+  body {
```
ferristseng: A lot of this styling no longer applies now that there is a separate component to handle the layout of the printout

client/app/hearings/components/hearingWorksheet/HearingWorksheetStream.jsx
```diff
@@ -4,33 +4,12 @@ import { bindActionCreators } from 'redux';
 import _ from 'lodash';
 import Button from '../../../components/Button';
 import TabWindow from '../../../components/TabWindow';
+import { filterCurrentIssues, filterPriorIssues, filterIssuesOnAppeal } from '../../utils';
 import { onAddIssue } from '../../actions/hearingWorksheetActions';
 import HearingWorksheetIssues from './HearingWorksheetIssues';
 
 class HearingWorksheetStream extends Component {
 
-  filterIssuesOnAppeal = (issues, appealId) =>
```
ferristseng: Refactor these filtering methods to a utils module. Not sure if that's the best place for these to go.

client/app/hearings/components/hearingWorksheet/HearingWorksheetPrinted.jsx
```diff
@@ -0,0 +1,206 @@
...
+};
+
+export class HearingWorksheetPrinted extends React.Component {
+
+  componentDidMount() {
+    document.title = getWorksheetTitle(this.props.worksheet);
+
+    const queryString = querystring.parse(window.location.search.slice(1));
+
+    if (!queryString.do_not_open_print_prompt) {
```
sharonwarner: I don't *think* this logic is working correctly. When I clicked from the case details page to the hearing worksheet, the URL was `/hearings/21301c7b-0daf-428e-8027-869b9b2557bb/worksheet/print?keep_open=true`, but the print prompt was still displayed and the page still closed when I clicked cancel.

app/serializers/hearings/legacy_hearing_serializer.rb
```diff
@@ -55,6 +55,7 @@ class LegacyHearingSerializer
   attribute :vbms_id
   attribute :venue
   attribute :veteran_age, if: for_worksheet, &:fetch_veteran_age
+  attribute :veteran_fi_last_formatted, if: for_worksheet
```
sharonwarner: FWIW, in hearings, we've defaulted to formatting names on the frontend. I think I prefer keeping it on the frontend (we use a bunch of different name formats, and it feels sloppy to me to have to pass names between the backend and the frontend in a bunch of different formats.) Either way though, if we want to move the name formatting default to the backend, do you mind creating a ticket to update that in other places as well?

sharonwarner: Alternatively, we could commit to only using 1-2 different name formats. Maybe a short name and a long name? That would need some design input though.

client/app/hearings/components/hearingWorksheet/HearingWorksheetPrinted.jsx
ferristseng: We had a test that was using `do_not_open_print_prompt`. I'll update this and the test to use `keep_open` instead.

app/serializers/hearings/legacy_hearing_serializer.rb
ferristseng: I like keeping the name formatting on the frontend. I'll do some digging to see how we're formatting the name in other places

PR #11210, opened 2019-06-25
Allow Case Review members to edit request issues

### LINKED ISSUE #11167: Add Correct Issues link to Case Details for members of the Case Review organization
The Case Review team will be processing the withdrawal of appeals tasks for AMA appeals. We've already changed the routing of these tasks to Case Review organization queue (which has been created).

To withdraw appeals, Case Review team users navigate to the edit issues screen to withdraw issues. When all issues are withdrawn, the appeal is considered withdrawn.

This issue will allow the Case Review team to more easily navigate to the Edit Issues screen.

## AC
- Add the `Correct issues` link to Case Details 
- Ensure Case Review team members can view and click that link, and navigate to Edit Issues
- When users finish editing issues, they are sent back to the Case Details screen

monfresh:
Resolves #11167

- Moved the logic for determining whether or not the user can edit
request issues from the User class to a policy class since this is
only used by the serializer.
- Added a conditional to check if the user is a member of the
Case Review team
- Added more tests to cover all scenarios

### DIFF
--- a/app/policies/appeal_request_issues_policy.rb
+++ b/app/policies/appeal_request_issues_policy.rb
@@ -7,20 +7,28 @@ class AppealRequestIssuesPolicy
   end
 
   def editable?
-    return true if case_review_team_member?
+    return true if editable_by_case_review_team_member?
 
-    appeal_has_in_progress_or_assigned_judge_or_attorney_tasks_assigned_to_user?
+    case_is_in_active_review_by_current_user?
   end
 
   private
 
   attr_reader :user, :appeal
 
-  def case_review_team_member?
+  def editable_by_case_review_team_member?
+    current_user_is_case_review_team_member? && case_is_not_in_active_review?
+  end
+
+  def current_user_is_case_review_team_member?
     BvaIntake.singleton.user_has_access?(user)
   end
 
-  def appeal_has_in_progress_or_assigned_judge_or_attorney_tasks_assigned_to_user?
+  def case_is_not_in_active_review?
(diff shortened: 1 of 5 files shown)


app/policies/appeal_request_issues_policy.rb
```diff
@@ -0,0 +1,30 @@
...
+class AppealRequestIssuesPolicy
+  def initialize(user:, appeal:)
+    @user = user
+    @appeal = appeal
+  end
+
+  def editable?
+    return true if case_review_team_member?
+
+    appeal_has_in_progress_or_assigned_judge_or_attorney_tasks_assigned_to_user?
```
aroltsch: woah

app/policies/appeal_request_issues_policy.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  private
+
+  attr_reader :user, :appeal
+
+  def case_review_team_member?
+    BvaIntake.singleton.user_has_access?(user)
+  end
+
+  def appeal_has_in_progress_or_assigned_judge_or_attorney_tasks_assigned_to_user?
```
aroltsch: Better name would be `case_is_in_active_review_by_attorney_or_judge?`

app/policies/appeal_request_issues_policy.rb
monfresh: hehe

app/policies/appeal_request_issues_policy.rb
monfresh: I like that

Issue #10804, opened 2019-05-16
Change text "Please contact the Caseflow Team"


Sjones352:
Since we have disabled the feedback button and Caseflow support phone, we need to change the text in our code that states `Please contact the Caseflow Team` / `please contact Caseflow support` because it very confusing to our users.

pkarman:
@laurjpeterson we should think about how we direct users to contact us now that we have changed to YourIT.

laurjpeterson:
Agree. Can you remind me where this text shows up in the application?

@msj0nes - I'll defer to you on language here. Thinking it could be "Please contact YourIT for support?"

pkarman:
https://github.com/department-of-veterans-affairs/caseflow/search?q=%22Please+contact%22&unscoped_q=%22Please+contact%22

msj0nes:
@laurjpeterson and @pkarman
>Please contact the Caseflow team via the VA Enterprise Service Desk at 855-673-4357 or by creating a ticket via YourIT, if experiencing technical difficulties.

Or something along those lines. If the word YourIT could be hyperlinked to **_https://yourit.va.gov/va_** that would be great.

pkarman:
Include the "Issues not Editable" error message for intake.

pkarman:
done merged and deployed.

PR #11212, opened 2019-06-25
Update support contact to point at YourIT

pkarman:
connects #10804 

Screen Shot 2019-06-25 at 3 21 30 PM


config/routes.rb
```diff
@@ -272,6 +272,8 @@
 
   # :nocov:
   namespace :test do
+    get "/error", to: "users#show_error"
```
pkarman: just to make it easier to visually test locally

pkarman:
@leikkisa agreed about the COPY.json file -- I will leave that for another PR.

pkarman:
it does not appear that links are used (possible?) in the COPY.json file.

leikkisa:
Yeah, I use a "_START" and "_END" suffix in COPY for links but it is a little awkward, and I generally don't feel strongly about using the COPY file. Some of them don't seem to have links too so maybe just for those.

PR #11213, opened 2019-06-25
Enable org queues to use task pagination API

lowellrex:
Connects #11054. Enables back-end pagination for certain organizations (VSOs, VLJ support staff, hearing management branch) when the current user has the `use_task_pages_api` feature toggle on. Still needs some cosmetic love, but it's far enough along that I'd like some feedback on how this approach can be improved.

caching-pagination


app/controllers/organizations/tasks_controller.rb
```diff
@@ -17,6 +17,8 @@ def index
   private
 
   def tasks
+    return [] if queue_config[:use_task_pages_api]
```
lowellrex: If we are using the back-end pagination API then we should not bother with returning the tasks here since we will use the ones from the queue config for the first page and the API for all subsequent pages.

app/models/organization.rb
```diff
@@ -24,6 +24,10 @@ def show_regional_office_in_queue?
     false
   end
 
+  def use_task_pages_api?
+    false
+  end
```
lowellrex: By default organizations will continue using front-end pagination.

app/models/organizations/hearings_management.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class HearingsManagement < Organization
+  def self.singleton
+    HearingsManagement.first || HearingsManagement.create(name: "Hearings Management", url: "hearings-management")
+  end
+
```
lowellrex: Moving the `singleton` definition to the top of the class definition so that it matches all of the other organizations.

app/models/queue_config.rb
```diff
@@ -17,17 +17,22 @@ def initialize(args)
...
+      tasks_per_page: TaskPager::TASKS_PER_PAGE,
+      use_task_pages_api: use_task_pages_api?,
+      tabs: tabs.map { |tab| attach_tasks_to_tab(tab, user) }
     }
   end
 
   private
 
+  def use_task_pages_api?
+    # return true
```
lowellrex: My manual toggle for easy testing locally. Will remove before this is merged.

app/models/queue_config.rb
```diff
@@ -37,6 +42,22 @@ def tabs
     ].compact
   end
 
+  def attach_tasks_to_tab(tab, user)
+    task_pager = TaskPager.new(assignee: organization, tab_name: tab[:name])
+
+    # Only return tasks in the configuration if we are using it to populate the first page of results.
+    # Otherwise avoid the overhead of the additional database requests.
+    tasks = use_task_pages_api? ? serialized_tasks_for_user(task_pager.paged_tasks, user) : []
```
lowellrex: Inverse of [the optimization above](https://github.com/department-of-veterans-affairs/caseflow/pull/11213/files#r297409795). Don't make additional requests to the database if we are just going to be using the tasks we get from some other method.

app/models/queue_config.rb
```diff
@@ -54,33 +75,24 @@ def include_tracking_tasks_tab?
...
-  # rubocop:disable Metrics/AbcSize
   def unassigned_tasks_tab
-    name = Constants.QUEUE_CONFIG.UNASSIGNED_TASKS_TAB_NAME
-    tasks = TaskPager.new(assignee: organization, tab_name: name).tasks_for_tab
-
     {
-      label: format(COPY::ORGANIZATIONAL_QUEUE_PAGE_UNASSIGNED_TAB_TITLE, tasks.count),
-      name: name,
+      label: COPY::ORGANIZATIONAL_QUEUE_PAGE_UNASSIGNED_TAB_TITLE,
+      name: Constants.QUEUE_CONFIG.UNASSIGNED_TASKS_TAB_NAME,
```
lowellrex: Was able to consolidate all this duplicated logic in [the new `attach_tasks_to_tab ` method](https://github.com/department-of-veterans-affairs/caseflow/pull/11213/files#diff-0e38eeb126d1e7d54ae84ff58a647e0cR45).

client/app/components/TablePagination.jsx
```diff
@@ -36,33 +36,28 @@ class TablePagination extends React.PureComponent {
 
   render() {
     const {
-      paginatedData,
+      casesPerPage,
       currentPage,
+      numberOfPages,
```
lowellrex: Pass `numberOfPages` as an argument instead of computing it from the `paginatedData` argument.

client/app/components/TablePagination.jsx
```diff
@@ -36,33 +36,28 @@ class TablePagination extends React.PureComponent {
...
-    // the amount of data in the current page.
-    const endingCaseNumber = numberOfPages > 0 && paginatedData[currentPage] ?
-      previousCaseCount + paginatedData[currentPage].length :
-      0;
-    // Create the range
-    let currentCaseRange = `${beginningCaseNumber}-${endingCaseNumber}`;
-    // Create the entire summary
-    const paginationSummary = `Viewing ${currentCaseRange} of ${totalCasesCount} total cases`;
+
+    const paginationSummary = `Viewing ${beginningCaseNumber}-${endingCaseNumber} of ${totalCasesCount} total cases`;
```
lowellrex: No significant change here other than we were able to replace the `paginatedData` argument (which relied on an array of array of tasks to determine this summary) with the `casesPerPage` argument (which is a number).

client/app/queue/OrganizationQueue.jsx
```diff
@@ -86,19 +87,25 @@ class OrganizationQueue extends React.PureComponent {
...
     return {
-      label,
+      label: tabConfig.label,
       page: <React.Fragment>
-        <p className="cf-margin-top-0">{description}</p>
+        <p className="cf-margin-top-0">{tabConfig.description}</p>
         { tabConfig.allow_bulk_assign && <BulkAssignButton /> }
         <TaskTable
-          customColumns={cols}
+          key={tabConfig.name}
```
lowellrex: Need to set a `key` on each `TaskTable` so that they can have their own state (React ended up sharing state across the sibling components without this key).

client/app/queue/OrganizationQueue.jsx
```diff
@@ -86,19 +87,25 @@ class OrganizationQueue extends React.PureComponent {
...
+          customColumns={this.columnsFromConfig(tabConfig)}
           tasks={tasks}
+          useTaskPagesApi={config.use_task_pages_api}
+          tasksPerPage={config.tasks_per_page}
+          numberOfPages={tabConfig.task_page_count}
+          totalTaskCount={tabConfig.total_task_count}
+          taskPagesApiEndpoint={tabConfig.task_page_endpoint_base_path}
```
lowellrex: Since we're adding 5 new arguments here maybe it makes sense to just make this a new component? What're y'all's thoughts?

client/app/queue/QueueTable.jsx
```diff
@@ -32,26 +37,15 @@ import { COLORS } from '../constants/AppConstants';
...
-  ease: 'outCube',
-  offset: -35
-}));
-
-const focusElement = (el = this) => {
-  if (el.tabIndex <= 0) {
-    el.setAttribute('tabindex', '-1');
-  }
-  el.focus();
-};
```
lowellrex: I got sick of the browser taking control of the scroll when I was testing this. Happy to put it back if we want, but I really hate this.

client/app/queue/QueueTable.jsx
```diff
@@ -285,11 +282,32 @@ export default class QueueTable extends React.PureComponent {
 
   updateCurrentPage = (newPage) => {
     this.setState({ currentPage: newPage });
-
-    scrollTo(this);
-    focusElement(this.elementForFocus);
+    this.requestNewPage(newPage);
   }
 
+  requestNewPage = (newPage) => {
```
lowellrex: This is the heart of the new behaviour. We make an API request if we are using the back-end pagination API.

lowellrex: This method seems more than a little hacky to me. Very open to ideas for how to improve it (Maybe it should be injected from `OrganizationQueue`? How would we set local state? Maybe it should just be a separate component?).

client/app/queue/utils.js
```diff
@@ -52,79 +52,85 @@ export const prepareMostRecentlyHeldHearingForStore = (appealId, hearing) => {
   };
 };
 
+const taskAttributesFromRawTask = (task) => {
```
lowellrex: Separating these functions into two steps each so we can get arrays for our task pages instead of the hashes that are sent to the global redux store.

app/models/queue_config.rb
```diff
@@ -54,33 +75,24 @@ def include_tracking_tasks_tab?
...
-      name: name,
+      name: Constants.QUEUE_CONFIG.TRACKING_TASKS_TAB_NAME,
       description: format(COPY::ALL_CASES_QUEUE_TABLE_TAB_DESCRIPTION, organization.name),
       columns: [
         Constants.QUEUE_CONFIG.CASE_DETAILS_LINK_COLUMN,
         Constants.QUEUE_CONFIG.ISSUE_COUNT_COLUMN,
         Constants.QUEUE_CONFIG.APPEAL_TYPE_COLUMN,
         Constants.QUEUE_CONFIG.DOCKET_NUMBER_COLUMN
       ],
-      task_group: Constants.QUEUE_CONFIG.TRACKING_TASKS_GROUP,
```
lowellrex: We never ended up using the `task_group` element.

app/models/organizations/hearings_management.rb
kevmo: `HearingAdmin` team could use same change.

app/models/organization.rb
pkarman: could we make this use the FeatureToggle rather than hardcoded bool?

`FeatureToggle.enabled?(:use_task_pages_api)`

app/models/organizations/colocated.rb
```diff
@@ -8,4 +8,8 @@ def self.singleton
   def next_assignee(options = {})
     ColocatedTaskDistributor.new.next_assignee(options)
   end
+
+  def use_task_pages_api?
+    true
```
pkarman: same thing here about feature toggle.

app/models/organizations/hearings_management.rb
```diff
@@ -9,7 +13,7 @@ def show_regional_office_in_queue?
     true
   end
 
-  def self.singleton
-    HearingsManagement.first || HearingsManagement.create(name: "Hearings Management", url: "hearings-management")
+  def use_task_pages_api?
+    true
```
pkarman: feature toggle?

app/models/organizations/vso.rb
```diff
@@ -1,3 +1,7 @@
 # frozen_string_literal: true
 
-class Vso < Representative; end
+class Vso < Representative
+  def use_task_pages_api?
+    true
```
pkarman: feature toggle?

app/models/queue_config.rb
```diff
@@ -17,17 +17,22 @@ def initialize(args)
...
+      use_task_pages_api: use_task_pages_api?,
+      tabs: tabs.map { |tab| attach_tasks_to_tab(tab, user) }
     }
   end
 
   private
 
+  def use_task_pages_api?
+    # return true
+    FeatureToggle.enabled?(:use_task_pages_api) && organization.use_task_pages_api?
```
pkarman: +1

app/models/queue_config.rb
```diff
@@ -37,6 +42,22 @@ def tabs
...
   end
 
+  def attach_tasks_to_tab(tab, user)
+    task_pager = TaskPager.new(assignee: organization, tab_name: tab[:name])
+
+    # Only return tasks in the configuration if we are using it to populate the first page of results.
+    # Otherwise avoid the overhead of the additional database requests.
+    tasks = use_task_pages_api? ? serialized_tasks_for_user(task_pager.paged_tasks, user) : []
+
+    tab.merge(
```
pkarman: nice!

app/models/task_pager.rb
```diff
@@ -46,14 +46,22 @@ def paged_tasks
   #   tasks
   # end
 
+  def task_page_count
```
pkarman: does the kaminari `total_pages` method work on `tasks_for_tab` here?

client/app/components/TablePagination.jsx
pkarman: Not sure if you looked at the Pagination component? I did something similar there for the /jobs pagination.

client/app/queue/OrganizationQueue.jsx
pkarman: a `PaginatedTaskTable` makes sense to me. I think the nonComp feature still uses the `TaskTable` so otherwise we'd want to make sure we don't introduce regressions.

client/app/queue/QueueTable.jsx
pkarman: hate is a good motivator.

client/app/queue/QueueTable.jsx
pkarman: This was the part I struggled with on the /jobs paginator too. I am not sure there is an elegant solution. In my case, I wasn't trying to adhere to SPA pattern so just reloaded the whole page.

https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/asyncableJobs/pages/JobsPage.jsx#L210

This looks readable to me, which is what I would care about. The only gotchas are the off-by-one with index and page counts. You may also want to guard against clicking on the current page (not sure if that affects here as it did for me on /jobs).

client/app/queue/OrganizationQueue.jsx
kevmo: If we believe we will ever want to use the `PaginatedTaskTable` in other views, I have a slight preference for creating it in its own file in `client/app/queue/components`.

client/app/queue/QueueTable.jsx
kevmo: I agree that this is readable, which is most important. The comments in the code are good.  I think this general approach is definitely fine for now.

app/models/organization.rb
lowellrex: FeatureToggle's are [enabled on a per-user basis](https://github.com/department-of-veterans-affairs/caseflow-commons/blob/8dde00d67b7c629e4b871f8dcb3617bfe989b3db/app/services/feature_toggle.rb#L15) and this flag allows us to enable it on a per-organization basis which gives us another level of control as we roll this out. When this code is release live nothing should change. No requests will use the task pages API. However, we can roll this out in a phased fashion as follows:

1. We can enable the feature toggle for a few members of teams with this flag set to true (VSOs, VLJ support, Hearings management) so they can test that it works and doesn't have any unexpected bugs without affecting how the rest of that team is using the task queue.
2. After individual folks have confirmed that the task pages API is working well we can roll it out to a whole team that needs the task pages API (we'd actually need to move the `enabled?` function call inside the organization's `use_task_pages_api?` function definition) without affecting any of the teams who are not experiencing team queue loading problems.
3. After we've caught some bugs with heavy full-team use we can roll it out to all organization queues (by removing this function entirely and making it the default).

This function at the organization enables phase 2 which I think is fairly important. Team queues work for most teams right now, so I'd love to limit the pagination changes to just those teams that are experiencing team queue loading problems before rolling the changes out to everybody.

app/models/organizations/hearings_management.rb
lowellrex: Totally agree. Going to postpone rearranging that class to a later PR so that this changeset stays small.

client/app/components/TablePagination.jsx
lowellrex: It's fairly easy to replace the only place we use `TablePagination` with the `Pagination` component so I'm going to go ahead and do that instead.

client/app/queue/QueueTable.jsx
```diff
@@ -286,6 +286,15 @@ export default class QueueTable extends React.PureComponent {
     return paginatedData;
   }
 
+  componentDidMount = () => {
+    this.setState({
+      tasksFromApiForPage: {
```
pkarman: as I'm reading this, the idea seems to be cache each paged task set on the browser, rather than re-fetching on each page click?

I can see the attraction of that for performance reasons. My only concern would be once sorting/filtering are enabled, that the cache invalidation will be tricky. Also, if someone updates a task in a different browser, that the user in this browser won't see the latest when they page "back" to that task set.

lowellrex: Correct. I think we'll have to handle caching some other way when we do sorting/filtering.

And you're totally right about the different browser window/tab causing cached tasks to be in the incorrect state (though we already have this now with entirely front-end pagination).

Happy to back out the caching of paged tasks if you think that's better.

pkarman: I think the caching falls under the category of premature optimization. Let's back it out. In theory the pagination itself should make things perform well enough that we don't need caching -- time will tell.

PR #11215, opened 2019-06-25
Close EP400 request issues

leikkisa:
We have confirmation from AMO that if an end product gets its claim type changed to an EP400, that we can close the request issues with no_decision.  Currently we are doing this manually case-by-case, so that the request issues can be intaken again on a new decision review.
https://dsva.slack.com/archives/CFBLH2LR4/p1561396691147800

I have started looking into doing this for EPs that are already cleared. Since we don't store the claim_type_code if it gets changed, it will be good to prevent new incoming ones so that we don't have to search for new ones, which requires fetching the EPE result from BGS and takes some time.

Note, we do establish sometimes establish 400 claims from Caseflow as well, but those do not have any request issues.

### DIFF
--- a/app/models/end_product_establishment.rb
+++ b/app/models/end_product_establishment.rb
@@ -201,6 +201,7 @@ class EndProductEstablishment < ApplicationRecord
       )
       sync_source!
       close_request_issues_if_canceled!
+      close_request_issues_if_converted_to_ep400!
     end
   rescue EstablishedEndProductNotFound, AppealRepository::AppealNotValidToReopen => error
     raise error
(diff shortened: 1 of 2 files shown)


app/models/end_product_establishment.rb
```diff
@@ -375,6 +376,13 @@ def close_request_issues_if_canceled!
     request_issues.all.find_each(&:close_after_end_product_canceled!)
   end
 
+  def close_request_issues_with_no_decision!
+    return unless status_cleared?
+    return unless result.claim_type_code.include?("400")
```
pkarman: could this ever generate false positives? should we use `claim_type_code =~ /^400/` to be more specific?

leikkisa: Yeah, so far all of the EP400s I've seen have it in the first three characters, so I think this is a good call in case other codes use that number in another way.

PR #11216, opened 2019-06-25
Lighthouse api docs controller

### LINKED ISSUE #11202: Proposed API endpoint for submitting Higher-Level Reviews
As part of the Lighthouse Appeals API project, we're adding an external API endpoint for submitting a Higher-Level Review (eventually we'll do the same for Supplemental Claims and Board Appeals). The goal is to have this endpoint be a single step--sending all the info needed as JSON in a single post--unlike the intake UI (multi-step with validation and returned populated data between steps). After having gone through the relevant controllers and models for Higher-Level Reviews, here is my proposed openapi spec (linked below) for this endpoint (we're currently still figuring out the URL namespacing, and  directory structure). An easy way to get a nicely formatted view of this spec is to paste it in to editor.swagger.io

Would love to get some feedback on this from all of you intake experts! --things we're missing/not thinking of, stuff we should look out for, misconceptions/bad assumptions we're making --really, anything we should consider. We're certain that we'll need to make tweaks to this spec with your input, and also as we just get more familiar with the codebase (and all the involved services) in general.

The standard for VA.gov APIs is for them to be documented with openapi (swagger), and that the shape of the JSON return should conform to JSON:API ( https://jsonapi.org/ )--which is why this spec isn't more 1-to-1 with the internal API being used by the intake form (and why it's a bit verbose (with types, etc)). A few examples:

* instead of naming the field `veteran_file_number` or `claimant` (or `participantID`) these are the `id` fields of the respective veteran and claimant objects (this fits better with JSON:API)
* no explicit `veteran_is_not_claimant` field - in the proposed spec, a claimant is an optional relationship to specify, and, if it is absent, the veteran is the claimant.

Thank you in advance for taking a look!

(A PR will be made shortly, and will be linked)

[decision_reviews.yaml.txt](https://github.com/department-of-veterans-affairs/caseflow/files/3326347/decision_reviews.yaml.txt)

### LINKED ISSUE #11214: API docs endpoint
Part of the Lighthouse API project is sharing OpenAPI specs for endpoints.  As we develop new API endpoints in caseflow, we want to have endpoints for serving up that documentation.

The new Lighthouse API work will be in the controllers/api directory.  The docs will be served from a new controller in a docs directory.  That directory will also contain open api specs to describe the APIs.

erluti:
Resolves #11214
Resolves #11202

### Description
Adding a docs_controller for serving OpenAPI specs to describe the API expansion for lighthouse project.
Included the first spec for a higher_level_review endpoint to be added later.

### Acceptance Criteria
- [ ] Code compiles correctly
- [ ] OpenAPI Spec accessible through docs/v3 route

### Testing Plan
1. Added request specs to describe expectations for OpenAPI spec responses.


edmangimelli:
LGTM

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -0,0 +1,397 @@
...
+                      type:
+                        type: string
+                        enum: [request_issue]
+                      attributes:
+                        type: object
+                        required:
+                          - decision_text
+                          - decision_date
+                          - ineligible_due_to_id
+                          - ineligible_reason
```
pkarman: note that the `ineligible_*` fields are typically set by the server side. I would not expect a client to send them.

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -0,0 +1,397 @@
...
+                            type: string
+                          ineligible_reason:
+                            type: string
+                          withdrawal_date:
+                            type: string
+                            format: date
+                          nonrating_issue_category:
+                            $ref: "#/components/schemas/nonrating_issue_category"
+      responses:
+        '200':
```
pkarman: you might consider `202` for success instead, just to emphasize that the actual establishment is async and will happen eventually.

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -0,0 +1,397 @@
...
+          description: Created
+        '500':
```
pkarman: probably should include a `400` here as well as that will indicate the request itself was invalid.

app/controllers/api/docs/v3/docs_controller.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class Api::Docs::V3::DocsController < ActionController::Base
+  protect_from_forgery with: :null_session
```
pkarman: if you inherit from `Api::ApplicationController` instead you might not need `protect_from_forgery`

config/routes.rb
```diff
@@ -29,6 +29,11 @@
     namespace :v2 do
       resources :appeals, only: :index
     end
+    namespace :docs do
```
pkarman: what would you think about `/api/v2/docs` as the namespace instead? I'm not sure we want to increment the version here since we're talking about a different endpoint space entirely (you're not doing a new version of `appeals` e.g.).

spec/requests/api/docs/v3/docs_controller_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "rails_helper"
```
pkarman: should not need that first line -- `rails_helper` is automatically required.

spec/requests/api/docs/v3/docs_controller_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+RSpec.describe Api::Docs::V3::DocsController, type: :request do
```
pkarman: I think the leading `RSpec` is unnecessary.

spec/requests/api/docs/v3/docs_controller_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+      get '/api/docs/v3/decision_reviews'
+      expect(response).to have_http_status(200)
+      json = JSON.parse(response.body)
+      expect(json["openapi"]).to eq('3.0.0')
+    end
+    describe '/higher_level_review documentation' do
+      before(:each) do
+        get '/api/docs/v3/decision_reviews'
+      end
+      let(:hlr_doc){
```
pkarman: I would expect code climate to suggest that multi-line blocks should use `do ... end` rather than braces.

app/controllers/api/docs/v3/docs_controller.rb
erluti: I wanted to start our own v3 Api controller.  When we add another endpoint i was thinking we'd start combining the shared logic between them.

config/routes.rb
erluti: this is the pattern other lighthouse projects use.  So when versions increment, the documents are found in the `api/docs/vN` path.  Our team discussed this a little bit, so I can go back to the other lighthouse teams to find out more about the standard

spec/requests/api/docs/v3/docs_controller_spec.rb
erluti: removed!

spec/requests/api/docs/v3/docs_controller_spec.rb
erluti: i don't know where I picked up the habit of doing let blocks in braces, but i fixed this!

config/routes.rb
erluti: Oh, as far as going to v3 goes, I was think we'd start all of our new work at v3.  The API we started a spec for was "decision_reviews", so I was originally thinking it'd be v0 but I understood our conversation earlier to be that we should start at v3.  I'm not very concerned with where we start, I just thought that's what we landed on in that meeting.

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -0,0 +1,397 @@
+openapi: 3.0.0
+info:
+  title: Decision Reviews
+  description: API version 0 for Decision Reviews provided by the Lighthouse Project
+  version: 3.0.0
+servers:
+  - url: /lighthouse_api/{version}
```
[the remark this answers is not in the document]
edmangimelli: fixed it

app/controllers/api/docs/v3/decision_reviews.yaml
edmangimelli: whoops. fixed

edmangimelli:
tickets from zenhub
Banana Peels team (Lighthouse Appeals API team)

https://app.zenhub.com/workspaces/developervagov-5ac272fff41bfa50c0e46c48/issues/department-of-veterans-affairs/vets-contrib/2356

https://app.zenhub.com/workspaces/developervagov-5ac272fff41bfa50c0e46c48/issues/department-of-veterans-affairs/vets-contrib/2359

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -0,0 +1,393 @@
...
+                    properties:
+                      type:
+                        type: string
+                        enum: [request_issue]
+                      attributes:
+                        type: object
+                        required:
+                          - decision_text
+                          - decision_date
+                          - withdrawal_date
```
pkarman: `withdrawal_date` is not required. neither is `nonrating_issue_category` unless the issue is a non-rating issue, in which case it must also have a description.

config/routes.rb
pkarman: ok namespaces as you've proposed here will work.

PR #11229, opened 2019-06-27
Refactor logic around BGS EP modifier

aroltsch:
Before I add 930 EP correction modifiers, I'd like to refactor the open modifier logic


app/models/end_product_establishment.rb
```diff
@@ -92,8 +85,6 @@ def create_contentions!
     records_ready_for_contentions = calculate_records_ready_for_contentions
     return if records_ready_for_contentions.empty?
 
-    set_establishment_values_from_source
```
aroltsch: I don't think we need to set modifiers when creating a contention. Peter or Sally if you could confirm.

app/models/ramp_review.rb
```diff
@@ -124,7 +124,6 @@ def new_end_product_establishment
       code: end_product_code,
       payee_code: payee_code,
       claimant_participant_id: claimant_participant_id,
-      valid_modifiers: valid_modifiers,
```
aroltsch: It seems as we don't need to set valid_modifiers here because calling `end_product_establishment.source.valid_modifiers` in `FindOpenEndProductModifier` should work correctly

app/models/end_product_establishment.rb
pkarman: we need modifier for creating an EP, but not a contention.

app/models/ramp_review.rb
pkarman: We might be done with RAMP anyway. Let me check on that.

app/workflows/find_open_end_product_modifier.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+# An EP modifier is a BGS requirement - it helps delineate EPs with the same claim_type_code
+# from one another. e.g. if you have multiple HLRs.
+class FindOpenEndProductModifier
```
pkarman: a naming nit: how about `EndProductModifierFinder` or just `EndProductModifier` ? IMO class names should be nouns.

aroltsch: EndProductModifierFinder sounds great to me!

PR #11236, opened 2019-06-27
Judge Task Reassigns

### LINKED ISSUE #11120: Allow Judges to Reassign JudgeTasks to other Judges
Judges have a missing ability that came out of this discussion in #10551

  - ~reassign any case assigned to another judge _to themselves_~                                                             
  - reassign any _of their cases_ to another judge                                                                      

Attorneys can already select which Judge to return a decision review to, so it only makes sense to empower the judge with the same option. Additionally this would help with covering any Judge periods of leave.

This would, at least initially, be limited to cases where the specific Judge is one of the assignees.

lomky:
Resolves #11120 

### Description
Judges now have the ability to reassign their judge tasks to another user

### Remaining Work
 - [x] Check assigning notes are available to the newly assigned Judge
 - [x] Tests


app/models/appeal.rb
```diff
@@ -111,7 +111,7 @@ def latest_attorney_case_review
   end
 
   def reviewing_judge_name
-    task = tasks.order(created_at: :desc).detect { |t| t.is_a?(JudgeTask) }
+    task = tasks.where.not(status: "cancelled").where(type: "JudgeDecisionReviewTask").order(created_at: :desc).first
```
lowellrex: Let's use `Constants.TASK_STATUSES.cancelled` and `JudgeDecisionReviewTask.name` here so that the application will fail at compile time (in the second example) if we ever decide to change the strings to which those variables refer instead of silently returning no tasks.

app/models/concerns/taskable.rb
```diff
@@ -4,10 +4,10 @@ module Taskable
   extend ActiveSupport::Concern
 
   def assigned_attorney
-    tasks.includes(:assigned_to).detect { |t| t.is_a?(AttorneyTask) }.try(:assigned_to)
+    tasks.where.not(status: "cancelled").order(created_at: :desc).includes(:assigned_to).detect { |t| t.is_a?(AttorneyTask) }.try(:assigned_to)
   end
 
   def assigned_judge
-    tasks.includes(:assigned_to).detect { |t| t.is_a?(JudgeTask) }.try(:assigned_to)
+    tasks.where.not(status: "cancelled").order(created_at: :desc).includes(:assigned_to).detect { |t| t.is_a?(JudgeTask) }.try(:assigned_to)
```
lowellrex: If we're looking for non-cancelled tasks often enough it might make sense to make that operation into a scope on the `Task` class. Up to you.

spec/feature/queue/ama_queue_spec.rb
```diff
@@ -715,5 +720,36 @@ def valid_document_id
...
+        visit "/queue"
+        expect(page).to have_content(format(COPY::JUDGE_CASE_REVIEW_TABLE_TITLE, "0"))
+
+        find(".cf-dropdown-trigger", text: COPY::CASE_LIST_TABLE_QUEUE_DROPDOWN_LABEL).click
+        expect(page).to have_content(COPY::JUDGE_ASSIGN_DROPDOWN_LINK_LABEL)
+        click_on COPY::JUDGE_ASSIGN_DROPDOWN_LINK_LABEL
+
+        click_on veteran_full_name
+      end
+    end
```
lowellrex: Can we add a test to confirm that this works for decision review tasks as well?

app/models/task.rb
```diff
@@ -43,6 +43,8 @@ class << self; undef_method :open; end
 
   scope :closed, -> { where(status: closed_statuses) }
 
+  scope :not_cancelled, -> { where(status: not_cancelled_statuses) }
```
lowellrex: Why this approach instead of `where.not(status: Constants.TASK_STATUSE.cancelled)`?

lomky: I got caught up following the existing scheme the other scopes were using.

PR #11239, opened 2019-06-28
Re-enable table filter names for org queues

lowellrex:
The move from using the `TaskTable` component to using the `QueueTable` component in PR #11213 left behind our friendly names that we display when we are filtering columns in that table. This PR adds those back in.

Before | After
--- | ---
image | image

There is probably a longer-term fix here to just use the column's `header` attribute [as defined in `TaskTable`](https://github.com/department-of-veterans-affairs/caseflow/blob/0083faa2ce9995302ce2d7717753b6cc26f121ce/client/app/queue/components/TaskTable.jsx#L43), but a quick glance at the code makes it appear as if that will be a fairly involved change.

### DIFF
--- a/client/app/queue/OrganizationQueue.jsx
+++ b/client/app/queue/OrganizationQueue.jsx
@@ -21,10 +21,7 @@ import {
 } from './selectors';
 import { tasksWithAppealsFromRawTasks } from './utils';
 import { clearCaseSelectSearch } from '../reader/CaseSelect/CaseSelectActions';
-import {
-  COLUMN_NAMES,
-  fullWidth
-} from './constants';
+import { fullWidth } from './constants';
 import QUEUE_CONFIG from '../../constants/QUEUE_CONFIG.json';
 
 const containerStyles = css({
@@ -112,7 +109,6 @@ class OrganizationQueue extends React.PureComponent {
           numberOfPages={tabConfig.task_page_count}
           totalTaskCount={tabConfig.total_task_count}
           taskPagesApiEndpoint={tabConfig.task_page_endpoint_base_path}
-          alternateColumnNames={COLUMN_NAMES}
           enablePagination
         />
       </React.Fragment>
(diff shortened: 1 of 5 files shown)


client/app/queue/OrganizationQueue.jsx
```diff
@@ -109,6 +112,7 @@ class OrganizationQueue extends React.PureComponent {
           numberOfPages={tabConfig.task_page_count}
           totalTaskCount={tabConfig.total_task_count}
           taskPagesApiEndpoint={tabConfig.task_page_endpoint_base_path}
+          alternateColumnNames={COLUMN_NAMES}
```
lowellrex: Heart of the change here.

client/app/queue/constants.js
```diff
@@ -170,7 +170,8 @@ export const COLOCATED_HOLD_DURATIONS = [15, 30, 45, 60, 90, 120, CUSTOM_HOLD_DU
 
 export const COLUMN_NAMES = {
   'appeal.caseType': 'Case Type',
-  'appeal.docketName': 'Docket Number',
+  'appeal.docketName': COPY.CASE_LIST_TABLE_DOCKET_NUMBER_COLUMN_TITLE,
+  'assignedTo.name': COPY.CASE_LIST_TABLE_APPEAL_LOCATION_COLUMN_TITLE,
```
lowellrex: Adds this alternative column name that was missing before.

client/app/queue/constants.js
```diff
@@ -170,7 +170,8 @@ export const COLOCATED_HOLD_DURATIONS = [15, 30, 45, 60, 90, 120, CUSTOM_HOLD_DU
 
 export const COLUMN_NAMES = {
   'appeal.caseType': 'Case Type',
-  'appeal.docketName': 'Docket Number',
+  'appeal.docketName': COPY.CASE_LIST_TABLE_DOCKET_NUMBER_COLUMN_TITLE,
```
lowellrex: Changes the text here from "Docket Number" to "Docket" since that is really what we are filtering on (legacy, hearing, evidence, direct).

hschallhorn:
Looks like both (and only) places where we implement `QueueTable` (`TaskTable` and now `OrganizationQueue`) are passing `COLUMN_NAMES` as `alternateColumnNames` which is only used by `FilterSummary`. Is there any reason we wouldn't want to just import the alternate column names in `FilterSummary` rather than passing it through multiple components?

Issue #11139, opened 2019-06-19
Cache SSN on veterans table


pkarman:
Similar to how we cache veteran name, cache the SSN on the veterans table.

This should hopefully reduce number of calls to BGS.

pkarman:
deployed to production. so far we can see a great deal of overlap in the SSN and the file number we have stored:
```
Veteran.connection.execute("select count(*) from veterans where file_number=ssn").getvalue(0,0)
=> 2228

Veteran.connection.execute("select count(*) from veterans where ssn is not null and  file_number != ssn").getvalue(0,0)
=> 437
```

pkarman:
cache has been building over time.
```
irb(main):048:0> Veteran.connection.execute("select count(*) from veterans where ssn is not null and  file_number != ssn").getvalue(0,0)
[2019-07-16 12:59:45 -0400]    (434.4ms)  select count(*) from veterans where ssn is not null and  file_number != ssn
=> 11313
irb(main):049:0> Veteran.connection.execute("select count(*) from veterans where file_number=ssn").getvalue(0,0)
[2019-07-16 12:59:56 -0400]    (114.6ms)  select count(*) from veterans where file_number=ssn
=> 49353
```
closing as complete.

PR #11245, opened 2019-06-28
Veteran ssn cache

pkarman:
connects #11139 

Adds local `ssn` column to save trips to BGS for commonly needed (but seldom updated) attribute.

*Schema Changes Only*

* [x] Column comments updated

### DIFF
--- a/app/models/veteran.rb
+++ b/app/models/veteran.rb
@@ -188,7 +188,7 @@ class Veteran < ApplicationRecord
   end
 
   def ssn
-    super || bgs_record[:ssn]
+    super || (bgs_record.is_a?(Hash) && bgs_record[:ssn])
   end
 
   def validate_address
--- a/app/models/veteran.rb
+++ b/app/models/veteran.rb
@@ -208,7 +208,9 @@ class Veteran < ApplicationRecord
 
     is_stale = (first_name.nil? || last_name.nil? || self[:ssn].nil?)
     [:first_name, :last_name, :middle_name, :name_suffix, :ssn].each do |name|
-      is_stale = true if self[name] != bgs_record[name]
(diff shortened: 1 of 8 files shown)


leikkisa:
I am a big fan of this idea.

lib/generators/random.rb
```diff
@@ -23,7 +23,7 @@ def ssn
     end
 
     def unique_ssn
-      my_ssn = ssn
+      my_ssn = ssn.to_s
```
pkarman: tests seem to all expect a string

app/models/veteran.rb
```diff
@@ -202,19 +206,32 @@ def validate_address
   def stale_name?
     return false unless accessible? && bgs_record.is_a?(Hash)
 
-    is_stale = (first_name.nil? || last_name.nil?)
-    [:first_name, :last_name, :middle_name, :name_suffix].each do |name|
+    is_stale = (first_name.nil? || last_name.nil? || self[:ssn].nil?)
+    [:first_name, :last_name, :middle_name, :name_suffix, :ssn].each do |name|
```
leikkisa: SSNs are not going to change, and for some veterans they will always be missing.  I think we added the stale_name? not only because we were backfilling it, but because the name can change.

I don't think there are a lot of downsides to having this here now, except that extra work will be done.  It'd be nice to backfill the data and eventually remove this.  Maybe a comment, or separating it into a backfill_ssn would be nice.

If you do want to keep this here, maybe rename the method from stale_name?

pkarman: renamed to `stale_attributes`

I think we have to check for stale ssn just like name. I don't expect it to change as often as name might (which also isn't often) but that's the price we pay for caching.

PR #11247, opened 2019-06-28
Enable sorting of queue columns on back-end

lowellrex:
Connects #11054. Implements front-end ability to pass back-end requests that include fields we want to sort tasks by.

Before (Front-end sorting) | After (Pagination API)
--- | ---
before | after

This PR establishes a new precedent:
* Behaviour of columns is defined on the front-end in the functions for each column in `TaskTable`. The back-end describes *which* columns are displayed on the front-end, but the front-end defines how the column behaves on the front-end and what that column's attributes are. The back-end does not set the `backendCanSort` attribute, for instance.


app/models/task_pager.rb
```diff
@@ -6,34 +6,46 @@ class TaskPager
...
     @page ||= 1
+    @sort_by ||= Constants.QUEUE_CONFIG.CASE_DETAILS_LINK_COLUMN
+    @sort_order ||= Constants.QUEUE_CONFIG.COLUMN_SORT_ORDER_ASC
 
     fail(Caseflow::Error::MissingRequiredProperty, message: errors.full_messages.join(", ")) unless valid?
   end
 
   def paged_tasks
-    tasks_for_tab.order(:created_at).page(page).per(TASKS_PER_PAGE)
+    sorted_tasks(tasks_for_tab).page(page).per(TASKS_PER_PAGE)
```
lowellrex: Added back-end code for testing purposes. This will be replaced by whatever comes out of #11052.

lowellrex: This is not a very Rails-y way of chaining function calls (by nesting instead of by actually chaining methods), but I'm at a loss for a more Rails-y way to do this. Would appreciate any suggestions.

client/app/queue/QueueTable.jsx
```diff
@@ -190,7 +190,7 @@ export default class QueueTable extends React.PureComponent {
     const { defaultSort } = this.props;
     const state = {
       sortAscending: true,
-      sortColIdx: null,
+      sortColName: null,
```
lowellrex: We are using the column's name instead of its numeric position in the list of columns so that we can send the name to the back-end as part of the API request for the next/sorted page of tasks.

client/app/queue/components/TaskTable.jsx
```diff
@@ -42,6 +43,7 @@ const collapseColumn = (requireDasRecord) => (task) => hasDASRecord(task, requir
 export const docketNumberColumn = (tasks, requireDasRecord) => {
   return {
     header: COPY.CASE_LIST_TABLE_DOCKET_NUMBER_COLUMN_TITLE,
+    name: QUEUE_CONFIG.DOCKET_NUMBER_COLUMN,
```
lowellrex: Set name attributes on each column that is sortable so that we can [address it by name](https://github.com/department-of-veterans-affairs/caseflow/pull/11247/files#r299727386) in `QueueTable`.

client/app/queue/components/TaskTable.jsx
```diff
@@ -205,6 +217,7 @@ export const daysWaitingColumn = (requireDasRecord) => {
           diff(task.placedOnHoldAt, 'days')} limit={task.onHoldDuration} warning /> : null }
       </React.Fragment>;
     },
+    backendCanSort: true,
```
lowellrex: Our flag that the back-end can sort this column.

app/models/task_pager.rb
```diff
@@ -5,35 +5,53 @@ class TaskPager
...
+  end
+
+  def sorted_tasks(tasks)
+    case sort_by
+    when Constants.QUEUE_CONFIG.DAYS_ON_HOLD_COLUMN, Constants.QUEUE_CONFIG.TASK_DUE_DATE_COLUMN
+      tasks.order(assigned_at: sort_order.to_sym)
+    when Constants.QUEUE_CONFIG.TASK_CLOSED_DATE_COLUMN
+      tasks.order(closed_at: sort_order.to_sym)
+    when Constants.QUEUE_CONFIG.TASK_TYPE_COLUMN
+      tasks.order(type: sort_order.to_sym, action: sort_order.to_sym, created_at: sort_order.to_sym)
```
lowellrex: We sort by the task type first and then the action here which results in a slightly different sort order than front-end sorting which sorts using the label (which is a combination of the type and action fields). I think this is a fine tradeoff for now since only `ColocatedTask`s use the action field so only the VLJ support staff team queue could possibly have this problem (that is captured in the screencaps in the summary of this PR) and they will likely have only `ColocatedTask`s and won't actually encounter this funkiness in practice.

lowellrex: Also, we are sorting on the class name stored in the type field of the database and not the classes' labels so there may be mis-ordering as a result of that. We should address this eventually, but I think the fix might be somewhat involved (joining to a transient table that maps between class names and human-friendly labels) and they will usually sort in the same order, so I think it is fine to release this as is and improve it down the road.

app/models/task_pager.rb
```diff
@@ -5,35 +5,53 @@ class TaskPager
 
   validates :tab_name, presence: true
   validate :assignee_is_user_or_organization
+  validate :sort_order_is_valid
 
-  attr_accessor :assignee, :tab_name, :page
-  # attr_accessor :filters, :sort_by, :sort_order
+  attr_accessor :assignee, :tab_name, :page, :sort_by, :sort_order
+  # attr_accessor :filters
```
kevmo: We are keeping `:filters` for future use, I assume.

app/models/task_pager.rb
```diff
@@ -5,35 +5,53 @@ class TaskPager
...
+    # APPEAL_TYPE_COLUMN
+    # CASE_DETAILS_LINK_COLUMN
+    # DOCUMENT_COUNT_READER_LINK_COLUMN
+    # DOCKET_NUMBER_COLUMN
+    # HEARING_BADGE_COLUMN
+    # ISSUE_COUNT_COLUMN
+    # REGIONAL_OFFICE_COLUMN
+    # TASK_ASSIGNEE_COLUMN
+    # TASK_ASSIGNER_COLUMN
+    # TASK_HOLD_LENGTH_COLUMN
```
kevmo: :+1:

client/constants/QUEUE_CONFIG.json
```diff
@@ -20,5 +20,10 @@
...
 
   "TAB_NAME_REQUEST_PARAM": "tab",
-  "PAGE_NUMBER_REQUEST_PARAM": "page"
+  "PAGE_NUMBER_REQUEST_PARAM": "page",
+  "SORT_COLUMN_REQUEST_PARAM": "sort_by",
+  "SORT_DIRECTION_REQUEST_PARAM": "order",
+  
+  "COLUMN_SORT_ORDER_ASC": "asc",
+  "COLUMN_SORT_ORDER_DESC": "desc"
 }
```
kevmo: Github complaining about no newlines.

Issue #11188, opened 2019-06-24
Allow editing contention text on intake


leikkisa:
### Story
Currently users are only allowed to edit contention text for request issues during the "edit" step.
During the 6/20 AMO meeting, they requested that we allow users to edit contention text during the initial intake as well.

### Acceptance criteria
During intake, for reviews that are processed in VBMS, a user can edit the contention text for a request issue on the add issues page. When the contention is created in VBMS, it reflects the text submitted by the user.

Note - we need to confirm that CAs are the appropriate users for editing the contention text, or whether we want that to only remain on the edit screen which is more likely to be used by VSRs.

pkarman:
Dupe of #11248 ?

laurjpeterson:
@leikkisa - we've completed this, right? if so, let's close it.

leikkisa:
Done! Closing.

PR #11251, opened 2019-06-28
Edit contention on intake

leikkisa:
connects #11188


spec/feature/intake/higher_level_review/edit_spec.rb
```diff
@@ -1473,30 +1473,6 @@ def click_cancel(visit_page)
         create(:request_issue, :rating, decision_review: higher_level_review, contested_issue_description: "PTSD")
       end
 
-      scenario "edit contention text" do
```
leikkisa: This test largely duplicates what's in the test below it

PR #11261, opened 2019-07-01
add metadata endpoint for lighthouse api

edmangimelli:
### Description
Add a metadata endpoint for the lighthouse api which lists the available versions and provides links to docs and healthcheck.

### Acceptance Criteria
- [ ] A metadata endpoint returns a json object with an array of describing the different api versions


app/controllers/api/metadata_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class Api::MetadataController < ActionController::Base
+  protect_from_forgery with: :null_session
+
+  VERSION_STATUS = {
```
bastosmichael: Wouldn't it be better to have this in just one place instead of replicating it here and in ::ApplicationController? https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/controllers/application_controller.rb#L22-L27

app/controllers/api/metadata_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    previous: "Previous Version",
+    deprecated: "Deprecated Version"
+  }.freeze
+
+  def index
+    render json: {
+      meta: {
+        versions: [
+          {
+            version: "3",
```
bastosmichael: I would check with Derek on this but I blieve there was a whole discussion about wanting to stick with semantic versioning on this one. https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/controllers/claims_api/metadata_controller.rb#L12

app/controllers/api/metadata_controller.rb
erluti: this change is to the caseflow repo

bastosmichael: Got it sorry didn't realize that until after I commented...

app/controllers/api/metadata_controller.rb
edmangimelli: I get that changing the major version number in semver implies breaking changes, but I think one would still expect some overlap between versions. For instance, hypothetically, the claims API might have 3 major versions, with breaking changes between each, *but* each version still deals with claims.
The way API versioning is working in caseflow right now, v1 and v2, I believe, have no relation at all.
api/v1  - the system that runs their chron jobs
api/v2  - appeals status, with no overlap with v1
api/v3  - the root they wanted us to use

so each version is more like a numerically indexed namespace than a version number, if that makes sense. 
So, we could do semver starting at v3, but it would suck if they continued that trend, and they carved out a v4 for something that had no relation to the appeals API at all. Although, I suppose, as long no one else uses the v3 namespace, I suppose we're technically semver-ing OK.

edmangimelli: that being said--
so you want us to change it from `3` to `3.0.0` ?

dvogel: Can someone link me to this discussion? To my mind there's no need for anything beyond a major version number. The purpose of minor and patch numbers is to allow users to roll back to old versions to avoid bugs introduced in specific versions. For web APIs there's no way for a consumer to roll back to an older version of an API unless we host multiple versions. There's no reason to host multiple versions unless they are incompatible. If a bug was introduced, we would handle rolling the deployment back to a non-bugged revision.

edmangimelli: I suppose, if a semver is only present in a description, and not in any paths, then maybe it can be a hint that things have been updated. If anything more than just the major version number ended up in a path, or in a param, that could be extremely annoying for an API user

PR #11262, opened 2019-07-01
Add Print All Button for Hearing Worksheets

### LINKED ISSUE #10293: Print All Worksheets Function for Hearing Prep Daily Docket
## User Story

As a judge, I need to print all my worksheets for the day in one place. This saves me time and allows me to get all of my materials ready without opening each case and printing from there. 

## Mocks
https://va-gov.invisionapp.com/share/JHRCZI2GZF8

## Acceptance Criteria
- [ ] Verify there is a button at the top of the Hearing Prep Daily Docket that when pressed, sends all the VLJ Hearing Worksheets for that day to the default printer to print. 

## Related Tickets
User suggested update that replicates VACOLS functionality. Ticket here: https://github.com/department-of-veterans-affairs/appeals-support/issues/4823

ferristseng:
Resolves #10293

### Description

* Add new container to get multiple hearings to print
* Create new route that takes an comma-separate list of IDs
  * Use the print route that accepts multiple IDs instead of the one that accepts a single ID for consistency
* Wrap hearing worksheet in a table as a workaround to get footers displaying correctly for the multi-worksheet print view

### Acceptance Criteria

- [x] Verify there is a button at the top of the Hearing Prep Daily Docket that when pressed, sends all the VLJ Hearing Worksheets for that day to the default printer to print.


app/assets/stylesheets/print/hearings_worksheet.scss
```diff
@@ -234,12 +242,23 @@ $body-font-size: 14px;
...
-    width: 100%;
-    position: fixed;
-    bottom: 0;
-    right: $cell-padding;
+  table {
+    thead {
+      .cf-print-footer {
+        float: right;
+        position: relative;
+        top: 988px;
```
ferristseng: Having a hard-coded position here was the only way I could get the footer to stick to the bottom of the page, without overlapping with different footers in the multi-page view. I think it's OK because the page should be the same size every time, but I'm curious if it works across platforms as well.

client/app/hearings/containers/HearingWorksheetPrintAllContainer.jsx
```diff
@@ -0,0 +1,91 @@
...
+
+    if (_.isEmpty(hearingIds)) {
+      return (
+        <React.Fragment>
+          <AppSegment filledBackground>
+            <StatusMessage
+              type="alert"
+              title="No Hearings Specified"
+              messageText="Could not find hearings worksheets because none were specified"
+            />
```
lomaxap: Are you thinking there are other ways to get to this page (besides the daily docket) where the user could specify what hearings to print? Otherwise, language could be "There are no hearings scheduled for this docket".

We could also remove the print buttons altogether from the Daily Docket if there are no hearings.

app/assets/stylesheets/print/hearings_worksheet.scss
lomaxap: You may have stumbled upon this, but [this guy](https://medium.com/@Idan_Co/the-ultimate-print-html-template-with-header-footer-568f415f6d2a) seems to think this approach is sound.

client/app/hearings/containers/HearingWorksheetPrintAllContainer.jsx
ferristseng: This was added mostly as a catch-all to prevent nothing from displaying. I thought about removing the print button from the daily docket, but I'm thinking now it might be better to just disable it.

ferristseng:
I think the artifacts are if you have the header and footer option selected in Chrome (I had that option disabled while testing). I think I can probably tweak the margins a little bit to avoid that.

PR #11264, opened 2019-07-01
Use before_save hook instead of before_update

### LINKED ISSUE #11243: task closed_at not being set when status is completed
See thread in https://dsva.slack.com/archives/C54QCEHAL/p1561740095098900

From production:
```
> Task.completed.where(closed_at: nil).count
[2019-06-28 14:00:49 -0400]    (30.2ms)  SELECT COUNT(*) FROM "tasks" WHERE "tasks"."status" = $1 AND "tasks"."closed_at" IS NULL  [["status", "completed"]]
=> 816
```

```
> Task.completed.where(closed_at: nil).order(updated_at: :desc).first.updated_at
[2019-06-28 14:01:50 -0400]   Task Load (20.8ms)  SELECT  "tasks".* FROM "tasks" WHERE "tasks"."status" = $1 AND "tasks"."closed_at" IS NULL ORDER BY "tasks"."updated_at" DESC LIMIT $2  [["status", "completed"], ["LIMIT", 1]]
=> Tue, 18 Jun 2019 17:42:42 UTC +00:00
```

The before_update hook is apparently not always working. https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/task.rb#L460

aroltsch:
Resolves #11243  

### Description
Update timestamps on create and on update for tasks. 

Todo:
Once this PR is merged, I will update existing production records


spec/controllers/tasks_controller_spec.rb
```diff
@@ -86,7 +86,7 @@
       it "should process the request succesfully" do
         get :index, params: { user_id: user.id, role: "colocated" }
         response_body = JSON.parse(response.body)["tasks"]["data"]
-        expect(response_body.size).to eq 3
+        expect(response_body.size).to eq 4
```
aroltsch: Now this is correct. It is returning a recently completed task since `closed_at` is set correctly on the create of a closed task.

app/models/task.rb
```diff
@@ -449,17 +449,17 @@ def set_assigned_at
...
+    return unless will_save_change_to_status?
+
+    case status_change_to_be_saved&.last&.to_sym
+    when :assigned
+      self.assigned_at = Time.zone.now if !will_save_change_to_assigned_at?
+    when :in_progress
+      self.started_at = Time.zone.now if !will_save_change_to_started_at?
+    when :on_hold
+      self.placed_on_hold_at = Time.zone.now if !will_save_change_to_placed_on_hold_at?
+    when :completed, :cancelled
```
pkarman: I'm reading this as, if the status is changing, and the relevant timestamp column is not also flagged to be saved, then set the relevant timestamp to `now`. Otherwise leave the timestamp alone.

Is that correct?

aroltsch: Almost correct. Basically previously we had a "bug" where if you did the following: `task.update(status: "completed", closed_at: 2.weeks.ago)`, it would ignore `closed_at` value and set it to updated_at/Time.zone.now regardless. Now with this change, it will not set it to Time.now if you are passing a timestamp.

Issue #10756, opened 2019-05-14
Send source contention ID for contentions on DTA claims


pkarman:
Required https://github.com/department-of-veterans-affairs/connect_vbms/issues/219

When we create contentions as a result of a DTA error on an HLR, we should send the original contention_ids (from the request issues that got a DTA error) when we create the contentions.

If there are multiple prior contentions resulting in a single contention, send an array of the prior contention IDs.

leikkisa:
This is planned to be available in prod on 8/10.

PR #11267, opened 2019-07-01
Send original contention IDs when contention is from a DTA decision

leikkisa:
connects #10756

### DIFF
--- a/spec/models/end_product_establishment_spec.rb
+++ b/spec/models/end_product_establishment_spec.rb
@@ -416,18 +416,22 @@ describe EndProductEstablishment do
         ]
       end
 
-      it "sends the original contention ids when creating the contention" do
-        subject
+      context "when send send_original_dta_contentions is enabled" do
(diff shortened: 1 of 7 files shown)


spec/models/end_product_establishment_spec.rb
```diff
@@ -391,6 +391,45 @@
...
+
+      it "sends the original contention ids when creating the contention" do
+        subject
+
+        expect(Fakes::VBMSService).to have_received(:create_contentions!).once.with(
+          veteran_file_number: veteran_file_number,
+          claim_id: end_product_establishment.reference_id,
+          contentions: array_including(
+            description: "I am contesting a dta decision",
+            original_contention_ids: [101, 121]
```
pkarman: I assume the `connect_vbms` library takes care of joining these into a space-delimited string.

leikkisa: Yes, I put that into the connect_vbms PR. I just remembered that I need to put this behind a feature flag, so I'll add that before I merge it.

PR #11269, opened 2019-07-02
if VACOLS vbms_id fails to locate Veteran, try SSN

pkarman:
See https://github.com/department-of-veterans-affairs/dsva-vacols/issues/62 and https://dsva.slack.com/archives/CHX8FMP28/p1561993593116200 for context.

### DIFF
--- a/app/models/legacy_appeal.rb
+++ b/app/models/legacy_appeal.rb
@@ -227,10 +227,14 @@ class LegacyAppeal < ApplicationRecord
   end
 
   def veteran
-    @veteran ||= Veteran.find_or_create_by_file_number_or_ssn(veteran_file_number)
+    @veteran ||= begin
+      Veteran.find_or_create_by_file_number_or_ssn(veteran_file_number) || Veteran.find_or_create_by_ssn(veteran_ssn)
+    end
   end
 
   def veteran_ssn
+    return case_record.correspondent.ssn if case_record&.correspondent&.ssn
+
     vbms_id.ends_with?("C") ? (veteran&.ssn) : sanitized_vbms_id
   end
 
--- a/app/models/legacy_appeal.rb
+++ b/app/models/legacy_appeal.rb
@@ -227,9 +227,7 @@ class LegacyAppeal < ApplicationRecord
   end
 
   def veteran
-    @veteran ||= begin
-      Veteran.find_or_create_by_file_number_or_ssn(veteran_file_number) || Veteran.find_or_create_by_ssn(veteran_ssn)
-    end
+    @veteran ||= VeteranFinder.find_or_create_all(veteran_file_number, veteran_ssn).first
   end
 
   def veteran_ssn
--- a/app/models/veteran.rb
+++ b/app/models/veteran.rb
@@ -239,14 +239,14 @@ class Veteran < ApplicationRecord
       end
     end
 
-    private
-
     def find_or_create_by_ssn(ssn, sync_name: false)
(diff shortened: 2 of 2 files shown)


app/models/legacy_appeal.rb
```diff
@@ -227,10 +227,14 @@ def veteran_if_exists
   end
 
   def veteran
-    @veteran ||= Veteran.find_or_create_by_file_number_or_ssn(veteran_file_number)
+    @veteran ||= begin
+      Veteran.find_or_create_by_file_number_or_ssn(veteran_file_number) || Veteran.find_or_create_by_ssn(veteran_ssn)
```
aroltsch: Hmmm, should not this method `find_or_create_by_file_number_or_ssn` also search by ssn?

pkarman: it does search by SSN, yes. But it is passed the `file_number` so it will fail.

We are essentially providing a 2nd try, passing the ssn the 2nd time instead of the file_number.

aroltsch: Should we utilize VeteranFinder class https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/services/veteran_finder.rb#L5, I think it should do it and it is a little cleaner. Also, just an FYI, `find_or_create_by_ssn` is a private method on Veteran

pkarman: `VeteranFinder` is just a wrapper around `Veteran.find_or_create_by_file_number_or_ssn`.

The trick here is that we are trying the look up by a 2nd value (the SSN) instead of just giving up when the file number fails.

I made `Veteran.find_or_create_by_ssn` public in this PR.

pkarman: I do see that VeteranFinder can take an array though, so you're right. I'll try that.

pkarman:
you're right @aroltsch -- much cleaner to use VeteranFinder.

app/models/legacy_appeal.rb
```diff
@@ -227,10 +227,12 @@ def veteran_if_exists
   end
 
   def veteran
-    @veteran ||= Veteran.find_or_create_by_file_number_or_ssn(veteran_file_number)
+    @veteran ||= VeteranFinder.find_or_create_all(veteran_file_number, veteran_ssn).first
   end
 
   def veteran_ssn
+    return case_record.correspondent.ssn if case_record&.correspondent&.ssn
```
aroltsch: This might generate some unnecessary queries to VACOLS.

app/models/legacy_appeal.rb
```diff
@@ -227,10 +227,12 @@ def veteran_if_exists
   end
 
   def veteran
-    @veteran ||= Veteran.find_or_create_by_file_number_or_ssn(veteran_file_number)
+    @veteran ||= VeteranFinder.find_or_create_all(veteran_file_number, veteran_ssn).first
```
aroltsch: Nice!

app/models/legacy_appeal.rb
aroltsch: thinking a little more

aroltsch: you should be able to use a `cached` value instead of making another request, we are setting it here here https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/services/appeal_repository.rb#L196. However, your code would return an appellant ssn even before it would check for a veteran ssn existence, is that correct?

pkarman:
the specific case that spawned this turned out to be bad data, so I think we should avoid this.

PR #11270, opened 2019-07-02
Move state management inside table filters

lowellrex:
Related to #11054. This PR moves state management into the column filters from the parent component. This change fell out of some changes we are making to allow filtering to happen on the back-end.


client/app/queue/QueueTable.jsx
```diff
@@ -193,7 +191,6 @@ export default class QueueTable extends React.PureComponent {
     const state = {
       sortAscending: true,
       sortColIdx: null,
-      areDropdownFiltersOpen: {},
```
lowellrex: The `QueueTable` does not need to keep track of which dropdown menus on columns are open. This PR offloads that work on to the column headers themselves.

client/app/components/TableFilter.jsx
```diff
@@ -35,6 +30,12 @@ import FilterOption from './FilterOption';
  */
 
 class TableFilter extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.state = { open: false };
```
kevmo: Should we be using `this.setState`?

lowellrex: When we initialize the state in the constructor we use `this.state = ...`. [React docs for reference](https://reactjs.org/docs/state-and-lifecycle.html#adding-local-state-to-a-class).

PR #11275, opened 2019-07-02
Remove SSN presence validation

### LINKED ISSUE #11257: Allow Intake for Veteran without a SSN
A Veteran may legitimately not have a SSN and that should not prevent Intake.

During the initial Intake search, a missing SSN on the BGS record currently prevents the user from proceeding.

Contact @pkarman for an example file number.

monfresh:
**Why**: Some veterans legitimately do not have an SSN, and we don't
want to prevent Intake from proceeding due to a missing SSN.

Resolves #11257

### DIFF
--- a/spec/feature/intake/intake_spec.rb
+++ b/spec/feature/intake/intake_spec.rb
@@ -213,7 +213,7 @@ feature "Intake" do
         expect(page).to have_current_path("/intake/search")
         expect(page).to have_content("Please fill in the following field(s) in the Veteran's profile in VBMS or")
         expect(page).to have_content(
-          "the corporate database, then retry establishing the EP in Caseflow: ssn, country."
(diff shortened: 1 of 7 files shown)


spec/feature/intake/intake_spec.rb
```diff
@@ -213,7 +213,7 @@
         expect(page).to have_current_path("/intake/search")
         expect(page).to have_content("Please fill in the following field(s) in the Veteran's profile in VBMS or")
         expect(page).to have_content(
-          "the corporate database, then retry establishing the EP in Caseflow: ssn, country."
+          "the corporate database, then retry establishing the EP in Caseflow: country."
```
monfresh: @leikkisa @pkarman Is this existing spec representative of what happened in production, or do we need another scenario?

pkarman: yes this is the existing scenario.

PR #11276, opened 2019-07-02
Update EvidenceWindowTask Timer

lomaxap:
`EvidenceWindowTask` can be a child of an `AssignHearingDispositionTask` or a `DistributionTask`. This PR ensures that we're always using the hearing date, if present, to determine when the task expires.


app/models/tasks/evidence_submission_window_task.rb
```diff
@@ -15,14 +15,20 @@ def when_timer_ends
...
     from_date ||= appeal.receipt_date
 
     from_date + 90.days
   end
 
+  def hearing
+    @hearing ||= if parent.is_a?(AssignHearingDispositionTask)
+                   parent.hearing
+                 elsif parent.is_a?(DistributionTask)
+                   parent.children.find_by(type: "HearingTask")&.hearing
```
pkarman: is it possible that the parent will have more then one child `HearingTask`?

app/models/tasks/evidence_submission_window_task.rb
```diff
@@ -15,14 +15,20 @@ def when_timer_ends
...
-    from_date = if parent.is_a?(AssignHearingDispositionTask)
-                  parent.hearing&.hearing_day&.scheduled_for
-                end
+    from_date = hearing.hearing_day&.scheduled_for if hearing.present?
     from_date ||= appeal.receipt_date
 
     from_date + 90.days
   end
 
+  def hearing
```
lowellrex: Can we just use the most recent hearing for a given appeal? (`appeal.hearings.order { |h| h.hearing_day.scheduled_for }.last` or something similar) Will we ever create an `EvidenceSubmissionWindowTask` for a hearing other than the one that was scheduled for the latest date? This would make it so the hearing is less reliant on the task tree (which does not have a direct relationship with hearings).

lowellrex: Also, it looks like `EvidenceSubmissionWindowTask`s are children of a few other types of tasks not mentioned here:

```sql
select count(*)
, type
from tasks
where id in (
  select parent_id
  from tasks
  where type = 'EvidenceSubmissionWindowTask'
)
group by 2;
```

https://caseflow-looker.va.gov/sql/sqtzhjr4mtwsmk: 

Count | Type
--- | ---
14 | HearingTask
703 | DistributionTask
1 | AttorneyTask
104 | AssignHearingDispositionTask

lomaxap: the latest `scheduled_for` might not be the correct hearing (a hearing can be "postponed" do an earlier date), but the last created will be correct. We didn't add a `created_at` column to hearings, but updated to sort by id:
```
def hearing
    appeal.hearings.order(&:id).last
  end
```

PR #11277, opened 2019-07-02
Fix appeal withdrawn flake

aroltsch:
Fixes https://circleci.com/gh/department-of-veterans-affairs/caseflow/71823

The test was failing due to too many distribution tasks on the same appeal, this PR fixes this problem


monfresh:
I think all we need is to remove `create(:distribution_task, appeal: appeal)` on line 239 and update the expectation on line 24 from `on_hold` to `assigned`

PR #11283, opened 2019-07-03
Move pagination URL parameters into shared constants

lowellrex:
Connects #11054. Reduces the size of changes that will come with the follow-on PR that introduces the first instance of sorting pagination requests (#11247).

### DIFF
--- a/client/constants/QUEUE_CONFIG.json
+++ b/client/constants/QUEUE_CONFIG.json
@@ -13,6 +13,7 @@
   "ISSUE_COUNT_COLUMN": "issueCountColumn",
   "REGIONAL_OFFICE_COLUMN": "regionalOfficeColumn",
   "TASK_ASSIGNEE_COLUMN": "assignedToColumn",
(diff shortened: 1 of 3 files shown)


client/constants/QUEUE_CONFIG.json
```diff
@@ -13,5 +13,7 @@
   "ISSUE_COUNT_COLUMN": "issueCountColumn",
   "REGIONAL_OFFICE_COLUMN": "regionalOfficeColumn",
   "TASK_ASSIGNEE_COLUMN": "assignedToColumn",
-  "TASK_TYPE_COLUMN": "taskColumn"
```
hschallhorn: Reasoning for this being removed when it's still used in [queue_config](https://github.com/department-of-veterans-affairs/caseflow/blob/7f116e03e4cad315bb4d5a99eefa2c8fb7920b50/app/models/queue_config.rb#L102)?

lowellrex: Oversight! Good catch!

Issue #11063, opened 2019-06-07
Geo matching to closest RO office, irrespective of state


pshahVA:
## Background
Current state: Geo matching to veterans allows the hearings team to determine the closest regional office to that veteran.  This regional office is usually limited to the same state as the veteran with some exceptions.  

## Goals
Investigate the level of effort to change this alogorithm to look for the nearest regional office, irrespective of the state the Veteran lives in.

## AC
- Investigate the level of effort for the following two updates to the geo matching the veteran to the nearest regional office.
   - Option 1.) Find the closest regional office or alternate hearing location to the veteran, regardless of state lines.
   - Option 2.) Find the closest regional office to the veteran, regardless of state lines.

lomaxap:
@pshahVA where are we with this?

pshahVA:
This is the email that was sent to the Board: 



>  I have talked to the engineering team to get a better understand of the relative level of effort for the two geo matching options that you have asked us to investigate.
> Option 1: Match Veteran to closest Regional office.
> This will keep the existing functionality, but match a Veteran to a regional office that is closest to their physical address, irrespective of state they live in. This is a smaller engineering change that is not too disruptive to the application.
> Option 2: Match the Veteran to the closest regional office or alternate regional office.
> This is a significant change in the flow of how Veterans are scheduled and would be a much larger effort because Caseflow uses the Regional Office as the main way to organize hearings.
> Here's why: 
> The current flow to build the schedule and then schedule a Veteran is as follows:
> 1.	Build the schedule, by allocating and assigning hearing days to each Regional Office, which occurs through the yearly scheduling exercise of uploading non-availability dates of RO’s etc.
> 2.	When ready to schedule a Veteran, the Hearing Coordinator picks a Regional Office that they want to schedule. All Veterans closest to that regional office (RO) that are eligible to be scheduled will be shown in a list.
> 3.	The Hearing Coordinator then selects the Veteran they want to schedule. Then, Caseflow suggests the Veteran's closest hearing location based on the Veteran's address and a list of alternate hearing locations. A coordinator can confirm the suggested location or pick a different one based on their judgement and review.
> o	Right now, an alternate hearing location can only be selected if it is associated with the regional office. This is because we group alternate hearing locations with the ROs that have used them before. For example, the Houston RO has X alternate hearing location. X alternate hearing location is associated with Houston in our code.
[...]

PR #11284, opened 2019-07-03
facility_id  Utility Functions for RegionalOffices

lomaxap:
Connects #11063 

For use in refactoring `va_dot_gov_address_validator`


lomaxap:
> These aren't being used yet right?

Right, I have a script that samples appeals using this methods so we can get a sense of how the geomatching data will change. One of these methods will probably get nixed.

Issue #10982, opened 2019-06-03
Investigation | Allow Lit Support users to assign Motion to Vacate tasks to judges


laurjpeterson:
Judges are required to decide Motions, including Motions to Vacate. Litigation Support first receive Motion to Vacate mail tasks, but then assign them to judges to decide. They often assign them to the judge who originally wrote the decision.

Here's a picture of what we think the task tree needs to look like:
<img width="206" alt="Screen Shot 2019-06-03 at 12 58 18 PM" src="https://user-images.githubusercontent.com/27704093/58819935-49091f00-85ff-11e9-98d3-5b36fd99cef9.png">

## AC
- First, investigate which front end components we could use for this

For AMA appeals only:
- Put this work behind a feature flag
- Create a new task called Draft Motion to Vacate
- Motion to Vacate tasks have the Action available to them - "Send to judge for Draft Motion to Vacate"
- When they click "Send to judge", Lit Support user is prompted to select a judge
   - Note: we could auto-populate the judge who originally signed the decision's name
- Lit Support is able to add notes to describe what the judge should do
   - Note: could we use the assign admin action patterns?
- Draft Motion to Vacate task appears in Judge Review queue

## Notes
- Subsequent tickets will delineate how judges decide

laurjpeterson:
@leikkisa - going to close this one because it's investigation! the other one is #11252

PR #11289, opened 2019-07-03
[WIP] draft motion to vacate

leikkisa:
connects #10982


app/models/tasks/draft_motion_to_vacate_task.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class DraftMotionToVacateTask < JudgeTask
```
aroltsch: I think we should distinguish between Assign DraftMotionToVacateTask and Review DraftMotionToVacateTask. At first a judge assigns the task to a drafting attorney. When a drafting attorney finishes the decision, he/she would send the decision back to the judge for a final review.

aroltsch: Actually I take this back. This is okay for now.

leikkisa:
Code was finished in another PR

PR #11290, opened 2019-07-03
Replace canceled with cancelled, repo wide

lomky:
Ran "git grep -l 'canceled' | xargs sed -i 's/canceled/cancelled/g'"

(backed out any changes to the `db/migrations` dir)

Resolves one existential crisis in our codebase.


pkarman:
FWIW there would need to be a lot of db migrations to rename columns and update enum column values as well.

lomky:
True. The migrations aren't that many, but I hadn't though of enums yet. 

There's a reason it's a draft. I may abandon it as too extreme to be worth it :)

PR #11293, opened 2019-07-03
When creating a task, status should be assigned - send Sentry alert for now

aroltsch:
This is the first PR to enforce initial status for Task creation. We will first send all offenses to Sentry instead of throwing an error so not to block work.

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -16,12 +16,13 @@ class Task < ApplicationRecord
   has_many :task_timers, dependent: :destroy
 
   validates :assigned_to, :appeal, :type, :status, presence: true
+  validate :status_is_valid_on_create, on: :create
 
   before_create :set_assigned_at
   after_create :create_and_auto_assign_child_task, if: :automatically_assign_org_task?
   after_create :tell_parent_task_child_task_created
 
-  before_save :set_timestamps
+  before_save :set_timestamp
   after_update :update_parent_status, if: :task_just_closed_and_has_parent?
   after_update :update_children_status_after_closed, if: :task_just_closed?
 
@@ -323,6 +324,7 @@ class Task < ApplicationRecord
       t.assigned_by_id = self.class.child_assigned_by_id(parent, current_user)
       t.assigned_to = self.class.child_task_assignee(parent, reassign_params)
       t.instructions = [instructions, reassign_params[:instructions]].flatten
+      t.status = Constants.TASK_STATUSES.assigned
       t.save!
     end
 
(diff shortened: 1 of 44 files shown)


app/models/task.rb
```diff
@@ -321,6 +322,7 @@ def reassign(reassign_params, current_user)
       t.assigned_by_id = self.class.child_assigned_by_id(parent, current_user)
       t.assigned_to = self.class.child_task_assignee(parent, reassign_params)
       t.instructions = [instructions, reassign_params[:instructions]].flatten
+      t.status = Constants.TASK_STATUSES.assigned
```
aroltsch: This is the only place I found so far where we are creating a task with status other than assigned

app/workflows/ihp_tasks_factory.rb
```diff
@@ -6,6 +6,8 @@ def initialize(parent)
   end
 
   def create_ihp_tasks!
+    return unless @parent
```
aroltsch: Now because of activated callbacks, lots of tests are calling this factory and a lot of tasks don't have parents in the tests. I think creating a bunch parents in order for this to work is an overkill.

spec/feature/hearings/daily_docket_spec.rb
```diff
@@ -109,12 +109,13 @@
...
       create(:assign_hearing_disposition_task,
+             :completed,
              parent: hearing_task_association.hearing_task,
-             appeal: hearing.appeal,
-             status: Constants.TASK_STATUSES.completed)
+             appeal: hearing.appeal)
     end
 
     scenario "User cannot update disposition" do
+      hearing_task_association.hearing_task.update(status: :in_progress)
```
aroltsch: This is a strange use case and I was not sure how to approach it. There is a callback on the hearing_task when all children get completed/cancelled, the hearing_task status gets updated to completed/cancelled. However, this test case assumes that hearing_task is in progress.

spec/models/legacy_hearing_spec.rb
```diff
@@ -65,16 +67,22 @@
...
       let!(:disposition_task) do
         create(:assign_hearing_disposition_task,
-               parent: hearing_task_association.hearing_task,
-               status: Constants.TASK_STATUSES.cancelled)
+               :cancelled,
+               parent: hearing_task_association.hearing_task)
+      end
+
+      before do
+        hearing_task_association.hearing_task.update(status: :in_progress)
```
aroltsch: Same here for the hearing_task.

app/workflows/ihp_tasks_factory.rb
pkarman: It's not clear to me why callbacks are activated now and not before?

Could we change the factorybot classes in the tests to always assign a parent?

I worry that silently returning here in the code to reflect a pattern in the tests will bite us eventually.

app/models/task.rb
```diff
@@ -458,12 +460,23 @@ def set_assigned_at
...
+  def set_timestamp
     return unless will_save_change_to_attribute?(:status)
 
     timestamp_to_update = STATUS_TIMESTAMPS[status_change_to_be_saved&.last&.to_sym]
     return if will_save_change_to_attribute?(timestamp_to_update)
 
     self[timestamp_to_update] = Time.zone.now
   end
+
+  def status_is_valid_on_create
```
pkarman: I like this.

app/workflows/ihp_tasks_factory.rb
aroltsch: To always a assign a parent causes other issues. For example, to always assign a parent to a hearing_task in the factory itself breaks a ton of tests because hearing_task has a root task as its parent. So if there is already a root task or a root task is created later, it throws an error.

aroltsch: I am going to merge this PR the way it is and I will think about this problem.

Issue #11113, opened 2019-06-18
Change routing of admin actions


laurjpeterson:
When attorneys identify admin actions, the VLJ Support Staff and the Board would like them to be routed directly to the team that needs to action on them, rather than going through VLJ Support Staff as a middleman. 

This is different than our first implementation and set of requirements, which required all admin actions to go through VLJ Support Staff. 

## AC
- Update the following `ColocatedTask` routing logic:
   - Missing hearing transcript admin actions should be routed to the Hearing Management Branch queue
   - Schedule hearing admin actions should be routed to the Hearing Management Branch queue
   - FOIA/Privacy Act request admin actions should be routed directly to the Privacy team queue
   - Translation tasks should be routed directly to the Translation team queue
- Update existing data

## Questions
- Is it low effort to update the name of this group of tasks to be `AdminActions` instead of `ColocatedTask`s?

laurjpeterson:
@lowellrex - can echo estimate this ticket during the next session? also, i think it might absorb this ticket which was already estimated - https://github.com/department-of-veterans-affairs/caseflow/issues/10915

I also think that VLJ Support would benefit from us manually moving their existing Missing hearing transcript, FOIA/Privacy Act, and Schedule hearing tasks to the Hearing Branch. Would we be open to doing this only one time, since we'd fix the routing in the above ticket? if so, I'll create another ticket for that transfer to be estimated.

lowellrex:
We can estimate this ticket and we'll decide whether it makes sense to split it out into multiple tickets during estimation and if it does then we'll write those tickets.

laurjpeterson:
@lowellrex - added one more AC. Routing Translation Tasks to Translation team queue.

lowellrex:
Reminder that after we make the code changes for this PR that we should updated existing tasks that would have gotten bypassed had this code been live at the time they were created (the Missing hearing transcript, FOIA/Privacy Act, and Schedule hearing tasks [LP mentions above](https://github.com/department-of-veterans-affairs/caseflow/issues/11113#issuecomment-503206375)). That work can take place as part of a different ticket if it begins to balloon into a big endeavor.

hschallhorn:
I've got a few questions after looking into this. Within Caseflow, it appears that VLJ Support Staff only have the ability to directly act on

[...]

Issue #9057, opened 2019-01-30
Remove action column from tasks table


lowellrex:
Currently we use the action column of the task table as a psuedo-class that we use to determine various behaviours. This ticket exists to remove the code's reliance on that column (and as a follow-on to a false start in PR #7686).

Since the `Task.action` field is currently referenced across a broad swath of the codebase, I propose doing this overhaul in a few steps to avoid introducing a mass of changes that cause some bug that would be hard to troubleshoot or undo if we did the whole overhaul at once.

Before we remove the column from the database we can use [the `ignored_columns` Rails method](https://api.rubyonrails.org/classes/ActiveRecord/ModelSchema/ClassMethods.html#method-i-ignored_columns) on the `Task` class to ignore that field to see if any application logic is affected.

## 🚫 Blocked by
* [x] Migrate ColocatedTasks to their respective subclasses #11337
* [x] Remove code with behaviour based on ColocatedTask action field #11526

## Acceptance criteria
* [x] Create named subclasses for the different `ColocatedTask`s currently defined by their action field (#11295)
* [x] Migrate existing `ColocatedTask`s to new subclasses (combination of `type` and `action` fields) (#11337)
* [x] Make sure LegacyTasks are handled properly (#11817)
* [x] Remove references to the `action` field from the codebase
  * [x] Change task type (#11818) - `action` is mostly used as an argument in this workflow.
    * client/app/queue/ChangeTaskTypeModal.jsx
    * app/models/task.rb (same_task_type? function)
    * app/models/tasks/colocated_task.rb
    * app/models/tasks/mail_task.rb
  * [x] `JudgeTask` creation (#11819)
    * app/workflows/judge_assign_task_creator.rb
    * app/models/docket.rb
  * [x] Hearing disposition change workflow - Does not look like this actually needs to change because `action` is only used as an argument in the workflow.
    * client/app/queue/ChangeHearingDispositionModal.jsx
    * client/app/queue/components/PostponeHearingModal.jsx
    * app/models/tasks/assign_hearing_disposition_task.rb
* [x] Ignore the column in the class using `ignored_columns`, deploy code and wait to see if any bugs arise with the column live in production (let folks on bat-team know when this goes out and to bug the person who made the changes if bugs happen) (#11820)
    * app/models/task_pager.rb
    * app/controllers/tasks_controller.rb
* [ ] Remove the `action` column from the database in a migration

laurjpeterson:
[...]

PR #11295, opened 2019-07-03
Create subclasses for each ColocatedTask action

lowellrex:
Connects #11113. Midway PR before updating routing rules.


lowellrex:
Related to https://github.com/department-of-veterans-affairs/caseflow/issues/9057

spec/models/tasks/colocated_task_spec.rb
```diff
@@ -160,7 +160,7 @@
       end
 
       context "at the same time" do
-        let(:params_list) { [task_params, task_params] }
+        let(:params_list) { [task_params, task_params.clone] }
```
hschallhorn: `params.delete(:action)` was affecting the second test

app/models/tasks/colocated_task.rb
```diff
@@ -23,18 +23,9 @@ def create_many_from_params(params_array, user)
...
-          new_task_type = ColocatedTask.subclasses.find do |task_class|
-            task_class.label == Constants::CO_LOCATED_ADMIN_ACTIONS[input_action]
-          end
-          # TODO: Fail if there is no new_task_type.
-          params.merge!(type: new_task_type.name)
-
-          # TODO: Determine where this should be assigned shortly.
-          params.merge!(assigned_to: Colocated.singleton)
+          new_task_type = find_subclass_by_action(params.delete(:action).to_s)
+          params.merge!(type: new_task_type&.name, assigned_to: Colocated.singleton)
```
hschallhorn: `new_task_type&.name` allows a nil (invalid) task type to pass through, which will be caught by `valid_action_or_type` when the task is created

spec/factories/task.rb
```diff
@@ -74,70 +74,174 @@
       parent { create(:generic_task) }
     end
 
-    factory :colocated_task, class: ColocatedTask do
-      type { ColocatedTask.name }
+    factory :colocated_task, traits: [Constants::CO_LOCATED_ADMIN_ACTIONS.keys.sample.to_sym] do
```
hschallhorn: Picks a random colocated subclass to create

app/models/tasks/colocated_task.rb
```diff
@@ -22,7 +22,13 @@ class << self
     def create_many_from_params(params_array, user)
       # Create all ColocatedTasks in one transaction so that if any fail they all fail.
       ActiveRecord::Base.multi_transaction do
-        team_tasks = super(params_array.map { |p| p.merge(assigned_to: Colocated.singleton) }, user)
+        params_array = params_array.map do |params|
+          # Find the task type for a given action.
+          new_task_type = find_subclass_by_action(params.delete(:action).to_s)
```
lowellrex: Maybe we should create a copy of `params` so we don't alter the input too much? `p = params.clone...`

app/models/tasks/colocated_task.rb
```diff
@@ -70,14 +80,14 @@ def available_actions(user)
   end
 
   def available_actions_with_conditions(core_actions)
-    if %w[translation schedule_hearing].include?(action) && appeal.is_a?(LegacyAppeal)
+    if %w[TranslationColocatedTask ScheduleHearingColocatedTask].include?(type) && appeal.is_a?(LegacyAppeal)
```
lowellrex: Now that these classes exist can we overwrite this function in those classes?

Also, so that we remain backwards-compatible with existing `ColocatedTask`s that have actions (before we migrate those to the specific subclasses) can we keep around the existing condition that checks the task's action?

app/models/tasks/colocated_task.rb
```diff
@@ -89,11 +99,11 @@ def actions_available?(_user)
   end
 
   def create_twin_of_type(params)
-    self.class.create!(
+    task_type = ColocatedTask.find_subclass_by_action(params[:action])
```
lowellrex: 👍

app/models/tasks/colocated_task.rb
```diff
@@ -107,7 +117,7 @@ def ama_translation_actions(core_actions)
   end
 
   def legacy_translation_or_hearing_actions(actions)
-    return legacy_schedule_hearing_actions(actions) if action == "schedule_hearing"
+    return legacy_schedule_hearing_actions(actions) if is_a?(ScheduleHearingColocatedTask)
```
lowellrex: Let's leave the condition that depends on the `action` here and overwrite this function in the subclass. We can remove the `action` condition after we've migrated all existing tasks.

db/seeds.rb
```diff
@@ -826,7 +826,7 @@ def create_task_at_judge_review(appeal, judge, attorney)
     FactoryBot.create(:attorney_case_review, task_id: child.id)
   end
 
-  def create_task_at_colocated(appeal, judge, attorney, task_attributes = {})
+  def create_task_at_colocated(appeal, judge, attorney, trait = Constants::CO_LOCATED_ADMIN_ACTIONS.keys.sample.to_sym)
```
lowellrex: Wow. I like this approach a lot more! Great improvement.

spec/factories/task.rb
lowellrex: Love it! Great fix for the factory!

hschallhorn:
Adding comments for `vacols_location` and `available_actions` to point to this PR to eventually remove once we migrate ColocatedTasks in production.

app/models/tasks/colocated_task.rb
```diff
@@ -177,18 +198,43 @@ def task_is_unique
       appeal_id: appeal_id,
       assigned_to_id: assigned_to_id,
       assigned_to_type: assigned_to_type,
-      action: action,
+      type: type,
```
lowellrex: Let's leave action (as well as type) in so that we account for unique tasks during the transition.

app/models/tasks/schedule_hearing_colocated_task.rb
```diff
@@ -0,0 +1,14 @@
...
+  def self.label
+    Constants.CO_LOCATED_ADMIN_ACTIONS.schedule_hearing
+  end
+
+  def available_actions_with_conditions(core_actions)
+    core_actions = super(core_actions)
+    if appeal.is_a?(LegacyAppeal)
+      return legacy_schedule_hearing_actions(core_actions)
+    end
+  end
```
lowellrex: I think we want to return `core_action` here if the appeal is not a `LegacyAppeal`, right?

hschallhorn:
Also connects #9057!

PR #11296, opened 2019-07-03
Fix Hearings Worksheet Link

### LINKED ISSUE #5390: Fix hearing worksheet link
When an attorney clicks the Hearing Worksheet link in Case Details, a new tab opens up and the print functionality immediately shows up. I think the expected result is to show the .pdf view of the hearing worksheet, rather than prompt the user to print.

Something like this shows up:
screen shot 2018-05-03 at 4 26 50 pm

## AC
- When the user clicks Hearing Worksheet, the Hearing Worksheet pdf opens in a new tab.

ferristseng:
Resolves #5390

### Description

* add way of overriding stylesheets for the hearings index file
  * move styling from print stylesheet to new overrides stylesheet, so the read-only view resembles the print view more closely
* add `keep_open` parameter to anywhere the print view of the hearings worksheet gets linked to


ferristseng:
The ticket says that it should open the hearing worksheet in PDF view. I did some research to see how feasible this was, and I think it's possible, but it would be very difficult to replicate what gets printed exactly.

Instead, I chose to copy the print CSS to a separate stylesheet, and have the print view show something similar to what would be printed. There's some things that get lost like the pagination for longer worksheets, but I think it does a good job of giving an idea of what the final printout will look like. The user also still has the option of printing the page or using the print dialogue to save the page as a PDF.

Issue #11109, opened 2019-06-17
Nightly data integrity checks job


pkarman:
We are starting to accrue the need for some nightly checks on data integrity. Let's have a single job that runs that performs the relevant queries and sends them to a Slack channel.

lomaxap:
What do you all think of a utility class like [this](https://github.com/department-of-veterans-affairs/caseflow/pull/11169/files) to check for duplicates.

pkarman:
As part of this it would be helpful if the `SlackService` library could route to different channels.

pkarman:
## Proposal

An abstract class called `DataIntegrityChecker`:

```ruby
class DataIntegrityChecker
  def call
  end

  def report?
  end

  def report
  end

  def slack_channel
    "appeals-app-alerts"
  end
end
```

A single job called `DataIntegrityChecksJob` that loads all subclasses of `DataIntegrityChecker` and runs them. E.g.

```ruby
class DataIntegrityChecksJob < CaseflowJob
  def perform
    list_of_data_integrity_checkers.each do |klass|
      checker = klass.new
      checker.call
      if checker.report?
        send_to_slack(checker.slack_channel, checker.report)
      end
    end
  end
end
```

pkarman:
@lomaxap @tomas-nava @lowellrex @aroltsch curious to know what you think of this proposal.

lowellrex:
Makes sense to me. Do we have a list of jobs that this model would replace?

pkarman:
to start,

* UntrackedLegacyAppealsReportJob
* the async job report currently hacked in to PrepareEstablishClaimTasksJob
* some combination of #10925 and #11169

aroltsch:
This looks good to me!

Issue #11183, opened 2019-06-24
Fix CASEFLOW-charged VACOLS appeals with no active tasks


lowellrex:
While working on resolving the problem of `ColocatedTask`s for `LegacyAppeal`s being [charged to the wrong VACOLS location](https://github.com/department-of-veterans-affairs/caseflow/issues/11152) when they were completed, I decided to look into the question of whether there were any appeals charged to the "CASEFLOW" VACOLS location that do not have active tasks in Caseflow and discovered that there are 10 appeals in this state.

One case charged to "CASEFLOW" in VACOLS does not have an associated `LegacyAppeal` in Caseflow. The other nine cases have associated `LegacyAppeal`s but the only tasks they have in Caseflow are `RootTask`s. Caseflow will never change the VACOLS location of these 10 appeals away from "CASEFLOW".

### Immediate recovery actions
* Charge these cases away from CASEFLOW in VACOLS (probably to the most recent prior location?)

### Long-term prevention actions
* Determine how these 10 appeals ended up in this situation and prevent it from reoccurring

It may be difficult to determine how this happened (prior location table may provide some clues though) and it may have little return on the time investment. A more efficient solution might be to set up a regular (daily?) alert that notifies us when there are any appeals charged to "CASEFLOW" in VACOLS so that we can check the logs when this situation reoccurs.

Commands run in production that discovered this data integrity issue:

```ruby
  rails c> vacols_ids = VACOLS::Case.where(bfcurloc: "CASEFLOW").pluck(:bfkey)
  rails c> vacols_ids.count
  # 28222
  rails c> appeal_ids = LegacyAppeal.where(vacols_id: vacols_ids).pluck(:id)
  rails c> appeal_ids.count
  # 28221
  # Which case charged to CASEFLOW in VACOLS does not exist in Caseflow?
  rails c> vacols_ids.sort - LegacyAppeal.where(id: appeal_ids).pluck(:vacols_id).sort
  # ["2832564"] <- bfkey
  rails c> appeal_with_tasks_ids = Task.where.not(type: [RootTask.name, TrackVeteranTask.name]).where(appeal_type: LegacyAppeal.name, appeal_id: appeal_ids).pluck(:appeal_id)
  rails c> appeal_with_tasks_ids = appeal_with_tasks_ids.uniq
  rails c> appeal_with_tasks_ids.count
  # 28212
  # Which appeals don't have active tasks?
[...]

PR #11297, opened 2019-07-03
Data integrity checks job

pkarman:
connects #11109 

This creates an abstract base class called `DataIntegrityChecker` that subclasses can implement. If registered with the daily job `DataIntegrityChecksJob` then they will run sequentially and report to Slack.

Also changes the `SlackService` to allow customizable channel and title of the message.


app/services/slack_service.rb
```diff
@@ -1,19 +1,42 @@
...
+  DEFAULT_CHANNEL = "#appeals-app-alerts"
 
-  attr_accessor :url
+  def initialize(url:)
+    @url = url
+  end
+
+  attr_reader :url
+
+  def send_notification(msg, title = "", channel = DEFAULT_CHANNEL)
```
pkarman: changes tested manually via rails console. evidence at https://dsva.slack.com/archives/C2TB5MH6F/p1562186636041800

pkarman:
I'm going to go ahead and merge. The actual cron job to hit this reporter won't be deployed for a day or 2 so we have some time to finish #11183

PR #11304, opened 2019-07-05
Migration to update timestamps for tasks

aroltsch:
Fixes production tasks that are missing timestamps


db/migrate/20190705172439_update_task_timestamps.rb
```diff
@@ -0,0 +1,9 @@
+class UpdateTaskTimestamps < ActiveRecord::Migration[5.1]
+  def change
+    Task.find_each do |task|
+      task.update(closed_at: task.updated_at) if !task.open? && task.closed_at.nil?
```
pkarman: maybe use `update!` so we fail on exception?

aroltsch: ok!

PR #11321, opened 2019-07-08
Only show tracking tasks tab for Field VSO queues

### LINKED ISSUE #9757: Limit queue table view for Field VSOs to tracking tasks tab
#9647 added the ability to easily create `FieldVso`s. This ticket exists to make it so their queue table view only shows the "All Cases" tab and none of the other tabs.

## Acceptance criteria
* [ ] Create new `QueueTab` class
* [ ] Only show the "All cases" tab in the organizational queue view for these VSOs
* [ ] Hide tab when there is only a single tab being shown

lowellrex:
Resolves #9757. Team queues for Field VSOs and Private Bar now only show the "All cases" tab.

Before (defaulted to "Unassigned" tab) | After
--- | ---
image | image

This PR includes some architectural changes to how we construct the `QueueConfig` object as well. Instead of defining the columns in the `QueueConfig` class itself, we now describe the tabs in dedicated classes that the `Organization` model instantiates and returns as an array of tabs from the `queue_tabs` method. Curious to get some early feedback on that design.


app/models/queue_config.rb
```diff
@@ -33,25 +33,20 @@ def use_task_pages_api?(user)
   end
 
   def tabs
-    [
-      include_tracking_tasks_tab? ? tracking_tasks_tab : nil,
-      unassigned_tasks_tab,
-      assigned_tasks_tab,
-      completed_tasks_tab
-    ].compact
+    organization.queue_tabs
```
lowellrex: This is probably a good place to start exploring the changes.

app/models/representative.rb
```diff
@@ -14,10 +14,27 @@ def can_receive_task?(_task)
     false
   end
 
+  def show_reader_link_column?
+    false
+  end
```
lowellrex: We could alternatively just overwrite the `unassigned_tasks_tab` method to have a false `show_reader_link_column` argument upon instantiation instead of adding this method. I don't have a strong preference between those approaches.

client/app/components/TabWindow.jsx
```diff
@@ -66,24 +66,26 @@ export default class TabWindow extends React.Component {
...
+                aria-label={this.getTabHeader(tab)}
+                disabled={Boolean(tab.disable)}
+              >
+                <span>
+                  {this.getTabHeaderWithSVG(tab)}
+                </span>
+              </button>
+            )}
+          </div>
+      }
```
kevmo: Not a blocker (obviously), but this diff seems like 8x more lines than it actually is!

app/models/organization.rb
```diff
@@ -71,6 +75,31 @@ def user_admin_path
...
+  end
+
+  def assigned_tasks_tab
+    ::AssignedTasksTab.new(assignee_name: name, show_regional_office_column: show_regional_office_in_queue?)
+  end
+
+  def completed_tasks_tab
+    ::CompletedTasksTab.new(assignee_name: name, show_regional_office_column: show_regional_office_in_queue?)
+  end
+
```
kevmo: I am a fan of this approach.

app/models/queue_tabs/tracking_tasks_tab.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def columns
+    [
+      Constants.QUEUE_CONFIG.CASE_DETAILS_LINK_COLUMN,
+      Constants.QUEUE_CONFIG.ISSUE_COUNT_COLUMN,
+      Constants.QUEUE_CONFIG.APPEAL_TYPE_COLUMN,
+      Constants.QUEUE_CONFIG.DOCKET_NUMBER_COLUMN
+    ]
+  end
+end
```
kevmo: It's not a blocker for this PR (and this was not introduced in this PR) but it's a little confusing how our project-specific nomenclature uses "all cases" and "tracking tasks" somewhat interchangeably, and it would be cool to change that sometime in the future.

spec/models/organizations/field_vso_spec.rb
```diff
@@ -1,10 +1,19 @@
...
     end
   end
+
+  describe ".queue_tabs" do
+    it "only returns a single tab with tracking tasks" do
+      tabs = vso.queue_tabs
+      expect(tabs.length).to eq(1)
+      expect(tabs.first).to be_a(TrackingTasksTab)
+    end
+  end
```
kevmo: Good test.

PR #11324, opened 2019-07-08
Lighthouse API higher level review status endpoint

### LINKED ISSUE #11323: Lighthouse API - expand spec to include status after HigherLevelReview is accepted
This PR introduced a spec for upcoming lighthouse API work: https://github.com/department-of-veterans-affairs/caseflow/pull/11216
We need to add support for the status of an accepted HigherLevelReview.

erluti:
Resolves #11323

### Description
Primary change updates specs to support a status endpoint.  Response for status is based on the [HLRStatusSerializer](https://github.com/department-of-veterans-affairs/caseflow/blob/3e0639a69202486a15e4b2b8bb763754c14b236e/app/models/serializers/v2/hlr_status_serializer.rb).
Also adds in some additional details for error response and makes small fixes with other document issues.


app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -136,11 +140,151 @@ paths:
...
+        '5XX':
           description: Failed
+          content:
+            application/vnd.api+json:
+              schema:
+                type: object
+                properties:
+                  errors:
+                    $ref: "#/components/schemas/errors"
+  /higher_level_reviews/status/{id}:
```
pkarman: we have typically exposed the `uuid` in user-facing URLs rather than the `id` -- I'm not sure if it matters here, but something to be aware of.

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -136,11 +140,151 @@ paths:
...
+      parameters:
+        - name: id
+          in: path
+          required: true
+          description: ID of an Accepted Higher-Level Review
+          schema:
+            type : integer
+            format: int64
+            minimum: 1
+      summary: After submitting a Higher-Level Review, you can check to see it's current status
```
pkarman: typo -- should be "its"

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -136,11 +140,151 @@ paths:
...
+      responses:
+        '200':
+          description: Status found
```
pkarman: do you need to differentiate between a HLR that has been created-but-not-yet-established?

app/controllers/api/docs/v3/decision_reviews.yaml
erluti: i forgot that a uuid is a string, i'll update this and clarify.  thanks!

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -136,11 +140,151 @@ paths:
...
+                    properties:
+                      id:
+                        type: string
+                      type:
+                        type: string
+                        description: Will be "higherLevelReview"
+                      attributes:
+                        type: object
+                        properties:
+                          appealIds:
```
leikkisa: What is this for?

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -136,11 +140,151 @@ paths:
...
+                            properties:
+                              type:
+                                type: string
+                              details:
+                                type: object
+                                properties:
+                                  description:
+                                    type: string
+                                  disposition:
+                                    type: string
```
leikkisa: For HLRs I believe we only have this at the issue level (different from legacy appeals which has it at the appeal and issue level)

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -136,11 +140,151 @@ paths:
...
+                                lastAction:
+                                  type: string
+                                  format: date-time
+                                date:
+                                  type: string
+                                  format: date-time
+                                description:
+                                  type: string
+                                diagnosticCode:
+                                  type: string
```
leikkisa: We don't currently capture this field, but I think you can get it from the rating issue

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -136,11 +140,151 @@ paths:
...
+                            type: array
+                            items:
+                              type: object
+                              properties:
+                                active:
+                                  type: boolean
+                                lastAction:
+                                  type: string
+                                  format: date-time
+                                date:
```
leikkisa: What date does this represent? My guess is the previous decision date.

app/controllers/api/docs/v3/decision_reviews.yaml
erluti: I don't believe so, the HigherLevelReview model takes that into account for the `status_hash` which is used a `status` in the serializer

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -136,11 +140,151 @@ paths:
...
+                            type: boolean
+                          active:
+                            type: boolean
+                          description:
+                            type: string
+                          location:
+                            type: string
+                          aoj:
+                            type: string
+                          programArea:
```
leikkisa: Is this benefit type?

app/controllers/api/docs/v3/decision_reviews.yaml
erluti: I was just describing what already existed in this serializer https://github.com/department-of-veterans-affairs/caseflow/blob/3e0639a69202486a15e4b2b8bb763754c14b236e/app/models/serializers/v2/hlr_status_serializer.rb
So I guess I don't actually know what it is?  Any feedback on what should be in this status response would be great.  I could make a v3 serializer and scale back what it returns.

app/controllers/api/docs/v3/decision_reviews.yaml
erluti: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/claim_review.rb#L186
This is where issues come from, so it's whatever `date` column is on DecisionReviews

erluti: Ah, so specifically this is where the issue properties come from:
https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/decision_review.rb#L454

edmangimelli:
lgtm with the exception of the remaining comments

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -423,8 +456,170 @@ paths:
...
+                            description: Will be empty https://github.com/department-of-veterans-affairs/caseflow/blob/c57d95f6ea11985b7658782cc82005107e9c3156/app/models/serializers/v2/appeal_status_serializer.rb#L37
+                            items:
+                              type: object
+                              nullable: true
+              examples:
+                received_status:
+                  value:
+                    data:
+                      id: HLR1
+                      type: 'higherLevelReview'
```
edmangimelli: I know we got some consensus from LightHouse engineers about using camel case _for object keys_. But do we really want to use it for constants? I can understand wanting to be consistent, *but* I'd much rather have constants match those that are already in caseflow.

app/controllers/api/docs/v3/decision_reviews.yaml
erluti: I think I'm going to pull back on using the existing serializer and use our own.  So I'm going to cut way back on what the status endpoint returns to reflect the jsonapi recommendation for asynchronous processing.  Thanks for your questions, they helped a lot

erluti:
Made some bigger changes, I meant for this all to conform with https://jsonapi.org/recommendations/#asynchronous-processing.  So I stripped back what the status endpoint is returning to a simple enum based on https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/concerns/asyncable.rb

erluti:
I was trying to use labels to communicate that I was going to make changes and a label to communicate that I was ready for it to be looked at again, I didn't realize it would auto-merge.  It's only spec changes, but please still leave any relevant feedback

Issue #11035, opened 2019-06-06
VBMS MissingVeteranIdentifier for FetchDocumentForReaderUserJob


ferristseng:
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/5274/events/361143/

Looks like this could be happening because the appeal ID that's recorded in the sentry event does not exist. Seeing a `nil` file number in the VBMS request, which is probably why it's getting an error back.

monfresh:
This was fixed in #11326

PR #11326, opened 2019-07-08
Don't look up nil file number in VBMS

monfresh:
While looking into a Sentry error, I noticed that our existing code was
not handling the case where the BGS lookup would return `nil`, which
resulted in another VBMS query with a `nil` file number, resulting in
yet another VBMS error due to the missing file number.

I took this opportunity to refactor the code to make it easier to read
and test by extracting the `fetch_documents_for` code into its own
class.


app/services/vbms_documents_for_appeal.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  def veteran_file_number_docs_request
+    VBMS::Requests::FindDocumentVersionReference.new(veteran_file_number)
+  end
+
+  def bgs_claim_number_docs_request
+    VBMS::Requests::FindDocumentVersionReference.new(bgs_claim_number)
+  end
+
+  def send_and_log_request(vbms_id, request)
```
pkarman: is this now duplicated in the ExternalAPI::VBMSService class?

monfresh: Yes, for now, although it's not exactly duplicated. This new class doesn't need the `override_vbms_client` argument. The goal is to eventually move all the public methods into their own classes. One sign that a class is doing too much or doesn't have a specific purpose is when it has a generic name like "VBMSService" and contains a collection of unrelated methods.

This leads to duplicated code inside each method like `@vbms_client ||= init_vbms_client` which is a hint that this might be better inside an `initialize` method. Additionally, having all these unrelated methods leads to code like the `send_and_log_request` method taking an optional `override_vbms_client` argument that only applies to certain methods. Instead of conditionally choosing which VBMS client to use, it's better to adhere to "Tell, Don't Ask" and initialize the proper client in the proper context.

Once all the methods are extracted, if there is exact duplication of the `send_and_log_request` method, then it might make sense to move it to a shared module. Otherwise, this kind of duplication is fine.

monfresh:
Fixes https://github.com/department-of-veterans-affairs/caseflow/issues/11035

app/services/vbms_documents_for_appeal.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
+
+class VbmsDocumentsForAppeal
```
pkarman: What do you think about keeping this class in the same `ExternalAPI` namespace (directory)?

app/services/vbms_documents_for_appeal.rb
pkarman: I generally agree that our BGS and VBMS external API service classes are following a junk drawer pattern and could use some refactoring love. It would be nice if they both ended up with similar interfaces.

app/services/vbms_documents_for_appeal.rb
monfresh: Sounds good to me. Done here: https://github.com/department-of-veterans-affairs/caseflow/pull/11326/commits/8f103b4c035c6143282fe6e6f450a4773b251045

I also had to add a missing spec that our test coverage check was complaining about. In the process, this revealed some design issues that made it hard to test the happy path, which I addressed by moving the document creation to a new class that we can mock in our tests since the document creation is already unit tested in `document_spec.rb`. 

While working on this, I also realized that this better design has the added benefit of not requiring Rails to run this test. If we only required `spec_helper`, these tests would run in 0.05 seconds version 7.5, which is 150x faster! 

I will prep a PR to remove the `--require rails_helper` flag from `.rspec` and instead manually require rails_helper only in tests that need it.

spec/services/vbms_documents_for_appeal_spec.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
+
+describe ExternalApi::VbmsDocumentsForAppeal do
+  context "when VBMS cannot find file number and BGS returns nil for claim number" do
+    it "raises and does not try to look up the nil number in VBMS" do
+      veteran_file_number = "12345678"
```
pkarman: this same set up code appears in all the tests. how about using `let` like our other specs do?

app/services/vbms_documents_for_appeal.rb
monfresh: I agree.

spec/services/vbms_documents_for_appeal_spec.rb
monfresh: Done.

PR #11330, opened 2019-07-09
add error responses to the Higher-Level Review spec

edmangimelli:
Resolves:
https://app.zenhub.com/workspaces/developervagov-5ac272fff41bfa50c0e46c48/issues/department-of-veterans-affairs/vets-contrib/2561

### Description
Enumerate all of the possible error responses for the Higher Level Review endpoint (Lighthouse) by adding example to openapi spec

### Acceptance Criteria
- [ ] HLR error responses are adequately documented

### Testing Plan
- [ ] `decision_reviews.yaml` renders correctly in editor.swagger.io

### DIFF
--- a/app/controllers/api/docs/v3/decision_reviews.yaml
+++ b/app/controllers/api/docs/v3/decision_reviews.yaml
@@ -138,7 +138,291 @@ paths:
         '202':
           description: Created
         '400':
-          description: Invalid Request
+          description: |-
+            400 response
+          content:
+            application/vnd.api+json:
+              examples:
+                malformed_request:
+                  value: {
+                    "errors": [
+                      {
+                        "status": "400",
+                        "title":  "Malformed request",
+                        "code":  "malformed_request"
+                      }
+                    ]
+                  }
+        '401':
+          description: |-
+            401 response
+          content:
+            application/vnd.api+json:
+              examples:
+                unauthenticated:
+                  value: {
+                    "errors": [
+                      {
+                        "status": "401",
+                        "title":  "Unauthenticated",
+                        "code":  "unauthenticated"
+                      }
+                    ]
+                  }
+        '403':
+          description: |-
+            403 response
+          content:
+            application/vnd.api+json:
+              examples:
+                veteran_not_accessible:
+                  value: {
+                    "errors": [
+                      {
+                        "status": "403",
+                        "title":  "Veteran File inaccessible",
+                        "code":  "veteran_not_accessible"
+                      }
+                    ]
+                  }
+        '404':
+          description: |-
+            404 response
+          content:
+            application/vnd.api+json:
+              examples:
+                veteran_not_found:
+                  value: {
+                    "errors": [
+                      {
+                        "status": "404",
+                        "title":  "Veteran File not found",
+                        "code":  "veteran_not_found"
+                      }
+                    ]
+                  }
+                isssue_not_found:
+                  value: {
+                    "errors": [
+                      {
+                        "status": "404",
+                        "title":  "Issue not found",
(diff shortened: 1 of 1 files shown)


edmangimelli:
Issue
https://github.com/department-of-veterans-affairs/caseflow/issues/11339

Lighthouse Tickets
https://app.zenhub.com/workspaces/developervagov-5ac272fff41bfa50c0e46c48/issues/department-of-veterans-affairs/vets-contrib/2561

https://app.zenhub.com/workspaces/developervagov-5ac272fff41bfa50c0e46c48/issues/department-of-veterans-affairs/vets-contrib/2589

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -138,7 +138,291 @@ paths:
...
+                  value: {
+                    "errors": [
+                      {
+                        "status": "404",
+                        "title":  "Veteran File not found",
+                        "code":  "veteran_not_found"
+                      }
+                    ]
+                  }
+                isssue_not_found:
```
leikkisa: There is a typo here

leikkisa: When does this error happen?

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -138,7 +138,291 @@ paths:
...
+                  value: {
+                    "errors": [
+                      {
+                        "status": "404",
+                        "title":  "Issue not found",
+                        "code":  "isssue_not_found"
```
leikkisa: Same typo

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -138,7 +138,291 @@ paths:
...
+                        "code":  "higher_level_review_to_higher_level_review"
+                      }
+                    ]
+                  }
+                appeal_to_appeal:
+                  value: {
+                    "errors": [
+                      {
+                        "status": "422",
+                        "title":  "Issue is ineligible because it was last processed as a Board Appeal and this can't be done twice in a row",
```
leikkisa: You only need this for appeals, so if this file is for HLRs you don't need it here, but if it's for decision reviews then it's fine.

app/controllers/api/docs/v3/decision_reviews.yaml
edmangimelli: At some point in the future, I'm imagining we'll provide a seperate endpoint which will return an array of all of the issues on file (in caseflow, vbms, etc) for a specific veteran. Whoever is using our API will use that array of issues, selecting whichever ones they want from it, and then add in issues manually (that weren't in any system), and then take that array of issues (some with IDs and some without) and use one of our decision review endpoints (like this one) and request a decision review.
That's when this error would occur --when they provide an ID that's invalid. And, like I mentioned, if an issue in that array doesn't have an ID, we know that this is a new issue that they are creating and the `issue_not_found` error is not applicable.
The `veteran_not_found` and `claimant_not_found` errors are analogous.

edmangimelli: @erluti @pkarman @leikkisa 
if you think this plan ^^^ won't work. please comment. thanks!

app/controllers/api/docs/v3/decision_reviews.yaml
edmangimelli: thanks! oversight

app/controllers/api/docs/v3/decision_reviews.yaml
edmangimelli: thanks!

PR #11341, opened 2019-07-09
Allow Judges to See All Hearing Days

### LINKED ISSUE #10563: Filtered Judge View of Hearing Schedule
Currently, judge view of the hearing schedule is (1) hearing days that are assigned to them, and (2) hearing days with individual hearings that are assigned to them. We could give judges the entire hearing schedule and allow them to filter. But, they wouldn't be able to see their assigned hearings (as opposed to hearing days), so **we think we should do this filtering for them.**  

## User Stories
As a judge, I need to see my schedule first when I log in so that I can focus on my hearing days and prepare accordingly. 

As a judge, I need to see the entire hearing schedule so that I can easily switch days with other judges. 

## Mocks
Original mocks can be found in Invision here: https://va-gov.invisionapp.com/share/PMR1ID7XSHT#/screens/353826170

Screen Shot 2019-04-25 at 5 43 25 PM

## Acceptance Criteria
- [ ] Verify the Schedule view lists AMA and Legacy video, CO, and travel board hearings that are by default filtered by that judge.
- [ ] Verify the judge has the ability to clear filters and see the entire schedule ahead of time

## Associated Tickets
#10139 - Hearing Prep Refactor

lomaxap:
Resolves #10563

- Adds button to Hearing Schedule List for judges: "Your Hearing Schedule", "Complete Hearing Schedule"
- Adds `show_all` param to `HearingDayController` to override filtering by judge's hearings


client/app/hearings/components/dailyDocket/DailyDocket.jsx
```diff
@@ -61,12 +61,26 @@ export default class DailyDocket extends React.Component {
...
+    const { hearings, user } = this.props;
+
+    return _.orderBy(Object.values(hearings).map((hearing) => (
+      {
+        ...hearing,
+        sortOrder: this.isJudgeHearing(user.userId, hearing) ? 0 : 1
+      }
+    )), ['sortOrder', 'scheduledTimeString']);
+  }
+
```
sharonwarner: I don't think I'm following this logic. Are we returning all hearings to the frontend for the judge view of the daily docket even if they aren't assigned to the judge? I feel a little weird about sending information to the frontend that shouldn't be visible to the user in the UI. Would it be significantly harder to move this logic to the backend?

lomaxap: need @Chingujo 's input on this

This logic doesn't hide anything from the user, it only sorts the hearings that are not assigned to the judge to the bottom of the list.

I made the assumption that if we are showing hearing days that are not assigned to the judge, when they click those days, they would also want to see the hearings for that day whether they were assigned to them or not.

Are there better ways of handling this @Chingujo

sharonwarner: Interesting - yeah, I think that could be confusing for the user since we don't include the VLJ on each row on the daily docket. I think we either have to add VLJ to each row or only display hearings assigned to that judge.

PR #11347, opened 2019-07-09
Add RequestIssue.contention method

pkarman:
This small change is in anticipation of more advanced error checking when a request issue is edited. We'll want to fetch the contention and make sure it exists before modifying the request issue.

see #11348

### DIFF
--- a/app/models/end_product_establishment.rb
+++ b/app/models/end_product_establishment.rb
@@ -146,6 +146,8 @@ class EndProductEstablishment < ApplicationRecord
     cached ? cached_result : fetched_result
   end
 
+  alias end_product result
+
   delegate :contentions, to: :cached_result
 
   def limited_poa_on_established_claim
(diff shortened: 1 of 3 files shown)


app/models/end_product_establishment.rb
```diff
@@ -146,6 +146,8 @@ def result(cached: false)
     cached ? cached_result : fetched_result
   end
 
+  alias end_product result
```
pkarman: just a mnemonic convenience. I can never remember at the console that the EndProductEstablishment is not operating on a `result` -- I always expect it to be an `end_product`.

app/models/request_issue.rb
```diff
@@ -525,6 +525,10 @@ def contention_disposition
     end
   end
 
+  def contention
+    end_product_establishment.end_product.contentions.find { |c| c.id.to_s == contention_reference_id.to_s }
```
pkarman: I used `end_product` here since that makes more sense in context. An End Product has Contentions. A `result` is not very intuitive.

app/models/end_product_establishment.rb
leikkisa: I'd be down to rename it altogether! but perhaps another time :-P

PR #11352, opened 2019-07-10
Don't automatically require rails_helper

monfresh:
**Why**: Our "rails_helper" file contains a lot of intensive work that
heavily slows down tests. By automatically requiring this file in
`.rspec`, it unnecessarily slows down tests that don't need it.

As an example, none of our specs in `mappers` require `rails_helper`.
When we run specs the old way, the specs run in 16.77 seconds. When we
stop requiring `rails_helper`, they run in 0.03745 seconds. That's
448x faster!

The recommended best practice is to only require `rails_helper` when
it is needed. The next step is to break down the different parts of
`rails_helper` into separate files in `spec/support` so that we can
further speed up tests. For example, some of the setup in `rails_helper`
only applies to feature specs.


monfresh:
I don't think it's related. The specs passed earlier. Danger is complaining because `spec_helper` contains the line `config.filter_run focus: true`.

PR #11367, opened 2019-07-10
Remove ability for attorneys to place tasks on hold

lowellrex:
`AttorneyTask`s should not be allowed to be put on hold. This PR removes the ability to put those tasks on hold. Related slack conversation: https://dsva.slack.com/archives/C6E41RE92/p1562784114188800


spec/feature/queue/attorney_queue_spec.rb
```diff
@@ -70,89 +70,6 @@
     end
   end
```
lomky: Do we want to explicitly test AttorneyTasks can't go on hold? I'm not super strong either way, but it would be nice to have the documented do-not-want explicitly in our test cases to match our business case.

PR #11372, opened 2019-07-10
API documentation

pkarman:
This does not currently describe all endpoints in detail, but should at least mention them all. In addition it describes the API key authn mechanism.


pkarman:
@laurjpeterson historically how have API consumers communicated to use that they need an API key?

laurjpeterson:
I don't know, I wasn't part of/privy to much of our Appeals Status work, which is what I think about when you say API consumers. It's possible there are still some people on the DSVA VA.gov team (I think it was team unicorns for some reason) that would know how Chris did this in the past that we could reach out to, though.

docs/caseflow-apis.md
```diff
@@ -0,0 +1,321 @@
...
+
+#### `GET /idt/api/v1/veterans`
+
+## VETText / Hearings
+
+The Hearings API returns all the hearings scheduled for a given day.
+
+```bash
+% curl -H 'Authorization: Token your-sekrit-key' \
+       https://appeals.cf.ds.va.gov/api/v2/hearings/2019-06-07
```
ferristseng: When I tried this locally, I had to add a `token=` to the authorization header:

```
curl -H 'Authorization: Token token=<sekrit-key>' localhost:3000/api/v2/hearings/2019-07-06
```

monfresh: Our controller tests, and the Rails code that parses the authorization header indicate that both syntaxes should work.
https://github.com/rails/rails/blob/9b35b9ff417e8ec9888deac1e66c2855677164cd/actionpack/lib/action_controller/metal/http_authentication.rb#L446-L463

ferristseng: Yeah, I just tried it without it and it worked. I must have been doing something else wrong when I was trying initially!

monfresh: Here's how I confirmed it locally:
1. In the Rails console, I ran `ApiKey.create!(consumer_name: "API Consumer").key_string`
2. I then copied the string and pasted it in the `curl` command, like this:
```console
curl -H "Authorization: Token 0b30595527394e5b8544b48801fcb64c" http://localhost:3000/api/v2/hearings/2019-07-06
```
I got this response:
```
{
  "errors": [
    {
      "status":"404",
      "title":"Hearing day not found",
      "detail":"No hearing days with that date were found."
    }
  ]
}
```

PR #11396, opened 2019-07-11
do not sync record on every veteran_ssn call

pkarman:
basically reverting back to `veteran.ssn` pattern because we are trying to avoid BGS calls. Instead we'll do a one-time backfill of ssn and rely on Intake to populate that field going forward.


pkarman:
We're going to want it for HLR and SC too so leaving it this way in the parent class.

PR #11399, opened 2019-07-12
Resolves extra carriage return in response in connect-vbms

Sjones352:
Resolves FindDocumentVersionReference and GetDocumentList API issue have been down.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ gem "bgs", git: "https://github.com/department-of-veterans-affairs/ruby-bgs.git"
 gem "bootsnap", require: false
 gem "business_time", "~> 0.9.3"
 gem "caseflow", git: "https://github.com/department-of-veterans-affairs/caseflow-commons", ref: "8dde00d67b7c629e4b871f8dcb3617bfe989b3db"
-gem "connect_vbms", git: "https://github.com/department-of-veterans-affairs/connect_vbms.git", ref: "94b439475edbe6d2ac9a9676b5e58361ffac70f3"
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -13,7 +13,7 @@ gem "bgs", git: "https://github.com/department-of-veterans-affairs/ruby-bgs.git"
 gem "bootsnap", require: false
 gem "business_time", "~> 0.9.3"
 gem "caseflow", git: "https://github.com/department-of-veterans-affairs/caseflow-commons", ref: "8dde00d67b7c629e4b871f8dcb3617bfe989b3db"
-gem "connect_vbms", git: "https://github.com/department-of-veterans-affairs/connect_vbms.git", ref: "279fac8a1f605487908f315a7157184588ed9a42"
+gem "connect_vbms", git: "https://github.com/department-of-veterans-affairs/connect_vbms.git", ref: "94b439475edbe6d2ac9a9676b5e58361ffac70f3"
```
pkarman: is this the right sha? I thought we wanted c7289957fef9c3548994d98c03a6cf5f0948fbe4

Sjones352: oh man you are right. Fixed

leikkisa: ➕ 

I was thinking another potential way to fix this is getting the boundary from the headers when it's available.

pkarman: that's a good point @leikkisa -- I see the `content-type` header has a `boundary` part that should communicate that. I hadn't looked at this code before; it's a shame we rolled our own since this seems like a ruby lib like https://rubygems.org/gems/multipart-parser would have been cleaner.

PR #11406, opened 2019-07-12
Warm more BGS caches

pkarman:
Renames the job to make it inclusive of more BGS cache warming.

We are using Hearings as a proxy for timeliness, which allows us to roughly load balance our BGS data pulls in a "just in time" pattern.


pkarman:
NOTE we should coordinate the merge of this PR with the corresponding name change in appeals-deployment

PR #11407, opened 2019-07-12
Lighthouse - make mock data for hlr create

### LINKED ISSUE #11328: Lighthouse API - mock response from creating a higher level review
The route and endpoint should be set up so developers can start testing their code.  It should return mocked data, but not actually create any database entries.

erluti:
Resolves #11328

### Description
Added an endpoint and a spec for creating a High Level Review.  The response is a static mock response for now.

### Acceptance Criteria
- [x] Code compiles correctly

### Testing Plan
 - new spec for accessing endpoint and expected jsonapi response


app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  def async_status(higher_level_review)
+    #REVIEW should this be in asyncable?
+    if higher_level_review.processed?
+      :processed
+    elsif higher_level_review.canceled?
+      :canceled
+    elsif higher_level_review.attempted?
+      :attempted
+    elsif higher_level_review.submitted?
+      :submitted
```
edmangimelli: Should there be a catch-all here?

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+        id: higher_level_review.uuid,
+        attributes: {
+          status: async_status(higher_level_review)
+        }
+      }
+    }
+  end
+
+  def async_status(higher_level_review)
+    #REVIEW should this be in asyncable?
```
edmangimelli: feels like it. What do you think @pkarman ?

pkarman: I'd be open to an `asyncable_status` method as part of the `Asyncable` concern, sure.

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
pkarman: the default would be something like `not_yet_submitted`

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    if higher_level_review.processed?
+      :processed
+    elsif higher_level_review.canceled?
```
edmangimelli: these methods mostly check timestamp columns of the same name (canceled_at, etc.). Is it safe to check processed first? Can a HLR have a processed_at timestamp and then later be canceled (so then it has both a processed_at and canceled_at timestamp)? If that situation were possible, than an HLR would erroneously return processed instead of canceled.
Will all HLR's start with a submitted_at timestamp, and then, depending on what happens, will have either a processed, canceled, or attempted timestamp?

pkarman: yes you should always check `processed?` first, because that's the last status to be set.

If something is `processed?` it should never also be `canceled?`. See https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/concerns/asyncable.rb#L145

asyncable models do not have a `submitted_at` value by default. They must have the `submit_for_processing` method called on them. So HLRs when they first are created are not yet submitted. Calling `submit_for_processing` is the last step in the intake process, after the data is all validated.

config/brakeman.ignore
```diff
@@ -132,8 +132,27 @@
       "user_input": null,
       "confidence": "High",
       "note": "The application does not force use of HTTPS: `config.force_ssl` is not enabled"
+    },
```
pkarman: rather than creating exceptions for each new API controller, could we just inherit from a common concern or controller that has the null_session exemption?

spec/models/concerns/asyncable_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+describe Asyncable do
+  class TestAsyncable
+    include ActiveModel::Model
+    include Asyncable
+
+    attr_accessor :submitted_at, :attempted_at, :processed_at, :canceled_at
+
+    # access attrs as if they were activerecord columns
+    def [](column)
```
pkarman: clever!

spec/models/concerns/asyncable_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+  end
+
+  describe "#asyncable_status" do
+    it "should return :not_yet_submitted if there is no timestamps" do
+      subject = TestAsyncable.new
+      expect(subject.asyncable_status).to eq :not_yet_submitted
+    end
+    it "should return :processed when the processed timestamp is set" do
+      t = Time.zone.now
```
pkarman: a nit about variable naming: we try and avoid single letters. How about `now` instead?

config/brakeman.ignore
erluti: The plan is to have a common controller for the api controllers, I was thinking we'd do it soon.  I'll add a story to our backlog to specifically consolidate these now that we have a couple controllers.

pkarman: I'm still not clear on why you're not just inheriting from `Api::ApplicationController`

erluti:
> I'm curious why we need to mock this - is it for some development needs outside of Caseflow?
> 
> Does it need to be live in production? Could we put a `FeatureToggle` around this code and just turn it on in UAT?
@pkarman

Is UAT an environment?  We don't expect any API we add to be accessible until we add it to the Kong Gateway.  This was requested to support va.gov beginning development before the features are live.
Would you recommend using FeatureToggle here?

spec/models/concerns/asyncable_spec.rb
edmangimelli: Ha! That is pretty nifty! Never seen mixins tested like this.

Issue #10824, opened 2019-05-20
Automatically end IHP tasks after a set period of time


lowellrex:
Currently, tasks to draft informal hearing presentations (IHPs) are [only completed when the assigned VSO marks them as such](https://github.com/department-of-veterans-affairs/caseflow/blob/cb3cf31bc2e86d418fc45a4fd724adec0585215e/app/models/tasks/informal_hearing_presentation_task.rb#L9). This ticket exists to automatically mark IHP tasks complete after the deadline has passed.

## Acceptance criteria
* [x] Determine what the deadline for IHPs is (see [related Slack discussion](https://dsva.slack.com/archives/CA1QBDLEL/p1558361145003700))
* [ ] Add code to mark IHP tasks complete after the deadline has passed (#11408 undone by #11426)
* [ ] Confirm that `TaskTimer`s are being created from ☝️ this change live in production
* [ ] Backfill `TaskTimer`s for existing IHP tasks.
* [ ] Confirm that IHP tasks that were past their deadlines were closed after their `TaskTimer`s were picked up by the [job that runs at the bottom of every hour](https://github.com/department-of-veterans-affairs/appeals-deployment/blob/61570d9bc78d85d0ec7e0340362c965efa7a80cb/ansible/vars/aws-config.yml#L529).

laurjpeterson:
On 5/18, we heard that the Chairman proposed 30 days for AODs and 120 for all other cases, starting June 1, but we were going to meet with the VSOs again before they finalized it.

lomky:
Would we want these to be marked `completed` or `cancelled` if the date has passed without the task being marked completed?

pkarman:
I would think `cancelled` makes more sense. Maybe even `auto_cancelled` to indicate the "how"?

laurjpeterson:
is `auto_cancelled` already an option? i thought the closed statuses were just `completed` and `cancelled` so far. if we can add another one, that'd be great! it'd be more clear than both

lowellrex:
We will need to backfill timers for a handful of existing IHP tasks. At the time of this writing, [there are 30 IHP tasks](https://caseflow-looker.va.gov/sql/ffhnq8b5mvrqxh) that are past due. We can use the following sql statement to confirm that the addition of the task timers to the existing IHP tasks was successfully by tracking whether these 30 IHP tasks are marked complete.

```sql
select id
, type
, status
, created_at
, appeal_type
, appeal_id
from tasks
[...]

PR #11408, opened 2019-07-15
Automatically end IHP tasks after deadline

lowellrex:
Connects #10824.


spec/models/tasks/informal_hearing_presentation_task_spec.rb
```diff
@@ -75,4 +75,52 @@
...
+    end
+
+    def ballpark_seconds(seconds, cushion = 10)
+      (seconds - cushion)..(seconds + cushion)
+    end
+
+    subject do
+      expect(task.task_timers.count).to eq(1)
+      timer = task.task_timers.first
+      expect(ballpark_seconds((timer.created_at + deadline_length.days).to_i)).to include(timer.submitted_at.to_i)
```
lowellrex: There can be a second or more of lag between when the task is created and the task timer is created so I created this `ballpark_seconds` function because the only thing we care about here is that we are within the same day that we expect. I would love to use some built-in utility for this date comparison but couldn't find anything after a little bit of searching.

spec/models/tasks/informal_hearing_presentation_task_spec.rb
```diff
@@ -75,4 +75,52 @@
...
+  end
+
+  describe ".when_timer_ends" do
+    let(:task) { FactoryBot.create(:informal_hearing_presentation_task) }
+
+    subject { task.when_timer_ends }
+
+    it "appends a message to the instructions field when we automatically close an IHP task" do
+      subject
+      expect(task.reload.instructions).to eq([COPY::IHP_TASK_REACHED_DEADLINE_MESSAGE])
```
lomky: Nitpick, this technically passes whether we append _or_ overwrite the instructions.

app/models/tasks/informal_hearing_presentation_task.rb
```diff
@@ -34,4 +32,23 @@ def available_actions(user)
...
+    if open?
+      update!(
+        status: Constants.TASK_STATUSES.cancelled,
+        instructions: instructions << COPY::IHP_TASK_REACHED_DEADLINE_MESSAGE
+      )
+    end
+  end
+
+  def timer_ends_at
+    created_at + deadline_length.days
```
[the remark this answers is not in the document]
lowellrex: Yup. `created_at` is populated when we create the row in the database whereas `started_at` is populated when the status of the task [is changed to `in_progress`](https://github.com/department-of-veterans-affairs/caseflow/blob/e07ec9a6dcac89e0cfb066a548cc1c2f0cfb6400/app/models/task.rb#L455).

client/COPY.json
```diff
@@ -260,6 +260,7 @@
   "NO_SHOW_HEARING_TASK_COMPLETE_MODAL_BODY": "Marking this case as complete will release it to a judge for review. A 90 day evidence hold may still apply to this case.",
   "NO_SHOW_HEARING_TASK_RESCHEDULE_FORBIDDEN_ERROR": "Can only reschedule NoShowHearingTasks",
   "HEARING_TASK_ASSOCIATION_NOT_UNIQUE_MESSAGE": "that is not closed already exists for %s %s",
+  "IHP_TASK_REACHED_DEADLINE_MESSAGE": "InformalHearingPresentationTask closed after reaching deadline",
```
lomky: Does the phrase `InformalHearingPresentationTask` convey meaning to our users, or is that developer-side jargon?

PR #11411, opened 2019-07-15
Add may_modify BGS call

### LINKED ISSUE #10087: Intake | User not authorized error, same RO
EP creation fails with "User not authorized" error.
Alternate error message: An error occurred while establishing the claim: Claim not established; Veteran is employed by this station.

VBMS investigation reveals that a User may not perform an Intake for a veteran who is an employee in the same RO as the User.

However, the sensitivity check performed early in the intake process does not catch this case.

I have reached out the user in question to find a work around for now. The VBMS suggestion was to use a different CSS account tied to a different RO, but after checking with support, most users do not have multiple accounts.

#10030 is related alert fix. See https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3467/events/277130/
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3467/events/294187/

pkarman:
Resolves #10087 

This new BGS service call forces the BGS library to use the more strict `claimants.find_flashes` endpoint. We should call this at the start of each Intake so that we prevent users from submitting Intakes that will result in Unauthorized errors from VBMS.

The `can_access?` service is call is more permissive because it is only *reading* from the Veteran file, while `may_modify?` checks on whether user may *write* to the file.

The new user-visible error looks like:

Screen Shot 2019-07-15 at 3 05 40 PM


spec/controllers/intakes_controller_spec.rb
```diff
@@ -101,6 +101,22 @@
         expect(bgs).to have_received(:fetch_veteran_info).exactly(1).times
       end
     end
+
+    context "veteran in BGS but user may not modify" do
```
pkarman: this is the only substantive change. everything else is refactor.

app/models/intake.rb
```diff
@@ -189,15 +168,11 @@ def complete_with_status!(status)
...
-      set_veteran_accessible_error
-    elsif veteran.incident_flash?
-      self.error_code = :incident_flash
-    elsif duplicate_intake_in_progress
+    validator = IntakeStartValidator.new(intake: self)
+
+    return false unless validator.validate
+
+    if duplicate_intake_in_progress
       self.error_code = :duplicate_intake_in_progress
```
leikkisa: Is there a reason not to also move this to the validator?

app/queries/intakes_flagged_for_manager_review_query.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class IntakesFlaggedForManagerReviewQuery
```
leikkisa: Should we remove this altogether? iirc we are not showing this on the manager page anymore.

pkarman: totally down to remove it, but I think there's other code that would need to be dropped too. I will create a separate ticket/PR for that.

app/models/intake.rb
pkarman: the `@error_data` variable meant exposing that attribute outside the `Intake` class, which felt a little weird. I can take another look.

app/queries/intakes_flagged_for_manager_review_query.rb
pkarman: #11414

PR #11412, opened 2019-07-15
Add created_by and updated_by fields to Hearings

### LINKED ISSUE #11351: add created_by/updated_by to hearings model
Similar to #11260 add a `created_by_id` and `updated_by_id` columns to the `hearings` model.

ferristseng:
Resolves #11351

### Description

  * Add `created_by` and `updated_by` relationships to hearing and legacy hearing models
    * Populate them with the user in the current context before creating or updating a record
  * Fix some tests to use the user factory
  * Reset the active user after each test
  * Update the user factory to populate `vacols_uniq_id`


spec/rails_helper.rb
```diff
@@ -181,6 +180,7 @@ def read_csv(klass, date_shift)
     Timecop.return
     Fakes::BGSService.clean!
     Time.zone = @spec_time_zone
+    User.unauthenticate!
```
ferristseng: Moved the call to unauthenticate to `after(:each)` to clear the user state before the next test runs. I was running into issues where subsequent tests would have `RequestStore` populated because that value wasn't getting reset.

spec/models/hearing_day_spec.rb
```diff
@@ -77,7 +77,7 @@
 
   context "update hearing" do
     let!(:hearing_day) do
-      FactoryBot.create(:hearing_day, request_type: HearingDay::REQUEST_TYPES[:video], regional_office: "RO18")
```
tomas-nava: I think we have a pattern of using the `FactoryBot` prefix in specs. If we're changing this we should do it everywhere in a single PR for consistency.

pkarman: I think we should definitely not be using the prefix. Our spec_helper is configured that way:

9b581d4ae6 (Moncef Belyamani 2019-07-11 10:01:15 -0400  27)   # Allows us to use shorthand FactoryBot methods.
9b581d4ae6 (Moncef Belyamani 2019-07-11 10:01:15 -0400  28)   config.include FactoryBot::Syntax::Methods

+1 to doing that in a single PR

Issue #11195, opened 2019-06-24
Allow decision reviews with cleared end products to be edited


leikkisa:
### Story
Currently, if a decision review has one or more cleared end products, we prevent the user from editing that claim.  Moving forward, if a user edits issues on a cleared end product, we will add those to a new corrective 930 claim.

### Acceptance criteria
* If a user clicks "Edit in Caseflow" in VBMS for a cleared claim, they are directed to that decision review's "Edit issues" page.
* This behavior is behind a feature toggle such as "claim_review_corrections"
* Users are not prevented from going to the Edit issues page if the claim has been cleared

### Notes
- Sally and LP are confirming whether users can actually click "Edit in Caseflow" in VBMS for a cleared claim. If they can't, this ticket is just for the feature toggle and just allowing users to navigate to the Edit page by some other mechanism, the Caseflow case search page.
- Higher Level Reviews and Supplemental Claims referred to as a group are called "claim reviews" in our code
- Higher Level Reviews, Supplemental Claims, and appeals referred to as a group are called "decision reviews" in our code

PR #11413, opened 2019-07-15
Allow editing cleared claims

leikkisa:
connects #11195


spec/feature/intake/higher_level_review/edit_spec.rb
```diff
@@ -1141,6 +1141,15 @@ def click_cancel(visit_page)
         expect(page).to have_content("Issues Not Editable")
         expect(page).to have_content(Constants.INTAKE_FORM_NAMES.higher_level_review)
       end
+
+      context "when correct_claim_reviews is enabled" do
+        before { FeatureToggle.enable!(:correct_claim_reviews) }
```
aroltsch: @leikkisa We should add an `after` block to disable this feature flag. Once you enable a feature flag in one test, it will be enabled for the rest of the tests and we should always ensure each test starts with a clean environment.

Issue #11093, opened 2019-06-14
async job errors should lead with exception class name


pkarman:
To make it simpler to group and report on stuck async jobs, we should always stringify errors with the exception class name first.

pkarman:
fixed in #11418

PR #11418, opened 2019-07-16
Simplify and unify error handling

### LINKED ISSUE #10541: Catch BGS transient errors
We already catch some of the transient errors but in different context (see https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/exceptions/bgs_sync_error.rb)
We should catch it and make them transient.  
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/4751/
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3936/
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/4052/

monfresh:
**Why**: In order to make it easier to handle errors from VBMS and BGS,
we moved the logic to those gems which now provide a handy `transient?`
method we can use to determine whether or not to send the error to
Sentry.

By adding `rescue_from` clauses in `ApplicationJob` and the parent
controllers, we ensure all transient errors can be caught.

There are likely still cases where we rescue from VBMS and BGS errors
directly in other files, but to keep this PR as small as possible, those
can be addressed later if we find we are still seeing transient errors
in Sentry.

Resolves #10541

Depends on these other PRs:
https://github.com/department-of-veterans-affairs/connect_vbms/pull/227
https://github.com/department-of-veterans-affairs/ruby-bgs/pull/80


app/jobs/decision_issue_sync_job.rb
```diff
@@ -11,15 +11,12 @@ def perform(request_issue_or_effectuation)
...
       request_issue_or_effectuation.sync_decision_issues!
-    rescue Rating::NilRatingProfileListError, Rating::LockedRatingError, Rating::BackfilledRatingError => error
-      request_issue_or_effectuation.update_error!(error.class.to_s)
-      # no Raven report, just noise. This just means nothing new has happened.
     rescue Errno::ETIMEDOUT => error
-      Rails.logger.error error
       # no Raven report. We'll try again later.
+      Rails.logger.error error
     rescue StandardError => error
-      request_issue_or_effectuation.update_error!(error.to_s)
```
pkarman: the `to_s` vs `inspect` method is subtly different iirc. I think we were using `to_s` because we want the error string to always be prefixed by the class name so that it is easier to read on the /jobs page and in reports.

app/models/rating.rb
```diff
@@ -3,9 +3,23 @@
 class Rating
   include ActiveModel::Model
 
-  class NilRatingProfileListError < StandardError; end
-  class LockedRatingError < StandardError; end
-  class BackfilledRatingError < StandardError; end
+  class NilRatingProfileListError < StandardError
+    def transient?
```
pkarman: I understand why we're calling these 'transient' in order to keep them from sentry, but they aren't really transient in the sense that they are due to unknown upstream service errors. They are due to data integrity issues and should eventually resolve themselves and which we don't need to worry ourselves unnecessarily.

Could we either comment to that effect or somehow indicate that these are not transient in the same sense that unknown HTTP/network errors are?

app/services/vbms_request.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
```
pkarman: I think we're trying to keep these in the `external_api` namespace.

app/jobs/decision_issue_sync_job.rb
monfresh: In other places, we are using `inspect`, so I thought I'd make it consistent, and also simplify the code. Before, the string displayed was different depending on the type of error. If it was a `Rating` error, it would only display the class name, such as `Rating::NilRatingProfileListError`. If it was a `StandardError`, it would only display the error message.

Now, the `decision_sync_error` is consistent for all error types. It displays both the class name and the error message, such as `#<Rating::NilRatingProfileListError: error message>` and `#<StandardError: error message>`. See the corresponding spec for the diff.

app/models/rating.rb
monfresh: Yes, we can. How about a more generic `ignorable?` that can apply to all types of errors we want to ignore, whether transient or not? I think our main goal here is to identify errors we don't care about, whether they are transient or some other type of error we don't want to send to Sentry, right?

app/services/vbms_request.rb
monfresh: Fixed here: https://github.com/department-of-veterans-affairs/caseflow/pull/11418/commits/62811953249c090321acce2954768ee53b10445a

monfresh:
@pkarman Thoughts on my comments?

app/models/rating.rb
pkarman: Yes `.ignorable?` seems good.

app/jobs/decision_issue_sync_job.rb
pkarman: As long as we're consistent with #11093 then I'm cool.

pkarman:
@monfresh let me know if I missed anything and if this is ready for a final 👍 -- I think the `.ignorable?` was just at the proposal phase yes?

monfresh:
I think this is ready for final 👍. I updated `transient?` to `ignorable?` in this PR, and opened 2 new PRs in `ruby-bgs` and `connect_vbms`. Once those PRs are merged, I will update this PR to point the Gemfile to the latest `ref`.

app/jobs/task_timer_job.rb
```diff
@@ -43,7 +45,7 @@ def process(task_timer)
     # Ensure errors are sent to Sentry, but don't block the job from continuing.
     # The next time the job runs, we'll process the unprocessed task timers again.
     task_timer.update_error!(error.inspect)
-    Raven.capture_exception(error)
+    capture_exception(error: error)
```
lowellrex: Why are we removing the namespace prefix here?

monfresh: This is a method defined in `ApplicationJob` that handles errors for all jobs so that we can ignore certain errors. See `app/jobs/application_job` above.

Issue #11349, opened 2019-07-10
Fix BulkAssignModal when front-end does not have access to all tasks


lowellrex:
The `BulkAssignModal` currently [populates values](https://github.com/department-of-veterans-affairs/caseflow/blob/a97f887101a5f6148798584e853b49edcf83c05c/client/app/queue/components/BulkAssignModal.jsx#L222) in the regional office and task type dropdown menus by collecting all unique values for those fields from the set of tasks available in the global redux store.

![image](https://user-images.githubusercontent.com/32683958/60971780-ea684c80-a2f2-11e9-94c0-5c620dac5d83.png)

This approach correctly contains all unique values for the appropriate tasks unless the front-end does not have access to all tasks that should be considered. Since we are limiting the number of tasks returned to the front-end currently with [a hard-limit in `Organizations::TasksController`](https://github.com/department-of-veterans-affairs/caseflow/blob/a97f887101a5f6148798584e853b49edcf83c05c/app/controllers/organizations/tasks_controller.rb#L26) and soon with [the pagination API](https://github.com/department-of-veterans-affairs/caseflow/blob/a97f887101a5f6148798584e853b49edcf83c05c/app/models/queue_config.rb#L49) for the hearings management branch, the list of values in the regional office and task type dropdown menus may exclude some values that should otherwise be included.

Since the `BulkAssignModal` already [makes a request to the back-end](https://github.com/department-of-veterans-affairs/caseflow/blob/a97f887101a5f6148798584e853b49edcf83c05c/client/app/queue/components/BulkAssignModal.jsx#L35) for the list of users who belong to the hearing management branch, we can piggy back off that request to get the complete list of possible hearing locations and task types even when we do not have all tasks available to us on the front-end. As a bonus we can simply move the `Organizations::MembersController` since the `BulkAssignModal` is currently the only place where we make a request to that endpoint (`caseflow $> git grep members client/`).

## Acceptance criteria
* [ ] `BulkAssignModal`'s regional office and task type dropdown menus contain unique values for correct fields for all tasks

kevmo:
2

lowellrex:
Query that will return the cross-section that we want to return to the front-end (https://caseflow-looker.va.gov/sql/rmzbrxwqjskgdd).

```sql
select count(*)
, tasks.type
, closest_regional_office
from tasks
left join (
  select closest_regional_office
  , id
  , 'LegacyAppeal' as type
  from legacy_appeals
  union
[...]

PR #11421, opened 2019-07-16
Move RegionalOffice hashes to constants

lowellrex:
Related to #11349 where we want to be able to map from the regional office key (e.g. RO34) to the name of the city on the front-end without having to pass that information from the back-end.


app/models/regional_office.rb
```diff
@@ -1415,14 +117,5 @@ def ro_facility_ids_for_state(state_code)
         ro[:facility_locator_id].present? && state_code == ro[:state]
       end.pluck(:facility_locator_id).uniq
     end
-
-    def find_ro_by_facility_id(facility_id)
```
lowellrex: We don't use this function anywhere so I removed it.

app/models/regional_office.rb
```diff
@@ -1389,7 +91,7 @@ def ros_with_hearings
 
     # Returns RegionalOffice objects for each RO that has the passed station code
     def for_station(station_key)
-      [STATIONS[station_key]].flatten.map do |regional_office_key|
+      [STATIONS[station_key.to_sym]].flatten.map do |regional_office_key|
```
lowellrex: This type casting feels a little hacky but I am struggling to think of a better way to handle this. Open to suggestions!

pkarman: I find your approach readable and can't think of a better one. Strings and symbols in Ruby are notoriously annoying.

app/models/regional_office.rb
```diff
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
-# rubocop:disable Metrics/ClassLength
+def convert_top_level_key_to_string(hash)
+  hash.keys.map { |key| [key.to_s, hash[key]] }.to_h
+end
```
lowellrex: These hashes were previously formatted with the outermost key as a string, and the nested hashes' keys as symbols. This functions transforms the input consistently formatted nested hashes to match the previous formatting.

app/models/regional_office.rb
lomaxap: ☝️ I'll be using this in a forthcoming PR

lowellrex: Ahh gotcha. Putting it back.

PR #11422, opened 2019-07-16
Add IDT examples

pkarman:
Adds curl examples for IDT API.


docs/caseflow-apis.md
```diff
@@ -259,20 +259,49 @@ On error returns a `400` code with an error message like:
...
+% curl -H 'TOKEN: your-idt-token' \
+       -H 'FILENUMBER: veteran-file-number' \
+       https://appeals.cf.ds.va.gov/idt/api/v1/appeals
+```
+
 #### `GET /idt/api/v1/appeals/:appeal_id`
 
+```bash
+% curl -H 'TOKEN: your-idt-token' \
+       https://appeals.cf.ds.va.gov/idt/api/v1/appeals?appeal_id=a-long-appeal-uuid
```
aroltsch: I think it should be `https://appeals.cf.ds.va.gov/idt/api/v1/appeals/uuid_or_vacols_id` (a slash instead of a question mark)

PR #11424, opened 2019-07-16
Fix BulkAssignModal when front-end does not have access to all tasks

### LINKED ISSUE #11349: Fix BulkAssignModal when front-end does not have access to all tasks

lowellrex:
Resolves #11349. Populates dropdown menus by response from API request instead of plucking off fields from tasks in the global Redux store.


client/app/queue/components/BulkAssignModal.jsx
```diff
@@ -32,11 +36,17 @@ class BulkAssignModal extends React.PureComponent {
...
-    ApiUtil.get(`/organizations/${this.organizationUrl()}/members.json`).then((resp) => {
-      this.setState({ users: resp.body.members.data });
+    this.setState({ loadingComponent: <LoadingScreen spinnerColor={LOGO_COLORS.QUEUE.ACCENT} /> });
+
+    ApiUtil.get(`/organizations/${this.organizationUrl()}/task_summary.json`).then((resp) => {
+      this.setState({
+        users: resp.body.members.data,
+        taskCountForTypeAndRegionalOffice: JSON.parse(resp.body.task_counts),
+        loadingComponent: null
+      });
```
lowellrex: Heart of the change starts here where we make the request for information to populate the dropdowns from the back-end.

client/app/queue/components/BulkAssignModal.jsx
```diff
@@ -212,18 +196,25 @@ class BulkAssignModal extends React.PureComponent {
...
-      this.generateNumberOfTaskOptions(), true)}
-  </QueueFlowModal>;
+  render = () => {
+    if (this.state.loadingComponent) {
+      return this.state.loadingComponent;
+    }
+
+    return <QueueFlowModal
+      pathAfterSubmit={`/organizations/${this.organizationUrl()}`}
+      button="Assign Tasks"
```
hschallhorn: Probably want to use `COPY::BULK_ASSIGN_BUTTON_TEXT` here

Issue #11260, opened 2019-07-01
hearing days created_by and updated_by should be FKs


pkarman:
Right now we store CSS id on the `hearing_days` model for `created_by` and `updated_by`. We should use AR associations instead.

e.g.
```
belongs_to :created_by, class_name: "User", foreign_key: "created_by_id"
```

This will require a schema change and data migration. Also, db seeds will likely need to change.

lomaxap:
Want to confirm the steps here @pkarman. HearingDay has an existing `created_by` column with that uses `css_id`. We want to replace that with a `created_by_id` foreign key column. 

1. schema change
```
add_foreign_key :hearing_days, :users, column: :created_by_id
```
2. migrate day by mapping `css_id` to `user.id` and filling `created_by_id`
3. add AR association and remove `created_by` column

pkarman:
yes that looks correct @lomaxap -- multiple PRs, with the 4th being the column drops

tomas-nava:
may need to consider changes to Tableau and Looker

PR #11430, opened 2019-07-16
use css_id for created_by and updated_by when creating HearingDays in seeds

tomas-nava:
`HearingDays` created in seeds were improperly setting `created_by` and `updated_by` to `User` objects instead of `css_id`s. Also fixed some minor linting errors.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -152,7 +152,9 @@ class SeedDB
         parent: parent_hearing_task,
         appeal: appeal
       )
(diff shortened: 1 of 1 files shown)


pkarman:
it was this bug that first drew my attention to what became #11260 -- why would we change seeds if we are going to change the models to use FKs?

tomas-nava:
I'm being deliberate in my implementation of #11260 and found it useful to have a representative "before" state in the seed data. This'll be revised in later PRs that connect to that work.

tomas-nava:
Yeah, good question. It'd be better if we did it consistently.

Issue #11206, opened 2019-06-25
Add link to decision reviews Edit page on the Case search page


leikkisa:
### Story
There is currently no way within Caseflow to navigate to the edit page for decision reviews that are processed in VBMS.  The case search page seems like a natural place to add a link.

### Acceptance criteria
1. On the case search page, for higher level reviews and supplemental claims that are listed for a veteran, add a link to the edit page for that higher level review or supplemental claim.

Related: https://github.com/department-of-veterans-affairs/caseflow/issues/10358

pkarman:
Amazed we don't have this yet! Good use case for the UUID in the URL.

PR #11431, opened 2019-07-17
added link to edit issues

Sjones352:
Connects #11206

### Description
Adds edit issues link to search page

Screen Shot 2019-07-16 at 6 36 28 PM


client/app/queue/OtherReviewsTable.jsx
```diff
@@ -63,9 +68,19 @@ class OtherReviewsTable extends React.PureComponent {
...
+            {review.claimantNames.length > 0 ? review.claimantNames.join(', ') : review.veteranFullName}
+          </span>
+          {review.processedInVbms && <Link
+            name="edit-issues"
+            href={review.editIssuesUrl}
+            target="_blank">
+            Edit issues
+          </Link>}
+        </React.Fragment>;
+      }
```
leikkisa: Can we make 'Higher Level Review' a link to the review instead of adding the extra text, down in "OTHER_REVIEWS_TABLE_REVIEW_TYPE_COLUMN_TITLE"?

leikkisa: I think we can have the link whether it's processed in VBMS or not, since it's using the UUID.

spec/feature/queue/search_spec.rb
```diff
@@ -71,6 +76,11 @@
             expect(page).to have_content(COPY::CASE_LIST_TABLE_EMPTY_TEXT)
             expect(page).to have_content(COPY::OTHER_REVIEWS_TABLE_TITLE)
           end
+
+          it "should show edit issues page" do
+            expect(page).to have_content("Edit issues")
+            click_on "Edit issues"
```
leikkisa: This looks like it's missing the next step to check that it actually navigates to the "Edit issues" screen.

client/app/queue/OtherReviewsTable.jsx
leikkisa: Let's get rid of "processedInVbms", because if it's false, this column will be blank.

PR #11436, opened 2019-07-17
Allow veterans to be scheduled on full hearing day

### LINKED ISSUE #11111: Add Veterans to the daily docket
## User Story
As a hearings coordinator, there are certain times when I need to add a veteran onto the docket at the last minute. I need certain hearings branch users to have these privileges.  The current docket no longer shows up as an available hearing day if the day is full.  

_TLDR: We need to allow hearing coordinators the ability to schedule outside of a docket limit. This proposes a search way to do this so we don't have to make major changes to the "Schedule Veterans" tables._

## Acceptance criteria:
- [ ] Verify that when a hearing coordinator searches a Veteran the schedule Veterans actions is in the schedule hearing task
- [ ] Verify that the date fields shows the list of available days for that Veteran to be scheduled 
- [ ] Verify that there is an "other" option that when selected lets them enter a custom day
- [ ] Verify that the date entered is a valid hearing day for that Veteran
- [ ] Verify a warning if that day is full with the following copy: 
     - Title: This hearing day is full
     - Body: You are about to schedule this Veteran on a full docket. Please verify before scheduling.
- [ ] Verify that user can confirm the Veteran into that daily docket
- [ ] Verify that the confirmation for schedule veteran is shown at the top of case details
- [ ] Verify that the Veteran shows in the appropriate daily docket

### Mocks
_Other is an an option in hearing date field and when selected a custom date entry appears_

_warning pops up if daily docket is full_

  

_Confirm message is the same_

ferristseng:
Resolves #11111

### Description

 * Moved some functions on `HearingDay` that took a date range as input to a separate class `HearingDayRange`.
   * Also moved some tests from the `HearingDay` spec to a new spec for `HearingDayRange`
 * Added attribute, `override_full_hearing_day_validation`, on `Hearing` to override the full hearing day validation check.
 * Some small refactors to address code climate warnings
 * Return all hearing dates for a regional office from API instead of just open ones, and rename route to `/regional_offices/:regional_office/hearing_dates` to reflect that.
 * Update UI to show filled vs. total slots in schedule veteran modal.
 * Add warning when the user selects a hearing day that is already full.

### Acceptance Criteria

- [x] Verify that when a hearing coordinator searches a Veteran the schedule Veterans actions is in the schedule hearing task
- [x] Verify that the date fields shows the list of available days for that Veteran to be scheduled 
- [x] ~~Verify that there is an "other" option that when selected lets them enter a custom day~~ Shows a list of all hearing days that are available even filled ones
- [x] ~~Verify that the date entered is a valid hearing day for that Veteran~~ Only shows a list of valid hearing days now, so this isn't necessary
- [x] Verify a warning if that day is full with the following copy: 
     - Title: This hearing day is full
     - Body: You are about to schedule this Veteran on a full docket. Please verify before scheduling.
- [x] Verify that user can confirm the Veteran into that daily docket
- [x] Verify that the confirmation for schedule veteran is shown at the top of case details
- [x] Verify that the Veteran shows in the appropriate daily docket


.reek.yml
```diff
@@ -5,7 +5,14 @@ detectors:
...
     exclude:
       - BulkTaskAssignment
+      - Hearing#override_full_hearing_day_validation
       - QueueConfig
+  BooleanParameter:
+    exclude:
+      - HearingRepository#slot_new_hearing
+  ControlParameter:
+    exclude:
+      - HearingRepository#slot_new_hearing
```
ferristseng: I think this is a false positive. Reek was warning that boolean parameters shouldn't be used to control the flow of a function, but this is being passed as a field to the model.

.reek.yml
```diff
@@ -5,7 +5,14 @@ detectors:
   Attribute:
     exclude:
       - BulkTaskAssignment
+      - Hearing#override_full_hearing_day_validation
```
ferristseng: Using this attribute to override the full hearing day validation, so I think this can be ignored.

app/services/hearing_day_range.rb
```diff
@@ -0,0 +1,152 @@
+# frozen_string_literal: true
+
+class HearingDayRange
```
ferristseng: Moved some of the hearing day methods that worked on a date range to a separate class. This helped me work through some of the code climate warnings that a couple of methods had too many arguments.

spec/feature/hearings/schedule_veteran_spec.rb
```diff
@@ -464,4 +464,48 @@ def navigate_to_schedule_veteran_modal
...
+    let!(:hearing_day) do
+      create(
+        :hearing_day,
+        scheduled_for: Time.zone.today + 5,
+        request_type: HearingDay::REQUEST_TYPES[:video],
+        regional_office: "RO17"
+      )
+    end
+    let!(:hearings) do
+      (1...13).map do |idx|
```
tomas-nava:
```suggestion
      (1..hearing_day.total_slots).map do |idx|
```

future proofing...

spec/repositories/hearing_repository_spec.rb
```diff
@@ -23,6 +23,43 @@
...
         .hearing_date.to_datetime.in_time_zone("UTC").hour).to eq(9)
     end
+
+    context "for a full hearing day" do
+      before do
+        Timecop.return
+      end
+
+      let!(:hearings) do
+        (1...13).map do |idx|
```
tomas-nava:
```suggestion
        (1..hearing_day.total_slots).map do |idx|
```
future proofing...

client/app/components/DataDropdowns/HearingDate.jsx
```diff
@@ -11,12 +11,11 @@ import ApiUtil from '../../util/ApiUtil';
 import _ from 'lodash';
 import { formatDateStr, getMinutesToMilliseconds } from '../../util/DateUtil';
 import LoadingLabel from './LoadingLabel';
-
 import SearchableDropdown from '../SearchableDropdown';
 
 class HearingDateDropdown extends React.Component {
   componentDidMount() {
-    setTimeout(this.getHearingDates, 0);
```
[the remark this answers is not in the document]
ferristseng: I wasn't completely sure about this. I thought `setTimeout(..., 0)` would do nothing, but apparently there are reasons to use it: https://stackoverflow.com/questions/779379/why-is-settimeoutfn-0-sometimes-useful. I'll add these back, but not totally sure why they are there in the first place.

.reek.yml
ferristseng: Yeah, I was a little bit worried about replicating the bulk of the logic for two different, but very similar, methods. Maybe can revisit this.

client/app/components/DataDropdowns/HearingDate.jsx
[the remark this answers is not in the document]
lomaxap: this is necessary here. it's a quirk of how redux actions work async. `componentDidMount` may be called before the props on the component are set from the redux state. `setTimeout` pushes the callback to the end of the call stack.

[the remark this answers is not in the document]
lomaxap: although. this was originally written when the hearingdate dropdown was rendered multiple times on the daily docket page, pushing the callback to the end of the callback guaranteed each component had the most up to date redux state when attempting to `getHearingDates`.

You could probably get away with removing this. Although, we could no longer use multiple on a single page without doing a request for each dropdown.

.reek.yml
tomas-nava: Yeah, I prefer the way you've done it.

Issue #11385, opened 2019-07-11
MVP 930s | Allow users to "correct" request issues


laurjpeterson:
Users need to correct request issues via 930 EPs.

## AC

Before save:
- Add/remove issues page contains all of the request issues from the former EP, **but do not allow users to update the contention text, withdraw, or remove them** 
- Add the option "Correct in VBMS, via 930 EP" in the dropdown (where withdraw/remove was), which adds a new request issue that's connected to that incorrect request issue
- If a user clicks "Correct in VBMS, via 930 EP", we will add the same request issue to the 930 EP when the user clicks Establish

After save:
- The former request issue (on the 030 or 040) and new request issue (on the 930) will be connected

PR #11443, opened 2019-07-18
Correct request issue on closed claim

leikkisa:
connects #11385


app/workflows/request_issue_correction.rb
```diff
@@ -0,0 +1,68 @@
...
+class RequestIssueCorrection
+  def initialize(user:, review:, request_issues_data:)
+    @user = user
+    @review = review
+    @request_issues_data = request_issues_data
+  end
+
+  delegate :end_product_establishment_for_issue, to: :review
+
+  EXCLUDED_ATTRIBUTES = %w[
```
aroltsch: I think we should do the opposite, be specific of which attributes to copy over. If in the future someone adds a new attribute to the request issues that we don't want to copy over and doesn't update this list, we will be mistakenly copying that over.

app/workflows/request_issue_correction.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+class RequestIssueCorrection
+  def initialize(user:, review:, request_issues_data:)
+    @user = user
```
aroltsch: I don't think we need `user` in this class?

app/workflows/request_issue_correction.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  def corrected_issues_data
+    return [] unless request_issues_data
+
+    request_issues_data.select { |ri| !ri[:correction_type].nil? && ri[:request_issue_id] }
+  end
+
+  def create_correction_issue!(original_issue)
+    issue_data = corrected_issues_data.find { |ri| ri[:request_issue_id] == original_issue.id.to_s }
+    correction_type = issue_data[:correction_type]
```
aroltsch: I am afraid this could be confusing for others when looking at this approach. Whenever there is a correction to the original issue, we are sending original request issue with the correction type value but then we don't set the correction type on the original and instead create a copy of the request issue. However, whenever we are creating a new correction issue, we use that passed correction type field differently. I think this is totally fine for now though.

app/models/request_issues_update.rb
```diff
@@ -14,23 +14,22 @@ class RequestIssuesUpdate < ApplicationRecord
...
-      process_edited_issues!
+      process_issues!
       review.mark_rating_request_issues_to_reassociate!
       update!(
         before_request_issue_ids: before_issues.map(&:id),
         after_request_issue_ids: after_issues.map(&:id),
         withdrawn_request_issue_ids: withdrawn_issues.map(&:id),
-        edited_request_issue_ids: edited_issues.map(&:id)
+        edited_request_issue_ids: edited_issues.map(&:id),
+        corrected_request_issue_ids: corrected_issues.map(&:id)
```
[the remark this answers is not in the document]
leikkisa: That's a good question. I use these IDs sometimes for investigation on failures.  I also need to double check if we need them for asynchronous processing. If we do, then I need to add something to RequestIssueCorrection to be able to use these IDs to get the issues at a later time than when the update was received.  That looks like it's missing from the new version of withdrawals too.

app/workflows/request_issue_correction.rb
leikkisa: Yeah, I was torn on that, whether a new field would be more likely to be included versus excluded. I am fine to switching it to an "include" versus "exclude" version, I don't have a strong preference.

Both cases are possible, I wonder if there's a good way to prevent issues on that, like maybe I should add the list straight to RequestIssue where it's more likely to be seen?  RequestIssue is already pretty big though.

app/workflows/request_issue_correction.rb
leikkisa: Yeah, both cases have some weirdness, one more on the front end, and this more on the backend.  I am still a bit torn on this too. Maybe we can ask Peter about this.

app/models/request_issues_update.rb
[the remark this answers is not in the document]
aroltsch: I think this is okay to keep them regardless.

app/workflows/request_issue_correction.rb
aroltsch: Yes, let's leave it on the BE for now.

app/workflows/request_issue_correction.rb
aroltsch: I think it is better to "include" because if someone adds a new field, they would probably add a validation on it that it is required. So the test would fail to create a request issue.

Issue #11451, opened 2019-07-18
Remove Edit Contention Button on Unidentified issues in UAT


Sjones352:
In Intake, we do not edit text on Unidentified issues but in VBMS we instruct user to Edit issue in Caseflow. We should specify that that we do not edit-contention text on unidentified issues.

![Screen Shot 2019-07-18 at 12.58.49 PM.png](https://images.zenhubusercontent.com/5b8fddd1b2cc7b6c78ece9f9/e72eb777-85ea-407b-b05c-a0dbbd6a85c2)

I also found a bug in UAT which shows the Edit-Contention-Text on Unidentified issues. We should remove this featuretoggle on Unidentified issues.
![Screen Shot 2019-07-18 at 12.48.45 PM.png](https://images.zenhubusercontent.com/5b8fddd1b2cc7b6c78ece9f9/333e4f79-1758-463d-880d-913d6f29b3cd)

msj0nes:
@Sjones352 Is this ticket now safe to close with the merge above or is additional work needed?

Sjones352:
@msj0nes yes

PR #11452, opened 2019-07-18
Remove Edit Contention Button on Unidentified issues

Sjones352:
Connects #11451

### Description
 The edit-contention-text button is enabled for unidentified issues which should not be true because user edits the title on Intake creation. Upon further research, while this feature is enabled unidentified issues, it does not allow you to save the edited text. I tested it both in local and UAT and this button is available. However it is not available in prod because we had an issue this morning with user wanting to edit the context text on an unidentified issues.

Screen Shot 2019-07-18 at 2 32 50 PM

**Removed**
Screen Shot 2019-07-18 at 2 35 29 PM


client/app/intake/pages/addIssues.jsx
```diff
@@ -154,7 +154,8 @@ export class AddIssuesPage extends React.Component {
         <div>
           { requestIssues.map((issue) => {
             const editableContentionText = Boolean(
-              formType !== FORM_TYPES.APPEAL.key && !issue.category && !issue.ineligibleReason);
+              formType !== FORM_TYPES.APPEAL.key &&
+              !issue.category && !issue.ineligibleReason && !issue.isUnidentified);
```
pkarman: looks good -- do we need a test?

Sjones352: Makes sense. I will write one.

Sjones352: added a test

spec/feature/intake/higher_level_review/edit_spec.rb
```diff
@@ -1493,6 +1493,19 @@ def click_cancel(visit_page)
...
+
+      context "when review has unidentified issues and non-ratings" do
+        let!(:issue) do
+          create(:request_issue, :unidentified, decision_review: higher_level_review,
+                                                contested_issue_description: "This is unidentified")
+        end
+
+        scenario "do not show edit contention text on unidentified issues" do
+          visit "higher_level_reviews/#{higher_level_review.uuid}/edit"
+          expect(RequestIssue.where(contested_issue_description: "This is unidentified")).to_not be_nil
```
pkarman: this db call is unnecessary here I think. we just want to make sure the text is on the page

```ruby
expect(page).to have_content("This is unidentified")
```

PR #11459, opened 2019-07-19
Cache BGS veteran info to prevent duplicate API calls

pkarman:
Currently a `Veteran` record call to `stale_attributes?` will generate 2 identical BGS API calls to fetch the veteran info: one via `can_access?` and one via `fetch_veteran_info`.

This PR caches the first response so that there is only 1 call per BGSService object.

Furthermore, the response is cached in Redis so that other `Veteran` objects can take advantage of the local cache. Note that `can_access?` should always check per-user so it ignores the Redis cache of the Veteran record in lieu of a sensitivity check.

### DIFF
--- a/app/services/external_api/bgs_service.rb
+++ b/app/services/external_api/bgs_service.rb
@@ -7,10 +7,10 @@ class ExternalApi::BGSService
   include PowerOfAttorneyMapper
   include AddressMapper
 
-  attr_accessor :client
+  attr_reader :client
 
-  def initialize
-    @client = init_client
+  def initialize(client: init_client)
+    @client = client
 
     # These instance variables are used for caching their
     # respective requests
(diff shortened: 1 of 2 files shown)


app/services/external_api/bgs_service.rb
```diff
@@ -58,10 +58,12 @@ def fetch_veteran_info(vbms_id)
     DBService.release_db_connections
 
     @veteran_info[vbms_id] ||=
-      MetricsService.record("BGS: fetch veteran info for vbms id: #{vbms_id}",
-                            service: :bgs,
-                            name: "veteran.find_by_file_number") do
-        client.veteran.find_by_file_number(vbms_id)
+      Rails.cache.fetch(fetch_veteran_info_cache_key(vbms_id), expires_in: 24.hours) do
```
lomky: I see the previous cache was for 24 hours, but that was a cache of authorization rather than of info which may change. Is there any data we're worried may become dangerously stale of 24 hours? 

As I look through the code base, our previous art for caching veteran/claimant info is done on a specific field (address, POA), but is also set for 24 hours.

Thoughts?

app/services/external_api/bgs_service.rb
```diff
@@ -172,15 +174,25 @@ def find_address_by_participant_id(participant_id)
...
 
       MetricsService.record("BGS: can_access? (find_by_file_number): #{vbms_id}",
                             service: :bgs,
                             name: "can_access?") do
-        client.can_access?(vbms_id, true)
+        record = client.veteran.find_by_file_number(vbms_id)
+        # local memo cache for this object
+        @veteran_info[vbms_id] ||= record
+        # persist cache for other objects
+        Rails.cache.write(fetch_veteran_info_cache_key(vbms_id), record, expires_in: 24.hours)
```
lomky: I see the previous cache was for 24 hours, but that was a cache of authorization rather than of info which may change. Is there any data we're worried may become dangerously stale of 24 hours? 

As I look through the code base, our previous art for caching veteran/claimant info is done on a specific field (address, POA), but is also set for 24 hours.

Thoughts?

app/services/external_api/bgs_service.rb
lomky: Should this one also be 10 minutes?

pkarman: yes! good catch

Issue #11348, opened 2019-07-09
Verify contentions exist in VBMS before allowing editing of request issues


pkarman:
The most frequent error in our async job queue for `RequestIssuesUpdate` is the `ContentionNotFound` error. We believe this is because users are manually deleting contentions directly in VBMS, making the reference ids in Caseflow point at non-existing records.

In order to prevent those kinds of errors, we want to (a) limit manual deletion in VBMS (we have an open request for that) and (b) verify that contentions exist before a user starts to edit the request issues on a claim review.

TBD what should happen when a user tries to edit a claim review with missing contentions? Some suggestions:

* automatically mark the request issue as `removed` with a status like `contention_missing`
* visually flag the request issue in the browser with some kind of alert about the missing contention
* both?

pkarman:
@leikkisa @jimruggiero @rutvigupta-design I have started on this in #11466 

Right now I am just raising an error to prevent editing of the issues on the decision review, but this feels inadequate to me, and least useful for the user.

I'm leaning toward the easiest solution in the description of this ticket, which would be marking the troublesome issues as `removed` with a status like `contention_missing`.

Thoughts?

leikkisa:
@pkarman I think that's the best way to go.  They can add the issue again if they need to.  Also it sounds like the VBMS fix for this is in the July release, and that will prevent this more contentions from being deleted in VBMS.

pkarman:
This has been merged. Moving to Validation to make sure we stop seeing those errors.

pkarman:
Most recent incident in Sentry is from 7/25.
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/5361/events/406708/

pkarman:
No more in Sentry. Considering this fixed.

PR #11466, opened 2019-07-19
Check for contentions before allowing claim review edit

pkarman:
A common error for async `RequestIssuesUpdate` failures is when the referenced contention cannot be found. Prevent those by checking before the editing session starts.


pkarman:
I agree this is less than ideal. See #11348 for discussion. I like the idea of just closing the request issue with status `removed` which we can do both at edit time and also at job run time.

pkarman:
@leikkisa this should be ready for another pass

app/controllers/claim_review_controller.rb
```diff
@@ -3,16 +3,17 @@
 class ClaimReviewController < ApplicationController
   before_action :verify_access, :react_routed, :set_application
 
+  EDIT_ERRORS = {
+    "RequestIssue::MissingDecisionDate" => COPY::CLAIM_REVIEW_EDIT_ERROR_MISSING_DECISION_DATE,
+    "StandardError" => COPY::CLAIM_REVIEW_EDIT_ERROR_DEFAULT
+  }.freeze
+
   def edit
-    # force sync on initial edit call so that we have latest EP status.
```
pkarman: this is just a refactor since we do not raise an error on contention not found.

spec/models/request_issue_contention_spec.rb
```diff
@@ -84,8 +84,9 @@
     context "when contention does not exist" do
       let(:contention_id) { "9999" }
 
-      it "raises ContentionNotFound error" do
-        expect { subject }.to raise_error(EndProductEstablishment::ContentionNotFound)
+      it "marks request issue as removed" do
```
pkarman: this is the main change

PR #11467, opened 2019-07-19
Skip VLJ support for schedule_hearing, missing_hearing_transcripts, translation, and foia colocated tasks

### LINKED ISSUE #11113: Change routing of admin actions

hschallhorn:
Resolves #11113

### Description
This PR removed VLJ support as the intermediary for schedule_hearing, missing_hearing_transcripts, translation, and foia. These tasks will be routed automatically to their respective locations

### Acceptance Criteria
When a colocated task is created with the action
- [ ] `missing_hearing_transcripts`, a `TranscriptionTask` is created and assigned to the HearingsManagement team
- [ ] `translation`, a `TranslationTask` is created and assigned to the Translation team
- [ ] `foia`, an `FoiaTask` is created and assigned to the Privacy team
- [ ] `schedule_hearing`, the task is routed to the HearingsManagement team where it can be decided whether or not a hearing needs to be scheduled
- [ ] Upon completion of a `ScheduleHearingColocatedTask`, a `ScheduleHearingTask` and its `HearingTask` parent are created.


spec/factories/task.rb
```diff
@@ -66,6 +66,11 @@
       appeal { create(:appeal) }
     end
 
+    factory :foi_task, class: FoiaTask do
+      type { FoiaTask.name }
+      appeal { create(:appeal) }
```
monfresh: Can we simplify this to `appeal`, without the block? See notes here: https://github.com/department-of-veterans-affairs/caseflow/wiki/Testing-Best-Practices#factories

app/models/task.rb
```diff
@@ -293,6 +304,10 @@ def when_child_task_created(child_task)
     update!(status: :on_hold) if !on_hold?
   end
 
+  def task_is_assigned_to_users_organization?(user)
```
hschallhorn: Moved from GenericTask

app/models/tasks/colocated_task.rb
```diff
@@ -51,9 +51,19 @@ def verify_user_can_create!(user, parent)
...
 
+    def default_assignee
+      Colocated.singleton
+    end
+
     def find_subclass_by_action(action)
       subclasses.find { |task_class| task_class.label == Constants::CO_LOCATED_ADMIN_ACTIONS[action] }
     end
+
+    def actions_assigned_to_colocated
```
hschallhorn: All subclasses where the default assignee is Colocated

app/models/tasks/schedule_hearing_colocated_task.rb
```diff
@@ -5,8 +5,31 @@ def self.label
...
   end
 
-  def available_actions_with_conditions(core_actions)
-    core_actions = super(core_actions)
-    appeal.is_a?(LegacyAppeal) ? legacy_schedule_hearing_actions(core_actions) : core_actions
+  def self.default_assignee
+    HearingsManagement.singleton
+  end
+
+  def available_actions(user)
```
hschallhorn: Gives HearingsManagement the ability to confirm whether or not a hearing needs to be scheduled before creating a ScheduleHearingTask

spec/controllers/tasks_controller_spec.rb
```diff
@@ -682,7 +684,7 @@
 
         assert_response :success
         response_body = JSON.parse(response.body)
-        expect(response_body["tasks"].length).to eq 3
+        expect(response_body["tasks"].length).to eq 4
```
hschallhorn: Accounts for automatically created child tasks of colocated actions

spec/feature/queue/colocated_task_queue_spec.rb
```diff
@@ -248,48 +248,6 @@
     end
   end
 
-  describe "translation task for AMA appeal" do
```
hschallhorn: translation tasks skip VLJ support now

app/models/task.rb
```diff
@@ -48,7 +48,18 @@ class << self; undef_method :open; end
...
-  scope :visible_in_queue_table_view, -> { where.not(type: [TrackVeteranTask.name, TimedHoldTask.name]) }
+  scope :visible_in_queue_table_view, lambda {
+    where.not(
+      type:
+        [
+          TrackVeteranTask.name,
+          TimedHoldTask.name,
+          FoiaColocatedTask.name,
+          TranslationColocatedTask.name,
+          MissingHearingTranscriptsColocatedTask.name
```
lowellrex: Why do we want to exclude these tasks from the queue table view?

hschallhorn: Discussed this with Lauren. As they automatically create actionable children assigned to the org and are never actionable themselves, we hide them

lowellrex: Makes sense to me!

Instead of explicitly calling out which task classes are not visible in the queue table view could we make make `hide_from_queue_table_view` a class method (as well as an instance method: `self.class.hide_from_queue_table_view`) and do something like `Task.descendants.select(&:hide_from_queue_table_view).map(&:name)` to get the names of the task classes?

hschallhorn: Sound's like a plan!

lowellrex:
Is the reason why we need to make these subclasses of `ColocatedTask` because the payload the front-end sends [includes the type of task](https://github.com/department-of-veterans-affairs/caseflow/blob/d9fefcbbc24577b27114d85add7ff3e52e85152e/client/app/queue/AddColocatedTaskView.jsx#L82) we want to create and that the back-end always [tells the front-end to create `ColocatedTask`s](https://github.com/department-of-veterans-affairs/caseflow/blob/d9fefcbbc24577b27114d85add7ff3e52e85152e/app/repositories/task_action_repository.rb#L193)? Are there other constraints that I am missing?

I am wondering if we could reduce the complexity of having to jump through the `ColocatedTask` just to create a different type of task by just sending the task type along with the action so that we just create the desired task type directly (we'd need to modify the `AddColocatedTaskView` slightly to make this work as well). Perhaps something like:

```ruby
# app/repositories/task_action_repository.rb
def legacy_and_colocated_task_add_admin_action_data(_task, _user)
  {
    redirect_after: "/queue",
    selected: nil,
    options: Constants::CO_LOCATED_ADMIN_ACTIONS.map do |key, value|
      {
        label: value,
        value: key,
        type: ColocatedTask.name # Or whatever the type we want to create
      }
    end
  }
end
```

If we go that route it might be helpful to have a helper method that gets us all of the `ColocatedTask` and non-`ColocatedTask` admin actions that we want to be available in the attorneys' "Add admin action" view. We could also reuse such a method when determining [when to charge legacy cases](https://github.com/department-of-veterans-affairs/caseflow/blob/d9fefcbbc24577b27114d85add7ff3e52e85152e/app/models/tasks/colocated_task.rb#L186) back to the correct VACOLS location.

```ruby
def all_admin_actions
  ColocatedTask.descendants.map(&:name) + [FoiaTask.name] + [TranscriptionTask.name] # ...
end
```

hschallhorn:
RE:

> I am wondering if we could reduce the complexity of having to jump through the ColocatedTask just to create a different type of task by just sending the task type along with the action so that we just create the desired task type directly

We opted not to go this route as we'd miss out on `ColocatedTask` logic such as

1. The tracking of on hold tasks in attorney (and soon judge) queue for legacy appeals
2. The updating of a legacy appeal's vacols location
3. Skipping `verify_user_can_update` and requiring a parent task (which a legacy appeal would not have)

app/models/tasks/foia_colocated_task.rb
```diff
@@ -1,7 +1,45 @@
...
+  end
+
+  def create_privacy_act_task
+    FoiaTask.create!(
+      assigned_to: assigned_to,
+      assigned_by: assigned_by,
+      instructions: instructions,
+      appeal: appeal,
+      parent: self
+    )
```
lowellrex: Can we create a `PrivacyActTask` here instead? How are `FoiaTask` and `PrivacyActTask` different?

hschallhorn: LP and I talked and she said that we needed a different task (other than just a `PrivacyActTask`) so the privacy team know that this is FOIA specific

app/models/task.rb
```diff
@@ -239,6 +243,10 @@ def hide_from_queue_table_view
     false
   end
 
+  def self.hide_from_queue_table_view
+    false
+  end
```
lowellrex: If we change the instance method to be the following, can we remove the instance method definition from all of the subclasses?

```ruby
def hide_from_queue_table_view
  self.class.hide_from_queue_table_view
end
```

hschallhorn: Yes. Yes yes. Forgot to go back and do this.

app/models/tasks/colocated_task.rb
```diff
@@ -173,7 +183,10 @@ def vacols_location
   def schedule_hearing_vacols_location
     # Return to attorney if the task is cancelled. For instance, if the VLJ support staff sees that the hearing was
     # actually held.
-    return assigned_by.vacols_uniq_id if children.all? { |child| child.status == Constants.TASK_STATUSES.cancelled }
+    if (children.present? && children.all? { |child| child.status == Constants.TASK_STATUSES.cancelled }) ||
+       status == Constants.TASK_STATUSES.cancelled
```
lowellrex: We need this additional condition because `ScheduleHearingColocatedTask` might not have children tasks?

hschallhorn: 🎯 and [].all? returns true

app/models/tasks/transcription_task.rb
```diff
@@ -3,10 +3,12 @@
 class TranscriptionTask < GenericTask
   before_create :check_parent_type
 
-  class NonDispositionTaskParent < StandardError; end
+  class NonDispositionTaskOrMissingHearingTranscriptsParent < StandardError; end
 
   def check_parent_type
-    fail NonDispositionTaskParent unless parent.is_a? AssignHearingDispositionTask
+    unless parent.is_a?(AssignHearingDispositionTask) || parent.is_a?(MissingHearingTranscriptsColocatedTask)
+      fail NonDispositionTaskOrMissingHearingTranscriptsParent
```
lowellrex: Might it make sense to use the existing `Caseflow::Error::InvalidParentTask` class here?

app/models/tasks/missing_hearing_transcripts_colocated_task.rb
```diff
@@ -1,7 +1,41 @@
...
 
 class MissingHearingTranscriptsColocatedTask < ColocatedTask
+  after_create :create_transcription_task
+
   def self.label
     Constants.CO_LOCATED_ADMIN_ACTIONS.missing_hearing_transcripts
   end
+
+  def self.default_assignee
+    HearingsManagement.singleton
```
lowellrex: Should this go to `TranscriptionTeam.singleton`?

hschallhorn: Original ticket says HearingsManagement, but we can clarify with @laurjpeterson

db/seeds.rb
```diff
@@ -847,7 +847,7 @@ def create_task_at_judge_review(appeal, judge, attorney)
     FactoryBot.create(:attorney_case_review, task_id: child.id)
   end
 
-  def create_task_at_colocated(appeal, judge, attorney, trait = Constants::CO_LOCATED_ADMIN_ACTIONS.keys.sample.to_sym)
+  def create_task_at_colocated(appeal, judge, attorney, trait = ColocatedTask.actions_assigned_to_colocated.sample.to_sym.to_sym)
```
lowellrex: Why do we need to chain `.to_sym` twice?

PR #11470, opened 2019-07-19
Speed up test suite

monfresh:
**How**: Our biggest bottleneck in specs is the VACOLS-related DB
cleaning in `database_cleaner.rb`. Since only specific tests need to
interact with VACOLS, it makes sense to only perform the DB cleaning
for those tests.

- Create separate `database_cleaner` and `vacols_database_cleaner`
files that only clean the Postgres or both Postgres and Oracle DB, respectively.
- Don't automatically require these 2 files in `rails_helper`. Instead,
require one or the other as needed in each spec.
- When the entire suite is run, once it encounters a `require` statement
to load `database_cleaner` or `vacols_database_cleaner`, it will run the
DB cleaning in both files for all subsequent tests. To avoid that, we
additionally specify that the Postgres cleaning should only happen in
specs that are tagged with `:postgres`, and the Oracle cleaning in tests
tagged with `:vacols`.
- Move the loading of the `Vftypes`, `Issref`, and `Actcode` VACOLS
tables out of a `before(:all)` block and into a separate rake task
since this only needs to run once for the lifetime of the test DB. We
don't want to put it in a `before(:suite)` block because that will slow
down running individual tests locally.
- Move other helper methods out of `rails_helper` into separate files
for better organization


.circleci/config.yml
```diff
@@ -86,11 +86,16 @@ jobs:
...
+          name: FACOLS test setup
+          command: |
+            ~/project/ci-bin/capture-log "bundle exec rake spec:setup_vacols"
+
       - run:
           name: RSpec
           command: |
             mkdir -p ~/test-results/rspec
-            testfiles=$(circleci tests glob "spec/**/*spec.rb" | circleci tests split --split-by timings)
+            testfiles=$(circleci tests glob "spec/**/*spec.rb" | circleci tests split --split-by=timings)
```
monfresh: I'm not sure if this makes a difference, but the documentation includes the `=` sign.

spec/controllers/api/v2/hearings_controller_spec.rb
```diff
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
+require "support/vacols_database_cleaner"
 require "rails_helper"
 
-RSpec.describe Api::V2::HearingsController, type: :controller do
+RSpec.describe Api::V2::HearingsController, :vacols, type: :controller do
```
aroltsch: I know it is a lot to change but I would rename it to be `postgres_and_vacols` to make it more clear

PR #11472, opened 2019-07-19
Lighthouse pb higher level review show

### LINKED ISSUE #11471: Lighthouse API - expand docs to have a show route for Higher Level Reviews
We need a new section in the openapi specs to describe the "show" route for a Higher Level Review.  This should give some meaningful details to a user who's intake has been processed and is waiting on the issues to be reviewed.

erluti:
Resolves #11471

### Description
Expanding OpenAPI spec for the show route on Higher Level Reviews.

### Acceptance Criteria
- [x] Code compiles correctly
- [x] Endpoint is described in openapi spec 

### Testing Plan
- includes test in docs route to be sure the doc for this endpoint is available


edmangimelli:
looks good!
personally, i would replace line 984 with
```
enum [IntakeStatus]
```
instead of the description

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -583,6 +583,134 @@ paths:
                 properties:
                   errors:
                     $ref: "#/components/schemas/errors"
+  /higher_level_reviews/{uuid}:
```
pkarman: all decision reviews will probably have the same request/response definition. Does it make sense to just have a single endpoint and return the decision review "type" as an attribute in the response? The UUID is unique by definition, so we don't need to worry that 2 decision reviews of different types will collide. Thoughts?

erluti: I think it makes technical sense.  I was thinking from the user's perspective, having a POST/GET parallel is easier to use.  I try to hold off on abstractions until I've actually started repeating myself, so I'd prefer to revisit this idea than code for it initially.

Issue #11386, opened 2019-07-11
MVP 930s | Users can add missing request issues/contentions so they can go on an 930 EP


laurjpeterson:
Users might need to add missing contentions/request issues to a cleared claim, and they will go on the 930 EP

Note: This is back end work, on `request_issues_update` and `end_product_establishment`

## AC
- Users can add issues, using the same front end mechanism we have today, which will then go onto a 930 EP
- User clicks save

PR #11474, opened 2019-07-20
Add new correction issue

leikkisa:
connects #11386

Allows adding rating and nonrating correction issues. Does not yet allow unidentified correction issues (do we want that?)


app/models/claim_review.rb
```diff
@@ -123,8 +123,12 @@ def sync_end_product_establishments!
...
   end
 
-  def cleared_eps
-    end_product_establishments.select { |ep| ep.status_cleared?(sync: true) }.map(&:code)
+  def cleared_end_products
+    end_product_establishments.select { |ep| ep.status_cleared?(sync: true) }
+  end
+
+  def cleared_end_product_types
+    cleared_end_products.map{ |ep| ep.rating? ? "rating" : "nonrating" }.uniq
```
aroltsch: Instead of hard coding these values, I would create two separate boolean methods: `contain_cleared_rating_end_product?` and `contain_cleared_nonrating_end_product?`. I think it would be easier to understand in the future.

app/models/contestable_issue.rb
```diff
@@ -38,7 +39,8 @@ def from_decision_issue(decision_issue, contesting_decision_review)
         decision_issue: decision_issue,
         source_request_issues: decision_issue.request_issues.active,
         source_decision_review: decision_issue.decision_review,
-        contesting_decision_review: contesting_decision_review
+        contesting_decision_review: contesting_decision_review,
+        issue_type: decision_issue.issue_type
```
aroltsch: what if we also have a boolean here? 
`rating: true`
`rating: false`

app/models/claim_review.rb
```diff
@@ -214,6 +211,18 @@ def end_product_establishment_for_issue(issue)
 
   private
 
+  def cleared_end_products
+    end_product_establishments.select { |ep| ep.status_cleared?(sync: true) }
```
aroltsch: Can we use memoization here so we don't make another call to the DB when we call this method for the second time? 
```@cleared_end_products ||= end_product_establishments.select { |ep| ep.status_cleared?(sync: true) }```

File: client/app/intake/components/UntimelyExemptionModal.jsx [hunk 4]
```diff
@@ -21,33 +21,38 @@ class UntimelyExemptionModal extends React.Component {
   onAddIssue = () => {
     const currentIssueData = this.props.intakeData.currentIssueAndNotes;
     const currentIssue = currentIssueData.currentIssue;
+    const isRatingCorrection = currentIssue.isRating && this.props.intakeData.hasClearedRatingEp;
```
aroltsch: It would be nice to move this logic somewhere so it can be reused. I can see the same logic is applied in LegacyOptInModal.jsx. Not too big of a deal though.

PR #11479, opened 2019-07-22
Refactor Error Handling for VaDotGovService

lomaxap:
`VaDotGovService` provides an interface to 2 `va.gov` API services

1. Vet360 - validates an address, returning standardized postal addresses and a geolocation
2. Facilities API - returns data related to VA facilities and, if lat/long coordinates are provided, distances to each facility

`VaDotGovAddressValidator` implements logic for finding and storing a veteran's closest regional office as well as distances to VA hearing facilities. 

Previously, application flow was controlled through raising and handling a series of exceptions. In addition to being [slower](https://www.honeybadger.io/blog/benchmarking-exceptions-in-ruby-yep-theyre-slow/), this approach made the code hard to follow and hard to add additional steps to the logic.

This PR refactors `VaDotGovService`/`VaDotGovAddressValidator` in the following ways:
- Moves all response handling to helpers classes `FacilitiesResponse` and `AddressValidationResponse`
- All public methods on `VaDotGovService` return a hash with an `error` property defined
  - `get_distance` - `{ error: Caseflow::Error, facilities: [] }`
  - `validate_address` - `{ error: Caseflow::Error, valid_address: {} }`
- Adds `ErrorHandler` to `VaDotGovAddressValidator` that handles creating admin actions for appeals where a closest regional office can't be found
- Adds a `Validations` module that checks for errors in results for `va.gov`


monfresh:
I like the idea behind this very much. I have a few initial comments, and I'm happy to pair with you next week to go over them in more detail:

- To make it easier and cleaner to test, we could use dependency injection in `VADotGovService` as opposed to using `Fakes::VADotGovService`, which feels like cheating. The real external service uses a private method `send_va_dot_gov_request`, and we are treating that like a public method by using `Fakes::VADotGovService`, which allows us to do things like this:
```ruby
allow(VADotGovService).to receive(:send_va_dot_gov_request)
  .and_return(HTTPI::Response.new(error_code, {}, {}.to_json))
```
I'm not a big fan of this approach because it ties us to the implementation. If we end up using something other than `HTTPI` in the future, we'll have to change our tests. Instead, I recommend using dependency injection so that we can pass in our own fake service, similar to `Fakes::VADotGovService`, except it would be much simpler and would only have methods like `body` and `code`.

If the external service somehow stops returning an object that responds to `body` and `code`, we would only need to make the change in one place, as opposed to in every test where we perform the stub above.

- The Reek complaint about the long parameter list is valid in this case. It points to a [Data Clump](https://refactoring.guru/smells/data-clumps) code smell, which we can fix by moving all that address data into a first class object called `VeteranAddress` for example. Then, instead of passing in all the address parameters individually from method to method, we can pass in an instance of the `VeteranAddress` class.

- The `AddressValidationResponse` class is not being consistent with error handling. If the response body contains some messages, then calling `error` on the `AddressValidationResponse` object will return an error class, but for certain response codes, like 429, it will immediately raise an exception.

- We could go a step further and not raise any exceptions at all, at least not for the expected validation errors. Since we expect that sometimes there will be validation errors, those should not be considered exceptions because they are expected to happen. The only place I can see where we need to validate the veteran address is during Intake. How about we move that closer to the controller, which would call a class that performs the address validation, and return to the controller the result without raising exceptions? Here is an example:
https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/case_reviews_controller.rb#L13-L23

app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -39,6 +39,7 @@ def perform
       begin
         geomatch_result = appeal.va_dot_gov_address_validator.update_closest_ro_and_ahls
         record_geomatched_appeal(appeal.external_id, geomatch_result[:status])
+        sleep 1
```
monfresh: Did you mean to leave this in here?

lomaxap: yes. This is here to ensure we don't go over the request limit for both endpoints

lib/caseflow/error.rb
```diff
@@ -29,6 +29,7 @@ class VaDotGovRequestError < VaDotGovAPIError; end
   class VaDotGovServerError < VaDotGovAPIError; end
   class VaDotGovLimitError < VaDotGovAPIError; end
   class VaDotGovAddressCouldNotBeFoundError < VaDotGovAPIError; end
+  class VaDotGovMissingFacilityError < VaDotGovAPIError; end
   class VaDotGovInvalidInputError < VaDotGovAPIError; end
   class VaDotGovMultipleAddressError < VaDotGovAPIError; end
   class VaDotGovNullAddressError < StandardError; end
```
monfresh: I don't think this error is used anymore, right?

lomaxap: ended up adding an exception for missing facility since it's pretty important to know when that happens.

app/services/va_dot_gov_address_validator/error_handler.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def initialize(appeal:, appellant_address:)
+    @appeal = appeal
+    @appellant_address = appellant_address
+  end
+
+  def handle(error)
+    if check_for_philippines_and_maybe_update
+      :philippines_exception
+    elsif verify_address_errors.any? { |klass| error.instance_of?(klass) }
```
monfresh: could we do `verify_address_errors.include?(error.class)`? It's about 1.5 times faster.

app/services/va_dot_gov_address_validator/error_handler.rb
```diff
@@ -0,0 +1,70 @@
...
+      create_admin_action_for_schedule_hearing_task(
+        instructions: "The appellant's address in VBMS does not exist, is incomplete, or is ambiguous.",
+        admin_action_type: HearingAdminActionVerifyAddressTask
+      )
+
+      :created_verify_address_admin_action
+    elsif foreign_veteran_errors.any? { |klass| error.instance_of?(klass) }
```
monfresh: Same comment here about using include.

app/services/va_dot_gov_address_validator/validations.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+module VaDotGovAddressValidator::Validations
```
monfresh: Will this module be included in other classes in the future? Right now, it looks like it's only included in `VaDotGovAddressValidator`. If it won't be included in other classes, then we should think about refactoring `VaDotGovAddressValidator` so that it's not doing so much. Moving code out of a class into a module, only to include the module back in the class is like shoving dirt under the rug :) It makes the class look smaller and neater, but it's not. We can address this later though.

app/services/va_dot_gov_address_validator/error_handler.rb
lomaxap: will that work? The errors i'm looping through are instances, so I need to evaluate with instance_of?

app/services/va_dot_gov_address_validator/validations.rb
lomaxap: lol, fair enough. Yea, I think I could split this up in a future PR.

spec/services/va_dot_gov_address_validator_spec.rb
```diff
@@ -0,0 +1,258 @@
...
+        context "and veteran's country is Philippines", skip: "fails CircleCi" do
+          before do
+            # this mocks get_facility_data call for ErrorHandler#check_for_philippines_and_maybe_update
+            philippines_response = ExternalApi::VADotGovService::FacilitiesResponse.new(mock_response)
+            allow(philippines_response).to receive(:data).and_return([mock_facility_data(id: "vba_358")])
+            allow(philippines_response).to receive(:error).and_return(nil)
+            allow(ExternalApi::VADotGovService).to receive(:get_facility_data)
+              .and_return(philippines_response)
+
+            Fakes::BGSService.address_records = Hash[appeal.veteran_file_number, { cntry_nm: "PHILIPPINES" }]
```
monfresh: Did you try `allow_any_instance_of(Fakes::BGSService).to receive(:find_address_by_participant_id).and_return` to see if that works better in Circle CI? Just a wild guess, but perhaps this is failing in Circle CI because `fetch_bgs_record` stores the data in Redis, and it might not be cleared.

app/services/va_dot_gov_address_validator/error_handler.rb
monfresh: I think it should work. when you call `.class` on the instance, you will get the class, and `verify_address_errors` is an array of classes.

Issue #11191, opened 2019-06-24
MVP 930s| Establish 930 EP + success page


laurjpeterson:
Part of the MVP flow for creating 930 EPs epic

## AC
- When user clicks Establish EP, they navigate to the confirmation page that explains they established a Control 930 EP in VBMS and they can make further corrections in VBMS
    - Note: we propose we make them all Control 930 EPs for the MVP
- 930 EP with the claim label from the original EP (030 or 040) is established in VBMS 
- Use confirmation page design + copy changes below

## Copy
Heading: **Establish 930 EP**
Body: You are now creating a [930 EP + claim label] in VBMS.
   - list the request issues that were added to the Control 930 EP

leikkisa:
Example copy:
A Higher-Level Review Rating Control 930 EP is being established

leikkisa:
control, national_quality_error, local_quality_error

PR #11486, opened 2019-07-23
Add 930s to confirmation page

leikkisa:
connects #11191


app/controllers/appeals_controller.rb
```diff
@@ -114,8 +114,8 @@ def update
       set_flash_success_message
 
       render json: {
-        issuesBefore: request_issues_update.before_issues.map(&:ui_hash),
-        issuesAfter: request_issues_update.after_issues.map(&:ui_hash),
+        beforeIssues: request_issues_update.before_issues.map(&:ui_hash),
+        afterIssues: request_issues_update.after_issues.map(&:ui_hash),
```
leikkisa: I renamed these to make them more consistent with elsewhere because it was tripping me up a little

client/app/intake/pages/decisionReviewEditCompleted.jsx
```diff
@@ -17,7 +18,7 @@ const leadMessageList = ({ veteran, formName, requestIssues, addedIssues }) => {
   const editMessage = () => {
     if (requestIssues.length === 0) {
       return 'removed';
-    } else if (_.every(addedIssues, (ri) => ri.withdrawalPending)) {
+    } else if (_.every(addedIssues, (ri) => ri.withdrawalPending || ri.withdrawalDate)) {
```
leikkisa: This was a bug where it wouldn't consider a review withdrawn if a review had been partially withdrawn previously.  It still doesn't fully account for ineligible issues, but this is an improvement.

client/app/intake/reducers/appeal.js
```diff
@@ -72,7 +73,7 @@ const updateFromServerIntake = (state, serverIntake) => {
     },
     veteranInvalidFields: {
       $set: {
-        veteranMissingFields: serverIntake.veteranInvalidFields.veteran_missing_fields.join(', '),
+        veteranMissingFields: _.join(serverIntake.veteranInvalidFields.veteran_missing_fields, ', '),
```
leikkisa: This is fixing a bug that affects hot reload on the front end.

client/app/intake/pages/decisionReviewEditCompleted.jsx
```diff
@@ -53,108 +54,64 @@ const leadMessageList = ({ veteran, formName, requestIssues, addedIssues }) => {
...
-      } else if (hasWithdrawnIssues) {
+      } else if (_.every(addedIssues, (ri) => ri.withdrawalPending || ri.withdrawalDate)) {
         return 'Review Withdrawn';
       }
 
       return 'Claim Issues Saved';
     };
 
+    const beforeEps = _.uniq(_.map(beforeIssues, 'endProductCode'));
+    const afterEps = _.uniq(_.map(afterIssues, 'endProductCode'));
```
leikkisa: I think right now this still includes withdrawn issues, so need to fix that.

spec/feature/intake/end_product_correction_spec.rb
```diff
@@ -45,12 +45,30 @@
     )
   end
 
+  let!(:test1) do
```
leikkisa: Oops, I'll take these out and write real tests. This was just for me.

app/models/request_issues_update.rb
```diff
@@ -90,7 +90,11 @@ def after_issues
   end
 
   def edited_issues
-    @edited_issues ||= edited_request_issue_ids ? fetch_edited_issues : calculate_edited_issues
+    edited_request_issue_ids ? fetch_edited_issues : calculate_edited_issues
```
aroltsch: Why did you remove the memoization part?

app/controllers/claim_review_controller.rb
```diff
@@ -60,14 +60,15 @@ def render_success
...
+                     afterIssues: request_issues_update.after_issues.map(&:ui_hash),
                      withdrawnIssues: request_issues_update.withdrawn_issues.map(&:ui_hash) }
     else
       render json: {
         redirect_to: nil,
-        issuesBefore: request_issues_update.before_issues.map(&:ui_hash),
-        issuesAfter: request_issues_update.after_issues.map(&:ui_hash),
+        beforeIssues: request_issues_update.before_issues.map(&:ui_hash),
+        afterIssues: request_issues_update.after_issues.map(&:ui_hash),
+        updatedIssues: request_issues_update.all_updated_issues.map(&:ui_hash),
```
aroltsch: Confused why you need all issues instead of just sending back correction issues?

app/models/request_issues_update.rb
```diff
@@ -191,7 +195,7 @@ def process_removed_issues!
 
   def correction
     @correction ||= RequestIssueCorrection.new(
-      review: review,
+      request_issues_update: self,
```
aroltsch: you could also pass `corrected_request_issue_ids` and `review` instead of passing the entire thing.

app/controllers/claim_review_controller.rb
leikkisa: On the confirmation page we need to know which EPs have changed, and what issues are on them.  "before" and "after" issues both include issues that haven't changed, and also don't include correction issues.  So updatedIssues includes the correction issues, removes "unchanged" issues, and also has their EP code data.

This is a refactor from what was there, due to a combination of there being bugs in the previous code :-/, and wanting to generalize the functionality to handle all of the EP types (before the confirmation page only grouped issues by rating/nonrating and didn't care about EP code).

PR #11487, opened 2019-07-23
Lighthouse - Higher Level Review happy path

edmangimelli:
### Description
A first implementation of the "create a Higher Level Review" api endpoint.

### Acceptance Criteria
- [ ] Tests pass


app/models/lighthouse_higher_level_review_request_deserialized.rb
```diff
@@ -0,0 +1,288 @@
...
+  #     nca
+  #   ]
+
+  # SUBCLASSES
+
+  class Veteran
+    attr_reader :file_number
+
+    def initialize(file_number)
+      fail unless file_number =~ /^\d{8,9}$/
```
erluti: instead of using `fail`s, raise a menaingful error.  sometimes it might make sense to create your own error class,  but here it might make sense to raise an `ArgumentError`.  Also have a matching test for it

app/models/lighthouse_higher_level_review_request_deserialized.rb
```diff
@@ -0,0 +1,288 @@
...
+    fails unless benefit_type.in? BENEFIT_TYPES.keys
+  end
+
+  def initialize_relationships(relationships)
+    @veteran = Veteran.new relationships["veteran"]["data"]["id"]
+
+    # a claimant is optional
+    id = relationships.dig("claimant", "data", "id")
+    code = relationships.dig("claimant", "data", "meta", "payeeCode")
+    fail if !!id ^ !!code # must have both or neither
```
erluti: nice, i've never needed to use an `xor`!  although the Rails helper `present?` is a little clearer than `!!` in my opinion

app/models/lighthouse_higher_level_review_request_deserialized.rb
```diff
@@ -0,0 +1,288 @@
...
+  end
+
+  class NewNonratingIssue
+    attr_reader :nonrating_issue_category
+
+    delegate :decision_text, :decision_date, to: :new_rating_issue
+
+    def initialize(options)
+      # options should have these keys: decision_text, decision_date, nonrating_issue_category, benefit_type
+      @new_rating_issue = NewRatingIssue.new options
```
erluti: should a `NewRatingIssue` be created in a `NewNonratingIssue`?

app/models/lighthouse_higher_level_review_request_deserialized.rb
```diff
@@ -0,0 +1,288 @@
...
+    arg.is_a?(TrueClass) || arg.is_a?(FalseClass)
+  end
+
+  def valid_date_format?(date)
+    # should be "YYYY-MM-DD"
+    Date.valid_date?(*(date.split("-").map { |s| Integer s }))
+  end
+
+  def initialize_attributes(attributes)
+    ATTRIBUTES.each { |k| instance_variable_set("@#{k}", attributes[k.camelize(:lower)]) }
```
erluti: I want explicit code preferred over metaprogramming.  This code leaves a lot unknown.  Are we storing some variables we aren't even using?  We need some explicit knowledge of what's in there to even know what to access.

app/models/lighthouse_higher_level_review_request_deserialized.rb
```diff
@@ -0,0 +1,288 @@
...
+  #     fiduciary
+  #     insurance
+  #     education
+  #     voc_rehab
+  #     loan_guaranty
+  #     vha
+  #     nca
+  #   ]
+
+  # SUBCLASSES
```
erluti: It's hard to evaluate the usefulness of these subclasses because you are using metaprogramming to instantiate them.  But it looks like they are just validating the input values.  I think that could simple be a method on the LighthouseHigherLevelReviewRequestDeserialized class.  But if these classes are carrying more weight then that, it's usually better to let them be their own class proper and have tests for them.  If we need a lot of these they can be in a subdirectory/namespace for them.

spec/models/lighthouse_higher_level_review_request_deserialized.rb
```diff
@@ -0,0 +1,62 @@
...
+            }
+          ]
+        }
+
+
+      )
+    end
+
+    it "should have a receipt_date" do
+      expect(subject.receipt_date).to eq("2019-07-10")
```
erluti: are we going to use this internally as a string or do we want to parse the date and test for a Date object?

app/models/lighthouse_higher_level_review_request_deserialized.rb
edmangimelli: agreed. this PR is still a WIP

app/models/lighthouse_higher_level_review_request_deserialized.rb
```diff
@@ -0,0 +1,288 @@
...
+      @new_nonrating_issue = NewNonratingIssue.new options
+      @id = Integer String options[:id]
+    end
+  end
+
+  # ALL OF THE AVAILABLE INSTANCE METHODS (not counting methods of subclasses)
+
+  attr_reader(*ATTRIBUTES, :veteran, :claimant, :issues)
+
+  def hash_for_review!
```
erluti: typically methods use `!` to signify they have some side-effeccts, but sense these don't you can drop the `!` from the method names

app/models/lighthouse_higher_level_review_request_deserialized.rb
edmangimelli: I think it's clean to have a constant describe what attributes are valid, but i'll rethink this

app/models/lighthouse_higher_level_review_request_deserialized.rb
edmangimelli: they're pretty intimately tied to their parent class, which is why I tucked them away. and, correct, they pretty much only provide validation. i started with methods, but it seemed to get ugly pretty fast. I think this organizes the validation logic, avoids hash pitfalls, and provides a clean syntax for accessing sub values
```
intake.veteran.file_number
```

app/models/lighthouse_higher_level_review_request_deserialized.rb
erluti: Using a constant to validate what attributes are valid or should be present is good.  I think the big win of Ruby is you can write code that's easy to understand and I think there's a lot of benefit to writing code that someone (especially ourselves) can just read and understand what's going on 6/12/18 months from now, and I think metaprogramming hides what's actually happening here.

spec/models/lighthouse_higher_level_review_request_deserialized.rb
edmangimelli: i'm letting the existing models decide. they're expecting a string

app/models/lighthouse_higher_level_review_request_deserialized.rb
edmangimelli: yeah, that was a conscious decision so they'd line up with the intake model methods they were meant for. yeah, maybe it's a bit deceiving

app/models/lighthouse_higher_level_review_request_deserialized.rb
erluti: Are we accessing those sub-values somewhere?  I do like that syntax, but if we aren't using it I'm worried we're adding some complexity.  Ruby style rarely supports using subclasses like this and I think if there's really a case to model these objects individually instead of using a hash, we should let them be their own class with their own tests

pkarman:
Some comments about naming:

* Can we avoid the use of the word "Lighthouse" in class names? While this API is going to be consumed by the Lighthouse project, it's a Caseflow API.
* The classes that sit in the "model" namespace aren't really models. We've been trying to reserve that space for things that are ActiveRecord models (though we haven't been consistent about it). Depending on the class, maybe `app/serializers` is more appropriate? Alternately, maybe a new space `app/models/api` could be used?

edmangimelli:
> Some comments about naming:
> 
> * Can we avoid the use of the word "Lighthouse" in class names? While this API is going to be consumed by the Lighthouse project, it's a Caseflow API.
> * The classes that sit in the "model" namespace aren't really models. We've been trying to reserve that space for things that are ActiveRecord models (though we haven't been consistent about it). Depending on the class, maybe `app/serializers` is more appropriate? Alternately, maybe a new space `app/models/api` could be used?

@pkarman, what do you think of the reorganization of the models. do they need to be wrapped in a module?

app/models/lighthouse_higher_level_review_request_deserialized.rb
edmangimelli: fixed. 👍

app/models/lighthouse_higher_level_review_request_deserialized.rb
edmangimelli: oof. i think i was tired when i wrote this. fixed. 👍

app/models/lighthouse_higher_level_review_request_deserialized.rb
edmangimelli: i like it. 👍

app/models/api/higher_level_review_request.rb
```diff
@@ -0,0 +1,211 @@
...
+#
+# examples:
+#
+#   * it ensures that a veteran file number is 8 or 9 digits,
+#     but does NOT ensure that it refers to an actual veteran file that the user has access to
+#
+#   * ensures that a date is an actual calendar date, but does not ensure that the date makes
+#     sense for where ever it's being used
+
+class HigherLevelReviewRequest
```
pkarman: These classes all seem to be validators, not models per se. If we move them to the `Api::` services space, you could create a common base class `Api::JsonValidator` with the utility class methods you have here.

Then I would keep the same class hierarchy as the AR models, to make it forward compatible.

`class Api::JsonValidator`
`class Api::DecisionReviewValidator < Api::JsonValidator` # empty at this point
`class Api::ClaimReviewValidator < Api::DecisionReviewValidator` # empty at this point
`class Api::HigherLevelReviewValidator < Api::ClaimReviewValidator` # rename `HigherLevelReviewRequest`

app/models/api/higher_level_review_request.rb
```diff
@@ -0,0 +1,211 @@
...
+
+  ISSUE_TYPES = %w[
+    rating_issue
+    nonrating_issue
+  ].freeze
+
+  ISSUE_CLASSES = %w[
+    RatingIssue
+    NonratingIssue
+    RatingIssueReference
```
pkarman: I don't understand the `*Reference` naming here.

app/models/api/higher_level_review_request.rb
```diff
@@ -0,0 +1,211 @@
...
+    NonratingIssueReference
+  ].freeze
+
+  file_to_json = lambda do |*path|
+    JSON.parse(File.read(File.join(Rails.root, *path)))
+  end
+
+  BENEFIT_TYPES = file_to_json["client", "constants", "BENEFIT_TYPES.json"]
```
pkarman: this could go in `Api::DecisionReviewValidator`

app/models/api/higher_level_review_request.rb
```diff
@@ -0,0 +1,211 @@
...
+  NONRATING_ISSUE_CATEGORIES = file_to_json["client", "constants", "ISSUE_CATEGORIES.json"]
```
pkarman: this could go in `Api::DecisionReviewValidator`

app/models/api/higher_level_review_request.rb
```diff
@@ -0,0 +1,211 @@
...
+
+  PAYEE_CODES = %w[
```
pkarman: could we create a constants file for this? seems like one should already exist.

app/models/api/higher_level_review_request/claimant.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class HigherLevelReviewRequest::Claimant
```
pkarman: how about `Api::ClaimantValidator < Api::JsonValidator`

app/models/api/higher_level_review_request/veteran.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class HigherLevelReviewRequest::Veteran
```
pkarman: `class Api::VeteranValidator < Api::JsonValidator`

app/models/api/higher_level_review_request.rb
pkarman: alternately,

`class Api::Validator::Base`
`class Api::Validator::DecisionReview`

etc, and put them in `app/services/api/validators`

app/models/api/higher_level_review_request/nonrating_issue.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+# for creating a new nonrating issue
+
+class HigherLevelReviewRequest::NonratingIssue < HigherLevelReviewRequest::RatingIssue
```
pkarman: this class hierarchy does not make sense to me. A NonratingIssue is not a kind of RatingIssue. It's a kind of RequestIssue.

app/models/api/higher_level_review_request/nonrating_issue_reference.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class HigherLevelReviewRequest::NonratingIssueReference < HigherLevelReviewRequest::Reference
```
pkarman: not sure what a Reference is

app/models/api/higher_level_review_request/rating_issue.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+# for creating a new rating issue
+
+class HigherLevelReviewRequest::RatingIssue < HigherLevelReviewRequest::Notes
```
pkarman: a RatingIssue it not a kind of Note afaik

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
```diff
@@ -4,6 +4,31 @@ class Api::V3::DecisionReview::HigherLevelReviewsController < ActionController::
   protect_from_forgery with: :null_session
 
   def create
+    begin
+      @request = HigherLevelReviewRequest.new params
```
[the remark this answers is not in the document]
edmangimelli: i thought maybe it might mess with the request method. i'll change it

app/models/lighthouse_higher_level_review_request_deserialized.rb
edmangimelli: You're absolutely right. I've strucken it. 👍

app/models/lighthouse_higher_level_review_request_deserialized.rb
edmangimelli: >Are we accessing those sub-values somewhere? 
so far just inside the classes
>I think if there's really a case to model these objects individually instead of using a hash, we should let them be their own class with their own tests
👍

app/models/api/higher_level_review_request.rb
edmangimelli: no longer applicable

edmangimelli:
@pkarman
@erluti

I think I've addressed all the issues highlighted, and thanks for iterating with me on this.

Please let me know what you think of this organization:


these error classes are meant to ensure that when error codes are used, they align with our spec
  they read the spec to create a list of errors to validate againts
  they format errors correctly for JSON:API
```
app/models/api

   DecisionReviewError
      ↓
   HigherLevelReviewError

                             HigherLevelReviewErrorCollection 
```


these are mostly intended to keep the validation needed organized,
but they also handle transforming data (from our JSON:API spec to the different hash shapes needed by the existing models)
```
app/serializers/api/v3

  main class:

    HigherLevelReviewPreintake

  sub classes:
  
    ClaimantPreintake

    VeteranPreintake

    RequestIssuePreintake
```
Do we definitely want to name these classe *Validator?
- the classes do a bit more than validation (HLR and RequestIssue do some transformation)
- they read in constants from spec and json files
- i like that preintake gives a hint where, chronologically, these classes fit in. validator doesn't immediately tell me at what point in the intake process these classes are used


```
app/serializers/api/v3/concerns

  helpers.rb

  validation.rb  <- these are small functions for validation types and values against arrays of constants
```


many tests have been written, but more need to be.

again. thanks for all of the feedback!

PR #11503, opened 2019-07-23
If any of the request issues have ratings, do not allow to edit them

### LINKED ISSUE #10571: Investigation | Contention cannot be deleted because it is connected to a rating issue
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/4669/events/310757/

Exact error message: `contention cannot be deleted because it is connected to a rating issue`

We thought this was fixed by syncing EPs at Claim Review edit time but apparently not. See #10102

pkarman:
Resolves #10571 

### Description
When a user attempts to edit a Claim Review, if any of the request issues are already decided, do not allow to edit them


client/app/intake/components/IssueList.jsx
```diff
@@ -30,27 +66,7 @@ export default class IssuesList extends React.Component {
           const editableContentionText = Boolean(formType !== FORM_TYPES.APPEAL.key &&
             !issue.category && !issue.ineligibleReason && !issue.endProductCleared && !issue.isUnidentified
           );
-          let issueActionOptions = [];
-
-          if (issue.correctionType && issue.endProductCleared) {
```
aroltsch: Moved this logic into a method above.

app/models/request_issue.rb
```diff
@@ -623,6 +624,15 @@ def contention
 
   private
 
+  # When a request issue already has a rating in VBMS, prevent user from editing it.
+  def contention_connected_to_rating?
+    return false unless contention_reference_id
+
+    matching_rating_issues.any? if end_product_establishment.associated_rating
+  rescue Rating::BgsRatingError
+    false
```
leikkisa: This could be true if the matching rating issue is locked (or possibly backfilled, but based on my understanding that wouldn't happen here).

app/jobs/decision_issue_sync_job.rb
```diff
@@ -11,7 +11,7 @@ def perform(request_issue_or_effectuation)
 
     begin
       request_issue_or_effectuation.sync_decision_issues!
-    rescue Rating::NilRatingProfileListError, Rating::LockedRatingError, Rating::BackfilledRatingError => error
+    rescue Rating::BgsRatingError => error
```
pkarman: I believe @monfresh has a WIP that will touch this too #11418

client/app/intake/components/IssueList.jsx
```diff
@@ -7,7 +7,43 @@ import Button from '../../components/Button';
...
+  fontStyle: 'Italic'
+});
+
 export default class IssuesList extends React.Component {
+
+  generateIssueActionOptions = (issue) => {
+    let options = [];
+
+    if (issue.uneditable) {
+      return options;
```
leikkisa: Should we hide the button, or put a disabled option instead if uneditable?

client/app/intake/components/IssueList.jsx
```diff
@@ -67,6 +83,10 @@ export default class IssuesList extends React.Component {
                 legacyAppeals={intakeData.legacyAppeals}
                 formType={formType} />
 
+              { _.isEmpty(issueActionOptions) && <div className="issue-action">
+                <span {...nonEditableIssueStyling}>Rating in process</span>
```
leikkisa: I like this message 👍

app/models/request_issue.rb
```diff
@@ -394,7 +394,8 @@ def ui_hash
       withdrawal_date: withdrawal_date,
       contested_issue_description: contested_issue_description,
       end_product_cleared: end_product_establishment&.status_cleared?,
-      end_product_code: end_product_code
+      end_product_code: end_product_code,
+      uneditable: contention_connected_to_rating?
```
pkarman: could we assert the boolean positively? i.e. rather than `uneditable: true` could we make it `editable: false`?

e.g.
```ruby
def editable?
  return false if contention_connected_to_rating?

  true
end
```

app/models/request_issue.rb
```diff
@@ -623,6 +624,15 @@ def contention
 
   private
 
+  # When a request issue already has a rating in VBMS, prevent user from editing it.
+  def contention_connected_to_rating?
+    return false unless contention_reference_id
+
+    matching_rating_issues.any? if end_product_establishment.associated_rating
+  rescue Rating::BgsRatingError
```
pkarman: I like this!

app/jobs/decision_issue_sync_job.rb
leikkisa: I think we still want to store the original class in error because they tell us different things.

client/app/intake/components/IssueList.jsx
aroltsch: It displays the message:
image

app/models/request_issue.rb
aroltsch: At first I implemented `editable`. Then I noticed on the FE it is easier to default to issue being editable by default unless specified (for example, when you add a new issue, you'd like to be able to remove it). I can change back to `editable` if you feel strongly about it.

app/models/request_issue.rb
aroltsch: Not very familiar with these types, are you saying the following:
NilRatingProfileListError => false 
LockedRatingError => true
BackfilledRatingError => true

leikkisa: I guess in the situations where we don't know, we could default to not allowing the contention to be edited.  However, in the case of a LockedRating, we won't get that error anymore, unless we change the call to BGS.

If we default to allowing the contention to be edited (easiest option), I'm wondering what the downside is. In that case, our request issue would be marked as removed, but the contention won't get removed in VBMS.  And then we won't sync it with decision issues later.

client/app/intake/components/IssueList.jsx
leikkisa: Thanks this is helpful! I was imagining an action button with no options, but this looks good.

app/models/request_issue.rb
leikkisa: For the locked rating and backfilled rating errors, we won't really know if this issue is on a "rating in progress", so maybe we should change that message?

For the locked rating thing, we'd have to change the BGS service to get the data to have allowLockedRatings set to false in order to know whether there was a locked rating.

I wonder if there's a case for catching the CannotDeleteContention error, and un-removing the request issue.

leikkisa: I can only really think of these two, I like the first one better.
Rating may be in process
Not editable

Rating in process is so much easier to communicate than "we don't know for sure, but maybe a rating is in process so we are locking edit for now" :-P.

aroltsch: I like "Rating may be in process" !

client/COPY.json
```diff
@@ -449,6 +449,7 @@
   "CLAIMANT_NOT_FOUND_END": " for assistance.",
   "INTAKE_EDIT_WITHDRAW_DATE": "Please include the date the withdrawal was requested",
   "INTAKE_WITHDRAWN_BANNER": "This review will be withdrawn. You can intake these issues as a different type of decision review, if that was requested.",
+  "INTAKE_RATING_MAY_BE_PROCESS": "Rating may be in process",
```
leikkisa: Maybe "in progress"? instead? I don't feel too strongly about it.

PR #11516, opened 2019-07-24
Avoid using serializer in hearing dates route

ferristseng:
### Description

  * Removes use of serializer from `all_hearing_days_with_hearings_hash`


app/services/hearing_day_range.rb
```diff
@@ -83,14 +83,16 @@ def all_hearing_days_with_hearings_hash(current_user_id = nil)
         all_hearings = (hearing_day.hearings || []) + (vacols_hearings_for_days[hearing_day.id.to_s] || [])
         scheduled_hearings = self.class.filter_non_scheduled_hearings(all_hearings || [])
 
-        self.class.hearing_day_hash_with_hearings(hearing_day, scheduled_hearings, current_user_id)
+        [hearing_day, scheduled_hearings]
```
lomaxap: from the name, would expect this method to return an array of `HearingDay`s. But I can't think of a quick solution to this right now, but flagging for later

PR #11523, opened 2019-07-24
Send UpdateCachedAppealAttributesJob runtime and volume metrics to Datadog

### LINKED ISSUE #11519: Instrument UpdateCachedAppealsAttributesJob
Before we turn the `UpdateCachedAppealsAttributesJob` on to run at some regular interval in the appeals-deployment repo (every 6 hours?) let's add some instrumentation [like we have in `UpdateAppellantRepresentationJob `](https://github.com/department-of-veterans-affairs/caseflow/blob/6e6aba8d30b95b49b1be4981bbed544bcd907fc9/app/jobs/update_appellant_representation_job.rb#L80) to this job so we can track how long it takes to run and how many appeals are added each run [in Datadog](https://app.datadoghq.com/metric/explorer?live=true&page=0&is_auto=false&from_ts=1556221026934&to_ts=1563997026934&tile_size=m&exp_metric=dsva_appeals.update_appellant_representation_job.runtime&exp_agg=max&exp_row_type=metric). We can use the metrics we collect to inform how often we should run this job and if we need to refactor it should the it start to run for too long.

## Acceptance criteria
* [ ] Send metrics from `UpdateCachedAppealsAttributesJob` to Datadog

kevmo:
Resolves #11519. 

The `UpdateCachedAppealAttributesJob` now sends job runtime data to Datadog. Added test coverage to job spec.


lowellrex:
Looks great! Can we add a counter for the number of appeals we are running on so we can see how the set of appeals this job is working on grows over time? Perhaps something like:

```ruby
def increment_appeal_count(count, appeal_type) # legacy vs ama
  count.times do
    DataDogService.increment_counter(
      app_name: APP_NAME,
      metric_group: METRIC_GROUP_NAME,
      metric_name: "appeals",
      attrs: {
        type: appeal_type
      }
    )
  end
end
```

app/jobs/update_cached_appeals_attributes_job.rb
```diff
@@ -55,4 +66,28 @@ def cache_legacy_appeal_vacols_data(legacy_appeals)
...
+      DataDogService.increment_counter(
+        app_name: APP_NAME,
+        metric_group: METRIC_GROUP_NAME,
+        metric_name: "appeals_to_cache",
+        attrs: {
+          type: appeal_type
+        }
+      )
+    end
+end
```
lowellrex: Let's pop this `end` in two spaces: `rubocop --auto-correct app/jobs/update_cached_appeals_attributes_job.rb`.

Issue #11083, opened 2019-06-13
Investigate request issues ineligible due to issue in active review


leikkisa:
These request issues are all listed as ineligible due to being a duplicate of an issue already in active review:
[57957, 62176, 52540, 58898, 55049, 59690, 63854, 70329]

They are all contesting different decision issues, but are all ineligible due to the same request issue (RI 5002):
[5723, 4021, 2396, 8546, 1427, 9884, 4553, 7249]

leikkisa:
I am closing this, the work on this is done, and I don't think data clean up is an option.

PR #11528, opened 2019-07-25
Do not check for duplicate issue by rating_issue_reference_id if nil

leikkisa:
connects #11083 

The eligibility check was marking issues as duplicate by rating_issue_reference_id if it was from a decision issue, but there was no duplicate decision issue and the rating issue reference id on both issues was nil.

What do we do for clean up on this?  We can't make the issues eligible again.  I guess to start it would be good to check how many issues are impacted.


app/models/request_issue.rb
```diff
@@ -833,7 +833,7 @@ def legacy_appeal_eligible_for_opt_in?
   end
 
   def check_for_active_request_issue_by_rating!
-    return unless rating?
+    return unless contested_rating_issue_reference_id
```
pkarman: how about `associated_rating_issue?` instead? that method is the same but is more meaningful to the logic.

PR #11531, opened 2019-07-25
Refactor JudgeTask to inherit from GenericTask

### LINKED ISSUE #7630: Refactor JudgeTask to inherit from GenericTask
Following the merge of #7585, JudgeTask will share a lot of similarities with GenericTask and should probably do that. Currently JudgeTask inherits from Task.

lowellrex:
Resolves #7630.

One step in the slow slog towards unifying `GenericTask` and `Task`! Only `ColocatedTask` and `AttorneyTask` remain after this (`caseflow $ git grep " < Task" app/models/`).


spec/models/tasks/judge_dispatch_return_task_spec.rb
```diff
@@ -9,7 +9,9 @@
   let(:dispatch_task) do
     create(:bva_dispatch_task, assigned_to: dispatch_user, parent: create(:root_task))
   end
-  let(:params) { { assigned_to: judge, appeal: dispatch_task.appeal, parent_id: dispatch_task.id } }
+  let(:params) do
+    { assigned_to_id: judge.id, assigned_to_type: User.name, appeal: dispatch_task.appeal, parent_id: dispatch_task.id }
```
lowellrex: We expect the `assigned_to_...` parameters instead of the `assigned_to` parameter. These tests were incorrect before.

spec/feature/queue/task_queue_spec.rb
```diff
@@ -604,7 +604,7 @@ def validate_pulac_cerullo_tasks_created(task_class, label)
     end
     let!(:decision_issue) { create(:decision_issue, decision_review: appeal, request_issues: appeal.request_issues) }
 
-    let(:root_task) { FactoryBot.create(:root_task) }
+    let(:root_task) { FactoryBot.create(:root_task, appeal: appeal) }
```
lowellrex: Previously this test was attaching tasks to two different appeals because `Task` uses the `appeal` parameter to determine the appeal while `GenericTask` uses the parent task's appeal to determine the new task's appeal. This change fixes the test.

PR #11553, opened 2019-07-26
Case distribution

pkarman:
connects #11713 

**Why**
Errors were logged but only to the Rails logs and not to Sentry. This hid the fact that we have ongoing problems. This first step is to log the full exceptions to Sentry to begin seeing patterns.

Also refactors slightly and adds more tests.

### DIFF
--- a/app/jobs/start_distribution_job.rb
+++ b/app/jobs/start_distribution_job.rb
@@ -9,6 +9,7 @@ class StartDistributionJob < ApplicationJob
     distribution.distribute!
   rescue StandardError => error
     handle_error(error)
+    # do not re-raise, since we only want to run once.
   end
 
   private
--- a/app/jobs/start_distribution_job.rb
+++ b/app/jobs/start_distribution_job.rb
@@ -8,13 +8,14 @@ class StartDistributionJob < ApplicationJob
     RequestStore.store[:current_user] = user if user
     distribution.distribute!
   rescue StandardError => error
-    Rails.logger.info "StartDistributionJob failed: #{error.message}"
-    Rails.logger.info error.backtrace.join("\n")
+    handle_error(error)
   end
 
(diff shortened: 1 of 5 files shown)


app/models/distribution.rb
```diff
@@ -20,16 +20,22 @@ class Distribution < ApplicationRecord
...
     return unless %w[pending error].include? status
 
     if status == "error"
       return unless valid?(context: :create)
     end
 
-    transaction do
-      # this might take awhile due to VACOLS, so set our timeout to 5 minutes.
-      ActiveRecord::Base.connection.execute "SET LOCAL statement_timeout = 300000"
+    multi_transaction do
```
pkarman: we update both VACOLS and Posgres in a single transaction.

app/jobs/start_distribution_job.rb
```diff
@@ -8,13 +8,15 @@ def perform(distribution, user = nil)
...
 
-  # :nocov:
-  def max_attempts
-    1
+  private
+
+  def handle_error(error)
+    Rails.logger.info "StartDistributionJob failed: #{error.message}"
+    Rails.logger.info error.backtrace.join("\n")
+    Raven.capture_exception(error)
```
pkarman: adding to Sentry

spec/factories/distribution.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :distribution do
+    association :judge, factory: :user
```
monfresh: Thank you for using this syntax!

monfresh:
As a follow-up, we can try to write a test to reproduce the scenario where the status is not being properly updated to `error` when a pending distribution runs into a failure.

PR #11555, opened 2019-07-26
Add error handling to UpdateCachedAppealsAttributesJob

kevmo:
Additionally, this PR adds the `false` flag to the array method `in_groups_of`, which is used for batching the appeals in groups of 1000, the maximum batch size for VACOLS db. The lack of this flag may have been causing errors, previously.

### DIFF
--- a/app/jobs/update_cached_appeals_attributes_job.rb
+++ b/app/jobs/update_cached_appeals_attributes_job.rb
@@ -16,7 +16,7 @@ class UpdateCachedAppealsAttributesJob < CaseflowJob
     cache_ama_appeals
     cache_legacy_appeals
 
-    record_runtime(start_time)s
+    record_runtime(start_time)
   rescue StandardError => error
     log_error(start_time, error)
   end
--- a/app/jobs/update_cached_appeals_attributes_job.rb
+++ b/app/jobs/update_cached_appeals_attributes_job.rb
@@ -2,7 +2,9 @@
 
 BATCH_SIZE = 1000
 
-class UpdateCachedAppealsAttributesJob < ApplicationJob
+class UpdateCachedAppealsAttributesJob < CaseflowJob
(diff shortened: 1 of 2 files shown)


kevmo:
Taking suggestions on how to formulate a test for this. This is a pretty non-destructive job, though.

app/jobs/update_cached_appeals_attributes_job.rb
```diff
@@ -57,7 +60,7 @@ def cache_legacy_appeal_postgres_data(legacy_appeals)
   end
 
   def cache_legacy_appeal_vacols_data(legacy_appeals)
-    legacy_appeals.pluck(:vacols_id).in_groups_of(BATCH_SIZE).each do |vacols_ids|
+    legacy_appeals.pluck(:vacols_id).in_groups_of(BATCH_SIZE, false).each do |vacols_ids|
```
lowellrex: Great catch! I had no idea about [the second parameter to `in_groups_of`](https://apidock.com/rails/v3.0.0/Array/in_groups_of).

app/jobs/update_cached_appeals_attributes_job.rb
```diff
@@ -90,4 +93,16 @@ def record_runtime(start_time)
...
   end
+
+  def log_error(start_time, err)
+    duration = time_ago_in_words(start_time)
+    msg = "UpdateCachedAppealsAttributesJob failed after running for #{duration}. Fatal error: #{err.message}"
+
+    Rails.logger.info(msg)
+    Rails.logger.info(err.backtrace.join("\n"))
+
+    slack_service.send_notification(msg)
```
lowellrex: I think we need to change `UpdateCachedAppealsAttributesJob` to inherit from `CaseflowJob` so that we have access to `slack_service`.

Issue #11192, opened 2019-06-24
Refactor TaskRows component


aroltsch:
Component `client/app/queue/components/TaskRows.jsx` has become overly complicated when we added withdrawn and cancel statuses to the task timeline. decisionDateTemplate() method has a lot of `if` statements which makes it difficult to understand. I propose that we move timeline logic into its container and think about how we can avoid so many if statements. I also propose that we create a new container for the legacy appeal timeline.

PR #11558, opened 2019-07-26
Show withdrawn icon on casetimeline

Sjones352:
Connects  #11192

Screen Shot 2019-07-26 at 2 26 40 PM


client/app/queue/components/DecisionDateTimeLine.jsx
```diff
@@ -60,11 +59,11 @@ class DecisionDateTimeLine extends React.PureComponent {
       {timeline && <tr>
         <td className="taskContainerStyling taskTimeTimelineContainerStyling">
           <CaseDetailsDescriptionList>
-            {showDecisionDate() || showWithdrawalDate()}
+            {appeal.decisionDate ? showDecisionDate() : showWithdrawalDate()}
```
aroltsch: Do we need this check here? I thought the problem was with the icon styling?

Sjones352: Removed check

PR #11563, opened 2019-07-26
Custom VACOLS attribute type for ASCII column

### LINKED ISSUE #8643: Encoding::UndefinedConversionError: U+2019 from UTF-8 to US-ASCII: INSERT INTO "VACOLS"."ISSUES"
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3307/

### LINKED ISSUE #10748: Best way to ensure data saved to VACOLS is in ASCII?
Here is a Sentry error due to attempting to save a string that couldn't be converted from UTF-8 to ASCII when saving to VACOLS:
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/4994/events/331753/

In this case, the data already existed in the Caseflow DB in the Certifications table. The bad string looks like this:
`Allsup Vet\u0092s Disability Appeals SVC 300 Allsup Pla`.

When the `certify_v2` endpoint was called, it tried to save this string to VACOLS, which resulted in this error:
```
ActiveRecord::StatementInvalid: Encoding::UndefinedConversionError: U+0092 from UTF-8 
to US-ASCII
```
I have fixed this one manually for now, but we need a longer term solution. One that I can think of is to have a middleware that calls `to_ascii` on all user input. Thoughts?

pkarman:
Resolves #8643 
Resolves #10748 

### Description
VACOLS is configured for ASCII7 encoding only, and we keep trying to send it codepoints that are valid UTF-8 but not valid ASCII.

### DIFF
--- a/app/mappers/queue_mapper.rb
+++ b/app/mappers/queue_mapper.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
-require "stringex/unidecoder"
-require "stringex/core_ext"
+require "helpers/ascii_converter"
 
 class QueueMapper
   COLUMN_NAMES = {
@@ -123,11 +122,11 @@ class QueueMapper
   end
 
   def truncate_notes_at_350_characters_and_convert_to_ascii
-    renamed_attributes[COLUMN_NAMES[:note]] = note[0..349].to_ascii if note
+    renamed_attributes[COLUMN_NAMES[:note]] = AsciiConverter.new(string: note).convert[0..349] if note
   end
 
   def truncate_comment_at_600_characters_and_convert_to_ascii
-    renamed_attributes[COLUMN_NAMES[:comment]] = comment[0..599].to_ascii if comment
+    renamed_attributes[COLUMN_NAMES[:comment]] = AsciiConverter.new(string: comment).convert[0..599] if comment
   end
 
   def add_modification_timestamp
--- a/app/mappers/queue_mapper.rb
+++ b/app/mappers/queue_mapper.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "helpers/ascii_converter"
-
 class QueueMapper
   COLUMN_NAMES = {
     case_id: :defolder,
@@ -86,8 +84,7 @@ class QueueMapper
     convert_quality_to_vacols_code
     convert_one_touch_initiative_to_vacols_code
     convert_deficiencies_to_vacols_code
-    truncate_notes_at_350_characters_and_convert_to_ascii
-    truncate_comment_at_600_characters_and_convert_to_ascii
+    assign_note_and_comment
     add_modification_timestamp
   end
 
@@ -121,12 +118,9 @@ class QueueMapper
     end
   end
 
-  def truncate_notes_at_350_characters_and_convert_to_ascii
-    renamed_attributes[COLUMN_NAMES[:note]] = AsciiConverter.new(string: note).convert[0..349] if note
-  end
-
-  def truncate_comment_at_600_characters_and_convert_to_ascii
-    renamed_attributes[COLUMN_NAMES[:comment]] = AsciiConverter.new(string: comment).convert[0..599] if comment
+  def assign_note_and_comment
+    renamed_attributes[COLUMN_NAMES[:note]] = note if note
+    renamed_attributes[COLUMN_NAMES[:comment]] = comment if comment
   end
 
   def add_modification_timestamp
--- a/config/initializers/activerecord_types.rb
+++ b/config/initializers/activerecord_types.rb
@@ -1,12 +1,15 @@
+# frozen_string_literal: true
+
 # Custom column types, particularly for VACOLS
 
 require "helpers/ascii_converter"
 
 class AsciiString < ActiveRecord::Type::Text
   private
+
   def cast_value(value)
     ascii_value = AsciiConverter.new(string: value.to_s).convert
-    limit ? ascii_value[0, limit-1] : ascii_value
+    limit ? ascii_value[0, (limit - 1)] : ascii_value
   end
 end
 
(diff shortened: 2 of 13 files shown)


config/initializers/activerecord_types.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+# Custom column types, particularly for VACOLS
+
+require "helpers/ascii_converter"
+
+class AsciiString < ActiveRecord::Type::Text
+  private
+
+  def cast_value(value)
```
pkarman: the private method is documented for this purpose. https://api.rubyonrails.org/classes/ActiveModel/Type/Value.html#method-i-cast_value

lib/helpers/ascii_converter.rb
```diff
@@ -0,0 +1,36 @@
...
+  private
+
+  attr_reader :str
+
+  def utf8?
+    str.valid_encoding? && str.encoding == Encoding::UTF_8
+  end
+
+  def ascii?
+    str.bytes.none? { |byte| byte > 127 }
```
monfresh: Should we use Ruby's built-in `ascii_only?` method? https://apidock.com/ruby/String/ascii_only%3F

lib/helpers/ascii_converter.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    str.to_ascii
+  end
+
+  private
+
+  attr_reader :str
+
+  def utf8?
+    str.valid_encoding? && str.encoding == Encoding::UTF_8
```
monfresh: Should we use `force_encoding` first before checking if the encoding is valid? I'm afraid the `str.valid_encoding?` without specifying the encoding might return true when it's not actually true. 
https://apidock.com/ruby/String/valid_encoding%3F

lib/helpers/ascii_converter.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def ascii?
+    str.bytes.none? { |byte| byte > 127 }
+  end
+
+  def cp1252?
+    (str.bytes & Array(128..159)).any?
```
monfresh: Should we use Ruby's `force_encoding("Windows-1252").valid_encoding?`

monfresh: On second thought, I don't think `force_encoding` would work if we get a character that can't be converted. I always refer to these articles when I can't remember how encoding works: 
https://www.justinweiss.com/articles/3-steps-to-fix-encoding-problems-in-ruby/
https://www.justinweiss.com/articles/how-to-get-from-theyre-to-theyre/

lib/helpers/ascii_converter.rb
monfresh: According to the first article I linked to below, `encoding` might not always return the actual encoding. For example:
```
"hi\x99!".encoding => #<Encoding:UTF-8>
```
which is wrong.

Also, if you call `valid_encoding?` on a string like "hi\x99!", it will blow up with a `Encoding::UndefinedConversionError`. I'm not sure how often that happens though. It seems like we mostly get UTF-8 and Windows 1252, right?

lib/helpers/ascii_converter.rb
pkarman: I didn't know about `ascii_only?` -- good call

lib/helpers/ascii_converter.rb
pkarman: I don't think we want to `force_encoding` because that implies that we know the string is valid ahead of time.

Here are some examples:

```
[7] pry(main)> s = "hi\x99!"
=> "hi\x99!"
[8] pry(main)> s.valid_encoding?
=> false
[9] pry(main)> ac = AsciiConverter.new(string: s)
=> #<AsciiConverter:0x00007fe8983e1538 @str="hi\x99!">
[10] pry(main)> ac.convert
=> "hi!"  # what we want
```

The string "hi∑" is valid UTF-8 so `valid_encoding?` should just work.

```
irb(main):001:0> s = "hi∑"
=> "hi∑"
irb(main):002:0> s.encoding
=> #<Encoding:UTF-8>
irb(main):003:0> s.valid_encoding?
=> true
```

pkarman: fwiw `"\x99" => "\x{2122}",    #TRADE MARK SIGN`

you could argue that "\x99" should be "(TM)" in ASCII, but that's not what stringex does.

```
[3] pry(main)> "\u{2122}".to_ascii
=> ""
```

lib/helpers/ascii_converter.rb
pkarman: this particular method is looking for suspect bytes in the specific range because those are used by Windows 1252 and *not* by Latin1 (ISO-8859-1).

UTF-8 might also use them which is why we should always check `utf8?` in combination with `cp1252?`. If a string returns `true` for `utf8?` then `cp1252?` is irrelevant.

`encoding` does tell us what the actual encoding of the bytes are, just what Ruby has flagged as the encoding of the `String` object. The `valid_encoding?` is what actually compares `encoding` vs the the bytes and responds accordingly.

I believe that `force_encoding` just changes the flag on the `String` object, which we can reasonably assume will be correct because of `cp1252?` but which I'd rather not trust. The advantage of calling `.encode(dest, src)` is that it will raise an error if something goes wrong, which I think we want.

https://ruby-doc.org/core-2.5.3/String.html#method-i-encode

app/mappers/queue_mapper.rb
```diff
@@ -123,11 +122,11 @@ def convert_deficiencies_to_vacols_code
   end
 
   def truncate_notes_at_350_characters_and_convert_to_ascii
-    renamed_attributes[COLUMN_NAMES[:note]] = note[0..349].to_ascii if note
+    renamed_attributes[COLUMN_NAMES[:note]] = AsciiConverter.new(string: note).convert[0..349] if note
```
monfresh: Instead of truncating here, what do you think about defining the attributes with a limit on the proper VACOLS tables? In this case, it would be `VACOLS::Decass` and the attributes would be `deatcom` with a limit of 350 and `debmcom` with a limit of 600.

monfresh: And we wouldn't need the ascii converter either, right?

pkarman:
@monfresh I think we got this one now.

Issue #11070, opened 2019-06-10
Hearings branch cases don't have actions available


pshahVA:
Problem:
There are cases that are visible with the hearing branch (location 57) in VACOLS but the hearing branch can't perform a task on the case in Caseflow.  The short term fix has been to modify the case by the engineering team via the rails console, but we need to get to the root cause of this.

Technical work around done:

lomaxap:
Write Data Integrity class (#11297) to find all appeals with an incomplete `HearingTask` that do not have an actionable child task

hschallhorn:
One such case:

Appeal 2707721 had no available actions due to a `ScheduleHearingTask` being on hold even though all its child tasks were completed.

```ruby
> LegacyAppeal.find_by(vacols_id: "2707721").structure_render(:id, :status, :closed_at)
LegacyAppeal 497775
└── RootTask 18860, on_hold, (closed_at)
    ├── HearingTask 85054, on_hold, (closed_at)
    │   └── ScheduleHearingTask 85055, on_hold, (closed_at)
    │       ├── HearingAdminActionVerifyAddressTask 96195, completed, 2019-02-26 20:43:27 UTC
    │       │   └── HearingAdminActionVerifyAddressTask 112972, completed, 2019-02-26 20:43:27 UTC
    │       ├── HearingAdminActionVerifyAddressTask 112975, completed, 2019-02-26 20:46:16 UTC
    │       │   └── HearingAdminActionVerifyAddressTask 112978, completed, 2019-02-26 20:46:16 UTC
    │       ├── HearingAdminActionVerifyAddressTask 112976, completed, 2019-02-26 20:54:00 UTC
    │       │   └── HearingAdminActionVerifyAddressTask 112984, completed, 2019-02-26 20:54:00 UTC
    │       ├── HearingAdminActionVerifyAddressTask 112979, completed, 2019-02-26 21:00:56 UTC
    │       │   └── HearingAdminActionVerifyAddressTask 113004, completed, 2019-02-26 21:00:56 UTC
    │       └── HearingAdminActionVerifyAddressTask 112980, completed, 2019-02-26 20:53:33 UTC
    │           └── HearingAdminActionVerifyAddressTask 112989, completed, 2019-02-26 20:53:33 UTC
    └── HearingTask 123835, completed, 2019-02-28 16:51:24 UTC
        └── ScheduleHearingTask 18862, completed, 2019-01-15 14:28:20 UTC
```

[...]

PR #11564, opened 2019-07-26
new data integrity check for open hearing tasks without active descendants

tomas-nava:
Connects #11070

### Description
Adds a `DataIntegrityChecker` process that checks for open `HearingTask`s that have no descendants with actionable statuses and sends a report to the `#appeals-hearings` channel on Slack if it finds them. When a `HearingTask` is in this state, users may be unable to schedule hearings.


app/services/open_hearing_tasks_without_active_descendants_checker.rb
```diff
@@ -0,0 +1,29 @@
...
+class OpenHearingTasksWithoutActiveDescendantsChecker < DataIntegrityChecker
+  def call
+    hearing_task_ids = ids_of_open_hearing_tasks_without_active_descendants
+    build_report(hearing_task_ids)
+  end
+
+  private
+
+  def ids_of_open_hearing_tasks_without_active_descendants
+    HearingTask.open.select { |task| task.descendants.map(&:active?).exclude?(true) }.map(&:id)
```
pkarman: could you use the existing `active` class scope here instead of `map` on each object? Mostly just curious.

tomas-nava: Good idea, I'll try it.

tomas-nava: I tried it out, but `descendants` returns an array rather than an ActiveRecord collection, so a scope won't work.

PR #11565, opened 2019-07-27
Fix non-comp filter

leikkisa:
Bug fix for:
https://dsva.slack.com/archives/CFBLH2LR4/p1563905024021200


leikkisa:
noncomp_filter

PR #11587, opened 2019-07-30
Do not include canceled claims in taken EP modifiers

leikkisa:
Cancelled claims should be released after they are cancelled, like cleared claims. We temporarily did this because many canceled claims were being affected by an external issue where the modifiers weren't being released (which also can affect cleared claims but was less common).

I want to fix this to the intended state (especially since there's no comment explaining the above).  Also, we are now trying to report the Veterans who have this issue to BGS so that they can fix them, but it's difficult to know which Veterans are affected unless they experience the error on a claim that wasn't closed recently.


monfresh:
Should we add a spec for this?

pkarman:
For context, this reverts 9774d381c6d74a337e9462e6a13cff04a85aa0f4 from #11065 addressing #11043

leikkisa:
@monfresh Hi! yes, I am planning on adding a spec. I should have added [WIP] because I wasn't intending for this to be reviewed yet. I have some investigation to do on the data side as well.

PR #11594, opened 2019-07-30
Fix Caseflow timeline for canceled withdrawal of appeals tasks 

### LINKED ISSUE #11498: Fix Caseflow timeline for canceled withdrawal of appeals tasks
A BVA user has canceled a withdrawal of appeal task (because it wasn't a withdrawal of appeal). In the Case Timeline, however, it looks like the mail task was both canceled and completed. 

withdrawal of appeal

https://appeals.cf.ds.va.gov/queue/appeals/dfe4db42-1792-487a-b343-dfb576a45ec0

## AC
- We should hide the task assigned to Mail team from Case Timeline
- We should mark the task assigned to Mail team as canceled, instead of completed

Sjones352:
Resolves #11498
Screen Shot 2019-08-01 at 6 17 08 PM


app/models/task.rb
```diff
@@ -445,10 +445,10 @@ def task_just_closed_and_has_parent?
     task_just_closed? && parent
   end
 
-  def update_status_if_children_tasks_are_complete(child_task)
+  def update_status_if_children_tasks_are_closed(child_task)
     if children.any? && children.open.empty? && on_hold?
       if assigned_to.is_a?(Organization) && cascade_closure_from_child_task?(child_task)
-        return update!(status: Constants.TASK_STATUSES.completed)
+        return update!(status: child_task.status)
```
leikkisa: We should confirm this behavior with @aroltsch since it affects other tasks too.  If we don't want it for other tasks, we can put this into AppealWithdrawalMailTask to override this behavior.

spec/feature/queue/case_details_spec.rb
```diff
@@ -1102,6 +1102,62 @@
...
+      end
+
+      let(:user) { create(:user) }
+
+      before do
+        OrganizationsUser.add_user_to_organization(user, BvaIntake.singleton)
+        User.authenticate!(user: user)
+      end
+
+      it "displays AppealWithdrawalMailTask in case timelinewhen" do
```
leikkisa: typo

spec/feature/queue/case_details_spec.rb
```diff
@@ -1102,6 +1102,62 @@
...
+        click_on "Search"
+        click_on appeal.docket_number
+
+        new_tasks = withdraw_mail_task.children
+        expect(new_tasks.length).to eq(1)
+
+        new_task = new_tasks.first
+        expect(new_task.status).to eq("cancelled")
+
+        expect(page).to have_content("AppealWithdrawalMailTask cancelled")
```
leikkisa: Should we add a check that it's only showing the task assigned to BVA intake, and not the one assigned to the Mail team?  Or that both are showing?
This is another question for @aroltsch in case the original ticket scope has changed.

app/models/task.rb
aroltsch: The logic should be the following:
* if all children tasks are cancelled, set the status to cancelled
* if all children tasks are completed, set the status to completed
* if there is a combination of completed and cancelled statuses, set the status to completed

spec/feature/queue/case_details_spec.rb
aroltsch: They should both be showing.

Sjones352: Fixed

app/models/task.rb
Sjones352: @aroltsch updated pr but my question is are there chances when there is a there is a combination of completed and cancelled statuses? If so would I check for ```if children.pluck(:status).uniq == [Constants.TASK_STATUSES.cancelled] && children.pluck(:status).uniq == [Constants.TASK_STATUSES.completed] &&
          return update!(status: Constants.TASK_STATUSES.completed)```.  I can't seem to recreate a combination of both to find out if that method is correct.

spec/feature/queue/case_details_spec.rb
leikkisa: This test doesn't reflect that they're both showing.

app/models/task.rb
```diff
@@ -445,10 +445,14 @@ def task_just_closed_and_has_parent?
...
-  def update_status_if_children_tasks_are_complete(child_task)
+  def update_status_if_children_tasks_are_closed(child_task)
     if children.any? && children.open.empty? && on_hold?
       if assigned_to.is_a?(Organization) && cascade_closure_from_child_task?(child_task)
-        return update!(status: Constants.TASK_STATUSES.completed)
+        if children.pluck(:status).uniq == [Constants.TASK_STATUSES.cancelled]
+          return update!(status: Constants.TASK_STATUSES.cancelled)
+        else
+          return update!(status: Constants.TASK_STATUSES.completed)
+        end
```
leikkisa: This does the job!  I think something like this might read more easily:
```
if children.all?(&:cancelled?)
  return update!(status: Constants.TASK_STATUSES.cancelled)
else
  return update!(status: Constants.TASK_STATUSES.completed)
end
```

spec/factories/task.rb
```diff
@@ -352,6 +352,20 @@
...
     end
 
+    factory :withdraw_mail_task, class: AppealWithdrawalMailTask do
+      type { AppealWithdrawalMailTask.name }
+      appeal { create(:appeal) }
+      assigned_to { MailTeam.singleton }
+      parent { create(:root_task, appeal: appeal) }
+    end
+
+    factory :withdraw_bva_task, class: AppealWithdrawalMailTask do
```
aroltsch: let's name both of these tasks the same as the class name `withdrawal` instead of `withdraw` and add appeal to it
`appeal_withdrawal_mail_task`

app/models/task.rb
aroltsch: Nice!

aroltsch:
@Sjones352 we would also need to create a migration to update existing production records. We can do it in the next PR.

app/models/task.rb
monfresh: That is indeed easier to read than `children.pluck(:status).uniq == [Constants.TASK_STATUSES.cancelled]`. However, this will load all children in memory and iterate through them. What I had initially proposed I think is the happy medium: both readability and performance by moving the logic to a well-named method:
```ruby
if all_children_cancelled?
  return update!(status: Constants.TASK_STATUSES.cancelled)
else
  return update!(status: Constants.TASK_STATUSES.completed)
end

def all_children_cancelled?
  children.pluck(:status).uniq == [Constants.TASK_STATUSES.cancelled]
end
```
The next step would be to move this stuff to a separate class, and come up with well-named methods for the other complex conditions.

Sjones352: Thanks @monfresh I have implemented this method

PR #11598, opened 2019-07-30
Lighthouse - Higher Level Review /happier/ path

edmangimelli:
the happy path to a higher level review endpoint sans most of the validation

issue: https://github.com/department-of-veterans-affairs/caseflow/issues/11734


app/serializers/api/v3/higher_level_review_preintake.rb
```diff
@@ -0,0 +1,139 @@
+# frozen_string_literal: true
```
pkarman: this is not a serializer. it's a validator and processor. It belongs in `app/services/api`

edmangimelli: 👍

edmangimelli: relatedly, would `higher_level_review_processor` be a better name?

pkarman: `Api::V3::HigherLevelReviewProcessor` sounds good to me.

edmangimelli:
@erluti what do you think about the error lookup? too much

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,173 @@
+# frozen_string_literal: true
+
+class Api::V3::HigherLevelReviewProcessor
+  attr_reader :intake, :errors
+
+  unless (CATEGORIES_BY_BENEFIT_TYPE = JSON.parse(File.read("client/constants/ISSUE_CATEGORIES.json")))
+    fail StandardError, "couldn't pull nonrating issue categories from ISSUE_CATEGORIES.json"
+  end
+
+  def initialize(user:, params:)
```
erluti: I think hash arguments are best when some of them are optional.  Personally I don't like the overhead of assigning names to the args I'm passing in, but I know there's a case for clarity when the arguments are all named like that.

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,173 @@
...
+    end
+  end
+
+  def errors?
+    !errors.empty?
+  end
+
+  def build_start_review_complete
+    transaction do
+      @intake = Intake.build(
```
erluti: build doesn't touch the database, so I think this could be moved into the initializer.  And then you wouldn't need instance variables for `@user` and `@veteran_file_number`.

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,173 @@
...
+
+      acc << request_issue
+    end
+  end
+
+  def errors?
+    !errors.empty?
+  end
+
+  def build_start_review_complete
```
erluti: I think this method should be named with a `!` because it's got some serious side-effects when it's called.

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,173 @@
...
+    @same_office = attributes[:same_office]
+    @legacy_opt_in_approved = attributes[:legacy_opt_in_approved]
+    @benefit_type = attributes[:benefit_type]
+
+    @veteran_file_number = relationships[:veteran][:data][:id]
+
+    @claimant_participant_id = relationships.dig :claimant, :data, :id
+    @claimant_payee_code = relationships.dig :claimant, :data, :meta, :payeeCode
+
+    @request_issues = included.reduce([]) do |acc, included_item|
```
erluti: this seems like it would be a good as a method, then there could be tests just for this logic

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,173 @@
...
+      next acc unless included_item[:type] == "RequestIssue"
+
+      request_issue, error = json_api_included_request_issue_to_intake_data_hash(included_item)
```
erluti: maybe `json_api_included_request_issue_to_intake_data_hash` could add it's errors to `@errors` directly and just return `nil` when it fails.  Then it doesn't need to return the array structure

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,173 @@
+# frozen_string_literal: true
+
+class Api::V3::HigherLevelReviewProcessor
+  attr_reader :intake, :errors
+
+  unless (CATEGORIES_BY_BENEFIT_TYPE = JSON.parse(File.read("client/constants/ISSUE_CATEGORIES.json")))
+    fail StandardError, "couldn't pull nonrating issue categories from ISSUE_CATEGORIES.json"
```
erluti: This should probably be in the test suit to ensure nobody has removed this file from source, but I don't want to have it cause an error at startup.  I guess rails auto-loading magic might hide this issue until this class is used, but I think we need to explicitly handle this error in a way that's localized to our code.  Does this change often?  I think it'd be okay to have this be a class method/variable

pkarman: you should not need to read constants files manually. that's all done for you. Check the `Constants` space. You just use them. See config/initializers/constants.rb for details.

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,173 @@
...
+        [422, "You don't have permission to intake this veteran"]
+      when "veteran_not_valid"
+        [422, "The veteran's profile has missing or invalid information required to create an EP."]
+      when "duplicate_intake_in_progress"
+        [409, "Intake In progress"]
+      when "reserved_veteran_file_number"
+        [422, "Invalid veteran file number"]
+      when "incident_flash"
+        [422, "The veteran has an incident flash"]
+      else; [nil, nil]
```
erluti: I think stylistically Ruby generally avoids semicolons.  But you can drop the else comepletely because a case statement returns `nil` when no condition is met.
Another approach might be to use the else instead of the `unless status || title` below.
```
else
  code = "unknown_error"
  [422, "Unkown error"]
end
```

app/services/api/v3/higher_level_review_processor.rb
edmangimelli: > probably be in the test suit to ensure nobody has removed this file
agreed
> I think it'd be okay to have this be a class method
cool. i'll keep it as a class variable then
> See config/initializers/constants.rb
nice! scratch all that

app/services/api/v3/higher_level_review_processor.rb
edmangimelli: >I think stylistically Ruby generally avoids semicolons. 

that's just an oversight by rubocop. idk why that happened

>else
  code = "unknown_error"
  [422, "Unkown error"]
end

so, i considered this, but decided against it. the case statement is acting like a map, and throwing a side effect in there seems unpalatable to me.

edmangimelli: i've tweaked it to be in more line with how I was thinking of it. let me know what you think

app/services/api/v3/higher_level_review_processor.rb
edmangimelli: it is a method :D  
i should probably just bring back the attr_readers to help testing

app/services/api/v3/higher_level_review_processor.rb
edmangimelli: I was aware, and this was a conscious decision.
Originally I had it outside, but decided to move it in.
It the made the most sense to me to keep them together. I think I did this for someone in the future reading through. All the DBy looking stuff is in a transaction 👍 I know if it was outside, I would be very curious that one DB step was excluded for some reason, and I'd start digging. and then I'd discover that build doesn't actually do stuff with the DB. I guess i felt like I was saving them a goose chase

app/services/api/v3/higher_level_review_processor.rb
edmangimelli: 👍

app/services/api/v3/higher_level_review_processor.rb
erluti: My bigger concern is argument-passing via instance variables.  It makes the code less clear and methods harder to test.  I'm also concerned that even though intake is a readable attribute, it's misleading that it's only set outside of the constructor.
I guess I'm just familiar with Rails enough to know offhand the difference between `build` and `create`, so now I understand why you thought to put this here.

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
```diff
@@ -4,6 +4,34 @@ class Api::V3::DecisionReview::HigherLevelReviewsController < ActionController::
...
+    uuid = higher_level_review.uuid
+
+    response.set_header(
+      "Content-Location",
+      "#{request.base_url}/api/v3/decision_review/higher_level_reviews/intake_status/#{uuid}"
+    )
+
+    render json: intake_status(higher_level_review), status: :accepted
+  rescue StandardError => error
+    error = Api::V3::HigherLevelReviewProcessor.error_from_error_code(
```
erluti: I get why you made this a class method, since Ruby is interpreted anyway it's okay to be an instance method.  Especially since I don't think it's expected to be used outside of this class.  But you can drop the module namespacing if you want to keep it as a class method.
I love the rewrite of the method though!  It looks great!

app/services/api/v3/higher_level_review_processor.rb
erluti: I meant like it could be `@request_issues = issues_from_included(included)` and then this logic is isolated and the constructor is clearer about what it's doing

app/services/api/v3/higher_level_review_processor.rb
edmangimelli: >My bigger concern is argument-passing via instance variables.

this sort ties in to why i'd rather not change `included_request_issue_to_intake_data_hash`. i'd rather keep it as a method with clear ins and outs.

>I'm also concerned that even though intake is a readable attribute, it's misleading that it's only set outside of the constructor.

this isn't an expectation i would have. This is a rails convention? To me a readable attribute is indistinguishable from a method, and I don't have many expectations as to what a standalone class's methods are

>I guess I'm just familiar with Rails enough to know offhand the difference between build and create

you wouldn't unless you looked into the intake model. lol
build gets overridden in the intake model. granted, conventions are followed.

app/services/api/v3/higher_level_review_processor.rb
edmangimelli: ah sorry. misunderstanding

app/services/api/v3/higher_level_review_processor.rb
edmangimelli: this has been switched around. to me, json_api_included... is clearer and more testable in a form closer to how I already have it, but take a look at the latest iteration

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
edmangimelli: Ah. this might be a habit I picked up doing React.
If an instance method isn't using self, why is it an instance method?
So, now, to me, a method being an instance method is a signal that it uses/modifies the internal state of the object. I noticed Rubyists are pretty loose with the role of an instance method though. You're right, very unlikely that it will be needed far from where an instance is being used.
switch it to an instance method?

edmangimelli: >But you can drop the module namespacing if you want to keep it as a class method.

i could be wrong, but I don't think that's true
scopes have to be reopened (?)

example:

```
# doesn't work

module A
  module B
    module C
    end
  end
end

class A::B::C::D
  def florgle
    E.krippik
  end
end

class A::B::E
  def self.krippik
    "The Cowabunga Lifestyle"
  end
end

d = A::B::C::D.new

d.florgle

# works

module A
  module B
    module C
    end
  end
end

module A
  module B
    module C
      class D
        def florgle
          E.krippik
        end
      end
    end
  end
end

module A
  module B
    class E
      def self.krippik
        "The Cowabunga Lifestyle"
      end
    end
  end
end

d = A::B::C::D.new

d.florgle
```

edmangimelli: https://stackoverflow.com/questions/3009477/what-is-rubys-double-colon/55027491#55027491

erluti: I don't think that's the same case as us.  I think since they are both in the `Api::V3` namespace, you can just refer to `HigherLevelReviewProcessor`.  I could be wrong though, I didn't test it in irb, lol

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,146 @@
...
+      identified = benefit_type.present? && category.present?
+      {
+        rating_issue_reference_id: (contests == "on_file_rating_issue") ? id : nil,
+        rating_issue_diagnostic_code: nil,
+        decision_text: decision_text,
+        decision_date: decision_date,
+        nonrating_issue_category: category,
+        benefit_type: benefit_type,
+        notes: notes,
+        is_unidentified: !identified,
```
erluti: if you use `blank?` instead of `present?` you don't need a variable here
```
is_unidentified: benefit_type.blank? || category.blank?
```

edmangimelli: for sure, but this is solely for readability's sake. i had it written that way originally, but decided it made more sense to break it out. i think it's sometimes hard to think about things in the negative

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
edmangimelli: looks like code climate agrees with my "if it doesn't use the instance's state, it shouldn't be an instance method" sentiment:
>Api::V3::HigherLevelReviewProcessor#attributes doesn't depend on instance state (maybe move it to another class?)

spec/requests/api/v3/higher_level_reviews_controller_spec.rb
```diff
@@ -36,14 +36,30 @@
...
-          errors:
-            [
-              error_a,
-              error_b
-            ]
+          errors: [error_a, error_b]
         },
         status: 422
       )
+      expect(subject.call([error_b, error_a])).to eq(
```
erluti: I've never seen this testing patten before where the method is the subject of the tests.  I think it's clearer what's going on if the object is the subject, and then you are calling the method on it.  I also think it's odd to test the meta-programming call instead of calling the method like it's used in code

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
```diff
@@ -4,29 +4,69 @@ class Api::V3::DecisionReview::HigherLevelReviewsController < ActionController::
...
+  #   end
 
-  def intake_status(higher_level_review)
-    {
-      data: {
-        type: "IntakeStatus",
-        id: higher_level_review.uuid,
-        attributes: {
-          status: higher_level_review.asyncable_status
+  class << self
```
erluti: why change this to a class method?

edmangimelli: 2 reasons:

1)
>looks like code climate agrees with my "if it doesn't use the instance's state, it shouldn't be an instance method" sentiment: _"Api::V3::HigherLevelReviewProcessor#attributes doesn't depend on instance state (maybe move it to another class?)"_

2)
testability

spec/requests/api/v3/higher_level_reviews_controller_spec.rb
edmangimelli: i did this because i was testing a class method.
are you saying the object should be the entire class?

erluti: Oh, for class methods the notation usually uses a period like `context .errors_to_render_args`.
The thing I noticed was putting a method as the subject to treat like an object was peculiar to me.  I don't feel like it's necessary to always have a subject block, and in the case of class methods I usually don't and just call them directly. like
```
expect(Api::V3::DecisionReview::HigherLevelReviewsController errors_to_render_args(error_b, error_a)).to eq(...)
```

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,281 @@
...
+  def start_review_complete!
+    ActiveRecord::Base.transaction do
+      start!
+      review!
+      complete!
+    end
+  end
+
+  def higher_level_review
+    intake.detail.reload
```
edmangimelli: is it weird that I had to put reload here @erluti ?

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
```diff
@@ -4,29 +4,71 @@ class Api::V3::DecisionReview::HigherLevelReviewsController < ActionController::
...
+      { json: { errors: errors }, status: errors.map { |e| Integer e.status }.max || 422 }
+    end
+
+    # given multiple objects, will return the error for the first error code it can find
+    def error_from_objects_error_code(*args)
+      args.each do |arg|
+        code = arg.try(:error_code)
+        return Api::V3::HigherLevelReviewProcessor.error_from_error_code(code) if code
+      end
+      Api::V3::HigherLevelReviewProcessor.error_from_error_code(nil)
```
erluti: Maybe we should make the "unknown error" directly accessible here so we don't have to do a lookup for `nil`

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,281 @@
...
+    params[:included].each do |included_item|
+      next unless included_item[:type] == "RequestIssue"
+
+      value = included_request_issue_to_intake_data_hash(included_item)
+      (value.is_a?(self.class::Error) ? @errors : @request_issues) << value
+    end
+  end
+
+  def errors?
+    !errors.empty?
```
erluti: `errors.any?` is the opposite of `errors.empty?`

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,281 @@
...
+  def start_review_complete!
+    ActiveRecord::Base.transaction do
+      start!
+      review!
+      complete!
+    end
+  end
+
+  def higher_level_review
+    intake.detail&.reload
```
erluti: is there a need for this to be reloaded?  if it's for testing you can just call reload in the test

spec/requests/api/v3/higher_level_reviews_controller_spec.rb
```diff
@@ -1,22 +1,265 @@
...
+            error_with_403_integer_status,
+            error_with_404_integer_status,
+            error_with_1000_integer_status,
+            error_with_403_string_status,
+            error_with_404_string_status,
+            error_with_1000_string_status
+          ], 1000
+        ]
+      ].each do |(array, status)|
+        expect(hlrc.errors_to_render_args(array)).to eq(json: { errors: array }, status: status)
```
erluti: could the `it` block be more descriptive if each of these expect-loops were their own test?

spec/requests/api/v3/higher_level_reviews_controller_spec.rb
```diff
@@ -1,22 +1,265 @@
...
+      end
+
+      array = [error_with_string_that_cant_quite_convert_to_int]
+      expect { hlrc.errors_to_render_args(array) }.to raise_error(ArgumentError)
+    end
+  end
+
+  context "#error_from_objects_error_code" do
+    let(:code_a) { :incident_flash }
+    let(:object_a) do
```
erluti: can we use `intake` and `start_error` instead of `object` to name these?

spec/requests/api/v3/higher_level_reviews_controller_spec.rb
```diff
@@ -1,22 +1,265 @@
 # frozen_string_literal: true
 
-describe Api::V3::DecisionReview::HigherLevelReviewsController, type: :request do
+require "rails_helper"
+require "support/intake_helpers"
+require "support/vacols_database_cleaner"
+require File.expand_path("../../../../app/services/api/v3/higher_level_review_processor.rb", __dir__)
+
+hlrp = Api::V3::HigherLevelReviewProcessor
```
erluti: was there a code climate error or something for not referring to this by the full classname?

spec/services/api/v3/higher_level_review_processor_spec.rb
```diff
@@ -0,0 +1,990 @@
...
+    ActionController::Parameters.new(
+      data: data,
+      included: included
+    )
+  end
+
+  context "Error" do
+    subject { Api::V3::HigherLevelReviewProcessor::Error }
+    it("is a Class") { expect(subject).to be_a(Class) }
+    it("creates Structs") { expect(subject.new).to be_a(Struct) }
```
erluti: I feel like this test is just testing Ruby functionality.  And I don't think we actually care if it's a struct, we just care that an error object responds to `is_a?(Error)`, right?  So I think these tests about it being a Class/Struct can be replaced with something that tests it will satisfy the conditional you use to handle `included_request_issue_to_intake_data_hash`.

spec/services/api/v3/higher_level_review_processor_spec.rb
```diff
@@ -0,0 +1,1592 @@
...
+    it("creates a new processor") { expect(subject).to be_a(hlrp) }
+    it("attached user to intake") { expect(subject.intake.user).to be(user) }
+    it("attached veteran to intake") { expect(subject.intake.veteran.file_number).to eq(veteran_file_number) }
+    it("hasn't committed intake to the DB") { expect(subject.intake.id).to be_nil }
+  end
+
+  context "#errors?" do
+    context "(good input)" do
+      subject { hlrp.new(params, user) }
+      it("is false with good input") do
```
erluti: so the context/describe block build a sentence for when the failure happens.
so if this fails, we'll get `Api::V3::HigherLevelReviewProcessor #errors? (good input) is false with good input`
All that to say you don't need to repeat the context in the test name, so you can drop the "with good input"

spec/requests/api/v3/higher_level_reviews_controller_spec.rb
```diff
@@ -1,22 +1,311 @@
...
+            error_with_403_integer_status,
+            error_with_404_integer_status,
+            error_with_1000_integer_status,
+            error_with_403_string_status,
+            error_with_404_string_status,
+            error_with_1000_string_status
+          ], 1000
+        ]
+      ].each do |(array, status)|
+        expect(hlrc.status_from_errors(array)).to eq(status)
```
edmangimelli: >could the it block be more descriptive if each of these expect-loops were their own test?

yes. sounds painful though. lol

spec/requests/api/v3/higher_level_reviews_controller_spec.rb
edmangimelli: could. i _think_ i'd prefer to err on the side of _name it what it is, over what it's used for_ --not sure though

spec/services/api/v3/higher_level_review_processor_spec.rb
edmangimelli: agreed. this was probably written in a sleepier moment

spec/services/api/v3/higher_level_review_processor_spec.rb
edmangimelli: nice

spec/requests/api/v3/higher_level_reviews_controller_spec.rb
edmangimelli: misunderstanding. resolved in DMs. switching back to explicitly stating the class everywhere

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,312 @@
...
+  ERRORS_BY_CODE = [
+    [403, :veteran_not_accessible, "You don't have permission to view this veteran's information"], # i
+    [404, :veteran_not_found, "Veteran not found"], # i
+    [409, :duplicate_intake_in_progress, "Intake in progress"], # i
+    [422, :adding_legacy_issue_without_opting_in, "To add a legacy issue, legacy_opt_in_approved must be true"],
+    [422, :decision_issue_id_cannot_be_blank, "Must specify a decision issue ID to contest a decision issue"],
+    [422, :incident_flash, "The veteran has an incident flash"], # i
+    [422, :intake_review_failed, "The review step of processing the intake failed"],
+    [422, :intake_start_failed, "The start step of processing the intake failed"],
+    [422, :invalid_file_number, "Veteran ID not found"], # i
```
leikkisa: This is different than not found, it means the file number isn't 8 or 9 numbers.

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,312 @@
...
+    [422, :legacy_issue_id_cannot_be_blank, "Must specify a legacy issue ID to contest a legacy issue"],
+    [422, :must_have_text_to_contest_other, "notes or decision_text must be non-blank when contesting \"other\""],
```
leikkisa: I don't think notes is required, just decision_text

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,312 @@
...
+    [422, :notes_cannot_be_blank_when_contesting_decision, "Notes cannot be blank when contesting a decision"],
+    [422, :notes_cannot_be_blank_when_contesting_legacy, "Notes cannot be blank when contesting a legacy issue"],
+    [422, :notes_cannot_be_blank_when_contesting_rating, "Notes cannot be blank when contesting a rating"],
```
leikkisa: Maybe I don't understand what you're using notes for.

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,312 @@
...
+    end
+
+    def blank_notes_error_for_contest_type(type)
+      "notes_cannot_be_blank_when_contesting_#{type}"
+    end
+
+    INTAKE_DATA_HASH_ID_KEY_BY_CONTEST_TYPE = {
+      decision: :contested_decision_issue_id,
+      rating: :rating_issue_reference_id,
+      legacy: :vacols_id
```
leikkisa: For legacy issues, you need a combination of the vacols_id and vacols_sequence_id.  Legacy issue opt-ins shouldn't happen for decision issues, but they could also be rating issues (or nonrating issues). I'm not sure I'm completely understanding all of this yet, but if it were from a rating issue, you'd still also want the rating_issue_reference_id.

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,312 @@
...
+        is_unidentified: false,
+        INTAKE_DATA_HASH_ID_KEY_BY_CONTEST_TYPE[contest_type] => id,
+        notes: notes
+      }
+    end
+
+    def contesting_other_to_intake_data_hash_for_category(request_issue, category)
+      notes, decision_date, decision_text = request_issue.values_at(:notes, :decision_date, :decision_text)
+
+      return error_from_error_code(:must_have_text_to_contest_other) if notes.blank? && decision_text.blank?
```
leikkisa: Notes are optional, but decision_date is not

app/services/api/v3/higher_level_review_processor.rb
edmangimelli: I took this directly from app/intake/pages/search.jsx
```
    const searchErrors = {
      invalid_file_number: {
        title: 'Veteran ID not found',
        body: 'Please enter a valid Veteran ID and try again.'
      },
```

edmangimelli: Should I keep it like the client has it, or should I change my message to reflect that we're expecting 8 or 9 digits?

app/services/api/v3/higher_level_review_processor.rb
edmangimelli: I just thought this was a sane default. To include a request issue, at a minimum, there should be some text, otherwise what is the request issue about? If an ID were included (allowing us to fill in the blanks for them), then this requirement isn't imposed. But, if they can't specify a category, or even a decision date, at a minimum, some text should be included.

app/services/api/v3/higher_level_review_processor.rb
edmangimelli: At some point I got the impression that notes were required. Not sure when --maybe I misunderstood Peter during a video chat?  But, yeah, I can see from the Caseflow UI it's optional. I'll fix this

app/services/api/v3/higher_level_review_processor.rb
edmangimelli: Ah. interesting. It seems like I'm misunderstanding something. It'd be great to chat about this synchronously

app/services/api/v3/higher_level_review_processor.rb
edmangimelli: decision date must always be specified? for all types of request issues?

app/services/api/v3/higher_level_review_processor/request_issue_params.rb
```diff
@@ -0,0 +1,164 @@
...
+#     for converting an API-style request issue params object to an
+#     IntakesController-style request issue params object
+#  
+#   RequestIssueParams::IntakesControllerShape.validate
+#     returns a an Error or nil
+#
+# for more info: 
+#   https://github.com/department-of-veterans-affairs/caseflow/issues/11734
+
+module Api::V3::HigherLevelReviewProcessor::RequestIssueParams
```
pkarman: the use of modules here confuses me.

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,135 @@
+# frozen_string_literal: true
+
+class Api::V3::HigherLevelReviewProcessor
```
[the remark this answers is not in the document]
pkarman: this is really an `IntakeProcessor` -- maybe it doesn't need to refer to higher level review or decision review at all?

app/services/api/v3/higher_level_review_processor/request_issue_params.rb
edmangimelli: Yeah, I was just using them for namespacing. They didn't feel like a class proper to me, nor a mixin

edmangimelli:
@pkarman 

how about this?

*make module `Parameters`, make class `RequestIssue` inside*
```
module Api::V3::HigherLevelReviewProcessor::RequestIssueParams
 ↓ ↓ ↓ ↓ ↓ ↓ ↓
class Api::V3::DecisionReview::Parameters::RequestIssue
```
(takes in API-style RequestIssue Parameters and can convert them
to IntakesController-style RequestIssue Parameters)


*make class*
```
class Api::V3::DecisionReview::Parameters::Intake
```
which utilizes `Parameters::RequestIssue`



*make class*
```
Api::V3::DecisionReview::IntakeSerializer
```

move:
```
Api::V3::DecisionReview::HigherLevelReviewsController.intake_status
 ↓ ↓ ↓ ↓ ↓ ↓ ↓
Api::V3::DecisionReview::IntakeSerializer.status
```



*make class*
```
Api::V3::DecisionReview::Error
```

*move*
```
Api::V3::DecisionReview::HigherLevelReviewsController.status_from_errors
 ↓ ↓ ↓ ↓ ↓ ↓ ↓
Api::V3::DecisionReview::Error.status_from_errors
```

*move*
```
Api::V3::DecisionReview::HigherLevelReviewsController.error_from_objects_error_code
 ↓ ↓ ↓ ↓ ↓ ↓ ↓
Api::V3::DecisionReview::Error.error_from_objects_error_code
```

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,135 @@
...
+  # this method fails by exception. some exceptions will have an error_code method
+  def start_review_complete!
+    ActiveRecord::Base.transaction do
+      start!
+      review!
+      complete!
+    end
+  end
+
+  class << self
```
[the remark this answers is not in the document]
edmangimelli: Agreed. Maybe in some ways I was overcorrecting here. I feel like one of the reasons my first attempt was poorly received is that everyone reviewing felt like I was making too many classes. Separating out all of the code for transforming the parameters was the idea of the `PreIntake` classes that I had before

edmangimelli:
**move everything in **
```
module Api::V3::HigherLevelReviewProcessor::Error
```
**to**
```
Class Api::V3::DecisionReview::Error
```

edmangimelli:
@pkarman

**Re:** moving code in HigherLevelReviewProcessor to IntakeProcessor
(disregarding the parameters extraction/transformation stuff which will be moved out into its own classes. IntakeProcessor will simply use those classes)

**and Re:**
> seems like a lot of these are not specific to Higher Level Reviews. Maybe Api::V3::DecisionReview::Error is a better name?


The big reason I didn't want to do this from the get-go is that it-- *for me to do it* --would have felt like abstracting too early.

At this point, in trying to get to MVP, it felt safest to just think in terms of HigherLevelReviews, then, when we got to iterating on the next lane (SupplementalClaims), we'd properly abstract things up one level.

Not being an intake expert, this seemed like the correct order to do things.

So I guess I was just waiting for this feedback. I'll abstract it. 👍 

therefore:

**move most everything in**
```
Api::V3::HigherLevelReviewProcessor
```
**to a new class**
```
Api::V3::DecisionReview::IntakeProcessor
```
**and make (?)**
```
Api::V3::DecisionReview::HigherLevelReviewProcessor
```
which inherits from `IntakeProcessor`

pkarman:
These proposed changes mostly look reasonable.

Here are some principles I'm operating from:

* class names in the `Api::V3::` space should have parity with the model structure in `app/model`. There are 3 distinct models: the `intake`, the `decision_review` (which comes in 3 flavors), and the `request_issue`. So I would expect names like `Api::V3::IntakeProcessor` and `Api::V3::DecisionReview`, etc. For errors and parameter handling, I think it's safe to use the namespaces like `Api::V3::Error` and `Api::V3::Parameters` though if you want to make those last specific to decision reviews, that's fine too.
* the pattern in this codebase seems to be only using `module` when a file is intended to be `include`'d somewhere else, as with helpers and concerns and other mixins. Not for namespacing.
* I understand the hesitation toward abstracting too early and being driven by the MVP. I think if we adhere to the first principle above, that is still possible. While this API represents a MVP, Caseflow itself has lots of existing patterns and the intake process is well known/named at this point, so following the existing naming hierarchy for models is a safe bet.

edmangimelli:
@pkarman, how far should I go with matching the models?

what about this:
```
namespace = Api::V3::DecisionReview::

  IntakeProcessor                   (most of existing Processor code)
  DecisionReviewIntakeProcessor     (empty)
  ClaimReviewIntakeProcessor        (empty)
  HigherLevelReviewIntakeProcessor  (mostly empty)
  
  
  IntakeError                   (most existing Error code)
  DecisionReviewIntakeError     (empty)
  ClaimReviewIntakeError        (empty)
  HigherLevelReviewIntakeError  (HLR specific errors)
  
  
  RequestIssueParams (request issue transformation code)
  
  IntakeParams (actually, a lot of what's in the Processor
                we'll probably end up in here)
```
I believe the namespace for the entire API is `Api::V3::DecisionReview`, not just `Api::V3`

At least, that's how we're treating our URLs (for instance: `/decision_review/higher_level_review`)

pkarman:
> I believe the namespace for the entire API is Api::V3::DecisionReview, not just Api::V3

I wouldn't necessarily draw a 1:1 between the API you expose publicly and the class structure in the `app/services/api/v3` space. Realistically I would be thinking about `app/models` and `app/services/api/v3` as occupying equivalent levels in the namespace hierarchy.

I wouldn't go as far as creating empty classes just to satisfy the model hierarchy. That *does* feel like overly abstract at this stage. I just wanted to emphasize keeping the distinction between `intake` and `decision_review` which I think you are.

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
```diff
@@ -2,29 +2,44 @@
...
+    end
+
+    processor.run!
+    if processor.errors?
+      render_errors(processor.errors)
+      return
+    end
 
-  private
+    response.set_header("Content-Location", (
```
pkarman: is this just redirecting? I would probably use the `redirect_to` method rather than poking into the `response` object.

That said, I don't typically see APIs that redirect on successful POST. I would expect a JSON response like I wrote in the comment above.

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
```diff
@@ -2,29 +2,44 @@
...
-    )
-    response.set_header(
-      "Content-Location",
-      # id returned is static, if a mock intake_status is created, this should match
-      "#{request.base_url}/api/v3/decision_review/higher_level_reviews/intake_status/999"
-    )
-    render json: intake_status(mock_hlr), status: 202
-  end
+    processor = Api::V3::DecisionReview::HigherLevelReviewProcessor.new(params, current_user)
+    if processor.errors?
```
pkarman: A pattern we often use in controllers is to instantiate the delegate objects in memoized private methods, and then put the rendering steps into their own private methods. So you'd end up with something like:

```ruby

def create
  render_processor_errors
  return if performed?

  processor.run!

  render_processor_errors
  return if performed?

  render_processor_success
end

private

def processor
  @processor ||= Api::V3::DecisionReview::HigherLevelReviewProcessor.new(params, current_user)
end

def render_processor_errors
  return unless processor.errors?

  render_errors(processor.errors)
end

def render_processor_success
  render json: { status: :accepted, location: new_location }, status: :accepted
end

def new_location
  "/api/v3/decision_review/higher_level_reviews/intake_status/#{processor.higher_level_review.uuid}"
end
```

See https://api.rubyonrails.org/classes/ActionController/Metal.html#method-i-performed-3F for the `performed?` method.

app/models/higher_level_review.rb
```diff
@@ -97,6 +97,16 @@ def available_review_options
     %w[supplemental_claim appeal]
   end
 
+  def api_intake_status
```
pkarman: This seems like a useful method, but I'm not sure it belongs in the model. It's more of a [decorator](https://www.thegreatcodeadventure.com/rails-refactoring-part-iii-the-decorator-pattern/) and specific to the API.

How about creating a `Api::V3::DecisionReview::Status` class like:

```ruby
class Api::V3::DecisionReview::Status
  def initialize(decision_review:, type:)
    @decision_review = decision_review
    @type = type
  end

  def call
    {
      data: {
        type: type,
        id: decision_review.uuid,
        attributes: { status: decision_review.asyncable_status }
      }
    }
  end

  private

  attr_reader :decision_review, :type
end
```

and use it like:

```ruby
Api::V3::DecisionReview::Status.new(decision_review: higher_level_review, type: "IntakeStatus").call
```

instead?

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
```diff
@@ -2,29 +2,44 @@
...
-    {
-      data: {
-        type: "IntakeStatus",
-        id: higher_level_review.uuid,
-        attributes: {
-          status: higher_level_review.asyncable_status
-        }
-      }
-    }
+    render json: higher_level_review.api_intake_status, status: :accepted
```
pkarman: see comment below about `api_intake_status` -- it's a view, which doesn't really belong on the model.

app/services/api/v3/higher_level_review_processor.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class Api::V3::DecisionReview::HigherLevelReviewProcessor < Api::V3::DecisionReview::IntakeProcessor
```
pkarman: Could we call this `Api::V3::DecisionReview::HigherLevelReviewIntakeProcessor` instead so it's clear from the name that it inherits from `IntakeProcessor` and operates on an `intake` and not a `decision_review`?

app/services/api/v3/intake_error.rb
```diff
@@ -0,0 +1,88 @@
...
+      when Symbol
+        obj
+      when String
+        obj.to_sym
+      else
+        obj.try(:error_code)
+      end
+    end
+
+    def first_error_code(array)
```
pkarman: this method name is a little ambiguous, and this class is doing some clever manipulation to try and make it easier to consume from the outside.

maybe `find_first_matching_error_code` would be clearer?

app/services/api/v3/intake_params.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+class Api::V3::DecisionReview::IntakeParams
```
pkarman: I really like this!

app/services/api/v3/intake_processor.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+class Api::V3::DecisionReview::IntakeProcessor
+  attr_reader :errors, :intake
+
+  def initialize(params, user, form_type)
```
pkarman: usually if we have more than 2 method arguments we use named parameters just to make it easier on our selves later. So `initialize(params:, user:, form_type:)`.

app/services/api/v3/intake_processor.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+class Api::V3::DecisionReview::IntakeProcessor
+  attr_reader :errors, :intake
```
pkarman: unless they need to be public accessors, we're trying to move toward private by default.

app/services/api/v3/intake_processor.rb
```diff
@@ -0,0 +1,48 @@
...
+    build_intake(user, form_type)
+  end
+
+  def errors?
+    errors.any?
+  end
+
+  # this method performs all of the intake steps which write to DBs.
+  # this method fails by exception. some exceptions will have an error_code method
+  def run!
```
pkarman: I like how clean this is.

app/services/api/v3/request_issue_params.rb
```diff
@@ -0,0 +1,159 @@
...
+    category: :nonrating_issue_category
+  )
+
+  ID_KEYS = [:decisionIssueId, :ratingIssueId, :legacyAppealId, :legacyAppealIssueId].freeze
+
+  IDENTIFYING_KEYS = [*ID_KEYS, :category].freeze
+
+  attr_reader :error_code
+
+  def initialize(request_issue, benefit_type, legacy_opt_in_approved)
```
pkarman: named parameters here? similar to comment above.

app/services/api/v3/request_issue_params.rb
```diff
@@ -0,0 +1,159 @@
...
+#       vacols_id
+#       vacols_sequence_id
+#       contested_decision_issue_id
+#       ineligible_reason
+#       ineligible_due_to_id
+#       edited_description
+#       correction_type
+#     }
+
+class Api::V3::DecisionReview::RequestIssueParams
```
pkarman: I really like how this class landed!

spec/requests/api/v3/higher_level_reviews_controller_spec.rb
```diff
@@ -1,22 +1,331 @@
 # frozen_string_literal: true
 
-describe Api::V3::DecisionReview::HigherLevelReviewsController, type: :request do
+require "rails_helper"
+require "support/intake_helpers"
+require "support/vacols_database_cleaner"
+require File.expand_path("../../../../app/services/api/v3/higher_level_review_processor.rb", __dir__)
```
pkarman: this should not be necessary. is the class you are testing not autoloading? that's a bug if so.

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
erluti: I don't believe this code redirects, it's just setting a header along with the returned json.  We're trying to follow the JSONAPI recommendation for 202 responses found here https://jsonapi.org/recommendations/#asynchronous-processing

pkarman: Understood.

Then I would use the `url_for` method rather than hardcoding the location path, and `set_header` as you've got here. https://apidock.com/rails/ActionController/Base/url_for

app/services/api/v3/intake_processor.rb
edmangimelli: These both need to be accessible publicly.

app/services/api/v3/intake_processor.rb
edmangimelli: I like it! I'll change it

app/services/api/v3/request_issue_params.rb
edmangimelli: 👍

app/services/api/v3/intake_error.rb
edmangimelli: i tweaked this. let me know if it's more clear or less clear

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
```diff
@@ -2,29 +2,43 @@
...
+  def intake_error_from_exception_or_processor(exception)
+    Api::V3::DecisionReview::IntakeError.new(exception, processor.try(:intake))
+  end
+
+  def render_errors(*errors)
+    render Api::V3::DecisionReview::IntakeErrors.new(*errors).render_hash
   end
 end
+
+# def mock_create
```
erluti: i think it's okay to just delete this instead of commenting it out

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
```diff
@@ -2,29 +2,43 @@
...
-    render json: intake_status(mock_hlr), status: 202
+    if processor.run!.errors?
+      render_errors(processor.errors)
+    else
+      response.set_header("Content-Location", url_for(:intake_status, processor.higher_level_review.uuid))
+      render Api::V3::DecisionReview::IntakeStatus.new(processor.intake).render_hash
+    end
+  rescue StandardError => error
+    # do we want something like intakes_controller's log_error here?
+    render_errors(intake_error_from_exception_or_processor(error))
```
erluti: what's the case we're handling here?  looks like `intake_error_from_exception_or_processor` is checking if the error has an error code that corresponds to one of the error codes, then if it doesn't it's checking the processor.  Will the processor collect errors but potentially throw an exception that won't have an error code?

app/services/api/v3/decision_review/intake_error.rb
```diff
@@ -0,0 +1,80 @@
...
+    [500, :intake_start_failed, "The start step of processing the intake failed for an unknown reason"],
+    [500, :reserved_veteran_file_number, "Invalid veteran file number"] # i
+  ].freeze
+
+  UNKNOWN_ERROR = [500, :unknown_error, "Unknown error"].freeze
+
+  KNOWN_ERRORS_BY_CODE = KNOWN_ERRORS.each_with_object({}) { |array, hash| hash[array.second] = array }
+
+  # Given multiple arguments, it uses the first argument that is/has an error code
+  def initialize(*args)
```
erluti: I don't like using the splat to cast things into an array when there's only one case of it needing to be more than one element.  I think this can be clearer if the initializer expected just a code.
Then you can create a builder method to handle accepting objects, and use this builder method to return nil if there is no code so in the one case where there's two args you can `||` them.
```
def initialize(code)
    @status, @code, @title = (
      KNOWN_ERRORS_BY_CODE[code] || UNKNOWN_ERROR
    )
end

def self.new_from_object(obj)
    return nil unless obj.respond_to?(:error_code)
    IntakeError.new(obj.error_code)
end
```

It looks like that `error_code` class method might be superfluous because you seem to never pass a string or symbol to the constructer.  Might be the result of all the refactoring.

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
edmangimelli: >Will the processor collect errors but potentially throw an exception that won't have an error code?
Yes. The intake/decision_review models throw at least one exception that has an error_code, but not all exceptions will. after start_review_complete!, the processor mostly fails by exception. trying to get an error code from the intake is a "just in case the exception has no info, but, whatever caused the exception _did_ set intake's error code." I don't know if this would ever happen or not, but it seems like i should check there anyway

app/services/api/v3/decision_review/intake_error.rb
edmangimelli: error_code isn't superfluous.
currently, when creating an IntakeError, you can give it one or multiple things, and it will use the first thing that is/has an error code.
originally it took just one object (that is/has an error_code), and handling multiple objects was a class method. I can see the value in going back to that approach, although I thought this was cleaner

edmangimelli: this class has been updated. let me know if you like this better

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
erluti: Make sure you include a test that covers this case then.  But if you can't imagine a scenario where this could happen, we should probably trim it back.

app/services/api/v3/decision_review/intake_error.rb
erluti: I definitely liked that you extracted that.  Much more readable.

>error_code isn't superfluous, it's used to a reduce a string/symbol/object to a symbol

When does that happen?  It looks like you are always passing in an object with an `error_code`

edmangimelli: ah. you're write. yeah, this has been refactored enough that this is overkill.

spec/requests/api/v3/decision_review/higher_level_reviews_controller_spec.rb
```diff
@@ -108,31 +109,45 @@
       included: included
     }
   end
+  let(:post_params) do
```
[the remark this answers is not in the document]
edmangimelli: whoops. couldn't see the method body. i thought you were referring to something else

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
```diff
@@ -2,29 +2,41 @@
...
+                     Api::V3::DecisionReview::HigherLevelReviewIntakeProcessor.new(params, current_user)
+                   rescue StandardError
+                     nil
+                   end
+  end
+
+  # Try to create an IntakeError from the exception, otherwise the processor's intake object.
+  # If neither has an error_code, the IntakeError will be IntakeError::UNKNOWN_ERROR
+  def intake_error_code_from_exception_or_processor(exception)
+    Api::V3::DecisionReview::IntakeError.from_first_potential_error_code_found([exception, processor.try(:intake)])
```
pkarman: rubocop may suggest the syntax `processor&.intake` instead.

app/services/api/v3/decision_review/higher_level_review_intake_processor.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class Api::V3::DecisionReview::HigherLevelReviewIntakeProcessor < Api::V3::DecisionReview::IntakeProcessor
+  def initialize(params, user)
+    super(params: params, user: user, form_type: "higher_level_review")
+  end
+
+  def higher_level_review
```
pkarman: `alias higher_level_review detail`

app/services/api/v3/decision_review/intake_params.rb
```diff
@@ -0,0 +1,179 @@
...
+  end
+
+  def veteran_file_number
+    veteran[:data][:id].to_s
+  end
+
+  # params for IntakesController#review
+  def review_params
+    ActionController::Parameters.new(
+      receipt_date: attributes[:receiptDate] || Time.zone.now.strftime("%Y-%m-%d"),
```
pkarman: `strftime("%F")` is the shortcut for this ISO8601 format.

app/services/api/v3/decision_review/intake_status.rb
```diff
@@ -0,0 +1,19 @@
...
+    @intake = intake
+  end
+
+  def render_hash
+    @intake.reload if @intake.id
+    {
+      json: {
+        type: @intake.detail_type,
+        id: @intake.detail&.uuid,
+        attributes: { status: @intake.detail&.asyncable_status }
```
pkarman:
```ruby

private

attr_reader :intake

delegate :detail, to: :intake
```

might DRY that up a bit.

spec/controllers/api/docs/v3/decision_reviews_spec.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+context "api/docs/v3/decision_reviews.yaml" do
+  let(:spec) { YAML.safe_load(File.read("app/controllers/api/docs/v3/decision_reviews.yaml")) }
```
pkarman: You may want/need `File.read(File.join(Rails.root, "app/controllers/api/docs/v3/decision_reviews.yaml")))` instead.

spec/exceptions/api/v3/decision_review/review_error_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+      user: build_stubbed(:user),
+      veteran_file_number: "64205050",
+      form_type: "higher_level_review"
+    )
+    intake.detail = HigherLevelReview.new
+    intake
+  end
+
+  context ".new" do
+    subject { Api::V3::DecisionReview::ReviewError }
```
pkarman: this is the same as the built-in `described_class`

spec/exceptions/api/v3/decision_review/review_error_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    it "creating the exception should not raise an exception" do
+      expect { subject.new(intake) }.not_to raise_error
```
pkarman: could be more idiomatically written as `expect(described_class.new(intake)).to be_a described_class`

spec/exceptions/api/v3/decision_review/start_error_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require "rails_helper"
```
pkarman: I believe these `require "rails_helper"` lines are superfluous. I know a lot of existing specs have them, but that file is always loaded by `.rspec` file.

spec/services/api/v3/decision_review/intake_error_spec.rb
```diff
@@ -0,0 +1,170 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+context Api::V3::DecisionReview::IntakeError do
+  # :reek:UtilityFunction:
```
pkarman: we've been trying to keep all the Reek exceptions in the `.reek.yml` file

spec/services/api/v3/decision_review/intake_params_spec.rb
```diff
@@ -0,0 +1,264 @@
...
+    context "bad request issue" do
+      let(:included) do
+        [
+          {
+            type: "RequestIssue",
+            attributes: {}
+          }
+        ]
+      end
+      it "should have code :" do
```
pkarman: is that ` :` a typo?

app/controllers/api/v3/decision_review/intake_statuses_controller.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class Api::V3::DecisionReview::IntakeStatusesController < Api::V3::BaseController
```
pkarman: not clear to me what this class is for?

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
edmangimelli: i'll switch it

app/services/api/v3/decision_review/intake_status.rb
edmangimelli: 👍

app/controllers/api/v3/decision_review/intake_statuses_controller.rb
edmangimelli: This is a placeholder at the moment. With another ticket, it will be fleshed out to return the status of an intake, eventually returning the URL for the higher level review once it's been processed

Issue #8368, opened 2018-12-19
Edit Hearing Day | Null state


sharonwarner:
Add text to the null state for selectors, "Please Select..."

laurenrussell:
<img width="1062" alt="dropdown-selector" src="https://user-images.githubusercontent.com/44249412/50237834-d88d7f00-038b-11e9-9d7e-b1e2e2e943d1.png">

Sjones352:
Ticket has already been done.

MeredithStewart:
@Sjones352 -- I don't see this in demo or in production working. I don't see the "please select."

Sjones352:
Interesting. Thought @oscarAtNava had already done. Re-opening it

ferristseng:
@lomaxap this should be null room right?

ferristseng:
We have a null constraint on the room column. Would we need to remove that constraint for this?

lomaxap:
Ah, ok. Then yes. We should remove the constraint for this.

PR #11601, opened 2019-07-30
Remove not null constraint on Hearing Day room column

ferristseng:
Needed for #8368

### Description

Remove not null constraint on Hearing Day room column, since we are allowing users to select a null room in the dropdown.

*Schema Changes Only*

* [x] Column comments updated


db/migrate/20190730193421_remove_not_null_constraint_from_hearing_day_room_column.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveNotNullConstraintFromHearingDayRoomColumn < ActiveRecord::Migration[5.1]
+  def change
+    change_column_null :hearing_days, :room, true
+    change_column_comment :hearing_days, :room, "The room where the hearing will take place"
```
sharonwarner: It might be worth clarifying that this is the room number at BVA. People sometimes confuse it for a room number at a regional office.

PR #11608, opened 2019-07-31
WIP: Sorting dockets numbers and types using cached data in /organizations/<org> view.

### LINKED ISSUE #11311: Enable back-end sorting of tasks by docket type and number
#11247 established the pattern for how the pagination API will sort tasks for the queue on the back-end but did not enable sorting by all fields. This ticket exists to enable sorting by docket type and number through the pagination API that mirrors [how we sort by that field on the front-end](https://github.com/department-of-veterans-affairs/caseflow/blob/0377b8d0a0d77a81a070c82c6bd336600885cfa3/client/app/queue/components/TaskTable.jsx#L65).

## Acceptance criteria
* [ ] Enable back-end sorting of tasks by docket type and number

kevmo:
Connects #11311.

This PR contains two changes:
* Sorts tasks by docket number the backend.  Passing a SQL string to `joins` the way we do here appears to be the most common way people do this in Rails, but open to a more Rails-y interpretation. A future iteration of this code may want to move the method doing this work onto the `Task` class.
* Additionally, turns on backend sorting for the days on hold column. The backend sorting logic was added in [PR 11614](https://github.com/department-of-veterans-affairs/caseflow/pull/11608), but the front-end flag was not set -- this PR also corrects that.


app/models/task_pager.rb
```diff
@@ -36,12 +36,14 @@ def sorted_tasks(tasks)
       tasks.order(type: sort_order, action: sort_order, created_at: sort_order)
     when Constants.QUEUE_CONFIG.DAYS_ON_HOLD_COLUMN
       tasks.order(placed_on_hold_at: sort_order)
+    when Constants.QUEUE_CONFIG.DOCKET_NUMBER_COLUMN
+      tasks_with_cached_appeal_attributes(tasks).order("cached_appeal_attributes.docket_number #{sort_order}")
```
lowellrex: I'm not sure if is possible, but can we do something like `.order("cached_appeal_attributes.docket_number ?",  sort_order)` to avoid the direct string interpolation and avenue for SQL injection?

client/app/queue/components/TaskTable.jsx
```diff
@@ -155,6 +156,7 @@ export const typeColumn = (tasks, requireDasRecord) => {
       renderAppealType(task.appeal) :
       <span {...redText}>{COPY.ATTORNEY_QUEUE_TABLE_TASK_NEEDS_ASSIGNMENT_ERROR_MESSAGE}</span>,
     span: (task) => hasDASRecord(task, requireDasRecord) ? 1 : 5,
+    backendCanSort: true,
```
lowellrex: I don't think the back-end can sort by appeal type (original, AOD, CAVC remand) yet, can it?

client/app/queue/components/TaskTable.jsx
```diff
@@ -237,7 +239,7 @@ export const daysOnHoldColumn = (requireDasRecord) => {
           diff(task.placedOnHoldAt, 'days')} />
       </React.Fragment>;
     },
-    backendCanSort: false,
+    backendCanSort: true,
```
lowellrex: Can we spin this change off into a different PR so that we can get it in immediately?

Issue #11442, opened 2019-07-17
TaskTimer Bug for Manually Closed Parent Tasks


lomky:
TaskTimerJobs are only run on TaskTimers that have active parent tasks. If the parent tasks are manually cancelled or otherwise closed, the Timers are left open and never triggered.

We should probably actively close out TaskTimers when their associated parent tasks are closed?

[Slack Context](https://dsva.slack.com/archives/CHX8FMP28/p1563395935454600)  
[Looker of potential cases](https://caseflow-looker.va.gov/explore/caseflow_rds_demo/appeals?qid=EeO2Ay8iaxVgaFziOgHaFy&origin_space=16&toggle=dat,vis)

## Acceptance criteria
* [x] Cancel related task timers when tasks are closed
* [x] Timer Job should check for tasks in the limbo state
* [x] Cancel existing task timers related to closed tasks currently in production
  - Don't blindly close all task timers in the Slack report since some may be real.
* [x] (Maybe) make the job that runs task timers skip (and cancel?) task timers related to closed tasks.

lomky:
As a part of this ticket, check for any open orphaned TaskTimers that need to be closed out in prod.

See [the looker](https://caseflow-looker.va.gov/looks/118?qid=lWSrxlza5N6wrWlxwqxtg2)

lowellrex:
1 - 4
2 - 8
3 - 0
4 - 0
5 - 0

Estimate at 2. Code changes should be straightforward, but manually updating of live data brings this to a 2 level of effort. May be a 1 if the manual live data change is easy.

PR #11611, opened 2019-07-31
Testing for TaskTimerJob success by emitting gauge to DataDog.

lomky:
Completes #11442 

### Description
Adds some DataDog reporting to the TaskTimerJob so we can see runtime, success, or any issues.


lomky:
Filed that issue in #11626

PR #11615, opened 2019-07-31
Show user session

pkarman:
# Description
Adds a new route `/test/users/me` that will dump the current user session for debugging purposes.

In addition, closes a security/performance issue where all the User records were being written to the response on the `/test/users` page in a production environment.

### DIFF
--- a/app/controllers/test/users_controller.rb
+++ b/app/controllers/test/users_controller.rb
@@ -164,14 +164,14 @@ class Test::UsersController < ApplicationController
   helper_method :user_session
 
   def test_users
-    return [] if Rails.env.prod?
+    return [] unless ApplicationController.dependencies_faked?
 
     User.all
   end
   helper_method :test_users
 
   def features_list
-    return [] if Rails.env.prod?
+    return [] unless ApplicationController.dependencies_faked?
 
     FeatureToggle.features
   end
--- a/app/controllers/test/users_controller.rb
+++ b/app/controllers/test/users_controller.rb
@@ -65,8 +65,7 @@ class Test::UsersController < ApplicationController
   ].freeze
 
   # :nocov:
-  def index
-  end
+  def index; end
 
   def show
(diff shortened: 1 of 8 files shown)


spec/feature/log_in_as_user_spec.rb
```diff
@@ -19,19 +19,37 @@
...
-    safe_click("#button-Log-in-as-user")
-    expect(page).to have_content("ANNE MERICA (DSUSER)")
-    expect(page).to have_content("Certification Help")
-    visit "test/users"
-    expect(page).not_to have_content("Log in as user")
-    find("a", text: "ANNE MERICA (DSUSER)").click
-    find("a", text: "Sign Out").click
-    expect(page).to have_content("DSUSER")
-    expect(page).to have_content("admin page")
+  context "user is Global Admin" do
```
pkarman: this first test is just a refactor to use `before/after` blocks.

app/controllers/test/users_controller.rb
```diff
@@ -158,6 +157,30 @@ def show_error
 
   private
 
+  def user_session
+    (params[:id] == "me") ? session : nil
```
pkarman: we can't see other users' sessions at this point but might in future allow for viewing a user's record or other data.

spec/feature/log_in_as_user_spec.rb
```diff
@@ -19,19 +19,37 @@
...
+    before do
+      Functions.grant!("Global Admin", users: ["DSUSER"])
+    end
+
+    after do
+      Functions.client.del("Global Admin")
+    end
+
+    scenario "user is able to log in as user" do
```
leikkisa: Should we stub "dependencies_faked?" here to make sure this still works in prod?

pkarman: good call. I will open another PR

PR #11616, opened 2019-07-31
Refactor Status API controller

pkarman:
No functional change.

Refactors large controller into a smaller service class. Leaves the HTTP specific code in the controller.


spec/services/api/v2/appeals_spec.rb
```diff
@@ -0,0 +1,127 @@
+# frozen_string_literal: true
+
+require "support/vacols_database_cleaner"
+require "rails_helper"
+
```
pkarman: this spec largely lifted from the larger feature test, which still contains all the existing coverage. This file is here mostly as a placeholder so that we can add to it over time to prevent regressions.

pkarman:
I am inclined to skip the codeclimate errors about duplicated code. It is intentionally duplicated test fixture data, to start with some known working tests from the feature test.

spec/services/api/v2/appeals_spec.rb
pkarman: refactored into helpers

pkarman:
@monfresh should be ready for your review

PR #11619, opened 2019-08-01
Allow VSO users to see the Feedback page

pkarman:
https://dsva.slack.com/archives/CHX8FMP28/p1564667715203000


spec/feature/send_feedback_spec.rb
```diff
@@ -15,4 +15,16 @@
...
+
+  context "User is part of VSO" do
+    before do
+      allow(current_user).to receive(:vso_employee?) { true }
+    end
+
+    scenario "Can see Feedback page" do
+      visit "/feedback"
+
+      expect(page).to have_content("YourIT")
```
monfresh: Could we use a controller spec instead since the main thing we're testing is the authorization, as opposed to the contents of the feedback page?

PR #11620, opened 2019-08-01
move feature spec to controller spec

pkarman:
#11619 re-write

### DIFF
--- a/spec/controllers/application_controller_spec.rb
+++ b/spec/controllers/application_controller_spec.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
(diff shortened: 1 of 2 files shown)


spec/controllers/application_controller_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe ApplicationController, type: :controller do
+  let(:user) { build(:user) }
+
+  before do
+    allow_any_instance_of(ApplicationController).to receive(:current_user).and_return(user)
```
monfresh: Should we also simulate that the user is a VSO user?

pkarman:
@monfresh VSO test added

Issue #9674, opened 2019-02-27
UI Fix: Filtering for Suggested Locations


ShellySmithDSVA:
Filtering for the Suggested Locations under the RO has a floating ghosted box and is off to the left of the column. This should be centered under the column so that visually the functions are connected. See screenshot below. 

![screen shot 2019-02-27 at 12 07 07 pm](https://user-images.githubusercontent.com/45463980/53509827-a33b9680-3a8a-11e9-95d8-5b8cc4cf84ae.png)

lomaxap:
will be resolved with #11285

PR #11629, opened 2019-08-01
Refactor Assign Hearings to use QueueTable instead of Table

### LINKED ISSUE #11285: Use Updated Table for Hearing Schedule
We currently use an older version of the `Table` Component for both `/hearings/schedule` and `hearings/schedule/assign`, we should update this to use [`QueueTable`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/QueueTable.jsx) 

this will also resolve #9674

ferristseng:
Resolves #11285

Also confirmed that the issue mentioned in #9674 was resolved too.

Screen Shot 2019-08-02 at 11 04 44 AM

### Description

  * Update `AssignHearingTabs` to use `QueueTable` instead of `Table` component
    * Remove code related to sorting table from `AssignHearingTabs`, since `QueueTable` has that built-in.
  * Add new options to `QueueTable`
    * `enableFilterTextTransform`, to disable auto-capitalization of filter options so that locations display better
    * `filterValueTransform ` (named `valueTransform ` in `TableFilter`), to transform non-string values to strings so they can be filtered.
  * Refactor `QueueTable` so `rowObjects` automatically gets passed into `tableData` for the `TableFilter` if `tableData` is not specified.


client/app/components/TableFilter.jsx
```diff
@@ -126,6 +141,9 @@ class TableFilter extends React.PureComponent {
       this.filterDropdownOptions(tableData, columnName) :
       // Keeping the historical prop `getFilterValues` for backwards compatibility,
       // will remove this once all apps are using this new component.
+      // 
+      // WARNING: If you use getFilterValues, it will cause some of the options to
+      // not display correctly when they are checked.
```
ferristseng: Going to create a follow-up PR to remove `getFilterValues`. I don't think it's being used anymore from a search.

client/app/queue/QueueTable.jsx
```diff
@@ -87,6 +93,8 @@ const HeaderRow = (props) => {
         if (!props.useTaskPagesApi && (column.enableFilter || column.getFilterValues)) {
           filterIcon = <TableFilter
             {...column}
+            tableData={column.tableData || props.rowObjects}
```
lomaxap: 👍

client/app/queue/QueueTable.jsx
```diff
@@ -243,9 +252,22 @@ export default class QueueTable extends React.PureComponent {
...
         // Only return the data point if it contains the value of the filter
         filteredData = filteredData.filter((row) => {
-          const cellValue = _.get(row, columnName);
+          let cellValue = _.get(row, columnName);
+
+          if (columnConfig && columnConfig.filterValueTransform) {
+            cellValue = columnConfig.filterValueTransform(cellValue);
+          }
 
           if (typeof cellValue === 'undefined' || cellValue === null) {
```
lomaxap: could change this to `_.isNil(cellValue)`

PR #11630, opened 2019-08-01
Available actions migrated to subclasses

### LINKED ISSUE #11526: Remove code with behaviour based on ColocatedTask action field
Since we are [now creating subclasses of `ColocatedTask`](https://github.com/department-of-veterans-affairs/caseflow/pull/11295) instead of instances of that parent class we can have logic rely on the class (preferably overridden in a subclass) instead of the `action` field of the instance we are dealing with. This ticket exists to replace existing `ColocatedTask` code ([like `available_actions_with_conditions `](https://github.com/department-of-veterans-affairs/caseflow/blob/005260203ae777fa8ea3671de553eba47478e1b0/app/models/tasks/colocated_task.rb#L84)) that relies on the `action` with code that offloads that logic to the subclasses.

## Acceptance criteria
* [ ] Remove code with behaviour based on `ColocatedTask` action field

lomky:
Resolves #11526 

### Description
Removing action code!

 - Removed `action` from the available_actions logic
 - Removed validation of `action`
 - Removed label based on `action`

### Acceptance Criteria
- [x] Tests keep passing
- [x] No UI changes
  - [x] Actions available on Translation Colocated Tasks same
  - [x] Actions available on Schedule Hearing Colocated Tasks same

### DIFF
--- a/app/models/tasks/colocated_task.rb
+++ b/app/models/tasks/colocated_task.rb
@@ -122,9 +122,9 @@ class ColocatedTask < Task
   def vacols_location
     # Break this out into respective subclasses once ColocatedTasks are migrated
     # https://github.com/department-of-veterans-affairs/caseflow/pull/11295#issuecomment-509659069
-    if action == "schedule_hearing" || type == ScheduleHearingColocatedTask.name
+    if type == ScheduleHearingColocatedTask.name
       schedule_hearing_vacols_location
-    elsif action == "translation" || type == TranslationColocatedTask.name
+    elsif type == TranslationColocatedTask.name
       translation_vacols_location
     else
       assigned_by.vacols_uniq_id
@@ -154,7 +154,6 @@ class ColocatedTask < Task
       appeal_id: appeal_id,
       assigned_to_id: assigned_to_id,
       assigned_to_type: assigned_to_type,
-      action: action,
       type: type,
       parent_id: parent_id,
       instructions: instructions
--- a/app/models/tasks/colocated_task.rb
+++ b/app/models/tasks/colocated_task.rb
@@ -67,6 +67,8 @@ class ColocatedTask < Task
     if assigned_to == user ||
        (task_is_assigned_to_user_within_organization?(user) && Colocated.singleton.user_is_admin?(user))
       actions = [
+        Constants.TASK_ACTIONS.COLOCATED_RETURN_TO_ATTORNEY.to_h,
+        Constants.TASK_ACTIONS.CHANGE_TASK_TYPE.to_h,
         Constants.TASK_ACTIONS.TOGGLE_TIMED_HOLD.to_h,
         Constants.TASK_ACTIONS.ASSIGN_TO_PRIVACY_TEAM.to_h,
         Constants.TASK_ACTIONS.CANCEL_TASK.to_h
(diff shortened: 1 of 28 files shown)


lomky:
I believe the CodeClimate FeatureEnvy is a false positive

app/models/tasks/colocated_task.rb
```diff
@@ -160,9 +122,9 @@ def appeal_in_caseflow_vacols_location?
   def vacols_location
     # Break this out into respective subclasses once ColocatedTasks are migrated
     # https://github.com/department-of-veterans-affairs/caseflow/pull/11295#issuecomment-509659069
-    if action == "schedule_hearing" || type == ScheduleHearingColocatedTask.name
+    if type == ScheduleHearingColocatedTask.name
       schedule_hearing_vacols_location
```
hschallhorn: We should move `schedule_hearing_vacols_location` to `vacols_location` in `ScheduleHearingColocatedTask`.

app/models/tasks/colocated_task.rb
```diff
@@ -160,9 +122,9 @@ def appeal_in_caseflow_vacols_location?
...
-    elsif action == "translation" || type == TranslationColocatedTask.name
+    elsif type == TranslationColocatedTask.name
       translation_vacols_location
```
hschallhorn: Same comment as above!

app/models/tasks/colocated_task.rb
```diff
@@ -66,35 +66,21 @@ def timeline_title
...
-    []
-  end
-
-  def available_actions_with_conditions(core_actions)
-    # Break this out into respective subclasses once ColocatedTasks are migrated
-    # https://github.com/department-of-veterans-affairs/caseflow/pull/11295#issuecomment-509659069
-    if %w[translation schedule_hearing].include?(action) && appeal.is_a?(LegacyAppeal)
-      return legacy_translation_or_hearing_actions(core_actions)
-    end
+      actions.unshift(Constants.TASK_ACTIONS.COLOCATED_RETURN_TO_ATTORNEY.to_h)
```
hschallhorn: Can probably move these up into the original actions array.

hschallhorn:
Can also remove[ this line](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/tasks/colocated_task.rb#L195) as well!

app/models/tasks/colocated_task.rb
lomky: I wasn't sure how to do so and preserve the ordering?

hschallhorn: I wonder how important the ordering of the actions are outside of having to update tests

lomky: also, the only person noticing an order change will be admins, with their Reassign appearing on top, so I think they can deal with it okay

app/models/tasks/colocated_task.rb
```diff
@@ -67,17 +67,16 @@ def available_actions(user)
...
         Constants.TASK_ACTIONS.CANCEL_TASK.to_h
       ]
 
       actions.unshift(Constants.TASK_ACTIONS.REASSIGN_TO_PERSON.to_h) if Colocated.singleton.user_is_admin?(user)
 
-      actions.unshift(Constants.TASK_ACTIONS.COLOCATED_RETURN_TO_ATTORNEY.to_h)
-      actions.unshift(Constants.TASK_ACTIONS.CHANGE_TASK_TYPE.to_h)
-
-      return actions
+      actions
```
hschallhorn: Found you! This should be `return actions`

PR #11636, opened 2019-08-02
Abstract v3 api controller

### LINKED ISSUE #11635: Lighthouse API - abstract controller for external api access
We need a controller that consolidates controller features in common with the v3 api work that is exposed to external users through the Lighthouse API gateway.  This should also remove extra brakeman ignore exceptions for new v3 controllers.

erluti:
Resolves #11635

### Description
Adding a controller to be shared between all the new v3 work related to supporting the external proxy access.

### Acceptance Criteria
- [ ] Code compiles correctly


app/controllers/api/external_proxy_controller.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class Api::ExternalProxyController < ActionController::Base
```
edmangimelli: Just out of curiosity, why the name "Api::ExternalProxyController" and something like "Api::Base"?

erluti: `ApplicationController` is more of the Rails convention and one already exists in the API namespace.  Since we're not using the name "Lighthouse" in our naming conventions, I figured since this controller was for APIs accessible through the external proxy (aka Kong), I'd call it that.

pkarman: `ExternalProxyController` is pretty meaningless inside Caseflow.

Why are we not simply inheriting from `Api::ApplicationController`?

edmangimelli: `Api::V3::BaseController` ?

PR #11644, opened 2019-08-02
Sort tasks by docket number in Organization Queue view.

kevmo:
Relates to #11311.  Allows user to sort tasks in their queue by docket number - sorting work is done on the backend.

Adds a Factorybot `cached_appeal` factory.

Also adds PropTypes checking in `TaskTable.jsx`


app/models/task_pager.rb
```diff
@@ -36,13 +36,15 @@ def sorted_tasks(tasks)
       tasks.order(type: sort_order, action: sort_order, created_at: sort_order)
     when Constants.QUEUE_CONFIG.TASK_HOLD_LENGTH_COLUMN
       tasks.order(placed_on_hold_at: sort_order)
+    when Constants.QUEUE_CONFIG.DOCKET_NUMBER_COLUMN
+      tasks_with_cached_appeal_attributes(tasks).order("cached_appeal_attributes.docket_number #{sort_order}")
```
kevmo: This is, to the best of my knowledge, the way we have to do this. However, our fears about SQL injection can perhaps be laid to rest by the fact that this code will error anytime the string is anything but `ASC` or `DESC` (see `validate :sort_order_is_valid`)

kevmo:
Interested in ideas about the best way to test this.

app/models/task_pager.rb
lowellrex: We can eliminate all chances of SQL injection by doing something like the following directly above this:

```ruby
# Do not directly insert string into SQL statement and verify that we are sorting as expected.
verified_sort_order = sort_order == "DESC" ? "DESC" : "ASC"
tasks_with_cached_appeal_attributes(tasks).order("cached_appeal_attributes.docket_number #{verified_sort_order}")
```

I favor that approach (if we truly cannot sanitize the input any other way) so that the hand-sanitization happens as close as possible to the SQL statement we pass off the database so that years from now we don't change some validation and allow for any string input in that field.

lowellrex:
This PR replaces the now-outdated PR #11608.

app/models/task_pager.rb
lowellrex: It looks like [the front-end sorts this column](https://github.com/department-of-veterans-affairs/caseflow/blob/32bdb1dca6311cd429c57a0a41dd2ad49e43e1bc/client/app/queue/components/TaskTable.jsx#L65) on the combination of docket name and docket number. Can we do the same thing when sorting on the back-end here?

lowellrex:
With regards to testing, let's add [some model tests](https://github.com/department-of-veterans-affairs/caseflow/blob/686b4e96e7608596a2d64cca5c5d6a15aba80b04/spec/models/task_pager_spec.rb#L116). I think those should be sufficient. Though if we want to be very thorough we could add feature tests that confirm the rows are sorted the same when we are using the pagination API as they are when we are doing all the sorting on the front-end.

kevmo:
I am looking for advice as to how to most efficiently mock the fact that these operations rely on the joining of a table that contains cached data.

PR #11646, opened 2019-08-02
Run UpdateCachedAppealsAttributesJob during data seeding.

kevmo:
Run `UpdateCachedAppealsAttributesJob.perform_now` in `seeds.rb` so that `cached_appeal_attributes` table is set up for local development along with other seed data.


db/seeds.rb
```diff
@@ -1256,6 +1256,8 @@ def seed
     create_dispatch_tasks(50)
     create_ramp_elections(9)
     create_api_key
+
+    UpdateCachedAppealsAttributesJob.perform_now
```
lowellrex: Actually, can we move this to sit next to `FetchHearingLocationsForVeteransJob.perform_now` since they both run for the same reason.

kevmo:
Updated. gonna slap the ready-to-merge label on once tests clear

Issue #11307, opened 2019-07-08
Allow organization queue table configuration to accept filtering


lowellrex:
Spinning this out of #11052.

#11049 uses tasks from each tab's `tasks` element in the configuration returned from the back-end instead of the tasks from the global redux store. This ticket exists to enable the back-end to return those tasks filtered by their properties (the ability for the front-end to send these requests will be completed as part of a later PR). A quick sketch of one way this could look is below.

If we receive a request to `/organizations/vlj-support-staff/tasks?tab=on_hold&filter[]=col%3Ddocket_type%26val%3Dlegacy&filter[]=col%3Dtask_action%26val%3Dtranslation` we would return tasks that correspond to the following ActiveRecord statement.

```ruby
Task.where(assigned_to: Colocated.singleton)
  .where(status: "on_hold") # for the tab
  .where(appeal_type: "LegacyAppeal") # First filter
  .where(action: "translation") # Second filter
```

## Acceptance criteria
* [ ] Allow organization queue tasks endpoint to returned filtered set of tasks.

PR #11648, opened 2019-08-02
Allow backend to accept Queue filters

lowellrex:
Connects #11307. Enables `TaskPager` to accept `filters` attribute for additional filtering that will be passed from the front-end.


app/models/task_pager.rb
```diff
@@ -18,14 +17,16 @@ def initialize(args)
...
     @sort_by ||= Constants.QUEUE_CONFIG.CASE_DETAILS_LINK_COLUMN
     @sort_order ||= Constants.QUEUE_CONFIG.COLUMN_SORT_ORDER_ASC
+    @filters ||= []
 
     fail(Caseflow::Error::MissingRequiredProperty, message: errors.full_messages.join(", ")) unless valid?
   end
 
   def paged_tasks
-    sorted_tasks(tasks_for_tab).page(page).per(TASKS_PER_PAGE)
+    sorted_tasks(filtered_tasks).page(page).per(TASKS_PER_PAGE)
```
lowellrex: This is the heart of the change. All subsequent code changes spring from here.

app/models/task_pager.rb
```diff
@@ -18,14 +17,16 @@ def initialize(args)
...
   end
 
+  # rubocop:disable Metrics/AbcSize
```
lowellrex: I'm at a loss for how to reduce the number of branching conditions here while maintaining the simplicity that currently exists. Option to suggestions though!

app/models/task_pager.rb
```diff
@@ -55,36 +57,14 @@ def sorted_tasks(tasks)
       tasks.order(created_at: sort_order)
     end
   end
+  # rubocop:enable Metrics/AbcSize
 
-  def tasks_sorted_by_docket_number(tasks)
-    # Additional guard against SQL Injection
```
lowellrex: Now that this code has moved up alongside everywhere else we use `sort_order` I think we're fine to insert that value as a string directly into a sql statement.

app/models/task_pager.rb
```diff
@@ -55,36 +57,14 @@ def sorted_tasks(tasks)
...
-    tasks_with_cached_appeal_attributes(tasks).order(order_sql)
-  end
-
-  def tasks_with_cached_appeal_attributes(tasks)
-    sql = "left join cached_appeal_attributes \
-          on cached_appeal_attributes.appeal_id = tasks.appeal_id \
-          and cached_appeal_attributes.appeal_type = tasks.appeal_type"
-
-    tasks.joins(sql)
+  def cached_attributes_join_clause
```
lowellrex: Factored this join clause because we need to use it multiple places (before filtering and before ordering) because `ActiveRecord` "forgets" about the transient joins after the statement it prepares is executed.

app/models/task_pager.rb
```diff
@@ -93,6 +73,11 @@ def total_task_count
     @total_task_count ||= tasks_for_tab.count
   end
 
+  def filtered_tasks
+    where_clause = QueueWhereClauseArgumentsFactory.new(filter_params: filters).arguments
```
lowellrex: I'm still a little unsure about whether spinning `QueueWhereClauseArgumentsFactory` and `QueueFilterParameter` into their own classes is worthwhile, so I'm curious to get your thoughts! Provides some handy validation code built right in to Rails, but also separates some code.

app/models/task_pager.rb
lomky: I'll think on it a bit, but I'm also not convinced there's a simpler & clearer way to write this type of function.

app/models/task_pager.rb
```diff
@@ -37,7 +38,8 @@ def sorted_tasks(tasks)
     when Constants.QUEUE_CONFIG.TASK_HOLD_LENGTH_COLUMN
       tasks.order(placed_on_hold_at: sort_order)
     when Constants.QUEUE_CONFIG.DOCKET_NUMBER_COLUMN
-      tasks_sorted_by_docket_number(tasks)
```
lomky: I have a slight preference for the semantic clarity of a named function `tasks_sorted_by_docket_number` here vs the inline `join(a).order(b,c)`, but only slight. 

As a bonus, that would reduce the complexity of the method and might satisfy rubocop/ABC. It would only count as 1 point (`tasks_sorted_by_docket_number()`) vs 5 points (`joins`, `cached_attributes_join_clause`, `order`, `docket_type`, `docket_number`). (Or maybe it's only 3 this way? I'm not sure how rubocop counts the quoted ones)

app/models/task_pager.rb
lomky: Since we validate the sort order as in comes in, did this actually catch anything before?

app/models/task_pager.rb
lomky: The validations combined with the clarity of the specs seem worth it to me, the calls are also quite clear, and this is a pretty expected practice in our code base. I'm on board for their distinct files.

spec/models/queue_where_clause_arguments_factory_spec.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe QueueWhereClauseArgumentsFactory, :postgres do
+  describe ".new" do
+    let(:filter_params) { nil }
```
lomky: This is overridden in every context block. Delete it?

spec/models/queue_where_clause_arguments_factory_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      it "instantiates without error" do
+        expect { subject }.to_not raise_error
+        expect(subject).to be_a(QueueWhereClauseArgumentsFactory)
+      end
+    end
+  end
+
+  describe ".arguments" do
+    let(:filter_params) { nil }
```
lomky: Ditto always-overwritten comment

spec/models/task_pager_spec.rb
```diff
@@ -235,4 +235,50 @@
       end
     end
   end
+
+  describe ".filtered_tasks" do
+    let(:assignee) { create(:organization) }
+    let(:tab_name) { Constants.QUEUE_CONFIG.UNASSIGNED_TASKS_TAB_NAME }
+    let(:filters) { [] }
```
lomky: Ditto always-overwritten comment

spec/models/task_pager_spec.rb
```diff
@@ -235,4 +235,50 @@
...
+
+        it "returns the same set of tasks for the filtered and unfiltered set" do
+          expect(subject.map(&:id)).to match_array(task_pager.tasks_for_tab.map(&:id))
+        end
+      end
+
+      context "when filter includes TranslationTasks" do
+        let(:filters) { ["col=#{Constants.QUEUE_CONFIG.TASK_TYPE_COLUMN}&val=#{TranslationTask.name}"] }
+
+        it "returns only translation tasks assigned to the current organization" do
```
lomky: :100: awesome defensive, comprehensive testing!

app/models/task_pager.rb
lomky: A small suggestion for the `DOCKET_NUMBER_COLUMN` case that might slip us under the complexity line, but I think the overall structure of the function is good as is.

app/models/task_pager.rb
lowellrex: I don't think it did. I suggested we add it to defend against our future selves who may want to allow any input for that field and not realize that we were using that value directly in a SQL statement. That was perhaps always too defensive, but now that we've moved that code up alongside where we use that value in a bunch of other places, I am solidly in the "unnecessarily defensive" camp.

app/models/task_pager.rb
hschallhorn: Re: reducing the complexity. As soon as we start hitting the rest of the sort, rails will start griping and we'll have to add it in anyway.

Issue #11176, opened 2019-06-21
[Epic] Prevent duplicate TranslationTasks for an appeal


lowellrex:
The [Translation team noticed](https://dsva.slack.com/archives/CHX8FMP28/p1560976113274800) that there are multiple tasks in their queue for the same appeal for two reasons.

1. The application automatically created a `TranslationTask` and Board employees later created another `TranslationTask` manually.
2. `ColocatedTask`s with open child `TranslationTask`s were marked as complete and additional `ColocatedTask`s with child `TranslationTask`s were created.

We were able to make these determinations by [running the following SQL query](https://caseflow-looker.va.gov/sql/xmsvb6j5h8xktp):
```sql
select appeal_id
, appeal_type
, count(*)
from tasks
where type = 'TranslationTask'
  and assigned_to_type = 'Organization'
group by 1, 2
having count(*) > 1;
```

Which revealed the following six appeals had multiple open `TranslationTask`s: `806`, `1271`, `1439`, `2222`, `2280`, and `2649`. Digging into the task trees for these appeals we saw two patterns.

First, `ColocatedTask`s are [able to be completed](https://github.com/department-of-veterans-affairs/caseflow/blob/8c862c8641f361801ff7a31988ffc0a5acd08d14/app/models/tasks/colocated_task.rb#L83) even if they have open child tasks. Two of the six appeals with duplicate `TranslationTask`s had the duplicate tasks created when there were already open `TranslationTask`s. The questions I'd like to answer is if we are not making this obvious enough for Caseflow users (that there is an open `TranslationTask`), if users know there is an open `TranslationTask` and just don't know what to do so they create another one, or if we should just prevent the `ColocatedTask` from being closed in the first place? I'm in favor of the last option here, but am curious what others think.

In order to clean up the duplicated `TranslationTask`s from this situation, I propose we cancel the `TranslationTask`s that are children of closed `ColocatedTask`s, leaving open the `TranslationTask`s with open `ColocatedTask`s.

[...]

PR #11654, opened 2019-08-05
Translation Task blocks appeal distribution

lomky:
Bumps #11176 

### Description
Translation Tasks are now child tasks on the Distribution task, preventing the appeal from being ready for distribution until they are complete


lomky:
@kevmo that is what the test confirming the parent task status of `on_hold` is doing! We already test that `on_hold` distribution tasks aren't grabbed.

PR #11659, opened 2019-08-05
FactoryBot ColocatedTask assignee fix

hschallhorn:
### Description
The factory for colocated tasks sets the correct assignee for the task. Removed unneeded assignee callouts to hopefully clear up confusion and fix seed data. 
Also added a test to ensure changing the task type of a Colocated task assigns the new task to the correct assignee.


app/models/tasks/colocated_task.rb
```diff
@@ -103,7 +103,7 @@ def create_twin_of_type(params)
       parent: parent,
       assigned_by: assigned_by,
       instructions: params[:instructions],
-      assigned_to: Colocated.singleton
+      assigned_to: task_type.default_assignee
```
hschallhorn: Fixes changing task types to a task that has a different default assignee than the original

db/seeds.rb
```diff
@@ -867,8 +867,7 @@ def create_task_at_colocated(appeal, judge, attorney, trait = ColocatedTask.acti
 
     org_task_args = { appeal: appeal,
                       parent: atty_task,
-                      assigned_by: attorney,
-                      assigned_to: Colocated.singleton }
```
hschallhorn: Was creating translation tasks in a weird state as detailed [here](https://dsva.slack.com/archives/CJL810329/p1565032723122900?thread_ts=1565032600.121500&cid=CJL810329).

PR #11661, opened 2019-08-05
Add user-friendly validation to `outcode` endpoint

### LINKED ISSUE #11354: IDT Outcode: Return more helpful error message when citation number already exists
Currently, when IDT tries to outcode an appeal with an existing citation number (tied to a different appeal), the controller returns a 200 because the `ActiveRecord::RecordNotUnique` error message is not rescued. 

Instead of raising exceptions for expected scenarios and rescuing them (an anti-pattern), we should be adding validations using `ActiveModel::Model` in `LegacyAppealDispatch` and `AmaAppealDispatch` and using our pattern of having the service class return a `FormResponse` that the controller can parse for errors.

monfresh:
Add user-friendly validation to `outcode` endpoint

**Why**: Previously, we were handling some validation errors by raising
exceptions and then rescuing them in various controllers in order to 
respond with the appropriate JSON. This is not optimal for various 
reasons:
- Raising errors is slow
- Exceptions should not be raised for flow of control
  (https://rubystyle.guide/#no-exceptional-flows)
- Exceptions should not be raised for expected scenarios, such as
invalid data from the user
- Leaving it up to Rails to raise exceptions makes it hard to see what 
exactly is being validated, and results in inconsistent behavior and 
bugs. For example, in some cases, we were returning JSON such as 
`{ message: "error message" }`. In other cases, we were returning a 
`Caseflow::Error`, which uses a different hash.

One common Sentry error we have been getting is
`ActiveRecord::RecordNotUnique` due to IDT trying to outcode an appeal
with an existing citation number (tied to a different appeal). This
error was not being rescued, and the controller was returning a 200
response nonetheless, which is misleading.

**How**:
- Use our pattern where the controller calls a class that performs
ActiveModel validations, and returns a `FormResponse` object that
responds to `success?` and `errors` so the controller can decide how
to respond.
- All the validations are now in one place and easily understandable
using the Rails validation conventions.
- No more raising exceptions for validation errors, making the 
controller faster

Resolves #11354


app/validators/decision_document_validator.rb
```diff
@@ -0,0 +1,23 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    validate :unique_citation_number
+    validates :citation_number, format: { with: /\AA?\d{8}\Z/i }
+    validates :decision_date, :redacted_document_location, :file, presence: true
+  end
+
+  def unique_citation_number?
+    DecisionDocument.find_by(appeal: appeal, citation_number: citation_number).nil?
```
aroltsch: I thought that citation number must be uniq in general and not on per appeal basis?

monfresh: You are absolutely right! Thank you for catching this. I will write a spec where the citation number exists on a different appeal.

monfresh: Fixed.

app/controllers/idt/api/v1/appeals_controller.rb
```diff
@@ -23,8 +19,12 @@ def details
   end
 
   def outcode
-    BvaDispatchTask.outcode(appeal, outcode_params, user)
-    render json: { message: "Success!" }
+    result = BvaDispatchTask.outcode(appeal, outcode_params, user)
+    if result.success?
```
aroltsch: Not a huge fan of if/else statements :). How about

```
return render json: { message: "Success!" } if result.success?
render json: { message: result.errors[0] }, status: :bad_request
```

app/validators/decision_document_validator.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module DecisionDocumentValidator
+  extend ActiveSupport::Concern
+
+  included do
+    validate :unique_citation_number
+    validates :citation_number, format: { with: /\AA?\d{8}\Z/i }
+    validates :decision_date, :redacted_document_location, :file, presence: true
```
aroltsch: I personally prefer having these validations at the model level. These are basic AR validations for presence and uniqueness

monfresh: This is the only place where DecisionDocuments get created, but I think in this case it might be better to have the validations on the model. We can initialize a new decision document and check if it's valid. This would allow us to remove the extra instance variables in the dispatch classes.

monfresh: I tried to do this and ran into issues because the model has other stuff in it that triggers validations. We can avoid that by specifying `on: :create` for the citation uniqueness validation, but I think it's better to run validations only when they are needed in the context where they are needed. If the model didn't have any other things that triggered validations, then it might make sense to move them there, but I think for now, it's better to have them in the dispatch classes, which is the only place they are needed.

app/controllers/idt/api/v1/appeals_controller.rb
monfresh: Works for me. I thought you were done reviewing, so I added the ready to merge label. I will open a new PR to fix this.

monfresh: Fixed here: https://github.com/department-of-veterans-affairs/caseflow/pull/11694

Issue #10894, opened 2019-05-28
[Tech spec] Improve maintainability of tasks and task actions


lowellrex:
Maintenance and extension of the task model and task actions continues to occupy an extensive amount of developer time. This epic exists to address some of the issues that cause tasks and actions to occupy so much developer time.

### Process to add new task action

The rough outline of the process for adding a new action to a task is as follows:
1. Caseflow user group determines need for some action needing to be taken on a task.
1. Caseflow product owners are made aware of the need and communicate to developers.
1. Developers clarify requirements and create new action (if it did not exist before).
1. Developers add code to make action appear for specific task type under the correct conditions.
1. Developers write tests to confirm it appears as expected.
1. Code is released to production.

### Initiative goals

This development cycle takes weeks to complete. The goals of this initiative include the following:
1. Reduce the length of time it takes to add new tasks and actions.
1. Reduce amount of time developers spend adding new tasks and actions.
1. Increase comprehensibility of task and actions code.
1. Increase transparency of tasks and actions for folks using Caseflow.

### Proposed approach

In order to accomplish these four goals I propose we **move conditional task action logic out of code** and into configurations stored outside of the code.

A quick sketch of an administrative interface for this might look something like the following (#10893):
![image](https://user-images.githubusercontent.com/32683958/58483702-bebc4900-812e-11e9-91f4-3b8e86e2defe.png)

### Code changes
* [ ] Move conditional blocks out of [`available_actions` method](https://github.com/department-of-veterans-affairs/caseflow/blob/949b06739d7b4c98e12f846bcd762892aab23a9a/app/models/tasks/generic_task.rb#L34) and into configuration hashes for each task class (#10888).
* [ ] Add support for mutually exclusive task actions that determine their label and value based on the state of the task (#10889).
* [ ] Move task actions configurations out of code entirely and into database.
* [ ] Create front-end interface for more accessible configuration of task actions (#10893).

### Drawbacks

**Incorrectly available actions**: Currently when developers add task actions we can identify potential situations when a given action should not be available and prevent it from being available (judge checkout should not be available until attorney checkout is completed, for example). If task action configurations are more easily modified then we are more likely to end up adding task actions to . We can address this by having the task actions know what conditions they are not allowed to be applied to so that they cannot be added to sets of conditions.

[...]

PR #11664, opened 2019-08-06
Allow judges to return cases to any attorney

### LINKED ISSUE #11610: Allow judges to return attorney decision drafts to attorneys on other judge teams
Judges occasionally want to return decision drafts to attorneys who are not on their judge team (like in the case of the AOD appeal mentioned in [this slack thread](https://dsva.slack.com/archives/CHX8FMP28/p1564497688137100)). The list of attorneys available for judges to return cases to is constrained by [this function associated with the task action](https://github.com/department-of-veterans-affairs/caseflow/blob/6d482605644d357d05d28e076d98bdf09d390996/app/repositories/task_action_repository.rb#L229).

## Acceptance criteria
* [ ] Judges can return attorney decision drafts to attorneys on other judge teams

lowellrex:
Resolves #11610. This change adds attorneys from all other judge teams to the list of options when a judge selects the "Return to attorney" action for a `JudgeDecisionReviewTask`.

Before | After
--- | ---
image | image


app/repositories/task_action_repository.rb
```diff
@@ -228,7 +228,12 @@ def schedule_veteran_data(_task, _user = nil)
 
     def return_to_attorney_data(task, _user = nil)
       assignee = task.children.select { |child| child.is_a?(AttorneyTask) }.max_by(&:created_at)&.assigned_to
-      attorneys = JudgeTeam.for_judge(task.assigned_to)&.attorneys || []
+
+      judge_team = JudgeTeam.for_judge(task.assigned_to)
+
+      # Include attorneys for all judge teams in list of possible recipients so that judges can send cases to
+      # attorneys who are not on their judge team.
+      attorneys = (judge_team&.attorneys || []) + JudgeTeam.where.not(id: judge_team&.id).map(&:attorneys).flatten
```
monfresh: What do you think about taking this opportunity to move this to a separate class for easier reading and testing? Maybe something like this:
```ruby
# frozen_string_literal: true

class ReturnToAttorneyData
  def initialize(task)
    @task = task
  end

  def call
    {
      selected: assignee,
      options: attorney_options,
      type: AttorneyRewriteTask.name
    }
  end

  private

  attr_reader :task

  def assignee
    @assignee ||= task.children.select { |child| child.is_a?(AttorneyTask) }.max_by(&:created_at)&.assigned_to
  end

  def attorney_options
    attorneys.map do |attorney|
      {
        label: attorney.full_name,
        value: attorney.id
      }
    end
  end

  # Include attorneys for all judge teams in list of possible recipients so that judges can send cases to
  # attorneys who are not on their judge team.
  def attorneys
    [judge_team_attorneys, attorneys_not_on_judge_team, assignee].flatten.compact
  end

  def judge_team_attorneys
    judge_team&.attorneys
  end

  def attorneys_not_on_judge_team
    JudgeTeam.where.not(id: judge_team&.id).flat_map(&:attorneys)
  end

  def judge_team
    @judge_team ||= JudgeTeam.for_judge(task.assigned_to)
  end
end
```

spec/feature/queue/judge_checkout_flow_spec.rb
```diff
@@ -88,6 +88,33 @@
...
+
+        click_dropdown(text: other_attorney.full_name)
+        expect(page).to have_content(other_attorney.full_name)
+
+        fill_in("instructions", with: "Returning case to different attorney")
+        click_on(COPY::MODAL_SUBMIT_BUTTON)
+
+        expect(page).to have_content(COPY::ASSIGN_TASK_SUCCESS_MESSAGE % other_attorney.full_name)
+      end
+    end
```
monfresh: What do you think about replacing this feature test with a unit test? We already test the dropdown in other feature tests. As long we have a unit test that makes sure the data returned to the front end is correct, we can be confident the dropdown will have the right attorneys.

lowellrex: Smart! I missed that we already had tests for this because we didn't used the string literal directly in the test. Fixing that as well as part of this PR.

app/repositories/task_action_repository.rb
lowellrex: I'd rather keep the code in this file. One foundational ideology of the `TaskActionRepository` class is that it should be transparent. I believe the task model has suffered (been prone to bugs, hard to read, hard to learn, etc.) because we have logic that lives all over the place. So, we sought to correct some of those historical mistakes by creating this `TaskActionRepository` class to contain all additional task action data and have each function return that hash of data directly. I'd like to continue with that pattern instead of moving logic out of this file.

For what it's worth, there are two functions (`toggle_timed_hold` and `review_decision_draft`) that do not adhere to "always directly return the hash" principle. I think I was misguided in the effort that introduced them and would like to see their returning hashes returned to this file (perhaps as part of work that will come out of #10894, specifically PR #10888).

monfresh: Moving functionality out of the Task model and into the TaskActionRepository was a great idea. Over time, this repository has grown to include many different concepts that are not closely related, making it hard to see which methods relate to each other. By moving related concepts into separate classes, we adhere to the Single Responsibility Principle. This makes it easier to read and test.

By keeping all the code inside one method, it makes it hard to read, and the next time we need to make a change, we'll keep adding more lines to the same method. When it gets complex enough that Rubocop complains, which it already would in this PR if we used the default Rubocop settings (something we have a PR out for), then what would we do? One solution would be to move some of the lines into separate methods in the repository, making the repository even longer. A better solution is to move the related code to a separate class. 

A line like `attorneys = (judge_team&.attorneys || []) + JudgeTeam.where.not(id: judge_team&.id).map(&:attorneys).flatten` creates a lot of cognitive overload, as opposed to a descriptive method name. Code is read far more often than it is written, so we want to optimize for readability. Would you agree with that?

I'm happy to approve this PR for now to allow cases to move forward, and we can improve the code in another PR.

lowellrex: One common thread I have observed long-lived software created and maintained by large teams is that abstraction introduced complexity. An abstraction that seems obvious to Billie may never occur to Alex. Everybody creates their own mental models of systems.

On the Caseflow team we already expect back-end developers to learn the abstractions of ruby and Rails (and all the benefits that come with those abstractions!). Additionally, developers have to learn how we practice the MVC model, our database structure, the business problems the application solves, and so on. As a system Caseflow has an immense number of interconnected components.

I believe that the fewer component a system has, the easier it is to mentally model. A coincident belief is that small functions should be avoided until there is prevalent duplication of the code they obscure. In this particular circumstance, while I do believe the following way to rearrange this code introduces indirection and abstraction, I think both are fine.

```ruby
judge_team_attorneys = judge_team&.attorneys || []
other_judge_teams_attorneys = JudgeTeam.where.not(id: judge_team&.id).map(&:attorneys).flatten
all_attorneys_with_judge_teams_first = judge_team_attorneys + other_judge_teams_attorneys
```

monfresh: What would you do if Code Climate complained about the complexity and/or length of the method?

lowellrex: I would ignore Code Climate.

PR #11680, opened 2019-08-06
Allow judges to correct issues for hearings they were assigned to

### LINKED ISSUE #10562: Allow judges to "Correct Issues" from Case Details when they are assigned a hearing
User story: A judge needs to be able to correct AMA issues in advance of or on the day of the hearing. 

This functionality was not deployed prior to checkout to a Judge. An appeal is assigned to a judge in a hearing, but that's prior to checkout for a decision.

Investigate how easy it is to turn the "correct issues" link on for judges/attorneys when they aren't assigned a task.

## Idea
- When a judge is assigned a hearing, they can see the Correct Issues link
   - use `appeal.hearing.last.judge`

ferristseng:
Resolves #10562

### Description

  - Modify appeal request issues policy to allow the assigned judge to edit issues
  - Feature is behind `:allow_judge_edit_issues` feature toggle
  - Fix issue in test that was failing because there was an ambiguous sort order


spec/feature/queue/case_details/edit_issues_spec.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+require "support/vacols_database_cleaner"
+require "rails_helper"
+
+RSpec.feature "Edit Issues", :all_dbs do
+  before do
+    FeatureToggle.enable!(:allow_judge_edit_issues)
+  end
+
```
tomas-nava:
```suggestion

after do
  FeatureToggle.disable!(:allow_judge_edit_issues)
end

```

Don't know how important it is, but there's a pattern of disabling flags after your tests have run.

PR #11682, opened 2019-08-06
feature toggle around api v3

### LINKED ISSUE #11681: Lighthouse API - FeatureToggle around new api v3
Use FeatureToggle to prevent api work from unexpectedly entering production

erluti:
Resolves #11681

### Description
Use FeatureToggle to prevent API v3 features from being accessible before we are ready to toggle the feature.
Updated the shared controller to have a method that can be used as a before_filter to lock out actions that aren't ready for prime time.

### Acceptance Criteria
- [ ] Code compiles correctly


app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class Api::V3::DecisionReview::HigherLevelReviewsController < Api::ExternalProxyController
+  before_action :api_released?
```
leikkisa: should this be in the external proxy controller?

app/controllers/api/external_proxy_controller.rb
```diff
@@ -2,4 +2,19 @@
 
 class Api::ExternalProxyController < ActionController::Base
   protect_from_forgery with: :null_session
+
+  def api_released?
+    return true if FeatureToggle.enabled?(:external_api_released)
```
leikkisa: A little quirk about our FeatureToggle - if there's any chance the FeatureToggle would only be enabled for specific users, you'll want this to look like: `FeatureToggle.enabled?(: external_api_released, user: the_appropriate_user)` where you set the_appropriate_user, otherwise it will return false (even if it's enabled for that user).

If not, then this looks good.

app/controllers/api/external_proxy_controller.rb
```diff
@@ -2,4 +2,19 @@
...
+    render json: {
+      errors: [
+        {
+          status: "501",
+          title: "Not Implemented",
+          detail: "This endpoint is not yet supported."
+        }
+      ]
+    },
+           status: :not_implemented
```
leikkisa: The indentation here is a little funky

erluti: Code Climate was giving me a hard time with the section and I finally just did a rubocop auto-correct on it and this is what it gave me and Code Climate was OK too.  So I agree with you, but I don't know how to get Code Climate to OK it and if you have a suggested format I'd be happy to use it!

app/controllers/api/external_proxy_controller.rb
erluti: Yeah, the plan was to be a blanket disable

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
erluti: I was thinking that this would be removed from controllers as they are made available in production.
But now I'm thinking I might be using FeatureToggle wrong, since I was expected it to always be disabled in production.  And just remove the check for APIs ready to release.
Adding lots of FeatureToggles to be checked for all time doesn't seem great.  Is there another way to go about this or is the "always false in production" approach make sense?

app/controllers/api/external_proxy_controller.rb
leikkisa: Ah, okay. I will defer to the linting robots.

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
leikkisa: Okay, I was imagining the whole API would be released or not.  It does seem like a strange way to use the FeatureToggle, where the FeatureToggle would always be off in production, but you'd remove this check for the individual controllers.  I don't know how many you expect, but I would prefer having individual FeatureToggles to this approach.

If you want it to always be "on" when not in production, then there's a method called dependencies_faked? in the application controller that might be useful.

erluti: I created a FeatureToggle for each controller now, but for now it's just on HigherLevelReviewsController, I don't think we want to turn off Docs ever.

app/controllers/api/external_proxy_controller.rb
```diff
@@ -2,4 +2,27 @@
...
+  end
+
+  # This method smells of :reek:ControlParameter
+  def controller_as_api_feature_name(cntrlr = self.class.name)
+    case cntrlr
+    when "Api::V3::DecisionReview::HigherLevelReviewsController"
+      :higher_level_review_api
+    end
+  end
```
edmangimelli: Would reek be happier if this were a hash tied to a constant instead of a method?

app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
edmangimelli: >Okay, I was imagining the whole API would be released or not.

Any reason why we wouldn't want to do it this way? --a feature toggle for the whole api, that is

edmangimelli: >  Is there another way to go about this or is the "always false in production" approach make sense?

Maybe I don't follow, but I think I like where you were going --explicitly enabling our API's featuretoggle(s) in the environments we want (dev and preprod?), and then do nothing in prod (implicitly false)

erluti:
I've updated to wrap the routes in a FeatureToggle.

config/environments/test.rb
```diff
@@ -87,4 +87,6 @@
   ENV["ENABLE_PUSHGATEWAY_HEALTHCHECK"] = "true"
 
   config.active_job.queue_adapter = :test
+
+  FeatureToggle.enable!(:api_v3)
```
pkarman: this doesn't belong here. It should be placed in a `before` block (and disabled in an `after` block) in whatever relevant spec files need it.

config/routes.rb
```diff
@@ -30,9 +30,11 @@
       resources :appeals, only: :index
       resources :hearings, only: :show, param: :hearing_day
     end
-    namespace :v3 do
-      namespace :decision_review do
-        resources :higher_level_reviews, only: :create
+    if FeatureToggle.enabled?(:api_v3)
```
pkarman: great!

spec/controllers/api/external_proxy_controller_spec.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+RSpec.describe Api::ExternalProxyController, type: :controller do
```
pkarman: empty test?

config/environments/test.rb
erluti: That doesn't work because at that point the routes have already been initialized

pkarman: ah you're right! A load order gotcha.

Probably better then to put the FeatureToggle check in the API base controller as a `before_action` instead of in the routes. We need to be able to toggle it in tests in order to remain consistent with our use elsewhere.

erluti: I put the `before_action` in our generic controller and made the feature `api_v3`.  When we work our task for using the current application controller I'll move the method over and drop our generic controller.

pkarman: I think it's fine to keep the API v3 base controller with this method. It doesn't belong in `Api::ApplicationController` since it only applies to v3.

How about just renaming the base controller to `Api::V3::BaseController` and having it inherit from `Api::ApplicationController`? Then you can move the `before_action` to `Api::V3::BaseController` so it affects that whole namespace.

erluti: I've made these changes and remove the `ExternalProxyController`

app/controllers/api/v3/base_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class Api::V3::BaseController < Api::ApplicationController
+  protect_from_forgery with: :null_session
```
pkarman: is `protect_from_forgery` needed here? or does it work to inherit the parent class?

spec/controllers/api/v3/base_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        expect { JSON.parse(response.body) }.to_not raise_error
+        parsed_response = JSON.parse(response.body)
+        expect(parsed_response["errors"]).to be_a Array
+        expect(parsed_response["errors"].first).to include("status", "title", "detail")
+      end
+    end
+
+    describe "when enabled" do
+      it "should return a 200 response" do
+        FeatureToggle.enable!(:api_v3)
```
pkarman: `FeatureToggle` needs to be wrapped in a `before` and `after` block because it basically sets a global variable for the entire test suite. See examples in other specs.

erluti: I used byebug and observed that it was always disabled at the beginning of each run.  But I went ahead and added the before/after hooks.

app/controllers/api/v3/base_controller.rb
erluti: it was added originally because of brakeman, but i removed the line and updated brakeman to ignore that it doesn't have a `protect_from_forgery`

PR #11684, opened 2019-08-07
Refactor upload_document_to_vbms to service class

monfresh:
**Why**: To separate the concerns and have each related VBMS call
defined in its own class. This follows the pattern we started by moving
document uploads to the `ExternalApi::VbmsDocumentsForAppeal` class.

The main reason for this refactor is to be able to try a different
file number (using the BGS claim number) if the veteran's file number
does not exist in VBMS. We were not being consistent with trying this
alternative number.

Ulimately, we want to find a better long-term solution for trying
alternative numbers for any VBMS call where the file number is required.

### DIFF
--- a/app/services/external_api/vbms_document_uploader.rb
+++ b/app/services/external_api/vbms_document_uploader.rb
@@ -1,0 +1,99 @@
+# frozen_string_literal: true
+
+class ExternalApi::VbmsDocumentUploader
+  def initialize(file_number:, document:, vbms_client: init_vbms_client, bgs_client: init_bgs_client)
+    @file_number = file_number
+    @document = document
+    @vbms_client = vbms_client
+    @bgs_client = bgs_client
+  end
+
+  def call
+    begin
+      upload_document_using_veteran_file_number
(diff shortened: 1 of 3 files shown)


app/services/external_api/vbms_document_uploader.rb
```diff
@@ -0,0 +1,99 @@
+# frozen_string_literal: true
+
+class ExternalApi::VbmsDocumentUploader
+  def initialize(file_number:, document:, vbms_client: init_vbms_client, bgs_client: init_bgs_client)
```
aroltsch: Would vbms_client and bgs_client ever change? I think we can move the initialization to lines 7-8

monfresh: They are there for dependency injection, so we can pass in fake clients in specs. Example: https://github.com/department-of-veterans-affairs/caseflow/blob/master/spec/services/vbms_documents_for_appeal_spec.rb#L14

pkarman:
going a different direction with this.

PR #11628, opened 2019-08-01
do not slack msg in uat

pkarman:
https://dsva.slack.com/archives/C4JECDLSE/p1564686049041800

PR #11703, opened 2019-08-08
Stop sending Slack job alerts from UAT

lowellrex:
Extends the work done in #11628. This PR will make it so that jobs that run in UAT do not send any notifications to Slack. Below is a snapshot of the `#appeals-app-alerts` channel showing notifications from jobs running in UAT and Prod. The second and third notifications need developer attention, but they are sandwiched by notifications about jobs running in UAT that do not need our attention. This PR aims to reduce the amount of noise we send to the `#appeals-app-alerts` channel.

image


monfresh:
Looks great! Can we add a test please?

app/services/slack_service.rb
```diff
@@ -10,7 +10,7 @@ def initialize(url:)
   attr_reader :url
 
   def send_notification(msg, title = "", channel = DEFAULT_CHANNEL)
-    return unless url
+    return unless url && (aws_env != "uat")
```
lowellrex: We only use `SlackService` in our jobs so I think it is fine to move this behaviour into this class. Alternatively we could create a function in `CaseflowJob` that encapsulates this behaviour instead.

```ruby
def notify_slack(*args)
  return if aws_env == "uat"
  slack_service.send_notification(args)
end
```

spec/services/slack_service_spec.rb
```diff
@@ -1,10 +1,22 @@
...
   it "posts to http" do
-    slack_service = SlackService.new(url: "http://www.example.com")
     allow_any_instance_of(HTTPClient).to receive(:post).and_return("response")
     response = slack_service.send_notification("hello")
     expect(response).to eq("response")
   end
+
+  context "when it is run in the uat environment" do
+    before do
+      allow_any_instance_of(SlackService).to receive(:aws_env).and_return("uat")
```
monfresh: What do you think about stubbing `ENV` instead of calling a private method, and moving the line inside the `it` block since there is only one test is this context?
```ruby
stub_const("ENV", { "DEPLOY_ENV" => "foo" })
```
Also, in general, using `allow_any_instance_of` is discouraged:
https://relishapp.com/rspec/rspec-mocks/docs/working-with-legacy-code/any-instance

spec/services/slack_service_spec.rb
```diff
@@ -1,10 +1,22 @@
...
+    end
+
+    it "does not make post request" do
+      expect_any_instance_of(HTTPClient).to_not receive(:post)
```
monfresh: I know this was already here, so it's not a suggestion for this PR, but in general, we want to avoid testing implementation details. If we end up using a different library to make HTTP calls, we'll have to change our tests. A better solution would be to use dependency injection to specify the HTTP client, so that the specs can pass in a barebones class that responds to things like `post` and `get`, which are fairly universal methods among the various clients out there.

PR #11706, opened 2019-08-08
Get Facility Ids to Geomatch from Outside state

### LINKED ISSUE #11063: Geo matching to closest RO office, irrespective of state

lomaxap:
Resolves #11063


client/constants/REGIONAL_OFFICE_INFORMATION.json
```diff
@@ -573,7 +573,7 @@
     "state": "DE",
     "timezone": "America/New_York",
     "hold_hearings": false,
-    "facility_locator_id": "vba_460",
+    "facility_locator_id": null,
```
lomaxap: remove Delaware from facility locator options

lomaxap:
> Is the idea that all the facilities are valid options now, so `facility_ids_to_geomatch` returns every RO facility id?

right, rather than just the RO facility IDs for a state. 

> Would you expect the performance of `VADotGovService.get_distance` to still be pretty similar?

i'd imagine it might take marginally more time, but not too concerned about it since this step only happens during a job

PR #11708, opened 2019-08-08
[WIP] Pulling Hearings Workflow logic out of Tasks#update_from_params

lomaxap:
In service of #12649 

Hearing Tasks (`ScheduleHearingTask`, `HearingTask`, `AssignHearingDispositionTask`, and `ChangeHearingDispositionTask`) makes awkward use of the `TaskControllers` update method and params to implement hearings-related workflows. 

As a result, it's often difficult to understand how hearings data is passed through task controllers, what exactly the impact of that data will be, and how the entire hearings task tree works. This PR attempts to simplify this by moving all workflow logic outside of `Task` models and into a `Hearings::WorkflowManager`, `Hearings::Schedule` and `Hearings::Disposition`. 

`HearingTasksController` provides a simple interface for `Task` modals to interact with and kick off hearings task logic. You can think of each method inside `HearingTasksController` as a thin layer of logic on top of `TasksController#update`. 

The general idea here is that `HearingTaskController method` to `WorkflowManager method` to `hearing task available action` should be 1:1:1. Each action has a corresponding route. 

With this update, we can then remove all hearings logic from `update_from_params` inside `Task` models, meaning `TaskController::update` method only updates the tasks's status ✨✨✨. After this PR is merged, we can slowly begin updating modals to point to new endpoints.


config/routes.rb
```diff
@@ -154,6 +154,17 @@
   get 'hearings/queue/appeals/:vacols_id', to: 'queue#index'
   get 'hearings/find_closest_hearing_locations', to: 'hearings#find_closest_hearing_locations'
 
+  scope path: "hearings/tasks/:id" do
+    get "/schedule", to: "hearing_tasks#schedule"
+    get "/withdraw", to: "hearing_tasks#withdraw"
+    get "/hold", to: "hearing_tasks#hold"
+    get "/cancel", to: "hearing_tasks#cancel"
+    get "/no_show", to: "hearing_tasks#now_show"
```
tomas-nava:
```suggestion
    get "/no_show", to: "hearing_tasks#no_show"
```

ferristseng:
I think this looks really good! I think it gives you a much better sense of all the actions you can take with regards to hearings tasks.

One thing that I think that could be more clear is which kinds of hearing tasks you can pass as input to `Hearings::Scheduler` and `Hearings::WorkflowManager`. Maybe some light documentation or type checking would make it more clear if that's the case. I'm assuming the scheduler works with schedule hearing tasks, and the workflow manager works with others, but I'm not sure if that's accurate.

Also, will there have to be some light glue connecting the workflow / scheduler classes to the task?

pkarman:
@ferristseng tagged you to shepherd this one.

ferristseng:
Not sure what still needed to be done for this, so I'm going to close it. I would like to revisit doing this refactor in the future since it would be nice to have all of the hearings actions in one place, but we do have hearings documentation now for #12649 (https://github.com/department-of-veterans-affairs/caseflow/wiki/Caseflow-Hearings)

PR #11709, opened 2019-08-08
Include "Assign team cases" dropdown for all vacols judges

### LINKED ISSUE #11570: Give acting judges dropdown link to assign cases view
#10719 unified the way we present queue dropdowns on the judge queue view to use organizations instead of the user's ID. However, one side effect of that change was acting judges no longer being presented with the dropdown if the person is identified as an acting judge in VACOLS but does not have a `JudgeTeam` on the Caseflow side (as discovered in #11567). This ticket exists to make it so that acting judges can navigate to the assign cases view in these circumstances.

## Acceptance criteria
* [ ] Acting judges are presented with a dropdown menu on their queue table view to visit the assign cases page.

hschallhorn:
Resolves #11570

### Description
Displays the "Assign cases" dropdown for any user with a judge vacols role.

|Before|After|
|---|---|
|Screen Shot 2019-08-08 at 3 13 16 PM|Screen Shot 2019-08-08 at 3 21 29 PM|

### Acceptance Criteria
- [ ] Acting judges are presented with a dropdown menu on their queue table view to visit the assign cases page

### Testing Plan
1. Pick a user that is not a judge in VACOLS (I'll be using BVADREINGER!)
```ruby
> User.find_by(css_id: "BVADREINGER").judge_in_vacols?
=> false
```
2. Visit their queue and confirm there is no "Assign team cases" option if they have a "Switch views" dropdown
3. Give this user an "Acting judge" role in vacols
```ruby
> FactoryBot.create(:staff, :attorney_judge_role, user: User.find_by(css_id: "BVADREINGER"))
> User.find_by(css_id: "BVADREINGER").judge_in_vacols?
=> true
```
4. Visit the user's queue again, ensure they have a "Switch views" dropdown with an "Assign team cases" option, and ensure clicking this takes you to their judge "Assign cases" view

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -247,11 +247,10 @@ class User < ApplicationRecord
 
   def selectable_organizations
     orgs = organizations.select(&:selectable_in_queue?)
-    judge_team = JudgeTeam.for_judge(self)
 
-    if judge_team
+    if judge_in_vacols?
       orgs << {
-        id: judge_team.id,
+        id: (JudgeTeam.for_judge(self) || JudgeTeam.create_for_judge(self)).id,
         name: "Assign",
         url: format("queue/%s/assign", id)
       }
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -250,7 +250,6 @@ class User < ApplicationRecord
 
     if judge_in_vacols?
       orgs << {
-        id: (JudgeTeam.for_judge(self) || JudgeTeam.create_for_judge(self)).id,
         name: "Assign",
         url: format("queue/%s/assign", id)
       }
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -248,7 +248,7 @@ class User < ApplicationRecord
   def selectable_organizations
     orgs = organizations.select(&:selectable_in_queue?)
 
-    if judge_in_vacols?
+    if JudgeTeam.for_judge(self) || judge_in_vacols?
       orgs << {
         name: "Assign",
         url: format("queue/%s/assign", id)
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -247,11 +247,9 @@ def user_info_for_idt
 
   def selectable_organizations
     orgs = organizations.select(&:selectable_in_queue?)
-    judge_team = JudgeTeam.for_judge(self)
 
-    if judge_team
+    if judge_in_vacols?
```
lowellrex: I like this approach! Maybe we could check for either condition, are they a judge in VACOLS or do they have a `JudgeTeam`. There may be no folks with judge teams who are not identified as judges in VACOLS, but allowing either condition will pave the way for us to eventually [get rid of our reliance on VACOLS for user roles](https://github.com/department-of-veterans-affairs/caseflow/issues/8436) (which will eventually happen I swear!)

app/models/user.rb
```diff
@@ -247,11 +247,9 @@ def user_info_for_idt
 
   def selectable_organizations
     orgs = organizations.select(&:selectable_in_queue?)
-    judge_team = JudgeTeam.for_judge(self)
 
-    if judge_team
+    if JudgeTeam.for_judge(self) || judge_in_vacols?
       orgs << {
-        id: judge_team.id,
```
lowellrex: Can we also remove the ID field from [the array we pass to the front-end](https://github.com/department-of-veterans-affairs/caseflow/blob/cddd5d7730f0225a8e7cb7d1ddb2f10cbf94cd04/app/views/queue/index.html.erb#L7) as well? It does not appear as if we actually use the ID field anywhere on the front-end so it'd be nice to not pass it for any organization so we don't mistakenly think that we do use it on the front-end.

PR #11712, opened 2019-08-08
11440 Search by Docket Number

### LINKED ISSUE #11440: Case search | Enhance to be able to search by docket number
Allow users to search for Veterans' cases by docket number.

## AC
- users can search by docket number for both AMA and legacy appeals

jcq:
Resolves #11440 

### Description
Added support for searching for cases by docket number in addition to existing veteran SSN or file number. This includes making both frontend and backend code more generic, supporting both kinds of searches via the existing /appeals endpoint.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...

### DIFF
--- a/app/controllers/appeals_controller.rb
+++ b/app/controllers/appeals_controller.rb
@@ -10,13 +10,17 @@ class AppealsController < ApplicationController
     respond_to do |format|
       format.html { render template: "queue/index" }
       format.json do
-        veteran_file_number = request.headers["HTTP_VETERAN_ID"]
+        case_search = request.headers["HTTP_CASE_SEARCH"]
 
-        result = CaseSearchResultsForVeteranFileNumber.new(
-          file_number_or_ssn: veteran_file_number, user: current_user
+        result1 = CaseSearchResultsForVeteranFileNumber.new(
+          file_number_or_ssn: case_search, user: current_user
         ).call
 
-        render_search_results_as_json(result)
+        result2 = CaseSearchResultsForDocketNumber.new(
+          docket_number: case_search, user: current_user
+        ).call
+
+        render_combined_results_as_json([result1, result2])
       end
     end
   end
@@ -132,6 +136,34 @@ class AppealsController < ApplicationController
       render json: result.extra[:search_results]
     else
       render json: result.to_h, status: result.extra[:status]
+    end
+  end
+
+  def render_combined_results_as_json(items)
+    success = false
+    errors = []
+    search_results = {
+      appeals: [],
+      claim_reviews: []
+    }
+    status = nil
+
+    items.each do |item|
+      hashed = item.to_h
+
+      if item.success?
+        success = true
+        search_results.each { |key, _val| search_results[key] += hashed[:search_results][key] }
+      else
+        errors += hashed[:errors]
+        status = item.extra[:status]
+      end
+    end
+
+    if success
+      render json: search_results
+    else
+      render json: { errors: errors }, status: status
     end
   end
 
--- a/app/services/appeal_finder.rb
+++ b/app/services/appeal_finder.rb
@@ -23,6 +23,25 @@ class AppealFinder
     )
   end
 
+  def find_appeals_by_docket_number(docket_number)
+    return [] if docket_number.empty?
+
+    # Take the first six digits as date, remaining as ID
+    parsed = docket_number.split("-")
+
+    # If we can't parse a valid date from search, return no results
+    begin
+      receipt_date = Date.strptime(parsed[0], "%y%m%d")
+
+      id = parsed[1]
+      appeals = Appeal.where(id: id, receipt_date: receipt_date)
+
+      return appeals
+    rescue ArgumentError => e
+      return []
+    end
+  end
+
   class << self
     def find_appeals_with_file_numbers(file_numbers)
       return [] if file_numbers.empty?
--- a/app/workflows/case_search_results_for_docket_number.rb
+++ b/app/workflows/case_search_results_for_docket_number.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+class CaseSearchResultsForDocketNumber < ::CaseSearchResultsBase
+  validate :docket_number_presence
+  validate :veterans_exist, if: :current_user_is_vso_employee?
+
+  def initialize(docket_number:, user:)
+    super(user: user)
+    @docket_number = docket_number.to_s if docket_number
+  end
+
+  protected
+
+  def appeals
+    @appeals = AppealFinder.new(user: user).find_appeals_by_docket_number(docket_number)
+    return @appeals
+  end
+
+  def claim_reviews
+    veteran_file_numbers = veterans_user_can_access.map(&:file_number)
+
+    ClaimReview.find_all_visible_by_file_number(*veteran_file_numbers)
+  end
+
+  private
+
+  attr_reader :docket_number
+
+  def docket_number_presence
+    return if docket_number
+
+    errors.add(:workflow, missing_docket_number_error)
+    @status = :bad_request
+  end
+
+  def missing_docket_number_error
+    {
+      "title": "Docket number missing",
+      "detail": "HTTP_CASE_SEARCH request header must include docket number"
+    }
+  end
+
+  def veterans_exist
+    return unless veterans_user_can_access.empty?
+
+    errors.add(:workflow, not_found_error)
+    @status = :not_found
+  end
+
+  def not_found_error
+    {
+      "title": "Docket ID not found",
+      "detail": "Could not find a case matching the docket number"
+    }
+  end
+
+  def veterans
+    # Determine vet that corresponds to docket number so we can validate user can access
+    @file_numbers_for_appeals ||= appeals.map(&:veteran_file_number).uniq
+    @veterans ||= VeteranFinder.find_or_create_all(@file_numbers_for_appeals)
+  end
+end
--- a/client/COPY.json
+++ b/client/COPY.json
@@ -1,9 +1,9 @@
 {
   "CASE_SEARCH_HOME_PAGE_HEADING": "Veteran Case Search",
-  "CASE_SEARCH_INPUT_PLACEHOLDER": "Enter a Veteran ID or SSN",
-  "CASE_SEARCH_INPUT_INSTRUCTION": "Please enter a valid claims file number or SSN to search for all available cases.",
+  "CASE_SEARCH_INPUT_PLACEHOLDER": "Enter a Veteran ID, SSN, or docket number",
+  "CASE_SEARCH_INPUT_INSTRUCTION": "Please enter a valid claims file number, SSN, or docket number to search for all available cases.",
   "CASE_SEARCH_ERROR_EMPTY_SEARCH_TERM": "Cannot search for no text",
-  "CASE_SEARCH_ERROR_INVALID_ID_HEADING": "Invalid Veteran ID “%s”",
+  "CASE_SEARCH_ERROR_INVALID_ID_HEADING": "Invalid search parameter “%s”",
   "CASE_SEARCH_ERROR_NO_CASES_FOUND_HEADING": "No cases found for “%s”",
   "CASE_SEARCH_ERROR_UNKNOWN_ERROR_HEADING": "Server encountered an error searching for “%s”",
   "CASE_SEARCH_ERROR_UNKNOWN_ERROR_MESSAGE": "Please retry your search and contact support if errors persist.",
--- a/client/app/queue/CaseList/CaseListActions.js
+++ b/client/app/queue/CaseList/CaseListActions.js
@@ -2,14 +2,8 @@ import { SEARCH_ERROR_FOR } from '../constants';
 import ApiUtil from '../../util/ApiUtil';
 import * as Constants from './actionTypes';
 import _ from 'lodash';
-import {
-  onReceiveAppealDetails,
-  onReceiveClaimReviewDetails
-} from '../QueueActions';
-import {
-  prepareAppealForStore,
-  prepareClaimReviewForStore
-} from '../utils';
+import { onReceiveAppealDetails, onReceiveClaimReviewDetails } from '../QueueActions';
+import { prepareAppealForStore, prepareClaimReviewForStore } from '../utils';
 
 export const clearCaseListSearch = () => ({
   type: Constants.CLEAR_CASE_LIST_SEARCH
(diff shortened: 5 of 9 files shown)


app/workflows/case_search_results_for_docket_number.rb
```diff
@@ -0,0 +1,62 @@
...
+      "title": "Docket ID not found",
+      "detail": "Could not find a case matching the docket number"
+    }
+  end
+
+  def veterans
+    # Determine vet that corresponds to docket number so we can validate user can access
+    @file_numbers_for_appeals ||= appeals.map(&:veteran_file_number).uniq
+    @veterans ||= VeteranFinder.find_or_create_all(@file_numbers_for_appeals)
+  end
```
leikkisa: I think since there is only one appeal, you could use the access? method on just that file_number before returning the appeal, and you don't need this method.

client/app/queue/CaseList/CaseListActions.js
```diff
@@ -90,38 +84,44 @@ export const fetchAppealUsingBackendError = (searchQuery, error) => ({
...
 
         if (returnedObject) {
-          isResponseEmpty = _.size(returnedObject.appeals) === 0 &&
-            _.size(returnedObject.claim_reviews) === 0;
+          isResponseEmpty = _.size(returnedObject.appeals) === 0 && _.size(returnedObject.claim_reviews) === 0;
         }
 
         if (!returnedObject || isResponseEmpty) {
-          dispatch(fetchedNoAppealsUsingVeteranId(veteranId));
+          dispatch(fetchedNoAppealsUsingVeteranId(search));
```
leikkisa: Not your fault but the naming of things is getting confusing to me. We're using veteranId to represent file number, or SSN, and now docket number.  Elsewhere sometimes we mean the ID to mean the Caseflow ID.

client/COPY.json
```diff
@@ -1,9 +1,9 @@
 {
   "CASE_SEARCH_HOME_PAGE_HEADING": "Veteran Case Search",
-  "CASE_SEARCH_INPUT_PLACEHOLDER": "Enter a Veteran ID or SSN",
-  "CASE_SEARCH_INPUT_INSTRUCTION": "Please enter a valid claims file number or SSN to search for all available cases.",
+  "CASE_SEARCH_INPUT_PLACEHOLDER": "Enter a Veteran ID, SSN, or AMA docket number",
```
leikkisa: Can we call this file number here instead of ID?

client/COPY.json
```diff
@@ -1,9 +1,9 @@
...
+  "CASE_SEARCH_INPUT_INSTRUCTION": "Please enter a valid claims file number, SSN, or AMA docket number to search for all available cases.",
   "CASE_SEARCH_ERROR_EMPTY_SEARCH_TERM": "Cannot search for no text",
-  "CASE_SEARCH_ERROR_INVALID_ID_HEADING": "Invalid Veteran ID “%s”",
+  "CASE_SEARCH_ERROR_INVALID_ID_HEADING": "Invalid search parameter “%s”",
```
leikkisa: 👍

client/app/queue/CaseList/CaseListActions.js
```diff
@@ -90,38 +84,44 @@ export const fetchAppealUsingBackendError = (searchQuery, error) => ({
...
       return reject();
     }
 
-    const veteranId = searchQuery.replace(/\D/g, '');
-    // Allow for SSNs (9 digits) as well as claims file numbers (7 or 8 digits).
+    // Allow numbers + hyphen (for docket number)
+    const search = searchQuery.trim().replace(/[^\d-]/g, '');
 
-    if (!veteranId.match(/\d{7,9}/)) {
+    // We should probably refactor these to be imports
```
leikkisa: Yeah seems useful. Nice additions though.

client/app/queue/CaseList/CaseListActions.js
```diff
@@ -90,38 +84,44 @@ export const fetchAppealUsingBackendError = (searchQuery, error) => ({
...
+    if (!validInput(search)) {
       dispatch(fetchAppealUsingInvalidVeteranIdFailed(searchQuery));
 
       return reject();
     }
 
     dispatch(requestAppealUsingVeteranId());
     ApiUtil.get('/appeals', {
-      headers: { 'veteran-id': veteranId }
+      headers: { 'case-search': search }
```
leikkisa: searchTerm would be clearer to me (my brain thinks of search as a verb) but don't feel strongly about it.

app/controllers/appeals_controller.rb
```diff
@@ -10,13 +10,17 @@ def index
     respond_to do |format|
       format.html { render template: "queue/index" }
       format.json do
-        veteran_file_number = request.headers["HTTP_VETERAN_ID"]
+        case_search = request.headers["HTTP_CASE_SEARCH"]
```
leikkisa: 👍

app/workflows/case_search_results_for_docket_number.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  protected
+
+  def appeals
+    @appeals = AppealFinder.new(user: user).find_appeals_by_docket_number(docket_number)
+    return @appeals
+  end
+
+  def claim_reviews
```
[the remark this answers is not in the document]
jcq: Ah, k.  I wasn't entirely certain how/if I should handle that, so good to know it's not applicable.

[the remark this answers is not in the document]
jcq: Actually, I still need a stub for it, due to the parent class expecting its presence. We could probably refactor things more in the future, depending on what other types of searches we might want to do and what their needs/returns might be.

app/services/appeal_finder.rb
```diff
@@ -23,6 +23,25 @@ def find_appeals_by_ssn_or_file_number(file_number_or_ssn)
...
+
+    # Take the first six digits as date, remaining as ID
+    parsed = docket_number.split("-")
+
+    # If we can't parse a valid date from search, return no results
+    begin
+      receipt_date = Date.strptime(parsed[0], "%y%m%d")
+
+      id = parsed[1]
+      appeals = Appeal.where(id: id, receipt_date: receipt_date)
```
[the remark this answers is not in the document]
jcq: Good point — I'm all for more efficient DB queries. The base class expects that `appeals` will be dealing with an array, but I can use Array.wrap() around the resulting single record.

app/workflows/case_search_results_for_docket_number.rb
jcq: Unfortunately, this is another case where the base class expects to be able to use the `veterans` variable (used in the `veterans_user_can_access` base class method). This does seem a bit superfluous since there will only be one element in the appeals array, but since `VeteranFinder.find_or_create_all` takes an array as an argument, it might actually be more verbose to not just keep everything as an array.

Thoughts?

app/controllers/appeals_controller.rb
```diff
@@ -10,13 +10,17 @@ def index
...
 
-        result = CaseSearchResultsForVeteranFileNumber.new(
-          file_number_or_ssn: veteran_file_number, user: current_user
+        result1 = CaseSearchResultsForVeteranFileNumber.new(
+          file_number_or_ssn: case_search, user: current_user
         ).call
 
-        render_search_results_as_json(result)
+        result2 = CaseSearchResultsForDocketNumber.new(
```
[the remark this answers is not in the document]
jcq: Yeah, it does seem a bit weird at first. The basic logic I was following was that we would check for invalid input client-side (via regex matches on ssn, file number, and docket number), then simply search based on all possible criteria and combine the results. If any of them succeed, then the entire operation succeeds and results are returned (the results could be empty as long as an error didn't occur). If all fail (due to server error or something), we return error.

Happy to rethink this as you see fit, however. Since my knowledge of the backend processes is still limited, I was aiming to fit this in with as little disruption to existing code as possible.

client/app/queue/CaseList/CaseListActions.js
jcq: I'm fine with changing this. Did you mean just updating the local variable, or do you also prefer we change to 'search-term' header as well?

client/COPY.json
jcq: Yup! Agree that makes it less confusing

app/services/appeal_finder.rb
```diff
@@ -23,6 +23,25 @@ def find_appeals_by_ssn_or_file_number(file_number_or_ssn)
     )
   end
 
+  def find_appeals_by_docket_number(docket_number)
+    return [] if docket_number.empty?
+
+    # Take the first six digits as date, remaining as ID
+    parsed = docket_number.split("-")
```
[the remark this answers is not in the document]
jcq: That is a good question; currently all the logic assumes (and relies on) the presence of the hyphen for docket numbers. Perhaps for now we should think about ways to indicate that is required in the help text...?

app/controllers/appeals_controller.rb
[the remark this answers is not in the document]
jcq: I ended up going with your suggestion here and just doing a regex check on the input, as that ended up greatly simplifying things. We will likely need to reassess the next time we add a new search type, but we can put that off until then.

client/app/queue/CaseList/CaseListActions.js
jcq: Did end up refactoring these

app/workflows/case_search_results_for_docket_number.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def appeals
+    Array.wrap(AppealFinder.new(user: user).find_appeal_by_docket_number(docket_number))
+  end
+
+  private
+
+  attr_reader :docket_number
+
+  def docket_number_presence
```
leikkisa: This is prevented in multiple places, on the front end, and in the controller. So this and the missing_docket_number error would never happen. On one hand, it feels like "good measure". On the other hand, it's never going to happen and feels like bloat (and could potentially be confusing).  I don't think you could get a test that shows this error.

I think this is an example of why it might be good to put the logic for CaseSearchResultsForDocketNumber into CaseSearchResultsForVeteranFileNumber and rename it (CaseSearchResultsForSearchTerm ?)

app/services/appeal_finder.rb
```diff
@@ -23,6 +23,33 @@ def find_appeals_by_ssn_or_file_number(file_number_or_ssn)
...
+
+      appeal
+      # # If VSO user, check to make sure they have rights to the requested appeal
+      # if user.vso_employee?
+      #   veterans = VeteranFinder.find_or_create_all(appeal.veteran_file_number)
+
+      #   filter_appeals_for_vso_user(appeals: Array.wrap(appeal), veterans: veterans)
+      # else
+      #   appeal
+      # end
```
leikkisa: Can this stuff go away?

client/app/queue/CaseList/CaseListActions.js
```diff
@@ -90,65 +92,70 @@ export const fetchAppealUsingBackendError = (searchQuery, error) => ({
...
+export const fetchAppealsBySearch = (searchTerm) => (dispatch) => {
+  return ApiUtil.get('/appeals', { headers: { 'case-search': searchTerm } }).
+    then((response) => {
+      const returnedObject = response.text ? JSON.parse(response.text) : null;
+      const isResponseEmpty = returnedObject && !size(returnedObject.appeals) && !size(returnedObject.claim_reviews);
 
-    if (!veteranId.match(/\d{7,9}/)) {
-      dispatch(fetchAppealUsingInvalidVeteranIdFailed(searchQuery));
+      if (!returnedObject || isResponseEmpty) {
+        dispatch(fetchedNoAppealsUsingVeteranId(searchTerm));
```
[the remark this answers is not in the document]
jcq: Yeah, totally agree — my brain did want to rename some of those actions, etc now that they are more generic, but I was trying to balance that with trying not to let myself get too far down the rabbit hole w/ this PR

That said, I fell quite a bit further down the rabbit hole than any of us expected when this ticket was assigned, so wouldn't really hurt to fix up some of the remaining nomenclature.

spec/feature/queue/search_spec.rb
```diff
@@ -396,6 +397,61 @@ def perform_search
...
   end
 
+  context "queue case search for appeals using docket number" do
+    let!(:veteran) { create(:veteran, first_name: "Testy", last_name: "McTesterson") }
+    let!(:appeal) { create(:appeal, veteran: veteran) }
+
+    context "when using invalid docket number" do
+      def perform_search
+        visit "/search"
+        fill_in "searchBarEmptyList", with: invalid_docket_number
```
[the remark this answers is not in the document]
jcq: Hah, totally. I think that ended up being an issue with copying and pasting a bunch of times before really considering how to DRY out the code. :)

spec/feature/queue/search_spec.rb
```diff
@@ -396,6 +397,61 @@ def perform_search
...
+      end
+    end
+
+    context "when appeal exists" do
+      def perform_search
+        visit "/search"
+        fill_in "searchBarEmptyList", with: appeal.docket_number
+        click_on "Search"
+      end
+
```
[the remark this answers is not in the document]
jcq: Weird — how did that not get caught / come up when running the tests?

app/services/appeal_finder.rb
jcq: Yup! Thought I had cleared that out. Good catch!

client/app/queue/CaseList/CaseListActions.js
```diff
@@ -90,65 +92,70 @@ export const fetchAppealUsingBackendError = (searchQuery, error) => ({
...
 
-      return reject();
-    }
+        return Promise.reject();
+      }
 
-    dispatch(requestAppealUsingVeteranId());
-    ApiUtil.get('/appeals', {
-      headers: { 'veteran-id': veteranId }
+      dispatch(onReceiveAppealsUsingVeteranId(returnedObject.appeals));
```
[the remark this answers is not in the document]
jcq: Sounds good

app/services/appeal_finder.rb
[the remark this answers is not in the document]
jcq: I updated the copy to this:
"Please enter a valid claims file number, SSN, or AMA docket number (with hyphen) to search for all available cases."

Think that is clear enough to users?

app/workflows/case_search_results_for_docket_number.rb
jcq: This should now be merged into a single CaseResultsforSearchTerm class!

app/workflows/case_search_results_for_docket_number.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def appeals
+    Array.wrap(AppealFinder.find_appeal_by_docket_number(docket_number))
+  end
+
+  private
+
+  attr_reader :docket_number
+
+  def docket_number_presence
```
leikkisa: I would delete this method

app/workflows/case_search_results_for_docket_number.rb
```diff
@@ -0,0 +1,66 @@
...
+    return if docket_number
+
+    errors.add(:workflow, missing_docket_number_error)
+    @status = :bad_request
+  end
+
+  def missing_docket_number_error
```
leikkisa: And then you can delete this too

app/workflows/case_search_results_for_docket_number.rb
```diff
@@ -0,0 +1,66 @@
...
+    {
+      "title": "Docket number missing",
+      "detail": "HTTP_CASE_SEARCH request header must include docket number"
+    }
+  end
+
+  def veterans_exist
```
leikkisa: You can delete this since it's in the base class

app/workflows/case_search_results_for_docket_number.rb
```diff
@@ -0,0 +1,66 @@
...
+    }
+  end
+
+  def veterans
+    # Determine vet that corresponds to docket number so we can validate user can access
+    @file_numbers_for_appeals ||= appeals.map(&:veteran_file_number)
+    @veterans ||= VeteranFinder.find_or_create_all(@file_numbers_for_appeals)
+  end
+
+  def prohibited_error
```
leikkisa: You can delete this, it's handled in the base class. The same error message should be valid still.

PR #11714, opened 2019-08-09
Remove use_bgs feature flag and refactor rep code into its own class

### LINKED ISSUE #11016: remove the use_representative_info_from_bgs feature flag
The `use_representative_info_from_bgs` flag is on by default for all users, so we should stop checking for it in code.

aroltsch:
Resolves #11016 

### Description
1. Remove `use_representative_info_from_bgs` feature flag
2. Move legacy appeal representative logic into its own class


app/workflows/legacy_appeal_representative_fetcher.rb
```diff
@@ -0,0 +1,60 @@
...
+  REPRESENTATIVE_METHOD_NAMES = [
+    :representative_name,
+    :representative_type,
+    :representative_address
+  ].freeze
+
+  REPRESENTATIVE_METHOD_NAMES.each do |method_name|
+    define_method(method_name) do
+      if use_representative_info_from_bgs?
+        power_of_attorney.send("bgs_#{method_name}".to_sym)
```
monfresh: If these are public methods on `power_of_attorney`, I think it's recommended to use `public_send`.

app/workflows/legacy_appeal_representative_fetcher.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+class LegacyAppealRepresentativeFetcher
```
monfresh: What do you think about dropping the `Fetcher` from the name and moving this to the `models` folder? This class represents a representative and it returns methods that correspond to the attributes of the rep. I don't think it represents a particular workflow.

aroltsch: I like this suggestion!

PR #11728, opened 2019-08-12
Hide view docs button in Case Details for RO and call center users

### LINKED ISSUE #9969: Update Case Details to remove "View Docs" link so ROs, call center folks, don't see it
Remove the "View Docs" link in Case Details for Regional Offices that are accessing their view of Caseflow Hearing Schedule (CSEM function, "RO ViewHearSched") 

Note that the user is prevented from accessing reader. They get a Drat message, but it would be more ideal if the link didn't appear.

tomas-nava:
Resolves #9969

### Description
Don't show the Veteran Documents link in Case Details if a user has the "Case Details" or "RO ViewHearSched" CSEM functions.

**Note:** The link should've already been hidden for "Case Details" users; I updated a flakey test that exercises this and confirmed in production with a selection of users.

Also renames variables and refactors some aspects of passing user roles to the front end.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -18,9 +18,6 @@ class User < ApplicationRecord
   # Ephemeral values obtained from CSS on auth. Stored in user's session
   attr_writer :regional_office
 
-  FUNCTIONS = ["Establish Claim", "Manage Claim Establishment", "Certify Appeal",
-               "Reader", "Hearing Prep", "Mail Intake", "Admin Intake", "Case Details"].freeze
-
   # Because of the function character limit, we need to also alias some functions
   FUNCTION_ALIASES = {
     "Manage Claims Establishme" => ["Manage Claim Establishment"],
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -71,6 +71,10 @@ class User < ApplicationRecord
     can_any_of_these_roles?(["Build HearSched", "Edit HearSched", "RO ViewHearSched", "VSO", "Hearing Prep"])
   end
 
+  def can_view_hearing_schedule?
+    can?("RO ViewHearSched") && !can?("Build HearSched") && !can?("Edit HearSched")
+  end
+
   def administer_org_users?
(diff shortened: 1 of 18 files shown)


app/models/user.rb
```diff
@@ -18,9 +18,6 @@ class User < ApplicationRecord
   # Ephemeral values obtained from CSS on auth. Stored in user's session
   attr_writer :regional_office
 
-  FUNCTIONS = ["Establish Claim", "Manage Claim Establishment", "Certify Appeal",
-               "Reader", "Hearing Prep", "Mail Intake", "Admin Intake", "Case Details"].freeze
-
```
tomas-nava: This is out of date and doesn't seem to be used anywhere.

pkarman: possibly related to the caseflow-commons app/services/functions.rb lib but I don't see how. If we remove it and nothing breaks, +1

app/views/hearings/index.html.erb
```diff
@@ -14,15 +14,12 @@
...
-      !current_user.can?('Build HearSched') && !current_user.can?('Edit HearSched'),
-  userRoleHearingPrep: current_user.can?('Hearing Prep'),
-  userInHearingsOrganization: (HearingsManagement.singleton.users.include?(current_user) ||
-      TranscriptionTeam.singleton.users.include?(current_user)),
+  userCanAssignHearingSchedule: current_user.can_assign_hearing_schedule?,
+  userCanBuildHearingSchedule: current_user.can?('Build HearSched'),
+  userCanViewHearingSchedule: current_user.can_view_hearing_schedule?,
+  userCanVsoHearingSchedule: current_user.can_vso_hearing_schedule?,
+  userHasHearingPrepRole: current_user.can?('Hearing Prep'),
+  userInHearingOrTranscriptionOrganization: current_user.in_hearing_or_transcription_organization?,
```
ferristseng: Nice! These naming changes look great! They make the permissions a lot clearer

PR #11732, opened 2019-08-12
Rename advanced_on_docket to advanced_on_docket?

lomky:
`Appeal`, `Legacy Appeal`, `Hearing`, `Claimant` and `Person` objects know about their advance on docket status with a function called advance**d**_on_docket. This PR renames this to the more railsy and search-friendly version `advance_on_docket?`, bringing it in line with the majority of the code base.

This PR does not touch:
 - past migrations referencing `advanced_on_docket`
 - the modal URLs currently using 'advanced'


lowellrex:
What's the advantage of the alias instead of renaming the function?

Issue #11733, opened 2019-08-12
Split TaskPager apart


lowellrex:
While working on [returning grouped counts of tasks to the front-end](https://github.com/department-of-veterans-affairs/caseflow/issues/11509), it became increasingly apparent that the `TaskPager` class was overloaded. Specifically, it is doing 4 different jobs:
• Paginated some set of tasks
• Sorting some set of tasks
• Filtering some set of tasks into a smaller set of tasks
• Fetching tasks for a given tab

This ticket exists to untangle these jobs from one another.

**Tasks for a given tab**
The logic for fetching tasks for a given tab could be offloaded to the `QueueTab` classes. Combining this with a factory could result in code that looks like the following sketch.

```ruby
tasks_for_tab = QueueTab.from_tab_name(tab_name_param).tasks_for(assignee)
```

**Sorting tasks**
Since we sort tasks on a given column it may make sense to create `QueueColumn` classes that encapsulate this logic. We could reuse those objects to help return the grouped task counts to the front-end as part of #11509.

## Acceptance criteria
* [x] Move logic for fetching tasks for a given tab out of `TaskPager` (#11735)
* [x] Move logic for sorting tasks out of `TaskPager` (#11750)
* [x] Move logic for filtering tasks out of `TaskPager` (#11816)

PR #11735, opened 2019-08-12
Allow QueueTab to return tasks

lowellrex:
Connects #11733. Moves logic for fetching tasks for a given tab from `TaskPager` to each individual `QueueTab`.

Three major changes included in this PR:
1. `TaskPager` no longer controls which tasks a given tab is supposed to return. That responsibility is transferred to the various `QueueTab` subclasses' `tasks` method.
1. `QueueTab`s now require an `assignee` argument when they are created.
1. Added a new `QueueTab.from_name` function for returning a tab class given the tab's name string we receive from the front-end.

### DIFF
--- a/app/models/task_pager.rb
+++ b/app/models/task_pager.rb
@@ -92,62 +92,16 @@ class TaskPager
     @task_page_count ||= paged_tasks.total_pages
   end
 
-  def total_task_count
-    @total_task_count ||= tasks_for_tab.count
-  end
-
   def filtered_tasks
     where_clause = QueueWhereClauseArgumentsFactory.new(filter_params: filters).arguments
     where_clause.empty? ? tasks_for_tab : tasks_for_tab.joins(cached_attributes_join_clause).where(*where_clause)
   end
 
   def tasks_for_tab
-    case tab_name
-    when Constants.QUEUE_CONFIG.TRACKING_TASKS_TAB_NAME
-      tracking_tasks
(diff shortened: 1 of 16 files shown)


app/models/task_pager.rb
```diff
@@ -92,63 +92,17 @@ def task_page_count
     @task_page_count ||= paged_tasks.total_pages
   end
 
-  def total_task_count
```
lowellrex: Was only used in [one place in `QueueConfig`](https://github.com/department-of-veterans-affairs/caseflow/pull/11735/files#diff-0e38eeb126d1e7d54ae84ff58a647e0cL52) that has been replaced with `tab.tasks.count`.

app/models/task_pager.rb
```diff
@@ -92,63 +92,17 @@ def task_page_count
...
-    when Constants.QUEUE_CONFIG.ASSIGNED_TASKS_TAB_NAME
-      assigned_child_tasks
-    when Constants.QUEUE_CONFIG.ON_HOLD_TASKS_TAB_NAME
-      on_hold_child_tasks
-    when Constants.QUEUE_CONFIG.COMPLETED_TASKS_TAB_NAME
-      recently_completed_tasks
-    else
-      fail(Caseflow::Error::InvalidTaskTableTab, tab_name: tab_name)
-    end
+    QueueTab.from_name(tab_name).new(assignee: organization).tasks
```
lowellrex: The rest of the changes spring from this one.

PR #11741, opened 2019-08-12
Cache Hearing Day id from VACOLS

### LINKED ISSUE #11377: Add Association between HearingDay and LegacyHearing
We currently cannot retrieve a list of `LegacyHearing`s for a given `HearingDay` without retrieving `VACOLS::CaseHearing.where(vdkey: hearingday.id)`. We should add an association between `HearingDay` and `LegacyHearing` to speed up queries where a trip to VACOLS is unnecessary

ferristseng:
Resolves #11377 

### Description

  - rename `hearing_day_id` to `hearing_day_vacols_id`
  - save hearing day id from VACOLS in `hearing_day_id`
  - added some tests for `hearing_day` method on `LegacyHearing`

### DIFF
--- a/app/models/legacy_hearing.rb
+++ b/app/models/legacy_hearing.rb
@@ -108,6 +108,14 @@ class LegacyHearing < ApplicationRecord
   end
 
   def hearing_day
+    if hearing_day_id.nil?
+      begin
+        update!(hearing_day_id: hearing_day_vacols_id)
+      rescue ActiveRecord::InvalidForeignKey
+        # Hearing day doesn't exist yet in DB. Ignore for now.
+      end
+    end
+
     # access with caution. this retrieves the hearing_day_id from vacols
     # then looks up the HearingDay in Caseflow
     @hearing_day ||= HearingDay.find_by_id(hearing_day_id)
--- a/app/models/legacy_hearing.rb
+++ b/app/models/legacy_hearing.rb
@@ -118,8 +118,6 @@ class LegacyHearing < ApplicationRecord
       end
     end
 
-    # access with caution. this retrieves the hearing_day_id from vacols
-    # then looks up the HearingDay in Caseflow
     @hearing_day ||= HearingDay.find_by_id(hearing_day_id)
   end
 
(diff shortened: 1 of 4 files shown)


app/models/legacy_hearing.rb
```diff
@@ -108,6 +108,16 @@ def external_id
   end
 
   def hearing_day
+    if hearing_day_id.nil?
+      begin
+        update!(hearing_day_id: hearing_day_vacols_id)
+      rescue ActiveRecord::InvalidForeignKey
+        # Hearing day doesn't exist yet in DB. Set hearing_day_id to nil
+        # explicitly for clarity (even though we already know it's nil).
+        hearing_day_id = nil
```
ferristseng: Decided to catch this error to preserve the old behavior where `hearing_day` would be `nil` if the hearing day does not exist yet in Caseflow.

app/models/legacy_hearing.rb
```diff
@@ -108,8 +108,16 @@ def external_id
   end
 
   def hearing_day
-    # access with caution. this retrieves the hearing_day_id from vacols
-    # then looks up the HearingDay in Caseflow
+    if hearing_day_id.nil?
+      begin
+        update!(hearing_day_id: hearing_day_vacols_id)
```
lomaxap: the `vdkey` could refer to either a row in VACOLS or the HearingDay id in caseflow.

it refers to a row in VACOLS 
a) if the hearing is a central office and scheduled before 01/01/2019
or
b) if the hearing is a video hearing and was scheduled before 4/1/2019

Are you thinking you'd store both here do a check for VACOLS/Caseflow elsewhere?

ferristseng: I was thinking this would replicate the old behavior, just avoid making the VACOLS call on subsequent requests because the id would be cached in Caseflow.

I think we could potentially add that logic here. Should we not try to do the update if `hearing_day_vacols_id` refers to a VACOLS row?

lomaxap: i think so. Otherwise, we're essentially turning this into a polymorphic reference without also designating what table we're referencing (with something like a `hearing_day_type` field). 

There will also be duplicate `id` values if we ever have more than ~40,000 legacy hearings.

PR #11743, opened 2019-08-12
Refactor SlackService

monfresh:
**Why**:
- To make the tests less brittle by using dependency injection. If we
end up using a different HTTP library, such as Faraday, we won't have
to change the tests
- To eliminate the use of `allow_any_instance_of`, which is discouraged
by the RSpec maintainers
- To use named keywords so that's it's clear what each argument
represents
- To fix the FeatureEnvy code smell by making the msg, title, and
channel arguments part of the initialization as opposed to arguments
for the `send_notification` method.

This PR also improves test coverage for SlackService, which revealed
a bug when the channel argument did not start with the # character.

Further improvements could be to rename the class from the generic
`SlackService` to something more precise such as `SlackNotification`
and then rename the `send_notification` method to `call`.

### DIFF
--- a/app/jobs/caseflow_job.rb
+++ b/app/jobs/caseflow_job.rb
@@ -1,11 +1,4 @@
 # frozen_string_literal: true
 
 class CaseflowJob < ApplicationJob
-  def slack_url
-    ENV["SLACK_DISPATCH_ALERT_URL"]
-  end
-
-  def slack_service
-    @slack_service ||= SlackService.new(url: slack_url)
-  end
 end
--- a/app/services/slack_service.rb
+++ b/app/services/slack_service.rb
@@ -3,16 +3,17 @@
 class SlackService
   DEFAULT_CHANNEL = "#appeals-app-alerts"
 
-  def initialize(url:)
-    @url = url
+  def initialize(msg:, title: "", channel: DEFAULT_CHANNEL, http_service: HTTPClient.new)
+    @msg = msg
+    @title = title
+    @channel = channel
+    @http_service = http_service
   end
 
-  attr_reader :url
+  def send_notification
+    return if aws_env == "uat" || url.blank?
 
-  def send_notification(msg, title = "", channel = DEFAULT_CHANNEL)
-    return unless url && (aws_env != "uat")
-
-    slack_msg = format_slack_msg(msg, title, channel)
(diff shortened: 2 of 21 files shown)


app/jobs/caseflow_job.rb
```diff
@@ -1,11 +1,5 @@
...
 
 class CaseflowJob < ApplicationJob
-  def slack_url
-    ENV["SLACK_DISPATCH_ALERT_URL"]
-  end
 
-  def slack_service
-    @slack_service ||= SlackService.new(url: slack_url)
-  end
 end
```
monfresh: Since this is empty now, should we remove it?

app/services/slack_service.rb
```diff
@@ -3,33 +3,37 @@
 class SlackService
   DEFAULT_CHANNEL = "#appeals-app-alerts"
 
-  def initialize(url:)
-    @url = url
+  def initialize(msg:, title: "", channel: DEFAULT_CHANNEL, http_service: HTTPClient.new)
```
pkarman: it's a little weird to me that the service is now instantiated and is really only call-able once. I.e. you would never call `send_notification` more than one time.

Also, the URL is now entirely a global variable which means the service can only be used with one URL ever per env.

how about we just change the named arguments in `send_notification` (maybe rename it `send`?) and allow `initialize` to receive both `url` and/or `http_service`.

I like the ability to support dependency injection, I just worry that we effectively loose all the flexibility we had with this class.

monfresh: I thought about that, and I noticed that we were always using the same URL. In the case of Slack, I don't think there would be multiple URLs per environment, so it probably makes sense to have it in an env var. 

As for calling `send_notification` multiple times, I can see how it would be helpful to send multiple notifications without instantiating a new object each time, but that doesn't seem to be how we use it. We're using it as a one-time notification to send job errors to the `appeals-app-alerts` channel in all cases except in the `DataIntegrityChecksJob`, where one of the 3 alerts goes to the `appeals-hearings` channel.

I'm happy to put the msg, title, and channel back with the `send_notification` method if you think that's better.

pkarman: what you've observed is how we use it in production, but we use a different url when developing (points at the #dev-null slack channel) and I like the idea of keeping this more general purpose.

PR #11745, opened 2019-08-12
Bump React

lomaxap:


client/test/helpers/index.js
```diff
@@ -13,6 +11,4 @@ import _ from 'lodash';
 //
 // To work around this, we manually find the element that's an actual DOM node,
 // and not a React element.
-export const findElementById = (wrapper, id) => wrapper.
-  find({ id }).
-  findWhere((node) => _.isString(node.type()) && node.getDOMNode().id === id);
+export const findElementById = (wrapper, id) => wrapper.find({ id }).hostNodes();
```
jcq: This was a quick fix for an issue that popped up with updated version of Enzyme; it accomplishes the same thing as the previous code did.

client/test/karma/components/LoadingDataDisplay-test.js
```diff
@@ -41,6 +41,9 @@ describe('LoadingDataDisplay', () => {
     );
 
     return wait().then(() => {
+      // Force component re-render due to async setState
+      wrapper.update();
```
jcq: All of these were failing with updated Enzyme due to it not noticing a re-render after state change. This forces Enzyme to re-render, allowing tests to pass as expected.

PR #11749, opened 2019-08-12
BGSService.may_modify depends on conflict checker

### LINKED ISSUE #11736: VBMSError User Not Authorized - supplemental claim intake
https://appeals.cf.ds.va.gov/asyncable_jobs/SupplementalClaim/jobs/90312

This problem should have been prevented during intake, so this ticket is now an investigation into why not.

pkarman:
Resolves #11736 

### Description
Implements additional sensitivity checks to prevent user not authorized errors during intake establishment.

The full algorithm is in https://github.com/department-of-veterans-affairs/caseflow/issues/10087#issuecomment-507783830

Tested in production with the following user_id/veteran_participant_id pairs that had previously failed:

```ruby
tests = [
  ["2321", "3309696"],
  ["3309", "4012373"],
  ["5290", "35206608"],
]

tests.each do |pair|
  RequestStore[:current_user] = User.find pair[0]
  checker = ExternalApi::BgsVeteranStationUserConflict.new(veteran_participant_id: pair[1])
  puts "#{pair} => #{checker.conflict?}" # should return true
end
```

### DIFF
--- a/app/services/external_api/bgs_veteran_station_user_conflict.rb
+++ b/app/services/external_api/bgs_veteran_station_user_conflict.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class ExternalApi::BgsVeteranStationUserConflict
   def initialize(veteran_participant_id:, client: nil)
     @veteran_participant_id = veteran_participant_id
@@ -12,7 +14,6 @@ class ExternalApi::BgsVeteranStationUserConflict
     return true if station_dtos.any? && !veteran_dto
 
     # otherwise we must check sensitivity reason
-
   end
 
   private
@@ -36,7 +37,7 @@ class ExternalApi::BgsVeteranStationUserConflict
 
     return false unless veteran_dto
 
-    return veteran_dto[:station_number] == current_user_station
+    veteran_dto[:station_number] == current_user_station
   end
 
   def veteran_station_id
@@ -60,9 +61,11 @@ class ExternalApi::BgsVeteranStationUserConflict
   end
 
   def sensitivity_level
-    @sensitivity_level ||= MetricsService.record("BGS: fetch sensitivity level by participant id: #{veteran_participant_id}",
-                                                 service: :bgs,
-                                                 name: "security.find_sensitivity_level_by_participant_id") do
+    @sensitivity_level ||= MetricsService.record(
+      "BGS: fetch sensitivity level by participant id: #{veteran_participant_id}",
+      service: :bgs,
+      name: "security.find_sensitivity_level_by_participant_id"
+    ) do
       client.security.find_sensitivity_level_by_participant_id(veteran_participant_id)
     end
   end
--- a/app/services/external_api/bgs_veteran_station_user_conflict.rb
+++ b/app/services/external_api/bgs_veteran_station_user_conflict.rb
@@ -1,0 +1,73 @@
+class ExternalApi::BgsVeteranStationUserConflict
+  def initialize(veteran_participant_id:, client: nil)
+    @veteran_participant_id = veteran_participant_id
+    @client = client
+  end
+
+  # logic detailed in https://github.com/department-of-veterans-affairs/caseflow/issues/10087#issuecomment-507783830
+  def call
+    DBService.release_db_connections
+
+    # simple case if DTOs exist and do not contain a Veteran record
+    return true if station_dtos.any? && !veteran_dto
+
+    # otherwise we must check sensitivity reason
+
+  end
+
+  private
+
+  attr_reader :veteran_participant_id
+
+  def current_user_station
+    RequestStore[:current_user].station_id.to_s
+  end
+
+  def station_dtos
+    [employee_dtos[:station]].flatten
+  end
+
+  def veteran_dto
+    station_dtos.find { |dto| dto[:ptcpnt_rlnshp_type_nm] == "Veteran" }
+  end
+
+  def veteran_at_same_station?
+    return false unless station_dtos
+
+    return false unless veteran_dto
+
+    return veteran_dto[:station_number] == current_user_station
+  end
+
+  def veteran_station_id
+    "#{sensitivity_level[:fclty_type_cd]}#{sensitivity_level[:cd]}"
+  end
+
+  def sensitivity_reason
+    sensitivity_level[:sntvty_reason_type_nm]
+  end
+
+  def violates_sensitivity_reason?
+    ["Relative of Local VA Employee", "VBA Employee", "Veteran", "Work Study"].include?(sensitivity_reason)
+  end
+
+  def employee_dtos
+    @employee_dtos ||= MetricsService.record("BGS: fetch employee by participant id: #{veteran_participant_id}",
+                                             service: :bgs,
+                                             name: "people.find_employee_by_participant_id") do
+      client.people.find_employee_by_participant_id(veteran_participant_id)
+    end
+  end
+
+  def sensitivity_level
+    @sensitivity_level ||= MetricsService.record("BGS: fetch sensitivity level by participant id: #{veteran_participant_id}",
+                                                 service: :bgs,
+                                                 name: "security.find_sensitivity_level_by_participant_id") do
+      client.security.find_sensitivity_level_by_participant_id(veteran_participant_id)
+    end
+  end
+
+  def client
+    @client ||= ExternalApi::BGSService.new.client
+  end
+end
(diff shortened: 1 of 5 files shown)


app/services/external_api/bgs_veteran_station_user_conflict.rb
```diff
@@ -0,0 +1,97 @@
...
+  # a good example in production is veteran_participant_id 3309696 and user id 2321
+  # "true" return value in this case means there is a conflict.
+  # "false" means no conflict.
+  def call
+    DBService.release_db_connections
+
+    return false unless station_dtos.any?
+
+    # simple case if DTOs exist and contain a Veteran record
+    return true if veteran_dto
```
leikkisa: Should this be veteran_at_same_station?

app/services/external_api/bgs_veteran_station_user_conflict.rb
```diff
@@ -0,0 +1,97 @@
...
+    # default is no conflict
+    false
+  end
+
+  private
+
+  attr_reader :veteran_participant_id
+
+  def current_user_station
+    RequestStore[:current_user].station_id.to_s
```
leikkisa: I think this is okay (do we reset the current user to the establishment user when this is being done by a job?).  If not, should we pass the user?

app/services/external_api/bgs_veteran_station_user_conflict.rb
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  def station_dtos
+    [employee_dtos[:station]].flatten
```
leikkisa: Not necessarily a change request, but I don't understand what a DTO is.  Maybe we can add it to WTF.

app/services/external_api/bgs_veteran_station_user_conflict.rb
```diff
@@ -0,0 +1,97 @@
...
+  def call
+    DBService.release_db_connections
+
+    return false unless station_dtos.any?
+
+    # simple case if DTOs exist and contain a Veteran record
+    return true if veteran_dto
+
+    # likewise for spouse
+    return true if spouse_dto
```
leikkisa: This is probably because I don't fully understand DTO, but should we be checking if the spouse is at the same station?

app/services/external_api/bgs_veteran_station_user_conflict.rb
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  # logic detailed in https://github.com/department-of-veterans-affairs/caseflow/issues/10087#issuecomment-507783830
+  # a good example in production is veteran_participant_id 3309696 and user id 2321
+  # "true" return value in this case means there is a conflict.
+  # "false" means no conflict.
+  def call
+    DBService.release_db_connections
+
+    return false unless station_dtos.any?
```
leikkisa: Do we still want to check sensitivity if there aren't station_dtos?

app/services/external_api/bgs_veteran_station_user_conflict.rb
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  def sensitivity_reason
+    sensitivity_level[:sntvty_reason_type_nm]
+  end
+
+  def violates_sensitivity_reason?
+    return false unless sensitivity_level
+
+    ["Relative of Local VA Employee", "VBA Employee", "Veteran", "Work Study"].include?(sensitivity_reason)
```
leikkisa: I'm guessing if there are other sensitivity access problems, it would happen in the can_access? check, so I think this is okay.

app/services/external_api/bgs_veteran_station_user_conflict.rb
```diff
@@ -0,0 +1,97 @@
...
+  def initialize(veteran_participant_id:, client: nil)
+    @veteran_participant_id = veteran_participant_id
+    @client = client
+  end
+
+  # logic detailed in https://github.com/department-of-veterans-affairs/caseflow/issues/10087#issuecomment-507783830
+  # a good example in production is veteran_participant_id 3309696 and user id 2321
+  # "true" return value in this case means there is a conflict.
+  # "false" means no conflict.
+  def call
```
leikkisa: Would it be bad form to call this conflict?  I am okay with this as is, but I think `!conflict_checker.call` is not the most intuitive to read.

spec/services/external_api/bgs_service_spec.rb
```diff
@@ -45,4 +47,84 @@
...
+          ptcpnt_id: veteran.participant_id,
+          station: { ptcpnt_id: "456", ptcpnt_rlnshp_type_nm: "VBA employee", station_number: user.station_id }
+        }
+      end
+      let(:sensitivity_level) do
+        {
+          cd: user.station_id[1..2],
+          fclty_type_cd: user.station_id[0],
+          scrty_level_type_cd: "6",
+          sntvty_reason_type_nm: "VBA Employee"
```
leikkisa: The veteran can still be modified if they're an employee, as long as they're not being processed at the same office right?

pkarman: correct. the sensitivity checks only relate to the same station as the user.

app/services/external_api/bgs_veteran_station_user_conflict.rb
pkarman: this class is called synchronously during intake, not in a job. https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/validators/intake_start_validator.rb#L28

We could pass the user in explicitly, sure. All other BGS service calls use `RequestStore[:current_user]` though so we'd have to pass it to the client too.

app/services/external_api/bgs_veteran_station_user_conflict.rb
pkarman: Data Type Object -- I asked too! :)

app/services/external_api/bgs_veteran_station_user_conflict.rb
pkarman: yes. This is one reason we see authz failures. The "Spouse" is not called out explicitly in https://github.com/department-of-veterans-affairs/caseflow/issues/10087#issuecomment-507783830 but it is one of the reasons that sensitivity checks fail.

pkarman: it falls under "If the Relationship Type is not “Veteran” the station should be restricted"

app/services/external_api/bgs_veteran_station_user_conflict.rb
pkarman: No. Because we only check sensitivity if there certain types of DTOs.

pkarman:
thanks @leikkisa -- renamed some methods for clarity (I agree with you -- I was adhering too closely to the BGS language).

app/services/external_api/bgs_veteran_station_user_conflict.rb
leikkisa: oh! If this is just happening during intake, then I wouldn't worry about passing the user. I was confusing checking at the beginning (which is what this is) versus establishing later.

app/services/external_api/bgs_veteran_station_user_conflict.rb
leikkisa: I guess if the person is the veteran's spouse they should be prevented from doing the intake no matter what the station. And then if the spouse is located at the same station as the user doing the intake, that should also be prevented?

leikkisa:
I'm sure this is somewhere else, but I'm adding it for my own benefit, so that I know how to read what's happening:

*findEmployeeByFileNumber (or ParticipantId)*

This will return an EmployeeDTO with a List of EmployeeStnDTOs. Just because something is returned in this list does not necessarily mean it must be restricted.

If the list of EmployeeStnDTO contains an EmployeeStnDTO with relationship type “Veteran” then the veteran’s Sensitivity level must be evaluated by making the next service call. *No need to make the second call if there is no DTO with type “Veteran”

*findSntvtyLevelByFileNumber (or ParticipantId)*

This one is a little tricky. You do not want the bdnStnNbr for this, you want the fcltyTypeCd + cd. You may see something like this
bdnStnNbr=349
cd=17
fcltyTypeCd=3
So in this case the potentially station is actually 317, not 349.
 
When you’ve gathered your list of EmployeeStnDTOs to evaluate, for each dto, you will need to check the following to determine if the station should be restricted: 
 
1.       If the Relationship Type is not “Veteran” the station should be restricted
2.  OR If the Sensitivity Reason is in the following list: 
Relative of Local VA Employee
VBA Employee
Veteran 
Work Study
**The following 2 reasons are not reasons to restrict access to a station:
Non-VBA Employee
Veteran Service 
3.  OR if the current user’s subjectId matches the participant Id of the DTO

app/services/external_api/bgs_veteran_station_user_conflict.rb
pkarman: Agreed that station shouldn't matter. I *think* the DTOs will include the Spouse entry regardless of station, but we would need an example in production to verify. The example I have from https://github.com/department-of-veterans-affairs/caseflow/issues/11736#issuecomment-520588469 is where the user is the Spouse at the same station as the Veteran.

PR #11750, opened 2019-08-12
Move logic for sorting tasks out of TaskPager

lowellrex:
Connects #11733. `TaskPager` no longer controls sorting of tasks. That responsibility is transferred to the new `TaskSorter` class and the various (also new!) `QueueColumn` subclasses.


app/models/task_pager.rb
```diff
@@ -26,66 +26,9 @@ def paged_tasks
...
-  def tasks_sorted_by_veteran_name(tasks)
-    tasks.joins(cached_attributes_join_clause).order("cached_appeal_attributes.veteran_name #{sort_order}")
-  end
-
-  def cached_attributes_join_clause
-    "left join cached_appeal_attributes "\
-    "on cached_appeal_attributes.appeal_id = tasks.appeal_id "\
-    "and cached_appeal_attributes.appeal_type = tasks.appeal_type"
+    column = QueueColumn.from_name(sort_by)
+    TaskSorter.new(tasks: tasks, sort_order: sort_order, column: column).sorted_tasks
```
lowellrex: This is the heart of the change. All of this logic moved to the new `TaskSorter` class.

app/models/queue_column.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+class QueueColumn
+  include ActiveModel::Model
+
+  class << self
+    def from_name(column_name)
+      column = subclasses.find { |subclass| subclass.column_name == column_name }
```
lowellrex: `QueueColumn` subclasses currently have only properties, no behaviour. Thus they are candidates for factoring into a `Constants` json file. I am on the fence about whether they should be classes on configurations in a json file, and am curious for your thoughts on the matter.

Right off the bat some arguments for each that I find compelling:

### Classes
* Subclasses can inherit from the parent class
* Can use class constants to refer to tables associated with Active Record objects like `CachedAppeal` and `Task`
* Makes it easier to add behaviour to `QueueColumn`s in the future should we want to
* `simplecov` knows to yell at us if we do not have sufficient code coverage

### Constants json file
* Inheritance is indirect and determining why things are behaving as they are can be confusing when reading code
* Defining all columns in a single file makes it easier to read
* All configurations can be shared with the front-end

Do you have any compelling arguments in favor of one approach or the other? Do you favor one over the other?

kevmo: I find the JSON file to be compelling for single-source-of-truth reasons. 

> It is better to have 100 functions operate on one data structure than to have 10 functions operate on 10 data structures.

app/models/task_pager.rb
```diff
@@ -94,7 +37,9 @@ def task_page_count
 
   def filtered_tasks
     where_clause = QueueWhereClauseArgumentsFactory.new(filter_params: filters).arguments
-    where_clause.empty? ? tasks_for_tab : tasks_for_tab.joins(cached_attributes_join_clause).where(*where_clause)
+    return tasks_for_tab if where_clause.empty?
+
+    tasks_for_tab.joins(TaskSorter.cached_attributes_join_clause).where(*where_clause)
```
lowellrex: When we move the filtering logic out of `TaskPager` I think a more natural location for the `cached_attributes_join_clause` method will appear and we can move it then.

client/constants/QUEUE_CONFIG.json
```diff
@@ -7,20 +7,92 @@
 
   "APPEAL_TYPE_COLUMN": "typeColumn",
   "CASE_DETAILS_LINK_COLUMN": "detailsColumn",
-  "DAYS_WAITING_COLUMN": "daysWaitingColumn",
   "DAYS_ON_HOLD_COLUMN": "daysOnHoldColumn",
-  "DOCUMENT_COUNT_READER_LINK_COLUMN": "readerLinkColumn",
+  "DAYS_WAITING_COLUMN": "daysWaitingColumn",
   "DOCKET_NUMBER_COLUMN": "docketNumberColumn",
+  "DOCUMENT_COUNT_READER_LINK_COLUMN": "readerLinkColumn",
```
lowellrex: Alphabetization.

client/constants/QUEUE_CONFIG.json
```diff
@@ -7,20 +7,92 @@
...
   "HEARING_BADGE_COLUMN": "hearingBadgeColumn",
   "ISSUE_COUNT_COLUMN": "issueCountColumn",
   "REGIONAL_OFFICE_COLUMN": "regionalOfficeColumn",
   "TASK_ASSIGNEE_COLUMN": "assignedToColumn",
   "TASK_ASSIGNER_COLUMN": "completedToNameColumn",
   "TASK_CLOSED_DATE_COLUMN": "completedDateColumn",
+  "TASK_CREATED_AT_COLUMN": "createdDateColumn",
```
lowellrex: Add new column for default sorting.

client/constants/QUEUE_CONFIG.json
```diff
@@ -7,20 +7,92 @@
...
   "TASK_DUE_DATE_COLUMN": "caseDueDateColumn",
   "TASK_HOLD_LENGTH_COLUMN": "caseDaysOnHoldColumn",
   "TASK_TYPE_COLUMN": "taskColumn",
 
+  "COLUMNS": {
```
lowellrex: After this is merged I will create a PR to migrate all uses of `QUEUE_CONFIG.*_COLUMN` to instead use `QUEUE_CONFIG.COLUMNS.*`.

hschallhorn: Done in #11928

Issue #9590, opened 2019-02-21
Clean up feature toggles


aroltsch:
## AC:
Go through feature toggles that are enabled for all in appeals-deployment in production env and delete those in Caseflow and from appeals-deployment.

#### Done previously (thanks @Sjones352 !)
- [x] create_board_grant_effectuations
- [x] decision_reviews
- [x] judge_case_review_checkout
- [x] ama_decision_issues
- [x] case_search_home_page
- [x] intake_reserved_file_number
- [x] vbms_include_user - set to `true`
- [x] claims_service_v5 - set to `true`

#### Remove from Caseflow code and appeals deployment. 
- [x] attorney_assignment_to_colocated - remove `featureToggles` from view

#### These only need to be removed from appeals deployment:
- [x] decision_document_upload
- [x] ama_acd_tasks
- [x] idt_ama_appeals
- [x] intake_legacy_opt_in
- [x] automatic_case_distribution

## Technical notes
A minimum of 2 PRs is required (more is cool too).
Feature toggles should not be removed from appeals_deployment until the PR removing their code from Caseflow has been merged for one day.

Removing feature toggles in appeals deployment is done here:
`appeals-deployment/ansible/vars/features-config.yml`

This is not necessary for this PR, but for reference, here are instructions for checking if a feature toggle has been turned on in production for over a month:
e.g. `git blame ansible//vars/features-config.yml | grep "decision_reviews"`

jimruggiero:
Estimated at a 3 in lieu of additional context regarding Anya's suggestion to do this in multiple PR's and not know enough about what's involved with removing these from appeals deployments.

jimruggiero:
@Sjones352 Can you provide any additional context to finalize estimation and next steps?

Sjones352:
@jimruggiero this may take a little while because they are multiple feature toggles and tests may need to change. I will look into it further on Monday

PR #11771, opened 2019-08-14
Remove case_search_home_page feature toggle

Sjones352:
Connects #9590

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -168,15 +168,12 @@ class ApplicationController < ApplicationBaseController
   end
 
   def case_search_home_page
-    if current_user&.authenticated?
-      return false if current_user.admin?
-      return false if current_user.organization_queue_user? || current_user.vso_employee?
-      return false if current_user.attorney_in_vacols? || current_user.judge_in_vacols?
-      return false if current_user.colocated_in_vacols?
+    return false if current_user.admin?
+    return false if current_user.organization_queue_user? || current_user.vso_employee?
+    return false if current_user.attorney_in_vacols? || current_user.judge_in_vacols?
+    return false if current_user.colocated_in_vacols?
 
-      return true
-    end
-    false
+    true
   end
   helper_method :case_search_home_page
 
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -168,7 +168,7 @@ def no_cache
   end
 
   def case_search_home_page
-    if feature_enabled?(:case_search_home_page)
+    if current_user&.authenticated?
```
Sjones352: By removing the feature toggle, I implemented a method to check if user is authenticated. I am guessing by my test failing, it should not check for user authentication.

leikkisa: Hey Sandra, yeah you should be able to take that out, it's already in verify_authentication. So you could remove the `if` altogether.  It feels like this shouldn't have caused your test to fail though, unless the user wasn't authenticated there, but still a good thing to catch!

app/controllers/home_controller.rb
```diff
@@ -4,9 +4,10 @@ class HomeController < ApplicationController
...
   def index
-    return redirect_to("/help") unless current_user&.authenticated?
-    return render("queue/index") if feature_enabled?(:case_search_home_page)
-
-    redirect_to("/help")
+    if current_user&.authenticated?
+      render("queue/index")
+    else
+      redirect_to("/help")
+    end
```
leikkisa: It'll look more like this:
```
return redirect_to("/help") unless current_user&.authenticated?
render("queue/index")
```

app/controllers/application_controller.rb
leikkisa:
```
return false if current_user.admin?
return false if current_user.organization_queue_user? || current_user.vso_employee?
return false if current_user.attorney_in_vacols? || current_user.judge_in_vacols?
return false if current_user.colocated_in_vacols?

true
```

app/controllers/home_controller.rb
Sjones352: Fixed

Sjones352: Thanks for the suggestions Sally. Right now all tests are passing

Issue #11509, opened 2019-07-24
Include filterable values in queue config response to front-end


lowellrex:
TODO: Finish writing this ticket.

## Acceptance criteria
* [ ] Back-end returns the values that are available for a field to filter on to the front-end when the queue table view is drawn

lowellrex:
Similar to work done on bulk task assign modal (TODO: Add link). Numbers in parentheses should be total counts in database, not just tasks loaded on the front-end.

```
1 | ||
2 | ||||
3 | |||||
5 | |
8 | 
```

Estimating at 3. Could be 1 because we are simply grouping by a single column in a single table most of the time. Testing could add a significant amount of effort here. The fact that we have to do this for each column means it is probably not totally trivial to do. 5 as a hedge in case there are performance issues that arise that we need to address.

PR #11777, opened 2019-08-14
Enable TaskTable to use API requests to filter tasks

### LINKED ISSUE #11309: Enable TaskTable to use API requests to filter tasks
Spun out of #11054.

The `TaskTable` component currently manages filtering only on the front-end. This ticket exists to enable us to pass filters from the front-end as part of the pagination API requests we make so that we can filter large queues. At some point we will migrate all queues to using the pagination API but are not doing that migration in a single step so we need to support both methods at the same time.

## 🚫 Blocked by
* [x] Allow organization queue table configuration to accept filtering (#11307)
* [x] Include filterable values in queue config response to front-end (#11509)
* [x] Enable remaining three task queue filters on back-end (#11888)

## Acceptance criteria
* [ ] `TaskTable` component's pagination, filtering, and sorting use API requests instead of happening on the front-end.
* [x] Use options returned by the back-end (#11509) to build the filter dropdown menus

## Non-goals
* In order to reduce the complexity of this ticket, let's not save any tasks in state initially. Caching will happen as part of #11310

hschallhorn:
Resolves #11309

### Description
Builds query string to request filtered tasks from the back end and displays the filtered tasks.

### Acceptance Criteria
- [ ] TaskTable component's pagination, filtering, and sorting use API requests instead of happening on the front-end.

### Testing Plan
1. Ensure pagination is enabled with `FeatureToggle.enable!(:use_task_pages_api)`
2. Change user to any VLJ support member and navigate to /organizations/vlj-support
3. Click the filter button next to any column and select a filter!
4. Ensure the tasks filter correctly!!!
5. Pay special attention to filtering by AOD status as well


app/controllers/organizations/task_pages_controller.rb
```diff
@@ -40,6 +32,17 @@ def organization_url
     params[:organization_url]
   end
 
+  def paged_tasks
```
hschallhorn: Moved out because of code climate

client/app/queue/QueueTable.jsx
```diff
@@ -313,7 +314,11 @@ export default class QueueTable extends React.PureComponent {
   // &page=2
   // &sort_by=detailsColumn
   // &order=desc
+  // &filter[]=col=docketNumberColumn&val=legacy,evidence_submission&filters[]=col=taskColumn&val=Unaccredited rep
```
hschallhorn: found by
```ruby
> data = { :filters => ["col=docketNumberColumn&val=legacy,evidence_submission", "col=taskColumn&val=Unaccredited rep"] }
> CGI.unescape(data.to_query)
=> "filters[]=col=docketNumberColumn&val=legacy,evidence_submission&filters[]=col=taskColumn&val=Unaccredited rep"
```

client/app/queue/QueueTable.jsx
```diff
@@ -325,9 +330,24 @@ export default class QueueTable extends React.PureComponent {
         QUEUE_CONFIG.COLUMN_SORT_ORDER_DESC;
     }
 
+    if (!_.isEmpty(filteredByList)) {
+      for (const columnName in filteredByList) {
+        if (!_.isEmpty(filteredByList[columnName])) {
+          const column = this.props.columns.find((col) => col.columnName === columnName);
```
hschallhorn: `columnName` here is the property of the task being used for the value. For example, in the case of `taskColumn`, `columnName` is 'label', not 'taskColumn' which is what the back end expects.

hschallhorn:
Currently blocked by #11509

hschallhorn:
This is ready to rock after #12031 is merged and AOD filters are tested

PR #11780, opened 2019-08-14
Sorting by Assignee Name

### LINKED ISSUE #11316: Enable back-end sorting of tasks by assignee name
#11247 established the pattern for how the pagination API will sort tasks for the queue on the back-end but did not enable sorting by all fields. This ticket exists to enable sorting by assignee name through the pagination API that mirrors [how we sort by that field on the front-end](https://github.com/department-of-veterans-affairs/caseflow/blob/0377b8d0a0d77a81a070c82c6bd336600885cfa3/client/app/queue/components/TaskTable.jsx#L180).

## Acceptance criteria
* [ ] Enable back-end sorting of tasks by assignee name

lomky:
Resolves #11316

Adds sorting based on task assignee name.


app/models/task_pager.rb
```diff
@@ -46,23 +47,30 @@ def sorted_tasks(tasks)
...
     end
   end
+  # rubocop:enable Metrics/MethodLength
   # rubocop:enable Metrics/AbcSize
   # rubocop:enable Metrics/CyclomaticComplexity
 
+  def tasks_sorted_by_assignee_name(tasks)
+    tasks.joins("left join users on tasks.assigned_to_id = users.id and tasks.assigned_to_type = '#{User.name}'")
+      .order("users.full_name #{sort_order}")
+  end
```
hschallhorn: I don't think this logic is quite right.
1. We will be passing tasks assigned to organizations to the front end and will need to sort on that as well.
2. The front end actually sorts on [appeal.assigned_to_location](https://github.com/department-of-veterans-affairs/caseflow/blob/a47e65c1cf97a722b24d075c02d66416f8cb7bec/app/models/serializers/work_queue/task_serializer.rb#L38) which [Appeal](https://github.com/department-of-veterans-affairs/caseflow/blob/a47e65c1cf97a722b24d075c02d66416f8cb7bec/app/models/appeal.rb#L83) and [LegacyAppeal](https://github.com/department-of-veterans-affairs/caseflow/blob/a47e65c1cf97a722b24d075c02d66416f8cb7bec/app/models/legacy_appeal.rb#L707) both have their own logic for.
3. RootTask, LegacyTask and Task each have their [own implementation](https://github.com/department-of-veterans-affairs/caseflow/search?q=assigned_to_label&unscoped_q=assigned_to_label) for assigned_to_label as well.

I think we'll need some context as to why the front end logic for assignee name is such a mess, but I'm definitely for simplifying it if we can. One option could be just changing how it works in [`task_column_serializer`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/serializers/work_queue/task_column_serializer.rb#L144) as pagination is the only place we use that.

PR #11781, opened 2019-08-14
When canceling an AssignHearingDispositionTask, only try to create an EvidenceSubmissionWindowTask if it doesn't already exist

### LINKED ISSUE #10554: Caseflow::Error::DuplicateOrgTask on duplicate ChangeHearingDispositionTasks
[Sentry issue](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/4673/events/)

The HearingDispositionChangeJob [is pulling in `ChangeHearingDispositionTask`s as well as `DispositionTask`s when collecting tasks to act on](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/jobs/hearing_disposition_change_job.rb#L18); it should only pull in `DispositionTask`s.

It's generating errors but leaving tasks in the state they should be in.

tomas-nava:
Resolves #10554

### Description
When canceling an `AssignHearingDispositionTask`, uses `find_or_create_by!` to only create a new `EvidenceSubmissionWindowTask` if one doesn't already exist on the task's `HearingTask`'s parent. This should resolve [this Sentry error](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/5946/events/420785/).


spec/models/tasks/assign_hearing_disposition_task_spec.rb
```diff
@@ -290,6 +289,20 @@
...
+            let!(:evidence_submission_window_task) do
+              create(:evidence_submission_window_task, appeal: appeal, parent: distribution_task)
+            end
+
+            it "does not attempt to create a new EvidenceSubmissionWindowTask" do
+              expect(EvidenceSubmissionWindowTask.count).to eq 1
+
+              expect { subject }.to_not raise_error
+
+              expect(EvidenceSubmissionWindowTask.count).to eq 1
```
monfresh: What do you think about using RSpec's change expectation, which I think would be the equivalent of these 3 expectations? 
```ruby
expect { subject }.to_not change(EvidenceSubmissionWindowTask, :count)
```
I might be off on the syntax.

spec/factories/task.rb
```diff
@@ -387,7 +387,7 @@
     factory :evidence_submission_window_task, class: EvidenceSubmissionWindowTask do
       type { EvidenceSubmissionWindowTask.name }
       appeal { create(:appeal) }
-      assigned_to { HearingsManagement.singleton }
+      assigned_to { MailTeam.singleton }
```
monfresh: This is fine for this PR since this is extra work not related to this PR, but we should open a new issue to create all the singletons at the start of the test suite, and exclude them from being cleaned up by database cleaner.

monfresh: I created the issue: https://github.com/department-of-veterans-affairs/caseflow/issues/11782

spec/models/tasks/assign_hearing_disposition_task_spec.rb
tomas-nava: you were not off on the syntax! fixed in [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/11781/commits/78cc472548b278d5e105e5a17be924e275444e24).

PR #11783, opened 2019-08-15
Removed intake-reserved-file-number-flag

Sjones352:
Connects #9590


spec/feature/intake/intake_spec.rb
```diff
@@ -321,14 +321,6 @@
...
 
-      before do
-        FeatureToggle.enable!(:intake_reserved_file_number, users: [current_user.css_id])
-      end
-
-      after do
-        FeatureToggle.disable!(:intake_reserved_file_number, users: [current_user.css_id])
-      end
-
       let(:veteran) do
```
Sjones352: I guess I should remove these tests since we are using `Rails.deploy_env?(:prod)` to check for production

leikkisa: Hi Sandra. I would keep the tests, but mock Rails.deploy_env?(:prod) to return true for the tests.

PR #11786, opened 2019-08-15
Fix bug for no upcoming hearing days for RO17 and RO18

### LINKED ISSUE #11764: Uncaught TypeError: Cannot read property 'room' of undefined for RO18 in test
In AssignHearings.jsx, selectedHearingDay.room is undefined so the following error: `Uncaught TypeError: Cannot read property 'room' of undefined` is thrown. Link: http://localhost:3000/hearings/schedule/assign?roValue=RO18

ferristseng:
Resolves #11764

### Description

  - Only return selected hearing day room if selected hearing day is defined
  - Test to make sure the above fix is working


spec/feature/hearings/schedule_veteran_spec.rb
```diff
@@ -509,4 +509,12 @@ def navigate_to_schedule_veteran_modal
       expect(page).to have_content(COPY::SCHEDULE_VETERAN_SUCCESS_MESSAGE_DETAIL)
     end
   end
+
+  context "No upcoming hearing days" do
```
rubaiyat22: I'm not sure if this test is necessary. Please let me know if this needs to be kept or discarded.

sharonwarner: I think it makes sense to keep!

client/app/hearings/components/AssignHearings.jsx
```diff
@@ -58,14 +58,25 @@ const UpcomingHearingDaysNav = ({
   </div>
 );
 
+UpcomingHearingDaysNav.propTypes = {
```
sharonwarner: 👍

PR #11788, opened 2019-08-15
Ensure case search returns legacy appeals for vso employees

### LINKED ISSUE #11650: Allow VSO Caseflow users to search for Legacy cases
### Use Case
As a VSO user of Caseflow, I can see legacy cases in the daily docket.  I should be able to search for all cases that I see in my docket.

### Acceptance Criteria
- [ ] The search functionality across Caseflow should return results from Legacy cases, in addition to the AMA cases that it currently returns.

hschallhorn:
Resolves #11650

### Description
Adds legacy appeals to the appeals returned from `veteran.accessible_appeals_for_poa`

|Before|After|
|---|---|
|Screen Shot 2019-08-15 at 2 37 48 PM|Screen Shot 2019-08-15 at 2 38 13 PM|

### Acceptance Criteria
- [ ] The search functionality for VSO users should return results from `LegacyAppeals`, in addition to the `Appeals` that it currently returns.

### Testing Plan
1. Log is as our good friend BILLIE_VSO
2. Find a veteran Billie would not normally have access to
```ruby
> u = User.find_by(css_id: "BILLIE_VSO")
> v = Veteran.find(46)
> CaseSearchResultsForVeteranFileNumber.new(file_number_or_ssn: v.file_number, user: u).call.extra[:search_results][:appeals].any?
=> false
```
3. Pretend Billie's VSO is the veteran's poa by changing [this participant_id](https://github.com/department-of-veterans-affairs/caseflow/blob/a47e65c1cf97a722b24d075c02d66416f8cb7bec/lib/fakes/bgs_service.rb#L554) to Billie's VSO's participant_id. (rails console and server may need to be restarted after this step)
```ruby
> u.organizations.first.participant_id
=> "2452415"
```
4. Ensure this appeal is now returned in the search
```ruby
CaseSearchResultsForVeteranFileNumber.new(file_number_or_ssn: v.file_number, user: u).call.extra[:search_results][:appeals].any?
=> true
```
5. Visit /search and ensure this case shows up when searching for the veteran's file number
```ruby
> v.file_number
=> "228081153"
```


app/models/veteran.rb
```diff
@@ -180,6 +187,18 @@ def accessible_appeals_for_poa(poa_participant_ids)
     end
   end
 
+  def accessible_legacy_appeals_for_poa(poa_participant_ids)
+    legacy_appeals = LegacyAppeal.fetch_appeals_by_file_number(file_number)
+
+    claimants_participant_ids = legacy_appeals.map { |appeal| appeal.veteran.participant_id }.flatten
+
+    poas = bgs.fetch_poas_by_participant_ids(claimants_participant_ids)
```
lowellrex: Do we ever call `accessible_appeals_for_poa` some place other than in the search results where we don't want to return `LegacyAppeal`s? If not, then could refactor this slightly to make one fewer call to BGS by collecting `claimants_participant_ids` for both AMA and legacy appeals before making this request to BGS.

hschallhorn: oo YES. nice catch.

Issue #11739, opened 2019-08-12
Motion to vacate dismissals and denials | Connect frontend and backend for judges to log motions to vacate decisions


laurjpeterson:
When judges deny or dismiss a Motion to Vacate, the Motions Attorney receives the case back to outcode and dispatch. We'll do this by logging the judge Motion to Vacate disposition (grant/den/dismiss) and sending the task back to the assigning Motions Attorney.

## AC
- Connect frontend and backend 
   - Front end ticket: #10325
   - Back end ticket: #11707 
- When judges dismiss or denies a motion to vacate, judge motion to vacate (MTV) task is completed and the motions attorney's previous task is put in-progress 

## Out of scope
- Granted motions to vacate - covered in #11253

aroltsch:
There is a WIP in progress PR that should be picked up by another engineer.

aroltsch:
TODO items:

Create attorney tasks if motion is granted
Return task to the motions attorney if motion is not granted

PR #11800, opened 2019-08-16
Allow adding dispositions to post-decision motions

monfresh:
Connects #11739

TODO: 
1. Accept assigned_to in the param
2. Create attorney tasks if motion is granted
3. Return task to the motions attorney if motion is not granted


aroltsch:
The endpoint should accept new assignee (if disposition is granted), the notes plus the disposition and the vacate type . Task ID would be the current judge's task id. Then based on the disposition, we would either create a new task to send it to the drafting attorney or send the appeal back to the motions attorney who originally sent the task to the judge.

leikkisa:
Still to do (in future PR):
Create attorney tasks if motion is granted
Return task to the motions attorney if motion is not granted

Issue #11713, opened 2019-08-09
Automatic Case Distribution failures


monfresh:
We noticed that some Distributions get stuck in the `pending` state because an error occurred along the way. The state should be changed to `error`, but for some reason, it is not. If a Distribution is `pending`, then it blocks judges from trying again because we have a validation that won't allow a Distribution to be created if there is already an existing one for that same judge in the `pending` state. While tracing the error for some of those `pending` Distributions, I noticed that for 2 of them, the error was the same:
How I searched the logs:
```
awslogs get dsva-appeals-certification-prod/opt/caseflow-certification/src/log/caseflow-certification.out ALL --start='8/7/2019 11:17:30' --end='8/7/2019 11:18:00' | grep "StartDistributionJob"
```
Error:
```
StartDistributionJob failed: PG::NotNullViolation: ERROR:  null value in column "appeal_id" violates not-null constraint
```
It looks like this happened when trying to create a JudgeAssignTask without an `appeal_id`:
https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/docket.rb#L72

I haven't yet figured out how that happened.

In addition, I looked up how many Distributions we have in `error` state, and there were 340. I looked up the last one and this was the error:
```
StartDistributionJob failed: PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_distributed_cases_on_case_id"
```

Slack thread:
https://dsva.slack.com/archives/CHX8FMP28/p1565121226268200

pkarman:
All the pending Distribution records were deleted when this ticket was opened.

pkarman:
`duplicate key value violates unique constraint "index_distributed_cases_on_case_id"` is happening frequently and is probably at the root of many failures.

pkarman:
Now we are seeing errors in Sentry: https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/6114/events/433162/

pkarman:
The problem is consistently with the SQL in https://github.com/department-of-veterans-affairs/caseflow/blame/master/app/models/vacols/case_docket.rb#L238 where VACOLS cases are being distributed more than once.

My suspicion is that some data has changed in VACOLS that is triggering whatever logic Caseflow uses to identify them. This may be a business flow bug, or a problem in the SQL, and it will require more investigation.

pkarman:
Comments from @amprokop 

[...]

Issue #11778, opened 2019-08-14
Standardize class structure


pkarman:
Per https://github.com/rubocop-hq/ruby-style-guide#consistent-classes and https://github.com/rubocop-hq/ruby-style-guide#def-self-class-methods we should standardize our class layouts so that we consistently order methods.

We should always prefer this style:

```ruby
class Foo
  class << self
    def first_method
      # body omitted
    end

    def second_method_etc
      # body omitted
    end
  end

  def initialize
    # something
  end
end
```

and should group all class methods first in the file, followed by instance methods.

monfresh:
There's also a new more aesthetically-pleasing website for the style guide: https://rubystyle.guide/#classes-modules

PR #11806, opened 2019-08-16
Add started_at and errored_at columns to Distribution

pkarman:
connects #11713 

A bunch of misc clean up in prep for understanding how Distributions work.

Moves some files, refactors a large method into into its own class, fixes a small js bug.

Adds new columns for the `Distribution` model so we can get a better sense of job timings.

*Schema Changes Only*

* [x] Column comments updated


client/app/components/Pagination.jsx
```diff
@@ -50,11 +50,11 @@ class Pagination extends React.PureComponent {
 
     // If there are no pages, there is no data, so the range should be 0-0.
     // Otherwise, the beginning of the range is the previous amount of cases + 1
-    const beginningCaseNumber = (currentPage * pageSize) - pageSize + 1;
+    const beginningCaseNumber = totalCases === 0 ? 0 : ((currentPage * pageSize) - pageSize + 1);
```
pkarman: this fixes a pagination bug where it was displaying `1-0 of 0 total`

app/controllers/distributions_controller.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class DistributionsController < ApplicationController
+  include RunAsyncable
```
pkarman: just uses the concern -- no functional change

app/jobs/start_distribution_job.rb
```diff
@@ -17,6 +17,6 @@ def perform(distribution, user = nil)
   def handle_error(error)
     Rails.logger.info "StartDistributionJob failed: #{error.message}"
     Rails.logger.info error.backtrace.join("\n")
-    Raven.capture_exception(error)
```
pkarman: new job style

app/models/distributed_case.rb
```diff
@@ -4,7 +4,7 @@ class DistributedCase < ApplicationRecord
   belongs_to :distribution
   belongs_to :task
 
-  validates :distribution, :case_id, :docket, :ready_at, presence: true
+  validates :case_id, :distribution, :docket, :ready_at, presence: true
```
pkarman: just some alphabetizing

app/models/distribution.rb
```diff
@@ -33,9 +33,14 @@ def distribute!
       return unless valid?(context: :create)
     end
 
+    # always set the timestamp, ok if the status rolls back in the transaction below.
+    update!(started_at: Time.zone.now)
```
pkarman: using the new columns

app/models/vacols/case.rb
```diff
@@ -253,18 +253,71 @@ class InvalidLocationError < StandardError; end
...
   }
 
-  # These scopes query VACOLS and cannot be covered by automated tests.
-  # :nocov:
-  def self.remands_ready_for_claims_establishment
-    VACOLS::Case.joins(:folder, :correspondent)
-      .where(WHERE_PAPERLESS_REMAND_LOC97)
-      .order("BFDDEC ASC")
-  end
+  class << self
```
pkarman: reorganized class methods per #11778 and refactored `batch_update_vacols_location` into its own class.

app/queries/vacols_location_batch_updater.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+class VacolsLocationBatchUpdater
```
pkarman: refactor from VACOLS::Case

spec/models/distribution_spec.rb
```diff
@@ -189,12 +192,37 @@
       expect(subject.distributed_cases.where(docket: Constants.AMA_DOCKETS.evidence_submission).count).to eq(2)
     end
 
+    context "when a nonpriority legacy case re-distribtution is attempted" do
```
pkarman: this is the meaningful test. it validates current behavior (which we might change later).

app/models/vacols/case.rb
lomky: Other than this, was the functionality changed at all?

pkarman: nope, same functionality. just a refactor to make a new Query class.

PR #11807, opened 2019-08-16
Remove using JSON.parse in favor of response body instead

### LINKED ISSUE #11797: Fix Double Parsing of JSON on Frontend
In many places where we are using `ApiUtil`, we are parsing the response text as JSON:

```
const responseObject = JSON.parse(response.text);
```

The `superagent` library we're using mentions in the documentation that for common response types like JSON, the body is automatically parsed

> Much like SuperAgent can auto-serialize request data, it can also automatically parse it. When a parser is defined for the Content-Type, it is parsed, which by default includes "application/json" and "application/x-www-form-urlencoded". The parsed object is then available via res.body.

We should be able to skip parsing, and just use `response.body` instead.

ferristseng:
Resolves #11797

### Description

- Remove the usage of `JSON.parse(_)` in favor of using `response.body` directly
- Move some functions that are duplicated to `intake.js`

### DIFF
--- a/client/app/certification/Certification.jsx
+++ b/client/app/certification/Certification.jsx
@@ -22,20 +22,18 @@ export class Certification extends React.Component {
     const loadPromise = new Promise((resolve, reject) => {
       const makePollAttempt = () => {
         ApiUtil.get(`/certifications/${this.props.vacolsId}`).
-          then(({ text }) => {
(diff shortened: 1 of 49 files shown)


ferristseng:
Merging #11860, which contains this PR, and fixes for the failed lint step.

Issue #11631, opened 2019-08-01
Write tech spec describing how to mark Caseflow users as inactive


lowellrex:
Rows in Caseflow's `users` table should never be deleted* from the database. This allows us to keep a record of users as long as the application lives. However, this practice of persisting `User` records in perpetuity presents some challenges with tasks. Two salient challenges are 1) people who are no longer working on Caseflow (perhaps because they are no longer employed at the VA or at a VSO) can still be assigned tasks and 2) we have no insight into how many active tasks are assigned to people who will never act on those tasks.

These challenges were identified in a [related Slack conversation](https://dsva.slack.com/archives/CHX8FMP28/p1564693034223000) and this ticket exists to explore how we can solve these problems and other problems that stem from the fact that we do not identify inactive Caseflow users.

## Acceptance criteria
* [x] Tech spec that outlines several possible paths forward for how to mark Caseflow users as inactive
* [x] Decision on how to proceed following discussion of the tech spec (which of the paths forward to implement)

Estimating this ticket at a 3 because that is our practice for all tickets where the end result is a tech spec.

*On rare occasions we have deleted rows from the `users` table. One example that comes to mind is [when we consolidated duplicate rows](https://github.com/department-of-veterans-affairs/caseflow/issues/9444) in that table.

pkarman:
fwiw I'd vote for the "soft delete" approach we follow in `HearingDay` with the `acts_as_paranoid` gem. It has some drawbacks, but I would enjoy the consistency of it in the codebase.

msj0nes:
Idea. If it ever boils down to a manual process where the Board has to notify us of staff leaving then it would be cool to have a ‘admin panel’ similar to ‘team management’ where we add the user/CSS ID. 

Adding the user/css ID to the listing would:
- Trigger the return of current cases assigned back to the assigning VLJ. 
- If the ex-attorney case is with admin before departure, the returned case from admin would go back to the VLJ. 
- A popup notification when users try to assign to these specific users. 

The VLJ from there makes the decision on what happens next.

hschallhorn:
### Things we're trying to solve

1. Develop a way to determine if a user is inactive
2. Make sure tasks aren't assigned to inactive users
3. Develop a way to mark a user inactive
4. Handle tasks assigned to user when they become inactive (this is similar to when a user leaves an org, which we also don't handle) (may want to spin out to a different spec to keep this at a 3!)

### Possible Solutions

[...]

PR #11810, opened 2019-08-16
Add active status for users

hschallhorn:
References #11631

### Description
Add status to the users column which defaults existing users and newly created users to "active

### Acceptance Criteria
- [x] Migration doesn't fail
- [x] All users are active after migration
- [x] Scope properly selects all active users
- [x] Creating a user defaults the status to active
- [x] Updating a user's status sets timestamp of change

### Testing Plan
1. Ensure migration causes no issues
2. Ensure all users are active after migration
```ruby
> User.count.eql?(User.active.count)`
=> true
```
3. Ensure creating a user defaults the status to active
```ruby
> FactoryBot.create(:user, station_id: 101, css_id: "NEW_USER")
=> #<User:0x00007fa8933bcef8
 id: 128,
 created_at: Fri, 16 Aug 2019 21:46:57 UTC +00:00,
 css_id: "NEW_USER",
 efolder_documents_fetched_at: nil,
 email: nil,
 full_name: "Lauren Roth",
 last_login_at: nil,
 roles: nil,
 selected_regional_office: nil,
 station_id: "101",
 updated_at: Fri, 16 Aug 2019 21:46:57 UTC +00:00,
 status: "active",
 status_updated_at: nil>
```
4. Ensure creating a user with an invalid status fails
```ruby
> FactoryBot.create(:user, station_id: 101, css_id: "NEW_USER_2",  status: "invalid")
ArgumentError: 'invalid' is not a valid status
```
5. Ensure updating a user's status updates the `status_updated_at` timestamp to now
```ruby
> User.last.update!(status: "inactive")
UPDATE "users" SET "status" = $1, "updated_at" = $2 WHERE "users"."id" = $3  [["status", "active"], ["updated_at", "2019-08-16 21:51:09.953541"], ["id", 128]]
UPDATE "users" SET "updated_at" = $1, "status_updated_at" = $2 WHERE "users"."id" = $3  [["updated_at", "2019-08-16 21:51:09.959955"], ["status_updated_at", "2019-08-16 21:51:09.959373"], ["id", 128]]
```

*Schema Changes Only*

* [x] Column comments updated


db/migrate/20190816194620_add_status_to_users.rb
```diff
@@ -0,0 +1,11 @@
+class AddStatusToUsers < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def change
+    add_column :users, :status, :string, comment: "Whether or not the user is an active user of caseflow"
+    add_column :users, :status_updated_at, :datetime, comment: "When the user's status was last updated"
+    add_index :users, :status, algorithm: :concurrently
+
+    User.update_all(status: "active")
```
pkarman: could we do this instead by setting the `default` value to "active" on the `add_column` call? that way the db is doing the work.

db/migrate/20190816210203_add_default_value_to_status.rb
```diff
@@ -0,0 +1,9 @@
+class AddDefaultValueToStatus < ActiveRecord::Migration[5.1]
+  def up
+    change_column_default :users, :status, "active"
```
pkarman: ah nm I see you did it here.

app/models/user.rb
```diff
@@ -26,6 +26,13 @@ class User < ApplicationRecord
 
   before_create :normalize_css_id
 
+  after_update :maybe_update_status_timestamp
```
pkarman: instead of a hook (which is a little flakey IME) could we just have an explicit method?

```ruby
def update_status!(new_status)
  update!(status: new_status, status_updated_at: Time.zone.now)
end
```

Issue #10511, opened 2019-04-23
Ability to intake decision reviews with missing ratings/issues 


leikkisa:
## Background
When a claims assistant is intaking an AMA claim for rating issues, Caseflow present a list of past rating issues to the user to select from.  Caseflow first fetches a list of the veteran's promulgated rating profiles, and then fetches the list of issues on each rating profile.  The issues indicate which issues changed from the previous rating profile (disabilities that were "at issue").

If a rating issue is not found, the user is currently directed to add an "Unidentified issue", which gets added to a rating end product.  The contention created in VBMS does not contain the issue description. Instead it indicates that the issue needs to be edited in Caseflow, where it can be removed and the correct issue can be added in its place if available.

For Supplemental Claims, there is not a timeliness requirement for contentions, so a veteran can submit a supplemental claim for decisions from any time in the past.  Some issue data, particularly for old issues, is not currently available.

Before 5/2/2003, ratings were stored on a user's individual computer.  Ratings created after that were created on the new system.  The old ratings can be backfilled, creating a backfilled rating.  Caseflow does not currently have access to the backfilled ratings, but it is something that can be added.  To do: add information about which backfilled ratings are technically considered promulgated (issues from the old system, issues that didn't result in a rating but were on a cleared EP).

Currently, for non-rating issues, we do not have access to the issue's historical data. We have users add them in Caseflow for the first time.  If that issue comes through AMA again, it is connected to that first issue and its history is tracked from there.

**Examples of missing ratings (to do: find specific examples)**

- Issues from before 2003
- Potentially issues after 2003 (denied issues that didn't result in a rating? Issues that just haven't changed since 2003?)
- Previously denied pension issues that do not have a rating

## Possible solutions 
(to do: add pros and cons, some information about what would be needed from Caseflow/BGS for implementation)

- Allow rating issues that aren't connected to their previous rating issue. We would collect the relevant information for this issue that would allow a VSR to confirm its decision.  Potential option to backfill data if we can get the data needed populated in BGS. **(Recommended)**
- Add rating issues that are connected to a disability if available. Not all missing issues will have this. This will not allow us to track the history of an issue before AMA. More investigation is needed on how to connect these.
[...]

PR #11813, opened 2019-08-16
Add RatingDecision for not-service-connected decisions

pkarman:
connects #10511 #11840 

### Description
BGS API returns not-service connected decisions. This adds a method `decisions` to the `Rating` model that returns `RatingDecision` records. These would be a superset of `RatingIssue` which seems to include only service connected decisions, or not-service-connected decisions that have explicit rating issue associated.

### DIFF
--- a/app/models/rating.rb
+++ b/app/models/rating.rb
@@ -72,13 +72,7 @@ class Rating
     return {} unless rating_profile[:disabilities]
 
     Array.wrap(rating_profile[:disabilities]).map do |disability|
-      decision = {}
-      decision[:type_name] = disability[:decn_tn]
-      decision[:rating_sequence_number] = disability[:disability_evaluations][:rating_sn]
-      decision[:disability_id] = disability[:dis_sn]
-      decision[:diagnostic_text] = disability[:disability_evaluations][:dgnstc_txt]
-      decision[:profile_date] = disability[:disability_evaluations][:prfl_dt]
-      RatingDecision.from_bgs_hash(decision)
+      RatingDecision.from_bgs_disability(self, disability)
     end
   end
 
--- a/app/models/rating.rb
+++ b/app/models/rating.rb
@@ -69,7 +69,7 @@ class Rating
   end
 
   def decisions
-    return {} unless rating_profile[:disabilities]
+    return [] unless rating_profile[:disabilities]
 
     Array.wrap(rating_profile[:disabilities]).map do |disability|
       RatingDecision.from_bgs_disability(self, disability)
--- a/app/models/rating.rb
+++ b/app/models/rating.rb
@@ -41,7 +41,8 @@ class Rating
       participant_id: participant_id,
       profile_date: profile_date,
       promulgation_date: promulgation_date,
-      issues: issues.map(&:serialize)
+      issues: issues.map(&:serialize),
+      decisions: decisions.map(&:serialize)
     }
   end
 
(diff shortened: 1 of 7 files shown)


spec/models/rating_issue_spec.rb
```diff
@@ -10,7 +10,6 @@
 
   let(:profile_date) { Time.zone.today - 30 }
   let(:promulgation_date) { Time.zone.today - 30 }
-  let(:profile_date) { Time.zone.today - 40 }
```
pkarman: zap duplicate line

app/models/rating.rb
```diff
@@ -67,6 +68,14 @@ def pension?
     associated_claims_data.any? { |ac| ac[:bnft_clm_tc].match(/PMC$/) }
   end
 
+  def decisions
```
leikkisa: Should we call this disabilities?  It doesn't seem like there's specifically a "Decision" section, and the disabilities section has some decision data, and the rating issues section has other decision data (such as decision text).

app/models/rating_decision.rb
```diff
@@ -0,0 +1,55 @@
...
+      new(
+        type_name: disability[:decn_tn],
+        rating_sequence_number: disability[:disability_evaluations][:rating_sn],
+        rating_reference_id: disability[:disability_evaluations][:rba_issue_id], # ok if nil
+        disability_id: disability[:dis_sn],
+        diagnostic_text: disability[:disability_evaluations][:dgnstc_txt],
+        diagnostic_type: disability[:disability_evaluations][:dgnstc_tn],
+        diagnostic_code: disability[:disability_evaluations][:dgnstc_tc],
+        disability_date: disability[:dis_dt],
+        profile_date: disability[:disability_evaluations][:prfl_dt],
```
leikkisa: I don't think this is related to the real profile date. The same disability will show up in multiple ratings and the profile date I think will just be the profile date of that rating. The disability date seems to do the same.

app/models/rating_decision.rb
```diff
@@ -0,0 +1,55 @@
...
+                :rating_sequence_number,
+                :rating_reference_id,
+                :type_name
+
+  class << self
+    def from_bgs_disability(rating, disability)
+      new(
+        type_name: disability[:decn_tn],
+        rating_sequence_number: disability[:disability_evaluations][:rating_sn],
+        rating_reference_id: disability[:disability_evaluations][:rba_issue_id], # ok if nil
```
leikkisa: This comment is useful! We could be more explicit and say it'll be nil if it's not service connected or if there's not a rating issue for it.

app/models/rating_decision.rb
```diff
@@ -0,0 +1,55 @@
...
+        disability_id: disability[:dis_sn],
+        diagnostic_text: disability[:disability_evaluations][:dgnstc_txt],
+        diagnostic_type: disability[:disability_evaluations][:dgnstc_tn],
+        diagnostic_code: disability[:disability_evaluations][:dgnstc_tc],
+        disability_date: disability[:dis_dt],
+        profile_date: disability[:disability_evaluations][:prfl_dt],
+        participant_id: rating.participant_id,
+        benefit_type: rating.pension? ? :pension : :compensation
```
leikkisa: Can we add the disability SN?  I think that will be useful when we show them in the list, because that data is also in the rating issues and might be useful for de-duping.

app/models/rating.rb
pkarman: I was intentionally trying to stay away from "disabilities" because even though that's what the top-level key is called, the particular bits we are pulling out are related to the decision. The name comes from the important ID `ratingSn`. The BGS docs say:

> Rating Decision – Rating Decision Identifier | A system generated identifier created to uniquely identify each instance of RATING DECISION. (Rating Sequence Number)

So I called it `decisions` returning `RatingDecision` objects to mirror `issues` (which returns `RatingIssue` objects).

There are just a few words in VA Land that are so overused (decision, issue, rating, appeal) that they are ambiguous, so I was trying for consistency to at least mitigate inevitable confusion.

app/models/rating_decision.rb
pkarman: I noticed in some of the [old results](https://github.com/department-of-veterans-affairs/caseflow/issues/10511#issuecomment-522163058) that this date could vary from the parent profile date, as did the disability date. So I wanted to capture it from the specific disability just to be sure. I can use the `rating.profile_date` instead like `RatingIssue` does, but in case it was different I thought it useful to capture.

app/models/rating_decision.rb
pkarman: I noticed for newer data that there were some not-service-connected decisions that had rating issue ids. It is not purely one or the other.

app/models/rating_decision.rb
pkarman: it's in there as `disability_id` a few lines above.

pkarman: agreed about de-duping. Right now I plan to filter out based on whether there is a `rating_reference_id` or not.

PR #11819, opened 2019-08-19
Stop creating JudgeTasks with action argument

lowellrex:
Connects #9057. Part of a larger effort to remove the `action` field from the tasks table. This PR stops adding superfluous fields to the statements that creates a new `JudgeAssignTask`.


app/models/docket.rb
```diff
@@ -72,12 +72,7 @@ def docket_appeals
   def assign_judge_tasks_for_appeals(appeals, judge)
     appeals.map do |appeal|
       Rails.logger.info("Assigning judge task for appeal #{appeal.id}")
-      task = JudgeAssignTask.create!(appeal: appeal,
-                                     parent: appeal.root_task,
-                                     appeal_type: Appeal.name,
```
lowellrex: Handled appropriately by single-table inheritance.

app/models/docket.rb
```diff
@@ -72,12 +72,7 @@ def docket_appeals
...
-                                     assigned_at: Time.zone.now,
```
lowellrex: Already covered by the base class: https://github.com/department-of-veterans-affairs/caseflow/blob/1b54c660784f522af08da102cd5703c16dfc9a48/app/models/task.rb#L490

PR #11820, opened 2019-08-19
Ignore action column of tasks table

lowellrex:
Connects #9057. Part of a larger effort to remove the `action` field from the tasks table. This PR paves the way for us to remove the column in the database after this has been live without issue for a little while (2 weeks?).


app/models/task.rb
```diff
@@ -60,6 +60,8 @@ class << self; undef_method :open; end
                                  )
                                }
 
+  self.ignored_columns = %w[action]
```
lomky: Just in case this slips priority and we take awhile to get back to it, can we add a comment referencing the ticket for context?

PR #11822, opened 2019-08-19
Datadog Service through Caseflow Job

### LINKED ISSUE #11626: Extend the CaseflowJob class with Duration Report Option
Extend the CaseflowJob class with the ability to record the runtime of the job and send it to DataDog.

Refactor the places we currently do this with this new class option.

[Existing implementations](https://github.com/department-of-veterans-affairs/caseflow/search?q=job_duration_seconds&unscoped_q=job_duration_seconds)

lomky:
Resolves #11626 

Integrates DataDogService into CaseflowJob


app/jobs/task_timer_job.rb
```diff
@@ -24,7 +25,7 @@ def perform
       cancel(task_timer)
     end
 
-    record_runtime(start_time)
+    datadog_service.report_runtime
```
lowellrex: What do you think of having the interface just be a call to `record_runtime`, so it would be simple to add it to other jobs? We could set a `start_time` attribute on the job when we instantiate it. An added benefit here is that we don't have to carry the overhead of an instance of `DatadogService` in memory (which could be an avenue for holding onto network connections for a while) throughout the entire lifetime of the job.

lomky: I don't object. I could see us in the future wanting more than just the `record_runtime` available through the interface, but it would be easy enough to extend that when we get there.

app/jobs/task_timer_job.rb
```diff
@@ -12,7 +12,7 @@ class TaskTimerJob < CaseflowJob
   #  grab task type and appeal from tasks
 
   def perform
-    start_time = Time.zone.now
+    set_job_start_time
```
lowellrex: We can make `start_time` a property of each instance and we won't have to explicitly set it. Perhaps something like:

```ruby
  def initialize(args)
    @start_time = Time.zone.now
  end
```

app/jobs/task_timer_job.rb
```diff
@@ -24,7 +24,7 @@ def perform
       cancel(task_timer)
     end
 
-    record_runtime(start_time)
+    datadog_report_runtime(app_name: APP_NAME, metric_group_name: METRIC_GROUP_NAME)
```
lowellrex: Let's follow the model we established with `slack_url` and make these parameters instance methods.

PR #11823, opened 2019-08-19
Update label for ColocatedTasks created by judges

### LINKED ISSUE #8759: Change mark task complete label for ColocatedTasks created by judges
Currently the action that marks a `ColocatedTask` complete is labelled "Send back to attorney". With #8608 we added the enabled judges to send tasks to the VLJ support staff, so this label is no longer always accurate. This ticket exists to change the label of this action to more precisely explain what will happen when the action is taken. I imagine this could be accomplished by conditionally changing the label to read "Send back to judge" when the judge was the one who created the task or just renaming this label to read "Mark task complete" to cover both cases (the confirmation box that pops.

Before | After | Alternative
--- | --- | ---
image | image | image

An argument in favor of the alternative ("Mark task complete") approach is that the confirmation box that pops up when the "Send back to attorney" option is selected already looks like this:

image

## Acceptance criteria
* [ ] Change the label for the "Send back to attorney" action to better describe what will happen when that action is taken

lowellrex:
Resolves #8759. Changes the label of the "mark task complete" action to "Send back to judge" if the creator of the task is a judge.

Before | After
--- | ---
image | image

### DIFF
--- a/app/models/tasks/colocated_task.rb
+++ b/app/models/tasks/colocated_task.rb
@@ -76,16 +76,9 @@ class ColocatedTask < Task
   def available_actions(user)
     if assigned_to == user ||
        (task_is_assigned_to_user_within_organization?(user) && Colocated.singleton.user_is_admin?(user))
-      
-      # Use assigner so that we handle creation of the ColcoatedTask from LegacyTasks gracefully.
-      return_action = if JudgeTeam.for_judge(assigned_by)
-                        Constants.TASK_ACTIONS.COLOCATED_RETURN_TO_JUDGE.to_h
-                      else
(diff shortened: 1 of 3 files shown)


app/models/tasks/colocated_task.rb
```diff
@@ -92,6 +93,15 @@ def available_actions(user)
     []
   end
 
+  def return_to_assigner_action
+    # Use assigner so that we handle creation of the ColcoatedTask from LegacyTasks gracefully.
+    if JudgeTeam.for_judge(assigned_by)
```
lomky: "If the person who assigned this has a judge team"? Am I reading that correctly?

lowellrex: Exactly!

Issue #11802, opened 2019-08-16
Hearing API missing facility data for some veterans


pkarman:
The Hearings API response is missing some facility information for some veterans.

Example participant ids:

30694233
31153080
35510455
6230049
3236701
23662439
2769203
32626624
8232736
8232736
8710546
34330243
29529926
36464464
38938523
11137301
38494203
12997228
300686
35115929
45123823
7682350
32770898

lomaxap:
For upcoming hearings

`LegacyHearing.where(id: [116731, 120768, 115536, 110825, 121824, 121584, 107827, 122758, 120158, 108901, 121825, 122272, 116322, 117246, 121823])` (15)

There are no `nil` locations for future AMA hearings

We did not make hearing locations a mandatory field until around April. @premal I think we can assume that if a hearing location is nil, then the hearing will be at the RO, but would like to confirm with someone at the branch. Can you reach out?

pkarman:
Possibly related to https://github.com/department-of-veterans-affairs/dsva-vacols/issues/74

pshahVA:
worked with @lomaxap and he determined that the first one above doesn't have a legacy or ama appeal.

PR #11841, opened 2019-08-20
Use hearing_location.name for regional_office in hearings API

### LINKED ISSUE #11834: Hearings API regional_office should reflect hearing location
The Hearings API returns a record like:

```json
{
  "address":"100 SW Main Street Floor 2",
  "city":"Portland",
  "appeal":"701dc7a2-732e-4eca-9dc5-f024adf0719d",
  "facility_id":"vba_348",
  "participant_id":"31634112",
  "regional_office":"Seattle regional office",
  "room":"7",
  "scheduled_for":"2019-08-23T12:30:00.000-07:00",
  "state":"OR",
  "timezone":"America/Los_Angeles",
  "zip_code":"97204"
}
```

Is this correct?

pkarman:
Resolves #11834 

### Description
use the hearing_location.name as the `regional_office` value.

### DIFF
--- a/app/serializers/api/v2/hearing_serializer.rb
+++ b/app/serializers/api/v2/hearing_serializer.rb
@@ -20,10 +20,10 @@ class Api::V2::HearingSerializer
   end
   # deprecated :regional_office
   attribute :regional_office do |hearing|
-    hearing.hearing_location&.name
+    hearing.hearing_location&.name || hearing.regional_office_name
   end
   attribute :hearing_location do |hearing|
-    hearing.hearing_location&.name
+    hearing.hearing_location&.name || hearing.regional_office_name
   end
   attribute :room do |hearing|
     hearing.hearing_day.room
--- a/app/serializers/api/v2/hearing_serializer.rb
+++ b/app/serializers/api/v2/hearing_serializer.rb
@@ -18,7 +18,11 @@ class Api::V2::HearingSerializer
   attribute :participant_id do |hearing|
     hearing.appeal.veteran.participant_id
   end
+  # deprecated :regional_office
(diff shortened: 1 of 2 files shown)


app/serializers/api/v2/hearing_serializer.rb
```diff
@@ -18,7 +18,9 @@ class Api::V2::HearingSerializer
   attribute :participant_id do |hearing|
     hearing.appeal.veteran.participant_id
   end
-  attribute :regional_office, &:regional_office_name
+  attribute :regional_office do |hearing|
```
lomaxap: i think "regional office" has a specific meaning in the VA. e.g the hearing location can be a health center or an employment center. is there any concern there?

pkarman: that's a good point. we should rename it. I'll need to give the VEText folks a heads up.

app/serializers/api/v2/hearing_serializer.rb
```diff
@@ -18,7 +18,9 @@ class Api::V2::HearingSerializer
...
+    hearing.hearing_location&.name
```
lomaxap: there are a few records that were created before hearing location mandatory/existed and the location is blank (#11802) so could do `hearing.hearing_location&.name || hearing.regional_office_name`

app/serializers/api/v2/hearing_serializer.rb
pkarman: I added the new key name and marked the old as deprecated. We can remove it in a week. I emailed the VEText folks to that effect.

pkarman:
thanks @lomaxap -- PR updated with your suggestions.

Issue #11840, opened 2019-08-20
Add rating decisions to Intake from before 2005


pkarman:
Originating from #10511 this ticket describes older BGS rating data that does not contain rating issue information for older (pre-2005) decisions.

See https://github.com/department-of-veterans-affairs/appeals-support/issues/5319 for support issue and referenced OIT tickets.

These past decisions appear in VBMS under the "Rating Profile" tab but are not populating in the Caseflow Add Issues screen during Intake.

Our current plans involve 2 threads:

* [x] Include the rating decisions currently returned by BGS as part of the rating_profile API response. These include non service connected and service connected decisions. We will include the non service connected decisions that do not have a `rating_issue_id` that are currently not included in the Add Issues Intake screen.
* [ ] Explore a new BGS API endpoint that may contain the missing `rating_issue_id` values for these non service connected decisions. This requires some additional infrastructure work in Caseflow in order to allow us to preview this API endpoint with production-like data while it is in development.

One or both solutions should allow us to include these missing rating decisions in Caseflow Intake and unblock users.

pkarman:
production participant ids to use for development and testing:

33380576
31140273
4563635

pkarman:
Now behind feature toggle. Enable for your user via prod rails console.
```
FeatureToggle.enable!(:contestable_rating_decisions, users: [you.css_id])
```

pkarman:
in production just now:

```
RequestIssue.where.not(contested_rating_decision_reference_id: nil).count
=> 243
```

17 of which have already been decided.

pkarman:
According to https://github.com/department-of-veterans-affairs/appeals-support/issues/5324 we may need to revisit the +/- 10 day date window comparison logic for whether a rating decision is contestable or not. See https://dsva.slack.com/archives/CHX8FMP28/p1568212423253600

pkarman:
New BGS endpoint at

```ruby
BGSService.new.client.rating_profile.find_in_date_range(participant_id: 13774779, start_date: '1999-01-01', end_date: '2019-09-24')
```

but currently returning SOAP error.

pkarman:
Ongoing list of rating issues still missing from intake screen: https://github.com/department-of-veterans-affairs/appeals-support/issues/5319#issuecomment-526672743

pkarman:
As part of this effort we should write a report showing the number of unidentified issues vs rating decision issues week-over-week. That should illustrate whether we are making any progress.

[...]

PR #11846, opened 2019-08-20
Contestable rating decisions

pkarman:
connects #11840 

### Description
Adds `contested_rating_decision_reference_id` to RequestIssue. This unique id applies to all BGS rating decisions regardless of whether they are attached to a rating issue.

*Schema Changes Only*

* [x] Column comments updated


app/models/contestable_issue.rb
```diff
@@ -43,13 +44,27 @@ def from_decision_issue(decision_issue, contesting_decision_review)
         is_rating: decision_issue.rating?
       )
     end
+
+    def from_rating_decision(rating_decision, contesting_decision_review)
+      new(
+        rating_issue_reference_id: rating_decision.rating_reference_id,
+        rating_issue_profile_date: rating_decision.profile_date.to_date,
+        rating_issue_sequence_id: rating_decision.rating_sequence_number,
+        approx_decision_date: rating_decision.disability_date.to_date,
```
leikkisa: I don't think we can use this here. From what I've looked at (cross referencing disabilities that we do have rating issue data for), the disability shows up in multiple rating profiles and the disability date matches the rating profile date.

So let's say a disability is from the year 1998, and since then there have been 2 rating profiles. One is from yesterday, and one is from a year ago.  The disability will look like it was decided yesterday in one of the profiles, and also a year ago in the other profile.

Since we use this for timeliness checks, I don't think we can use this here since it doesn't seem to have anything to do with the decision date.

leikkisa: We might need to have users enter this.

pkarman: good point. I think we should use the promulgation_date from the parent rating profile, just like we do for rating_issues. I think that will be the most consistent.

pkarman:
> I don't understand what you want to use the ratingSn for

I was looking for a unique reference id we could use to refer to these rating decisions. I'm happy to use the `disSn` instead. I changed the column name to `contested_rating_decision_reference_id` so that it is more clear what we are trying to capture, and we can store whatever value there is most useful.

app/models/contestable_issue.rb
```diff
@@ -91,6 +106,7 @@ def title_of_active_review
   def contested_by_request_issue
     RequestIssue.active.find_by(
       contested_rating_issue_reference_id: rating_issue_reference_id,
+      contested_rating_decision_reference_id: rating_decision_reference_id,
```
leikkisa: I have more questions on this since the rating_decision_reference_id may not be stable for the same disability over time, but I think it makes sense to worry about this more when we start looking at handling the duplicate nature of these.

PR #11850, opened 2019-08-21
Upgrade diff npm lib

pkarman:
Revert "Revert "Bump diff from 3.2.0 to 3.5.0 in /client (#11549)" (#11574)"

This reverts commit fa5127828a286d580e3f58371d69cf9d96432293.

supercedes #11549 which was reverted.


pkarman:
one of several GH security alerts I can see on the repo.

jcq:
Yeah, I guess I was wondering why it was originally reverted; I don't see anything about breaking changes, but I do see folks discussing the security issue addressed in 3.5.0.

pkarman:
there were a whole series of npm updates that got reverted because one of them was causing a deployment failure and we didn't know which one it was. The root case turned out to be webpack-dev-server https://github.com/department-of-veterans-affairs/caseflow/network/alert/client/yarn.lock/webpack-dev-server/open

PR #11854, opened 2019-08-21
disability_evaluations may be plural array

pkarman:
regression bug


app/models/rating_decision.rb
```diff
@@ -21,16 +21,17 @@ class RatingDecision
 
   class << self
     def from_bgs_disability(rating, disability)
+      latest_evaluation = Array.wrap(disability[:disability_evaluations]).max_by { |dis_eval| dis_eval[:dis_dt] } || {}
```
pkarman: this logic comes from Rating model

Issue #11849, opened 2019-08-21
Properly handle old versions of automatically routed ColocatedTasks


hschallhorn:
As noted in [this](https://dsva.slack.com/archives/CHX8FMP28/p1566347997498800?thread_ts=1566340326.497600&cid=CHX8FMP28) slack conversation, when #11467 was merged, `FoiaColocatedTask`s, `MissingHearingTranscriptColocatedTask`s, and `TranslationColocatedTask`s that were already "assigned" or "on_hold" in production became hidden in the queue table and task snapshot. This ticket exists to either migrate them to their new state (assigned to the default_assignee with a child `FoiaTask`, `TranscriptionTask`, `TranslationTask` respectively) or fix the current code to allow these tasks to progress as they used to, then clean up the code once they've all been completed.

hschallhorn:
### Translation Colocated Tasks

```ruby
> colocated_org = Colocated.singleton
> old_translation_tasks = TranslationColocatedTask.open.where(assigned_to_type: Organization.name, assigned_to_id: colocated_org.id);
> old_translation_tasks.count
=> 14
> old_translation_child_tasks = TranslationColocatedTask.open.where(parent_id: old_translation_tasks.pluck(:id));
>  old_translation_child_tasks.group(:status).count
=> {"on_hold"=>3, "in_progress"=>1, "assigned"=>10}
> Task.open.where(parent_id: old_translation_child_tasks.on_hold.pluck(:id)).group(:type).count
=> {"TranslationTask"=>3}
```

### Transcription Colocated Tasks

```ruby
> old_transcription_tasks = MissingHearingTranscriptsColocatedTask.open.where(assigned_to_type: Organization.name, assigned_to_id: colocated_org.id);
> old_transcription_tasks.count
=> 79
> old_transcrpition_child_tasks = MissingHearingTranscriptsColocatedTask.open.where(parent_id: old_transcription_tasks.pluck(:id));
>  old_transcrpition_child_tasks.group(:status).count
=> {"in_progress"=>1, "on_hold"=>9, "assigned"=>69}
> Task.open.where(parent_id: old_transcrpition_child_tasks.on_hold.pluck(:id)).group(:type).count
=> {"PrivacyActTask"=>1, "TimedHoldTask"=>8}
```

### FOIA Colocated Tasks

```ruby
> old_foia_tasks = FoiaColocatedTask.open.where(assigned_to_type: Organization.name, assigned_to_id: colocated_org.id);
[...]

PR #11855, opened 2019-08-21
Create pre routed subclass for old style colocated tasks in production

hschallhorn:
References #11849

### Description
Creates classes to transition old style colocated tasks that have new routing logic back to their original colocated logic so they may be worked in caseflow by VLJ support staff


app/models/tasks/pre_routing_colocated_task.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class PreRoutingColocatedTask < ColocatedTask
```
lowellrex: Can we add a comment in the code that explains why we need to make the distinction between pre- and post-automatically routing `ColocatedTask`?

lowellrex:
Can we add a comment in `ColocatedTask.find_subclass_by_action` that explains that we are only searching for direct subclasses of `ColocatedTask` instead of all `descendant` classes in order to avoid instantiating any `PreRoutingColocatedTask`s?

Alternatively, we could make the code more "self-documenting" by doing something like selecting all descendant classes and adding a test for whether we expect to create new instances of that class that `PreRoutingColocatedTask` could overwrite. This might be more explicit than relying on us remembering that the `subclasses` method only gets direct child classes instead of all descendants.

```ruby
  # in pre_routing_colocated_task.rb
  def self.can_create_new_instances?
    false
  end

  # in colocated_task.rb
  def self.can_create_new_instances?
    true
  end

  def find_subclass_by_action(action)
    descendants.find { |task_class| task_class.can_create_new_instances? && task_class.label == Constants::CO_LOCATED_ADMIN_ACTIONS[action] }
  end
```

PR #11856, opened 2019-08-21
Match format of veteran name as saved in appeals attributes table

hschallhorn:
### Description
Match test of name format to [how veteran names are saved](https://github.com/department-of-veterans-affairs/caseflow/blob/086e0d822c6966712edea67f8e52e1dc47444ec6/app/jobs/update_cached_appeals_attributes_job.rb#L177) in the cache appeal attributes job

### DIFF
--- a/spec/models/task_pager_spec.rb
+++ b/spec/models/task_pager_spec.rb
@@ -278,16 +278,20 @@ describe TaskPager, :all_dbs do
           create(
             :cached_appeal,
             appeal_id: task.appeal_id,
-            veteran_name: "#{last_name.split(' ').last}, #{first_name.split(' ').first}"
+            veteran_name: "#{last_name.split(' ').last}, #{first_name}"
           )
         end
       end
 
       it "sorts by veteran last and first name" do
         expected_order = created_tasks.sort_by do |task|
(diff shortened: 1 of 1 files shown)


spec/models/task_pager_spec.rb
```diff
@@ -278,16 +278,20 @@
...
         expected_order = created_tasks.sort_by do |task|
-          "#{task.appeal.veteran_last_name.split(' ').last}, #{task.appeal.veteran_first_name.split(' ').first}"
+          "#{task.appeal.veteran_last_name.split(' ').last}, #{task.appeal.veteran_first_name}"
         end
-        expect(subject.map(&:appeal_id)).to eq(expected_order.map(&:appeal_id))
+        expect(subject.map do |task|
+          "#{task.appeal.veteran_last_name.split(' ').last}, #{task.appeal.veteran_first_name}"
+        end).to eq(expected_order.map do |task|
+          "#{task.appeal.veteran_last_name.split(' ').last}, #{task.appeal.veteran_first_name}"
+        end)
```
hschallhorn: Added bonus of more descriptive failures for [this flakey](https://github.com/department-of-veterans-affairs/caseflow/issues/10516#issuecomment-523050358) test as it does not fail locally

lomky: could we name the enumeration variables `subject_task` and `expected_task`, or something like that?

PR #11864, opened 2019-08-22
Add UNSAFE_ prefix to deprecated React lifecycle methods

jcq:
### Description
Various React lifecycle hooks [have been deprecated in recent versions](https://reactjs.org/blog/2019/08/08/react-v16.9.0.html#renaming-unsafe-lifecycle-methods), and will be removed completely in v17.x. This is the first step to fixing up our code accordingly; we prefix these deprecated methods with `UNSAFE_`, which will silence the warnings for the time being.

I've created a follow-up ticket to go through and remove the deprecated methods in favor of better practices:
https://github.com/department-of-veterans-affairs/caseflow/issues/11866

### Acceptance Criteria
- [ ] Code compiles correctly


jcq:
> I can override the codeclimate `TODO` flags, but I think you'll need to add `/* eslint-disable react/prop-types */` to the top of the rest of the files unless @pkarman can override circleci failures

Yeah, I think I'd prefer to not add the comment to disable eslint in order to not forget to do it as part of the later refactoring if possible.

PR #11865, opened 2019-08-22
Geomatch All Facilities

lomaxap:
Changes geomatching to find the closest VA facility (which can be an RO, VHA etc) then find the regional office associated with that facility.

- Adds `VaDotGovAddressValidator::ClosestRegionalOfficeFinder` class to manage new process


client/constants/REGIONAL_OFFICE_INFORMATION.json
```diff
@@ -258,8 +258,7 @@
       "vha_657GO",
       "vha_626GH",
       "vc_0719V",
-      "vc_0701V",
-      "vc_0719V"
+      "vc_0701V"
```
lomaxap: vc_0719V was duplicated

app/services/external_api/va_dot_gov_service.rb
```diff
@@ -46,7 +46,7 @@ def send_facilities_requests(ids:, query:)
       response = nil
 
       until remaining_ids.empty? || response.try(:next?) == false || response.try(:success?) == false
-        response = send_facilities_request(query: query.merge(page: page, per_page: 100)).merge(response)
+        response = send_facilities_request(query: query.merge(page: page, per_page: 200)).merge(response)
```
lomaxap: there are currently 158 total facility ids. Upping the number of response per page to get all facilities in a single go. Will need to test in UAT to make sure this works.

app/services/va_dot_gov_address_validator/closest_regional_office_finder.rb
```diff
@@ -0,0 +1,44 @@
...
+class VaDotGovAddressValidator::ClosestRegionalOfficeFinder
+  attr_reader :facilities, :closest_facility_id
+
+  def initialize(facilities:, closest_facility_id:)
+    @facilities = facilities
+    @closest_facility_id = closest_facility_id
+  end
+
+  def call
+    possible_regional_offices_distances.min_by { |ro| ro[:distance] }.dig(:regional_office_key)
```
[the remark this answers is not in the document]
lomaxap: this should never happen. if it does, there was an issue upstream at the Facilities API. We should know when it does though, so I added a failure here

PR #11833, opened 2019-08-19
WIP Send filter options to front-end from back-end

lowellrex:
Resolves #11509.

PR #11871, opened 2019-08-22
Prepare frontend for backend column structure change

lowellrex:
Allows the front-end to accept an array of either strings or objects in anticipation of the data structure changes that back-end will return when #11833 is merged (will change from strings to objects then). Since the javascript bundles are cached by the browser we need to get this front-end change into production before the back-end change to avoid page rendering errors that would result from the old front-end code being called with the new data structure.

### DIFF
--- a/client/app/queue/OrganizationQueue.jsx
+++ b/client/app/queue/OrganizationQueue.jsx
@@ -4,6 +4,7 @@ import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import { css } from 'glamor';
 import { sprintf } from 'sprintf-js';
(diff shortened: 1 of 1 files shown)


client/app/queue/OrganizationQueue.jsx
```diff
@@ -159,6 +162,19 @@ class OrganizationQueue extends React.PureComponent {
   };
 }
 
+OrganizationQueue.propTypes = {
```
lomky: For future record: this change is to satisfy our newer eslint 'undefined properties requirement, and unrelated to the meat of the PR.

PR #11883, opened 2019-08-23
Bump React Router

jcq:
### Description
This bumps the dependencies for react-router (& related) here, as well as bumping to a newer version of caseflow-frontend-toolkit which includes updated react, react-router deps.

Under the old version of react-router, we were encountering an error in QueueApp due to its use of an array for the `path` prop:
https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/QueueApp.jsx#L291

That syntax is only supported in more recent versions, and was causing a cascade of other errors that prevented the page from rendering.

### Acceptance Criteria
- [ ] Code compiles correctly
- [ ] Organization queue view loads correctly

### Testing Plan

1. Choose a user who has a Queue organization (such as JOLLY_POSTMAN) and go to their queue
2. Choose "Switch Views" and choose their organization's queue of cases
3. Make sure page loads

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -28,7 +28,7 @@
     "client/node_modules"
   ],
   "dependencies": {
-    "@department-of-veterans-affairs/caseflow-frontend-toolkit": "https://github.com/department-of-veterans-affairs/caseflow-frontend-toolkit#d386d03",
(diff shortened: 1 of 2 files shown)


jcq:
This is sort of dependent upon this PR for caseflow-frontend-toolkit:
https://github.com/department-of-veterans-affairs/caseflow-frontend-toolkit/pull/52

(It doesn't actually have to have landed in master, as we are simply referencing that dependency via commit hash in this package.json)

PR #11884, opened 2019-08-23
Fix pager sort to ignore case and consistent ruby-vs-sql

pkarman:
If you ask Ruby which comes first alphabetically, `Newark` or `New York`, you'll get a different answer than if you ask PostgreSQL the same question. Substituting an underscore for a space in Ruby makes them sort the same way, predictably.

### DIFF
--- a/app/models/task_sorter.rb
+++ b/app/models/task_sorter.rb
@@ -31,8 +31,24 @@ class TaskSorter
 
   private
 
+  def sort_requires_case_norm?(col)
+    return false if col =~ /_at$/
+    return false if col =~ /^is_/
+    return false if col =~ /_count$/
+
+    true
+  end
+
   def order_clause
-    column.sorting_columns.map { |col| "#{column.sorting_table}.#{col} #{sort_order}" }.join(", ")
+    clauses = column.sorting_columns.map do |col|
+      tbl_clause = if sort_requires_case_norm?(col)
+                     "UPPER(#{column.sorting_table}.#{col})"
+                   else
+                     "#{column.sorting_table}.#{col}"
+                   end
+      "#{tbl_clause} #{sort_order}"
+    end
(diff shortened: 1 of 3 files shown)


app/models/task_sorter.rb
```diff
@@ -31,8 +31,24 @@ def sorted_tasks
...
+    return false if col =~ /^is_/
+    return false if col =~ /_count$/
+
+    true
+  end
+
   def order_clause
-    column.sorting_columns.map { |col| "#{column.sorting_table}.#{col} #{sort_order}" }.join(", ")
+    clauses = column.sorting_columns.map do |col|
+      tbl_clause = if sort_requires_case_norm?(col)
```
lowellrex: What's the disadvantage in always normalizing casing?

pkarman: some columns do not support it: timestamps, integers, booleans. if we put `UPPER` on everything we get psql errors.

app/models/task_sorter.rb
```diff
@@ -31,8 +31,24 @@ def sorted_tasks
 
   private
 
+  def sort_requires_case_norm?(col)
+    return false if col =~ /_at$/
+    return false if col =~ /^is_/
+    return false if col =~ /_count$/
```
lowellrex: Can we add a code comment here that explains that we're using column name as a heuristic for the data type?

app/models/task_sorter.rb
tomas-nava: Do these regexes capture all the timestamps, integers, and booleans that'll come through the task sorter? We definitely have columns of all three types that are not named according to those patterns.

pkarman: they cover all the use cases we currently test. I kept adding them till all the tests in `spec/models/task_sorter_spec.rb` passed.

Issue #11907, opened 2019-08-27
Alert Intake users when their async jobs fail


pkarman:
## Background

AMO has indicated that stuck async jobs are one of their highest priorities.

One pain point is the lack of visibility to users when their intakes are not successfully established in VBMS. Often they will re-do the intake, which will also fail if the underlying cause of failure is due to some data misalignment upstream in CorpDB.

One suggested pattern is to surface async job failures as alerts in the UI so that users are aware they have pending jobs that have not yet succeeded.

As a first iteration of this, create an Inbox style message queue and use the existing Alert style component to make Intake users aware.

## User story

As a claims assistant (intake user), I would like to be alerted if a claim I have established is stuck and will not complete.

## AC

* [ ] If an async claim review establishment is still failing after 24 hours, a message will be sent to the intake user. The content of the message will depend on whether an incident ticket has already been opened or not for the job error. We'll know if an incident has been opened based on a job note. See #12214 for details on adding notes to jobs. Message text TBD in consult with AMO.
* [ ] On the eventual success of a job that had previously taken longer than 24 hours, a confirmation message will be sent indicating establishment was successful and there is no further action necessary.
* [ ] If a job is cancelled manually (usually because it will never succeed), a job note should be created by whomever cancels the job and the note will be sent as a message to the user indicating whether any further action is required. See #12214 for details on job notes.

## Inbox

![Screen Shot 2019-08-27 at 10 57 18 AM](https://user-images.githubusercontent.com/1205061/63787493-89540280-c8b9-11e9-8340-6703ea3b85eb.png)

## Intake alert

![Screen Shot 2019-08-27 at 11 34 04 AM](https://user-images.githubusercontent.com/1205061/63790304-9d4e3300-c8be-11e9-828c-60ab3282d8c7.png)

pkarman:
Based on conversations with Jane C, Lauren S and Fred S, we think the following messaging pattern will work:

* No messaging sent to `Admin Intake` supervisors at this point. Only the Intake user will receive messages.
* If an async claim review establishment is still failing after 24 hours, a message will be sent to the intake user. The content of the message will depend on whether an incident ticket has already been opened or not for the job error. We'll know if an incident has been opened based on a job note. See #12214 for details on adding notes to jobs.
* On the eventual success of a job, a confirmation message will be sent indicating establishment was successful and there is no further action necessary.
[...]

PR #11913, opened 2019-08-27
Inbox and intake messages

pkarman:
connects #11907 

### Description
Adds a new endpoint called "inbox" where user messages are managed.

Feature toggle `:inbox`.

*Schema Changes Only*

* [x] Column comments updated

Screen Shot 2019-08-29 at 10 07 08 AM


pkarman:
thanks for the comments re: paginated jobs, @jcq -- yes that has never seen a designer's eyes and likely needs a total re-think. Out of scope here.

PR #11928, opened 2019-08-28
Use column names from columns rather than individual constants

hschallhorn:
### Description
As of #11750 we added more logic for each of the queue table columns. This PR simply uses the new logic and removes the old constants.

### Acceptance Criteria
- [ ] Use  `Constants.QUEUE_CONFIG.COLUMNS.<column>.name` over `Constants.QUEUE_CONFIG.<column>_COLUMN`

### Testing Plan
1. Nothing breaks 🤞


spec/models/task_sorter_spec.rb
```diff
@@ -106,7 +106,7 @@
       end
 
       context "when sorting by due date" do
-        let(:column_name) { Constants.QUEUE_CONFIG.TASK_DUE_DATE_COLUMN }
+        let(:column_name) { Constants.QUEUE_CONFIG.COLUMNS.TASK_DUE_DATE.name }
 
         before do
           tasks.each do |task|
```
lomky: One missed `TASK_ASSIGNER_COLUMN` down on line 124!

hschallhorn: Fixed!

Issue #11933, opened 2019-08-28
Intake Stats Page | Deprecate /intake/stats


pkarman:
The /intake/stats page reports on RAMP which is now deprecated.

Mark the route as deprecated and return a 404 and see who complains. There are a very low number of hits according on Google Analytics.

pkarman:
Code is merged and deployed. Starting the countdown timer for when we can remove the code.

If we don't hear anything in 60 days, let's delete.

PR #11934, opened 2019-08-28
Deprecate intake stats manager

pkarman:
connects #11933 

### Description
The 2 routes /intake/stats and /intake/manager contain code that is not actively used. Mark them as deprecated (commented out) for now, to be removed after we confirm there are no active users relying on them.


app/controllers/intake_stats_controller.rb
```diff
@@ -7,6 +7,11 @@ class IntakeStatsController < ApplicationController
   before_action :verify_access
 
   def show
+    # deprecated 2019/08/28
+    # either remove this controller entirely or update it to reflect current AMA models.
+    render "errors/404", layout: "application", status: :not_found
```
aroltsch: It would be nice to have a page that show deprecated message instead of 404. 404 might be confusing to users. This is okay as well.

pkarman: this was the quick and dirty approach. I agree a more formal deprecation page would be nice. Maybe next time?

I think the reality is that no one uses it and we can just delete eventually.

Issue #11836, opened 2019-08-20
EP 930s | Show previous decision ratings


jimruggiero:
###Scenario: a user identified a decision was incorrectly made, for example, they granted an issue instead of denying it.

AC:
1. Contestable issue generator: add all decision reviews (display all issues) - finalize_decision_issues_before_receipt_date method should also return decisions issues for that review

Assumptions:
1. Find all and add to list of contestable issues (existing add issues modal)

leikkisa:
Look in `contestable issue generator`

Add the current claim review's decision issues to the list of contestable issues

Currently it only includes these:
`finalized_decision_issues_before_receipt_date`

We want that to include something more like this:
`finalized_decision_issues_before_receipt_date + claim_review_decision_issues`

where the claim_review_decision_issues would equal `review.decision_issues`

jimruggiero:
AHHH

PR #11937, opened 2019-08-28
Return All Decision Issues, Regardless of Receipt Date

jcq:
Connects #11836 

### Description
Previously, we filtered out decision issues where the approx_decision_date < receipt_date. This removes that filter and includes all relevant decision issues, regardless of date.

### Acceptance Criteria
- [ ] Code compiles correctly
- [ ] Decision issues after the receipt date show up in the list of contestable issues


app/models/claim_review.rb
```diff
@@ -63,6 +63,11 @@ def caseflow_only_edit_issues_url
     "/#{self.class.to_s.underscore.pluralize}/#{uuid}/edit"
   end
 
+  def finalized_decision_issues
+    DecisionIssue.where(participant_id: veteran.participant_id, benefit_type: benefit_type)
```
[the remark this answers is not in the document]
jcq: Alrighty, I've updated code to combine `review.decision_issues` with the results of `finalized_decision_issues_before_receipt_date`. Let me know if that logic seems OK.

app/workflows/contestable_issue_generator.rb
```diff
@@ -41,7 +41,8 @@ def from_ratings
   end
 
   def from_decision_issues
-    @from_decision_issues ||= finalized_decision_issues_before_receipt_date.map do |decision_issue|
+    issues = finalized_decision_issues_before_receipt_date | review.decision_issues
```
aroltsch: Two things I am thinking about: 
1. I think we'd only like to include finalized decision issues for the review
2. We'd need to make sure this logic is hidden behind the feature flag: `correct_claim_reviews`, aka we should only display decision issues after the receipt_date if correct_claim_reviews is enabled for the current user

jcq: Good points both. Pushing updates now.

app/workflows/contestable_issue_generator.rb
```diff
@@ -41,7 +41,14 @@ def from_ratings
   end
 
   def from_decision_issues
-    @from_decision_issues ||= finalized_decision_issues_before_receipt_date.map do |decision_issue|
+    issues = finalized_decision_issues_before_receipt_date
+
+    # If correction support enabled, we include all decision issues for the issue
+    if FeatureToggle.enabled?(:correct_claim_reviews, user: RequestStore[:current_user])
+      issues |= review.decision_issues.select(&:finalized?)
```
leikkisa: This should be a +=, so it would include the regular decision issues from before the receipt date, plus decision issues connected to the review

spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -1168,7 +1168,39 @@ def complete_higher_level_review
...
+          expect(page).to_not have_content("Left knee granted")
+        end
+      end
+
+      context "correct_claim_reviews feature toggle enabled" do
+        before { FeatureToggle.enable!(:correct_claim_reviews) }
+        scenario "Contestable issues include future finalized decision issues" do
+          hlr, = start_higher_level_review(veteran, is_comp: false)
+          create(:decision_issue,
+                 decision_review: hlr,
```
leikkisa: This would be from another decision review since it was decided before this decision review existed.

spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -1168,7 +1168,39 @@ def complete_higher_level_review
...
+                 caseflow_decision_date: receipt_date + 1.day,
+                 benefit_type: hlr.benefit_type,
+                 decision_text: "something was decided in the future",
+                 participant_id: veteran.participant_id)
+
+          visit "/intake/add_issues"
+          click_intake_add_issue
+
+          expect(page).to have_content("something was decided in the past")
```
leikkisa: If you didn't fix the |=, but fixed the decision review to be from a prior decision review, this part would fail.

spec/models/decision_review_spec.rb
```diff
@@ -205,11 +216,26 @@ def find_serialized_issue(serialized_contestable_issues, ref_id_or_description)
...
+        end
+      end
+
+      context "with correct_claim_reviews feature toggle" do
+        before { FeatureToggle.enable!(:correct_claim_reviews) }
+
+        it "does include decision issues in the future that correspond to same review" do
+          expect(subject.map(&:serialize)).to include(hash_including(description: "decision issue 3"))
+          expect(subject.map(&:serialize)).to include(hash_including(description: "future decision issue 1"))
+          expect(subject.map(&:serialize)).to_not include(hash_including(description: "future decision issue 2"))
```
leikkisa: Maybe in the description you can say "from another review" or "from the same review" to make it clearer what's happening

spec/models/appeal_spec.rb
```diff
@@ -171,10 +171,10 @@
              end_product_last_action_date: receipt_date - 1.day)
     end
 
-    it "does not return Decision Issues in the future" do
-      expect(subject.count).to eq(1)
-      expect(subject.first.decision_issue.id).to eq(past_decision_issue.id)
-    end
+    # it "does not return Decision Issues in the future" do
```
leikkisa: I suppose we can leave this off for now, we prevent edit on outcoded appeals and they shouldn't be able to navigate to the edit page.

app/workflows/contestable_issue_generator.rb
jcq: Oh! I did the `|=` in order to combine without duplicates. Wouldn't `+=` run the risk of having multiple of the same issue?

jcq: Just switching to concatenating followed by `.uniq` to be clear as to what we're doing

app/workflows/contestable_issue_generator.rb
```diff
@@ -37,7 +37,14 @@ def from_ratings
   end
 
   def contestable_decision_issues
-    @contestable_decision_issues ||= finalized_decision_issues_before_receipt_date.map do |decision_issue|
+    issues = finalized_decision_issues_before_receipt_date
+
+    # If correction support enabled, we include all decision issues for the issue
```
leikkisa: It's actually "all decision issues for the review"

spec/feature/intake/end_product_correction_spec.rb
```diff
@@ -107,6 +107,56 @@
...
+          click_intake_add_issue
+          expect(page).to have_content("Left knee granted")
+          expect(page).to have_content("something was decided in the past for past claim review")
+          expect(page).to have_content("something was decided in the future for this claim review")
+        end
+
+        it "does not show decision issues from future from past claim review" do
+          visit edit_path
+          click_intake_add_issue
+          expect(page).to_not have_content("something was decided in the future for past claim review")
```
leikkisa: I think it makes more sense to put this line in the above "it" statement, it'll be faster.

spec/models/appeal_spec.rb
```diff
@@ -171,10 +171,10 @@
...
+    #   expect(subject.count).to eq(1)
+    #   expect(subject.first.decision_issue.id).to eq(past_decision_issue.id)
+    # end
```
leikkisa: This can be uncommented... hopefully this test still passes since the feature toggle isn't on!

PR #11940, opened 2019-08-29
Use queue config filter options to populate filter dropdowns

hschallhorn:
Prep for #11309

### Description
Use the queue config filter options for columns on the front end

### Acceptance Criteria
- [ ] Filter dropdowns use filter values from the back end rather than calculated ones on the front end

### Testing Plan
1. TODO when master stops bugging out


spec/feature/queue/task_queue_spec.rb
```diff
@@ -440,46 +440,90 @@ def validate_pulac_cerullo_tasks_created(task_class, label)
       User.authenticate!(user: organization_user)
       create_list(:generic_task, unassigned_count, :in_progress, assigned_to: organization)
       create_list(:generic_task, assigned_count, :on_hold, assigned_to: organization)
-      visit(organization.path)
     end
 
-    it "shows the right organization name" do
-      expect(page).to have_content(organization.name)
-    end
+    context "when not using pagination" do
```
hschallhorn: highly recommended ignoring white space changes to look at this

client/app/queue/OrganizationQueue.jsx
```diff
@@ -56,29 +56,30 @@ class OrganizationQueue extends React.PureComponent {
     return config;
   }
 
-  createColumnObject = (column, config, tasks) => {
+  filterValuesForColumn = (columnObj) => columnObj && columnObj.filterable && columnObj.filter_options;
```
lowellrex: Can we pass `config` as an argument here as well so we can add `config.use_task_pages_api` to this condition? I think we still want to hide the filters from organization queues using the task pages API because clicking the filter options will not actually filter the rows since we have not finished that work.

client/app/queue/OrganizationQueue.jsx
```diff
@@ -56,29 +56,30 @@ class OrganizationQueue extends React.PureComponent {
...
+    return functionForColumn[columnName];
   }
 
   columnsFromConfig = (tabConfig, tasks) => {
     return tabConfig.columns.map((column) => {
-      const columnName = typeof (column) === 'string' ? column : column.name;
-
-      return this.createColumnObject(columnName, tabConfig, tasks);
+      return typeof (column) === 'string' ? this.createColumnObject(column, null, tabConfig, tasks) :
+        this.createColumnObject(column.name, column, tabConfig, tasks);
```
lowellrex: We added this ternary condition in #11871 in order to be as non-disruptive as possible, but now since the back-end is returning an object we should be able to remove it. If we do that we can leave the function signature the same and just expect an object instead of a string. Could look something like the following:

```jsx
columnsFromConfig = (tabConfig, tasks) =>
  tabConfig.columns.map((column) => this.createColumnObject(column, tabConfig, tasks));
```

And the variables `columnObj` and `columnName` in `createColumnObject()` would become `column` and `column.name` respectively.

client/app/components/TableFilter.jsx
```diff
@@ -59,13 +59,22 @@ class TableFilter extends React.PureComponent {
   }
 
   filterDropdownOptions = (tableDataByRow, columnName) => {
-    const { customFilterLabels, enableFilterTextTransform } = this.props;
+    const { customFilterLabels, enableFilterTextTransform, manualFilterOptions } = this.props;
+    const filtersForColumn = _.get(this.props.filteredByList, columnName);
+
+    if (manualFilterOptions && manualFilterOptions.length) {
```
lowellrex: What do you think of renaming this variable to something like `precalculatedFilterOptions` or `filterOptionsFromApi`?

client/app/queue/QueueTable.jsx
```diff
@@ -98,6 +95,13 @@ const HeaderRow = (props) => {
             valueTransform={column.filterValueTransform}
             updateFilters={(newFilters) => props.updateFilteredByList(newFilters)}
             filteredByList={props.filteredByList} />;
+        } else if (props.useTaskPagesApi && column.filterOptions) {
+          filterIcon = <TableFilter
+            {...column}
+            tableData={column.tableData || props.rowObjects}
+            filterOptionsFromApi={props.useTaskPagesApi && column.filterOptions}
+            updateFilters={(newFilters) => props.updateFilteredByList(newFilters)}
+            filteredByList={props.filteredByList} />;
```
lowellrex: Can we hide the filter icon by commenting this block out until we update `updateFilteredByList()` to correctly make requests to the back-end for the task pages API? Right now clicking the dropdown if we are using the task pages does not filter the list.

Issue #11622, opened 2019-08-01
Refactor intake issue.js file


aroltsch:
In Intake issue.js, there are 3 very similar methods: formatRequestIssues, formatIssues and formatAddedIssues and they are all called on the request issue edit page. Investigate how these methods can be combined so we don't repeat the same logic of assigning the same attributes. Also, some of the logic separating rating, nonrating and unidentified could probably be combined too. Create better names.

PR #11943, opened 2019-08-29
Refactor intake issue file

Sjones352:
Connects #11622
Still a WIP.

### DIFF
--- a/client/app/intake/util/issues.js
+++ b/client/app/intake/util/issues.js
@@ -112,13 +112,11 @@ export const formatRequestIssues = (requestIssues, contestableIssues) => {
   return requestIssues.map((issue) => {
     const issueDate = new Date(issue.rating_issue_profile_date);
 
-    // Nonrating issues
     if (issue) {
       return {
         id: String(issue.id),
         isRating: false,
         benefitType: issue.benefit_type,
-        category: issue.category,
         decisionIssueId: issue.contested_decision_issue_id,
         description: issue.description,
         decisionDate: formatDateStr(issue.approx_decision_date),
@@ -135,7 +133,6 @@ export const formatRequestIssues = (requestIssues, contestableIssues) => {
         endProductCode: issue.end_product_code,
         withdrawalDate: formatDateStrUtc(issue.withdrawal_date),
         editable: issue.editable,
-        // new code
         isUnidentified: issue.is_unidentified,
         notes: issue.notes,
         index: contestableIssueIndexByRequestIssue(contestableIssues, issue),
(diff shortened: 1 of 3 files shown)


client/app/intake/util/issues.js
```diff
@@ -133,57 +131,22 @@ export const formatRequestIssues = (requestIssues, contestableIssues) => {
...
-        vacolsIssue: issue.vacols_issue,
-        endProductCleared: issue.end_product_cleared,
-        endProductCode: issue.end_product_code,
-        withdrawalDate: formatDateStrUtc(issue.withdrawal_date),
-        editable: issue.editable
+        notes: issue.notes,
+        category: issue.category,
+        index: contestableIssueIndexByRequestIssue(contestableIssues, issue),
+        isRating: !issue.category,
+        rating: !issue.category,
```
aroltsch: What is the difference between `rating` and `isRating`?

client/app/intake/util/issues.js
```diff
@@ -110,14 +110,12 @@ export const formatRequestIssues = (requestIssues, contestableIssues) => {
   }
 
   return requestIssues.map((issue) => {
+    const issueDate = new Date(issue.rating_issue_profile_date);
```
aroltsch: Do we want to include this inside the if statement? To parse the date only if the issue is present

client/app/intake/util/issues.js
```diff
@@ -110,14 +110,12 @@ export const formatRequestIssues = (requestIssues, contestableIssues) => {
...
 
-    // Nonrating issues
-    if (issue.category) {
+    if (issue) {
```
aroltsch: It is better to return if issue is not present instead of having a very big if block. 

e.g. 
```
if (!issue) { return; } 
```

client/app/intake/util/issues.js
leikkisa: It looks like `rating` is coming from the back end, so you probably don't want to set it using the issue category, since that will be missing for nonrating decision issues.  isRating was being set manually for new request issues that didn't have a category (because new nonrating issues have a category). Part of the reason the formatting is separated is due to the differences in the issue data before they become issues on the backend.  I tots agree it needs refactored, but we will need to be careful.

client/app/intake/util/issues.js
```diff
@@ -392,7 +354,8 @@ export const formatAddedIssues = (intakeData, useAmaActivationDate = false) => {
         endProductCleared: issue.endProductCleared,
         editedDescription: issue.editedDescription,
         correctionType: issue.correctionType,
-        editable: issue.editable
+        editable: issue.editable,
+        rating: !issue.category
```
leikkisa: this should be issue.rating

Sjones352: removed rating

client/app/intake/util/issues.js
Sjones352: Fixed. Ready for another review

client/app/intake/util/issues.js
Sjones352: Fixed

client/app/intake/util/issues.js
Sjones352: Removed rating

PR #11944, opened 2019-08-29
Async jobs page style

pkarman:
connects #11907 

### Description
Improves UX of the /jobs page:

* fits content better horizontally on the /jobs page, reducing scroll and table overflow
* adds Restart button to details job page
* changes Job Type filter to a dropdown button instead of links (better horizontal fit)
* shows only the exception name on the /jobs page, full error stack on the details page

Screen Shot 2019-08-29 at 1 01 33 PM

Screen Shot 2019-08-30 at 9 27 50 AM

### DIFF
--- a/client/app/asyncableJobs/components/AsyncModelNav.jsx
+++ b/client/app/asyncableJobs/components/AsyncModelNav.jsx
@@ -1,21 +1,33 @@
 import React from 'react';
 import moment from 'moment';
+import PropTypes from 'prop-types';
+import DropdownButton from '../../components/DropdownButton';
 
 const DATE_TIME_FORMAT = 'ddd MMM DD HH:mm:ss YYYY';
 
 export default class AsyncModelNav extends React.PureComponent {
   modelNameLinks = () => {
-    let links = [];
+    const links = [];
+    const modelNames = this.props.models.sort();
+    const numLinks = modelNames.length;
 
-    for (let modelName of this.props.models.sort()) {
-      let modelLink = <span key={modelName} className="cf-model-jobs-link">
-        <a href={`/asyncable_jobs/${modelName}/jobs`}>{modelName}</a>
-      </span>;
+    for (let modelName of modelNames) {
+      const url = `/asyncable_jobs/${modelName}/jobs`;
+      let modelLink;
+
+      if (numLinks > 4) {
+        modelLink = {
+          title: modelName,
+          target: url
+        };
+      } else {
+        modelLink = <span key={modelName} className="cf-model-jobs-link"><a href={url}>{modelName}</a></span>;
+      }
 
       links.push(modelLink);
     }
 
-    return links;
+    return numLinks > 4 ? <DropdownButton lists={links} label="Job Type" /> : links;
   }
 
   render = () => {
@@ -23,8 +35,13 @@ export default class AsyncModelNav extends React.PureComponent {
     return <div>
       <strong>Last updated:</strong> {moment(this.props.fetchedAt).format(DATE_TIME_FORMAT)}
       &nbsp;&#183;&nbsp;
-      <a href="/jobs">All jobs</a>
+      <a href="/jobs" className="cf-link-btn">All jobs</a>
       <div>{this.modelNameLinks()}</div>
     </div>;
   }
 }
+
+AsyncModelNav.propTypes = {
+  models: PropTypes.array,
+  fetchedAt: PropTypes.string
+};
--- a/client/app/asyncableJobs/components/AsyncModelNav.jsx
+++ b/client/app/asyncableJobs/components/AsyncModelNav.jsx
@@ -27,16 +27,17 @@ export default class AsyncModelNav extends React.PureComponent {
       links.push(modelLink);
     }
 
-    return numLinks > 4 ? <DropdownButton lists={links} label="Job Type" /> : links;
+    return numLinks > 4 ? <DropdownButton lists={links} label="Filter by Job Type" /> : links;
   }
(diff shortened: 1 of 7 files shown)


client/app/asyncableJobs/components/AsyncModelNav.jsx
```diff
@@ -1,30 +1,48 @@
...
   render = () => {
 
     return <div>
       <strong>Last updated:</strong> {moment(this.props.fetchedAt).format(DATE_TIME_FORMAT)}
-      &nbsp;&#183;&nbsp;
-      <a href="/jobs">All jobs</a>
-      <div>{this.modelNameLinks()}</div>
+      <div>
+        {this.modelNameLinks()}
+        <a style={{ float: 'right' }} href="/jobs" className="cf-link-btn">All jobs</a>
```
jcq: This seems a bit weird, from a UX perspective. This is effectively the "back" button on a detail page, but here it's used to remove the filter. Obviously this might be something that can wait until a designer has time to take a crack at it, but it might be clearer to the user to simply have an "All" (effectively blank) option in the "Filter by Job Type" dropdown.

client/app/asyncableJobs/pages/JobPage.jsx
```diff
@@ -61,6 +70,10 @@ class AsyncableJobPage extends React.PureComponent {
             <th>User</th>
             <td><a href={`/intake/manager?user_css_id=${job.user}`}>{job.user}</a></td>
           </tr>
+          <tr>
+            <th></th>
+            <td><JobRestartButton job={job} page={this} /></td>
```
jcq: This looks a bit odd, given the vertical orientation of the table. How about in this case removing the <th> and having colspan="2" on the <td>?  We'd probably want to center or right-align the content in that case. 🤷‍♂ 

This is really a question for a designer I suppose — functionality is what matters on a screen like this, and  it seems like that's working.

client/app/asyncableJobs/components/AsyncModelNav.jsx
jcq: For some reason I just noticed that the filter dropdown is present on the detail page as well. Perhaps we could hide the AsyncModelNav on the detail page and simply have a "< Back" sort of link that takes user back to the full list?

Again, totally fine to wait for a designer pass.

client/app/asyncableJobs/components/JobRestartButton.jsx
```diff
@@ -0,0 +1,120 @@
...
+    this.sendRestart();
+  }
+
+  sendRestart = () => {
+    const button = this;
+    const page = this.props.page;
+    const job = this.props.job;
+    const url = `/asyncable_jobs/${job.klass}/jobs/${job.id}`;
+
+    ApiUtil.patch(url, {}).
```
jcq: Do you mean to have no payload here? Are you just interpreting any PATCH to this endpoint as a restart?

client/app/asyncableJobs/components/JobRestartButton.jsx
```diff
@@ -0,0 +1,120 @@
...
+  restartJob = () => {
+    this.setState({ restarting: true });
+    this.sendRestart();
+  }
+
+  sendRestart = () => {
+    const button = this;
+    const page = this.props.page;
+    const job = this.props.job;
+    const url = `/asyncable_jobs/${job.klass}/jobs/${job.id}`;
```
jcq: I have to know... what's the origin of "klass"?

client/app/asyncableJobs/components/JobRestartButton.jsx
```diff
@@ -0,0 +1,120 @@
...
+            restarting: false
+          });
+
+          // trigger table rerender
+          const jobsRestarted = page.state.restarted + 1;
+
+          page.setState({ restarted: jobsRestarted });
+        },
+        (error) => {
+          throw error;
```
jcq: Why throw and then immediately catch the error?

client/app/asyncableJobs/components/JobRestartButton.jsx
pkarman: I've been leaving it blank because the only action right now is restart. There's also the potential of adding a `cancel` action, in which case we'd want to have a payload to indicate that. Whether the business agrees remains TBD.

For context: the /jobs page was intended as a stopgap measure at AMA day, and is only used by a handful of people at present. The reason it's getting attention now is that we have heard from business that they need more visibility into stuck jobs. Priority high, our design resources low.

client/app/asyncableJobs/components/JobRestartButton.jsx
pkarman: `class` is a reserved word in Ruby, and I think we carried over the convention is JS.

client/app/asyncableJobs/components/JobRestartButton.jsx
pkarman: I think originally we were logging the error too. I can remove this now.

client/app/asyncableJobs/components/AsyncModelNav.jsx
pkarman: We will be linking directly to the detail page from the /intake space in future, so there won't be a sense of "back".

agreed about the nav.

client/app/asyncableJobs/pages/JobPage.jsx
pkarman: I moved the button outside the table. You're right, it was odd placement.

Screenshot updated in PR description.

Issue #11882, opened 2019-08-23
Investigate Hearings Management Queue Results


pshahVA:
## Hearings management queue may be limiting results

- The hearings management queue is only returning about 200 tasks, but there were thousands prior to temporarily limiting the volume for performance reason.

- Some engineering work has been done to close admin tasks that are corrected, but we still need to check if all tasks are showing up.

pkarman:
There is not a numerical limit. There is a filter on AOD tasks:

https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/organizations/tasks_controller.rb#L24

pshahVA:
Thanks.  @lomaxap said he updated the code so that many tasks may have been closed,  we may be able to remove this filter now.  He said he will look.

rubaiyat22:
@lomaxap @pshahVA 

I did a benchmark test on retrieving all tasks of a current user of Hearing Management organization.
``` 
RequestStore[:current_user] = User.system_user
org = HearingsManagement.singleton
timing = Benchmark.measure { GenericQueue.new(user: org).tasks }
```
Task count: 5546

Run 1: 
```
=> #<Benchmark::Tms:0x000055663a2d8598 @label="", @real=18.471204397999827, @cstime=0.0, @cutime=0.0, @stime=0.41777, @utime=12.132155000000001, @total=12.549925000000002>
```
Run 2:
```
=> #<Benchmark::Tms:0x00005566340b9790 @label="", @real=17.060635664000074, @cstime=0.0, @cutime=0.0, @stime=0.40796600000000005, @utime=11.025085, @total=11.433051>
```
Run 3:
```
=> #<Benchmark::Tms:0x000055663555bc48 @label="", @real=18.687685814999895, @cstime=0.0, @cutime=0.0, @stime=0.33956600000000003, @utime=12.767992, @total=13.107558>
```
Real value is the elapsed time in minutes.

On average of **3** runs, **5546** tasks took **18.0731752923** minutes.

pshahVA:
Does this mean that the volume coming back on the UI is too small?

ferristseng:
@pshahVA Are you talking about with the fix in place? I think we're returning 1000 right now. Do you think we should be returning more?

We were trying to judge how removing the fix would affect prod performance. It looks like there might still be timeout issues if we do revert it.

pshahVA:
[...]

PR #11945, opened 2019-08-29
Find active child TimedHoldTask without New Query

lomaxap:
Issue surfaced in #11882, `children.open` runs a new query. For `GenericQueues` this results in an an n+1 query. Filter open child tasks in find block  instead of using the `open` method


tomas-nava:
😢 the test failures in circle are happening for me locally as well; I'm not sure why, as the method returns identical results during these tests both before and after your change. Maybe the original query causes cached objects to reload?

lomaxap:
> Maybe the original query causes cached objects to reload?

I added a bunch of reload statements to the failing tests and this resolved the issue. This method effectively ensured tasks were always up-to-date.

ferristseng:
Is there any documentation about what `open` is supposed to do? I couldn't find any searching around.

lomaxap:
> Is there any documentation about what `open` is supposed to do? I couldn't find any searching around.

I don't think there's any documentation on it, but `open` is a [scope method](https://api.rubyonrails.org/classes/ActiveRecord/Scoping/Named/ClassMethods.html) that [limits tasks](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/task.rb#L44) to those with open statuses. So instead of having to write `Task.where(status: open_statuses)` all the time, you can just write `Task.open`.

PR #11947, opened 2019-08-30
Regional Office Timezone for DC No Longer Blank

### LINKED ISSUE #11796: Hearing API regional office and timezone blank for DC
Regional Office and Time Zone are always blank for hearings that are in DC.

```
HearingDay.where(regional_office: nil).count
=> 317
```

## AC

* [ ] Hearing API should always have a timezone and regional office value

aplybeah:
Resolves #11796 

Updated methods in `hearing_serializer` and `legacy_hearing` so that the data specific to the Central office can be accessed from the hash.


app/models/hearing.rb
```diff
@@ -157,15 +156,25 @@ def worksheet_issues
...
+                          end
+  end
+
+  def regional_office_key
+    hearing_day_regional_office || "C"
+  end
 
   def regional_office_name
-    RegionalOffice::CITIES[regional_office_key][:label] unless regional_office_key.nil?
+    regional_office.to_h[:label]
```
pkarman: could we make these `.to_h` calls into actual methods? so we could call `regional_office.name` and `regional_office.timezone` and use `:delegate` instead?

Issue #11927, opened 2019-08-28
Claim review establishment stats


pkarman:
AMO would like to know some statistics about mean time to establish SC, HLR, RequestIssuesUpdate async jobs.

pkarman:
```csv
type,total,cancelled,processed,median,avg,max,min
supplemental_claims,115833,7,115826,00:00:16,00:37:30,1710:43:22,00:00:00
higher_level_reviews,24065,3,24062,00:00:09,01:21:06,2643:32:36,00:00:00
request_issues_updates,16372,180,16194,00:00:12,00:14:38,791:54:59,00:00:00
```

PR #11955, opened 2019-08-30
Optimize run time by avoiding multiple unnecessary SQL calls

pkarman:
connects #11927 

The async hash loads lots of unneeded associations, so avoid it for these large data sets.


pkarman:
merged in #11956

PR #11959, opened 2019-09-03
Change Fetch Hearing Locations Query to Get Distinct Appeals

lomaxap:
joining `available_hearing_locations` results in duplicate appeal records being returned for query. This means that the same appeal being geomatched multiple times in a single run of this job. 
- Changes query to aggregate hearing location updated_at dates, and group by appeal id

### DIFF
--- a/app/jobs/fetch_hearing_locations_for_veterans_job.rb
+++ b/app/jobs/fetch_hearing_locations_for_veterans_job.rb
@@ -33,6 +33,7 @@ class FetchHearingLocationsForVeteransJob < ApplicationJob
 
     appeals.each do |appeal|
       begin
+        appeal.reload # we only selected id and ahl_update_at, reload all columns
         geomatch_result = geomatch(appeal)
         record_geomatched_appeal(appeal.external_id, geomatch_result[:status])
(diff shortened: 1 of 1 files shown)


app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -15,9 +15,11 @@ def find_appeals_ready_for_geomatching(appeal_type)
     ).pluck(:appeal_id)
 
     appeal_type.left_outer_joins(:available_hearing_locations)
+      .select(:id)
+      .select("MIN(available_hearing_locations.updated_at) as ahl_updated_at")
       .where(id: appeal_ids)
-      .order("available_hearing_locations.updated_at nulls first")
-      .limit(QUERY_LIMIT)
+      .group(:id)
+      .order("ahl_updated_at nulls first")
```
ferristseng: Is this line here to order it so the appeals without ahls are processed first?

lomaxap: yes. it also orders the oldest updated appeals first.

PR #11962, opened 2019-09-03
Anya/11838 correct decision issues with dta supp claims

### LINKED ISSUE #11838: EP 930s | Remove incorrectly added issues from DTA SC
**Scenario: a user identified a decision was incorrectly made. For example, they DTA'd an issue instead of denying it. The DTA Supplemental Claim EP was not finalized and decided.**

image

## AC
- When EP 930 on the original claim is established, find the associated DTA Supplemental Claim. Close the incorrectly added request issue from that DTA supplemental claim. 
   - Suggest a new `closed_status`
- Delete contentions in VBMS 
- If all request issues are removed from supplemental claim, cancel EP
   - There's already code in place for this, but ensure it's working for this

aroltsch:
Resolves #11838 

### Description
Remove incorrectly added DTA request issues


app/models/request_issue.rb
```diff
@@ -912,7 +918,7 @@ def check_for_active_request_issue_by_rating!
 
   def check_for_active_request_issue_by_decision_issue!
     return unless contested_decision_issue_id
-
+    return if correction?
```
aroltsch: Do you think this is correct, Sally?

app/workflows/contestable_issue_generator.rb
```diff
@@ -41,7 +41,7 @@ def from_ratings
   end
 
   def from_decision_issues
-    @from_decision_issues ||= finalized_decision_issues_before_receipt_date.map do |decision_issue|
+    @from_decision_issues ||= (finalized_decision_issues_before_receipt_date | review.decision_issues).map do |decision_issue|
```
aroltsch: Ignore this code. This will be removed. JC is working on it.

app/models/decision_issue.rb
```diff
@@ -19,7 +19,7 @@ class DecisionIssue < ApplicationRecord
   has_many :remand_reasons, dependent: :destroy
   belongs_to :decision_review, polymorphic: true
   has_one :effectuation, class_name: "BoardGrantEffectuation", foreign_key: :granted_decision_issue_id
-  has_one :contesting_request_issue, class_name: "RequestIssue", foreign_key: "contested_decision_issue_id"
+  has_many :contesting_request_issues, class_name: "RequestIssue", foreign_key: "contested_decision_issue_id"
```
aroltsch: I don't think it was used anywhere.

app/models/request_issue_closure.rb
```diff
@@ -18,6 +18,14 @@ def with_no_decision!
     end
   end
 
+  def remove_issue_with_corrected_decision!
```
leikkisa: I decided to move some logic to RequestIssueClosure to ensure that there is no attempt to remove the contention and cancel the EP if the request issue was already closed. Also Moncef thought this would be a good workflow for request issue closures, but we didn't move them all here from the get go.

app/workflows/request_issue_closure.rb
```diff
@@ -18,6 +18,14 @@ def with_no_decision!
     end
   end
 
+  def remove_issue_with_corrected_decision!
```
leikkisa: I decided to move some logic to RequestIssueClosure to ensure that there is no attempt to remove the contention and cancel the EP if the request issue was already closed. Also Moncef thought this would be a good workflow for request issue closures, but we didn't move them all here from the get go.

app/workflows/request_issue_closure.rb
```diff
@@ -18,6 +18,14 @@ def with_no_decision!
...
+    close!(status: :removed) do
+      canceled!
+      RequestIssueContention.new(request_issue).remove!
+      request_issue.end_product_establishment&.cancel_unused_end_product!
+    end
+  end
+
```
leikkisa: I decided to move some logic to RequestIssueClosure to ensure that there is no attempt to remove the contention and cancel the EP if the request issue was already closed. Also Moncef thought this would be a good workflow for request issue closures, but we didn't move them all here from the get go.

PR #11964, opened 2019-09-03
Nil check for tasks on change type

hschallhorn:
### Description
Legacy colocated org tasks do not always have parents
```
LegacyAppeal 198515
├── RootTask
└── OtherColocatedTask
    └── OtherColocatedTask
        └── TimedHoldTask
```
so the[ check for task type equality ](https://github.com/department-of-veterans-affairs/caseflow/blob/15bdbfaa4b8f2c7c31354d260c2640907d020c95/app/models/task.rb#L221)fails with a NoMethodError.

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -227,7 +227,7 @@ class Task < ApplicationRecord
   end
 
   def same_task_type?(task_to_check)
-    type.eql?(task_to_check.type)
+    type.eql?(task_to_check&.type)
   end
 
   def cancel_descendants


app/models/task.rb
```diff
@@ -227,7 +227,7 @@ def last_descendant_of_type
   end
 
   def same_task_type?(task_to_check)
-    type.eql?(task_to_check.type)
+    type.eql?(task_to_check&.type)
```
pkarman: if `task_to_check` is `nil` that would happen most often when there is no parent assigned?

if the `type` is also `nil` then that means this method returns `true`?

I guess w/o a test it's hard to tell why this might be happening. Is it ok that `task_to_check` is nil?

hschallhorn: This is happening because legacy colocated org tasks don't have parents. If `task_to_check` is nil here, `same_task_type?` will return false, which is what we want. When I wrote this I was unaware that legacy colocated tasks may not have parents.

The type of a task should never be nil. 
```ruby
> Task.where(type: nil).count
=> 0
```

PR #11977, opened 2019-09-04
Allow AVLJs to perform judge tasks on legacy appeals

### LINKED ISSUE #11971: Allow AVLJs to be assigned legacy cases
Started from [this slack thread](https://dsva.slack.com/archives/CHX8FMP28/p1567534420110500). 

For attorneys and judges, we determine what type of task (and therefor what actions are available to them) for a legacy appeal assigned to them [based on their vacols role](https://github.com/department-of-veterans-affairs/caseflow/blob/a86c1e04e52faf6911b99281908576790fc692df/app/models/queues/legacy_work_queue.rb#L30). This causes an issue for acting judges as their first role is always "attorney" and they do not receive the ability to dispatch a drafted appeal.

This ticket is to determine and implement how we decide which actions an acting judge sees.

### Expand conditional to include judge user

By using `user&.vacols_roles&.include? "judge"  we would select all judges and acting judges. However, this would display a judge task and actions to every acting judge that is assigned a legacy task, regardless if it was assigned intending to be a judge task or an attorney task (essentially the opposite of what happens now).

### Using JudgeTeam ownership to determine acting judges

We do this in [various other places](https://github.com/department-of-veterans-affairs/caseflow/search?q=JudgeTeam.for_judge&unscoped_q=JudgeTeam.for_judge) in the code where we want to allow AVLJs to have judge functionality. This would require us to keep the JudgeTeams for acting judges up to date, but would allow us to have some control over who is currently working on drafting a decision vs evaluating one.

### Using issues with decided dispositions

An appeal where all issues have non nil dispositions for their issues is guaranteed to have gone through attorney checkout. This would cause some weirdness when adding issues after some of them have had dispositions added.

### Using appeals with decision documents

An appeal with a decision document is guaranteed to have gone through attorney checkout. This would break as soon as a judge sends the case back to the attorney.

## Actually creating the tasks and assigning them to users as we do with ama appeals

Not exactly a viable solution for the AVLJs working AOJ cases who needed a solution to this
around yesterday.

## Thoughts

This was mostly to get some things from my mind before stepping away for the day. Will revisit in the morning for feedback and ideas from engineering, support, and product.

hschallhorn:
Resolves #11971

### Description
Due to the way we determine [task type for legacy work queues](https://github.com/department-of-veterans-affairs/caseflow/blob/a86c1e04e52faf6911b99281908576790fc692df/app/models/queues/legacy_work_queue.rb#L30), acting judges have been receiving judge tasks, but cannot act upon them. This PR classes any appeal with an associated AttorneyCaseReview assigned to someone with a judge vacols_role as a JudgeTask, allowing AVLJs to review legacy cases.

### DIFF
--- a/app/controllers/legacy_tasks_controller.rb
+++ b/app/controllers/legacy_tasks_controller.rb
@@ -26,7 +26,7 @@ class LegacyTasksController < ApplicationController
                               name: "LegacyTasksController.index") do
           tasks = LegacyWorkQueue.tasks_for_user(user)
           render json: {
-            tasks: json_tasks(tasks, current_role)
+            tasks: json_tasks(tasks, user, current_role)
           }
         end
       end
@@ -98,7 +98,7 @@ class LegacyTasksController < ApplicationController
     ::WorkQueue::LegacyTaskSerializer.new(task)
   end
 
-  def json_tasks(tasks, role)
-    ::WorkQueue::LegacyTaskSerializer.new(tasks, is_collection: true, params: { role: role })
+  def json_tasks(tasks, user, role)
+    ::WorkQueue::LegacyTaskSerializer.new(tasks, is_collection: true, params: { user: user, role: role })
   end
 end
--- a/app/models/legacy_appeal.rb
+++ b/app/models/legacy_appeal.rb
@@ -734,11 +734,7 @@ class LegacyAppeal < ApplicationRecord
   end
 
   def attorney_case_review
-    AttorneyCaseReview.find_by(task_id: "#{vacols_id}-#{vacols_case_review_creation_date_in_string_format}")
-  end
-
-  def vacols_case_review_creation_date_in_string_format
-    VacolsHelper.day_only_str(vacols_case_review.created_at)
+    AttorneyCaseReview.find_by(task_id: "#{vacols_id}-#{VacolsHelper.day_only_str(vacols_case_review.created_at)}")
   end
 
   def vacols_case_review
--- a/app/models/legacy_appeal.rb
+++ b/app/models/legacy_appeal.rb
@@ -733,6 +733,18 @@ class LegacyAppeal < ApplicationRecord
     file_type.eql? "Paper"
   end
 
+  def attorney_case_review
+    AttorneyCaseReview.find_by(task_id: "#{vacols_id}-#{vacols_case_review_creation_date_in_string_format}")
+  end
+
+  def vacols_case_review_creation_date_in_string_format
+    VacolsHelper.day_only_str(vacols_case_review.created_at)
+  end
+
+  def vacols_case_review
+    VACOLS::CaseAssignment.latest_task_for_appeal(vacols_id)
+  end
+
   private
 
   def soc_date_eligible_for_opt_in?(receipt_date)
--- a/app/models/legacy_tasks/judge_legacy_decision_review_task.rb
+++ b/app/models/legacy_tasks/judge_legacy_decision_review_task.rb
@@ -9,8 +9,8 @@ class JudgeLegacyDecisionReviewTask < JudgeLegacyTask
     end
   end
 
-  def available_actions(current_user, role)
-    return [] if role != "judge" || current_user != assigned_to
+  def available_actions(current_user, _role)
+    return [] if current_user.vacols_roles.none? { |vrole| vrole.casecmp("judge").zero? } || current_user != assigned_to
 
     [
       Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h,
--- a/app/models/legacy_tasks/judge_legacy_decision_review_task.rb
+++ b/app/models/legacy_tasks/judge_legacy_decision_review_task.rb
@@ -10,7 +10,7 @@ class JudgeLegacyDecisionReviewTask < JudgeLegacyTask
   end
 
   def available_actions(current_user, _role)
-    return [] if current_user.vacols_roles.none? { |vrole| vrole.casecmp("judge").zero? } || current_user != assigned_to
+    return [] if current_user != assigned_to || !current_user.judge_in_vacols?
 
     [
       Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h,
--- a/app/models/queues/legacy_work_queue.rb
+++ b/app/models/queues/legacy_work_queue.rb
@@ -27,7 +27,7 @@ class LegacyWorkQueue
       vacols_tasks.zip(vacols_appeals).map do |task, appeal|
         user = validate_or_create_user(user, task.assigned_to_css_id)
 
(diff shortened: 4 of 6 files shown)


app/models/queues/legacy_work_queue.rb
```diff
@@ -40,5 +40,13 @@ def validate_or_create_user(user, css_id)
         User.find_by_css_id_or_create_with_default_station_id(css_id)
       end
     end
+
+    def issues_have_dispositions(appeal)
+      appeal.issues.all? { |issue| !issue.disposition.nil? }
+    end
```
hschallhorn: If all of the issues for an appeal have a disposition, the decision has already been drafted by an attorney.  We are then assuming that this task has been assigned to an AVLJ for the purpose of them performing a judge task.

Is there a better metric to use for this?

app/controllers/legacy_tasks_controller.rb
```diff
@@ -98,7 +98,7 @@ def json_task(task)
     ::WorkQueue::LegacyTaskSerializer.new(task)
   end
 
-  def json_tasks(tasks, role)
-    ::WorkQueue::LegacyTaskSerializer.new(tasks, is_collection: true, params: { role: role })
+  def json_tasks(tasks, user, role)
+    ::WorkQueue::LegacyTaskSerializer.new(tasks, is_collection: true, params: { user: user, role: role })
```
hschallhorn: Ensures the current user gets passed to `task.available_actions`

app/models/queues/legacy_work_queue.rb
```diff
@@ -27,7 +27,7 @@ def tasks_from_vacols_tasks(vacols_tasks, user = nil)
       vacols_tasks.zip(vacols_appeals).map do |task, appeal|
         user = validate_or_create_user(user, task.assigned_to_css_id)
 
-        task_class = (user&.vacols_roles&.first&.downcase == "judge") ? JudgeLegacyTask : AttorneyLegacyTask
+        task_class = (user_is_judge(user) && issues_have_dispositions(appeal)) ? JudgeLegacyTask : AttorneyLegacyTask
```
leikkisa: Based on your comment below, should this be an || ?

app/models/queues/legacy_work_queue.rb
leikkisa: I think `issue.disposition.present?` would be better than the double negative.

app/models/queues/legacy_work_queue.rb
```diff
@@ -40,5 +40,13 @@ def validate_or_create_user(user, css_id)
         User.find_by_css_id_or_create_with_default_station_id(css_id)
       end
     end
+
+    def issues_have_dispositions(appeal)
+      appeal.issues.all? { |issue| !issue.disposition.nil? }
+    end
+
+    def user_is_judge(user)
+      user&.vacols_roles&.any? { |role| role.casecmp("judge").zero? }
```
leikkisa: I like that you're not just checking the first one now! Good bug prevention

app/models/queues/legacy_work_queue.rb
hschallhorn: So we're actually changing the way we classify a user as a judge here. Previously we were classifying all acting judges (who will have the vacols_roles ["attorney", "judge"]) as just attorneys. So here we're now checking that the user has a judge role.

But, as acting judges still can receive attorney tasks, we want to ensure we aren't incorrectly giving them judge tasks for decisions that haven't yet been drafted, so we want to make sure both of these conditionals are true.

app/models/legacy_tasks/judge_legacy_decision_review_task.rb
```diff
@@ -9,8 +9,8 @@ def review_action
     end
   end
 
-  def available_actions(current_user, role)
-    return [] if role != "judge" || current_user != assigned_to
+  def available_actions(current_user, _role)
+    return [] if current_user.vacols_roles.none? { |vrole| vrole.casecmp("judge").zero? } || current_user != assigned_to
```
pkarman: I would love to see this logic on the `User` model instead, encapsulated in a well-named method.

maybe something like `current_user.acts_like_vacols_judge?` or something?

app/models/queues/legacy_work_queue.rb
pkarman: this method seems like it belongs on the `appeal` class, not as a helper in this one.

app/models/queues/legacy_work_queue.rb
pkarman: again, this method could go on the `User` model.

app/models/legacy_tasks/judge_legacy_decision_review_task.rb
hschallhorn: 🤦‍♀  or the already existant `judge_in_vacols?`

app/models/legacy_appeal.rb
```diff
@@ -733,6 +733,18 @@ def paper_case?
...
+  end
+
+  def vacols_case_review_creation_date_in_string_format
+    VacolsHelper.day_only_str(vacols_case_review.created_at)
+  end
+
+  def vacols_case_review
+    VACOLS::CaseAssignment.latest_task_for_appeal(vacols_id)
+  end
+
```
hschallhorn: Moved relevant case review stuff out of update_legacy_attorney_case_review to here

app/models/queues/legacy_work_queue.rb
```diff
@@ -27,7 +27,7 @@ def tasks_from_vacols_tasks(vacols_tasks, user = nil)
       vacols_tasks.zip(vacols_appeals).map do |task, appeal|
         user = validate_or_create_user(user, task.assigned_to_css_id)
 
-        task_class = (user&.vacols_roles&.first&.downcase == "judge") ? JudgeLegacyTask : AttorneyLegacyTask
+        task_class = (user.judge_in_vacols? && appeal.attorney_case_review) ? JudgeLegacyTask : AttorneyLegacyTask
```
hschallhorn: Crux of the change. If a legacy case is assigned to an avlj in vacols and the case has already been reviewed by an attorney, we will consider this task to be a JudgeLegacyTask.

app/workflows/update_legacy_attorney_case_review.rb
```diff
@@ -30,38 +30,26 @@ def call
...
   def vacols_case_review_exists
     return if vacols_case_review
 
     errors.add(:document_id, "Could not find a legacy Attorney Case Review with id #{id}")
   end
 
   def vacols_case_review
-    @vacols_case_review ||= VACOLS::CaseAssignment.latest_task_for_appeal(id)
+    @vacols_case_review ||= appeal.vacols_case_review
   end
```
hschallhorn: Moved all this to legacy_appeal

PR #11979, opened 2019-09-04
Prefer veteran.file_number over vbms_id

pkarman:
## Description

LegacyAppeal assumes that `vbms_id` correlates with what BGS/VBMS think of as a file_number, but that isn't always true.

See https://dsva.slack.com/archives/CHX8FMP28/p1567615019135700

### DIFF
--- a/app/models/legacy_appeal.rb
+++ b/app/models/legacy_appeal.rb
@@ -230,7 +230,7 @@ class LegacyAppeal < ApplicationRecord
   end
 
   def veteran
-    @veteran ||= Veteran.find_or_create_by_file_number_or_ssn(veteran_file_number)
+    @veteran ||= Veteran.find_or_create_by_file_number_or_ssn(sanitized_vbms_id)
   end
 
   def veteran_ssn
--- a/app/models/legacy_appeal.rb
+++ b/app/models/legacy_appeal.rb
@@ -653,6 +653,9 @@ class LegacyAppeal < ApplicationRecord
   # Alias sanitized_vbms_id becauase file_number is the term used VBA wide for this veteran identifier
   def veteran_file_number
     vacols_file_number = sanitized_vbms_id
+
+    return vacols_file_number unless veteran
+
     caseflow_file_number = veteran.file_number
     if vacols_file_number != caseflow_file_number
       Raven.capture_message("legacy appeal #{external_id} has file_number mismatch with VACOLS and Caseflow")
--- a/app/models/legacy_appeal.rb
+++ b/app/models/legacy_appeal.rb
@@ -650,7 +650,10 @@ class LegacyAppeal < ApplicationRecord
     LegacyAppeal.veteran_file_number_from_bfcorlid vbms_id
   end
 
-  # Alias sanitized_vbms_id becauase file_number is the term used VBA wide for this veteran identifier
+  # The sanitized_vbms_id may be a SSN value, which may or may not be a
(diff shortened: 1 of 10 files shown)


app/models/legacy_appeal.rb
```diff
@@ -652,7 +652,15 @@ def sanitized_vbms_id
 
   # Alias sanitized_vbms_id becauase file_number is the term used VBA wide for this veteran identifier
```
lomky: Could we update this comment to match the new behavior?

spec/models/legacy_hearing_spec.rb
```diff
@@ -177,17 +177,15 @@
   context "#to_hash_for_worksheet" do
     subject { hearing.to_hash_for_worksheet(nil).with_indifferent_access }
 
+    let(:vbms_id) { "12345678" }
+    let!(:veteran) { create(:veteran, file_number: vbms_id) }
```
lomky: This line is used indirectly through the shared `creates` using the `vbms_id`, correct?

pkarman: yes. the important change here is to pre-create the `Veteran` record so that when `LegacyAppeal` calls `.veteran` the record exists and is not created on-the-fly, which will result in an attempt to re-use the file_number.

client/COPY.json
```diff
@@ -466,6 +466,9 @@
   "INTAKE_EDIT_WITHDRAW_DATE": "Please include the date the withdrawal was requested",
   "INTAKE_WITHDRAWN_BANNER": "This review will be withdrawn. You can intake these issues as a different type of decision review, if that was requested.",
   "INTAKE_RATING_MAY_BE_PROCESS": "Rating may be in progress",
+  "INTAKE_SEARCH_ERROR_INVALID_FILE_NUMBER": "Please enter a valid Veteran ID and try again.",
```
pkarman: just some clean up I noticed while fixing tests.

app/models/veteran.rb
```diff
@@ -271,7 +271,8 @@ def find_by_file_number_or_ssn(file_number_or_ssn, sync_name: false)
 
     def find_or_create_by_file_number_or_ssn(file_number_or_ssn, sync_name: false)
       if file_number_or_ssn.to_s.length == 9
-        find_or_create_by_ssn(file_number_or_ssn, sync_name: sync_name) ||
+        find_and_maybe_backfill_name(file_number_or_ssn, sync_name: sync_name) ||
```
lomky: What led to this behavior change here?

pkarman: if the existing `file_number` value is a SSN (9 digits) (which might happen and can be valid) then we were ignoring it and immediately looking for the value in the `ssn` column instead. This change let's us first find (but not create) by `file_number`, then if that is nil, find_or_create by `ssn`.

pkarman: I discovered this behavior because tests started to fail when `LegacyAppeal.veteran_file_number` called internally to `.veteran` with a 9 digit vbms_id.

lomky: Ah, so this function is perhaps more explicitly `find_by_file_number_and_maybe_backfill_name`? I feel like there's a possible refactor here, but up to you if you want to dig into that here, file an issue to look into it later, or pass!

pkarman: I thought the same thing! I will fix in a follow up PR.

PR #11981, opened 2019-09-04
move tests from daily_docket_spec.rb to daily_docket/{role} for each roles

rubaiyat22:
Resolves 11592

### Description
- move user related tests from daily_docket_spec to spec files corresponding with each role
- not ready to be merged, working on it right now


rubaiyat22:
Hi @tomas-nava, is this what you had in mind?

Issue #11851, opened 2019-08-21
EP 930s | Update current validation to ensure request issue corrections to not be treated as duplicates


jimruggiero:
AC:
If a request issue has a correction type, then it is a correction issue and we don't want to prevent a veteran from requesting that issue on an active review (so we don't want to consider it a duplicate). If it doesn't have a correction type, then it's already on an active review, and it is ineligible.

Assumptions:
1. Conflicts here will be handled outside of Caseflow on the VBMS process side.

leikkisa:
This starts with `check_for_active_request_issue_by_decision_issue` in `RequestIssue`

If a request issue has a correction_type, then it is a correction issue and we don't want to prevent a veteran from requesting that issue on an active review (so we don't want to consider it a duplicate).  If it doesn't have a correction type, then it's already on an active review, and it is ineligible.

Code change example:
`RequestIssue.active.find_by(contested_decision_issue_id: contested_decision_issue_id, correction_type: nil)`

aroltsch:
If we have the following scenario: 
Request issue A gets a remanded decision issue which in turn creates a DTA supplemental claim with a new request issue B
then we'd like to correct a decision issue using the edit page which in turn would create a request issue C, we make a request. 

`RequestIssue.active.find_by(contested_decision_issue_id: contested_decision_issue_id, correction_type: nil)` would return request issue B making request issue C ineligible which is wrong. 

I think the code should be something like that: 

```
def check_for_active_request_issue_by_decision_issue!
    return unless contested_decision_issue_id
    return if correction?
    add_duplicate_issue_error(
      RequestIssue.active.find_by(contested_decision_issue_id: contested_decision_issue_id)
    )
  end
```

leikkisa:
We should be closing request issue B if the decision issue gets corrected, so it won't be active.

aroltsch:
@Sjones352 I chatted with Sally and my comment about DTA supplemental claim should be in a separate ticket. You ticket is for the following scenario. 

```Request issue A gets a decision issue A with a denied disposition
Veteran wants to appeal decision issue A
At the same time VBMS submits a correction issue on decision issue A which creates a correction request issue B 
We should allow a veteran to create a request issue C```

The code that @leikkisa suggested is the right approach.

PR #11983, opened 2019-09-05
EP 930s | Update current validation to ensure request issue corrections to not be treated as duplicates 

Sjones352:
Connects #11851


app/models/request_issue.rb
```diff
@@ -914,7 +914,8 @@ def check_for_active_request_issue_by_decision_issue!
     return unless contested_decision_issue_id
 
     add_duplicate_issue_error(
-      RequestIssue.active.find_by(contested_decision_issue_id: contested_decision_issue_id)
+      RequestIssue.active.find_by(contested_decision_issue_id: contested_decision_issue_id,
+                                  correction_type: correction_type)
```
aroltsch: Could we add a comment why we are doing this.

PR #11986, opened 2019-09-05
Timezone for Central Office

### LINKED ISSUE #11796: Hearing API regional office and timezone blank for DC

aplybeah:
Resolves #11796, #11802 

Updated methods in hearing_serializer and legacy_hearing so that the data specific to the Central office can be accessed from the hash.

- Delegation methods are used in `hearing.rb` 
- The `regional_office` model now has updated methods that that pull the timezones and the office names from the hash.


app/models/hearing.rb
```diff
@@ -45,6 +44,8 @@ class HearingDayFull < StandardError; end
   delegate :external_id, to: :appeal, prefix: true
   delegate :regional_office, to: :hearing_day, prefix: true
   delegate :hearing_day_full?, to: :hearing_day
+  delegate :regional_office_name, to: :regional_office
```
lomaxap: this line should be deleted

Issue #11760, opened 2019-08-13
Investigate cache job runtime increase


lowellrex:
Runtimes for [the `UpdateCachedAppealsAttributesJob`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/jobs/update_cached_appeals_attributes_job.rb) increased from ~2.5 minutes to ~7.5 minutes over this past weekend. Short runtimes allow us to run the job more frequently and keep the cache more up-to-date. This ticket exists to determine why runtimes increased and address it if possible. It may not be possible to reduce the job's runtime for any number of reasons (say, we are not querying VACOLS when we weren't before), in which case we will truck on with the ~7.5 minute runtime.

![image](https://user-images.githubusercontent.com/32683958/62971297-77656080-bddf-11e9-8d2f-cc532f81466d.png)
_[Source: Datadog](https://app.datadoghq.com/dashboard/t5t-exj-m32/cronjob-statistics)_

## Acceptance criteria
* [ ] Determine the cause of the increased runtime length
* [ ] Determine if the runtime length can be easily optimized to reduce runtime
* [ ] Create another ticket to actually make the optimization (if we determine it makes sense to)

lowellrex:
_[What is this graph?](https://github.com/department-of-veterans-affairs/caseflow/issues/11533)_
```
      1 | ||||||
      2 | ||||||
      3 | 
      5 | 
      8 | 
```

This is purely an investigation ticket. Additional ticket to optimize the code could be the result of this ticket.

Estimating this at 2. Possible 1 because we are SQL experts and the cause of the runtime increase may be obvious to us upon initial investigation.

lowellrex:
AMA appeal attribute caching takes ~10 seconds while legacy appeal attribute caching takes ~1000 seconds. We should focus our efforts there. I suspect we are taking this long primarily because network round trips to VACOLS are very slow.

![image](https://user-images.githubusercontent.com/32683958/64699375-a28caf80-d472-11e9-8d98-c1ebb2ca0963.png)
[Source: Datadog](https://app.datadoghq.com/metric/explorer?live=true&page=0&is_auto=false&from_ts=1568120433813&to_ts=1568206833813&tile_size=m&exp_metric=dsva_appeals.cache_ama_appeals.runtime%2Cdsva_appeals.cache_legacy_appeals.runtime&exp_agg=max&exp_row_type=metric)

lowellrex:
[...]

PR #12008, opened 2019-09-09
Add instrumentation to the updated cached appeals job

lowellrex:
Connects #11760. Adds instrumentation to determine which particular part of the job is taking a long time to run so we can better know where to devote developer time.

### DIFF
--- a/app/jobs/update_cached_appeals_attributes_job.rb
+++ b/app/jobs/update_cached_appeals_attributes_job.rb
@@ -11,14 +11,17 @@ class UpdateCachedAppealsAttributesJob < CaseflowJob
   METRIC_GROUP_NAME = UpdateCachedAppealsAttributesJob.name.underscore
 
   def perform(_args = {})
-    start_time = Time.zone.now
-
+    ama_appeals_start = Time.zone.now
     cache_ama_appeals
+    time_segment(segment: "cache_ama_appeals", start_time: ama_appeals_start)
+
+    legacy_appeals_start = Time.zone.now
     cache_legacy_appeals
+    time_segment(segment: "cache_legacy_appeals", start_time: legacy_appeals_start)
 
     datadog_report_runtime(metric_group_name: METRIC_GROUP_NAME)
   rescue StandardError => error
-    log_error(start_time, error)
+    log_error(@start_time, error)
   end
 
   # rubocop:disable Metrics/MethodLength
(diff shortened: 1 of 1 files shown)


app/jobs/update_cached_appeals_attributes_job.rb
```diff
@@ -11,14 +11,17 @@ class UpdateCachedAppealsAttributesJob < CaseflowJob
   METRIC_GROUP_NAME = UpdateCachedAppealsAttributesJob.name.underscore
 
   def perform(_args = {})
-    start_time = Time.zone.now
```
lowellrex: This variable is already available as `@start_time` thanks to [the parent class](https://github.com/department-of-veterans-affairs/caseflow/blob/adbfc6ea699c0a00d83baec66a5158f9a9768faa/app/jobs/caseflow_job.rb#L6).

app/jobs/update_cached_appeals_attributes_job.rb
```diff
@@ -155,6 +158,17 @@ def log_error(start_time, err)
 
   private
 
+  def time_segment(segment:, start_time:)
```
lowellrex: Nicked this from a similar method in [the `DataDogService` class](https://github.com/department-of-veterans-affairs/caseflow/blob/adbfc6ea699c0a00d83baec66a5158f9a9768faa/app/services/data_dog_service.rb#L14).

app/jobs/update_cached_appeals_attributes_job.rb
```diff
@@ -155,6 +158,17 @@ def log_error(start_time, err)
...
+    job_duration_seconds = Time.zone.now - start_time
+
+    DataDogService.emit_gauge(
+      app_name: "caseflow_job_segment",
```
hschallhorn: What is "caseflow_job_segment" here? Can't find anything on it.

kevmo: We set the `app_name` argument using the `APP_NAME` in other places in this code, may be more consistent to set this argument using a constant here, too.

lowellrex: I don't actually know where `app_name` is used in DataDog but I included it in here because we use it elsewhere. Happy to remove it from this PR if you'd like though.

lowellrex:
> It may be worth refactoring the Datadog example group in `update_cached_appeals_attributes_job_spec` in response to this change.

I'm intending for this change to be a temporary addition (like, say for a day) so that we can see where best to spend our time when attempting to speed this job up. I suspect that the VACOLS requests occupy the overwhelming majority of the time it takes to do this job, but wanted to get some hard data to confirm.

PR #12020, opened 2019-09-09
Resolve issue with server not starting

ferristseng:
Seeing this error when starting the rails server:

```
...
/home/ftseng/Documents/Code/caseflow/lib/fakes/bgs_service_record_maker.rb:28:in `block in process_csv': undefined method `to_sym' for nil:NilClass (NoMethodError)
```

### Description

  * Looks like we weren't handling `veteran_exists` in the old case statement, so I removed it from the test CSV. https://github.com/department-of-veterans-affairs/caseflow/blob/7e893eb3e30538df3f41c78cfe60fe3d52d91a35/lib/fakes/bgs_service.rb#L49-L50
  * Only call method if `bgs_key` is not `nil`


lib/fakes/bgs_service_record_maker.rb
```diff
@@ -25,7 +25,9 @@ def process_csv
 
       method_name = row_hash["bgs_key"]
 
-      send(method_name.to_sym, veteran)
+      if method_name
```
pkarman: I worry this will bite us if we typo something in the csv file

could we just add a `veteran_exists` method instead?

pkarman: ah nm. I see that this is just checking for a nil value, not whether it is defined.

PR #12023, opened 2019-09-09
may_modify uses both find_flashes and station conflict check

pkarman:
See thread in https://dsva.slack.com/archives/CCB97HZJL/p1568060189094500

We had previously switched away from `find_flashes` because it didn't catch all cases. Now we are adding it back because it does catch some cases that the station conflict logic misses.


pkarman:
I'm a little reluctant to cache `may_modify` because the use case is different. It's only called once per user per intake.

PR #12030, opened 2019-09-10
When filtering on RO join on legacy appeals as well

aroltsch:
### Description
Bulk assignment: when filtering on a RO we should join on legacy appeals as well

### DIFF
--- a/app/workflows/bulk_task_assignment.rb
+++ b/app/workflows/bulk_task_assignment.rb
@@ -43,10 +43,10 @@ class BulkTaskAssignment
         .limit(task_count).order(:created_at)
       if regional_office
         tasks = tasks.joins(
-          "INNER JOIN appeals ON appeals.id = appeal_id AND appeal_type = 'Appeal'"
+          "INNER JOIN appeals ON appeals.id = appeal_id AND appeal_type = '#{Appeal.name}'"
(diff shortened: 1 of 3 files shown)


app/workflows/bulk_task_assignment.rb
```diff
@@ -43,8 +43,10 @@ def tasks_to_be_assigned
         .limit(task_count).order(:created_at)
       if regional_office
         tasks = tasks.joins(
-          "INNER JOIN appeals ON appeals.id = appeal_id AND tasks.appeal_type = 'Appeal'"
-        ).where("appeals.closest_regional_office = ?", regional_office)
+          "INNER JOIN appeals ON appeals.id = appeal_id AND appeal_type = 'Appeal'"
+        ).where("closest_regional_office = ?", regional_office) +
+                tasks.joins("INNER JOIN legacy_appeals ON legacy_appeals.id = appeal_id \
+                  AND appeal_type = 'LegacyAppeal'").where("closest_regional_office = ?", regional_office)
```
lowellrex: Looks good to me! Slight preference for `appeal_type = '#{LegacyAppeal.name}'` (and similar for `Appeal` as well) so that we can be resilient to future changes (though the name of those classes strikes me as an unlikely thing we will ever change).

PR #12034, opened 2019-09-10
stub User.api_user for parallel testing

edmangimelli:
A test in "spec/requests/api/v3/decision_review/higher_level_reviews_controller_spec.rb" was working intermittently. @erluti determined this was due to a combination of caching api_user, running parallel tests, and the database cleanup between tests.
At @pkarman 's suggestion, stubbing method.


edmangimelli:
now that I've skipped the flakey test, all tests are passing, but now my controller has less than 90% coverage which counts as a fail.
Switching this to the stubbed solution Peter mentioned.
https://dsva.slack.com/archives/CKUCFQAHJ/p1568148869020000?thread_ts=1568133904.016100&cid=CKUCFQAHJ

Issue #12037, opened 2019-09-10
Prevent creation of child tasks of TimedHoldTasks


lowellrex:
`TimedHoldTask`s [should not have any child tasks](https://caseflow-looker.va.gov/sql/dns58vwdwyszry) (because they themselves are simply indications that the parent task is on hold for some period of time) but that is not strictly enforced in the code anywhere. This ticket exists to add a guardrail to prevent the creation of child tasks of `TimedHoldTask`s in the code. If we come up with a generic solution here, consider reusing it to satisfy #11996 by making the code "self-documenting".

## Acceptance criteria
* [ ] Prevent creation of child tasks of `TimedHoldTask`s in the code

hschallhorn:
_[What is this graph?](https://github.com/department-of-veterans-affairs/caseflow/issues/11533)_
```
1 | |
2 | |||||
3 | ||||||
5 | 
8 | 
```

Purely preventative. This has not actually happened. Scope of the ticket is pretty straightforward. Estimating at a three if we attempt to generalize this to the base model early. Also is a different way to restrict parent/child relationships.

PR #12038, opened 2019-09-10
Exclude TimedHoldTasks from bulk assignment

lowellrex:
Resolves bug exposed by [this sentry alert](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/6516/) (and [related slack conversation](https://dsva.slack.com/archives/C4JECDLSE/p1568136167020400)).

The bug here is that the application attempts to create child tasks for `TimedHoldTask`s assigned to the `HearingsManagement` organization. That attempt will fail because `TimedHoldTask`s [should not have any child tasks](https://caseflow-looker.va.gov/sql/dns58vwdwyszry) (ticketed in #12037) and, more specifically, the request from the front-end does not pass a field required by the `TimedHoldTask` (`days_on_hold`) so validation thankfully fails to create the invalid task.

This PR prevents this issue by excluding `TimedHoldTask`s from being returned to the front-end as an option in the dropdown menu.

### DIFF
--- a/app/controllers/organizations/task_summary_controller.rb
+++ b/app/controllers/organizations/task_summary_controller.rb
@@ -13,19 +13,23 @@ class Organizations::TaskSummaryController < OrganizationsController
       left join (
         select closest_regional_office
         , id
-        , 'LegacyAppeal' as type
+        , '#{LegacyAppeal.name}' as type
         from legacy_appeals
         union
         select closest_regional_office
         , id
-        , 'Appeal' as type
+        , '#{Appeal.name}' as type
         from appeals
       ) all_appeals
         on tasks.appeal_id = all_appeals.id
         and tasks.appeal_type = all_appeals.type
       where assigned_to_id = #{organization.id}
-        and assigned_to_type = 'Organization'
-        and status in ('assigned', 'in_progress')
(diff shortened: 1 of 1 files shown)


app/controllers/organizations/task_summary_controller.rb
```diff
@@ -13,19 +13,23 @@ def index
...
         and tasks.appeal_type = all_appeals.type
       where assigned_to_id = #{organization.id}
-        and assigned_to_type = 'Organization'
-        and status in ('assigned', 'in_progress')
+        and assigned_to_type = '#{Organization.name}'
+        and status in ('#{Constants.TASK_STATUSES.assigned}', '#{Constants.TASK_STATUSES.in_progress}')
+      -- Exclude TimedHoldTasks from being available for bulk assignment because they should not have child tasks
+      -- and should only be created as children of NoShowHearingTasks, and those tasks will become available for
+      -- bulk assignment after the timed hold has been completed.
+        and tasks.type <> '#{TimedHoldTask.name}'
```
lowellrex: The addition of this line (32) is the only real change to the PR. The other changes are using class names and adding comments.

PR #12040, opened 2019-09-10
Unify sorting of task type on the front and back end

### LINKED ISSUE #11319: Unify how we sort task types on front- and back-end
As [noted in the pull request](https://github.com/department-of-veterans-affairs/caseflow/pull/11247#pullrequestreview-257757985) that introduced back-end sorting of tasks by task type, the front-end and back-end do not return tasks in the same order when we request they be sorted by task type. The front-end uses the task's label field while the back-end uses a combination of task type name and the action field.

## Acceptance criteria
* [ ] Sorting by the "task type" column on the front-end returns tasks in the same order as that same sort on the back-end.

hschallhorn:
Resolves #11319

### Description
Sorts tasks on the back end by their label

### Acceptance Criteria
- [ ] Back end sorts taskColumn by the task's label 

### Testing Plan
1. Ensure sorting the task column results in alphabetical order task labels when sorting with the pagination api

### DIFF
--- a/app/models/task_sorter.rb
+++ b/app/models/task_sorter.rb
@@ -41,6 +41,10 @@ class TaskSorter
   end
 
   def order_clause
+    if column.eql?(Constants.QUEUE_CONFIG.Task_TYPE_COLUMN)
+      return task_type_order_clause
+    end
+
     clauses = column.sorting_columns.map do |col|
       tbl_clause = if sort_requires_case_norm?(col)
                      "UPPER(#{column.sorting_table}.#{col})"
@@ -50,6 +54,10 @@ class TaskSorter
       "#{tbl_clause} #{sort_order}"
     end
     clauses.join(", ")
+  end
+
+  def task_type_order_clause
+    "TODO"
   end
 
   def column_is_valid
--- a/app/models/task_sorter.rb
+++ b/app/models/task_sorter.rb
@@ -41,7 +41,7 @@ class TaskSorter
   end
 
   def order_clause
-    if column.eql?(Constants.QUEUE_CONFIG.Task_TYPE_COLUMN)
+    if column.name.eql?(Constants.QUEUE_CONFIG.COLUMNS.TASK_TYPE.name)
       return task_type_order_clause
     end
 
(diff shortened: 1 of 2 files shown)


app/models/task_sorter.rb
```diff
@@ -41,6 +41,10 @@ def sort_requires_case_norm?(col)
   end
 
   def order_clause
+    if column.name.eql?(Constants.QUEUE_CONFIG.COLUMNS.TASK_TYPE.name)
+      return task_type_order_clause
+    end
+
     clauses = column.sorting_columns.map do |col|
```
lomky: Since we've introduced a branch option up top, could we refactor this second 'branch' option into another function we invoke? Thoughts?

lomky: maybe `default_order_clause`?

app/models/task_sorter.rb
```diff
@@ -52,6 +56,11 @@ def order_clause
     clauses.join(", ")
   end
 
+  def task_type_order_clause
+    task_types_sorted_by_label = Task.descendants.sort_by(&:label).map(&:name)
+    "position(type in '#{task_types_sorted_by_label.join(',')}') #{sort_order}"
```
lomky: There's a lot going on in this line! Could we break it up a bit or drop a comment?

I also would like a little more context for where this sentence return statement comes from, and what it is flowing into? Previously we were returning the relatively straightforward "table ASC/DESC", but now we have a bit more complexity. Maybe just a comment, or if a variable assignment makes sense, a well named var?

spec/models/task_pager_spec.rb
```diff
@@ -187,10 +187,31 @@
 
     context "when sorting by task type" do
```
lomky: Is there an instance in which one of these tests (`task_pager_spec` & `task_sorter_spec`) would fail and the other wouldn't? Can we cut one of them?

hschallhorn: That was my thought as well. There's a ton of overlap in `task_pager_spec` and `task_sorter_spec. I'll look at them tomorrow and determine if we want to condense them or not. May spin out another PR on it

PR #12041, opened 2019-09-10
Add test demonstrating zero legacy cases results in full AMA dockets

pkarman:
Demonstrate existing business logic that zero legacy cases eligible for distribution will result in AMA cases filling the pool.


spec/models/distribution_spec.rb
```diff
@@ -241,6 +241,35 @@
         expect(subject.distributed_cases.count).to eq(15)
       end
     end
+
+    context "when there are zero legacy cases eligible" do
+      let!(:legacy_priority_cases) { [] }
+      let!(:legacy_nonpriority_cases) { [] }
+      let!(:same_judge_nonpriority_hearings) { [] }
```
lomky: Any reason we don't need to overwrite `same_judge_priority_hearings`?

spec/models/distribution_spec.rb
```diff
@@ -241,6 +241,35 @@
...
+        expect(subject.statistics["total_batch_size"]).to eq(45)
+        expect(subject.statistics["priority_count"]).to eq(1)
+        expect(subject.statistics["legacy_proportion"]).to eq(0.0)
+        expect(subject.statistics["direct_review_proportion"]).to be_within(0.01).of(0.13)
+        expect(subject.statistics["evidence_submission_proportion"]).to be_within(0.01).of(0.43)
+        expect(subject.statistics["hearing_proportion"]).to be_within(0.01).of(0.43)
+        expect(subject.statistics["pacesetting_direct_review_proportion"]).to eq(0.1)
+        expect(subject.statistics["interpolated_minimum_direct_review_proportion"]).to eq(0.067)
+        expect(subject.statistics["nonpriority_iterations"]).to be_between(2, 3)
+        expect(subject.distributed_cases.count).to eq(15)
```
lomky: optional: is there anyway to break these into more digestable `it` blocks with some contextual strings?

spec/models/distribution_spec.rb
pkarman: I only overwrote enough fixtures to get the test passing.

spec/models/distribution_spec.rb
pkarman: probably? but this whole spec file would need the same treatment. that's bigger scope that this PR

PR #12046, opened 2019-09-11
Replace arrays with hashes in appeal cache job

### LINKED ISSUE #11760: Investigate cache job runtime increase

lowellrex:
Connects #11760. 

While digging through this file I had a hard time understanding what the logic was actually doing so I've undertaken a slight refactor to improve clarity of what is going on in the `cache_legacy_appeal_vacols_data` function. Does not change any business logic.


app/jobs/update_cached_appeals_attributes_job.rb
```diff
@@ -102,24 +102,37 @@ def cache_legacy_appeal_postgres_data(legacy_appeals)
...
-      vacols_ids = ids.map { |id| id[0] }
-      vacols_folders = VACOLS::Folder.where(ticknum: vacols_ids).pluck(:ticknum, :tinum, :ticorkey)
-      issue_counts_to_cache = issues_counts_for_vacols_folders(vacols_ids)
-      veteran_names_to_cache = veteran_names_for_correspondent_ids(vacols_folders.map { |folder| folder[2] })
-      aod_status_to_cache = VACOLS::Case.aod(vacols_folders.map { |folder| folder[0] })
-      case_status_to_cache = case_status_for_vacols_id(vacols_folders.map { |folder| folder[0] })
-      appeal_assignees_to_cache = assignees_for_vacols_id(vacols_folders.map { |folder| folder[0] })
-
-      values_to_cache = vacols_folders.map do |vacols_folder|
+    all_vacols_ids = legacy_appeals.pluck(:vacols_id).flatten
```
lowellrex: First major change: Collecting the VACOLS IDs into an array that we then batch. We never used the `:id` field we plucked out previously, and flattening this collection allows us to reuse it as `batch_vacols_ids` inside of the loop below.

app/jobs/update_cached_appeals_attributes_job.rb
```diff
@@ -102,24 +102,37 @@ def cache_legacy_appeal_postgres_data(legacy_appeals)
...
+      vacols_folders = VACOLS::Folder
+        .where(ticknum: batch_vacols_ids)
+        .pluck(:ticknum, :tinum, :ticorkey)
+        .map do |folder|
+        {
+          vacols_id: folder[0],
+          docket_number: folder[1],
+          correspondent_id: folder[2]
+        }
+      end
```
lowellrex: Second major change: Collecting the `vacols_folders` as hashes instead of arrays. Primarily done so we don't have to refer back to where we created the array to understand what field `2` refers to, instead we can use the name we assign it in this map.

app/jobs/update_cached_appeals_attributes_job.rb
```diff
@@ -102,24 +102,37 @@ def cache_legacy_appeal_postgres_data(legacy_appeals)
...
+
+      issue_counts_to_cache = issues_counts_for_vacols_folders(batch_vacols_ids)
+      aod_status_to_cache = VACOLS::Case.aod(batch_vacols_ids)
+      case_status_to_cache = case_status_for_vacols_id(batch_vacols_ids)
+      appeal_assignees_to_cache = assignees_for_vacols_id(batch_vacols_ids)
+
+      correspondent_ids = vacols_folders.map { |folder| folder[:correspondent_id] }
+      veteran_names_to_cache = veteran_names_for_correspondent_ids(correspondent_ids)
```
lowellrex: Major change 3: Re-ordered the function calls so they are organized by the parameter they accept.

lowellrex:
> This looks good to me! Could you drop a comment explaining how this resolves #11760 for our institutional memory?

Good catch! It actually doesn't resolve the ticket so I updated the description to "connects" now.

PR #12052, opened 2019-09-11
Filtering cases not in Caseflow

### LINKED ISSUE #12011: Only Return legacy ScheduleHearingTasks where Case location is Caseflow
related to #10820

In order to prevent deduped appeals from appearing in the assign hearings list, we should only [query legacy appeals](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/tasks/schedule_hearing_task.rb#L30) that are in `CASEFLOW`. We do this currently only for central office hearings

sharonwarner:
Resolves #12011 

### Description
This PR updates the tasks_for_ro logic so that we only return appeals that are in Caseflow.


pkarman:
@ferristseng assigning to you to shepherd.

pkarman:
+ @tomas-nava -- please pick up or close.

ferristseng:
I'll pick this one up...this can be something done after pagination lands. The code for this file is moving to the pager. Closing this PR in favor of a future one.

PR #12061, opened 2019-09-11
Add ability to look up BGS participant IDs before logging into Caseflow

lowellrex:
Allows Martina to look up BGS participant IDs before folks log into Caseflow at which point they would become available in the existing dropdown menu. This allows us to create organizations for private attorneys and start populating their queues with `TrackVeteranTask`s so they have something to look at the first time they log in to Caseflow.

image


pkarman:
love this!

client/app/queue/LookupParticipantIdModal.jsx
```diff
@@ -57,16 +80,23 @@ class LookupParticipantIdModal extends React.Component {
...
     this.setState({ representedOrganizations: null });
     this.clearAlerts();
 
     return this.props.requestGet(url).
       then((resp) => this.setState({ representedOrganizations: resp.body.represented_organizations })).
-      catch((err) => this.props.showErrorMessage({ title: 'Error',
-        detail: err.message }));
+      catch(
+        // Errors caught and displayed in requestGet().
+      );
```
lowellrex: Fixes an existing bug with errors not displaying properly.

app/controllers/user_info_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+      )
+    end
+
+    css_id
+  end
+
+  def station_id
+    station_id = params[:station_id]
+
+    unless station_id.is_a?(String) && !station_id.empty? && station_id.match(/[^0-9]/).nil?
```
pkarman: `unless !station_id.empty? && station_id.match(/[^0-9]/).nil?` is logically a little hard to parse.

how about instead

`unless station_id.present? && station_id =~ /^[0-9]+$/`

app/services/external_api/bgs_service.rb
```diff
@@ -276,12 +276,16 @@ def find_all_relationships(participant_id:)
   end
 
   def get_participant_id_for_user(user)
+    get_participant_id_for_css_id_and_station_id(user.css_id, user.station_id)
+  end
+
```
pkarman: nice refactor!

lib/generators/random.rb
```diff
@@ -13,6 +13,11 @@ def word_characters(len = 256)
       from_set(set, len)
     end
 
+    def special_characters(len = 16)
+      set = (" ".."/").to_a + (":".."@").to_a + ("[".."`").to_a + ("{".."~").to_a
```
pkarman: devious!

app/controllers/user_info_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    render json: { represented_organizations: vsos_user_represents }
+  end
+
+  private
+
+  def css_id
+    css_id = params[:css_id]
+
+    unless css_id.is_a?(String) && !css_id.empty? && css_id.match(/[^a-zA-Z0-9_]/).nil?
```
pkarman: how about instead

`unless css_id.present? && css_id =~ /^\w+$/`

lowellrex:
> I tried it locally and I admit the UX is a little odd when the user cannot be found. It says both "Success!" and "User does not belong to any POA organizations." which might be correct but will that make sense to the user?

Right now the only person who uses this feature is Martina and, as far as I know, she does not seem bothered by that oddity. I'm open to other ideas about how that should be handled though!

app/controllers/user_info_controller.rb
lowellrex: Great call! Turns out we don't even need the `present?` condition because null values will be correctly handled by the regex match alone!

PR #12079, opened 2019-09-12
Try harder to get claimant name for decision review task

pkarman:
Observed in production that the claimant's name on the non comp detail page displayed `null` sometimes.

This may be mitigated now that we are backfilling the Person table, but this change would at least show something not-null in the worst case, and explicitly check for a present name in the best base.


app/models/serializers/work_queue/decision_review_task_serializer.rb
```diff
@@ -19,7 +23,7 @@ def self.claimant_name(object)
   def self.claimant_relationship(object)
     return "self" unless decision_review(object).veteran_is_not_claimant
 
-    decision_review(object).claimants.first.try(:relationship)
+    claimant_with_name(object).try(:relationship) || "claimant"
```
aroltsch: Does it make sense to display "claimant" if relationship is nil under relationship as well?

pkarman: good question! added test and fix for that too.

PR #12105, opened 2019-09-13
Alert on /intake if unread inbox messages

pkarman:
connects #11907 

### Description
Add alert to /intake page when there are async job errors.

Note this PR doesn't create any messages, just creates the infrastructure for when they do exist, which will be a follow on PR.

Screen Shot 2019-09-13 at 6 07 47 PM

New Inbox nav link with icon
Screen Shot 2019-09-16 at 9 50 19 AM

New Intake nav link with icon, on the Inbox page only for now.
Screen Shot 2019-09-16 at 10 52 03 AM


client/app/components/InboxLink.jsx
```diff
@@ -0,0 +1,46 @@
...
+});
+const textStyling = css({
+  color: COLORS.PRIMARY,
+  display: 'table-cell',
+  fontSize: '1.7rem',
+  lineHeight: '4em',
+  paddingLeft: '0.5rem',
+  marginBottom: 0
+});
+const youveGotMailStyle = css({
```
leikkisa: You've got mail!

carodew:
Inbox is a pretty new concept, is that right? I may need some more context in order to have a drive-by design opinion – aka, what kinds of messages go in there? how is this different from Queue, how does the rest of Intake function (I'm probably least familiar with that product than any other part of Caseflow at the moment). 

My initial curiosities would be – what are the interaction rules that govern what appears in inbox versus other parts of the application, and is this a pattern that can or should be extended to other parts of Caseflow? (The latter probably would need some research). Also we need to get y'alls a designer. 

I see the PR is already merged, so no need to move backwards. For future reference I would need a day or preferably two turnaround time on Design Opinions™, when that's possible.

Issue #12110, opened 2019-09-16
Prevent DecisionReviewTask from work when zero active request issues


pkarman:
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/6584/

If a task's decision review no longer has any active request issues, the task is supposed to be closed (see #11963). However, we are seeing instances where that did not happen.

If a task has zero actionable request issues, it should not show in the LOB noncomp queue for work.

## AC

- [x] Hide noncomp DecisionReviewTasks that cannot be worked on.
- [x] Create DataIntegrityCheck job to proactively close tasks that cannot be worked.

leikkisa:
I canceled the following "DecisionReviewTasks" that were assigned and had no active request issues:
[227629, 171353, 270379, 225103, 225579, 402032, 293867, 336230, 338052, 382189]

All of them had one ineligible request issue

pkarman:
I think we're proactive enough on this now to close.

PR #12112, opened 2019-09-16
Filter inactive appeals from decision review tasks queue

pkarman:
connects #12110 

### Description
When a task is open but its associated appeal has zero active request issues, hide the task from the queue.

A follow on PR will identify those tasks and close them.


spec/feature/non_comp/reviews_spec.rb
```diff
@@ -45,6 +51,11 @@
                appeal: hlr_b,
                assigned_to: non_comp_org,
                assigned_at: today),
+        create(:higher_level_review_task,
```
pkarman: this task is filtered out and so the tests remain intact.

pkarman:
agree about fix and prevent. this is a stop gap so users aren't wasting time.

PR #12114, opened 2019-09-16
Don't create DecisionReviewTask if there are only ineligible issues

leikkisa:
connects #12110

### Description
- Do not create DecisionReviewTask if there are no eligible request issues
- Run create_business_line_tasks! if there are new request issues, in case there are request issues that don't yet have a VeteranRecordRequest task for appeals
- Only create one VeteranRecordRequest task per business line


app/models/appeal.rb
```diff
@@ -664,9 +664,13 @@ def maybe_create_translation_task
...
   end
 
-  def create_business_line_tasks
-    request_issues.select(&:requires_record_request_task?).each do |req_issue|
-      business_line = req_issue.business_line
+  def create_business_line_tasks!
+    issues_for_tasks = request_issues.select(&:requires_record_request_task?)
+    business_lines = issues_for_tasks.map(&:business_line).uniq
+
+    business_lines.each do |business_line|
```
leikkisa: I decided to do this per business line instead of per request issue, because the Task isn't request_issue specific, so there could otherwise be duplicate tasks per business line

spec/models/request_issues_update_spec.rb
```diff
@@ -254,6 +254,7 @@ def allow_update_contention
         it "saves update, adds issues, and calls create contentions" do
           allow_create_contentions
           allow_associate_rating_request_issues
+          expect(review).to receive(:create_business_line_tasks!).once
```
leikkisa: It turns out that there's already work and tests for removing the active tasks

app/models/appeal.rb
```diff
@@ -647,6 +647,22 @@ def finalized_decision_issues_before_receipt_date
       end
   end
 
+  def create_business_line_tasks!
+    issues_needing_tasks = request_issues.select(&:requires_record_request_task?)
+    business_lines = issues_needing_tasks.map(&:business_line).uniq
```
leikkisa: Previously this was doing this per request issue, but that would create duplicate VeteranRecordRequests if there were two request issues on the same appeal

app/models/appeal.rb
```diff
@@ -647,6 +647,22 @@ def finalized_decision_issues_before_receipt_date
       end
   end
 
+  def create_business_line_tasks!
+    issues_needing_tasks = request_issues.select(&:requires_record_request_task?)
```
leikkisa: requires_record_request_task? checks if the request issue is eligible

app/models/claim_review.rb
```diff
@@ -268,6 +268,7 @@ def incomplete_tasks?
 
   def create_decision_review_task!
     return if tasks.any? { |task| task.is_a?(DecisionReviewTask) } # TODO: more specific check?
+    return if request_issues.active.blank?
```
leikkisa: This is the main line for preventing creating the task if there aren't any eligible request issues

app/models/appeal.rb
pkarman: On a Board Appeal the benefit type (LOB) is assigned per Request Issue, so there may legitimately be multiple tasks per Appeal.

app/models/decision_review.rb
```diff
@@ -338,6 +338,10 @@ def withdrawn_request_issues
     request_issues.withdrawn
   end
 
+  def create_business_line_tasks!
```
pkarman: nice

spec/factories/task.rb
```diff
@@ -500,6 +500,13 @@
       assigned_by { nil }
     end
 
+    factory :decision_review_task, class: DecisionReviewTask do
```
pkarman: I don't think this is correct.

One, DecisionReviewTask is not assigned a parent RootTask, on purpose, since it is not part of the Queue task chain.

Two, there are already factories for `:appeal_task` and `:higher_level_review_task` that do this.

app/models/appeal.rb
pkarman: This is a good change.

app/models/appeal.rb
```diff
@@ -647,6 +647,22 @@ def finalized_decision_issues_before_receipt_date
...
+    business_lines = issues_needing_tasks.map(&:business_line).uniq
+
+    business_lines.each do |business_line|
+      next if tasks.any? { |task| task.is_a?(VeteranRecordRequest) && task.assigned_to == business_line }
```
pkarman: how would this ever be true if we're already got a unique list of business lines?

app/models/appeal.rb
leikkisa: Updated my comment above to indicate it would be duplicated per business line

app/models/appeal.rb
leikkisa: I was thinking if we run create_business_line_tasks! during request issues update, if there are new request issues, that this would prevent a duplicate.

pkarman: ah! good call. I hadn't thought about RequestIssuesUpdate

app/models/request_issues_update.rb
```diff
@@ -30,6 +30,7 @@ def perform!
         edited_request_issue_ids: edited_issues.map(&:id),
         corrected_request_issue_ids: corrected_issues.map(&:id)
       )
+      create_business_line_tasks! if added_issues.present?
```
pkarman: is there a test for this?

spec/factories/task.rb
leikkisa: Ah, good to know. I had searched for decision_review_task, not these. Will update.

app/models/request_issues_update.rb
leikkisa: I added a check to the request issues update spec that it gets called (it's a one line change to that spec below).  I didn't test the create_business_line_tasks! functionality in the RIU test since that has its own tests. I will also add another check that it doesn't get called when there aren't added issues.

PR #12116, opened 2019-09-17
Remove Prometheus service since we will be using Datadog

aroltsch:
### Description
Since we are actively using Datadog we decided to remove Prometheus. 

### Testing Plan
1. Deploy to UAT
2. Ensure the site is up and running
3. Check DD healthy host count for UAT servers https://app.datadoghq.com/dashboard/pww-yy2-j9d/load-balancer-healthy-host-counts?from_ts=1568725645197&fullscreen_widget=262865873&live=true&tile_size=m&to_ts=1568729245197&fullscreen_section=overview
4. Check Kibana logs for prometheus requests receiving route not found


app/middleware/metrics_collector.rb
```diff
@@ -1,68 +0,0 @@
-# frozen_string_literal: true
-
-# Simple middleware that collects gauge metrics whenever
-# a GET /metrics request is made. This ensures we regularly
-# get a snapshot of instance information
-class MetricsCollector
```
aroltsch: MetricsCollector was only used to expose metrics to the `/metrics` endpoint which was used by the Prometheus server to collect metrics

app/middleware/metrics_collector.rb
```diff
@@ -1,68 +0,0 @@
...
-    conns = VACOLS::Record.connection_pool.connections
-    active = conns.count { |c| c.in_use? && c.owner.alive? }
-    dead = conns.count { |c| c.in_use? && !c.owner.alive? }
-    idle = conns.count { |c| !c.in_use? }
-
-    PrometheusService.vacols_db_connections.set({ type: "active" }, active)
-    PrometheusService.vacols_db_connections.set({ type: "dead" }, dead)
-    PrometheusService.vacols_db_connections.set({ type: "idle" }, idle)
-
-    emit_datadog_point("vacols", "active", active)
```
aroltsch: I wonder if this is something we are still interested to be sent to DD.

aroltsch: I am not sure if this information about connections is accurate. I don't think it is counting connections from the jobs.

app/controllers/health_checks_controller.rb
```diff
@@ -2,6 +2,7 @@
 
 class HealthChecksController < ActionController::Base
   include TrackRequestId
+  include CollectDataDogMetrics
```
aroltsch: Since we don't have `/metrics` endpoint, we can collect metrics when `/health_checks` endpoint is hit which I think is used by the load balancer.

aroltsch:
Testing:
-[ ] Make sure Prometheus is scrapping metrics in UAT by looking at the logs:

image

-[ ] Deploy by branch to UAT that removes Prometheus service from Caseflow, make sure `metrics` endpoint is not available

image


-[ ] Make sure app is up and running in Caseflow

-[ ] Make sure Promtheus is no longer scrapping in UAT

PR #12117, opened 2019-09-17
Refactor queue selection and legacy task eager loading

### LINKED ISSUE #12059: Shows tasks in Attorney queue
> Attorney's queue page must pull both types of cases: legacy appeals from Caseflow tasks table and legacy appeals from Vacols that match attorney's location

Ensure that the attorney's queue is returning the newly created `AttorneyTask`s from #12022

`AttorneyLegacyTask` is the old way of tracking the task after it gets assigned to an attorney. We shouldn't return these tasks for the new workflow: 

https://github.com/department-of-veterans-affairs/caseflow/blob/3ad02facb015b7228fb646eb75118546876ba430/app/models/legacy_tasks/attorney_legacy_task.rb#L23-L25

The query to fetch tasks:

https://github.com/department-of-veterans-affairs/caseflow/blob/3ad02facb015b7228fb646eb75118546876ba430/app/models/queues/legacy_work_queue.rb#L22-L37

Here's where it gets called:

https://github.com/department-of-veterans-affairs/caseflow/blob/3ad02facb015b7228fb646eb75118546876ba430/app/controllers/legacy_tasks_controller.rb#L24-L31

If the `FeatureToggle` is enabled, we could utilize the existing `AttorneyQueue` to get tasks for the user.

The request gets routed from `/queue/:user_id`:

https://github.com/department-of-veterans-affairs/caseflow/blob/1bc8f57748b8834d8dadef658715cf1a4904fb63/config/routes.rb#L221

ferristseng:
### Description

- Refactor which queue to use to `QueueForRole` class
- Refactor legacy task eager loading to method on serializer


app/workflows/queue_for_role.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class QueueForRole
```
lomaxap: 💯 this class is a great idea!

Issue #11982, opened 2019-09-04
Intake | Alphabetize Intake drop down menu items


jimruggiero:
*** Story
As an intake user, I want to see menu items in all drop downs in my intake workflows listed in numeric, then alphabetical order, so that I can find what I need more quickly. 

## AC:
1. Scenario: Login and initial form selection is required
     [ ] 1. Select form, list appears sorted as:
          - Decision Review Request: Board Appeal (Notice of Disagreement) - VA Form 10182
          - Decision Review Request: Higher Level Review - VA Form 20-0996
          - Decision Review Request: Supplemental Claim - VA Form 20-0995
          - RAMP Opt-In Election Form
          - RAMP Selection (VA Form 21-4138)

2. Scenario: HLR (or SC) and Comp are selected, Other than vet? = Y
     [ ] 4. Add Issues modal screen 1, list appears sorted as:
          - "Other non-rating", then 
          - "Unknown issue category", then 
          - Sort remaining list items A-Z ascending

3. Scenario: HLR (or SC) and Pension are selected, Other than vet? = Y
     [ ] 4. Add Issues screen modal screen 1, the Issue Category list appears sorted as:
          - Apportionment
          - Burial Benefits - NCS Burial
          - Burial Benefits - Plot or Internment Allowance
          - Burial Benefits - Transportation Allowance
          - Burial Benefits - VA Hospitalization Death
          - Dependent Eligibility | Adoption
          - Dependent Eligibility | Stepchild
          - Dependent Eligibility | School child
          - Dependent Eligibility | Validity of Marriage
          - Dependent Eligibility | Parent(s)
          - Dependent Eligibility | Other
          - Effective date | Liberalizing Legislation
          - Effective date | 3.400(b)(ii)(B)
          - Eligibility | Wartime service
          - Eligibility | Veteran Status
          - Income/Net Worth | Countable Income
          - Income/Net Worth | Medical Expense Deductions
          - Income/Net Worth | Residential Lot Size
          - Overpayment | Validity of debt
          - Overpayment | Waiver
          - Penalty Period
          - Post Award Audit
          - Survivors pension eligibility

4. The payee code drop down in screen "3. Review" when HLR or SC and Comp or Pension benefit type are selected, and "Other than claimant? = Y": No menu changes; keep list as-is.

jimruggiero:
Per Lauren S on 9/18 via email, AMO is good with these changes.

PR #12127, opened 2019-09-17
Intake | Alphabetize Intake drop down menu items

Sjones352:
Connects #11982
Screen Shot 2019-09-17 at 12 36 26 PM

Screen Shot 2019-09-16 at 1 54 22 PM


jcq:
Yeah, in general we'd want to avoid doing things like sorting in something that is called for every render due to performance issues. You could move the sorting outside of render and do only once, or memoize the sorting if it is dependent upon some sort of prop or state variable. 

However, @leikkisa's suggestion avoids the need for all of those things, so would be the best solution given the current requirements. (This also allows you to avoid pulling in Lodash, which is always good.)

jcq:
How about in the test using the existing array, running a sort (to guarantee order for testing purposes) and checking against first and last element of that? That way the test still works if something is added/removed from the options.

spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -669,6 +668,15 @@ def complete_higher_level_review
...
 
+      scenario "form sorted in ascending order" do
+        visit "/intake"
+
+        first_form = Constants.INTAKE_FORM_NAMES.appeal
+        second_form = Constants.INTAKE_FORM_NAMES.higher_level_review
+        expect(first_form).to eq("Decision Review Request: Board Appeal (Notice of Disagreement) — VA Form 10182")
+        expect(second_form).to eq("Decision Review Request: Higher-Level Review — VA Form 20-0996")
+      end
+
```
leikkisa: Honestly, I don't feel a strong need to test this.  But if you do, I think a more intuitive location for this test would be: caseflow/spec/feature/intake/intake_spec.rb

I am also okay with not having this test.

Issue #11757, opened 2019-08-13
Add Pagination to Assign Hearings


lomaxap:
#### AC
- [ ] Verify that all assign hearings tabs paginate
- [ ] Verify that user is able to sort and filter across tabs

lomaxap:
- We need to add an endpoint (i don't think we can change the existing `organization/tasks` endpoint to meet our needs) for assign hearings table that returns a QueueConfig response with [tabs](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/queue_config.rb#L40-L46) that are specific to hearings. A Queue tab that looks something like:

```ruby
class LegacyVeteransWaitingTab < QueueTab
 def label
    "Legacy Veterans Waiting"
  end

  def self.tab_name
    "legacyAppeals"
  end

  def description
    ""
  end

  def tasks
    ## ScheduleHearingTask#tasks_for_ro
  end

  def column_names
    [
      Constants.QUEUE_CONFIG.CASE_DETAILS_LINK_COLUMN,
      Constants.QUEUE_CONFIG.APPEAL_TYPE_COLUMN,
      Constants.QUEUE_CONFIG.DOCKET_NUMBER_COLUMN,
      Constants.QUEUE_CONFIG.HEARING_SUGGESTED_LOCATION_COLUMN # this doesn't exist yet
    ].compact
  end
end
```
- The `#tasks` method on the tab should be some variation of [`ScheduleHearingTasks#task_for_ro`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/tasks/schedule_hearing_task.rb#L13-L26)

- We need some way to alter `QueueConfig#tabs` to return different tabs for the same hearings user based on the page they are on (`/organization/<org-name>` vs `/hearings/schedule/assign`)

### Questions

- What do we need to change for `TaskPager` and `TaskFilter` to make it work for `suggestedLocation`?
- Once we've add our tabs, can we just pass the configuration objects to the QueueTable in the same way we're doing [here](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/OrganizationQueue.jsx#L109)?

ferristseng:
Docs are here: https://github.com/department-of-veterans-affairs/caseflow/wiki/Caseflow-Hearings

chandracarney:
Path we chose for pagination: no feature flag, deploy straight to production. Why? Request for a feature flag came in after development started and would add ~3+ days to dev time.
[...]

PR #12128, opened 2019-09-17
Add closest_regional_office_key to cached attributes

lomaxap:
connects #11757 

- Adds `closest_regional_office_key` to both ama and legacy cached attributes


app/jobs/update_cached_appeals_attributes_job.rb
```diff
@@ -42,6 +42,7 @@ def cache_ama_appeals
         assignee_label: appeal_assignees_to_cache[appeal.id],
         case_type: appeal.type,
         closest_regional_office_city: regional_office ? regional_office[:city] : COPY::UNKNOWN_REGIONAL_OFFICE,
+        closest_regional_office_key: regional_office ? appeal.closest_regional_office : COPY::UNKNOWN_REGIONAL_OFFICE,
```
hschallhorn: Is whatever you're going to use this fo going to want to see "Unknown" here or nil?

lomaxap: For my use, I will always be filtering on this column, so it should never return either. Copied the above code for consistency. Do you have a preference?

hschallhorn: No preference!

Issue #11310, opened 2019-07-08
Cache responses from queue pagination API


lowellrex:
Implement some way to store responses from the pagination API in state (result sets keyed off of the combination of filters/sorts/pages?) so we don't have to make an API request for all new page requests, filter changes, and re-sorting with a loading spinner when those API requests are in flight. This caching allows us to prevent making additional API calls when the pagination API has already returned a set of tasks that fit on a single page, so we have all of the information we need to apply additional sorts and filters without hitting the server.

## Blocked by
* [x] Enable TaskTable to use API requests to filter tasks #11309 (only makes sense to cache these requests when we are making them)

## Acceptance criteria
* [ ] Avoid making redundant requests to pagination API.

PR #12131, opened 2019-09-17
Implement naive caching for queue pagination API

lowellrex:
Connects #11310. Adds naive caching based only on the URL of the endpoint we are making the task page request to. Later on, we can add the ability to bypass the API request when there is only a single page of results (thus the front-end already has access to all of the rows that the back-end could possibly return so we can do all of the filtering on the front-end without needing to involve the back-end).

In the screen capture below, notice that the second time we make a request for hearing clarification tasks we do not make an API request.

naive_caching


hschallhorn:
I think we should cache the tasks from viewing the default view (page 1, no filters, no sort) as well, so we don't get the request after clearing the filters in the screen grab

lowellrex:
> I second the desire to cache our initial empty params tasks too, but happy for that to be in this PR or a follow on!

Will cache initial set of tasks in another PR.

PR #12132, opened 2019-09-17
Add DecisionReviewTasksForInactiveAppealsChecker

pkarman:
connects #12110 

### Description
Sends Slack message when we have open Business Line tasks for inactive appeals.

### DIFF
--- a/app/services/decision_review_tasks_for_inactive_appeals_checker.rb
+++ b/app/services/decision_review_tasks_for_inactive_appeals_checker.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+class DecisionReviewTasksForInactiveAppealsChecker < DataIntegrityChecker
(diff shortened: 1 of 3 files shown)


app/services/decision_review_tasks_for_inactive_appeals_checker.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class DecisionReviewTasksForInactiveAppealsChecker < DataIntegrityChecker
+  def call
+    tasks_with_inactive_appeals.each do |task|
+      add_to_report "#{task.type} #{task.id} should be canceled"
```
leikkisa: I'm going to lean away from Abby's training on using "canceled" because I think at some point folks agreed on cancelled? :-P

PR #12149, opened 2019-09-17
Rubaiyat/12022 create attorney task for assigned legacy cases

### LINKED ISSUE #12022: Create AttorneyTask when a judge assigns a legacy case to an attorney
> Create AttorneyTask when a judge assigns a legacy case to attorney only if the feature toggle is enabled and there exists a JudgeAssignTask for that appeal in Caseflow

It looks like this is the area where we are assigning the case to an attorney:

https://github.com/department-of-veterans-affairs/caseflow/blob/476ae88ec068fcb3815ad9933e20490b11233b86/app/models/judge_case_assignment_to_attorney.rb#L12-L22

The method creates a record in DECASS.

https://github.com/department-of-veterans-affairs/caseflow/blob/476ae88ec068fcb3815ad9933e20490b11233b86/app/repositories/queue_repository.rb#L119-L144

There is also a method to `reassign_to_attorney!`, which updates the record in DECASS. We can update this too to create a new `AttorneyTask` instead of touching DECASS.

---

The entry point to this code is the `legacy_tasks_controller`, which is being called here:

https://github.com/department-of-veterans-affairs/caseflow/blob/a5ff9639e15bbe85f460561ae61dc57b077db8ad/client/app/queue/UnassignedCasesPage.jsx#L41

here:

https://github.com/department-of-veterans-affairs/caseflow/blob/a5ff9639e15bbe85f460561ae61dc57b077db8ad/client/app/queue/AssignedCasesPage.jsx#L62

and here:

https://github.com/department-of-veterans-affairs/caseflow/blob/a5ff9639e15bbe85f460561ae61dc57b077db8ad/client/app/queue/AssignToAttorneyModalView.jsx#L18-L37

rubaiyat22:
Resolves #12022

### Description
- continuation of das deprecation workflow
- create AttorneyTask for Legacy appeals if legacy_das_depcrecation FeatureToggle is enabled & there exists a JudgeAssignTask for the appeal
- create a reassign method in which a judge can reassign an AttorneyTask to another attorney


client/app/queue/QueueActions.js
```diff
@@ -399,6 +399,35 @@ export const bulkAssignTasks =
...
+  if (oldTask.appealType === 'Appeal') {
+    const amaTasks = resp.tasks.data;
+
+    dispatch(onReceiveAmaTasks(
+      amaTasks
+    ));
+  } else if (isInitial && (oldTask.appealType === 'LegacyAppeal' && !oldTask.isLegacy)) {
+    //For das depractaion, legacy_task_controller #create returns tasks not a task
+    const tasks = resp.tasks.data;
+    const allTasks = prepareAllTasksForStore(tasks);
```
lomaxap: since the only thing that changes here is the value of task(s), we could conditionally set that value and only have 1 `else` statement.

```javascript
const taskData = resp.tasks.data;
const tasks = isInitial && (oldTask.appealType === 'LegacyAppeal' && !oldTask.isLegacy) ? taskData : [taskData];
```

rubaiyat22: Good idea! I totally missed this!

PR #12163, opened 2019-09-19
Remove loofah package from Gemfile.

kevmo:
Remove loofah package from Gemfile, as it's a dependency for a package, active_model_serializers, we are no longer using.


pkarman:
@kevmo this one could be merged after code thaw?

Issue #11193, opened 2019-06-24
Add HLR and SC 930 claim labels to end product codes


leikkisa:
### Story
Users will be able to edit DTA claims that are already cleared in order to make corrections. These claims should be established under the appropriate 930 end product codes.

### Acceptance criteria
- The below end product codes are added to the tree of end product codes referenced in request_issue.rb.  This could under "correction", and organized similarly to non-930 codes where dta and original are broken out into separate branches on the tree.
- Logic for choosing the end product code is updated to choose the dta correction end product code.  The review's "decision_review_remanded" will be useful for determining this logic.
- The end product codes should already be added to end_product.rb and end_product_codes.json, but please confirm
- Bonus points: The code for selecting the correct end product code is refactored into a separate workflow, and/or the tree is refactored into another format to make end product code selection simpler.

Please see PR for the actual list of codes - I am removing the short list from here because the actual list is quite long.

leikkisa:
Waiting on a new list of EP codes, it seems like we are missing some

jimruggiero:
@leikkisa can you add a screen shot or gif to this ticket when it's ready to be merged?

leikkisa:
@jimruggiero sure, I can add a screenshot. I think I'll add one of the confirmation screen, because this ticket is all backend.

I am going to keep this in progress until we hear more about some missing dta correction pension codes.

leikkisa:
![Screen Shot 2019-09-20 at 4.52.00 PM.png](https://images.zenhubusercontent.com/58f6b6eb51aa2c30577bb60e/03cc355a-9a73-4353-97b1-1a27001064d7)

leikkisa:
I am moving this issue over, because the "missing EP codes" weren't in the original requirements for VBMS, and I think it might take a while to resolve that. I'll open a new (blocked) issue to handle that once we get some answers on it.

PR #12169, opened 2019-09-19
Refactor request issue end_product_code, add codes for 930s

leikkisa:
connects #11193

### Description
- Add known 930 codes for DTA claims. Note, there are some issues with this list, but I want to go ahead and get this in for VBMS testing, and it will not cause problems in prod because it's behind a feature flag.
- Refactors the code for selecting the EP code for a request issue
- Refactors the test for EP code selection too.  This is an unusual pattern for tests, but it's very easy to make mistakes while copying and pasting these (it has happened multiple times in the past), so I wanted to be able to test them against a "CSV" style structure made from a spreadsheet of them instead of copy/pasting them individually into the tests.  Since the 930's add a permutation type (correction type), it substantially increases the number of EP codes added.  Feedback welcome.
- Removes "temp_benefit_type" because removing it was a TODO (per Code Climate)

### Documentation Updates


app/models/request_issue.rb
```diff
@@ -664,6 +547,11 @@ def remanded?
     decision_review.try(:decision_review_remanded?)
   end
 
+  # TODO: use request issue benefit type once it's populated for request issues on build
+  def temp_find_benefit_type
```
leikkisa: It turns out this is only used in the EP code selector, so I'll move it over too

spec/feature/queue/motion_to_vacate_spec.rb
```diff
@@ -264,7 +264,6 @@ def address_motion_to_vacate(user:, appeal:, judge_task:)
 end
 
 def format_judge_instructions(notes:, disposition:, vacate_type:, hyperlink: nil)
-  binding.pry
```
Sjones352: Forgot a binding pry here

leikkisa: This is a removal. I think JC also caught it in one of his PRs but I saw it in my branch too when searching for my binding.pry's

app/workflows/end_product_code_selector.rb
```diff
@@ -0,0 +1,222 @@
+# frozen_string_literal: true
+
+##
+# Selects the appropriate end product code for a request issue
+
+# The first branch differentiates between primary issues and correction issues
+# For corrections, the next branch is correction type
+# From there, the next branch is original or dta, then benefit type, then review type, then issue type
+
+class EndProductCodeSelector
```
Sjones352: Great job on moving these in there own class.

app/workflows/end_product_code_selector.rb
```diff
@@ -0,0 +1,222 @@
+# frozen_string_literal: true
+
+##
+# Selects the appropriate end product code for a request issue
+
+# The first branch differentiates between primary issues and correction issues
```
leikkisa: I wanted to call non-correction issues "original", but we're already doing that to differentiate dta versus non-dta, so I picked "primary". Feedback on naming is welcome though!

app/models/claim_review.rb
```diff
@@ -80,19 +80,7 @@ def finalized_decision_issues_before_receipt_date
...
-        issue.update!(
-          end_product_establishment: end_product_establishment_for_issue(issue),
-          benefit_type: benefit_type,
-          veteran_participant_id: veteran.participant_id
-        )
-      end
-      RequestIssueCorrectionCleaner.new(issue).remove_dta_request_issue! if issue.correction?
-      issue.create_legacy_issue_optin if issue.legacy_issue_opted_in?
-    end
+    new_issues.each(&:create_for_claim_review!)
```
leikkisa: This was a code climate fix because this method called "issue." so much

client/constants/END_PRODUCT_CODES.json
```diff
@@ -103,5 +103,6 @@
   "930ASCRNQPMC": "PMC Supplemental Claim Correction of Rating NQE 930",
   "930ASNRCPMC": "PMC Supplemental Claim Non-Rating Control 930",
   "930RC": "Rating Control",
-  "930RCPMC": "PMC-Rating Control"
+  "930RCPMC": "PMC-Rating Control",
+  "930AMASDO": "AMA Supplemental Difference of Opinion"
```
leikkisa: We're not using this anywhere yet, I'm not sure what it is, but I added it from their new list

spec/workflows/end_product_code_selector_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+           benefit_type: benefit_type,
+           correction_type: correction_type,
+           is_unidentified: is_unidentified)
+  end
+
+  context "#call" do
+    subject { EndProductCodeSelector.new(request_issue).call }
+
+    context "for an unidentified issue" do
+      let(:is_unidentified) { TRUE }
```
tomas-nava:
```suggestion
      let(:is_unidentified) { true }
```

I get a warning on this line: `warning: constant ::TRUE is deprecated`

app/models/request_issue.rb
```diff
@@ -928,35 +820,6 @@ def check_for_active_request_issue_by_decision_issue!
...
   end
 
-  def original_end_product_code
-    choose_end_product_code(END_PRODUCT_CODES[:original][temp_find_benefit_type.to_sym])
-  end
-
-  # TODO: use request issue benefit type once it's populated for request issues on build
-  def temp_find_benefit_type
-    decision_review.benefit_type || benefit_type || contested_benefit_type
-  end
```
leikkisa: I confirmed in prod that request issues all have a benefit type populated

app/models/request_issue.rb
```diff
@@ -300,12 +185,19 @@ def attributes_look_like_contested_issue?(data)
 
   delegate :veteran, to: :decision_review
 
+  def create_for_claim_review!
+    return unless decision_review.is_a?(ClaimReview)
+
+    update!(benefit_type: decision_review.benefit_type, veteran_participant_id: veteran.participant_id)
+    update!(end_product_establishment: decision_review.end_product_establishment_for_issue(self))
```
leikkisa: This update needs to happen separately from the above update, because benefit type needs to be populated before choosing the end product code for creating the end product establishment.

spec/workflows/end_product_code_selector_spec.rb
```diff
@@ -0,0 +1,130 @@
+# frozen_string_literal: true
+
+require "support/database_cleaner"
+require "rails_helper"
+
```
tomas-nava: I like the efficiency of writing the tests this way, but I'd appreciate a comment here describing the contents of `EP_CODES` so it'll be easier for someone else to update them later.

leikkisa:
Adding a note regarding the "issues" with the 930 list of EPs.  We did ask for the latest list, so this is the latest one.  It does not have all of the expected pension EP codes.

app/models/claim_review.rb
tomas-nava: I like it better this way anyway!

app/models/request_issue.rb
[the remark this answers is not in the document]
leikkisa: There was a little bit of a weird code block before, where it didn't add the end product establishment in some cases as well.  I moved those conditions to end_product_establishment_for_issue there because I thought it was more intuitive.

```
if processed_in_caseflow? || !issue.eligible?	
      else	
        issue.update!(	
          end_product_establishment: end_product_establishment_for_issue(issue),	
          benefit_type: benefit_type,	
          veteran_participant_id: veteran.participant_id	
        )	
      end
```

spec/workflows/end_product_code_selector_spec.rb
leikkisa: Good call!  Will do

app/models/request_issue.rb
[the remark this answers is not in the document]
leikkisa: I see it! Will only update if it's not nil.

Issue #11996, opened 2019-09-06
Task model documentation


pkarman:
We need some top-of-the-file docs at the level of the individual Task model in the code.

I would find the following helpful comments in each `app/models/tasks/*.rb` file:

* what milestone in the overall Caseflow decision review process does this task map to?
* under what conditions would a record of this type be created?
* what parent record types would it potentially have? are any required?
* what child record types would it potentially have? are any required?
* what kind of user would be assigned to this task?
* under what conditions would the status change? from what value to what value?

PR #12172, opened 2019-09-19
Update JudgeAssignTask documentation

lowellrex:
Connects #11996. Starting the in-file documentation update for tasks with `JudgeAssignTask` as an example.


app/models/tasks/judge_assign_task.rb
```diff
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
 
 ##
-# Task for a judge to assign tasks to attorneys.
+# Task assigned to judge from which they will assign the associated appeal to one of their attorneys by creating a
```
pkarman: what is the name of the "task to draft a decision on the appeal"?

app/models/tasks/judge_assign_task.rb
```diff
@@ -1,7 +1,11 @@
...
+# task to draft a decision on the appeal.
+# Task is created as a result of case distribution.
+# Task should always have a RootTask as its parent.
+# Task can have no child tasks, AttorneyTask children, or ColocatedTask children.
```
pkarman: are these statements redundant? Is it enough to say "Task can have no child tasks" -- or is there some nuance here I'm missing?

app/models/tasks/judge_assign_task.rb
lowellrex: `AttorneyTask` is the name of the task but I intentionally avoided using the name for 2 reasons. First, the task name could change (I think it would make sense to make it more precise) and if the name of the task changes, I think it is likely we would miss that change in the documentation (same reason I prefer `AttorneyTask.name` to the string `"AttorneyTask"`). And second, `AttorneyTask` is a base class for other types of tasks attorneys have to do in addition to being a task that is directly instantiated itself, so there may be some confusion about whether we mean `AttorneyTask` in the sense of `AttorneyTask` and all of its sub-classes, or just instances of `AttorneyTask`. This is a confusing situation that should also probably be resolved in the code itself.

Happy to change "task to draft a decision on the appeal" to `AttorneyTask` if you'd like though.

app/models/tasks/judge_assign_task.rb
lowellrex: Great catch! I'll clarify this language.

app/models/tasks/judge_assign_task.rb
pkarman: gotcha.

Maybe "by creating a task (often an AttorneyTask or one of its subclasses) to draft a decision on the appeal" would clarify?

lowellrex: I think it would be more accurate to say "by creating a task (an AttorneyTask but not any of its subclasses) to draft a decision on the appeal". Does that look good to you?

pkarman: great!

PR #12174, opened 2019-09-19
Lock mutli-task creation into transaction

### LINKED ISSUE #8437: Put task creation functions inside of a transaction block
Currently, when we create a parent task successfully but fail to create a child task for whatever reason we raise an error but persist the parent task to the database. We should not create any tasks in that lineage if we fail to create a child task. [`ColocatedTask`](https://github.com/department-of-veterans-affairs/caseflow/blob/6bea2aab4a10d6bf3de6b4b26ab1f229c9f5637a/app/models/tasks/colocated_task.rb#L16) already puts creation attempts inside a transaction block which avoids that problem for that particular task type. We should put creation attempts for all task types in transaction blocks to avoid this problem universally (perhaps [in `Task.create_many_from_params`](https://github.com/department-of-veterans-affairs/caseflow/blob/6bea2aab4a10d6bf3de6b4b26ab1f229c9f5637a/app/models/task.rb#L82)?).

## Acceptance criteria
* [ ] If we fail to create a child task we should not create the parent task

lomky:
Resolves #8437 

### Description
Lock mutli-task creation into transaction

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -251,7 +251,7 @@ class Task < ApplicationRecord
   end
 
   def self.create_many_from_params(params_array, current_user)
-    ActiveRecord::Base.multi_transaction do
+    multi_transaction do
       params_array.map { |params| create_from_params(params, current_user) }
     end
   end
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -251,7 +251,9 @@ class Task < ApplicationRecord
   end
 
   def self.create_many_from_params(params_array, current_user)
-    params_array.map { |params| create_from_params(params, current_user) }
+    ActiveRecord::Base.multi_transaction do
+      params_array.map { |params| create_from_params(params, current_user) }
+    end
   end
 
   def self.modify_params(params)
--- a/app/models/tasks/no_show_hearing_task.rb
+++ b/app/models/tasks/no_show_hearing_task.rb
@@ -6,12 +6,14 @@ class NoShowHearingTask < Task
   before_validation :set_assignee
 
   def self.create_with_hold(parent_task)
-    create!(parent: parent_task, appeal: parent_task.appeal).tap do |no_show_hearing_task|
-      TimedHoldTask.create_from_parent(
-        no_show_hearing_task,
-        days_on_hold: 25,
-        instructions: ["Mail must be received within 14 days of the original hearing date."]
-      )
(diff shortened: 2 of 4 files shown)


app/models/task.rb
```diff
@@ -251,7 +251,9 @@ def self.incomplete_or_recently_closed
   end
 
   def self.create_many_from_params(params_array, current_user)
-    params_array.map { |params| create_from_params(params, current_user) }
+    ActiveRecord::Base.multi_transaction do
```
pkarman: fwiw I *think* you can just call `multi_transaction` here with no class prefix since all our AR models inherit from `ActiveRecord::Base`

app/models/tasks/no_show_hearing_task.rb
```diff
@@ -6,12 +6,14 @@ class NoShowHearingTask < Task
   before_validation :set_assignee
 
   def self.create_with_hold(parent_task)
-    create!(parent: parent_task, appeal: parent_task.appeal).tap do |no_show_hearing_task|
-      TimedHoldTask.create_from_parent(
-        no_show_hearing_task,
-        days_on_hold: 25,
-        instructions: ["Mail must be received within 14 days of the original hearing date."]
-      )
+    ActiveRecord::Base.multi_transaction do
```
pkarman: same thing here

app/models/tasks/special_case_movement_task.rb
```diff
@@ -13,17 +13,19 @@ class SpecialCaseMovementTask < Task
...
-                            assigned_to: assigned_to,
-                            assigned_by: assigned_by,
-                            instructions: instructions)
-    # We don't want the judge to have to worry about the SpecialCaseMovementTask,
-    #   so we assign it to the SCM user that assigned this.
-    update!(status: Constants.TASK_STATUSES.completed, assigned_to: assigned_by)
-    # For now, we expect the parent to always be the distribution task
-    #   so we don't worry about distribution task explicitly
-    parent.update!(status: Constants.TASK_STATUSES.completed)
+    ActiveRecord::Base.multi_transaction do
```
pkarman: this one is inside an instance method but could also be `Task.multi_transaction`

also, I think the `multi_transaction` is when we write to both postgresql and VACOLS in a single block. If we're not writing to VACOLS, can just use `transaction`.

Issue #12093, opened 2019-09-13
Use claim date for submit date when creating contentions


leikkisa:
### Story
When we create contentions in VBMS, we set the "submit date" to today.  We have been asked to set that to the claim date of the end product establishment.

### Acceptance criteria
When we create contentions in VBMS, the submit_date is set to the claim date

### Technical notes
This will require multiple PRs.

1. First in the connect_vbms repo, update create_contentions to accept a claim_date. Test this against VBMS UAT before merging.
2. Update caseflow's repo to use the new connect_vbms gem. This can be done in its own small PR if desired.  Then when creating contentions in Caseflow, include an additional field called "claim_date" and set it to the end product establishment's "claim_date".  Look at the "build_contentions" method for where to add this data.

PR #12178, opened 2019-09-20
Use claim date for submit date when creating contentions

Sjones352:
Connects #12093


app/models/ramp_refiling.rb
```diff
@@ -90,7 +90,8 @@ def create_contentions_in_vbms
       veteran_file_number: veteran_file_number,
       claim_id: end_product_establishment.reference_id,
       contentions: contention_descriptions_to_create.map { |desc| { "description": desc } },
-      user: intake_processed_by
+      user: intake_processed_by,
+      claim_date: claim_date
```
leikkisa: You could put receipt date here directly instead of aliasing it with claim date in ramp review.

Issue #12088, opened 2019-09-13
Paginated filtering still shows additional pages


hschallhorn:
When filtering multiple pages of tasks with the pagination API, the later pages will display additional unfiltered tasks.

### Filtering without pagination

<kbd><img width="840" alt="Screen Shot 2019-09-13 at 10 21 23 AM" src="https://user-images.githubusercontent.com/45575454/64870767-e538be00-d611-11e9-9e93-eecb61dd7a22.png"></kbd>

_Only shows one available page_
_"Viewing 1-2 of 2 total"_

### Filtering with pagination

<kbd><img width="833" alt="Screen Shot 2019-09-13 at 10 14 18 AM" src="https://user-images.githubusercontent.com/45575454/64870790-f255ad00-d611-11e9-95d6-f9a5f1812150.png"></kbd>
<kbd><img width="845" alt="Screen Shot 2019-09-13 at 10 14 35 AM" src="https://user-images.githubusercontent.com/45575454/64870793-f386da00-d611-11e9-9373-48557f82ac8a.png"></kbd>

_Shows 2 available pages_
_"Viewing 1-2 of 22 total"_

## AC

- [ ] Page filtered results correctly
  - [ ] "Viewing X-X of X total" should be correct for the number of returned tasks
  - [ ] Only shows available pages for the number of returned tasks

hschallhorn:
_[What is this graph?](https://github.com/department-of-veterans-affairs/caseflow/issues/11533)_
```
1 | |
2 | ||||||||
3 | |||
5 | 
8 | 
```

Believe this could be a 1 if we simply aren't updating the total number of tasks and pages on the front end when we get tasks from the pagination API. Could be a 3 if the above reasons are not the cause of the problem.

hschallhorn:
Closed in #12213

hschallhorn:
Closed in #12213

PR #12181, opened 2019-09-20
[WIP] Add Pagination to Assign Hearings Page

### LINKED ISSUE #11757: Add Pagination to Assign Hearings

lomaxap:
resolves #11757


app/controllers/organizations/task_pages_controller.rb
```diff
@@ -25,7 +25,9 @@ class Organizations::TaskPagesController < OrganizationsController
 
   def index
     render json: {
-      tasks: json_tasks(paged_tasks)
+      tasks: json_tasks(task_pager.paged_tasks),
+      task_page_count: task_pager.task_page_count,
+      total_task_count: task_pager.total_task_count
```
lomaxap: @hschallhorn I'm adding 2 new properties to this response. This will allow me to use the QueueTable without having to get the QueueConfig data from a separate request (there is only ever one configuration / tab configs for this page).

client/app/queue/QueueTable.jsx
```diff
@@ -231,6 +231,10 @@ export default class QueueTable extends React.PureComponent {
       this.setState({ cachedTasks: { ...this.state.cachedTasks,
         [this.requestUrl()]: this.props.rowObjects } });
     }
+
+    if (this.props.eager) {
+      this.requestTasks();
+    }
```
lomaxap: also note i'm adding a prop the `QueueTable` that loads tasks after mount

app/controllers/organizations/task_pages_controller.rb
hschallhorn: I believe we were going to have to start doing this anyway as we think it's what's causing the issue in #12088, so add away!

pkarman:
@ferristseng @hschallhorn should this PR stay alive?

hschallhorn:
Oh boy. No commits since sept? Don't think this would be worthwhile to keep open. A lot could have changed in that time.

ferristseng:
We can close this PR. @rubaiyat22 or I are planning on picking this up, and will create a new branch

PR #12185, opened 2019-09-20
Add missing timestamp columns

### LINKED ISSUE #9793: Add created_at and updated_at to active models
If a table doesn't have updated_at and created_at already, add those fields to our tables

### LINKED ISSUE #12070: Add updated_at to the cached_appeal_attributes table to inform bug investigations
While investigating the job failing to run in #12069 we realized it would be nice to have an `updated_at` and `created_at`  value for the rows, especially since some rows get touched twice per run of the job!

This ticket is to add that column & make sure it gets updated appropriately by the job.

pkarman:
Resolves #9793 & Resolves #12070 

### Description
All rails tables should have timestamps.

*Schema Changes Only*

* [x] Column comments updated


pkarman:
Auto generated with:

```ruby
ApplicationRecord.descendants.reject(&:abstract_class?).reject { |k| k.ancestors.include?(VACOLS::Record) }.map { |klass| klass.table_name unless klass.new.attributes.include?("updated_at") }.compact.uniq.each { |t| puts "add_column :#{t}, :updated_at, :datetime" }
```

app/models/tasks/task_business_payload.rb
```diff
@@ -1,5 +0,0 @@
-# frozen_string_literal: true
-
-class TaskBusinessPayload < ApplicationRecord
```
lomky: A bit off that I see the migration creating this table, but not how it got removed from the schema?a

app/models/distribution.rb
```diff
@@ -53,10 +53,6 @@ def distribute!
 
   private
 
-  def attributes
```
lomky: This deletion is making distribution controller throw a fit without a direct call. Is it possible we were offering this info externally, even though we weren't invoking it directly?

app/models/tasks/task_business_payload.rb
pkarman: db/migrate/20190211164106_remove_business_payload_table.rb

app/models/distribution.rb
pkarman: we were overriding a built-in public method and making it private, seemingly to get the kind of json serialization we wanted. Instead, I added an actual Serializer class.

Issue #12167, opened 2019-09-19
[Tech Spec] Task event history


pkarman:
## Problem

The Bat team often finds itself trying to debug the state of an Appeal or LegacyAppeal, whether it's part of [Automatic Case Distribution](#11713) or in response to some support request. While at any given moment we can see the state of an Appeal through viewing the structure of its tasks, we cannot easily see the chronology of how the tasks arrived at their current state: who triggered what events at what time to result in their current status and assignment and individual column values.

## Proposal

First proposed by @lowellrex in #7254 we should track the event history of each `Task` record as an append-only log. This would allow for an audit trail of actions on each task over time.

The [PaperTrail gem](https://github.com/paper-trail-gem/paper_trail) provides this functionality. As an MVP, we could enable it for our `Task` model to start tracking changes to our `tasks` table.

## Implementation

Enable the PaperTrail gem per its README, creating a new `versions` table and configuring our `Task` model.

Add a `.history` method to both the `Appeal` and `Legacy` models.

```ruby
def history
  AppealTaskHistory.new(appeal: self)
end
```

Create an `AppealTaskHistory` service class:

```ruby
class AppealTaskHistory
  def initialize(appeal:)
    @appeal = appeal
  end

  def events
    @events ||= build_events
  end

  private

  attr_reader :appeal

  def build_events
    # iterate through appeal.tasks.versions and build a single array
    # of changes in reverse chronological order (most recent first).
    # returns that array as TaskEvent objects
  end
end
```

The `TaskEvent` model class is a simple wrapper/delegator class around `PaperTrail::Version`.

```ruby
class TaskEvent
  include ActiveModel::Model

  attr_accessor :version

  def diff
    # see https://github.com/paper-trail-gem/paper_trail#3c-diffing-versions
    version.changeset
  end

  def who
    version.whodunnit
  end

  def when
    version.created_at
  end

  def summary
    "[#{when}] [#{who}]\n#{diff.pretty_inspect}"
  end
end
```

Then from the console, example usage like:

```ruby
> appeal.history.each { |event| puts event.summary }
```

## Future

Should this MVP prove useful, appeal history could be exposed in the app as a simple changelog.

## References

Similar pattern developed in the 18F C2 app (also a workflow tracker Rails app). E.g. https://github.com/18F/C2/pull/1180

pkarman:
verified this is working in production as deployed.

PR #12192, opened 2019-09-24
Enable paper_trail for Tasks

pkarman:
connects #12167 

### Description
Enables the paper_trail gem for Task model.

Note we use version 8 so must use the docs at
https://github.com/paper-trail-gem/paper_trail/blob/v8.1.2/README.md

*Schema Changes Only*

* [ ] Column comments updated

### DIFF
--- a/config/initializers/paper_trail.rb
+++ b/config/initializers/paper_trail.rb
@@ -8,3 +8,22 @@ module PaperTrail
     end
   end
 end
+
+# make sure whodunnit is set in console
+Rails.application.configure do
+  console do
+    PaperTrail.whodunnit = ->() {
+      @paper_trail_whodunnit ||= (
+        until RequestStore[:current_user].present? do
+          puts "=" * 80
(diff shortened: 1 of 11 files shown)


config/initializers/paper_trail.rb
```diff
@@ -8,3 +8,22 @@ def user
     end
   end
 end
+
+# make sure whodunnit is set in console
+Rails.application.configure do
```
lomky: Do we have the option here to record who actually did the thing by interrogating Session's `global_admin`, which stores the user that is impersonating, if any.

See [user controller](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/test/users_controller.rb#L148)

pkarman: this code is just for the Rails console. The controller code looks at `current_user` which is going to record whomever that is. So if some is impersonating, it will look like the impersonated person. I think that's what we want, with the understanding that if you are impersonating another user, you really shouldn't be changing data. That's pretty basic sudo protocol.

Issue #12150, opened 2019-09-17
Monthly metrics report job


pkarman:
A CaseflowJob run monthly that sends a message to Slack with useful metrics for reporting up to OIT. See #11994 and #11198 for details.

## AC

Send 2 sets of numbers via Slack message to the #appeals-job-alerts channel.

* number of Appeals established during the previous month.
* using ClaimReviewAsyncStatsReporter, number of claim reviews established within 7 days (count + %), plus total, cancelled, processed

PR #12193, opened 2019-09-24
Monthly metrics job

pkarman:
connects #12150 

### Description
The start of monthly OIT metrics


app/services/claim_review_async_stats_reporter.rb
```diff
@@ -69,6 +96,14 @@ def build
   # rubocop:enable Metrics/AbcSize
   # rubocop:enable Metrics/MethodLength
 
+  def established_within_seven_days(completion_times)
+    completion_times.select { |span| span.fdiv(86_400).to_i < 7 }.count
+  end
+
+  def percent_established_within_seven_days(completion_times)
+    ((established_within_seven_days(completion_times) / completion_times.count.to_f) * 100).round(2)
```
alisan16: Is this meant to not include claims that have not yet completed or will not complete? I see in the test spec that the cancelled claims are not included in the denominator here, is that desired behavior?

app/jobs/monthly_metrics_report_job.rb
```diff
@@ -0,0 +1,45 @@
...
+    slack_service.send_notification(msg, self.class.to_s)
+  end
+
+  # rubocop:disable Metrics/LineLength
+  def build_report(appeals, async_stats)
+    sc_stats = async_stats.stats[:supplemental_claims]
+    hlr_stats = async_stats.stats[:higher_level_reviews]
+    report = []
+    report << "Monthly report #{start_date} to #{end_date}"
+    report << "Appeals established within 7 days: #{appeals} (100%)"
```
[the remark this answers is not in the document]
pkarman: just for parity with the other 2 claim review types. Appeals are always established immediately (no async to VBMS).

app/services/claim_review_async_stats_reporter.rb
pkarman: yes. if they are cancelled, that means they will never be established. What we want to measure is how many got "stuck" and took longer than 7 days. The cancelled ones are often cancelled long after 7 days, so they drive the number up but not in a helpful way, since they reflect intakes that will need to be redone. The number cancelled over time is very very small.

Issue #12003, opened 2019-09-06
Investigate Number of LegacyHearings without VACOLS records


lomaxap:

jimruggiero:
@lomaxap @pshahVA  Can one of you please add a user story and AC's (if appropriate) to this ticket?

pshahVA:
@ferristseng , do you have any context?

ferristseng:
Yes, this has caused a lot of issues while I was on batteam. This was happening when people were deleting hearings in VACOLS. The `LegacyHearing` instances we had would get orphaned, and the application would error out. I don't know if there's a good fix for this outside of making sure that people don't delete hearings in VACOLS.

ferristseng:
Related issues that deal with a similar problem with LegacyAppeal: #12791, #12810
PR to add job to discover LegacyAppeals in the dangling state: #13501

tomas-nava:
related #13208

PR #12195, opened 2019-09-24
Hearing Task Association should only return a hearing if the vacols record exists

lomaxap:
Related to #12003 

### Description

  * Refactor duplicated code for hearing task association to `HasHearingTask` concern
  * Refactor hearings-related code from `task.rb` to separate module
  * Add `LegacyHearing#vacols_hearing_exists?`
  * Update `most_recent_closed_hearing_on_appeal` to be aware of whether or not the hearings record is orphaned or not


pkarman:
@ferristseng assigned this one to you to figure out if we should patch and proceed.

PR #12196, opened 2019-09-24
Add Hearings Page: Screen Reader Modal Navigation Fix

### LINKED ISSUE #12121: Add Hearings Page: content behind modal is read by a screen reader, but shouldn't be.
If a user selects the Add hearing Date element, a modal window is displayed visually. Page content can be accessed by assistive technology users when this modal window is displayed. Visually users know a modal is displayed and cannot access other features in the background while this window has focus. (New: 08/27/2019) you can see in the screen shot below that focus is on the “digital service at VA” verses in the modal box

Suggested fix: https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html

aplybeah:
Resolves #12121 

This fix selects the first focusable element and the last selectable element in the modal and cycles the user through each option when they press `Tab`.

In addition, an extra focus target has been added to `compnentWillUnmount`. This will place the focus back on to the button that triggered the modal to open instead of navigating to `Digital Services at VA` at the bottom of the page.


client/app/components/Modal.jsx
```diff
@@ -14,20 +14,24 @@ export default class Modal extends React.Component {
     super(props);
     this.buttonIdPrefix = `${this.props.title.replace(/\s/g, '-')}-button-id-`;
   }
-
   handleTab = (event) => {
     let lastButtonId = this.buttonIdPrefix + (this.props.buttons.length - 1);
     let firstButton = document.getElementById(`${this.buttonIdPrefix}close`);
     let lastButton = document.getElementById(lastButtonId);
+    // a more specific selector
+    let endButton = document.querySelectorAll('#modal_id .usa-button')[1];
```
lomaxap: are we sure that buttons will always have a `usa-button` class? Since this components [lets you pass in your own button components or configuration through props](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/components/Modal.jsx#L83-L109), we don't necessarily know what classes they will have. It just so happens most people have been adding `usa-button` to the props they pass, but I found at least one instance where no `usa-button` class was given (see [here](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/intake/components/RemoveIssueModal.jsx#L43)) 

Also didn't realize we were [generating a button id](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/components/Modal.jsx#L77) so I think you can just change this to
``let endButton = document.querySelector(`#modal_id #${lastButtonId}`)``

Issue #11811, opened 2019-08-16
Tech Spec: Handle tasks assigned to users that become inactive or are removed from organizations


hschallhorn:
As noted in #11631, when a user becomes inactive or when a user is removed from an organization, there may be tasks assigned to them left in limbo. Currently, a team admin or caseflow support need to be notified of a user's inactive status to manually reassign or cancel all tasks in limbo. This ticket exists to tech spec the automation for handling these tasks

## Blocked by 🚳 
 - [x] Add active status for users #11810

## AC
- [ ]  When a user is removed from an organization or marked as inactive, how do we automatically cancel or reassign the tasks assigned to them
  - Default case?
  - block with message if there are open tasks assigned to them?

lowellrex:
_[What is this graph?](https://github.com/department-of-veterans-affairs/caseflow/issues/11533)_
```
      1 | |
      2 | ||
      3 | |||||
      5 | ||||
      8 | 
```

Will likely need input from design since there will be enough edge cases that user interaction will often be required.

Estimating this at 3. Possible 1 because the decision about how we want to handle this could be simple (always assign to team admin when somebody leaves). Possible 2 because even if the implementation is simple we will need exhaustive test coverage. Possible 5 because this may require changes to the task model (new status? new workflow?) as a result of design and user research considerations.

Will need release notes.

lomky:
Morphed into Tech Spec task for next sprint, in collab with Product design

lomky:
@lowellrex @hschallhorn I see a finished PR for the tech spec and an issue for implementation spun out from there. Is this task complete?

lowellrex:
I think the only thing left is to connect the script (added in #12483) with the front-end that removes folks and a back-end hook (or regular job monitor?) that re-assigns tasks when somebody leaves an organization. Do we have a ticket written to do this work?

Issue #12173, opened 2019-09-19
Reassign ColocatedTasks assigned to non VLJ support team members


hschallhorn:
VSCLHALL and ADJEWILH were added to the Colocated org in Caseflow although they do not work for the VLJ support team at the board. This has left them with automatically assigned `ColocatedTask`s due to the fact that tasks assigned to the colocated org get automatically assigned to anyone in the organization.

## Blocked By
- [x] These non team members should be removed from the Colocated org in the teams management page by one of the VLJ Support team admins. (@stwards11 should notify when this has happened)

## AC
- [ ] All `ColocatedTask`s previously assigned to `VSCLHALL` and `ADJEWILH` are reassigned to other members of the colocated staff.

kevmo:
1 | |||||||||
2 | |||
3 | 
5 |
8 |

1.

hschallhorn:
See [this tech spec](https://github.com/department-of-veterans-affairs/caseflow/pull/12197/files#diff-33a91cd5134bf0c5a4522e456e900342R38) for some guidance. Buddy up on this!

hschallhorn:
- [ ] Start with assigned and in_progress (active) colocated tasks
```ruby
VSCLHALL = User.find_by(css_id: "VSCLHALL")
active_tasks = ColocatedTask.active.where(assigned_to: VSCLHALL)

# Let's test on one task
task = active_tasks.first
puts task.reload.appeal.structure_render(:id, :assigned_to_id, :status)
# LegacyAppeal 884327
# ├── RootTask 305239, 2, assigned
# └── ScheduleHearingColocatedTask 308530, 23, on_hold
#     └── ScheduleHearingColocatedTask 308531, 10425, assigned

task.parent.assigned_to.next_assignee
# <User id: 7798, ...
# This is who we would assume the next task will be assigned to

# Alright, let's test the reassigning!
task.reassign({ assigned_to_type: User.name, assigned_to_id: task.parent.assigned_to.next_assignee.id }, task.assigned_by)
puts task.reload.appeal.structure_render(:id, :assigned_to_id, :status)
# LegacyAppeal 884327
# ├── RootTask 305239, 2, assigned
# └── ScheduleHearingColocatedTask 308530, 23, on_hold
#     ├── ScheduleHearingColocatedTask 308531, 10425, cancelled
#     └── ScheduleHearingColocatedTask 456129, 7798, assigned

# Looks good, let's test that next assignee was incremented
[...]

PR #12197, opened 2019-09-24
[Tech spec] Inactive user task re-assignment

lowellrex:
Connects #11811.

## Work remaining to be done:
* [ ] Determine which approach(es) we want to pursue
* [ ] Write tickets to do the implement the chosen approach(es)


docs/11811_tech_spec.md
```diff
@@ -0,0 +1,309 @@
...
+    - Expect that determination to be made by the product team in consultation with Board employees, as described in [this Slack post by LP](https://dsva.slack.com/files/T03FECE8V/FN5GLN4ES?origin_team=T03FECE8V) (reproduced as a standalone document in this PR)
+
+## Assumptions
+
+* This document assumes that this process begins after the application has identified the inactive user (or user who just left an organization) whose open tasks we want to handle.
+* The existing reassign() function will satisfy the mechanical problem of reassigning tasks even when they are not the terminal node in a branch of the task tree. That is, if a user becomes inactive, tasks that are children of tasks assigned to the newly-inactive user will still be available for action.
+* The application does not care if the task's assignee is no longer a member of the organizationally-assigned parent task is assigned when we are re-assigning their tasks. That is, if we are re-assigning the inactive user Billie's `AojColocatedTask` which has a parent `AojColocatedTask` assigned to the `Colocated` organization, we do not care if Billie is still a member of the `Colocated` organization or not.
```
yoomlam: Line 22: Hard to parse that first sentence. In the second sentence, should the child and parent tasks be the same?

docs/11811_tech_spec.md
```diff
@@ -0,0 +1,309 @@
...
+     2 | PendingScanningVbmsColocatedTask                 | PendingScanningVbmsColocatedTask                 | Organization     | VSCLHALL
+     2 | ScheduleHearingColocatedTask                     | ScheduleHearingColocatedTask                     | Organization     | VSCLHALL
+     1 | PowerOfAttorneyRelatedMailTask                   | PowerOfAttorneyRelatedMailTask                   | Organization     | VSCLHALL
+     1 | TimedHoldTask                                    | OtherColocatedTask                               | User             | VSCLHALL
+     1 | TimedHoldTask                                    | StayedAppealColocatedTask                        | User             | VSCLHALL
+(37 rows)
+```
+
+## Footnotes
+
```
yoomlam: Are these task relationships enforced in the code to prevent future code changes from breaking these constraints?

docs/11811_tech_spec.md
```diff
@@ -0,0 +1,309 @@
...
+    if task.parent.assigned_to.is_a?(Organization)
+      task.parent.update!(status: Constants.TASK_STATUSES.assigned)
+      task.update!(status: Constants.TASK_STATUSES.cancelled)
+    end
+
+    ...
+  end
+end
+```
+
```
yoomlam: Without an understanding of task states, it doesn't make intuitive sense for the parent task to become `assigned`. What was the state before that? Perhaps `needsReview` or `subtasksChanged` conveys the need for reassignment of child tasks.

docs/11811_tech_spec.md
```diff
@@ -0,0 +1,309 @@
...
+      # Update the instructions with a note detailing what we've done.
+      task.update!(instructions: [task.instructions, "Re-assigning open task for inactive user"].flatten)
+    end
+  end
+end
+```
+
+## Additional considerations
+
+Additionally, it may be worthwhile to alert folks via email (or some means other than simply task re-assignment which relies on somebody checking a particular queue). Electing to send an email in addition to or in place of re-assigning a task does not change this implementation in any meaningful way since we will still need to identify the recipient of such an email (a `Organization`'s administrator, for instance) and will rely on the task tree to identify that recipient.
```
yoomlam: Any automated reassignments should trigger a notification to someone, depending on the impact of the reassignment (i.e., based on task type probably). That way, if the recipient wants to change the reassignment, they can.

hschallhorn:
Some thoughts that have spun out of actually doing some of the reassignments [here](https://github.com/department-of-veterans-affairs/caseflow/issues/12173#issuecomment-539197954):
1. next_assignee is S L O O O W W W. We opted to somewhat batch the reassignments by only fetching next_assignee once per assignee, but a batch reassignment function may be necessary
2. The reassign function will want a [`current_user`](https://github.com/department-of-veterans-affairs/caseflow/blob/5fc7d88144b0003e17daed97d1cacfd5977973e0/app/models/task.rb#L135) passed to it to set the [`assigned_by_id`](https://github.com/department-of-veterans-affairs/caseflow/blob/5fc7d88144b0003e17daed97d1cacfd5977973e0/app/models/task.rb#L470) of the new task, so that should be passed in as well rather than nil.

hschallhorn:
When talking to design on #12312, it was mentioned that it may be nice for there to be a delay before the actual reassignment happens. Maybe we can put this in a job to give users some time to undo the status change if need be

docs/11811_tech_spec.md
lowellrex: Great catch. Fixing now.

docs/11811_tech_spec.md
lowellrex: Definitely agree that this particular scenario needs refinement. We are relying on somebody (a team administrator perhaps) monitoring the team's queue to notice that a new case has appeared in that queue. Assuming that somebody is monitoring a team queue may prove to be a bad assumption and we may need a better way to notify the team administrator that an appeal has popped into their queue.

I think it may be beyond the scope of this ticket, but perhaps we can add a [data integrity check](https://github.com/department-of-veterans-affairs/caseflow/blob/0170d3e7771d942d2ce788e02e62b20415b0a592/app/jobs/data_integrity_checks_job.rb#L3) that alerts us (or team admins?) when an appeal has not changed status in some number of days (appeals with `JudgeAssignTask`s that have not yet been dispatched, where the most recent `updated_at` value for that appeal's tasks is some number of days ago)?

lowellrex: Ticketed: https://github.com/department-of-veterans-affairs/caseflow/issues/12395

lowellrex:
Created #12394 as implementation ticket for this tech spec so I'm closing it out.

Issue #11401, opened 2019-07-12
Set correct contention type when creating contention


leikkisa:
### Story
When we create contentions in VBMS, we hard code the contention type to "REP".  Users are then expected to update the contention type when they're working on the claim in VBMS.

VBMS added new contention types on August 10 so that we can now set it when we create the contentions.

### Acceptance criteria
1. When a contention is created for a request issue on an HLR, the contention type is set to "HLR"
1. When it is on an SC, the contention type is set to "SUP"

### Technical notes
This will require multiple PRs.
1. The first PR will be in the connect_vbms repo. For create_contentions, we should add the ability to send an optional contention type to the "contentions" object. If it is not present, we will default to "REP".  This should be tested against VBMS UAT before merging.
1. Then the caseflow repo can be updated. This should include updating the connect_vbms gem, and setting the contention type for any contentions created on HLRs or SCs.  The primary work for this is in a method called "build_contentions".

PR #12203, opened 2019-09-25
Set correct contention type when creating contention

Sjones352:
Connects  #11401


app/models/request_issue.rb
```diff
@@ -278,6 +276,14 @@ def special_issues
     return specials unless specials.empty?
   end
 
+  def contention_type
+    # binding.pry
```
leikkisa: This can get cleaned up

client/constants/CONTENTION_TYPES.json
```diff
@@ -0,0 +1,5 @@
+{
+  "higher_level_review": "HLR",
+  "supplemental_claim": "SUP",
+  "rep": "REP"
```
leikkisa: Can we call this "default"? I think that's more descriptive

PR #12204, opened 2019-09-25
Daily check for stuck appeals

pkarman:
Data integrity check for AMA Appeals that appear "stuck":

* no tasks
* all tasks on hold
* dispatched but not yet closed


spec/services/stuck_appeals_checker_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    appeal
+  end
+
+  describe "#call" do
+    it "reports 3 appeals stuck" do
+      subject.call
+
+      expect(subject.report?).to eq(true)
+      expect(subject.report).to match(/^Stuck Appeals: 3 reported/)
+    end
```
kevmo: Good test.

app/queries/appeals_with_no_tasks_or_all_tasks_on_hold_query.rb
```diff
@@ -2,17 +2,43 @@
...
+      .where(type: "RootTask", status: on_hold))
+      .where(id: completed_dispatch_tasks("Appeal"))
+  end
+
   def stuck_appeals
     stuck_query("Appeal")
   end
 
+  def completed_dispatch_tasks(klass_name)
+    tasks_for(klass_name).where(type: %w[BvaDispatchTask QualityReviewTask], status: completed)
```
lowellrex: What do you think of referencing these names as methods of the class instead of string literals? (`type: [BvaDispatchTask.name, QualityReviewTask.name]`) If we ever changed the name of these tasks, the application would error at compile-time with the methods and would not ever error with these strings.

Less important with `Appeal` and `RootTask` since those are less likely to change, but I still think it's a handy practice to adopt (using methods instead of strings where possible).

lowellrex:
What do you think of adding a `.stuck?()` method on `Appeal` and `LegacyAppeal` that calls these functions so that we can interrogate individual appeals to determine whether they are stuck or not. Maybe something like the following:

```ruby
def ama_appeal_stuck?(id)
  (stuck_query(Appeal.name).where(appeal_type: Appeal.name, appeal_id: id).any? ||
    appeals_with_zero_tasks.where(appeal_type: Appeal.name, appeal_id: id).any? ||
    dispatched_appeals_on_hold.where(appeal_type: Appeal.name, appeal_id: id).any?)
end
```

Which `Appeal` could call:

```ruby
def stuck?
  AppealsWithNoTasksOrAllTasksOnHoldQuery.ama_appeal_stuck?(id)
end
```

PR #12205, opened 2019-09-25
Batch calls to users table

hschallhorn:
### Description
Removes cacophony of calls to find individual users when calling `find_all_attorneys` or `find_all_judges`

### Testing Plan
1. Checkout master
2. Run benchmark for old method and view said cacophony
```ruby
> require "benchmark"
> Benchmark.measure { AttorneyRepository.find_all_attorneys }
=> #<Benchmark::Tms:0x00007fc2abc1b7c0
 @cstime=0.0,
 @cutime=0.0,
 @label="",
 @real=2.404171000001952,
 @stime=0.13165599999999955,
 @total=1.6080920000000063,
 @utime=1.4764360000000067>
```
3. Checkout hschallhorn/speed-up-user-repos
4. Run benchmark for new method
```ruby
> reload!
> Benchmark.measure { AttorneyRepository.find_all_attorneys }
=> #<Benchmark::Tms:0x00007fc2b4d21aa8
 @cstime=0.0,
 @cutime=0.0,
 @label="",
 @real=0.2092659999616444,
 @stime=0.030889999999999418,
 @total=0.1086920000000049,
 @utime=0.07780200000000548>
```
5. Run all flake tests in dev to ensure they pass


app/repositories/attorney_repository.rb
```diff
@@ -2,9 +2,13 @@
...
-      User.find_by_css_id_or_create_with_default_station_id(record.sdomainid)
-    end
+    css_ids = VACOLS::Staff.where(sactive: "A").where.not(sattyid: nil).where.not(sdomainid: nil)
+      .pluck("UPPER(sdomainid)")
+
+    # Create users in caseflow table
+    new_user_css_ids = css_ids - User.where(css_id: css_ids).pluck(:css_id)
+    User.create(new_user_css_ids.map { |css_id| { css_id: css_id, station_id: User::BOARD_STATION_ID } })
+
+    User.where(css_id: css_ids)
```
hschallhorn: One thing to note is that in our dev environment we have a few repeat `sdomainid`s in our vacols records
```ruby
> staff = VACOLS::Staff.where(sactive: "A").where.not(sattyid: nil).where.not(sdomainid: nil)
> staff.order('count_all desc').group(:sdomainid).count
=> {"CSS_ID18"=>6,
    "CSS_ID22"=>6,
    "CSS_ID27"=>6,
    "CSS_ID11"=>6,
    "CSS_ID13"=>6,
    "CSS_ID14"=>6,
    "CSS_ID24"=>6,
    "CSS_ID16"=>6,
    "CSS_ID21"=>6,
    "CSS_ID12"=>6,
    "CSS_ID19"=>6,
    "CSS_ID25"=>6,
    "CSS_ID17"=>6,
    "CSS_ID15"=>6,
    "BVADSPINKA"=>1,
    ...}
> css_ids = staff.pluck("UPPER(sdomainid)")
> css_ids.count
=> 914
> css_ids.uniq.count
=> 844
```
This change will only return the 844 users in dev whereas we were returning 914 (6 of the same `User` for each duplicate `css_id`). All `sdomainid`s are unique in VACOLS in prod, so this is not something that will create user facing changes.

app/repositories/attorney_repository.rb
```diff
@@ -2,9 +2,13 @@
...
-    records = VACOLS::Staff.where(sactive: "A").where.not(sattyid: nil)
-    records.select(&:sdomainid).map do |record|
-      User.find_by_css_id_or_create_with_default_station_id(record.sdomainid)
-    end
+    css_ids = VACOLS::Staff.where(sactive: "A").where.not(sattyid: nil).where.not(sdomainid: nil)
+      .pluck("UPPER(sdomainid)")
+
+    # Create users in caseflow table
+    new_user_css_ids = css_ids - User.where(css_id: css_ids).pluck(:css_id)
+    User.create(new_user_css_ids.map { |css_id| { css_id: css_id, station_id: User::BOARD_STATION_ID } })
```
lowellrex: Can we move this function into the `User` model so that the `JudgeRepository` can reuse it as well? Maybe something like the following

```ruby
def self.batch_find_by_css_ids_or_create_with_default_station_id(css_ids)
  new_user_css_ids = css_ids - User.where(css_id: css_ids).pluck(:css_id)
  User.create(new_user_css_ids.map { |css_id| { css_id: css_id, station_id: User::BOARD_STATION_ID } }) if new_user_css_ids.any?
  User.where(css_id: css_ids)
end
```

hschallhorn: I like!

app/models/user.rb
```diff
@@ -412,6 +412,12 @@ def find_by_css_id_or_create_with_default_station_id(css_id)
       find_by_css_id(css_id) || User.create(css_id: css_id.upcase, station_id: BOARD_STATION_ID)
     end
 
+    def batch_find_by_css_id_or_create_with_default_station_id(css_ids)
+      new_user_css_ids = css_ids - User.where(css_id: css_ids).pluck(:css_id)
+      User.create(new_user_css_ids.map { |css_id| { css_id: css_id, station_id: User::BOARD_STATION_ID } })
+      User.where(css_id: css_ids)
+    end
```
lowellrex: Can we add the upper-casing logic we have in the `.pluck()` statements into this function? (something like `upper_css_ids = css_ids.map(&:upcase)` should work).

As this function is currently written if the input `css_ids` include CSS IDs with lowercase letters, we will [uppercase those CSS IDs before creating](https://github.com/department-of-veterans-affairs/caseflow/blob/907c5644469ae217746d82938ca0c8568e7db072/app/models/user.rb#L28) them in the database, but not return them because the upper-cased CSS ID does not match the one in the input `css_ids` list. Additionally, our where clause will not find `User` records for the same CSS ID if the input CSS ID is lowercase so the `create()` on the following line will fail.

Some tests to demonstrate:

```ruby
describe ".batch_find_by_css_id_or_create_with_default_station_id", focus: true do
  subject { User.batch_find_by_css_id_or_create_with_default_station_id(css_ids) }

  context "when the input list of CSS IDs includes a lowercase CSS ID" do
    let(:lowercase_css_id) { Generators::Random.from_set(("a".."z").to_a, 16) }
    let(:css_ids) { [lowercase_css_id] }

    context "when the User record for the lowercase CSS ID does not yet exist in the database" do
      it "returns the newly created User record for that CSS ID" do
        expect(subject.length).to eq(1)
      end
    end

    context "when the User record for the lowercase CSS ID already exists in the database" do
      before { User.create(css_id: lowercase_css_id, station_id: User::BOARD_STATION_ID) }
      it "returns the existing User record for that CSS ID" do
        expect(subject.length).to eq(1)
      end
    end
  end
end
```

hschallhorn: Good call!

app/repositories/user_repository.rb
```diff
@@ -63,14 +63,6 @@ def css_id_by_full_name(full_name)
...
     # :nocov:
 
-    def find_all_hearing_coordinators
-      coordinator_records = VACOLS::Staff.where(sdept: "HRG", sactive: "A")
-
-      coordinator_records.select(&:sdomainid).map do |record|
-        User.find_by_css_id_or_create_with_default_station_id(record.sdomainid)
-      end
-    end
-
```
hschallhorn: Was looking for other instances we could batch this function, and this whole one seems to be [unused elsewhere.](https://github.com/department-of-veterans-affairs/caseflow/search?q=find_all_hearing_coordinators&unscoped_q=find_all_hearing_coordinators)

app/models/user.rb
hschallhorn: All fixed! (unless that was a subtle hint to add in some tests for `batch_find_by_css_id_or_create_with_default_station_id` 😬

lowellrex: Yeah, can we just add these tests wholesale to the `user_spec`?

PR #12220, opened 2019-09-25
Intake alerts quiet after 4 hours

pkarman:
Help improve signal:noise ratio in #appeals-intake-alerts channel.

Also extends the re-try window for decision issue sync to a full month.

### DIFF
--- a/app/jobs/decision_review_process_job.rb
+++ b/app/jobs/decision_review_process_job.rb
@@ -37,7 +37,7 @@ class DecisionReviewProcessJob < CaseflowJob
   attr_reader :decision_review
 
   def ok_to_ping_sentry?
-    decision_review.sort_by_last_submitted_at > (Time.zone.now - 1.day)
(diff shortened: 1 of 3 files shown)


app/jobs/decision_review_process_job.rb
```diff
@@ -28,7 +34,13 @@ def perform(decision_review)
 
   private
 
-  def add_extra_context_to_sentry(decision_review)
+  attr_reader :decision_review
+
+  def ok_to_ping_sentry?
+    decision_review.sort_by_last_submitted_at > (Time.zone.now - 4.hours)
```
pkarman: 4 hours means it will alert at most 2 times and maybe only once.

leikkisa: I think last_submitted_at gets replaced for retries.  Does it only get replaced when a user restarts a job?  If the former is true, I don't think this will accomplish getting fewer Sentry errors.

PR #12226, opened 2019-09-26
Cache of hearing facility addresses

pkarman:
Hearings API relies on accurate street address. We have a bug where we are truncating street address in the `hearing_location.address` field. This change pulls the street address from a .json constants file instead.

### DIFF
--- a/app/models/hearing_location.rb
+++ b/app/models/hearing_location.rb
@@ -2,4 +2,9 @@
 
 class HearingLocation < ApplicationRecord
   belongs_to :hearing, polymorphic: true
+
+  def street_address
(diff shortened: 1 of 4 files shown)


ferristseng:
I think an empty string is causing it:

```
{"address_1"=>"<> Bay Pines Blvd.",
 "address_2"=>"",
 "address_3"=>nil,
 "city"=>"St. Petersburg",
 "state"=>"FL",
 "zip"=>"33744"}
```

pkarman:
good catch @ferristseng -- I added tests to catch blanks and nulls.

PR #12227, opened 2019-09-26
Fix formatting of full address

ferristseng:
https://dsva.slack.com/archives/C2ZAMLK88/p1569524988002100

### Description

AHL addresses were being stored incorrectly because of an issue that caused duplicate characters to be collapsed.

  * Adds tests to verify address formatting
  * Fix AHL address spacing issue and duplicate character issue


pkarman:
We should definitely have @lomaxap look at this before we merge. I suspect the `.squeeze` was trying to solve a specific problem that I'm not sure we understand.

pkarman:
@ferristseng the code looks good.

What I'm not sure about is how the `full_address` concatenation is being used, and whether the slight changes to the formatting (e.g. spaces between address lines) is significant or not. Let's wait for @lomaxap to weigh in.

Issue #12186, opened 2019-09-21
For DTA Supplemental claims, only send the review's decision issues in contestable issues


leikkisa:
### Story
For DTA Supplemental claims, request issues are automatically created based on the DTA'd decisions.  We want to allow users to edit the contention text on those automatically created request issues, but we do not want to allow them to add or remove issues.  However, in the case where a user is editing a DTA claim to make a correction on one of the decisions for that review, we want the user to be able to add those issues.

So instead of disabling the "Add issue" button while for editing DTA Supplemental Claims, we should only send the review's decision issues as contestable issues, and only show the Add Issue button if those exist.

### Acceptance criteria
- While editing a DTA claim, a user can only add that claim's decision issues
- Hide "Add issues" button if that is not an option
- Hide button that says "None of these match" on the first add issue modal.  If that is available, they could navigate to the "add nonrating" screen, which would allow the user to add a new request issue.

related
https://github.com/department-of-veterans-affairs/caseflow/issues/11011
https://github.com/department-of-veterans-affairs/caseflow/pull/12104

PR #12228, opened 2019-09-27
DTA corrections - prevent invalid new issues from being added

leikkisa:
connects #12186

### Description
When a user is editing a correction DTA claim, show that claim's decision issues as potential issues that can get added, but do not allow adding other issues. Hide the "Add issues" button if there are no contestable issues.


spec/feature/intake/end_product_correction_spec.rb
```diff
@@ -419,10 +419,35 @@
...
     context "when the end product is cleared" do
-      it "allows edit and shows the add issue button" do
-        visit edit_path
-        expect(page).to have_content("Edit Issues")
-        expect(page).to have_css("#button-add-issue")
+      context "when the review has no decision issues" do
+        it "does not allow the user to add issues" do
+          visit edit_path
+          expect(page).to have_content("Edit Issues")
+          expect(page).to_not have_css("#button-add-issue")
```
Sjones352: I see the add button. Should this be hidden
Screen Shot 2019-10-04 at 4 41 33 PM

Sjones352: This test fails on my local

PR #12240, opened 2019-09-27
Accessibility Fix for Dropdown Menus

### LINKED ISSUE #12125: Switch View button not focusable + Fix order of buttons that are focused upon clicking tab on Hearings schedule pages
After the Apply button the Switch view element does not receive focus. Focus skips to the Download Current View element. (New: 08/27/2019) The focus does shift around in the wrong order on this page in particular

- Switch View button is not being read by AT
- The order of which buttons are focused on while navigating the page via tab is strange and should be made more orderly (i.e. focus on ""Schedule Veterans"" before ""Build Schedule"" because it is visually to the left)"

aplybeah:
Resolves #12125 

This PR:
- adjusts the tab order of the `Schedule Veterans` and the `Build Schedule` buttons to match what is visually present
- Allows the `Switch View` button to be selected (via the Tab key)and be read by a screen reader
- Allows the items in the `Switch View` dropdown to be selected(via the Tab key) and be read by a screen reader
- Escape/Enter keys toggle the `Switch View` menu when the `Switch View` button has focus

Refactor Notes (Will be pursued at a later time):
- Following the Mac's command prompts for selecting options (eg. "clicking" on Complete Hearing Schedule without using a mouse) will change the page. As of right now, pressing the `Enter` key does not do the same thing.
- The focus in the dropdown options sort of "floats" in the options list after an option has been selected instead of returning back to the `Switch View` button


client/app/components/DropdownButton.jsx
```diff
@@ -38,14 +43,24 @@ export default class DropdownButton extends React.Component {
     }
   }
 
+  keyHandler = (event) => {
```
alisan16: You shouldn't need this - adding role="button" to the DropdownButton (see render()) allows the button to be toggled via Enter/Esc

client/app/components/DropdownButton.jsx
```diff
@@ -59,21 +74,21 @@ export default class DropdownButton extends React.Component {
...
         </li>)}
     </ul>;
-  }
+  };
 
   render() {
     const { label } = this.props;
 
-    return <div className="cf-dropdown" ref={this.setWrapperRef} {...dropdownBtnContainer}>
+    return <div className="cf-dropdown" ref={this.setWrapperRef} {...dropdownBtnContainer} aria-haspopup="true">
```
alisan16: I don't think you need aria-haspopup here. Just adding role="button" should work
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role

client/app/components/DropdownButton.jsx
```diff
@@ -38,14 +43,24 @@ export default class DropdownButton extends React.Component {
...
+  }
   onMenuClick = () => {
     this.setState((prevState) => ({
       menu: !prevState.menu
     }));
   };
 
   dropdownLink = (list) => {
-    return <Link className="usa-button-secondary usa-button"
+    return <Link className="usa-button-secondary usa-button" aria-activedescendant="cf-dropdown">
```
alisan16: I don't think this is the correct usage of aria-activedescendant. https://www.w3.org/TR/wai-aria/#aria-activedescendant seems to say that it's used to indicate that "cf-dropdown" is a descendent of the current element, but in this case, it should be reversed. "cf-dropdown" is the dropdown button that contains multiple dropdownLinks

client/app/components/DropdownButton.jsx
```diff
@@ -59,21 +74,21 @@ export default class DropdownButton extends React.Component {
   }
 
   dropdownButtonList = () => {
-    return <ul className="cf-dropdown-menu active" {...dropdownList}>
+    return <ul className="cf-dropdown-menu active" {...dropdownList} aria-expanded="true">
       {this.props.lists.map((list, index) =>
-        <li key={index}>
+        <li key={index} role="option" aria-selected="true" tabIndex="0" aria-activedescendant="cf-dropdown">
```
alisan16: Same comment about aria-activedescendant.
I don't think the items in the list are options, that usually refers to an option within a form (same with aria-selected="true"): https://www.w3schools.com/tags/tag_option.asp, whereas this button opens 2 different links that reload the table.
Do you need the tabIndex attribute here?

aplybeah:
## Updates to this PR:
I'm trying to adopt the principle of "No ARIA is better than Bad ARIA" in the future for working with accessibility features. 
_What this means to me:_
*  the focus should be on using as many native HTML elements as possible before resorting to `aria` tags. (i.e. be as semantic as possible)
* the primary use for `aria` tags should be to ensure screen readers give a descriptive readout (e.g. "List of three items, press Escape to close" vs "List item")

I've also done a lot more research on `aria` roles and what they do. Specifically with the `button` role one of biggest things I've noticed is that, [it doesn't assign keyboard functionality](https://stackoverflow.com/questions/57398579/when-and-when-not-to-use-role-button); [it just tells screen readers what the element should be able to function as](https://stackoverflow.com/questions/6562407/what-does-role-button-mean).

For this dropdown, I used the practices listed [here](https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html).

---

In the previous rendition of this PR, I was using a nested `<div>` and adding roles to it to make it a custom button. I believe the cause for that is [that some elements such as `button` make their children presentational](https://www.w3.org/TR/wai-aria-practices/#children_presentational). To get around this, I added the `aria-haspopup` and `aria-expanded` attributes. 
* `aria-haspopup` adds the descendants of the `<button>` attribute to the accessibility tree, meaning that the screen reader will now be able to read the links in the list
* `aria-expanded` is the attribute that will have the screen reader notify users that there was an expanded list opened, and gives a prompt for closing it.


### Pain Points:
* screen reader cursor won't return to the "switch view" button, but instead will stay in the spot where the list item was instead after an option was clicked.
* the above also happens when reversing the tab order after the dropdown list has been opened.

client/app/components/DropdownButton.jsx
```diff
@@ -50,7 +60,7 @@ export default class DropdownButton extends React.Component {
   }
 
   dropdownAction = (list) => {
-    return <a onClick={() => {
+    return <a href="#" onClick={() => {
```
alisan16: Right now the links are reading as "visited, link, ...". You can resolve this by making a custom href, something like `href={"#" + list.value}` so that the screenreader doesn't always think the link has already been visited.

Screen Shot 2019-10-10 at 12 11 48 PM

client/app/components/DropdownButton.jsx
```diff
@@ -24,17 +24,27 @@ export default class DropdownButton extends React.Component {
...
+    document.removeEventListener('keydown', this.onClickOutside);
+  }
   setWrapperRef = (node) => this.wrapperRef = node
 
   onClickOutside = (event) => {
     if (this.wrapperRef && !this.wrapperRef.contains(event.target) && this.state.menu) {
       this.setState({
         menu: false
       });
+    } else if (event.key === 'Escape') {
```
alisan16: What is this for?

aplybeah: @alisan16  Ordinarily, pressing the escape key doesn't close the dropdown. This adds a listener for the escape key so that functionality returns.

PR #12252, opened 2019-09-30
Async job notes

### LINKED ISSUE #12214: [Tech Spec] Async job notes
## Problem

When async claim establishment jobs (for HLR, SC and RequestIssueUpdate) fail with errors, we usually need to create incident tickets in YourIT. We need a reliable way of communicating that a stuck async job is being addressed in an upstream system, so that anyone looking at the job in Caseflow can know that (a) someone has seen the problem, (b) they have created a ticket and (c) what the URL of the ticket is so that they can follow up with our partner services.

## Proposal

Create a new table `job_notes` with a schema like:

```ruby
    create_table :job_notes do |t|
      t.belongs_to :job, polymorphic: true, null: false, comment: "The job to which the note applies"
      t.belongs_to :user, null: false, comment: "The user who created the note"
      t.timestamps null: false, comment: "Default created_at/updated_at"
      t.text :note, null: false, comment: "The note"
      t.boolean :send_to_intake_user, default: false, comment: "Should the note trigger a message to the job intake user"
    end
```

Add a form to the individual async job page that allows any user able to view that page to also add a note.

The form has a checkbox with the label "Send to :username:" where the username is the intake user.

* Notes cannot be edited.
* Notes cannot be deleted.
* Notes may optionally trigger a `Message` to the user who created the Intake (job).

pkarman:
Resolves #12214 

### Description

Adds job notes for ad hoc annotation of async jobs.

This is intended for tracking errors and creating xrefs to GH tickets, YourIT tickets, etc.

*Schema Changes Only*

* [x] Column comments updated

### DIFF
--- a/app/controllers/asyncable_jobs_controller.rb
+++ b/app/controllers/asyncable_jobs_controller.rb
@@ -26,6 +26,16 @@ class AsyncableJobsController < ApplicationController
     render json: job.asyncable_ui_hash
   end
 
+  def add_note
+    job_note = JobNote.create!(
+      job: job,
+      user: current_user,
+      note: allowed_params[:note],
+      send_to_intake_user: allowed_params[:send_to_intake_user]
+    )
+    render json: job_note.ui_hash
+  end
+
   private
 
   helper_method :jobs, :job, :allowed_params, :pagination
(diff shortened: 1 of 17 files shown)


pkarman:
@leikkisa good bug find. Fixed.

Checkbox commented out till the server supports it.

PR #12258, opened 2019-10-01
Return queue config to the front end for user queues

### LINKED ISSUE #11697: Return QueueConfig to front-end for individual queues
In preparation for defining individual queue configurations on the back-end, we should start returning the `QueueConfig` to the front-end [from `TasksController.index`](https://github.com/department-of-veterans-affairs/caseflow/blob/99edafe333201710f276317da22d05bc66df5acd/app/controllers/tasks_controller.rb#L39) (similar to [how we do for organizations](https://github.com/department-of-veterans-affairs/caseflow/blob/99edafe333201710f276317da22d05bc66df5acd/app/controllers/organizations/tasks_controller.rb#L13)) so that we can build the tabs dynamically on the front-end down the line.

## 🚫 Blocked by
* [x] Extend QueueConfig to work for individual (non-org) users #11696

## Acceptance criteria
* [ ] Return `QueueConfig` to front-end for individual task queues

hschallhorn:
Resolves #11697

### Description
Returns queue config to the front end for user queues

Screen Shot 2019-10-01 at 10 58 22 AM

### Acceptance Criteria
- [ ] Returns queue config to the front end for user queues

### Testing Plan
1. Go to a user's queue
2. Ensure queue config is returned to the front end for user's queue

### DIFF
--- a/app/controllers/tasks_controller.rb
+++ b/app/controllers/tasks_controller.rb
@@ -39,7 +39,7 @@ class TasksController < ApplicationController
   #      GET /tasks?user_id=xxx&role=judge
   def index
     tasks = QueueForRole.new(user_role).create(user: user).tasks
-    render json: { tasks: json_tasks(tasks) }
+    render json: { tasks: json_tasks(tasks), queue_config: queue_config }
   end
 
   # To create colocated task
@@ -152,6 +152,10 @@ class TasksController < ApplicationController
   end
 
   private
+
+  def queue_config
+    QueueConfig.new(assignee: user).to_hash_for_user(current_user)
+  end
 
   def verify_task_access
     if current_user.vso_employee? && task_classes.exclude?(InformalHearingPresentationTask.name.to_sym)
--- a/app/models/queue_config.rb
+++ b/app/models/queue_config.rb
@@ -18,10 +18,9 @@ class QueueConfig
 
   def to_hash_for_user(user)
     table_title = COPY::COLOCATED_QUEUE_PAGE_TABLE_TITLE
-    active_tab = Constants.QUEUE_CONFIG.UNASSIGNED_TASKS_TAB_NAME
     {
       table_title: assignee_is_org? ? format(COPY::ORGANIZATION_QUEUE_TABLE_TITLE, assignee.name) : table_title,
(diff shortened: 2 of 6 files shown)


app/models/queue_config.rb
```diff
@@ -18,9 +18,10 @@ def initialize(args)
 
   def to_hash_for_user(user)
     table_title = COPY::COLOCATED_QUEUE_PAGE_TABLE_TITLE
+    active_tab = Constants.QUEUE_CONFIG.UNASSIGNED_TASKS_TAB_NAME
     {
       table_title: assignee_is_org? ? format(COPY::ORGANIZATION_QUEUE_TABLE_TITLE, assignee.name) : table_title,
-      active_tab: Constants.QUEUE_CONFIG.UNASSIGNED_TASKS_TAB_NAME,
+      active_tab: assignee_is_org? ? active_tab : Constants.QUEUE_CONFIG.ASSIGNED_TASKS_TAB_NAME,
```
kevmo: This ternary using a variable that points to one of our Constants strings in the truthy case vs directly calling the Constant string in the falsy case could be made a little more readable by using a constant style for the return values.

app/controllers/tasks_controller.rb
```diff
@@ -39,7 +39,7 @@ def set_application
   #      GET /tasks?user_id=xxx&role=judge
   def index
     tasks = QueueForRole.new(user_role).create(user: user).tasks
-    render json: { tasks: json_tasks(tasks) }
+    render json: { tasks: json_tasks(tasks), queue_config: queue_config }
```
kevmo: Would be good to add a test to make sure this controller returns `queue_config` (and perhaps that it returns the correct values for given inputs).

app/models/queue_config.rb
kevmo:
``` 
active_unassigned_tab = Constants.QUEUE_CONFIG.UNASSIGNED_TASKS_TAB_NAME
active_assigned_tab = Constants.QUEUE_CONFIG.ASSIGNED_TASKS_TAB_NAME
```
then in the ternary that is in the hash: 
`active_tab: assignee_is_org? ? active_unassigned_tab : active_assigned_tab`

PR #12281, opened 2019-10-02
Appeals API intake status show

edmangimelli:
Implement the show endpoint for IntakeStatuses

resolves https://github.com/department-of-veterans-affairs/caseflow/issues/12292

An IntakeStatus (the asyncable status of an intake's DecisionReview) is returned upon successful creation of a DecisionReview (an intake) (already implemented), and also when _getting_ the intake_statuses endpoint with a DecisionReview's uuid (what this PR is implementing).

The the parameters/shape of this endpoint are already described in decision_review.yaml
See issue for instructions for viewing this openapi spec


app/controllers/api/v3/decision_review/higher_level_reviews_controller.rb
```diff
@@ -1,31 +1,54 @@
...
     @processor ||= Api::V3::DecisionReview::HigherLevelReviewIntakeProcessor.new(params, User.api_user)
   end
 
+  def intake_status
+    @intake_status ||= Api::V3::DecisionReview::IntakeStatus.new(processor.intake)
+  end
+
+  # following https://jsonapi.org/recommendations/#asynchronous-processing
+  # the first status returned is 202, not 200
+  def creation_http_status
```
erluti: If the IntakeStatus model had a parameter in the constructor for "is_new", then when then the helper method above could set it to true and then this logic could exist in the model

edmangimelli: i like it 👍

app/services/api/v3/decision_review/higher_level_review_intake_processor.rb
```diff
@@ -5,5 +5,7 @@ def initialize(params, user)
     super(params: params, user: user, form_type: "higher_level_review")
   end
 
-  alias higher_level_review detail
+  def higher_level_review
+    @intake&.detail
```
pkarman: typically we avoid using instance variables directly and instead define reader methods.

```ruby
def higher_level_review
  intake&.detail
end

attr_reader :intake
```

app/services/api/v3/decision_review/intake_status.rb
```diff
@@ -1,25 +1,55 @@
...
+    }
+  end
+
+  def no_decision_review_json
+    {
+      errors: [
+        {
+          status: NO_DECISION_REVIEW_HTTP_STATUS,
+          code: :intake_not_connected_to_a_decision_review,
+          title: "This intake is not connected to a decision review."
```
pkarman: Could we put this string in a constants file? `client/constants/API_V3.json` e.g.

spec/requests/api/v3/decision_review/higher_level_reviews_controller_spec.rb
```diff
@@ -120,7 +120,10 @@ def post_params
   describe "#create" do
     describe "general cases" do
       it "should return a 202 on success" do
-        allow(User).to receive(:api_user) { build(:user) }
+        allow(User).to receive(:api_user) do
```
pkarman: this repeated code seems ripe for a little method

spec/requests/api/v3/decision_review/intake_statuses_controller_spec.rb
```diff
@@ -0,0 +1,222 @@
...
+        detail: detail
+      )
+    end
+  end
+
+  let(:new_higher_level_review) do
+    HigherLevelReview.create!(veteran_file_number: veteran_file_number)
+  end
+
+  # mocked intake and decision review (no DB)
```
pkarman: why use database_cleaner if you are going to mock the db?

pkarman: and why mock the db with fake classes? I'm confused as to what that gains us.

spec/services/api/v3/decision_review/intake_status_spec.rb
```diff
@@ -1,29 +1,109 @@
...
 
 require "support/database_cleaner"
+require "rails_helper"
 
 context Api::V3::DecisionReview::IntakeStatus, :postgres do
-  let(:intake) do
-    intake = Intake.build(
+  let(:veteran_file_number) { "123456789" }
+
+  def fake_intake(detail = nil)
```
pkarman: I'm confused the `fake_` naming there -- in what way are these objects "fake"?

app/services/api/v3/decision_review/higher_level_review_intake_processor.rb
edmangimelli: oversight 👍

app/services/api/v3/decision_review/intake_status.rb
edmangimelli: Would you prefer a constants file, or should i move that hash up to the top of the class with the other constants?

erluti: We haven't used a constants file yet.  Is it only for strings?  What's the advantage to having one?  Is it worth creating it for this single string?

spec/requests/api/v3/decision_review/higher_level_reviews_controller_spec.rb
edmangimelli: 👍

spec/requests/api/v3/decision_review/intake_statuses_controller_spec.rb
edmangimelli: The first test of the #show method is done just fine with the mocked DB, and that's what I do (the "should return...NOT_SUBMITTED..." test).

For the second test of #show ("...should return...SUBMITTED..."), I needed to mock a DecisionReview that had asyncable_status of :submitted. Whenever I mocked this with allow, it seemed like, at some point in the code (i _think_ deeper than my code), my DB object was getting reloaded, which was getting rid of my mocked methods. That is, i would mock asyncable status, i could verify that the mock was working, but then, when I made my request, and the controller would do its fetching, I would lose my mocks. So, I opted to instead forgo the mocked DB and completely mock an intake and DecisionReview with plain old classes.

spec/services/api/v3/decision_review/intake_status_spec.rb
edmangimelli: 👍 agreed. These objects certainly aren't fake in the same way as the ones I mocked in the controller speck

app/services/api/v3/decision_review/intake_status.rb
pkarman: The constants files for the rest of the app allow strings to be defined as constants and shared between front end and back end code. This is particularly useful for testing the UI.

For the API it is less important, but I'm leaning toward yes, using a constants file just for consistency. I could be convinced otherwise, since I'm not sure other API code uses that convention.

spec/requests/api/v3/decision_review/intake_statuses_controller_spec.rb
pkarman: I guess I'm saying that you can create AR objects and just not write them to the db. That's what `build` is for (as opposed to `create`).

edmangimelli: Yeah, but the controller is expecting to be able to look them up in the DB.  But.... I did end up mocking my lookup method..

pkarman: If the controller tests need the data in the db, then why bother with mocking at all? I.e. we have 1000s of tests that write data to the db in order to test controller and model actions. Why not here?

```
create(:higher_level_review, establishment_submitted_at: Time.zone.now)
```

pkarman: Or even better:

```ruby
create(:higher_level_review, :requires_processing)
```

edmangimelli: >we have 1000s of tests that write data to the db in order to test controller and model actions. Why not here?

>Whenever I mocked this with allow, it seemed like, at some point ... my DB object was getting reloaded, which was getting rid of my mocked methods.

but, I'll give the factory 👍 and report back

app/services/api/v3/decision_review/intake_status.rb
```diff
@@ -1,25 +1,65 @@
 # frozen_string_literal: true
 
 class Api::V3::DecisionReview::IntakeStatus
```
erluti: Should this move to `/services/api/v3`?

spec/requests/api/v3/decision_review/intake_statuses_controller_spec.rb
edmangimelli: worked 👍 thanks!

app/services/api/v3/decision_review/intake_status.rb
erluti: It doesn't look like the other APIs are using the constants files.  Since this text isn't going to be shared anywhere I think it's clearer to keep it close to the code that uses it.

app/services/api/v3/decision_review/intake_status.rb
edmangimelli: I'm good with that. @pkarman any thoughts?

pkarman: this doesn't seem like a serializer class. It has as serialization method on it `to_json` but also implements business logic. I would probably leave it where it is.

spec/services/api/v3/decision_review/intake_status_spec.rb
```diff
@@ -1,28 +1,157 @@
...
-      expect(subject.render_hash).to eq(
-        json: {
-          type: intake.detail_type,
-          id: intake.detail.uuid,
-          attributes: { status: intake.detail.asyncable_status }
-        },
-        status: :accepted
+  let(:veteran_file_number) { "123456789" }
+
+  let(:create_intake) do
```
pkarman: We name our spec fixture data as nouns, not verbs, and we prefer to use the FactoryBot methods in specs. So this should be written as:

```ruby
let(:intake) do
  create(:intake, veteran_file_number: veteran_file_number, detail: detail)
end
```

spec/services/api/v3/decision_review/intake_status_spec.rb
```diff
@@ -1,28 +1,157 @@
...
+    lambda do |detail = nil|
+      Intake.create!(
+        user: Generators::User.build,
+        veteran_file_number: veteran_file_number,
+        detail: detail
+      )
+    end
+  end
+
+  let(:create_higher_level_review) do
```
pkarman: similarly,

```ruby
let(:higher_level_review) do
  create(:higher_level_review, veteran_file_number: veteran_file_number)
end
```

spec/services/api/v3/decision_review/intake_status_spec.rb
```diff
@@ -1,28 +1,157 @@
...
+
+  let(:create_supplemental_claim) do
+    SupplementalClaim.create!(veteran_file_number: veteran_file_number)
+  end
+
+  let(:create_appeal) do
+    Appeal.create!(veteran_file_number: veteran_file_number)
+  end
+
+  let(:new_intake_status) do
```
pkarman: this could be written as

```ruby
subject { described_class.new(intake) }
```

spec/services/api/v3/decision_review/intake_status_spec.rb
```diff
@@ -1,28 +1,157 @@
...
+    ->(intake) { Api::V3::DecisionReview::IntakeStatus.new(intake) }
+  end
+
+  context "#to_json" do
```
pkarman: the pattern we try to follow is `describe` the method, `context` with a `before` to set up the scene, and then the `it`. So

```ruby
describe "#to_json" do
  context "when the intake has a decision review" do
    let(:detail) { higher_level_review }
    let(:asyncable_status) { "dog" }
    let(:uuid) { "cat" }

    before do
      allow(detail).to receive(:asyncable_status) { asyncable_status }
      detail.uuid = uuid
    end

    it "returns correctly formatted JSON::API response" do
      json = subject.to_json

      expect(json).to be_a(Hash)
      # etc
    end
  end
end
```

spec/services/api/v3/decision_review/intake_status_spec.rb
```diff
@@ -1,28 +1,157 @@
...
+      expect(intake_status.to_json.keys).to contain_exactly(:errors)
+      expect(intake_status.to_json[:errors]).to be_a(Array)
+      expect(intake_status.to_json[:errors].length).to eq(1)
+      expect(intake_status.to_json[:errors][0]).to be_a(Hash)
+      expect(intake_status.to_json[:errors][0].keys).to contain_exactly(:status, :code, :title)
+      expect(intake_status.to_json[:errors][0][:status]).to be > 399
+    end
+  end
+
+  context "#http_status" do
```
pkarman: similar pattern here. If you find yourself writing "when..." in the `it` description, that part properly belongs in a `context` block.

spec/requests/api/v3/decision_review/intake_statuses_controller_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+  let!(:api_key) do
+    ApiKey.create!(consumer_name: "ApiV3 Test Consumer").key_string
+  end
+
+  let(:veteran_file_number) do
+    "64205050"
+  end
+
+  let(:new_intake) do
+    lambda do |detail = nil|
```
pkarman: a lambda inside a `let` is pretty unusual, since a `let` is itself a kind of lambda. See my example in the other specs.

spec/requests/api/v3/decision_review/higher_level_reviews_controller_spec.rb
```diff
@@ -36,6 +36,13 @@
     User.authenticate!(roles: ["Admin Intake"])
   end
 
+  let(:mock_api_user) do
```
pkarman: this should be a `before` block not a `let`. You want it to apply to every spec and get called automatically for you.

spec/requests/api/v3/decision_review/higher_level_reviews_controller_spec.rb
```diff
@@ -36,6 +36,13 @@
...
+    allow(User).to receive(:api_user) do
+      val = "ABC"
+      User.create!(station_id: val, css_id: val, full_name: val)
```
pkarman: prefer the factory: `create(:user ...)`

spec/requests/api/v3/decision_review/higher_level_reviews_controller_spec.rb
```diff
@@ -120,7 +127,7 @@ def post_params
   describe "#create" do
     describe "general cases" do
       it "should return a 202 on success" do
-        allow(User).to receive(:api_user) { build(:user) }
+        mock_api_user
```
pkarman: this should be called for you in the `before`

spec/requests/api/v3/decision_review/higher_level_reviews_controller_spec.rb
```diff
@@ -178,7 +186,8 @@ def post_params
       describe "reserved_veteran_file_number" do
         let(:veteran_file_number) { "123456789" }
         it "should return 500/reserved_veteran_file_number" do
-          allow(Rails).to receive(:deploy_env?).and_return(true)
+          mock_api_user
+          allow(Rails).to receive(:deploy_env?).with(:prod).and_return(true)
```
pkarman: I expect to see `allow` in a `before` block inside a `context` block. You set up your scenario in a `context` then run the test in an `it`.

spec/services/api/v3/decision_review/intake_status_spec.rb
```diff
@@ -1,28 +1,157 @@
...
+        detail: detail
+      )
+    end
+  end
+
+  let(:create_higher_level_review) do
+    HigherLevelReview.create!(veteran_file_number: veteran_file_number)
+  end
+
+  let(:create_supplemental_claim) do
```
pkarman: are the `supplemental_claim` and `appeal` even used here? are we ready to test those?

config/routes.rb
```diff
@@ -32,7 +32,9 @@
     end
     namespace :v3 do
       namespace :decision_review do
-        resources :higher_level_reviews, only: :create
+        resources :higher_level_reviews, only: [:create, :show]
+        resources :supplemental_claims, only: [:create, :show]
```
pkarman: are we ready to add `suppemental_claims` and `appeals`? I thought those were post-MVP.

app/models/decision_review.rb
```diff
@@ -57,6 +57,20 @@ def canceled_at_column
     def review_title
       to_s.underscore.titleize
     end
+
+    # Find the DecisionReview that has the given uuid, whether it's an Appeal, HigherLevelReview,
+    # etc. --any non-abstract descendant of DecisionReview.
+    # Purposely trying to not clobber find_by_uuid
+    def by_uuid(uuid)
```
pkarman: I like this!

app/controllers/api/v3/decision_review/supplemental_claims_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
```
pkarman: are we ready for this class? I don't see any tests and I thought this claim type was post-MVP.

app/controllers/api/v3/decision_review/appeals_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
```
pkarman: are we ready for this class? I don't see any tests and I thought this claim type was post-MVP.

config/routes.rb
edmangimelli: I _stubbed_ those controllers so that this method (from intake_statuses_controller):
```
  def decision_review_url
    url_for(
      controller: decision_review_controller,
      action: :show,
      id: decision_review.uuid
    )
  end
```
would work for any type of decision review. I suppose they could be cut --should i cut them?

app/controllers/api/v3/decision_review/supplemental_claims_controller.rb
edmangimelli: see above
https://github.com/department-of-veterans-affairs/caseflow/pull/12281#discussion_r334769796

app/controllers/api/v3/decision_review/appeals_controller.rb
edmangimelli: see above
https://github.com/department-of-veterans-affairs/caseflow/pull/12281#discussion_r334769796

edmangimelli:
@pkarman i've updated the tests. let me know if they need more love

config/routes.rb
pkarman: I can see the attraction of creating placeholders for where future work needs to happen. I raised the question because I wasn't sure that we were ready to create routes that were non-functional. I.e. as an API user I would expect a 404 response to a route that wasn't functional, rather than a 200 but nothing happening.

Leave them for now. It's aspirational but who doesn't need something to aspire to? ;)

spec/requests/api/v3/decision_review/higher_level_reviews_controller_spec.rb
```diff
@@ -109,41 +114,41 @@
...
 
-    describe "test error case: unknown_category_for_benefit_type" do
+    describe "when unknown_category_for_benefit_type" do
       let(:category) { "Words ending in urple" }
-      it "should return a 422 on this failure" do
-        post_params
+      it "should return a 422 error" do
+        allow(User).to receive(:api_user).and_return(mock_api_user)
+        post(*post_params)
         error = Api::V3::DecisionReview::IntakeError.new(:request_issue_category_invalid_for_benefit_type)
```
pkarman: why not just

```ruby
expect(response).to have_http_status(422)
```

like the `it` description claims? Are we concerned that the http status codes will change in future and we would have to update the test?

spec/requests/api/v3/decision_review/intake_statuses_controller_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+        hlr
+      end
+
+      it "returns status SUBMITTED_HTTP_STATUS" do
+        expect(response).to have_http_status(
+          Api::V3::DecisionReview::IntakeStatus::SUBMITTED_HTTP_STATUS
+        )
+      end
+
+      it("is correctly shaped") do
```
pkarman: idiomatically we don't typically use `it()` syntax with parentheses. Not a blocker.

spec/requests/api/v3/decision_review/higher_level_reviews_controller_spec.rb
edmangimelli: i suppose I was worried about making the test brittle. I think I'd rather change the `it` description to "should return :request_issue_category_invalid_for_benefit_type error"

Issue #12115, opened 2019-09-17
Allow Judges to Cancel AttorneyTasks & Return Cases to JudgeAssignTask


lomky:
Spun out of this discussion in slack: https://dsva.slack.com/archives/C6E41RE92/p1568660503005300

There's been scenarios where a Judge or SSC wants to take a case back from an attorney while BVA clarifies something. They don't want to reassign to someone else just hold it until clarification is done and then address assigning it to an attorney later. 

This **shouldn't** include any cases with a task created by the Attorney for VLJ support or any other dept opened, but only for cases where the `AttorneyTask` is a leaf node in the task tree, with status either `assigned` or `in progress`

Questions:
  - Should this be available if there are any closed VLJ support tasks under the attorney task?
  - Should this be available for `AttorneyRewriteTasks`, `AttorneyQRTask`, and `AttorneyDRTask`s?

kevmo:
1 |
2 | |||||
3 | ||||||
5 | |
8 |

Answer to first question is probably yes.  Second question - does it matter?  But probably yes.

Make sure you check Slack messages for commands run.

3.  Potential 5 - inserting this into user flow could be more complex than manually changing database.

hschallhorn:
Currently, `AttorneyTask`s are not created in caseflow for legacy appeals, however, there is work to start doing this with the depreciation of DAS in #12022. It seems unwise to attempt to wrestle with vacols and case assignment and reassignments if we are going to be using `AttorneyTask`s in the near future. Also, of all the cases where the Bat Team has manually done this, each appeal has been an AMA appeal. Proposing to either 
1. Call this done once ^ is merged
2. Add #12022 as a blocker

kevmo:
Blocked by #12022. 

Once #12022 is completed and PR #12295 is merged, this functionality described by this ticket _should_ exist.  We are leaving this ticket open as a reminder to check back in and make sure that is actually the case.

lomky:
#12022 is closed & #12295 complete! How's this looking @hschallhorn @kevmo ?

hschallhorn:
<img width="1026" alt="Screen Shot 2019-11-04 at 9 49 55 AM" src="https://user-images.githubusercontent.com/45575454/68129955-82b0bf80-fee8-11e9-8800-86392882f950.png">
🚀

PR #12295, opened 2019-10-07
Allow AMA AttorneyTasks to be cancelled

hschallhorn:
Connects #12115

### Description
Adds task action to attorney task to allow them to be cancelled and sent back to the assigning judge to be assigned again.

Screen Shot 2019-10-07 at 11 46 57 AM
Screen Shot 2019-10-07 at 11 47 04 AM
Screen Shot 2019-10-07 at 11 47 15 AM
Screen Shot 2019-10-07 at 11 47 28 AM

### Acceptance Criteria
- [ ] Attorneys can cancel AMA `AttorneyTask`s assigned to them

### Testing Plan
1. Sign in as "BVASCASPER1"
2. Go to any assigned task in the user's queue
3. Ensure they have the 'Cancel Task' action
4. Ensure cancelling the task removes it from the user's queue and sends the task back to the judge


spec/models/tasks/attorney_task_spec.rb
```diff
@@ -83,13 +85,29 @@
...
-        Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h
+        Constants.TASK_ACTIONS.ADD_ADMIN_ACTION.to_h,
+        Constants.TASK_ACTIONS.CANCEL_TASK.to_h
       ]
 
       expect(subject).to eq(expected_actions)
     end
   end
+
+  context "when cancelling the task" do
```
lomky: :100:!

app/models/tasks/attorney_task.rb
```diff
@@ -55,4 +58,21 @@ def assigned_to_role_is_valid
   def assigned_by_role_is_valid
     errors.add(:assigned_by, "has to be a judge") if assigned_by && !assigned_by.judge_in_vacols?
   end
+
+  def task_just_cancelled?
```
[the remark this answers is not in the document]
hschallhorn: Great point. It does not look like there is any place in the code where we cancel attorney tasks.

app/models/tasks/attorney_task.rb
```diff
@@ -14,16 +14,22 @@ class AttorneyTask < Task
   validate :assigned_to_role_is_valid
   validate :child_attorney_tasks_are_completed, on: :create
 
+  after_update :send_back_to_judge_assign, if: :task_just_cancelled?
+
   def available_actions(user)
     if parent.is_a?(JudgeTask) && parent.assigned_to == user
```
lomky: The way we do the Judge tasks, this lets the Judge who is the Decision Review judge (`parent.assigned_to`) cancel the task. But in the case that the Assigning Judge was different, they don't have the ability to cancel. Do we want to try to explicitly enable the Assigning Judge's ability to cancel?

hschallhorn: Great point! As the code works now, the [assigning judge is also the reviewing judge](https://github.com/department-of-veterans-affairs/caseflow/blob/5fc7d88144b0003e17daed97d1cacfd5977973e0/app/controllers/judge_assign_tasks_controller.rb#L33). Judge tasks can be [reassigned](https://github.com/department-of-veterans-affairs/caseflow/blob/e82803c2cdb863b53ece9796ab1a2585739d1e5b/app/models/tasks/judge_task.rb#L13), but this happens before an attorney task is created. Also, an attorney can return a decision to a different judge, but this "reassignment" occurs after the attorney task is completed.

lomky: Looks like we can reassign JudgeDecisionReviewTasks that are `on_hold` (i.e. assigned out to an attorney). We should be able to determine the assigning judge via the `assigned_by` on the AttorneyTask

PR #12306, opened 2019-10-08
Fix warning appearing in attorney checkout feature tests

ferristseng:
### Description

Occasionally seeing tests timeout, like this one:

https://app.circleci.com/jobs/github/department-of-veterans-affairs/caseflow/86975/parallel-runs/3?filterBy=FAILED

There isn't a lot of debug info, but I was seeing this warning:

```
Checking for expected text of nil is confusing and/or pointless since it will always match. Please specify a string or regexp instead.
```

The warning shows up 4 times locally and on successful runs in CircleCI, but only once in the failed run, so I think fixing it might help.


ferristseng:
Just got this warning during the build :(. Going to merge it still since it fixes a warning

Issue #12194, opened 2019-09-24
Appeals API - Issues endpoint


edmangimelli:
## Update (10/8)

Peter, Philip, and I had a video conference to discuss this. Sending back the ContestableIssue in the shape that is currently being received by the Intake UI was agreed upon by all parties as the correct way to go. Afterwards, Kelly Lein was updated and my misunderstanding about returning all issues (contestable or not) was cleared up --that was _not_ a requirement of this endpoint (I misunderstood). The proposal below has now been documented in the API spec https://github.com/department-of-veterans-affairs/caseflow/pull/12328 with only minor differences (the contestable bool has been removed, descriptions of the fields have been added, type was changed to ContestableIssue)

## Update (9/30)

After discussing with Philip, and weighing all of the concerns and goals, we've decided the best route is to return `ContestableIssue` objects (and uncontestable objects in a similar shape), specifying the type as simply "Issue", and removing the `id` field. This compromise takes us one step away from proper JSON:API format, but keeps us more closely inline with how the Intake UI is receiving data (at Peter's suggestion), and, most importantly, will provide the best experience for an API consumer **--formatting the objects this way allows an API consumer to instantly use them as RequestIssues without any transformation.**

## Proposal
In order to contest issues via a decision review, an API consumer will need a way to retrieve all of the issues for a specific veteran.

Below is my proposal for an _issues_ endpoint.
An API consumer will _get_ `/issues`, supplying 1) `the veteran's file number or SSN` and 2) `the receipt date of a hypothetical (or actual) decision review` as headers, and will receive a JSON response with an array of issues (described in detail below).

**Notable differences:**

Returns a flat array of issues. That is, legacy issues will not be nested. Each element of the returned array (top level field `data` below) is an _atomic issue_ in that it could potentially be contested on a decision review (if it is contestable) (see the original proposal for more details).

All issues--contestable and uncontestable--are returned. All issues will have a `contestable` bool to indicate whether or not they are contestable.
_A note on the implementation of this:_
ContestableIssueGenerator houses (or references) all the logic needed for deciding which issues are contestable. To return all non-contestable issues as well, ContestableIssueGenerator will be used first to identifiy the contestable issues. Then:
```
uncontestable_decision_issues = all_of_the_veteran's_decision_issues - veteran's_contestable_decision_issues
uncontestable_rating_issues = all_of_the_veteran's_rating_issues - veteran's_contestable_rating_issues
and so on...
```

[...]

PR #12328, opened 2019-10-09
Update API spec for the /issues endpoint - Appeals API

edmangimelli:
Updates the docs (decision_review.yaml openapi spec) for issue: https://github.com/department-of-veterans-affairs/caseflow/issues/12194

--the proposed Appeals API issues endpoint


edmangimelli:
I'd especially love some eyes on my descriptions @leikkisa @pkarman 

`legacyAppealId:` (vacols_id) "The ID of the LegacyAppeal in VACOLS"
`legacyAppealIssueId:` (vacols_sequence_id) "The ID of the specific issue of the referenced LegacyAppeal (One does not contest an entire LegacyAppeal, but, rather, a specific issue on a LegacyAppeal)"
`nodDate:` "(yyyy-mm-dd) \"Notice of Disagreement\" date"
`description:` "Description of the issue"
`disposition:` "Disposition of the issue"
`closeDate:` "Date of disposition"
`note:` "Note"
`ratingIssueId:` "RatingIssue ID"
`ratingIssueProfileDate:` "(yyyy-mm-dd) RatingIssue date"
`ratingIssueDiagnosticCode:` "RatingIssue diagnostic code"
`ratingDecisionId:` "The BGS ID for the contested rating decision. This may be populated while ratingIssueId is nil"
`decisionIssueId:` "DecisionIssue date"
`approxDecisionDate:` "(yyyy-mm-dd) Approximate decision date"
`description:` "Description"
`rampClaimId:` "RampClaim ID"
`titleOfActiveReview:` "Title of active review"
`sourceReviewType:` "Source review type"
`timely:` "An issue is timely if it has been less than 372 days since its receipt date. Untimely issues (timely == false), must accompanied by a timeliness exemption"
`latestIssuesInChain:` "Shows the chain of decision and rating issues that preceded this issue. Only the most recent issue can be contested (the object itself that contains the latestIssuesInChain attribute)."
`isRating:` "Whether or not this is a RatingIssue"

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -973,6 +973,140 @@ paths:
...
+                      anyOf:
+                        - type: object
+                          properties:
+                            type:
+                              type: string
+                              enum: ["ContestableIssue"]
+                            attributes:
+                              type: object
+                              properties:
+                                legacyAppealId: # vacols_id
```
leikkisa: I'm guessing you need to send legacies issues along with the contestable issues for the next processing step.  But I wanted to comment just in case.

Legacy issues aren't contestable issues themselves.  They are more like VACOLS request issues that are getting converted to Caseflow request issues, and request issues don't contest other request issues.

Users first pick a contestable issue, which is from a rating issue or a past Caseflow decision (not a legacy appeal).  Then that contestable issue gets connected to a legacy issue.

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -973,6 +973,140 @@ paths:
...
+                                  nullable: true
+                                  description: "Description"
+                                rampClaimId:
+                                  type: string
+                                  nullable: true
+                                  description: "RampClaim ID"
+                                titleOfActiveReview:
+                                  type: string
+                                  nullable: true
+                                  description: "Title of active review"
```
leikkisa: This is used for eligibility checks.  If an issue is still active on another Decision Review, then it's ineligible.  This title is used for messaging.

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -973,6 +973,140 @@ paths:
...
+                                sourceReviewType:
+                                  type: string
+                                  nullable: true
+                                  description: "Source review type"
```
leikkisa: This is also used for eligibility checks.  This is the type of Decision Review the issue was last decided on (if any).  If an issue was just decided on an Appeal, it can't be added to another Appeal or HLR.  If it was just on an HLR, it can't be added to another HLR.

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -973,6 +973,140 @@ paths:
...
+                                timely:
+                                  type: boolean
+                                  description: "An issue is timely if it has been less than 372 days since its receipt date. Untimely issues (timely == false), must accompanied by a timeliness exemption"
```
leikkisa: An issue is timely if the receipt date is within 372 dates of the decision date.

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -973,6 +973,140 @@ paths:
...
+                                  nullable: true
+                                  description: "RatingIssue diagnostic code"
+                                ratingDecisionId:
+                                  type: string
+                                  nullable: true
+                                  description: "The BGS ID for the contested rating decision. This may be populated while ratingIssueId is nil"
+                                decisionIssueId:
+                                  type: string
+                                  nullable: true
+                                  description: "DecisionIssue date"
```
leikkisa: This is an ID instead of a date, and an integer type

app/controllers/api/docs/v3/decision_reviews.yaml
edmangimelli: ah... so you never contest a _vacols_sequence_id + vacols_id_ ?
you select a legacy issue (sequence), and then associate it with decision or rating?

app/controllers/api/docs/v3/decision_reviews.yaml
edmangimelli: how about:
"Title of DecisionReview that this issue is still active on"
?

app/controllers/api/docs/v3/decision_reviews.yaml
edmangimelli: Ah. Ok this ties into a conversation we're having with Peter right now in slack
https://dsva.slack.com/archives/CKUCFQAHJ/p1570587547013000?thread_ts=1570561492.009000&cid=CKUCFQAHJ

app/controllers/api/docs/v3/decision_reviews.yaml
edmangimelli: whoops! thanks!

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -997,6 +997,140 @@ paths:
...
+          content:
+            application/vnd.api+json:
+              schema:
+                type: object
+                properties:
+                  data:
+                    type: array
+                    items:
+                      anyOf:
+                        - type: object
```
erluti: can we add a description to these two objects that could be in the array to give a hint as to what they are describing

edmangimelli:
After speaking with @leikkisa yesterday, it seems like this approach won't work for Legacy Issues and begs some bigger questions that need looking into and stakeholder input (should this API be able to associate Legacy Issues at all, if so, what does that mean for our consumers? etc). Consequently, I'm going to strike Legacy Issues from this PR

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -997,6 +997,104 @@ paths:
...
+                              nullable: true
+                              description: "The BGS ID for the contested rating decision. This may be populated while ratingIssueId is nil"
+                            decisionIssueId:
+                              type: integer
+                              nullable: true
+                              description: "DecisionIssue ID"
+                            approxDecisionDate:
+                              type: string
+                              nullable: true
+                              pattern: '^\d{4}-\d{2}-\d{2}$'
```
erluti: instead of using a pattern jscon schema supports `format: date`
https://json-schema.org/understanding-json-schema/reference/string.html#built-in-formats

edmangimelli: oh nice. i didn't realize their date format was ISO8601  😎 
Would you mind if that was a separate PR? (to get the rest of the document) (to avoid merge conflicts potentially)

PR #12335, opened 2019-10-09
Filter out VACOLS hearings not associated with a Case

### LINKED ISSUE #12321: NotNullViolation when getting CO hearing days
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/6962/events/505923/

Seeing this error on `/hearings/hearing_day.json`. It's causing this issue: https://dsva.slack.com/archives/CHX8FMP28/p1570555326407700

The issue happens when initializing the LegacyAppeal that's supposed to be associated with the hearing. The `folder_nr` field on the case hearing from VACOLS is null, and tripping a constraint in the Caseflow DB.

ferristseng:
Resolves #12321 

### Description

  * Updates queries to filter out case hearings where `folder_nr` is `null`
  * Adds some tests

### DIFF
--- a/spec/repositories/hearing_repository_spec.rb
+++ b/spec/repositories/hearing_repository_spec.rb
@@ -9,7 +9,7 @@ describe HearingRepository, :all_dbs do
     Time.zone = "America/Chicago"
   end
 
-  context ".fetch_hearings_for_parent", focus: true do
+  context ".fetch_hearings_for_parent" do
     let(:hearing_date) { Time.utc(2019, 5, 2) }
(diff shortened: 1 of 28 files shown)


spec/repositories/hearing_repository_spec.rb
```diff
@@ -9,6 +9,34 @@
     Time.zone = "America/Chicago"
   end
 
+  context ".fetch_hearings_for_parent" do
+    let(:hearing_date) { Time.utc(2019, 5, 2) }
```
pkarman: important that it is after 2019-01-01

PR #12338, opened 2019-10-09
Added Makefile documentation & removed defunct FACOLS command

lomky:
Added descriptions to some of the makefile commands. Removed the non-functional FACOLS command.

### DIFF
--- a/Makefile.example
+++ b/Makefile.example
@@ -42,7 +42,7 @@
 whats-next:  ## TODO
 	scripts/whats-next
 
-ready:  ## TODO
+ready:  ## Helper command that watches the docker containers to finish start up
 	h=`docker-compose ps | grep healthy`; while [ "$$h" == "" ]; do h=`docker-compose ps | grep healthy`; echo $$h; sleep 1; done
 
 up:  ## Start the docker containers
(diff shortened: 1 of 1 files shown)


Makefile.example
```diff
@@ -61,12 +61,9 @@ clean:  ## Clean logs, compiled webpack assets, and temporary files
 	rm -rf public/assets
 	rm -rf tmp/capybara
 
-vacols:  ## TODO
+facols:  ## Connect to your local FACOLS instance
 	docker exec --tty -i VACOLS_DB bash
 
-facols:  ## TODO
```
pkarman: this sqlplus command is a handy way to connect directly to the oracle db. You could do it in 2 steps with `make vacols` and then running `sqlplus` w/in the docker container. But `facols` and `vacols` are 2 different targets.

Maybe `facols` should be renamed as `facols-prompt` or something to indicate you go directly to oracle cli.

Makefile.example
```diff
@@ -61,10 +61,10 @@ clean:  ## Clean logs, compiled webpack assets, and temporary files
 	rm -rf public/assets
 	rm -rf tmp/capybara
 
-vacols:  ## TODO
+facols-box:  ## Connect to the docker FACOLS instance
```
pkarman: nit: maybe `facols-bash` is a little more mnemonic?

PR #12341, opened 2019-10-09
set label to 'Search'

### LINKED ISSUE #12101: Search label text in Queue > Reader reads "Search small" by AT. It should just read "Search"
The label for the search element is incorrectly read as Search Small for assistive technology users. Visually the label reads: Type to Search. This defect occurs throughout the web hosted application with the search function improperly being read as Search Small. (New: 08/27/2019) Ensure you can hear what is visually being seen to ensure compliance

Noting – we should also check the search tips and make sure those are also associated with the search input as well.

yoomlam:
Resolves #12101 

### Description
Setting the label as "Search" (when a title is not specified).
Tried to set label to empty string to cause VoiceOver to skip the label but that caused rspec test failures (due to lack of label for input form).

### Acceptance Criteria
- [ ] VoiceOver no longer says "Search small"

### Testing Plan
1. Use VoiceOver to navigate to the search field, listening for erroneous screen reading.


yoomlam:
Not sure why there are so many RSpec test failures. 

Regarding the lint test failing, here's what JC had to say:
> OK, the camel case warning there is just silly. The native React function is not camel case, so you obviously won't be able to pass that.
However, this is something that should be eventually updated (as that lifecycle method will be removed in React 17). If you're feeling ambitious, look at suggested methods for migrating to a different supported lifecycle method.
getDerivedStateFromProps is a usual candidate for migrating from componentWillReceiveProps, but it's a static method and you can't call other instance methods in it. I'd have to look more to see what would make sense in this specific context

Resolving this is beyond by knowledge at this point and is irrelevant to the ticket.

alisan16:
Can you update the PR description? It doesn't look like you're setting the label to empty string anymore.

Verified AC.

Created an issue to update the eslint rules: https://github.com/department-of-veterans-affairs/caseflow/issues/12359

Issue #12352, opened 2019-10-09
User.api_user needs to be able to intake any veteran


edmangimelli:
From this thread:
https://dsva.slack.com/archives/CKUCFQAHJ/p1569851414024700

The Appeals API's api_user (User.api_user), which is the user used for all Appeals API requests (request made by any API consumer), needs to be able to intake any veteran file. Currently, when attempting to do fake intakes in the UAT environment, :veteran_not_modifiable errors are being returned. 

A separate authorization layer for the Appeals API will be created in the future, but, for MVP (supporting VA.gov), a veteran will only be intaking their own file, and that authorization will occur via VA.gov.

### Acceptance criteria
- feature toggled
- tested
- all parties agree on the approach

PR #12353, opened 2019-10-09
Appeals API - ignore station ID checks for User.api_user

edmangimelli:
WIP
User.api_user needs to be able to intake any veteran. Details in https://github.com/department-of-veterans-affairs/caseflow/issues/12352

resolves https://github.com/department-of-veterans-affairs/caseflow/issues/12352

Is this tiny change the right solution? Or is this the wrong approach?


app/models/validators/intake_start_validator.rb
```diff
@@ -71,6 +71,6 @@ def file_number_reserved?
   end
 
   def user_may_modify_veteran_file?
-    BGSService.new.may_modify?(veteran_file_number, veteran.participant_id)
+    (FeatureToggle.enabled?(:api_v3) && intake.user == User.api_user) || BGSService.new.may_modify?(veteran_file_number, veteran.participant_id)
```
erluti: Does this need to be wrapped in a feature flag?  If so, maybe the feature flag should be in the `User.api_user` and raise an exception so all attempts to use the api_user blow up?

edmangimelli: > Does this need to be wrapped in a feature flag?

safety first 🤷‍♂ 

>  If so, maybe the feature flag should be in the User.api_user and raise an exception so all attempts to use the api_user blow up

I'm not sure how that would work. Even without the feature flag, i'd want `intake.user == User.api_user` always being tested, and throwing an exception in that scenario seems bad. So raising an exception for "all attempts to use the api_user" doesn't seem good. But I probably don't have the full picture of what you're sayign

edmangimelli: this thread:
https://dsva.slack.com/archives/CKUCFQAHJ/p1570665708005000?thread_ts=1570659417.004700&cid=CKUCFQAHJ

has a couple of verbose refactors of my one line change. but neither hits what you're saying

edmangimelli: Yeah, testing in many places as to whether a request is an appeals api request seems wrong. Maybe intake_start_validator's initializer needs an appeals_api option or something. 🤔

edmangimelli: instead of an exception, maybe false or nil? that might be ok:
```
    def api_user
      return nil unless FeatureToggle.enabled?(:api_v3)

      @api_user ||= find_or_create_by(
        station_id: "101",
        css_id: "APIUSER",
        full_name: "API User"
      )
    end
```
@erluti whaddya think?

erluti: I appreciate your concern for security, but I think since the `api_user` needs to be explicitly called there's not much danger of it being used outside of the feature flag.  I think we can rely on PR reviews outside of the API and the routes being wrapped with the feature flag to protect from someone misusing this user.

app/models/validators/intake_start_validator.rb
```diff
@@ -71,6 +71,6 @@ def file_number_reserved?
   end
 
   def user_may_modify_veteran_file?
-    BGSService.new.may_modify?(veteran_file_number, veteran.participant_id)
+    intake.user == User.api_user || BGSService.new.may_modify?(veteran_file_number, veteran.participant_id)
```
pkarman: could we write this on 2 lines instead? I would find it easier to read.

```ruby
return true if intake.user == User.api_user

BGSService.new.may_modify?(veteran_file_number, veteran.participant_id)
```

spec/models/validators/intake_start_validator_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      create(:intake, veteran_file_number: veteran.file_number, user: user)
+    end
+
+    let(:validator) do
+      described_class.new(intake: intake)
+    end
+
+    it "returns true when BGS allows modification" do
+      allow_any_instance_of(BGSService).to receive(:may_modify?) { true }
+      expect(validator.send(:user_may_modify_veteran_file?)).to be true
```
pkarman: the fact that we must use `send` here indicates that either (a) we shouldn't be testing a private method in this way, and/or (b) we should make the method public. The use of `send` is a code smell IMO.

pkarman: Since the lone public method is `validate` we should test that method instead, and look for anticipated error codes instead. I think that's preferable over making the private method public, but I could be convinced otherwise.

edmangimelli: my gut instinct is that
```
def a
  ...
  b
  ...
end

private 

def b
  ...
  c
  ...
end

def c
  ...
end
```
if the only way to test `c` is by testing `a`, maybe the class could use some refactor-love. Indirect testing like that feels no different than trying to test:
```
def a
   ...
   ...
   ...20 more lines...
   (
     a few lines
     that I want
     to test
   )
   ...
   ...
end
```
But, discussing with Philip and weighing the alternatives, testing `validate` is probably the way to go as refactoring this class is certainly beyond the scope of this ticket. And, agreed, making the private method public to make it testable seems like a worse alternative to testing `validate` (incidentally, I think I end up defaulting to class methods to increase testing surface area (which is probably not much better)).

spec/models/validators/intake_start_validator_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      expect(validate_error_code).to be nil
+    end
+
+    it "sets error_code \"veteran_not_modifiable\" when BGS does not allow modification" do
+      allow_any_instance_of(BGSService).to receive(:may_modify?) { false }
+      expect(validate_error_code).to eq "veteran_not_modifiable"
+    end
+
+    it "sets no error_code when user is User.api_user" do
+      user = User.api_user
```
pkarman: one nit:

```ruby
context "user is api_user" do
  let(:user) { User.api_user }

  it "sets no error code" do
    expect(validate_error_code).to be nil
  end
end
```

edmangimelli: ugh. Yeah, that's a big nit. Right after I pushed I thought, "hmmm. better doublecheck that `user = ...` is doing what I want (properly scoped)". Tested `user = "abc"` and got no error 😞 
and then i see your comment. will update. will purchase book on rspec.

Issue #11938, opened 2019-08-29
Add date picker to Step 3 (Review screen) in Caseflow Intake


jimruggiero:
### Context: 
Feedback from User Research session held on 8/12 with VBA Intake user included a request to add a date picker to this screen, noting multiple prior instances of incorrect dates being added (found during quality process checks) and that date picker functionality is preferred because it is a highly familiar pattern to users. They use a date picker in all other applications required to complete their work and find manual date entry to be cumbersome. 

### Story:

As a VBA Intake user, specializing in the processing of HLR and SC compensation appeals claims, I want the ability to use a date picker to enter the forms' receipt date, so that I can reduce the risk of manual data entry error. 

### AC: 
- Current manual date entry functionality is unchanged.
- A date picker icon appears and functions per current spec (as coded elsewhere in Caseflow); date in form will continue to be editable once populated via date picker selections, until the "Continue to the next step" button is clicked, or when users navigate back to the page via the browser back button.
- Date picker icon appears for date entry on intake screen 3 (Review) and in all applicable add issue modals.

PR #12363, opened 2019-10-11
 Intake add date-picker

Sjones352:
Connects #11938 
**Receipt Date** 
Screen Shot 2019-10-11 at 1 38 19 PM

**Decision Date**
Screen Shot 2019-10-11 at 1 38 58 PM

**Withdraw Date**
Screen Shot 2019-10-11 at 1 39 48 PM

**Dates formatted**
Screen Shot 2019-10-11 at 1 37 12 PM


spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -683,13 +682,11 @@ def complete_higher_level_review
         fill_in "Issue category", with: "Apportionment"
         find("#issue-category").send_keys :enter
 
-        fill_in "Decision date", with: "13/04/2019"
-        expect(page).to have_content("Please enter a valid decision date")
+        fill_in "Decision date", with: "12/04/2019"
+        expect(page).to have_content("Decision date cannot be in the future")
 
         Timecop.return
         fill_in "Decision date", with: Time.zone.tomorrow.mdY
```
tomas-nava: I don't think you need these lines after moving the expectation from below.

spec/feature/intake/appeal/edit_spec.rb
```diff
@@ -579,12 +571,8 @@
 
       fill_in "withdraw-date", with: "14/20/2019"
```
tomas-nava: Because we're using the date picker component, `fill_in` doesn't behave the way this test expects it to. If the component has no value, `fill_in` will work *but* it will validate in place. So if I have a blank component:

Screen Shot 2019-10-11 at 5 27 34 PM

...and I say ```fill_in "withdraw-date", with: "14/20/2019"``` I'll get

Screen Shot 2019-10-11 at 5 28 41 PM

...because the component auto-corrected `14` to `12`, a valid date. After that, everything I pass to the component via `fill_in` will go straight to the year field, leaving the month and day values untouched. So if the field looks like the above and I say ```fill_in "withdraw-date", with: "01/01/2123"```, we'll get:

Screen Shot 2019-10-11 at 5 33 23 PM

All that is to say, we shouldn't use `fill_in` to test the date-picker component.

spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -683,13 +682,11 @@ def complete_higher_level_review
         fill_in "Issue category", with: "Apportionment"
         find("#issue-category").send_keys :enter
 
-        fill_in "Decision date", with: "13/04/2019"
-        expect(page).to have_content("Please enter a valid decision date")
+        fill_in "Decision date", with: "12/04/2019"
+        expect(page).to have_content("Decision date cannot be in the future")
```
tomas-nava: I think this `fill_in` will work, but this test will start failing in 62 days when 12/04/2019 is no longer in the future.

Sjones352: Fixed it

spec/feature/intake/appeal/edit_spec.rb
```diff
@@ -579,12 +571,8 @@
...
 
-      expect(page).to have_content("We cannot process your request. Please enter a valid date.")
+      expect(page).to have_content("We cannot process your request")
       expect(page).to have_button("Save", disabled: true)
```
tomas-nava: The `fill_in` calls in this test are still not doing what this test expects them to, as noted in [this comment](https://github.com/department-of-veterans-affairs/caseflow/pull/12363#discussion_r334210972). If you clear the contents of the field with:
```
find("#withdraw-date").native.clear
```
before the `fill_in` on line `567`, that one will behave as expected.

But the `fill_in` on line `572` is entering an invalid date and expecting to see an error message because of that. What's actually happening is that the component is validating on the fly, auto-correcting the date to `12/20/2019`, and we're seeing the error message because that date's still in the future. So, I'd just get rid of this last `fill_in` and set of expectations, because it's just retesting the preceding expectation.

client/app/intake/util/issues.js
```diff
@@ -123,7 +110,7 @@ export const formatRequestIssues = (requestIssues, contestableIssues) => {
       vacolsIssue: issue.vacols_issue,
       endProductCleared: issue.end_product_cleared,
       endProductCode: issue.end_product_code,
-      withdrawalDate: formatDateStrUtc(issue.withdrawal_date),
+      withdrawalDate: issue.withdrawal_date,
```
tomas-nava: I don't know why this date was converted to UTC before, but I don't think it will be anymore. Should it be?

client/app/intake/pages/higherLevelReview/review.jsx
```diff
@@ -104,6 +106,29 @@ class Review extends React.PureComponent {
...
+  veteranName: PropTypes.string,
+  receiptDate: PropTypes.string,
+  receiptDateError: PropTypes.string,
+  benefitType: PropTypes.string,
+  benefitTypeError: PropTypes.string,
+  informalConference: PropTypes.string,
+  informalConferenceError: PropTypes.string,
+  sameOffice: PropTypes.string,
+  sameOfficeError: PropTypes.string,
+  legacyOptInApproved: PropTypes.string,
```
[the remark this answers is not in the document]
leikkisa: ➕ 
Also, informalConference and sameOffice are also "bool".

client/app/intake/pages/supplementalClaim/review.jsx
```diff
@@ -81,6 +83,30 @@ class Review extends React.PureComponent {
   }
 }
 
+Review.propTypes = {
+  veteranName: PropTypes.string,
+  receiptDate: PropTypes.string,
+  receiptDateError: PropTypes.string,
+  benefitType: PropTypes.string,
+  benefitTypeError: PropTypes.string,
+  informalConference: PropTypes.string,
```
leikkisa: informalConference, sameOffice, and legacyOptInApproved props are bool

client/app/intake/util/issues.js
Sjones352: Good catch. That was a mistake on my part but it should be `withdrawalDate: issue.withdrawal_date`

leikkisa: I checked the git blame to see if this was there for any special reason (like a bug fix), and it wasn't, it was the first time withdrawalDate was added here. I'm guessing it accomplished the same formatting as formatDateStr so there wasn't discrimination between the two.

I like moving the use of formatDateStr instead of here to the front end code, and that should convert the dates into local time (per a comment I saw elsewhere).

client/app/intake/util/issues.js
```diff
@@ -327,7 +314,7 @@ export const formatAddedIssues = (intakeData, useAmaActivationDate = false) => {
         referenceId: issue.id,
         text: issue.description,
         // formatDatStr converts to local time instead of UTC
```
leikkisa: This comment can be removed

spec/feature/intake/appeal/edit_spec.rb
leikkisa: Good catch.

client/app/intake/pages/supplementalClaim/review.jsx
Sjones352: This method converts sameOffice to string once user clicks radio button `sameOffice === null ? null : sameOffice.toString()`. Please check error when I inspect console
Screen Shot 2019-10-17 at 11 06 19 AM

client/app/intake/util/issues.js
Sjones352: Fixed and moved to frontend

client/app/intake/pages/supplementalClaim/review.jsx
leikkisa: Got it! I thought the prop was bool, since we're getting it from the backend, and that it was converted to string for the radio button.

PR #12366, opened 2019-10-11
instead of sleep, cache geomatch API responses

pkarman:
Our `rake db:seed` task was very slow because we had two `sleep 1` calls in a loop, as part of the `FetchHearingLocationsForVeteransJob` process.

This PR adds some logging to the seed process so it's easier to spot which steps are happening in the Rails logs, and eliminates the `sleep` calls. Instead identical API calls to the geomatching service are cached for 24 hours, which avoids the need I think the `sleep` was filling (throttling our use of that va.gov API). My assumption with the caching logic is that geomatching is a relatively fixed thing (stations don't physically move often).

The speed up is pretty dramatic:
```
before: 5m52.478s
after: 2m11.036s
```

### DIFF
--- a/app/services/external_api/va_dot_gov_service.rb
+++ b/app/services/external_api/va_dot_gov_service.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "json"
+require "digest"
 
 class ExternalApi::VADotGovService
   BASE_URL = ENV["VA_DOT_GOV_API_URL"] || ""
@@ -32,10 +33,13 @@ class ExternalApi::VADotGovService
     private
 
     def send_facilities_request(query:)
-      response = send_va_dot_gov_request(
-        query: query,
-        endpoint: FACILITIES_ENDPOINT
-      )
+      cache_key = Digest::SHA1.hexdigest query.to_json
+      response = Rails.cache.fetch(cache_key, expires_in: 24.hours) do
+        send_va_dot_gov_request(
+          query: query,
+          endpoint: FACILITIES_ENDPOINT
+        )
+      end
 
       ExternalApi::VADotGovService::FacilitiesResponse.new(response)
     end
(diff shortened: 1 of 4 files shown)


app/services/external_api/va_dot_gov_service.rb
```diff
@@ -32,10 +33,13 @@ def validate_address(address)
     private
 
     def send_facilities_request(query:)
-      response = send_va_dot_gov_request(
-        query: query,
-        endpoint: FACILITIES_ENDPOINT
-      )
+      cache_key = Digest::SHA1.hexdigest query.to_json
```
pkarman: during `db:seed` this was generating lots of identical API calls.

db/seeds.rb
```diff
@@ -1252,36 +1252,43 @@ def create_inbox_messages
...
-
-    create_intake_users
-
-    create_inbox_messages
-
+  def perform_seeding_jobs
     # Active Jobs which populate tables based on seed data
     FetchHearingLocationsForVeteransJob.perform_now
     UpdateCachedAppealsAttributesJob.perform_now
+    NightlySyncsJob.perform_now
```
pkarman: this is a new job. everything else is just refactor.

app/services/external_api/va_dot_gov_service.rb
lomaxap: We geomatch based on the veteran's address's lat/long coordinates, so these would never be the same for geomatching. For requests to get only facility data, this will be very helpful. 

The sleep is there to limit the rate of requests (I think the limit is 200 per second?). I think I put it there out of an abundance of caution. Since geomatching requires a BGS call and often a VACOLS call, the whole transaction takes around half a second, so there probably isn't any use for it. We're sending limit errors to data dog, so I'll keep an eye on that after this gets deployed

PR #12371, opened 2019-10-11
Add Back Hearing view

lomaxap:
When consolidating hearing prep with hearings schedule, we removed a method that logged that a hearing worksheet was viewed by a judge. This adds back that functionality and removes a conditional that limited the worksheet link to worksheets viewed by that judge


client/app/queue/CaseHearingsDetail.jsx
```diff
@@ -41,7 +42,7 @@ class CaseHearingsDetail extends React.PureComponent {
       </React.Fragment>
     }];
 
-    if (hearing.viewedByJudge && !userIsVsoEmployee) {
```
pkarman: should we add a test for this link's visibility? That was what sparked this PR initially IIRC.

PR #12376, opened 2019-10-11
Improve rating decision dates

pkarman:
connects #11840 

### Description
Per the logic described in https://github.com/department-of-veterans-affairs/caseflow/issues/11840#issuecomment-536614144 expand the definition of the `decision_date` to be more inclusive and hopefully reduce unidentified request issues.

**NOTE** this should not be merged until it can be tested against legit known production data


app/models/rating_decision.rb
```diff
@@ -68,7 +75,7 @@ def contestable?
 
     return false unless decision_date
 
-    effective_date_near_promulgation_date?
+    effective_date_near_promulgation_date? || effective_date_near_profile_date?
```
leikkisa: Will this filter out the old issues?

pkarman: No. That will be a in a separate PR, addressing `app/workflows/contestable_issue_generator.rb`

app/models/rating_decision.rb
```diff
@@ -93,16 +100,40 @@ def service_connected?
...
+  # we have to make educated guesses because there are a variety of "date" attributes
+  # on a rating decision and they are not populated consistently.
+
+  def effective_start_date_of_original_decision
+    original_denial_date || converted_begin_date || begin_date
+  end
+
   def effective_date
-    original_denial_date || converted_begin_date || begin_date || disability_date
+    return disability_date if effective_start_date_prior_to_modern_recordkeeping?
```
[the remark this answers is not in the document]
pkarman: disability_date does move. Each `disability` entry has a unique `disability_date`. The constant is the `disability_id`. Even the `diagnostic_text` changes from entry to entry sometimes, so I'm not sure it's possible to tell if "nothing on the disability has changed". I.e. if the claimant kept listing the same entry on each subsequent appeal, it would keep showing up, even with the same decision being made (e.g. not service connected).

PR #12377, opened 2019-10-13
change filtering to UNION results, not INTERSECT

### LINKED ISSUE #12082: Filtering by Case Type should be additive
image

## Acceptance criteria
* [ ] Selecting options in addition to "AOD" in the "Type(s)" column shows those options even if they are not AOD for organization queues using the pagination API

yoomlam:
Resolves #12082 

### Description
Filtering by type should union (rather than intersect) the sets of rows that satisfy each filter selection. Specifically, in queries that select tasks, the where clauses should be joined by an "OR", not an "AND".

### Acceptance Criteria
- [ ] For the case type column, relevant tables should show all rows that satisfy any filter selection.

### Testing Plan
1. Verify all components that use the changed TaskFilter behave as desired, i.e., more filter selections should add to the number of rows.
2. Also check filtering selections across columns.


app/models/task_filter.rb
```diff
@@ -37,7 +37,7 @@ def where_clause
     where_arguments = filters.map(&:values).reject(&:empty?)
 
     if filter_params.any? { |filter_string| filter_string[/typeColumn&val=.*is_aod/] }
-      where_string << "#{where_string.present? ? ' AND ' : ''}cached_appeal_attributes.is_aod = true"
+      where_string << "#{where_string.present? ? ' OR ' : ''}cached_appeal_attributes.is_aod = true"
     end
```
yoomlam: 1. Why is `is_aod` being treated specially?
2. Why isn't a constant for `is_aod` being used?

yoomlam: Because `is_aod` is not actually a case type. It is an attribute directly on the case.

Given `filter_params = ["col=typeColumn&val=TranslationTask,is_aod"]`, we get:
```
where_string => cached_appeal_attributes.case_type IN (?) OR cached_appeal_attributes.is_aod = true
where_arguments => [["TranslationTask"]]
```

yoomlam:
Ticket #12082 wants behavior that is opposite the behavior desired in this test: https://github.com/department-of-veterans-affairs/caseflow/blob/365647d7018d5ba480a62d40c8686865cadeec22/spec/models/task_filter_spec.rb#L316

Which behavior is desired?

lomky:
When we implemented the refactor & related specs a couple months ago we sought parity rather than best behavior. So I would suggest we work towards the desired behavior now (what #12082 is asking for) rather than what we previously implemented (what the spec suggests) and update the spec accordingly.

app/models/task_filter.rb
```diff
@@ -46,7 +46,7 @@ def where_clause
   def self.where_string_from_filters(filters)
     filters.map do |filter|
       filter.values.present? ? "#{table_column_from_name(filter.column)} IN (?)" : nil
-    end.compact.join(" AND ")
+    end.compact.join(" OR ")
```
hschallhorn: Pretty sure this should stay an "AND". We'd want the resulting rows to satisfy anything in a single column, but everything across columns. If I was buying shirts and selected the colors green and blue and a size of small, I should be shown all small green and blue shirts, not all green and blue shirts in any size, and small shirts in any color.

hschallhorn:
Agreed with Kat! I was the one who wrote this and had a mild misunderstanding of the desired behavior, so this test is wrong 😬.

app/models/task_filter.rb
yoomlam: Of course. My bad.

spec/models/queue_filter_parameter_spec.rb
```diff
@@ -55,7 +55,7 @@
       let(:filter_string) { "col=#{Constants.QUEUE_CONFIG.COLUMNS.APPEAL_TYPE.name}&val=is_aod,#{good_value}" }
 
       it "rejects the is_aod value" do
-        expect(subject.values).to match_array([good_value])
```
hschallhorn: We can remove this whole 'when typeColumn has 'is_aod' filter" test as `queue_filter_parameter` no longer handles "is_aod"

app/models/queue_filter_parameter.rb
```diff
@@ -21,10 +21,6 @@ def self.from_string(filter_string)
     filter_hash = Rack::Utils.parse_query(filter_string)
     values = filter_hash["val"]&.split(",")&.map { |value| escaped_value(value) }
 
-    if filter_hash["col"].eql?(Constants.QUEUE_CONFIG.COLUMNS.APPEAL_TYPE.name)
-      values.reject! { |value| value.eql?("is_aod") }
-    end
-
```
hschallhorn: 👍 Love moving this out of queue_filter_parameter

app/models/task_filter.rb
```diff
@@ -36,19 +36,35 @@ def where_clause
...
 
+  def self.create_where_clause(filter)
+    clause = "#{table_column_from_name(filter.column)} IN (?)"
+    case filter.column
+    when Constants.QUEUE_CONFIG.COLUMNS.APPEAL_TYPE.name
+      filter_selections = filter.values
+      if filter_selections.include?(Constants.QUEUE_CONFIG.FILTER_OPTIONS.IS_AOD.key)
+        clause = extract_aod_clause(filter, clause)
+      end
+    end
```
hschallhorn: Using a switch statement is a bit overkill with just the single case. Could condense this to 

```ruby
    if filter.column.eql?(Constants.QUEUE_CONFIG.COLUMNS.APPEAL_TYPE.name) &&
       filter.values.include?(Constants.QUEUE_CONFIG.FILTER_OPTIONS.IS_AOD.key)
      clause = extract_aod_clause(filter, clause)
    end
```

spec/models/task_filter_spec.rb
```diff
@@ -274,6 +274,16 @@
         end
       end
 
+      let(:aod_cases) do
```
hschallhorn: Mildly nitpicky, but can we rename this to `aod_case_ids` so it's evident in the union below?

app/models/task_filter.rb
yoomlam: Got rid of case statement but Codeclimate complains: `Use '#value?' instead of '#values.include?'. [rubocop]` because it thinks `values` is a method on an array object. We could change filter's `values` attribute to something else if we really wanted to.

PR #12385, opened 2019-10-15
Tableau SQL data sources

pkarman:
connects https://github.com/department-of-veterans-affairs/dsva-vacols/issues/70

### Description

Updates Tableau basic data source SQL to reflect latest BVA progress categories, and to use the `cached_user_attributes` table instead of cross-schema join to VACOLS.

Adds tests for the above.

Adds SQL formatting tool for consistent pretty print.


rajinderbhanot:
Good

Issue #12211, opened 2019-09-25
Add task action "Death Dismissal" to LegacyAppeals


hschallhorn:
Specifically for FNOD - Deceased veterans are being scheduled hearings (receiving letters, taking up hearing spots for other vets). As VACOLS users are no longer able to move cases out of caseflow with vacols, we need to give them this option within the caseflow application.
Selecting this action will change the case's location in VACOLS to location 66 and cancel all open tasks for judge to start working on dismissing the appeal. 

[Relevant slack words](https://dsva.slack.com/archives/CJL810329/p1568913385077700?thread_ts=1568912922.076400&cid=CJL810329).

## AC
- [x] Display "Death Dismissal" task action for legacy appeals with
  - [x] VACOLS location code of "CASEFLOW"
  - [x] A non nil value for `SFNOD` on the veteran's `CORRES` record
- [x] Selecting this action will confirm with the user that all open tasks will be cancelled and the appeal's location code will be updated to 66
- [x] Confirming this action
  - [x] Cancels all open tasks
  - [x] Updated the appeal's location code in VACOLS to 66

hschallhorn:
@laurjpeterson Anything to add on this?

laurjpeterson:
Looks good to me! Fixed minor wording in main entry.

kevmo:
1 |
2 |
3 | |||||||||
5 | |||
8 |

3.  Straight-forward, but involves VACOLS, lending uncertainty and some 5 votes.

hschallhorn:
When cancelling all tasks, ensure that we do not have situation where we leave zombie open tasks that may have been opened when a task is cancelled.

lomky:
- Who should be able to do this? 
   - The Judge on the case. Not the Attorney.
 - When should the Judge be able to do this? Anytime the case is at the Judge or Attorney.

PR #12387, opened 2019-10-15
LegacyAppeal Dismissal Due to FNOD

lomky:
Completes #12211 

### Description

Adds the ability to do a Death Dismissal action on Legacy Appeals when the VACOLs SFNOD is set to a date.

### Acceptance Criteria
- [ ] Colocated Admins can do the action for any Legacy Appeal with an SFNOD assigned to the colocated team

### Testing Plan
1. Mark a LegacyAppeal with an open Colocated task with a SFNOD on the console
    `appeal.notice_of_death_date = 5.days.ago`
2. Sign in as a Colocated Admin & navigate to that appeal
3. Confirm you have the option `Death Dismissal` in the drop down
4. Confirm the action pops up a confirmation modal
5. Submitting the death dismissal results in a success message
6. On the console, confirm the appeal location is now 66 & the appeal tasks are all cancelled.


config/routes.rb
```diff
@@ -122,6 +122,7 @@
       resources :special_issues, only: [:create, :index]
       resources :advance_on_docket_motions, only: [:create]
       get 'tasks', to: "tasks#for_appeal"
+      post 'death_dismissal', to: "appeals#death_dismissal"
```
hschallhorn: this line can just be `post :death_dismissal`. And then your test would be `subject { post :death_dismissal }` as well.

app/models/legacy_tasks/attorney_legacy_task.rb
```diff
@@ -2,7 +2,11 @@
 
 class AttorneyLegacyTask < LegacyTask
   def available_actions(current_user, role)
-    return [] if role != "attorney" || current_user != assigned_to
+    if assigning_judge(current_user, role) && appeal.notice_of_death_date
+      return [Constants.TASK_ACTIONS.DEATH_DISMISSAL.to_h]
```
yoomlam: So when a NOD occurs, the judge has no other action except to DEATH_DISMISSAL. Why not do it automatically and just notify the judge?

app/models/legacy_tasks/attorney_legacy_task.rb
```diff
@@ -23,4 +27,14 @@ def timeline_title
...
+
+  private
+
+  def assigning_judge(current_user, role)
+    role == "judge" && current_user.id == assigned_by.pg_id
+  end
+
+  def not_assigned_attorney(current_user, role)
+    role != "attorney" || current_user != assigned_to
+  end
```
yoomlam: Thank you for making the intent explicit.

app/models/legacy_tasks/attorney_legacy_task.rb
lomky: It's SFNOD or FNOD; NOD is already used for Notice of Disagreement :smile: (I don't know what the SF or F stand for...).

Mostly because that wasn't the ask. Judges actually have the ability to do this in VACOLS, as well, and this mirrors the option. I'm not sure we want to be pulling things out of users work queues without their knowledge, as we don't really have a way to inform them what happened.

lomky: We should be careful to call it SFNOD or FNOD; NOD is already used for Notice of Disagreement :smile: (I don't know what the SF or F stand for...).

Mostly because that wasn't the ask. Judges actually have the ability to do this in VACOLS, as well, and this mirrors the option. I'm not sure we want to be pulling things out of users work queues without their knowledge, as we don't really have a way to inform them what happened.

lowellrex: I believe FNOD stands for "final notice of death". The VACOLS convention for naming columns is to prefix the column name with the first letter or two of the table name (like `brieff.bfcurloc`)... That doesn't necessarily explain why all of the `corres` column names are prefixed with an "s" though...

app/repositories/appeal_repository.rb
```diff
@@ -358,6 +358,11 @@ def self.update_location_after_dispatch!(appeal:)
     appeal.case_record.update_vacols_location!(location)
   end
 
+  def self.update_location_for_death_dismissal!(appeal:)
+    location = "66"
```
lowellrex: What do you think about referencing the element of the `LegacyAppeal::LOCATION_CODES` [we added in this PR](https://github.com/department-of-veterans-affairs/caseflow/pull/12387/files?file-filters%5B%5D=.json&file-filters%5B%5D=.jsx&file-filters%5B%5D=.rb#diff-5a94be9eb6a667117856c5bd2d145da7R161) instead of the literal "66" here?

app/models/legacy_tasks/attorney_legacy_task.rb
```diff
@@ -2,7 +2,11 @@
 
 class AttorneyLegacyTask < LegacyTask
   def available_actions(current_user, role)
-    return [] if role != "attorney" || current_user != assigned_to
+    if assigning_judge(current_user, role) && appeal.notice_of_death_date
```
lowellrex: It looks like we have similar logic to determine if an appeal is eligible for a death dismissal (and if a given user can execute that death dismissal) in a few different tasks. What do you think about making that interrogation of the appeal explicit? Perhaps with a method that looks something like:

```ruby
# appeal.rb/legacy_appeal.rb
def eligible_for_death_dismissal?(user)
  user_has_actionable_open_tasks = tasks.open.where(assigned_to: user) || tasks.open.where(assigned_by: user, type: [AttorneyTask.name, ...])
  !notice_of_death_date.nil? && user_has_actionable_open_tasks
end 
```

lowellrex:
Since death dismissal happens to an entire appeal and not any specific task, I wonder if it makes sense to implement the task action similar to what we've done for [incoming mail on `RootTask`s](https://github.com/department-of-veterans-affairs/caseflow/blob/edc8f49bfdd1a00219c9a07f564d37b3980e91c3/app/models/tasks/root_task.rb#L31). Instead of adding a task action for death dismissal to a bunch of different task types, we could instead add it to the `RootTask` and make that action available when applicable through some means other than the task actions dropdown (perhaps we could re-use the ... I think [the existing front-end logic](https://github.com/department-of-veterans-affairs/caseflow/blob/edc8f49bfdd1a00219c9a07f564d37b3980e91c3/client/app/queue/TaskSnapshot.jsx#L37) might just work).

image

lowellrex:
I'm finding this PR a little difficult to digest and wonder if it might be possible to break this up into several smaller, more bite-sized PRs that tackle a single piece of this ticket. That would allow the reviewers to focus on the content of the changes and make sure the code coverage was thorough while still ensuring the app is always deployable. One division that seems sensible to me is below:

1. Determination of whether an appeal is eligible for a death dismissal for a given user
    * I think this encompass most of the files in `app/models/tasks`
2. Death dismissal logic (location code change, task cancellation)
    * `LegacyAppeal` and `AppealRepository`
3. API routes and controllers
4. Front-end components
5. Adding the `DEATH_DISMISSAL` task action to the various tasks

@lomky What do you think of that approach?

PR #12398, opened 2019-10-16
add to the make file's clean directive commands to remove node_modules and packag…

edmangimelli:
very tiny update to Makefile.example

do those 2 rm statements belong in clean? or should belong in some separate, more forceful clean directive?

resolves https://github.com/department-of-veterans-affairs/caseflow/issues/12399


Makefile.example
```diff
@@ -60,6 +60,8 @@ clean:  ## Clean logs, compiled webpack assets, and temporary files
 	rm -f app/assets/javascripts/*webpack*
 	rm -rf public/assets
 	rm -rf tmp/capybara
+	rm -rf client/node_modules
```
pkarman: maybe a new target? `make realclean` is one convention for this kind of destructive move.

something like:

```
realclean: clean
    rm -rf client/node_modules
    rm -f client/package-lock.json
```

Issue #12316, opened 2019-10-08
Create VirtualHearing Class


lomaxap:
### Background
related to #12315

### Objective
Create an ActiveRecord class that allows us to query virtual_hearings. We should be able to access virtual hearing data from the parent hearing with `hearing.virtual_hearing`. We should also be able to test if a hearing is virtual with `hearing.virtual?` where there is a related `virtual_hearing` record that is `active`

Hearing serializers should return boolean values for `virtual` attribute

`VirtualHearing.active` should [scope](https://api.rubyonrails.org/classes/ActiveRecord/Scoping/Named/ClassMethods.html) virtual hearings to those where `active: true` 

For legacy_hearings, after a virtual hearing is created, the `hearing_type` in `VACOLS` should be updated to `VIRTUAL`

ferristseng:
> For legacy_hearings, after a virtual hearing is created, the hearing_type in VACOLS should be updated to VIRTUAL

@lomaxap Testing this, and it looks like `hearing_type` has to be 1 character. Any ideas for what this should be? cc @pshahVA

PR #12417, opened 2019-10-17
Add Virtual Hearings Model

ferristseng:
Related to #12316 

### Description

  - Add model for virtual hearings table
  - Add associations b/w virtual hearings and hearing/legacy hearing
  - Add `VirtualHearingRepository`, and method to find virtual hearings that can be deleted from Pexip
  - Add tests for `VirtualHearingRepository`
  - Update serializer to return `virtual` attribute

### AC

  - [x] Create an ActiveRecord class that allows us to query virtual_hearings. We should be able to access virtual hearing data from the parent hearing with `hearing.virtual_hearing`. 
  - [x] We should also be able to test if a hearing is virtual with `hearing.virtual?` where there is a related virtual_hearing record that is active
  - [x] Hearing serializers should return boolean values for virtual attribute
  - [x] `VirtualHearing.active` should scope virtual hearings to those where active: true
  - [x] ~For legacy_hearings, after a virtual hearing is created, the hearing_type in VACOLS should be updated to `VIRTUAL`~


app/repositories/virtual_hearing_repository.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class VirtualHearingRepository
+  class << self
+    def ready_for_deletion
```
lomaxap: 👍

spec/models/concerns/has_virtual_hearing_examples.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+shared_examples_for "a model that can have a virtual hearing" do
```
lomaxap: 😮

PR #12419, opened 2019-10-17
Update reassign function

### LINKED ISSUE #12365: Update the task reassign function
[As found](https://github.com/department-of-veterans-affairs/caseflow/issues/12173#issuecomment-541136486) when reassigning tasks in #12173, There are two pitfalls with our current reassign function.
1. If the task to be reassigned has a child `TimedHoldTask`, it will be [cancelled](https://github.com/department-of-veterans-affairs/caseflow/blob/e1a5d717fb51a34aeac10f276fc86a895266969f/app/models/task.rb#L581) before it is [moved under the new task](https://github.com/department-of-veterans-affairs/caseflow/blob/e1a5d717fb51a34aeac10f276fc86a895266969f/app/models/task.rb#L479).
2.  If the task to be reassigned has children, and is therefore on hold, the new task is not put on hold

## AC
- [ ] When a task is reassigned, children `TimedHoldTask`s are correctly moved under the new task
- [ ] When a task is reassigned, children `TimedHoldTask`s are not cancelled
- [ ] When a task with children is reassigned, the new task is put on hold

hschallhorn:
Resolves #12365

### Acceptance Criteria
- [ ]  When a task is reassigned, children TimedHoldTasks are correctly moved under the new task
- [ ]  When a task is reassigned, children TimedHoldTasks are not cancelled
- [ ]  When a task with children is reassigned, the new task is put on hold


app/models/task.rb
```diff
@@ -474,9 +474,10 @@ def reassign(reassign_params, current_user)
       task.save!
     end
 
-    update!(status: Constants.TASK_STATUSES.cancelled)
+    sibling.on_hold! if children.open.any?
+    children.open.update_all(parent_id: sibling.id)
```
lowellrex: Could we add a code comment about why we are doing this (specifically updating the status if there are children tasks) and/or maybe rearrange this code which might make the intent clearer?

```ruby
if children.open.any?
  sibling.on_hold!
  children.open.update_all(parent_id: sibling.id)
end
```

app/models/task.rb
```diff
@@ -474,9 +474,13 @@ def reassign(reassign_params, current_user)
       task.save!
     end
 
-    update!(status: Constants.TASK_STATUSES.cancelled)
+    # Preserve the open children and on hold status of the old task
+    if children.open.any?
+      children.open.update_all(parent_id: sibling.id)
+      sibling.on_hold!
```
lowellrex: One other thought, can we just copy over the previous task's status?

hschallhorn: Yeah I was thinking about that as well. Couldn't come up with any scenarios where we wouldn't want that to happen

PR #12427, opened 2019-10-18
Save end product claim type code updates

leikkisa:
### Description
This table is to save end product claim_type_code updates when they are changed outside of Caseflow manually (in Share).  These are causing us errors in processing remand supplemental claims, and are potentially causing eligibility check errors and other issues as well.

The initial purpose of this table is to be able to do some investigation and reporting on the problem.  We do not yet have solutions for how to resolve the problems, but it will hopefully useful for implementing solutions. If not, it may be a temporary table.

### Documentation Updates

*Schema Changes Only*

* [x] Column comments updated


leikkisa:
Regarding language should I more consistently use the label "claim_type_code"?  We just call it "code" in end product establishments. I have been calling it end product code, but am open to changing it to claim_type_code if that makes more sense.

Issue #12420, opened 2019-10-17
SCM: Extend the organizations_user with judge & attorney roles 


lomky:
In order to allow others to act on behalf of Judges, add clear roles to members of Judge Teams.

## Acceptance Criteria
 - [ ] `organizations_users` table has a column `judge_team_role`, valid values of nil, `judge`, and `attorney`
 - [ ] Judge teams can only have only have one judge

This ticket _does not_ include updating the code base to do anything with these fields, nor updating the existing teams.

Part of #11801

kevmo:
Estimating at 1.

lowellrex:
@lomky Love this idea! Super pumped to stop overloading the `admin` column for `JudgeTeam`s!

How do you feel about moving the new `judge_team_role` column into a different table since it will only apply to judge teams? That field will only have values for `OrganzationUser`s associated with `JudgeTeam`s and right now less than a third of the rows in that table are associated with `JudgeTeam`s. That proportion will likely decrease over time since we will add more teams and organizations, and judges and attorneys are already well populated in that table.

```ruby
rails c> OrganizationsUser.count
# 3144
rails c> OrganizationsUser.where(organization_id: JudgeTeam.pluck(:id)).count
# 954
```

What do you think of a table definition that looks like the following:
```sql
judge_team_users (
  organziations_users_id    int -- could instead use organization and user IDs
  role                      enum(judge, attorney)
)
```

Some advantages to this approach include:
1. Avoiding `null` roles. If a user is in a judge team they will have a role.
2. All column in `OrganizationsUser` are relevant to all rows. Currently there are only three columns in this table that are not used for linking the user and the organization. `created_at` and `updated_at` provide timestamps and `admin` is a boolean which describes whether someone is has administrative privileges or not. All organizations make use of administrative privileges in some way (user management), so this applies to all rows in the table.

lowellrex:
We're going to proceed as follows:
1. Create new `JudgeTeamRole` single-table inheritance models (slightly modified approach described above).
2. Start creating these objects in the application code.
3. Backfill data where `JudgeTeamRole`s should exist.
4. Cut over to identifying judges and attorneys using `JudgeTeamRole`s (#12422)

lomky:
@lowellrex number 3 is a follow-on ticket, as well. #12421

lowellrex:
Closing this ticket since all remaining work will happen in #12481, #12421, and #12422.

PR #12449, opened 2019-10-22
Create JudgeTeamRole table

lowellrex:
Connects #12420. Creates new `JudgeTeamRole` single-table inheritance models.

### DIFF
--- a/db/migrate/20191021215017_create_judge_team_roles.rb
+++ b/db/migrate/20191021215017_create_judge_team_roles.rb
@@ -1,0 +1,10 @@
+class CreateJudgeTeamRoles < ActiveRecord::Migration[5.1]
+  def change
+    create_table :judge_team_roles do |t|
+      t.string :type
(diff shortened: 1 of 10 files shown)


db/migrate/20191021215017_create_judge_team_roles.rb
```diff
@@ -0,0 +1,10 @@
+class CreateJudgeTeamRoles < ActiveRecord::Migration[5.1]
```
lomky: Could we add comments for the table & columns?

lowellrex: Great call. Will add them in a follow-on PR (https://github.com/department-of-veterans-affairs/caseflow/issues/12420#issuecomment-545043363).

lowellrex: Actually just adding that now so we can kick off a new build.

Issue #12318, opened 2019-10-08
MTV - Add option for partial grant and select issues


leikkisa:
### Story
Sometimes, judges partially grant Veterans' motions to vacate. This means that only some, not all, of the issues on the original appeal decision will be vacated and decided again on the new decision.

Judges (or potentially other users, to be determined) need to select some, not all, issues that are vacated, to be decided on the new decision. (And then, follow the typical assign to attorney, draft decision, attorney checkout, judge checkout etc. path.)

### Acceptance criteria
- On the Judge Address Motion to Vacate modal, a new option is available to allow for a partial grant
- When the judge selects the partial grant option, a list of decision issue descriptions from the previous appeal are populated. 
- Users can select from that list of issues

### Out of scope
- Action on submit
- Additional processing on the selected issues
- Connectivity to timeline display data is out of scope
- Vacated issues get a new request issue created for them on the same appeal - https://github.com/department-of-veterans-affairs/caseflow/issues/12373
   - The vacated decisions are connected to the new request issue with vacated_by_request_issue_id
   - The request issues get assigned to a decisions attorney to write, to make new decisions on

### Design
![Screen Shot 2019-10-08 at 11.32.06 AM.png](https://images.zenhubusercontent.com/58f6b6eb51aa2c30577bb60e/8f955f9a-4aa3-448b-9880-145d075220d4)

### Notes
Ticket to pull motions to vacate issues only into this workflow handled under a separate ticket

laurjpeterson:
@leikkisa @jimruggiero @geronimoramos @rutvigupta-design - i updated this ticket for clarity. 

If a judge selects "partial vacatur", all issues from the previously decided decision (that has the request for vacatur) should show in the checkboxes you display in the design. This way, the judge can choose which of those issues are actually being vacated. 

So the flow would look something like this. Does this make sense?

- An appeal with 4 issues was decided. 
- Veteran sends in a request for motion to vacate.
- Motions attorney assigns MTV to judge
- Judge decides that only 3 of the 4 warrant vacatur - and thus chooses partial vacatur
- Judge selects those 3 issues from the screen above, and only those 3 issues become request issues in the subsequent new decision
- Judge assigns task to attorney (either Vacate and De Novo or Straight Vacate and Readjudication) so a new decision, with just the 3 vacated issues, as request issues

leikkisa:
[...]

PR #12454, opened 2019-10-22
MTV | Add option for partial grant and select issues

jcq:
Connects #12318

### Description
Updated various bits related to judge "Address Motion to Vacate" view to support "partial vacatur" as a disposition. This also enables selection of issues in case of "partial" disposition.

This PR also adds backend functionality for allowing "partial" disposition so that things don't break. Backend handling of the selected issues will be handled in a future PR.

### Acceptance Criteria
- [ ] Code compiles correctly
- [ ] Address Motion to Vacate supports `partial` disposition
- [ ] Address Motion to Vacate displays and mandates selection of one or more issues when `partial` disposition is selected


jcq:
### Screenshot of Updated UI
Screen Shot 2019-10-28 at 17 40 48

client/app/queue/mtv/MTVJudgeDisposition.jsx
```diff
@@ -112,16 +124,24 @@ export const MTVJudgeDisposition = ({
...
+          <MTVIssueSelection
+            issues={appeal.issues}
+            onChange={({ issueIds: newIssueIds }) => setIssueIds(newIssueIds)}
+          />
+        )}
+
         {disposition && disposition === 'denied' && <MissingDenialDraftAlert to={JUDGE_RETURN_TO_LIT_SUPPORT.value} />}
 
-        {disposition && disposition === 'granted' && (
+        {disposition && isGrantType() && (
```
leikkisa: I suppose you could just put isGrantType() here (and the other places it's used) since that function is checking disposition

client/app/queue/mtv/MTVJudgeDisposition.jsx
```diff
@@ -112,16 +124,24 @@ export const MTVJudgeDisposition = ({
...
           onChange={(val) => {
             setVacateType(null);
             setDisposition(val);
           }}
           value={disposition}
         />
 
+        {disposition && disposition === 'partial' && (
+          <MTVIssueSelection
+            issues={appeal.issues}
```
leikkisa: This should be the appeal's decision_issues. It looks like issues returns both decision issues and request issues, but the request issue is what is showing on the screen.

jcq: Aha! That's an important distinction. Thanks for catching!

client/app/queue/mtv/MTVJudgeDisposition.jsx
jcq: True, that could shorten a couple of them. The logic wouldn't work for the `!isGrantType()` one, so I'm sorta inclined to just keep them all consistent, but I agree that shortening logic like this is a general positive

client/app/queue/mtv/MTVDispositionSelection.jsx
```diff
@@ -2,9 +2,13 @@ import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import RadioField from '../../components/RadioField';
 
-import { DISPOSITION_OPTIONS } from '../../../constants/MOTION_TO_VACATE.json';
+import { DISPOSITION_OPTIONS, PARTIAL_GRANT_OPTION } from '../../../constants/MOTION_TO_VACATE.json';
 
-export const MTVDispositionSelection = ({ label = '', value: initialVal = null, onChange }) => {
+// In certain cases (judge view) we want to include a "partial grant" option
```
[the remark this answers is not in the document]
jcq: The "partial" option is only applicable to judge view, so we want to keep it out of the attorney view where this component is also used. And, since it's being inserted into the middle, we have this slightly annoying logic rather than just adding to the end of the array.

We could certainly combine them if we added additional properties to the objects that specify if the given option is applicable to judge/atty view, but I thought for the limited cases this might be clearer. Happy to change if you think otherwise though!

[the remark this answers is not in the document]
leikkisa: Ah okay, thanks for clearing that up for me.

PR #12459, opened 2019-10-22
Lighthouse pb implement hlr show endpoint

### LINKED ISSUE #12458: API v3 - show endpoint for HigherLevelReviews
### User story
As an api-consumer
I want to see details of my higher level review
So that I can inform my users about it's details

### Acceptance criteria
Returns the details of the Higher Level Review

Implements endpoint designed here https://github.com/department-of-veterans-affairs/caseflow/issues/12001

Ticket from vets-contrib: https://app.zenhub.com/workspaces/developervagov-5ac272fff41bfa50c0e46c48/issues/department-of-veterans-affairs/vets-contrib/2728

erluti:
Resolves #12458

### Description
Adding serializers and `show` endpoint for higher_level_reviews controller.
Made a few small changes to accommodate the serializers.
Small tweak to docs to reflect name change of attribute.


spec/requests/api/v3/decision_review/higher_level_reviews_controller_spec.rb
```diff
@@ -107,21 +103,212 @@
...
-      { params: params, headers: { "Authorization" => "Token #{api_key}" } }
-    ]
+  describe "#show" do
+    let!(:higher_level_review) do
+      processor = Api::V3::DecisionReview::HigherLevelReviewIntakeProcessor.new(
+        ActionController::Parameters.new(data: data, included: included),
+        create(:user)
+      )
+      processor.run!
+      processor.higher_level_review
```
edmangimelli: for chaining purposes, i made `run!` return the processor.
so you could have written `processor.run!.higher_level_review`
(not saying that's better. just saying it's a possibility)

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -971,7 +971,7 @@ paths:
                                   type: string
                                   format: date
                                   nullable: true
-                                ratingDecisionReferenceId:
+                                ratingDecisionId:
```
pkarman: is it important for this name to match some kind of JSON API spec? Deviating from the model name will probably confuse Caseflow developers later.

app/serializers/api/v3/request_issue_serializer.rb
```diff
@@ -0,0 +1,31 @@
...
+class Api::V3::RequestIssueSerializer
+  include FastJsonapi::ObjectSerializer
+  set_key_transform :camel_lower
+
+  self.record_type = "RequestIssue"
+
+  attributes :diagnostic_code, :description, :contention_text, :notes,
+             :is_unidentified, :ramp_claim_id, :ineligible_reason,
+             :ineligible_due_to_id, :withdrawal_date, :contested_issue_description,
+             :end_product_code, :title_of_active_review
```
pkarman: I'm not sure `title_of_active_review` is what you want here. Most of the time it's going to be `nil` unless the request issue was ineligible. In that case, maybe the API field name would be more helpful if it included the word "ineligible" in it somehow, or was conditional based on `ineligible_due_to_id` or something.

app/controllers/api/docs/v3/decision_reviews.yaml
erluti: We were trying to have variable names that are easier for the API consumers to understand.  Does `Reference` distinguish it from another kind of Id?

pkarman: yes the "reference" convention in our schema means it is a pointer to a record in another system. So a `foo_reference_id` indicates that `foo` is elsewhere, whereas `foo_id` indicates that the tables `foos` has a record with a matching `id` value.

app/serializers/api/v3/request_issue_serializer.rb
pkarman: again, that would vary based on the value of `ineligible_reason` -- do you want to have different object names for each potential reason?

If you wanted to combine them, how about:

```
ineligible: {
  due_to_id: 123,
  reason: "something",
  title_of_active_review: "foo"  # might be nil
}
```

app/controllers/api/docs/v3/decision_reviews.yaml
erluti: I think to the consumer of the API they aren't invested in where the id comes from as long as the API respects the ids provided.  The name is already changed from its internal name and in other places we've dropped the word "reference", so I'd like to drop it here too.

app/serializers/api/v3/request_issue_serializer.rb
erluti: I went with your suggestion to include the reason in an ineligible object.  Thanks for the help, I really like how this turned out.

Issue #12408, opened 2019-10-16
Decision issue - convert ui_hash to serializer


leikkisa:
### Description
As a tech improvement ticket, in order to improve consistency and code quality, and reduce the size of our models, we can convert ui_hash method to use serializers instead.

### Acceptance criteria
 - Decision issues use a serializer instead of ui_hash

### Background
This is a small conversion to start. We'll be expanding this to the other ui_hash methods in the code

PR #12460, opened 2019-10-23
WIP| Decision issue - convert ui_hash to serializer

Sjones352:
Connects #12408

### DIFF
--- a/app/serializers/intake/decision_issue_serializer.rb
+++ b/app/serializers/intake/decision_issue_serializer.rb
@@ -1,0 +1,8 @@
+class DecisionIssueSerializer
+  include FastJsonapi::ObjectSerializer
+  attribute :id, &:id
+  attribute :requestIssueId, &:request_issues&.first&.id
+  attribute :description, &:description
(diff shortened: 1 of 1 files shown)


app/models/decision_issue.rb
```diff
@@ -105,14 +105,8 @@ def remanded?
...
 
-  def ui_hash
-    {
-      id: id,
-      requestIssueId: request_issues&.first&.id,
-      description: description,
-      disposition: disposition,
-      approxDecisionDate: approx_decision_date
-    }
+  def decision_issue_serializer
```
leikkisa: I am still not loving this name for this, let's keep it named `ui_hash` for now until we think of a better pattern

app/serializers/decision_issue_serializer.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class DecisionIssueSerializer
```
leikkisa: Can you add a top line comment for this new file?

PR #12468, opened 2019-10-23
Update functions.md

rajinderbhanot:
Moved to wiki


docs/functions.md
```diff
@@ -1,18 +1,2 @@
...
-
-For this particular case in certification, demo environment, it would grant CF_Q_283 and VHAISASOMEONE users and it would deny CF_SCOTTY_283 with System Admin function.
-To add or deny function to a certain user, you would have to add the user's CSS_ID to this file, make a PR for this code change and merge it. Make sure you use the right environment(s) and that you are preserving the existing list (if that's your intention). The empty list will remove all users who were granted/denied the function. Also, the same CSS_ID cannot be both granted and denied the same function.
-
-Note: For the function/roles, Caseflow currently ignores System Admin function from CSUM/CSEM users.
-
-For manual handling of Functions check out Commons project:
-https://github.com/department-of-veterans-affairs/caseflow-commons
+moved to wiki
+https://github.com/department-of-veterans-affairs/caseflow/wiki/functions
```
leikkisa: I think putting a generic link here may be more future proof, because if the wiki page name changes, this link will break:
`[This page has migrated to the Caseflow wiki](https://github.com/department-of-veterans-affairs/caseflow/wiki)`

pkarman: Possibly more future proof but not very user friendly. The chances that the URL changes are non-zero but much much lower than the pain caused by having to follow a link and then search again for the page.

PR #12474, opened 2019-10-23
Folder org, topline and folder comments

leikkisa:
### Description
- Adds model folders for: Intake, Stats
  - For models and model specs
- Moves one dispatch file into the dispatch folder
- Adds some topline comments
- Adds some folder readme files


pkarman:
@leikkisa should we press on with this or re-approach?

pkarman:
@leikkisa this is a good idea. maybe break it down into smaller PRs to get merged?

leikkisa:
@pkarman Oh this is from the offsite. I think that's a good idea - I'll go ahead and close this PR. I think it makes sense to do the topline comments separate from folder re-org.

Issue #12481, opened 2019-10-25
Create JudgeTeamRole objects in application


lowellrex:
Follow-on ticket from #12420. That ticket created the tables and classes, this ticket exists to create those objects in the application code. 

## Acceptance criteria
* [x] Create `JudgeTeamLead` wherever we currently [create a "judge" for a `JudgeTeam`](https://github.com/department-of-veterans-affairs/caseflow/blob/c82d3f8e65ad7d5d079468ff3a79964bfbaddbdd/app/models/organizations/judge_team.rb#L8). (#12484)
    * Assume judge will be `JudgeTeamLead`
* [x] Whenever an attorney is added to a `JudgeTeam`, create a corresponding `DecisionDraftingAttorney` instance. (#12493)
    * Assume non-judges will be `DecisionDraftingAttorney`
* [x] Ensure `JudgeTeam`s have only one `User` with the `JudgeTeamLead` role. (#12497)
* [x] Ensure `JudgeTeamLead`s are always identified as admins in the associated `OrganizationsUser` model. (#12497)

Since we do not ever expect to find the application in the state described by these last two acceptance criteria (that is, we don't expect the application to ever have a `JudgeTeam` with anything other than a single `JudgeTeamLead`), consider implementing these acceptance criteria as [data integrity checks](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/jobs/data_integrity_checks_job.rb). Implementing these as data integrity checks will allow the application to continuing functioning while we investigate how the application ended up in this state.

lomky:
Estimating while in flight at a 2

yoomlam:
All tickets associated with Acceptance Criteria are complete.

yoomlam:
All associated sub-tickets have been merged.

PR #12484, opened 2019-10-25
Create JudgeTeamLead when creating JudgeTeams for judges

lowellrex:
Connects #12481. Implements 1 part of 4 for the base ticket by creating a `JudgeTeamLead` instance associated with the `OrganizationsUser` object when a new `JudgeTeam` is created for a `User`.

### DIFF
--- a/app/models/organizations/judge_team.rb
+++ b/app/models/organizations/judge_team.rb
@@ -6,6 +6,8 @@ class JudgeTeam < Organization
   end
 
   def self.create_for_judge(user)
+    fail(Caseflow::Error::DuplicateJudgeTeam, user_id: user.id) if JudgeTeam.for_judge(user)
+
     create!(name: user.css_id, url: user.css_id.downcase).tap do |org|
       org_user = OrganizationsUser.make_user_admin(user, org)
       JudgeTeamLead.create!(organizations_user: org_user)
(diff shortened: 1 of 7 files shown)


app/models/organizations/judge_team.rb
```diff
@@ -6,8 +6,11 @@ def self.for_judge(user)
   end
 
   def self.create_for_judge(user)
+    fail(Caseflow::Error::DuplicateJudgeTeam, user_id: user.id) if JudgeTeam.for_judge(user)
```
lowellrex: Added for extra safety. Not directly related to this PR.

app/models/organizations_user.rb
```diff
@@ -13,7 +13,9 @@ def self.add_user_to_organization(user, organization)
   end
 
   def self.make_user_admin(user, organization)
-    add_user_to_organization(user, organization).update!(admin: true)
+    add_user_to_organization(user, organization).tap do |org_user|
+      org_user.update!(admin: true)
+    end
```
lowellrex: Made this change (and added [the related test](https://github.com/department-of-veterans-affairs/caseflow/pull/12484/files#diff-0569e73dfafe4bd28bab5c0d0d7260f4R40)) so that the `make_user_admin` function returns an instance of `OrganizationsUser` instead of the boolean `true` indicating that the `update!` method was successful.

app/models/organizations/judge_team.rb
```diff
@@ -6,8 +6,11 @@ def self.for_judge(user)
...
+
     create!(name: user.css_id, url: user.css_id.downcase).tap do |org|
-      OrganizationsUser.make_user_admin(user, org)
+      org_user = OrganizationsUser.make_user_admin(user, org)
+      JudgeTeamLead.create!(organizations_user: org_user)
```
lowellrex: Heart of the change here.

Issue #11607, opened 2019-07-30
Unify GenericTask and Task classes


lowellrex:
The `GenericTask` class was [created last summer](https://github.com/department-of-veterans-affairs/caseflow/pull/6713) to extend the `Task` model in a more... generic way than the existing `Task` subclasses so that we could quickly roll out queues to different work groups at the Board. Once all `Task` subclasses inherit from `GenericTask`, we should combine those two classes into a single class to remove an unnecessary class and reduce confusion around why they both exist and how they are different.

## Blocked by 
* [x] Refactor `JudgeTask` to inherit from `GenericTask` #7630
* [x] Refactor `ColocatedTask` to inherit from `GenericTask` #11605
* [x] Refactor `AttorneyTask ` to inherit from `GenericTask` #11606 

## Acceptance criteria
* [x] Move all logic in `GenericTask` into `Task`
* [x] All classes that currently inherit from `GenericTask` inherit from `Task` (#12485)
* [x] Stop creating `GenericTask`s in the application (see https://github.com/department-of-veterans-affairs/caseflow/issues/11607#issuecomment-546942653) (#12487)
* [x] Update tests and generators to create `GenericTask`s instead of `Task`s (#12487)
* [x] Change type of [833 `GenericTask`s in production](https://caseflow-looker.va.gov/sql/9ctxpwxvrbvbs3) to `Task`
* [x] Remove `GenericTask` class
* ~[ ] Rename [`:generic_task` factory](https://github.com/department-of-veterans-affairs/caseflow/blob/5453cdcfd5d9c1bf283fad51dc92160e4cd29de6/spec/factories/task.rb#L97) to `:task`~

lowellrex:
_[What is this graph?](https://github.com/department-of-veterans-affairs/caseflow/issues/11533)_
```
1 | ||||||
2 | |||
3 | 
5 | 
8 | 
```

Estimating this at 1. Possible 2 because we anticipate a fairly significant amount of test changes.

lowellrex:
Looks like our application is still creating `GenericTask`s (perhaps by way of [the `AssignToView`](https://github.com/department-of-veterans-affairs/caseflow/blob/5834be47e5145a31bf0f928a70b87d7d504f78d1/client/app/queue/AssignToView.jsx#L72)?). Adding that as an acceptance criteria.

[_Source: Looker_](https://caseflow-looker.va.gov/sql/4jpxsz3rcyfmjy)
[...]

PR #12487, opened 2019-10-28
Remove references to GenericTask

lowellrex:
Connects #11607. Removes references to `GenericTask` from application code and tests, and creates `Task`s instead of `GenericTask`s when assigning tasks without types.

### DIFF
--- a/spec/factories/task.rb
+++ b/spec/factories/task.rb
@@ -94,8 +94,8 @@ FactoryBot.define do
       end
     end
 
-    factory :generic_task, class: GenericTask do
-      type { GenericTask.name }
+    factory :generic_task, class: Task do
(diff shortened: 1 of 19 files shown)


spec/factories/task.rb
```diff
@@ -94,8 +94,8 @@
       end
     end
 
-    factory :generic_task, class: GenericTask do
-      type { GenericTask.name }
+    factory :generic_task, class: Task do
```
kevmo: It could be worth considering whether this factory should continue to be called `generic_task`.

lowellrex: Agreed. I think it's worth doing that in a different PR though. Will add it as an acceptance criteria on the main ticket.

PR #12488, opened 2019-10-28
Controllers for Virtual Hearings

### LINKED ISSUE #12332: Add Virtual Hearings Parameters to Hearings Controller
### Background
When a user switches a hearing from central or video to virtual, we need to store information related to the virtual hearing and kick off an async process to send emails and schedule a conference via Pexip API

### Objective
Modify Hearings controller to handle parameters for virtual hearings that allow us to `create`, `update`, or "`delete`" virtual hearings.

---
 hearing controller should now accept the following params
```ruby
{
  hearing: {
    virtual_hearing: {
      veteran_email: "",
      judge_email: "",
      representative_email: ""
    }
  }
}
```

create should `find_or_create_by` the `hearing_id` and `hearing_type`

when a virtual hearing is created for the first time, it should kick off th VirtualHearingConferenceJob (#12331)

for an update, if any of the emails have changed:

```
veteran_email: "",
judge_email: "",
representative_email: ""
```

we should update the corresponding `email_sent` column to `false` and invoke the `VirturalHearingsConferenceJob` to resend emails.

---

to "delete" a virtual_hearings, a `active: false` parameter should be sent to the hearings controller,
all `email_sent` columns should be set to `false` and a `VirtualHearingsConferenceJob` should be kicked off to send cancellation emails and delete the Pexip conference

### Considerations

This may require a large refactor of this controller to be able to handle more parameters. Consider using a [`FormResponse`](https://forum.upcase.com/t/form-objects/2267) (see also [this](https://thoughtbot.com/blog/activemodel-form-objects)) to handle controller logic. 

I can think of a handful of scenarios where we may not to send emails and cancel conferences outside of the controller. For that reason, we should consider writing all logic that invokes the `VirtualHearingsConferenceJob` to methods on the `VirtualHearing` class

ferristseng:
Resolves #12332 

### Description

  * Add `validates_email_format_of` gem
  * Refactor hearings controller update logic to `LegacyHearingUpdateForm` and `HearingUpdateForm`
  * Refactor logic to extract params from request body
  * Add logic to create a virtual hearing from the hearings controller

### DIFF
--- a/app/controllers/hearings_controller.rb
+++ b/app/controllers/hearings_controller.rb
@@ -17,7 +17,7 @@ class HearingsController < HearingsApplicationController
   end
 
   rescue_from ActiveRecord::RecordNotUnique do |_error|
-    render json: { "errors": ["message": "Virtual hearing already exists", code: 1002] }, status: :conflict 
+    render json: { "errors": ["message": "Virtual hearing already exists", code: 1002] }, status: :conflict
   end
 
   def show
@@ -25,10 +25,10 @@ class HearingsController < HearingsApplicationController
   end
 
   def update
-    if hearing.is_a?(LegacyHearing)
-      form = LegacyHearingUpdateForm.new(update_params_legacy)
+    form = if hearing.is_a?(LegacyHearing)
+      LegacyHearingUpdateForm.new(update_params_legacy)
     else
-      form = HearingUpdateForm.new(update_params)
+      HearingUpdateForm.new(update_params)
     end
     form.update
 
--- a/app/controllers/hearings_controller.rb
+++ b/app/controllers/hearings_controller.rb
@@ -16,44 +16,23 @@ class HearingsController < HearingsApplicationController
     render json: { "errors": ["message": error.message, code: 1001] }, status: :not_found
   end
 
+  rescue_from ActiveRecord::RecordNotUnique do |_error|
+    render json: { "errors": ["message": "Virtual hearing already exists", code: 1002] }, status: :conflict 
+  end
+
   def show
     render json: hearing.to_hash_for_worksheet(current_user.id)
   end
 
   def update
-    update_hearing
-    update_advance_on_docket_motion unless advance_on_docket_motion_params.empty?
-
-    if params.has_key?(:virtual_hearing) && !create_virtual_hearing
-      return render(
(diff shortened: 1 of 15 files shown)


app/controllers/hearings_controller.rb
```diff
@@ -11,7 +11,15 @@ class HearingsController < HearingsApplicationController
     render json: { "errors": ["message": error.message, code: 1000] }, status: :not_found
   end
 
-  rescue_from ActiveRecord::RecordInvalid, Caseflow::Error::VacolsRepositoryError do |error|
+  rescue_from ActiveRecord::RecordNotUnique do |_error|
+    render json: { "errors": ["message": "Virtual hearing already exists", code: 1003] }, status: :conflict
+  end
+
+  rescue_from ActiveRecord::RecordInvalid do |error|
+    render json: { "errors": ["message": error.message, code: 1002] }, status: :bad_request
```
lomaxap: what scenarios would this happen in? Wondering how we would display these differently on the front end and if we need to have more human friendly messages be sent back

app/controllers/hearings_controller.rb
```diff
@@ -74,65 +64,68 @@ def hearing_external_id
...
-                                       :problem_type, :requested_remedy,
-                                       :sent_to_transcriber_date, :task_number,
-                                       :transcriber, :uploaded_to_vbms_date
-                                     ],
-                                     hearing_issue_notes_attributes: [:id, :allow, :deny, :remand,
-                                                                      :dismiss, :reopen, :worksheet_notes])
+    params
+      .require(:hearing)
+      .permit(
+        *COMMON_HEARING_ATTRIBUTES,
```
lomaxap: 👍

app/models/hearings/forms/base_hearing_update_form.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+  end
+
+  protected
+
+  def update_hearing; end
+
+  private
+
+  def update_advance_on_docket_motion
```
lomaxap: this will actually only happen on AMA hearings (`advanced_on_docket_params` can only be sent from an AMA hearing's form), so you could move this to `hearing_update_form`

Issue #12450, opened 2019-10-22
Add Support for "Help" Text to RadioField


jcq:
### Description
We want the ability to optionally display "help" text under `RadioField` options

### Acceptance criteria
- Add new `help` prop to `RadioField` of type `string` that, if set, will display under the option
- Add example of it to styleguide 

### Designs

<img width="527" alt="mtv_radio_help" src="https://user-images.githubusercontent.com/2581274/67313290-bd6e2d00-f4d0-11e9-8951-a87d8cbc7359.png">

### Design Notes
The styles for this (provided by @rutvigupta-design) are:

```css
color: #5B616B
font-family: 'Source Sans Pro'
font-weight: 400
font-size: 17px
line-height: 1.5em
```

leikkisa:
<img width="412" alt="Screen Shot 2019-10-28 at 11 47 05 AM" src="https://user-images.githubusercontent.com/6589610/67708082-d0717900-f978-11e9-8f1b-471069570a92.png">

jcq:
### Screenshot of implementation:
<img width="381" alt="Screen Shot 2019-10-28 at 11 48 55" src="https://user-images.githubusercontent.com/2581274/67708330-365e0080-f979-11e9-8bdf-b3ca1e0214fb.png">

PR #12489, opened 2019-10-28
Add Support for "Help" Text to RadioField

jcq:
Connects #12450

### Description
Added optional support for "help" text for the `RadioField` component. This should be set via a `help` property on the objects passed via the `options` prop.

### Acceptance Criteria
- [ ] Styles for new "help" text are correct

### Testing Plan
1. Go to Caseflow Styleguide
2. Click on "Radio Buttons" option
3. Look at the "With help text" example

### DIFF
--- a/client/app/components/RadioField.jsx
+++ b/client/app/components/RadioField.jsx
@@ -12,13 +12,12 @@ import StringUtil from '../util/StringUtil';
  */
 
 export default class RadioField extends React.Component {
-
   isVertical() {
     return this.props.vertical || this.props.options.length > 2;
   }
 
   render() {
-    let {
+    const {
       id,
       className,
       label,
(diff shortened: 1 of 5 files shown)


client/app/containers/StyleGuide/RadioFieldExamples/Example5.jsx
```diff
@@ -0,0 +1,32 @@
+import React, { useState } from 'react';
```
leikkisa: You updated the StyleGuide! So much awesome

client/app/components/RadioField.jsx
```diff
@@ -12,13 +27,12 @@ import StringUtil from '../util/StringUtil';
...
 
 export default class RadioField extends React.Component {
-
   isVertical() {
     return this.props.vertical || this.props.options.length > 2;
   }
 
   render() {
-    let {
+    const {
```
leikkisa: Yay, thanks for getting these as always :-D

jcq:
### Screenshot of implementation:
Screen Shot 2019-10-28 at 11 48 55

Issue #12492, opened 2019-10-28
Prevent judges from being removed from their JudgeTeams


lowellrex:
Right now, any admin of a `JudgeTeam` can remove the judge associated with that `JudgeTeam`. This ticket exists to create a control in the application that prevents folks from doing that.

![remove-aabshire](https://user-images.githubusercontent.com/32683958/67708328-58b04800-f992-11e9-839c-7d02dca130c7.gif)
Admins should not be able to do this ☝️ 

## Acceptance criteria
* [ ] `JudgeTeam` admins can no longer remove the judge of a `JudgeTeam` from that team

hschallhorn:
Did we determine this would be front end or back end or both? I remember discussing it in estimation, but it looks like that did not make it into the ticket

lowellrex:
> Did we determine this would be front end or back end or both? I remember discussing it in estimation, but it looks like that did not make it into the ticket

Definitely on the back-end. Also on the front-end if we're going for extra credit.

PR #12493, opened 2019-10-28
Create DecisionDraftingAttorney rows when adding users to JudgeTeams

lowellrex:
Connects #12481. Whenever a user is added to a `JudgeTeam`, create a corresponding `DecisionDraftingAttorney` instance.

For now, we are assuming all users added to `JudgeTeam`s will be attorneys. This may change in the future when we roll out more functionality for DVCs. When we do that, we may need to modify this function.

### DIFF
--- a/app/models/organizations/judge_team.rb
+++ b/app/models/organizations/judge_team.rb
@@ -18,7 +18,8 @@ class JudgeTeam < Organization
   # Use the size of the organization to determine if we have just created a JudgeTeam for a judge,
   # or just added a user to an existing JudgeTeam. We assume the first user will always be a judge. All subsequent
   # members of the team will be attorneys.
-  def user_added_to_organization(org_user)
+  def add_user(user)
+    org_user = super
     class_name = (users.count == 1) ? JudgeTeamLead : DecisionDraftingAttorney
     class_name.create!(organizations_user: org_user)
   end
--- a/app/models/organizations/judge_team.rb
+++ b/app/models/organizations/judge_team.rb
@@ -9,15 +9,14 @@ class JudgeTeam < Organization
     fail(Caseflow::Error::DuplicateJudgeTeam, user_id: user.id) if JudgeTeam.for_judge(user)
 
     create!(name: user.css_id, url: user.css_id.downcase).tap do |org|
-      # Add the JudgeTeamLead record in org.user_added_to_organization which gets triggered by
-      # OrganizationsUser.inform_organization_user_added.
+      # Add the JudgeTeamLead record in org.add_user which gets triggered by OrganizationsUser.make_user_admin
       OrganizationsUser.make_user_admin(user, org)
     end
   end
 
   # Use the size of the organization to determine if we have just created a JudgeTeam for a judge,
-  # or just added a user to an existing JudgeTeam. We assume the first user will always be a judge. All subsequent
-  # members of the team will be attorneys.
+  # or just added a user to an existing JudgeTeam. We assume the first user will always be a judge.
+  # All subsequent members of the team will be attorneys.
   def add_user(user)
     super.tap do |org_user|
       class_name = (users.count == 1) ? JudgeTeamLead : DecisionDraftingAttorney
--- a/app/models/organizations/judge_team.rb
+++ b/app/models/organizations/judge_team.rb
@@ -9,9 +9,18 @@ class JudgeTeam < Organization
     fail(Caseflow::Error::DuplicateJudgeTeam, user_id: user.id) if JudgeTeam.for_judge(user)
 
     create!(name: user.css_id, url: user.css_id.downcase).tap do |org|
-      org_user = OrganizationsUser.make_user_admin(user, org)
-      JudgeTeamLead.create!(organizations_user: org_user)
+      # Add the JudgeTeamLead record in org.user_added_to_organization which gets triggered by
+      # OrganizationsUser.inform_organization_user_added.
+      OrganizationsUser.make_user_admin(user, org)
     end
+  end
+
+  # Use the size of the organization to determine if we have just created a JudgeTeam for a judge,
+  # or just added a user to an existing JudgeTeam. We assume the first user will always be a judge. All subsequent
+  # members of the team will be attorneys.
+  def user_added_to_organization(org_user)
+    class_name = (users.count == 1) ? JudgeTeamLead : DecisionDraftingAttorney
+    class_name.create!(organizations_user: org_user)
   end
 
   def judge
--- a/app/models/organizations/judge_team.rb
+++ b/app/models/organizations/judge_team.rb
@@ -19,9 +19,10 @@ class JudgeTeam < Organization
   # or just added a user to an existing JudgeTeam. We assume the first user will always be a judge. All subsequent
   # members of the team will be attorneys.
   def add_user(user)
-    org_user = super
-    class_name = (users.count == 1) ? JudgeTeamLead : DecisionDraftingAttorney
-    class_name.create!(organizations_user: org_user)
+    super.tap do |org_user|
+      class_name = (users.count == 1) ? JudgeTeamLead : DecisionDraftingAttorney
+      class_name.create!(organizations_user: org_user)
+    end
   end
 
   def judge
--- a/.reek.yml
+++ b/.reek.yml
@@ -14,6 +14,7 @@ detectors:
       - Veteran
   ControlParameter:
     exclude:
+      - Fakes::BGSService
       - HearingRepository#slot_new_hearing
       - Task#actions_allowable?
       - TaskSorter#sort_requires_case_norm?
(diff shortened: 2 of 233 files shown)


app/models/organizations/judge_team.rb
```diff
@@ -9,11 +9,20 @@ def self.create_for_judge(user)
...
-      JudgeTeamLead.create!(organizations_user: org_user)
+      # Add the JudgeTeamLead record in org.user_added_to_organization which gets triggered by
+      # OrganizationsUser.inform_organization_user_added.
+      OrganizationsUser.make_user_admin(user, org)
     end
   end
 
+  # Use the size of the organization to determine if we have just created a JudgeTeam for a judge,
+  # or just added a user to an existing JudgeTeam. We assume the first user will always be a judge. All subsequent
+  # members of the team will be attorneys.
```
lowellrex: These code comments are fairly verbose. Curious about your thoughts on keeping vs removing them?

app/models/organizations/judge_team.rb
```diff
@@ -9,11 +9,20 @@ def self.create_for_judge(user)
     fail(Caseflow::Error::DuplicateJudgeTeam, user_id: user.id) if JudgeTeam.for_judge(user)
 
     create!(name: user.css_id, url: user.css_id.downcase).tap do |org|
-      org_user = OrganizationsUser.make_user_admin(user, org)
-      JudgeTeamLead.create!(organizations_user: org_user)
+      # Add the JudgeTeamLead record in org.user_added_to_organization which gets triggered by
+      # OrganizationsUser.inform_organization_user_added.
+      OrganizationsUser.make_user_admin(user, org)
```
lowellrex: @hschallhorn curious about your thoughts on this approach. Adding the correct `JudgeTeamRole` record is quite indirect in this design. Do you think that is alright? Do any more optimal approaches occur to you?

lowellrex: One solution for this may be to allow `Organization`s more control over when users are added. Right now we add the relationship of a user to an organization by way of the `OrganizationsUser` class ([like in `Organization::UsersController`](https://github.com/department-of-veterans-affairs/caseflow/blob/5453cdcfd5d9c1bf283fad51dc92160e4cd29de6/app/controllers/organizations/users_controller.rb#L18) for example) instead of adding a user to an organization. Since the `OrganizationsUser.add_user_to_organization` method already requires an instance of both a `User` and an `Organization`, a traditionally object-oriented approach would suggest that we make that function into a method on one of those two objects.

Perhaps we could have have `User`s added directly to `Organization`s:
```ruby
# app/models/organization.rb

def add_user(user)
  OrganizationsUser.add_user_to_organization(user, self)
end
```

What do you think of that approach? Does it actually solve the indirection problem? Is the indirectness a problem at all?

hschallhorn: I like the idea of creating a `add_user` method (mostly because I always assume there is one and have to track down how we normally do it). This this would require some updating of code elsewhere to make sure we're hitting the right method.

app/models/organizations/judge_team.rb
hschallhorn: I think if we go with an `add_user` function, this logic may change (as we'll only be calling `add_user` for attorneys), so I'll not comment til we decide which route to go

app/models/organizations/judge_team.rb
lowellrex: Created a ticket to add `Organization.add_user` instance method (#12496). Happy to delay this ticket until we have implemented that ticket. Or we could continue progress on this ticket and come back and clean this implementation up after we've implemented that ticket. Curious about your thoughts @hschallhorn.

yoomlam: Expanding on the idea of associating two object types...
Currently an organization distinguishes between admin and non-admin members via an admin boolean flag in the `organizations_users` table. I propose having a label that describes the relationship of the member to the organization.

**Problem 1**: Currently, a user can be assigned to both the judge and attorney roles. When adding this user to a JudgeTeam organization, we assume the first user added to the team is always a judge and subsequently added users (if any) are attorneys. Not only does this add domain knowledge to the model, it also restricts the operations that can be done on the organization so that we don't arrive in an unexpected state. Some hypothetical questions: What happens if the judge for the JudgeTeam needs to be replaced? What if we want to form a team of attorneys before a specific judge is known? 

**Problem 2**: In some cases we want to perform certain actions (i.e., hooks) when a certain type of member is added to an organization. Since a user who is an acting judge has both the judge and attorney roles, we cannot determine semantically (i.e., by querying the user) what type of role that user is playing in the organization to which they are being added. 

**Motivation**: Members/Users can have different roles in different organizations. For instance, an acting judge can be the judge (aka admin) in one JudgeTeam and an attorney in another JudgeTeam. 

**Benefits**: Replacing a boolean flag with a string label provides more richness and flexibility. As long as we are careful with what labels are allowed (by using an enumeration), extending our model for future features and tickets will be easier since the label will explicitly specify what role the user is playing for the particular organization: `organization.add_user_as(Roles.Attorney, user)`.

(Background: I've worked with graph databases for at least 6 years, and I always specify the relationship label when linking two objects. The relationship label is not a characteristics of either objects alone, so it should not be an attribute on the objects being linked.)

yoomlam: I wanted to add that this approach allows a user to assume multiple roles in an organization. E.g., A judge can be the [`JudgeTeamLead`](https://github.com/department-of-veterans-affairs/caseflow/blob/5453cdcfd5d9c1bf283fad51dc92160e4cd29de6/app/models/judge_team_lead.rb) as well as the `admin` (so that general calls to get an organization's admin still works).

This idea encourages creating distinct and clearly defined roles. For example, an `admin` manages organization membership, whereas a `JudgeTeamLead` assigns cases to members. So code that manages membership only cares about the `admin` role, and code that deals with assigning cases only cares if the user is a `JudgeTeamLead` -- I think this contextualizes the code and makes it easier to understand. While it may be the case that current judges play both roles, this may change in the future.

lowellrex:
🚫 Blocked by #12496.

app/models/organizations/judge_team.rb
lomky: Just a quick note:

app/models/organizations/judge_team.rb
lomky: I don't think we can assume everyone on a judge team is either a JudgeTeamLead or a DraftingAttorney. It's possible someone will be added to a team as an organization admin to act on behalf of the judge, but this person would not be a drafting attorney.

lowellrex: You're absolutely right. I think the goal for this ticket is just to reach feature parity with a new implementation and then to adapt for new scenarios. As it currently stands, all members of a `JudgeTeam` are assumed to be either judges or attorneys. This PR does not attempt to address other scenarios.

yoomlam:
> 🚫 Blocked by #12496.

That issue is now closed.

app/models/organizations/judge_team.rb
```diff
@@ -9,8 +9,19 @@ def self.create_for_judge(user)
     fail(Caseflow::Error::DuplicateJudgeTeam, user_id: user.id) if JudgeTeam.for_judge(user)
 
     create!(name: user.css_id, url: user.css_id.downcase).tap do |org|
-      org_user = OrganizationsUser.make_user_admin(user, org)
-      JudgeTeamLead.create!(organizations_user: org_user)
+      # Add the JudgeTeamLead record in org.user_added_to_organization which gets triggered by
+      # OrganizationsUser.inform_organization_user_added.
```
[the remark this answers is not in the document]
yoomlam: Good catch! Done.

app/models/organizations/judge_team.rb
```diff
@@ -9,8 +9,19 @@ def self.create_for_judge(user)
...
+      OrganizationsUser.make_user_admin(user, org)
+    end
+  end
+
+  # Use the size of the organization to determine if we have just created a JudgeTeam for a judge,
+  # or just added a user to an existing JudgeTeam. We assume the first user will always be a judge. All subsequent
+  # members of the team will be attorneys.
+  def add_user(user)
+    super.tap do |org_user|
+      class_name = (users.count == 1) ? JudgeTeamLead : DecisionDraftingAttorney
```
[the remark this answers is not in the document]
yoomlam: I remember discussing this in a call: i.e., that we couldn't disable the Remove Admin button in the frontend b/c the frontend doesn't know the org is a JudgeTeam and that it would be easier to implement in the backend. I don't recall any action items resulting from that conversation.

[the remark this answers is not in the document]
lowellrex: We have a ticket to prevent judges from being removed from their `JudgeTeam`s: https://github.com/department-of-veterans-affairs/caseflow/issues/12492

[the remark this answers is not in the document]
hschallhorn: Knew I wasn't going crazy. Batteam is loosening a few too many screws

PR #12495, opened 2019-10-28
Add meat to OrganizationsController.add_user_to_vso code comment

lowellrex:


yoomlam:
Looks good as long as it is safe to assume that the user is a VSO staff and not someone else.
I see there is a `verify_organization_access` and `verify_business_line` check, so we are also assuming that if the user has access to the organization, it is acceptable to add them to the organization.

PR #12508, opened 2019-10-29
Look up Legacy Appeal veteran file number from BGS if looks like SSN

pkarman:
See https://dsva.slack.com/archives/CHX8FMP28/p1572035200274400

### Description
in Legacy Appeal, when the related Veteran model has a SSN value in the `file_number` column, look up the actual file number from BGS.


app/models/veteran.rb
```diff
@@ -254,7 +254,7 @@ def find_by_ssn(ssn, sync_name: false)
       end
       return found_locally if found_locally
 
-      file_number = BGSService.new.fetch_file_number_by_ssn(ssn)
+      file_number = bgs.fetch_file_number_by_ssn(ssn)
```
pkarman: small unrelated optimization, re-using existing BGS object.

Issue #12370, opened 2019-10-11
MTV - Save partial grant to database with selected issues


leikkisa:
### User or job story
This is a backend ticket. This story covers what happens when a Judge "Submits" a partially granted motion to vacate with issues selected.

### Acceptance criteria
- The post decision motion disposition is partial_grant
- When the user submits the Address Motion to Vacate task, the IDs of the selected issues are saved in the post decision motions table
- Document new db column

### Technical notes
- Need to create a new db column

### Out of scope
- Additional processing on the partially granted issues (e.g. creating the new request issues)

### Release notes
See ticket https://github.com/department-of-veterans-affairs/caseflow/issues/11669

#### Designs
Not applicable (backend only).

#### Background/context
Not applicable.

jcq:
The AC on here specifies that the new disposition be `partial_grant`, but all the others are past tense (`granted`, `denied`, etc). I had previously implemented it as just `partial`, which doesn't actually fit that pattern either. Should we match the AC? Should we match the pattern? Is `partially_granted` too long?

leikkisa:
To the extent that I can make it up, I would say partially_granted is fine. We should check with @jimruggiero too, in case there is other disposition-related knowledge I'm not thinking of.

jimruggiero:
partially_granted works

PR #12513, opened 2019-10-30
MTV | Save partial grant to database with selected issues

jcq:
Connects #12370

### Description
Updated DB schema to add `vacated_issue_ids` to `post_decisions_motions` table, and added support for saving the selected issue IDs when submitting from a judge "address motion to vacate" view.

### Acceptance Criteria
- [ ] Code compiles correctly
- [ ] Tests updated to validate new functionality

*Schema Changes Only*

* [x] Column comments updated
* [ ] ~~Query profiling performed (eyeball Rails log, check bullet and fasterer output)~~

### DIFF
--- a/app/controllers/post_decision_motions_controller.rb
+++ b/app/controllers/post_decision_motions_controller.rb
@@ -28,6 +28,6 @@ class PostDecisionMotionsController < ApplicationController
   end
 
   def motion_params
-    params.permit(:disposition, :task_id, :vacate_type, :instructions, :assigned_to_id)
+    params.permit(:disposition, :task_id, :vacate_type, :instructions, :assigned_to_id, vacated_issue_ids: [])
   end
 end
--- a/app/controllers/post_decision_motions_controller.rb
+++ b/app/controllers/post_decision_motions_controller.rb
@@ -28,6 +28,6 @@ class PostDecisionMotionsController < ApplicationController
   end
 
   def motion_params
-    params.permit(:disposition, :task_id, :vacate_type, :instructions, :assigned_to_id, vacated_issue_ids: [])
+    params.permit(:disposition, :task_id, :vacate_type, :instructions, :assigned_to_id, vacated_decision_issue_ids: [])
   end
 end
--- a/app/models/post_decision_motion.rb
+++ b/app/models/post_decision_motion.rb
@@ -3,8 +3,11 @@
 class PostDecisionMotion < ApplicationRecord
   belongs_to :task, optional: false
 
+  # has_many :decision_issues as: :vacated_issues
+
   validates :disposition, presence: true
   validate :vacate_type_is_present_if_granted
+  validate :vacated_issues_present_if_partial
 
   enum disposition: {
     granted: "granted",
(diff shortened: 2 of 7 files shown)


app/models/post_decision_motion.rb
```diff
@@ -19,11 +22,23 @@ class PostDecisionMotion < ApplicationRecord
...
 
+  def vacated_issues
+    return [] unless vacated_issue_ids
+
+    DecisionIssue.find(vacated_issue_ids)
+  end
+
   private
 
   def vacate_type_is_present_if_granted
```
leikkisa: I think we'll need to include the vacate type if it's partially_granted too

app/workflows/post_decision_motion_updater.rb
```diff
@@ -24,13 +24,19 @@ def process
   def create_motion
     motion = PostDecisionMotion.new(
       task: task,
-      disposition: params[:disposition],
+      disposition: disposition,
       vacate_type: params[:vacate_type]
     )
+
+    if params.key?(:vacated_issue_ids)
+      motion.vacated_issue_ids = params[:vacated_issue_ids].map(&:to_i)
```
leikkisa: I think ruby will do the "to_i" magic when saving to the DB

app/workflows/post_decision_motion_updater.rb
```diff
@@ -71,7 +77,12 @@ def create_abstract_task
   end
 
   def disposition
-    params[:disposition]
+    case params[:disposition]
```
leikkisa: We could just call this partially_granted in the first place on the front end to make this simpler.

app/controllers/post_decision_motions_controller.rb
```diff
@@ -28,6 +28,6 @@ def task
   end
 
   def motion_params
-    params.permit(:disposition, :task_id, :vacate_type, :instructions, :assigned_to_id)
+    params.permit(:disposition, :task_id, :vacate_type, :instructions, :assigned_to_id, vacated_issue_ids: [])
```
leikkisa: If it's not a hassle, on second thought calling this vacated_decision_issue_ids would be more descriptive.  It's possible to surmise this from the vacated_issues method too, but I see below they got confused so it might be better to be explicit here.

db/migrate/20191028200741_add_issues_to_post_decision_motions.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddIssuesToPostDecisionMotions < ActiveRecord::Migration[5.1]
+  def change
+    add_column :post_decision_motions, :vacated_issue_ids, :integer, array: true, comment: "An array of the request issue IDs that were chosen for partial vactur in this post-decision motion."
```
leikkisa: This should be decision issue IDs.  

I think this description sounds a little like issues chosen were partially vacated (instead of the appeal decision being partially vacated), which doesn't quite make sense, but that's how the English reads to me.

Not sure how much better this is, but maybe something like:  "When a motion to vacate is partially granted, this includes an array of the appeal's decision issue IDs that were chosen for vacatur in this post-decision motion".

app/models/post_decision_motion.rb
jcq: Good catch. Will make that an OR.

app/workflows/post_decision_motion_updater.rb
jcq: True. I did that because the front end and a bunch of other things were coded with just `partial`. I can switch if you think it's better...

app/workflows/post_decision_motion_updater.rb
jcq: Ah, k. I had done this in an attempt to get it to actually save, but the problem was elsewhere (I didn't have my syntax correct for permitting array params in the controller).

db/migrate/20191028200741_add_issues_to_post_decision_motions.rb
jcq: Works for me! And good catch. I'd prefer my comments didn't sow confusion. :\

app/controllers/post_decision_motions_controller.rb
jcq: I switched things to `vacated_decision_issue_ids`

PR #12527, opened 2019-10-31
Preserve line breaks added to task instructions

lowellrex:
Before | After
--- | ---
image | image

I noticed [the `nl2br` utility function](https://github.com/department-of-veterans-affairs/caseflow/blob/edc8f49bfdd1a00219c9a07f564d37b3980e91c3/client/app/util/StringUtil.js#L85) while motoring around the code yesterday so I thought it'd be nice to use it to preserve new lines that VA employees put into the task instructions field.

### DIFF
--- a/client/app/queue/components/TaskRows.jsx
+++ b/client/app/queue/components/TaskRows.jsx
@@ -1,6 +1,7 @@
 import { css } from 'glamor';
 import React from 'react';
 import moment from 'moment';
+import PropTypes from 'prop-types';
 import Button from '../../components/Button';
 import COPY from '../../../COPY.json';
 import { GrayDot, GreenCheckmark, CancelIcon } from '../../components/RenderFunctions';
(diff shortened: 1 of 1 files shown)


client/app/queue/components/TaskRows.jsx
```diff
@@ -186,7 +187,7 @@ class TaskRows extends React.PureComponent {
 
     return <React.Fragment key={`${task.uniqueId} fragment`}>
       {task.instructions.map((text) => <React.Fragment key={`${task.uniqueId} span`}>
-        <span key={`${task.uniqueId} instructions`}>{text}</span><br /></React.Fragment>)}
+        <span key={`${task.uniqueId} instructions`}>{StringUtil.nl2br(text)}</span><br /></React.Fragment>)}
```
lowellrex: This is the only meaningful change in this PR. The rest is fixing linting complaints.

lowellrex:
cc @jimruggiero that the display of these task instructions will be improved tomorrow!

Issue #12378, opened 2019-10-14
Add decision date to unidentified issue


leikkisa:
### User or job story
As a user intaking or editing a decision review, I can add a decision date to unidentified issues.

### Acceptance criteria
- This work is behind a feature toggle - unidentified_issue_decision_date
- On the Unidentified issues modal the user can enter a decision date
- The decision date is displayed with the unidentified issue after the issue is added
- The decision date is saved in the request issue's "decision_date" field
- The decision date field is optional for unidentified issues
- The request_issue table decision_date db column comment is updated to reflect that it includes the decision date for unidentified issues and new nonrating issues
- The decision date uses the date picker
- Please include a screenshot of the unidentified issue modal with the decision date field

### Out of scope
- timeliness eligibility check on unidentified issues

### Release notes
Users can optionally enter a decision date when they are adding an unidentified issue to a decision review.

#### Background/context
When a user cannot find a rating issue matching the veteran's request, they intake an unidentified issue.  Currently, we do not collect the decision date for these issues.  Intaking the decision date will better enable us to determine which rating issue it should be matched to and the reason that it is missing.

Some veterans do not include decision dates on their form, so this field should be optional.

PR #12528, opened 2019-10-31
Sandra/12378 decision date undentified issue

Sjones352:
Connects  #12378
Screen Shot 2019-10-31 at 11 27 01 AM
**Add Issue Page**
Screen Shot 2019-10-31 at 11 27 18 AM
**Edit page**
Screen Shot 2019-10-31 at 2 33 25 PM

### DIFF
--- a/app/models/request_issue.rb
+++ b/app/models/request_issue.rb
@@ -325,19 +325,18 @@ class RequestIssue < ApplicationRecord
       contested_issue_description: contested_issue_description,
       end_product_cleared: end_product_establishment&.status_cleared?,
       end_product_code: end_product_code,
-      editable: editable?,
-      decisionDate: decision_date
+      editable: editable?
     }
   end
 
   def approx_decision_date_of_issue_being_contested
-    return if is_unidentified
-
     if contested_issue
       contested_issue.approx_decision_date
     elsif decision_date
       decision_date
     else
+      return if is_unidentified
+
       # in theory we should never get here
       fail MissingDecisionDate, id
     end
--- a/app/models/request_issue.rb
+++ b/app/models/request_issue.rb
@@ -325,7 +325,8 @@ class RequestIssue < ApplicationRecord
       contested_issue_description: contested_issue_description,
       end_product_cleared: end_product_establishment&.status_cleared?,
       end_product_code: end_product_code,
-      editable: editable?
+      editable: editable?,
+      decisionDate: decision_date
     }
   end
 
--- a/app/models/request_issue.rb
+++ b/app/models/request_issue.rb
@@ -578,6 +578,10 @@ class RequestIssue < ApplicationRecord
     contested_decision_issue&.remanded?
   end
 
+  def title_of_active_review
+    duplicate_of_issue_in_active_review? ? ineligible_due_to.review_title : nil
+  end
+
   private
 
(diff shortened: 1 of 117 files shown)


client/app/intake/components/UnidentifiedIssuesModal.jsx
```diff
@@ -16,12 +18,13 @@ class UnidentifiedIssuesModal extends React.Component {
   }
 
   onAddIssue = () => {
-    // this.props.addUnidentifiedIssue(this.state.description, this.state.notes, correctionType);
-    const { description, notes } = this.state;
+    // this.props.addUnidentifiedIssue(this.state.description, this.state.notes, decisionDate, correctionType);
```
leikkisa: This comment is leftover from the modal refactor, let's take it out.

client/app/intake/util/issues.js
```diff
@@ -98,7 +98,7 @@ export const formatRequestIssues = (requestIssues, contestableIssues) => {
       benefitType: issue.benefit_type,
       decisionIssueId: issue.contested_decision_issue_id,
       description: issue.description,
-      decisionDate: issue.approx_decision_date,
+      decisionDate: issue.is_unidentified ? issue.decisionDate : issue.approx_decision_date,
```
leikkisa: Let's leave this at what it used to be. This is pulling data from the server, so I don't think there will be the camelCase decisionDate field.

db/migrate/20191030222631_update_decision_date_comment_on_request_issues.rb
```diff
@@ -0,0 +1,5 @@
+class UpdateDecisionDateCommentOnRequestIssues < ActiveRecord::Migration[5.1]
+  def change
+  	change_column_comment(:request_issues, :decision_date, "Either the rating issue's promulgation date, the decision issue's approx decision date or unidentified decision date entered by the user")
```
leikkisa: I was looking at the schema and noticed this is used for nonrating issues too, maybe we can update this to:
`Either the rating issue's promulgation date, the decision issue's approx decision date or the decision date entered by the user (for nonrating and unidentified issues)`

Updating it will require rolling back and re-migrating

app/models/request_issue.rb
```diff
@@ -325,7 +325,8 @@ def ui_hash
       contested_issue_description: contested_issue_description,
       end_product_cleared: end_product_establishment&.status_cleared?,
       end_product_code: end_product_code,
-      editable: editable?
+      editable: editable?,
+      decisionDate: decision_date
```
leikkisa: It's probably better to take this out and use approx_decision_date, but that means the method below needs to get refactored a bit:
```
  def approx_decision_date_of_issue_being_contested
    if contested_issue
      contested_issue.approx_decision_date
    elsif decision_date
      decision_date
    else
      return if is_unidentified
      # in theory we should never get here
      fail MissingDecisionDate, id
    end
  end
```

app/models/request_issue.rb
```diff
@@ -330,13 +330,13 @@ def ui_hash
...
 
   def approx_decision_date_of_issue_being_contested
-    return if is_unidentified
-
     if contested_issue
       contested_issue.approx_decision_date
     elsif decision_date
       decision_date
     else
+      return if is_unidentified
```
leikkisa: 👍

PR #12529, opened 2019-10-31
Created LegacyAppeal `eligible_for_death_dismissal`

lomky:
Starts #12211 

### Description
LegacyAppeals now can determine if a given Appeal & User are eligible for death dismissal

### Acceptance Criteria
- [ ] Tests pass

A part of a larger set of PRs.

### DIFF
--- a/app/models/appeal.rb
+++ b/app/models/appeal.rb
@@ -678,7 +678,7 @@ class Appeal < DecisionReview
     AppealsWithNoTasksOrAllTasksOnHoldQuery.new.ama_appeal_stuck?(self)
   end
 
-  def eligible_for_death_dismissal?(user)
+  def eligible_for_death_dismissal?(_user)
     # Death dismissal processing is only for VACOLs/Legacy appeals
     false
   end
--- a/app/models/appeal.rb
+++ b/app/models/appeal.rb
@@ -678,6 +678,11 @@ class Appeal < DecisionReview
     AppealsWithNoTasksOrAllTasksOnHoldQuery.new.ama_appeal_stuck?(self)
   end
 
+  def eligible_for_death_dismissal?(user)
+    # Death dismissal processing is only for VACOLs/Legacy appeals
+    false
+  end
+
   private
 
   def most_recently_assigned_to_label(tasks)
--- a/app/models/legacy_appeal.rb
+++ b/app/models/legacy_appeal.rb
@@ -767,9 +767,8 @@ class LegacyAppeal < ApplicationRecord
   def eligible_for_death_dismissal?(user)
     return false if notice_of_death_date.nil?
 
-    user_has_actionable_open_tasks = tasks.open.where(type: ColocatedTask.subclasses.map(&:name)).any?
-
-    user_has_actionable_open_tasks && Colocated.singleton.user_is_admin?(user)
+    user_has_relevent_open_tasks = tasks.open.where(type: ColocatedTask.subclasses.map(&:name)).any?
+    user_has_relevent_open_tasks && Colocated.singleton.user_is_admin?(user)
   end
 
   private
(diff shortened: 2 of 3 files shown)


app/models/legacy_appeal.rb
```diff
@@ -764,6 +764,14 @@ def vacols_case_review
     VACOLS::CaseAssignment.latest_task_for_appeal(vacols_id)
   end
 
+  def eligible_for_death_dismissal?(user)
+    return false if notice_of_death_date.nil?
```
yoomlam: `if not notice_of_death_date` is easier to understand for me.

app/models/legacy_appeal.rb
```diff
@@ -764,6 +764,14 @@ def vacols_case_review
...
+
+    user_has_actionable_open_tasks = tasks.open.where(type: ColocatedTask.subclasses.map(&:name)).any?
```
yoomlam: I would have to presume that any subclass of `ColocatedTask` is an _actionable task_. If not, rename variable.

app/models/legacy_appeal.rb
```diff
@@ -764,6 +764,14 @@ def vacols_case_review
...
+
```
yoomlam: Delete blank line to keep semantically-related lines together.

spec/models/legacy_appeal_spec.rb
```diff
@@ -2393,6 +2393,59 @@
     end
   end
 
+  context "#eligible_for_death_dismissal?" do
+    let(:correspondent) { create(:correspondent, sfnod: 4.days.ago) }
+    let(:vacols_case) { create(:case, correspondent: correspondent) }
+    let(:appeal) { create(:legacy_appeal, vacols_case: vacols_case) }
+    let(:colocated_user) { create(:user) }
+    let(:attorney) { create(:user) }
```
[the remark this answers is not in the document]
lomky: No, I'm dropping that ability, as it wasn't requested and was adding complexity to the logic.

app/models/legacy_appeal.rb
lomky: I think this is the more 'ruby' way to write it, but I'm interested in a tie-breaker opinion!

app/models/legacy_appeal.rb
```diff
@@ -764,6 +764,13 @@ def vacols_case_review
     VACOLS::CaseAssignment.latest_task_for_appeal(vacols_id)
   end
 
+  def eligible_for_death_dismissal?(user)
```
lowellrex: Will AMA `Appeal`s ever be eligible for death dismissal? If so, can we add a method that articulates those conditions? If not, can we add a method named similarly to this one that always returns `false` so that we maintain the same interface for legacy and AMA appeals?

app/models/legacy_appeal.rb
lowellrex: I don't have a strong preference here, but the Caseflow practice to date is to prefer `... if some_var.nil?` over `... unless some_var`.

Another option is to include `notice_of_death_date` in the final expression alongside the two other conditions.

app/models/legacy_appeal.rb
lomky: Not to my knowledge, and if so, it will look nothing like this flow.

app/models/appeal.rb
```diff
@@ -678,6 +678,11 @@ def stuck?
     AppealsWithNoTasksOrAllTasksOnHoldQuery.new.ama_appeal_stuck?(self)
   end
 
+  def eligible_for_death_dismissal?(user)
```
lowellrex: I think we usually prefix unused function arguments with an underscore.

lomky:
```suggestion
  def eligible_for_death_dismissal?(_user)
```

app/models/legacy_appeal.rb
```diff
@@ -764,6 +764,13 @@ def vacols_case_review
...
+    return false if notice_of_death_date.nil?
+
+    user_has_relevent_open_tasks = tasks.open.where(type: ColocatedTask.subclasses.map(&:name)).any?
```
lowellrex: This could also be written `ColocatedTask.open.where(appeal: self).any?`. Not sure if that's any clearer though.

PR #12530, opened 2019-10-31
Cleanup and comments from Death Dismissal

lomky:
Unrelated cleanup and comments encountered while working on Death Dismissal

### DIFF
--- a/Makefile.example
+++ b/Makefile.example
@@ -61,7 +61,7 @@ clean:  ## Clean logs, compiled webpack assets, and temporary files
 	rm -rf public/assets
 	rm -rf tmp/capybara
 
-realclean: clean
+realclean: clean ## TODO
 	rm -rf client/node_modules
 	rm -f client/package-lock.json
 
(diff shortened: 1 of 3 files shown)


Makefile.example
```diff
@@ -61,7 +61,7 @@ clean:  ## Clean logs, compiled webpack assets, and temporary files
 	rm -rf public/assets
 	rm -rf tmp/capybara
 
-realclean: clean
+realclean: clean ## TODO
```
hschallhorn: Que es?

app/models/legacy_tasks/judge_legacy_decision_review_task.rb
```diff
@@ -10,6 +10,8 @@ def review_action
   end
 
   def available_actions(current_user, _role)
+    # This must check judge_in_vacols? rather than role as judge, or acting
```
yoomlam: Replace "or" with "otherwise"?

Makefile.example
lomky: I dunno! but if you have no `## comment`, it won't list at all in the help section when you run `make`, but this at least shows the option exists. I believe Lighthouse team made it?  https://dsva.slack.com/archives/C2ZAMLK88/p1571377264125700?thread_ts=1571377264.125700

app/models/legacy_tasks/judge_legacy_decision_review_task.rb
lomky:
```suggestion
    # This must check judge_in_vacols? rather than role as judge, otherwise acting
```

PR #12532, opened 2019-10-31
Persist Fake Ratings storage

pkarman:
Similar to #12024 store fakes BGS Ratings in Redis rather than as class methods.

This allows us to persist across local app restarts, and allows us to inspect the same data via the console as is available to the browser.

### DIFF
--- a/.reek.yml
+++ b/.reek.yml
@@ -14,6 +14,7 @@ detectors:
       - Veteran
   ControlParameter:
     exclude:
+      - Fakes::BGSService
       - HearingRepository#slot_new_hearing
       - Task#actions_allowable?
       - TaskSorter#sort_requires_case_norm?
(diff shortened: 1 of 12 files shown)


.reek.yml
```diff
@@ -14,13 +14,15 @@ detectors:
       - Veteran
   ControlParameter:
     exclude:
+      - Fakes::BGSService
```
pkarman: these are all to quiet existing warnings, not the new code

lib/fakes/persistent_store.rb
```diff
@@ -40,10 +40,50 @@ def fetch_and_inflate(key)
...
+      nil
+    when ActiveSupport::JSON::DATE_REGEX
+      begin
+        Date.parse(data)
+      rescue ArgumentError
+        data
+      end
+    when ActiveSupport::JSON::DATETIME_REGEX
+      begin
+        Time.zone.parse(data).utc.to_datetime
```
pkarman: the only change from original

PR #12541, opened 2019-11-01
If a veteran has a duplicate record, show an error alert when a user searches for a veteran in intake

### LINKED ISSUE #12077: Show alert to user if Duplicate Veteran record found
### Story
When a veteran has a duplicate record, with a different participant ID, establishing their claim fails.

Related #11742

### Acceptance criteria
1. When a user searches for a veteran in intake, if that veteran has a duplicate record show the user an error alert
1. The error alert should include both participant IDs for the veteran, instructions to create an incident in YourIt, and to try again after that has been resolved

### Copy (draft)
This Veteran has a duplicate record in CorpDB.  Please submit an incident in YourIT requesting that the duplicate record be removed.  Include the veteran's first initial, last name, the last 4 digits of their file number or SSN, and both of the following participant IDs: [participant_id1, participant_id2].

Note, please make YourIT a link.  Feel free to make improvements to the copy.

### Technical notes
- Check if the SSN matches the BGS record soc_sec field
- If they do not match, then you can use BGSService find_by_ssn, to get the other participant ID with the soc_sec number.  If the participant ID does not match the veteran's participant_id in Caseflow, then show the alert

tomas-nava:
Resolves #12077

### Description
Adds a validation for duplicate veteran records to the intake start validator, and shows an error with instructions to submit a YourIT ticket if a duplicate record is found.

The validation is behind a feature toggle so we can enable it gradually if desired.

Screen Shot 2019-11-01 at 2 34 35 PM

### DIFF
--- a/app/models/validators/intake_start_validator.rb
+++ b/app/models/validators/intake_start_validator.rb
@@ -60,7 +60,7 @@ class IntakeStartValidator
   def duplicate_veteran_records_in_corpdb
     return false unless FeatureToggle.enabled?(:alert_duplicate_veterans, user: RequestStore[:current_user])
 
-    pids = veteran.participant_ids
+    pids = DuplicateVeteranParticipantIDFinder.new(veteran: veteran).call
     if pids.count > 1
       intake.store_error_data(pids: pids)
       return true
--- a/app/models/validators/intake_start_validator.rb
+++ b/app/models/validators/intake_start_validator.rb
@@ -61,7 +61,7 @@ class IntakeStartValidator
     pids = veteran.participant_ids
     if pids.count > 1
       intake.store_error_data(pids: pids)
-      true
+      return true
     end
     false
   end
--- a/app/models/validators/intake_start_validator.rb
+++ b/app/models/validators/intake_start_validator.rb
@@ -58,6 +58,8 @@ class IntakeStartValidator
   end
 
   def duplicate_veteran_records_in_corpdb
+    return false unless FeatureToggle.enabled?(:alert_duplicate_veterans, user: RequestStore[:current_user])
+
     pids = veteran.participant_ids
     if pids.count > 1
       intake.store_error_data(pids: pids)
--- a/app/models/validators/intake_start_validator.rb
+++ b/app/models/validators/intake_start_validator.rb
@@ -52,7 +52,7 @@ class IntakeStartValidator
   end
 
   def set_veteran_accessible_error
-    return unless !veteran.accessible?
+    return if veteran.accessible?
 
     intake.error_code = veteran.multiple_phone_numbers? ? :veteran_has_multiple_phone_numbers : :veteran_not_accessible
   end
--- a/app/models/validators/intake_start_validator.rb
+++ b/app/models/validators/intake_start_validator.rb
@@ -29,6 +29,8 @@ class IntakeStartValidator
       intake.error_code = :veteran_not_modifiable
     elsif veteran.incident_flash?
       intake.error_code = :incident_flash
+    elsif duplicate_veteran_records_in_corpdb
+      intake.error_code = :veteran_has_duplicate_records_in_corpdb
     end
   end
 
@@ -53,6 +55,15 @@ class IntakeStartValidator
     return unless !veteran.accessible?
 
     intake.error_code = veteran.multiple_phone_numbers? ? :veteran_has_multiple_phone_numbers : :veteran_not_accessible
+  end
+
+  def duplicate_veteran_records_in_corpdb
+    pids = veteran.participant_ids
+    if pids.count > 1
+      intake.store_error_data(pids: pids)
+      true
+    end
+    false
   end
 
   def duplicate_intake_in_progress
--- a/app/models/veteran.rb
+++ b/app/models/veteran.rb
@@ -56,6 +56,13 @@ class Veteran < ApplicationRecord
     participant_id: :ptcpnt_id
   }.freeze
 
+  # TODO: cache :soc_sec_number & :ssn_nbr?
+  BGS_SSN_FIELD_NAMES = [
+    :ssn,
+    :soc_sec_number,
+    :ssn_nbr
+  ].freeze
+
   # TODO: get middle initial from BGS
   def name
     FullName.new(first_name, "", last_name)
@@ -201,6 +208,11 @@ class Veteran < ApplicationRecord
 
   def participant_id
     super || ptcpnt_id
+  end
+
+  def participant_ids
+    ssns = ([ssn] + BGS_SSN_FIELD_NAMES.map { |field_name| bgs_record[field_name] }).compact.uniq
+    ([participant_id] + ssns.map { |ssn| BGSService.new.client.people.find_by_ssn(ssn)[:ptcpnt_id] }).compact.uniq
   end
 
   def ssn
--- a/app/models/veteran.rb
+++ b/app/models/veteran.rb
@@ -56,7 +56,6 @@ class Veteran < ApplicationRecord
     participant_id: :ptcpnt_id
   }.freeze
 
-  # TODO: cache :soc_sec_number & :ssn_nbr?
   BGS_SSN_FIELD_NAMES = [
     :ssn,
     :soc_sec_number,
--- a/app/models/veteran.rb
+++ b/app/models/veteran.rb
@@ -56,12 +56,6 @@ class Veteran < ApplicationRecord
     participant_id: :ptcpnt_id
   }.freeze
 
-  BGS_SSN_FIELD_NAMES = [
-    :ssn,
-    :soc_sec_number,
-    :ssn_nbr
-  ].freeze
-
   # TODO: get middle initial from BGS
   def name
     FullName.new(first_name, "", last_name)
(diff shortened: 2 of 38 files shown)


client/COPY.json
```diff
@@ -508,6 +508,7 @@
   "NO_RELATIONSHIPS": "This Veteran currently has no known relationships. If you have access, please add the claimant to the Corporate Database to continue processing this intake. If you do not have access, please ",
   "DUPLICATE_PHONE_NUMBER_TITLE": "Veteran Record Alert",
   "DUPLICATE_PHONE_NUMBER_MESSAGE": "Please contact VLJ Support to request that this veteran's information be updated to remove duplicate phone numbers currently on file to resolve this error.",
+  "DUPLICATE_CORPDB_RECORD_MESSAGE": "This Veteran has a duplicate record in CorpDB. Please submit an incident in <a href=\"https://yourit.va.gov\" target=\"_blank\" rel=\"noopener noreferrer\">YourIT</a> requesting that the duplicate record be removed. Include the veteran's first initial, last name, the last 4 digits of their file number or SSN, and the following participant IDs: [%s].",
```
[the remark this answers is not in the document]
leikkisa: I am reluctant to add this considering how often our "instructions" change - but in a meeting today heard that these should get routed to the PHI SME Compensation Corporate Records Group.

Do you think we should add this to the YourIT instructions?

client/app/components/Alert.jsx
```diff
@@ -20,11 +21,18 @@ export default class Alert extends React.Component {
     return attrs;
   }
 
+  messageDiv() {
+    let messageText = this.props.children || this.props.message;
+
+    return <div
+      className="usa-alert-text"
+      dangerouslySetInnerHTML={{ __html: DOMPurify.sanitize(messageText) }}
```
leikkisa: I think this is breaking other alerts, like the incidentFlash alert which has the body set to a React element

pkarman: I have used the dangerous inner HTML hack in the past if I have raw HTML I want to display. Since then, though, we've added support for Markdown syntax in text display, so that might be an easier way to go. The constant string would be easier to write as well.

Look for `ReactMarkdown` in the `client/app` space for example.

app/models/veteran.rb
```diff
@@ -203,6 +209,11 @@ def participant_id
     super || ptcpnt_id
   end
 
+  def participant_ids
```
pkarman: the naming of this might be a little misleading since we have a `participant_id` attribute that is native to the Veteran, and what we're doing here is looking for duplicates (bad CorpDB data).

Maybe `duplicate_participant_ids` would be clearer that this is a Bad State? Or even a service class, like:

```ruby
class DuplicateVeteranFinder
  def initialize(participant_id:, ssn:)
    @participant_id = participant_id
    @ssn = ssn
  end

  def call
    # the current participant_ids code
  end

  private

  attr_reader :ssn, :participant_id
end
```

tomas-nava: I like this idea, will implement.

spec/feature/intake/intake_spec.rb
```diff
@@ -269,6 +269,29 @@
       expect(page).to have_content("David Schwimmer already started processing this form")
     end
 
+    context "" do
```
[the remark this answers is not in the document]
tomas-nava: fixed with [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/12541/commits/1ab63b93097dca0b155260f62eed5d50984a3480)

client/COPY.json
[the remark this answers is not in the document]
tomas-nava: I will add it, and we can see how it looks.

geronimoramos:
@tomas-nava looks good to me from copy perspective! It might be the screen shot but a little bit more padding between header and body would make it perfect
Screen Shot 2019-11-01 at 1 33 14 PM

client/app/components/Alert.jsx
tomas-nava: Addressed in [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/12541/commits/ae1d2cae73861203f60e4feed64dba90d26607d0).

app/models/veteran.rb
tomas-nava: Addressed in [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/12541/commits/de88a093d39b2333ca39ac4d7117af30fde4b90e).

client/COPY.json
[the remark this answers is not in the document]
tomas-nava: added in [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/12541/commits/5e6807720df25421ffca76021f6bf3e0ee5924d7), and updated the screenshot.

tomas-nava:
@geronimoramos I created [a new issue for the spacing issue in caseflow-commons](https://github.com/department-of-veterans-affairs/caseflow-commons/issues/166).

app/models/validators/intake_start_validator.rb
```diff
@@ -50,11 +52,22 @@ def validate_file_number
   end
 
   def set_veteran_accessible_error
-    return unless !veteran.accessible?
+    return if veteran.accessible?
```
pkarman: good! I find myself cleaning up double negatives too.

app/services/duplicate_veteran_participant_id_finder.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class DuplicateVeteranParticipantIDFinder
+  BGS_SSN_FIELD_NAMES = [
+    :ssn,
+    :soc_sec_number,
+    :ssn_nbr
+  ].freeze
+
+  def initialize(veteran:)
```
pkarman: I like this interface

client/app/intake/pages/search.jsx
```diff
@@ -89,6 +91,17 @@ class Search extends React.PureComponent {
         title: 'The Veteran has multiple active phone numbers',
         body: COPY.DUPLICATE_PHONE_NUMBER_MESSAGE
       },
+      veteran_has_duplicate_records_in_corpdb: {
+        title: 'Duplicate veteran records',
+        body: <React.Fragment key="alert-error-body">
+          {'This Veteran has a duplicate record in CorpDB. Please submit an incident in '}
+          { YourITLink }
```
pkarman: cleaner!

app/models/validators/intake_start_validator.rb
```diff
@@ -50,11 +52,22 @@ def validate_file_number
...
 
     intake.error_code = veteran.multiple_phone_numbers? ? :veteran_has_multiple_phone_numbers : :veteran_not_accessible
   end
 
+  def duplicate_veteran_records_in_corpdb
+    return false unless FeatureToggle.enabled?(:alert_duplicate_veterans, user: RequestStore[:current_user])
+
+    pids = DuplicateVeteranParticipantIDFinder.new(veteran: veteran).call
```
leikkisa: Oh, nice way to separate this out 👍

spec/services/duplicate_veteran_participant_id_finder_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    stub_const("BGSService", ExternalApi::BGSService)
+    RequestStore[:current_user] = create(:user)
+
+    ssn = "987654321"
+    file_number = "111111111S"
+    participant_id = "123456"
+    second_participant_id = "765432"
+    veteran = create(:veteran, ssn: ssn, participant_id: participant_id, file_number: file_number)
+
+    allow_any_instance_of(Veteran).to receive(:fetch_bgs_record).and_return({})
```
pkarman: instead of `allow_any_instance_of` could you mock our particular `veteran`?

```ruby
allow(veteran).to receive(:fetch_bgs_record).and_return({})
```

I'm trying to follow the advice of using `allow_any_instance_of` more sparingly and instead mocking/injecting specific double/instances.

spec/feature/intake/intake_spec.rb
```diff
@@ -269,6 +269,31 @@
...
 
+    context "the alert_duplicate_veterans feature toggle is enabled" do
+      before { FeatureToggle.enable!(:alert_duplicate_veterans, users: [current_user.css_id]) }
+      after { FeatureToggle.disable!(:alert_duplicate_veterans) }
+
+      scenario "Search for a veteran who has duplicate records in CorpDB" do
+        participant_id1 = "123456"
+        participant_id2 = "789012"
+
+        allow_any_instance_of(DuplicateVeteranParticipantIDFinder).to receive(:call).and_return(
```
leikkisa: 👍

spec/services/duplicate_veteran_participant_id_finder_spec.rb
tomas-nava: Addressed in [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/12541/commits/ba43ee012690c220988344029fbbafe345f0007a).

tomas-nava:
> It looks good to me! Would you mind adding the screenshot to the issue, that is a request from Product (I don't think they look at the PRs much). Thank you!

added the screenshot in a comment on the issue: https://github.com/department-of-veterans-affairs/caseflow/issues/12077#issuecomment-549465366

PR #12542, opened 2019-11-01
Create design-story.md

carodew:
### Description
Proposal for an issue template for design, research, or writing stories. 

### Acceptance Criteria
- [ ] Product managers review this and are willing to give it a try.

### Testing Plan
1. Use it. Iterate.

### Documentation Updates
- [ ] Not relevant.


jimruggiero:
@carodew are you thinking this template would only be used if the Design team is originating tickets to research a new problem, address design debt or initiate UX improvements? Curious to hear your thoughts on when this template would be used instead of the existing story template.

carodew:
> @carodew are you thinking this template would only be used if the Design team is originating tickets to research a new problem, address design debt or initiate UX improvements? Curious to hear your thoughts on when this template would be used instead of the existing story template.

@jimruggiero I'm thinking this would be a design 'task' template that is one step smaller than a traditional user story. For example, if the user story is "As a hearings coordinator, I want to be able to schedule Veterans for virtual hearings as part of the pilot" the research task stemming off of that might follow this template, and state something like "Conduct user research with Veterans and representatives involved in the virtual hearings pilot" and a design task might be "design the interface for hearings coordinators to switch hearing type." Does that make sense?

jimruggiero:
@Carolyn Dew <carolyn@navapbc.com> That does make sense. Thank you for the
clarification!

carodew:
Made two changes based on feedback – 

1. moved 'not in scope' section to be right below problem statement
2. renamed the file 'design-task.md' to clarify that's it's a level below user stories.

carodew:
@jimruggiero I made one more change (made out of scope a required field). If this looks good to you can you change your review and possibly hit the green merge button? ty!

jimruggiero:
Looks good!

PR #12549, opened 2019-11-01
Remove actions on on-hold tasks for judges

### LINKED ISSUE #12412: Do not show actions for MTV Judge tasks on_hold
### User or job story
As a Judge user, when my motions to vacate tasks are on hold, I shouldn't be able to perform actions on them.

### Open questions
Was the default "on_hold" behavior overwritten in JudgeTask on purpose? Please check with the person who created "actions_available?" in JudgeTask, or someone who is familiar with the desired behavior for one of the other judge tasks.  Another potential place to look for an answer is the specs.

### Acceptance criteria
- A Judge cannot perform actions on their MTV tasks while they are on hold, unless it is a timed hold (default task behavior)
  - JudgeAddressMotionToVacateTask
  - JudgeSignMotionToVacateTask
- Update top line comments in JudgeTask with the MTV JudgeTasks

### Technical notes
- The default task behavior in "actions_available?" is to not show actions if the task is on_hold (unless it's a timed hold)
- This was overwritten in JudgeTask, which is a parent of the above two tasks
- Depending on the answer to the open question above, if it was on purpose, this can be implemented for the two specific tasks. If it wasn't, then it can be for all Judge tasks (this may fix a bug).  If we don't know, the default path forward is to only apply it to the MTV tasks.

### Release notes
N/A

#### Designs
N/A

#### Background/context
Discussion thread for whether on_hold tasks should have actions:
https://dsva.slack.com/archives/C2ZAMLK88/p1571268452113200

nanotone:
Resolves #12412 

### Description
Removes `JudgeTask`'s `actions_available?` method so that it reverts to the default implementation, which is to return false when the task is on hold. I consulted with @tomas-nava who is more familiar with this area and believes the override was an out-of-date work-around that's no longer needed.

### Acceptance Criteria
- [ ] A Judge cannot perform actions on their MTV tasks while they are on hold, unless it is a timed hold (default task behavior)
- [ ] Update top line comments in JudgeTask with the MTV JudgeTasks

### Testing Plan
- [ ] Manual smoke-test to confirm that judges' on-hold tasks do not allow any actions (the default behavior)

### Documentation Updates
- Updated topline comment in `judge_task.rb` to reflect the new MTV subclasses

### DIFF
--- a/app/models/tasks/judge_task.rb
+++ b/app/models/tasks/judge_task.rb
@@ -2,8 +2,12 @@
 
 ##
 # Parent class for all tasks to be completed by judges, including
-# JudgeQualityReviewTasks, JudgeDecisionReviewTasks,
-# JudgeDispatchReturnTasks, and JudgeAssignTasks.
+# - JudgeQualityReviewTasks
+# - JudgeDecisionReviewTasks
+# - JudgeDispatchReturnTasks
+# - JudgeAssignTasks
+# - JudgeAddressMotionToVacateTasks
+# - JudgeSignMotionToVacateTasks
 
 class JudgeTask < Task
   def available_actions(user)
@@ -13,10 +17,6 @@ class JudgeTask < Task
       Constants.TASK_ACTIONS.REASSIGN_TO_JUDGE.to_h,
       additional_available_actions(user)
     ].flatten
-  end
-
-  def actions_available?(user)
-    assigned_to == user
   end
 
   def additional_available_actions(_user)
--- a/app/models/tasks/judge_task.rb
+++ b/app/models/tasks/judge_task.rb
@@ -19,9 +19,11 @@ class JudgeTask < Task
     ].flatten
   end
 
+  # :nocov:
   def additional_available_actions(_user)
     fail Caseflow::Error::MustImplementInSubclass
   end
+  # :nocov:
 
   def timeline_title
     COPY::CASE_TIMELINE_JUDGE_TASK
--- a/spec/models/tasks/judge_task_spec.rb
+++ b/spec/models/tasks/judge_task_spec.rb
@@ -22,9 +22,17 @@ describe JudgeTask, :all_dbs do
 
     subject { subject_task.available_actions_unwrapper(user) }
 
-    context "the task is not assigned to the current user" do
-      let(:user) { judge2 }
+    context "when the task is on hold" do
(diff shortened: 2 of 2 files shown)


spec/models/tasks/judge_task_spec.rb
```diff
@@ -22,9 +22,17 @@
 
     subject { subject_task.available_actions_unwrapper(user) }
 
-    context "the task is not assigned to the current user" do
-      let(:user) { judge2 }
+    context "when the task is on hold" do
```
leikkisa: It looks like the judge task spec didn't have a specific test for the previous differing behavior in judge task.  Normally I would say to remove that spec (if it existed), since this PR is removing that method.

Since it's now inheriting from task, I think the test for this method is already covered here:
https://github.com/department-of-veterans-affairs/caseflow/blob/8fb70d203fa98ce57b61850e33fbdf7df60a3025/spec/models/task_spec.rb#L566

If you wanted to add an extra layer of diligence, you could add something to the MTV feature test - when a judge's task is on hold, to visit to the case details screen and check the UI if there are available actions.  I don't feel strongly about adding this here since that's the default behavior and this is a bug fix.

app/models/tasks/judge_task.rb
```diff
@@ -15,10 +19,6 @@ def available_actions(user)
     ].flatten
   end
 
-  def actions_available?(user)
```
leikkisa: Based on the git blames, this code for actions available is older than that in task.rb, so I think this was just outdated, and this looks like the right move!

spec/models/tasks/judge_task_spec.rb
nanotone: Yeah, I'm on board with just removing the test and calling it a day. However, it looks like removing it will take `judge_task.rb` test coverage under 90% 🤦‍♂

I might fiddle around with improving coverage elsewhere in that file, but let me know if you have thoughts.

nanotone:
Now that the overriding `actions_available?` method is gone, there's almost no real functionality in JudgeTask (which is a good thing). I'm marking the abstract method `additional_available_actions` with `nocov` to make CircleCI happy.

yoomlam:
5 Why's meeting to dig into this ticket: https://dsva.slack.com/archives/C2ZAMLK88/p1574866744386400

Issue #12284, opened 2019-10-03
Lit Support: Route case assignments for inactive attorney profiles to the general work queue 


jimruggiero:
### User or job story
User story: When a Motions Attorney, who was originally working a given case no longer works at BVA or in Lit Support, I want the the case returned to the team queue, so that it can be re-assigned to a different attorney.

### Acceptance criteria
- This work is behind a feature toggle: "review_motion_to_vacate".
- When motions attorney profiles are marked as inactive in Caseflow, all cases assigned to that attorney are automatically moved to the respective generic team queue(s).
- Cases moved into generic team queues are visible and can by reassigned by team leads.
- Once reassigned, cases are visible by the newly assigned attorney. 
- Newly assigned attorneys can see reassigned cases in their work queues.
- A screenshot or .gif of what is coded and merged to Production is attached to this issue prior to it being closed.

### Release notes
_Write what should be included in release notes (Caseflow uses Headway), updated when the story is built, before it's deployed._

Caseflow now supports the ability for cases assigned to attorneys no longer working at The Board or in Litigation Support, to be routed to a teams' general work queue for reassignment to a new attorney.

#### Designs
https://www.figma.com/file/VtOG0hibbyJ6T5UcYnn5EP/Motions-to-Vacate_Phase-II_Epic-8547?node-id=0%3A1

#### Background/context
This story specifically drives value for BVA team leads. It allows them to quickly reassign cases to prevent unnecessary delays in the process for veterans and ensures productivity metrics are met by their respective teams.

#### Success criteria
See ACs

#### Technical notes
- One approach would be to do this through an after save hook on users.
- Close the task associated with the inactive attorney and open a new task
- User tasks should all have parent organizations tasks

#### Open questions
- Does this need to support more than Lit Support (i.e. when returning tasks to generic queues, would there be different queues for different teams [motions v decisions attorney teams])?
- Can we extend this to judge reassignment flows detailed in other tickets?

#### Other notes
- Newly assigned attorneys can see the prior (inactive) attorney assignment and the new assignment of the case (to them) in Case Timeline - out of scope for this epic

#### Resources/other links
Related PR for reassigning tasks for inactive users: https://github.com/department-of-veterans-affairs/caseflow/pull/12483

jimruggiero:
Please reference work that Echo is doing to hook into inactive user/general queue task assignments (see ticket XXXXX).

jcq:
[...]

PR #12567, opened 2019-11-01
MTV | Add org tasks to MTV task tree to account for potentially inactive users

jcq:
Connects #12284

### Description
We reorganized MTV task tree to add organization tasks (of same type) as parents to the various tasks that are spawned by `PostDecisionMotionUpdater`. This allows the org to handle things in case the assigned attorney goes inactive.

### Acceptance Criteria
- [ ] Code compiles correctly
- [ ] We create org task for `DeniedMotionToVacateTask`
- [ ] We create org task for `DismissedMotionToVacateTask`
- [ ] We create org task for `VacateAndDeNovoTask`
- [ ] We create org task for `StraightVacateAndReadjudicationTask`
- [ ] Org tasks are completed if their child atty task is completed (but reopened if atty task is cancelled)

### Testing Plan
1. Go to ...

### Documentation Updates

### DIFF
--- a/app/models/tasks/decided_motion_to_vacate_task.rb
+++ b/app/models/tasks/decided_motion_to_vacate_task.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+class DecidedMotionToVacateTask < Task
+  before_create :automatically_create_org_task
+
+  def available_actions(user)
+    actions = super(user)
+
+    actions.push(Constants.TASK_ACTIONS.LIT_SUPPORT_PULAC_CERULLO.to_h)
+
+    actions
+  end
+
+  def self.label
+    COPY::DENIED_MOTION_TO_VACATE_TASK_LABEL
+  end
+
+  def update_status_if_children_tasks_are_closed(_child_task)
+    if assigned_to.is_a?(Organization)
+      return update!(status: :completed)
+    end
+
+    super
+  end
+
+  def automatically_create_org_task
+    if assigned_to.is_a?(Organization)
+      return
+    end
+
+    org_task = dup.tap do |new_task|
+      new_task.assigned_to_type = "Organization"
+      new_task.assigned_to = org
+      new_task.save!
+    end
+    self.parent = org_task
+  end
+
+  def org
+    LitigationSupport.singleton
+  end
+end
--- a/app/models/tasks/decided_motion_to_vacate_task.rb
+++ b/app/models/tasks/decided_motion_to_vacate_task.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class DecidedMotionToVacateTask < Task
+  self.abstract_class = true
+
   before_create :automatically_create_org_task
 
   def available_actions(user)
--- a/app/models/tasks/decided_motion_to_vacate_task.rb
+++ b/app/models/tasks/decided_motion_to_vacate_task.rb
@@ -34,8 +34,8 @@ class DecidedMotionToVacateTask < Task
     self.parent = org_task
   end
 
-  def org(_user)
-    self.class.org(_user)
+  def org(user)
+    self.class.org(user)
   end
 
   class << self
--- a/app/models/tasks/decided_motion_to_vacate_task.rb
+++ b/app/models/tasks/decided_motion_to_vacate_task.rb
@@ -35,6 +35,12 @@ class DecidedMotionToVacateTask < Task
   end
 
   def org
-    fail Caseflow::Error::MustImplementInSubclass
+    self.class.org
+  end
+
+  class << self
+    def org
+      fail Caseflow::Error::MustImplementInSubclass
+    end
   end
 end
--- a/app/models/tasks/decided_motion_to_vacate_task.rb
+++ b/app/models/tasks/decided_motion_to_vacate_task.rb
@@ -15,7 +15,7 @@ class DecidedMotionToVacateTask < Task
 
   def update_status_if_children_tasks_are_closed(_child_task)
     if assigned_to.is_a?(Organization)
-      return update!(status: :completed)
+      return update!(status: :completed) unless all_children_cancelled?
     end
 
     super
@@ -28,18 +28,18 @@ class DecidedMotionToVacateTask < Task
 
     org_task = dup.tap do |new_task|
       new_task.assigned_to_type = "Organization"
-      new_task.assigned_to = org
+      new_task.assigned_to = org(assigned_by)
       new_task.save!
     end
     self.parent = org_task
   end
 
-  def org
-    self.class.org
+  def org(_user)
+    self.class.org(_user)
   end
 
   class << self
-    def org
+    def org(_user)
       fail Caseflow::Error::MustImplementInSubclass
     end
   end
--- a/app/models/tasks/decided_motion_to_vacate_task.rb
+++ b/app/models/tasks/decided_motion_to_vacate_task.rb
@@ -13,10 +13,6 @@ class DecidedMotionToVacateTask < Task
     actions
   end
 
-  def self.label
-    COPY::DENIED_MOTION_TO_VACATE_TASK_LABEL
-  end
-
   def update_status_if_children_tasks_are_closed(_child_task)
     if assigned_to.is_a?(Organization)
       return update!(status: :completed)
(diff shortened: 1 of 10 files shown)


jcq:
Initial implementation added logic to `PostDecisionMotionUpdater`, but perhaps it may be wise to rewrite to implement the org-task-creation logic into a new abstract task class from which the various ones inherit.

spec/models/decided_motion_to_vacate_task_spec.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require "support/vacols_database_cleaner"
+require "rails_helper"
+
+RSpec.describe DecidedMotionToVacateTask, :all_dbs, type: :model do
```
leikkisa: This shouldn't be using VACOLS, so I think it's okay to do the postgres version of DB cleanup here. The reason this got added in the first place is because the vacols database cleanup was really slow.

spec/workflows/post_decision_motion_updater_spec.rb
```diff
@@ -38,31 +42,82 @@
...
+
           attorney_task = StraightVacateAndReadjudicationTask.find_by(assigned_to_id: assigned_to_id)
           expect(attorney_task).to_not be nil
-          expect(attorney_task.parent).to eq abstract_task
+          expect(attorney_task.parent).to eq org_task
           expect(attorney_task.assigned_by).to eq task.assigned_to
           expect(attorney_task.status).to eq Constants.TASK_STATUSES.assigned
         end
+
+        it "should create JudgeSignMotionToVacateTask" do
```
leikkisa: This seems redundant with the above test

spec/workflows/post_decision_motion_updater_spec.rb
```diff
@@ -38,31 +42,82 @@
       context "when vacate type is straight vacate and readjudication" do
         let(:vacate_type) { "straight_vacate_and_readjudication" }
 
-        it "should create straight vacate and readjudication attorney task" do
+        it "should create straight vacate and readjudication attorney task with correct structure" do
           subject.process
           expect(task.reload.status).to eq Constants.TASK_STATUSES.completed
           abstract_task = AbstractMotionToVacateTask.find_by(parent: task.parent)
```
leikkisa: With respect to this comment from https://github.com/department-of-veterans-affairs/caseflow/issues/12284#issuecomment-548153118

> Suggested we replace AbstractMotionToVacateTask with JudgeSignMotionToVacateTask in the grant/partial cases, as that follows how other judge tasks are handled. 

It looks like we're keeping AbstractMotionToVacateTask for all of the decided MTVs, but adding JudgeSignMotionToVacate in between that and the two granted paths.

app/workflows/post_decision_motion_updater.rb
```diff
@@ -91,7 +107,7 @@ def denied_or_dismissed?
   end
 
   def assigned_to
-    @assigned_to ||= (denied_or_dismissed? ? prev_motions_attorney_or_org : User.find_by(id: params[:assigned_to_id]))
+    @assigned_to ||= (denied_or_dismissed? ? prev_motions_attorney : User.find_by(id: params[:assigned_to_id]))
```
leikkisa: I think you should keep this as `prev_motions_attorney_or_org`.

spec/models/decided_motion_to_vacate_task_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  describe ".available_actions" do
+    subject { task.available_actions(attorney) }
+
+    let(:attorney) { create(:user) }
+    let(:task) { create(:denied_motion_to_vacate_task) }
+
+    it "should include Pulac Cerullo action in available actions" do
+      actions = subject
+      expect(actions).to include(Constants.TASK_ACTIONS.LIT_SUPPORT_PULAC_CERULLO.to_h)
```
leikkisa: You could just put subject here (not important)

app/models/tasks/decided_motion_to_vacate_task.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def available_actions(user)
+    actions = super(user)
+
+    actions.push(Constants.TASK_ACTIONS.LIT_SUPPORT_PULAC_CERULLO.to_h)
+
+    actions
+  end
+
+  def update_status_if_children_tasks_are_closed(_child_task)
```
leikkisa: I don't think we want to complete the org task if, for example, the user task gets canceled when their user becomes inactive.  But it does make sense to complete the org task if the user task is completed.  I believe closed includes both canceled and completed.

spec/workflows/post_decision_motion_updater_spec.rb
```diff
@@ -99,26 +154,45 @@
...
-        expect(attorney_task.parent).to eq abstract_task
+        expect(attorney_task.parent).to eq org_task
         expect(attorney_task.assigned_by).to eq task.assigned_to
         expect(attorney_task.assigned_to).to eq mtv_mail_task.assigned_to
         expect(attorney_task.status).to eq Constants.TASK_STATUSES.assigned
       end
 
-      it "should assign new task to org if prev atty is inactive" do
-        motions_atty.update_status!(Constants.USER_STATUSES.inactive)
+      # it "should only assign new task to org if prev atty is inactive" do
```
leikkisa: Let's keep this functionality around

spec/models/decided_motion_to_vacate_task_spec.rb
jcq: Ahh...  I had been getting errors from dupes before w/o the `:all_dbs` symbol, but the key was adding the correct `require "support/database_cleaner"` bit at the top.

Thx!

spec/workflows/post_decision_motion_updater_spec.rb
jcq: Yup — it was going to be its own test, but then I decided it would be duplicative and just moved the stuff into the existing test. This is now superfluous.

app/workflows/post_decision_motion_updater.rb
jcq: The `prev_motions_attorney_or_org` bit was in there before we had separate org tasks — we don't want to separately possibly assign the "User" task to the org

app/models/tasks/decided_motion_to_vacate_task.rb
jcq: Very good point — my brain read that as completed, but that would also cover cancelled as well, which is not the logic we want. I'm updating.

spec/workflows/post_decision_motion_updater_spec.rb
jcq: Re-added, though adjusted it slightly

spec/workflows/post_decision_motion_updater_spec.rb
```diff
@@ -99,26 +147,47 @@
...
-        expect(attorney_task.parent).to eq abstract_task
-        expect(attorney_task.assigned_by).to eq task.assigned_to
-        expect(attorney_task.assigned_to).to eq LitigationSupport.singleton
-        expect(attorney_task.status).to eq Constants.TASK_STATUSES.assigned
+
+        org_task = DeniedMotionToVacateTask.find_by(assigned_to_id: lit_support_team)
+        expect(org_task).to_not be nil
+        expect(org_task.parent).to eq abstract_task
+
+        # The following check should be added back in once logic is in place to prevent assignment to inactive users
```
leikkisa: I think you can take these comments out since we're not going to incorporate that right now

spec/workflows/post_decision_motion_updater_spec.rb
```diff
@@ -129,26 +198,47 @@
...
-        expect(attorney_task.parent).to eq abstract_task
-        expect(attorney_task.assigned_by).to eq task.assigned_to
-        expect(attorney_task.assigned_to).to eq LitigationSupport.singleton
-        expect(attorney_task.status).to eq Constants.TASK_STATUSES.assigned
+
+        org_task = DismissedMotionToVacateTask.find_by(assigned_to_id: lit_support_team)
+        expect(org_task).to_not be nil
+        expect(org_task.parent).to eq abstract_task
+
+        # The following check should be added back in once logic is in place to prevent assignment to inactive users
```
leikkisa: Same here

PR #12576, opened 2019-11-04
Do not reopen RootTasks when child created

lowellrex:
Resolves #12574.

PR #12578, opened 2019-11-04
Send message to Sentry when child task created for closed task

lowellrex:
Adds an alert when the application creates a child task of a closed task as suggested by this comment on a recent PR: https://github.com/department-of-veterans-affairs/caseflow/pull/12576#pullrequestreview-311225292

### DIFF
--- a/spec/models/task_spec.rb
+++ b/spec/models/task_spec.rb
@@ -976,7 +976,7 @@ describe Task, :all_dbs do
       allow(Raven).to receive(:capture_message)
     end
 
-    context "when the RootTask is active" do
+    context "when the task is active" do
       it "does not send a message to Sentry" do
         expect(parent_task.status).to eq(Constants.TASK_STATUSES.assigned)
         expect(parent_task.children.count).to eq(0)
@@ -989,7 +989,7 @@ describe Task, :all_dbs do
       end
     end
 
-    context "when the RootTask is closed" do
+    context "when the task is closed" do
       before { parent_task.update!(status: Constants.TASK_STATUSES.completed) }
 
       it "sends a message to Sentry" do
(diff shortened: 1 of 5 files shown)


spec/models/task_spec.rb
```diff
@@ -966,4 +966,36 @@ def next_assignee(_options = {})
...
+
+        expect(parent_task.status).to eq(Constants.TASK_STATUSES.on_hold)
+        expect(parent_task.children.count).to eq(1)
+      end
+    end
+
+    context "when the RootTask is closed" do
+      before { parent_task.update!(status: Constants.TASK_STATUSES.completed) }
+
+      it "does sends a message to Sentry" do
```
hschallhorn: "does sends" -> "does send" or "sends"

spec/models/task_spec.rb
```diff
@@ -966,4 +966,36 @@ def next_assignee(_options = {})
...
     end
   end
+
+  describe ".when_child_task_created" do
+    let(:parent_task) { create(:task, appeal: create(:appeal)) }
+
+    subject { create(:task, parent: parent_task, appeal: parent_task.appeal) }
+
+    context "when the RootTask is active" do
+      it "does not send a message to Sentry" do
```
hschallhorn: We might want to write a test to confirm that

spec/models/task_spec.rb
hschallhorn: Also may want a test that the message gets sent to sentry here as well

hschallhorn:
Do we also want to be informed when a child task is created under a root task? As of #12576, we still create the child, but just don't update the `RootTask`'s status.

spec/models/tasks/root_task_spec.rb
```diff
@@ -129,6 +133,22 @@
         expect(root_task.status).to eq(Constants.TASK_STATUSES.completed)
         expect(root_task.children.count).to eq(1)
       end
+
+      context "when the child task is a generic 'ol task" do
```
hschallhorn: >generic task

:eye_twitch:

PR #12582, opened 2019-11-04
Add upsert to AODMotion

### LINKED ISSUE #12577: Duplicate AOD records
In #11124 we attempted to implement a unique constraint on `advance_on_docket_motions` records.

However, we appear to continue getting "duplicate" records.
https://github.com/department-of-veterans-affairs/dsva-vacols/issues/85

lomaxap:
resolves #12577

### DIFF
--- a/app/models/advance_on_docket_motion.rb
+++ b/app/models/advance_on_docket_motion.rb
@@ -24,10 +24,10 @@ class AdvanceOnDocketMotion < ApplicationRecord
       ).any?
     end
 
-    def upsert(person_id:, **args)
+    def upsert_by_person_id(person_id, attrs)
       motion = find_or_create_by(person_id: person_id)
 
-      motion.update(**args)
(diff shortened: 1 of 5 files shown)


app/models/advance_on_docket_motion.rb
```diff
@@ -23,5 +23,11 @@ def granted_for_person?(person_id, appeal_receipt_date)
         person_id: person_id
       ).any?
     end
+
+    def upsert_by_person_id(person_id, attrs)
```
pkarman: upsert? is this a forward compatible naming for Rails 6?

lomaxap: according to rails doc, upsert

> does not instantiate any models nor does it trigger Active Record callbacks or validations

so this wouldn't be consistent with that behavior. I'll change this to `create_or_update_by_person_id`

PR #12588, opened 2019-11-04
Do not save two claimants during intake, use last claimant if there are two

leikkisa:
### Description
If someone goes back and edits the claimant during intake (on the review request page), a second claimant is added to the decision review.  The second claimant is the correct claimant, before we were referencing the first claimant.

This PR:
- Fixes the bug where multiple claimants are created for a decision review
- Changes the code to use the last claimant instead of the first claimant. This is to be safe for the small number of decision reviews that were already created with two claimants.

### Testing Plan
1. Start a decision review
2. Add a claimant, proceed to add issues page
3. Go back, change the claimant, proceed to add issues page
4. Check that the second claimant is the reviews claimant, and that there are no other claimants

Also:
If there are multiple claimants, check that the last claimant is used for establishing a claim and creating a decision document.


app/controllers/intakes_controller.rb
```diff
@@ -141,16 +141,19 @@ def intake
...
+
   def veteran_file_number
     # param could be file number or SSN. Make sure we return file number.
     veteran = Veteran.find_by_file_number_or_ssn(params[:file_number], sync_name: true)
     veteran ? veteran.file_number : params[:file_number]
   end
 
   def success_message
-    detail = intake.detail
     claimant_name = detail.veteran_full_name
```
leikkisa: `claimant_name = detail.veteran_full_name` is really only needed for RAMP intakes, for AMA reviews, claimant.try(:name) is equivalent to to veteran_full_name when the veteran is the claimant

app/models/claimant.rb
```diff
@@ -13,7 +13,7 @@ class Claimant < ApplicationRecord
             uniqueness: { scope: [:decision_review_id, :decision_review_type],
                           on: :create }
 
-  def self.create_from_intake_data!(participant_id:, payee_code:)
```
leikkisa: This was quite the misnomer

spec/controllers/idt/api/appeals_controller_spec.rb
```diff
@@ -36,7 +36,7 @@
 
         let!(:ama_appeals) do
           [
-            create(:appeal, veteran: veteran1, number_of_claimants: 2),
```
leikkisa: I was torn about leaving these tests in (if in the future we really will support multiple claimants per decision review). I decided to take it out for now since we don't and it's confusing. The table (engineered for the future) has already been a source of confusion.  I am open to discussion.

app/models/claimant.rb
```diff
@@ -13,7 +13,7 @@ class Claimant < ApplicationRecord
...
+  def self.create_without_intake!(participant_id:, payee_code:)
```
leikkisa: Unfortunately we can't use this during intake because there's some extra validations whose errors we want to send back to the UI

app/models/serializers/work_queue/veteran_record_request_serializer.rb
```diff
@@ -8,18 +8,13 @@ def self.decision_review(object)
   end
 
   def self.claimant_name(object)
-    if decision_review(object).veteran_is_not_claimant
-      # TODO: support multiple?
-      decision_review(object).claimants.first.try(:name)
-    else
-      decision_review(object).veteran_full_name
```
leikkisa: if the veteran is the claimant, the claimant name is the same as the veteran full name.

spec/models/claim_review_intake_spec.rb
```diff
@@ -0,0 +1,167 @@
+# frozen_string_literal: true
```
leikkisa: Created this due to Code climate complaining about duplicate code. #review is not implemented separately for SCs or HLRs, it's actually in the claim_review_intake file.  So I took it out of their individual specs and added one here.

app/models/decision_review.rb
```diff
@@ -162,35 +162,31 @@ def start_review!
...
 
+  # Creates claimants for automatically generated decision reviews
   def create_claimants!(participant_id:, payee_code:)
     remove_claimants!
-    claimants.create_from_intake_data!(participant_id: participant_id, payee_code: payee_code)
+    claimants.create_without_intake!(participant_id: participant_id, payee_code: payee_code)
   end
 
   def remove_claimants!
-    claimants.destroy_all unless claimants.empty?
```
leikkisa: This wasn't actually destroying the claimant in the appeal intake model test, I think maybe because of the associations.  I don't think we need/want to remove associations here anyway.

app/models/appeal.rb
```diff
@@ -227,9 +227,7 @@ def advanced_on_docket?
            :last_name,
            :name_suffix, to: :veteran, prefix: true, allow_nil: true
 
-  def appellant
-    claimants.first
-  end
+  alias appellant claimant
```
tomas-nava: cool, didn't know you could do this.

app/models/board_grant_effectuation.rb
```diff
@@ -135,7 +135,7 @@ def build_end_product_establishment
       source: decision_document,
       veteran_file_number: veteran.file_number,
       claim_date: decision_document.decision_date,
-      claimant_participant_id: claimaint_participant_id,
+      claimant_participant_id: claimant_participant_id,
```
tomas-nava: 😨

app/models/decision_review.rb
tomas-nava: This seems right, since we're using `participant_id` to identify a `Person` and associate them with a `Claimant`; if a claimant record is deleted it can be recreated when needed without data loss.

spec/controllers/idt/api/appeals_controller_spec.rb
tomas-nava: I agree with this decision.

Issue #12348, opened 2019-10-09
Add Virtual Hearing Type to Column on Hearing Schedule List


lomaxap:
![Screenshot 2019-10-09 16.49.08.png](https://images.zenhubusercontent.com/5cd05a11cdc0ba05ef4f13f6/60a0939a-9835-4a2a-8c96-fb963382e70a)

The Your Hearing Schedule page displays Virtual in the Type column, and Virtual is included as a value in the column filter.
- If there are multiple Hearing Types for one day, all types are listed, separated by a comma.

pshahVA:
@ferristseng, can we please add a screenshot, either in the ticket or the pull request.

ferristseng:
@pshahVA added it to the pull request in the description

PR #12598, opened 2019-11-05
Add Virtual type to daily docket

### LINKED ISSUE #12349: Add Virtual Hearing Type to Daily Docket Print
The print view of the Daily Docket displays the Type.

https://va-gov.invisionapp.com/share/F7U0FOLT63R#/screens/385699963

### Objective

When a user downloads a printable version of their daily docket, it should reflect that the hearing is virtual

ferristseng:
Resolves #12349

### Description

  * Move `readable_request_type` to serializer for hearing days
  * Update `readable_request_type` to consider virtual hearings
  * Fix issue where `representative_email` was not an optional field

---

### Screenshots

On Hearing Schedule:

Screen Shot 2019-11-05 at 1 10 01 PM

---

On Daily Docket:

Screen Shot 2019-11-05 at 1 10 29 PM

---

On Printed Daily Docket:

Screen Shot 2019-11-05 at 1 11 36 PM

### DIFF
--- a/app/repositories/virtual_hearing_repository.rb
+++ b/app/repositories/virtual_hearing_repository.rb
@@ -17,24 +17,5 @@ class VirtualHearingRepository
 
       virtual_hearings_for_ama_hearings + virtual_hearings_for_legacy_hearings
     end
-
-    def hearing_day_has_virtual_hearing?(hearing_day)
-      virtual_hearing_for_ama_hearing_exists = VirtualHearing
-        .where(hearing_type: Hearing.name)
-        .joins("INNER JOIN hearings ON hearings.id = virtual_hearings.hearing_id")
-        .where("hearings.hearing_day_id = :hearing_day_id", hearing_day_id: hearing_day.id)
-        .exists?
-
-      # Small optimization: avoids a second query, if there is already an AMA hearing!
-      return true if virtual_hearing_for_ama_hearing_exists
-
-      virtual_hearing_for_legacy_hearing_exists = VirtualHearing
-        .where(hearing_type: LegacyHearing.name)
-        .joins("INNER JOIN legacy_hearings ON legacy_hearings.id = virtual_hearings.hearing_id")
-        .where("legacy_hearings.hearing_day_id = :hearing_day_id", hearing_day_id: hearing_day.id)
-        .exists?
-
-      virtual_hearing_for_legacy_hearing_exists
-    end
   end
 end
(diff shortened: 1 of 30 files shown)


app/serializers/hearings/hearing_day_serializer.rb
```diff
@@ -13,7 +13,13 @@ class HearingDaySerializer
...
   attribute :lock
   attribute :notes
-  attribute :readable_request_type
+  attribute :readable_request_type do |hearing_day|
+    if VirtualHearingRepository.hearing_day_has_virtual_hearing?(hearing_day)
+      "Video, Virtual"
+    else
+      Hearing::HEARING_TYPES[hearing_day.request_type.to_sym]
+    end
+  end
```
ferristseng: Does it make sense to do this on the backend?

app/repositories/virtual_hearing_repository.rb
```diff
@@ -17,5 +17,24 @@ def ready_for_deletion
...
+      virtual_hearing_for_legacy_hearing_exists = VirtualHearing
+        .where(hearing_type: LegacyHearing.name)
+        .joins("INNER JOIN legacy_hearings ON legacy_hearings.id = virtual_hearings.hearing_id")
+        .where("legacy_hearings.hearing_day_id = :hearing_day_id", hearing_day_id: hearing_day.id)
+        .exists?
+
+      virtual_hearing_for_legacy_hearing_exists
+    end
   end
 end
```
lomaxap: i think this is going to slow down the Schedule list page by a lot since it would introduce as many as 2 queries per hearing day. 

I've been trying to think of clever ways around this, but I'm starting to think the best solution is the add a column to `HearingDay` and update it in an `after_update` hook on the `VirtualHearing`s model. This feels like a lot of engineering for one field in a table, but I don't we should use this query in a serializer for all hearing days. 

I'd be a fan of removing the #12348 from this PR and addressing it separately.

ferristseng: I'll remove changes for #12348 for now.

Another idea I had was to pass a list of associated hearings as context to the serializer, since most of the routes in the hearing day controller are already fetching that info. The column approach would work too, but I agree it feels like a disproportionate amount of engineering for such a small change to the UI.

lomaxap: Yea, with the exception of when we only need to call `HearingDayRange#load_days` if the user isn't a VSO or Judge, we generally do always also query Hearings. I think we could start with that approach and keep an eye on load times.

PR #12599, opened 2019-11-05
Remove modern recordkeeping, always use promulgation date as decision date

pkarman:
connects #11840

See https://dsva.slack.com/archives/C754H751T/p1572906326036300?thread_ts=1572904848.036200&cid=C754H751T

### Description
Continue refining the definition of `contestable?` rating decision to be more inclusive. We now include all rating decisions where the disability date is near the promulgation or profile date, and ignore the arbitrary "modern recordkeeping" definition implemented earlier.

The goal is to increase the number of rating decisions visible on the Add Issues screen during Intake.

### DIFF
--- a/app/models/rating_decision.rb
+++ b/app/models/rating_decision.rb
@@ -70,6 +70,8 @@ class RatingDecision
   def contestable?
     return true if rating_issue?
 
+    return false unless disability_date
+
     date_near_promulgation_date?(disability_date.to_date) || date_near_profile_date?(disability_date.to_date)
   end
 
--- a/app/models/rating_decision.rb
+++ b/app/models/rating_decision.rb
@@ -63,16 +63,20 @@ class RatingDecision
     service_connected? ? service_connected_decision_text : not_service_connected_decision_text
   end
 
+  # the decision date is the formal date under which this decision will be grouped in the UI.
+  # We currently use promulgation_date to be consistent with RatingIssue, even though that might
+  # not be the actual decision date. See effective_date for why.
   def decision_date
     promulgation_date
   end
 
   def contestable?
-    return true if rating_issue?
+    return false if rating_issue?
 
-    return false unless disability_date
+    # TODO
+    #return false unless disability_date
 
-    date_near_promulgation_date?(disability_date.to_date) || date_near_profile_date?(disability_date.to_date)
+    true
   end
 
   def rating_issue?
@@ -95,6 +99,11 @@ class RatingDecision
     type_name == "Service Connected"
   end
 
+  # The effective date is what we call the date when the rating decision appears to have been made,
+  # even if the official legal date is the promulgation date. Because of the way disability evaluations
+  # are presented in the API response from BGS, it's not clear if each evaluation represents a new
+  # decision or not. We err on the side of always including each evaluation listed, in order to make sure
+  # all issues can be Intaken.
   def effective_date
     effective_start_date_of_original_decision || disability_date
   end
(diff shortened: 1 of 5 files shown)


app/models/rating_decision.rb
```diff
@@ -100,6 +93,10 @@ def service_connected?
     type_name == "Service Connected"
   end
 
+  def effective_date
```
pkarman: made this public method for future use in UI

app/models/rating_decision.rb
```diff
@@ -67,17 +64,15 @@ def decision_text
   end
 
   def decision_date
-    return promulgation_date if rating_issue? || effective_date_near_promulgation_date?
-
-    effective_date
+    promulgation_date
```
nanotone: Since there was a discussion about this change here, and the method is now a bit sparse, I'm wondering if it could use a one- or two-line comment about why decision date is now just the promulgation date?

app/models/rating_decision.rb
```diff
@@ -110,32 +109,20 @@ def effective_start_date_of_original_decision
...
+  def date_near_promulgation_date?(the_date)
+    the_date.between?((the_promulgation - GRACE_PERIOD.days), (the_promulgation + GRACE_PERIOD.days))
   end
 
-  def effective_start_date_prior_to_modern_recordkeeping?
-    return false unless effective_start_date_of_original_decision
-
-    effective_start_date_of_original_decision.to_date < MODERN_RECORDKEEPING_CUTOFF_DATE.to_date
+  def the_promulgation
+    @the_promulgation ||= promulgation_date.to_date
```
nanotone: Curious about these `the_*` private methods -- what's the intention behind this (seemingly dramatic) naming?

app/models/rating_decision.rb
```diff
@@ -67,17 +64,15 @@ def decision_text
...
-
-    effective_date
+    promulgation_date
   end
 
   def contestable?
     return true if rating_issue?
 
-    return false unless decision_date
+    return false unless disability_date
```
leikkisa: I don't think this hurts anything, but this line could be removed.

app/models/rating_decision.rb
```diff
@@ -67,17 +64,15 @@ def decision_text
...
 
   def decision_date
-    return promulgation_date if rating_issue? || effective_date_near_promulgation_date?
-
-    effective_date
+    promulgation_date
   end
 
   def contestable?
     return true if rating_issue?
```
leikkisa: I thought this would be false (it's not contestable if it's a rating issue). I see that they get rejected in the contestable issue generator, but it seems like we could exclude them here. Otherwise, do we need the `contestable?` method at all?

app/models/rating_decision.rb
```diff
@@ -67,17 +64,15 @@ def decision_text
...
 
-    return false unless decision_date
+    return false unless disability_date
 
-    effective_date_near_promulgation_date? || effective_date_near_profile_date?
+    date_near_promulgation_date?(disability_date.to_date) || date_near_profile_date?(disability_date.to_date)
```
leikkisa: Do we want to get rid of these, so that all of the disabilities have at least one version come through?

app/models/rating_decision.rb
leikkisa: I think we can get rid of all of these methods, something like 97% of them were between profile and promulgation (probably more with the buffer), and I don't see a reason for excluding the ones that were older than that.

Here's one example from the spreadsheet:
> Service connection for tinnitus is granted with an evaluation of 10 percent effective March 21, 2011.
This one has profile/promulgation dates of 2014 but a disability date of 2011.  For the ones that were outside of the range, I noticed a pattern of the disability date being aligned with the effective date.

PR #12611, opened 2019-11-05
Rubaiyat/create virtual hearing conference job

### LINKED ISSUE #12331: VirtualHearingConferenceJob
### Background
[Full tech spec](https://docs.google.com/document/d/1dbpHbywtPlxk1SzWnW-SuHzzzmDcM1Y7-OopAacWStM/edit). When a hearing is switched from central or video to a virtual hearing, we need to send an email to the veteran, the representative, and the judge and schedule a conference via the Pexip API (#12314)

When a virtual hearing is cancelled or postponed, we need to be able to send an email to the veteran, representative, and judge that the hearing was cancelled and remove the conference from Pexip.

### Objective
Create a job to be performed async (but not scheduled) that 

1. Creates a conference via the Pexip API (if it hasn't already been) and updates `virtual_hearings` table `conference_created` to `true` and stores the id.
2. Sends a confirmation email to the veteran (if one has not already been sent) and updates `virtual_hearing` table `veteran_email_sent` to `true`
3. Sends a confirmation email to the judge  (if one has not already been sent) and updates `virtual_hearing` table `judge_email_sent` to `true`
4. Sends a confirmation email to the representative  (if one has not already been sent) and updates `virtual_hearing` table `representative_email_sent` to `true`

Create a job to be performed async (but not scheduled) that 

1. Removes a conference via the Pexip API (if it hasn't already been) and updates `virtual_hearings` table `conference_deleted` to `true` and stores the id.
2. Sends a cancellation email to the veteran (if one has not already been sent) and updates `virtual_hearing` table `veteran_email_sent` to `true`
3. Sends a cancellation email to the judge  (if one has not already been sent) and updates `virtual_hearing` table `judge_email_sent` to `true`
4. Sends a cancellation email to the representative  (if one has not already been sent) and updates `virtual_hearing` table `representative_email_sent` to `true`

### Considerations
In a separate issue, we will address scenarios where we're unable to create a conference or 1 of the 3 emails fail (see #12337).

We are using the `..email_sent` columns to determine whether or not a confirmation/cancellation emails should or should not be sent. See #12332 for details

### Acceptance Criteria
- [x] Verify that `perform_now` updates appropriate fields on the `virtual_hearings` table

rubaiyat22:
Resolves #12331

### Description
- write a job to create virtual hearing conferences
- send emails to users if a conference has been created
- unit tests
- write module PexipClient & class SendEmail which the delete job can also use


app/jobs/virtual_hearings/send_email.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+module VirtualHearings::SendEmail
+  attr_reader :virtual_hearing
+
+  def send_emails(virtual_hearing:, type:)
+    @virtual_hearing ||= virtual_hearing
```
lomaxap: Since this is a module, you have access to the instance variables that you set in the job, so you don't need to pass the virtual_hearing param.

You may want to at some point consider making this a service that you initialize with a `virtual_hearing` and `type` and is invoke with a `call` method.

app/jobs/virtual_hearings/send_email.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+module VirtualHearings::SendEmail
```
lomaxap: I'm not sure this should be a module if no other class is using this.

rubaiyat22: I wanted delete_conference_job to use it since it would also send emails to users

app/jobs/virtual_hearings/send_email.rb
rubaiyat22: I like the service idea! and I implemented it! Now this service can be used by both create and delete job

app/jobs/virtual_hearings/create_conference_job.rb
```diff
@@ -0,0 +1,52 @@
...
+      )
+
+      if resp.error
+        Rails.logger.info "CreateConferenceJob failed: #{resp.error.message}"
+        fail resp.error
+      end
+
+      virtual_hearing.conference_id = resp.data[:conference_id]
+      virtual_hearing.status = :active
+      virtual_hearing.save
```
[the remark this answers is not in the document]
rubaiyat22: That's my bad, i don't think i can save symbol in an active record like this. Fixed it!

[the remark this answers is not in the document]
rubaiyat22: Yes, it does coerce it to a string!

Issue #12520, opened 2019-10-30
Async job CSV export


pkarman:
### User story
As a Caseflow user, I need to export Async Job metadata as a CSV in order to perform analysis outside of Caseflow.

### Acceptance criteria
- The /jobs page has a "Download as CSV" link, when clicked responds with a file named in the convention `async-jobs-YMD.csv`

#### Background/context
Requested by AMO during breakout session 10/23.

#### Success criteria
Include as needed, especially for issues that aren't part of epics.

#### Technical notes
The AsyncableJobsReporter class already has a CSV export function. Should be able to wire that up to a new endpoint on the AsyncableJobsController.

PR #12626, opened 2019-11-06
Async job CSV export

nanotone:
Connects #12520

### Description
Adds
- a `Download as CSV` button on the `/jobs` page, and
- a CSV format responder for the `/jobs` endpoint. I went this route because the content (a list of async jobs) is the same as what `asyncable_jobs#index` already serves, and the only difference is in the content type, which ought to be handled with HTTP semantics.

Screenshots of the new jobs page layout as per design advice from @rutvigupta-design:

With many job types:
Screen Shot 2019-11-07 at 12 50 05

With a few job types:
Screen Shot 2019-11-07 at 12 59 16

Note: the many vs few job types UI was previously in place, and I tried not to change it when tweaking the layout to accommodate the new button. 

### Acceptance Criteria
- The `/jobs` page has a "Download as CSV" link, which when clicked returns a file named in the convention `async-jobs-YMD.csv`

### Testing Plan
1. Load the `/jobs` page, ideally with some jobs in the DB.
2. Click the `Download as CSV` button.
3. Verify that the CSV file has a `async-jobs-YMD.csv` name, and the expected contents.
4. New controller spec, new feature spec.

### DIFF
--- a/app/controllers/asyncable_jobs_controller.rb
+++ b/app/controllers/asyncable_jobs_controller.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "date"
-
 class AsyncableJobsController < ApplicationController
   include PaginationConcern
 
@@ -18,7 +16,7 @@ class AsyncableJobsController < ApplicationController
       format.html
       format.csv do
         jobs_as_csv = AsyncableJobsReporter.new(jobs: jobs).as_csv
-        filename = Time.now.utc.strftime("async-jobs-%Y%m%d.csv")
+        filename = Time.zone.now.strftime("async-jobs-%Y%m%d.csv")
         send_data jobs_as_csv, filename: filename
       end
     end
--- a/app/controllers/asyncable_jobs_controller.rb
+++ b/app/controllers/asyncable_jobs_controller.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "date"
+
 class AsyncableJobsController < ApplicationController
   include PaginationConcern
 
@@ -11,6 +13,14 @@ class AsyncableJobsController < ApplicationController
   def index
     if allowed_params[:asyncable_job_klass]
       @jobs = asyncable_job_klass.potentially_stuck.limit(page_size).offset(page_start)
+    end
+    respond_to do |format|
+      format.html
+      format.csv do
+        jobs_as_csv = AsyncableJobsReporter.new(jobs: jobs).as_csv
+        filename = Date.today.strftime("async-jobs-%Y%m%d.csv")
+        send_data jobs_as_csv, filename: filename
+      end
     end
   end
 
--- a/app/controllers/asyncable_jobs_controller.rb
+++ b/app/controllers/asyncable_jobs_controller.rb
@@ -18,7 +18,7 @@ class AsyncableJobsController < ApplicationController
       format.html
       format.csv do
         jobs_as_csv = AsyncableJobsReporter.new(jobs: jobs).as_csv
-        filename = Date.today.strftime("async-jobs-%Y%m%d.csv")
+        filename = Time.now.utc.strftime("async-jobs-%Y%m%d.csv")
         send_data jobs_as_csv, filename: filename
       end
     end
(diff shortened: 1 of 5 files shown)


leikkisa:
@nanotone Maybe we should put the "All jobs" to the left, closer to the filter, since that's what it's acting on.  I like the location to the right and at the top for downloading a CSV, but I agree that design might have better input.

nanotone:
@rutvigupta-design @geronimoramos -- @jcq suggested tagging y'all for any quick feedback on the design for this admin page. At minimum, I imagine the `Download as CSV` button could use a margin to set it apart from `All jobs`?

app/controllers/asyncable_jobs_controller.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "date"
```
pkarman: should not need this explicitly. Rails always loads `date` module.

app/controllers/asyncable_jobs_controller.rb
```diff
@@ -12,6 +14,14 @@ def index
     if allowed_params[:asyncable_job_klass]
       @jobs = asyncable_job_klass.potentially_stuck.limit(page_size).offset(page_start)
     end
+    respond_to do |format|
+      format.html
+      format.csv do
+        jobs_as_csv = AsyncableJobsReporter.new(jobs: jobs).as_csv
+        filename = Time.now.utc.strftime("async-jobs-%Y%m%d.csv")
```
pkarman: I'd suggest `Time.zone.now.strftime` instead, so that it is relative to the timezone of the app (ET). This will be the least surprising thing for folks in PT.

client/app/asyncableJobs/components/AsyncModelNav.jsx
```diff
@@ -37,6 +37,7 @@ export default class AsyncModelNav extends React.PureComponent {
       <div>
         {this.modelNameLinks()}
         <a style={{ float: 'right' }} href="/jobs" className="cf-link-btn">All jobs</a>
+        <a style={{ float: 'right' }} href="/jobs.csv" className="cf-link-btn">Download as CSV</a>
```
pkarman: this will always load all jobs, and not just the jobs on the current page. That will surprise folks, who might expect to see only in the CSV what they can see on the webpage.

I think you want "the-current-url.csv"

rutvigupta-design:
@nanotone (sorry if this is too late) but here are my suggestions: 
- I agree with Sally, I think moving "All Jobs" next to the filter makes sense. If this isn't possible, I would put "All Jobs" to the right of "Download as CSV" and put a .5em margin between them. 
- **Not sure if this is out of scope for this:** I would also increase the space between 
a) the "Last updated" text and the "Filter by Job title" button and b) the buttons overall and the divider below them

Let me know if I can further clarify any of this!

nanotone:
Tweaked the jobs page layout and updated the screenshots in the PR description at top.

app/controllers/asyncable_jobs_controller.rb
```diff
@@ -12,6 +12,14 @@ def index
     if allowed_params[:asyncable_job_klass]
       @jobs = asyncable_job_klass.potentially_stuck.limit(page_size).offset(page_start)
     end
+    respond_to do |format|
+      format.html
+      format.csv do
+        jobs_as_csv = AsyncableJobsReporter.new(jobs: jobs).as_csv
```
pkarman: `job` here is paginated, which is probably not what we want. I.e. when we click Download we expect all the jobs in the current filter, not just the ones on the current page.

OTOH for some filters (like request issues) we'd get a CSV of 10k+ rows which will not perform very well.

client/app/asyncableJobs/components/AsyncModelNav.jsx
```diff
@@ -31,18 +31,23 @@ export default class AsyncModelNav extends React.PureComponent {
...
+      '/jobs';
 
     return <div>
       <strong>Last updated:</strong> {moment(this.props.fetchedAt).format(DATE_TIME_FORMAT)}
-      <div>
+      <div style={{ marginTop: '.5em' }}>
+        <a style={{ marginRight: '.5em' }} href="/jobs" className="cf-link-btn">All jobs</a>
         {this.modelNameLinks()}
-        <a style={{ float: 'right' }} href="/jobs" className="cf-link-btn">All jobs</a>
+        <a style={{ float: 'right' }} href={`${currentPath}.csv`} className="cf-link-btn">Download as CSV</a>
```
pkarman: +1

PR #12630, opened 2019-11-06
Always fetch user queue

### LINKED ISSUE #12621: TypeError - app/queue/ColocatedTaskListView in length error: Cannot read property 'length' of undefined
[Sentry alert](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/7294/events/latest/). Looks like `tasksForTab` is returning undefined. My first guess would be that `queueConfig` is not getting set properly on the front end. Assumed related to #12499. Unable to reproduce locally. 

### What we know

- Started happening after our deploy on [Monday, November 4th, 2019](https://dsva.slack.com/archives/C7VNGMYUQ/p1572882316105300)
- Has only occurred with 3 users (2124, 3678, 3772)
- All users are VSOs
- Not all users belong to the same VSO
- Not all users are admins for their VSO
- All errors are thrown by following these steps:
1. Visiting the VSO org page
2. Selecting a case
3. Assigning OR reassigning that case
4. Automatically getting returned to the user's queue

lowellrex:
Resolves #12621.

We were encountering the "white screen of death" in the scenario below because we were holding on to the `queueConfig` from the organization queue. This change makes it so we always refetch the queue config.

### Original reproduction steps
1. Log in as BILLIE_VSO
1. Navigate to http://localhost:3000/organizations/veterans-service-organization?tab=unassigned&page=1
1. Select a case and assign it to anyone
1. **Via the back button**, navigate back to the unassigned tab
1. Select a case and assign the case to anyone
1. White screen of death.


spec/feature/queue/ama_queue_spec.rb
```diff
@@ -859,4 +859,54 @@ def valid_document_id
       end
     end
   end
+
+  describe "Navigating between team and individual queues" do
```
lowellrex: I wrote this test to mirror exactly what we do manually in the browser to demonstrate this bug, but for some reason this test doesn't fail on master. That is, the individual queue displays after the second task assignment. I can't figure out why but I figure I'll leave this test in here because it shows how we can recreate this situation in our local development environments.

Issue #12421, opened 2019-10-17
SCM: Migrate existing Judge Teams to use Judge & Attorney Roles


lomky:
Followup to #12420 to update the existing JudgeTeams to mark their Judges & Attorneys before we add any new admins to teams.

### Acceptance Criteria
 - [x] Current JudgeTeam admins have the role `JudgeTeamLead`
 - [x] Current JudgeTeam non-admins have the role `DecisionDraftingAttorney`

Part of #11801

kevmo:
Estimating at 2.  Do as a Rake task.

lomky:
Dry run in production found a small bug with passing in an argument for an alert message. 

Also discovered there are two extant judge teams with multiple admins already (!!). These two teams will not be affected by the rake task, and will be manually corrected instead as they are in what is currently considered a bad state.

Notes:
```
# Existing State for JudgeTeamLead & DecisionDraftingAttorney
irb(main):001:0> JudgeTeamLead.first
=> nil
# No JudgeTeamLeads
irb(main):007:0> DecisionDraftingAttorney.all       
[...]

PR #12636, opened 2019-11-07
Rake task to assign all existing JudgeTeam Members a JudgeTeamRole

lomky:
Connects #12421 

### Description
Creates a Rake task to process all the JudgeTeams and assign the Admin `JudgeTeamLead` and the nonadmins `DecisionDraftingAttorney`s. 

Will alert & not process empty judge teams, teams with no admin, teams with multiple admins.

## Test Plan
Rake task run on dev without any issues.

For production run, task should be run with `dry_run` and results reviewed before invoking final migration.

Ultimately we are populating a currently empty table with data, and no data is being deleted. Risk is low.


spec/workflows/bulk_task_reassignment_spec.rb
```diff
@@ -3,7 +3,7 @@
 require "support/database_cleaner"
 require "rails_helper"
 
-describe BulkTaskAssignment, :postgres do
+describe BulkTaskReassignment, :postgres do
```
lomky: Cleanup; Unrelated to PR

app/workflows/assign_judgeteam_roles.rb
```diff
@@ -0,0 +1,55 @@
...
+        if @dry_run
+          puts "Would make #{message}"
+        else
+          org_user = OrganizationsUser.existing_record(admin, judge_team)
+          JudgeTeamLead.find_or_create_by(organizations_user: org_user)
+          Rails.logger.info("Setting #{message}")
+        end
+      end
+      nonadmins = judge_team.attorneys
+      nonadmins.each do |atty|
```
yoomlam: Could be shortened to `judge_team.nonadmins.each do |atty|`?

Issue #12518, opened 2019-10-30
Motions attorney completes Deny or Dismiss motion to vacate


leikkisa:
### User or job story
As a motions attorney, I want the ability to close a review motion to vacate (denial or dismissal) task, so that I can generate and send a letter to the Veteran.

### Acceptance criteria
- This work is behind a feature toggle: "review_motion_to_vacate".
- Motions attorneys have the action "Mark task as complete" on their Deny or Dismiss Motions to Vacate tasks
- When the Motions attorney completes the motion to vacate, all open parent tasks are also completed
- The user gets routed to their queue with a flash alert indicating that the case has been completed
- Please include final screenshots in the issue

### Open questions
- What happens to the instructions or context the Motions attorney enters?

### Release notes
Caseflow now supports Motions Attorneys' ability to complete denied or dismissed motions to vacate.

#### Designs
![Screen Shot 2019-10-30 at 12.43.35 PM.png](https://images.zenhubusercontent.com/58f6b6eb51aa2c30577bb60e/0450ff89-3d06-493a-9f90-ccc2cf5aafa2)

![Screen Shot 2019-10-30 at 12.43.44 PM.png](https://images.zenhubusercontent.com/58f6b6eb51aa2c30577bb60e/d1688e41-17c7-4de8-84b1-712996961c85)

![Screen Shot 2019-10-30 at 12.43.58 PM.png](https://images.zenhubusercontent.com/58f6b6eb51aa2c30577bb60e/bddb0589-7432-4dce-a1d4-ceb6a85a5f79)

jimruggiero:
@leikkisa @geronimoramos @rutvigupta-design C/Should MTV notes go here?
![MTV_Notes](https://user-images.githubusercontent.com/52474028/67912197-8dbdc580-fb5f-11e9-9284-c1e0fdc4747c.jpg)

jimruggiero:
@rutvigupta-design @geronimoramos Can you confirm that the instructions text box in the modal above is optional (i.e. does not require content) to Mark complete?

rutvigupta-design:
@jimruggiero my understanding was the notes appear as **task instructions** for the judge/attorneys (i.e. under the "assigned by" next to the Actions drop-down. The notes mentioned in the image refer to notes on specific issues that the veteran is requesting, so the pattern of relating that to tasks in this case might be confusing. 

Yes, the Mark complete instructions text box should be optional.

leikkisa:
[...]

PR #12638, opened 2019-11-07
MTV | Atty completes denied/dismissed tasks

jcq:
Connects #12518

### Description
This adds support for the "Mark Task Complete" workflow for denied/dismissed MTV tasks. We use the normal `MARK_COMPLETE` task action, but in this case we needed the success message to not reference the `assigned_by`, but instead needed it to reference Lit Support. This adds support for tasks specifying the value to go in there via optional implementation of `completion_contact` method.

### Acceptance Criteria
- [ ] Code compiles correctly


jcq:
### Updated Screen
This includes the explicit callout to Litigation Support. 
Screen Shot 2019-11-07 at 10 10 48

**Note:** this shows a UX issue with the placement of the "Switch Views" dropdown that should be addressed in a separate PR (as it isn't unique to this implementation)

spec/models/tasks/dismissed_motion_to_vacate_task_spec.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+require "support/database_cleaner"
```
leikkisa: This test is an example of one that isn't using the DB (also doesn't need rails_helper), but thanks for adding a spec.

spec/feature/queue/motion_to_vacate_spec.rb
```diff
@@ -439,6 +439,56 @@
...
+      # create(:staff, :judge_role, sdomainid: judge.css_id)
+      lit_support_team.add_user(motions_attorney)
+      # judge_team.add_user(drafting_attorney)
+      # ["John Doe", "Jane Doe"].map do |name|
+      #   judge_team.add_user(create(:user, full_name: name))
+      # end
+      FeatureToggle.enable!(:review_motion_to_vacate)
+
+      vacate_motion_mail_task.update(status: Constants.TASK_STATUSES.completed)
+      judge_address_motion_to_vacate_task.update(status: Constants.TASK_STATUSES.completed)
```
leikkisa: You could set this to completed when you're creating the judge_address_motion_to_vacate_task

spec/feature/queue/motion_to_vacate_spec.rb
```diff
@@ -439,6 +439,56 @@
...
+    end
+
+    after { FeatureToggle.disable!(:review_motion_to_vacate) }
+
+    it "shows modal" do
```
leikkisa: This test looks like it's covering the full completion now so the name could cast a wider net

spec/feature/queue/motion_to_vacate_spec.rb
```diff
@@ -474,6 +524,20 @@ def judge_send_to_dispatch(user:, appeal:)
     find("div", class: "Select-option", text: "Ready for Dispatch").click
   end
 
+  def complete_motion_to_vacate(user:, appeal:, task:)
```
leikkisa: I'm guessing you'll be re-using this elsewhere (like testing a flow for completing dismissed), but if not, it's more straightforward to put this code directly into the test.

Some of this code is generic enough (the marking complete) that it could potentially go into a task_helper file for specs.  You don't have to do that here but we could consider that as a tech improvement ticket, for some of the repetitive actions in task specs, so that the tests can be focused on what's specific to this test.

spec/feature/queue/motion_to_vacate_spec.rb
```diff
@@ -474,6 +524,20 @@ def judge_send_to_dispatch(user:, appeal:)
...
+    visit "/queue/appeals/#{appeal.uuid}"
+
+    find(".Select-placeholder", text: COPY::TASK_ACTION_DROPDOWN_BOX_LABEL).click
+    find("div", class: "Select-option", text: Constants.TASK_ACTIONS.MARK_COMPLETE.label).click
+
+    expect(page.current_path).to eq("/queue/appeals/#{appeal.uuid}/tasks/#{task.id}/modal/mark_task_complete")
+
+    click_on "Mark complete"
+
+    expect(page).to have_content(format(COPY::MARK_TASK_COMPLETE_CONFIRMATION_DETAIL, task.completion_contact))
```
leikkisa: Let's check that the status on all of the relevant tasks gets updated to completed here (in the DB) too.

spec/feature/queue/motion_to_vacate_spec.rb
```diff
@@ -439,6 +439,56 @@
...
+    before do
+      # create(:staff, :judge_role, sdomainid: judge.css_id)
+      lit_support_team.add_user(motions_attorney)
+      # judge_team.add_user(drafting_attorney)
+      # ["John Doe", "Jane Doe"].map do |name|
+      #   judge_team.add_user(create(:user, full_name: name))
+      # end
+      FeatureToggle.enable!(:review_motion_to_vacate)
+
+      vacate_motion_mail_task.update(status: Constants.TASK_STATUSES.completed)
```
leikkisa: This should actually get completed by completing the denied motion to vacate task.  Let's add a check for this at the end of the test.  All of the preceding tasks should get completed once this last step is complete.

spec/feature/queue/motion_to_vacate_spec.rb
```diff
@@ -439,6 +439,56 @@
...
+        appeal: appeal,
+        assigned_by: judge,
+        assigned_to: motions_attorney
+      )
+    end
+
+    before do
+      # create(:staff, :judge_role, sdomainid: judge.css_id)
+      lit_support_team.add_user(motions_attorney)
+      # judge_team.add_user(drafting_attorney)
```
leikkisa: Looks like you don't need these

app/models/tasks/dismissed_motion_to_vacate_task.rb
```diff
@@ -8,4 +8,8 @@ def self.label
   def self.org(_user)
     LitigationSupport.singleton
   end
+
+  def completion_contact
+    "the Litigation Support team"
+  end
```
leikkisa: Not a comment for now because I don't know the design for the granted flows yet, but these might eventually belong in DecidedMotionToVacateTask.

Do you think we should put these into the COPY file instead?

client/app/queue/components/CompleteTaskModal.jsx
```diff
@@ -110,8 +99,13 @@ class CompleteTaskModal extends React.Component {
...
     return `${String.fromCodePoint(assignedBy.firstName.codePointAt(0))}. ${assignedBy.lastName}`;
   };
+
+  getTaskConfiguration = () => {
+    return taskActionData(this.props) || {};
+  };
+
   getContentArgs = () => ({
-    assignerName: this.getTaskAssignerName(),
+    contact: this.getTaskConfiguration().contact || this.getTaskAssignerName(),
```
leikkisa: Cool, I could see this being useful in future tasks too

spec/feature/queue/motion_to_vacate_spec.rb
```diff
@@ -439,6 +439,56 @@
     end
   end
 
+  describe "Attorney Completes Denied Motion to Vacate Task" do
```
leikkisa: You could add these steps into the existing test for denied motion to vacate, and make it a full end-to-end test.  You could even put returning the draft in there, and then do the completion once the attorney has re-submitted.

leikkisa:
I found another thing we should work on while reviewing this, but the motions attorney had access to actions on the judge's task, so we should look through and fix action permissions at some point. Will write a ticket!

spec/feature/queue/motion_to_vacate_spec.rb
jcq: This is a good flag — adding the extra check surfaced the issue that parent tasks only autoclose if all of their children close *and* if the parent is an org task. I've added additional logic to account for that.

spec/feature/queue/motion_to_vacate_spec.rb
jcq: Yeah, I broke this out for likely use in other testing flows, and have now put it to use on doing dismissed as well as denied (for sake of full coverage).

Agreed that some stuff might make sense for a future refactor

spec/feature/queue/motion_to_vacate_spec.rb
jcq: Turns out it's best to not set this manually, as we want to ensure that things are actually getting completed programmatically.

app/models/tasks/dismissed_motion_to_vacate_task.rb
jcq: I'm not sure if I prefer it in COPY or not, but since that's how most of the rest of the bits are being done, I'm going to update the code to use it.

app/repositories/task_action_repository.rb
```diff
@@ -166,6 +166,10 @@ def address_motion_to_vacate_data(task, _user = nil)
       }
     end
 
+    def complete_motion_to_vacate_data(_task, _user = nil)
+      {}
+    end
+
```
leikkisa: I don't think this needs to be here

spec/feature/queue/motion_to_vacate_spec.rb
```diff
@@ -491,6 +549,28 @@ def judge_send_to_dispatch(user:, appeal:)
...
+
+    expect(page).to have_content(format(COPY::MARK_TASK_COMPLETE_CONFIRMATION_DETAIL, task.completion_contact))
+
+    expect(task.reload.status).to eq Constants.TASK_STATUSES.completed
+
+    org_task = task.reload.parent
+    expect(org_task.status).to eq Constants.TASK_STATUSES.completed
+
+    expect(abstract_motion_to_vacate_task.reload.status).to eq Constants.TASK_STATUSES.completed
+    expect(vacate_motion_mail_task.reload.status).to eq Constants.TASK_STATUSES.completed
```
leikkisa: 🎉

app/models/tasks/dismissed_motion_to_vacate_task.rb
```diff
@@ -8,4 +8,8 @@ def self.label
   def self.org(_user)
     LitigationSupport.singleton
   end
+
+  def completion_contact
+    COPY::CANCEL_TASK_CONTACT_MTV_DISMISSED
```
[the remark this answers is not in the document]
jcq: Was having trouble deciding the clearest way to consolidate them, but will go with `CANCEL_TASK_CONTACT_LIT_SUPPORT`.

Issue #12640, opened 2019-11-07
Async job CSV export unpaginated


nanotone:
### User
As a Caseflow user, I need to export all Async Jobs as a CSV (possibly filtered by job type), in order to perform analysis outside of Caseflow without concatenating many CSVs.

### Acceptance criteria
- The CSV-format responses for `/jobs` and `asyncable_jobs/JobType/jobs` are not paginated.

#### Background/context
Users of the current Download as CSV functionality will find that the downloaded CSVs are smaller than they expect. Analysis on the asyncable jobs data will normally be done on the entire dataset.

#### Resources/other links
Previous issue: #12520 addressed by #12626

PR #12660, opened 2019-11-08
Disable pagination when listing async jobs for CSV export

nanotone:
Connects #12640 

### Description
CSV exports of async jobs via the `/jobs` and `/asyncable_jobs/JobType/jobs` endpoints should not be subject to pagination. Data analysis using these exports will generally want the entire dataset without having to concatenate CSVs.

### Acceptance Criteria
- [ ] The CSV-format responses for `/jobs` and `/asyncable_jobs/JobType/jobs` are not paginated.

### Testing Plan
1. Added a test to create more than `page_size` jobs and verify that the CSV isn't truncated
2. Verified that the test fails without the corresponding controller update


spec/controllers/asyncable_jobs_controller_spec.rb
```diff
@@ -51,6 +51,20 @@
       end
 
       let(:user) { create(:default_user) }
+      let(:page_size) { 50 }
```
nanotone: Not a big fan of duplicating the `50` here, but I wasn't sure about having the test reach into the PaginationConcern to get it, either.

PR #12664, opened 2019-11-09
Search User name in autocomplete, fix zero match bug

### LINKED ISSUE #12536: User autocomplete error
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/7077/events/529834/

The org management XHR autocomplete feature has a bug the org cannot be excluded.

`NoMethodError: undefined method `-' for nil:NilClass`

pkarman:
Resolves #12536 

### Description
Looking at production logs, it is clear our users search for other users by name as well as CSS id. So support that expectation where possible.

In addition, there was a bug where zero matches were causing a 500 error due to nil object.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -5,9 +5,9 @@ class UsersController < ApplicationController
 
   def index
     return filter_by_role if params[:role]
-    return filter_by_css_id if params[:css_id]
+    return filter_by_css_id_or_name if params[:css_id]
 
-    render json: {}, status: 200
+    render json: {}, status: :ok
   end
 
   def search_by_css_id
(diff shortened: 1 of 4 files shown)


app/controllers/users_controller.rb
```diff
@@ -54,9 +54,10 @@ def filter_by_role
     end
   end
 
-  def filter_by_css_id
-    finder = UserFinder.new(css_id: css_id)
-    users = finder.users
+  def filter_by_css_id_or_name
+    # the param name is css_id for convenience but we are more generous in what we search.
+    finder = UserFinder.new(css_id: css_id, name: css_id)
+    users = finder.users || []
```
pkarman: this was the zero match bug in the sentry alert

app/services/user_finder.rb
```diff
@@ -41,6 +47,18 @@ def users_matching_css_id
...
 
+  def users_matching_name
+    return [] if name.blank?
+
+    name_segments = name.split(/\W/)
+
+    User.where(
+      "full_name ILIKE (?) OR full_name ILIKE (?)",
+      "%#{name_segments.join('%')}%",
+      "%#{name_segments.reverse.join('%')}%"
```
leikkisa: Very nice

PR #12667, opened 2019-11-11
Quiet job alerts that are not actionable

pkarman:
The daily job alerts channel should be more actionable, so make it clearer which things require attention.

### DIFF
--- a/app/services/asyncable_jobs_reporter.rb
+++ b/app/services/asyncable_jobs_reporter.rb
@@ -5,13 +5,14 @@ require "csv"
 class AsyncableJobsReporter
   attr_reader :jobs
 
-  def initialize(jobs:)
+  def initialize(jobs:, verbose: false)
     @jobs = jobs
+    @verbose = verbose
   end
 
   def as_csv
     CSV.generate do |csv|
-      csv << %w[type id submitted last_submitted attempted_at error participant_id]
+      csv << %w[type id submitted last_submitted attempted error participant_id]
       jobs.each do |job|
         klass = job.class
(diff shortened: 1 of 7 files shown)


app/services/asyncable_jobs_reporter.rb
```diff
@@ -5,13 +5,14 @@
...
-  def initialize(jobs:)
+  def initialize(jobs:, verbose: false)
     @jobs = jobs
+    @verbose = verbose
   end
 
   def as_csv
     CSV.generate do |csv|
-      csv << %w[type id submitted last_submitted attempted_at error participant_id]
+      csv << %w[type id submitted last_submitted attempted error participant_id]
```
pkarman: grammatical nit for consistency

leikkisa: It looks like there's a spec where this needs to get updated too:
asyncable_jobs_controller_spec.rb:66

spec/services/asyncable_jobs_reporter_spec.rb
```diff
@@ -85,6 +87,18 @@
...
+
+    context "when verbose is false" do
+      let(:verbose) { false }
+
+      it "is more terse" do
+        msg = <<~HEREDOC
+          HigherLevelReview has 1 total jobs in queue
+          SupplementalClaim has 5 total jobs in queue
+        HEREDOC
+        expect(subject.summarize).to eq(msg.chomp)
```
leikkisa: `chomp` is my new favorite method name ;-)

PR #12673, opened 2019-11-11
Appeal status

### LINKED ISSUE #12624: Add AppealStatus model
In order to programmatically enforce/reflect business rules for reporting about the status of an Appeal, create a new `AppealStatus` model that reflects the current business logic represented in https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/sql/bva-decision-progress.sql

Example:

```ruby
class AppealStatus
  STATUS_DEFS = {
    foo: "Human string for status"
  }

  def initialize(appeal)
    @appeal = appeal
  end

  def compute
    # task analysis here, set the internal `@status` attribute
  end

  def to_s
    STATUS_DEFS[status]
  end

  private

  attr_reader :appeal, :status
end

class Appeal
  def status
    AppealStatus.new(self)
  end
end
```

## Background

Current B3 SQL being used to compute status. Evaluate for comparison https://github.com/department-of-veterans-affairs/dsva-vacols/issues/91

pkarman:
Resolves #12624 

### Description
Adds new service `AppealStatus` to determine the BVA status value from the tasks on an Appeal.

### DIFF
--- a/app/services/appeal_status.rb
+++ b/app/services/appeal_status.rb
@@ -59,6 +59,10 @@ class AppealStatus
     DEFINITIONS[status]
   end
 
+  def as_json(_args)
+    to_sym
+  end
+
   private
 
   attr_reader :appeal
--- a/app/services/appeal_status.rb
+++ b/app/services/appeal_status.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+# Determine the status of an Appeal (symbol and string) based on its Tasks.
+
 class AppealStatus
   attr_reader :status
 
--- a/app/services/appeal_status.rb
+++ b/app/services/appeal_status.rb
@@ -14,7 +14,8 @@ class AppealStatus
     dispatched: "8. Decision dispatched",
     cancelled: "CANCELLED",
     misc: "MISC",
-    on_hold: "ON HOLD"
+    on_hold: "ON HOLD",
+    unknown: "UNKNOWN"
   }.freeze
 
   MISC_TASK_NAMES = %w[
@@ -88,6 +89,8 @@ class AppealStatus
       :misc
     elsif open_timed_hold_task?
       :on_hold
+    else
+      :unknown
     end
   end
   # rubocop:enable Metrics/MethodLength
--- a/app/services/appeal_status.rb
+++ b/app/services/appeal_status.rb
@@ -1,0 +1,164 @@
+# frozen_string_literal: true
+
+class AppealStatus
+  attr_reader :status
+
+  DEFINITIONS = {
+    not_distributed: "1. Not distributed",
+    distributed_to_judge: "2. Distributed to judge",
+    assigned_to_attorney: "3. Assigned to attorney",
+    assigned_to_colocated: "4. Assigned to colocated",
+    in_progress: "5. Decision in progress",
+    ready_for_signature: "6. Decision ready for signature",
+    signed: "7. Decision signed",
+    dispatched: "8. Decision dispatched",
+    cancelled: "CANCELLED",
+    misc: "MISC",
+    on_hold: "ON HOLD"
+  }.freeze
+
+  MISC_TASK_NAMES = %w[
+    JudgeQualityReviewTask
+    JudgeDispatchReturnTask
+    AttorneyQualityReviewTask
+    AttorneyDispatchReturnTask
+  ].freeze
+
+  ATTORNEY_TASK_NAMES = %w[
+    AttorneyTask
+    AttorneyRewriteTask
+  ].freeze
+
+  class << self
+    def colocated_task_names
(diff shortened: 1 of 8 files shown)


app/services/appeal_status.rb
```diff
@@ -0,0 +1,173 @@
+# frozen_string_literal: true
+
+# Determine the status of an Appeal (symbol and string) based on its Tasks.
+
+class AppealStatus
+  attr_reader :status
+
+  DEFINITIONS = {
```
pkarman: these definitions come from https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/sql/bva-decision-progress.sql

lowellrex:
Do the use cases of this class overlap with any of the use cases for existing appeal status code in the application? ([`ApiStatusAlerts`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/api_status_alerts.rb) or [the legacy](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/serializers/v2/legacy_appeal_status_serializer.rb) or AMA [`AppealStatusSerializer`s](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/serializers/v2/appeal_status_serializer.rb)). Do we plan on migrating the logic in the existing places where we interrogate appeal status to use this new class?

pkarman:
> Do we plan on migrating the logic in the existing places where we interrogate appeal status to use this new class?

Maybe?

Despite the name collision, the "Appeals Status API" audience is Veterans using va.gov. This new `AppealStatus` service audience is BVA employees. I think the status categories are different enough that they do not overlap except in the broadest terms.

The API, for example, just [performs a boolean check](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/serializers/v2/appeal_status_serializer.rb#L26) if an Appeal is active or not based on [`active_status?`](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/appeal.rb#L332)

This new class parses all active appeals into one of 12 buckets, for identifying workflow steps.

lowellrex:
> Despite the name collision, the "Appeals Status API" audience is Veterans using va.gov. This new AppealStatus service audience is BVA employees. I think the status categories are different enough that they do not overlap except in the broadest terms.

Gotcha gotcha.

Is the API status check a superset of some more precise Board statuses? If so, could we point the API status check to this class as well confirming that the appeal is in one of several more precise statuses?

I'm a little concerned about providing an opening for drift in the codebase. As implemented now, the VA.gov status lists an appeal as active at the Board if the appeal has finished intake. I can imagine us being tasked to delineate between whether an appeal is in case storage or being actively adjudicated. If we do this 9 months from now, I am much more confident we would re-use this code for this API extension if the logic lived in the same file.

app/services/appeal_status.rb
lomaxap: for my context, how ere these statuses settled on? Are these appeal states that the board has specifically requested they wanted status for? Curious about why we're not interested in the steps in between `not_distributed` and `distributed`

pkarman: Yes, these status categories are specific to what BVA wants to see in their reports.

pkarman:
> Is the API status check a superset of some more precise Board statuses?

Yes and no. The `Appeal.active?` method is an abbreviated superset of Board status values, but there are other checks in `Appeal.active_status?` that go beyond Board categories (by taking into account actual decisions and/or remands).

>  If so, could we point the API status check to this class as well confirming that the appeal is in one of several more precise statuses?

Would a comment in the API code suffice? As above, it's not a strict overlap.

> I can imagine us being tasked to delineate between whether an appeal is in case storage or being actively adjudicated.

Given that the Appeals Status API has no product owner/manager and was entirely developed at the request of the va.gov team, I'm not sure we can anticipate what kinds of values would be meaningful to that feature set in the future. Is "case storage" even a category under AMA?

I hear what you're saying about drift, but aside from the name I'm not sure there's any overlap here.

pkarman:
@lowellrex would it help if I renamed this service class to something like `BVAAppealStatus`?

lowellrex:
> @lowellrex would it help if I renamed this service class to something like `BVAAppealStatus`?

Sure, yeah I think that makes sense!

Alternatively we could add a specific method for internal Board use that could be replicated should we ever decide to incorporate the VA.gov logic in this class as well. Perhaps something like `AppealStatus.for_bva` and `AppealStatus.for_va_dot_gov`.

pkarman:
I renamed the class, preferring the pattern of small classes that do one thing.

pkarman:
@lomaxap @lowellrex thanks for the feedback. Ok to give this an official 💚 +1?

lowellrex:
> @lomaxap @lowellrex thanks for the feedback. Ok to give this an official 💚 +1?

Sure thing! Taking a look now.

Issue #12447, opened 2019-10-22
Choose correct claimant and payee code for board grant effectuations


leikkisa:
### User or job story
As a user who is processing a board grant effectuation, if the payee code is wrong then I have to manually create a new claim with the correct data, and the prior incorrect claim needs to be canceled.

### Acceptance criteria
- When creating a board grant effectuation EP, the claimant_participant_id should be set to the appeal's claimant's participant_id
- If the claimant has a payee_code set, the end product establishment should use that payee_code
  - Note: this will not be set for existing appeals, but I think we should set it for future appeals (which would be another ticket)
- If the claimant does not have the payee_code set, it should be calculated.  If the claimant is the veteran, it can be set to "00". If the claimant is a dependent, it should be set accordingly (see technical notes).

### Out of scope
Backfilling data

### Release notes
Board grant effectuations can support non-veteran claimants and payee codes.

#### Designs
Include screenshots or links to designs if applicable.

#### Background/context
We do not have the payee_code selector for appeals during intake, because what we've heard is that people at the board do not know what payee_codes are and wouldn't be able to choose one.

We may want to re-examine that assumption now that we can "default" the payee_code value, and now that we know that we shouldn't just use the veteran as the claimant for board grant effectuations.

Providing that option would only apply to new appeals going forward.  In order for this to apply to appeals that have already been completed, this should be a back-end fix.

#### Technical notes
- The end product establishment for board grant effectuations is created in board_grant_effectuation.rb > build_end_product_establishment
- For board grant effectuations, the end product establishment "source" is a decision document. The decision document is connected to an appeal, and the appeal has the claimant.
- There is a method that chooses the default payee code for a specific veteran relationship, located in the relationship.rb > default_payee_code
  - This method is currently used to choose a default payee code for the front-end payee_code selector.

#### Other notes
Example of this is claim_id 73192881, where the appeal had a non-veteran claimant. Claim 74791005 was manually established to fix the issue.

#### Resources/other links
https://dsva.slack.com/archives/CFBLH2LR4/p1571749027058100?thread_ts=1571168692.017200&cid=CFBLH2LR4

PR #12675, opened 2019-11-12
Choose correct claimant and payee code for board grant effectuations

tomas-nava:
Connects #12447

### Description
Choose the correct `claimant_participant_id` and `payee_code` when creating a board grant effectuation end product.

### DIFF
--- a/app/models/board_grant_effectuation.rb
+++ b/app/models/board_grant_effectuation.rb
@@ -130,12 +130,9 @@ class BoardGrantEffectuation < ApplicationRecord
     )
   end
 
-  # the end product establishment for board grant effectuations is created here
-  # decision_document.appeal.claimaints
-  # decision_document.appeal.veteran_is_not_claimant
   def build_end_product_establishment
     EndProductEstablishment.create!(
-      source: decision_document, # the decision document is connected to an appeal, and the appeal has the claimant
+      source: decision_document,
       veteran_file_number: veteran.file_number,
       claim_date: decision_document.decision_date,
       claimant_participant_id: claimaint_participant_id,
--- a/app/models/board_grant_effectuation.rb
+++ b/app/models/board_grant_effectuation.rb
@@ -151,15 +151,11 @@ class BoardGrantEffectuation < ApplicationRecord
   def claimant_payee_code
     claimant = appeal&.claimants&.first
 
-    payee_code = if claimant&.payee_code.present?
-                   claimant.payee_code
-                 elsif appeal&.veteran_is_not_claimant
-                   veteran&.relationships&.find do |relationship|
-                     relationship.participant_id == claimant.participant_id
-                   end&.default_payee_code
-                 end
-
-    payee_code || EndProduct::DEFAULT_PAYEE_CODE
+    if claimant&.payee_code.present?
+      claimant.payee_code
+    elsif appeal&.veteran_is_not_claimant
+      veteran&.relationship_with_participant_id(claimant.participant_id)&.default_payee_code
+    end || EndProduct::DEFAULT_PAYEE_CODE
   end
 
   def end_product_code
--- a/app/models/board_grant_effectuation.rb
+++ b/app/models/board_grant_effectuation.rb
@@ -152,15 +152,15 @@ class BoardGrantEffectuation < ApplicationRecord
     claimant = decision_document&.appeal&.claimants&.first
     veteran = decision_document&.appeal&.veteran
 
-    if claimant&.payee_code.present?
-      claimant.payee_code
-    elsif decision_document&.appeal&.veteran_is_not_claimant
-      veteran&.relationships&.find do |relationship|
-        relationship.participant_id == claimant.participant_id
-      end&.default_payee_code
-    else
-      "00"
-    end
+    payee_code = if claimant&.payee_code.present?
+                   claimant.payee_code
+                 elsif decision_document&.appeal&.veteran_is_not_claimant
+                   veteran&.relationships&.find do |relationship|
+                     relationship.participant_id == claimant.participant_id
+                   end&.default_payee_code
+                 end
+
+    payee_code || EndProduct::DEFAULT_PAYEE_CODE
   end
 
   def end_product_code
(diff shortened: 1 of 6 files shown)


app/models/board_grant_effectuation.rb
```diff
@@ -135,14 +135,34 @@ def build_end_product_establishment
...
+  def claimant_payee_code
+    claimant = decision_document&.appeal&.claimants&.first
+    veteran = decision_document&.appeal&.veteran
+
+    if claimant&.payee_code.present?
+      claimant.payee_code
+    elsif decision_document&.appeal&.veteran_is_not_claimant
+      veteran&.relationships&.find do |relationship|
+        relationship.participant_id == claimant.participant_id
+      end&.default_payee_code
```
tomas-nava: What should the `payee_code` be when the veteran is not the claimant and there are no relationships that match the claimant's `participant_id`?

tomas-nava:
`spec/models/decision_document_spec.rb:214` is a test regression, I will fix it.

app/models/board_grant_effectuation.rb
```diff
@@ -135,14 +135,34 @@ def build_end_product_establishment
...
     )
   end
 
+  def claimaint_participant_id
+    decision_document&.appeal&.claimants&.first&.participant_id
+  end
+
+  def claimant_payee_code
+    claimant = decision_document&.appeal&.claimants&.first
+    veteran = decision_document&.appeal&.veteran
```
leikkisa: veteran is delegated to the appeal so you should be able to get away without this variable.

app/models/board_grant_effectuation.rb
```diff
@@ -135,14 +135,34 @@ def build_end_product_establishment
...
+      payee_code: claimant_payee_code,
       code: end_product_code,
       station: end_product_station,
       benefit_type_code: veteran.benefit_type_code,
       user: User.system_user
     )
   end
 
+  def claimaint_participant_id
+    decision_document&.appeal&.claimants&.first&.participant_id
```
leikkisa: There's a `belongs_to :appeal` which would allow you to leave out `decision_document` in these calls.

app/models/board_grant_effectuation.rb
```diff
@@ -135,14 +135,34 @@ def build_end_product_establishment
...
+  end
+
+  def claimant_payee_code
+    claimant = decision_document&.appeal&.claimants&.first
```
leikkisa: Just an FYI, I am planning on adding a `claimant` method to decision reviews in my other PR, but I can update this in that PR

spec/models/board_grant_effectuation_spec.rb
```diff
@@ -129,6 +136,97 @@
       )
     end
 
+    it "is created with the correct claimant participant id and payee code" do
```
leikkisa: Nice tests!

app/models/board_grant_effectuation.rb
```diff
@@ -135,14 +135,34 @@ def build_end_product_establishment
...
+    veteran = decision_document&.appeal&.veteran
+
+    payee_code = if claimant&.payee_code.present?
+                   claimant.payee_code
+                 elsif decision_document&.appeal&.veteran_is_not_claimant
+                   veteran&.relationships&.find do |relationship|
+                     relationship.participant_id == claimant.participant_id
+                   end&.default_payee_code
+                 end
```
nanotone: I'm still learning Ruby so this isn't a blocker, but does this feel stylistically unwieldy? The heavy indenting (here due to the find-block inside the if/elsif inside the assignment) is normally a style smell for me.

pkarman: indented if/else usually indicates a new method is needed.

leikkisa: If the claimant model had access to the relationship_type, then the default_payee_code could be a concern that's shared by relationship and claimant.

app/models/board_grant_effectuation.rb
tomas-nava: fixed in [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/12675/commits/655596da072b8a3585076f3ba3a0bb90a9ebe374).

app/models/board_grant_effectuation.rb
tomas-nava: fixed in [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/12675/commits/655596da072b8a3585076f3ba3a0bb90a9ebe374).

app/models/board_grant_effectuation.rb
tomas-nava: I made this more concise with [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/12675/commits/4cf1e4c914ced4d63ab0a6db3e1e92810a8d1029).

I looked into using a payee_code concern on relationships and claimants, but it's complex enough that I think it's out of scope for this PR: the claimant model _does_ have access to the `relationship_type` (called `relationship` there), but it also needs to know the `gender` and have access to the veteran to properly derive a default_payee_code.

app/models/board_grant_effectuation.rb
tomas-nava: fixed in [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/12675/commits/655596da072b8a3585076f3ba3a0bb90a9ebe374).

app/models/veteran.rb
```diff
@@ -377,11 +383,10 @@ def fetch_end_products
...
   def fetch_relationships
-    relationships = bgs.find_all_relationships(
-      participant_id: participant_id
-    )
-    relationships_array = Array.wrap(relationships)
-    relationships_array.map { |relationship_hash| Relationship.from_bgs_hash(self, relationship_hash) }
+    relationship_hashes = Array.wrap(bgs.find_all_relationships(participant_id: participant_id))
+    relationship_hashes.map do |relationship_hash|
+      Relationship.from_bgs_hash(self, relationship_hash)
+    end
```
tomas-nava: Just a refactor, no functionality changes.

PR #12679, opened 2019-11-12
v3 api - update spec - receive contact info when intaking HLRs

edmangimelli:
**no code changes**

_the easiest way to view this new version of the spec:_
paste the contents of
https://raw.githubusercontent.com/department-of-veterans-affairs/caseflow/cd470c763d27558fe968a789e78383c319cde426/app/controllers/api/docs/v3/decision_reviews.yaml

into http://editor.swagger.io/

example json:
```
{
  "data": {
    "type": "HigherLevelReview",
    "attributes": {
      "receiptDate": "2019-07-10",
      "informalConference": true,
      "sameOffice": false,
      "legacyOptInApproved": true,
      "benefitType": "compensation",
      "veteran": {
        "fileNumberOrSsn": "123456789",
        "addressLine1": "123 Street St",
        "addressLine2": "Apt 4",
        "city": "Chicago",
        "stateProvinceCode": "IL",
        "zipPostalCode": "60652",
        "phoneNumber": "4432924565",
        "emailAddress": "someone@example.com"
      },
      "claimant": {
        "participantId": "44444444",
        "payeeCode": "10"
      }
    },
    "included": [
      {
        "type": "RequestIssue",
        "attributes": {
          "decisionIssueId": 205
        }
      }
    ]
  }
}
```

resolves https://github.com/department-of-veterans-affairs/caseflow/issues/12705

>[Form 20-0996](https://www.va.gov/decision-reviews/forms/higher-level-review-20-0996.pdf) has a section for updating a veteran's contact info. Currently, our only decision review creation endpoint, _create a higher_level_review_, does not support this. 

I considered a few different sources when designing this:
[Form 20-0996](https://www.va.gov/decision-reviews/forms/higher-level-review-20-0996.pdf)
[The Benefits API](https://developer.va.gov/explore/benefits/docs/claims)
[Vet360 endpoints](https://app.swaggerhub.com/apis-docs/annaswims/va-gov_api/1.0.0#/profile/postVet360EmailAddress) (va.gov API /v0/profile)
and the [Address Validation API](https://developer.va.gov/explore/verification/docs/address_validation)

At this moment, we're going to receive this contact info solely for generating a _filled-out_ pdf of 20-0996, but, in the future, we'll use it to update the veteran's address in vet360.
My original design leaned very closely to how va.gov's profile endpoints / the benefits api works. In discussing this with @erluti, I'm erring on the side of replicating 20-0996. It's nearly field for field similar to the paper form, but I did decide to add a phoneNumberExt field in accordance with how vet360 takes in phone numbers. My design is a compromise --the goal was to keep it simple and not too nested.

*NOTE:*
I'm also introducing another change.
To be more [JSON:API-y](https://jsonapi.org/format/) I've moved `veteran` and `claimant` out of `relationships` and into `attributes`. 1) there's currently no plan and we don't foresee a plan to create endpoints that would allow one to manipulate veteran or claimant records --using the `relationships` field seem misleading in this respect. 2) with the contact info changes, this endpoint will be performing multiple actions with one request (creating a HLR, and updating vet360 with new contact info), something JSON:API is not well-suited to do. Doing multiple things at once is already a perversion of JSON:API, and doing it within `relationships` seemed even worse.

### DIFF
--- a/app/controllers/api/docs/v3/decision_reviews.yaml
+++ b/app/controllers/api/docs/v3/decision_reviews.yaml
@@ -162,16 +162,25 @@ paths:
                               type: string
                               pattern: '^[0-9]+$'
                               description: >-
-                                Update a veteran's phone number.
-                                Example: "18446982311".
-                                Format: Include only digit characters. Include both country code and area code.
+                                Update a veteran's phone number (include area code).
+                                Example: "8446982311".
+                                Format: Include only digit characters. Include area code.
                                 Max length: 14 characters.
                                 Corresponds to "10. TELEPHONE NUMBER" on form 20-0996.
+
+                            phoneNumberCountryCode:
+                              type: string
+                              pattern: '^[0-9]+$'
+                              description: >-
+                                Update a veteran's phone number country code.
+                                Note: If not specified, will assume "1".
+                                Example: "20".
+                                Format: Include only digit characters.
 
                             phoneNumberExt:
                               type: string
                               description: >-
-                                Update a veteran's phone number extension.
+                                Update a veteran's phone number extension (if needed).
                                 Example: "123".
                                 Max length: 10 characters.
 
@@ -243,7 +252,7 @@ paths:
                         city: Chicago
                         stateProvinceCode: IL
                         zipPostalCode: "60652"
-                        phoneNumber: "14432924565"
+                        phoneNumber: "4432924565"
                         emailAddress: someone@example.com
                       claimant:
                         participantId: "44444444"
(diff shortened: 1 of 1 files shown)


app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -80,52 +81,129 @@ paths:
...
+                                Max length: 10 characters.
+
+                            emailAddress:
+                              type: string
+                              pattern: '^.+@.+\..+$'
+                              description: >-
+                                Update a veteran's email address.
+                                Example: "linda@example.com".
+                                Max length: 255 characters.
+                                Regex: /^.+@.+\..+$/i
```
erluti: Is the regex necessary if `pattern` is being used?

edmangimelli: Probably overkill. I'll clean it up in next PR

PR #12684, opened 2019-11-12
Always enable database_cleaner

pkarman:
Reverts part of 7e95cc5faeb08e9484b38bc3b0a1b5ae2002697a to proactively address flakey tests.


pkarman:
Average Rspec run time on each CI node is about `8:15`

PR #12687, opened 2019-11-12
Show POA address on Case Details Page

### LINKED ISSUE #12102: Add POA email address to hearings details
### User Story:
As a hearings branch member, I need to be able to schedule a virtual hearing.  This requires Caseflow to have the POA email.   Add the POA email to the information that is displayed in the CaseDetails page, similar to how we have it for the Veteran.  In addition, the hearings team needs to be able to update the POA email address just for use with this hearing.

### Designs:
Display-only POA email address on Case Details:

### Acceptance criteria: 
The email is visible if available.
### Feature Toggle: 
no
### Technical notes:  
Please look in the person record in BGS under :email_addr=>

ferristseng:
Resolves #12102 

### Description

  - Fetch POA email address from BGS person
  - Display POA email address on Case Details page

### Screenshots

Screen Shot 2019-11-12 at 2 58 58 PM

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -37,7 +37,6 @@ class Person < ApplicationRecord
     FullName.new(first_name, "", last_name).formatted(:readable_short)
   end
 
-  # TODO: Should this be cached?
   def email_address
     bgs_person[:email_address]
   end
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -37,6 +37,11 @@ class Person < ApplicationRecord
     FullName.new(first_name, "", last_name).formatted(:readable_short)
   end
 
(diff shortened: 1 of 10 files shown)


app/models/person.rb
```diff
@@ -37,6 +37,10 @@ def name
     FullName.new(first_name, "", last_name).formatted(:readable_short)
   end
 
+  def email_address
```
ferristseng: Should this be cached? Someone's email probably changes more frequently than their name, but I still think it wouldn't be that often. Right now, the only place this would trigger a call to BGS is in the POA route. It shouldn't be super difficult to add caching in the future if it does end up slowing things down.

Issue #12519, opened 2019-10-30
Refactor Hearing Details to Only Send Edited Data


lomaxap:
Currently, the hearing details sends all form data regardless of whether the data has changed. https://github.com/department-of-veterans-affairs/caseflow/blob/20ae9aa3ea0cb77a10b7608db2f481c5dc6aa224/client/app/hearings/components/Details.jsx#L121-L127

#12488 is written to expect only data that has changed in the params. 

AC
- [ ] Only params that have been changed should be sent to hearings controller

PR #12696, opened 2019-11-13
Rubaiyat/integrate caseflow with virtual hearings

### LINKED ISSUE #12334: Integrate Virtual Hearing Modal with Controller
### Background
#12333 created a modal that appears after a user switches from central to virtual hearings and vice versa, we now need to send `virtual_hearings` params to the HearingsController (#12332) 

### Objective
There will be 2 requests when virtual hearings data changes. One to the virtual_hearing endpoint and another to the hearings endpoint. These requests should be chained in that order. If the virtual hearing request fails, the user is presented with a failure and no hearings data will be saved. If the hearings request fails, it's OK that the virtual hearing data was saved, we simply want the error message to accurately reflect what failed. 

- [ ] Set value of hearing type dropdown based on `hearing` data
- [ ] Add method that posts virtual_hearing params when a hearing is changed from anything to `virtual`

params:
```
{
  virtual_hearing: {
     hearing_id: 123,
     hearing_type: "",
     veteran_email: "",
     judge_email: "",
     representative_email: "",
     active: true
  }
}
```
- [ ] Add method that sends `{ virtual_hearing: { active: false } }` to endpoint when a hearing is changed from `virtual` to `central` or `video`

rubaiyat22:
Resolves #12334
Relates to #12519
### Description
- integrate changes on hearing details page with backend, sends patch request to hearings_controller
- integrate changes on daily docket row with backed, sends patch request to hearings_controller
- modify tests on hearing_prep_spec.rb and hearing_details_spec.rb

### DIFF
--- a/app/serializers/hearings/hearing_serializer.rb
+++ b/app/serializers/hearings/hearing_serializer.rb
@@ -71,6 +71,11 @@ class HearingSerializer
   attribute :veteran_gender, if: for_full
   attribute :veteran_last_name
   attribute :virtual, &:virtual?
+  attribute :virtual_hearing do |object|
+    if object.virtual?
+      VirtualHearingSerializer.new(object.virtual_hearing).serializable_hash[:data][:attributes]
+    end
+  end
   attribute :witness
   attribute :worksheet_issues
 end
--- a/app/serializers/hearings/hearing_serializer.rb
+++ b/app/serializers/hearings/hearing_serializer.rb
@@ -70,7 +70,7 @@ class HearingSerializer
   attribute :veteran_first_name
   attribute :veteran_gender, if: for_full
   attribute :veteran_last_name
-  attribute :virtual, &:virtual?
+  attribute :is_virtual, &:virtual?
   attribute :virtual_hearing do |object|
     if object.virtual?
       VirtualHearingSerializer.new(object.virtual_hearing).serializable_hash[:data][:attributes]
--- a/client/app/hearings/components/dailyDocket/DailyDocketRow.jsx
+++ b/client/app/hearings/components/dailyDocket/DailyDocketRow.jsx
@@ -51,7 +51,7 @@ const inputSpacing = css({
   }
 });
 
-class HearingActions extends React.Component {
+class DailyDocketRow extends React.Component {
   constructor(props) {
     super(props);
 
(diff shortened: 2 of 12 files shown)


client/app/hearings/containers/DailyDocketContainer.jsx
```diff
@@ -98,6 +98,18 @@ export class DailyDocketContainer extends React.Component {
     });
   };
 
+  formatVirtualHearing = (hearing) => {
```
rubaiyat22: this function is not working as expected but left it as a placeholder for future changes

app/serializers/hearings/hearing_serializer.rb
```diff
@@ -71,6 +71,11 @@ class HearingSerializer
   attribute :veteran_gender, if: for_full
   attribute :veteran_last_name
   attribute :virtual, &:virtual?
```
lomaxap: i think we need to change this to is_virtual

client/app/hearings/components/dailyDocket/DailyDocketRow.jsx
```diff
@@ -256,14 +267,15 @@ class HearingActions extends React.Component {
         {this.getRightColumn()}
       </div>
       {/* This is where the modal for the edit virtual hearing modal is supposed to appear*/}
-      {(user.userCanScheduleVirtualHearings && this.state.virtualHearingModalActive) &&
+      {(user.userCanScheduleVirtualHearings && this.state.virtualHearingModalActive && hearing.isVirtual) &&
```
lomaxap: i'm surprised this works. Does `hearing.isVirtual` return anything but `undefined`?

client/app/hearings/containers/DailyDocketContainer.jsx
lomaxap: will be covered with #12519

client/app/hearings/containers/DailyDocketContainer.jsx
```diff
@@ -98,6 +98,18 @@ export class DailyDocketContainer extends React.Component {
...
+    if (!virtualHearing.isVirtual){
```
lomaxap: this should be `hearing.isVirtual`

PR #12699, opened 2019-11-13
Add virtual_hearings_establishment table

ferristseng:
Schema changes for #12696

### Description

We are adding the `Asyncable` concern to a virtual hearing using a related model similar to what is being done for `TaskTimer`. There's some overlap of naming and status tracking that would makes it confusing to implement the necessary columns on the `VirtualHearing` model directly.

A follow up PR will follow with the implementation of the model.

*Schema Changes Only*

* [x] Column comments updated

### DIFF
--- a/db/migrate/20191113160247_create_virtual_hearing_establishments.rb
+++ b/db/migrate/20191113160247_create_virtual_hearing_establishments.rb
@@ -1,0 +1,15 @@
+class CreateVirtualHearingEstablishments < ActiveRecord::Migration[5.1]
+  def change
+    create_table :virtual_hearing_establishments do |t|
+      t.belongs_to :virtual_hearing, null: false, comment: "Virtual Hearing the conference is being established for."
+      t.datetime :last_submitted_at, comment: "Async timestamp for most recent job start."
+      t.datetime :submitted_at, comment: "Async timestamp for initial job start."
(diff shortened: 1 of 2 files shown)


db/migrate/20191113160247_create_virtual_hearing_establishments.rb
```diff
@@ -0,0 +1,15 @@
...
+  def change
+    create_table :virtual_hearing_establishments do |t|
+      t.belongs_to :virtual_hearing, null: false, comment: "Virtual Hearing the conference is being established for."
+      t.datetime :last_submitted_at, comment: "Async timestamp for most recent job start."
+      t.datetime :submitted_at, comment: "Async timestamp for initial job start."
+      t.datetime :attempted_at, comment: "Async timestamp for most recent attempt to run."
+      t.datetime :processed_at, comment: "Timestamp for when the virtual hearing was successfully processed."
+      t.datetime :canceled_at, comment: "Timestamp when job was abandoned."
+      t.string :error, comment: "Async any error message from most recent failed attempt to run."
+      t.datetime "created_at", null: false, comment: "Automatic timestamp when row was created."
```
pkarman: fyi I think `t.timestamps, null: false, comment: "Automatic timestamps for row creation/update"` could be used instead. not a big deal either way.

PR #12702, opened 2019-11-13
Add Asyncable to Virtual Hearings Establishment

### LINKED ISSUE #12685: Use asyncable concern for virtual hearings

Virtual hearings reaches out to external job to 
1.) Send emails
2.) Connect the Virtual hearings solution (pexip)
The caseflow team has developed   [asyncable concern](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/concerns/asyncable.rb) to allow external dependencies that are not reliable to be retried over an extended period of time.  Investigate using them in virtual hearings.

## Technical notes

Related to #12337

ferristseng:
Resolves #12685 

### Description

  * Adds `Asyncable` concern to `VirtualHearingEstablishment`
  * Updates form to create/restart `VirtualHearingEstablishment`
  * Updates create conference job to modify `VirtualHearingEstablishment`

### Screenshots

Screen Shot 2019-11-14 at 3 25 58 PM

Screen Shot 2019-11-14 at 3 26 19 PM

### DIFF
--- a/app/jobs/virtual_hearings/create_conference_job.rb
+++ b/app/jobs/virtual_hearings/create_conference_job.rb
@@ -16,27 +16,39 @@ class VirtualHearings::CreateConferenceJob < ApplicationJob
   def perform(hearing_id:)
     @virtual_hearing = VirtualHearing.find_by(hearing_id: hearing_id)
 
-    if !virtual_hearing.conference_id && !virtual_hearing.alias
-      assign_virtual_hearing_alias
+    virtual_hearing.establishment.attempted!
 
-      resp = client.create_conference(
-        host_pin: rand(1000..9999),
-        guest_pin: rand(1000..9999),
-        name: virtual_hearing.alias
-      )
-
-      if resp.error
-        Rails.logger.info "CreateConferenceJob failed: #{resp.error.message}"
-        fail resp.error
-      end
-
-      virtual_hearing.update(conference_id: resp.data[:conference_id], status: :active)
-    end
+    create_conference if !virtual_hearing.conference_id && !virtual_hearing.alias
 
     VirtualHearings::SendEmail.new(virtual_hearing: virtual_hearing, type: :confirmation).call
+
+    if virtual_hearing.active? && virtual_hearing.all_emails_sent?
+      virtual_hearing.establishment.clear_error!
+      virtual_hearing.establishment.processed!
+    end
   end
 
   private
+
+  def create_conference
+    assign_virtual_hearing_alias
+
+    resp = client.create_conference(
+      host_pin: rand(1000..9999),
+      guest_pin: rand(1000..9999),
+      name: virtual_hearing.alias
+    )
+
+    if resp.error
+      Rails.logger.warn "CreateConferenceJob failed: #{resp.error.message}"
+
+      virtual_hearing.establishment.update_error!(resp.error.message)
+
+      fail resp.error
+    end
+
+    virtual_hearing.update(conference_id: resp.data[:conference_id], status: :active)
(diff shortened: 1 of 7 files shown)


app/models/hearings/virtual_hearing_establishment.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  # :nocov:
+  # Implements Asyncable
+  def veteran
+    virtual_hearing.hearing.appeal&.veteran
+  end
+
+  # Implements Asyncable
+  def asyncable_user
+    virtual_hearing.created_by.css_id
```
nanotone: Incidentally, I've been touching this code a fair amount, and in #12707 did a tiny refactor to have `#asyncable_user` return the full User object, which to me is more useful and less surprising.

The automerge will be silently broken, but I'll gladly be the steward here for whichever PR merges later -- it doesn't look like any other part of this PR uses it.

app/jobs/virtual_hearings/create_conference_job.rb
```diff
@@ -16,28 +16,40 @@ class VirtualHearings::CreateConferenceJob < ApplicationJob
   def perform(hearing_id:)
     @virtual_hearing = VirtualHearing.find_by(hearing_id: hearing_id)
 
-    if !virtual_hearing.conference_id && !virtual_hearing.alias
-      assign_virtual_hearing_alias
+    virtual_hearing.establishment.attempted!
```
rubaiyat22: Should attempted timestamp be set even if the job was never going to kick off (i.e perform_now is only called to send email updates)?

ferristseng: What I was thinking is that creating the conference and sending all the initial emails are all involved with establishing the virtual hearing. The class does seem a bit overloaded because it also is handling the status of the job for when any of the emails associated with the virtual hearings change.

ferristseng: It seems like the asyncable class is tied to the job though, so from the code it makes sense IMO

rubaiyat22: Yeah it makes sense to me as well. I was just wondering if there was a way to indicate which part of the job was in progress/completed (i.e creating conference and sending email, just sending email), not sure if it's really necessary

ferristseng: Yeah, I think it would be nice if the UI could display more granularity. The virtual hearing has that info, but we would need some way to expose it to the frontend

rubaiyat22: One way we could display what is in progress is many by adding a new row `status` or something similar saying something like "Creating Conference" or "sending emails" because those two things are done one after another. If a conference is being created, status could be "Creating conference". "If a conference is created successfully, status would be set to "Sending emails." We could add a method to `VirtualHearingEstablishment`similar to `veteran` which sets the status according to the values in `VirtualHearing` table (i.e if `active == false`=>`status="creating conference"`, if `active == true &&  !all_emails_sent` => `status = "sending emails"`)

ferristseng: Yeah, we could probably override `asyncable_status` for that functionality. Unfortunately that doesn't get passed to the frontend (`asyncable_ui_hash`), but seems like it could be a good candidate to put that logic in. Maybe we can create a follow up ticket.

app/models/hearings/virtual_hearing_establishment.rb
ferristseng: FYI merging this!

PR #12703, opened 2019-11-13
Disallow judge removal from their own teams

### LINKED ISSUE #12492: Prevent judges from being removed from their JudgeTeams

hschallhorn:
Resolves #12492 

### Description
Displays an error when a user tries to remove a judge from their team

### Acceptance Criteria
- [ ] Admins cannot remove a judge from their judge team

### Testing Plan
1. Sign in a TEAM ADMIN
1. Go to /organizations/bvaaabshire/users
1. Press "Remove user" next to "Judge with hearings and cases (BVAAABSHIRE)"
1. Ensure an error is displayed

### User Facing Changes
Screen Shot 2019-11-13 at 3 52 37 PM

 ### Notes
This can be circumvented by removing admin rights from the judge and then removing them as a user

### DIFF
--- a/app/controllers/organizations/users_controller.rb
+++ b/app/controllers/organizations/users_controller.rb
@@ -36,6 +36,11 @@ class Organizations::UsersController < OrganizationsController
   end
 
   def destroy
+    pp organization
+    pp organization.judge
+    pp user_to_modify
+    pp organization.is_a?(JudgeTeam)
+    pp organization.judge.eql?(user_to_modify)
     if organization.is_a?(JudgeTeam) && organization.judge.eql?(user_to_modify)
       fail Caseflow::Error::ActionForbiddenError, message: COPY::JUDGE_TEAM_REMOVE_JUDGE_ERROR
     end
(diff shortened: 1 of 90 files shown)


app/controllers/organizations/users_controller.rb
```diff
@@ -36,6 +36,10 @@ def update
   end
 
   def destroy
+    if organization.is_a?(JudgeTeam) && organization.judge.eql?(user_to_modify)
+      fail Caseflow::Error::ActionForbiddenError, message: COPY::JUDGE_TEAM_REMOVE_JUDGE_ERROR
+    end
```
kevmo: I would move these lines of code into their own function that is then either called in this controller (`fail_if_judge_team_judge_admin`) or in an ÒrganizationUser method. It will keep the controller more maintainable.

Issue #12215, opened 2019-09-25
Legacy Cases being moved back to Location 55


ferristseng:
https://dsva.slack.com/archives/CHX8FMP28/p1569353446139100

Investigate this further. @lomaxap suggested it could be caused by the `hearing_disposition_change_job`.

I noticed that one of the tasks seemed like it was modified by the job because the `updated_at` timestamp was a few minutes after the job ran:

```
AssignHearingDispositionTask updated_at

  2019-08-15 05:01:45 UTC


HearingDispositionChangeJob ran at 1:08 AM EST on 08-15

HearingDispositionChangeJob completed after running for 7 minutes. Processed 17 No show hearings. Processed 89 Held hearings. Processed 16 Cancelled hearings. Processed 11 Postponed hearings. Processed 3 Between one and two days old hearings. Processed 1 Stale hearings. Processed 0 Unknown disposition hearings. Encountered errors for 2 hearings.

https://dsva.slack.com/archives/C4JECDLSE/p1565845684086500 


Hearing took place on

  2019-08-13 16:30:00 UTC
```

Every `AssignHearingDispositionTask` for the appeals mentioned in the Slack message has a `cancelled` status:

```
irb(main):020:0> appeals.map { |a| a.tasks }.flatten.select { |t| t.is_a? AssignHearingDispositionTask }.map { |t| t.status }
=> ["cancelled", "cancelled", "cancelled", "cancelled", "cancelled", "cancelled", "cancelled", "cancelled", "cancelled", "cancelled", "cancelled", "cancelled", "cancelled"]
```

I'm thinking that this line:

https://github.com/department-of-veterans-affairs/caseflow/blob/17583e6fc78bdfc29494a748b99b8275bc5228a2/app/models/task.rb#L32

ends up executing this:

https://github.com/department-of-veterans-affairs/caseflow/blob/17583e6fc78bdfc29494a748b99b8275bc5228a2/app/models/tasks/hearing_task.rb#L49-L59

ferristseng:
The appeals also are all have a disposition of "Dismissed, Withdrawn" and status of "Complete":

```
[...]

PR #12706, opened 2019-11-13
Fail if attempting to close hearing task not in caseflow

lomaxap:
related to #12215: Cases that mysteriously move locations after the case is resolve. Maybe as a result of duplicate hearing tasks or maybe because a location is moved manually in VACOLS with the expectation that it will no longer be worked in Caseflow. 

Since we only move locations in hearings on the Hearing Task, fail when we attempt to move an appeal that is not currently in Caseflow

### DIFF
--- a/app/models/tasks/hearing_task.rb
+++ b/app/models/tasks/hearing_task.rb
@@ -10,6 +10,8 @@ class HearingTask < Task
   delegate :hearing, to: :hearing_task_association, allow_nil: true
   before_validation :set_assignee
 
+  class HearingTaskNotCompletable < StandardError; end
+
   def self.label
     "All hearing-related tasks"
   end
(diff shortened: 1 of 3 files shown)


app/models/tasks/hearing_task.rb
```diff
@@ -34,10 +36,20 @@ def verify_org_task_unique
     true
   end
 
+  def hearing_task_ready_for_completion?
+    return false if !appeal.tasks.open.where(type: HearingTask.name).empty?
```
ferristseng: Small nit: I think you could use `exists?` here instead

tomas-nava: Although `exists?` is valid, I actually think this is more readable, and consistent with the way we've done similar checks elsewhere in the code:
```ruby
return false if appeal.tasks.open.where(type: HearingTask.name).any?
```

pkarman:
@ferristseng tagged you to shepherd this one.

ferristseng:
Planning on closing this because it seems to break the flow for changing the disposition for a hearing.

This test reveals the issue:

The problem occurs because canceling a hearing task will update the location of a case to "case storage" (VACOLS location code 81). When a change hearing disposition task is created, the location isn't kicked back to `CASEFLOW`, so a subsequent update to the disposition causes an error because the VACOLS location is no longer `CASEFLOW`, which is behavior that this PR introduces.

We could potentially update change hearing disposition to kick the appeal back to `CASEFLOW`. In either circumstance, it seems like we aren't respecting the `CASEFLOW` location.

@tomas-nava Do you think it's worth trying to make that change, or just close? I think you might have more context on how this should work.

Issue #12524, opened 2019-10-30
Async job notes trigger a message to the user


leikkisa:
### User or job story
As a user, when my asynchronous job gets a note added, I'd like to get a message in my inbox.

### Acceptance criteria
- When a job note is added to an async job with the "send_to_intake_user" flag set to true, the user connected to the job gets a message delivered to their inbox with the note
- The user does not get a message if they're the one that added the note
- Include a screenshot of the inbox message a user gets when a job note got added

### Release notes
Users will receive a message in their inbox when a note gets added to one of their asynchronous jobs by another user, such as the engineering team.

#### Designs
N/A

#### Background/context
This is breaking off a piece of the scope for issue: 
https://github.com/department-of-veterans-affairs/caseflow/issues/12214

and is a follow on to this PR
https://github.com/department-of-veterans-affairs/caseflow/pull/12252

#### Technical notes
Please see the inbox app to see how messages are currently being sent.

PR #12707, opened 2019-11-13
Enable sending inbox messages during job note creation

nanotone:
Connects #12524 

### Description
This enables the "Send as message to user" checkbox in the job note creation form. New messages will end up in the Caseflow Inbox of the job's asyncable user (its current "owner").

### Acceptance Criteria
- [ ] Code compiles correctly
- [ ] Adding a note with `send_to_intake_user: true` sends a message to the user's inbox
- [ ] Adding a note with `send_to_intake_user: false` doesn't send any message

### Testing Plan
1. Go to `/jobs` with at least one async job created.
2. Click into a job.
3. Write a job note, make sure the checkbox is selected, and click "Add Note"
4. As the owner of that job (possibly a different user), go to `/inbox`.
5. Verify that a new inbox message was created from that job note.
6. Click on the link in the message to verify that it goes back to the job note.

### User Facing Changes

This checkbox, originally made by @pkarman, has been uncommented back in:
Screen Shot 2019-11-13 at 17 14 20

Caseflow Inbox has already been built, but this is the first time it will be populated with actual messages. A job-note message would look like:
Screen Shot 2019-11-17 at 12 11 56
 
### Implementation Notes

- The Message object has a polymorphic detail field, currently unused. For job note messages, I'm passing the job note itself, which feels like the most specific and informative option, should we need it one day.
- Changed `Asyncable#asyncable_user` to return a User object rather just the `css_id`, which is more surprising and less useful in most situations.
- Changed the checkbox to be disabled only during `saveInProgress`, not when the text field is empty. An empty text field should only disable the "Add Note" button.


app/controllers/asyncable_jobs_controller.rb
```diff
@@ -41,6 +41,13 @@ def add_note
       note: allowed_params[:note],
       send_to_intake_user: allowed_params[:send_to_intake_user]
     )
+    if ActiveRecord::Type::Boolean.new.deserialize(allowed_params[:send_to_intake_user]) && job.asyncable_user
```
pkarman: seems like either the job.create and the message.create should be wrapped in a single transaction, and/or the optional message creation should be moved to the `JobNote` model or a new service class that takes in the `allowed_params` and operates on them.

i.e. feels like the controller is starting to get too fat.

app/controllers/asyncable_jobs_controller.rb
```diff
@@ -41,6 +41,13 @@ def add_note
...
+      Message.create!(
+        detail: job_note,
+        text: allowed_params[:note],
```
pkarman: this was not specified in the ticket AC, but I am not sure we want to use the note text as the message body. I was imagining that the message would be something that pointed to the job note. e.g.

> A new note has been added to your {job_type} job. [Click here](link to job details page) to view the note.

My assumption is that the job note text will have little to no meaning in the context of the inbox itself and will just confuse the user.

This assumption is why I created the `Message.detail` association, to allow inbox messages to link to other objects.

That all said, those were just my (unwritten) assumptions and have zero user research behind them.

nanotone:
@leikkisa I pushed a few commits to address your smaller concerns about testing. Do you think checking `intake.user.messages` in the add-note feature test is sufficient, or should it also switch user and nav to the inbox? That feels like a pretty big multi-feature test, which I'm hesitant to do, but maybe some feature tests already do this?

nanotone:
@pkarman Thanks for all the comments! Agree with everything you wrote, I'm just gradually getting a sense for thin controllers.

The suggestion about message text is good and I think makes sense in this situation of somebody adding an arbitrarily long note on the job. For the related job-note/inbox work, I think we could be smarter about surfacing more info in the message, so that the user can glance at their inbox and take action without needing to click on links.

I'm not totally clear on when we'd need the polymorphic `detail` object if we're crafting links into the message text itself. Do we also have a Linkable concern (or, not sure what Rails convention here is)?

pkarman:
> For the related job-note/inbox work, I think we could be smarter about surfacing more info in the message, so that the user can glance at their inbox and take action without needing to click on links.

Since job notes are ad hoc human-authored, we'd need to be smarter as a matter of protocol, not code.

My assumption is that the action to be taken is to visit the Job Details page and take a screen shot. That's the CA protocol AMO has agreed to for stuck async jobs, which right now is the primary use case for Job Notes. So providing that link in the message *is* the action afaik. I.e. "show me the thing that this inbox message refers to."

Maybe it's a both/and? Include the job note text *and* a link to the job detail?

> I'm not totally clear on when we'd need the polymorphic detail object if we're crafting links into the message text itself. Do we also have a Linkable concern (or, not sure what Rails convention here is)?

The `.detail` is to allow for a Job Note now, but I believe Hearings will have some use for Inbox messaging in the near future, so I made it polymorphic. This is mostly for data integrity and forward compatibility for inbox.

What I'm imagining is saving the message text in Markdown format, and then displaying in inbox via ReactMarkdown like we do for JobNote. We don't have to do that in this PR. Let's stay with what you've got for now, and we can re-visit in a follow on PR if needed.

nanotone:
As per @pkarman's suggestion, the bulk of the message-creating logic has been moved into a new service class, `AsyncableJobMessaging`.

I chose the service class route, and this name in particular, with an eye toward the other events in a job's lifecycle that might trigger messaging: failure after 24 hours, success after 24 hours, and manual cancellation -- it didn't feel right to burden the `JobNote` model with all this non-jobnote logic.

The content of the inbox message has also been updated to just be a link back to the job note, for now. The intention is to have `AsyncableJobMessaging` be the arbiter of message text, which could change to include a job note preview (depending on CA protocol).

One more thing that could go into this PR is a feature test that switches user, visits the inbox, and clicks on the link to verify its destination, but that feels like it shouldn't live in the async jobs feature spec. Happy to receive some guidance on this.

pkarman:
> One more thing that could go into this PR is a feature test that switches user, visits the inbox,

That could go in a different PR. Agreed it should be a feature test of the Inbox, not the JobNote per se.

app/services/asyncable_job_messaging.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+# AsyncableJobMessaging handles the decision-making around how and whether events in the
```
pkarman: I like this

Issue #12711, opened 2019-11-14
Request Issue reporting


pkarman:
## Description
In order to know whether #11840 solutions are improving the rate of unidentified issues (by decreasing them), we need a week-by-week breakdown of request issues by type: rating, nonrating and unidentified.

We are looking for trends that we can corollate to the various changes we have made over the last 3 months.

pkarman:
```csv
week_of,rating,nonrating,unidentified,unidentified_percent
2019-02-25,745,13,9,1.17
2019-03-04,1707,55,18,1.01
2019-03-11,2035,33,66,3.09
2019-03-18,3045,97,43,1.35
2019-03-25,4261,160,83,1.84
2019-04-01,3682,186,62,1.58
2019-04-08,3850,220,58,1.41
2019-04-15,3637,222,68,1.73
2019-04-22,5087,249,56,1.04
2019-04-29,4685,216,58,1.17
2019-05-06,6732,414,86,1.19
2019-05-13,9211,395,88,0.91
2019-05-20,10253,513,95,0.87
2019-05-27,10528,518,203,1.8
2019-06-03,19699,936,497,2.35
2019-06-10,23201,1068,643,2.58
2019-06-17,21441,1021,640,2.77
2019-06-24,16236,720,290,1.68
2019-07-01,12884,390,210,1.56
2019-07-08,15090,677,313,1.95
2019-07-15,17984,671,1098,5.56
2019-07-22,17470,760,977,5.09
2019-07-29,17555,593,923,4.84
2019-08-05,20597,883,983,4.38
2019-08-12,24858,1101,996,3.7
2019-08-19,23260,1333,1010,3.94
2019-08-26,15866,1265,877,4.87
2019-09-02,13199,1027,610,4.11
2019-09-09,16346,1174,727,3.98
2019-09-16,15587,1091,820,4.69
2019-09-23,16173,1048,726,4.05
2019-09-30,15455,1090,577,3.37
2019-10-07,14009,818,487,3.18
2019-10-14,14174,802,562,3.62
2019-10-21,25809,1346,885,3.16
2019-10-28,26448,1309,899,3.14
2019-11-04,23979,1195,795,3.06
2019-11-11,13187,646,398,2.8
```

PR #12714, opened 2019-11-14
request issue report week-over-week by type

pkarman:
connects #12711 

### Description

Summarize request issues created week-over-week by type.

### DIFF
--- a/app/services/request_issue_reporter.rb
+++ b/app/services/request_issue_reporter.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
-# rr = RequestIssueReporter.new(start_date: 4.weeks.ago)
+# Reports the number of Request Issues by origin type week-over-week
+
 class RequestIssueReporter
   attr_reader :stats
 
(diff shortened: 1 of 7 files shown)


app/services/request_issue_reporter.rb
```diff
@@ -0,0 +1,64 @@
...
+      nonrating: RequestIssue.nonrating.where("created_at >= ? AND created_at < ?", start_week, end_week).count,
+      unidentified: RequestIssue.unidentified.where("created_at >= ? AND created_at < ?", start_week, end_week).count
+    }
+  end
+
+  def rating_request_issues
+    RequestIssue.where.not(
+      contested_rating_issue_reference_id: nil
+    ).or(RequestIssue.where.not(contested_rating_decision_reference_id: nil))
+  end
```
nanotone: Seems that this method could be a `RequestIssue` class method?

app/services/request_issue_reporter.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+  end
+
+  private
+
+  attr_reader :start_date, :end_date
+
+  def build
+    stats = {}
+    week_of = start_date.next_week # start the first Monday after the start_date
```
nanotone: From this PR I'm not sure how/when the reporter is used, but is this the desired behavior even if `start_date` happens to be a Monday?

app/services/request_issue_reporter.rb
pkarman: yes I started with `RequestIssue.rating` class method but that includes unidentified issues in its definition (since unidentified issues eventually become rating issues).

I tried adding a `rating_issue_or_rating_decision_issue` class method on `RequestIssue` but it wasn't clear to me if that would confuse folks who came later or not, so I kept it scoped to this class only for now.

app/services/request_issue_reporter.rb
pkarman: good catch! I will fix. I don't think it matters much for this particular use case, but it will be less surprising later.

app/services/request_issue_reporter.rb
leikkisa: I don't know if this is due to data hygiene or just how we capture contested decision issues, but there are a bunch of rating issues that don't get captured with this. On a preliminary look I get 2814 of them, RI 241374 is an example.

app/services/request_issue_reporter.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+# rr = RequestIssueReporter.new(start_date: 4.weeks.ago)
```
leikkisa: I think this got left in, but it would be good to put a comment up there

PR #12720, opened 2019-11-14
APIv3 - HigherLevelReview#show to match up with specs

### LINKED ISSUE #12719: APIv3 - Specs do not align with show response
## Description
HigherLevelReview#show is returning lower case types in the included section of the response.  Also part of the response is not included in the spec.

Discussion took place in DSVA slack here https://dsva.slack.com/archives/CKUCFQAHJ/p1573513535008500

erluti:
Resolves #12719

### Description
* The relationships for a HigherLevelReview had to be updated to ensure the types were being set in upper-Camelcase.
* Specs for included values were missing and added to reflect API response

### DIFF
--- a/app/controllers/api/docs/v3/decision_reviews.yaml
+++ b/app/controllers/api/docs/v3/decision_reviews.yaml
@@ -968,7 +968,7 @@ paths:
                                       type: string
                                       enum: ["RequestIssue"]
                                     id:
-                                      type: integer
+                                      type: string
                           decisionIssues:
                             type: object
                             properties:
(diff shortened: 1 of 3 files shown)


app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -1023,18 +1023,76 @@ paths:
...
+                                middleName:
+                                  type: string
+                                  nullable: true
+                                lastName:
+                                  type: string
+                                  nullable: true
+                                payeeCode:
+                                  allOf:
+                                    - $ref: "#/components/schemas/PayeeCode"
+                                    - nullable: true
```
edmangimelli: nifty. i had to read up on this and play around with it. editor.swagger.com ended up suggesting exactly what you did

erluti: That's how I found out about it!  I just took a stab at it and then it suggested this.

PR #12721, opened 2019-11-14
kick off virtual conference job from base_hearing_update_form

### LINKED ISSUE #12537: Add VirtualHearingJob to Hearings Controller
Kick off job from hearing controller after a virtual hearing is created, emails are updated, or status is changed to cancelled.

rubaiyat22:
Resolves #12537

### Description
- kick off `CreateConferencJob` from `BaseHearingUpdateForm` if virtual hearing is created or if any of the email fields are updated
- kick of `DeleteConferencesJob` if virtual hearing is `cancelled`

### DIFF
--- a/app/models/hearings/forms/base_hearing_update_form.rb
+++ b/app/models/hearings/forms/base_hearing_update_form.rb
@@ -16,7 +16,14 @@ class BaseHearingUpdateForm
 
       if !virtual_hearing_attributes.blank?
         create_or_update_virtual_hearing
-        # TODO: Start the job to create the Pexip conference here?
+
+        if hearing.virtual_hearing.status == "pending"
+          VirtualHearings::CreateConferenceJob.perform_now(hearing_id: hearing.id)
+        end
+        
+        if virtual_hearing_attributes[:status] == "cancelled"
+          VirtualHearings::DeleteConferencesJob.perform_now
+        end
       end
     end
   end
--- a/app/models/hearings/forms/base_hearing_update_form.rb
+++ b/app/models/hearings/forms/base_hearing_update_form.rb
@@ -62,6 +62,9 @@ class BaseHearingUpdateForm
       updates = virtual_hearing_attributes.compact.merge(emails_sent_updates)
 
       virtual_hearing.update(updates)
+      virtual_hearing.establishment.restart!
+    else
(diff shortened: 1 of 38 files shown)


app/models/hearings/forms/base_hearing_update_form.rb
```diff
@@ -27,6 +27,15 @@ def update_hearing; end
 
   private
 
+  def start_async_job
+    if hearing.virtual_hearing.status == "pending"
+      VirtualHearings::CreateConferenceJob.perform_now(hearing_id: hearing.id)
+    elsif hearing.virtual_hearing.status == "cancelled"
+      # DeleteConferencesJob should be kicked off in a scheduled time in the future
+      VirtualHearings::DeleteConferencesJob.perform_now
```
ferristseng: I think it's probably fine to exclude this logic for now, since we'd probably want to remove it when we setup the job scheduling in the near future. The scheduling for jobs is in the appeals-lambda repo.

See:

https://github.com/department-of-veterans-affairs/appeals-lambdas/blob/master/async-jobs-trigger/serverless.yml
https://github.com/department-of-veterans-affairs/appeals-deployment/wiki/How-to-Create-a-Caseflow-cron-job

spec/controllers/hearings_controller_spec.rb
```diff
@@ -155,6 +155,15 @@
           expect(VirtualHearing.first.representative_email).to eq("new_representative_email@caseflow.gov")
         end
 
+        it "kicks off CreateConferenceJob and updates virtual_hearing table" do
+          subject
```
ferristseng: Nice!

app/models/hearings/forms/base_hearing_update_form.rb
rubaiyat22: Excluded it for now! Will work on this in upcoming PR.

spec/feature/hearings/hearing_details_spec.rb
```diff
@@ -87,7 +87,6 @@
       legacy_hearing.reload
       expect(VirtualHearing.count).to eq(1)
       expect(legacy_hearing.virtual?).to eq(true)
-      expect(legacy_hearing.virtual_hearing.status).to eq("pending")
```
rubaiyat22: Removed this specific test because i'm testing this logic in other places

PR #12726, opened 2019-11-15
Ignore non-actionable tasks when reporting last assignee

### LINKED ISSUE #12514: Ignore tasks that are not actionable when determining the assigned to location of an appeal
[Related slack thread](https://dsva.slack.com/archives/CJL810329/p1572358195003400).

When [displaying who a case is assigned to](https://github.com/department-of-veterans-affairs/caseflow/blob/20ae9aa3ea0cb77a10b7608db2f481c5dc6aa224/app/models/legacy_appeal.rb#L732) we typically show the assignee of the most [recently assigned task](https://github.com/department-of-veterans-affairs/caseflow/blob/20ae9aa3ea0cb77a10b7608db2f481c5dc6aa224/app/models/task.rb#L140). We currently have some tasks that are used to track veterans of private attorneys (`TrackVeteranTask`s) and tasks that determine if another task is on hold (`TimedHoldTask`s). Neither of these are actionable tasks and should be ignored when determining the most recently assigned tasks. Showing these task assignees as "who is in charge of the case" is jarring for users who do not expect a private attorney to ever be assigned the case, as would be displayed for a `TrackVeteranTasks`

### AC
- [ ] `TimedHoldTask`s and `TrackVeteranTask`s are ignored when determining the `assigned_to_location`

yoomlam:
Resolves #12514 

### Description
LegacyAppeal and Appeal classes ignore TimedHoldTasks and TrackVeteranTasks when determining the assigned_to_location.

### Acceptance Criteria
- [ ] TimedHoldTasks and TrackVeteranTasks are ignored when determining the assigned_to_location

### Testing Plan
1. Relevant RSpec tests pass


spec/models/legacy_appeal_spec.rb
```diff
@@ -2527,13 +2527,42 @@
...
+            expect(appeal.assigned_to_location).to eq(COPY::CASE_LIST_TABLE_CASE_STORAGE_LABEL)
+          end
+        end
+
+        describe "when there is an assigned actionable task" do
+          let(:task_assignee) { create(:user) }
+          let!(:task) { create(:ama_attorney_task, :in_progress, assigned_to: task_assignee, appeal: appeal) }
+
+          it "returns the actionable task's label and does not include nonactionable tasks in its determinations" do
+            expect(appeal.assigned_to_location).to eq(task_assignee.css_id)
```
yoomlam: Rspec failed once on this line. Might be a flakey test but I can't figure out why.

app/models/appeal.rb
```diff
@@ -80,18 +80,23 @@ def type
...
     return COPY::CASE_LIST_TABLE_POST_DECISION_LABEL if root_task&.status == Constants.TASK_STATUSES.completed
 
-    active_tasks = tasks.active.visible_in_queue_table_view
-    return most_recently_assigned_to_label(active_tasks) if active_tasks.any?
-
-    on_hold_tasks = tasks.on_hold.visible_in_queue_table_view
-    return most_recently_assigned_to_label(on_hold_tasks) if on_hold_tasks.any?
+    recently_updated_task = Appeal.any_recently_updated(
+      tasks.active.actionable.visible_in_queue_table_view,
+      tasks.on_hold.actionable.visible_in_queue_table_view
```
hschallhorn: The `active` scope includes [assigned and in_progress](https://github.com/department-of-veterans-affairs/caseflow/blob/c6521544fd8af71d0f7bae750f60d84c5fb84b0b/app/models/task.rb#L83). The `open` scope includes both of these as well as [on_hold](https://github.com/department-of-veterans-affairs/caseflow/blob/c6521544fd8af71d0f7bae750f60d84c5fb84b0b/app/models/task.rb#L87). These lines could be reduced to 
```ruby
recently_updated_task = tasks.open.actionable.visible_in_queue_table_view.most_recently_updated
```

app/models/appeal.rb
```diff
@@ -80,18 +80,23 @@ def type
...
 
+    # Should following 2 lines be removed?
     # this condition is no longer needed since we only want active or on hold tasks
-    return most_recently_assigned_to_label(tasks) if tasks.any?
+    return tasks.actionable.most_recently_updated&.assigned_to_label if tasks.any?
 
     fetch_status.to_s.titleize
   end
 
+  def self.any_recently_updated(*tasks_arrays)
```
hschallhorn: This could be removed with the above suggested changes

app/models/legacy_appeal.rb
```diff
@@ -740,13 +740,17 @@ def external_id
 
   def assigned_to_location
     return location_code unless location_code_is_caseflow?
-    return active_tasks.most_recently_assigned.assigned_to_label if active_tasks.any?
-    return on_hold_tasks.most_recently_assigned.assigned_to_label if on_hold_tasks.any?
+
+    recently_updated_task = Appeal.any_recently_updated(
```
hschallhorn: Same suggestion as the above

app/models/task.rb
```diff
@@ -62,6 +62,12 @@ class << self; undef_method :open; end
     )
   }
 
+  scope :not_timed_hold_task, -> { where.not(type: :TimedHoldTask) }
+
+  scope :not_track_veteran_task, -> { where.not(type: :TrackVeteranTask) }
+
+  scope :actionable, -> { not_timed_hold_task.not_track_veteran_task }
```
hschallhorn: I don't think we need these scopes.  These two classes are already excluded by `visible_in_queue_table_view`

spec/models/legacy_appeal_spec.rb
hschallhorn: CI or local rspec?

hschallhorn: This isn't the reason this failed (most likely not), but `:ama_attorney_task` is used for `Appeal`s. `Legacy_Appeals` don't have attorney tasks. May want to use `:colocated_task` instead

app/models/appeal.rb
yoomlam: If an on_hold task is found before an active task do we want to show assignee of the active task?

app/models/task.rb
yoomlam: Is the concept of an 'actionable' task interlinked with (or the same as) the 'visible_in_queue_table_view' concept?

hschallhorn: 💯

app/models/appeal.rb
hschallhorn: You're absolutely right. Had a mild brain fart here. Ignore the previous comment!

app/models/appeal.rb
hschallhorn: So, ignore this as well. However, since this method is only dealing with tasks, I think it's better suited to reside in tasks.rb.

yoomlam: Moved to `task.rb`.

app/models/task.rb
yoomlam: Ok. Leaving as is.

spec/models/legacy_appeal_spec.rb
yoomlam: Changed to `:colocated_task`.

app/models/task.rb
hschallhorn: I still think that we should rely on  `visible_in_queue_table_view` here rather than create a new scope. We used to call out [`TrackVeteranTask` and `TimedHoldTask` individually for this scope](https://github.com/department-of-veterans-affairs/caseflow/pull/11467/files#diff-c95c9451cf097bdd58f73d0f9b114f97L51) but listing each task to hide individually got unwieldy when adding more "non actionable" tasks.

yoomlam: I'll get rid of the `actionable`, `not_track_veteran_task`, and `not_timed_hold_task` scopes, and simply use `visible_in_queue_table_view` as synonymous with `actionable`.

app/models/appeal.rb
```diff
@@ -80,14 +80,15 @@ def type
...
+    recently_updated_task = Task.any_recently_updated(
+      tasks.active.actionable.visible_in_queue_table_view,
+      tasks.on_hold.actionable.visible_in_queue_table_view
+    )
+    return recently_updated_task.assigned_to_label if recently_updated_task
 
+    # Should following 2 lines be removed?
     # this condition is no longer needed since we only want active or on hold tasks
-    return most_recently_assigned_to_label(tasks) if tasks.any?
+    return tasks.actionable.most_recently_updated&.assigned_to_label if tasks.any?
```
yoomlam: Should I remove these lines?

hschallhorn: I would say keep them. Just because we don't expect to reach that point doesn't guarantee that we won't.

PR #12727, opened 2019-11-15
Remove json_hearing_day and json_hearing_days

### LINKED ISSUE #12622: Remove json_hearing_days from HearingDayController
`json_hearing_days` adds attributes for regional office city and room label, we should remove this in favor of attributes on a serializer.

ferristseng:
Resolves #12622

### Description

I decided to do this in such a way where the structure of the data that the API returned was intact.

This led to some issues because `HearingDay#to_hash` was being used in `assign_judges_to_hearing_days.rb`. The change required updating `assign_judges_to_hearing_days.rb` to no longer depend on the behavior of `HearingDay#to_hash`. The file was tested pretty extensively, so I'm hoping the unit tests catch any regressions.

#### Changes

  * Remove `HearingDayController#json_hearing_days` and `HearingDayController#json_hearing_day`
  * Remove `HearingDay#create_hearing_day`, which chained together `HearingDay#create` and `HearingDay#to_hash`. IMO, ideally, `HearingDay#to_hash` should primarily be used as a representation of `HearingDay` to the frontend. We should avoid using a hash to represent a `HearingDay` in the backend portions of the code.
  * Update `HearingDaySerializer` to serialize `regional_office` and `room` and keep the structure of the data intact.
  * Refactor `assign_judges_to_hearing_days.rb` to use HearingDay model instead of hash
  * Remove `app/services/hearing_schedule/generate_co_hearing_days_schedule.rb`. This was an unused piece of code that called `HearingDay#create_hearing_day`
  * Update tests to expect the `HearingDay` model instead of a hash


app/services/hearing_schedule/assign_judges_to_hearing_days.rb
```diff
@@ -30,7 +30,7 @@ def match_hearing_days_to_judges
     evenly_assign_judges_to_hearing_days
     assign_remaining_hearing_days
     verify_assignments
-    @assigned_hearing_days.sort_by { |day| day[:scheduled_for] }
+    @assigned_hearing_days.sort_by(&:scheduled_for_as_date)
```
lomaxap: 👍 thanks for doing the extra step of not using the hash here

PR #12730, opened 2019-11-15
v3 api - remove RequestIssues abstraction from HLR endpoint

edmangimelli:
**no code changes**

resolves: https://github.com/department-of-veterans-affairs/caseflow/issues/12733


app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -248,34 +248,90 @@ paths:
...
-                          category:
-                            $ref: "#/components/schemas/NonratingIssueCategory"
-                          decisionDate:
-                            type: string
-                          decisionText:
+                            nullable: true
+                          ratingDecisionIssueId:
                             type: string
+                            nullable: true
+                          # category: # commented out for MVP
```
erluti: Instead of commenting this out, go ahead and make a ticket for this work and put this in there.

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -248,34 +248,90 @@ paths:
...
+                            nullable: true
+                            type: array
+                            description: LegacyAppealIssues Array
+
+                            items:
+                              type: object
+                              description: >-
+                                OPTIONAL. _LegacyAppealIssues_ that you are associating with
+                                this _ContestableIssue_. To associate _LegacyAppealIssues_,
+                                `legacyOptInApproved` must be `true`. At this time, only one
```
erluti: instead of "only one" you could say "at most one" and remove the parenthetical comment.

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -248,34 +248,90 @@ paths:
...
+                                _LegacyAppealIssue_ can be associated with a _ContestableIssue_
+                                (i.e., if the `legacyAppealIssues` array is included, it can have
+                                0 or 1 members). Lastly, a _LegacyAppealIssue_ can only be
```
erluti: Maybe change the last sentence to:
```A _LegacyAppealIssue_ can only be associated with one _ContestableIssue_.```

File: app/controllers/api/docs/v3/decision_reviews.yaml [hunk 4]
```diff
@@ -248,34 +248,90 @@ paths:
                               $ref: "#/components/schemas/PayeeCode"
                 included:
                   type: array
+                  description: ContestableIssues Array
                   items:
                     type: object
+                    description: >-
+                      A Higher-Level Review, in large part, can be thought of as a
```
erluti: This first sentence can be removed

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -248,34 +248,90 @@ paths:
...
+                      collection of issues being contested.
+                      Use the `/contestable_issues` endpoint to pull a list of
+                      a veteran's contestable issues to select from, and include
+                      the _ContestableIssue_ objects that you want to contest in this
+                      array.
+                      When including a _ContestableIssue_ obtained from the
```
erluti: I think this "When including a _ContestableIssue_" paragraph is unnecessary.  The spec is spelling out what we expect to see in the request, I don't think we should make some commitment about implementation here in case we end up implementing this in a way where random fields aren't accepted.

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -248,34 +248,90 @@ paths:
...
+                      `/contestable_issues` endpoint, only the ID fields of the
+                      _ContestableIssue_ need to be included. The other fields of a
+                      _ContestableIssue_ will be ignored --modifying a
+                      _ContestableIssue_ cannot be done using this endpoint.
+                      An exception: one can include a `notes` field to provide
+                      additional information.
+                      **Additionally**, a `legacyAppealIssues` array can be added to a
```
erluti: I think we can reduce this section to something like this
```
Associating a _LegacyAppealIssue_ to a _ContestableIssue_ pulls the _LegacyAppealIssue_ out of the legacy appeal system and into AMA. See the `legacyAppealIssues` array below for more info on opting-in legacy issues.
```

app/controllers/api/docs/v3/decision_reviews.yaml
edmangimelli: Both?

edmangimelli: updated existing ticket
https://app.zenhub.com/workspaces/developervagov-5ac272fff41bfa50c0e46c48/issues/department-of-veterans-affairs/vets-contrib/3456

app/controllers/api/docs/v3/decision_reviews.yaml
edmangimelli: I like the parenthetical, but we can drop it 👍

app/controllers/api/docs/v3/decision_reviews.yaml
edmangimelli: I arrived at the same conclusion 👍 nice

edmangimelli: I think I'm less in favor for this change. Removing all parentheticals in this paragraph, i think this reads confusingly

app/controllers/api/docs/v3/decision_reviews.yaml
edmangimelli: Seems reasonable. i just wanted to drive home the point that these fields couldn't be used to update existing records (which might be important to clarify now, with the veteran fields possibly influencing expectations).

edmangimelli: I think it's important to leave in the sentence about `notes` (and also to verify that that is indeed how `notes` is used)

app/controllers/api/docs/v3/decision_reviews.yaml
edmangimelli: ok

app/controllers/api/docs/v3/decision_reviews.yaml
edmangimelli: I'm not in favor of this change. I would like to be very explicit that this field (legacyAppealIssues) is something that the consumer is adding to a ContestableIssue (same with the notes field (but I'd like to doublecheck this with the Caseflow team)). I like having "**Additonally..." and on in addition to the documentation further down with the array.

app/controllers/api/docs/v3/decision_reviews.yaml
erluti: I just meant that we don't need commented out code in our specs (also makes collapsing annoying in edittor.swagger.io), but if it's worth saving we should save the work in a ticket

erluti:
My comments on the copy changes were mostly looking to see where we can trim it down.  I think it's good to document things and lay it out for the user, but we found out in that initial hallway testing that if there's walls of text they don't read it.  My hope was to try to compact it down so they do read it and get the most important information.

app/controllers/api/docs/v3/decision_reviews.yaml
edmangimelli: 👍

PR #12732, opened 2019-11-15
Lighthouse pb issues spec changes

### LINKED ISSUE #12731: APIv3 - Create spec for /legacy_appeal_issues
Describe the request and response for getting a list of issues to associate with a ContestableIssue when creating a decision review.
## Description
LegacyAppealIssues should be added to ContestableIssues on the creation of a decision review, so this endpoint needs to be specified for documentation and design.

erluti:
Resolves #12731

### Description
Describes the request and response for getting a list of issues to associate with a ContestableIssue when creating a decision review.


app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -1230,6 +1230,10 @@ paths:
             type: string
             format: date
       summary: Returns all contestable issues for a specific veteran.
+      description: >-
+        Returns all issues a Veteran could contest in a Decision Review
+        as of the `receiptDate`.  Associate these resutls when creating
```
edmangimelli:
```suggestion
        as of the `receiptDate`.  Associate these results when creating
```

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -1359,6 +1363,102 @@ paths:
                       status: 500,
                       code: unknown_error
                       title: "Unknown error"
+  /legacy_appeal_issues:
+    get:
+      tags:
+        - Issues
+      parameters:
+        - name: veteranId
```
edmangimelli: I switch from `id` to `fileNumberOrSsn` in the create endpoint. Do you want to go with veteranId everywhere?

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -1359,6 +1363,102 @@ paths:
...
+          content:
+            application/vnd.api+json:
+              schema:
+                type: object
+                properties:
+                  data:
+                    type: array
+                    items:
+                      type: object
+                      description: "A legacy appeal issue (to Opt-In these issues to AMA, you include it as a LegacyIssue when submitting ContestableIssues to a new DecisionReview)"
```
edmangimelli:
```suggestion
                      description: "A legacy appeal issue. To opt-in these issues to AMA, you associate them with ContestableIssues when creating a DecisionReview (see the documentation for POST /higher_level_reviews for more details)."
```

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -1359,6 +1363,102 @@ paths:
...
+                        type:
+                          type: string
+                          enum: ["LegacyAppealIssue"]
+                        attributes:
+                          type: object
+                          properties:
+                            legacyAppealId:
+                              type: string
+                            legacyAppealIssueId:
+                              type: string
```
edmangimelli: What about these fields:
```
nodDate
description
disposition
closeDate
note
```
?
https://github.com/department-of-veterans-affairs/caseflow/issues/12194

app/controllers/api/docs/v3/decision_reviews.yaml
erluti: I think we should go with something like `veteranId`.  I like that it's a shorter name and broader term

app/controllers/api/docs/v3/decision_reviews.yaml
erluti: I didn't see those fields on the ContestableIssue I was generating in test.
Looks like I missed `description` though

erluti: I was focused on what we needed from legacy issues to add to create and over looked the data the user would want for displaying them.  Thanks!

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -1359,6 +1363,114 @@ paths:
...
+        - name: receiptDate
+          in: header
+          required: true
+          description: (yyyy-mm-dd) In order to determine contestability of issues, the receipt date of a hypothetical Decision Review must be supplied.
+          schema:
+            type: string
+            format: date
+      summary: Returns all legacy appeal issues for a specific veteran.
+      description: >-
+        Returns all the legacy appeals for a veteran that could be added to
```
pkarman: "legacy appeal issues"

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -1359,6 +1363,114 @@ paths:
                       status: 500,
                       code: unknown_error
                       title: "Unknown error"
+  /legacy_appeal_issues:
```
pkarman: name nit: maybe "legacy issues" instead of "legacy appeal issues" -- a `LegacyAppeal` has a definite meaning in Caseflow context, and so "legacy appeal issues" might be misleading.

edmangimelli: During the happy path ticket (POST /higher_level_reviews), I introducted an abstraction that I thought would make things simpler for our consumers:

That there are `LegacyAppeals` that have `LegacyAppealIssues`.

At the time, I argued to change:
```
vacols_id
vacols_sequence_id
```
to:
```
legacyAppealId
legacyAppealIssueId
```
(respectively),

and it seemed like we all signed off on that idea.

So, I just wanted to clarify,

is our use of legacyAppeal and legacyAppealIssue incompatible with how those terms are currently being used?

The idea is to return, essentially, an array of sequence ids:

```
[
  {legacyAppealId: 3849384, legacyAppealIssueId: 1},
  {legacyAppealId: 3849384, legacyAppealIssueId: 2},
  {legacyAppealId: 3849384, legacyAppealIssueId: 3},
  {legacyAppealId: 7777798, legacyAppealIssueId: 1},
  {legacyAppealId: 7777798, legacyAppealIssueId: 2},
  {legacyAppealId: 7777798, legacyAppealIssueId: 3},
]
```
and not something nested:
```
{
  "3849384": [1,2,3],
  "7777798": [1,2,3],
}
```
because, as Sally explained to us, you associate a sequence id / vacols id combo, and not an entire legacy appeal, and we wanted to return an array of objects that were instantly associatable to our consumers.

leikkisa: Peter made a naming comment, to maybe call them legacyIssue instead of legacyAppealIssue.

I think the data structure you proposed looks easier to work with for these purposes, since they are connected to request issue at the legacy issue level, not the appeal level.

It is a lot more verbose, but I don't think there would ever be too many of them (only certain ones are filtered from all of the veteran's legacy issues).

edmangimelli: @leikkisa  is the way were using `legacyAppeal` different from how the rest of caseflow is using it? If so, could you describe how legacy appeal is currently being used.

Should we rather:

A `legacyAppeal` has `legacyIssues`?

leikkisa: Haha, yeah there are so many issues!

I personally do not have a strong opinion on the renaming, but what you're proposing:
> A `legacyAppeal` has `legacyIssues`
 is aligned with Peter's comment, and that makes perfect sense to me.  We call them legacy issues in our opt-in code.

I think the way you're using legacyAppeal is aligned with us.  The one thing that might be confusing is that there is an ID in Caseflow, which is different from the vacols_id. I think we do usually think of the vacols_id as the legacy appeal ID (I've gotten caught up on that before, where I tried the Caseflow ID and realized the ID was actually a vacols_id).  It might be a little mixed up in our code too.

pkarman: In our existing models, we have a `LegacyAppeal` that has 2 columns on it: `id` and `vacols_id`.

So when I see the phrase "legacy appeal id" I immediately think `LegacyAppeal.id` which is the caseflow-specific autoincrementing PK column for the `legacy_appeals` table.

Which is definitely *not* the same as a `vacols_id` (as Sally mentions).

Maybe we just need some developer comments in the code that references to "legacy appeal id" should be interpreted as `vacols_id` in the code, just to disambiguate.

To be clear: I'm fine calling it "legacy appeal" and "legacy appeal issue" in the outward-facing API docs. I was just calling attention to the fact that "legacy appeal" has a techncal meaning to Caseflow developers that is different than what it means here.

erluti: When we implement this endpoint, I hope the use of the serializers and things will make it obvious to a caseflow-developer that `legacyAppealId` is actually `vacols_id`, but the spec is for the external consumers and our focus is making it more usable to someone who's not up to speed on caseflow internals.

app/controllers/api/docs/v3/decision_reviews.yaml
erluti: thanks, fixed

edmangimelli:
lgtm

PR #12734, opened 2019-11-18
Link to Virtual Hearing

### LINKED ISSUE #12347: Add Virtual Hearing Links to Daily Docket and Hearing Details
Hearing Details

https://va-gov.invisionapp.com/share/8AUO95JMRW7#/screens/391733361

Daily Docket

https://va-gov.invisionapp.com/share/8AUO95JMRW7#/screens/391731332

The Daily Docket displays the VLJ Virtual Hearing Link in the Actions column. Clicking the link/label opens the virtual hearing URL in a new tab.

aplybeah:
Resolves #12347 

- Create `VirtualHearingLink` and added the logic to make it render dynamically. 

**How it _should_ work:**
- This is a functional component. All of the props from hearing/virtualHearing get passed to the link component and populate the link with the appropriate fields. This then generates the link. dynamically. 
- currently, the address is hardcoded as `care.evn.va.gov`. This is the address for the dev environment as defined [here](https://github.com/department-of-veterans-affairs/appeals-team/blob/155200dcbb1479cded6acb38b7cef83510abd8f5/Project%20Folders/Caseflow%20Projects/Hearings/Telehearings/tech-specs/Integrating%20Caseflow%20with%20Pexip.md#joining-a-conference).

**Roadblocks:**
*  I'm still not solid on how to test virtual hearings. What I have tried so far is by creating a few in the rails console to have some saved. The idea here was to have a few virtual hearings saved as a baseline to test that the props I want are going through and how the feature itself get integrated with the database. This was met with varying degrees of success. 
    * Most consistent error was `ActiveRecord::RecordNotUnique: PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_users_unique_css_id" `; even with self-defined css_id that would have overridden the randomly generated one. Additionally, using `FactoryBot` generated this error also.

* Testing by switching a hearing to a virtual hearing is met by a `400`
Screen Shot 2019-11-18 at 10 33 24 AM

I do have plans to revisit this.


client/app/hearings/components/Details.jsx
```diff
@@ -185,7 +186,12 @@ class HearingDetails extends React.Component {
           <h2>Hearing Details</h2>
           <DetailsOverview hearing={this.props.hearing} />
           <div className="cf-help-divider" />
-
+          { success &&
```
lomaxap: `success` is only updated after the hearing is saved, so this will be `false` when the page is loaded, so you should remove this conditional

client/app/hearings/components/Details.jsx
```diff
@@ -185,7 +186,12 @@ class HearingDetails extends React.Component {
...
+            <VirtualHearingLink
+              hearing={this.props.hearing}
+              virtualHearing={virtualHearingForm}
+            />
```
lomaxap: i think it's great that you made this its own component!

We should move this inside `DetailsInputs` component next to the `HearingType` dropdown

client/app/hearings/components/VirtualHearingLink.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const VirtualHearingLink = (props) =>
+  <div >
```
lomaxap: since we're passing the `hearing` as a prop, we can just do the `hearing.isVirtual` check inside this component and return null if it's false. This way we won't have to add that conditional on every page that we include this link.

client/app/hearings/components/VirtualHearingLink.jsx
```diff
@@ -0,0 +1,27 @@
...
+    {props.hearing.isVirtual &&
+    <a href={`https://care.evn.va.gov/webapp/?conference=${props.hearing.alias}
+    &pin=${props.hearing.pin}&join=1&role=${props.hearing.role}`}
```
lomaxap: you're referencing the hearing here, but `pin`, `role`, and `alias` are defined on on the `virtualHearing`

client/app/hearings/components/dailyDocket/DailyDocketRow.jsx
```diff
@@ -238,6 +239,7 @@ class DailyDocketRow extends React.Component {
     const inputProps = this.getInputProps();
 
     return <div {...inputSpacing}>
+      <VirtualHearingLink hearing={hearing} />
```
lomaxap: you need to also pass the `virtualHearing`, that's actually [serialized nested](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/serializers/hearings/hearing_serializer.rb#L74-L78) under `hearing`, so that can just be `virtualHearing={hearing.virtualHearing}`

pshahVA:
Can you please add a screenshot of the link on the UI from this ticket?

Issue #12689, opened 2019-11-12
Send users a message when their asyncable job hasn't succeeded after 24 hours


leikkisa:
## User or job story
As a user, I'd like to know when a job that I submitted hasn't succeeded after 24 hours, and why it hasn't succeeded. This will let me know if I should contact support.

## Acceptance criteria
- [ ] When a job has been re-tried more than 24 hours after it has been submitted, and it still fails, send a message to the users inbox (see copy below)
- [ ] In order to prevent duplicates, do not send this message to the user if they have already received one
- [ ] Add message type to schema (check with Peter)
- [ ] Include a screenshot in the Github issue of the inbox message

## Release notes

A message will be sent to a users inbox if their job hasn't succeeded after 24 hours with more information on why it didn't succeed.  This will help inform users whether they need to create a support ticket.

### Designs
Copy:
The Intake processing job for {claim_review_type} {job_id} was unable to complete because of an error: {error_name}.

If this error reflects a network problem of some kind, it will be automatically re-tried at {estimated_retry_time}.

If this error reflects a data integrity issue such as duplicate records in BGS and/or VBMS, it will require manual resolution via a YourIT ticket. The Intake processing job cannot continue until the error is resolved.

### Background/context
This ticket has been broken out from: https://github.com/department-of-veterans-affairs/caseflow/issues/11907 which includes implementation of the inbox feature and further details and designs.

See #12214 for details on job notes.

PR #12740, opened 2019-11-18
Add message type column to messages db schema

nanotone:
Connects #12689 

### Description
As part of #12689 (if a job fails after 24 hours, message the user once) this PR is a schema change proposal to add a `message_type` column to the `messages` table. Some benefits of adding this column would be:
- Primary benefit: Easy to check whether a job has triggered any `job_failing` messages yet -- compared to inspecting the text, or doing timestamp math with assumptions
- Easy to regenerate message prose, in case we make any changes to the English (see discussion in #12707) and want them to be retroactive
- Easy to implement inbox filtering by message type, since we'd like Caseflow Inbox to be useful for non-job-related messaging

As a taste of things to come (see various acceptance criteria in #11907) I am also proposing a set of `message_type` enum values, and welcoming thoughts on better enum values or even a better alternative to all this.

### Schema Changes
* [x] Column comments updated

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -8,7 +8,7 @@ class Message < ApplicationRecord
     job_note_added: "job_note_added",
     job_failing: "job_failing",
     failing_job_succeeded: "failing_job_succeeded",
-    job_canceled: "job_canceled"
+    job_cancelled: "job_cancelled"
   }
 
   scope :read, -> { where.not(read_at: nil) }
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -4,6 +4,13 @@ class Message < ApplicationRecord
   belongs_to :user
   belongs_to :detail, polymorphic: true
 
+  enum message_type: {
+    job_note_added: "job_note_added",
+    job_failing: "job_failing",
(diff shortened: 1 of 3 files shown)


app/models/message.rb
```diff
@@ -4,6 +4,13 @@ class Message < ApplicationRecord
   belongs_to :user
   belongs_to :detail, polymorphic: true
 
+  enum message_type: {
+    job_note_added: "job_note_added",
+    job_failing: "job_failing",
+    failing_job_succeeded: "failing_job_succeeded",
+    job_canceled: "job_canceled"
```
leikkisa: This is a funny thing, because the founding intake folks felt strongly about spelling this with one L, but since thing the prevailing winds in the app overall say we should use two L's - cancelled.

Working on creating consistency in the app is still oustanding, but I'm going to suggest using the new guidance here.

lomaxap: there was a rumour started that the single L was the "American" way of spelling it.

nanotone: >rumour
this is too much for me

Issue #12397, opened 2019-10-16
Create table to connect a request issue to many legacy issues


leikkisa:
### User or job story
As a user, I would like to be able to connect multiple legacy issues to a request issue.

### Acceptance criteria
- Create a new table called request_legacy_issues
  - Add a table DB comment (see "copy" below)
- Add the following fields to the new table, with DB comments
  - id (primary key)
  - timestamps
  - request_issue_id (required, foreign key)
  - vacols_id (required)
  - vacols_sequence_id (required)
  - legacy_issue_optin_id (optional, foreign key)

**Note - the following two bullets can be separated into a separate chunk of work if desired.**
- the current data from request_issues and legacy_issue_optins is copied to the new table via a migration
- when a new request issue is created and it's connected to a legacy issue, there is a hook that saves a copy of the relevant data to the new table

### Copy
New table comment (feel free to edit):

When a user creates a request issue for a veteran, they have the opportunity to connect it to one or more legacy issues from VACOLS.  If the veteran also chooses to opt their legacy issues into AMA, the issues are closed in VACOLS through a legacy issue opt-in. This table stores the legacy issues connected to each request issue, and the record for opting them into AMA (if applicable).

### Release notes
N/A

#### Designs
N/A

#### Background/context
This is a technical ticket to support a one-to-many relationship between request issues and legacy issues.  

Not all vacols issues that are connected to a request issue get opted-in.  They only get opted in if the veteran selects to have their legacy issues opted in to AMA.

The first step will be creating and maintaining a copy of the current data (in request issues and legacy issue optins) in the new table.  Once that is ready, the next step is to switch the logic to using the the new table.  Then the vacols_id and vacols_sequence_id fields can be removed from the request_issues table.

After the data structure is ready, the final ticket for supporting multiple legacy issues is here:
See: https://github.com/department-of-veterans-affairs/caseflow/issues/11103

#### Technical notes
A before_save or after_save hook will help populate the data not populated during the migration to backfill the data.

The existing table and column comments in the schema may be useful for coming up with the column comments on the new tables.

pkarman:
The story in #11103 says

> As a user, I would like to be able to opt-in multiple legacy issues per request issue.

which seems different than "connecting multiple". AFAIK the current Add Issues flow only gives the modal for connecting legacy issues when the opt-in box is checked, effectively making all connections == opt-ins.

> Not all vacols issues that are connected to a request issue get opted-in.

[...]

PR #8470, opened 2018-12-28
Remove vacols_id and vacols_sequence_id columns from legacy_issue_optins

pkarman:
connects #8223 

### Description
DRY up some db column duplication.

PR #12746, opened 2019-11-18
Add table for multiple legacy issues per request issue

leikkisa:
connects #12397

### Description
This creates a table for storing multiple legacy issues connected to a single request issue.

Data will be backfilled and hooks added to save to this table in a follow up PR.  In order to backfill the data I'd like to use a method create_legacy_issue! in a database migration, so I am adding that method in this PR.  Database migrations are run before the code is updated, so calling this method in the database migration would likely break deployment.
 
### Database Changes
*Only for Schema Changes*

* [x] Column comments updated


db/migrate/20191118195627_create_legacy_issue_table.rb
```diff
@@ -0,0 +1,11 @@
...
+      t.belongs_to :request_issue, null: false, comment: "The request issue the legacy issue is being connected to."
+      t.references :legacy_issue_optin, foreign_key: true, comment: "If the veteran selected to opt their legacy issues into AMA, and the legacy issue was eligible for opt-in, a legacy issue-optin is created."
+      t.string :vacols_id, null: false, comment: "The VACOLS ID of the legacy appeal that the legacy issue is part of."
+      t.integer :vacols_sequence_id, null: false, comment: "The sequence ID of the legacy issue on the legacy appeal. The vacols_id and vacols_sequence_id form a composite key to identify a specific legacy issue."
+      t.timestamps null: false, comment: "Default created_at/updated_at"
+    end
+  end
+end
```
leikkisa: I didn't add any uniqueness constraints. legacy_issue_optin can be null, and if an issue is removed, then the same vacols_id and vacols_sequence_id could get added again elsewhere.

Lmk if there are any migration improvement ideas though!

pkarman: we originally had the `vacols_id` and `vacols_sequence_id` columns on the `legacy_issue_optins` table, then dropped them when we decided it was one-to-one rather than one-to-many. Now that we are resurrecting one-to-many, I'm curious why not just adding those columns back? I'm not clear on why a new table is needed?

pkarman: see #8470

leikkisa: @pkarman Snap - I remember replying to this, but I don't see it here 🤦‍♀

The reason we didn't move to storing it on legacy issue opt-in is because not all legacy issue "connections" get an opt-in, it depends on the veteran's opt-in selection and whether the issue is eligible.  We talked about having it in both places, but you used delegation which totally handled it (and is the first time I learned about that!).

There are a couple of alternative options:
1. We could store an array for both vacols fields in RequestIssue.  That might get interesting if the two issues have different levels of eligibility. It's also less easy to take advantage of the data when it's stored that way.
1. We could create legacy issue opt-ins for all legacy issue connections, whether they're opted in or not.  I think that would be confusing and inefficient, plus it'll be harder to backfill.

I talked with @tomas-nava about this first and he suggested this approach, which seems much better than the alternatives I could think up.

leikkisa: I don't remember ever originally thinking it was one-to-many. I just remember thinking we couldn't only store the vacols fields on legacy issue-opt-in, and wasn't excited about duplicating them on two tables.

pkarman: Ah ok, I missed the part about needing to connect legacy issues with request issues *without* the opt-in. Teaches me to read the issue more carefully.

app/models/request_issue.rb
```diff
@@ -570,6 +570,19 @@ def title_of_active_review
     duplicate_of_issue_in_active_review? ? ineligible_due_to.review_title : nil
   end
 
+  # Temporary method for populating legacy_issue table until we fully switch over
+  def create_legacy_issue!
```
nanotone: When / in what context will this method be called?

leikkisa: This will be called in the database migration where we backfill the new table.

It will potentially be used in the code when we're saving request issues and creating legacy issue optins (although that will probably be a little different), so that we're saving the data in both places at the same time (with the goal of achieving parity between the old and new data stores).

app/models/request_issue.rb
```diff
@@ -570,6 +570,19 @@ def title_of_active_review
...
+    return unless vacols_id
```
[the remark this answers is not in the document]

leikkisa:
I'm cool with doing the data migration outside of a DB migration, we used to just post our scripts in the issue.  The DB migrations are nice for having a solid record, but I think it's easier to do outside of a migration.

PR #12747, opened 2019-11-18
Remove actionNameOfTask() helper function

lowellrex:
Removes a utility function made redundant when we [created subclasses for each ColocatedTask action](https://github.com/department-of-veterans-affairs/caseflow/pull/11295) instead of relying on the deprecated `action` field in the `tasks` table.

### Testing Plan
1. Check out `origin/master`
1. Log in as `BVALSPORER`
1. Go to your personal queue, notice that the task types in the "Type(s)" column appear normal
1. Check out this branch
1. Return to your personal queue, notice that the task types stay the same.


client/app/queue/utils.js
```diff
@@ -553,8 +552,6 @@ export const taskActionData = ({ task, match }) => {
   return null;
 };
 
-export const actionNameOfTask = (task) => CO_LOCATED_ADMIN_ACTIONS[task.label] || task.label;
```
lowellrex: `CO_LOCATED_ADMIN_ACTIONS[task.label]` is now always returning `undefined` since the task labels are being returning from the back-end as the values instead of the keys in `CO_LOCATED_ADMIN_ACTIONS.json`.

```
caseflow $> git grep -A1 "def .*label" app/models/tasks/*colocated_task.rb
```

PR #12753, opened 2019-11-19
Only send updated params to /hearings on DailyDocket / HearingDetails

### LINKED ISSUE #12519: Refactor Hearing Details to Only Send Edited Data

lomaxap:
resolves #12519 

- all requests from Hearing Details and Daily Docket now only send parameters that have been updated
  - adds `dateDiff` to `client/hearings/utils` that compares 2 objects returns an object with the properties that have been changed
- refactors `HearingTypeDropdown` into separate component
- moves virtual hearing feature tests to `spec/features/hearings/virtual_hearings`


app/models/hearing.rb
```diff
@@ -85,8 +85,6 @@ def readable_location
   end
 
   def readable_request_type
-    return "Virtual" if virtual?
```
lomaxap: i think we should use `virtual?` / `isVirtual` to determine this. Pretty sure there's existing logic in the front end that assumes this can only be Central or Video.

I'm thinking of `virtual` as a type of `video` hearing and not as a request type.

client/app/components/common/reducers.js
```diff
@@ -50,20 +50,13 @@ const formsReducer = (state = {}, action = {}) => {
...
       [action.payload.formName]: {
-        $set: {
-          ...formState,
-          ...action.payload.formData
-        }
+        $set: action.payload.formData === null ?
+          {} : {
+            ...formState,
+            ...action.payload.formData
+          }
```
lomaxap: fixing some random redux plumbing. this may be unnecessary, but i didn't like deleting an object property like i was before

client/app/hearings/components/Details.jsx
```diff
@@ -42,56 +43,61 @@ class HearingDetails extends React.Component {
...
     };
-  }
 
-  componentDidMount() {
-    this.setInitialFormData();
+    this.updateAllFormData(initialFormData);
   }
 
-  setInitialFormData = () => {
+  getInitialFormData = () => {
```
lomaxap: `formData` takes a subset of serialized hearing data that we allow to be edited.

client/app/hearings/containers/DailyDocketContainer.jsx
```diff
@@ -98,66 +98,37 @@ export class DailyDocketContainer extends React.Component {
...
-      transcript_requested: hearing.transcriptRequested,
-      notes: hearing.notes,
-      hearing_location_attributes: hearing.location ? ApiUtil.convertToSnakeCase(hearing.location) : null,
-      virtual_hearing_attributes: virtualHearing ? ApiUtil.convertToSnakeCase(virtualHearing) : null,
-      scheduled_time_string: hearing.scheduledTimeString,
-      prepped: hearing.prepped,
-      ...legacyValues,
-      ...amaHearingValues
-    }, _.isNil);
+  formatHearingFormData = (hearingFormData) => {
```
lomaxap: since we're only getting the data that's been changed, we no longer need to determine what data to send for legacy hearings vs. ama hearings.

app/models/hearing.rb
ferristseng: I'm usually interpreting this field as "request type that we display to the user", since the actual request type is stored as a code (which isn't super readable). I'm fine with tweaking that assumption, but maybe there needs to be a comment on how we use this field because I find the naming of it to be a bit confusing.

aplybeah: This logic is referenced in several spots in the front end. A few places to start would be `onRequestTypeChange` in `HearingDayAddModal` and  `DetailsInputs`

PR #12758, opened 2019-11-20
Use request issues update user for establishment and cancellations

leikkisa:
### Description
When a user edits the issues on a claim, they can add and remove issues.  This can subsequently result in adding or removing end products.  These establishments or cancellations should be associated to the user who performed the update.

Currently, the user is manually being set to the system user. This PR sets the user to the RequestIssuesUpdate user, if a RequestIssuesUpdate is the source of the DecisionReviewProcess job.

Please note: This is the first (small change), of what may be many changes with respect to stopping from using the system user. This specific request is for EP cancellations, so the user has been updated for RequestIssuesUpdates only in this PR. More time and investigation is necessary before making additional changes, such as automatically generated decision reviews, EPs established from intakes, or whatever else BGS brings to our attention.

### DIFF
--- a/app/jobs/decision_review_process_job.rb
+++ b/app/jobs/decision_review_process_job.rb
@@ -10,7 +10,7 @@ class DecisionReviewProcessJob < CaseflowJob
     @decision_review = thing_to_establish
 
     # If establishment is for a RequestIssuesUpdate, use the user on the update
-    establishment_user = decision_review.try(:user) || User.system_user
+    @establishment_user = decision_review.try(:user) || User.system_user
 
     # restore whatever the user was when we finish, in case we are not running async (as during tests)
     current_user = RequestStore.store[:current_user]
@@ -37,7 +37,7 @@ class DecisionReviewProcessJob < CaseflowJob
 
   private
 
-  attr_reader :decision_review
+  attr_reader :decision_review, :establishment_user
 
   def ok_to_ping_sentry?
     decision_review.sort_by_last_submitted_at > (Time.zone.now - 4.hours)
--- a/app/jobs/decision_review_process_job.rb
+++ b/app/jobs/decision_review_process_job.rb
@@ -9,9 +9,12 @@ class DecisionReviewProcessJob < CaseflowJob
   def perform(thing_to_establish)
     @decision_review = thing_to_establish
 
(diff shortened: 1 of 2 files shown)


app/jobs/decision_review_process_job.rb
```diff
@@ -9,9 +9,12 @@ class DecisionReviewProcessJob < CaseflowJob
   def perform(thing_to_establish)
     @decision_review = thing_to_establish
 
+    # If establishment is for a RequestIssuesUpdate, use the user on the update
+    @establishment_user = decision_review.try(:user) || User.system_user
```
pkarman: I think the rubocop recommendation now is `decision_review&.user` but I could be wrong.

spec/jobs/decision_review_process_job_spec.rb
```diff
@@ -77,6 +77,24 @@ def sort_by_last_submitted_at; end
     end
   end
 
+  context ".establishment_user" do
```
pkarman: you've made the `establishment_user` a private method, so testing it explicitly like this is a little unusual.

You could make `establishment_user` a public `attr_reader` to do the easiest thing (and no more need for `send`).

Alternately, you could just test the `RequestStore.store[:current_user]` after the test is done instead of the `send(:establishment_user)`.

pkarman: What you really want to test here, but haven't, is that the `establishment_subject` gets the expected user to send to VBMS. I think that would require some poking into the `establishment_subject` with an observer like `expect().to have_received...` or similar.

app/jobs/decision_review_process_job.rb
leikkisa: Try works if there isn't a "user" method, which is true for decision reviews.

spec/jobs/decision_review_process_job_spec.rb
leikkisa: Yeah, if it's okay to save this until tomorrow I can set up the test to actually check what gets sent to VBMS.  It requires some additional test setup to get a request issue update to want to cancel an EP and I am a little short on time for tonight.

I couldn't figure out how to test RequestStore.store[:current_user] directly since it gets changed back to whatever it was within the perform method.

Issue #12237, opened 2019-09-27
Audit *_views tables to determine best storage options


pkarman:
We are accruing millions of db records in these tables:

* api_views
* appeals_views
* document_views
* hearing_views

Investigate whether these tables are still necessary and what need(s) they are helping to fulfill. If the original need was reporting statistics, perhaps some other data storage would be appropriate (e.g. DataDog, Google Analytics, separate RDS instance).

aroltsch:
1. A PR that adds `appeal_views` used by Reader: https://github.com/department-of-veterans-affairs/caseflow/pull/2349/files#diff-e952e317643190646a1a08902bd13111R6 
1. A PR that adds `api_views` to record every time the API is accessed, record the VBMS id, API key, and time: https://github.com/department-of-veterans-affairs/caseflow/pull/4540
1. A PR that adds `document_views` to mark opened documents as read https://github.com/department-of-veterans-affairs/caseflow/pull/1166
1. A PR to add `hearing_views` to record worksheet views https://github.com/department-of-veterans-affairs/caseflow/pull/4198

I think some of these metrics can be migrated to DD or google analytics.

pkarman:
Adding Dangerfile warning about these tables if we keep them and need to modify them.

https://github.com/department-of-veterans-affairs/caseflow/pull/12764/files

PR #12764, opened 2019-11-20
Danger warn on large table migration

pkarman:
We've had multiple times recently when creating indices on large tables caused deployment failures due to connection timeouts. Try and prevent that my warning us about tables known to be large.

Pasting screenshot here (and removing the migration that triggered it).

Screen Shot 2019-11-20 at 10 19 11 AM

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -74,15 +74,16 @@ KNOWN_LARGE_TABLES = %w[
   document_views
   hearing_views
   versions
-]
+].freeze
 
+large_table_migration_pattern = /(add_index|add_column) :(#{KNOWN_LARGE_TABLES.join('|')})/
 migrations_on_large_tables = git.diff.flat_map do |chunk|
   chunk.patch.lines.grep(/^\+\s*\w/).select do |added_line|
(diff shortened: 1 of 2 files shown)


Dangerfile
```diff
@@ -61,3 +61,30 @@ if !new_specs.empty?
...
+KNOWN_LARGE_TABLES = %w[
+  annotations
+  api_views
+  appeal_views
+  claims_folder_searches
+  documents
+  documents_tags
+  document_views
+  hearing_views
+  versions
```
nanotone: Can you comment briefly on how these large tables were identified -- i.e. everything above a certain row count? And how do we revisit this list periodically?

pkarman: I manually ran `select count(*) from ...` on each one. All of them were in the multi-millions.

Revisiting would be good. I think we'll have a better sense of proactively identifying table size as part of the metrics work we are doing. We've also raised the issue of whether we should be storing this kind of data in the db in the first place. See #12237

nanotone: Thanks for the context. It does look like many of these are write-only (or read only for reporting) and should be shipped out of the app for logging/metrics/analysis to be done elsewhere. But this PR lgtm.

PR #12767, opened 2019-11-20
Message user about first job failure after 24 hours

nanotone:
Connects #12689 

### Description
User gets an Inbox message when any of their jobs encounters a failure after 24 hours of trying.

### Acceptance Criteria
- [ ] When a job has been re-tried more than 24 hours after it has been submitted, and it still fails, send a message to the users inbox
- [ ] In order to prevent duplicates, do not send this message to the user if they have already received one

### Testing Plan
Use a feature spec that already creates a fake job with appropriately old timestamps, e.g.
1. Instrument and run `spec/feature/asyncable_jobs/asyncable_jobs_index_spec.rb:61`
2. `pry> hlr.update_error!("oops!")`
3. Switch to the user mentioned in the resulting `Message` object
4. Visit `/inbox` and verify that message landed in the inbox 
5. Also added a group of tests for the `AsyncableJobMessaging` service class, for all the edge cases

### User Facing Changes
Screen Shot 2019-11-20 at 13 57 38

### Notes
- Is there a way to fully simulate the processing of a fake job, in the tests?
- The messaging logic is currently triggered by `Asyncable#update_error!` which feels a bit tightly coupled to the storage side of the ORM. However, any alternative would require each Asyncable implementation to explicitly trigger messaging, which is brittle and non-DRY. Thoughts?


nanotone:
Also fixing a really 🤦‍♂ spec filename.

spec/services/asyncable_job_messaging_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+        message_count = owner.messages.count
+        subject
+        expect(owner.messages.count).to eq message_count
+      end
+    end
+
+    context "when a failure first happens after 24 hours" do
+      it "sends a message to the job owner" do
+        job
+        Timecop.travel(Time.zone.now.tomorrow)
```
leikkisa: I've never seen this before - time travel!

app/services/asyncable_job_messaging.rb
```diff
@@ -26,10 +28,37 @@ def add_job_note(text:, send_to_intake_user:)
         Message.create!(
           detail: job_note,
```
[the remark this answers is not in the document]
nanotone: I think it could go either way since the `detail` is polymorphic. I originally went with `job_note` because it's more informative/relevant for a `job_note_added` message, and you can always get the job from the note.

In any case, the `detail` attribute isn't currently used, since `text` serves all of the UI's needs.

app/services/asyncable_job_messaging.rb
```diff
@@ -26,10 +28,37 @@ def add_job_note(text:, send_to_intake_user:)
...
       job_note
     end
   end
+
+  def handle_job_failure(err:)
+    return unless messaging_enabled_for_job_attempt?
+    return if Message.where(detail: job, message_type: "job_failing").any?
+
+    if err.length > 80
+      err = err[0..80] + "..."
```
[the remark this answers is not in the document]
nanotone: This is related to the way `#update_error!` gets called with just a string, which in the codebase currently can be
- `"No payee code"`
- `"DTA SC creation failed"`, or
- `error.inspect`

The truncation to 80 chars is just a safeguard.

app/services/asyncable_job_messaging.rb
```diff
@@ -26,10 +28,37 @@ def add_job_note(text:, send_to_intake_user:)
...
+          message_type: "job_note_added"
         )
       end
       job_note
     end
   end
+
+  def handle_job_failure(err:)
+    return unless messaging_enabled_for_job_attempt?
+    return if Message.where(detail: job, message_type: "job_failing").any?
```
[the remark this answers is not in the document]
nanotone: Nice, TIL some more ActiveRecord magic ✨

nanotone:
@leikkisa Thanks for the comments; I added responses and will look at getting some feature test coverage for this. Any suggestions for which feature test to add to (or should I create a new spec for messaging)?

app/models/concerns/asyncable.rb
```diff
@@ -217,6 +217,7 @@ def clear_error!
 
   def update_error!(err)
     update!(self.class.error_column => err)
+    AsyncableJobMessaging.new(job: self).handle_job_failure(err: err)
```
pkarman: I would just override `update_error!(err)` in the `ClaimReview` model. That's the only place we should be sending messages right now.

app/services/asyncable_job_messaging.rb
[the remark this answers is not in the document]
pkarman: Look at `AsyncableJobsReporter#sanitize_error` for an example. For async jobs that fail in this use case, they should be getting errors from `error.inspect` which should lead with the (hopefully helpful) error class name.

I would not truncate based on length because (a) there may be PII in the first 80 chars, and (b) we warn "just enough" information for that user to click the link to the Job Details page.

app/services/asyncable_job_messaging.rb
```diff
@@ -26,10 +28,37 @@ def add_job_note(text:, send_to_intake_user:)
...
+      text: message_text,
+      user: job.asyncable_user,
+      message_type: "job_failing"
+    )
+  end
+
+  private
+
+  def messaging_enabled_for_job_attempt?
+    return false unless job.submitted? && job.asyncable_user
```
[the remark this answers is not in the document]
pkarman: I would not put the `ClaimReview` check here, since this class should not need to know too much about the `job`. If you override `update_error!` in ClaimReview, then `job` will never (at this point) be anything else.

app/services/asyncable_job_messaging.rb
```diff
@@ -26,10 +28,37 @@ def add_job_note(text:, send_to_intake_user:)
...
+
+  def handle_job_failure(err:)
+    return unless messaging_enabled_for_job_attempt?
+    return if Message.where(detail: job, message_type: "job_failing").any?
+
+    if err.length > 80
+      err = err[0..80] + "..."
+    end
+    message_text = <<-EOS.strip_heredoc
+      The Intake processing job for <a href="#{job.path}">#{job.class} #{job.id}</a> was unable to complete because of an error: #{err}
```
pkarman: wordsmithing: maybe avoid "Intake processing job" to make it a little more general use case? How about instead just "job".

spec/services/asyncable_job_messaging_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+      it "doesn't send a message to the job owner" do
+        message_count = owner.messages.count
+        subject
+        expect(owner.messages.count).to eq message_count
+      end
+    end
+
+    context "when a failure first happens after 24 hours" do
+      it "sends a message to the job owner" do
+        job
```
pkarman: if you put a `let!(:job)` above, you don't need to call the `job` fixture method explicitly.

spec/services/asyncable_job_messaging_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+        expect(owner.messages.count).to eq message_count + 1
+        expect(owner.messages.last.text).to match("unable to complete because of an error")
+        expect(owner.messages.last.text).to match(job.path)
+      end
+    end
+
+    context "when multiple failures happen after 24 hours" do
+      it "sends only one message to the job owner" do
+        job
+        Timecop.travel(Time.zone.now.tomorrow)
```
pkarman: I typically see `Timecop.travel` with a block so that the scope of the travel is limited. Not sure if it matters (maybe the `it` block scope limits the effects?) but might be clearer to next person that the effect is limited to the current context.

leikkisa:
@nanotone I think the Inbox feature spec might be a good place

app/services/asyncable_job_messaging.rb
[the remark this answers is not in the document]
leikkisa: Yeah, I see what you're saying.  It is "lossy" to connect it to the job directly.

I noticed it was a little unintuitive for me to get all of the messages on a job, or get to the job from the message.  If we end up thinking that's useful to have though, we could handle those in methods, that way we can access data easily AND preserve the connection.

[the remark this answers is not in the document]
nanotone: Yeah, it's true that `job.messages` doesn't get you the job's notes' messages, and you need either `message.detail` or `message.detail.job` to get the job from the message. I'm 👍 on handling that in methods later, since there's currently no need to for this (i.e. filtering inbox for all messages related to one job).

nanotone:
Updated to address all comments, including a small refactor to move error sanitization from the jobs reporter class to the job itself (with concrete implementation in ClaimReview) so that job messaging can also use it. I'm not in love with ClaimReview's assumptions about error string format, but this shouldn't be any worse than before.

Also note: added an inbox feature test for this. Thanks for the advice on `DecisionReviewProcessJob` and mocking, @leikkisa!

app/models/claim_review.rb
```diff
@@ -110,6 +111,16 @@ def establish!
     processed!
   end
 
+  def update_error!(err)
+    super
+    AsyncableJobMessaging.new(job: self).handle_job_failure
+  end
+
+  def sanitized_error
```
pkarman: this could go on the Asyncable concern if we wanted to make it available everywhere.

app/models/concerns/asyncable.rb
```diff
@@ -190,6 +190,10 @@ def asyncable_user
     nil # abstract method intended to be overridden
   end
 
+  def sanitized_error
+    nil # abstract method intended to be overridden
```
pkarman: could just implement this as ClaimReview does. If there are models setting error strings that are *not* `Exception.inspect` then we should fix those in any case.

app/services/asyncable_jobs_reporter.rb
```diff
@@ -21,7 +21,7 @@ def as_csv
           job[klass.submitted_at_column],
           job[klass.last_submitted_at_column],
           job[klass.attempted_at_column],
-          sanitize_error(job),
+          job.sanitized_error,
```
pkarman: great!

app/models/concerns/asyncable.rb
nanotone: I believe there are (commented earlier https://github.com/department-of-veterans-affairs/caseflow/pull/12767#discussion_r349134072) but it's minor and I'm happy to do that in a separate PR. For now, I'll just promote `ClaimReview#sanitized_error` to here.

PR #12772, opened 2019-11-21
Handle Various Virtual Hearing Confirmation Messages

### LINKED ISSUE #12759: Virtual Hearing Confirmation Messages
Create confirmation when a hearing is change to / from virtual.

- When changing to Virtual for the first time:
https://va-gov.invisionapp.com/share/8AUO95JMRW7#/screens/391731030
- When changing back to video / central from virtual:
https://va-gov.invisionapp.com/share/8AUO95JMRW7#/screens/391798749
- When updating 1 or 2 emails
https://va-gov.invisionapp.com/share/8AUO95JMRW7#/screens/391798214

### AC
- [ ] correct confirmation shows after save for each scenariou

lomaxap:
resolves #12759 

- creates `UserAlert` model that corresponds to `Alert.jsx` props (`type`, `title`, `message`)
- creates generic component / reducers / actions for displaying alerts from server
  - `UserAlerts` component expires alerts after 30 seconds, instead of manually having to flush alerts. 
- creates `VirtualHearingUserAlertGenerator` to built virtual hearing alerts based on `virtual_hearing_attributes` passed to hearings controller
- ❗️ `update` actions on hearing controller now returns
```ruby
{
  data: hearing_data,
  alerts: array_of_alerts
}
```


app/models/hearings/forms/base_hearing_update_form.rb
```diff
@@ -40,18 +46,20 @@ def email_sent_flag(attr_key)
     !(status_changed || virtual_hearing_attributes.key?(attr_key))
   end
 
-  def create_or_update_virtual_hearing
-    created = false
+  def created?
+    @created ||= false
```
lomaxap: need access to whether or not hearing was created when determining what message to return. i don't love this.

ferristseng: alternatively, you could return `created` from `create_or_update_virtual_hearing`, and pass it to `add_virtual_hearing_alert`.

PR #12777, opened 2019-11-21
Tooltip Accessibility Fix

### LINKED ISSUE #12141: Mouseover tool tips not accessible to keyboard-only or other AT users – they should be
https://www.sarasoueidan.com/blog/accessible-tooltips/

Information regarding a hearing is displayed with Mouse Overs. This information cannot be viewed with a keyboard or read by assistive technology programs. (New: 08/21/2019)

aplybeah:
Resolves #12141 
I tried my best to follow the functionality of [this design pattern](https://cdpn.io/Moiety/debug/LaPvWy); specifically the pattern labelled **Tooltip with Additional Information** . I achieved this by adding `tabIndex=0` to the main tooltip component, and adding a few attributes from the [`react-tooltip` library that we are using](https://react-tooltip.netlify.com/).

**_Attributes:_**
* `tabIndex=0`: adds the tooltip container to the tab order, making it keyboard accessible
* `data-event`/ `data-event-off`: These attributes activate/deactivate the tooltip based on widow event handlers.
* `aria-describedby`: [W3C standard](https://w3c.github.io/aria-practices/#tooltip) states that the element referencing the tooltip content must use this attribute. This attribute was designed to be more descriptive than `aria-labelledby`

Currently the screen reader will announce: "<Element Name/Label> , tooltip, <tooltip content>. You are currently on a tooltip"

_Nit_
If there is an element that has a tooltip that is already in the tab order (e.g. button and anchor tags) information gets read out twice.


lomaxap:
@alisan16 @carodew see alsia's note:
> Currently the screen reader will announce: "<Element Name/Label> , tooltip, . You are currently on a tooltip"
> Nit
> If there is an element that has a tooltip that is already in the tab order (e.g. button and anchor tags) information gets read out twice.

Does this meet the AC for this ticket?

alisan16:
Can you share a screenshot/link to a screen where the tooltips are read twice?

lomaxap:
If we're using a tooltip library that isn't itself accessible, i vote we switch to a lib that's actually accessible instead of trying to hack this one. That would be a larger lift for this ticket, but I think the ideal solution.

client/app/components/Tooltip.jsx
```diff
@@ -25,11 +26,20 @@ const Tooltip = (props) => {
   });
 
   return <React.Fragment>
-    <span data-tip data-for={id}>{props.children}</span>
+    <span data-tip data-for={id} role="tooltip" data-event="focus mouseenter"
+      data-event-off="keydown mouseleave" tabIndex={0} aria-describedby={id}>{props.children}</span>
     <span {...tooltipStyling} >
       <ReactTooltip effect="solid" id={id} offset={offset} place={position} multiline>{text}</ReactTooltip>
```
lomaxap: if i'm understanding this right, i think you can create a new id like `a11y-${id}` and write this like
```
<span data-tip data-for={id} role="tooltip" data-event="focus mouseenter"
      data-event-off="keydown mouseleave" tabIndex={0} aria-describedby={`a11y-${id}`}>{props.children}</span>
<p id={`a11y-${id}`} hidden >{text}</p>
```

aplybeah:
@alisan16 The tooltip information isn't read out twice; it's more of the content of the element. I was mostly expecting a readout that would express that a button would generate a tooltip, rather than announcing a tooltip then announcing a button.
Screen Shot 2019-11-25 at 12 07 39 PM
Screen Shot 2019-11-25 at 12 03 52 PM

client/app/components/Tooltip.jsx
lomaxap: i don't think i'm understanding this right `<Element Name/Label> , tooltip, . You are currently on a tooltip`, do you expect it to read the element name?

aplybeah: @lomaxap , no...I was expecting a readout more like `<Element Name/Label>, <Tooltip info>, button, tooltip. You are currently on a button`.  I wanted to make sure that the readout was clear and understandable. At the same time, W3 hasn't reached an agreement on what an accessible tooltip looks like yet. [Design pattern note](https://www.w3.org/TR/wai-aria-practices/#tooltip)

pkarman:
@alisan16 could this be merged?

PR #12788, opened 2019-11-22
ETL syncer

### LINKED ISSUE #12659: ETL syncer service
Once we have created the ETL database and models, we will need a basic syncer service class structure that can take an origin model and target model and keep them in sync.

Example:

```ruby
class ETL::Syncer
  def initialize(origin:, target:)
    @origin = origin
    @target = target
  end

  def call
    instances_needing_update.each do |original|
      target.sync_with_original(original)
    end
  end

  private

  attr_reader :origin, :target 

  def instances_needing_update
    # looked at origin.`updated_at` vs Time.now e.g.
  end
end
```

## AC

- [x] Syncer classes for
  * [x] Organization
  * [x] Person
  * [x] OrganizationsUser
  * [x] User
  * [x] Appeal
  * [x] Task

pkarman:
connects #12659 

### Description
ETL models must be kept in sync with their corresponding original models. Ordinarily this is an incremental process (sync everything that has changed since the last time we synced). Sometimes we will need a full sync (as for the first time, or for schema/logic changes).

This PR introduces the `ETL::Syncer` service and its subclasses. Each ETL model has a corresponding Syncer class.

* The Syncer is in charge of determining what needs syncing and iterating over rows
* The Model is in charge of the sync mappings between models

*NOTE* that Appeal and Task ETL models are not in this PR. They will come in a follow-on PR.

### Testing Plan
1. Automated tests pass.
2. Run manually in UAT after deployment.

### Documentation Updates


Makefile.example
```diff
@@ -95,7 +95,7 @@ c:  ## Start rails console
 	bundle exec rails console
 
 etl-migrate:	## Migrate ETL database
-	DB=etl bundle exec rake db:migrate
+	DB=etl bundle exec rake db:migrate db:test:prepare
```
pkarman: apply migrations to test db

app/services/etl/syncer.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def target_class
+    fail "Must override abstract method target_class"
+  end
+
+  private
+
+  attr_reader :since
+
+  def incremental?
```
lomaxap: what scenarios would this be incremental?

app/models/etl/record.rb
```diff
@@ -5,4 +5,25 @@
 class ETL::Record < ApplicationRecord
   self.abstract_class = true
   establish_connection :"etl_#{Rails.env}"
+
+  class << self
+    def sync_with_original(original)
```
lomaxap: is there ever an instance where this wouldn't be overridden by the child class?

lomaxap: `organization_users`, `people`, and `organization` models?

app/services/etl/syncer.rb
pkarman: I expect we'll do an incremental (hourly?) job to keep ETL updated throughout the week, maybe a full sync from scratch on the weekends.

PR #12794, opened 2019-11-23
refactor local:destroy and local:build

tomas-nava:
### Description
Drop test & development databases in local:destroy
Reorder & remove redundant commands in local:build
add more logging to make it easier to figure out what's going on


lib/tasks/local/build.rake
```diff
@@ -3,38 +3,38 @@
 namespace :local do
   desc "build local development environment"
   task :build do
-    puts "Downloading facols image from ECR"
+    puts ">>> BEGIN local:build"
+    puts ">>> 01/08 Creating development and test caseflow databases"
+    system("RAILS_ENV=development bundle exec rake db:create") || abort
```
pkarman: I don't think this can run till after the `docker-compose up -d` has finished. Otherwise the postgresql server is not running yet.

lib/tasks/local/build.rake
```diff
@@ -3,38 +3,38 @@
...
 
       print "."
       sleep 1
     end
-    # Add a new line so that this scipt's output is more readable.
+    # Add a new line so that this script's output is more readable.
     puts ""
 
-    puts "Creating local caseflow dbs"
-    system("bundle exec rake db:create db:schema:load") || abort
```
pkarman: I think this order is important (after docker is ready).

lib/tasks/local/destroy.rake
```diff
@@ -3,11 +3,18 @@
 namespace :local do
   desc "destroy local development environment to have a clean slate when rebuilding"
   task :destroy do
-    puts "Tearing down docker volumes"
+    puts ">>> BEGIN local:destroy"
+
+    puts ">>> 01/02 Dropping development and test caseflow databases"
+    system("RAILS_ENV=development bundle exec rake db:drop") || abort
```
pkarman: good

lib/tasks/local/build.rake
```diff
@@ -3,38 +3,38 @@
...
     system("RAILS_ENV=development bundle exec rake local:vacols:seed") || abort
 
-    puts "Seeding FACOLS TEST"
+    puts ">>> 06/08 Seeding FACOLS TEST"
     system("RAILS_ENV=test bundle exec rake spec:setup_vacols") || abort
 
-    puts "Enabling feature flags"
-    system("bundle exec rails runner scripts/enable_features_dev.rb") || abort
+    puts ">>> 07/08 Loading schema and seeding local caseflow database"
+    system("RAILS_ENV=development bundle exec rake db:schema:load db:seed") || abort
```
pkarman: `db:setup` was doing both these things, so this prevents `db:seed` from running 2x.

lib/tasks/local/build.rake
tomas-nava: fixed in [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/12794/commits/41dedcd09c232d0025b1a6244876e639f1cea034).

lib/tasks/local/build.rake
```diff
@@ -3,38 +3,38 @@
...
       print "."
       sleep 1
     end
-    # Add a new line so that this scipt's output is more readable.
+    # Add a new line so that this script's output is more readable.
     puts ""
 
-    puts "Creating local caseflow dbs"
-    system("bundle exec rake db:create db:schema:load") || abort
+    puts ">>> 04/08 Creating development and test caseflow databases"
```
pkarman: I think by specifying the `RAILS_ENV` here that the "test" db is *not* going to be created.

So either need a separate command with `RAILS_ENV=test` or drop the explicit env var.

tomas-nava: I don't know if it's just me, but this command creates both databases:
```
$ RAILS_ENV=development bundle exec rake db:create
Created database 'caseflow_certification_development'
Created database 'caseflow_certification_test'
```

tomas-nava: This is [expected behavior](https://github.com/rails/rails/issues/27299#issuecomment-295855603):

> You are assuming that this behavior is a bug. It is not, it is the behavior that we (the Rails core) wants to the framework. I'm sorry that the behavior don't work for you but it is not going to change.

pkarman: TIL!

PR #12798, opened 2019-11-25
Only assignee can act on JudgeTasks

lowellrex:
Reverts change made in PR #12549 that allowed anyone to act on `JudgeTask`s. This PR still ensures that `on_hold` `JudgeTask`s are unable to be acted upon. [Related Slack thread](https://dsva.slack.com/archives/CJL810329/p1574691024052000).

### DIFF
--- a/app/models/tasks/judge_task.rb
+++ b/app/models/tasks/judge_task.rb
@@ -19,6 +19,10 @@ class JudgeTask < Task
     ].flatten
   end
 
+  def actions_allowable?(user)
+    super && assigned_to == user
+  end
+
   # :nocov:
   def additional_available_actions(_user)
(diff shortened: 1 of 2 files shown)


lowellrex:
> It may be helpful to add a comment on the JudgeTask method about why it's needed, as this was not evident to me or Tomas when I made the original change (in service of completely unrelated MTV stuff). But ✅

I'd be in favor of a code comment. Do you have any thoughts about what language we should include in the comment?

nanotone:
I'm not too familiar with the logic behind attorneys seeing or not seeing actions on JudgeTasks, so perhaps something to that effect, as well as something to explain why this extra check applies only to JudgeTasks and not tasks in general?

Issue #12725, opened 2019-11-15
Clean both databases between every test


lowellrex:
During [engineering huddle last week](https://github.com/department-of-veterans-affairs/caseflow/issues/10569#issuecomment-547026308), the team decided to explore rolling back the change to selectively clean the test databases (introduced in #11470) because we believe the selective cleaning has introduced additional "flaky" tests and not reduced the runtime of our rspec test suite in Circle CI.

## Acceptance criteria
* [ ] Determine how to measure whether cleaning the tests slows down rspec tests in Circle CI
* [ ] Change code to clean databases between each test.

lowellrex:
In order to determine whether these changes will have an effect on the running time of our rspec tests in Circle CI we need to collect some benchmarks. We collected the benchmarks by visiting the public webpage that listed the results of our `master` branch builds: https://circleci.com/gh/department-of-veterans-affairs/caseflow/tree/master?page=0. We ran the following script in the browser's Javascript console on that page and the 14 subsequent pages to collect data about each rspec build for the past month.

```js
buildRows = document.getElementsByClassName("build");
for (idx = 0; idx < buildRows.length; idx++ ) {
  let row = buildRows.item(idx);

  // Figured these paths out by walking the DOM manually.
  let buildName = row.children[2].children[1].children[1].textContent
  if (buildName === "rspec") {
    let rawDate = row.children[3].children[0].children[0].title;
    let duration = row.children[3].children[0].children[1].children[1].textContent;
    let result = row.classList[1];

    console.log(`"${rawDate}", "${duration}", "${result}"`);
  }
}
```

We pasted the results of each of these into a text editor, finessed them into a CSV format, [uploaded them to Google Docs](https://docs.google.com/spreadsheets/d/1ORnvebEnhbAPLfgA8XUYWPKALUdjGcVLu_Dpg3TWLgM/edit#gid=535542463), and created the graph below.

![image](https://user-images.githubusercontent.com/32683958/68907866-43b81080-0717-11ea-9bfa-f1e7a121ab47.png)
_Rolling 5 trial average in red_

From this we can see that our rspec builds have consistently taken ~17 minutes to run on Circle CI over the past month.

Follow up work:
* [ ] Gather data for the month before/after #11470 went live and compare.
* [ ] Review runtimes after rolling back changes in #11470.

lowellrex:
[...]

PR #12800, opened 2019-11-25
Database cleaner cleanup

pkarman:
Connects #12725 

### Description
Consolidate db cleaner config into one file, which is autoloaded by rails_helper.

If we want to make VACOLS db cleaner optional by tag in the future, we can modify this single file. The `require` and the spec tagging are separate changes.

This also adds db cleaner for the new `:etl` specs. These require explicit tagging, but they occupy a small known corner of the spec suite and should not leak across specs like FACOLS and Caseflow dbs do.


pkarman:
for posterity: remove the 2nd line of a line in-place where the second line is an extra newline:

```
pcregrep -M -rl 'frozen_string_literal: true\n\n\n'  spec | xargs -n 1 perl -i -ne 'print unless $. == 2'
```

monfresh:
@pkarman The reason for separating VACOLS and Postgres into separate files and not requiring them by default was to avoid cleaning the DB when running tests that don't use the DB at all, or tests that don't use VACOLS. Even if you decide to use tagging again, the `before(:suite)` block will still run by default before running any single test, slowing down those that don't need it by 5 seconds (as measured last July). The example I used was the specs in the `spec/mappers` folder. By not automatically requiring database_cleaner, those tests ran 448 times faster!

Another thing I didn't realize at the time as that starting with Rails 5.1, the `database_cleaner` gem is no longer needed. The functionality is built into Rails as long as you set  `config.use_transactional_fixtures = true`. I tested this with the login.gov code and it works. I'm not sure if it works with Oracle, but it's worth a try. Could you try setting `config.use_transactional_fixtures = true` in `rails_helper.rb` and commenting out `database_cleaner.rb` and see if tests fail due to DB cleanup issues? Or you could pick any test that uses VACOLS and run it twice in a row.

PR #12809, opened 2019-11-26
Aggregrate Geomatching Errors in Datadog

lomaxap:
send all geomatching errors to data dog and create a datadog alert if errors exceed X / hour


app/jobs/fetch_hearing_locations_for_veterans_job.rb
```diff
@@ -40,8 +40,8 @@ def perform
         record_geomatched_appeal(appeal.external_id, "limit_error")
         break
       rescue StandardError => error
-        capture_exception(error: error, extra: { appeal_external_id: appeal.external_id })
```
pkarman: I am reluctant to stop capturing things in Sentry. Seems like capturing in *both* places would be preferable for now?

pkarman:
we're not seeing enough sentry noise to make this necessary right now.

Issue #12813, opened 2019-11-26
v3 API - /legacy_appeal_issues spec has incorrect type for description field


edmangimelli:
`GET /legacy_appeal_issues` spec has incorrect type for the `description` field (should be an array of strings).

In discussing with @erluti, decided we should also include the `codes` and `labels` fields

PR #12814, opened 2019-11-26
v3 API - fix LegacyAppealsIssue spec

edmangimelli:
Resolves https://github.com/department-of-veterans-affairs/caseflow/issues/12813

**no code changes**


app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -1485,10 +1485,20 @@ paths:
...
+                              items:
+                                type: string
+                            labels:
+                              type: array
+                              items:
+                                type: string
                             disposition:
                               type: string
-                            close_date:
+                            closeDate:
```
erluti: Good catch!

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -1485,10 +1485,20 @@ paths:
                               type: string
                               format: date
                             description:
-                              type: string
+                              type: array
+                              items:
+                                type: string
+                            codes:
```
erluti: I feel like codes and labels might benefit from a description to explain how they connect to `description`.

PR #12815, opened 2019-11-26
API v3 - contestable issues endpoint

### LINKED ISSUE #12812: API v3 - Implement Issues endpoint
Implementation the issues endpoint based on the [spec](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/api/docs/v3/decision_reviews.yaml) and designed in this [issue](https://github.com/department-of-veterans-affairs/caseflow/issues/12194).
This endpoint will return contestable issues that could be used for creating a decision review.

## Description
A "stand-in" decesion review will be used with the ContestableIssueGenerator to find a list of issues a consumer could use in a decision review.
The returned `ContestableIssue`s will be given attributes useful for consumers and Ids necessary for creating a new decision review.

erluti:
Resolves #12812

### Description
* Introduces `contestable_issues_controller` to return ContestableIssues to consumers.
* Uses a "stand-in" decision review for the ContestableIssueGenerator to return possible issues for a new Decision Review to contest.
  * Currently only supports HigherLevelReviews with "compensation" benefit types.
* Since these don't have IDs, a serializing method is in the controller instead of using a JSONAPI serializer.

### Acceptance Criteria
- [x] returns Contestable Issues that can be used in a Decision Review


app/controllers/api/v3/decision_review/contestable_issues_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+class Api::V3::DecisionReview::ContestableIssuesController < Api::V3::BaseController
+  def index
+    # REVIEW move these before filters?
+    veteran = Veteran.find_by_file_number(request.headers['veteranId'])
+    unless veteran
+      render_error(
+        status: 404,
+        code: :veteran_not_found,
+        title: "Veteran not found"
+      )
```
edmangimelli: we have a class for pulling error data based on code (that is, returning the status and title based on the code). i'm thinking that class should be used (though maybe it should either be renamed or refactored)

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+      description: contestable_issue.description,
+      isRating: contestable_issue.is_rating,
+      latestIssuesInChain: contestable_issue.latest_contestable_issues.collect { |latest| { id: latest.decision_issue&.id, approxDecisionDate: latest.approx_decision_date } },
+      decisionIssueId: contestable_issue.decision_issue&.id,
+      ratingDecisionId: contestable_issue.rating_decision_reference_id,
+      approxDecisionDate: contestable_issue.approx_decision_date,
+      rampClaimId: contestable_issue.ramp_claim_id,
+      titleOfActiveReview: contestable_issue.title_of_active_review,
+      sourceReviewType: contestable_issue.source_review_type
+      # TODO fix based on private attributes
```
edmangimelli: could you elaborate on this

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+      # timely: contestable_issue.timely?
```
edmangimelli: was this meant to be uncommented?

edmangimelli:
looking good!
I feel like maybe there's a bit too much in the controller (that maybe should be in a service class)

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
erluti: this is WIP

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
erluti: That class is named `IntakeErrors` and this `render_error` method is a view (in all but name), that actually handles setting the response status.  Maybe this method could accept an `IntakeErrors` as an argument?  We could make a new ticket for proposed refactorings.

edmangimelli: not talking about the method (render_error), just having all of that data in on place --status, code, title. so for the convention has been to use a code to lookup title and status

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+  def set_receipt_date_from_header
+    @receipt_date = Date.parse(request.headers["receiptDate"])
+    if @receipt_date < Constants::DATES["AMA_ACTIVATION"].to_date || Time.zone.today < @receipt_date
+      render_error(
+        status: 422,
+        code: :bad_receipt_date,
+        title: "Bad receipt date"
+      )
+    end
+  end
```
edmangimelli: would a rescue be nice?
```
  def set_receipt_date_from_header
    @receipt_date = Date.parse(request.headers["receiptDate"])
    render_bad_receipt_date_error if receipt_date_is_before_ama_date? || receipt_date_is_in_the_future?
  rescue ArgumentError
    render_bad_receipt_date_error
  end

  def receipt_date_is_before_ama_date?
    @receipt_date < Constants::DATES["AMA_ACTIVATION"].to_date
  end

  def receipt_date_is_in_the_future
    Time.zone.today < @receipt_date
  end

  def render_bad_receipt_date_error
    render_error(
      status: 422,
      code: :bad_receipt_date,
      title: "Bad receipt date"
    )
  end
```

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+      approxDecisionDate: contestable_issue.approx_decision_date,
+      rampClaimId: contestable_issue.ramp_claim_id,
+      titleOfActiveReview: contestable_issue.title_of_active_review,
+      sourceReviewType: contestable_issue.source_review_type,
+      timely: contestable_issue.timely?
+    }.reject { |_, value| value.nil? }
+
+    { type: "ContestableIssue", attributes: attributes }
+  end
+end
```
edmangimelli: will this stay here, or will it (at a later time) be moved somewhere else?

spec/requests/api/v3/decision_review/contestable_issues_controller_spec.rb
```diff
@@ -0,0 +1,221 @@
...
+  after { FeatureToggle.disable!(:api_v3) }
+
+  describe "#index" do
+    let(:veteran) { create(:veteran) }
+
+    let!(:api_key) do
+      ApiKey.create!(consumer_name: "ApiV3 Test Consumer").key_string
+    end
+
+    def get_issues(veteran_id: veteran.file_number, receipt_date: Time.zone.today)
```
edmangimelli: i thought code climate threw a fit when we did this (define a method in a test)

spec/requests/api/v3/decision_review/contestable_issues_controller_spec.rb
```diff
@@ -0,0 +1,221 @@
...
+    it "should return a list of issues" do
+      Generators::Rating.build(
+        participant_id: veteran.ptcpnt_id,
+        profile_date: Time.zone.today - 10.days # must be before receipt_date
+      ) # this is a contestable_rating_issues
+      get_issues
+      issues = JSON.parse(response.body)
+      expect(issues).to be_an Array
+      expect(issues.count > 0).to be true
+    end
```
edmangimelli: it's really nice to see examples of how to create some of this fake data. nice job

edmangimelli: should `it "should return a 200 OK"` and `it "should return a list of issues"` test that the array returned has changed from the first test to the second? (like maybe just the count)

spec/requests/api/v3/decision_review/contestable_issues_controller_spec.rb
```diff
@@ -0,0 +1,221 @@
...
+                dgnstc_tc: "123456"
+              }
+            }
+          ],
+          profile_date: date - 10.days # must be before receipt_date
+        ) # this is a contestable_rating_issues
+        another_decision_issue # instantiate this
+        get_issues
+        issues = JSON.parse(response.body)
+        expect(issues.count > 0).to be true
```
edmangimelli:
```suggestion
        expect(issues).not_to be_empty
```

edmangimelli: also, didn't know you could throw an `expect` in a `let`

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
edmangimelli: i like how dead simple this is --easier to read than the serializers we've made

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+      isRating: contestable_issue.is_rating,
+      latestIssuesInChain: latest_issues_in_chain,
+      decisionIssueId: contestable_issue.decision_issue&.id,
+      ratingDecisionId: contestable_issue.rating_decision_reference_id,
+      approxDecisionDate: contestable_issue.approx_decision_date,
+      rampClaimId: contestable_issue.ramp_claim_id,
+      titleOfActiveReview: contestable_issue.title_of_active_review,
+      sourceReviewType: contestable_issue.source_review_type,
+      timely: contestable_issue.timely?
+    }.reject { |_, value| value.nil? }
```
edmangimelli:
```suggestion
    }.compact
```
or is there something I'm not thinking of?

spec/requests/api/v3/decision_review/contestable_issues_controller_spec.rb
```diff
@@ -0,0 +1,221 @@
...
+        issues.each do |issue|
+          expect(issue["attributes"].keys).to include("isRating")
+          expect(issue["attributes"]["isRating"]).to be_in([true, false])
+        end
+      end
+      it "should have latestIssuesInChain attribute" do
+        issues.each do |issue|
+          expect(issue["attributes"].keys).to include("latestIssuesInChain")
+          expect(issue["attributes"]["latestIssuesInChain"]).to be_a Array
+          expect(issue["attributes"]["latestIssuesInChain"].first.keys).to include("id", "approxDecisionDate")
```
edmangimelli: should we test `.all?`

spec/requests/api/v3/decision_review/contestable_issues_controller_spec.rb
```diff
@@ -0,0 +1,221 @@
...
+        end
+      end
+      it "should have decisionIssueId attribute" do
+        issue_with_decision_issue = issues.find { |i| i["attributes"].key?("decisionIssueId") }
+        expect(issue_with_decision_issue).to be_present
+        expect(issue_with_decision_issue["attributes"]["decisionIssueId"]).to be_a Integer
+      end
+      context "with contestable rating decisions enabled" do
+        before { FeatureToggle.enable!(:contestable_rating_decisions) }
+        after { FeatureToggle.disable!(:contestable_rating_decisions) }
```
edmangimelli: interesting. nice to know

edmangimelli:
love it.
It's also really cool to see how this stuff works (in the big spec)

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
edmangimelli: (because the parse might fail with bad input, and "Bad receipt date" would be a more descriptive error than probably whatever would be returned otherwise)

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
edmangimelli: (wait --did i already ask this?)

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
erluti: i wasn't thinking of compact, thanks!  (when I wrote this I was thinking it was weirdly unfamiliar, lol)

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+        status: 404,
+        code: :veteran_not_found,
+        title: "Veteran not found"
+      )
+    end
+  end
+
+  def set_receipt_date_from_header
+    @receipt_date = Date.parse(request.headers["receiptDate"])
+    if @receipt_date < Constants::DATES["AMA_ACTIVATION"].to_date || Time.zone.today < @receipt_date
```
pkarman: there is other logic for date evaluation in `DecisionReview.ama_activation_date` -- should use that existing method instead on the `standin_claim_review`

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+      render_error(
+        status: 422,
+        code: :bad_receipt_date,
+        title: "Bad receipt date"
+      )
+    end
+  end
+
+  def contestable_issue_data(contestable_issue)
```
pkarman: this method seems like it should be in its own class, since it is for serialization.

spec/requests/api/v3/decision_review/contestable_issues_controller_spec.rb
pkarman: should be fine to create a method in a test.

spec/requests/api/v3/decision_review/contestable_issues_controller_spec.rb
```diff
@@ -0,0 +1,221 @@
...
+          ],
+          profile_date: date - 10.days # must be before receipt_date
+        ) # this is a contestable_rating_issues
+        another_decision_issue # instantiate this
+        get_issues
+        issues = JSON.parse(response.body)
+        expect(issues.count > 0).to be true
+        issues
+      end
+      it "should have ratingIssueId attribute" do
```
pkarman: one line of whitespace between `it` blocks please.

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+    # currently only supporting HLR for "compensation"; going to have to do this for each type
+    standin_claim_review = HigherLevelReview.new(
+      veteran_file_number: @veteran.file_number,
+      receipt_date: @receipt_date,
+      # must be in ClaimantValidator::BENEFIT_TYPE_REQUIRES_PAYEE_CODE for can_contest_rating_issues?
+      benefit_type: "compensation"
+    )
+    issues = ContestableIssueGenerator.new(standin_claim_review).contestable_issues
+    # this generates this error in UAT:
+    #  BGS::PublicError (Logon ID APIUSER Not Found in the Benefits Gateway Service (BGS). Contact your
```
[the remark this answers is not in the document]
erluti: I made this comment when I was SSHed into the UAT box and testing out if this approach made sense.  I've recently discovered that the base controller we're using sets the user made in BGS calls with this method https://github.com/department-of-veterans-affairs/caseflow/blob/068fe87c3d24906f2c3d416796ff2a9261423760/app/controllers/api/application_controller.rb#L31
Does that have any impact on our discussion around the username?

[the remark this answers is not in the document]
pkarman: yes and no. The `system_user` should be able to read all the things (like when a Veteran wants to check their Appeal status) but I don't know yet if we want it to write all the things (as for this new API).

But here, for reading the list of BGS ratings, the `system_user` should be used since that is already set up to work with BGS.

spec/requests/api/v3/decision_review/contestable_issues_controller_spec.rb
erluti: good change!  but I'm thinking about taking it out since it is unexpected

spec/requests/api/v3/decision_review/contestable_issues_controller_spec.rb
erluti: they're separate tests. for 200 OK I don't care how many results come back.  For the other one I care that it's an array of data

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
erluti: going to extract this into a serializer class per Peter's suggestion, but I'm going to keep this simple I think

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
erluti: So it looks like rails is noticing the ISO8601 format and automatically making it a `Date`.  I added a test for non-iso8601 formats and took this rescue approach, but when I realized it was already transforming it, along with some other refactoring changes I ended up not needing it.

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+class Api::V3::DecisionReview::ContestableIssuesController < Api::V3::BaseController
+  before_action :set_veteran_from_header
+
+  def index
+    @receipt_date = request.headers["receiptDate"]
+    if receipt_date_is_bad
```
pkarman: idiomatically I would expect this method to be named `bad_receipt_date?` (or more charitably, `invalid_receipt_date?`)

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+      bad_receipt_date
```
pkarman: maybe `render_invalid_receipt_date` would be more clear what this method is doing.

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def receipt_date_is_bad
+    !@receipt_date.is_a?(Date) ||
+      @receipt_date < standin_claim_review.ama_activation_date ||
+      Time.zone.today < @receipt_date
+  end
+
+  def set_veteran_from_header
+    @veteran = Veteran.find_by_file_number(request.headers["veteranId"])
```
pkarman: you probably want `VeteranFinder.find_best_match(request.headers["veteranId"])` instead. If this is the first time Caseflow has seen this veteran, then `find_by_file_number` will return nil.

app/serializers/api/v3/contestable_issue_serializer.rb
```diff
@@ -0,0 +1,37 @@
...
+      isRating: contestable_issue.is_rating,
+      latestIssuesInChain: latest_issues_in_chain,
+      decisionIssueId: contestable_issue.decision_issue&.id,
+      ratingDecisionId: contestable_issue.rating_decision_reference_id,
+      approxDecisionDate: contestable_issue.approx_decision_date,
+      rampClaimId: contestable_issue.ramp_claim_id,
+      titleOfActiveReview: contestable_issue.title_of_active_review,
+      sourceReviewType: contestable_issue.source_review_type,
+      timely: contestable_issue.timely?
+    }.compact
```
pkarman: do we really want `.compact` here? I would think `nil` values are meaningful?

spec/requests/api/v3/decision_review/contestable_issues_controller_spec.rb
```diff
@@ -0,0 +1,247 @@
...
+      expect(issues).to be_an Array
+      expect(issues.count).to be > 0
+    end
+
+    context "returned issues" do
+      let(:source) { create(:higher_level_review, veteran_file_number: veteran.file_number, same_office: false) }
+      let(:claim_id) { "12345" }
+      let(:rating_issue_reference_id) { "99999" }
+      let(:end_product_establishment) do
+        EndProductEstablishment.new(
```
pkarman: why not use the factory here? it will also create the EndProduct to match.

spec/requests/api/v3/decision_review/contestable_issues_controller_spec.rb
```diff
@@ -0,0 +1,247 @@
...
+      expect(response).to have_http_status(:not_found)
+    end
+
+    it "should return a 422 when the receipt is before AMA" do
+      get_issues(receipt_date: Time.zone.today - 1000.years)
+      expect(response).to have_http_status(:unprocessable_entity)
+    end
+
+    it "should return a 422 when the receipt date after today" do
+      get_issues(receipt_date: Time.zone.today + 1.year)
```
pkarman: or just `Time.zone.tomorrow`

app/controllers/api/v3/decision_review/contestable_issues_controller.rb
erluti: i went back and forth on the name, i'll use your suggestion

app/serializers/api/v3/contestable_issue_serializer.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class Api::V3::ContestableIssueSerializer
+  def initialize(contestable_issues)
+    @contestable_issues = contestable_issues
+  end
+
+  def to_json(_ = {})
```
pkarman: do you need to set a default value if ignoring the param? mostly curious.

erluti: I don't!  I'll pull that.  I was just learning that it needed to accept a hash of opts to work with `render json:`

edmangimelli:
Lg!

PR #12816, opened 2019-11-26
ETL sync appeals tasks

pkarman:
connects #12659

### Description
Follow on to #12788 for Appeal and Task models.


app/models/etl/record.rb
```diff
@@ -2,15 +2,26 @@
...
+# "target" == ETL db record
+#
+# Note about schema Rails automatic meta columns:
+#  * if the table is a 1:1 mirror:
+#    * the "id" is the same in original and target
+#    * the "created_at" and "updated_at" are the same on original and target
+#  * if the table is transformed in any way:
+#    * the "id" of the original is mapped to ":original_id" (e.g. appeal_id)
+#    * the "created_at" may be the same on original and target, if not, there is a ":original_created_at"
+#    * the "updated_at" should be modified every time target (ETL) is modified
```
alisan16: Can you clarify this last comment for “updated_at”? I don’t quite understand how the original and target updated_at fields will differ when the table has been transformed based on the explanation here.

PR #12819, opened 2019-11-26
Appeal Status Decorator

pkarman:
### Description
Refactor only.

The [Rails Decorator pattern](https://www.thegreatcodeadventure.com/rails-refactoring-part-iii-the-decorator-pattern/) allows for extending model behavior for specific presentations.

In this case, the Appeal model had several methods designed for the Appeals Status API. This PR pulls those out into a separate class.


app/models/appeal_events.rb
```diff
@@ -16,6 +16,8 @@ def all
       hlr_events
     when Appeal
       appeal_events
+    when AppealStatusDecorator
+      appeal_events
     end
   end
```
lomky: Optional: line comment explaining this model would be nice!

app/models/api_status_alerts.rb
```diff
@@ -35,14 +37,16 @@ def post_decision
     return unless Time.zone.today <= decision_review.due_date_to_appeal_decision
     return if decision_review.is_a?(Appeal) && Time.zone.today > decision_review.cavc_due_date
 
+    is_appeal = decision_review.is_a?(AppealStatusDecorator) || decision_review.is_a?(Appeal)
```
ferristseng: Does the line above this also need to do a similar check? Could this logic be in a method on the appeal?

Issue #12241, opened 2019-09-27
Certification: On Matching Dates Page, invisible cancel button shares focus with the Continue or Refresh buttons


alisan16:
On the Matching Dates page an invisible cancel button and the Continue button both have a visual indication of focus. Only one element should have a visual indication of focus (the dotted line or the black outline). (New: 08/22/2019)

<img width="754" alt="Screen Shot 2019-09-27 at 5 08 43 PM" src="https://user-images.githubusercontent.com/1198106/65802433-a5152780-e149-11e9-91b3-b7ae588ae053.png">

jcq:
### Updated Screenshot
<img width="1063" alt="Screen Shot 2019-11-26 at 15 49 51" src="https://user-images.githubusercontent.com/2581274/69681908-71c91900-1064-11ea-893c-f749abbade6e.png">

PR #12821, opened 2019-11-26
Fix Layout of Certification Footer (Fixes Mult Focused Element Issue)

jcq:
Connects #12241

### Description
This reworks the certification/Footer component to avoid nesting a <button> within an anchor ("a") tag. In doing so, we avoid case where both button and link are focused at the same time. 

### Acceptance Criteria
- [ ] Only one element is focused when "Continue" button has focus


jcq:
### Updated Screenshot
Screen Shot 2019-11-26 at 15 49 51

PR #12823, opened 2019-11-27
resend emails when user changes virtual hearin…

### LINKED ISSUE #12761: Resend Emails When Hearing Time Changes
We need to resend emails after the hearing time changes on the daily docket.
Once #12753 is merged, [check for `scheduled_time_string` changes in hearing updates](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/hearings/forms/base_hearing_update_form.rb#L38) when evaluating `email_sent_flag`.

### AC
- [ ] Emails are sent to veteran, representative and/or judge when hearing time changes

rubaiyat22:
Resolves #12761 
- reset *_email_sent_flags and kick off async job
- modify COPY to send alert 
- disable TimeRadioButtons until async job is completed

Screen Shot 2019-11-27 at 6 11 20 PM


client/app/hearings/components/dailyDocket/DailyDocketRow.jsx
```diff
@@ -195,12 +195,14 @@ class DailyDocketRow extends React.Component {
...
       <HearingLocationDropdown {...inputProps} regionalOffice={regionalOffice} />
       <StaticHearingDay hearing={hearing} />
-      <TimeRadioButtons {...inputProps} regionalOffice={regionalOffice} update={(values) => {
-        this.update(values);
-        if (values.scheduledTimeString !== null) {
-          this.openVirtualHearingModal();
-        }
-      }} />
+      <TimeRadioButtons {...inputProps} regionalOffice={regionalOffice}
+        readOnly={!hearing.virtualHearing.jobCompleted}
```
ferristseng: Might also need to check if there is a virtual hearing -- getting an error with this line when there is no virtual hearing associated with the hearing.

Issue #12793, opened 2019-11-23
MTV - add Straight Vacate option to flow


leikkisa:
## User or job story
As a judge user, I would like to have the option to vacate decision issues without creating new decisions on them.

## Acceptance criteria
- [ ] Please put this work behind the feature toggle: [review_motion_to_vacate]
- [ ] Update copy and code to rename Straight Vacate and Readjudication to Vacate and Readjudication
- [ ] Add a Straight Vacate option to the MTVJudgeDisposition modal
- [ ] When submitted, this creates a post decision motion with the new "straight_vacate" type
- [ ] When submitted, a StraightVacateTask is created (following current structure, as DecidedMotionToVacateTask with parent org task, and assigned to the drafting attorney)
- [ ] Existing pattern is followed for partial/full grant and saving issue ids
- [ ] Include screenshot(s) in the Github issue if there are front-end changes
- [ ] Add topline comments for any new files
       - Are there examples of these; best practices documented?

## Release notes
When a judge is deciding a motion to vacate, they will have the option to Straight Vacate, which will not result in new decisions.

### Out of scope
- The attorney's next step
- Creating any issues

### Designs
![Screen Shot 2019-11-22 at 4.53.29 PM.png](https://images.zenhubusercontent.com/58f6b6eb51aa2c30577bb60e/dad39caf-c8d3-44b5-86c0-2df2d2b037d1)

### Background/context
We learned about a new MTV flow. We believe this may be to vacate decision issues that happened in AMA, when the issue was still being processed in the legacy system (and therefore getting decided in the legacy system).

### Technical notes
Some things will need to be renamed - files, class names, variables, in order to update StraightVacateAndReadjudicate to just VacateAndReadjudicate.

jcq:
## Updated Screen Shot
<img width="557" alt="Screen Shot 2019-11-27 at 16 27 24" src="https://user-images.githubusercontent.com/2581274/69767821-e5d1f280-1132-11ea-9e5d-a245a640bb30.png">

PR #12827, opened 2019-11-28
Add StraightVacateTask & rename existing to VacateAndReadjudicationTask

jcq:
Connects #12793

### Description
Rename StraightVacateAndReadjudication to VacateAndReadjudication, and implement new StraightVacate task

### Acceptance Criteria
- [ ] Code compiles correctly


jcq:
## Updated Screen Shot
Screen Shot 2019-11-27 at 16 27 24

app/models/tasks/straight_vacate_task.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class StraightVacateTask < DecidedMotionToVacateTask
+  def self.label
+    COPY::STRAIGHT_VACATE_TASK_LABEL
+  end
+
+  def self.org(user)
+    JudgeTeam.for_judge(user.reload)
+  end
```
tomas-nava: Not a blocker, but [we're trying to standardize class method definition](https://github.com/department-of-veterans-affairs/caseflow/issues/11778) using `class << self` blocks instead of `def self.method_name`.

app/models/tasks/straight_vacate_task.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class StraightVacateTask < DecidedMotionToVacateTask
+  def self.label
+    COPY::STRAIGHT_VACATE_TASK_LABEL
+  end
+
+  def self.org(user)
+    JudgeTeam.for_judge(user.reload)
```
tomas-nava: We're assuming here (and elsewhere) that `user` will be a judge who's an admin on a `JudgeTeam`, but it looks like [we get possible users by fetching a list of judges from VACOLS](https://github.com/department-of-veterans-affairs/caseflow/blob/jc/12793-mtv-straight-vacate/app/repositories/task_action_repository.rb#L155). These two aren't always the same, as explained in #8436. It's out of scope for this PR to fix the issue, but maybe we could have a more deliberate error for cases where `JudgeTeam.for_judge(user.reload)` returns `nil`. We'll get an error anyway, because the new task [will fail validation](https://github.com/department-of-veterans-affairs/caseflow/blob/jc/12793-mtv-straight-vacate/app/models/task.rb#L24), but a custom error that highlights the issue here might be useful.

app/models/tasks/straight_vacate_task.rb
jcq: No prob — updated those!

spec/feature/queue/motion_to_vacate_spec.rb
```diff
@@ -182,10 +182,44 @@
...
+        disposition: "granted",
+        vacate_type: "vacate_and_readjudication"
+      )
+      new_task = VacateAndReadjudicationTask.find_by(assigned_to: drafting_attorney)
+      expect(new_task).to_not be_nil
+      expect(new_task.label).to eq COPY::VACATE_AND_READJUDICATION_TASK_LABEL
+      expect(new_task.available_actions(motions_attorney)).to include(
+        Constants.TASK_ACTIONS.LIT_SUPPORT_PULAC_CERULLO.to_h
+      )
+      expect(new_task.instructions.join("")).to eq(instructions)
```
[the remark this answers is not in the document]
jcq: It seems like we could navigate back in to the case details page and check the task type, but @leikkisa and I thought it probably makes more sense to do the verification via the DB records since it's faster, and we could test the display of task types in some future PR that might be more design-focused.

app/models/tasks/straight_vacate_task.rb
jcq: I added some code and a new `NonexistantJudgeTeam` error; let me know if you think that helps with this issue.

jcq:
@tomas-nava Alrighty, I think I addressed all of the comments in replies. I also updated the schema to change the comment text — that was a good flag for completeness.

lib/caseflow/error.rb
```diff
@@ -203,6 +203,12 @@ def initialize(args)
     end
   end
 
+  class NonexistantJudgeTeam < StandardError
```
tomas-nava: I'm sorry to report that it's spelled `Nonexistent`

spec/controllers/post_decision_motions_controller_spec.rb
```diff
@@ -58,7 +61,7 @@ def create_task_without_unnecessary_models
       :judge_address_motion_to_vacate_task,
       appeal: appeal,
       parent: parent,
-      assigned_to: user,
+      assigned_to: judge,
       assigned_by: assigned_by
     )
   end
```
tomas-nava: Could we get a small test that expects the `NonexistentJudgeTeam` error?

lib/caseflow/error.rb
jcq: Yikes, that's an embarrassing typo. Glad you noticed it!

spec/controllers/post_decision_motions_controller_spec.rb
jcq: Yup!

PR #12837, opened 2019-12-02
Add instructions when bulk reassigning tasks

### LINKED ISSUE #12506: Add instructions to reassigned tasks of inactive users
Spun out of #12483.

[Some slack chatter.](https://dsva.slack.com/archives/CJL810329/p1572369656018000?thread_ts=1572295303.028100&cid=CJL810329)

We should be documenting reasoning why we are cancelling or reassigning tasks of inactive users so that the history exists in the database. An additional benefit of doing this is so that users can see why they are receiving a task or why downstream children were cancelled.

### AC
- [ ] When a user's tasks are reassigned or cancelled in the reassign rake task, add instructions as to why the task was created, reassigned, or cancelled

hschallhorn:
Resolves #12506

### Description
Adds a descriptive instructions to tasks that have been assigned, reassigned, or cancelled due to bulk task reassignment.

### Acceptance Criteria
- [ ] Assigned, reassigned, or cancelled tasks due to bulk task reassignment include instructions on why this occurred


app/workflows/bulk_task_reassignment.rb
```diff
@@ -198,18 +206,26 @@ def reassign_judge_review_tasks
...
       end
     end
   end
 
+  def reassign_judge_review_task(task, atty_task)
+    reassigned_tasks = task.reassign(
+      {
+        assigned_to_type: User.name,
+        assigned_to_id: new_supervising_judge_from_task(atty_task).id,
+        instructions: reassignment_instructions("reassigned")
```
kevmo: `Constants.TASK_STATUSES.reassigned` :)

app/workflows/bulk_task_reassignment.rb
```diff
@@ -71,6 +71,14 @@ def new_supervising_judge_from_task(task)
     assignee_judge_teams.where.not(name: user.css_id).first&.judge
   end
 
+  def reassignment_instructions(status)
+    format(COPY::BULK_REASSIGN_INSTRUCTIONS, status, user.css_id)
+  end
+
+  def update_task_status_with_instructions(task, status)
+    task.update_with_instructions(status: status, instructions: reassignment_instructions(status))
```
kevmo: Musings:

I don't love how much `status` there is in this function call.  It initially seems like  `Task#update_with_instructions` should be refactored, but
```
  def update_with_instructions(params)
    params[:instructions] = flattened_instructions(params)
    update!(params)
  end
```
is OK code (and pre-dates this PR).  I guess we could argue that `update_with_instructions` should be deleted and `flattened_instructions` to be called elsewhere, but that is beyond scope of this PR, so... Do with these thoughts what you will!

app/workflows/bulk_task_reassignment.rb
```diff
@@ -198,18 +206,26 @@ def reassign_judge_review_tasks
...
-
-          # Move the attorney task under the new JudgeDecisionReviewTask assigned to the new judge
-          atty_task.update!(parent: reassigned_tasks.first)
+          reassign_judge_review_task(task, atty_task)
         end
       end
     end
   end
 
+  def reassign_judge_review_task(task, atty_task)
```
kevmo: Nitpick: I would rename that `attorney_task`

app/workflows/bulk_task_reassignment.rb
hschallhorn: Eh, I don't know if I agree with this. Reassigned isn't exactly a task status, just the verb we want to use for these instructions.

app/workflows/bulk_task_reassignment.rb
hschallhorn: > predates this PR
Was spun out of this PR 😄 
I am also not a huge fan of these functions, but I've been playing with and ruminating over this ticket of like 2 weeks, and this is the cleanest I've got. So I'm open to suggestions. Was trying to cut down on a bunch of repeating code.

kevmo: I say merge this, then, and make a tech debt ticket.

app/workflows/bulk_task_reassignment.rb
kevmo: Oh yeah... great point!

kevmo: OK, i figured out what I should have said: I would add this string to just be part of `COPY::BULK_REASSIGN_INSTRUCTIONS`.

Issue #12836, opened 2019-12-02
Bug: React Ref Bug in EvaluateDecisionView


jcq:
## Description
If the form is submitted when complexity of case is selected but "quality of work" is null (no radio selected), this error is tossed:
`Argument appears to not be a ReactComponent. Keys: current`

## Background/context/resources
- [Sentry alert](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/6477/)

## Technical notes
This appears to be due to an issue with refs not being passed or initialized correctly for that component.

PR #12838, opened 2019-12-03
Bug Fix: React refs in EvaluateDecisionView & JudgeCaseQuality

jcq:
Connects #12836

### Description
Fixes bug noticed in this [Sentry alert](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/6477/)

### Acceptance Criteria
- [ ] Code compiles correctly
- [ ] If submit button is pressed when required fields aren't filled, it properly scrolls to field and/or alert
- [ ] No errors in JS console


client/app/queue/caseEvaluation/JudgeCaseQuality.jsx
```diff
@@ -114,17 +123,28 @@ export const JudgeCaseQuality = ({ highlight = false, qualityAlertRef = null, on
...
-          Please provide more details about <b>quality of work</b>.{' '}
-          If none of these apply to this case, please share{' '}
-          <b>additional comments</b> below.
+        <Alert
+          ref={qualityAlertRef}
+          type="info"
+          scrollOnAlert={false}
+          styling={qualityOfWorkAlertStyling}
+        >
+          Please provide more details about <b>quality of work</b>. If none of
```
tomas-nava: do you need a `{' '}` at the end of this line?

client/app/queue/caseEvaluation/EvaluateDecisionView.jsx
```diff
@@ -79,20 +80,20 @@ class EvaluateDecisionView extends React.PureComponent {
...
-      this.scrollTo(this.qualityLabel);
+      this.scrollTo(this.qualityLabel.current);
 
       return false;
     }
 
     // eslint-disable-next-line camelcase
     if (qualityIsDeficient(this.state.quality) && _.every([areas_for_improvement, factors_not_considered], _.isEmpty)) {
-      this.scrollTo(this.deficientQualityAlert);
+      this.scrollTo(this.deficientQualityAlert.current);
```
leikkisa: I wonder if it would make sense to rename this to qualityAlert here and above on line 59. I notice it was a little bit of a mix before and you renamed it below.  I don't have a full view of how the two files relate to each other or what's going on so no strong feelings on this.

jcq: I'm a bit torn, as it does indicate that the reason for the alert is that we're below a certain level of quality. However, I like shorter variable names, and that fits better with the naming scheme of the other things. I've gone ahead and updated.

client/app/queue/caseEvaluation/JudgeCaseQuality.jsx
jcq: I think that got automatically added when that line got auto-formatted due to being too long. However, it's not too long now, so we can remove that and just pull the single dangling word up. Looks cleaner that way.

Issue #12733, opened 2019-11-16
forgo RequestIssues abstraction - v3 api


edmangimelli:
After discussing the importance of the shape of RequestIssues with @leikkisa, @erluti and I, after more discussion, decided that a RequestIssue is _not_ a useful abstraction for our API consumers. This abstraction, however well-suited it may be for storing data on what is being contested on a decision review, seems to us to be unintuitive and unnecessary for an API consumer. Our proposal is instead to treat HLRs as collections of ContestableIssues, and, when wanting to associate a legacy issue (vacols_sequence_id), append that info as a new field on the contestable issue.

so, instead of:
```
# array of request issues:
[
  {
    decision_issue_id: 1,
    notes: "Hello",
    vacols_id: 2,
    vacols_sequence_id: 3,
  },
  {
    rating_issue_id: 4,
    notes: "Goodbye",
  }
]
```
this:
```
# array of contestable issues:
[
  {
    decisionIssueId: 1,
    notes: "Hello",
    legacyAppealIssues: [
      {
        legacyAppealId: 2,
        legacyAppealIssueId: 3,
      }
    ]
  },
  {
    ratingIssueId: 4,
    notes: "Goodbye",
  },
  ...
]
```

## Motivations

A big motivation was allowing consumers to use ContestableIssues returned from the /contestable_issues endpoint immediately, without modification.
How different is adding a legacyAppealIssues array as opposed to just adding a legacyAppealId and legacyAppealIssueId field to an object --not huge, but 1) we know from Sally that soon one can associate multiple legacy issues to one contestable issue (the array was unavoidable), 2) not modifying the ContestableIssue object at all seemed nice, but we also didn't want too much nesting:
```
# an idea for a request issue
{
  contestableIssue: {...},
  legacyAppealIssues: [...],
}
```

PR #12839, opened 2019-12-03
v3 api - update post HLR endpoint for spec changes (less classes)

edmangimelli:
Removes the RequestIssue abstraction from the POST HLR endpoint ([issue #12733](https://github.com/department-of-veterans-affairs/caseflow/issues/12733)) and removes the ability to associate legacy appeals.

The meat of this PR is these 4 files:

[contestable_issue_params.rb](https://github.com/department-of-veterans-affairs/caseflow/blob/v3-api-update-post-endpoint-for-spec-changes/app/services/api/v3/decision_review/contestable_issue_params.rb):  replaces `request_issue_params.rb`

[contestable_issue_finder.rb](https://github.com/department-of-veterans-affairs/caseflow/blob/v3-api-update-post-endpoint-for-spec-changes/app/services/api/v3/decision_review/contestable_issue_finder.rb)

[higher_level_review_intake_params.rb](https://github.com/department-of-veterans-affairs/caseflow/blob/v3-api-update-post-endpoint-for-spec-changes/app/services/api/v3/decision_review/higher_level_review_intake_params.rb):  fixed for new abstraction (`ContestableIssueParams`) and refactored

[hash_path_validator.rb](https://github.com/department-of-veterans-affairs/caseflow/blob/v3-api-update-post-endpoint-for-spec-changes/app/services/api/v3/decision_review/hash_path_validator.rb):  a tiny helper service class for validating a *single* path of a hash (allows us to [describe all the excepted types very succinctly](https://github.com/department-of-veterans-affairs/caseflow/blob/v3-api-update-post-endpoint-for-spec-changes/app/services/api/v3/decision_review/intake_params.rb#L143-L252) while providing a useful error message when the shape is wrong


app/services/api/v3/decision_review/intake_params.rb
```diff
@@ -41,149 +40,198 @@ def complete_params
...
 
-  def claimant_payee_code
-    claimant && claimant[:data][:meta][:payeeCode]
+  def describe_shape_error
+    "payload must be an object" unless @params.respond_to?(:dig)
+  
+    types_and_paths.find do |(types, path)|
+      error_msg = DigError.new(hash: @params, path: path, values: types).to_s
+      break error_msg if error_msg
+      false
```
erluti: this could be an `each` loop where you `return error_msg if error_msg` instead of using find

edmangimelli: (i thought i already responded to this? must've been different PR)

code climate doesn't like early returning either from `each` or `find` as I recall, and I agree, it seems like `find` is more appropriate (semantically) for what I'm trying to do

app/services/api/v3/decision_review/contestable_issue_params.rb
```diff
@@ -0,0 +1,125 @@
...
+  CATEGORIES_BY_BENEFIT_TYPE = Constants::ISSUE_CATEGORIES.slice("compensation")
+
+  attr_reader :error_code
+
+  def initialize(params:, benefit_type:, legacy_opt_in_approved:)
+    @attributes = params[:attributes].slice(*PERMITTED_KEYS)
+    # ^^^ params shape / types of fields is mostly checked in IntakeParams (not here)
+    @benefit_type = benefit_type
+    @legacy_opt_in_approved = legacy_opt_in_approved
+    @error_code = determine_error_code
```
erluti: this is a large amount of work that this class is doing.  Maybe we can move this to a public method for testing; perhaps it could be named `validate` or something like that

app/services/api/v3/decision_review/hash_path_validator.rb
```diff
@@ -0,0 +1,112 @@
...
+  end
+
+  # may throw exception --use path_is_valid? first
+  def dig
+    hash.dig(*path)
+  end
+
+  def path_is_valid?
+    @path_is_valid ||= begin
+                       allowed_values.any? { |av| av === dig }
```
erluti: indentation looks weird on github

edmangimelli: i'll see if rubocop changes it up if i wrap it with parentheses

app/services/api/v3/decision_review/contestable_issue_params.rb
edmangimelli: fwiw, i went ahead and tested the private methods of that class (for this reason)

app/services/api/v3/decision_review/intake_params.rb
erluti: find usually returns an element from the Enumerable you are running find on, and this is returning a new object.
I'm surprised at code climate, especially since `break` is returning out of the loop.  Sending arguments via break does make it seem less like an `each`.

if `to_s` is doing the work, then maybe it could be something like this?
```
error = types_and_paths.collect { |(types, path)| DigError.new(hash: @params, path: path, values: types) }.find { |digger| !digger.to_s.nil? }
```

app/services/api/v3/decision_review/contestable_issue_params.rb
erluti: i didn't mean just for testing, but also for clarity when using ContestableIssueParams, you can see that validating is an important process

app/services/api/v3/decision_review/intake_params.rb
edmangimelli: >find usually returns an element from the Enumerable you are running find on, and this is returning a new object.

correct this, is overriding that behavior

edmangimelli: I'm pretty far from a premature optimization person, but _not short-circuiting_* this feels kind of wrong

*_transforming the entire array, and then using find to short-circuit_

edmangimelli: Here's where this method currently stands (with the unintuitive `find`):
```
def describe_shape_error
  return "payload must be an object" unless @params.respond_to?(:dig)

  types_and_paths.find do |(types, path)|
    validator = Api::V3::DecisionReview::HashPathValidator.new(hash: @params, path: path, allowed_values: types)
    break validator.error_msg if !validator.path_is_valid?

    false
  end
end
```

What about this compromise?:
```
def describe_shape_error
  return "payload must be an object" unless @params.respond_to?(:dig)

  # find the first [types, path] combo that doesn't work (if any)
  bad_type_and_path_combo = types_and_paths.find do |(types, path)|
    !Api::V3::DecisionReview::HashPathValidator.new(
      hash: @params,
      path: path,
      allowed_values: types
    ).path_is_valid?
  end

  return nil unless bad_type_and_path_combo

  types, path = bad_type_and_path_combo

  Api::V3::DecisionReview::HashPathValidator.new(
    hash: @params,
    path: path,
    allowed_values: types
  ).error_msg
end
```

app/services/api/v3/decision_review/contestable_issue_params.rb
edmangimelli: excluding comments, this class is only about 100 lines, which i thought was a pretty good side.
It _does_ cover everything related to ContestableIssueParams, which amounts to 2 things:
1) validation of the params input
2) transformation of a ContestableIssue (+any associated LegacyIssues) into a RequestIssue

You're saying it would be better if that was 2 classes?

app/services/api/v3/decision_review/intake_params.rb
edmangimelli: Wait, I was thinking that making a new HPV was an expensive operation, but it's not — it doesn’t do anything (validating) until you tell it to.

Yours is better. I’ll change it up

edmangimelli: went with map, but ended up with something a little different. let me know

app/services/api/v3/decision_review/contestable_issue_params.rb
erluti: I meant `determine_error_code` is a lot of what this class is doing.  The class is a good size.  I was trying to say there's two advantages to making `determine_error_code` into a public method.
1. tests are easier (no need to `send` or anything)
2. using the class is clearer (below is some theoretical code, and i like it because it's more explicit with what it's doing.)
```
cip = Api::V3::DecisionReview::ContestableIssueParams.new(...)
return cip.error_message unless cip.validate
use_params(cip.intakes_controller_params
```

app/services/api/v3/decision_review/intake_params.rb
erluti: nice!

app/services/api/v3/decision_review/contestable_issue_params.rb
pkarman: if you must test a method, it should be public. Using `send()` is definitely a code smell.

pkarman: I think this should be 2 classes. `ContestableIssueParams` and `ContestableIssueParamsValidator`.

app/services/api/v3/decision_review/hash_path_validator.rb
```diff
@@ -0,0 +1,116 @@
...
+#       path: [:a, :b],
+#       allowed_values: [String, Integer]
+#     )
+#
+#     v.path_is_valid? # false
+#
+#     puts v.error_msg
+#     # [:a][:b] should be one of [String, Integer]. Got: 44.4.
+
+class Api::V3::DecisionReview::HashPathValidator
```
pkarman: I like this! Nice delegation validation pattern.

Where is it used though? Maybe I missed seeing it.

app/services/api/v3/decision_review/contestable_issue_params.rb
```diff
@@ -0,0 +1,93 @@
...
+      decision_date: contestable_issue&.approx_decision_date,
+      benefit_type: @benefit_type,
+      ramp_claim_id: contestable_issue&.ramp_claim_id,
+      contested_decision_issue_id: contestable_issue&.decision_issue&.id
+    )
+  end
+
+  private
+
+  def lookup
```
pkarman: Class names should be nouns. Could we call this `ContestableIssueFinder` instead?

app/services/api/v3/decision_review/higher_level_review_intake_params.rb
```diff
@@ -0,0 +1,324 @@
...
+    attributes.dig("veteran", "fileNumberOrSsn").to_s.strip
+  end
+
+  def attributes
+    attributes? ? @params["data"]["attributes"] : {}
+  end
+
+  def attributes?
+    params? &&
+      @params["data"].respond_to?(:has_key?) &&
```
pkarman: `respond_to` is a code smell. Are you testing for duck-typing? Maybe just `try(:has_key)` would do it for you?

app/services/api/v3/decision_review/lookup_contestable_issue.rb
```diff
@@ -0,0 +1,76 @@
...
+  def matches_rating_decision_issue_id?(contestable_issue)
+    contestable_issue&.rating_decision_reference_id.to_s.strip == @rating_decision_issue_id
+  end
+
+  def contestable_issues_for_veteran_and_form_type
+    @contestable_issues_for_veteran_and_form_type ||= contestable_issue_generator.contestable_issues
+  end
+
+  def contestable_issue_generator
+    @contestable_issue_generator ||= ContestableIssueGenerator.new(
```
pkarman: typically we use `attr_reader` and define private methods instead of using instance vars directly.

app/services/api/v3/decision_review/lookup_contestable_issue.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
```
pkarman: I get what this class is doing. I would just name it differently.

app/services/api/v3/decision_review/contestable_issue_params.rb
```diff
@@ -0,0 +1,93 @@
...
+        veteran:,
+        receipt_date:,
+        benefit_type:,
+        params:
+      )
+    @decision_review_class = decision_review_class
+    @veteran = veteran
+    @receipt_date = receipt_date
+    @benefit_type = benefit_type
+    @ids = params.as_json["attributes"]
```
erluti: if you are trying to get away from the `ActionController::Parameters` object you could do something like this:
```
@id = params.permit('attributes')
              .to_hash.slice("ratingIssueId", "decisionIssueId", "ratingDecisionIssueId")
             .symbolize_keys
```

I think the use of `as_json` is a little misleading here since you aren't trying to use json.  `permit` will return a `ActionController::Parameters` object that you can call `to_hash` on because it will only contain the permitted values.

app/services/api/v3/decision_review/higher_level_review_intake_params.rb
```diff
@@ -0,0 +1,324 @@
...
+    validate
+  end
+
+  def intake_errors?
+    !@intake_errors.empty?
+  end
+
+  # params for IntakesController#review
+  def review_params
+    ActionController::Parameters.new(
```
erluti: Why is this a Parameters instead of a Hash?

app/services/api/v3/decision_review/lookup_contestable_issue.rb
```diff
@@ -0,0 +1,76 @@
...
+# answers whether or not the provided ids:
+#   rating_issue_id
+#   decision_issue_id
+#   rating_decision_issue_id
+#
+# reference a contestable issue (`valid?` method)
+# and returns that contestable issue (`contestable_issue` method)
+
+class Api::V3::DecisionReview::LookupContestableIssue
+  def initialize(opts)
```
erluti: Maybe we could formalize these to make it a little clearer.
perhaps like:
```
def initialize(decision_review_class:, veteran:, receipt_date:, benefit_type:, issues_by_id:)
```
The first four are "required' for setting up a ContestableIssueGenerator, and then the issues_by_id will be better described with a name for what it's doing.

app/services/api/v3/decision_review/lookup_contestable_issue.rb
```diff
@@ -0,0 +1,76 @@
...
+        :receipt_date,
+        :benefit_type
+      )
+    @rating_issue_id,
+      @decision_issue_id,
+      @rating_decision_issue_id = [
+        :ratingIssueId,
+        :decisionIssueId,
+        :ratingDecisionIssueId
+      ].map { |key| opts[key].to_s.strip }
```
erluti: why use `map` here instead of declaring them explicitly?

app/services/api/v3/decision_review/contestable_issue_params.rb
edmangimelli: i went from LookupContestableIssue -> ContestableIssueFinder -> LookupContestableIssue

👍

app/services/api/v3/decision_review/higher_level_review_intake_params.rb
edmangimelli: could do `dig` and catch the exception, alternatively. But, a missing object (or sub-object) seems like a common enough occurrence that it shouldn't throw exceptions

app/services/api/v3/decision_review/lookup_contestable_issue.rb
edmangimelli: even for this caching pattern?

app/services/api/v3/decision_review/contestable_issue_params.rb
edmangimelli: i like it +1

app/services/api/v3/decision_review/higher_level_review_intake_params.rb
edmangimelli: i'm using the exact same model methods that the Intakes Controller is using, therefore I want to pass the exact same type of thing (i.e. those methods expect parameters not hashes)

app/services/api/v3/decision_review/lookup_contestable_issue.rb
edmangimelli: i switched from that to this. let me revisit why i wanted to do this more dynamically

app/services/api/v3/decision_review/lookup_contestable_issue.rb
edmangimelli: brevity

app/services/api/v3/decision_review/hash_path_validator.rb
edmangimelli: https://github.com/department-of-veterans-affairs/caseflow/blob/c1dc993afeb32b10e8b44248d8c577f23ba77054/app/services/api/v3/decision_review/higher_level_review_intake_params.rb#L160

app/services/api/v3/decision_review/higher_level_review_intake_params.rb
erluti: I'm hoping to leave parameters out of the models and leave them in the controller.  I personally feel like they have some "surprise" features I don't like working around and would like to not have to deal with them in models and tests.

app/controllers/api/docs/v3/decision_reviews.yaml
```diff
@@ -681,73 +520,33 @@ paths:
...
-                    ]
-                  }
-                request_issue_must_have_at_least_one_ID_field:
-                  value: {
-                    "errors": [
-                      {
-                        "status": "422",
-                        "title": "A request issue must have at least one of the following: decisionIssueId, ratingIssueId, legacyAppealId, legacyAppealIssueId",
-                        "code": "request_issue_must_have_at_least_one_ID_field"
+                        "title": "Couldn't find ContestableIssue",
```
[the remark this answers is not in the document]
edmangimelli: Sure.  Whyn't?
¯\_(ツ)_/¯

edmangimelli:
https://github.com/department-of-veterans-affairs/caseflow/pull/12839#discussion_r362054306
@pkarman

>  I suspect that we have conveniently treated them like Hash when they are not, strictly.

You're correct. I'm being overly paranoid I think

app/services/api/v3/decision_review/contestable_issue_params.rb
```diff
@@ -0,0 +1,93 @@
...
+      benefit_type: @benefit_type,
+      ramp_claim_id: contestable_issue&.ramp_claim_id,
+      contested_decision_issue_id: contestable_issue&.decision_issue&.id
+    )
+  end
+
+  private
+
+  def lookup
+    @lookup ||= Api::V3::DecisionReview::ContestableIssueFinder.new(
```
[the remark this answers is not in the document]
edmangimelli: 👍

app/services/api/v3/decision_review/contestable_issue_finder.rb
```diff
@@ -0,0 +1,77 @@
...
+  def matches_rating_decision_issue_id?(contestable_issue)
+    contestable_issue&.rating_decision_reference_id.to_s.strip == @rating_decision_issue_id
+  end
+
+  def contestable_issues_for_veteran_and_form_type
+    @contestable_issues_for_veteran_and_form_type ||= contestable_issue_generator.contestable_issues
+  end
+
+  def contestable_issue_generator
+    @contestable_issue_generator ||= ContestableIssueGenerator.new(
```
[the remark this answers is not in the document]
edmangimelli: 👍

pkarman:
> the models really do expect parameters (the models use methods specific to parameters)

then we should definitely change that, but it's ok to do it in a separate PR.

PR #12842, opened 2019-12-03
Use SimpleDelegator for a delegation pattern

pkarman:
Small refactor to reduce code. The SimpleDelegator standard library allows for the delegation pattern in less code.

Also renames the test to match the tested class path.


pkarman:
@nanotone I would start with the docs: https://ruby-doc.org/stdlib-2.5.1/libdoc/delegate/rdoc/Delegator.html and https://ruby-doc.org/stdlib-2.5.1/libdoc/delegate/rdoc/SimpleDelegator.html

> SimpleDelegator, not being a module or concern, seems like it can't be used if you actually want to inherit from some other class for business logic, right?

Inheritance and delegation are 2 different OO patterns for extending the functionality of existing classes. So it's not correct IMO to talk about them as mutually exclusive for business logic. They just have pros/cons in different contexts and can be used together.

You can use delegation to extend a class and implement business logic, as the class does in this PR.

> It's also not possible to delegate methods to multiple destinations, or delegate only some method names?

If you wanted to delegate only some method names, then you would probably use the built-in Ruby `delegate` instead. Same pattern, different mechanism.

In this case, we were delegating so many methods on the RequestIssue object that it just made more sense to me to delegate all of them by default, since we weren't actively trying to *not* delegate some.

PR #12845, opened 2019-12-03
disable hearing dropdowns while async job is running

### LINKED ISSUE #12830: disable HearingTypeDropdown & DispositionDropdown until async job is completed
User should not change hearing type or cancel a virtual hearing while the background job is still running. The field `jobCompleted` in `VirtualHearingSerializer` indicates whether or not background job was completed.

[Screen Shot 2019-12-02 at 1.56.08 PM.png](https://images.zenhubusercontent.com/5d518fc3e04e3134e624281f/f4508456-fd53-4a3c-9787-3ed271a00ee2)
[Screen Shot 2019-12-02 at 1.53.44 PM.png](https://images.zenhubusercontent.com/5d518fc3e04e3134e624281f/d8e33352-0120-4653-8a12-51087a6b1af0)

rubaiyat22:
Resolves #12830

### Description
- disable HearingType dropdown on hearing details page until async job is completed
- disable Disposition dropdown on daily docket page until async job is completed
- write feature tests


client/app/hearings/components/dailyDocket/DailyDocketRow.jsx
```diff
@@ -188,15 +188,15 @@ class DailyDocketRow extends React.Component {
   }
 
   defaultRightInputs = () => {
-    const { hearing, regionalOffice } = this.props;
+    const { hearing, regionalOffice, readOnly } = this.props;
```
lomaxap: i would expect `readOnly` to disable all inputs for the form. why isn't `!jobCompleted` sufficient?

rubaiyat22: @lomaxap  The reason why i needed to default to `readOnly` is because of the different access users have based on their role. Just using !jobCompleted isn't sufficient when a user these roles `["RO ViewHearSched", "VSO"]` because it does not disable the dropdown.

PR #12847, opened 2019-12-03
Do not raise FilenumberDoesNotExist during document prefetch

pkarman:
Should silence some non-actionable Sentry alerts in the #appeals-reader channel.

### DIFF
--- a/app/workflows/fetch_documents_for_reader_job.rb
+++ b/app/workflows/fetch_documents_for_reader_job.rb
@@ -10,6 +10,11 @@ class FetchDocumentsForReaderJob
   def process
     setup_debug_context
     appeals.each { |appeal| fetch_for_appeal(appeal) }
+  rescue VBMS::FilenumberDoesNotExist => error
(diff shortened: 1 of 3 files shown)


app/workflows/fetch_documents_for_reader_job.rb
```diff
@@ -10,6 +10,11 @@ def initialize(user:, appeals:)
   def process
     setup_debug_context
     appeals.each { |appeal| fetch_for_appeal(appeal) }
+  rescue VBMS::FilenumberDoesNotExist => error
+    # there is nothing actionable here, since it reflects data changes on the VBMS side.
+    # we do not want to retry since it will never work.
+    log_error
```
jcq: This will simply log "ERROR" without the context of what kind of error it is, right? Is that OK to lose?

pkarman: good catch. added explicit error logging.

PR #12850, opened 2019-12-03
Create function to handle updating task instructions

hschallhorn:
### Description
Removed redundant code where we append new instructions when updating a task. (spun out of #12837)


spec/models/tasks/colocated_task_spec.rb
```diff
@@ -511,7 +511,7 @@
         assigned_by: initial_assigner
       )
     end
-    let(:colocated_task) { org_task.children.first }
+    let!(:colocated_task) { org_task.children.first }
```
hschallhorn: `Colocated.singleton.next_assignee.id` on line 519 was returning the same user `colocated_task` was assigned to as this task had not been created yet, triggering a duplicate task error. Because we force this task to be created before setting the reassign params, `next_assignee` iterates to the next colocated user.

spec/feature/queue/attorney_queue_spec.rb
```diff
@@ -140,12 +140,13 @@
           assigned_by: attorney
         )
       end
-      let(:colocated_person_task) { colocated_org_task.children.first }
+      let!(:colocated_person_task) { colocated_org_task.children.first }
 
       before do
+        Colocated.singleton.add_user(create(:user))
```
hschallhorn: There needs to be another person in the colocated org for `Colocated.singleton.next_assignee` to return someone other than the user `colocated_person_task` is assigned to. Throws a duplicate task error if not.

spec/feature/queue/colocated_task_queue_spec.rb
```diff
@@ -67,7 +67,7 @@
       # verify that the instructions from the VLJ appear on the case timeline
       scroll_to("#case_timeline-section")
       view_text = COPY::TASK_SNAPSHOT_VIEW_TASK_INSTRUCTIONS_LABEL
-      page.find_all("table#case-timeline-table button", text: view_text).each(&:click)
+      page.find("table#case-timeline-table button", text: view_text).click
```
hschallhorn: `find_all` will not fail if it finds no matching element. If the page does not load right away, we won't actually open the task instructions, which I believe was causing the flake on the line below. By changing this to `find`, we wait until the "View instructions" element is loaded, then click it.

spec/models/tasks/colocated_task_spec.rb
yoomlam: Why is `next_assignee` reliant on the existence of any task at all? My intuition for `next_assignee` is that is returns the next available user. Since no options were provided to `next_assignee` it shouldn't be dependent on any task information.

spec/feature/queue/colocated_task_queue_spec.rb
yoomlam: Nice! I created a wiki page https://github.com/department-of-veterans-affairs/caseflow/wiki/Flakey-Test-Remedies and added your tip. Feel free to modify it.

app/models/task.rb
```diff
@@ -377,7 +377,11 @@ def update_from_params(params, current_user)
...
 
-    params["instructions"] = flattened_instructions(params)
+    update_with_instructions(params)
+  end
+
+  def update_with_instructions(params)
+    params[:instructions] = flattened_instructions(params)
     update!(params)
 
     [self]
```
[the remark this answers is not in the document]
hschallhorn: Fantastic point! will update.

spec/models/tasks/colocated_task_spec.rb
hschallhorn: The issue here is that when the task is created, `next_assignee` returns the same person as has been set as the assignee in the `reassign_params` (resulting in the duplicate task as they're assigned to the same person). If we create the task before we set the `reassign_params`, `next_assignee` returns a different person. `next_assignee` is reliant on what tasks have already been assigned to users. It finds the most recently assigned task, checks the assignee, and returns the next user in the organization.

spec/feature/queue/attorney_queue_spec.rb
hschallhorn: Previously, the reassign function only flattened the array of instructions, but did not get rid of `nil`s. This test was previously passing because the instructions on the old task were `["instructions"]` and the instructions on the new task were `["instructions", nil]`. `flatten_instructions` flattens and compacts the old and new instructions, which tosses out `nil`s. As the assignee was the same, and now the instructions are the same, a duplicate task error is thrown

spec/feature/queue/colocated_task_queue_spec.rb
hschallhorn: Oh awesome! Thanks!

hschallhorn: TIL about diff syntax highlighting in markdown!

spec/feature/queue/attorney_queue_spec.rb
yoomlam: Duplicate task error is checked at https://github.com/department-of-veterans-affairs/caseflow/blob/30add3bab5652ffc14ecafdb2a94a1892572daef/app/models/tasks/colocated_task.rb#L160.

We only check for duplicates for ColocatedTasks and its descendants. We also check for duplicate org tasks: https://github.com/department-of-veterans-affairs/caseflow/blob/30add3bab5652ffc14ecafdb2a94a1892572daef/app/models/task.rb#L218

Issue #12672, opened 2019-11-11
Message user when their asyncable job has been canceled


leikkisa:
## User or job story
As a user, I'd like to know when my asyncable job has been canceled.

## Acceptance criteria
- [ ] Add a method such as "cancel_job_with_note" that cancels an asyncable job and sends a job note to the users inbox with a message, indicating whether further action is necessary.
- [ ] The message should automatically include a link to the job detail page where the user can see job details and notes
- [ ] The message should be from the current user (the user canceling the job)
- [ ] Please add a comment above the method instructing the user to add a message explaining why the job was canceled, and describing any additional action necessary
- [ ] Please add a process note to this ticket, with instructions on using the new method when canceling an asyncable job: #12276
- [ ] Include a screenshot in the Github issue of the message about cancellation

## Release notes

When an asyncable job has been canceled, the user will automatically receive a message indicating that it has been canceled.

### Designs
As per PR #12854, here are sample screenshots of job cancellation in action.

Inbox message notifying the user responsible for the job:
<img width="832" alt="Screen Shot 2019-12-03 at 17 49 48" src="https://user-images.githubusercontent.com/282869/70098381-c7b83680-15f9-11ea-8b1c-1c837f8bdc55.png">

A new job note appended to the job:
<img width="409" alt="Screen Shot 2019-12-03 at 17 49 02" src="https://user-images.githubusercontent.com/282869/70098321-a1929680-15f9-11ea-9394-b0f333f76c74.png">

### Background/context
Currently, job cancellations should only happen manually by an engineer.

This ticket has been broken out from: https://github.com/department-of-veterans-affairs/caseflow/issues/11907 which includes implementation of the inbox feature and further details and designs.

See #12214 for details on job notes.

### Technical notes
The asyncable concern has a method called canceled! that cancels asyncable jobs.  The new method will cancel the job and add an Job note. #12276 has instructions for adding a job note.

Issue #12276, opened 2019-10-02
Asynchronous Claims | Background Jobs


leikkisa:
### Description
This epic is here to connect all of the individual upstream error issues and describe the overall process for handling them.

This process is new and evolving, so please feel free to make improvements.

### Goals
1. Identify errors and ensure that they all have Incident reports, GitHub issues and job notes. We will use these as a form of communication to clients and users.
1. Ultimately, solve all of the issues!

### Process
1. Look for asynchronous jobs that haven't been accounted for yet (see code below)
1. Look for the related GitHub issue, which should be tagged with "upstream-error" and part of this epic.
    1. If there is no GitHub issue yet, please create one, connect it to this epic and tag it with "upstream-error" and post it in the #foxtrot channel in Slack
1. Add a note to the asynchronous job notes, referencing the GitHub issue.
1. Add the impacted job to the ticket in a markdown table, with the INC number and the YourIT link
1. If the asynchronous job needs to be cancelled, it can be done so in the Rails console, e.g.:
```
pry> job = HigherLevelReview.find(SOME_ID)
pry> user = User.find(YOUR_USERID)
pry> job.cancel_with_note!(current_user: user, note: "the reason for cancellation")
```

### Finding new errors
```
engineer_user_ids = [11110, 2197, 7850, 1376] # Sally, Peter, Tomas, Sandra
scs=SupplementalClaim.expired_without_processing.where(establishment_canceled_at: nil)
hlrs=HigherLevelReview.expired_without_processing.where(establishment_canceled_at: nil)
reviews = scs + hlrs

# Reviews without a note from one of us
todo = reviews.select{ |r| r.job_notes.select{ |jn| engineer_user_ids.include?(jn.user_id) }.empty? }
```

### Adding notes to errors once you have completed the process for them:
```
# method
def add_job_note(user, job, github_issue_number=nil, user_note=nil)
  github_issue_note = github_issue_number ? "To track progress on this issue, please see: https://github.com/department-of-veterans-affairs/caseflow/issues/#{github_issue_number}" : nil
  note = [user_note, github_issue_note].compact.join("\n")

  fail "Github issue number or user note is required" unless note

  JobNote.create!(
    job: job,
    user: user,
    note: note,
    send_to_intake_user: false
  )
end

# Example using it for a single job
my_user_id=1234
my_user=User.find(my_user_id)
job=todo.first # This is for whatever job you're looking at, such as a specific SC or HLR
github_issue_number=1234
[...]

PR #12854, opened 2019-12-03
Add inbox messaging for when a job is manually cancelled

nanotone:
Connects #12672 

### Description
Added a method called "cancel_job_with_note!" that cancels an asyncable job and sends a job note to the users inbox with a message, indicating whether further action is necessary.

### Acceptance Criteria
- [x] The message should automatically include a link to the job detail page where the user can see job details and notes
- [x] The message should be from the current user (the user canceling the job)
- [x] Please add a comment above the method instructing the user to add a message explaining why the job was canceled, and describing any additional action necessary
- [x] Please add a process note to this ticket, with instructions on using the new method when canceling an asyncable job: #12276
- [x] Include a screenshot in the Github issue of the message about cancellation

### Testing Plan
1. The cancellation itself must be done in Rails console:
```
pry> job = HigherLevelReview.find(SOME_ID)
pry> user = User.find(SOME_ADMIN_USERID)
pry> job.cancel_with_note!(current_user: user, note: "some reason")
```
2. As the job's asyncable user, visit `/inbox` and verify that a message was sent.
3. Click on the link in the message to visit the job, and verify that a job note was added.

### User Facing Changes

Sample inbox message notifying the asyncable user of the cancellation:
Screen Shot 2019-12-03 at 17 49 48

Sample job note appended to the job:
Screen Shot 2019-12-03 at 17 49 02
 
### Documentation Updates
- [ ] Comment added to the new method, describing what circumstances a job may be cancelled (manually by engineer) and what's required (the canceller's user object, and sufficient description of the reason).

### DIFF
--- a/app/models/claim_review.rb
+++ b/app/models/claim_review.rb
@@ -121,6 +121,17 @@ class ClaimReview < DecisionReview
     AsyncableJobMessaging.new(job: self).handle_job_failure
   end
 
+  # Cancel an unprocessed job, add a job note, and send a message to the job user's inbox.
+  # Currently this only happens manually by an engineer, and requires a message explaining
+  # why the job was cancelled and describing any additional action necessary.
+  def cancel_with_note!(current_user:, note:)
+    fail Caseflow::Error::ActionForbiddenError, message: "Acting user must be specified" unless current_user
+    fail Caseflow::Error::ActionForbiddenError, message: "Processed job cannot be cancelled" if processed?
+
+    canceled!
+    AsyncableJobMessaging.new(job: self, current_user: current_user).add_job_cancellation_note(text: note)
+  end
+
   def invalid_modifiers
     end_product_establishments.map(&:modifier).reject(&:nil?)
   end
--- a/app/models/job_note.rb
+++ b/app/models/job_note.rb
@@ -15,4 +15,8 @@ class JobNote < ApplicationRecord
       sent_to_intake_user: send_to_intake_user
     }
   end
+
+  def path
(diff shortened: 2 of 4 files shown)


app/models/claim_review.rb
```diff
@@ -121,6 +121,17 @@ def update_error!(err)
     AsyncableJobMessaging.new(job: self).handle_job_failure
   end
 
+  # Cancel an unprocessed job, add a job note, and send a message to the job user's inbox.
+  # Currently this only happens manually by an engineer, and requires a message explaining
+  # why the job was cancelled and describing any additional action necessary.
+  def cancel_with_note!(current_user:, note:)
```
leikkisa: Nice error handling :-)

nanotone:
@leikkisa I agree that we should revisit whether async messaging is available to jobs other than ClaimReview subclasses. I think a big question here is end-user workflows, e.g. whether the `asyncable_user` of an Appeal or VirtualHearingEstablishment would have any use for this.

app/models/claim_review.rb
```diff
@@ -121,6 +121,17 @@ def update_error!(err)
...
+    fail Caseflow::Error::ActionForbiddenError, message: "Acting user must be specified" unless current_user
+    fail Caseflow::Error::ActionForbiddenError, message: "Processed job cannot be cancelled" if processed?
+
+    canceled!
```
pkarman: I think we want to call `cancel_establishment!` here instead, since that also cancels things besides the job.

Issue #12410, opened 2019-10-16
Convert ui_hash to serializers (issues)


leikkisa:
### Description
As a tech improvement ticket, in order to improve consistency and code quality, and reduce the size of our models, we can convert ui_hash method to use serializers instead.

### Acceptance criteria

- "Issue" files use serializers instead of ui_hash:
    - ramp_issue
    - rating_issue
    - request_issue
    - rating
- Any new files include top line comments for documentation

Please note that decision_issue is in another ticket

related:
https://github.com/department-of-veterans-affairs/caseflow/issues/12408
https://github.com/department-of-veterans-affairs/caseflow/issues/12409
https://github.com/department-of-veterans-affairs/caseflow/issues/12355

PR #12861, opened 2019-12-04
Convert ui_hash to serializers (issues)

Sjones352:
Connects #12410


app/serializers/rating_serializer.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class RatingSerializer
+  include FastJsonapi::ObjectSerializer
+  set_id(&:participant_id)
```
leikkisa: I would set this to the profile_date, because that will be unique per rating per veteran

Sjones352:
@nanotone that is a method that would is called in so many places and would require tracing where we call `ui_hash` and change it to serialize

app/serializers/request_issue_serializer.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class RequestIssueSerializer
+  include FastJsonapi::ObjectSerializer
+
+  # converts ui_hash into a serializer
```
leikkisa: I don't think this comment is needed

app/models/rating.rb
```diff
@@ -30,22 +30,12 @@ def ignorable?
...
   TWO_LIFETIMES_DAYS = 250.years
 
-  def ui_hash
-    serialize
-  end
-
   # If you change this method, you will need
   # to clear cache in prod for your changes to
   # take effect immediately.
   # See DecisionReview#cached_serialized_timely_ratings.
```
leikkisa: I think this comment belongs in the new serializer file, or it could be removed since isn't likely to change and the cache is temporary.

app/models/ramp_issue.rb
```diff
@@ -18,7 +18,7 @@ def contention_text
     Contention.new(description).text
   end
 
-  def ui_hash
-    { id: id, description: description }
+  def serialize
```
leikkisa: You'll need to find where ui_hash was being called on Ramp Issues and update this there too. Check out `ramp_refiling_intake.rb`

app/models/request_issue.rb
```diff
@@ -313,36 +313,9 @@ def withdrawal_date
...
-      contested_decision_issue_id: contested_decision_issue_id,
-      withdrawal_date: withdrawal_date,
-      contested_issue_description: contested_issue_description,
-      end_product_cleared: end_product_establishment&.status_cleared?,
-      end_product_code: end_product_code,
-      editable: editable?
-    }
-  end
-  # rubocop:enable Metrics/MethodLength
+  def serialize
```
leikkisa: The `ui_hash` method for request issues was actually called in a handful of places, so you'll need to find those and update them.

Generally, I would do a search in the codebase for `ui_hash` and see if it seems to be for an "issue" that you are handling here - a rating issue, request issue, etc, and you can update those to serialize.

PR #12868, opened 2019-12-04
remove task's action field references

### LINKED ISSUE #12279: Remove references to tasks classed by action
As of #12272, we no longer store the action field of a task as it is no longer used. However, we still have [many](/department-of-veterans-affairs/caseflow/blob/master/app/controllers/tasks_controller.rb#L211) leftover [references](https://github.com/department-of-veterans-affairs/caseflow/blob/b5a9df63f9403f0e7d1081a6b30bef1c853fb46e/app/controllers/tasks_controller.rb#L221) to [the](https://github.com/department-of-veterans-affairs/caseflow/blob/b5a9df63f9403f0e7d1081a6b30bef1c853fb46e/app/models/task.rb#L111) action [field](https://github.com/department-of-veterans-affairs/caseflow/blob/b5a9df63f9403f0e7d1081a6b30bef1c853fb46e/app/models/tasks/colocated_task.rb#L34) (This is just a sample, there are a few more that need tracking down).

## Acceptance criteria
- [ ] Make sure there are no references to the now-defunct action field [more of a satisfaction criteria than action item]

yoomlam:
Resolves #12279

### Description
Replaced uses of `:action` with `:type` where appropriate; otherwise removed its use.

### Acceptance Criteria
- Make sure there are no references to the now-defunct action field [more of a satisfaction criteria than action item]


yoomlam:
My remaining concern is if I caught all occurrences of the `action` field for tasks. See the original ticket where I show my work to assess if I may have missed something.

app/models/task.rb
```diff
@@ -101,11 +101,10 @@ def create_many_from_params(params_array, current_user)
       end
     end
 
-    def modify_params(params)
+    def modify_params_for_create(params)
```
hschallhorn: Dig this rename 👍

app/models/tasks/colocated_task.rb
```diff
@@ -114,7 +119,7 @@ def actions_available?(_user)
   end
 
   def create_twin_of_type(params)
-    task_type = ColocatedTask.find_subclass_by_action(params[:action])
+    task_type = ColocatedTask.find_subclass_by_type(params[:type])
```
hschallhorn: `params[:type]` should already be the `task_type` we pass to  `create!`. W should not need to do a lookup here. We can pass `params[:type]` directly to `type:`

app/models/tasks/colocated_task.rb
```diff
@@ -31,7 +31,7 @@ def create_many_from_params(params_array, user)
         params_array = params_array.map do |params|
           # Find the task type for a given action.
           create_params = params.clone
-          new_task_type = find_subclass_by_action(create_params.delete(:action).to_s)
+          new_task_type = find_subclass_by_type(create_params[:type])
```
hschallhorn: This could be replaced by `Object.const_get(params[:type])` (the reverse method of `<class>.name`)

app/controllers/tasks_controller.rb
```diff
@@ -209,18 +210,15 @@ def task
 
   def create_params
     @create_params ||= [params.require("tasks")].flatten.map do |task|
-      task = task.permit(:type, :instructions, :action, :label, :assigned_to_id,
+      task = task.permit(:type, :instructions, :label, :assigned_to_id,
```
hschallhorn: We could remove `:label` from the permitted params here if we don't pass it to the [back end here](https://github.com/department-of-veterans-affairs/caseflow/blob/6e5725d9e5bcf4e6b51060592d69d2a2288cdac3/client/app/queue/AddColocatedTaskView.jsx#L82). This would also allow us to remove line 221

app/models/tasks/colocated_task.rb
```diff
@@ -69,6 +69,11 @@ def find_subclass_by_action(action)
       subclasses.find { |task_class| task_class.label == Constants::CO_LOCATED_ADMIN_ACTIONS[action] }
     end
 
+    def find_subclass_by_type(type)
+      subclasses.find { |task_class| task_class.name == type }
+    end
```
hschallhorn: I think this method could be removed if we use `const_get` above and the redundant type lookup below

app/repositories/task_action_repository.rb
```diff
@@ -203,13 +203,12 @@ def legacy_and_colocated_task_add_admin_action_data(_task, _user)
...
-        options: Constants::CO_LOCATED_ADMIN_ACTIONS.map do |key, value|
+        options: ColocatedTask.subclasses.map do |subclass|
           {
-            label: value,
-            value: key
+            label: subclass.label,
+            value: subclass.name
           }
-        end,
-        type: ColocatedTask.name
```
hschallhorn: Because we remove the task type from the task action repository data here, we can remove [`taskActionData(this.props).type || `](https://github.com/department-of-veterans-affairs/caseflow/blob/6e5725d9e5bcf4e6b51060592d69d2a2288cdac3/client/app/queue/AddColocatedTaskView.jsx#L84).

spec/controllers/tasks/change_type_controller_spec.rb
```diff
@@ -67,7 +67,7 @@
         end
 
         context "that needs reassigning" do
-          let(:new_task_type) { :foia }
+          let(:new_task_type) { FoiaColocatedTask }
           let(:new_task_class) { FoiaColocatedTask }
```
hschallhorn: Can we remove one of these variables and reuse the other?

spec/controllers/tasks_controller_spec.rb
```diff
@@ -425,7 +423,7 @@
           end
 
           it "should be successful" do
-            expect(AppealRepository).to receive(:update_location!).exactly(1).times
+            expect(AppealRepository).to receive(:update_location!).exactly(2).times
```
hschallhorn: Do we know why this increased?

yoomlam:
From Hunter: 
> Re: why you need ColocatedTask to call subclasses: It could be trying to call `subclasses` on a subclass of ColocatedTask, which would return nothing.

Obviously.  ;)

app/controllers/tasks_controller.rb
yoomlam: I also refactored to get rid of `:external_id` and line 218.

hschallhorn: get 👏 those 👏 line 👏 deletion 👏 numbers 👏 up 👏

app/models/tasks/colocated_task.rb
yoomlam: I thought about that but I also thought it may be safer to restrict the possible classes to the subclasses of `ColocatedTask`. What do you think? Is that being too paranoid?

app/models/tasks/colocated_task.rb
yoomlam: What about `assigned_to: task_type&.default_assignee` on line 128? We'll still need to look it up to get the `default_assignee`, right?

app/models/tasks/colocated_task.rb
hschallhorn: The only way we could get to this point in `ColocatedTask` is if the task type is a [colocated subclass](https://github.com/department-of-veterans-affairs/caseflow/blob/7b5251aaf8200942262a802744eb8a9ff2d0f571/app/controllers/tasks_controller.rb#L74). So maybe a smidge too paranoid 😄

app/models/tasks/colocated_task.rb
hschallhorn: You're completely right. Missed that we use the default assignee here as well

spec/controllers/tasks/change_type_controller_spec.rb
yoomlam: LOL. `:new_task_class` wasn't even used, so yea.

spec/controllers/tasks_controller_spec.rb
yoomlam: Sort of (I'm not confident in my knowledge of the underlying calls). Originally, the `external_id` and `type` were the same, and now the task `type` is different. So the count seems to reflect the number of unique types:

- If I add another task with the same `type`, the count remains 2.
- If I add another task with a different `type`, the count becomes 3.

This doesn't completely make sense to me but I don't have a good understanding of this.

app/repositories/task_action_repository.rb
yoomlam: I made the edit, but after a bit of digging, I still don't understand why. I'm gonna ask you to walk me through this reasoning via a Slack Chat to save me some time. I feel like I've spent twice the amount of time than was estimated for the ticket.

yoomlam:
Woohoo! All tests passed ... this time.

app/models/tasks/colocated_task.rb
```diff
@@ -69,6 +69,7 @@ def find_subclass_by_action(action)
       subclasses.find { |task_class| task_class.label == Constants::CO_LOCATED_ADMIN_ACTIONS[action] }
     end
 
+    # Is this method still relevant given ticket #12279 and related tickets?
```
hschallhorn: This is only used in seeds and testing. We could spin out a ticket to move it to someplace more appropriate, perhaps task.rb in factories. But definitely sill relevant as sometimes we want to get a random task we know will be assigned to the colocated team (FoiaColocatedTasks, MissingHearingTranscriptsColocatedTasks, and TranslationColocatedTasks get assigned to other teams and behave differently than all other ColocatedTasks)

app/models/tasks/colocated_task.rb
```diff
@@ -114,14 +115,14 @@ def actions_available?(_user)
...
-    task_type = ColocatedTask.find_subclass_by_action(params[:action])
+    task_type = Object.const_get(params[:type])
     ColocatedTask.create!(
       appeal: appeal,
       parent: parent,
       assigned_by: assigned_by,
       instructions: params[:instructions],
       assigned_to: task_type&.default_assignee,
-      type: task_type
+      type: params[:type]
```
hschallhorn: Wondering if we want to leave this as `task_type`, `task_type&.name`, or `params[:type]`. I don't know if I have a preference, but I do think what we use in [`create_params`](https://github.com/department-of-veterans-affairs/caseflow/pull/12868/files#diff-8b82658bb308177c50466acc7dd20d8dR35) should match

spec/controllers/tasks_controller_spec.rb
hschallhorn: yep, you're 100% correct.

app/models/tasks/colocated_task.rb
yoomlam: I'll let you write up that ticket since you have more context.

hschallhorn:
I would also remove [this line](https://github.com/department-of-veterans-affairs/caseflow/blob/e83a655282c92471688da5c48327057fd3ed1ea9/client/app/queue/AddColocatedTaskView.jsx#L82) as we are no longer using label on the back end

app/models/tasks/colocated_task.rb
yoomlam: Done.

app/repositories/task_action_repository.rb
```diff
@@ -203,13 +203,12 @@ def legacy_and_colocated_task_add_admin_action_data(_task, _user)
       {
         redirect_after: "/queue",
         selected: nil,
-        options: Constants::CO_LOCATED_ADMIN_ACTIONS.map do |key, value|
+        options: ColocatedTask.subclasses.map do |subclass|
```
yoomlam: `ColocatedTask.subclasses.count` returns 19.
`Constants::CO_LOCATED_ADMIN_ACTIONS.count` returns 18.
Causing failure: https://circleci.com/gh/department-of-veterans-affairs/caseflow/96078

yoomlam: Still don't know why this fails inconsistently!

Issue #12691, opened 2019-11-12
Fix action permissions for MTV tasks


leikkisa:
## User or job story
As a user processing an MTV task, I should not be able to take action on another user's task, or a task that I don't have permissions to act on.

Screenshot of a user being able to act on both a judge task and an attorney task
![Screen Shot 2019-11-08 at 4.19.24 PM.png](https://images.zenhubusercontent.com/58f6b6eb51aa2c30577bb60e/fe66dc32-b27e-4d3c-87c5-ab487e21c5ce)

## Acceptance criteria
- [ ] Only the assigned judge should be able to take action on judge tasks (check if team admin can also act on these)
- [ ] Only the assigned attorney should be able to take action on attorney tasks
- [ ] For organization tasks, only an organization admin should be able to take action on those tasks
- [ ] Include a screenshot in the Github issue of an MTV task tree with both a Judge and Attorney task, signed in as one type of user.  The user should only be able to take action on their task.

## Open questions
- Clarify if for organization tasks, should the only available action be the "assign task"?
  - _QR team did not want this; often the admin will act on the organization task_
- Should judge's be able to modify tasks for attorney's on their teams?
  - _for decision drafting, they want to be able to_
- Should org admins be able to act on their team's tasks?
  - _by default anyone on an org can act on an org task; is the above referring to tasks assigned to the org, or assigned to individuals in the org?_

## Release notes
N/A this is a bug fix for a feature that is not yet released

### Designs
N/A

### Technical notes

If someone wants to create a Pulac-cerullo task, they should do so off of a task that is assigned to them.

I believe there is a tech improvement ticket out there for improving available_actions, actions_available, and additional_available_actions, but those are all good methods to take a look at in task world.

leikkisa:
Closing this ticket, we think it's not necessary.

PR #12869, opened 2019-12-05
Fix action permissions for mtv tasks

tomas-nava:
Connects #12691

### Description
Sets strict permissions on who can see actions on MTV tasks:
- if the task is assigned to a user, only that user may see actions on the task
- if the task is assigned to an organization, only an admin in that organization may see actions on the task

I addressed the issue this way because none of the requested changes were universal; for example:
- For decision drafting, judges want to be able to act upon tasks assigned to attorneys on their teams
- Historically, anyone belonging to an organization can act on that organization's tasks

### Questions this PR does not address
An open question on #12691 asks whether "Assign task" should be the only available action on organization tasks; this PR does not address that question.

### Screenshots
_as judge JUDGE_1_
as-judge

_as attorney CSS_ID1_
as-attorney


tomas-nava:
> I think these are working right:
> An attorney can't act on their judge's task

Currently with this PR, yes, this will be the case (if it wasn't already).

> An org user can act on its org's tasks, whether or not they're an admin
> An org admin can act on its org's user's task

Currently with this PR, neither of these will be true.

> The second open question makes it sound like a judge should be able to act on their attorney's tasks, but right now that's not working.

I mentioned in #12691 that, for decision drafting, judges do want to be able to modify their attorney's tasks. My understanding is that for MTV tasks specifically, we don't want that to be the case (see the 2nd acceptance criterion: _Only the assigned attorney should be able to take action on attorney tasks_).

tomas-nava:
closing because we realized we don't need this functionality after all

Issue #12765, opened 2019-11-20
Don't Allow User to Send Empty Form Data


lomaxap:
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/7446/

suspected scenario: user changes something then changes it back to original value. the “save” button is still there, but no new data will be sent to backend when user clicks save.

we need a better [way of determining if the save button should be shown](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/hearings/components/dailyDocket/DailyDocketRow.jsx#L75) now that we're only sending saved data

tomas-nava:
@sahalliburton is this issue resolved by #12870?

PR #12870, opened 2019-12-05
Save Disposition better

### LINKED ISSUE #12797: Handle Disposition Updating and Saving better
### Background

We made the decision to only send changed fields in update requests so we could better implement virtual hearings logic that only happens if a field was changed. In order to do this we could have either:

1. stored updated values in a separate state from initial states on update, something like
```
<input value={hearing.updated.field || hearing.field} update={(val) => update({ field: val }) />
```
2. Store the entire initial state of the hearing and compare the initial value and changed value at the time of the request

We went with option 2 because it was more easily integrated with the various ways we've implemented forms.

### Issue

On the daily docket, the initial state for the hearing is stored on the `DailyDocketRow`
https://github.com/department-of-veterans-affairs/caseflow/blob/1b6879a99a42922cea5940e089e2f5ea510e0576/client/app/hearings/components/dailyDocket/DailyDocketRow.jsx#L62-L64

and then the comparison is done when the hearing is saved
https://github.com/department-of-veterans-affairs/caseflow/blob/1b6879a99a42922cea5940e089e2f5ea510e0576/client/app/hearings/components/dailyDocket/DailyDocketRow.jsx#L156-L174

This works fine except in the instance when the hearing is `cancelled` or `postponed`. "Currently Scheduled" and "Previously Scheduled" split hearings where the dispositions is null, `held` or `no_show` from those that are `cancelled` or `postponed` into separate `DailyDocketRows`:
https://github.com/department-of-veterans-affairs/caseflow/blob/44405358f7fe40e8d8175ad6e655ed21679690ca/client/app/hearings/components/dailyDocket/DailyDocket.jsx#L205-L212
https://github.com/department-of-veterans-affairs/caseflow/blob/44405358f7fe40e8d8175ad6e655ed21679690ca/client/app/hearings/components/dailyDocket/DailyDocket.jsx#L222-L230

This would be fine except the dropdown that controls disposition opens a modal before save:
https://github.com/department-of-veterans-affairs/caseflow/blob/44405358f7fe40e8d8175ad6e655ed21679690ca/client/app/hearings/components/dailyDocket/DailyDocketRowInputs.jsx#L29-L46

This means that when the currently scheduled hearings are changed to postponed, they are unmounted / removed from the first set of `DailyDocketRows` and mounted on the second, the modal opens, and when the user goes to `saveHearing`, the row no longer exists.

### Solutions
Since this issue only occurs when we're changing the disposition dropdown, we shouldn't complicate the entire save logic for one instance. Let's just change the disposition dropdown to handle this scenario:

```javascript
    onChange={(option) => {
      if (!option) {
        return;
      }

      const fromDisposition = hearing.disposition;
      const toDisposition = option.value;

      openDispositionModal({
        hearing,
        fromDisposition,
        toDisposition,
        onConfirm: saveThenUpdateDisposition
    }}
```
where `saveThenUpdateDisposition` is a slight variation on `saveHearing`, something like
```javascript
// class method on DailyDocketRow.jsx
saveThenUpdateDisposition = (toDisposition) => {
  const hearingWithDisp = { 
    ...this.props.hearing,
    disposition: toDisposition
  };

  const hearing = deepDiff(initialState, hearingWithDisp);

  this.props.saveHearing(hearing.externalId, hearing).then(() => {
    // update redux
    update(hearingWithDisp);
    // only update initialState if hearing isn't postponed or cancelled
    if (['postponed', 'cancelled'].indexOf(toDisp) === -1) {
     this.setState({ initialState: hearingWithDisp });
    }
  });
}
```

aplybeah:
Resolves #12797 

Changes the flow when hearing dispositions get updated on the `Daily Docket` page.
> Before:
* When users clicked `Confirm` on the modal, the disposition would save and the user wouldn't have the chance to make other edits.

> After:
* Disposition will only save once the `Save` button is clicked.

> Nit: 
*  I was unable to find where the `You have successully updated...` alert is triggered. I feel that there are too many alerts now, and that can lead to confusion. (ongoing)


ferristseng:
This introduces a regression, but I'm not sure why. It looks like the payload that gets sent to the server correctly, but the UI doesn't get updated. In fact, it looks like it gets reverted? It's hard to describe, so I captured it in a GIF:

out

ferristseng:
@tomas-nava Tagged you as a reviewer.

I think this is supposed to fix an issue where when you update the disposition from the dropdown, it updates the hearing disposition in the state, and pushes the daily docket row into the bottom section before you hit save.

sahalliburton:
Also addresses showing save button only when data changed: [#12765 ](https://github.com/department-of-veterans-affairs/caseflow/issues/12765)

client/app/hearings/components/dailyDocket/DailyDocketModals.jsx
```diff
@@ -89,67 +97,79 @@ AodModal.propTypes = {
...
+      <div className="cf-modal-scroll">
+        <Modal
+          closeHandler={onCancel}
+          confirmButton={this.confirmButton()}
+          cancelButton={this.cancelButton()}
+          title={this.COPY[hearingType].title}
+        >
+          <div>
+            <p>
+              Previous Disposition:{' '}
```
ferristseng, line 163: This syntax reads really strangely to me -- Is this something the linter put in?

client/app/hearings/utils.js
```diff
@@ -34,60 +35,110 @@ export const getWorksheetAppealsAndIssues = (worksheet) => {
...
+  // Determine whether the initial value is falsy
+  const falsy = init === null ? false : init;
+
+  // Handle the value comparison
+  switch (current) {
+  // Empty strings should be treated the same as false and null
+  case '':
+  case false:
+    return current != falsy;
+    // Default to compare the initial with the current value
```
ferristseng, line 108: Should this comment have the same indent as the `default`

client/app/hearings/utils.js
```diff
@@ -34,60 +35,110 @@ export const getWorksheetAppealsAndIssues = (worksheet) => {
...
-export const virtualHearingRoleForUser = (user, hearing) => (
-  user.userCanAssignHearingSchedule || user.userId === hearing.judgeId ?
-    VIRTUAL_HEARING_HOST :
-    VIRTUAL_HEARING_GUEST
-)
+/**
+ * Method to override falsy values for comparison
+ * @param {*} init -- Initial value to compare against
+ * @param {*} current -- Current Value to compare
+ * @returns {boolean} -- Whether the object has been changed
```
ferristseng, line 96: Nice! Great documentation!

client/app/hearings/components/dailyDocket/DailyDocketRow.jsx
```diff
@@ -70,23 +76,30 @@ class DailyDocketRow extends React.Component {
...
       edited: false,
+      editedFields: [],
       virtualHearingModalActive: false,
       startPolling: null
     };
   }
 
   update = (values) => {
+    // Determine whether the form has been edited and which fields
+    const { edited, editedFields } = handleEdit(this.state.initialState, values, this.state.editedFields);
```
ferristseng, line 87: Cool solution to determining if fields were edited!

client/app/hearings/components/dailyDocket/DailyDocketModals.jsx
sahalliburton: @ferristseng the syntax is a bit strange, it took me to get used to it at first, however this was applied by the linter based on the eslint config and it is React best practice for adding spaces after text in JSX: https://stackoverflow.com/questions/40264084/best-practice-when-adding-whitespace-in-jsx#answer-41142436

client/app/hearings/utils.js
sahalliburton: thanks for catching this! not sure why it was different

client/app/hearings/components/dailyDocket/DailyDocketModals.jsx
ferristseng: Cool, thanks for the link!

PR #12893, opened 2019-12-06
Fix intake dev data

leikkisa:
### Description
Dev data for intake has been broken. This is because we added use of a BGS service that was not available in Fakes::BGSService because we were calling it directly from the bgs code in ruby-bgs.

This adds the method to our BGS service (real and fake).

### DIFF
--- a/app/services/external_api/bgs_service.rb
+++ b/app/services/external_api/bgs_service.rb
@@ -88,6 +88,19 @@ class ExternalApi::BGSService
     }
   end
 
+  def fetch_person_info_by_ssn(ssn)
+    DBService.release_db_connections
+
+    @people_by_ssn[ssn] ||=
+      MetricsService.record("BGS: fetch person by ssn: #{ssn}",
+                            service: :bgs,
+                            name: "people.find_by_ssn") do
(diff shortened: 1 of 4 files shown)


pkarman:
I noticed this too and fixed it on a local branch. I think there's some duplicate code that can be removed here. I'm fine to merge this as-is and I can clean up in my PR.

app/services/external_api/bgs_service.rb
```diff
@@ -88,6 +88,19 @@ def fetch_person_info(participant_id)
...
+    DBService.release_db_connections
+
+    @people_by_ssn[ssn] ||=
+      MetricsService.record("BGS: fetch person by ssn: #{ssn}",
+                            service: :bgs,
+                            name: "people.find_by_ssn") do
+        client.people.find_by_ssn(ssn)
+      end
+
+    @people_by_ssn[ssn]
```
nanotone: I think this line isn't needed?

pkarman:
my PR #12895

pkarman:
let's close this since my PR merged already.

PR #12904, opened 2019-12-07
flakey test fix

tomas-nava:
Connects #10516

### Description
Fixes a flakey feature test by expecting a section title and making sure we're clicking the button we mean to click.

### DIFF
--- a/spec/feature/queue/colocated_task_queue_spec.rb
+++ b/spec/feature/queue/colocated_task_queue_spec.rb
@@ -65,9 +65,10 @@ RSpec.feature "ColocatedTask", :all_dbs do
       # Click into case details page. Expect to see draft decision option.
       click_on(appeal.veteran.name.formatted(:readable_full))
       # verify that the instructions from the VLJ appear on the case timeline
+      expect(page).to have_content("Case Timeline")
       scroll_to("#case_timeline-section")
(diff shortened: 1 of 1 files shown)


spec/feature/queue/colocated_task_queue_spec.rb
```diff
@@ -8,22 +8,14 @@
...
-
     let!(:atty_task) do
-      create(
-        :ama_attorney_task,
-        appeal: appeal,
-        parent: root_task,
-        assigned_by: judge_user,
-        assigned_to: attorney_user
-      )
+      create(:ama_attorney_task, appeal: appeal, parent: root_task, assigned_by: judge_user, assigned_to: attorney_user)
```
pkarman: nit: the multi-line `let` are often there because it makes it easier to add attributes later on refactor. if we were to add just a few more characters then line 16 would probably trigger rubocop rule.

tomas-nava: You're right, it would've triggered robocop with 1 more character. I returned it to multi-line with [this commit](https://github.com/department-of-veterans-affairs/caseflow/pull/12904/commits/d60ac1ffd7cad654404750b977f6efd349d01e13).

Issue #12680, opened 2019-11-12
Inactive JudgeTeams


lomky:
Right now there are ~25 JudgeTeams with no members. We should implement a way to show these teams are inactive, and when querying JudgeTeams in the app, only request active teams.

# AC
 - [x] The codebase recognizes Inactive JudgeTeams
 - [x] Code requesting lists of JudgeTeams only receives active teams
 - [ ] Existing empty JudgeTeams in prod are fixed
 - [x] File follow on ticket to hook up Inactivation of JudgeTeams linked to Judge inactivation.

lomky:
Fell out from #12421, see output from the run [on slack](https://dsva.slack.com/archives/CJL810329/p1573574528009800)

lomky:
_[What is this graph?](https://github.com/department-of-veterans-affairs/caseflow/issues/11533)_
```
1 | 
2 | ||||||||||||||
3 | ||||
5 | 
8 | 
```

What's the downside to keeping these? Nothing other than cruft. Maybe we want to keep this, as we just never wanna delete anything. Maybe we do an inactive flag instead.Valuable to mark these different somehow. 

Reworking this ticket to be discovery & implementation of marking JudgeTeams that are currently empty as inactive and updating places that pull JudgeTeams to only pull active JudgeTeams.

Forking off the work to handle inactiving current active JudgeTeams to another ticket.

Default scopes on the Org model could make this really easy.

Estimating at a 2. Might be a 3 because gut feeling, general unfamiliarity.

hschallhorn:
#11810 for reference on adding active status to user

kevmo:
There are 4 teams with 1 member, 25 with 0. Going to mark only those with 0 as `inactive`.

hschallhorn:
@kevmo What's left to do on this ticket?

hschallhorn:
Closing! https://dsva.slack.com/archives/CJL810329/p1587499066010100

PR #12905, opened 2019-12-09
Add status field to organizations. 

kevmo:
References #12680

### Description

Add `status` and `status_updated_at` to the `organizations` table.  Defaults existing orgs and newly created orgs to "active".

### Acceptance Criteria
- [ ] Migration doesn't fail
- [ ] All organizations are active after migration
- [ ] Default scope properly selects all active organizations
- [ ] Creating an organization defaults the status to active
- [ ] Updating an organizations's status sets timestamp of change

### Database Changes
Add `status` and `status_updated_at` to the `organizations` table.  Defaults existing orgs and newly created orgs to "active".

* [x] Column comments updated


app/models/organization.rb
```diff
@@ -13,6 +13,13 @@ class Organization < ApplicationRecord
 
   before_save :clean_url
 
+  enum status: {
+    Constants.USER_STATUSES.active.to_sym => Constants.ORGANIZATION_STATUSES.active,
```
pkarman: why mapping `USER_STATUSES` to `ORGANIZATION_STATUSES`?

app/models/organization.rb
```diff
@@ -13,6 +13,13 @@ class Organization < ApplicationRecord
...
+    Constants.USER_STATUSES.inactive.to_sym => Constants.ORGANIZATION_STATUSES.inactive
+  }
+
+  default_scope { where(status: Constants.ORGANIZATION_STATUSES.active) }
```
pkarman: IIRC the `enum` gives you some methods for free:
```ruby
default_scope { active }
```

app/models/organization.rb
```diff
@@ -124,6 +131,10 @@ def ama_task_serializer
     WorkQueue::TaskSerializer
   end
 
+  def update_status!(new_status)
+    update!(status: new_status, status_updated_at: Time.zone.now)
+  end
+
```
pkarman: might want to explicitly override the `enum` methods too to make sure the date is set:
```ruby
def active!
  self.status_updated_at = Time.zone.now
  super
end

def inactive!
  self.status_updated_at = Time.zone.now
  super
end
```

for whichever developer in the future (like me) thinks "oh this is an enum, I can just use the shortcut method"

spec/models/organization_spec.rb
```diff
@@ -31,6 +31,54 @@
...
+        end
+      end
+
+      context "with a valid status" do
+        let(:status) { Constants.ORGANIZATION_STATUSES.inactive }
+
+        it "succeeds and updates the status_updated_at column" do
+          expect(subject).to eq true
+          expect(org.reload.status).to eq status
+          expect(org.status_updated_at.to_s).to eq Time.zone.now.to_s
```
pkarman: why the string comparison of dates?

app/models/organization.rb
kevmo: Ah, that's a typo.

db/migrate/20191206192345_add_status_to_organizations.rb
```diff
@@ -0,0 +1,12 @@
+class AddStatusToOrganizations < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def change
+    add_column :organizations, :status, :string, comment: "Whether organization is active, inactive, or in some other Status."
+    add_column :organizations, :status_updated_at, :datetime, comment: "Track when organization status last changed."
+
+    add_index :organizations, :status, algorithm: :concurrently
+
+     Organization.update_all(status: "active")
```
hschallhorn: This should be `Organization.unscoped.update_all(status: "active")`

app/models/organization.rb
hschallhorn: to be fair, using `update!(status: status)` will also get around this.

spec/models/organization_spec.rb
hschallhorn: Might have been leftover in someone else's (_cough_) tests that were failing due to the differences between how postgres stores and returns time to how `Time.zone.now` returns time

kevmo:
@pkarman Mind taking at look at those RSpec errors?  All 3 are some flavor of 
````ActiveModel::UnknownAttributeError:
  unknown attribute 'status' for ETL::Organization.``

File: db/etl/migrate/20191212221851_org_status_change.rb [hunk 6]
```diff
@@ -0,0 +1,12 @@
+class OrgStatusChange < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def change
+    add_column :organizations, :status, :string, comment: "Whether organization is active, inactive, or in some other Status."
+    add_column :organizations, :status_updated_at, :datetime, comment: "Track when organization status last changed."
+
+    add_index :organizations, :status, algorithm: :concurrently
+
+     Organization.unscoped.all.active!
```
pkarman: no need for this in ETL -- it will sync with whatever is in the main Org table.

app/models/organization.rb
hschallhorn: I'm now also remembering why I opted to use ` update!(status: new_status)`. When this (I assume eventually) gets added to a controller, It's easier to just pass through the status sent from the front end (as we do with `adminFlag` in OrganizationUsers.jsx), rather than checking the status to call `active!` or `inactive!` or having to do multiple checks if we ever add more statuses. Not a blocker on this PR and can be updated if and when we actually have a "make orgs inactive" flow

PR #12906, opened 2019-12-09
Reduce ETL SQL calls by 66%

pkarman:
Optimize ETL Task syncer by caching Users and Organizations in-memory.

### DIFF
--- a/app/models/etl/task.rb
+++ b/app/models/etl/task.rb
@@ -10,15 +10,36 @@ class ETL::Task < ETL::Record
 
     private
 
+    def org_cache(org_id)
+      return if org_id.blank?
+
+      @org_cache ||= {}
+      @org_cache[org_id] ||= Organization.find(org_id)
+    end
+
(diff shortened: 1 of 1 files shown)


app/models/etl/task.rb
```diff
@@ -10,15 +10,36 @@ def origin_primary_key
...
+
     # rubocop:disable Metrics/MethodLength
     # rubocop:disable Metrics/AbcSize
     def merge_original_attributes_to_target(original, target)
       # memoize to save SQL calls
-      appeal = original.appeal
-      assigned_to = original.assigned_to
-      assigned_by = original.assigned_by
+      assigned_to = fetch_assigned_to(original)
+      assigned_by = user_cache(original.assigned_by_id)
```
alisan16: Is a task always assigned by a user type?

pkarman: yes. `assigned_by_id` has no corresponding `assigned_by_type` because it is always a user.

Issue #12411, opened 2019-10-16
Convert ui_hash to serializers (tasks and other)


leikkisa:
### Description
As a tech improvement ticket, in order to improve consistency and code quality, and reduce the size of our models, we can convert ui_hash method to use serializers instead.

### Acceptance criteria
- The following files use serializers instead of ui_hash:
    - business_line_task
    - decision_review_task
    - job_note
    - relationship
- Any new files include top line comments for documentation

related:
https://github.com/department-of-veterans-affairs/caseflow/issues/12408
https://github.com/department-of-veterans-affairs/caseflow/issues/12409
https://github.com/department-of-veterans-affairs/caseflow/issues/12355
https://github.com/department-of-veterans-affairs/caseflow/issues/12410

PR #12914, opened 2019-12-09
Convert ui_hash to serializers (tasks and other)

Sjones352:
Connects #12411


app/serializers/intake/job_note_serializer.rb
```diff
@@ -0,0 +1,13 @@
...
+class Intake::JobNoteSerializer
+  include FastJsonapi::ObjectSerializer
+
+  attribute :id
+  attribute :created_at
+  attribute :note
+
+  attribute :user do |object|
+    object.user.css_id
+  end
```
leikkisa: I think this is missing the send_to_intake_user attribute

PR #12915, opened 2019-12-09
AMA Quality Review percentage update & cap

### LINKED ISSUE #12867: Need to update AMA case QR code so that it replicates Legacy case QR code
## User story
User story: As a member of the Quality Review (QR) team at the Board, I need the AMA quality review monthly limit and selection probability to match was is currently set in Caseflow for Legacy, so that I can reliably receive 130+ AMA cases to review monthly, up to 137 case per month.

130+ of each, both capped at 137...code for legacy and AMA should be identical when this work is complete.

## Acceptance criteria
- [x] The monthly limit of Quality Reviews is set at 137
- [x] The Quality Review selection probability is set at 0.032

## Release notes
Caseflow now supports sending approximately 130+ AMA cases to QR for review monthly.

### Out of scope
1. Changes to legacy QR code.

### Background/context
It is believed that when AMA QR code was initially introduced to Caseflow in early 2019 that there would be a limited "ramp up" period during which AMA inventory would accumulate to levels sufficient enough to facilitate random selection and routing of AMA cases to the QR team at the Board for review at roughly 130+ cases per month, similar to what is currently coded for Legacy. After minimum AMA inventory was high enough to support sending 130+ monthly to QR, it was intended that AMA QR code would be updated to ensure 130+ would be sent each month thereafter. This work didn't happen. No change was made to the initial AMA QR code. As a result, the Board is currently seeing only 40+ cases be randomly selected and routed to QR for review monthly. This ticket is intended to correct this and bring the Board back into compliance with its QR policies ASAP. 

### Resources/other links
Legacy code lives here: https://github.com/department-of-veterans-affairs/caseflow/blob/555ac6727ea665985dd42a2f325cb498686c68f0/app/models/judge_case_review.rb

Current AMA code lives here: https://github.com/department-of-veterans-affairs/caseflow/blob/555ac6727ea665985dd42a2f325cb498686c68f0/app/models/quality_review_case_selector.rb

Relevant Slack thread: https://dsva.slack.com/archives/C2ZAMLK88/p1572561323019500

lomky:
Resolves #12867 

### Description
Creates functionality to make AMA behave the same as legacy for selecting and capping QR tasks.

### Acceptance Criteria
- [x] AMA and Legacy QR selection uses the same numbers
- [x] A larger refactor ticket is filed
- [x] Tests are written for this code!

### DIFF
--- a/app/models/quality_review_case_selector.rb
+++ b/app/models/quality_review_case_selector.rb
@@ -1,9 +1,20 @@
 # frozen_string_literal: true
 
 class QualityReviewCaseSelector
-  QUALITY_REVIEW_SELECTION_PROBABILITY = 0.04
+  # AMA Limit and Probability. See Legacy numbers at app/models/judge_case_review.rb:28
+  # As of Dec 2019, we want AMA and Legacy to use the same numbers
+  MONTHLY_LIMIT_OF_QUAILITY_REVIEWS = 137
+  QUALITY_REVIEW_SELECTION_PROBABILITY = 0.032
 
-  def self.select_case_for_quality_review?
-    rand < QUALITY_REVIEW_SELECTION_PROBABILITY
+  class << self
+    def select_case_for_quality_review?
+      reached_monthly_limit_in_quality_reviews? ? false : rand < QUALITY_REVIEW_SELECTION_PROBABILITY
+    end
+  end
+
+  private
+
+  def reached_monthly_limit_in_quality_reviews?
+    QualityReviewTask.this_month.size >= MONTHLY_LIMIT_OF_QUAILITY_REVIEWS
   end
 end
--- a/app/models/quality_review_case_selector.rb
+++ b/app/models/quality_review_case_selector.rb
@@ -10,11 +10,9 @@ class QualityReviewCaseSelector
     def select_case_for_quality_review?
       reached_monthly_limit_in_quality_reviews? ? false : rand < QUALITY_REVIEW_SELECTION_PROBABILITY
     end
-  end
 
-  private
-
-  def reached_monthly_limit_in_quality_reviews?
-    QualityReviewTask.this_month.size >= MONTHLY_LIMIT_OF_QUAILITY_REVIEWS
+    def reached_monthly_limit_in_quality_reviews?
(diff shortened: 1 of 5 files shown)


app/models/quality_review_case_selector.rb
```diff
@@ -1,9 +1,18 @@
 # frozen_string_literal: true
 
 class QualityReviewCaseSelector
-  QUALITY_REVIEW_SELECTION_PROBABILITY = 0.04
+  # AMA Limit and Probability. See Legacy numbers at app/models/judge_case_review.rb:28
+  # As of Dec 2019, we want AMA and Legacy to use the same numbers
+  MONTHLY_LIMIT_OF_QUAILITY_REVIEWS = 137
+  QUALITY_REVIEW_SELECTION_PROBABILITY = 0.032
```
hschallhorn: We wouldn't necessarily want this to be the same number correct? As we don't see the same number of completed ama cases. Regardless, this is higher than what we would probably need, so they'll hit their monthly limit anyway, so maybe just ignore me.

app/models/tasks/quality_review_task.rb
```diff
@@ -4,6 +4,8 @@
 # Task to track when an appeal has been randomly selected to be quality reviewed by the Quality Review team.
 
 class QualityReviewTask < Task
+  scope :this_month, -> { where(created_at: Time.zone.now.beginning_of_month..Time.zone.now.end_of_month) }
```
hschallhorn: Can we update this to `created_this_month` to be a bit more clear?

spec/models/tasks/quality_review_task_spec.rb
```diff
@@ -74,4 +74,16 @@
       expect(qr_task.status).to eq(Constants.TASK_STATUSES.assigned)
     end
   end
+
+  describe "after reaching the monthly limit" do
```
hschallhorn: Seems like this should be implemented in quality_review_case_selector_spec.rb

lomky:
@jimruggiero This is now set to merge and should be released tomorrow morning!

jimruggiero:
Thank you so much!

Issue #12210, opened 2019-09-25
Improve messaging on edit when review is not yet established


leikkisa:
### Story
A user went to case search, and clicked on the link that brings them to the edit page.  The review was still being established.  Currently this brings them to a "Something went wrong" page, with a note that the review is not yet established and a link to the job page.  This is the currently expected behavior, but the messaging is misleading and results in support tickets.

### Acceptance criteria
- [ ] When a review is not yet established, if the user navigates to the edit screen they are directed to a message page instead of the error page.
- [ ] The intake page for adding issues is not impacted

### Copy
Title: Review not editable
Message: The review has not yet been established in VBMS. Check [the job page](link/to/job/page) for details.  You may try to edit the review again once it has been established.

### User Facing Changes
Screenshot of new messaging:
<img width="1096" alt="Screen Shot 2019-12-10 at 16 04 20" src="https://user-images.githubusercontent.com/282869/70568904-d86f2c00-1b66-11ea-8f85-4c36ca934114.png">

### Technical notes
The `validate_prior_to_edit` method in the ClaimReview model throws the error if the claim review is not yet processed.  Instead, it might be good to send `processed?` to the front end via the ui_hash.

The edit page navigates here for both first-time intakes and edits. This should only impact "edit" sessions, where `processed?` is false for the review
`caseflow/client/app/intake/pages/addIssues.jsx`

`caseflow/client/app/intakeEdit/IntakeEditFrame.jsx` redirects some edit attempts to a message page:
`caseflow/client/app/intakeEdit/pages/message.jsx`

`caseflow/app/controllers/claim_review_controller.rb` handles current logic for the failure, with the link to the job page.

leikkisa:
This is the current screen:
![image (5).png](https://images.zenhubusercontent.com/58f6b6eb51aa2c30577bb60e/71ea3caa-b7f3-472b-9a0f-277515f80311)

PR #12918, opened 2019-12-10
Add message for editing a non-established claim review

nanotone:
Connects #12210 

### Description
When a user attempts to edit a ClaimReview (HigherLevelReview or SupplementalClaim) that has not been established yet, display an informative message instead of the current error message.

#### Notes:
With guidance from @jcq I took this opportunity to shift the backend/frontend responsibilities here a bit. Instead of letting `#validate_prior_to_edit` raise an exception and triggering Rails to render an error, the ClaimReview ui_hash is now usable before job processing finishes. This allows the frontend to decide how to render the edit page pre-establishment, with a few consequences:
- The frontend now makes use of the `asyncJobUrl` (which was already in the ui_hash) to assemble the friendly message.
- The frontend uses a new `/edit/not_editable` route in a style that's consistent with the other edit-page states.
- The backend has to be a little more careful about what fields to include in the ui_hash, pre- or post-establishment. This part I don't particularly love, and it shows in the diff for `claim_review.rb`.

### Acceptance Criteria
- [x] When a review is not yet established, if the user navigates to the edit screen they are directed to a message page instead of the error page.
- [x] The intake page for adding issues is not impacted
- [x] The edit page for appeals is not impacted

### Testing Plan
1. With a seeded database, get the UUID of an HLR without an `establishment_processed_at` (using Postgres or ActiveRecord), visit `/higher_level_reviews/<UUID>/edit`, and verify that the new messaging is shown.
1. Update feature tests to verify that the new informational page doesn't have any unexpected error messaging.
1. Existing feature tests already thoroughly exercise the add/edit-issue UI for new intakes, and for existing appeals.

### User Facing Changes
 - [x] Screenshots of UI changes added to PR & Original Issue

Screenshot of new messaging:
Screen Shot 2019-12-10 at 16 04 20

### DIFF
--- a/app/models/claim_review.rb
+++ b/app/models/claim_review.rb
@@ -24,7 +24,6 @@ class ClaimReview < DecisionReview
   self.abstract_class = true
 
   class NoEndProductsRequired < StandardError; end
-  class NotYetProcessed < StandardError; end
 
   class << self
     def find_by_uuid_or_reference_id!(claim_id)
--- a/app/models/claim_review.rb
+++ b/app/models/claim_review.rb
@@ -52,13 +52,13 @@ class ClaimReview < DecisionReview
   end
 
   def validate_prior_to_edit
-    fail NotYetProcessed unless processed?
+    if processed?
+      # force sync on initial edit call so that we have latest EP status.
+      # This helps prevent us editing something that recently closed upstream.
+      sync_end_product_establishments!
 
-    # force sync on initial edit call so that we have latest EP status.
-    # This helps prevent us editing something that recently closed upstream.
-    sync_end_product_establishments!
-
-    verify_contentions
+      verify_contentions
+    end
 
     # this will raise any errors for missing data
     ui_hash
--- a/app/models/claim_review.rb
+++ b/app/models/claim_review.rb
@@ -45,10 +45,15 @@ class ClaimReview < DecisionReview
     super.merge(
       asyncJobUrl: async_job_url,
       benefitType: benefit_type,
-      payeeCode: payee_code,
-      hasClearedRatingEp: cleared_rating_ep?,
-      hasClearedNonratingEp: cleared_nonrating_ep?
-    )
+      payeeCode: payee_code
+    ).tap do |hash|
+      if processed?
+        hash.update(
+          hasClearedRatingEp: cleared_rating_ep?,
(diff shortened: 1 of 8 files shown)


app/models/claim_review.rb
```diff
@@ -45,20 +45,25 @@ def ui_hash
     super.merge(
       asyncJobUrl: async_job_url,
       benefitType: benefit_type,
-      payeeCode: payee_code,
-      hasClearedRatingEp: cleared_rating_ep?,
-      hasClearedNonratingEp: cleared_nonrating_ep?
-    )
+      payeeCode: payee_code
+    ).tap do |hash|
```
Sjones352: As much as I want to break this into a serializer method, I think it looks good.

nanotone: One day very soon... 😉

jcq:
Frontend looks good, and whole thing seems to work, per the tests. Good work on incorporating the necessary route definition and redirect into the existing structure.

app/models/claim_review.rb
leikkisa: Haha

spec/feature/intake/supplemental_claim/edit_spec.rb
```diff
@@ -690,6 +690,8 @@ def click_cancel(visit_page)
         visit "#{url_path}/#{decision_review.uuid}/edit"
 
         expect(page).to have_content("Review not yet established in VBMS. Check the job page for details.")
+        expect(page).not_to have_content("Something went wrong")
```
leikkisa: I think it'd be better to check for the new title "Review not editable" than to check that the former content is absent.

app/models/claim_review.rb
```diff
@@ -45,20 +45,25 @@ def ui_hash
...
+        hash.update(
+          hasClearedRatingEp: cleared_rating_ep?,
+          hasClearedNonratingEp: cleared_nonrating_ep?
+        )
+      end
+    end
   end
 
   def validate_prior_to_edit
-    fail NotYetProcessed unless processed?
```
leikkisa: I think since this error isn't getting raised anywhere anymore, it can be removed from line 7 on the ClaimReviewController, and 27 on ClaimReview.

PR #12921, opened 2019-12-10
hotfix for colocated action count mismatch

yoomlam:
Resolves bug introduced in PR #12868 

### Description
See https://github.com/department-of-veterans-affairs/caseflow/pull/12868/files#r356064882

`ColocatedTask.subclasses.count` returns 19.
`Constants::CO_LOCATED_ADMIN_ACTIONS.count` returns 18.
Causing failure: https://circleci.com/gh/department-of-veterans-affairs/caseflow/96078

The list of colocated tasks offered to the user is manually maintained. Not all subclasses of `ColocatedTask` should be offered to the user.

Remember to update both:
- `client/constants/CO_LOCATED_ADMIN_ACTIONS.json`
- `client/constants/CO_LOCATED_ADMIN_ACTION_CLASSES.json`

### Acceptance Criteria
- [ ] Code compiles correctly and tests pass


app/repositories/task_action_repository.rb
```diff
@@ -203,10 +203,10 @@ def legacy_and_colocated_task_add_admin_action_data(_task, _user)
       {
         redirect_after: "/queue",
         selected: nil,
-        options: ColocatedTask.subclasses.map do |subclass|
+        options: Constants::CO_LOCATED_ADMIN_ACTION_CLASSES.map do |key, value|
```
hschallhorn: We could continue to use the `CO_LOCATED_ADMIN_ACTIONS` and `ColocatedTask.find_subclass_by_action(key)` to map values we already have to their classes without creating a new file.

hschallhorn: Because `PreRoutedColocatedTask`'s label is not in `CO_LOCATED_ADMIN_ACTIONS.json`, it should not be returned

PR #12922, opened 2019-12-10
Clean up feature toggles

Sjones352:
Connects #9590


hschallhorn:
I’m not seeing any place on the queue front end where `featureToggles` is actually _used_. We pull it out of the state in a few places, but don't actually ever look at the toggles. My gut feeling is to remove it. We check certain feature toggles mostly on the back end at this point and pass this with other objects (such as `use_task_pages_api` passed to the front end in `queue_config`). We’ve been making a large effort towards using `queue_config` as the one stop shop for all queue related “settings” and will most likely continue to do so in the future.

TLDR: I agree with removing `featureToggles` from the front end queue code.

kevmo:
I don't feel super-strongly, but my take is:

FeatureToggles are a thing we will certainly be using in the future, if I'm not mistaken. I think it makes sense to keep the field and point to an empty hash - otherwise, we will probably wind up re-adding it.

yoomlam:
> FeatureToggles are a thing we will certainly be using in the future, if I'm not mistaken. I think it makes sense to keep the field and point to an empty hash - otherwise, we will probably wind up re-adding it.

Re-adding it for Queue: https://dsva.slack.com/archives/CQCUMFA0H/p1585145509009300

PR #12924, opened 2019-12-10
Enable caseflowdemo caching for fakes, etc

pkarman:
Turn on redis caching on caseflowdemo.

We'll see this is working of the /test/users page loads more quickly after the initial view.


pkarman:
Yes the functionality is already in the code at https://github.com/department-of-veterans-affairs/caseflow/blob/master/config/environments/development.rb#L25

Issue #12824, opened 2019-11-27
Duplicate tasks causing case to return to user's queue


alisan16:
See discussion here: https://dsva.slack.com/archives/CHX8FMP28/p1574876465434700

Before resolution
```
Appeal 13066 [id, status, assigned_by_id, assigned_to_id, created_at, closed_at]
└── RootTask 290215, completed, (assigned_by_id), 2, 2019-07-24 15:18:35 UTC, 2019-11-20 18:51:23 UTC
    ├── TrackVeteranTask 290216, completed, (assigned_by_id), 286, 2019-07-24 15:18:36 UTC, (closed_at)
    ├── DistributionTask 290217, completed, (assigned_by_id), 2, 2019-07-24 15:18:36 UTC, 2019-09-13 19:43:34 UTC
    ├── JudgeAssignTask 406723, cancelled, (assigned_by_id), 915, 2019-09-13 19:43:34 UTC, 2019-09-17 12:58:59 UTC
    ├── JudgeAssignTask 413322, completed, 915, 812, 2019-09-17 12:58:59 UTC, 2019-09-20 16:42:46 UTC
    ├── JudgeDecisionReviewTask 423685, cancelled, (assigned_by_id), 812, 2019-09-20 16:42:46 UTC, 2019-10-29 18:05:46 UTC
    │   └── AttorneyTask 423688, cancelled, 812, 711, 2019-09-20 16:42:46 UTC, 2019-10-29 18:05:46 UTC
    ├── JudgeDecisionReviewTask 423691, completed, (assigned_by_id), 812, 2019-09-20 16:44:17 UTC, 2019-11-18 15:37:04 UTC
    │   └── AttorneyTask 423692, completed, 812, 711, 2019-09-20 16:44:17 UTC, 2019-10-28 16:06:41 UTC
    ├── JudgeAssignTask 494336, on_hold, (assigned_by_id), 812, 2019-10-29 18:05:46 UTC, (closed_at)
    │   ├── OtherColocatedTask 495782, completed, 812, 23, 2019-10-30 14:55:37 UTC, 2019-11-26 19:26:11 UTC
    │   │   └── OtherColocatedTask 495783, completed, 812, 1782, 2019-10-30 14:55:44 UTC, 2019-11-26 19:26:11 UTC
    │   │       └── TimedHoldTask 508609, completed, 1782, 1782, 2019-11-07 13:02:33 UTC, 2019-11-22 13:30:17 UTC
[...]

PR #12925, opened 2019-12-10
raise error when creating second open JudgeDecisionReviewTask

### LINKED ISSUE #12802: Prevent creation of redundant JudgeDecisionReviewTasks
## Description
Occasionally when we create the JudgeDecisionReviewTask & its child AttorneyTask, we do this twice. This can lead to confusion on the part of the Attorney and/or Judge this is assigned to, and will have unknown behavior if the Judge pushes through checkout twice.

Tomas has done some research and found that we are doing two distinct POSTs, creating the redundant task branch.  

[Investigation](https://dsva.slack.com/archives/CJL810329/p1573842570123600)
[POST discussion](https://dsva.slack.com/archives/CAM9FJ85P/p1573866025061200)

Possible places to look:
- is the front end allowing multiple form submissions?
- is the front end mistakenly sending multiple POST requests for a single form submission?
- is something in our networking setup glitching and duplicating POST requests?

## Acceptance Criteria
 - [ ] Any outstanding affected appeals are corrected in Prod
 - [ ] Root cause of the double DecisionReview branch is addressed

yoomlam:
Resolves #12802 and probably #12824

### Description
Prevent duplicate open JudgeDecisionReviewTask since there should only ever be one open JudgeDecisionReviewTask at a time for an appeal.

Added `verify_user_task_unique` check `before_create` of `JudgeDecisionReviewTask` instances.

Added new tests.

### Acceptance Criteria
- [ ] Code compiles correctly and tests pass


app/models/tasks/judge_decision_review_task.rb
```diff
@@ -16,4 +18,23 @@ def additional_available_actions(_user)
   def self.label
     COPY::JUDGE_DECISION_REVIEW_TASK_LABEL
   end
+
+  # Use the existence of another open JudgeDecisionReviewTask to prevent duplicates since there should only
+  # ever be one open JudgeDecisionReviewTask at a time for an appeal.
+  def verify_user_task_unique
+    return if !open?
```
lomky: is this checking if the task being created is open?

yoomlam: Yes, copied it from https://github.com/department-of-veterans-affairs/caseflow/blob/3f8f454f0ed930ff381ebfd10eb91a6b58b089be/app/models/task.rb#L213

spec/models/tasks/judge_task_spec.rb
```diff
@@ -124,7 +124,51 @@
...
+      let!(:jdrt) do
+        create(:ama_judge_decision_review_task, appeal: root_task.appeal, parent: root_task, assigned_to: judge)
+      end
+
+      subject { JudgeDecisionReviewTask.create!(appeal: root_task.appeal, parent: root_task, assigned_to: judge) }
+
+      context "when an open (assigned) JudgeDecisionReviewTask already exists" do
+        it "should fail creation of second JudgeDecisionReviewTask" do
+          expect { subject }.to raise_error(Caseflow::Error::DuplicateUserTask)
+          expect(root_task.appeal.tasks.count).to eq(2), root_task.appeal.tasks.to_a.to_s
```
lomky: nit pick: I think I'd prefer we stored how many tasks exist on the appeal in a variable before calling `subject`, and then check it hasn't changed after? That feels more robust if we ever change the factory, and gives a meaningful name over just a number. (ditto across the rest of the tasks). I really do like being robust to check the task tree hasn't changed!

yoomlam: Done

spec/models/tasks/judge_task_spec.rb
```diff
@@ -127,42 +127,32 @@
...
-          expect(root_task.appeal.tasks.count).to eq(2), root_task.appeal.tasks.to_a.to_s
-        end
-      end
-
-      context "when an open (on_hold) JudgeDecisionReviewTask already exists" do
-        it "should fail creation of second JudgeDecisionReviewTask" do
-          jdrt.update(status: Constants.TASK_STATUSES.on_hold)
-          expect { subject }.to raise_error(Caseflow::Error::DuplicateUserTask)
-          expect(root_task.appeal.tasks.count).to eq(2), root_task.appeal.tasks.to_a.to_s
+      Task.open_statuses.each do |o_status|
```
lomky: Beautiful! :tada:

PR #12933, opened 2019-12-10
Add Confirmation / Cancellation Copy

### LINKED ISSUE #12330: Create ActionMailers for Virtual Hearing Emails
### Background
related to: #12329
When a hearing is switched from central or video to virtual, a confirmation email with instructions will need to be sent to veteran, representative, and judge. 

### Objective
Create Mailers and Mailer layouts for 
- Notice that virtual hearing was cancelled to Judge, Veteran, and Representativee
- Confirmation that virtual hearing was scheduled

[reference](https://guides.rubyonrails.org/action_mailer_basics.html)

### Acceptance Criteria
- [ ] Verify that email is sent to test recipient in UAT
- [ ] Verify calendar invite is attached to email

lomaxap:
resolves #12330


app/models/hearings/forms/base_hearing_update_form.rb
```diff
@@ -40,7 +40,10 @@ def virtual_hearing_form_or_hearing_time_was_updated?
   def start_async_job
     if hearing.virtual_hearing.status == "pending" || !hearing.virtual_hearing.all_emails_sent?
       hearing.virtual_hearing.establishment.submit_for_processing!
-      VirtualHearings::CreateConferenceJob.perform_now(hearing_id: hearing.id)
+      VirtualHearings::CreateConferenceJob.perform_now(
+        hearing_id: hearing.id,
+        time_changed: scheduled_time_string.present? && !created?
+      )
```
lomaxap: don't love having to pass this flag down some many steps. open to suggestions

app/views/virtual_hearing_mailer/cancellation.html.erb
```diff
@@ -1,3 +1,24 @@
-Hi <%= @recipient.full_name %>,
+Dear <%= @recipient.name %>,
```
ferristseng: Should this be wrapped in a `<p>` similar to the confirmation email?

app/views/virtual_hearing_mailer/confirmation.html.erb
```diff
@@ -1,3 +1,107 @@
-Hi <%= @recipient.full_name %>,
+<p>
+  Dear<% if @recipient.title == VirtualHearingMailer::RECIPIENT_TITLES[:judge] %>Judge<% end %> <%= @recipient.name %>,
```
ferristseng: How does this display? Looks like it might need whitespace between "Dear" and "Judge", but not sure.

app/views/virtual_hearing_mailer/confirmation.html.erb
```diff
@@ -1,3 +1,107 @@
...
+      no application download is required. The hearing will launch automatically in your web browser after you click the link.
+    </li>
+  </ul>
+<% end %>
+
+<h3>
+  Help Desk
+</h3>
+<p>
+  If you are experiencing technical difficulties, contact the National Telehealth Technology HelpDesk (NTTHD) at 855-519-7116 to get support.
```
ferristseng: Missing space between "HelpDesk"?

app/views/virtual_hearing_mailer/confirmation.html.erb
```diff
@@ -1,3 +1,107 @@
-Hi <%= @recipient.full_name %>,
+<p>
+  Dear<% if @recipient.title == VirtualHearingMailer::RECIPIENT_TITLES[:judge] %>Judge<% end %> <%= @recipient.name %>,
+</p>
 
-Your virtual hearing on <%= "virtual_hearing_date" %> is confirmed.
+<% if @recipient.title == VirtualHearingMailer::RECIPIENT_TITLES[:representative] %>
```
ferristseng: I find this a bit difficult to know what the final email will look like just based off this file because of all the branching. Maybe there could be separate templates for the confirmation and time changed confirmation? I don't mind it if some of the copy is duplicated. Not a super urgent fix either way.

app/mailers/virtual_hearing_mailer.rb
```diff
@@ -14,12 +14,33 @@ class VirtualHearingMailer < ActionMailer::Base
...
+    @link = link
+    @time_changed = time_changed
+    mail(to: recipient.email, subject: confirmation_subject)
+  end
+
+  def confirmation_subject
+    return "Updated time: Your virtual hearing with the Board of Veterans' Appeals" if @time_changed
+
+    case recipient.title
+    when RECIPIENT_TITLES[:veteran], RECIPIENT_TITLES[:veteran]
```
ferristseng: Should one of these be for the representative? Looks like both cases are for the veteran

app/mailers/virtual_hearing_mailer.rb
```diff
@@ -14,12 +14,33 @@ class VirtualHearingMailer < ActionMailer::Base
...
+    when RECIPIENT_TITLES[:judge]
+      "Confirmation: Your virtual hearing"
+    end
+  end
+
+  def link
+    (recipient.title == RECIPIENT_TITLES[:judge]) ? virtual_hearing.host_link : virtual_hearing.guest_link
+  end
+
+  def veteran_full_name
```
ferristseng: Don't think this is being used anywhere

app/models/hearings/forms/base_hearing_update_form.rb
ferristseng: Hmm, yeah I don't love passing the flag down either. One option could be creating a different job to handle the time changed scenario. I think changing the time doesn't fall under creating the conference, so maybe the jobs could also be distinct.

lomaxap: i like this idea. I think we can split making another job into a follow-on PR. I've split the `updated_time_confirmation` into a separate mailer method to make that change eaiser

PR #12939, opened 2019-12-11
HearingsShowRate Metric

### LINKED ISSUE #12390: Instrument Quarterly Hearings Show Rate metric
This metric is required by OIT (Metric ID: 1812216039) and is critical to understand how the Caseflow Hearings products are helping to improve the appeals hearings process.

Metric definition: (Total scheduled meetings held – postposed hearings) / total scheduled meetings from Caseflow Hearing Schedule

Example SQL for AMA appeals from https://caseflow-looker.va.gov/looks/193
```
SELECT 
	hearings.disposition  AS "hearings.disposition",
	COUNT(DISTINCT appeals.id ) AS "appeals.count"
FROM public.hearings  AS hearings
LEFT JOIN public.appeals  AS appeals ON hearings.appeal_id = appeals.id 
LEFT JOIN public.hearing_days  AS hearing_days ON hearings.hearing_day_id = hearing_days.id 

WHERE 
	(((hearing_days.scheduled_for ) >= (DATE(DATE '2019-06-01')) AND (hearing_days.scheduled_for ) < (DATE(DATE '2019-06-30'))))
GROUP BY 1
ORDER BY 2 DESC
LIMIT 500

-- sql for creating the total and/or determining pivot columns
SELECT 
	COUNT(DISTINCT appeals.id ) AS "appeals.count"
FROM public.hearings  AS hearings
LEFT JOIN public.appeals  AS appeals ON hearings.appeal_id = appeals.id 
LEFT JOIN public.hearing_days  AS hearing_days ON hearings.hearing_day_id = hearing_days.id 

WHERE 
	(((hearing_days.scheduled_for ) >= (DATE(DATE '2019-06-01')) AND (hearing_days.scheduled_for ) < (DATE(DATE '2019-06-30'))))
LIMIT 1
```

Once this metric is instrumented, ensure delivery to OIT with #12389

lomaxap:
Resolves #12390

### DIFF
--- a/app/services/metrics/date_range.rb
+++ b/app/services/metrics/date_range.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class Metrics::DateRange
+  attr_reader :start_date, :end_date
+
+  def initialize(start_date, end_date)
+    @start_date = start_date
+    @end_date = end_date
+  end
+end
--- a/app/services/metrics/base.rb
+++ b/app/services/metrics/base.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 class Metrics::Base
-  self.abstract_class = true
   attr_reader :date_range
 
   delegate :start_date, :end_date, to: :date_range
--- a/app/services/metrics/base.rb
+++ b/app/services/metrics/base.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+class Metrics::Base
+  self.abstract_class = true
+  attr_reader :date_range
+
+  delegate :start_date, :end_date, to: :date_range
+
+  def initialize(date_range)
+    @date_range = date_range
+  end
+
+  def call; end
+
+  def id; end
+
+  def name; end
+end
--- a/app/services/metrics/hearings_show_rate.rb
+++ b/app/services/metrics/hearings_show_rate.rb
@@ -6,7 +6,7 @@
 
 class Metrics::HearingsShowRate < Metrics::Base
   def call
-    (hearings_by_disposition["held"] - hearings_by_disposition["postponed"]) / hearings.count
+    (hearings_by_disposition["held"] - hearings_by_disposition["postponed"]) / hearings.count.to_f
   end
 
   def name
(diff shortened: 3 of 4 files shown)


alisan16:
Can we gut check this on the past 3 months for example? Back in 2018, this metric was about 75% (see page 16 of this file: https://drive.google.com/file/d/1iw3QTGzy0I5lZx_xeWwvqkmPEvPyxfQc/view)

lomaxap:
the number is 49% for this implementation. The language in the PDF

> The percentage of scheduled hearings that are held, omitting postponed hearings.

makes it sound like postponed hearings should be excluded from the query all together, so the definition would be

`held hearings / (total hearings - postponed hearings)`

That number is 79%.

app/services/metrics/date_range.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class Metrics::DateRange
+  attr_reader :start_date, :end_date
```
pkarman: there are subtleties to using `Time` objects for date ranges, because `Time` will default to midnight.

e.g. https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/jobs/monthly_metrics_report_job.rb#L13

So while this particular implementation is fine as is, folks who use code like this:

`range = Metrics::DateRange.new(start_date: Time.zone.yesterday, end_date: Time.zone.today)`

will get values that effectively cover one 24-hour period (yesterday at midnight till today at midnight) when piped into a SQL query.

So, you may want to do some validation here that the arguments are of an expected object type. Otherwise you get subtle bugs where you are actually measuring 29 days out of a 30-day month.

alisan16:
Great that looks much more reasonable! @D-L-Ware fyi - I'll follow up with Susan Kagan to update the metric definition because it looks like the guideline they gave us was wrong.

app/services/metrics/date_range.rb
lomaxap: updated to coerce all inputs to dates and validate inputs

D-L-Ware:
Agree with the changes it makes sense to take postponements out as we ultimately do not know if they will show up or not in the end.  A postponement is a veteran taking advantage of there legal write to ask for a rescheduled hearing.

app/services/metrics/date_range.rb
alisan16: Is there a different way to validate the dates rather than .respond_to? It's throwing a CodeClimate smell now. Fine with ignoring if this is the easiest way, but just wanted to check. https://codeclimate.com/github/department-of-veterans-affairs/caseflow/pull/12939

pkarman: I think `try(:to_date)` would give you what you want?

Issue #12355, opened 2019-10-10
Convert ui_hash to serializers (decision review files)


leikkisa:
### Description
As a tech improvement ticket, in order to improve consistency and code quality, and reduce the size of our models, we can convert ui_hash method to use serializers instead.

Acceptance criteria
- Decision review files use serializers instead of ui_hash:
    - decision_review
    - claim_review
    - higher_level_review
    - supplemental_claim
    - appeal
- Any new files include top line comments for documentation

related:
https://github.com/department-of-veterans-affairs/caseflow/issues/12408
https://github.com/department-of-veterans-affairs/caseflow/issues/12409

PR #12948, opened 2019-12-12
Convert ui_hash to serializers (decision review files)

Sjones352:
Connects  #12355


app/models/claim_review.rb
```diff
@@ -40,19 +40,8 @@ def find_all_visible_by_file_number(*file_numbers)
     end
   end
 
-  def ui_hash
-    super.merge(
-      asyncJobUrl: async_job_url,
-      benefitType: benefit_type,
-      payeeCode: payee_code
-    ).tap do |hash|
```
Sjones352: Not sure how I can use tap in the new serializer I just implemented.

Sjones352: Fixed

app/models/claim_review.rb
```diff
@@ -259,18 +248,18 @@ def can_contest_rating_issues?
...
 
-  private
-
-  def cleared_end_products
-    @cleared_end_products ||= end_product_establishments.select { |ep| ep.status_cleared?(sync: true) }
-  end
-
   def cleared_rating_ep?
-    cleared_end_products.any?(&:rating?)
+    processed? && cleared_end_products.any?(&:rating?)
```
leikkisa: I'm thinking a better place for this check would be in `EndProductEstablishment.status_cleared?(sync: false)` if the end product establishment isn't yet processed (and leaving processed out of these methods).

app/serializers/intake/appeal_serializer.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class Intake::AppealSerializer < Intake::DecisionReviewSerializer
+  include FastJsonapi::ObjectSerializer
+  set_key_transform :camel_lower
+
+  attribute :docket_type, &:docket_type
```
leikkisa: I think `&:docket_type` is not needed here

app/serializers/intake/decision_review_serializer.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+class Intake::DecisionReviewSerializer
+  include FastJsonapi::ObjectSerializer
+  set_key_transform :camel_lower
+
+  attribute :claimant, &:claimant_participant_id
+  attribute :veteran_is_not_claimant, &:veteran_is_not_claimant
```
leikkisa: I don't think you need the `&:attribute_name` here unless it gets a different name in the serializer. So it makes sense for some of these, like `claimant_participant_id`, and `processed_in_caseflow?`, but for some of these it's redundant

app/serializers/intake/higher_level_review_serializer.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class Intake::HigherLevelReviewSerializer < Intake::ClaimReviewSerializer
+  include FastJsonapi::ObjectSerializer
+  set_key_transform :camel_lower
+
+  attribute :informal_conference, &:informal_conference
+  attribute :same_office, &:same_office
```
leikkisa: Same comment!

app/serializers/intake/supplemental_claim_serializer.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class Intake::SupplementalClaimSerializer < Intake::ClaimReviewSerializer
+  include FastJsonapi::ObjectSerializer
+  set_key_transform :camel_lower
+
+  attribute :is_dta_error, &:decision_review_remanded?
+  attribute :formType do
```
leikkisa: this can be spelled as "form_type" since you have the transform, for consistency

app/serializers/intake/appeal_serializer.rb
```diff
@@ -0,0 +1,13 @@
...
+  attribute :is_outcoded, &:outcoded?
+
+  attribute :formType do
```
nanotone: Could this be `form_type`?

Same question for `HigherLevelReviewSerializer` and `SupplementalClaimSerializer`.

app/models/claim_review.rb
nanotone: Is the concern about a potential race condition where the EP is cleared but the asyncable hasn't gotten its processed-time yet? Or perhaps just a general feeling that these methods ought to be a separate concern from asyncable job lifecycle?

app/models/claim_review.rb
Sjones352: @leikkisa would prefer I do it like this ```attribute :has_cleared_rating_ep do |object|
   object.processed? && object.cleared_rating_ep?
  end```

app/models/claim_review.rb
leikkisa: I was thinking more that if an end product isn't even established, it doesn't make sense to sync it, and all of the status checks would also return false.  I think `EndProductEstablishment.sync!` is actually the right place.

This is a higher level location for this, closer to the "root", and seems more resilient.

app/models/claim_review.rb
aroltsch: I trust Sally's judgement :)

leikkisa:
Just writing a note that we ended up deciding to keep "processed?" in the claim review checks for EPs, but it would be better to put in the end product establishment model and not try to sync if an ep hasn't been established.  However, this unexpectedly impacted RAMP code, which is some interesting tech-debt-y stuff that doesn't make sense to tackle in this PR.

PR #12950, opened 2019-12-12
search for appeal with both ssn and file_number

rubaiyat22:
### Description
- related to issue is appeals-batteam https://dsva.slack.com/archives/CHX8FMP28/p1575986812109200
- AppealFinder now looks for veteran case using both ssn and file_number
- write tests

### DIFF
--- a/spec/feature/queue/search_spec.rb
+++ b/spec/feature/queue/search_spec.rb
@@ -415,7 +415,7 @@ RSpec.feature "Search", :all_dbs do
     end
 
     context "when appeal is associated with ssn not file number" do
-      let!(:veteran) { create(:veteran, ssn: Generators::Random.unique_ssn) }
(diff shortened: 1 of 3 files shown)


pkarman:
change looks good -- tests?

rubaiyat22:
> change looks good -- tests?

Yes, will be writing tests before asking for a review

spec/feature/queue/search_spec.rb
```diff
@@ -413,6 +413,23 @@ def perform_search
         expect(page).to have_content(COPY::IS_PAPER_CASE)
       end
     end
+
+    context "when appeal is associated with ssn not file number" do
+      let!(:veteran) { create(:veteran, ssn: Generators::Random.unique_ssn) }
```
pkarman: I know it's rare, but having a random ssn and a random file_number can eventually lead to a collision where they are they same, and negate the test. Could we have explicit `file_number` and `ssn` values here so it's more clear that they are *not* the same value?

PR #12951, opened 2019-12-12
Disable trailing comma in literal rubocop rule

alisan16:
Disabling this rubocop rule across caseflow repos, which did not allow the use of trailing commas in literals.
See https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/1144

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -142,3 +142,6 @@ Lint/UriEscapeUnescape:
 
 Layout/MultilineMethodCallIndentation:
   EnforcedStyle: indented
+
+Style/TrailingCommaInLiteral:
+  Enabled: false


alisan16:
caseflow uses Rubocop version 0.67.2 whereas caseflow-efolder is using version 0.52.1

If we want to take this one step further and actually enforce trailing commas, we would need to add
```
EnforcedStyleForMultiline: comma
```
to both of the rules

https://www.rubocop.org/en/stable/cops_style/#styletrailingcommainarrayliteral
https://www.rubocop.org/en/stable/cops_style/#styletrailingcommainhashliteral

pkarman:
I am not in favor of requiring the trailing comma. I'd like to see us enforce it by convention, but there are plenty of instances where it makes less sense.

PR #12958, opened 2019-12-12
Metric: Percent of non-denial decisions with an EP created within 7 days

### LINKED ISSUE #12551: Quarterly Metric: Caseflow Dispatch % 7-day non-denial decisions
This metric is required by OIT (Metric ID: 1709076052).

Caseflow Dispatch - Percent of non-denial decisions with an EP created within 7 days
Metric description: Caseflow facilitates the transfer of cases to the Regional Office following issuance of a Board of Veterans’ Appeals decision, with an end product (EP) created to track work. Not creating an EP increases the possibility of miscommunication and slows average processing times, delaying ratings adjustment and working of remand orders.

Metric definition: (Number EPs with date<=7 days after outcoding date/number non-denial decisions)

lomaxap:
resolves #12551 

### notes
- Since `outcoding_date` only applies to Appeals, this metric only pulls decisions where the `decision_review_type` is `Appeal`. 

- `outcoding_date` is assumed to be `bva_dispatch_task.closed_at` and is used for the metric's date range

- For date range `2019-11-29` to `2019-12-29`:  49(`end_products_created_within_7_days_of_outcoding.count`)/ 297 (`non_denial_decisions.count`) = 0.16498


D-L-Ware:
We should use BVA Dispatch date not date of decision.  This metric is a little weird because it is not about our primary customer the Board but about VBA and how quickly they establish an EP in VBMS following a Board decision.  VBA is not notified of the decision until dispatch.  As such, the 7 days should be measured from time of dispatch.

D-L-Ware:
Also, are we accounting for the fact that any decisions dispatched 7 days or less before the end date range should not be included as we would not have data yet on whether they had been established in 7 days.

D-L-Ware:
Finally, it will not be surprising to find this metric returns a low percentage.  It is heavily dependent on VBA resourcing, specifically the AMO, and when we monitored it before it fluctuated greatly because AMO shifts resources a lot.

spec/services/metrics/non_denial_decisions_spec.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+describe Metrics::NonDenialDecisions, :postgres do
+  let(:user) { create(:user) }
+  let(:start_date) { Time.zone.now - 45.days }
+  let(:end_date) { Time.zone.now - 15.days }
+  let(:date_range) { Metrics::DateRange.new(start_date, end_date) }
+  let(:number_of_decisions_in_range) { 25 }
+  let(:number_of_end_products_created_in_7_days) { 10 }
+  before do
```
lomaxap: i couldn't quite figure out a way using existing workflow to get an appeal to a state where there were end product establishments for outcoded BVA Dispatch Tasks. i looked [here](https://github.com/department-of-veterans-affairs/caseflow/blob/master/spec/feature/queue/case_details_spec.rb#L1112-L1137) for reference, and used `BvaDipatchTask.outcode`, but there must be additional steps. may need to revisit this test.

spec/services/metrics/non_denial_decisions_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    BvaDispatchTask.where(status: Constants.TASK_STATUSES.completed).update_all(closed_at: Time.zone.now - 30.days)
+    decision_issues.sample(number_of_end_products_created_in_7_days).each do |decision|
+      create(:end_product_establishment, source: decision.decision_review.decision_documents.first)
+    end
+  end
+
+  subject { Metrics::NonDenialDecisions.new(date_range).call }
+
+  it do
```
pkarman: does this `it` need a title? might be hard to spot in output w/o it?

Issue #12926, opened 2019-12-10
Follow up job messaging work


leikkisa:
## User or job story
User story: As an intake user, I need to be able to communicate to veterans when their end product establishment is failing, through uploading a screenshot of their job details page. I would like the job details page to include the inbox messages in notes, because the inbox messages page includes information for other veterans.

## Acceptance criteria
- [x] When we add an inbox message for a job failing after 24 hours, subsequently succeeding, or being canceled, also add a job note (job note messages are already shown)
- [x] Read Inbox messages should be hidden after 120 days
- [x] Update copy (see below for new requested copy)
- [x] Use veteran file numbers when labeling async jobs in Inbox
- [x] Include a screenshot in this GitHub issue of an inbox message appearing on the job details page

## Release notes
Inbox messages related to asyncable jobs will appear as job notes on the job detail page.

### Designs
New copy:
- For messaging a job failure after 24 hours, add:
>No further action is necessary as the (IT) support team has been notified.  You will receive a separate message in your inbox when the issue has resolved.
- For messaging a job being manually cancelled, add:
>No further action is necessary. Please see the job details page for more information on why this job has been cancelled.

Sample screenshot of job page with new job notes that get automatically created alongside Inbox messages (for failures and successes after 24 hours):
<img width="401" alt="Screen Shot 2019-12-18 at 10 51 19" src="https://user-images.githubusercontent.com/282869/71101457-bc7e1280-2184-11ea-843a-589221e82a2b.png">

leikkisa:
Moving this to "Ready for Dev" since @nanotone is getting a head start on it, with the caveat that we are still waiting for some info from AMO.

PR #12978, opened 2019-12-14
Tweak copy for a few async message types

nanotone:
Connects #12926 

As per AMO's feedback on async job messaging, a few message types are getting slightly fleshed out.

### Designs
Sample screenshot of what the new language will look like, in the inbox:
Screen Shot 2019-12-16 at 16 57 10

### DIFF
--- a/app/services/asyncable_job_messaging.rb
+++ b/app/services/asyncable_job_messaging.rb
@@ -33,7 +33,8 @@ class AsyncableJobMessaging
       job_note = JobNote.create!(job: job, user: current_user, note: text, send_to_intake_user: send_to_intake_user)
       if send_to_intake_user
         message_text = <<-EOS.strip_heredoc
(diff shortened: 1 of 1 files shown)


nanotone:
@leikkisa I seem to recall that you wanted to make some additional tweaks to the requested copy for job cancellations. What are your thoughts on the current copy here?

leikkisa:
@nanotone If we cancel a job, it is usually because there is no need to continue with the EP establishment.  The one case I can remember, the EP actually successfully established and went on through to get cleared, but generating the tracked item failed. Because the EP was already cleared, there was no need to complete processing on that job.  How about we use this copy?

> No further action is necessary.  Please see the job details page for more information on why this job has been canceled.

nanotone:
@leikkisa That makes sense! Updated the code and screenshot, gonna go ahead and mark this as ready-to-merge.

PR #12983, opened 2019-12-16
Disallow the assignment of tasks to inactive users

### LINKED ISSUE #12943: Ensure new tasks are only assigned to active users
Since the ability to mark users inactive was deployed (#12212), we have removed inactive users from organizations that automatically assign tasks to their members (#12303) to ensure they do not get assigned more tasks. We will also remove them from all other organizations to cut down on ways they can be assigned tasks while inactive ( #12940). At task creation, we should validate that the assignee is actually active and task creation should fail if they are inactive.

## AC
- [ ] New tasks cannot be assigned to inactive users

hschallhorn:
Resolves #12943

### Description
Validates that a user is active before creating a task assigned to them. Removes inactive users from round robin distribution.

### Acceptance Criteria
- [ ] Tasks cannot be created assigned to inactive users
- [ ] The assignee pool for the round robin distributer does not include inactive users

#### On task creation
1. Mark a user of the translation team inactive
```ruby
User.find_by(css_id: "ORG_QUEUE_USER_0").inactive!
```
1. Sign in as ORG_QUEUE_USER_1
1. Go to the translation team's queue http://localhost:3000/organizations/translation
1. Click into any case in the unassigned tab
1. Select "Assign to person" from the task action dropdown
1. Attempt to assign to "Translation team member 0"
1. Ensure error
#### Automatic task creation
1. Make a VLJ support team member inactive and check how many tasks are assigned to them
```ruby
user = User.find_by(css_id: "BVALSPORER")
user.inactive!
Task.where(assigned_to: user).count
=> 2
```
1. Sign in as an attorney (such as "BVASCASPER1")
1. Go to the user's queue and click into any case
1. Select "Add admin action from the task action dropdown
1. Add some admin actions, preferably more than 2
1. Ensure none of these were automatically assigned to the user
```ruby
Task.where(assigned_to: user).count
=> 2
```

### User Facing Changes

Screen Shot 2019-12-17 at 11 10 01 AM


app/models/task.rb
```diff
@@ -653,5 +654,13 @@ def status_is_valid_on_create
 
     true
   end
+
+  def assignee_status_is_valid_on_create
+    if assigned_to.is_a?(User) && !assigned_to.active?
```
hschallhorn: specifically checking "not active" over "is inactive" as we may have other statuses in the future

hschallhorn:
@carodew How is the wording on this error? We want to make sure that no tasks are created for a user that is no longer active at the board.

carodew:
Question – if a user is inactive, should we prevent them from showing up in any selection drop downs, thus invalidating the need for this error? Or are we looking to prevent any concurrency errors, as in someone made this user inactive at the same time someone was trying to assign this translation task? 

I would opt for something a little more conversational. Let's say the users' name is Carol King. 

big text: "Uh oh! We're unable to assign this to Carol King."
little text: Carol King is marked as inactive in Caseflow. Please select someone else or [what should they do?] if you believe you're getting this message in error." 

What I don't know is who they should check with / what they should do if they think that Carol King shouldn't be marked as inactive. Submit a helpdesk ticket? Check with this person's manager? 

cc-ing @RyanSibley who is very busy and may not have a chance to weigh in, but if she does, she should feel free to edit the text in this suggestion.

app/models/task.rb
```diff
@@ -537,6 +538,10 @@ def assigned_to_label
     assigned_to.is_a?(Organization) ? assigned_to.name : assigned_to.css_id
   end
 
+  def child_must_have_active_assignee?
+    true
+  end
+
```
hschallhorn: Context on this in [slack](https://dsva.slack.com/archives/C54QCEHAL/p1576602568044100)

PR #12985, opened 2019-12-16
import sqls

### LINKED ISSUE #12068: Migrate Looker reports and alerts
Looker is going to be decommissioned. We need to migrate our existing SQL reports and alerts to some other mechanism.

### Proposal

Move the SQL to individual `.sql` files under the new `app/sql` directory as with #12044 and write a job that runs each one nightly just as Looker currently does.

enriquemanuel:
Resolves #12068

### Description
Since we are deprecating Looker, I found a way to automate and bring all the "Looks" aka SQL

I have created 2 folders:
`dashboards/*`
`looks_sqls/*`

The Dashboards are essentially *looks* grouped by. if you want to recreate a dashboard, you can just use the sql inside or get the queries accordingly.

### Code used
```
# config.ini
# API version is required
api_version=3.1
# Base URL for API. Do not include /api/* in the url
base_url=http://<looker ip or url for api>:19999
# API 3 client id
client_id=<get yours from the Looker Admin page>
# API 3 client secret
client_secret=<get yours from the Looker Admin page>
# Set to false if testing locally aganst self-signed certs. Otherwise leave True
```
```
# creationqueries.py
from looker_sdk import client, models, error
sdk = client.setup('config.ini')
looker_api_user = sdk.me()
looks_ids = list()

for look in sdk.all_looks(fields='id'):
	looks_ids.append(look.id)

for id in looks_ids:
	try:
		sql = sdk.run_look(id, 'sql')
		lookobj = sdk.look(id, fields='title')
		title = lookobj.title
		f = open("sqls/%s.sql"  % (title), "w+")
		f.write(sql)
		
	except:
		continue
	finally:
		f.close()

``

### DIFF
--- a/app/sql/looker/dashboards/AMA Appeals/AMA decisions by attorneys and judges.sql
+++ b/app/sql/looker/dashboards/AMA Appeals/AMA decisions by attorneys and judges.sql
@@ -1,0 +1,155 @@
+WITH appeal_task_status AS (SELECT *,
+    (select max(tasks.updated_at)
+            FROM tasks  AS tasks
+            where tasks.appeal_id = appeals.id
+            limit 1
+          ) as task_max_updated_at,
+          (select tasks.id
+            FROM tasks  AS tasks
+            where tasks.appeal_id = appeals.id AND tasks.type = 'AttorneyTask' AND tasks.appeal_type='Appeal'
+            order by tasks.assigned_at desc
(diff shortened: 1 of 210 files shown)


pkarman:
I put these into appeals-team instead, with a README disclaimer that they are not maintained.

PR #12986, opened 2019-12-16
Looking at flakes

ferristseng:
Changes:

  * `spec/feature/queue/ama_queue_spec.rb`
    
    I think this might be due to an ambiguity because that button shows up twice on the page. I don't know why it doesn't throw an error, but the page clearly has the appropriate button displaying as evidenced by the error message, which is showing the "View task instructions" text.

  * `spec/feature/queue/task_queue_spec.rb`

    Unclear what is going on here, but it looks like the update task type call is creating new tasks instead of doing just an update. You can tell that is happening because of this screenshot from a failed test: https://97199-51449239-gh.circle-artifacts.com/0/home/circleci/project/tmp/capybara/screenshot_2019-12-16-16-15-40.053.png. The tabs show a total of 24 tasks that get created, when the setup to the `before` block should only create 20. 

    I made 2 changes to these tests:
    * Combine the tests into 1. One was testing if the right filter options are available, and the other was testing if the filtering actually works. I think it makes sense to have these part of the same feature test. It should also hopefully make it more resilient to flakes.
    * Use `let!` construction instead of `build` and `update`. I don't understand why the update would be generating new data, but using a `let!` here seems more idiomatic and predictable.


spec/feature/queue/task_queue_spec.rb
```diff
@@ -435,12 +435,16 @@ def validate_pulac_cerullo_tasks_created(task_class, label)
...
+      create_list(:privacy_act_task, unassigned_count, :in_progress, assigned_to: organization)
+    end
+    let!(:assigned_tasks) do
+      create_list(:privacy_act_task, assigned_count, :on_hold, assigned_to: organization)
+    end
 
     before do
       organization.add_user(organization_user)
       User.authenticate!(user: organization_user)
-      create_list(:privacy_act_task, unassigned_count, :in_progress, assigned_to: organization)
```
pkarman: this is a good catch. doing fixtures in `before` is problematic because they *always* run, whereas `let()` allows you to override per context.

spec/feature/queue/task_queue_spec.rb
```diff
@@ -508,24 +512,39 @@ def validate_pulac_cerullo_tasks_created(task_class, label)
...
-          expect(find("tbody").find_all("tr").length).to eq(unassigned_count)
-          page.find_all("path.unselected-filter-icon-inner").first.click
-          page.find("label", text: "#{TranslationTask.label.humanize} (#{translation_task_count})").click
-          expect(find("tbody").find_all("tr").length).to eq(translation_task_count)
+        it "filters are correct, and filter as expected" do
+          step "check if there are the right number of rows for the unassigned table" do
+            expect(find("tbody").find_all("tr").length).to eq(unassigned_count)
+          end
+
+          step "check if the filter options are as expected" do
```
pkarman: I like this use of `step` -- makes it easier to skip later if necessary too.

PR #12993, opened 2019-12-17
Bump Frontend Deps 2019-12

jcq:
Connects #12996

### Description
Bumped various dependencies for minor patches and other updates:
- webpack
- babel
- babel-loader
- style-loader
- sass-loader
- node-sass
- react
- react-dom
- react-redux
- react-virtualized
- moment
- moment-timezone
- sprintf-js
- react-tooltip
- enzyme-adapter-react-16

This also adds `prettier-eslint` to allow for (optional) automatic code formatting in editors based on our eslint rules.

### Acceptance Criteria
- [ ] Code compiles correctly
- [ ] Tests pass
- [ ] No new runtime errors

### Testing
```
cd client
rm -rf node_modules
yarn install
```

Then launch dev environment


client/package.json
```diff
@@ -45,88 +70,63 @@
     "immutability-helper": "^3.0.1",
     "immutable": "^3.8.2",
     "imports-loader": "^0.7.1",
-    "jsdom": "9.8.3",
     "lodash": "^4.17.13",
     "mark.js": "^8.11.0",
-    "moment": "^2.18.1",
-    "moment-timezone": "^0.5.13",
-    "node-sass": "^4.9.2",
+    "moment": "^2.24.0",
```
leikkisa: Do we need moment, can we switch everything to moment-timezone? I think it's confusing that sometimes moment refers to moment, and sometimes it refers to moment-timezone.  It looks like moment-timezone includes moment 2.9.0

jcq: Moment is still a required dependency of moment-timezone (though it could be brought in a sub-dependency if we did want to take it out of main package.json). It's true that we *could* use moment-timezone only, but that woul:
- require updating a bunch of other files that import it
- encourage further overuse of an already very weighty module

While I am planning to add stuff to our Webpack config that strips a lot of unnecessary stuff out of Moment before it ends up in our bundle as an interim step, I would actually prefer to move away from it altogether. My strong preference would be to use the [date-fns](https://github.com/date-fns/date-fns) library, as it's easy to import just the individual functions that are needed rather than an entire large library. This would greatly improve our overall JS library footprint, which is always a good thing.

Anyway, this is a long-winded way of saying that I'd say we leave things as they are for now, but address in multiple ways moving forward. :)

Issue #12709, opened 2019-11-14
Visualize Appeal Task tree


pkarman:
## User story
As a BVA Tableau user, I would like to see a visualization of the Task tree for any Appeal.

## Acceptance criteria
- [x] Please put this work behind the feature toggle: `:appeal_viz`
- [x] This feature should be accessible to the same users who can view the Appeal
- [x] Include screenshot(s) in the Github issue if there are front-end changes

### Designs
![appeal-task-tree-viz](https://user-images.githubusercontent.com/1205061/68877479-c61ae300-06cb-11ea-906f-bb4759bc432b.png)

### Background/context
This was designed originally for direct hosting on the Tableau server, but given the existing infrastructure and db connectivity that Caseflow has, it's more maintainable in the app.

[Relevant Tableau docs on custom views](https://help.tableau.com/current/pro/desktop/en-gb/embed_code_custom_view.htm)

### Technical notes

Should use a new erb "view" layout for this feature, since we don't want the usual Caseflow "skin" on it.

I suggest the "/viz" name space for new controllers.

Existing Perl script that implements this feature. The things for re-use include the CSS and data model logic.

```perl
#!/usr/bin/perl
#
use strict;
use warnings;
use Scalar::Util qw(blessed dualvar isdual readonly refaddr reftype
                        tainted weaken isweak isvstring looks_like_number
                        set_prototype);
use DBI;
​
use Data::Dumper;
#
my $my_appid;
my $driver   = "SQLite"; 
my $database = "/var/www/html/database/caseflow_tasks.db";
my $dsn = "DBI:$driver:dbname=$database";
my $userid = "";
my $password = "";
my $dbh = DBI->connect($dsn, $userid, $password, { RaiseError => 1 }) 
   or die $DBI::errstr;
​
print "Content-type: text/html\n\n";
​
# $my_appid = $ARGV[0];
​
$my_appid = $ENV{'QUERY_STRING'};
​
if( !looks_like_number($my_appid) ) {
   print "<html><title>Error on Input</title><body>\n";
   print "<H1>Appeal ID must be numeric</H1>\n";
   print "</body></html>\n";
   exit(-1);
}
​
my $stmt = qq[
    SELECT id, appeal_id, status, type, assigned_at, parent_id, closed_at, receipt_date,
    Cast (( JulianDay(COALESCE(date(closed_at),date('now'))) - JulianDay(date(receipt_date))) As Integer) as days_aged,
[...]

PR #13008, opened 2019-12-17
Move task tree to its own controller

pkarman:
connects #12709 

### Description

Move Appeal task tree to its own controller. This should make it easier to address from Tableau per task, where the appeal type is more explicit.

Also adjusts the HTML display to include specific attributes/calculations.

### User Facing Changes

 - [x] Screenshots of UI changes added to PR & Original Issue
 
Screen Shot 2019-12-18 at 5 26 29 PM


app/models/appeal.rb
```diff
@@ -44,6 +44,8 @@ class Appeal < DecisionReview
   UUID_REGEX = /^\h{8}-\h{4}-\h{4}-\h{4}-\h{12}$/.freeze
   STATE_CODES_REQUIRING_TRANSLATION_TASK = %w[VI VQ PR PH RP PI].freeze
 
+  alias_attribute :nod_date, :receipt_date # LegacyAppeal parity
```
tomas-nava: Is this a requirement from Tableau?

pkarman: no, just makes it easier to calculate day math in the view.

PR #13020, opened 2019-12-18
ETL::OrganizationsUser sync by unique cols index

pkarman:
Discovered while syncing in production, organizations_users rows are deleted and re-created for the same org/user pair. That will change the primary key id value (since it's a new record). The ETL table will get out of sync if we rely on 1:1 match for `id` columns.

Instead we use the unique index columns to match. The `id` can therefore change for the "same" record but that's ok because this table is a m2m joining table and the `id` is ephemeral.

### DIFF
--- a/app/models/etl/organizations_user.rb
+++ b/app/models/etl/organizations_user.rb
@@ -3,4 +3,13 @@
 # copy of OrganizationsUser model
 
 class ETL::OrganizationsUser < ETL::Record
+  class << self
+    private
+
+    # original records get deleted and re-created as users change orgs.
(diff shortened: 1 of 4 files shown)


app/models/etl/organizations_user.rb
```diff
@@ -3,4 +3,13 @@
 # copy of OrganizationsUser model
 
 class ETL::OrganizationsUser < ETL::Record
+  class << self
+    private
+
+    # original records get deleted and re-created as users change orgs.
+    # so we avoid using the PK autoincrementing "id" to sync and instead rely on the unique index cols.
+    def find_by_primary_key(original)
```
pkarman: the fact that tests continue to pass means we had proper encapsulation of this private method.

Issue #12373, opened 2019-10-11
MTV - Create request issues for vacated decisions


leikkisa:
### User or job story
When a judge grants a motion to vacate for an issue, create a new request issue for that decision on the same appeal.

### Acceptance criteria
- [ ] Request issues are automatically created for granted motion to vacate issues, on the same appeal
- [ ] These request issues contest the decision issue that is being vacated (contested_decision_issue_id)
- [ ] The request issue data should be populated similarly to if a contested_decision_issue were added through intake (description, decision date, rating information, etc).

### Release notes
See https://github.com/department-of-veterans-affairs/caseflow/issues/11669

#### Designs
N/A, backend ticket

### Tech notes
Partial IDs are saved in the post decision motion, full grants apply to all decision issues on the appeal.

PR #13023, opened 2019-12-18
Create request issues from selected MTV decision issues

jcq:
Connects #12373

### Description
This automatically creates request issues for granted motions to vacate based on the selected decision issue(s).

### Acceptance Criteria
- [ ] Code compiles correctly


app/models/post_decision_motion.rb
```diff
@@ -29,6 +29,23 @@ def vacated_issues
     DecisionIssue.find(vacated_decision_issue_ids)
   end
 
+  def create_request_issues
+    vacated_decision_issue_ids.map do |decision_issue_id|
```
leikkisa: You can map the `vacated_issues` directly here since that method is available.

app/models/post_decision_motion.rb
```diff
@@ -29,6 +29,23 @@ def vacated_issues
     DecisionIssue.find(vacated_decision_issue_ids)
```
leikkisa: In this method, you could return the task.appeal's decision issues if it is a full grant, and otherwise find the vacated_decision_issue_ids.

The other option is saving the decision_issue_ids for a full grant too, that could be done in `PostDecisionMotionUpdater.create_motion` - if it's a full grant, save the IDs of the `task.appeal.decision_issues`.  This would require an update to the DB comment.

I could go either way. It might be more intuitive to save all of the IDs since the field is not called partially_vacated_decision_issue_ids (yikes, that's long).  It can be derived, and it takes up more space this way, but I don't think anyone really cares about that.  I suppose if the decision issues somehow change in the future, this will also keep a more reliable snapshot of what they were at the time (although the request issues created would also capture that).

app/models/post_decision_motion.rb
```diff
@@ -29,6 +29,23 @@ def vacated_issues
...
   end
 
+  def create_request_issues
```
leikkisa: There will be a second round of creating request issues for vacate and readjudicate, and vacate and de novo.  So it might be better to be more specific for this method name, like `create_request_issues_for_vacature`. Something to differentiate these from the new request issues that are being created for getting a new decision.

app/models/post_decision_motion.rb
```diff
@@ -29,6 +29,23 @@ def vacated_issues
...
+    vacated_decision_issue_ids.map do |decision_issue_id|
+      prev_decision_issue = DecisionIssue.find(decision_issue_id)
+      RequestIssue.create!(
+        decision_review: prev_decision_issue.decision_review,
+        decision_review_type: prev_decision_issue.decision_review_type,
+        contested_decision_issue_id: prev_decision_issue.id,
+        contested_rating_issue_reference_id: prev_decision_issue.rating_issue_reference_id,
+        contested_rating_issue_profile_date: prev_decision_issue.rating_profile_date,
+        contested_issue_description: prev_decision_issue.description,
```
leikkisa: Let's add the rating_promulgation_date too.

app/models/post_decision_motion.rb
```diff
@@ -29,6 +29,23 @@ def vacated_issues
...
+        nonrating_issue_category: prev_decision_issue.nonrating_issue_category,
+        benefit_type: prev_decision_issue.benefit_type,
+        decision_date: prev_decision_issue.caseflow_decision_date
+      )
```
leikkisa: It looks like we should add veteran_participant_id to the new request issue too.

spec/models/post_decision_motion_spec.rb
```diff
@@ -1,5 +1,55 @@
...
+  let(:hyperlink) { "https://va.gov/fake-link-to-file" }
+  let(:instructions) { "formatted instructions from judge" }
+  let(:vacated_decision_issue_ids) { orig_decision_issues.map(&:id) }
+  let(:post_decision_motion) do
+    create(
+      :post_decision_motion,
+      task: task,
+      disposition: disposition,
+      vacate_type: vacate_type,
+      vacated_decision_issue_ids: vacated_decision_issue_ids
```
leikkisa: Right now this wouldn't happen for a full grant, so you might have to update the tests depending on which approach you decide to take above.  If you put the logic for picking the decision issues for `granted` versus `partially_granted` in `PostDecisionMotion`, it'd be good to include both cases in this test.  If it's in `PostDecisionMotionUpdater`, then this looks good, but that model test will need an update.

app/models/post_decision_motion.rb
jcq: Yup, that makes sense!

app/models/post_decision_motion.rb
jcq: Oops, forgot to mention that I tried adding this only to realize that `rating_promulgation_date` is on decision issues but not on request issues. Is this something we add somewhere else, or just need to make sure to add it to the new decision issues once they get created?

spec/models/post_decision_motion_spec.rb
jcq: added test in post_decision_motion_updater_spec

app/models/post_decision_motion.rb
jcq: I initially implemented this logic on the front end so we send the issue IDs either way, but decided to switch things around and just implemented the logic in the PostDecisionMotionUpdater. That way we save the issue IDs no matter what, and as you say, it can serve as a snapshot of what the decision issues were at that given moment.

app/models/post_decision_motion.rb
leikkisa: Ah, it's saved in the `decision_date` field on request issues. That's not intuitive, but yes, please add it there.

app/workflows/post_decision_motion_updater.rb
```diff
@@ -30,6 +30,9 @@ def create_motion
 
     if params.key?(:vacated_decision_issue_ids)
       motion.vacated_decision_issue_ids = params[:vacated_decision_issue_ids]
+    elsif disposition == "granted"
```
leikkisa: Can we move this to before the above "if" statement?  There's a little bit of a bug, where if I select a couple of decision issues (let's say I select "partially granted" first), but then change it to granted, it's going with the selected decision issues instead of all of them.

Related, can we only save the vacated_decision_issue_ids from the params if it's got a disposition of "partially_granted"?.  The same bug impacts if they switch from partially granted to denied as well.

app/models/post_decision_motion.rb
leikkisa: Don't forget to update the DB comment ;-)

app/models/post_decision_motion.rb
```diff
@@ -29,6 +29,23 @@ def vacated_issues
     DecisionIssue.find(vacated_decision_issue_ids)
   end
 
+  def create_request_issues_for_vacature
+    vacated_issues.map do |prev_decision_issue|
```
leikkisa: 👍

app/workflows/post_decision_motion_updater.rb
leikkisa: To tack on to my previous comment, I guess it's less about the order, and more about making the decision issue IDs saved based on the disposition, rather than the presence of vacated_decision_issue_ids.

app/models/post_decision_motion.rb
jcq: Just to clarify here... this means we should set `decision_date` on the new request issue to `rating_promulgation_date` from the decision issue? I had assumed we would set that field to the `caseflow_decision_date` from the decision issue... do we not need to propagate that one?

app/workflows/post_decision_motion_updater.rb
jcq: Yeah, that logic should be updated. I'm just going to switch it to be based on disposition for both, which should be fine since we're validating the presence of the `vacated_decision_issue_ids` param for partial grant before we get here anyway.

app/models/post_decision_motion.rb
leikkisa: You're right, it should be the decision date, which for appeal issues will always be the `caseflow_decision_date`.

leikkisa:
Regarding the code climate notice about referring to prev_decision_issue so many times, they are implying the method should be in the request issue model. There are also ways to copy an object with a selected list of attributes included or excluded, which may not trigger that since the request issue would only get referred to once.

Functionally it works fine as it and I don't have a strong preference on it. But if you do decide to move it to RequestIssue, you can also get the veteran's participant ID from the previous request issue (so you don't have to pass anything to the RequestIssue model.

jcq:
@leikkisa I did a quick update whereby we move the logic for creating a request issue from a decision issue into the `DecisionIssue` model. Does that seem like a reasonable solution? It feels a bit cleaner, but I'm happy to go with whatever you think best. You had mentioned putting it into `RequestIssue`, but that would have the same issue of referring to the `DecisionIssue` model more than itself, right?

app/models/decision_issue.rb
```diff
@@ -155,6 +155,21 @@ def associated_request_issue
     request_issues.first
   end
 
+  def new_request_issue_from!
```
leikkisa: How about calling it `create_contesting_request_issue!`

jcq: Yeah, I wasn't thrilled with the name I had given it. I'll happily accept this change. :)

PR #13024, opened 2019-12-18
Save copy of legacy issue data to new table

leikkisa:
connects #12397

### Description
This ticket saves a copy of new legacy issues to the legacy issue table, and connects legacy opt ins to that legacy issue.

Once the new table is backfilled (will add scripts below), then the next step is to switch over and remove the vacols_id and vacols_sequence_id on the request issues table.


leikkisa:
Here is a script for backfilling data:
```
def backfill_legacy_issues(ris)
  ris.each do |ri|
    next if ri.legacy_issues.present?
    next unless ri.vacols_id && ri.vacols_sequence_id

    LegacyIssue.create!(
      request_issue: self,
      vacols_id: vacols_id,
      vacols_sequence_id: vacols_sequence_id
    ).tap do |li|
      ri.legacy_issue_optin&.update!(legacy_issue: li)
    end
  end
end
```

After the above, we can see if there are remaining opt-ins that aren't connected to a legacy issue.
`LegacyIssueOptin.where(legacy_issue_id: nil)`

I think these will represent the few "multiples" that were manually added.  I think we'll need to look at the vacols data for those to backfill them because we don't have the vacols_id and vacols_sequence_id saved for them. They are typically the same vacols_id as the other request issue's opt in, so we need to look at the legacy appeal and find the sequence_id of the other opted in issue(s).

tomas-nava:
I would recommend writing the backfill as a rake script so you can test, dry run, and log changes. Here's a draft of what it might look like based on your example above and [this script for changing the task type](https://github.com/department-of-veterans-affairs/caseflow/blob/master/lib/tasks/tasks.rake#L22):

```ruby
namespace :legacy_issues do
  class NoIssuesToBackfill < StandardError; end

  # usage:
  # Backfill request issues matching passed ids (dry run)
  #   $ bundle exec rake legacy_issues:backfill[12,13,14,15,16,17]"
  # Backfill request issues matching passed ids (execute)
  #   $ bundle exec rake legacy_issues:backfill[false,12,13,14,15,16,17]"
  # Backfill all request issues (dry run)
  #   $ bundle exec rake legacy_issues:backfill[]"
  # Backfill all request issues (execute)
  #   $ bundle exec rake legacy_issues:backfill[false]"
  desc "backfill eligible legacy issues"
  task :backfill, [:dry_run] => :environment do |_, args|
    Rails.logger.tagged("rake legacy_issues:backfill") do
      Rails.logger.info("Invoked with: #{args.to_a.join(', ')}")
    end
    extras = args.extras
    dry_run = args.dry_run&.to_s&.strip&.upcase != "FALSE"
    if dry_run && args.dry_run.to_i > 0
      extras.unshift(args.dry_run)
    end

    if dry_run
      puts "*** DRY RUN"
      puts "*** pass 'false' as the first argument to execute"
    end

    candidate_issues = if extras.any?
                         RequestIssue.find(extras)
                       else
                         RequestIssue.all
                       end

    eligible_issues = candidate_issues.select do |ri|
      ri.legacy_issues.empty? || (ri.vacols_id.present? && ri.vacols_sequence_id.present?)
    end

    if eligible_issues.count == 0
      fail NoIssuesToBackfill, "There aren't any eligible request issues to backfill."
    end

    id_string = eligible_issues.map(&:id).sort.join(",")
    change = dry_run ? "Would backfill" : "Backfilling"
    message = "#{change} #{eligible_issues.count} of #{candidate_issues.count} " \
      "RequestIssues with ids #{id_string}"
    puts message

    if !dry_run
      Rails.logger.tagged("rake legacy_issues:backfill") { Rails.logger.info(message) }

      eligible_issues.each do |ri|
        LegacyIssue.create!(
          request_issue: ri,
          vacols_id: ri.vacols_id,
          vacols_sequence_id: ri.vacols_sequence_id
        ).tap do |li|
          ri.legacy_issue_optin&.update!(legacy_issue: li)
        end
      end
    end
  end
end
```

PR #13027, opened 2019-12-18
Turn on config.force_ssl for links when FORCE_ENV env var set

pkarman:
Now that our demo site has a SSL cert, let's require it.

### DIFF
--- a/docker-bin/env.sh
+++ b/docker-bin/env.sh
@@ -2,6 +2,7 @@
 
 # env vars required to run the appeals-app docker container on caseflowdemo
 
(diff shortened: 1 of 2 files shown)


docker-bin/env.sh
```diff
@@ -2,6 +2,7 @@
 
 # env vars required to run the appeals-app docker container on caseflowdemo
 
+export FORCE_SSL=true
```
enriquemanuel: thanks!

pkarman:
#13028 is enough (we think, for now).

PR #13028, opened 2019-12-18
Have LetsEncrypt for demo using NGINX

enriquemanuel:
move some files around and have cert and configurations done accordingly.
This is now being manually tested in demo


local/nginx/config/nginx/site-confs/default
```diff
@@ -0,0 +1,70 @@
...
+        location / {
+                try_files $uri @rails;
+        }
+
+        location @rails {
+                proxy_set_header  X-Real-IP  $remote_addr;
+                proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
+                proxy_set_header Host $http_host;
+                proxy_redirect off;
+                proxy_pass http://rails_app;
```
pkarman: if we proxy `http://` to the rails app, then we definitely don't want #13027 to force ssl. Otherwise the Rails app itself will be expecting https connection.

pkarman: I'm wrong. I think. TIAS. https://stackoverflow.com/a/26738410

PR #13031, opened 2019-12-18
Allow PreRoutingColocatedTasks to be reassigned

### LINKED ISSUE #12856: Bug: Validation Fails for PreRoutingFoiaColocatedTask
## Description
We see an alert indicating that validation failed during task reassignment with this error: 
```Validation failed: Colocated subtype is not included in the list```

- [Sentry Alert](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/6953/)
- [Code that performs validation](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/tasks/colocated_task.rb#L181)
- [PR that *might* be tangentially related](https://github.com/department-of-veterans-affairs/caseflow/pull/11904) (This handles a different circumstance, but implies that `PreRoutingFoiaColocatedTask` should often be handled in the same way that `PreRoutingColocatedTask` does.)

## Technical notes
Using `ColocatedTask.descendants` instead of `ColocatedTask.subclasses` would allow `PreRoutingFoiaColocatedTask` to pass that validation, but might have unintended consequences.

### PR DESCRIPTION
**Pull request by hschallhorn**
Resolves #12856

### Description
Allows the reassigning of  `PreRoutingColocatedTask`s, but still throws errors when creating from params or changing task type.

### Acceptance Criteria
- [ ] Cannot `create_from_params` a `PreRoutingColocatedTask`
- [ ] Cannot `change_task_type` to a `PreRoutingColocatedTask`
- [ ] Can `reassign` an existing `PreRoutingColocatedTask`

### CODE REVIEW

File: spec/models/tasks/colocated_task_spec.rb [hunk 1]
```diff
@@ -530,4 +540,29 @@
...
+    let(:task_class) { PreRoutingFoiaColocatedTask }
+    let(:parent_task) do
+      PreRoutingFoiaColocatedTask.create(
+        assigned_by: attorney,
+        assigned_to: colocated_org,
+        parent: appeal_1.root_task,
+        appeal: appeal_1
+      )
+    end
+    let!(:child_task) { parent_task.children.first }
```
jcq: I'm not sure I know where/how the child task is being created, but this seems to work, so... 🤷‍♀️

hschallhorn: Organizations that use a round robin distributer[ automatically create a child task](https://github.com/department-of-veterans-affairs/caseflow/blob/5410ee8d75d925727d955e2b85601f216be5eda6/app/models/task.rb#L31) and [assign it to a team member](https://github.com/department-of-veterans-affairs/caseflow/blob/5410ee8d75d925727d955e2b85601f216be5eda6/app/models/task.rb#L571). The Colocated org is [one of these](https://github.com/department-of-veterans-affairs/caseflow/blob/9495744b4dc2975523c24b9a4ae2b2bcea4212f0/app/models/organizations/colocated.rb#L8) orgs!

Issue #13021, opened 2019-12-18
ETL Users VACOLS columns


pkarman:
> For VA users for vacols, need their stitle and smemgrp fields to be able to resolve attorneys into SSC groups and DVCs.

per @vatboyda

pkarman:
```
irb(main):013:0> ETL::User.where.not(stitle: nil).count
[2019-12-20 15:35:11 -0500]    (3.8ms)  SELECT COUNT(*) FROM "users" WHERE ("users"."stitle" IS NOT NULL)
=> 1711
```

PR #13039, opened 2019-12-19
Add index to document_views.user_id

pkarman:
We have a unique index on `document_id, user_id` but not one just on `user_id`.

`SELECT COUNT(*) FROM "document_views" WHERE "document_views"."user_id" = ?` will timeout after 30 seconds.

This new index may (or may not) help that performance but it's a start.

### Database Changes
*Only for Schema Changes*

* [ ] Column comments updated

### DIFF
--- a/db/migrate/20191219155741_add_user_id_index_document_views.rb
+++ b/db/migrate/20191219155741_add_user_id_index_document_views.rb
@@ -1,0 +1,13 @@
+class AddUserIdIndexDocumentViews < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def change
+    ActiveRecord::Base.connection.execute "SET statement_timeout = 1800000" # 30 minutes
+
+    add_index :document_views, :user_id, algorithm: :concurrently
+
+  ensure
+    # always restore the timeout value
+    ActiveRecord::Base.connection.execute "SET statement_timeout = 30000" # 30 seconds
+  end
+end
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -198,7 +198,9 @@
     t.string "sattyid"
     t.string "sdomainid", null: false
     t.string "slogid", null: false
+    t.string "smemgrp", limit: 8
```
pkarman: need to merge #13021 first

db/migrate/20191219155741_add_user_id_index_document_views.rb
```diff
@@ -0,0 +1,13 @@
...
+  disable_ddl_transaction!
+
+  def change
+    ActiveRecord::Base.connection.execute "SET statement_timeout = 1800000" # 30 minutes
+
+    add_index :document_views, :user_id, algorithm: :concurrently
+
+  ensure
+    # always restore the timeout value
+    ActiveRecord::Base.connection.execute "SET statement_timeout = 30000" # 30 seconds
```
nanotone: Since it doesn't seem like these timeout increases are avoidable going forward, is there any way to put the two `SET statement_timeout` lines in a method/block or other decorator pattern, so that there's one standardized way to wrap heavy indexing work, and the important lines pop out more?

Or is code reuse not so available in migration scripts?

pkarman: good question. I'll open a new ticket.

PR #13041, opened 2019-12-19
modify VirtualHearigModal to retrieve poa and vet email

### LINKED ISSUE #12891: Pre-populate rep & vet email on virtual hearing modal if exists in system
Design: https://va-gov.invisionapp.com/share/8AUO95JMRW7#/screens/391731033

Currently, the veteran and representative emails are not pre-populated on the VirtualHearingModal. 
The case details page is populating representative and veteran data [here](https://github.com/department-of-veterans-affairs/caseflow/blob/064d4c876d5ce18cc50f848ec029590db397a49d/client/app/queue/CaseDetailsView.jsx#L118). We can use this [method](https://github.com/department-of-veterans-affairs/caseflow/blob/dd83ad9453d86f86fc4993136cd325ea5f54fee6/client/app/queue/QueueActions.js#L153) to retrieve the relevant data.

rubaiyat22:
Resolves #12891 

### Description
- change hearing and legacy hearing serializers to return `veteran_email_address` & `representative_email_address`
- add methods to Appeal and LegacyAppeal to return both email addresses
- cache `email_address` to `Person` to Postgres and Etl
- write feature test for ama hearing

### Database Changes
*Only for Schema Changes*

Add `email_address` field to `Person`
- [ ] Timestamps (created_at, updated_at) for new tables
- [x] Column comments updated
- [ ] Query profiling performed (eyeball Rails log, check bullet and fasterer output)
- [ ] Appropriate indexes added (especially for foreign keys, polymorphic columns, and unique constraints)
- [x] #appeals-schema notified with summary and link to this PR

Screen Shot 2019-12-20 at 11 37 48 AM


client/app/hearings/components/details/DetailsInputs.jsx
```diff
@@ -79,7 +79,7 @@ class DetailsInputs extends React.Component {
               onChange={(veteranEmail) => updateVirtualHearing({ veteranEmail })}
             />
             <TextField
-              name="POA/Representive Email"
+              name="POA/Representative Email"
```
rubaiyat22: fixing my spelling mistake here 😓

app/serializers/hearings/hearing_serializer.rb
```diff
@@ -56,6 +56,7 @@ class HearingSerializer
   attribute :regional_office_timezone
   attribute :representative, if: for_full
   attribute :representative_name
+  attribute :representative_email_address
```
lomaxap: couple of notes about adding these to the serializer:
1. while we cache POA data [for 24 hours](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/services/bgs_power_of_attorney.rb#L29), for every hearing note seen in the last day, this will add a BGS call on the daily docket even though this information will not be used. I'm slightly concerned about impact on speed here.
2. veteran bgs attributes are cached every day [for in this job](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/jobs/warm_bgs_caches_job.rb) for upcoming hearings, so as long as we added the email address to the cached attributes, we won't have any issues.

@ferristseng do you think we should add a line in the `warm_bgs_cache_job` that retrieves the representative email address in this PR? Or figure out another solution later?

rubaiyat22: I'm also concered about spped. My initial idea was to do what HearingWorksheet is doing in legacyHearingSerializer, ` attribute :veteran_gender, if: for_worksheet, &:fetch_veteran_gender`. I'm not exactly sure how this works but maybe we could something similar for details so we avoid calling `representative_email_address` in daily docket

ferristseng: > do you think we should add a line in the warm_bgs_cache_job that retrieves the representative email address in this PR? Or figure out another solution later?

@lomaxap Yeah, I think that's a good idea.

Looking at the code, would it already do this? It looks like `warm_ro_participant_caches` calls `open_hearing_days_with_hearings_hash`, which calls into the hearing serializer. Do we get this for free already? (I haven't looked too deeply into how the caching job works)

ferristseng: Oh, maybe we need to add `email_address` as a cached attribute on the `Person` model (kind of like the veteran model).

ferristseng: > I'm also concerned about speed. My initial idea was to do what HearingWorksheet is doing in legacyHearingSerializer, attribute :veteran_gender, if: for_worksheet, &:fetch_veteran_gender. I'm not exactly sure how this works but maybe we could something similar for details so we avoid calling representative_email_address in daily docket

I think that might work too. `DetailsContainer` sets up by calling the hearings route, which serializes all the fields for the worksheet. `DailyDocketContainer` calls the hearings day route, which uses `hearing.quick_to_hash`. I think there is a naming issue right now with `for_worksheet`, since that serializer sets up the front ends for both the worksheet and the case details page it seems like?

pkarman:
I am wondering if we need to cache it on both Person and Veteran. Maybe just Person?

One of the things I have been observing is that Veteran table seems to be a subset of Person. It didn't start out that way but now we appear to be caching PII data in both places.

spec/feature/hearings/virtual_hearings/hearing_details_spec.rb
```diff
@@ -74,33 +85,27 @@
...
-    click_dropdown(name: "hearingType", index: 1)
-    fill_in "vet-email", with: "veteran@testingEmail.com"
-    fill_in "rep-email", with: "rep@testingEmail.com"
-    click_button(COPY::VIRTUAL_HEARING_CHANGE_HEARING_BUTTON)
-
-    visit "hearings/" + hearing.external_id.to_s + "/details"
-
-    expect(page).to have_field("Veteran Email", with: "veteran@testingEmail.com")
-    expect(page).to have_field("POA/Representive Email", with: "rep@testingEmail.com")
+  context "User can see and edit veteran and poa emails" do
```
rubaiyat22: modify flaky test

spec/jobs/warm_bgs_caches_job_spec.rb
```diff
@@ -56,7 +56,7 @@
       expect(Rails.cache.exist?(address_cache_key)).to eq(true)
       expect(appeal.veteran.reload[:ssn]).to_not be_nil
       expect(@slack_msg).to be_nil
-      expect(@people_sync).to eq(5)
```
ferristseng: Why did this test result change?

rubaiyat22: It used to result to a 5 because we were caching these five attributes `[:first_name, :last_name, :middle_name, :name_suffix, :date_of_birth]`. It changed because we are now caching one more attribute `email_address`. That was my understanding

ferristseng: I think `people_sync` is a counter for the number of times `update_cached_attributes!` is called, which doesn't necessarily correspond to the number of attributes there are. You can change `let!(:people) { create_list(:person, 5) }` to create 7 users instead of 5, and it breaks the test.

rubaiyat22: It seems that the line `let(:bgs_poa) { BgsPowerOfAttorney.new }` is creating another `Person` object. I'm not sure yet why this result changed now

ferristseng: Oh, ok. I think that makes sense. Maybe that wasn't being touched by the caching job before, and now it is because the serializer is getting the POA email. Could you add a comment that says that you're expecting that to create a `Person`?

PR #13042, opened 2019-12-19
Add more models that refer to Users

pkarman:
Whenever we add columns that refer to `User` model by foreign key associations, we should update this service class. It is useful for [merging user records](https://github.com/department-of-veterans-affairs/caseflow/wiki/Merging-User-Accounts).


pkarman:
* link updated
* FK introspection on the models can work, sometimes, but in prod models are not always autoloaded so we found explicit declaration necessary.

pkarman:
I will add this to the schema danger file checklist.

PR #13061, opened 2019-12-27
Fix race condition for remand supplemental claims

leikkisa:
connects #9948

### Description
Previously, if the last two request issues on an end product establishment had their decision issue sync processed at the same time, the remand supplemental claims wouldn't get created due to a race condition.  We have been manually checking for these and establishing them.

This PR moves the step for kicking off `on_decision_issues_sync_processed` outside of the transaction where decision issues are created, which involves calls to both BGS and VBMS which would slow it down.  It only runs once all request issues are processed, instead of when all other issues were processed except the current issue (previous behavior).

It is less likely that `on_decision_issues_sync_processed` would experience a race condition now that it's outside of the longer transaction.  But if it does experience a race condition now, it will get called twice instead of not getting called.  The first time it's called, if all of the end product establishment;s request issues haven't been created yet, then the second time it's called then it will call `create_contentions` and `associate_rating_issues` for the new issue, but otherwise is idempotent.


app/models/decision_review.rb
```diff
@@ -230,7 +230,7 @@ def create_remand_supplemental_claims!
       delay = rsc.receipt_date.future? ? (rsc.receipt_date + PROCESS_DELAY_VBMS_OFFSET_HOURS.hours).utc : 0
       rsc.submit_for_processing!(delay: delay)
 
-      unless rsc.processed? || rsc.receipt_date.future?
```
leikkisa: I removed this because if this method called twice, after the first time it will register as "processed?".  At that point, and end product will be established, and contentions will be created on it.  But it might get a new request issue the second time.  Processing it again does not establish a new end product, but it will add a contention for the new request issue.

app/models/claim_review.rb
```diff
@@ -276,8 +276,8 @@ def incomplete_tasks?
   end
 
   def create_decision_review_task!
-    return if tasks.any? { |task| task.is_a?(DecisionReviewTask) } # TODO: more specific check?
-    return if request_issues.active.blank?
+    return if tasks.any? { |task| task.is_a?(DecisionReviewTask) }
+    return if request_issues.active.empty?
```
leikkisa: The request issues is an array even if there aren't any, so I thought `empty?` sounded more intuitive here.

app/models/end_product_establishment.rb
```diff
@@ -296,10 +296,10 @@ def sync_decision_issues!
     end
   end
 
-  def on_decision_issue_sync_processed(processing_request_issue)
```
leikkisa: The "processing_request_issue" isn't necessary anymore.  It used to be used to check "Is every other issue processed except myself?".  Now, we're just checking if every issue is processed.

app/models/request_issue.rb
```diff
@@ -393,11 +393,12 @@ def sync_decision_issues!
     transaction do
       return unless create_decision_issues
 
-      end_product_establishment.on_decision_issue_sync_processed(self)
       clear_error!
       close_decided_issue!
       processed!
     end
+
+    end_product_establishment.on_decision_issue_sync_processed
```
leikkisa: The transaction above could take a while, because `create_decision_issues` makes requests to VBMS and BGS.  Moving this step outside of that transaction is the main change to prevent the race condition.

I think it's okay to mark the decision_sync as processed, before remand supplemental claims are created, which is really something that happens at the end product establishment level anyway.

spec/models/appeal_spec.rb
```diff
@@ -77,69 +77,6 @@
     end
   end
 
-  context "#create_remand_supplemental_claims!" do
```
leikkisa: Moving this test because the method is in decision_review.rb

spec/models/decision_review_spec.rb
```diff
@@ -83,6 +85,98 @@
...
+      )
+      expect(caseflow_remand.request_issues.count).to eq(1)
+      expect(caseflow_remand.request_issues.first).to have_attributes(
+        contested_decision_issue: remanded_decision_issue_processed_in_caseflow
+      )
+      expect(caseflow_remand.end_product_establishments).to be_empty
+      expect(caseflow_remand.tasks.first).to have_attributes(assigned_to: BusinessLine.find_by(url: "nca"))
+    end
+
+    context "when it gets runs a second time with a new decision issue" do
```
leikkisa: This test is new, the above test was previously in `appeal_spec`.

app/models/end_product_establishment.rb
nanotone: This is so, so much better.

app/models/request_issue.rb
nanotone: Describing my understanding of the bug, as a sanity check:

So the race condition was happening when two DecisionIssueSyncJobs would kick off overlapping transactions here, and neither of them would think they were the "last" issue to be processed since the other one hadn't committed yet.

leikkisa: That's correct.

Issue #12694, opened 2019-11-13
Allow unidentified issues to be connected to legacy issues


leikkisa:
## User or job story
As an intake user, when I add an unidentified request issue, I should be able to connect legacy issues to it for opt-in.

## Acceptance criteria
- [ ] Please put this work behind the feature toggle: verify_unidentified_issue
- [ ] When adding an unidentified issue, trigger the legacy opt-in modal after the user adds an unidentified issue
- [ ] This should happen in the same order it is happening elsewhere, after correction type and before untimely exemption (as usual)
- [ ] The unidentified issue should get the same legacy issue eligibility checks (as usual)
- [ ] The legacy issue(s) should be saved as connected to the unidentified issue (as usual)
- [ ] Opt-ins should be created for the unidentified issue if is not ineligible (as usual)
- [ ] Include screenshot in the GH issue of the add issues page with an unidentified issue with a legacy issue opt-in connected to it

## Release notes

Unidentified issues can be connected to legacy issue opt-ins.  This is being added to support the use of unidentified issues to handle missing ratings.

### Out of scope
Being able to process multiple legacy opt-ins.  That is being handled in other tickets, with the final one being #12693.  This work can happen in parallel, however it would be good to pay attention to the current status of that work for merge conflicts or other impacts.

### Background/context
Historically we tried to force users to remove unidentified issues, and did not allow them to continue to be decided on.  If we do allow them to continue to be processed and decided, then we need to allow them to also opt-in legacy issues if appropriate.

### Technical notes

On the front-end, the AddIssueManager, AddedIssue, and LegacyOptInModal may be good files to start to look at.

I am marking some AC with `(as usual)` to indicate that code may exist that does this for unidentified issues already, or for other types of issues where you can see the pattern

Sjones352:
Connected to legcay issue
<img width="1143" alt="Screen Shot 2019-12-31 at 12 42 37 AM" src="https://user-images.githubusercontent.com/23080951/71611415-95511900-2b66-11ea-90ec-5defea41218a.png">

<img width="1013" alt="Screen Shot 2019-12-31 at 12 38 59 PM" src="https://user-images.githubusercontent.com/23080951/71630012-bc433580-2bce-11ea-8ff3-b20ff53582d5.png">

PR #13070, opened 2019-12-30
Allow unidentified issues to be connected to legacy issues

Sjones352:
Connects #12694
Screen Shot 2019-12-31 at 12 42 37 AM

Screen Shot 2019-12-31 at 12 38 59 PM

Screen Shot 2019-12-31 at 12 45 28 PM


client/app/intake/components/AddIssueManager.jsx
```diff
@@ -231,6 +231,10 @@ class AddIssueManager extends React.Component {
           if (isCorrection(true, this.props.intakeData)) {
             this.setState({ currentIssue,
               currentModal: 'CorrectionTypeModal' });
+          } else if (featureToggles.verifyUnidentifiedIssue && this.hasLegacyAppeals()) {
```
leikkisa: Can you also change the button text when there are legacy issues?
This is similar to line 127'ish, but will also have the feature toggle:
`submitText: this.hasLegacyAppeals() ? 'Next' : 'Add this issue',`

client/app/intake/util/issues.js
```diff
@@ -177,6 +177,7 @@ const formatUnidentifiedIssues = (state) => {
   return (state.addedIssues || []).
     filter((issue) => issue.isUnidentified).
     map((issue) => {
+      console.log("vacols", issue.vacolsId);
```
leikkisa: This can be removed

client/app/intake/util/issues.js
```diff
@@ -187,7 +188,9 @@ const formatUnidentifiedIssues = (state) => {
         correction_type: issue.correctionType,
         untimely_exemption: issue.untimelyExemption,
         untimely_exemption_notes: issue.untimelyExemptionNotes,
-        ineligibleReason: issue.ineligible_reason
+        ineligibleReason: issue.ineligible_reason,
```
leikkisa: I think this should be snake_cased, it looks like it is for Rating issues, which I expect would be the same here.

spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -1362,6 +1362,63 @@ def complete_higher_level_review
...
+            ineligible_checklist = find("ul.cf-issue-checklist")
+            expect(ineligible_checklist).to have_content(
+              ineligible_constants.legacy_issue_not_withdrawn.to_s
+            )
+
+            expect(RequestIssue.find_by(
+                     is_unidentified: true,
+                     ineligible_reason: :legacy_issue_not_withdrawn,
+                     vacols_id: "vacols1",
+                     vacols_sequence_id: "1"
```
leikkisa: Nice!

PR #13080, opened 2019-12-31
Only established appeals in search results

### LINKED ISSUE #13079: Include only fully established Appeals in search results
If an Appeal is not yet established (in the middle of Intake, so it exists in the db but Intake is not finished) it will still appear in search results. This causes an error determining AOD.

https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/7534/events/584978/

pkarman:
Resolves #13079 

### Description
Include only established Appeals in search results (based on `established_at` column).

Also refactors AppealFinder tests into their own spec file, and some internal refactoring of methods not specific to the Veteran model.


app/services/appeal_finder.rb
```diff
@@ -1,36 +1,14 @@
 # frozen_string_literal: true
 
 class AppealFinder
-  def initialize(user:)
```
pkarman: just reorganized the file structure to match our pattern: class methods first.

app/services/appeal_finder.rb
```diff
@@ -61,17 +39,35 @@ def find_appeal_by_docket_number(docket_number)
...
+      self.class.find_appeals_with_file_numbers(
+        [veterans.map(&:file_number), veterans.map(&:ssn)].flatten.compact.uniq
+      )
+    end
+  end
+
   private
 
-  attr_accessor :user
+  attr_reader :user
```
pkarman: no need to write it

spec/services/appeals_for_poa_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      described_class.new(
+        veteran_file_number: veteran.file_number,
+        poa_participant_ids: [vso_participant_id, "other vso participant id"]
+      )
+    end
+
+    it "returns only the case with vso assigned to it" do
+      returned_appeals = subject.call
+      expect(returned_appeals.count).to eq 2
+      expect(returned_appeals.first).to eq appeals.first
```
ferristseng: Curious if this is guaranteed to preserve the ordering? Would it be better to just check for inclusion?

spec/services/appeal_finder_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+describe AppealFinder, :all_dbs do
+  let(:veteran) { legacy_appeal.veteran }
+  let!(:appeal) { create(:appeal, veteran: veteran) }
+  let(:unknown_docket_number) { "012345-000" }
+  let(:invalid_veteran_id) { "obviouslyinvalidveteranid" }
+  let(:invalid_docket_number) { "invaliddocket-number" }
+  let!(:legacy_appeal) { create(:legacy_appeal, :with_veteran, vacols_case: create(:case)) }
+
+  describe ".find_appeal_by_docket_number" do
+    subject { described_class.find_appeal_by_docket_number(docket_number) }
```
ferristseng: Cool! I didn't know about `described_class`!

app/models/veteran.rb
```diff
@@ -184,31 +184,6 @@ def decision_issues
     DecisionIssue.where(participant_id: participant_id)
   end
 
-  def accessible_appeals_for_poa(poa_participant_ids)
```
pkarman: moved these methods to a service class.

spec/feature/queue/search_spec.rb
```diff
@@ -42,20 +42,6 @@
       end
     end
 
-    context "when using AppealFinder" do
```
pkarman: these are unit tests, not feature tests (no browser involved) so moved them to a unit test spec file.

spec/services/appeals_for_poa_spec.rb
pkarman: yes good catch. I failed to see that when copy/pasting.

PR #13082, opened 2020-01-02
Caseflow::Migration class with add_safe_index method

pkarman:
## Description

Custom migration class to make it easier to create indexes. Handles the extension of statement_timeout for big tables, and automatically reverts indices on exceptions (since we do not run in transactions).

Example:

```ruby
class TestIndexRollback < Caseflow::Migration
  def change
    add_safe_index :users, :foobar # will raise exception, no such column
  end
end
```


lib/caseflow/migration.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+# Migration with built-in timeout extensions for adding indexes
+
+class Caseflow::Migration < ActiveRecord::Migration[5.1]
```
nanotone: What do you think about updating the Dangerfile to advocate using `Caseflow::Migration` for index-adding migrations going forward (assuming it's straightforward enough to implement that in the Dangerfile)?

lib/caseflow/migration.rb
```diff
@@ -0,0 +1,30 @@
...
+    table, columns, options = *args
+    options ||= {}
+    options[:algorithm] ||= :concurrently
+
+    add_index(table, columns, options)
+
+  rescue => error
+    say "Caught #{error}, rolling back index"
+    options[:column] = columns unless options[:name]
+    remove_index(table, options)
```
nanotone: It seems to me that the error should be re-raised so that whoever's running the migration can see that it failed. Also, the name `add_index_with_timeout` doesn't suggest any extra error-wrapping.

pkarman: good point @nanotone 

Does the method name `add_safe_index` better communicate?

nanotone: That method name sounds good to me. But, I'm still unclear on what happens if an error is caught and the index is rolled back (during a deployment, say). Will the migration script behave as if it succeeded?

pkarman: By re-raising the error (as you suggested) it should force the migration to halt (which is what we want). The auto-rollback of the index is mimicing what would happen if we were running inside a transaction (which we explicitly are not doing, to avoid locking).

You can try it locally with the example migration in the PR summary.

```
bundle exec rake db:migrate
[2020-01-02 16:17:18 -0600]    (0.5ms)  SELECT pg_try_advisory_lock(2554694829110551050)
[2020-01-02 16:17:18 -0600]    (1.8ms)  SELECT "schema_migrations"."version" FROM "schema_migrations" ORDER BY "schema_migrations"."version" ASC
[2020-01-02 16:17:18 -0600] Migrating to TestIndexRollback (20200101)
-- disable_ddl_transaction is true
== 20200101 TestIndexRollback: migrating ======================================
-- Extending statement_timeout to 30 minutes
[2020-01-02 16:17:18 -0600]    (0.6ms)  SET statement_timeout = 1800000
-- add_index(:users, :foobar, {:algorithm=>:concurrently})
[2020-01-02 16:17:18 -0600]    (1.4ms)  CREATE  INDEX CONCURRENTLY "index_users_on_foobar" ON "users"  ("foobar")
-- Caught PG::UndefinedColumn: ERROR:  column "foobar" does not exist
: CREATE  INDEX CONCURRENTLY "index_users_on_foobar" ON "users"  ("foobar"), rolling back index
-- remove_index(:users, {:algorithm=>:concurrently, :column=>:foobar})
-- Restoring statement_timeout to 30 seconds
[2020-01-02 16:17:18 -0600]    (0.6ms)  SET statement_timeout = 30000
-- disable_ddl_transaction is true
[2020-01-02 16:17:18 -0600]    (0.6ms)  SELECT pg_advisory_unlock(2554694829110551050)
rake aborted!
StandardError: An error has occurred, all later migrations canceled:

No indexes found on users with the options provided.
/Users/pkarman/va-appeals/caseflow/lib/caseflow/migration.rb:24:in `rescue in add_safe_index'
/Users/pkarman/va-appeals/caseflow/lib/caseflow/migration.rb:13:in `add_safe_index'
/Users/pkarman/va-appeals/caseflow/db/migrate/20200101_test_index_rollback.rb:3:in `change'
/Users/pkarman/.rbenv/versions/2.5.3/bin/bundle:23:in `load'
/Users/pkarman/.rbenv/versions/2.5.3/bin/bundle:23:in `<main>'

Caused by:
ArgumentError: No indexes found on users with the options provided.
/Users/pkarman/va-appeals/caseflow/lib/caseflow/migration.rb:24:in `rescue in add_safe_index'
/Users/pkarman/va-appeals/caseflow/lib/caseflow/migration.rb:13:in `add_safe_index'
/Users/pkarman/va-appeals/caseflow/db/migrate/20200101_test_index_rollback.rb:3:in `change'
/Users/pkarman/.rbenv/versions/2.5.3/bin/bundle:23:in `load'
/Users/pkarman/.rbenv/versions/2.5.3/bin/bundle:23:in `<main>'

Caused by:
ActiveRecord::StatementInvalid: PG::UndefinedColumn: ERROR:  column "foobar" does not exist
: CREATE  INDEX CONCURRENTLY "index_users_on_foobar" ON "users"  ("foobar")
/Users/pkarman/va-appeals/caseflow/lib/caseflow/migration.rb:20:in `add_safe_index'
/Users/pkarman/va-appeals/caseflow/db/migrate/20200101_test_index_rollback.rb:3:in `change'
/Users/pkarman/.rbenv/versions/2.5.3/bin/bundle:23:in `load'
/Users/pkarman/.rbenv/versions/2.5.3/bin/bundle:23:in `<main>'

Caused by:
PG::UndefinedColumn: ERROR:  column "foobar" does not exist
/Users/pkarman/va-appeals/caseflow/lib/caseflow/migration.rb:20:in `add_safe_index'
/Users/pkarman/va-appeals/caseflow/db/migrate/20200101_test_index_rollback.rb:3:in `change'
/Users/pkarman/.rbenv/versions/2.5.3/bin/bundle:23:in `load'
/Users/pkarman/.rbenv/versions/2.5.3/bin/bundle:23:in `<main>'
Tasks: TOP => db:migrate
(See full trace by running task with --trace)
make: *** [db-migrate] Error 1
```

nanotone: Thanks, super informative! I think I was just looking at the wrong diff and didn't realize the re-raise got added.

pkarman:
@nanotone thoughts on this now that it's baked awhile?

PR #13084, opened 2020-01-02
VBMS efolder pagination

### LINKED ISSUE #13047: Fix VBMS::DocumentTooBig errors
https://github.com/department-of-veterans-affairs/connect_vbms/pull/241 implements a new PagedDocuments VBMS service.

### AC
- [ ] Upgrade connect_vbms gem
- [ ] Use the new service in our `VBMSService` class instead of `VBMS::Requests::FindDocumentVersionReference`

### Testing

Test in prod with PID 3978641 and any others.

### Related

#10665 
#10901 
https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/1160

pkarman:
Resolves #13047 

### Description
Updates VBMS and BGS gems.

Adds feature toggle `:vbms_pagination` to control whether we use the new VBMS paged documents service to allow fetching VBMS efolders where there are 5k+ documents.

### Testing Plan
1. Identify a Veteran with 5k+ efolder in UAT (https://dsva.slack.com/archives/CK466V65V/p1577992111002300?thread_ts=1577990211.002000&cid=CK466V65V)
2. In Rails console, toggle the `:vbms_pagination` feature for yourself.
3. In Rails console, call the `ExternalApi::VbmsDocumentsForAppeal` for the Veteran.
4. In Rails console, call the `ExternalApi::VBMSService.fetch_document_series_for` method for the Veteran.

Repeat in production after deployment.

### DIFF
--- a/.reek.yml
+++ b/.reek.yml
@@ -16,11 +16,13 @@ detectors:
       - Veteran
   ControlParameter:
     exclude:
+      - ExternalApi::VBMSService#self.send_and_log_request
       - Fakes::BGSService
       - HearingRepository#slot_new_hearing
       - LegacyDocket#distribute_appeals
       - Task#actions_allowable?
       - TaskSorter#sort_requires_case_norm?
+      - VBMSCaseflowLogger#log
       - Veteran
   UncommunicativeVariableName:
     exclude:
@@ -80,6 +82,7 @@ detectors:
   LongParameterList:
     exclude:
       - Address#initialize
+      - ExternalApi::VBMSService#self.create_contentions!
       - LegacyDocket#new_distributed_case
   ManualDispatch:
     exclude:
@@ -130,6 +133,7 @@ detectors:
       - OrganizationTrackingTasksTab#column_names
       - Fakes::BGSService
       - Fakes::PexipService
+      - VBMSCaseflowLogger#log
       - VirtualHearings::CalendarTemplateHelper#formatted_date_time_for_zone
 
 ### Directory specific configuration
--- a/spec/services/vbms_documents_for_appeal_spec.rb
+++ b/spec/services/vbms_documents_for_appeal_spec.rb
@@ -145,13 +145,24 @@ describe ExternalApi::VbmsDocumentsForAppeal do
       it "looks up the BGS claim number in VBMS" do
         bgs_claim_number = "87654321"
 
-        allow(vbms_client).to receive(:send_request).and_raise(nonexistent_file_number_error)
+        # mock service to fail for both numbers
+        allow(pagination_service).to receive(:call)
+          .with(file_number: veteran_file_number).and_raise(nonexistent_file_number_error)
+        allow(pagination_service).to receive(:call)
+          .with(file_number: bgs_claim_number).and_raise(nonexistent_file_number_error)
+
+        # mock bgs service to return claim number
         allow(bgs_client).to receive(:fetch_veteran_info).with(veteran_file_number)
           .and_return(claim_number: bgs_claim_number)
+
+        # inject our mocked pagination service
         allow(docs).to receive(:vbms_paged_documents_service) { pagination_service }
 
-        expect(vbms_client).to receive(:send_request).exactly(:twice)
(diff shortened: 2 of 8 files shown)


.reek.yml
```diff
@@ -16,11 +16,13 @@ detectors:
       - Veteran
   ControlParameter:
     exclude:
+      - ExternalApi::VBMSService#self.send_and_log_request
```
pkarman: existing violations

spec/services/vbms_documents_for_appeal_spec.rb
```diff
@@ -91,4 +91,59 @@
...
+      it "looks up the BGS claim number in VBMS" do
+        bgs_claim_number = "87654321"
+
+        allow(vbms_client).to receive(:send_request).and_raise(nonexistent_file_number_error)
+        allow(bgs_client).to receive(:fetch_veteran_info).with(veteran_file_number)
+          .and_return(claim_number: bgs_claim_number)
+        allow(docs).to receive(:vbms_paged_documents_service) { pagination_service }
+
+        expect(vbms_client).to receive(:send_request).exactly(:twice)
+        expect { docs.fetch }.to raise_error(VBMS::FilenumberDoesNotExist)
```
leikkisa: Is this error getting raised because of line 139? It looks in the code like it raises this error if the claim number isn't different, but if it is different, it attempts to use `bgs_claim_number_docs_request`

spec/services/vbms_documents_for_appeal_spec.rb
```diff
@@ -91,4 +91,59 @@
...
+
+        result_hash = {
+          manifest_vbms_fetched_at: nil,
+          manifest_vva_fetched_at: nil,
+          documents: docs_from_vbms_array
+        }
+
+        expect(bgs_client).to_not receive(:fetch_veteran_info)
+        expect(vbms_client).to_not receive(:send_request)
+        expect(docs.fetch).to eq result_hash
```
leikkisa: I'm having some trouble understanding this test.  If the pagination service is turned on, why is not returning the pagination response.  It looks like mocking `DocumentsFromVbmsDocuments` is overriding the result, independent of whether the pagination service or the other service is being called.

pkarman: The internal pagination response, mocked, is `documents: [1, 2, 3]`.

The internal `docs_from_vbms` object, mocked, converted that to `[4, 5, 6]`.

What we are testing is that we call the pagination service (mocked) internally, rather than the older non-pagination service. The mocks allow us to avoid making actual HTTP calls, and instead define known return values. The return values themselves are meaningless.

spec/services/vbms_documents_for_appeal_spec.rb
pkarman: yes, line 139 mocks the exception.

The trigger for trying a 2nd time with the bgs number is the exception. If the 2 numbers are the same or bgs number is nil, the trigger is ignored and a second exception raised.

spec/services/vbms_documents_for_appeal_spec.rb
pkarman: I have commented and renamed things to hopefully be more clear. I basically just started with the existing test above and then changed it to use the pagination service.

spec/services/vbms_documents_for_appeal_spec.rb
pkarman: I realize in copy/pasting the existing test I failed to update to account for the different service call. I will fix.

pkarman: updated, with comments.

pkarman:
thanks @leikkisa -- updated tests with comments to hopefully make them more clear. Mocks and injections are hard to follow, IME.

spec/services/vbms_documents_for_appeal_spec.rb
leikkisa: Okay. This is still still confusing to me, but I see that it's still testing that the new pagination service is being called.

Issue #13081, opened 2019-12-31
Missing Ratings | Handle response for decisions at issue service


leikkisa:
## Description
This is to handle the response from the new BGS service for fetching decisions at issue, rating_profile.find_in_date_range.

## Acceptance criteria
- [ ] Add rating_profile.find_in_date_range to Caseflow
- [ ] Handle the response via a feature toggle [decisions_at_issue_service]
- [ ] Implement in a way that the prior services being used are replaced when using the new service (either one or another is used)

PR #13089, opened 2020-01-04
Missing Ratings | Implement Decisions at issue BGS Service

leikkisa:
connects #13081

### Description
This PR mimics the services currently used to fetch rating profiles (one for a list of the ratings within the date range, and another to fetch the profiles for each of them), but using the new BGS service.

The new service includes the rating profiles in the response, instead of making two calls. There are also a number of slight differences in organization or field names (e.g. prfl_dt vs prfil_dt).

This work is behind the feature toggle: ratings_at_issue

Any feedback or questions are welcome.

### Acceptance Criteria
- [x] Code compiles correctly

### Testing Plan
The testing is interesting here... like where to separate PromulgatedRating and RatingAtIssue (vs. putting stuff into the Rating spec) in tests.
 
### Documentation Updates


leikkisa:
To get example data in prod:
`BGSService.new.client.rating_profile.find_in_date_range(participant_id: v.participant_id, start_date: Date.new(2000,1,1), end_date: Date.today)`

Veteran participant ID 7101147 has no ratings to see an example of a veteran with no ratings. Most veterans can be used for an example with ratings.  I also tested with making up a veteran that doesn't exist (like participant ID 55555), which provides the same response for no ratings.

app/models/rating.rb
```diff
@@ -1,39 +1,77 @@
...
+    issues_data = Array.wrap(rating_profile[:rating_issues] || rating_profile.dig(:rba_issue_list, :rba_issue))
+
+    issues_data.map do |issue_data|
+      issue_data[:dgnstc_tc] = diagnostic_codes.dig(issue_data[:dis_sn], :dgnstc_tc)
+      RatingIssue.from_bgs_hash(self, issue_data)
+    end
+  end
+
+  def decisions
+    return [] unless FeatureToggle.enabled?(:contestable_rating_decisions, user: RequestStore[:current_user])
```
pkarman, line 66: do we still need this feature toggle? I think `contestable_rating_decisions` has been around for awhile now.

leikkisa: I want to keep it until we finish discussions with AMO about missing ratings. It was originally intended to be a temporary solution, and AMO is saying that the same issue appearing under multiple decision dates has been confusing to users. But we haven't fully decided how to handle it  yet, there's a lot of different elements of the missing ratings work that is under discussion.

app/models/concerns/latest_rating_disability_evaluation.rb
```diff
@@ -5,8 +5,8 @@ module LatestRatingDisabilityEvaluation
 
   class_methods do
     def latest_disability_evaluation(disability)
-      latest = Array.wrap(disability[:disability_evaluations]).max_by do |dis_eval|
-        dis_eval[:conv_begin_dt] || dis_eval[:begin_dt] || dis_eval[:dis_dt]
+      latest = Array.wrap(disability[:disability_evaluations] || disability[:disability_evaluation]).max_by do |eval|
+        eval[:conv_begin_dt] || eval[:begin_dt] || eval[:dis_dt]
```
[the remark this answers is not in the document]
leikkisa: Really good feedback!

Issue #12991, opened 2019-12-16
Dispatch Stats Page | Deprecate stats pages


leikkisa:
## Description
The stats pages aren't being visited, and their code is old.  In addition, they are a source of accessibility audit violations.

## Acceptance criteria
- [ ] Remove stats pages
- [x] User sees a 404 message if they try to navigate to them

## Background/context/resources
GH issue for deprecating intake stats page:
https://github.com/department-of-veterans-affairs/caseflow/issues/11933

Sandra double checked in Google Analytics whether these pages were getting visits, and there were 0 visits since October 1st, 2019.

msj0nes:
@Sjones352 is this ticket entirely complete? Can you close it if so.

Sjones352:
@msj0nes I removed the Intake stat code from caseflow and disabled the Dispatch/ Certification code to return 404. This was done in phases as the it was a lot of files to remove. I will remove the rest of this code this sprint.

PR #13095, opened 2020-01-07
Deprecate Stat pages

Sjones352:
Connects #12991
I do not think we should remove stats code at the moment until we are sure that users do not complain about missing these pages.
Screen Shot 2020-01-06 at 7 26 47 PM


spec/feature/certification/certification_stats_spec.rb
```diff
@@ -155,7 +155,7 @@
     expect(page).to have_content("Overall (median)")
   end
 
-  scenario "Unauthorized user access" do
+  fscenario "Unauthorized user access" do
```
leikkisa: This focus can be removed.

spec/feature/dispatch/dispatch_stats_spec.rb
```diff
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 
-RSpec.feature "Dispatch Stats Dashboard", :postgres do
+RSpec.feature "Dispatch Stats Dashboard", :postgres, skip: "deprecated" do
   before do
     Timecop.freeze(Time.utc(2015, 1, 1, 17, 55, 0, rand(1000)))
   end
 
-  context ".show#daily" do
+  fcontext ".show#daily" do
```
leikkisa: This focus can be removed

app/controllers/intake_stats_controller.rb
```diff
@@ -8,41 +8,11 @@ class IntakeStatsController < ApplicationController
 
   def show
     # deprecated 2019/08/28
-    # either remove this controller entirely or update it to reflect current AMA models.
```
leikkisa: You should be able to remove this whole file.

spec/services/cache_manager_spec.rb
```diff
@@ -8,32 +8,17 @@
     allow(Rails).to receive(:cache).and_return(redis_store)
   end
 
-  let(:stats_cache_key) { "IntakeStats-last-calculated-timestamp" }
   let(:ro_schedule_period) { build_stubbed(:ro_schedule_period) }
   let(:ro_cache_key) { "RoSchedulePeriod-#{ro_schedule_period.id}-cached-submitting_to_vacols" }
 
-  describe "#all_cache_keys" do
-    it "returns all keys in the cache" do
-      IntakeStats.throttled_calculate_all!
```
leikkisa: So you want to keep this test, but remove this line (17), and remove `stats_cache_key` from line 20

Issue #12634, opened 2019-11-07
Rails 6 upgrade


pkarman:
With the release of Rails 6 we are now a few revisions behind. We should plot out what's involved in an upgrade.

PR #13097, opened 2020-01-07
Upgrade Rails to 5.2.4.1

### LINKED ISSUE #9943: Upgrade to Rails 5.2
Fixes security issue CVE-2019-5420 among other things.

ferristseng:
Resolves #9943

### Description

  * Upgrade Rails to 5.2.4.1 and run `bundle update rails`

Used this guide for upgrade instructions (doesn't seem to require any changes): https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-1-to-rails-5-2

This is an interim step for upgrading Rails to 6 #12634

### References

  * https://github.com/rails/rails/blob/v5.2.4.1/activerecord/CHANGELOG.md


pkarman:
Nice!

Similar started in https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/1166

spec/factories/supplemental_claim.rb
```diff
@@ -17,7 +17,12 @@
 
     after(:create) do |sc, evaluator|
       if evaluator.number_of_claimants
-        sc.claimants = create_list(:claimant, evaluator.number_of_claimants, decision_review: sc)
+        sc.claimants = create_list(
```
ferristseng: This is very strange, but one of the changes seems to have "fixed" the validation, so now this line is breaking for multiple tests because the `payee_code` isn't getting set.

app/repositories/virtual_hearing_repository.rb
```diff
@@ -48,13 +48,13 @@ def hearing_day_has_virtual_hearing?(hearing_day)
...
+        .where(<<-SQL)
           (
-            virtual_hearings.judge_email_sent = ?
-            OR virtual_hearings.veteran_email_sent = ?
+            NOT virtual_hearings.judge_email_sent
+            OR NOT virtual_hearings.veteran_email_sent
             OR (
               NOT virtual_hearings.representative_email = null
-              AND virtual_hearings.representative_email = ?
+              AND NOT virtual_hearings.representative_email_sent
```
ferristseng: I spent some time tracking down why this suddenly started to fail. I couldn't find a direct change in the change log that is responsible for this, but I did notice there was a change to how booleans get serialized for sqlite3, and thought maybe it could have had some other implications.

Here is the relevant log of changes I was browsing: https://edgeguides.rubyonrails.org/5_2_release_notes.html#active-record-notable-changes

spec/factories/claimant.rb
```diff
@@ -4,7 +4,7 @@
   factory :claimant do
     sequence(:decision_review_id)
     sequence(:participant_id)
-    decision_review_type { "Appeal" }
+    association :decision_review, factory: :appeal
```
ferristseng: This change in `master` makes the validation "work". It seemed like there was a really deep exception that was being swallowed because an appeal didn't exist for some claimants, which was necessary for validation.

spec/controllers/case_reviews_controller_spec.rb
```diff
@@ -38,15 +38,22 @@
...
+                  },
+                  {
+                    "description": "great moments",
+                    "disposition": "remanded",
+                    "benefit_type": "vha",
+                    "diagnostic_code": "5001",
+                    "request_issue_ids": [request_issue2.id],
+                    "remand_reasons": [{ "code": "va_records", "post_aoj": true }]
+                  }
+                ]
```
ferristseng: I am **completely** baffled by why this change is necessary. For some reason, when the payload gets posted to the server, the disposition shifts to the first issue:

```
{
...
 "issues"=>
  [{"disposition"=>"remanded",   # WHY is this here???
    "description"=>"wonderful life",
    "benefit_type"=>"pension",
    "diagnostic_code"=>"5001",
    "request_issue_ids"=>["1", "2"]},
   {"description"=>"great moments",
    "benefit_type"=>"vha",
    "diagnostic_code"=>"5001",
    "request_issue_ids"=>["3"],
    "remand_reasons"=>[{"code"=>"va_records", "post_aoj"=>"true"}]}],
,,,}
```

I confirmed the params are being passed right to the controller, but the controller is reordering the parameters. Going to research if this is a known issue, but it is very strange indeed

ferristseng: Ok, I figured this out. Seems to be a bug when unwrapping parameters that were URL-encoded. The fix was to update this to serialize as JSON, which should be what the controller expects anyways.

ferristseng:
Deployed this branch successfully to UAT, and doing some smoke tests there.

ferristseng:
@pkarman Could you take a look at this again when you have time? I tested this in UAT a few days ago, and didn't notice any regressions just clicking through the app. I think we can try deploying this to production.

config/initializers/content_security_policy.rb
```diff
@@ -0,0 +1,25 @@
+# Be sure to restart your server when you modify this file.
```
pkarman: should we add a comment at the top to indicate why this file is present but completely commented out?

config/initializers/new_framework_defaults_5_2.rb
```diff
@@ -0,0 +1,38 @@
+# Be sure to restart your server when you modify this file.
```
pkarman: do we need this file?

spec/controllers/case_reviews_controller_spec.rb
```diff
@@ -38,20 +38,27 @@
...
-                "issues": [{ "description": "wonderful life",
-                             "benefit_type": "pension",
-                             "diagnostic_code": "5001",
-                             "request_issue_ids": [request_issue1.id, request_issue3.id] },
-                           { "disposition": "remanded", "description": "great moments",
-                             "benefit_type": "vha",
-                             "diagnostic_code": "5001",
-                             "request_issue_ids": [request_issue2.id],
-                             "remand_reasons": [{ "code": "va_records", "post_aoj": true }] }]
+                "issues": [
```
pkarman: this seems to be whitespace only?

spec/models/claim_review_spec.rb
```diff
@@ -1024,9 +1024,10 @@ def epe
   end
 
   describe "#search_table_ui_hash" do
-    let!(:claimants) { [create(:claimant), create(:claimant)] }
     let!(:appeal) { create(:appeal) }
-    let!(:sc) { create(:supplemental_claim, veteran_file_number: appeal.veteran_file_number, claimants: claimants) }
+    let!(:sc) do
+      create(:supplemental_claim, veteran_file_number: appeal.veteran_file_number, number_of_claimants: 2)
```
pkarman: +1

spec/controllers/case_reviews_controller_spec.rb
ferristseng: yeah, I think this is whitespace only

spec/models/claim_review_spec.rb
ferristseng: this is a slight change to use `number_of_claimants` to setup the test data vs. `claimants`. I don't think there's any functional change

config/initializers/new_framework_defaults_5_2.rb
ferristseng: I included these incase we needed them in the future. I'm fine with deleting them

pkarman: yes let's delete

PR #13100, opened 2020-01-07
Batch assign ama tasks in one request

### LINKED ISSUE #13086: Investigate why multiple JudgeDecisionReviewTasks are being created
Followup to #12802. We've now got [4 sentry alerts](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/7612/) of duplicate tasks we blocked the creation of.

## AC
- [x] Determine what is causing caseflow to create duplicate `JudgeDecisionReviewTask`s

hschallhorn:
~Random thought that may solve inconsistent request store tasks on the front end that may have given users the ability to create multiple assign requests~ Resolves #13086

### Description
The judge assign tasks controller was always able to handle multiple task assignments, but for some reason we only would send a request for each task. This PR batch assigns ama tasks to only receive and set the new tasks in the redux store once.

### Acceptance Criteria
- [ ] All tests still pass
- [ ] One request for all AMA task assignments is sent to the back end

### Testing Plan
1. Sign in as BVAAAAAABSHIRE and go to http://localhost:3000/queue/3/assign
1. Select multiple legacy appeals and ama appeals and assign them to an attorney
1. Confirm
   1. Button says "Assign 0 cases"
   1. We still send 1 request per each legacy appeal
   1. We send 1 request for _all_ the selected ama appeals
   1. Case numbers are still correctly updated in the attorney headers
   1. Assigned cases load correctly in the attorney assigned table (http://localhost:3000/queue/3/assign/<attorney_id>)
   1. Throw in a refresh for good measure to confirm numbers and tasks are all in their correct places

### DIFF
--- a/client/app/queue/QueueActions.js
+++ b/client/app/queue/QueueActions.js
@@ -402,7 +402,7 @@ export const bulkAssignTasks =
 // isInitial is only used for das deprecation
 const dispatchOldTasks = (dispatch, oldTasks, resp, isInitial = false) => {
   // Ama request is batched
-  if (Array.isArray(oldTasks)) {
+  if (Array.isArray(oldTasks) || oldTasks.appealType === 'Appeal') {
     dispatch(onReceiveAmaTasks(resp.tasks.data));
   } else {
     // For das deprecation, legacy_task_controller#create returns tasks, not a task
@@ -455,17 +455,19 @@ export const initialAssignTasksToUser = ({
     };
   });
 
-  return Promise.all(legacyParams.concat(amaParams).map((params) => {
+  const paramsArray = amaParams.requestParams.data.tasks.length ? legacyParams.concat(amaParams) : legacyParams;
+
+  return Promise.all(paramsArray.map((params) => {
     const { requestParams, url } = params;
 
     return ApiUtil.post(url, requestParams).
       then((resp) => resp.body).
       then((resp) => {
-        const assignedTasks = [requestParams.data.tasks].flat();
+        const assignedTasks = requestParams.data.tasks;
 
-        dispatchOldTasks(dispatch, requestParams.data.tasks, resp, true);
+        dispatchOldTasks(dispatch, assignedTasks, resp, true);
 
-        assignedTasks.forEach((oldTask) => {
+        [assignedTasks].flat().forEach((oldTask) => {
           dispatch(setSelectionOfTaskOfUser({
             userId: previousAssigneeId,
             taskId: oldTask.uniqueId,
--- a/client/app/queue/QueueActions.js
+++ b/client/app/queue/QueueActions.js
@@ -455,11 +455,7 @@ export const initialAssignTasksToUser = ({
     };
   });
 
-  const paramsArray = legacyParams.concat(amaParams);
-
-  debugger;
-
-  return Promise.all(paramsArray.map((params) => {
+  return Promise.all(legacyParams.concat(amaParams).map((params) => {
     const { requestParams, url } = params;
 
     return ApiUtil.post(url, requestParams).
--- a/client/app/queue/QueueActions.js
+++ b/client/app/queue/QueueActions.js
@@ -406,7 +406,7 @@ const dispatchOldTasks = (dispatch, oldTasks, resp) => {
     dispatch(onReceiveAmaTasks(resp.tasks.data));
   } else {
     // For das deprecation, legacy_task_controller#create returns tasks, not a task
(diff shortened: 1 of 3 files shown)


client/app/queue/QueueActions.js
```diff
@@ -418,49 +421,66 @@ const dispatchOldTasks = (dispatch, oldTask, resp, isInitial=false) => {
 
 export const initialAssignTasksToUser = ({
   tasks, assigneeId, previousAssigneeId
```
hschallhorn: Probably best to view diff without whitespace changes.

hschallhorn:
Considerations:
1. Does this impact the time between "Assign X cases" button press and finishing of the loading screen?

yoomlam:
I was able to confirm these:
>     Button says "Assign 0 cases"
>     We still send 1 request per each legacy appeal
>     We send 1 request for all the selected ama appeals
>     Case numbers are still correctly updated in the attorney headers
>     Assigned cases load correctly in the attorney assigned table (http://localhost:3000/queue/3/assign/<attorney_id>)
>     Throw in a refresh for good measure to confirm numbers and tasks are all in their correct places

However, when assigning multiple legacy appeals, I get an error in a banner. The appeals appear to be correctly assigned to selected attorney, however the case count for that attorney did not update (and is not correct until I refresh the page).

yoomlam:
Typo: "deatails" in error message.

image

client/app/queue/QueueActions.js
```diff
@@ -399,16 +399,16 @@ export const bulkAssignTasks =
...
   });
 
-// isInitial is only used for das deprecation 
-const dispatchOldTasks = (dispatch, oldTask, resp, isInitial=false) => {
-  if (oldTask.appealType === 'Appeal') {
-    dispatch(onReceiveAmaTasks(resp.tasks.data)); 
+// isInitial is only used for das deprecation
+const dispatchOldTasks = (dispatch, oldTasks, resp) => {
+  // Ama request is batched
+  if (Array.isArray(oldTasks) || oldTasks.appealType === 'Appeal') {
```
yoomlam: I don't see `oldTasks.appealType` being set. So, we determine AMA vs Legacy processing based on if `oldTasks` is an array.

client/app/queue/QueueActions.js
```diff
@@ -399,16 +399,16 @@ export const bulkAssignTasks =
...
+    dispatch(onReceiveAmaTasks(resp.tasks.data));
   } else {
-    //For das deprecation, legacy_task_controller#create returns tasks, not a task
-    const tasks = isInitial && (oldTask.appealType === 'LegacyAppeal' && !oldTask.isLegacy) ? resp.tasks.data : [resp.task.data];
-
+    // For das deprecation, legacy_task_controller#create returns tasks, not a task
+    const tasks = [resp.task.data].flat();
```
yoomlam: `flat()` would be needed if `resp.task.data` is an array. Would `resp.task.data` ever be an array if `oldTasks` is not an array?

client/app/queue/QueueActions.js
```diff
@@ -418,49 +418,62 @@ const dispatchOldTasks = (dispatch, oldTask, resp, isInitial=false) => {
...
+    }
+  }));
+
+  const paramsArray = amaParams.requestParams.data.tasks.length ? legacyParams.concat(amaParams) : legacyParams;
+
+  return Promise.all(paramsArray.map((params) => {
+    const { requestParams, url, taskIds } = params;
+
+    return ApiUtil.post(url, requestParams).
+      then((resp) => resp.body).
```
yoomlam: What is the purpose of `(resp) => resp.body`? Possible reason: if there is no `body`, then subsequent `then`s in the chain do not run

client/app/queue/QueueActions.js
hschallhorn: `reassignTasksToUser` also calls `dispatchOldTasks` with a single task for AMA. This is when `oldTasks.appealType` is checked

client/app/queue/QueueActions.js
hschallhorn: Yep. `legacy_task_controller` returns an array of tasks when das deprecation is turned on even though `oldTasks` is a single task

client/app/queue/QueueActions.js
hschallhorn: I actually don't know. I was just trying to make as few changes to the code we had as possible to get this fix out

PR #13101, opened 2020-01-07
IntakeReporter

pkarman:
### Description
Adds new class IntakeReporter and a small refactor for existing *Reporter classes to DRY them up.

### Documentation Updates

### DIFF
--- a/app/services/asyncable_jobs_reporter.rb
+++ b/app/services/asyncable_jobs_reporter.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require "csv"
+# Summarize models (jobs) that consume the Asyncable concern.
 
 class AsyncableJobsReporter
   attr_reader :jobs
--- a/app/services/concerns/reporter.rb
+++ b/app/services/concerns/reporter.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+# Mixin methods for *Reporter classes. Statistics, dates, etc.
+
+require "csv"
+
+module Reporter
+  extend ActiveSupport::Concern
+
+  def seconds_to_hms(secs)
+    [secs / 3600, secs / 60 % 60, secs % 60].map { |segment| segment.to_s.rjust(2, "0") }.join(":")
+  end
+
+  def median(times)
+    return 0 if times.empty?
+
+    len = times.length
+    (times[(len - 1) / 2] + times[len / 2]) / 2.0
+  end
+
+  def average(times)
+    return 0 if times.empty?
+
+    times.sum.to_f / times.length
+  end
+end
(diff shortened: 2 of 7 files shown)


app/services/intake_reporter.rb
```diff
@@ -0,0 +1,53 @@
...
+    @start_date = start_date
+    @end_date = end_date
+    @type = type
+  end
+
+  # rubocop:disable Metrics/AbcSize
+  def report
+    establishment_durations = []
+    intakes.each do |intake|
+      establishment_durations << (intake.completed_at - intake.started_at).to_i
```
leikkisa: The calculation for establishment time would be `intake.detail.establishment_processed_at - intake.detail.establishment_submitted_at`.  The metric here (completed_at - started_at) is more of how long it took a CA to complete the intake from starting to submitting. If that's the desired metric, I think it makes sense to give it a different label.

app/services/asyncable_jobs_reporter.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require "csv"
```
leikkisa: Should this still be in there since Reporter isn't included?

pkarman: yep csv lib is loaded by Rails by default, so it was superfluous to begin with.

pkarman:
Thanks @leikkisa this is ready for a review now.

app/services/concerns/reporter.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+# Mixin methods for *Reporter classes. Statistics, dates, etc.
+
+require "csv"
```
nanotone: Probably don't need this, right?

app/services/concerns/reporter.rb
```diff
@@ -0,0 +1,26 @@
...
+    len = times.length
+    (times[(len - 1) / 2] + times[len / 2]) / 2.0
+  end
+
+  def average(times)
+    return 0 if times.empty?
+
+    times.sum.to_f / times.length
+  end
+end
```
nanotone: This only nets a tiny bit of DRYness, but what about a little `percent` method, which I think ClaimReviewAsyncStatsReporter, RequestIssueReporter, and probably some future reporters, could use?

app/services/concerns/reporter.rb
```diff
@@ -23,4 +21,8 @@ def average(times)
 
     times.sum.to_f / times.length
   end
+
+  def percent(nominator, denominator)
+    ((nominator / denominator.to_f) * 100).round(2)
+  end
```
nanotone: Ooh nice ✨

PR #13105, opened 2020-01-08
V3 API - ensure request issues are being populated correctly

edmangimelli:
NOTE: this was not branched off of master

To keep inline with how Caseflow records RequestIssue, I, as an Appeals API developer, need to make sure that, when given the ID of a rating or decision, etc., I retrieve all of the details of that rating or decision, and included those details when saving the RequestIssue to the DB

### Acceptance Criteria
- [x] Tests pass


edmangimelli:
@erluti 
cleaned up that test (simplified it a bit)
better?

PR #13107, opened 2020-01-08
Decouple UpcomingHearingDays from AssignHearingTabs

ferristseng:
Prerequisite to #11757 

### Description

  * Moves `UpcomingHearingsTable` to a separate component, which doesn't use `AssignHearingTable`
  * Fix bug in `PowerOfAttorneyDetail` where the component thought it was constantly being loaded
  * Pass the current row to the `filterValueTransform`. This acts as context, and can optionally be handled

I'm making this change because it is not necessary to paginate `UpcomingHearingsTable`. The data that powers the `UpcomingHearingsTable` is bounded by the number of hearings for a given day (usually fewer than 20). `UpcomingHearingsTable` leverages `AssignHearingsTable`, and the plan right now is to add pagination to just `AssignHearingsTable`. It doesn't make sense to make `AssignHearingsTable` flexible enough to handle both cases...and I think it's probably just better to separate them at this point.

### DIFF
--- a/client/app/queue/QueueTable.jsx
+++ b/client/app/queue/QueueTable.jsx
@@ -29,8 +29,9 @@ import QUEUE_CONFIG from '../../constants/QUEUE_CONFIG.json';
  *     an argument and returns the value of the cell for that column.
  *   - @valueName {string} if valueFunction is not defined, cell value will use
  *     valueName to pull that attribute from the rowObject.
- *   - @filterValueTransform {function(any)} function that takes the value of the
- *     column, and transforms it into a string for filtering.
+ *   - @filterValueTransform {function(any, any)} function that takes the value of the
+ *     column, and transforms it into a string for filtering. The row is passed in as
+ *     a second argument.
  *   - @filterOptions {array[object]} array of value - displayText pairs to override the
  *     generated filter values and counts in <TableFilter>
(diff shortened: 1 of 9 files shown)


client/app/queue/QueueTable.jsx
```diff
@@ -29,8 +29,9 @@ import QUEUE_CONFIG from '../../constants/QUEUE_CONFIG.json';
  *     an argument and returns the value of the cell for that column.
  *   - @valueName {string} if valueFunction is not defined, cell value will use
  *     valueName to pull that attribute from the rowObject.
- *   - @filterValueTransform {function(any)} function that takes the value of the
- *     column, and transforms it into a string for filtering.
+ *   - @filterValueTransform {function(any, any)} function that takes the value of the
```
aplybeah: The second argument in `@filterVaulueTransform` may cause merge conflicts with #13055. [A more specific example here](https://github.com/department-of-veterans-affairs/caseflow/pull/13055/commits/2594f1bbea945e5db667169c1bbab24b2960c7a2#diff-5e2a12727bf8a39e8ae77ec332be2ee1R179). I _don't think_ Javascript has a way to make some arguments optional (I'm thinking along the lines of `*args` in python)

PR #13108, opened 2020-01-08
AMO metrics job

### LINKED ISSUE #12510: AMO bi-weekly intake numbers
### User story

As an AMO member, I need bi-weekly numbers reflecting Intake stats.

### Acceptance criteria
- A new cron job uses the ClaimReviewAsyncStatsReporter to send CSV output to the #caseflow-vbms-intake channel, similar to how MonthlyMetricsReportJob works.
- Job should also report current number of stuck SC and HLR async jobs.

pkarman:
Resolves #12510 

### Description
Adds AMO stats cron API entry.

Adds number of stuck jobs to report.

Background: https://dsva.slack.com/archives/CFBLH2LR4/p1578066110054800


app/jobs/amo_metrics_report_job.rb
```diff
@@ -46,7 +46,11 @@ def build_report(async_stats)
     report = []
     report << "AMO metrics report #{start_date} to #{end_date.to_date}"
     report << "Supplemental Claims #{sc_stats[:total]} established, median #{sc_med} average #{sc_avg}"
+    report << "Supplemental Claims newly stuck: #{sc_stats[:expired]}"
+    report << "Supplemental Claims total stuck: #{SupplementalClaim.expired_without_processing.with_error.count}"
```
nanotone: If I'm reading this correctly, "total stuck" is equivalent to "newly stuck" but with an extra `with_error` scope. Does that mean the "total stuck" count will always be less than or equal to the "newly stuck" count?

pkarman: total has no date restriction.
newly is limited to the current start/end date range.

`SupplementalClaim.expired_without_processing.with_error.count` has no date conditions on it.

Issue #12380, opened 2019-10-14
Add checkbox for verifying unidentified issue


leikkisa:
### User or job story
As a VSR, when I am editing a decision review's unidentified issues, I can click a checkbox to confirm that I verified that the issue is a valid past decision.

### Acceptance criteria
- [ ] This work should be behind a feature toggle: `verify_unidentified_issue`
- [ ] On the edit screen only, the unidentified issues modal shows a checkbox that the user can check to indicate that they've verified the issue has a valid past decision
- [ ] Above the checkbox are instructions to the user (see copy below)
- [ ] When the checkbox is checked, the submit button is disabled until all fields are populated
- [ ] When the checkbox is not checked, only the description is required (current behavior)
- [ ] This version of the modal is not available during regular intake, only on edit
- [ ] Include a screenshot of the new unidentified issues modal

### Copy
**Instructions above checkbox**
Please look for a record of the prior decision matching the description and decision date of the issue that was submitted by the veteran.

If you were unable to find a record of the prior decision, please leave the checkbox unchecked and fill in the description and decision date submitted by the veteran.

If found, use the prior decision's information to enter the description and decision date.  Update the notes with information on where you verified the record, such as VBMS or a decision document.  Please include identifying information for the record, such as a document title, that would allow another user to find it.

**Checkbox label**
Verify record of prior decision

### Out of scope
- Backend processing of the issue
- Eligibility check for issue

### Release notes
Users are able to add verified unidentified issues when editing a decision review, so that the issues can continue to be processed.

#### Designs
TBD

Sjones352:
Current Screenshot
<img width="511" alt="Screen Shot 2020-01-09 at 1 14 47 AM" src="https://user-images.githubusercontent.com/23080951/72080811-673ca900-32cb-11ea-8c30-598e70b3535d.png">

PR #13109, opened 2020-01-09
Add checkbox for verifying unidentified issue

Sjones352:
Connects #12380
Screen Shot 2020-01-14 at 12 42 27 PM
Screen Shot 2020-01-14 at 12 43 26 PM

Screen Shot 2020-01-14 at 12 36 24 PM


client/app/intake/components/UnidentifiedIssuesModal.jsx
```diff
@@ -112,9 +121,35 @@ class UnidentifiedIssuesModal extends React.Component {
...
+    Update the notes with information on where you verified the record, such as VBMS or a decision document.
+    Please include identifying information for the record, such as a document title,
+    that would allow another user to find it.
+        </p>
+        <Checkbox
+          label={<strong>Verify record of prior decision</strong>}
+          name="verify_prior_record"
+          value={this.isCheckboxChecked()}
+          onChange={this.onCheckboxChange}
+          // disabled={this.state.saveInProgress}
```
leikkisa: I'm pretty sure you can remove this since it's not getting "saved" yet.

client/app/intake/components/UnidentifiedIssuesModal.jsx
```diff
@@ -112,9 +121,35 @@ class UnidentifiedIssuesModal extends React.Component {
     );
   }
 
+  getCheckbox() {
+    return (
+      <React.Fragment>
+        <p>Please look for a record of the prior decision matching the description
+    and decision date of the issue that was submitted by the veteran.</p>
+        <br />
+        <p>If you were unable to find a record of the prior decision,
```
leikkisa: I wanted to include this chunk of text "If you were unable to find a record of the prior decision...", but since the text is looking pretty long, I think it's okay to remove this paragraph.

Sjones352: Should I also remove the third paragraph because it is a follow-up of the second paragraph.

Sjones352: Should I also remove the third paragraph as it is a follow-up of the second

leikkisa: Please keep that paragraph. Maybe we can shorten it:
If found, please check below that it is verified. Use the prior decision's information to enter the description and decision date.  Update the notes with information on the record, such as the location, ID, or document title.

leikkisa: There's a high likelihood that AMO will request changes to the copy once we present it to them, but this will get the idea and design across.

spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -1418,6 +1427,46 @@ def complete_higher_level_review
 
             expect(page).to_not have_content(COPY::VACOLS_OPTIN_ISSUE_CLOSED)
           end
+
+          fscenario "Verify checkbox on unidentified issues modal on edit page is enabled" do
```
leikkisa: There's a focus on this that should be removed. Thanks for adding this test!

client/app/intake/components/UnidentifiedIssuesModal.jsx
```diff
@@ -64,6 +65,21 @@ class UnidentifiedIssuesModal extends React.Component {
     }
   };
 
+  onCheckboxChange = (event) => {
+    this.setState({ checkboxSelected: event });
+  };
+
+  saveDisable =() => {
```
leikkisa: Just a nit, but I think I would prefer this to be called `saveDisabled`

client/app/intake/components/UnidentifiedIssuesModal.jsx
```diff
@@ -13,7 +14,8 @@ class UnidentifiedIssuesModal extends React.Component {
     this.state = {
       description: '',
       notes: '',
-      disabled: true
+      disabled: true,
```
leikkisa: I don't think we need this to be a prop in state anymore since we're using a function to calculate it. You can use the function in getModalButtons too.

PR #13111, opened 2020-01-09
Do not show inactive users as options for task assignee 

### LINKED ISSUE #12949: Do not show inactive users as options for task assignee
When a user sees a selection of other users to assign a task to, they should not see inactive users as an option.

## AC
- [ ] Limit available options of task assignees to active users on the front end

## Notes
[Here's some possible context](https://github.com/department-of-veterans-affairs/caseflow/issues/11631#issuecomment-522193687)

kevmo:
Resolves #12949

### Description
In "Assign to Person"  task assignment modals, we were offering inactive users 

### Acceptance Criteria
- [ ] Test suite passes.

### Testing Plan
1. Log in as translation team member #1, e.g.  ORG_QUEUE_USER_1
2. Navigate to http://localhost:3000/organizations/translation, open a case (e.g. http://localhost:3000/queue/appeals/2e7d0f5b-f9eb-4767-9b41-40426d9b0562)
3. In task actions, go to "Assign to Person".  You should see Translation team member #5. 
4. Now make Translation team member #5 as inactive in the rails console: 
`User.find_by_css_id("ORG_QUEUE_USER_5").inactive!`
5. Repeat steps 1 through 3.  You should no longer see Team Member 5.

### User Facing Changes
BEFORE: 
Screen Shot 2020-01-13 at 9 54 00 AM

AFTER:
Screen Shot 2020-01-13 at 9 58 46 AM


client/COPY.json
```diff
@@ -230,7 +230,7 @@
   "ASSIGN_WIDGET_NO_TASK_DETAIL": "Please select a task.",
   "ASSIGN_WIDGET_SUCCESS": "%(verb)s %(numCases)s %(casePlural)s",
   "ASSIGN_WIDGET_ASSIGNMENT_ERROR_TITLE": "Error assigning tasks",
-  "ASSIGN_WIDGET_ASSIGNMENT_ERROR_DETAIL": "One or more tasks couldn't be assigned. Reassign tasks to a judge on the case deatils screen",
+  "ASSIGN_WIDGET_ASSIGNMENT_ERROR_DETAIL": "One or more tasks couldn't be assigned. Reassign tasks to a judge on the case details screen",
```
kevmo: just a typo I found.

spec/factories/user.rb
```diff
@@ -32,6 +32,10 @@
       roles { ["Mail Intake"] }
     end
 
+    trait :inactive do
+      status { "inactive" }
+    end
```
hschallhorn: We should get this for free with `user.inactive?`

app/repositories/task_action_repository.rb
```diff
@@ -344,5 +338,17 @@ def users_to_options(users)
         }
       end
     end
+
+    def potential_task_assignees(task)
+      if task.assigned_to.is_a?(Organization)
+        task.assigned_to.users.reject(&:inactive?)
```
hschallhorn: Enums come with scopes for free as well. This can be simplified to `task.assigned_to.users.active`

app/repositories/task_action_repository.rb
```diff
@@ -344,5 +338,17 @@ def users_to_options(users)
...
+      elsif task.parent&.assigned_to.is_a?(Organization)
+        task.parent.assigned_to.users.reject do |check_user|
```
hschallhorn: Same thing here, we can keep the original logic and just add `.active` after `users`

app/repositories/task_action_repository.rb
kevmo: Tried to go that route, but it doesn't work because `task.assigned_to.users` is a Ruby array, not an Active Record object.

spec/factories/user.rb
hschallhorn: Blarg, didn't see this was in the factory. Ignore.

PR #13134, opened 2020-01-10
RootTask closes itself if all child tasks will be closed after_save

### LINKED ISSUE #13133: Appeals with post-dispatch tasks remain open
## Description

When post-dispatch tasks are opened on an Appeal (e.g. Mail tasks, or `InformalHearingPresentationTask`), the Root Task may remain on hold indefinitely and the Appeal appears "stuck" with no definite status. This is because [the status of the Appeal is derived from the state of its task tree](https://github.com/department-of-veterans-affairs/caseflow/issues/12713#issuecomment-554015297)

#12238

The `AppealsWithNoTasksOrAllTasksOnHoldQuery` job reveals about 30 of these are consistently open.

## Technical notes

The existing mechanism for `update_parent_status` should be used to close the Root task where necessary. One way to do this would be to define a `Task.post_dispatch_task?` boolean method on the superclass and override it where necessary in the relevant child task classes. Then the `RootTask.when_child_task_completed` method could interrogate each child to see if the RootTask should close itself.

pkarman:
Resolves #13133 

### Description
Use existing callbacks to allow RootTask to close itself when there are only auto-close-able child tasks remaining.

This applies primarily to tasks opened *after* a case has been dispatched.

Tested in production, the `.post_dispatch_task?` correctly identifies 28 of the 30 stuck appeals.
```ruby
stuck_appeals.select { |appeal| appeal.tasks.any? { |t| t.open? && t.post_dispatch_task? } }
```

### DIFF
--- a/app/models/tasks/root_task.rb
+++ b/app/models/tasks/root_task.rb
@@ -45,7 +45,7 @@ class RootTask < Task
   def update_children_status_after_closed
     transaction do
       # do not use update_all since that will avoid callbacks and we want versions history.
-      children.open.where(type: CHILD_TASK_TYPES_TO_CLOSE_AFTER_CLOSED).to_a.each(&:completed!)
+      children.open.where(type: CHILD_TASK_TYPES_TO_CLOSE_AFTER_CLOSED).find_each(&:completed!)
     end
   end
 
--- a/app/models/tasks/root_task.rb
+++ b/app/models/tasks/root_task.rb
@@ -9,11 +9,19 @@ class RootTask < Task
   # Set assignee to the Bva organization automatically so we don't have to set it when we create RootTasks.
   after_initialize :set_assignee, if: -> { assigned_to_id.nil? }
 
+  CHILD_TASK_TYPES_TO_CLOSE_AFTER_CLOSED = [TrackVeteranTask.name].freeze
+
   def set_assignee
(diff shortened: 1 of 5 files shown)


app/models/tasks/root_task.rb
```diff
@@ -35,7 +43,14 @@ def when_child_task_created(child_task)
   end
 
   def update_children_status_after_closed
-    children.open.where(type: TrackVeteranTask.name).update_all(status: Constants.TASK_STATUSES.completed)
+    transaction do
+      # do not use update_all since that will avoid callbacks and we want versions history.
+      children.open.where(type: CHILD_TASK_TYPES_TO_CLOSE_AFTER_CLOSED).to_a.each(&:completed!)
```
hschallhorn: Don't need to convert to array here.

pkarman: I thought so too, but codeclimate would prefer I use `find_each` instead and that isn't what we want.

hschallhorn: 😾 damn code climate

pkarman: I mean, I guess we could use `find_each` but it seemed less intuitive to me. happy to switch if you'd prefer.

pkarman: I changed it.

PR #13143, opened 2020-01-13
unset session pg_user_id if user not found

pkarman:
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/7634/

### Description
If a Rails session `pg_user_id` is stale and no longer points at a valid User row, unset the session variable.

### DIFF
--- a/.reek.yml
+++ b/.reek.yml
@@ -69,6 +69,8 @@ detectors:
       - TaskTimerJob
       - UpdateCachedAppealsAttributesJob#assignees_for_vacols_id
       - UserReporter#report_user_related_records
+      - User#current_case_assignments_with_views
+      - User#appeal_has_task_assigned_to_user?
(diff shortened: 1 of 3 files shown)


.reek.yml
```diff
@@ -78,6 +80,7 @@ detectors:
       - Task
       - CaseflowJob
       - Api::V3::DecisionReview::HigherLevelReviewIntakeProcessor
+      - User
```
jcq: For my own understanding, what does adding something in here do?

pkarman: it quiets the reek (codeclimate) warnings.

in this case we probably wouldn't have seen new ones since codeclimate keeps track of those we have manually ignored in the past, but locally when I run it, it was warning about those. I knew we would likely never change them, so explicitly marked them in .reek.yml to keep track of the explicit exception.

Issue #12166, opened 2019-09-19
BGS::ShareError No Records Returned


leikkisa:
## Description
We're getting alerts for BGS::ShareError: No records returned, when syncing an established end product. This may be related to de-duping veteran records. When I looked up the veteran by participant ID, I got a different record for them. The current filenumber and SSN don't match, and no longer return a veteran record.

- [ ] Add an error class to ruby-bgs so that this doesn't come to us as a generic BGS::ShareError, and can be rescued and handled
- [ ] Investigation: Look at Sentry cases and check if we can reliably find a new file number for them.

## Out of scope
There is another ticket that is a partial dupe (for implementing a fix) #7734.  I think it's worth adding the BGS error class, and investigating the solution first, and then moving onto implementation.

## Background/context/resources
Example: https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/6641/events/570362/
https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/6641/events/482296/

Related: https://github.com/department-of-veterans-affairs/caseflow/issues/7734

nanotone:
Related to #12158

nanotone:
1. PR for the new ShareError subclass: department-of-veterans-affairs/ruby-bgs#86
2. I investigated 5 recent "No records returned" errors, by looking up the participant ID in BGS and seeing if the returned record was usable for each veteran:
- In 4 out of 5 cases, the new record matched our veteran's full name.
  - For 3 of these, the SSN/file number was clearly a corrected typo -- 2 had adjacent digits swapped, 1 was off-by-one).
  - For the last one, the SSN/file number was significantly different, but the birthdate matched so I'm confident it's the same person.
- In 1 case out of 5, the participant ID also didn't exist in BGS.

PR #13150, opened 2020-01-14
BGS NoRatingsExistForVeteran error

pkarman:
Updates the BGS gem to use the new NoRatingsExistForVeteran exception, to mirror changes upstream.

https://dsva.slack.com/archives/CHX8FMP28/p1579008773084900

see also #12166

### DIFF
--- a/app/models/rating.rb
+++ b/app/models/rating.rb
@@ -138,7 +138,7 @@ class Rating
       end
 
       unsorted.sort_by(&:promulgation_date).reverse
-    rescue Savon::Error
+    rescue Savon::Error, BGS::ShareError
       []
     end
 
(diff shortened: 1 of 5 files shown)


app/models/rating.rb
```diff
@@ -138,7 +138,7 @@ def fetch_in_range(participant_id:, start_date:, end_date:)
       end
 
       unsorted.sort_by(&:promulgation_date).reverse
-    rescue Savon::Error
+    rescue Savon::Error, BGS::ShareError
```
nanotone: Did you intend to trap all ShareErrors, or just the new NoRatingsExistForVeteran?

pkarman: all of them.

PR #13169, opened 2020-01-15
Fix inactive judge team error and selectable_orgs flake

hschallhorn:
Resolves error seen by judges when their team is made inactive.

### Description
When grabbing or users and mapping to the related organization, inactive teams are returned as `nil` causing an error when the name of the org is [queried](https://github.com/department-of-veterans-affairs/caseflow/blob/e21282bd6a16402c20171a6f7f23f29cf360cfdf/app/controllers/application_controller.rb#L41).

### Acceptance Criteria
- [ ] No errors when a users team is made inactive

### Testing Plan
1. Make a user's team they admin inactive
```ruby
user = User.find_by(css_id: "BVAAABSHIRE")
JudgeTeam.for_judge(user).inactive!
```
2. Navigate to that user's queue
3. Ensure no errors

### DIFF
--- a/app/models/organizations/judge_team.rb
+++ b/app/models/organizations/judge_team.rb
@@ -23,7 +23,7 @@ class JudgeTeam < Organization
     end
 
     def use_judge_team_roles?
-      FeatureToggle.enabled?(:use_judge_team_role)
+      FeatureToggle.enabled?(:judge_admin_scm)
     end
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -262,7 +262,7 @@ class User < ApplicationRecord
   end
 
   def administered_teams
-    organizations_users.select(&:admin?).map(&:organization)
+    organizations_users.select(&:admin?).map(&:organization).compact
   end
 
   def administered_judge_teams
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -319,10 +319,9 @@ describe User, :all_dbs do
 
     context "when the user is a judge team admin" do
       let(:judge_team) { create(:judge_team, :has_judge_team_lead_as_admin) }
-      let(:judge) { judge_team.judge }
+      let!(:judge) { judge_team.judge }
 
       before do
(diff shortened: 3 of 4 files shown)


app/models/user.rb
```diff
@@ -262,7 +262,7 @@ def current_case_assignments
   end
 
   def administered_teams
-    organizations_users.admin.map(&:organization)
+    organizations_users.admin.map(&:organization).compact
```
hschallhorn: Real change

app/models/user.rb
```diff
@@ -343,7 +343,7 @@ def show_reader_link_column?
   private
 
   def inactive_judge_team
-    JudgeTeam.unscoped.inactive.find_by(id: organizations_users.select(&:admin?).pluck(:organization_id))
+    JudgeTeam.unscoped.inactive.find_by(id: organizations_users.admin.pluck(:organization_id))
```
hschallhorn: Missed this change in old PR

app/models/user.rb
kevmo: sick and tired of establishment code, looking for that real change

spec/models/user_spec.rb
```diff
@@ -318,8 +318,8 @@
     end
 
     context "when the user is a judge team admin" do
-      let(:judge) { create(:user) }
-      let!(:judge_team) { create(:judge_team) }
+      let(:judge_team) { create(:judge_team, :has_judge_team_lead_as_admin) }
+      let(:judge) { judge_team.judge }
```
hschallhorn: Should fix flakes on `selectable_organizations`

app/models/organizations/judge_team.rb
```diff
@@ -23,7 +23,7 @@ def create_for_judge(user)
     end
 
     def use_judge_team_roles?
-      FeatureToggle.enabled?(:use_judge_team_role)
+      FeatureToggle.enabled?(:judge_admin_scm)
```
hschallhorn: Updating judge team roles to be under SCM flag

Issue #12942, opened 2019-12-12
Mark judge teams inactive when the judge becomes inactive


hschallhorn:
In an effort to remove a user from any organizations they are a part of (#12940), `JudgeTeam`s should be handled separately. Rather than removing an inactive judge from a `JudgeTeam`, we should mark these `JudgeTeam`s as inactive

## AC
- [ ] When a judge is made inactive, so is their corresponding `JudgeTeam`

lomky:
_[What is this graph?](https://github.com/department-of-veterans-affairs/caseflow/issues/11533)_
```
1 | |||||
2 | ||||||||||
3 |  
5 | 
8 | 
```
Estimated at a 2.

**Notes**

Do users have versioning? Not currently. Some disagreement on the team whether we want to track this in versions vs log analysis. Do we need to keep the entire version ledger in prods db? Later topic.

Inactive Organizations is new, just finishing PR. Two new fields on Orgs (`status` & `status_updated_at`), largely mirrors the inactive user. 

If a judge gets marked inactive, go update the status in their judge_team to inactive. This will need a couple tests, but otherwise straightforward. Do not empty inactive judge_teams of their atty/admin members.

Should we affect Atty task work ability if they belong to an inactive judge team? No, that should be handled elsewhere.

Its a little strange that we remove inactive users from active orgs, but not active users from inactive orgs.  We should follow up with the board what the preferred business process would be here. We could implement behavior if they do know what they want to happen.

PR #13172, opened 2020-01-16
remove inactive users from all orgs except judges from JudgeTeams

### LINKED ISSUE #12940: Remove inactive users from all organizations
As a followup to #12303, we would like to remove a user from their organizations when they are marked inactive. This excludes removing judges from their `JudgeTeam`s as that will be handled in another ticket

## AC
- [ ] When a user is made inactive, they are removed from every organization they are a member or admin of, other than judges from `JudgeTeam`s

## Notes
As judge teams are the only organization that somewhat rely on there being a dedicated admin (at least until SCM epic is finished), there should be no ill effects of an organization that loses an admin.

yoomlam:
Resolves #12940 

### Description
Remove a user from their organizations when they are marked inactive, except do not remove judges from their JudgeTeam (they are marked inactive #12942).

Changes:
```diff
-    remove_user_from_auto_assign_orgs
+    remove_user_from_orgs
```

### Acceptance Criteria
- [x] Code compiles correctly and tests pass

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -275,7 +275,7 @@ class User < ApplicationRecord
 
   def selectable_organizations
     orgs = organizations.select(&:selectable_in_queue?)
-    judge_team_judges = (JudgeTeam.for_judge(self) || judge_in_vacols?) ? [self] : []
+    judge_team_judges = judge? ? [self] : []
     judge_team_judges |= administered_judge_teams.map(&:judge) if FeatureToggle.enabled?(:judge_admin_scm)
 
     judge_team_judges.each do |judge|
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -362,9 +362,9 @@ class User < ApplicationRecord
 
   def remove_user_from_orgs
     removal_orgs = organizations
-    removal_orgs = removal_orgs.reject { |org| org.is_a?(JudgeTeam) && org.judge == self } if judge?
-    removal_orgs.each do |organization|
-      OrganizationsUser.remove_user_from_organization(self, organization)
+    my_judge_team = JudgeTeam.for_judge(self)
+    removal_orgs.each do |org|
+      OrganizationsUser.remove_user_from_organization(self, org) unless org == my_judge_team
     end
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -289,7 +289,7 @@ class User < ApplicationRecord
   end
 
   def judge?
-    JudgeTeam.for_judge(self) || judge_in_vacols?
+    !!JudgeTeam.for_judge(self) || judge_in_vacols?
   end
 
   def update_status!(new_status)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -288,6 +288,14 @@ class User < ApplicationRecord
     orgs
   end
 
+  def is_judge?
+    if FeatureToggle.enabled?(:use_judge_team_role)
+      JudgeTeam.for_judge(self) || judge_in_vacols?
+    else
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -288,6 +288,14 @@ def selectable_organizations
     orgs
   end
 
+  def judge?
+    if FeatureToggle.enabled?(:use_judge_team_role)
```
hschallhorn: Even if we are not using judge team roles, we still want to check if there is a judge team before . we make a call to vacols. I'd remove this check.

app/models/user.rb
```diff
@@ -352,13 +360,14 @@ def user_reactivation
...
+    remove_user_from_orgs
     JudgeTeam.for_judge(self)&.inactive!
   end
 
-  def remove_user_from_auto_assign_orgs
-    auto_assign_orgs = organizations.select(&:automatically_assign_to_member?)
-    auto_assign_orgs.each do |organization|
+  def remove_user_from_orgs
+    removal_orgs = organizations
+    removal_orgs = removal_orgs.reject { |org| org.is_a?(JudgeTeam) } if judge?
```
hschallhorn: Here we're only checking if the user is a member of a judge team. Instead, we should reject the team that the user is the judge for. I'd add a check here to make sure `org.judge == self`

hschallhorn: There are a lot of places in the code where we check `org.is_a?(JudgeTeam)` in a loop. We should offload this to the database. `removal_orgs.where.not(type: JudgeTeam.name)`. I'
ll file a ticket to clean up the rest of the code.

yoomlam: In the new ticket, we should be careful of potential future subclasses of `JudgeTeam` that won't be found by `.where.not(type: JudgeTeam.name)`.

hschallhorn: But also, that second comment I made was nullified by the first 🤷‍♂comment.

yoomlam: The `.where.not(type: JudgeTeam.name)` also does not catch organizations with `type=nil` in tests that `create(:organization)`.

hschallhorn: Hm, we should update factory bot to include types then. On that now.

hschallhorn: #13177

hschallhorn:
Let's user the new `judge?` function [here](https://github.com/department-of-veterans-affairs/caseflow/pull/13172/files#diff-4676c008b11a5480d73d4a6de01e45b9L278) as well.

Adding `before { allow(user).to receive(:judge?).and_return(false) }` [here](https://github.com/department-of-veterans-affairs/caseflow/pull/13172/files#diff-12b107c16792b9ecba685e51b51826f1R750) may solve the flake as well

app/models/user.rb
```diff
@@ -362,9 +362,9 @@ def user_inactivation
...
   def remove_user_from_orgs
     removal_orgs = organizations
-    removal_orgs = removal_orgs.reject { |org| org.is_a?(JudgeTeam) && org.judge == self } if judge?
-    removal_orgs.each do |organization|
-      OrganizationsUser.remove_user_from_organization(self, organization)
+    my_judge_team = JudgeTeam.for_judge(self)
+    removal_orgs.each do |org|
+      OrganizationsUser.remove_user_from_organization(self, org) unless org == my_judge_team
     end
   end
```
yoomlam: @hschallhorn Significantly changed the code after your review. What do you think about this current version?

hschallhorn: 👌 I like it! This will probably error for an inactive judge team until #13169 is merged

Issue #13069, opened 2019-12-30
Motion to Vacate | Granted | Add vacated decision issues to new appeal stream


leikkisa:
## User or job story
User story: As a Judge, after I address a motion to vacate, if there are granted issues, the new appeal stream should come pre-populated with the request issues contesting the decisions marked for vacature. Those should also get decision issues with the disposition of "vacated".

## Acceptance criteria
- [ ] Please put this work behind the feature toggle: [review_motion_to_vacate]
- [ ] When a new appeal stream is created from a PostDecisionMotion, automatically add the request issues (contesting the original decision issues)
- [ ] Automatically add decision issues for each of those request issues with the disposition of "vacated" (see copy below)

## Open questions
- Final copy for decision issue description and decision text

## Out of scope
UI for the next screen after clicking "Continue"

## Release notes
N/A - Backend ticket

### Designs
Copy for vacated decision issue descriptions:
The issue: [issue description] has been vacated.

### Technical notes
- Please see #12373 - this ticket has the code for creating the request issues needed, but the action is not currently being triggered.

Decision issue fields to populate:
- disposition: "vacated"
- description: The decision: "[decision issue description]" has been vacated.
- participant_id
- decision_review_type: "Appeal"
- decision_review_id: [appeal.id]
- benefit_type: [use same benefit type as request issue]
- caseflow_decision_date: [Date.today]

leikkisa:
@jcq This work is still relevant, so let's keep this ticket open until we get to the point of adding decision issues.

leikkisa:
The scope of this ticket has been updated to reflect adding these issues to a new appeal stream.

PR #13173, opened 2020-01-16
Create Vacated Decision Issues from Request Issues

jcq:
Connects #13069

## Description
This adds code to create vacated decision issues from given request issues


app/models/post_decision_motion.rb
```diff
@@ -27,8 +27,12 @@ def vacated_issues
     DecisionIssue.find(vacated_decision_issue_ids)
   end
 
-  def create_request_issues_for_vacature
-    vacated_issues.map(&:create_contesting_request_issue!)
+  def request_issues_for_vacature
+    @request_issues_for_vacature ||= vacated_issues.map(&:create_contesting_request_issue!)
```
leikkisa: This memoization won't persist, and we wouldn't want it to create contesting request issues again if it gets called again later.  You might want to keep the "create" method, and different code when accessing them.
`RequestIssue.where(contested_decision_issue_id: vacated_issues.map(&:id))`

leikkisa: This makes me think we should also use `find_or_create_by` in `create_contesting_request_issue!`, or return if the `decision_issue` already has a `contesting_request_issue` as a safety.

app/models/post_decision_motion.rb
```diff
@@ -27,8 +27,12 @@ def vacated_issues
...
+  end
+
+  def vacated_decision_issues
+    @vacated_decision_issues ||= request_issues_for_vacature.map(&:create_vacated_decision_issue!)
```
leikkisa: Same comments here, it's probably better to separate creating the decision issues from getting them later. It'd be good to make create_vacated_decision_issue! idempotent.

app/models/request_issue.rb
```diff
@@ -471,6 +471,18 @@ def create_decision_issue_from_params(decision_issue_param)
...
+      decision_review: decision_review,
+      decision_review_type: decision_review_type,
+      disposition: "vacated",
+      description: "The decision: #{description} has been vacated.",
+      caseflow_decision_date: Time.zone.today,
+      benefit_type: benefit_type,
+      participant_id: decision_review.veteran.participant_id
+    )
+  end
+
```
leikkisa: This also needs to be linked to the request issue being decided on. I missed adding that on the ticket.  There's a separate table for this called RequestDecisionIssue.  Since you're in the RequestIssue model, instead of `DecisionIssue.create!` you could use `decision_issues.create!`, and it will add that association.

spec/models/post_decision_motion_spec.rb
```diff
@@ -46,10 +48,22 @@
...
-      post_decision_motion.create_request_issues_for_vacature
+      expect(appeal.request_issues.size).to eq 0
+      post_decision_motion.request_issues_for_vacature
       appeal.reload
-      expect(appeal.request_issues.length).to eq 3
+      expect(appeal.request_issues.size).to eq 3
+    end
+  end
+
+  context "handles creation of vacated decision issues" do
```
leikkisa: For these unit tests, I think the norm is to name the spec after the method.
`context "#vacated_decision_issues" do`, and set the subject for the context `subject { post_decision_motion.vacated_decision_issues }`.

Then in your `it` block, you can call `subject` instead of `post_decision_motion.vacated_decision_issues`. That's particularly useful on larger tests. This feedback similarly applies to the above test.

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20191223175715) do
+ActiveRecord::Schema.define(version: 20200102193042) do
```
leikkisa: I don't see a migration, so might want to leave this out.

jcq: Ugh, yeah. That was due to merge conflicts during code freeze. I'll try to just revert that piece, since I think they've cleaned up master in that regard now.

jcq: Actually... that's the same version as master now, so things should just work themselves out by merging that in. Whew!

app/models/decision_issue.rb
```diff
@@ -160,7 +160,7 @@ def associated_request_issue
   end
 
   def create_contesting_request_issue!
-    RequestIssue.create!(
+    RequestIssue.find_or_create_by!(
       decision_review: decision_review,
```
leikkisa: These are now going to be added onto the new appeal stream instead of the original appeal

app/models/request_issue.rb
```diff
@@ -472,6 +472,18 @@ def create_decision_issue_from_params(decision_issue_param)
     )
   end
 
+  def create_vacated_decision_issue!
+    decision_issues.find_or_create_by!(
+      decision_review: decision_review,
```
leikkisa: This should be okay since the request issue will be on the new appeal stream

app/controllers/post_decision_motions_controller.rb
```diff
@@ -25,6 +25,17 @@ def return_to_lit_support
     render json: { tasks: ::WorkQueue::TaskSerializer.new(appeal_tasks, is_collection: true) }
   end
 
+  def create_issues
```
leikkisa: I think it makes sense to move creating issues to the PostDecisionMotionUpdater, when the new appeal stream is getting created

jcq: Yeah, if their creation is automatic we don't need an endpoint for it.

Methinks it makes sense to wait until your PR (#13325) lands and then incorporate upon new stream creation.

app/models/decision_issue.rb
```diff
@@ -160,8 +160,9 @@ def associated_request_issue
   end
 
   def create_contesting_request_issue!
-    RequestIssue.create!(
-      decision_review: decision_review,
+    vacate_appeal_stream = Appeal.find_by(stream_type: "vacate", stream_docket_number: decision_review.docket_number)
```
leikkisa: This is good for now, we'll probably update it in Sandra's ticket to pass in the decision review so it can be used for De Novo's too.

config/routes.rb
```diff
@@ -301,6 +301,7 @@
   end
 
   post "post_decision_motions/return", to: "post_decision_motions#return_to_lit_support"
+  post "post_decision_motions/:id/create_issues", to: "post_decision_motions#create_issues"
```
leikkisa: Is this needed or being used?

spec/controllers/post_decision_motions_controller_spec.rb
```diff
@@ -84,6 +86,51 @@
     end
   end
 
+  describe "#create_issues" do
```
leikkisa: I think this can get cleaned up (removed) since there's not an endpoint for this anymore

spec/workflows/post_decision_motion_updater_spec.rb
```diff
@@ -65,6 +65,8 @@
           expect(task.reload.status).to eq Constants.TASK_STATUSES.completed
           verify_vacate_stream
         end
+
+        it "should create decision issues on new vacate"
```
leikkisa: This looks like a stray line

spec/workflows/post_decision_motion_updater_spec.rb
```diff
@@ -235,4 +242,13 @@ def verify_vacate_stream
...
     expect(attorney_task).to_not be_nil
   end
+
+  def verify_decision_issues_created
+    motion = PostDecisionMotion.first
+    request_issues = vacate_stream.request_issues
+    expect(request_issues.size).to eq(motion.decision_issues_for_vacatur.size)
+
+    decision_issues = vacate_stream.decision_issues
+    expect(decision_issues.size).to eq(motion.decision_issues_for_vacatur.size)
```
leikkisa: 👍

config/routes.rb
jcq: Nope — will clean this out!

Issue #13165, opened 2020-01-15
BVA Intake | No BVA Policy Regarding Same RO or Station | VBA CorpDB Rule To Be Removed


msj0nes:
## Description
Issue escalated to product team. Issue noted in slack [thread](https://dsva.slack.com/archives/C54QCEHAL/p1579009181283200). BVA Intake user experiencing VBA [Corp DB] error message seen below. Error is due to VBA based rules not related to the Board. This rule or restriction should only be associated to 'reviews' and not 'appeals'.

Resolution Test User CSS ID:  BVAIMBUTLER 
Resolution Test Case: https://appeals.cf.ds.va.gov/queue/appeals/3596372 

![image](https://user-images.githubusercontent.com/41164406/72462698-a70a0180-379f-11ea-8f48-de6609aa20ca.png)

## Background/context/resources

**_Board confirmation received on January 15, 2020 12:55 PM from J. White [email]_**

Please remove the restriction as soon as possible.  Appeals are only handled at the Board so there is no possibility of restricting same station employees.  Thank you.

[**_API Info_**](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/services/external_api/bgs_service.rb#L193-L210)
**_Related Tickets_**
- https://github.com/department-of-veterans-affairs/caseflow/issues/10087#issuecomment-507783830)
**_Related Slack_**
- https://dsva.slack.com/archives/C54QCEHAL/p1579009181283200
- https://dsva.slack.com/archives/CFBLH2LR4/p1565816330125600

## Technical notes
N/A

msj0nes:
# OIT Consulted For Advisement
COR contacted for advisement as issue is now policy based requiring stakeholder business process changes.


**OIT Update Provided** 
From:  S. Salvato (OIT)
Wednesday, January 22, 2020 9:04 AM

Details to be discussed with T. Hamilton for next steps.

PR #13180, opened 2020-01-16
Skip same-station conflict check when intaking appeals

nanotone:
Connects #13165 

### Description
This PR disables the same-station conflict check when intaking appeals, as requested by the Board. Some decisions I made:

- Doing the check for appeals up in the models/validators layer, rather than the BGSService layer. This felt like application logic that our library code for external dependencies shouldn't have to worry about.
- Replacing `may_modify?` with `station_conflict?` and adjusting the logic accordingly. This makes the method name clearly describe what it's trying to detect. I felt that `may_modify?` was coupling station conflicts with sensitivity levels, and conflating these with the desired end effect (preventing failed establishments).

Happy to hear what folks' thoughts are on this small refactor; I'm still getting to understand the division of labor between Caseflow app logic, BGSService, and BGS itself.

### Acceptance Criteria
- [ ] Same-station conflict check remains in place for ClaimReviews
- [ ] Same-station conflict check is skipped for Appeals

### Testing Plan
1. All `may_modify?` tests updated to use `station_conflict?`
2. New validator test to check that appeals bypass the conflict check

### DIFF
--- a/spec/models/validators/intake_start_validator_spec.rb
+++ b/spec/models/validators/intake_start_validator_spec.rb
@@ -6,9 +6,11 @@ describe IntakeStartValidator, :postgres do
 
     let(:veteran) { create(:veteran) }
 
+    let(:review) { HigherLevelReview.new }
+
     let(:intake) do
       # IntakeStartValidator expects an uncommitted intake (hence new)
-      Intake.new(veteran_file_number: veteran.file_number, user: user)
+      HigherLevelReviewIntake.new(veteran_file_number: veteran.file_number, detail: review, user: user)
     end
 
     let(:validator) do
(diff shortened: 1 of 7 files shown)


nanotone:
@leikkisa @pkarman As per request from @msj0nes (see [Slack thread](https://dsva.slack.com/archives/C54QCEHAL/p1580400290138300) for context), I've put the same-station bypass behind a `:allow_same_station_appeals` feature toggle for now, and this PR should be ready to go once approved.

spec/models/validators/intake_start_validator_spec.rb
```diff
@@ -21,15 +23,37 @@
...
-    it "sets error_code \"veteran_not_modifiable\" when BGS does not allow modification" do
-      allow_any_instance_of(BGSService).to receive(:may_modify?) { false }
+    it "sets error_code \"veteran_not_modifiable\" when BGS shows a station conflict" do
+      allow_any_instance_of(BGSService).to receive(:station_conflict?) { true }
       expect(validate_error_code).to eq "veteran_not_modifiable"
     end
 
+    context "intaking an Appeal" do
+      let(:intake) do
+        AppealIntake.new(veteran_file_number: veteran.file_number, detail: review, user: user)
```
pkarman: why avoid the `intake` factory here?

nanotone: This was modeled after the `HigherLevelReviewIntake.new` further up in the same file, which is commented with `IntakeStartValidator expects an uncommitted intake (hence new)`.

PR #13187, opened 2020-01-16
Fix Hearing Time display on Virtual Hearing Emails

### LINKED ISSUE #13156: Virtual Hearings incorrect time on email reminder
- Observed: Hearing time (5:30pm EST) displayed in Veteran, Judge, and POA emails doesn't match the time (12:30pm EST) displayed in Caseflow. 
- Expected: Hearing time displayed in emails matches the time displayed in Caseflow.

image

ferristseng:
Resolves #13156

### Description

The cause of this issue is a discrepancy between the timezone of the server (EST), and the timezone of the worker environment (UTC).

Currently, Rails is interpreting the `scheduled_time` value we are saving as an EST value because the server's timezone is EST. The value is actually saved in UTC though, but it has the EST offset on it (`12:30` becomes `17:30` in the DB). Since the `scheduled_time` actually represents the local time at the regional office, we have to undo the EST offset first.

```
irb(main):015:0> puts Hearing.all.map(&:attributes_before_type_cast).map { |h| h["scheduled_time"] }
[2020-01-17 13:28:58 -0500]   Hearing Load (1.7ms)  SELECT "hearings".* FROM "hearings"
17:30:00.  # 12:30
17:30:00.  # 12:30
17:30:00.  # 12:30
20:30:00.  #  3:30
14:30:00   # etc...
14:00:00
17:30:00
13:30:00
``` 

The current `Hearing#scheduled_for` implementation works on the server side because Rails also automatically converts the `scheduled_time` to EST, so the time values have the appropriate correction:

```
DateTime.new.in_time_zone(regional_office_timezone).change(
      year: hearing_day.scheduled_for.year,
      month: hearing_day.scheduled_for.month,
      day: hearing_day.scheduled_for.day,
      # !! Hour has the offset applied on the server because the server is EST, but is not elsewhere !!
      hour: scheduled_time.hour,                               
      min: scheduled_time.min,
      sec: scheduled_time.sec
)
```

However in Rails console, and presumably in the worker's environment, since UTC is the default time zone, these values get skewed. The solution is to always apply the EST offset to get this working in environments where the default timezone is not set to EST.

---

Confirmed working in UAT:

```
irb(main):038:0>     Time.use_zone(regional_office_timezone) do
irb(main):039:1*       Time.zone.local(
irb(main):040:2*         hearing_day.scheduled_for.year,
irb(main):041:2*         hearing_day.scheduled_for.month,
irb(main):042:2*         hearing_day.scheduled_for.day,
irb(main):043:2*         scheduled_for_est.hour,
irb(main):044:2*         scheduled_for_est.min,
irb(main):045:2*         scheduled_for_est.sec
irb(main):046:2>       )
irb(main):047:1>     end
=> Mon, 20 Jan 2020 12:30:00 EST -05:00
```


app/models/hearing.rb
```diff
@@ -142,14 +142,26 @@ def advance_on_docket_motion
...
-    )
+    # The server timezone is America/New_York. The server interprets
+    # the time as an EST time when it gets saved, but the scheduled time
+    # value is saved in UTC (e.g. 12:30 gets saved as 17:30).
+    #
+    # This method undoes all the time conversions, and returns the final value
+    # in the timezone of the regional office. To do that, it needs to consider
+    # the fact that Rails will apply the local offset (EST) to the time
+    # value before it is saved.
+    scheduled_for_est = scheduled_time.utc.in_time_zone("America/New_York")
```
pkarman: we should make the server timezone explicit via config rather than hardcoding here. Maybe we can determine it via system config? that way if we ever move servers or env changes, it is automatically reflected here.

ferristseng: Looks like `Time.zone.getlocal.zone` can do this.

```
irb(main):024:0> Time.now.getlocal.zone
=> "EST"
irb(main):025:0> Time.zone.name
=> "UTC"
```

nanotone:
Just kibitzing here, but,
>The value is actually saved in UTC though, but it has the EST offset on it (12:30 becomes 17:30 in the DB)
if I understand the situation correctly, this goes against all time-data advice I know. If the Postgres column is tz-aware and stores UTC times, shouldn't all localization happen outside of the storage layer?

app/views/virtual_hearing_mailer/sections/_date_time.html.erb
```diff
@@ -2,5 +2,5 @@
   Date and Time
 </h3>
 <p>
-  <%= @virtual_hearing.hearing.scheduled_for.strftime("%A, %d %B %Y at %I:%M%P %Z") %>
+  <%= @virtual_hearing.hearing.time.local_time.strftime("%A, %d %B %Y at %I:%M%P %Z") %>
```
pkarman: so you've switched which method is used entirely for email presentation? That `.time` is a delegator to `HearingTimeService` which has hardcoded logic in for `America/New_York` already I see. So is the change to `scheduled_for` the right place to put the fix? Or should all the timezone shifting shenanigans be in the `HearingTimeService` class?

ferristseng:
> Just kibitzing here, but,
> 
> > The value is actually saved in UTC though, but it has the EST offset on it (12:30 becomes 17:30 in the DB)
> 
> if I understand the situation correctly, this goes against all time-data advice I know. If the Postgres column is tz-aware and stores UTC times, shouldn't all localization happen outside of the storage layer?

Yeah, I think the right approach to storing the time, would be to apply the timezone before it gets saved (but still store it as UTC in the database). So, the client sends `12:30`, then the server should explicitly interpret that as `12:30` in the TZ of the regional office before it gets saved (rather than implicitly interpreting it as `12:30` in server time).

app/models/hearing.rb
```diff
@@ -142,14 +142,27 @@ def advance_on_docket_motion
...
-    )
+    # The server timezone is America/New_York. The server interprets
+    # the time as an EST time when it gets saved, but the scheduled time
+    # value is saved in UTC (e.g. 12:30 gets saved as 17:30).
+    #
+    # This method undoes all the time conversions, and returns the final value
+    # in the timezone of the regional office. To do that, it needs to consider
+    # the fact that Rails will apply the local offset (EST) to the time
+    # value before it is saved.
+    system_tz = Time.now.getlocal.zone
```
pkarman: this seems to work. When I run it locally I get `CST` which is where I'm sitting.

spec/mailers/virtual_hearing_mailer_spec.rb
```diff
@@ -40,14 +54,71 @@
...
+      let(:hearing) do
+        create(
+          :legacy_hearing,
+          case_hearing: case_hearing,
+          hearing_day: hearing_day,
+          hearing_day_id: hearing_day.id
+        )
+      end
+
+      it_should_behave_like "it has the correct time in the email body", "10:30am", "7:30am"
```
pkarman: the time is not frozen in this test, but I think we'll miss a daylight savings time bug with those particular times of day (i.e. if you picked 2am then the offset isn't always guaranteed to be 3 hours).

app/views/virtual_hearing_mailer/sections/_date_time.html.erb
ferristseng: Yeah, the hardcoded logic in `HearingTimeService` is appropriate because it's handling the case for legacy hearings, where the time is stored in EST.

From what I can tell, the expectation of `HearingTimeService` is that `Hearing#scheduled_for` returns the scheduled time in the timezone of the regional office. The issue was that this assumption was broken when the timezone wasn't configured as EST. There might be other places where a similar fix might be necessary though.

app/models/hearing.rb
pkarman: in production `getlocal` is in fact `EST`

ferristseng: This appears to be where we're setting the timezone:

https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/application_controller.rb#L209-L212

Commenting the `Time.zone = ...` line out, makes `zone` come back as `UTC` from the server. It's setup so the timezone of the request is always that of the user making the request.

If that's the case, it makes sense that the jobs run with a TZ of UTC since there isn't a current user. But, it makes me think there might be a possible bug where if someone from the west coast looks at a hearing scheduled by someone in the east coast, the hearing time will be wrong. I will try to reproduce.

ferristseng: Unfortunately, there are also just places where we set the timezone explicitly to EST, but never change it back. Just loading the hearings page forces the application to use EST timezone.

https://github.com/department-of-veterans-affairs/caseflow/blob/8b65814b28a99ff5c198092ed679db3aeaa9aaf1/app/helpers/vacols_helper.rb#L49

nanotone: Whoa, this is just a shared global variable? Might it be worth dangefiling this (e.g. anytime `Time.zone =` gets added within `/app`)?

ferristseng: Yeah, I like that idea. We probably shouldn't be touching `Time.zone` directly -- `Time.use_zone` seems to be a good alternative.

spec/controllers/home_controller_spec.rb
```diff
@@ -14,18 +14,28 @@
...
 
-      it "should set timezone in session if is not set" do
+      it "should set timezone in session if is not set and restore it" do
+        initial_time_zone = Time.zone
         expect(@request.session[:timezone]).to eq nil
+
         get :index
-        expect(Time.zone.name).to eq current_user.timezone
+        allow(Time).to receive(:zone=).and_call_original
+        allow(Time).to receive(:zone=).with(current_user.timezone).once.and_call_original
```
tomas-nava: I don't think these allows (or those below) have any effect on the test; it passes whether they're there or not. Should they be expectations?

spec/controllers/home_controller_spec.rb
```diff
@@ -14,18 +14,28 @@
...
         expect(@request.session[:timezone]).to eq current_user.timezone
+
+        expect(Time.zone).to eq initial_time_zone
       end
```
tomas-nava: When I run this test locally, the initial value of `Time.zone` and `current_user.time_zone` are both `America/New_York`; maybe we should set them explicitly to different values?

spec/controllers/home_controller_spec.rb
ferristseng: I think the 2nd one is important. I can probably change the current user so their timezone isn't EST to make the test expectation more explicit

ferristseng: Updated `allow` to `expect`

spec/controllers/home_controller_spec.rb
ferristseng: Updated the current user's RO to the Chicago RO, so now their timezone is not America/New_York

app/helpers/vacols_helper.rb
```diff
@@ -46,12 +46,12 @@ def self.validate_presence(hash, required_fields)
...
 
   def self.day_only_str(date_time)
-    Time.zone = "Eastern Time (US & Canada)"
-
-    Time.zone.local(
-      date_time.year,
-      date_time.month,
-      date_time.day
-    ).strftime("%Y-%m-%d")
+    Time.use_zone("America/New_York") do
```
pkarman: could we define this TZ string as a constant with a comment so that it is clear why we are using ET explicitly?

app/models/vacols/record.rb
```diff
@@ -15,15 +15,16 @@ def self.relative_vacols_date(date_diff)
   end
 
   def self.rounded_current_time
-    Time.zone = "Eastern Time (US & Canada)"
-    current_time = Time.zone.now
+    Time.use_zone("America/New_York") do
```
pkarman: same comment re: constant (maybe re-use the same constant?)

app/services/hearing_time_service.rb
```diff
@@ -66,14 +66,20 @@ def central_office_time_string
...
     return @hearing.scheduled_for if @hearing.is_a?(Hearing)
 
-    # since regional office is from HearingDay, it can only be nil if it's a
-    # central office hearing in ET
-    regional_office = @hearing.regional_office_timezone || "America/New_York"
+    # for legacy hearings, convert to the regional office's time zone
+
+    # if the hearing's regional_office_timezone is nil, assume this is a
+    # central office hearing (eastern time)
+    regional_office_timezone = @hearing.regional_office_timezone || "America/New_York"
```
pkarman: same comment re: constant

app/helpers/vacols_helper.rb
```diff
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
 
 module VacolsHelper
+  VACOLS_DEFAULT_TIMEZONE = "America/New_York"
```
pkarman: maybe a comment here to explain why we are setting it explicitly instead of defaulting to `config/initializers/set_time_zone.rb`?

the rationale is that this is the timezone of the VACOLS server in the VA data center, not the timezone of the Rails server (even though they might, at times, be the same value).

app/models/hearing.rb
```diff
@@ -127,14 +127,41 @@ def advance_on_docket_motion
...
+    # whenever it's read from the database. Because that date did not fall
+    # during daylight savings time, the conversion remains the same no matter
+    # what time of year it's done.
+    #
+    # So when we need to display the time the hearing is scheduled for, we have
+    # to explicitly convert it to the time zone of the person who scheduled it,
+    # then assemble and return a TimeWithZone object cast to the regional
+    # office's time zone.
+
+    updated_by_timezone = updated_by&.timezone || Time.zone.name
```
pkarman: do we state anywhere in the UI that the user should be setting the time in *their* time zone, rather than the user's/RO's? I get the logic here, just wondering from a UX perspective if we are just assuming that the user isn't doing time zone math themselves trying to be "helpful".

spec/controllers/hearings_controller_spec.rb
```diff
@@ -294,12 +299,44 @@
...
+    let(:expected_time_zone) { "America/New_York" }
+    # for "America/New_York", "-04:00" or "-05:00" depending on daylight savings time
+    let(:utc_offset) do
+      hours, minutes = Time.zone.now.in_time_zone(expected_time_zone).utc_offset.divmod(60)[0].divmod(60)
+      hour_string = (hours < 0) ? format("%03i", hours) : format("+%02i", hours)
+      "#{hour_string}:#{format('%02i', minutes)}"
+    end
 
-    it "returns hearing details" do
+    subject do
```
pkarman: nice

spec/controllers/hearings_controller_spec.rb
```diff
@@ -312,6 +349,7 @@
       let!(:appeal) { create(:appeal) }
 
       it "returns an address" do
+        # Baltimore, MD regional office (eastern time)
         get :find_closest_hearing_locations,
             as: :json,
             params: { appeal_id: appeal.external_id, regional_office: "RO13" }
```
pkarman: maybe instead of a comment, just create a `let(:baltimore_ro_code) { "RO13" }` so that the code documents itself?

spec/controllers/home_controller_spec.rb
```diff
@@ -12,20 +12,33 @@
     end
 
     context "when visitor is logged in" do
-      let!(:current_user) { User.authenticate! }
+      let!(:current_user) do
+        # RO28 is the Chicago Regional Office
+        User.authenticate!(user: create(:user, regional_office: "RO28", station_id: 328))
```
pkarman: same comment for Chicago: `let(:chicago_ro_code) { "RO28" }`

app/models/hearing.rb
tomas-nava: The interface offers pre-baked times based on the RO that are assumed to be in the RO's timezone. The part where it gets saved in the user's timezone is a bug that this PR is a bandaid for, and that will require more extensive work to correct properly.

app/models/hearing.rb
```diff
@@ -127,14 +127,41 @@ def advance_on_docket_motion
...
+    # field. The time is converted to UTC upon save *relative to the timezone of
+    # the user who saved it*, not relative to the timezone of the RO where the
+    # veteran will attend the hearing. For example, if a user in New York
+    # schedules an 8:30am hearing for a veteran in Los Angeles, the time will be
+    # saved as 13:30 (with the eastern time +5 offset added) rather than 16:30
+    # (with the pacific time +8 offset added).
+    #
+    # The offset will *always* be +5, regardless of when the hearing is
+    # scheduled, because a date of 01 Jan 2000 is associated with the time
+    # whenever it's read from the database. Because that date did not fall
```
nanotone: Does the word `*always*` apply only to the example user in New York from the previous paragraph?

It's also a little confusing as to how the 2000/01/01 reference date is related. It sounds like that's just baked-in behavior from a lower level of abstraction?

app/helpers/vacols_helper.rb
tomas-nava: addressed in 50f76e3394b1bb0fbcba5eee03baa1844a088b23

spec/controllers/hearings_controller_spec.rb
tomas-nava: addressed in 50f76e3394b1bb0fbcba5eee03baa1844a088b23

spec/controllers/home_controller_spec.rb
tomas-nava: addressed in 50f76e3394b1bb0fbcba5eee03baa1844a088b23

app/models/hearing.rb
tomas-nava: addressed in 9b19aee5c8d42efc36275bbaf00496a56e7c93a0

Yes, the 2000/01/01 appears to be an (arbitrary?) date that Rails associates with a postgres `time` field. I couldn't find official documentation of this behavior with a quick search, just a lot of StackOverflow conversations [like this one](https://stackoverflow.com/questions/42538468/rails-why-time-datatype-start-at-1-1-2000).

Issue #12289, opened 2019-10-03
Allow user to add issues after the receipt date


leikkisa:
### User or job story
User story: As a Claims Assistant, when intaking a veteran's form, I need to be able to add issues whose promulgation date are in the future of the receipt date.

### Acceptance criteria
- [ ] This is behind a feature toggle - show_future_ratings
- [ ] During intake rating issues with promulgation dates after the receipt date are shown in the list of issues
- [ ] This does not apply to other types of contestable issues such as decision issues
- [ ] ~Checking for duplicates with decision issues should apply to all decision issues, not just ones before the receipt date~ - No need to check 
- [ ] Please include a screenshot in the PR with a "future issue" showing on the add issues modal

### Release notes
During intake or editing a decision review, users will now be able to see issues decided after the receipt date.  This will allow users to add issues where the decision date is incorrectly showing as in the future.

### Background
Issues that are backfilled and then promulgated are getting promulgation dates at that time instead of when the decision was actually made.  The preferred way to handle "missing ratings" is to backfill them. So when users are backfilling missing ratings, they are not showing up due to the date being after the receipt date.

We met with AMO to discuss this and got confirmation that promulgating backfilled ratings automatically sets the promulgation date which causes this problem. They have approved showing issues with future promulgation dates for this reason.

#### Technical notes
Currently issues in the future of the receipt date are filtered out from the list of contestable issues unless they are decisions from the same review.  So we can remove that filter in the list of contestable issues.

See the ContestableIssueGenerator from_ratings method.

Currently duplicates with decision issues are checked against contestable decision issues from before the receipt date, but we should filter out any rating issues that have a decision issue at all.

#### Resources/other links
Examples of occurrences
https://dsva.slack.com/archives/CHX8FMP28/p1570124151302100
https://dsva.slack.com/archives/C754H751T/p1571067626001500
https://dsva.slack.com/archives/CFBLH2LR4/p1575639629381400
https://dsva.slack.com/archives/CFBLH2LR4/p1576533492022400
https://dsva.slack.com/archives/CHX8FMP28/p1578325651240000
https://dsva.slack.com/archives/CHX8FMP28/p1579118731109700

Thread to discuss:
https://dsva.slack.com/archives/CFBLH2LR4/p1577740568037100

### Outstanding questions
How do we communicate this to the users? We would have to explain that their issue might show a decision date in the future, when the issue being contested could be from 1990.

pkarman:
[...]

PR #13195, opened 2020-01-17
Allow user to add issues after the receipt date

Sjones352:
Connects #12289 
Screen Shot 2020-01-17 at 4 35 23 PM

Screen Shot 2020-01-17 at 4 35 04 PM


app/workflows/contestable_issue_generator.rb
```diff
@@ -14,6 +14,18 @@ def contestable_issues
     contestable_rating_issues + contestable_decision_issues + contestable_rating_decisions
   end
 
+  def from_ratings
+    return [] unless receipt_date
+
+    issues = rating_issues
+
+    unless FeatureToggle.enabled?(:show_future_ratings, user: RequestStore[:current_user])
+      issues.reject! { |issue| issue.profile_date && issue.profile_date.to_date > receipt_date }
```
nanotone: In the unlikely but plausible situation that the `rating_issues` array is persistent, this `reject!` line would mutate that array and may cause subtle bugs down the road. What do you think about using just `reject` with an assignment, to be safe?

spec/workflows/contestable_issue_generator_spec.rb
```diff
@@ -107,4 +107,15 @@
...
+
+  describe "#contestable_issues with future ratings" do
+    subject { ContestableIssueGenerator.new(review).from_ratings }
+    before { FeatureToggle.enable!(:show_future_ratings) }
+    after { FeatureToggle.disable!(:show_future_ratings) }
+
+    it "when show_future_ratings feature toggle is enabled" do
+      expect(subject.count).to eq(3)
+      expect(subject.first.description).to eq "Future Rating issue"
+    end
```
nanotone: Perhaps it'd be worth it to have a sibling test for when `:show_future_ratings` isn't enabled?

app/controllers/intakes_controller.rb
```diff
@@ -89,7 +89,8 @@ def index_props
         rampIntake: FeatureToggle.enabled?(:ramp_intake, user: current_user),
         editContentionText: FeatureToggle.enabled?(:edit_contention_text, user: current_user),
         unidentifiedIssueDecisionDate: FeatureToggle.enabled?(:unidentified_issue_decision_date, user: current_user),
-        verifyUnidentifiedIssue: FeatureToggle.enabled?(:verify_unidentified_issue, user: current_user)
+        verifyUnidentifiedIssue: FeatureToggle.enabled?(:verify_unidentified_issue, user: current_user),
+        showFutureRatings: FeatureToggle.enabled?(:show_future_ratings, user: current_user)
```
leikkisa: The feature toggle operates in the backend only, so it's not necessary to send it to the front end.

app/views/appeals/edit.html.erb
```diff
@@ -12,7 +12,8 @@
       editContentionText: FeatureToggle.enabled?(:edit_contention_text, user: current_user),
       correctClaimReviews: FeatureToggle.enabled?(:correct_claim_reviews, user: current_user),
       unidentifiedIssueDecisionDate: FeatureToggle.enabled?(:unidentified_issue_decision_date, user: current_user),
-      verifyUnidentifiedIssue: FeatureToggle.enabled?(:verify_unidentified_issue, user: current_user)
+      verifyUnidentifiedIssue: FeatureToggle.enabled?(:verify_unidentified_issue, user: current_user),
+      showFutureRatings: FeatureToggle.enabled?(:show_future_ratings, user: current_user)
```
leikkisa: Same as first comment

app/views/higher_level_reviews/edit.html.erb
```diff
@@ -12,7 +12,8 @@
       editContentionText: FeatureToggle.enabled?(:edit_contention_text, user: current_user),
       correctClaimReviews: FeatureToggle.enabled?(:correct_claim_reviews, user: current_user),
       unidentifiedIssueDecisionDate: FeatureToggle.enabled?(:unidentified_issue_decision_date, user: current_user),
-      verifyUnidentifiedIssue: FeatureToggle.enabled?(:verify_unidentified_issue, user: current_user)
+      verifyUnidentifiedIssue: FeatureToggle.enabled?(:verify_unidentified_issue, user: current_user),
+      showFutureRatings: FeatureToggle.enabled?(:show_future_ratings, user: current_user)
```
leikkisa: Same as first comment

app/views/supplemental_claims/edit.html.erb
```diff
@@ -12,7 +12,8 @@
       editContentionText: FeatureToggle.enabled?(:edit_contention_text, user: current_user),
       correctClaimReviews: FeatureToggle.enabled?(:correct_claim_reviews, user: current_user),
       unidentifiedIssueDecisionDate: FeatureToggle.enabled?(:unidentified_issue_decision_date, user: current_user),
-      verifyUnidentifiedIssue: FeatureToggle.enabled?(:verify_unidentified_issue, user: current_user)
+      verifyUnidentifiedIssue: FeatureToggle.enabled?(:verify_unidentified_issue, user: current_user),
+      showFutureRatings: FeatureToggle.enabled?(:show_future_ratings, user: current_user)
```
leikkisa: Same as first comment

client/app/intake/reducers/featureToggles.js
```diff
@@ -16,6 +16,9 @@ const updateFromServerFeatures = (state, featureToggles) => {
     },
     verifyUnidentifiedIssue: {
       $set: Boolean(featureToggles.verifyUnidentifiedIssue)
+    },
+    showFutureRatings: {
```
leikkisa: Same as first comment

client/app/intake/reducers/featureToggles.js
```diff
@@ -26,7 +29,8 @@ export const mapDataToFeatureToggle = (data = { featureToggles: {} }) => (
     editContentionText: false,
     correctClaimReviews: false,
     unidentifiedIssueDecisionDate: false,
-    verifyUnidentifiedIssue: false
+    verifyUnidentifiedIssue: false,
+    showFutureRatings: false
```
leikkisa: Same as first comment

spec/workflows/contestable_issue_generator_spec.rb
```diff
@@ -107,4 +107,15 @@
       end
     end
   end
+
+  describe "#contestable_issues with future ratings" do
```
leikkisa: I think you could put a context within the "#contestable_issues" context, where the feature toggle is enabled, since this is still for testing that same method.

app/workflows/contestable_issue_generator.rb
leikkisa: Yang's right, it's mutating rating_issues.

`issues.reject!` -> `issues = issues.reject` (one example)

spec/workflows/contestable_issue_generator_spec.rb
```diff
@@ -107,4 +107,15 @@
...
+    subject { ContestableIssueGenerator.new(review).from_ratings }
```
leikkisa: This could stay as `.contestable_issues` per the spec name.

app/workflows/contestable_issue_generator.rb
```diff
@@ -14,6 +14,18 @@ def contestable_issues
     contestable_rating_issues + contestable_decision_issues + contestable_rating_decisions
   end
 
+  def from_ratings
```
leikkisa: This can stay as a private method if you adjust the spec subject.

app/workflows/contestable_issue_generator.rb
```diff
@@ -31,9 +31,13 @@ def contestable_rating_decisions
   def from_ratings
     return [] unless receipt_date
 
-    rating_issues
-      .select { |issue| issue.profile_date && issue.profile_date.to_date <= receipt_date }
-      .map { |rating_issue| ContestableIssue.from_rating_issue(rating_issue, review) }
+    issues = rating_issues
+
+    unless FeatureToggle.enabled?(:show_future_ratings, user: RequestStore[:current_user])
+      issues = issues.reject { |issue| issue.profile_date && issue.profile_date.to_date > receipt_date }
```
nanotone: I should have caught this earlier, but I'm now realizing that this PR changes the behavior even if `:show_future_ratings` isn't enabled.

Previously, the `.select` step would ignore any issues without a `profile_date`. Now, even if the toggle isn't on, the new `.reject` step will actually allow issues without a `profile_date`.

I'm not sure how many issues are missing a profile date, or what the desired behavior is in that case, but it might be better to stick with the `.select` style? I defer to @leikkisa here.

leikkisa: There won't be any rating issues without a profile date, because the ratings are keyed by profile date. However, I think it's still a good idea in terms of being safer, and it is also more readable to keep the original logic, especially since the reason for using this logic (using `reject!`) is no longer the case.

The original logic looked like this:
`.select { |issue| issue.profile_date && issue.profile_date.to_date <= receipt_date }`

spec/workflows/contestable_issue_generator_spec.rb
```diff
@@ -106,5 +106,22 @@
...
+        expect(subject.first.description).to eq "Future Rating issue"
+      end
+    end
+
+    context "#contestable_issues with no future ratings" do
+      it "when show_future_ratings feature toggle is not enabled" do
+        expect(subject.count).to eq(3)
+        expect(subject.first.description).to eq "Rating issue"
+      end
+    end
```
leikkisa: Thanks for moving this inside of the #contestable_issues context.

spec/workflows/contestable_issue_generator_spec.rb
```diff
@@ -106,5 +106,22 @@
...
+
+      it "when show_future_ratings feature toggle is enabled" do
+        expect(subject.count).to eq(4)
+        expect(subject.first.description).to eq "Future Rating issue"
+      end
+    end
+
+    context "#contestable_issues with no future ratings" do
+      it "when show_future_ratings feature toggle is not enabled" do
+        expect(subject.count).to eq(3)
```
Sjones352: This test fails if I change this method ```unless FeatureToggle.enabled?(:show_future_ratings, user: RequestStore[:current_user])
     issues.select { |issue| issue.profile_date && issue.profile_date.to_date <= receipt_date }
    end``` because the feature toggle is always set.

leikkisa: @Sjones352 I don't completely understand the feature toggle shouldn't always be set.
Are you still doing `issues = issues.select` ... with your update?

Issue #13198, opened 2020-01-17
Motion to Vacate | Allow AddressMotionToVacate Task to Appear in Judge Queue


jcq:
## Description
By default, while a judge can act on an `AddressMotionToVacateTask` from the case details page, it does not show up in their queue due to existing filtering rules. We need to alter things so that these show up correctly.

## Background/context/resources
Existing Queue code filters this list to just a few specific task types instead of displaying all tasks that are assigned to a judge user. Why?

## Technical notes
- This is being filtered out in the [judgeDecisionReviewTasksSelector](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/selectors.js#L284
) redux code

jcq:
@lomky or @hschallhorn, could one of you shed light upon the existing logic (link in tech notes above)? I can obviously just add `AddressMotionToVacateTask` to the list of allowed tasks in that filter, but I figure it's better to understand the purpose before just adding stuff willy-nilly.

hschallhorn:
😄 This got lost in my emails. This selector exists because we technically have 2 queues for judges, their "assign" queue and their "decision" queue. Their assign page lists only judge assign tasks, both legacy and ama and their decision queue holds all other tasks they can take action on (reviewing and modifying decisions). `judgeDecisionReviewTasksSelector` grabs all tasks that should be shown in this second queue.

jcq:
>`judgeDecisionReviewTasksSelector` grabs all tasks that should be shown in this second queue.

This makes sense. I wonder if it might make sense to refactor things a bit with the goal of being a bit more future proof, but that's not at all critical to do at the moment.

PR #13199, opened 2020-01-18
Allow `AddressMotionToVacateTask` to Appear in Judge Queue

jcq:
Connects #13198

### Description
Updated whitelist in `judgeDecisionReviewTasksSelector` to allow `AddressMotionToVacateTask` to appear in judge queue.

### Acceptance Criteria
- [ ] `AddressMotionToVacateTask` shows up in queue for assigned judge

### DIFF
--- a/client/app/queue/selectors.js
+++ b/client/app/queue/selectors.js
@@ -291,7 +291,8 @@ export const judgeDecisionReviewTasksSelector = createSelector(
     // eslint-disable-next-line no-undefined
     return [null, undefined, COPY.JUDGE_DECISION_REVIEW_TASK_LABEL,
       COPY.JUDGE_QUALITY_REVIEW_TASK_LABEL,
-      COPY.JUDGE_DISPATCH_RETURN_TASK_LABEL].includes(task.label);
+      COPY.JUDGE_DISPATCH_RETURN_TASK_LABEL,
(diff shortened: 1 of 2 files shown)


client/app/queue/selectors.js
```diff
@@ -283,14 +283,16 @@ export const judgeDecisionReviewTasksSelector = createSelector(
...
+        COPY.JUDGE_ADDRESS_MOTION_TO_VACATE_TASK_LABEL].includes(task.label)) &&
         (task.status === TASK_STATUSES.in_progress || task.status === TASK_STATUSES.assigned);
     }
 
     // eslint-disable-next-line no-undefined
     return [null, undefined, COPY.JUDGE_DECISION_REVIEW_TASK_LABEL,
       COPY.JUDGE_QUALITY_REVIEW_TASK_LABEL,
-      COPY.JUDGE_DISPATCH_RETURN_TASK_LABEL].includes(task.label);
+      COPY.JUDGE_DISPATCH_RETURN_TASK_LABEL,
+      COPY.JUDGE_ADDRESS_MOTION_TO_VACATE_TASK_LABEL].includes(task.label);
```
nanotone: If these two lists (other than the null and undefined) are going to stay identical for the foreseeable future, might there be value in DRYing this up slightly, into a shared constant?

jcq: Sorry, only saw this after it had merged. That's a good suggestion, though I also wonder if it might make sense to do additional refactoring to make this structure more future proof.

PR #13204, opened 2020-01-21
Realistic name seeds

lomky:
### Description
Seed data updated to give test data users with permissions or test data realistic names. 

Before: 
`Role Description`
e.g. `Judge with hearings and cases`

After:
`GivenName RoleDescription FamilyName`
e.g. `Aaron JudgeWithHearingsAndCases Abshire`

### Acceptance Criteria
- [ ] Code still works
- [ ] Other team members like the change

### Testing Plan
1. Go to [`/test/users`](http://localhost:3000/test/users), look over the names. 
1. Browse as various users and see if you prefer the UI
   - Attorney list on judge Assign
   - Judge list on reassign to judge
   - Names on case details page
   - Names on hearings schedule & details
   - Other locations too, probably? Everywhere we show users names!

### User Facing Changes
Dev only


leikkisa:
This is a fun PR. I like this idea. I haven't looked at it in the UI yet, but have a preliminary comment that I think just a first name would be sufficient (if it's for our memory or for communicating with each other). I guess the "role" would kind of be a last name in that case.

wturne:
I like the idea of the first name with the role as the last name.  I think this is a great idea.

lomky:
@wturne here are some selected screen shots of what the Middle Name setup looks like. In particular how it shows in the dropdowns (helping us find via keyword), but looks like a natural first/last in the hearings.


Old | New
------------ | -------------
Old - User Selector | New - User Selector
Old - Assign | New - Assign
Old - Case Details | New - Case Details
Old - Hearing Sched | New - Hearing Sched
Old - Docket | New - Docket

PR #13209, opened 2020-01-21
ETL calculate AOD based on age every day

pkarman:
### Description
Optimizes the AOD calculation based on age by storing it as a column.

### Database Changes
*Only for Schema Changes*

* [x] Column comments updated

### DIFF
--- a/app/services/etl/builder.rb
+++ b/app/services/etl/builder.rb
@@ -19,23 +19,24 @@ class ETL::Builder
 
   def initialize(since: checkpoint_time)
     @since = since
+    @built = 0
   end
 
-  attr_reader :since
+  attr_reader :since, :built
 
   def incremental
-    built = 0
     checkmark
     syncer_klasses.each do |klass|
-      built += klass.new(since: since).call
+      self.built += klass.new(since: since).call
     end
+    post_build_steps
     built
   end
 
   def full
-    built = 0
     checkmark
-    syncer_klasses.each { |klass| built += klass.new.call }
+    syncer_klasses.each { |klass| self.built += klass.new.call }
+    post_build_steps
     built
   end
 
@@ -44,6 +45,8 @@ class ETL::Builder
   end
 
   private
+
+  attr_writer :built
 
   def syncer_klasses
     ETL_KLASSES.map { |klass| "ETL::#{klass}Syncer".constantize }
@@ -64,5 +67,13 @@ class ETL::Builder
 
   def checkmark
     Rails.cache.write(CHECKPOINT_KEY, Time.zone.now.to_s)
+  end
+
+  def post_build_steps
+    mark_aod_for_today
+  end
+
+  def mark_aod_for_today
+    ETL::Appeal.where("claimant_dob <= ?", 75.years.ago).where(aod_due_to_dob: false).update_all(aod_due_to_dob: true)
   end
 end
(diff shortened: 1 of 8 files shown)


app/services/etl/builder.rb
```diff
@@ -64,4 +67,12 @@ def checkpoint
   def checkmark
     Rails.cache.write(CHECKPOINT_KEY, Time.zone.now.to_s)
   end
+
+  def post_build_steps
+    mark_aod_for_today
+  end
+
+  def mark_aod_for_today
+    ETL::Appeal.where("claimant_dob <= ?", 75.years.ago).where(aod_due_to_dob: false).update_all(aod_due_to_dob: true)
```
edmangimelli: in `person.rb`
`< 75` is used instead of `<= 75`:
```
  def advanced_on_docket_based_on_age?
    date_of_birth && date_of_birth < 75.years.ago
  end
```
is that intentional?

pkarman: good catch and good question.

Based on my experiments, I don't think it matters for Ruby, because `75.years.ago` returns a `DateTime` but `date_of_birth` is a `Date`.

```
[2] pry(main)> dob = 75.years.ago.to_date
=> Tue, 23 Jan 1945
[3] pry(main)> dob.to_s
=> "1945-01-23"
[4] pry(main)> dob < 75.years.ago
=> true
[5] pry(main)> dob <= 75.years.ago
=> true
[10] pry(main)> dob < (Time.zone.tomorrow - 75.years)
=> true
[11] pry(main)> dob <= (Time.zone.tomorrow - 75.years)
=> true
[12] pry(main)> dob < (Time.zone.yesterday - 75.years)
=> false
[13] pry(main)> dob <= (Time.zone.yesterday - 75.years)
=> false
```

However for SQL, it does matter:

```
[15] pry(main)> Person.where("date_of_birth < ?", DateTime.parse("1920-01-01"))
[2020-01-23 09:23:28 -0600]   Person Load (1.6ms)  SELECT "people".* FROM "people" WHERE (date_of_birth < '1920-01-01 00:00:00')
=> []
[16] pry(main)> Person.where("date_of_birth <= ?", DateTime.parse("1920-01-01"))
[2020-01-23 09:23:33 -0600]   Person Load (1.4ms)  SELECT "people".* FROM "people" WHERE (date_of_birth <= '1920-01-01 00:00:00')
=> [#<Person:0x00007fe437af1018
  id: 340,
  created_at: Tue, 14 Jan 2020 20:49:22 UTC +00:00,
  date_of_birth: Thu, 01 Jan 1920,
  first_name: "Tom",
  last_name: "Brady",
  middle_name: "Edward",
  name_suffix: nil,
  participant_id: "673564",
  updated_at: Tue, 14 Jan 2020 20:51:33 UTC +00:00,
  email_address: "tom.brady@caseflow.gov">]
```

So yes, it is intentional.

edmangimelli: interesting
so this:
```
dob < 75.years.ago
```
probably evaluates to something like this:
```
Tue, 23 Jan 1945, 00:00 < Tue, 23 Jan 1945, 09:24
```
makes sense

so, in the Ruby comparison, you have
```
Date < DateTime
```
becomes
```
DateTime < DateTime
```

whereas, in the SQL the passed in DateTime is turning into a Date

Thanks for explaining!

pkarman:
Add "active-only" qualifier for appeals. Don't toggle aod-on-dob for inactive appeals.

PR #13223, opened 2020-01-22
Correcting the Timezone Display

### LINKED ISSUE #12768: VEText timezone should be of the hearing location
## User or job story
User story: As a Veteran, the text message should indicate the timezone of my hearing location, not my regional office.

Current text message for a Veteran that lives in MST, but the regional office is CST.  He was scheduled for an alternate regional office that is in MST. 
Caseflow sent the following message:
You have an upcoming Board of Veterans' Appeals hearing with a Veterans Law Judge at the Fort Meade VA Medical Center at 113 Comanche Road, Room 3, Fort Meade, SD, 57741-1002 on Nov 15, 2019 at 8:30 AM CST. 

This data is technically correct but since the Veteran's hearing is in MST, it would make more sense to show 7:30 MST.

- [ ] Please test this by investigating the output for a few cities where the regional office will be in a different timezone than the hearing location.  

- [ ] This feature should be accessible to the following user groups:
- [ ] Include screenshot(s) in the Github issue if there are front-end changes

### Background/context
This will make the message clearer for the Veteran.

aplybeah:
Resolves #12768 

### Description
- Added a `timezone` field to the `REGIONAL_OFFICE_FACILITY_ADDRESS.json`
- updated the `hearing_location` model to have a `timezone` field, which pulls the relevant information from the updated json
- Updated the `hearings_controller_spec` to test the following scenarios:
                  - A hearing in the central office
                  - A hearing not held on the east coast
                  - A hearing with a hearing location that doesn't match the original regional office
- made sure the `facility_id` in the test factory was saved for a more accurate representation of data.


spec/factories/hearing_location.rb
```diff
@@ -11,6 +11,7 @@
         ro = evaluator.regional_office
         facility_id = RegionalOffice.facility_ids_for_ro(ro).first || "VACO"
         facility_address = Constants::REGIONAL_OFFICE_FACILITY_ADDRESS[facility_id]
+        hearing_location.facility_id = facility_id
```
ferristseng: nice catch!

Issue #12697, opened 2019-11-13
Data Lake


pkarman:
## Background/context

Caseflow must provide its data to the BVA Data Warehouse (live on demand storage) and Data Lake (persistent long term snapshot storage).

The BVA data lake is an S3 bucket where Caseflow needs to deposit CSV files of raw db data.

See [Data Warehouse Epic](https://github.com/department-of-veterans-affairs/caseflow/issues/12643)

## Goals

Raw db deposits should reflect point-in-time snapshots of Caseflow production data, with enough metadata to support exploration and discovery.

## Success criteria

Data Lake is daily populated with Caseflow data.

## Stakeholders
* BVA
* B3

pkarman:
Infra ticket: https://github.com/department-of-veterans-affairs/appeals-deployment/issues/2316

PR #13226, opened 2020-01-22
CsvToS3Exporter service

pkarman:
connects #12697 

### Description
Service class to:

* export a single postgresql table to a csv file,
* verify its contents,
* write it to S3 bucket

### Testing Plan
1. In UAT, set AWS credentials as env vars for the appropriate Data Lake bucket. Credentials are currently in 1Password.
1. In UAT, run the exporter for a single table from the console.

### Documentation Updates

### Example

From Rails console:

```
[1] pry(main)> e = CsvToS3Exporter.new(test: "/tmp/appeals.csv", table: 'appeals', bucket: 'foo')
=> #<CsvToS3Exporter:0x00007feaabe4aa28
 @bucket="foo",
 @compress=true,
 @date="2020-01-22",
 @region="us-gov-west-1",
 @table="appeals",
 @test="/tmp/appeals.csv">
[2] pry(main)> e.call
psql $POSTGRES_URL -c '\copy appeals to /var/folders/vy/cpsrccmx7x3g1j9vv8rxm58m0000gn/T/2020-01-22-caseflow-appeals.csv20200122-16147-1npg7s2 with (format csv, header, force_quote *)'
wc -l /var/folders/vy/cpsrccmx7x3g1j9vv8rxm58m0000gn/T/2020-01-22-caseflow-appeals.csv20200122-16147-1npg7s2
sha256sum /var/folders/vy/cpsrccmx7x3g1j9vv8rxm58m0000gn/T/2020-01-22-caseflow-appeals.csv20200122-16147-1npg7s2
cp /var/folders/vy/cpsrccmx7x3g1j9vv8rxm58m0000gn/T/2020-01-22-caseflow-appeals.csv20200122-16147-1npg7s2 /tmp/appeals.csv
echo '{"rows":278,"time":"2020-01-22T23:08:56Z","checksum":"5d5eb2685df9829ea3e94c4389a1fa1da67501381f0e27f4ad613250ac6e0343"}' > /tmp/appeals.csv.meta
=> {:rows=>278, :time=>"2020-01-22T23:08:56Z", :checksum=>"5d5eb2685df9829ea3e94c4389a1fa1da67501381f0e27f4ad613250ac6e0343"}
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,6 +99,17 @@ jobs:
             ~/project/ci-bin/capture-log "bundle exec rake spec:setup_vacols"
 
       - run:
+          name: Install dependencies
+          command: |
+            wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo apt-key add -
+            sudo sh -c 'echo "deb http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google-chrome.list'
+            sudo apt-get update
+            sudo apt-get install google-chrome-stable
+            bundle exec rake webdrivers:chromedriver:remove
+            bundle exec rake webdrivers:chromedriver:update
+            sudo apt-get install postgresql-client
+
+      - run:
           name: RSpec
           command: |
             mkdir -p ~/test-results/rspec
--- a/spec/feature/reader/reader_spec.rb
+++ b/spec/feature/reader/reader_spec.rb
@@ -496,7 +496,7 @@ RSpec.feature "Reader", :all_dbs do
       add_comment("A")
 
       click_on "Edit"
-      find("#editCommentBox-2").send_keys(:backspace)
+      find("#editCommentBox-2").send_keys(:delete)
       click_on "Save"
 
       # Delete modal should appear
--- a/.reek.yml
+++ b/.reek.yml
@@ -11,6 +11,7 @@ detectors:
   BooleanParameter:
     exclude:
       - AsyncableJobsReporter#initialize
+      - CsvToS3Exporter#initialize
       - HearingRepository#slot_new_hearing
       - LegacyDocket#distribute_appeals
       - Veteran
@@ -86,6 +87,7 @@ detectors:
   LongParameterList:
     exclude:
       - Address#initialize
+      - CsvToS3Exporter#initialize
       - ExternalApi::VBMSService#self.create_contentions!
       - LegacyDocket#new_distributed_case
       - Api::V3::DecisionReview::ContestableIssueFinder#initialize
@@ -117,6 +119,7 @@ detectors:
   TooManyInstanceVariables:
     exclude:
       - AmaAppealDispatch
+      - CsvToS3Exporter
       - DuplicateUserTask
       - LegacyAppealDispatch
       - Address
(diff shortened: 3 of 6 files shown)


spec/feature/reader/reader_spec.rb
```diff
@@ -496,7 +496,7 @@ def expect_dropdown_filter_to_be_visible
       add_comment("A")
 
       click_on "Edit"
-      find("#editCommentBox-2").send_keys(:backspace)
+      find("#editCommentBox-2").send_keys(:delete)
```
pkarman: related to the chrome driver update at CircleCI

.circleci/config.yml
```diff
@@ -98,6 +98,17 @@ jobs:
           command: |
             ~/project/ci-bin/capture-log "bundle exec rake spec:setup_vacols"
 
+      - run:
+          name: Install dependencies
```
pkarman: I copied over this CI step from efolder, where we had already updated the chrome driver. Now we have parity.

I added the postgresql-client to caseflow only since it is required for this PR.

enriquemanuel:
This is pretty cool and impressive and solves the need, but I would love to see this as a start of migrating the monolithic application that Caseflow is to a Microservices example that other engineers can use.

We can use the `appeals-lambdas` and invoke it from a cron. The code can be in any language and used in any future and it can read from a SQS.

Why I say this?
Because it would be nice that this could potentially be open source; then we could start creating tests to our lambdas, start creating new components outside the big main app and make it simpler to deploy and test across different areas.

Finally, I believe that this is a core example of something that can be used outside of the core application.

I know there are very good reasons to keep it inside the code, but right now we do have some QA and CI problems because our application is a large one, so at least in my head, create new functionality outside the core application is a good idea.

Open for thoughts.

pkarman:
In order to move this functionality outside this repo, we would need infrastructure to support:

* setup/teardown a test database the `psql` command can read from,
* automated testing via CI

AFAIK, the `appeals-lambdas` repo does not currently offer either of those.

More broadly:

> a start of migrating the monolithic application that Caseflow is to a Microservices example

Let me state categorically that there is no intention of migrating Caseflow to a microservices pattern. It might be possible to carve off small pieces of this app if they do not require access to the database models. This particular task (export a table as CSV to an S3 bucket) is a fine candidate for that, assuming the needs above are addressed. But broadly speaking, we have no plans to break up the monolith to use microservices.

enriquemanuel:
> Let me state categorically that there is no intention of migrating Caseflow to a microservices pattern. It might be possible to carve off small pieces of this app if they do not require access to the database models. This particular task (export a table as CSV to an S3 bucket) is a fine candidate for that, assuming the needs above are addressed. But broadly speaking, we have no plans to break up the monolith to use microservices.

I completely agree that we **should not** migrate the app; what I tried to mention is that for new features, like this one, they are prime candidates to not overload the application and start creating side services (or micro services) that can be invoked from within the app or by a placing items on a queue.

We can, and its the intent that we should create create a CI for the Appeals Lambdas, and we are headed that way.

My only thought about this is that we should start creating those services so our application that is big, doesn't overgrow us and then might be too difficult to handle and/or test as well takes longer to deploy.

alisan16:
> In order to move this functionality outside this repo, we would need infrastructure to support:
> - setup/teardown a test database the psql command can read from,

What would it take to set this up? I agree with both of you that this is a fine candidate for modeling a new microservices pattern in Caseflow. Since this task is related to the Data Lake, there shouldn't be a rush to get this in the "easy" way via bloating Caseflow with another tangential Rails class.

Issue #12947, opened 2019-12-12
Add "Job note with message" method


leikkisa:
## Description
As an engineer, I often like to add job notes from the console when reviewing stuck claims (instead of in the Caseflow UI).  Adding job notes from here doesn't add the Inbox message though.  It'd be good to make this available through a non-controller method, such as in the JobNote model or a workflow.

## Technical notes
See the AsyncableJobController add_note for current code on how to add a job note with an inbox message.

leikkisa:
Open to conversation on whether we should do this if there's no place we're calling it in the code (I can update my personal script). 

This can potentially be used in the future for handling known upstream errors, I think that depends on whether we need to individually add cases to YourIT tickets, and it'd be good to ask Peter's opinion.

msj0nes:
@nanotone Is this issue fully complete? If so please close the ticket.

PR #13235, opened 2020-01-23
Default AsyncableJobMessaging user to RequestStore current user

nanotone:
Connects #12947 

### Description
This PR every-so-slightly tidies up the way AsyncableJobMessaging detects the current user, to be in line with the rest of the codebase.

When I first wrote that class, I didn't realize that setting `RequestStore[:current_user]` was common practice for folks working in the Rails console. Now, instead of being forced to provide a user object, the engineer can do either

```
> ajm = AsyncableJobMessaging.new(job: some_job)
> ajm.add_job_note(text: "the note", send_to_intake_user: true)
```

or if they want to provide a user object other than what's in RequestStore,

```
> ajm = AsyncableJobMessaging.new(job: some_job, user: someone_else)
> ajm.add_job_note(text: "the note", send_to_intake_user: true)
```

### DIFF
--- a/app/models/claim_review.rb
+++ b/app/models/claim_review.rb
@@ -121,12 +121,12 @@ class ClaimReview < DecisionReview
   # Cancel an unprocessed job, add a job note, and send a message to the job user's inbox.
   # Currently this only happens manually by an engineer, and requires a message explaining
   # why the job was cancelled and describing any additional action necessary.
-  def cancel_with_note!(current_user:, note:)
-    fail Caseflow::Error::ActionForbiddenError, message: "Acting user must be specified" unless current_user
+  def cancel_with_note!(user: RequestStore[:current_user], note:)
+    fail Caseflow::Error::ActionForbiddenError, message: "Acting user must be specified" unless user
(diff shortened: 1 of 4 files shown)


nanotone:
@leikkisa: Note that this doesn't directly address the connected issue. I think the issue could be closed anyway, now that I'm reminded that AsyncableJobsMessaging already does JobNote+Message adding in a single method (see PR description above).

Would you still prefer a new method to be added to one or more models, for ergonomic reasons?

app/models/claim_review.rb
```diff
@@ -121,12 +121,12 @@ def update_error!(err)
   # Cancel an unprocessed job, add a job note, and send a message to the job user's inbox.
   # Currently this only happens manually by an engineer, and requires a message explaining
   # why the job was cancelled and describing any additional action necessary.
-  def cancel_with_note!(current_user:, note:)
+  def cancel_with_note!(current_user: RequestStore[:current_user], note:)
```
edmangimelli: just out of curiosity, why not
`def cancel_with_note!(user: RequestStore[:current_user], note:)`
?

nanotone: Good catch! Updated.

Issue #13211, opened 2020-01-22
Motion to Vacate | Granted | Support multiple AMA appeal streams


leikkisa:
## User or job story
At the board, multiple appeal streams need to be supported in order to support post-decision motions.

## Acceptance criteria
**Docket Number**
- [ ] Create a new field on Appeal, called docket_number
- [ ] Populate docket_number when creating an appeal (at the same time as populating the receipt date), with the current format `"#{receipt_date.strftime('%y%m%d')}-#{id}"`
- [ ] Backfill docket_number for existing appeals
- [ ] Use the new field docket_number, instead of the method
- [ ] Remove the docket_number method in the Appeal model

**Type**
- [ ] Create a new field on Appeal, called type
- [ ] When creating a new appeal from Intake, populate "type" with "Original"
- [ ] Backfill "type" for existing appeals
- [ ] Use new field instead of the method
- [ ] Remove "type" method in Appeal model

### Background/context

Appeal streams indicate work on the same docket number, but happening on different appeals.  This will mimic the structure being used for legacy appeals.  New appeal streams will have a different docket type, depending on the work being done on that appeal.

To see an example of a Legacy Appeal with multiple docket types under the same docket number, see Legacy Appeals with the following vacols_ids:
["3017739", "3703912", "3726007", "3726018", "3996075"]

You can look at the `docket_number` and `type` for these legacy appeals.

### Technical notes
This can be done in multiple PRs
1. Add the new DB fields and backfill them
1. Switch to using the new fields instead of the methods

hschallhorn:
Related! #11678

nanotone:
I'm going to go ahead and call this issue done for the sake of sprint planning; the one remaining open PR is blocked externally, isn't very relevant to upcoming MTV work, and is outside of what we expected with the original 3-point estimate.

PR #13241, opened 2020-01-23
Add docket_number and stream_type columns to Appeals table

nanotone:
Connects #13211 

### Description
This is the first of multiple PRs for adding (MTV-required) support for multiple appeal streams with the same docket number. Previously, each appeal had a unique docket number and a type of `Original`. Going forward, the MTV flow will create a new appeal with the same docket number as the original, but with type `Vacate`.

This PR just adds the two DB columns and updates the `Appeal` model to use the values of those columns when not nil, or fall back on the existing behavior of hard-coded ephemeral values.

Future work:
- Update existing appeal-creation code to persist values to DB
- Backfill these two columns for all existing appeals
- After backfilling, remove the logic that supplies default values when DB values are nil

Note:
ActiveRecord is not happy about a DB column named `type`, so I named it `stream_type` (`docket_type` is used for a different purpose). We may want to consider changing the Appeal model's method name anyway, for the sake of consistency and because "type" is quite generic.

### Acceptance Criteria
- [ ] Code compiles correctly
- [ ] `appeals` table contains new columns
- [ ] `Appeal#docket_number` and `#type` use existing behavior when new columns are nil
- [ ] `Appeal#docket_number` and `#type` use values in new columns when they're not nil

### Testing Plan
1. Tested in Rails console:
```
> a = Appeal.first
> a.docket_number
=> "191208-1"
> a.type
=> "Original"
> a = Appeal.new(docket_number: "foo", stream_type: "Vacate")  # or use psql to insert
> a.docket_number
=> "foo"
> a.type
=> "Vacate"
```

### Database Changes
*Only for Schema Changes*

* [x] Column comments updated

### DIFF
--- a/db/migrate/20200123200907_add_multiple_appeal_streams.rb
+++ b/db/migrate/20200123200907_add_multiple_appeal_streams.rb
@@ -1,0 +1,6 @@
+class AddMultipleAppealStreams < ActiveRecord::Migration[5.1]
+  def change
+    add_column :appeals, :docket_number, :string, comment: "Multiple appeals with the same docket number indicate separate appeal streams, mimicking the structure of legacy appeals."
+    add_column :appeals, :stream_type, :string, comment: "When multiple appeals have the same docket number, they are differentiated by appeal stream type, depending on the work being done on each appeal."
+  end
+end
--- a/db/migrate/20200123200907_add_multiple_appeal_streams.rb
+++ b/db/migrate/20200123200907_add_multiple_appeal_streams.rb
@@ -1,6 +1,6 @@
 class AddMultipleAppealStreams < ActiveRecord::Migration[5.1]
   def change
-    add_column :appeals, :docket_number, :string, comment: "Multiple appeals with the same docket number indicate separate appeal streams, mimicking the structure of legacy appeals."
+    add_column :appeals, :stream_docket_number, :string, comment: "Multiple appeals with the same docket number indicate separate appeal streams, mimicking the structure of legacy appeals."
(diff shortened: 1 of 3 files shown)


nanotone:
Updated to make `stream_type` an enum, as per @pkarman's suggestion.

nanotone:
@pkarman This is ready :bowtie: for another look. The two new DB fields are `stream_docket_number` and `stream_type`, which should not affect the SQL tests now.

Question: Do I need the ETL DB migration at all?

db/migrate/20200123200907_add_multiple_appeal_streams.rb
```diff
@@ -0,0 +1,6 @@
+class AddMultipleAppealStreams < ActiveRecord::Migration[5.1]
+  def change
+    add_column :appeals, :stream_docket_number, :string, comment: "Multiple appeals with the same docket number indicate separate appeal streams, mimicking the structure of legacy appeals."
```
leikkisa: I like this approach a lot overall.

Non-blocking comment, but the name `stream_docket_number` implies that each stream might have a different docket number, which is kind of the opposite.  I don't have a better suggestion though (besides an inane comment that multiple streams combined form a river). I'd be reluctant to use ama_ as a prefix.

nanotone: I definitely get your concern about naming, too. One silver lining is that these are largely confined to the storage layer for now, and hopefully my DB column comments are helpful on that front. In Ruby land, `docket_number` and `type` will still be the preferred methods.

Once the dust settles on this whole thing (since we have to be extra careful on order of operations so as not to break things ETL-side), we can revisit the issue.

PR #13258, opened 2020-01-25
Persist stream docket number and type to new database columns

nanotone:
Connects #13211 

### Description
Second in a series of PRs (first: #13241) to support appeal streams by storing the docket number and type in the DB.

This PR adds a `before_save` hook to Appeal which persists accurate values for docket number and type to the DB. After this PR, the DB fields will be populated for new and touched appeals, including MTV appeal when they land.

N.B.: with docket number, there's a tension in the fact that a non-nil stored value overrides the computed value, which in turn overrides a nil stored value — possible if the Appeal is new or only recently got its `id` or `receipt_date` — and this has some subtle order-of-operations implications. 

I've added a test for edge cases, even though the happy path for appeal creation is pretty well-behaved. Other tests are somewhat less well-behaved.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Added a model test, to stress-test the order-of-operations subtleties mentioned above.

### DIFF
--- a/app/models/appeal.rb
+++ b/app/models/appeal.rb
@@ -30,6 +30,8 @@ class Appeal < DecisionReview
     "De Novo": "De Novo"
   }
 
+  before_save :set_stream_fields
+
   with_options on: :intake_review do
     validates :receipt_date, :docket_type, presence: { message: "blank" }
     validate :validate_receipt_date
@@ -297,7 +299,7 @@ class Appeal < DecisionReview
 
   def docket_number
     return stream_docket_number if stream_docket_number
-    return "Missing Docket Number" unless receipt_date
+    return "Missing Docket Number" unless receipt_date && id.present?
 
     "#{receipt_date.strftime('%y%m%d')}-#{id}"
   end
@@ -422,6 +424,13 @@ class Appeal < DecisionReview
   end
 
   private
+
+  def set_stream_fields
+    if receipt_date && id.present?
+      self.stream_docket_number ||= docket_number
+    end
+    self.stream_type ||= type
+  end
 
   def maybe_create_translation_task
     veteran_state_code = veteran&.state
--- a/app/models/appeal.rb
+++ b/app/models/appeal.rb
@@ -30,7 +30,7 @@ class Appeal < DecisionReview
     "De Novo": "De Novo"
   }
 
-  before_save :set_stream_fields
+  before_save :set_stream_docket_number_and_stream_type
 
   with_options on: :intake_review do
     validates :receipt_date, :docket_type, presence: { message: "blank" }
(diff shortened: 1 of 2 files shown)


app/models/appeal.rb
```diff
@@ -297,7 +299,7 @@ def create_issues!(new_issues)
 
   def docket_number
     return stream_docket_number if stream_docket_number
-    return "Missing Docket Number" unless receipt_date
+    return "Missing Docket Number" unless receipt_date && id.present?
```
pkarman: I think `.persisted?` is the built-in way to do `id.present?`

spec/models/appeal_spec.rb
```diff
@@ -302,10 +302,10 @@
   context "#docket_number" do
     context "when receipt_date is defined" do
       let(:appeal) do
-        create(:appeal, receipt_date: Time.new("2018", "04", "05").utc)
+        create(:appeal, id: 1234, receipt_date: Time.new("2018", "04", "05").utc)
```
pkarman: should not need to explicitly add an id here, if you are using `.persisted?` -- I think.

nanotone:
@pkarman You're absolutely right on both comments. Pushed an update.

app/models/appeal.rb
```diff
@@ -423,6 +425,13 @@ def eligible_for_death_dismissal?(_user)
 
   private
 
+  def set_stream_fields
```
tomas-nava: I would prefer a more verbose method name, like `set_stream_docket_number_and_type`

app/models/appeal.rb
```diff
@@ -423,6 +425,13 @@ def eligible_for_death_dismissal?(_user)
...
+    if receipt_date && persisted?
+      self.stream_docket_number ||= docket_number
```
tomas-nava: Is it possible that the docket number will get stuck as `"Missing Docket Number"`? Should we check for that value and re-generate the ID if it's saved in the db?

app/models/appeal.rb
nanotone: It's a bit more verbose than my personal preference, but I'm okay with it here because this method is not really intended to be called by anybody other than the model hook.

app/models/appeal.rb
nanotone: The `receipt_date && persisted?` check should protect against `Missing Docket Number` ever making it into the DB. It's intentionally the same check as in `#docket_number`, but I realize that this can read more like a coincidence. Do you think it's worth preserving in a separate `has_docket_number?` method?

tomas-nava: Ok, I see now. Nah, it's fine as is.

PR #13260, opened 2020-01-26
[WIP] Test for DB cleanup issues

monfresh:
Trying out a script to detect DB cleanup issues.


monfresh:
Looks like PRs from forks can't run because of the AWS credentials needed to download the Docker images. It would be great if the Docker images can be open sourced. Any idea if that's possible, @aroltsch or @enriquemanuel ?

PR #13270, opened 2020-01-27
Alsia 13219 lock time buttons

### LINKED ISSUE #13219: Time of virtual hearing is editable after hearing date
### Summary
I noticed in UAT that I'm still able to (1) edit the hearing time and (2) send emails _after_ the date on which the hearing occurred. I would expect that the hearing time would not be editable after the hearing date. (If the hearing was postponed and rescheduled, and therefore the hearing date was in the future, I would expect the time to be editable again.)

### Observed
_(screenshots from UAT)_

1. Log in to UAT as CF_RIKER_317 (a Hearing Coordinator user)
1. Go to the Hearing Schedule overview page and select a Daily Docket with a virtual hearing on a date in the past
1. Edit the hearing time and save.
    
1. See that the blue info alert displays.
    
1. See that emails are sent.
    

### Expected
Hearing time is _not_ editable (field is disabled) for virtual hearing dates that are in the past.

aplybeah:
Resolves #13219 

### Description
- added a boolean attribute `scheduled_for_is_past?` which compares the current date and the date in `scheduled_for`.
- changed the conditions for `readOnly` in the Daily Docket Row component
- Time Radio buttons are locked for both legacy and ama hearings when the date of the hearing has passed.

#### Past Hearings
_AMA_
Screen Shot 2020-01-27 at 12 11 41 PM

_Legacy_
Screen Shot 2020-01-27 at 12 21 00 PM

---

#### Future Hearings
_AMA_
Screen Shot 2020-01-27 at 12 19 22 PM

_Legacy_
Screen Shot 2020-01-27 at 12 14 31 PM

There is still a discrepancy between the displayed `<StaticHearingDay>` and the actual day that appears on the docket. However, in the console the correct date and time are returned:
Screen Shot 2020-01-27 at 12 16 28 PM


app/models/legacy_hearing.rb
```diff
@@ -184,6 +184,11 @@ def scheduled_pending?
     scheduled_for && !closed?
   end
 
+  # pulls scheduled_for from hearing day instead of VACOLS
+  def scheduled_for_past?
+    hearing_day.scheduled_for < DateTime.now.in_time_zone(regional_office_timezone)
```
ferristseng: I'm wondering if we want to just changed `LegacyHearing::scheduled_for` to pull from the hearing day consistently, or if there's better logic to determine whether or not to use the scheduled for date from VACOLS vs Caseflow. Maybe that can be a follow up tech debt PR?

aplybeah:
@ferristseng  From what I was able to see, in Legacy Hearings the time buttons lock the day after the hearing was held and AMA locks immediately. I spoke with Nichole and she says the scenario where hearing times lock the day after is the better one

aplybeah:
Just a note for the future: `scheduled_for_past?` in both legacy and ama hearing models originally had `DateTime.now........`. However, when the actual day of the hearing matches the `scheduled_for` or `hearing_day.scheduled_for` attributes the time buttons would lock. This caused a lot of test regressions because the tests (for the most part) assume that `scheduled_for` and `hearing_day` are the same. Changing to `DateTime.yesterday.......` resolves this and also gives extra time to fiddle with the hearing time before the buttons get locked.

PR #13280, opened 2020-01-28
Backfill docket number and type into new stream fields, via migration

nanotone:
Connects #13211 

### Description
Third in a series of PRs (previously: #13241 #13258) to support appeal streams by storing the docket number and type in the DB.

This PR backfills the `stream_docket_number` and `stream_type` fields for existing appeals. Since every existing appeal has type "Original" and an intrinsic docket number (derived from its own receipt date and Rails ID), this can be done with a migration rather than ActiveRecord queries.

As a backfill script, the migration is rollback-less; the two new fields are currently not being used anyway. After the backfill, all appeals present and future will have populated stream fields.

### Testing Plan
1. Tested on local Postgres, starting with Appeal records with every combination of {receipt_date, stream_docket_number, stream_type} presence.

### Database Changes
This PR uses a migration but does not change the database schema.


db/migrate/20200128153700_populate_stream_fields.rb
```diff
@@ -0,0 +1,17 @@
+class PopulateStreamFields < ActiveRecord::Migration[5.1]
+  def up
+    safety_assured do
+      execute "UPDATE appeals SET stream_type='Original' WHERE stream_type IS NULL"
```
pkarman: do you want to also `change_default` here after the update?

db/migrate/20200128153700_populate_stream_fields.rb
```diff
@@ -0,0 +1,17 @@
...
+      execute <<-EOS.strip_heredoc
+        UPDATE appeals
+          SET stream_docket_number=to_char(receipt_date, 'YYMMDD-') || id
```
pkarman: I thought the default docket number was concatenate receipt_date and id? Or is the `||` syntax a fancy shorthand for that?

nanotone: I got the `||` syntax from our very own SQL code for simulating docket numbers. Turns out it's standard stuff: https://www.postgresql.org/docs/9.4/functions-string.html

pkarman: TIL!

db/migrate/20200128153700_populate_stream_fields.rb
nanotone: Good idea, I'll set the default for `stream_type` to `Original`.

PR #13141, opened 2020-01-13
Implement Pagination for Assign Hearings

ferristseng:
Resolves #11757

### Description
This PR implements pagination for assign hearings table by taking existing pagination setup for queue tables and organizing and fitting hearings data around it. Idea is to paginate the Ama assign hearings table and legacy assign hearings table and ensure that `Suggested Location` and `Power of Attorney` columns are filterable.

The changes include frontend refactoring and introducing new backend endpoints. This [PR](https://github.com/department-of-veterans-affairs/caseflow/pull/13285) adds two new fields: `suggestedHearingLocation` and `powerOfAttorneyName` to the `CachedAppeals` table. We join `CachedAppeals` table with `ScheduleHearingTasks` by `closest_regional_office_key` to returns rows of tasks for the assign hearings table either for `amaAssignHearingsTab` or `legacyAssignHearingsTab.`

#### Backend Changes 🔙 

- Remove `tasks#ready_for_hearing_schedule` endpoint from `TasksController`. `ScheduleHearingTasks` in the `AssignHearingsTable` will no longer be populated through this endpoint.
- Create `ScheduleHearingTasksController` endpoint for retrieving tasks for given page as well returning any filtered rows.  
- Create `ScheduleHearingTasksColumnsController` endpoint for returning `filter_options` for `Suggested Location` and `Power of Attorney` column. `filter_options` is calculated in the backend and returned to `AssignHearingsTable` component in the frontend. 
- Remove `HearingCoordinatorScheduleQueue` and move functionalities to `AssignHearing`
- Create `AssignHearing` class which mimics a queue_tab and acts as `amaAssignHearingsTab` and `legacyAssignHearingsTab` but also takes the functionalities of `HearingCoordinatorScheduleQueue`. It returns tasks by joining `ScheduleHearingTasks` with `CachedAppeal` table by `regional_office_key`. It also calculates the `filter_options` for `Suggested Location` and `Power Of Attorney` columns. 
- Create `ScheduleHearingTaskPager` class which inherits from `TaskPager.` This is where we start using Queue's backend functionalities for pagination.  We don't user `TaskSorter` because none of our columns will be sortable on the frontend. We automatically sort the tasks here based on `CAVC`, `AOD` and `DocketNumber.`
- Make few changes to `TaskFilter` and `QueueFilterParameter` so that it works with  our filterable column data.
- Serialize `suggestedHearingLocation` for Tasks so we can avoid calculating on the frontend
[...]

PR #13285, opened 2020-01-28
Add cols to cached_appeal_attributes for pagination

rubaiyat22:
connects #13141

### Description
- add two columns to cached_appeal_attributes table: `power_of_attorney_name` and `suggested_hearing_location` which will be used for filtering in hearings table
- add the newly created columns to update_cached_appeals_attributes_job
- create serializer for suggested_hearing_location which will be used in frontend in pagination work

### Database Changes
*Only for Schema Changes*

* [x] Column comments updated


.reek.yml
```diff
@@ -74,6 +74,7 @@ detectors:
       - VeteranAttributeCacher
       - WarmBgsCachesJob
       - ExternalApi::PexipService#send_pexip_request
+      - UpdateCachedAppealsAttributesJob#format_suggested_hearing_location
```
rubaiyat22: Ignoring feature_envy codeclimate error for now. Still thinking of how to resolve it

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20200102193042) do
+ActiveRecord::Schema.define(version: 20200127223143) do
```
pkarman: might need merge conflict resolved -- be sure the result is the highest value.

app/models/hearings/available_hearing_locations.rb
```diff
@@ -3,4 +3,25 @@
...
+    when "va_benefits_facility"
+      return "(BVA) " if loc.facility_id == "vba_372"
+
+      ro_or_vba
+    else
+      ""
+    end
+  end
+
+  def ro_or_vba
```
pkarman: maybe `determine_vba_facility_type` would be a more meaningful name?

.reek.yml
pkarman: Probably should be a method on the `appeal` maybe with a decorator pattern? see https://github.com/department-of-veterans-affairs/caseflow/wiki/Backend-Pattern-Decorator

app/models/hearings/available_hearing_locations.rb
```diff
@@ -3,4 +3,30 @@
...
+  # this will eventually replace getFacilityType in AppealHearingLocations
+  # in upcoming pagination work
+  def formatted_facility_type
+    case facility_type
+    when "vet_center"
+      "(Vet Center) "
+    when "va_health_facility"
+      "(VHA) "
+    when "va_benefits_facility"
+      return "(BVA) " if facility_id == "vba_372"
```
ferristseng: Small nit pick, but these all have a space at the end of the string that I'm not sure is needed

db/migrate/20200127223143_add_poa_name_and_suggested_hearing_loc_to_cached_appeal_attributes.rb
```diff
@@ -0,0 +1,6 @@
+class AddPoaNameAndSuggestedHearingLocToCachedAppealAttributes < ActiveRecord::Migration[5.1]
+  def change
+    add_column :cached_appeal_attributes, :power_of_attorney_name, :string, comment: "'Firstname, Lastname' of power of attorney"
+    add_column :cached_appeal_attributes, :suggested_hearing_location, :string, comment: "Suggested hearing location in 'City, State (Facility Type)' format"
```
ferristseng: Nice comments for the new columns!

Small nit though is that I think it should be 'Firstname Lastname' for the POA (no comma)

app/serializers/helpers/appeal_hearing_helper.rb
```diff
@@ -24,4 +25,9 @@ def hearings(appeal)
       AppealHearingSerializer.new(hearing).serializable_hash[:data][:attributes]
     end
   end
+
+  # returns the closest hearing location
+  def suggested_hearing_location(appeal)
+    available_hearing_locations(appeal).min_by { |loc| loc[:distance] }
```
ferristseng: One suggestion: `suggested_hearing_location` could live on the `Appeal` and `LegacyAppeal` as a shared concern instead so this logic isn't repeated here and in the update job.

Issue #12382, opened 2019-10-15
Missing Ratings | Handle save for verified unidentified issue


leikkisa:
### User or job story
As a user, when I create a verified unidentified issue, if a contention is created in VBMS, its contention text matches the issue's description.

### Acceptance criteria
- [ ] This work is behind the feature toggle: `verify_unidentified_issue`
  - [ ] This feature toggle was added in #12380
- [ ] The request_issues table has a new column called verified_unidentified_issue which is a boolean
- [ ] Saving a verified unidentified issue sets this value to true
- [ ] Contentions created for verified unidentified issues use the actual description or edited contention text for contention text (instead of "UNIDENTIFIED ISSUE - Please click *Edit in Caseflow* button to fix")
- [ ] Verified unidentified issues return true for `rating?`
- [ ] `is_unidentified` is set to false when creating verified unidentified issues

### Release notes
When a user adds a verified unidentified issue and it gets a contention in VBMS, the contention text will be set to the issue's description instead of "UNIDENTIFIED ISSUE - Please click *Edit in Caseflow* button to fix".

#### Designs
This is a backend ticket.

#### Background/context
When the contention text says "Unidentified issue", users cannot continue processing that issue.  Verifying the issue basically converts it from a regular unidentified issue, to one which can continue to getting a decision.

#### Technical notes
Since these are added on the Edit screen, they will get processed using request issues update.

PR #13287, opened 2020-01-28
 Handle save for verified unidentified issues

Sjones352:
Connects #12382

**Description:** This pr adds a verified_unidentified_issue column on request_issues table. The verified unidentified issue column allows an issue whose rating data is missing to be intaken as a regular rating issue.

Screen Shot 2020-01-28 at 5 30 22 PM
Screen Shot 2020-01-28 at 5 30 36 PM
Screen Shot 2020-01-27 at 7 38 52 PM

### Database Changes
*Only for Schema Changes*

* [x] Column comments updated


db/migrate/20200128033405_add_verified_unidentified_issue_to_request_issues.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddVerifiedUnidentifiedIssueToRequestIssues < ActiveRecord::Migration[5.1]
+  def change
+    add_column :request_issues, :verified_unidentified_issue, :boolean,
+               comment: "A verified unidentified issue allows an issue whose rating data is missing
+               to be intaken as a regular rating issue. In order to be marked as verified, a VSR needs
+               to confirm that they were able to find the record of the decision for the issue."
```
nanotone: This inserts a bunch of newlines and spaces into the comment, which I think is probably not your intention. It seems that for migrations, it's okay to have long lines for column comments (see `db/migrate/20190326202730_add_intake_database_comments.rb`)

spec/models/request_issue_spec.rb
```diff
@@ -923,6 +923,11 @@
       end
     end
 
+    context "verified unidentified issue returns true for rating" do
+      let(:verified_unidentified_issue) { true }
+      it { is_expected.to be true }
+    end
```
nanotone: Not sure if you're still working on this test, but I don't think it's testing what you want to test, since it doesn't seem to have a `subject`.

spec/feature/intake/higher_level_review_spec.rb
```diff
@@ -1481,6 +1481,27 @@ def complete_higher_level_review
...
+            safe_click "#Number-of-issues-has-changed-button-id-1"
+
+            expect(page).to have_current_path(
+              "/higher_level_reviews/#{higher_level_review.uuid}/edit/confirmation"
+            )
+
+            verified_issue = RequestIssue.find_by(verified_unidentified_issue: true,
+                                                  ineligible_reason: :legacy_appeal_not_eligible,
+                                                  vacols_id: "vacols1",
+                                                  vacols_sequence_id: "1")
```
leikkisa: `is_unidentified` should also be set to "false" here.

app/models/request_issue.rb
```diff
@@ -233,7 +235,8 @@ def rating?
     !!associated_rating_issue? ||
       !!previous_rating_issue? ||
       !!associated_rating_decision? ||
-      !!contested_decision_issue&.rating?
+      !!contested_decision_issue&.rating? ||
+      verified_unidentified_issue
```
leikkisa: I forgot to mention in my review, can you add this to the model test?

client/app/intake/util/issues.js
```diff
@@ -223,7 +225,8 @@ const formatRatingRequestIssues = (state) => {
 };
 
 const formatNonratingRequestIssues = (state) => {
-  return (state.addedIssues || []).filter((issue) => !issue.isRating && !issue.isUnidentified).map((issue) => {
+  return (state.addedIssues || []).filter((issue) => !issue.isRating && !issue.isUnidentified && !issue.verifiedUnidentifiedIssue).
```
nanotone: This is such a tiny nitpick, but while you're doing another pass... 😅🙏 
Can this line be broken up right before `filter`? It's getting very long, and the filter is its own step in the chain. Just to make it a bit more readable.

db/migrate/20200128033405_add_verified_unidentified_issue_to_request_issues.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddVerifiedUnidentifiedIssueToRequestIssues < ActiveRecord::Migration[5.1]
+  def change
+    add_column :request_issues, :verified_unidentified_issue, :boolean, comment: "A verified unidentified issue allows an issue whose rating data is missing to be intaken as a regular rating issue. In order to be marked as verified, a VSR needs to confirm that they were able to find the record of the decision for the issue."
```
pkarman: is there any need to track *who* verified the record?

Sjones352: @pkarman in order to avoid all this debacle of missing rates, VSRs need to be accountable while creating a verified unidentified issue because that's how Caseflow will know the difference between a rating and verified issue therefore; yes I do believe there is a need to track who verified the record.

pkarman: so the assumption is that whomever is checking the box is tracked already as either the intake user or the request_issues_update user?

Sjones352: Since this feature is available only on the edit page unidentified modal, I would say request_issues_update user

PR #13288, opened 2020-01-28
Appeal status nil decision date fix

### LINKED ISSUE #13286: Appeals Status API with nil decision date
Veteran id: 110169

https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/7763/

Raising error:
```
ArgumentError
comparison of ActiveSupport::TimeWithZone with nil failed
```

pkarman:
Resolves #13286 

### Description
When comparing a Time and nil object, Rails balks. Solution is to cast both to integers (nil casts to `0`).

```
> Time.zone.now < nil
Traceback (most recent call last):
        2: from (irb):51
        1: from (irb):51:in `<'
ArgumentError (comparison of ActiveSupport::TimeWithZone with nil failed)
> nil.to_i
=> 0
> Time.zone.now.to_i
=> 1580250229
> Time.zone.now.to_i < nil.to_i
=> false
```

Confirmed this patch fixes the production issue via console monkey patch.


spec/models/appeal_series_spec.rb
```diff
@@ -106,6 +106,13 @@
         ]
       end
 
+      it "handles nil decision_date comparison" do
+        allow(appeals.first).to receive(:decision_date) { nil }
+        allow(appeals.last).to receive(:decision_date) { Time.zone.now }
+
+        expect(series.latest_appeal).to_not be_nil
```
yoomlam: Should you ensure that `latest_active_appeal_by_last_location_change_date` returns nil so it doesn't shortcut calling `latest_appeal_by_decision_date` (for example by mocking `appeal.last_location_change_date`)?

pkarman: To ensure `latest_active_appeal_by_last_location_change_date` returns `nil` we either need to (1) make it a public method so we can mock it with `allow` or (2) make sure the data reflects a state where that method would return nil.

In this case we are following (2), since `latest_active_appeal_by_last_location_change_date` selects only `.active?` appeals and these appeals are completed (closed).

PR #13289, opened 2020-01-28
Show emails on hearing details

### LINKED ISSUE #13158: Virtual Hearings previous email is not shown correctly
- Observed: When a virtual hearing is changed back to video, the emails that were stored in the virtual hearing are no longer displayed.

- To be (expected): If the hearing was virtual at one point, and then converted back to video, the email addresses should be displayed, but not editable. 
image

rubaiyat22:
Resolves #13158 

### Description
- show email fields on details page if a hearing was previously a virtual hearing
- add `was_virtual?` method to hearing
- create serializer for `was_virtual?` and use it in frontend to determine whether to show email fields

Screen Shot 2020-01-29 at 5 29 21 PM


app/serializers/hearings/hearing_serializer.rb
```diff
@@ -74,10 +74,11 @@ class HearingSerializer
   attribute :veteran_email_address
   attribute :is_virtual, &:virtual?
   attribute :virtual_hearing do |object|
-    if object.virtual?
+    if object.virtual? || object.was_virtual?
```
ferristseng: You might need to check if/how we're using `virtualHearing` on the frontend. I noticed one regression related to how we use `virtualHearing` where some of the fields on the daily docket are disabled after changing a hearing back to video:

Screen Shot 2020-01-30 at 10 11 18 AM

rubaiyat22: Good catch! I think since now `virtualHearing` is always being returned that logic is no longer valid.

client/app/hearings/components/dailyDocket/DailyDocketRow.jsx
```diff
@@ -252,7 +252,7 @@ class DailyDocketRow extends React.Component {
           cancelUpdate={this.cancelUpdate}
           saveHearing={this.saveHearing}
           openDispositionModal={openDispositionModal}
-          readOnly={readOnly || (hearing.virtualHearing && !hearing.virtualHearing.jobCompleted)} />
+          readOnly={readOnly || (hearing.isVirtual && !hearing.virtualHearing.jobCompleted)} />
```
rubaiyat22: @ferristseng made changes here to look for `isVirtual` instead of `virtualHearing`

PR #13290, opened 2020-01-29
Use claimant POA to check access to appeal data

### LINKED ISSUE #10431: Investigate how to pull appellant POA
When the appellant is not the Veteran, we still need to show the appellant's POA. We've gotten one [support ticket](https://github.com/department-of-veterans-affairs/appeals-support/issues/4977) where VBMS is showing the appellant's POA, but Caseflow show there is no POA. 

## AC
- If the appellant is not the Veteran, populate appellant POA in the POA section of Case Details.
- This applies to both Legacy and AMA cases

## Other notes
- VSOs need to see cases where the appellant is not the Veteran, and they need to see their appellant's cases. So, we probably need to enhance the `TrackVeteranTask` to populate for appellants, too. 🤔

hschallhorn:
Resolves inability of VSO users to see their cases in Caseflow where the vet is deceased #10431

### Description
If the user cannot access the veteran's information in BGS because the veteran is not the claimant, checks to see if the user represents the claimant's POA before disallowing access.

### Acceptance Criteria
- [ ] VSO users who represent the claimant can access case details

### Testing
Test proposed change in production
```ruby
# Find an appeal where the claimant is not the veteran
appeal = Appeal.find_by(uuid: "42dde0a6-384b-495f-b1bc-8ccb72df4b10")

# Lets find the rep
vso = appeal.representatives.first

# Grab a user from the org
user = vso.users.first

# See if they can access this veteran's info in BGS
RequestStore[:current_user] = user
BGSService.new.can_access?(appeal.veteran_file_number)
=> false
# Cool, this is what is stopping the VSO from seeing the case

# Ensure fix would allow them access
appeal.veteran_is_not_claimant
=> true
appeal.representatives.any? { |rep| rep.user_has_access?(user) }
=> true
```


app/models/appeal.rb
```diff
@@ -271,6 +271,10 @@ def advanced_on_docket?
            :zip,
            :state, to: :appellant, prefix: true, allow_nil: true
 
+  def appellant_is_not_veteran
+    !!veteran_is_not_claimant
+  end
```
hschallhorn: alias `!!veteran_is_not_claimant` to
1. Pull out of serializers
2. Have common method between legacy and ama appeals

spec/controllers/appeals_controller_spec.rb
```diff
@@ -454,14 +454,13 @@
   end
 
   describe "GET appeals/:id" do
+    let!(:user) { User.authenticate!(roles: ["System Admin"]) }
```
kevmo: What's this for?

spec/controllers/appeals_controller_spec.rb
```diff
@@ -477,6 +476,33 @@
...
+        let!(:vso) { create(:vso, participant_id: appeal.claimant[:representative][:participant_id]) }
+
+        before do
+          allow_any_instance_of(LegacyAppeal).to receive(:appellant_is_not_veteran).and_return(true)
+          allow_any_instance_of(BGSService).to receive(:fetch_poas_by_participant_id)
+            .and_return([appeal.claimant[:representative]])
+        end
+
+        it "returns an error but does not send a message to Sentry" do
+          expect(Raven).to receive(:capture_exception).exactly(0).times
```
kevmo: Nice.

spec/controllers/appeals_controller_spec.rb
```diff
@@ -477,6 +476,33 @@
...
+          subject
+          expect(response.response_code).to eq(403)
+        end
+
+        context "when feature is enabled" do
+          before { FeatureToggle.enable!(:vso_calimant_representative) }
```
kevmo: I think this is supposed to be `claimant`

app/models/concerns/appeal_concern.rb
```diff
@@ -45,6 +45,12 @@ def appellant_last_first_mi
     end
   end
 
+  def user_represents_claimant_not_veteran
+    return false unless FeatureToggle.enabled?(:vso_calimant_representative)
```
kevmo: `claimant`

spec/controllers/appeals_controller_spec.rb
hschallhorn: moved from here: https://github.com/department-of-veterans-affairs/caseflow/pull/13290/files#diff-014acdeaec177c0d00b5630a510b75cfL463

app/models/concerns/appeal_concern.rb
```diff
@@ -45,6 +45,12 @@ def appellant_last_first_mi
     end
   end
 
+  def user_represents_claimant_not_veteran
+    return false unless FeatureToggle.enabled?(:vso_claimant_representative)
+
+    appellant_is_not_veteran && representatives.any? { |rep| rep.user_has_access?(current_user) }
```
yoomlam, line 51: @hschallhorn @araposo-tistatech 
Do we want allow VSO users to see case details if the appellant is the Veteran? [context](https://dsva.slack.com/archives/CHX8FMP28/p1593548945204400?thread_ts=1593543051.200400&cid=CHX8FMP28)

hschallhorn: This is handled by `BGSService.new.can_access?(appeal.veteran_file_number)`. Because this is checked as an OR, if the vso represents the veteran and the veteran is the appellant, they should have have access to the veteran in BGS and the first part of the or statement returns true.

yoomlam: In the case of this support ticket, `can_access?` returned `false`. 🤷‍♂️ 
Ann-Marie is checking with the Board.

Issue #13059, opened 2019-12-26
Legacy issue opt-in: adjust design for "None of these match"


leikkisa:
## User or job story
User story: As an intake user, when I'm selecting the legacy issue to connect to a request issue, the "None of these match" option seems like it belongs to a specific legacy appeal, and isn't an obvious option. I'd like it to be a clearer option.

## Acceptance criteria
- [ ] Add a new section on the legacy opt in modal, titled "None of these match"
- [ ] Change the radio option text for the "None of these match" radio option to "No VACOLS issues were found matching the issue: [issue description]"
- [ ] Include a screenshot of the new modal in the Github issue

## Release notes
When selecting legacy issues that match a request issue, the "None of these match" option is more apparent.

### Background/context
This option was intentionally de-emphasized early on to prevent users from selecting it without checking the legacy issues first.  It should stay at the bottom, but should still be apparent as an available option.

Due to this, and the button validation, I decided not to make "None of these match" its own button for proceeding, because we want to prevent users from proceeding too quickly.

### Design
This is the current design, not the new design
![Screen Shot 2019-12-12 at 2.41.17 PM.png](https://images.zenhubusercontent.com/58f6b6eb51aa2c30577bb60e/f16d3939-ed03-43a8-80fa-fd1923aabc07)

### Resources/other links
https://dsva.slack.com/archives/CFBLH2LR4/p1576190705464900

jcq:
Screenshot of Updated UI:
<img width="748" alt="Screen Shot 2020-01-29 at 10 32 06" src="https://user-images.githubusercontent.com/2581274/73385808-a8006000-4282-11ea-9fd5-cfbb1f7d1a60.png">

jcq:
Upping the estimate from 1 to 2 due to having to deal with updating numerous tests.

PR #13295, opened 2020-01-29
Updated LegacyOptInModal to put "None of these match" option in its own section

jcq:
Connects #13059

### Description
Updated `LegacyOptInModal` to give "None of these match" option its own section.


jcq:
Updated ticket with screenshot of updated UI:
Screen Shot 2020-01-29 at 10 32 06

PR #13296, opened 2020-01-29
Use stream_docket_number for docket number in SQL code

nanotone:
Connects #13211

### Description
Epilogue in a series of PRs (previously: #13241 #13258 #13280) to support appeal streams by storing the docket number and type in the DB.

Now that `stream_docket_number` is created and accurately populated for all existing and future appeals, the SQL code no longer needs to mirror the old Rails code in computing docket numbers on the fly.

One broad question: we do want this diff, right? Assuming Tableau wants accurate docket numbers once MTV appeals land.

And one very narrow concern:

An Appeal that's been saved just once, with a receipt_date and a PG-generated ID, will have a logical docket number but not a `stream_docket_number`, since Rails can't know the ID pre-save. Rails-side, this isn't any worse than before, but this simplified SQL query would now see a NULL.

In practice, it's not currently a problem since the appeal is created by `IntakesController#create` and receipt_date lands later in `#review`. Should we protect against the hypothetical bug, though? Either by revisiting this SQL code, or making Rails do a double-save.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. All tests in `spec/sql/` still passing.

### DIFF
--- a/app/models/appeal.rb
+++ b/app/models/appeal.rb
@@ -32,6 +32,7 @@ class Appeal < DecisionReview
   }
 
   before_save :set_stream_docket_number_and_stream_type
+  after_save :set_stream_docket_number_from_new_id
 
   with_options on: :intake_review do
     validates :receipt_date, :docket_type, presence: { message: "blank" }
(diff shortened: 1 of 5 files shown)


nanotone:
@leikkisa I think I'm going to try adding an extra `after_save` hook rather than change it from `before_save`. Here's how I'm thinking of them:

- `before_save` is for when you have `{id: 123, receipt_date: nil, stream_docket_number: nil}` and you set the `receipt_date` (which is what we currently do, I believe). This way, you can update `receipt_date` and `stream_docket_number` in one save.
- `after_save` is for when you have `{id: nil, receipt_date: some_date, stream_docket_number: nil}`, do a save, and the new ID is enough to determine the docket number (so we do a double-save).
- If I understand your `tap` suggestion correctly, that puts the onus of double-saving on whoever's using the model, which I'm not a big fan of because it's safer to assume they're not familiar with the subtleties of the stream docket number.

New code and accompanying test pushed.

Issue #13037, opened 2019-12-19
AMA QR code requires a second urgent update


jimruggiero:
User story: As a member of the Quality Review (QR) team at the Board, I need the AMA quality review monthly limit and selection probability to be updated to .188, so that I can reliably receive 130+ AMA cases to review monthly, up to **133** case per month, not 137 as is currently set.

### Acceptance criteria
- [ ] The monthly limit of Quality Reviews is set at 133
- [ ] The Quality Review selection probability is set at 0.188

### Release notes
Caseflow now supports sending sending up to 133 AMA cases to QR for review monthly.

### Out of scope
Changes to legacy QR code.

### Background/context
It is believed that the prior update made in #12867 was insufficient. The Board on had 7 cases directed to QR last week and was expecting a significantly higher volume. 

### Resources/other links
Legacy code lives here: https://github.com/department-of-veterans-affairs/caseflow/blob/555ac6727ea665985dd42a2f325cb498686c68f0/app/models/judge_case_review.rb

hschallhorn:
#12915 was merged 9 days ago, lets see what the spread looks like in prod
```ruby
week_qr_count  = QualityReviewTask.where(appeal_type: Appeal.name, assigned_to_type: Organization.name, created_at: 7.days.ago..Time.zone.now).count
week_dispatch_count = BvaDispatchTask.where(appeal_type: Appeal.name, assigned_to_type: Organization.name, created_at: 7.days.ago..Time.zone.now).count
week_completed_cases_count = week_qr_count + week_dispatch_count
month_completed_cases_count = week_completed_cases_count / 7.0 * 30
# Let's see what our number should produce
month_completed_cases_count * QualityReviewCaseSelector:: QUALITY_REVIEW_SELECTION_PROBABILITY
=> 31.54285714285714
# Let's see what the new number would give us
month_completed_cases_count * 0.188
=> 185.31428571428572
# Let's confirm this for the last few months
6.times do |number_of_months_ago|
  month = (number_of_months_ago + 1).months.ago
  month_qr_count  = QualityReviewTask.where(appeal_type: Appeal.name, assigned_to_type: Organization.name, created_at: month.beginning_of_month..month.end_of_month).count
  month_dispatch_count = BvaDispatchTask.where(appeal_type: Appeal.name, assigned_to_type: Organization.name, created_at: month.beginning_of_month..month.end_of_month).count
  month_completed_cases_count = month_qr_count + month_dispatch_count
[...]

PR #13304, opened 2020-01-30
Do not double count cancelled or child tasks when determining monthly limit for QR tasks

hschallhorn:
Resolves issue where QR has stopped getting tasks at only 95.
References #13037

### Description
Stopped double counting tasks. We previously would count the child tasks as well.
```ruby
# This is the code that we check the limit against
QualityReviewTask.created_this_month.size
=> 168

# I wonder if this includes legacy appeals
QualityReviewTask.created_this_month.group(:appeal_type).count
=> {"Appeal"=>168}
# Nope

# I wonder if this includes user tasks?
QualityReviewTask.created_this_month.group(:assigned_to_type).count
=> {"Organization"=>77, "User"=>91}
# YEP
```

### Acceptance Criteria
- [ ] Cancelled QR tasks are not counted when calculating the monthly limit
- [ ] More then one QR task on an appeal are not counted when calculating the monthly limit

### DIFF
--- a/spec/models/quality_review_case_selector_spec.rb
+++ b/spec/models/quality_review_case_selector_spec.rb
@@ -24,6 +24,8 @@ describe QualityReviewCaseSelector, :all_dbs do
     end
 
     context "after reaching the monthly limit" do
+      before { stub_const("QualityReviewCaseSelector::MONTHLY_LIMIT_OF_QUAILITY_REVIEWS", 5) }
+
       let!(:qr_tasks) { create_list(:qr_task, QualityReviewCaseSelector::MONTHLY_LIMIT_OF_QUAILITY_REVIEWS) }
 
       it "returns true" do
--- a/spec/models/quality_review_case_selector_spec.rb
+++ b/spec/models/quality_review_case_selector_spec.rb
@@ -2,6 +2,7 @@
 
 describe QualityReviewCaseSelector, :all_dbs do
   describe ".reached_monthly_limit_in_quality_reviews?" do
+    subject { QualityReviewCaseSelector.reached_monthly_limit_in_quality_reviews? }
     context "when a realistic number of cases are completed" do
       # Pulled from prod with:
       # Task.where(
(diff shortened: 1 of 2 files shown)


spec/models/quality_review_case_selector_spec.rb
```diff
@@ -25,14 +24,29 @@
     end
 
     context "after reaching the monthly limit" do
-      let!(:qr_tasks) do
-        QualityReviewCaseSelector::MONTHLY_LIMIT_OF_QUAILITY_REVIEWS.times do
-          QualityReviewTask.create(assigned_to: create(:user), appeal: create(:appeal), created_at: Time.zone.now)
+      let!(:qr_tasks) { create_list(:qr_task, QualityReviewCaseSelector::MONTHLY_LIMIT_OF_QUAILITY_REVIEWS) }
```
lomky: I was pondering this a bit last time I was here - do we save any runtime if we hardset the value of 
`QualityReviewCaseSelector::MONTHLY_LIMIT_OF_QUAILITY_REVIEWS` to be lower?

spec/models/quality_review_case_selector_spec.rb
```diff
@@ -25,14 +24,29 @@
...
+
+      it "returns true" do
+        expect(subject).to be(true)
+      end
+
+      context "but some tasks are cancelled" do
+        before do
+          allow_any_instance_of(BvaDispatch).to receive(:next_assignee).and_return(false)
```
lomky: what's this doing here? A comment explaining why we're talking to BvaDispatch would be nice. Are we trying to cut off the callbacks? If so we could call `update_column` instead of `update!`, and not have to use `allow_any_instance_of`?

spec/models/quality_review_case_selector_spec.rb
hschallhorn: We definitely would! Runtime on this was concerning me too. Great idea!

Issue #13222, opened 2020-01-22
Motion to Vacate | Granted | After AddressMotionToVacate, create appeal stream if granted


leikkisa:
## User or job story
When the Judge addresses a motion to vacate, and it is granted, it should move onto a new appeal stream.

## Acceptance criteria
- [ ] After the AddressMotionToVacateTask, if the judge selects grant, or partial grant, create a new appeal stream
- [ ] Add the original appeal's claimants to the new appeal
- [ ] Add a RootTask
- [ ] Add a JudgeAssignTask, which is a child of RootTask, and assigned to the Judge
- [ ] Clean up code in PostDecisionMotionUpdater that creates other tasks for granted appeals

## Out of scope
- Adding request or decision issues (see #13069)

## Questions
What happens to the additional instructions/context from the AddressMotionToVacateTask?

## Appeal details
Create a copy of the original appeal:
- veteran_file_number: same as original
- legacy_opt_in_approved: same as original
- veteran_is_not_claimant: same as original
- docket_type: same as original
- docket_number: same as original
- type: "Vacate"

## Release notes
N/A

leikkisa:
Blocking currently until we learn whether denials/dismissals should get their own appeal stream, since in the current flow they do not get a decision document.

If they do _not_ get a new appeal stream, the contents of this ticket will be replaced by starting the new appeal stream only after AddressMotionToVacate.

leikkisa:
Moved to current sprint since due to priority, but we marked it as not originally being pulled in.

PR #13325, opened 2020-02-01
Create a Vacate appeal stream for a Motion to Vacate

leikkisa:
connects #13222

### Description
- Creates a new Vacate appeal stream for granted motions to vacate
- Adds initial tasks to the new appeal stream
- Cleans up tasks that are no longer used

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...

### User Facing Changes

 - [ ] Screenshots of UI changes added to PR & Original Issue
 
### Documentation Updates

### Database Changes
*Only for Schema Changes*

* [ ] Column comments updated


app/workflows/initial_tasks_factory.rb
```diff
@@ -9,7 +9,7 @@ def initialize(appeal)
   def create_root_and_sub_tasks!
     create_vso_tracking_tasks
     ActiveRecord::Base.transaction do
-      create_subtasks!
+      create_subtasks! if @appeal.Original?
```
leikkisa: Should we lowercase this in the DB, and just upcase it on the frontend?

app/models/decision_issue.rb
```diff
@@ -246,7 +246,7 @@ def create_remand_supplemental_claim!
     )
     fail AppealDTAPayeeCodeError, decision_review.id unless dta_payee_code
 
-    sc.create_claimants!(
+    sc.create_claimant!(
```
leikkisa: Small rename to be more accurate

app/workflows/post_decision_motion_updater.rb
```diff
@@ -54,85 +71,71 @@ def create_new_tasks
...
   end
 
   def task_type
-    grant_type? ? params[:vacate_type] : "#{params[:disposition]}_motion_to_vacate"
+    "#{disposition}_motion_to_vacate"
   end
 
   def grant_type?
-    %w[granted partial].include? params[:disposition]
+    %w[granted partially_granted].include? disposition
```
leikkisa: I decided to make the front-end version of the disposition, partial -> partially_granted

app/workflows/post_decision_motion_updater.rb
```diff
@@ -54,85 +71,71 @@ def create_new_tasks
...
+    task.assigned_to
+  end
+
+  def attorney_user
+    @attorney_user ||= (denied_or_dismissed? ? prev_motions_attorney : User.find_by(id: params[:assigned_to_id]))
   end
 
   def prev_motions_attorney
-    mtv_mail_task.assigned_to
+    task.parent.assigned_to
```
leikkisa: I think this is clear enough without making a method for it.

spec/controllers/post_decision_motions_controller_spec.rb
```diff
@@ -31,44 +31,55 @@
...
       end
     end
 
-    context "when there is no judge team" do
-      it "raises NonexistentJudgeTeam error" do
+    context "when the motion is valid" do
+      let!(:attorney_staff) { create(:staff, :attorney_role, sdomainid: user.css_id) }
+      before do
+        allow(judge).to receive(:judge_in_vacols?).and_return(true)
+        allow_any_instance_of(PostDecisionMotionUpdater).to receive(:judge_user).and_return(judge)
```
leikkisa: I added this so that the PostDecisionMotionUpdater's judge_user would return true for judge_in_vacols?  This is extra validation that is done on AttorneyTask

app/workflows/initial_tasks_factory.rb
nanotone: Revealing my Rails newbie status here... where is `Original?` even defined? I don't see it as a database field or as a method in `appeal.rb`, but it is clearly available on the Appeal object.

app/workflows/post_decision_motion_updater.rb
Sjones352: Ok newbie question but why are we changing it from mtv_mail_task to task.parent

app/workflows/initial_tasks_factory.rb
leikkisa: It's magic that happens when you enumerate a field, which you did for stream_type.

app/workflows/post_decision_motion_updater.rb
leikkisa: Not a newbie question! This was a completely unnecessary change, I just thought that task.parent.assigned_to was just as clear and the mtv_mail_task method was unneeded.

If you think it's harder to read now, I can put it back.

app/workflows/initial_tasks_factory.rb
nanotone: Thanks for the explanation, truly magical. Yes, if I understand correctly (that the key is what's used Rails-side, and the value is what's used DB-side) then it'd be better for the keys to all look like method names. Including renaming the key `De Novo` to `de_novo`.

app/workflows/post_decision_motion_updater.rb
nanotone: I think this works. My original inclination was to keep `mtv_mail_task` if `task.parent` is mentioned anywhere else in this file, but the one other place is in `create_abstract_task` which isn't too topical to `mtv_mail_task`.

app/workflows/post_decision_motion_updater.rb
```diff
@@ -1,26 +1,45 @@
...
+  delegate :appeal, to: :task
+
   def process
     ActiveRecord::Base.transaction do
-      motion = create_motion
-      return unless motion
+      return unless post_decision_motion
+
+      create_new_tasks if denied_or_dismissed?
+      if grant_type?
```
nanotone: Are `grant_type?` and `denied_or_dismissed?` the universe here? I saw some documentation mention a withdraw disposition, but not sure if that flow enters this method.

app/workflows/initial_tasks_factory.rb
jcq: I vote for lowercase anytime it's a property that we might be accessing. We'd want to display it uppercase when actually being displayed to a user on the frontend, but even in React components we'd want to be able to access/assume it as lowercase, I'd say.

app/workflows/post_decision_motion_updater.rb
leikkisa: Good question. Withdrawals seem to get the "dismissed" disposition, according to designs.

Screen Shot 2020-02-04 at 3 49 37 PM

app/workflows/initial_tasks_factory.rb
nanotone: Here's the one spot on the FE I changed in my parallel diff: https://github.com/department-of-veterans-affairs/caseflow/blob/413674dcbaacaf7826baa266d84129aa932785df/client/app/queue/utils.js#L122

app/workflows/post_decision_motion_updater.rb
jcq: The earliest designs had "withdrawn" as a separate disposition, but that got folded into "dismissed" fairly early on, as I recall.

app/workflows/initial_tasks_factory.rb
nanotone: Resolving this thread because we decided to punt on lowercasing: #13358

app/workflows/post_decision_motion_updater.rb
nanotone: Wow, so much context, thanks! The only reason I bring this up, is that these two lines feel like an if/else pair, but it's not obvious from reading the code whether `denied_or_dismissed?` and `grant_type?` are exactly opposites of each other. There may not be a better way to go, just readabality for thought.

nanotone: Aha, I wasn't completely off base. Code climate thinks this method has about 2x too much cognitive complexity 🙃

app/workflows/post_decision_motion_updater.rb
```diff
@@ -1,26 +1,45 @@
...
 
+##
+# The PostDecisionMotionUpdater validates and creates a PostDecisionMotion from the JudgeAddressMotionToVacateTask,
+# and creates the subsequent tasks or appeal streams.
+
 class PostDecisionMotionUpdater
   include ActiveModel::Model
 
-  attr_reader :task, :params
+  attr_reader :task, :params, :disposition, :instructions
```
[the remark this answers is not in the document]
nanotone: It just means Ruby (not Rails, iiuc) will automatically make `pdmu.instructions` and `pdmu.instructions=` methods for you, to decouple it from the `@instructions` internal attribute.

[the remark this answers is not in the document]
leikkisa: I think since we're using it in two places below, I thought it was nicer to just use "instructions" instead of `params[:instructions]`. But it's really not a meaningful or necessary change. I don't even think it's less code.  It's nicer for `disposition` though, which is used several times.

`attr_reader` (and related: `attr_writer`, and `attr_accessor` combines read/write) are useful if want something that's not saved in DB columns to be an attribute. Some models don't even have DB tables, so all of their attributes are defined this way (see EndProduct.rb)

Issue #13007, opened 2019-12-17
Motion to Vacate | Granted | Update attorney review screen


jimruggiero:
As a decision drafting attorney I want to see a screen that allows me to review granted motions to vacate from a judge. 

## Acceptance criteria
- [ ] When the appeal stream_type is Vacate, after a decision attorney selects the "Decision ready for review" action on their AttorneyTask, they should be directed to `/appeals/[uuid]/tasks/[id]/draft_decision/vacatures` (ultimately up to the engineer)
  - Note: This page will use the same component as the `draft_decision/dispositions` page.

When the URL matches that pattern, the component implements the following changes:
- [ ] Update header copy on the "Draft decisions" screen to "Review vacated decision issues"
- [ ] Update the copy for the text below the header (see Copy)
- [ ] Hide the "Add decision" button
- [ ] Hide "Delete" for now (added to scope afterwards, differs from screenshot)

## Copy
Header: Review vacated decision issues
Text: Review the prior decision issues here that have been marked for vacatur by the Judge. Please remember to include them in your decision letter if you haven't already.

### Out of scope
- "return to judge" banner
- Top bar showing steps 1 and 2 (this varies per vacate type)

### Designs
https://app.mural.co/t/workqueue2001/m/workqueue2001/1573574515518/2c771474e7259260c0ebc68b44c78bb98f295b04

![b872c96b-7f39-4c0c-8df1-5d15f76f2d0a.png](https://images.zenhubusercontent.com/58f6b6eb51aa2c30577bb60e/0b032e49-7d8c-4e4d-9995-29f4d936b631)

### Tech notes
This is happening on the AttorneyTask, which is a child of JudgeDecisionReviewTask
The Draft Decisions URL that this page shows up on:
`/queue/appeals/887309dd-2bda-4d38-b4e9-7545099b34b3/tasks/9/draft_decision/dispositions`

Original copy of text:
"Add decisions to each issue the Veteran is appealing. Any decisions you add must be connected to at least one issue on appeal. This helps Caseflow gather metrics on BVA decisions and meet AMA reporting requirements."

### Open Questions
N/A

### Resources/other links
N/A

jimruggiero:
![image (9).png](https://images.zenhubusercontent.com/5d2748c531297d7d65f8d26b/b872c96b-7f39-4c0c-8df1-5d15f76f2d0a)

jcq:
[...]

PR #13345, opened 2020-02-04
MTV | Attorney Review Decision Issues View

jcq:
Connects #13007

### Description
Added `ReviewVacatedDecisionsIssuesView` into structure of `MotionToVacateFlowContainer` to allow attorney to review the selected decision issues for vacatur.

### Acceptance Criteria
- [ ] Code compiles correctly


app/repositories/task_action_repository.rb
```diff
@@ -325,6 +325,9 @@ def review_decision_draft(task, user)
       action = Constants.TASK_ACTIONS.REVIEW_LEGACY_DECISION.to_h
       action = Constants.TASK_ACTIONS.REVIEW_AMA_DECISION.to_h if task.ama?
 
+      # TODO: Update this to access task.appeal.stream_type == vacate
+      action = Constants.TASK_ACTIONS.REVIEW_VACATE_DECISION.to_h if task.ama?
```
leikkisa: I think this task action is actually going away altogether

jcq: Why would it be going away? We'll still be triggering this from the `AttorneyTask`, right?

leikkisa: Hmm, I'm going to stay out of it for now to avoid micromanaging, especially since we had some good conversations about the url paths, and you might want to handle those through different task actions instead of using the existing "decision ready for review one".

spec/feature/queue/motion_to_vacate_spec.rb
```diff
@@ -385,6 +385,70 @@
...
+    end
+    let!(:post_decision_motion) { post_decision_motion_updater.process }
+    let(:vacate_stream) { Appeal.find_by(stream_docket_number: appeal.docket_number, stream_type: "vacate") }
+    let(:attorney_task) { AttorneyTask.find_by(assigned_to: drafting_attorney) }
+
+    before do
+      judge_team.add_user(drafting_attorney)
+      FeatureToggle.enable!(:review_motion_to_vacate)
+
+      judge_address_motion_to_vacate_task.update(status: Constants.TASK_STATUSES.completed)
```
leikkisa: This could go where you're creating the task in the first place

spec/feature/queue/motion_to_vacate_spec.rb
```diff
@@ -385,6 +385,70 @@
...
+
+    let(:post_decision_motion_params) do
+      {
+        instructions: "I am granting this",
+        disposition: "granted",
+        vacate_type: "straight_vacate",
+        assigned_to_id: drafting_attorney
+      }
+    end
+    let(:post_decision_motion_updater) do
```
leikkisa: This is all good, but for future reference, you can add onto to the previous feature tests instead of breaking off a new one for each PR. That'll probably simplify things when we're starting to test the differences between vacate types, and test setup as well.

client/app/queue/mtv/mtvCheckoutSteps.js
```diff
@@ -0,0 +1,30 @@
+export const views = {
+  review_vacatures: { title: 'Review vacated decision issues' },
+  add_decisions: { title: 'Add decisions' },
+  submit: { title: 'Submit draft decision for review' }
+};
```
leikkisa: Nice

client/app/queue/mtv/ReviewVacatedDecisionIssuesView.jsx
```diff
@@ -0,0 +1,75 @@
...
+      getNextStepUrl={() => ctx.getNextUrl('review_vacatures')}
+      getPrevStepUrl={() => ctx.getPrevUrl('review_vacatures')}
+    >
+      <h1>{COPY.MTV_CHECKOUT_REVIEW_VACATURES_TITLE}</h1>
+      <p>{COPY.MTV_CHECKOUT_REVIEW_VACATURES_EXPLANATION}</p>
+      <hr />
+      <AmaIssueList requestIssues={appeal.issues} errorMessages={{}}>
+        <DecisionIssues
+          decisionIssues={ctx.decisionIssues}
+          //   openDecisionHandler={this.openDecisionHandler}
```
leikkisa: Is this for the "Edit" modal?

client/app/queue/mtv/MotionToVacateFlowContainer.jsx
```diff
@@ -1,34 +1,50 @@
...
-      <MotionToVacateContextProvider>
+      <MotionToVacateContextProvider initialState={initialState}>
         {/* MTV Progress Bar (#13319) Here */}
 
         <Switch>
           <Route path={`${path}/review_vacatures`}>
             {/* Insert component from #13007 here */}
-            {/* <ReviewVacatedDecisionIssuesView appeal={appeal} /> */}
-            <></>
+            <ReviewVacatedDecisionIssuesView appeal={appeal} />
```
leikkisa: I think the comment above this is handled now in this PR

client/app/queue/mtv/mtvCheckoutSteps.js
```diff
@@ -0,0 +1,30 @@
...
+
+// This might be more elegantly modeled w/ a finite state machine lib like xstate
```
leikkisa: I think it might make more sense to put this comment into a tech improvement ticket, unless you are planning on addressing it in this current round of work.

client/app/queue/mtv/mtvCheckoutSteps.js
```diff
@@ -0,0 +1,30 @@
...
+export const getSteps = ({ type, vacateType }) => {
+  switch (vacateType?.toLowerCase()) {
```
leikkisa: I think the question mark is extra, and don't think the `toLowerCase` is necessary although I can't imagine it hurts. But lmk if I'm missing knowledge of a JS feature.

spec/feature/queue/motion_to_vacate_spec.rb
```diff
@@ -385,6 +385,70 @@
...
+
+      find(".Select-placeholder", text: COPY::TASK_ACTION_DROPDOWN_BOX_LABEL).click
+      find("div", class: "Select-option", text: Constants.TASK_ACTIONS.REVIEW_VACATE_DECISION.label).click
+
+      path = [
+        "/queue/appeals/#{vacate_stream.uuid}/tasks/#{attorney_task.id}",
+        "motion_to_vacate_checkout/review_vacatures"
+      ].join("/")
+
+      expect(page.current_path).to eq(path)
```
leikkisa: I tested this locally, but it might be good to add a little more to this feature test to make make it apparent in the test that the correct page is rendering.

client/app/queue/mtv/mtvCheckoutSteps.js
```diff
@@ -0,0 +1,30 @@
...
+  case 'straight_vacate':
+  case 'vacate_and_de_novo':
+    return ['review_vacatures', 'submit'];
+  case 'vacate_and_readjudicate':
+    return ['review_vacatures', 'add_decisions', 'submit'];
+  default:
+    return type?.toLowerCase() === 'de_novo' ? ['add_decisions', 'submit'] : [];
```
leikkisa: I like this structure. I have the same ? about the ? as above :-P.

This assumes we are not sending the vacateType for de novos, but if it's in the serializer you might want to add the check for the "de_novo" type above the vacateType checks.

client/app/queue/mtv/RemoveDecisionIssueModal.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import COPY from '../../../COPY';
+import Modal from '../../components/Modal';
+
+export const RemoveDecisionIssueModal = ({ onCancel, onSubmit }) => {
```
leikkisa: We decided to remove the "delete" link from this page for now, but this is still useful code to have if you're not going with re-using the existing code.  I have a preference for re-using existing code for when someone hits "delete", but haven't compared the options and don't know the arguments for creating a new modal for this.

If you're going to keep it, can you add it to the feature test?

client/app/queue/mtv/RemoveDecisionIssueModal.jsx
```diff
@@ -0,0 +1,29 @@
...
+    { classNames: ['usa-button', 'usa-button-primary'],
+      name: 'Yes, delete decision',
+      onClick: onSubmit }
+  ];
+
+  return (
+    <Modal buttons={buttons} closeHandler={onCancel} title="Delete decision">
+      <span className="delete-decision-modal">
+        {COPY.DECISION_ISSUE_CONFIRM_DELETE}
+        {/* {toDeleteHasConnectedIssue && COPY.DECISION_ISSUE_CONFIRM_DELETE_WITH_CONNECTED_ISSUES} */}
```
leikkisa: I'm not sure what this is, a To Do for the next round?

client/app/queue/mtv/MotionToVacateFlowContainer.jsx
```diff
@@ -1,34 +1,50 @@
...
+
           <Route path={`${path}/add_decisions`}>
             {/* Insert component from #13071 here */}
             {/* <AddDecisionsView appeal={appeal} /> */}
-            <></>
+            <h1>add_decisions</h1>
+          </Route>
+
+          <Route path={`${path}/submit`}>
+            {/* Insert component from #13071 here */}
```
leikkisa: This might be a copy and paste from above, but this will be a different screen than the AddDecisions view.

client/app/queue/mtv/ReviewVacatedDecisionIssuesView.jsx
```diff
@@ -0,0 +1,75 @@
...
+  };
+
+  const onCancelDelete = () =>
+    setState({
+      ...state,
+      deleteModal: false,
+      issueId: null
+    });
+
+  const onDeleteSubmit = () => {
```
leikkisa: The modal says it'll save deletes, but it doesn't. It looks like the intention is to wire it up in this PR, but I'm not sure if that is actually the intention.

client/app/queue/mtv/MotionToVacateFlowContainer.jsx
jcq: Yeah, I have updated that to note the new correct ticket (#13385).

client/app/queue/mtv/RemoveDecisionIssueModal.jsx
jcq: This was in the delete modal for normal flows — it would output this text:
```
Deleting this decision will also remove it from other issues you've added it to.
```
But I'm pretty sure that wouldn't be relevant here and we should just remove

client/app/queue/mtv/mtvCheckoutSteps.js
jcq: Good flag — I assumed it would be `null` when `type` is "de_novo", but wasn't positive. I've checked with Yang and he confirmed it would be `null`.

client/app/queue/mtv/ReviewVacatedDecisionIssuesView.jsx
jcq: Yeah, that was a reminder of how we would implement "Edit", but since we removed all of the rest of the code, it's silly to leave that in.

client/app/queue/mtv/ReviewVacatedDecisionIssuesView.jsx
jcq: Decent point — I'm switching to different copy that simply refers to "removing" the decision rather than implying that confirming will perform a delete.

client/app/queue/mtv/RemoveDecisionIssueModal.jsx
jcq: We're maintaining a different data structure, so we need different handlers to update things. That said, you're right about currently removing it from the AC, so I've gone ahead and made it conditional upon a prop that defaults to off.

spec/feature/queue/motion_to_vacate_spec.rb
jcq: Hmm... Yeah, it might make sense to restructure motion_to_vacate_spec to break things down in different ways. Instead of having chunks based around the type of task and then blocks under that dealing with the different vacate types, etc, it might be better to describe what type of decision is being done and then lay out the whole flow, switching users as need be.

That seems like it's probably out of scope for this particular ticket, but if you think it would be a good improvement, perhaps we should create a new one for that?

spec/feature/queue/motion_to_vacate_spec.rb
jcq: Oops. I remember now why that was put where it was; tasks can only be created with `status` of "assigned". Reverting. 😕

client/app/queue/mtv/ReviewVacatedDecisionIssuesView.jsx
jcq: (The functionality in here removes it from the list of decision issues stored in the container context, but it doesn't actually persist that to the server.)

spec/feature/queue/motion_to_vacate_spec.rb
leikkisa: I was mentioning it for future tickets, so you know that's an option. I do think it'll be a good idea to review the tests after we're done (and add things like the delete modal when it's accessible).

Issue #13333, opened 2020-02-03
Motion to Vacate | Serialize appeal stream information


nanotone:
## Description

In order to support UI flows for MTV, the frontend needs serialized appeal objects to include the appeal's stream type and vacate type.

## Acceptance criteria

- [ ] Appeal serializer includes `type`
- [ ] Appeal serializer includes `vacate_type` when stream type is `Vacate`

## Background/context

Appeal streams are separate Appeal objects that share the same docket number but different stream types, and indicate work that can be done in parallel on a case. For a Motions to Vacate flow, the stream type is `Vacate`, and the vacate type for a granted motion indicates whether it is a straight vacate, vacate and readjudication, or vacate and de novo.

Because much of the MTV flow is shared with that of a normal Appeal, both in terms of frontend UI and backend data modeling, we want to expose stream type and vacate type information in the Appeal serializer so that the frontend can display the correct UI.

## Technical notes

The Appeal serializer lives in `app/serializers/intake/appeal_serializer.rb`. The stream type is available via the `Appeal#type` method.

The vacate type, which should only be present when the stream type is `Vacate`, is a bit more complicated to retrieve: it exists as `PostDecisionMotion#vacate_type` via the post-decision motion whose task is a `JudgeAddressMotionToVacateTask` on the appeal's task tree.

jcq:
We will also need to provide the `vacated_issues` for the `PostDecisionMotion` associated with the task/appeal to the frontend. Does it make sense to add that here?

If this seems out of scope, we could always create another ticket for it.

**Context**
https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/post_decision_motion.rb#L24

nanotone:
@jcq Good catch! Yeah, it could totally go here -- would it be okay to include as an extra field on the serialized appeal?

jcq:
After talking to @leikkisa, my question might actually be moot. While obtaining them from the *original* appeal is a bit complicated (and what I was thinking with my first question), it's trivial to get them from the *new* appeal stream — they will just be *the* decision issues for that appeal.

PR #13346, opened 2020-02-04
Serialize appeal stream information for vacate-type appeals

nanotone:
Connects #13333 

### Description
This PR adds `type` and `vacate_type` fields to the Appeal serializer, so that the frontend can show the appropriate UI for MTV-related appeal streams.

### Acceptance Criteria
- [ ] Appeal serializer includes `type`
- [ ] Appeal serializer includes `vacate_type` when stream type is `Vacate`

### Testing Plan
- Added test for `vacate_type` when appeal is vacate and non-vacate
  - Added `:straight_vacated` trait to appeal factorybot, to help with this and future MTV tests

### DIFF
--- a/app/models/appeal.rb
+++ b/app/models/appeal.rb
@@ -97,6 +97,12 @@ class Appeal < DecisionReview
     end
   end
 
+  def vacate_type
+    return nil unless vacate?
+
+    post_decision_motion&.vacate_type
+  end
+
   # Returns the most directly responsible party for an appeal when it is at the Board,
   # mirroring Legacy Appeals' location code in VACOLS
   def assigned_to_location
@@ -459,5 +465,11 @@ class Appeal < DecisionReview
   ensure
     distribution_task = tasks.open.find_by(type: DistributionTask.name)
     TranslationTask.create_from_parent(distribution_task) if STATE_CODES_REQUIRING_TRANSLATION_TASK.include?(state_code)
+  end
+
+  # Non-vacate Appeals are not expected to have a PDM, but this method makes a
+  # best-effort attempt in either situation, and returns nil if none is found.
+  def post_decision_motion
+    PostDecisionMotion.where(task: tasks).first
   end
 end
--- a/app/models/appeal.rb
+++ b/app/models/appeal.rb
@@ -470,6 +470,6 @@ class Appeal < DecisionReview
   # Non-vacate Appeals are not expected to have a PDM, but this method makes a
   # best-effort attempt in either situation, and returns nil if none is found.
   def post_decision_motion
-    PostDecisionMotion.where(task: tasks).first
+    PostDecisionMotion.find_by(task: tasks)
   end
 end
--- a/app/models/appeal.rb
+++ b/app/models/appeal.rb
@@ -98,8 +98,6 @@ class Appeal < DecisionReview
   end
 
   def vacate_type
-    return nil unless vacate?
-
     post_decision_motion&.vacate_type
   end
 
--- a/app/serializers/intake/appeal_serializer.rb
+++ b/app/serializers/intake/appeal_serializer.rb
@@ -9,6 +9,4 @@ class Intake::AppealSerializer < Intake::DecisionReviewSerializer
   attribute :form_type do
     "appeal"
   end
-  attribute :type
-  attribute :vacate_type, if: proc { |appeal| appeal.vacate? }
 end
--- a/app/serializers/intake/appeal_serializer.rb
+++ b/app/serializers/intake/appeal_serializer.rb
@@ -9,4 +9,6 @@ class Intake::AppealSerializer < Intake::DecisionReviewSerializer
   attribute :form_type do
     "appeal"
   end
+  attribute :type
+  attribute :vacate_type, if: Proc.new { |appeal| appeal.vacate? }
 end
--- a/app/serializers/intake/appeal_serializer.rb
+++ b/app/serializers/intake/appeal_serializer.rb
@@ -10,5 +10,5 @@ class Intake::AppealSerializer < Intake::DecisionReviewSerializer
     "appeal"
   end
   attribute :type
-  attribute :vacate_type, if: Proc.new { |appeal| appeal.vacate? }
+  attribute :vacate_type, if: proc { |appeal| appeal.vacate? }
 end
--- a/spec/factories/appeal.rb
+++ b/spec/factories/appeal.rb
@@ -196,12 +196,14 @@ FactoryBot.define do
           appeal: appeal,
           parent: appeal.root_task,
           assigned_at: evaluator.active_task_assigned_at,
-          assigned_to: evaluator.associated_judge)
+          assigned_to: evaluator.associated_judge
+        )
         params = {
           disposition: "granted",
           vacate_type: "straight_vacate",
           instructions: "some instructions",
-          assigned_to_id: task.assigned_to.id}
+          assigned_to_id: task.assigned_to.id
+        }
         PostDecisionMotionUpdater.new(task, params).process
       end
     end
--- a/spec/factories/appeal.rb
+++ b/spec/factories/appeal.rb
@@ -189,7 +189,7 @@ FactoryBot.define do
     end
 
     trait :straight_vacated do
-      stream_type { "Vacate" }
+      stream_type { "vacate" }
 
       after(:create) do |appeal, evaluator|
         task = JudgeAddressMotionToVacateTask.create!(
--- a/spec/factories/appeal.rb
+++ b/spec/factories/appeal.rb
@@ -187,5 +187,23 @@ FactoryBot.define do
         ).call
       end
     end
+
+    trait :straight_vacated do
+      stream_type { "Vacate" }
+
+      after(:create) do |appeal, evaluator|
+        task = JudgeAddressMotionToVacateTask.create!(
+          appeal: appeal,
+          parent: appeal.root_task,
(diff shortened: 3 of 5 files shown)


app/models/appeal.rb
```diff
@@ -447,4 +453,10 @@ def maybe_create_translation_task
     distribution_task = tasks.open.find_by(type: DistributionTask.name)
     TranslationTask.create_from_parent(distribution_task) if STATE_CODES_REQUIRING_TRANSLATION_TASK.include?(state_code)
   end
+
+  # Non-vacate Appeals are not expected to have a PDM, but this method makes a
+  # best-effort attempt in either situation, and returns nil if none is found.
+  def post_decision_motion
+    PostDecisionMotion.where(task: tasks).first
```
nanotone: Fortunately, Rails can fetch this in one SQL roundtrip, but I'm still starting to think that it's awkward having the PDM associated with a task. I can see why we originally decided to do so, but in the new world of separate appeal streams, wouldn't it be better for the PDM to be associated with the appeal instead?

app/models/appeal.rb
```diff
@@ -26,9 +26,9 @@ class Appeal < DecisionReview
   has_many :record_synced_by_job, as: :record
 
   enum stream_type: {
-    "Original": "Original",
-    "Vacate": "Vacate",
-    "De Novo": "De Novo"
+    "original": "Original",
+    "vacate": "Vacate",
```
nanotone: @leikkisa My understanding is that this is what you're intending in #13325 as well. I went ahead and switched to the new style here, because I use `vacate?` twice in this PR (and git is good about merging identical hunks) but lemme know if you intended something different so I can follow suit.

spec/factories/appeal.rb
```diff
@@ -187,5 +187,23 @@
...
+          appeal: appeal,
+          parent: appeal.root_task,
+          assigned_at: evaluator.active_task_assigned_at,
+          assigned_to: evaluator.associated_judge)
+        params = {
+          disposition: "granted",
+          vacate_type: "straight_vacate",
+          instructions: "some instructions",
+          assigned_to_id: task.assigned_to.id}
+        PostDecisionMotionUpdater.new(task, params).process
```
nanotone: @jcq: I hope I'm using the `PostDecisionMotionUpdater` more-or-less correctly here! It's missing a few keys in the params, but seems to be happy with what I gave it.

app/models/appeal.rb
jcq: I'd be interested in @leikkisa thoughts on this. That change does seem to make more sense, though would it be associated with the original appeal? The new stream? Both?

spec/factories/appeal.rb
jcq: I think that looks right... 

Is `:straight_vacated` the only trait we're adding to the factory? What about other vacate types?

nanotone: Other vacate types are absolutely welcome in the factory. I just didn't need them for the tests in this particular PR, and am not a fan of having unused code lying around. Do you think they'd be a trivial addition? My gut sense, especially on De Novo, was no.

app/models/appeal.rb
nanotone: My thought was that it'd be associated with the new Vacate stream, so that the three Grant lanes all have the PDM in the same place. And for a post-MTV motion type, it would probably feel more natural to have each PDM associated with the stream that branched off of the original appeal (rather than lots of PDMs associated with a single Original appeal).

Definitely just spitballin here, though.

app/serializers/intake/appeal_serializer.rb
```diff
@@ -9,4 +9,6 @@ class Intake::AppealSerializer < Intake::DecisionReviewSerializer
   attribute :form_type do
     "appeal"
   end
+  attribute :type
```
leikkisa: This serializer is used for Intake. I don't think it's the same serializer that's used for Queue since this serializer is also relatively new.  I think this is the right one, and it has "type": `caseflow/app/models/serializers/work_queue/appeal_serializer.rb`

app/serializers/intake/appeal_serializer.rb
```diff
@@ -9,4 +9,6 @@ class Intake::AppealSerializer < Intake::DecisionReviewSerializer
...
+  attribute :vacate_type, if: proc { |appeal| appeal.vacate? }
```
leikkisa: It might be best to include this attribute whether or not it's a vacate stream (and have it be nil if it's not), depending on how it's used in the front end components. Will we want to check if it exists before using it?

spec/factories/appeal.rb
```diff
@@ -187,5 +187,25 @@
...
       end
     end
+
+    trait :straight_vacated do
+      stream_type { "Vacate" }
+
+      after(:create) do |appeal, evaluator|
+        task = JudgeAddressMotionToVacateTask.create!(
+          appeal: appeal,
+          parent: appeal.root_task,
```
leikkisa: I think this is not the root_task currently, although we could discuss.  The address motion to vacate task does look at its parent for some things. It might end up being be necessary to make the parent MTV mail task then.

spec/factories/appeal.rb
leikkisa: I think it might be good for us to have a conversation as to whether the PostDecisionMotion should save an appeal_id instead of a task_id. Then you wouldn't have to mock out the whole previous task structure (the MTV mail task and the Address MTV task), you could theoretically create the original appeal, the new appeal, and the post decision motion.

app/models/appeal.rb
```diff
@@ -447,4 +453,10 @@ def maybe_create_translation_task
     distribution_task = tasks.open.find_by(type: DistributionTask.name)
     TranslationTask.create_from_parent(distribution_task) if STATE_CODES_REQUIRING_TRANSLATION_TASK.include?(state_code)
   end
+
+  # Non-vacate Appeals are not expected to have a PDM, but this method makes a
+  # best-effort attempt in either situation, and returns nil if none is found.
+  def post_decision_motion
+    PostDecisionMotion.find_by(task: tasks)
```
leikkisa: I don't think this will work on the new appeal stream, because the task on the PostDecisionMotion is on the Original appeal.

app/models/appeal.rb
```diff
@@ -84,6 +84,12 @@ def type
     stream_type || "Original"
   end
 
+  def vacate_type
+    return nil unless vacate?
```
leikkisa: This works. It won't give us the vacate_type for on the "De Novo" stream, but we won't need it there either since the stream_type will say it all.

There's a separate conversation we should have about accessing the PDM, which might be useful for us to be able to access from all of the appeal streams.

app/serializers/intake/appeal_serializer.rb
nanotone: That would explain why we were seeing different fields getting serialized. Doesn't explain why the codebase has serializers in all sorts of different locations though 😕

client/app/queue/utils.js
```diff
@@ -119,7 +119,7 @@ const appealAttributesFromRawTask = (task) => ({
   externalId: task.attributes.external_appeal_id,
   docketName: task.attributes.docket_name,
   isLegacyAppeal: task.attributes.docket_name === 'legacy',
-  caseType: task.attributes.case_type,
+  caseType: _.capitalize(task.attributes.case_type),
```
Sjones352: @nanotone why are we capitalizing the case_type attribute?

nanotone: As part of this PR, I changed the appeal types to be lowercase instead of capitalized, so that Rails magic gives us the `Appeal#original?` and `Appeal#vacate?` methods -- this was modeled after what @leikkisa is doing in #13325.

nanotone: I decided to do this complicated lowercasing operation in a future tech-debt ticket, so this has been reverted.

app/serializers/intake/appeal_serializer.rb
nanotone: Good point; I've changed this attribute to be always present and sometimes-nil. Gonna go ahead and resolve this thread.

app/models/appeal.rb
nanotone: Punted in favor of #13358

app/models/appeal.rb
nanotone: We later decided that the PostDecisionMotion would be attached to the Vacate stream. This method will have to change anyway, but for this PR I'll leave it as-is.

spec/factories/appeal.rb
nanotone: I agree that this is un-ideal, but since it's just a test factory, I'd like to get this PR merged in and then use your updated PDMU to fix it, given what we decided about PDMs earlier.

nanotone:
@leikkisa @jcq I ended up not making any changes other than the obvious ones, because the remaining improvements will be much easier once
- the factory uses Sally's new PDMU, and
- we go ahead and tackle #13355.

Could I get a re-review, with that in mind?

app/models/appeal.rb
```diff
@@ -84,6 +84,16 @@ def type
     stream_type || "Original"
   end
 
+  def vacate?
```
leikkisa: I'm planning on downcasing this in my other PR, so you shouldn't need this method anymore since stream_type is enumerated.

app/models/appeal.rb
leikkisa: This method could be simplified to:
`post_decision_motion&.vacate_type`

Issue #10815, opened 2019-05-17
Difference of Opinion | Handle dispositions


leikkisa:
### Story
The basic point of a higher level review (HLR) is for a senior adjudicator to review a claim and potentially make a different determination based on the same evidence. In some cases, a higher level reviewer will determine that evidence is missing from a claim, even though it's not a DTA error. In such cases, the HLR reviewer will need to close out the HLR and establish a supplemental claim. The HLR reviewer will select a difference of opinion disposition that indicates the claim was closed and sent back due to a difference of opinion rather than a DTA error.
**Related**: https://github.com/department-of-veterans-affairs/caseflow/issues/12700

### Acceptance criteria
- [x] The disposition "Difference of Opinion" is added to the list of remand dispositions for a Decision Issue
- [x] Suggestion: Add a "remand type" to remand supplemental claims to differentiate between board remands, duty to assist errors, and difference of opinion
- [x] When an issue gets decided with the disposition "Difference of Opinion" on an HLR, Caseflow automatically creates a remand supplemental claim for those issues (similar to DTA errors).
- [x] If a decision review gets both DTA and DOO dispositions, those go onto separate EPs
- [x] The correct EP code is selected for each issue on the remand supplemental claim, and any 930 correction claims

### Codes and Claim Labels
Code | Claim Label
--- | ---
EP040AMADOR | 040 AMA Difference of Opinion - Rating
EP040AMADONR | 040 AMA Difference of Opinion - NR
EP040ADORPMC | 040 PMC AMA Difference of Opinion - Rating
EP040ADONRPMC | 040 PMC AMA Difference of Opinion - NR
930AMADOR | 930 AMA Difference of Opinion – Rating
930AMADONR | 930 AMA Difference of Opinion – NR
930DORPMC | 930 PMC AMA Difference of Opinion – Rating
930DONRPMC | 930 PMC AMA Difference of Opinion – NR

#### Resources/other links
https://dsva.slack.com/archives/CFBLH2LR4/p1571753040061000

### Technical notes
We consider board remands, DTA errors, and DoOs all "remand" supplemental claims, because remand means "to order back", and these are all claims that get sent back.

Some places to update end product codes:
- caseflow/app/models/end_product.rb
- caseflow/client/constants/END_PRODUCT_CODES.json
- caseflow/app/workflows/end_product_code_selector.rb

Other files to look at:
- DecisionIssue
- SupplementalClaim
- DecisionReview
[...]

PR #13349, opened 2020-02-05
Handle difference of opinion dispositions

leikkisa:
connects #10815

### Description
- Adds difference of opinion EP codes
  - Note: duplicates EP codes for 930 DOOs, for control, local quality, and national quality. We will remove the modal for that and restructure the EP codes in a future PR
- Adds "remand_type" to request issue
  - handles corrections of remand type by looking at the previous request issue

### User Facing Changes
N/A
 
### Documentation Updates

### Database Changes
N/A


app/workflows/end_product_code_selector.rb
```diff
@@ -79,25 +95,41 @@ class EndProductCodeSelector
...
-        dta: {
-          compensation: {
-            appeal: {
-              rating: "930AMARRC",
-              nonrating: "930AMARNRC"
+        remand: {
+          duty_to_assist: {
+            compensation: {
+              appeal: {
+                rating: "930AMARRC",
```
leikkisa: The 930 DOO codes are being duplicated right now for Control, Local Quality Error, and National Quality Error. I am asking the client about this because they didn't get broken out. I'd prefer to go ahead and put in duplicates now, and if we're not going to get them broken out, then handle that in a different PR (which would involve some front end changes as well).

https://dsva.slack.com/archives/CFBLH2LR4/p1580861034093800

app/models/decision_issue.rb
```diff
@@ -21,16 +21,26 @@ class DecisionIssue < ApplicationRecord
...
+  DIFFERENCE_OF_OPINION = "Difference of Opinion"
+  DTA_ERROR = "DTA Error"
+  DTA_ERROR_EXAM_MO = "DTA Error - Exam/MO"
   DTA_ERROR_FED_RECS = "DTA Error - Fed Recs"
   DTA_ERROR_OTHER_RECS = "DTA Error - Other Recs"
-  DTA_ERROR_EXAM_MO = "DTA Error - Exam/MO"
-  DTA_ERROR = "DTA Error"
-  REMAND = "remanded"
+  DTA_ERROR_PMR = "DTA Error - PMRs"
+  REMANDED = "remanded"
```
leikkisa: This was some alphabetizing from Tomas.

app/workflows/end_product_code_selector.rb
leikkisa: Client has confirmed that we don't need the correction type for these, but I want to remove those in a future PR because it's a different assumption around correction issues, and probably requires some more front-end and back-end work (and this doesn't hurt much for now, as most of these will be automatically generated and not from corrections).

msj0nes:
# Stakeholder Demo Request

Contact AMO when ready to demo (prior to deploy).

PR #13352, opened 2020-02-05
Add member_of_organization method to User model

ajspotts:
### Description
The User model was lacking a method to check if a user was a member of a particular organization.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -288,6 +288,10 @@ class User < ApplicationRecord
     orgs
   end
 
+  def member_of_organization?(org)
+    organizations.pluck(:id).include?(org&.id)
+  end
+
   def judge?
     !!JudgeTeam.for_judge(self) || judge_in_vacols?
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -289,7 +289,7 @@ class User < ApplicationRecord
   end
 
   def member_of_organization?(org)
-    organizations.pluck(:id).include?(org&.id)
+    organizations.include?(org)
   end
 
   def judge?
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -288,6 +288,10 @@ def selectable_organizations
     orgs
   end
 
+  def member_of_organization?(org)
+    organizations.pluck(:id).include?(org&.id)
```
pkarman: is the `id` comparison necessary? I would have thought this would also work:

`organizations.include?(org)`

ajspotts: it does! thanks for the catch, refactoring

hschallhorn:
Quick Q! What's the motivation behind using `user.member_of_organization?(org)` over `org.can_access?(user)`

ajspotts:
> Quick Q! What's the motivation behind using `user.member_of_organization?(org)` over `org.user_has_access?(user)`

the idea to write that method came about when Kat, Yoom & I were pairing during my SCM ticket and at some point there was a determination made that there seemed to me a need for that function on the User model.  That said if you find it to be redundant I’m happy to remove the change.
Also, apologies for not having some more specific context for you.  We also decided at the time that adding the function was outside the scope of the issue so I punted it until after the completion of that ticket.

spec/models/user_spec.rb
```diff
@@ -349,6 +349,26 @@
     end
   end
 
+  context "#member_of_organization?" do
+    let(:org) { create(:organization) }
```
yoomlam: How about adding another test for the case where `org=nil` ?

ajspotts: added!

Issue #13062, opened 2019-12-27
When establishing a board grant or remand claim, set the status to "RFD"


leikkisa:
## Description

In order to process board grants and board remands more quickly, we would like to set the status of the claim to "RFD" when it's established (instead of "PEND").

## Acceptance criteria
- [ ] The EndProduct.to_vbms_hash method sets the status type code to "RFD" if the end product code is for a board grant, or board remand
- [ ] Othewise, it sets the status code to "PEND". This is the default value in connect_vbms, so it doesn't have to be set here, but if it's not, the status code shouldn't be a key in the hash

## Technical notes
The board grant EP codes can be found in `caseflow/app/models/board_grant_effectuation.rb`

The board remand codes can be found in `caseflow/app/workflows/end_product_code_selector.rb`, in the tree where it is a DTA and for an appeal (such as 040BDERPMC).  See `EndProductEstablishment.rating?` for an example of automatically pulling these from the tree.

PR #13361, opened 2020-02-06
When establishing a board grant or remand claim, set the status to "RFD"

Sjones352:
Connects  #13062
Screen Shot 2020-02-06 at 12 29 26 AM


app/models/end_product.rb
```diff
@@ -238,10 +239,19 @@ def to_vbms_hash
...
       claimant_participant_id: claimant_participant_id,
       limited_poa_code: limited_poa_code,
-      limited_poa_access: limited_poa_access
+      limited_poa_access: limited_poa_access,
+      status_type_code: set_status_type_code
     }
   end
 
+  def set_status_type_code
+    if BoardGrantEffectuation::END_PRODUCT_CODES || EndProductCodeSelector::END_PRODUCT_CODES
```
leikkisa: Here, you'll want to check if the `BoardGrantEffectuation::END_PRODUCT_CODES` includes the claim_type_code, although here in `to_vbms_hash` it will make more sense to include the end product attribute directly (so instead of putting a method here:
`status_type_code: set_status_type_code`, it'd be more like `status_type_code: status_type_code`. And when you're initializing the End Product, like in the EndProductEstablishment model or elsewhere, that'll be where you want to determine whether it's set to "RFD" or "PEND".

We're going to also want to include the board remand codes:
    "040BDENR" => "Board DTA Error - Non-Rating",
    "040BDER" => "Board DTA Error - Rating",
    "040BDENRPMC" => "PMC Board DTA Error - Non-Rating",
    "040BDERPMC" => "PMC Board DTA Error - Rating"

These are currently mixed in with the DTA codes in EndProduct, so it'll make sense to separate those out.

app/models/end_product_establishment.rb
```diff
@@ -459,10 +459,21 @@ def end_product_with_modifier(the_modifier = nil)
       gulf_war_registry: false,
       station_of_jurisdiction: station,
       limited_poa_code: limited_poa_code,
-      limited_poa_access: limited_poa_access
+      limited_poa_access: limited_poa_access,
+      status_type_code: set_status_type_code
     )
   end
 
+  def set_status_type_code
```
leikkisa: I'd be okay just calling this "status_type_code". The method itself isn't technically "setting" anything, just returning a value.

leikkisa: Up to you, but do you think we should add a small comment that EPs originating from the board are set to "Ready for Decision" in order to expedite processing?

spec/models/end_product_establishment_spec.rb
```diff
@@ -315,7 +318,8 @@
             gulf_war_registry: false,
             suppress_acknowledgement_letter: false,
             limited_poa_code: "ABC",
-            limited_poa_access: true
+            limited_poa_access: true,
+            status_type_code: "RFD"
```
leikkisa: This one should be PEND

app/models/end_product_establishment.rb
leikkisa: Can you add a model test for this?

app/models/end_product_establishment.rb
```diff
@@ -459,10 +469,13 @@ def end_product_with_modifier(the_modifier = nil)
       gulf_war_registry: false,
       station_of_jurisdiction: station,
       limited_poa_code: limited_poa_code,
-      limited_poa_access: limited_poa_access
+      limited_poa_access: limited_poa_access,
+      status_type_code: status_type_code
     )
   end
 
+  # In order to expedite processing, EPs originating from the board are set to "Ready for Decision"
```
leikkisa: I would move this to above the "status_type" method, that's where we usually put comments that are adding context to the method.

Issue #10732, opened 2019-05-13
Retry VBMSError::FilenumberDoesNotExist with claim number?


monfresh:
A common VBMS error we see in Sentry is `VBMSError::FilenumberDoesNotExist`. I see that in one case (in [fetch_documents_for](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/services/external_api/vbms_service.rb#L43)), we are rescuing the error and trying an alternative claim number instead of the SSN. I’m wondering if we should apply the same logic to other methods, like `upload_document_to_vbms`.

From the description of the [PR that added the rescue](https://github.com/department-of-veterans-affairs/caseflow/pull/7755), it sounds like we should be fetching the correct number from BGS every time we send over a potentially incorrect file number:
```
In some situations the file number in VACOLS (the bfcorlid column) is incorrect. 
It will be the SSN where it should be a claim number. When hitting the VBMS 
endpoint to get a list of files, we use the wrong number, and get an error with 
text: "File Number does not exist within the system." To solve this problem we 
can make a call to BGS to find the correct file number and then call VBMS with 
that instead.
```

cc @sharonwarner  and @sunil who participated in the original [issue](https://github.com/department-of-veterans-affairs/caseflow/issues/4567)

lomky:
This is happening a lot over the past month.

hschallhorn:
[Also this dude](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/5059/
)

pkarman:
some stats:

```
irb(main):011:0> Veteran.where("file_number = ssn").count
[2019-10-30 11:58:55 -0400]    (510.4ms)  SELECT COUNT(*) FROM "veterans" WHERE (file_number = ssn)
=> 141828
irb(main):012:0> Veteran.where("file_number != ssn").count
[2019-10-30 11:59:01 -0400]    (83.9ms)  SELECT COUNT(*) FROM "veterans" WHERE (file_number != ssn)
=> 34362
irb(main):015:0> Veteran.count
[2019-10-30 11:59:32 -0400]    (72.0ms)  SELECT COUNT(*) FROM "veterans"
=> 485479
irb(main):017:0> Veteran.where("char_length(file_number) = 9").count
[2019-10-30 12:00:43 -0400]    (1089.7ms)  SELECT COUNT(*) FROM "veterans" WHERE (char_length(file_number) = 9)
=> 380899
[...]

PR #13364, opened 2020-02-06
Update ExternalApi::VBMSService#fetch_document_series_for to fallback to BGS claim ID 

ferristseng:
Related to #10732 

### Description

  * Refactors retry logic to base class
  * Add `VbmsDocumentSeriesForAppeal` to retry VBMS call using BGS claim id

### DIFF
--- a/app/services/external_api/vbms_document_series_for_appeal.rb
+++ b/app/services/external_api/vbms_document_series_for_appeal.rb
@@ -12,7 +12,7 @@ class ExternalApi::VbmsDocumentSeriesForAppeal < ExternalApi::VbmsRequestWithFil
       service = VBMS::Service::PagedDocuments.new(client: vbms_client)
 
       ExternalApi::VBMSService.call_and_log_service(
-        service: service, 
+        service: service,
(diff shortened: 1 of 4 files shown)


pkarman:
See also https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/1137

app/services/external_api/vbms_document_series_for_appeal.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class ExternalApi::VbmsDocumentSeriesForAppeal < ExternalApi::VbmsRequestWithFileNumber
+  def fetch
+    request_with_retry
+  end
+
+  protected
+
+  def do_request(file_number_or_claim_number)
```
pkarman: we should name this variable `ssn_or_claim_number` or simply `file_number`.

A file number may be either a SSN or claim number.

ferristseng: Done!

PR #13365, opened 2020-02-06
Add more detail to task event summary

pkarman:
Observed while in production console with:

```ruby
appeal.history.summary { |ev| puts ev }
```

Added the task type/id for better detailing.


app/models/task_event.rb
```diff
@@ -11,12 +11,20 @@ def diff
   end
 
   def who
-    User.find(version.whodunnit)
+    version.whodunnit ? User.find(version.whodunnit) : User.new
+  end
+
+  def original
+    @original ||= reified
```
leikkisa: TIL a new word! Although still not sure what it means in code :-P.

`reified` is undefined for me and it looks like that's happening in rspec too. `version.reify` does something, not sure what the desire behavior is though.

pkarman: gah! my typo. `version.reify` is what I wanted.

Issue #12554, opened 2019-11-01
Monthly Metric: Caseflow certifications percentage of total certifications


alisan16:
This metric is required by OIT (Metric ID: 1709076050).

Metric description: Caseflow - Percentage of all cases certified with Caseflow, which assists Agency Original Jurisdiction in uploading all required documents of record to a shared electronic file repository and automatically prepares the Certification of Appeal to transfer the appeal to the Board of Veterans’ Appeals.

Metric definition: (Number of appeals transferred to the Board using Caseflow/Total number of appeals transferred to the Board (manual certifications of paper appeals and Caseflow certifications))

alisan16:
Certification usage notebook from 2017: https://department-of-veterans-affairs.github.io/appeals-data/notebooks/intake/certification-usage.nb.html

pkarman:
Background: https://github.com/department-of-veterans-affairs/appeals-data/blob/master/notebooks/intake/certification-usage.Rmd

pkarman:
#13082 implemented this

documented in https://github.com/department-of-veterans-affairs/caseflow/wiki/Writing-DB-migrations

alisan16:
Based on the comment, I think this was closed on accident. Re-opening cc @pkarman

pkarman:
These ROs are not listed in the [list of ROs](https://github.com/department-of-veterans-affairs/appeals-data/blob/master/data/ro.csv) that the older reporting code used to identify [how VACOLS certifications were measured](https://github.com/department-of-veterans-affairs/appeals-data/blob/master/notebooks/intake/certification-usage.Rmd#L25).

```
RO61 Houston Foreign Cases
RO64 Columbia Fiduciary Hub
RO65 Indianapolis Fiduciary Hub
RO66 Lincoln Fiduciary Hub
RO67 Louisville Fiduciary Hub
RO68 Milwaukee Fiduciary Hub
RO69 Western Area Fiduciary Hub
RO70 Louisville CLCW
RO78 Unknown
RO79 St. Paul Regional Loan Center
RO97 ARC
```

Do they matter? is the list from the original code just out of date?

pkarman:
Run in production:

```
irb(main):142:0> sept = Metrics::DateRange.new('2019-09-01', '2019-09-30')

irb(main):143:0> octo = Metrics::DateRange.new('2019-10-01', '2019-10-31')

[...]

Issue #12748, opened 2019-11-19
Caseflow OIT Metrics


alisan16:
## Background/context
We are required to deliver these 7 metrics to OIT on the scheduled reporting frequency. Metrics were approved in June 2019. 

Metrics listed on this [VA Sharepoint site](https://dvagov.sharepoint.com/:x:/r/sites/epmoepmd/benefits/_layouts/15/guestaccess.aspx?e=sIhwv4&share=EUgfXGIOUjNDgl67Ni70fVcBr5qddsswVWjhnbI3kC9DCw)

Metric ID | 2. Metric Description | 3. Unit of Measure | 4. Performance Measurement Category Mapping | 5. Agency Baseline Capability | 6. 2018 Target | 7. 2019 Target | 8. Measurement Condition | 9. Reporting Frequency
-- | -- | -- | -- | -- | -- | -- | -- | --|
1909040936 | Caseflow - Percentage of all cases certified with Caseflow, which assists Agency Original Jurisdiction in uploading all required documents of record to a shared electronic file repository and automatically prepares the Certification of Appeal to transfer the appeal to the Board of Veterans’ Appeals. | (Number of appeals transferred to the Board using Caseflow/Total number of appeals transferred to the Board (manual certifications of paper appeals and Caseflow certifications)) | Strategic and Business Results | 85.7 | 90 | 90 | Over target | Monthly
1909040937 | Caseflow Dispatch - Percent of non-denial decisions with an EP created within 7 days. Caseflow facilitates the transfer of cases to the Regional Office following issuance of a Board of Veterans’ Appeals decision, with an end product (EP) created to track work. Not creating an EP increases the possibility of miscommunication and slows average processing times, delaying ratings adjustment and working of remand orders. | (Number EPs with date<=7 days after outcoding date/number non-denial decisions) | Strategic and Business Results | 79.6 | 80 | 80 | Over target | Quarterly
1909040938 | Increase Veteran show rates at Board hearings using Caseflow Hearing Schedule, which enables a standardized scheduling process providing the ability to schedule hearings at both regional offices and alternate locations closer to Veterans’ homes. | (Total scheduled meetings held – postposed hearings/total scheduled meetings from Caseflow Hearing Schedule) | Strategic and Business Results |   | N/A | 75 | Over target | Quarterly
[...]

PR #13367, opened 2020-02-06
Metric: Certifications via Caseflow

pkarman:
connects #12554 

### Description
OIT metric for Certifications via Caseflow.

### Testing Plan
1. Monkeypatch and run in production console for verifying logic.

 
### Documentation Updates

### DIFF
--- a/app/models/vacols/case.rb
+++ b/app/models/vacols/case.rb
@@ -315,6 +315,14 @@ class VACOLS::Case < VACOLS::Record
     # :nocov:
   end
 
+  def paperless?
+    folder.tivbms == "Y"
+  end
+
+  def certified_with_caseflow?
+    bfdcertool.present?
+  end
+
   # The attributes that are copied over when the case is cloned because of a remand
   def remand_clone_attributes
     slice(
--- a/app/services/metrics/certification_usage.rb
+++ b/app/services/metrics/certification_usage.rb
@@ -15,7 +15,8 @@ class Metrics::CertificationUsage < Metrics::Base
       certified_total: certified_total,
       certified_paperless: certified_paperless,
       certified_with_caseflow: certified_with_caseflow,
-      metric: percent(certified_with_caseflow, certified_total)
+      total_metric: percent(certified_with_caseflow, certified_total),
+      paperless_metric: percent(certified_with_caseflow, certified_paperless)
     }
   end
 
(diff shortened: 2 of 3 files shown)


app/services/metrics/certification_usage.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def certified_paperless
+    @certified_paperless ||= certifications.select(&:paperless?).count
+  end
+
+  def certified_with_caseflow
+    @certified_with_caseflow ||= certifications.select(&:certified_with_caseflow?).count
+  end
+
+  def regional_office_codes
```
pkarman: TBD whether this is needed or not

app/models/vacols/case.rb
```diff
@@ -315,6 +315,14 @@ def remand_return_date(vacols_ids)
     # :nocov:
   end
 
+  def paperless?
+    folder.tivbms == "Y"
+  end
+
+  def certified_with_caseflow?
```
pkarman: https://dsva.slack.com/archives/C2ZAMLK88/p1485890347000118

app/services/metrics/certification_usage.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+# Metric ID: 1709076050
```
alisan16: Can you update the metric ID to the new ID here: https://github.com/department-of-veterans-affairs/caseflow/issues/12748 I think 17### was from the old contract.

spec/services/metrics/certification_usage_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+        vacols_case = create(:case, :type_original, bf41stat: 3.days.ago)
+        vacols_case.folder.update!(tivbms: "N")
+      end
+    end
+
+    it "reports total and paperless metrics" do
+      expect(subject).to eq(
+        certified_total: 25,
+        certified_paperless: 15,
+        certified_with_caseflow: 15,
```
alisan16: How is this metric 15? Can paper cases be certified with caseflow?

pkarman: I added some comments.

Note that for the sake of the code/tests, we are just measuring what's in VACOLS. In a practical real-world sense, no, paper cases would not be certified with caseflow, but afaik nothing prevents that from happening, technically, and VACOLS could reflect that situation if it happened.

PR #13379, opened 2020-02-10
show searchable dropdown error in an error banner instead of as a validation error

### LINKED ISSUE #11814: style error shown when a searchable dropdown component with dynamically-loaded contents is unable to load
Currently if a searchable dropdown component with dynamically-loaded contents fails to load, the error is displayed in the same format as a validation error:

current-error

Instead, as per @geronimoramos we should display such errors in an alert below the dropdown, like this:

new-error

[Related slack thread](https://dsva.slack.com/archives/C2ZAMLK88/p1565895805366700)

_The error seen above can be reproduced in the "but facilities api throws an error" test in `spec/feature/hearings/schedule_veteran/build_hearsched_spec.rb` in the `master` branch_

tomas-nava:
Resolves #11814

### Description
Styles service error messages on the appeal hearing locations searchable dropdown in an alert banner instead of as a validation error.

### How to review
Check out the branch and run 
```
bundle exec rspec spec/feature/hearings/schedule_veteran/build_hearsched_spec.rb:144
```

longer-alert

Regular validation errors are still displayed as expected

validation-error


client/app/components/DataDropdowns/AppealHearingLocations.jsx
```diff
@@ -126,18 +127,23 @@ class AppealHearingLocationsDropdown extends React.Component {
     const {
       name, label, onChange, readOnly, placeholder, errorMessage,
       appealHearingLocations: { isFetching, options, errorMsg } } = this.props;
+    const validationErrorMessage = errorMessage;
+    const serviceErrorMessage = errorMsg;
```
ferristseng: Nice! This naming makes it a lot clearer

Issue #13355, opened 2020-02-05
Motion to Vacate | Associate PostDecisionMotion to Appeal


nanotone:
## Description

Change the PostDecisionMotion schema so that it is associated with an appeal with type "Vacate", not a task.

## Acceptance criteria

- [ ] ~~Remove post_decision_motions.task_id~~ 
- [ ] Add post_decision_motions.appeal_id
- [ ] For motions to vacate, the associated appeal should be the "Vacate" stream

## Context

The PostDecisionMotion is currently associated with a task, because until recently, MTV-related tasks were all part of the task tree of the original appeal. The switch to creating a separate appeal stream with type "Vacate" now makes it easier to tell which tasks are related to the original appeal and which are related to MTV. The new appeal stream is therefore a more convenient object for the PostDecisionMotion to reference.

## Technical notes

A non-exhaustive list of places in the codebase to update:
- DB migration to remove the old column and add the new
- Appeal, wherever it deals with a PostDecisionMotion
- PostDecisionMotion and PostDecisionMotionUpdater

PR #13386, opened 2020-02-11
Associate PostDecisionMotion with appeal rather than task

nanotone:
Connects #13355 

### Description
This PR adds an `appeal_id` column to the PostDecisionMotion table and model. The logic surrounding the relationship between appeals/tasks and PDMs is cleaned up accordingly.

### Acceptance Criteria
- [ ] Add `post_decision_motions.appeal_id`
- [ ] For motions to vacate, the associated appeal should be the "Vacate" stream
- ~~Rails migrate suggested removing post_decision_motions.task_id in a separate migration.~~ We decided not to remove it after all.

### Testing Plan
1. Updated tests to match new association, and corrected the factory for appeals with a vacate stream
2. Manually tested migrate, rollback, and re-migrate for 4 PDMs, one with each disposition.

### Database Changes
*Only for Schema Changes*

### DIFF
--- a/app/models/appeal.rb
+++ b/app/models/appeal.rb
@@ -23,6 +23,7 @@ class Appeal < DecisionReview
   has_many :remand_supplemental_claims, as: :decision_review_remanded, class_name: "SupplementalClaim"
 
   has_one :special_issue_list
+  has_one :post_decision_motion
   has_many :record_synced_by_job, as: :record
 
   enum stream_type: {
@@ -98,6 +99,8 @@ class Appeal < DecisionReview
   end
 
   def vacate_type
+    return nil unless vacate?
+
     post_decision_motion&.vacate_type
   end
 
@@ -463,11 +466,5 @@ class Appeal < DecisionReview
   ensure
     distribution_task = tasks.open.find_by(type: DistributionTask.name)
     TranslationTask.create_from_parent(distribution_task) if STATE_CODES_REQUIRING_TRANSLATION_TASK.include?(state_code)
-  end
-
-  # Non-vacate Appeals are not expected to have a PDM, but this method makes a
-  # best-effort attempt in either situation, and returns nil if none is found.
-  def post_decision_motion
-    PostDecisionMotion.find_by(task: tasks)
   end
 end
--- a/app/models/post_decision_motion.rb
+++ b/app/models/post_decision_motion.rb
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 
 class PostDecisionMotion < ApplicationRecord
-  belongs_to :task, optional: false
+  self.ignored_columns = ["task_id"]
+
+  belongs_to :appeal, optional: false
 
   validates :disposition, presence: true
   validate :vacate_type_is_present_if_granted
--- a/app/models/post_decision_motion.rb
+++ b/app/models/post_decision_motion.rb
@@ -1,9 +1,8 @@
 # frozen_string_literal: true
 
 class PostDecisionMotion < ApplicationRecord
-  self.ignored_columns = ["task_id"]
-
-  belongs_to :appeal, optional: false
+  belongs_to :appeal
+  belongs_to :task, optional: false
 
   validates :disposition, presence: true
   validate :vacate_type_is_present_if_granted
--- a/app/workflows/post_decision_motion_updater.rb
+++ b/app/workflows/post_decision_motion_updater.rb
@@ -7,6 +7,8 @@
 class PostDecisionMotionUpdater
   include ActiveModel::Model
 
+  class InvalidMotionUpdaterArguments < StandardError; end
+
   attr_reader :task, :params, :disposition, :instructions
 
   def initialize(task, params)
@@ -16,20 +18,24 @@ class PostDecisionMotionUpdater
     @instructions = @params[:instructions]
   end
 
-  delegate :appeal, to: :task
+  delegate :appeal, to: :task, prefix: "original"
 
   def process
     ActiveRecord::Base.transaction do
-      return unless post_decision_motion
+      if post_decision_motion
+        handle_denial_or_dismissal
+        handle_grant
+      end
 
-      handle_denial_or_dismissal
-      handle_grant
-
-      return if errors.messages.any?
+      raise InvalidMotionUpdaterArguments if errors.messages.any?
 
       task.update(status: Constants.TASK_STATUSES.completed)
       post_decision_motion
     end
+  rescue InvalidMotionUpdaterArguments
+    # Slightly leaky abstraction: vacate_stream may have already been created and
+    # committed to DB before we realize the transaction should be rolled back.
+    @vacate_stream&.destroy!
   end
 
   private
@@ -38,9 +44,23 @@ class PostDecisionMotionUpdater
     @post_decision_motion ||= create_motion
   end
 
+  def vacate_stream
+    return nil unless grant_type?
+
+    @vacate_stream ||= original_appeal.create_stream(:vacate)
+  end
+
+  def appeal
+    if denied_or_dismissed?
+      original_appeal
+    elsif grant_type?
+      vacate_stream
+    end
+  end
+
   def create_motion
     motion = PostDecisionMotion.new(
-      task: task,
+      appeal: appeal,
       disposition: disposition,
       vacate_type: params[:vacate_type]
     )
@@ -49,7 +69,7 @@ class PostDecisionMotionUpdater
       motion.vacated_decision_issue_ids = params[:vacated_decision_issue_ids]
     elsif disposition == "granted"
       # For full grant, auto populate all decision issue IDs
-      motion.vacated_decision_issue_ids = appeal.decision_issues.map(&:id)
+      motion.vacated_decision_issue_ids = original_appeal.decision_issues.map(&:id)
     end
 
     unless motion.valid?
@@ -85,7 +105,6 @@ class PostDecisionMotionUpdater
   def handle_grant
     return unless grant_type?
 
-    vacate_stream = appeal.create_stream(:vacate)
     create_new_stream_tasks(vacate_stream)
     post_decision_motion.create_request_issues_for_vacatur
     post_decision_motion.create_vacated_decision_issues
--- a/app/workflows/post_decision_motion_updater.rb
+++ b/app/workflows/post_decision_motion_updater.rb
@@ -22,12 +22,8 @@ class PostDecisionMotionUpdater
 
   def process
     ActiveRecord::Base.transaction do
-      if post_decision_motion
-        handle_denial_or_dismissal
-        handle_grant
-      end
-
-      raise InvalidMotionUpdaterArguments if errors.messages.any?
+      create_new_models
+      fail InvalidMotionUpdaterArguments if errors.messages.any?
 
       task.update(status: Constants.TASK_STATUSES.completed)
       post_decision_motion
@@ -35,7 +31,7 @@ class PostDecisionMotionUpdater
   rescue InvalidMotionUpdaterArguments
     # Slightly leaky abstraction: vacate_stream may have already been created and
     # committed to DB before we realize the transaction should be rolled back.
-    @vacate_stream&.destroy!
+    vacate_stream&.destroy!
   end
 
   private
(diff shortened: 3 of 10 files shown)


app/models/appeal.rb
```diff
@@ -98,6 +99,8 @@ def create_stream(stream_type)
   end
 
   def vacate_type
+    return nil unless vacate?
```
nanotone: I slotted this back in, because denied/dismissed PDMs don't have a fresh stream and are attached to the Original appeal.
Also, if we have future motions, they might not all be related to vacatur.

app/models/post_decision_motion.rb
```diff
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 
 class PostDecisionMotion < ApplicationRecord
-  belongs_to :task, optional: false
+  self.ignored_columns = ["task_id"]
```
nanotone: This was suggested by Rails migrate as part of a two-phase migration, so that we're not moving PDMs to appeals and removing `task_id` all at once.

See, e.g. https://withatwist.dev/backward-compatible-database-migrations.html

app/workflows/post_decision_motion_updater.rb
```diff
@@ -16,19 +18,23 @@ def initialize(task, params)
...
 
-      return if errors.messages.any?
+      raise InvalidMotionUpdaterArguments if errors.messages.any?
 
       task.update(status: Constants.TASK_STATUSES.completed)
     end
+  rescue InvalidMotionUpdaterArguments
+    # Slightly leaky abstraction: vacate_stream may have already been created and
+    # committed to DB before we realize the transaction should be rolled back.
+    @vacate_stream&.destroy!
```
nanotone: I'm not in love with the entirety of this, but I do think it's worthwhile to accept a bit of uglyness here so that we can have
- a nice and clean `vacate_stream` appeal, which happens to use a separate transaction to create-if-necessary
- a stronger guarantee of a no-op in case any of the arguments were invalid. (I test for this too)

I'm open to any and all ideas on tidier ways to achieve this.

db/schema.rb
```diff
@@ -906,7 +906,7 @@
   create_table "people", force: :cascade do |t|
     t.datetime "created_at", null: false
     t.date "date_of_birth"
-    t.string "email_address"
+    t.string "email_address", comment: "Person email address, cached from BGS"
```
nanotone: This seems to have been some unrelated untidiness from a manual merge coming out of code freeze. Maybe the new Rails version finally picked up on it.

app/models/appeal.rb
leikkisa: That's fine, it seems like it's helpful for readability. I think my previous comment was more because it would return nil anyway if there's not a post decision motion or a vacate type on the post decision motion.

app/workflows/post_decision_motion_updater.rb
leikkisa: This doesn't need the @ here. I can't remember if I checked if the vacate stream was still created when the transaction failed before, but I think your idea makes sense.

app/workflows/post_decision_motion_updater.rb
```diff
@@ -7,6 +7,8 @@
 class PostDecisionMotionUpdater
   include ActiveModel::Model
 
+  class InvalidMotionUpdaterArguments < StandardError; end
+
   attr_reader :task, :params, :disposition, :instructions
 
   def initialize(task, params)
```
jcq: Are we leaving task here for backward compatibility? Why not update the initialization params to just take the appeal?

app/models/appeal.rb
```diff
@@ -23,6 +23,7 @@ class Appeal < DecisionReview
   has_many :remand_supplemental_claims, as: :decision_review_remanded, class_name: "SupplementalClaim"
 
   has_one :special_issue_list
+  has_one :post_decision_motion
```
leikkisa: Can you add the `belongs_to` to PostDecisionMotion as well?

spec/factories/appeal.rb
```diff
@@ -188,23 +190,27 @@
...
+    trait :with_straight_vacate_stream do
       after(:create) do |appeal, evaluator|
-        task = JudgeAddressMotionToVacateTask.create!(
+        mail_task = VacateMotionMailTask.create!(
           appeal: appeal,
           parent: appeal.root_task,
           assigned_at: evaluator.active_task_assigned_at,
           assigned_to: evaluator.associated_judge
         )
+        addr_task = JudgeAddressMotionToVacateTask.create!(
```
leikkisa: I think we talked about it before and decided not to worry about it for the sake of test setup, but the assigned_to on the task above would be a motions attorney. Not a request, just mentioning it in case I'm wrong about our previous conversation.

spec/workflows/post_decision_motion_updater_spec.rb
```diff
@@ -104,10 +104,12 @@
         let(:vacate_type) { "vacate_and_de_novo" }
         let(:assigned_to_id) { nil }
 
-        it "should not create an attorney task" do
+        it "should not create a motion, vacate stream, or attorney task" do
           subject.process
           expect(subject.errors[:assigned_to].first).to eq "can't be blank"
           expect(task.reload.status).to eq Constants.TASK_STATUSES.in_progress
+          expect(Appeal.vacate.count).to eq 0
```
leikkisa: Good addition!

app/workflows/post_decision_motion_updater.rb
```diff
@@ -37,9 +43,23 @@ def post_decision_motion
...
+      original_appeal
+    elsif grant_type?
+      vacate_stream
+    end
+  end
+
   def create_motion
     motion = PostDecisionMotion.new(
-      task: task,
+      appeal: appeal,
```
leikkisa: It makes sense that we would want a way to connect to the original appeal for denied/dismissed flows, when looking at the data, otherwise the PDM is kind of floating. We missed that case in our conversation.

I wonder if it makes sense to keep the task_id instead. If there were multiple post decision motions on the appeal, that would make it easier to connect to what originated it. We didn't have to worry about that when this was just for the new vacate stream, since every (granted) post decision motion would result in a new vacate stream.

Too bad we don't know too much about the other PostDecisionMotions yet.

app/workflows/post_decision_motion_updater.rb
nanotone: This was originally a suggestion from @leikkisa, but after consciously thinking about it (thanks!) I think it makes sense:
- The PDMU makes use of the original task, to set its status to complete and to create its sibling abstract task in case of deny/dismiss.
- Thinking mechanically/ergonomically, the task originated the PDMU, even if the appeal "owns" it.

**But** that reminds me, I do need to make sure the abstract task belongs to the new appeal when vacatur is granted.

app/workflows/post_decision_motion_updater.rb
```diff
@@ -37,9 +43,23 @@ def post_decision_motion
...
+    @vacate_stream ||= original_appeal.create_stream(:vacate)
+  end
+
+  def appeal
+    if denied_or_dismissed?
+      original_appeal
+    elsif grant_type?
+      vacate_stream
+    end
+  end
```
leikkisa: If you decide to use the new field just for the vacate stream, you won't need this method anymore.

app/models/post_decision_motion.rb
```diff
@@ -1,7 +1,9 @@
...
+
+  belongs_to :appeal, optional: false
```
leikkisa: Depending on what we decide to store in the new field, this might not be required.

db/migrate/20200206161309_link_motions_to_appeals.rb
```diff
@@ -0,0 +1,43 @@
+class LinkMotionsToAppeals < ActiveRecord::Migration[5.1]
+  def up
+    # Index creation and task_id removal will be a separate Caseflow::Migration
+    add_reference :post_decision_motions, :appeal, foreign_key: true, index: false
+    safety_assured do
+      execute <<-EOS.strip_heredoc
```
leikkisa: This will also be impacted by the decision.

spec/workflows/post_decision_motion_updater_spec.rb
```diff
@@ -116,10 +118,12 @@
         let(:vacate_type) { nil }
         let(:assigned_to_id) { attorney.id }
 
-        it "should not create an attorney task" do
+        it "should not create a motion, vacate stream, or attorney task" do
           subject.process
           expect(subject.errors[:vacate_type].first).to eq "is required for granted disposition"
           expect(task.reload.status).to eq Constants.TASK_STATUSES.in_progress
+          expect(Appeal.vacate.count).to eq 0
+          expect(PostDecisionMotion.count).to eq 0
```
leikkisa: There's probably somewhere in the test where you add a check that the post decision motion is connected to the vacate_stream, maybe in verify_vacate_stream

app/workflows/post_decision_motion_updater.rb
```diff
@@ -16,19 +18,23 @@ def initialize(task, params)
     @instructions = @params[:instructions]
   end
 
-  delegate :appeal, to: :task
+  delegate :appeal, to: :task, prefix: "original"
```
leikkisa: Love this

spec/factories/appeal.rb
nanotone: Thanks for pointing this out! I think I may create a later PR to tidy up a lot of this factory stuff (after I read up on all the bells and whistles and sugar of factorybot).

There's a lot of sorta-overlapping setup code in multiple factories (appeal, task, user; not surprising when changing PDM internals) and in PDMU_spec, which could probably get a nice refactor.

spec/factories/appeal.rb
```diff
@@ -188,23 +190,27 @@
       end
     end
 
-    trait :straight_vacated do
-      stream_type { "vacate" }
-
+    trait :with_straight_vacate_stream do
```
jcq: Definitely looking forward to using this!

app/workflows/post_decision_motion_updater.rb
```diff
@@ -85,21 +101,20 @@ def handle_denial_or_dismissal
...
     AbstractMotionToVacateTask.new(
-      appeal: appeal,
+      appeal: original_appeal,
       parent: task.parent,
       assigned_to: judge_user
     )
   end
 
-  def create_new_task(parent)
+  def create_new_task(parent, appeal: vacate_stream)
```
leikkisa: This change is a little bit misleading, it's only used on the deny/dismiss flows right now with the original appeal. It doesn't seem necessary to add the appeal argument and default it to the vacate stream.

db/migrate/20200206161309_link_motions_to_appeals.rb
```diff
@@ -0,0 +1,26 @@
...
+        UPDATE post_decision_motions SET appeal_id = (
+          SELECT appeals.id FROM appeals
+            WHERE
+              appeals.stream_type = 'Vacate' AND
+              appeals.stream_docket_number = (
+                SELECT appeals.stream_docket_number
+                  FROM tasks JOIN appeals ON tasks.appeal_id = appeals.id
+                  WHERE tasks.id = post_decision_motions.task_id
```
leikkisa: I haven't tested this query, but can you access post_decision_motions.task_id without an explicit join?

db/migrate/20200206161309_link_motions_to_appeals.rb
```diff
@@ -0,0 +1,26 @@
...
+            )
+            LIMIT 1
+        )
+      EOS
```
leikkisa: I was curious so I looked at the data in prod and UAT, and the good news is there aren't any PostDecisionMotions yet, so this backfill isn't necessary atm. The upside of new feature development 😸 

It looks like in UAT the stream_docket_number didn't get fully backfilled, but prod looks good! I'll look around at the data some more and update those.

db/migrate/20200206161309_link_motions_to_appeals.rb
nanotone: Yep, the `post_decision_motions` namespace in the sub-query falls back to the record currently being `UPDATE`d.

app/workflows/post_decision_motion_updater.rb
nanotone: Yeah I don't like this either. I think I may have been juggling a changing requirement with a CodeClimate complaint at the same time. Will massage it some more.

PR #13397, opened 2020-02-12
ensure nil case_record won't break when getting rep name

rubaiyat22:
Relates #13369 
- reference [slack thread](https://dsva.slack.com/archives/CJL810329/p1581433711251600)
- use safe navigation when evalulating `case_record` for `legacy_appeal`

### DIFF
--- a/app/models/concerns/appeal_available_hearing_locations.rb
+++ b/app/models/concerns/appeal_available_hearing_locations.rb
@@ -7,7 +7,7 @@ module AppealAvailableHearingLocations
     # return the closest hearing location
     available_hearing_locations&.min_by do |loc|
       next if !loc.distance.nil?
-      
+
       loc.distance
     end
   end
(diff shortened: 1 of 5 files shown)


app/models/hearings/available_hearing_locations.rb
```diff
@@ -37,7 +37,7 @@ def formatted_facility_type
   end
 
   def determine_vba_facility_type
-    classification.include?("Regional") ? "(RO)" : "(VBA)"
+    classification&.include?("Regional") ? "(RO)" : "(VBA)"
```
rubaiyat22: Also using safe navigation here

app/models/concerns/appeal_available_hearing_locations.rb
```diff
@@ -5,6 +5,10 @@ module AppealAvailableHearingLocations
 
   def suggested_hearing_location
     # return the closest hearing location
-    available_hearing_locations&.min_by { |loc| loc.distance }
+    available_hearing_locations&.min_by do |loc|
```
rubaiyat22: added this to prevent errors in case of a nil distance

rubaiyat22:
@hschallhorn made some small changes after your approval because I noticed places where the job could potentially break

Issue #13074, opened 2019-12-30
Record ETL build stats


pkarman:
Create a `etl_builds` table with something like:

* created_at, updated_at
* statistics (json column like `distributions.statistics`)

and create a new record for each build. This way we have an easy audit trail for when data was last changed, and how many records were updated for each class.

The model class `ETL::Build` would be created by `ETL::Builder`.

rajinderbhanot:
Following tables/structures will be good to capture ETL load stats
**ETL_BATCH**  - Table to capture ETL batch load information
BATCH_ID  -  Batch ID
BATCH_NAME -  Batch Name
STARTED_AT - Batch load started at
FINISHED_AT - Batch load finished at
STATUS - Batch load status
TABLE_LOAD_CNT - Number of tables loaded
COMMENTS - Batch load comments

**ETL_JOBS** Log information at table/file level
JOB_ID -  Job ID
BATCH_ID -  Batch ID
TABLE_NAME -  Table Name
STARTED_AT -  Load started at
FINISHED_AT -  Load finished at
STATUS - Load Status
SOURCE_INFO -  Source data file or tables
ROWS_INS -  Rows Inserted
ROWS_UPD -  Rows Updated
ROWS_DEL -  Rows Deleted
ROWS_REJ - Rows Rejected
COMMENTS - Load comment/error message

LOAD_STATUSES
R | Running
C | Complete
S | Suspended
E | Error
A | Aborted

pkarman:
Ran successfully today for the first time w/o error:

```
irb(main):006:0> pp build
#<ETL::Build:0x000000000aac7268
 id: 1,
 started_at: Tue, 18 Feb 2020 18:20:41 UTC +00:00,
 finished_at: Tue, 18 Feb 2020 18:25:00 UTC +00:00,
 status: "complete",
 comments: nil,
 created_at: Tue, 18 Feb 2020 18:20:41 UTC +00:00,
 updated_at: Tue, 18 Feb 2020 18:25:00 UTC +00:00>
=> #<ETL::Build id: 1, started_at: "2020-02-18 18:20:41", finished_at: "2020-02-18 18:25:00", status: "complete", comments: nil, created_at: "2020-02-18 18:20:41", updated_at: "2020-02-18 18:25:00">
irb(main):007:0> pp build.etl_build_tables
[...]

PR #13401, opened 2020-02-12
ETL Build tables

pkarman:
connects #13074 

### Description
Adds new tables to track metadata about ETL builds, mostly timings and number of rows touched.

### Documentation Updates

### Database Changes
*Only for Schema Changes*

* [x] Column comments updated

### DIFF
--- a/spec/factories/appeal.rb
+++ b/spec/factories/appeal.rb
@@ -48,9 +48,9 @@ FactoryBot.define do
 
     transient do
       associated_attorney do
-        judge = User.find_or_create_by(css_id: "BVAAABSHIRE", station_id: 101)
+        judge = User.find_or_create_by(css_id: "BVAAABSHIRE", full_name: "BVAAABSHIRE", station_id: 101)
         judge_team = JudgeTeam.for_judge(judge) || JudgeTeam.create_for_judge(judge)
-        attorney = User.find_or_create_by(css_id: "BVAEERDMAN", station_id: 101)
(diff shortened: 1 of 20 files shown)


spec/factories/appeal.rb
```diff
@@ -48,9 +48,9 @@
 
     transient do
       associated_attorney do
-        judge = User.find_or_create_by(css_id: "BVAAABSHIRE", station_id: 101)
+        judge = User.find_or_create_by(css_id: "BVAAABSHIRE", full_name: "BVAAABSHIRE", station_id: 101)
```
pkarman: the `User.full_name` method will trigger a write via VACOLS lookup when accessed if the value is nil, so always set it (which is normal now via CSEM log in anyway).

spec/services/etl/appeal_syncer_spec.rb
```diff
@@ -73,7 +77,7 @@
       end
 
       it "syncs" do
-        subject
+        subject # no error proves the test passes
```
tomas-nava: Would the `.to_not raise_error()` pattern work here?

pkarman: yes!

I was thinking of https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/UnspecifiedException but re-read it and remembered this is the one we don't want:

```
expect { subject }.to raise_error
```

I will change.

Issue #13071, opened 2019-12-30
Motion to Vacate | Granted | Readjudicate | Decision Attorney | Go to Add Decisions screen


leikkisa:
## User or job story
User story: As a decisions attorney, after I review the vacated decision issues selected by the judge, if the vacature is a Vacate and Readjudicate, I need to be able to add new decisions.

## Acceptance criteria
- [ ] Clicking "Continue" on the "review vacated decision issues" for a **Vacate and Readjudicate** vacature type, directs the user to the "Add Decisions page" (draft_decision/dispositions)
- [ ] Update the copy for the Add Decisions page (see copy)
- [ ] Hide the "Edit" and "Delete" buttons on the already existing decision issues with the disposition of "vacated"

## Copy
Add decisions to the items listed here that have been reissued because of the motion to vacate. Any decisions you add must be connected to at least one issue on appeal. This helps Caseflow gather metrics on BVA decisions and meet AMA reporting requirements.

## Technical Notes
In existing flows, this uses a `QueueFlowPage`, which wraps the list of list of issues, which uses `AmaIssueList` component.

### Designs
![image.jpeg](https://images.zenhubusercontent.com/58f6b6eb51aa2c30577bb60e/662b1754-eefd-4fd3-bb25-af3d4da71d44)

PR #13402, opened 2020-02-13
MTV | Attorney Checkout | Add Decisions View

jcq:
Connects #13071

### Description
This implements the "Add decisions" screen of the MTV attorney checkout flow. This view only applies to `vacate_and_readjudication` vacate types.

### Acceptance Criteria
- [ ] Code compiles correctly


client/app/queue/mtv/AddDecisionIssuesView.jsx
```diff
@@ -0,0 +1,91 @@
...
+  );
+
+  const closeModals = () => setState({ ...defaultState });
+
+  const openAddIssueModal = (requestIssueId, decisionIssue) => () => {
+    const requestIssue = appeal?.issues?.find((issue) => issue.id === requestIssueId);
+    const newDecisionIssue = {
+      id: `temporary-id-${uuid.v4()}`,
+      description: '',
+      disposition: requestIssue.closed_status,
```
leikkisa: Closed status is actually different from disposition. I think it makes more sense to initialize this as not populated

client/app/queue/mtv/AddDecisionIssuesView.jsx
```diff
@@ -0,0 +1,91 @@
...
+      benefit_type: requestIssue.program,
+      diagnostic_code: requestIssue.diagnostic_code,
```
leikkisa: For this, can you add populating the diagnostic code of the new request issue in `create_contesting_request_issue!`

client/app/queue/mtv/AddDecisionIssueModal.jsx
```diff
@@ -0,0 +1,152 @@
...
+import BENEFIT_TYPES from '../../../constants/BENEFIT_TYPES';
+import DIAGNOSTIC_CODE_DESCRIPTIONS from '../../../constants/DIAGNOSTIC_CODE_DESCRIPTIONS';
+import SearchableDropdown from '../../components/SearchableDropdown';
+
+import cx from 'classnames';
+import styles from './AddDecisionIssueModal.module.scss';
+
+const isValid = ({ disposition, description }) => disposition && description;
+
+export const AddDecisionIssueModal = ({
```
leikkisa: I'm not feeling very good about re-making a new version of this component, compared to adapting the existing one. There's a lot of duplicate code and I'm worried about maintainability. I'm also worried about missing functionality for what's not duplicated that I might not completely understand (for example I don't understand the "staging" process for appeals).

I felt this way about the delete modal too, but this is much bigger so it's feeling more significant.

I can't think of any differences we want in appearance or functionality. Is the primary reason for this because of the way routing for displaying this modal is handled?  Can we explore re-using the SelectDispositionsView component instead?

client/app/queue/mtv/AddDecisionIssueModal.module.scss
```diff
@@ -0,0 +1,12 @@
+.example {
```
leikkisa: Just commenting that the modal is much narrower than it should be.

client/app/queue/mtv/AddDecisionIssuesView.jsx
jcq: That was logic copied from the existing `SelectDispositionsView`:
https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/SelectDispositionsView.jsx#L136

However, I agree that seems odd, since the values don't actually match up. I'm happy to change it to null, since `disposition` is required in the modal anyway.

client/app/queue/mtv/AddDecisionIssuesView.jsx
jcq: It looks like `diagnostic_code` for `RequestIssue` is just a method that pulls it from the contested rating issue:
https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/request_issue.rb#L506

Is that something I can actually initialize in `create_contesting_request_issue!`?

client/app/queue/mtv/AddDecisionIssueModal.jsx
jcq: I totally agree about the code duplication. I had intended to follow this up with another ticket that makes both this MTV flow and `SelectDispositionsView` use this new modal component. Even without this new MTV work, that should properly have been split out into its own component anyway instead of just being defined amidst the rest of `SelectDispositionsView`. 

I'll make sure this has everything it needs to handle all needs, and then we can decide if it makes sense to do the refactoring of `SelectDispositionsView` as part of this or break that into a separate ticket.

client/app/queue/mtv/AddDecisionIssueModal.module.scss
jcq: Yup, good catch! Fixing that to match what it was before.

client/app/queue/components/DecisionIssues.jsx
```diff
@@ -58,84 +59,103 @@ const noteDiv = css({
...
+          <div {...outerDiv} key={decisionIssue.id} className="decision-issue">
+            <div {...grayLine} />
+            <div {...decisionIssueDiv}>
+              <div {...flexContainer}>
+                Decision
+                <div>
+                  {openDeleteAddedDecisionIssueHandler && (
+                    <span>
+                      <Button
+                        name="Delete"
```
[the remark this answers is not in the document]
jcq: I could have sworn that was in there, but maybe I reverted/lost something before pushing. Anyhoo, it's implemented and working now!

jcq:
> The plan is to do another refactor in a future PR to de-duplicate the AddDecisionsModal.

The ticket for the future refactoring:
https://github.com/department-of-veterans-affairs/caseflow/issues/13503

Issue #13389, opened 2020-02-11
Same Station Conflict Check | Add CSEM Station 101 Restriction | Feature Toggle to Station Controlled


msj0nes:
### Description
Apply CSEM access station 101 restriction when handling bypass of same-station conflict check for intaking appeals. 

Currently Caseflow access is managed via CSEM (CSS) and segmented via VA stations. Station 101 is specifically dedicated to cover access for BVA. Per OIT COR, OBPI has approved BVA to bypass the '_[Same-station conflict](https://github.com/department-of-veterans-affairs/caseflow/issues/13165)_' as of 2/11/20 under the condition that the bypass only applies to users belonging to station 101. In addition to the conditions outlined in PR https://github.com/department-of-veterans-affairs/caseflow/commit/4a150bddfd48b51c097f6b38eb93ce3e3b486575, a CSEM station access based restriction needs to be applied to ensure the OBPI agreement is upheld as it relates to the issue. 

**Associated Ticket** https://github.com/department-of-veterans-affairs/caseflow/issues/13165 

### Acceptance Criteria
- [ ] Same-station conflict check remains in place for ClaimReviews (Higher-Level, Supplemental)
- [ ] Same-station conflict check is skipped for Appeals
  - [ ] Same-station conflict check is skipped only for users belonging to CSEM ‘station 101’ and on the MailTeam
- [ ] Remove the feature toggle

### Testing Plan
- [ ] A MailTeam user at station 101 can start an appeal intake, even if `station_conflict?` returns true
- [ ] A user at station 101 that is not a MailTeam user cannot start an appeal intake if `station_conflict?` returns true
- [ ] A MailTeam user not at station 101 user cannot start an appeal intake if `station_conflict?` returns true

### Tech notes
See Intake model validation

msj0nes:
@leikkisa or @nanotone let me know your thoughts on this and if anything additional is needed.

leikkisa:
@msj0nes This looks great! I'll look if there are tech notes regarding BVA permissions and the testing plan which you noted.

leikkisa:
The BvaIntake organization doesn't seem like it's being appropriately used in the code right now. It only has 3 users. I checked the last 1000 Appeals intaken, and it was done by 33 unique users at 14 different stations and 12 different regional offices.

I think before we complete this work, we should make sure that all of the appropriate users are added to the BvaIntake organization.

msj0nes:
[...]

PR #13409, opened 2020-02-13
Implement same-station bypass for MailTeam users at Station 101

nanotone:
Connects #13389 

### Description
Implement same-station bypass for intake users who are both
- MailTeam members
- at Station 101

### Testing Plan
1. Updated validator test.

### Documentation Updates
- Broke out the two user requirements (listed above) into a separate method, with comments linking to the GitHub issues where we discussed this policy implementation.


nanotone:
@leikkisa Ooh I think I _will_ be extra. Coming right up.

PR #13411, opened 2020-02-13
Fix undefined appellant address

### LINKED ISSUE #13321: Scheduling Veterans shows undefined address
## User or job story
User story: As a caseflow hearings branch user, I need to be able to see the address of the veteran when I try to schedule. The following error is displayed:

image

Link to example:
https://appeals.cf.ds.va.gov/queue/appeals/2732956/?hearingDate=2020-01-31&regionalOffice=RO13

## Acceptance criteria
- [ ] The address should be displayed on the schedule veteran screen.

## Release notes
n/a-bug fix

ferristseng:
Resolves #13321

### Description

  * Fix bug where representative's participant id is being used to fetch the appellant's address for LegacyAppeal
  * Update tests to setup veteran and appellant address data in BGS for legacy appeals
  * Remove some unused functions and tests
  * Small variable naming change to remove a lint comment


app/models/legacy_appeal.rb
```diff
@@ -703,21 +708,6 @@ def cavc
     type == "Court Remand"
   end
 
-  # Adding anything to this to_hash can trigger a lazy load which slows down
-  # welcome gate dramatically. Don't add anything to it without also adding it to
-  # the query in VACOLS::CaseAssignment.
-  def to_hash(viewed: nil, issues: nil, hearings: nil)
```
ferristseng: Can't find this being used anywhere. I'm assuming this has been replaced by the corresponding serializer

app/models/legacy_appeal.rb
```diff
@@ -963,17 +961,6 @@ def convert_file_number_to_vacols(file_number)
       fail Caseflow::Error::InvalidFileNumber
     end
 
-    # Because SSN is not accurate in VACOLS, we pull the file
-    # number from BGS for the SSN and use that to look appeals
-    # up in VACOLS
-    def vbms_id_for_ssn(ssn)
```
ferristseng: Couldn't find this being used or even tested

app/models/legacy_appeal.rb
ferristseng: Nevermind, removing this is too difficult since a lot of tests depend on it (but I don't think any code does).

client/app/queue/components/AssignHearingModal.jsx
```diff
@@ -188,13 +188,12 @@ class AssignHearingModal extends React.PureComponent {
   render = () => {
     const { appeal, openHearing } = this.props;
     const { showErrorMessages, showFullHearingDayWarning } = this.state;
-    const { address_line_1, city, state, zip } = appeal.appellantAddress || {};
+    const { address_line_1: addressLine1, city, state, zip } = appeal.appellantAddress || {};
```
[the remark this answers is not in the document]
ferristseng: This change led to some flagging by the linter

app/mappers/address_mapper.rb
```diff
@@ -28,20 +28,6 @@ def get_address_from_corres_entry(corres_entry)
...
-    {
-      address_line_1: veteran.address_line1,
-      address_line_2: veteran.address_line2,
-      address_line_3: veteran.address_line3,
-      city: veteran.city,
-      state: veteran.state,
-      country: veteran.country,
-      zip: veteran.zip_code
-    }
-  end
```
ferristseng: Removing this because it's no longer used

ferristseng:
One Q: Should we fallback to address data in VACOLS?

spec/factories/legacy_appeal.rb
```diff
@@ -4,11 +4,29 @@
...
+      if evaluator.veteran_address.present?
+        veteran = create(:veteran, file_number: appeal.sanitized_vbms_id)
+
+        (BGSService.address_records ||= {}).update(veteran.participant_id => evaluator.veteran_address)
+      end
+
+      if evaluator.appellant_address.present?
+        # Creating a veteran has a side effect of populating `BGSService.veteran_store`.
+        # BGS should be setup to return the appellant's participant ID from their SSN.
+        appellant = create(:veteran, ssn: appeal.appellant_ssn)
```
tomas-nava: I believe the way things are currently set up, `appeal.appellant_ssn` will always be `nil`. So here you're creating a Veteran object with a nil `ssn`. Either way you'll get a generated `participant_id` that you can use on the next line. But it seems like you could just use `create(:veteran)` in the next line in place of `appellant` and get the same result.

spec/models/legacy_appeal_spec.rb
```diff
@@ -1917,52 +1924,6 @@
...
-      it "includes viewed boolean in hash" do
-        expect(subject["viewed"]).to be_truthy
-      end
-
-      it "includes issues in hash" do
-        expect(subject["issues"]).to eq(issues.map(&:attributes))
-      end
-    end
-  end
-
```
tomas-nava: Can you say more about why we don't need these tests anymore?

app/models/legacy_appeal.rb
```diff
@@ -345,7 +350,7 @@ def claimant
         middle_name: appellant_middle_initial,
         last_name: appellant_last_name,
         name_suffix: appellant_name_suffix,
-        address: get_address_from_corres_entry(case_record.correspondent),
+        address: appellant_address,
```
tomas-nava: Can you say more about why we're switching from looking at the correspondent address from VACOLS to the appellant address from BGS?

app/models/legacy_appeal.rb
```diff
@@ -354,7 +359,7 @@ def claimant
         middle_name: veteran_middle_initial,
         last_name: veteran_last_name,
         name_suffix: veteran_name_suffix,
-        address: get_address_from_veteran_record(veteran) || get_address_from_corres_entry(case_record.correspondent),
+        address: appellant_address,
```
tomas-nava: What happens when the veteran and the appellant are not the same person?

spec/models/legacy_appeal_spec.rb
ferristseng: Oops, good catch. I tried removing `LegacyAppeal#to_hash` but I think it's only used in test files! I will add it back

app/models/legacy_appeal.rb
ferristseng: My understanding was the correspondent and the appellant are interchangeable judging from the implementation of `LegacyAppeal#appellant`, which acts as an alias to `LegacyAppeal#claimant`. The chain of logic I'm following is `appellant -> claimant -> correspondent`. I wasn't 100% sure if this matched existing assumptions. Either way, right now, calling `LegacyAppeal#appellant[:address]` and `LegacyAppeal#appellant_address` is returning the address from 2 separate places (and different results), which I thought was an inconsistency.

app/models/legacy_appeal.rb
ferristseng: `bgs_address_service` should be instantiated with the veteran's participant id instead of the appellant's. This will happen if `appellant_is_not_veteran` is `false`, which mirrors the condition in method.

spec/factories/legacy_appeal.rb
```diff
@@ -4,11 +4,29 @@
...
+      veteran_address { nil }
+      appellant_address { nil }
     end
 
     vacols_id { vacols_case&.bfkey || "123456" }
     vbms_id { vacols_case&.bfcorlid }
 
+    after(:create) do |appeal, evaluator|
+      if evaluator.veteran_address.present?
+        veteran = create(:veteran, file_number: appeal.sanitized_vbms_id)
```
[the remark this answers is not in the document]
ferristseng: When I looked, it seemed like the `find_or_create_by` pattern doesn't exist in FactoryBot, so I decided to just create duplicate instances. Either way, wasn't sure the best approach to take with this. I will continue to look into this.

spec/factories/legacy_appeal.rb
ferristseng: Hm, interesting. Maybe the if should check what the appellant's SSN is too. I think this depends on the SSN being set, since the BGS call relies on SSN

app/models/legacy_appeal.rb
ferristseng: It feels like there's something wrong here maybe. I don't understand why the appellant is an alias for claimant

pkarman: appellant and claimant are 2 words for the same role, afaik.

https://www.knowva.ebenefits.va.gov/system/templates/selfservice/va_ssnew/help/customer/locale/en-US/portal/554400000001018/content/554400000014081/M21-1,-Part-I,-Chapter-5,-Section-A---General-Information-on-Appeals#1a

app/models/legacy_appeal.rb
pkarman: The code looks fine.

My only point is that we are now always pulling addresses from BGS instead of VACOLS, and that may surprise some long-time VACOLS users who see one address in VACOLS and a different one in Caseflow. Just something to be aware of from a support POV.

ferristseng:
> It's not the most elegant, but maybe we could prefix with appellant_or_veteran_?

@tomas-nava Noted. I'm a little concerned that might be a bigger change since it looks like it is supposed to mirror `Appeal#appellant_*`. I'm going to hold off on it and we can make that refactor in the future if it's still confusing. I'm curious if the `Appeal` appellant code can return the veteran too, or if it only returns something if the veteran is not the appellant. We might want to make those behave similarly!

app/models/legacy_appeal.rb
ferristseng: Ok, I can notify support about this change so they know

app/models/legacy_appeal.rb
hschallhorn: @ferristseng May be worthwhile to just toss back in `|| get_address_from_corres_entry(case_record.correspondent)` for when `appellant_address` is nil

PR #13426, opened 2020-02-14
Allow nullish coalescing operator in eslint

jcq:
### Description
This provides support for the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) in ESLint.

### Background
We already had the Babel plugin enabled to support this operator, but ESLint was throwing an error upon seeing it.

This is similar in a lot of respects to using `||` to provide default values in an assignment statement without the pitfall of things like empty strings being coerced into falsey values and triggering the default.

See [this blog post](https://dev.to/bamartindev/es2020-nullish-coalescing-with-babel-plugins-1j8n) for more info and examples.

### Acceptance Criteria
- [ ] Code compiles correctly


jcq:
Closing in favor of #13797

Issue #13385, opened 2020-02-11
Motion to Vacate | Granted | Readjudicate | Decision Attorney | Submit Decisions Screen


jcq:
## Description
This is the final step/route in the checkout workflow (see #13007, #13071). This could either reuse or replace the `SubmitDecisionView` component that is used in non-MTV workflows. It serves as a final confirmation step, and handles triggering of the save to the API.

## Technical notes
The current behavior (using `SubmitDecisionView`) creates a payload using `buildCaseReviewPayload` in `utils.js`.  It then submits to the ```/case_reviews/${taskId}/complete``` endpoint. If our payload matches, we might be able to use the same one, but there's a decent chance our backend needs are different.

PR #13440, opened 2020-02-15
Implement "Submit" Step of MTV Attorney Checkout

jcq:
Connects #13385

### Description
This implements the final step of attorney checkout flow for Vacate motions. For this one, we've implemented a light wrapper around the existing `SubmitDecisionView` component; we place our updated decision issues into the Redux store and let the existing component handle it from there.

### Acceptance Criteria
- [ ] Code compiles correctly


jcq:
Ideally we should merge #13402 first, though no conflicts should occur.

PR #13441, opened 2020-02-17
Retry BGS ignorable errors

pkarman:
When background jobs encounter ignorable BGS transient errors, re-try once rather than dying immediately.

e.g. https://dsva.slack.com/archives/CN3FQR4A1/p1581919300017200

### DIFF
--- a/app/models/concerns/associated_bgs_record.rb
+++ b/app/models/concerns/associated_bgs_record.rb
@@ -55,10 +55,20 @@ module AssociatedBgsRecord
   end
 
   def bgs_record
-    @bgs_record ||= (fetch_bgs_record || :not_found)
+    @bgs_record ||= (try_and_retry_bgs_record || :not_found)
   end
 
   private
+
+  def try_and_retry_bgs_record
+    fetch_bgs_record
+  rescue BGS::ShareError => error
+    if error.ignorable?
+      fetch_bgs_record
+    else
+      fail error # re-raise if we can't try again
+    end
+  end
 
   def load_bgs_record!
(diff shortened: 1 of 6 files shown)


app/models/concerns/associated_bgs_record.rb
```diff
@@ -55,11 +55,21 @@ def accessible?
...
 
+  def try_and_retry_bgs_record
+    fetch_bgs_record
+  rescue BGS::ShareError => error
+    if error.ignorable?
+      fetch_bgs_record
+    else
+      raise error # re-raise if we can't try again
+    end
+  end
```
tomas-nava: We're not limiting the number of times we try or waiting between tries; is that cause for concern, or will stopping for non-`ignorable?` errors be a suitable safeguard?

pkarman: This method isn't recursive, so we're limited to re-trying once, automatically. Calling code can implement its own loops if it wishes (so far none does).

There's already a retry for transport/network related errors, and it does a `sleep 1`. https://github.com/department-of-veterans-affairs/ruby-bgs/blob/master/lib/bgs/base.rb#L153. So I don't think we need additional wait time here. The kinds of `ignorable?` errors that we've identified in https://github.com/department-of-veterans-affairs/ruby-bgs/blob/master/lib/bgs/errors.rb seem to not be transport-related but more specific to BGS infra hiccups.

It may very well be that this change does not effectively work around the problems because the problems are not, in fact, transient. But effectively it means N+1 tries before dying, rather than simply N, and perhaps that will help some of our long-running jobs get a little further.

Issue #13311, opened 2020-01-30
Clean up new lines in description before saving contention text


leikkisa:
## Description
When saving an issue's description, if the rating issue has new lines `\r\n` or characters like that, when we create a contention in VBMS, the description is failing to save. They are getting saved as a blank description, and then cannot be removed in the UI because blank descriptions are not allowed.

## Acceptance criteria
- [ ] Do not create contentions with newline characters in the text

## Background/context/resources
There are a couple more of these in the batteam, but here is one case:
https://dsva.slack.com/archives/CFBLH2LR4/p1580315700024000

**Example problem description:**
contested_issue_description:
  "Service connection for post traumatic stress disorder (claimed as acuired psychiatric disorder to include but not limitied to PTSD, bi-polar, depression, and anxiety due to MST) is granted with an evaluation of 70 percent effective January 31, 2017.\r\n" +
  "\r\n" +
  "An evaluation of 100 percent is assigned because of hospitalization over 21 days from March 7, 2017. (38 CFR 3.401, 38 CFR 4.29)\r\n" +
  "\r\n" +
  "An evaluation of 70 percent is assigned from May 1, 2017.",

Here's one caused by a newline in a decision issue:
"A rating higher than 0 percent from August 16, 2017 through December 27, 2017 for bilateral hearing loss is denied. \n\nA 20-percent rating effective December 28, 2017 for bilateral hearing loss is granted.\n\nA 50-percent rating effective January 14, 2019 for bilateral hearing loss is granted. "

## Technical notes
This can be done in a few places - when pulling the rating issue, saving the contested issue description, or creating the contention text.

See `Contention.text`, or RatingIssue.

leikkisa:
@nanotone FYI I'm seeing this happening originating from Caseflow too, with \n\n in a decision issue description, from the issue that Martina reported here:
https://dsva.slack.com/archives/C54QCEHAL/p1581710985442800

I think that's a potential reason for making the change in the Contention model.

nanotone:
@leikkisa Yes, I think the Contention model should be the one-stop shop for readying any contention text, whether it needs to be cleared of newlines or truncated to 255 bytes.

PR #13446, opened 2020-02-18
Handle VBMS requirement for newline-free contention text

nanotone:
Connects #13311 

### Description
This PR removes newlines from contention text before it is submitted to VBMS, which will otherwise save the text as blank.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Added a new test to check that newlines are removed
2. Existing test updated

### DIFF
--- a/spec/models/contention_spec.rb
+++ b/spec/models/contention_spec.rb
@@ -2,7 +2,7 @@
 
 describe Contention do
   let(:utf8_text) do
-    "The claim of entitlement to compensation under 38 U.S.C. § 1151 for " \
+    "The claim of entitlement\nto compensation under\r\n38 U.S.C. § 1151\rfor " \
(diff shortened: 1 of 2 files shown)


spec/models/contention_spec.rb
```diff
@@ -2,17 +2,23 @@
 
 describe Contention do
   let(:utf8_text) do
-    "The claim of entitlement to compensation under 38 U.S.C. § 1151 for " \
+    "The claim of entitlement\nto compensation under\r\n38 U.S.C. § 1151\rfor " \
```
jcq: Out of curiosity, why are all the extra characters there? Is this also stripping certain characters, or simply truncating?

(I realize that's out of the scope of this PR)

nanotone: I think you're referring to the long string of non-ASCII characters on the next line. That's for testing that the truncation method lops off whole unicode characters (which, in UTF-8 encoding, can take up more than 1 byte each) until the whole thing comes in at 255 bytes or less -- in this case it ends up being 254 bytes, since the 255th byte isn't a valid character in UTF-8.

Issue #9588, opened 2019-02-21
Deprecate DAS Tech Spec


aroltsch:
Original tech spec : https://docs.google.com/document/d/1hPTu7ihUCVhlV44tssILLQGUhO0F6j2hJXARwAD409U/edit?ts=5c6ef86a#

Reference: https://github.com/department-of-veterans-affairs/caseflow/issues/7809

AC:
1. Understand the path forward
2. Create engineering tickets where possible

# Tech Spec

## Background

Earlier this year, VA was planning to move VACOLS to a different data center since there was no clear timeline for deprecation. In response to this, the Caseflow team gave VA a ballpark timeline of March 2020. Although BVA is optimistic about Appeals Modernization Act readiness at this point, there are some signs of skepticism about the overall project of VACOLS deprecation. Some people believe that VACOLS could be around for another ten years. 

After February 14th, the Caseflow team should begin a strategic shift towards VACOLS deprecation to ensure long-term appeals program health. This lays out some thinking about the first major phase of that process, deprecating the piece of the system attorneys and judges use to review cases (DAS, or Decision Assignment Sheet) and deprecating task management from the judge step of the process to when an appeal leaves the Board (Dispatch).

Also, Queue's DAS integration is a constant thorn in our side. Some of the problems it has caused are:

1. DAS tasks do not have unique ids
2. Our frontend code is needlessly complex, handling both Caseflow and Legacy (DAS) tasks
3. While our Task model allows multiple tasks to be assigned at once, DAS does not
4. We have had several bugs with our integration (Duplicate DECASS recrods, OMO attorney credit, etc.)

Deprecating DAS will make our code simpler, more maintainable, reduce bugs, and move us one step towards our long term goal of deprecating all of VACOLS.

## Goals
* Make substantial progress on deprecating VACOLS — deprecate DAS before the end of the contract
* Maintain judge trust in and adoption of Caseflow

## Strategy/Approach
* Communicate with users to surface concerns well in advance (think 1 month) of any deprecation that affects them.
* Make very small steps at first, as confidence and trust in the process goes, make larger steps.
* Make only reversible steps.
* Where possible, always pilot with a team first.
* Don't migrate data, divert it.

### Communicate with judges to build adoption (end of Feb)

Judge trust and adoption has been an issue in the past. Judges are a senior group at the Board, and we’ll have to get some degree of buy-in from this group to move forward. So we should communicate our plans and seek feedback early by sending an email like:

[...]

PR #13457, opened 2020-02-19
OMO request is no longer supported

pkarman:
### Description
OMO no longer supported. Do not limit menu options based on that VACOLS value.

https://github.com/department-of-veterans-affairs/caseflow/issues/9588#issuecomment-492251230

https://dsva.slack.com/archives/CHX8FMP28/p1582123368372100


pkarman:
I figured we'd leave the constant defined as a way of documenting VACOLS.

Issue #13077, opened 2019-12-31
XHR response uncaught exception


pkarman:
## Description
XHR requests that have not-ok responses (e.g. 404) are caught in a general exception handling routine that generates not-actionable Sentry/Slack alerts.

Example: https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/2991/events/584914/

We should:

* [ ] handle the not-ok XHR responses in the code that generates the XHR request and not allow it to bubble up to superclient
* [ ] decide how we want 404 responses in particular to be handled: is there user action needed? bat team response? data clean up?

pkarman:
More example: https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/7774/events/008130426d174724bb989f7aef1f7f81/

Issue #13137, opened 2020-01-10
PropType Error for QueueTable


jcq:
## Description
We routinely see PropTypes-related errors in the JS console caused by `QueueTable` and its related components. This is low priority, as it doesn't impede normal usage, but we should fix so it doesn't pollute our logs and cause us to miss other errors.

## Background/context/resources
- [Slack ticket](https://dsva.slack.com/archives/CQNPUG8EQ/p1578604615001300) from @yoomlam 
- These errors can be reproduced by just loading the main Queue page for a user with cases (such as BVAAABSHIRE).

## Technical notes
`HeaderRow`, `FooterRow`, `Row`, `BodyRows`, and `QueueTable` components are all defined within the QueueTable.jsx file. However, all are set to [share the same set of PropTypes](https://github.com/department-of-veterans-affairs/caseflow/blob/master/client/app/queue/QueueTable.jsx#L598). Perhaps this shouldn't be the case?

## Example errors
```
backend.js:6 Warning: Failed prop type: The prop `rowObjects` is marked as required in `FooterRow`, but its value is `undefined`.
    in FooterRow (created by QueueTable)
    in QueueTable (created by TaskTableUnconnected)
    in TaskTableUnconnected (created by ConnectFunction)
    in ConnectFunction (created by JudgeDecisionReviewTaskListView)
    in section
    in AppSegment (created by JudgeDecisionReviewTaskListView)
    in JudgeDecisionReviewTaskListView (created by ConnectFunction)
    in ConnectFunction (created by Context.Consumer)
    in LoadingDataDisplay (created by QueueLoadingScreen)
    in div (created by QueueLoadingScreen)
    in QueueLoadingScreen (created by ConnectFunction)
    in ConnectFunction (created by Context.Consumer)
    in Route (created by PageRoute)
    in PageRoute (created by Context.Consumer)
    in withRouter(PageRoute) (created by QueueApp)
    in div (created by QueueApp)
    in main (created by AppFrame)
    in AppFrame (created by QueueApp)
    in div
    in NavigationBar (created by CaseflowNavigationBar)
    in CaseflowNavigationBar (created by Context.Consumer)
    in withRouter(CaseflowNavigationBar) (created by QueueApp)
    in Router (created by BrowserRouter)
    in BrowserRouter (created by QueueApp)
    in QueueApp (created by ConnectFunction)
    in ConnectFunction (created by Queue)
    in Provider (created by ReduxBase)
[...]

PR #13458, opened 2020-02-19
Uncaught js XHR error

pkarman:
connects #13077 

sentry error https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/7774

is an uncaught non-200 XHR response. Instead of letting it get to Raven as a generic exception, log it to the console (at least) with a more helpful error.

This happens when a hearing has been deleted in VACOLS and Caseflow is out of sync. That should be caught by an actionable data integrity checker.

Example console output:

Screen Shot 2020-02-19 at 3 29 15 PM

### DIFF
--- a/client/app/queue/QueueTable.jsx
+++ b/client/app/queue/QueueTable.jsx
@@ -604,7 +604,7 @@ export default class QueueTable extends React.PureComponent {
           rowClassNames={rowClassNames}
           bodyStyling={bodyStyling}
           {...this.state} />
-        <FooterRow columns={columns} />
+        <FooterRow rowObjects={[]} columns={columns} />
       </table>;
 
     return <div className="cf-table-wrapper" ref={(div) => {
--- a/client/app/queue/components/HearingBadge.jsx
+++ b/client/app/queue/components/HearingBadge.jsx
@@ -47,6 +47,8 @@ class HearingBadge extends React.PureComponent {
     if (!this.props.mostRecentlyHeldHearingForAppeal && !this.props.hearing && this.props.externalId) {
       ApiUtil.get(`/appeals/${this.props.externalId}/hearings`).then((response) => {
         this.props.setMostRecentlyHeldHearingForAppeal(this.props.externalId, response.body);
+      }, () => {
+        console.error(`There was an error getting hearings for ${this.props.externalId}`);
       });
     }
   }
--- a/client/app/queue/components/HearingBadge.jsx
+++ b/client/app/queue/components/HearingBadge.jsx
@@ -49,6 +49,7 @@ class HearingBadge extends React.PureComponent {
         this.props.setMostRecentlyHeldHearingForAppeal(this.props.externalId, response.body);
       }, () => {
         const error = new Error(`There was an error getting hearings for appeal ${this.props.externalId}`);
+
         if (window.Raven) {
           window.Raven.captureException(error);
         }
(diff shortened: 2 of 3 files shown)


client/app/queue/QueueTable.jsx
```diff
@@ -604,7 +604,7 @@ export default class QueueTable extends React.PureComponent {
           rowClassNames={rowClassNames}
           bodyStyling={bodyStyling}
           {...this.state} />
-        <FooterRow columns={columns} />
+        <FooterRow rowObjects={[]} columns={columns} />
```
pkarman: quiets prop warning

jcq: This should indeed quiet the props warning, but the overall issue is documented more here:
https://github.com/department-of-veterans-affairs/caseflow/issues/13137

client/app/queue/components/HearingBadge.jsx
```diff
@@ -47,6 +47,12 @@ class HearingBadge extends React.PureComponent {
     if (!this.props.mostRecentlyHeldHearingForAppeal && !this.props.hearing && this.props.externalId) {
       ApiUtil.get(`/appeals/${this.props.externalId}/hearings`).then((response) => {
         this.props.setMostRecentlyHeldHearingForAppeal(this.props.externalId, response.body);
+      }, () => {
+        const error = new Error(`There was an error getting hearings for appeal ${this.props.externalId}`);
+        if (window.Raven) {
+          window.Raven.captureException(error);
+        }
+        console.log(error);
```
jcq: The behavior is essentially the same, but if all we're doing is logging an error, it's semantically a little better to use `console.error()`

client/app/queue/components/HearingBadge.jsx
```diff
@@ -47,6 +47,12 @@ class HearingBadge extends React.PureComponent {
     if (!this.props.mostRecentlyHeldHearingForAppeal && !this.props.hearing && this.props.externalId) {
       ApiUtil.get(`/appeals/${this.props.externalId}/hearings`).then((response) => {
         this.props.setMostRecentlyHeldHearingForAppeal(this.props.externalId, response.body);
+      }, () => {
+        const error = new Error(`There was an error getting hearings for appeal ${this.props.externalId}`);
+        if (window.Raven) {
+          window.Raven.captureException(error);
+        }
+        console.error(error);
```
nanotone: This new code isn't specific to HearingBadge or even the hearings endpoint in general. It feels like we should consider pushing it into `ApiUtil` so that all XHR calls get the benefit of capture. @jcq does that seem like a long-term FEWG effort, or overstepping?

client/app/queue/components/HearingBadge.jsx
```diff
@@ -47,6 +47,12 @@ class HearingBadge extends React.PureComponent {
     if (!this.props.mostRecentlyHeldHearingForAppeal && !this.props.hearing && this.props.externalId) {
       ApiUtil.get(`/appeals/${this.props.externalId}/hearings`).then((response) => {
         this.props.setMostRecentlyHeldHearingForAppeal(this.props.externalId, response.body);
+      }, () => {
```
jcq: I tend to prefer chaining a `.catch()` as more visually straightforward than using the second argument to `.then()` as an error-handling callback.

client/app/queue/components/HearingBadge.jsx
jcq: The drawback to putting it into `ApiUtil` is just not knowing what behavior might break due to that change. While it's very tempting to handle everything in one place, it also might be better (in some cases) from a user perspective to allow the components that are making the calls to handle the errors, as they might want to display some sort of notice to the user.

This might be a good case for [error boundaries](https://reactjs.org/docs/error-boundaries.html), as it would allow us to stop errors from bubbling up past the point where we want to handle/notify user.

